PR #3872, opened 2021-07-20
Use equal operator results as arguments

qarkai:


ihhub:
@qarkai , thanks for this code cleanup!

PR #3870, opened 2021-07-19
SpellStorage doesn't inherit from std::vector anymore

vincent-grosbois:
This is cleaner because now it's impossible to directly modify the underlying std::vector
But it required several refactors in different places, so not sure it's worth it overall

PR #3878, opened 2021-07-21
Use erase-remove idiom

qarkai:
Didn't touch `src/fheroes2/spell/spell_book.cpp` since it's handled in #3870.


ihhub:
@qarkai , thank you very much for these changes!

PR #3887, opened 2021-07-22
First implementation of Artifact Effect

vincent-grosbois:
This should be the base for Artifact Effects

I listed all the possible effects in the enum, but not all of them are
implemented.
The only ones that are fully implemented (and compatible with the game)
are the ones related to luck/morale


ihhub:
Hi @vincent-grosbois and @oleg-derevenetz , if I understood correctly from our conversation we decided to go with inheritance and dynamic casting rather than with unions as the easiest to read approach. Please correct me in this.

ihhub:
We are closing this pull request because it is incomplete and the author decided not not pursue with the project development.

PR #3888, opened 2021-07-22
clang-tidy: disable cppcoreguidelines-pro-type-union-access

qarkai:
Suggested `variant` is C++17 feature. Related to #3887.


ihhub:
@qarkai , a bunch of thanks for improving our static analyser!

Issue #3509, opened 2021-05-24
Campaign, some names of bonuses don't fit the box and should be shortened


Branikolog:
The names of bonuses in "Choice" box are sometimes too long:
![image](https://user-images.githubusercontent.com/55348946/119353626-ca09c700-bcab-11eb-85a4-c4241ff4c2cd.png)
![image](https://user-images.githubusercontent.com/55348946/119332345-20b5d780-bc91-11eb-8540-c40a29c29199.png)
![image](https://user-images.githubusercontent.com/55348946/119332252-01b74580-bc91-11eb-8078-c5c4255a00f4.png)
![image](https://user-images.githubusercontent.com/55348946/119336045-c408eb80-bc95-11eb-9c42-0b7e2c82b43a.png)
![image](https://user-images.githubusercontent.com/55348946/119338603-e05a5780-bc98-11eb-95ad-deb447291f39.png)
![image](https://user-images.githubusercontent.com/55348946/119339384-c10ffa00-bc99-11eb-943a-2f2ff8d66969.png)
![image](https://user-images.githubusercontent.com/55348946/119341902-01bd4280-bc9d-11eb-80f2-9654bf5dd749.png)
![image](https://user-images.githubusercontent.com/55348946/119351874-a9d90880-bca9-11eb-8964-59cb8f7dd816.png)
![image](https://user-images.githubusercontent.com/55348946/119351987-cbd28b00-bca9-11eb-8099-407a114e3502.png)

Branikolog:
In the original "Choice" looks like:
![image](https://user-images.githubusercontent.com/55348946/119353344-70090180-bcab-11eb-99a5-45178fb1ab2c.png)
![image](https://user-images.githubusercontent.com/55348946/119353397-80b97780-bcab-11eb-95bf-e68bb94198c5.png)
![image](https://user-images.githubusercontent.com/55348946/119353448-93cc4780-bcab-11eb-823c-be8876ea8395.png)

![image](https://user-images.githubusercontent.com/55348946/119352463-659a3800-bcaa-11eb-9d59-2b64a9cc539e.png)
![image](https://user-images.githubusercontent.com/55348946/119352533-78ad0800-bcaa-11eb-8e7c-939c14f5d366.png)
![image](https://user-images.githubusercontent.com/55348946/119352575-8793ba80-bcaa-11eb-8b35-81b83961c43b.png)
[...]

PR #3891, opened 2021-07-24
Fix English phrasing and punctuation

dimag0g:
I found several cases of awkward English phrasing while working on translations, which I would like to fix:

- Replace "Fizbin Medal" with "Fizbin of Misfortune", as in the original game.
- Expand incomplete phrases like "MirrorImage created", "%{name} skipping turn" etc. to proper sentences.
- Replace "up, to" with "up to", finish victory conditions with a dot to match the loss conditions.
- A few other fixes.

### DIFF
--- a/src/fheroes2/campaign/campaign_scenariodata.cpp
+++ b/src/fheroes2/campaign/campaign_scenariodata.cpp
@@ -327,7 +327,7 @@ namespace
         case Artifact::DIVINE_BREASTPLATE:
             return _( "Breastplate" );
         case Artifact::FIZBIN_MISFORTUNE:
-            return _( "Fizbin Medal" );
+            return _( "Fizbin of Misfortune" );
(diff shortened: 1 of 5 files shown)


ihhub:
@dimag0g , thank you very much for these fixes!

src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -327,7 +327,7 @@ namespace
         case Artifact::DIVINE_BREASTPLATE:
             return _( "Breastplate" );
         case Artifact::FIZBIN_MISFORTUNE:
-            return _( "Fizbin Medal" );
+            return _( "Fizbin of Misfortune" );
```
qarkai, line 330: This was shortened to fit campaign's bonuses box, see #3509

PR #3895, opened 2021-07-24
Remove unused AlphaBlit overload

### LINKED ISSUE #3890: Suspicious conversion
https://github.com/ihhub/fheroes2/blob/cea9150e74594f0dacc48cbe9531500eec33e7dd/src/engine/image.cpp#L816
This `AlphaBlit` overload expects `uint8_t alphaValue` as 9th argument but receives value of `bool flip`. Considering that `bool flip` is 10th parameter, `bool -> uint8_t` conversion looks suspicious. If that was intentional `flip` could be explicitly casted to `uint8_t` at least.

qarkai:
Fixes #3890.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -808,14 +808,6 @@ namespace fheroes2
         }
     }
 
-    void AlphaBlit( const Image & in, const Point & inPos, Image & out, const Point & outPos, const Size & size, bool flip )
-    {
-        if ( inPos.x < 0 || inPos.y < 0 )
-            return;
-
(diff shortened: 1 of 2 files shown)


qarkai:
Hi @oleg-derevenetz, clang-tidy-pr-comments [failed](https://github.com/ihhub/fheroes2/actions/runs/1062460971) on this PR.

oleg-derevenetz:
@qarkai Yeah, it looks like if there is no warnings at all, yml file just not generated (even the empty one, like in other similar cases).

ihhub:

PR #3896, opened 2021-07-24
Don't run clang-tidy-pr-comments if there is no fixes.yml file

oleg-derevenetz:
Should fix this:

https://github.com/ihhub/fheroes2/runs/3150484374?check_suite_focus=true


ihhub:
@oleg-derevenetz , a lot of thanks for the fix!

Issue #2211, opened 2020-11-11
Add support for localizations


Baltix:
Hi,

It seems changing interface translation language in this fheroes2 fork doesn't work - I put 
lang = lt
into fheroes2.cfg but this doesn't change interface language to Lithuanian.
I've tried with fheroes2_linux_sdl2.zip ver 0.8.3

I've compiled lt.po file with msgfmt command and copied this file to ~/.fheroes2/files/lang/lt.mo
Also tried to put this file to extracted files/lang subfolder of location where I've extracted fheroes2_linux_sdl2.zip file

Lithuanian translation works fine with original fheroes2 latest SVN code from http://sourceforge.net/projects/fheroes2/

I'm attaching zipped fheroes2.cfg file
[fheroes2.cfg.zip](https://github.com/ihhub/fheroes2/files/5525729/fheroes2.cfg.zip), which changes interface language to Lithuanian with original fheroes2.cfg but doesn't work with your fork - line "lang = lt" is removed automatically when I start your fheroes2 fork :(

Thanks for improving such nice game :)

ihhub:
Hi @Baltix ,

We rewrote the whole rendering engine which right now doesn't support any translations. This is not a priority at the moment as many things should be fixed before we could even start translating the game to another languages. Such feature could be supported closer to 1.0 release or after it.

Baltix:
It's very sad to hear, that you don't have any plans for localization until 1.0 release :(
Is it so hard to add an ability to use existing translations from lang folder with new rendering engine?

P.S. I think label should be "enhancement" or "localization", not "question" ;)

ihhub:
Sveikas @Baltix ,

The playable game is more important than a buggy one but with localization like it was at Sourceforge. That's why our priority is to make the full game first and only then add a support of extra features. The problem with localization is that the old engine utilizes an external fonts with combination of SDL. We removed SDL from image generation completely except the part of rendering the image on screen. In this case the new engine doesn't support SDL readable fonts but needs some conversion from SDL images to fheroes2 image format. To make things right we need to generate needed alphabets instead of using external fonts to have the same look in the game as in the original.

On top of this we changed a lot of text in the game so old translation is not valid for many places.

ihhub:
Also, @Baltix , would you mind to support Lithuanian language localization once it's being implemented? Most likely it'll be added after New Year.

dimag0g:
[...]

PR #3897, opened 2021-07-24
Add language selection window

ihhub:
relates to #2211


ihhub:
Hi @dimag0g and @undef21 , I need your help with Russian language. Could you please check CRC32 value for Russian (Buka, isn't it?) version of the game and tell me the value so I can put into this pull request?

undef21:
Hello @ihhub.
My version of game:
CRC32:
9db30b69 heroes2.agg
7a500037 heroes2x.agg

MD5:
12175fb91b3f7714a9d99e5a1bbc69ac  heroes2.agg
a63eb93ef26965b5bc3240dc27ea09dd heroes2x.agg

SHA1:
df8df73dff4b38c758d4bece94e58d0e98099bc0   heroes2.agg
5dc379f5041f6f67b9a35b11446e05d629233813 heroes2x.agg

SHA256:
da08a14cc545f6708bd2b746edd7ea7fa0eb7d0ed26f6e54ab8e6ccb2e735e19  heroes2.agg
68f12a2ca2dd1a000e1136ad70f38b19116686c2a7a0e9a0860528d998b52afd  heroes2x.agg

ihhub:
Hi @undef21 , I was asking about CRC32 computed within the game just for font. Please make a breakpoint at 44 in ui_language.cpp file to get the value.

undef21:
> Hi @undef21 , I was asking about CRC32 computed within the game just for font. Please make a breakpoint at 44 in ui_language.cpp file to get the value.

crc32 == d5cf8af3

src/fheroes2/dialog/dialog_language_selection.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        buttonGroup.button( 0 ).press();
+        optionButtonGroup.draw();
+
+        // This text MUST be in English. No translation is applied for it.
+        const Text title( "Choose game language:", { FontSize::NORMAL, FontColor::YELLOW } );
+        title.draw( windowRoi.x + ( windowRoi.width - title.width() ) / 2, windowRoi.y + 10, display );
+
+        for ( int32_t i = 0; i < languageCount; ++i ) {
+            const Text languageName( getLanguageEnglishName( languagues[i] ), { FontSize::NORMAL, FontColor::WHITE } );
```
oleg-derevenetz, line 89: Since the language is present in `languages`, then it is supported by game assets, so we may show name of the additional language in a native language here, not in English.

src/fheroes2/system/settings.cpp
```diff
@@ -654,7 +654,7 @@ std::string Settings::String() const
     os << std::endl << "# auto combat spell casting: on/off" << std::endl;
     os << "auto spell casting = " << ( opt_global.Modes( GLOBAL_BATTLE_AUTO_SPELLCAST ) ? "on" : "off" ) << std::endl;
 
-    os << std::endl << "# game language" << std::endl;
+    os << std::endl << "# game language (empty field is for English)" << std::endl;
```
oleg-derevenetz, line 657: I'd rephrase this as "an empty value means English" :)

oleg-derevenetz (changes requested):
Hi @ihhub I left a few comments here, could you please take a look at them on occasion?

src/fheroes2/dialog/dialog_language_selection.cpp
ihhub: To display it we have to load the translation for this language and what if the translation doesn't exist?

oleg-derevenetz: If `.mo` file doesn't exist for this language, then language name will be shown in English anyway, because `Translation::current` will be `nullptr`.

BTW, if we detect that `.mo` file is not present for this language, we may just don't show the language selection dialog on the first game run. If we see that assets has some additional supported language, we try to load it first. If we fail, then nothing happens. On success, we show the language selection dialog, and, if user still selects English, we just call `Translation::reset()`.

ihhub: I've added the requested code. Please take a look.

ihhub:
Hi @oleg-derevenetz , what was the problem with Polish language? As far as I know this is the only language which uses all 255 characters.

oleg-derevenetz:
@ihhub

> what was the problem with Polish language for mo file generation? As far as I know this is the only language which uses all 255 characters.

iconv complained about illegal input sequence:

https://github.com/oleg-derevenetz/fheroes2/runs/3154349216?check_suite_focus=true

It looks like there is some wrong UTF-8 sequence in the `pl.po` file.

src/fheroes2/game/game_mainmenu.cpp
```diff
@@ -160,6 +161,12 @@ fheroes2::GameMode Game::MainMenu( bool isFirstGameRun )
     // image background
     fheroes2::drawMainMenuScreen();
     if ( isFirstGameRun ) {
+        fheroes2::SupportedLanguage supportedLanguage = fheroes2::getSupportedLanguage();
+        if ( supportedLanguage != fheroes2::SupportedLanguage::English && conf.isTranslationFilePresent( fheroes2::getLanguageAbbreviation( supportedLanguage ) ) ) {
```
oleg-derevenetz, line 165: I suppose we may just return `bool` from `Settings::setGameLanguage()`. So we may do this:

```
if ( [...] && conf.setGameLanguage( fheroes2::getLanguageAbbreviation( supportedLanguage ) ) ) {
    [...]
}
```

So if we have the appropriate translation file, we will load it automatically before showing the language menu. And if we don't have one, there is no need to show that menu. What do you think?

oleg-derevenetz (changes requested):
Hi @ihhub I left just one proposal regarding the new approach. Please take a look when you have time.

src/fheroes2/game/game_mainmenu.cpp
ihhub: Hi @oleg-derevenetz , it's done.

src/fheroes2/gui/ui_language.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    class LanguageSwitcher
+    {
+    public:
+        LanguageSwitcher() = delete;
+
+        LanguageSwitcher( const LanguageSwitcher & ) = delete;
+        LanguageSwitcher( const LanguageSwitcher && ) = delete;
+        LanguageSwitcher & operator=( const LanguageSwitcher & ) = delete;
+
```
oleg-derevenetz, line 44: You can add deleted move assignment operator here if you wish to silence clang-tidy. But it's totally optional of course.

oleg-derevenetz (approved):

src/fheroes2/gui/ui_language.cpp
ihhub: Done.

dimag0g:
> iconv complained about illegal input sequence:
> 
> https://github.com/oleg-derevenetz/fheroes2/runs/3154349216?check_suite_focus=true
> 
> It looks like there is some wrong UTF-8 sequence in the `pl.po` file. Maybe it's the consequences of the previous `sed` call though.

FIY, the position 424 where the error happens is the first UTF-8 character: `Ł` inside the line 

    "Last-Translator: Łukasz Cieliński <lukasc@mp.pl>\n"

On my system (MinGW-w64/Windows10), this line is identical in `pl.po` and `pl.po.tmp`. Would it be possible to check the contents of the `pl.po.tmp` file which produces the failure?

oleg-derevenetz:
Hi @dimag0g Here is a test example with extended logging (environment, `iconv` version, etc) and an artifact containing all the files in `files/lang` dir after executing `make`:

https://github.com/oleg-derevenetz/fheroes2/actions/runs/1083165590

I'm currently trying to figure out what's wrong here. Looks like the `pl.po` (and derived files) contain valid UTF-8, but `iconv` still complains.

oleg-derevenetz:
Hmm, it seems that `iconv` is not able to transliterate this symbol (apply `//TRANSLIT`) for some reason. If I convert this file to `utf-32`, `iconv` is able to do it.

PR #3886, opened 2021-07-21
Remove Unicode support

oleg-derevenetz:

PR #3898, opened 2021-07-24
Remove TextInterface

qarkai:
`TextInterface` is redundant since #3886.
Hide `TextAscii`.


ihhub:
@qarkai , thanks a lot for this cleanup!

PR #3907, opened 2021-07-25
Restore background music after playing short-term music effects

### LINKED ISSUE #1175: If music from world map objects was played twice, it stops current terrain theme
When player visit two times the same object (it can be Witch Hut or Troll Bridge) and hear the special sound of this object twice, terrain theme stops, until hero starts moving. Happens only with the CD music, surely, as MIDI doesn't have world map objects sounds...

oleg-derevenetz:
fix #1175


ihhub:
@oleg-derevenetz , a bunch of thanks for fixing this issue!

PR #3908, opened 2021-07-26
Code style check workflow: use system clang-format-diff

oleg-derevenetz:
No need to supply `clang-format-diff.py` in the `script/tools` directory.


ihhub:
@oleg-derevenetz thank you so much for this change!

PR #3912, opened 2021-07-26
Correct player icon in the Game Info window.

### LINKED ISSUE #3475: Scenario info window, incorrect graphics and information displayed for the scenario information window during the game compared to before the game
In fheroes2, there are incorrect graphics and information displayed for the scenario information window during the game compared to before the game.

In original game:

Wrong players' graphics 003

Wrong players' graphics 004

In fheroes2:

Wrong players' graphics 001

Wrong players' graphics 002

Arthusppp:
I fixed a bug in the game info window that incorrectly displayed the player icon for computer-only players. For more details, please read my comment in the #3475. It fixes #3475.

### DIFF
--- a/src/fheroes2/gui/player_info.cpp
+++ b/src/fheroes2/gui/player_info.cpp
@@ -190,7 +190,7 @@ void Interface::PlayersInfo::RedrawInfo( bool show_play_info ) const /* show_pla
             // comp only
             if ( fi.ComputerOnlyColors() & player.GetColor() ) {
             if ( show_play_info ) {
-                index = ( player.isPlay() ? 3 : 15 ) + Color::GetIndex( player.GetColor() );
+                index = ( player.isPlay() ? 15 : 3 ) + Color::GetIndex( player.GetColor() );
             }
             else
                 index = 15 + Color::GetIndex( player.GetColor() );
--- a/src/fheroes2/gui/player_info.cpp
+++ b/src/fheroes2/gui/player_info.cpp
@@ -188,22 +188,11 @@ void Interface::PlayersInfo::RedrawInfo( bool show_play_info ) const /* show_pla
             index = 9 + Color::GetIndex( player.GetColor() );
         else
             // comp only
-            if ( fi.ComputerOnlyColors() & player.GetColor() ) {
-            if ( show_play_info ) {
-                index = ( player.isPlay() ? 15 : 3 ) + Color::GetIndex( player.GetColor() );
-            }
-            else
-                index = 15 + Color::GetIndex( player.GetColor() );
-        }
+            if ( fi.ComputerOnlyColors() & player.GetColor() )
+            index = 15 + Color::GetIndex( player.GetColor() );
         else
-        // comp/human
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/player_info.cpp
```diff
@@ -190,7 +190,7 @@ void Interface::PlayersInfo::RedrawInfo( bool show_play_info ) const /* show_pla
             // comp only
             if ( fi.ComputerOnlyColors() & player.GetColor() ) {
             if ( show_play_info ) {
-                index = ( player.isPlay() ? 3 : 15 ) + Color::GetIndex( player.GetColor() );
+                index = ( player.isPlay() ? 15 : 3 ) + Color::GetIndex( player.GetColor() );
```
oleg-derevenetz, line 193: Hmm. According to this change, if computer-only player will be vanquished, its icon will be changed to the fully colored? I suppose it shouldn't work like that. These icons either should not depend on `isPlay()` result at all, or we should use those fancy "dead human/computer player" icons instead.

Arthusppp: Agree, I have resolved one bug and introduced a new one. 

The algorithm should not contain any player.isPlay() function here. 
The index should be:
* 3 + player color - for human or computer player
* 9 + player color - for selected/playing human
* 15 + player color - for computer only player

I will fix it shortly.

The OG in the Game Info window does not recognise if a player is playing or is vanquished. 

In the future, in a dedicated issue, we should add "_one human only player in a map and one player is about to start a game_" - in this case, the OG uses the "21" index. And, in the same way, I believe we can/should use those fancy "vanquished player" icons.

src/fheroes2/gui/player_info.cpp
```diff
@@ -188,22 +188,11 @@ void Interface::PlayersInfo::RedrawInfo( bool show_play_info ) const /* show_pla
...
         else
             // comp only
-            if ( fi.ComputerOnlyColors() & player.GetColor() ) {
-            if ( show_play_info ) {
-                index = ( player.isPlay() ? 3 : 15 ) + Color::GetIndex( player.GetColor() );
-            }
-            else
-                index = 15 + Color::GetIndex( player.GetColor() );
-        }
+            if ( fi.ComputerOnlyColors() & player.GetColor() )
```
ihhub, line 191: Could we please modify the code to be little more readable such as:
```
        
        if ( humans_colors & player.GetColor() ) {
            // Current human.
            index = 9 + Color::GetIndex( player.GetColor() );
        }
        else if ( fi.ComputerOnlyColors() & player.GetColor() ) {
            // Computer only.
            index = 15 + Color::GetIndex( player.GetColor() );
        }
        else {
            // Computer or human.
            index = 3 + Color::GetIndex( player.GetColor() );
        }
```

Sadly I can't do this myself as this is a patch-based pull request.

Arthusppp:
Withdrawing this pull request, as I have deteled the source branch. I will create a new pull request shortly (with the change described above).

PR #3918, opened 2021-07-27
AI: the first target of Chain Lighting can't be immune to it

### LINKED ISSUE #3915: Chain lightning must not touch Black Dragons
Dragons are magic resistant but somehow Chain Lightning spell goes through them which is incorrect. Load the save file, attack by Phoenixes and see what will happen.
[4_TEARS_0150.zip](https://github.com/ihhub/fheroes2/files/6886402/4_TEARS_0150.zip)

oleg-derevenetz:
fix #3915


ihhub:
@oleg-derevenetz , thanks a lot for fixing this!

Issue #1272, opened 2020-08-19
Game settings window, opinion and suggestions regarding the options


LeHerosInconnu:
I did a little study of the different options available in the "Game Settings" window.
Here is my opinion and suggestions regarding these options.

**Mandatory by default:**

- game: autosave on
Currently this option is enabled by default.

- world: use unique artifacts for secondary skills
Currently this option is not enabled by default.
This option affects artifacts related to the secondary skills Scouting, Mysticism and Necromancy.
In the original game, if the hero has multiple copies of one of these artifacts, the bonus is only counted once (hero's vision area, etc.).
**Scouting**
--Telescope: +1 square to hero's viewing range.
**Mysticism**
--Power Ring: +2 additional spell points restored to the hero (a total of 3 spell points restored with the spell point restored by the hero at the start of each turn).
**Necromancy**
--Spade of Necromancy: +10% of Skeletons generated at the end of a combat.
When this option is not activated, all the bonuses of artifacts related to these three secondary skills are counted multiple times.
This is a very interesting option but it should be possible to set it separately for each secondary skill to allow precise tuning according to the user's wishes (separate settings for Scouting, Mysticism and Necromancy). (https://github.com/ihhub/fheroes2/issues/2336)

- heroes: remember move points for retreat/surrender result
Currently this option is not enabled by default.
In the original game, the hero recovers all of his movement points when the hero is recruited.
I consider this to be a bug.
In fheroes2, it is possible to use this option to remedie this.

**Must have:**

- game: remember last focus
This option makes the last hero/castle selected when the player ends his turn automatically be selected in the next turn.

- battle: show damage info
This option can be improved, I already have some ideas. (https://github.com/ihhub/fheroes2/issues/1624)
Edit: Should be integrated into the program with the possibility to be activated/deactivated with the keyboard key "F5".

- world: show visited content from objects
Currently this option is enabled by default.
It may be desirable to leave this option enabled by default, even if it does not match how the original game works, as it is a very nice game improvement.
This option also acts, in a hidden way, on the display of the terrain penalty in the information window in the squares adjacent to the hero.
This feature should have a separate setting (https://github.com/ihhub/fheroes2/issues/2575).

- world: Wind/Water Mills and Magic Garden can be captured
I really like this option, because it removes unnecessary and tedious micromanagement from the game.

- world: scouting skill show extended content info
This option needs to be improved, and I already have a more complete idea for it. (https://github.com/ihhub/fheroes2/issues/1506)

[...]

PR #3922, opened 2021-07-29
Remove some useless experimental options

ihhub:
- world: show visited content from objects (this is enabled by default now)
- world: ban for WeekOf/MonthOf Monsters
- world: ban plagues months
- world: Months Of Monsters do not place creatures on map
- world: Starting heroes as Loss Conditions for Human Players
- world: Only 1 hero can be hired by the one player every week
- world: Each castle allows one hero to be recruited every week

relates to #1272

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -525,7 +525,6 @@ u32 * Castle::GetDwelling( u32 dw )
 
 void Castle::ActionNewWeek( void )
 {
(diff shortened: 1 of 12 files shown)


oleg-derevenetz:
Hi @ihhub it worth to remove LOSS_STARTHERO from here as well:

https://github.com/ihhub/fheroes2/blob/18dfc0b8b827ed0172484a142e3ad2fc395b2276/src/fheroes2/game/game_over.cpp#L438

PR #3924, opened 2021-07-29
Don't play ambient sounds during AI turn, move all battle sound-related stuff to Battle::Interface

### LINKED ISSUE #1189: World map structure's sound continues playing during AI turn
When we choose hero standing close to any world map structure, that has its own sound playing and press "End turn" button - the sound continues playing during whole AI turn.

oleg-derevenetz:
fix #1189


ihhub:
@oleg-derevenetz , thanks a lot for fixing this issue!

PR #3927, opened 2021-07-29
Remove unused functions

qarkai:
Found by [xunused](https://github.com/mgehre/xunused).
Skipped const counterparts and some other overloads.


src/fheroes2/game/game_newgame.cpp
```diff
@@ -269,50 +269,6 @@ fheroes2::GameMode Game::NewPriceOfLoyaltyCampaign()
     return gameChoice;
 }
 
-fheroes2::GameMode Game::NewNetwork()
```
ihhub, line 272: Please revert this change. We'll need it for the future.

src/fheroes2/monster/monster_info.cpp
```diff
@@ -605,54 +605,6 @@ namespace fheroes2
         return "";
     }
 
-    std::string getMonsterDescription( const int monsterId )
```
ihhub, line 608: Please rever this change. It's useful for tools.

src/fheroes2/ai/ai_base.cpp
```diff
@@ -37,16 +37,6 @@ namespace AI
         return "base";
     }
 
-    const char * Base::License() const
-    {
-        return "GPL-2.0";
-    }
-
-    int Base::GetPersonality() const
```
ihhub, line 45: Please revert this method. We are going to add personality very soon.

ihhub (changes requested):
Hi @qarkai , I left several remarks here. Could you please take a look?

src/fheroes2/monster/monster_info.cpp
qarkai: But there is no tools which use this functions :)

qarkai:
Those functions could be restored on demand but whatever.

src/fheroes2/monster/monster_info.cpp
ihhub: For my usage and tools :)

ihhub:
@qarkai , huge thanks for code cleanup!

PR #3928, opened 2021-07-29
GitHub Actions: add Nintendo Switch as build target

### LINKED ISSUE #3215: Add Nintendo Switch build in GitHub Actions
We have a full support of Nintendo Switch and we would like to ease life of users by adding an automated compilation of Nintendo Switch package.

oleg-derevenetz:
close #3215

Unfortunately, I have no idea how to make a package properly (now I just create a ZIP file with `fheroes2` binary file and all usual stuff) and there is no separate README for Switch, so only generic README.txt it is included in the package.


ihhub:
Hi @dimag0g , could you please take a look at this pull request?

ihhub:
> Hi @dimag0g , could you please take a look at this pull request?

@dimag0g , could you please explain why the package should be named as `*.nro`?

dimag0g:
@ihhub NRO is a "Nintendo Resource Object" which is a custom archive containing an ELF file, an icon and a few description fields. This is mostly used by homebrew SW, the eShop games usually come in NSP files (Nintendo Switch Package) which contain a mountable filesystem image. It's also possible to package homebrew in an NSP, but it only makes sense if game resources can be distributed together with the executable, which is not the case for fheores2.

Here's a `make` line I use in my `switch` branch: https://github.com/dimag0g/fheroes2/blob/switch/Makefile

oleg-derevenetz:
Hi @dimag0g

```
/opt/devkitpro/tools/bin/elf2nro $(TARGET) $(TARGET).nro --icon=heroes2.jpg --nacp=$(TARGET).nacp
```

I suppose we need an icon? I wasn't able to find `heroes2.jpg` in your repo in the `switch` branch. What are the requirements for the icon? Is it possible to make NRO without it?

dimag0g:
@oleg-derevenetz You'll need a 256x256 picture. If you don't supply one, you'll get a default icon from devkitPro which I suppose is free to use. I didn't include the [icon](https://iconarchive.com/show/mega-games-pack-28-icons-by-3xhumed/Heroes-II-of-Might-and-Magic-2-icon.html) I use in the repo because its license is not GPL-compatible.

oleg-derevenetz:
Hi @ihhub @dimag0g Now the NRO file is produced and packed into the ZIP archive with all other stuff.

> You'll need a 256x256 picture. If you don't supply one, the NRO will get a default icon from devkitPro which I suppose is free to use.

I copied `icon.png` from PSV for now (into `files/images/platform/switch`). It is 128x128, but it seems that it's OK for `elf2nro` - at least it doesn't complain. However, I don't know how it looks like on the device. I can upscale it, but if there is some source image in vector format, it is better to create an icon using it instead.

@dimag0g if you have a Switch, could you please test the produced NRO? The test release is available here:

https://github.com/oleg-derevenetz/fheroes2/releases/tag/fheroes2-switch-sdl2_dev

Also there are still no installation instructions that could be included in the ZIP package.

ihhub:
Hi @oleg-derevenetz  and @dimag0g , if we need 256x256 image we can use this:
logo_256

dimag0g:
@oleg-derevenetz The game runs on my Switch just fine. As I explained, the lowest available resolution is a bit too high and the default button mapping is a bit unusual ("A" is mouse right click, "B" is mouse left click), but the game is fully usable otherwise.

ihhub:
@oleg-derevenetz , a huge thanks for implementing this!

@dimag0g , could you please assist to complete the support of Nintendo Switch: documentation perhaps and extra resolutions like we did for PlayStation Vita?

dimag0g:
@ihhub @oleg-derevenetz Upon further testing I discovered that the game tries to use the root folder of the SD card as the home directory, so saving a game fails and language files are not loaded:

```
10:59:31: [ERROR]	open:  fheroes2.key
10:59:34: [ERROR]	setGameLanguage:  Translation file ru.mo is not found.
11:00:15: [ERROR]	open:  /files/save/Àë__åð___0001.sav
```

I remember working on this problem in my Switch releases. I will make a PR soon with the changes (custom resolution, button mapping, home directory detection and the README file I have written). For now, it's a great progress to have official Switch builds, thanks to both of you for this effort!

PR #3911, opened 2021-07-26
Highlight cells from where a monster will attack

ihhub:
close #3575

PR #3931, opened 2021-07-30
Fix an assertion call during battle shadow rendering

ihhub:
for wide monsters which are at the end of the battlefield

regression from #3911

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1647,11 +1647,23 @@ void Battle::Interface::RedrawCover()
 
             if ( _currentUnit->GetTailIndex() != -1 ) {
                 int tailDirection = _currentUnit->isReflect() ? RIGHT : LEFT;
-                const Cell * tailAttackerCell = Board::GetCell( attackerCell->GetIndex(), tailDirection );
+                const Cell * tailAttackerCell = nullptr;
+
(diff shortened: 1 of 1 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1646,8 +1646,13 @@ void Battle::Interface::RedrawCover()
             highlightCells.emplace( attackerCell );
 
             if ( _currentUnit->GetTailIndex() != -1 ) {
-                const int tailDirection = _currentUnit->isReflect() ? RIGHT : LEFT;
+                int tailDirection = _currentUnit->isReflect() ? RIGHT : LEFT;
                 const Cell * tailAttackerCell = Board::GetCell( attackerCell->GetIndex(), tailDirection );
+                if ( tailAttackerCell == nullptr ) {
```
oleg-derevenetz, line 1651: I suppose you can use Board::isValidDirection() here as well, this should look more "transparent". But it's up to you :)

oleg-derevenetz, line 1651: I suppose you can use Board::isValidDirection() here as well, this should look more "transparent". But it's up to you :)

ihhub: Fixed.

oleg-derevenetz (commented):
Hi @ihhub I simplified the logic a bit, please take a look.

PR #3932, opened 2021-07-30
iconv: fix transliteration issues for certain languages

oleg-derevenetz:
Related to #3897

I have no idea why transliteration works fine almost everywhere with `C.UTF-8` (on my macOS, on my Gentoo, on my Cygwin too) except GitHub Ubuntu, but it looks like Ubuntu need to set a language-specific locale for `iconv` to make transliteration work.


ihhub:
@oleg-derevenetz , well done! Thanks a lot for this fix!

PR #3933, opened 2021-07-31
allow player to immediately edit save file name without clicking on the input field

### LINKED ISSUE #3103: Save File list, allow to type new name without selecting the bottom bar
It would be nice it make bottom bar always in selected state, so player can press any key and change file name without pointing it.
image

eos428:
fix #3103

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -238,7 +238,7 @@ std::string SelectFileListSimple( const std::string & header, const std::string 
     fheroes2::Button buttonOk( rt.x + 34, rt.y + 315, ICN::REQUEST, 1, 2 );
     fheroes2::Button buttonCancel( rt.x + 244, rt.y + 315, ICN::REQUEST, 3, 4 );
 
-    bool edit_mode = true;
+    bool edit_mode = editor;
 
     MapsFileInfoList lists = GetSortedMapsFileInfoList();
     FileInfoListBox listbox( rt.getPosition(), edit_mode );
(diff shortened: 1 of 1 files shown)


src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -238,7 +238,7 @@ std::string SelectFileListSimple( const std::string & header, const std::string
     fheroes2::Button buttonOk( rt.x + 34, rt.y + 315, ICN::REQUEST, 1, 2 );
     fheroes2::Button buttonCancel( rt.x + 244, rt.y + 315, ICN::REQUEST, 3, 4 );
 
-    bool edit_mode = false;
+    bool edit_mode = true;
```
ihhub, line 241: This change allows to edit save files for Save file Loading screen.

ihhub (changes requested):

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -238,7 +238,7 @@ std::string SelectFileListSimple( const std::string & header, const std::string
     fheroes2::Button buttonOk( rt.x + 34, rt.y + 315, ICN::REQUEST, 1, 2 );
     fheroes2::Button buttonCancel( rt.x + 244, rt.y + 315, ICN::REQUEST, 3, 4 );
 
-    bool edit_mode = false;
+    bool edit_mode = editor;
```
ihhub, line 241: The original issue requires input field to always have an active state while this change will make it only for the time of opening the dialog and after clicking on an item the focus is gone. Also **_** symbol is not present.

ihhub (changes requested):

src/fheroes2/dialog/dialog_selectfile.cpp
eos428: Yeah you're right, we probably don't need `editMode` anymore in this case and we might as well change up the `listbox` so that it doesn't need `editMode` anymore.

ihhub:
@eos428 , huge thanks for adding this feature!

PR #3936, opened 2021-07-31
Add normal Game settings dialog

ihhub:
to change resolution, language and experimental options

image

### DIFF
--- a/src/fheroes2/gui/ui_language.cpp
+++ b/src/fheroes2/gui/ui_language.cpp
@@ -27,7 +27,6 @@
 
 #include <cassert>
 #include <map>
-#include <string>
 
 namespace
 {
--- a/src/fheroes2/gui/ui_language.cpp
+++ b/src/fheroes2/gui/ui_language.cpp
@@ -25,9 +25,7 @@
 #include "tools.h"
 #include "translations.h"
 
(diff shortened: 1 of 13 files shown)


src/fheroes2/gui/ui_language.cpp
```diff
@@ -121,4 +130,22 @@ namespace fheroes2
...
     }
+
+    SupportedLanguage getLanguageFromAbbreviation( const std::string & abbreviation )
+    {
+        if ( abbreviation.empty() ) {
+            return SupportedLanguage::English;
+        }
+
+        std::string name( abbreviation );
+        std::transform( name.begin(), name.end(), name.begin(), []( const unsigned char c ) { return std::tolower( c ); } );
```
qarkai, line 141: No need for lambda, can use `::tolower` (see `StringLower` in `tools.cpp`) or explicit overload `(int ( * )( int ))std::tolower`.

oleg-derevenetz (approved):
LGTM. Only I have the "Experimental Settings" icon associated with "please don't go here"? :)

ihhub:
> LGTM. Only I have the "Experimental Settings" icon associated with "please don't go here"? :)

Which icon? I'd love to change it if it's better.

PR #3937, opened 2021-07-31
Fix attacking after turning back in some cases

oleg-derevenetz:
Found when testing the #3931. In the following situation:

Wide units attack fail

neither of these two wide units will attack each other, they just will do one step forward instead (the same is observed when AI controls them during auto battle). This PR is intended to fix this.


ihhub:
@oleg-derevenetz , good job! Thanks for fixing it!

PR #3939, opened 2021-07-31
Remove sortTiles flag from Maps::GetAroundIndexes

qarkai:
Inline sort by distance in only caller.


ihhub:

PR #3946, opened 2021-08-01
Provide issue templates for the most common issue types

oleg-derevenetz:
People often forget to attach a screenshot or a save file. You can test these templates in my repo:

https://github.com/oleg-derevenetz/fheroes2/issues

Just click on the "New issue" button.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -7,9 +7,9 @@ body:
   attributes:
     label: Preliminary checks
     options:
-    - label: I have checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic
+    - label: I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
       required: true
-    - label: I have checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds)
+    - label: I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).
       required: true
 - type: dropdown
   attributes:
@@ -24,38 +24,16 @@ body:
     - Other
   validations:
     required: true
-- type: input
-  attributes:
-    label: Additional platform info
-    placeholder: Describe your platform if you chose "Other" in the previous section.
 - type: textarea
   attributes:
     label: Describe the bug
-    placeholder: Provide a clear and concise description of what the bug is.
-  validations:
-    required: true
-- type: textarea
-  attributes:
-    label: Steps to reproduce
-    placeholder: |
-      Steps to reproduce the bug:
-      1. Go to ...
-      2. Click on ...
-      3. See error ...
-  validations:
-    required: true
-- type: textarea
-  attributes:
-    label: Expected behavior
-    placeholder: Provide a clear and concise description of what you expected to happen.
-  validations:
-    required: true
-- type: textarea
-  attributes:
-    label: Screenshots
     placeholder: >
-      If possible, provide screenshots to illustrate the issue.
-      You can add them by dragging and dropping the image file to this section.
+      Provide a clear and concise description of what the bug is, including
+      the list of steps to reproduce the bug. If possible, provide screenshots
+      to illustrate the issue. You can add screenshots by dragging and dropping
+      the image file to this section.
+  validations:
+    required: true
 - type: textarea
   attributes:
     label: Save file
@@ -63,7 +41,12 @@ body:
       If possible, provide a save file to make it easier for developers to
       reproduce the issue. You can add it by placing your save file into a
       ZIP archive and then dragging and dropping the resulting archive file
-      into this section.
+      into this section. If you do not want to provide a save file (for example,
+      because it is not required to reproduce the issue or is not available),
+      please indicate here the reason why the save file cannot be attached to
+      this bug report.
+  validations:
+    required: true
 - type: textarea
   attributes:
     label: Additional info
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,60 @@
+name: Bug report
+description: Something is not working or is going wrong
+labels: [ bug ]
+
+body:
+- type: dropdown
+  attributes:
+    label: Platform
+    description: Choose your platform
+    options:
+    - Windows
+    - macOS
+    - Linux
+    - Other
+  validations:
+    required: true
+- type: input
+  attributes:
+    label: Additional platform info
+    placeholder: Describe your platform if you chose "Other" in the previous section.
+- type: textarea
+  attributes:
+    label: Describe the bug
+    placeholder: Provide a clear and concise description of what the bug is.
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Steps to reproduce
+    placeholder: |
+      Steps to reproduce the bug:
+      1. Go to ...
+      2. Click on ...
+      3. See error ...
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Expected behavior
+    placeholder: Provide a clear and concise description of what you expected to happen.
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Screenshots
+    placeholder: >
+      If possible, provide screenshots to illustrate the issue.
+      You can add them by dragging and dropping the image file to this section.
+- type: textarea
+  attributes:
+    label: Save file
+    placeholder: >
+      If possible, provide a save file to make it easier for developers to
+      reproduce the issue. You can add it by placing your save file into a
+      ZIP archive and then dragging and dropping the resulting archive file
+      into this section.
+- type: textarea
+  attributes:
+    label: Additional info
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,70 @@
...
+labels: [ bug ]
+
+body:
+- type: checkboxes
+  attributes:
+    label: Preliminary checks
+    options:
+    - label: I have checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic
+      required: true
+    - label: I have checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds)
```
ihhub, line 12: Could we please add a full stop at the end of each sentence? It'll feel like a complete statement.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,35 @@
+name: Feature request
+description: New feature, request or improvement
+labels: [ improvement ]
+
+body:
+- type: checkboxes
+  attributes:
+    label: Preliminary checks
+    options:
+    - label: I have checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic
```
ihhub, line 10: Could we please add a full stop at the end of this sentence as well?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,70 @@
...
+    - Linux
+    - PlayStation Vita
+    - Nintendo Switch
+    - Other
+  validations:
+    required: true
+- type: input
+  attributes:
+    label: Additional platform info
+    placeholder: Describe your platform if you chose "Other" in the previous section.
```
ihhub, line 30: I'm looking from an ordinary user view: this field looks useless for most of people. We don't have any other platforms which we (officially) support. Also we haven't seen any so far :) I'd recommend to remove this field to keep things simple.

.github/ISSUE_TEMPLATE/bug_report.yml
ihhub: [very minor thing] we're mixing formal and informal language: it should be `I've` :)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,70 @@
...
+- type: textarea
+  attributes:
+    label: Steps to reproduce
+    placeholder: |
+      Steps to reproduce the bug:
+      1. Go to ...
+      2. Click on ...
+      3. See error ...
+  validations:
+    required: true
```
ihhub, line 46: I'm suggesting to combine these 2 fields. I know that description and steps to reproduce aren't the same but the more fields we have the more reluctant people would be to open issues. I'm not saying that it should be this way. I'm just suggesting so I'm happy for a discussion.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,70 @@
...
+      If possible, provide screenshots to illustrate the issue.
+      You can add them by dragging and dropping the image file to this section.
+- type: textarea
+  attributes:
+    label: Save file
+    placeholder: >
+      If possible, provide a save file to make it easier for developers to
+      reproduce the issue. You can add it by placing your save file into a
+      ZIP archive and then dragging and dropping the resulting archive file
+      into this section.
```
ihhub, line 66: Should we combine these fields too? Something like `Attachments`? Many players struggle to find save files (based on requests in GitHub and Discord).

ihhub (changes requested):
Hi @oleg-derevenetz , I left few comments and suggestions here. Could you please take a look?

ihhub:
Hi @Branikolog could you please check this pull request as well and give your opinion as a user?

.github/ISSUE_TEMPLATE/bug_report.yml
oleg-derevenetz: Should we have the "Other" item in the list of platforms then?

.github/ISSUE_TEMPLATE/bug_report.yml
oleg-derevenetz: You see, the whole point of this form is to encourage people to provide as many information as possible. People often are not aware of importance of screenshots and save files, and they may have no idea how to attach them. That's why all these separate inputs. "Attachments" looks too generic and doesn't encourage people to attach something specific.

.github/ISSUE_TEMPLATE/bug_report.yml
ihhub: Yes, we should keep it.

.github/ISSUE_TEMPLATE/bug_report.yml
ihhub: Solid reasons. Let's keep it as it is.

Branikolog:
Hi, @oleg-derevenetz and @ihhub 
I love the "preliminary check" section. It will really help to prevent appearing already fixed bugs.
Not sure that all "Describe the bug", "Steps to reproduce" and "Expected behavior" fields must be compulsory... Sometimes there are no required steps to reproduce the bug. We could leave all three these sections, so people know basic steps for describing the problem, but lets allow them make it in more free way, rather than restrict with three these fields.
Talking about savefiles and screenshots fields. It's much more useful, when we can place screenshot or file next to certain description in text. Nevertheless, the way of attaching them is a useful information (Which is written currently on default background in these fields). So I prefer this information (Drag-n-drop pics and attaching zipped files) to be simply written above description field, so issue creator would notice it and add later, if possible. 
I don't know, if it's possible, but it would be nice to add two buttons ("Add screenshot" and "Add file") that would place a needed file at cursor's position in text. (Probably showing pop-up dialog, with the rules of adding files into a field)

oleg-derevenetz:
Hi @ihhub and @Branikolog I have made an effort to address your comments and reduce the number of form inputs, please take a look. Also, I made the "Save file" mandatory in the "Bug report" form - user will need either to provide a savefile, or describe why the savefile is absent.

> We could leave all three these sections, so people know basic steps for describing the problem, but lets allow them make it in more free way, rather than restrict with three these fields.

I made only one section instead with a descriptive placeholder. Hope that this will be enough.

> So I prefer this information (Drag-n-drop pics and attaching zipped files) to be simply written above description field, so issue creator would notice it and add later, if possible.

These descriptions (even if they will be added above the input field) will not be shown after issue is already submitted. Only regular issue interface will be available. All this decorations (inputs, selection boxes, descriptions, etc) are available only on initial submit.

> I don't know, if it's possible, but it would be nice to add two buttons ("Add screenshot" and "Add file") that would place a needed file at cursor's position in text. (Probably showing pop-up dialog, with the rules of adding files into a field)

It is not currently possible to add buttons or popups to this form - I suppose because all this implies having some code (which will respond to button clicks or show popups), only basic codeless inputs are currently available.

Branikolog:
Hi, @oleg-derevenetz 

> > So I prefer this information (Drag-n-drop pics and attaching zipped files) to be simply written above description field, so issue creator would notice it and add later, if possible.
> 
> These descriptions (even if they will be added above the input field) will not be shown after issue is already submitted. Only regular issue interface will be available. All this decorations (inputs, selection boxes, descriptions, etc) are available only on initial submit.
> 

I meant, that an issue creator, after he read the description of the way to add screenhot or savefile, later, while he will typing some text in a field, will have in mind, that addition of savefiles and screenshots is possible.

Anyway, the way you've done is really nice!

One more question regarding these fields. Default text is nice and clear, but I think, there're too much words. Is it possible to make bold or underline some key words in this text to point issue creator the way he should act? For example, in "Description of the bug" field we can highlight such words as "description", "list of steps to reproduce" and "screenshots".
(Also, we can highlight "ZIP archive" in the second field, so creator would instantly notice what he is required to do.)

I've forgot to inspect other buttons on issue creation, before. :)

Talking about "Feature request" option I would like to watch there a small tip for people about our imprevement policy. To prevent people from requesting "zounds" of new low-quality castles/creatures or inappropriate features...
Something short, like: "Current project improvement policy is mostly concentrated on making gameplay more convinient and easy-to-interact. We would like to develop features, that make the game more user-friendly but not overflow overall interface. Be sure, that the feature you're going to request will be maximum useful for the majority of engine-users/developers"

oleg-derevenetz:
@Branikolog

> Default text is nice and clear, but I think, there're too much words. Is it possible to make bold or underline some key words in this text to point issue creator the way he should act?

It is not possible in placeholders, they do not support formatting. This may be done in descriptions (outside of text fields), but I would not want to transfer all these texts to descriptions precisely because they are outside of text fields :) This will bloat the form, and explanations in placeholders are pretty simple, there is no need to have them in front of your eyes all the time. May be the "Save file" is the only exception - however, its only task is to ensure that the user will upload the save file, so there is no need to write some long and complex things in this field.

> Talking about "Feature request" option I would like to watch there a small tip for people about our improvement policy.

I added something like a warning to it.

Branikolog:
@oleg-derevenetz 

> I added something like a warning to it.

It's perfect! )

ihhub:
Hi @oleg-derevenetz and @Branikolog , are we good with these changes?

ihhub (approved):

oleg-derevenetz:
Hi @ihhub Well, all I may say is that everything should work here :)

ihhub:
@oleg-derevenetz , great! Then let's go with this solution.

PR #3950, opened 2021-08-02
Battle::Board::GetPath(): minor optimization

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks so much for this small optimization!

PR #3951, opened 2021-08-02
Make captured object event messages to contain yellow titles

### LINKED ISSUE #2154: Adventure map, message windows, option "world: Wind/Water Mills and Magic Garden can be captured", display the title in yellow
I suggest to change the title color to yellow for these windows.

In fheroes2:

Plural title window fh2 01

Plural title window fh2 02

Plural title window fh2 03

Similar issue: https://github.com/ihhub/fheroes2/issues/1919.

ihhub:
close #2154


src/fheroes2/game/game_io.cpp
```diff
@@ -200,18 +188,24 @@ fheroes2::GameMode Game::Load( const std::string & fn )
...
     fz >> binver;
 
     // check version: false
     if ( binver > CURRENT_FORMAT_VERSION || binver < LAST_SUPPORTED_FORMAT_VERSION ) {
-        std::ostringstream os;
-        os << "usupported save format: " << binver << std::endl
-           << "game version: " << CURRENT_FORMAT_VERSION << std::endl
-           << "last supported version: " << LAST_SUPPORTED_FORMAT_VERSION;
-        Dialog::Message( "Error", os.str(), Font::BIG, Dialog::OK );
+        std::string warningMessage( _( "Usupported save format: " ) );
```
oleg-derevenetz, line 198: I believe it's more `errorMessage` than `warningMessage` :) But it's just a nitpick.

src/fheroes2/world/world.cpp
```diff
@@ -1533,10 +1533,17 @@ bool EventDate::isAllow( int col, u32 date ) const
...
-    return msg << obj.resource << obj.computer << obj.first << obj.subsequent << obj.colors << obj.message;
+    return msg << obj.resource << obj.computer << obj.first << obj.subsequent << obj.colors << obj.message << obj.title;
 }
 
 StreamBase & operator>>( StreamBase & msg, EventDate & obj )
 {
-    return msg >> obj.resource >> obj.computer >> obj.first >> obj.subsequent >> obj.colors >> obj.message;
+    msg >> obj.resource >> obj.computer >> obj.first >> obj.subsequent >> obj.colors >> obj.message;
+
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_096_RELEASE, "Remove temp variable usage here." );
```
oleg-derevenetz, line 1543: I suppose that it should look like "remove this check", because there is no any temp variable.

oleg-derevenetz (changes requested):
Hi @ihhub I left a few minor comments, please take a look when it will be convenient for you.

Also, may be turn off the `cppcoreguidelines-pro-bounds-constant-array-index` warning for the `clang-tidy`? Non-constant array subscripts are widely used in this project (they may be eventually replaced by `std::vector` or `std::array` but that's another story), and we are definitely not going to use GSL (are we?).

ihhub:
> Hi @ihhub I left a few minor comments, please take a look when it will be convenient for you.
> 
> Also, may be turn off the `cppcoreguidelines-pro-bounds-constant-array-index` warning for the `clang-tidy`? Non-constant array subscripts are widely used in this project (they may be eventually replaced by `std::vector` or `std::array` but that's another story), and we are definitely not going to use GSL (are we?).

Hi @oleg-derevenetz , no, we aren't going to use `std::gsl` code. It looks unnatural. I think we have to run off the check about array indexes for now.

oleg-derevenetz (approved):

PR #3952, opened 2021-08-02
Remove duplicated arena size defines

qarkai:
Defined in `battle_board.h`.


ihhub:

PR #3953, opened 2021-08-02
Nintendo Switch build fixes

dimag0g:
- add custom 848x480 resolution, fix touchscreen scaling
- hardcode home directory as `/switch/fheroes2`
- convert the icon file to JPG (PNG is not working)
- custom controls layout, described in the Readme file


docs/README_switch.md
```diff
@@ -0,0 +1,61 @@
...
+         +--- maps
+         +--- music
+         +--- fheroes2.nro
+
+
+[3] RUNNING
+-----------
+
+This build of Free Heroes 2 was tested on 12.0.3|AMS M.19.4|S (FAT32). exFAT is not recommended.
+It's possible to use USB mice / keyboards with this build.
```
ihhub, line 44: Could we rephrase this sentence into `USB mice and keyboards are supported too.`?

ihhub (changes requested):
Hi @dimag0g , I left multiple comments here. Could you please take a look?

ihhub:
@dimag0g , can't the icon have transparent background? If not I can provide an image with a proper background in this case.

Makefile.switch
```diff
@@ -25,16 +25,21 @@ PROJECT_TITLE := fheroes2
 PROJECT_AUTHOR := fheroes2 resurrection team
 PROJECT_VERSION := 0.9.5
 
+PLATFORM = switch
```
oleg-derevenetz, line 28: Is this variable used anywhere? The sub-`make` command receives separate `PLATFORM=switch` argument.

Makefile.switch
```diff
@@ -25,16 +25,21 @@ PROJECT_TITLE := fheroes2
...
 
 $(TARGET).nro: $(TARGET).elf $(TARGET).nacp
-	elf2nro $(TARGET).elf $(TARGET).nro --icon=files/images/platform/switch/icon.png --nacp=$(TARGET).nacp
+	elf2nro $(TARGET).elf $(TARGET).nro --icon=files/images/platform/switch/icon.jpg --nacp=$(TARGET).nacp
 
+.PHONY: $(TARGET).nacp
 $(TARGET).nacp:
 	nacptool --create "$(PROJECT_TITLE)" "$(PROJECT_AUTHOR)" "$(PROJECT_VERSION)" $(TARGET).nacp
 
+.PHONY: $(TARGET).elf
```
oleg-derevenetz, line 42: Why the `$(TARGET).nacp` and `$(TARGET).elf` are declared as phony targets? These are real files and they can be tracked by `make` as usual. Is there some reason I don't know about?

oleg-derevenetz (changes requested):
Hi @dimag0g I left a few comments, could you take a look when it is convenient for you?

src/engine/localevent.cpp
```diff
@@ -216,6 +216,28 @@ namespace
...
+#undef SDL_CONTROLLER_BUTTON_DPAD_LEFT
+#undef SDL_CONTROLLER_BUTTON_DPAD_RIGHT
+#undef SDL_CONTROLLER_BUTTON_DPAD_UP
+#undef SDL_CONTROLLER_BUTTON_DPAD_DOWN
+#define SDL_CONTROLLER_BUTTON_A 1
+#define SDL_CONTROLLER_BUTTON_B 0
+#define SDL_CONTROLLER_BUTTON_DPAD_LEFT 13
+#define SDL_CONTROLLER_BUTTON_DPAD_RIGHT 14
+#define SDL_CONTROLLER_BUTTON_DPAD_UP 11
+#define SDL_CONTROLLER_BUTTON_DPAD_DOWN 12
```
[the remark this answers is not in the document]
dimag0g: The purpose of these `#undef/#define` is to redefine existing constants, in order to keep intact the code that uses them. I'm pretty sure redefining them as enumeration will not work.

Makefile.switch
dimag0g: `$(TARGET).nacp` depends on `$(PROJECT_VERSION)` and other variables. It will not be rebuild if these variables change, unless declared as `.PHONY`. As an option, it could be declared to depend on the file which defines the current version, e.g. `Makefile.switch`.

`$(TARGET).elf` is a copy of `src/dist/$(TARGET).elf`, which in turn depends on all `.cpp`/`.h` files inside `src`. It will not be rebuilt when the source code changes, unless the sub-make is run unconditionally.

Is there a better way to force `make` targets to always be rebuilt?

oleg-derevenetz: Yeah, fair point - they depend on non-files and files that are not tracked by this Makefile.

> As an alternative, we could simply list all the commands inside an existing .PHONY target, all

Yes, it will look better this way. BTW, we can list `clean` as phony as well.

oleg-derevenetz: BTW, could you please implement similar changes in `Makefile.vita` as well? `$(TARGET).elf` depends on external files here and `param.sfo` depends on variables.

docs/README_switch.md
```diff
@@ -0,0 +1,61 @@
+This is a port of Free Heroes 2 (free implementation of Heroes of Might and Magic 2 engine) to Nintendo Switch.
```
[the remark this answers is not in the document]
dimag0g: Should I include `README_switch.md` along with `README.txt`, or should I only include `README_switch` and add missing parts of `README.txt` to it?

Makefile.switch
dimag0g: @ihhub I've tried to implement the requested fixes. I left the `#undef/#define` part unchanged, modified the Makefile.switch to run the packaging commands unconditionally, and replaced the icon with a white background version. I never managed a PNG icon to work on a Switch, so keeping the transparent background is not possible AFAIK.

Unfortunately, I will be unavailable for the next few days. If there's a trivial change required which you could do yourself, it might be better to do so instead of waiting for me to respond.

dimag0g:
@ihhub I've tried to implement the requested fixes. I left the `#undef/#define` part unchanged, modified the Makefile.switch to run the packaging commands unconditionally, and replaced the icon with a white background version. I never managed a PNG icon to work on a Switch, so keeping the transparent background is not possible AFAIK.

Unfortunately, I will be unavailable for the next few days. If there's a trivial change required which you could do yourself, it might be better to do so instead of waiting for me to respond.

docs/README_switch.md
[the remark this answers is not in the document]
ihhub: Both files should be present like you did.

oleg-derevenetz:
Hi @ihhub are you sure you want to delete JPG and not PNG?

ihhub:
> Hi @ihhub are you sure you want to delete JPG and not PNG?

Yes, I wanted to replace JPEG but I couldn't so I made another commit to add JPEG file. Now it's in master branch.

ihhub (approved):

ihhub:
Hi @oleg-derevenetz , do you have any more comments about this pull request?

oleg-derevenetz:
Hi @ihhub No, I suppose we can merge this. We can make a separate PRs then if needed.

ihhub:
@dimag0g , a huge thanks for these changes!

PR #3955, opened 2021-08-02
Fixup .PHONY targets in makefiles

oleg-derevenetz:


ihhub:

PR #3956, opened 2021-08-03
Convert Kingdoms::size() method to static member

qarkai:
Use range-based for to iterate over kingdoms.

### DIFF
--- a/src/fheroes2/kingdom/kingdom.h
+++ b/src/fheroes2/kingdom/kingdom.h
@@ -223,13 +223,12 @@ public:
 
     void AddTributeEvents( CapturedObjects & captureobj, const uint32_t day, const int objectType );
 
-    u32 size( void ) const;
-
 private:
     friend StreamBase & operator<<( StreamBase &, const Kingdoms & );
     friend StreamBase & operator>>( StreamBase &, Kingdoms & );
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/kingdom/kingdom.h
```diff
@@ -223,13 +223,12 @@ class Kingdoms
...
     void AddTributeEvents( CapturedObjects & captureobj, const uint32_t day, const int objectType );
 
-    u32 size( void ) const;
-
 private:
     friend StreamBase & operator<<( StreamBase &, const Kingdoms & );
     friend StreamBase & operator>>( StreamBase &, Kingdoms & );
 
-    Kingdom kingdoms[KINGDOMMAX + 1];
+    static constexpr uint32_t size = KINGDOMMAX + 1;
```
ihhub, line 230: To separate members of the class/structure from the rest of variable we usually start the name from `_` symbol.

ihhub (commented):
Hi @qarkai , I left just one comment. If you think it's not worth to change then let's go with these changes.

src/fheroes2/kingdom/kingdom.h
qarkai: Ah, right!

ihhub:

PR #3960, opened 2021-08-03
Add links to README_switch.md in INSTALL.md and DEVELOPMENT.md, rename "PS Vita" to "PlayStation Vita"

oleg-derevenetz:
As usual, you can see how it looks like in the website mode here:

Also I added links to PSV and Switch-specific READMEs to INSTALL.md (they still contain useful data about which data should be copied to where, although this does not apply to devices with official firmware), if it is better not to do this, I can remove them.


ihhub:
@oleg-derevenetz , good job!

PR #3962, opened 2021-08-03
Remove extra class specifiers in static member function calls, ERROR_LOG already prints __FUNCTION__, so we shouldn't print function name once again

oleg-derevenetz:


ihhub:

PR #3963, opened 2021-08-03
Clang-Tidy: disable cppcoreguidelines-pro-bounds-array-to-pointer-decay

oleg-derevenetz:
Reason: `char *` is widely used in this project to represent null-terminated strings, and `char[]` literals are often passed as arguments to functions that accept `char *`. At the same time we are not going to use GSL yet, so `gsl::span` is not available anyway, and neither `std::span` (from C++20) nor `std::string_view` (from C++17) are available.


ihhub:

PR #3965, opened 2021-08-04
Add language choice to Windows installer

qarkai:
Screenshot 2021-08-04 141326

### DIFF
--- a/script/windows/fheroes2.iss
+++ b/script/windows/fheroes2.iss
@@ -19,6 +19,20 @@ OutputDir={#BuildDir}
 ArchitecturesInstallIn64BitMode=x64
 #endif
 
+[Components]
+Name: "lang"; Description: "Localizations"; Types: full
+Name: "lang\pt"; Description: "Brazilian Portuguese"; Types: full
+Name: "lang\cs"; Description: "Czech"; Types: full
+Name: "lang\nl"; Description: "Dutch"; Types: full
+Name: "lang\fr"; Description: "French"; Types: full
+Name: "lang\hu"; Description: "Hungarian"; Types: full
+Name: "lang\lt"; Description: "Lithuanian"; Types: full
+Name: "lang\pl"; Description: "Polish"; Types: full
+Name: "lang\ru"; Description: "Russian"; Types: full
+Name: "lang\es"; Description: "Spanish"; Types: full
+Name: "lang\sv"; Description: "Swedish"; Types: full
+Name: "lang\tr"; Description: "Turkish"; Types: full
+
 [Files]
 Source: "{#BuildDir}\{#AppName}.exe"; DestDir: "{app}"
 Source: "{#BuildDir}\lib*.dll"; DestDir: "{app}"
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @qarkai , I have a worry that this might mislead users thinking that once localization is installed fheroes2 can support these languages.

qarkai:
Hi @ihhub, so what's your suggestion then?

ihhub:
> Hi @ihhub, so what's your suggestion then?

We could keep as it is for now and introduce such change once we have a full support of all languages even those which were not officially supported.

@oleg-derevenetz , what do you think?

qarkai:
@ihhub then why do we install them if we don't support these languages?

oleg-derevenetz:
> > Hi @ihhub, so what's your suggestion then?
> 
> We could keep as it is for now and introduce such change once we have a full support of all languages even those which were not officially supported.
> 
> @oleg-derevenetz , what do you think?

I agree. People might think that as long as they are selected French or Russian language, they will get the whole game in that language, but they are unaware that they will need also localized versions of game assets, and this may be surprising.

qarkai:
No comment, no option rename, nothing, just "leave as it is"? I remind that installer brings a bunch of unneeded files.

> they are unaware that they will need also localized versions of game assets
Let's make people aware what they need.

oleg-derevenetz:
> @ihhub then why do we install them if we don't support these languages?

We have, but only partially - we still need at least fonts from official assets. Also IIRC there are images with pre-rendered localized text on them in these assets.

oleg-derevenetz:
> @qarkai , we have a partial support of Polish, Russian and French languages. The rest of languages right now are not in use. If the problem is in extra files let's just copy these 2 languages for now but without the menu to select languages.

People still may edit fheroes2.cfg by hand and get another language... Well, sort of :)

ihhub:
> > @qarkai , we have a partial support of Polish, Russian and French languages. The rest of languages right now are not in use. If the problem is in extra files let's just copy these 2 languages for now but without the menu to select languages.
> 
> People still may edit fheroes2.cfg by hand and get another language... Well, sort of :)

-_- That's why I suggested to leave as it is for now :)

ihhub:
I'm moving this pull request to 1.0 release as we still don't have an ability to support multiple languages (for now) based on game's assets. I have a plan to generate language assets based on English letters but it'll take some time.

ihhub:
I converted this pull request to Draft for now as we haven't agreed on the solution. Once we add support of other languages out of box we can add this change.

ihhub:
Soon we're going to be able to support all languages so the question is whether we really want to allow users to select certain languages? I think in the current era of high volume discs 1 MB shouldn't matter much.

@qarkai  and @oleg-derevenetz , what do you guys think?

oleg-derevenetz:
@ihhub

> Soon we're going to be able to support all languages so the question is whether we really want to allow users to select certain languages? I think in the current era of high volume discs 1 MB shouldn't matter much.

I believe this is not necessary.

qarkai:
@ihhub I don't care. I'm tired of team's attitude.

ihhub:
> @ihhub I don't care. I'm tired of team's attitude.
Oh, that's sad to hear. I didn't know that something went wrong but still thanks a lot for your contribution to the project.

ihhub:
We are closing this pull request as we greatly expanded the list of languages by default.

PR #3969, opened 2021-08-04
Ignore .mo files

qarkai:


ihhub:
@qarkai , a bunch of thanks for this change!

PR #3972, opened 2021-08-05
Properly handle cross-moat attacks

### LINKED ISSUE #2589: Combat, troops can be attacked, while standing right behind the moat
Troops, that enter moat shouldn't be able to attack enemy behind the moat. This can happen only if attacking troop spends one turn standing in a moat. 
image

oleg-derevenetz:
fix #2589


oleg-derevenetz:
BTW, here is an example of not quite the correct logic of an AI pathfinder:

Before Berserk Turn

The Crusader here is under the Berserk spell and should attack Unicorn in a straight line from the left (from cell 95), because this is the only reachable cell for the Crusader near the Unicorn (and without this PR he could even immediately attack the Unicorn!). But he choose to attack it from the cell 86 instead, because `_pathfinder.hexIsPassable()` (called from the `Arena::hexIsPassable()`) returns `true` for it! As a result, his path is eventually corrected by the `FindNearestReachableCell()` and he ends his journey here:

After Berserk Turn

src/fheroes2/battle/battle_board.h
```diff
@@ -75,6 +75,8 @@ namespace Battle
         static bool isMoatIndex( s32 index, const Unit & b );
         static bool isBridgeIndex( s32 index, const Unit & b );
         static bool isOutOfWallsIndex( s32 );
+        static bool isIndexToTheLeftOfTheMoat( int32_t index );
```
qarkai, line 78: Could be private.

src/fheroes2/battle/battle_board.h
```diff
@@ -117,6 +118,9 @@ namespace Battle
         bool GetPathForWideUnit( const Unit & unit, const Position & destination, const uint32_t remainingSteps, const int32_t currentHeadCellId,
                                  const int32_t prevHeadCellId, std::vector<bool> & visitedCells, Indexes & result ) const;
         void StraightenPathForUnit( const int32_t currentCellId, Indexes & path ) const;
+
+        static bool isIndexToTheLeftOfTheMoat( int32_t index );
+        static bool isIndexToTheRightOfTheMoat( int32_t index );
```
qarkai, line 123: Also `index` should be `const` according to code style.

src/fheroes2/battle/battle_board.cpp
```diff
@@ -709,6 +708,19 @@ bool Battle::Board::isMoatIndex( s32 index, const Unit & b )
...
+    return ( ( index <= 6 ) || ( 11 <= index && index <= 17 ) || ( 22 <= index && index <= 27 ) || ( 33 <= index && index <= 38 ) || ( 44 <= index && index <= 48 )
+             || ( 55 <= index && index <= 60 ) || ( 66 <= index && index <= 71 ) || ( 77 <= index && index <= 83 ) || ( 88 <= index && index <= 94 ) );
+}
+
+bool Battle::Board::isIndexToTheRightOfTheMoat( const int32_t index )
+{
+    return ( ( 8 <= index && index <= 10 ) || ( 19 <= index && index <= 21 ) || ( 29 <= index && index <= 32 ) || ( 40 <= index && index <= 43 )
+             || ( 50 <= index && index <= 54 ) || ( 62 <= index && index <= 65 ) || ( 73 <= index && index <= 76 ) || ( 85 <= index && index <= 87 )
+             || ( 96 <= index && index <= 98 ) );
+}
```
ihhub, line 722: Can we please avoid hardcoded values if possible? You can take a look at battle_board.h file, there's an enumeration which might help to avoid some of hardcoded values. For example:
`( 8 <= index && index <= 10 )`
will be
`( CASTLE_FIRST_TOP_WALL_POS <= index && index <= ARENAW )`

ihhub (commented):
Hi @oleg-derevenetz , I left just one suggestion here. Would you mind to check it?

ihhub:

src/fheroes2/battle/battle_board.cpp
oleg-derevenetz: Hi @ihhub It seems that this enumeration shouldn't help much in this case :) But if you insist I can replace a few of these values.

ihhub: If we can replace without a lot of work then let's please do it.

Branikolog:
It is really difficult to compare the original behaviour and our engine, as the original game has really much movement bugs during a siege...
For example: in the OG Cavalry can attack boars standing far over a moat...
image
So, I'm not sure, that in fheroes2 Champions should be able to attack Phoenixes through moat...
image
Also, I'm not sure that Cyclops should be able to attack Medusas from this angle: 
image
Medusas cannot attack Champions through moat... If moat movement penalty is equal for defenders and attackers, we should allow both to attack. Or prevent the attack for both Champions and Medusas through moat...

There's also showed wrong angle attack through moat:
image
image
Medusas couldn't reach shown place but nevertheless proceeded attack:
image

I'm not sure the issue is related to current topic...
Medusas couldn't step back here:
image
Just skipped their turn.
As well as Hydras trying to reach the cell above them:
image
But can proceed movement if I point the cell inside moat:
image
image

oleg-derevenetz:
@Branikolog 

> Also, I'm not sure that Cyclops should be able to attack Medusas from this angle:

Medusas are in a moat here (partially), so why not?

> Medusas cannot attack Cyclops through moat...

And there neither of them is in the moat (even partially), so cross-moat attacks are prohibited.

Branikolog:
> @Branikolog
> 
> > Also, I'm not sure that Cyclops should be able to attack Medusas from this angle:
> 
> Medusas are in a moat here (partially), so why not?

I not sure, that cyclops are able to reach that cell...  But they definitely should be able to attack medusas. 

> > Medusas cannot attack Cyclops through moat...
> 
> And there neither of them is in the moat (even partially), so cross-moat attacks are prohibited.

So, the issue above with champions attacking phoenixes through moat is a bug, right?

oleg-derevenetz:
@Branikolog

> I not sure, that cyclops are able to reach that cell...

He is able - through a diagonal move from a cell under the break in the wall right before the Medusas.

> So, the issue above with champions attacking phoenixes through moat is a bug, right?

Yes, it seems so - they are not counted as standing "on either side of the moat", because it's the moat cell under their tail. I'll try to fix this.

oleg-derevenetz:
@Branikolog

> Medusas couldn't reach shown place but nevertheless proceeded attack:

They are able to reach that cell - it would be their tail cell. It's just a wrong highlighting.

oleg-derevenetz:
@Branikolog

> Just skipped their turn.
> As well as Hydras trying to reach the cell above them:

Interesting, I can't reproduce this:

Before
After

Hydras are able to move both when I point to the moat cell and to the ground cell. Turn was never skipped.

oleg-derevenetz:
> Medusas couldn't step back here:

Yes, issue is confirmed and I (hopefully) have a fix.

oleg-derevenetz:
@Branikolog

> > So, the issue above with champions attacking phoenixes through moat is a bug, right?
> 
> Yes, it seems so - they are not counted as standing "on either side of the moat", because it's the moat cell under their head. I'll try to fix this.

Hmm, I made a few more tests and it seems that in OG bridge is considered as a part of the moat, so unit on the bridge is able to attack units on both sides of the moat:




So it seems that my initial behavior was correct.

oleg-derevenetz:
A short feedback summary so far.

> So, I'm not sure, that in fheroes2 Champions should be able to attack Phoenixes through moat...
> image

Yes, in OG bridge [is considered as a part of the moat](https://github.com/ihhub/fheroes2/pull/3972#issuecomment-898641875), so unit on the bridge is able to attack units on both sides of the moat.

> Also, I'm not sure that Cyclops should be able to attack Medusas from this angle:
> image

They are able - through a diagonal move from a cell under the break in the wall right before the Medusas.

> Medusas cannot attack Cyclops through moat...
> image

They shouldn't - neither of them is in the moat (even partially), so cross-moat attacks are prohibited.

> There's also showed wrong angle attack through moat:
> image
> image
> Medusas couldn't reach shown place but nevertheless proceeded attack:
> image

Looks like a cell highlighting issue, the attack itself is performed from the correct position.

> I'm not sure the issue is related to current topic...
> Medusas couldn't step back here:
> image
> Just skipped their turn.

Fixed by cf55fc6.

> As well as Hydras trying to reach the cell above them:
> image
> But can proceed movement if I point the cell inside moat:
> image
> image

I [was not able](https://github.com/ihhub/fheroes2/pull/3972#issuecomment-898558872) to reproduce this.

Branikolog:
@oleg-derevenetz 
> 
> Hmm, I made a few more tests and it seems that in OG bridge is considered as a part of the moat, so unit on the bridge is able to attack units on both sides of the moat:
> 
> So it seems that my initial behavior was correct.

Hmm, I'm not sure it looks good, when a troop can attack someone while standing on a bridge, and cannot, if he's on a ground:
image
image

I believe, this is some kind of bug in the OG.

Also, I'm not sure, that we should allow here Cyclops to attack Griffins:
image
As they should walk one cell in a moat, preventing them from attacking troops on the other side of moat. (Note, that cyclops cannot enter castle in a case those cells'd be free.)
image

Branikolog:
Regarding Hydras... I'll retest it tomorrow, as I was able to reproduce the issue several times the time I created the issue at my work machine...

oleg-derevenetz:
Hi @Branikolog

> I believe, this is some kind of bug in the OG.

So what's the actual logic should be?

> Also, I'm not sure, that we should allow here Cyclops to attack Griffins:

The rules of all this is highly non-obvious. For example, in OG Crusader is not able to attack from the moat here:



but he is able to attack from the moat here:



Why? I don't know.

> Regarding Hydras... I'll retest it tomorrow, as I was able to reproduce the issue several times the time I created the issue at my work machine...

I tried to reproduce it many times, I even lined up other units in the same way (not quite the same units, but units of the same size) - no luck. BTW, you may get a new build for this PR with fix for Medusas that won't step backwards:

https://ci.appveyor.com/project/ihhub/fheroes2/builds/40374877

Branikolog:
@oleg-derevenetz If Crusaders would be able to attack Gargoyles, we can easily abuse this by placing all our troops in a moat and attack one by one the troop behind the moat, while only the opposite creature should be able to attack. Imagine, that I attack with a weak creature (Standing somewhere in a moat nearby) before Crusader's turn. After the creature dies I can proceed an attack with Crusaders without retaliation..
Regarding the case, where Crusaders and Gargoyles are on the same side of a moat: I'm really not sure, whether we should allow them to attack from that angle... We can create a special rule for 100% cases: Creature cannot attack while stepping into a moat or making one step in a water... Even backwards... 
image

This are only my thoughts... I think, this behaviour is quite predictable and can be easily accepted by the majority of players, while a creature, standing somewhere in a moat, coming close, stepping into a moat and attacking in 1 turn seems to be not fair.)

oleg-derevenetz:
Hi @Branikolog I simplified rules a bit: attack from the moat cell is allowed only if attacker is already near the target (and also (hopefully) fixed battle shadow rendering in cases like with Medusas above), you can test this:

https://ci.appveyor.com/project/ihhub/fheroes2/builds/40377119

But there is currently an issue with this approach: for example, wide units from castle are unable to attack non-nearby units from their tail cells if these cells will be in the moat. They are able to attack these units from their head cells (which are on the ground), but not from tail cells, which looks odd. Need to think about how best to handle this...

Branikolog:
Hi, @oleg-derevenetz 
It's a very strange, but I can't reproduce the issue with Hydras using yesterday's version... Don't know, what is happening with this reptile...
I'd better test your latest version then...

Branikolog:
Using the latest version also noticed that 2-hex creatures are able to attack, while their back moves into a moat.
image
Note, that this only happens in a case, when champions go around phoenixes...
image

There're also some problems with highlighting cells and the position, taken by Champions during an attack:
image
image
But it seems to be another issue with choosing the position for attack of 2-hex creatures...

oleg-derevenetz:
@Branikolog

> Using the latest version also noticed that 2-hex creatures are able to attack, while their back moves into a moat.

That's what I said [here](https://github.com/ihhub/fheroes2/pull/3972#issuecomment-898772210) :) Currently I don't know how to properly handle this.

> There're also some problems with highlighting cells and the position, taken by Champions during an attack:

What's the problem exactly? Cell highlighting and attack position seems to be correct - unit will always attack from its head cell if possible (and it is obviously possible here). Champions will attack from their head cell while their tail cell is in the moat (and they currently are not able to attack from their tail cell, because it's in the moat - due to an issue described above).

Branikolog:
@oleg-derevenetz 
image
Looking at the screenshot I expect Champions take following position:
image
Without making a step into a moat...

Currently we're missing one position, while attacking, where Champions make far diagonal attack from the position:
image
But we can attack only from bottom position.
image

oleg-derevenetz:
@Branikolog I suppose we can leave the current behavior as is - wide unit is able to attack from the cell that will be on the ground (after the movement) but not from the cell that will be in the moat. It seems that there is no ideal solution - each has it's own downsides. If we disable attack completely if any part of the wide unit will be in a moat after movement, it will look rather odd too.

oleg-derevenetz:
@Branikolog

> Looking at the screenshot I expect Champions take following position:

No, it doesn't work this way. Units will always attack from their "main" cell (and it's a head cell) if target cell CAN become their head cell after movement (and in this case it obviously can).

> Currently we're missing one position, while attacking, where Champions make far diagonal attack from the position:

That's because instead of "attack the cell from a direction" mechanics another mechanics was initially chosen - "one cell attacks another cell". All mechanics is different and I suppose nobody will rewrite this all from scratch in the foreseeable future :)

Branikolog:
@oleg-derevenetz 
I understand, that it's not a trivial task to rewrite the code (Or to add one more attack direction to the existing), But in some cases I would like my 2-hex creatures to position themselves around target at 100% cells:
If champions can attack from this angle:
image
Why can't they do from phoenixes' back?
image
Even in this particular case I prefer taking the upper left position, to cover Medusas and don't let them go down:
image
At least now we have highlighted cells, and the taken position won't be a surprise for a player... :)

oleg-derevenetz:
@Branikolog

> I understand, that it's not a trivial task to rewrite the code

Rewriting the code itself is easy: slam bang - and it's dоnе :) However, rewrite it CORRECTLY is a huge job (it took more than a year to catch MOST bugs in a current approach), noone wants to go back to the 0.8.x state regarding bugs (especially users), so we have to make compromises.

What do you think regarding my proposal to allow wide units to attack from their ground cells even if they are partially in a moat? They will not able to attack cross-moat, and this behavior will at least have some straight logic.

Branikolog:
Your solution looks like a compromise, but 2-hex creatures would gain an advantage being able to proceed an attack from a moat cell and being able to enter a castle the next turn, while 1-hex creatures should firstly step into a moat, and only second turn they would be able either to attack or enter a castle through a moat.

Ideally, Champions should be able to attack Phoenixes from the current position. 
Champion Diagonal attack

The same for a bottom line of cells:
image

When choosing moat cells - the attack couldn't be proceeded.

oleg-derevenetz:
@Branikolog

> Your solution looks like a compromise, but 2-hex creatures would gain an advantage being able to proceed an attack from a moat cell and being able to enter a castle the next turn, while 1-hex creatures should firstly step into a moat, and only second turn they would be able either to attack or enter a castle through a moat.

Huh, in OG wide units are even able to perform cross-moat attacks:

cross_moat

:) I suppose just because their "main" cell (tail cell in OG) will be on the ground during the attack. At least they will not be able to do this in fheroes2.

> Ideally, Champions should be able to attack Phoenixes from the current position.

They are able to do this in OG because "main" cell is tail cell here. So with main tail there is no problem to take this position. With main head Champions inevitably will step into the moat and will not able to attack Phoenixes, which is illogical. That's why my compromise.

Branikolog:
@oleg-derevenetz 
Let's not look at the OG anymore regarding the problem.) Ton of bugs is happening during a battle there. :)

Sorry for annoying you with my complains. ;)
I just hope, we could find the best solution for this problem.

Look at the screenshot.
image
Cavalry is going to attack Hydras, but with current implementation, it is not possible without moving. 
During an attack Cavalry steps back into a moat and Crusaders are able to hit them... I think we should find a method to allow such diagonal attacks from that angle (The same I was showing before at screenshots with Phoenixes and Champions).

oleg-derevenetz:
@Branikolog unfortunately, engine doesn't have a psychic powers to decide what the player is GOING to do here - is he going to attack from the tail cell or he is going to slightly move and attack from the head cell (to be able to attack someone on the other side of the moat on the next turn for example). Both variants are legitimate here. But at least castle defenders are always able to attack units in the moat without moving to a moat, because their head cells a turned to the moat side while they are inside a castle.

Branikolog:
@oleg-derevenetz 
I believe, the key to this problem lays in direct attacks, we're currently missing. 
image
image

In the OG we can properly attack from the front diagonal and from the back too.

ihhub:
Hi @oleg-derevenetz  and @Branikolog , first of all, thank you very much for your efforts and fruitful conversation!

If the only problem left wide creature top and bottom attacks then we can proceed with these changes as the direct attacks is another issue not covered in this pull request.

ihhub (approved):

oleg-derevenetz:
Hi @ihhub

> If the only problem left wide creature top and bottom attacks then we can proceed with these changes as the direct attacks is another issue not covered in this pull request.

Yes, I believe so.

ihhub:
@oleg-derevenetz , big thanks for fixing this issue!

PR #3975, opened 2021-08-05
Audio thread-safety fixes, various bugfixes

### LINKED ISSUE #3974: Adventure map, music, playing while AI turn, plays from the beginning each time new AI starts his turn
### Preliminary checks

### Platform

Windows

### Describe the bug

While playing with allies, each time new ally starts his turn, "end turn" theme plays from the beginning.

### Save file

It is obvious, while playing "Fortress Isle" map.
[Music loop.zip](https://github.com/ihhub/fheroes2/files/6938803/Music.loop.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #3974

* Reset environment sounds and terrain music theme at the end of the human turn instead of the beginning of the AI turn;
* Make currentMusic atomic, since it could be read and modified from `AGG::AsyncSoundManager` worker thread as well as from main thread in other parts of the code;
* Combine Mixer, Music and Audio, use the mutex to avoid potential race conditions because functions from these namespaces might be called from `AGG::AsyncSoundManager` worker thread as well as from main thread in other parts of the code, modernize code a bit;
* Various minor fixes.

---

~~Also, I really don't like that functions from `Mixer` and `Music` namespaces may be called from `AGG::AsyncSoundManager` worker thread and other parts of the code simultaneously while having no synchronization, it smells like potential race conditions - I doubt that even `SDL_mixer` itself is thread-safe, not to mention auxiliary variables in those modules. I added mutexes to protect those functions so far~~.

~~P.S. In a good way, `Mixer` and `Music` should be combined - they both use the same `SDL_mixer` and call each other's functions, and since they use separate mutexes, this potentially may cause a deadlock - if one thread calls function from `Music` that calls function from `Mixer` and the other thread calls function from `Mixer` that calls function from `Music`.~~


ihhub:
Hi @oleg-derevenetz , could you please resolve merge conflict for this pull request?

oleg-derevenetz:
> Hi @oleg-derevenetz , could you please resolve merge conflict for this pull request?

Yes of course, WIP :)

ihhub:
> 
> 
> > Hi @oleg-derevenetz , could you please resolve merge conflict for this pull request?
> 
> Hi @ihhub Yes of course, WIP :)

Ah, okay. Let's not take a risk and target it for 0.9.7. I'm going to make 0.9.6 release pretty soon so no rush on fixing this pull request then.

oleg-derevenetz:
@ihhub

> Ah, okay. Let's not take a risk and target it for 0.9.7. I'm going to make 0.9.6 release pretty soon so no rush on fixing this pull request then.

I agree. The observed bug with AI sound is minor, but potential race conditions doesn't looks good. I had a few crashes (very rare) on my macOS inside SDL_mixer, and I blamed the library, but now I think that maybe it was caused by race conditions...

oleg-derevenetz:
I think it's at least better than it was before. A few notes:

* `AGG::PlayMusicInternally()` eventually (via `Settings::GetLastFile()`) calls `getenv()` which is guaranteed to be thread-safe by C++11 unless there is no calls to `setenv()`/`putenv()` from competing threads;
* It also eventually calls `readdir()` which is not guaranteed to be thread-safe by POSIX (and in older `libc`s it wasn't), but now in fact it is because `readdir_r()` is shit and deprecated everywhere;
* It also still accesses `Settings` data in one place via `GetRootDirs()`, but only to get `path_program`, which is set just once on app startup and then doesn't change.

In general, all this is still quite fragile in terms of thread safety, but it looks now MUCH better than before. Maybe in the future we somehow rewrite all this to call only functions directly related to sound from threads. The simpler the thread worker code, the better.

src/fheroes2/agg/agg.cpp
```diff
@@ -186,9 +180,9 @@ namespace AGG
         std::condition_variable _workerNotification;
         std::condition_variable _masterNotification;
 
-        std::queue<std::pair<int, bool> > _musicTasks;
-        std::queue<int> _soundTasks;
-        std::queue<std::vector<int> > _loopSoundTasks;
+        std::queue<std::tuple<int, MusicSource, bool>> _musicTasks;
+        std::queue<std::tuple<int, int>> _soundTasks;
+        std::queue<std::tuple<std::vector<int>, int>> _loopSoundTasks;
```
ihhub, line 185: Can we please try to avoid tuple usage? It's useful for code but hard to read the code after. For tuples with 2 parameters we can use std::pair.

ihhub (commented):
Hi @oleg-derevenetz , all changes are good. I left just one comment in respect to std::tuple. Could you please take a look?

ihhub:
@oleg-derevenetz , a bunch of thanks for these useful changes!

PR #3976, opened 2021-08-05
Remove extra check: res should have no other value than END_TURN here

oleg-derevenetz:


src/fheroes2/game/game_startgame.cpp
```diff
@@ -646,23 +646,21 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                 // CONTROL_AI turn
                 default:
-                    if ( res == fheroes2::GameMode::END_TURN ) {
```
ihhub, line 649: If we are sure about this shall we add an assertion to verify this?

ihhub:

PR #3981, opened 2021-08-06
Clang-Tidy: disable cppcoreguidelines-avoid-non-const-global-variables

oleg-derevenetz:
Reason: namespace-scope and module-scope variables are routinely used in this project.


ihhub:
@oleg-derevenetz , thanks for this change!

PR #3983, opened 2021-08-06
Move duplicated code to filterByName function

qarkai:
Also reduce number of comparisons using `StrCmp` type.


ihhub:
@qarkai , thank you very much for this code duplication reduction!

PR #3984, opened 2021-08-06
Post-release removing of unused headers and forward declarations

qarkai:


ihhub:
@qarkai , huge thanks for this code cleanup!

PR #3990, opened 2021-08-08
Replace TinyConfig::AddEntry with simple insert

qarkai:
`uniq` flag never was `true`.

### DIFF
--- a/src/engine/tinyconfig.cpp
+++ b/src/engine/tinyconfig.cpp
@@ -71,7 +71,7 @@ bool TinyConfig::Load( const std::string & cfile )
             left = StringTrim( left );
             right = StringTrim( right );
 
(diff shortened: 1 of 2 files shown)


src/engine/tinyconfig.cpp
```diff
@@ -71,23 +71,13 @@ bool TinyConfig::Load( const std::string & cfile )
             left = StringTrim( left );
             right = StringTrim( right );
 
-            AddEntry( left, right, false );
+            insert( { ModifyKey( left ), right } );
```
ihhub, line 74: Could we please use `emplace` instead?

ihhub (changes requested):

ihhub:
@qarkai , thanks a lot for this code cleanup!

PR #3992, opened 2021-08-08
Game settings dialog: reset current language to English if game resources doesn't support other languages

### LINKED ISSUE #3991: Impossible to set English after switching from localized version to English version
### Preliminary checks

### Platform

Windows

### Describe the bug

Run localized version of the game resources. Choose the language which it supports except English. Run fheroes2 only on English based resources. Go to Main Menu --> Configuration --> Click change language and you're unable to do so. Also it shows that the version is English after all but in reality it is not.

### Save file

N.A.

### Additional info

_No response_

oleg-derevenetz:
Fix #3991

This approach still allows to set unsupported languages (like Spanish or Lithuanian) via `fheroes2.cfg`. However, user shouldn't try to choose the language through the GUI in this case.

### DIFF
--- a/src/fheroes2/dialog/dialog_game_settings.cpp
+++ b/src/fheroes2/dialog/dialog_game_settings.cpp
@@ -207,22 +207,26 @@ namespace fheroes2
             case SelectedWindow::Language: {
                 Settings & conf = Settings::Get();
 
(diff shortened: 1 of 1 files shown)


ihhub:

PR #3993, opened 2021-08-09
A not visited obelisk remains not visited after the UA has been found.

### LINKED ISSUE #438: Obelisk status after digging artifact up
After digging ultimate artifact up all obelisk show "Already visited" status. Even those, you haven't visited before.
Actually, i'm not sure how it works in original game...
Obelisk

Arthusppp:
A "not visited" obelisk remains "not visited" after the Ultimate Artifact has been found. Fixes #438.

In this PR, I am fixing a bug when all obelisks become visited when the Ultimate Artifact is found (even if a player never visited any obelisk). For future enhancements, pls visit #438 or create a new issue.


ihhub:
@Arthusppp , huge thanks for fixing this issue!

PR #3994, opened 2021-08-09
Fix installation guide for pacman-based Linux

qarkai:
Add command for SDL2 build.


ihhub:
@qarkai , thanks a lot for this change!

PR #3995, opened 2021-08-09
Deduplicate settings code

qarkai:
Use setters in `Settings::Read` in instead of copypaste.
Change scroll speed enum values to simplify read/write.
Use `clamp` and `min` instead of manual comparisons.
Also fix `battle_speed` initial value.


ihhub:
@qarkai , a bunch of thanks for this code cleanup!

PR #4001, opened 2021-08-10
Do not let the player get a random artifact that allows him to win the game

### LINKED ISSUE #3999: Random artifact is same as quest goal.
### Preliminary checks

### Platform

Linux

### Describe the bug

Random artifact from grave is same as artifact in victory condition. In that case we get fast victory.

### Save file

[bug2.zip](https://github.com/ihhub/fheroes2/files/6964177/bug2.zip)

Search the grave for fast victory.

### Additional info

_No response_

oleg-derevenetz:
fix #3999

### DIFF
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -839,14 +839,14 @@ u32 GoldInsteadArtifact( int obj )
     return 0;
 }
 
-void ResetArtifactStats()
+void fheroes2::ResetArtifactStats()
 {
     for ( artifactstats_t & item : artifacts ) {
         item.bits = 0;
     }
 }
 
-void ExcludeArtifactFromRandom( int art )
+void fheroes2::ExcludeArtifactFromRandom( int art )
 {
     const size_t id = static_cast<size_t>( art );
 
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -56,11 +56,11 @@ enum
 
 struct artifactstats_t
 {
-    u8 bits;
-    u8 extra;
-    u8 type;
-    const char * name;
-    const char * description;
+    uint8_t bits;
+    const uint8_t extra;
+    const uint8_t type;
+    const char * const name;
+    const char * const description;
 };
 
 artifactstats_t artifacts[] = {
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -846,9 +846,9 @@ void fheroes2::ResetArtifactStats()
     }
 }
 
-void fheroes2::ExcludeArtifactFromRandom( int art )
+void fheroes2::ExcludeArtifactFromRandom( const int artifactID )
 {
-    const size_t id = static_cast<size_t>( art );
+    const size_t id = static_cast<size_t>( artifactID );
 
     assert( id < artifacts.size() );
 
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -21,6 +21,7 @@
  ***************************************************************************/
 
 #include <algorithm>
+#include <array>
 #include <string>
 #include <vector>
 
@@ -42,7 +43,7 @@
 
 enum
 {
-    ART_DISABLED = 0x01,
+    ART_RNDDISABLED = 0x01,
     ART_RNDUSED = 0x02
 };
 enum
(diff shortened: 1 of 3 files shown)


src/fheroes2/resource/artifact.h
```diff
@@ -205,8 +205,12 @@ class Artifact
 
 StreamBase & operator<<( StreamBase &, const Artifact & );
 StreamBase & operator>>( StreamBase &, Artifact & );
+
 u32 GoldInsteadArtifact( int );
 
+void ResetArtifactStats();
```
ihhub, line 211: Could we please avoid using global functions? We can put them in `fheroes2` namespace to reduce their scope.

src/fheroes2/resource/artifact.h
```diff
@@ -205,8 +205,12 @@ class Artifact
...
+void ExcludeArtifactFromRandom( int art );
```
ihhub, line 212: I would recommend to make this input argument `const` and name it as `artifactID` :)

src/fheroes2/resource/artifact.cpp
```diff
@@ -56,156 +58,162 @@ enum
...
-    case Artifact::ENDLESS_CORD_WOOD:
-        return _n( "The %{name} provides %{count} unit of wood per day.", "The %{name} provides %{count} units of wood per day.", count );
-    case Artifact::ENDLESS_CART_ORE:
-        return _n( "The %{name} provides %{count} unit of ore per day.", "The %{name} provides %{count} units of ore per day.", count );
-    case Artifact::ENDLESS_POUCH_CRYSTAL:
-        return _n( "The %{name} provides %{count} unit of crystal per day.", "The %{name} provides %{count} units of crystal per day.", count );
-    default:
-        break;
+    std::array<artifactstats_t, 104> artifacts = { {
+        { 0, 12, TYPE3, _( "Ultimate Book of Knowledge" ), _( "The %{name} increases your knowledge by %{count}." ) },
```
ihhub, line 71: Could we please remove translation calls here since we do the same in the code below and also it's a global variable which is initialized at the very beginning of the application.

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -80,6 +80,18 @@ bool World::LoadMapMP2( const std::string & filename )
     Reset();
     Defaults();
 
+    // clear artifact flags to correctly generate random artifacts
+    ResetArtifactStats();
```
ihhub, line 84: What's about loading a save file? I think we need to reset there as well.

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -80,6 +80,18 @@ bool World::LoadMapMP2( const std::string & filename )
...
+
+    const Settings & conf = Settings::Get();
+
+    // do not let the player get a random artifact that allows him to win the game
+    if ( ( conf.ConditionWins() & GameOver::WINS_ARTIFACT ) == GameOver::WINS_ARTIFACT && !conf.WinsFindUltimateArtifact() ) {
+        const Artifact art = conf.WinsFindArtifactID();
+
+        ExcludeArtifactFromRandom( art.GetID() );
+    }
```
ihhub, line 93: Shouldn't we need to put this check just before `ProcessNewMap` call?

ihhub (changes requested):
Hi @oleg-derevenetz , there are multiple comments. Could you please take a look at them?

src/fheroes2/world/world_loadmap.cpp
oleg-derevenetz: @ihhub As far as I know, random artifacts are replaced by actual ones on the start of a new game (when MP2 is processed), so in save files random artifacts are not random anymore :)

ihhub: That's a good point :)

src/fheroes2/resource/artifact.cpp
oleg-derevenetz: I believe that translation calls are needed here to properly generate `.po` file. They don't really work but they are just markers for `gettext` parser.

ihhub: Oh, yes. I totally forgot about this. We can leave as it is or use `gettext_noop` call instead.

oleg-derevenetz: I believe `_()` looks shorter.

ihhub: The difference between `_()` and `gettext_noop` is that the last function does exactly nothing. Anyway, I'm marking this comment as resolved.

ihhub:
@oleg-derevenetz , good job! Thanks for the fix!

PR #4008, opened 2021-08-12
Finish idle animation during battle action

### LINKED ISSUE #1229: Idle animation stops while any action in a battle is performing
Currently, when any creature makes any movement, all idle animations stop at one certain frame. That could look weird. For example: if any creature closes his eyes and another one makes shoot - the first creature keeps standing with its eyes closed till the animation of shooting is finished. Also, idle animation freezes while AI makes any actions during its turn.
In the OG, idle animation interrupts and sprite returns to the default (standing) state.
That would be nice, if we can make creatures continue idle animations, or finish at least the last one.

vasilenkoalexey:
Finish idle animation during battle action.  fix #1229

### DIFF
--- a/src/fheroes2/battle/battle_army.h
+++ b/src/fheroes2/battle/battle_army.h
@@ -68,7 +68,7 @@ namespace Battle
         int GetControl( void ) const;
         uint32_t GetSurrenderCost( void ) const;
         Troops GetKilledTroops( void ) const;
-        bool animateIdleUnits( const bool restart = true );
+        bool animateIdleUnits( const bool forceRestartIdleAnimation );
         void resetIdleAnimation();
 
         void NewTurn( void );
--- a/src/fheroes2/battle/battle_army.h
+++ b/src/fheroes2/battle/battle_army.h
@@ -70,6 +70,7 @@ namespace Battle
         Troops GetKilledTroops( void ) const;
         bool animateIdleUnits( const bool forceRestartIdleAnimation );
         void resetIdleAnimation();
+        void resetAnimation();
 
         void NewTurn( void );
         void SyncArmyCount();
--- a/src/fheroes2/battle/battle_army.h
+++ b/src/fheroes2/battle/battle_army.h
@@ -68,7 +68,7 @@ namespace Battle
         int GetControl( void ) const;
         uint32_t GetSurrenderCost( void ) const;
         Troops GetKilledTroops( void ) const;
-        bool animateIdleUnits();
+        bool animateIdleUnits( const bool restart = true );
         void resetIdleAnimation();
 
         void NewTurn( void );
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -2710,15 +2710,13 @@ void Battle::Interface::AnimateUnitWithDelay( Unit & unit, uint32_t delay )
     while ( le.HandleEvents( false ) ) {
         CheckGlobalEvents( le );
 
-        const bool redrawNeeded = IdleTroopsAnimation( false );
-
         if ( Game::validateCustomAnimationDelay( frameDelay ) ) {
             Redraw();
             if ( unit.isFinishAnimFrame() )
                 break;
             unit.IncreaseAnimFrame();
         }
-        else if ( redrawNeeded ) {
+        else if ( IdleTroopsAnimation( false ) ) {
             Redraw();
         }
     }
@@ -2748,15 +2746,13 @@ void Battle::Interface::RedrawTroopDefaultDelay( Unit & unit )
     while ( le.HandleEvents( false ) ) {
         CheckGlobalEvents( le );
 
-        const bool redrawNeeded = IdleTroopsAnimation( false );
-
         if ( Game::validateAnimationDelay( Game::BATTLE_FRAME_DELAY ) ) {
             Redraw();
             if ( unit.isFinishAnimFrame() )
                 break;
             unit.IncreaseAnimFrame();
         }
-        else if ( redrawNeeded ) {
+        else if ( IdleTroopsAnimation( false ) ) {
             Redraw();
         }
     }
@@ -2798,8 +2794,6 @@ void Battle::Interface::RedrawMissileAnimation( const fheroes2::Point & startPos
     while ( le.HandleEvents( false ) && pnt != points.end() ) {
         CheckGlobalEvents( le );
 
-        const bool redrawNeeded = IdleTroopsAnimation( false );
-
         if ( Game::validateAnimationDelay( Game::BATTLE_MISSILE_DELAY ) ) {
             RedrawPartialStart();
             if ( isMage ) {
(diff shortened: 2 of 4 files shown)


src/fheroes2/battle/battle_army.h
```diff
@@ -68,7 +68,7 @@ namespace Battle
         int GetControl( void ) const;
         uint32_t GetSurrenderCost( void ) const;
         Troops GetKilledTroops( void ) const;
-        bool animateIdleUnits();
+        bool animateIdleUnits( const bool restart = true );
```
ihhub, line 71: Please do not use default values for input arguments. Could we please remove it?

ihhub: Also could we please name this input argument something like `forceRestartIdleAnimation` to be more specific about the purpose of this variable?

src/fheroes2/battle/battle_interface.h
```diff
@@ -274,7 +274,7 @@ namespace Battle
         void RedrawTargetsWithFrameAnimation( int32_t dst, const TargetsInfo & targets, int icn, int m82, int repeatCount = 0 );
         void RedrawTargetsWithFrameAnimation( const TargetsInfo &, int, int, bool );
 
-        bool IdleTroopsAnimation( void );
+        bool IdleTroopsAnimation( const bool restart = true );
```
ihhub, line 277: The same comment here: could we please avoid default parameters for input arguments?

ihhub: And also could we please make the variable more specifically as right now it's unknown what we want to restart.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4805,6 +4962,9 @@ void Battle::Interface::RedrawTroopWithFrameAnimation( Unit & b, int icn, int m8
             }
             ++frame;
         }
+        else if ( redrawNeeded ) {
```
ihhub, line 4965: We don't need to create another variable which is used only in one place. We can directly call the method here. Also it'll be faster since we won't call the method if the above condition is true. This applies for all cases in these changes.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4169,6 +4251,9 @@ void Battle::Interface::RedrawActionStoneSpell( const Unit & target )
             alpha += 10;
             ++frame;
         }
+        else if ( redrawNeeded ) {
+            Redraw();
```
ihhub, line 4255: Could you please check how it will look during a spell casting? I think the monster might be moving being stoned at the same time which is not allowed to be.

ihhub (changes requested):
Hi @vasilenkoalexey , I left multiple comments here. Would you mind to take a look at them? Also please verify my idea that monsters on which a spell is applied are still moving.

src/fheroes2/battle/battle_interface.cpp
vasilenkoalexey: Let me explain...
I tried to cover case when we should draw both idle and battle animation, 
```
 while ( le.HandleEvents() && currentAlpha < 220 ) {
        CheckGlobalEvents( le );

        const bool redrawNeeded = IdleTroopsAnimation( false );

        if ( Game::validateAnimationDelay( Game::BATTLE_SPELL_DELAY ) ) {
            currentAlpha += 20;
            target.SetCustomAlpha( currentAlpha );
            Redraw();
        }
        else if ( redrawNeeded ) {
            Redraw();
        }
    }
```
In this case we call Redraw() once

But for this implementation
```
 while ( le.HandleEvents() && currentAlpha < 220 ) {
        CheckGlobalEvents( le );

        if ( Game::validateAnimationDelay( Game::BATTLE_SPELL_DELAY ) ) {
            currentAlpha += 20;
            target.SetCustomAlpha( currentAlpha );
            Redraw();
        }
        if ( IdleTroopsAnimation( false ) ) {
            Redraw();
        }
    }
```
we call Redraw()  twice, 
So, i dont see another simple implementation without additional var.

But if you think this has little performance impact, I will fix

src/fheroes2/battle/battle_interface.cpp
vasilenkoalexey: Here we animate idle
```
bool Battle::Force::animateIdleUnits( const bool forceRestartIdleAnimation )
{
    bool redrawNeeded = false;

    for ( Force::iterator it = begin(); it != end(); ++it ) {
        Unit & unit = **it;

        // check if alive and not paralyzed
        if ( unit.isValid() && !unit.Modes( SP_BLIND | IS_PARALYZE_MAGIC ) ) {
            if ( unit.isIdling() ) {
                if ( unit.isFinishAnimFrame() ) {
                    redrawNeeded = unit.SwitchAnimation( Monster_Info::STATIC ) || redrawNeeded;
                }
                else {
                    unit.IncreaseAnimFrame();
                    redrawNeeded = true;
                }
            }
```
and
```
        SP_PARALYZE = 0x40000000,
        SP_STONE = 0x80000000,

        IS_PARALYZE_MAGIC = 0xC0000000,
```
And `IS_PARALYZE_MAGIC` looks like `SP_PARALYZE | SP_STONE`. So, all should be fine.
We can replace it for example by IS_FROZEN = SP_BLIND | SP_SP_PARALYZE | SP_STONE or leave it as it is :)

vasilenkoalexey: But spell applied after redraw. I think then we should switch animation to STATIC for all targets before redrawing.

```
    if ( interface ) {
        interface->RedrawActionSpellCastPart1( spell, dst, current_commander, targets );
        for ( const TargetInfo & target : resistTargets ) {
            interface->RedrawActionResistSpell( *target.defender, playResistSound );
        }
    }

    TargetsApplySpell( current_commander, spell, targets );
```

src/fheroes2/battle/battle_interface.cpp
ihhub: I suggested to make
```cpp
else if ( IdleTroopsAnimation( false ) ) {
    Redraw();
}
```

ihhub: Besides this please verify that idle animation won't change its speed while we're drawing a spell with **different** speed.

vasilenkoalexey: > I suggested to make

Ok, done.

> Besides this please verify that idle animation won't change its speed while we're drawing a spell with **different** speed.

Should be ok, since we do not affect the change of idle frames.

src/fheroes2/battle/battle_interface.cpp
vasilenkoalexey: > Could you please check how it will look during a spell casting?

I switch all spell targets to STATIC before redrawing.

src/fheroes2/battle/battle_army.h
vasilenkoalexey: Done

Branikolog:
Everything seems to work well.

ihhub:
Hi @vasilenkoalexey , please try to apply a Resurrect spell on a unit which neighbors a unit with idling animation. The spell is flicking. For example in this scenario I was applying the spell on Titans when Bone Dragons were having idling animation:
image

ihhub:
Hi @vasilenkoalexey , shall we rework this pull request as we made significant changes in this part of the code?

PR #4009, opened 2021-08-12
Make Windows script to automatically copy & extract all necessary assets from original HoMM2 distributions

### LINKED ISSUE #4002: Preferable version of the original game
### Preliminary checks

There's an issue with video files in dosbox version of the original game.
Installation guide lacks the information, that win version is required for fheroes2 engine.

We should note, that windows version of the original game is preferable for fheroes2 installation.

### Additional info

_No response_

oleg-derevenetz:
Related to #4002

Currently only GOG version may be autodetected. I need `.reg` file with contents of the HoMM2 section in the `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall` or `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall` registry keys to autodetect other installed distributions.


oleg-derevenetz:
Hi @ihhub what do you think about this approach? We can install this script along with `demo_windows.ps1` and add it to the Start Menu group so user may just run it if he has one of the official HoMM2 distributions installed. BTW, do you have Ubisoft HoMM2 version installed somewhere? If yes, can you please provide the `.reg` file for `Uninstall` registry subkey for this distribution?

ihhub:
> 
> 
> Hi @ihhub what do you think about this approach? We can install this script along with `demo_windows.ps1` and add it to the Start Menu group so user may just run it if he has one of the official HoMM2 distributions installed. BTW, do you have Ubisoft HoMM2 version installed somewhere? If yes, can you please provide the `.reg` file for `Uninstall` registry subkey for the Ubisoft distribution? This script will be able to detect Ubisoft HoMM2 installations as well.

Hi @oleg-derevenetz , I don't have Ubisoft version, only GoG for now.

My thoughts about making this script more user-friendly:
- bundle it with the installer
- at the time of installation check if specific gog file is present but no **anim** directory exists. If this is the case ask user if he wants to extract files as he probably owns DOS version of the game. If yes then we proceed with the script help.
What do you think about this way?

I'm not sure that users will understand the meaning of this script so adding it to Main Menu might not be the best solution.

After all it's a temporary solution as in the future we have to add C/C++ library to extract into memory these files during application runtime.

Also in few days I'll test this script on my old Win 7 system to verify that it works with PowerShell 2.0.

script/homm2/homm2_windows.ps1
```diff
@@ -0,0 +1,143 @@
...
+                $path = $subkey.GetValue("InstallLocation").TrimEnd("\")
+
+                if (Test-HoMM2DirectoryPath -Path $path) {
+                    $homm2Path = $path
+
+                    break
+                }
+            }
+        }
+    }
```
ihhub, line 85: This might be not the best solution as people could have Ubisoft (as mentioned by you) and even Gold version before GoG existence. As I understood we're trying to find the location of the HoMM2 installation. In this case user knows it. Based on my installation suggestion we could ask user to point the original HoMM2 installation path making this portion of the code obsolete.

script/homm2/homm2_windows.ps1
```diff
@@ -0,0 +1,143 @@
...
+        Write-Host -ForegroundColor Yellow "WARNING: Unable to determine HoMM2 directory"
+
+        $homm2Path = Read-Host -Prompt "Please enter the full path to the HoMM2 directory (e.g. C:\GOG Games\HoMM 2 Gold)"
+
+        if (-Not (Test-HoMM2DirectoryPath -Path $homm2Path)) {
+            Write-Host -ForegroundColor Red "FATAL ERROR: Unable to find HoMM2 directory"
+
+            return
+        }
+    }
```
ihhub, line 97: Just a comment: this is good enough. Let's don't make any extra assumptions about user's installation and give only one option to enter the path.

script/homm2/homm2_windows.ps1
```diff
@@ -0,0 +1,143 @@
...
+
+    Write-Host -ForegroundColor Green (-Join("HoMM2 directory: ", (Resolve-Path $homm2Path).Path))
+
+    Write-Host "[3/3] copying assets"
```
ihhub, line 101: To be more user-friendly I'd call it as **copying game's resources**.

ihhub (changes requested):
Hi @oleg-derevenetz , I left few comments. Also as I see this script serves 2 purposes:
- copy files from the original game installation directory into fheroes2 directory
- extract animation files from gog files

I think we might need to separate these 2 operations as user might choose to install fheroes2 into the original HoMMII directory.

script/homm2/homm2_windows.ps1
oleg-derevenetz: Hi @ihhub I can remove the publisher check here and just search for installations that contains `heroes2.exe` and typical assets directories. In this case script will support literally everything. It will make more checks though but I think it doesn't really matter in this case. Users always like when something works automagically and they don't need to make extra things - like torment with viewing properties of shortcuts, copy/paste etc. If auto detection fails, then they will have an opportunity to enter the path manually.

oleg-derevenetz:
@ihhub 

> I think we might need to separate these 2 operations as user might choose to install fheroes2 into the original HoMMII directory.

I can make a separate mini script that will just extract animations from `homm2.gog` in the current directory, that's all. I doubt that many people will install fheroes2 in the directory with the original game though, but why not.

ihhub:
> @ihhub
> 
> > I think we might need to separate these 2 operations as user might choose to install fheroes2 into the original HoMMII directory.
> 
> I can make a separate mini script that will just extract animations from `homm2.gog` in the current directory, that's all. I doubt that many people will install fheroes2 in the directory with the original game though, but why not.

Based on given feedback many people install fheroes2 into the original game directory. Having a separate script should be fine.

ihhub:
@oleg-derevenetz , for full scale solution I was thinking something like [this library](http://www.libarchive.org/) but a single file solution. I found some before but I can't access the list right now.

ihhub:
How about [this one](https://github.com/erincandescent/lib9660)?

oleg-derevenetz:
Hi @ihhub it seems that `homm2.gog` isn't the plain ISO9660 image - it cannot be mounted by Windows (Windows 10 is able to mount ISO and VHD images) and WinRAR is not able to open it either.

undef21:
Hello @ihhub. homm2.gog is raw cd image. Sample code for converting to iso:
```
#include <stdio.h>
int main(void)
{
    FILE *fsrc, *fdst;
    char buf[2352];
    if((fsrc = fopen("homm2.gog", "rb")) != NULL && (fdst = fopen("homm2.iso", "wb")) != NULL)
        while(fread(buf, 2352, 1, fsrc))
            fwrite(&buf[buf[15] == 2 ? 24 : 16], 2048, 1, fdst);
}
```
Some versions of homm2.gog (eg en) are CD MODE1 images, but some (eg de) are MODE2 Form1 images.

oleg-derevenetz:
Hi @ihhub After some thought, I suggest not integrating the `.gog` file support into the core of the game, but just write an extraction script instead, just like in Windows. I suppose shell wouldn't be enough, but there are also widely supported Python, Tcl or Perl.

ihhub:

ihhub (changes requested):

script/windows/fheroes2.iss
```diff
@@ -45,9 +49,19 @@ Name: desktopicon; Description: "Desktop shortcut"
...
 Name: "{group}\Download demo version files"; Filename: "{app}\demo_windows.bat"; WorkingDir: "{app}"
+Name: "{group}\Extract game resources from the original HoMM2 distribution"; Filename: "{app}\homm2_windows.bat"; WorkingDir: "{app}"
 Name: "{group}\Game data files"; Filename: %WINDIR%\explorer.exe; Parameters: """%APPDATA%\{#AppName}"""
 Name: "{group}\Uninstall"; Filename: "{uninstallexe}"
 Name: "{commondesktop}\Free Heroes of Might & Magic II"; Filename: "{app}\{#AppName}.exe"; WorkingDir: "{app}"; Tasks: desktopicon
 
 [Run]
-Filename: "{app}\demo_windows.bat"; Description: "Download demo version files"; Flags: postinstall runascurrentuser
+Filename: "{app}\demo_windows.bat"; Description: "Download demo version files"; Flags: postinstall runascurrentuser; Check: IsNotInstalledToTheDirectoryOfTheOriginalGame
+Filename: "{app}\anim_extract_windows.bat"; Description: "Extract animation resources"; Flags: postinstall runascurrentuser; Check: FileExists(ExpandConstant('{app}\homm2.gog'))
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub How about the "Extract video files from the original HoMM2 distribution"?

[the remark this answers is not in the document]
ihhub: Works fine too.

ihhub:
Hi @oleg-derevenetz , I'm still thinking about a solution not to provide these script during installation but use them for unpack video files. It means ship them as a part of the installation and use them to unpack files but don't copy them into destination folder. I'm thinking it'll be more user-friendly approach and less questions will be raised. What do you think?

oleg-derevenetz:
Hi @ihhub suppose that user installed fheroes2 first with demo assets, he liked the game and then he decided to buy the full version and use assets from it. In this case it would be useful to have the extraction script already installed (and reachable via Start Menu).

script/homm2/homm2_windows.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
```
ihhub, line 1: We should rename this script. As a user I'm reading this as a script to run homm2, not to copy files.

ihhub:
Hi @oleg-derevenetz , one small correction to do if possible: in case of existing DATA directory with at least heroes2.agg file we should uncheck checkbox for downloading Demo version to avoid unnecessary file overriding.
image

ihhub:
Also interesting fact: we're creating fheroes2 directory within destination directory. I think it's a bug.

ihhub:
> 
> 
> Hi @oleg-derevenetz , one small correction to do if possible: in case of existing DATA directory with at least heroes2.agg file we should uncheck checkbox for downloading Demo version to avoid unnecessary file overriding. image

Ignore this please. I made a mistake during installation.

ihhub:
@oleg-derevenetz , excellent work! Thanks a lot for these changes!

oleg-derevenetz:
@ihhub

> Also interesting fact: we're creating fheroes2 directory within destination directory. I think it's a bug.

AFAIK when user selects the destination directory, the app directory itself is not supposed to exist yet, that's why installer adds "fheroes2" to the selected directory - apparently it is supposed that user should select something like "C:\Program Files", "D:\Program Files", "C:\Games", etc, and only then app directory name should be added to it. The final path may be edited by hand.

PR #4012, opened 2021-08-14
Optimize Battle Arena related code

qarkai:
Remove `invert` flag from `GetCommander()`, add `getEnemyCommander()` as replacement.


ihhub:
@qarkai , thank you very much for this opimization!

PR #4013, opened 2021-08-14
Make sure that a random filename will not clash with the names of existing files

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks a lot for this extra check!

Issue #3621, opened 2021-06-04
Castle screen, Creature separation window, the sentence "Fast separation into slots:" and the graphic "2" are not displayed when there are only 2 creatures in the starting slot during the distribution between the hero's army and the garrison


LeHerosInconnu:
In fheroes2, the sentence "Fast separation into slots:" and the graphic "2" are not displayed when there are only 2 creatures in the starting slot during the distribution between the hero's army and the garrison.

In fheroes2:

![Creature separation window 001](https://user-images.githubusercontent.com/43583994/120869941-e7863d00-c597-11eb-8859-c34379096bc8.png)

![Creature separation window 002](https://user-images.githubusercontent.com/43583994/120869947-ebb25a80-c597-11eb-89b6-18d010b8b649.png)

https://user-images.githubusercontent.com/43583994/120869999-13a1be00-c598-11eb-9aaf-814309d2a729.mp4

PR #4016, opened 2021-08-15
Fix too many free slots in cross-army splitting

### LINKED ISSUE #4015: Not correct splitting of stack into 6 parts
### Preliminary checks

### Platform

Windows

### Describe the bug

Splitting is not working correctly when we split hero's army into 6 stacks, and there's already one slot is occupied in a castle.
image
image

### Save file

[Split issue.zip](https://github.com/ihhub/fheroes2/files/6988000/Split.issue.zip)

### Additional info

_No response_

eos428:
fix #4015

### DIFF
--- a/src/fheroes2/dialog/dialog_selectcount.cpp
+++ b/src/fheroes2/dialog/dialog_selectcount.cpp
@@ -313,9 +313,6 @@ bool Dialog::InputString( const std::string & header, std::string & res, const s
 
 int Dialog::ArmySplitTroop( uint32_t freeSlots, const uint32_t redistributeMax, const bool savelastTroop, uint32_t & redistributeCount, bool & useFastSplit )
 {
-    if ( savelastTroop )
-        ++freeSlots;
-
     assert( freeSlots > 0 );
 
     fheroes2::Display & display = fheroes2::Display::instance();


ihhub:
Hi @Branikolog , could you please test this change for all cases?

Branikolog:
@ihhub 
Everything seems to be ok now.

ihhub:
@eos428 , thanks a lot for the fix!

oleg-derevenetz:
I believe this will break #3621 once again.

ihhub:
Okay, I'll check today every case.

Branikolog:
@oleg-derevenetz 
I've tested as intensely, as I could... I've managed to find another issue, but the problem you've pointed above doesn't occur.

oleg-derevenetz:
Hi @Branikolog I managed to reproduce the #3621 on the latest build:

bug

Try to drag&drop Archers to the garrison.

Branikolog:
Hi, @oleg-derevenetz 
True!
Rare case, when we try to split the only stack in hero army... My fault... Missed this problem...
The maximum result stack number lacks 1 in a case we split the only troop in army.
Splitting orcs, which are NOT the only troop in hero army:
image
Splitting orcs, which are the only troop in hero army:
image

PR #4018, opened 2021-08-15
Battle pathfinding: store passability info separately for the head and tail of wide units, improve AI pathfinding a bit

oleg-derevenetz:
Today I ran into an interesting problem:

bug

Although the Cavalry can occupy each of the highlighted cells separately, it can't occupy them BOTH AT THE SAME TIME. If I press the left mouse button here, Cavalry just skips the turn, and this is completely logical - Cavalry can't reach those cells from inside the castle because of the moat, and it can't reach those cells from outside, because the head cell (the rightmost one) is too far.

This problem apparently cannot be solved within the current pathfinding paradigm, because each of both cells have the valid `direction` field, so I decided to introduce separate info for the head (`_reachableForHead`) and tail (`_reachableForTail`) of units and also replace direction (`int`) by reachability (`bool`) since all that was done with them was a comparison for equality/inequality with `UNKNOWN`, no direction info was ever used anywhere. This approach seems to work just fine:

However, it needs an extensive testing with all unit types (regular units, wide units, flying units, wide flying units) in different circumstances (siege, obstacles, etc).

Also I improved AI pathfinding a bit in relation to the attack - since AI may decide to attack a distant unit, it may select the wrong cell (unreachable on the current turn) for this, and attack fails even if attacker is near the target - because he is literally one step away from the desired cell. In this PR I implemented a logic that allows attacker to attack the target if possible, even if desired cell wasn't reached.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -168,7 +168,7 @@ namespace AI
         assert( targetCell != nullptr );
 
         // Target cell is already reachable
-        if ( targetCell->GetHeadDirection() != UNKNOWN ) {
+        if ( targetCell->isReachableForHead() ) {
             return Board::FixupTargetCellForUnit( currentUnit, target );
         }
 
@@ -177,7 +177,7 @@ namespace AI
 
         // Search for the nearest reachable cell
         for ( const Cell & cell : *Arena::GetBoard() ) {
-            if ( cell.GetHeadDirection() != UNKNOWN ) {
+            if ( cell.isReachableForHead() ) {
                 const uint32_t distance = Board::GetDistance( target, cell.GetIndex() );
 
                 if ( distance < nearestDistance ) {
@@ -202,7 +202,7 @@ namespace AI
             if ( Board::isValidDirection( from, tailDirection ) ) {
                 const Cell * tailCell = Board::GetCell( Board::GetIndexDirection( from, tailDirection ) );
 
-                if ( tailCell != nullptr && tailCell->GetTailDirection() != UNKNOWN && ( tailCell->GetUnit() == nullptr || tailCell->GetUnit() == &attacker ) ) {
+                if ( tailCell != nullptr && tailCell->isReachableForTail() && ( tailCell->GetUnit() == nullptr || tailCell->GetUnit() == &attacker ) ) {
                     headIndex = from;
                     tailIndex = tailCell->GetIndex();
                 }
@@ -215,7 +215,7 @@ namespace AI
                 if ( Board::isValidDirection( from, headDirection ) ) {
                     const Cell * headCell = Board::GetCell( Board::GetIndexDirection( from, headDirection ) );
 
-                    if ( headCell != nullptr && headCell->GetHeadDirection() != UNKNOWN && ( headCell->GetUnit() == nullptr || headCell->GetUnit() == &attacker ) ) {
+                    if ( headCell != nullptr && headCell->isReachableForHead() && ( headCell->GetUnit() == nullptr || headCell->GetUnit() == &attacker ) ) {
                         headIndex = headCell->GetIndex();
                         tailIndex = from;
                     }
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -227,7 +227,7 @@ namespace AI
         }
 
         // Check that the attacker is actually capable of attacking the target from this position
-        std::array<int32_t, 2> indexes = { headIndex, tailIndex };
+        const std::array<int32_t, 2> indexes = { headIndex, tailIndex };
 
         for ( const int32_t idx : indexes ) {
             if ( idx == -1 ) {
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -34,6 +34,7 @@
 #include "settings.h"
 #include "speed.h"
 
+#include <array>
 #include <cassert>
 #include <cmath>
 #include <cstdint>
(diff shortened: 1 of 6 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1702,25 +1706,18 @@ void Battle::Interface::RedrawCover()
...
             if ( isApplicable ) {
-                if ( displayMoveShadow && ( highlightCell->GetDirection() == UNKNOWN || highlightCell->GetUnit() != nullptr ) ) {
-                    fheroes2::Blit( sf_shadow, _mainSurface, highlightCell->GetPos().x, highlightCell->GetPos().y );
-                }
-
                 fheroes2::Blit( sf_cursor, _mainSurface, highlightCell->GetPos().x, highlightCell->GetPos().y );
             }
-            else {
-                fheroes2::Blit( sf_shadow, _mainSurface, highlightCell->GetPos().x, highlightCell->GetPos().y );
-            }
```
oleg-derevenetz, line 1723: Hi @ihhub Could you tell me please why the removed code was needed here - given that `RedrawCoverBoard()` has already been called at the beginning of this function and all movement shadow has already been drawn in it? I removed this code and doesn't see any issues.

ihhub: Hi @oleg-derevenetz , there were some cases which I covered here but I don't remember which ones.

ihhub:
Hi @Branikolog , could you please test these changes including shadow drawings?

Branikolog:
I'm not sure about the relation of some problems to this PR, so, I'll be posting all suspicious cases regarding battle pathfinding I find.
At this particular case Cavalry stands in a moat and can attack phoenixes.
image
But at the same time cavalry cannot attack cyclops from the same place...
image

oleg-derevenetz:
@Branikolog Cavalry can attack Phoenixes here, because it is attacking from the ground cell. It is not able to attack Cyclops because it attacking them from the moat cell.

Branikolog:
> @Branikolog Cavalry can attack Phoenixes here, because it is attacking from the ground cell. It is not able to attack Cyclops because it attacking them from the moat cell.

I clearly understand currently implemented logic.)
But, looking at the situation disregarding that "head-butt division": Cavalry standing on the same two hexes can attack one creature but cannot attack another one. I find it a bit out of logic... Only if we consider a moat as some sort of obstacle, that prevents attack itself... Which is arguable, as water in moat affects only legs of creatures. :) (Well, I, actually, count it more "prevent-move" obstacle, so while standing into a moat with any part of body a creature stops and cannot proceed any action)

oleg-derevenetz:
@Branikolog With wide units there will be always something "wrong" in this regard. If attack will be completely prohibited if either part of the wide unit is in the moat, there will be another "inconsistencies", I believe we have already seen one in the PR regarding cross-moat attacks.

Branikolog:
@oleg-derevenetz 
True. I just hope we could make this logic maximum predictable and comfortable for players.

oleg-derevenetz:
@Branikolog There is also a funny AI behavor here:



If I press "Auto" button here, Paladins just do one step forward, not trying to attack anyone (however at least Unicorns and Druids are reachable for them). At first I thought that it's a pathfinding issue, but this is not really the case. AI just don't want to attack anyone for some reason.

Branikolog:
> @Branikolog There is also a funny AI behavior here:

I think, they are aware of attacking both these stacks, and focused on sprites. Due to the fact, they are unable to reach them, they just made one step, to beat them up next turn. ;)

ihhub:
Hi @Branikolog , as usual you would like to have an ideal solution in one shot. I will rephrase my request: are these changes making the logic better?

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -186,7 +187,64 @@ namespace AI
...
+                    }
+                }
+            }
+        }
+        else {
+            headIndex = from;
+        }
+
+        // Check that the attacker is actually capable of attacking the target from this position
+        std::array<int32_t, 2> indexes = { headIndex, tailIndex };
```
ihhub, line 230: [Optional] we could make this variable `const`.

oleg-derevenetz:
> Hi @Branikolog , as usual you would like to have an ideal solution in one shot. I will rephrase my request: are these changes making the logic better?

Or, what is much more important, do they break anything :)

ihhub:
> 
> 
> > Hi @Branikolog , as usual you would like to have an ideal solution in one shot. I will rephrase my request: are these changes making the logic better?
> 
> Or, what is much more important, do they break anything :)

Precisely! We don't need to fix everything at once. We can do step by step to achieve the goal.

src/fheroes2/battle/battle_cell.h
```diff
@@ -56,14 +56,19 @@ namespace Battle
...
         void ResetQuality( void );
-        void ResetDirection( void );
+        void ResetReachability();
 
         void SetObject( int );
-        void SetDirection( int );
         void SetQuality( u32 );
 
+        void SetReachableForHead();
+        void SetReachableForTail();
```
ihhub, line 65: To be consistent could we please name newly created method from non-capital letter?

src/fheroes2/battle/battle_interface.cpp
ihhub: Ah, I recollected. Please try with and without battlefield shadow option. So we display different type of cursor shadow based on this option being on or off.

ihhub (commented):
Hi @oleg-derevenetz , I left just few comments here. Could you please take a look?

Branikolog:
> > > Hi @Branikolog , as usual you would like to have an ideal solution in one shot. I will rephrase my request: are these changes making the logic better?
> > 
> > 
> > Or, what is much more important, do they break anything :)
> 
> Precisely! We don't need to fix everything at once. We can do step by step to achieve the goal.

Okay-okay. :) Gameplay has become better. No crashes or incorrect behaviour. )

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: Hi @ihhub I just made a few screenshots:

---

With battle movement shadow, this branch:

This_Branch_With_Movement_Shadow

With battle movement shadow, `master` branch:

Master_Branch_With_Movement_Shadow

---

Without movement shadow, this branch:

This_Branch_Without_Movement_Shadow

Without movement shadow, `master` branch:

Master_Branch_Without_Movement_Shadow

---

I don't see any difference here. Maybe I'm looking in the wrong place?

ihhub: Hmm... let's then proceed with this change if it does nothing wrong :)

ihhub:
@oleg-derevenetz , great thanks for these fixes!

PR #4023, opened 2021-08-17
Drop duplicated enum and replace some magic numbers with its values

qarkai:
Move enum from `Battle::Board` to `Battle::Arena`.
Fix typo FORTH -> FOURTH


ihhub:
@qarkai , thank you very much for this code correction!

Issue #4024, opened 2021-08-17
Catapult shoots 4 times with Exert Ballistics and the "Ballista of Quickness"


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

While possessing an artifact "Ballista of Quickness" a hero with expert Ballistics makes catapult shoot 4 times in a row during a siege. While in the original game Catapult makes 3 shots maximum.
![image](https://user-images.githubusercontent.com/55348946/129730222-89013a67-6b67-4863-a9c2-c70998430b97.png)

### Save file

Campaign save file:
[Evil_6_0034.zip](https://github.com/ihhub/fheroes2/files/7000106/Evil_6_0034.zip)

PR #4025, opened 2021-08-17
Fix max ballista shots

tau3:
Hi! Here is a fix for #4024

### DIFF
--- a/src/fheroes2/battle/battle_catapult.cpp
+++ b/src/fheroes2/battle/battle_catapult.cpp
@@ -56,11 +56,6 @@ Battle::Catapult::Catapult( const HeroBase & hero )
 
     if ( hero.HasArtifact( Artifact::BALLISTA ) )
         catShots += Artifact( Artifact::BALLISTA ).ExtraValue();
-
-    const int maxCatShots = 3;
-    if ( catShots > maxCatShots ) {
-        catShots = maxCatShots;
-    }
 }
 
 u32 Battle::Catapult::GetDamage() const
--- a/src/fheroes2/battle/battle_catapult.cpp
+++ b/src/fheroes2/battle/battle_catapult.cpp
@@ -58,7 +58,7 @@ Battle::Catapult::Catapult( const HeroBase & hero )
         catShots += Artifact( Artifact::BALLISTA ).ExtraValue();
 
     const int maxCatShots = 3;
-    if( catShots > maxCatShots) {
+    if ( catShots > maxCatShots ) {
         catShots = maxCatShots;
     }
 }
--- a/src/fheroes2/battle/battle_catapult.cpp
+++ b/src/fheroes2/battle/battle_catapult.cpp
@@ -56,6 +56,11 @@ Battle::Catapult::Catapult( const HeroBase & hero )
 
     if ( hero.HasArtifact( Artifact::BALLISTA ) )
         catShots += Artifact( Artifact::BALLISTA ).ExtraValue();
+
+    const int maxCatShots = 3;
+    if( catShots > maxCatShots) {
+        catShots = maxCatShots;
+    }
 }
 
 u32 Battle::Catapult::GetDamage() const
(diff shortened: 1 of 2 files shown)


src/fheroes2/battle/battle_catapult.cpp
```diff
@@ -56,6 +56,11 @@ Battle::Catapult::Catapult( const HeroBase & hero )
 
     if ( hero.HasArtifact( Artifact::BALLISTA ) )
         catShots += Artifact( Artifact::BALLISTA ).ExtraValue();
+
+    const int maxCatShots = 3;
+    if ( catShots > maxCatShots ) {
+        catShots = maxCatShots;
+    }
```
qarkai, line 63:
```suggestion
    catShots = std::max( catShots, maxCatShots );
```

oleg-derevenetz, line 63: Hi @tau3 may be just change `extra` value for `Ballista of Quickness` in `resource/artifact.cpp` from 2 to 1? Hi @Branikolog I suppose this artifact should just give one extra shot?

tau3: Hi! I was afraid of changing that static configuration and went the other way, but yes, your suggestion is better :)
Lets wait for a response

oleg-derevenetz: @tau3 This artifact has a misleading description - what does "lets your catapult fire twice per round" mean: is this mean "one additional shot" or "two additional shots".

Branikolog:
Hi, @oleg-derevenetz !
In the original game this artifact adds I've extra shot, allowing to shot twice. If hero has expert Ballistics, catapult shots 3 times.

oleg-derevenetz:
Hi @Branikolog

> In the original game this artifact adds I've extra shot, allowing to shot twice. If hero has expert Ballistics, catapult shots 3 times. 

So it seems that currently this artifact has wrong `extra` value of 2, and we can replace it by 1 instead...

Branikolog:
> Hi @Branikolog
> 
> > In the original game this artifact adds I've extra shot, allowing to shot twice. If hero has expert Ballistics, catapult shots 3 times.
> 
> So it seems that currently this artifact has wrong `extra` value of 2, and we can replace it by 1 instead...

Correct!
Even without "Ballistics" catapult shoots 3 times.

oleg-derevenetz:
@Branikolog 

> Correct!
> Even without "Ballistics" catapult shoots 3 times.

So with this artifact catapult always shoots 3 times, and Ballistics skill doesn't matter?

Branikolog:
> @Branikolog
> 
> > Correct!
> > Even without "Ballistics" catapult shoots 3 times.
> 
> So with this artifact catapult always shoots 3 times, and Ballistics skill doesn't matter?

Yes. In the OG Catapult shoots 3 times maximum. 
4 shots looks kind of unfair. )

LeHerosInconnu:
Hello everyone,

> @tau3 This artifact has a misleading description - what does "lets your catapult fire twice per round" mean: is this mean "one additional shot" or "two additional shots".

The "Ballista of Quickness" gives "one additional shot".

Here's how it should work:

Ballistics; Increases the catapult's shots, damage and accuracy in siege combat.
Basic Level: 1 shot with extra damage.
Advanced Level: 2 shots with extra damage.
Expert Level: 2 shots with max damage.

Without Ballistics: 1 shot.
Without Ballistics and with Ballista of Quickness: 1 shot + 1 shot = 2 shots.
With Ballistics Basic: 1 shot with extra damage.
With Ballistics Basic and Ballista of Quickness: 1 shot with extra damage + 1 shot = 2 shots with extra damage.
With Balllistics Advanced: 2 shots with extra damage
With Ballistics Advanced and Ballista of Quickness: 2 shots with extra damage + 1 shot = 3 shots with extra damage.
With Ballistics Expert: 2 shots with max damage.
With Ballistics Expert and Ballista of Quickness: 2 shots with max damage + 1 shot = 3 shots with max damage.

The extra value for "Ballista of Quickness" should be set to "1".
When the hero has Ballistics, logically, the Ballista of Quickness's extra shot should also have the extra damage bonus and the maximum damage bonus.

oleg-derevenetz (approved):

ihhub:
@tau3 , thank you very much for the fix!

PR #4026, opened 2021-08-17
Highlight Cells refactoring

tau3:
Hello! 

Saw a code smell warning on sonarcloud. Decided to extract `Battle::Interface::CalculateHighlightCells` and split into smaller members without logic changes. This is only the first iteration, some of resulting members are still bloated and should be refactored, but current changeset is already not so small.

Please, review

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1529,56 +1529,141 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
     text.Blit( sx + ( bar.width() - text.w() ) / 2, sy, _mainSurface );
 }
 
+std::set<const Battle::Cell *> Battle::Interface::CalculateHighlightCellsOnValidSpell( const Cell * cell ) const
+{
+    std::set<const Battle::Cell *> highlightCells;
+
+    switch ( humanturn_spell.GetID() ) {
+    case Spell::COLDRING: {
+        const Indexes around = Board::GetAroundIndexes( index_pos );
+        for ( size_t i = 0; i < around.size(); ++i ) {
+            const Cell * aroundCell = Board::GetCell( around[i] );
+            if ( aroundCell != nullptr ) {
+                highlightCells.emplace( aroundCell );
+            }
+        }
+        break;
+    }
+    case Spell::FIREBALL:
+    case Spell::METEORSHOWER: {
+        highlightCells.emplace( cell );
+        const Indexes around = Board::GetAroundIndexes( index_pos );
+        for ( size_t i = 0; i < around.size(); ++i ) {
+            const Cell * aroundCell = Board::GetCell( around[i] );
+            if ( aroundCell != nullptr ) {
+                highlightCells.emplace( aroundCell );
+            }
+        }
+        break;
+    }
+    case Spell::FIREBLAST: {
+        highlightCells.emplace( cell );
+        const Indexes around = Board::GetAroundIndexes( index_pos );
+        for ( size_t i = 0; i < around.size(); ++i ) {
+            const Cell * aroundCell = Board::GetCell( around[i] );
+            if ( aroundCell != nullptr ) {
+                highlightCells.emplace( aroundCell );
+            }
+
+            const Indexes aroundTwice = Board::GetAroundIndexes( around[i] );
+            for ( size_t j = 0; j < aroundTwice.size(); ++j ) {
+                const Cell * aroundCellTwice = Board::GetCell( aroundTwice[j] );
+                if ( aroundCellTwice != nullptr ) {
+                    highlightCells.emplace( aroundCellTwice );
+                }
+            }
+        }
+        break;
+    }
+    default:
+        highlightCells.emplace( cell );
+    }
+    return highlightCells;
+}
+
+int CursorTypeToDirection( const int cursorType )
+{
+    int direction = 0;
+    if ( cursorType == Cursor::SWORD_TOPLEFT ) {
+        direction = Direction::BOTTOM_RIGHT;
+    }
+    else if ( cursorType == Cursor::SWORD_TOPRIGHT ) {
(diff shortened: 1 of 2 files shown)


tau3:
maybe all this cursor deciding logic can be moved out to a separate class because `Interface` already has to much members and functions

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1521,190 +1529,228 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
...
-            else {
-                assert( 0 );
-            }
+    }
+    return highlightCells;
+}
 
-            const Cell * attackerCell = Board::GetCell( index_pos, direction );
-            assert( attackerCell != nullptr );
+std::set<const Battle::Cell *> Battle::Interface::CalculateHighlightCellsForBigUnit( const Cell * cell ) const
```
qarkai, line 1661: AFAIK it's called wide unit,

ihhub:
Hi @tau3 , could you please solve merge conflict in this pull request?

tau3:
well, that merge was tough

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1540,6 +1540,156 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
...
+
+    switch ( humanturn_spell.GetID() ) {
+    case Spell::COLDRING: {
+        const Indexes around = Board::GetAroundIndexes( index_pos );
+        for ( const int32_t index : around ) {
+            const Cell * aroundCell = Board::GetCell( index );
+            if ( aroundCell != nullptr ) {
+                highlightCells.emplace( aroundCell );
+            }
+        }
```
qarkai, line 1555: This repeats at least four times. Could you move it to function please?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1540,6 +1540,156 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
...
+    case Spell::FIREBALL:
+    case Spell::METEORSHOWER: {
+        highlightCells.emplace( cell );
+        const Indexes around = Board::GetAroundIndexes( index_pos );
+        for ( const int32_t index : around ) {
+            const Cell * aroundCell = Board::GetCell( index );
+            if ( aroundCell != nullptr ) {
+                highlightCells.emplace( aroundCell );
+            }
+        }
```
qarkai, line 1567: This could be replaced with `CalculateHighlightCellsOnAreaShot`

src/fheroes2/battle/battle_interface.cpp
ihhub: I agree on this. If we want to refactor we should remove duplicated code.

src/fheroes2/battle/battle_interface.h
```diff
@@ -293,6 +294,12 @@ namespace Battle
         int GetBattleCursor( std::string & ) const;
         int GetBattleSpellCursor( std::string & ) const;
 
+        std::set<const Cell *> CalculateHighlightCells( const Cell *, int ) const;
+        std::set<const Cell *> CalculateHighlightCellsOnValidSpell( const Cell * ) const;
+        std::set<const Cell *> CalculateHighlightCellsOnSwordCursor( int ) const;
+        std::set<const Cell *> CalculateHighlightCellsForWideUnit() const;
+        std::set<const Cell *> CalculateHighlightCellsOnAreaShot( const Cell * ) const;
```
ihhub, line 301: Please add the names of input arguments and name the methods from non-capital letter per our coding standard. Also `int` arguments could be **const** as well.

ihhub: Actually we don't calculate anything in these methods. We could call them something like `getHightlighedCells` for example.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1540,6 +1540,156 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
...
+        direction = Direction::RIGHT;
+    }
+    else if ( cursorType == Cursor::SWORD_RIGHT ) {
+        direction = Direction::LEFT;
+    }
+    else {
+        assert( 0 );
+    }
+    return direction;
+}
```
ihhub, line 1620: Please put this function into unnamed namespace within this file.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1540,6 +1540,156 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
...
+        highlightCells.emplace( cell );
+    }
+    return highlightCells;
+}
+
+int CursorTypeToDirection( const int cursorType )
+{
+    int direction = 0;
+    if ( cursorType == Cursor::SWORD_TOPLEFT ) {
+        direction = Direction::BOTTOM_RIGHT;
```
ihhub, line 1599: We don't need to have `direction` variable in this function. We can directly return here without assignment.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1540,6 +1540,156 @@ void Battle::Interface::RedrawTroopCount( const Unit & unit )
...
+    }
+    else {
+        assert( 0 );
+    }
+    return direction;
+}
+
+std::set<const Battle::Cell *> Battle::Interface::CalculateHighlightCellsOnSwordCursor( int cursorType ) const
+{
+    int direction = CursorTypeToDirection( cursorType );
```
ihhub, line 1624: This variable could be `const`.

ihhub (changes requested):
Hi @tau3 , I left few comments here as well with @qarkai . Could you please take a look?

ihhub:
Hi @tau3 , if we're still going with this change please fix merging conflicts for this pull request.

PR #4030, opened 2021-08-20
Fix multiple SonarQube Smells

ihhub:
- fix rendering of text broken at #3898
- remove potentially dangerous constructors from main classes
- fix multiple-statements-per-line SonarQube smells
- fix empty-function-body SonarQube smells
- add missing member initializations of few classes
- fix incorrect casting of variables causing object copying


ihhub:
We have to ignore all these clang-tidy comments as changes were targeted for different issues.

oleg-derevenetz (changes requested):
Hi @ihhub I left just one suggestion here (I believe it should also fix the corresponding clang-tidy issue).

PR #4034, opened 2021-08-20
Game::EnvironmentSoundMixer(): remove extra Audio::isValid() check that blocks the main thread

### LINKED ISSUE #4031: The game freezes while hero crossing a new terrain
### Preliminary checks

### Platform

Windows

### Describe the bug

Run the latest version and walk by a hero through different terrains with MIDI. Notice lags while the hero switches between terrains.

### Save file

Choose Roxana and try to attack Master Swordsmen monsters on the left from the hero.
[Wlr.zip](https://github.com/ihhub/fheroes2/files/7021975/Wlr.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4031

Hi @ihhub I was unable to reproduce this issue on my own hardware (apparently it's too fast), so I borrowed an ancient Lenovo T420, but it was taken back from me recently. Do you have a hardware slow enough to test this patch?

### DIFF
--- a/src/fheroes2/game/game.cpp
+++ b/src/fheroes2/game/game.cpp
@@ -351,10 +351,6 @@ u32 & Game::CastleAnimationFrame( void )
 /* play all sound from focus area game */
 void Game::EnvironmentSoundMixer( void )
 {
-    if ( !Audio::isValid() ) {
-        return;
-    }
-
     const fheroes2::Point abs_pt( Interface::GetFocusCenter() );
     std::fill( reserved_vols.begin(), reserved_vols.end(), 0 );
 


ihhub:
> 
> 
> fix #4031
> 
> Hi @ihhub I was unable to reproduce this issue on my own hardware (apparently it's too fast), so I borrowed an ancient Lenovo T420, but it was taken back from me recently. Do you have a hardware slow enough to test this patch?

Hi @oleg-derevenetz , I have brand new Asus VivoBook S14/S15 which has such problem :) I'll test it soon once I finish changes for another bug fix.

ihhub:
@oleg-derevenetz , thanks a lot for this fix! I tested on my machine and it works as a charm.

vincent-grosbois:
Hello @oleg-derevenetz  @ihhub 
just thought about this yesterday, sorry I saw the PR is closed now but I think it's worth commenting:
the root cause seems to be that Audio::IsValid is trying to acquire the lock, just to check if the bool isValid is set or not.
I think, instead, what can be done is that Audio::IsValid doesn't lock anything, and just replace the bool isValid with a std::atomic<bool> isValid. With this, all checks to Audio::IsValid will never wait for the lock from audio thread.

I don't think there is any race condition issues about doing this, since the valid flag for audio is only set up once at startup and at cleanup

oleg-derevenetz:
Hi @vincent-grosbois I already thought about making `valid` atomic, but I think currently it doesn't worth it: if some code checks `isValid()`, then a few lines later something from `Mixer::` or `Music::` will be called and lock will be required anyway. This place was an exception (with useless check I should say, because `isValid()` would be checked anyway a bit later).

vincent-grosbois:
mmm... I think it's worth it anyway as Audio::isValid is public, so this kind of bug is bound to happen anyway in the future (ie someone calling this function in main thread without thinking it will lock)

oleg-derevenetz:
@vincent-grosbois Calling `isValid()` is usually a prerequisite for calling something from `Mixer::` or `Music::`. I suppose that it was called from `EnvironmentSoundMixer()` because of call to `Mixer::MaxVolume()`, but it doesn't require this check because it returns a constant. But why not, I can make it atomic. All functions from `Mixer::` or `Music::` have their own guard locks, so it wouldn't do any harm.

PR #4038, opened 2021-08-21
Add object count for weekly object income message

### LINKED ISSUE #2153: Adventure map, message windows, option "world: Wind/Water Mills and Magic Garden can be captured", display the plural names in the title
In fheroes2, when the option "world: Wind/Water Mills and Magic Garden can be captured" is activated, display the plural names in the title when there are several waterwheels, magic gardens and windmills controlled.

In fheroes2:

Plural title window fh2 01

Plural title window fh2 02

Plural title window fh2 03

ihhub:
close #2153

Also add translation support for some missing strings.


ihhub:
Hi @Branikolog and @LeHerosInconnu, could you please check this change as well? As we don't have plural forms for object types on map I added " (N places)" to the message's title if the number of objects with income more than 1. I know this is not the best solution and we can change name but I think it's the best in terms of not touching the code and adding complexity for extra translations.

LeHerosInconnu:
@ihhub.

> Hi @Branikolog and @LeHerosInconnu, could you please check this change as well? As we don't have plural forms for object types on map I added " (N places)" to the message's title if the number of objects with income more than 1. I know this is not the best solution and we can change name but I think it's the best in terms of not touching the code and adding complexity for extra translations.

I think the word "places" is not really necessary, the number in parenthesis is enough.
Also, as it is rather common to have several of the same place, the plural can be used by default in the title.
So:
"Water Wills (N)"
"Windmills (N)"
"Magic Gardens (N)"

I also made a proposal to combine the three windows into one: https://github.com/ihhub/fheroes2/discussions/4047.

ihhub:
Hi @LeHerosInconnu , as I wrote before making words in **plural** form is troublesome. Hence this change is proposed.

LeHerosInconnu:
In this case put the singular with only the number in parenthesis.

So:
"Water Will (N)"
"Windmill (N)"
"Magic Garden (N)"

Branikolog:
@ihhub 
I can't offer any good solutions without changing the title or adding some new text into this window... :(
May be, "Magic Garden (2x)"? But I'm far from satisfaction with this...

ihhub:
I will keep this pull request open for any other future ideas :)

oleg-derevenetz (approved):

src/fheroes2/world/world.cpp
```diff
@@ -298,22 +298,22 @@ void CapturedObjects::ResetColor( int color )
     }
 }
 
-Funds CapturedObjects::TributeCapturedObject( int color, int obj )
+void CapturedObjects::tributeCapturedObjects( const int playerColorId, const int objectType, Funds & funds, int & objectCount )
```
qarkai, line 301: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f21-to-return-multiple-out-values-prefer-returning-a-struct-or-tuple

oleg-derevenetz: Tuples are not welcome in this project because of their anonymous nature (it is not obvious what all these tuple fields are intended for), and making a special struct that will be used just in one place may be a bit too much.

PR #4042, opened 2021-08-21
make.yml: make translation generation a separate step, use make -C instead of cd && make

oleg-derevenetz:


ihhub:

PR #4044, opened 2021-08-21
Audio: make valid atomic

oleg-derevenetz:
Discussed in #4034


ihhub:

PR #4048, opened 2021-08-21
Don't set the fade-in for music on Windows

### LINKED ISSUE #2901: Horse's steps are fading while crossing terrain border using MIDI sound
When hero moves from one terrain type to another and the music is changing, horse's steps disappear for a while and then fade up in a few seconds.

oleg-derevenetz:
fix #2901

This solution is far from ideal, but it should work. Hi @Branikolog could you please verify that it works for you, and horse steps are not fading anymore?

### DIFF
--- a/src/fheroes2/game/fheroes2.cpp
+++ b/src/fheroes2/game/fheroes2.cpp
@@ -163,7 +163,11 @@ int main( int argc, char ** argv )
                 Mixer::Volume( -1, Mixer::MaxVolume() * conf.SoundVolume() / 10 );
 
                 Music::Volume( Mixer::MaxVolume() * conf.MusicVolume() / 10 );
+#if !defined( _MSC_VER ) && !defined( __MINGW32__ )
+                // On Windows, the MIDI music volume is used as the master volume,
+                // so this fade-in would be set for everything, including sound effects
                 Music::SetFadeIn( 900 );
+#endif
             }
 
             fheroes2::Display & display = fheroes2::Display::instance();


Branikolog:
Hi, @oleg-derevenetz !
I cannot hear steps fading anymore, but after terrain theme changes I hear a gap in any sound game plays: horse steps disappear for a bit, and then I can hear some sort of cut step sound, in a case when hero changes a terrain and stops right after that.
Also, if a hero stands near any object, that plays its own sound, and we switch to another hero, after switching we hear object sound being cut.

oleg-derevenetz:
@Branikolog

> I cannot hear steps fading anymore, but after terrain theme changes I hear a gap in any sound game plays: horse steps disappear for a bit, and then I can hear some sort of cut step sound, in a case when hero changes a terrain and stops right after that.

Yes, unfortunately, currently this gap still exists because mixer is busy processing new terrain music. Although this processing is performed in a separate thread and does not affect the game UI itself, it still affects other sounds, including horse steps, ambient sounds, etc. A possible solution would be to distribute the processing of music and sound effects across different threads, but AFAIK SDL_mixer is not thread-safe.

> Also, if a hero stands near any object, that plays its own sound, and we switch to another hero, after switching we hear object sound being cut.

What do you mean? When you switching to another hero, you shouldn't be able to hear object sounds near the previous hero anyway, because they are not near anymore :)

Branikolog:
> What do you mean? When you switching to another hero, you shouldn't be able to hear object sounds near the previous hero anyway, because they are not near anymore :)

Try to fast switch between hero near object, and the second hero standing on the other terrain. You can hear a small, short piece of object sound, while choosing second hero...

oleg-derevenetz:
@Branikolog

> Try to fast switch between hero near object, and the second hero standing on the other terrain. You can hear a small, short piece of object sound, while choosing second hero...

Ah, that's what you're talking about. I suppose that it's maybe because mixer manages to play a piece of an old object sound in the interval between loading new music/sounds. If it would be able to load new sounds faster (or if it was possible to parallelize it) this would not have happened.

Does this PR make things better than before?

Branikolog:
@oleg-derevenetz 
Well... Cannot tell, that changes are essential...
Before we could hear some cut sounds while choosing hero near object with sound... Now we can hear similar problem, while switching from the hero with sounds, to another hero...
So midi sound worked with issues... Now it has become... not good...
Horse steps are still interrupted...

oleg-derevenetz:
@Branikolog well, let's then postpone this change for a while :)

Branikolog:
@oleg-derevenetz 
Anyway, I really appreciate your effort! :)

PR #4054, opened 2021-08-21
Fix invalid reference in FilterResolutions

a1exsh:
The reference returned from `front()` is no longer valid after use of
`emplace_back`, so we should store current resolution by value instead.


a1exsh:
E.g. on my system the only resolution detected is `3840 x 2160` so I was getting that one and the default `640 x 480` in the list.  After changing to value instead of reference I'm seeing all the default possible resolutions, as expected.

oleg-derevenetz:
Hi @a1exsh

> The reference returned from `front()` is no longer valid after use of
> `emplace_back`, so we should store current resolution by value instead.

It is no longer valid only if reallocation occurs (vector capacity is changed). However, it seems that in this case vector capacity is always changed, because initial set contains only 3 resolutions, and `possibleResolutions` adds five variants. Anyway, storing a reference here is unnecessary and dangerous, no gain at all, copy is much safer.

ihhub:
@a1exsh , thank you so much for the fix!

PR #4055, opened 2021-08-21
Ensure strict ordering in SortResolutions

a1exsh:
Use of "greater or equal" at the last place is not correct.

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -70,7 +70,7 @@ namespace
 
     bool SortResolutions( const fheroes2::Size & first, const fheroes2::Size & second )
     {
-        return first.width > second.width || ( first.width == second.width && first.height >= second.height );
+        return first.width > second.width || ( first.width == second.width && first.height > second.height );
     }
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
@a1exsh It's not that I'm against this change, but does it really matter? Since two `Size` structs are equal, why the exact order matters? Does this break anything?

a1exsh:
It hasn't broken anything for me, but I guess it violates the `comp(a,a) == false` requirement: https://en.cppreference.com/w/cpp/named_req/Compare

Let's be on the safe side. ;)

oleg-derevenetz:
> It hasn't broken anything for me, but I guess it violates the `comp(a,a) == false` requirement: https://en.cppreference.com/w/cpp/named_req/Compare
> 
> Let's be on the safe side. ;)

Well, I agree. Since it is used in `std::sort()`, it should meet the `Compare` requirement.

ihhub:
@a1exsh , thanks a lot for this change!

PR #4056, opened 2021-08-21
Make check_code_format script executable

a1exsh:
To be able to run it just like `./script/tools/check_code_format.sh`.


ihhub:
@a1exsh , thank you very much for this fix!

PR #4035, opened 2021-08-21
Remember UI position in Kingdom Overview dialog

ihhub:
close #1846

PR #4057, opened 2021-08-21
clang-tidy: enable AllowSoleDefaultDtor option for cppcoreguidelines-special-member-functions check

oleg-derevenetz:
For example, in #4035 we have an abstract class with virtual defaulted destructor, no need to add copy/move constructors and assignment operators in such cases.


ihhub:

PR #4058, opened 2021-08-21
Add monsters into High Score dialog

### LINKED ISSUE #954: Add monsters in High Score board
Here is a look of our score board:
image

And this is how it should look:
image

Arthusppp:
Monster sprite added to the High Scores. Fixes #954.

Hello! My first real commit. Please be kind to a newbie programmer like me ;) 

What I have fixed: I have added a monster to the High Scores dialog. See my example below:
7AzZ774HdO

My (zipped) test high score file: 
[fheroes2.hgs.zip](https://github.com/ihhub/fheroes2/files/7026235/fheroes2.hgs.zip)

The issues/questions I have:
1. the monsterAnimationSequence variable is the same as in map_tiles.cpp:67. How to share it elegantly?
2. is there any better way to handle the "rating to monster" relationship?
3. I do not know if I handled the Redraw part correctly. I was trying to reuse map_tiles as much as possible, but... pls let me know.


src/fheroes2/game/game_highscores.cpp
```diff
@@ -181,10 +186,262 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
+Monster HGSData::GetMonsterByRating( u32 rating )
+{
+    const std::array<Monster::monster_t, 229> rating2monster = { Monster::PEASANT,
+                                                                 Monster::PEASANT,
+                                                                 Monster::PEASANT,
+                                                                 Monster::PEASANT,
+                                                                 Monster::GOBLIN,
+                                                                 Monster::GOBLIN,
+                                                                 Monster::GOBLIN,
+                                                                 /*7*/ Monster::GOBLIN,
```
ihhub, line 214: Do we need these comments with numbers?

src/fheroes2/game/game_highscores.cpp
```diff
@@ -181,10 +186,262 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
+                                                                 Monster::TITAN,
+                                                                 Monster::TITAN,
+                                                                 Monster::BLACK_DRAGON };
+
+    if ( rating > 228 )
+        rating = 228;
+
+    Monster monster = Monster( rating2monster[rating] );
+
+    return monster;
```
ihhub, line 442: We could directly return from the function as `return Monster( rating2monster[rating] );`. No need to create a temporary variable for this.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -257,6 +514,14 @@ fheroes2::GameMode Game::HighScores()
 
         if ( le.MouseClickLeft( buttonExit.area() ) || HotKeyCloseWindow )
             return fheroes2::GameMode::MAIN_MENU;
+
+        if ( Game::validateAnimationDelay( Game::MAPS_DELAY ) ) {
+            uint32_t & frame = Game::MapsAnimationFrame();
+            ++frame;
```
ihhub, line 520: Let's not use reference to World Map animation. We can have an internal counter within `HGSData` class and increment it within `RedrawList` method.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -99,6 +102,8 @@ class HGSData
 
 private:
     std::vector<hgs_t> list;
+    const std::array<uint8_t, 15> monsterAnimationSequence = { 0, 0, 1, 2, 1, 0, 0, 0, 3, 4, 5, 4, 3, 0, 0 };
+    Monster GetMonsterByRating( u32 rating );
```
ihhub, line 106: Please put this method at the top of private section with separation by an empty line from class members. Also use `uint32_t` instead of **u32**. Since this method does nothing to class it should be **static** with signature:
```
static Monster getMonsterByRating( const uint32_t rating );
```
Notice that the first letter is non-capital.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -99,6 +102,8 @@ class HGSData
 
 private:
     std::vector<hgs_t> list;
+    const std::array<uint8_t, 15> monsterAnimationSequence = { 0, 0, 1, 2, 1, 0, 0, 0, 3, 4, 5, 4, 3, 0, 0 };
```
ihhub, line 105: This is a duplication code from `maps_tiles.cpp`. Please create a separate function in monster_anim.h and monster_anim.cpp classes under fheroes2 namespace to return a const reference to the same container. This container should be located in anonymous (with no name) namespace at the beginning of monster_anim.cpp file. Then here and in maps_tile.cpp use newly created function.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -181,10 +186,262 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
         text.Set( std::to_string( hgs.rating ) );
         text.Blit( ox + 480, oy + 70 );
 
+        Monster monster = GetMonsterByRating( hgs.rating );
+        uint32_t spriteIndex = monster.GetSpriteIndex() * 9;
+        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::MINIMON, spriteIndex );
+        fheroes2::Blit( sprite, display, sprite.x() + ox + 500 + 32, sprite.y() + oy + 70 + 24 );
+
+        /*+hgs.days+ox+oy - to start from "random" animation frame when a monster occurs more than once in the high score list*/
```
ihhub, line 194: Please use single line comment.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -181,10 +186,262 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
         text.Set( std::to_string( hgs.rating ) );
         text.Blit( ox + 480, oy + 70 );
 
+        Monster monster = GetMonsterByRating( hgs.rating );
+        uint32_t spriteIndex = monster.GetSpriteIndex() * 9;
```
ihhub, line 190: These variables can be const.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -181,10 +186,262 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
+        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::MINIMON, spriteIndex );
+        fheroes2::Blit( sprite, display, sprite.x() + ox + 500 + 32, sprite.y() + oy + 70 + 24 );
+
+        /*+hgs.days+ox+oy - to start from "random" animation frame when a monster occurs more than once in the high score list*/
+        uint32_t animateIndex
+            = spriteIndex + 1
+              + monsterAnimationSequence[( static_cast<unsigned long long>( hgs.days ) + ox + oy + Game::MapsAnimationFrame() ) % monsterAnimationSequence.size()];
```
ihhub, line 197: This variable can be const.

ihhub, line 197: Why do we convert `hgs.days` into **unsigned long long**?

src/fheroes2/game/game_highscores.cpp
```diff
@@ -29,9 +30,11 @@
 #include "agg.h"
 #include "agg_image.h"
 #include "audio.h"
+#include "bin_info.h"
```
ihhub, line 33: Do we need this file inclusion?

ihhub (changes requested):
Hi @Arthusppp , I left multiple comments here. Would you mind to take a look?

LeHerosInconnu:
Hello @Arthusppp and @ihhub,

Would it be possible to adjust the position of the elements in these screens (at least the horizontal and vertical position of the creature sprites in this PR).

In original game:

High scores 20

High scores 21

src/fheroes2/game/game_highscores.cpp
qarkai: Or even `return { rating2monster[rating] };`

src/fheroes2/game/game_highscores.cpp
```diff
@@ -181,10 +186,262 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
+                                                                 Monster::RED_DRAGON,
+                                                                 Monster::RED_DRAGON,
+                                                                 Monster::RED_DRAGON,
+                                                                 /*225*/ Monster::TITAN,
+                                                                 Monster::TITAN,
+                                                                 Monster::TITAN,
+                                                                 Monster::BLACK_DRAGON };
+
+    if ( rating > 228 )
+        rating = 228;
```
qarkai, line 438: Please don't use magic numbers. It's `rating2monster.size() - 1`. Moreover to avoid duplicating it could be written as `rating = std::min(rating, rating2monster.size() - 1);`.

src/fheroes2/game/game_highscores.cpp
Arthusppp: If not, then my Visual says:

> Warning	C26451	Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2)

How I should handle it? Should I ignore this warning?
Or maybe sth like:

> + monsterAnimationSequence[( (size_t)ox + oy + hgs.days + Game::MapsAnimationFrame() ) % monsterAnimationSequence.size()];

ihhub: Yes, please ignore it for now. It's not a compilation warning after all.

src/fheroes2/game/game_highscores.cpp
Arthusppp: No we dont. Removed.

src/fheroes2/game/game_highscores.cpp
Arthusppp: Many thanks for this! Implemented with this:

`return Monster( rating2monster[std::min( rating, rating2monster.size() - 1 )] );`

src/fheroes2/game/game_highscores.cpp
Arthusppp: No problem :) I "have copied" the World Map animation algorithm. Please take a look and let me know.

src/fheroes2/game/game_highscores.cpp
Arthusppp: I have changed to size_t as the std::min claims that uint32_t (and most probably u32 as well) is "_template parameter _Ty is ambiguous_". What I try to achieve here is:

`return Monster( rating2monster[std::min( rating, rating2monster.size() - 1 )] );`

Please let me know whether this is good approach or not.

src/fheroes2/game/game_highscores.cpp
Arthusppp: Done as described above.

src/fheroes2/game/game_highscores.cpp
Arthusppp: Done.

src/fheroes2/game/game_highscores.cpp
Arthusppp: Done. I believe I have changed all variables like this to be const.

src/fheroes2/game/game_highscores.cpp
Arthusppp: Done.

Arthusppp: Ok, ignored and removed "static_cast" part.

src/fheroes2/game/game_highscores.cpp
Arthusppp: No, we do not this in the current implementation. Removed.

src/fheroes2/game/game_highscores.cpp
Arthusppp: Many thanks for this. I have implemented the following:

`return Monster( rating2monster[std::min( rating, rating2monster.size() - 1 )] );`

ihhub:
Hi @Arthusppp  and @LeHerosInconnu , let fix the alignment of elements in a separate pull request to simplify the changes here.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -257,6 +525,11 @@ fheroes2::GameMode Game::HighScores()
 
         if ( le.MouseClickLeft( buttonExit.area() ) || HotKeyCloseWindow )
             return fheroes2::GameMode::MAIN_MENU;
+
+        if ( hgs.validateAnimationDelay() ) {
```
ihhub, line 529: Why can't we use `Game::validateAnimationDelay( Game::MAPS_DELAY )` without introducing duplicated code?

src/fheroes2/game/game_highscores.cpp
```diff
@@ -96,9 +99,15 @@ class HGSData
...
     void RedrawList( int32_t ox, int32_t oy );
+    bool validateAnimationDelay( void );
 
 private:
     std::vector<hgs_t> list;
+
+    static Monster getMonsterByRatingStandardGame( const size_t rating );
+
+    std::chrono::time_point<std::chrono::steady_clock> _prevTime = std::chrono::steady_clock::now();
+    u32 hgs_animation_frame = 0;
```
ihhub, line 110: I think we should remove this code based on suggestion about **validateAnimationDelay** call.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -170,21 +179,280 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
+{
+    const std::chrono::duration<double> time = std::chrono::steady_clock::now() - HGSData::_prevTime;
+    const uint64_t passedMs = static_cast<uint64_t>( time.count() * 1000 + 0.5 );
+    if ( passedMs >= 200 ) {
+        _prevTime = std::chrono::steady_clock::now();
+        hgs_animation_frame++;
+        return true;
+    }
+    return false;
+}
```
ihhub, line 454: I think we should remove this code based on suggestion about **validateAnimationDelay** call.

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -2100,6 +2099,7 @@ std::pair<uint32_t, uint32_t> Maps::Tiles::GetMonsterSpriteIndices( const Tiles
     }
     else {
         const fheroes2::Point & mp = Maps::GetPoint( tileIndex );
+        const std::array<uint8_t, 15> monsterAnimationSequence = fheroes2::MonsterAnimationSequence();
```
ihhub, line 2102: Please use const reference to avoid object copy.

src/fheroes2/monster/monster_anim.cpp
```diff
@@ -134,6 +136,13 @@ namespace fheroes2
         _updateFrameInfo();
     }
 
+    const std::array<uint8_t, 15> monsterAnimationSequence = { 0, 0, 1, 2, 1, 0, 0, 0, 3, 4, 5, 4, 3, 0, 0 };
+
+    std::array<uint8_t, 15> MonsterAnimationSequence()
```
ihhub, line 141: This function must return const reference.

ihhub, line 141: Please rename it into `getMonsterAnimationSequence` to be even more easy to understand :)

src/fheroes2/monster/monster_anim.cpp
```diff
@@ -134,6 +136,13 @@ namespace fheroes2
         _updateFrameInfo();
     }
 
+    const std::array<uint8_t, 15> monsterAnimationSequence = { 0, 0, 1, 2, 1, 0, 0, 0, 3, 4, 5, 4, 3, 0, 0 };
```
ihhub, line 139: Please put this variable in unnamed namespace at the top of this file after header inclusions.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -170,21 +179,280 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
 
         text.Set( std::to_string( hgs.days ) );
-        text.Blit( ox + 420, oy + 70 );
+        text.Blit( ox + 79 + 324, oy + 70 );
 
         text.Set( std::to_string( hgs.rating ) );
-        text.Blit( ox + 480, oy + 70 );
+        text.Blit( ox + 79 + 405, oy + 70 );
+
+        const Monster monster = HGSData::getMonsterByRatingStandardGame( hgs.rating );
```
ihhub, line 193: We're constructing `const std::array<Monster::monster_t, 229> rating2monster` multiple times (even on stack memory). Why can't we do this before the loop and then just put the logic from the method here?

src/fheroes2/game/game_highscores.cpp
```diff
@@ -170,21 +179,280 @@ void HGSData::RedrawList( int32_t ox, int32_t oy )
...
-        text.Blit( ox + 260, oy + 70 );
+        text.Blit( ox + 79 + 165, oy + 70 );
 
         text.Set( std::to_string( hgs.days ) );
-        text.Blit( ox + 420, oy + 70 );
+        text.Blit( ox + 79 + 324, oy + 70 );
 
         text.Set( std::to_string( hgs.rating ) );
-        text.Blit( ox + 480, oy + 70 );
+        text.Blit( ox + 79 + 405, oy + 70 );
```
ihhub, line 191: Please combine all numbers. They're all magic numbers so no point to separate them even knowing the compiler will optimize the code.

ihhub (changes requested):
Hi @Arthusppp , here are multiple comments left. Would you mind to take a look?

ihhub:
@Arthusppp , thank you so much for this feature!

PR #4059, opened 2021-08-22
Don't ignore in<>out mismatch in Image Transpose

a1exsh:
Passing unexpected images is a programming error, we should bail out instead
of pretending nothing happened.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -2065,10 +2065,12 @@ namespace fheroes2
 
     void Transpose( const Image & in, Image & out )
     {
-        assert( !in.empty() );
         assert( !out.empty() );
-        assert( in.width() == out.height() && in.height() == out.width() );
 
+        if ( in.empty() || in.width() != out.height() || in.height() != out.width() ) {
+            out.reset();
+            return;
+        }
         const int32_t width = in.width();
         const int32_t height = in.height();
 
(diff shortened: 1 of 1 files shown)


src/engine/image.cpp
```diff
@@ -2065,8 +2065,9 @@ namespace fheroes2
 
     void Transpose( const Image & in, Image & out )
     {
-        if ( in.empty() || out.empty() || in.width() != out.height() || in.height() != out.width() )
-            return;
+        assert( !in.empty() );
+        assert( !out.empty() );
+        assert( in.width() == out.height() && in.height() == out.width() );
```
ihhub, line 2070: In this case we can't catch any issues in release mode. What's the benefit doing this in debug build only? Also there could be cases when some images could be missing (due to pirated version of the game) and if we do this we're allowing unexpected behavior on user's systems.

a1exsh: For release mode we may have a fallback that makes the out image a black rectangle.  With the latest released version it would be some uninitialized memory grarbage instead.

ihhub: Having a black area is better than a potential crash during application runtime. A crash is the last thing what users expect. Also we apply these functions while generating another images and some of them might be empty after all which shouldn't lead to a crash in debug mode.

a1exsh: Makes sense.  I've only left assert for the out image not being empty: it should always be there and such an error is better caught in debug mode.

ihhub:
@a1exsh , thank you very much for this change!

PR #4051, opened 2021-08-21
Fix some issues found by clang UBSan

oleg-derevenetz:

PR #4061, opened 2021-08-22
Eradicate more reinterpret_cast calls

oleg-derevenetz:
* Replace `reinterpret_cast` by `static_cast` or `dynamic_cast` where appropriate;
* Remove `reinterpret_cast` usage from the `bin2txt` tool - this led to unaligned accesses at least.

I compared the `bin2txt` output from `master` branch and from this branch for some `.bin` files - it seems that nothing has been broken.

### DIFF
--- a/src/tools/bin2txt.cpp
+++ b/src/tools/bin2txt.cpp
@@ -1,34 +1,9 @@
-#include <algorithm>
 #include <fstream>
 #include <iostream>
 #include <string>
-#include <type_traits>
 #include <vector>
 
-#include "endian_h2.h"
-
-namespace
-{
-    template <typename T, typename = typename std::enable_if<std::is_integral<T>::value || std::is_floating_point<T>::value>::type>
-    T getValue( const char * data, const size_t base, const size_t offset = 0 )
-    {
-        const char * begin = data + base + offset * sizeof( T );
-        const char * end = begin + sizeof( T );
-
-        T result;
-
-        // Data is originally stored using the little-endian byte order
-#if BYTE_ORDER == LITTLE_ENDIAN
-        std::copy( begin, end, reinterpret_cast<char *>( &result ) );
-#elif BYTE_ORDER == BIG_ENDIAN
-        std::reverse_copy( begin, end, reinterpret_cast<char *>( &result ) );
-#else
-        static_assert( false, "Unknown byte order" );
-#endif
-
-        return result;
-    }
-}
+#include "serialize.h"
 
 int main( int argc, char ** argv )
 {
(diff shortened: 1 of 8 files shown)


src/tools/bin2txt.cpp
```diff
@@ -1,8 +1,22 @@
...
 #include <vector>
 
+namespace
+{
+    template <typename T>
+    T getValue( const char * data, const size_t base, const size_t offset = 0 )
+    {
+        T result;
+
+        memcpy( &result, data + base + offset * sizeof( T ), sizeof( T ) );
```
a1exsh, line 14: This approach might be fixing alignment problems, but is still prone to little/big-endian mismatch, IMO.  It's better to extract individual bytes and construct the resulting value manually in a platform-independent way.

oleg-derevenetz: I already thought about this at the time, but I think that probability of running this game on a bigendian platform (ancient Solaris SPARC?) is very low. And if we will support platforms with different endianness, then, naturally, we will have to do something with this, and, I suppose, not just in this place :)

oleg-derevenetz: @a1exsh I already thought about this, but I believe that probability of running this game on bigendian platform (are there modern bigendian platforms around? ancient Solaris SPARC?) is very low :)

ciplogic: ARM is big endian

a1exsh: What about the recent Macs — these are ARM-based, so that might be different enough to justify caring about endianness.

oleg-derevenetz: @a1exsh

> What about the recent Macs — these are ARM-based, so that might be different enough to justify caring about endianness.

All modern ARM-based systems are running in the little-endian mode (and M1-based systems too, there are people that successfully play this game on M1). Big-endian systems are currently almost the same as ones' complement systems :) And, as I already said, support for big-endian systems is currently literally absent in the game as a whole.

oleg-derevenetz: @a1exsh I suppose if recent Macs would be big-endian, there would be a big headache with Rosetta on them :)))

oleg-derevenetz: @a1exsh However, you know what? You're right, let's implement proper endianness support along with fixing alignment issues. I'll do this today.

a1exsh: Don't know if that's worth the hassle right now. If you say people can play it on M1 Macs, I'm fine. :)

oleg-derevenetz: @a1exsh It was more hassle with CMake than with the implementation itself. What do you think about the current approach?

a1exsh: LGTM

a1exsh (approved):
Really nice!

oleg-derevenetz:
@ihhub BTW, I always playing fheroes2 with UB sanitizer enabled (`-fsanitize=undefined`) in a recent days and all that it has found so far - is what is fixed in this PR and in #4051. Not bad at all for a project of this size.

ihhub:
> 
> 
> @ihhub BTW, I always playing fheroes2 with UB sanitizer enabled (`-fsanitize=undefined`) in a recent days and all that it has found so far - is what is fixed in this PR and in #4051. Not bad at all for a project of this size.

Cool, we spent so much time fixing all possible warnings and bugs found by compilers so I'm not really surprised :)

oleg-derevenetz:
> Cool, we spent so much time fixing all possible warnings and bugs found by compilers so I'm not really surprised :)

I also used thread sanitizer to check the sound worker, but it's hard to use it all the time because it's a bit slow. There is also a memory sanitizer, but it is still experimental and not supported on macOS.

ihhub:
@oleg-derevenetz , thanks you very much for these fixes!

PR #4069, opened 2021-08-22
Calculate the number of stack-to-stack splitting options using the total quantity from both stacks when splitting to the stack with the same troop type

### LINKED ISSUE #3621: Castle screen, Creature separation window, the sentence "Fast separation into slots:" and the graphic "2" are not displayed when there are only 2 creatures in the starting slot during the distribution between the hero's army and the garrison

### LINKED ISSUE #4020: The number of stack-to-stack splitting options is limited with the quantity of base stack
### Preliminary checks

We're trying to split 2 archers over the stack of 4.
image
So we can only create two same size stacks:
image
While trying to split larger base stack...
image
...gives us an opportunity to create more same size stacks:
image

In a case we split one stack over another we should sum the overall creature quantity and offer a player all possible options.
Ex: if we split 2 archers over 4, we should be able to get up to 6 stacks of the same size.

### Additional info

_No response_

oleg-derevenetz:
close #4020
fix #3621 once again

### DIFF
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -44,15 +44,18 @@ namespace
         const bool isSameTroopType = troopTarget.isValid() && troopFrom.GetID() == troopTarget.GetID();
         const uint32_t overallCount = isSameTroopType ? troopFrom.GetCount() + troopTarget.GetCount() : troopFrom.GetCount();
 
-        if ( overallCount <= 1 ) {
-            // cross-army split logic - prevent splits where we'd lose the last stack of a hero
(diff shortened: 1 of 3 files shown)


ihhub:

PR #4070, opened 2021-08-22
My first PR, use vector instead of a list

ciplogic:
It is quicker to create `std::vector` than `std::list` but this is also made as a PR to get used with the flow of contributing.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -211,7 +211,7 @@ std::string System::GetMessageLocale( int length /* 1, 2, 3 */ )
         // 2: en_us
         // 1: en
         if ( length < 3 ) {
-            std::list<std::string> list = StringSplit( locname, length < 2 ? "_" : "." );
+            std::vector<std::string> list = StringSplit( locname, length < 2 ? "_" : "." );
             return list.empty() ? locname : list.front();
         }
     }
--- a/src/engine/tinyconfig.cpp
+++ b/src/engine/tinyconfig.cpp
@@ -56,9 +56,9 @@ bool TinyConfig::Load( const std::string & cfile )
     if ( !sf.open( cfile, "rb" ) )
         return false;
 
-    std::list<std::string> rows = StringSplit( sf.toString(), "\n" );
+    std::vector<std::string> rows = StringSplit( sf.toString(), "\n" );
 
-    for ( std::list<std::string>::const_iterator it = rows.begin(); it != rows.end(); ++it ) {
+    for ( auto it = rows.begin(); it != rows.end(); ++it ) {
         std::string str = StringTrim( *it );
         if ( str.empty() || str[0] == comment )
             continue;
--- a/src/engine/tinyconfig.cpp
+++ b/src/engine/tinyconfig.cpp
@@ -58,9 +58,8 @@ bool TinyConfig::Load( const std::string & cfile )
 
     std::vector<std::string> rows = StringSplit( sf.toString(), "\n" );
 
-    for (auto & row : rows) {
-row
-        std::string str = StringTrim( *it );
+    for ( auto & row : rows ) {
+        std::string str = StringTrim( row );
         if ( str.empty() || str[0] == comment )
             continue;
 
--- a/src/engine/tinyconfig.cpp
+++ b/src/engine/tinyconfig.cpp
@@ -58,7 +58,8 @@ bool TinyConfig::Load( const std::string & cfile )
 
     std::vector<std::string> rows = StringSplit( sf.toString(), "\n" );
 
-    for ( auto it = rows.begin(); it != rows.end(); ++it ) {
+    for (auto & row : rows) {
+row
         std::string str = StringTrim( *it );
         if ( str.empty() || str[0] == comment )
             continue;
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -163,9 +163,9 @@ void StringReplace( std::string & dst, const char * pred, int value )
     StringReplace( dst, pred, std::to_string( value ) );
 }
 
-std::list<std::string> StringSplit( const std::string & str, const std::string & sep )
+std::vector<std::string> StringSplit( const std::string & str, const std::string & sep )
 {
-    std::list<std::string> list;
+    std::vector<std::string> list;
     size_t pos1 = 0;
     size_t pos2 = std::string::npos;
 
--- a/src/engine/tools.h
+++ b/src/engine/tools.h
@@ -38,7 +38,7 @@ int Sign( int );
 std::string StringTrim( std::string );
 std::string StringLower( std::string );
 
-std::list<std::string> StringSplit( const std::string &, const std::string & );
+std::vector<std::string> StringSplit( const std::string &, const std::string & );
 
 void StringReplace( std::string &, const char *, const std::string & );
 void StringReplace( std::string &, const char *, int );
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -157,13 +157,12 @@ struct mofile
             buf.seek( offset2 );
             std::vector<std::string> tags = StringSplit( buf.toString( length2 ), "\n" );
 
-            for (auto & tag : tags) {
-tagtag
+            for ( auto & tag : tags ) {
                 if ( encoding.empty() )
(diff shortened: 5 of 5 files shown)


oleg-derevenetz:
> It is quicker to create `std::vector` than `std::list`

This is a controversial statement. When you need to append multiple items to the container (moreover, if their number is unknown in advance) `std::list` may be more efficient because, unlike `std::vector`, it does not need to perform constant reallocations. And if the elements are accessed only sequentially, I don't see a reason to replace list by vector here.

ciplogic:
@oleg-derevenetz "constant rellocations" do not happen, For illustration for 10 items to be added (excluding if we call "reserve" on a vector), by default there are 3 allocations: 4 items, 8 items and 16 items, and the cost of rellocation is much smaller than allocation itself. Yet for a list of 10 items, the cost is of 10 allocations.

Lists also have different other issues like they increase memory fragmentation and conversely will slow down memory allocation.

oleg-derevenetz:
> @oleg-derevenetz "constant rellocations" do not happen, For illustration for 10 items to be added (excluding if we call "reserve" on a vector), by default there are 3 allocations: 4 items, 8 items and 16 items, and the cost of rellocation is much smaller than allocation itself. Yet for a list of 10 items, the cost is of 10 allocations.

So we need to move 4 items to a newly allocated memory chunk first (after first realloc 4->8), and move additional 8 items to a newly allocated memory chunk later once again (after second realloc 8->16).

> Lists also have different other issues like they increase memory fragmentation and conversely will slow down memory allocation.

This PR currently looks like "change for the sake of change" for me, sorry. If it doesn't solve some specific explicit and measurable issue, I strongly doubt that it will be merged.

ciplogic:
This is fine not to merge it. 

I did it because on "development page" is written the following: 

"Contribution
We welcome and appreciate any help, even if it is a tiny text or code change. "

As for me the sake of the PR was written in my first change: yes, excluding that it will technically will make the code marginally faster, was to check the flow of adding a PR, and how the community works.

I am not here to say that this change will improve performance (it does, I can make you a micro-benchmark if this is what you care), but it would make it sure that I can add more changes.

I did test for 10 items, and for 100K calls the timings with VS 2019/Release are the following: (List is first, vector is second):
```
T1:328
T2:63
```
For the following test code:
```
static void UseList() {
    int counter = 0;
    for (int t = 0; t < 100000; t++) {
        std::list<std::string> items;
        for (int i = 0; i < 10; i++) {
            items.push_back(std::to_string(i));
        }
        counter += items.size();
    }
    std::cout << counter << '\n';
}

static void UseVector() {
    int counter = 0;
    for (int t = 0; t < 100000; t++) {
        std::vector<std::string> items;
        for (int i = 0; i < 10; i++) {
            items.push_back(std::to_string(i));
        }
        counter += items.size();
    }
    std::cout << counter << '\n';

}
```

oleg-derevenetz:
> I did it because on "development page" is written the following:
> 
> "Contribution
> We welcome and appreciate any help, even if it is a tiny text or code change. "

There are lots (380+) of **_real_** issues opened, any help with them will be really appreciated.

> I did test for 10 items, and for 100K calls the timings with VS 2019/Release are the following: (List is first, vector is second):

I didn't mean some synthetic tests. By "measurable issue" I mean some issue that may be observed "in the field" for current application. For example "in this case game is working slow because of this, so we need to do that, and this issue will be fixed". All these premature optimizations based on synthetic tests frankly looks useless.

ciplogic:
As I said, this is not a "premature optimization", was a PR just to check the community. It is an optimization but if you read a more recent C++ book of C++, it is recommended to use vectors over lists. I can say this as a hired for cash C++ developer with training.

You introduced the argument that the code I am introducing is slower. It isn't.

As for reference I can do larger refactors if I can do simpler ones. 

I am one person that touched this code-base on my own fork, and I have patches (obviously if I cannot pass simple patches, my interest will whine quickly to do larger ones) working on the same codebase. Please check here: 
https://github.com/ciplogic/fheroes2enh

As for fixing real issues, I will do it, obviously, as you do it, but if critical issues are the concern, I can say (and I checked especially if you are that picky) that I have seen PRs that do simple cleanups and nothing is fixed or improved (in fact it looks that one commit broke stuff) like this PR:
https://github.com/ihhub/fheroes2/pull/4061

Kind regards!

oleg-derevenetz:
> You introduced the argument that the code I am introducing is slower. It isn't.

That's not what I said. What I said is that no need to make changes in the sake of changes.

> I am one person that touched this code-base on my own fork, and I have patches (obviously if I cannot pass simple patches, my interest will whine quickly to do larger ones) working on the same codebase. Please check here:
> https://github.com/ciplogic/fheroes2enh

Any patches that will solve real issues are really appreciated.

> As for fixing real issues, I will do it, obviously, as you do it, but if critical issues are the concern, I can say (and I checked especially if you are that picky) that I have seen PRs that do simple cleanups and nothing is fixed or improved (in fact it looks that one commit broke stuff) like this PR:
> #4061

This PR fixes multiple **_real_** issues (caught by clang UBSan) with potential unaligned accesses on platforms where unaligned access turns into a CPU trap.

ciplogic:
```That's not what I said. ```

This is what you said:
```
This is a controversial statement. When you need to append multiple items to the container (moreover, if their number is unknown in advance) std::list may be more efficient because, unlike std::vector, it does not need to perform constant reallocations. And if the elements are accessed only sequentially, I don't see a reason to replace list by vector here.
```

For me it was enough 

Enjoy doing your game great again!

**Comment by oleg-derevenetz** [comment]
> Enjoy doing your game great again!

That's what I do in a daily basis :) You are welcome to join.

**Comment by ciplogic** [comment]
I rather not! But thanks for the invitation!

It was a pleasure!

**Comment by ihhub** [comment]
Hi @ciplogic , as per @oleg-derevenetz remark before doing such micro-optimizations (if they are needed) we have to measure them within the real application. Each container has pros and cons. Also theoretical knowledge sometimes does not apply for all situations as compiler is smart enough to optimize the code better.

Taking some remark about code personally is not a good habit and GitHub is definitely not a place for personal issues to come. Every developer here usually receives a bunch of comments in respect to their code and we shouldn't ever take them personally as the motivation of review is to get at the end the best code with less efforts, readable and fast enough.

I still repeat what has been said before you're always welcome to contribute to this project.

**Comment by ciplogic** [comment]
Yes, I grant you this @ihhub 

As for the claim: I personally think that micro-optimizations do not matter per se, every one taken individually, but it missed the point at least for me as for PR

What I found it frustrating are the following: 

- the PR was to test waters and it was according to the guidelines of contributing acceptable. Many small fixes do get at the end better runtime. And this is not only "just for show", really, many small fixes do add-up to a different product and code-base. This is why you have hundreds of small issues and they add up to make a polished software (and codebase)

- I personally don't care if it is a vector or a list if the code works, yet the first reasoning and double down-ing on it was that: "list is faster than vector" which is a false statement. This is also important for me on other level: "Also theoretical knowledge sometimes does not apply for all situations as compiler is smart enough to optimize the code better." - this is not true yet of 2021 and I know this writing a compiler, writing transpilers and optimizing compilers to not be the case. 

- small fixes like this should be accepted and even they are not "in the 380 issues", as because if added many enough, people get comfortable to fix larger items. For example, if you know that the person can fix small sized issues, the same person later you trust on medium sized issues

- last, but probably not least: your contribution guidelines does promote small fixes, and I found it to be false, and people depending on their time and energy can do these first. To me was to test waters, and I don't find them comfortable enough to swim in them.

Thank you for a detailed answers. 

Yes, it is not optimal to argue, yet the best approach is to do what you are happy to do it and I think that it is reasonable to say that you are not happy, and I was not happy, so as you see, I closed m PR, I deleted my fork and I will not argue.

Another independent source:
```Conclusion
To conclude, we can get some facts about each data structure:

std::vector is insanely faster than std::list to find an element
std::vector performs always faster than std::list with very small data
std::vector is always faster to push elements at the back than std::list
std::list handles very well large elements, especially for sorting or inserting in the front
This draw simple conclusions on usage of each data structure:

Number crunching: 

use std::vector
Linear search: use std::vector
Random Insert/Remove: use std::list (if data size very small (< 64B on my computer), use std::vector)
Big data size: use std::list (not if intended for searching)
```
https://baptiste-wicht.com/posts/2012/11/cpp-benchmark-vector-vs-list.html

PR #4072, opened 2021-08-23
Restart of the battle should lead to the same results

### LINKED ISSUE #3996: Luck/Morale effects happen randomly in battle on each re-load.
### Preliminary checks

### Platform

Windows

### Describe the bug

Steps:
1. Start a battle by hero with some luck/morale bonus (save the game before battle).
2. Make some moves to make luck/morale effect happened.
For example, hero's gargoyles received additional move cause of morale bonus.
Then, next hero's unit - centaurs - did not receive it.
3. Quit battle, load saved game.

Result:
 - In original Heroes 2 this sequence of events will happen on every re-load (gargoyles received morale effect, centaurs not). Game could be re-loaded million times - you will receive exactly the same sequence (if your moves were the same too).
 - On fheroes2 sequence is random. For example, on first re-load gargoyles will receive morale effect, on second re-load - centaurs will, on third - nobody (or both), etc..

It is very important to support original behavior.

### Save file

Experienced this issue in almost every battle.

### Additional info

fheroes 0.9.6, Win10, GOG dosbox

vincent-grosbois:
close #3996


src/fheroes2/battle/battle_arena.cpp
```diff
@@ -282,6 +284,24 @@ Battle::Arena::~Arena()
     arena = nullptr;
 }
 
+// compute a new seed from a list of actions, so random actions happen differently depending on user inputs
+size_t UpdateRandomSeed( const size_t seed, const Battle::Actions & actions )
```
oleg-derevenetz, line 288: I suppose it's better to create an anonymous namespace in the beginning of the source file and move it here (along with `GetCovr()`) may be.

src/engine/rand.h
```diff
@@ -96,6 +103,32 @@ namespace Rand
     private:
         int32_t Get( const std::function<uint32_t( uint32_t )> & randomFunc );
     };
+
+    // Specific random generator used so that battle randomness is deterministic
+    class BattleRandomGenerator
```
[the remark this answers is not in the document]
oleg-derevenetz: `Engine` contains some generic classes and functions, not bound to the game logic itself. Since this random generator may be used not only in a battle, may be rename it in a more generic way - `DeterministicRandomGenerator` may be?

oleg-derevenetz (changes requested):
Hi @vincent-grosbois I left a few comments here, would you mind to take a look?

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -386,6 +386,8 @@ namespace AI
         _enemySpellStrength = 0;
         _highestDamageExpected = 0;
         _considerRetreat = false;
+        _randomGenerator = &arena.GetRandomGenerator();
+        assert( _randomGenerator );
```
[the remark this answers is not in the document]
vincent-grosbois: This pointer will outlive the arena, but it can only ever be re-entered via this method, so it's going to be re-set to a correct value on the next battle. It's the same thing that is happening for _commander member, where it's also dangling.

I think the correct fix for this would be a refactor of this class (create a new instance for each battle) rather than fixing this specific member

oleg-derevenetz (approved):

src/fheroes2/battle/battle_arena.cpp
```diff
@@ -211,9 +234,9 @@ Battle::Arena::Arena( Army & a1, Army & a2, s32 index, bool local )
...
     if ( castle ) {
         // init
-        towers[0] = castle->isBuild( BUILD_LEFTTURRET ) ? new Tower( *castle, TWR_LEFT ) : nullptr;
-        towers[1] = new Tower( *castle, TWR_CENTER );
-        towers[2] = castle->isBuild( BUILD_RIGHTTURRET ) ? new Tower( *castle, TWR_RIGHT ) : nullptr;
+        towers[0] = castle->isBuild( BUILD_LEFTTURRET ) ? new Tower( *castle, TWR_LEFT, _randomGenerator ) : nullptr;
+        towers[1] = new Tower( *castle, TWR_CENTER, _randomGenerator );
+        towers[2] = castle->isBuild( BUILD_RIGHTTURRET ) ? new Tower( *castle, TWR_RIGHT, _randomGenerator ) : nullptr;
         const bool fortification = ( Race::KNGT == castle->GetRace() ) && castle->isBuild( BUILD_SPEC );
         catapult = army1->GetCommander() ? new Catapult( *army1->GetCommander() ) : nullptr;
```
oleg-derevenetz, line 241: @vincent-grosbois BTW catapult probably needs your random generator too.

vincent-grosbois:
Youre right!

Envoyé de mon iPhone

oleg-derevenetz:
@vincent-grosbois How about making this deterministic combat optional (I mean, instant combat should still be replayed as is, but allow each specific battle to take place in its own way)? In other words, an option to not derive initial seed from map cell index and world's map seed?

src/engine/rand.h
```diff
@@ -96,6 +103,32 @@ namespace Rand
...
+
+    // Specific random generator that keeps and update its state
+    class DeterministicRandomGenerator
+    {
+    public:
+        DeterministicRandomGenerator( const size_t initialSeed );
+
+        // prevent accidental copies
+        DeterministicRandomGenerator( const DeterministicRandomGenerator & ) = delete;
+        DeterministicRandomGenerator & operator=( const DeterministicRandomGenerator & ) = delete;
```
ihhub, line 115: Could we please add deleted destructor to please clang-tidy?

src/engine/rand.cpp
```diff
@@ -125,3 +125,23 @@ int32_t Rand::Queue::GetWithSeed( uint32_t seed )
...
+
+size_t Rand::DeterministicRandomGenerator::GetSeed() const
+{
+    return _currentSeed;
+}
+
+void Rand::DeterministicRandomGenerator::UpdateSeed( const size_t seed )
+{
+    _currentSeed = seed;
+}
```
ihhub, line 141: Absolutely optional: we could put the body of these methods inside header making them inline methods.

src/fheroes2/battle/battle_arena.cpp
```diff
@@ -164,14 +164,16 @@ Battle::Arena::Arena( Army & a1, Army & a2, s32 index, bool local )
...
 {
     const Settings & conf = Settings::Get();
     usage_spells.reserve( 20 );
 
     assert( arena == nullptr );
     arena = this;
-    army1 = new Force( a1, false );
-    army2 = new Force( a2, true );
+
+    army1 = new Force( a1, false, _randomGenerator );
```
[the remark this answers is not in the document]
ihhub: Hi @oleg-derevenetz , should we shut this warning too?

src/fheroes2/ai/normal/ai_normal_battle.cpp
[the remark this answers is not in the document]
ihhub: Can we please just add a TODO comment for the future refactoring but not to do this in the current pull request?

src/engine/rand.h
```diff
@@ -96,6 +103,32 @@ namespace Rand
...
+
+        size_t GetSeed() const;
+        void UpdateSeed( const size_t seed );
+
+        uint32_t Get( uint32_t from, uint32_t to = 0 ) const;
```
ihhub, line 120: These input arguments could be `const`.

src/fheroes2/battle/battle_army.h
```diff
@@ -25,6 +25,7 @@
 
 #include "army.h"
 #include "bitmodes.h"
+#include "rand.h"
```
ihhub, line 28: Here we can go with forward declaration while putting the header into the source file.

src/fheroes2/battle/battle_troop.h
```diff
@@ -31,6 +31,7 @@
 #include "battle_cell.h"
 #include "bitmodes.h"
 #include "players.h"
+#include "rand.h"
```
ihhub, line 34: Here we can go with forward declaration while putting the header into the source file.

ihhub (commented):
Hi @vincent-grosbois , good job! I left just few comments here. Would you mind to take a look?

src/engine/rand.h
oleg-derevenetz: Defaulted destructor you mean? :) I don't think that deleted destructor would be useful here.

ihhub: Oh, my bad. I have more sleep. You're absolutely right.

vincent-grosbois:
> @vincent-grosbois How about making this deterministic combat optional (I mean, instant combat should still be replayed as is, but allow each specific battle to take place in its own way)? In other words, an option to not derive initial seed from map cell index and world map seed?

why not but how do you want to do it? with a config option?

ihhub:
@vincent-grosbois , thank you so much for this change!

PR #4073, opened 2021-08-23
Fix the logic in Army::GetMoraleModificator()

qarkai:
Fix the undead morale penalty:
* all undead get no modificator
* artifact "Arm of the Martyr" discards "All race troops +1" bonus

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -997,29 +997,29 @@ int Army::GetMorale( void ) const
 
 int Army::GetMoraleModificator( std::string * strs ) const
 {
-    int result = Morale::NORMAL;
-
     // different race penalty
     std::set<int> races;
     bool hasUndead = false;
+    bool allUndead = true;
 
     for ( const Troop * troop : *this )
         if ( troop->isValid() ) {
             races.insert( troop->GetRace() );
             hasUndead = hasUndead || troop->isUndead();
+            allUndead = allUndead && troop->isUndead();
         }
 
+    if ( allUndead )
+        return Morale::NORMAL;
+
     const int count = static_cast<int>( races.size() );
-    const bool allUndead = hasUndead && AllTroopsAreUndead();
+    int result = Morale::NORMAL;
 
     switch ( count ) {
     case 0:
     case 2:
         break;
     case 1:
-        if ( allUndead )
-            return 0;
-
         if ( !AllTroopsAreTheSame() ) {
             ++result;
             if ( strs ) {
@@ -1044,7 +1044,7 @@ int Army::GetMoraleModificator( std::string * strs ) const
     }
 
     // undead in life group or artifact "Arm of the Martyr"
-    if ( !allUndead && ( hasUndead || ( GetCommander() && GetCommander()->HasArtifact( Artifact::ARM_MARTYR ) ) ) ) {
+    if ( hasUndead || ( GetCommander() && GetCommander()->HasArtifact( Artifact::ARM_MARTYR ) ) ) {
         result -= 1;
         if ( strs ) {
             strs->append( _( "Some undead in group -1" ) );
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -1024,7 +1024,7 @@ int Army::GetMoraleModificator( std::string * strs ) const
             ++result;
             if ( strs ) {
                 std::string str = _( "All %{race} troops +1" );
-                StringReplace( str, "%{race}", Race::String( *races.begin() ) );
+                StringReplace( str, "%{race}", *races.begin() == Race::NONE ? _( "Multiple" ) : Race::String( *races.begin() ) );
                 strs->append( str );
                 *strs += '\n';
             }
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -1012,15 +1012,18 @@ int Army::GetMoraleModificator( std::string * strs ) const
     if ( allUndead )
         return Morale::NORMAL;
 
-    const int count = static_cast<int>( races.size() );
     int result = Morale::NORMAL;
 
+    // artifact "Arm of the Martyr" adds the undead morale penalty
+    hasUndead = hasUndead || ( GetCommander() && GetCommander()->hasArtifact( Artifact::ARM_MARTYR ) );
+
+    const int count = static_cast<int>( races.size() );
     switch ( count ) {
     case 0:
     case 2:
         break;
     case 1:
-        if ( !hasUndead && !AllTroopsAreTheSame() ) { // Neutral troops army can have undead Ghost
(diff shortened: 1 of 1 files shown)


src/fheroes2/army/army.cpp
```diff
@@ -995,131 +995,56 @@ int Army::GetMorale( void ) const
...
         }
         break;
     }
 
-    // undead in life group
-    if ( ( hasDifferentTroops && ( count_necr || ghost_present ) && ( count_kngt || count_barb || count_sorc || count_wrlk || count_wzrd || count_bomg ) ) ||
-         // or artifact Arm Martyr
-         ( GetCommander() && GetCommander()->HasArtifact( Artifact::ARM_MARTYR ) ) ) {
+    // undead in life group or artifact "Arm of the Martyr"
+    if ( hasUndead || ( GetCommander() && GetCommander()->HasArtifact( Artifact::ARM_MARTYR ) ) ) {
```
vincent-grosbois, line 1047: you changed the logic here
Now if you have only undead you'll get -1 morale penalty

qarkai: Nope, method returns from `switch` if army has only undead. Conditions were redundant.

a1exsh:
@qarkai could you please add a brief description of the intent of this PR?  What was not optimal and how the new version is better?  Thanks!

oleg-derevenetz:
> @qarkai could you please add a brief description of the intent of this PR?  What was not optimal and how the new version is better?  Thanks!

The new version is at least shorter.

qarkai:
@a1exsh I don't understand. If you see something wrong say it directly. I believe that code is self-descriptive.

a1exsh:
@qarkai when I read "optimize" my first thought is performance optimization.  Not knowing the optimization target, it's hard to tell if the actual change pays into it.

Here, I understand, we are talking about making code shorter or simpler.  Maybe we should call it that way then? ;)  E.g. "Simplify XXX" or "Make XXX shorter".

I also believe that a good commit message should answer two questions: the what (in the summary) and the why (in the description).  This helps at the review time and especially when looking back to understand why something was changed in particular way.

That's just my opinion, of course. :grinning:

src/fheroes2/army/army.cpp
a1exsh: What if you have a `NECR` and a Ghost?  They count as different races, as far as I understand, so the count of races will be 2 in the switch above?

qarkai:
@a1exsh it doesn't matter how we call PRs, @ihhub changes them on merge as he wish.

src/fheroes2/army/army.cpp
qarkai: I don't have ghosts to check but you probably right. Original code has same issue.

a1exsh:
> @a1exsh it doesn't matter how we call PRs, @ihhub changes them on merge as he wish.

But PRs are reviewed by more than just one person.  It helps, well, me at least, to have a meaningful PR description. ;)

src/fheroes2/army/army.cpp
vincent-grosbois: this is potentially a bug though, I wonder if Ghost shouldn't be counted as Necro for this method

ihhub:
Hi @qarkai and @a1exsh , I usually rename commits before merging or pull requests to be understandable for the most of people, not because of my personal preference :) Also it's needed when I'm constructing the list of changes for a release since players don't need to know implementation details but what was fixed / added.

src/fheroes2/army/army.cpp
a1exsh: I also thought about it, but what if you have a pleasant, a skeleton and a ghost — that should still count as 3 different alignments, right?

a1exsh:
Looks like everything is covered now, but sadly there is now a merge conflict...

src/fheroes2/army/army.cpp
LeHerosInconnu: Hello @qarkai and @oleg-derevenetz,

> @LeHerosInconnu how about the fact that multiple "Multiple" (LOL) troops give +1 morale? :) I'm not saying that this behavior should be changed, but it's a bit illogical. What a Genie and a Thief (sorry, Rogue) have in common? :)

They are all "town orphans", that's what brings them closer and so they are happy to be together, except for the Ghost who scares them all except the Elementals. :D

> Also you didn't answer about Goblin situation. He removes +1 bonus for Sorceress army on your screenshot.

It is normal that the Goblin removes the "All Sorceress troops" bonus because he is not from the Sorceress castle, so all troops are not from the same castle anymore.

qarkai: > This is to show that in case the troops are all either from the same castle or all neutral troops, as soon as the Ghost considered as "Undead" is in the army, "All same troops" is not taken into account anymore.

"This is to show that in case the troops are all either from the same castle or all neutral troops, as soon as the Goblin considered as... "Goblin" is in the army, "All same troops" is not taken into account anymore."
I'm tired to show lack of logic in Ghost exception.

BTW how does "Arm of the Martyr" and "All <alignment> troops>" combination work in PoL?

LeHerosInconnu: Hello @qarkai,
 
> BTW how does "Arm of the Martyr" and "All troops>" combination work in OG PoL?

The Arm of the Martyr artifact gives the Undead Morale penalty.

In the original game:

Ghost Undead Neutral 02 001
Ghost Undead Neutral 02 002
Ghost Undead Neutral 05 001
Ghost Undead Neutral 05 002
Ghost Undead Neutral 05 003

qarkai: Finally unambiguous example of 'presence of undead discards +1 morale for "All <alignment> troops"' rule! Was it too hard? Thank you anyway.

LeHerosInconnu: @qarkai.

> Finally unambiguous example of 'presence of undead discards +1 morale for "All troops"' rule! Was it too hard? Thank you anyway.

Glad you finally understand how it works. :D

qarkai: @LeHerosInconnu as soon as you explained it without ambiguity.

LeHerosInconnu: @qarkai.

> @LeHerosInconnu as soon as you explained it without ambiguity.

It's okay, some people need more time. :)

a1exsh (approved):
The logic is not trivial, in the end.  Pity that we don't have unit tests: they could save us from fixing this another time. ;)

qarkai:
> The logic is not trivial, in the end. Pity that we don't have unit tests: they could save us from fixing this another time. ;)

More pity that this logic is not documented anywhere.

ihhub:
Hi @qarkai , are we good now to merge this change?

qarkai:
> Hi @qarkai , are we good now to merge this change?

Yes.

ihhub:

qarkai:
> @qarkai , thank you very much for this code cleanup!

FYI It's not just cleanup.

PR #4075, opened 2021-08-24
Modify Necromancer Captain Quarters look

### LINKED ISSUE #3841: Add missing Captain's Quarters part in Necromancer town as a separate image
If Necromancer town does not have built castle but contains Captain's Quarters it's visible that a part of the Quarters is missing. This is the issue in the original game assets. We would like to add missing portion of Quarters but creating an extra image which will be a part of fheroes2 assets. In other words we need to draw an image which will be attached to existing Quarters sprite making it as a completed building rather than truncated in the middle.

This is the task for artists but anyone is welcome to try. Once you're done with image creation either post it here or approach ihhub in [Discord](https://discord.com/invite/xF85vbZ) for assistance in converting image into readable by game format.

image

ihhub:
close #3841

This is how it looks (no bridge to nowhere):
image

### DIFF
--- a/src/fheroes2/castle/castle_building.cpp
+++ b/src/fheroes2/castle/castle_building.cpp
@@ -39,6 +39,87 @@ fheroes2::Rect CastleGetCoordBuilding( int, building_t, const fheroes2::Point & 
 void CastlePackOrdersBuildings( const Castle &, std::vector<building_t> & );
 fheroes2::Rect CastleGetMaxArea( const Castle &, const fheroes2::Point & );
 
+namespace
+{
+    bool isBuildingConnectionNeeded( const Castle & castle, const uint32_t buildId, const bool constructionInProgress )
+    {
+        const int race = castle.GetRace();
+
+        if ( race == Race::BARB ) {
+            if ( buildId & BUILD_MAGEGUILD ) {
+                const int mageGuildLevel = castle.GetLevelMageGuild();
+                if ( constructionInProgress ) {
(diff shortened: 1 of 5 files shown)


src/fheroes2/castle/castle_building.cpp
```diff
@@ -39,6 +39,87 @@ fheroes2::Rect CastleGetCoordBuilding( int, building_t, const fheroes2::Point &
...
+        const int race = castle.GetRace();
+
+        if ( race == Race::BARB ) {
+            if ( buildId & BUILD_MAGEGUILD ) {
+                const int mageGuildLevel = castle.GetLevelMageGuild();
+                if ( constructionInProgress ) {
+                    return mageGuildLevel == 0 || buildId > ( BUILD_MAGEGUILD1 << ( mageGuildLevel - 1 ) );
+                }
+                else {
+                    return buildId == ( BUILD_MAGEGUILD1 << ( mageGuildLevel - 1 ) );
```
oleg-derevenetz, line 55: I suppose it is not possible for `mageGuildLevel` to be less than 1 here?

oleg-derevenetz (commented):
Hi @ihhub I just left one question, please take a look when appropriate.

src/fheroes2/castle/castle_building.cpp
ihhub: If the construction isn't in progress it means that the building is built. To be safe I can add an assertion.

PR #4077, opened 2021-08-24
Remove unused LOSS_STARTHERO loss condition

oleg-derevenetz:
It is unused after merging of #3922


ihhub:
@oleg-derevenetz , thank you very much for cleanup!

Issue #4064, opened 2021-08-22
Combat during a siege, if the combat is restarted after the instant combat, the Captain has zero spell points


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, for a combat during a siege, if the combat is restarted after the instant combat, the Captain has zero spell points

In fheroes2:

![Bug restart siege spell points captain 001](https://user-images.githubusercontent.com/43583994/130361180-963f0805-d90e-4e2f-bc28-850347e89045.png)

![Bug restart siege spell points captain 002](https://user-images.githubusercontent.com/43583994/130361183-50e93b71-ade0-4df2-800d-739d54534d81.png)

End the turn.
Restart the siege combat (third combat).

### Save file

Here is the save file:
[Bug captain zero spell points_.zip](https://github.com/ihhub/fheroes2/files/7027653/Bug.captain.zero.spell.points_.zip)

### Additional info

Version 0.9.6 3976.

ihhub:
And that's why our AI is weak!

oleg-derevenetz:
> And that's why our AI is weak!

This issue related only to captains, heroes will currently get their mana back as it should.

PR #4078, opened 2021-08-24
Fix AI behavior during siege

oleg-derevenetz:
I've seen similar AI behavior during a siege many times:

Paladin in the moat

If I press the "AUTO" button now, Paladin does this:

Paladin fail

Utter failure. My initial (not a very thorough) investigation showed that this is not a pathfinding problem, but now I did a proper investigation and it showed that I was wrong - AI pathfinder really believes that Paladin is not able to reach Unicorn here(?!). So I replaced AI pathfinder by my pathfinder from `Battle::Board` (for the purpose of making decisions on the reachability of units for attack), and...

Paladin success

The pathetic unicorns were defeated, as they should have been.

Hi @ihhub regarding your sentence from #4064:

> And that's why our AI is weak!

**_That's why it is weak_** - at least, one of the real reasons. Pathfinding issues will ruin even the most sophisticated battle planner.

Also I fixed an issue when archers are trying to "kite" melee attackers by "moving" to the very same position that they currently occupy which is expressed in errors like this:

```
23:57:39: [DBG_ENGINE]	main:  Free Heroes of Might and Magic II, version: 0.9.6
23:57:57: [DBG_BATTLE]	GetPath:  Path is not found for Unit: [ 11 Druids, Red, pos: 54, -1, reflect ], destination: (head cell ID: 54, tail cell ID: -1)
23:57:57: [DBG_BATTLE]	ApplyActionMove:  path empty, Unit: [ 11 Druids, Red, pos: 54, -1, reflect ] to dst: 54
23:57:57: [DBG_BATTLE]	GetPath:  Path is not found for Unit: [ 11 Druids, Red, pos: 54, -1, reflect ], destination: (head cell ID: 54, tail cell ID: -1)
23:57:57: [DBG_BATTLE]	ApplyActionMove:  path empty, Unit: [ 11 Druids, Red, pos: 54, -1, reflect ] to dst: 54
```

P.S. However, I'm not sure about the latter issue with kiting - is it worth to fix it, may be just print warning as it happens now... The thing is, AI battle planner recommends the unit to kite to the definitely unreachable cell - unreachable to such an extent that nearest reachable cell turns out to be the current unit cell (I suppose that recommended cell is very close to the current one and reachable only in such a way that unit should leave the castle first and then return back to the castle via a moat, which is impossible during the same turn, that's why such an effect). In theory, this should not happen at all... On the other hand, of course, AI pathfinder is good at the strategic level (like an owl from a famous joke), but not at the tactical level, so it is natural that pathfinder from `Battle::Board` is used to correct it at the tactical level.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -88,7 +88,7 @@ namespace AI
             MeleeAttackOutcome current;
             current.positionValue = Board::GetCell( cell )->GetQuality();
             current.attackValue = Board::OptimalAttackValue( attacker, defender, cell );
-            current.canAttackImmediately = Board::CanAttackUnitFromCell( attacker, defender, cell );
+            current.canAttackImmediately = Board::CanAttackUnitFromPosition( attacker, defender, cell );
 
             // Pick target if either position has improved or unit is higher value at the same position quality
             if ( IsOutcomeImproved( current, bestOutcome ) ) {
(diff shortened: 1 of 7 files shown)


oleg-derevenetz:
I played a siege a few dozen times with these changes and I should say that castle siege (as well as castle defense) performed by AI looks much better now, no obviously stupid moves.

ihhub:

PR #4082, opened 2021-08-25
Rename HeroBase's HasArtifact to artifactCount for readability

qarkai:
And add separate `bool hasArtifact`.
And simplify some calculations related to `artifactCount`.


ihhub:

PR #4094, opened 2021-08-27
Simplify passability logic a bit, clarify function descriptions

oleg-derevenetz:
* Simplify logic a bit: cells occupied by the unit are always reachable for the same unit, if cell is occupied by another unit, it is unreachable anyway;
* Make it clear that passability-related functions will work only with "current unit" - the unit for which cell passability information is currently calculated.


ihhub:
@oleg-derevenetz , good job on making the code easier to read!

PR #4095, opened 2021-08-27
Heroes::inCastle(): respect GUARDIAN mode

oleg-derevenetz:
This fixes an ability to upgrade units in castle guardian armies (and, may be, something else).


ihhub:

PR #4104, opened 2021-08-28
Remove redundant member initializations

qarkai:
Fix clang-tidy [readability-redundant-member-init](https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-member-init.html) warnings.
Disable check for unnecessary base class initializations within copy constructors since gcc generates warning if it removed.


ihhub:
@qarkai , thank you for this code correction!

PR #4106, opened 2021-08-28
Display message about Ultimate Artifact after the battle

### LINKED ISSUE #1511: After the combat, message for capturing an Ultimate Artifact
In the original game, a message is displayed when a hero captures an opponent's Ultimate Artifact to inform the player that the hero has captured an Ultimate Artifact but the captured artifact suddenly disappears.
The hero therefore does not keep this artifact.

In fheroes2, Ultimate Artifacts are not kept either, but no message is displayed.

In original game:

Capturing Ultimate Artifact 01

Capturing Ultimate Artifact 02

Capturing Ultimate Artifact 03

In fheroes2:

Capturing Ultimate Artifact 04

In the original game, this message is displayed right after the combat on the adventure map.
I suggest that for fheroes2 this message be displayed directly in the combat screen following the other artifact capture messages and in the same window style.

So following these windows, and in the same window style:

Capturing Ultimate Artifact 01

There is also a bug in the original game.
When it is the opponent who captures an Ultimate Artifact from the player controlled hero, the same message is displayed.
While this message should be tailored to the situation.

In original game, the same message is displayed not tailored to the situation:

Capturing Ultimate Artifact 06

I suggest when this happens, to display this message in the combat screen following the defeat message and in the same window style: "As your enemy reach for the Ultimate Cloak of Protection, it mysteriously disappears."

So following this window, and in the same window style:

Capturing Ultimate Artifact 05

I have planned an artifact that allows you to keep Ultimate Artifacts after a combat.
And it could also be added as a game option: "game: Ultimate Artifacts captured from the opponent are kept by the hero". :)
I will open new issues for these proposals later.

a1exsh:
Fix #1511

The capture of human player's ultimate artifact is harder to handle, because it happens in a different place in the code: https://github.com/ihhub/fheroes2/blob/master/src/fheroes2/battle/battle_main.cpp#L238-L257

This logic is duplicated in the dialog code that this PR changes, but there it is interleaved with display code...

### DIFF
--- a/src/fheroes2/battle/battle.h
+++ b/src/fheroes2/battle/battle.h
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "army.h"
+#include "artifact.h"
 #include "gamedefs.h"
 #include "m82.h"
 
@@ -67,6 +68,8 @@ namespace Battle
     };
 
     Result Loader( Army &, Army &, s32 );
+
+    typedef std::list<std::pair<BagArtifacts::iterator, BagArtifacts::iterator>> ArtifactsPickup;
 
     struct TargetInfo
     {
--- a/src/fheroes2/battle/battle.h
+++ b/src/fheroes2/battle/battle.h
@@ -27,7 +27,6 @@
 #include <vector>
 
 #include "army.h"
-#include "artifact.h"
 #include "gamedefs.h"
 #include "m82.h"
 
--- a/src/fheroes2/battle/battle.h
+++ b/src/fheroes2/battle/battle.h
@@ -69,7 +69,7 @@ namespace Battle
 
     Result Loader( Army &, Army &, s32 );
 
-    typedef std::list<std::pair<BagArtifacts::iterator, BagArtifacts::iterator>> ArtifactsPickup;
+    using ArtifactsPickup = std::list<std::pair<BagArtifacts::iterator, BagArtifacts::iterator>>;
 
     struct TargetInfo
     {
--- a/src/fheroes2/battle/battle.h
+++ b/src/fheroes2/battle/battle.h
@@ -69,7 +69,8 @@ namespace Battle
 
     Result Loader( Army &, Army &, s32 );
 
-    using ArtifactsPickup = std::list<std::pair<BagArtifacts::iterator, BagArtifacts::iterator>>;
+    using SingleArtifactPickup = std::pair<BagArtifacts::iterator, BagArtifacts::iterator>;
+    using ArtifactsPickup = std::list<SingleArtifactPickup>;
 
     struct TargetInfo
     {
--- a/src/fheroes2/battle/battle.h
+++ b/src/fheroes2/battle/battle.h
@@ -69,11 +69,6 @@ namespace Battle
 
     Result Loader( Army &, Army &, s32 );
 
-    // a single artifact pickup action is a pair of pointers (dst, src) into the artifact bags of the two heroes
-    // dst may be null, if the artifact cannot be picked up
-    using SingleArtifactPickup = std::pair<Artifact *, Artifact *>;
-    using ArtifactsPickup = std::list<SingleArtifactPickup>;
-
     struct TargetInfo
     {
         Unit * defender;
--- a/src/fheroes2/battle/battle.h
+++ b/src/fheroes2/battle/battle.h
@@ -69,7 +69,9 @@ namespace Battle
 
     Result Loader( Army &, Army &, s32 );
 
-    using SingleArtifactPickup = std::pair<BagArtifacts::iterator, BagArtifacts::iterator>;
+    // a single artifact pickup action is a pair of pointers (dst, src) into the artifact bags of the two heroes
+    // dst may be null, if the artifact cannot be picked up
+    using SingleArtifactPickup = std::pair<Artifact *, Artifact *>;
     using ArtifactsPickup = std::list<SingleArtifactPickup>;
 
     struct TargetInfo
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -484,15 +484,15 @@ bool Battle::Arena::DialogBattleSummary( const Result & res, const ArtifactsPick
     }
 
     if ( !artifacts.empty() ) {
-        HeroBase * hero_wins = ( res.army1 & RESULT_WINS ? army1->GetCommander() : ( res.army2 & RESULT_WINS ? army2->GetCommander() : nullptr ) );
-        HeroBase * hero_loss = ( res.army1 & RESULT_LOSS ? army1->GetCommander() : ( res.army2 & RESULT_LOSS ? army2->GetCommander() : nullptr ) );
+        HeroBase * winner = ( res.army1 & RESULT_WINS ? army1->GetCommander() : ( res.army2 & RESULT_WINS ? army2->GetCommander() : nullptr ) );
+        HeroBase * loser = ( res.army1 & RESULT_LOSS ? army1->GetCommander() : ( res.army2 & RESULT_LOSS ? army2->GetCommander() : nullptr ) );
 
         // Can't transfer artifacts
-        if ( hero_wins == nullptr || hero_loss == nullptr )
+        if ( winner == nullptr || loser == nullptr )
             return false;
 
-        const bool isWinnerHuman = hero_wins && hero_wins->isControlHuman();
-        const BagArtifacts::const_iterator winnerBagEnd = hero_wins->GetBagArtifacts().end();
+        const bool isWinnerHuman = winner && winner->isControlHuman();
+        const BagArtifacts::const_iterator winnerBagEnd = winner->GetBagArtifacts().end();
 
         btn_ok.setICNInfo( isEvilInterface ? ICN::WINCMBBE : ICN::WINCMBTB, 0, 1 );
         btn_ok.setPosition( pos_rt.x + 121, pos_rt.y + 410 );
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -492,7 +492,6 @@ bool Battle::Arena::DialogBattleSummary( const Result & res, const ArtifactsPick
             return false;
 
         const bool isWinnerHuman = winner && winner->isControlHuman();
-        const BagArtifacts::const_iterator winnerBagEnd = winner->GetBagArtifacts().end();
 
         btn_ok.setICNInfo( isEvilInterface ? ICN::WINCMBBE : ICN::WINCMBTB, 0, 1 );
         btn_ok.setPosition( pos_rt.x + 121, pos_rt.y + 410 );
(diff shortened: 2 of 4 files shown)


src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -497,31 +497,39 @@ bool Battle::Arena::DialogBattleSummary( const Result & res, const bool transfer
         btn_ok.setICNInfo( isEvilInterface ? ICN::WINCMBBE : ICN::WINCMBTB, 0, 1 );
         btn_ok.setPosition( pos_rt.x + 121, pos_rt.y + 410 );
 
-        for ( size_t i = 0; i < bag2.size(); ++i ) {
-            Artifact & art = bag2[i];
+        BagArtifacts & sortedBag2( bag2 );
+        // all ultimate artifacst should be displayed last
+        std::sort( sortedBag2.begin(), sortedBag2.end(),
```
a1exsh, line 502: We need to at least use `std::stable_sort` here, or better don't have the logic in two places.  Will check what can be done about that.

a1exsh:
Added handling of computer trying to capture an ultimate artifact:
image

In the end, the logic is still in two places, but the dialog treats the artifact bags as "read-only".  This could be further extracted into a single place, but requires more effort than I can spend right now.

a1exsh:
> This could be further extracted into a single place, but requires more effort than I can spend right now.

I have an idea.  We could separate the logic into two functions:
1. Calculate anticipated artifact transfer after the battle, without actually modifying anything.
2. If a human player was involved in the battle, the battle summary dialog is shown where the artifact transfer is only displayed, but not yet executed.
3. Ultimately (he-he), the transfer is executed, regardless if a human player is involved, based on the data calculated in the first step.

The requried data structure can be as simple as a list of `(src, dst)` iterator pairs.

What do you think?

a1exsh:
Implementation looks like this: https://github.com/ihhub/fheroes2/pull/4106/commits/46eb6890333d05c1def9139450170fbdfd35d87a

I'm also noticing an opportunity to make the code in `Battle::Loader()` significantly shorter by using a `do { ... } while ()`.  Should I try to do that in this PR or better a separate one? :)

src/fheroes2/battle/battle_main.cpp
```diff
@@ -244,24 +238,50 @@ Battle::Result Battle::Loader( Army & army1, Army & army2, s32 mapsindex )
...
+}
+
+void Battle::ExecArtifactsPickup( const ArtifactsPickup & action )
+{
+    for ( ArtifactsPickup::const_iterator it = action.begin(); it != action.end(); ++it ) {
+        BagArtifacts::iterator wi = it->first;
+        BagArtifacts::iterator li = it->second;
+        Artifact & art = *li;
+
+        // FIXME: abstraction leak
```
a1exsh, line 280: Not sure what to do about this one.  We need to signal "no slot" on the left (winner) side somehow, but that's not possible with generic iterators, I guess.

src/fheroes2/battle/battle_main.cpp
```diff
@@ -41,7 +41,8 @@
 
 namespace Battle
 {
-    void PickupArtifactsAction( HeroBase &, HeroBase & );
+    ArtifactsPickup CalcArtifactsPickup( HeroBase & hero_wins, HeroBase & hero_loss );
+    void ExecArtifactsPickup( const ArtifactsPickup & actions, BagArtifacts::const_iterator wend );
```
ihhub, line 45: Let please put these functions in anonymous namespace within this file to reduce its scope.

src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -483,45 +483,46 @@ bool Battle::Arena::DialogBattleSummary( const Result & res, const bool transfer
...
+                }
+                else {
+                    if ( isWinnerHuman ) {
+                        artMsg = _( "As you reach for the %{name}, it mysteriously disappears." );
+                    }
+                    else {
+                        artMsg = _( "As your enemy reaches for the %{name}, it mysteriously disappears." );
+                    }
+                    StringReplace( artMsg, "%{name}", art.GetName() );
+                }
```
ihhub, line 523: Should we do the opposite logic to avoid NOT operation?
```
if ( art.isUltimate() ) {
    // do something
}
else {
    // do anything else
}
```

It's not about optimization but a reader can easily doesn't notice "!" while reading the code.

src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -483,45 +483,46 @@ bool Battle::Arena::DialogBattleSummary( const Result & res, const bool transfer
...
-        BagArtifacts & bag1 = hero1->GetBagArtifacts();
-        BagArtifacts & bag2 = hero2->GetBagArtifacts();
+        bool isWinnerHuman = hero_wins && hero_wins->isControlHuman();
 
         btn_ok.setICNInfo( isEvilInterface ? ICN::WINCMBBE : ICN::WINCMBTB, 0, 1 );
         btn_ok.setPosition( pos_rt.x + 121, pos_rt.y + 410 );
 
-        for ( size_t i = 0; i < bag2.size(); ++i ) {
-            Artifact & art = bag2[i];
+        for ( ArtifactsPickup::const_iterator it = artifacts.begin(); it != artifacts.end(); ++it ) {
```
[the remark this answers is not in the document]
ihhub: Let's do range loop here but without  **auto**:
```
for ( const ArtifactsPickup & artifactPickup : artifacts )
```

src/fheroes2/battle/battle_main.cpp
```diff
@@ -244,24 +238,52 @@ Battle::Result Battle::Loader( Army & army1, Army & army2, s32 mapsindex )
...
 
-void Battle::PickupArtifactsAction( HeroBase & hero1, HeroBase & hero2 )
+Battle::ArtifactsPickup Battle::CalcArtifactsPickup( HeroBase & hero_wins, HeroBase & hero_loss )
 {
-    BagArtifacts & bag1 = hero1.GetBagArtifacts();
-    BagArtifacts & bag2 = hero2.GetBagArtifacts();
+    ArtifactsPickup result;
+
+    BagArtifacts & bag_wins = hero_wins.GetBagArtifacts();
+    BagArtifacts & bag_loss = hero_loss.GetBagArtifacts();
```
ihhub, line 246: Please use camelCase variable naming. For example `winnerBag` and `loserBag`.

src/fheroes2/battle/battle.h
```diff
@@ -68,6 +69,8 @@ namespace Battle
 
     Result Loader( Army &, Army &, s32 );
 
+    using ArtifactsPickup = std::list<std::pair<BagArtifacts::iterator, BagArtifacts::iterator>>;
```
ihhub, line 72: Hmm... I think the solution using iterators is not the best. `Artifact` is a very lightweight class so we can just store instead or even pointers to Artifact class as **BagArtifacts** is just a storage for them. What do you think?

ihhub (changes requested):
Hi @a1exsh , I left few comments here but before addressing them could you please explain the logic behind the code as reading iterator based code is a bit troublesome?

a1exsh:
@ihhub thanks for having a look!

The goal of is to extract the logic for artifacts pickup after the battle, so that it's not duplicated in the dialog and in the code that runs e.g. when both heroes involved in the battle are played by the AI.  My idea was to separate logic into two steps:
1. Prepare the plan for artifact pickup.
2. Execute the actual transfer, according to the plan.

Both steps are always taken, regardless if human player is involved, which requires less checks in the code.  In case of a human player, the plan is used to show captured/lost artifacts in the battle summary dialog, but actual transfer happens at a later step.

Most of the logic is in the function for preparing the plan, and execution is simply following it.  The plan (`ArtifactsPickup`) consists of pairs of references to places in the corresponding heroes artifact bags.  The task of executing the transfer is just going through the list of pairs, copying artifact values from one place to the other (unless there is no place to copy to, which is the case with ultimate artifacts or artifacts for which there was no free space in the winning hero's bag), finally removing the artifact from the losing hero bag (this is done always).

I've chosen iterators, but we could also use pointers.  Use of pointers has the advantage that we don't need additional parameter for sentinel value, we can just use a null pointer for that.

a1exsh:
@ihhub here's how it looks with the pointers instead of iterators: https://github.com/ihhub/fheroes2/pull/4106/commits/31ea0b7eedfd7be1538e8c273fe705bfdc2ba75c

ihhub:

src/fheroes2/battle/battle_main.cpp
```diff
@@ -41,11 +41,53 @@
...
 namespace Battle
 {
-    void PickupArtifactsAction( HeroBase &, HeroBase & );
     void EagleEyeSkillAction( HeroBase &, const SpellStorage &, bool, const Rand::DeterministicRandomGenerator & randomGenerator );
     void NecromancySkillAction( HeroBase & hero, const uint32_t, const bool isControlHuman, const Battle::Arena & arena );
 }
 
+namespace
+{
+    Battle::ArtifactsPickup PlanArtifactsPickup( BagArtifacts & winnerBag, BagArtifacts & loserBag )
```
ihhub, line 50: What if we simplify this code without using pointers at all? Artifact is lightweight class so it won't take much more space than a normal pointer.
```cpp
std::vector<Artifact> planArtifactTransfer( const BagArtifacts & winnerBag, const BagArtifacts & loserBag )
{
    std::vector<Artifact> artifacts;

    // Calculate how many places the winner has in his bag.
    size_t availableArtifactSlots = 0;
    for ( const Artifact & artifact : winnerBag ) {
        if ( !artifact.isValid() ) {
            ++availableArtifactSlots;
        }
    }

    for ( const Artifact & artifact : loserBag ) {
        if ( artifact.isValid() && art.GetID() != Artifact::MAGIC_BOOK ) {
            if ( art.isUltimate() ) {
                artifacts.emplace_back( artifact );
            }
            else if ( availableArtifactSlots > 0 ) {
                artifacts.emplace_back( artifact );
                --availableArtifactSlots;
            }
        }
    }

    return artifacts;
}
```
then inside `transferArtifacts` function which is `ExecArtifactsPickup` right now will look like:
```cpp
void transferArtifacts( BagArtifacts & winnerBag, BagArtifacts & loserBag, const std::vector<Artifact> & artifacts )
{
    // Clear loser's artifact bag.
    for ( Artifact & artifact : loserBag ) {
        if ( artifact.isValid() && art.GetID() != Artifact::MAGIC_BOOK ) {
            artifact = Artifact::UNKNOWN;
        }
    }

    size_t artifactPos = 0;

    for ( Artifact & artifact : winnerBag ) {
        if ( artifact.isValid() ) {
            continue;
        }

        for ( ; artifactPos >= artifacts.size(); ++artifactPos ) {
            if ( !artifacts[artifactPos].isUltimate() ) {
                artifact = artifacts[artifactPos];
                ++artifactPos;
                break;
            }
        }

        if ( artifactPos >= artifacts.size() ) {
            break;
        }
    }
}
```

In `Arena::DialogBattleSummary` method we pass `const std::vector<Artifact> & artifacts` which will combine both ultimate and normal artifacts. The condition
```cpp
if ( ( isWinnerHuman && nullptr != pickup.first ) || art.isUltimate() ) { // always show the message for ultimate artifacts
```
will simply be changed into
```cpp
if ( isWinnerHuman || art.isUltimate() ) { // always show the message for ultimate artifacts
```

We loose a bit here in performance (1 extra loop) but it should be fine as this code is executed very rarely. However, we'd gain in code readability. What do you think?

ihhub (changes requested):

src/fheroes2/battle/battle_main.cpp
a1exsh: Yeah, we can also get along with copying the artifacts.  The ultimate artifacts should still be moved to the end of the list, though, that makes the display more logical, IMO.

I will apply your suggestion and test how it works, thanks!

a1exsh: OK, seems like it works like expected.

Test map: [ARTICAPT.zip](https://github.com/ihhub/fheroes2/files/7147814/ARTICAPT.zip)
1. Play Blue to defeat Green's hero that has the Ultimate Book of Knowledge.
2. Play Green to see Ultimate Book disappear as Blue (computer) tries to get it.

ihhub:

PR #3724, opened 2021-06-19
Maps::FileInfo: don't save the full path to the map file, save only the basename

oleg-derevenetz:
fix #3722

May somewhat conflict with #3710 because of removed `Settings::GetListFiles()`, so one of them should be updated after merging of another one.

PR #4110, opened 2021-08-29
Add short save file description

ihhub:
useful to find save file location.

This is how it looks:
image

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -67,6 +67,15 @@ namespace
 
         return text.size();
     }
+
+    std::string ResizeToShortName( const std::string & str )
+    {
+        std::string res = System::GetBasename( str );
+        size_t it = res.rfind( '.' );
+        if ( std::string::npos != it )
+            res.resize( it );
+        return res;
+    }
 }
 
 std::string SelectFileListSimple( const std::string &, const std::string &, const bool );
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @Branikolog  and @LeHerosInconnu, could you please check this new small feature once build is completed?

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -83,14 +92,25 @@ class FileInfoListBox : public Interface::ListBox<Maps::FileInfo>
...
     void ActionListDoubleClick( Maps::FileInfo & ) override;
     void ActionListSingleClick( Maps::FileInfo & ) override;
 
-    void ActionListPressRight( Maps::FileInfo & ) override
+    void ActionListPressRight( Maps::FileInfo & info ) override
     {
-        // Do nothing.
+        // On some OS like Windows path contains '\' symbol. This symbol doesn't exist in the resources.
+        // To avoid this we have to replace all '\' symbols by '/' symbols.
+        std::string fullPath = info.file;
```
oleg-derevenetz, line 104: Hi @ihhub You won't get the full path here on relatively new saves (after merging of #3724). Settings will contain the full path to the current map after `Game::Load()` call, but not the `FileInfo` itself.

oleg-derevenetz: Ah, sorry, this is not the case. In the `Game::LoadSAV2FileInfo()` the `file` field of the `FileInfo` is forcibly replaced by the full path to the save file:

https://github.com/ihhub/fheroes2/blob/060c31eca0867e26a861c7e0ba355aadf86bcd13/src/fheroes2/game/game_io.cpp#L313

So everything should work here.

oleg-derevenetz (approved):

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -83,14 +92,25 @@ class FileInfoListBox : public Interface::ListBox<Maps::FileInfo>
...
+        StringReplace( fullPath, "\\", "/" );
+
+        std::string description( _( "Map: " ) );
+        description += info.name;
+        description += "\n \n";
```
a1exsh, line 109: Why not `\n\n` (without space)?  Is the space important for rendering somehow?

ihhub: This is old code / class usage and it can't handle `\n\n` properly but since we have concerns I'll use the newest classes for this.

LeHerosInconnu:
Hello @ihhub,

> Hi @Branikolog and @LeHerosInconnu, could you please check this new small feature once build is completed?

I tested your feature, it works fine for me.

Branikolog:
> Hi @Branikolog and @LeHerosInconnu, could you please check this new small feature once build is completed?

Hi, @ihhub!
Looks awesome! Just one question, which I can't examine on my machine: What if save file location wouldn't fit three lines. Will info window expand then?

ihhub:
> 
> 
> > Hi @Branikolog and @LeHerosInconnu, could you please check this new small feature once build is completed?
> 
> Hi, @ihhub!
> Looks awesome! Just one question, which I can't examine on my machine: What if save file location wouldn't fit three lines. Will info window expand then?

Window is not expandable in width for now. The line will be divided in the middle of words.

Issue #3721, opened 2021-06-19
AI: add hero's roles


ihhub:
Currently all AI heroes have the same role which is incorrect. Based on @idshibanov proposal we could make an enhancement on existing AI hero logic:

- [x] use existing `enum class Role` for roles. Enumeration entries could be renamed to be more readable
- [x] for now we will focus only on 3 roles: Scouting/gatherer, Hunter and Champion
- [x] designating hero a **champion** based on raw primary and secondary skills (XP don't really matter). Heroes get a bonus to give a champion army/artifacts. Skew his task estimates/priorities on visiting castles and fighting. Champion should be greatly overpower other heroes by skills
- [x] **Hunters** as a regular heroes as they are right now. Can fight, can just harvest resources, most of them are this role.
- [x] Designated scout/resource **gatherer** role if a lot of heroes. Leave them a stack or two of fastest troops. Have boost on fog of war scouting. Once nothing to scout (no fog tiles that can path to) convert to a hunter/logistics

All roles should be reevaluated at each AI turn. The main logic should be done for `ObjectValueStorage` class which should have hero's type and in `bool HeroesValidObject( const Heroes & hero, s32 index )` function.

For setting roles we need to use `void Normal::KingdomTurn( Kingdom & kingdom )` method just before **HeroesTurn( heroes );** call.

ihhub:
Hi @zenseii , you're right. I'm closing it.

PR #4112, opened 2021-08-30
Add base roles for AI heroes

ihhub:
relates to #3721


oleg-derevenetz (approved):

idshibanov (approved):
Overall a great start. You might consider writing a generic `getObjectValue` method and refactor getHunterObjectValue/getFighterObjectValue calls to specify special rules only and call a generic one at the end if no condition was met.

ihhub:
> 
> 
> Overall a great start. You might consider writing a generic `getObjectValue` method and refactor getHunterObjectValue/getFighterObjectValue calls to specify special rules only and call a generic one at the end if no condition was met.

Hi  @idshibanov , I made a TODO comment about this in the code.

PR #4113, opened 2021-08-30
Cleanup GetKingdomStartingResources

qarkai:
* Remove unused `kingdom_starting_resource` array.
* Remove intermediate redundant `UpdateStartingResource` method.
* Make `GetKingdomStartingResources` private and format according to codestyle.

### DIFF
--- a/src/fheroes2/kingdom/kingdom.cpp
+++ b/src/fheroes2/kingdom/kingdom.cpp
@@ -69,7 +69,7 @@ void Kingdom::Init( int clr )
     if ( Color::ALL & color ) {
         heroes.reserve( GetMaxHeroes() );
         castles.reserve( 15 );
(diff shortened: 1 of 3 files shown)


src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -46,6 +46,11 @@
 
 #include <cassert>
 
+namespace GameStatic
+{
+    extern cost_t kingdom_starting_resource[];
```
ihhub, line 51: This is the only place which bothers me. `extern` is very tricky which comes from C world. It's unknown for code reader where the location of this variable is. We could just add a function which will return a pointer to this array (which by right must be const after our recent changes).

What do you think @qarkai ?

qarkai: Probably simply move this array to `kingdom.cpp` after #4179.

PR #4118, opened 2021-09-01
Optimize Battle Arena related code

qarkai:
Similar to #4012:
* remove `invert` flag from `GetForce()`
* add `getEnemyForce()` as replacement
* format `GetForce()` according to code style

### DIFF
--- a/src/fheroes2/battle/battle_arena.h
+++ b/src/fheroes2/battle/battle_arena.h
@@ -74,7 +74,8 @@ namespace Battle
 
         Force & GetForce1( void );
         Force & GetForce2( void );
-        Force & GetForce( int color, bool invert = false );
+        Force & getForce( const int color );
+        Force & getEnemyForce( const int color );
         Force & GetCurrentForce( void );
 
         int GetArmyColor1( void ) const;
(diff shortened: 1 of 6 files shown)


src/fheroes2/battle/battle_arena.h
```diff
@@ -74,7 +74,8 @@ namespace Battle
 
         Force & GetForce1( void );
         Force & GetForce2( void );
-        Force & GetForce( int color, bool invert = false );
+        Force & getForce( const int color );
+        Force & getEnemyForce( const int color );
         Force & GetCurrentForce( void );
```
a1exsh, line 79: The change is good in general, but the renamed methods stand out now as they start with the small letter...

a1exsh: OK, but maybe it makes sense to rename the related GetForce methods as well, while changing these ones? Otherwise it introduces inconsistency.

ihhub: It is good from one side but from another keeping change minimal leads to easier code reviews :)

ihhub:
@qarkai , thanks a lot for this code change!

PR #4125, opened 2021-09-01
Return only unique maps (based on map file name) from PrepareMapsFileInfoList()

### LINKED ISSUE #4117: Scenario selection window, the name of each scenario is displayed twice
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the name of each scenario is displayed twice.

In fheroes2 :

Twice the same scenario name 01

### Save file

None.

### Additional info

Version 0.9.6 4111

oleg-derevenetz:
fix #4117

The root cause of this issue is that the list of maps can be read from the `MAPS` subdirectory of current directory, and if we run the `fheroes2.exe` from the directory that contains the `MAPS` subdirectory, then the files in this `MAPS` subdirectory are processed twice - the first time as maps from the current directory, and the second time as maps from the app installation directory. I really don't think that we should process any files from the current directory in any case, so this PR contains an optional commit that removes the processing of files from the current directory (I can revert it if there are some real reasons to do this).

Also move local functions in game_scenarioinfo.cpp to anonymous namespace.

P.S. Hi @ihhub you may want to sync [fheroes2/clang-tidy-pr-comments](https://github.com/fheroes2/clang-tidy-pr-comments) with the upstream once again - there are (minor) fresh changes were merged recently.

### DIFF
--- a/src/fheroes2/game/game_scenarioinfo.cpp
+++ b/src/fheroes2/game/game_scenarioinfo.cpp
@@ -144,7 +144,7 @@ namespace
     }
 }
 
-fheroes2::GameMode Game::SelectScenario( void )
+fheroes2::GameMode Game::SelectScenario()
 {
     return fheroes2::GameMode::SCENARIO_INFO;
 }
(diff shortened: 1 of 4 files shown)


ihhub:
@oleg-derevenetz , thanks a lot for this fix! **fheroes2/clang-tidy-pr-comments** was updated as well.

Issue #1890, opened 2020-10-03
Scenario Info window, the window is not centered vertically and horizontally in the screen


LeHerosInconnu:
In fheroes2, the window is not centered vertically and horizontally in the screen.

In fheroes2:

![Scenario info Evil fh2 01](https://user-images.githubusercontent.com/43583994/95000049-b941ff00-05bd-11eb-8a43-172c0b7cd2ae.png)

PR #4137, opened 2021-09-02
Alignment fix of scenario info window

obrem:
Fix for issue #1890 
The horizontal window alignment didn't take the border width into consideration when calculating its position.
The vertical position was misaligned due to sprite reporting a height of 476px when the actual size is 472px

box-height
actual-box-height

### DIFF
--- a/src/fheroes2/dialog/dialog_gameinfo.cpp
+++ b/src/fheroes2/dialog/dialog_gameinfo.cpp
@@ -45,7 +45,8 @@ void Dialog::GameInfo( void )
 
     const fheroes2::Sprite & box = fheroes2::AGG::GetICN( ICN::SCENIBKG, 0 );
 
-    fheroes2::Point pt( ( display.width() - box.width() ) / 2, ( display.height() - box.height() ) / 2 );
+    // the visual graphics of the box sprite height is 4 pixels less than given value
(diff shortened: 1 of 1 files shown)


a1exsh:
I guess the height simply includes the shadow that is cast below the box?..

obrem:
> I guess the height simply includes the shadow that is cast below the box?..

That makes sense, I doubled checked and it also happens to be so that the left shadow is exactly 16px wide, the same as the left border. I was making the assumption it was the border that was making the offset. But visually the window is centered.

Maybe the `-BORDERWIDTH` should be replaced with a value instead since its not correct?

a1exsh:
@obrem it is likely that the shadow is not part of the original image resource and is added by our code (look for `addShadow()`).  If that's the case, we should know how much X and Y offset for the shadow was used, so we can take that into account when aligning the dialog window.

ihhub:
Hi @obrem and @a1exsh , regarding the shadow and `ICN::SCENIBKG` image. The shadow is embedded into the image itself. It's not generated by our application. You can either use out tools to extract resources or call `Save` function within agg_image.cpp code to save your needed image.
Regarding the image itself: X offset is 16 pixels which is equivalent of **BORDERWIDTH** and Y offset is 4 pixels so @obrem is absolutely correct about the changes.

ihhub:
Hi @obrem , I slightly corrected the code so everyone else will understand why we use these magic numbers :)

ihhub:
@obrem , thank you so much for your contribution to the project!

PR #4140, opened 2021-09-03
Remove flags from most of the captured objects when losing

### LINKED ISSUE #1640: Adventure map, resource generators, there should not be a gray flag on the resource generators when a player loses the game
In fheroes2, when a player loses the game, the resource generators that were under his control have the flag color changed to gray.
In the original game the flags are removed, there is no gray flag.

In original game:

Resource generator flag H2 01

Resource generator flag H2 02

Resource generator flag H2 03

In fheroes2:

Resource generator flag fh2 01

Resource generator flag fh2 02

Resource generator flag fh2 03

a1exsh:
Fixes #1640

The original code needs rework, IMO, as it's abstraction leaks all the way down, but for now just fix the reported issue.

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1756,20 +1756,20 @@ void Maps::Tiles::CorrectFlags32( const int col, const u32 index, const bool up 
 {
     if ( col == Color::NONE ) {
         removeFlags();
+        return;
     }
-    else {
-        TilesAddon * taddon = FindFlags();
 
-        // replace flag
-        if ( taddon )
-            taddon->index = index;
-        else if ( up )
-            // or new flag
-            addons_level2.emplace_back( TilesAddon::UPPER, World::GetUniq(), 0x38, index );
-        else
-            // or new flag
-            addons_level1.emplace_back( TilesAddon::UPPER, World::GetUniq(), 0x38, index );
-    }
+    TilesAddon * taddon = FindFlags();
+
+    // replace flag
+    if ( taddon )
+        taddon->index = index;
+    else if ( up )
+        // or new flag
+        addons_level2.emplace_back( TilesAddon::UPPER, World::GetUniq(), 0x38, index );
+    else
+        // or new flag
+        addons_level1.emplace_back( TilesAddon::UPPER, World::GetUniq(), 0x38, index );
 }
 
 void Maps::Tiles::FixedPreload( Tiles & tile )
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1752,7 +1752,7 @@ void Maps::Tiles::CaptureFlags32( const MP2::MapObjectType objectType, int col )
     }
 }
 
-void Maps::Tiles::CorrectFlags32( int col, u32 index, bool up )
+void Maps::Tiles::CorrectFlags32( const int col, const u32 index, const bool up )
 {
     if ( col == Color::NONE ) {
         removeFlags();
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1694,31 +1694,31 @@ void Maps::Tiles::CaptureFlags32( const MP2::MapObjectType objectType, int col )
     switch ( objectType ) {
     case MP2::OBJ_WINDMILL:
         index += 42;
-        CorrectFlags32( index, false );
+        CorrectFlags32( col, index, false );
         break;
     case MP2::OBJ_WATERWHEEL:
         index += 14;
-        CorrectFlags32( index, false );
+        CorrectFlags32( col, index, false );
         break;
     case MP2::OBJ_MAGICGARDEN:
         index += 42;
-        CorrectFlags32( index, false );
+        CorrectFlags32( col, index, false );
         break;
 
     case MP2::OBJ_MINES:
         index += 14;
-        CorrectFlags32( index, true );
+        CorrectFlags32( col, index, true );
         break;
     case MP2::OBJ_LIGHTHOUSE:
         index += 42;
-        CorrectFlags32( index, false );
+        CorrectFlags32( col, index, false );
         break;
 
     case MP2::OBJ_ALCHEMYLAB: {
         index += 21;
         if ( Maps::isValidDirection( _index, Direction::TOP ) ) {
             Maps::Tiles & tile = world.GetTiles( Maps::GetDirectionIndex( _index, Direction::TOP ) );
-            tile.CorrectFlags32( index, true );
+            tile.CorrectFlags32( col, index, true );
         }
         break;
     }
@@ -1727,7 +1727,7 @@ void Maps::Tiles::CaptureFlags32( const MP2::MapObjectType objectType, int col )
         index += 28;
         if ( Maps::isValidDirection( _index, Direction::TOP_RIGHT ) ) {
             Maps::Tiles & tile = world.GetTiles( Maps::GetDirectionIndex( _index, Direction::TOP_RIGHT ) );
-            tile.CorrectFlags32( index, true );
+            tile.CorrectFlags32( col, index, true );
         }
         break;
     }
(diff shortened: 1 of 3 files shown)


a1exsh:
This map may be useful for testing: [ALLTEST.MP2.zip](https://github.com/ihhub/fheroes2/files/7103974/ALLTEST.MP2.zip)

Grab those Phoenixes and head to the East! ;)

a1exsh:
Before green has lost:
image

After (flags removed from all captured objects):
image

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1752,19 +1752,24 @@ void Maps::Tiles::CaptureFlags32( const MP2::MapObjectType objectType, int col )
     }
 }
 
-void Maps::Tiles::CorrectFlags32( u32 index, bool up )
+void Maps::Tiles::CorrectFlags32( int col, u32 index, bool up )
```
oleg-derevenetz, line 1755: Since function signature is modified here, it is useful to add `const` to parameter types.

src/fheroes2/maps/maps_tiles.h
```diff
@@ -305,7 +305,7 @@ namespace Maps
         TilesAddon * FindFlags( void );
 
         // correct flags, ICN::FLAGS32 vesion
-        void CorrectFlags32( u32 index, bool );
+        void CorrectFlags32( int col, u32 index, bool );
```
oleg-derevenetz, line 308: The same is here - it would be useful to add `const`.

src/fheroes2/world/world.cpp
```diff
@@ -292,8 +292,10 @@ void CapturedObjects::ResetColor( int color )
         ObjectColor & objcol = ( *it ).second.objcol;
 
         if ( objcol.isColor( color ) ) {
-            objcol.second = Color::UNUSED;
-            world.GetTiles( ( *it ).first ).CaptureFlags32( static_cast<MP2::MapObjectType>( objcol.first ), objcol.second );
+            MP2::MapObjectType objectType = static_cast<MP2::MapObjectType>( objcol.first );
```
oleg-derevenetz, line 295: I suppose, `objectType` can be declared as `const` here.

oleg-derevenetz (changes requested):
Hi @a1exsh I left a few minor comments, please take a look when it's convenient for you.

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1752,19 +1752,24 @@ void Maps::Tiles::CaptureFlags32( const MP2::MapObjectType objectType, int col )
...
-        taddon->index = index;
-    else if ( up )
-        // or new flag
-        addons_level2.emplace_back( TilesAddon::UPPER, World::GetUniq(), 0x38, index );
-    else
-        // or new flag
-        addons_level1.emplace_back( TilesAddon::UPPER, World::GetUniq(), 0x38, index );
+    if ( col == Color::NONE ) {
+        removeFlags();
+    }
```
qarkai, line 1759:
```suggestion
    if ( col == Color::NONE ) {
        removeFlags();
        return;
    }
```
And revert `else` branch whitespace changes. Patch will be smaller.

src/fheroes2/world/world.cpp
a1exsh: Yes, that makes sense.

a1exsh:
```suggestion
            const MP2::MapObjectType objectType = static_cast<MP2::MapObjectType>( objcol.first );
```

src/fheroes2/maps/maps_tiles.h
a1exsh: Any chance to skip consts on pass-by-value parameters in declarations?..

I like the symmetry and all, but it becomes really noisy very fast, IMO.

src/fheroes2/maps/maps_tiles.cpp
a1exsh:
```suggestion
void Maps::Tiles::CorrectFlags32( const int col, const u32 index, const bool up )
```

src/fheroes2/maps/maps_tiles.cpp
a1exsh: Maybe...

src/fheroes2/maps/maps_tiles.h
oleg-derevenetz: Hi @a1exsh in this project `const` is used in parameter types for the same reason as for local variables - to ensure that they are not accidentally changed somewhere. There may be use cases with non-const parameters - for example, if parameter value should be somehow capped and you don't want to add a separate variable just to cap the parameter passed by value, but in general `const` parameters are welcome in a new code.

a1exsh: I totally get that, but what I'm pointing out is that it's possible to use `const` parameters in function definitions, but omit the modifiers in the declaration, like in the line above.

The declaration is for the caller, and the caller doesn't care if implementation modifies passed by value parameters like int, bool, etc.  Of course, in case of pointers, references, composite types or containers, `const` should be used in the declarations as well.

oleg-derevenetz: Ah, you mean just function declarations... You see, when people (for example, me) write a new function, they write definition first and then just copy definition to the place where declaration resides (usually it's a header file) - or, may be, vice versa. In either case, they should either remove `const` from declaration or add them to the definition, which is extra action. I suppose that just keeping them completely identical is much easier.

a1exsh:
```suggestion
        void CorrectFlags32( const int col, const u32 index, const bool );
```

src/fheroes2/maps/maps_tiles.h
```diff
@@ -305,7 +305,7 @@ namespace Maps
         TilesAddon * FindFlags( void );
 
         // correct flags, ICN::FLAGS32 vesion
-        void CorrectFlags32( u32 index, bool );
+        void CorrectFlags32( const int col, const u32 index, const bool );
```
a1exsh, line 308:
```suggestion
        void CorrectFlags32( const int col, const u32 index, const bool up );
```

Since we are changing it anyway

oleg-derevenetz (approved):

ihhub:
Hi @a1exsh , I think we already have a solution for this. Please take a look at `Maps::Tiles::removeFlags` method inside. What we need is to call this method on all objects own by the player except castles and towns where we should set gray flags.

a1exsh:
@ihhub hey, but that's exactly what this PR does!

ihhub:
> 
> 
> @ihhub hey, but that's exactly what this PR does!

Ah, my bad. I missed one line here. My apologies. I'll review it very soon.

src/fheroes2/world/world.cpp
```diff
@@ -292,8 +292,10 @@ void CapturedObjects::ResetColor( int color )
         ObjectColor & objcol = ( *it ).second.objcol;
 
         if ( objcol.isColor( color ) ) {
-            objcol.second = Color::UNUSED;
-            world.GetTiles( ( *it ).first ).CaptureFlags32( static_cast<MP2::MapObjectType>( objcol.first ), objcol.second );
+            const MP2::MapObjectType objectType = static_cast<MP2::MapObjectType>( objcol.first );
+
+            objcol.second = objectType == MP2::OBJ_CASTLE ? Color::UNUSED : Color::NONE;
+            world.GetTiles( ( *it ).first ).CaptureFlags32( objectType, objcol.second );
```
ihhub, line 298: We could do this way:
```cpp
const MP2::MapObjectType objectType = static_cast<MP2::MapObjectType>( objcol.first );
const bool isCastle = ( objectType == MP2::OBJ_CASTLE ) || ( objectType == MP2::OBJN_CASTLE );
if ( isCastle ) {
    world.GetTiles( it->first ).CaptureFlags32( objectType, Color::UNUSED );
}
else {
    world.GetTiles( it->first ).removeFlags();
}
```
without any changes outside this method. Of course it's the same way as before but less hacky way without extra conditions in methods. What do you think?

Also please notice that `OBJN_CASTLE` should be checked too.

a1exsh: I think I've tried that initially, but it didn't work for the Sawmill and Alchemist Lab, because these two objects have special handling in that they set the flag on a _different_ tile.  So I had to move the handling all the way down to the `CorrectFlags32` function, even though that makes the change bigger.

That is `OBJN_CASTLE` and why don't have have a `case` for it in `CaptureFlag32`?

ihhub: OBJ_CASTLE is the main entrance while OBJN_CASTLE is for the rest of tiles related to castle. I think your solution is correct.

a1exsh: I also think that assignment of `objcol.second` is essential in this method and should not be removed (as in your proposed change).

ihhub:
@a1exsh , thanks a lot for this fix!

PR #4153, opened 2021-09-03
Allow hero to traverse Stone Liths and Whirlpools even if he has no movement points left

### LINKED ISSUE #4146: Adventure map, Stone Liths wastes some move points while travelling from grass to snow terrain
### Preliminary checks

### Platform

Windows

### Describe the bug

A hero with 100 movement points left enters Stone Liths. While travelling (using space) from grass to grass terrain he still has 100 points. But when a hero appears on snow 100 points are gone.
To reproduce a bug enter a Stone Liths and press space till the moment, hero appears on snow. After that you cannot travel anymore, because of empty movement bar.
image

### Save file

[Stone Liths Penalty.zip](https://github.com/ihhub/fheroes2/files/7105493/Stone.Liths.Penalty.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4146

Hi @Branikolog in fact, hero still has 100MP after teleporting to the Snow terrain (you can make sure of this by viewing hero's quick info), so no movement penalty is in fact applied. I suppose that hero still should be able to traverse through teleports if he has non-zero movement points, even if he is not able to make a step over the terrain on which teleport resides due to this terrain's penalty? If yes, then this PR should fix the issue. If no, then there is no issue :)

### DIFF
--- a/src/fheroes2/gui/interface_events.cpp
+++ b/src/fheroes2/gui/interface_events.cpp
@@ -388,18 +388,12 @@ void Interface::Basic::EventDefaultAction( void )
 
     if ( hero ) {
         // 1. action object
-        if ( MP2::isActionObject( hero->GetMapsObject(), hero->isShipMaster() ) && ( !MP2::isMoveObject( hero->GetMapsObject() ) || hero->CanMove() ) ) {
-            const Maps::Tiles & tile = world.GetTiles( hero->GetIndex() );
-
+        if ( MP2::isActionObject( hero->GetMapsObject(), hero->isShipMaster() ) ) {
             hero->Action( hero->GetIndex(), true );
-            if ( MP2::OBJ_STONELITHS == tile.GetObject( false ) || MP2::OBJ_WHIRLPOOL == tile.GetObject( false ) )
-                SetRedraw( REDRAW_HEROES );
-            SetRedraw( REDRAW_GAMEAREA );
         }
     }
-    else
+    else if ( GetFocusCastle() ) {
         // 2. town dialog
-        if ( GetFocusCastle() ) {
         Game::OpenCastleDialog( *GetFocusCastle() );
     }
 }
--- a/src/fheroes2/maps/mp2.cpp
+++ b/src/fheroes2/maps/mp2.cpp
@@ -1209,20 +1209,6 @@ bool MP2::isProtectedObject( const MapObjectType objectType )
     return isCaptureObject( objectType );
 }
 
-bool MP2::isMoveObject( const MapObjectType objectType )
-{
-    switch ( objectType ) {
-    case OBJ_STONELITHS:
-    case OBJ_WHIRLPOOL:
-        return true;
-
-    default:
-        break;
-    }
-
-    return false;
-}
-
 bool MP2::isAbandonedMine( const MapObjectType objectType )
 {
     return objectType == MP2::OBJN_ABANDONEDMINE || objectType == MP2::OBJ_ABANDONEDMINE;
--- a/src/fheroes2/maps/mp2.h
+++ b/src/fheroes2/maps/mp2.h
@@ -587,7 +587,6 @@ namespace MP2
     bool isHeroUpgradeObject( const MapObjectType objectType );
     bool isMonsterDwelling( const MapObjectType objectType );
     bool isRemoveObject( const MapObjectType objectType );
-    bool isMoveObject( const MapObjectType objectType );
     bool isAbandonedMine( const MapObjectType objectType );
     bool isProtectedObject( const MapObjectType objectType );
 


Branikolog:
> fix #4146
> 
> Hi @Branikolog in fact, hero still has 100MP after teleporting to the Snow terrain (you can make sure of this by viewing hero's quick info), so no movement penalty is in fact applied. I suppose that hero still should be able to traverse through teleports if he has non-zero movement points, even if he is not able to make a step over the terrain on which teleport resides due to this terrain's penalty? If yes, then this PR should fix the issue.

Hello, @oleg-derevenetz !
Well, here, I suppose, a question regarding our current movement meter would appear...
Actually, I forgot to check if hero still had that 100 points. Thanks for your investigation.
While teleporting through Stone Liths (there're about 5-6 of them at the map) I experienced an issue: I made a several jumps till the moment I appeared at snow. So then I couldn't jump again, which was disappointing, as I didn't make any move with the hero... So then I reloaded and the first time hero jumped into a Stone Liths he appeared at snow again and that's it. So I had to reload every time I appeared on snow terrain, till I reached the needed destination.
So, I assume, a hero's movement meter was recalculated considering snow penalty. That's why we cannot jump again, as hero has ot enough movement bar to move.
The question, that I have now, is are we going to allow yellow bar appear and disappear in movement meter, while jumping through Stone Liths between different terrain penalties? That looks kind of strange, actually.
Also, after your fix, which works quite well, hero is able to jump even without movement bar in his meter (while having some movement points). But if a hero has exactly 0 points - he can't jump. That can rise lots of questions from players... They can wonder why do hero can jump with empty bar in some case, and cannot in other.
Talking about movement meter operation I always compare with H3 (as you do, I noticed :) )
We have some differences here. In H3 movement bar has tiny green piece (In H2 it's yellow) in a case hero is able to make a movement at least to one direction. But when a hero tries to move to the direction, that requires more points that he has, the rest of the points are reset to zero. So hero cannot move anymore. 
I'm not trying to copy-paste all H3 mechanics, but our current movement mechanics is not ideal and stimulates appearing such issues...

oleg-derevenetz:
@Branikolog I suppose that if we make the movement indicator to show a yellow bar even if hero is not able to move on the current terrain, then there will be another questions from user "why my movement indicator is not empty even if I can't move?". I suppose the current behavior was designed to prevent asking questions of just this type. According to this - yes, if hero can't move over the current terrain, his movement indicator should be empty, but if he teleports to the terrain where he is able to move, his movement indicator should become yellow again.

Branikolog:
Well, movement indicator and its behaviour looks like a nice topic for discussion. ;)
Talking about current fix:
It's really not obvious, especially for newcommers, why at one save point we can jump through Stone Liths multiple times, and cannot do the same at other save point:
[Stone Liths inf jumps.zip](https://github.com/ihhub/fheroes2/files/7108460/Stone.Liths.inf.jumps.zip)
[Stone Liths 1 jump.zip](https://github.com/ihhub/fheroes2/files/7108461/Stone.Liths.1.jump.zip)
How do you think, how can we handle this case?

Original H2 doesn't allow to jump using spacebar. While H3 does. I believe in future, we can create our own functioning of teleportation mechanics... For example, apply small movement consuming (like 10-50 pts) to initiate teleportation and also allow player to choose destination point from all available. (It's just my dreams for now, nothing more. ;) )

oleg-derevenetz:
@Branikolog I think that current movement bar mechanics is OK - users will move more often than teleport, so it's better to focus it on movement (if movement is not possible, then it should be empty, just like it works now). To get the exact MP value, there is always a quick info window. As an alternative, we can allow to teleport using space bar even with zero MP, but I'm not sure about that.

LeHerosInconnu:
Hello everyone.

This is how I see it:
If the hero has at least 1 movement point (ex: 0001/1000), then he can use the teleporter with the space bar.
If the hero has 0 movement point, a message could be displayed:
"Your hero is too tired to use this Stone Liths today. Try again tomorrow."

oleg-derevenetz:
Hi @Branikolog @LeHerosInconnu technically I can even think of allowing movement via teleports using the space bar even with zero MP. If MP are not spent for this anyway, why not really? Hi @ihhub what's your opinion?

Branikolog:
> Hi @Branikolog @LeHerosInconnu technically I can even think of allowing movement via teleports using the space bar even with zero MP. If MP are not spent for this anyway, why not really? With all other objects, user is allowed to use space bar even with zero MP. Hi @ihhub what's your opinion?

Well, I expect the same behaviour for the both cases I showed in a savefiles above. It's not obvious, why player can use space to teleport in one case, but cannot in another.

LeHerosInconnu:
> Hi @Branikolog @LeHerosInconnu technically I can even think of allowing movement via teleports using the space bar even with zero MP. If MP are not spent for this anyway, why not really? With all other objects, user is allowed to use space bar even with zero MP. Hi @ihhub what's your opinion?

Good point.

> Well, I expect the same behaviour for the both cases I showed in a savefiles above. It's not obvious, why player can use space to teleport in one case, but cannot in another.

Good point too.

I'm okay with the hero still being able to use Stone Liths with the space bar even with 0000/1000 movement points.
(But it is a bit of a "cheat" for the human player, because in Heroes 2 a hero can't attack another hero who is positioned on a Stone Liths while in Heroes 3 this is possible (if I remember correctly). :D
But Heroes 2 must remain Heroes 2, so no attack through a Stone Liths.)

oleg-derevenetz:
Hi @Branikolog now hero should be able to traverse Stone Liths and Whirlpools (via space bar) even if he have no MP left. Also I removed extra `SetRedraw()` calls, hope this wouldn't break anything (according to my tests, it's not, but who knows).

ihhub:

PR #4156, opened 2021-09-04
Fix black edges on view world

### LINKED ISSUE #3157: View World screen, it should not be possible to display the adventure map beyond the edges
In fheroes2, it should not be possible to display the adventure map beyond the edges (wide black border).

In original game:

View World no bord H2 02

In fheroes2:

View World no bord fh2 01

View World no bord fh2 02

a1exsh:
Fix #3157

This prevents displaying beyond the edges of the map, ~but seems to introduce a problem when changing position by clicking on the world: it might jump around unpredictably.  Not sure why that would be the case.~

In general with big screen resolution and a small map we will display back area around the map anyway, so maybe it's not critical to try to fix it at all.

### DIFF
--- a/src/fheroes2/kingdom/view_world.cpp
+++ b/src/fheroes2/kingdom/view_world.cpp
@@ -388,24 +388,47 @@ ViewWorld::ZoomROIs::ZoomROIs( const ViewWorld::ZoomLevel zoomLevel, const fhero
     : _zoomLevel( zoomLevel )
     , _center( centerInPixels )
 {
+    updateZoomLevels();
+    updateCenter();
+}
+
+void ViewWorld::ZoomROIs::updateZoomLevels()
+{
     for ( int i = 0; i < 4; ++i ) {
         _roiForZoomLevels[i] = computeROI( _center, static_cast<ViewWorld::ZoomLevel>( i ) );
     }
 }
 
+bool ViewWorld::ZoomROIs::updateCenter()
+{
+    return ChangeCenter( _center );
+}
+
 bool ViewWorld::ZoomROIs::ChangeCenter( const fheroes2::Point & centerInPixels )
 {
     const fheroes2::Rect currentRect = GetROIinPixels();
-    const fheroes2::Point newCenter( clamp( centerInPixels.x, currentRect.width / 2, world.w() * TILEWIDTH - currentRect.width / 2 ),
-                                     clamp( centerInPixels.y, currentRect.height / 2, world.h() * TILEWIDTH - currentRect.height / 2 ) );
+    const fheroes2::Size worldSize( world.w() * TILEWIDTH, world.h() * TILEWIDTH );
+    fheroes2::Point newCenter;
+
+    if ( worldSize.width <= currentRect.width ) {
+        newCenter.x = worldSize.width / 2;
+    }
+    else {
+        newCenter.x = clamp( centerInPixels.x, currentRect.width / 2, worldSize.width - currentRect.width / 2 );
+    }
+
+    if ( worldSize.height <= currentRect.height ) {
+        newCenter.y = worldSize.height / 2;
+    }
+    else {
+        newCenter.y = clamp( centerInPixels.y, currentRect.height / 2, worldSize.height - currentRect.height / 2 );
+    }
 
     if ( newCenter == _center ) {
         return false;
     }
     _center = newCenter;
-    for ( int i = 0; i < 4; ++i ) {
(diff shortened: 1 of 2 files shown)


oleg-derevenetz:
> In general with big screen resolution and a small map we will display back area around the map anyway, so maybe it's not critical to try to fix it at all.

+1

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -395,7 +395,9 @@ ViewWorld::ZoomROIs::ZoomROIs( const ViewWorld::ZoomLevel zoomLevel, const fhero
 
 bool ViewWorld::ZoomROIs::ChangeCenter( const fheroes2::Point & centerInPixels )
 {
-    const fheroes2::Point newCenter( clamp( centerInPixels.x, 0, world.w() * TILEWIDTH ), clamp( centerInPixels.y, 0, world.h() * TILEWIDTH ) );
+    const fheroes2::Rect currentRect = GetROIinPixels();
+    const fheroes2::Point newCenter( clamp( centerInPixels.x, currentRect.width / 2, world.w() * TILEWIDTH - currentRect.width / 2 ),
+                                     clamp( centerInPixels.y, currentRect.height / 2, world.h() * TILEWIDTH - currentRect.height / 2 ) );
```
ihhub, line 400: How about this solution:
```cpp
    const fheroes2::Rect currentRect = GetROIinPixels();
    const fheroes2::Size worldSize( world.w() * TILEWIDTH, world.h() * TILEWIDTH );

    fheroes2::Point newCenter;

    if ( worldSize.width <= currentRect.width ) {
        newCenter.x = worldSize.width / 2;
    }
    else {
        newCenter.x = clamp( centerInPixels.x, currentRect.width / 2, worldSize.width - currentRect.width / 2 );
    }

    if ( worldSize.height <= currentRect.height ) {
        newCenter.y = worldSize.height / 2;
    }
    else {
        newCenter.y = clamp( centerInPixels.y, currentRect.height / 2, worldSize.height - currentRect.height / 2 );
    }
```
It solves the problem with small maps for big resolutions :)

LeHerosInconnu:
Hello everyone.

In cases the map can be displayed completely in the higher resolutions, the map should always be centered, without the possibility to move it by clicking on the mini map (what would be the use of that?).
In cases the map cannot be displayed completely, the black borders should not be displayed.

ihhub:
> 
> 
> Hello everyone.
> 
> In cases the map can be displayed completely in the higher resolutions, the map should always be centered, without the possibility to move it by clicking on the mini map (what would be the use of that?).
> In cases the map cannot be displayed completely, the black borders should not be displayed.

Hi @LeHerosInconnu , the proposed solution by me covers all your concerns :)

src/fheroes2/kingdom/view_world.cpp
ihhub: One thing I missed in this solution: we have to update the position when we change zoom.

LeHerosInconnu:
Okay @ihhub, we were writing "at the same time". :)

a1exsh:
@ihhub I've applied your suggested change and further added handling for the case when showing the world for the first time (centered on hero): it could still show the black edge, but now that's also fixed.

I'm noticing that we don't have the border rendered below the `EXIT` button, where some black area is still visible (as well as when entire world fits on the screen):
image

Should we try to render the gray stone pattern like we do on the actual adventure map instead?
image

ihhub:
Hi @a1exsh , if I click on any place for a small map which fits on a screen the map jumps to corners instead of staying in the center.

a1exsh:
> Hi @a1exsh , if I click on any place for a small map which fits on a screen the map jumps to corners instead of staying in the center.

@ihhub that was an issue in my first attempt, but it is resolved by applying your suggestion.  Please use the updated branch at: https://github.com/ihhub/fheroes2/pull/4156/commits/180a3f4d56186e9f49a311306bc744e3245cabaa

ihhub:
> 
> 
> > Hi @a1exsh , if I click on any place for a small map which fits on a screen the map jumps to corners instead of staying in the center.
> 
> @ihhub that was an issue in my first attempt, but it is resolved by applying your suggestion. Please use the updated branch at: [180a3f4](https://github.com/ihhub/fheroes2/commit/180a3f4d56186e9f49a311306bc744e3245cabaa)

ah, yes. GitHub Desktop didn't pull latest commit.

ihhub:
@a1exsh , thank you so much for fixing this issue!

PR #4161, opened 2021-09-04
Fix passability for tiles with shadows only

### LINKED ISSUE #4149: Adventure map, a hero cannot reach area near rock
### Preliminary checks

### Platform

Windows

### Describe the bug

image
Hero cannot even attack skeletons.

### Save file

[Unreachable area.zip](https://github.com/ihhub/fheroes2/files/7106055/Unreachable.area.zip)

### Additional info

_No response_

ihhub:
close #4149


src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1619,10 +1619,10 @@ bool Maps::Tiles::isStream( void ) const
     return tileICN == ICN::STREAM || ( tileICN == ICN::OBJNMUL2 && objectIndex < 14 );
 }
 
-bool Maps::Tiles::isShadow( void ) const
+bool Maps::Tiles::isShadow() const
 {
     return isShadowSprite( objectTileset, objectIndex )
-           && addons_level1.size() != static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) );
+           && addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) );
```
oleg-derevenetz, line 1625: Yeah, logic of `isShadow()` was suspicious long ago, that's why I didn't use it in `Maps::Tiles::GoodForUltimateArtifact()` at the time (and replicated the current logic). BTW, can the tile with `objectTileset == 0` have any shadow addons? I vaguely remember something like this...

oleg-derevenetz (commented):
Hi @ihhub I just left one question, could you please take a look when it's convenient for you?

src/fheroes2/maps/maps_tiles.cpp
ihhub: Yes and no. The structure of map is still not fully known to me and I saw tiles having something in addonss but not in the main tile structure.

oleg-derevenetz (approved):

PR #4163, opened 2021-09-04
Update Polish translation

Arthusppp:
Updated PL translation


ihhub:
@Arthusppp , thank you so much for these changes! I verified them with my limited knowledge of Polish :)

PR #4168, opened 2021-09-05
audio_music: log the reason why mixer could not be created

trofi:
On NixOS SDL2_mixer was not configured fully against timidity,
but it was not immediately clear:
    10:02:38: [ERROR]       PlayMusic:  music parameter was NULL

With the patch error reporting is more straightforward:
    10:03:45: [ERROR]       Play:  Couldn't open /etc/timidity/freepats.cfg

The change makes error handling in
    void Music::Play( const std::vector<uint8_t> & v, const bool loop )
    void Music::Play( const std::string & file, const bool loop )
identical.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -378,7 +378,12 @@ void Music::Play( const std::vector<uint8_t> & v, const bool loop )
 #endif
         SDL_FreeRW( rwops );
 
-        PlayMusic( mix, loop );
+        if ( !mix ) {
+            ERROR_LOG( Mix_GetError() );
+        }
+        else {
+            PlayMusic( mix, loop );
+        }
     }
 }
 


src/engine/audio.cpp
```diff
@@ -378,7 +378,12 @@ void Music::Play( const std::vector<uint8_t> & v, const bool loop )
 #endif
         SDL_FreeRW( rwops );
 
-        PlayMusic( mix, loop );
+        if ( !mix ) {
+            ERROR_LOG( Mix_GetError() );
+        }
+        else {
+            PlayMusic( mix, loop );
+        }
```
ihhub, line 386: I'd suggest to do more logical (opposite) way:
```cpp
        if ( mix != nullptr ) {
            PlayMusic( mix, loop );
            
        }
        else {
            ERROR_LOG( Mix_GetError() );
        }
```
so it's easier to understand by readers.

ihhub (changes requested):

src/engine/audio.cpp
ihhub: But since we have an identical code let's leave it as it is for now.

ihhub:
@trofi , thank you so much for your contribution to the project!

PR #4183, opened 2021-09-09
Fix Maps::getAroundIndexes()

### LINKED ISSUE #4182: Adventure map, sound of sea waves plays when hero is close to the edges of the map
### Preliminary checks

### Platform

Windows

### Describe the bug

In a case we move hero close to the left or right edges of the map we can hear sea sound on a custom created map.
image

### Save file

[Edge map sound.zip](https://github.com/ihhub/fheroes2/files/7135835/Edge.map.sound.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4182

### DIFF
--- a/src/fheroes2/game/game.cpp
+++ b/src/fheroes2/game/game.cpp
@@ -65,11 +65,12 @@ namespace
 
     std::vector<int> reserved_vols( LOOPXX_COUNT, 0 );
 
-    u32 GetMixerChannelFromObject( const Maps::Tiles & tile )
+    uint32_t GetMixerChannelFromObject( const Maps::Tiles & tile )
     {
-        // force: check stream
-        if ( tile.isStream() )
+        // check stream first
+        if ( tile.isStream() ) {
             return 13;
+        }
 
         return M82::GetIndexLOOP00XXFromObject( tile.GetObject( false ) );
     }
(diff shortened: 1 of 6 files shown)


src/fheroes2/game/game.cpp
```diff
@@ -347,24 +357,26 @@ u32 & Game::CastleAnimationFrame( void )
...
+// play environment sounds from the game area in focus
+void Game::EnvironmentSoundMixer()
 {
     const fheroes2::Point abs_pt( Interface::GetFocusCenter() );
     std::fill( reserved_vols.begin(), reserved_vols.end(), 0 );
 
-    // scan 4x4 square from focus
-    for ( s32 yy = abs_pt.y - 3; yy <= abs_pt.y + 3; ++yy ) {
-        for ( s32 xx = abs_pt.x - 3; xx <= abs_pt.x + 3; ++xx ) {
+    // scan 4x4 area in focus
```
ihhub, line 366: Actually it's 7x7 area :)

src/fheroes2/game/game.cpp
```diff
@@ -65,13 +65,23 @@ namespace
...
+        // check stream first
+        if ( tile.isStream() ) {
             return 13;
+        }
 
-        return M82::GetIndexLOOP00XXFromObject( tile.GetObject( false ) );
+        MP2::MapObjectType objectType = tile.GetObject( false );
+
+        // ignore the coast tiles at the edge of the map
+        if ( ( x == 0 || y == 0 || x == world.w() - 1 || y == world.h() - 1 ) && objectType == MP2::OBJ_COAST ) {
```
ihhub, line 80: I think the problem lies in incorrect map format as we have not only sound issue but popup window shows Beach as well. This is the issue with the original map editor which doesn't check that a tile is actually at the edge of the map. Or our `Maps::Tiles::setAsEmpty()` method is incorrect.

oleg-derevenetz: Hi @ihhub IIRC the `setAsEmpty()` is called only when there was some pickable object on coast that was picked by hero. On this map (or at least with this savefile) this isn't the case, we have multiple `OBJ_COAST` tiles from nowhere on the edge of the map. Do you want to fixup this on map load instead?

Hi @Branikolog could you please provide a original map file for this issue as well to investigate this?

src/fheroes2/game/game.cpp
oleg-derevenetz: Yes, I just paraphrased this comment without going into the details. I'll update it properly.

src/fheroes2/game/game.cpp
Branikolog: Hi, @oleg-derevenetz !
Sure! Here it is: 
[Map_0051.zip](https://github.com/ihhub/fheroes2/files/7142505/Map_0051.zip)

oleg-derevenetz: Hi @ihhub Not that the problem was directly in the `setAsEmpty()`, but sort of :)

src/fheroes2/maps/maps.cpp
```diff
@@ -288,59 +289,38 @@ int32_t Maps::GetIndexFromAbsPoint( const int32_t x, const int32_t y )
     return y * world.w() + x;
 }
 
-Maps::Indexes Maps::GetAroundIndexes( int32_t center )
+Maps::Indexes Maps::GetAroundIndexes( const int32_t tileIndex )
 {
-    Indexes result;
```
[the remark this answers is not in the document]
oleg-derevenetz: It wrapped around near edges, giving incorrect results.

ihhub:

PR #4184, opened 2021-09-09
Reset focus when hero was dismissed, use KEY_RETURN for EVENT_OPENFOCUS by default

### LINKED ISSUE #4138: Adventure map, [ENTER] / [SPACE] hotkeys behavior
### Preliminary checks

### Platform

Linux

### Describe the bug

According to the default hotkeys config, `ENTER` should open focused element, which can be a hero or a town:
```
# 2.3 game events:
...
# open focus		- (default keycode:  13 = 'return')
```

This is also how it works in the original game.

In fheroes2 pressing `ENTER` doesn't open either of the dialogs.  Pressing `SPACE` (to which the `default action` is bound by default) opens the currently selected town, but does not open the currently selected hero.

Both hotkeys should open town/hero depending on the current focus.

### Save file

No save file needed.

### Additional info

_No response_

### LINKED ISSUE #4178: Adventure map, hero scrollbar doesn't update to be centered, after some hero disappeared
### Preliminary checks

### Platform

Windows

### Describe the bug

While running fheroes2 at low resolution (up to 800x600), scrollbar doesn't become centered, in a case we dismiss a hero when all heroes doesn't fit available boxes in hero column.
Here we have 8 heroes for 7 visible slots:
image
We dismiss some heroes:
image
Scrollbar is still sticked to upper arrow. It cannot be moved.
But when we save and reload the game bar appears correctly:
image
Note, that when we loose hero in battle everything works correctly.

### Save file

[Test Scroll.zip](https://github.com/ihhub/fheroes2/files/7128550/Test.Scroll.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4138
fix #4178

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -231,48 +231,46 @@ void Game::OpenHeroesDialog( Heroes & hero, bool updateFocus, bool windowIsGameW
     const KingdomHeroes & myHeroes = hero.GetKingdom().GetHeroes();
     KingdomHeroes::const_iterator it = std::find( myHeroes.begin(), myHeroes.end(), &hero );
 
-    if ( it != myHeroes.end() ) {
-        int result = Dialog::ZERO;
+    int result = Dialog::ZERO;
 
-        while ( Dialog::CANCEL != result ) {
-            result = ( *it )->OpenDialog( false, needFade, disableDismiss );
-            if ( needFade )
-                needFade = false;
+    while ( it != myHeroes.end() && result != Dialog::CANCEL ) {
+        result = ( *it )->OpenDialog( false, needFade, disableDismiss );
+        if ( needFade )
+            needFade = false;
 
-            switch ( result ) {
-            case Dialog::PREV:
-                if ( it == myHeroes.begin() )
-                    it = myHeroes.end();
-                --it;
-                break;
+        switch ( result ) {
+        case Dialog::PREV:
+            if ( it == myHeroes.begin() )
+                it = myHeroes.end();
+            --it;
+            break;
 
-            case Dialog::NEXT:
-                ++it;
-                if ( it == myHeroes.end() )
-                    it = myHeroes.begin();
-                break;
-
-            case Dialog::DISMISS:
-                AGG::PlaySound( M82::KILLFADE );
-
-                ( *it )->GetPath().Hide();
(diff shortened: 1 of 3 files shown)


a1exsh:
I'm noticing an issue with the focus after closing the castle dialog.  Will provide details and a save file soon.

a1exsh:
> I'm noticing an issue with the focus after closing the castle dialog. Will provide details and a save file soon.

Ah, actually all is fine: the focus jumps to the hero in the castle, if there is one, after closing the castle dialog.  I think that's helpful and is by design. :)

oleg-derevenetz:
> Ah, actually all is fine: the focus jumps to the hero in the castle, if there is one, after closing the castle dialog.  I think that's helpful and is by design. :)

This PR doesn't touch the castle dialog logic. If no one has complained about this behavior yet, I believe it is correct.

a1exsh:
> This PR doesn't touch the castle dialog logic. If no one has complained about this behavior yet, I believe it is correct.

Yes, I've compared with the HEAD and this part works the same way there. I haven't checked the behavior on dismiss, though.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -231,48 +231,46 @@ void Game::OpenHeroesDialog( Heroes & hero, bool updateFocus, bool windowIsGameW
...
-            if ( needFade )
-                needFade = false;
-
-            switch ( result ) {
-            case Dialog::PREV:
-                if ( it == myHeroes.begin() )
-                    it = myHeroes.end();
-                --it;
-                break;
+    int result = Dialog::ZERO;
```
ihhub, line 234: I think besides this change we have to modify `void SetCurrentVisible` method to move **UpdateScrollbarRange();** call from `if ( _currentId >= 0 && ( _topId > _currentId || _topId + maxItems <= _currentId ) ) { // out of view` condition. Just this fix solves the whole problem with scrollbar. What do you think?

oleg-derevenetz: @ihhub May be. `ResetFocus()` also calls `ListBox::Reset()` which properly sets scrollbar range. If you wish, you can add commit with this change to this PR, or I can do it tomorrow - I need to move out now :)

ihhub:
@oleg-derevenetz , a bunch of thanks for fixing these issues!

PR #4191, opened 2021-09-11
Use bitwise OR instead of addition in deserialization code

a1exsh:
The result is the same, but OR makes more sense for bit operations (and should
be slightly more performant, though that's probably never an issue for the
data volumes we process).

### DIFF
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -394,7 +394,7 @@ u8 StreamBuf::get8()
 u16 StreamBuf::getBE16()
 {
     u16 result = ( get8() << 8 );
-    result += get8();
+    result |= get8();
 
     return result;
 }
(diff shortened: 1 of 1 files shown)


a1exsh:
One of the checks failed with `Error sending AppVeyor compilation message: The remote server returned an error: (500) Internal Server Error`:
https://ci.appveyor.com/project/ihhub/fheroes2/builds/40729157/job/ycgu877asny32ras#L223

Can someone re-run it please?

oleg-derevenetz:
> Can someone re-run it please?

You can re-run it by force-pushing the same commit (well, not exactly the same, but with different commit metadata - for example, commit time, something to change the commit hash).

a1exsh:
OK, now it says "All checks have passed" :)

ihhub:
@a1exsh , thank you very much for this improvement!

PR #4194, opened 2021-09-11
Update Polish translation

Arthusppp:
Update Polish translation


ihhub:
@Arthusppp , thank you so much for this improvement!

PR #4195, opened 2021-09-12
Fix post battle crash

ihhub:
introduced in #4106


ihhub:
Hi @a1exsh , just for your information. We forgot to add extra verification.

a1exsh:
Oops!

PR #4204, opened 2021-09-12
Add hotkey for View World dialog

a1exsh:
Use the same default key as in the Original Game.

Especially useful for testing local changes. ;)


ihhub:
@a1exsh , thank you so much for this improvement!

Issue #3903, opened 2021-07-25
Remove move point and resource cost from spells


vincent-grosbois:
Hello
it seems that in spells defintion and code, there is the possibility that each spells also costs either resources or move points.
I think it's unused
Can we remove it?

oleg-derevenetz:
Hi @vincent-grosbois AFAIK only in H3 spells may consume movement points (for example, Dimension Door consumes them), but in H2 it seems to not be the case. However, I'm not an expert here. @Branikolog @LeHerosInconnu do you have any advice?

ihhub:
Isn't Dimension Door eating move points as well as Town Gate? For example, function `bool HeroesTownGate( Heroes & hero, const Castle * castle )` has **hero.ApplyPenaltyMovement( townGatePenalty );**.

LeHerosInconnu:
Hello everyone,

The penalty for using the spells "Dimension Door", "Town Gate" and "Town Portal" is 225 movement points.
The hero can nevertheless cast the spell if he has at least 1 movement point left (not 100 points for a horizontal or vertical movement point on the grass terrain for example, but really 1 point (0001)).

Related closed issue: https://github.com/ihhub/fheroes2/issues/1325.

oleg-derevenetz:
> The hero can nevertheless cast the spell if he has at least 1 movement point left

Ah, that's why I decided that DD doesn't consume movement.

LeHerosInconnu:
Hello @oleg-derevenetz,

> The hero can nevertheless cast the spell if he has at least 1 movement point left (not 100 points for a horizontal or vertical movement point on the grass terrain for example, but really 1 point (0001)).

In the original game, I was sure that the hero could teleport even if he couldn't move on the adventure map, but I was wrong about the minimum value of hero's movement points needed for that.
After some tests, it seems that the minimum value of movement points of the hero is as long as there is a yellow portion left in the hero's movement indicator.
The spell can be cast with 75 movement points that's for sure, but I don't know what is the minimum number of points for a yellow portion to be displayed in the hero's movement indicator, with 50 movement points there is no more yellow portion in the movement indicator and the hero can't cast a teleportation spell, between the two values, I don't know.
However, in the original game, if the hero has at least a yellow portion left in the hero's movement indicator and even if he can't move on the adventure map, he can still teleport.

In the example, in the original game, the hero has 175 movement points left.
[...]

PR #4205, opened 2021-09-13
Add Spell description window in monster info dialog

ihhub:
Besides this:
- fix few spell messages in the game
- optimize ground type translation calls (no need to translate all in once)
- remove MAXU16 macro
- remove resource cost and state of spells. They were always 0. Relates to #3903
- add translations to some missing strings
- fix few compilation warnings
- fix one Polish sentence (missing capital letter)

image
image

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -566,7 +566,7 @@ namespace AI
 
         // Walkers: move closer to the castle walls during siege
         if ( _attackingCastle && target.cell == -1 ) {
-            uint32_t shortestDist = 65535;
+            uint32_t shortestDist = UINT32_MAX;
 
             for ( const int wallIndex : underWallsIndicies ) {
                 if ( !arena.hexIsPassable( wallIndex ) ) {
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -566,7 +566,7 @@ namespace AI
 
         // Walkers: move closer to the castle walls during siege
         if ( _attackingCastle && target.cell == -1 ) {
-            uint32_t shortestDist = MAXU16;
+            uint32_t shortestDist = 65535;
 
             for ( const int wallIndex : underWallsIndicies ) {
                 if ( !arena.hexIsPassable( wallIndex ) ) {
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -994,8 +994,8 @@ bool ArtifactsBar::ActionBarLeftMouseDoubleClick( Artifact & art )
 
             if ( spell.MovePoint() ) {
                 text += '\n';
-                text.append( _( "Move points: " ) );
-                text += std::to_string( spell.MovePoint() );
+                text.append( _( "Move points: %{mp}" ) );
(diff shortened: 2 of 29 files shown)


ihhub:
Hi @LeHerosInconnu and @Branikolog , since we're showing spell information there's one spell which doesn't exist in the Magic Book: Medusa's stone gaze. I named the spell as `Petrification` with description `Turns the targeted creature into stone.  A petrified creature receives half damage from a direct attack.` Could you please let me know if you have a better idea about the spell?

Branikolog:
> Hi @LeHerosInconnu and @Branikolog , since we're showing spell information there's one spell which doesn't exist in the Magic Book: Medusa's stone gaze. I named the spell as `Petrification` with description `Turns the targeted creature into stone. A petrified creature receives half damage from a direct attack.` Could you please let me know if you have a better idea about the spell?

Hi, @ihhub !
In descriptions of other spells in game "targeted" is often written in different words. For example: "selected" or "affected". Since the effect is not applied by a hero (Only Medusas have such ability for now), wouldn't it be better to change "targeted" for "affected"?

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -566,7 +566,7 @@ namespace AI
 
         // Walkers: move closer to the castle walls during siege
         if ( _attackingCastle && target.cell == -1 ) {
-            uint32_t shortestDist = MAXU16;
+            uint32_t shortestDist = 65535;
```
oleg-derevenetz, line 569: (optional) We can use either `UINT32_MAX` from `<cstdint>` or `std::numeric_limits<uint32_t>::max` from `<limits>` here and in other similar places.

src/fheroes2/resource/artifact.cpp
```diff
@@ -986,27 +986,19 @@ bool ArtifactsBar::ActionBarLeftMouseDoubleClick( Artifact & art )
...
 
             StringReplace( text, "%{spell}", spell.GetName() );
             StringReplace( text, "%{sp}", spell.SpellPoint() );
 
             if ( spell.MovePoint() ) {
-                text.append( "\n" );
-                text.append( "Move point: %{mp}" );
-                StringReplace( text, "%{mp}", spell.MovePoint() );
+                text += '\n';
+                text.append( _( "Move points: " ) );
```
oleg-derevenetz, line 997: In general, formatted strings (in form of "Move points: %{mp}") may be preferable in comparison with basic string concatenation in case we ever add support for right-to-left writing languages. Although of course such formatting requires a certain accuracy from translators.

oleg-derevenetz (approved):
Hi @ihhub LGTM in general, I left only two optional comments.

LeHerosInconnu:
Hello @ihhub and @Branikolog,

> > Hi @LeHerosInconnu and @Branikolog , since we're showing spell information there's one spell which doesn't exist in the Magic Book: Medusa's stone gaze. I named the spell as `Petrification` with description `Turns the targeted creature into stone. A petrified creature receives half damage from a direct attack.` Could you please let me know if you have a better idea about the spell?
> 
> Hi, @ihhub !
> In descriptions of other spells in game "targeted" is often written in different words. For example: "selected" or "affected". Since the effect is not applied by a hero (Only Medusas have such ability for now), wouldn't it be better to change "targeted" for "affected"?

"Petrification" is right on for me.
And yes, "affected" is a better use here, as it is not a "voluntary" action by Medusa.

Issue #4206, opened 2021-09-13
Deterministic battles cause different outcome in instant and manual battles


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

![image](https://user-images.githubusercontent.com/55348946/133069577-ac469721-643e-4274-98e0-5feddb491e34.png)
While attacking the same creature we have different outcome each reload:
![image](https://user-images.githubusercontent.com/55348946/133069658-7a084b01-d38f-4611-a1bf-3458fbe5526a.png)
![image](https://user-images.githubusercontent.com/55348946/133069720-b55ddc4e-83d5-406d-b78c-79459e250ae2.png)
![image](https://user-images.githubusercontent.com/55348946/133069804-b0a1a518-dda1-49dd-8fa6-d15eda75b249.png)

Also, correct me if I'm wrong, deterministic option being turned on should lead to the same battle results: the same damage, morale effects and the same AI behaviour.
Currently the same battle goes completely different on a reload:
![image](https://user-images.githubusercontent.com/55348946/133070451-7a3592f5-dc5f-46b0-8497-ead696eb9914.png)
![image](https://user-images.githubusercontent.com/55348946/133070476-c2e74e54-5df6-4a10-af66-84761e1477a1.png)

Replaying the battle manually and allowing AI make all by himself also leads to different outcome.
![image](https://user-images.githubusercontent.com/55348946/133070739-33e5e751-3970-4771-98e6-29653a1d030b.png)
![image](https://user-images.githubusercontent.com/55348946/133070832-31eb27d4-9556-46e7-9b74-e57b5284c1eb.png)

### Save file

[Non deterministic battle.zip](https://github.com/ihhub/fheroes2/files/7153627/Non.deterministic.battle.zip)

vincent-grosbois:
that's strange... do you know what is the first thing that AI does differently on different reloads ? it could be something that I forgot to make "deterministic" and that is still fully random

Branikolog:
Hi, @vincent-grosbois 
Even damage dealt by hobbits on 5-6 screenshot was different in a same circumstances.

vincent-grosbois:
[...]

PR #4208, opened 2021-09-13
Fix deterministic battle outcome

vincent-grosbois:
Fix for #4206


src/fheroes2/battle/battle_arena.h
```diff
@@ -51,6 +51,22 @@ namespace Battle
     {
     };
 
+    class TroopsUidGenerator
+    {
+    private:
+        uint32_t _id{ 0 };
```
ihhub, line 57: Could you please move private section after public section to keep the same structure of classes?

ihhub:

Branikolog:
Hi, @vincent-grosbois & @ihhub.
With "Deterministic" option enabled everything seems to work correctly from the first sight.
I don't clearly remember the initial "deterministic" issue, but if I'm not mistaken even with the option disabled instant battles should also have the same outcome. Which is not the case with this fix...

oleg-derevenetz:
@Branikolog

> I don't clearly remember the initial "deterministic" issue, but if I'm not mistaken even with the option disabled instant battles should also have the same outcome. Which is not the case with this fix...

With or without this option instant battles should have the same outcome **_with auto battles_** (if the player still chooses manual battle after the instant battle and presses the "AUTO" button immediately). There is the only guarantee in non-deterministic mode.

Branikolog:
> > I don't clearly remember the initial "deterministic" issue, but if I'm not mistaken even with the option disabled instant battles should also have the same outcome. Which is not the case with this fix...
> 
> With or without this option instant battles should have the same outcome **_with auto battles_** (if the player still chooses manual battle after the instant battle and presses the "AUTO" button immediately). This is [the only guarantee](https://github.com/ihhub/fheroes2/pull/4072#issuecomment-907208142) in non-deterministic mode.

Got it. So there's a problem then.
Also, I've found an issue with deterministic option: The first turn my creature has Morale triggered. So it is triggered each time, no matter where do I place the creature on reload. Is it ok?
Also, If I completely rearrange my army, morale is still triggered for the creature. Only if dismiss ALL other creatures Morale wouldn't trigger.
[Music loop.zip](https://github.com/ihhub/fheroes2/files/7170287/Music.loop.zip)
Here's a savefile, where morale always triggers for Boars. (Except the case I dismiss the whole army except of boars)

oleg-derevenetz:
@Branikolog

> Got it. So there's a problem then.

In non-deterministic mode the outcomes of instant and following "full auto" battles are different?

> Also, I've found an issue with deterministic option: The first turn my creature has Morale triggered. So it is triggered each time, no matter where do I place the creature on reload. Is it ok?

As I see, in the current implementation the initial seed depends only on the battle location (map cell number), the composition of the army does not matter. So currently the first morale event does not depend on the unit's location, only on the turn order (which depends mostly on the unit's speed).

Branikolog:
> In non-deterministic mode the outcomes of instant and following "full auto" battles are different?

The outcome is different on reloading the same game and making an instant battle. I suppose, the outcome in this case should be the same, so people won't abuse reloading and making instant battles, right?


 
> As I see, in the current implementation the initial seed depends only on the battle location (map cell number), the composition of the army does not matter. So currently the first morale event does not depend on the unit's location, only on the turn order (which depends mostly on the unit's speed).

If I split a unit into two - the first gets morale triggered too. 
I assume, if I initially had two stacks of the same type and unite them - morale would trigger anyway. I think this can be abused.

There's also one more issue:
Instant battle in deterministic mode (I suppose for both modes) doesn't consider battle dwarves' resistance: I manage to get two different outcomes in "Auto Resolve" mode (Using spells).

vincent-grosbois:
Sorry im lost, how many issues are remaining?:D
I see:
- issue that the seed only depends on map location and not army itself (can be fixed easily)
- issue about battle resistance for dwarves
- another one? Im not sure

Envoyé de mon iPhone

Branikolog:
@vincent-grosbois 
In non-deterministic mode the result of Instant battle should be the same each reload (Curently it's the the same only within instant and auto battle (in manual mode)) . Otherwise, reloading can be abused. We don't want to allow players cheating like that. ;)

oleg-derevenetz:
@Branikolog

> In non-deterministic mode the result of Instant battle should be the same each reload

No, it was already discussed [here](https://github.com/ihhub/fheroes2/pull/4072#issuecomment-907208142). It should only be the same after each reload in the deterministic mode.

ihhub:
Hi @Branikolog , @oleg-derevenetz and @vincent-grosbois , let's please focus in this pull request only on the current issue. If we see that something we need to change we will create a new issues (or few of them) and fix them separately. I think in this way we can reduce the time spending on debates :)

@Branikolog, if this pull request fixes the issue described in #4206 then we should proceed with merging.

oleg-derevenetz:
Hi @ihhub There is still possibility of different outcome in case of dwarves spell resistance, so this should be 100% fixed before merging.

vincent-grosbois:
I can do that one in another pr though
I agree it's probably better to merge this fix already

Envoyé de mon iPhone

Branikolog:
Okay then. I missed the discussion of previous PR, probably I'm wrong and we should allow different outcome for instant battles in non-deterministic mode.
I'll read the thread later. For now only Resistance and army composition are considered as issues.

Branikolog:
@oleg-derevenetz 
Well, I got your point. Sorry for such an offensive politics from my side. I missed that PR with the discussions in there. 
Anyway, I still feel something wrong in that non-deterministic logic. :)
If we allow different outcome on reload, so we should make the whole behaviour non-deterministic.
I consider the non-deterministic mode as hardcore way to play. The idea is: if I'm not satisfied with the outcome the instant battle offers me, so then I enter manual mode and start playing on my own. After battlefield loaded and I realized, that enemy forces are too strong - there's no way back: I should make whole battle on my own. Pressing the "auto" (to repeat the instant battle outcome) is like too easy way ;) So results I expect be different in that case. My idea is to make each battle unpredictable. 
Deterministic battles are replayable (It can be achieved by reload or auto battle inside manual mode. Also, I know in H3 there exists "fast auto battle" option, which allows to finish battle in Auto mode instantly. So engagement of this feature in the beginning of the battle would bring the same  to Instant battle outcome.) .
So non-deterministic ones shouldn't allow to bring a player a predictable outcome. 
I can accept different outcome on reload for non-deterministic battles, but auto battle should also bring different outcome. What is the point to start manual battle and allow AI to make all work by himself? :)
Anyway, that's just my thoughts, which I should better post at some discussion threads.;)

oleg-derevenetz:
@Branikolog so, as I understand, you want to make "full auto" results different from the results of an instant battle in the non-deterministic mode and use the AUTO button as a second chance to get different results from the instant battle without manual intervention. Technically in can be done easily, but there are some downsides:

1. It happens quite often that you get unexpectedly good results from instant battle and it's just curious how these results were achieved by the AI. Now you can just replay the full battle (even in non-deterministic mode) to see this. If full auto battle will not be able to reproduce the instant battle, it will not be possible to do this.
2. Player can start a manual battle just to see if there is possible to do something with the outcome of the instant battle. He can start the battle in AUTO mode and then turn it off if he sees that something is going wrong (or not turn it off if it is obvious that nothing is possible to do here). If full auto battle will not be able to reproduce the instant battle, this approach would not work either.

Branikolog:
@oleg-derevenetz 
All your statements are true. 
But if you're here to rise your skills - you'd better use "cheating" deterministic mode. ;) 
Real "Pro" would make the outcome much better, than AI can. ;) So no need to use Auto at all. 
Frankly speaking, I've never (for about 20 years) used that auto battle.)

oleg-derevenetz:
@Branikolog I suppose it would be really better to discuss it separately with other project participants, since it looks like this discussion is out of this PR's topic :) Currently at least in this regard everything works as intended.

vincent-grosbois:
@oleg-derevenetz @Branikolog I pushed 2 commits to fix the 2 remaining issues

src/fheroes2/battle/battle_action.cpp
```diff
@@ -818,7 +818,7 @@ Battle::TargetsInfo Battle::Arena::GetTargetsForSpells( const HeroBase * hero, c
         for ( auto & tgt : targets ) {
             const uint32_t resist = tgt.defender->GetMagicResist( spell, hero ? hero->GetPower() : 0 );
 
-            if ( 0 < resist && 100 > resist && resist >= Rand::Get( 1, 100 ) ) {
+            if ( 0 < resist && 100 > resist && resist >= _randomGenerator.Get( 1, 100 ) ) {
```
felix642, line 821: From what I've seen in this project, we usually put the variable at the left of the operator and the constant at the right. 
My brain doesn't quite like it when you switch those.
Would you mind changing it?

src/fheroes2/battle/battle_troop.h
```diff
@@ -75,7 +75,7 @@ namespace Battle
     class Unit : public ArmyTroop, public BitModes, public Control
     {
     public:
-        Unit( const Troop &, s32 pos, bool reflect, const Rand::DeterministicRandomGenerator & randomGenerator );
+        Unit( const Troop &, s32 pos, bool reflect, const Rand::DeterministicRandomGenerator & randomGenerator, const uint32_t uid );
```
felix642, line 78: Since you changed that line, would you mind changing the s32 to int32_t ?

felix642 (commented):
Hi @vincent-grosbois, I know I'm a bit late to the party (: 
I left 2 minor comments on your pull request.
Really good job by the way, I really like the code you wrote!

vincent-grosbois:
@ihhub what do you think about @felix642 remarks ? can we merge as-is ?

ihhub:
@vincent-grosbois , thanks a lot for fixing this issue!

PR #4210, opened 2021-09-14
Battle command cleanup

oleg-derevenetz:
* Remove unused battle commands;
* Remove the unused variant of MSG_BATTLE_MOVE;
* Change the type of the battle command type to a named enum class.


vincent-grosbois:
Hi @oleg-derevenetz 
for ease of debug, can you also change the type of command type variable so instead of being an int, it's the type of the command enum type?

ihhub:
@oleg-derevenetz , a bunch of thanks for this code cleanup!

Issue #2406, opened 2020-12-15
Creature separation window, the number of creatures should be updated when the player clicks on the buttons of the number of troops to be separated


LeHerosInconnu:
In fheroes2, the number of creatures should be updated when the player clicks on the buttons of the number of troops to be separated.

In the example, the number "12" is not updated when the player clicks the buttons "3", "4" and "5", the same when the number is changed to "16".

![Separation number not updated fh2 01](https://user-images.githubusercontent.com/43583994/102272117-8d79b880-3f20-11eb-90f6-c147d4b1a887.gif)

PR #4211, opened 2021-09-14
Troops split logic change

### LINKED ISSUE #776: Min/Max buttons priority in separation troop dialog window
While separating units dialog window has default "MAX" button.
I think it should be "MIN" on default, as users usually separate 1 unit from the main stack. 
So now we have to press button MAX/MIN twice, to choose 1 unit.
If "MIN" button would be default, we have to press once to separate 1 unit from stack.
separation

Arthusppp:
Edit: the implemented algorithm is described in [the following comment](https://github.com/ihhub/fheroes2/pull/4211#issuecomment-965673032). 

-----
The army split logic after this change:

Case#1 splitting a troop into an empty slot -> the fhereoes2 moves a troop into the destination slot
(no change here)

Case#2 splitting troops into an empty slot -> by default, the fheroes2 proposes to split one troop into the destination slot (the main idea of the issue)
(before the change, the fheroes2 proposed "fast split" (divides troops equally into both slots)

Case#3 splitting troops into the same troop type -> by default system proposes "fast split" (divides troops equally into both slots)
(before the change, the fheroes2 moved one troop from the source slot to the destination slot)

The rest remains unchanged.

My recommendation: the designer should review the whole logic here. I.e. "fast split" is an awesome option, however, it is a bit blurry due to the UI redistribute number is displayed as enabled. (imho when a player uses "fast split", the number of troops to move should be disabled; also, for example, I can see two options here: "fast split troops equally into all available slots"; "fast split one troop into all empty slots").

Fixes #776

### DIFF
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -62,6 +62,11 @@ namespace
             troopFrom.SetCount( 1 );
             troopTarget.Set( troopFrom.GetMonster(), 1 );
         }
+        else if ( isSameTroopType && troopFrom.isValid() && troopFrom.GetCount() == 1 && troopTarget.isValid() && troopTarget.GetCount() == 1 ) {
+            // a player splits the same troop type and both count one; move a monster from the source slot to the target slot.
+            troopFrom.Reset();
+            troopTarget.SetCount( 2 );
+        }
         else {
             uint32_t freeSlots = static_cast<uint32_t>( 1 + armyTarget->Size() - armyTarget->GetCount() );
 
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -69,22 +69,9 @@ namespace
                 ++freeSlots;
 
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
-            uint32_t redistributeCount = troopFrom.GetCount() / 2;
-            const uint32_t halfOverallCount = overallCount / 2;
+            uint32_t redistributeCount = isSameTroopType ? 1 : troopFrom.GetCount() / 2;
 
-            if ( isSameTroopType ) {
-                // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by
-                // default)
-                redistributeCount = halfOverallCount > troopTarget.GetCount() ? halfOverallCount - troopTarget.GetCount() : troopTarget.GetCount() - halfOverallCount;
-
-                // possible when merging two slots of the same unit and there is a count difference, ie. 2 troops + 14 troops
-                if ( redistributeCount > maxCount )
-                    redistributeCount = maxCount;
-            }
-
-            // if splitting to the same troop type, use this bool to turn on fast split option at the beginning of the dialog
-            bool useFastSplit = true;
-
+            bool useFastSplit = !isSameTroopType;
             const uint32_t slots = Dialog::ArmySplitTroop( ( freeSlots > overallCount ? overallCount : freeSlots ), maxCount, redistributeCount, useFastSplit );
 
             if ( slots < 2 || slots > 6 )
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -70,7 +70,7 @@ namespace
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
             uint32_t redistributeCount = troopFrom.GetCount() / 2;
 
-            // if same type, then display how amny troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by default)
+            // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by default)
             if ( isSameTroopType ) {
                 redistributeCount
                     = ( overallCount / 2 ) > troopTarget.GetCount() ? ( overallCount / 2 ) - troopTarget.GetCount() : troopTarget.GetCount() - ( overallCount / 2 );
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -70,7 +70,8 @@ namespace
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
             uint32_t redistributeCount = troopFrom.GetCount() / 2;
 
-            // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by default)
+            // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by
+            // default)
             if ( isSameTroopType ) {
                 redistributeCount
                     = ( overallCount / 2 ) > troopTarget.GetCount() ? ( overallCount / 2 ) - troopTarget.GetCount() : troopTarget.GetCount() - ( overallCount / 2 );
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -42,6 +42,7 @@ namespace
         const Army * armyFrom = troopFrom.GetArmy();
         const bool saveLastTroop = armyFrom->SaveLastTroop() && armyFrom != armyTarget;
         const bool isSameTroopType = troopTarget.isValid() && troopFrom.GetID() == troopTarget.GetID();
+        const bool isTargetEmpty = troopTarget.isEmpty();
         const uint32_t overallCount = isSameTroopType ? troopFrom.GetCount() + troopTarget.GetCount() : troopFrom.GetCount();
 
         assert( overallCount > 0 );
@@ -64,10 +65,10 @@ namespace
                 ++freeSlots;
 
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
-            uint32_t redistributeCount = isSameTroopType ? 1 : troopFrom.GetCount() / 2;
+            uint32_t redistributeCount = (isSameTroopType || isTargetEmpty) ? 1 : troopFrom.GetCount() / 2;
 
-            // if splitting to the same troop type, use this bool to turn off fast split option at the beginning of the dialog
-            bool useFastSplit = !isSameTroopType;
+            // if splitting to the same troop type, use this bool to turn on fast split option at the beginning of the dialog
+            bool useFastSplit = isSameTroopType && !isTargetEmpty;
             const uint32_t slots = Dialog::ArmySplitTroop( ( freeSlots > overallCount ? overallCount : freeSlots ), maxCount, redistributeCount, useFastSplit );
 
             if ( slots < 2 || slots > 6 )
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -64,11 +64,29 @@ namespace
             if ( isSameTroopType )
                 ++freeSlots;
 
-            const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
-            uint32_t redistributeCount = ( isSameTroopType || isTargetEmpty ) ? 1 : troopFrom.GetCount() / 2;
+            const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() : troopFrom.GetCount() - 1;
 
-            // if splitting to the same troop type, use this bool to turn on fast split option at the beginning of the dialog
-            bool useFastSplit = isSameTroopType && !isTargetEmpty;
+            // by default, the game proposes the fast split option
+            bool useFastSplit = true;
+
+            // if splitting troops and the destination slot contains the same troop type, then we want to divide the sum of all troops from both slots
+            // the default redistribute value represents how many troops should be moved to achieve equal slots
+            uint32_t redistributeCount = 1;
+
+            if ( useFastSplit ) {
+                if ( isSameTroopType ) {
+                    redistributeCount
+                        = ( overallCount / 2 ) > troopTarget.GetCount() ? ( overallCount / 2 ) - troopTarget.GetCount() : troopTarget.GetCount() - ( overallCount / 2 );
+
+                    // possible when merging two slots of the same unit and there is a count difference, ie. 2 Titan + 14 Titans
+                    if ( redistributeCount > maxCount )
+                        redistributeCount = maxCount;
+                }
+                else {
+                    redistributeCount = troopFrom.GetCount() / 2;
+                }
+            }
+
             const uint32_t slots = Dialog::ArmySplitTroop( ( freeSlots > overallCount ? overallCount : freeSlots ), maxCount, redistributeCount, useFastSplit );
 
             if ( slots < 2 || slots > 6 )
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -57,34 +57,31 @@ namespace
 
             Army::SwapTroops( troopFrom, troopTarget );
         }
+        else if ( !troopTarget.isValid() && troopFrom.GetCount() == 2 ) {
+            troopFrom.SetCount( 1 );
+            troopTarget.Set( troopFrom.GetMonster(), 1 );
+        }
         else {
             uint32_t freeSlots = static_cast<uint32_t>( 1 + armyTarget->Size() - armyTarget->GetCount() );
 
             if ( isSameTroopType )
                 ++freeSlots;
 
-            const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() : troopFrom.GetCount() - 1;
+            const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
+            uint32_t redistributeCount = troopFrom.GetCount() / 2;
 
-            // by default, the game proposes the fast split option
-            bool useFastSplit = true;
+            // if same type, then display how amny troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by default)
+            if ( isSameTroopType ) {
+                redistributeCount
+                    = ( overallCount / 2 ) > troopTarget.GetCount() ? ( overallCount / 2 ) - troopTarget.GetCount() : troopTarget.GetCount() - ( overallCount / 2 );
 
-            // if splitting troops and the destination slot contains the same troop type, then we want to divide the sum of all troops from both slots
-            // the default redistribute value represents how many troops should be moved to achieve equal slots
-            uint32_t redistributeCount = 1;
-
-            if ( useFastSplit ) {
-                if ( isSameTroopType ) {
-                    redistributeCount
-                        = ( overallCount / 2 ) > troopTarget.GetCount() ? ( overallCount / 2 ) - troopTarget.GetCount() : troopTarget.GetCount() - ( overallCount / 2 );
-
-                    // possible when merging two slots of the same unit and there is a count difference, ie. 2 Titan + 14 Titans
-                    if ( redistributeCount > maxCount )
-                        redistributeCount = maxCount;
-                }
-                else {
-                    redistributeCount = troopFrom.GetCount() / 2;
-                }
+                // possible when merging two slots of the same unit and there is a count difference, ie. 2 Titan + 14 Titans
+                if ( redistributeCount > maxCount )
+                    redistributeCount = maxCount;
             }
+
+            // if splitting to the same troop type, use this bool to turn on fast split option at the beginning of the dialog
+            bool useFastSplit = true;
 
             const uint32_t slots = Dialog::ArmySplitTroop( ( freeSlots > overallCount ? overallCount : freeSlots ), maxCount, redistributeCount, useFastSplit );
 
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -58,6 +58,7 @@ namespace
             Army::SwapTroops( troopFrom, troopTarget );
         }
         else if ( !troopTarget.isValid() && troopFrom.GetCount() == 2 ) {
+            // a player splits a slot with two monsters into an empty slot; move one monster into the source slot to the target slot.
             troopFrom.SetCount( 1 );
             troopTarget.Set( troopFrom.GetMonster(), 1 );
         }
@@ -69,14 +70,14 @@ namespace
 
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
             uint32_t redistributeCount = troopFrom.GetCount() / 2;
+            const uint32_t halfOverallCount = overallCount / 2;
 
-            // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by
-            // default)
             if ( isSameTroopType ) {
-                redistributeCount
-                    = ( overallCount / 2 ) > troopTarget.GetCount() ? ( overallCount / 2 ) - troopTarget.GetCount() : troopTarget.GetCount() - ( overallCount / 2 );
+                // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by
+                // default)
+                redistributeCount = halfOverallCount > troopTarget.GetCount() ? halfOverallCount - troopTarget.GetCount() : troopTarget.GetCount() - halfOverallCount;
 
-                // possible when merging two slots of the same unit and there is a count difference, ie. 2 Titan + 14 Titans
+                // possible when merging two slots of the same unit and there is a count difference, ie. 2 troops + 14 troops
                 if ( redistributeCount > maxCount )
                     redistributeCount = maxCount;
             }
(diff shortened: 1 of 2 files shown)


src/fheroes2/army/army_bar.cpp
```diff
@@ -64,10 +65,10 @@ namespace
                 ++freeSlots;
 
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
-            uint32_t redistributeCount = isSameTroopType ? 1 : troopFrom.GetCount() / 2;
+            uint32_t redistributeCount = ( isSameTroopType || isTargetEmpty ) ? 1 : troopFrom.GetCount() / 2;
 
-            // if splitting to the same troop type, use this bool to turn off fast split option at the beginning of the dialog
-            bool useFastSplit = !isSameTroopType;
+            // if splitting to the same troop type, use this bool to turn on fast split option at the beginning of the dialog
+            bool useFastSplit = isSameTroopType && !isTargetEmpty;
```
a1exsh, line 71: Looks like the second part is not needed: if we it's the same troop type, that means the target cannot empty.  Or am I missing something?

Branikolog:
Hello, @Arthusppp !
Initially issue #776 was made to ease the case, when we want to split 1 unit from the stack.
The #2 case you've proposed makes splitting 1 troop from the main stack on default. 
Previously we divided stack 50/50. Here we should consider the most common situation: do players split 1 unit or divide stack 50\50 more often? I believe that 50\50 is more often case, as Heroes 2 is not Heroes 3, where we have 7 slots. In H2 we have just 5, so we rarely put all single units in slots, as we have to hold enough army. 
Secondly, at splitting window we have both buttons: "MIN" to set 1 unit to split and [2] to make 50\50 splitting. 
image
"MIN" button is larger, so it's much easier to click on it. comparing with [2] button. So, if we consider dividing 50\50 and splitting only 1 unit from a stack as equal cases , pressing larger "MIN" button is more comfortable, than smaller [2].
Here I should mention #776. On a default currently we have MAX button. So to split 1 unit from a stack we should press the "MAX/MIN" button twice. Making "MIN" on default will make splitting 1 unit easy and fast.

Talking about #3 case. 
Adding 1 unit to an existing stack or making both stack equal are the cases that both happen quite often. BUT splitting 1 unit was preferred on default to avoid the situation, when we split over the same size stack:
image
image
We press OK on a default dialog and nothing changes.

src/fheroes2/army/army_bar.cpp
Arthusppp: No, you are not missing anything, you are correct. Basically, the useFastSplit will always be false (instead of the "logic" I introduced here).

As I stated in the PR's description, the "fast split" is an awesome feature, however, it should be holistically reworked. In this issue, the "useFastSplit" must be false (to prevent fast splitting when a player simply clicks the OK button when splitting into an empty slot).

oleg-derevenetz:
Honestly this looks like a "matter-of-personal-taste" change. For example, regarding case N2 - if you want to make single-unit stacks (to minimize the retaliation feedback from the enemy stacks) you can use the ctrl + left mouse button, which is specifically intended for this.

Branikolog:
> use the ctrl + left mouse button, which is specifically intended for this.

Hi, @oleg-derevenetz I have to disagree with your statement, as we should offer a player the same functionality With or Without using keyboard (or at least try to). But on subject, I believe, that default MIN button in this window will be better, that making default split into 1, as we're too limited in army slots to full them with "ones" so often. I personally use 50\50 splitting way more often.
That's my personal point and it is debatable.
But the case#3 is definitely working not properly.

@Arthusppp 
>  it is a bit blurry due to the UI redistribute number is displayed as enabled. (imho when a player uses "fast split", the number of troops to move should be disabled;
We have an opened issue about splitting dialog here: #2406
Also, we would like to add dynamic title to this window. Originally posted by @LeHerosInconnu in https://github.com/ihhub/fheroes2/discussions/2821#discussioncomment-386746_
Which is still in discussion.

> I can see two options here: "fast split troops equally into all available slots"

I believe, that the maximum number of splitting stacks ( [5] on a screenshot ) makes the same thing.

oleg-derevenetz:
@Branikolog

> we should offer a player the same functionality With or Without using keyboard (or at least try to).

Sometimes it's just not possible without extra overloading the GUI. Let's take the "press space bar to retry object visiting" mechanics as an example - there is just no option to do it via GUI.

Branikolog:
> Sometimes it's just not possible without extra overloading the GUI. Let's take the "press space bar to retry object visiting" mechanics as an example - there is just no option to do it via GUI.

Unfortunately, you're right. At least we should strive, at the same time trying to keep GUI balanced.
In the example above we have that option placed on a spacebar as the largest and the most comfortable button on a keyboard, so player wouldn't drop his sight to press it. That's a compromise, but I hope in future we can implement a solution to enter objects using only mouse.

Arthusppp:
Given the above, how I implemented:
Case#1 splitting a troop into an empty slot -> the fhereoes2 moves a troop into the destination slot without the split window.

Case#2 splitting troops into an empty slot -> by default, the fheroes2 proposes the fast split option and by default, the Min button is displayed. When a player clicks the Min button, then the game will change the number of troops to split to 1 (one).
The fast split option here will propose almost half of the source troops (ie. when splitting 8, by default we will see 4; when splitting 11, by default we will see 5).
Before split:
image
Split window:
image
Default outcome:
image


Case#3 splitting troops into the same troop type -> by default, the fhereoes2 proposes to fast split into two slots. The default outcome is to have two slots with half of both slots. By default, the Min button is displayed (with the minor exception described below).

Example#1: if we split a slot with 15 Titans into a slot with 1 Titan, then the system proposes the fast split. The number of troops to split is 7. As a default result (fast split), a hero will have 2 slots with 8 Titans each.

Before split:
image
Split window:
image
Default outcome:
image



Example#2: if we split a slot with 1 Titan into a slot with 15 Titans, then the system proposes (displays) to split 0 Titans. This is counter-intuitive, however, as a result, a hero will have 2 slots with 8 Titans each. A player cannot change this number, only the fast split option is possible here. The fast split operates on the sum of both slots (ie, in this case, fast split into 4 slots will split into 4 slots with 4 Titans each).
Again, the Min button is displayed by default (as earlier). However, when slots are almost equal (ie the first slot contains 5 Titans, and the second contains 6 Titans), then the proposed amount of troops to move is 1, and the Max button is displayed by default (as there is no point in giving the Min button because there is already a min value proposed).

Before split:
image
Split window:
image
Default outcome:
image



Points to improve:
1. I believe the final solution should use a different UI approach. For example what has been done in Homm3: to display both slots with two separate numbers of units after the change in each slot.
2. When splitting a slot with 2 troops into an empty slot, I believe we should skip the dialog and simply do the split. 
3. When changing the fast split slots amount, I believe we should update the number of units to split. But... again, this is not elegant, we should consider a slightly different UI/UX approach.
4. I have not addressed the splitting between two slots with the same amount. I am not sure how to tackle when the first slot contains 5 units and the second slot contains 4 units. The current behavior is:
Before split:
image
Split window - here a player can click the Max button, in this example, the Max button proposes to move 7 peasants. As a result, a hero will have 1 peasant in the first slot and 15 peasants in the second slot.
image
Default outcome (none):
image

src/fheroes2/army/army_bar.cpp
Arthusppp: Resolved.

LeHerosInconnu:
Hello everyone,

> Given the above, how I implemented:
> Case#1 splitting a troop into an empty slot -> the fhereoes2 moves a troop into the destination slot without the split window.

So no change from the current operation, okay.

> Case#2 splitting troops into an empty slot -> by default, the fheroes2 proposes the fast split option and by default, the Min button is displayed. When a player clicks the Min button, then the game will change the number of troops to split to 1 (one).
> The fast split option here will propose almost half of the source troops (ie. when splitting 8, by default we will see 4; when splitting 11, by default we will see 5).
> Before split:
> image
> Split window:
> image
> Default outcome:
> image

Okay.

> Case#3 splitting troops into the same troop type -> by default, the fhereoes2 proposes to fast split into two slots. The default outcome is to have two slots with half of both slots. By default, the Min button is displayed (with the minor exception described below).
> 
> Example#1: if we split a slot with 15 Titans into a slot with 1 Titan, then the system proposes the fast split. The number of troops to split is 7. As a default result (fast split), a hero will have 2 slots with 8 Titans each.
> 
> Before split:
> image
> Split window:
> image
> Default outcome:
> image

I think that here the number displayed should match the result and that "8" should be displayed, indicating 2 troops (selected button) of 8 creatures each.
In case of several troops with different numbers of creatures the "average" number should be displayed.
For example, with:
"8-8-8-7" : "8" is displayed
"8-7-7-7" : "7" is displayed
"8-8-7-7" : "8" is displayed

> Example#2: if we split a slot with 1 Titan into a slot with 15 Titans, then the system proposes (displays) to split 0 Titans. This is counter-intuitive, however, as a result, a hero will have 2 slots with 8 Titans each.

This is not good, because, as you said, it's counter intuitive.
Here again, the number displayed should match the result and "8" should be displayed.

> A player cannot change this number, only the fast split option is possible here.

This is not good, the player can and should be able to change his mind, everything should work in the window.
(And adding dynamic titles would be a plus.)

> The fast split operates on the sum of both slots (ie, in this case, fast split into 4 slots will split into 4 slots with 4 Titans each).
> Again, the Min button is displayed by default (as earlier). However, when slots are almost equal (ie the first slot contains 5 Titans, and the second contains 6 Titans), then the proposed amount of troops to move is 1, and the Max button is displayed by default (as there is no point in giving the Min button because there is already a min value proposed).
> 
> Before split:
> image
> Split window:
> image
> Default outcome:
> image
> 
> Points to improve:
> 
> 1. I believe the final solution should use a different UI approach. For example what has been done in Homm3: to display both slots with two separate numbers of units after the change in each slot.

At the origin, it was not recommended to modify too much the original window. :)
But indeed, it would be a good idea to add, for example, slots with the small sprites of the creatures to show the final result in the hero's army (all types of creatures in the hero's army).
(And display the average number at the same time.)

> 2. When splitting a slot with 2 troops into an empty slot, I believe we should skip the dialog and simply do the split.

Yes, this would be better.

> 3. When changing the fast split slots amount, I believe we should update the number of units to split. But... again, this is not elegant, we should consider a slightly different UI/UX approach.

Like addind slots with the small sprites of the creatures to show the final result in the hero's army. :)

> 4. I have not addressed the splitting between two slots with the same amount. I am not sure how to tackle when the first slot contains 5 units and the second slot contains 4 units. The current behavior is:
>    Before split:
>    image
>    Split window - here a player can click the Max button, in this example, the Max button proposes to move 7 peasants. As a result, a hero will have 1 peasant in the first slot and 15 peasants in the second slot.
>    image
>    Default outcome (none):
>    image

I think that when both troops contain the same number of creatures (or almost), the "2" button should not be selected by default and the number of creatures should be "1".
By default, if nothing is changed, and the player clicks on the "OKAY" button one creature is transferred.
The player can always click on the "CANCEL" button if he does not want to change anything, both troops will remain equal in the presented case (or will remain for example 5-4).
If the player clicks on the "2" button and validates with the "OKAY" button, the troops also remain unchanged if they are equal.
If the player clicks on the "2" button and the troops are organized as "4-5", they are changed to "5-4" (if the initial troop of the action is the one with "4" creatures), as it is currently the case.

Arthusppp:
Hello @LeHerosInconnu, @Branikolog, @oleg-derevenetz , @ihhub and everyone!

So, to sum up everything, I have the following proposition:

### Case#1 splitting a troop into an empty slot
-> the fhereoes2 moves a troop into the destination slot without the split window.

### Case#2 splitting troops into an empty slot
-> by default, the fheroes2 proposes the fast split option and by default, the Min button is displayed. When a player clicks the Min button, then the game will change the number of troops to split to 1 (one).
The fast split option here will propose almost half of the source troops (ie. when splitting 8, by default we will see 4; when splitting 11, by default we will see 5).

(case#1 and case#2 are the same as earlier)

### Case#3 splitting troops into the same troop type
-> by default, the fhereoes2 proposes to fast split into two slots. The default outcome is to have two slots with half of both slots. By default, the Min button is displayed. A player can move from 1 to the number of source troops; by default, the game displays the number of troops that should be shifted from the bigger into the lower slot
Example#1: from 15 Titans to 2 Titans: by default, it is divided into two slots: source slot will have 9 Titans, destination slot will have 8 Titans; by default, the game displays "6 Titans to move" and the fast split into 2 slots option is selected; manually, a player can move from 1 to 15 Titans
Example#2: from 2 Titans to 15 Titans: by default, it is divided into two slots: source slot will have 9 Titans, destination slot will have 8 Titans; by default, the game displays "6 Titans to move" and the fast split into 2 slots option is selected; manually, a player can move 1 or 2 Titans. Please notice: in this case (source troops count is lower than the destination troops count), when a player clicks the Up or Down arrow, then the system will change to the maximum troops that can be moved (here: 2). Again, when the fast split into two is selected, the number of troops to move will be "6".

In my opinion, this is not ideal, but close enough and more intuitive. In the previous proposition, a player could move manually 8 troops when there was 1 troop in a source slot - I personally dislike this approach. 

Exception in Case#3: "small troop count difference" ("small" is 0 or 1)
-> by default, a single troop is proposed to move, a user can move all troops from the source slot, the Max button is the default, the fast split is possible (if a user clicks it)
Example#3: from 3 Titans to 3 Titans - the game proposes to move a single Titan (no fast split by default). A player can move 1 or 2 or 3 Titans. If the fast split into two slots is requested, then the result does not change anything (3-3).
Example#4: from 5 Titans to 6 Titans - the game proposes to move a single Titan (default outcome is 4-6). A player can move from 1 to 5 Titans. The fast split 

The fast split into 2 slots when "small troop count difference" - as proposed above:
* when splitting 4-4, then it does nothing (two slots with 4 troops each remain),
* when splitting 5-4, then it swaps number of troops -> result is 4-5,
* when splitting 4-5, then it swaps number of troops -> result is 5-4.

In addition: 
* splitting 1 troop into a slot with the same troop type will simply merge it without displaying the split window (this will cut off few corner cases of the above),
* splitting 2 troops into an empty slot will split into 1-1 without displaying the split window (as stated earlier, this is a QoL feature),
* the number of troops to split should be updated when the fast split option is selected (now it remains unchanged no matter which fast split option is selected); the number represents the number of troops in the first slot after fast split (ie splitting 17 troops into 4 slots, the outcome is 5-4-4-4, so "5" is displayed).

Branikolog:
> > 1. When splitting a slot with 2 troops into an empty slot, I believe we should skip the dialog and simply do the split.
> 
> Yes, this would be better.

I think it's a good idea.


> 3\. When changing the fast split slots amount, I believe we should update the number of units to split. But... again, this is not elegant, we should consider a slightly different UI/UX approach.

As I've talked above for implementing such logic we already have a discussion here: https://github.com/ihhub/fheroes2/discussions/2821#discussioncomment-384241

Players will be confused without changing the title of the dialog. I know, that current implementation works similar but we always offer player to split 1 unit (on default). So players are used to this "1" up in there and while initiating fast split button they consider, that their action changes this dialog behaviour. 
Split implementation of the build from this PR always shows different amount of creatures, so player (me in particular) doesn't know whether we move some creatures or just making some equal splitting....
Example:
image
Here we're trying to split 36 over 42 orcs. The window says, that we're trying to move 3 units, which isn't the case, as the default 50\50 fast split option enabled will bring 3 units FROM the target stack to the initial one.
The possible solution to this situation we've discussed here as well, https://github.com/ihhub/fheroes2/discussions/2821#discussioncomment-384241 , in a form of dynamic name of the dialog. Or, probably, making the name of the dialog semi-transparent (While clicking on fast split buttons) so the player will notice, that changes in a creature counter are correlated with fast split button being initiated.
Anyway, the box shows sometimes random numbers while splitting stack over stack:
image
image
Why do we have same 5 in these both dialogs?


> In my opinion, this is not ideal, but close enough and more intuitive. In the previous proposition, a player could move manually 8 troops when there was 1 troop in a source slot - I personally dislike this approach.

In the current implementation we can move  1 unit while splitting on a stack of 15:
image
(the same to adding 1 to 15) If we don't want to move 1 unit we press cancel. (Or can use fast split options.)
At the build from this PR we cannot add 1 unit to 15:
image
Only fast split options are making effect.
So previous implementation is more flexible.


> * when splitting 4-4, then it does nothing (two slots with 4 troops each remain),

I think it's not ok if we offer no differences on default splitting action. 
Also, If I want to add only single creature I need to press up and down arrows to disable fast split buttons. (Or press max and min button twice.
image


> * the number of troops to split should be updated when the fast split option is selected (now it remains unchanged no matter which fast split option is selected); the number represents the number of troops in the first slot after fast split (ie splitting 17 troops into 4 slots, the outcome is 5-4-4-4, so "5" is displayed).

For this we need to rework completely UI of this window.
I will be off for weekend. If you wish, @Arthusppp, you can contact me in Discord next week and we can discuss in real time this issue.

Arthusppp:
Hello @LeHerosInconnu , @Branikolog , @oleg-derevenetz , @ihhub  and everyone!

Gosh, this "good first issue" tackles so complicated matter :) 

Therefore, I propose the below in order to satisfy the #776 issue and the 0.9.8 milestone:

1. Splitting a troop into an empty slot 
-> this simply moves a troop into the destination slot.

2. Splitting two troops into an empty slot
-> this simply divides into two slots (one troop each)

3. Splitting troops into an empty slot
-> by default, the fheroes2 proposes the fast split option and by default, the Min button is displayed. When a player clicks the Min button, then the game will change the number of troops to split to 1 (one) and the fast split is not selected.
The fast split option here will propose almost half of the source troops (ie. when splitting 8, by default we will see 4; when splitting 11, by default we will see 5).

4. Splitting troops into the same troop type
-> by default, the fheroes2 proposes the fast split option and by default, the Min button is displayed. A player can move manually from 1 to max source troops count.
When a player clicks the Min button, then the game will change the number of troops to split to 1 (one) and the fast split becomes unselected.
Issue here: when splitting from a slot with 4 troops into a slot with 14 troops (5 troops to move to be equal), then by default, the number of troops to move shows "4" instead of "5" and the fast split button is selected. It is because the current logic of redistributing count forbids selecting more than the troop count of the source slot. This for sure will be resolved when we will get the final "fast split option" implementation.


The rest of the split logic remains unchanged and will be a subject of deeper discussion in:
- Discussion: https://github.com/ihhub/fheroes2/discussions/2821
- Issue: https://github.com/ihhub/fheroes2/issues/2406

Arthusppp:
@ihhub  If I may suggest adding this to the 0.9.8 milestone. This PR is not a final solution of the troops split. However, this PR covers the most common scenarios of the troops split. And this fixes #776.

ihhub:
> 
> 
> @ihhub If I may suggest adding this to the 0.9.8 milestone. This PR is not a final solution of the troops split. However, this PR covers the most common scenarios of the troops split. And this fixes #776.

Hi @Arthusppp , I'm okay to make it happen to 0.9.8 after @Branikolog approval.

@Branikolog please give your feedback again as I see that we haven't finished here.

src/fheroes2/army/army_bar.cpp
```diff
@@ -57,17 +57,33 @@ namespace
 
             Army::SwapTroops( troopFrom, troopTarget );
         }
+        else if ( !troopTarget.isValid() && troopFrom.GetCount() == 2 ) {
+            troopFrom.SetCount( 1 );
```
ihhub, line 61: Could we please add a comment here as the logic of this function becomes complex? Something like this:
```cpp
// Move one monster from a stack of 2 creatures to an empty slot.
```
so that any code reader will know what's going on here.

src/fheroes2/army/army_bar.cpp
```diff
@@ -57,17 +57,33 @@ namespace
...
 
             if ( isSameTroopType )
                 ++freeSlots;
 
             const uint32_t maxCount = saveLastTroop ? troopFrom.GetCount() - 1 : troopFrom.GetCount();
-            uint32_t redistributeCount = isSameTroopType ? 1 : troopFrom.GetCount() / 2;
+            uint32_t redistributeCount = troopFrom.GetCount() / 2;
+
+            // if the same type, then display how many troops should be moved from one slot to another (if redistributeCount > 1 then the Min button is displayed by
+            // default)
```
ihhub, line 74: Please put this comment inside if-body.

ihhub (changes requested):
Hi @Arthusppp , I left just few comments here. Could you please check them?

src/fheroes2/army/army_bar.cpp
```diff
@@ -57,17 +57,33 @@ namespace
...
+            if ( isSameTroopType ) {
+                redistributeCount
```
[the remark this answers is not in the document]
Arthusppp: Done.

src/fheroes2/army/army_bar.cpp
Arthusppp: Done.

src/fheroes2/army/army_bar.cpp
Arthusppp: Done.

Branikolog:
Hi, @Arthusppp & @ihhub.
Actually, I prefer the logic of splitting stack-over-stack implemented before, except instant distribution of stack with only two troops into two (1+1) and "MIN" default button for fast choice of separating a single troop from a stack, which @Arthusppp made within this PR already.

If we want to move 1 troop from one stack to another, while splitting two equal stacks, we need to press "MAX\MIN" button twice here.
image

Adding of 1 single troop is simple and clear in the cases we split stack-over-stack of any size.

If the majority of players and our team decides, that equalizing of two stacks (implemented by @Arthusppp within this PR) on default is the best and most common behaviour here so we should at least change "MAX\MIN" logic a bit, as it currently works on its own:
image
Here we split 6 over 9 Orcs. "MIN" button is default one and we can easily choose 1 troop and start rolling a scrollbar to add (2, 3, 4, ...etc) troops to the right stack, which is ok in this case.
But if we choose to split 9 Orcs over 6, "MAX" button is a default one, so we need to press it twice if we want to add just 1 troop.
image

The numbers in countbox are really confusing. Until the moment we can show the exact amount of transferring creatures and change the heading of the window dynamically, can we turn off this feature? It's completely not clear how it works in the current build.
image
image

So, summarizing my feedback:
1) I suggest leave splitting logic of stack with 2 troops instantly into 1+1. (Nice improvement of current split logic)

2) Default "MIN" button while splitting on an empty slot works ok now (after fix of @Arthusppp). 
(I remind you that the original issue #776 was exactly about this logic)

3) Leave changes of default equalizing (Stack-over-stack) for the future, when we implement a new split dialog with dynamically changing heading of the window and probably new type of buttons.

Arthusppp:
Hello @Branikolog 

I believe the latest commit solves the #776 issue. The changes I proposed:
1. When a player splits troops (>2) into an empty slot, by default the fheroes2 proposes half of them, and the Min button is a default button.
2. When a player splits two troops into an empty slot, by default the fheroes2 moves one troop into a destination slot.


The #776 resolution:
Starting from, I want to split goblins:
fheroes2_mV88FzT06o
And now splitting into an empty slot:

fheroes2_qhwXa4mgQB



Also, I have added the following minor feature: when a player changes the number of troops to move from 2 to 1, then the Max button is a default. (currently, 0.9.9, any manual up/down change does not affect Min/Max buttons). 

Example: I am splitting 17 goblins into the 3 goblins slot. Lowering the split number:
fheroes2_IjgukasA3y

Then once more pressing the Down button (the number lowers from 2 to 1, the Max button becomes default)

fheroes2_TG8yJvZkzb

And now I am pressing the Max button.
fheroes2_nQKsViK80b


To be clear: I have removed all previously proposed features. Everything else (not stated in this comment) works in the same way as in the 0.9.9.

ihhub:
Hi @Branikolog , could you please verify these changes if they're aligned with your proposed logic?

Branikolog:
Hi, @Arthusppp 
Found 1 issue:
While splitting 1 over 1 monster game shows a dialog window:
image
image
Which is useless. We should unite both creatures at once.

Arthusppp:
Hello @Branikolog 

Many thanks for your input. I have added the above. 

To sum up changes of this PR:

1. When a player splits troops (>2) into an empty slot, by default the fheroes2 proposes half of them, and the Min button is a default button.
2. When a player splits two troops into an empty slot, by default the fheroes2 moves one troop into a destination slot.
3. When a player splits one troop into one troop of the same type, then the fhereos2 moves the one troop into the target slot without the split window (as a result we have two troops in the target slot).

What do you think about expanding your last proposition (point 3) into the following: "When a player moves a single troop (count=1) into the same troop type slot (count=any), then this one monster is added to the target slot by default (without the split dialog)"?

Branikolog:
@Arthusppp 
Thank YOU for contributing to fheroes2 project. :)

>What do you think about expanding your last proposition (point 3) into the following: "When a player moves a single troop (count=1) into the same troop type slot (count=any), then this one monster is added to the target slot by default (without the split dialog)"?

Negative. We should have an ability to split both stacks into 2/3/4 equal.

>To sum up changes of this PR:
>
>When a player splits troops (>2) into an empty slot, by default the fheroes2 proposes half of them, and the Min button is a default button.
>When a player splits two troops into an empty slot, by default the fheroes2 moves one troop into a destination slot.
>When a player splits one troop into one troop of the same type, then the fhereos2 moves the one troop into the target slot without the split window (as a result we have two troops in the target slot).
>

Everything is good now.
I suppose, if we implement another splitting UI, in future some of the default choices could be changed here. But now everything seems proper and convenient.

Arthusppp:
Hi @Branikolog 

I have rejected this idea, but I have used a different explanation: what I proposed, was to introduce different behaviour of the split functionality that depends on if you split 1 troop to 23 troops or 23 troops to 1 troop. I generally dislike ideas like this.

@ihhub, please review the introduced code again, and I propose to add this to the 0.9.10 (current) milestone (as we will continue the "fast split logic discussion").

ihhub (approved):

ihhub:
@Arthusppp , thank you very much for these changes!

PR #4213, opened 2021-09-15
Fix some code smells introduced in recent commits

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thank you very much for keeping the code clean!

PR #4216, opened 2021-09-17
[WiP] Better cache locality in image Copy()

a1exsh:
First copy the image layer part, then the transform part, if any.
I find that this way the code is also slightly easier to read.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1092,17 +1092,16 @@ namespace fheroes2
         uint8_t * imageOutY = out.image() + offsetOutY;
         const uint8_t * imageInYEnd = imageInY + height * widthIn;
 
-        if ( out.singleLayer() ) {
-            for ( ; imageInY != imageInYEnd; imageInY += widthIn, imageOutY += widthOut ) {
-                memcpy( imageOutY, imageInY, static_cast<size_t>( width ) );
-            }
+        for ( ; imageInY != imageInYEnd; imageInY += widthIn, imageOutY += widthOut ) {
+            memcpy( imageOutY, imageInY, static_cast<size_t>( width ) );
         }
-        else {
+
+        if ( !out.singleLayer() ) {
             const uint8_t * transformInY = in.transform() + offsetInY;
             uint8_t * transformOutY = out.transform() + offsetOutY;
+            const uint8_t * transformInYEnd = transformInY + height * widthIn;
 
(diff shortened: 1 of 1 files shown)


a1exsh:
This is purely theoretical, of course, I have not tried to use profiler to check if there is any difference, or, for that matter — improvement.

If someone has a profiling setup, esp. for low-end hardware, it'd be nice to test it. There are other places in image.cpp that could see the same treatment, if it turns out to be beneficial.

felix642:
I haven't done any test either but looking at the code it is probably slower than what we already have. 

If we look at the new method we have to define a new variable transformInYEnd (which it not a big deal), but we have to iterate through the image twice instead of once. Which means that your method is probably slower than what we already have. 

I do not think it really changes the readability of the code (doesn't improve it nor worsen it from my point of view).

I'm sorry to say that but I do not think we should merge this pull request 😢 unless there's something I've missed? (And I would be really happy to know if so)

ihhub:
Hi @a1exsh , theoretically it might benefit but I have doubts about this. As you mentioned we have to do the profiling. The application has an ability to display FPS counter. This is the best indicator for performance. The most critical place for performance is hero's movement on speed 9 which must give 120 FPS at maximum. The bigger resolution you set the slower application becomes. So you can test with fullHD resolution before and after your changes for hero's movement on exactly same file save.

ihhub:
This change will only affect cases when a destination image is **not** single layer image. Display image is a single layer image. There are some cases when destination image is not single layer image. These cases can be divides into 2 groups:
- those which should be single layer image but we just forgot to mark them
- those which must remain double layer images

If we want to speed up some places without assumptions and relying on CPU cache (the size of CPU cache depends on CPUs as well) we might need to focus on the first group of images and just mark them single-layer images.

a1exsh:
@ihhub good points, thank you.  I've tested it on 1920x1080 resolution and I see no performance difference: both versions give me ~100 FPS, which is probably expected, according to your explanation about single layer images.

So, I'd say this is just something to keep in mind if we come to a point where need to squeeze extra bit of performance. ;)

a1exsh:
> but we have to iterate through the image twice instead of once. Which means that your method is probably slower than what we already have.

@felix642 well, we iterate through the first half once, and then once through the second half, so it's still the same number of iterations in total. ;)

ihhub:
@a1exsh , let's for now make this pull request as WIP till we find places where these changes improve performance.

ihhub:
Hi @a1exsh , shall we still keep this pull request or close it for now?

a1exsh:
Let's close it — doesn't look like we have performance issues at the moment.

PR #4218, opened 2021-09-18
Flatten HeroBase::CanCastSpell and fix result without error output

qarkai:
In previous variant spell's move points requirement wasn't checked if `res` argument was omitted.

### DIFF
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -412,35 +412,38 @@ double HeroBase::GetSpellcastStrength( const double armyLimit ) const
 
 bool HeroBase::CanCastSpell( const Spell & spell, std::string * res ) const
 {
-    if ( res ) {
-        res->clear();
-
-        if ( HaveSpellBook() ) {
-            if ( HaveSpell( spell ) ) {
-                if ( HaveSpellPoints( spell ) ) {
-                    if ( spell.MovePoint() <= move_point ) {
-                        return true;
-                    }
-                    else {
-                        *res += _( "Not enough move points." );
-                    }
-                }
-                else {
-                    *res += _( "That spell costs %{mana} mana. You only have %{point} mana, so you can't cast the spell." );
-                }
-            }
-            else {
-                *res += _( "The spell is not found." );
-            }
+    if ( !HaveSpellBook() ) {
+        if ( res ) {
+            *res = _( "Spell book is not present." );
         }
-        else {
-            *res += _( "Spell book is not present." );
-        }
-
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


src/fheroes2/heroes/heroes_base.cpp
```diff
@@ -412,35 +412,38 @@ double HeroBase::GetSpellcastStrength( const double armyLimit ) const
...
+            *res = _( "The spell is not found." );
         }
-        else {
-            *res += _( "Spell book is not present." );
+        return false;
+    }
+
+    if ( !HaveSpellPoints( spell ) ) {
+        if ( res ) {
+            *res = _( "That spell costs %{mana} mana. You only have %{point} mana, so you can't cast the spell." );
```
a1exsh, line 431: Not that original code didn't have this problem, but I'm pretty sure that "mana" is called "spell points" throughout the OG...

a1exsh (approved):
Nice readability improvement!

src/fheroes2/heroes/heroes_base.cpp
qarkai: Don't have OG to check but let's fix it in separate PR.

src/fheroes2/heroes/heroes_base.cpp
```diff
@@ -412,35 +412,38 @@ double HeroBase::GetSpellcastStrength( const double armyLimit ) const
...
+            *res = _( "Not enough move points." );
+        }
         return false;
     }
 
-    return HaveSpellBook() && HaveSpell( spell ) && HaveSpellPoints( spell );
+    if ( res ) {
+        res->clear();
+    }
+    return true;
```
felix642, line 446: Do we really need to discard the value of res if we return true? 
I do not think this method should clear this variable if it doesn't use it?
What do you think?

felix642 (commented):
Hi @qarkai! I really like how you changed that method. I left one small comment on which I wanted to have your opinion. 
Would you mind taking a look ? :)

src/fheroes2/heroes/heroes_base.cpp
qarkai: Why not? Method rewrites `res` in case of error. It will be inconsistent if we leave previous data in `res` otherwise.

src/fheroes2/heroes/heroes_base.cpp
ihhub: This is the same as in the original game.

src/fheroes2/heroes/heroes_base.cpp
felix642: Well, you're right, we don't really need to clear res since we're not using it when the method returns true, but clearing it allows us to stay consistent.

ihhub:
@qarkai , a bunch of thanks for these changes!

PR #4220, opened 2021-09-19
give 15 char limit when inputing player name in highscores

### LINKED ISSUE #4167: The player name (at least for the High Score purpose) should not exceed 15 chars
### Preliminary checks

### Platform

Windows

### Describe the bug

The High Score does not support player name longer than 15 chars. Please see row no 8 and 9:

image

### Save file

To end this game, get the Black Pearl artifact. Then the player name input is shown (at least if your score allows to place you on your High Score list).

[No_Name_21_0005.zip](https://github.com/ihhub/fheroes2/files/7111264/No_Name_21_0005.zip)

### Additional info

The OG accepts max 15 chars here.

eos428:
fix #4167

### DIFF
--- a/src/fheroes2/dialog/dialog.h
+++ b/src/fheroes2/dialog/dialog.h
@@ -103,7 +103,7 @@ namespace Dialog
     int ResourceInfo( const std::string &, const std::string &, const Funds &, int buttons = Dialog::OK );
     int SelectSkillFromArena( void );
     bool SelectCount( const std::string &, u32 min, u32 max, u32 & res, int step = 1 );
-    bool InputString( const std::string & header, std::string & result, const std::string & title = std::string(), const size_t charLimit = 99 );
+    bool InputString( const std::string & header, std::string & result, const std::string & title = std::string(), const size_t charLimit = 0 );
     Troop RecruitMonster( const Monster &, u32 available, bool );
     void DwellingInfo( const Monster &, u32 available );
     bool SetGuardian( Heroes &, Troop &, CapturedObject &, bool readonly );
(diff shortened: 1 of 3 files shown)


src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -213,7 +213,7 @@ bool Dialog::SelectCount( const std::string & header, u32 min, u32 max, u32 & cu
     return result == Dialog::OK;
 }
 
-bool Dialog::InputString( const std::string & header, std::string & res, const std::string & title )
+bool Dialog::InputString( const std::string & header, std::string & res, const std::string & title, const size_t charLimit )
```
ihhub, line 216: [unrelated to this change but] Could we please change lines 225-226 into:
```cpp
res.clear();
```
?
The same check is done in `clear` call actually.

src/fheroes2/dialog/dialog.h
```diff
@@ -103,7 +103,7 @@ namespace Dialog
     int ResourceInfo( const std::string &, const std::string &, const Funds &, int buttons = Dialog::OK );
     int SelectSkillFromArena( void );
     bool SelectCount( const std::string &, u32 min, u32 max, u32 & res, int step = 1 );
-    bool InputString( const std::string &, std::string &, const std::string & title = std::string() );
+    bool InputString( const std::string & header, std::string & result, const std::string & title = std::string(), const size_t charLimit = 99 );
```
ihhub, line 106: What if we don't want to have a limit? Shall we set this value to **0** by default and modify the condition
```cpp
if ( charLimit > res.size() || le.KeyValue() == KeySym::KEY_BACKSPACE )
```
into
```cpp
if ( charLimit == 0 || charLimit > res.size() || le.KeyValue() == KeySym::KEY_BACKSPACE )
```
?

ihhub (changes requested):
Hi @eos428 , I left few comments here. Would you mind to take a look?

src/fheroes2/dialog/dialog.h
eos428: Yeah sure, this should work too. Originally I wanted -1 but I forgot `size_t` is unsigned :D

ihhub:
@eos428 , thank you so much for this fix!

PR #4235, opened 2021-09-20
Optimize artifacts income calculation in Kingdom::GetIncome

qarkai:
Iterate over heroes just once.


a1exsh (approved):
Nice!

a1exsh:
@qarkai some builds are failing with `'artifacts' uses undefined class 'std::array<int,10>'`.

Maybe `#include <array>` is missing?

ihhub:
@qarkai , thank you very much for this small improvement!

PR #4248, opened 2021-09-22
Changed shipwreck message

### LINKED ISSUE #4246: Adventure map, Shipwreck, the text should be adapted when the hero does not fight the ghosts
### Discussed in https://github.com/ihhub/fheroes2/discussions/4239

<div type='discussions-op-text'>

<sup>Originally posted by **LeHerosInconnu** September 21, 2021</sup>
When a hero visits a Shipwreck where the ghosts have already been defeated, the text should be adapted because the hero does not fight the ghosts.
It is the same in the original game.

The beginning of the text "Upon defeating the Ghosts" should simply be removed.

In fheroes2:

No combat with ghosts 01

Here is the save file:
[Text shipwreck 01.zip](https://github.com/ihhub/fheroes2/files/7202831/Text.shipwreck.01.zip)

Version 0.9.7 4210</div>

felix642:
Fix #4246 

Removed "Upon defeating the ghosts" from the message and also change the other messages that were similar.

### DIFF
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -1516,17 +1516,17 @@ void ActionToPoorMoraleObject( Heroes & hero, const MP2::MapObjectType objectTyp
     switch ( objectType ) {
     case MP2::OBJ_GRAVEYARD:
         ask = _( "You tentatively approach the burial ground of ancient warriors. Do you want to search the graves?" );
-        msg = _( "Upon defeating the Zombies you spend several hours searching the graves and find nothing. Such a despicable act reduces your army's morale." );
+        msg = _( "You spend several hours searching the graves and find nothing. Such a despicable act reduces your army's morale." );
         win = _( "Upon defeating the Zombies you search the graves and find something!" );
         break;
     case MP2::OBJ_SHIPWRECK:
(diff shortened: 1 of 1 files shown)


a1exsh (approved):
:+1:

Arthusppp:
Should this change also update all *.po files?

ihhub:
> 
> 
> Should this change also update all *.po files?

Hi @Arthusppp , please wait till this pull request will be merged and then you can update po files.

ihhub:
Hi @felix642 , everything looks fine. The one thing which we have to do is to change *.po files located at `files/lang` directory. We have 12 files there and we need to modify the sentences there as well otherwise translations (even with extra words) won't appear at all.

felix642:
Hi @ihhub, I saw that you updated the different .po files. Thank you.. 
We also have to update the different translations to match the new sentences. What do you think is the best way to do so?

ihhub:
> 
> 
> Hi @ihhub, I saw that you updated the different .po files. Thank you.. We also have to update the different translations to match the new sentences. What do you think is the best way to do so?

Hi @felix642 , the only problem which we have to fix is a fact that after our changes we have 2 identical strings and translation compilation fails. I think we need to ask help of other people to help with translations here.

PR #4249, opened 2021-09-23
Unselect initial town in the Town Portal spell list

### LINKED ISSUE #843: Click over selected castle causes immediate teleportation in "Town Portal" spell list
If player makes simple click over "yellow" castle spell activates.
TP

a1exsh:
Fix #843

### DIFF
--- a/src/fheroes2/heroes/heroes_spell.cpp
+++ b/src/fheroes2/heroes/heroes_spell.cpp
@@ -526,6 +526,7 @@ bool ActionSpellTownPortal( Heroes & hero )
     listbox.SetAreaMaxItems( 5 );
     listbox.SetAreaItems( fheroes2::Rect( area.x + 6, area.y + 49, 250, 160 ) );
     listbox.SetListContent( castles );
+    listbox.Unselect();
     listbox.RedrawBackground( area.getPosition() );
     listbox.Redraw();
 


a1exsh:
I have this one-line change, but wondering now what's the fastest way to test it? :sweat_smile: 

There doesn't seem to be a way to select spells in the OG editor (for a hero, a town or a shrine).  Does turning on some debug level gives all possible spells to the human player, for example?

LeHerosInconnu:
Hello @a1exsh,

> I have this one-line change, but wondering now what's the fastest way to test it? 
> 
> There doesn't seem to be a way to select spells in the OG editor (for a hero, a town or a shrine). Does turning on some debug level gives all possible spells to the human player, for example?

It is not possible to do this in the scenario editor of the original game.
One way around it is to place a Spell Scroll with the desired spell on the adventure map, or to place many castles with a fully built Mage Guild.
I don't know if there is a debug level in fheroes2 for that.
In the meantime here is a test scenario file with a spell scroll:
[0 9956d.zip](https://github.com/ihhub/fheroes2/files/7216991/0.9956d.zip)

vincent-grosbois:
> 
> 
> I have this one-line change, but wondering now what's the fastest way to test it? 😅
> 
> There doesn't seem to be a way to select spells in the OG editor (for a hero, a town or a shrine). Does turning on some debug level gives all possible spells to the human player, for example?

Usually I just run the game in debug mode, put a breakpoint somewhere and edit whatever data structure so it contains all spells
What you can do otherwise is just to add a few lines of codes when you compile such that whenever a spellbook object is created it adds the town portal spell, etc

a1exsh:
Spell Scroll works the best — thank you, this way one can also test it in the OG.

I've tested the change and it works as I expected: the initial town is not selected, so clicking once doesn't immediately teleport the hero.  The issue should be fixed by this.

ihhub:
Hi @a1exsh , I think the original issue is based on a fact that we allow single mouse clicking and interpret 2 independent clicks as a double click (due to design). What we really need is to implementation of mouse double click support. I think the proposed change doesn't solve anything.
@Branikolog , please correct me if I'm wrong.

a1exsh:
> Hi @a1exsh , I think the original issue is based on a fact that we allow single mouse clicking and interpret 2 independent clicks as a double click (due to design).

It works the same way in OG and I think it's fine.

@ihhub Didn't you write this on the issue ticket? ;)
> I think the correct way would be to do not highlight first town in a list at the beginning of the spell.
https://github.com/ihhub/fheroes2/issues/843#issuecomment-646993374

It does solve the issue and I think it's a reasonable approach to this problem..

ihhub:
Oops, my bad🙂

Branikolog:
Hi, @ihhub & @a1exsh.
In the OG we need to press OK button to initiate the spell. Currently in fheroes2 we simply need to click over selected (yellow) castle name. I've already talked about it in https://github.com/ihhub/fheroes2/issues/843#issuecomment-893361546
We can always misclick if it was previously selected. So no town selected on initiating the spell helps a bit (with the first town in a list) but not solves the problem.

ihhub:
@a1exsh , thank you very much for this change!

PR #4251, opened 2021-09-24
Fix scrollbar rendering in select lists

### LINKED ISSUE #1371: Combat screen, logs list, the border of the buttons and the scroll bar does not have a yellow border on the left
The border of the buttons and the scroll bar does not have a yellow border on the left (1 pixel wide).

Combat screen, logs list, yellow border 01

a1exsh:
This fixes rendering of the list in Town Portal spell and Select
Hero/Skill/Army/Artifact/etc. lists in Battle-Only setup.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -39,26 +39,26 @@ public:
         , ok( false )
     {
         RedrawBackground( rt.getPosition() );
-        SetScrollButtonUp( ICN::LISTBOX, 3, 4, fheroes2::Point( rt.x + rt.width - 24, rt.y + 25 ) );
-        SetScrollButtonDn( ICN::LISTBOX, 5, 6, fheroes2::Point( rt.x + rt.width - 24, rt.y + rt.height - 55 ) );
+        SetScrollButtonUp( ICN::LISTBOX, 3, 4, fheroes2::Point( rt.x + rt.width - 25, rt.y + 25 ) );
+        SetScrollButtonDn( ICN::LISTBOX, 5, 6, fheroes2::Point( rt.x + rt.width - 25, rt.y + rt.height - 55 ) );
 
-        SetScrollBar( fheroes2::AGG::GetICN( ICN::LISTBOX, 10 ), fheroes2::Rect( rt.x + rt.width - 20, rt.y + 48, 14, rt.height - 107 ) );
+        SetScrollBar( fheroes2::AGG::GetICN( ICN::LISTBOX, 10 ), fheroes2::Rect( rt.x + rt.width - 21, rt.y + 48, 14, rt.height - 107 ) );
         SetAreaMaxItems( 5 );
         SetAreaItems( fheroes2::Rect( rt.x + 10, rt.y + 30, rt.width - 30, rt.height - 70 ) );
     }
 
     void RedrawBackground( const fheroes2::Point & dst ) override
     {
-        Dialog::FrameBorder::RenderOther( fheroes2::AGG::GetICN( ICN::CELLWIN, 1 ), fheroes2::Rect( dst.x, dst.y + 25, rtAreaItems.width + 5, rtAreaItems.height + 10 ) );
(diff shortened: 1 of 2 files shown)


a1exsh:
Screenshots:
image
image

a1exsh:
One of the builds have failed with `fatal error C1083: Cannot open include file: 'SDL.h': No such file or directory`.
This is due to failed installation of dependencies:
```
Exception calling "DownloadFile" with "2" argument(s): "Unable to connect to the remote server"
At line:1 char:1
+ (New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.or ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
...
```
Which happily reported at the end:
```
File not found - SDL2_image.dll
0 File(s) copied
File not found - SDL2_image.lib
0 File(s) copied
File not found - zlib1.dll
0 File(s) copied
2 File(s) copied
2 File(s) copied
2 File(s) copied
"SUCCESS! Installation is completed"
```
Someone who understands Windows scripting should probably have a look. ;)

LeHerosInconnu:
Hello @a1exsh,

I think it would be nice to add a yellow border to the left of the scrollbar in this window too.
You will also have to add a "white" border to the left of the yellow border to close the inner frame of the cities.

Yellow border missing left scrollbar 01

Yellow border missing left scrollbar 02

ihhub:
> 
> 
> One of the builds has failed with `fatal error C1083: Cannot open include file: 'SDL.h': No such file or directory`.
> This is due to failed installation of dependencies:
> 
> ```
> Exception calling "DownloadFile" with "2" argument(s): "Unable to connect to the remote server"
> At line:1 char:1
> + (New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.or ...
> + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
>     + FullyQualifiedErrorId : WebException
> ...
> ```
> 
> Which happily reported at the end:
> 
> ```
> File not found - SDL2_image.dll
> 0 File(s) copied
> File not found - SDL2_image.lib
> 0 File(s) copied
> File not found - zlib1.dll
> 0 File(s) copied
> 2 File(s) copied
> 2 File(s) copied
> 2 File(s) copied
> "SUCCESS! Installation is completed"
> ```
> 
> Someone who understands Windows scripting should probably have a look. ;)

In this case you can just ask me to restart the build :)

a1exsh:
> In this case you can just ask me to restart the build :)

I mean it would make more sense if failed dependencies didn't report "SUCCESS" and stopped the build at that stage already. ;)

a1exsh:
@LeHerosInconnu 
> I think it would be nice to add a yellow border to the left of the scrollbar in this window too.

Let me give it a try.

a1exsh:
Hm, actually I don't think that we need the _yellow_ border there.  What we need is rendering like in the OG (is it only the editor where these are used?)
image

There is no yellow border, but there is 2px thick border: 1px comes from the list box area, and another 1px from the scrollbar.  We just need to position them precisely to get the same look.

a1exsh:
This is how it should be done, IMO:
image

Maybe window should be extended a little, it was already looking a bit to narrow for the list.

a1exsh:
After extending the window 280 -> 290px:
image

a1exsh:
The only other place where the `LISTBOX` icons are used (from the code) seems to be battle-only dialog:
image

Here the rendering of the scrollbar is also not ideal, I would say.

It also looks a bit strange due to the double border, but that's a separate issue.

a1exsh:
That's also easy to fix:
image

Still not ideal, but better than before. ;)

a1exsh:
I've decided to scope it differently, as settings dialog should probably use a different scrollbar in the first place, and the battle log scrollbar looks out of place: I'll see what could be done there, but separately from this PR.

ihhub:
@a1exsh , a bunch for thanks for this change!

PR #4259, opened 2021-09-25
Fix cmake library include directories for engine target

PierceLBrooks:
This also introduces the package location "_ROOT" suffix hint policy to allow for proper discovery behavior when libraries are manually specified.


ihhub:
@PierceLBrooks , thank you very much for your contribution to the project!

PR #4261, opened 2021-09-26
Add Campaign high scores and fix inability to add new entries with similar scores

### LINKED ISSUE #4166: The High Score list - add Campaign
### Preliminary checks

Add the Campaign High Score list.

This issue is a continuation of the #954. The #954 added only single map support. Also, the #954 contains many details on how the list is made in the OG (ie. position of elements on a screen, campaign duration and its rating monster).

image

Implement the feature as in the OG and in the #954.

### Additional info

_No response_

### LINKED ISSUE #5165: High score does not list two equal high scores like the original game
### Preliminary checks

### Platform

Windows

### Describe the bug

Unless a high score is unique, it will not be listed (Both versions' player is called Unknown Hero):
Original game:
Original

fheroes2 where two similar scores should show:
fheroes2

I imagine we would want to list every high score created, even thinking of a case where you play the same map years in between and end up with the exact same score.

Also note that currently fheroes2 does not highlight new scores in yellow like OG. I've looked into solving this separately.

### Save file

I can provide a save file for fheroes2. For original game just press 911 to win immediately (by cheating).

### Additional info

_No response_

eos428:
fix #4166, fix #5165

In summary:
- Rename hgs_t to HighScoreStandardData, also create HighScoreCampaignData
- Rename HGSData to HighScoreDataContainer which now also contains a vector of HighScoreCampaignData
- Remove display logic from HGSData and place it inside anonymous namespace in game_highscores.cpp
- Create highscores.cpp and highscores.h for pure high-score logic, game's highscore logic will be in game_highscores.cpp
- Alot of code cleanup here and there, probably there's still more we can do here
- Set campaign names for PoL campaigns and add a getter for the campaign name, this one is relatively minor

I'll need some help on the correct sprite index for high score campaign's background. I will make a comment on the specific line needing the changes. And then regarding about the change in file formatting which we'd look to make closer to a save file's, I'd put that in another PR after this to avoid overscoping. This PR is already a lot more than what I expected in the first place.

Also, this is totally not a good first issue :P


src/fheroes2/game/game_highscores.cpp
```diff
@@ -49,406 +52,196 @@
...
-    hdata.setbigendian( true );
-    hdata << static_cast<u16>( HGS_ID ) << list;
-    if ( hdata.fail() || !hdata.write( fn ) )
-        return false;
+        fheroes2::Display & display = fheroes2::Display::instance();
 
-    return true;
-}
+        // image background
+        fheroes2::Blit( fheroes2::AGG::GetICN( ICN::HSBKG, 4 ), display, ox, oy );
```
eos428, line 113: This is the wrong background issue I wrote about in the PR summary. Need help to get the correct sprite index on this one.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+            }
+        }
+
+        std::pair<size_t, Monster::monster_t> lastData = monsterDays.back();
+
+        if ( dayCount >= lastData.first )
+            return Monster( lastData.second );
+
+        Monster::monster_t monster = Monster::PEASANT;
+        for ( size_t i = 0; i < monsterDays.size(); ++i ) {
```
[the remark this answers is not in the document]
vincent-grosbois: I'm sure you can use somethin like std::lower_bound instead to have the result in O(log(n)) instead of O(n)
it may not matter though

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+#include <string>
+
+#include "highscores.h"
+#include "zzlib.h"
+
+namespace
+{
+    const uint16_t HGS_ID1 = 0xF1F3;
+    const uint16_t HGS_ID2 = 0xF1F4;
+    const uint8_t HGS_MAX = 10;
```
vincent-grosbois, line 32: would suggest renaming this or putting a comment on what it is

eos428: Yeah, I suppose we could rename these, the IDs probably still need a comment in addition to the renaming though haha

ihhub, line 32: This variable must be `size_t` type to avoid implicit conversion.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+    bool DaysSort( const HighScore::HighScoreCampaignData & h1, const HighScore::HighScoreCampaignData & h2 )
+    {
+        return h1._days < h2._days;
+    }
+}
+
+namespace HighScore
+{
+    HighScoreDataContainer & HighScoreDataContainer::Get()
```
ihhub, line 81: Do we need this method? It's used only for singleton classes while in our cases we can just create a normal instance of this class as it's used only in one place after all.

src/fheroes2/game/highscores.h
```diff
@@ -0,0 +1,101 @@
...
+
+#include <string>
+
+#include "monster.h"
+
+class StreamBase;
+
+namespace HighScore
+{
+    struct HighScoreStandardData
```
ihhub, line 32: The only difference between `HighScoreStandardData` and `HighScoreCampaignData` structures is presence of `_rating` member in the last structure. To reduce code duplication we should at least derive `HighScoreCampaignData` from `HighScoreStandardData` structure.
Moreover, we can calculate rating during runtime (not a big deal to call small function 10 times). In this case we'll need only one structure to hold all needed data. What do you think?

src/fheroes2/game/highscores.h
```diff
@@ -0,0 +1,101 @@
...
+            , _localTime( 0 )
+            , _days( 0 )
+            , _rating( 0 )
+        {}
+
+        bool operator==( const HighScoreStandardData & other ) const;
+
+        std::string _player;
+        std::string _scenarioName;
+        uint32_t _localTime;
```
ihhub, line 46: I recommend to change this type to `uint64_t` to avoid conversion warnings.

src/fheroes2/game/highscores.h
```diff
@@ -0,0 +1,101 @@
...
+        bool Save( const std::string & fileName ) const;
+        void RegisterScoreStandard( const std::string & playerName, const std::string & scenarioName, const uint32_t days, const uint32_t rating );
+        void RegisterScoreCampaign( const std::string & playerName, const std::string & campaignName, const uint32_t days );
+
+        const std::vector<HighScoreStandardData> & GetHighScoresStandard()
+        {
+            return _highScoresStandard;
+        }
+
+        const std::vector<HighScoreCampaignData> & GetHighScoresCampaign()
```
ihhub, line 86: These 2 methods could be `const`.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+        for ( size_t i = 0; i < monsterDays.size(); ++i ) {
+            if ( dayCount <= monsterDays[i].first ) {
+                monster = monsterDays[i].second;
+                break;
+            }
+        }
+
+        return Monster( monster );
+    }
+}
```
ihhub, line 296: Please add an empty line at the end of this file.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+                else if ( monster == Monster::BATTLE_DWARF ) {
+                    dayIncrementCount = 200;
+                }
+                // >= 5801
+                else if ( monster == Monster::PEASANT ) {
+                    dayIncrementCount = 1;
+                }
+
+                daySoFar += dayIncrementCount;
+                monsterDays.emplace_back( std::make_pair( daySoFar, monstersInRanking[i] ) );
```
ihhub, line 277: Please try to remove `std::make_pair` here and emplace_back should work properly without it.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        return false;
+    }
+
+    bool HighScoreDataContainer::Save( const std::string & fileName ) const
+    {
+        ZStreamFile hdata;
+        hdata.setbigendian( true );
+        hdata << static_cast<uint16_t>( HGS_ID2 ) << _highScoresStandard << _highScoresCampaign;
```
ihhub, line 144: Please remove static_cast call here. It'd redundant based on a fact that `HGS_ID2` is uint16_t type.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+        if ( hdata.fail() || !hdata.write( fileName ) )
+            return false;
+
+        return true;
```
ihhub, line 148: We could combine these lines into
```cpp
return !hdata.fail() && hdata.write( fileName ) );
```

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    void HighScoreDataContainer::RegisterScoreStandard( const std::string & playerName, const std::string & scenarioName, const uint32_t days, const uint32_t rating )
+    {
+        HighScoreStandardData highScore;
+        std::vector<HighScoreStandardData> & highScores = _highScoresStandard;
```
ihhub, line 154: Why do we need this extra variable if we can use `_highScoresStandard` directly?

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+        highScores.emplace_back( highScore );
+        std::sort( highScores.begin(), highScores.end(), RatingSort );
+        if ( highScores.size() > HGS_MAX )
+            highScores.resize( HGS_MAX );
+    }
+
+    void HighScoreDataContainer::RegisterScoreCampaign( const std::string & playerName, const std::string & campaignName, const uint32_t days )
+    {
+        HighScoreCampaignData highScore;
+        std::vector<HighScoreCampaignData> & highScores = _highScoresCampaign;
```
ihhub, line 175: Why do we need this extra variable if we can use `_highScoresCampaign` directly?

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+        ZStreamFile hdata;
+        hdata.setbigendian( true );
+        hdata << static_cast<uint16_t>( HGS_ID2 ) << _highScoresStandard << _highScoresCampaign;
+        if ( hdata.fail() || !hdata.write( fileName ) )
+            return false;
+
+        return true;
+    }
+
+    void HighScoreDataContainer::RegisterScoreStandard( const std::string & playerName, const std::string & scenarioName, const uint32_t days, const uint32_t rating )
```
ihhub, line 151: if we're going to use the same structure we might need to create another private function which will be called from this method and the method below to reduce code duplication.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+                else if ( monster == Monster::GREATER_DRUID ) {
+                    ratingIncrementCount = 3;
+                }
+                // >= 228
+                else if ( monster == Monster::BLACK_DRAGON ) {
+                    ratingIncrementCount = 1;
+                }
+
+                ratingSoFar += ratingIncrementCount;
+                monsterRatings.emplace_back( std::make_pair( ratingSoFar, monstersInRanking[i] ) );
```
ihhub, line 221: We can remove std::make_pair call from this line.

src/fheroes2/game/highscores.cpp
```diff
@@ -0,0 +1,296 @@
...
+            }
+        }
+
+        std::pair<size_t, Monster::monster_t> lastData = monsterRatings.back();
```
ihhub, line 225: We're making a copy of std::pair here. Please use const reference here.

ihhub (changes requested):
Hi @eos428 , I left several comments here. Could you please take a look at them and see that my suggestions are useful?

src/fheroes2/game/highscores.cpp
eos428: At the moment this is used in both HighScoresStandard() and HighScoresCampaign(). I don't wish to make a new instance every time these two functions are called since we have to initialize the pairs for monster ratings and days,

src/fheroes2/game/highscores.h
eos428: Yes I agree with calculating rating during runtime. In that case, I think we need to store the game difficulty too?

src/fheroes2/game/highscores.cpp
eos428: Sorry, personal coding style in c# when I use vscode. The idea is to minimize direct references to class' variables by creating a local variable. But no worries I will change this one and the one above.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -49,406 +52,196 @@
...
-    u32 rating;
-};
-
-StreamBase & operator<<( StreamBase & msg, const hgs_t & hgs )
+namespace
 {
-    return msg << hgs.player << hgs.land << hgs.localtime << hgs.days << hgs.rating;
-}
+    const std::string highScoreDataPath = "fheroes2.hgs";
+    static uint32_t monsterAnimationFrameId = 0;
```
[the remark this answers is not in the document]
eos428: Yeah, I guess we can create local variables in both HighScoresStandard() and HighScoresCampaign()

src/fheroes2/game/highscores.cpp
eos428: Yes I think we should create a template function for std::vector<T> where T is either high score standard or high score campaign

src/fheroes2/game/highscores.h
eos428: Also while the variables are almost the same, I think we should still make the distinction between the string's variable name. One is campaignName and another is scenarioName.

src/fheroes2/game/highscores.h
eos428: I took a look and it seems we don't have a method to serialize or deserialize uint64_t yet. Should we do this the serialization thing in another PR?

src/fheroes2/game/highscores.cpp
eos428: On second thought I think I get what you mean, this should be inside the anonymous namespace in `game_highscores.cpp`

src/fheroes2/game/game_highscores.cpp
eos428: I see... thanks for the info. There's an issue where the button is still `Campaign` when we are viewing campaign high scores.  I thought it was the background but maybe the problem is somewhere else?

src/fheroes2/game/game_highscores.cpp
```diff
@@ -457,45 +249,46 @@ fheroes2::GameMode Game::HighScores()
...
 
     fheroes2::Display & display = fheroes2::Display::instance();
     const fheroes2::Point top( ( display.width() - back.width() ) / 2, ( display.height() - back.height() ) / 2 );
     const fheroes2::StandardWindow border( display.DEFAULT_WIDTH, display.DEFAULT_HEIGHT );
 
-    hgs.RedrawList( top.x, top.y );
+    RedrawHighScoresCampaign( top.x, top.y, monsterAnimationFrameId );
 
-    fheroes2::Button buttonCampain( top.x + 8, top.y + 315, ICN::HISCORE, 0, 1 );
+    fheroes2::Button buttonStandard( top.x + 8, top.y + 315, ICN::HISCORE, 2, 1 );
```
undef21, line 268: Sprites of button "Standard"  have indices 2, 3.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -49,406 +52,195 @@
...
+
+    LocalEvent & le = LocalEvent::Get();
+
+    // highscores loop
+    while ( le.HandleEvents() ) {
+        // key code info
+        if ( Settings::Get().Debug() == 0x12 && le.KeyPress() )
+            Dialog::Message( "Key Press:", std::to_string( le.KeyValue() ), Font::SMALL, Dialog::OK );
+        if ( buttonCampaign.isEnabled() ) {
+            le.MousePressLeft( buttonCampaign.area() ) ? buttonCampaign.drawOnPress() : buttonCampaign.drawOnRelease();
```
undef21, line 214: I think problem with "Campaign" button here...

src/fheroes2/game/game_highscores.cpp
```diff
@@ -508,23 +301,25 @@ fheroes2::GameMode Game::HighScores()
         // key code info
         if ( Settings::Get().Debug() == 0x12 && le.KeyPress() )
             Dialog::Message( "Key Press:", std::to_string( le.KeyValue() ), Font::SMALL, Dialog::OK );
-        if ( buttonCampain.isEnabled() ) {
-            le.MousePressLeft( buttonCampain.area() ) ? buttonCampain.drawOnPress() : buttonCampain.drawOnRelease();
+        if ( buttonStandard.isEnabled() ) {
+            le.MousePressLeft( buttonStandard.area() ) ? buttonStandard.drawOnPress() : buttonStandard.drawOnRelease();
```
undef21, line 305: ...and here. Comment this lines and you can see.

eos428: This doesn't seem to solve the issue... It seems the background image in  `fheroes2::Blit( fheroes2::AGG::GetICN( ICN::HSBKG, 0 ), display, top.x, top.y );` in `RedrawHighScoresCampaign()` also includes the `Campaign` button, which is not what we are looking for and I think that's the root cause of the wrong button issue.

Maybe there's another high score background sprite that we don't know yet?

undef21: This is not quite an original resource. IСN was converted from original format. It's same as game screenshot.

eos428: Sorry for late response.

IMO the best solution would be to draw the background just once. But there's the problem where the background gets dirty due to the monster sprite.

Alternatively we could draw the campaign button again on top of the background, but it seems a bit hacky to say the least.

ihhub:
Hi @eos428 , **game_highscores.cpp** contains 2 functions:
- **fheroes2::GameMode Game::HighScoresStandard()**
- **fheroes2::GameMode Game::HighScoresCampaign()**

from what I can see these functions are extremely identical except few minor things. We have to reduce code duplication here and we can do this way: add a new function in unnamed namespace within this file which will accept a single parameter indicating whether it is a campaign or a normal scenario (**const bool isCampaign** is sufficient). Then the remaining logic inside function will depend on this variable. Some variables will just have a common name.

What do you think?

eos428:
Hi @ihhub, I think that's something we can try. Hopefully the only param we have to pass is just `isCampaign`.

zenseii:
Hi guys! What is currently remaining for this PR to be merged? It would seem that working on issues such as these https://github.com/ihhub/fheroes2/issues/5269, https://github.com/ihhub/fheroes2/issues/5239, https://github.com/ihhub/fheroes2/issues/5165 would have to wait until this PR has been merged, right?

ihhub (approved):

ihhub:
Hi @oleg-derevenetz , I completed this pull request so I marked it approved. Please take a look for some obvious mistakes if any.

src/fheroes2/game/game_highscores.cpp
```diff
@@ -50,422 +49,102 @@
...
-    ScoreRegistry( "Barock", "Wastelands", 240, 60 );
-}
-
-fheroes2::GameMode Game::HighScores()
+fheroes2::GameMode Game::DisplayHighScores( const bool isCampaign )
 {
 #ifdef WITH_DEBUG
     if ( IS_DEVEL() && world.CountDay() ) {
-        std::string msg = std::string( "Developer mode, not save! \n \n Your result: " ) + std::to_string( GetGameOverScores() );
+        std::string msg = std::string( "Developer mode, not save! \n \n Your result: " ) + std::to_string( Campaign::CampaignSaveData::Get().getDaysPassed() );
```
oleg-derevenetz, line 147: I suppose we need to use the `Game::GetGameOverScores()` here in case of a regular single player game.

oleg-derevenetz (changes requested):
Hi @ihhub I left only one obvious notice, please take a look when you have time :)

ihhub:
@eos428 , thank you very much for these changes!

PR #4262, opened 2021-09-26
Fix #4260

### LINKED ISSUE #4260: Combat: AI Hydra logic
### Preliminary checks

### Platform

Linux

### Describe the bug

When attacking with Hydras, AI doesn't always choose the optimal spot.

Example:

All Wolves skipped and stay where they are, Hydra's turn:

Hydra attacks top left Wolf. Instead, in this situation I think AI should have moved Hydra between the two Wolves in the topmost row and attacked both.

When evaluating possible moves, AI should take into account that a Hydra attacks all neighboring creatures and select the hex where it can maximize damage to the whole enemy army.

### Save file

[savefile.zip](https://github.com/ihhub/fheroes2/files/7230245/savefile.zip)

### Additional info

I have replayed this battle a couple of times and noticed this kind of behavior in multiple situations. Game was set to Normal difficulty.

vincent-grosbois:


src/fheroes2/battle/battle_board.cpp
```diff
@@ -523,6 +523,45 @@ int32_t Battle::Board::OptimalAttackValue( const Unit & attacker, const Unit & t
...
         return target.GetScoreQuality( attacker ) + DoubleCellAttackValue( attacker, target, from, targetCell );
     }
+    if ( attacker.isAllAdjacentCellsAttack() ) {
+        Indexes aroundAttacker;
+        if ( !attacker.isWide() ) {
+            aroundAttacker = GetAroundIndexes( from );
+        }
+        else {
+            // finding out where the tail of the unit will be
+            const Cell * currentHead = attacker.GetPosition().GetHead();
```
ihhub, line 533: Could you please cache `attacker.GetPosition()` call here?

src/fheroes2/battle/battle_board.cpp
```diff
@@ -523,6 +523,45 @@ int32_t Battle::Board::OptimalAttackValue( const Unit & attacker, const Unit & t
...
+            const int32_t currentPositionHeadX = currentHead->GetIndex() % ARENAW;
+            const int32_t newPositionHeadX = from % ARENAW;
+            const int32_t newPositionTailIndex
+                = ( newPositionHeadX == currentPositionHeadX )
+                      ? ( from + attacker.GetPosition().GetTail()->GetIndex() % ARENAW - attacker.GetPosition().GetHead()->GetIndex() % ARENAW )
+                      : ( newPositionHeadX < currentPositionHeadX ? from + 1 : from - 1 );
```
ihhub, line 539: Please avoid multiple ternary operations within the same statement as it impacts code readability.

ihhub, line 539: Isn't this statement depends on a side where the monster starts the battle? I think we should take into account `isReflect` as well.

ihhub (changes requested):
Hi @vincent-grosbois , I put few more comments here. Would you mind to check them?

src/fheroes2/battle/battle_board.cpp
oleg-derevenetz: There is also the `Battle::Position::GetCorrect()` static method that returns the correct `Position` of wide unit in such cases, considering the battlefield obstacles and other units. I'd suggest using it instead of homegrown logic of such kind.

src/fheroes2/battle/battle_cell.h
```diff
@@ -107,7 +107,8 @@ namespace Battle
         bool isReflect( void ) const;
         bool contains( int cellIndex ) const;
 
-        static Position GetCorrect( const Unit &, s32 );
+        // Returns the Position member that unit would get, if it moved to headIndex
+        static Position GetPositionWhenMoved( const Unit & unit, const int32_t headIndex );
```
oleg-derevenetz, line 111: Strictly speaking, the last parameter is not always the head index :) It is mostly, but it may be the tail index as well - for example, if there is some obstacle on the neighboring cell. In the `GetReachable()` I named it as `dst`. But it's just a cosmetic nitpick and may be changed later.

src/fheroes2/monster/monster.h
```diff
@@ -178,6 +178,7 @@ class Monster
     bool isTwiceAttack( void ) const;
     bool isRegenerating( void ) const;
     bool isDoubleCellAttack( void ) const;
+    bool isAllAdjacentCellsAttack( void ) const;
```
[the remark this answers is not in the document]
ihhub: clang-tidy is correct about this. Please remove redundant `void` parameter.

src/fheroes2/battle/battle_board.cpp
```diff
@@ -523,6 +523,25 @@ int32_t Battle::Board::OptimalAttackValue( const Unit & attacker, const Unit & t
         const int32_t targetCell = OptimalAttackTarget( attacker, target, from );
         return target.GetScoreQuality( attacker ) + DoubleCellAttackValue( attacker, target, from, targetCell );
     }
+    else if ( attacker.isAllAdjacentCellsAttack() ) {
```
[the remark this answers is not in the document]
ihhub: I'd agree with this change as well. Let's remove `else` keyword from here.

src/fheroes2/battle/battle_board.cpp
```diff
@@ -523,6 +523,25 @@ int32_t Battle::Board::OptimalAttackValue( const Unit & attacker, const Unit & t
...
+                unitsUnderAttack.insert( cell.GetUnit() );
+            }
+        }
+
+        int32_t attackValue = 0;
+        for ( const Unit * unit : unitsUnderAttack ) {
+            attackValue += unit->GetScoreQuality( attacker );
+        }
+        return attackValue;
+    }
```
ihhub, line 544: Hmm... I recently made the optimization changes here. Could you please put them back?

ihhub (changes requested):
Hi @vincent-grosbois , I added few more comments here. Would you mind to take a look at them?

src/fheroes2/battle/battle_cell.h
vincent-grosbois: I can rename it no worries
But bth I really don't understand what this method is doing then.
I thought the point was to get the position object if you moved the unit to headIndex
How can the logic work if you can't be sure if the input is the head or the tail index?

src/fheroes2/battle/battle_board.cpp
vincent-grosbois: I can't find them in the PR :(

src/fheroes2/battle/battle_cell.h
oleg-derevenetz: @vincent-grosbois When player points to some cell using the mouse cursor (for example, why flying, moving or using the Teleport spell), the cell he points to is mostly the head cell (in case of wide units), but there may be situations when this cell can be the tail cell only - for example, when this cell has an obstacle on the cell next to it (that would be the tail cell of the unit). I didn't write this method, but, according to its logic, it should return the correct `Position` object in such cases.

ihhub:
@Branikolog , please test this pull request as a priority. It must go to the upcoming release.

src/fheroes2/monster/monster.h
ihhub, line 181: Please remove `void` here. It's useless.

ihhub (commented):
Hi @vincent-grosbois , I left just one very minor comment. The rest looks good to me.

Branikolog:
Hello, @vincent-grosbois & @ihhub !
I'm not sure it's related to current PR, but I find a few issues with pathfinding algorithm for Hydras:
Hydra Issue
While placing my upper stack of Hydras (26) to that thin place, both enemy upper stacks (20+20) start behaving incorrectly: They skip their turn sometimes. Start moving backwards\forwards only for 1 hex inside that crack.
[Crash.zip](https://github.com/ihhub/fheroes2/files/7272960/Crash.zip)
Also, I've noticed here, that enemy constantly changes the position of attacking hydras, which causes an issue, when he blocks two available attack directions for his allied troops with one stack in a thin places, like on a screenshot above.

One more:
Hydra Issue #2
While placing upper 34 stack a bit lower, both upper enemy Hydras will skip their turn until the moment, lower AI Hydras will free a path under the crack.
[Patfnding.zip](https://github.com/ihhub/fheroes2/files/7272968/Patfnding.zip)

vincent-grosbois:
I think this can be due to me using incorrectly the call to GetCorrect() method, which is  still don't understand

Basically I'm using it as a method that should tell me, "if I move a `unit `to a given `index`, what should be its `position `after the move?" (so, including change of direction and that handles wide units) Apparently this function doesn't do this
Do we have a function that does something similar ?

oleg-derevenetz:
> Basically I'm using it as a method that should tell me, "if I move a unit to a given index, what should be its position after the move?" (so, including change of direction and that handles wide units) Apparently this function doesn't do this

It looks like this function should do exactly this (with some reservations in case of some corner cases though). For wide units it basically checks that unit will fit if the given cell is a head cell, and if not, it tries to fit it by turning the given cell into the tail cell.

vincent-grosbois:
Hi @Branikolog 
I tried to reproduce what you mentionned  but didn't find anything specific
Maybe if you tell us exactly where to move the units etc, it can be reproduced

Branikolog:
Hi, @vincent-grosbois 
At the "Crash" savefile just start attacking the village and then place Hydras right here:
image
image
You can watch in a battle log, that enemy Hydras skip.

vincent-grosbois:
Sorry I can't reproduce it on my end... I did the same thing you mentionned and the hydras never skip turns

vincent-grosbois:
> 
> 
> Hi, @vincent-grosbois. Sorry for the late answer. Check the video:
> 2021-10-07.11-25-25.mp4

Thanks !
I was able to check but it's completely unrelated to my commit

The issue is that in that situation, the hydra can't find any path to move to any of the enemy hydras, which is clearly not what should happen.
The specific line where pathfinding fails is : https://github.com/ihhub/fheroes2/blob/99f33c52e170ea6c12ed0082f67bdd2e8f6e7e2c/src/fheroes2/ai/normal/ai_normal_battle.cpp#L542

I suggest you open another issue to track this bug, as it's unrelated to this commit

PR #4273, opened 2021-09-27
Improve Week class

vincent-grosbois:
- using correct types
- don't inherit from std::pair
- use enum class
- moved relevant methods from Monster class to Week class

This will make it simpler to have "deterministic week change"


src/fheroes2/kingdom/week.cpp
```diff
@@ -26,77 +26,271 @@
...
 
     return "Unnamed";
 }
 
-int Week::WeekRand( void )
+Week Week::RandomWeek( const bool isNewMonth )
 {
-    return ( 0 == ( world.CountWeek() + 1 ) % 3 ) ? static_cast<int>( MONSTERS ) : Rand::Get( ANT, CONDOR );
+    const WeekName weekName = isNewMonth ? MonthRand( world ) : WeekRand( world );
+    const Monster::monster_t monster = ( weekName == WeekName::MONSTERS ) ? ( isNewMonth ? RandomMonsterMonthOf() : RandomMonsterWeekOf() ) : Monster::UNKNOWN;
```
ihhub, line 279: Could we please not to use multiple ternary operators withing one statement? We could go with if-else conditions.

src/fheroes2/kingdom/week.cpp
```diff
@@ -26,77 +26,271 @@
...
+    return sb;
 }
 
-StreamBase & operator>>( StreamBase & sb, Week & st )
+StreamBase & operator<<( StreamBase & stream, const Week & week )
 {
-    return sb >> st.first >> st.second;
-}
+    return stream << static_cast<int32_t>( week.GetType() ) << static_cast<int32_t>( week.GetMonster() );
+}
```
ihhub, line 296: Please add a new empty line at the end of the file.

ihhub (commented):
Hi @vincent-grosbois , I left just 2 comments here. Could you please take a look?

ihhub:
@vincent-grosbois , thanks a lot for these changes!

PR #4290, opened 2021-09-30
Show map type icons in New Map lists

### LINKED ISSUE #1411: Choice of original scenarios or expansion scenarios
In the original game, there is a window that allows the user to select the type of scenarios (original or expansion) that will be displayed in the scenario selection window.
In fheroes2, this window does not exist, and all the scenarios are displayed without distinction in the scenario selection window.

In original game:

Choice of scenarios type 01

In fheroes2:

Choice of scenarios type 02

ihhub:
Since we do not have sorting by victory/loss conditions or by number of player adding extra buttons for Map Type makes no sense unless we make a brand new dialog for all options.

close #1411

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -994,6 +994,28 @@ namespace fheroes2
                     }
                 }
                 return true;
+            case ICN::MAP_TYPE_ICON: {
+                // TODO: add a new icon for the Resurrection add-on map type.
+                _icnVsSprite[id].resize( 2 );
+                for ( Sprite & icon : _icnVsSprite[id] ) {
+                    icon.resize( 17, 17 );
+                    icon.fill( 0 );
+                }
+
+                const Sprite & successionWarsIcon = GetICN( ICN::ARTFX, 6 );
+                const Sprite & priceOfLoyaltyIcon = GetICN( ICN::ARTFX, 90 );
+
+                if ( !successionWarsIcon.empty() ) {
+                    Resize( successionWarsIcon, 0, 0, successionWarsIcon.width(), successionWarsIcon.height(), _icnVsSprite[id][0], 1, 1, 15, 15 );
+                }
+
+                if ( !priceOfLoyaltyIcon.empty() ) {
+                    Resize( priceOfLoyaltyIcon, 0, 0, priceOfLoyaltyIcon.width(), priceOfLoyaltyIcon.height(), _icnVsSprite[id][1], 1, 1, 15, 15 );
+                }
+
+                return true;
+            }
+
             default:
                 break;
             }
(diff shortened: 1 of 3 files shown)


LeHerosInconnu:
@ihhub.

I saw that you added icons for the "The Succession Wars" and "The Price of Loyalty" scenarios. Nice. :)
Scenario info icon 01

Looking at the screen, I thought the shield icon was for "The Succession Wars" scenarios and the crown icon was for "The Price of Loyalty" scenarios, but it's actually the other way around.
And this is confusing.
Although the Battle Garb of Enduran was introduced in the PoL expansion, I think a shield icon would be more appropriate for "The Succession Wars" scenarios.
Indeed, the box of the original game "The Succession Wars" looks like this:
Heroes-of-Might-and-Magic-II

And the original program icon looks like this:
Program Icone 01

While the box of the "The Price of Loyalty" expansion looks like this:
Heroes_of_Might_and_Magic_II_The_Price_of_Loyalty

For the time being (before we have maybe a specially designed icon), I think the "Sword of Anduran" or the "Helmet of Anduran" might be more appropriate for "The Price of Loyalty" scenarios.
What do you think?
And what the other team members think about it?

Branikolog:
Hi all.
For me personally icons are so tiny, so I can hardly distinguish what is inside those icons. I simply prefer icons look as much different, as they could: even colorful squares would be ok, but icons with something inside them is always better. ;)  If icon suits the version of the game, compatible maps are from, it's ok. But if it doesn't suit much - I don't really care.
I could even accept if all map compatibility icons will look like a garb (which looks like the game launch icon), but with different color (color of background and griffin on a shield, for example).
But in a future, I believe, we'd completely rework this window, adding new buttons to filter maps, so not sure, we should spend ton of time improving those small icons, we have now. ;)

LeHerosInconnu:
@Branikolog.

> Hi all. For me personally icons are so tiny, so I can hardly distinguish what is inside those icons. I simply prefer icons look as much different, as they could: even colorful squares would be ok, but icons with something inside them is always better. ;) If icon suits the version of the game, compatible maps are from, it's ok. But if it doesn't suit much - I don't really care. I could even accept if all map compatibility icons will look like a garb (which looks like the game launch icon), but with different color (color of background and griffin on a shield, for example). But in a future, I believe, we'd completely rework this window, adding new buttons to filter maps, so not sure, we should spend ton of time improving those small icons, we have now. ;)

Your remark certainly concerns for resolutions higher than the original resolution of the game.
I find that for the 640 x 480 resolution the icons are okay although they could be a bit more precise and defined (simpler) like the victory and defeat condition icons (the icons with a green and red background).
I agree that the background color of these icons should be quite distinctive, orange or yellow and blue for example.

ihhub:
Hi @LeHerosInconnu and @Branikolog , thanks for the feedback! I agree that icons are far from ideal but we don't have an artist to do the work. We'll definitely improve this part.

PR #4295, opened 2021-09-30
Add monster double attack cell highlighting

### LINKED ISSUE #1629: Combat, clearly indicate which hexagons will be affected by a creature attack that attacks on two hexagons
For creatures that attack on two hexagons, I suggest clearly indicating which hexagons will be affected in gray.
Currently, only one of the hexes affected by the attack is grayed out.

In fheroes2:

Indicate both attack hexes for creatures 01

ihhub:
close #1629

plus very minor code cleanup

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1677,30 +1677,15 @@ void Battle::Interface::RedrawCover()
 
             if ( _currentUnit->isDoubleCellAttack() ) {
                 // We have to invert the direction.
-                int attackDirection = 0;
-                if ( direction == TOP_LEFT ) {
-                    attackDirection = BOTTOM_RIGHT;
-                }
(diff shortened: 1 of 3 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1674,6 +1674,37 @@ void Battle::Interface::RedrawCover()
 
                 highlightCells.emplace( pos.GetTail() );
             }
+
+            if ( _currentUnit->isDoubleCellAttack() ) {
+                // We have to invert the direction.
```
oleg-derevenetz, line 1679: There is the `Battle::Board::GetReflectDirection()` static method that may be used here to reduce the code duplication.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion here, please take a look when it's convenient for you.

PR #4311, opened 2021-10-03
Fix infinite loop for AI trying to use Stoneliths

ihhub:
It happens when AI can't go anywhere else or the passage is blocked

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -20,6 +20,7 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include <algorithm>
 #include <cassert>
 
 #include "agg.h"
@@ -887,23 +888,35 @@ namespace AI
         DEBUG_LOG( DBG_AI, DBG_INFO, hero.GetName() );
     }
 
-    void AIToTeleports( Heroes & hero, s32 index_from )
+    void AIToTeleports( Heroes & hero, const int32_t startIndex )
     {
-        int indexTo = world.NextTeleport( index_from );
-
         const Route::Path & path = hero.GetPath();
-        if ( !path.empty() ) {
-            const int dest = path.front().GetIndex();
-            while ( indexTo != dest ) {
-                indexTo = world.NextTeleport( index_from );
-                const Maps::Tiles & tile = world.GetTiles( indexTo );
-                if ( index_from == indexTo || tile.isFog( hero.GetColor() ) || tile.GetObject() == MP2::OBJ_HEROES )
-                    break;
-            }
+        if ( path.empty() ) {
+            // Hero is not moving anywhere.
+            return;
         }
 
-        if ( index_from == indexTo ) {
-            DEBUG_LOG( DBG_AI, DBG_WARN, "teleport unsuccessfull, can't find exit lith" );
+        const int32_t endIndex = path.front().GetIndex();
+        const int32_t heroColor = hero.GetColor();
+
+        // To avoid infinite loops we have to verify that the list of tile indexes contains at least one reachable tile.
+        // So we get the list of them and remove those which are unsuitable.
+        MapsIndexes teleports = world.GetTeleportEndPoints( startIndex );
+        teleports.erase( std::remove_if( teleports.begin(), teleports.end(), [endIndex, heroColor]( const int32_t index ) {
+            if ( endIndex == index ) {
+                return false;
+            }
+            const Maps::Tiles & tile = world.GetTiles( index );
+            return !tile.isFog( heroColor ) && tile.GetObject() != MP2::OBJ_HEROES; } ), teleports.end() );
+
+        if ( teleports.empty() ) {
+            DEBUG_LOG( DBG_AI, DBG_WARN, "AI hero " << hero.GetName() << " has nowhere to go through stone liths." );
+            return;
+        }
+
+        const int32_t indexTo = Rand::Get( teleports );
+        if ( startIndex == indexTo ) {
+            DEBUG_LOG( DBG_AI, DBG_WARN, "AI hero " << hero.GetName() << " has teleportation tile the same as starting position: " << startIndex );
             return;
         }
 


oleg-derevenetz:
Hi @ihhub

> It happens when AI can't go anywhere else or the passage is blocked

Are you sure? This case should be handled correctly because in this case `world.NextTeleport()` returns `index_from` and loop exits via `break`.

oleg-derevenetz (commented):
Hi @ihhub

> It happens when AI can't go anywhere else or the passage is blocked

Are you sure? This case should be handled correctly because in this case `world.NextTeleport()` returns `index_from` and loop is exited via `break`.

ihhub:
> 
> 
> Hi @ihhub
> 
> > It happens when AI can't go anywhere else or the passage is blocked
> 
> Are you sure? This case should be handled correctly because in this case `world.NextTeleport()` returns `index_from` and loop is exited via `break`.



> 
> 
> Hi @ihhub
> 
> > It happens when AI can't go anywhere else or the passage is blocked
> 
> Are you sure? This case should be handled correctly because in this case `world.NextTeleport()` returns `index_from` and loop is exited via `break`.

Hi @oleg-derevenetz , `world.NextTeleport()` returns `index_from` only when no available teleport points are available. However, when it returns non-empty array we don't check every element in the array. So in my case this method was returning an array of 8 elements and none of them fulfilled the conditions in the loop making it running forever. I replace the loop logic by a single call of `world.NextTeleport()` and removing the indexes which are invalid.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -887,23 +888,38 @@ namespace AI
...
+        MapsIndexes teleports = world.GetTeleportEndPoints( startIndex );
+        teleports.erase( std::remove_if( teleports.begin(), teleports.end(),
+                                         [endIndex, heroColor]( const int32_t index ) {
+                                             if ( endIndex == index ) {
+                                                 return false;
+                                             }
+                                             const Maps::Tiles & tile = world.GetTiles( index );
+                                             return !tile.isFog( heroColor ) && tile.GetObject() != MP2::OBJ_HEROES;
+                                         } ),
+                         teleports.end() );
```
oleg-derevenetz, line 913: It's hard for me to say if this is OK, but I suppose that this code and its previous version are not equivalent. In previous version AI may teleport only to the following places: next cell of his `path`, some tile in the fog (have no idea why) or some tile with a hero on it (it's hard for me to say why this check for a hero is here, because the `World::GetTeleportEndPoints()` cannot return tile with a hero on it due to special check in that method). Only one of these tiles may be selected. However, in the new version AI may teleport to any free exit, not only to the next cell of his `path`.

oleg-derevenetz (commented):
Hi @ihhub I left another comment here, please take a look when it's convenient for you.

src/fheroes2/ai/ai_hero_action.cpp
ihhub: The previous code is:
```cpp
            const int dest = path.front().GetIndex();
            while ( indexTo != dest ) {
                indexTo = world.NextTeleport( index_from );
                const Maps::Tiles & tile = world.GetTiles( indexTo );
                if ( index_from == indexTo || tile.isFog( hero.GetColor() ) || tile.GetObject() == MP2::OBJ_HEROES )
                    break;
            }
```
which we can interpret as:
```cpp
            const int dest = path.front().GetIndex();
            while ( 1 ) {
                indexTo = world.NextTeleport( index_from );
                const Maps::Tiles & tile = world.GetTiles( indexTo );
                if ( indexTo == dest || index_from == indexTo || tile.isFog( hero.GetColor() ) || tile.GetObject() == MP2::OBJ_HEROES )
                    break;
            }
```
the condition `index_from == indexTo` must only occur if `world.NextTeleport` has no tiles available for teleports or in other words `teleports.empty()`.

Then later in the code we check if `index_from == indexTo` is true (or `teleports.empty()` logically) and then return from the function.

Coming to other cases and modifying `index_from == indexTo` condition in the code we have:
```cpp
            const int dest = path.front().GetIndex();
            while ( 1 ) {
                MapsIndexes teleports = world.GetTeleportEndPoints( index_from );
                if ( teleports.empty() ) {
                    indexTo = index_from;
                }
                else {
                    indexTo = Rand::Get( teleports );
                }

                const Maps::Tiles & tile = world.GetTiles( indexTo );
                if ( indexTo == dest || tile.isFog( hero.GetColor() ) || tile.GetObject() == MP2::OBJ_HEROES )
                    break;
            }
```

Then with empty teleports array we exit the loop and hit `if ( index_from == indexTo ) {` condition immediately returning from the function. Assuming that we handle this case before the loop so we have
```cpp
            const int dest = path.front().GetIndex();
            while ( 1 ) {
                MapsIndexes teleports = world.GetTeleportEndPoints( index_from );
                indexTo = Rand::Get( teleports );

                const Maps::Tiles & tile = world.GetTiles( indexTo );
                if ( indexTo == dest || tile.isFog( hero.GetColor() ) || tile.GetObject() == MP2::OBJ_HEROES )
                    break;
            }
```
If we have all elements in the array can't meet the condition we have the infinite loop. With my changes we remove all these **invalid** elements from the array and then pick any.

oleg-derevenetz: Suppose that we have three teleports on tiles 10, 20 and 30, hero is on tile 30 and his next tile in path should be tile 20.

With previous version of the code:

```
int indexTo = world.NextTeleport( index_from );
```

Suppose that `indexTo` is 10 here.

```
const int dest = path.front().GetIndex();
```

`dest` is 20.

```
while ( indexTo != dest ) {
```

10 is not equal to 20, entering the loop.

```
indexTo = world.NextTeleport( index_from );
```

Suppose that `indexTo` is 20 here.

```
if ( index_from == indexTo || tile.isFog( hero.GetColor() ) || tile.GetObject() == MP2::OBJ_HEROES )
```

30 is not equal to 20 and suppose that destination tile is not in a fog and there is no hero on it.

```
while ( indexTo != dest ) {
```

20 equals 20, exiting the loop.

```
if ( index_from == indexTo ) {
```

30 is not equal to 20, continuing execution with `indexTo` equal 20.

With new version of the code:

```
const int32_t endIndex = path.front().GetIndex();
```

`endIndex` is 20.

```
teleports.erase( std::remove_if( teleports.begin(), teleports.end(),
                                         [endIndex, heroColor]( const int32_t index ) {
                                             if ( endIndex == index ) {
                                                 return false;
                                             }
                                             const Maps::Tiles & tile = world.GetTiles( index );
                                             return !tile.isFog( heroColor ) && tile.GetObject() != MP2::OBJ_HEROES;
                                         } ),
                         teleports.end() );
```

`teleports` will contain tiles 10 and 20 after this filter.

```
const int32_t indexTo = Rand::Get( teleports );
```

Suppose that `indexTo` becomes 10 here.

```
if ( startIndex == indexTo ) {
```

30 is not equal to 10, continuing execution with `indexTo` equal 10.

ihhub: Sure. I'll do.

oleg-derevenetz: @ihhub I suppose this can be done in another PR :) Since this is merely an improvement, and this PR is a bugfix.

ihhub: Done.

PR #4312, opened 2021-10-03
Use within-loop variables in repeatPattern()

a1exsh:
Follow-up on #4203


ihhub:
@a1exsh , thank you very much for this code correction!

PR #4314, opened 2021-10-04
Update battle screenshot on the home page

a1exsh:
After fixing the border rendering in #4203.

I've just used GIMP to copy the correct pieces over from the correctly
rendered border. ;)


a1exsh:
Not sure how did you manage to make the original image file so much smaller in size: 321K vs. 508K.  Did you use some PNG optimization program for that?

ihhub:
Hi @a1exsh , we shouldn't do any editing for the existing screenshots. They must come straight from the game itself. We also fixed few more issues related to UI so I think a fresh screenshot is more welcomed. I asked @Branikolog to spend time for this.

a1exsh:
> Hi @a1exsh , we shouldn't do any editing for the existing screenshots. They must come straight from the game itself. We also fixed few more issues related to UI so I think a fresh screenshot is more welcomed. I asked @Branikolog to spend time for this.

OK, could see this coming. Reproducing the same battle screen is not easy, but at the same time it doesn't have to be exactly the same. 😅

ihhub:
Closing this pull request as we shouldn't edit screenshots from the game.

PR #4320, opened 2021-10-04
improve german localization

undef21:
1. add support of ß, ä, ö, ü letters;
2. switch translate file de.po to CP1252 encoding.


src/fheroes2/agg/agg_image.cpp
```diff
@@ -179,6 +179,19 @@ namespace fheroes2
                     imageArray.pop_back();
                     imageArray.erase( imageArray.begin() + 128 + 0xC0 - 0x80 );
                 }
+                // German version uses CP1252
+                verifiedFontWidth = ( id == ICN::FONT ) ? 10 : 7;
+                if ( imageArray.size() == 103 && imageArray[99].width() == verifiedFontWidth ) {
+                    imageArray.insert( imageArray.begin() + 0x80 - 0x20, 0xFC - 0x80, imageArray[0] );
```
ihhub, line 185: As discovered for Russian language please put final integer values everywhere.

ihhub (changes requested):
Hi @undef21 , I left few comments here despite of fact this request being as a Draft. Can you please check them? Also what do we need to do to proceed with this pull request further?

dimag0g:
Just FYI, I'm still working on a German translation. Let me know if I should push whatever changes I already have, so that other contributors don't repeat the work.

a1exsh:
@undef21 @ihhub how can we make progress here?

I'm checking the German translation from @dimag0g, and I'd pretty much like to use the umlaut-enabled version of the game for that.

undef21:
@dimag0g, I have the GOG version. font.icn and smalfont.icn files contain ä, ö, ü, ß, Ä, Ö, Ü letters. I don't generate anything.

a1exsh:
> I have the GOG version. font.icn and smalfont.icn files contain ä, ö, ü, ß, Ä, Ö, Ü letters. I don't generate anything.

@undef21 I cannot seem to find these letters in my GOG version (installed as English).  Could you please share the indexes at which these letters should be found in the ICNs?  In my version both end at index `095`.

undef21:
@a1exsh, these letters have indexes 96-102 in the German version of the game.
I guess each national version (except France) has ASCII letters with indexes 0-96 and language-specific letters after the ASCII part.

a1exsh:
> @a1exsh, these letters have indexes 96-102 in the German version of the game. I guess each national version (except France) has ASCII letters with indexes 0-96 and language-specific letters after the ASCII part.

Hm, interesting.

The game now supports many languages out of the box, e.g. Russian, and the font in that case is all generated, as I understand. It should be relatively straightforward to do the same for German, I guess?

@ihhub what do you think?

undef21:
@a1exsh, if game's localization same as selected language then the game uses the original font else the game uses generated one.

ihhub:
> > @a1exsh, these letters have indexes 96-102 in the German version of the game. I guess each national version (except France) has ASCII letters with indexes 0-96 and language-specific letters after the ASCII part.
> 
> Hm, interesting.
> 
> The game now supports many languages out of the box, e.g. Russian, and the font in that case is all generated, as I understand. It should be relatively straightforward to do the same for German, I guess?
> 
> @ihhub what do you think?

We already non-ideally generate these letters for German language.

a1exsh:
> We already non-ideally generate these letters for German language.

Hm, but how can I see it in action?  Do I need to build from the tip of this PR branch to see it?

undef21:
@a1exsh, you need to build this branch and German GOG resources.

a1exsh:
> @a1exsh, you need to build this branch and German GOG resources.

But then it's going to rely on the letters from resources.  I'm able to use Russian localization w/o OG-localized resources, for example, and I believe it should be possible to do the same for German. I think will just give it a try and see.

a1exsh:
@undef21 @ihhub so it doesn't seem to work without German GOG resources.  Here is what I see when building from this PR branch and adding a translation with `ö` for "Auflösung":
image

What am I missing?  Where in the code do we generate the umlauts and how to make them render?

heroesiiifan:
I don't see the generation in source either.

But I don't find it so extremely important. Normally you will use the german resources anyway. Otherwise you have a language mishmash because of the untranslated buttons.

I have also looked at the letter generation. but that is unfortunately not so self-explanatory for an outsider too add the 7 letters.

But, I think you should merge that anyway already, so that you have in the interaction with #4905 already times a basic German, working version (if you use the German resources).

We will sooner or later have to rework something anyway.

a1exsh:
> I don't see the generation in source either.

Looks like this is the place to add umlauts: https://github.com/ihhub/fheroes2/blob/bbb68b95e330003a09fa589cb4b4c35b3537994e/src/fheroes2/agg/agg_image.cpp#L195

For example, like done here: https://github.com/ihhub/fheroes2/blob/bbb68b95e330003a09fa589cb4b4c35b3537994e/src/fheroes2/agg/agg_image.cpp#L439

Laserlicht:
@undef21 Can you finish the draft? So that it can be merged?

Right now it is a bit messy. Only if you have english resources, the letters are displayed. 😅

Laserlicht:
@ihhub Is it possible to merge this PR?
So that the german resource files now also work correctly.

ihhub:
@undef21 , thank you very much for these changes!

PR #4330, opened 2021-10-06
Add Editor base code

ihhub:
We need to fill information about over 900 objects present in the game in order to have fully workable editor. This code is just a prototype but it could help to proceed with filling more information about objects.

Most likely we will need help of non-development team in order to get full information. In this case we have to store such structures in separate files (which we can put in h2d file.


ihhub:
Hi @idshibanov and @oleg-derevenetz , this is an excellent opportunity to bring JSON support to live. With this feature we can store all required info in very useful format. What do you think?

ihhub:
Closing this pull request as it was serving as a POC.

PR #4349, opened 2021-10-09
Fix button rendering in Restart battle dialog

### LINKED ISSUE #3513: Instant battle window, window frame is patchy near "OKAY" and "Restart" buttons
Dark part of window frame looks patchy under the buttons.
image
image

a1exsh:
Fix #3513


a1exsh:
> NB: currently both buttons are missing the shadow. This should be added separately, there are more instances like that (e.g. CONFIG button in the map info dialog).

How the shadows can be added: #4350

Branikolog:
Brilliant work!
We're still missing shadows for these buttons, but the lowest part of window looks like there's no shadow before for sure.

Issue #2261, opened 2020-11-26
Combat screen, Surrender window, "ACCEPT" and "DECLINE" buttons have no shadow


LeHerosInconnu:
In fheroes2, the "ACCEPT" and "DECLINE" buttons have no shadow.
The same flaw is present in the original game.

In fheroes2:

![Surrender fh2 02](https://user-images.githubusercontent.com/43583994/100390339-4a6aaa80-3030-11eb-8b12-40893f9a6ce0.png)

For fheroes2, I suggest vertically moving these buttons the same distance from the bottom edge of the window and adding a shadow in the same way as in this window:

![Surrendered fh2 01](https://user-images.githubusercontent.com/43583994/100391218-0cbb5100-3033-11eb-8fa6-0d059aa3fba8.png)

ihhub:
Hmmm... we then have to change the sprites... I'll take a look.

ihhub:
This issue was fixed in #4350.

PR #4350, opened 2021-10-09
Add missing button shadows in dialogs

a1exsh:
Relates to #4349


a1exsh:
Original:
image

Changed:
image

Branikolog:
Hi, @a1exsh 
Looks fantastic! :)
But I've noticed, that the original central button has its shadow (which is a part of image) merged with the frame darkening of the frame. Is it possible  to get rid of bright line (between shadow and the frame darkening) here?
config shadow

a1exsh:
@Branikolog good catch.  I think we can fix that by extending the shadow by just one more pixel (currently it is -4,4).

Branikolog:
@a1exsh 
Also, it feels like shadow to the left contrasts more with the background, comparing to the same place near central button. Is it just due to another background texture?

a1exsh:
> @a1exsh Also, it feels like shadow to the left contrasts more with the background, comparing to the same place near central button. Is it just due to another background texture?

@Branikolog that is because the `OKAY` button and the shadow beneath it are part of the original resource, but we add shadow manually for our custom button, and it doesn't match 100%.  Now that I think about it, in this case both buttons are of the same size, so what we can try is copying the original button together with the shadow instead.

Branikolog:
@a1exsh 
It's perfect!

a1exsh:
Experimental settings dialog:
image

Game config dialog:
Screenshot from 2021-10-12 18-09-08
image

Branikolog:
Hi, @a1exsh.
One more case found.
image

a1exsh:
Still quick (as in "hack"), but less dirty corners:
image

a1exsh:
Not sure if there are button shadows for this dialog in OG, will check later:
image

a1exsh:
There are no shadows, but I guess it might have looked a bit better with them:
heroes2_006

a1exsh:
Fixed gift resources buttons position and added shadows.

Before:
image

After:
Screenshot from 2021-10-30 13-15-23

It is kinda frustrating that every dialog has its own way of calculating position and border and placing the buttons on the window...  Are there any plans to start working on a proper UI toolkit? :) @ihhub

Branikolog:
> Not sure if there are button shadows for this dialog in OG, will check later: image

We have an opened issue for that. :)
#2261
Just not to forget to close it after fixing. ;)

a1exsh:
It really becomes an art at some point: every single button is unique %)

image

src/fheroes2/gui/ui_button.cpp
```diff
@@ -172,6 +182,16 @@ namespace fheroes2
...
     }
 
+    const Sprite & ButtonBase::_getShadow() const
+    {
+        if ( !_shadowSprite ) {
+            // TODO: button corners have to be cleaned up for the perfect shadow
+            const Sprite & shadow = fheroes2::makeShadow( _getReleased(), fheroes2::Point( -4, 6 ), 3 );
+            _shadowSprite.reset( new Sprite( shadow ) );
+        }
+        return *_shadowSprite.get();
```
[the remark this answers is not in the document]
a1exsh: Not sure about this one: I've just copied the approach from `_getDisabled()`.

a1exsh:
While working on this I've noticed another glitch related to the Evil "show hero" / "show market" buttons used in Surrender dialog as well as when an army is willing to join the hero.

To make the pressed state we replace all back pixels (color index 36) with transform 1, but the there are pixels with the same index also in the middle of the pressed Evil buttons, not only around the border.  This results in background leaking through the drawing, which looks like this:
image

A have a fix in the works, will either push here or open a new PR, depending on the resulting size.

a1exsh:
Alright, I'm now content with this change myself — please have a look when you have time. :)

Branikolog:
Hi, @a1exsh !
I've noticed one thing, related to shadows. While pressing "Hero info" or "Marketplace" buttons, shadow becomes more bold. While common "Yes"/"NO" buttons have same shadow in a pressed state:
image
image
Actually, this PR becomes really huge. So this could be fixed in a separate one, if you wish. Anyway, you've done a ton of work with buttons already! Thanks!

I've also found out, that "GIFT" and "EXIT" buttons at evil interface become surrounded with gray frame once being pressed:
image
The same to Surrender buttons:
image
image
And "GIFT" window:
image
image
A small piece of background changes while pressing OK button at setting window:
image
Corner
And same to evil:
image

a1exsh:
@Branikolog thanks for having a look — I have a fix for the button corners as well, just didn't push it yet to avoid bloating the PR. Will check the rest of your findings later. :)

Branikolog:
@a1exsh 
Thank You!

a1exsh:
> While pressing "Hero info" or "Marketplace" buttons, shadow becomes more bold.
 image

Good catch — will fix it.

> I've also found out, that "GIFT" and "EXIT" buttons at evil interface become surrounded with gray frame once being pressed:
image

This was the case before changes from this PR — let's fix it separately (I already know how to do that).

> The same to Surrender buttons:
image

Also was the case before, need to check how to fix it.

> And "GIFT" window:
image

Was the case, fixing seems easy.

> A small piece of background changes while pressing OK button at setting window:
image

Here I set a trap for myself: need to figure out why I did change the button ICN in this code :sweat_smile:

a1exsh:
> While pressing "Hero info" or "Marketplace" buttons, shadow becomes more bold.
> A small piece of background changes while pressing OK button at setting window:

@Branikolog fixed for both evil and good modes —- please have a look.  Let's fix other issues separately. :)

Branikolog:
@a1exsh 

> @Branikolog fixed for both evil and good modes —- please have a look. Let's fix other issues separately. :)

Sure!

a1exsh:
@Branikolog would you have time for another look on this? :)

Branikolog:
Hi, @a1exsh !
I'm not sure, if all the cases are related to current PR. So be free to postpone some of them to future.

I've noticed, that Creature info window has smaller shadows for buttons.
image
Also, unavailable "ACCEPT" button has a small black line along left and bottom border.
image
Pretty sure, we haven't discussed it before, but I'm not sure it is related to the current PR.
And the same to Dig and supposedly other system buttons.
image

a1exsh:
> Hi, @a1exsh ! I'm not sure, if all the cases are related to current PR. So be free to postpone some of them to future.
> 
> I've noticed, that Creature info window has smaller shadows for buttons.
image

Right.  This inconsistency comes from OG resources.  Not sure if we would try to make all of UI consistent one day, but at least we should be consistent within a single dialog window.

> Also, unavailable "ACCEPT" button has a small black line along left and bottom border.
image
> Pretty sure, we haven't discussed it before, but I'm not sure it is related to the current PR.
> And the same to Dig and supposedly other system buttons.
image

Nicely spotted — I can see it as well.  This is caused by our generic logic of producing disabled button sprites by applying a darkening palette.  All evil disabled buttons might suffer from it.  We might want to tackle it separately, if at all.

a1exsh:
@Branikolog I'd like to wrap this change up and move on to the next one: I've got a few in the pipeline, but they depend on this one being merged first. ;)

Branikolog:
@a1exsh 

> @Branikolog I'd like to wrap this change up and move on to the next one: I've got a few in the pipeline, but they depend on this one being merged first. ;)

Please, don't hesitate making small changes one my one, rather than merging huge part of new code.
You know, I'm always complaining on new issues being noticed. Ignore it! :D

a1exsh:
@ihhub please have another look when you have a minute. :)

a1exsh:
@ihhub maybe you could have a look on the code — the changes related to shadows are now complete, but I have more related improvements in pipeline. :)

src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -887,13 +876,17 @@ bool Battle::DialogBattleSurrender( const HeroBase & hero, u32 cost, Kingdom & k
             btnMarket.disable();
         }
         else {
+            btnMarket.drawShadow();
+            btnMarket.captureBackground();
```
ihhub, line 880: Can you please explain the logic for this code? Are we drawing a shadow and then we're copying a background? If yes shouldn't it be easier to create a factory function which will accept a background image and 2 button images? The function will just return either a pair of images or a a button? We know that the background is not going to be changed.

ihhub (changes requested):
Hi @a1exsh , I raised few design related questions here. Would you mind to take a look at them?

src/engine/image.cpp
```diff
@@ -1797,6 +1797,22 @@ namespace fheroes2
...
+        const int32_t width = in.width();
+        const int32_t height = in.height();
+
+        const uint8_t * transformIn = in.transform();
+        const uint8_t * transformInEnd = transformIn + height * width;
+        uint8_t * transformOut = out.transform();
+        for ( ; transformIn != transformInEnd; ++transformIn, ++transformOut ) {
+            *transformOut = *transformIn;
+        }
+    }
```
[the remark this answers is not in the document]
a1exsh: Good point :sweat_smile:  I think I've had some extra logic there, but after removing that it's not really different from plain `memcpy()`.

src/fheroes2/battle/battle_dialogs.cpp
a1exsh: That's a good question: now I will have to recall everything, it has been a few weeks since I was last working on this...

ihhub: I think a single function should fulfill all our needs in this greatly reducing the code complexity.

a1exsh: @ihhub thanks a lot for your suggestions — I think I've addressed them all and the code is indeed much cleaner now.  I've added a special `AutoShadowButton` class to work with the shadows.  Please have another look!

a1exsh:
@Branikolog I did some code refactoring per @ihhub 's request.  To my eye everything looks good, but a slight problem remains with the pressed state of the Hero and Marketplace buttons in the Level Up and Army Joining dialogs (and that precisely because we do not draw shadows for them there, so also not capturing the background to restore it before drawing the pressed state).  I prefer to address this separately, as it is not related to adding shadows, which is the title of the PR. :)

Branikolog:
@a1exsh 
Sure! I believe it's much easier to work on fixes in separate PRs (even separate for each button) as well as test the changes. 
I'll check the PR once again in a few hours.

a1exsh:
> a slight problem remains with the pressed state of the Hero and Marketplace buttons in the Level Up and Army Joining dialogs

Hm, I just went through the changes once again and I see that this was handled by custom code before, so it can be seen as a regression.  I will still try to fix it right here, it shouldn't be too much of additional code.

a1exsh:
>  I will still try to fix it right here, it shouldn't be too much of additional code.

Fixed :)

a1exsh:
@Branikolog I've found and fixed an issue with the `OKAY` button shadow in the battle summary dialog, introduced by refactoring.
@ihhub please have another look at the code. :)

Branikolog:
Hi, @a1exsh !
Found an issue with "Town Portal" buttons being pressed.
image

a1exsh:
> Hi, @a1exsh ! Found an issue with "Town Portal" buttons being pressed. image

If the issue you're referring to is with button corners — it is not new and I'm waiting for this PR to be merged before addressing these. I think it was discussed already in the comments here.

Branikolog:
Sorry, @a1exsh. I have too much tasks now. :) Missed this moment.
I was just confused by the fact, the latest commit version doesn't have these corners.

a1exsh:
@Branikolog 
> I was just confused by the fact, the latest commit version doesn't have these corners.

It may be, but the issue is definitely present in the latest release and it's not directly related to the shadows, so let's address it separately.

a1exsh:
@ihhub please find the time to have a look — I've extracted the relevant code to new subclasses of the base button class.

src/engine/image.h
```diff
@@ -267,4 +270,9 @@ namespace fheroes2
     Image Stretch( const Image & in, int32_t inX, int32_t inY, int32_t widthIn, int32_t heightIn, int32_t widthOut, int32_t heightOut );
 
     void Transpose( const Image & in, Image & out );
+
+    // Generates a new image with a shadow of the shape of existing image. Shadow must have only (-x, +y) offset.
+    Sprite addShadow( const Sprite & in, const Point & shadowOffset, const uint8_t transformValue );
+
+    Sprite makeShadow( const Sprite & in, const Point & shadowOffset, const uint8_t transformValue );
```
ihhub, line 277: Please put these functions to places to have alphabetical order of function names. You can ignore that these function names don't start from a capital letter.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1441,6 +1459,40 @@ namespace fheroes2
...
+
+                const int releasedIndex = ( id == ICN::EVIL_ARMY_BUTTON ) ? 0 : 4;
+                _icnVsSprite[id][0] = GetICN( ICN::ADVEBTNS, releasedIndex );
+                AddTransparency( _icnVsSprite[id][0], 36 );
+
+                Sprite pressed = GetICN( ICN::ADVEBTNS, releasedIndex + 1 );
+                AddTransparency( pressed, 36 );
+                AddTransparency( pressed, 61 ); // remove the extra brown border
+
+                Sprite fixed( pressed.width(), pressed.height(), pressed.x(), pressed.y() );
```
ihhub, line 1487: Can we place name this variable by more understandable name? Right now `fixed` is a confusing name.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1441,6 +1459,40 @@ namespace fheroes2
...
+                fixed.reset();
+                // put back pixels that actually should be black
+                Fill( fixed, 1, 4, 31, 31, 36 );
+                Blit( pressed, fixed );
+                _icnVsSprite[id][1] = fixed;
```
ihhub, line 1492: Please use `std::move` instead of copying the image.

ihhub: Even better: we can directly use `_icnVsSprite[id][1]` instead of `fixed`.

src/fheroes2/gui/ui_button.cpp
```diff
@@ -240,7 +240,76 @@ namespace fheroes2
         return _disabled;
     }
 
-    ButtonGroup::ButtonGroup( const Rect & area, int buttonTypes )
+    AutoBackgroundButton::AutoBackgroundButton( const Image & in, int32_t offsetX, int32_t offsetY, int icnId, uint32_t releasedIndex, uint32_t pressedIndex )
+        : ButtonSprite( offsetX, offsetY, fheroes2::AGG::GetICN( icnId, releasedIndex ), fheroes2::AGG::GetICN( icnId, pressedIndex ) )
+    {
+        _captureBackground( in );
+    }
```
ihhub, line 247: This portion of the code worries me a lot. We're using `ButtonSprite` class as a base class which accepts images as constructor arguments. Here we're using ICN indexes to get images which later we copy to internal ButtonSprite members and then we're overriding them with new images within `_captureBackground` method. Also it's unknown what **in** image is for.

Name `AutoBackgroundButton` is misleading. Is it a button with automated background?

The same set of remarks is applied for `AutoShadowButton` class.

I'm suggesting to just create 2 or 4 functions, each will return ButtonSprite object. For example:
```cpp
ButtonSprite generateButtonWithShadow( int32_t offsetX, int32_t offsetY, const Sprite & released, const Sprite & pressed, const Sprite & disabled, const Image & background )
{
    // Prepare images
    ... // some magic code mostly the same like in `_captureBackground` method

    return { /* ButtonSprite constructor parameters */ };
}
```

On top of this `ButtonGroup` shouldn't know about shadow thing. We can pass ButtonSprite objects to it.

What do you think @a1exsh ?

ihhub (changes requested):
Hi @a1exsh , I left few more comments in this pull request. Could you please take a look at them and check if they make sense, especially from architectural point of view?

src/fheroes2/gui/ui_button.cpp
a1exsh: Sure, we can do it with a function instead.

The trouble with `ButtonGroup` and adding buttons to it explicitly is that the code to calculate buttons positions is not reused.  I would like us to use the shared code more.

a1exsh: @ihhub well, that's nearly not as easy as it sounded.  Instances of `ButtonBase` or derived classes cannot be copied, due to embedded `unique_ptr`.

With the following code
```cpp
    fheroes2::ButtonSprite btnCancel = fheroes2::makeButtonWithShadow( pos_rt.x + buttonOffset + 129, pos_rt.y + 410, fheroes2::AGG::GetICN( buttonCancelICN, 0 ),
                                                                       fheroes2::AGG::GetICN( buttonCancelICN, 1 ), display );
```
I'm getting this error:
```
../fheroes2/battle/battle_dialogs.cpp: In member function ‘bool Battle::Arena::DialogBattleSummary(const Battle::Result&, const std::vector<Artifact>&, bool) const’:
../fheroes2/battle/battle_dialogs.cpp:501:125: error: use of deleted function ‘fheroes2::ButtonSprite::ButtonSprite(const fheroes2::ButtonSprite&)’
  501 |                                                                        fheroes2::AGG::GetICN( buttonCancelICN, 1 ), display );
      |                                                                                                                             ^
In file included from ../fheroes2/battle/battle_interface.h:33,
                 from ../fheroes2/battle/battle_dialogs.cpp:31:
../fheroes2/gui/ui_button.h:99:11: note: ‘fheroes2::ButtonSprite::ButtonSprite(const fheroes2::ButtonSprite&)’ is implicitly deleted because the default definition would be ill-formed:
   99 |     class ButtonSprite : public ButtonBase
      |           ^~~~~~~~~~~~
../fheroes2/gui/ui_button.h:99:11: error: use of deleted function ‘fheroes2::ButtonBase::ButtonBase(const fheroes2::ButtonBase&)’
../fheroes2/gui/ui_button.h:30:11: note: ‘fheroes2::ButtonBase::ButtonBase(const fheroes2::ButtonBase&)’ is implicitly deleted because the default definition would be ill-formed:
   30 |     class ButtonBase : public ActionObject
      |           ^~~~~~~~~~
../fheroes2/gui/ui_button.h:30:11: error: use of deleted function ‘std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) [with _Tp = fheroes2::Sprite; _Dp = std::default_delete<fheroes2::Sprite>]’
In file included from /usr/include/c++/10/memory:83,
                 from ../fheroes2/system/players.h:26,
                 from ../fheroes2/army/army.h:30,
                 from ../fheroes2/battle/battle_dialogs.cpp:27:
/usr/include/c++/10/bits/unique_ptr.h:468:7: note: declared here
  468 |       unique_ptr(const unique_ptr&) = delete;
      |       ^~~~~~~~~~
```

Any idea how to overcome it?

ihhub: Ah, we have to define move-constructor and move operator but delete copy-constructor and copy operator for ButtonSprite class.

a1exsh: @ihhub I managed to convince the compiler, but not sure I can convince myself the code is now better :joy: 

This is getting really out of hand, and is really outside of my C++ capabilities...

ihhub: We all learn something new every day :) I'll take a look.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -531,16 +531,29 @@ bool ActionSpellTownPortal( Heroes & hero )
...
+        new fheroes2::ButtonSprite( fheroes2::makeButtonWithShadow( area.x + border, area.y + area.height - border - buttonOkIcn.height(), buttonOkIcn,
+                                                                    fheroes2::AGG::GetICN( icnId, 2 ), display ) ) );
+
+    std::unique_ptr<fheroes2::ButtonBase> buttonCancel( new fheroes2::ButtonSprite(
+        fheroes2::makeButtonWithShadow( area.x + area.width - border - buttonCancelIcn.width(), area.y + area.height - border - buttonCancelIcn.height(), buttonCancelIcn,
+                                        fheroes2::AGG::GetICN( icnId, 4 ), display ) ) );
+
+    fheroes2::ButtonGroup btnGroup;
+    btnGroup.createButton( buttonOk, Dialog::OK );
+    btnGroup.createButton( buttonCancel, Dialog::CANCEL );
```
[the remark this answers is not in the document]
ihhub: `createButton` method should have this signature
```cpp
void addButton( ButtonSprite && button, const int returnValue )
{
    addButton( new ButtonSprite( std::move( button ) ),  returnValue );
}
```
In this case we don't need unique_ptr usage but can directly invoke `makeButtonWithShadow` call within `addButton` method.

ihhub (changes requested):
Hi @a1exsh , I put few more comments here. I feel we're almost there. Can you please check the comments?

src/engine/image.cpp
```diff
@@ -1803,6 +1847,15 @@ namespace fheroes2
         }
     }
 
+    void ReplaceTransform( const Image & in, Image & out )
```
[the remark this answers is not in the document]
a1exsh: :heavy_check_mark:

src/fheroes2/gui/ui_button.cpp
```diff
@@ -429,4 +454,47 @@ namespace fheroes2
             _button[i]->unsubscribe();
         }
     }
+
+    ButtonSprite makeButtonWithBackground( int32_t offsetX, int32_t offsetY, const Sprite & released, const Sprite & pressed /*, const Sprite & disabled*/,
```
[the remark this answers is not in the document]
a1exsh: Oops, I thought I got rid of that one.

src/fheroes2/gui/ui_button.h
```diff
@@ -101,8 +100,13 @@ namespace fheroes2
     public:
         ButtonSprite( int32_t offsetX = 0, int32_t offsetY = 0 );
         ButtonSprite( int32_t offsetX, int32_t offsetY, const Sprite & released, const Sprite & pressed, const Sprite & disabled = Sprite() );
+        ButtonSprite( ButtonSprite & ) = delete;
+        ButtonSprite( ButtonSprite && o ) noexcept;
```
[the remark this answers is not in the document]
a1exsh: OK

src/fheroes2/heroes/heroes_spell.cpp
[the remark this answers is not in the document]
a1exsh: I had the same concerns, but didn't know how to solve it — thanks! :)

a1exsh:
@ihhub it complies and works, but I will not pretend I understood this `std::move()` magic :sweat_smile:

src/fheroes2/gui/ui_button.cpp
```diff
@@ -429,4 +479,45 @@ namespace fheroes2
...
+        Sprite releasedWithBackground( croppedBackground, 0, 0 );
+        Blit( released, releasedWithBackground, released.x(), released.y() );
+
+        Sprite pressedWithBackground( croppedBackground, 0, 0 );
+        Blit( pressed, pressedWithBackground, pressed.x(), pressed.y() );
+
+        Sprite disabled( released );
+        ApplyPalette( disabled, PAL::GetPalette( PAL::PaletteType::DARKENING ) );
+
+        Sprite disabledWithBackground( croppedBackground, 0, 0 );
```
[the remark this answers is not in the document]
ihhub:
```cpp
Sprite disabledWithBackground( std::move( croppedBackground ) );
disabledWithBackground.setPosition( 0, 0 );
```

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -531,16 +531,24 @@ bool ActionSpellTownPortal( Heroes & hero )
...
 
-    fheroes2::ButtonGroup btnGroups;
-    btnGroups.createButton( area.x + 5, area.y + 222, isEvilInterface ? ICN::NON_UNIFORM_EVIL_OKAY_BUTTON : ICN::NON_UNIFORM_GOOD_OKAY_BUTTON, 0, 1, Dialog::OK );
-    btnGroups.createButton( area.x + 187, area.y + 222, isEvilInterface ? ICN::NON_UNIFORM_EVIL_CANCEL_BUTTON : ICN::NON_UNIFORM_GOOD_CANCEL_BUTTON, 0, 1,
-                            Dialog::CANCEL );
-    btnGroups.draw();
+    const int32_t border = 10;
+    const int icnId = isEvilInterface ? ICN::SYSTEME : ICN::SYSTEM;
+    const fheroes2::Sprite & buttonOkIcn = fheroes2::AGG::GetICN( icnId, 1 );
+    const fheroes2::Sprite & buttonCancelIcn = fheroes2::AGG::GetICN( icnId, 3 );
```
ihhub, line 537: These aren't ICNs but images. It's good to call them `buttonOkImage` and `buttonCancelmage`.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -512,14 +512,14 @@ bool ActionSpellTownPortal( Heroes & hero )
         return false;
     }
 
-    std::unique_ptr<fheroes2::StandardWindow> frameborder( new fheroes2::StandardWindow( 290, 250 ) );
```
ihhub, line 515: I remember that was done on purpose to avoid some drawing issues. Can you please explain why are we changing it?

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -230,9 +230,21 @@ void Dialog::MakeGiftResource( Kingdom & kingdom )
     ResourceBar info2( funds2, box.x + 25, box.y + 150 );
     info2.Redraw();
 
-    fheroes2::ButtonGroup btnGroups( box, Dialog::OK | Dialog::CANCEL );
-    btnGroups.button( 0 ).disable();
-    btnGroups.draw();
+    const int32_t border = 10;
+    const int icnId = Settings::Get().ExtGameEvilInterface() ? ICN::SYSTEME : ICN::SYSTEM;
+    const fheroes2::Sprite & buttonOkIcn = fheroes2::AGG::GetICN( icnId, 1 );
+    const fheroes2::Sprite & buttonCancelIcn = fheroes2::AGG::GetICN( icnId, 3 );
```
ihhub, line 236: These aren't ICNs but images. Shall we call them `buttonOkImage` and `buttonCancelImage` respectively?

ihhub (changes requested):
Hi @a1exsh , I left just few more optimization related comments. Could you please check them too?

ihhub:
> @ihhub it complies and works, but I will not pretend I understood this `std::move()` magic 😅

Move semantics are used to avoid memory copy and only for objects which are going to be destroyed just after the call. This is useful when you can just swap values instead of copy them. In case of images it's essential to avoid extensive memory manipulations.

a1exsh:
> Move semantics are used to avoid memory copy and only for objects which are going to be destroyed just after the call.

Well, that is clear.  What I don't get is why in some places it's OK to just call `new ButtonSprite( ... )` and in other places you have to specify `std::move()` explicitly...

src/fheroes2/heroes/heroes_spell.cpp
a1exsh: @ihhub you are right, we should not change it.  I now see why it was done, after consulting the commit history: https://github.com/ihhub/fheroes2/commit/438257c7b7f7b206c319c514d83c2d83dc3f801c

I will put it back and add a comment for the next developer who comes across it and thinks we don't need the pointer there :D

src/fheroes2/dialog/dialog_giftresources.cpp
a1exsh: True.  They are actually Sprites, so called them like that.

src/fheroes2/gui/ui_button.cpp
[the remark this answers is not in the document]
a1exsh: This does feel like a micro-optimization, but doesn't seem to hurt readability too much. ;)

ihhub:
> Well, that is clear. What I don't get is why in some places it's OK to just call `new ButtonSprite( ... )` and in other places you have to specify `std::move()` explicitly...

std::move enforces in some sense to end object's lifetime within your code block. If you don't do this then you're passing by normal reference rather than by r-value.

src/fheroes2/heroes/heroes_spell.cpp
ihhub: Great!

src/fheroes2/gui/ui_button.cpp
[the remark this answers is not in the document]
ihhub: Haha. It's actually a huge performance boost as we don't make system calls to allocate and free memory on stack.

ihhub (approved):

ihhub:
@Branikolog please retest the latest build again as we changed a lot of internal code.

a1exsh:
appveyor build failed with an internal error, please re-trigger:
```
MSBUILD : error MSB4017: The build stopped unexpectedly because of an unexpected logger failure.
Microsoft.Build.Exceptions.InternalLoggerException: The build stopped unexpectedly because of an unexpected logger failure. ---> Microsoft.Build.Exceptions.InternalLoggerException: The build stopped unexpectedly because of an unexpected logger failure. ---> System.Net.WebException: The remote server returned an error: (500) Internal Server Error.
```

ihhub:
@a1exsh , a huge thanks for these changes!

PR #4372, opened 2021-10-10
Battle::Position::GetPositionWhenMoved(): the destination index is not necessarily the head index

oleg-derevenetz:
Minor style nits after #4262


ihhub:

PR #4374, opened 2021-10-10
fixing files installation properties with cmake on windows

PierceLBrooks:
fixing https://github.com/ihhub/fheroes2/commit/2daacebbd238dba9fc50eaa6ebe004610514c69c#r57787888 with separate call to install function in root-level cmake script


ihhub:
@PierceLBrooks , thank you very much for fixing this issue!

PR #4395, opened 2021-10-11
Maps::Tiles::PlaceMonsterOnTile(): don't push an existing MONS32.ICN to addons

### LINKED ISSUE #4361: Adventure map, it should be possible to dig at this square
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, it should be possible to dig at this square.

In the original game, once the resource pile and the wandering troop are removed from the adventure map, it is possible to dig at these two squares.
In fheroes2, it is not possible to dig at the square where the wandering troop was located.

In original game:

Why no digging 001
Why no digging 002
Why no digging 003
Why no digging 004
Why no digging 005
Why no digging 006

In fheroes2:

Why no digging 007
Why no digging 008
Why no digging 009
Why no digging 010

### Save file

Here are the save files:
[Why no digging 01.zip](https://github.com/ihhub/fheroes2/files/7317625/Why.no.digging.01.zip)

### Additional info

Version 0.9.8 4311

oleg-derevenetz:
fix #4361

I'm not sure if this is the right fix, but it looks like it should be done like this.

### DIFF
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -940,8 +940,9 @@ void Maps::Tiles::PlaceMonsterOnTile( Tiles & tile, const Monster & mons, const 
 {
     tile.SetObject( MP2::OBJ_MONSTER );
 
-    // if there was another sprite here (shadow for example) push it down to Addons
-    if ( tile.objectTileset != 0 && tile.objectIndex != 255 ) {
+    // if there was another sprite here (shadow for example) push it down to Addons,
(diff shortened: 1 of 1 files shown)


ihhub:
> 
> 
> Hi @oleg-derevenetz I think the logic is correct (it's what I wanted to implemented when I had a look at this)
> 
> I this possible however to use some enum values insted of '48' as int ?

Sadly we don't have any enumeration values for this.

oleg-derevenetz:
Hi @vincent-grosbois

> I this possible however to use some enum values insted of '48' as int ?

I don't see any enum suitable for the tileset.

PR #4413, opened 2021-10-13
isShadow doesn't allocate

vincent-grosbois:
currently the call to isShadow is allocating / deallocating on each call

This can be fixed easily
If you're ok with this change, I'll change all the other tilesets with
the same trick next


oleg-derevenetz:
Hi @vincent-grosbois why not just declare `shadows` as `static` instead?

ihhub:
Even better. Should be then make every bitset of the size of 256?

PR #4419, opened 2021-10-14
Add hotkeys for selection of SW/PoL campaign

a1exsh:
Use the same hotkeys as in OG: "O" for "Original", "E" — for "Expansion".


ihhub:
@a1exsh , a bunch of thanks for making this improvement!

PR #4420, opened 2021-10-14
Handle default exit hotkey in Campaign scenario select / info

a1exsh:
Otherwise you have to click the `Cancel` button with the mouse.  The hotkey
saves a lot time when testing changes.


ihhub:
@a1exsh , thank you very much for this improvement!

Issue #4409, opened 2021-10-13
Campaign info window, "VIEW INTRO" button doesn't have "pressed" state


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Button doesn't change when we press, "VIEW INTRO".
![image](https://user-images.githubusercontent.com/55348946/137150659-dc4d9ed4-e31d-4896-8e84-159268ed6280.png)

### Save file

Start any campaign and check scenario info.

pflugs30:
@Branikolog I believe that image failed to upload.

pflugs30:
> Anyway, We cannot hide buttons, as they are a part of the window itself. (Well, we can, but it really not easy task) So disabled button seems natural and easy solution here.

It seems that hiding a button is possible and relatively straight-forward.  In fact, that's how we don't show the **RESTART** button until post-launch: `buttonRestart.hide();`

I appreciate the response, and the logic makes sense.  I will create a PR for the fix to keep this change small and tight.

PR #4431, opened 2021-10-15
Hide the OKAY button in campaign info window

### LINKED ISSUE #4377: Campaign info window, shadow of the "OKAY" button
### Preliminary checks

### Platform

Windows

### Describe the bug

Start campaign and open scenario info window.
"OKAY" button has too bold shadow in this window. (For both good and evil campaigns)
image
Related issue with fixing shadow was here:
https://github.com/ihhub/fheroes2/issues/1412

In the OG we don't have this button disabled:
image
"OKAY" AND "CANCEL" buttons work in the same way: simply cancel the window.
But actually, I prefer fheroes2 implementation, as making two different buttons work the same is senseless and while looking at the disabled button player wouldn't consider this window as scenario selection window.

### Save file

[Evil_1_Complete.zip](https://github.com/ihhub/fheroes2/files/7320483/Evil_1_Complete.zip)

### Additional info

_No response_

a1exsh:
Fix #4377

Also disable the `RESTART` button when it's hidden: otherwise clicking on the
empty area where the button should be leads to the action unexpectedly.

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -614,13 +614,15 @@ fheroes2::GameMode Game::SelectCampaignScenario( const fheroes2::GameMode prevMo
 
     if ( allowToRestart ) {
         buttonOk.disable();
+        buttonOk.hide();
         buttonRestart.draw();
     }
     else {
+        buttonRestart.disable();
         buttonRestart.hide();
+        buttonOk.draw();
     }
 
-    buttonOk.draw();
     buttonCancel.draw();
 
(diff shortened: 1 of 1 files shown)


Branikolog:
> Fix #4377
> 
> Also disable the `RESTART` button when it's hidden: otherwise clicking on the empty area where the button should be leads to the action unexpectedly.

Good catch, @a1exsh! :)

ihhub:
@a1exsh , thank you so much for fixing these issues!

pflugs30:
I agree, well done!  A good improvement over the OG.

For the record, some additional discussion on #4409.

PR #4443, opened 2021-10-17
More dialogs to exit with READY hotkey

a1exsh:
Hero, Castle and Kingdom dialogs have only one button: `EXIT`.  Especially
when playing with one hand holding the mouse (to the right of the keyboard) it
is more convenient to exit these screens by hitting the num-pad `ENTER`.

Otherwise one has to either use the mouse to click on the button or use one's
left hand to hit `ESC`.

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -280,7 +280,7 @@ void Battle::DialogBattleSettings( void )
             RedrawBattleSettings( optionAreas );
             saveConfiguration = true;
         }
-        else if ( Game::HotKeyPressEvent( Game::EVENT_DEFAULT_EXIT ) || le.MouseClickLeft( btn_ok.area() ) ) {
+        else if ( HotKeyCloseWindow || le.MouseClickLeft( btn_ok.area() ) ) {
             break;
         }
     }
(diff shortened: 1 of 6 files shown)


a1exsh:
For me personally, convenience is one of the major reasons to play fheroes2 — otherwise I'd just stick with the OG ;)

ihhub:
@a1exsh , thanks for this change! Now it became clearer to me :)

minuspinus:
Maybe I misunderstood this request, but now on many windows where there is only an okay button, you can't close the window by ESC.

minuspinus:
> You close them by default with the return key. This can be remapped in the hotkeys. The default cancel and the default confirm key should be different.

If there is only one okay button in the window, it is not a confirmation.
I.e. when there is a choice - then RETURN is a confirmation (by default).
And when there is only one button, ESC should also work.  

telegram-cloud-photo-size-2-5406739174666661111-y

Otherwise, it violates my user experience. 
On the city screen, I can ESC out of the mages guild, but I can't ESC out of the tavern.

I can get out of the settings menu (both ESC and RETURN), but I can't get out of the tavern (only RETURN)
telegram-cloud-photo-size-2-5406739174666661118-y
telegram-cloud-photo-size-2-5406739174666661119-y

minuspinus:
> @minuspinus. You need to look at the buttons. Buttons that say Okay are triggered with one key, buttons that say Cancel/Exit with another.

How should the hotkeys be set up so I can close the tavern window and ESC and RETURN? As it is done with the settings menu.

image

PR #4452, opened 2021-10-18
First Draft Norwegian Translation

zenseii:
Hi and first thank you so much for this great project, again!

I've been working on this for a while and figured I should share it so that it is available publicly, especially now since I've translated all the strings that I found possible to translate right now and I've  checked it several times.

This contains the `.po`  translation file of the first draft of my translation of fheroes2 to Norwegian.

The makefile in `files\lang` was changed so that in the compilation to `.mo` the locale-specific letters `Æ, Ø,  Å` get changed to `Ae, Oe , Aa`, which are the correct ways to write these letters when you lack these specific letters used in Norwegian. `É` is also simply changed to `E` because the current general method present in the makefile changes it to `'E` with the dash accent which is not good in the case of Norwegian.

`Translations.cpp` was changed to make use of the plural forms of the nouns in ngettext.

Translations have been done in an as literal and faithful way as possible, but inevitably differences due to the natures of the languages are present. Encyclopedias, dictionaries and input from other native speakers have assisted in the translation.
This will obviously still need polish and further attention from others.

<details>
<summary>Some more indepth discussion of the translation and perspectives/issues (press to drop down)</summary>
<br>

> I am aware of the fact that the original English version uses upper case letters for most nouns and in titles which is valid for writing in especially older more fantasy English (_Ex. Sword of Might)_. Meanwhile, this Norwegian translation in its current state does not follow this pattern because this looks very very bad in Norwegian (Ex. Instead of _Sverdet av Makt_ with a capital M I wrote _Sverdet av makt_). To put it in other words you would probably get a D (where A is best) as a grade if you write like in English in a paper in high school or in years even below that. 
> 
> **Having said that** I am understanding and agreeing of the sentiment to keep things as standardized between languages as possible in this project. My personal preference would be to keep the words in lower case, because this keeps the flow of reading natural and fluid/fast. This also makes the translation feel more like an authentic Norwegian tranlation to the game, in my opinion. But I am not stubborn and will listen to reason.
> 
> Some restrictions with the translation currently, apart from the above-mentioned letters and accents, are for instance the use of definite nouns which work differently in Norwegian compared to English and especially that the definite article “the” does not exist in the same way but is a suffix to the noun itself. 
> 
> Ex. `A mighty sword`<-> `Et mektig sverd` ; ` The mighty sword `<-> ` Det mektige sverdet` . 
> 
> Notice that the adjective (`mektig`->`mektigE`) is changed as well. This means that many artifacts have their names in a definite form rather than indefinite, but having gone back and forth several times I found this to be the currently best approach in terms of grammar and readability.
> 
> There are also 3 genders: feminine, masculine and neutrum, which creates interesting problems when translating, but generally I have solved this in the translation. A better way than the currently applied one most likely exists and hopefully someone can share some ideas.
> 
> Maybe a way would be to make it possible to define the definite case of a noun so that it can be specifically called whenever needed, in a somewhat similar fashion to how plural forms are specified right now.
> 

</details>

PS. The correct ISO standard abbreviation of the written language variant of Norwegian I used in this translation, called Bokmål, is `nb`, which is why this is used rather than the seemingly more obvious choice of `no`. This is in case someone down the line would like to translate to the other variant called Nynorsk `nn`. I doubt this, but at least this future-proofs it.

Hopefully this can prove helpful for the translations in the other Northern-Germanic languages, like the currently available incomplete Swedish one and the Danish template.


zenseii:
The PR is now up-to-date with my last changes along with the fixes from your feedback. I changed the sizes ,ex. (36 x 36), to have them just the same as the English way, but without the /n since I believe adding this is something that can be done as a general change among all the translations along with changing them to (XX x XX). If this is fine then for me this PR is complete.

ihhub:
@zenseii , thank you so much for your contribution to the project!

PR #4454, opened 2021-10-18
Consider the protected tile as an obstacle when recalculating the hero's path

### LINKED ISSUE #4430: Hero attacked by spawned monsters
### Preliminary checks

### Platform

Linux

### Describe the bug

End turn, the Trolls are spawned on the map.  Sarakin continues movement, but gets ambushed by the group of Trolls in the woods.

In this case the Trolls are spawned on the road and everything works fine:
image

When they are spawned two tiles to the East, they ambush the hero.

### Save file

[Good_3_0028.zip](https://github.com/ihhub/fheroes2/files/7353623/Good_3_0028.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4430

### DIFF
--- a/src/fheroes2/heroes/route.cpp
+++ b/src/fheroes2/heroes/route.cpp
@@ -428,7 +428,7 @@ bool StepIsObstacle( const Route::Step & s )
         break;
     }
 
-    // consider the protected tile is an obstacle because the hero will not be able to step on it without a battle
+    // consider the protected tile as an obstacle because the hero will not be able to step on it without a battle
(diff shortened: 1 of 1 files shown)


ihhub:
I think we need to test AI behavior with this change to make sure that AI doesn't become dumber...

oleg-derevenetz:
Hi @ihhub, I suggest to merge this, and if AI gets dumber (although I doubt it), we will investigate the cause.

ihhub:

PR #4482, opened 2021-10-23
Use original alignment for Rating, Map Size in gameinfo window

zenseii:
Original:
dossceninfolines
Current fheroes2:
fhersceninfo
Fixed:
fixedfheroes2alginment

The `PO` files will need to reflect this change:
```
msgid ""
" \n"
"Rating"
msgstr ""

msgid ""
" \n"
"Map Size"
msgstr ""
```
This fixes potential issues with longer-worded translations:
Before/After                                                                                      
misalignedfheroes2         fixtranslatedgameinfo


ihhub:
Hi @zenseii , we must update all translation files with this changes also can you please explain why we added a space before \n symbol?

zenseii:
> Hi @zenseii , we must update all translation files with this changes also can you please explain why we added a space before \n symbol?

Yes, when I wrote `_("\nRating")` the string would not break to the next line and it lead to this:
image

I believe the preceeding line needs to contain something to be able to break to the next line. `\n\n` neither moves it down. Not even by one line.

zenseii:
Hi, @ihhub! All changes have been committed and this is ready for your review and potenial merge.

ihhub:

Issue #3346, opened 2021-05-05
Hero Meeting screen, small problems with exchange buttons


LeHerosInconnu:
There are small problems with the shadows of the arrows:
- the shadows are not updating correctly (there is a fault when the button is released), it's quite visible with the top arrow.
- there seems to be several shadows for the arrows (a shadow which is "stuck" to the arrow and which moves with the arrow), it is well visible with the bottom arrow.
- both arrows can be activated at the same time.

In fheroes2:

https://user-images.githubusercontent.com/43583994/117145913-f2408d00-adb3-11eb-8bb8-201f3cdcf512.mp4

![vlcsnap-2021-05-05-14h47m24s421](https://user-images.githubusercontent.com/43583994/117145573-96760400-adb3-11eb-95ac-c5a5bb7d07d6.png)

![vlcsnap-2021-05-05-14h47m41s237](https://user-images.githubusercontent.com/43583994/117145603-9ece3f00-adb3-11eb-9615-67ca2d406d5b.png)

![vlcsnap-2021-05-05-14h48m11s170](https://user-images.githubusercontent.com/43583994/117145630-a55cb680-adb3-11eb-91f3-c5bead054255.png)

Open the three images in different tabs in your browser, enlarge the images and switch between the three tabs alternately.

![vlcsnap-2021-05-05-14h47m24s421 02](https://user-images.githubusercontent.com/43583994/117146021-0edcc500-adb4-11eb-8c8a-bb862a2a091c.png) ![vlcsnap-2021-05-05-14h47m41s237 02](https://user-images.githubusercontent.com/43583994/117146026-0f755b80-adb4-11eb-9e5a-391a237cefab.png) ![vlcsnap-2021-05-05-14h48m11s170 02](https://user-images.githubusercontent.com/43583994/117146031-100df200-adb4-11eb-8e4b-e3b74f15160c.png)

PR #4518, opened 2021-10-29
Fix heroes meeting swap arrows

### LINKED ISSUE #4503: Meeting heroes, swap army / artifact buttons rendering
### Preliminary checks

### Platform

Linux

### Describe the bug

The button sprites include a piece of background and their areas are overlapping.  The button which is rendered last overwrites pieces of the other button:
- normal state, lower button's background cut into the shadow of the upper button:
   image
- after pressing the upper button, the shadow is fully rendered again, but it cuts a few pixels from the lower button itself:
   image

Another small issue is that the arrow width is defined incorrectly in the code: it should be 44 pixels instead of 43.

The code for creating these buttons from original resources is duplicated for left vs. right arrow, but is almost identical, could be reduced in half: https://github.com/ihhub/fheroes2/blob/eaba68a729cda59dff4d54baea801b72e047cec5/src/fheroes2/agg/agg_image.cpp#L796-L886

### Save file

No save file needed: just let any two heroes meet.

### Additional info

Discovered while working on #4350.  I will try to come up with a fix for this issue first.

a1exsh:
Fix #4503, #3346

There is still a slight issue with rendering of background for the bottom two arrows (for artifact exchange), but it doesn't seem like we have any better patch in the background image we're using (`STONEBAK`):

image


ihhub:
Hi @ LeHerosInconnu and @Branikolog , could you please take a look at these changes?

a1exsh:
@ihhub let's merge the shadows PR first — I would like to reuse the code here as well, as here we are also working with shadows.

LeHerosInconnu:
> Hi @LeHerosInconnu and @Branikolog , could you please take a look at these changes?

> @ihhub let's merge the shadows PR first — I would like to reuse the code here as well, as here we are also working with shadows.

I'll wait for the shadows PR to be merged then. :)

a1exsh:
@LeHerosInconnu , @Branikolog please have a look when you have time: the code is updated to use the new function from the other shadows PR ;)

LeHerosInconnu:
Hello @a1exsh,

> @LeHerosInconnu , @Branikolog please have a look when you have time: the code is updated to use the new function from the other shadows PR ;)

The first time the top arrow for creatures is pressed, there is a defect that appears under the bottom arrow for creatures when the button is released.
And each time the bottom arrow for creatures is pressed and released there is a defect that appears under the bottom arrow for creatures.

Also, if immediately after the meeting a troop is selected, there is a default that appears under the bottom arrow for creatures.

In fheroes2:

https://user-images.githubusercontent.com/43583994/144718225-64c7fe42-b898-4665-8bda-e402fb2903ba.mp4

https://user-images.githubusercontent.com/43583994/144718233-b0966b08-5a66-4dcd-8825-97e499e48854.mp4

a1exsh:
@LeHerosInconnu nicely spotted — I'll have another look!

a1exsh:
@LeHerosInconnu fixed

LeHerosInconnu:
@a1exsh.

> @LeHerosInconnu fixed

That was a quick change.
Works like a charm.
Good job. :)

Branikolog:
Hello, @a1exsh 
I've noticed, that in this PR arrows run away in different directions... 
Here's a comparison with OG:
ezgif com-gif-maker (27)

Also, We've problems with background near artifact arrows.. But this was the case before PR.

LeHerosInconnu:
Hello @Branikolog, @a1exsh and @ihhub,

> Hello, @a1exsh I've noticed, that in this PR arrows run away in different directions...

Oh, I didn't see that, I was too focused on the graphic glitches. :D
Good catch!

Although these arrows are not properly centered in the original game either. :)
Before they were part of the background, maybe we should center them properly now that they are detached from the background?
What do you think about it?

a1exsh:
> I've noticed, that in this PR arrows run away in different directions...

We pretty much have to do that so that the button areas do not overlap, which was causing the originally reported issues.

> Also, We've problems with background near artifact arrows.. But this was the case before PR.

Yes.  Unfortunately, I haven't found a suitable resource in the OG to make it look any better.

a1exsh:
@LeHerosInconnu 
> Before they were part of the background, maybe we should center them properly now that they are detached from the background?

I did try to center them.  Do you see a problem with the latest version?

LeHerosInconnu:
@a1exsh.

> @LeHerosInconnu
> 
> > Before they were part of the background, maybe we should center them properly now that they are detached from the background?
> 
> I did try to center them. Do you see a problem with the latest version?

Indeed, the bottom ones are centered as much as possible, good.
The top ones are not centered vertically, see in relation to the red frame.

In fheroes2:
(From the screenshot above.)
Meeting yellow arrows cadre rouge 01

a1exsh:
> The top ones are not centered vertically, see in relation to the red frame.

I was going to write that we cannot move them down further, but then I had an idea.  Let me try it :D

a1exsh:
Centered the army swap arrows (the shadow sticks below, but I think this is how it should be):
image

LeHerosInconnu:
@a1exsh.

> Centered the army swap arrows (the shadow sticks below, but I think this is how it should be):
> image

Yes, it's better to center the yellow arrows without taking their shadows into account.
All seems good now. :)

LeHerosInconnu:
@a1exsh.

I did a check with the last change, I think it's good. :)

ihhub:
@a1exsh , a bunch of thanks for these fixes!

PR #4531, opened 2021-11-01
Do not reset music type to MUSIC_MIDI_ORIGINAL too early

oleg-derevenetz:
At the moment when `Settings::Read()` is called, `AGGInitializer` doesn't exist yet, so `isPriceOfLoyaltySupported()` always returns `false` and therefore `music = expansion` setting from the `fheroes2.cfg` is never read correctly.


ihhub:
@oleg-derevenetz , thank you very much for fixing this issue!

PR #4532, opened 2021-11-01
Add buttons in castle construction window

### LINKED ISSUE #4335: Castle construction window, "show next/previous town" buttons can be useful here
### Preliminary checks

To switch between castles we need to cancel the construction screen.
image

image

This can be useful, if player is going to proceed a construction of several buildings, but is not sure, whether he has enough resources. With this buttons player can check another towns in a few seconds.

We have enough space and similar design of this window to add here the same buttons, we have at the main window,

### Additional info

_No response_

ihhub:
close #4335

Also reduce memory usage for castle window and fix swap and meeting buttons during hero fading animation.

This is how it looks:
image

### DIFF
--- a/src/fheroes2/castle/buildinginfo.cpp
+++ b/src/fheroes2/castle/buildinginfo.cpp
@@ -20,6 +20,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include <cassert>
+
 #include "buildinginfo.h"
 #include "agg.h"
 #include "agg_image.h"
@@ -56,6 +58,8 @@ namespace
         case Race::NECR:
             return fheroes2::Point( 35, 10 );
         default:
+            // Did you add a new race?
+            assert( 0 );
             return fheroes2::Point();
         }
     }
@@ -258,7 +262,7 @@ int GetIndexBuildingSprite( u32 build )
     return 0;
 }
 
-BuildingInfo::BuildingInfo( const Castle & c, building_t b )
+BuildingInfo::BuildingInfo( const Castle & c, const building_t b )
     : castle( c )
     , building( b )
     , area( 0, 0, 135, 70 )
--- a/src/fheroes2/castle/buildinginfo.h
+++ b/src/fheroes2/castle/buildinginfo.h
@@ -31,7 +31,7 @@ class StatusBar;
 class BuildingInfo
 {
 public:
-    BuildingInfo( const Castle &, building_t );
+    BuildingInfo( const Castle & c, const building_t b );
 
     uint32_t getBuilding( void ) const;
     void SetPos( s32, s32 );
--- a/src/fheroes2/castle/castle.h
+++ b/src/fheroes2/castle/castle.h
@@ -112,8 +112,8 @@ public:
         Close, // Close the dialog.
         NextCastle, // Open main dialog of the next castle.
         PreviousCastle, // Open main dialog of the previous castle.
-        NextTown, // Open building purchase dialog of the next castle.
-        PreviousTown // Open building purchase dialog of the previous castle.
+        NextCostructionWindow, // Open construction dialog of the next castle.
+        PreviousCostructionWindow // Open construction dialog of the previous castle.
     };
 
     Castle();
@@ -178,7 +178,7 @@ public:
 
     void DrawImageCastle( const fheroes2::Point & pt ) const;
 
-    CastleDialogReturnValue OpenDialog( const bool readonly );
+    CastleDialogReturnValue OpenDialog( const bool readOnly, const bool openConstructionWindow );
 
     int GetAttackModificator( const std::string * ) const;
     int GetDefenseModificator( const std::string * ) const;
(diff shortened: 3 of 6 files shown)


ihhub:
Hi @Branikolog , could you please test these changes thoughtfully?

Branikolog:
Hi, @ihhub.
Town switch works OK.
The only thing I've noticed: there's no info "pop ups" while right clicking on these buttons.
image

ihhub:
Hi @Branikolog , I've added missing info dialogs for newly added buttons.

oleg-derevenetz (approved):

LeHerosInconnu:
Hello @ihhub and @Branikolog,

Wouldn't it be possible to make it so that when the user starts cycling castles from the castle construction screen and the next city is a town, if the user uses the next (or previous) button it continues to go to the castle construction screen (instead of the town "home" screen) until the player presses the "EXIT" button to exit the cycle of construction screens (or exit the home town screen when it is a town).

So, castle 01 construction screen, castle 02 construction screen, town 03 home screen, castle 04 construction screen (instead of castle 04 home screen), castle 05 construction screen, town 06 home screen, town 07 home screen, castle 08 construction screen, etc.
The cycle is only interrupted when the user presses the "EXIT" button in a castle construction screen.
If the player presses the "EXIT" button in a town home screen, the cycle is interrupted and the player also exits the town home screen and returns to the adventure map as usual.
If the player upgrades a town into a castle and uses the previous or next button, the cycle continues.
If the player after upgrading a town to a castle goes to another castle and then returns to the castle he just built (with the next button and then the previous button, or the other way around), the construction screen of this new castle is displayed.
The player, after upgrading a town to a castle, can also access the construction screen of that castle as usual, without interrupting the cycle.

Branikolog:
Hi, @LeHerosInconnu 
In my opinion using buttons within same window should switch between exactly the same windows: it would be rather confusing, if player, while pressing "next" button in a town view will jump back to construction screen. I find this behaviour inconsistent. Definitely going back to construction screen is more convenient, but the logic will be complex and non-obvious to players.

LeHerosInconnu:
@Branikolog.

> Hi, @LeHerosInconnu In my opinion using buttons within same window should switch between exactly the same windows: it would be rather confusing, if player, while pressing "next" button in a town view will jump back to construction screen. I find this behaviour inconsistent. Definitely going back to construction screen is more convenient, but the logic will be complex and non-obvious to players.

Imagine the situation for two minutes.
The player opens the construction screen to review his castles and build a building when possible, this may even take several "turns" of review depending on priorities and available resources.
The player begins cycling through the castles and arrives at a town, which takes him out of the "cycling loop", forcing him to re-enter the "cycling loop" as soon as he enters a castle again.
Now imagine several towns interspersed between several castles and it gets really annoying.
The function then becomes more annoying to use than it is beneficial, because it is done "halfway".

Branikolog:
@LeHerosInconnu 
Agree with you. Regarding convenience - your proposal looks quite relevant. But at the same time this brings some sort of mess with switching: what if player performs any action after being switched to town and then presses "next" button again.
I understand your logic completely. But I'm deeply convinced, that once we moved back to the main window, "Next town" "Previous town" should work initially as they should at this window.
Just imagine this:
we're switching for another town...
image
image
Close your eyes. Open your eyes.
What action do you expect from the button being pressed at this window?
image

LeHerosInconnu:
@Branikolog.

> @LeHerosInconnu Agree with you. Regarding convenience - your proposal looks quite relevant.

And that's the purpose of this feature, yes?

> But at the same time this brings some sort of mess with switching: what if player performs any action after being switched to town and then presses "next" button again.

All the actions that a player can perform in the town home screen should be possible.
And then pressing the next button should continue what has been started, cycling through the castle construction screens.

> I understand your logic completely. But I'm deeply convinced, that once we moved back to the main window, "Next town" "Previous town" should work initially as they should at this window. Just imagine this: we're switching for another town...
> Close your eyes. Open your eyes. What action do you expect NOW from the button being pressed at this window?

The player is intending to access all the castle construction screens, it seems logical and expected to continue this.

ihhub:
Hi @LeHerosInconnu and @Branikolog , I think both cases are not ideal and both are acceptable. Regarding implementation for switching back to construction screen: it's possible but not trivial. What I did was the easiest solution at the moment.

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu and @Branikolog , I think both cases are not ideal and both are acceptable. Regarding implementation for switching back to construction screen: **it's possible but not trivial. What I did was the easiest solution at the moment.**

So you chose the dark side. :rofl:

Luke: "Vader. Is the dark side stronger?"

Yoda: "No, no, no. Quicker, easier, more seductive."

a1exsh:
What if we display the "Build castle" dialog when switching from a castle to a town with the new arrow buttons?  It will be immediately clear what's going on, IMO.

We might want to adjust the dialog's event loop so that it can also be switched to the next/previous town using the arrows — not sure how ugly that's going to be in the code, though.

LeHerosInconnu:
Hello @a1exsh,

> What if we display the "Build castle" dialog when switching from a castle to a town with the new arrow buttons? It will be immediately clear what's going on, IMO.

You mean replace "Show next town" and "Show previous town" by "Build next town" and "Build previous town" in the info bar at the bottom of the screen when the player positions the mouse cursor over the new buttons.
Brilliant and simple. :)

> We might want to adjust the dialog's event loop so that it can also be switched to the next/previous town using the arrows — not sure how ugly that's going to be in the code, though.

When the player accesses the home screen of a city from the adventure map, the info bar at the bottom of the screen shows "Show next town" and "Show previous town" when the player positions the mouse cursor over the right and left arrow buttons.

When the player accesses the city building screen, the info bar at the bottom of the screen shows "Build next town" and "Build previous town" when the player positions the mouse cursor over the left and right arrow buttons.

The town screens are considered to be "home" and "construction" at the same time, it is the info bar that tells the player if he is in a "construction screen" or "home screen" loop.

Yes, I want this. :)

a1exsh:
@LeHerosInconnu what I mean is that I would like a sequence of screens like this (all by clicking the next town arrow button):

image

Screenshot from 2021-11-08 08-51-53

image

Right now the middle screen will open without any indication of the "mode" we've started with, which breaks the sequence: the Knight castle will also open in a mode different from the first one.

I hope that makes it clear? :)

LeHerosInconnu:
Hello @a1exsh,

> @LeHerosInconnu what I mean is that I would like a sequence of screens like this (all by clicking the next town arrow button):
> 
> image
> 
> Screenshot from 2021-11-08 08-51-53
> 
> image
> 
> Right now the middle screen will open without any indication of the "mode" we've started with, which breaks the sequence: the Knight castle will also open in a mode different from the first one.
> 
> I hope that makes it clear? :)

Okay, I see. :)
But in this case, could the user continue to press the previous and next buttons without having to close the town to castle upgrade window?
If it is possible, then I think it can work. :)

a1exsh:
> But in this case, could the user continue to press the previous and next buttons without having to close the town to castle upgrade window?

That is the idea, yes.  I just don't know how hard it is to implement that.

PR #4543, opened 2021-11-02
ResetMixer is async by default

### LINKED ISSUE #4463: Fix lag of 0.1 sec when loading a map due to music reset
### Preliminary checks

When loading a new map, we call Mixer::Reset(), this method will synchronously call code from SDL_Mix to stop all sounds.
However on windows with MIDI music, this is notoriously slow to due to several issues on Windows side.
This is was fixed in fheroes2 by using AsyncSoundManager.
However Mixer::Reset doesn't benefit from this and so it will block the main thread for like 0.1 sec while the MIDI file unloads

AsyncSoundManager should also handle tasks such a "stop all music" / "stop all sounds" in async fashion

### Additional info

I can have a look at it

vincent-grosbois:
ResetMixer is suffering from the same issues with music player:
for MIDI files, it will call SDL that will add a lag of 0.1 sec for every
action on MIDI files. This means that when calling ResetMixer during loading,
we waste 0.1 sec as well.

The fix is simply to have the call to Music::Reset be done in the
dedicated audio thread so it doesn't block the main thread.

We can put the "stop music" task into the stack of existing _musicTasks,
because it behaves the same way, ie the logic has to process in priority
the latest task and can discard all the previous ones.

fixes #4463

### DIFF
--- a/src/fheroes2/agg/agg.cpp
+++ b/src/fheroes2/agg/agg.cpp
@@ -113,8 +113,24 @@ namespace AGG
             }
         }
 
+        void pushStopMusic()
+        {
+            _createThreadIfNeeded();
+
+            std::lock_guard<std::mutex> mutexLock( _mutex );
+
+            while ( !_musicTasks.empty() ) {
+                _musicTasks.pop();
+            }
+
+            _musicTasks.emplace( -1, MUSIC_MIDI_ORIGINAL, true );
+            _runFlag = 1;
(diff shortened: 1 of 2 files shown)


vincent-grosbois:
I think the app veyor test is failing for no reason but I don't know how I can re-trigger it

ihhub:
@vincent-grosbois , thank you so much for the fix!

PR #4544, opened 2021-11-04
Fix Stonehenge building requirements

idshibanov:
Bringing building requirements back to the original values. Was reported by users in discord.


ihhub:
@idshibanov , thank you very much for fixing this issue!

PR #4547, opened 2021-11-04
Do not leak the hotkey state from the experimental settings dialog to the scenario info dialog

### LINKED ISSUE #4546: Exiting experimental settings dialog within a game by hotkey closes scenario info window as well
### Preliminary checks

### Platform

Linux

### Describe the bug

1. Start or load a game and open scenario information window:
   image
2. Click 'Config'.  The "Experimental Game Settings" window appears:
   image
3. Hit `ESC` or `ENTER` to close the settings window.
4. **Bug:** both windows are closed.
5. When clicking 'Okay' instead of using a hotkey, only the settings window is closed as expected.

### Save file

No save file needed: just start any game.

### Additional info

1. The same dialog doesn't close the parent window when opened from main menu: no bug is present.
2. I've seen the same problem in at least one other place, but couldn't reproduce it after seeing it once so far.

oleg-derevenetz:
fix #4546

### DIFF
--- a/src/fheroes2/dialog/dialog_gameinfo.cpp
+++ b/src/fheroes2/dialog/dialog_gameinfo.cpp
@@ -125,12 +125,12 @@ void Dialog::GameInfo( void )
         le.MousePressLeft( buttonOk.area() ) ? buttonOk.drawOnPress() : buttonOk.drawOnRelease();
         le.MousePressLeft( buttonCfg.area() ) ? buttonCfg.drawOnPress() : buttonCfg.drawOnRelease();
 
+        if ( le.MouseClickLeft( buttonOk.area() ) || HotKeyCloseWindow )
+            break;
+
         if ( le.MouseClickLeft( buttonCfg.area() ) ) {
             Dialog::ExtSettings( true );
             display.render();
         }
-
-        if ( le.MouseClickLeft( buttonOk.area() ) || HotKeyCloseWindow )
-            break;
     }
 }


ihhub:

src/fheroes2/dialog/dialog_gameinfo.cpp
```diff
@@ -125,12 +125,12 @@ void Dialog::GameInfo( void )
         le.MousePressLeft( buttonOk.area() ) ? buttonOk.drawOnPress() : buttonOk.drawOnRelease();
         le.MousePressLeft( buttonCfg.area() ) ? buttonCfg.drawOnPress() : buttonCfg.drawOnRelease();
 
+        if ( le.MouseClickLeft( buttonOk.area() ) || HotKeyCloseWindow )
```
a1exsh, line 128: Wow, this is subtle!

Is there a way to make it work regardless of the order of `if`s here?  Like resetting the status of the hotkey once it is handled, for example.

oleg-derevenetz: Currently, there is no way to do this, and, moreover, sometimes such resetting is not desirable. All events should be handled with caution.

a1exsh: Well, I see the point, because e.g. MouseClick resets automatically and it's confusing, but we could reset events unconditionally when closing a dialog, for example.

oleg-derevenetz: @a1exsh

> but we could reset events unconditionally when closing a dialog, for example.

There is `LocalEvent::GetClean()` method that clears the current state of a shared event (except `MOUSE_WHEEL` events for some reason), but it should be used with caution.

PR #4549, opened 2021-11-04
Fix Stonehenge gold cost

zenseii:
The Stonehenge costs 2500 gold in OG:

image


ihhub:
@zenseii , thanks a lot for the fix!

PR #4553, opened 2021-11-04
Allow to use mouse wheel to change system options

a1exsh:
Mouse wheel is enabled on the following options:

- Music
- Effects
- Hero Speed
- AI Speed
- Scroll Speed

Any adjustments made using the wheel do not wrap over max. value, as opposed
to clicking.  Clicking still works as before.

Also unify the value wrapping math.

### DIFF
--- a/src/fheroes2/dialog/dialog_system_options.cpp
+++ b/src/fheroes2/dialog/dialog_system_options.cpp
@@ -254,7 +254,6 @@ namespace
 
         display.render();
 
-        bool redraw = false;
         bool saveConfig = false;
 
         // dialog menu loop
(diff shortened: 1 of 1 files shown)


src/fheroes2/dialog/dialog_system_options.cpp
```diff
@@ -266,56 +265,102 @@ namespace
...
+            bool saveSoundVolume = false;
+            if ( Audio::isValid() ) {
+                if ( le.MouseClickLeft( musicVolumeRoi ) ) {
+                    conf.SetMusicVolume( ( conf.MusicVolume() + 1 ) % 10 );
+                    saveMusicVolume = true;
+                }
+                else if ( le.MouseWheelUp( musicVolumeRoi ) ) {
+                    conf.SetMusicVolume( std::min( conf.MusicVolume() + 1, 9 ) );
+                    saveMusicVolume = true;
+                }
```
ihhub, line 279: Shouldn't we combine these two conditions (through logical OR) as they're doing the same thing?

I think the same question for the rest of options.

ihhub (changes requested):
Hi @a1exsh , I left one comment here. Could you please take a look?

src/fheroes2/dialog/dialog_system_options.cpp
a1exsh: Not exactly: clicking wraps over max value, while scrolling doesn't.

See the linked discussion, maybe we change the behavior as the outcome, then we could simplify the code here. But I do like the implemented behavior better without wrapping on scroll. ;)

ihhub: Got it. Thanks for the explanation!

ihhub:
@a1exsh , thanks a lot for this improvement!

Issue #4557, opened 2021-11-05
Combat screen, terrain music is played during the combat


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the terrain music is played during the combat.

In fheroes2:

Activate the sound of the video.

https://user-images.githubusercontent.com/43583994/140564637-69d97357-8345-46ad-8326-b3431cf81490.mp4

### Save file

Here is the save file:
[Before combat terrain music_.zip](https://github.com/ihhub/fheroes2/files/7488344/Before.combat.terrain.music_.zip)

### Additional info

Game started with version 0.9.8 4544 and continued with version 0.9.9 4551.

oleg-derevenetz:
Regression after #4543.

vincent-grosbois:
that sucks... I'll try to see if I can fix it without having to revert

oleg-derevenetz:
@vincent-grosbois this PR is not included into 0.9.9, so no hurry here.

vincent-grosbois:
@oleg-derevenetz 
I couldn't reproduce it but I based on the video it seems that when the "prebattle music" plays, the game music isn't stopped
Based on this, it seems that it's due to the call to sync() in AGG::PlaySound that is probably erasing the "stop music" task that couldn't be dequeued in time

zenseii:
I can add that the same happens in Battle Only mode, where you'll keep hearing the Main Menu theme.

oleg-derevenetz:
I believe we can consider this as fixed by #4558.

PR #4558, opened 2021-11-05
Fix terrain music being played during battle

vincent-grosbois:
Here's a tentative fix for #4557 which I couldn't reproduce

I think the issue is as follows:
before this PR, we call PlaySound() with aync = false
This forces a call to "sync()", which will remove all pending music tasks. I think at this point it can potentially cancel the "stop music" tasks generated by the call to Agg::ResetMixer that maybe was not dequeued yet

NB : it's unclear to me why AGG::PlaySound in synch mode will erase all pending async music tasks, seems incorrect IMO

### DIFF
--- a/src/fheroes2/agg/agg.cpp
+++ b/src/fheroes2/agg/agg.cpp
@@ -627,7 +627,7 @@ std::vector<u8> AGG::LoadBINFRM( const char * frm_file )
     return AGG::ReadChunk( frm_file );
 }
 
-void AGG::ResetMixer( bool asyncronizedCall /* = true */ )
+void AGG::ResetMixer( bool asyncronizedCall /* = false */ )
 {
     if ( asyncronizedCall ) {
         g_asyncSoundManager.pushStopMusic();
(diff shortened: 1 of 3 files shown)


oleg-derevenetz:
Hi @vincent-grosbois are you sure that there will be no race condition between playing M82::PREBATTL from background thread and waiting for the completion of this playback here:

https://github.com/ihhub/fheroes2/blob/960d7b594ece35610de746e11586284b70f93046/src/fheroes2/battle/battle_arena.cpp#L288

?

ihhub:
Hi @vincent-grosbois , I missed one point: we shouldn't use default values for input arguments. I'd insist to remove default value for `void ResetMixer( bool asyncronizedCall = true );` function. Or at least the default value should follow the same standard: all other functions have `false` as default value.

This will be somehow a move to revert changes introduced in the previous pull request. Next step is to explicitly call `ResetMixer` in places where it's safe to do so.

vincent-grosbois:
@oleg-derevenetz  yes I'm quite sure it's unrelated

@ihhub  ok
I've pushed a new version of this commit where I turned everything "sync" by default except during map loading

ihhub:
> 
> 
> @oleg-derevenetz yes I'm quite sure it's unrelated
> 
> @ihhub ok I've pushed a new version of this commit where I turned everything "sync" by default except during map loading

Thanks!

ihhub:
@vincent-grosbois , thank you so much for fixing this issue!

PR #4560, opened 2021-11-06
Set starting army in Archibald 7 scenario

### LINKED ISSUE #4379: Archibald 7 scenario missing starter bonus
### Preliminary checks

### Platform

Windows

### Describe the bug

In the OG in this scenario player should have 5th level hero (with 5000 exp. plus random secondary skills) and some army.
For Necromancers:
image
For Warlocks:
image
For Barbarians:
image

### Save file

[Evil_6_Complete.zip](https://github.com/ihhub/fheroes2/files/7320773/Evil_6_Complete.zip)

### Additional info

_No response_

a1exsh:
Fix #4379

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -740,8 +740,8 @@ fheroes2::GameMode Game::SelectCampaignScenario( const fheroes2::GameMode prevMo
             // starting faction scenario bonus has to be called before players.SetStartGame()
             if ( scenarioBonus._type == Campaign::ScenarioBonusData::STARTING_RACE || scenarioBonus._type == Campaign::ScenarioBonusData::STARTING_RACE_AND_ARMY ) {
                 // but the army has to be set after starting the game, so first only set the race
-                const Campaign::ScenarioBonusData temp( Campaign::ScenarioBonusData::STARTING_RACE, scenarioBonus._subType, scenarioBonus._amount );
-                SetScenarioBonus( campaignSaveData.getCampaignID(), chosenScenarioID, temp );
+                SetScenarioBonus( campaignSaveData.getCampaignID(), chosenScenarioID,
+                                  { Campaign::ScenarioBonusData::STARTING_RACE, scenarioBonus._subType, scenarioBonus._amount } );
             }
 
             Players & players = conf.GetPlayers();
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -226,66 +226,6 @@ namespace
         }
     }
 
-    void SetScenarioBonus( const Campaign::ScenarioBonusData & scenarioBonus )
-    {
-        const Players & sortedPlayers = Settings::Get().GetPlayers();
-        for ( const Player * player : sortedPlayers ) {
-            if ( player == nullptr ) {
-                continue;
-            }
-
-            if ( !player->isControlHuman() )
-                continue;
-
-            Kingdom & kingdom = world.GetKingdom( player->GetColor() );
-            Heroes * bestHero = kingdom.GetBestHero();
-
-            switch ( scenarioBonus._type ) {
-            case Campaign::ScenarioBonusData::RESOURCES:
-                kingdom.AddFundsResource( Funds( scenarioBonus._subType, scenarioBonus._amount ) );
-                break;
-            case Campaign::ScenarioBonusData::ARTIFACT: {
-                Heroes * hero = kingdom.GetBestHero();
-                assert( hero != nullptr );
-                if ( hero != nullptr ) {
-                    hero->PickupArtifact( Artifact( scenarioBonus._subType ) );
-                }
-                break;
-            }
-            case Campaign::ScenarioBonusData::TROOP:
-                kingdom.GetBestHero()->GetArmy().JoinTroop( Troop( Monster( scenarioBonus._subType ), scenarioBonus._amount ) );
-                break;
-            case Campaign::ScenarioBonusData::SPELL: {
-                KingdomHeroes & heroes = kingdom.GetHeroes();
-                assert( !heroes.empty() );
-                if ( !heroes.empty() ) {
-                    // TODO: make sure that the correct hero receives the spell. Right now it's a semi-hacky way to do this.
-                    heroes.back()->AppendSpellToBook( scenarioBonus._subType, true );
-                }
-                break;
-            }
-            case Campaign::ScenarioBonusData::STARTING_RACE:
-                Players::SetPlayerRace( player->GetColor(), scenarioBonus._subType );
-                break;
-            case Campaign::ScenarioBonusData::SKILL_PRIMARY:
-                assert( bestHero != nullptr );
-                if ( bestHero != nullptr ) {
-                    for ( uint32_t i = 0; i < scenarioBonus._amount; ++i )
-                        bestHero->IncreasePrimarySkill( scenarioBonus._subType );
-                }
-                break;
-            case Campaign::ScenarioBonusData::SKILL_SECONDARY:
-                assert( bestHero != nullptr );
-                if ( bestHero != nullptr ) {
-                    bestHero->LearnSkill( Skill::Secondary( scenarioBonus._subType, scenarioBonus._amount ) );
-                }
-                break;
-            default:
-                assert( 0 );
-            }
-        }
-    }
-
     void replaceArmy( Army & army, const std::vector<Troop> & troops )
     {
         army.Clean();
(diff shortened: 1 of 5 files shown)


a1exsh:
Here is just a very dirty hack to make it work.  I don't see a suitable bonus type or an award that could help us here (unless we abuse "carry over troops" somehow).

I'm now wondering if maybe the starting army is part of the map file, but just don't use this data?

zenseii:
> I'm now wondering if maybe the starting army is part of the map file, but we just don't use this data?

I can check this for you quickly with the OG editor if you want?

a1exsh:
> I can check this for you quickly with the OG editor if you want?

How do you open campaign maps in OG editor?  The map file is called `CAMPE07.H2C` and the title is `Evil 7`, which I don't see in the list in my editor.  Is there a trick?

zenseii:
> How do you open campaign maps in OG editor? The map file is called `CAMPE07.H2C` and the title is `Evil 7`, which I don't see in the list in my editor. Is there a trick?

Yep, just rename the .H2C to .mp2 and you're good. So this specific becomes: CAMPE07.mp2

a1exsh:
> Yep, just rename the .H2C to .mp2 and you're good. 

Thanks, will give it a try :D

a1exsh:
Doesn't look like the army is set in the editor — there is just random town and hero, no special army (maybe it's still part of the file, but not something that editor exposes).

ihhub:
> 
> 
> Doesn't look like the army is set in the editor — there is just random town and hero, no special army (maybe it's still part of the file, but not something that editor exposes).

No, the army is set within the game.

a1exsh:
> No, the army is set within the game.

@ihhub alright.  Do you maybe have a suggestion how to organize the code better?  I'm thinking, could we make the starting army an optional field in the `STARTING_RACE` bonus?  We also don't need to save the army to the file, I guess, as it only matters when starting the scenario — the file format can be kept as is.

a1exsh:
Note to self: the starting hero characteristics should be also part of the bonus, not only the army.

ihhub:
> 
> 
> > No, the army is set within the game.
> 
> @ihhub alright. Do you maybe have a suggestion how to organize the code better? I'm thinking, could we make the starting army an optional field in the `STARTING_RACE` bonus? We also don't need to save the army to the file, I guess, as it only matters when starting the scenario — the file format can be kept as is.

Yes, it's good to add a new Bonus type called `STARTING_RACE_AND_ARMY` (or something more human readable) which will have the same code as this:
```cpp
case ScenarioBonusData::STARTING_RACE:
    objectName = Race::String( _subType );
    break;
```

Then add logic inside `void SetScenarioBonus( const Campaign::ScenarioBonusData & scenarioBonus )` and modify `std::vector<Campaign::ScenarioBonusData> getArchibaldCampaignBonusData( const int scenarioID )` function to use newly added type. However, the structure of bonuses doesn't allow us to add multiple monsters as well as hero stats so we hardcode these values within `SetScenarioBonus` (or within another function which is called from this function).

I think it make the code clean.

a1exsh:
> Yes, it's good to add a new Bonus type called STARTING_RACE_AND_ARMY

Tried that.

> I think it make the code clean.

Didn't happen so far :)  There is an additional trouble that setting the race has to be done before starting the game, but adding army to the best player's hero — after.  Either we will end up calling `SetScenarioBonus` twice, or we need to call a different function for setting the army.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -225,7 +226,51 @@ namespace
         }
     }
 
-    void SetScenarioBonus( const Campaign::ScenarioBonusData & scenarioBonus )
+    void replaceArmy( Army & army, const std::vector<Troop> & troops )
+    {
+        army.Clean();
+        for ( uint32_t i = 0; i < troops.size(); ++i )
```
ihhub, line 232: Please use `size_t` here as **size()** returns such type.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -225,7 +226,51 @@ namespace
...
+                case Race::BARB:
+                    startingTroops.emplace_back( Monster::ORC_CHIEF, 12 );
+                    startingTroops.emplace_back( Monster::OGRE, 18 );
+                    startingTroops.emplace_back( Monster::GOBLIN, 40 );
+                    break;
+                default:
+                    assert( 0 ); // bonus changed?
+                }
+                replaceArmy( hero->GetArmy(), startingTroops );
+                uint32_t exp = hero->GetExperience();
```
ihhub, line 263: Please make this variable `const` as we don't modify it.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -225,7 +226,51 @@ namespace
...
+    void replaceArmy( Army & army, const std::vector<Troop> & troops )
+    {
+        army.Clean();
+        for ( uint32_t i = 0; i < troops.size(); ++i )
+            army.GetTroop( i )->Set( troops[i] );
+    }
+
+    void setHeroAndArmyBonus( Heroes * hero, const int campaignID, const uint32_t currentScenarioID )
+    {
+        switch ( campaignID ) {
```
ihhub, line 238: This switch case requires at least **default** case. Please add it with an assertion to remind a future developer to add appropriate logic.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -225,7 +226,51 @@ namespace
...
+        case Campaign::ARCHIBALD_CAMPAIGN:
+            switch ( currentScenarioID ) {
+            case 6:
```
ihhub, line 241: I think switch for this case is not needed. We can do like this:
```cpp
// Only one scenario in Archibald campaign has this bonus.
if ( currentScenarioID != 6 ) {
    return;
}
```

src/fheroes2/game/game_campaign.cpp
```diff
@@ -225,7 +226,51 @@ namespace
...
+                    break;
+                case Race::BARB:
+                    startingTroops.emplace_back( Monster::ORC_CHIEF, 12 );
+                    startingTroops.emplace_back( Monster::OGRE, 18 );
+                    startingTroops.emplace_back( Monster::GOBLIN, 40 );
+                    break;
+                default:
+                    assert( 0 ); // bonus changed?
+                }
+                replaceArmy( hero->GetArmy(), startingTroops );
```
ihhub, line 262: I think we can do like this:
```cpp
switch ( hero->GetRace() ) {
case Race::NECR:
    replaceArmy( hero->GetArmy(), { { Monster::SKELETON, 50 }, { Monster::ROYAL_MUMMY, 18 }, { Monster::VAMPIRE_LORD, 8 } } );
    break;
case Race::WRLK:
    replaceArmy( hero->GetArmy(), { { Monster::CENTAUR, 40 }, { Monster::GARGOYLE, 24 }, { Monster::GRIFFIN, 18 } } );
    break;
case Race::BARB:
    replaceArmy( hero->GetArmy(), { { Monster::ORC_CHIEF, 12 }, { Monster::OGRE, 18 }, { Monster::GOBLIN, 40 } } );
    break;
default:
    assert( 0 ); // bonus changed?
}
```
In this case the lifetime of std:;vector will be much shorter.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -244,28 +289,35 @@ namespace
...
-                KingdomHeroes & heroes = kingdom.GetHeroes();
-                assert( !heroes.empty() );
-                if ( !heroes.empty() ) {
-                    // TODO: make sure that the correct hero receives the spell. Right now it's a semi-hacky way to do this.
-                    heroes.back()->AppendSpellToBook( scenarioBonus._subType, true );
+                // TODO: make sure that the correct hero receives the spell. Right now it's a semi-hacky way to do this.
+                assert( bestHero != nullptr );
+                if ( bestHero != nullptr ) {
+                    bestHero->AppendSpellToBook( scenarioBonus._subType, true );
                 }
```
ihhub, line 309: I think this will break one of scenarios as best hero is not always the last one.

ihhub (changes requested):
Hi @a1exsh , I left several comments here. Would you mind to take a look?

src/fheroes2/game/game_campaign.cpp
a1exsh: Hm, sure, but I wonder why do we have clang-tidy robot and it doesn't report this (but tends to find all sorts of non-issues) ;)

src/fheroes2/game/game_campaign.cpp
a1exsh: Ah, vector literals: that's a really nice trick, thank you! :)

src/fheroes2/game/game_campaign.cpp
a1exsh: OK, will change it back.  I just had the impression that all cases in this function were added independently without looking at the rest, so it became quite messy: tried to clean it up.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -696,8 +738,11 @@ fheroes2::GameMode Game::SelectCampaignScenario( const fheroes2::GameMode prevMo
             conf.SetCurrentFileInfo( mapInfo );
 
             // starting faction scenario bonus has to be called before players.SetStartGame()
-            if ( scenarioBonus._type == Campaign::ScenarioBonusData::STARTING_RACE )
-                SetScenarioBonus( scenarioBonus );
+            if ( scenarioBonus._type == Campaign::ScenarioBonusData::STARTING_RACE || scenarioBonus._type == Campaign::ScenarioBonusData::STARTING_RACE_AND_ARMY ) {
+                // but the army has to be set after starting the game, so first only set the race
+                const Campaign::ScenarioBonusData temp( Campaign::ScenarioBonusData::STARTING_RACE, scenarioBonus._subType, scenarioBonus._amount );
```
ihhub, line 743: Just a small correction: should we name this variable as **tempBonus**?

a1exsh: I've removed it completely: no need to improve the name if it doesn't have one ;)

ihhub:
@a1exsh , good job and keep it up!

PR #4561, opened 2021-11-06
Use the path from the AI pathfinder when determining the nearest reachable cell instead of simply comparing distances

### LINKED ISSUE #4559: Combat AI hangs for pathfinding
### Preliminary checks

### Platform

Windows

### Describe the bug

Pathfinding doesn't work on AI turn leading to infinite loop in the battle. Open fheroes2 in debug mode showing battles and observe the hang. It looks like AI doesn't take into account objects on the battle.

Here is a screenshot of the situation when AI hangs:
image

Once Archers are out of arrow they'll hit Dwarves and Dwarves with Golems will stand in infinite loop.

### Save file

[Freeze.zip](https://github.com/ihhub/fheroes2/files/7489402/Freeze.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4559

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -256,7 +256,7 @@ namespace AI
             DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Melee phase end, targetCell is " << target.cell );
 
             if ( target.cell != -1 ) {
-                const int32_t reachableCell = Board::FindNearestReachableCell( currentUnit, target.cell );
+                const int32_t reachableCell = arena.GetNearestReachableCell( currentUnit, target.cell );
 
                 DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Nearest reachable cell is " << reachableCell );
 
@@ -453,7 +453,7 @@ namespace AI
                 if ( target.cell != -1 ) {
                     DEBUG_LOG( DBG_BATTLE, DBG_INFO, currentUnit.GetName() << " archer kiting enemy, moving to " << target.cell );
 
-                    const int32_t reachableCell = Board::FindNearestReachableCell( currentUnit, target.cell );
+                    const int32_t reachableCell = arena.GetNearestReachableCell( currentUnit, target.cell );
 
                     DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Nearest reachable cell is " << reachableCell );
 
@@ -700,7 +700,7 @@ namespace AI
                     if ( targetCell != -1 ) {
                         DEBUG_LOG( DBG_BATTLE, DBG_INFO, currentUnit.GetName() << " is under Berserk spell, moving to " << targetCell );
 
-                        const int32_t reachableCell = Board::FindNearestReachableCell( currentUnit, targetCell );
+                        const int32_t reachableCell = arena.GetNearestReachableCell( currentUnit, targetCell );
 
                         DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Nearest reachable cell is " << reachableCell );
 
--- a/src/fheroes2/battle/battle_arena.cpp
+++ b/src/fheroes2/battle/battle_arena.cpp
@@ -672,6 +672,37 @@ Battle::Indexes Battle::Arena::getAllAvailableMoves( uint32_t moveRange ) const
     return _pathfinder.getAllAvailableMoves( moveRange );
 }
 
+int32_t Battle::Arena::GetNearestReachableCell( const Unit & currentUnit, const int32_t dst ) const
+{
+    const Position dstPos = Position::GetReachable( currentUnit, dst );
+
+    if ( dstPos.GetHead() != nullptr && ( !currentUnit.isWide() || dstPos.GetTail() != nullptr ) ) {
+        // Destination cell is already reachable
+        return dstPos.GetHead()->GetIndex();
+    }
+
+    const Indexes path = _pathfinder.buildPath( dst );
+
+    // Destination cell is unreachable in principle according to the ArenaPathfinder
+    if ( path.empty() ) {
+        return -1;
+    }
+
+    // Search for the reachable cell nearest to the end of the path
+    const Cell * nearestCell = nullptr;
+
+    for ( auto it = path.crbegin(); it != path.crend(); ++it ) {
+        const Position pos = Position::GetReachable( currentUnit, *it );
+
+        if ( pos.GetHead() != nullptr && ( !currentUnit.isWide() || pos.GetTail() != nullptr ) ) {
+            nearestCell = pos.GetHead();
+            break;
+        }
+    }
+
+    return nearestCell ? nearestCell->GetIndex() : -1;
+}
+
 Battle::Unit * Battle::Arena::GetTroopBoard( s32 index )
 {
     return Board::isValidIndex( index ) ? board[index].GetUnit() : nullptr;
--- a/src/fheroes2/battle/battle_arena.cpp
+++ b/src/fheroes2/battle/battle_arena.cpp
@@ -689,18 +689,15 @@ int32_t Battle::Arena::GetNearestReachableCell( const Unit & currentUnit, const 
     }
 
     // Search for the reachable cell nearest to the end of the path
-    const Cell * nearestCell = nullptr;
-
     for ( auto it = path.crbegin(); it != path.crend(); ++it ) {
         const Position pos = Position::GetReachable( currentUnit, *it );
 
         if ( pos.GetHead() != nullptr && ( !currentUnit.isWide() || pos.GetTail() != nullptr ) ) {
-            nearestCell = pos.GetHead();
-            break;
+            return pos.GetHead()->GetIndex();
         }
     }
 
-    return nearestCell ? nearestCell->GetIndex() : -1;
+    return -1;
 }
 
 Battle::Unit * Battle::Arena::GetTroopBoard( s32 index )
--- a/src/fheroes2/battle/battle_arena.h
+++ b/src/fheroes2/battle/battle_arena.h
@@ -124,6 +124,10 @@ namespace Battle
         Indexes CalculateTwoMoveOverlap( int32_t indexTo, uint32_t movementRange = 0 ) const;
         Indexes GetPath( const Unit &, const Position & ) const;
 
+        // Returns the cell nearest to the end of the path to the cell with the given index (according to the ArenaPathfinder)
+        // and reachable for the current unit (to which the current board passability information relates)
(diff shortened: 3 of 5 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , looks good to me! Thank you!

Before proceeding with merging we need an extensive test of these changes. @LeHerosInconnu and @Branikolog , guys, could you please help to test these changes? Essentially we need to have no AI freezes during battles. If you need a compiled debug version of the game I can help.

oleg-derevenetz:
@ihhub I believe that current AI implementation is in principle vulnerable to the "infinite battle" issues. Suppose that one side has a weak, but fast unit and selects a defensive strategy (elude the attack), while the other side has a strong, but slow unit. In this case battle may take an infinite amount of turns. I suppose we should impose a limit on the number of moves in AI vs AI battles (not in this PR, of course).

LeHerosInconnu:
Hello @ihhub,
 
> Before proceeding with merging we need an extensive test of these changes. @LeHerosInconnu and @Branikolog , guys, could you please help to test these changes? Essentially we need to have no AI freezes during battles. If you need a compiled debug version of the game I can help.

So yet another boring test. :D

How to test this efficiently?
Is it okay to attack wandering troops and play the combat in automatic mode?
In this case saves can be made before a combat.
Or rather run a scenario in debug mode and watch the AI play?
I think a recent debug version of the game would be necessary, so make it available.

oleg-derevenetz:
Hi @LeHerosInconnu

> So yet another boring test. :D
> 
> How to test this efficiently? Is it okay to attack wandering troops and play the combat in automatic mode? In this case saves can be made before a combat. Or rather run a scenario in debug mode and watch the AI play? I think a recent debug version of the game would be necessary, so make it available.

I suppose you can just play the game as usual, and, if it freezes, it will be noticeable :) Such freezes may occur only in AI vs AI battles, so there is no need to make saves before each combat. AUTOSAVE should be enough.

LeHerosInconnu:
Hello @oleg-derevenetz,

> Hi @LeHerosInconnu
> 
> > So yet another boring test. :D
> > How to test this efficiently? Is it okay to attack wandering troops and play the combat in automatic mode? In this case saves can be made before a combat. Or rather run a scenario in debug mode and watch the AI play? I think a recent debug version of the game would be necessary, so make it available.
> 
> I suppose you can just play the game as usual, and, if it freezes, it will be noticeable :) Such freezes may occur only in AI vs AI battles (or in auto/instant battles).

So playing a scenario with the build program of this PR is enough?
Is it possible to continue a scenario already started with the standard program, or is it better to start a scenario from the beginning?

oleg-derevenetz:
@LeHerosInconnu

> So playing a scenario with the build program of this PR is enough?

I suppose it would be better to play as usual, but use a debug build with `DBG_BATTLE` turned on (so you will see all battles, even AI vs AI, I suppose, @ihhub is able to provide such a build) to be able to make a screenshot of a freeze and describe in general what to do to make this particular battle happen.

> Is it possible to continue a scenario already started with the standard program, or is it better to start a scenario from the beginning?

No, there is no need to start a new scenario, you can continue to play any existing scenario.

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu
> 
> > So playing a scenario with the build program of this PR is enough?
> 
> I suppose it would be better to play as usual, but use a debug build with `DBG_BATTLE` turned on (so you will see all battles, even AI vs AI, I suppose, @ihhub is able to provide such a build) to be able to make a screenshot of a freeze and describe in general what to do to make this particular battle happen.
> 
> > Is it possible to continue a scenario already started with the standard program, or is it better to start a scenario from the beginning?
> 
> No, there is no need to start a new scenario, you can continue to play any existing scenario.

Okay, thanks.
I'll wait for the debug build then and I'll start a test tomorrow.

oleg-derevenetz:
@LeHerosInconnu

> I'll wait for the debug build then and I'll start a test tomorrow.

Thank you very much for your invaluable support of this project.

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -453,7 +453,7 @@ namespace AI
                 if ( target.cell != -1 ) {
                     DEBUG_LOG( DBG_BATTLE, DBG_INFO, currentUnit.GetName() << " archer kiting enemy, moving to " << target.cell );
 
-                    const int32_t reachableCell = Board::FindNearestReachableCell( currentUnit, target.cell );
+                    const int32_t reachableCell = arena.GetNearestReachableCell( currentUnit, target.cell );
```
ihhub, line 456: Shouldn't we check if **reachableCell** is not a negative value in the condition below?

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -256,7 +256,7 @@ namespace AI
             DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Melee phase end, targetCell is " << target.cell );
 
             if ( target.cell != -1 ) {
-                const int32_t reachableCell = Board::FindNearestReachableCell( currentUnit, target.cell );
+                const int32_t reachableCell = arena.GetNearestReachableCell( currentUnit, target.cell );
 
                 DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Nearest reachable cell is " << reachableCell );
```
ihhub, line 261: Shouldn't we check if **reachableCell** is not a negative value in the conditions below?

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -700,7 +700,7 @@ namespace AI
                     if ( targetCell != -1 ) {
                         DEBUG_LOG( DBG_BATTLE, DBG_INFO, currentUnit.GetName() << " is under Berserk spell, moving to " << targetCell );
 
-                        const int32_t reachableCell = Board::FindNearestReachableCell( currentUnit, targetCell );
+                        const int32_t reachableCell = arena.GetNearestReachableCell( currentUnit, targetCell );
 
                         DEBUG_LOG( DBG_BATTLE, DBG_INFO, "Nearest reachable cell is " << reachableCell );
```
ihhub, line 705: Shouldn't we check if **reachableCell** is not a negative value in the conditions below?

src/fheroes2/battle/battle_arena.h
```diff
@@ -124,6 +124,10 @@ namespace Battle
         Indexes CalculateTwoMoveOverlap( int32_t indexTo, uint32_t movementRange = 0 ) const;
         Indexes GetPath( const Unit &, const Position & ) const;
 
+        // Returns the cell nearest to the end of the path to the cell with the given index (according to the ArenaPathfinder)
+        // and reachable for the current unit (to which the current board passability information relates)
```
ihhub, line 128: Can we then please add `In case of unreachable cells -1 is returned`? It's absolutely optional and you can skip it.

src/fheroes2/battle/battle_arena.cpp
```diff
@@ -672,6 +672,37 @@ Battle::Indexes Battle::Arena::getAllAvailableMoves( uint32_t moveRange ) const
...
+    }
+
+    // Search for the reachable cell nearest to the end of the path
+    const Cell * nearestCell = nullptr;
+
+    for ( auto it = path.crbegin(); it != path.crend(); ++it ) {
+        const Position pos = Position::GetReachable( currentUnit, *it );
+
+        if ( pos.GetHead() != nullptr && ( !currentUnit.isWide() || pos.GetTail() != nullptr ) ) {
+            nearestCell = pos.GetHead();
```
ihhub, line 698: Should we just return from here instead of using extra variable? In this case **nearestCell** won't be needed and at the end of the function we just return **-1**.

src/fheroes2/battle/battle_arena.cpp
```diff
@@ -672,6 +672,37 @@ Battle::Indexes Battle::Arena::getAllAvailableMoves( uint32_t moveRange ) const
...
+int32_t Battle::Arena::GetNearestReachableCell( const Unit & currentUnit, const int32_t dst ) const
+{
+    const Position dstPos = Position::GetReachable( currentUnit, dst );
+
+    if ( dstPos.GetHead() != nullptr && ( !currentUnit.isWide() || dstPos.GetTail() != nullptr ) ) {
+        // Destination cell is already reachable
+        return dstPos.GetHead()->GetIndex();
+    }
+
+    const Indexes path = _pathfinder.buildPath( dst );
```
ihhub, line 684: A common comment not related to these changes: I have doubts that `Indexes` class has move constructor even if it's derived from **std::vector**. In this case we're calling copy constructor.

ihhub (commented):
Hi @oleg-derevenetz , after taking second view on this pull request I left few very minor comments. Could you please take a look and check if they make sense?

ihhub:
@LeHerosInconnu , here is compiled debug version:
[fheroes2.zip](https://github.com/ihhub/fheroes2/files/7491735/fheroes2.zip)

src/fheroes2/ai/normal/ai_normal_battle.cpp
oleg-derevenetz: @ihhub if `reachableCell` is negative here, then it means that something is wrong either with AI battle planner or AI pathfinder (because AI battle planner commands the unit to move to the unreachable cell - unreachable not just on the current turn, which is normal, but unreachable in principle, and this should not happen). In this case `Battle::Arena::ApplyActionMove()` will print a handy warning and unit just stays in place.

src/fheroes2/battle/battle_arena.cpp
oleg-derevenetz: @ihhub I believe copy elision (namely NRVO) will do its job here and no extra copy constructors will be called:

https://godbolt.org/z/6a8cn73T9

src/fheroes2/battle/battle_arena.cpp
oleg-derevenetz: Yes, I adapted the previous version of the code and overlooked this, now it could be done easier.

ihhub (approved):

LeHerosInconnu:
Hello everyone,

> Okay, thanks. I'll wait for the debug build then and I'll start a test tomorrow.

I have completed the test.
So, I played a complete scenario with the debug version without any problems (Broken Alliance).
I also played for a few weeks a scenario where the AI battles a considerable number of wandering creatures, again with the debug version, without encountering any problems either (Alteris Altered).

[AlteAlte.zip](https://github.com/ihhub/fheroes2/files/7498392/AlteAlte.zip)

ihhub:
@LeHerosInconnu , marvellous! Thank you so much helping with testing!

ihhub:

Issue #4550, opened 2021-11-04
Adventure map, several sounds played at the same time when the hero gains a level after fighting the guardians of an artifact


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, several sounds are played at the same time when the hero gains a level after fighting the guardians of an artifact.

In fheroes2:

Activate the sound of the videos.

Normal:

https://user-images.githubusercontent.com/43583994/140333693-866553ae-cfbe-4159-9aea-61f55315c9a2.mp4

Two sounds:

https://user-images.githubusercontent.com/43583994/140333745-0f554870-e788-4a54-b9a4-a37949b8b47e.mp4

### Save file

Here is the save file:
[Before two sounds artifact up.zip](https://github.com/ihhub/fheroes2/files/7475626/Before.two.sounds.artifact.up.zip)

### Additional info

Version 0.9.8 4530

vincent-grosbois:
I'm confused... should we just play the "level up" sound and never any sound for artifact pickup ?

LeHerosInconnu:
Hello @vincent-grosbois,

> I'm confused... should we just play the "level up" sound and never any sound for artifact pickup ?

This is how it is handled in the original game:

Activate the sound of the videos.

No condition:

https://user-images.githubusercontent.com/43583994/140610240-9621aac7-bf67-4dcf-a460-055363d8c545.mp4

Fighting the guardians:

https://user-images.githubusercontent.com/43583994/140610213-470187e3-7223-4468-a48c-ecf50d3a3426.mp4

Wise person condition:

https://user-images.githubusercontent.com/43583994/140610247-960c48eb-032b-43f4-9132-463a78e7e655.mp4

But, I think it's possible to do better in fheroes2. :)

The sound of the hero level up should be played when this window appears:

![vlcsnap-2021-11-06-13h19m49s999](https://user-images.githubusercontent.com/43583994/140610152-84b4873a-da68-40b8-b5c0-ef663387a546.png)

The sound of capturing the artifact should be played when this window appears:

[...]

PR #4564, opened 2021-11-06
Fix sound effects when receiving artifacts from world map

vincent-grosbois:
fix for #4550


ihhub:
@vincent-grosbois , thank you very much for these changes!

PR #4575, opened 2021-11-07
Add hotkey to Okay button in Campaign select window

a1exsh:
Improve convenience: ESC is bound to Cancel, but Okay could only be clicked.


ihhub:
@a1exsh , thank you so much for this small but valuable improvement!

Issue #4567, opened 2021-11-06
Combat, AI behavior, AI hero casts the Disrupting Ray spell continually on the Titans (0.9.9 4551)


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, Myrini casts the Disrupting Ray spell continually on the Titans.
Myrini (who is also a quite strong hero) could also flee towards the end of the combat rather than let the opposing hero capture her artifacts.

In fheroes2:

https://user-images.githubusercontent.com/43583994/140625794-2001ab4a-5275-46c7-a76a-6db9b58655ad.mp4

Halon attacks Myrini.

### Save file

Here is the save file:
[Before attack disrupting ray_.zip](https://github.com/ihhub/fheroes2/files/7491344/Before.attack.disrupting.ray_.zip)

### Additional info

Game started with version 0.9.8 4544 and continued with version 0.9.9 4551.

vincent-grosbois:
I will have a look but I think the issue of hero fleeing could be done in another issue.

PR #4580, opened 2021-11-07
Better heuristic for AI usage of Disrupting Ray

vincent-grosbois:
Relates to #4567

I changed the logic a little:
- if enemy has already defense < 4, we reduce the score of disrupting ray
(really rare corner case)
- if we have a weak army, it means we will do weak direct damage, so
it reduces the usefulness of disrupting ray. In that case I reduce the
score

If I replay the fight of #4567 it will make the AI stop spamming DR
and cast lightning bolt on the titans

NB : in theory the "enemy strength" should take into account the actual
enemy defense, that would also get lower as we cast DR on the enemy.
But the logic change I'm doing is still valid IMO


idshibanov (approved):
LGTM.

ihhub:
@vincent-grosbois , thanks a lot for this improvement!

PR #4581, opened 2021-11-08
Some improve Funds class

vasilenkoalexey:
Some improvements for `Funds` class and near...


src/fheroes2/castle/buildinginfo.cpp
```diff
@@ -65,134 +65,225 @@ namespace
...
-
-payment_t BuildingInfo::GetCost( u32 build, int race )
+payment_t BuildingInfo::GetCost( const uint32_t build, const int race )
 {
-    payment_t payment;
-    const buildstats_t * ptr = &_builds[0];
-
-    while ( BUILD_NOTHING != ptr->id2 && !( ptr->id2 == build && ( !race || ( race & ptr->race ) ) ) )
-        ++ptr;
+    switch ( build ) {
```
a1exsh, line 70: Hm, not sure this is really an improvement: data beats code any day of the week, they say.

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -871,29 +871,29 @@ bool Kingdom::IsTileVisibleFromCrystalBall( const int32_t dest ) const
     return false;
 }
 
-cost_t Kingdom::_getKingdomStartingResources( const int difficulty )
+payment_t Kingdom::_getKingdomStartingResources( const int difficulty )
 {
     if ( isControlAI() )
-        return { 10000, 30, 10, 30, 10, 10, 10 };
+        return { 30, 30, 10, 10, 10, 10, 10000 };
```
a1exsh, line 877: This looks alarming: do we really have cost and payment types that take all the same resource amount parameters, but in subtly different order?..

src/fheroes2/resource/resource.h
```diff
@@ -67,68 +50,64 @@ class Funds
...
+    void Reset();
+    std::string String() const;
+
+    int32_t wood;
+    int32_t mercury;
+    int32_t ore;
+    int32_t sulfur;
+    int32_t crystal;
+    int32_t gems;
+    int32_t gold;
```
a1exsh, line 84: 1. Maybe order these in the same way they are listed in constructor?
2. Do we really need these as negative values at any point?  If not, let's use an unsigned type?

src/fheroes2/kingdom/kingdom.cpp
vincent-grosbois: yes.... also notice that the big change in building info is always switching the order of resources etc, so pretty error prone

a1exsh: Reading the change further I see that we remove `cost_t` altogether — it'd be helpful to note that in the PR description to align the expectations ;)

PR #4582, opened 2021-11-08
Kingdom::LossPostActions(): don't (indirectly) call erase() in the std::for_each() functor

### LINKED ISSUE #4579: A hero still on a boat after the player was vanquished
### Preliminary checks

### Platform

Linux

### Describe the bug

The Orange player was vanquished couple of weeks ago, but his last hero is still on a boat (sitting on the Whirlpool in the southern part of the sea).

### Save file

[Evil_11_0095.zip](https://github.com/ihhub/fheroes2/files/7492938/Evil_11_0095.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4579

`std::vector::erase()` should not be called while walking over the vector using `std::for_each()` because it invalidates current iterator as well as pre-calculated `end()`.

### DIFF
--- a/src/fheroes2/kingdom/kingdom.cpp
+++ b/src/fheroes2/kingdom/kingdom.cpp
@@ -127,14 +127,20 @@ void Kingdom::LossPostActions( void )
     if ( isPlay() ) {
         Players::SetPlayerInGame( color, false );
 
-        if ( !heroes.empty() ) {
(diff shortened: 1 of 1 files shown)


ihhub:
@oleg-derevenetz , a huge thanks for this fix! I am amazed how the previous code works without any crashes.

oleg-derevenetz:
@ihhub that's because vector storage is not reallocated after `erase()`, so all invalidated vector elements still point to the valid `Heroes` objects. However, since `erase()` shifts vector elements to the left, some `Heroes` objects (I suppose, about half of them) may be skipped and `SetFreeman()` is not called for these objects.

PR #4584, opened 2021-11-08
Load palette from AGG file

ihhub:
icn2img tool now requires original kb.pal file.

The solution is not ideal as Display class needs a valid palette at the initialization.

### DIFF
--- a/fheroes2-vs2015.vcxproj
+++ b/fheroes2-vs2015.vcxproj
@@ -176,6 +176,7 @@
     <ClCompile Include="src\engine\core.cpp" />
     <ClCompile Include="src\engine\dir.cpp" />
     <ClCompile Include="src\engine\image.cpp" />
+    <ClCompile Include="src\engine\image_palette.cpp" />
     <ClCompile Include="src\engine\image_tool.cpp" />
     <ClCompile Include="src\engine\localevent.cpp" />
     <ClCompile Include="src\engine\logging.cpp" />
@@ -390,6 +391,7 @@
     <ClInclude Include="src\engine\core.h" />
     <ClInclude Include="src\engine\dir.h" />
     <ClInclude Include="src\engine\image.h" />
+	<ClInclude Include="src\engine\image_palette.h" />
     <ClInclude Include="src\engine\image_tool.h" />
     <ClInclude Include="src\engine\logging.h" />
     <ClInclude Include="src\engine\localevent.h" />
--- a/src/engine/image_palette.cpp
+++ b/src/engine/image_palette.cpp
@@ -36,7 +36,7 @@ namespace
             std::copy_n( kb_pal, paletteSize, gamePalette.begin() );
         }
 
-        std::array<uint8_t, 768> gamePalette;
+        std::array<uint8_t, paletteSize> gamePalette;
     };
 
     PaletteHolder paletteHolder;
--- a/src/engine/image_palette.cpp
+++ b/src/engine/image_palette.cpp
@@ -1,0 +1,61 @@
+/***************************************************************************
+ *   Free Heroes of Might and Magic II: https://github.com/ihhub/fheroes2  *
+ *   Copyright (C) 2021                                                    *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ *   This program is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU General Public License for more details.                          *
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#include "image_palette.h"
+#include "palette_h2.h"
+
+#include <algorithm>
+#include <array>
+#include <cassert>
+
+namespace
+{
+    const size_t paletteSize = 768;
+
+    struct PaletteHolder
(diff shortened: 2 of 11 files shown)


src/engine/image_palette.cpp
```diff
@@ -0,0 +1,61 @@
...
+    const size_t paletteSize = 768;
+
+    struct PaletteHolder
+    {
+        PaletteHolder()
+        {
+            std::copy_n( kb_pal, paletteSize, gamePalette.begin() );
+        }
+
+        std::array<uint8_t, 768> gamePalette;
```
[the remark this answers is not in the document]
oleg-derevenetz: I'd suggest to do this as follows:

```
constexpr size_t paletteSize = 768;
[...]
std::array<uint8_t, paletteSize> gamePalette;
```

src/tools/icn2img.cpp
```diff
@@ -26,52 +26,55 @@
...
 {
-    if ( argc < 3 ) {
-        std::cout << argv[0] << " infile.icn extract_to_dir" << std::endl;
-        return EXIT_SUCCESS;
+    if ( argc < 4 ) {
+        std::cout << argv[0] << " inputFile.icn destinationDirectory paletteFileName.pal" << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    std::string intpuFileName( argv[1] );
```
oleg-derevenetz, line 41: I suppose there is a typo: `intpuFileName` -> `inputFileName`.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions here, please take a look when you have a free time.

src/engine/image_palette.cpp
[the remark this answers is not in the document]
ihhub: Ah, yes. I missed it after third work redo...

src/engine/screen.cpp
```diff
@@ -126,13 +126,15 @@ namespace
         return indexes;
     }
 
-    const uint8_t * PALPAlette()
+    const uint8_t * PALPalette()
```
a1exsh, line 129: Any chance to keep only this function or the one with the same name from `image_tool.cpp`?

src/tools/icn2img.cpp
```diff
@@ -26,52 +26,55 @@
...
 #include "image_tool.h"
 #include "serialize.h"
 #include "system.h"
 
 int main( int argc, char ** argv )
 {
-    if ( argc < 3 ) {
-        std::cout << argv[0] << " infile.icn extract_to_dir" << std::endl;
-        return EXIT_SUCCESS;
+    if ( argc < 4 ) {
```
a1exsh, line 36:
```suggestion
    if ( argc != 4 ) {
```

Extra parameters should not be accepted either.

src/tools/icn2img.cpp
```diff
@@ -26,52 +26,55 @@
...
+    std::vector<uint8_t> palette = paletteFile.getRaw();
+    if ( palette.size() != 768 ) {
+        std::cout << "Invalid palette size of " << palette.size() << " instead of 768." << std::endl;
+        return EXIT_FAILURE;
+    }
 
-    std::string shortname( *ptr );
-    ++ptr;
-    std::string prefix( *ptr );
+    fheroes2::setGamePalette( palette );
```
a1exsh, line 58: I'd pass palette as a parameter to `SaevImage()` rather than relying on the global value.

src/tools/icn2img.cpp
```diff
@@ -102,17 +105,11 @@ int main( int argc, char ** argv )
             dstfile += ".png";
             shortdstfile += ".png";
 #endif
-            if ( fheroes2::Save( image, dstfile, 23 ) ) {
-                fs << " <sprite index=\"" << ii + 1 << "\" name=\"" << shortdstfile.c_str() << "\" ox=\"" << head.offsetX << "\" oy=\"" << head.offsetY << "\"/>"
```
a1exsh, line 106: I know nobody likes XML, but the offset information is important for loading the sprites back to the game from a set of images: we should not just discard it.  Maybe just print it on the console so that the user can capture it to a text file?

a1exsh (commented):
@ihhub please have a look on my comments when you have time. :)

src/tools/icn2img.cpp
ihhub: The original palette is not suitable for this. All values in this palette are in 4 times smaller they should be.

a1exsh: I'm not sure you got my point. Currently, the result of calling SaveImage depends on the globally set palette. I suggest to pass it as a parameter to SaveImage instead of using the global definition (relying on side-effects).

src/engine/screen.cpp
ihhub: icn2img tool shouldn't even have an access to screen code. We have to remove such dependency. Also this function has a static member while image_tool's doesn't. This function must be removed completely but as stated in the description of this pull request we still have a problem on using palette before even reading it.

The purpose of this pull request is to remove `palette_h2.h` completely as it must not belong to this project.

ihhub:
Hi @a1exsh , please take a look an my answers as well as I addressed some of your comments.

fheroes2-vs2015.vcxproj
```diff
@@ -390,6 +391,7 @@
     <ClInclude Include="src\engine\core.h" />
     <ClInclude Include="src\engine\dir.h" />
     <ClInclude Include="src\engine\image.h" />
+	<ClInclude Include="src\engine\image_palette.h" />
```
a1exsh, line 394: Is this a TAB char? ;)

src/tools/icn2img.cpp
```diff
@@ -102,17 +105,13 @@ int main( int argc, char ** argv )
...
-                   << std::endl;
-            }
+            std::cout << "Image " << ii + 1 << " has offset of [" << head.offsetX << ", " << head.offsetY << "]" << std::endl;
```
a1exsh, line 108:
```suggestion
            std::cout << "Image " << ii + 1 << " has offset of [" << static_cast<int32_t>( head.offsetX ) << ", " << static_cast<int32_t>( head.offsetY ) << "]" << std::endl;
```
Otherwise negative offsets look really funny (-1 becomes 65535 and so on) ;)

a1exsh (commented):
Spotted two small issues: please have a look.

Issue #4129, opened 2021-09-02
Combat, AI behavior, AI casts not necessary spells in combat, here Slow and Haste (0.9.6 4111)


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

In fheroes2, AI casts not necessary spells in combat, here Slow and Haste.
AI casts the Slow spell on the troop of Hydras which is already the slowest troop and attacks this slowed troop in close combat anyway.
AI casts the Haste spell on his troops who are already faster, and who can already reach the opposing troop.

In fheroes2:

https://user-images.githubusercontent.com/43583994/131835286-9fe53b2f-6a45-47a9-9675-8f7d5bc3b7c6.mp4

Flint attacks Roxana.

Here is the save file:
[AI casts slow no need 01.zip](https://github.com/ihhub/fheroes2/files/7098467/AI.casts.slow.no.need.01.zip)

### Describe the possible solution

AI should economize his spell points rather than casting unnecessary spells.

ihhub:
This issue should be retested on the latest version.

LeHerosInconnu:
Hello @ihhub and @vincent-grosbois,

> This issue should be retested on the latest version.

I did some tests with version 0.9.9 4584.

With the save file "AI casts slow no need 01.sav", the AI still uses the Slow and Haste spells.

In fheroes2:

https://user-images.githubusercontent.com/43583994/141683600-82bd77eb-31f1-495b-82ba-0dfeb6270b3c.mp4

With a scenario started from the beginning, it's different.
In this particular case, where the AI is adopting a defensive strategy, the Slow spell could be used at start of the combat. :)
Also, the AI does not use Dispel Magic, Magic Arrow or Curse, and still use the spell Haste at the end of the combat.

In fheroes2:

![AI spells 02](https://user-images.githubusercontent.com/43583994/141685071-d672251c-0b2d-402e-8969-505b4ed34f3c.png)

https://user-images.githubusercontent.com/43583994/141684872-1d6ac4ee-d6eb-4ab0-aea7-800cf1ecebe1.mp4

Regarding this remark:

> This should somehwat improve issue https://github.com/ihhub/fheroes2/issues/4129 even though the problem will still remain that enemy will cast a slow spell even though it has very low score, if it's the only spell available

[...]

PR #4599, opened 2021-11-09
Fix heuristic for slow/haste spell in battle

vincent-grosbois:
I fixed a few bugs around speed computation, where enemy speed was
incorrectly computed. I also change the "average enemy speed" so it's weighted by monster
strength

I also changed the heuristic a bit so monster that are already slow/fast
don't really benefit that much from the spell

This should somehwat improve issue #4129 even though the problem will still
remain that enemy will cast a slow spell even though it has very low score,
if it's the only spell available

### DIFF
--- a/src/fheroes2/battle/battle_troop.h
+++ b/src/fheroes2/battle/battle_troop.h
@@ -132,7 +132,7 @@ namespace Battle
         int GetCurrentOrArmyColor() const; // current unit color (if valid), color of the unit's army otherwise
         int GetCurrentControl() const;
         uint32_t GetMoveRange() const;
-        u32 GetSpeed( bool skip_standing_check ) const;
+        u32 GetSpeed( bool skip_standing_check, bool skip_moved_check ) const;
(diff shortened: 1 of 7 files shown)


src/fheroes2/battle/battle_troop.h
```diff
@@ -132,7 +132,7 @@ namespace Battle
         int GetCurrentOrArmyColor() const; // current unit color (if valid), color of the unit's army otherwise
         int GetCurrentControl() const;
         uint32_t GetMoveRange() const;
-        u32 GetSpeed( bool skip_standing_check ) const;
+        u32 GetSpeed( bool skip_standing_check, bool skip_moved_check ) const;
```
ihhub, line 135: Please use `camelCase` variable naming. `skip_standing_check` is from old code.

ihhub (changes requested):

ihhub:
@vincent-grosbois , a huge thanks for these fixes and improvements!

PR #4600, opened 2021-11-09
Well : Clean up, optimize and make consistent with OG

zenseii:
This first part currently only affects peasants since they are the only creatures where min damage is the same as max damage. In the fheroes2 monster info window their damage is already reported as just "1", so this also makes it more consistent.

OG:
image

fheroes2:
image

This is most likely why Damage is written "Damg" in OG:
OG:
image
fheroes2:
image

### DIFF
--- a/src/fheroes2/castle/castle_well.cpp
+++ b/src/fheroes2/castle/castle_well.cpp
@@ -35,8 +35,8 @@
 #include "monster_anim.h"
 #include "resource.h"
 #include "speed.h"
-#include "text.h"
 #include "translations.h"
+#include "ui_text.h"
 
 namespace
 {
@@ -188,10 +188,10 @@ void Castle::OpenWell( void )
                     }
                 }
                 if ( isCreaturePresent ) {
-                    Dialog::Message( "", _( "Not enough resources to buy monsters." ), Font::BIG, Dialog::OK );
+                    Dialog::Message( "", _( "Not enough resources to buy monsters." ), int (fheroes2::FontSize::LARGE), Dialog::OK );
                 }
                 else {
-                    Dialog::Message( "", _( "No monsters available for purchase." ), Font::BIG, Dialog::OK );
+                    Dialog::Message( "", _( "No monsters available for purchase." ), int( fheroes2::FontSize::LARGE ), Dialog::OK );
                 }
             }
             else if ( Dialog::YES == Dialog::ResourceInfo( _( "Buy Monsters" ), str, total, Dialog::YES | Dialog::NO ) ) {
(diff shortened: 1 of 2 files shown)


src/fheroes2/castle/castle_well.cpp
```diff
@@ -336,7 +336,15 @@ void Castle::WellRedrawInfoArea( const fheroes2::Point & cur_pt, const std::vect
...
-        str = std::string( _( "Damage" ) ) + ": " + std::to_string( monster.GetDamageMin() ) + "-" + std::to_string( monster.GetDamageMax() );
+        str = std::string( _( "Damg" ) ) + ": ";
+
+        if ( monster.GetDamageMin() != monster.GetDamageMax() ) {
+            str.append( std::to_string( monster.GetDamageMin() ) + "-" + std::to_string( monster.GetDamageMax() ) );
+        }
+        else {
+            str.append( std::to_string( monster.GetDamageMin() ) );
+        }
+
```
ihhub, line 347: Since we touched this code I'd suggest to do some optimization to avoid extra memory copying:
```cpp
str = _( "Damg" );
str += ": ";

const uint32_t monsterMinDamage = monster.GetDamageMin();
const uint32_t monsterMaxDamage = monster.GetDamageMax();

str += std::to_string( monsterMinDamage );

if ( monsterMinDamage != monsterMaxDamage ) {
    str += '-';
    str += std::to_string( monsterMaxDamage );
}
```

What do you think?

ihhub (changes requested):

src/fheroes2/castle/castle_well.cpp
zenseii: Yes, this much better.

zenseii:
@ihhub I see that this code optimization can be applied for more of the strings for the well so I'll extend the scope of this PR to change those as well and while I'm at it I'll replace `text.h` with `ui_text.h.`

src/fheroes2/castle/castle_well.cpp
```diff
@@ -311,78 +313,147 @@ void Castle::WellRedrawInfoArea( const fheroes2::Point & cur_pt, const std::vect
...
+        const std::string monsterMultiName = monster.GetMultiName();
+
+        text.set( monsterMultiName, statsFontType );
+        dst_pt.x = pt.x + 122 - text.width() / 2;
+        dst_pt.y = pt.y + 19;
+        text.draw( dst_pt.x, dst_pt.y, display );
+
+        // stats declarations
+
         std::string str;
```
ihhub, line 335: Please move the declaration of this variable to line 341 just where we assign the value to it.

src/fheroes2/castle/castle_well.cpp
```diff
@@ -311,78 +313,147 @@ void Castle::WellRedrawInfoArea( const fheroes2::Point & cur_pt, const std::vect
...
-        // attack
+        const std::string monsterMultiName = monster.GetMultiName();
+
+        text.set( monsterMultiName, statsFontType );
+        dst_pt.x = pt.x + 122 - text.width() / 2;
+        dst_pt.y = pt.y + 19;
+        text.draw( dst_pt.x, dst_pt.y, display );
+
+        // stats declarations
+
```
ihhub, line 334: Please remove this empty line.

src/fheroes2/castle/castle_well.cpp
```diff
@@ -311,78 +313,147 @@ void Castle::WellRedrawInfoArea( const fheroes2::Point & cur_pt, const std::vect
...
+        const int32_t statsInitialOffsetY = 22;
+        int32_t statsOffsetY = statsInitialOffsetY;
+
+        // attack
+        str = _( "Attack" );
+        str += ": ";
+
+        const uint32_t monsterAttack = monster.GetAttack();
+
+        str += std::to_string( monsterAttack );
```
ihhub, line 346: This variable is used only in the line below. Please use the call directly to avoid extra memory copy.

src/fheroes2/castle/castle_well.cpp
```diff
@@ -311,78 +313,147 @@ void Castle::WellRedrawInfoArea( const fheroes2::Point & cur_pt, const std::vect
...
+
         std::string str;
-        str = std::string( _( "Attack" ) ) + ": " + std::to_string( monster.GetAttack() );
-        text.Set( str );
-        dst_pt.x = pt.x + 268 - text.w() / 2;
-        dst_pt.y = pt.y + 22;
-        text.Blit( dst_pt.x, dst_pt.y );
+        const int32_t statsOffsetX = 269;
+        const int32_t statsInitialOffsetY = 22;
+        int32_t statsOffsetY = statsInitialOffsetY;
```
ihhub, line 338: Please move these variable declaration just before the place where they used. It is much easier to read the code when declaration follows by usage.

ihhub (changes requested):
Hi @zenseii , I put few comments here. I didn't repeat them to avoid polluting this pull request. Could you please take a look?

ihhub (changes requested):
Hi @zenseii , I left one more comment here. Could you please take a look again?

src/fheroes2/castle/castle_well.cpp
```diff
@@ -188,10 +188,10 @@ void Castle::OpenWell( void )
                     }
                 }
                 if ( isCreaturePresent ) {
-                    Dialog::Message( "", _( "Not enough resources to buy monsters." ), Font::BIG, Dialog::OK );
+                    Dialog::Message( "", _( "Not enough resources to buy monsters." ), int( fheroes2::FontSize::LARGE ), Dialog::OK );
```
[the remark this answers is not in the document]
zenseii: An alternative could be to write "2" instead of Font::BIG so you don't have to include text.h just for this. Not optimal, but I imagine it is not as error prone.

zenseii:
You're welcome and thank you for taking the time and having the patience to go through this noob code! 😅

PR #4609, opened 2021-11-11
Fix max music/sound volume

a1exsh:
1. The max value is 10, but we were going only up to 9 previously.
2. Move min/max logic from dialog to settings (it was partially there already).

### DIFF
--- a/src/fheroes2/dialog/dialog_system_options.cpp
+++ b/src/fheroes2/dialog/dialog_system_options.cpp
@@ -270,15 +270,15 @@ namespace
             bool saveSoundVolume = false;
             if ( Audio::isValid() ) {
                 if ( le.MouseClickLeft( musicVolumeRoi ) ) {
-                    conf.SetMusicVolume( ( conf.MusicVolume() + 1 ) % 10 );
+                    conf.SetMusicVolume( ( conf.MusicVolume() + 1 ) % 11 );
                     saveMusicVolume = true;
                 }
                 else if ( le.MouseWheelUp( musicVolumeRoi ) ) {
-                    conf.SetMusicVolume( std::min( conf.MusicVolume() + 1, 9 ) );
+                    conf.SetMusicVolume( conf.MusicVolume() + 1 );
                     saveMusicVolume = true;
                 }
                 else if ( le.MouseWheelDn( musicVolumeRoi ) ) {
-                    conf.SetMusicVolume( std::max( 0, conf.MusicVolume() - 1 ) );
+                    conf.SetMusicVolume( conf.MusicVolume() - 1 );
                     saveMusicVolume = true;
                 }
                 if ( saveMusicVolume ) {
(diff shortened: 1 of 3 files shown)


src/fheroes2/system/settings.cpp
```diff
@@ -646,17 +646,7 @@ void Settings::setFullScreen( const bool enable )
...
-    case SCROLL_NORMAL:
-    case SCROLL_FAST1:
-    case SCROLL_FAST2:
-        scroll_speed = speed;
-        break;
-    default:
-        scroll_speed = SCROLL_NORMAL;
-        break;
-    }
+    scroll_speed = clamp( static_cast<int>( speed ), static_cast<int>( SCROLL_SLOW ), static_cast<int>( SCROLL_FAST2 ) );
```
ihhub, line 649: Please remove static casting from **speed** as it's already **int**.

ihhub: Also declare the enumeration containing scroll values as `int` to drop static casts everywhere.

ihhub (changes requested):
Hi @a1exsh , I left just 2 comments. Could you please take a look?

src/fheroes2/system/settings.cpp
a1exsh: > Also declare the enumeration containing scroll values as `int` to drop static casts everywhere.

I've tried now and I'm running into this error:
```
error: declaration of ‘ScrollSpeed Settings::ScrollSpeed() const’ changes meaning of ‘ScrollSpeed’ [-fpermissive]
```

May we change the function's name to `GetScrollSpeed()`?  We would have to rename all related functions as well though.

a1exsh: This is the new enum:
```
enum class ScrollSpeed : int
{
    SLOW = 1,
    NORMAL = 2,
    FAST1 = 3,
    FAST2 = 4
};
```

a1exsh: Or we can name the enum differently, to avoid the conflict, e.g. `MapScrollSpeed`.

Still it doesn't eliminate the need for all static casts, e.g. to do arithmetic one still has to cast to `int` and then back to enum type.  Or am I doing something wrong? :)

ihhub: Hi @a1exsh , what I meant is to declare the enumeration as `enum : int` instead of `enum` without adding a name. Yes, it's a half solution but it solves the issue with casting.

a1exsh: Ah, I haven't seen this use before (also not in this project).  My compiler is still unhappy about it with the casts when calling `clamp()`:
```
error: no matching function for call to ‘clamp(int&, <unnamed enum>, <unnamed enum>)’
  650 |     scroll_speed = clamp( speed, SCROLL_SLOW, SCROLL_FAST2 );
```

The arithmetic seems to just work though.

ihhub: -_- fine. Let's leave static cast for now. We can come back to this enumeration later.

src/fheroes2/dialog/dialog_system_options.cpp
```diff
@@ -351,15 +351,15 @@ namespace
             // set scroll speed
             bool saveScrollSpeed = false;
             if ( le.MouseClickLeft( scrollSpeedRoi ) ) {
-                conf.SetScrollSpeed( SCROLL_FAST2 > conf.ScrollSpeed() ? conf.ScrollSpeed() + 1 : SCROLL_SLOW );
+                conf.SetScrollSpeed( conf.ScrollSpeed() % 4 + 1 );
```
ihhub, line 354: Should we use `SCROLL_FAST2` here instead?

a1exsh: Eh, why not ;)

ihhub:
@a1exsh , thank you very much for this fix!

PR #4610, opened 2021-11-11
Allow to change Battle Speed using mouse wheel

### LINKED ISSUE #4608: Missing mouse wheel handling in Battle options
### Preliminary checks

Overlooked when working on #4553 , battle animation speed should also be possible to change with the mouse wheel:

image

We should add the same handling as in system options.

### Additional info

_No response_

a1exsh:
Fix #4608

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -248,31 +248,35 @@ void Battle::DialogBattleSettings( void )
     while ( le.HandleEvents() ) {
         le.MousePressLeft( btn_ok.area() ) ? btn_ok.drawOnPress() : btn_ok.drawOnRelease();
 
-        bool saveBattleSpeed = false;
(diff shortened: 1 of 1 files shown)


src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -308,6 +313,14 @@ void Battle::DialogBattleSettings( void )
...
             break;
         }
+
+        if ( saveBattleSpeed ) {
+            Game::UpdateGameSpeed();
+            fheroes2::Blit( dialog, display, pos_rt.x, pos_rt.y );
+            RedrawBattleSettings( optionAreas );
+            display.render();
+            saveConfiguration = true;
+        }
```
ihhub, line 323: Shall we also do a small refactoring to avoid copying the same code all over the function?

ihhub:
@a1exsh , thanks a lot of making this improvement!

PR #4631, opened 2021-11-13
Improve marketplace gift button placement

zenseii:
This relates to discussion https://github.com/ihhub/fheroes2/discussions/4621.

This makes the gift button always available in the marketplace instead of conditionally hidden.

Code for updating the trade area had to be added to the gift condition to avoid being able to gift resources and then trade those already gifted resources again and therefore twice. Without the adjustments, this would be possible if you select two resources to trade right before you open the gift dialog.

### DIFF
--- a/src/fheroes2/dialog/dialog_marketplace.cpp
+++ b/src/fheroes2/dialog/dialog_marketplace.cpp
@@ -552,7 +552,8 @@ void Dialog::Marketplace( Kingdom & kingdom, bool fromTradingPost )
 
             firstExchange = false;
 
-            resourceTo = resourceFrom = Resource::UNKNOWN;
+            resourceTo = Resource::UNKNOWN;
+            resourceFrom = Resource::UNKNOWN;
             gui.ShowTradeArea( kingdom, resourceFrom, resourceTo, 0, 0, 0, 0, fromTradingPost, firstExchange );
 
             fundsFrom = kingdom.GetFunds();
--- a/src/fheroes2/dialog/dialog_marketplace.cpp
+++ b/src/fheroes2/dialog/dialog_marketplace.cpp
@@ -144,8 +144,11 @@ public:
         buttonLeft.setICNInfo( tradpost, 3, 4 );
         buttonRight.setICNInfo( tradpost, 5, 6 );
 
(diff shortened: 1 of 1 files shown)


a1exsh:
LGTM :+1:

src/fheroes2/dialog/dialog_marketplace.cpp
```diff
@@ -425,10 +428,19 @@ void Dialog::Marketplace( Kingdom & kingdom, bool fromTradingPost )
         if ( le.MouseClickLeft( buttonExit.area() ) || HotKeyCloseWindow )
             break;
 
+        // gift resources
         if ( buttonGift.isEnabled() && le.MouseClickLeft( buttonGift.area() ) ) {
             Dialog::MakeGiftResource( kingdom );
+
+            resourceTo = resourceFrom = Resource::UNKNOWN;
```
ihhub, line 435: Please use one operation per line:
```cpp
resourceTo = Resource::UNKNOWN;
resourceFrom = Resource::UNKNOWN;
```

ihhub (changes requested):

src/fheroes2/dialog/dialog_marketplace.cpp
zenseii: Sure! I saw that it was used like this but that must have been part of the old code. I fixed this where I saw it as well now.

ihhub:
Hi @Branikolog , are you okay with this change?

ihhub (approved):

Branikolog:
Fix works well, but in my opinion, this rarely used button takes too "front" place in this window. On the other hand I cannot suggest anything better without complete reworking of this window... So... 👍

zenseii:
> Fix works well, but in my opinion, this rarely used button takes too "front" place in this window. On the other hand I cannot suggest anything better without complete reworking of this window... So... 👍

Yes, I agree with these points and that is also why I proposed to implement av way for the gift button to only appear in case you're playing a multiplayer scenario, maybe even specifically only if there are other human players since I'm not sure there really is much point in gifting to a friendly AI? Maybe if you're playing a scenario where you're depending on your AI teammate to survive or you want to help them survive? This would all be part of a different PR though.

zenseii:
@ihhub this PR has gotten a good-to-go by enough people now that I feel confident in saying that it is ready to be merged.

ihhub:
@zenseii , thanks a lot for this change!

PR #4635, opened 2021-11-14
Remove BUILD_RELEASE macro

### LINKED ISSUE #4629: WITH_DEBUG and BUILD_RELEASE should be grouped into one macro
### Preliminary checks

### Platform

Windows

### Describe the bug

The code has two macros `WITH_DEBUG` and `BUILD_RELEASE`. These flags are mutually exclusive but somehow we still have them. We must have only one macro for this. Also CMake build has a bug: it sets only `BUILD_RELEASE` macro for release and none for debug build.

### Save file

Nothing to attach.

### Additional info

_No response_

oleg-derevenetz:
fix #4629


ihhub:
@oleg-derevenetz , thank you so much for fixing it!

PR #4652, opened 2021-11-17
Make Unix script to automatically copy & extract all necessary assets from original HoMM2 distributions

### LINKED ISSUE #4647: Original files extraction for Linux systems
### Preliminary checks

Installation of fheroes2 using GOG version requires game files.

People owning GOG version need an instructions.

### Additional info

_No response_

oleg-derevenetz:
close #4647

Currently `python` (even 2.x will do) and `bsdtar` from `libarchive` (installed almost everywhere) are required if there is a need to extract animations from `homm2.gog`.


ihhub:
@oleg-derevenetz , a bunch of thanks for adding this script!

PR #4684, opened 2021-11-23
Update Norwegian translation

zenseii:
This adds my last improvements and a lot of small fixes like typos and references being spelled wrong. 

All the newly added strings have been translated. 
Hero names and experimental options have also been translated now. 

With this PR the translation has been updated against the latest PO template file.


files/lang/nb.po
```diff
@@ -1498,7 +1491,7 @@ msgid "Maze"
 msgstr "Labyrint"
 
 msgid "Upg. Maze"
-msgstr "Oppg. Labyrint"
```
ihhub, line 1501: Doesn't Norwegian allow to use nouns with capital letters?

files/lang/nb.po
```diff
@@ -5916,20 +5930,32 @@ msgstr "Gullgruve"
 msgid "Mine"
 msgstr "Gruve"
 
+msgid "Burma shave."
+msgstr "Barberskum"
```
ihhub, line 5934: Haha. Really? People won't understand the joke :) How about some local phrase related to Norwegian? For example a neutral one like "Talk to frogs". Burma Shave is related to billboards https://en.wikipedia.org/wiki/Burma-Shave

ihhub (commented):
Hi @zenseii , I left just 2 questions. Could you please check them?

files/lang/nb.po
zenseii: No, nouns should be lower case, unless it's the start of a sentence. In this case it's after an abbreviation so it should still be lower case. Anyways I'm not too convinced about lower case or not on these upgraded buildings so I could be fine having it either way.

files/lang/nb.po
zenseii: Haha yeah I completely forgot to try to come up with something funny for that one! I'll try to think about something and push a commit.

zenseii:
Made some small adjustments as I was wrapping up a map in Norwegian. I've finished making the commits for this PR as long as you approve the changes. I even added some alternative "jokes" in case they're prefered or needed at a later point.

ihhub:
@zenseii , thank you so much for these updates!

PR #4688, opened 2021-11-23
Properly handle the day-week-month routine

### LINKED ISSUE #4657: Adventure map, Stables, Stables bonus is only removed at the beginning of day 2 instead of the beginning of day 1
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the Stables bonus is only removed at the beginning of day 2 instead of the beginning of day 1.

In fheroes2:

Stables multi solo 005

Stables multi solo 006

End the turn.

### Save file

Here is the save file:
[Stables day 1.zip](https://github.com/ihhub/fheroes2/files/7562224/Stables.day.1.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4657

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -530,6 +530,11 @@ u32 * Castle::GetDwelling( u32 dw )
 
 void Castle::ActionNewWeek( void )
 {
+    // skip the first week
+    if ( world.CountWeek() < 2 ) {
+        return;
+    }
+
     const bool isNeutral = GetColor() == Color::NONE;
 
     // increase population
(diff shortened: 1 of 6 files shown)


vincent-grosbois:
So, where was the bug ? :D

oleg-derevenetz:
> So, where was the bug ? :D

This particular bug was caused by the fact that the daily routine was processed first, and the weekly routine was processed later, so, when the hero's turn points were calculated, stables had not yet been removed from the list of objects visited by the hero. This PR fixes another potential bugs as well related to freeman heroes (for example, if hero was dismissed, his bonuses related to visited objects stayed until he was rehired, so, for example, if a hero with Stables bonus was dismissed on day 7 and rehired by another player by day 2 on the next week, he will still have the Stables bonus until the end of the next week).

ihhub:

PR #4704, opened 2021-11-26
Add default castle's name is none is specified by editor

### LINKED ISSUE #4624: Voyage Home, scenario 4 has a town with no name leading to an assertion crash during the gameplay
### Preliminary checks

### Platform

Windows

### Describe the bug

Voyage Home, scenario 4 has one town with no name just left to the start castle. Once a hero enters the village the game crashes during to an assertion in Text class within `getLineWidth` function call.

### Save file

[attack_village_glitch.zip](https://github.com/ihhub/fheroes2/files/7529186/attack_village_glitch.zip)

### Additional info

_No response_

ihhub:
close #4624

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -21,6 +21,7 @@
  ***************************************************************************/
 
 #include <algorithm>
+#include <array>
 
 #include "agg.h"
 #include "agg_image.h"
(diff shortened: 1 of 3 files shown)


src/fheroes2/castle/castle.cpp
```diff
@@ -2265,6 +2285,26 @@ const std::string & Castle::GetName( void ) const
     return name;
 }
 
+void Castle::setName( const std::set<std::string> & usedNames )
+{
+    assert( name.empty() && !usedNames.empty() );
```
oleg-derevenetz, line 2290: What if all the castles on the map don't have names? In this case `usedNames` will be empty and this assertion will fail.

oleg-derevenetz (commented):

src/fheroes2/castle/castle.cpp
ihhub: I've removed it. You're right.

PR #4706, opened 2021-11-26
Play sounds from closest surrounding objects first

### LINKED ISSUE #4573: "[ERROR] PlayChunk:  No free channels available" appears often
### Preliminary checks

### Platform

Linux

### Describe the bug

In a version built from recent `HEAD` the following errors appear on the console:
```
10:09:19: [DBG_ENGINE]	main:  Free Heroes of Might and Magic II, version: 0.9.9
10:50:12: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:28: [ERROR]	PlayChunk:  No free channels available
10:56:56: [ERROR]	PlayChunk:  No free channels available
10:56:56: [ERROR]	PlayChunk:  No free channels available
10:56:58: [ERROR]	PlayChunk:  No free channels available
10:56:59: [ERROR]	PlayChunk:  No free channels available
11:00:33: [ERROR]	PlayChunk:  No free channels available
11:00:33: [ERROR]	PlayChunk:  No free channels available
11:18:15: [ERROR]	PlayChunk:  No free channels available
11:18:16: [ERROR]	PlayChunk:  No free channels available
11:18:22: [ERROR]	PlayChunk:  No free channels available
```

This was not the case until recently.

### Save file

Not sure what triggers it.  Will add a save file if I find when the error appears.

### Additional info

Haven't noticed any sounds missing so far.

ihhub:
close #4573

### DIFF
--- a/src/fheroes2/game/game.cpp
+++ b/src/fheroes2/game/game.cpp
@@ -345,6 +345,11 @@ u32 & Game::MapsAnimationFrame( void )
 // play environment sounds from the game area in focus
 void Game::EnvironmentSoundMixer()
 {
+    size_t availableChannels = Mixer::getChannelCount();
+    if ( availableChannels == 0 ) {
+        return;
+    }
+
     const fheroes2::Point abs_pt( Interface::GetFocusCenter() );
     std::fill( reserved_vols.begin(), reserved_vols.end(), 0 );
 
@@ -376,6 +381,14 @@ void Game::EnvironmentSoundMixer()
         if ( channel < reserved_vols.size() ) {
             const double distance = std::sqrt( pos.x * pos.x + pos.y * pos.y );
             const int32_t volume = static_cast<int32_t>( ( ( maxDistance - distance ) / maxDistance ) * maxVolume + minVolumeOnMaxDistance );
+
+            if ( reserved_vols[channel] == 0 ) {
+                if ( availableChannels == 0 ) {
+                    // No new channel can be added.
+                    continue;
+                }
+                --availableChannels;
+            }
 
             if ( volume > reserved_vols[channel] ) {
                 reserved_vols[channel] = volume;
--- a/src/fheroes2/game/game.cpp
+++ b/src/fheroes2/game/game.cpp
@@ -359,7 +359,8 @@ void Game::EnvironmentSoundMixer()
     }
 
     // Sort positions by distance to the center.
-    std::sort( positions.begin(), positions.end(), []( const fheroes2::Point & p1, const fheroes2::Point & p2 ) { return p1.x * p1.x + p1.y * p1.y < p2.x * p2.x + p2.y * p2.y; } );
+    std::sort( positions.begin(), positions.end(),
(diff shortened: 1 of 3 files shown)


a1exsh (commented):
1. Maybe use stable sort to avoid possible issues when two or more objects are at the same distance?
2. Should we calculate volume based on actual euclidian distance instead of max of x,y?

ihhub:
> 1. Maybe use stable sort to avoid possible issues when two or more objects are at the same distance?
> 
>     2. Should we calculate volume based on actual euclidian distance instead of max of x,y?
1. We always have more than 1 object (except the center) with the same distance. We still calculate the distance, it doesn't matter where the object locates: from left or from right. In other words we don't care what the order for objects on the same distance from the center. Since std::sort is faster than std::stable_sort so we can keep it as it is.
2. Let me try to refactor a bit more as the code for volume is horrible as well...

a1exsh:
> we don't care what the order for objects on the same distance from the center.

That'd be true if we could play them all, but we sort precisely because we want the closest N first: meaning the rest will not be played. If sort is not stable, the first N objects may differ from one run of the function to the other, given the same surroundings.

ihhub:
> > we don't care what the order for objects on the same distance from the center.
> 
> That'd be true if we could play them all, but we sort precisely because we want the closest N first: meaning the rest will not be played. If sort is not stable, the first N objects may differ from one run of the function to the other, given the same surroundings.

Yes, you're right. I've fixed.

ihhub:
Okay. I'll add tomorrow the code to handle the number of available channels so we'll have a full solution.

oleg-derevenetz (approved):

src/fheroes2/game/game.cpp
```diff
@@ -345,23 +345,53 @@ u32 & Game::MapsAnimationFrame( void )
...
+    for ( int32_t y = -maxOffset; y <= maxOffset; ++y ) {
+        for ( int32_t x = -maxOffset; x <= maxOffset; ++x ) {
+            if ( Maps::isValidAbsPoint( x + abs_pt.x, y + abs_pt.y ) ) {
+                positions.emplace_back( x, y );
+            }
+        }
+    }
+
+    // Sort positions by distance to the center.
+    std::stable_sort( positions.begin(), positions.end(),
```
a1exsh, line 369: @ihhub now thinking about it a little bit more, I think we could avoid a sort altogether if we traversed the positions in a different order, in a form of a growing spiral in the loop above.  For that we would probably have to hard-code the 49 possible offsets to check in the desired order.

ihhub: I think it'd be over engineering 🙂 Let's keep it in a simpler to modify and read form. Once we face performance issue we can modify this code.

PR #4709, opened 2021-11-27
Add V-Sync option for SDL2

### LINKED ISSUE #1984: Adventure map, adventure map display that "makes waves" while changing adventure map view
In fheroes2, when the hero moves around on the adventure map or the view of the adventure map is changed with the arrow shaped mouse cursor with the edges of the screen, the displayed adventure map "makes waves".

It is possible to clearly see the problem when the "Hero Speed" is set to "1" and when the "Scroll Speed" is set to "4" in the System Options window and the program is in full screen mode (test in 640x480 resolution).

### LINKED ISSUE #4500: Flickering animations in fullscreen mode
### Preliminary checks

### Platform

Windows

### Describe the bug

I have flickered animations of units, water  (and other thing, they just less noticible) when game in full screen mode. If game in window mode, no matter normal or maximized - all good.

OS: Windows 10 x64 LTSC
GeForce 750M

### Save file

It reproducible in the game independently of  game, map, level or anything else.

### Additional info

_No response_

### LINKED ISSUE #4572: [FR] Add an VYSNC option (Easy SDL2 code included)
### Preliminary checks

Hi!

Would it be possible to have an VSYNC toggle option on the menu?
It's very simple to implement on the SDL2 side of things: here in this exact line in `src/engine/screen.cpp`:

https://github.com/ihhub/fheroes2/blob/058d675daa31195f552432a30e2c356bd9f967e4/src/engine/screen.cpp#L987

...if the VSYNC option on the game menu is ON, it should return:

`return (SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC);
`
instead of:

`return (SDL_RENDERER_ACCELERATED);`

Also, game internal waits for video refresh should be disabled when VSYNC is ON, because **VSYNC should be used as the speed limiter when it's active**.
Yes, I know there are those 144Hz monitors, etc, but these can do without VSYNC as until now.

It's very simple to implement on the SDL2 side of things: here in this exact line in `src/engine/screen.cpp`:

https://github.com/ihhub/fheroes2/blob/058d675daa31195f552432a30e2c356bd9f967e4/src/engine/screen.cpp#L987

...if the VSYNC option on the game menu is ON, it should return:

`return (SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC);
`
instead of:

`return (SDL_RENDERER_ACCELERATED);`

Also, game internal waits for video refresh should be disabled when VSYNC is ON, because **VSYNC should be used as the speed limiter when it's active**.

### Additional info

VSYNC is universally supported in SDL2, every backend has it working unlike in the old SDL1.x days.

I always do my own builds with forced vsync ON, it works perfectly and prevents tearing and flickering like in this issue here:
https://github.com/ihhub/fheroes2/issues/4500

ihhub:
close #4572
close #4500
close #1984

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -797,6 +797,11 @@ namespace
             return new RenderEngine;
         }
 
+        void setVSync( const bool enable ) override
+        {
+            _isVSyncEnabled = enable;
+        }
+
     protected:
         RenderEngine()
             : _window( nullptr )
@@ -804,6 +809,7 @@ namespace
             , _renderer( nullptr )
             , _texture( nullptr )
             , _prevWindowPos( SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED )
+            , _isVSyncEnabled( false )
         {}
 
         void clear() override
@@ -984,8 +990,14 @@ namespace
 
         fheroes2::Size _windowedSize;
 
+        bool _isVSyncEnabled;
+
         int renderFlags() const
         {
+            if ( _isVSyncEnabled ) {
+                return ( SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC );
+            }
+
             return SDL_RENDERER_ACCELERATED;
         }
 
--- a/src/engine/screen.h
+++ b/src/engine/screen.h
@@ -71,6 +71,11 @@ namespace fheroes2
             return fheroes2::Size();
         }
 
+        virtual void setVSync( const bool )
+        {
+            // Do nothing.
+        }
+
     protected:
         BaseRenderEngine()
             : _isFullScreen( false )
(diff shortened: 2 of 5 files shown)


LeHerosInconnu:
Hello @ihhub,

I did a quick test for this for https://github.com/ihhub/fheroes2/issues/1984.
The wave problem seems to be solved with this, but the graphics are blurry when scrolling.
This does not happen when the adventure map view is moved with the mouse from the minimap, although the scrolling is more jerky.
The hero's movement on the adventure map also seems less fluid.

oleg-derevenetz:
I suppose that scrolling/movement is less smooth with this change because the whole application is now blocked in the `SDL_RenderPresent()` until the vsync event happens.

ihhub:
Hi @LeHerosInconnu  and @oleg-derevenetz , you're both right. This option isn't an ideal one but it fixes flickering on some systems. Since it's a blocking call FPS will be capped at 60. This explains why the motion is less fluid (at speed 9 without this option we have 125 FPS). If given a choice between heavy flickering or less fluid animation people will choose the second one at least to be able to play the game.

Issue #4367, opened 2021-10-10
Castle defenders pop-up position of troops / text


a1exsh:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Linux

### Describe the bug

The positioning of troops and text in the castle defenders pop-up window seems to be off, touching the bottom border:
![image](https://user-images.githubusercontent.com/489601/136707168-aa641212-91a1-4c72-b442-f61db00e4d9f.png)

### Save file

[The_Clearing_0055.zip](https://github.com/ihhub/fheroes2/files/7317923/The_Clearing_0055.zip)

zenseii:
I'll have a go at this one.

PR #4711, opened 2021-11-27
Fix position of armies for quick info window for both castle and enemy hero

zenseii:
This relates to and gives a solution for issue https://github.com/ihhub/fheroes2/issues/4367

The OG implementation has its flaws so trying to replicate it 100% is no use. Here's an example of the original Heroes 2:
image

This is the same enemy castle in fheroes2 after my fix:
image

Changes made:
* Moved the creatures in the castle quick info higher up and  more to the center of the window
* Made the creature quantity and the quantity text (:swarm) appear right below the center of the creature icon (which do note is not always beneath a bipedal creature's feet!)
* Changed to fheroes2::Text and made necessary adaptments
* ~~Added parantheses in a short-hand if else inline statement for readability~~

The current implementation in the code (which I believe is part of the old code) doesn't let you use a different space between the first and second row of creatures in an army and in a castle - the space is shared between them. This is why creatures in an enemy army appear slightly closer than before my fix. In the current code you can only differentiate where to display the entire bulk of the army uniquely for each window.

Here is an enemy hero army after my fix:

image

The best would be to have one amount of spacing between first and second row for the castle and one for the enemy army, but I wasn't able to find any way of implementing that. There is already this for a castle where a hero is a guardian so I could create a similar solution to that maybe, but that would mean adding another parameter to the DrawMons32Line function, which seems a bit overkill.

Black Dragons are used because they take up the most space from what I could tell. The Titan's mohawk is at a good distance from whatever is above it now too.


zenseii:
It would be great if @LeHeroesInconnu could have a look at this with his pixel perfect sight 👍

LeHerosInconnu:
Hello @zenseii,

> It would be great if @LeHerosInconnu could have a look at this with their pixel perfect sight +1

I'll take a look at this tomorrow, it's time to quit. :)

zenseii:
> Hello @zenseii,
> 
> > It would be great if @LeHerosInconnu could have a look at this with their pixel perfect sight +1
> 
> I'll take a look at this tomorrow, it's time to quit. :)

Thanks! Yeah, don't feel any pressure - do it when you feel like it 👍

src/fheroes2/army/army.cpp
```diff
@@ -679,27 +680,28 @@ void Troops::DrawMons32Line( int32_t cx, int32_t cy, uint32_t width, uint32_t fi
         if ( !compact )
             cx += chunk / 2;
 
-        Text text;
```
ihhub, line 682: I'm thinking that this method should not belong to this file at all. This is the only function which does some UI stuff while the rest of code is for logic. It's always good to separate logic and UI. Can we please do this:
- create 2 files: **army_ui_helper.h** and **army_ui_helper.cpp**
- in the header file we add similar method within **fheroes2** namespace
- in the source file we add the code
- in this **army.cpp** we include newly added header and call the respective method

?

I know it looks a bit of counterproductive but it eases the code readability. What do you think?

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -684,20 +685,21 @@ void Dialog::QuickInfo( const Castle & castle, const fheroes2::Rect & activeArea
 
     cur_rt = fheroes2::Rect( cur_rt.x + 22, cur_rt.y + 9, 192, 154 );
     fheroes2::Point dst_pt;
-    Text text;
+    fheroes2::Text text;
```
ihhub, line 688: Please remove this line and declare **text** at line 692 instead of **set** call.

src/fheroes2/army/army.cpp
```diff
@@ -679,27 +680,28 @@ void Troops::DrawMons32Line( int32_t cx, int32_t cy, uint32_t width, uint32_t fi
...
                         fheroes2::Blit( monster, fheroes2::Display::instance(), cx + offset, cy + offsetY + monster.y() );
-                        text.Blit( cx + chunk - text.w() - offset, cy + 23 );
+                        text.draw( cx + chunk - text.width() - offset, cy + 23, display );
                     }
                     else {
-                        const int offsetY = 30 - monster.height();
-                        fheroes2::Blit( monster, fheroes2::Display::instance(), cx - monster.width() / 2 + monster.x(), cy + offsetY + monster.y() );
-                        text.Blit( cx - text.w() / 2, cy + 29 );
+                        const int offsetY = 28 - monster.height();
+                        fheroes2::Blit( monster, fheroes2::Display::instance(), cx - monster.width() / 2 + monster.x() + 2, cy + offsetY + monster.y() );
```
ihhub, line 703: Please use **display** here instead of `fheroes2::Display::instance()`.

ihhub: I'd even go further and accept **Image & output** parameter in this function. This will be useful for the future upgrade of rendering when we're going to implement up-scaling.

ihhub (changes requested):
Hi @zenseii , I put few comments in this pull request. Would you mind to take a look at them?

src/fheroes2/army/army.cpp
zenseii: Yes, this sounds like a good solution. I'm working on it. I noticed that there are two instances of DrawMons32Line in two classes, Army and Troop. Should I bring them over like this to the new namespace in the source file? There's also the method DrawMonsterLine which I've brought over to the new header and source file.

LeHerosInconnu:
Hello @zenseii,

> > Hello @zenseii,
> > > It would be great if @LeHerosInconnu could have a look at this with their pixel perfect sight :+1:
> > 
> > 
> > I'll take a look at this tomorrow, it's time to quit. :)
> 
> Thanks! Yeah, don't feel any pressure - do it when you feel like it :+1:

Here are my comments.

- The vertical position of the creature icons and creature numbers in the hero information window can be moved down one pixel vertically, which will put them at the same distance from the bottom of the window as in the city information window, and at the same time will move the top of the Dragon and Titan sprites away from the line of text just above them.

Info window PR test coupé 006

Anyway this window will need to be enlarged later, if only to put the frame around the hero's portrait. :) (https://github.com/ihhub/fheroes2/discussions/3330, https://github.com/ihhub/fheroes2/discussions/4687)

- In the city information window, the city name can be moved down one pixel vertically, which will put it at the same distance from the top of the window as the hero name in the hero information window.

Info window PR test coupé 007

Everything else seems fine.
Nice. :)

src/fheroes2/army/army.cpp
zenseii: So the problem I have been struggling with is that two of the methods in `army.h` depend on the class `Troops` by taking this as a parameter: `const Troops & troops`. The methods are `Army::DrawMons32Line` and `Army::DrawMonsterLine`, they are found in the class `Army` which inherits the class `Troops`.

Here is what I currently have for  `army_ui_helper.h` (the class identifiers are WIP):
```c++
#pragma once

#include <string>
#include <vector>

#include "army.h"

namespace fheroes2
{
    class TroopMonsters : public Troops
    {
    public:
        void DrawMons32Line( int32_t, int32_t, uint32_t, uint32_t, uint32_t, uint32_t, bool, bool ) const;
    };

    class ArmyMonsters : public Army, public TroopMonsters
    {
    public:
        static void DrawMons32Line( const Troops &, s32, s32, u32, u32 = 0, u32 = 0 );
        static void DrawMonsterLines( const Troops & troops, int32_t posX, int32_t posY, uint32_t lineWidth, uint32_t drawPower, bool compact = true,
                                      bool isScouteView = true );
    };
}
```

The error message is not surprisingly : 'DrawMons32Line' is not a member of 'Troops'.
Changing the method parameters to `const TroopMonsters & troops` to give access to `DrawMons32Line` gives the issue where other code in the project that uses this function expects to be able to call other methods found in the original `Troops` when using the object `troops`.

src/fheroes2/army/army.cpp
zenseii: Sure! I'll do this once I'm able to refactor `Troop::DrawMons32Line` from "army.h" into army_"ui_helper.h".

src/fheroes2/army/army.cpp
ihhub: Actually I asked to move only this function and then call this function within `void Army::DrawMons32Line` and `void Army::DrawMonsterLines` methods. Please revert changes for these function but move this function into newly created files.

ihhub (changes requested):

zenseii:
@ihhub I've attempted to move the declaration and definition of `void Troops:DrawMons32Line` and call it from the new location `void fheroes2::DrawTroops::DrawMonsLine32` within `Army::DrawMons32Line` and `Army::DrawMonsterLine`. 

I made it into a class so that I could inherit from `Troops` because if not the functions local to Eroops like `GetCount()` and `isValid()` would not be accessible.

I mostly understand what the issues are, like the definition of `Army::DrawMons32Line` ends up with an unused parameter `const Troops & troops`, but changing this to `const fheroes2::DrawTroops` gives this error from functions that used `Troops::DrawMons32Line`: "cannot convert argument 1 from 'const Troops' to 'const fheroes2::DrawTroops &'. Maybe I shouldn't be passing by reference.

I also need to figure out how to pass `Image & output` as an argument in the places needed, because initializing it like I have done is obviously wrong.

src/fheroes2/army/army.cpp
```diff
@@ -1257,22 +1216,25 @@ void Army::DrawMons32Line( const Troops & troops, s32 cx, s32 cy, u32 width, u32
     troops.DrawMons32Line( cx, cy, width, first, count, Skill::Level::EXPERT, false, true );
```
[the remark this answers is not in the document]
zenseii: @ihhub this here is basically the problem with moving the method.

src/fheroes2/army/army.cpp
zenseii: Yes, I've done this now and left a comment further down about the issues I've run into.

src/fheroes2/army/army_ui_helper.h
```diff
@@ -0,0 +1,32 @@
...
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#pragma once
+
+#include "army.h"
+
+namespace fheroes2
+{
+    class DrawTroops : public Troops
```
ihhub, line 27: Please remove this class completely here. This function must be just a part of fheroes2 namespace.

src/fheroes2/army/army_ui_helper.h
```diff
@@ -0,0 +1,32 @@
...
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#pragma once
+
+#include "army.h"
```
ihhub, line 23: We don't need this file inclusion after removing the class.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    if ( isValid() ) {
+        if ( 0 == count )
+            count = GetCount();
+
+        const int chunk = width / count;
+        if ( !compact )
+            cx += chunk / 2;
+
+        for ( const_iterator it = begin(); it != end(); ++it ) {
```
ihhub, line 42: This function should receive an extra parameter `const Army & army` from where we can get all individual troops and iterate through them.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "army_ui_helper.h"
+#include "agg_image.h"
+#include "army_troop.h"
+#include "game.h"
+#include "icn.h"
+#include "image.h"
+#include "monster.h"
+#include "screen.h"
+#include "ui_text.h"
+
```
ihhub, line 30: Please remove this empty line.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,70 @@
...
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#include "army_ui_helper.h"
+#include "agg_image.h"
+#include "army_troop.h"
+#include "game.h"
```
ihhub, line 24: Do we need this file header inclusion?

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "icn.h"
+#include "image.h"
+#include "monster.h"
```
ihhub, line 27: Do we need this file header inclusion?

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "screen.h"
+#include "ui_text.h"
+
+
+void fheroes2::DrawTroops::DrawMons32Line( int32_t cx, int32_t cy, uint32_t width, uint32_t first, uint32_t count, uint32_t drawPower, bool compact, bool isScouteView ) const
+{
+    if ( isValid() ) {
```
ihhub, line 34: We can do this:
```cpp
if ( !army.isValid() ) {
    return;
}

...
// the rest of the code.
```
In this case we don't need to have text indentation.

src/fheroes2/army/army.cpp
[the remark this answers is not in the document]
ihhub: Hi @zenseii , as I explained in a comment below we shouldn't make this function as a method of a class. Please check my suggestions.

ihhub (changes requested):
Hi @zenseii , I left several more comments here. Would you mind to check them?

src/fheroes2/army/army_ui_helper.cpp
zenseii: game.h is needed for this part

```cpp
fheroes2::Text text( isScouteView ? Game::CountScoute( ( *it )->GetCount(), drawPower, compact )
				   : Game::CountThievesGuild( ( *it )->GetCount(), drawPower ),
```

src/fheroes2/army/army_ui_helper.cpp
zenseii: It seems not.

src/fheroes2/army/army_ui_helper.cpp
ihhub: Oh, okay.

zenseii:
@ihhub I've looked at this code up and down for quite a while now and I'm still not able to come up with a solution.

Here's a summary of what the 2 outstanding problems are as far as I understand:

Problem 1:

The for loop cannot iterate through the creatures to draw because const_iterator and the functions begin() and end() are inaccessible due to the class Troops having these from protected inheritance from std::vector<Troop *> :

https://github.com/ihhub/fheroes2/blob/52453798ce3572577b45520f7b9b802a07ccbadf/src/fheroes2/army/army.h#L42

https://github.com/ihhub/fheroes2/blob/52453798ce3572577b45520f7b9b802a07ccbadf/src/fheroes2/army/army.cpp#L731




Problem 2:

The Army::DrawMonsterLine32() cannot take class Army as a parameter because it needs to use const Troops, because other places in the code refering to this function use the class type Troops. 
Example of where this happens in battle_dialogs.cpp:

https://github.com/ihhub/fheroes2/blob/52453798ce3572577b45520f7b9b802a07ccbadf/src/fheroes2/battle/battle_dialogs.cpp#L382

https://github.com/ihhub/fheroes2/blob/52453798ce3572577b45520f7b9b802a07ccbadf/src/fheroes2/battle/battle_dialogs.cpp#L471-L472

Visual studio would give this message if using const Army & as parameter: "no suitable user-defined conversion from "const Troops" to "const Army" exists."


My current solution to this is to give the new fheroes2::DrawMonsterLine32 the const Troops & parameter instead of const Army:

```cpp
namespace fheroes2
{
    void DrawMons32Line( const Troops & troops, int32_t cx, int32_t cy, uint32_t width, uint32_t first, uint32_t count, uint32_t drawPower, bool compact,
                         bool isScouteView );
}
```

ihhub:
Hi @zenseii , 

Problem 1: the class has `size_t Size( void ) const;` method which returns the number of slots used by monsters and `const Troop * GetTroop( size_t ) const;` method to get a pointer to a certain troop.

Problem 2: yes, we're expecting only `Troops` class instance. `Army` is a child class of it so we should use `Troops`.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,75 @@
...
+    if ( !compact ) {
+        cx += chunk / 2;
+    }
+
+    for ( size_t slot = 0; slot <= troops.Size(); ++slot ) {
+        const Troop * troop = troops.GetTroop( slot );
+        if ( troop && troop->isValid() ) {
+            if ( 0 == first && count ) {
+                const fheroes2::Sprite & monster = fheroes2::AGG::GetICN( ICN::MONS32, troop->GetSpriteIndex() );
+                fheroes2::Text text( isScouteView ? Game::CountScoute( troop->GetCount(), drawPower, compact )
```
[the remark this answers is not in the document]
zenseii: @ihhub Is there anything I need to do about this warning and the one below?

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,75 @@
...
+                                                  : Game::CountThievesGuild( troop->GetCount(), drawPower ),
```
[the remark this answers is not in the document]
zenseii: See above.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "agg_image.h"
+#include "army.h"
+#include "army_troop.h"
+#include "game.h"
+#include "icn.h"
+#include "image.h"
+#include "screen.h"
+#include "ui_text.h"
+
+void fheroes2::DrawMons32Line( const Troops & troops, int32_t cx, int32_t cy, uint32_t width, uint32_t first, uint32_t count, uint32_t drawPower, bool compact,
```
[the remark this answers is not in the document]
zenseii: @ihhub Is there anything I need to do about this one? I'm working on adding Image & output as a parameter to this function like you requested too, so in the end there will be 32 parameters.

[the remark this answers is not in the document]
ihhub: Nope.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    const int chunk = width / count;
+    if ( !compact ) {
+        cx += chunk / 2;
+    }
+
+    for ( size_t slot = 0; slot <= troops.Size(); ++slot ) {
+        const Troop * troop = troops.GetTroop( slot );
+        if ( troop && troop->isValid() ) {
```
ihhub, line 49: To reduce complexity highlighted by clang-tidy please do this trick:
```cpp
if ( troop == nullptr || !troop->isValid() ) {
    continue;
}
... // the rest of the code
```

It will greatly improve code readability as well.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,77 @@
...
+                else {
+                    const int offsetY = 28 - monster.height();
+                    fheroes2::Blit( monster, output, cx - monster.width() / 2 + monster.x() + 2, cy + offsetY + monster.y() );
+                    text.draw( cx - text.width() / 2, cy + 29, output );
+                }
+                cx += chunk;
+                --count;
+            }
+            else
+                --first;
```
ihhub, line 74: Please add {} around this code block:
```cpp
else {
    --first;
}
```

src/fheroes2/army/army_ui_helper.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <cstdint>
+
+class Troops;
+
+namespace fheroes2
+{
+    class Image;
+
+    void DrawMons32Line( const Troops & troops, int32_t cx, int32_t cy, uint32_t width, uint32_t first, uint32_t count, uint32_t drawPower, bool compact,
```
ihhub, line 31: I am thinking that this name is not human-readable (not blaming you for this as it is just copy-paste). Since we are modifying this code I suggest to name the function as `drawMiniMonsters`. No point to specify magical number 32. What you think?

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -708,20 +709,21 @@ void Dialog::QuickInfo( const Castle & castle, const fheroes2::Rect & activeArea
...
+    const fheroes2::FontType smallWhiteFont{ fheroes2::FontSize::SMALL, fheroes2::FontColor::WHITE };
 
     // castle name
-    text.Set( castle.GetName(), Font::SMALL );
-    dst_pt.x = cur_rt.x + ( cur_rt.width - text.w() ) / 2;
-    dst_pt.y = cur_rt.y;
-    text.Blit( dst_pt.x, dst_pt.y );
+    text.set( castle.GetName(), smallWhiteFont );
```
ihhub, line 716: Please combine these 3 lines into one:
```cpp
fheroes2::Text text( castle.GetName(), fheroes2::FontType::smallWhite() );
```

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -782,10 +784,10 @@ void Dialog::QuickInfo( const Castle & castle, const fheroes2::Rect & activeArea
     // show guardian
     if ( isGuardianVisible ) {
         // hero name
-        text.Set( guardian->GetName(), Font::SMALL );
-        dst_pt.x = cur_rt.x + ( cur_rt.width - text.w() ) / 2;
+        text.set( guardian->GetName(), smallWhiteFont );
```
ihhub, line 787: Please use this call instead of the current:
```cpp
text.set( guardian->GetName(), fheroes2::FontType::smallWhite() );
```

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -796,34 +798,34 @@ void Dialog::QuickInfo( const Castle & castle, const fheroes2::Rect & activeArea
         }
     }
     else {
-        text.Set( _( "Defenders:" ) );
-        dst_pt.x = cur_rt.x + ( cur_rt.width - text.w() ) / 2;
-        dst_pt.y += castleIcon.height() + 5;
-        text.Blit( dst_pt.x, dst_pt.y );
+        text.set( _( "Defenders:" ), smallWhiteFont );
```
ihhub, line 801: Please `fheroes2::FontType::smallWhite()` and for the rest of places in the code below.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    const int chunk = width / count;
+    if ( !compact ) {
+        cx += chunk / 2;
+    }
+
+    for ( size_t slot = 0; slot <= troops.Size(); ++slot ) {
+        const Troop * troop = troops.GetTroop( slot );
+        if ( troop && troop->isValid() ) {
+            if ( 0 == first && count ) {
```
ihhub, line 50: Also to reduce the complexity of this code we can do like this:
```cpp
if ( first != 0 || count == 0 ) {
    --first;
    continue;
}
... // the rest of the code without else part.
```

src/fheroes2/army/army_ui_helper.h
ihhub, line 31: Please make the names of input arguments the same as in implementation of the function.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,77 @@
...
+                const fheroes2::Sprite & monster = fheroes2::AGG::GetICN( ICN::MONS32, troop->GetSpriteIndex() );
+                fheroes2::Text text( isScouteView ? Game::CountScoute( troop->GetCount(), drawPower, compact ) : Game::CountThievesGuild( troop->GetCount(), drawPower ),
+                                     { fheroes2::FontSize::SMALL, fheroes2::FontColor::WHITE } );
```
ihhub, line 53: Please use `fheroes2::FontType::smallWhite()` instead of this code.

src/fheroes2/army/army_ui_helper.cpp
ihhub: With the proposed above complexity reduction approach we do not need to do this change.

ihhub (changes requested):
Hi @zenseii , I left multiple comments in this pull request. Could you please take a look at them?

src/fheroes2/army/army_ui_helper.h
zenseii: @ihhub I believe the names of the input arguments are the same between the header and the code file, but maybe I'm missing something?

zenseii:
Thank you so much for the comments and feedback, @ihhub . I have addressed them and checked that everything works well. When it suits you it would be nice if you could have another look and see if this code is now ready.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,78 @@
...
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#include "army_ui_helper.h"
+#include "agg_image.h"
+#include "army.h"
+#include "army_troop.h"
+#include "game.h"
+#include "icn.h"
+#include "image.h"
```
ihhub, line 27: We can remove this header inclusion as it is included in screen.h file.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -0,0 +1,78 @@
...
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#include "army_ui_helper.h"
+#include "agg_image.h"
+#include "army.h"
```
ihhub, line 23: Do we need this file inclusion?

ihhub (commented):
Hi @zenseii , I left just 2 comments related to file header inclusion. The rest looks good. Could you please check them?

src/fheroes2/army/army_ui_helper.cpp
zenseii: `army.h` is needed for the class type Troops, while `army_troop.h` is needed for the class type Troop.

src/fheroes2/army/army_ui_helper.cpp
zenseii: Done! Thanks!

zenseii: Oh wait I did the opposite...

zenseii: Although it seems to be the logical choice to only include image.h since we don't need the extra stuff from screen.h anymore.

ihhub:
@zenseii , thanks a lot for fixing this issue!

PR #4712, opened 2021-11-27
Use the OK button for the Tavern window

zenseii:
Currently in the Tavern there's the "Yes" button. The "OK" button is used in the OG and this is also what makes sense.


ihhub:
@zenseii , thanks a lot for spotting this bug and fixing it!

PR #4720, opened 2021-11-28
Allow to select player class using mouse wheel

a1exsh:
Makes it easier to go back (and forward) through the player classes.

### DIFF
--- a/src/fheroes2/gui/player_info.cpp
+++ b/src/fheroes2/gui/player_info.cpp
@@ -19,6 +19,7 @@
  ***************************************************************************/
 
 #include <algorithm>
+#include <cassert>
 
 #include "agg_image.h"
 #include "dialog.h"
@@ -59,6 +60,8 @@ namespace
             player.SetRace( Race::KNGT );
             break;
         default:
+            // Did you add a new race? Add the logic above
+            assert( 0 );
             break;
         }
     }
@@ -88,6 +91,8 @@ namespace
             player.SetRace( Race::NECR );
             break;
         default:
+            // Did you add a new race? Add the logic above
+            assert( 0 );
             break;
         }
     }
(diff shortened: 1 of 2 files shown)


src/fheroes2/gui/player_info.cpp
```diff
@@ -295,8 +356,21 @@ bool Interface::PlayersInfo::QueueEventProcessing( void )
...
+        }
+    }
+    else if ( le.MouseWheelDn() ) {
+        if ( nullptr != ( player = GetFromClassClick( le.GetMouseCursor() ) ) ) {
+            if ( conf.AllowChangeRace( player->GetColor() ) ) {
+                changeRaceToNext( *player );
+            }
+        }
+    }
+    else { // if ( le.MouseClickLeft() )
```
ihhub, line 373: Please remove this commented code.

ihhub: Actually it'll be even better if we add this condition explicitly.

src/fheroes2/gui/player_info.cpp
```diff
@@ -295,8 +356,21 @@ bool Interface::PlayersInfo::QueueEventProcessing( void )
...
                 Font::BIG );
     }
-    // le.MouseClickLeft()
-    else {
+    else if ( le.MouseWheelUp() ) {
+        if ( nullptr != ( player = GetFromClassClick( le.GetMouseCursor() ) ) ) {
+            if ( conf.AllowChangeRace( player->GetColor() ) ) {
+                changeRaceToPrev( *player );
+            }
+        }
```
ihhub, line 364: Please don't do two commands in one line (I totally understand it was copy-pasted from another code):
```cpp
        player = GetFromClassClick( le.GetMouseCursor() );
        if ( player != nullptr ) ) {
            if ( conf.AllowChangeRace( player->GetColor() ) ) {
                changeRaceToPrev( *player );
            }
        }
```
The same applies for the condition below.

src/fheroes2/gui/player_info.cpp
```diff
@@ -32,6 +32,67 @@
...
+            player.SetRace( Race::WRLK );
+            break;
+        case Race::NECR:
+            player.SetRace( Race::WZRD );
+            break;
+        case Race::RAND:
+            player.SetRace( Race::NECR );
+            break;
+        default:
+            break;
```
ihhub, line 91: I'd suggest to add an assertion here:
```cpp
// Did you add a new race? Add the logic above
assert( 0 );
```

src/fheroes2/gui/player_info.cpp
```diff
@@ -32,6 +32,67 @@
...
+            player.SetRace( Race::NECR );
+            break;
+        case Race::NECR:
+            player.SetRace( Race::RAND );
+            break;
+        case Race::RAND:
+            player.SetRace( Race::KNGT );
+            break;
+        default:
+            break;
```
ihhub, line 62: I'd suggest to add an assertion here:
```cpp
// Did you add a new race? Add the logic above
assert( 0 );
```

ihhub (changes requested):
Hi @a1exsh , I put several comments in this pull request. Can you please check them once you have time?

src/fheroes2/gui/player_info.cpp
a1exsh: Was my idea as well, but we have a `const LocalEvent` here and somehow of all the checks that we do here `MouseClickLeft()` is not declared `const`.  I don't really want to open that can of worms :)

a1exsh: @ihhub if you have ideas what to do here, please suggest

ihhub:

PR #4722, opened 2021-11-28
Fix Enter button pressed state in gift dialog

### LINKED ISSUE #4717: Gift resource dialog: ENTER hotkey is leaked
### Preliminary checks

### Platform

Linux

### Describe the bug

Go to a town where you have a marketplace and plan a gift to some player.
After selecting the count, use `ENTER` hotkey instead of clicking the button with the mouse:
image

The count selection dialog is closed, but it also closes the gift dialog, as if one clicked the `OKAY` button.  This is because the events from the buttons are handled after all other events in the loop: https://github.com/ihhub/fheroes2/blob/master/src/fheroes2/dialog/dialog_giftresources.cpp#L276

### Save file

The issue can be reproduced in any game, no save file needed.

### Additional info

_No response_

vincent-grosbois:


ihhub:
@vincent-grosbois , thank you so much for the fix!

PR #4728, opened 2021-11-30
Fix Summon Boat logic

### LINKED ISSUE #2919: Adventure map, Summon Boat spell, the boat is summoned at a location not accessible by the hero while another location is available that would allow him to board immediately
In fheroes2, the boat is summoned at a location not accessible by the hero while another location is available that would allow him to board immediately.

In fheroes2:
The boat can be summoned at three locations.
This is fine and normal.
Summon Boat locations fh2 01

The boat could have been summoned at the bottom location, rather than at the top location and be directly accessible.
Summon Boat locations fh2 02

https://user-images.githubusercontent.com/43583994/110320064-0b28fe00-8010-11eb-99cb-564333780dde.mp4

Here is the save file:
End the turn.
[Before summon boat 1000.zip](https://github.com/ihhub/fheroes2/files/6101242/Before.summon.boat.1000.zip)

Related closed issue: https://github.com/ihhub/fheroes2/issues/2172.

ihhub:
close #2919

### DIFF
--- a/src/fheroes2/heroes/heroes_spell.cpp
+++ b/src/fheroes2/heroes/heroes_spell.cpp
@@ -365,9 +365,6 @@ bool ActionSpellSummonBoat( const Heroes & hero )
     }
 
     const int32_t center = hero.GetIndex();
-    const int32_t worldWidth = world.w();
-    const int32_t centerX = center % worldWidth;
-    const int32_t centerY = center / worldWidth;
 
     const int tilePassability = world.GetTiles( center ).GetPassable();
 
--- a/src/fheroes2/heroes/heroes_spell.cpp
+++ b/src/fheroes2/heroes/heroes_spell.cpp
@@ -365,30 +365,54 @@ bool ActionSpellSummonBoat( const Heroes & hero )
     }
 
     const int32_t center = hero.GetIndex();
-    const fheroes2::Point & centerPoint = Maps::GetPoint( center );
+    const int32_t worldWidth = world.w();
+    const int32_t centerX = center % worldWidth;
+    const int32_t centerY = center / worldWidth;
 
-    // find water
-    int32_t dst_water = -1;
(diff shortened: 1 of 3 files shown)


src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -365,30 +365,54 @@ bool ActionSpellSummonBoat( const Heroes & hero )
...
+        possibleBoatPositions.emplace_back( center - worldWidth - 1 );
+    }
+
+    if ( ( tilePassabilty & Direction::BOTTOM_RIGHT ) != 0 && Maps::isValidAbsPoint( centerX + 1, centerY + 1 ) ) {
+        possibleBoatPositions.emplace_back( center + worldWidth + 1 );
+    }
+
+    if ( ( tilePassabilty & Direction::BOTTOM_LEFT ) != 0 && Maps::isValidAbsPoint( centerX - 1, centerY + 1 ) ) {
+        possibleBoatPositions.emplace_back( center + worldWidth - 1 );
+    }
```
oleg-derevenetz, line 404: Are there any reasons not to use `Maps::getAroundIndexes()` here?

oleg-derevenetz (commented):
Hi @ihhub I left just one question here, could you take a look?

src/fheroes2/heroes/heroes_spell.cpp
ihhub: `getAroundIndexes` doesn't take into account the current tile passability. The current tile might now allow to move to the top while `getAroundIndexes` will still return the tile above.

oleg-derevenetz: I believe we can check it later in the cycle by getting a direction to the candidate tile using the `Maps::GetDirection()` and comparing this direction with the `tilePassabilty`. I believe there is a typo BTW: `tilePassabilty`-> `tilePassability` :)

oleg-derevenetz (approved):

Issue #4438, opened 2021-10-16
Scenario window, title of the "OKAY" button information window should be "Okay" instead of "OK"


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the title of the "OKAY" button information window should be "Okay" instead of "OK".

In fheroes2:

![Okay scenario 01](https://user-images.githubusercontent.com/43583994/137589701-6f3bef39-9de2-4043-a623-55c8611658fb.png)

### Save file

Not needed.

ihhub:
We have 4 places with OK and only 1 place with Okay. Should we then keep everything the same?

LeHerosInconnu:
Hello @ihhub,

> We have 4 places with OK and only 1 place with Okay. Should we then keep everything the same?

Yes, all titles should be "Okay" to stay consistent with the other titles like, for example, "Cancel".

In fheroes2:

![Title Cancel 01](https://user-images.githubusercontent.com/43583994/137598050-1d626a91-2f61-4ea3-a14d-aa9fe2167d43.png)

PR #4736, opened 2021-12-01
Add popup window to OK button in game settings

ihhub:
tiny but consistent feature


LeHerosInconnu:
Hello @ihhub,

"In fheroes2, the title of the "OKAY" button information window should be "Okay" instead of "OK"." :)
Related issue: https://github.com/ihhub/fheroes2/issues/4438.

In fheroes2:

Okay instead OK 01

There are other buttons that do not have a right-click information window.
Do you prefer that this be reported in a single issue or in separate issues?

PR #4740, opened 2021-12-03
Update background music on the fly

ihhub:
when user changes the type of music

### DIFF
--- a/src/fheroes2/dialog/dialog_system_options.cpp
+++ b/src/fheroes2/dialog/dialog_system_options.cpp
@@ -310,12 +310,11 @@ namespace
                 if ( type == MUSIC_MIDI_EXPANSION && !conf.isPriceOfLoyaltySupported() )
                     ++type;
 
-                const int currentMusic = Game::CurrentMusic();
+                const Game::MusicRestorer musicRestorer;
 
                 conf.SetMusicType( type > MUSIC_EXTERNAL ? 0 : type );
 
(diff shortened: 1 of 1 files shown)


src/fheroes2/dialog/dialog_system_options.cpp
```diff
@@ -310,7 +310,13 @@ namespace
                 if ( type == MUSIC_MIDI_EXPANSION && !conf.isPriceOfLoyaltySupported() )
                     ++type;
 
+                const int currentMusic = Game::CurrentMusic();
+
                 conf.SetMusicType( type > MUSIC_EXTERNAL ? 0 : type );
+
+                Game::SetCurrentMusic( -1 );
+                AGG::PlayMusic( currentMusic, true );
+
```
oleg-derevenetz, line 319: It's optional, but I suppose that we can use the `MusicRestorer` class like this:

```
const Game::MusicRestorer musicRestorer;

conf.SetMusicType( type > MUSIC_EXTERNAL ? 0 : type );

Game::SetCurrentMusic( MUS::UNKNOWN );
```

oleg-derevenetz (commented):
Hi @ihhub I just left one optional suggestion, please take a look when it's convenient for you.

src/fheroes2/dialog/dialog_system_options.cpp
ihhub: I've changed it.

Issue #1632, opened 2020-09-11
Hero, Magic Book, spell info window, add the "Game: Display additional spell info in Magic Book" option


LeHerosInconnu:
In fheroes2, it would be nice to add an option that display some additional information about the spell in the spell info windows.

A sentence is added in the spell's information window to indicate additional information about the effect of the spell depending on the hero who uses it.
An additional sentence is added when the hero has an artifact that modifies the effect of the spell.

For the damage spells "Magic Arrow", "Death Ripple", etc.:
Example with the "Lightning Bolt" spell:
Without artifact that modifies the effect of the spell:
"This spell does 50 points of damage when Natasha casts it."

With the artifact "Lighting Rod":
"Natasha has the Lighting Rod artifact which increases the damage of this spell by 50%.
This spell does 75 damage when Natasha casts it."

For the "Cure" and "Mass Cure" spells:
Example with the "Cure" spell:
"This spell restores 10 HP when Natasha casts it."

For the resurrection spells "Animate Dead", "Resurrect", etc.:
Example with the "Resurrect True" spell:
With no artifact that modifies the effect of the spell:
"This spell restores 100 HP when Natasha casts it."

With the artifact "Ankh":
"Natasha has the Ankh artifact which doubles the effectiveness of this spell.
This spell restores 200 HP when Natasha casts it."

For "Set Guardian" spells:
Example with the spell "Set Earth Guardian":
"This spell summons 8 Earth Elementals to guard the mine when Natasha casts it."

For the "Town Gate" spell:
The name of the nearest village/castle and the possibility to welcome the hero are indicated.

Example with a castle occupied by another hero:
“The nearest castle is Algary.
This castle is occupied by your hero Halon."

Example with a town which is not occupied:
"The nearest town is Sorpigal.
This town can welcome you."

For "Summon Elemental" spells:
Example with the spell "Summon Air Elemental":
Without artifact which modifies the effect of the spell:
"This spell summons 6 Air Elementals when Natasha casts it."

With the "Book of Elements" artifact:
"Natasha has the Book of Elements artifact which doubles the effectiveness of this spell.
This spell summons 12 Air Elementals when Natasha casts it."

With the "Book of Elements" and "Elemental Ring" artifacts:
"Natasha has the Book of Elements artifact which doubles the effectiveness of this spell and the Elemental Ring which halves the cost of this spell.
This spell summons 12 Air Elementals when Natasha casts it for only half the cost of the spell."

For "Earthquake" spell:
[...]

PR #4761, opened 2021-12-07
Add additional spell info for spells owned by a hero

ihhub:
relates to #1632

image

### DIFF
--- a/fheroes2-vs2015.vcxproj
+++ b/fheroes2-vs2015.vcxproj
@@ -373,6 +373,7 @@
     <ClCompile Include="src\fheroes2\resource\resource.cpp" />
     <ClCompile Include="src\fheroes2\spell\spell.cpp" />
     <ClCompile Include="src\fheroes2\spell\spell_book.cpp" />
+    <ClCompile Include="src\fheroes2\spell\spell_info.cpp" />
     <ClCompile Include="src\fheroes2\spell\spell_storage.cpp" />
     <ClCompile Include="src\fheroes2\system\bitmodes.cpp" />
     <ClCompile Include="src\fheroes2\system\players.cpp" />
@@ -544,6 +545,7 @@
     <ClInclude Include="src\fheroes2\resource\resource.h" />
     <ClInclude Include="src\fheroes2\spell\spell.h" />
     <ClInclude Include="src\fheroes2\spell\spell_book.h" />
+    <ClInclude Include="src\fheroes2\spell\spell_info.h" />
     <ClInclude Include="src\fheroes2\spell\spell_storage.h" />
     <ClInclude Include="src\fheroes2\system\bitmodes.h" />
     <ClInclude Include="src\fheroes2\system\gamedefs.h" />
--- a/fheroes2-vs2019.vcxproj
+++ b/fheroes2-vs2019.vcxproj
@@ -374,6 +374,7 @@
     <ClCompile Include="src\fheroes2\resource\resource.cpp" />
     <ClCompile Include="src\fheroes2\spell\spell.cpp" />
     <ClCompile Include="src\fheroes2\spell\spell_book.cpp" />
+    <ClCompile Include="src\fheroes2\spell\spell_info.cpp" />
     <ClCompile Include="src\fheroes2\spell\spell_storage.cpp" />
     <ClCompile Include="src\fheroes2\system\bitmodes.cpp" />
     <ClCompile Include="src\fheroes2\system\players.cpp" />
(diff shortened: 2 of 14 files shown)


oleg-derevenetz (approved):

LeHerosInconnu:
@ihhub.

> Hi @Branikolog and @LeHerosInconnu, please take a look at this implementation.
> 
> The description doesn't depend on artifacts (you won't see extra info about special artifact effect) and I emitted hero's name as this spell's info dialog can be shown only from a hero dialog / active hero. Earthquake spell I left as it is for now.
> 
> Overall, if you're okay with the way how it's implemented we can proceed later with fine tuning.

I did some tests, everything seems to work fine. :)
And about the Hypnotize, Haunt, Set Elemental Guardians spells, Town Gate (indicate the name of the nearest Town/Castle according to the position of the hero). ;)

Branikolog:
@ihhub 
Looks nice for me.

@LeHerosInconnu 

>And about Hypnotize, Haunt, Set Elemental Guardians spells and Town Gate (indicate the name of the nearest Town/Castle according to the position of the hero). ;)

Also Stone/Steel skin, Bloodlust and Dragon Slayer need description, that shows the exact number of Primary Skills rised. :)

ihhub:
Hi @LeHerosInconnu and @Branikolog , I added extra description for Town Gate and guardian spells. The rest of spells I highly recommend to do in a separate pull request as they require the base description rework.

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu and @Branikolog , I added extra description for Town Gate and guardian spells. The rest of spells I highly recommend to do in a separate pull request as they require the base description rework.

To avoid the spell failed message, it would be nice to specify that the castle is already occupied by one of the player's heroes, with an empty separating line between sentences:

“The nearest castle is Westfork.

This castle is occupied by your hero Rialdo."

Also it would be nice to specify in the sentence if it is a "town" or a "castle" depending on the case:

“The nearest town is Westfork.

This town is occupied by your hero Rialdo."

In fheroes2:

Nearest town occupied 02

Nearest town occupied 01

For Hypnotize, a sentence in the same vein as for the Resurrect True spell may be adequate:

"This spell controls up to 700 HP."

Or:

"This spell controls 700 HP."

In fheroes2:

Hypnotize spell info 01

Resurrect True spell info 01

ihhub:
Hi @LeHerosInconnu , I think specifying town or castle for **Town** Gate spell might be unnecessary. I suggest to leave the implementation as it is. Some users might ask why not just disable spell then instead of showing extra info?

Regarding Hypnotize spell: this spell has not only HP limit but also duration. Are we going to display both? I think we can do this in a separate pull request if you're okay with this.

ihhub:
I've added Hypnotize spell description.

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu , I think specifying town or castle for **Town** Gate spell might be unnecessary. I suggest to leave the implementation as it is.

The precision of the town or castle is important for decision making because, for example, the player may want to send his hero to a castle where he can recover his spell points thanks to the presence of a Mage Guild or recruit creatures, and this is more likely to be possible in a castle than in a town (so the player can evaluate this without having to exit the Magic Book window and go check).

The Town Portal spell allows to have this kind of information because a list of towns and castles is displayed allowing to right-click on the city graphic to have additional information (creatures in garrison for example).

And adding the graphic of the town or castle in the Town Gate spell information window would not allow to right click on this graphic to have additional information when this window is already opened with the right click (temporary window without "OKAY" button).

> Some users might ask why not just disable spell then instead of showing extra info?

I have already planned to not allow spells to be used directly in the hero's Magic Book, under certain conditions, for adventure spells and combat spells.
For example, for the Town Gate spell, the nearest castle is already occupied by one of the player's other heroes, or the hero trying to cast the spell is already in the nearest castle, or the player no longer has a castle, etc.
This is part of my long list of suggestions not yet posted. :)

But it does tell the player why he can' t do it, and it does so directly in the spell info window.

So, with the current functioning, if the player sees in the spell information window that the castle is occupied for example by the hero to whom he wants to cast the spell, he will not do it and will not waste 10 spell points stupidly (in the current version of fheroes2, if the hero who wants to cast the Town Gate spell is already in the nearest town he can still cast the spell). :)

> Regarding Hypnotize spell: this spell has not only HP limit but also duration. Are we going to display both? I think we can do this in a separate pull request if you're okay with this.

The duration of the Hypnotize spell (duration depending on the hero Spell Power, limited to one attack) should rather be specified in the spell description (to be done in another PR).
And if a modification of the duration or the initial functioning of the spell is altered, it would be better to add one or more sentences to describe the effect of this or these changes.

ihhub:
Hi @LeHerosInconnu , I think all changes for Town Gate spell should be done separately :) @Branikolog even created an issue for this and we can expand the changes there.

Duration of spells will be done separately as I'd like to keep changes minimal.

PR #4768, opened 2021-12-08
Fix regression in campaign bonuses logic

ihhub:
related to #4560


ihhub:
Hi @a1exsh , we missed this part. New bonus type is only for scenario 7.

a1exsh:
> Hi @a1exsh , we missed this part. New bonus type is only for scenario 7.

Dev-oops...

PR #4769, opened 2021-12-10
Fix image Copy function

### LINKED ISSUE #4762: Button rendering artifacts after changing resolution
### Preliminary checks

### Platform

Linux

### Describe the bug

First reported by @ihhub in https://github.com/ihhub/fheroes2/pull/4748#issuecomment-986262150

> if I change resolution of the game OKAY button has incorrect rendering: image

### Save file

No save file needed.

### Additional info

It is not always immediately visible, one might need to change resolution more than once to start seeing it:
Screenshot from 2021-12-08 07-50-14

Also, it seems that all buttons that use the new background-capturing code from #4350 are affected:
Screenshot from 2021-12-08 07-51-14

ihhub:
close #4762

Within `makeButtonWithShadow` function we were calling **Crop** method which by default creates 2-layer image and calls **Copy** function. The last one copies only image layer from Display (as Display is a single-layer image) without touching transform layer. Newly created image has undefined values within transform layer leading to such bizarre output in debug mode.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1105,18 +1105,36 @@ namespace fheroes2
 
         const int32_t offsetOutY = outY * widthOut + outX;
         uint8_t * imageOutY = out.image() + offsetOutY;
-        const uint8_t * imageInYEnd = imageInY + height * widthIn;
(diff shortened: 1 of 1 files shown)


a1exsh:
@ihhub nice corner case! :D

PR #4774, opened 2021-12-10
Fix crash due to out of memory boundary access during resolution change

ihhub:
relates to #4762

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -1287,6 +1287,8 @@ namespace fheroes2
         // deallocate engine resources
         _engine->clear();
 
+        _prevRoi = {};
+
         // allocate engine resources
         if ( !_engine->allocate( width_, height_, isFullScreen ) ) {
             clear();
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz , can you please verify this fix on your side as I can't reproduce the crash?

oleg-derevenetz:
Hi @ihhub Yes, of course, I'll check this later today. It usually doesn't crash but can corrupt some memory and this leads to visual artifacts - like that screen dimming artifact after battle mentioned in #4672.

oleg-derevenetz:
LGTM. I noticed that there is also the `Display::release()` method, should we clear the `_prevRoi` in this method too or this is not necessary?

ihhub:
> LGTM. I noticed that there is also the `Display::release()` method, should we clear the `_prevRoi` in this method too or this is not necessary?

Done. `release` is called at the end of application when we release all resource but it's good to have full cleanup just in case.

PR #4777, opened 2021-12-10
Add hotkey for settings in main menu

a1exsh:
Use the same hotkey as in "New Game" dialog.

Also rename the variable in the code (used to be called `resolutionArea`, as
this previously opened resolutions dialog).

Not sure why `T` is the default hotkey.  I would find `O` more logical, as:
1. It's already used for "Options" within game and battle.
2. It's the next letter in the word "Config" (`C` already stands for "Credits" or "Campaign", but `O` is not yet taken).


ihhub:
@a1exsh , a bunch of thanks for this feature!

PR #4778, opened 2021-12-10
Don't reset the hero's path in SLEEPER mode, reset the SLEEPER mode only with the beginning of the hero's movement

### LINKED ISSUE #4764: Adventure map, do not remove the sleeping hero state when the user left-clicks on the hero portrait in the hero list
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, if the user left-clicks on the hero's portrait in the hero list when this hero is in sleeping state and the hero is not already selected, the sleeping state is removed.
The hero sleeping state should be changed only if the hero moves on the adventure map.

In fheroes2:

https://user-images.githubusercontent.com/43583994/145228256-c0d7773d-2c5c-4c13-8b97-cce9267cbb77.mp4

### Save file

Not needed.

### Additional info

_No response_

oleg-derevenetz:
fix #4764

~~I decided not to do this when moving the hero because currently the existing hero path is reset when the hero goes to the SLEEPER mode, so he can't have a paved path and be in the SLEEPER mode at the same time.~~

### DIFF
--- a/src/fheroes2/ai/ai_base.cpp
+++ b/src/fheroes2/ai/ai_base.cpp
@@ -153,7 +153,7 @@ namespace AI
 
     bool Base::HeroesCanMove( const Heroes & hero )
     {
-        return hero.MayStillMove( false ) && !hero.Modes( Heroes::MOVED );
+        return hero.MayStillMove( false, false ) && !hero.Modes( Heroes::MOVED );
     }
 
     void Base::HeroTurn( Heroes & hero )
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -446,7 +446,7 @@ namespace
         }
 
         hero->ResetModes( Heroes::WAITING | Heroes::MOVED | Heroes::SKIPPED_TURN );
-        if ( !hero->MayStillMove( false ) ) {
+        if ( !hero->MayStillMove( false, false ) ) {
             hero->SetModes( Heroes::MOVED );
         }
         else {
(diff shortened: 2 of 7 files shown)


LeHerosInconnu:
Hello @oleg-derevenetz,

> fix #4764
> 
> I decided not to do this when moving the hero because currently the existing hero path is reset when the hero goes to the SLEEPER mode, so he can't have a paved path and be in the SLEEPER mode at the same time.

This is annoying because if the player just wants to display a predefined path for a sleeping hero to check something, for example to be able to program the position of another hero for a future meeting, but does not want to remove the sleeping state of the hero, he is forced to put this hero back into sleeping mode.
Whereas if the sleeping state of the hero is removed only if he really moves on the adventure map, it is not necessary to put the hero back into sleeping mode.
Which is how it was intended to work in the related issue in the first place. :)

oleg-derevenetz:
> This is annoying because if the player just wants to display a predefined path for a sleeping hero to check something, for example to be able to program the position of another hero for a future meeting, but does not want to remove the sleeping state of the hero, he is forced to put this hero back into sleeping mode. Whereas if the sleeping state of the hero is removed only if he really moves on the adventure map, it is not necessary to put the hero back into sleeping mode. Which is how it was intended to work in the related issue in the first place. :)

Currently putting a hero to the sleeping mode clears his predefined path, so it would be logical to not allow a hero to be in a sleeping mode and have a predefined path at the same time. I suppose that's how it was intended to work originally, otherwise I don't see a reason to clear the predefined path.

LeHerosInconnu:
@oleg-derevenetz.

Yes, it is not necessary to remove the predefined hero path, it is better this way. :)

ihhub:

Issue #4212, opened 2021-09-15
[Vita] Space button


Kiiro-Yakumo:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

The space button as probably everyone know makes the hero re-visit the field he or she stands on as if arriving there without moving away. On PC it's easy to do so with said space button. Vita doesn't have this feature as far as I know but would provide useful in many situations on the map.

### Describe the possible solution

Since most of the possible buttons are taken there is little left to use. In case of PS Vita it could be back touch area, any of it or specific part. In case of PS TV which technically lacks such back touch area it could be L3 and/or R3 on the connected PS4 Gamepad.

### Additional info

Taking lesson from other projects I guess an additional option in the config file to enable/disable back touch area could prove useful for those who have problems with holding PS Vita or for whatever other reason.

vincent-grosbois:
Do we have the "space button" in H2 ? I thought it was a H3 thing

Kiiro-Yakumo:
We do in both, it comes handy especially in situations where you have a big map with one type of portal multiplied and you don't want to constantly make a move back and forth to just pass through the portal. It works in both Heroes 2, Heroes 3 and probably in Heroes 4 as well.

oleg-derevenetz:
@vincent-grosbois in original HoMM2 space button is not really supported, but in fheroes2 it is (as an interface enhancement).

vincent-grosbois:
BTW was it really thought out, what are the gameplay changes of allowing this?
Because there are a few maps where you have a monoliths that leads to lots of different other liths, and it makes the game much easier if you can press space until you find the correct lith exit, without spending move points. Also if it's the same for AI, it can lead to some gameplay changes by the AI.

For instance, there is a map in POL campaign where the enemy has 3 liths on the deserts in his zone, and it's connected to 1 lith in the player's zone. Because the enemy starts in the desert zone, he has only 1 chance out of 3 to reach the player zone when taking a lith, and because the terrain is desert it's quite hard to retry. However with the new change, the AI can reach the player's lith exit on the first try

oleg-derevenetz:
@vincent-grosbois only the liths of the same type (having the same sprite) are interconnected, so if those 3 liths in that map have different sprites, there is no way to teleport to the endpoint of the another lith using space button.

Branikolog:
[...]

PR #4781, opened 2021-12-11
PS Vita updates

Northfear:
Updated `Makefile.vita` to make sure that add-on data files and translations are included in VPK file.
Also removed DPad map scrolling (since it's redundant) and added a gamepad binding for re-visiting current hero location (solves https://github.com/ihhub/fheroes2/issues/4212)

### DIFF
--- a/Makefile.vita
+++ b/Makefile.vita
@@ -30,13 +30,21 @@ all: package
 
 package: $(TARGET).vpk
 
-$(TARGET).vpk: eboot.bin param.sfo
+$(TARGET).vpk: eboot.bin param.sfo translations
 	vita-pack-vpk -s param.sfo -b eboot.bin \
 		--add files/images/platform/psv/sce_sys/icon0.png=sce_sys/icon0.png \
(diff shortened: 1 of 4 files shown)


Northfear:
Sorry for the mess. Forgot to test it with PC build

Makefile.vita
```diff
@@ -30,13 +30,21 @@ all: package
 
 package: $(TARGET).vpk
 
-$(TARGET).vpk: eboot.bin param.sfo
+$(TARGET).vpk: eboot.bin param.sfo translations
```
oleg-derevenetz, line 33: I think we also need to add `translations` to `.PHONY` targets as well, because it's not a file-related target.

oleg-derevenetz (changes requested):
Hi @Northfear I just left one `Makefile`-related suggestion, please take a look when it's convenient for you.

Makefile.vita
Northfear: Done

ihhub:
@Northfear , thank you very much for these changes!

PR #4798, opened 2021-12-17
Update Polish translation

Arthusppp:


ihhub:
@Arthusppp , thank you so much for this change!

PR #4802, opened 2021-12-19
Fix handling of national symbols in the paths to music files on Windows

### LINKED ISSUE #4755: cannot find the music (error Cyrillic directory path)
### Preliminary checks

### Platform

Windows

### Describe the bug

If the path to the game directory contains Cyrillic, then the game cannot find the music

image

### Save file

-

### Additional info

_No response_

oleg-derevenetz:
fix #4755

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -421,15 +421,12 @@ std::string System::FileNameToUTF8( const std::string & str )
     acpToUtf8[str] = str;
 
     auto getLastErrorStr = []() {
-        LPVOID msgBuf;
+        LPTSTR msgBuf;
 
-        const DWORD msgLen = FormatMessage( FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, GetLastError(),
-                                            MAKELANGID( LANG_NEUTRAL, SUBLANG_DEFAULT ), reinterpret_cast<LPTSTR>( &msgBuf ), 0, nullptr );
-
-        if ( msgLen > 0 ) {
-            LPTSTR msgStr = static_cast<LPTSTR>( msgBuf );
-
-            const std::string result( msgStr, msgStr + msgLen );
+        if ( FormatMessage( FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, GetLastError(),
+                            MAKELANGID( LANG_NEUTRAL, SUBLANG_DEFAULT ), reinterpret_cast<LPTSTR>( &msgBuf ), 0, nullptr )
+             > 0 ) {
+            const std::string result( msgBuf );
 
             LocalFree( msgBuf );
 
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -23,11 +23,13 @@
 #include <algorithm>
 #include <cstdlib>
 #include <fstream>
+#include <memory>
 
 #if defined( ANDROID ) || defined( _MSC_VER )
 #include <clocale>
 #endif
 
+#include "logging.h"
 #include "system.h"
 #include "tools.h"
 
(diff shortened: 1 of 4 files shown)


src/engine/system.cpp
```diff
@@ -400,3 +403,80 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
...
+{
+#if defined( __MINGW32__ ) || defined( _MSC_VER )
+    static std::map<std::string, std::string> acpToUtf8;
+
+    if ( str.empty() ) {
+        return str;
+    }
+
+    const auto iter = acpToUtf8.find( str );
+
```
ihhub, line 417: **You can ignore it if you like**: we can remove an extra space from variable definition and it's usage here and on lines 442, 458 and 466.

src/engine/system.cpp
```diff
@@ -400,3 +403,80 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
...
+    if ( WideCharToMultiByte( CP_UTF8, WC_ERR_INVALID_CHARS | WC_NO_BEST_FIT_CHARS, wStr.get(), -1, uStr.get(), uLen, nullptr, nullptr ) != uLen ) {
+        ERROR_LOG( getLastErrorStr() );
+
+        return str;
+    }
+
+    const std::string result( uStr.get() );
+
+    // Put the final result to the cache
+    acpToUtf8[str] = result;
```
ihhub, line 476: As of now we can ignore but I think let's put a TODO comment above this code:
```
// TODO: use a single variable for memory storage without copying it to a temporary string.
```

src/engine/system.cpp
```diff
@@ -400,3 +403,80 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
...
+
+    return result;
```
ihhub, line 478: I'd recommend to do this way to avoid extra memory copy:
```cpp
const inter insertedIter = acpToUtf8.insert( std::string result( uStr.get() ) );
return insertedIter->second;
```

ihhub (changes requested):
Hi @oleg-derevenetz , I left few minor comments here. Would you mind to take a look at them?

src/engine/system.cpp
```diff
@@ -400,3 +403,81 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
     return true;
 }
 #endif
+
+std::string System::FileNameToUTF8( const std::string & str )
+{
+#if defined( __MINGW32__ ) || defined( _MSC_VER )
+    static std::map<std::string, std::string> acpToUtf8;
+
+    if ( str.empty() ) {
```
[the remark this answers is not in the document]
ihhub: It's optional. You can ignore it.

src/engine/system.cpp
oleg-derevenetz: @ihhub I suppose we can't really use `insert()` here because there is already an item for this key in map. There is `insert_or_assign()`, but it requires C++17.

ihhub: How about this?
```cpp
// In case of any issues, the original string will be returned, so let's put it to the cache right away
auto insertedIter = acpToUtf8.insert( str );
...
// some code here
...
insertedIter->second = std::string( result( uStr.get() ) );
return insertedIter->second;
```

oleg-derevenetz: @ihhub Honestly, it looks like there will be no benefit. Consider:

```
S f1(const S &s)
{
    static std::map<int, S> m;

    m[0] = s; // Default constructor for S and then operator=(const S&) will be called

    S s1;

    m[0] = s1; // operator=(const S&) will be called once again

    return s1; // Copy elision
}
```

```
S f2(const S &s)
{
    static std::map<int, S> m;

    auto emplaceRes = m.emplace(0, s); // Copy constructor will be called for S

    emplaceRes.first->second = S(); // operator=(S&&) will be called

    return emplaceRes.first->second; // No copy elision, copy constructor will be called once again
}
```

So we have two copies in the first case (via `operator=()`), and one movement and the same two copies in the second case (via copy constructor). In fact, it looks like we even have [more operations in the second case](https://godbolt.org/z/aqhadbjbn) :)

src/engine/system.cpp
oleg-derevenetz: @ihhub Temporary string encourages copy elision, while other approaches do not (see below).

src/engine/system.cpp
ihhub: Ah, I forgot that we're returning the value, not const reference. Then yes, you're absolutely right.

oleg-derevenetz: @ihhub I already thought about returning a const reference here, but I don't like this because of non-Windows version of this function. Consider some clever guy that likes to live dangerously (image with Austin Powers included) and writes code like this:

```
const std::string & s = FileNameToUTF8( someFunctionReturningString() );
```

In Windows version it will always be OK, because currently we get a reference to temporary with lifetime extension, and in case of `const std::string &` as return type we will get just a reference to the static container element. However, in case of non-Windows and `const std::string &` as return type, we will have a dangling reference.

ihhub:
@oleg-derevenetz , huge thanks for these fixes!

PR #4807, opened 2021-12-19
Windows: Try to set the console codepage to the system codepage

oleg-derevenetz:
Related to #4755

Before:

before

After:

after

Quick and dirty solution.

### DIFF
--- a/src/engine/logging.cpp
+++ b/src/engine/logging.cpp
@@ -18,12 +18,48 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-#include "logging.h"
 #include <ctime>
+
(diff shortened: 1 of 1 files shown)


src/engine/logging.cpp
```diff
@@ -18,12 +18,48 @@
...
+            }
+        }
+
+        ConsoleCPSwitcher & operator=( const ConsoleCPSwitcher & ) = delete;
+
+    private:
+        const UINT _consoleOutputCP;
+    };
+
+    ConsoleCPSwitcher consoleCPSwitcher;
```
ihhub, line 61: Let's please make this variable `const` as we don't modify it.

ihhub (commented):
Hi @oleg-derevenetz , I left just one comment. I can do the change myself :)

ihhub:

PR #4811, opened 2021-12-20
Audio: replace SDL_GetError() by Mix_GetError()

oleg-derevenetz:
It is recommended by manual to use the `Mix_GetError()` to get the last SDL mixer error, not the `SDL_GetError()`.


ihhub:

PR #4812, opened 2021-12-20
Build system improvements

oleg-derevenetz:
* Ensure that apt-get update is running in unattended mode;
* Run brew update before brew install;
* Windows: migrate from SDL2-2.0.12 to SDL2-2.0.18.


ihhub:
@oleg-derevenetz , good job!

PR #4828, opened 2021-12-23
Add popup dialog for map information

ihhub:
Useful for finding the location of the file and checking the map type
image

### DIFF
--- a/src/fheroes2/dialog/dialog_selectscenario.cpp
+++ b/src/fheroes2/dialog/dialog_selectscenario.cpp
@@ -119,7 +119,7 @@ void ScenarioListBox::ActionListDoubleClick( Maps::FileInfo & )
 
 void ScenarioListBox::ActionListPressRight( Maps::FileInfo & info )
 {
-    // On some OS like Windows path contains '\' symbol. This symbol doesn't exist in the resources.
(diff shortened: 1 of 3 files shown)


src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -113,6 +117,35 @@ void ScenarioListBox::ActionListDoubleClick( Maps::FileInfo & )
     selectOk = true;
 }
 
+void ScenarioListBox::ActionListPressRight( Maps::FileInfo & info )
+{
+    // On some OS like Windows path contains '\' symbol. This symbol doesn't exist in the resources.
```
yvesgurcan, line 122: I noticed a small typo in the comment and took the opportunity to improve the comment a little.

```suggestion
    // On some OSes like Windows, the path may contain '\' symbols. This symbol doesn't exist in the resources.
```

yvesgurcan (approved):
Awesome! 🚀

src/fheroes2/dialog/dialog_selectscenario.cpp
ihhub: Ah, thanks for spotting it!

PR #4834, opened 2021-12-24
Fix French translations

yvesgurcan:
- Replace "Petite cartes" with "Petites cartes"
- Replace "multi-joueur" and "multi-joueurs" with "multijoueur"
- Replace "taux de difficulté" with "niveau de difficulté" and replace "note" with "taux"
- Add translation to map type description and title dialog

See discussion: https://github.com/ihhub/fheroes2/discussions/4827


zenseii:
My suggestion would be to add more of the needed changes to the French translation to this PR while we are at it.

Also I'm linking this relevant discussion where this was first mentioned:
https://github.com/ihhub/fheroes2/discussions/4827

yvesgurcan:
> My suggestion would be to add more of the needed changes to the French translation to this PR while we are at it.
> 
> Also I'm linking this relevant discussion where this was first mentioned: #4827

Absolutely. This is just a draft PR right now. Just getting a feel of CI/CD stuff and getting myself set up. More changes soon.

dimag0g:
What was the purpose of "moving things around"? Apparently the PO file was formatted with longer lines, which wrapped some words from one line to the next in multiline entries. This may become a problem during a merge, and when someone with old formatting setup commits their changes.

yvesgurcan:
Hey @dimag0g, this was  a work in progress I set up to get used to the repository. I'm unsure how the wrapping changed so much. I created a different branch and that one didn't worry about longer lines. I closed the PR to avoid further confusions.

ihhub:
Hi @yvesgurcan , please don't close pull request like this :) I'm sure there's a way to update your file via tools to remove these line changes. Also don't worry about the number of commits per pull request. We squash all commits into one upon merging a pull request. I might try on my side to get the correct line lengths.

@dimag0g , any ideas what could be wrong? I think it could be incorrect version of editor...

yvesgurcan:
Sorry, @ihhub! Don't worry, it's not lost work. I saw @dimag0g PR at https://github.com/ihhub/fheroes2/pull/4851/files and I got the project running locally so success. I'm glad to hear commits are squashed, that's a good idea!

dimag0g:
@yvesgurcan I use the following settings in Poedit and never saw the wrapping problems:
Untitled

Though I think this last option (preserve formatting) could cause problems in some cases. I.e. you could edit the PO file manually (implicitly creating a longer line) and then Poedit might decide the formatting it has to "preserve" is 85 characters per line...

yvesgurcan:
Thanks for sharing your configuration of poedit. It seems similar to mine. I most definitely edited the `.po` file manually and then used poedit afterwards, which sounds exactly like what you mentioned. It seems like poedit wanted then to wrap at 80 characters instead of 79.

About "Nécroman," I'm glad that it works given the space there is for faction names. Is there no way for us to make these dialog windows bigger? It's probably not easy with the art used in the background, huh. If I remember correctly, I found a reliable source the other day that mentioned that "Nécroman" is the medieval spelling of the word. The only thing I can find right now is this entry from the Wiktionary: https://fr.wiktionary.org/wiki/n%C3%A9croman. So, technically, this is not wrong, it's just that the translation choice that was made due to size constraints results in using an older spelling of the word, which might be surprising to a French audience. In fact, it seems that "Nécromant" is itself also an older variant and that the modern word to use would be "Nécromancien." But, as you mentioned, this has already been discussed. An annotation system would be ideal! Can we use comments with `#` in `.po` files?

As concerns "Expérimental," that's an understandable issue. It seems that, without the noun that the adjective is qualifying, French tends to default to the masculine singular form, which might explain why people might be tempted to change it to that form. It looks like the French language is not necessarily good at filling gaps and might need to be a little more explicit about the context if you want to use other forms (feminine and/or plural). I suspect that when francophones read "expérimentaux," their mind does not go to "paramètres expérimentaux" and this might create a little bit of confusion while the reader is mentally looking for the missing noun and might not guess what it is. Whereas "expérimental" seems to lead towards a more loosely defined "quelque chose d'expérimental" (something experimental) without focusing on what (and what gender) this "something" that's experimental is. That seems a little easier on the brain and the main point of the text (namely, that you're dealing with something experimental in this menu so, be careful about it) is still conveyed to the reader. I might be lacking the expertise in linguistics to describe this in a better way but hopefully my explanation offers a satisfying reason as to why people might want to use "Expérimental" instead of "Expérimentaux."

Thank you for your apology, @dimag0g. It means a lot to me. I'm genuinely not mad or anything. It's an open source project so collaborating and sharing the burden is even more importatnt. I can't wait to see your PR merged and added to the next build!

PR #4842, opened 2021-12-25
Fallback to `find_library` for smacker when pkgconfig is unable to find it

nE0sIghT:
Smacker doesn't provides pkgconfig file.
Some distros provide it, but others - doesn't.


nE0sIghT:
Looks like this is useless based on #4843

Issue #4797, opened 2021-12-17
Translation, problems encountered with the French language


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

So, here are the problems I noticed.

The letter "é" in "Résolution" and "Expérimental".
The word "Language" should be "Langage".
The word "Experimental" should be "Expérimentaux".
All words certainly not translated.
![French translation 001](https://user-images.githubusercontent.com/43583994/146010111-ac04a9db-f22f-4a53-aef1-0fee9fd6a9b4.png)

Problem with "Random".
**Edit 2021/12/23:** The word "Random" could be replaced by the word "Hasard".
![French translation 008](https://user-images.githubusercontent.com/43583994/146010224-df286a78-735e-4fc1-bc73-ca3a6557e455.png)

The word "Connaissance" is too long.
The word "Connaissance" could be replaced by "Savoir".
And "Pt de magie" could be replaced by "Mana".
![French translation 009](https://user-images.githubusercontent.com/43583994/146010281-8242037a-289d-4b0e-b4dc-90c9d0b6c3db.png)

The word "Plusieurs" is too long.
The word "Plusieurs" could be replaced by "Quelques" a slightly shorter word.
(It would be better to use the corresponding numbers "5-9".) :D
![French translation 018](https://user-images.githubusercontent.com/43583994/146010466-1bc4d4ee-ba8e-4341-bda6-d48d6335b3c5.png)

Apart from that, there are still texts in English, errors with masculine and feminine forms, articles "le", "la" not accorded with the noun, etc. (https://github.com/ihhub/fheroes2/discussions/4827)

### Save file

Not needed.

dimag0g:
I tried to fix most of the things in https://github.com/ihhub/fheroes2/pull/4851, except for the following:

"Savoir" is nice, but "Connaissance" is used in a lot of places, including artefact names. I'd rather not rename these, googling for "ultime livre du savoir" and finding nothing is not cool.

"Quelques" sounds good, but I would like a second opinion on this one before making the changes, as "Plusieurs" is used in the GoG version.

Regarding (#4827):
[...]

PR #4851, opened 2021-12-26
Update French translation

dimag0g:
- translate new strings

- remove incorrect c-string specifiers. I have added `sed -i -e 's/, c-format//' $(POT)` to `src/dist/Makefile`, which should get rid of them in other translations as well

- minor fixes


files/lang/fr.po
```diff
@@ -2094,26 +2294,22 @@ msgid "Quit out of Free Heroes of Might and Magic II."
...
 
-#, fuzzy
 msgid "Resolution"
-msgstr "Choisir la résolution d'écran"
+msgstr "Résolution"
 
-#, fuzzy
 msgid "Experimental"
-msgstr "Esprit"
+msgstr "Experimental"
```
felix642, line 2303: Expérimental

files/lang/fr.po
```diff
@@ -2094,26 +2294,22 @@ msgid "Quit out of Free Heroes of Might and Magic II."
...
 
-#, fuzzy
 msgid "Select Game Language"
-msgstr "Choix de la langue :"
+msgstr "Choix de la langue"
 
-#, fuzzy
 msgid "Change language of the game."
-msgstr "Charger la résolution du jeu."
+msgstr "Charger la langue du jeu."
```
felix642, line 2312: Changez

files/lang/fr.po
```diff
@@ -2124,14 +2320,15 @@ msgstr "Charger la résolution du jeu."
...
 msgid "Attention"
-msgstr "Annexion"
+msgstr "Attention"
 
 msgid ""
 "Your version of Heroes of Might and Magic II does not support any languages "
 "except English."
 msgstr ""
+"Votre version de Heroes of Might and Magic II ne prend en charge aucune "
+"langue autre que l’anglais."
```
felix642, line 2331: That's a bit weird to say but okay ? 😁

files/lang/fr.po
```diff
@@ -2176,30 +2373,28 @@ msgstr ""
...
+msgstr "Vos fonds"
 
 msgid "Planned Gift"
-msgstr ""
+msgstr "Don planifié"
 
-#, fuzzy
 msgid "Gift from %{name}"
-msgstr "Recruter %{name}"
+msgstr "Cadeau de la part de %{name}"
```
felix642, line 2397: On one translation it's "cadeau" and on the other it's "don". I think that both of them should be "cadeau"? What do you think ?

files/lang/fr.po
```diff
@@ -2868,20 +3067,19 @@ msgstr ""
...
-msgstr ""
+msgstr "Dernière version supportée : "
 
 msgid "This saved game is localized to '"
-msgstr ""
+msgstr "Cette partie sauvegardée est localisée en '"
 
 msgid "' language, but the current language of the game is '"
-msgstr ""
+msgstr "' mais la langue actuelle du jeu est '"
```
felix642, line 3082: We lost the first word here. Do you know what the complete sentence is suppose to be ?

files/lang/fr.po
```diff
@@ -2980,9 +3177,11 @@ msgstr "Meilleurs scores"
 msgid "View the high scores screen."
 msgstr "Voir l'écran des meilleurs scores."
 
-#, fuzzy
-msgid "Game settings."
-msgstr "Salutations !"
+msgid "Change language, resolution and settings of the game."
+msgstr "Charger la langue, la résolution et les paramètres du jeu."
```
felix642, line 3181: Changez

files/lang/fr.po
```diff
@@ -3098,9 +3297,8 @@ msgstr ""
 "Vous avez trouvé le %{name}.\n"
 "Votre quête est terminée."
 
-#, fuzzy
 msgid "Ultimate Artifact"
-msgstr "Trouver l'ultime artéfact."
+msgstr "Ultime artéfact"
```
felix642, line 3301: Artéfact Ultime ?

files/lang/fr.po
```diff
@@ -4333,6 +4522,8 @@ msgid ""
 "You've encountered a strange person with a hat and an owl on it. He tells is "
 "you that he is willing to give %{art} if you have %{skill}."
```
felix642, line 4523: There's a mistake in the id ? 
"He tells is you that he is"

files/lang/fr.po
```diff
@@ -4793,6 +4984,9 @@ msgid ""
 "The captain looks at you with surprise and says:\n"
 "\"You already have all the maps I know about. Let me fish in peace now.\""
 msgstr ""
+"Le capitaine vous regarde avec surprise et dit :\n"
+"\"Vous avez déjà toutes les cartes que je connais. Laissez-moi pêcher en "
+"paix maintenant.\""
```
felix642, line 4989: We should use french quotation marks here.

files/lang/fr.po
```diff
@@ -6827,7 +7012,6 @@ msgstr "Double attaque"
 msgid "Double damage to Undead"
 msgstr "Double dégâts aux morts-vivants"
 
-#, c-format
```
felix642, line 6830: Hmm, can you actually remove this without breaking anything ?

files/lang/fr.po
```diff
@@ -8325,6 +8499,32 @@ msgstr "Fermer le Grimoire"
...
 msgstr "Voir %{spell}"
 
+msgid "This spell does %{damage} points of damage."
+msgstr "Ce sort fait %{damage} points de dégâts."
+
+msgid ""
+"This spell summons\n"
+"%{count} %{monster}."
+msgstr ""
+"Ce sort convoque\n"
```
felix642, line 8509: "invoque" ?

felix642 (commented):
Hi @dimag0g, I left a few comments regarding the french translation. Would you mind taking a look ?

files/lang/fr.po
dimag0g: "changer" is used in 8 translations, while "changez" is not used at all. Shall we stick to the existing phrasing?

files/lang/fr.po
dimag0g: "Cette partie sauvegardée est localisée en 'en' mais la langue actuelle du jeu est 'fr'" I could add the word "langue" before the locale name if you like

files/lang/fr.po
dimag0g: cadeau it is...

files/lang/fr.po
dimag0g: It's "L'Ultime Artefact" in the GoG version.

files/lang/fr.po
dimag0g: Good catch!

files/lang/fr.po
dimag0g: I suggest we do this as a separate PR. There are plenty of translations to modify, and the makefile must be adapted too, in order to replace the French quotation marks back to English one, as French marks will not work in the game anyway.

files/lang/fr.po
dimag0g: This comment actually **must** be removed: in all these strings the % character is a literal percent sign, not a printf format specifier. With the comments in place, my translation checker rightfully complains about broken format specifier in translated text.

files/lang/fr.po
felix642: Yes, my bad. I meant "Changer" instead of "Charger". The word is incorrect in this translation.

files/lang/fr.po
felix642: No, now that I know what the full sentence is I think that's fine like this ! :) 
I wanted to make sure that you simply didn't forget the word when translating it.

files/lang/fr.po
felix642: The same applies here, it should be changer instead of changez. My bad..

files/lang/fr.po
felix642: Then "l'ultime artéfact" it is!

files/lang/fr.po
dimag0g: I now see these comments come back after every merge. I'll see if I can deal with them at the source. Worst case, I suggest we run the fheroes2.pot through `sed -i s/, c-format//`. AFAIK we don't use `printf` and friends anywhere in the project.

files/lang/fr.po
felix642: Oh, that's not exactly what I meant.. We have to fix it yes but by doing so you will break this translation. I took a look at the code and the mistake appears to be in every single file. That means that to fix it we will have to change the id in every .po file and in the heroes_action.cpp file.

I feel like this a bit out of the scope of this pull request and it think it would be better to do it in a different pull request what do you think @dimag0g ? 
I can also do it if you're unsure how to fix it.

files/lang/fr.po
felix642: Oh, I was not aware of that. That's perfect then we'll wait for that one!

files/lang/fr.po
felix642: Wow, I didn't realize that! Good catch.

files/lang/fr.po
dimag0g: I added a change to  heroes_action.cpp to this PR. Indeed this will break other translations until they are merged with the new POT template, but I guess that's OK. No translation file is 100% complete in a project where the code is being worked on.

files/lang/fr.po
dimag0g: I just machine-translated it. If you have a better phrasing in mind, let me know.

felix642: No, I simply meant that it's weird to tell a player that his game only supports english in french. If the player sees this message that means that the game definitely supports other languages.
There's nothing wrong with your translation here!

files/lang/fr.po
felix642: > I now see these comments come back after every merge. I'll see if I can deal with them at the source. Worst case, I suggest we run the fheroes2.pot through `sed -i s/, c-format//`. AFAIK we don't use `printf` and friends anywhere in the project.

You probably now how .pot files work better than me. But I can confirm, we are not using the printf family (other than fprintf to display error messages to the console) so I don't think that would be an issue.

felix642 (approved):

files/lang/fr.po
dimag0g: AFAIK this message is displayed when French-specific characters such as `à` are not found in the game resources. French translation will still be there of course, but it will not display text correctly.

felix642:
@dimag0g, one last question.
Can we close #4797 with this PR or not ?

dimag0g:
@felix642 Thanks for pointing me to this bug report. I'll go though it and fix whatever I can, hopefully later today.

files/lang/fr.po
yvesgurcan: Yay! French quotation marks are awesome!

files/lang/fr.po
```diff
@@ -4946,6 +5140,8 @@ msgid ""
 "As you enter the Alchemist's Tower, a hobbled, graying man in a brown cloak "
 "makes his way towards you."
 msgstr ""
+"Lorsque vous entrez dans la Tour de l’Alchimiste, un homme entravé et "
```
yvesgurcan, line 5143: I would suggest "boiteux" (https://en.wiktionary.org/wiki/boiteux) here instead of "entravé."
```suggestion
"Lorsque vous entrez dans la Tour de l’Alchimiste, un homme boiteux et "
```

files/lang/fr.po
```diff
@@ -4970,7 +5165,11 @@ msgid_plural ""
 "After you consent to pay the requested amount of gold, the alchemist grabs "
 "all cursed artifacts and throws them into his magical cauldron."
 msgstr[0] ""
+"Après avoir consenti à payer la quantité d’or demandée, l’alchimiste prends "
+"l’artefact maudit et le jette dans son chaudron magique."
```
yvesgurcan, line 5169: So, to my surprise, it seems that the word for "artifact" in French has two different spellings: "artéfact" and "artefact." Since it looks like the French strings use "artéfact" everywhere else, I would suggest keeping that spelling for the translation of the word "artifact" here and in the diff below.
```suggestion
"l’artéfact maudit et le jette dans son chaudron magique."
```

files/lang/fr.po
```diff
@@ -5110,8 +5309,8 @@ msgid ""
 "Looking somewhat disappointed, the Sphinx sighs. You've answered my riddle "
 "so here's your reward. Now begone."
 msgstr ""
-"Le Sphinx soupire, l'air un peu déçu. \"Vous avez répondu à mon énigme, "
-"voici votre récompense. Maintenant, partez.\" "
+"Le Sphinx soupire, l'air un peu déçu. Vous avez répondu à mon énigme, voici "
```
yvesgurcan, line 5312: Getting rid of these quotes or not is a tough call. They most likely should be present in English but they aren't. So, the question is, do we want to remain faithful to the English text or would we rather be typographically correct in French?

files/lang/fr.po
```diff
@@ -5162,11 +5361,10 @@ msgstr ""
 "cherchez.\""
 
 msgid "Spell book is not present."
-msgstr ""
+msgstr "Le livre de sorts n'est pas présent."
```
yvesgurcan, line 5364: I remember seeing the word "grimoire" to translate "spell book" somewhere else in this file. While I'm tempted to use "livre de sorts" instead of "grimoire," I'm a little on the fence and can't decide which one would work better. They both do the job well. I guess the only thing is that I suspect that it would be better to remain consistent with our translation of "spell book" and pick one of the two options throughout the entire file.

files/lang/fr.po
```diff
@@ -5162,11 +5361,10 @@ msgstr ""
...
 
-#, fuzzy
 msgid "The spell is not found."
-msgstr "Ce sort est déjà utilisé."
+msgstr "Le sort n'est pas trouvé."
```
yvesgurcan, line 5367: It seems that there might be something a little funky going on with the English text. I'm tempted to say that it should be either "Spell not found." or "The spell has not been found." or "The spell wasn't found." but I might be incorrect.
No matter what, "Le sort n'est pas trouvé." seems awkward in French. I would suggest:
```suggestion
msgstr "Le sort n'a pas été trouvé."
```
Or:
```suggestion
msgstr "Sort introuvable."
```

files/lang/fr.po
```diff
@@ -5854,10 +6050,10 @@ msgid "Multi"
 msgstr "Multi"
 
 msgid "race|Random"
-msgstr ""
+msgstr "Hasard"
```
yvesgurcan, line 6053: I've noticed these prefixed strings before and I suspect that it might be necessary to keep the prefix? I'm not entirely sure.
```suggestion
msgstr "race|Hasard"
```

files/lang/fr.po
```diff
@@ -6019,20 +6215,19 @@ msgid "Mine"
 msgstr "Mine"
 
 msgid "Burma shave."
-msgstr ""
+msgstr "Burma shave."
```
yvesgurcan, line 6218: It looks like "shave" was maybe kept untranslated? Is it supposed to be "Burma shaves." in English?
```suggestion
msgstr "Burma se rase."
```

files/lang/fr.po
```diff
@@ -6019,20 +6215,19 @@ msgid "Mine"
...
+msgstr "Panneau suivant : 50 miles."
 
-#, fuzzy
 msgid "See Rock City."
-msgstr "Cité des arbres"
+msgstr "Voir Rock City."
 
 msgid "This space for rent."
-msgstr ""
+msgstr "Cet espace est à louer."
```
yvesgurcan, line 6227: To convey the succinct aspect of the phrase conveyed by the lack of "is" in "This space for rent." (instead of "This space is for rent."), I would suggest to keep the phrase shorter in French as well and drop the verbe "est" as well as "Cet."
```suggestion
msgstr "Espace à louer."
```

files/lang/fr.po
```diff
@@ -2745,108 +2951,101 @@ msgid "Please make sure that campaign files are correct and present."
...
-msgstr ""
+msgstr "Projet original avant 0.7"
 
 msgid "Support us at"
-msgstr ""
+msgstr "Soutenez-nous au"
 
 msgid "Connect with us at"
-msgstr ""
+msgstr "Connectez-vous avec nous sur"
```
yvesgurcan, line 2978: The phrase "Connectez-vous avec nous" doesn't seem to work here since it implies to literally use a device to connect to the internet (I think) whereas the English word "connect" here is about getting in touch with somebody. Or maybe this is an intentional use of the word "connect" to both mean connecting through a device and getting in touch with somebody? :) If that's the case, the use of a single word to mean both things does not seem possible in French, so if we were to take "connect" as "use your machine to go online," it seems that the proper translation would be "Connectez-vous sur."

```suggestion
msgstr "Entrez en contact avec nous sur"
```

files/lang/fr.po
```diff
@@ -6265,42 +6458,35 @@ msgstr "Tourbillon"
...
-msgstr ""
+msgstr "Événement"
 
 msgid "Random Monster"
 msgstr "Monstre Aléatoire"
 
-#, fuzzy
 msgid "Random Ultimate Artifact"
-msgstr "Trouver l'ultime artéfact."
+msgstr "Ultime artéfact aléatoire."
```
yvesgurcan, line 6489: We might want to stick to capitalized first letter for now? And no period.
```suggestion
msgstr "Artéfact Ultime Aléatoire"
```

files/lang/fr.po
```diff
@@ -8292,14 +8463,15 @@ msgid "Sets Water Elementals to guard a mine against enemy armies."
...
-#, fuzzy
 msgid "Petrification"
-msgstr "Fortifications"
+msgstr "Pétrification"
 
 msgid ""
 "Turns the affected creature into stone.  A petrified creature receives half "
 "damage from a direct attack."
 msgstr ""
+"Transforme la créature affectée en pierre.  Une créature pétrifiée reçoit la "
```
yvesgurcan, line 8473: For clarity purposes, I would suggest to move "en pierre" closer to "Transforme" as to avoid confusion about "affectée" having something to do with "en pierre" (even if this is not very likely that such a confusion might happen, really).
```suggestion
"Transforme en pierre la créature affectée. Une créature pétrifiée reçoit la "
```

yvesgurcan (changes requested):
Hooray for a better French translation! I'm a little sad that you stole my thunder as I was working on some of the things you addressed but, at the same time, this work is more robust than mine and I was working on getting set up anyway. Great job! :)

files/lang/fr.po
felix642: It doesn't really matter @yvesgurcan, I simply thought it was funny to say that the game only supports english in french.
If we really consider this to be an issue then we should do that in a different pull request. Since the fix for that would be to remove the translation of this string by changing the file dialog_game_Settings.cpp. But I feel like this is a bit out of the scope of this PR and isn't a major issue. :)

files/lang/fr.po
```diff
@@ -326,7 +324,7 @@ msgid "Luck"
 msgstr "Chance"
 
 msgid "Spell Points"
-msgstr "Pt de magie"
+msgstr "Mana"
```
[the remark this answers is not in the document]
felix642: @yvesgurcan,  Take a look at  #4797,  this is where they proposed this change!

files/lang/fr.po
felix642: This translation appears on signs in the game. I think you should take a look at that: https://en.m.wikipedia.org/wiki/Burma-Shave ;)

files/lang/fr.po
```diff
@@ -121,13 +121,11 @@ msgstr "Peu"
 msgid "All %{race} troops +1"
 msgstr "+1 pour toutes les troupes %{race}"
 
-#, fuzzy
 msgid "Multiple"
-msgstr "Multi"
+msgstr "Multiple"
```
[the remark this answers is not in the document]
dimag0g: This for a player who has several kinds of castles at once. It's translated as "Varié" in GoG version, perhaps we reuse that?
Untitled

files/lang/fr.po
```diff
@@ -121,13 +121,11 @@ msgstr "Peu"
...
 
-#, fuzzy
 msgid "Troops of %{count} alignments -%{penalty}"
-msgstr "Troupes de 3 alignements -1"
+msgstr "Troupes de %{count} alignements -%{penalty}"
```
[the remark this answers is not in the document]
dimag0g: Moral modifiers, e.g. "Troops of 3 alignments -1"

files/lang/fr.po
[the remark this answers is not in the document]
dimag0g: @yvesgurcan It's a hard choice between an ugly shortening and liberal word choice. However, the French GoG version of the game uses "Mana" in a lot of places, which tilted the balance towards this word.

files/lang/fr.po
```diff
@@ -624,43 +621,37 @@ msgid "Human"
...
 msgid "%{race2} %{name2}"
-msgstr "Échanger %{name1} avec %{name2}"
+msgstr "%{race2} le %{name2}"
 
 msgid "Monsters"
 msgstr "Monstres"
 
 msgid "N/A"
-msgstr ""
+msgstr "N/A"
```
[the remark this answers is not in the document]
dimag0g: I didn't think much about this one. What about "N/C" as in "non concerné"?

files/lang/fr.po
```diff
@@ -1313,16 +1303,14 @@ msgstr "Parchemin majeur"
...
 msgid "Power Axe"
 msgstr "Puissante Hache"
 
-#, fuzzy
 msgid "Stealth Shield"
-msgstr "Ultime Bouclier"
+msgstr "Bouclier furtif"
 
 msgid "Thunder Mace"
 msgstr "Massue"
```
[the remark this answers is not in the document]
dimag0g: When it comes to artifact names, I try to follow the GoG translation as close as possible.  "Bouclier furtif" is spelled exactly like that in this version, no idea why they didn't capitalize the second word. And thanks for bringing up the Thunder Mace - it's actually "Massue du Tonnerre".

files/lang/fr.po
```diff
@@ -1365,6 +1353,222 @@ msgstr "Impossible de construire %{name}."
 msgid "Build %{name}."
 msgstr "Construire %{name}."
 
+msgid "Blackridge"
```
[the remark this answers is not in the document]
dimag0g: This was done for languages like Russian, to let the translator spell "Blackridge" with Cyrillic letters. For languages with Latin alphabet no translation is planned AFAIK.

files/lang/fr.po
```diff
@@ -1782,34 +1986,33 @@ msgstr ""
...
 msgid "Click to show previous town."
-msgstr "Montrer le village précédent"
+msgstr "Cliquez pour afficher le village précédent."
 
 msgid "Show previous town"
 msgstr "Montrer le village précédent"
 
 msgid "Army joining"
-msgstr ""
+msgstr "L'armée à joindre"
```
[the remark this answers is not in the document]
dimag0g: Would "L'armée à rallier" be better? This is a dialog title which is displayed when you move the hero to the garrison.

files/lang/fr.po
```diff
@@ -2176,30 +2373,28 @@ msgstr ""
...
-msgstr ""
+msgstr "Cadeau planifié"
 
-#, fuzzy
 msgid "Gift from %{name}"
-msgstr "Recruter %{name}"
+msgstr "Cadeau de la part de %{name}"
 
 msgid "Set Guardian"
 msgstr "Gardien"
```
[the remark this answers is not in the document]
dimag0g: AFAIK "Set Guardian" refers to an army protecting a mine. "Régler" wouldn't sound right here, "affecter", "fixer", "assigner" ou "nommer" would be better.

I don't think it's a good idea to mark manual translations as "fuzzy" - this is what you get when `make merge` tries to translate stuff for you, usually with rather poor results (eg. "Max" -> "Maximus"). It is of much worse quality than a poor manual translation.

files/lang/fr.po
```diff
@@ -2868,20 +3067,19 @@ msgstr ""
...
 msgid "This saved game is localized to '"
-msgstr ""
+msgstr "Cette partie sauvegardée est localisée en '"
 
 msgid "' language, but the current language of the game is '"
-msgstr ""
+msgstr "' mais la langue actuelle du jeu est '"
 
 msgid "Hot Seat"
 msgstr "Tournante"
```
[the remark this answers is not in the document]
dimag0g: Thanks for catching this! A gang rape game is indeed not what HoMM users would expect ;-)
I'll put "Hot Seat" in. Just FYI, the GoG version calls it "jeu à plusieurs" which I personally find confusing, because it essentially means "multiplayer".

files/lang/fr.po
dimag0g: Prefixes should not be kept, exactly because of the character coding of the French version of the game, which maps `|` to `ê`. As a result, the code which removed prefixes during rendering of the translated text was removed, and "race|Hasard" will be printed as "raceêHasard".

The code which removes prefixes from the English text is still there, so "race|Random" is still rendered as "Random"

files/lang/fr.po
dimag0g: How about "Prenez contact avec nous sur"?

files/lang/fr.po
dimag0g: I'd put quotes in English text, where they belong (and in all translations as a consequence). @ihhub What do you think?

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Oh, I see! Yes, "Multiple" seems to work. :)

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Oh, so, the alignment is like the kind of the monster (necromantic, barbarians, etc?).

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: That's fair. Would it be complicated to make the size of these dialog windows bigger either for all languages or depending on the language? It seems probable that this problem can occur in other languages as well. Alternatively, we can drop "Pt de" and just go for "Magie," which should be sufficient enough to imply that this is an amount of something (just like "Mana" here implies that these are "Pt de Mana").

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: It does not seem that "non communiqué" conveys the same meaning than "not applicable." I like the Quebec alternative of "sans object" as it's closer to the idea that something is irrelevant but I suspect that this phrase could surprise francophones from France, especially if we abbreviate it to "s.o." Maybe keeping it closer to the English phrase with "n/a" ("non applicable") is the best solution here.
```suggest
"n/a"
```

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Oh, great, that translation for the Mace is better than mine. As concerns the case of "furtif," well, if you want to keep it lowercase it doesn't hurt much but a capitalized "F" would be more consistent with other translated artifact names. I'm wondering about the GOG translation. How did they do it? Was it done by professional translators?

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Ok! That's good to know.

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Oh, of course! Now that you mentioned that this is a dialog title, this translation makes a lot more sense. It seems that the English phrase means that you are somewhat joining two armies together. In that case, "Armée à joindre" seems to work.

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Oooh, these are great suggestions. "Assigner Guardien" sounds good.

files/lang/fr.po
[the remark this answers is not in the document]
yvesgurcan: Great! Yes, "Jeu à plusieurs" sounds confusing because it does convey the meaning that it is indeed a multiplayer game but it does not convey the meaning that players are sharing a seat to play.

files/lang/fr.po
yvesgurcan: Ooooh, I get it. This is a way to signify that this is a string that applies to a specific case with a specific translation.

files/lang/fr.po
yvesgurcan: lol!
In this case, it looks like both the English and French strings should be "Burma-Shave."

files/lang/fr.po
yvesgurcan: It seems like it would work. 🙂

yvesgurcan:
Thank you for putting up with my comments. This translation is going to look stellar!

files/lang/fr.po
yvesgurcan: Haha! That's fair. :)

dimag0g:
I have pushed the changes implemented so far. I left out the changes related to quotes ("the Sphynx sighs"), I prefer we do it as a separate PR later, perhaps together with transition from English quotes to French ones.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1805,7 +1805,7 @@ void ActionToArtifact( Heroes & hero, s32 dst_index )
                 else {
                     // Did you add a new condition? If yes add a proper if-else branch.
                     assert( 0 );
-                    msg = _( "You've encountered a strange person with a hat and an owl on it. He tells is you that he is willing to give %{art} if you have %{skill}." );
```
ihhub, line 1808: Please make an appropriate change in the rest of po files, otherwise other languages will have non-translated string for this case.

files/lang/fr.po
ihhub: We can put any French-related innocent joke here. Like "Baget est là" (Google translation).

ihhub (changes requested):
Hi @dimag0g , I left few comments here. Would you mind to take a look?

src/fheroes2/heroes/heroes_action.cpp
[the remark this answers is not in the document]
dimag0g: I suggest we leave the translation in for now. Later, when the useless message is removed from the game, it will simply not be used for a while, until the unused strings are cleaned from translation files.

I can merge the new string in the rest of the translation files, but I'd rather not update translations in languages I don't speak. It's preferable that this string gets marked as "fuzzy" by the merge (because the match is not exact), and translators will know they have to check the phrasing.

files/lang/fr.po
```diff
@@ -2373,25 +2566,38 @@ msgstr "Êtes vous sûr de vouloir effacer le fichier :"
...
+msgstr ""
+"\n"
+"\n"
+"Type de carte :\n"
+
+msgid "The Succession Wars"
+msgstr "Les guerres de succession"
+
+msgid "The Price of Loyalty"
+msgstr "Le prix de la loyauté"
```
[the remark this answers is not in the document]
dimag0g: These strings are translated in some places but not others. I'll make sure they are translated everywhere, with caps as in English, unless others object to the idea.

dimag0g:
I'm not familiar with the "Baget est là" joke, could someone enlighten me, or perhaps suggest a better-known one?

dimag0g:
I rebased and force-pushed my old commits, together with two additional ones:
- undo Price of Loyalty / Succession Wars translations. Let's keep them in English just like the French GoG version did.
- merge all translation files with the latest POT. Some strings moved around, but as far as I can tell this reflects changes to the source code.

ihhub:
Hi @dimag0g , are we good merging this pull request?

ihhub (approved):

dimag0g:
@ihhub I have nothing to add, so I think we can merge this!

files/lang/fr.po
yvesgurcan: That seems like a good idea!

files/lang/fr.po
yvesgurcan: Haha! I like the idea of putting a joke that's more relevant to French in the translation. However, I'm not sure I know what "Baget est là" would refer to. 😋 What's the original text?

yvesgurcan (approved):
Amazing work! This is such a great improvement!

ihhub:
@dimag0g , thank you so much for this work!

PR #4853, opened 2021-12-27
Moved trading resources slightly to the left

### LINKED ISSUE #4803: Marketplace and Trading Post windows, the element block with the text "Your Resources" and the resource icons is not positioned correctly
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the element block with the text "Your Resources" and the resource icons is not positioned correctly.
The block of elements indicated by the red rectangle should be moved 2 pixels to the left, to be centered horizontally in relation to the "GIFT" button and to be at the same distance from the nearest edge ot the Marketplace window frame compared to the block of resource icons in the "Available Trades" block.

In fheroes2:

Market your ressources cadre rouge 01

### Save file

Not needed.

### Additional info

_No response_

felix642:
closes #4803 

The x position of the left block was considered to be 0 relative to the frame. So I have instead move the right block 2 pixels to the left with the buttons to center them.

### DIFF
--- a/src/fheroes2/dialog/dialog_marketplace.cpp
+++ b/src/fheroes2/dialog/dialog_marketplace.cpp
@@ -22,6 +22,7 @@
 
 #include <string>
 #include <vector>
+#include <iostream>
 
 #include "agg_image.h"
 #include "cursor.h"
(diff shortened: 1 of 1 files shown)


LeHerosInconnu:
@felix642.

> The x position of the left block was considered to be 0 relative to the frame. So I have instead move the right block 2 pixels to the left with the buttons to center them.

After this change, the "GIFT" button should be moved horizontally 4 pixels to the right to be centered in relation to the "Your Resources" block of resource icons and to be at the same distance from the edge of the nearest frame as the "EXIT" button.

In fheroes2:

Marketplace gift button position 01

LeHerosInconnu:
Hello @felix642,

I checked, all is fine.
Thanks for the change. :)

ihhub:
@felix642 , thank you very much for these changes!

PR #4855, opened 2021-12-28
Enhancement, add feature to click and drag to swap monsters

### LINKED ISSUE #4850: Hero screens, enhancement to "click and hold and move" the mouse cursor
### Preliminary checks

In fheroes2, since it is possible to click and hold and move the mouse cursor to another location to perform an action (e.g. to split a troop into two troops by moving the mouse cursor to a free location), I suggest that when the player performs this action, by moving the mouse cursor to a troop of another type and releasing the mouse button, the two troops are swapped.

In the example, by clicking and holding on the Liches troop and moving the mouse cursor to the Phoenixes troop and releasing the mouse button, the two troops are swapped.

In fheroes2:

Clic and move over other troop 01

Clic and move over other troop 02

Allow to "click and hold and move" the mouse cursor to a troop of another type to swap these two troops.

### Additional info

_No response_

felix642:
closes #4850

### DIFF
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -592,7 +592,8 @@ bool ArmyBar::ActionBarLeftMouseRelease( ArmyTroop & destTroop, ArmyTroop & sele
         RedistributeArmy( selectedTroop, destTroop, _army );
         return true;
     }
-    else if ( destTroop.isValid() ) {
+
+    if ( destTroop.isValid() ) {
         Army::SwapTroops( selectedTroop, destTroop );
         return true;
     }
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -574,6 +574,9 @@ bool ArmyBar::ActionBarLeftMouseRelease( ArmyTroop & troop )
             if ( isSelected() )
                 ResetSelected();
         }
+        else if ( troop.isValid() ) {
+            Army::SwapTroops( *troopPress, troop );
+        }
     }
 
     return true;
@@ -587,6 +590,10 @@ bool ArmyBar::ActionBarLeftMouseRelease( ArmyTroop & destTroop, ArmyTroop & sele
     // cross-army drag split
     if ( selectedTroop.isValid() && ( !destTroop.isValid() || selectedTroop.GetID() == destTroop.GetID() ) ) {
         RedistributeArmy( selectedTroop, destTroop, _army );
+        return true;
+    }
+    else if ( destTroop.isValid() ) {
+        Army::SwapTroops( selectedTroop, destTroop );
         return true;
     }
 
--- a/src/fheroes2/dialog/dialog_marketplace.cpp
+++ b/src/fheroes2/dialog/dialog_marketplace.cpp
@@ -245,11 +245,11 @@ void TradeWindowGUI::ShowTradeArea( const Kingdom & kingdom, int resourceFrom, i
         }
         const TextBox displayMessage( message, Font::BIG, fheroes2::Rect( pos_rt.x, pos_rt.y + 30, pos_rt.width, 100 ) );
         const fheroes2::Sprite & sprite_from = fheroes2::AGG::GetICN( ICN::RESOURCE, Resource::GetIndexSprite2( resourceFrom ) );
-        dst_pt.x = pos_rt.x + ( pos_rt.width - sprite_from.width() + 1 ) / 2 - 70;
+        dst_pt.x = pos_rt.x + ( pos_rt.width - sprite_from.width() ) / 2 - 70;
         dst_pt.y = pos_rt.y + 115 - sprite_from.height();
         fheroes2::Blit( sprite_from, display, dst_pt.x, dst_pt.y );
         const fheroes2::Sprite & sprite_to = fheroes2::AGG::GetICN( ICN::RESOURCE, Resource::GetIndexSprite2( resourceTo ) );
-        dst_pt.x = pos_rt.x + ( pos_rt.width - sprite_to.width() + 1 ) / 2 + 70;
+        dst_pt.x = pos_rt.x + ( pos_rt.width - sprite_to.width() ) / 2 + 70;
         dst_pt.y = pos_rt.y + 115 - sprite_to.height();
         fheroes2::Blit( sprite_to, display, dst_pt.x, dst_pt.y );
         const fheroes2::Sprite & sprite_fromto = fheroes2::AGG::GetICN( tradpost, 0 );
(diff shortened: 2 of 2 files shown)


ihhub:
Hi @Branikolog , could you please test this issue since you know all possible cases for it?

Branikolog:
Hi, @ihhub, @felix642 and @LeHerosInconnu.
Operating in a manner, @LeHerosInconnu proposed in #4850, became not so consistent now regarding to other army reorganizing logic. Definitely, Drag-and-drop is quite comfortable in way it was implemented in current PR. But while we "drop" monster into an empty slot game initiates fast split dialog, which offers completely other experience, comparing to implemented here. 
I suggest to advance the logic to the case we drag-and-drop target monster into empty slot.  So we can reorganize army much faster. Also we should initiate fast split dialog for the cases, it was implemented before, but only with "Shift" button being pressed. In this way it will be operating in the same, traditional manner (I mean holding Shift to split stacks), all players got used to. Also, while drag-and-drop on the same type we should unite both stacks. So, in other words, I expect drag-and-drop working in a completely same manner, single left click works.

Returning to PR.
I've found a few issues. @felix642, could you, please, look?
If we select one stack and then try to move it over another - the army slot still keep being selected:
image
image
I think we should cancel selection, while using drag-and-drop for any cases.

Also, when one stack is being selected in garrison or hero army and we try to drag-and-drop creatures in other army - nothing happens. 
image
image

felix642:
> Hi, @ihhub, @felix642 and @LeHerosInconnu. Operating in a manner, @LeHerosInconnu proposed in #4850, became not so consistent now regarding to other army reorganizing logic. Definitely, Drag-and-drop is quite comfortable in way it was implemented in current PR. But while we "drop" monster into an empty slot game initiates fast split dialog, which offers completely other experience, comparing to implemented here. I suggest to advance the logic to the case we drag-and-drop target monster into empty slot. So we can reorganize army much faster. Also we should initiate fast split dialog for the cases, it was implemented before, but only with "Shift" button being pressed. In this way it will be operating in the same, traditional manner (I mean holding Shift to split stacks), all players got used to. Also, while drag-and-drop on the same type we should unite both stacks. So, in other words, I expect drag-and-drop working in a completely same manner, single left click works.
> 
> Returning to PR. I've found a few issues. @felix642, could you, please, look? If we select one stack and then try to move it over another - the army slot still keeps being selected: image image I think we should cancel selection, while using drag-and-drop for any cases.
> 
> Also, when one stack is being selected in garrison or hero army and we try to drag-and-drop creatures in other army - nothing happens. image image While it still works for cross army dragging.

I completely agree with what @Branikolog suggested. We should be able to move units around using drag-and-drop and to split them using shift+drag-and-drop. This way we will stay consistent with what we have and still allow players to easily split their army if needed.

I will wait for @ihhub and @LeHerosInconnus opinions before implementing this change.

LeHerosInconnu:
Hello everyone,

> > Hi, @ihhub, @felix642 and @LeHerosInconnu. Operating in a manner, @LeHerosInconnu proposed in #4850, became not so consistent now regarding to other army reorganizing logic. Definitely, Drag-and-drop is quite comfortable in way it was implemented in current PR. But while we "drop" monster into an empty slot game initiates fast split dialog, which offers completely other experience, comparing to implemented here. I suggest to advance the logic to the case we drag-and-drop target monster into empty slot. So we can reorganize army much faster. Also we should initiate fast split dialog for the cases, it was implemented before, but only with "Shift" button being pressed. In this way it will be operating in the same, traditional manner (I mean holding Shift to split stacks), all players got used to. Also, while drag-and-drop on the same type we should unite both stacks. So, in other words, I expect drag-and-drop working in a completely same manner, single left click works.
>
> I completely agree with what @Branikolog suggested. We should be able to move units around using drag-and-drop and to split them using shift+drag-and-drop. This way we will stay consistent with what we have and still allow players to easily split their army if needed.
> 
> I will wait for @ihhub and @LeHerosInconnus opinions before implementing this change.

Since it is also possible to proceed to the separation and reorganization of the troops by selecting a troop and using the right mouse button (so not only by using the left click or "drag and drop" with the "shift" key), I think it could be possible to change this although I am used to the current operation (the use of the right mouse button must be maintained).
Otherwise having to go back to using only the "shift" key for the separation operation would be a step backward for me (any operation should be able to be done only with the use of the mouse).
I think it would be better to test this in normal use (during a game) to see if it is not too difficult to adapt to this change.
It should also be taken into account that it has been working this way for a long time now and that the players have integrated this into their way of using the interface.

Branikolog:
Hi, @LeHerosInconnu.

>Otherwise having to go back to using only the "shift" key for the separation operation would be a step backward for me (any operation should be able to be done only with the use of the mouse).

Same for me. But game logic should be consistent and predictable for players. That's why I suggested the above. Anyway, for fast splitting with drag-and-drop we can implement later the same operation for right click as well.

LeHerosInconnu:
@Branikolog.

> Hi, @LeHerosInconnu.
> 
> > Otherwise having to go back to using only the "shift" key for the separation operation would be a step backward for me (any operation should be able to be done only with the use of the mouse).
> 
> Same for me. But game logic should be consistent and predictable for players. That's why I suggested the above. Anyway, for fast splitting with drag-and-drop we can implement later the same operation for right click as well.

The right click already works for the use of fast separation into slots, it should not be removed.

Branikolog:
@LeHerosInconnu 

> The right click already works for the use of fast separation into slots, it should not be removed.

I mean allow to hold right mouse button, move cursor to an empty slot to initiate fast splitting dialog, the same, as it works with single clicking on a troop proceeded with right click on empty slot.

LeHerosInconnu:
@Branikolog.

> @LeHerosInconnu
> 
> > The right click already works for the use of fast separation into slots, it should not be removed.
> 
> I mean allow to hold right mouse button, move cursor to an empty slot to initiate fast splitting dialog, the same, as it works with single clicking on a troop proceeded with right click on empty slot.

But right-clicking and holding on a troop displays the creature's information window as long as the right click is held. :)

Branikolog:
> @Branikolog.
> 
> > @LeHerosInconnu
> > > The right click already works for the use of fast separation into slots, it should not be removed.
> > 
> > 
> > I mean allow to hold right mouse button, move cursor to an empty slot to initiate fast splitting dialog, the same, as it works with single clicking on a troop proceeded with right click on empty slot.
> 
> But right-clicking and holding on a troop displays the creature's information window as long as the right click is held. :)

Completely forgot about that "rarely" used feature. :D
Need to open the second can of beer to think about that thoughtfully.

felix642:
I will close this pull request for now since we're not 100% sure to know how we want to implement this. If anyone is interested you can take a look at #4855 where @Branikolog and @LeHerosInconnu have started to discuss a possible implementation of this new logic. When we all agree on the new implementation I will open a new pull request with the requested changes.

PR #4866, opened 2021-12-30
Changed captain's name

### LINKED ISSUE #4859: Combat screen, Captain's Options window, change the title
### Preliminary checks

In fheroes2, the name of the city is indicated as if it was the name of the Captain.

In the original game, the title of the window is simply "Captain".

In original game:

Captain title combat H2 01

In fheroes2:

Captain info buttons 001

I suggest to change the title of the Captain's Options windows to "Captain of the 'name of the city'".

In the example:

"Captain of Fenton"

Instead of:

"Fenton the Barbarian"

### Additional info

_No response_

felix642:
closes #4859

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -738,8 +738,9 @@ int Battle::Arena::DialogBattleHero( const HeroBase & hero, const bool buttons, 
     std::string str;
     Text text;
     text.Set( Font::SMALL );
-    str = _( "Captain of %{name}" );
+    str = hero.isCaptain() ? _( "Captain of %{name}" ) : _( "%{name} the %{race}" );
     StringReplace( str, "%{name}", hero.GetName() );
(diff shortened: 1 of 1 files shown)


yvesgurcan:
Why?

yvesgurcan:
Wouldn't that make all similar dialog titles, even for heroes, "Captain of something?" Or does this only apply to captains?

felix642:
> Wouldn't that make all similar dialog titles, even for heroes, "Captain of something?" Or does this only apply to captains?

Good catch, for some reasons I didn't think of that ?

yvesgurcan:
Oh, ok! Would it be possible to use some logic to only apply the change in the string to situations where the player is looking at the info related to a captain, not a hero?

yvesgurcan:
Awesome! I'm glad we were able to catch this. 🙂

yvesgurcan (approved):

ihhub:
@felix642 , good job! Thanks a lot for this!

PR #4871, opened 2022-01-01
Improve AI behavior during castle defence

ihhub:
- fix calculation of tower strength for neutral castle
- don't set AI in defensive strategy if no towers are present

### DIFF
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -31,6 +31,7 @@
 #include "battle_army.h"
 #include "battle_cell.h"
 #include "battle_interface.h"
+#include "battle_tower.h"
 #include "battle_troop.h"
 #include "game_static.h"
 #include "logging.h"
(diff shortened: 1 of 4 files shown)


src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1142,7 +1143,11 @@ s32 Battle::Unit::GetScoreQuality( const Unit & defender ) const
     }
     // Otherwise heavy penalty for hiting our own units
     else if ( attacker.GetArmyColor() == defender.GetArmyColor() ) {
-        attackerThreat *= -2;
+        const bool isTower = ( dynamic_cast<const Battle::Tower *>( this ) != nullptr );
+        if ( !isTower ) {
```
oleg-derevenetz, line 1147: Hi @ihhub why there is no penalty for towers for hitting own units? :)

ihhub: I don't remember that tower ever hit your own units. The problem here is happens only when a neutral army is defending a castle so Tower (which is a Unit) color which is always NONE matches defending army color which isn't the case for normal AI.

oleg-derevenetz: @ihhub So in this case it so happens that `attacker.GetArmyColor()` is `NONE` (because it's the tower) and `defender.GetArmyColor()` is also `NONE` if `defender` is from the castle defenders (after all, neutral armies cannot besiege the castle), so the penalty is applied. I still doesn't see the issue :(

oleg-derevenetz (approved):

ihhub:
@oleg-derevenetz , there's another issue which I'll address separately: we calculate the power of towers based on the current unit meaning that we use our own army as a target which is wrong.

src/fheroes2/battle/battle_troop.cpp
oleg-derevenetz: Ah, I got this. It's due to the peculiarities of the AI strength evaluation, and applies only to towers.

oleg-derevenetz:
@ihhub Yeah, AI features is still a somewhat unexplored area for me :)

PR #4878, opened 2022-01-01
Minor Route::Path cleanup

oleg-derevenetz:
Follow-up after #4846

PR #4846, opened 2021-12-25
Rework the logic of re-calculating the paths of heroes on the adventure map

oleg-derevenetz:
fix #4838

Alternative approach to the #4839 (more risky in terms of possible breaking changes). Fixes also the following cases:

---

Before this change:

Obstacle of Heroes Before 1
Obstacle of Heroes Before 2

After this change:

Obstacle of Heroes After 1
Obstacle of Heroes After 2

---

Before this change:

Obstacle of Monsters Before 1
Obstacle of Monsters Before 2

After this change:

Obstacle of Monsters After 1
Obstacle of Monsters After 2

PR #4881, opened 2022-01-01
Route::Path::GetFrontDirection(): return Direction::UNKNOWN if path is empty and path destination is not valid

oleg-derevenetz:
Another follow-up after #4846 and #4878


ihhub:
@oleg-derevenetz , thanks again for this code cleanup!

PR #4886, opened 2022-01-02
Update documentation related to MacOS

githappens:
As GitHub does not provide macOS runners on the ARM platform, zip archives are not created for Apple Silicon chips. This patch clarifies that x86-64 binaries on MacOS will not run on ARM64 platform and points affected users to compatible installation methods.


ihhub:
@githappens , thank you so much for your contribution to the project!

PR #4890, opened 2022-01-03
Fix sed -i behavior on some systems

oleg-derevenetz:
GNU `sed` is incompatible with BSD `sed` in this regard: the GNU `sed` has the `-i[SUFFIX]` syntax, while the BSD `sed` has the `-i extension` syntax, so we have undesirable behavior at least on the latest macOS (files like `fheroes2.pot-e` for example).

### DIFF
--- a/files/lang/Makefile
+++ b/files/lang/Makefile
@@ -42,25 +42,25 @@ ru.mo: ru.po
 # German has locale-specific transliteration rules (ä -> ae, etc)
 de.mo: de.po
 	sed -e 's/ä/ae/g' $< > $<.tmp
-	sed -i -e 's/ö/oe/g' $<.tmp
-	sed -i -e 's/ü/ue/g' $<.tmp
-	sed -i -e 's/Ö/Oe/g' $<.tmp
-	sed -i -e 's/Ü/Ue/g' $<.tmp
+	sed -i~ -e 's/ö/oe/g' $<.tmp
+	sed -i~ -e 's/ü/ue/g' $<.tmp
+	sed -i~ -e 's/Ö/Oe/g' $<.tmp
+	sed -i~ -e 's/Ü/Ue/g' $<.tmp
 	iconv -f utf-8 -t ascii//TRANSLIT $<.tmp > $<.2.tmp
 	msgfmt $<.2.tmp -o $@
-	
+
 # Norwegian has locale-specific transliteration rules (æ -> ae, etc)
 nb.mo: nb.po
 	sed -e 's/æ/ae/g' $< > $<.tmp
-	sed -i -e 's/ø/oe/g' $<.tmp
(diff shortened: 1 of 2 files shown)


files/lang/Makefile
```diff
@@ -42,25 +42,25 @@ ru.mo: ru.po
...
 de.mo: de.po
 	sed -e 's/ä/ae/g' $< > $<.tmp
-	sed -i -e 's/ö/oe/g' $<.tmp
-	sed -i -e 's/ü/ue/g' $<.tmp
-	sed -i -e 's/Ö/Oe/g' $<.tmp
-	sed -i -e 's/Ü/Ue/g' $<.tmp
+	sed -i~ -e 's/ö/oe/g' $<.tmp
+	sed -i~ -e 's/ü/ue/g' $<.tmp
+	sed -i~ -e 's/Ö/Oe/g' $<.tmp
+	sed -i~ -e 's/Ü/Ue/g' $<.tmp
```
a1exsh, line 48: I wonder if the transliteration is still relevant for the German locale, given the recent improvements in fonts?

oleg-derevenetz: Honestly, I don't know :)

ihhub: Hi @dimag0g , can you please help us with this? What do you think about this change?

dimag0g: Is the longer `--in-place` option recognized correctly on MacOS? I think that would improve readability: `-i` is already a bit criptic, `-i~` made me google the man page.

@a1exsh the `-i` option for `sed` is not exclusive to German, so even if German ends up not using it, the question is still valid.

oleg-derevenetz: @dimag0g macOS use BSD `sed`, it doesn't have longer options (unlike GNU `sed`). This page is from FreeBSD, but on macOS it's just the same:

https://nixdoc.net/man-pages/FreeBSD/man1/sed.1.html

Previous macOS versions have `sed` that works correctly with parameters like `sed -i -e ...`, in this case extension was recognized as zero-length. However, `sed` from latest macOS does not recognize this case as zero-length anymore, it thinks that extension is `-e`. There is a variant to use `sed -i "" -e ...`, but it is incompatible with GNU `sed`, because GNU `sed` recognizes `""` as a filename:

```bash
oleg@gentoo ~/Temp $ sed -i "" -e "s/.//" tt.sh
sed: can't read : No such file or directory
oleg@gentoo ~/Temp $ sed -i -e "s/.//" tt.sh
oleg@gentoo ~/Temp $
```

dimag0g: Frankly, none of the options we spoke about so far seem to have clear advantages over `-i~`, so I suggest we just keep your changes.

ihhub:
@oleg-derevenetz , thank you so much for this new addition to the code!

Issue #4868, opened 2021-12-31
Missing letters in Russian translation on Nintendo Switch


dimag0g:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Nintendo Switch

### Describe the bug

This happens with Buka data files. The PC version works correctly, the issue seems to be Switch-specific. In some of the Russian translations, letters "т", "в", "й" and some others are rendered as spaces.

![2021123118373400](https://user-images.githubusercontent.com/10945319/147834430-603d8135-73b7-416c-8af4-bc5bee35f01a.jpg)
![2021123118374700](https://user-images.githubusercontent.com/10945319/147834431-09387f76-3dd7-448c-ba1d-1628841d061c.jpg)

This is not systematic, i.e. some of those letters are missing in one sentence, but may be present in the next one.

### Save file

The issue can be seen on a build screen of any castle, so no saved game is necessary.

dimag0g:
My investigation so far shows that `StringLower` function is to blame, which uses the current locale to decide which characters are upper/lower case. Apparently, on the PC the locale is set to "C" which means `tolower` only ever modifies ASCII characters, leaving other code pages intact. This is already a bug, as Russian letters remain uppercase after a call to `StringLower`, and I believe the same problem affects Polish and German languages (which use CP1250/CP1252) as well.

What's worse, `std::string` contains elements of type `char`, while `tolower` only works with `unsigned char`. This isn't a problem on Windows/Linux/MacOS, but it appears to be a problem on Switch:

> int tolower( int ch );

> ch - character to be converted. If the value of ch is not representable as unsigned char and does not equal EOF, the behavior is undefined

Casting the contents of a string to unsigned chars fixes the issue on the Switch:
```
/* convert to lower case */
std::string StringLower( std::string str )
{
    std::transform( str.begin(), str.end(), str.begin(), []( unsigned char c ){ return std::tolower( c ); } );
    return str;
}
```

I'll make a PR with just this change, unless we want to make `StringLower` locale-aware right away.

ihhub:
Closing this issue as it was fixed in #4891 .

PR #4891, opened 2022-01-03
added missing casts to unsigned char for isspace, isdigit, and tolower

dimag0g:
The reasons for this change are described in https://github.com/ihhub/fheroes2/issues/4868


ihhub:
@dimag0g , thanks a lot for this fix!

PR #4894, opened 2022-01-03
AIWorldPathfinder: encourage the AI-controlled heroes to overcome water obstacles using boats

### LINKED ISSUE #4464: Efficient boat and coast usage by AI
### Preliminary checks

Current AI implementation discourages AI to use boats and come back to coast tiles already visited. Moreover, AI treats boats and coast tiles as action object which at some point makes sense but from another they aren't actions at all.

 We have pathfinder implementation which depends on pathfinding skill of a hero, his start position and color. We need to remember maximum move points available (`GetMaxMovePoints()`) and the current move points (`GetMovePoints`). During scanning tiles for path we can add 2 cases:
- if a hero is on land and there's an empty boat - reduce as many move point as needed to end the current day
- if a hero is on water and there's a coast tile - reduce as many move point as needed to end the current day

Also remove AI behavior regarding boat and coast, it needs to ignore them as action tiles inside `ai_normal_hero.cpp` file. Such approach will give AI much more freedom to plan its expansion.

This logic must be applied only for AI heroes.

### Additional info

_No response_

oleg-derevenetz:
close #4464

Hi @ihhub here is the test map file: [WTRTEST.ZIP](https://github.com/ihhub/fheroes2/files/7803018/WTRTEST.ZIP). I didn't touch AI part however, but it seems that AI is trying to sail to neighboring islands and even to the island with my castle. I need your advice in regard to further AI-related changes.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -80,7 +80,7 @@ namespace
         case MP2::OBJ_WHIRLPOOL:
             return hero.isShipMaster() && !hero.isVisited( tile );
         case MP2::OBJ_COAST:
-            // Coast is not an action object. If this assertion blows up then something wrong with logic above.
+            // Coast is not an action object. If this assertion blows up then something wrong with the logic above.
             assert( 0 );
             return false;
 
@@ -671,7 +671,7 @@ namespace AI
             return 5000;
         }
         else if ( objectType == MP2::OBJ_COAST ) {
-            // Coast is not an object. If this assertion blows up something is wrong the the logic.
+            // Coast is not an object. If this assertion blows up something is wrong with the logic.
             assert( 0 );
             return -dangerousTaskPenalty;
         }
@@ -684,7 +684,7 @@ namespace AI
             return -dangerousTaskPenalty; // no point to even loose the army for this
         }
         else if ( objectType == MP2::OBJ_BOAT ) {
-            // Boat is not considered by AI as an action object. If this assertion blows up something is wrong the the logic.
+            // Boat is not considered by AI as an action object. If this assertion blows up something is wrong with the logic.
             assert( 0 );
             return -dangerousTaskPenalty;
         }
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -55,9 +55,7 @@ namespace
         const Army & army = hero.GetArmy();
         const Kingdom & kingdom = hero.GetKingdom();
 
-        // AI heroes must not take into account coast tile as an action tile. They should calculate paths through them.
-        // TODO: remove the condition for coast once AI is capable to create a path between land and water.
-        if ( !MP2::isActionObject( objectType ) && objectType != MP2::OBJ_COAST ) {
+        if ( !MP2::isActionObject( objectType ) ) {
             return false;
         }
 
@@ -82,7 +80,9 @@ namespace
         case MP2::OBJ_WHIRLPOOL:
             return hero.isShipMaster() && !hero.isVisited( tile );
         case MP2::OBJ_COAST:
-            return hero.isShipMaster() && !hero.isVisited( tile ) && tile.GetRegion() != hero.lastGroundRegion();
+            // Coast is not an action object. If this assertion blows up then something wrong with logic above.
+            assert( 0 );
+            return false;
 
         case MP2::OBJ_SAWMILL:
         case MP2::OBJ_MINES:
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1066,9 +1066,7 @@ namespace AI
 
         int priorityTarget = -1;
         maxPriority = lowestPossibleValue;
-#ifdef WITH_DEBUG
         MP2::MapObjectType objectType = MP2::OBJ_ZERO;
-#endif
 
         // pre-cache the pathfinder
         _pathfinder.reEvaluateIfNeeded( hero );
@@ -1123,9 +1121,7 @@ namespace AI
                 if ( dist && value > maxPriority ) {
                     maxPriority = value;
                     priorityTarget = node.first;
-#ifdef WITH_DEBUG
                     objectType = static_cast<MP2::MapObjectType>( node.second );
-#endif
 
                     DEBUG_LOG( DBG_AI, DBG_TRACE,
                                hero.GetName() << ": valid object at " << node.first << " value is " << value << " ("
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -395,6 +395,9 @@ namespace
             return AIShouldVisitCastle( hero, index );
 
         case MP2::OBJ_BOAT:
+            // AI should never consider a boat as a destination point. It uses them only to make a path.
+            return false;
+
         case MP2::OBJ_STONELITHS:
             // check later
             return true;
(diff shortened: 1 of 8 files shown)


ihhub:
Hi @oleg-derevenetz , in **ai_normal_hero.cpp** file at lines 60, 84, 671 and 913 we use `MP2::OBJ_COAST` as an object. We should remove it. Once it's done AI without your changes will never disembark the boat from sea. To get save samples you can take a look at #4649.

ihhub:
Also lines 690-693 should be removed as well:
```cpp
        else if ( objectType == MP2::OBJ_BOAT ) {
            // de-prioritize the water movement even harder
            return -5000.0;
        }
```

oleg-derevenetz:
Hi @ihhub Honestly, I don't like the result so far. I will push my debug variant now, it will fail the build due to "unused variable" warnings, but you can try it however with this savefile: [AI_Debug.zip](https://github.com/ihhub/fheroes2/files/7811433/AI_Debug.zip).

First of all, you can see how the new AI pathfinding (pre-alpha) works. This is an example:

Pathfinding in AI mode

This works only for Maximus in map debug mode (I use `debug = 49824`). You should use the right mouse button to make the new AI pathfinding work in the "planning" mode. The left mouse button works as usual.

Also, you can see the behavior of AI-controlled heroes (you will get the "real" vision on the whole map when you use the right mouse button for the first time). I'd say it's still very dumb. For example, they like to get into boats and disembark from them unnecessarily.

oleg-derevenetz:
@ihhub I have an impression that we should give the AI kingdoms "full vision" (no fog at all).

oleg-derevenetz:
@ihhub I enabled the full vision mode for AI (you will need to press the right mouse button with Maximus selected to enable it). I'd say that with full vision AI not so dumb :)

ihhub:
Hi @oleg-derevenetz , you have to remove the code in **ai_normal_hero.cpp** related to COAST, otherwise AI'll be doing some stupid things like embarking and disembarking all the time :( I'll take a look at the code at my evening.

oleg-derevenetz:
@ihhub AI behaves very purposeful even without further changes if it sees the whole map, no time wasted. It captures the neutral town and my town very quickly if the full AI vision is enabled (by pressing right mouse button).

oleg-derevenetz:
Complex AI path

oleg-derevenetz:
@ihhub

> Hi @oleg-derevenetz , you have to remove the code in **ai_normal_hero.cpp** related to COAST, otherwise AI'll be doing some stupid things like embarking and disembarking all the time :( I'll take a look at the code at my evening.

Hmm, the mighty hero with 5 Titans continues to embark & disembark endlessly on the `!WTRTST3` map even when I removed the condition regarding `OBJ_COAST`.

ihhub:
Hi @oleg-derevenetz , I found the root of the problem. I'm finishing my testings.

oleg-derevenetz:
@ihhub you are free to add your commits to this PR.

oleg-derevenetz:
@ihhub Is there anything else that would be desirable to do on the AI part? I suppose that AI shouldn't consider `OBJ_BOAT` and `OBJ_COAST` as objects, because AI pathfinder should be able to pave the path through all this automatically (at least in the "planning mode").

ihhub:
@oleg-derevenetz , I'm done with my changes.

oleg-derevenetz:
@ihhub BTW does the AI know how to build new boats or use the Summon Boat spell?

ihhub:
> @ihhub BTW does the AI know how to build new boats or use the Summon Boat spell?

We have the code related to building boats:
```cpp
        // Call internally checks if it's valid (space/resources) to buy one
        if ( castle.GetKingdom().GetFunds() >= PaymentConditions::BuyBoat() * ( islandOrPeninsula ? 2 : 4 ) )
            castle.BuyBoat();
```
However, AI doesn't know how to use Summon Boat spell as of now.

Branikolog:
I definitely saw AI building boats at shipyard.

oleg-derevenetz:
I suppose we should now test this with wider audience on a real big maps with water.

ihhub:
> I suppose we should now test this with wider audience on a real big maps with water.

Please rebase from master to have Whirlpool fix in.

idshibanov (approved):
Great work, all changes are quite straightforward.

We could disable map fog for AI on higher difficulties, it will definitely improve the early game. Might have a small performance hit during AI turns though.

oleg-derevenetz:
It seems that AI still loves Whirlpools very much:

https://user-images.githubusercontent.com/32623900/148379068-09a7f243-81f0-4265-a923-4d1fe28ef9f3.mp4

[AI-Whirlpool.zip](https://github.com/ihhub/fheroes2/files/7821516/AI-Whirlpool.zip)

ihhub:
> It seems that AI still loves Whirlpools very much:
> Free.Heroes.of.Might.and.Magic.II.version_.0.9.11.2022-01-06.14-37-05.mp4
> 
> [AI-Whirlpool.zip](https://github.com/ihhub/fheroes2/files/7821516/AI-Whirlpool.zip)

Not again... I'll check it.

oleg-derevenetz:
@ihhub

> Not again... I'll check it.

As I can see, the `AIWorldPathfinder` is able to pave the path through Stone Liths, but not through Whirlpools. We can implement the similar logic for Whirlpools too (the only issue is that Whirlpool exit tile is not deterministic, so the "real" path may differ from the planned, and AI should be ready for this). I can try to implement necessary changes for `AIWorldPathfinder`, and then there will be need to implement logic on the AI part (similar to the logic that AI uses for teleporters). What do you think?

idshibanov:
You could combine coast and whirlpool objects together to use them only if there's no other action possible. I thought to tag them as 'visited' before but it's kind of tricky because they consist of multiple tiles.

oleg-derevenetz:
> Correct. You could use the same approach but we have to discourage AI from using it anyway...

That's the problem. If pathfinder will use it automatically, how can we discourage the AI to use it? The pathfinder always offers the shortest path, and if the shortest path from A to B is via the whirlpool, then the pathfinder will always offer the path through the whirlpool.

ihhub:
> You could combine coast and whirlpool objects together to use them only if there's no other action possible. I thought to tag them as 'visited' before but it's kind of tricky because they consist of multiple tiles.
> 
> > As I can see, the AIWorldPathfinder is able to pave the path through Stone Liths, but not through Whirlpools.
> 
> Correct. You could use the same approach but we have to discourage AI from using it anyway...

Every tile of a single Whirlpool has a UID. We need to mark all of them as visited if needed.

ihhub:
Hi @oleg-derevenetz  and @idshibanov , I'm suggesting to use Whirlpools in the same way as Liths but inside `getPriorityTarget` we'll add a penalty depending on possible army loss. Also to make sure that strong heroes won't loose army for nothing inside pathfinding we can add movement penalty (like terrain) depending on army loss. For example, if a hero has 100 Black dragons only then the movement penalty will be (random number) 100 / 3 * 300, where 300 is a single dragon strength. Of course this algorithm is far from ideal but it allows to fix whirlpools by weak heroes if needed and to avoid army losses. Please let me know your thoughts.

oleg-derevenetz:
@ihhub Honestly I don't like an idea to mix movement penalty and army loss penalty, it's a different beasts. Instead I propose to implement a different modes of `AIPathfinder`, one of them will be able to pave the path through whirlpools and the other will not. This way AI will be able to select the mode according to some "external" preferences, which should be more agile.

ihhub:
> @ihhub Honestly I don't like an idea to mix movement penalty and army loss penalty, it's a different beasts. Instead I propose to implement a different modes of `AIPathfinder`, one of them will be able to pave the path through whirlpools and the other will not. This way AI will be able to select the mode according to some "external" preferences, which should be more agile.

This is even better. I'm okay with this. The condition of usage the certain type of pathfinding will be determined externally.

oleg-derevenetz:
@ihhub also we can do the whirlpool-related stuff later in a separate PR to not make this one too complex.

ihhub:
@oleg-derevenetz , I'm totally fine with a separate pull request for Whirlpools.

ihhub (approved):

ihhub:
@oleg-derevenetz , excellent work! Thanks a lot for such enormous improvement!

PR #4897, opened 2022-01-04
MusicRestorer: prevent attempts to play the old music by new instances while the music being restored is starting in the background

oleg-derevenetz:
This should fix sound issues like this: [sound-after-battle-bug.zip](https://github.com/ihhub/fheroes2/files/7810935/sound-after-battle-bug.zip). Set the Battles to Auto Resolve and Music Type to External, skip the turn and press OKAY in the battle result window. You will hear the `MUS::BATTLELOSE` instead of `MUS::COMPUTER_TURN` while the AI hero gets into the boat.


ihhub:

oleg-derevenetz:
Hope this will not break anything.

PR #4902, opened 2022-01-05
Fix the Whirlpools

oleg-derevenetz:
Regression after #4878. The Whirlpools are not working now.


ihhub:
@oleg-derevenetz , thanks for fixing it!

PR #4910, opened 2022-01-09
CMake improvements

oleg-derevenetz:
* Now CMake is able to regenerate and install translation files (not the ideal solution I believe, because it uses custom targets to generate the `src/dist/fheroes2.pot` and run `make` in the `files/lang` to update existing `.po` files and generate `.mo` files, and changing anything in the source tree is bad manners for CMake), but it's better than nothing. Not supported on WIN32 though.
* Now CMake is able to install the `fheroes2.key`, what was previously overlooked.


ihhub:

PR #4925, opened 2022-01-13
Reduce SDL dependency and the amount of warnings in Visual Studio

ihhub:
from 75 to 50. Little more efforts and we can make all warnings to be treated as errors.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -476,7 +476,7 @@ namespace fheroes2
         if ( !empty() ) {
             const size_t totalSize = static_cast<size_t>( _width * _height );
             std::fill( image(), image() + totalSize, value );
-            std::fill( transform(), transform() + totalSize, 0 );
+            std::fill( transform(), transform() + totalSize, static_cast<uint8_t>( 0 ) );
         }
     }
 
@@ -504,8 +504,8 @@ namespace fheroes2
     {
         if ( !empty() ) {
             const size_t totalSize = static_cast<size_t>( _width * _height );
-            std::fill( image(), image() + totalSize, 0 );
-            std::fill( transform(), transform() + totalSize, 1 ); // skip all data
+            std::fill( image(), image() + totalSize, static_cast<uint8_t>( 0 ) );
+            std::fill( transform(), transform() + totalSize, static_cast<uint8_t>( 1 ) ); // skip all data
         }
     }
 
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -228,16 +228,16 @@ std::vector<u8> LoadFileToMem( const std::string & path )
     std::vector<uint8_t> data( length );
 
     size_t dataToRead = static_cast<size_t>( length );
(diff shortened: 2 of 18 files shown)


src/engine/image.cpp
```diff
@@ -476,7 +476,7 @@ namespace fheroes2
         if ( !empty() ) {
             const size_t totalSize = static_cast<size_t>( _width * _height );
             std::fill( image(), image() + totalSize, value );
-            std::fill( transform(), transform() + totalSize, 0 );
+            std::fill( transform(), transform() + totalSize, static_cast<uint8_t>( 0 ) );
```
oleg-derevenetz, line 479: If the problem here (and in other such places) is the sign incompatibility, we just can use the unsigned integer literal (`0U`) instead of the `static_cast`.

src/engine/tools.cpp
```diff
@@ -198,40 +196,50 @@ int Sign( int s )
...
+    std::streamoff length = file.tellg();
+    if ( length < 1 ) {
+        return {};
     }
 
-    SDL_RWclose( rw );
+    std::vector<uint8_t> data( length );
+
+    size_t dataToRead = static_cast<size_t>( length );
+    size_t dataReaded = 0;
```
oleg-derevenetz, line 231: I'd rename this as `dataAlreadyRead` or something like this. AFAIK there is no such form as "readed" :) The past tense of "read" is also "read".

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2990,7 +2990,7 @@ void Battle::Interface::RedrawActionWincesKills( TargetsInfo & targets, Unit * a
     LocalEvent & le = LocalEvent::Get();
 
     // targets damage animation
-    int finish = 0;
+    ptrdiff_t finish = 0;
```
oleg-derevenetz, line 2993: Using `ptrdiff_t` for a plain counter that has nothing to do with pointer arithmetic looks a bit odd. I suggest to use `int` and just convert the result of the `std::count_if()` call below to `int` using `static_cast`.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few notes, please take a look when you have time.

src/engine/tools.cpp
ihhub: Okay, this is  a dumb mistake 🤣

src/engine/image.cpp
ihhub: Sure. Let me try if VS will complain about this.

src/fheroes2/battle/battle_interface.cpp
ihhub: Yeah, it'll work too. I wanted to avoid casting but you're right about pointer related types.

src/engine/image.cpp
ihhub: Nope. Visual Studio still complains about them.

oleg-derevenetz (approved):

PR #4927, opened 2022-01-14
Improve the AI fog discovery logic

oleg-derevenetz:
Discussed in #4904.

The proposed approach to expanding the list of objects that AI can visit for the purpose of fog discovery.

Hi @ihhub what do you think?

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -325,7 +325,7 @@ namespace AI
             AIToMagellanMaps( hero, dst_index );
             break;
 
-            // teleports
+        // teleports
         case MP2::OBJ_STONELITHS:
             AIToTeleports( hero, dst_index );
             break;
@@ -341,7 +341,7 @@ namespace AI
             AIToPrimarySkillObject( hero, objectType, dst_index );
             break;
 
-            // experience modification
+        // experience modification
         case MP2::OBJ_GAZEBO:
             AIToExperienceObject( hero, objectType, dst_index );
             break;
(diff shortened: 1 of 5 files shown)


ihhub:
Hi @oleg-derevenetz , the call `GetObject( true )` which should have **false** value, otherwise we're ignoring heroes. I think we should try an opposite approach to list objects which require fights for heroes. The list will be much smaller.

oleg-derevenetz:
@ihhub The signature of `Maps::Tiles::getObject()` is:

 ```cpp
 ​MP2::MapObjectType ​GetObject​( ​bool​ ignoreObjectUnderHero = ​true​ ) ​const​;
```

So if this parameter is `true`, it means that on tiles with heroes the `OBJ_HEROES` will be returned, and not an object under hero.

ihhub:
> @ihhub The signature of `Maps::Tiles::GetObject()` is:
> 
> ```c++
> ​MP2::MapObjectType ​GetObject​( ​bool​ ignoreObjectUnderHero = ​true​ ) ​const​;
> ```
> 
> So if this parameter is `true`, it means that on tiles with heroes the `OBJ_HEROES` will be returned, and not an object under hero.

It always confuses me... Yes, you're right.

oleg-derevenetz:
I'm just not sure that we need to consider resource objects, bonus objects (such as Gazebo) or monster dwellings as suitable for a fog discovery by itself. Such objects have separate logic AFAIK - for some of them hero should have a free army slot, for some of them he should have money/resources, and so on.

ihhub:
> I'm just not sure that we need to consider resource objects, bonus objects (such as Gazebo) or monster dwellings as suitable for a fog discovery by itself. Such objects have separate logic AFAIK - for some of them hero should have a free army slot, for some of them he should have money/resources, and so on.

You have a valid point. Then let's add those which don't impact AI heroes in any bad way such as fights, resource usage or choosing something. Signs are definitely these kind of places :)

ihhub:

PR #4928, opened 2022-01-15
Rename WhirlpoolTroopLooseEffect() -> WhirlpoolTroopLoseEffect(), minor improvements

oleg-derevenetz:
Minor improvements while I'm here:

* Make the `Troops::GetWeakestTroop()` `const`;
* Rewrite the `WhirlpoolTroopLoseEffect()` in the same manner as the `AIWhirlpoolTroopLoseEffect()` by changing variable names to more telling ones.

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -457,10 +457,10 @@ Troop * Troops::GetFirstValid( void )
     return it == end() ? nullptr : *it;
 }
 
-Troop * Troops::GetWeakestTroop( void )
+Troop * Troops::GetWeakestTroop() const
 {
-    iterator first = begin();
-    iterator last = end();
+    const_iterator first = begin();
+    const_iterator last = end();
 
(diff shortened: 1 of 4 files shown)


src/fheroes2/army/army.cpp
```diff
@@ -457,10 +457,10 @@ Troop * Troops::GetFirstValid( void )
     return it == end() ? nullptr : *it;
 }
 
-Troop * Troops::GetWeakestTroop( void )
+Troop * Troops::GetWeakestTroop() const
```
ihhub, line 460: I'm surprised that compilers don't complain about this as we return a pointer to a member of class.

oleg-derevenetz: @ihhub `Troops` is the container of pointers to `Troop`, not the container of `Troop` by itself.

ihhub:
@oleg-derevenetz , thank you very much for these changes!

PR #4931, opened 2022-01-18
Redesign the hero recruitment process

### LINKED ISSUE #3191: Tavern, Falagar surrender previous turn but is not available in tavern
In fheroes2, Falagar surrender previous turn but is not available in tavern.

In the original game, if one or more heroes surrender, they can all be recruited at the tavern, even if it's a new week that begins.

In fheroes2:

Surrender but not in tavern fh2 01
Surrender but not in tavern fh2 02
Surrender but not in tavern fh2 03

End the turn.
Surrender with Falagar.
Gem cannot surrender, the button is grayed. (https://github.com/ihhub/fheroes2/issues/3190)
Gem flees.

The next turn a new week begins.
Gem can be recruited at the tavern.
In the original game this is possible, but the likelihood of it happening is quite low. (Not correct, it was a confusion with a defeated hero.)

Falagar surrenders the previous turn, but he cannot be recruited at the tavern.
In the original game, if a hero surrenders, he is always recruitable (at least for one full week, and this even if it is a new week that begins).

Here is the save file:
[No surrender with enough gold.zip](https://github.com/ihhub/fheroes2/files/6313940/No.surrender.with.enough.gold.zip)

### LINKED ISSUE #3662: Castle screen, heroes in tavern could change if AI player hires them in its turn
Seems that AI has same heroes for hire in tavern. That's why hero in tavern of a player could disappear.
First day.
image

Second day.
image
[Hero poll.zip](https://github.com/ihhub/fheroes2/files/6631004/Hero.poll.zip)

oleg-derevenetz:
fix #3191
fix #3662

Here are some more notes on this topic - what this PR fixes and what it DOESN'T fix:

* ~~Previously, hiring heroes in castles was done by all kingdoms from a single pool, and the heroes offered for hire could intersect in different kingdoms at the same time - for example, the same hero could be offered for hire in the Blue Kingdom and in the Green one. Now the set of heroes is fixed for the whole week (except for the surrendered heroes) - related to the #3662. The problem here is that there is not enough heroes in the game. We have only 9 heroes in each of the 6 fractions, so if there are 6 players in the game and each of them will hire 8 heroes (max number), then that's what will happen for certain players:~~

No heroes for hire

~~So it's impossible to have non-intersecting heroes for hire for each player and avoid this situation. Hi @ihhub @Branikolog @LeHerosInconnu any thoughts?~~ (P.S. reverted this part, because it seems that OG allows the same heroes to be available for hire in different kingdoms at the same time).

* Previously, just one fled/surrendered hero was remembered. Now two heroes of such kind are remembered, and the second fled hero will not overwrite the first one (but the third one will overwrite one of previous heroes). AFAIK that's how it works in the OG.
* Previously, the fled/surrendered hero was remembered exactly within 7 days from the date of surrender. Now it's not the case - such hero will be overwritten at the beginning of a new week, EXCEPT if he surrendered on Sunday - in this case he will be available for hire next week as well. I believe this logic is closer to the original game.
* ~~This PR DOESN't fix jails. The problem with jails is that in current implementation they really "contain" heroes and these heroes don't get into the general hiring pool until the jail is open. Therefore, if there are a lot of jails on the map, then there may not be enough heroes in the hiring pool.~~

### DIFF
--- a/src/fheroes2/heroes/heroes_recruits.h
+++ b/src/fheroes2/heroes/heroes_recruits.h
@@ -23,28 +23,58 @@
 #ifndef H2RECRUITS_H
 #define H2RECRUITS_H
 
+#include <cstdint>
 #include <utility>
+
+#include "heroes.h"
 
 class Heroes;
 class StreamBase;
 
-class Recruits : public std::pair<int, int>
+struct Recruit
+{
+    Recruit()
+        : id( Heroes::UNKNOWN )
+        , surrenderDay( 0 )
+    {}
+
+    Recruit( const Heroes & hero )
+        : id( hero.GetID() )
+        , surrenderDay( 0 )
+    {}
+
+    Recruit( const Heroes & hero, const uint32_t heroSurrenderDay )
+        : id( hero.GetID() )
+        , surrenderDay( heroSurrenderDay )
+    {}
+
+    int id;
+    uint32_t surrenderDay;
+};
+
+class Recruits : public std::pair<Recruit, Recruit>
 {
 public:
     Recruits();
 
-    void Reset( void );
+    void Reset();
 
-    int GetID1( void ) const;
-    int GetID2( void ) const;
+    int GetID1() const;
+    int GetID2() const;
 
-    Heroes * GetHero1( void );
-    Heroes * GetHero2( void );
+    Heroes * GetHero1() const;
+    Heroes * GetHero2() const;
+
+    uint32_t getSurrenderDay1() const;
+    uint32_t getSurrenderDay2() const;
 
     void SetHero1( Heroes * hero );
     void SetHero2( Heroes * hero );
+
+    void appendSurrenderedHero( Heroes & hero, const uint32_t heroSurrenderDay );
 };
 
-StreamBase & operator>>( StreamBase &, Recruits & );
+StreamBase & operator<<( StreamBase & msg, const Recruit & recruit );
+StreamBase & operator>>( StreamBase & msg, Recruit & recruit );
 
 #endif
--- a/src/fheroes2/heroes/heroes_recruits.h
+++ b/src/fheroes2/heroes/heroes_recruits.h
@@ -68,12 +68,12 @@ public:
     uint32_t getSurrenderDay1() const;
     uint32_t getSurrenderDay2() const;
 
-    void SetHero1( Heroes * hero );
-    void SetHero2( Heroes * hero );
+    void SetHero1( const Heroes * hero );
+    void SetHero2( const Heroes * hero );
 
-    void SetHero2Tmp( Heroes * hero, const uint32_t heroSurrenderDay );
+    void SetHero2Tmp( const Heroes * hero, const uint32_t heroSurrenderDay );
 
-    void appendSurrenderedHero( Heroes & hero, const uint32_t heroSurrenderDay );
+    void appendSurrenderedHero( const Heroes & hero, const uint32_t heroSurrenderDay );
 };
 
 StreamBase & operator<<( StreamBase & msg, const Recruit & recruit );
--- a/src/fheroes2/heroes/heroes_recruits.h
+++ b/src/fheroes2/heroes/heroes_recruits.h
@@ -64,8 +64,8 @@ public:
     Heroes * GetHero1() const;
     Heroes * GetHero2() const;
 
-    uint32_t getSurrenderDay1() const;
-    uint32_t getSurrenderDay2() const;
+    uint32_t getSurrenderDayOfHero1() const;
+    uint32_t getSurrenderDayOfHero2() const;
 
     void SetHero1( const Heroes * hero );
     void SetHero2( const Heroes * hero );
--- a/src/fheroes2/heroes/heroes_recruits.h
+++ b/src/fheroes2/heroes/heroes_recruits.h
@@ -71,6 +71,8 @@ public:
     void SetHero1( Heroes * hero );
     void SetHero2( Heroes * hero );
 
+    void SetHero2Tmp( Heroes * hero, const uint32_t heroSurrenderDay );
+
     void appendSurrenderedHero( Heroes & hero, const uint32_t heroSurrenderDay );
 };
 
--- a/src/fheroes2/heroes/heroes_recruits.h
+++ b/src/fheroes2/heroes/heroes_recruits.h
@@ -41,8 +41,8 @@ public:
     Heroes * GetHero1( void );
     Heroes * GetHero2( void );
 
-    void SetHero1( const Heroes * );
-    void SetHero2( const Heroes * );
+    void SetHero1( Heroes * hero );
+    void SetHero2( Heroes * hero );
 };
 
 StreamBase & operator>>( StreamBase &, Recruits & );
--- a/src/fheroes2/heroes/heroes_recruits.h
+++ b/src/fheroes2/heroes/heroes_recruits.h
@@ -38,14 +38,13 @@ struct Recruit
         , surrenderDay( 0 )
     {}
 
-    Recruit( const Heroes & hero )
-        : id( hero.GetID() )
-        , surrenderDay( 0 )
-    {}
-
     Recruit( const Heroes & hero, const uint32_t heroSurrenderDay )
         : id( hero.GetID() )
         , surrenderDay( heroSurrenderDay )
+    {}
+
+    Recruit( const Heroes & hero )
+        : Recruit( hero, 0 )
     {}
 
     int id;
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1725,8 +1725,8 @@ namespace AI
 
             Heroes * prisoner = world.FromJailHeroes( tileIndex );
 
-            if ( prisoner && prisoner->Recruit( hero.GetColor(), Maps::GetPoint( tileIndex ) ) ) {
-                prisoner->ResetModes( Heroes::JAIL );
+            if ( prisoner ) {
+                prisoner->Recruit( hero.GetColor(), Maps::GetPoint( tileIndex ) );
             }
         }
     }
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -232,7 +232,7 @@ namespace AI
         sortedCastleList.SortByBuildingValue();
 
         if ( ( moreTasksForHeroes && heroes.size() < static_cast<size_t>( heroLimit ) ) || heroes.empty() ) { // safe to cast as heroLimit is > 0
-            Recruits & rec = kingdom.GetRecruits();
+            const Recruits & rec = kingdom.GetRecruits();
             Castle * recruitmentCastle = nullptr;
             int lowestHeroCount = heroLimit;
 
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -754,17 +754,6 @@ Heroes * Castle::RecruitHero( Heroes * hero )
     if ( !hero->Recruit( *this ) )
         return nullptr;
 
-    // actually update available heroes to recruit
-    const Colors colors( Settings::Get().GetPlayers().GetActualColors() );
-
-    for ( const int kingdomColor : colors ) {
-        Kingdom & kingdom = world.GetKingdom( kingdomColor );
-        if ( kingdom.GetLastLostHero() == hero )
-            kingdom.ResetLastLostHero();
-
-        kingdom.GetRecruits();
-    }
-
     Kingdom & currentKingdom = GetKingdom();
     currentKingdom.OddFundsResource( PaymentConditions::RecruitHero( hero->GetLevel() ) );
 
--- a/src/fheroes2/castle/castle_town.cpp
+++ b/src/fheroes2/castle/castle_town.cpp
@@ -358,9 +358,7 @@ Castle::ConstructionDialogResult Castle::openConstructionDialog( uint32_t & dwel
     Kingdom & kingdom = GetKingdom();
 
     Heroes * hero1 = kingdom.GetRecruits().GetHero1();
-
-    Heroes * lastLostHero = kingdom.GetLastLostHero();
-    Heroes * hero2 = lastLostHero && lastLostHero != hero1 ? lastLostHero : kingdom.GetRecruits().GetHero2();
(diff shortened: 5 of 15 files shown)


LeHerosInconnu:
Hello @oleg-derevenetz,

> So it's impossible to have non-intersecting heroes for hire for each player and avoid this situation. Hi @ihhub @Branikolog @LeHerosInconnu any thoughts? (P.S. reverted this part for a while).

> This PR DOESN't fix jails. The problem with jails is that in current implementation they really "contain" heroes and these heroes don't get into the general hiring pool until the jail is open. Therefore, if there are a lot of jails on the map, then there may not be enough heroes in the hiring pool.

Not sure about this (this is just an observation from playing the game), but I believe that in the original game, AI players only recruit 7 heroes at most.

And in the original game they solved these problems by allowing the same hero to be in a game multiple times. :)

In original game:

Kalindra two heroes 01

Here are the test scenario file and the save files:
[Full heroes 01.zip](https://github.com/ihhub/fheroes2/files/7897638/Full.heroes.01.zip)

Related discussion: https://github.com/ihhub/fheroes2/discussions/3199.

oleg-derevenetz:
Hi @LeHerosInconnu

> Not sure about this (this is just an observation from playing the game), but I believe that in the original game, AI players only recruit 7 heroes at most.

Although AI may actually do that, humans (in a multiplayer) are not obliged to do that :) If all 6 players recruit 8 heroes each, there is just 8 heroes left - and if there is a jail on the map, like on Broken Alliance, we have just 7 heroes left, which is clearly not enough for having unique set of heroes per player (even if the same hero will be available for hire in both castle slots). I believe that we should allow one hero to be available for hire in multiple kingdoms at the same time, it seems inevitable. Thus #3662 is not a bug. As a comparison, in HoMM3 there are minimum 15 heroes of each fraction, so in HoMM3 it is possible to afford the luxury of having unique set of recruits for each kingdom.

LeHerosInconnu:
@oleg-derevenetz.

> Hi @LeHerosInconnu
> 
> > Not sure about this (this is just an observation from playing the game), but I believe that in the original game, AI players only recruit 7 heroes at most.
> 
> Although AI may actually do that, humans (in a multiplayer) are not obliged to do that :) If all 6 players recruit 8 heroes each, there is just 8 heroes left - and if there is a jail on the map, like on Broken Alliance, we have just 7 heroes left, which is clearly not enough for having unique set of heroes per player (even if the same hero will be available for hire in both castle slots). I believe that we should allow one hero to be available for hire in multiple kingdoms at the same time, it seems inevitable.

The heroes in the Jails, if they are personalized, will not be available in the tavern (not sure, certainly to some extent)(and the personalization also allows to use the portraits added with the Price of Loyalty).
And since a player can't have more than 8 heroes in his ranks, we stay within the limit of 48 heroes on the adventure map.

Is there any way to know how this is handled in the original game?

One solution would be to add portraits.
Another would be to allow the use of the same portrait but with a different name when necessary.

> Thus #3662 is not a bug.

When a hero is available in the tavern of a player's castle, he must not disappear, for whatever reason, before the end of the current week (unless that player recruits this hero).

> As a comparison, in HoMM3 there are minimum 15 heroes of each fraction, so in HoMM3 it is possible to afford the luxury of having unique set of recruits for each kingdom.

I think this could also happen with Heroes 3, because heroes can be garrisoned and, if I remember correctly, these are not counted in the 8 heroes on the adventure map for the player.

oleg-derevenetz:
@LeHerosInconnu

> When a hero is available in the tavern of a player's castle, he must not disappear, for whatever reason, before the end of the current week (unless that player recruits this hero).

But there is not enough heroes for this behavior. We need at least 10 heroes per fraction for that, but we have just 9 (without heroes available in campaign only), and even if we allow the same hero to present itself in both castle slots, this will not help as well if we have jails on the map. Or we should somehow support the situation with a shortage of heroes in the UI - show something instead of those black squares when there is no hero to hire.

oleg-derevenetz:
@LeHerosInconnu

> And since a player can't have more than 8 heroes in his ranks, we stay within the limit of 48 heroes on the adventure map.

The issue with jails is that they should contain the hero of particular (configured via the map editor) fraction, so we need to reserve a hero from this fraction in advance. If we do not do this, then the situation may turn out, for example, when we know that there is a Knight in this particular jail, but all Knights are already hired by the players (easy, because there are just 9 Knights in the game). FAIL :) Also currently AFAIK the number of heroes is fixed in the game, so we can't dynamically generate arbitrary new heroes from the specified fraction.

oleg-derevenetz:
@LeHerosInconnu

> When a hero is available in the tavern of a player's castle, he must not disappear, for whatever reason, before the end of the current week (unless that player recruits this hero).

In the OG the same hero can present in recruits in multiple kingdoms simultaneously. Consider:

https://user-images.githubusercontent.com/32623900/150198004-ef9a8ab1-2757-46ca-a2c1-74688d249b95.mp4

So #3662 is not a bug.

LeHerosInconnu:
Hello @oleg-derevenetz,

> @LeHerosInconnu
> 
> > When a hero is available in the tavern of a player's castle, he must not disappear, for whatever reason, before the end of the current week (unless that player recruits this hero).
> 
> But there is not enough heroes for this behavior. We need at least 10 heroes per fraction for that, but we have just 9 (without heroes available in campaign only), and even if we allow the same hero to present itself in both castle slots, this will not help as well if we have jails on the map. Or we should somehow support the situation with a shortage of heroes in the UI - show something instead of those black squares when there is no hero to hire.

> @LeHerosInconnu
> 
> > And since a player can't have more than 8 heroes in his ranks, we stay within the limit of 48 heroes on the adventure map.
> 
> The issue with jails is that they should contain the hero of particular (configured via the map editor) fraction, so we need to reserve a hero from this fraction in advance. If we do not do this, then the situation may turn out, for example, when we know that there is a Knight in this particular jail, but all Knights are already hired by the players (easy, because there are just 9 Knights in the game). FAIL :) Also currently AFAIK the number of heroes is fixed in the game, so we can't dynamically generate arbitrary new heroes from the specified fraction.

I think that the custom portrait of the heroes in the jails are first "reserved", but can be changed if needed (in the original game).
I did the test of putting the same hero portrait in two jails, and the portrait of the first released hero was indeed the custom one, but for the second released hero, the portrait was changed (it was the portrait of another hero) and thus was not the custom portrait.

> @LeHerosInconnu
> 
> > When a hero is available in the tavern of a player's castle, he must not disappear, for whatever reason, before the end of the current week (unless that player recruits this hero).
> 
> In the OG the same hero can present in the set of recruits in multiple kingdoms simultaneously. Consider:
> 
>  DOSBox.0.74-2.1.Cpu.speed_.max.100.cycles.Frameskip.0.Program_.HEROES2.2022-01-19.22-06-06.mp4 
> So #3662 is not a bug. Given the lack of heroes, I'd say that this behavior is inevitable.

What happens when one of the two players dismisses one of his heroes and recruits the hero present in the tavern of both players?
Surely the portrait of the hero present in the taverns is replaced by the portrait of the previously dismissed hero?

In case the limit of available hero portraits is reached it is understandable to proceed this way.
But in the course of the game, as long as the limit of available hero portraits is not reached, this should not happen.

oleg-derevenetz:
@LeHerosInconnu

> I think that the custom portrait of the heroes in the jails are first "reserved", but can be changed if needed (in the original game).

The question is not the portrait, but the fraction (race), which is a mandatory characteristic of a prisoner hero. AFAIK you can't put a Knight in a jail, but get, for example, a Wizard from it. One Knight will be reserved for this jail and will not be available for hire via Tavern.

> Surely the portrait of the hero present in the taverns is replaced by the portrait of the previously dismissed hero?

Yes, if there is a lack of heroes.

> But in the course of the game, as long as the limit of available hero portraits is not reached, this should not happen.

I'm not sure about that, I believe that there is always such a probability, but this probability is the lower the larger the number of free heroes.

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu
> 
> > I think that the custom portrait of the heroes in the jails are first "reserved", but can be changed if needed (in the original game).
> 
> The question is not the portrait, but the fraction (race), which is a mandatory characteristic of a prisoner hero. AFAIK you can't put a Knight in a jail, but get, for example, a Wizard from it. One Knight will be reserved for this jail and will not be available for hire via Tavern.
> 
> > Surely the portrait of the hero present in the taverns is replaced by the portrait of the previously dismissed hero?
> 
> Yes, if there is a lack of heroes.
> 
> > But in the course of the game, as long as the limit of available hero portraits is not reached, this should not happen.
> 
> I'm not sure about that, I believe that there is always such a probability, but this probability is the lower the larger the number of free heroes.

So, the jails reserve a specific portrait as a priority (and it can eventually be replaced by a portrait of a same hero type if needed).
We can then consider that the jails still present on the adventure map don't really have a defined portrait but a portrait of a hero type.
This means that if a portrait is needed in a tavern (the limit of the number of hero portraits on the adventure map is reached), the jails don't need a specific portrait.
I think that in the original game, there is always a portrait of each hero type available in reserve no matter what.
All players together, there should only be eight heroes of the same type recruited at the same time and therefore always one portrait of each hero type available for the tavern or jail.

oleg-derevenetz:
@LeHerosInconnu

> This means that if a portrait is needed in a tavern (the limit of the number of hero portraits on the adventure map is reached), the jails don't need a specific portrait.
I think that in the original game, there is always a portrait of each hero type available in reserve no matter what.
All players together, there should only be eight heroes of the same type recruited at the same time and therefore always one portrait of each hero type available for the tavern or jail.

Consider the following situation: there are two jails on the map with a Knight in each (no specific portrait) and there are two players on this map. Consider also that one player has 4 Knights recruited, and another player also has 4 Knights (he also may have 5 Knights recruited BTW, because why not). If both of them have 8 Knights recruited (4+4), then there will be no Knight for the second jail, and if both if them have 9 Knights recruited (4+5), then there will be no Knight even for the first jail :) So it seems that there is a need to reserve the same amount of heroes as the number of jails with prisoners of a corresponding race, and just one hero in reserve is not enough.

LeHerosInconnu:
Hello @oleg-derevenetz,

> @LeHerosInconnu
> 
> > This means that if a portrait is needed in a tavern (the limit of the number of hero portraits on the adventure map is reached), the jails don't need a specific portrait.
> > I think that in the original game, there is always a portrait of each hero type available in reserve no matter what.
> > All players together, there should only be eight heroes of the same type recruited at the same time and therefore always one portrait of each hero type available for the tavern or jail.
> 
> Consider the following situation: there are two jails on the map with a Knight in each (no specific portrait) and there are two players on this map. Consider also that one player has 4 Knights recruited, and another player also has 4 Knights (he also may have 5 Knights recruited BTW, because why not). If both of them have 8 Knights recruited (4+4), then there will be no Knight for the second jail, and if both of them have 9 Knights recruited (4+5), then there will be no Knight even for the first jail :) So it seems that there is a need to reserve the same amount of heroes as the number of jails with prisoners of a corresponding race, and just one hero in reserve is not enough.

I did some testing for this.
- First case, four Knight heroes on the adventure map for blue and green players, with two Knights heroes (without customizing the portrait) in both jails.
- Second case, five Knight heroes on the adventure map for blue and green players, with two Knights heroes (without customizing the portrait) in both jails.
- Third case, ten jails with all Knight heroes (without customizing the portrait).

In original game:
First case.
Heroes portrait jail 001
Heroes portrait jail 002
Heroes portrait jail 003
Heroes portrait jail 004

In fheroes2:
First case.
Heroes portrait jail 009

In original game:
Second case.
Heroes portrait jail 005
Heroes portrait jail 006
Heroes portrait jail 007
Heroes portrait jail 008

In fheroes2:
Second case.
Heroes portrait jail 010

In original game:
Third case.
Heroes portrait jail 011
Heroes portrait jail 012

In fheroes2:
Third case.
Heroes portrait jail 013

In the first two cases, the Knight hero type is preserved for the heroes in the jails, but the Knight hero type is not preserved for the heroes on the adventure map.
So we can deduce that in the original game the hero type is preserved first in the jails, but the hero type is changed for the hero on the adventure map if necessary.
In the third case, as soon as all nine Knight heroes are present on the adventure map, the type of heroes in the jails is no longer respected.
It seems to work almost the same way in Heroes 2 and in fheroes2 in the current version.

Here are the test scenario files:
[Heroes portrait jail 01.zip](https://github.com/ihhub/fheroes2/files/7916659/Heroes.portrait.jail.01.zip)

oleg-derevenetz:
Hi @LeHerosInconnu, yeah, it seems that both in OG and fheroes2 jails have a precedence in terms of reserving heroes from a certain race (heroes are reserved for jails first, and if there is not enough heroes from the given race for all jails, other races are used). Well, so much the better :) I'll update the description for this PR.

ihhub (approved):

oleg-derevenetz:
Hi @ihhub pls wait a second, just one more improvement... :)

oleg-derevenetz:
Hi @ihhub @Branikolog I would like to double-check the `TYPE_HIREABLE_HERO` campaign reward. Do you happen to have a savefile of the last day of the (almost completed) mission with such a reward (that is, in the next mission there should be a special hero as a reward)? :)

oleg-derevenetz:
> Hi @ihhub @Branikolog @LeHerosInconnu I would like to double-check the `TYPE_HIREABLE_HERO` campaign reward. Do you happen to have a savefile of the last day of the (almost completed) mission with such a reward (that is, in the next mission there should be a special hero as a reward)? :)

Nevermind, I already found such a savefile :)

ihhub:
@oleg-derevenetz , are we good to merge it?

src/fheroes2/heroes/heroes_recruits.h
```diff
@@ -23,28 +23,59 @@
...
+    void Reset();
+
+    int GetID1() const;
+    int GetID2() const;
+
+    Heroes * GetHero1() const;
+    Heroes * GetHero2() const;
+
+    uint32_t getSurrenderDay1() const;
+    uint32_t getSurrenderDay2() const;
```
ihhub, line 68: Are these related to hero1 and hero2? If yes should we name them as `getSurrenderDayHero1` and `getSurrenderDayHero2` because now it can be read as day 1 and day 2?

oleg-derevenetz:
Hi @ihhub

> @oleg-derevenetz , are we good to merge it?

I believe yes, I checked the campaign heroes, everything seems to work.

src/fheroes2/heroes/heroes_recruits.h
oleg-derevenetz: Yes, this should look better. I'll rename them.

ihhub:
@oleg-derevenetz , huge thanks for these valuable changes!

PR #4937, opened 2022-01-21
German font generation (support Umlaut & Eszett)

Laserlicht:
Generation of the chars äöüÄÖÜß from other chars


a1exsh:
@Laserlicht nice — it seems to work! :)

Can we maybe shift the umlaut over small "u" at least one pixel up?
image

Closer look:
image

There are some extra shadow pixels in the ß letter in the normal font:
image

image

Laserlicht:
Hi @a1exsh 
Thank you for testing. 😀

Problem with the Shadow is fixed by: c254ee178f9f0f5f1602f7a8a242ba424df447b6
And the äöü with: dc4d2c84537e7865e33bf9028d831404c2e550a5

image

Should be okay now! 😃

a1exsh:
> image

Already much better, but I still notice small flaws, which I tried to correct here:
image

1. ä: the right dot should look more like the left one.
2. ö: the left dot and its shadow are missing one pixel (at the bottom-left).
3. ß: the shadow needs correction.

Also, I'm not sure about the form of the dots over Ä, Ö, Ü: have you tried to use the same form (2x2) as for the small letters?

Laserlicht:
Now it look like this:
image
image

Ä Ö Ü -> I've adapted it from the original german font
image
image
Okay the dots are one pixel higher.

What would be the best solution here?

a1exsh:
> Ä Ö Ü -> I've adapted it from the original german font

I see.  Let's keep them like that then.  Can you also share how the letter `ß` looks in the original game resources?  I feel like our variant could be improved a bit.

Laserlicht:
@a1exsh Here is some comparison
It seems that fheroes2 modify the chars a little bit.

image
image

image
image

Laserlicht:
#4320 should be merged together with this one. Otherwise, there are no umlauts with German assets.

Or should we do the generation of the letters in any case?

dimag0g:
> #4320 should be merged together with this one. Otherwise, there are no umlauts with German assets.

The charset of the translation file depends only on the Makefile, you should overtake at least that change, assuming your generated characters correspond to CP1252.

I don't know what the rest of #4320 is about. I thought it was supposed to generate the German characters, so if this PR already does that, there may be no need to merge.

Laserlicht:
> The charset of the translation file depends only on the Makefile, you should overtake at least that change, assuming your generated characters correspond to CP1252.
It has directly nothing to do with the encoding.

This pull request generates the letters. But currently fheroes2 only does this if no german assets were found.

If german assets were found it tries to load them. This is what #4320 is for, to move the original assets to the right places according to CP1252 encoding.

Laserlicht:
I have now changed it. it generates the characters now always, even if there would be in the resources.

So #4320 is no longer necessary.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2823,5 +3012,17 @@ namespace fheroes2
 
             return false;
         }
+
+        bool SkipAlphabetRessource( const SupportedLanguage language )
```
ihhub, line 3016: Could you please explain why do we need this function? I can clearly see that German version of HoMM2 has all required symbols.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -180,20 +180,209 @@ namespace
...
+            fheroes2::Copy( font[223 - 32], 4, 3 + 3, font[223 - 32], 3, 8 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 4, 3 + 3, font[223 - 32], 3, 9 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 4, 3 + 3, font[223 - 32], 2, 10 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 3, 6 + 3, font[223 - 32], 2, 7 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 3, 6 + 3, font[223 - 32], 2, 8 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 3, 6 + 3, font[223 - 32], 2, 9 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 3, 6 + 3, font[223 - 32], 1, 10 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 7, 4 + 3, font[223 - 32], 4, 8 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 7, 4 + 3, font[223 - 32], 3, 10 + 3, 1, 1 );
+            fheroes2::Copy( font[223 - 32], 8, 5 + 3, font[223 - 32], 4, 9 + 3, 1, 1 );
```
ihhub, line 302: I strongly believe that we can copy much bigger chunks of images than 1 pixel only. This letter is a slight modification of B. In this case what we need is to erase certain parts of the image and copy few pixels only. Please use `fheroes2::FillTransform` to erase certain parts of the image.

The same applies for small font.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -180,20 +180,209 @@ namespace
...
+            font[228 - 32].reset();
+            fheroes2::Copy( font[65], 0, 0, font[228 - 32], 0, 3, font[65].width(), font[65].height() );
+            fheroes2::Copy( font[228 - 32], 3, 0 + 3, font[228 - 32], 3, 0, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 3, 0 + 3, font[228 - 32], 6, 0, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 2, 1 + 3, font[228 - 32], 3, 1, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 2, 1 + 3, font[228 - 32], 2, 0, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 2, 1 + 3, font[228 - 32], 5, 0, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 2, 1 + 3, font[228 - 32], 6, 1, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 1, 0 + 3, font[228 - 32], 2, 1, 1, 1 );
+            fheroes2::Copy( font[228 - 32], 1, 0 + 3, font[228 - 32], 5, 1, 1, 1 );
```
ihhub, line 237: Do we need to copy even shadow-related pixels? `updateShadow` will add all missing transform pixels.

ihhub (commented):
Hi @Laserlicht , I left few comments here. Would you mind to take a look?

src/fheroes2/agg/agg_image.cpp
ihhub: On top of this we can just copy the same pixels from one place to another so no need to repeat 1 by 1 pixel copy for the second dot of the character.

src/fheroes2/agg/agg_image.cpp
Laserlicht: Yes, the german version has all the letters. But I think the generation has certain advantages. The large letters do not have to be squeezed (y-axis) together, as in the original. See also the example images above.

So you can discuss what's better.

src/fheroes2/agg/agg_image.cpp
Laserlicht: Okay, I did not know the function. Will have a look at it.

src/fheroes2/agg/agg_image.cpp
Laserlicht: These are partly different shades of gray that are copied. So actually no shadows.

src/fheroes2/agg/agg_image.cpp
ihhub: I would insist for now to keep the original version of German letters. We do the same for other languages even our letters sometimes better than the original.

Laserlicht: Okay... I've changed it back.

Because of this #4320 needs to merged too.

ihhub:
@Laserlicht , thank you so much for your contribution to the project!

PR #4943, opened 2022-01-22
Update de.po

a1exsh:
Spotted some minor issues, such as missing umlauts, some more cases of Sie/du,
etc.


a1exsh:
@dimag0g FYI

a1exsh:
I have doubts about this one:
image

For one, it should be "Einfaches Pfadfinden", to be consistent with "Einfache Diplomatie".  On the other hand, that still might be not correct in other contexts.  How about we sidestep the issues, by writing it in this form: "Pfadfinden, verbessert" (maybe without comma is also OK)?  This way the context matters less and we don't have to differentiate on the m/f/n for the adjectives.

dimag0g:
I took a look at the changes you did so far and have no remarks or corrections (not surprising considering my rather poor German).

Regarding skills translation: "Einfach Pfadfinden" is a translation of "Basic Pathfinding", so there should be no problem to fix it to "Einfaches Pfadfinden", just like it was done for "Diplomatie".

"Einfach" on the skill icon is an independent translation of "skill|Basic". If you think "Pfadfinden \n Einfach" on the skill icon is unacceptable we could change it to "Pfadfinden \n verbessert". All other translations featuring "Basic" would have to be fixed as well.

dimag0g:
While you're at it, could you also:

- replace Aussichtsturm with Beobachtungsturm for consistency
- replace Schiffbruch with Schiffswrack for consistency
- replace "Überlebenden eines Schiffbruchs" with "Schiffbrüchiger" for consistency
- replace Truhe with Kiste when talking about "Schatzkiste"/"Seekiste"
- replace Brunnen with Quelle when talking about artesian springs (I feel the "Artesischer Brunnen" from the GoG translation is wrong)
- replace Quelle with Brunnen when talking about fountains (again, GoG translating "Fountain" as "Quelle" doesn't feel right)
- replace "Magische Gärten" with "Zaubergärten" for consistency
- replace Getränk with Schluck (I seem to have missed some)

a1exsh:
@dimag0g nicely spotted!  I will open this for review after the English update is merged.

a1exsh:
Marking this ready for review.

ihhub:
@a1exsh , thank you very much for your changes!

PR #4947, opened 2022-01-22
Modify army order setting

ihhub:
Now it's a part of battle settings window and can be changed anytime during battle. The current unit has the same cycling color as the current monster. Moved creatures have grayed icons. Border of rectangles associates with the color of monsters.

Also drawings now are faster.

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -274,7 +274,6 @@ void Battle::DialogBattleSettings( void )
         //     Dialog::Message( _( "Monster Info" ), _( "Toggle the monster info window, which shows information on the active and targeted monsters." ), Font::BIG );
         // }
 
-        
         bool saveShowArmyOrder = false;
         if ( le.MouseClickLeft( optionAreas[1] ) ) {
             conf.setBattleShowArmyOrder( !conf.BattleShowArmyOrder() );
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -274,25 +274,26 @@ void Battle::DialogBattleSettings( void )
         //     Dialog::Message( _( "Monster Info" ), _( "Toggle the monster info window, which shows information on the active and targeted monsters." ), Font::BIG );
         // }
 
+        
         bool saveShowArmyOrder = false;
+        if ( le.MouseClickLeft( optionAreas[1] ) ) {
+            conf.setBattleShowArmyOrder( !conf.BattleShowArmyOrder() );
+            saveShowArmyOrder = true;
+        }
+        else if ( le.MousePressRight( optionAreas[1] ) ) {
+            Dialog::Message( _( "Army Order" ), _( "Toggle to display army order during the battle." ), Font::BIG );
+        }
+
+        bool saveAutoSpellCast = false;
         if ( le.MouseClickLeft( optionAreas[2] ) ) {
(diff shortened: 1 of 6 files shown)


src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -273,10 +274,10 @@ void Battle::DialogBattleSettings( void )
         //     Dialog::Message( _( "Monster Info" ), _( "Toggle the monster info window, which shows information on the active and targeted monsters." ), Font::BIG );
         // }
 
-        bool saveAutoSpellCast = false;
+        bool saveShowArmyOrder = false;
         if ( le.MouseClickLeft( optionAreas[2] ) ) {
             conf.setBattleAutoSpellcast( !conf.BattleAutoSpellcast() );
-            saveAutoSpellCast = true;
+            saveShowArmyOrder = true;
```
oleg-derevenetz, line 280: I believe there is some inconsistency. Why we are setting the `saveShowArmyOrder` here but calling the `conf.setBattleAutoSpellcast()`?

src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -285,6 +286,15 @@ void Battle::DialogBattleSettings( void )
                 Font::BIG );
         }
 
+        bool saveAutoSpellCast = false;
+        if ( le.MouseClickLeft( optionAreas[1] ) ) {
+            conf.setBattleShowArmyOrder( !conf.BattleShowArmyOrder() );
+            saveAutoSpellCast = true;
```
oleg-derevenetz, line 292: The same inconsistency as earlier. We are setting the `saveAutoSpellCast` here but calling the `conf.setBattleShowArmyOrder()`.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few comments, please take a look when you have time.

oleg-derevenetz:
Is there no more suitable picture for this setting? I can't even say for sure what is drawn in the picture used now :)

src/fheroes2/battle/battle_dialogs.cpp
ihhub: It's a bug.

ihhub:
> Is there no more suitable picture for this setting? I can't even say for sure what is drawn in the picture used now - some sort of a paper with cooking recipe may be :)

Haha! You're right. It feels ugly. I'll try to come up with some background image for this.

ihhub:
Hi @oleg-derevenetz , I honestly out of ideas to replace these images with something better :(

oleg-derevenetz:
@ihhub

> Hi @oleg-derevenetz , I am honestly out of ideas to replace these images with something better :(

Well, let's leave it as is for now :) May be in future some volunteer artist will be able to create something.

oleg-derevenetz:
Hi @ihhub BTW what is the procedure of making a new entry in `resurrection.h2d`? What are requirements for images? If I have some BMP/JPG/PNG file, how can I add it to the `resurrection.h2d` and use it as an sprite in the game?

ihhub:
Hi @oleg-derevenetz ,

First of all you need to convert your image into readable by the game format. This can be done by `fheroes2::Load()` call. To see how it looks like in the same you can save it right after loading by calling `fheroes2::Save()`.

Once your image is loaded you need to load `resurrection.h2d` file. For this purpose use `fheroes2::H2RReader` class. Then create a new instance of `fheroes2::H2Writer` class and call `add` method with `fheroes2::H2RReader` instance. After this use `fheroes2::writeImageToH2D` function to add your newly loaded image into H2RWriter and then call `write` method of the class.

Issue #4265, opened 2021-09-27
New week, week of the creature, available heroes, generated resources, etc. should be the same if a save file is loaded


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

The available heroes, generated resources, etc., etc., etc. should be the same if a save file is loaded.
All data needed and used for these events should be generated at the beginning of the scenario.
For example, the generation of resources for the various generators on the adventure map should be determined and be the same in all cases.
For the available heroes, if an action is repeated in the same way after loading a save file (with no heroes recruited by any of the players), the same result should be obtained.

In fheroes2:
![New week 003](https://user-images.githubusercontent.com/43583994/134916578-18ebc7ac-29d7-43fc-869d-56440eda133a.png)
![New week 004](https://user-images.githubusercontent.com/43583994/134916596-7be2f528-2665-493f-ab38-a3658a531092.png)
![New week 005](https://user-images.githubusercontent.com/43583994/134916611-79bc21bc-ca9b-4812-8b67-d3f26e41d555.png)
![New week 001](https://user-images.githubusercontent.com/43583994/134916713-29558507-9712-4364-8cfd-d5de37fa74ef.png)

In fheroes2 (after reloading):
![New week 006](https://user-images.githubusercontent.com/43583994/134916655-07629239-a03e-40cb-9782-5dad65f72ecf.png)
![New week 007](https://user-images.githubusercontent.com/43583994/134916667-07a01dfd-f57b-4f84-991f-edc802ea9d71.png)
![New week 008](https://user-images.githubusercontent.com/43583994/134916674-3e69d14e-4a7b-409f-88a7-2396e3d88807.png)
![New week 002](https://user-images.githubusercontent.com/43583994/134916814-9d5c40ac-f7aa-4942-8948-d50f95739197.png)

End the turn.
Reload and end the turn.

### Save file

Here is the save file:
[Different heroes in tavern 01.zip](https://github.com/ihhub/fheroes2/files/7236538/Different.heroes.in.tavern.01.zip)

### Additional info

Version 0.9.7 4208

vincent-grosbois:
[...]

PR #4948, opened 2022-01-22
Fix the "week of monsters" and a few other issues

### LINKED ISSUE #4944: Month of monsters — no monsters spawned on adventure map
### Preliminary checks

### Platform

Linux

### Describe the bug

Playing the "Voyage Home" campaign, I'm noticing that no monsters are appearing on the map on the first day of then second month:
image

Load the game from the attached file and end the turn.  The monsters should have appeared on the map, but I don't see any single one (almost 100% of the map is already visible to the player).

### Save file

[AUTOSAVE.zip](https://github.com/ihhub/fheroes2/files/7918194/AUTOSAVE.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #4944 - regression after #4688 (week type was calculated too late).

This PR also fixes potential dangling pointer issue with the `World::_rumor` (if `World::vec_rumors` was cleared and then new rumors were loaded at the different address, we could get a dangling pointer) and also removes the unused `World::week_next` (partially, because it is present in a savefile, and I don't want to clash with #4931 yet, so I left a TODO comment for now).


src/fheroes2/world/world.cpp
```diff
@@ -1419,8 +1413,11 @@ StreamBase & operator<<( StreamBase & msg, const World & w )
     const uint16_t width = static_cast<uint16_t>( w.width );
     const uint16_t height = static_cast<uint16_t>( w.height );
 
+    // TODO: remove this
```
ihhub, line 1416: Shouldn't we then not to even write week data? We can still read it for older versions.

ihhub (changes requested):
Hi @oleg-derevenetz , I left one comment here. Could you please take a look?

src/fheroes2/world/world.cpp
oleg-derevenetz: Hi @ihhub Yes, I'll do this for sure, but the problem is in #4931 - it also messes with savefiles, so I do not want to add the code here that I'll need to rewrite anyway when #4931 will be merged. So we can either wait until #4931 will be merged, rebase this PR and add the corresponding logic, or we can make a separate PR for this.

ihhub: Let's merge #4931 first. I left there just one comment. Please take a look.

oleg-derevenetz:
Hi @ihhub I have an idea to make the world week seed dependent only on map seed and number of the week.

Pros:

* We can calculate different things (such as week type and rumor) on the fly in a deterministic manner. No need to save this data in the savefile and periodically recalculate (what is possible to forget about);
* Such an approach can (in perspective) be used to deal with issues described in the #4265.

Cons:

* This approach does not depend on the player's actions, and thus is somewhat more predictable.

What do you think?

ihhub:
> Hi @ihhub I have an idea to make the world week seed dependent only on map seed and number of the week.
> 
> Pros:
> 
>     * We can calculate different things (such as week type and rumor) on the fly in a deterministic manner. No need to save this data in the savefile and periodically recalculate (what is possible to forget about);
> 
>     * Such an approach can (in perspective) be used to deal with issues described in the [New week, week of the creature, available heroes, generated resources, etc. should be the same if a save file is loaded #4265](https://github.com/ihhub/fheroes2/issues/4265).
> 
> 
> Cons:
> 
>     * This approach does not depend on the player's actions, and thus is somewhat more predictable.
> 
> 
> What do you think?

I think it should be dependent on monsters on the map. If I remember correctly this is how it works in the original game.

oleg-derevenetz:
@ihhub

> I think it should be dependent on monsters on the map. If I remember correctly this is how it works in the original game.

I don't see any dependency on the map monsters in week of/month of generation code (`kingdom/week.cpp`) As I can see, it depends only on a week/month number and the seed.

ihhub:
> @ihhub
> 
> > I think it should be dependent on monsters on the map. If I remember correctly this is how it works in the original game.
> 
> I don't see any dependency on the map monsters in week of/month of generation code (`kingdom/week.cpp`) As I can see, it depends only on a week/month number and the seed.
> 
> P.S. So if I kill some monster of the map, it should somehow affect the week/month type?

You should. @Branikolog  can you please shed some light on this question?

oleg-derevenetz:
@ihhub

> You should. @Branikolog can you please shed some light on this question?

OK, I reverted this part, now the seed used for the calculation of week/month type depends on map state as before.

ihhub:
@oleg-derevenetz , hold on please with reverting back changes :) We didn't even hear the expert @Branikolog . I wrote "should" based on my observations and I still might be wrong.

@LeHerosInconnu could you please tell us whether new week type depends on monsters on the map?

oleg-derevenetz:
Hi @ihhub I just did a few tests - started the multiplayer game on a apecially crafted map for 2 players, just always skipped turns for every player (no hero moves, no attacks, etc), and loaded the AUTOSAVE several times on 121 (thus effectively repeating the new week's event). Each time I got a new week of a different type. Thus, it seems that in OG the type of a new week is not deterministic. You can try it yourself: load [NEWWEEK.zip](https://github.com/ihhub/fheroes2/files/7952751/NEWWEEK.zip) and just skip the turn, then load it again and skip the turn once again and so on. In each case, you will receive a different type of new week.

LeHerosInconnu:
Hello @ihhub and @oleg-derevenetz,

> @oleg-derevenetz , hold on please with reverting back changes :) We didn't even hear the expert @Branikolog . I wrote "should" based on my observations and I still might be wrong.
> 
> @LeHerosInconnu could you please tell us whether new week type depends on monsters on the map?

There is information about "Week of" and "Month of" from the linked page (towards the end of the page) in this discussion: https://github.com/ihhub/fheroes2/discussions/4676. :)

oleg-derevenetz:
Hi @LeHerosInconnu

> There is information about "Week of" and "Month of" from the linked page (towards the end of the page) in this discussion: #4676. :)

Yes, but the information I need is not there :) I need to know is the probability of every single week of/month of fully random or deterministic, and, if it is deterministic, what does it depend on. My experiments above have shown that in OG this probability is (apparently) completely random, but I have a proposal to make it deterministic (depending on the map seed and week/month number).

LeHerosInconnu:
Hello @oleg-derevenetz,

> Hi @LeHerosInconnu
> 
> > There is information about "Week of" and "Month of" from the linked page (towards the end of the page) in this discussion: #4676. :)
> 
> Yes, but the information I need is not there :) I need to know is the probability of every single week of/month of fully random or deterministic, and, if it is deterministic, what does it depend on. My experiments above have shown that in OG this probability is (apparently) completely random, but I have a proposal to make it deterministic (depending on the initial map seed and week/month number).

It's random in the original game but not sure if it is "fully random".
When reloading saves, the "Week of" and "Month of" changes.
I always found it a bit too easy to influence these events in this way (reloading saves), and I would prefer a deterministic way of working. :)
I think it would be a good improvement over the original game.
But it's okay to keep the randomness from the original game.

Here are some save files to test this:
[New week and month 01.zip](https://github.com/ihhub/fheroes2/files/7961941/New.week.and.month.01.zip)

oleg-derevenetz:
> It's random in the original game but not sure if it is "fully random".

By "fully random" I mean that it doesn't depend on player's actions. Since I always start from he same map state (by loading the same savefile) and then repeat the same set of actions (or rather, no actions) and get different results each time, then it's "fully random" :)

LeHerosInconnu:
Hello @oleg-derevenetz,

> > It's random in the original game but not sure if it is "fully random".
> 
> By "fully random" I mean that it doesn't depend on player's actions. Since I always start from the same map state (by loading the same savefile) and then repeat the same set of actions (or rather, no actions) and get different results each time, then it's "fully random" :)

One thing I noticed in the original game is that for the Month of the Plague it is sometimes possible to avoid it by reloading a save and moving a hero or buying creatures but other times no matter what you change the Month of the Plague always happens.

ihhub:
Hi @oleg-derevenetz  and @LeHerosInconnu , am I right to say that dependency on monsters is considered more random than dependency from date only making it close to the original? If yes then we can proceed with merging.

oleg-derevenetz:
Hi @ihhub I'd say that in OG there is no dependency on anything :) Also I understand so that @LeHerosInconnu is not against an additional determinism in this regard.

ihhub:
In this case @oleg-derevenetz huge thanks for this work!

PR #4950, opened 2022-01-23
Fix a few code smells detected by the SonarCloud

oleg-derevenetz:


ihhub:

PR #4951, opened 2022-01-25
Add "Battle Only" button generation for German version of the game

### LINKED ISSUE #3116: unneccessary button in new game menu
When you open the new menu there is a button that has a totally messed up description. Clicking on it has the same effect as clicking on the "multiplayer" button:
fheroes092newgamemenubug

fheroes 0.9.2 ubuntu_sdl2 version; game assets from the german gog dos version

Laserlicht:
Generation of "Kampf" button.

image

close #3116


src/fheroes2/agg/agg_image.cpp
```diff
@@ -33,6 +33,7 @@
 #include "image_tool.h"
 #include "pal.h"
 #include "screen.h"
+#include "settings.h"
```
ihhub, line 36: This file inclusion is the main concern from my side. It mustn't be included as agg_image logic should not depend on settings of the game. Also the proposed solution won't work for cases when a player switches between languages. Once ICN is loaded it doesn't change.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1508,6 +1509,80 @@ namespace fheroes2
                 ApplyPalette( _icnVsSprite[icnId][i], palette );
             }
         }
+        
+        void generateICNEnglish( int id )
```
ihhub, line 1513: Why don't we detect CRC32 of each English and then choose the logic suitable for the detected resources? Having English based game and using German language will lead to very interesting issues. The same applies for English language within German resources. Could you please verify this?

ihhub (changes requested):
Hi @Laserlicht , I left few comments and suggestions regarding this pull request. Could you please have a look at them?

src/fheroes2/agg/agg_image.cpp
Laserlicht: Okay, I see. Of course, I'm not that into it yet.

src/fheroes2/agg/agg_image.cpp
Laserlicht: I changed it and used the function from ui_language.cpp for it. Is that better?

ihhub: Yes, it is.

ihhub: Yes, it is.

ihhub:
@Laserlicht , thank you very much for your contribution to the project!

PR #4952, opened 2022-01-25
Fix a few recent code smells

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks a lot for cleaning up the code!

PR #4957, opened 2022-01-26
Made language area and campaign choice selection to cover text too

### LINKED ISSUE #4940: Click item text to select from a list of choices
### Preliminary checks

In the UI elements with a list of choices the item can only be selected by clicking on the radio button, e.g:
image
image

It would be nicer if clicking on the item text also worked to select that item.

### Additional info

_No response_

ihhub:
close #4940

### DIFF
--- a/src/fheroes2/dialog/dialog_language_selection.cpp
+++ b/src/fheroes2/dialog/dialog_language_selection.cpp
@@ -56,6 +56,7 @@ namespace fheroes2
         const int32_t languageAreaWidth = 200;
         const int32_t languageAreaHeight = 25;
         const int32_t offsetFromBorders = 10;
+        const int32_t offsetFromButton = 15;
 
         const int32_t languageCount = static_cast<int32_t>( languages.size() );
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/dialog/dialog_language_selection.cpp
```diff
@@ -108,6 +108,12 @@ namespace fheroes2
 
         SupportedLanguage chosenLanguage = languages[selectionId];
 
+        std::vector<fheroes2::Rect> languageArea( languages.size() );
+        for ( size_t i = 0; i < languages.size(); ++i ) {
+            languageArea[i] = buttonGroup.button( i ).area();
+            languageArea[i].width += languageAreaWidth + offsetFromBorders;
```
oleg-derevenetz, line 114: Why `offsetFromBorders` and not `15`? I suppose the distance between radio button sprite and text is `15` pixels?

oleg-derevenetz (commented):
Hi @ihhub I just left one question, please take a look.

src/fheroes2/dialog/dialog_language_selection.cpp
ihhub: Ooops. You're right.

oleg-derevenetz (approved):

PR #4958, opened 2022-01-26
Fix race condition on initialization of paletteHolder

dimag0g:
There is a race condition between the initialization of two global variables:

    image_palette.cpp:42    PaletteHolder paletteHolder;
    screen.cpp:184          const uint8_t * currentPalette = PALPalette();

If the compiler choses to initialize `currentPalette` first, it uses data from uninitialized `paletteHolder`, typically resulting in all-black palette. This is what happens e.g. with Mingw64 toolchain.

The proposed solution is to access `paletteHolder` via a singleton.

### DIFF
--- a/src/engine/image_palette.cpp
+++ b/src/engine/image_palette.cpp
@@ -31,22 +31,28 @@ namespace
 
     struct PaletteHolder
     {
+    private:
         PaletteHolder()
         {
             std::copy_n( kb_pal, paletteSize, gamePalette.begin() );
         }
 
+    public:
+        static PaletteHolder & instance()
+        {
+            static PaletteHolder paletteHolder;
+            return paletteHolder;
+        }
+
         std::array<uint8_t, paletteSize> gamePalette;
     };
-
-    PaletteHolder paletteHolder;
 }
 
 namespace fheroes2
 {
     const uint8_t * getGamePalette()
     {
-        return paletteHolder.gamePalette.data();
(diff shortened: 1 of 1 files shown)


src/engine/image_palette.cpp
```diff
@@ -36,17 +36,21 @@ namespace
             std::copy_n( kb_pal, paletteSize, gamePalette.begin() );
         }
 
+        static PaletteHolder & singleton()
```
ihhub, line 39: Could you please name this method as `instance`? This is done for any newly added code.

src/engine/image_palette.cpp
```diff
@@ -36,17 +36,21 @@ namespace
             std::copy_n( kb_pal, paletteSize, gamePalette.begin() );
         }
```
ihhub, line 37: To avoid any future wrongdoings by other developers could we please move this constructor into private section?

ihhub (changes requested):
Hi @dimag0g , the change looks good. I left 2 comments here. Can you take a look at them?

dimag0g:
@ihhub All done!

ihhub (approved):

dimag0g:
@ihhub Thanks for your fixes. Is there anything else for me to do here?

ihhub:
@dimag0g , my apologies. I got busy and had no time to merge it. Thanks a lot for fixing this issue!

PR #4964, opened 2022-01-30
Fix a few recently detected code smells

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thank you so much for code cleanup!

PR #4965, opened 2022-01-30
Migrate seeds from size_t to uint32_t

oleg-derevenetz:
Currently there is some inconsistency: most of the `namespace Rand` use `uint32_t` seeds, while `DeterministicRandomGenerator` and `hashCombine()` use `size_t` seeds (I suppose that's because `std::hash<...>::operator()` returns `size_t`). This leads to numerous `static_casts` back and forth. I suggest using `uint32_t` everywhere.

Pros of `size_t`:

* ???

Cons of `size_t`:

* `size_t` is implementation-defined (its size can vary from 16 bits to infinity), so we can't reliably save it to the savefiles, we should convert it to some fixed type anyway;
* Most of the existing `Rand` library uses `uint32_t`.


ihhub:

PR #4967, opened 2022-01-31
Implement a heuristic for hiring heroes that does not allow to offer the same heroes for hire in several kingdoms at the same time, if there are enough free heroes

### LINKED ISSUE #3662: Castle screen, heroes in tavern could change if AI player hires them in its turn

oleg-derevenetz:
fix #3662

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -652,6 +652,7 @@ bool Heroes::Recruit( const int col, const fheroes2::Point & pt )
         return false;
     }
 
+    ResetModes( RECRUIT );
     ResetModes( JAIL );
 
     SetColor( col );
@@ -1924,6 +1925,19 @@ Heroes * AllHeroes::GetFreeman( const int race, const int heroIDToIgnore ) const
         return nullptr;
     }
 
+    // Try to get a list of freeman heroes who are not yet available for recruitment in any kingdom
+    std::vector<int> freemanHeroesNotRecruits = freeman_heroes;
+
+    freemanHeroesNotRecruits.erase( std::remove_if( freemanHeroesNotRecruits.begin(), freemanHeroesNotRecruits.end(),
+                                                    [this]( const int heroID ) { return at( heroID )->Modes( Heroes::RECRUIT ); } ),
+                                    freemanHeroesNotRecruits.end() );
+
+    if ( !freemanHeroesNotRecruits.empty() ) {
+        return at( Rand::Get( freemanHeroesNotRecruits ) );
+    }
+
+    // There is no freeman heroes who are not yet available for recruitment, allow
+    // heroes to be available for recruitment in several kingdoms at the same time
     return at( Rand::Get( freeman_heroes ) );
 }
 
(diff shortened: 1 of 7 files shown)


src/fheroes2/heroes/heroes.cpp
```diff
@@ -1449,34 +1450,40 @@ bool Heroes::isFreeman( void ) const
 void Heroes::SetFreeman( int reason )
 {
     if ( !isFreeman() ) {
-        bool savepoints = false;
```
ihhub, line 1452: I assume this method just has code cleanup, doesn't it?

src/fheroes2/heroes/heroes_recruits.cpp
```diff
@@ -28,6 +28,37 @@
...
+{}
+
+Recruit::Recruit( const Heroes & hero )
+    : Recruit( hero, 0 )
+{}
+
+int Recruit::getID() const
+{
+    return _id;
+}
```
ihhub, line 48: Since it's a new method we can make it inlined.

src/fheroes2/heroes/heroes_recruits.cpp
```diff
@@ -28,6 +28,37 @@
...
+
+uint32_t Recruit::getSurrenderDay() const
+{
+    return _surrenderDay;
+}
```
ihhub, line 53: This method can be inlined as well.

src/fheroes2/world/world.cpp
```diff
@@ -23,6 +23,7 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
+#include <set>
```
ihhub, line 26: I think we can remove this inclusion since it's included in `kingdom.h` file.

ihhub (commented):
Hi @oleg-derevenetz , I left few very minor comments. Could you please check them?

src/fheroes2/heroes/heroes.cpp
oleg-derevenetz: @ihhub Not quite. Since the `Kingdom::appendSurrenderedHero()` now modifies hero's flags, its call should be moved later, after these flags are cleared in this method. Code cleanup was made along with this change :)

src/fheroes2/world/world.cpp
oleg-derevenetz: @ihhub I usually include the first-class include files (such as stdlib-related files) in the file where corresponding types or templates are used (EXCEPT when they are used in `module.cpp` and this file was already included in the `module.h`) to avoid dependencies on other modules in this regard. But if you insist I can remove this inclusion.

ihhub: We can keep it as it is.

ihhub:
@oleg-derevenetz , excellent work! Thank you very much for fixing it!

PR #4969, opened 2022-02-01
Fix the crash on an invalid assertion when the "Army order" setting is turned on and someone casts Berserker spell

oleg-derevenetz:
I chose black for Berserkers, but it's a matter of taste.


ihhub:
@oleg-derevenetz , thanks for fixing it!

PR #4970, opened 2022-02-01
Unit with ALL_ADJACENT_CELL_MELEE_ATTACK ability (e.g. Hydra) should be able to attack adjacent units from his own army if he is affected by Berserk or Hypnotize spells

oleg-derevenetz:
Currently his "splash" melee attack always targets only units from the opposing army.

Also, when choosing a cell from which the unit affected by Berserk spell will attack the target unit, always try to choose the nearest one (according to the AI pathfinder).

P.S. What's funny, in this situation:

Hydra Berserk AI Pathfinder

AI pathfinder thinks that Hydra (under Berserk spell) can move to the cell 34, and then happens this:

```
22:42:13: [DBG_BATTLE]	berserkTurn:  Hydra is under Berserk spell, moving to 34
22:42:13: [DBG_BATTLE]	berserkTurn:  Nearest reachable cell is -1
22:42:13: [DBG_BATTLE]	ApplyActionMove:  incorrect param: uid: 3, dst: -1
22:42:13: [DBG_BATTLE]	ApplyActionEnd:  uid: 3 moved
```

because cell 34 is the tail cell of Red Dragons, and there is absolutely no possibility for Hydra to move to this cell. But that's a relatively minor issue.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -718,11 +718,10 @@ namespace AI
                 }
                 else {
                     int targetCell = -1;
-                    const Indexes & around = Board::GetAroundIndexes( *targetUnit );
-                    for ( const int cell : around ) {
-                        if ( arena.hexIsPassable( cell ) ) {
+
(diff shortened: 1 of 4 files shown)


a1exsh:
Ah, good catch!

ihhub:

PR #4971, opened 2022-02-02
Allow a short name for all campaign bonus artifacts and spells

a1exsh:
This is needed to support translations, such as Russian and German.

### DIFF
--- a/src/fheroes2/campaign/campaign_scenariodata.cpp
+++ b/src/fheroes2/campaign/campaign_scenariodata.cpp
@@ -391,10 +391,10 @@ namespace
             return _( "short|Resurrect" );
         case Spell::STEELSKIN:
             return _( "short|Steel Skin" );
-        case Spell::VIEWHEROES:
-            return _( "short|View Heroes" );
         case Spell::SUMMONEELEMENT:
             return _( "Summon Earth" );
+        case Spell::VIEWHEROES:
+            return _( "short|View Heroes" );
         default:
             return Spell( spellId ).GetName();
         }
--- a/src/fheroes2/campaign/campaign_scenariodata.cpp
+++ b/src/fheroes2/campaign/campaign_scenariodata.cpp
@@ -324,49 +324,49 @@ namespace
     {
         switch ( artifactId ) {
         case Artifact::BALLISTA:
-            return _( "Ballista" );
+            return _( "campaignBonus|Ballista" );
         case Artifact::BLACK_PEARL:
-            return _( "short|Black Pearl" );
+            return _( "campaignBonus|Black Pearl" );
         case Artifact::CASTER_BRACELET:
-            return _( "Caster's Bracelet" );
+            return _( "campaignBonus|Caster's Bracelet" );
         case Artifact::DEFENDER_HELM:
-            return _( "Defender Helm" );
+            return _( "campaignBonus|Defender Helm" );
         case Artifact::DIVINE_BREASTPLATE:
-            return _( "Breastplate" );
+            return _( "campaignBonus|Breastplate" );
         case Artifact::DRAGON_SWORD:
-            return _( "Dragon Sword" );
+            return _( "campaignBonus|Dragon Sword" );
         case Artifact::FIZBIN_MISFORTUNE:
-            return _( "Fizbin Medal" );
+            return _( "campaignBonus|Fizbin Medal" );
         case Artifact::FOREMOST_SCROLL:
-            return _( "Foremost Scroll" );
+            return _( "campaignBonus|Foremost Scroll" );
         case Artifact::HIDEOUS_MASK:
-            return _( "short|Hideous Mask" );
+            return _( "campaignBonus|Hideous Mask" );
         case Artifact::MAGE_RING:
-            return _( "Mage's Ring" );
+            return _( "campaignBonus|Mage's Ring" );
         case Artifact::MAJOR_SCROLL:
-            return _( "Major Scroll" );
+            return _( "campaignBonus|Major Scroll" );
         case Artifact::MEDAL_HONOR:
-            return _( "short|Medal of Honor" );
+            return _( "campaignBonus|Medal of Honor" );
         case Artifact::MEDAL_VALOR:
-            return _( "short|Medal of Valor" );
+            return _( "campaignBonus|Medal of Valor" );
         case Artifact::MINOR_SCROLL:
-            return _( "Minor Scroll" );
+            return _( "campaignBonus|Minor Scroll" );
         case Artifact::NOMAD_BOOTS_MOBILITY:
-            return _( "Nomad Boots" );
+            return _( "campaignBonus|Nomad Boots" );
         case Artifact::POWER_AXE:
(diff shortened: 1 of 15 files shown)


a1exsh:
Ah, we will need to do the same to `getSpellCampaignName`...

src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -325,6 +325,8 @@ namespace
         switch ( artifactId ) {
         case Artifact::BALLISTA:
             return _( "Ballista" );
+        case Artifact::BLACK_PEARL:
```
ihhub, line 328: Could you please explain why do we have some items with prefix name like `short` and not others?

src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -361,8 +375,26 @@ namespace
...
+            return _( "short|Mass Curse" );
+        case Spell::MASSHASTE:
+            return _( "short|Mass Haste" );
+        case Spell::MIRRORIMAGE:
+            return _( "short|Mirror Image" );
+        case Spell::RESURRECT:
+            return _( "short|Resurrect" );
+        case Spell::STEELSKIN:
+            return _( "short|Steel Skin" );
         case Spell::SUMMONEELEMENT:
```
ihhub, line 394: The same question here: why do we have some items with `short` while others don't? We must be consistent.

ihhub (changes requested):
Hi @a1exsh , I left a question here. Could you please take a look? Also since we're modifying translation entries we have to modify them in pot files.

src/fheroes2/campaign/campaign_scenariodata.cpp
ihhub: `short` is not very informative name. Can we use `campaignArtifact|`?

a1exsh: > `short` is not very informative name. Can we use `campaignArtifact|`?

Yes, I would actually use `campaignBonus|`, as it should also apply to spells and other bonuses.

src/fheroes2/campaign/campaign_scenariodata.cpp
a1exsh: Initially, only the items that didn't fit in the untranslated English form were put here.  As such, they had to be different from the full form, and didn't need the distinguishing prefix.  But actually, the prefix doesn't matter for English, as far as I understand, as it's simply discarded at run time, correct?

a1exsh:
> Also since we're modifying translation entries we have to modify them in pot files.

@ihhub here I'd appreciate a hint on how to do that.  I've run `make` in `files/lang` and now I've got every `.po` file modified.  Do I have to commit these changes and add to this PR?  Do I have to add translations for every language or ask respective translators to provide updates?

ihhub:
> > Also since we're modifying translation entries we have to modify them in pot files.
> 
> @ihhub here I'd appreciate a hint on how to do that. I've run `make` in `files/lang` and now I've got every `.po` file modified. Do I have to commit these changes and add to this PR? Do I have to add translations for every language or ask respective translators to provide updates?

My apologies for a late reply. I am very busy in recent weeks.
We need to modify translation files by replacing old entries in them without touching translation string. For example:
```
msgid "Breastplate"
msgstr "Napierśnik"
```
should be changed into:
```
msgid "campaignBonus|Breastplate"
msgstr "Napierśnik"
```
and that's it. No need to ask translator to do the work as we're touching only English version of the text.

As a conclusion, you have to commit changes in po files to keep translation up to date.

src/fheroes2/campaign/campaign_scenariodata.cpp
ihhub: Yes, You're absolutely right. The prefix will be discarded upon rendering the text on a screen.

a1exsh:
@ihhub no worries at all, but please also bear with me :)

I think I must be doing something wrong.  I've fetched the latest changes from the `master` branch and merged into this PR's branch locally.  Then I did a clean build and got the `fheroes2.pot` file updated with all the new translatable strings.  Finally I ran `make` in the `files/lang` directory and got these much changes:
```
14 files changed, 711 insertions(+), 306 deletions(-)
files/lang/cs.po |  49 +++++++++-----
files/lang/de.po |  62 +++++++++++------
files/lang/es.po |  66 ++++++++++++------
files/lang/fr.po |  58 +++++++++++-----
files/lang/hu.po |  61 +++++++++++------
files/lang/it.po |  53 ++++++++++-----
files/lang/lt.po |  54 ++++++++++-----
files/lang/nb.po |  58 +++++++++++-----
files/lang/nl.po |  48 +++++++++-----
files/lang/pl.po | 199 +++++++++++++++++++++++++++++++++++++------------------
files/lang/pt.po |  52 ++++++++++-----
files/lang/ru.po | 145 ++++++++++++++++++++++++++++++----------
files/lang/sv.po |  64 ++++++++++++------
files/lang/tr.po |  48 +++++++++-----
```

If we instead modify the `.po` files manually, don't we have to keep the old `msgid` together with the prefixed one, as we have references from the code to both forms?

ihhub:
> 

Let's modify it automatically instead. poedit sorts strings so I won't be surprised if we have so many changes.

a1exsh:
@ihhub I think finally got around it.  I've used my knowledge of Russian and German (and very rudimentary of French and Polish) to update the `.po` files after re-generating them — the rest are just automatically generated.

Please have a look and let me know how we should proceed.

ihhub:

a1exsh:
@ihhub hm, I'm not sure what's wrong here, but it looks like the context is only properly handled in case there's no space in the translatable strings.  Otherwise I see the untranslated one, e.g:
image

Similar happens for "Defender Helm", for example.  At the same time "Mass Curse" it seems that "campaignBonus|Mass Curse" first (somehow internally?) becomes "Mass Curse", for which there is a translation, which is picked up, but it's not the one that we provide for this special context of campaign...

I've checked the `gettext` manual and I see that there is a separate function to work with contexts: any reason we are not using this approach?

oleg-derevenetz:
Hi @a1exsh

> hm, I'm not sure what's wrong here

It seems that you are just running an old binary with new translations.

a1exsh:
> 
> It seems that you are just running an old binary with new translations.

Damn, you are right — I forgot to rebuild after merging from master. :facepalm: 
Seems to work as expected now.

> > I've checked the gettext manual and I see that there is a separate function to work with contexts: any reason we are not using this approach?
> 
> fheroes2 doesn't use `gettext`.

I mean, we are still using the `msgfmt` facility and the file formats of `gettext`, so it'd make sense to use the supported way of encoding translation context, even if we are not linking with the `gettext` lib, no?

oleg-derevenetz:
> I mean, we are still using the msgfmt facility and the file formats of gettext, so it'd make sense to use the supported way of encoding translation context, even if we are not linking with the gettext lib, no?

I believe that the task of 100% replicating gettext in all respects is simply not on the agenda :)

PR #4972, opened 2022-02-02
Rework the Berserker's logic, bringing it closer to the logic of a regular AI-controlled unit

oleg-derevenetz:
Fix the issue described in #4970

Before (note that the Hydra affected by the Berserk spell skips a turn):

https://user-images.githubusercontent.com/32623900/152168694-f1a71c38-036a-4c0a-9c0a-1f4714576d1f.mp4

After:

https://user-images.githubusercontent.com/32623900/152168713-421d7bcb-39ff-41c2-8738-d4020cf60b58.mp4

---

Also fixed a bug that caused the castle towers not to target berserkers.


ihhub:
@oleg-derevenetz , great job! Well done!

PR #4975, opened 2022-02-02
Don't use ESC key as a retreat key during battle, because it conflicts with the hotkey that closes the hero dialog

oleg-derevenetz:
That's what happens when I press the `ESC` key to close the hero dialog:

https://user-images.githubusercontent.com/32623900/152235272-91fb71b0-5b6b-47f3-8990-5f4e230de838.mp4

I propose to use the `KEY_r` instead by default.

### DIFF
--- a/fheroes2.key
+++ b/fheroes2.key
@@ -70,7 +70,7 @@
 #
 # 2.5 battle events:
 # battle cast spell	- (default keycode:  99 = 'c')
-# battle retreat	- (default keycode:  27 = 'esc')
+# battle retreat	- (default keycode: 114 = 'r')
 # battle surrender	- (default keycode: 115 = 's')
 # battle auto switch	- (default keycode:  97 = 'a')
 # battle options	- (default keycode: 111 = 'o')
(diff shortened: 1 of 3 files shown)


a1exsh:
@oleg-derevenetz isn't the `ESC` key just leaked from the hero options dialog here?

oleg-derevenetz:
@a1exsh No.

a1exsh:
But it looks exactly like the case for leaking.  If you click the close button, it just closes the hero options dialog.  If you press `ESC`, the dialog is closed and next you are asked if you want to retreat.

I wouldn't like to change the behavior of `ESC` when during the battle and no hero options dialog is open: it should still prompt for a retreat like it does now.

oleg-derevenetz:
@a1exsh It has nothing to do with leaking. The problem is that here:

https://github.com/ihhub/fheroes2/blob/bb5e9a2d88cb7a299fe68f0adc80236121659038/src/fheroes2/battle/battle_dialogs.cpp#L857-L858

`result` absolutely legitimate becomes 2. The real problem is that `ESC` is used as a "retreat" key, and it is used in this sense in the hero dialog as well, while most users (including me) expect that `ESC` will just dismiss this dialog.

ihhub:
@oleg-derevenetz , thanks a lot for this code change!

PR #4976, opened 2022-02-02
Fixes with lowercase conversion of translations

dimag0g:
This PR addresses three somewhat related problems regarding `StringLower` applied to translated strings, e.g. "Archers" -> "A few archers":

1. In German, nouns are always uppercase. There is no way to implement this feature via translation files, `StringLower` has to be aware of the language.
2. `StringLower` uses `std::tolower` which by default only works with ASCII. Letters from CP12XX codepages will remain uppercase, e.g. "Mało Łucznicy" in Polish or "Мало Стрелки" in Russian.
3. The variable `locale` in `translations.cpp` is only updated when the language is first loaded (via `bindDomain`), it will not change after successive calls. Changing from "Language 1" to "Language 2" and then back to "Language 1" in the game will result in some locale-specific rules from "Language 2" to be applied to "Language 1". This currently mostly concerns plural forms, but after (1) and (2) were implemented, it affected lowercase rules as well.

The problems are resolves as follows:

1. In addition to `StringLower` an additional `Translations::StringLower` function is implemented, which is only applied to translated in-game strings like monster names.
2. `Translations::StringLower` uses a lookup table to convert strings when the current encoding is set to `CP1250` or `CP1251`. There are a few conflicting characters which should be lowercased differently in two encodings (notably, `ß` which is used in German and thus never lower-cased, and `Ј` and `Ґ` characters which are currently not used by any of the translations).
3. The `locale` variable is made part of the `mofile` structure, and `bindDomain` sets it correctly for the current translation. Incidentally, `setDomain` function became useless and was removed. The game always used to call `bindDomain && setDomain` together anyway.

### DIFF
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -20,6 +20,7 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include <array>
 #include <list>
 #include <map>
 #include <string>
@@ -69,8 +70,10 @@ struct mofile
     uint32_t offset_strings2;
     uint32_t hash_size;
     uint32_t hash_offset;
+    int locale;
     StreamBuf buf;
     std::map<u32, chunk> hash_offsets;
+    std::string domain;
     std::string encoding;
     std::string plural_forms;
     u32 nplurals;
@@ -236,7 +239,6 @@ namespace Translation
 
     mofile * current = nullptr;
     std::map<std::string, mofile> domains;
-    int locale = LOCALE_EN;
     char context = 0;
 
     void setStripContext( char strip )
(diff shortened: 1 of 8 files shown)


src/engine/translations.cpp
```diff
@@ -413,4 +417,40 @@ namespace Translation
...
+            // Linux needs the language and the country code as well
+            // Country codes are usually language codes capitalized, except for "en"
+            if ( ( current->domain[0] == 'e' ) && ( current->domain[1] == 'n' ) )
+                localeName = std::string() + "en_US." + current->encoding;
+            else
+                localeName = std::string() + current->domain + "_" + StringUpper( current->domain ) + "." + current->encoding;
+#endif
+
+#if defined( ANDROID ) || defined( __APPLE__ )
+            setlocale( LC_CTYPE, localeName.data() );
```
oleg-derevenetz, line 444: Hi @dimag0g AFAIK `setlocale()` is able to return the current locale if second parameter is `nullptr`. In theory, it would be possible to save the previous locale and then reset it back to the previous one instead of "C".

dimag0g: @oleg-derevenetz True, but for now no other locale than "C" is used, so resetting to "C" will do. Besides, I wonder whether we should use `setlocale` at all. Windows comes with hundreds of pre-configured locales, but in Linux a German-speaking user might just configure `de_DE.UTF-8` and switching to `de_DE.CP1252` will fail. And that's a pretty typical case, think about someone from Austira/Belgium with `de_AT`/`de_BE`, or (God forbid!) someone with a Nintendo Switch with whatever is supported there. It's just not reliable, and I think we should settle for something like [ICU](https://icu.unicode.org/) or maybe just put our own lookup table.

oleg-derevenetz: Yeah, that's true, desired locale might just not be installed.

src/engine/tools.cpp
```diff
@@ -66,6 +66,13 @@ std::string StringLower( std::string str )
     return str;
 }
 
+/* convert to upper case */
+std::string StringUpper( std::string str )
```
a1exsh, line 70:
```suggestion
std::string StringUpper( const std::string & str )
```
Looking around, other similar functions could also use this signature.

src/fheroes2/army/army.cpp
```diff
@@ -120,7 +120,7 @@ std::string Army::TroopSizeString( const Troop & troop )
         break;
     }
 
-    StringReplace( str, "%{monster}", StringLower( troop.GetMultiName() ) );
+    StringReplace( str, "%{monster}", Translation::StringLower( troop.GetMultiName() ) );
```
a1exsh, line 123: Instead of introducing a twist here, wouldn't it be simpler to just throw away lowercase conversion altogether?  What good does it do to have it?

src/engine/translations.cpp
```diff
@@ -413,4 +417,40 @@ namespace Translation
...
+            std::string result = ::StringLower( str );
+            setlocale( LC_CTYPE, "C" );
+#else
+            std::setlocale( LC_CTYPE, localeName.data() );
+            std::string result = ::StringLower( str );
+            std::setlocale( LC_CTYPE, "C" );
```
a1exsh, line 450: This looks really fishy.  I'd rather not do that.  Esp. see the note https://en.cppreference.com/w/cpp/locale/setlocale:

`Because setlocale modifies global state which affects execution of locale-dependent functions, it is undefined behavior to call it from one thread, while another thread is executing any of the following functions: [std::fprintf](https://en.cppreference.com/w/cpp/io/c/fprintf), ...`

src/engine/translations.cpp
oleg-derevenetz: I tend to the variant with our own table within the engine.

src/fheroes2/army/army.cpp
dimag0g: Do you mean we should just keep uppercase letters, as in "guarded by A pack of Ghosts"?

src/engine/translations.cpp
dimag0g: I agree. The only question is, what do we use as a replacement? We could use a proper encoding library, or perhaps just put in a bunch of lookup tables. In fact, dealing with non standard encodings (e.g. French, where `lowercase('$') == '#'`) can only be done with a custom LUT.

src/fheroes2/army/army.cpp
a1exsh: > Do you mean we should just keep uppercase letters, as in "guarded by A pack of Ghosts"?

Well, that's not ideal either.  I was thinking more of a case like "Move Archers here": it's still OK with the uppercase.

Maybe we need to introduce special translations for "in the middle of a sentence", because proper spelling of it is language-dependent?  E.g. in German, it would help with the case of adjective + noun.  Consider: "Grüne Drachen kaufen" vs. "Wollen Sie grüne Drachen kaufen?"

I've never worked with `gettext` before, btw — maybe it already has facilities to support something like that?

src/engine/translations.cpp
a1exsh: Don't get me wrong: I'm not against use of `setlocale()` as such, but I think it should be done only at startup or when changing the game language, in a controlled environment that guarantees no concurrent calls to the affected functions.

What I'm objecting is switching locale forth and back for the purpose of a single lowercase call.

ihhub:
I don't know why GitHub marked my previous comment as pending so I'm posting here:
I agree with @oleg-derevenetz about the usage of tables for these cases. We don't have many languages to support so creation of such tables won't be a difficult task.

src/engine/translations.cpp
ihhub: LUT might be the solution for `toLower` and `toUpper` calls for well-used languages. For some unknown languages which aren't supported yet we can fallback for `setlocale` solution.

oleg-derevenetz: We can just ignore unsupported languages (i.e. just use `std::tolower()` and `std::toupper()` without calling `setlocale()` in case of some unsupported language). Modification of global state for such a local use is really doesn't look good.

dimag0g:
@ihhub @oleg-derevenetz 

I ended up making a single lookup table for both CP1250 and 1251. There are a few clashes (as noted in the comments) but none of the conflicting characters are used in translations (expcept for ß, which is used in German, but lowercasing German strings is a no-op, so it doesn't matter). If it becomes necessary, we could make one lookup table per encoding.

Could someone enlighten me how to format the initializer list in a clang-compliant way without putting each value on a new line?

oleg-derevenetz:
@dimag0g

> Could someone enlighten me how to format the initializer list in a clang-compliant way without putting each value on a new line?

`monsterIcnIds` and `binFileName` in the [monster/monster_info.cpp](https://github.com/ihhub/fheroes2/blob/3dae29e127b6a31aa3aacd6b322f934ce3b78bfc/src/fheroes2/monster/monster_info.cpp#L43) look like the right thing.

src/fheroes2/army/army.cpp
dimag0g: @a1exsh As far as I know, there is no such support. `gettext` manual keeps stressing out on every page that you should always use it with complete sentences.

dimag0g:
I'd be grateful if people fluent in Polish/German/Russian could test this.

src/engine/translations.cpp
```diff
@@ -256,87 +258,86 @@ namespace Translation
...
-        else if ( str == "sv" || str == "swedish" )
-            locale = LOCALE_SV;
-        else if ( str == "tr" || str == "turkish" )
-            locale = LOCALE_TR;
-
-        return domains[domain].open( file );
-    }
+        // Search for already loaded domain or load from file
+        std::map<std::string, mofile>::iterator it = domains.find( str );
+        if ( it == domains.end() ) {
```
ihhub, line 265: To reduce changes in this pull request as well as the complexity of the function I am proposing to do the following way:
```cpp
if ( it != domains.end() ) {
    current = &( *it ).second;
    return true;
}
...
// the rest of the code.
```

src/engine/translations.cpp
```diff
@@ -413,4 +414,82 @@ namespace Translation
 
         return stripContext( n == 1 ? str : plural );
     }
+
+    // Character lookup table for custom tolower
+    // Compatible with ASCII, custom French encoding, CP1250 and CP1251
+    const std::array<unsigned char, 256> tolowerLUT
```
ihhub, line 420: Please put this table into anonymous namespace at the top of this file to reduce the scope and keep global (to the file) variables in one place.
```cpp
namespace
{
    const std::array<unsigned char, 256> tolowerLUT = { ... };
}
```

src/engine/translations.cpp
```diff
@@ -413,4 +414,82 @@ namespace Translation
...
+
+    std::string StringLower( std::string str )
+    {
+        if ( current ) {
+            // With German, lowercasing strings does more harm than good
+            if ( current->locale == LOCALE_DE )
+                return str;
+
+            // For CP1250/1251 codepages a custom lowercase LUT is implemented
+            if ( ( current->encoding == "CP1250" ) || ( current->encoding == "CP1251" ) ) {
```
ihhub, line 488: Here we specify that we apply the LUT for C1250 and CP1251 encodings while the LUT covers French language as well.

src/engine/translations.cpp
```diff
@@ -413,4 +414,82 @@ namespace Translation
 
         return stripContext( n == 1 ? str : plural );
     }
+
+    // Character lookup table for custom tolower
+    // Compatible with ASCII, custom French encoding, CP1250 and CP1251
+    const std::array<unsigned char, 256> tolowerLUT
+        = { 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F, // Non-printable characters
```
ihhub, line 421: I can see that CP1252 which is used for many European languages can be covered by this LUT as well. Should we add the support of it also?

ihhub (changes requested):
Hi @dimag0g , I left several comments in this pull request. Would you mind to take a look at them and see if they make sense?

src/engine/translations.cpp
dimag0g: Currently CP1252 is only used for German, for which no case change is done and the LUT is not used. It's indeed quite similar and could be supported. Nevertheless, I don't think it makes sense to support additional encodings which are not actually used, it brings maintenance cost and adds no value.

src/engine/translations.cpp
dimag0g: French is actually adequately covered by the original `std::tolower`: all special characters used as French letters are lowercase so they never change. This is why the original `::StringLower` is called for French instead of using the LUT. A LUT would be required if we ever need a `::StringUpper`, which would have to transform `@`, corresponding to the French `à`, into `A`.

If you like, I can add an extra clause `( current->locale == LOCALE_FR )`, to make sure the LUT is used for French. This change will have no functional effect though.

src/engine/translations.cpp
dimag0g: I'll put all local-only stuff into that namespace then, not just the LUT.

ihhub (approved):

ihhub:
@dimag0g, excellent work! Thanks a lot for these changes!

PR #4981, opened 2022-02-04
Change the way spell casts are logged

oleg-derevenetz:
Unifies the spell logging format for all spells. Previously, a number of spells (for example, Summon Elemental, Teleport or Earthquake) were not logged at all.

Also fixes the following issues:

* Earthquake may "heal" castle walls (pay attention to the condition of the wall at the bottom right):

https://user-images.githubusercontent.com/32623900/152602691-8ac7968a-1c79-4c28-915b-b725b9619f2f.mp4

* Spells are incorrectly logged if they are applied to the tail cell of a wide unit:

https://user-images.githubusercontent.com/32623900/152602941-850cd40d-c214-4264-b759-27e77054893c.mp4

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -1059,17 +1059,34 @@ void Battle::Arena::ApplyActionSpellMirrorImage( Command & cmd )
         if ( it != distances.end() ) {
             const Position pos = Position::GetPositionWhenMoved( *troop, *it );
             const s32 dst = pos.GetHead()->GetIndex();
+
             DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "set position: " << dst );
-            if ( interface )
+
+            if ( interface ) {
+                const HeroBase * commander = GetCurrentCommander();
+                assert( commander != nullptr );
+
+                TargetInfo targetInfo;
+                targetInfo.defender = troop;
+
+                TargetsInfo targetsInfo;
+                targetsInfo.push_back( targetInfo );
+
+                interface->RedrawActionSpellCastStatus( Spell( Spell::MIRRORIMAGE ), who, commander->GetName(), targetsInfo );
                 interface->RedrawActionMirrorImageSpell( *troop, pos );
+            }
+
             Unit * mirror = CreateMirrorImage( *troop, dst );
-            if ( mirror )
+            if ( mirror ) {
                 mirror->SetPosition( pos );
+            }
         }
         else {
-            if ( interface )
-                interface->SetStatus( _( "spell failed!" ), true );
             DEBUG_LOG( DBG_BATTLE, DBG_WARN, "new position not found!" );
+
+            if ( interface ) {
+                interface->SetStatus( _( "Spell failed!" ), true );
+            }
         }
     }
     else {
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -1017,13 +1017,17 @@ void Battle::Arena::ApplyActionSpellTeleport( Command & cmd )
 
 void Battle::Arena::ApplyActionSpellEarthQuake( const Command & /*cmd*/ )
 {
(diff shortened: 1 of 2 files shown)


src/fheroes2/battle/battle_action.cpp
```diff
@@ -986,8 +989,19 @@ void Battle::Arena::ApplyActionSpellTeleport( Command & cmd )
...
-        if ( interface )
+        if ( interface ) {
+            const HeroBase * commander = GetCurrentCommander();
+            assert( commander != nullptr );
+
+            TargetInfo targetInfo;
+            targetInfo.defender = b;
+
+            TargetsInfo targetsInfo;
+            targetsInfo.push_back( targetInfo );
```
ihhub, line 1000: Could we please use `emplace_back( std::move( targetInfo ) );` to save some CPU ticks? :)

src/fheroes2/battle/battle_action.cpp
```diff
@@ -1041,17 +1067,34 @@ void Battle::Arena::ApplyActionSpellMirrorImage( Command & cmd )
...
+
+            if ( interface ) {
+                const HeroBase * commander = GetCurrentCommander();
+                assert( commander != nullptr );
+
+                TargetInfo targetInfo;
+                targetInfo.defender = troop;
+
+                TargetsInfo targetsInfo;
+                targetsInfo.push_back( targetInfo );
```
ihhub, line 1081: The same comment here as well: please use `emplace_back( std::move( targetInfo ) );` combination.

ihhub (changes requested):
Hi @oleg-derevenetz , I left just 2 small optimization related comments. Would you mind to take a look?

src/fheroes2/battle/battle_action.cpp
```diff
@@ -1001,17 +1015,29 @@ void Battle::Arena::ApplyActionSpellTeleport( Command & cmd )
...
     std::vector<int> targets = GetCastleTargets();
+
     if ( interface ) {
+        interface->RedrawActionSpellCastStatus( Spell( Spell::EARTHQUAKE ), -1, commander->GetName(), {} );
         interface->RedrawActionEarthQuakeSpell( targets );
     }
 
-    const HeroBase * commander = GetCurrentCommander();
-    const std::pair<int, int> range = commander ? getEarthquakeDamageRange( commander ) : std::make_pair( 0, 0 );
+    const std::pair<int, int> range = getEarthquakeDamageRange( commander );
```
ihhub, line 1028: Can we please modify `getEarthquakeDamageRange` function to return `std::pair<uint32_t, uint32_t>` pair? In this case we won't need any casting. Also right now we implicitly do the casting at line 1034.

src/fheroes2/battle/battle_action.cpp
oleg-derevenetz: @ihhub This would make sense if this would be some container on the heap - instead of copying the container's contents, we would just transfer the ownership of the contents to the new container. However, this is just struct that contains int and pointer, and move constructor would just copy them anyway to the new object, just like the copy constructor, because there is no other options. I'm not sure that there will be any gain :) I'll take a look tomorrow on an assembly output of both variants, I suppose that they will be identical.

Such tricks may be justified in generic code with templates, because we don't know in advance what kind of object we will move. But in this case... I'm not sure :)

ihhub: Fair point. Let's leave it as it is then.

src/fheroes2/battle/battle_action.cpp
oleg-derevenetz: @ihhub Yes, I just decided not to touch it, but I can rewrite it tomorrow.

ihhub: It's okay. I can do it myself since this is the only change is needed.

ihhub:
@oleg-derevenetz , thank you so much for these great changes!

PR #4982, opened 2022-02-04
Migrate clang analyzer to clang 12

oleg-derevenetz:


src/engine/image.cpp
```diff
@@ -1247,6 +1247,10 @@ namespace fheroes2
                 if ( *inX > 0 && *inX < 6 ) { // 1 is to skip, 2 - 5 types of shadows
                     if ( ( x > 0 && *( inX - 1 ) == 0 ) || ( x < reducedWidth && *( inX + 1 ) == 0 ) || ( isNotTopRow && *( inX - width ) == 0 )
                          || ( isNotBottomRow && *( inX + width ) == 0 ) ) {
+#ifdef __clang_analyzer__
+                        // FIXME: apparently, a false positive from the Clang Analyzer
+                        assert( outImageY != nullptr );
```
ihhub, line 1252: Actually SonarQube complains about these 2 places too.

ihhub: I'll take a look at these places. Something fishy is there.

oleg-derevenetz: @ihhub

> Actually SonarQube complains about these 2 places too.

That's because SonarQube uses clang analyzer too :)

ihhub: I have a feeling it generates due to `reset` method usage. What if we apply this trick there (line 1228)?

oleg-derevenetz: @ihhub I moved this assertion a bit earlier, let's see. But in fact it doesn't matter, because due to `#ifdef __clang_analyzer__` this assertion is visible only for `scan-build` and not for the "regular" compiler. There will be no real `assert()` call in the generated code.

ihhub: If we move the assertion out of the loop it's good to remove these if-def wrapper code. No harm to have one extra assertion  like we usually do.

oleg-derevenetz: It looks like it does help, let's see what happens if we move another assertion out of the loop.

ihhub:
It's a miracle. It works!

oleg-derevenetz:
@ihhub

> It's a miracle. It works!

We can try to check them for `!empty()` BTW.

ihhub:
@oleg-derevenetz , well done! Now we have no bugs from static analyzers!

PR #4984, opened 2022-02-05
Try to fix SonarQube and Clang found bugs

ihhub:
relates to #4982

### DIFF
--- a/.github/workflows/clang_analyzer.yml
+++ b/.github/workflows/clang_analyzer.yml
@@ -15,10 +15,10 @@ jobs:
     - name: Install dependencies and clang-tools
       run: |
         sudo apt-get -y update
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @oleg-derevenetz , does this trick fix one of clang false positives?

oleg-derevenetz:
@ihhub now we have just these 2 bugs left :)

ihhub:
> @ihhub now we have just these 2 bugs left :)

Nooooo! Absolutely unacceptable! Hehe

We still have quite a bunch of warning in Visual Studio (48 to be precise). I have a fix for some of them but not for all. The goal is to fix them all and then interpret all warnings as bugs leading to less error prone code.

PR #4987, opened 2022-02-05
Slightly refine the code

oleg-derevenetz:
* Rename variables in the code related to the order of units in battle - for better clarity (all those `part1`, `part2` looks a bit clunky);
* Move this code from `Battle::Force` to the anonymous namespace in the `battle_arena.cpp`, because it doesn't have any dependencies on `Battle::Force`, is not relevant to it and is needed only in one place;
* Fix code smells related to the implicit slicing of `Force` to `Units` (most of remaining "bugs" from SonarCloud);
* Fix code smell related to the implicit slicing of `Players` to `std::vector<Player *>` (another "bug" from SonarCloud);
* Fix typo in debugging output.


ihhub:
@oleg-derevenetz , thank you so much for the constant code cleanup!

PR #4991, opened 2022-02-07
Use localtime_r() or localtime_s() instead of localtime()

oleg-derevenetz:
At least in one place (`DEBUG_LOG` macro) `localtime()` could be called from the non-main thread (via `AGG::AsyncSoundManager`) - only being built with DEBUG, but nonetheless.


ihhub:
@oleg-derevenetz , thank you very much for this code improvement!

PR #4994, opened 2022-02-08
Do not count on a specific type of attacking unit when applying built-in dispel

oleg-derevenetz:
Because in the new ideology, we no longer rely on the type of unit itself, but on unit's capabilities.

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -140,19 +140,17 @@ void Battle::Arena::BattleProcess( Unit & attacker, Unit & defender, int32_t dst
 
             targets = GetTargetsForSpells( attacker.GetCommander(), spell, defender.GetHeadIndex() );
 
-            bool validSpell = true;
-            if ( attacker == Monster::ARCHMAGE && !defender.Modes( IS_GOOD_MAGIC ) )
-                validSpell = false;
-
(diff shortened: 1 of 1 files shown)


src/fheroes2/battle/battle_action.cpp
```diff
@@ -136,23 +136,21 @@ void Battle::Arena::BattleProcess( Unit & attacker, Unit & defender, int32_t dst
...
-            if ( attacker == Monster::ARCHMAGE && !defender.Modes( IS_GOOD_MAGIC ) )
-                validSpell = false;
-
-            if ( !targets.empty() && validSpell ) {
+            // The built-in dispel should only remove beneficial spells from the target
+            if ( !targets.empty() && ( spell.GetID() != Spell::DISPEL || defender.Modes( IS_GOOD_MAGIC ) ) ) {
                 if ( interface ) {
+                    const std::string name( attacker.GetName() );
+
                     interface->RedrawActionSpellCastStatus( spell, defender.GetHeadIndex(), name, targets );
```
ihhub, line 146: A tiny optimization: we can use `attacker.GetName()` directly within this call.

ihhub (commented):
Hi @oleg-derevenetz , I left just one tiny optimization comment. If you think it's useless then we can proceed with merging.

idshibanov (approved):

PR #4998, opened 2022-02-09
Heroes::GetRangeRouteDays(): remove the special logic related to the mechanics of the last step, because now pathfinder can handle it on its own

oleg-derevenetz:
Now this logic is incorrect.


idshibanov (approved):

ihhub:
@oleg-derevenetz , thank you so much for cleaning up the code!

PR #5000, opened 2022-02-10
Fix neutral monster growth formula

### LINKED ISSUE #4893: Question / bug on monster growth
### Preliminary checks

### Platform

Windows

### Describe the bug

cf https://github.com/ihhub/fheroes2/blob/84b4d57dd0aacec2f955cb962bf7506f47979e50/src/fheroes2/maps/maps_tiles_quantity.cpp#L1104

- Monster growth is set at 14% weekly (8/7)
- Due to integer division, the stack will never grow week after week if the initial stack size is < 7

My questions:
- Is this value of 14% weekly accurate ? Do you know what is the behavior in OG ? Is it random ? I know in HOMM3 it also depend on the difficulty level I think
- What should be the correct behavior when stack size is small ? We could instead increase the stack size by one each week randomly (14% of growing by 1 unit if only 1 unit is there, to 85% of chance of growing by 1 unit if 6 units are there)

### Save file

none

### Additional info

_No response_

idshibanov:
Fixes #4893 .

Currently if monster stack is smaller than 7 it won't grow at all. Adjusting the logic to follow the original.

### DIFF
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -1103,7 +1103,7 @@ void Maps::Tiles::UpdateMonsterPopulation( Tiles & tile )
         tile.MonsterSetCount( troop.GetRNDSize( false ) );
     }
     else if ( !tile.MonsterFixedCount() ) {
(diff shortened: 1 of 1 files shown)


ihhub:
@idshibanov , thanks a lot for fixing this bug!

PR #5002, opened 2022-02-10
Another round of removing unused code

oleg-derevenetz:
Also `MapPosition::isPosition()` was made virtual, because it is overridden in `Castle`.

### DIFF
--- a/src/engine/xmi2mid.cpp
+++ b/src/engine/xmi2mid.cpp
@@ -313,7 +313,7 @@ StreamBuf & operator<<( StreamBuf & sb, const MidiChunk & event )
     return sb;
 }
 
-struct MidiEvents : std::vector<MidiChunk>
+struct MidiEvents : public std::vector<MidiChunk>
 {
     uint32_t trackTempo = 0;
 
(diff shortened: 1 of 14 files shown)


src/engine/xmi2mid.cpp
```diff
@@ -433,9 +433,6 @@ struct MidTrack
     IFFChunkHeader mtrk;
     MidiEvents events;
 
-    MidTrack()
```
ihhub, line 436: Shall we then make this constructor as delete to avoid default constuctor creation by a compiler?

ihhub (commented):
Hi @oleg-derevenetz , I left one comment here as well. Please take a look if it makes sense.

src/engine/xmi2mid.cpp
oleg-derevenetz: @ihhub Default constructors are generated only if we don't have any user-defined constructors. Since we have at least one user-defined constructor here, default constructor [is not generated](https://godbolt.org/z/oPsqdGvh7).

ihhub:
@oleg-derevenetz , thank you so much for this cleanup!

PR #5005, opened 2022-02-12
Fix multiple compilation warnings

ihhub:
- reduce the amount of static_cast usage
- use correct variable types
- add extra checks to avoid future code breaks

### DIFF
--- a/src/engine/pal.cpp
+++ b/src/engine/pal.cpp
@@ -140,16 +140,21 @@ namespace
     }
 }
 
-std::vector<uint8_t> PAL::GetCyclingPalette( int stepId )
+std::vector<uint8_t> PAL::GetCyclingPalette( uint32_t stepId )
 {
     std::vector<uint8_t> palette = PAL::GetPalette( PaletteType::STANDARD );
 
-    const std::vector<CyclingColorSet> & set = GetCyclingColors();
-    for ( std::vector<CyclingColorSet>::const_iterator it = set.begin(); it != set.end(); ++it ) {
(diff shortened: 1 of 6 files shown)


src/engine/pal.cpp
```diff
@@ -140,16 +140,21 @@ namespace
...
+        for ( uint32_t id = 0; id < colorSet.length; ++id ) {
+            if ( colorSet.forward ) {
+                const uint32_t newColorID = colorSet.start + ( ( id + stepId ) % colorSet.length );
+                palette[colorSet.start + id] = static_cast<uint8_t>( newColorID );
+            }
+            else {
+                const uint32_t lastColorID = colorSet.length - 1;
+                const uint32_t newColorID = colorSet.start + lastColorID - ( ( lastColorID + stepId - id ) % colorSet.length );
+                palette[colorSet.start + id] = static_cast<uint8_t>( newColorID );
+            }
```
oleg-derevenetz, line 157: We can take declaration of the `newColorID` and palette update out of the `if` statement to avoid code duplication:

```cpp
uint32_t newColorID;

if ( colorSet.forward ) {
[...]
}
else {
[...]
}

palette[colorSet.start + id] = static_cast<uint8_t>( newColorID );
```

But it's a matter of taste.

oleg-derevenetz (changes requested):
Hi @ihhub I just left two recommendations here, please take a look when it's convenient for you.

idshibanov (approved):
Tentative approval. Good to merge once you address the comment above.

oleg-derevenetz (approved):

PR #5006, opened 2022-02-12
Add base support for text mode

ihhub:
This mode is extremely useful for people with visual impairment. A special software (out of the scope of this project) should be used to process the output of the console and interpret it into a speech.

This is a prototype code and could be changed over time based of a given feedback.


ihhub:
Hi @oleg-derevenetz , @idshibanov  and @dimag0g , could you please take a look and express your ideas about the proposed solution? I feel that the current implementation is a bit clunky and I miss something obvious to make it better.

dimag0g:
@ihhub This is really cool! Can't wait to feed this to orca and listen to the narrative while I play.

I noticed a somewhat unfortunate behavior of fheroes2 while working on translations: in some dialogs (castle, heroes, etc.) a status string is continuously re-rendered either all the time or while hovering certain elements, causing thousands of calls to `gettext`. I was thinking about optimizing these dialogs, but the performance gain was so small I dropped the idea. Now we might want to rework these dialogs not for the sake of performance, but to limit the amount of text output.

oleg-derevenetz:
To be honest, I do not know how to make this less clumsy. Since we have almost no "UI controls" in the game (in the sense of independent objects that handle events by themselves) it seems that we will not be able to make the console output "centralized". Also, although it is relatively clear how this will work in dialog windows and so on, I still don't quite understand how this expected to work in battle mode and on the adventure map. How can I help the player to focus on a certain tile? Should this mode print all information regarding a specific tile when the mouse is over that tile? What about the new objects on the adventure map that will appear when the fog of war is dispersed during the hero movement - will there a way to promptly report about them to let the player to make decisions regarding the next movements of the hero? Also on Windows the console window won't be with us forever (I suppose), so it's unclear to me where to output the text if (or when) this console is no longer with us. I'm not sure that all this can be done in a really convenient way for a player.

ihhub:
> @ihhub This is really cool! Can't wait to feed this to espeak and listen to the narrative while I play.
> 
> I noticed a somewhat unfortunate behavior of fheroes2 while working on translations: in some dialogs (castle, heroes, etc.) a status string is continuously re-rendered either all the time or while hovering certain elements, causing thousands of calls to `gettext` every minute. I was thinking about optimizing these dialogs, but the performance gain was so small I dropped the idea. Now we might want to rework these dialogs not for the sake of performance, but to limit the amount of text output.

We should avoid any cases like this. While performance wise it is not noticeable on desktops it can be visible on low-end devices like Raspberry Pi. So please go ahead and fix these places.

dimag0g:
@oleg-derevenetz On Windows we may want to output text to Microsoft Speech API rather than the console.

dimag0g:
@ihhub As it turns out, the optimization is already done in places such as `StatusBar::ShowMessage`, which only render the message if it's different from the previous one. It should handle text output as well.

The message still goes through `gettext` repeatedly, but optimizing this will be difficult to achieve: some compound strings (e.g `Dismiss %{name} the %{race}`) have to go through `gettext` *before* the substitution is done. It's still possible to optimize this for simple strings, but, like I said, the performance gain is indeed very small.

BlindGuyNW:
Hey all,

As one of (if not the only) blind players currently interested in this, I can say that the idea of output to, say, SAPI or ESpeak hadn't occurred to me. It would probably solve a number of problems if things could be done that way. There are a number of accessibility output solutions, such as [this one](https://github.com/accessibleapps/accessible_output2), which might help with doing this in a reasonably straight-forward way. I always favored [Tolk](https://github.com/dkager/tolk), but the end result is largely equivalent.

The main problem at the moment is that in order to read output from the console I need to move away from the game window because screen readers only really support getting the data from the active window by default. If we speak it ourselves, that sidesteps the issue.

Sorry to throw a wrench into things :)

ihhub:
Hi @BlindGuyNW , are you suggesting to move from text based solution to an integration with an audio output API, primary on Windows? This can be done but this pull request must be closed. I personally think that the support of an audio output API is a right thing to do.

Also #5054 pull request is aimed to standardize the dialogs in the game which means that the support for text mode / speaking API mode will be more broad for the end users.

@oleg-derevenetz and @dimag0g , I would like to hear your opinions about this topic.

dimag0g:
I personally think that console output is the right thing to do as a first step: it will be useful for debugging/testing. Later on, we could implement an option to redirect the text output to `speech-dispatcher` on Debian or `SAPI` on Windows, and that's what the end users will use, but console output option is useful.

ihhub:
Hi @dimag0g and @oleg-derevenetz , I modified the logic to make the code faster and easier to read. I think this is a better solution. Could you please take a look again?

oleg-derevenetz (approved):

PR #5011, opened 2022-02-13
Allow AI fog discovery only for teleport endpoints with reachability information

### LINKED ISSUE #5008: Freeze during AI turn
### Preliminary checks

### Platform

Windows

### Describe the bug

Game freezes during AI turn.
Issue was found by one of our players.

### Save file

[error.zip](https://github.com/ihhub/fheroes2/files/8055422/error.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5008

Also fully assign the assessment of the army defending the tile to the `AIWorldPathfinder::processCurrentNode()`.

### DIFF
--- a/src/fheroes2/ai/ai.h
+++ b/src/fheroes2/ai/ai.h
@@ -60,7 +60,7 @@ namespace AI
     };
 
     const double ARMY_STRENGTH_ADVANTAGE_SMALL = 1.3;
-    const double ARMY_STRENGTH_ADVANTAGE_MEDUIM = 1.5;
+    const double ARMY_STRENGTH_ADVANTAGE_MEDIUM = 1.5;
     const double ARMY_STRENGTH_ADVANTAGE_LARGE = 1.8;
 
     class Base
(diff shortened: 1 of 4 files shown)


idshibanov (approved):

oleg-derevenetz:
The problem here is that this assessment is not 100% reliable, because of random nature of teleports. Suppose that teleport has two endpoints, one is guarded and another is not. In this case a weak AI hero can try to teleport to the second endpoint, but it will be teleported to the first one. In this case I suppose it will not be able to move away from the teleport tile until he is defeated by another hero (or his kingdom is vanquished).

idshibanov:
> AI just stands still being trapped by Unicorns, while in OG it will sacrifice itself. Maybe in the future it will be possible to add some logic for such cases.

I think we should cover this separately since you already mentioned this issue was already present. Let's make sure there's no freezes at least.

oleg-derevenetz:
Hi @idshibanov

> I think we should cover this separately since you already mentioned this issue was already present. Let's make sure there's no freezes at least.

Freezes (infinite loop) was caused by the fact that `getFogDiscoveryTile()` returned a tile that was protected by Griffins, and AI pathfinder was unable to create a path for that tile (because hero has not enough army to defeat the guardians), hero was not moved anywhere, it still has the movement points, and process repeated _ad infinitum_. Now this shouldn't happen.

ihhub:
@oleg-derevenetz , well done! Kudos to you!

PR #5012, opened 2022-02-13
Fix player's allies configuration for 'Defeat the other side' victory condition (resolves #5001).

### LINKED ISSUE #5001: Incorrect allies set in games with 'Defeat the other side' victory condition (Voyage Home #4 and some other maps affected)
### Preliminary checks

### Platform

Windows

### Describe the bug

Steps to reproduce:
a) Start Voyage Home, map 4: load 'vh2_Complete.savc' from zip, select map 4.
b) You can see Red player map is visible to you

1) I met this issue during playing Voyage Home map 4. In FHeroes2 Red computer player acts as an ally to a human player. You can see that instantly as all their territory is opened to you, then they fight on your side.
image
This felt wrong and broken. I cannot check it in the original game, since it has a different save file format (if I understand correctly). But according to the walk through below, Red indeed should be your enemy.
https://mightandmagic.fandom.com/wiki/Blood_is_Thicker

2) Allies with 'Defeat the other side' victory conditions are determined by wins1 field in the map structure (offset 0x20..0x21) and FHeroes2 handles it and treats it like index of a color which starts the second union, i.e. all the colors <wins1 - side1, all colors >= wins1 - side2. This seems incorrect and correct meaning is a number of players in side1. This often does not make a difference, but when players colors are assigned not in sequential order this does make a difference.

It can be proven in OG Map Editor on 'Last Hope' map for example. There are three players: Blue, Orange and Purple. Original union settings is Blue vs Orange+Purple, wins1=1. However after setting in Map Editor specs Blue+Orange vs Purple it can be seen that wins1 is set to 2. So it proves that wins1 is a number of players in side1.
image
image

3) So I think it should be fixed in Maps::FileInfo::ReadMP2:
image

With that change Voyage Home map 4 does not have Red played allied with a human:
image

### Save file

[saves.zip](https://github.com/ihhub/fheroes2/files/8044179/saves.zip)

### Additional info

_No response_

imiron13:


src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -367,21 +367,23 @@ bool Maps::FileInfo::ReadMP2( const std::string & filename )
     fs.seek( 0x76 );
     description = fs.toString( mapDescriptionLength );
 
-    // fill unions
+    // fill unions, see issue #5001, #4576
```
ihhub, line 370: We do not state issue numbers in the code. Please revert this change.

src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -367,21 +367,23 @@ bool Maps::FileInfo::ReadMP2( const std::string & filename )
...
 
+        const int numPlayersSide1 = wins1; // 'wins1' map spec parameter means number of players in side1
+        int playerIdx = 0; // index of a player among available players/colors
         for ( const int color : availableColors ) {
-            if ( Color::GetIndex( color ) < wins1 )
+            if ( playerIdx < numPlayersSide1 )
                 side1 |= color;
             else
                 side2 |= color;
+            playerIdx++;
```
ihhub, line 386: Please use pre-increment form instead of post-increment as per our standard.

src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -367,21 +367,23 @@ bool Maps::FileInfo::ReadMP2( const std::string & filename )
...
     if ( conditions_wins == VICTORY_DEFEAT_OTHER_SIDE && !skipUnionSetup ) {
         int side1 = 0;
         int side2 = 0;
 
         const Colors availableColors( kingdom_colors );
 
         assert( !availableColors.empty() );
-        wins1 += Color::GetIndex( availableColors.front() );
 
+        const int numPlayersSide1 = wins1; // 'wins1' map spec parameter means number of players in side1
```
ihhub, line 379: Please also add an assertion:
```cpp
assert( ( numPlayersSide1 > 0 ) && ( numPlayersSide1 < static_cast<int>( availableColors ) ) );
```
to make sure that the number is valid.

ihhub, line 379: I think this comment is not needed. We can drop it.

ihhub (changes requested):
Hi @imiron13 , thank you very much for your pull request. I left few comments here. Could you please take a look at them?

src/fheroes2/maps/maps_fileinfo.cpp
oleg-derevenetz: Hi @ihhub assertions are good when catching programming errors, caused by a programmer (such as calling a function with invalid parameter or violations in some internal data structures). However, here we will be dealing with invalid external data (invalid map file). I'm not sure that a "controlled crash" is a good option here. We can handle this in a "regular" way instead - by calling `DEBUG_LOG( DBG_GAME, DBG_WARN, ... )` and returning `false` from `ReadMP2()`.

idshibanov: I agree with Oleg on this one, I think assertion is overkill in this case. Logging the error and ignoring should be sufficient imo.

imiron13: I also prefer DEBUG_LOG here as map is an external resource. Added DEBUG_LOG+return false. With injection of incorrect value it prints a debug message and also does not show a corrupted map in the list of maps in New Game->Standard Game menu, which is nice behavior: corrupted map skipped, but the game does not crash.

ihhub:
@imiron13 , thank you so much for your contribution to the project!

imiron13:
Hooray!
Thank you guys for the great project. Keep it up.

PR #5013, opened 2022-02-14
Improve path calculation by WorldPathfinder & AIPathfinder

oleg-derevenetz:
* Additional assertions to catch potential programming errors early;
* Extended check for potential pathfinding loops (only if built without `NDEBUG`, i.e. with assertions enabled). The previous check for loops was more simplistic;
* Early determination of whether there is a path to a given index instead of `clear()` call at the end.


ihhub:
@oleg-derevenetz , thanks a lot for these code corrections!

PR #5007, opened 2022-02-12
Various sound-related fixes

oleg-derevenetz:
fix #4511

Also this fixes an issue with `Game::restoreSoundsForCurrentFocus()` in MIDI mode if there is no sound for the current terrain. Before (unmute the sound in this video):

https://user-images.githubusercontent.com/32623900/153728485-729adb0f-d4f3-4898-965b-f2dc9f65eb5b.mp4

After:

https://user-images.githubusercontent.com/32623900/153728488-16dd4c8c-9733-4256-9a46-210b65c98179.mp4

Also this PR removes the only asynchronous `AGG::ResetMixer()` call and reverts the commit that introduced asynchronous `AGG::ResetMixer()` variant because it's unreliable and it did work only because it was followed by the non-async `AGG::ResetMixer()` call in the `Interface::Basic::HumanTurn()` (now `Interface::Basic::StartGame()`).

PR #5017, opened 2022-02-14
Reset environment sounds and music theme at the beginning and at the end of the human turn only

oleg-derevenetz:
To avoid repeating the AI music for each AI player from scratch. I forgot about it.

Regression after #5007.


ihhub:
@oleg-derevenetz , great job! Thanks a lot for your efforts!

My apologies that I don't make a progress these days as I am being extremely busy with personal issues :(

PR #5018, opened 2022-02-14
Fix a few recent code smells

oleg-derevenetz:


ihhub:

PR #5019, opened 2022-02-14
Run the SonarCloud Analyzer workflow only for the main repo, not for forks

oleg-derevenetz:
I'm a bit tired of seeing a workflow failure notification after every fork sync :)


ihhub:
@oleg-derevenetz , if it works best for you then it's fine :)

PR #5022, opened 2022-02-15
Rename "Lean To" -> "Lean-To"

oleg-derevenetz:
In the OG:

Lean-To


ihhub:

PR #5023, opened 2022-02-16
Force to use only the original palette from the game

ihhub:
At maximum we can use only those colors which are defined by the application icon and the error message. No more than that.


src/engine/image_palette.cpp
```diff
@@ -43,7 +42,39 @@ namespace
     private:
         PaletteHolder()
         {
-            std::copy_n( kb_pal, paletteSize, gamePalette.begin() );
+            // This is a temporary palette which is used ONLY for cursor and game startup error message. A real palette must be loaded from AGG file.
+            gamePalette
+                = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
ihhub, line 47: Notice that this palette has a lot of zeroes in comparison with previous one.

oleg-derevenetz (approved):

PR #5025, opened 2022-02-16
Maps::FileInfo validation improvement

oleg-derevenetz:
Follow-up to #5012

* `Maps::FileInfo::ReadMP2()`: return false if map contains invalid list of kingdom colors instead of an assertion failure;
* `Campaign::ScenarioData::loadMap()`: validate result of the `ReadMP2()`.


ihhub:
@oleg-derevenetz , thanks a lot for these improvements!

PR #5026, opened 2022-02-17
Fix possible errors for translation related code

ihhub:
regression from #4976

### DIFF
--- a/src/fheroes2/monster/monster_anim.h
+++ b/src/fheroes2/monster/monster_anim.h
@@ -75,7 +75,7 @@ namespace fheroes2
         explicit RandomMonsterAnimation( const Monster & monster );
 
         RandomMonsterAnimation( const RandomMonsterAnimation & ) = delete;
-        RandomMonsterAnimation( RandomMonsterAnimation && ) = default;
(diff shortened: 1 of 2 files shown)


oleg-derevenetz (approved):

src/fheroes2/monster/monster_anim.h
```diff
@@ -75,7 +75,7 @@ namespace fheroes2
         explicit RandomMonsterAnimation( const Monster & monster );
 
         RandomMonsterAnimation( const RandomMonsterAnimation & ) = delete;
-        RandomMonsterAnimation( RandomMonsterAnimation && ) noexcept = default;
```
oleg-derevenetz: @ihhub Unfortunately we need at least C++14 for this :)

ihhub: Sad. Once the code is stable enough we can switch to C++14.

ihhub: You know what? Maybe it's time to do this. Maybe we even consider using GitHub action for this as it's a faster solution? There's a small amount of developers who use older VS but none of us does it. At the beginning of the project it was reasonable since it started in 2019. Now it's 2022 so we should be good to go.

PR #5029, opened 2022-02-17
World::ComputeStaticAnalysis(): reset the region information for all tiles at the very beginning

oleg-derevenetz:
This should fix the following crash due to the invalid assertion:

```
Assertion failed: (regionID < _regions.size()), function KingdomTurn, file ai_normal_kingdom.cpp, line 112.
```

How to reproduce:

1. Run fheroes2;
2. Load the following savefile: [!1.zip](https://github.com/ihhub/fheroes2/files/8089177/1.zip);
3. Load the following savefile: [!2.zip](https://github.com/ihhub/fheroes2/files/8089182/2.zip);
4. Skip the turn.

Maybe this can be fixed in some other way, perhaps @idshibanov will advise something.

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1051,7 +1051,7 @@ void Maps::Tiles::UpdateRegion( uint32_t newRegionID )
         _region = newRegionID;
     }
     else {
-        _region = 0;
+        _region = REGION_NODE_BLOCKED;
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1050,6 +1050,9 @@ void Maps::Tiles::UpdateRegion( uint32_t newRegionID )
     if ( tilePassable ) {
         _region = newRegionID;
     }
+    else {
+        _region = 0;
```
ihhub, line 1054: Should we use here `REGION_NODE_BLOCKED`?

idshibanov (approved):
Good solution, glad we had an assertion there. Obstacle tiles were not overridden when second game loaded, other stuff shouldn't be affected.

ihhub:

PR #5031, opened 2022-02-17
Clang-Tidy: disable readability-function-cognitive-complexity

oleg-derevenetz:
This diagnostic doesn't work reliably anyway because it will not be reported if function declaration (name/return type/arguments) is not changed as part of PR. It is better to use the SonarCloud for this.


ihhub:
@oleg-derevenetz , good job for making developer's life easier!

PR #5032, opened 2022-02-17
Interface::Basic::EventDefaultAction(): reset and redraw the focus after hero's action

oleg-derevenetz:
This should fix the issue where the sound of the terrain does not change when teleporting using Stone Liths with a space bar. Before:

https://user-images.githubusercontent.com/32623900/154559062-d52317b9-3020-49d3-8f8f-700623884a8b.mp4

After:

https://user-images.githubusercontent.com/32623900/154559089-6136ac21-181c-4526-8907-ef46fa6750a2.mp4


ihhub:

PR #5035, opened 2022-02-18
Migrate Windows build from AppVeyor and VS2015 to GitHub and VS2019

oleg-derevenetz:
What's new:

* MSVC 2015 -> 2019;
* Inno Setup 5 -> 6;
* MSVC14 redistributable DLLs are removed - it seems that they are not needed anymore;
* Zlib DLL is removed from SDL2 distribution, because it seems that it is not needed - zlib is linked statically.

Please note:

* Build numbers are changed (decreased compared to those that were before), so **_if you have some interim Windows pre-release installed and you want to update it to another interim pre-release with the same release number, you will need to uninstall it and reinstall from scratch_**. **_This is a one-time action_**. This does not apply to ordinary users who have "regular" releases installed, because release numbers are senior compared to build numbers.

Drawbacks:

* It seems that MinGW `make` is not able to deal with UTF-8 correctly, so I have to install Cygwin just to build translation files, because `files/lang/Makefile` contains UTF-8 symbols (AppVeyor has the same issue but it has Cygwin preinstalled). This increases build time significantly (however it's still better than with AppVeyor due to parallel builds). If someone knows a reliable way to make MinGW `make` work with  UTF-8 symbols, please tell me how;
* GitHub build artifacts (generated for pull requests to match the former AppVeyor behavior) are [zipped on-the-fly when artifact is downloaded](https://github.com/actions/upload-artifact#zipped-artifact-downloads), so currently we have EXE inside ZIP for Inno Setup packages and ZIP inside ZIP for standalone packages. This is currently unavoidable and cannot be treated.


ihhub:
Hi @oleg-derevenetz , one thing which I forgot to change for the past few releases: can we please name exe file as `fheroes2_windows_{platform}_{SDL-version}_installer.exe` as it is done for releases? In this case it is more obvious for players to determine the difference (and also easier for me while making a release).

oleg-derevenetz:
Hi @ihhub

> one thing which I forgot to change for the past few releases: can we please name an exe file as `fheroes2_windows_{platform}_{SDL-version}_installer.exe` as it is done for releases? In this case it is more obvious for players to determine the difference (and also easier for me while making a release).

Done.

oleg-derevenetz:
Also I gave this workflow a more generic name "MSVC", because in future we can move to the VS2022 for example, but `GITHUB_RUN_NUMBER` is linked to the workflow with the specific name. Therefore, we will have a bad choice between leaving the old name (VS 2019) and resetting the build counter.

ihhub:
@oleg-derevenetz , one more thing which I realized is important is an ability for our QA team to test changes in a pull request. AppVeyor stores artifacts for each build (check [this page](https://github.com/ihhub/fheroes2/wiki/Download-Windows-latest-compiled-packages-of-a-pull-request)) while I do not see an option for GitHub Actions. We can store them somewhere outside GitHub if needed.

ihhub:
> @oleg-derevenetz , one more thing which I realized is important is an ability for our QA team to test changes in a pull request. AppVeyor stores artifacts for each build (check [this page](https://github.com/ihhub/fheroes2/wiki/Download-Windows-latest-compiled-packages-of-a-pull-request)) while I do not see an option for GitHub Actions. We can store them somewhere outside GitHub if needed.

False alarm. GitHub Action stores all artifacts in Actions page.

.github/workflows/clang_analyzer.yml
```diff
@@ -10,8 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-      with:
```
ihhub, line 13: Could you please tell me the default clone depth for this action? Is it 50? If bigger than it slows down cloning of the repository.

oleg-derevenetz: It's 1.

ihhub: Then it's good!

.github/workflows/vs2019.yml
```diff
@@ -9,26 +11,99 @@ jobs:
...
-    - name: Install dependencies
+    - uses: actions/checkout@v2.4.0
+    - name: Install dependencies, cygwin and zip
       run: |
         cd script/windows
         ./install_packages.bat
-        cd ../..
-    - name: Add MSBuild to PATH
-      uses: microsoft/setup-msbuild@v1.0.2
+        choco.exe install cygwin cyg-get zip -y
```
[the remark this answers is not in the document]
ihhub: Let's skip caching for now. It's still faster then AppVeyor.

ihhub:
As of now I leave AppVeyor webhook active as we have pull requests older than this one. Once we merge some of them I will remove the hook.

PR #5037, opened 2022-02-19
Update license for source files

ihhub:
Every single file was modified within the project. To avoid confusion for source code readers it is important to have an up to date license information.

All links and emails are preserved.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   Free Heroes of Might and Magic II: https://github.com/ihhub/fheroes2  *
- *   Copyright (C) 2022                                                    *
+ *   Copyright (C) 2019 - 2022                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2008 by Andrey Afletdinov <fheroes2@gmail.com>          *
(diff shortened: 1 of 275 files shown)


src/engine/audio.cpp
```diff
@@ -1,8 +1,9 @@
 /***************************************************************************
- *   Copyright (C) 2008 by Andrey Afletdinov <fheroes2@gmail.com>          *
+ *   Free Heroes of Might and Magic II: https://github.com/ihhub/fheroes2  *
+ *   Copyright (C) 2022                                                    *
```
oleg-derevenetz, line 3: Wouldn't it be better to specify a range: 2019 - 2022?

oleg-derevenetz (commented):
Hi @ihhub I just left one question, please take a look when you get the chance :)

src/engine/audio.cpp
ihhub: This is an excellent question. I thought about this. If we use a range of years then we have to update all files every year (even only for changed files). So far there's no a perfect solution but I am open to hear your point of view.

oleg-derevenetz: I use a range of years in my own projects, but I'm too lazy to include the header in each source file, so update a few `LICENSE`/`COPYING` files once a year is not an issue for me :) With header in each file things getting more complicated, of course. In theory, we can make a separate check in Actions (or integrate it into the existing code style check) that will verify that each updated source file in the PR has valid copyright header with valid copyright date. So it's up to you, I'm fine with both solutions :)

ihhub: Let's make in your proposed way and we will have a whole year to decide what to do next.

ihhub: It's changed. Please take a look.

oleg-derevenetz (approved):

PR #5038, opened 2022-02-19
AI should save current unit if retreating

### LINKED ISSUE #4656: Hero nuke himself if he is fleeing and has armageddon spell
### Preliminary checks

### Platform

Windows

### Describe the bug

- create a hero with 1 unit and only armagedon spell
- have him fight a more powerful army
- hero will cast armageddon, even if that means killing himself

this is because in the code that evaluates the effect of mass spells, we don't check if the spell will destroy our own army. In that case, we should probably never cast the spell

### Save file

none

### Additional info

_No response_

idshibanov:
Fixes #4656 .

Happens when AI makes last spell cast before fleeing the battle and planning to invoke power mass spell like Armageddon.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_spell.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_spell.cpp
@@ -41,7 +41,7 @@ namespace AI
         return Board::DistanceFromOriginX( unit.GetHeadIndex(), unit.isReflect() );
     }
 
-    SpellSelection BattlePlanner::selectBestSpell( Arena & arena, bool retreating ) const
+    SpellSelection BattlePlanner::selectBestSpell( Arena & arena, const Battle::Unit & currentUnit, bool retreating ) const
     {
         // Cast best spell with highest heuristic on target pointer saved
         SpellSelection bestSpell;
@@ -87,7 +87,7 @@ namespace AI
                 continue;
 
             if ( spell.isDamage() ) {
-                checkSelectBestSpell( spell, spellDamageValue( spell, arena, friendly, enemies, retreating ) );
+                checkSelectBestSpell( spell, spellDamageValue( spell, arena, currentUnit, friendly, enemies, retreating ) );
             }
             else if ( spell.isEffectDispel() ) {
                 checkSelectBestSpell( spell, spellDispellValue( spell, friendly, enemies ) );
@@ -114,7 +114,8 @@ namespace AI
         return bestSpell;
     }
 
-    SpellcastOutcome BattlePlanner::spellDamageValue( const Spell & spell, Arena & arena, const Units & friendly, const Units & enemies, bool retreating ) const
+    SpellcastOutcome BattlePlanner::spellDamageValue( const Spell & spell, Arena & arena, const Battle::Unit & currentUnit, const Units & friendly, const Units & enemies,
+                                                      bool retreating ) const
     {
         SpellcastOutcome bestOutcome;
         if ( !spell.isDamage() )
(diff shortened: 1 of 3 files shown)


src/fheroes2/ai/normal/ai_normal_spell.cpp
```diff
@@ -150,18 +151,30 @@ namespace AI
                 spellHeuristic += damageHeuristic( enemy );
             }
             for ( const Unit * unit : friendly ) {
-                spellHeuristic -= damageHeuristic( unit );
+                const double valueLost = damageHeuristic( unit );
+                // check if we're retreating and will lose current unit
+                if ( retreating && unit->isUID( currentUnit.GetUID() ) && std::fabs( valueLost - unit->GetStrength() ) < 0.001 ) {
```
oleg-derevenetz, line 156: Hi @idshibanov just curious, is there a reason to use `isUID( GetUID() )` in such cases instead of pointer comparison? I see that UIDs are used mostly within commands (I suppose to support gaming over the network, because pointers cannot be passed over the network), and in other places pointers are simply compared.

idshibanov: I always thought that pointer comparison is unreliable especially since `currentUnit` reference is passed in from somewhere else and not directly a part of `friendly` collection.

oleg-derevenetz: Passing by reference doesn't change an address of an object, so we can apply `&` to the reference to an object and get the same pointer as if we applied `&` to the object itself, because references [are not separate objects](https://isocpp.org/wiki/faq/references#reseating-refs) and cannot be considered separately from the objects they refer to. However if somewhere in the call chain object was passed via a copy, then it's naturally a different story :) But AFAIK it's not the case in the battle-related code. Anyway it's up to you of course.

idshibanov: > if somewhere in the call chain object was passed via a copy, then it's naturally a different story

That's what I'm trying to avoid exactly. Although not the case today this is outside of the BattlePlanner class realm. Plus UIDs are already there and perfectly fit the need :)

oleg-derevenetz: @idshibanov Yeah, that's true. I'll propose a PR today that explicitly disables copy and move constructors and assignment operators for `Battle::Unit` and descendants, because current code in many places relies on the fact that all units are in heap and their addresses can be freely compared.

ihhub: Hi @idshibanov and @oleg-derevenetz , since copy and move constructors are disabled should we change the code here?

src/fheroes2/ai/normal/ai_normal_spell.cpp
```diff
@@ -150,18 +151,30 @@ namespace AI
...
             // Area of effect spells like Fireball
-            auto areaOfEffectCheck = [&damageHeuristic, &bestOutcome]( const TargetsInfo & targets, const int32_t index, int myColor ) {
+            auto areaOfEffectCheck = [&damageHeuristic, &bestOutcome, &currentUnit, &retreating]( const TargetsInfo & targets, const int32_t index, int myColor ) {
                 double spellHeuristic = 0;
                 for ( const TargetInfo & target : targets ) {
                     if ( target.defender->GetCurrentColor() == myColor ) {
-                        spellHeuristic -= damageHeuristic( target.defender );
+                        const double valueLost = damageHeuristic( target.defender );
+                        // check if we're retreating and will lose current unit
+                        if ( retreating && target.defender->isUID( currentUnit.GetUID() ) && std::fabs( valueLost - target.defender->GetStrength() ) < 0.001 ) {
```
ihhub, line 173: I think the same question highlighted by @oleg-derevenetz is applied for this line as well.

ihhub (changes requested):
Hi @idshibanov , I am marking this pull request as requires changes just to notify you about highlighted by @oleg-derevenetz possible improvement.

src/fheroes2/ai/normal/ai_normal_spell.cpp
oleg-derevenetz: Hi @ihhub this is not necessary, I'm OK with both variants.

ihhub: Well, in this case let's proceed with merging :)

PR #5039, opened 2022-02-19
Explicitly disable copy (and therefore move) constructor of Battle::Unit and Battle::Tower

oleg-derevenetz:
In addition to already disabled copy (and therefore move) assignment for these classes. Current battle-related code in many places relies on the fact that all units are in heap and their addresses can be freely compared - so at least such a unit cannot be accidentally passed as a function argument via a copy (or move) instead of pointer or reference.


ihhub:
@oleg-derevenetz , thanks a lot for this code correction!

PR #5044, opened 2022-02-21
Compile with -pthread, not just link with it

oleg-derevenetz:
At least with GCC, `-pthread` implicitly adds `-D_REENTRANT`, which is the compile-time flag:

```
oleg@gentoo ~ $ gcc -dumpspecs | grep pthread
%{posix:-D_POSIX_SOURCE} %{pthread:-D_REENTRANT}
[...]
```

so using it just with linker is not enough (at least in general).

This issue came out when I tried to build fheroes2 as WASM using Emscripten (just out of curiosity). Emscripten was not able to link the final executable with `-pthread` if there were object files that used threads or atomics, but built without `-pthread`.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -43,7 +43,7 @@ jobs:
         brew install ${{ matrix.config.dependencies }}
     - name: Build
       run: |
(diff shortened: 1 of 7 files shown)


ihhub:
@oleg-derevenetz , thanks a lot for changing it!

Issue #4298, opened 2021-10-01
Adventure map, Ultimate Artifact, not possible to dig for the Ultimate Artifact if the scenario creator has placed it on a river


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, if the Ultimate Artifact is manually placed on a river by the scenario creator, it is not possible to dig for it.
This is possible in the original game.

In original game:

![Ultimate on river H2 001](https://user-images.githubusercontent.com/43583994/135605439-63666167-2504-4edd-ae98-9cc9abd4681e.png)
![Ultimate on river H2 002](https://user-images.githubusercontent.com/43583994/135605443-d55f461e-247f-4594-9ecf-e20a1f65059b.png)
![Ultimate on river H2 003](https://user-images.githubusercontent.com/43583994/135605449-5d7b49e0-b1f8-4b5e-90e0-9a00eec58f07.png)
![Ultimate on river H2 004](https://user-images.githubusercontent.com/43583994/135605454-add5176a-f8f9-447b-97e6-207e1ff22ac4.png)
![Ultimate on river H2 007](https://user-images.githubusercontent.com/43583994/135605461-f7ffaf44-3a53-4046-a703-94edc084e415.png)
![Ultimate on river H2 008](https://user-images.githubusercontent.com/43583994/135605469-62da61e4-5c0e-4c67-b491-a6874d6d819d.png)

In fheroes2:

![Ultimate on river 001](https://user-images.githubusercontent.com/43583994/135605567-b1adc452-f645-4f57-bec1-05cacb1a0cb5.png)
![Ultimate on river 002](https://user-images.githubusercontent.com/43583994/135605570-2d25e7dd-07f3-4980-8e91-5cfdec4384d9.png)
![Ultimate on river 003](https://user-images.githubusercontent.com/43583994/135605576-3113a46c-e45c-4bf8-ac2b-771bcf8661e9.png)
![Ultimate on river 004](https://user-images.githubusercontent.com/43583994/135605581-464e21ed-3eba-4bf5-8fb8-f1c3cae93be4.png)
[...]

PR #5045, opened 2022-02-21
#4298 You can now dig in streams if there is an artifact there

DeiNile:
Adding a check to `Maps::Tiles::GoodForUltimateArtifact()` did the trick.

However, I am not entirely convinced that this is the best way to do it? For one, it is not entirely obvious to me what is meant by `TilesAddon::isShadow`. Conceptually speaking, what is a shadow? I know that `isValidShadowSprite()` ultimately determines this, but it is not clear to me exactly what is considered a shadow. Furthermore, since the original check was `addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) )` there is an assumption here that you can only dig in shadows. This is a little counter-intuitive to me, since you can obviously dig in more spots than shadows. You can dig on most grass tiles etc., so there must be some kind of misconception I have about what a "shadow" is here. If anyone can shed some light this, that would be helpful.

Several things already use `Maps::TilesAddon::isShadow()` so I did not want to change the implementation of that, lest I introduce bugs elsewhere too.

Now, the other thing I am curious about is why the original check - `addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) )` - assumed that `addons_level1` contained only shadow elements? In my testing that is now always the case. When a road intersects a stream, for instance, there are multiple elements in this list. I changed the condition `addons_level1.size() >= ...` to take this into account, which is demonstrated in my screenshots below. With that being said, I am not sure if I violated some core assumption that might have knock-on effects elsewhere.

stream-digging

My original implementation did not account for when a stream intersects with a road. The original game allows this.

*Original game*
stream-digging-puzzle-original

If you try to dig in the intersection and there is no artifact it yells at you:

stream-digging-bad-intersection-original

But if there is an artifact there, it will let you dig:

stream-digging-good-intersection-original

*fheroes 2*

stream-digging-bad-intersection-fheroes

stream-digging-good-intersection-fheroes2

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -589,6 +589,11 @@ bool Maps::TilesAddon::isShadow( const TilesAddon & ta )
     return Tiles::isShadowSprite( ta.object, ta.index );
 }
 
+bool Maps::TilesAddon::isStream( const TilesAddon & ta ) 
+{
+    return MP2::GetICNObject( ta.object ) == ICN::STREAM;
+}
+
 bool Maps::Tiles::isShadowSprite( const int icn, const uint8_t icnIndex )
 {
     return isValidShadowSprite( icn, icnIndex );
@@ -1620,7 +1625,8 @@ bool Maps::Tiles::GoodForUltimateArtifact() const
     }
 
     if ( objectTileset == 0 || isShadowSprite( objectTileset, objectIndex ) ) {
-        return addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) );
+        return ( addons_level1.size() >= static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) ) )
+            || ( addons_level1.size() >= static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isStream ) ) );
     }
 
     return false;
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -80,6 +80,7 @@ namespace Maps
         std::string String( int level ) const;
 
         static bool isShadow( const TilesAddon & );
+        static bool isStream( const TilesAddon & );
 
         static bool isResource( const TilesAddon & );
         static bool isArtifact( const TilesAddon & );


oleg-derevenetz:
@DeiNile

> Furthermore, since the original check was `addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) )` there is an assumption here that you can only dig in shadows.

That's not quite true. What's actually being checked there is that this tile either has no level 1 add-ons at all or all add-ons are shadows. There are level1 add-ons that you can't dig on - for example, AFAIK, you can't dig on roads, which are add-ons as well. Also some parts of large terrain objects are add-ons as well. You need to be extremely careful and make sure that you still can't dig where you shouldn't be able to do it. I suppose @ihhub is able to add something on this topic.

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1620,7 +1625,8 @@ bool Maps::Tiles::GoodForUltimateArtifact() const
     }
 
     if ( objectTileset == 0 || isShadowSprite( objectTileset, objectIndex ) ) {
-        return addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) );
+        return ( addons_level1.size() >= static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) ) )
```
oleg-derevenetz, line 1628: It seems that check of this kind is pointless because `addons_level1.size()` is ALWAYS greater or equal than the right part of this expression :) The whole is always greater than or equal to some  part of this whole.

DeiNile: Derp... You are of course correct @oleg-derevenetz  :p For reference I originally did
```c++
return addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) )
    || addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isStream ) );
```

That works for this specific issue, but it does not work when a road intersects with a stream. So I was attempting to solve for both problems at the same time. And when a road intersects a stream, `addons_level1.size() == 2`, so it seems problematic to do a strict equality check against the number addons. Especially when I am not 100% clear on what exactly a shadow is.

I could of course rewrite it to be
```c++
return addons_level1.size() == static_cast<size_t>( std::count_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isShadow ) )
    || std::find_if( addons_level1.begin(), addons_level1.end(), TilesAddon::isStream ) != addons_level1.end();
```

That should better retain the previous behavior, while being stricter about the stream check. I guess my question about that would be: is it "bad" to not consider the number of addons for this particular portion of the check?

oleg-derevenetz: Hi @DeiNile

In its current form this PR is undoubtedly leads to some regressions. Consider:

Dig

You should not be able to dig in places like that. The structure of such places is as follows:

```
----------------:>>>>>>>>
Tile index      : 1617, point: (33, 22)
uniq            : 1602
mp2 object      : 0, (No object)
tileset         : 176, (TREFALL.ICN)
object index    : 3, (animated: 0)
level           : 2
region          : 9
ground          : Grass, (isRoad: 1)
shadow          : 1
passable        : center, top, right, bottom-right, bottom, bottom-left, left
quantity 1      : 3
quantity 2      : 0
quantity 3      : 0
----------------1--------
uniq            : 673
tileset         : 122, (ROAD.ICN)
index           : 0
level           : 3, (3)
shadow          : 0
----------------I--------
----------------:<<<<<<<<
```

so we have a tileset with a shadow on this tile (that in itself is not an obstacle for digging), BUT there is also an addon with the road, and this is not allowed.

DeiNile: I see, thanks for the insight @oleg-derevenetz. I misunderstood what the underlying issue was then. Digging in streams, roads etc. are really edge cases that are only allowed when the ultimate artifact has been placed "under" such objects. I think I already have a solution partially working, I just have to double check what addon tiles you can place the artifact under, which will take a little bit of time.

DeiNile: Alright, so I looked at it a bit more, and I have identified a bunch of different "shapes" or features on the ground that you are normally not allowed to dig in, but if the ultimate artifact is there you should be able to. In the picture below you can see all the shapes (not sure what the best term for these are, please let me know) to the right of the town.

Screenshot 2022-04-14 100110

So now it is just a matter of determining if an area has any of these shapes, and this is where I am running into a few issues. On grass I have determined that the valid icons are all `ICN::OBJNGRAS`, `ICN::OBJNGRA2` AND `ICN::OBJNMUL2`, but these are obviously not accurate enough to tell me exactly what shapes are placed on the tile. When I look at `Maps::TilesAddon::isRoad()` I see a lot of checks against `index`, and something very similar happens in `Maps::Tiles::isStream()`. How do you know what `index` corresponds to which "shape"? For instance, how would I know if a crater is placed on a tile? @oleg-derevenetz , any guidance you can provide to determine this information would be very helpful.

oleg-derevenetz: Hi @DeiNile I believe @ihhub can tell you more about this than I can :) He is much more familiar with the design of adventure maps than I am.

ihhub: The current implementation as stated by @oleg-derevenetz works like this:
- if no objects of level 1 present allow to dig
- if all objects of level 1 present allow to dig
- if another case do not allow to dig

Besides division by two levels there is another parameter responsible for extra categories: `level` within **TilesAddon** class. For roads, streams, cracks and other spots this value is **3**. Moreover, you can take a look at this code:
```cpp
// Count how many objects are there excluding shadows, roads and river streams.
const std::ptrdiff_t validLevel1ObjectCount = std::count_if( addons_level1.begin(), addons_level1.end(), []( const TilesAddon & addon ) {
    if ( TilesAddon::isShadow( addon ) ) {
        return false;
    }

    const int icnType = MP2::GetICNObject( addon.object );
    return icnType != ICN::ROAD && icnType != ICN::STREAM;
} );
```
which covers most of cases.

In conclusion, since we want to have the same behavior as in the original game the code must be modified like this:
```cpp
bool Maps::Tiles::GoodForUltimateArtifact() const
{
    if ( isWater() || !isPassableFrom( Direction::CENTER, false, true, 0 ) ) {
        return false;
    }

    if ( objectTileset != 0 && isShadowSprite( objectTileset, objectIndex ) ) {
        return false;
    }

    const std::ptrdiff_t validLevel1ObjectCount = std::count_if( addons_level1.begin(), addons_level1.end(), []( const TilesAddon & addon ) {
        if ( TilesAddon::isShadow( addon ) ) {
            return false;
        }

        const int icnType = MP2::GetICNObject( addon.object );
        return icnType != ICN::STREAM;
    } );

    return validLevel1ObjectCount == 0;
}
```

Now this function should be used only for **setting** an Ultimate Artifact randomly which works exactly as in the original game. However, for real digging of the artifact we might need a separate function. @DeiNile , could you please check how the original game shows places for digging when the artifact is placed there?

ihhub:
Closing this pull request in favor of #5720 .

PR #5048, opened 2022-02-22
WIP: add direct up and down attacks for wide monsters

### LINKED ISSUE #501: Direct up and down attacks not handled correctly for double-hex units
While attacking from a certain angle Phoenix make one wing swing and then attacks. But he stays the same place.
Flying animation
Flying animation2
Flying animation3

ihhub:
close #501

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -181,13 +181,94 @@ namespace
             }
         }
     }
+
+    bool isSwordAttackCursorTheme( const int theme )
+    {
+        switch ( theme ) {
+        case Cursor::SWORD_TOPRIGHT:
+        case Cursor::SWORD_RIGHT:
+        case Cursor::SWORD_BOTTOMRIGHT:
+        case Cursor::SWORD_BOTTOMLEFT:
+        case Cursor::SWORD_LEFT:
+        case Cursor::SWORD_TOPLEFT:
+        case Cursor::SWORD_TOP:
+        case Cursor::SWORD_BOTTOM:
+            return true;
+        default:
+            break;
+        }
+
+        return false;
+    }
+
+    bool isAttackCursorTheme( const int theme )
+    {
+        if ( isSwordAttackCursorTheme( theme ) ) {
+            return true;
+        }
+
+        switch ( theme ) {
+        case Cursor::WAR_ARROW:
+        case Cursor::WAR_BROKENARROW:
+            return true;
+        default:
+            break;
+        }
+
+        return false;
+    }
+
+    int GetSwordCursorDirection( const int direction, const bool isWideMonster, const bool isReflected )
+    {
+        switch ( direction ) {
+        case Battle::BOTTOM_RIGHT:
+            return ( isWideMonster && isReflected ) ? Cursor::SWORD_TOP : Cursor::SWORD_TOPLEFT;
+        case Battle::BOTTOM_LEFT:
+            return ( isWideMonster && !isReflected ) ? Cursor::SWORD_TOP : Cursor::SWORD_TOPRIGHT;
+        case Battle::RIGHT:
+            return Cursor::SWORD_LEFT;
+        case Battle::TOP_RIGHT:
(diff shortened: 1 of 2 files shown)


oleg-derevenetz:
Hi @ihhub are you sure that supporting these "direct attacks" is really worth it? According to my experiments, these are just mysterious and non-obvious variants of cell-cell attacks (I mean, they are actually cell-cell attacks, but with a non-obvious direction of attack for the player).

ihhub:
> Hi @ihhub are you sure that supporting these "direct attacks" is really worth it? According to my experiments, these are just mysterious and non-obvious variants of cell-cell attacks (I mean, they are actually cell-cell attacks, but with a non-obvious direction of attack for the player).

I agree with you on this. Adding these directions is not trivial and it leads to more confusion. I can close this pull request and the corresponding issue since you proposed the solution for the rest of attack direction problems.

oleg-derevenetz:
Hi @ihhub should I implement something specific for this?

ihhub:
@oleg-derevenetz , I am not sure that we really need some extra implementation for this. @Branikolog what do you think about NOT implementing such behavior at all to avoid player's confusion?

Branikolog:
Hi, @oleg-derevenetz & @ihhub 
I think, we should handle these attacks, as they provide more tactical opportunities. 

https://user-images.githubusercontent.com/55348946/159165217-b7df3bb2-8645-4f00-b9d2-2585104dff48.mp4

Here we can choose between three attack directions, while trying to attack zombies.

oleg-derevenetz:
Hi @Branikolog

> Here we can choose between three attack directions, while trying to attack rogues.

In fheroes2 the direct attack from above and attack from the top right cell are currently absolutely the same, because the main cell of the wide unit is the head cell, like in HoMM3. The "direct attack" in OG is the way to force the wide unit to attack from the head cell instead of the tail cell (which is the main cell in OG).

Branikolog:
Hi, @oleg-derevenetz !
With the current fheroes2 implementation we're restricted in attacking only from two directions (in my example above). Sure, we can implement attacking from a tail in other manner(with the use of keyboard, for instance), but the original implementation seems quite comfortable to me.

oleg-derevenetz:
Hi @Branikolog

> With the current fheroes2 implementation we're restricted in attacking only from two directions (in my example above). Sure, we can implement attacking from a tail in other manner(with the use of keyboard, for instance), but the original implementation seems quite comfortable to me.

In OG these directions are:

1. Attack from cell - this attack means that the target unit will be attacked from the main cell of the attacking unit - his tail cell (if it is possible, of course):

https://user-images.githubusercontent.com/32623900/159256546-7245fd68-8277-4169-a025-69e6ab23aa0a.mp4

2. Attack "from above" or "from below" - this attack means that target unit will be attacked from the secondary cell of the attacking unit - his head cell:

https://user-images.githubusercontent.com/32623900/159256758-57702f04-0bd4-4749-a3b1-cb5a0d15defc.mp4

Since, unlike OG, in fheroes2 the main cell of a wide unit is head cell, I propose making the second type of attack an attack from the secondary cell too - the tail cell. So in this case:

Before attack

the unit should take the following position and attack should be performed in the following direction:

Attack

This will not require global alterations in terms of changing the main cell of wide units to the tail cell instead of the head cell (I still think the head cell is the best choice for the main cell of wide units) and it will give the necessary level of flexibility.

Branikolog:
@oleg-derevenetz 
In the original game, in a case of direct up/down attack, second creature being affected by a breath attack is the one, that is closer to the tail cell of an attacking creature.
image
image
I believe, this was made to be able to proceed breath attack over the troops, close to the starting point (in a case there's no enough room to attack from a right diagonal cell):
image
This is what I want to mention, considering direct up/down attacks.

In the screenshots, you've provided, There should be a possibility to place Phoenixes at three different places (like the OG allowed as well):
1:
image
2:
image
3:
image

oleg-derevenetz:
@Branikolog

> In the original game, in a case of direct up/down attack, second creature being affected by a breath attack is the one, that is closer to the tail cell of an attacking creature.

Yeah, that's because the direct up/down attack is performed from the "non-main" head cell:

Attack

> There should be a possibility to place Phoenixes at three different places (like the OG allowed as well):

This is impossible without changing the "main" cell of wide units to the tail cell.

PR #5049, opened 2022-02-22
Include translation context for the "Defender" scenario

a1exsh:
At least in Russian the existing translation for this string, which is
displayed in the battle summary, doesn't work in the context of the scenario
name.


a1exsh:
Not updating the `.po` files for now — that creates a lot of diffs :(

zenseii (approved):

ihhub:
@a1exsh , thank you for the change!

PR #5055, opened 2022-02-24
GitHub Actions: add script to check copyright headers

oleg-derevenetz:
Discussed in #5037. Sample is available [here](https://github.com/oleg-derevenetz/fheroes2/runs/5312654307?check_suite_focus=true).


ihhub:
@oleg-derevenetz , thanks a lot for this new check!

PR #5059, opened 2022-02-27
Fix directory and file detection on Windows

ihhub:
It saves a bit of time.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -223,7 +223,23 @@ char * System::GetOptionsArgument( void )
 
 bool System::IsFile( const std::string & name, bool writable )
 {
+    if ( name.empty() ) {
+        // An empty path cannot be a file.
+        return false;
+    }
+
 #if defined( _MSC_VER )
+    const DWORD fileAttributes = GetFileAttributes( name.c_str() );
+    if ( fileAttributes == INVALID_FILE_ATTRIBUTES ) {
+        // This path doesn't exist.
+        return false;
+    }
+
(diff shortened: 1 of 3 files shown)


oleg-derevenetz (commented):
Hi @ihhub There are also `System::IsFile()` and `System::IsDirectory()`. While with POSIX both of them work correctly (by checking the inode type), it seems that with Windows and other systems they do not check the file type at all. We can fix them and use here. What do you think?

ihhub:
> Hi @ihhub There are also `System::IsFile()` and `System::IsDirectory()`. While with POSIX both of them work correctly (by checking the inode type), it seems that with Windows and other systems they do not check the file type at all. We can fix them and use here. What do you think?

Sure. We can do this.

oleg-derevenetz (approved):

src/engine/system.cpp
```diff
@@ -223,7 +223,23 @@ char * System::GetOptionsArgument( void )
...
+    if ( fileAttributes == INVALID_FILE_ATTRIBUTES ) {
+        // This path doesn't exist.
+        return false;
+    }
+
+    if ( ( fileAttributes & FILE_ATTRIBUTE_DIRECTORY ) != 0 ) {
+        // This is a directory.
+        return false;
+    }
+
```
oleg-derevenetz, line 242: BTW we can add TODO comments for `ANDROID` and `FHEROES2_VITA` cases to not forget to add the similar checks in future here and in the `System::IsDirectory()`.

ihhub: It is done.

PR #5060, opened 2022-02-27
Fix the versioning of toshimaru/auto-author-assign

oleg-derevenetz:
It seems that this Action doesn't follow the [recommended versioning practice](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-release-management-for-actions), so specifying just a major version is not enough :(


ihhub:
@oleg-derevenetz , thanks for the fix!

PR #5062, opened 2022-03-03
Use move constructor instead of copying strings by value

ihhub:
In this case we avoid extra memory copy. Also this pull request contains some small code cleanup.


src/fheroes2/gui/ui_dialog.h
```diff
@@ -111,6 +111,8 @@ namespace fheroes2
     public:
         ResourceDialogElement( const int32_t resourceType, const std::string & text );
 
+        ResourceDialogElement( const int32_t resourceType, std::string && text );
```
oleg-derevenetz, line 114: Instead of these two constructors we can have just one here and in other similar places:

```cpp
ResourceDialogElement( const int32_t resourceType, std::string text );
```

and initialize member `_text` using `std::move()`:

```cpp
_text( std::move( text ) )
```

In this case, the code would be a little simpler (and we will not need the separate `init()`), BUT we will perform two movements where we currently do just one. Considering that moving is a cheap operation, the simpler code may be worth it, but it's up to you.

src/fheroes2/gui/text.h
```diff
@@ -125,7 +125,7 @@ class TextSprite : protected Text
 class TextBox : protected fheroes2::Rect
 {
 public:
-    TextBox();
+    TextBox() = delete;
```
oleg-derevenetz, line 128: Since we have other user-defined constructors here, we can just remove this.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few optional suggestions, please take a look when it's convenient for you.

src/fheroes2/gui/ui_dialog.h
ihhub: This is quite tricky. On the one hand, you are absolutely right. We must reduce code duplication. On the other hand, other developer might tend to change passing by value to passing by reference. Also static analyzers might complain about this.

What do you think?

src/fheroes2/gui/text.h
ihhub: It is removed.

src/fheroes2/gui/ui_dialog.h
oleg-derevenetz: Regarding static analyzers, `clang-tidy` is not against this idiom and suggests it in its `modernize-pass-by-value` check:

https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html

But the current approach would work as well, but it's up to you :)

ihhub: Tough decision :)

ihhub:
@oleg-derevenetz , I decided to keep the same style we have right now as I can still see some small potential issues with other developers not being used move operators :(

PR #5071, opened 2022-03-05
Fix Sirens logic, AI interaction and UI

ihhub:
- AI did not have any action while moving onto Sirens object while it should
- sirens must remove at least one unit from a stack which has more than 1 unit
- add experience UI element to Sirens dialog (close #4699 )

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1769,13 +1769,15 @@ namespace AI
 
     void AIToSirens( Heroes & hero, const MP2::MapObjectType objectType, const int32_t objectIndex )
     {
-        if ( !hero.isObjectTypeVisited( objectType ) ) {
-            const uint32_t experience = hero.GetArmy().ActionToSirens();
-            hero.IncreaseExperience( experience );
-
-            hero.SetVisited( objectIndex );
-
-            DEBUG_LOG( DBG_GAME, DBG_INFO, hero.GetName() << " visited Sirens and got " << experience << " experience." )
+        if ( hero.isObjectTypeVisited( objectType ) ) {
+            return;
         }
+
+        const uint32_t experience = hero.GetArmy().ActionToSirens();
+        hero.IncreaseExperience( experience );
+
+        hero.SetVisited( objectIndex );
+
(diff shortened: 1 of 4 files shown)


src/fheroes2/army/army.cpp
```diff
@@ -1250,21 +1250,33 @@ void Army::JoinStrongestFromArmy( Army & army2 )
...
-    u32 res = 0;
+    uint32_t experience = 0;
 
-    for ( iterator it = begin(); it != end(); ++it )
-        if ( ( *it )->isValid() ) {
-            const u32 kill = ( *it )->GetCount() * 30 / 100;
+    for ( Troop * troop : *this ) {
+        assert( troop != nullptr );
+
+        if ( troop->isValid() ) {
```
oleg-derevenetz, line 1260: We can flatten the code a bit here by changing this to:

```cpp
if ( !troop->isValid() ) {
    continue;
}
```

oleg-derevenetz (changes requested):
Hi @ihhub I just left one minor suggestion, please take a look when you have a free time.

src/fheroes2/army/army.cpp
ihhub: Done.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1762,4 +1766,16 @@ namespace AI
 
         DEBUG_LOG( DBG_GAME, DBG_INFO, hero.GetName() << " visited Alchemist Tower to remove " << cursed << " artifacts." );
     }
+
+    void AIToSirens( Heroes & hero, const MP2::MapObjectType objectType, const int32_t objectIndex )
+    {
+        if ( !hero.isObjectTypeVisited( objectType ) ) {
```
oleg-derevenetz, line 1772: Oh, we can flatten the code here as well BTW :) And I recommend to sync this PR with `master` since #5072 is now merged.

oleg-derevenetz (changes requested):
Hi @ihhub I just found another place where the code may be flatten :)

src/fheroes2/ai/ai_hero_action.cpp
ihhub: Fixed and updated.

oleg-derevenetz (approved):

PR #5072, opened 2022-03-05
Increase the fetch-depth for clang_tidy.yml and code_style_check.yml workflows

oleg-derevenetz:
These workflows access commits other than the current one (such as `HEAD^`) and default `fetch-depth` (1) is not enough for them.

P.S. I recommend to sync existing PRs with `master` after merging this one.


ihhub:

PR #5075, opened 2022-03-06
Remember the last open page in Spell Book during battle

### LINKED ISSUE #1565: During combat and on the adventure map, remember the last page of the magic book opened by the hero
One enhancement I would like to see in fheroes2 is that the program remembers the last page opened by the hero when he uses his magic book.

This would avoid, for example during the combat, that the player would have to change page to use for the third time in a row the spell located in the fifth page of the magic book.

In fheroes2:

Remember last page of magic book 01

Remember last page of magic book 02

ihhub:
partially addresses #1565


ihhub:
Converting this into draft based on @Branikolog feedback.

src/fheroes2/spell/spell_book.cpp
```diff
@@ -163,38 +163,43 @@ namespace
...
+                       const std::function<void( const std::string & )> * statusCallback ) const
 {
     if ( !hero.HaveSpellBook() ) {
         Dialog::Message( "", _( "You have no Magic Book, so you cannot cast a spell." ), Font::BIG, Dialog::OK );
         return Spell::NONE;
     }
 
-    Filter currentFilter = displayableSpells == Filter::ALL ? Filter::ADVN : displayableSpells;
-    SpellStorage displayedSpells = SetFilter( currentFilter, &hero );
+    if ( !restorePreviousState || ( displayableSpells != Filter::ALL && _spellFilter != displayableSpells ) ) {
```
oleg-derevenetz, line 174: AFAIK we should here set the `_spellFilter` to `Filter::ADVN` if `displayableSpells` is `Filter::ALL`, otherwise filter will not work properly when we open the spell book from the hero's inventory screen. With this change all spells will be shown in the spell book, while in reality we want to show only adventure spells first but with an ability to switch to the combat spells. Am I wrong?

oleg-derevenetz (commented):
Hi @ihhub I just left one question, please take a look when it is convenient for you.

src/fheroes2/spell/spell_book.cpp
ihhub: It is fixed.

oleg-derevenetz (approved):

PR #5078, opened 2022-03-06
Remember map size selection

### LINKED ISSUE #3260: Scenario selection window, scenario map size arrangement should be memorized
image
Map size arrangement should be saved, when we choose scenario and then return to the list back. Currently, in fheroes2 each time we enter scenario list, upper "map size" buttons never save their state.
Also, in the OG these buttons save their state even if we start game and then return back to main menu.

ihhub:
close #3260

### DIFF
--- a/src/fheroes2/dialog/dialog_selectscenario.cpp
+++ b/src/fheroes2/dialog/dialog_selectscenario.cpp
@@ -44,9 +44,6 @@
 
 namespace
 {
-    // This variable is used to remember the selection of map size throught the game.
-    int selectedMapSize = Maps::mapsize_t::ZERO;
-
     void mapInfo( const Maps::FileInfo & info )
     {
         // On some OSes like Windows, the path may contain '\' symbols. This symbol doesn't exist in the resources.
@@ -355,6 +352,9 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & all )
     fheroes2::Button buttonSelectAll( rt.x + 285, rt.y + 22, ICN::REQUESTS, 17, 18 );
 
     fheroes2::ButtonBase * currentPressedButton = nullptr;
+
+    // This variable is used to remember the selection of map size throught the game.
+    static int selectedMapSize = Maps::mapsize_t::ZERO;
 
     switch ( selectedMapSize ) {
     case Maps::SMALL:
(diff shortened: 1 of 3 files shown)


src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -334,8 +350,47 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & all, siz
     fheroes2::Button buttonSelectXLarge( rt.x + 223, rt.y + 22, ICN::REQUESTS, 15, 16 );
     fheroes2::Button buttonSelectAll( rt.x + 285, rt.y + 22, ICN::REQUESTS, 17, 18 );
 
-    buttonSelectAll.press();
-    fheroes2::ButtonBase * currentPressedButton = &buttonSelectAll;
+    fheroes2::ButtonBase * currentPressedButton = nullptr;
+
+    // This variable is used to remember the selection of map size throught the game.
```
oleg-derevenetz, line 355: I suppose there is a typo: `throught` -> `through` :)

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -359,8 +414,33 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & all, siz
...
     listbox.SetAreaMaxItems( 9 );
-    listbox.SetAreaItems( fheroes2::Rect( rt.x + 55, rt.y + 55, 270, 175 ) );
-    listbox.SetListContent( const_cast<MapsFileInfoList &>( all ) );
+    listbox.SetAreaItems( { rt.x + 55, rt.y + 55, 270, 175 } );
+
+    size_t selectedId = 0;
+
+    switch ( selectedMapSize ) {
+    case Maps::SMALL:
+        listbox.SetListContent( const_cast<MapsFileInfoList &>( small ) );
```
oleg-derevenetz, line 423: Are you sure we need `const_cast` here and in other similar places? `all` is the reference to const, but `small`, `medium`, etc are not.

oleg-derevenetz (changes requested):
Hi @ihhub I left a few suggestions here, please take a look when it's convenient for you.

src/fheroes2/dialog/dialog_selectscenario.cpp
ihhub: You're right!

src/fheroes2/dialog/dialog_selectscenario.cpp
ihhub: Yup, copy-paste mistake.

oleg-derevenetz (approved):

PR #5083, opened 2022-03-06
Check for the git failures in pipelines

oleg-derevenetz:
Related to #5072

Previously, `git` failures in pipelines were not detected and everything went like the `git` output was just empty. Now such failures will be detected like [here](https://github.com/ihhub/fheroes2/runs/5439323484?check_suite_focus=true) and [here](https://github.com/ihhub/fheroes2/runs/5439323619?check_suite_focus=true) - just in case if `fetch-depth: 50` will not be deep enough.


ihhub:
@oleg-derevenetz , well done! Now our pipeline is super stable :)

Issue #4233, opened 2021-09-20
Add support for GOG FLAC music


lleonis:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

GOG provides a "Soundtrack (FLAC)" package in the extra downloads. It would be nice to support it, because it allows to losslessly enjoy Paul Anthony Romero's wonderful compositions.

### Describe the possible solution

I guess this must not be that difficult, because I have successfully hacked it to work like this:

- GOG package contains a single mp3 with the MIDI main theme and a cover jpg - delete them
- Track 25 - 26 - 27 (Exploration III) needs to be copied to have 3 identical files
- Rename all flac files to Track02.flac, Track03.flac, ...
- Modify src/fheroes2/agg/mus.cpp GetString: under OGG_MUSIC_TYPE::WIN_VERSION, replace the last string ".ogg" with ".flac" (line 98)

Of course, this is just a quick & dirty demonstration, there might be a cleaner way to make this work. Also, SDL FLAC dependency might need to be stated somewhere.

### Additional info

This package includes town themes from both Succession Wars and Price of Loyalty. PoL themes are the extra 6 tracks at the end. It would be great to have a config option to switch between them. GOG DOS version provides a similar option to choose music. Coolest would be to add a third option to select tracks automatically depending on original/expansion map played.

File names in the package are the following:

> Heroes of Might and Magic II OST - 01 - Main Menu (Midi Version).mp3
> Heroes of Might and Magic II OST - 02 - Battle I.flac
> Heroes of Might and Magic II OST - 03 - Battle II.flac
> Heroes of Might and Magic II OST - 04 - Battle III.flac
> Heroes of Might and Magic II OST - 05 - Town - Sorceress.flac
> Heroes of Might and Magic II OST - 06 - Town - Warlock.flac
> Heroes of Might and Magic II OST - 07 - Town - Necromancer.flac
> Heroes of Might and Magic II OST - 08 - Town - Knight.flac
> Heroes of Might and Magic II OST - 09 - Town - Barbarian.flac
> Heroes of Might and Magic II OST - 10 - Town - Wizard.flac
> Heroes of Might and Magic II OST - 11 - Terrain - Lava.flac
> Heroes of Might and Magic II OST - 12 - Terrain - Wasteland.flac
[...]

PR #5095, opened 2022-03-09
Added search for FLAC files if Ogg files aren't found

Kreeblah:
This adds a search for FLAC files if Ogg files aren't found.  There don't seem to need to be any other changes, as SDL_mixer takes care of playback, though it does depend on SDL_mixer being compiled with FLAC support.

### DIFF
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -22,6 +22,11 @@
  ***************************************************************************/
 
 #include <algorithm>
+
+#if !defined( __APPLE__ ) || !defined( __MACH__ )
+#include <cassert>
+#endif
+
 #include <cctype>
 #include <cmath>
 #include <cstdlib>
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -417,6 +417,8 @@ namespace fheroes2
 
     void replaceStringEnding( std::string & output, const char * originalEnding, const char * correctedEnding )
     {
+        assert( originalEnding != nullptr && correctedEnding != nullptr );
+
         const size_t originalEndingSize = strlen( originalEnding );
         const size_t correctedEndingSize = strlen( correctedEnding );
         if ( output.size() < originalEndingSize ) {
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -22,11 +22,7 @@
  ***************************************************************************/
 
 #include <algorithm>
-
-#if !defined( __APPLE__ ) || !defined( __MACH__ )
 #include <cassert>
-#endif
-
 #include <cctype>
 #include <cmath>
 #include <cstdlib>
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -414,4 +414,21 @@ namespace fheroes2
 
         return ~crc;
     }
+
+    void replaceStringEnding( std::string & output, const char * originalEnding, const char * correctedEnding )
+    {
+        const size_t originalEndingSize = strlen( originalEnding );
+        const size_t correctedEndingSize = strlen( correctedEnding );
+        if ( output.size() < originalEndingSize ) {
+            // The original string is smaller than the ending.
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


src/fheroes2/agg/agg.cpp
```diff
@@ -550,22 +550,35 @@ void AGG::PlayMusicInternally( const int mus, const MusicSource musicType, const
         std::string filename = Settings::GetLastFile( prefix_music, MUS::GetString( mus, MUS::OGG_MUSIC_TYPE::DOS_VERSION ) );
 
         if ( !System::IsFile( filename ) ) {
-            filename = Settings::GetLastFile( prefix_music, MUS::GetString( mus, MUS::OGG_MUSIC_TYPE::WIN_VERSION ) );
+            StringReplace( filename, ".ogg", ".flac" );
```
ihhub, line 553: Since we increased the number of operations at strings I am proposing to create a new function which will be used only for such cases.
```cpp
void replaceStringEnding( std::string & output, const char * originalEnding, const char * correctedEnding )
{
    const size_t originalEndingSize = strlen( originalEnding );
    if ( output.size() < originalEndingSize ) {
        // The original string is smaller than the ending.
        return;
    }

    if ( memcmp( output.data() + output.size() - originalEndingSize, originalEnding, originalEndingSize ) != 0 ) {
         // The string does not have the required ending.
         return;
    }

    output.replace( output.size() - originalEndingSize, originalEnding, originalEndingSize, correctedEnding );
}
````
and then use this function for replacing the ending of files. The function should be declare in **tools.h** file. I am not sure that I wrote the function properly as I did this in GitHub editor :)

What do you think?

**Review by ihhub — COMMENTED** [review]
Hi @Kreeblah , thanks for creating this pull request. I left one code change proposal. Could you please take a look at it?

File: src/fheroes2/agg/agg.cpp (hunk 1 above)
Comment (Kreeblah) [reply src/fheroes2/agg/agg.cpp]: That makes a lot of sense.  I just updated my request with a method matching that signature in `tools.h`.  It just needed a slight change to the parameters passed in to `output.replace()`.

I did, however, make one additional change.  I'd missed adding calls to `GetLastFile()` in my original commit, which seems to be required for the files to actually be loaded.

File: src/engine/tools.cpp [hunk 2]
```diff
@@ -414,4 +414,21 @@ namespace fheroes2
 
         return ~crc;
     }
+
+    void replaceStringEnding( std::string & output, const char * originalEnding, const char * correctedEnding )
+    {
+        const size_t originalEndingSize = strlen( originalEnding );
+        const size_t correctedEndingSize = strlen( correctedEnding );
```
ihhub, line 421: I would also recommend to add these assertions just in case:
```cpp
assert( originalEnding != nullptr && correctedEnding != nullptr );
```
so we can catch some incorrect calls of this function.

ihhub (approved):
Hi @Kreeblah , everything looks good for me. I added one more optional suggestion. It is up to you to decide.

zenseii:
I believe this PR should close this issue:
https://github.com/ihhub/fheroes2/issues/4233

src/engine/tools.cpp
Kreeblah: That makes sense to me.  I just made another commit with that check prior to the local variable declarations and `strlen()` calls.

src/engine/tools.cpp
```diff
@@ -414,4 +414,23 @@ namespace fheroes2
 
         return ~crc;
     }
+
+    void replaceStringEnding( std::string & output, const char * originalEnding, const char * correctedEnding )
+    {
+        assert( originalEnding != nullptr && correctedEnding != nullptr );
```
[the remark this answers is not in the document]
ihhub: Hi @Kreeblah , please add `#include <cassert>` to fix the compilation on g++.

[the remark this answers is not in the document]
Kreeblah: Ah, that's a g++ thing?  I figured it was a Mac/other platforms thing.  I'll update the commit I just made, then, with some platform ifdef guards.

ihhub:
@Kreeblah , thank you so much for your contribution to the project!

PR #5103, opened 2022-03-09
Code cleanup

ihhub:
- make many obvious methods inline
- fix some compilation warnings on Visual Studio (yes, static_cast is only the way)
- remove some redundant experimental options

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -663,7 +663,7 @@ bool Heroes::Recruit( const int col, const fheroes2::Point & pt )
 
     Kingdom & kingdom = world.GetKingdom( col );
 
-    if ( !kingdom.AllowRecruitHero( false, 0 ) ) {
+    if ( !kingdom.AllowRecruitHero( false ) ) {
         return false;
     }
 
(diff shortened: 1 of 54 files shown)


src/fheroes2/heroes/heroes.cpp
```diff
@@ -760,13 +760,7 @@ void Heroes::ReplenishSpellPoints()
 
     // in castle?
     if ( castle && castle->GetLevelMageGuild() ) {
-        // restore from mage guild
-        if ( Settings::Get().ExtCastleGuildRestorePointsTurn() ) {
-            curr += maxp * GameStatic::GetMageGuildRestoreSpellPointsPercentDay( castle->GetLevelMageGuild() ) / 100;
```
oleg-derevenetz, line 765: I suppose we can now remove the `GameStatic::GetMageGuildRestoreSpellPointsPercentDay()` (and corresponding static data from the `game_static.cpp`) as well.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one proposal for further cleanup, please take a look when it's convenient for you.

oleg-derevenetz (approved):

PR #5109, opened 2022-03-11
Add MP3 file support

### LINKED ISSUE #5063: Heroes 2 Compendium version .mp3 music not detected by fheroes2
### Preliminary checks

### Platform

Windows

### Describe the bug

A user reported on the project's discord channel that they only get midi sound even after having copied the .mp3 audio from their Heroes 2 Compendium CD to the music folder inside the fheroes2 folder. They've tried to name the audio files the same as the GOG audio files.

### Save file

None

### Additional info
They're on Windows 7 64bit and have tried both the SDL1 and 2 latest builds in the .zip archive/not installer.

ihhub:
Reworked the way how we read music files. Previous implementation was slow and incorrect. We were expecting ogg files to be present while searching flac files.

Expanded the support of file types to MP3 ( close #5063 ).

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -133,8 +133,8 @@ void Audio::Init()
         const int initializationFlags = MIX_INIT_FLAC | MIX_INIT_MOD | MIX_INIT_MP3 | MIX_INIT_OGG;
         const int initializedFlags = Mix_Init( initializationFlags );
         if ( ( initializedFlags & initializationFlags ) != initializationFlags ) {
-            DEBUG_LOG( DBG_ENGINE, DBG_TRACE, "Expected music initialization flags as " << initializationFlags << " but received "
-                       << ( initializedFlags & initializationFlags ) )
+            DEBUG_LOG( DBG_ENGINE, DBG_TRACE,
+                       "Expected music initialization flags as " << initializationFlags << " but received " << ( initializedFlags & initializationFlags ) )
         }
 #endif
         hardware.freq = 22050;
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -133,7 +133,7 @@ void Audio::Init()
         const int initializationFlags = MIX_INIT_FLAC | MIX_INIT_MOD | MIX_INIT_MP3 | MIX_INIT_OGG;
         const int initializedFlags = Mix_Init( initializationFlags );
         if ( ( initializedFlags & initializationFlags ) != initializationFlags ) {
-            DEBUG_LOG( DBG_ENGINE, DBG_TRACE,
+            DEBUG_LOG( DBG_ENGINE, DBG_WARN,
                        "Expected music initialization flags as " << initializationFlags << " but received " << ( initializedFlags & initializationFlags ) )
         }
 #endif
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -130,7 +130,12 @@ void Audio::Init()
 
     if ( fheroes2::isComponentInitialized( fheroes2::SystemInitializationComponent::Audio ) ) {
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
-        Mix_Init( MIX_INIT_OGG | MIX_INIT_MP3 | MIX_INIT_MOD );
+        const int initializationFlags = MIX_INIT_FLAC | MIX_INIT_MOD | MIX_INIT_MP3 | MIX_INIT_OGG;
+        const int initializedFlags = Mix_Init( initializationFlags );
+        if ( ( initializedFlags & initializationFlags ) != initializationFlags ) {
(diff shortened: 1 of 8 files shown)


ihhub:
Hi @Kreeblah, please take a look at this pull request.

Kreeblah:
That's definitely an improved implementation and simplifies things quite a bit.  I do have a couple of thoughts on it, though they're more around organization than specific code.

First, since it's easier now to add a new format or re-order the existing formats, it might be worth changing the search order to FLAC, then Ogg, then MP3.  My thinking in this is that if somebody's gone through the trouble of creating FLAC files, they're likely to be making a deliberate choice about the audio quality vs. space tradeoff, so it would make sense to give them the highest-quality audio available.  Then, for the lossy formats, if somebody has both, Ogg has a bit-for-bit quality advantage over MP3.  Obviously, this assumes that both the Ogg and MP3 files were encoded at the same bitrates, but barring testing each file and making a determination based on the encoding characteristics (which seems like way too much effort here), having Ogg take precedence over MP3 seems like a good rule of thumb.

Second, with the methods here no longer making an assumption that the music is encoded in Ogg format, and requiring a file extension to be passed in, it might make sense to also rename the `OGG_MUSIC_TYPE` enum to something that doesn't imply an inherent Ogg format.

src/fheroes2/agg/agg.cpp
```diff
@@ -46,6 +46,67 @@
...
+        for ( const std::string & filename : possibleFilenames ) {
+            if ( System::IsFile( filename ) ) {
+                return filename;
+            }
+        }
+
+        // None of OGG files found. Try MP3 files.
+        for ( std::string & filename : possibleFilenames ) {
+            fheroes2::replaceStringEnding( filename, ".ogg", ".mp3" );
+
```
oleg-derevenetz, line 91: I suppose we should either remove this empty line or add an empty line between L99 and L100 below :)

src/fheroes2/agg/agg.cpp
```diff
@@ -547,54 +608,28 @@ void AGG::PlayMusicInternally( const int mus, const MusicSource musicType, const
...
-                        }
-                    }
-                }
-            }
+        if ( filename.empty() ) {
+            filename = getExternalMusicFile( mus, MUS::OGG_MUSIC_TYPE::MAPPED, musicDirectories );
         }
 
-        if ( !filename.empty() ) {
+        if ( filename.empty() ) {
```
oleg-derevenetz, line 622: I suppose we can just call the `Music::Play()` if filename is not empty and return. In this case we can remove the `isSongFound` completely.

src/fheroes2/agg/agg.cpp
```diff
@@ -547,54 +608,28 @@ void AGG::PlayMusicInternally( const int mus, const MusicSource musicType, const
...
+            DEBUG_LOG( DBG_ENGINE, DBG_WARN, "Cannot find a file for " << mus << " track." )
+        }
+        else {
             Music::Play( filename, loop );
             isSongFound = true;
 
             Game::SetCurrentMusic( mus );
+
+            DEBUG_LOG( DBG_ENGINE, DBG_TRACE, MUS::getFileName( mus, MUS::OGG_MUSIC_TYPE::MAPPED, ".ogg" ) )
```
oleg-derevenetz, line 631: While just not print the `filename` here? The resulting music file is not necessary an OGG file, it might as well be MP3 or FLAC.

src/engine/audio.cpp
```diff
@@ -130,7 +130,12 @@ void Audio::Init()
 
     if ( fheroes2::isComponentInitialized( fheroes2::SystemInitializationComponent::Audio ) ) {
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
-        Mix_Init( MIX_INIT_OGG | MIX_INIT_MP3 | MIX_INIT_MOD );
+        const int initializationFlags = MIX_INIT_FLAC | MIX_INIT_MOD | MIX_INIT_MP3 | MIX_INIT_OGG;
+        const int initializedFlags = Mix_Init( initializationFlags );
+        if ( ( initializedFlags & initializationFlags ) != initializationFlags ) {
+            DEBUG_LOG( DBG_ENGINE, DBG_TRACE,
```
oleg-derevenetz, line 136: I'd use the `DBG_WARN` here instead of the `DBG_TRACE`, because it's serious inconsistency.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions here, please take a look when you have a free time.

ihhub:
Hi @Kreeblah and @oleg-derevenetz , I addressed all given comments. Please take a look again.

Kreeblah:
@ihhub - That looks good to me!  :^)

oleg-derevenetz (approved):

PR #5112, opened 2022-03-11
Do not render extra frame while playing an audio only video file

ihhub:
We should not waste resources to do unnecessary work for rendering.

### DIFF
--- a/src/fheroes2/game/game_video.cpp
+++ b/src/fheroes2/game/game_video.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   Free Heroes of Might and Magic II: https://github.com/ihhub/fheroes2  *
- *   Copyright (C) 2020                                                    *
+ *   Copyright (C) 2020 - 2022                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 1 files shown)


src/fheroes2/game/game_video.cpp
```diff
@@ -40,6 +40,15 @@ namespace
         fheroes2::DrawRect( image, roi, color );
         fheroes2::DrawRect( image, fheroes2::Rect( roi.x - 1, roi.y - 1, roi.width + 2, roi.height + 2 ), color );
     }
+
+    void playAudio( const std::vector<std::vector<uint8_t>> & audioChannels )
+    {
+        for ( const std::vector<uint8_t> & audio : audioChannels ) {
+            if ( !audio.empty() ) {
+                Mixer::Play( &audio[0], static_cast<uint32_t>( audio.size() ), -1, false );
```
oleg-derevenetz, line 48: The last two arguments of `Mixer::Play()` are `-1` and `false` by default, so we can just omit them.

src/fheroes2/game/game_video.cpp
```diff
@@ -76,6 +85,17 @@ namespace Video
         if ( video.frameCount() < 1 ) // nothing to show
             return 0;
 
+        const std::vector<std::vector<uint8_t>> & audioChannels = video.getAudioChannels();
+        const bool hasAudio = Audio::isValid() && !audioChannels.empty();
+        if ( action == VideoAction::IGNORE_VIDEO ) {
+            // Since no video is rendered play audio if available.
```
oleg-derevenetz, line 91: Just curious - what is the purpose of playing the soundtrack from the video while we are not playing the video itself?

oleg-derevenetz (commented):
Hi @ihhub I just left one optional suggestion and one question, please take a look when it's convenient for you.

src/fheroes2/game/game_video.cpp
ihhub: We have two problems:
- there are video files that do not have any reasonable video content
- there are video files that suppose to be played in tandem with other video sequences (like glowing crystal in campaigns) but I haven't implemented the proper way to do it

ihhub:
Hi @oleg-derevenetz , I addressed your given comment. Please take a look.

oleg-derevenetz (approved):

PR #5120, opened 2022-03-13
Fix Town Gate spell logic

### LINKED ISSUE #4771: Adventure map, Town Gate spell shouldn't be proceed if hero casts it from the castle
### Preliminary checks

### Platform

Windows

### Describe the bug

While casting "Town Gate" spell standing inside a town hero makes teleportation  inside town wasting spell points to nowhere.
While in the OG pop up message appears.
image
I also want to suggest to change a bit message to: "You're already in a town. Spell failed".

### Save file

[Uploading Town Gate.zip…]()

### Additional info

_No response_

ihhub:
close #4771


ihhub:
Hi @Branikolog and @LeHerosInconnu, please test these changes.

oleg-derevenetz:
Hi @ihhub Currently, if spell book is opened from the hero's screen, some spells can be grayed out, like here:

SpellBook

I believe there is some inconsistency, because it's not entirely clear for the user why they are grayed out. If they are grayed out because they cannot be cast, then, for example, battle spells should be grayed out as well, but that's not the case.

Branikolog:
Hi.
I agree with @oleg-derevenetz . In a case we browse book through hero inventory no need to "gray" spells names.

LeHerosInconnu:
Hello @ihhub, @oleg-derevenetz and @Branikolog,

> Hi @Branikolog and @LeHerosInconnu, please test these changes.

It's okay, the hero cannot cast the spell if the nearest town is already occupied.

In fheroes2:

Town Gate 001

Town Gate 002

(If the player attempts to cast the spell anyway, this window is displayed.)
And I think it's important to indicate ", so you can't cast the spell." to the player in this window, as it is in the original game.
Also I will add a blank separating line between the sentences:

"The nearest town is Hillstone.

This town is occupied by your hero Circas, so you can't cast the spell."

Town Gate 003

In original game:

Town Gate 006

> Hi @ihhub Currently, if spell book is opened from the hero's screen, some spells can be grayed out, like here:
> 
> SpellBook
> 
> I believe there is some inconsistency, because it's not entirely clear for the user why they are grayed out. If they are grayed out because they cannot be cast, then, for example, battle spells should be grayed out as well, but that's not the case.

> Hi. I agree with @oleg-derevenetz . In a case we browse book through hero inventory no need to "gray" spells names.

Adventure and combat spells that cannot be cast (i.e. on the adventure map and in combat) are already grayed out in the original game when the hero consults the Magic Book.
It should stay that way.
It's very practical.
This directly indicates, for example, the spells that could be cast during a combat and this even before engaging in combat.

In original game:

Town Gate 004

Town Gate 005

oleg-derevenetz:
Hi @LeHerosInconnu

> Adventure and combat spells that cannot be cast (i.e. on the adventure map and in combat) are already grayed out in the original game when the hero consults the Magic Book.

AFAIK they are grayed out only if hero could not cast them due to the lack of mana, like on your screenshots. But with Town Gate it's not the case - currently it is grayed out because hero can't cast it, while battle spells are not grayed out even though hero cannot cast them right now (because he is not participating in battle). This looks misleading.

LeHerosInconnu:
Hello @oleg-derevenetz,

> @LeHerosInconnu
> 
> > If all combat spells were grayed out when not in combat, the information will no longer be available.
> 
> That's why my proposal is to not gray out ANY spells when spell book is opened from the hero's inventory, because neither of those spells can be casted anyway in this case. Spells can be grayed out only when they may be really casted (when hero's spell book is opened during the battle or from the adventure map).

What I mean with the spells grayed out in the Magic Book at any point in the game is that the player can take into consideration that some adventure and combat spells are not available (for whatever reason) and make decisions based on that before deciding on the hero's movement on the adventure map and before engaging in combat.
This is really handy, and would be a plus compared to the original game which only grays spells according to the number of spell points.

ihhub:
Hi @oleg-derevenetz , @LeHerosInconnu and @Branikolog , I updated the popup window and added an empty line as requested. I am proposing to make graying / not graying decision to be out of this pull request as we have debates about it. My thinking is that a spell which cannot be performed by other reasons than not enough mana must be visually separated from others. For example, we can gray-out the spell icon too.

Honestly, I am okay with both ways.

oleg-derevenetz (approved):

LeHerosInconnu:
Hello @ihhub,

So, I checked this out, here are my remarks.

When the hero tries to cast the spell on the adventure map anyway.

In fheroes2:

This one is okay:
Town Gate spell 03

For this one, I suggest:
Town Gate spell 01

"The nearest town is Hillstone.

Your hero Circas is already in Hillstone, so you can't cast the spell."

Or:

"The nearest town is Hillstone.

Your hero Circas is already in the nearest town, so you can't cast the spell."

Or:

"The nearest town is Hillstone.

Your hero is already in Hillstone, so you can't cast the spell."

Or:

"The nearest town is Hillstone.

Your hero is already in the nearest town, so you can't cast the spell."

For this other one, I suggest:
Town Gate spell 02

"The nearest town is Hillstone.

This town is occupied by your hero Liliam, so you can't cast the spell."

In the Magic Book in the hero's screen.

In fheroes2:

This one is okay:
Town Gate spell 04

For this one, I suggest (the player does not have any town or castle. I think it should also be indicated in this case):
Town Gate spell 05

"Returns the caster to any town or castle currently owned.

You do not currently own any town or castle."

[0 9936.zip](https://github.com/ihhub/fheroes2/files/8342306/0.9936.zip)

Version 0.9.13 5161

LeHerosInconnu:
@ihhub,

> Hi @oleg-derevenetz , @LeHerosInconnu and @Branikolog , I updated the popup window and added an empty line as requested. I am proposing to make graying / not graying decision to be out of this pull request as we have debates about it. My thinking is that a spell which cannot be performed by other reasons than not enough mana must be visually separated from others. For example, we can gray-out the spell icon too.
> 
> Honestly, I am okay with both ways.

For me, the grayed text of the spell is enough.
The player can get the explanation by right clicking on the spell graphic.
And it would be a shame to renounce the beautiful colored graphics of the spells. :)

PR #5126, opened 2022-03-14
Italian language selection

zenseii:
This PR makes it possible to choose the Italian language from the Language selection options. Accents etc will be added in a separate PR.


ihhub:
Hi @zenseii , everything looks fine but I would insist to add a real code to at least generate by copy-pasting some letters like it was done for French language at first. In this case users will have all letters even if some of them are wrong.

files/lang/it.po
```diff
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: fheroes2\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2022-02-17 08:59+0100\n"
-"PO-Revision-Date: 2021-11-23 23:57+0100\n"
+"PO-Revision-Date: 2022-03-16 19:22+0100\n"
```
ihhub, line 11: Could we please revert the changes in this file as they do absolutely nothing? :)

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1401,6 +1401,230 @@ namespace
...
+    void generateItalianAlphabet()
+    {
+        // Resize fonts.
+        for ( const int icnId : { ICN::FONT, ICN::SMALFONT } ) {
+            _icnVsSprite[icnId].resize( 96 );
+        }
+
+        // Normal font.
+        {
+            std::vector<fheroes2::Sprite> & font = _icnVsSprite[ICN::FONT];
```
ihhub, line 1413: Isn't this code exactly the same as for French alphabet? If yes why do we even generate it?

ihhub (commented):
Hi @zenseii , I left just 2 comments here for further discussion. Could you please take a look?

zenseii:
Hi @ihhub ! I've now made the first batch of small and normal fonts. All letters with diacritics are included and have been tested. I should still make a change and try to create a diacritic like grave accent `` ` `` once instead of for every special character in the small fonts. Many of the same diacritics are found in French, although some where missing and I have added them in Italian like the `È` and  `É`.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1401,6 +1401,251 @@ namespace
...
+            _icnVsSprite[icnId].resize( 96 );
+            // Italian uses CP1252 for special characters so we need to extend the array.
+            _icnVsSprite[icnId].insert( _icnVsSprite[icnId].end(), 160, _icnVsSprite[icnId][0] );
+        }
+
+        // Normal font.
+        {
+            std::vector<fheroes2::Sprite> & font = _icnVsSprite[ICN::FONT];
+
+            font[6].resize( font[85].width(), font[85].height() + 3 );
```
ihhub, line 1417: This letter is not generated for small font.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1401,6 +1401,251 @@ namespace
...
+
+            // U with grave accent `.
+            font[217 - 32].resize( font[53].width(), font[53].height() + 3 );
+            font[217 - 32].reset();
+            fheroes2::Copy( font[53], 0, 0, font[217 - 32], 0, 3, font[53].width(), font[53].height() );
+            font[217 - 32].setPosition( font[53].x(), font[53].y() - 3 );
+            fheroes2::Copy( font[6], 4, 0, font[217 - 32], 5, 0, 4, 2 );
+            fheroes2::updateShadow( font[217 - 32], { -1, 2 }, 2 );
+
+            // a with grave accent `.
```
ihhub, line 1469: Based on provided by you screenshots Italian font contains small letters at positions 96 - 100 (Discord server, translations channel, 18th March 2022). Why are they different here?

ihhub (changes requested):
Hi @zenseii , I left few comments here. Also with these changes do you see that the description of maps is still valid? As we found out Italian font doesn't have letters related to CP1252.

src/fheroes2/agg/agg_image.cpp
zenseii: Hi @ihhub ! Basically I found out that the game does indeed use CP1252 in the code, even if the resources as seen in the screenshot you mentioned are from 96-100. Here is that screenshot:
https://cdn.discordapp.com/attachments/903468433402900480/954119761598513152/unknown.png

However, I could not find any place where apart from UI where these where used, so in no Italian original map files. I managed to show these letters in the original game by using a hexeditor and inserting these special characters, using their CP1252 values. I then did another test by editing the game in ram using cheat engine and found that I could show the special letters using CP1252 values for them.

In other words, I'm using these values in order to support the text in the Italian map files. I haven't actually found any instance of them in any maps, but I haven't looked through all the text in all map and campaign files.

The logical choice for me was to use CP1252 values to maintain support for the original game in case any maps use the special characters.

Hex-edited CP1252 values in map description:
image


Original game showing this description (note the original game's resources does not contain the special characters `ÀÈÉÌÒÙé` which is why they are all replaced with `ù`):
image

src/fheroes2/agg/agg_image.cpp
zenseii: This letter is generated to create the grave accent `` ` ``. This accent was a pure copy paste from generateFrenchAlphabet(). This is why font[6] is copied for all the characters with grave accent `` ` ``.

zenseii: But yes, I should  just generate the grave accent once for small font and then use that for every special character, just like done for normal font by setting it to font[6].

src/fheroes2/agg/agg_image.cpp
ihhub: Got it. Thanks for a detailed explanation!

src/fheroes2/agg/agg_image.cpp
ihhub: Then why do we need font[6] in first place if we could generate it for letter 192-32 straight away?

zenseii: We could wouldn't that mean that we would have to copy all this code below for every character that uses the grave accent `` ` ``

```cpp
            fheroes2::Copy( font[85], 0, 0, font[6], 0, 3, font[85].width(), font[85].height() );
            fheroes2::Copy( font[6], 2, 3, font[6], 4, 0, 1, 1 );
            fheroes2::Copy( font[6], 2, 4, font[6], 5, 0, 1, 1 );
            fheroes2::Copy( font[6], 2, 3, font[6], 5, 1, 1, 1 );
            fheroes2::Copy( font[6], 2, 4, font[6], 6, 1, 1, 1 );
            fheroes2::Copy( font[6], 3, 3, font[6], 7, 1, 1, 1 );   
```

Instead of just repeating this line:
```cpp
            fheroes2::Copy( font[6], 4, 0, font[192 - 32], 6, 0, 4, 2 );
```

ihhub: No, We can apply the change for 192-32 and then copy from 192-32 for other characters.

zenseii: Aha I see. So generate it for `À` in this case, and then copy only the `` ` `` part of that `font[192-32]`.

ihhub: That's right. We will avoid having an extra image.

zenseii: I'll try to have it implemented by this evening. One question I have is if the x and y coordinates for copying from a sprite are the usual where 0,0 is in the upper left corner?

ihhub: Yes, it is a top-left corner.

zenseii:
@ihhub I have addressed your comments and this PR is ready for review when you have the opportunity. I would like to ask that you look at the generation of the small font lower case `ì` and check if it is done in an acceptable way. 

I also fixed the issue of the it.po having a `\v` as one user reported giving a warning during compilation.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1401,6 +1401,229 @@ namespace
...
+            fheroes2::Copy( font[69], 0, 0, font[232 - 32], 0, 3, font[69].width(), font[69].height() );
+            font[232 - 32].setPosition( font[69].x(), font[69].y() - 3 );
+            fheroes2::Copy( font[192 - 32], 7, 0, font[232 - 32], 3, 0, 4, 2 );
+            fheroes2::updateShadow( font[232 - 32], { -1, 2 }, 2 );
+
+            // i with grave accent `.
+            font[236 - 32] = font[73];
+            fheroes2::FillTransform( font[236 - 32], 0, 0, font[236 - 32].width(), 2, 1 );
+            fheroes2::Copy( font[192 - 32], 7, 0, font[236 - 32], 1, 0, 4, 2 );
+            fheroes2::updateShadow( font[236 - 32], { -1, 1 }, 2 );
```
ihhub, line 1481: I think this is a mistake: shadow offset must be the same for every letter.

ihhub (changes requested):
Hi @zenseii , I left one more comment. Could you please take a look?

src/fheroes2/agg/agg_image.cpp
zenseii: Yes, you're right, it's a human error 😅

ihhub:
@zenseii , huge thanks for your changes!

PR #5137, opened 2022-03-19
Add libmpg DLL into Windows packages

zenseii:
This library for MP3 playback was missing for Windows similarly to for the FLAC files.

Slightly unrelated to this PR:
There is some quite obvious popping sounds when using the MP3 files, most noticable during opera singing. No popping is present when using FLAC or OGG, and neither when playing the MP3 files outside of fheroes2.


ihhub:
Hi @zenseii , could you please explain what problem or missing feature we solve by adding this library into our package? We don't play MPG files as the game has a special video format.

ihhub:
@zenseii , thanks a lot for this change!

I think there are 2 possible root causes for MP3 playback:
- `void Audio::Init()` function: we set frequency and format different from MP3
- `void Music::Play()` function: we call `Mix_LoadMUS_RW` function which is supports **Ogg and MikMod specific currently**. We should use overloaded method where the type of music is specified.

oleg-derevenetz:
Hi @zenseii do these popping sounds appear at the beginning of the playback or during the entire playback?

zenseii:
> Hi @zenseii do these popping sounds appear at the beginning of the playback or during the entire playback?

Hi @ihhub and @oleg-derevenetz . The sounds are usually not there in the beginning or if they are then they are not very obvious. I will have to make a recording of it, but my impression is that all tracks start getting them after having played a while, both adventure map and castle tracks.

PR #5148, opened 2022-03-20
Build player influence map to inform AI kingdom development decisions

### LINKED ISSUE #5046: AI should upgrade magic guild more often
### Preliminary checks

The end game of heroes 2 is sometimes a bit to easy. Although the game is notoriously hard on some maps.
I was watching a stream and this expert player suggested that the AI should focus more on building the magic guild.

If the AI has more spells the end-game will be harder because the spells are so good in this game.

### Additional info

I know this is a balance change but since you changed the AI I thought this was reasonable as well.

idshibanov:
Fixes #5046 .

AI players will upgrade magic guild more often when there's heroes with appropriate Wisdom skill and current map region is considered to be safe.

Also includes bug fixes related to region generation and AI map scan at the beginning of the turn

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_castle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_castle.cpp
@@ -174,9 +174,6 @@ namespace AI
 
     void Normal::CastleTurn( Castle & castle, bool defensive )
     {
-        const uint32_t regionID = world.GetTiles( castle.GetIndex() ).GetRegion();
-        const RegionStats & stats = _regions[regionID];
-
         if ( defensive ) {
             Build( castle, GetDefensiveStructures() );
 
@@ -184,6 +181,9 @@ namespace AI
             OptimizeTroopsOrder( castle.GetArmy() );
         }
         else {
+            const uint32_t regionID = world.GetTiles( castle.GetIndex() ).GetRegion();
+            const RegionStats & stats = _regions[regionID];
+
             CastleDevelopment( castle, stats.safetyFactor, stats.spellLevel );
         }
     }
--- a/src/fheroes2/ai/normal/ai_normal_castle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_castle.cpp
@@ -134,7 +134,7 @@ namespace AI
         return false;
     }
 
-    bool CastleDevelopment( Castle & castle )
+    bool CastleDevelopment( Castle & castle, int safetyFactor, int spellLevel )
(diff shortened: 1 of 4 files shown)


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -92,11 +92,81 @@ namespace AI
...
+                stats.evaluated = true;
+                regionsToCheck.emplace_back( regionID, 100 );
+                ++lastPositive;
+            }
+            else {
+                stats.safetyFactor = 0;
+                stats.evaluated = false;
+            }
+        }
+        std::sort( regionsToCheck.begin(), regionsToCheck.end(), []( const auto & x, const auto & y ) { return x.second > y.second; } );
```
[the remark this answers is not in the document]
oleg-derevenetz: Yeah... It would be good to use C++17 features in this project (generic lambdas with `auto` parameters is the C++14 feature), but...

src/fheroes2/ai/normal/ai_normal_castle.cpp
```diff
@@ -166,14 +174,17 @@ namespace AI
 
     void Normal::CastleTurn( Castle & castle, bool defensive )
     {
+        const uint32_t regionID = world.GetTiles( castle.GetIndex() ).GetRegion();
+        const RegionStats & stats = _regions[regionID];
```
ihhub, line 178: We could move these 2 variables into else branch because it is the only place where they're in use.

src/fheroes2/world/world_regions.cpp
```diff
@@ -355,5 +355,13 @@ void World::ComputeStaticAnalysis()
                 reg._neighbours.insert( vec_tiles[exitIndex].GetRegion() );
             }
         }
+
+        // Fix missing references
+        for ( uint32_t adjacent : reg._neighbours ) {
+            std::set<uint32_t> & otherList = _regions[adjacent]._neighbours;
+            if ( otherList.find( reg._id ) == otherList.end() ) {
+                otherList.insert( reg._id );
```
ihhub, line 363: Since this is a set wouldn't it more efficient to just use `otherList.insert( reg._id );` without the check since set stores only unique values?

ihhub (changes requested):
Hi @idshibanov , I left few comments here. Could you please check them?

ihhub:
@idshibanov , thank you so much for these changes!

PR #5150, opened 2022-03-21
Remove memory copy while reading data from AGG file

ihhub:
We were wasting memory resources by doing nothing. Now loading of resources on low-profile machines will be even faster.

### DIFF
--- a/src/fheroes2/agg/agg.cpp
+++ b/src/fheroes2/agg/agg.cpp
@@ -392,7 +392,8 @@ bool AGG::ReadDataDir( void )
 std::vector<uint8_t> AGG::ReadChunk( const std::string & key )
 {
     if ( heroes2x_agg.isGood() ) {
-        const std::vector<uint8_t> & buf = heroes2x_agg.read( key );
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

src/fheroes2/agg/agg.cpp
```diff
@@ -392,7 +392,8 @@ bool AGG::ReadDataDir( void )
 std::vector<uint8_t> AGG::ReadChunk( const std::string & key )
 {
     if ( heroes2x_agg.isGood() ) {
-        const std::vector<uint8_t> & buf = heroes2x_agg.read( key );
+        // Make sure that the below container is not const-reference so returning it from the function will invoke a move constructor instead of copy constructor.
+        std::vector<uint8_t> buf = heroes2x_agg.read( key );
```
oleg-derevenetz, line 396: Not only should this container not be a `const` reference, it also shouldn't be a `const` itself. If it will be const, no move constructor will be used (but NRVO can be used though):

https://godbolt.org/z/bG6xYvTb4

ihhub: Great point! I will update the comments.

PR #5155, opened 2022-03-22
Optimize fighter hero priorities

idshibanov:
Adjust object values for fighter hero as of right now they are simply chasing monsters around the map/not even picking up artifacts at times.


ihhub:
@idshibanov , thank you very much for the improvements!

PR #5157, opened 2022-03-22
Reduce the number of storage I/O calls while searching for data

ihhub:
- ListDirs structure was useless. It is removed from now
- all getenv calls are cached
- GetRootDirs is cached as paths cannot be changed during the application runtime


src/fheroes2/system/settings.cpp
```diff
@@ -500,42 +500,50 @@ void Settings::SetProgramPath( const char * argv0 )
         path_program = argv0;
 }
 
-ListDirs Settings::GetRootDirs()
+const std::set<std::string> & Settings::GetRootDirs()
```
oleg-derevenetz, line 503: I believe that the `std::set` should be replaced by something that retains an order of insertion (e.g. `std::vector`), because order of directories is important here - for example, for the `Settings::GetLastFile()`.

src/fheroes2/system/settings.cpp
```diff
@@ -534,14 +541,22 @@ ListDirs Settings::GetRootDirs()
...
-    const std::string & data = System::GetDataDirectory( "fheroes2" );
-    if ( !data.empty() && ( std::find( dirs.cbegin(), dirs.cend(), data ) == dirs.cend() ) )
-        dirs.push_back( data );
+    std::string dataPath = System::GetDataDirectory( "fheroes2" );
+    if ( !dataPath.empty() ) {
+        dirs.emplace_back( std::move( dataPath ) );
+    }
+
+    // Remove all duplicates.
+    dirs.erase( std::unique( dirs.begin(), dirs.end() ), dirs.end() );
```
oleg-derevenetz, line 556: Hi @ihhub please note that `std::unique()` eliminates only CONSECUTIVE equivalent elements. If equivalent elements are not consecutive, duplication removal [will not work](https://godbolt.org/z/bv8Po1oaP).

src/fheroes2/system/settings.cpp
```diff
@@ -500,32 +500,39 @@ void Settings::SetProgramPath( const char * argv0 )
...
 
     // from app path
-    dirs.push_back( System::GetDirname( Settings::Get().path_program ) );
+    dirs.emplace_back( System::GetDirname( Settings::Get().path_program ) );
 
     // os-specific directories
-    dirs.splice( dirs.end(), System::GetOSSpecificDirectories() );
+#if defined( FHEROES2_VITA )
+    dirs.emplace_back( "ux0:app/FHOMM0002" );
+#endif
```
oleg-derevenetz, line 527: Hi @ihhub are you sure it is worth giving up on the `System::GetOSSpecificDirectories()`? Handling OS-specific things in engine looks right. We can just either make it return `std::vector` as well and add that vector to the `dirs` using `insert()` or accept `dirs` by reference as a parameter and `emplace_back()` in it (and may be call it `AppendOSSpecificDirectories()`).

src/fheroes2/system/settings.cpp
```diff
@@ -540,21 +542,18 @@ const std::vector<std::string> & Settings::GetRootDirs()
...
-    // user config directory
+    // User config directory.
     std::string configPath = System::GetConfigDirectory( "fheroes2" );
-    if ( !configPath.empty() ) {
+    if ( !configPath.empty() && std::find( dirs.begin(), dirs.end(), configPath ) == dirs.end() ) {
         dirs.emplace_back( std::move( configPath ) );
     }
 
-    // user data directory (may be the same as user config directory, so check this to avoid unnecessary work)
+    // User data directory (may be the same as user config directory, so check this to avoid unnecessary work)
```
oleg-derevenetz, line 551: Since now we always check for duplicates, this phrase in parentheses looks redundant :)

oleg-derevenetz (changes requested):
Hi @ihhub I left just one minor note regarding code comment, please take a look when possible :)

PR #5160, opened 2022-03-23
Fix typo in CMakeLists

oleg-derevenetz:
Catches my eye.


ihhub:
@oleg-derevenetz , even a tiny correction is good enough :)

PR #5161, opened 2022-03-24
Optimize GitHub Actions workflows

oleg-derevenetz:
Now we have a complex set of workflows with a couple dozen of jobs, and there are the following problems with them:

* Slow feedback. Quick but important checks (such as code style check or Clang-Tidy) may be started too late.
* Even if one of the essential checks fails, other workflows still continue to run, which is pointless.
* After updating the PR branch, the checks already running continue to run, which is pointless and slows down checks for the updated PR as well as other PRs.

This PR is designed to try to address these problems using the following techniques:

* Make feedback faster by running fast, but essential checks first, and failing fast the build in case of any issues.
* Automatically cancel already running workflows after updating the branch and before running a new workflow instance.
* Fail fast on matrix jobs.

That's how it looks like:

https://github.com/ihhub/fheroes2/actions/runs/2031938649

If the code style check fails, no other workflows will run. If MSVC build (which is relatively fast) fails, there is no point to run Make or CMake builds. Analyzers are launched first to speed up feedback.

### DIFF
--- a/.github/workflows/clang_analyzer.yml
+++ b/.github/workflows/clang_analyzer.yml
@@ -7,6 +7,7 @@ jobs:
   clang:
     name: Clang Analyzer
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
     - uses: actions/checkout@v2
     - name: Install dependencies and clang-tools
--- a/.github/workflows/clang_analyzer.yml
+++ b/.github/workflows/clang_analyzer.yml
@@ -1,8 +1,7 @@
 name: Clang Analyzer
 
 on:
-  pull_request:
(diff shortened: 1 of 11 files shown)


ihhub:
Hi @oleg-derevenetz , I also noticed that sometimes our runners are stuck for hours. Shall we add `timeout-minute` limit to prevent such? Even 30 minutes is way too long for any job not to finish.

ihhub:
@oleg-derevenetz , thanks a lot of polishing our pipeline!

oleg-derevenetz:
Hi @ihhub it would be good to update the list of required checks for PRs :)

PR #5170, opened 2022-03-24
Fix the calculation of the cost of surrender

### LINKED ISSUE #5168: Combat screen, summoned elementals should not be taken into account when calculating the surrender cost
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the summoned elementals should not be taken into account when calculating the surrender cost.

In fheroes2:

Elemental no cost surrender 001

Elemental no cost surrender 002

### Save file

Here is the save file:
[Before full bar elemental 01.zip](https://github.com/ihhub/fheroes2/files/8344308/Before.full.bar.elemental.01.zip)

### Additional info

Version 0.9.13 5161

oleg-derevenetz:
fix #5168

Also fixed an issue when resurrected (not truly resurrected) units were taken into account when calculating the cost of surrender.

### DIFF
--- a/src/fheroes2/army/army_troop.cpp
+++ b/src/fheroes2/army/army_troop.cpp
@@ -124,12 +124,12 @@ bool Troop::isEmpty( void ) const
     return !isValid();
 }
 
-payment_t Troop::GetCost( void ) const
+payment_t Troop::GetCost() const
 {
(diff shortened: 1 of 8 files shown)


ihhub:

PR #5171, opened 2022-03-25
Remove unnecessary call to play VICTORY music during Campaign intro

zenseii:
I noticed the VICTORY track getting called right after "choosing your lord" in the original campaigns for no apparent reason. You don't hear it getting played and it shouldn't be played here.


ihhub:
@zenseii , thanks a lot for finding and fixing this issue!

Issue #2918, opened 2021-03-08
End of scenario/campaign animation, show score text after dragon's fire breath and fade to black after input


LeHerosInconnu:
In the original game, after the dragon animation, the screen stays on the last frame of the animation and the player must click to exit the animation screen:
**Edit 2021/03/25.**
**This step is working correctly now.**
![vlcsnap-2021-03-08-12h20m31s420](https://user-images.githubusercontent.com/43583994/110317352-54774e80-800c-11eb-971f-4fb00dfda498.png)
![vlcsnap-2021-03-08-12h20m42s716](https://user-images.githubusercontent.com/43583994/110317359-593c0280-800c-11eb-853c-b921e62c618c.png)

When the player makes a click, there is a fade effect to switch to a black and white image:
![vlcsnap-2021-03-08-12h21m27s173](https://user-images.githubusercontent.com/43583994/110317387-61943d80-800c-11eb-8e7d-5de1ab793fc5.png)

Then a window is displayed to enter the desired name:
![vlcsnap-2021-03-08-12h21m36s492](https://user-images.githubusercontent.com/43583994/110317402-678a1e80-800c-11eb-8a20-1e8700b62804.png)

After validating the name, there is a fade effect and the score screen is displayed:
The mouse cursor is then displayed back on the screen.
![vlcsnap-2021-03-08-12h21m53s286](https://user-images.githubusercontent.com/43583994/110317427-6e189600-800c-11eb-8ad0-2d27595f76c2.png)

In original game:

https://user-images.githubusercontent.com/43583994/110316655-512f9300-800b-11eb-8eee-8d50e40438fc.mp4

In fheroes2:

https://user-images.githubusercontent.com/43583994/110316760-73291580-800b-11eb-874e-ae41aa066279.mp4

Here is the save file:
[Anim Score End mouse curso.zip](https://github.com/ihhub/fheroes2/files/6101097/Anim.Score.End.mouse.curso.zip)

ihhub:
Hi @LeHerosInconnu , could you please update this issue as we added waiting user input code.

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu , could you please update this issue as we added waiting user input code.

First post updated.

ihhub:
I'm marking this issue as high priority because this is an essential missing feature for many players.

PR #5174, opened 2022-03-27
Fix High Score Music after winning scenario

zenseii:
This plays the Victory theme when winning a scenario after the Dragon animation like OG, instead of the Main Menu theme like  it does currently, and it makes sure that the Main Menu theme continues to play if opening the High Score from the Main Menu, like in the original game.

This also future-proofs for when the Victory theme will be played while showing the Congratulations text behind the painting after the Dragon animation, but before the High Score menu is displayed.

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -816,6 +816,9 @@ fheroes2::GameMode Game::CompleteCampaignScenario( const bool isLoadingSaveFile 
 
         AGG::ResetAudio();
         Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
+        // TODO : Implement function that displays the last frame of win.smk with score
+        // and a dialog for name entry. AGG:PlayMusic is run here in order to start
+        // playing before displaying the high score.
         AGG::PlayMusic( MUS::VICTORY, true, true );
         return fheroes2::GameMode::HIGHSCORES;
     }
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -816,6 +816,7 @@ fheroes2::GameMode Game::CompleteCampaignScenario( const bool isLoadingSaveFile 
 
         AGG::ResetAudio();
         Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
+        AGG::PlayMusic( MUS::VICTORY, true, true );
         return fheroes2::GameMode::HIGHSCORES;
     }
 
--- a/src/fheroes2/game/game_over.cpp
+++ b/src/fheroes2/game/game_over.cpp
@@ -343,7 +343,7 @@ fheroes2::GameMode GameOver::Result::LocalCheckGameOver()
                 else {
                     AGG::ResetAudio();
                     Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
-                    AGG::PlayMusic( MUS::VICTORY, true, true );
+                    AGG::PlayMusic( MUS::VICTORY, true, false );
 
                     res = fheroes2::GameMode::HIGHSCORES;
 
--- a/src/fheroes2/game/game_over.cpp
+++ b/src/fheroes2/game/game_over.cpp
@@ -343,7 +343,7 @@ fheroes2::GameMode GameOver::Result::LocalCheckGameOver()
                 else {
                     AGG::ResetAudio();
                     Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
-                    AGG::PlayMusic( MUS::VICTORY, true, false );
+                    AGG::PlayMusic( MUS::VICTORY, true, true );
 
                     res = fheroes2::GameMode::HIGHSCORES;
 
--- a/src/fheroes2/game/game_over.cpp
+++ b/src/fheroes2/game/game_over.cpp
@@ -343,6 +343,7 @@ fheroes2::GameMode GameOver::Result::LocalCheckGameOver()
                 else {
                     AGG::ResetAudio();
                     Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
+                    AGG::PlayMusic( MUS::VICTORY, true, true );
 
                     res = fheroes2::GameMode::HIGHSCORES;
 
--- a/src/fheroes2/game/game_over.cpp
+++ b/src/fheroes2/game/game_over.cpp
@@ -343,6 +343,9 @@ fheroes2::GameMode GameOver::Result::LocalCheckGameOver()
                 else {
                     AGG::ResetAudio();
                     Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
+                    // TODO : Implement function that displays the last frame of win.smk and
+                    // a dialog for name entry. AGG:PlayMusic is run here in order to start playing
+                    // before displaying the high score.
                     AGG::PlayMusic( MUS::VICTORY, true, true );
 
                     res = fheroes2::GameMode::HIGHSCORES;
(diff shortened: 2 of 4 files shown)


src/fheroes2/game/game_startgame.cpp
```diff
@@ -646,7 +646,12 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     res = HumanTurn( loadedFromSave );
 
-                    // reset environment sounds and music theme at the end of the human turn
+                    // Audio has already been reset if scenario is won and WIN.SMK has played.
+                    if ( Game::CurrentMusic() == MUS::VICTORY ) {
```
oleg-derevenetz, line 650: Hi @zenseii there is a race condition. If `AGG::PlayMusic()` is called in async mode, then `Game::SetCurrentMusic()` will be called in the background thread, and there is no guarantee that it will be set to the `MUS::VICTORY` here.

zenseii: Thank you for your comment @oleg-derevenetz ! You're absolutely right. I initially thought of using the `fheroes2::GameMode` value of `res` as a condition, but it would become more complex. So the simplest approach would be to set run async to false in this case.

The best would maybe be to run `HighScores()` after `PlayMusic( MUS::VICTORY)` to avoid checking this condition I wrote before ending every single turn, but this presents it's own set of issues when leaving the high scores menu.

oleg-derevenetz: Why just not check that if `res` is `GameMode::HIGHSCORES`, then we do not reset sounds? Like this:

```cpp
// Audio has already been reset if scenario is won and WIN.SMK has played.
if ( res == fheroes2::GameMode::HIGHSCORES) {
  break;
}
```

Please also note that `WIN.SMK` could be played (and highscores screen is shown) when you won a campaign, this is handled in different place in the campaign-related code.

zenseii: Excuse me, I just saw your comment after commiting. Yes, the only issue here is that we could have `res = fheroes2::GameMode::Cancel` in the case of using the experimental option to continue playing the game after winning which is at the end of this if statement below.

```
                    AGG::PlayMusic( MUS::VICTORY, true, false );

                    res = fheroes2::GameMode::HIGHSCORES;

                    if ( conf.ExtGameContinueAfterVictory() && myKingdom.isPlay() ) {
                        if ( Dialog::YES == Dialog::Message( "", _( "Do you wish to continue the game?" ), Font::BIG, Dialog::YES | Dialog::NO ) ) {
                            continueAfterVictory = true;
                            // Game::HighScores() calls ResetResult()
                            Game::HighScores();
                            Interface::Basic & I = Interface::Basic::Get();
                            I.ResetFocus( GameFocus::HEROES );
                            I.SetRedraw( Interface::REDRAW_ALL );
                            res = fheroes2::GameMode::CANCEL;
                        }
                    }
 ```

In the case of `DIALOG::YES `we will still see the High Score before continuing to play the map, this would mean that the music would be reset since `res` would now be set to `fheroes2::GameMode::CANCEL`.

Comment (oleg-derevenetz) [reply src/fheroes2/game/game_startgame.cpp:650]: @zenseii If we continue the game after victory, this means that we do not exit to the main menu after showing the highscores (and we remain on the adventure map), so there is no reason to continue to play the `MUS::VICTORY`.

Comment (zenseii) [reply src/fheroes2/game/game_startgame.cpp:650]: @oleg-derevenetz Yes, so it would still be beneficial to run `ResetAudio()` after exiting the highscore if we continue to play, but the problem is that the music would be reset right before opening the High Scores and there would be no music if you choose to continue playing.

Comment (oleg-derevenetz) [reply src/fheroes2/game/game_startgame.cpp:650]: @zenseii

> but the problem is that the music would be reset right before opening the High Scores

Why? If we choose to continue to play, then highscores screen will be show immediately in the same piece of code you quoted above.

> and there would be no music if you choose to continue playing

There will be the terrain music on the adventure map as usual.

Comment (zenseii) [reply src/fheroes2/game/game_startgame.cpp:650]: > @zenseii
> 
> > but the problem is that the music would be reset right before opening the High Scores
> 
> Why? If we choose to continue to play, then highscores screen will be shown immediately in the same piece of code you quoted above.


There would be no music because at this point  `ResetAudio()` would be called immediately after when going back to `game_startgame.cpp` as seen below this comment.
> > and there would be no music if you choose to continue playing
> 
> There will be the terrain music on the adventure map as usual.

Yes, I realized my comment was ambiguous so I edited it. The adventure map music will indeed make sure that new audio is played, the problem is as stated above.

Comment (oleg-derevenetz) [reply src/fheroes2/game/game_startgame.cpp:650]: BTW if the problem is just that in the highscores screen should be played the `MUS::VICTORY` instead of the `MUS::MAINMENU`, then the change would be a one-liner in the `game_highscores.cpp`:

```cpp
AGG::PlayMusic( ( gameResult.GetResult() & GameOver::WINS ) ? MUS::VICTORY : MUS::MAINMENU, true, true );
```

Wouldn't this be enough?

oleg-derevenetz: @zenseii 

> There would be no music because at this point ResetAudio() would be called immediately after when going back to game_startgame.cpp as seen below this discussion.

Yes, but the `Game::HighScores()` will be called first.

zenseii: > BTW if the problem is just that in the highscores screen should be played the `MUS::VICTORY` instead of the `MUS::MAINMENU`, then the change would be a one-liner in the `game_highscores.cpp`:
> 
> ```c++
> AGG::PlayMusic( ( gameResult.GetResult() & GameOver::WINS ) ? MUS::VICTORY : MUS::MAINMENU, true, true );
> ```
> 
> Wouldn't this be enough?

@oleg-derevenetz Yes, this was my intial thought too, but this turned out to not be future-proof for when we will implement the correct behavior for the Dragon Animation, because `MUS::VICTORY` starts playing when you have the last frame of the Dragon animation still showing and the High Scores has not been shown yet and you are inputting name, as can be seen in the video in this issue:
https://github.com/ihhub/fheroes2/issues/2918

Basically ``MUS::VICTORY`` needs to be played before the High Scores menu.

zenseii: > @zenseii
> 
> > There would be no music because at this point ResetAudio() would be called immediately after when going back to game_startgame.cpp as seen below this discussion.
> 
> Yes, but the `Game::HighScores()` will be called first.

Yes, actually you are right. I don't know what I was thinking here but using `res == fheroes2::GameMode::HIGHSCORES` as the condition solves the problem. I just tested it. 
Thank you for you patience and explanation.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -816,6 +816,7 @@ fheroes2::GameMode Game::CompleteCampaignScenario( const bool isLoadingSaveFile
 
         AGG::ResetAudio();
         Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
+        AGG::PlayMusic( MUS::VICTORY, true, true );
```
oleg-derevenetz, line 819: BTW we can add the TODO comment here with the brief explanation why the `AGG::PlayMusic()` is called here and that it should be replaced by a more complex approach in future.

src/fheroes2/game/game_over.cpp
```diff
@@ -343,6 +343,7 @@ fheroes2::GameMode GameOver::Result::LocalCheckGameOver()
                 else {
                     AGG::ResetAudio();
                     Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT );
+                    AGG::PlayMusic( MUS::VICTORY, true, true );
```
oleg-derevenetz, line 346: And here.

src/fheroes2/game/game_campaign.cpp
zenseii: Yes, great idea 👍

ihhub (approved):

src/fheroes2/game/game_over.cpp
zenseii: Done!

oleg-derevenetz (approved):

ihhub:
@zenseii , huge thanks for fixing this bug!

Issue #5166, opened 2022-03-24
Combat screen, Army order bar, reduce the number of visible slots on the screen, add arrow buttons at the edges of the bar


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, I suggest to reduce the number of visible slots on the screen to a number of fourteen slots, and to add arrow buttons at the edges of the bar to display the non visible slots.

Using the left and right arrows on the keyboard also allows to display the non visible slots.

When the mouse is positioned over the area of the Army order bar and the arrow buttons at the edges of the bar, the use of the mouse wheel up and down also allows to display the non visible slots.
Wheel up to display the hidden slots on the left side of the screen and wheel down to display the hidden slots on the right side of the screen.

The slot of the creature that has the hand must always be visible in the Army order bar (except when the player is using the bar).

In fheroes2:

Reducing the number of visible slots on the screen to fourteen allows to add arrow buttons on the edges of the bar to display the non visible slots:
![Army order bar 001](https://user-images.githubusercontent.com/43583994/159983006-2e22fedb-4c3c-4e58-b793-314a7b519f14.png)

Problems occur with fifteen slots.
The left edge of the yellow frame of the first bar slot is no longer visible:
![Army order bar 003](https://user-images.githubusercontent.com/43583994/159983100-1ded3976-c522-483d-ab1e-3fdab5cf1b03.png)

With sixteen slots, the first and last slots of the bar are cut off:
![Army order bar 004](https://user-images.githubusercontent.com/43583994/159983273-6efe9bf6-4f7e-4b0d-a497-8095fd0ed017.png)

With seventeen slots, the slot of the active troop is no longer visible, and the slot of the first troop (Champion) is not grayed even though this troop has already played:
![Army order bar 006](https://user-images.githubusercontent.com/43583994/159983606-77cae22c-a299-4a06-8dc1-185e98e53b75.png)

### Save file

Here is the save file:
[Before full bar elemental 01.zip](https://github.com/ihhub/fheroes2/files/8344308/Before.full.bar.elemental.01.zip)

### Additional info

Version 0.9.13 5161

[...]

PR #5176, opened 2022-03-28
Fix army order option related crash

### LINKED ISSUE #5167: Combat screen, Army order bar, the game crashes when the action bar goes over the edge of the screen and a hero flees or surrenders
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the game crashes when the action bar goes over the edge of the screen and a hero flees or surrenders.

In fheroes2:

Army order bar 004

### Save file

Here is the save file:
[Before full bar elemental 01.zip](https://github.com/ihhub/fheroes2/files/8344308/Before.full.bar.elemental.01.zip)

### Additional info

Version 0.9.13 5161

Related issue: https://github.com/ihhub/fheroes2/issues/5166.

ihhub:
This fixes #5167 . Also for cases when the number of units to display is more than the screen the order is floating (relates to #5166 ) which means that the bar will show only limited number of units but shifts the order during the progression of the battle.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -911,23 +911,47 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
     }
 
     const int32_t validUnitCount = static_cast<int32_t>( std::count_if( orders->begin(), orders->end(), []( const Unit * unit ) { return unit->isValid(); } ) );
+    const int32_t maximumUnitsToDraw = area.width / armyOrderMonsterIconSize;
 
-    int32_t ox = area.x + ( area.width - armyOrderMonsterIconSize * validUnitCount ) / 2;
-    int32_t oy = area.y;
+    int32_t offsetX = area.x;
 
-    fheroes2::Rect::x = ox;
-    fheroes2::Rect::y = oy;
+    if ( validUnitCount > maximumUnitsToDraw ) {
+        offsetX += ( area.width - armyOrderMonsterIconSize * maximumUnitsToDraw ) / 2;
+    }
+    else {
+        offsetX += ( area.width - armyOrderMonsterIconSize * validUnitCount ) / 2;
+    }
+
+    fheroes2::Rect::x = offsetX;
+    fheroes2::Rect::y = area.y;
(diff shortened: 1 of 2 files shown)


oleg-derevenetz:
Hi @ihhub crash is not observed anymore, but there is problem with rendering:

Screenshot

Cavalry unit in upper left corner is never grayed out, although he skipped the turn.

ihhub:
> Hi @ihhub crash is not observed anymore, but there is problem with rendering:
> 
> Screenshot
> 
> Cavalry unit in upper left corner is never grayed out, although he skipped the turn.

Hi @oleg-derevenetz , I addressed it. Please take a look.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -911,23 +911,47 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
...
-            fheroes2::Rect::width += armyOrderMonsterIconSize;
+    int32_t unitsDrawn = 0;
+    int32_t unitsProcessed = 0;
+
+    for ( const Unit * unit : *orders ) {
+        if ( unitsDrawn == maximumUnitsToDraw ) {
+            break;
+        }
+
+        if ( unit == nullptr || !unit->isValid() ) {
```
oleg-derevenetz, line 939: If `unit` from `orders` is `nullptr` then something is definitely wrong - it is already assumed that it may not be `nullptr` when `validUnitCount` is initialized. I suggest to add an assertion regarding this either on `validUnitCount` initialization or both here and on `validUnitCount` initialization, and remove this condition, because it's too late to check it here anyway.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion here, could you please take a look?

src/fheroes2/battle/battle_interface.cpp
ihhub: Hi @oleg-derevenetz , I've added in both places.

oleg-derevenetz (approved):

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu, I partially addressed your suggestion for #5166 here.

I tried it, works fine. :)

PR #5177, opened 2022-03-28
Avoid extra memory reallocations while handling stream buffers

ihhub:
This significantly increases loading of translation files, save files and AGG files.


oleg-derevenetz (changes requested):
Hi @ihhub it would be nice to convince the static analyzer that there is no out-of-bounds write by adding an assertion.

oleg-derevenetz (approved):

PR #5178, opened 2022-03-28
Use move operators for Stream Buffer

ihhub:
Reduce the amount of memory being used (especially within StreamFile::toStreamBuf method).

### DIFF
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -211,7 +211,8 @@ StreamBuf::~StreamBuf()
 }
 
 StreamBuf::StreamBuf( StreamBuf && st ) noexcept
-    : itbeg( nullptr )
+    : StreamBase( std::move( st ) )
+    , itbeg( nullptr )
     , itget( nullptr )
     , itput( nullptr )
     , itend( nullptr )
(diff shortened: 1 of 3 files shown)


src/engine/serialize.cpp
```diff
@@ -210,13 +210,16 @@ StreamBuf::~StreamBuf()
         delete[] itbeg;
 }
 
-StreamBuf::StreamBuf( const StreamBuf & st )
+StreamBuf::StreamBuf( StreamBuf && st ) noexcept
```
oleg-derevenetz, line 213: Hmm, since `StreamBuf` is inherited from the `StreamBase`, shouldn't we implement and call inherited constructors and assignment operators from the `StreamBase`? I see at least `flags` field in `StreamBase`, it is not initialized here and is not swapped with `st`.

oleg-derevenetz:
Hi @ihhub we have similar move assignment operator for the `Sprite` class:

```cpp
    Sprite & Sprite::operator=( Sprite && sprite ) noexcept
    {
        if ( this != &sprite ) {
            Image::operator=( std::move( sprite ) );

            std::swap( _x, sprite._x );
            std::swap( _y, sprite._y );
        }

        return *this;
    }
```

We can do here in a similar way. The `bugprone-use-after-move` warning from Clang-Tidy is bogus here - while use after move may be bugprone, in this case it's perfectly legal.

oleg-derevenetz:
Hi @ihhub it seems that I used an ambiguous wording :) What I meant is to make the move constructor for the `StreamBuffer` class like the move constructor for the `Sprite` class, not the vice versa :) But the reverse is okay too.

ihhub:
> Hi @ihhub it seems that I used an ambiguous wording :) What I meant is to make the move constructor for the `StreamBuffer` class like the move constructor for the `Sprite` class, not the vice versa :) But the reverse is okay too.

I've reserved the code :)

oleg-derevenetz (approved):

PR #5179, opened 2022-03-29
GitHub Actions: use shallow clone for VDPM, fix broken PSV build, improve Makefiles

oleg-derevenetz:
Make the build process a bit faster, disable some warnings specific for PSV and issued for the SDL library code, use `CCFLAGS` (Common CFLAGS) for both C and C++ compilers instead of using just `CFLAGS` and removing the C++ specific options via the `filter-out` construct, add copyright headers to makefiles.

Hi @ihhub could you please remove the Clang-Tidy from the list of required checks? It does not run if there is no changes in source files, like in this PR.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -120,7 +120,7 @@ jobs:
     - name: Setup Vita SDK
       continue-on-error: true
       run: |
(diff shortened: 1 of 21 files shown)


ihhub:
@oleg-derevenetz , thanks a lot for fixing the issue!

PR #5183, opened 2022-03-30
Fix the build for PSV, add multiple warning-related compiler flags to the Makefile, remove useless -Weffc++ from the waiting list

oleg-derevenetz:
Also this PR fixes the build for PSV (once again).

In GCC, `-Weffc++` is broken and useless. For example, even the latest GCC [reports](https://godbolt.org/z/WdKKdsvWz) the following warning:

```
<source>: In constructor 'C::C()':
<source>:6:5: warning: 'C::s' should be initialized in the member initialization list [-Weffc++]
    6 |     C() {};
      |     ^
```

for the following code:

```cpp
#include <string>

class C
{
public:
    C() {};
private:
    std::string s;
};

int main()
{
    C c;
}
```

despite the fact that `std::string` has perfectly safe default constructor.

### DIFF
--- a/src/Makefile.vita
+++ b/src/Makefile.vita
@@ -21,7 +21,6 @@
 AR := arm-vita-eabi-gcc-ar
 CC := arm-vita-eabi-gcc
 CXX := arm-vita-eabi-g++
-LIBS := -lz
 
 ifdef FHEROES2_STRICT_COMPILATION
 CCFLAGS := $(CCFLAGS) -Wno-error=pedantic -Wno-error=float-equal
--- a/src/Makefile.vita
+++ b/src/Makefile.vita
@@ -33,9 +33,9 @@ LDFLAGS := $(LDFLAGS) -L$(VITASDK)/arm-vita-eabi/lib -L$(VITASDK)/lib -Wl,--sort
 
 CCFLAGS_TP := $(CCFLAGS_TP) $(ARCH)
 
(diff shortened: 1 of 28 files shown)


src/tools/82m2wav.cpp
```diff
@@ -66,7 +66,7 @@ int main( int argc, char ** argv )
         wavHeader.putLE32( 0x61746164 ); // DATA
         wavHeader.putLE32( size ); // size
 
-        fd_body.write( (const char *)wavHeader.data(), wavHeader.size() );
+        fd_body.write( reinterpret_cast<const char *>( wavHeader.data() ), wavHeader.size() );
```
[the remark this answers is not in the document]
ihhub: I think we might need to mute this warning. It makes no sense to not to use reinterpret_cast.

src/Makefile.vita
```diff
@@ -21,7 +21,6 @@
 AR := arm-vita-eabi-gcc-ar
 CC := arm-vita-eabi-gcc
 CXX := arm-vita-eabi-g++
-LIBS := -lz
```
ihhub, line 24: Why are we removing `lz` library which is zlib for PSVita?

oleg-derevenetz: @ihhub because it's already added to the `LIBS` in the main `Makefile`:

https://github.com/oleg-derevenetz/fheroes2/blob/b79005538534d0e4ff4522e3911a7858eaa3a5d9/src/Makefile#L55

ihhub: Got it!

ihhub:
@oleg-derevenetz , great job!

PR #5185, opened 2022-03-30
Clang-Tidy: disable the cppcoreguidelines-pro-type-reinterpret-cast check

oleg-derevenetz:
Requested by @ihhub

Although `reinterpret_cast` conversion is dangerous in general, it's perfectly legal to use it in certain cases (for example, when accessing a byte representation of an object by casting a pointer to it to the `char *`).


ihhub:
@oleg-derevenetz , thank you so much for this correction!

PR #5186, opened 2022-03-30
Fix regression for AGG file search

ihhub:
Case-sensitive filesystems do not support the current approach so we have to revert back to the original but with small modifications.

Regression of #5157 

Notes:
- h2d file related logic can remain as it is as we distribute the file and we know the actual name
- the same problem is present for music files but I'll fix it in a separate pull request.

### DIFF
--- a/src/fheroes2/agg/agg.cpp
+++ b/src/fheroes2/agg/agg.cpp
@@ -657,28 +657,63 @@ AGG::AGGInitializer::~AGGInitializer()
 
 bool AGG::AGGInitializer::init()
 {
-    std::string fullPath;
-    if ( Settings::findFile( "data", "heroes2.agg", fullPath ) ) {
-        if ( !heroes2_agg.open( fullPath ) ) {
-            return false;
-        }
-
-        if ( !g_midiHeroes2AGG.open( fullPath ) ) {
-            // How is it even possible that for the second time the file is not readable?
-            assert( 0 );
-            return false;
-        }
-    }
-    else {
+    const ListFiles aggFileNames = Settings::FindFiles( "data", ".agg", false );
+    if ( aggFileNames.empty() ) {
         return false;
     }
 
-    // Both AGG files must be located in the same directory. In this case we modify the ending of the existing full path to avoid extra logic.
-    fheroes2::replaceStringEnding( fullPath, ".agg", "x.agg" );
+    const std::string heroes2AggFileName( "heroes2.agg" );
+    std::string heroes2AggFilePath;
+    std::string aggLowerCaseFilePath;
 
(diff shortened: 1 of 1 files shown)


src/fheroes2/agg/agg.cpp
```diff
@@ -657,28 +657,64 @@ AGG::AGGInitializer::~AGGInitializer()
...
+        assert( 0 );
+        return false;
+    }
+
+    // Find "heroes2x.agg" file.
+    std::string heroes2XAggFilePath;
+    fheroes2::replaceStringEnding( aggLowerCaseFilePath, ".agg", "x.agg" );
+
+    for ( const std::string & path : aggFileNames ) {
+        if ( path.size() != aggLowerCaseFilePath.size() ) {
```
oleg-derevenetz, line 704: I'm not sure about this. In theory, with multibyte encodings like UTF-8, uppercase string and its lowercase version may differ in size (in bytes).

oleg-derevenetz (commented):
Hi @ihhub I just left one comment, please take a look when it's convenient for you.

src/fheroes2/agg/agg.cpp
ihhub: Hmmm... you're right!

ihhub: I've changed the condition.

src/fheroes2/agg/agg.cpp
```diff
@@ -657,28 +657,64 @@ AGG::AGGInitializer::~AGGInitializer()
...
+        assert( 0 );
+        return false;
+    }
+
+    // Find "heroes2x.agg" file.
+    std::string heroes2XAggFilePath;
+    fheroes2::replaceStringEnding( aggLowerCaseFilePath, ".agg", "x.agg" );
+
+    for ( const std::string & path : aggFileNames ) {
+        if ( path.size() < aggLowerCaseFilePath.size() ) {
```
oleg-derevenetz, line 704: @ihhub Hmm. Suppose we have some UTF-8 path like `some/set/of/multibyte/characters/heroes2x.agg` in `aggLowerCaseFilePath` and `path` is `sOmE/sEt/oF/MultIbYte/chArActeRs/heroes2x.agg`. The second string may be longer than the first one (in bytes), shorter than the first one or equal in size. No guarantees of any kind are provided. I suppose that we just shouldn't perform the comparison here at all. Any comparison should be done only after converting `path` to lowercase and `std::string::operator==()` will do this automatically.

ihhub: Oh, yeah. I lowercase letters always comes after upper case. Yes, I should remove the check.

oleg-derevenetz (approved):

PR #5187, opened 2022-03-31
Add Norwegian language option and font generation

zenseii:
This adds Norwegian as a language choice in the game and generates Norwegian fonts.

Marking as draft since I need to generate small fonts and copy the normal fonts É and é from Italian. 

One could technically say that this will generate the Danish fonts too...

image


src/fheroes2/agg/agg_image.cpp
```diff
@@ -1400,6 +1400,91 @@ namespace
...
+            font[197 - 32].setPosition( font[33].x(), font[33].y() - 5 );
+            fheroes2::updateShadow( font[197 - 32], { -1, 2 }, 2 );
+
+            // a with circle on top.
+            font[229 - 32].resize( font[65].width(), font[65].height() + 6 );
+            font[229 - 32].reset();
+            fheroes2::Copy( font[65], 0, 0, font[229 - 32], 0, 5, font[65].width(), font[65].height() );
+            fheroes2::Copy( font[197 - 32], 7, 0, font[229 - 32], 3, 0, 4, 4 );
+            font[229 - 32].setPosition( font[65].x(), font[65].y() - 5 );
+            fheroes2::updateShadow( font[229 - 32], { -1, 2 }, 2 );
```
ihhub, line 1435: To avoid the same mistake done in Italian language related pull request can we please do some changes in this file:
- create 2 functions in anonymous namespace
```cpp
void updateNormalFontLetterShadow( fheroes2::Image & letter )
{
    fheroes2::updateShadow( letter, { -1, 2 }, 2 );
}

void updateSmallFontLetterShadow( fheroes2::Image & letter )
{
    fheroes2::updateShadow( letter, { -1, 1 }, 2 );
}
```
- replace all `updateShadow` calls by newly added functions.

What do you think?

zenseii: Yes, that sounds like a great idea and avoids dumb errors. I'll make the change here then and wait for the Italian PR to get merged first and then fix it.

zenseii: Hmm I'm thinking if there's a way to call the new function at the end and update all the letters' shadows through a loop or iterating through them, or if there's a way to add all used letters at once as arguments to the function?

zenseii: I went ahead and replaced all instances of them. Not sure if implementing a way of avoiding having to repeat them is necessary or worth the time to make it.

zenseii:
Hi @ihhub ! This PR is finally ready for review when you get an available moment. There is one issue about the language choice menu because I decided to create a proper lowercase k for both fonts. 
The issue can be seen in these images, basically choosing Norwegian changes the lowercase k's used in Polish and for generating small Russian k:
image
Look at the k in Polish in particular.
image

I'm ready to try to address this myself, since this issue is not acceptable. I have a somewhat grasp on what's going on, but I haven't looked thoroughly into it.

ihhub:
Hi @zenseii , I do not observer such behavior on my English version of the game in relation to small k. It is good that you created a separate pull request for this.

ihhub:
@zenseii , excellent work! Well done on all translation related stuff!

PR #5189, opened 2022-03-31
Fix the calculation of the cost of upgrading troops

### LINKED ISSUE #5188: Troop upgrade increases gold reserves
### Preliminary checks

### Platform

Windows

### Describe the bug

Latest commit version.
Upgrading more than 1 troop game shows the wrong message (without gold picture) and after confirmation gold amount increases in kingdom resource box.
image
image
image
While upgrading 1 troop everything works as intended.
image
On released 0.9.13  version the bug does not occures. 

### Save file

[Upgrade bug.zip](https://github.com/ihhub/fheroes2/files/8387855/Upgrade.bug.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5188

Regression after #5170

The problem is that currently in many relative classes (which inherit from each other) there are non-virtual methods that have the same names, but do different things that are not very compatible with each other. The previous approach was to make the `GetCost()` and `GetUpgradeCost()` `virtual` in related classes (such as `Monster`, `Troop` and `Battle::Unit`), but this may be error-prone in current circumstances, because the wrong method may be called easily (which is what in fact happened). This PR offers the second approach - I have changed the names of these methods to different ones, I hope that this approach is less error-prone.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -657,7 +657,7 @@ namespace AI
             }
         }
         else if ( join.reason == NeutralMonsterJoiningCondition::Reason::ForMoney ) {
-            const int32_t joiningCost = troop.GetCost().gold;
+            const int32_t joiningCost = troop.GetTotalCost().gold;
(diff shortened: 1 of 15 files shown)


idshibanov (approved):

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for fixing these issues!

PR #5192, opened 2022-04-01
Makefile: add options to build the project with UBSAN, ASAN or TSAN

oleg-derevenetz:
I usually do this by patching the `Makefile`, so I decided to add options instead. [UBSAN](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)+[ASAN](https://clang.llvm.org/docs/AddressSanitizer.html) introduce relatively small runtime overhead (about 2x), so I use this combination on the everyday basis, while UBSAN+[TSAN](https://clang.llvm.org/docs/ThreadSanitizer.html) introduce significantly larger overhead (5x - 15x).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 #
 option(USE_SYSTEM_LIBSMACKER "Use system libsmacker instead of the bundled version" OFF)
 option(ENABLE_STRICT_COMPILATION "Enable strict compilation mode (turns warnings into errors)" OFF)
-option(ENABLE_IMAGE "Enable SDL Image support (requires libpng)" ON)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 #
 option(USE_SYSTEM_LIBSMACKER "Use system libsmacker instead of the bundled version" OFF)
 option(ENABLE_STRICT_COMPILATION "Enable strict compilation mode (turns warnings into errors)" OFF)
-option(ENABLE_IMAGE "Enable SDL/SDL2 Image support (requires libpng)" ON)
+option(ENABLE_IMAGE "Enable SDL Image support (requires libpng)" ON)
```
ihhub, line 38: This option is valid only for SDL 2 as SDL 1 doesn't support any image formats except Bitmap. I will update the code to clearly indicate this.

ihhub:
@oleg-derevenetz , thanks a lot for adding these options!

PR #5193, opened 2022-04-02
Remove all erroneous \v in nb.po and it.po

zenseii:
These somehow snuck their way into a couple of strings. Did a search through all .po files and these were all.

They give warnings during compilation as reported by a user on Discord.

They're caused by holding shift and pressing enter, which was done by accident.


ihhub:
Hi @zenseii , to avoid any warnings in the future it is good to treat all warnings for translations as errors. According to [logs](https://github.com/ihhub/fheroes2/runs/5788519930?check_suite_focus=true) these warnings occur during **msgmerge** run. However, based on the [number of options](https://man7.org/linux/man-pages/man1/msgmerge.1.html) for this tool I cannot find the one which is responsible of treating warnings as errors.

ihhub:
@zenseii , thanks for cleaning up translations!

PR #5197, opened 2022-04-02
Regularize preprocessor symbols, sync warning-related options in CMakeLists with Makefile

oleg-derevenetz:
* Remove scraps of the Android-related code, because we in fact currently do not support Android;
* Subordinate names of target-related preprocessor symbols to a specific system (`TARGET_XXX`). Do not define custom symbols with names starting with underscores (like `__SWITCH__`), because such names are reserved;
* Sync warning-related options in CMakeLists.txt with Makefile.


.github/workflows/make.yml
```diff
@@ -154,7 +151,7 @@ jobs:
       if: ${{ github.event_name == 'push' }}
       run: |
         cp docs/README.txt docs/README_PSV.md .
-        zip fheroes2_psv_sdl2.zip fheroes2.vpk LICENSE fheroes2.key changelog.txt README.txt README_PSV.md files/lang/*.mo files/data/*.h2d
```
ihhub, line 157: Could we please add a comment here stating that `fheroes2.vpk` package contains translation and fheroes2 data files? It took me some time to understand this change.

src/Makefile.mingw32
```diff
@@ -1,37 +0,0 @@
-###########################################################################
```
ihhub, line 1: I foresee that some people start to complain that we remove mingw support but honestly it is a correct move.

ihhub (commented):
Hi @oleg-derevenetz , I left just one suggestion here. Could you please take a look?

src/Makefile.mingw32
oleg-derevenetz: I only removed the support for `mingw32` platform, which is not used neither here:

https://github.com/ihhub/fheroes2/blob/8bbdf8e3751ca7fab6c46d50c37589ff9ab81be6/src/Makefile#L107-L128

nor here:

https://github.com/ihhub/fheroes2/blob/8bbdf8e3751ca7fab6c46d50c37589ff9ab81be6/src/dist/Makefile#L36-L42

`mingw` is still supported though. Someone could use the removed platform by running `make -C src PLATFORM=mingw32`, but I really doubt that (and I also doubt that in this case something worthwhile came out).

ihhub:
@oleg-derevenetz , thank you so much for these little but very important changes to move forward!

PR #5198, opened 2022-04-02
Initial implementation of dimension door for AI

idshibanov:
Adding basic functionality to support AI using the Dimension door spell. Path building algorithm can be polished later, but for most common cases this should be sufficient.

Also spell movement penalties are now part of the appropriate struct rather than static variables.

dimension-siege2


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1235,12 +1235,21 @@ namespace AI
...
-            _pathfinder.reEvaluateIfNeeded( *bestHero );
-            bestHero->GetPath().setPath( _pathfinder.buildPath( bestTargetIndex ), bestTargetIndex );
-
             const size_t heroesBefore = heroes.size();
 
-            HeroesMove( *bestHero );
+            // check if we want to use Dimension Door spell or move regularly
+            const std::list<Route::Step> & dimensionPath = _pathfinder.getDimensionDoorPath( *bestHero, bestTargetIndex );
+            uint32_t dimensionDoorDistance = AIWorldPathfinder::calculatePathPenalty( dimensionPath );
+            uint32_t moveDistance = _pathfinder.getDistance( bestTargetIndex );
```
oleg-derevenetz, line 1243: Hi @idshibanov shouldn't we call `_pathfinder.reEvaluateIfNeeded()` before calculating the distance here? Hero may cast the dimension door as a previous move and pathfinder cache may become invalid. Or maybe I got it wrong?

idshibanov: I think you are correct, although it was fine in all cases I tested it's safer to run it before the check anyway. Moved.

src/fheroes2/maps/maps.h
```diff
@@ -70,6 +70,7 @@ namespace Maps
     Indexes ScanAroundObject( const int32_t center, const MP2::MapObjectType objectType, const bool ignoreHeroes );
     Indexes GetFreeIndexesAroundTile( const int32_t center );
 
+    bool isValidForDimensionDoor( int32_t targetIndex, bool isWater );
```
ihhub, line 73: Small correction: these input arguments could be `const`.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -713,6 +722,82 @@ std::vector<IndexObject> AIWorldPathfinder::getObjectsOnTheWay( const int target
...
 
+std::list<Route::Step> AIWorldPathfinder::getDimensionDoorPath( const Heroes & hero, int targetIndex ) const
+{
+    std::list<Route::Step> path;
+
+    const Spell dimensionDoor( Spell::DIMENSIONDOOR );
+    if ( !hero.HaveSpell( dimensionDoor ) || !Maps::isValidAbsIndex( targetIndex ) )
+        return path;
+
+    const uint32_t spCost = std::max( 1U, dimensionDoor.SpellPoint() );
```
ihhub, line 733: I would recommend to call `dimensionDoor.SpellPoint( &hero ) instead. I know that Dimension Door spell points are fixed but how knows. Maybe in the future we will add an artifact which reduces or even increases the cost of this spell.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1684,6 +1684,30 @@ namespace AI
         }
     }
 
+    bool HeroesCastDimensionDoor( Heroes & hero, const int32_t targetIndex )
+    {
+        const Spell dimensionDoor( Spell::DIMENSIONDOOR );
+        if ( !Maps::isValidAbsIndex( targetIndex ) || hero.GetMovePoints() < dimensionDoor.MovePoint() || hero.GetSpellPoints() < dimensionDoor.SpellPoint()
```
ihhub, line 1690: The same comment as for **world_pathfinding.cpp**: it is good to call `dimensionDoor.SpellPoint( &hero )`.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -713,6 +722,82 @@ std::vector<IndexObject> AIWorldPathfinder::getObjectsOnTheWay( const int target
...
+    const uint32_t movementCost = std::max( 1U, dimensionDoor.MovePoint() );
+
+    const uint32_t maxCasts = std::min( hero.GetSpellPoints() / spCost, hero.GetMovePoints() / movementCost );
+    if ( maxCasts == 0 )
```
ihhub, line 737: Since we want to use the spell only when at least 2 casts are allowed should we change this condition to
```cpp
if ( maxCasts < 2 )
```
?
We can put the comment below about this here.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -713,6 +722,82 @@ std::vector<IndexObject> AIWorldPathfinder::getObjectsOnTheWay( const int target
...
+        targetIndex = Maps::GetDirectionIndex( targetIndex, Direction::BOTTOM );
+        if ( !Maps::isValidAbsIndex( targetIndex ) )
+            return path;
+    }
+
+    const fheroes2::Point startPoint = Maps::GetPoint( hero.GetIndex() );
+    const fheroes2::Point targetPoint = Maps::GetPoint( targetIndex );
+
+    fheroes2::Point current = startPoint;
+    fheroes2::Point difference = targetPoint - startPoint;
```
ihhub, line 750: I think we can avoid using extra variable by doing this:
```cpp
    const fheroes2::Point targetPoint = Maps::GetPoint( targetIndex );

    fheroes2::Point current = Maps::GetPoint( hero.GetIndex() );
    fheroes2::Point difference = targetPoint - current;
```

It's a very tiny optimization but might help in case of hundreds of calls.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -713,6 +722,82 @@ std::vector<IndexObject> AIWorldPathfinder::getObjectsOnTheWay( const int target
...
+
+        const int32_t anotherNodeIdx = Maps::GetIndexFromAbsPoint( another );
+        bool found = Maps::isValidForDimensionDoor( anotherNodeIdx, water );
+
+        if ( !found ) {
+            for ( size_t i = 0; i < directions.size(); ++i ) {
+                if ( !Maps::isValidDirection( anotherNodeIdx, directions[i] ) )
+                    continue;
+
+                const int newIndex = anotherNodeIdx + _mapOffset[i];
```
ihhub, line 771: Shall we put this check first before calculating **newPoint**?
```cpp
if ( !Maps::isValidForDimensionDoor( newIndex, water ) ) {
    continue;
}
```

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -713,6 +722,82 @@ std::vector<IndexObject> AIWorldPathfinder::getObjectsOnTheWay( const int target
...
+                     && Maps::isValidForDimensionDoor( newIndex, water ) ) {
+                    path.emplace_back( newIndex, currentNodeIdx, Direction::CENTER, movementCost );
+                    current = newPoint;
+                    found = true;
+                    break;
+                }
+            }
+
+            if ( !found )
+                return std::list<Route::Step>();
```
ihhub, line 783: We could use a short form here `return {};` but this is optional.

ihhub (changes requested):
Hi @idshibanov , I left multiple suggestions here. Could you please take a look?

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1235,12 +1235,21 @@ namespace AI
...
             _pathfinder.reEvaluateIfNeeded( *bestHero );
-            bestHero->GetPath().setPath( _pathfinder.buildPath( bestTargetIndex ), bestTargetIndex );
 
-            const size_t heroesBefore = heroes.size();
+            // check if we want to use Dimension Door spell or move regularly
+            const std::list<Route::Step> & dimensionPath = _pathfinder.getDimensionDoorPath( *bestHero, bestTargetIndex );
+            uint32_t dimensionDoorDistance = AIWorldPathfinder::calculatePathPenalty( dimensionPath );
+            uint32_t moveDistance = _pathfinder.getDistance( bestTargetIndex );
+            if ( dimensionDoorDistance && ( !moveDistance || dimensionDoorDistance < moveDistance / 2 ) ) {
+                HeroesCastDimensionDoor( *bestHero, dimensionPath.front().GetIndex() );
```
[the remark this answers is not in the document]
idshibanov: There's nothing to do in this case even if it returns false. I added return value for the future.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1684,6 +1684,30 @@ namespace AI
...
+             || !hero.HaveSpell( dimensionDoor ) )
+            return false;
```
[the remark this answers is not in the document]
idshibanov: If this function is called that check is least likely to be false. This is action execution, not planning. Also HaveSpell() check is slowest since it loops through artifact and spell list.

ihhub:
Hi @idshibanov , could you please rebase from master branch? I can see changes were merged recently.

ihhub (approved):

oleg-derevenetz (approved):

PR #5200, opened 2022-04-03
Play just the M82::TREASURE sound when we picked up an artifact due to an event

### LINKED ISSUE #5199: Sea event operation: multiple sounds playing
### Preliminary checks

### Platform

Windows

### Describe the bug

Two different sounds are playing while initiating sea event.
image
I suggest leaving only harp sound.
Edit: Sound is playing currently only for event containing artifact. I don't remember, if any sound is playing in the OG, but I think, for all other events (containing text or resource) we should also play some music.

### Save file

[Sea event sound.zip](https://github.com/ihhub/fheroes2/files/8404278/Sea.event.sound.zip)
Just move ship 1 cell to left.

### Additional info

_No response_

oleg-derevenetz:
fix #5199

I have no idea why we played both `M82::TREASURE` and pickup sound in this case. Let's play just `M82::TREASURE`.


ihhub:

PR #5201, opened 2022-04-03
Further Makefile improvements

oleg-derevenetz:
* Do not duplicate build options taken from platform-specific rules, allow to prevent `sdl-config` from running if it isn't needed;
* Remove unused LINK and LINK_FLAGS from `Makefile.osx`;
* `CMakeLists.txt`: remove `WITH_IMAGE` from `target_compile_definitions` for `icn2img` and `til2img` (related to #5194);
* Properly add `-lSDL_Mixer`/`-lSDL2_Mixer` to the list of linked libs (should go before the `-lSDL` due to the `SDL_LoadWAV_RW` dependency);
* Build the `Linux SDL2 Debug` job with `SDL2_Image`, remove `SDL_Image` dependencies from other jobs (this should speedup the dependency installation phase a bit);
* Build the `Clang Analyzer` and `SonarCloud` workflows with `FHEROES2_WITH_IMAGE` to expand the scope of the code being tested;
* Allow to set the `FHEROES2_DATA` via command-line/environment variable instead of setting it via `-DFHEROES2_DATA` in `CFLAGS` as before;
* List available options for Nintendo Switch and PS Vita in corresponding Makefiles;
* Tune the set of required libraries for Nintendo Switch and PS Vita;
* Fix build for PS Vita if `FHEROES2_WITH_IMAGE` is defined - currently build fails because `libwebp` is not linked.


ihhub:

src/Makefile
```diff
@@ -27,11 +27,11 @@ TARGET := fheroes2
 # Common flags for both C and C++ compilers
 CCFLAGS := -fsigned-char -pthread
 # Flags for the C compiler
-CFLAGS := $(CFLAGS)
+CFLAGS :=
```
undef21, line 30: @oleg-derevenetz could you please explain removing external flags? This change and similar ones below disable passing user- or system-defined CFLAGS, CXXFLAGS, LDFLAGS.

oleg-derevenetz: @undef21 because currently in fact different (although intersecting) set of flags is used to built the fheroes2 code and third-party code (third-party code has a more relaxed set of flags), so I don't want to provoke people to pass user-defined set of `CFLAGS` thinking that they are used just for the fheroes2 code itself.

oleg-derevenetz: @undef21 I don't mind to bring back the previous behavior regarding the user-defined `CFLAGS`, `CXXFLAGS`, `LDFLAGS` and probably `CCFLAGS`, because [some build systems use it](https://scons.org/doc/production/HTML/scons-man.html#cv-CCFLAGS), so you can offer a PR if you find this reasonable. I just hope that people will avoid passing additional warning-related options or maybe fheroes2-specific defines via `CFLAGS`/`CXXFLAGS`, but the system-related options, such as `-march`, can be passed, because they can be used both for third-party and fheroes2-specific code.

undef21: @oleg-derevenetz Also distro-specific build systems use exported FLAGS for building packages.
For example, Debian build helper adds hardening options (-fsta) and -g option for building dbg symbols package.
```
CFLAGS=-g -O2 -fdebug-prefix-map=~/Projects/debian/fheroes2-0.9.15/debian/patches=. -fstack-protector-strong -Wformat -Werror=format-security
CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2
CXXFLAGS=-g -O2 -fdebug-prefix-map=~/Projects/debian/fheroes2-0.9.15/debian/patches=. -fstack-protector-strong -Wformat -Werror=format-security
```
I guess we can stay project-specific CCFLAGS and other default values untouched and add CFLAGS, CXXFLAGS and CPPFLAGS for adding or redefining default values. Also we can add warnings "Redefining of flags is at your own risk" in README.:)

oleg-derevenetz: @undef21

> I guess we can stay project-specific CCFLAGS and other default values untouched and add CFLAGS, CXXFLAGS and CPPFLAGS for adding or redefining default values.

We can do like this in the `src/Makefile` then:

```
#
# Common build options for fheroes2 and third-party libraries
#

# Common flags for both C and C++ compilers
CCFLAGS := -fsigned-char -pthread
# Flags for the C compiler
CFLAGS := $(CFLAGS)
# Flags for the C++ compiler
CXXFLAGS := $(CXXFLAGS)
# Flags for the linker
LDFLAGS := $(LDFLAGS) -pthread
```

In this case these flags will be used for all code, including 3rd-party code, which is usually required for the system-wide build flags.

Issue #3720, opened 2021-06-19
AI: if AI looses a newly purchased hero another one could be purchased only at the next turn


ihhub:
Current logic in `void Normal::KingdomTurn( Kingdom & kingdom )` method could be described in such manner:
- move heroes
- buy heroes if needed
- move heroes
- develop castles

Such logic works perfectly in most of situations. However, we could describe some when this logic will fail (like in the description). The solution behind this is to create a while loop and check statuses of move heroes action, buying heroes action and developing castles action. If all these 3 actions are done then turn is completed. Also if a hero defeated another hero the danger ratio should be reevaluated which means we have to stop move heroes action and do some adjustments.

ihhub:
Fixed in #5202 .

Issue #2498, opened 2020-12-27
Game, recruitment of opponent heroes improved by AI players


LeHerosInconnu:
In fheroes2, it's too easy to recruit heroes improved by AI players compared to the original game.

In this game, I have already recruited heroes from the opposing players and there is still one hero available for recruitment in the tavern.

In fheroes2:

![Easy AI hero fh2 01](https://user-images.githubusercontent.com/43583994/103178186-e29ebe00-4880-11eb-82ce-f99b2e8224cc.png)

![Easy AI hero fh2 02](https://user-images.githubusercontent.com/43583994/103178187-e599ae80-4880-11eb-8759-90acea1600dc.png)

Here is the save file:
[Enemy hero tavern easy fh2 01.zip](https://github.com/ihhub/fheroes2/files/5745058/Enemy.hero.tavern.easy.fh2.01.zip)

vincent-grosbois:
I don't get it, what's the intended behavior?

LeHerosInconnu:
Hello @vincent-grosbois,

> I don't get it, what's the intended behavior?

In the original game, AI players seem to recruit their defeated heroes more often, so the human player is less likely to be able to recruit them.
Someone should check the source code of the original game to see how this is handled.

Branikolog:
I agree with @LeHerosInconnu. Enemy heroes with high level or with artifacts seem to appear in tavern too often. Either AI doesn't buy heroes that had run from a battle, or AI heroes were defeated too often...

idshibanov:
Currently AI buys hero only at the start of the turn, have to add another check at the end after all moves.

MicroWombat:
I keep running into this in version 1.1.16 as well: on almost every map, it’s possible very early on — sometimes literally in week 2 — to hire a leveled-up hero whom the previous owner somehow forgot about after they died.

PR #5202, opened 2022-04-04
Changed AI's turn to be repeated multiple times

felix642:
Improved AI's logic. At the end of every turn we check if new heroes have been recruited. If so we redo every steps to possibly reach new locations / recruit new heroes / Collect resources / etc.

Relates to #2498 and #3720


ihhub:
Hi @felix642 , please fix code styling issues in order to make the pipeline to succeed.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -241,127 +240,162 @@ namespace AI
...
-        // Step 1. Scan visible map (based on game difficulty), add goals and threats
-        std::vector<std::pair<int, const Army *>> enemyArmies;
+        uint8_t purchasedHeroesCount;
 
-        const int mapSize = world.w() * world.h();
-        _mapObjects.clear();
-        _regions.clear();
-        _regions.resize( world.getRegionCount() );
+        do {
+            purchasedHeroesCount = 0;
```
oleg-derevenetz, line 246: Hi @felix642 this value is never used because it will be always overwritten at line 361 (the Clang Analyzer also complains about this). I suggest to remove this line completely, because it's useless. Or you may even turn this loop into the infinite loop, remove the declaration at line 243 and do something like this:

```cpp
while ( true ) {
[...]
  uint8_t purchasedHeroesCount = purchaseNewHeroes(...);
[...]
  if ( purchasedHeroesCount == 0 ) {
    break;
  }
}
```

idshibanov: I would prefer infinite loop as well.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -403,30 +437,9 @@ namespace AI
             // target found, buy hero
             if ( recruitmentCastle && recruitHero( *recruitmentCastle, !slowEarlyGame, false ) ) {
                 ++availableHeroCount;
+                return true;
```
idshibanov, line 440: Change `purchaseNewHeroes` method definition so it returns a boolean instead of uint8_t.

Also `++availableHeroCount;` does nothing in this case, you can remove it.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -241,127 +240,162 @@ namespace AI
...
+
+        } while ( purchasedHeroesCount != 0 ); // If we have bought new heroes during this turn let's repeat the whole process.
+    }
+
+    uint8_t Normal::purchaseNewHeroes( VecHeroes & heroes, const std::vector<AICastle> & sortedCastleList, const std::set<int> & castlesInDanger,
+                                       int32_t availableHeroCount )
+    {
+        const bool slowEarlyGame = world.CountDay() < 5 && sortedCastleList.size() == 1;
+        int32_t heroLimit = world.w() / Maps::SMALL + 1;
+        const bool moreTasksForHeroes = HeroesTurn( heroes );
```
idshibanov, line 393: `const bool moreTasksForHeroes = HeroesTurn( heroes );`
This line executes all hero moves and definitely shouldn't be in this method.

Change method definition to pass in `const bool moreTasksForHeroes` instead of `VecHeroes & heroes` and move it back into turn loop.

idshibanov (changes requested):
This will probably be merged after the upcoming release as we'll need some time to test this change out. I'm not convinced we should repeat all turn calculations every time, this might slow down AI turns considerably.

Test on L and XL maps as hero counts will be higher there. I doubt there will be any impact on Small though.

felix642:
Hi @idshibanov,
I definitely agree with you, we must test this change thoroughly before merging this branch. We don't want to break anything. I did test a bit the new logic to see how it goes and it seems to be working pretty well for now but I haven't made enough tests to make sure that everything is working perfectly.

> I'm not convinced we should repeat all turn calculations every time, this might slow down AI turns considerably.

I have updated to scan the visible map only once instead of doing this multiple times. It should improve the AI's performances but I still need to test before pushing those changes.

idshibanov (approved):

src/fheroes2/ai/normal/ai_normal.h
```diff
@@ -155,6 +156,8 @@ namespace AI
         void CastleTurn( Castle & castle, bool defensive ) override;
         void BattleTurn( Battle::Arena & arena, const Battle::Unit & currentUnit, Battle::Actions & actions ) override;
         bool HeroesTurn( VecHeroes & heroes ) override;
+        bool purchaseNewHeroes( const std::vector<AICastle> & sortedCastleList, const std::set<int> & castlesInDanger, int32_t availableHeroCount,
```
ihhub, line 159: This method should be at least `private` as no other external code calls it.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -238,6 +237,9 @@ namespace AI
         KingdomHeroes & heroes = kingdom.GetHeroes();
         const KingdomCastles & castles = kingdom.GetCastles();
 
+        std::vector<AICastle> sortedCastleList;
+        std::set<int> castlesInDanger;
```
ihhub, line 241: Please move these variables to the place just before the loop where they are used. It would be easier to read the code.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -329,39 +331,67 @@ namespace AI
...
+            }
 
-        // Step 3. Do some hero stuff.
+            // Step 4. Reassign heroes roles
+            setHeroRoles( heroes );
 
-        // If a hero is standing in a castle most likely he has nothing to do so let's try to give him more army.
-        for ( Heroes * hero : heroes ) {
-            HeroesActionComplete( *hero );
+            status.RedrawTurnProgress( 6 );
```
ihhub, line 355: Won't we have a moving backwards progress sand hourglasses? At least it would just from 7 to 6 again.

ihhub: Since we don't have a call of `RedrawTurnProgress( 8 );` I suggest to do this way:
```cpp
// before the loop
uint32_t progressStatus = 6;

// inside the loop
if ( progressStatus == 6 ) {
    status.RedrawTurnProgress( 6 );
    ++progressStatus;
}
else {
    status.RedrawTurnProgress( 8 );
}

...
if ( progressStatus == 7 ) {
    status.RedrawTurnProgress( 7 );
}
else {
    status.RedrawTurnProgress( 8 );
}
```
What do you think?

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -329,39 +331,67 @@ namespace AI
...
+        int32_t heroLimit = world.w() / Maps::SMALL + 1;
 
-        // Step 4. Buy new heroes, adjust roles, sort heroes based on priority or strength
-        std::vector<AICastle> sortedCastleList = getSortedCastleList( castles, castlesInDanger );
+        if ( _personality == EXPLORER )
+            ++heroLimit;
+        if ( slowEarlyGame )
+            heroLimit = 2;
 
         if ( availableHeroCount < heroLimit ) {
```
ihhub, line 396: Please use fast return logic:
```cpp
if ( availableHeroCount >= heroLimit ) {
    return false;
}
...
// the rest of the code.
```

This makes the code more readable.

ihhub (changes requested):
Hi @felix642 , I left few comments here. Would you mind to take a look?

felix642:
Hi @ihhub, sorry for the delay. I have resolved the comments on this PR.

ihhub (approved):

ihhub:
@felix642 , thank you so much for the improvement!

PR #5204, opened 2022-04-05
Restore original set of letters from assets when changing language

zenseii:
This PR aims to solve an issue described in https://github.com/ihhub/fheroes2/issues/5187#issuecomment-1086843531 and which is also the root problem for why the changes made by `generateFrenchAlphabet()` to the standard set of letters presist across all languages when changing to any other language other than either English or the language present in the assets, like when having German original assets present, after French has been used as language.

The issue can be observed in any current fheroes2 release by looking at the English Clouds of Xeen map description and switching between French and other languages. In French the key `&` is replaced by `ù`. You would need other assets than the French to test this.

Original description with `&` intact, because French has not been set:
image

Language set to any other language than either English or the one provided with assets (in my case German assets are present) after having set language to French at any point. Polish is used here:
image

The issue also presents itself when trying to type the keys `" # ¤ &` etc. and they get replaced with `à ô û ù` etc. after French language has been set at some point.
Polish language being used after French and letters `" # ¤` have been typed:
image

This issue of typing letters is also fixed by the changes in this PR.

Here the lowercase normal font k does not change when Norwegian language is set, and you can see that `Norsk` aka. Norwegian uses its lowercase k by comparing it to the one in `polski`. Russian small k is also preserved:
image

Compared to before:
image


src/fheroes2/agg/agg_image.cpp
```diff
@@ -1626,6 +1626,9 @@ namespace
 
     void generateAlphabet( const fheroes2::SupportedLanguage language )
     {
+        // Restore original 0-96 letters when changing language in case of changes to them.
+        alphabetPreserver.restore();
```
ihhub, line 1630: We already have a function which calls this function:
```cpp
void updateAlphabet( const SupportedLanguage language, const bool loadOriginalAlphabet )
        {
            if ( loadOriginalAlphabet || !isAlphabetSupported( language ) ) {
                alphabetPreserver.restore();
            }
            else {
                alphabetPreserver.preserve();
                generateAlphabet( language );
            }
        }
```
Wouldn't it better to put the call just after `alphabetPreserver.preserve();`?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1626,6 +1626,9 @@ namespace
 
     void generateAlphabet( const fheroes2::SupportedLanguage language )
     {
+        // Restore original 0-96 letters when changing language in case of changes to them.
```
ihhub, line 1629: This comment is not entirely correct: restoring the original alphabet might restore more than 96 letters. Please try German for example.

ihhub (changes requested):
Hi @zenseii , I left few comments here. Also this approach is not ideal: what if we use French assets? Will the problem be there?

src/fheroes2/agg/agg_image.cpp
zenseii: Yes, this is a much better placement. Not sure why I didn't consider this first. I've tested it and it works.

zenseii:
Hi @ihhub ! Thanks for the review. I've adressed the comments and tested the PR with German and French assets and the problem isn't observed anymore.

ihhub:
@zenseii , a bunch of thanks for fixing this issue!

PR #5209, opened 2022-04-05
Fix switching fullscreen mode in macOS for certain resolutions

oleg-derevenetz:
If we call `SDL_SetWindowSize()` before turning off the fullscreen mode, the size of the resulting window may be wrong. This issue may be observed not only in macOS, but on other systems where `SDL_WINDOW_FULLSCREEN_DESKTOP` is used (e.g. Linux).

Before:

https://user-images.githubusercontent.com/32623900/161860527-0295a3cc-4dbb-467d-826e-4536b9f5a4ca.mp4

After:

https://user-images.githubusercontent.com/32623900/161860814-1c36bf88-c006-43df-8b03-5120f136b88f.mp4

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -709,10 +709,6 @@ namespace
             uint32_t flags = SDL_GetWindowFlags( _window );
             if ( ( flags & SDL_WINDOW_FULLSCREEN ) == SDL_WINDOW_FULLSCREEN || ( flags & SDL_WINDOW_FULLSCREEN_DESKTOP ) == SDL_WINDOW_FULLSCREEN_DESKTOP ) {
                 flags = 0;
-
-                if ( _windowedSize.width != 0 && _windowedSize.height != 0 ) {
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub BTW do you know why `SDL_WINDOW_FULLSCREEN` is used on Windows and not `SDL_WINDOW_FULLSCREEN_DESKTOP`? I suppose, the latter may help with #5015.

ihhub:
@oleg-derevenetz , I remember there was a reason behind this as Windows behaves differently with these 2 flags. I need to recollect why.

ihhub:
@oleg-derevenetz , for Windows:
- **SDL_WINDOW_FULLSCREEN** is a true full screen mode where OS switches monitor resolution, including scaled mouse cursor
- **SDL_WINDOW_FULLSCREEN_DESKTOP** is an interpolation of the full screen mode, like border-less application. In this case the monitor keep the original resolution but cursor become tiny if you choose a very tiny resolution, like 640x480.

So both modes have issues.

ihhub:

PR #5212, opened 2022-04-06
WIP: add memory pool

ihhub:


oleg-derevenetz:
Hi @ihhub did you tried the [alternative allocators](https://gcc.gnu.org/onlinedocs/libstdc++/manual/api.html#table.extension_allocators)? Like this:

https://godbolt.org/z/Y9rExsr7j

ihhub:
> Hi @ihhub did you tried the [alternative allocators](https://gcc.gnu.org/onlinedocs/libstdc++/manual/memory.html#allocator.ext)? Like this:
> 
> https://godbolt.org/z/8qdqT81KW

No, I haven't. Memory allocators are a huge area of research. There are many ways to implement a memory allocator. I took this implementation from a previous project so it's proven that it works properly. Also for such small amount of memory we don't need to have an ideal solution but having any memory allocator will be more beneficial. However, I am open for discussion and even to close this pull request if it has no value. As of now it's just a concept.

oleg-derevenetz:
@ihhub I believe you are worried about the potential memory fragmentation on the gaming platforms with relatively small amount of RAM (like PS Vita)?

ihhub:
> @ihhub I believe you are worried about the potential memory fragmentation on the gaming platforms with relatively small amount of RAM (like PS Vita)?

More like from performance side. We also have players on Raspberry Pi and the future release on mobile systems where energy is important. Of course it's subjective and we might not need at all. I'm marking it for 1.0 release as it has low priority.

oleg-derevenetz:
Hi @ihhub

> More like from performance side. We also have players on Raspberry Pi and the future release on mobile systems where energy is important. Of course it's subjective and we might not need at all. I'm marking it for 1.0 release as it has low priority.

I just wanted to know if there are any complaints on specific platforms, and are there people who can test introduced changes to verify that the relevant problem has been solved (or not solved)? If the proposed changes really give a tangible performance gain that can be measured and that is critical on some platforms, then I am completely in favor.

ihhub:
Closing this pull request for now to focus on other more important things.

PR #5215, opened 2022-04-07
Remove "battle: reverse wait order (fast, average, slow)" option

ihhub:
It's logical that slowest units after waiting should go first. Also AI has no knowledge of such option.


ihhub:
> @ihhub AFAIK AI is not able to utilize soft wait at all, does it? :)

Yes :)

oleg-derevenetz (approved):

Issue #5114, opened 2022-03-12
Add "Battle Only" button for Polush version of the game


J013k:
### Your question

Hi!

Is there a chance to replace a strange entry in the main menu with the one from an original EN version?

Or translate it:
Battle Only -> Tylko Bitwa

### Additional info

Here is how it looks like in PL version:
![https://i.postimg.cc/R0c57xCK/rre-1.png](https://i.postimg.cc/R0c57xCK/rre-1.png)
Standard Game -> Gra Standardowa
Campaign Game -> Kampania
Multi-Player Game -> Gra Dla Wielu Osób
**Battle Only -> ?**
Config -> Konfiguracja
Cancel -> Anuluj

Here is how it looks like in EN version:
![https://i.postimg.cc/fT0HtjR0/fheroes2-menu.png](https://i.postimg.cc/fT0HtjR0/fheroes2-menu.png)

oleg-derevenetz:
Duplicate of #3116

J013k:
Thank you!

There is no need to duplicate the same issue.

Laserlicht:
With #3116 only the german version is fixed.

https://github.com/ihhub/fheroes2/pull/4951 has to be done for PL (and other languages) as well.

zenseii:
I'll have a look at doing this for at least Italian and Norwegian after my current work regarding those languages is finished.

PR #5219, opened 2022-04-08
Generate French Battle Only button

zenseii:
This generates a Battle Only button if French resources are available:
image

Relates to this:
https://github.com/ihhub/fheroes2/issues/5114

I tried to center and align as good as possible - The box is an odd number so a perfect centering is not possible.
I don't want to bother you since I know you're quite busy, but if you should have the time to give a quick glance with your talent for spotting at the above image @LeHerosInconnu , then I would be very grateful! Don't mind that the buttons themselves are not aligned specifically in that image, that is just a quick cut and paste of them - the text is what matters.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -2045,20 +2045,8 @@ namespace fheroes2
                     // Add 'T'
                     Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 21, out, 50 - i, secondLine, 12, 15 );
                     // Clean up 'AT'
-                    // remove dot in top left corner
-                    Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 18, out, 50 - i, 28, 1, 1 );
-                    // White three white pixels on left hanging part of T
-                    /*Blit( GetICN( ICN::BTNDC, 2 + i ), 92 - i, 26, out, 49 - i, 32, 1, 3 );*/
-               
-                    //Blit( GetICN( ICN::BTNDC, 2 + i ), 84 - i, 32, out, 49 - i, 34, 1, 1 );
-                    // hidden non-pressed, visible pressed
-                    
-                    /*Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 22, out, 49 - i, secondLine + 4, 1, 1 );*/
-                    //Bottom right part of T ?
-                    /*Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 21, out, 50 - i, secondLine + 9, 1, 1 );*/
-                   
-                    
-                    
+                    Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 18, out, 50 - i, secondLine, 1, 1 );
+                    Blit( GetICN( ICN::BTNDC, 2 + i ), 92 - ( 5 * i ), 27 - i, out, 49 - i, secondLine + 4 + i, 1, 3 );
                     // Add 'AI'.
                     Blit( GetICN( ICN::BTNMP, 6 + i ), 56 - i, 13, out, 62 - i, secondLine, 18, 15 );
                     // Clean up 'TA'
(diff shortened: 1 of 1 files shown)


zenseii:
Hi @ihhub ! The button fonts are clean and I tried to be as greedy as possible with the generation of it, but if you see need for further optimization and ways to optimize it then I'm all ears. Also this font has 4 "layers" (I wish it had just 2 like German") which means I had to use some tricks like multiplication for the 'T' to get it clean and properly generated.

Also there is some slight dithering in the deepest/brightest layer for the 'Mode' part and for the 'AI' part. Unless there's a magic trick to uniform these colors I don't really see any need to put effort into fixing that.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2013,6 +2013,55 @@ namespace fheroes2
...
+            switch ( id ) {
+            case ICN::BTNBATTLEONLY:
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::BTNNEWGM, 6 + i );
+                    // Clean the button
+                    Image uniform( 88, 23 );
+                    uniform.fill( ( i == 0 ) ? GetColorId( 216, 184, 152 ) : GetColorId( 184, 136, 96 ) );
+                    Copy( uniform, 0, 0, out, 25, 18, uniform.width(), uniform.height() );
```
ihhub, line 2034: We can use something like this:
```cpp
const uint8_t fillColor = ( i == 0 ) ? GetColorId( 216, 184, 152 ) : GetColorId( 184, 136, 96 );
Fill( out, 25, 18, 88, 23, fillColor );
```
At the time of initial button creation (it was actually changing the old image generation code) Fill function didn't exist.

This solution is much faster and easier to read. Could you please also do this for German and English?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2013,6 +2013,55 @@ namespace fheroes2
                     Blit( GetICN( ICN::BTNNEWGM, 4 + i ), 36 - i, 23, out, 78 - i, 23, 10, 14 );
                 }
                 break;
+
+            default:
+                // You're calling this function for non-specified ICN id. Check your logic!
+                assert( 0 );
+                break;
+            }
+        }
```
ihhub, line 2022: Please add an empty line to separate 2 functions.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2013,6 +2013,55 @@ namespace fheroes2
...
+                    // Add 'T'
+                    Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 21, out, 50 - i, secondLine, 12, 15 );
+                    // Clean up 'AT'
+                    Blit( GetICN( ICN::BTNDC, 2 + i ), 89 - i, 18, out, 50 - i, secondLine, 1, 1 );
+                    Blit( GetICN( ICN::BTNDC, 2 + i ), 92 - ( 5 * i ), 27 - i, out, 49 - i, secondLine + 4 + i, 1, 3 );
+                    // Add 'AI'.
+                    Blit( GetICN( ICN::BTNMP, 6 + i ), 56 - i, 13, out, 62 - i, secondLine, 18, 15 );
+                    // Clean up 'TA'
+                    Blit( GetICN( ICN::BTNBAUD, 2 + i ), 51 - i, 40, out, 60 - i, secondLine + 12, 3, 3 );
+                    // add 'LLE'
```
ihhub, line 2054: To be consistent: please make **a** as a capital letter.

ihhub (changes requested):
Hi @zenseii , I put few comments here. Would you mind to take a look at them?

src/fheroes2/agg/agg_image.cpp
zenseii: Sure, I was actually wondering why Fill() was not used.

ihhub (approved):

ihhub:
@zenseii , a lot of thanks for working on translations!

PR #5224, opened 2022-04-09
Generate Italian Battle Button

zenseii:
Relates to this: https://github.com/ihhub/fheroes2/issues/5114

Had to use some more letter spacing to avoid overlapping when pressing the button.

image

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -1894,9 +1894,9 @@ namespace
         }
     }
 
-    uint8_t getButtonFillingColor( const int32_t isReleasedState )
+    uint8_t getButtonFillingColor( const bool isReleasedState )
     {
-        return isReleasedState ? fheroes2::GetColorId( 184, 136, 96 ) : fheroes2::GetColorId( 216, 184, 152 );
+        return isReleasedState ? fheroes2::GetColorId( 216, 184, 152 ) : fheroes2::GetColorId( 184, 136, 96 );
     }
 }
 
@@ -1968,7 +1968,7 @@ namespace fheroes2
                     Sprite & out = _icnVsSprite[id][i];
                     out = GetICN( ICN::BTNNEWGM, 6 + i );
                     // Clean the button
-                    Fill( out, 25, 18, 88, 23, getButtonFillingColor( i ) );
+                    Fill( out, 25, 18, 88, 23, getButtonFillingColor( i == 0 ) );
                     // Add 'ba'
                     Blit( GetICN( ICN::BTNCMPGN, i ), 41 - i, 28, out, 30 - i, 13, 28, 14 );
                     // Add 'tt'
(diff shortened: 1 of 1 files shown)


src/fheroes2/agg/agg_image.cpp
```diff
@@ -2101,6 +2101,59 @@ namespace fheroes2
...
+        void generateItalianSpecificImages( const int id )
+        {
+            switch ( id ) {
+            case ICN::BTNBATTLEONLY:
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::BTNNEWGM, 6 + i );
+                    // clean the button
+                    const uint8_t fillColor = ( i == 0 ) ? GetColorId( 216, 184, 152 ) : GetColorId( 184, 136, 96 );
```
ihhub, line 2113: Can we please create a helper function in anonymous namespace to avoid this duplicated code? Something like this:
```cpp
uint8_t getButtonFillingColor( const isReleasedState )
{
    return isReleasedState ? GetColorId( 216, 184, 152 ) : GetColorId( 184, 136, 96 );
}
```

Lesser magical numbers easier to read the code.

ihhub (changes requested):

src/fheroes2/agg/agg_image.cpp
zenseii: Yes, I'll have a look at this. I can probably also remove the comments about Clean up since it should be obvious from the Fill() function call.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1893,6 +1893,11 @@ namespace
             // Other transform values are not relevant for transparency checks.
         }
     }
+
+    uint8_t getButtonFillingColor( const int32_t isReleasedState )
```
ihhub, line 1897: This input argument should be just `bool`

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1963,8 +1968,7 @@ namespace fheroes2
                     Sprite & out = _icnVsSprite[id][i];
                     out = GetICN( ICN::BTNNEWGM, 6 + i );
                     // Clean the button
-                    const uint8_t fillColor = ( i == 0 ) ? GetColorId( 216, 184, 152 ) : GetColorId( 184, 136, 96 );
-                    Fill( out, 25, 18, 88, 23, fillColor );
+                    Fill( out, 25, 18, 88, 23, getButtonFillingColor( i ) );
```
ihhub, line 1971: Here and in another places: `i == 0` instead of `i`.

ihhub (changes requested):

ihhub (approved):

ihhub:

PR #5227, opened 2022-04-09
Fix video file loading for case-sensitive filesystems

ihhub:
Previous approach was bug-prone as we assumed that file names are in all capital letters which might not be always true.

regression from #5125

### DIFF
--- a/src/fheroes2/game/game_video.cpp
+++ b/src/fheroes2/game/game_video.cpp
@@ -68,9 +68,7 @@ namespace Video
                     targetFileName = StringLower( targetFileName );
 
                     for ( const std::string & filePath : videoFiles ) {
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz , if you have time could you please test it on MacOS?

oleg-derevenetz:
> Hi @oleg-derevenetz , if you have time could you please test it on MacOS?

Hi @ihhub I'll test this later today, but please note that macOS default filesystem mode (including my own devices) is case-insensitive. If you want to check the behavior with case-sensitive filesystems, then it would be better to run the tests on Linux with EXT3 for example.

ihhub:
@oleg-derevenetz , I tested on my Linux virtual machine and it looks fine.

Issue #5225, opened 2022-04-09
Yellow colored word gets cut mid-word in introduction message


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

The yellow-colored word gets cut mid-word if it arrives at the end of the line. This is viewable in several of the languages. This is the message that comes after setting the resolution.

In the image the word should be `porte`.

![image](https://user-images.githubusercontent.com/12501091/162562892-9d0128e6-bf1d-4184-8380-4f9080377ed2.png)

You would basically need to change the Multicolored text function in ui_text.cpp to have whole words preserved when breaking a line like is already done for single-colored text.

### Save file

Just start the game with First run being set in fheroes2.cfg. This happens for several languages. French is one of them.

ihhub:
Hi @zenseii , it is a translation issue. We have this code:
```cpp
body.add( { _( "You can always change game resolution by clicking on the " ), fheroes2::FontType::normalWhite() } );
body.add( { _( "door" ), fheroes2::FontType::normalYellow() } );
body.add( { _( " on the left side of main menu.\n\nTo switch between windowed and full screen modes\npress " ), fheroes2::FontType::normalWhite() } );
body.add( { _( "F4" ), fheroes2::FontType::normalYellow() } );
body.add( { _( " key on the keyboard.\n\nEnjoy the game!" ), fheroes2::FontType::normalWhite() } );
```

So the translation should use `\n` where appropriate.

zenseii:
I'll check and update the langauges that this regards.

PR #5228, opened 2022-04-09
Fix line breaking in welcome message for various languages

zenseii:
Fixes issue: https://github.com/ihhub/fheroes2/issues/5225

For German, Polish, French, Norwegian and Italian.

The automatic formatting also fixed formatting issues in the German de.po.

The relevant line is the translation of `"You can always change game resolution by clicking on the"`

Depending on what was necessary and best either whitespace or \n was used in the language.

Note that the translated string has to end with a whitespace since the source string does that.


ihhub:
@zenseii , great thanks for fixing it!

PR #5231, opened 2022-04-09
Simplify the version update process by getting the version number from a single file

oleg-derevenetz:
The idea is to set the app version in a single file (`version.txt`) and get it from this file everywhere. Unfortunately, it seems that `version.h` and `sonar-project.properties` still need to be updated by hand (or using some utility script).

Also fixed an issue when C++ specific warning options (such as `-Woverloaded-virtual`) may be applied to the C code. This doesn't break the build, but GCC still complains about that.

Also fixed CMake build on Windows (with the current settings, it was impossible to build CMake project with MSVC in strict mode) and added the GitHub Actions pipeline for CMake on Windows. Rarely used build configurations (like any unused code) are gradually becoming unreliable, and they need to be constantly checked using pipelines.

Also replaced the `GITHUB_RUN_NUMBER` by `GITHUB_RUN_ID` for the purpose of calculating the build number. The thing is, `GITHUB_RUN_NUMBER` is described as "a unique number for each run of a particular workflow in a repository", but it turns out that this is not quite true if the workflow in question was called from the another workflow. In this case it will be taken from the caller workflow, which leads to inconsistent build number generation if the workflow in question was called from the multiple different workflows (like `MSVC` workflow which in our case may be called from `Push` and `Pull Request` workflows). `GITHUB_RUN_ID`, on the other hand, is workflow-independent.


ihhub:
@oleg-derevenetz , great work! Thank you for putting efforts for our automation!

PR #5235, opened 2022-04-10
Always grab the mouse in fullscreen mode to properly support devices with notch

### LINKED ISSUE #5233: Fullscreen panning does not work with Mac with notch in the upper area screen of the notch
### Preliminary checks

### Platform

Mac

### Describe the bug

The same user from this issue https://github.com/ihhub/fheroes2/issues/5203  with a Mac with the notch reports that full screen panning does not work in the upper area of the screen just as before as seen in the issue video posted in the linked to issue.

### Save file

Fullscreen mode

### Additional info

_No response_

oleg-derevenetz:
fix #5233

It seems that black bar with notch is not a part of the app's window.

Hi @zenseii can the reporting user test this change?

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -1052,12 +1052,8 @@ namespace
 
         void _toggleMouseCaptureMode()
         {
-            if ( SDL_GetNumVideoDisplays() < 2 ) {
-                // Less than 2 monitors in the system. Nothing to do.
-                return;
-            }
-
-            // This is a multi-display device. To properly support fullscreen mode it is important to lock mouse within application window area.
+            // To properly support fullscreen mode on devices with multiple displays or devices with notch,
+            // it is important to lock the mouse in the application window area.
             if ( isFullScreen() ) {
                 SDL_SetWindowGrab( _window, SDL_TRUE );
             }


zenseii:
Hi @oleg-derevenetz ! The user had this to say about using the artifact packages to install fheroes2:
"However, I do not think the zip archive has the binaries I need for apple silicon. I have only installed fheroes2 via homebrew"

I'm not familiar with Mac but are the binary provided in the artifacts be easily usable on a M1 mac? What is the general procedure to do this so I can share it with the user?

oleg-derevenetz:
@zenseii

> I'm not familiar with Mac but are the binary provided in the artifacts be easily usable on a M1 mac? What is the general procedure to do this so I can share it with the user?

Mac on M1 has Rozetta emulation layer that should be able to run x86 code. However, if there are issues, he can try to clone my fork, checkout the `macos-notch-scroll-fix` branch and build the app from source.

ihhub:
@oleg-derevenetz yup, that's the way :) I completely forgot that we did such in the past.

oleg-derevenetz:
No feedback so far, but I believe that this change at least won't hurt anyone.

ihhub:

PR #5263, opened 2022-04-11
Fix mage guild level reading from a save file

### LINKED ISSUE #5232: Cannot build upgraded mausoleum in Fool's Gold scenario
### Preliminary checks

### Platform

Other

### Describe the bug

The upgraded mausoleum cannot be built even though resources and prerequisites are in place. The town had mage guild level 4 at first when captured by the human player, and the upgrade could not be build. Building level 5 did not help.
<details>
  <summary>Click here for image</summary>

image
</details>

This town is set like this in the map editor:
https://cdn.discordapp.com/attachments/906187653144268800/962562414635782144/unknown.png
There's another town on the map with the same configuration but building upgraded mausoleum there was possible looking at the save.

### Save file

[Fool_0099.zip](https://github.com/ihhub/fheroes2/files/8458256/Fool_0099.zip)

Use the picture above to find the correct town.

### Additional info

This was reported by a user Hydd on Discord using Mac. I confirmed the issue on Windows.

ihhub:
close #5232

I still have no idea how it happened.

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -2475,6 +2475,20 @@ StreamBase & operator>>( StreamBase & msg, Castle & castle )
 
     msg >> static_cast<MapPosition &>( castle ) >> castle.modes >> castle.race >> castle.building >> castle.captain >> color >> castle.name >> castle.mageguild;
 
+    // Make sure that Mage Guild has all previous levels.
+    if ( castle.building & BUILD_MAGEGUILD5 ) {
+        castle.building |= BUILD_MAGEGUILD4;
+    }
+    if ( castle.building & BUILD_MAGEGUILD4 ) {
+        castle.building |= BUILD_MAGEGUILD3;
+    }
+    if ( castle.building & BUILD_MAGEGUILD3 ) {
+        castle.building |= BUILD_MAGEGUILD2;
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
@ihhub

> I still have no idea how it happened.

I have an idea. AI is able to build mage guilds of upper levels without building the mage guild of the lower level due to insufficient prerequisites check for mage guilds. For example, here:

https://github.com/ihhub/fheroes2/blob/ffddfd612290d8e09d4eb1e9ca054e79ed188518/src/fheroes2/ai/normal/ai_normal_castle.cpp#L160-L166

If you comment out all other build logic in this function and left just the `BUILD_MAGEGUILD5` in this list, you will see that the AI will build the level 5 of Mage Guild at the first day. But you will not be able to build even the non-upgraded Mausoleum due to the lack of level 1 of Mage Guild. Here is the test map I used to check this:

[!NECTOWN.zip](https://github.com/ihhub/fheroes2/files/8621675/NECTOWN.zip)

ihhub:
Hi @oleg-derevenetz , I think we can drop this PR since we fixed the root of the problem. What do you think?

oleg-derevenetz:
Hi @ihhub I believe yes, we can close this PR.

PR #5279, opened 2022-04-15
Add Black/White cursor support

ihhub:
For now only through configuration file.

image


ihhub:
Hi @oleg-derevenetz , I think we still have issues with compilation for PS Vita.

oleg-derevenetz:
>  I think we still have issues with compilation for PS Vita.

Hi @ihhub these errors seems legitimate - alignment requirements of `int16_t` are higher than alignment requirements of `char`, it may be dangerous to just use `reinterpret_cast` in this case.

ihhub:
> > I think we still have issues with compilation for PS Vita.
> 
> Hi @ihhub these errors seems legitimate - alignment requirements of `int16_t` are higher than alignment requirements of `char`, it may be dangerous to just use `reinterpret_cast` in this case.

It's a fair point. I'll change the logic.

oleg-derevenetz:
> > > I think we still have issues with compilation for PS Vita.
> > 
> > Hi @ihhub these errors seems legitimate - alignment requirements of `int16_t` are higher than alignment requirements of `char`, it may be dangerous to just use `reinterpret_cast` in this case.
> 
> It's a fair point. I'll change the logic.

@ihhub `serialize.h` has the `getLEValue()` for such cases to safely load some numeric value stored in the little-ending format from the data file.

oleg-derevenetz:
However streams should do the job as well.

oleg-derevenetz (commented):
Hi @ihhub we can call these cursors "monochrome cursor" or `MONO_CURSOR_XXX`. What do you think?

ihhub:
> Hi @ihhub we can call these cursors "monochrome cursor" or `MONO_CURSOR_XXX`. What do you think?

Yes, it is a good idea.

src/fheroes2/agg/icn.h
```diff
@@ -963,6 +964,10 @@ namespace ICN
         EVIL_ARMY_BUTTON,
         EVIL_MARKET_BUTTON,
 
+        MONO_CURSOR_ADVMBW,
+        MONO_CURSOR_SPELBW,
+        MONO_CURSOR_CMSSBW,
+
```
oleg-derevenetz, line 970: Just curious, what those BW mean at the end?

oleg-derevenetz (approved):

src/fheroes2/agg/icn.h
ihhub: Black-White :) It comes from the naming of the original resources.

Issue #5270, opened 2022-04-12
Missing dialogs from text mode


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

This is an overview of all the things I've found that do not output to the console when playing with text mode enabled. Include these events:
	
Adventure map related:
- [ ] Arena dialog
- [ ] Level up dialog - Also needs hotkey between skills ex left arrow/right arrow
- [ ] Treasure chests
- [ ] Finding resources on ground Ex "You find a small quantity of ..."
- [ ] Followers dialog - A group of …
- [ ] Quick info text when right-clicking friendly hero and castle
- [ ] Day week month text when new day and be able to find out what the current day is.
- [ ] Any messages involving gaining resources don't report what and how much ex. Campfire
- [ ] Any visit giving you a primary skill bonus should report what you gain like Defence +1 for fort or Stonehenge +1 power
- [ ] Be able to find out what your current resources are.

Hero screen:
- [ ] Dialog for right-clicking dismiss hero.
- [ ] Report that formation has been changed to grouped or spread.
- [ ] Exit button dialog

Menus:
- [ ] Main menu's Config menu
- [ ] Resolution options
- [ ] Experiemental options
- [ ] Player input name dialog in scenario selection

Battle related:
- [ ] Battle results dialog
- [ ] Battle log
- [ ] Surrender
- [ ] Probably lots more to be honest.

Castle:
- [ ] Recruit dialog - valid for adventure map dwellings too.
- [ ] Castle build options: Right clickable info on building information
- [ ] Recruit hero dialog
- [ ] Creature stats/info dialog when right clicking on it in castle or battle.

Others:
- [ ] Fheroes2 resurrection team presents
- [ ] Close spellbook dialog
- [ ] Original campaign: "Choose your lord" between archibald and roland
- [ ] Campaign bonus choice names (artifact, gold, starting race) and hotkey  for selection of bonuses
- [ ] Campaign replay intro hotkey and text
- [ ] Report AI actions during AI turn within charted territory (this is one that likely will need time)
- [ ] Hero meeting dialog - Probably the same as battle only mode.
- [ ] Credits
- [ ] High scores

Discussion ideas:

- Right-clicking on a map artifact doesn't show what the name of it is, when for people without vision impairments they can easily tell what artifact it is like seeing that the artifact is the telescope.
[...]

PR #5281, opened 2022-04-15
Add text support mode for campaign selection dialogs

ihhub:
Video file playback now does not require ROI to be added. This was a corner case only available for The Succession Wars campaign selection.

relates to #5270


src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -412,8 +412,16 @@ void Game::HotKeysDefaults( void )
...
+    key_events[EVENT_NEW_CAMPAIGN_SELECTION_SUCCESSION_WARS] = KEY_o;
+    key_events[EVENT_NEW_CAMPAIGN_SELECTION_PRICE_OF_LOYALTY] = KEY_e;
+
+    key_events[EVENT_NEW_ROLAND_CAMPAIGN] = KEY_1;
+    key_events[EVENT_NEW_ARCHIBALD_CAMPAIGN] = KEY_2;
+
+    key_events[EVENT_NEW_PRICE_OF_LOYALTY_CAMPAIGN] = KEY_1;
+    key_events[EVENT_NEW_VOYAGE_HOME_CAMPAIGN] = KEY_2;
+    key_events[EVENT_NEW_WIZARDS_ISLE_CAMPAIGN] = KEY_3;
+    key_events[EVENT_NEW_DESCENDANTS_CAMPAIGN] = KEY_4;
```
oleg-derevenetz, line 424: It would be nice to mention this to fheroes2.key as well.

oleg-derevenetz (changes requested):

src/fheroes2/game/game_hotkeys.cpp
ihhub: It is done. We should remove this file as generate it like we do for the configuration file as many entries are missing. As of now I polished the code and added new entries into the hotkey file.

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -430,14 +374,9 @@ bool Game::HotKeyHoldEvent( const int eventID )
...
-const char * Game::getHotKeyNameByEventId( const int eventID )
+std::string Game::getHotKeyNameByEventId( const int eventID )
 {
-    auto it = keyName.find( key_events[eventID] );
-    if ( it != keyName.end() ) {
-        return it->second;
-    }
-
-    return "???";
+    return StringUpper( KeySymGetName( key_events[eventID] ) );
```
ihhub, line 379: SDL has the same logic of converting key values into human-readable string so this code was modified.

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -27,276 +27,212 @@
...
 
+#include <array>
+#include <cassert>
 #include <map>
 
+namespace
+{
+    std::array<KeySym, Game::EVENT_LAST> key_events{ KEY_NONE };
+
+    const char * getEventName( const int eventId )
```
ihhub, line 40: Some entries were missing in this function.

oleg-derevenetz (approved):

PR #5285, opened 2022-04-16
Fix versioning of MSVC builds

oleg-derevenetz:
Follow-up to #5231

Unfortunately replacing `GITHUB_RUN_NUMBER` with the `GITHUB_RUN_ID` did not bring the desired results, because, although `GITHUB_RUN_ID` is workflow-independent, it seems that there is no requirement for it to increase monotonically, so it is not suitable as the build number. I removed generation of the `FHEROES2_BUILD_NUMBER` from the MSVC workflow for a while and used the `ignoreversion` [flag](https://jrsoftware.org/ishelp/index.php?topic=filessection) in the Inno Setup instead. May be in future I'll come up with a better solution.

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -54,7 +54,6 @@ jobs:
     - name: Generate version information
       run: |
         echo "FHEROES2_APP_VERSION=$(cat version.txt)" >> "$GITHUB_ENV"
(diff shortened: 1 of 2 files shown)


ihhub:
@oleg-derevenetz , thank you for the fix!

PR #5286, opened 2022-04-16
Reset the mixer volume before playing video soundtracks

### LINKED ISSUE #5282: Volume during campaign movies reduced a lot when seen right after winning a scenario
### Preliminary checks

### Platform

Windows

### Describe the bug

The volume for the all campaign scenario movies gets reduced quite a lot (very hard to hear the voice) if you finish a scenario in a campaign (not sure how if this happens with the final scenario in a campaign). After this point, any campaign video's volume will be lowered, even if started from the main menu. Their sound depend on the effects volume, not music. However, it seems even if you set sound effects to 10 and finish the map the sound will still be lowered, but it's louder than if you had it set to 6. The volume gets corrected if you just start a standard game scenario and leave back to main menu by new game/load game. The volume returns to normal if you restart fheroes2 too.

It seems to be more evident in the save file attached below for completing the first scenario of the evil campaign. Tested with completing PoL scenario 2 and the sound wasn't lowered as much. Maybe because I didn't level up after the final battle, or maybe it was just the nature of the audio in the PoL movie.

This was tested with both French and English assets on latest snapshot by myself, and confirmed by another user on Discord on public release 0.9.14 as well.

### Save file

[Evil_1_Right_beforFinish map.zip](https://github.com/ihhub/fheroes2/files/8497515/Evil_1_Right_beforFinish.map.zip)
1. Load the save file
2. Attack the castle, preferably with quick battle on
3. Confirm all windows until you get to the video and the sound will be lowered.

### Additional info

_No response_

oleg-derevenetz:
fix #5282

Mixer is usually used to play ambient sounds, which may have different volume depending on the distance. We should reset the volume of mixer channels to the default one before playing video soundtracks.

### DIFF
--- a/src/fheroes2/game/game_video.cpp
+++ b/src/fheroes2/game/game_video.cpp
@@ -44,6 +44,8 @@ namespace
 
     void playAudio( const std::vector<std::vector<uint8_t>> & audioChannels )
     {
+        Mixer::Volume( -1, Mixer::MaxVolume() * Settings::Get().SoundVolume() / 10 );
+
         for ( const std::vector<uint8_t> & audio : audioChannels ) {
             if ( !audio.empty() ) {
                 Mixer::Play( &audio[0], static_cast<uint32_t>( audio.size() ) );


ihhub:

PR #5294, opened 2022-04-17
Update libsmacker to 1.2.0r43

oleg-derevenetz:
Currently, the bundled version of `libsmacker` is an ancient one, so I propose to replace it by the latest version from the upstream (at the moment).


ihhub:
Hi @oleg-derevenetz , please test it carefully as we made some fixes in the code within the library.

oleg-derevenetz:
Yeah, there are big issues with this version. For example, it is unable to play any of the expansion campaign videos, for example, `MIXPOL1.SMK`.

But there is a silver lining, I found and fixed another bug in the [clang-tidy-pr-comments](https://github.com/platisd/clang-tidy-pr-comments) action. Hi @ihhub you may want to sync your fork with the upstream.

ihhub:
@oleg-derevenetz , I synced with the upstream.

ihhub:
`MIXPOL1.SMK` file might be audio only file. There is a check in game_video.cpp to validate the number of frames.

oleg-derevenetz:
> `MIXPOL1.SMK` file might be an audio only file. There is a check in game_video.cpp to validate the number of frames.

No, it's a video, but very "static" one (just a few moving fragments). Libsmacker is not able to load this file because it fails the size check in the `smk_huff16_build()`.

oleg-derevenetz:
Hi @ihhub it seems that I fixed all issues with audio only SMK files (like "Choose your lord") and those strange small files with Huffman trees (like `MIXPOL1.SMK`), so this new version of libsmacker now works. However the advantages of the new library are not obvious (except that it is new and has slightly improved error handling), therefore, it is up to you to decide whether it is worth updating or not.

I'm also wondering if there is any point in having the `USE_SYSTEM_LIBSMACKER` option in CMake, because the original libsmacker is clearly unusable in our case.

oleg-derevenetz:
Hi @ihhub

> could you please review and merge pull requests which you think are good and obvious enough to be merged? I need a break to solve my personal issues so I will be away for a while. However, the project shouldn't be stopped as many players are waiting new changes.

Yes, of course, I will make an effort to do this. I wish you to successfully deal with all these issues.

Issue #4804, opened 2021-12-19
Marketplace and Trading Post windows, Select Recipients window, wrong elements position


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, there is a wrong elements position in the Select Recipients window.

The distance between the green and red squares representing the players to be selected is not the same as between the other squares, 21 pixels instead of 22 pixels.
The block of colored squares representing the players to be selected is not centered horizontally in relation to the edges of the window frame.
The two blocks of resource icons for "Your Funds" and "Planned Gift" are not centered horizontally in relation to the edges of the window frame.

In fheroes2:
![Select recipients window 01](https://user-images.githubusercontent.com/43583994/146682193-c1b1590a-a386-467b-96f5-f129a6a02fe8.png)

### Save file

Not needed.

srmoll:
Hi, could I be assigned this issue?

ihhub:
Hi @srmoll , sure!

PR #5295, opened 2022-04-17
Fix spacing of player squares in Select Recipients window and horiz. center resource icons for "Your Funds" and "Planned Gift"

srmoll:
This commit was written for issue #4804. It replaces the call to Game::GetStep4Player(...) with an inlined change that uses integer ceiling arithmetic to replace Game::GetStep4Player(...). The functional requirement was to ensure that the colored squares representing players would be evenly spaced was met.

### DIFF
--- a/src/fheroes2/dialog/dialog_giftresources.cpp
+++ b/src/fheroes2/dialog/dialog_giftresources.cpp
@@ -34,6 +34,11 @@
 #include "ui_window.h"
 #include "world.h"
 
+namespace
+{
+    const fheroes2::Size giftDialogSize( 320, 234 );
+}
+
 int32_t GetIndexClickRects( const std::vector<fheroes2::Rect> & rects )
 {
     LocalEvent & le = LocalEvent::Get();
@@ -64,16 +69,17 @@ struct SelectRecipientsColors
         , recipients( 0 )
     {
         positions.reserve( colors.size() );
-        const fheroes2::StandardWindow frameborder( 320, 234 );
-        const fheroes2::Rect box( frameborder.activeArea() );
+        const fheroes2::Display & display = fheroes2::Display::instance();
+        const fheroes2::Rect box( ( display.width() - giftDialogSize.width ) / 2, ( display.height() - giftDialogSize.height ) / 2, giftDialogSize.width,
+                                  giftDialogSize.height );
 
         const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::CELLWIN, 43 );
         const int32_t colorCount = static_cast<int32_t>( colors.size() ); // safe to cast as the number of players <= 8.
-        int32_t playerContainerWidth = colorCount * sprite.width() + ( colorCount - 1 ) * 22; // original spacing = 22
-        int32_t startx = box.x + 160 - playerContainerWidth / 2; // fixed half width of brown box = 160
+        const int32_t playerContainerWidth = colorCount * sprite.width() + ( colorCount - 1 ) * 22; // original spacing = 22
+        const int32_t startX = box.x + ( giftDialogSize.width - playerContainerWidth ) / 2;
         for ( int32_t i = 0; i < colorCount; ++i ) {
-            int32_t posx = startx + i * ( 22 + sprite.width() );
-            positions.emplace_back( posx, pos.y, sprite.width(), sprite.height() );
+            const int32_t posX = startX + i * ( 22 + sprite.width() );
+            positions.emplace_back( posX, pos.y, sprite.width(), sprite.height() );
         }
     }
 
@@ -88,6 +94,7 @@ struct SelectRecipientsColors
 
         for ( Colors::const_iterator it = colors.begin(); it != colors.end(); ++it ) {
             const fheroes2::Rect & pos = positions[std::distance( colors.begin(), it )];
+
             fheroes2::Blit( fheroes2::AGG::GetICN( ICN::CELLWIN, 43 + Color::GetIndex( *it ) ), display, pos.x, pos.y );
             if ( recipients & *it )
                 fheroes2::Blit( fheroes2::AGG::GetICN( ICN::CELLWIN, 2 ), display, pos.x + 2, pos.y + 2 );
@@ -138,7 +145,7 @@ struct ResourceBar
         fheroes2::Display & display = fheroes2::Display::instance();
         fheroes2::Text text( std::to_string( count ), fheroes2::FontType::smallWhite() );
         const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::TRADPOST, 7 + Resource::getIconIcnIndex( type ) );
-        fheroes2::Blit( sprite, fheroes2::Display::instance(), posx, posy );
+        fheroes2::Blit( sprite, display, posx, posy );
         text.draw( posx + ( sprite.width() - text.width() ) / 2, posy + sprite.height() - 10, display );
     }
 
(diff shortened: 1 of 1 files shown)


srmoll:
@oleg-derevenetz I wanted to figure out the other functional requirements for issue #4804.  ie: "Your Funds" and "Planned Gift" are not centered horizontally in relation to the edges of the window frame. I looked at the code, and I'm unsure on where in the code this discrepancy comes from, but I was thinking would it be from adding 25 (px) to the width in the ResourceBar for each?

zenseii:
> @oleg-derevenetz I wanted to figure out the other functional requirements for issue #4804. ie: "Your Funds" and "Planned Gift" are not centered horizontally in relation to the edges of the window frame. I looked at the code, and I'm unsure on where in the code this discrepancy comes from, but I was thinking would it be from adding 25 (px) to the width in the ResourceBar for each?

Hi @srmoll ! You are right about this assumption. Here is what happens if you set it to + 0 instead of 25
image

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -64,12 +65,14 @@ struct SelectRecipientsColors
     {
         positions.reserve( colors.size() );
 
+        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::CELLWIN, 43 );
         const int32_t colorCount = static_cast<int32_t>( colors.size() ); // safe to cast as the number of players <= 8.
+        int32_t playerCtnrWidth = 2 * sprite.width() + ( colorCount - 1 ) * 22; // original spacing = 22
```
zenseii, line 70: How about dividing by 2 on both sides of the plus symbol so that we don't have to divide by 2 in the line below?

Also, what does Ctnr stand for? Maybe we could just write it out? I'm not sure it's intuitive.

srmoll: The two is a typo, I put that there to test out how it would look with 2 players. It should say colorCount. Ctnr = Container, and yes, writing it out is more intuitive

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -63,13 +64,16 @@ struct SelectRecipientsColors
         , recipients( 0 )
     {
         positions.reserve( colors.size() );
+        const fheroes2::StandardWindow frameborder( 320, 234 );
+        const fheroes2::Rect box( frameborder.activeArea() );
```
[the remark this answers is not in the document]
ihhub: If the intention was to get the future position of the dialog then please do this way:
```cpp
const fheroes2::Display & display = fheroes2::Display::instance();
const fheroes2::Rect box( ( display.width() - 320 ) / 2, ( display.height() - 234 ) / 2, 320, 234 );
```

and since we use 320 and 234 in another place I recommend to create a separate variable in anonymous namespace after header inclusions:
```cpp
namespace
{
    const fheroes2::Size giftDialogSize( 320, 234 );
}
```

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -63,13 +64,16 @@ struct SelectRecipientsColors
...
+        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::CELLWIN, 43 );
         const int32_t colorCount = static_cast<int32_t>( colors.size() ); // safe to cast as the number of players <= 8.
-
+        int32_t playerContainerWidth = colorCount * sprite.width() + ( colorCount - 1 ) * 22; // original spacing = 22
+        int32_t startx = box.x + 160 - playerContainerWidth / 2; // fixed half width of brown box = 160
         for ( int32_t i = 0; i < colorCount; ++i ) {
-            const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::CELLWIN, 43 );
-
-            positions.emplace_back( pos.x + Game::GetStep4Player( i, sprite.width() + 15, colorCount ), pos.y, sprite.width(), sprite.height() );
+            int32_t posx = startx + i * ( 22 + sprite.width() );
```
ihhub, line 75: This variable can be `const`.

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -132,10 +135,11 @@ struct ResourceBar
 
     static void RedrawResource( int type, s32 count, s32 posx, s32 posy )
     {
-        Text text( std::to_string( count ), Font::SMALL );
+        fheroes2::Display & display = fheroes2::Display::instance();
+        fheroes2::Text text( std::to_string( count ), fheroes2::FontType::smallWhite() );
         const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::TRADPOST, 7 + Resource::getIconIcnIndex( type ) );
         fheroes2::Blit( sprite, fheroes2::Display::instance(), posx, posy );
```
ihhub, line 141: Please use `display` here instead of **fheroes2::Display::instance()**.

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -211,26 +215,25 @@ void Dialog::MakeGiftResource( Kingdom & kingdom )
 
     Funds funds1( kingdom.GetFunds() );
     Funds funds2;
-    Text text;
 
-    text.Set( _( "Select Recipients" ) );
-    text.Blit( box.x + ( box.width - text.w() ) / 2, box.y + 5 );
+    const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::TRADPOST, 7 );
+    int32_t posx = box.x + 160 - ( sprite.width() + 240 ) / 2; // 160 = half width of brown box, fixed
```
ihhub, line 220: Please name this variable using `camelCase` and also make it `const`.

ihhub: I think 160 could be used as `giftDialogSize.width / 2` to avoid hardcoded values. Then we won't need this comment.

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -30,6 +30,7 @@
 #include "text.h"
```
ihhub, line 30: I think we can remove this file inclusion since we replaced all instances of the old class.

ihhub (dismissed):
Hi @srmoll , I left multiple comments here. Could you please take a look at them?

srmoll:
text.h cannot yet be deleted due to the uses of Font::BIG, which I am not sure how to replace, would fheroes2::FontType::largeWhite() work?

zenseii:
> text.h cannot yet be deleted due to the uses of Font::BIG, which I am not sure how to replace, would fheroes2::FontType::largeWhite() work?

Hi @srmoll ! I do not believe that function does not exist and infact most of the places where text.h hasn't been removed is because of these instances of text.h's fontsize call.

oleg-derevenetz (changes requested):
Hi @srmoll Good job! I have only one additional suggestion regarding getting rid of another anonymous literal. Could you please take a look when you have some free time?

src/fheroes2/dialog/dialog_giftresources.cpp
```diff
@@ -63,13 +69,17 @@ struct SelectRecipientsColors
...
     {
         positions.reserve( colors.size() );
+        const fheroes2::Display & display = fheroes2::Display::instance();
+        const fheroes2::Rect box( ( display.width() - giftDialogSize.width ) / 2, ( display.height() - giftDialogSize.height ) / 2, giftDialogSize.width,
+                                  giftDialogSize.height );
 
+        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::CELLWIN, 43 );
         const int32_t colorCount = static_cast<int32_t>( colors.size() ); // safe to cast as the number of players <= 8.
-
+        const int32_t playerContainerWidth = colorCount * sprite.width() + ( colorCount - 1 ) * 22; // original spacing = 22
```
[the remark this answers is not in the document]
srmoll: Thank you for the feedback, just pushed that change

oleg-derevenetz:
@srmoll Thank you very much and welcome to the project contributors!

srmoll:
> @srmoll Thank you very much and welcome to the project contributors!

Thank you @oleg-derevenetz and thank you also to @ihhub and @zenseii for all of their support!

PR #5301, opened 2022-04-18
Reserve half of spell points for dimension door

idshibanov:
Added a new variable controlling how much spell points are reserved for adventure spells, half by default. Reset to 0 if there's nothing to do, so "stuck" heroes should teleport out.


ihhub:

PR #5302, opened 2022-04-18
AI using View All to scout fog

idshibanov:


src/fheroes2/heroes/heroes_base.cpp
```diff
@@ -339,23 +339,23 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res ) const
         return false;
     }
 
-    if ( !HaveSpell( spell ) ) {
+    if ( !HaveSpellPoints( spell ) ) {
```
oleg-derevenetz, line 342: Hi @idshibanov I just wonder, why the order of checks was changed here? Isn't it more logical to first check if the hero has a spell before checking the sufficiency of mana and movement points?

idshibanov: This is to just short-circuit the check. `HaveSpell` loops through both spells and artifacts while move/spell points checks should fail often too.

oleg-derevenetz: I hope this won't affect the spellcasting UI. I took a quick look - it doesn't seem like it should.

oleg-derevenetz:
Hi @idshibanov could you please resolve the merge conflicts when you have free time so that we can proceed with this PR?

oleg-derevenetz:
Hi @idshibanov Merged, thank you!

PR #5305, opened 2022-04-19
Fix MP3 support in SDL1 MSVC packages, minor improvements for workflows

oleg-derevenetz:
`SDL_mixer-1` win32 package doesn't have the `libmpg123`, but it has the `smpeg`.

Plus minor improvements for workflows: speed up installation in the MSVC workflow a bit (don't install and use the `cyg-get` package, use only Chocolatey), update `cache` action to v3.


ihhub:

Issue #5283, opened 2022-04-16
French smallfont c with cedille ç is lowered


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In `Français`:
![image](https://user-images.githubusercontent.com/12501091/163670713-2b269bad-217a-4d96-929e-a6f885341832.png)

Probably this character didn't have the position set or something similar in its font generation, and it wasn't noticed because normal font was used here before. Using commit https://github.com/ihhub/fheroes2/commit/04011c5b5db00c1bfb96dbfdb657f6a9486ddb24 with English assets.

### Save file

Just open config menu.

zenseii:
I'll have a look at this one now.

PR #5313, opened 2022-04-22
Fix French smallfont Cedilla ç

zenseii:
Addresses issue https://github.com/ihhub/fheroes2/issues/5283.


oleg-derevenetz:
@zenseii Thank you!

Issue #5246, opened 2022-04-10
Make more strings use lowercase


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Now that lowercasing of strings were implemented in https://github.com/ihhub/fheroes2/pull/4976 more strings can be made lowercase:

- [ ] Select {Monster}
![image](https://user-images.githubusercontent.com/12501091/162641630-e8b328b9-78da-4a52-a48f-1e65c14d7b6e.png)
- [ ] Recruit {Monster} . There are at least two of them:
![image](https://user-images.githubusercontent.com/12501091/162641659-23e465ab-bf64-4b52-828b-05affe9465ef.png)
![image](https://user-images.githubusercontent.com/12501091/162641669-a8804e4e-e15b-4ef3-8037-990c020439be.png)
- [ ] This could be done for buildings too. Build {Building}:
![image](https://user-images.githubusercontent.com/12501091/162641843-4d738f89-6321-47b0-a957-6357b7a43e27.png)

### Describe the possible solution

You basically just need to use `Translation::StringLower()` like done in https://github.com/ihhub/fheroes2/pull/4976/files#diff-a4e630f2f0de5508d00515b54754d4a2e5db85014707a61735b2bc5ef4024769R287

There might need to be made extra contexts for the various strings to avoid a sentence starting with a lowercased string.

If we want to retain some of these strings in uppercase to keep in line with the OG, then might I suggest making it possible to specify it for different languages.

PR #5321, opened 2022-04-23
Make several strings lowercased with Translation::StringLower()

zenseii:
This addresses issue: https://github.com/ihhub/fheroes2/issues/5246
Following the implementation from this PR: https://github.com/ihhub/fheroes2/pull/4976

This adds many strings to be lowercased. Marking this as a draft because although many strings are easy to see that they should be lowercased, many others might need to be reverted for the moment since we want to have titles in English use uppercase for more than the first word. This is according to the discussion in above issue and general grammar rules.

Didn't want to revert them before publishing because at least now they're documented for an eventual later reference.


zenseii:
I'm closing this because it would take longer to solve the conflicts than to just make a new PR based on the current source.

The changes in this PR are still valid so this closed PR can still be used for reference to find the relevant strings.

PR #5322, opened 2022-04-23
AI should replenish lost spell points more frequently

idshibanov:
Heroes are often running out of spell points especially in case of heavy Dimension door use. Also includes small pathfinding fix.

This won't affect early game.


oleg-derevenetz:
@idshibanov Merged, thank you!

Issue #5237, opened 2022-04-10
Missing Gift button for other assets than English


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

No other language assets have a generated gift button:
![image](https://user-images.githubusercontent.com/12501091/162635273-152f7a63-907a-49d4-ae76-3d7d17d078b6.png)

### Describe the possible solution

Generate a gift button. Same approach as for Battle Only button I'd imagine.

zenseii:
I'll have a look at this one, but I need help to figure out what to call that button in German, French, Polish and Russian.

zenseii:
From our French users on Discord the word "Donner" was proposed for the button.

LeHerosInconnu:
Hello @zenseii,

> From our French users on Discord the word "Donner" was proposed for the button.

Wouldn't "DON" or "CADEAU" be more appropriate?

zenseii:
> Hello @zenseii,
> 
> > From our French users on Discord the word "Donner" was proposed for the button.
> 
> Wouldn't "DON" or "CADEAU" be more appropriate?

Hi @LeHerosInconnu , better late than never to the party for this discussion 😅.

No one has complained since DONNER got implemented in April 😅 But the English word GIFT on the button is something we (or the previous team?) put there and it isn't the best word to use. It would be better to use SEND or DONATE probably.

Note that GIFT refers to the verb and not the noun "a gift". So "to gift someone resources". That's why the verb DONNER was chosen.

LeHerosInconnu:
@zenseii.

> ... But the English word GIFT on the button is something we (or the previous team?) put there and it isn't the best word to use. It would be better to use SEND or DONATE probably.

I would choose "DONATE" or "GIVE". :)

PR #5331, opened 2022-04-24
Generate French 'GIFT' button

zenseii:
This addresses the French part of issue: https://github.com/ihhub/fheroes2/issues/5237

This makes a French 'Gift' button with the text 'Donner'. That's the word that was approved by French-speaking users.

The Gift button generated for English had to be moved to `generateEnglishSpecificImages()`, similar to what was done for the Battle Only button.

Added logic to fill with Evil button color to `getButtonFillingColor()`.

The current state of the generated buttons:
image

I had to refer to some different ICNs for the Evil button generation than than the Good, because the corresponding ones I used for the Good button didn't have the same background color as other evil ICNs (ex. campxtrg and campxtre).

There could be a reason to go back and make the Good button with the counterpart ICNs as the Evil button if this means you could avoid having one switch case for each button and instead choose the ICN according to which interface type is in use. Not doing this would just mean keeping the implementation like it is, I'm just proposing in order to avoid duplication of code whenever there's need to generate both Good and Evil buttons.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -500,7 +500,7 @@ namespace fheroes2
                     Fill( out, offsetXD + 9 - i, offsetY + i, 1, 1, getButtonFillingColor( i == 0, false ) );
                     // Add 'O'
                     const int32_t offsetXO = 10;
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

oleg-derevenetz:
Hi @zenseii Merged, thank you!

PR #5337, opened 2022-04-26
Add Translation Context for Secondary Skill Campaign Bonuses

### LINKED ISSUE #5214: Missing context for skill bonuses for campaigns
### Preliminary checks

### Platform

Other

### Describe the bug

Currently the skill choices for campaign bonuses do not have their own context and in other langauges than English this leads to these strings being too long for the bonus area:
image

The below discussion shows the awards for reference, but every skill should have its award context for future-proofing when more campaigns are added and even in English you could potentially make a campaign with a bonus of Advanced Pathfinding which would be too long:
https://github.com/ihhub/fheroes2/discussions/4541

This should be a good first issue since you could most likely follow this PR's approach:
https://github.com/ihhub/fheroes2/pull/4971/files#diff-14a58fb744cf2a6f1030e616fb38bd64411b3dab4cbdda3a099e7b7062dd0811

### Save file

Here's a save for having completed the first Voyage Home scenario. It has two skill bonuses.
[Voyage Home 1 Complete.zip](https://github.com/ihhub/fheroes2/files/8439996/Voyage.Home.1.Complete.zip)

To test others you can edit a po file and change the first scenario's awards to see how they are in game.

### Additional info

Not really relevant for fixing this issue but currently awards should be no longer than 16-18 characters (whitespace included) depending on what letters are used.

zenseii:
fix #5214

Added all current possible secondary skills to futureproof it.

### DIFF
--- a/src/fheroes2/campaign/campaign_scenariodata.cpp
+++ b/src/fheroes2/campaign/campaign_scenariodata.cpp
@@ -382,6 +382,140 @@ namespace
         }
     }
 
+    const std::string getSecondarySkillCampaignName( const uint32_t secondarySkillId, const uint32_t secondarySkillLevel ) 
+    {
+        switch ( secondarySkillId ) {
+        case Skill::Secondary::ARCHERY:
+            switch ( secondarySkillLevel ) {
+            case Skill::Level::BASIC:
+                return _( "campaignBonus|Basic Archery" );
+            case Skill::Level::ADVANCED:
+                return _( "campaignBonus|Advanced Archery" );
+            case Skill::Level::EXPERT:
+                return _( "campaignBonus|Expert Archery" );
+            }
+        case Skill::Secondary::BALLISTICS:
+            switch ( secondarySkillLevel ) {
+            case Skill::Level::BASIC:
+                return _( "campaignBonus|Basic Ballistics" );
+            case Skill::Level::ADVANCED:
+                return _( "campaignBonus|Advanced Ballistics" );
+            case Skill::Level::EXPERT:
+                return _( "campaignBonus|Expert Ballistics" );
+            }
+        case Skill::Secondary::DIPLOMACY:
+            switch ( secondarySkillLevel ) {
+            case Skill::Level::BASIC:
+                return _( "campaignBonus|Basic Diplomacy" );
+            case Skill::Level::ADVANCED:
+                return _( "campaignBonus|Advanced Diplomacy" );
+            case Skill::Level::EXPERT:
+                return _( "campaignBonus|Expert Diplomacy" );
+            }
+        case Skill::Secondary::EAGLEEYE:
+            switch ( secondarySkillLevel ) {
+            case Skill::Level::BASIC:
+                return _( "campaignBonus|Basic Eagle Eye" );
+            case Skill::Level::ADVANCED:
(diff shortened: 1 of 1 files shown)


src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -382,6 +382,140 @@ namespace
...
+        switch ( secondarySkillId ) {
+        case Skill::Secondary::ARCHERY:
+            switch ( secondarySkillLevel ) {
+            case Skill::Level::BASIC:
+                return _( "campaignBonus|Basic Archery" );
+            case Skill::Level::ADVANCED:
+                return _( "campaignBonus|Advanced Archery" );
+            case Skill::Level::EXPERT:
+                return _( "campaignBonus|Expert Archery" );
+            }
```
oleg-derevenetz, line 396: Hi @zenseii Please do not allow the `switch` to fall through here by adding `break` after this line (and preferably add the `default` case in the inner `switch`es with `assert(0)`, because, I suppose, there cannot be some skill level other than basic, advanced or expert).

oleg-derevenetz: @zenseii Or maybe even better - we can add the `assert(0)` and then `return {}` after every inner `switch`.

zenseii: So basically we `return{}` instead of `break`. I've written it like this for every case:
```cpp
        case Skill::Secondary::ARCHERY:
            switch ( secondarySkillLevel ) {
            case Skill::Level::BASIC:
                return _( "campaignBonus|Basic Archery" );
            case Skill::Level::ADVANCED:
                return _( "campaignBonus|Advanced Archery" );
            case Skill::Level::EXPERT:
                return _( "campaignBonus|Expert Archery" );
            }
            assert( 0 );
            return {};
        case Skill::Secondary::BALLISTICS:
```

oleg-derevenetz: We can also add comments above these assertions explaining them (like "There cannot be other valid values of the skill level").

oleg-derevenetz:
@zenseii

> This addresses issue: https://github.com/ihhub/fheroes2/issues/5214

Small hint: just write "fix #NNNN" or "close #NNNN" and this issue will be linked to this PR automatically (and will be closed as soon as this PR is merged).

src/fheroes2/campaign/campaign_scenariodata.cpp
zenseii: Thanks, done! Now since I added all current secondary skills I imagine we should `assert( 0 )` in the outer `default` statement too since that should never theoretically be executed as well? 

```cpp
                return _( "campaignBonus|Expert Wisdom" );
            }
            // There cannot be other valid values of the skill level
            assert( 0 );
            return {};
        default:
            assert( 0 );
            return Skill::Secondary( secondarySkillId, secondarySkillLevel ).GetName();
        }
        }
```

oleg-derevenetz: We can return an empty string (`{}`) in the `default` branch as well, if it shouldn't be executed anyway.

oleg-derevenetz:
@zenseii Merged, thank you!

PR #5339, opened 2022-04-26

oleg-derevenetz:


zenseii:
Thanks for fixing these @oleg-derevenetz. I'll know not to do these mistakes the next time..

oleg-derevenetz:
@zenseii I would not call it "mistakes" (there are no errors), it's more a code style issues forced by SonarCloud. However, it seems that there are currently errors in all these `generateLanguageSpecificImages()` machinery. Don't you think, for example, that currently app will crash on the failed assertion while attempting to load the `ICN::NON_UNIFORM_GOOD_MIN_BUTTON` with German game resources?

zenseii:
> @zenseii I would not call it "mistakes" (there are no errors), it's more an issues with the SonarCloud-forced code style. However, it seems that there are currently errors in all these `generateLanguageSpecificImages()` machinery. Don't you think, for example, that currently app will crash on the failed assertion while attempting to load the `ICN::NON_UNIFORM_GOOD_MIN_BUTTON` with German game resources?

I made sure to check this before putting it up for review and I just checked again and it doens't happen:
image

This also makes sense considering this line in the code that calls these generated modified assets where it defaults to the English code for generation:
https://github.com/ihhub/fheroes2/blob/7b4dccb000fe6b267494095a64308726cf355e93/src/fheroes2/agg/agg_image.cpp#L656-L673

oleg-derevenetz:
@zenseii What I mean is the following: suppose that we have German resources - not German language is selected in the game, but we have game resources in German (`fheroes2::getResourceLanguage()` returns `fheroes2::SupportedLanguage::German`). In this case `generateGermanSpecificImages()` will be called and, since it doesn't have the code to generate the `ICN::NON_UNIFORM_GOOD_MIN_BUTTON`, `assert(0)` will be called.

zenseii:
> @zenseii What I mean is the following: suppose that we have German resources - not German language selected in the game settings, but we have game resources in German (`fheroes2::getResourceLanguage()` returns `fheroes2::SupportedLanguage::German`). In this case `generateGermanSpecificImages()` will be called and, since it doesn't have the code to generate the `ICN::NON_UNIFORM_GOOD_MIN_BUTTON`, `assert(0)` will be called.

Yes that should be the case, yet as you can see I am using the German resources from the 'EXIT' button being 'ENDE' in the picture.

zenseii:
I managed to get the assert( 0 ) to kick in by setting breakpoints but I'm surprised this didn't happen before when not all assets had a 'battle only' button.

oleg-derevenetz:
@zenseii May be you built the app in with the `NDEBUG` defined - in this case assertions doesn't work.

zenseii:
Hmm I always build debug build with Visual Studio and they worked other times.

oleg-derevenetz:
@zenseii What if you download and run binary from our CI? Will it crash on the "GIFT" button for example? This button is not generated for German resources.

zenseii:
I'll check it now, but I'm sure it will. I'm still surprised it did not crash before when I've been testing it for such a while, but I understand why there were no problems when only the German and English Battle Only buttons were generated because this was the code then:
```cpp
      void generateLanguageSpecificImages( int id )
        {
            switch ( fheroes2::getResourceLanguage() ) {
            case fheroes2::SupportedLanguage::German:
                generateGermanSpecificImages( id );
                break;
            default:
                generateEnglishSpecificImages( id );
                break;
            }
```

This made me realize that any button that gets moved from `LoadModifiedICN()` to `generateEnglishSpecificImages()` will need to be generated for every langauge that has been included in `generateLanguageSpecificImages()`.

I can do this for the buttons so far generated, but it would seem like some logic to avoid this would be needed in order to avoid having to always generate for every langauge at once.

PR #5340, opened 2022-04-27
Add placeholder buttons to German, Italian and Polish

zenseii:
This addresses the assertion crashes discovered in https://github.com/ihhub/fheroes2/pull/5339.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -395,6 +395,34 @@ namespace fheroes2
                     Blit( GetICN( ICN::BTNNEWGM, 4 + i ), 36 - i, 23, out, 78 - i, 23, 10, 14 );
                 }
                 break;
+            case ICN::BTNGIFT_GOOD:
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::TRADPOST, 17 + i );
+                    // clean the button
+                    Fill( out, 33, 5, 31, 16, getButtonFillingColor( i == 0 ) );
+                }
+                break;
+            case ICN::BTNGIFT_EVIL:
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::TRADPOSE, 17 + i );
+                    // clean the button
+                    Fill( out, 33, 5, 31, 16, getButtonFillingColor( i == 0, false ) );
+                }
+                break;
+            case ICN::NON_UNIFORM_GOOD_MIN_BUTTON:
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::RECRUIT, 4 + i );
+                    // Clean the button and leave 'M'
+                    Fill( out, 31 - 2 * i, 5 + i, 25, 15, getButtonFillingColor( i == 0 ) );
+                    Fill( out, 29 - 2 * i, 17 + i, 2, 2, getButtonFillingColor( i == 0 ) );
+                }
+                break;
             default:
                 // You're calling this function for non-specified ICN id. Check your logic!
                 assert( 0 );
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
I had thoughts about making these functions return `false` if they are unable to generate images (and fallback to the `generateEnglishSpecificImages()` in this case) but may be the variant with empty buttons is better.

zenseii:
> I had thoughts about making these functions return `false` if they are unable to generate images (and fallback to the `generateEnglishSpecificImages()` in this case) but may be the variant with empty buttons is better.

I yours would be the best solution going forwards because if we start generating buttons for more languages then the work of copying empty buttons grows immensly. This PR was simply to a quick way to make sure that current snapshots don't crash on other assets than English.

I see that I have to do minor adjustments to what button is chosen as empty button for these languages since right now they aren't cleaned due to asset differences.
E.g. German assets right now:
image


It doesn't look any worse than what it was before so not really sure it matters.

oleg-derevenetz:
> Yours would be the best solution going forwards because if we start generating buttons for more languages then the work of copying empty buttons grows immensly.

The drawback of this is that default-generated buttons "in English" will contain a mess. Now they are at least clean.

oleg-derevenetz:
> So if false is returned a button of the same size that is empty will be used instead of creating such garbled buttons?

We should create a separate function then (as a fallback function) that will create these clean images/buttons for ALL images for which `generateLanguageSpecificImages()` may be called.

zenseii:
> > So if false is returned a button of the same size that is empty will be used instead of creating such garbled buttons?
> 
> We should create a separate function then (as a fallback function) that will create these clean images/buttons for ALL images for which `generateLanguageSpecificImages()` may be called. In this case we can make all other `generateXXXSpecificImages()` optional.

Yes so we basically have button 11 and 12 in `system` and `systeme` which can work as placeholder for that size of buttons since both good and evil exist, but I'm not sure there are already empty for the two other sizes of buttons ( i.e. big buttons like in the main menu and the smaller MAX button in recruit dialog).
image

zenseii:
> > So if false is returned a button of the same size that is empty will be used instead of creating such garbled buttons?
> 
> We should create a separate function then (as a fallback function) that will create these clean images/buttons for ALL images for which `generateLanguageSpecificImages()` may be called. In this case we can make all other `generateXXXSpecificImages()` optional.

@oleg-derevenetz, should we make this function in a separate PR and still merge this one as well as soon as it is declared ready since the changes in this PR are still valid?

oleg-derevenetz:
@zenseii we can make all `generateXXXSpecificImages()` like this:

```cpp
bool generateXXXSpecificImages(...)
{
  switch (...) {
  case SOME_SUPPORTED_ID:
    ...
    return true;
  default:
    break;
  }
  return false;
} 
```

So the only change will be all these `return true/false`, and place all logic related to the generation of empty buttons to a separate fallback function. This should help to avoid the code duplication. You can make that in this PR, I suppose that this will be easier than repeat the same empty image generation logic for each language.

zenseii:
Hi @oleg-derevenetz ! I've added the function to generate the missing button. I've looked through the code and I'm not sure if the complaint about recursion is still valid.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -284,7 +284,38 @@ namespace fheroes2
             }
         }
 
-        void generateEnglishSpecificImages( const int id )
+        void generateMissingButtons( const int id )
+        {
+            switch ( id ) {
+            case ICN::BTNGIFT_GOOD:
```
oleg-derevenetz, line 290: It would be nice to also add generation of some "default" variant of the `ICN::BTNBATTLEONLY` image here. I do not know for which languages there are localizations of original game resources, but in theory in the future a new languages may appear.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -637,31 +688,40 @@ namespace fheroes2
...
             }
+            return false;
         }
 
         void generateLanguageSpecificImages( int id )
         {
             switch ( fheroes2::getResourceLanguage() ) {
             case fheroes2::SupportedLanguage::German:
-                generateGermanSpecificImages( id );
+                if ( !generateGermanSpecificImages( id ) ) {
```
oleg-derevenetz, line 702: To reduce the code duplication, we can do something like this:

```cpp
// Language-specific image generators, may fail
switch ( fheroes2::getResourceLanguage() ) {
case fheroes2::SupportedLanguage::German:
  if ( generateGermanSpecificImages( id ) ) {
    return;
  }
  break;
[...]
}

// Image generator of a last resort, must provide the generation of the "default" variant
// for all image ids for which this function can be called, and must not fail.
generateDefaultImages( id );
```

src/fheroes2/agg/agg_image.cpp
```diff
@@ -284,7 +284,38 @@ namespace fheroes2
             }
         }
 
-        void generateEnglishSpecificImages( const int id )
+        void generateMissingButtons( const int id )
```
oleg-derevenetz, line 287: Since this function can (in future) generate more than just buttons, I propose to call it `generateDefaultImages` instead.

oleg-derevenetz (changes requested):
Hi @zenseii I just left a few suggestions, please take a look when it's convenient for you.

Regarding the recursion warning, it's legitimate, since all these functions call the `GetICN()`, which in turn may call the `GetMaximumICNIndex()`, which in turn may call the `LoadModifiedICN()`, which in turn may call the `generateLanguageSpecificImages()`, which in turn may call these functions.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -637,31 +688,40 @@ namespace fheroes2
...
                 break;
             case fheroes2::SupportedLanguage::Italian:
-                generateItalianSpecificImages( id );
+                if ( !generateItalianSpecificImages( id ) ) {
+                    generateMissingButtons( id );
+                }
                 break;
             default:
-                generateEnglishSpecificImages( id );
+                if ( !generateEnglishSpecificImages( id ) ) {
```
oleg-derevenetz, line 722: BTW, since we now will have the generator of a last resort, we can call this not for `default` case, but for `fheroes2::SupportedLanguage::English`. In this case for a some potential new language the placeholder images will be generated by default, instead of the unreadable mess.

zenseii:
@oleg-derevenetz, Thanks for the advice. I've implemented them and now what remains for me to do is to look at solving the recursion.

I also thought that we might want to have at least something be printed on top of these now empty buttons. To work across different languages we could either use `fheroes2::Text` since `ui_text.h` is already included or programatically use a smaller sprite that could be relevant for the button and be understandable across languages. Like for MIN button it could be a downwards arrow or triangle, for battle only it could be a sword etc.

The point is to have at least something so that the players will not be wondering why there is an empty button where there used to be a MAX button before for example.

zenseii:
> @zenseii The recursion is not an issue if these functions will not try to call the `GetICN()` with the ID of an image they should generate (and normally they shouldn't do that). 


I see and that is true. Should I try to implement a check that this does not happen?



> Regarding the content of the default buttons, we can add it in a separate PR when there will be any ideas. The purpose of this PR is to fix the assertion failures in certain (quite widespread) situations.

Yes it should be better for another PR since it will require some input from other members.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -284,7 +284,48 @@ namespace fheroes2
...
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::RECRUIT, 4 + i );
+                    Fill( out, 11 - i, 4 + i, 52, 17, getButtonFillingColor( i == 0 ) );
+                }
+                break;
+            default:
+                // You're calling this function for non-specified ICN id. Check your logic!
+                assert( 0 );
```
zenseii, line 323:
```suggestion
            default:
                // You're calling this function for non-specified ICN id. Check your logic!
                // Did you add a new image for one language without generating a default
                // for other languages?
                assert( 0 );
```

oleg-derevenetz (approved):

oleg-derevenetz:
@zenseii merged, thank you!

PR #5344, opened 2022-04-28
Add the toolset script to extract additional game resources from the original distribution of HoMM2

### LINKED ISSUE #5310: Automate importing of GoG flac music
### Preliminary checks

At the moment when we want to use flac music downloaded from GoG, we have to download the soundtrack, uncompress it, change the name of each track and then copy all .flac files to music directory. This procedure is carried out manually. This issue is related to the issue #4233 

The procedure described above could be automatized using powershell script. The script would be stored in the root directory of fheroes2 (like extract_homm2_resources.ps1), and it would use zip file containing the soundtrack in flac format downloaded from GoG. It would allow to specify the location of the zip file and also would detect it, if it was copied to the root directory.

### Additional info

The script we could use to carry out the task could be like this:

```PowerShell
$ErrorActionPreference = "Stop"

try {
    Write-Host "This script will extract and copy music files with flac extension to music directory" -ForegroundColor Yellow
    $flacZipUri = Get-ChildItem -Path ".\*flac*.zip" | Select-Object -ExpandProperty FullName
    $promptMsg = $null
    $flacZipExtractedDir = "extracted_flac"
    $musicDir = "music"
    if ($flacZipUri) {
        Write-Host "The following zip file was found:"
        Write-Host "$flacZipUri" -ForegroundColor Yellow
        $promptMsg = "If it contains the games soundtrack press 'enter', otherwise type in the location of zip file containing the soundtrack"
    } else {
        $promptMsg = "Type in the location of zip file containing the soundtrack"
    }
    $flacZipPrompt = Read-Host $promptMsg
    if (-not $flacZipPrompt) {
        $flacZipPrompt = $flacZipUri
    }
    if ($flacZipPrompt -and (Test-Path -Path "$flacZipPrompt" -PathType Leaf)) {
        if (Test-Path -Path ".\$flacZipExtractedDir" -PathType Container) {
            Remove-Item -Path ".\$flacZipExtractedDir" -Recurse -Force
        }
        Expand-Archive -Path "$flacZipPrompt" -DestinationPath ".\$flacZipExtractedDir" -Force
        if (Test-Path -Path ".\$musicDir" -PathType Container) {
            Remove-Item -Path ".\$musicDir" -Recurse -Force
        }
        New-Item -Path ".\$musicDir" -ItemType Directory -Force | Out-Null
        $extension = "flac"
        Get-ChildItem -Path ".\$flacZipExtractedDir\*.${extension}" -Recurse | 
        ForEach-Object { $orderNumber = (Select-String -InputObject $_.Name -Pattern "[0-9]{2}").Matches.Value;
                         $newName = "Track${orderNumber}.${extension}";
                         #Rename-Item -Path $_.FullName -NewName $newName;
                         Copy-Item -Path $_.FullName -Destination ".\$musicDir\$newName"
                       }
        if (Test-Path -Path ".\$flacZipExtractedDir" -PathType Container) {
            Remove-Item -Path ".\$flacZipExtractedDir" -Recurse -Force
        }
        $flacZipName = Get-ChildItem -Path "$flacZipPrompt" -Name
        if (Test-Path -Path ".\$flacZipName" -PathType Leaf) {
            Remove-Item -Path ".\$flacZipName" -Force
        }
    } else {
        Write-Host -ForegroundColor Red "FATAL ERROR: Unable to find zip file with soundtrack"
        return
    }
} catch {
    Write-Host -ForegroundColor Red (-Join("FATAL ERROR: ", ($_ | Out-String)))
} finally {
    Write-Host "Press any key to exit..."
    [void][System.Console]::ReadKey($true);
}
```

oleg-derevenetz:
close #5310

It even has the GUI with drag&drop support. For now, only extraction of FLAC or MP3 OST files from the GOG archive is supported - extraction is intentionally performed to a separate directory (`music.flac` or `music.mp3`) in order not to mix different types of soundtracks in the same directory and allow the user to decide which one he wants. The script was tested with PowerShell 2.0, but not on Windows 7 (I don't have one).


zenseii:
> close #5319
> 
> It even has the GUI. For now, only extraction of FLAC music files from the GOG archive is supported.

Hi, @oleg-derevenetz! Thanks for making this. Also, I believe you tagged the wrong issue/PR in the original message.

oleg-derevenetz:
Hi @ihhub @zenseii Does anyone have a Windows 7 machine to test this better?

zenseii:
> Hi @ihhub @zenseii Does anyone have a Windows 7 machine to test this better?

I have. I can run a test now.

zenseii:
Hi @oleg-derevenetz ! On my 32bit win 7 machine I grabbed the both .exe and .zip snapshots from this PR, and the resources script first gives me "WARNING: Unable to determine the HoMM2 diretory" and then when I manually enter the correct directory, which anyways was the standard "C:\GOG Games\HoMM 2 Gold" I get "FATAL ERROR: Unable to find the HoMM2 directory".

oleg-derevenetz:
> The toolset script works fine for both flac and mp3, only that the resulting folder for flac files is named `music.flac`, and for mp3 `music.mp3`.

This is intentional. Currently there is no way for a user to manually select the desired music format - other than replacing the directory itself. So the user will need to rename one of these folders to `music` by hand.

oleg-derevenetz:
> I get this message when running the fheroes 2 installer .exe.

Yes. You need to have a proper installation of original HoMM2 to extract mandatory resources (animations, data files, maps), not just the archives or something like that.

zenseii:
> > I get this message when running the fheroes 2 installer .exe.
> 
> Yes. You need to have a proper installation of original HoMM2 to extract mandatory resources (animations, data files, maps), not just the archives or something like that. Basically, you need a directory which contains `HEROES2.EXE`, `DATA`, `MAPS` and optionally `ANIM` and `MUSIC`.

Yes, I have all of these in the folder. I tried running the fheroes2 installation and resource extraction scripts right after having done an installation of the Polish dosbox gog file downloaded just now from GOG.

oleg-derevenetz:
> and "Heroes2w.exe"

That's the problem. Installation script expects DOS file too, because all installations I have seen before included either `HEROES2.EXE` or both `HEROES2.EXE` and `HEROES2W.EXE`.

zenseii:
> > and "Heroes2w.exe"
> 
> That's the problem. Installation script expects DOS file, because all installations I have seen before included either `HEROES2.EXE` alone or both `HEROES2.EXE` and `HEROES2W.EXE`.

There is only `Heroes2w.exe` in this one so yes that must be the problem. Anyways the script for this PR works as intended. It was only a coincidence that I chose to grab the Polish version from GoG.

ihhub:

PR #5349, opened 2022-04-29
Populate resolution list based till the lowest resolution

### LINKED ISSUE #5293: Windows Size
### Your question

Hello, 

I've just installed the FHeroes 2 with GOG full game.
But when I want to play the game, the windows is too big:

When I'm on windowed mode, no matter the resolution I choose, the windows heigh it too big, and I can't see the bottom of the windows.
In fullscreen, the game is on 16:9 size, but it's cropped on the side + bottom too. 

I'm on Windows 11, and my monitor is a 32:9 5120*1440p resolution
The resolution I can choose are: 
5120 x1440p (WAY)too big on side + bottom)
1920x1440 (WAY too big on bottom)
1920x1200 (too big on bottom)
1600x1200 (too big on bottom)

There is any way to fix it ? Like manual windows resize ? Or manual resolution ? 

### Additional info

_No response_

ihhub:
SDL really doesn't work properly with resolutions so we have to do some hacky way like this.

close #5293

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -102,12 +102,12 @@ namespace
             = { { 640, 480 },   { 800, 600 },  { 1024, 768 },  { 1152, 864 }, { 1280, 600 }, { 1280, 720 },  { 1280, 768 }, { 1280, 960 },
                 { 1280, 1024 }, { 1360, 768 }, { 1400, 1050 }, { 1440, 900 }, { 1600, 900 }, { 1680, 1050 }, { 1920, 1080 } };
 
-        const fheroes2::Size currentResolution = resolutions.back();
-        for ( size_t i = 0; i < possibleResolutions.size(); ++i ) {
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

ihhub:
Hi @oleg-derevenetz , I can see that the build for PS Vita fails with this error:
```
system.cpp: In function 'int System::MakeDirectory(const string&)':
system.cpp:114:12: error: 'sceIoMkdir' was not declared in this scope
  114 |     return sceIoMkdir( path.c_str(), 0777 );
      |            ^~~~~~~~~~
```
We haven't changed anything in the past. This looks weird.

oleg-derevenetz:
Hi @ihhub maybe something was changed in the include files of the vitasdk due to some update. Honestly, I don't know why use this function at all, because it seems that vitasdk has "normal" `mkdir()` in the `sys/stat.h`. I can try to remove this call, but I suppose it is there for a reason, so this change should be tested on the real device.

Issue #5244, opened 2022-04-10
Account for other languages with two lines for several headers


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Some languages have longer headers than for English and need two lines:

- [ ] Every option in the Adventure Map options (This just shows for Polish - other languages need the others):
![image](https://user-images.githubusercontent.com/12501091/162637316-742ecb0f-beea-43f3-a5a9-c1ca83d8d52b.png)
- [ ] Class choice when starting a scenario:
![image](https://user-images.githubusercontent.com/12501091/162637569-4a9f1c9a-1288-427c-a7c4-ac269da6a57b.png)
- [ ] Others?

### Describe the possible solution

This should probably just need the same approach used for Map Type, Map Difficulty and for the Combat Options.

PR #5350, opened 2022-04-29
Add double-lined headers to adventure map options.

zenseii:
This addresses one part of issue: https://github.com/ihhub/fheroes2/issues/5244

image

This also swaps position of "Settings" and "Experimental" in the main menu's system options since that fits in with the other headers: Language and Resolution.

image

Replaced use of `text.h` with `ui_text.h`.


ihhub:
@zenseii , a bunch of thanks for fixing this issue!

PR #5352, opened 2022-04-30
Add Ukrainian and Belarusian languages

zenseii:
This adds empty PO files to these langauges and adds their plural forms to `translations.cpp`.


ihhub:
@zenseii , thanks a lot for handling translations!

shprotru:
@zenseii prefix uk should be ua

ihhub:
Hi @shprotru , are you sure about this? We are talking about language abbreviation, not internet domain :)

shprotru:
@ihhub hm, by this standart used `uk` is correct: [List_of_ISO_639-1_codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)

PR #5353, opened 2022-04-30
Fix build on PSV

oleg-derevenetz:


ihhub:
@oleg-derevenetz , great thanks for fixing the build!

PR #5354, opened 2022-05-01
Add Belarusian, Bulgarian and Ukrainian languages support

ihhub:
These languages need to have some translation in order to appear in the list of supported languages.


ihhub:
Hi @zenseii , there was a small mistake: Ukrainian language abbreviation is **ua** instead of **uk** :)

zenseii:
> Hi @zenseii , there was a small mistake: Ukrainian language abbreviation is **ua** instead of **uk** :)

That was weird. I checked the ISO abbreviations and it was `uk`:

image


Where did you find the correct one?

ihhub:
@zenseii , hmm... I think I misunderstood... ua is for website domain. You're right! My bad...

PR #5356, opened 2022-05-01
Redesign Hot Key event file handling

ihhub:
Now this file should be generated on a first startup of the game. This is a list of changes done in this pull request:
- do proper naming of events for easier reading
- remove some unused events
- the file is generated if no file is present
- the file does not require to enter magical numeric numbers but human-readable strings
- remove file copying stage for Visual Studio. We do not need to copy configuration files anymore
- decouple the code by moving it to a separate header file


src/fheroes2/game/game_hotkeys.h
```diff
@@ -0,0 +1,134 @@
...
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ ***************************************************************************/
+
+#pragma once
+
+#include <string>
+
+namespace Game
+{
+    enum HotKeyEvent : int32_t
```
oleg-derevenetz, line 27: As far as I can see, we can replace this `enum` by the `enum class` - there is no need to make arithmetic or bitwise operations with members of this `enum`, right?

oleg-derevenetz (commented):
Hi @ihhub I just left one optional proposal, please take a look when you have some free time.

src/fheroes2/game/game_hotkeys.h
ihhub: It is used as index to access **hotKeyEventInfo** array elements.

PR #5357, opened 2022-05-02
Encourage AI to explore more

ihhub:
Current AI does fog discovery is nothing else to do which makes AI heroes to walk around visible territory and not to explore the map. This is a weak strategy as uncovering fog usually gives more resources and advantage. Now AI heroes have fog discovery value which is rated very low but still encourages them to explore.

With such change we can introduce a new type of AI heroes: explorers who would focus on fog discovery.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1161,7 +1161,7 @@ namespace AI
 
         double fogDiscoveryValue = getFogDiscoveryValue( hero );
         const int fogDiscoveryTarget = _pathfinder.getFogDiscoveryTile( hero );
-        if ( fogDiscoveryTarget >=0 ) {
+        if ( fogDiscoveryTarget >= 0 ) {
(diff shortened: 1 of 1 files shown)


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1089,26 +1136,7 @@ namespace AI
...
-
-                    if ( castle && ( castle->GetGarrisonStrength( &hero ) <= 0 || castle->GetColor() == hero.GetColor() ) )
-                        value -= dangerousTaskPenalty / 2;
-                    else
-                        value -= dangerousTaskPenalty;
-                }
+                getObjectValue( node.first, dist, value );
 
                 if ( dist > leftMovePoints ) {
                     // Distant object which is out of reach for the current turn must have lower priority.
```
idshibanov, line 1142: You moved distance penalty logic into the `getObjectValue` call but left this here, so distance penalty will be applied twice.

ihhub: You're absolutely right!

idshibanov (approved):

PR #5367, opened 2022-05-03
Properly extract animations from the GOG distribution of HoMM2 for Windows

oleg-derevenetz:
They reside not in `ANIM` but (for some reason) in `HEROES2/ANIM`.


ihhub:

Issue #5140, opened 2022-03-19
Text suggestions from recent PR


catadropa:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

On PR #5081, I made some comments which weren't picked up until after the PR was merged. Creating an issue here as advised, to track those suggestions.

1. [First comment, line 794](https://github.com/ihhub/fheroes2/pull/5081/files/1e2ddb5533b51578c8948d3743bb7e74c1134f86#diff-12416991db2e725cff5940191fe9b00e96db71af67ea53e4724d6ed09d2933fd) -- is this true, will the monsters always run away or is it possible that they attack?
2. [Second comment, line 808](https://github.com/ihhub/fheroes2/pull/5081/files/1e2ddb5533b51578c8948d3743bb7e74c1134f86#diff-12416991db2e725cff5940191fe9b00e96db71af67ea53e4724d6ed09d2933fd) -- remove "always"? Or " will join your army if you ask."? But maybe this follows OG.
3. [Third comment, line 511](https://github.com/ihhub/fheroes2/pull/5081/files/1e2ddb5533b51578c8948d3743bb7e74c1134f86#diff-14a58fb744cf2a6f1030e616fb38bd64411b3dab4cbdda3a099e7b7062dd0811) -- swap amount/additional: better as "...will have %{amount} additional %{resource} at..."?
4. [Fourth comment, line 534](https://github.com/ihhub/fheroes2/pull/5081/files/1e2ddb5533b51578c8948d3743bb7e74c1134f86#diff-14a58fb744cf2a6f1030e616fb38bd64411b3dab4cbdda3a099e7b7062dd0811) -- Better as: "...will have +%{count} %{skill} at..."? (note +-sign)

### Describe the possible solution

Please review suggestions, thanks.

zenseii:
I think these corrections sound good. For the second comment I think it would be more correct to say "Will always be willing to join your army.". Not sure where this message appears, might have to do with scouting on neutral armies.

zenseii:
One addtion to this:
Spellbook written in one word in "Close spellbook" - should it also not be magic book?

ihhub:
Most of things are addressed recently. Closing this pull request for now.

PR #5369, opened 2022-05-04
Update campaign UI

### LINKED ISSUE #5364: Do not create a scenario completion save by the end of a last scenario in a campaign
### Preliminary checks

### Platform

Windows

### Describe the bug

We create a scenario completion save by the end of a last scenario in a campaign which is incorrect as loading it instantly brings us to high scores window.

### Save file

Do not save at the end of a campaign.

### Additional info

_No response_

ihhub:
There are many changes in this pull request because one thing were dependent on another and so on. The changes are:
- remove code duplication for Monster sprite rendering by introducing `ui_monster` files
- all campaign related data is based on int32_t instead of uint32_t. This was needed to avoid useless static_casts and support negative values for some bonuses
- change some descriptions of awards and bonuses (relates to #5140)
- do not auto save after the completion of the last scenario in a campaign (close #5364)
- add images and full
- fix some campaign bonuses, some were incorrect
- display None if no awards available for the current scenario
- add hot keys to select bonuses
- fix rendering of campaign UI window as it was very slow on response on changing bonuses
- fix missing kingdom income value for scenarios with additional bonus resources
- game interface for campaign window must be unique and should not depend on overall game interface


src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -696,41 +696,42 @@ namespace Campaign
...
-            StringReplace( description, "%{artifact}", Artifact( static_cast<int>( _subType ) ).GetName() );
+            StringReplace( description, "%{artifact}", Artifact( _subType ).GetName() );
             return description;
         }
         case ScenarioBonusData::RESOURCES: {
-            std::string description( _( "The kingdom will have additional %{amount} %{resource} at the start of the scenario." ) );
-            StringReplace( description, "%{amount}", std::to_string( _amount ) );
-            StringReplace( description, "%{resource}", Resource::String( static_cast<int>( _subType ) ) );
+            std::string description( _amount > 0 ? _( "The kingdom will have by %{amount} more %{resource} at the start of the scenario." )
+                                                 : _( "The kingdom will have by %{amount} less %{resource} at the start of the scenario." ) );
```
zenseii, line 704:
```suggestion
            std::string description( _amount > 0 ? _( "The kingdom will receive %{amount} additional %{resource} at the start of the scenario." )
                                                 : _( "The kingdom will have %{amount} less %{resource} at the start of the scenario." ) );
```

src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -700,7 +700,7 @@ namespace Campaign
             return description;
         }
         case ScenarioBonusData::RESOURCES: {
-            std::string description( _amount > 0 ? _( "The kingdom will have by %{amount} more %{resource} at the start of the scenario." )
+            std::string description( _amount > 0 ? _( "The kingdom will receive by %{amount} more %{resource} at the start of the scenario." )
```
zenseii, line 703: The problem here and elsewhere is that "by" is not correct in this nor in case there is less.  Also "additional" is better than "more" in this case:
"The kingdom will receive 500 additional gold at the start of the scenario." This works if the resource is also one unit.

```suggestion
            std::string description( _amount > 0 ? _( "The kingdom will receive by %{amount} more %{resource} at the start of the scenario." )
```

src/fheroes2/campaign/campaign_scenariodata.cpp
```diff
@@ -700,7 +700,7 @@ namespace Campaign
...
                                                  : _( "The kingdom will have by %{amount} less %{resource} at the start of the scenario." ) );
```
zenseii, line 704: Same as the above comment
```suggestion
                                                 : _( "The kingdom will have %{amount} less %{resource} at the start of the scenario." ) );
```

oleg-derevenetz (approved):

files/lang/be.po
```diff
@@ -1083,10 +1083,10 @@ msgstr ""
 msgid "\"%{spell}\" spell will be carried over the scenario."
 msgstr ""
 
-msgid "%{hero} hero can be hired in the scenario."
+msgid "%{hero} can be hired in the scenario."
 msgstr ""
 
-msgid "Defeated %{hero} hero will not appear in the subsequent scenarios."
+msgid "Defeated %{hero} will not appear in the subsequent scenarios."
```
zenseii, line 1089: This would be better as ex. "Lord Corlagon has been defeated and will not appear in the ....."
```suggestion
msgid "%{hero} has been defeated and will not appear in the subsequent scenarios."
```

ihhub: It is not always true as the hero could be defeated few scenarios ago.

src/fheroes2/campaign/campaign_data.cpp
```diff
@@ -800,38 +810,41 @@ namespace Campaign
...
-            std::string description( _( "%{hero} hero can be hired in the scenario." ) );
-            StringReplace( description, "%{hero}", Heroes( static_cast<int>( _subType ), 0 ).GetName() );
+            std::string description( _( "%{hero} can be hired in the scenario." ) );
+            StringReplace( description, "%{hero}", Heroes( _subType, 0 ).GetName() );
             return description;
         }
         case CampaignAwardData::TYPE_DEFEAT_ENEMY_HERO: {
-            std::string description( _( "Defeated %{hero} hero will not appear in the subsequent scenarios." ) );
-            StringReplace( description, "%{hero}", Heroes( static_cast<int>( _subType ), 0 ).GetName() );
+            std::string description( _( "Defeated %{hero} will not appear in the subsequent scenarios." ) );
```
zenseii, line 841: Ex. "Lord Corlagon has been defeated and will not appear in the subsequent scenarios."
This award only appears once you've defeated that hero in a scenario in a campaign.
```suggestion
            std::string description( _( "%{hero} has been defeated and will not appear in the subsequent scenarios." ) );
```

Issue #4515, opened 2021-10-28
Add a Romanian translation


pflugs30:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

No Romanian translation exists

### Describe the possible solution

Add a Romanian translation

dimag0g:
@pflugs30 Do you have anyone in mind who is capable of doing the work?

pflugs30:
> @pflugs30 Do you have anyone in mind who is capable of doing the work?

I created this issue to track work done by **BlueRider** from the FreeHeroes2 Discord server.

PR #5372, opened 2022-05-04
Add initial support for Romanian

zenseii:
Relates to https://github.com/ihhub/fheroes2/issues/4515.
This adds intial support for Romanian.
The special diacritics `ĂăÂâÎîȘșȚț` in their CP1250 encodings are replaced by `AaAaIiSsTt`.

### DIFF
--- a/src/fheroes2/gui/ui_font.cpp
+++ b/src/fheroes2/gui/ui_font.cpp
@@ -42,42 +42,42 @@ namespace
 
             original.resize( 96 );
             original.insert( original.end(), 128, original[0] );
-            original[108] = original[51];
-            original[111] = original[58];
-            original[124] = original[83];
-            original[127] = original[90];
-            original[131] = original[44];
-            original[133] = original[33];
+            original[140 - 32] = original[83 - 32];
+            original[143 - 32] = original[90 - 32];
+            original[156 - 32] = original[115 - 32];
+            original[159 - 32] = original[122 - 32];
+            original[163 - 32] = original[76 - 32];
+            original[165 - 32] = original[65 - 32];
             // Uppercase S with cedilla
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/ui_font.cpp
```diff
@@ -1845,6 +1845,39 @@ namespace
             updateSmallFontLetterShadow( font[75] );
         }
     }
+
+    void generateRomanianAlphabeth( std::vector<std::vector<fheroes2::Sprite>> & icnVsSprite )
```
zenseii, line 1849:
```suggestion
    void generateRomanianAlphabet( std::vector<std::vector<fheroes2::Sprite>> & icnVsSprite )
```

src/fheroes2/gui/ui_font.cpp
```diff
@@ -1873,6 +1906,9 @@ namespace fheroes2
         case SupportedLanguage::Norwegian:
             generateNorwegianAlphabet( icnVsSprite );
             break;
+        case SupportedLanguage::Romanian:
+            generateRomanianAlphabeth( icnVsSprite );
```
zenseii, line 1910:
```suggestion
            generateRomanianAlphabet( icnVsSprite );
```

src/fheroes2/gui/ui_font.cpp
```diff
@@ -48,18 +48,38 @@ namespace
             original[127] = original[90];
```
ihhub, line 48: Can we please then use the same way of indexing for polish letters as well to be consistent? For example, 159 - 32 instead of 127. In this case it would be much easier to read the code.

ihhub:
@zenseii , thanks a lot for adding this language!

PR #5376, opened 2022-05-05
Add HotKey window and expand the list of system options

ihhub:
image
image

### DIFF
--- a/src/fheroes2/dialog/dialog_game_settings.cpp
+++ b/src/fheroes2/dialog/dialog_game_settings.cpp
@@ -151,7 +151,7 @@ namespace
 
     void drawHotKeyOptions( const fheroes2::Rect & optionRoi )
     {
-        drawOption( optionRoi, _( "Hot Keys" ), _( "In-game" ), ICN::CSPANEL, 4 );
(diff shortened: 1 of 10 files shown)


src/fheroes2/dialog/dialog_game_settings.cpp
```diff
@@ -359,17 +431,28 @@ namespace fheroes2
...
                 Dialog::ExtSettings( false );
                 windowType = SelectedWindow::Configuration;
                 break;
+            case SelectedWindow::HotKeys:
+                fheroes2::openHotkeysDialog();
+                windowType = SelectedWindow::Configuration;
+                break;
+            case SelectedWindow::CursorType:
+                conf.setMonochromeCursor( !conf.isMonochromeCursorEnabled() );
+                ::Cursor::Refresh();
```
oleg-derevenetz, line 447: Why we don't call this from the `Settings::setMonochromeCursor()` - after all , we would already be calling `Cursor::setMonochromeCursor()` from it?

oleg-derevenetz (commented):
Hi @ihhub I left just one question, could you please take a look when you have time?

oleg-derevenetz (approved):

PR #5377, opened 2022-05-05
Do not allow higher-level Mage Guilds to be built if the previous-level Mage Guild has not been built yet

### LINKED ISSUE #5232: Cannot build upgraded mausoleum in Fool's Gold scenario

oleg-derevenetz:
fix #5232

Previously, AI could in some situations "jump over" the levels of the Mage Guild during construction, which in turn led to unrecoverable violations of the requirements for the construction of buildings (like with Mausoleum in the related issue).

Related to #5263.

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -1182,50 +1182,61 @@ u32 Castle::GetBuildingRequirement( u32 build ) const
     return requirement;
 }
 
-/* check allow buy building */
-int Castle::CheckBuyBuilding( u32 build ) const
+int Castle::CheckBuyBuilding( const uint32_t build ) const
 {
-    if ( build & building )
+    if ( build & building ) {
         return ALREADY_BUILT;
+    }
 
     switch ( build ) {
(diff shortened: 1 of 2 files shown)


ihhub:
@oleg-derevenetz , thanks a lot for fixing this bug!

PR #5378, opened 2022-05-05
Battle Only mode: always assign a random spell to the Spell Scroll

oleg-derevenetz:
To fix the crash caused by an assertion failure.


ihhub:
@oleg-derevenetz , a bunch of thanks for the fix!

PR #5381, opened 2022-05-06
Make AI to be aware about map conditions

ihhub:
Map conditions must affect the behavior of AI.


src/fheroes2/world/world_loadmap.cpp
```diff
@@ -575,6 +575,24 @@ bool World::LoadMapMP2( const std::string & filename )
...
+    // Make AI aware of a special castle which is a winning condition for a human player.
+    if ( ( conf.ConditionWins() & GameOver::WINS_TOWN ) == GameOver::WINS_TOWN ) {
+        Castle * castle = vec_castles.Get( conf.WinsMapsPositionObject() );
+        assert( castle != nullptr );
+        if ( castle != nullptr ) {
+            castle->SetModes( Castle::CAPITAL );
+        }
+    }
+
+    // Make AI aware of a special castle to be defended in order not to loose it.
```
oleg-derevenetz, line 587: AFAIK there can not be a map with `LOSS_TOWN` and town in question belonging to a AI-controlled player - I just tried to make such a map, and in OG you will lose at the end of the first turn (and in fheroes2 you will lose immediately). Also this condition is not respected for multiplayer maps (both in OG and in feroes2) - it is just ignored and everything looks like the `LOSS_ALL` condition is in effect. However, may be the `CAPITAL` flag is used as a suggestion for the AI to capture this particular human-controlled castle ASAP?

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few notes regarding this PR, could you please take a look when it is convenient for you?

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -698,6 +699,15 @@ void World::ProcessNewMap()
         }
     }
 
+    // Make AI aware of a special castle which is a winning condition for a human player.
```
oleg-derevenetz, line 702: Please note that this castle may be a winning condition for AI too - usually if this is a neutral castle and "Comp also wins" map setting is in effect. In this case, the AI and the human player(s) should try to capture it in a race. BTW with this setting, if this castle is already belong to an AI player, AI immediately wins.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -35,6 +36,18 @@
 
 namespace
 {
+    bool isCastleLossConditionForHuman( const Castle * castle )
```
oleg-derevenetz, line 39: Please note that this castle will be a loss condition in a single-player only. In a multiplayer, it's just an ordinary castle. In the `GameOver::Result::LocalCheckGameOver()`, single-player mode is detected in the following way:

```cpp
const bool isSinglePlayer = ( Colors( Players::HumanColors() ).size() == 1 );
```

You might want to add the corresponding check here, otherwise in a multi-player mode AI will try to capture this castle like a maniac, despite the fact that it will not give him any advantage.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few notes, please take a look when possible.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -35,6 +36,25 @@
...
+        assert( castle != nullptr );
+
+        if ( Colors( Players::HumanColors() ).size() > 1 ) {
+            // This is a multiplayer mode. Castle loss condition is not applied for such games.
+            return false;
+        }
+
+        const Settings & conf = Settings::Get();
+        // This is either a town which human might lose or AI should capture it to win.
+        if ( ( conf.ConditionLoss() & ( GameOver::LOSS_TOWN | GameOver::LOSS_ENEMY_WINS_TOWN ) ) == 0 ) {
```
oleg-derevenetz, line 50: Hi @ihhub please note that `LOSS_ENEMY_WINS_*` values are used only as return values from the `World::CheckKingdomLoss()` to let the logic inside the `GameOver::Result::LocalCheckGameOver()` know why exactly did this particular kingdom lose. They do not participate in the map win/loss conditions (stored by the `Maps:FileInfo`) by themselves.

ihhub: Yeah, I understand this. That's why I converted to WIP state as many things are in mess...

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -35,6 +36,29 @@
 
 namespace
 {
+    bool isCastleLossConditionForHuman( const Castle * castle )
+    {
+        assert( castle != nullptr );
+
+        const Settings & conf = Settings::Get();
+        if ( conf.CurrentFileInfo().isMultiPlayerMap() ) {
+            // This is a multiplayer mode. Town related conditions are not applied for such game.
```
oleg-derevenetz, line 45: This is not quite true. `WINS_TOWN` condition is applied in the multiplayer mode just fine (all players, possibly including AI players - depending on the "Comp also wins via Special VC" setting, should capture the town in question in a race).

`LOSS_TOWN` condition, however, does not seem to apply. OG in the single-player mode:

https://user-images.githubusercontent.com/32623900/169367172-9eca5fd7-dd79-47f6-8277-9b6c16226b20.mp4

OG in the multiplayer mode on the same map:

https://user-images.githubusercontent.com/32623900/169367691-c9791b55-9181-4db4-8b00-4ce288e0f216.mp4

The castle in question is considered as a regular castle, loss of this castle by the starting player is not considered as the loss of the game. Currently in fheroes2 `LOSS_TOWN` is not applied in multiplayer mode as well.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few notes, please take a look when you have time.

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -698,6 +699,27 @@ void World::ProcessNewMap()
...
+        if ( ( conf.ConditionLoss() & GameOver::LOSS_TOWN ) != 0 ) {
+            // It is a loss town condition for human.
+            Castle * castle = vec_castles.Get( conf.LossMapsPositionObject() );
+            assert( castle != nullptr );
+            if ( castle != nullptr ) {
+                castle->SetModes( Castle::CAPITAL );
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
ihhub: Nope. We don't.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -35,6 +36,26 @@
...
 namespace
 {
+    bool isCastleLossConditionForHuman( const Castle * castle )
+    {
+        assert( castle != nullptr );
+
+        const Settings & conf = Settings::Get();
+        const bool isSinglePlayer = ( Colors( Players::HumanColors() ).size() == 1 );
+
+        if ( isSinglePlayer && ( ( conf.ConditionLoss() & GameOver::LOSS_TOWN ) != 0 ) && ( castle->GetCenter() == conf.LossMapsPositionObject() ) ) {
```
oleg-derevenetz, line 46: It looks like there are too many parentheses here :) Wouldn't it be better this way:

```cpp
if ( isSinglePlayer && ( conf.ConditionLoss() & GameOver::LOSS_TOWN ) != 0 && castle->GetCenter() == conf.LossMapsPositionObject() ) {
```

?

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -35,6 +36,26 @@
...
+            // It is a loss town condition for human.
+            return true;
+        }
+
+        if ( ( conf.WinsCompAlsoWins() && ( conf.ConditionWins() & GameOver::WINS_TOWN ) != 0 ) && ( castle->GetCenter() == conf.WinsMapsPositionObject() ) ) {
```
oleg-derevenetz, line 51:
```cpp
if ( conf.WinsCompAlsoWins() && ( conf.ConditionWins() & GameOver::WINS_TOWN ) != 0 && castle->GetCenter() == conf.WinsMapsPositionObject() ) {
```

?

oleg-derevenetz (commented):
Hi @ihhub LGTM, just one (optional) suggestion regarding the abundance of parentheses :)

src/fheroes2/ai/normal/ai_normal_hero.cpp
ihhub: I've removed them.

PR #5382, opened 2022-05-07
Update all .po against .pot

zenseii:
Also set the correct plural forms for Bulgarian and Turkish in the .po files. These will be fixed in the code in a separate PR.


ihhub:
@zenseii , thank you so much for this correction!

PR #5384, opened 2022-05-07
Tidy up and fix gettext plural forms for languages

zenseii:
This tidies up the Switch statement for gettext plural forms and reverts some changes to the order of cases. In other words I made the various locales use the same plural form as before, but I only checked the locales we currently have translation files for to see what plural expression they use.

Couldn't find a Locale for Hungarian even though a hungarian .po exists with quite a few translated strings.

These are used in cases like these for Russian so that the correct plural form is used:
image

Before this the Russian translation was using the Lithaunian plural form expression, which is not the same for Russian. Not sure if this was intentional.

### DIFF
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -440,7 +440,7 @@ namespace Translation
             switch ( current->locale ) {
             case LocaleType::LOCALE_AF:
             case LocaleType::LOCALE_BE:
-                return current->ngettext( str, (n % 10 == 1 && n % 100 != 11 ? 0 : n % 10 >= 2 && n % 10 <= 4 && ( n % 100 < 12 || n % 100 > 14 ) ? 1 : 2) );
+                return current->ngettext( str, ( n % 10 == 1 && n % 100 != 11 ? 0 : n % 10 >= 2 && n % 10 <= 4 && ( n % 100 < 12 || n % 100 > 14 ) ? 1 : 2 ) );
             case LocaleType::LOCALE_EU:
             case LocaleType::LOCALE_ID:
             case LocaleType::LOCALE_LA:
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -452,6 +452,7 @@ namespace Translation
                 return current->ngettext( str, ( n != 1 ) );
             case LocaleType::LOCALE_DA:
             case LocaleType::LOCALE_DE:
+                return current->ngettext( str, ( n != 1 ) );
             case LocaleType::LOCALE_ES:
                 return current->ngettext( str, ( n != 1 ) );
             case LocaleType::LOCALE_ET:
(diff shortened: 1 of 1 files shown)


src/engine/translations.cpp
```diff
@@ -440,24 +440,27 @@ namespace Translation
...
             case LocaleType::LOCALE_ET:
             case LocaleType::LOCALE_FI:
             case LocaleType::LOCALE_GL:
             case LocaleType::LOCALE_HE:
             case LocaleType::LOCALE_IT:
-                return current->ngettext( str, 0 );
+                return current->ngettext( str, ( n != 1 ) );
             case LocaleType::LOCALE_NL:
             case LocaleType::LOCALE_RO:
                 return current->ngettext( str, ( n == 1 ? 0 : n == 0 || ( n != 1 && n % 100 >= 1 && n % 100 <= 19 ) ? 1 : 2 ) );
```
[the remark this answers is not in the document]
zenseii: Of course...

ihhub:

src/engine/translations.cpp
```diff
@@ -439,30 +439,31 @@ namespace Translation
...
-                return current->ngettext( str, ( n == 1 ? 0 : n == 0 || ( n != 1 && n % 100 >= 1 && n % 100 <= 19 ) ? 1 : 2 ) );
             case LocaleType::LOCALE_SV:
+            case LocaleType::LOCALE_TR:
                 return current->ngettext( str, ( n != 1 ) );
+            case LocaleType::LOCALE_EL:
+            case LocaleType::LOCALE_FR:
+            case LocaleType::LOCALE_PT:
+                return current->ngettext( str, ( n > 1 ) );
+            case LocaleType::LOCALE_AR:
+                return current->ngettext( str, ( n == 0 ? 0 : n == 1 ? 1 : n == 2 ? 2 : n % 100 >= 3 && n % 100 <= 10 ? 3 : n % 100 >= 11 && n % 100 <= 99 ? 4 : 5 ) );
```
oleg-derevenetz, line 464: Hi @zenseii is there a chance that `n % 100` will ever be `> 99`?

zenseii: Hi @oleg-derevenetz, I'm not entirely sure about the formula for modulus but to me it seems like this would never be the case.

These expressions should all be taken from a database of standard expressions for finding the correct plural form for every language, so I haven't come up with any of them myself. I checked the formula for Arabic and it seems the one that was already present in the engine was wrong, although I found it applied automatically by poedit if setting it to use default expressions for Arabic.

https://poedit.net/trac/wiki/Doc/PluralForms
http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms#f1
https://arabeyes.org/Plural_Forms

From these it seems like the correct expression to use for arabic should be:
`n == 0 ? 0 : n == 1 ? 1 : n == 2 ? 2 : n % 100 >= 3 && n % 100 <= 10 ? 3 : n % 100 >= 11 ? 4 : 5`

zenseii: I think we can just use the more correct one above. Not sure why poedit has a default one that is wrong.

This could mean that we should check the other more complex expressions too. We'd only need to check the ones we have up against the database linked to above.

PR #5387, opened 2022-05-07
GitHub Actions: update translation files on push

### LINKED ISSUE #5383: Automate translation files being up to date check
### Preliminary checks

We have a plenty of translations and we constantly update string to be translated: some are being added, others are changed. A translator who intends to update the file for a particular language needs to manually generate fheroes2.pot file and then update the translation file. However, the root of the problem is that a developer who modified the code related to translations does not update translation files.
To generate .pot file you need to execute the following commands:
```bash
cd src/dist
make clean
make pot
```

Then to update translations which are .mo files the below steps must be executed:
```bash
cd ../../files/lang
make clean
make
```

Before proceeding with any automation we need to fix following issues:
- we put `*.pot` to be ignored within **.gitignore** file. This is easy to fix
- fheroes2.pot file has comments related to source code lines. This can be fixed by adding `--no-location` argument for **xgettext**
- newly generated fheroes2.pot file will have updated dates which would differ from the old files so a simple byte comparison will not help
- updated .po files will have the same date issue as pot file.

### Possible automation steps
- add fheroes2.pot file as a part of the repository
- edit .gitignore file
- edit makefile to add `--no-location` argument
- for any commit run a stage which will save an existing fheroes2.pot file (rename it), generates a new fheroes2.pot and compares the content, ignoring date difference
- if the file needs an update fail the stage showing the difference in files
- if no update needed proceed with po files and do the same

### Additional info

_No response_

oleg-derevenetz:
close #5383

Instead of forcing poor Windows developers to install Cygwin and `gettext`, deal with `bash`, `make`, and so on, I suggest just automatically regenerate `fheroes2.pot` and `*.po` on each push event. This will increase number of commits in log, though.

### DIFF
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -38,9 +38,9 @@ jobs:
         git diff -U0 HEAD^ | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
     - name: Save PR metadata
       run: |
-        echo ${{ github.event.number }} > clang-tidy-result/pr-id.txt
-        echo ${{ github.event.pull_request.head.repo.full_name }} > clang-tidy-result/pr-head-repo.txt
-        echo ${{ github.event.pull_request.head.ref }} > clang-tidy-result/pr-head-ref.txt
+        echo "${{ github.event.number }}" > clang-tidy-result/pr-id.txt
+        echo "${{ github.event.pull_request.head.repo.full_name }}" > clang-tidy-result/pr-head-repo.txt
+        echo "${{ github.event.pull_request.head.ref }}" > clang-tidy-result/pr-head-ref.txt
     - uses: actions/upload-artifact@v2
       with:
         name: clang-tidy-result
--- a/.github/workflows/clang_tidy_comments.yml
+++ b/.github/workflows/clang_tidy_comments.yml
@@ -36,13 +36,13 @@ jobs:
       run: |
         mkdir clang-tidy-result
         unzip clang-tidy-result.zip -d clang-tidy-result
-        echo "pr_id=$(cat clang-tidy-result/pr-id.txt)" >> "$GITHUB_ENV"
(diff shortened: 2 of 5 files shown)


ihhub:
Hi @oleg-derevenetz , is it correct that the bot will add a commit after every push to master?

oleg-derevenetz:
> is it correct that the bot will add a commit after every push to master?

Yes - if there is anything to commit. I'm currently thinking about reducing the number of commits by ignoring changes only in dates (`POT-Creation-Date`, `PO-Revision-Date`). BTW, in case of several consecutive pushes in a relatively small time frame there will be just one commit added, because previous jobs will be canceled.

oleg-derevenetz:
Hi @ihhub now the new commit will be created only if regenerated `*.po` files have other changes except changes in `POT-Creation-Date`. This should reduce the number of additional commits.

ihhub:
@oleg-derevenetz , right now it is impossible to directly commit to master branch without creation of a pull request. I made this to avoid any mistakes to push directly to master. I can disable this condition.

oleg-derevenetz:
@ihhub we can try to enable pushes, but disallow force pushes for everyone including administrators. This should protect from accidental data loss good enough. As an alternative, I can try to create pull requests each time after the push instead of creating commits in `master` and pushing them, but this has its own drawbacks - for example, after a few pushes there will be several almost identical PRs for the same po files.

ihhub:
@oleg-derevenetz , I think pull request create might be the better idea than just populate git history with extra commits.

oleg-derevenetz:
Hi @ihhub

> I think pull request create might be the better idea than just populate git history with extra commits.

OK, now this workflow will create a temporary branch, push changes to it and create a pull request to merge this branch (and also add proper labels and assign this PR to you).

ihhub:
@oleg-derevenetz , huge thanks for such improvements! Let's hope it'd speed up translations.

PR #5395, opened 2022-05-09
Don't try to add fheroes2.key to packages: now it is generated automatically

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks a lot for cleaning up!

PR #5396, opened 2022-05-10
Make Artifact evaluation for AI more granular

### LINKED ISSUE #4226: Add individual Artifact value for AI
### Preliminary checks

Right now all artifacts for AI are differentiated by 4 group values. This is not correct as some artifacts much more valuable than others even within the same group.

We differentiate the value of artifacts only based on it's Level in `int Artifact::getArtifactValue() const` method.

For each artifact we have to define its value which could be used by AI to evaluate the quality of certain action.

### Additional info

_No response_

felix642:
closes #4226.

Improved the getArtifactValue() function based on what @LeHerosInconnu said in #4226.

Artifacts value is based on multiple factors :
+1 per skill level
+1 per spell tier

The other effects add +1 to the value based on the weakest artifact.

For example, Gold Income gives +1 per 800 gold/day


src/fheroes2/resource/artifact.cpp
```diff
@@ -235,24 +235,76 @@ int Artifact::Level( void ) const
...
+        case fheroes2::ArtifactBonusType::HYPNOTIZE_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::DEATH_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::BERSERK_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::BLIND_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::PARALYZE_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::HOLY_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::DISPEL_SPELL_IMMUNITY:
+        case fheroes2::ArtifactBonusType::ENDLESS_AMMUNITION:
+        case fheroes2::ArtifactBonusType::NO_SHOOTING_PENALTY:
+        case fheroes2::ArtifactBonusType::VIEW_MONSTER_INFORMATION:
```
felix642, line 281: Not a fan of this case, but I feel like the issue is related to the enum and not my function. 

I feel like *_IMMUNITY should be similar to ADD_SPELL and only have 1 enum (SPELL_IMMUNITY) with the corresponding spell.
The same logic should apply for *_EXTRA_EFFECTIVENESS_PERCENT AND *_DAMAGE_REDUCTION

src/fheroes2/resource/artifact.cpp
```diff
@@ -235,24 +236,76 @@ int Artifact::Level( void ) const
...
-    else if ( level & ART_LEVEL3 ) {
-        return 3;
-    }
-    else if ( level & ART_ULTIMATE ) {
-        return 5;
-    }
+    int artifactValue = 0;
+    const std::vector<fheroes2::ArtifactBonus> & bonuses = fheroes2::getArtifactData( id ).bonuses;
+    const std::vector<fheroes2::ArtifactCurse> & curses = fheroes2::getArtifactData( id ).curses;
+    artifactValue += std::accumulate( bonuses.begin(), bonuses.end(), 0, []( int sum, const fheroes2::ArtifactBonus & bonus ) {
```
ihhub, line 242: Shouldn't a normal switch-case code be sufficient? For example,
```cpp
int artifactValue = 0;
for ( const fheroes2::ArtifactBonus & bonus : bonuses ) {
    switch ( bonus ) {
    case fheroes2::ArtifactBonusType::ADD_SPELL:
        artifactValue += Spell( bonus.value ).Level();
        break;
    case fheroes2::ArtifactBonusType::GOLD_INCOME:
        artifactValue -= bonus.value / 1000;
        break;
    ...
    default:
        // Did you add a new artifact bonus type? Add the logic above!
        assert( 0 );
        break;
    }
}
```
This code will be also faster as no lambda-function would be called. Also we must handle unknown bonuses and not to add any value for them. What do you think?

ihhub (changes requested):
Hi @felix642 , I left 2 comments in this pull request. Would you mind to check them in your free time?

src/fheroes2/resource/artifact.cpp
```diff
@@ -235,24 +236,76 @@ int Artifact::Level( void ) const
 
 int Artifact::getArtifactValue() const
 {
-    // TODO: this method should return a value of the artifact based on its bonuses and curses.
```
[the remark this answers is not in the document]
felix642: Here's the list of artifacts with their corresponding values.

| Artifact's id | Value |
|--|--|
| 0 | 12 |
| 1 | 12 |
| 2 | 12 |
| 3 | 12 |
| 4 | 12 |
| 5 | 12 |
| 6 | 16 |
| 7 | 10 |
| 8 | 4 |
| 9 | 2 |
| 10 | 2 |
| 11 | 3 |
| 12 | 1 |
| 13 | 1 |
| 14 | 1 |
| 15 | 1 |
| 16 | -2 |
| 17 | 1 |
| 18 | 1 |
| 19 | 1 |
| 20 | 1 |
| 21 | 1 |
| 22 | 2 |
| 23 | 3 |
| 24 | 2 |
| 25 | 3 |
| 26 | 2 |
| 27 | 3 |
| 28 | 4 |
| 29 | 5 |
| 30 | 1 |
| 31 | 0 |
| 32 | 0 |
| 33 | 2 |
| 34 | 1 |
| 35 | 1 |
| 36 | 1 |
| 37 | 1 |
| 38 | 1 |
| 39 | 2 |
| 40 | 2 |
| 41 | 1 |
| 42 | 1 |
| 43 | 2 |
| 44 | 1 |
| 45 | 1 |
| 46 | 1 |
| 47 | 1 |
| 48 | 1 |
| 49 | 1 |
| 50 | 1 |
| 51 | 1 |
| 52 | 2 |
| 53 | 2 |
| 54 | 1 |
| 55 | 1 |
| 56 | 1 |
| 57 | 1 |
| 58 | 1 |
| 59 | 1 |
| 60 | 1 |
| 61 | 1 |
| 62 | 1 |
| 63 | 1 |
| 64 | 1 |
| 65 | 1 |
| 66 | 5 |
| 67 | 2 |
| 68 | 1 |
| 69 | -1 |
| 70 | -1 |
| 71 | 1 |
| 72 | 1 |
| 73 | 1 |
| 74 | 1 |
| 75 | 1 |
| 76 | 1 |
| 77 | 2 |
| 78 | 4 |
| 79 | 2 |
| 80 | 4 |
| 81 | 0 |
| 82 | 0 |
| 83 | 0 |
| 84 | 0 |
| 85 | 0 |
| 86 | 0 |
| 87 | 2 |
| 88 | 5 |
| 89 | -1 |
| 90 | 26 |
| 91 | 1 |
| 92 | 0 |
| 93 | 0 |
| 94 | 5 |
| 95 | 5 |
| 96 | 8 |
| 97 | 2 |
| 98 | 3 |
| 99 | 5 |
| 100 | 5 |
| 101 | 5 |
| 102 | 1 |
| 103 | 0 |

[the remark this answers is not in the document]
ihhub: The original range of values was [0;5] while we have [-2;26]. This will definitely change AI behavior in general. Also `ENDLESS_BAG_GOLD` and `ENDLESS_PURSE_GOLD` have values of 0 which shouldn't be the case. We need to switch from **int** to **double** as a first step so that some artifacts will have more than 0 values. Then update `BagArtifacts::getArtifactValue()` method to return **double** as well.

src/fheroes2/resource/artifact.cpp
```diff
@@ -235,24 +236,91 @@ int Artifact::Level( void ) const
...
+        case fheroes2::ArtifactBonusType::VIEW_MONSTER_INFORMATION:
+            artifactValue += 1;
+            break;
+        case fheroes2::ArtifactBonusType::MAXIMUM_MORALE:
+        case fheroes2::ArtifactBonusType::DISABLE_ALL_SPELL_COMBAT_CASTING:
+        case fheroes2::ArtifactBonusType::MAXIMUM_LUCK:
+            artifactValue += 3;
+            break;
+        default:
+            artifactValue += bonus.value;
```
ihhub, line 299: Here we shouldn't do anything but to add an assertion to handle unknown bonuses.

src/fheroes2/resource/artifact.cpp
```diff
@@ -235,24 +236,91 @@ int Artifact::Level( void ) const
...
+        case fheroes2::ArtifactCurseType::COLD_SPELL_EXTRA_DAMAGE_PERCENT:
+        case fheroes2::ArtifactCurseType::FIRE_SPELL_EXTRA_DAMAGE_PERCENT:
+            artifactValue -= curse.value / 100;
+            break;
+        case fheroes2::ArtifactCurseType::NO_JOINING_ARMIES:
+        case fheroes2::ArtifactCurseType::UNDEAD_MORALE_PENALTY:
+            artifactValue -= 1;
+            break;
+        default:
+            artifactValue -= curse.value;
```
ihhub, line 318: Here we shouldn't do anything but to add an assertion to handle unknown curses.

src/fheroes2/resource/artifact.cpp
```diff
@@ -235,24 +236,91 @@ int Artifact::Level( void ) const
...
-    }
-    else if ( level & ART_LEVEL3 ) {
-        return 3;
-    }
-    else if ( level & ART_ULTIMATE ) {
-        return 5;
-    }
+    int artifactValue = 0;
+    const std::vector<fheroes2::ArtifactBonus> & bonuses = fheroes2::getArtifactData( id ).bonuses;
+    const std::vector<fheroes2::ArtifactCurse> & curses = fheroes2::getArtifactData( id ).curses;
```
ihhub, line 241: It is better to cache a return value of `getArtifactData` function to make the code faster.

```cpp
const fheroes2::ArtifactData & data = fheroes2::getArtifactData( id );
const std::vector<fheroes2::ArtifactBonus> & bonuses = data.bonuses;
const std::vector<fheroes2::ArtifactCurse> & curses = data.curses;
```

ihhub (changes requested):
Hi @felix642 , I left few comments here. Would you mind to take a look?

felix642:
> The original range of values was [0;5] while we have [-2;26]. This will definitely change AI behavior in general. Also ENDLESS_BAG_GOLD and ENDLESS_PURSE_GOLD have values of 0 which shouldn't be the case. We need to switch from int to double as a first step so that some artifacts will have more than 0 values. Then update BagArtifacts::getArtifactValue() method to return double as well.

@ihhub, do you think  we should normalize the values between something like [-5, 5] ?

ihhub:
> > The original range of values was [0;5] while we have [-2;26]. This will definitely change AI behavior in general. Also ENDLESS_BAG_GOLD and ENDLESS_PURSE_GOLD have values of 0 which shouldn't be the case. We need to switch from int to double as a first step so that some artifacts will have more than 0 values. Then update BagArtifacts::getArtifactValue() method to return double as well.
> 
> @ihhub, do you think we should normalize the values between something like [-5, 5] ?

No we shouldn't. The negative effect of artifacts is not that big as a positive. Let's do step by step. We will reach the ideal state eventually :)

src/fheroes2/resource/artifact.cpp
```diff
@@ -233,26 +234,96 @@ int Artifact::Level( void ) const
...
-    }
-    else if ( level & ART_LEVEL2 ) {
-        return 2;
-    }
-    else if ( level & ART_LEVEL3 ) {
-        return 3;
+    for ( const fheroes2::ArtifactBonus & bonus : bonuses ) {
+        switch ( bonus.type ) {
+        case fheroes2::ArtifactBonusType::ADD_SPELL:
+            artifactValue += Spell( bonus.value ).Level();
```
ihhub, line 247: Spell Scroll has value of 0 so this will not work. I suggest to just assign 1 to the value as a rough estimation. Also it would be more accurate because while picking scrolls on map AI shouldn't know their real value.

src/fheroes2/resource/artifact.cpp
```diff
@@ -233,26 +234,96 @@ int Artifact::Level( void ) const
...
+            break;
+        case fheroes2::ArtifactBonusType::GOLD_INCOME:
+            artifactValue += static_cast<double>( bonus.value ) / 1000.0;
```
ihhub, line 250: I would say that division by 800 would be more accurate to bring Golden Goose to the same value as the rest of Ultimate Artifacts. Also Gold is a very valuable resource in the game. For 250 gold you can hire a decent monster which is better than +1 to a primary skill.

src/fheroes2/resource/artifact.cpp
```diff
@@ -233,26 +234,96 @@ int Artifact::Level( void ) const
...
+            break;
+        case fheroes2::ArtifactBonusType::SEA_MOBILITY:
+            artifactValue += static_cast<double>( bonus.value ) / 500.0;
+            break;
+        case fheroes2::ArtifactBonusType::LAND_MOBILITY:
+            artifactValue += static_cast<double>( bonus.value ) / 300.0;
```
ihhub, line 256: 200 would be even better. Land mobility is much more valuable than sea mobility. Nomad Boots is a very powerful artifact.

src/fheroes2/resource/artifact.cpp
```diff
@@ -233,26 +234,96 @@ int Artifact::Level( void ) const
...
+            break;
+        }
     }
-    else if ( level & ART_ULTIMATE ) {
-        return 5;
+
+    for ( const fheroes2::ArtifactCurse & curse : curses ) {
+        switch ( curse.type ) {
+        case fheroes2::ArtifactCurseType::GOLD_PENALTY:
+            artifactValue -= static_cast<double>( curse.value ) / 250.0;
```
ihhub, line 309: 200 as for the same ability.

ihhub (changes requested):
Hi @felix642 , I left few suggestions. Could you please update the table of artifact values after this so other members evaluate these numbers and we can adjust them properly?

src/fheroes2/resource/artifact.cpp
idshibanov: Also take into account that Endless Sack of Gold (+1000 gold) is a major tier 3 artifact (in map editor) and should be comparable to at least other +3's.

idshibanov:
> > > The original range of values was [0;5] while we have [-2;26]. This will definitely change AI behavior in general. Also ENDLESS_BAG_GOLD and ENDLESS_PURSE_GOLD have values of 0 which shouldn't be the case. We need to switch from int to double as a first step so that some artifacts will have more than 0 values. Then update BagArtifacts::getArtifactValue() method to return double as well.
> > 
> > 
> > @ihhub, do you think we should normalize the values between something like [-5, 5] ?
> 
> No we shouldn't. The negative effect of artifacts is not that big as a positive. Let's do step by step. We will reach the ideal state eventually :)

Don't worry about normalizing ultimate artifacts since they are never placed on the map. I think AI should be ok even without additional changes.

src/fheroes2/resource/artifact.cpp
```diff
@@ -233,26 +234,94 @@ int Artifact::Level() const
...
 
-    if ( level & ART_LEVEL1 ) {
-        return 1;
-    }
-    else if ( level & ART_LEVEL2 ) {
-        return 2;
-    }
-    else if ( level & ART_LEVEL3 ) {
-        return 3;
+    for ( const fheroes2::ArtifactBonus & bonus : bonuses ) {
```
ihhub, line 244: Some artifact bonuses are not covered here like `KNOWLEDGE_SKILL` or `ATTACK_SKILL`. Please add them and run
```cpp
for ( int i = 0; i < Artifact::UNKNOWN; ++i ) {
    ERROR_LOG( Artifact( i ).GetName() << " - " << Artifact( i ).getArtifactValue() )
}
```
somewhere in the code to verify the results.

ihhub (changes requested):

felix642:
Hi @ihhub, sorry for the delay I forgot I had an open PR..

I added the required logic for the missing artifacts.

ihhub:
Hi @felix642 , could you please update your branch from master branch as we made a lot of changes and on Visual Studio it's impossible to test without rerunning all scripts?

ihhub (approved):

ihhub:
@felix642 , thank you so much for this improvement!

PR #5400, opened 2022-05-10
Update Polish translation

michkry:


files/lang/pl.po
```diff
@@ -2739,29 +2733,23 @@ msgstr "Nieznany"
...
+msgstr "Moc zaklęć:"
 
-#, fuzzy
 msgid "Knowledge:"
-msgstr "Wiedza"
+msgstr "Wiedza:"
 
-#, fuzzy
 msgid "Spell Points:"
-msgstr "Ptk. magii"
```
zenseii, line 2760: You should be able to write these out without abbreviations after the recent PR that made more space available. You should have space for around 15 characters (i.e. `Reconnaissance :` ).

files/lang/pl.po
```diff
@@ -3923,26 +3911,25 @@ msgid "Norwegian"
...
 msgid ", FPS: "
 msgstr ", FPS: "
 
 msgid "%{object} robber"
-msgstr ""
+msgstr "Rabuś %{object}"
 
 msgid "%{object} raided"
-msgstr ""
+msgstr "%{object} splądrowany"
```
Arthusppp, line 3932: I suggest "%{object} splądrowana" as this refers only to: 

> case MP2::OBJ_PYRAMID:
> case MP2::OBJN_PYRAMID: {
>     std::string modRaided = _( "%{object} raided" );



image

files/lang/pl.po
```diff
@@ -8552,6 +8542,12 @@ msgid ""
 "Mercury, Sulfur and Crystal are rare magical resources used for the most "
 "powerful creatures and buildings."
 msgstr ""
+"W Heroes 2 jest siedem surowców, używanych do budowy i ulepszania miast, "
+"kupowania jednostek i rekrutowania bohaterów. Złoto jest najbardziej "
+"pospolitym surowcem i jest potrzebne praktycznie do wszystkiego. Drewno i "
+"ruda są używane do budowy większości budynków. Klejnoty, rtęć, siarka i "
+"kryształy są rzadkimi magicznymi surowcami uzywanymi do budowy potężnych "
+"istot i budynków."
```
Arthusppp, line 8550: What about: "do budowy potężnych budynków i rekrutowania potężnych istot"? 
It's because we are not building Dragnos nor Titans - we are recruiting them ;)

files/lang/pl.po
```diff
@@ -9122,48 +9117,47 @@ msgid "world: allow to set guardian to objects"
...
 msgid "battle: deterministic events"
 msgstr "walka: deterministyczne zdarzenia"
 
 msgid "game: show system info"
-msgstr "game: pokaż informacje systemowe"
+msgstr "gra: pokaż informacje o systemie"
 
 msgid "game: autosave will be made at the beginning of the day"
-msgstr "Gra: Autozapis nastąpi na początku każdego dnia"
+msgstr "gra: Autozapis nastąpi na początku każdego dnia"
```
Arthusppp, line 9150: "autozapis"? Why capital letter?

files/lang/pl.po
zenseii: I see! As long as you don't feel like their shortening makes it hard for the player to understand what they mean then it is fine. 

We could always make the distance of the column of numbers further to the right. I was thinking we might just want to have a distance that adjusts itself according to the length of the text in the left column, but this would most likely require a bit of a rework of the logic here.

files/lang/pl.po
michkry:
```suggestion
msgstr "Splądrowana %{object}"
```

files/lang/pl.po
```diff
@@ -2739,29 +2733,23 @@ msgstr "Nieznany"
...
 
-#, fuzzy
 msgid "Knowledge:"
-msgstr "Wiedza"
+msgstr "Wiedza:"
 
-#, fuzzy
 msgid "Spell Points:"
-msgstr "Ptk. magii"
+msgstr "Ptk. magi:"
```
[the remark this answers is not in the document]
michkry: Corrected

files/lang/pl.po
michkry: I think we can leave the shorten version, at least for the moment. And in the future, if the logic is reworked, we will change the format of those items.

files/lang/pl.po
michkry: Changed to "do budowy potężnych budynków i rekrutowania potężnych istot"

files/lang/pl.po
michkry: Corrected

Arthusppp (approved):

michkry:
I resolved all discussions. If there are no further discussions, this PR is ready to be merged into the master branch

files/lang/pl.po
```diff
@@ -1089,13 +1089,12 @@ msgstr "Wyspy Piratów"
...
 msgstr "Rozbitek"
 
-#, fuzzy
 msgid ""
 "Capture the town on the island off the southeast shore in order to construct "
 "a boat and travel back towards the mainland. Do not lose the hero Gallavant."
 msgstr ""
 "Przejmij miasto na wyspie południowo-wschodniego wybrzeża aby zbudować łódź "
-"i wrócić z powrotem na ląd."
+"i wrócić z powrotem na ląd. Nie możesz stracić Gallavanta."
```
zenseii, line 1097: Hi @michkry! This is the only thing I noticed when looking over your PR. Would it be better to mention that Gallavant is a hero like in the original text here? Not sure it could be confused with something else like a town or similar, but it might at least make it clearer for the player.

michkry: I changed the whole translation. I took the text from the official polish version of the game.

zenseii (approved):

ihhub:
@michkry , thank you so much for updating Polish translation!

PR #5403, opened 2022-05-11
Fix character filtering when generating the save file name

### LINKED ISSUE #5390: Automatically generated save game naming fails, when map name is not properly recognized
### Preliminary checks

Some user generated maps have unrecognizable names, i.e. question marks, instead of the original language text (I assume, cyricil or other). When saving progress playing these maps, the question marks are not used in the automatically generated save filename, and instead has [...]_[...]_001 format, where [...] is empty space. Attempting to save under such filename results in an error "There was an issue during saving".

Generate a savename for such maps without unallowed characters?

### Additional info

_No response_

oleg-derevenetz:
fix #5390


ihhub:

PR #5413, opened 2022-05-14
Add Portuguese language option and initial font generation

zenseii:
This replaces all special characters with their most similar latin counterpart.

These letters were added: `Ã, Â, Ç, Ê, Õ, Ô, ã, â, ç, ê, õ, ô`.


ihhub:
@zenseii , thank you so much for adding this support!

altiereslima:
missing í and Í

zenseii:
> missing í and Í

Did you test it? These were added when Spanish were added, that's why they are not specifically mentioned here.

PR #5415, opened 2022-05-15
fheroes2::getLEValue(): replace static_assert() by #error

oleg-derevenetz:
While all modern compilers I aware of fire this `static_assert()` immediately when compiling the template (what is in fact is needed here if there is no valid byte order info present), formally (according to the standard) the template for which no valid specialization can be generated is "ill-formed, no diagnostic required" (and, formally, no valid specialization can be generated for the template with `static_assert( false, ... )`). So let's just replace it with the `#error` preprocessor directive - just in case.


ihhub:
@oleg-derevenetz , thanks a lot for extra tiny improvement!

Issue #4637, opened 2021-11-14
Adventure map, Multiplayer mode, instant combat should not be activated when a hero attacks another hero


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, instant combat should not be activated when a hero attacks another hero in multiplayer mode.

In fheroes2:

https://user-images.githubusercontent.com/43583994/141685719-dac48d29-418d-4b1f-a544-fe12805660c0.mp4

Here is the test scenario file:
[0 9951.zip](https://github.com/ihhub/fheroes2/files/7533852/0.9951.zip)

Also, if the green player attacks first, the message of the instant battle is intended for the blue player (the winner).

### Save file

Here is the save file (multiplayer):
[Before attack multiplayer 01.zip](https://github.com/ihhub/fheroes2/files/7533949/Before.attack.multiplayer.01.zip)

a1exsh:
> In fheroes2, instant combat should not be activated when a hero attacks another hero in multiplayer mode.

Why not? Do you think the defeated player will always insist on replaying the battle?

LeHerosInconnu:
Hello @a1exsh,

> > In fheroes2, instant combat should not be activated when a hero attacks another hero in multiplayer mode.
> 
> Why not? Do you think the defeated player will always insist on replaying the battle?

We can see that you are not familiar with the behavior of some players in multiplayer. :D

In the presented case, the player controlling the attacked hero might have the Blind spell or other spells, and the combat might have another outcome.
The player controlling the attacked hero may also want to cast a damage spell before retreating his hero.
And so on.
What if you were the one controlling the attacked hero? :D

oleg-derevenetz:
Let's treat this as an suggestion for improvement so far, because OG doesn't have the instant combat at all. I think that (at least in hotseat mode) this is acceptable, because the losing side may agree to the outcome of the auto combat, if this hero is expendable for him.

a1exsh:
@LeHerosInconnu 
> We can see that you are not familiar with the behavior of some players in multiplayer. :D

That's exactly my point: why should we change the game behavior for all players just because some like it differently?.. :)

[...]

PR #5417, opened 2022-05-15
Impose a limit of turns for an attacking AI-controlled hero

### LINKED ISSUE #5342: Two trolls cannot kill each other making instant battle last infinitely
### Preliminary checks

### Platform

Windows

### Describe the bug

Trolls have HP regeneration, which makes battle 1vs1 an endless fight to "no death".
image

### Save file

[Infnite Troll fight.zip](https://github.com/ihhub/fheroes2/files/8582945/Infnite.Troll.fight.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5342

Currently the logic is as follows:

* The limit is triggered if the number of dead in both armies does not change within 50 turns;
* The limit is applied only to the **_attacking AI-controlled army_** (the concept of "AI-controlled" in this particular case includes instant or auto battle mode). It is never applied to the defending army (it doesn't matter whether is controlled by AI or not), and also it is not applied to the attacking army if it is **_directly_** controlled by a human player at the moment (neither instant battle or auto battle are currently in effect);
* If the attacking army is controlled by AI player, and the limit of turns is exceeded, then the attacking hero will retreat (and lose his army);
* If the attacking army is controlled by human player, and there is an **_instant battle_** in progress, then the attacking hero will retreat as well;
* If the attacking army is controlled by human player, and there is an **_auto battle_** in progress, the auto battle will be turned off when the limit of turns is exceeded, and the attacker will not be able to turn it on again (it will be turned back off immediately).

That's how it currently works in case of an attacking AI-controlled hero:

https://user-images.githubusercontent.com/32623900/169033796-ff186097-9b0f-46d8-a705-db2f5f401a3f.mp4

### DIFF
--- a/src/fheroes2/ai/ai.h
+++ b/src/fheroes2/ai/ai.h
@@ -69,16 +69,12 @@ namespace AI
     class Base
     {
     public:
-        virtual void KingdomTurn( Kingdom & kingdom );
-        virtual void CastleTurn( Castle & castle, bool defensive );
-        virtual void BattleTurn( Battle::Arena & arena, const Battle::Unit & unit, Battle::Actions & actions );
-        virtual void HeroTurn( Heroes & hero );
-        virtual bool HeroesTurn( VecHeroes & )
-        {
-            return true;
-        }
+        virtual void KingdomTurn( Kingdom & kingdom ) = 0;
+        virtual void CastleTurn( Castle & castle, bool defensive ) = 0;
+        virtual void BattleTurn( Battle::Arena & arena, const Battle::Unit & unit, Battle::Actions & actions ) = 0;
+        virtual bool HeroesTurn( VecHeroes & heroes ) = 0;
 
-        virtual void revealFog( const Maps::Tiles & tile );
+        virtual void revealFog( const Maps::Tiles & tile ) = 0;
 
         virtual void HeroesAdd( const Heroes & hero );
         virtual void HeroesRemove( const Heroes & hero );
@@ -98,7 +94,6 @@ namespace AI
         virtual void CastlePreBattle( Castle & castle );
         virtual void CastleAfterBattle( Castle & castle, bool attackerWins );
 
-        virtual const char * Type() const;
         virtual int GetPersonality() const; // To be utilized in future.
         virtual std::string GetPersonalityString() const;
 
--- a/src/fheroes2/ai/ai.h
+++ b/src/fheroes2/ai/ai.h
@@ -105,6 +105,10 @@ namespace AI
         virtual void Reset();
         virtual void resetPathfinder() = 0;
 
+        // Should be called at the beginning of the battle even if no AI-controlled players are
+        // involved in the battle - because of the possibility of using instant or auto battle
+        virtual void battleBegins() = 0;
+
         virtual ~Base() = default;
 
     protected:
--- a/src/fheroes2/ai/ai_base.cpp
+++ b/src/fheroes2/ai/ai_base.cpp
@@ -18,26 +18,13 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-#include "agg.h"
 #include "ai.h"
-#include "battle_arena.h"
-#include "battle_command.h"
-#include "battle_troop.h"
-#include "game_interface.h"
 #include "heroes.h"
-#include "kingdom.h"
-#include "logging.h"
-#include "mus.h"
 #include "serialize.h"
 #include "translations.h"
 
 namespace AI
 {
-    const char * Base::Type() const
-    {
-        return "base";
-    }
-
     int Base::GetPersonality() const
     {
         return _personality;
@@ -56,7 +43,7 @@ namespace AI
             break;
         }
 
-        return Type();
+        return _( "None" );
     }
 
     void Base::Reset()
@@ -70,11 +57,6 @@ namespace AI
     }
 
     void Base::CastleAfterBattle( Castle &, bool )
-    {
-        // Do nothing.
-    }
-
-    void Base::CastleTurn( Castle &, bool )
     {
         // Do nothing.
     }
@@ -119,11 +101,6 @@ namespace AI
         // Do nothing.
     }
 
-    void Base::revealFog( const Maps::Tiles & )
-    {
-        // Do nothing.
-    }
-
     std::string Base::HeroesString( const Heroes & )
     {
         return std::string();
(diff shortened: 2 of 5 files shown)


src/fheroes2/ai/ai.h
```diff
@@ -61,6 +61,10 @@ namespace AI
         EXPLORER
     };
 
+    // When this limit of turns is exceeded for an attacking AI-controlled hero,
+    // the auto battle should be interrupted (one way or another)
+    const uint32_t BATTLE_TURN_LIMIT = 49;
```
ihhub, line 66: The problem with just counting turns till the limit is that some very rare battles can be that long. We need to separate battles with turns that make no changes in army's HP / strength and those which do. For example, attacking by Grand Elves with Ammo Cart an army of Hydras. Assume that we have 1 Elf and 100 Hydras. So the Elf technically can defeat Hydras by moving and shooting. It might take a lot of turns but it is possible.

However, the cases described in the original issue make no change in army's health leading to no changes within turns and this is where we need to impose such limit.

oleg-derevenetz: Hi @ihhub Technically, relying on HP change (or the lack of such change) is also unreliable, no HP changes during the several turns doesn't prove that situation is hopeless. Suppose that our AI has a concept of "soft wait" and it should defeat 100 Dwarves with the single Sprite. All it should do is wait for the morale event to attack the Dwarves without retaliation and then move away from them at the beginning of the next turn, because of their higher speed. Sprite will 100% win with this tactic, but how long it will take? Morale event may arrive every 5 turns or once in 50 turns - it's random after all. My proposal is to impose the hard limit on the AI/auto battle - it may be large enough, say, 100 turns or may be more - as the kill switch.

oleg-derevenetz: @ihhub I'll think about it a bit more.

LeHerosInconnu:
Hello @oleg-derevenetz and @ihhub,

> * If the attacking army is controlled by AI player, and the limit of turns is exceeded, then the attacking hero will retreat (and lose his army);
> * If the attacking army is controlled by human player, and there is an **_instant battle_** in progress, then the attacking hero will retreat as well;

What is this based on in the first place (appart from HoTA rules)?
It's a bit arbitrary and not in the spirit of Heroes 2.
Let's take an example with two identical heroes, without Morale and Luck modifiers and with 1 Black Dragon each.

In all cases, the hero who attacks first wins the combat, whether the damage of each attack is 1 point or 1000 points.
This is the spirit of Heroes 2, the attacker always has the advantage.

> * If the attacking army is controlled by human player, and there is an **_auto battle_** in progress, the auto battle will be turned off when the limit of turns is exceeded, and the attacker will not be able to turn it on again (it will be turned back off immediately).

I'm not sure this is a good thing, plus it should be made very clear to the player.
The auto battle in this case should end the combat automatically with the same rules as the instant battle.

oleg-derevenetz:
Hi @LeHerosInconnu

> What is this based on in the first place

This is based on the fact that attacker is an active participant, so he should take all responsibility for his actions. Also this is based on the fact that attacker is always able to retreat, while defender is not always able to retreat.

> The auto battle in this case should end the combat automatically with the same rules as the instant battle.

So the attacker should also retreat even with auto battle turned on? I believe that shouldn't be the case, because, unlike the instant battle, this action is non-reversible. I believe that auto battle should stop and let the player decide what to do.

LeHerosInconnu:
@oleg-derevenetz.

> Hi @LeHerosInconnu
> 
> > What is this based on in the first place
> 
> This is based on the fact that attacker is initiator of the battle, so he should take all responsibility for his actions. If even with all his advantages he is not able to win, then he miscalculated something and he should retreat.

So this means that a powerful hero with a big army should surrender because he can't engage a sprite.
This is really not in the spirit of Heroes 2.

What's so complicated about making the strongest army win?
Everything should happen as if the fight should be forced at that moment, as if the battlefield had a dimension of 5 hexagons by 5 hexagons.

> Also this is based on the fact that attacker is always able to retreat, while defender is not always able to retreat.

I don't see a problem with that.
Whatever the combat mode, manual (one or two human players), instant combat (should only work when a human player is the attacker, and with an option (or a change) for multiplayer mode: https://github.com/ihhub/fheroes2/issues/4637), auto combat (this mode can be interrupted at any time, with one or two human players), it will always be the same.
Also, making sure that the most powerful army wins the combat, and that in case of a tie, the attacking hero wins the combat, solves the problem.

> > The auto battle in this case should end the combat automatically with the same rules as the instant battle.
> 
> So the attacker should also retreat even with auto battle turned on?

No, the strongest army wins, and in case of a tie, the attacking hero wins the combat. :)

LeHerosInconnu:
@oleg-derevenetz.

> > What's so complicated about making the strongest army win?
> 
> If this "strongest army" is not able to engage the sprite, then this army cannot win.

The sprite can't win either, so the strongest army should win.
It is precisely to solve this kind of case that it has to happen this way. :)

> Just like @ihhub mentioned there, a single Elf with Ammo Cart is able to eliminate an unlimited number of Hydras in one stack, despite the fact that so-called "value" of these Hydras undoubtedly exceed the "value" of that Elf.

The fight is not blocked in this case, the elf continues to make damage to the hydras.

> BTW, a single Sprite is able to eliminate those Hydras too using soft wait and positive morale. So no, I'm strongly against this "strongest army wins" approach. If it would win, then it should win on the battlefield. Bigger "value" is the guarantee of nothing.

The same for this case, the fight is not blocked, since the sprite continues to make damages to the hydras.

oleg-derevenetz:
@LeHerosInconnu 

> The sprite can't win either, so the strongest army should win.

There is no such thing as "strongest army". Attacker has the initiative and all the profits, if he cannot win with all these benefits, then he failed and should retreat. That's how things work in HotA, I propose to take it for a sample.

> The same for this case, the fight is not blocked, since the sprite continues to make damages to the hydras.
What I'm trying to say is that even if there will be no morale for 50 or 100 consecutive turns, it doesn't mean that Hydras automatically win.

oleg-derevenetz:
@LeHerosInconnu

> HoTA is not the Holy Grail. :)

HoTA is the well-known flavor with strict balance rules due to the emphasis on the multiplayer. I see no reason not to borrow some of the things they have thought out and polished. 

> Yes, but the fight goes on here, it is not endless; and the duration of the combat should not be based on a number of rounds.

How do we know that the fight is still goes on here if there is no damage for 50-100 rounds? That's the problem. We should stop somewhere. I still believe that "army value" per se has nothing to do with battle results. Attacker should take the responsibility because he has the initiative and he was able to avoid the fight, while the defender wasn't able to do this.

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu
> 
> > HoTA is not the Holy Grail. :)
> 
> HoTA is the well-known flavor with strict balance rules due to the emphasis on the multiplayer. I see no reason not to borrow some of the things they have thought out and polished.

Heroes 2 is not a strictly balanced game.
You have to take the whole game in consideration, the "balance" also takes into account the size of the map.
Some types of heroes are better on small maps, others on medium maps, others are good at the beginning and at the end of the game and are average in the middle of the game.
This is the charm of Heroes 2, there are always different ways to play.
Also, Heroes 2 was created with single player game in mind, not multiplayer.

Let's base it on what Heroes 2 does in this case.
Let's have two identical heroes fight without Morale and Luck modifiers, with each creature in the game as an army against the same creature (1 peasant against 1 peasant, etc.).
In most cases (i.e. when the fight is not endless), the creature that attacks first wins the fight for the hero who commands it.
Let's apply the rules according to this which is more related to Heroes 2 than any other game or add-on of the series.

> > Yes, but the fight goes on here, it is not endless; and the duration of the combat should not be based on a number of rounds.
> 
> How do we know that the fight is still goes on here if there is no damage for 50-100 rounds? That's the problem. We should stop somewhere.

A rule can be put in place in this case.
Let's say that if the creatures' hit points don't change for 5 turns (or a bit more), the fight is considered endless and ends (with the rules of an endless combat applied).

> I still believe that "army value" per se has nothing to do with battle results. Attacker should take the responsibility because he has the initiative and he was able to avoid the fight, while the defender wasn't able to do this.

I think we have a endless fight here. :rofl: 
What is your army?
Mine is composed of 1 peasant. He is a nice guy. :)
Yours is certainly much better.
Wait a minute! It doesn't really matter.
Since you have engaged the fight, you must retreat.
So, I won with my peasant. I will offer him some gold coins. :1st_place_medal:

LeHerosInconnu:
@zenseii.

> > Also, Heroes 2 was created with single player game in mind, not multiplayer.
> 
> I'm not sure about this statement, @LeHerosInconnu ? Wasn't hotseat present in the game from the beginning? I seem to remember reading stories that the devs tested the game a lot by playing matches against each other and in that way they also learned what they should keep in mind when creating the AI?

Yes, hotseat was present when the game was released.
I read on the internet, in an interview with the creator of the game, Jon Van Caneghem, (I don't have the reference of the source at hand) that the game was created primarily to make the single player game fun. :)

> (Also, I totally didn't write my reply as an edit to your comment at first... Sorry about that.)

No worries. :)

LeHerosInconnu:
@zenseii.

> > Yes, hotseat was present when the game was released. I read on the internet, in an interview with the creator of the game, Jon Van Caneghem (I don't have the reference of the source at hand), that the game was created primarily to make the single player game fun. :)
> 
> I see. Then I stand corrected. Multiplayer or not, I'm not sure it changes much, should the AI become as good as or better than a human player at some point.

I think it's easier to reduce the effectiveness of a too strong AI, compared to playing with a too weak AI.
There is more flexibility in this case. :)

> > No worries. :)
> 
> Great! I'll make sure it doesn't happen again... Anyways, it might be better to continue this discussion in the relevant issue thread to avoid the problem of a PR becoming too crowded and long? It's just a suggestion.

Okay.

src/fheroes2/ai/ai.h
oleg-derevenetz: Hi @ihhub I addressed your concerns by tracking the number of killed units in both armies (I do not track the army HP because there are may be difficulties with this due to possible changes in the order of unit's turns). If there were no changes in the numbers of killed units in both armies during the latest 50 turns, then the "killer switch" logic is applied as described in the PR description.

ihhub:
@oleg-derevenetz , huge thanks for improving the logic!

PR #5418, opened 2022-05-16
Cache AI hero's and neutral monsters' strength during AI turn

### LINKED ISSUE #3840: AI: cache monster strength once per turn
Currently, AI evaluates reachable monster strength for every AI hero action. We waste CPU resources by recalculating these values. What we need to do is to calculate them only once per turn of an AI player and update it only if a hero fights monsters.

ihhub:
Calculation of army strength is a heavy operation and we should avoid calling it over and over again. However, caching should be error-prone to avoid incorrect estimations of strength by AI. close #3840 

Also this pull request fixes a case when AI was ignoring shrines with "View All" spell.

### DIFF
--- a/src/fheroes2/monster/monster_info.cpp
+++ b/src/fheroes2/monster/monster_info.cpp
@@ -931,9 +931,4 @@ namespace fheroes2
 
         return spellResistance;
     }
-
-    bool isAbilityPresent( const int monsterId, const MonsterAbilityType abilityType )
-    {
-        return ::isAbilityPresent( getMonsterData( monsterId ).battleStats.abilities, abilityType );
-    }
 }
--- a/src/fheroes2/monster/monster_info.cpp
+++ b/src/fheroes2/monster/monster_info.cpp
@@ -215,9 +215,9 @@ namespace
             { M82::UNKNOWN, M82::UNKNOWN, M82::UNKNOWN, M82::UNKNOWN, M82::UNKNOWN, M82::UNKNOWN, M82::UNKNOWN, M82::UNKNOWN }, // Random Monster 4
         };
 
-        // Monster abilities and weaknesses will be added later.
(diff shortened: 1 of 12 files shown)


src/fheroes2/monster/monster_info.cpp
```diff
@@ -139,78 +218,78 @@ namespace
...
-            { 8, 8, 4, 5, 50, Speed::SLOW, 0, {}, {} }, // Earth Elemental
-            { 7, 7, 2, 8, 35, Speed::VERYFAST, 0, {}, {} }, // Air Elemental
-            { 8, 6, 4, 6, 40, Speed::FAST, 0, {}, {} }, // Fire Elemental
-            { 6, 8, 3, 7, 45, Speed::AVERAGE, 0, {}, {} }, // Water Elemental
-            { 0, 0, 0, 0, 0, Speed::VERYSLOW, 0, {}, {} }, // Random Monster
-            { 0, 0, 0, 0, 0, Speed::VERYSLOW, 0, {}, {} }, // Random Monster 1
-            { 0, 0, 0, 0, 0, Speed::VERYSLOW, 0, {}, {} }, // Random Monster 2
-            { 0, 0, 0, 0, 0, Speed::VERYSLOW, 0, {}, {} }, // Random Monster 3
-            { 0, 0, 0, 0, 0, Speed::VERYSLOW, 0, {}, {} }, // Random Monster 4
+            { 0, 0, 0, 0, 0, Speed::VERYSLOW, 0, 0, {}, {} }, // Unknown Monster
```
oleg-derevenetz, line 221: Please update the comment above to reflect the new field in the initializer list.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one minor comment, could you please take a look when you have time?

oleg-derevenetz (approved):

PR #5420, opened 2022-05-18
Speed up AI turn

ihhub:
With recent changes for the past release AI became slower and it is noticeable on slower machines. Since we are going to add more logic to AI it is important to keep balance between performance and logic complexity. This pull request improves 2 places for AI turn:
- during `RedrawTurnProgress` method call render only status window instead of re-rendering the whole frame. It releases ~10% of overall AI turns
- `ClearFog` and `getFogTileCountToBeRevealed` where relying on an internal function which creates a vector, resize it due to item insertion and frees the memory. In particular `getFogTileCountToBeRevealed` can be called a lot of time during a single AI hero's movement calculation if there are many discovered cells.

### DIFF
--- a/src/fheroes2/gui/interface_status.cpp
+++ b/src/fheroes2/gui/interface_status.cpp
@@ -109,79 +109,82 @@ void Interface::StatusWindow::SetState( const StatusType status )
 void Interface::StatusWindow::Redraw( void ) const
 {
     const Settings & conf = Settings::Get();
+    if ( conf.ExtGameHideInterface() && !conf.ShowStatus() ) {
+        // The window is hidden.
+        return;
+    }
+
     const fheroes2::Rect & pos = GetArea();
 
-    if ( !conf.ExtGameHideInterface() || conf.ShowStatus() ) {
-        if ( conf.ExtGameHideInterface() ) {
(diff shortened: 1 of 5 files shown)


src/fheroes2/gui/interface_status.cpp
```diff
@@ -109,79 +109,82 @@ void Interface::StatusWindow::SetState( const StatusType status )
 void Interface::StatusWindow::Redraw( void ) const
 {
     const Settings & conf = Settings::Get();
+    if ( conf.ExtGameHideInterface() && !conf.ShowStatus() ) {
```
ihhub, line 112: This function was modified to reduce code nesting.

src/fheroes2/gui/interface_status.cpp
```diff
@@ -339,61 +342,58 @@ void Interface::StatusWindow::DrawArmyInfo( int oh ) const
 
 void Interface::StatusWindow::DrawAITurns( void ) const
 {
-    const Settings & conf = Settings::Get();
-
-    if ( !conf.ExtGameHideInterface() || conf.ShowStatus() ) {
```
ihhub, line 344: This check was useless as the parent function takes care of it.

oleg-derevenetz (approved):

PR #5424, opened 2022-05-19
Add Swedish in-game option and font support

zenseii:
All necessary fonts were already present in CP1252.

Note that for some reason the person who made the current Swedish PO chose to replace all instances of special characters with similar latin characters. This would have to be fixed by someone competent in the language.


ihhub:
@zenseii , thank you so much for extending the list of supported languages!

PR #5425, opened 2022-05-19
Do not show languages with empty translations

ihhub:
If a language doesn't have a translation then we shouldn't even show it in the list of languages.

### DIFF
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -244,17 +244,17 @@ namespace
         bool open( const std::string & file )
         {
             StreamFile sf;
-
-            if ( !sf.open( file, "rb" ) )
+            if ( !sf.open( file, "rb" ) ) {
                 return false;
+            }
 
             {
-                size_t size = sf.size();
+                const size_t size = sf.size();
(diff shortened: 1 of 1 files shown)


src/engine/translations.cpp
```diff
@@ -277,48 +277,61 @@ namespace
             }
 
             // parse encoding and plural forms
-            if ( count ) {
+            if ( count > 0 ) {
```
zenseii, line 280: Hi @ihhub! I have only got a question regarding this if statement.
I can't seem to find out when `count` is ever less than 1. Even if I remove all translations and all obsolete translations, I still get count == 1. I tested with `fr.mo`.

ihhub: Yes, this is correct. At least one translation piece should be present but I am not 100% that piece would be empty. I think it is a flaw in mo file generation. That is why I added a check for empty lines.

zenseii: I see! The check later on works as intended,  I was just confused by this one.

zenseii (approved):

PR #5431, opened 2022-05-20
Some fixes for the victory/defeat verification logic, further improvements in informing AI about the map conditions

oleg-derevenetz:
Follow-up to #5381

* Do not allow AI to re-hire heroes related to the `WINS_HERO` or `LOSS_HERO` conditions;

* Properly handle the situation when a hero, whose loss means defeat for a human player (`LOSS_HERO`), is already hired by an AI-controlled kingdom (lose immediately);

* Properly handle the situation when a hero, whose defeat means victory for a human player (`WINS_HERO`), is already hired by a human-controlled kingdom (win immediately);

* Certain methods should only be called for a human-controlled kingdom;

* It seems that in OG AI prefers to ignore heroes related to the `WINS_HERO` condition, because AI is not able to win due to this condition. In this video blue hero can easily defeat the purple hero (specified as target hero for `WINS_HERO` condition) by his Phoenixes, but he doesn't:

https://user-images.githubusercontent.com/32623900/169627812-3f4df237-e2b8-4da2-af08-432ad3be09c0.mp4

In this PR I made an effort to replicate this behavior:

https://user-images.githubusercontent.com/32623900/169627967-5c920b2b-aca1-4088-ba14-d1a3d6e40083.mp4

Heroes from other AI-controlled kingdoms will not try to attack him or the castle where he resides, and will not overly facilitate the job of a human player. I hope that the valuable hero himself is smart enough not to get into some kind of trouble (this may be addressed later separately).

---

* I applied similar logic to the `WINS_ARTIFACT` condition - AI-controlled heroes should ignore the artifact in question, since AI is not able to win due to this condition as well:

https://user-images.githubusercontent.com/32623900/169628321-971c1ad3-7f1b-45ea-8d50-cedc9b4a5edb.mp4

It seems that OG has the similar logic:

https://user-images.githubusercontent.com/32623900/169667088-0e0781cf-fcc8-4951-a7e8-f63b3aa05882.mp4

Fortunately, AI in fheroes2 does not know how to dig Ultimate Artifact yet, but I left TODO comment for the future.

---

* I removed the `killer_color` property of `Heroes` class, which was related to the `WINS_HERO` condition. AI should not attack the related hero anymore, and if this hero still commits suicide (for example, due to an unsuccessful attack on neutrals or lack of castles), then the victory will be awarded to the first human-controlled player in turn queue. In the OG:

https://user-images.githubusercontent.com/32623900/169650632-2645b4eb-8ff0-4b25-b586-177cb12cc1f3.mp4

In the fheroes2:

https://user-images.githubusercontent.com/32623900/169650644-6d833119-0884-49e1-a416-79041b4c8aa5.mp4

In the fheroes2 if the target hero is attacked by a human-controlled player:

https://user-images.githubusercontent.com/32623900/169650747-c12ab361-eb7e-4d00-b15b-a1f14d85eb3d.mp4


src/fheroes2/game/game_highscores.cpp
```diff
@@ -144,12 +145,12 @@ fheroes2::GameMode Game::DisplayHighScores( const bool isCampaign )
...
-        std::string msg;
+        std::string msg = "Developer mode is active, the result will not be saved! \n \n Your result: %{result}";
         if ( isCampaign ) {
-            msg = std::string( "Developer mode, not save! \n \n Your result: " ) + std::to_string( Campaign::CampaignSaveData::Get().getDaysPassed() );
+            StringReplace( msg, "%{result}", std::to_string( Campaign::CampaignSaveData::Get().getDaysPassed() ) );
         }
         else {
-            msg = std::string( "Developer mode, not save! \n \n Your result: " ) + std::to_string( GetGameOverScores() );
+            StringReplace( msg, "%{result}", std::to_string( GetGameOverScores() ) );
         }
```
ihhub, line 154: I think we can do even better:
```cpp
        std::string msg = "Developer mode is active, the result will not be saved! \n \n Your result: ";
        if ( isCampaign ) {
            msg += std::to_string( Campaign::CampaignSaveData::Get().getDaysPassed();
        }
        else {
            msg += std::to_string( GetGameOverScores() );
        }
```

src/fheroes2/system/save_format_version.h
```diff
@@ -25,6 +25,7 @@
 enum SaveFileFormat : uint16_t
 {
     // TODO: if you're adding a new version you must assign it to CURRENT_FORMAT_VERSION located at the bottom.
+    FORMAT_VERSION_0916_RELEASE = 9900,
     FORMAT_VERSION_0912_RELEASE = 9803,
```
ihhub, line 29: Usually when we add a new version we kill the last old one so I think we can remove `FORMAT_VERSION_PRE_097_RELEASE` as it is a very old version also.

ihhub (changes requested):
Hi @oleg-derevenetz , I left 2 suggestions. Would you mind to take a look at them?

src/fheroes2/game/game_highscores.cpp
oleg-derevenetz: Hi @ihhub BTW should this message be translated? If no, let's do how you suggested. If yes, I propose to add `_()` and use the placeholder because in different languages this score may be in different places of the sentence.

ihhub: I think it shouldn't as it's only for developers.

PR #5434, opened 2022-05-20
Flatpak Documentation

Laserlicht:


ihhub:
My main concern is who is responsible for this distribution as it is not a part of the standard distribution. I understand that https://github.com/flathub/io.github.ihhub.Fheroes2 repository exists just for this purpose but we should avoid using manual way. I'm thinking we need to grant access to that repository our project bot `fheroes2-deploy` which will do necessary update of files upon creation of a new version.

What do you think @oleg-derevenetz ?

oleg-derevenetz:
Hi @ihhub That's what package maintainers are for. Bot is not able to test the new package version properly, it is not able to track possible changes in dependencies, etc. etc. This should be done by a human - package maintainer, which use this package or at least this package system on a regular basis.

Laserlicht:
Flathub has its own bot that checks for new tags. As soon as it detects a new tag, the YAML file is updated and the build system from there tries to create a new package for x86-64 and AArch64. The package is then automatically made available on Flathub.

https://github.com/flathub/io.github.ihhub.Fheroes2/blob/37f5a35ef58f9033064455c0e459524edc496c84/io.github.ihhub.Fheroes2.yaml#L37

oleg-derevenetz:
Hi @Laserlicht what if this bot fail to build another version? Who will be responsible for supporting this package?

Laserlicht:
Hi @oleg-derevenetz 

If the bot can't build a new version, then the worst-case scenario is that it stays with the old version.

In case of problems, you can make a pull request to the repo, which also triggers the bot for a test build.

Flatpak-specific problems should then also be created as an issue in the repo. Then I can look for it. But in principle not so many problems are to be expected. Compared to other applications, fheroes2 is manageable.

docs/INSTALL.md
```diff
@@ -192,6 +193,21 @@ be missing depending on the version of the original game) from the original game
 `~/.local/share/fheroes2`) directory. Otherwise, you can download and install the demo version of the original game by running the
 `/usr/share/fheroes2/download_demo_version.sh` script.
 
+<a name="flatpak-linux"></a>
+### Flatpak
+
+If you are using [**Flatpak**](https://flatpak.org) and [**Flathub**](https://flathub.org), you can install the game by running the following command:
+
+```flatpak install flathub io.github.ihhub.Fheroes2```
```
ihhub, line 201: Is it a flatpak or bash command? If it's a bash command please use this:
```code
```sh
flatpak install flathub io.github.ihhub.Fheroes2
```
```

File: docs/INSTALL.md [hunk 2]
```diff
@@ -192,6 +193,21 @@ be missing depending on the version of the original game) from the original game
 `~/.local/share/fheroes2`) directory. Otherwise, you can download and install the demo version of the original game by running the
 `/usr/share/fheroes2/download_demo_version.sh` script.
 
+<a name="flatpak-linux"></a>
+### Flatpak
+
+If you are using [**Flatpak**](https://flatpak.org) and [**Flathub**](https://flathub.org), you can install the game by running the following command:
```
ihhub, line 199: Please keep the length of a string to the same length as other strings. Move `following command:` to another line.

docs/INSTALL.md
```diff
@@ -192,6 +193,21 @@ be missing depending on the version of the original game) from the original game
...
+
+```flatpak install flathub io.github.ihhub.Fheroes2```
+
+And launch from the start menu or the console:
+
+```flatpak run io.github.ihhub.Fheroes2```
+
+Alternatively, you can use an application manager like Discover, which is also available on Steam Deck.
+
+You have to copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` from the original game or demo directory to the `~/.var/app/io.github.ihhub.Fheroes2/data/fheroes2` directory after the first launch.
```
ihhub, line 209: Are we asking users to run the application once, it will create a path and then copy all resource files and run the application again?

ihhub (changes requested):
Hi @Laserlicht , I left few comments in this pull request. Could you please take a look?

docs/INSTALL.md
```diff
@@ -192,6 +193,26 @@ be missing depending on the version of the original game) from the original game
...
+
+And launch from the start menu or the console:
+
+```sh
+flatpak run io.github.ihhub.Fheroes2
+```
+
+Alternatively, you can use an application manager like Discover, which is also available on Steam Deck.
+
+You have to copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` from the original game or demo directory to the `~/.var/app/io.github.ihhub.Fheroes2/data/fheroes2` directory (created after the first launch).
```
ihhub, line 214: Why can't this directory be created at the time of application installation?

ihhub:
Hi @Laserlicht , my question was whether it is possible to create needed directories by Flatpak itself to avoid double fheroes2 startup as the current approach is not user friendly?

Laserlicht:
> Hi @Laserlicht , my question was whether it is possible to create needed directories by Flatpak itself to avoid double fheroes2 startup as the current approach is not user friendly?

Hi @ihhub 

I have adjusted it a bit. This is the current behavior:

https://user-images.githubusercontent.com/13953785/172055242-3a768529-af6e-4d0d-b31c-1f3bf388bdbb.mp4

docs/INSTALL.md
```diff
@@ -192,6 +193,26 @@ be missing depending on the version of the original game) from the original game
...
+
+And launch from the start menu or the console:
+
+```sh
+flatpak run io.github.ihhub.Fheroes2
+```
+
+Alternatively, you can use an application manager like Discover, which is also available on Steam Deck.
+
+You have to copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` from the original game or demo directory to the `~/.var/app/io.github.ihhub.Fheroes2/data/fheroes2` directory. You will be asked when you start the application.
```
ihhub, line 214: Please cut the line after `directory to the` to maintain similar line width.

ihhub (approved):

oleg-derevenetz (approved):

ihhub:
@Laserlicht , thank you so much for this change!

Issue #2332, opened 2020-12-06
Adventure screen, make a left click and hold it on the adventure map (or on the graphic of a castle, or on the portrait of a hero, or anywhere in the adventure screen) and moving the cursor over the minimap area should not change the position of the crosshairs of the minimap


LeHerosInconnu:
In fheroes2, make a left click and hold it on the adventure map (or on the graphic of a castle, or on the portrait of a hero, or anywhere in the adventure screen) and moving the cursor over the minimap area should not change the position of the crosshairs of the minimap.

The view of the minimap should change only when the left click is performed in the minimap area.

This flaw is not present in the original game.

In fheroes2:

![Move minimap 01](https://user-images.githubusercontent.com/43583994/101290599-a5a45600-3803-11eb-8df2-0db180a5b754.gif)

![Move minimap 02](https://user-images.githubusercontent.com/43583994/101290610-b5239f00-3803-11eb-9930-a1c7f51f2542.gif)

![Move minimap 03](https://user-images.githubusercontent.com/43583994/101290612-bce34380-3803-11eb-8a86-3d90453bb843.gif)

Also, make a left click and hold it on the minimap and exiting the minimap area and then returning to the minimap area should not change the position of the minimap crosshairs.

The view of the minimap should only change when the left click is performed in the minimap area.

This flaw is also present in the original game.

In fheroes2:

![Move minimap 04](https://user-images.githubusercontent.com/43583994/101290626-ce2c5000-3803-11eb-99f1-522c59954625.gif)

Laserlicht:
@ihhub I think this is solved now with https://github.com/ihhub/fheroes2/pull/5878 😊

Issue #1860, opened 2020-10-01
Adventure map, allow the user to "grab and move" the adventure map to change the displayed area


LeHerosInconnu:
A maintained left click allows the user to grab the map and move it to change the displayed area.
As for the mini map but directly on the adventure map.

In fheroes2:

![Adventure map grab and move 01](https://user-images.githubusercontent.com/43583994/94833833-e51f8200-040f-11eb-9978-b275c7da20a4.png)

Branikolog:
Hi, @LeHerosInconnu!
You're suggesting to "grab and move" even, when cursor is over world map?

LeHerosInconnu:
Hello @Branikolog,

> Hi, @LeHerosInconnu!
> You're suggesting to "grab and move" even, when cursor is over world map?

Yes, directly on the adventure map, where the hero moves and interacts with objects.

Branikolog:
Humm...
I'm so got used to default moving cursor to the edges of the screen, so I even can't imagine the other way of changing the display area...
Well, your suggestion can be useful, while playing in window mode...

LeHerosInconnu:
@Branikolog.

> Humm...
> I'm so got used to default moving cursor to the edges of the screen, so I even can't imagine the other way of changing the display area...
> Well, your suggestion can be useful, while playing in window mode...

You will forget the change of the map area with the arrows on the edges of the screen when you will have used it. :)

PR #5435, opened 2022-05-20
Scroll with mouse key on gamearea

Laserlicht:
With this you can move the section on the gamearea with the right mouse button.

Facilitates the operation on the Steam Deck immensely.

Unfortunately, I'm not that much of a C++ expert at the moment and my knowledge of the codebase is also limited.

**Done**
~~Could still need some support here for:~~
- ~~Clean integration (not so hacky)~~
- ~~Faster refresh when scrolling, currently it's a bit laggy~~
- ~~quickinfos outside the gamearea should not hide the menu after moving the mouse while pressing the right mouse button~~
- ~~ignore scrolling at the edge of the screen while pressing the right mouse button~~

### DIFF
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -293,7 +293,7 @@ bool ArmyBar::ActionBarCursor( ArmyTroop & troop )
 
         if ( &troop == troop2 ) {
             msg = _( "View %{name}" );
-            StringReplace( msg, "%{name}", troop.GetName() );
+            StringReplace( msg, "%{name}", Translation::StringLower( troop.GetName() ) );
         }
         else if ( !troop.isValid() ) {
             if ( !read_only ) {
@@ -304,24 +304,24 @@ bool ArmyBar::ActionBarCursor( ArmyTroop & troop )
                     msg = _( "Move or right click to redistribute %{name}" );
                 }
 
-                StringReplace( msg, "%{name}", troop2->GetName() );
+                StringReplace( msg, "%{name}", Translation::StringLower( troop2->GetName() ) );
             }
         }
         else if ( troop.GetID() == troop2->GetID() ) {
             if ( !read_only ) {
                 msg = _( "Combine %{name} armies" );
-                StringReplace( msg, "%{name}", troop.GetName() );
+                StringReplace( msg, "%{name}", Translation::StringLower( troop.GetName() ) );
             }
         }
         else if ( !read_only ) {
             msg = _( "Exchange %{name2} with %{name}" );
-            StringReplace( msg, "%{name}", troop.GetName() );
-            StringReplace( msg, "%{name2}", troop2->GetName() );
+            StringReplace( msg, "%{name}", Translation::StringLower( troop.GetName() ) );
+            StringReplace( msg, "%{name2}", Translation::StringLower( troop2->GetName() ) );
         }
     }
     else if ( troop.isValid() ) {
         msg = _( "Select %{name}" );
-        StringReplace( msg, "%{name}", troop.GetName() );
+        StringReplace( msg, "%{name}", Translation::StringLower( troop.GetName() ) );
     }
 
     return false;
@@ -334,14 +334,14 @@ bool ArmyBar::ActionBarCursor( ArmyTroop & destTroop, ArmyTroop & selectedTroop 
     if ( destTroop.isValid() ) {
         if ( destTroop.GetID() != selectedTroop.GetID() ) {
             msg = _( "Exchange %{name2} with %{name}" );
-            StringReplace( msg, "%{name}", destTroop.GetName() );
-            StringReplace( msg, "%{name2}", selectedTroop.GetName() );
+            StringReplace( msg, "%{name}", Translation::StringLower( destTroop.GetName() ) );
+            StringReplace( msg, "%{name2}", Translation::StringLower( selectedTroop.GetName() ) );
         }
         else if ( save_last_troop )
             msg = _( "Cannot move last troop" );
         else {
             msg = _( "Combine %{name} armies" );
-            StringReplace( msg, "%{name}", destTroop.GetName() );
+            StringReplace( msg, "%{name}", Translation::StringLower( destTroop.GetName() ) );
         }
     }
     else if ( save_last_troop )
@@ -354,7 +354,7 @@ bool ArmyBar::ActionBarCursor( ArmyTroop & destTroop, ArmyTroop & selectedTroop 
             msg = _( "Move or right click to redistribute %{name}" );
         }
 
-        StringReplace( msg, "%{name}", selectedTroop.GetName() );
+        StringReplace( msg, "%{name}", Translation::StringLower( selectedTroop.GetName() ) );
     }
 
     return false;
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -842,7 +842,7 @@ void Battle::ArmiesOrder::QueueEventProcessing( std::string & msg, const fheroes
         const fheroes2::Rect unitRoi = unitPos.second + offset;
         if ( le.MouseCursor( unitRoi ) ) {
             msg = _( "View %{monster} info" );
-            StringReplace( msg, "%{monster}", unitPos.first->GetName() );
+            StringReplace( msg, "%{monster}", Translation::StringLower( unitPos.first->GetName() ) );
         }
 
         const Unit & unit = *( unitPos.first );
@@ -2180,7 +2180,7 @@ int Battle::Interface::GetBattleCursor( std::string & statusMsg ) const
         if ( b_enemy ) {
             if ( _currentUnit->GetCurrentColor() == b_enemy->GetColor() || ( _currentUnit == b_enemy ) ) {
                 statusMsg = _( "View %{monster} info" );
-                StringReplace( statusMsg, "%{monster}", b_enemy->GetMultiName() );
+                StringReplace( statusMsg, "%{monster}", Translation::StringLower( b_enemy->GetMultiName() ) );
                 return Cursor::WAR_INFO;
             }
             else {
@@ -2188,7 +2188,7 @@ int Battle::Interface::GetBattleCursor( std::string & statusMsg ) const
                     statusMsg = _( "Shoot %{monster}" );
                     statusMsg.append( " " );
                     statusMsg.append( _n( "(1 shot left)", "(%{count} shots left)", _currentUnit->GetShots() ) );
-                    StringReplace( statusMsg, "%{monster}", b_enemy->GetMultiName() );
+                    StringReplace( statusMsg, "%{monster}", Translation::StringLower( b_enemy->GetMultiName() ) );
                     StringReplace( statusMsg, "%{count}", _currentUnit->GetShots() );
 
                     return arena.IsShootingPenalty( *_currentUnit, *b_enemy ) ? Cursor::WAR_BROKENARROW : Cursor::WAR_ARROW;
@@ -2240,7 +2240,7 @@ int Battle::Interface::GetBattleCursor( std::string & statusMsg ) const
                         const int cursor = GetSwordCursorDirection( currentDirection );
 
                         statusMsg = _( "Attack %{monster}" );
-                        StringReplace( statusMsg, "%{monster}", b_enemy->GetName() );
+                        StringReplace( statusMsg, "%{monster}", Translation::StringLower( b_enemy->GetName() ) );
 
                         return cursor;
                     }
@@ -2249,7 +2249,7 @@ int Battle::Interface::GetBattleCursor( std::string & statusMsg ) const
         }
         else if ( cell->isReachableForHead() || cell->isReachableForTail() ) {
             statusMsg = _currentUnit->isFlying() ? _( "Fly %{monster} here" ) : _( "Move %{monster} here" );
-            StringReplace( statusMsg, "%{monster}", _currentUnit->GetName() );
+            StringReplace( statusMsg, "%{monster}", Translation::StringLower( _currentUnit->GetName() ) );
             return _currentUnit->isFlying() ? Cursor::WAR_FLY : Cursor::WAR_MOVE;
         }
     }
@@ -2298,7 +2298,7 @@ int Battle::Interface::GetBattleSpellCursor( std::string & statusMsg ) const
         else if ( b_stats && b_stats->AllowApplySpell( spell, _currentUnit->GetCurrentOrArmyCommander() ) ) {
             statusMsg = _( "Cast %{spell} on %{monster}" );
             StringReplace( statusMsg, "%{spell}", spell.GetName() );
-            StringReplace( statusMsg, "%{monster}", b_stats->GetName() );
+            StringReplace( statusMsg, "%{monster}", Translation::StringLower( b_stats->GetName() ) );
             return GetCursorFromSpell( spell.GetID() );
         }
         else if ( !spell.isApplyToFriends() && !spell.isApplyToEnemies() && !spell.isApplyToAnyTroops() ) {
@@ -3155,7 +3155,7 @@ void Battle::Interface::RedrawActionAttackPart2( Unit & attacker, const TargetsI
                 msg.append( " " );
                 msg.append( _n( "1 %{defender} perishes.", "%{count} %{defender} perish.", target.killed ) );
                 StringReplace( msg, "%{count}", target.killed );
-                StringReplace( msg, "%{defender}", target.defender->GetPluralName( target.killed ) );
+                StringReplace( msg, "%{defender}", Translation::StringLower( target.defender->GetPluralName( target.killed ) ) );
             }
         }
 
@@ -3320,11 +3320,11 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
 
 #ifdef DEBUG_LOG
     std::string msg = _( "Moved %{monster}: %{src}, %{dst}" );
-    StringReplace( msg, "%{monster}", unit.GetName() );
+    StringReplace( msg, "%{monster}", Translation::StringLower( unit.GetName() ) );
     StringReplace( msg, "%{src}", unit.GetHeadIndex() );
 #else
     std::string msg = _( "Moved %{monster}" );
-    StringReplace( msg, "%{monster}", unit.GetName() );
+    StringReplace( msg, "%{monster}", Translation::StringLower( unit.GetName() ) );
 #endif
 
     _currentUnit = nullptr;
@@ -3403,7 +3403,7 @@ void Battle::Interface::RedrawActionFly( Unit & unit, const Position & pos )
     }
 
     std::string msg = _( "Moved %{monster}: %{src}, %{dst}" );
-    StringReplace( msg, "%{monster}", unit.GetName() );
+    StringReplace( msg, "%{monster}", Translation::StringLower( unit.GetName() ) );
     StringReplace( msg, "%{src}", unit.GetHeadIndex() );
 
     Cursor::Get().SetThemes( Cursor::WAR_POINTER );
@@ -3499,7 +3499,7 @@ void Battle::Interface::RedrawActionResistSpell( const Unit & target, bool playS
         AudioManager::PlaySound( M82::RSBRYFZL );
     }
     std::string str( _( "The %{name} resist the spell!" ) );
-    StringReplace( str, "%{name}", target.GetName() );
+    StringReplace( str, "%{name}", Translation::StringLower( target.GetName() ) );
     status.SetMessage( str, true );
     status.SetMessage( "", false );
 }
@@ -3512,7 +3512,7 @@ void Battle::Interface::RedrawActionSpellCastStatus( const Spell & spell, int32_
 
     if ( target && ( target->GetHeadIndex() == dst || ( target->isWide() && target->GetTailIndex() == dst ) ) ) {
         msg = _( "%{name} casts %{spell} on the %{troop}." );
-        StringReplace( msg, "%{troop}", target->GetName() );
+        StringReplace( msg, "%{troop}", Translation::StringLower( target->GetName() ) );
     }
     else {
         msg = _( "%{name} casts %{spell}." );
@@ -3827,8 +3827,8 @@ void Battle::Interface::RedrawActionMonsterSpellCastStatus( const Spell & spell,
         break;
     }
 
-    StringReplace( msg, "%{attacker}", attacker.GetMultiName() );
-    StringReplace( msg, "%{target}", target.defender->GetName() );
+    StringReplace( msg, "%{attacker}", Translation::StringLower( attacker.GetMultiName() ) );
+    StringReplace( msg, "%{target}", Translation::StringLower( target.defender->GetName() ) );
 
     status.SetMessage( msg, true );
     status.SetMessage( "", false );
@@ -3842,7 +3842,7 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
     const fheroes2::Rect & pos = unit.GetRectPosition();
 
     std::string msg = isGoodLuck ? _( "Good luck shines on the %{attacker}." ) : _( "Bad luck descends on the %{attacker}." );
-    StringReplace( msg, "%{attacker}", unit.GetName() );
+    StringReplace( msg, "%{attacker}", Translation::StringLower( unit.GetName() ) );
     status.SetMessage( msg, true );
 
     Cursor::Get().SetThemes( Cursor::WAR_POINTER );
@@ -3910,13 +3910,13 @@ void Battle::Interface::RedrawActionMorale( Unit & b, bool good )
 
     if ( good ) {
         msg = _( "High morale enables the %{monster} to attack again." );
-        StringReplace( msg, "%{monster}", b.GetName() );
+        StringReplace( msg, "%{monster}", Translation::StringLower( b.GetName() ) );
         status.SetMessage( msg, true );
         RedrawTroopWithFrameAnimation( b, ICN::MORALEG, M82::GOODMRLE, NONE );
     }
     else {
         msg = _( "Low morale causes the %{monster} to freeze in panic." );
-        StringReplace( msg, "%{monster}", b.GetName() );
+        StringReplace( msg, "%{monster}", Translation::StringLower( b.GetName() ) );
         status.SetMessage( msg, true );
         RedrawTroopWithFrameAnimation( b, ICN::MORALEB, M82::BADMRLE, WINCE );
     }
@@ -3954,7 +3954,7 @@ void Battle::Interface::RedrawActionTowerPart2( const Tower & tower, const Targe
         msg += ' ';
         msg.append( _n( "1 %{defender} perishes.", "%{count} %{defender} perish.", target.killed ) );
         StringReplace( msg, "%{count}", target.killed );
-        StringReplace( msg, "%{defender}", target.defender->GetPluralName( target.killed ) );
+        StringReplace( msg, "%{defender}", Translation::StringLower( target.defender->GetPluralName( target.killed ) ) );
     }
 
     if ( !isMirror ) {
--- a/src/fheroes2/campaign/campaign_scenariodata.cpp
+++ b/src/fheroes2/campaign/campaign_scenariodata.cpp
@@ -709,7 +709,7 @@ namespace Campaign
         case ScenarioBonusData::TROOP: {
             std::string description( _( "The main hero will have %{count} %{monster} at the start of the scenario." ) );
             StringReplace( description, "%{count}", std::to_string( _amount ) );
-            StringReplace( description, "%{monster}", Monster( _subType ).GetPluralName( _amount ) );
+            StringReplace( description, "%{monster}", Translation::StringLower( Monster( _subType ).GetPluralName( _amount ) ) );
             return description;
         }
         case ScenarioBonusData::SPELL: {
--- a/src/fheroes2/castle/castle_dialog.cpp
+++ b/src/fheroes2/castle/castle_dialog.cpp
@@ -110,7 +110,7 @@ namespace
 
         const Monster monster( race, buildingId );
         std::string msgStatus = _( "Recruit %{name}" );
-        StringReplace( msgStatus, "%{name}", monster.GetMultiName() );
+        StringReplace( msgStatus, "%{name}", Translation::StringLower( monster.GetMultiName() ) );
         return msgStatus;
     }
 
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -1865,7 +1865,7 @@ void ActionToArtifact( Heroes & hero, int32_t dst_index )
                 else {
                     msg = _(
                         "Through a clearing you observe an ancient artifact. Unfortunately, it's guarded by a nearby %{monster}. Do you want to fight the %{monster} for the artifact?" );
-                    StringReplace( msg, "%{monster}", troop->GetName() );
+                    StringReplace( msg, "%{monster}", Translation::StringLower( troop->GetName() ) );
                     battle = ( Dialog::YES == Dialog::Message( title, msg, Font::BIG, Dialog::YES | Dialog::NO ) );
                 }
             }
@@ -2257,7 +2257,7 @@ void ActionToDwellingJoinMonster( Heroes & hero, const MP2::MapObjectType object
 
     if ( troop.isValid() ) {
         std::string message = _( "A group of %{monster} with a desire for greater glory wish to join you. Do you accept?" );
-        StringReplace( message, "%{monster}", troop.GetMultiName() );
+        StringReplace( message, "%{monster}", Translation::StringLower( troop.GetMultiName() ) );
 
         AudioManager::PlaySound( M82::EXPERNCE );
 


LeHerosInconnu:
Hello @Laserlicht,

Is this feature similar to this one: https://github.com/ihhub/fheroes2/issues/1860?

I think that using the left click and hold would be better, because with the right click and hold an information window is displayed.
Using the left mouse button is also more natural and convenient because it is the most used button by the user.
And many other video games use the left click and hold for this functionality. :)

In fheroes2:
(Information window with right click and hold.)

Right clic info window adventure map 01

zenseii:
> Hello @LeHerosInconnu
> 
> Thank you very much for your advice. 🙂
> 
> Yes, this is similar to your issue.
> 
> I decided to use the right mouse button, because the HoMM3 HD patch also introduced this feature this way a few months ago. It would be so that the information window is displayed and if you move the mouse while holding the button, it disappears again and you can scroll.
> 
> For the Steam Deck, the right mouse button would probably also be better, so that you do not immediately move the screen section with imprecise clicks.
> 
> But sure, that's debatable. 🙂

Good evening @LeHerosInconnu and @Laserlicht 😊

Both of your points are very good and I would just like to add that what button to use could simply be made user-configurarable in what is currently the list of "hotkeys", since that is basically the first steps of key configuration.

Also, great job implementing this @Laserlicht!

LeHerosInconnu:
Hello @Laserlicht and @zenseii,

> Both of your points are very good and I would just like to add that what button to use could simply be made user-configurarable in what is currently the list of "hotkeys", since that is basically the first steps of key configuration.

Indeed, it would be even better if the user could choose the mouse button that best suits him to perform the action.

> Also, thanks for implementing this @Laserlicht!

As @zenseii said, thanks @Laserlicht! :)

Laserlicht:
I have implemented it now times with the right mouse button. An option for the right mouse button could be added in a 2nd expansion step.

Here you can see how it behaves. With the right mouse button still and moving. The animation is not as smooth as the original due to the GIF format.

Peek 2022-05-21 16-19

LeHerosInconnu:
Hello @Laserlicht,

So I tested this in several resolutions.
It works well, good job. :)

But, in my case, the use of the left mouse button would be preferable, because I find that the right-click information window, appearing even surreptitiously, is annoying.

https://user-images.githubusercontent.com/43583994/169689234-4d0e6bde-9956-4f68-b19e-dab61690f002.mp4

So you can implement the 2nd expansion step for the use of the left mouse button. :)

Also, would it be possible to prevent the mouse cursor from blinking when the user releases the mouse button on one of the screen edges, where the arrow cursor of the screen edge should appear?

https://user-images.githubusercontent.com/43583994/169689292-528d671f-6bd4-48ca-92ea-61e1d8cb6267.mp4

Laserlicht:
Hi @LeHerosInconnu 
The bug with the flickering cursor should be gone.

I have optimized the whole thing a bit. For a left mouse button now only four lines or so would have to be adjusted. I have already tested this with me. Also works well so far.

However, I currently do not know how the settings system of fheroes2 works...
The best would be a tristate setting. Off, left and right mouse button...

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -60,6 +60,18 @@ namespace
         COUT( info )
     }
 
+    void quickInfoLoop()
+    {
+        LocalEvent & le = LocalEvent::Get();
+        fheroes2::Point mousePos = le.GetMouseCursor();
+        Interface::GameArea gameArea = Interface::Basic::Get().GetGameArea();
```
felix642, line 67: This should be a const reference.

felix642 (changes requested):
Hey @Laserlicht, I left a small comment, would you mind taking a look ? :)

LeHerosInconnu:
Hello @Laserlicht,

> Hi @LeHerosInconnu The bug with the flickering cursor should be gone.
> 
> I have optimized the whole thing a bit. For a left mouse button now only four lines or so would have to be adjusted (only in interface_gamearea.cpp). I have already tested this local. Also works well so far.

Okay, I'll take a look at it tomorrow. :)

> However, I currently do not know how the settings system of fheroes2 works...

Maybe @zenseii could advise you about it.

> The best would be a tristate setting. Off, left and right mouse button...

Good idea for the three-state setting. :)

src/fheroes2/dialog/dialog_quickinfo.cpp
Laserlicht: Hey @felix642

Thanks for your input.

Do you mean before the `LocalEvent`? When I do that, I get an error when compiling.

Sorry C++ is normally not my world. 😅

```
error: passing ‘const LocalEvent’ as ‘this’ argument discards qualifiers [-fpermissive]
   70 |         while ( le.HandleEvents() && le.MousePressRight()
```

Eldorias-Piano-Covers:
Great, thanks ! Could it be possible to make this work with the middle click (click on the wheel) ? Like Starcraft 2 style ?

LeHerosInconnu:
Hello @Laserlicht,

> Hello @Laserlicht,
> 
> > Hi @LeHerosInconnu The bug with the flickering cursor should be gone.
> > I have optimized the whole thing a bit. For a left mouse button now only four lines or so would have to be adjusted (only in interface_gamearea.cpp). I have already tested this local. Also works well so far.
> 
> Okay, I'll take a look at it tomorrow. :)

I did a test in several resolutions, both in full screen and in windowed mode, and the cursor flickering problem is gone.
Nice. :)

However, I noticed a small problem. :)
When the user releases the mouse button on the right edge of the screen and moves the mouse over the area where the main interface is located (the area with the minimap, heroes and castles lists, command buttons, etc.) and tries to scroll the adventure map area by moving the mouse cursor to the edge of the screen (where the mouse cursor should then turn into an arrow before scrolling the screen) it does not work.
The user must first move the mouse cursor to the adventure map display area before moving the mouse cursor to the edge of the screen to activate the arrow cursor which then scrolls the screen.
The same goes for the small portion of the main interface area at the top and bottom right for scrolling up and down the screen.

In fheroes2:

https://user-images.githubusercontent.com/43583994/169756333-024d0aa8-7a94-4ae6-a990-969b2a48c975.mp4

zenseii:
> > However, I currently do not know how the settings system of fheroes2 works...
> 
> Maybe @zenseii could advise you about it.

I've honestly not looked at this part of the code so I wouldn't be able to help much I'm afraid. Sorry :P

src/fheroes2/dialog/dialog_quickinfo.cpp
felix642: If you can't make the GameArea const it should at least be a reference since this object is pretty big and we don't want to copy it, but instead reference it.

felix642:
Hey @LeHerosInconnu, @Laserlicht 

I was taking a look at the setting for this new feature and I was wondering where you wanted to add the option ?
With what we currently have it's not trivial to add the new setting since we need to support a tri-state button and the only menu that support this kind of feature is this one :

image

As you can see, this menu doesn't have space for a new button unless we decide to change the layout. We would also need to add a new Icon which would require a moderate amount of work.

So before going further I would like to know where we should add the setting for this new feature ?

src/fheroes2/dialog/dialog_quickinfo.cpp
Laserlicht: Thank you! I made it a reference.

Laserlicht:
> Great, thanks ! Could it be possible to make this work with the middle click (click on the wheel) ? Like Starcraft 2 style ?

Could be pssible by extending `localevent.cpp` with some functions for the middle button.

Laserlicht:
@LeHerosInconnu 
> However, I noticed a small problem. :)

Should be gone. :)

Laserlicht:
Hi @felix642 
> As you can see, this menu doesn't have space for a new button unless we decide to change the layout. We would also need to add a new Icon which would require a moderate amount of work.

That's a general question about how you want to deal with it in the future. Maybe a second page? But yes, that is not a little effort to change.

We could also take the experimental submenu. But there it doesn't quite fit and there only binary settings are supported.

Hmm...

LeHerosInconnu:
@Laserlicht.

> @LeHerosInconnu
> 
> > However, I noticed a small problem. :)
> 
> Should be gone. :)

Tested, and yes it's gone.
Nice. :)

Laserlicht:
> Could it be set in the fheroes2.cfg file for now?
> 
> "# scroll the adventure map with the mouse button: 0 - 3 mouse button = 1"
> 
> 0: disabled 1: left mouse button 2: right mouse button 3: middle mouse button
> 
> Can be set to "0" (same operation as the original game) or to "1" (to avoid "spurious appearance" of right click windows) by default.

I am currently not familiar with the cfg file. And therefore I can only badly judge whether this is the right way.

But yes, the 4 options looks good. :)

Laserlicht:
> > Could it be set in the fheroes2.cfg file for now?
> > "# scroll the adventure map with the mouse button: 0 - 3 mouse button = 1"
> > 0: disabled 1: left mouse button 2: right mouse button 3: middle mouse button
> > Can be set to "0" (same operation as the original game) or to "1" (to avoid "spurious appearance" of right click windows) by default.
> 
> I am currently not familiar with the cfg file. And therefore I can only badly judge whether this is the right way.
> 
> But yes, the 4 options looks good. :)

I have added an option for testing: Everything is possible. 😀
It's active right now, too.

Now we just need a solution to get the variable here from somewhere else...
`const int key = 4;`

Branikolog:
Hello, everyone!
My 5 cents about this enhancement. 
I definitely agree with @LeHerosInconnu : left click feels more natural, since we mostly use left click to create a path for our hero or creature. At the same time we use single short left clicks throughout the game, so adding scroll on holding left click takes its relevant place as a unique mechanic.
While right click is mostly for info. And we mostly hold right click continuously, so we will often come into the situation, when we want to check info, hold right mouse button and accidentally move mouse a bit, what makes screen move. 
That's why in my personal point of view, right click shouldn't be used for screen scrolling at all, as it creates a mess for players. Yes, I know, that many games have exactly right click implementation of screen scrolling, but for Heroes 2 left click looks much, much more suitable. (Also, central button seems more like a more possible solution, comparing to the right click). So I don't even think it's worth making the right click scrolling available for user, as it feels not good enough.
Returning to the current implementation... 
I think, after scrolling was initiated by holding mouse button, the cursor should be fixed on the same place on the screen, while we make some scrolling activity. So we we won't be restricted by a small area of the adventure map and can scroll the entire map (limited only by our mouse pad).
Ideally, this fixed cursor should change its form, so player can recognize that scrolling is initiated.
For cursor options we have already implemented color/monochrome adjustment. So we an convert that item in menu into a separate submenu, where cursor color, scrolling manner and even cursor speed could be customized.

Laserlicht:
Hello all,

yes, I must honestly say that the left mouse button does not feel bad.

Just find it unattractive that H3 HD uses the right one. Then the operations would be different...

That you accidentally close the quick menu is unlikely. There is an offset that scrolls only after x pixels, which should prevent exactly that.

Yes, the previous implementation has the disadvantage that you can move maximum the section size on a time. But it has the advantage that you can adapt it directly to touch devices (maybe this works out of the box? - don't have a windows touch device...).

@LeHerosInconnu 
That's something I can't really influence at all. SDL decides whether it is a click or a press-and-hold.

Sure, you can introduce a minimal mouse movement (as with the right button). But that's a bit of a tradeoff between clean scrolling and usibility.

LeHerosInconnu:
Hello @Laserlicht,

> @LeHerosInconnu That's something I can't really influence at all. SDL decides whether it is a click or a press-and-hold.
> 
> Sure, you can introduce a minimal mouse movement (as with the right button). But that's a bit of a tradeoff between clean scrolling and usibility.

We could leave it as it is for now until we find a solution or get rid of SDL. :)

Laserlicht:
There is now a setting option via the cfg file.

Laserlicht:
@LeHerosInconnu Can you test the changes?

LeHerosInconnu:
Hello @Laserlicht,

> Hello @Laserlicht,
> 
> > @LeHerosInconnu Can you test the changes?
> 
> I have some bugs to report and I'll take a look at it later today or tomorrow. :)

I did some testing and this is what I noticed.

With the "left" and "middle" settings, when the right mouse button is pressed and held, and the left (or middle, depending on the setting) mouse button is clicked, the adventure map view changes.

https://user-images.githubusercontent.com/43583994/172378361-0df4adec-cdf0-461d-91ee-83bfcf682699.mp4

With the "left" setting, when the left mouse button is pressed and held from the adventure map area, and then released over the minimap area, it behaves as if a click was made on the minimap and the adventure map view is changed.
This problem is not specific to the "drag scroll" mode and is already reported here: https://github.com/ihhub/fheroes2/issues/2332.

https://user-images.githubusercontent.com/43583994/172378468-11165ea4-8e79-4634-b7ed-5f9648a082bc.mp4

LeHerosInconnu:
Hello @Laserlicht,

Thinking about the problems encountered, I have some suggestions. :)

To solve the problem presented with "drag scroll" (for example with the minimap area), wouldn't it be possible to make it so that when the mouse button is pressed and held from the area of the adventure map (in the example the red frame) the mouse cursor cannot leave the area of the red frame (so, as long as the mouse button is not released)?

Adventure area cadre rouge 01

Also for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button), would it be possible to make the adventure map area start scrolling only when the mouse button is clicked and held AND the mouse cursor has moved at least 10 pixels (or another value)?

LeHerosInconnu:
Hello @Branikolog,

> Hi, @Laserlicht . I believe, current implementation of drag-scrolling could be quite comfortable for devices with touch screen. But at PC we don't have a restriction regarding scrolling area (I mean physical screen size), so it would be much more effective to lock cursor at single place and initiate infinite screen scrolling while moving mouse into single direction.

What exactly do you mean by "lock cursor at single place and initiate infinite screen scrolling while moving mouse into single direction."?
It looks like the same scrolling of the adventure map as with the edges of the screen but from a point defined with the mouse cursor, right?

The advantage of "drag scroll" is that it is much more precise and convenient to use than a function that would be similar to scrolling like with the screen edges.

LeHerosInconnu:
Hello @Branikolog,

> Hello, @LeHerosInconnu Sorry for my thought being explained not clear. :)

No worries. :)

> Current implementation has a limitation: we press mouse button, move mouse, but the scrolling will stop, when the cursor reaches the edges of the adventure map screen. In a case we press mouse button, while cursor is close to the left edge, for instance, and start moving it to the left side of the screen, we can move the visible area just on a little distance.
> 
> > lock cursor at single place and initiate infinite screen scrolling while moving mouse into single direction
> 
> By this, I mean the next behaviour: We press mouse button and start moving mouse. On the screen we will see, that cursor stays at the same place regarding the edges of the monitor, but scrolling will be possible to any direction until we stop moving mouse.
> 
> My idea is to lock the cursor on the same place of the screen and proceed scrolling without the limitation of the adventure map frames.

Okay, I see.
It works exactly the same way, but it is no longer the position of the mouse cursor that drives this, but the movement pulses sent by the mouse as a device.
Not sure if this is really necessary.
And what about from a programming point of view?

The "drag scroll" is convenient to observe the surroundings close to the hero or the surroundings a little further away:

https://user-images.githubusercontent.com/43583994/172885774-22a91c18-8e3d-4b7c-962a-69710d896fe9.mp4

Or, for example, observe the adventure map according to the path a hero will follow:

https://user-images.githubusercontent.com/43583994/172886130-c3ed16e9-e21d-44a9-a72d-30f08925069d.mp4

To view larger parts of the adventure map, using the minimap's "drag scroll" is much more convenient:

https://user-images.githubusercontent.com/43583994/172886333-2b899c03-b24e-4fd5-87c9-e38a0e3cd55e.mp4

Laserlicht:
@LeHerosInconnu All your bugs should be gone. For the left click I think I also found a solution based on your idea. 😄
I had to remove the 'all' option only because of this. It was too much work for me to implement it cleanly. But that should not be a problem (except for testing).

@Branikolog I understand how you mean that. But I don't think it's that easy to implement from a program point of view. You would have to go deep into the event system (to reset or hold the cursor, for example). That's why I'm not going to implement it that way. If there is a need, this can still be implemented as a selectable option.

I think the current variant is better. It is also simply more familiar. From smartphones/tablets and other software such as Google Maps, for example.
Sure, in a small window, the variant would have great advantages. But I think most play full screen or maximized. There, the advantage is limited to pressing/releasing the button when you lift the mouse.

LeHerosInconnu:
Hello @Laserlicht,

> Many thanks to you!

You are welcome. :)
 
> @LeHerosInconnu All your bugs should be gone. For the left click I think I also found a solution based on your idea. smile

Nicely done. :)

> I had to remove the 'all' option only because of this. It was too much work for me to implement it cleanly. But that should not be a problem (except for testing).

It's all right for me.

> Can you test all the options? blush

Here are my observations after some tests.

**With the "left" setting.**

If the left mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.

https://user-images.githubusercontent.com/43583994/173229121-78f139c8-2f70-4774-a561-7cddc8d802e6.mp4

The "drag scroll" can be initiated if the left mouse button is pressed and held from the interface area on the right side of the adventure map (minimap, heroes and castles lists, command buttons area, status window on the bottom right of the screen).

https://user-images.githubusercontent.com/43583994/173229175-1a4d8c06-ae4c-4c17-9101-b68a9372ec50.mp4

The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" works perfectly.
Good job! :)

**With the "right" setting.**

If the right mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.
(Same as with the "left" setting.)

The "drag scroll" can be initiated if the right mouse button is pressed and held from some parts of the interface area on the right side of the adventure map (except minimap, heroes and castles lists, command buttons area, status window on the bottom right of the screen).
(Partially the same as with the "left" setting.)

https://user-images.githubusercontent.com/43583994/173229798-5d2a0fd7-983b-4f70-b9e7-89c5a874d0b2.mp4

The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "right" setting.

**With the "middle" setting.**

If the middle mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.
(Same as with the "left" and "right" settings.)

The "drag scroll" can be initiated if the middle mouse button is pressed and held from the interface area on the right side of the adventure map (minimap, heroes and castles lists, command buttons area, status window on the bottom right of the screen).
(Same as with the "left" setting.)

The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "middle" setting.

This bug is still present:
> With the "left" and "middle" settings, when the right mouse button is pressed and held, and the left (or middle, depending on the setting) mouse button is clicked, the adventure map view changes.
> 
> https://user-images.githubusercontent.com/43583994/172378361-0df4adec-cdf0-461d-91ee-83bfcf682699.mp4

This problem is back:
> However, I noticed a small problem. :)
> When the user releases the mouse button on the right edge of the screen and moves the mouse over the area where the main interface is located (the area with the minimap, heroes and castles lists, command buttons, etc.) and tries to scroll the adventure map area by moving the mouse cursor to the edge of the screen (where the mouse cursor should then turn into an arrow before scrolling the screen) it does not work.
> The user must first move the mouse cursor to the adventure map display area before moving the mouse cursor to the edge of the screen to activate the arrow cursor which then scrolls the screen.
> The same goes for the small portion of the main interface area at the top and bottom right for scrolling up and down the screen.
> 
> In fheroes2:
> 
> https://user-images.githubusercontent.com/43583994/169756333-024d0aa8-7a94-4ae6-a990-969b2a48c975.mp4

This is not too annoying for the left, the top and the bottom of the screen (in fact I prefer this mode of operation as the user is not "surprised" by the two modes working "at the same time"), the problem comes from the right side of the screen because the user has to make a big back and forth with the mouse cursor to be able to activate again the mode of operation with the edge of the screen.

Laserlicht:
> If the left mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.
Hmm...

I have removed the radar fix. Otherwise it will bring new problems.

I now think that we should not deal with this bug here. Since it actually has nothing to do with the drag scroll.

Theoretically the bug would be easy to fix. But the input layer of fheroes2 is there... not so good...

Currently there are the functions:
- MouseClick - One cycle active - when a click is done (mouse press and release).
- MousePress - Permanently active - when button is pressed
- MouseRelease - One cycle active - when button is released

one is missing:
- "MouseOnPress" - One cycle active - when button is pressed.

But then you should rename the others too, otherwise it's not really nice. This is then a bigger change.

But with the MouseOnPress one could determine whether the button was pressed first on the radar/minimap and not already somewhere else.

> The "drag scroll" can be initiated if the left mouse button is pressed and held from the interface area on the right side of the adventure map (minimap, heroes and castles lists, command buttons area, status window on the bottom right of the screen).

Same problem as with the radar. But does not really bother. I think.

> If the right mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.
(Same as with the "left" setting.)

Should be gone by removing the fix.

> The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "right" setting.

It should. However, as long as the quickinfo is still open. It only closes when there is a "larger" movement.

> The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "middle" setting.

That's right. Didn't really make sense to me. Because you only press the middle button when you want to do the drag scroll. 😉

> This bug is still present:

Is now gone.

> This problem is back:

It's gone again. Could take it out again with the removal of the radar fix.

LeHerosInconnu:
Hello @Laserlicht,

> Thank you for your support!
> 
> > If the left mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.
> > Hmm...
> 
> I have removed the radar fix. Otherwise it will bring new problems.
> 
> I now think that we should not deal with this bug here. Since it actually has nothing to do with the drag scroll.
> 
> Theoretically the bug would be easy to fix. But the input layer of fheroes2 is there... not so good...
> 
> Currently there are the functions:
> 
> * MouseClick - One cycle active - when a click is done (mouse press and release).
> * MousePress - Permanently active - when button is pressed
> * MouseRelease - One cycle active - when button is released
> 
> one is missing:
> 
> * "MouseOnPress" - One cycle active - when button is pressed.
> 
> As an electrician, one would say that the rising edge is missing. smile
> 
> But then you should rename the others too, otherwise it's not really nice. This is then a bigger change.
> 
> But with the MouseOnPress one could determine whether the button was pressed first on the radar/minimap and not already somewhere else.

Okay, I got it. :)

> > The "drag scroll" can be initiated if the left mouse button is pressed and held from the interface area on the right side of the adventure map (minimap, heroes and castles lists, command buttons area, status window on the bottom right of the screen).
> 
> Same problem as with the radar. But does not really bother. I think.

It's not too annoying because it doesn't interfere with other actions, it was just to approach "perfection" and "peace of mind". :)

> > If the right mouse button is pressed and held from the adventure map area and the mouse cursor is moved over the minimap area and the mouse button is released, the left mouse click over the minimap area to change the minimap view is not taken into account, the right mouse click over the minimap area is correctly taken into account. You have to go back to the adventure map area and then back to the minimap area for everything to work as expected.
> > (Same as with the "left" setting.)
> 
> Should be gone by removing the fix.

Yes, it's gone.

> > The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "right" setting.
> 
> It should. However, as long as the quickinfo is still open. It only closes when there is a "larger" movement.

With the "right" setting, the right-click information window no longer works on the adventure map area.

https://user-images.githubusercontent.com/43583994/173310523-5ff3f03c-1067-42f3-995c-099cf83dbdc0.mp4

> > The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "middle" setting.
> 
> That's right. Didn't really make sense to me. Because you only press the middle button when you want to do the drag scroll. wink

Yes, you are right, the middle mouse button is not currently used on the adventure map (if I don't forget anything :) ), it will always be possible to complete it later if necessary.

> > This bug is still present:
> 
> Is now gone.

Yes, it's fixed.

> > This problem is back:
> 
> It's gone again. Could take it out again with the removal of the radar fix.

Okay.

Laserlicht:
right should work again

LeHerosInconnu:
Hello @Laserlicht,

> With the "right" setting, the right-click information window no longer works on the adventure map area.

> right should work again

Yes, everything works fine.

> > > The condition of moving the mouse cursor (for the problem of the not very precise click of the user (he clicks on a place and moves the mouse cursor slightly before releasing the mouse button)) to trigger the "drag scroll" does not seem to be applied with the "right" setting.
> > 
> > 
> > It should. However, as long as the quickinfo is still open. It only closes when there is a "larger" movement.

Compared to the "left" setting, the "right" setting does not seem to take into account a distance covered by the mouse cursor before performing the "drag scroll" of the adventure map.

In fheroes2:

https://user-images.githubusercontent.com/43583994/174318875-75a1da53-8a63-45c2-b15a-7e0ec412eb92.mp4

Laserlicht:
@LeHerosInconnu 
>Compared to the "left" setting, the "right" setting does not seem to take into account a distance covered by the mouse cursor before performing the "drag scroll" of the adventure map.

The right mouse button also has this function. However, it works a little differently. The offset is taken into account when quickinfo is open. So if you have moved at least 4 pixels, the window closes.

Of course, the 4 pixels are less than the 10 with the left mouse button. But I think that the quickinfo window then disturbs less.

https://github.com/ihhub/fheroes2/blob/2738da6f03b08ca409284f51c91b894ec5ccf3ac/src/fheroes2/dialog/dialog_quickinfo.cpp#L71

LeHerosInconnu:
Hello @Laserlicht,

> @LeHerosInconnu
> 
> > Compared to the "left" setting, the "right" setting does not seem to take into account a distance covered by the mouse cursor before performing the "drag scroll" of the adventure map.
> 
> The right mouse button also has this function. However, it works a little differently. The offset is taken into account when quickinfo is open. So if you have moved at least 4 pixels, the window closes.
> 
> Of course, the 4 pixels are less than the 10 with the left mouse button. But I think that the quickinfo window then disturbs less.
> 
> https://github.com/ihhub/fheroes2/blob/2738da6f03b08ca409284f51c91b894ec5ccf3ac/src/fheroes2/dialog/dialog_quickinfo.cpp#L71

Okay then.
Let's leave it like that for now and see if there is any feedback from the players.

zenseii:
Hi @Laserlicht! What is the status of this PR? Is it ready for review?

src/engine/localevent.cpp
```diff
@@ -1521,6 +1521,16 @@ bool LocalEvent::MouseReleaseLeft() const
...
+bool LocalEvent::MousePressMiddle() const
+{
+    return ( modes & MOUSE_PRESSED ) && SDL_BUTTON_MIDDLE == mouse_button;
+}
+
+bool LocalEvent::MouseReleaseMiddle() const
+{
+    return ( modes & MOUSE_RELEASED ) && SDL_BUTTON_MIDDLE == mouse_button;
+}
+
```
ihhub, line 1533: Please make these methods inline: move them to the header.

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -60,6 +60,18 @@ namespace
         COUT( info )
     }
 
+    void quickInfoLoop()
+    {
+        LocalEvent & le = LocalEvent::Get();
+        fheroes2::Point mousePos = le.GetMouseCursor();
```
ihhub, line 66: Please make this variable const reference to avoid memory copy.

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -60,6 +60,18 @@ namespace
...
+        Interface::GameArea & gameArea = Interface::Basic::Get().GetGameArea();
+        bool overGamearea = le.MouseCursor( gameArea.GetROI() );
```
ihhub, line 68: This variable can be const. Also since we use an opposite value we have to cache it to avoid extra computations.

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -60,6 +60,18 @@ namespace
...
+
+        while ( le.HandleEvents() && le.MousePressRight()
+                && ( ( abs( mousePos.x - le.GetMouseCursor().x ) < 4 && abs( mousePos.y - le.GetMouseCursor().y ) < 4 ) || !overGamearea || !gameArea.isDragScroll() ) )
+            ;
```
ihhub, line 72: This loop should be modified to be much more readable:
```cpp
while ( le.HandleEvents() ) {
    if ( !le.MousePressRight() ) {
        break;
    }

    if ( outsideGameArea ) {
        continue;
    }

    if ( gameArea.isDragScroll() && ( abs( mousePos.x - le.GetMouseCursor().x ) >= 4 || abs( mousePos.y - le.GetMouseCursor().y ) >= 4 ) {
        break;
    }
}

```

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -638,38 +638,92 @@ void Interface::GameArea::SetScroll( int direct )
...
 
+bool Interface::GameArea::isDragScroll()
+{
+    if ( !isDragButtonPressed() )
+        dragScroll = false;
+
+    return dragScroll;
+}
+
+void Interface::GameArea::SetDragScroll( bool v )
```
ihhub, line 649: This method should be inlined.

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -638,38 +638,88 @@ void Interface::GameArea::SetScroll( int direct )
...
+
+bool Interface::GameArea::isDragButtonPressed()
+{
+    LocalEvent & le = LocalEvent::Get();
+    const Settings & conf = Settings::Get();
+
+    if ( conf.DragScroll() == DRAG_SCROLL_NONE ) {
+        return false;
+    }
+    else if ( conf.DragScroll() == DRAG_SCROLL_LEFT ) {
```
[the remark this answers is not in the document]
ihhub: clang-tidy is right about this. **else** should be removed in this method.

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -638,38 +638,92 @@ void Interface::GameArea::SetScroll( int direct )
...
+    if ( !dragScrollPressedLast && isDragButtonPressed() )
+        dragScrollPos = mp;
+
+    if ( isDragButtonPressed() ) {
+        if ( conf.DragScroll() == DRAG_SCROLL_LEFT ) {
+            if ( abs( dragScrollPos.x - mp.x ) > 10 || abs( dragScrollPos.y - mp.y ) > 10 )
+                SetDragScroll( true );
+        }
+        else
+            SetDragScroll( true );
```
ihhub, line 690: Please add parentheses here.

src/fheroes2/gui/interface_gamearea.h
```diff
@@ -147,6 +150,9 @@ namespace Interface
 
         int32_t _prevIndexPos;
         int scrollDirection;
+        bool dragScroll;
+        bool dragScrollPressedLast;
+        fheroes2::Point dragScrollPos;
```
ihhub, line 155: Please name all new members with an underscore characters per our standard.

src/fheroes2/system/settings.cpp
```diff
@@ -459,6 +493,9 @@ std::string Settings::String() const
     os << std::endl << "# enable monochrome (black and white) cursors in the game" << std::endl;
     os << "monochrome cursor = " << ( opt_global.Modes( GLOBAL_MONOCHROME_CURSOR ) ? "on" : "off" ) << std::endl;
 
+    os << std::endl << "# drag scroll: none, left, right, middle" << std::endl;
```
ihhub, line 496: Could you please explain why we need this option in settings?

src/fheroes2/dialog/dialog_quickinfo.cpp
ihhub: Also could you please explain the magic number 4?

src/fheroes2/game/game_startgame.cpp
```diff
@@ -890,7 +890,7 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
             break;
         }
 
-        if ( fheroes2::cursor().isFocusActive() ) {
+        if ( fheroes2::cursor().isFocusActive() && !gameArea.isDragScroll() ) {
```
ihhub, line 893: I am curious: essential dragging is the same as scrolling. Why are we putting this condition here?

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -638,38 +638,92 @@ void Interface::GameArea::SetScroll( int direct )
...
+    return false;
+}
+
 void Interface::GameArea::QueueEventProcessing()
 {
     LocalEvent & le = LocalEvent::Get();
     const fheroes2::Point & mp = le.GetMouseCursor();
+    const Settings & conf = Settings::Get();
+
+    if ( !dragScrollPressedLast && isDragButtonPressed() )
```
ihhub, line 681: Shouldn't the logic be like this:
- check if a middle mouse button present. If not proceed with the existing logic
- if it is then check if the state was remembered. If it was move the area to the position based on distance difference
- if it is not remember the mouse position

I don't understand why we need this complex solution. Could you please explain / update the pull request description?

ihhub (changes requested):
Hi @Laserlicht , I left several comments here. Could you please take a look at them?

zenseii:
Hi everyone,
I just realized that in fheroes2 we already have drag n move with the left mouse button implemented in all of the "View artifacts/resources/etc." spell windows. To be consistent within the UI we might want to use the left-mouse button, and therefore go ahead with this PR rather than the other one?

In the original game there is no such functionality in those screens, althrough the mini-map is always used with clicking the left mouse button and dragging.

This is just a thought and a suggestion.

PR #5438, opened 2022-05-22
Improved error message if hero tries to open spell book and doesn't have one

### LINKED ISSUE #5437: Artefact Text - Spell Scroll Description is confusing
### Preliminary checks

- [ ] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

The description of the "Spell Scroll" type artifacts indicates the following text: "This spell scroll gives your hero the ability to cast the [...] spell".
However, if the hero does not have a magic book but has one of these artefacts, the icon for casting a spell is disabled, and when the "C" key is pressed to cast a spell, a popup warns that there is "no spell to to cast".

Should this artefact gives the possibility to cast a spell even without magic book ? Or does it should add & write a spell "In" the magic book ?

The texts must surely be readjusted to avoid any confusion.

### Save file

[spellscrolltext.zip](https://github.com/ihhub/fheroes2/files/8747414/spellscrolltext.zip)

### Additional info

In savegame : 
Ector Hero does not have any Magic Book
Ector Hero is carrying Spell Scroll for Dimension Door spell.
Ector Hero has 10 mana, and has movement points left.
Description of Spell Scroll : "This spell scroll gives your hero the ability to cast the Dimension Door spell"
Description of Magic Book : "The Magic Book enables you to cast spells".

felix642:
References #5437 

Currently, if the player tries to open his spell book and doesn't have one he will receive the message "No spell to cast.". This is true in most case, but misleading if the heroes has spell artifacts. So I changed the message from "No spell to cast." to "Spell book is not present."
image

Also, this text is already part of the game so it doesn't require any further work to translate it : 
image

### DIFF
--- a/files/lang/be.po
+++ b/files/lang/be.po
@@ -8192,6 +8192,9 @@ msgid ""
 "damage from a direct attack."
 msgstr ""
 
+msgid "You have no Magic Book, so you cannot cast a spell."
+msgstr ""
+
 msgid "No spell to cast."
 msgstr ""
 
--- a/files/lang/be.po
+++ b/files/lang/be.po
@@ -7509,7 +7509,7 @@ msgstr ""
 msgid "Spell Scroll"
 msgstr ""
 
-msgid "This %{name} gives your hero the ability to cast the %{spell} spell."
+msgid "This %{name} gives your hero the ability to cast the %{spell} spell if your hero has a Magic Book."
 msgstr ""
 
 msgid ""
@@ -8191,6 +8191,7 @@ msgid ""
 "Turns the affected creature into stone.  A petrified creature receives half "
 "damage from a direct attack."
 msgstr ""
+
 
 msgid "You have no Magic Book, so you cannot cast a spell."
 msgstr ""
--- a/files/lang/bg.po
+++ b/files/lang/bg.po
@@ -8178,6 +8178,9 @@ msgid ""
 "damage from a direct attack."
 msgstr ""
 
+msgid "You have no Magic Book, so you cannot cast a spell."
+msgstr ""
+
 msgid "No spell to cast."
 msgstr ""
 
--- a/files/lang/bg.po
+++ b/files/lang/bg.po
@@ -7495,7 +7495,7 @@ msgstr ""
 msgid "Spell Scroll"
 msgstr ""
 
-msgid "This %{name} gives your hero the ability to cast the %{spell} spell."
+msgid "This %{name} gives your hero the ability to cast the %{spell} spell if your hero has a Magic Book."
 msgstr ""
 
 msgid ""
--- a/files/lang/cs.po
+++ b/files/lang/cs.po
@@ -8400,6 +8400,9 @@ msgid ""
 "damage from a direct attack."
 msgstr ""
 
+msgid "You have no Magic Book, so you cannot cast a spell."
+msgstr ""
+
 msgid "No spell to cast."
 msgstr ""
 
--- a/files/lang/cs.po
+++ b/files/lang/cs.po
@@ -7715,7 +7715,7 @@ msgstr ""
 msgid "Spell Scroll"
 msgstr ""
 
-msgid "This %{name} gives your hero the ability to cast the %{spell} spell."
+msgid "This %{name} gives your hero the ability to cast the %{spell} spell if your hero has a Magic Book."
 msgstr ""
 
 msgid ""
--- a/files/lang/cs.po
+++ b/files/lang/cs.po
@@ -8400,6 +8400,7 @@ msgid ""
 "damage from a direct attack."
 msgstr ""
 
+#, fuzzy
 msgid "You have no Magic Book, so you cannot cast a spell."
 msgstr ""
 
--- a/files/lang/de.po
+++ b/files/lang/de.po
@@ -8769,7 +8769,8 @@ msgstr "Dummy 4"
 msgid "Spell Scroll"
 msgstr "Zauberspruchrolle"
 
-msgid "This %{name} gives your hero the ability to cast the %{spell} spell."
+#, fuzzy
+msgid "This %{name} gives your hero the ability to cast the %{spell} spell if your hero has a Magic Book."
 msgstr ""
 "Diese %{name} verleiht Ihrem Helden die Fähigkeit, den %{spell} Zauber zu "
 "wirken."
--- a/files/lang/de.po
+++ b/files/lang/de.po
@@ -9621,6 +9621,10 @@ msgstr ""
 "Verwandelt die betroffene Kreatur in Stein.  Eine versteinerte Kreatur "
 "erleidet bei einem direkten Angriff die Hälfte des Schadens."
 
+#, fuzzy
+msgid "You have no Magic Book, so you cannot cast a spell."
+msgstr ""
+
 msgid "No spell to cast."
 msgstr "Kein Zauber zu wirken."
 
--- a/files/lang/es.po
+++ b/files/lang/es.po
@@ -9272,9 +9272,8 @@ msgid ""
 "damage from a direct attack."
 msgstr ""
 
-#, fuzzy
 msgid "You have no Magic Book, so you cannot cast a spell."
-msgstr ""
+msgstr "No tienes un Libro de Hechizos, por lo que no puedes lanzar un hechizo."
 
 msgid "No spell to cast."
 msgstr "No hay hechizo que lanzar."
--- a/files/lang/es.po
+++ b/files/lang/es.po
@@ -8459,8 +8459,9 @@ msgstr "Momia"
 msgid "Spell Scroll"
 msgstr "Pegamino del Hechizo"
 
-msgid "This %{name} gives your hero the ability to cast the %{spell} spell."
-msgstr ""
+#, fuzzy
+msgid "This %{name} gives your hero the ability to cast the %{spell} spell if your hero has a Magic Book."
+msgstr "Este %{name} le da a tu héroe la habilidad de lanzar el hechizo %{spell} si tu héroe tiene un Libro de Hechizos"
 
 msgid ""
 "You find an elaborate container which houses an old vellum scroll. The runes "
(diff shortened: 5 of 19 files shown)


felix642:
@zenseii @Eldorias-Piano-Covers

I would like to know what you think of this change ?

LeHerosInconnu:
Hello @felix642,

Here are some proposals to stay in the mood of the game texts:

"No Magic Book to cast a spell."

"You have no Magic Book to cast a spell."

"You have no Magic Book."

"You don't have a Magic Book to cast a spell."

"You don't have a Magic Book."

"No Magic Book equipped."

"No Magic Book available."

"No Magic Book."

In fheroes2:

No magic book message 01

zenseii:
Hi @LeHerosInconnu  and @felix642 !

I agree with this change.

It should be noted that the original game uses "spell book" and "magic book" interchangeably, like here when buying a "spell book":
<details>

<summary> Large picture </summary>

image
</details>

However, I believe we should go for consistency so my suggestion is to solely use "magic book".

The wording should reflect that the player tries to cast a spell, and so the feedback should be simple and clear why this doesn't work. That's why for example "No Magic Book." in my opinion leaves a new player only with questions.

"You have no Magic Book and therefore no spells to cast." of "No spells to cast - You do not have a Magic Book.". (we should avoid using shortenened "don't" etc.)

Maybe I'm wrong about when this message pops up?

LeHerosInconnu:
@felix642 and @zenseii.
 
> Note that the original game uses "spell book" and "magic book" interchangeably, like here when buying a "spell book":
> However, I believe we should go for consistency so my suggestion is to solely use "magic book" since it's more frequently found.

Yes, Magic Book should be used everywhere.
It can also be considered that when no spell is written in the book, it is a Spell Book (spells are written right after the book is purchased).
This would also be valid for the three types of shrines where it is possible to buy a book when the option "heroes: allow buy a spellbook from Shrines" is activated.

> The wording should reflect that the player tries to cast a spell, and so the feedback should be simple and clear why this doesn't work. That's why for example "No Magic Book." in my opinion leaves a new player only with questions.
> 
> "You have no Magic Book and therefore no spells to cast." of "No spells to cast - You do not have a Magic Book.". (we should avoid using shortenened "don't" etc.)

This one should be fine:

"You have no Magic Book, so you can't cast a spell."

In fheroes2 (when the hero cannot cast a spell):

You can't cast a spell 01

zenseii:
@LeHerosInconnu 

I suggest to use cannot since we shouldn't write informally "You have no magic book, so you cannot cast a spell." 

Also this message pops up in the original when you press C on the adventure map. The OG message is simply "No spells to cast."

felix642:
Hi @zenseii @LeHerosInconnu,

I was using this message since it was already present in the game and reflected more accurately the reason why the player could not open his spell book. 

If you feel like it does not currently fit in the mood of the game, I can change it I don't mind, but we have to keep in mind that we'll need to translate the new message.

> Also @felix642 , I believe you tagged the wrong issue? The one you are referencing is about another problem.

I think I did reference the right issue ? I changed the message since in the issue the player actually had a spell to cast but no magic book, so the message was confusing. 

This PR was mainly use to clarify the error. I do not think a player should be able to cast a spell without a spell book (even if he has an artifact). But I don't think everyone agrees with me, so I'll leave the issue open until we decide how we want to handle that case.

LeHerosInconnu:
@zenseii.

> @LeHerosInconnu
> 
> I suggest to use cannot since we shouldn't write informally "You have no magic book, so you cannot cast a spell."

Why not.
And for the other texts already present in the game and where "can't" is used (as for the example above with the "Town Gate" spell)?

> Also this message pops up in the original when you press C on the adventure map. The OG message is simply "No spells to cast."

There is also the case when a Magic Book has been given to a Knight or a Barbarian hero in the editor and this one is therefore without spell as long as the hero doesn't visit a city with a mage guild, or a shrine on the adventure map, or doesn't get a spell scroll, or doesn't visit a pyramid (if he has the required wisdom level).
So the text "No spell to cast." should be conserved.

felix642:
> > @LeHerosInconnu
> > Also this message pops up in the original when you press C on the adventure map. The OG message is simply "No spells to cast."
> 
> There is also the case when a Magic Book has been given to a Knight or a Barbarian hero in the editor and this one is therefore without spell as long as the hero doesn't visit a city with a mage guild, or a shrine on the adventure map, or doesn't get a spell scroll, or doesn't visit a pyramid (if he has the required wisdom level). So the text "No spell to cast." should be conserved.

Yes, the message "No spell to cast." will still be displayed if the hero has a spell book but no spells in it.

zenseii:
@felix642 

The issue you referenced is about changing the text description when right clicking a spell scroll/parchment, whereas the text you changed is what happens when you press the hand icon on the adventure map, or press 'C' to "Cast". I think "No spells to cast" is more fitting as a reply to a player who just attempted to "Cast", and not just open a Magic Book to look at the spells, from what I understand. Maybe I'm overlooking something?

felix642:
@zenseii, to quote the issue,

felix642:
@zenseii to quote the issue :
> However, if the hero does not have a magic book but has one of these artefacts, the icon for casting a spell is disabled, and when the "C" key is pressed to cast a spell, a popup warns that there is "no spell to to cast".

The message "No spell to cast" is misleading in that case. The hero actually has a spell to cast (The one provided by the artifact), but does not have a spell book. So the error message "No spell to cast" is not accurate.

LeHerosInconnu:
@felix642.

> I do not think a player should be able to cast a spell without a spell book (even if he has an artifact). But I don't think everyone agrees with me, so I'll leave the issue open until we decide how we want to handle that case.

Yes, the hero can only cast a spell if he has a magic book (it works this way in the original game).
So if he has artifacts that give him spells, he can't cast them until he has a magic book.
On the other hand, if the hero has a magic book, he can cast spells of any level he has obtained in a spell scroll (or artifact), even if he does not have the necessary level of wisdom.

zenseii:
The confusing text that @Eldorias-Piano-Covers said was confusing is this one:
image.

You can add a "... if your hero has a Magic Book." or something of that vein to that text because this text is misleading the player to think that they can cast a spell as long as they only have the spell scroll.

That's why I suggested to at least change this text to not mislead the player, as it clearly did because @Eldorias-Piano-Covers believed it was a bug that he could not cast the spell.

Yes, later on we will have to decide if having a spell scroll should grant you the possibility to cast a spell, even without a magic book, but this simple text issue can be solved now before we undertake what might turn out to be a lengthy discussion about whether to change the implementation or not.

felix642:
> The text that @Eldorias-Piano-Covers said was confusing is this one: image.
> 
> You can add a "... if your hero has a Magic Book." or something of that vein to that text because this text is misleading the player to think that they can cast a spell as long as they only have the spell scroll.
> 
> That's why I suggested to at least change this text to not mislead the player, as it clearly did because @Eldorias-Piano-Covers believed it was a bug that he could not cast the spell.

I kind of disagree on this one. I feel like needing a spell book to cast spells is trivial and adding the info in the popup windows simply adds redundant information. 
I think that the main reason why @Eldorias-Piano-Covers  was confused was because the error message was misleading and not accurate (Hence why I changed it). 

But again, this is my personal opinion. If @LeHerosInconnu or someone else agrees with your opinion @zenseii I'll add this information to the popup window.

Eldorias-Piano-Covers:
Hello, yes I was surprised, because I picked up the "spell scroll" artifact which allows to cast the  adventure spell "Dimensional Door". My hero has enough mana points, but doesn't have a spellbook.
So I was wondering A) if this was a bug and my hero should be able to cast this spell even without a magic book,
or B) if the artifact's description text was correct.

zenseii:
> I kind of disagree on this one. I feel like needing a spell book to cast spells is trivial and adding the info in the popup windows simply adds redundant information. 
I think that the main reason why @Eldorias-Piano-Covers  was confused was because the error message was misleading and not accurate (Hence why I changed it). 


<details>
<summary> Initial Discord bug report </summary>

image

</details>

Anyways, @Eldorias-Piano-Covers will be able to verify themselves what they meant the problem was 😅.

I guess we have to agree to disagree, @felix642 . I believe that saying "gives your hero the ability to" is at best very ambiguous, but honestly just a lie if you don't have a Magic Book.

Eldorias-Piano-Covers:
For now, yes, we can juste edit the description text. For the future, I think it would be cool to be able to cast a spell with its spell scroll artefact, consireding those artefact as tiny magic books with one spell inside ;)

felix642:
> > I kind of disagree on this one. I feel like needing a spell book to cast spells is trivial and adding the info in the popup windows 
> I guess we have to agree to disagree, @felix642 . I believe that saying "gives your hero the ability to" is at best very ambiguous, but honestly just a lie if you don't have a Magic Book.

It's not that I don't find the text misleading is just that adding "if your hero has a Magic Book" sounds weird to me.

But I don't have any better idea right now and @Eldorias-Piano-Covers seems to agree with you (and was the one originally confused by the message). So I'll change the message and mark the translations as fuzzy for now.

Thank you for your input @zenseii @LeHerosInconnu @Eldorias-Piano-Covers :)

Eldorias-Piano-Covers:
Perfect, thanks ! (happy to discover this one ^^)

LeHerosInconnu:
Hello @Eldorias-Piano-Covers,

> For the future, I think it would be cool to be able to cast a spell with its spell scroll artefact, consireding those artefact as tiny magic books with one spell inside ;)

It could actually be interesting to play with the possibility to cast spells contained in artifacts (and thus also in spell scrolls) without the hero having a magic book.

This could be done with an option.
I suggest you to open a discussion (not an issue) on the subject, so that it is not forgotten.
Or if you want I can do it by myself. :)

Eldorias-Piano-Covers:
Hello, as I am not familiar with this, could you create the associated discussion ? :) Thanks !

LeHerosInconnu:
@Eldorias-Piano-Covers.

> Hello, as I am not familiar with this, could you create the associated discussion ? :) Thanks !

Done here: https://github.com/ihhub/fheroes2/discussions/5445. :)

PR #5442, opened 2022-05-22
Improved text if only one troop is selected.

### LINKED ISSUE #5358: Army bar: Don't show text about redistributing troops when there is only one unit in the troop
### Preliminary checks

### Platform

Other

### Describe the bug

If there is only one unit in a troop the message shouldn't say that you can right-click to redistribute since you can't.

It should only say `"Move the %{name}"`

image

This is found in two places in the army_bar.cpp for the castle garrison and hero army.

Also "the" should be added for the plural case as seen in the picture above.

### Save file

Just open any castle/hero army.

### Additional info

_No response_

felix642:
closes #5358


felix642:
Huh, I formatted the whole file by mistake >:(.
I need to revert this.

ihhub:
@felix642 , thank you so much for fixing this issue!

PR #5446, opened 2022-05-23
Unify code for monsters to avoid hardcoded monster types

ihhub:
- return messages during battle should be based on monster's abilities
- monster quality and strength calculation depends purely on abilities
- do not show random spells in battle only mode
- rename Stone spell into Petrify spell to be consistent with its description

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -218,7 +218,7 @@ Heroes::Heroes( int heroid, int rc )
         magic_point = 120;
 
         // all spell in magic book
-        for ( u32 spell = Spell::FIREBALL; spell < Spell::PETRIFY; ++spell )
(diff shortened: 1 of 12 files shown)


src/fheroes2/heroes/heroes.cpp
```diff
@@ -218,7 +218,7 @@ Heroes::Heroes( int heroid, int rc )
         magic_point = 120;
 
         // all spell in magic book
-        for ( u32 spell = Spell::FIREBALL; spell < Spell::STONE; ++spell )
+        for ( u32 spell = Spell::FIREBALL; spell < Spell::PETRIFY; ++spell )
```
oleg-derevenetz, line 221: @ihhub BTW `DEBUG_HERO` also doesn't need to have `RANDOM` spells in the spell book, so I suggest to replace the `Spell::PETRIFY` here by `Spell::RANDOM` as well.

PR #5454, opened 2022-05-24
Fix a few recent code smells

oleg-derevenetz:
* Copy constructor of the `HighscoreData`: add `const`;
* Remove `AI::Base::HeroesCanMove()` and `Heroes::MOVED` flag, because it was checked only in the `AI::Base::HeroesCanMove()`, which never has been called;
* Remove `AI::Base::HeroesGetTask()`;
* Make `AI::Normal::CastleTurn()` and `AI::Normal::HeroesTurn()` private and remove corresponding virtual functions from the `AI::Base`, because they never has been called from the outside and therefore they may be considered as an implementation detail.


ihhub:
@oleg-derevenetz , a bunch of thanks for cleaning up the code!

PR #5456, opened 2022-05-25
Add army swap and move armies in castle

### LINKED ISSUE #4084: Hotkeys for moving armies between heroes/castle garrison
### Preliminary checks

One of the annoying parts of the original game is the necessity to manually move each army unit bought in a castle or town to the hero army troops.  This requires a lot of clicking and makes the process rather boring.

UPDATE 03.01.25:
Tasks before completion:
- [x] Swap troops between guest hero and town garrison
- [ ] UI feature to move troops between hero and town garrison
- [ ] UI feature to swap troops between hero and town garrison

Original Post:

Newer games in the series (esp. Heroes IV and V) have introduced hotkeys (`space`, `up`, `down`) for solving this issue that we could also use.

I see the following possible setups and propose the following workflows:

1. A hero and a guardian in a castle:

  - Swap heroes and their armies (unchanged) using `space` key.
    This just adds a hotkey for the existing functionality.

  - Move all units from the visiting hero to the guardian using `up` key.
    The existing logic from the "meeting heroes" can be used.

  - Move all units from the guardian to the visiting hero using `down` key.
    Same logic can be used.

2. A hero visiting a castle (no active guardian):

  - Make hero a guardian of the castle using `space` key.
    Any existing troops in the castle will join hero's army, but the check needs to be made first that all troops will fit, and otherwise — the move should be rejected.
    This logic already exists, so just a hotkey needs to be added.

  - Move (almost) all units from the visiting hero to the castle using `up` key.
    The hero needs to keep at least one troop.  Logic exists.

  - Move all units from the castle to the visiting hero using `down` key.

3. There is only a guardian in the castle:

  - The guardian may leave the castle together with the army using `space` key.
    Just adding a hotkey here as well.

4. No heroes in the castle: no army can be moved.

### Additional info

_No response_

zenseii:
Close #4084 

This is not implemented/done yet:

- [ ] Mouse drag and drop input for moving armies in castle.
- [ ] Swap armies heroes meeting screen.
- [ ] Move hotkeys for heroes meeting screen.
- [ ] Fix move logic so that moved troops keep their positions in the army and so that similar troops are not unnecessarily merged.
- [ ] Swap armies with heroes as guardians option enabled.
- [ ] Move armies with heroes as guardians option enabled.

This is currently implemented:

- [ ] Swap in castle by pressing SPACE or by clicking the Crest icon.
- [ ] Move all troops by pressing up or down arrow in castle.


src/fheroes2/game/game_hotkeys.h
```diff
@@ -130,6 +132,7 @@ namespace Game
         NO_EVENT,
     };
 
+    // TODO: Implement HotKeyReleaseEvent.
```
zenseii, line 135:
```suggestion
    // TODO: Implement HotKeyReleaseEvent to avoid repeated presses of the hotkey when held down for a while
```

zenseii:
Hi, @ihhub!
This PR is ready for your review when you have the time and feel like it. The only part I wasn't able to do was replace use of `ArmyBar::SetArmy()` with `swapArmies()` and `MoveTroops()` in `Castle_Dialog.cpp`. I've tried to replace this old method but no matter what the troops will end up with the original hero when swapping between the guest and guard hero.

Aside from this the missing part now is the graphical side of it so you don't just move the cursor in the air with no real feedback. Unless this can be easily implemented, I think it might be more fit for another PR? For me either way is fine.

src/fheroes2/army/army.cpp
```diff
@@ -382,31 +382,133 @@ void Troops::JoinTroops( Troops & troops2 )
...
-        if ( troop && troop->isValid() ) {
-            ++validTroops;
+    // Check if there are troops to move.
+    if ( from.GetCount() == 0 || ( !moveAll && from.GetCount() == 1 && from.GetFirstValid()->GetCount() == 1 ) ) {
+        return;
+    }
+    // Modify the order in which to move the troops.
+    std::vector<Troop *> troopFromOrder;
+    uint32_t troopPointerIndex = 0;
+    if ( selectedTroopIndex != 4 ) {
```
ihhub, line 400: This logic needs an explanation. Why when the last stack is selected we cannot do the movement? A short comment would be satisfying.

src/fheroes2/army/army.cpp
```diff
@@ -382,31 +382,133 @@ void Troops::JoinTroops( Troops & troops2 )
...
+    // Will be changed later on if the receiving army's troops get merged. This avoids unnecessary merges.
+    bool preferEmptySlot = false;
+
+    // 'from' army might have more than 0 if it is a hero army with one troop left.
+    while ( from.GetCount() > 0 || ( !moveAll && from.GetCount() == 1 && from.GetFirstValid()->GetCount() == 1 ) ) {
+        // Step 1: Attempt to move troops directly to the same slot in the receiving army to preserve formation.
+        for ( uint32_t slot = 0; slot < ARMYMAXTROOPS; ++slot ) {
+            Troop * troop = troopFromOrder.at( slot );
+            // Make sure the selected troop is not moved before troops that could not be moved in Step 1.
+            if ( slot == 4 && GetCount() > 1 && isSelected ) {
```
ihhub, line 425: I would recommend to create a variable in anonymous namespace for this hardcoded value.

src/fheroes2/army/army.cpp
```diff
@@ -382,31 +382,133 @@ void Troops::JoinTroops( Troops & troops2 )
         }
 }
 
-void Troops::MoveTroops( const Troops & from )
+void Troops::MoveTroops( const Troops & from, const uint32_t selectedTroopIndex, const bool isSelected, const bool moveAll )
```
ihhub, line 385: The logic in this pull request is extremely complex to understand. What if we have 2 methods:
```cpp
void Troops::MoveSingleStackTroops( ... );
void Troops::MoveMultipleStackTroops( ... );
```

Will it solve the complexity of this code?

src/fheroes2/army/army.cpp
ihhub: Even better: what if we use `ARMYMAXTROOPS - 1` here?

src/fheroes2/army/army.h
```diff
@@ -82,9 +82,10 @@ class Troops : protected std::vector<Troop *>
     bool CanJoinTroops( const Troops & ) const;
 
     // Used only for moving full army in hero's meeting dialog.
-    void MoveTroops( const Troops & from );
+    void MoveTroops( const Troops & from, const uint32_t selectedTroopIndex = 4, const bool isSelected = false, const bool moveAll = false );
 
-    void MergeTroops();
+    // Defaults to 4 since the 5th troop is being merged into.
+    void MergeTroops( const uint32_t mergeNumber = 4 );
```
ihhub, line 88: Please avoid default values for parameters unless it is extremely necessary.

ihhub, line 88: Please avoid default values for parameters unless it is extremely necessary.

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -46,7 +46,8 @@ namespace
         WORLD_MAP,
         BATTLE,
         CASTLE,
-        MONSTER
+        MONSTER,
+        ARMY
```
ihhub, line 50: Monster group was designed exactly for moving army as well. I think we could keep the newly added hotkey in the same category.

src/fheroes2/heroes/heroes_meeting.cpp
```diff
@@ -400,14 +400,18 @@ void Heroes::MeetingDialog( Heroes & otherHero )
...
+        if ( le.MousePressLeft( moveArmyToHero2.area() ) || HotKeyHoldEvent( Game::HotKeyEvent::MOVE_RIGHT ) ) {
             moveArmyToHero2.drawOnPress();
             moveArmyToHero1.drawOnRelease();
         }
-        else if ( le.MousePressLeft( moveArmyToHero1.area() ) ) {
+        else if ( le.MousePressLeft( moveArmyToHero1.area() ) || HotKeyHoldEvent( Game::HotKeyEvent::MOVE_LEFT ) ) {
             moveArmyToHero1.drawOnPress();
             moveArmyToHero2.drawOnRelease();
         }
+        else if ( le.MousePressRight( moveArmyToHero2.area() ) || le.MousePressRight( moveArmyToHero1.area() ) || HotKeyHoldEvent( Game::HotKeyEvent::SWAP_ARMIES ) ) {
```
ihhub, line 411: Are we doing an action by right clicking on arrows? We shouldn't do this as it is not intuitive from user's perspective.

src/fheroes2/heroes/heroes_meeting.cpp
```diff
@@ -547,8 +551,24 @@ void Heroes::MeetingDialog( Heroes & otherHero )
 
             display.render();
         }
-        else if ( le.MouseClickLeft( moveArmyToHero2.area() ) ) {
-            otherHero.GetArmy().MoveTroops( GetArmy() );
+        else if ( le.MouseClickLeft( moveArmyToHero2.area() ) || HotKeyPressEvent( Game::HotKeyEvent::MOVE_RIGHT ) ) {
+            otherHero.GetArmy().MoveTroops( GetArmy().getTroops(), selectArmy1.isSelected() ? selectArmy1.GetSelectedIndex() : 4, selectArmy1.isSelected() );
```
ihhub, line 555: Please replace 4 by `ARMYMAXTROOPS - 1`.

src/fheroes2/heroes/heroes_meeting.cpp
```diff
@@ -547,8 +551,24 @@ void Heroes::MeetingDialog( Heroes & otherHero )
...
+            selectArmy1.Redraw();
+            selectArmy2.Redraw();
+
+            moraleIndicator1.Redraw();
+            moraleIndicator2.Redraw();
+
+            display.render();
+        }
+        else if ( le.MouseClickLeft( moveArmyToHero1.area() ) || HotKeyPressEvent( Game::HotKeyEvent::MOVE_LEFT ) ) {
+            GetArmy().MoveTroops( otherHero.GetArmy().getTroops(), selectArmy2.isSelected() ? selectArmy2.GetSelectedIndex() : 4, selectArmy2.isSelected() );
```
ihhub, line 571: Please replace 4 by `ARMYMAXTROOPS - 1`.

Also I would suggest to rename `ARMYMAXTROOPS` into `TROOP_STACKS_PER_ARMY` or something like this to be more understandable.

src/fheroes2/army/army.cpp
```diff
@@ -509,19 +617,27 @@ const Troop * Troops::GetSlowestTroop() const
     return *lowest;
 }
 
-void Troops::MergeTroops()
+void Troops::MergeTroops( const uint32_t mergeNumber )
```
ihhub, line 620: This is not entirely clear what `mergeNumber` stands for. Is it the number of first troop stacks from left to right?

src/fheroes2/army/army.cpp
```diff
@@ -509,19 +617,27 @@ const Troop * Troops::GetSlowestTroop() const
...
 {
+    // You are requesting to merge 0 or less troops.
+    assert( !( mergeNumber <= 0 ) );
```
ihhub, line 623: `assert( mergeNumber > 0 );` would be sufficient.

src/fheroes2/army/army.cpp
```diff
@@ -382,31 +382,133 @@ void Troops::JoinTroops( Troops & troops2 )
...
+
+void Army::swapArmies( Army & swapArmy )
+{
+    // Hero armies cannot end up with zero troops.
+    if ( GetCount() == 0 || swapArmy.GetCount() == 0 ) {
+        return;
+    }
+
+    // Either army can have 5 stacks so iterate through max possible.
+    for ( size_t slot = 0; slot < ARMYMAXTROOPS; ++slot ) {
```
ihhub, line 510: In some cases we use `size_t` while in others we utilize `uint32_t`. Please use the same variable type everywhere - `size_t` to be consistent in the code.

src/fheroes2/army/army.cpp
```diff
@@ -382,31 +382,133 @@ void Troops::JoinTroops( Troops & troops2 )
...
+        if ( troopCountPreMerge == GetCount() ) {
+            return;
+        }
+        preferEmptySlot = true;
+    }
+}
+
+void Army::swapArmies( Army & swapArmy )
+{
+    // Hero armies cannot end up with zero troops.
```
ihhub, line 504: Shall we try this approach
```cpp
std::swap( reinterpret_cast<Troops &>( *this ), reinterpret_cast<Troops &>( swapArmy ) );
```

Moveover, we could accept `Troops &` as a parameter so we can do this:
```cpp
std::swap( reinterpret_cast<Troops &>( *this ), swapArmy );
```

Going even further why do we need this method to be a member of this class since we are swapping armies of `Troops` class?
```cpp
void Troops::swapArmies( Troops & swapArmy )
{
    std::swap( *this, swapArmy );
}
```

ihhub (changes requested):
Hi @zenseii , I left several comments in this pull request. Could you please check them?

src/fheroes2/army/army.cpp
oleg-derevenetz: If it is decided after all to use cast, please don't use the `reinterpret_cast` unless absolutely necessary. `Army` is the descendant of `Troops` so `static_cast` will work [just fine](https://godbolt.org/z/Gv4W5ez9z).

ihhub: I think adding a method to Troops class is even better🙂

zenseii: I initially wanted to have it in troops as it would be more logical, but I ran into the problem that `GetArmy().getTroops()`, which is what will be called in all cases where swap is needed, returns `const Troops` and std::swap cannot work with consts, from what I understand.

src/fheroes2/army/army.cpp
zenseii: I see! I wasn't sure if `size_t` was part of the old code, but this is no problem.

src/fheroes2/army/army.cpp
zenseii: Yes, it did become quite complex. I've been thinking of ways to potentially simplify it. I'll look into if making two such functions according to stack number would remove some complexity.

src/fheroes2/army/army.cpp
zenseii: Yes, this is much better.

src/fheroes2/army/army.h
zenseii: The reason I added a default value is because MergeTroops is already used elsewhere in the code and in those cases all possible merges of troops are expected to happen when merging. I believe this part can be reworked to not have a default value. I would just have to change my approach a bit.

src/fheroes2/heroes/heroes_meeting.cpp
zenseii: I see! I will have to ask for some advice from @LeHerosInconnu and @Branikolog for how to swap troops in the heroes meeting screen using the mouse. The way I went with was right clicking on either the left or the right arrow, and this would make both arrows get pressed down.

src/fheroes2/army/army.cpp
oleg-derevenetz: Hi @zenseii

> but I ran into the problem that `GetArmy().getTroops()`, which is what will be called in all cases where swap is needed

It shouldn't really. `GetArmy().getTroops()` is intended to make object slicing explicit. Object slicing is when a copy is made of only a part of an object that is inherited from one of its ancestors - namely, when (explicitly or implicitly) a copy constructor or `operator=()` is called for one of the ancestors, and a descendant object is passed to it as an argument. For example:

```cpp
struct Control {};

struct Troops
{
    int f = 0;
};

struct Army: public Troops, public Control
{
    int f = 0; // this f is totally different from above
};

int main()
{
  Army a;

  // Slicing
  Troops t1 = a;
  
  // Not a slicing
  Troops & t2 = a;
}
```

Object slicing may be intentional and safe, but may be arbitrary and treacherous, this depends on situation. In you case, `swapArmies()` accepts a reference, `std::swap()` also accepts a reference, so no slicing is happens in your code. Slicing will be performed in the insides of the `std::swap()` due to calls of copy constructors and/or `operator=()` for `Troops` instead of `Army`, but I suppose that noone will complain about that.

[This is an example](https://godbolt.org/z/7bboh8YGd). As you can see, "troops" (`Troops::f`) of armies are swapped just fine, and Clang-Tidy doesn't complain.

src/fheroes2/heroes/heroes_meeting.cpp
LeHerosInconnu: Hello @zenseii and @ihhub,

> I see! I will have to ask for some advice from @LeHerosInconnu and @Branikolog for how to swap troops in the heroes meeting screen using the mouse. The way I went with was right clicking on either the left or the right arrow, and this would make both arrows get pressed down.
> 
> I really have no better ideas unless we decide to change the design of this screen but I'm not much in favor of that since it changes the original look.

We already had a discussion about this with @Branikolog here: https://github.com/ihhub/fheroes2/discussions/4327, and we managed to agree that adding an extra button in the same style as the other big yellow arrows, as there is enough space to do so, would be the best solution. :)

src/fheroes2/army/army.cpp
zenseii: I've addressed this, @ihhub , but I'll leave it to you to resolve this conversation if the comment is now satisfactory.

src/fheroes2/army/army.cpp
zenseii: Hi @oleg-derevenetz !

I've tried to implement the swap function like this now but to me it seems like the swap does not replace the previous troops and just adds them. You'll have to forgive me if this is very obvious, but right now it is like this for the castles in `castle_dialog.cpp Line 469`:

```
      GetArmy().swapArmies( heroes.Guest()->GetArmy() );
```

Here is an ingame example of what happens:
Hero army before swap:
image
Castle army before swap:
image

Hero army after swap:
image

Castle army after swap:
image


It goes without saying that opening either castle screen or hero screen after this makes the game crash.

Here is the message if I try to use the explicit `getTroops()`:
image

image

I'm sorry if this is all very obvious to you, but as a noob it's hard to see what I can do to address this.

Thank you for taking the time to look at this!

oleg-derevenetz: Hi @zenseii that's because `Troops` has the custom `operator=()` (and `std::swap()` just calls that operator):

https://github.com/ihhub/fheroes2/blob/63be73ef3018c2d6660d06a556803aa2117da3c7/src/fheroes2/army/army.cpp#L169-L175

This operator assumes that current state of underlying vector (`Troops` is inherited from the `std::vector`) is empty (there are no troops), which is obviously not the case in your case (and obviously that's definitely non-canonical behavior for this operator). That's the way this code has been written by the original team :)

zenseii: Aha that explains a lot... I never would have thought about not being able to trust the `=` operator, but there's still a lot to learn. I figure the natural course of action for me is to fix the behavior of this operator, since sooner or later that will have to be done.

Either that or just make up my own swap function.

src/fheroes2/army/army.cpp
```diff
@@ -168,9 +168,11 @@ Troops::Troops( const Troops & troops )
...
 {
-    reserve( rhs.size() );
-    for ( const_iterator it = rhs.begin(); it != rhs.end(); ++it )
-        push_back( new Troop( **it ) );
+    if ( this != &rhs ) {
+        reserve( rhs.size() );
+        for ( const_iterator it = rhs.begin(); it != rhs.end(); ++it )
+            push_back( new Troop( **it ) );
+    }
     return *this;
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @zenseii the removal of assignment operators and copy/move constructors is only done for objects that should not be copied by design. For example, instances of `Heroes` class should not be copied, because the set of heroes is somewhat static, they are created once in the heap at the game startup, and in the future the game operates only with pointers to them. Most of objects that are located in heap (i.e. created using `new`, `make_unique()`, etc) usually do not need to be copied, so a good habit is to remove their copy/move constructors and assignment operators so that they cannot be copied _accidentally by mistake_ (for example, by passing them to a function via value instead of pointer or reference). But, if you need to swap the _contents_ of an object with contents of another object of the same class using the `std::swap()`, this class still need to have at least a move constructor and move assignment operator.

The previous team implemented their own (non-canonical) set of methods for `Troops` which are not compatible with C++ STL. You can either use their home-grown set of methods or (in addition) properly implement the standard set of methods so they can be safely used with standard STL functions, such as `std::swap()`.

> and deleting this assignment operator shows that it is only actually used in two places

The first place is the copy constructor, which may be by itself used in a few more places.

PR #5460, opened 2022-05-25
Fix false highscore entry while cancelling to continue a campaign

### LINKED ISSUE #5269: Opening High Scores after exiting from campaign scenario selection opens name entry prompt dialog
### Preliminary checks

### Platform

Windows

### Describe the bug

If you exit from the scenario selection window of a campaign (with the awards and bonus etc) to the main menu, and then open up the high scores, you will get the dialog for name entry.

image

### Save file

Just load up any campaign's complete save file.
[Good_1_Complete.zip](https://github.com/ihhub/fheroes2/files/8476750/Good_1_Complete.zip)

### Additional info

_No response_

ihhub:
close #5269


oleg-derevenetz:
Hi @ihhub it seems that this solution doesn't work if I load save of any completed scenario (`XXX_Completed.savc`) - in this case if I cancel it after the appearance of the campaign scenario window and then open the High Scores screen, I'm still being prompted for a name to save the highscore.

oleg-derevenetz:
@ihhub I figured out what the problem is - I use the developer mode, and in this mode the fact of the end of the game was determined incorrectly (by checking the `world.CountDay()` instead of querying the `GameOver::Result`).

Also (unrelated to this issue with campaign scores) in developer mode the game result was never reset after displaying, therefore, when reopening the High Score screen after the win, you always get the same dialog window with your latest results and no High Scores.

I fixed both issues in the separate commit in this PR, please take a look when possible.

oleg-derevenetz (approved):

ihhub:
@oleg-derevenetz , thank you so much for fixing the debug mode case!

PR #5464, opened 2022-05-25
Fix code smells related to calls of the virtual functions from constructor or destructor

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks a lot for fixing these smells / warnings!

PR #5465, opened 2022-05-25
Remove types.h and remove (void) from function parameters for all relevant functions

oleg-derevenetz:
This is a comprehensive change, but I believe that everyone is already tired of replacing type aliases from `engine/types.h` with standard types and cleaning `(void)` from function parameters 😱

Also in this PR I fixed an issue in the `check_copyright_headers.py` when it tries to check the copyright headers of a removed file and also fixed multiple issues found by Clang-Tidy in the process of preparing this PR.


ihhub:
@oleg-derevenetz , thanks a lot for cleaning this mess!

PR #5467, opened 2022-05-26
Fix a few more code smells

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks a lot for constantly polishing the code!

PR #5468, opened 2022-05-26
Take into account AI hero's role during meeting

ihhub:
The higher AI hero's role the higher need to have a better army and set of artifacts.


oleg-derevenetz (changes requested):
Hi @ihhub I found just one typo in the comment, LGTM in general.

oleg-derevenetz (approved):

PR #5472, opened 2022-05-28
Migrate to C++17

oleg-derevenetz:
With all the benefits.

I have converted the VS2015 project files to VS2017 in case someone wants to build the game for WinXP.


ihhub:
I would suggest to move this pull request for the next release and merge it after the current one.

oleg-derevenetz:
Hi @ihhub

> I would suggest to move this pull request for the next release and merge it after the current one.

No problem, I'll move it to 0.9.17.

fheroes2-vs2017.vcxproj
```diff
@@ -1,577 +1,577 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
ihhub, line 1: Hi @oleg-derevenetz , this is interesting why the whole file is marked as changed but you modified only few line.

src/fheroes2/kingdom/luck.cpp
```diff
@@ -21,6 +21,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include <algorithm>
+
 #include "luck.h"
 #include "tools.h"
```
ihhub, line 27: We can remove this header as it was used for `clamp` function.

src/fheroes2/kingdom/morale.cpp
```diff
@@ -21,6 +21,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include <algorithm>
+
 #include "morale.h"
 #include "tools.h"
```
ihhub, line 27: We can remove this header as it was used for `clamp` function.

ihhub (changes requested):
Hi @oleg-derevenetz , I left few comments here. Could you please take a look? Also for files where we changed to `std::clamp` we might need to remove `tool.h` file inclusion.

fheroes2-vs2017.vcxproj
oleg-derevenetz: @ihhub That's because the previous version always included the `\r` character at the end of lines, even on Linux - `git` was unable to properly recode it for some reason. Now this should work properly - `\r\n` on Windows and `\n` on Linux.

ihhub: Oh, it was actually created by Visual Studio itself.

ihhub:
Hi @oleg-derevenetz , could you please resolve merge conflict here? I think it is because I added few files into the project.

fheroes2-vs2017.vcxproj
oleg-derevenetz: @ihhub I have no idea why `git` was unable to properly recode it - maybe it thinks that this was binary file. Nevertheless, this is just an XML and can be recoded just fine.

ihhub (approved):

src/fheroes2/system/save_format_version.h
```diff
@@ -25,6 +25,7 @@
 enum SaveFileFormat : uint16_t
 {
     // TODO: if you're adding a new version you must assign it to CURRENT_FORMAT_VERSION located at the bottom.
+    FORMAT_VERSION_0917_RELEASE = 10000,
```
ihhub, line 28: Mmm... since we decided to push it for 0.9.16 should we rename `FORMAT_VERSION_0916_RELEASE` into FORMAT_VERSION_PRE_0916_RELEASE` and then add `FORMAT_VERSION_0916_RELEASE`?

oleg-derevenetz: @ihhub Well yes, we can do this way then.

ihhub: Thank you!

fheroes2-vs2017.vcxproj
oleg-derevenetz: @ihhub probably `git` was unable to recode this file because these `vcxproj` files have Unicode BOM at start (however it's a bit strange because BOM by itself is not a reason to consider the file as binary and not recode between CRLF and LF AFAIK). I just added the BOM to the VS2017 project file just in case.

ihhub:
@oleg-derevenetz , well done! I hope we don't break anything :)

oleg-derevenetz:
Hi @ihhub

> I hope we don't break anything :)

It shouldn't be, but of course a ton of code smells have been added :) Mostly about `[[nodiscard]]` and code modernization - usage of structured bindings, init-statements, etc. Sorry about that :)

ihhub:
Nooooo... hehe🙂

ihhub:
@oleg-derevenetz , I deactivated both checks in SonarQube. For now they are polluting the results.

PR #5475, opened 2022-05-29
disable merging on every rebuild

undef21:
Source code should be unchanged after `make all; make clean`.
Currently .po files are changed on every rebuild. This PR fix it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,18 +77,7 @@ add_subdirectory(src)
 #
 if(NOT WIN32)
 	find_program(MAKE_EXECUTABLE make)
-	find_program(SED_EXECUTABLE sed)
-	find_program(XGETTEXT_EXECUTABLE xgettext)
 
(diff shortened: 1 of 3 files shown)


oleg-derevenetz (changes requested):
Hi @undef21

Changes in the Makefile are fine (except that this `for` loop will have issues with filenames containing spaces, but GNU Make [does not properly support](https://savannah.gnu.org/bugs/?712) such filenames anyway).

The `Translation update` workflow uses the explicit `make merge`, so there should be no issues with it as well.

The `CMakeLists.txt` in the root folder of the project, however, still assumes that `translations` target depends on `${PROJECT_NAME}.pot`, which is not the case anymore and can be fixed by removing the `DEPENDS ${PROJECT_NAME}.pot` from that target.

The `docs/TRANSLATION.md` still assumes that you just need to run the `make *.po clean` command to update the translatable strings. Since this command will not update `*.po` anymore, it should now be replaced by the `make merge` command.

oleg-derevenetz (approved):

ihhub:
@undef21 , great thanks for this improvement!

PR #5484, opened 2022-05-30
update debian packaging script

undef21:
In the source root dir:
```
ln -s script/packaging/debian
dpkg-buildpackage -us -uc -b
```


ihhub:
@undef21 , thank you so much for these improvements!

PR #5485, opened 2022-05-31
Properly deal with human-only colors in single player mode

### LINKED ISSUE #2578: Main menu, game allows to choose the color, which is not supposed to be human player
fheroes2:
image
The same map in OG:
image

oleg-derevenetz:
fix #2578


ihhub:
@oleg-derevenetz , great fix! Thank you so much for it!

PR #5486, opened 2022-05-31
Cleanup of Audio related code

ihhub:
- move `Audio::isValid()` checks to the appropriate places
- remove some invalid TODO comments
- remove useless audio related calls to slightly improve the performance

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -369,7 +369,7 @@ void Mixer::SetChannels( const int num )
 
     const int channelsCount = Mix_AllocateChannels( num );
     if ( num != channelsCount ) {
-        ERROR_LOG( "Failed to the required amount of channels for sound. Required " << num << " but received " << channelsCount )
+        ERROR_LOG( "Failed to allocate the required amount of channels for sound. The required number of channels " << num << " but allocated only " << channelsCount )
     }
 
     if ( channelsCount > 0 ) {
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -362,11 +362,6 @@ void Mixer::SetChannels( const int num )
         return;
     }
 
-    if ( !savedMixerVolumes.empty() ) {
-        // Why are you allocating channels again?
-        assert( 0 );
-    }
-
     const int channelsCount = Mix_AllocateChannels( num );
     if ( num != channelsCount ) {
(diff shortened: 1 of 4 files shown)


src/engine/audio.cpp
```diff
@@ -365,14 +362,22 @@ void Mixer::SetChannels( const int num )
         return;
     }
 
-    const size_t channelsCount = static_cast<size_t>( Mix_AllocateChannels( num ) );
+    if ( !savedMixerVolumes.empty() ) {
+        // Why are you allocating channels again?
+        assert( 0 );
+    }
```
oleg-derevenetz, line 368: This check looks a bit strange IMHO. First of all, it will not work if sound was never muted before. Also, it's a completely legitimate operation to change the number of channels in runtime (this is even specifically mentioned in the `Mix_AllocateChannels()` manual).

src/fheroes2/game/fheroes2.cpp
```diff
@@ -238,7 +238,8 @@ int main( int argc, char ** argv )
 
         if ( Audio::isValid() ) {
             Mixer::SetChannels( 32 );
-            Mixer::setVolume( -1, 100 * conf.SoundVolume() / 10 );
+            // Set the volume for all channels to 0. This is required to avoid random volume spikes at the beginning of the game.
+            Mixer::setVolume( -1, 0 );
```
oleg-derevenetz, line 242: Hmm, interesting, where do these random spikes come from? AFAIK we always set the volume before starting the playback on any channel, don't we?

oleg-derevenetz (commented):
Hi @ihhub I just left two questions, would you mind to take a look?

src/fheroes2/game/fheroes2.cpp
ihhub: Yes, we do. I observed it with ocean sounds, just a couple of time. It is better to prevent any possible noise.

ihhub: From another side I might be just wrong but we actually set volume **after** starting a playback:
```cpp
Play()
// here we have a gap in execution.
Pause()
setVolume()
Resume()
```

oleg-derevenetz (approved):

src/fheroes2/game/fheroes2.cpp
ihhub: Yes, with having volume as 0 and then setting volume to something above 0 should work.

PR #5491, opened 2022-06-01
Fix issues with Romanian

zenseii:
Relates to #4515 
This adds a Romanian PO, uses correct encoding in the makefile for MO compilation, removes Romanian from CP1250 font since that was wrong although they share some letters.


ihhub:

PR #5498, opened 2022-06-03
simplify lang/Makefile

undef21:


files/lang/Makefile
```diff
@@ -18,107 +18,42 @@
...
 # and transliterate the rest with default iconv rules
 %.mo: %.po
-	sed -e 'y/äëïöőüűÄËÏŐÖÜŰ/aeioouuAEIOOUU/' $< > $<.tmp
-	iconv -f utf-8 -t ascii//TRANSLIT $<.tmp > $<.2.tmp
-	msgfmt $<.2.tmp -o $@
+	sed -e 'y/äëïöőüűÄËÏŐÖÜŰ/aeioouuAEIOOUU/' $< | iconv -f utf-8 -t ascii//TRANSLIT | msgfmt - -o $@
 
 clean:
-	rm -f *.mo *.po~ *.tmp *.tmp~
+	rm -f *.mo
```
oleg-derevenetz, line 59: We should remove there `*.po~` as well, because `msgmerge` will create backup copies of the original `po` files on `make merge`.

files/lang/Makefile
```diff
@@ -18,107 +18,42 @@
 #   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             #
 ###########################################################################
 
+recode = sed -e '1,20 s/UTF-8/$(1)/' $< | iconv -f utf-8 -t $(1) | msgfmt - -o $@
```
oleg-derevenetz, line 21: Let's call it something like `MSGFMT` to follow the overall style of variable naming in our Makefiles (uppercase) and better reflect what this command is in fact doing (not just recode, isn't it?).

Also please add the newline after this line to visually separate variable definition from the subsequent target.

oleg-derevenetz (commented):
Hi @undef21 I just left a few suggestions, please take a look when you have time.

ihhub:
@undef21 , thank you so much for this improvement!

PR #5499, opened 2022-06-03
CMake updates (Mainly for Windows/MSYS2/MINGW) and translations in CMake

hgdagon:
Current issues:

Probably many, but the generated languages work in-game, so that's a plus.

Ninja generator convert converts the symbols in `sed` to some weird characters. ~~Don't know what to do about that yet (I think the was in `ENCODING` override in CMake, though, will need to look).~~ Using escape sequences, while ugly, seems to be the only workaround so far.

The MSVC generator also has an issue with the sed command, which I can reproduce in command prompt. Will probably need different character escapes. Working on it now.

~~The Romanian translation was added while I was working on this. It was mentioned in the Makefile, but the file itself was missing. This script currently doesn't support it. It will probably generate garbage or fail to build.~~ Romanian works fine now.


zenseii:
Hi @hgdagon! Thank you for looking into this. I see that you are making changes to the makefile in `files/lang`, are they compatible with this PR https://github.com/ihhub/fheroes2/pull/5498 by @undef21 ?

hgdagon:
@zenseii Oh, neat! No, it's not compatible, but I think this will be ongoing for some time. The checks will probably find many more issues. If/when that PR is approved, I can then make changes on my side.

files/lang/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+		string(REPLACE "UTF-8" "${charset}" PO_CONTENT "${PO_CONTENT}")
+	endif()
+	file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${PO} "${PO_CONTENT}")
+	execute_process(
+		COMMAND ${ICONV_EXE} -f UTF-8 -t ${charset} ${CMAKE_CURRENT_BINARY_DIR}/${PO}
+		OUTPUT_VARIABLE PO_CONTENT
+		ENCODING NONE
+		)
+	file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${PO} "${PO_CONTENT}")
+	# file(GENERATE OUTPUT ${PO} CONTENT "${PO_CONTENT}" NEWLINE_STYLE UNIX)
```
ihhub, line 76: Do we need this line?

src/fheroes2/game/game_io.cpp
```diff
@@ -311,7 +311,7 @@ bool Game::LoadSAV2FileInfo( const std::string & fn, Maps::FileInfo & finfo )
 
 std::string Game::GetSaveDir()
 {
-    return System::ConcatePath( System::ConcatePath( System::GetDataDirectory( "fheroes2" ), "files" ), "save" );
```
ihhub, line 314: What is the reason behind this change?

oleg-derevenetz: Please also note that this is a breaking change. Users literally will not be able to load their savefiles or their highscores after this change - they are being updated to the new version and all their files are gone. I doubt they'll like it.

hgdagon:
@ihhub I don't seem to have any more free time recently. I have since forgotten what I even changed here. Please go ahead and close this. My apologies for wasting everybody's time.

PR #5502, opened 2022-06-04
Do not auto-save at the end of a turn if the player has been vanquished on that turn

oleg-derevenetz:
Currently game may be auto-saved in certain cases - for example, when player dismissed his last hero by hand, however it will not be auto-saved if player attacks by the same last hero and loses the battle. This seems a bit unfair. With this PR game will not be auto-saved at the end of the turn in all cases when player loses/vanquished on that turn. Let's give him another chance.


ihhub:

PR #5504, opened 2022-06-05
Improve loop sound playback logic

ihhub:
The original approach is based on using last in an array elements. This leads to some "jumping" sounds during hero's walk. Such effect exists due to a fact that we were trying to replace angles and volumes by absolutely different values.

For example, we have existing sounds:
```
sound A, angle 0, volume 20
```

Then we want to play these sounds:
```
sound A, angle 0, volume 20
sound A, angle 80, volume 70
```
In this case **sound A, angle 0, volume 20** would be replaced by **sound A, angle 80, volume 70** which is absolutely incorrect.

The proposed approach contains logic to first detect sounds which are exact copies of existing played sounds and keep existing sounds without touching them. Then, find angles which are closest to each other and update the existing sound so the listener would hear a smallest change in sound effects.

This correction requires further testing but so far results are much better than before, especially for 3D audio.


src/fheroes2/audio/audio_manager.cpp
```diff
@@ -530,6 +535,36 @@ namespace
...
+        int bestAngleDiff = 360;
+        int bestVolumeDiff = 100;
+
+        for ( size_t soundToAddId = 0; soundToAddId < soundToAdd.size(); ++soundToAddId ) {
+            for ( size_t soundToReplaceId = 0; soundToReplaceId < soundToReplace.size(); ++soundToReplaceId ) {
+                const int angleDiff = std::abs( soundToAdd[soundToAddId].angle - soundToReplace[soundToReplaceId].angle );
+                const int volumeDiff
+                    = std::abs( static_cast<int>( soundToAdd[soundToAddId].volumePercentage ) - static_cast<int>( soundToReplace[soundToReplaceId].volumePercentage ) );
+                if ( bestAngleDiff > angleDiff ) {
+                    if ( bestAngleDiff == angleDiff && volumeDiff > bestVolumeDiff ) {
```
oleg-derevenetz, line 554: Hmm, if we are already here, then `bestAngleDiff` is already greater than `angleDiff`, so expression `bestAngleDiff == angleDiff` is always false.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -552,14 +587,27 @@ namespace
             return;
         }
 
+        // Sort all sounds by angle. This will help with search methods below.
```
oleg-derevenetz, line 590: Do you mean it will help with `std::find()` when you are searching for an existing sound? Maybe, if `std::find()` has some heuristics about sorted data, but maybe not. Maybe it worth to think about replacing the `std::vector` by a more suitable container? How about the `std::multimap`? It has `find` to find an exact match and `lower_bound`/`upper_bound` that may be used to find the "close enough" match.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -530,6 +535,36 @@ namespace
         DEBUG_LOG( DBG_ENGINE, DBG_TRACE, XMI::GetString( xmi ) )
     }
 
+    void getClosestSoundIdPairByAngle( const std::vector<AudioManager::AudioLoopEffectInfo> & soundToAdd, const std::vector<ChannelAudioLoopEffectInfo> & soundToReplace,
```
oleg-derevenetz, line 538: How about returning the `std::pair<size_t, size_t>` here, since `bestSoundToAddId` and `bestSoundToReplaceId` are rewritten anyway in any case?

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions, please take a look when you have time.

src/fheroes2/audio/audio_manager.cpp
ihhub: Oh, yes. It supposed to be `>=`

src/fheroes2/audio/audio_manager.cpp
ihhub: There are 2 options:
- do this in the current pull request to increase the amount of changes
- remove this sorting for now

I would suggest to remove and add a TODO comment for a possible optimization. What do you think?

src/fheroes2/audio/audio_manager.cpp
ihhub: Reading `std::pair` as a return value is not clear what these variables are.

src/fheroes2/audio/audio_manager.cpp
oleg-derevenetz: Yeah, I suppose leaving the `TODO` comment would be enough for now.

src/fheroes2/audio/audio_manager.cpp
zenseii: Wouldn't `>=` always be true anyways here due to already having checked that in the previous if condition.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -568,18 +605,38 @@ namespace
...
+
+            size_t effectsToReplaceCount = std::min( effectsToAdd.size(), effectsToReplace.size() );
+
+            while ( effectsToReplaceCount > 0 ) {
+                --effectsToReplaceCount;
 
-            size_t elementSize = std::min( effects.size(), existingEffects.size() );
-            while ( elementSize > 0 ) {
-                --elementSize;
+                // Find the most closest angles to those which are going to be added.
```
zenseii, line 629: Just a small correction.
```suggestion
                // Find the closest angles to those which are going to be added.
```

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -552,14 +587,16 @@ namespace
             return;
         }
 
+        // TODO: use another containers for sound effects to support more efficient sort and find operations based on the code below.
```
zenseii, line 590: Either the suggested change or `use other containers`.
```suggestion
        // TODO: use another container for sound effects to support more efficient sort and find operations based on the code below.
```

zenseii (commented):
Hi @ihhub! I've read through the code and found some typos, and I tested the changes and they do indeed improve alot when moving along a coastline.

src/fheroes2/audio/audio_manager.cpp
ihhub: Nope, `angleDiff` can be bigger than `bestAngleDiff`.

zenseii: I see! I thought the change was about the nested if statement 👍

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -530,6 +535,36 @@ namespace
...
+                const int angleDiff = std::abs( soundToAdd[soundToAddId].angle - soundToReplace[soundToReplaceId].angle );
+                const int volumeDiff
+                    = std::abs( static_cast<int>( soundToAdd[soundToAddId].volumePercentage ) - static_cast<int>( soundToReplace[soundToReplaceId].volumePercentage ) );
+                if ( bestAngleDiff >= angleDiff ) {
+                    if ( bestAngleDiff == angleDiff && volumeDiff > bestVolumeDiff ) {
+                        // The existing best pair has lower volume difference.
+                        continue;
+                    }
+
+                    bestAngleDiff = angleDiff;
```
oleg-derevenetz, line 559: I see the place where `bestAngleDiff` is updated, but I do not see a place where `bestVolumeDiff` is updated, it always remains `100`.

oleg-derevenetz (changes requested):
Hi @ihhub I found another suspicious place, please take a look to ensure that everything is by design.

src/fheroes2/audio/audio_manager.cpp
ihhub: Yes! This is a bug. I didn't notice this since all values were unique per angle.

ihhub: and I also should stop coding while almost sleeping...

ihhub:
Hi @oleg-derevenetz , the logic is correct according to design.

oleg-derevenetz (approved):

ihhub:
@oleg-derevenetz , thanks a lot for finding bugs in my code :)

Issue #5047, opened 2022-02-22
Pause and resume MIDI music during a turn


djk12587:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Music is one of the best parts of this game. The castle music is phenomenal, i'd like to hear more of the middle and ending parts of the castle music. Currently, i feel like i listen to the first 0-30 seconds over and over and over again.

### Describe the possible solution

Create an experimental feature that pauses and resumes castle music when switching between your castles during a single turn.

When switching between 2 or more towns of the same type, the music will continue. When you switch between different town types, the music will pause and resume.

Once your turn is over, all castle music will restart from the beginning.

### Additional info

- [x] MIDI music support
- [x] External music support

_No response_

Branikolog:
Hi, @djk12587 !
I believe your proposal is completely the same to the existing issue: #1068

djk12587:
@Branikolog it could be, but #1068 is kind of hard to understand and lacks detail.

ihhub:
We have 2 types of music to pause and resume: MIDI and external. I am creating tasks for each of them:
- [ ] MIDI music support
- [ ] External music support

ihhub:
Hi @oleg-derevenetz , do we have anything else to be done for this issue.

oleg-derevenetz:
Hi @ihhub currently this works absolutely the same on Windows and non-Windows OSes: MIDI music playback still cannot be resumed from an arbitrary position because of SDL Mixer limitations, other music types are resumed just fine, it does not depend on the OS.

PR #5505, opened 2022-06-05
Add music resume ability for MP3, OGG and FLAC formats

### LINKED ISSUE #5255: Replace SDL Mixer with more appropriate open source audio library
### Preliminary checks

We face many issues related to audio playback. For instance, the game cannot resume music where it was paused while switching between castles. An alternative solution is to implement another library as a replacement of SDL Mixer for most platforms.

One of the most promising candidates for this problem is [OpenAL](https://www.openal.org/) which is widely used for other games. We need to investigate all capabilities of this library before starting its integration. This is a list of requirements for any library to be used in the project:
- [ ] MIDI support
- [ ] WAV support
- [ ] OGG support
- [ ] FLAC support
- [ ] MP3 support
- [ ] multi-channel sound
- [ ] ability to start music playback from any place
- [ ] OS-independent
- [ ] GPL-2.0 or higher license

### Additional info

_No response_

ihhub:
Add Music resume ability for **MP3**, **OGG** and **FLAC** tracks. MIDI tracks are much more complex so I left them untouched but the code is ready for them as well.

relates to #5047 
close #5255 since the only concern not to use SDL was inability to restart tracks from random places.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -97,15 +97,18 @@ namespace
     // This mutex protects all operations with audio
     std::recursive_mutex audioMutex;
 
-    void FreeChannel( const int channel )
+    // This is the callback function set by Mix_ChannelFinished()
+    void freeChannel( const int channelId )
     {
-        Mix_Chunk * sample = Mix_GetChunk( channel );
+        const std::lock_guard<std::recursive_mutex> audioGuard( audioMutex );
+
+        Mix_Chunk * sample = Mix_GetChunk( channelId );
 
         if ( sample ) {
             Mix_FreeChunk( sample );
         }
 
-        audioEffectLedger.unregisterChannel( channel );
+        audioEffectLedger.unregisterChannel( channelId );
     }
 
     Mix_Chunk * LoadWAV( const uint8_t * ptr, const uint32_t size )
@@ -118,20 +121,20 @@ namespace
         return sample;
     }
 
-    int PlayChunk( Mix_Chunk * sample, const int channel, const bool loop )
+    int PlayChunk( Mix_Chunk * sample, const int channelId, const bool loop )
     {
         assert( sample != nullptr );
 
-        if ( channel >= 0 ) {
-            audioEffectLedger.unregisterChannel( channel );
+        if ( channelId >= 0 ) {
+            audioEffectLedger.unregisterChannel( channelId );
         }
 
-        const int channelId = Mix_PlayChannel( channel, sample, loop ? -1 : 0 );
-        if ( channelId < 0 ) {
-            ERROR_LOG( "Failed to play an audio chunk for channel " << channelId << ". The error: " << Mix_GetError() )
+        const int channel = Mix_PlayChannel( channelId, sample, loop ? -1 : 0 );
+        if ( channel < 0 ) {
+            ERROR_LOG( "Failed to play an audio chunk for channel " << channel << ". The error: " << Mix_GetError() )
         }
 
-        return channelId;
+        return channel;
     }
 
     void addSoundEffect( const int channelId, const int16_t angle, uint8_t volumePercentage )
@@ -148,20 +151,20 @@ namespace
         }
     }
 
-    int PlayChunkFromDistance( Mix_Chunk * sample, const int channel, const bool loop, const int16_t angle, uint8_t volumePercentage )
+    int PlayChunkFromDistance( Mix_Chunk * sample, const int channelId, const bool loop, const int16_t angle, uint8_t volumePercentage )
     {
         assert( sample != nullptr );
 
-        const int channelId = Mix_PlayChannel( channel, sample, loop ? -1 : 0 );
+        const int channel = Mix_PlayChannel( channelId, sample, loop ? -1 : 0 );
 
-        if ( channelId < 0 ) {
-            ERROR_LOG( "Failed to play a mix chunk for original channel " << channel << ". The error: " << Mix_GetError() )
-            return channelId;
+        if ( channel < 0 ) {
+            ERROR_LOG( "Failed to play an audio chunk for channel " << channel << ". The error: " << Mix_GetError() )
+            return channel;
         }
 
-        addSoundEffect( channelId, angle, volumePercentage );
+        addSoundEffect( channel, angle, volumePercentage );
 
-        return channelId;
+        return channel;
     }
 
     struct MusicInfo
@@ -247,6 +250,7 @@ namespace
 
     void PlayMusic( const uint64_t musicUID, const Music::PlaybackMode playbackMode );
 
+    // This is the callback function set by Mix_HookMusicFinished()
     void replayCurrentMusic()
     {
         const std::lock_guard<std::mutex> musicLooperGuard( musicLooperMutex );
@@ -443,13 +447,16 @@ void Audio::Init()
         isInitialized = true;
     }
 
-    // This function holds its own mutex, and we can't call it while holding our own mutexes
+    // These functions hold their own mutexes under the hood while waiting for the corresponding callback
+    // function to finish, so we can't call them while holding our own mutexes because of a possible deadlock
     Mix_HookMusicFinished( replayCurrentMusic );
+    Mix_ChannelFinished( freeChannel );
 }
 
 void Audio::Quit()
 {
-    // This function holds its own mutex, and we can't call it while holding our own mutexes
+    // This function holds its own mutex under the hood while waiting for the corresponding callback function
+    // to finish, so we can't call it while holding our own mutexes because of a possible deadlock
     Mix_HookMusicFinished( nullptr );
 
     const std::lock_guard<std::mutex> musicLooperGuard( musicLooperMutex );
@@ -576,6 +583,7 @@ int Mixer::Play( const uint8_t * ptr, const uint32_t size, const int channelId, 
     }
 
     const std::lock_guard<std::recursive_mutex> audioGuard( audioMutex );
+
     if ( !isInitialized ) {
         return -1;
     }
@@ -585,7 +593,6 @@ int Mixer::Play( const uint8_t * ptr, const uint32_t size, const int channelId, 
         return -1;
     }
 
-    Mix_ChannelFinished( FreeChannel );
     return PlayChunk( sample, channelId, loop );
 }
 
@@ -598,6 +605,7 @@ int Mixer::PlayFromDistance( const uint8_t * ptr, const uint32_t size, const int
     }
 
     const std::lock_guard<std::recursive_mutex> audioGuard( audioMutex );
+
     if ( !isInitialized ) {
         return -1;
     }
@@ -607,13 +615,13 @@ int Mixer::PlayFromDistance( const uint8_t * ptr, const uint32_t size, const int
         return -1;
     }
 
-    Mix_ChannelFinished( FreeChannel );
     return PlayChunkFromDistance( sample, channelId, loop, angle, volumePercentage );
 }
 
 int Mixer::applySoundEffect( const int channelId, const int16_t angle, uint8_t volumePercentage )
 {
     const std::lock_guard<std::recursive_mutex> audioGuard( audioMutex );
+
     if ( !isInitialized ) {
         return -1;
     }
@@ -623,7 +631,7 @@ int Mixer::applySoundEffect( const int channelId, const int16_t angle, uint8_t v
     return channelId;
 }
 
-int Mixer::setVolume( const int channel, const int volumePercentage )
+int Mixer::setVolume( const int channelId, const int volumePercentage )
 {
     const int volume = normalizeToSDLVolume( volumePercentage );
 
@@ -634,10 +642,10 @@ int Mixer::setVolume( const int channel, const int volumePercentage )
     }
 
     if ( !isMuted ) {
-        return normalizeFromSDLVolume( Mix_Volume( channel, volume ) );
+        return normalizeFromSDLVolume( Mix_Volume( channelId, volume ) );
     }
 
-    if ( channel < 0 ) {
+    if ( channelId < 0 ) {
         if ( savedMixerVolumes.empty() ) {
             return 0;
         }
@@ -649,7 +657,7 @@ int Mixer::setVolume( const int channel, const int volumePercentage )
         return normalizeFromSDLVolume( prevVolume );
     }
 
-    const size_t channelNum = static_cast<size_t>( channel );
+    const size_t channelNum = static_cast<size_t>( channelId );
 
     if ( channelNum >= savedMixerVolumes.size() ) {
         return 0;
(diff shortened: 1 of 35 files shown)


ihhub:
Hi @oleg-derevenetz , can we please disable `[[nodiscard]]` check for clang-tidy for now?

oleg-derevenetz:
Hi @ihhub

> can we please disable `[[nodiscard]]` check for clang-tidy for now?

Yes, I believe so.

src/engine/thread.cpp
```diff
@@ -0,0 +1,84 @@
...
+            std::unique_lock<std::mutex> mutexLock( manager->_mutex );
+            manager->_workerNotification.wait( mutexLock, [manager] { return manager->_runFlag == 1; } );
+            mutexLock.unlock();
+
+            if ( manager->_exitFlag )
+                break;
+
+            manager->_mutex.lock();
+
+            manager->doStuff();
```
oleg-derevenetz, line 75: All this behavior with calling the `unlock()` in the separate function looks error-prone of course. Maybe separate the `doStuff()` in two distinct functions and call the first of them in a separate block with `std::lock_guard` (BTW this can be done in lines 66-68 as well) or at least add the TODO comment to think about this later?

oleg-derevenetz (commented):
Hi @ihhub I just left a few comments, please take a look when you have time.

oleg-derevenetz:
> now we are keeping all music tracks in memory, do we need to do this?

If we have enough memory, then why not.

> should all songs in the game be resumed while played?

No. For example, `MUS::BATTLEWIN`, `MUS::BATTLELOSE`, music during battle, `MUS::PUZZLE`, objects music (such as `MUS::TREEHOUSE`, `MUS::EXPERIENCE`, etc, etc) should be played from scratch every time.

> should we use cached music tracks before even checking their presence in a storage?

Why not. If user removed the file from storage while the music is cached, then it's the user's problem.

src/engine/audio.cpp
```diff
@@ -164,26 +161,177 @@ namespace
...
+            return iter->second;
+        }
+
+        void update( const uint64_t musicUID, Mix_Music * mix, const double position )
+        {
+            assert( mix != nullptr );
+
+            auto iter = _musicCache.find( musicUID );
+            if ( iter == _musicCache.end() ) {
+                _musicCache.try_emplace( musicUID, MusicInfo{ mix, position } );
```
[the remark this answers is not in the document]
ihhub: The compiler is unhappy about this so I added `MusicInfo`

src/engine/thread.cpp
ihhub: I fixed it.

ihhub:
Hi @Branikolog , could you please help to test this pull request?

src/fheroes2/audio/audio_manager.h
```diff
@@ -61,6 +61,9 @@ namespace AudioManager
 
     void playLoopSounds( std::map<M82::SoundType, std::vector<AudioLoopEffectInfo>> soundEffects, bool asyncronizedCall );
     void PlaySound( int m82, bool asyncronizedCall = false );
-    void PlayMusic( int mus, bool loop = true, bool asyncronizedCall = false );
+
+    void PlayMusic( const int trackId, const bool loop, bool rewindToStart = false );
```
oleg-derevenetz, line 65: Hi @ihhub how about making the `rewindToStart = true` for non-async variant (that is used primarily for dialog/object sounds) and `rewindToStart = false` for the async variant (`PlayMusicAsync`), since it is used for background music? Currently a lot of sounds that should always start from the beginning are not covered - for example, here:

https://github.com/ihhub/fheroes2/blob/7f21ff123dbd547d1348326ad83fb5610776de7e/src/fheroes2/battle/battle_dialogs.cpp#L419-L434

and here:

https://github.com/ihhub/fheroes2/blob/7f21ff123dbd547d1348326ad83fb5610776de7e/src/fheroes2/heroes/heroes_action.cpp#L259

and here:

https://github.com/ihhub/fheroes2/blob/7f21ff123dbd547d1348326ad83fb5610776de7e/src/fheroes2/heroes/heroes_action.cpp#L959

etc, etc. Or maybe somehow link this up to the `loop`, because I don't know a single case when non-looped music should be resumed (but I know the reverse examples - for example, the music in battle).

ihhub: Hi @oleg-derevenetz , all music track which are not looped **always** start from the beginning.

ihhub: Two ways we can do:
- add a comment explaining the cases
- add an enumeration instead of these 2 variables

To be safe it's good to go with enumeration. What do you think?

oleg-derevenetz: @ihhub I think that enum is obviously better, because this reduces the number of arguments and outlines "incompatible" variations.

ihhub: It is done.

src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -418,19 +418,19 @@ bool Battle::Arena::DialogBattleSummary( const Result & res, const std::vector<A
...
     }
     else if ( army1->GetControl() & CONTROL_HUMAN ) {
         GetSummaryParams( res.army1, res.army2, army1->GetCommander(), res.exp1, sequence, title, msg );
-        AudioManager::PlayMusic( MUS::BATTLELOSE, false );
+        AudioManager::PlayMusic( MUS::BATTLELOSE, AudioManager::MusicPlaybackMode::REWIND_AND_PLAY_INFINITE );
     }
     else if ( army2->GetControl() & CONTROL_HUMAN ) {
         GetSummaryParams( res.army2, res.army1, army2->GetCommander(), res.exp2, sequence, title, msg );
-        AudioManager::PlayMusic( MUS::BATTLELOSE, false );
+        AudioManager::PlayMusic( MUS::BATTLELOSE, AudioManager::MusicPlaybackMode::REWIND_AND_PLAY_INFINITE );
```
oleg-derevenetz, line 433: Hi @ihhub We need to use `PLAY_ONCE` here and in other places above, because first `false` meant `loop = false` (unless you deliberately want to change it).

ihhub: Oh, nice catch!

src/engine/audio.h
```diff
@@ -63,8 +63,16 @@ namespace Mixer
 
 namespace Music
 {
-    void Play( const std::vector<uint8_t> & v, const bool loop );
-    void Play( const std::string & file, const bool loop );
+    // Music UID is used to cache existing songs. It is caller's responsibility to generate them.
+
+    // This function return true in case of music track for corresponding Music UID is cached.
+    bool Play( const uint64_t musicUID, const bool loop, const bool rewindToStart );
```
oleg-derevenetz, line 69: @ihhub I propose to use the same approach with `enum` here as well, because `loop = false` is incompatible with `rewindToStart = false`.

ihhub: Eh... Sure, I will move the enumeration here.

src/engine/audio.cpp
```diff
@@ -164,26 +161,182 @@ namespace
...
+            if ( iter == _musicCache.end() ) {
+                return {};
+            }
+
+            return iter->second;
+        }
+
+        void update( const uint64_t musicUID, Mix_Music * mix, const double position )
+        {
+            assert( mix != nullptr );
```
zenseii, line 186: Hi @ihhub!
I just built from the latest commit and whenever I try to load up a scenario or exit the game this assertion fails:
`Assertion failed: mix != nullptr, file D:\Documents\GitHub\fheroes2-zenseii\src\engine\audio.cpp, line 186`

I have the sound files in flac and they're named `homm2_01.flac` up to 42.

ihhub: Hi @zenseii , can you please attach a call stack to see where it is called? We have few places where this method can be called.

ihhub:
Somehow the code doesn't work for SDL 2...

zenseii (approved):

src/engine/audio.cpp
```diff
@@ -164,26 +161,187 @@ namespace
...
+        if ( isResumeSupported && musicInfo.position > 1 ) {
+            // Set music position only when at least 1 second of the track has been played.
+            Mix_RewindMusic();
+            if ( Mix_SetMusicPosition( musicInfo.position ) == -1 ) {
+                ERROR_LOG( "The codec does not support music resuming from a custom place." )
+            }
+        }
+
+        musicSettings.currentTrack = musicInfo;
+        musicSettings.currentTrackUID = musicUID;
```
oleg-derevenetz, line 344: Suppose that currently played track is ended and thread that should run the `replayCurrentMusic()` is currently waiting/sleeping at line 247. Suppose that `currentTrackUID` is now, say, 5, as well as `asyncTrackUID`. While that thread is sleeping, this function is called by the main thread or by another auxiliary thread to switch to track with id, say, 10. It sets the `asyncTrackUID` to 10 at line 315 of this function, then starts the playback of track 10, then sets the `currentTrackUID` to 10 here and finishes. And then, say, after 5 seconds (because why not) that thread with `replayCurrentMusic()` is woken up and continued execution. It will call this function again and will rewind the track 10 to the start, and will not able to detect that track has been already changed.

Need to think how to better implement the looping to avoid situations like this...

oleg-derevenetz (changes requested):
Hi @ihhub I expressed my concerns regarding possible issues with music looping implementation. I'll try to think regarding possible solutions...

src/engine/audio.cpp
oleg-derevenetz: Hi @ihhub I just added the commit that should fix situations like this, please take a look when you have time. If you wish, you are free to revert it.

oleg-derevenetz:
What's interesting, `Mix_HookMusicFinished()` has its own mutex (what is not mentioned anywhere in the documentation), which is held at least while the handler is executing and which can cause a deadlock in the situation like this:

1. `Music::Stop()` is called, it acquires the `audioMutex`;
2. Hook handler is started by another thread, it waits on `audioMutex`;
3. `Music::Stop()` calls the `Mix_HookMusicFinished( nullptr )`, it tries to acquire its internal mutex, but it is not able to, because hook handler is currently waiting on the `audioMutex`;
4. Deadlock.

oleg-derevenetz:
BTW, we have the same issues currently with the `Mix_ChannelFinished()` hook: it can be executed from the background thread, it calls SDL_Mixer functions, and it now messing with `audioEffectLedger` without holding the `audioMutex`.

oleg-derevenetz:
Hi @ihhub it's probably good for your review.

ihhub:
> Hi @ihhub just one more thing (C) lieutenant Columbo: currently there is an issue with sound from streams (at first I thought that that's my fault, but this is observed on the `master` branch too). How to reproduce:
> 
>     1. Start any map with streams nearby, for example, "Broken Alliance";
> 
>     2. Ensure that stream sound is loud and clear (you may even set the hero right on the stream);
> 
>     3. Open Castle screen for any castle and then exit from it;
> 
>     4. Select the hero standing on stream. You will not be able to hear stream sound anymore - any sounds (sawmill, magic shrine, campfire) will remain just fine except that one. Even if you move the hero over stream, sound of the stream won't be back.

This is extremely weird but I think I know why.

src/engine/audio.cpp
```diff
@@ -209,84 +371,105 @@ namespace
...
-
-    audioSpecs.freq = frequency;
-    audioSpecs.format = format;
-    audioSpecs.channels = static_cast<uint8_t>( channels );
-
-    isInitialized = true;
+    // These functions hold their own mutexes under the hood while waiting for the corresponding callback
+    // function to finish, so we can't call them while holding our own mutexes because of a possible deadlock
+    Mix_HookMusicFinished( replayCurrentMusic );
+    Mix_ChannelFinished( freeChannel );
```
oleg-derevenetz, line 453: Since we don't hold a lock here, in theory if there is a some background thread trying to call, say, `Mixer::Play()` or `Music::Play()` during the `Audio::Init()` execution, there may be a sort of race and these callbacks may in theory not be called, if `Mixer::Play()` or `Music::Play()` will finish first and the audio track will be short enough. But in practice this is highly unlikely, because when `Audio::Init()` is called, there should not be such threads yet.

src/engine/audio.cpp
```diff
@@ -148,42 +151,201 @@ namespace
...
+
+            musicLooperThread->join();
+            musicLooperThread.reset();
+        }
+
+        const std::lock_guard<std::recursive_mutex> audioGuard( audioMutex );
+
+        if ( !isInitialized ) {
+            return;
+        }
```
ihhub, line 269: In SDL2 this function will be called every time we stop current music playback. Then we create a new thread where we check whether the current track is okay to play. I feel it is some sort of overkill by doing this. I think we need to add at least these checks before creating a thread:
```cpp
// The current track managed to change during the start of this thread
if ( currentTrackChangeCounter != musicSettings.currentTrackChangeCounter ) {
    return;
}

// All other cases should be handled by the Mix_PlayMisic() itself
if ( musicSettings.currentTrackPlaybackMode != Music::PlaybackMode::CONTINUE_TO_PLAY_INFINITE ) {
    return;
}
```

@oleg-derevenetz , what do you think?

ihhub: I edited the first post.

ihhub:
Hi @oleg-derevenetz , besides this change I don't have any more remarks. I will hold it a bit and review again tomorrow morning to see if we missed something.

ihhub:
@oleg-derevenetz , I could say more that I didn't expect audio related changes to be so huge when I had started cleaning it up. Anyway, we are almost there to complete this chapter of code changes.

src/engine/audio.cpp
ihhub: Yes, it is safe to so such in our codebase.

oleg-derevenetz:
BTW wile testing all this music stuff, I noticed another issue - sound effect volume is not changing "on the fly" anymore while increasing/decreasing its value in the adventure map's settings window. They actually changed only when I move the hero.

ihhub:
> BTW wile testing all this music stuff, I noticed another issue - sound effect volume is not changing "on the fly" anymore while increasing/decreasing its value in the adventure map's settings window. They actually changed only when I move the hero. I suppose this is somehow related to the new implementation of the `EnvironmentSoundMixer`.

I know why. I will fix it in a separate pull request.

ihhub:
Hi @oleg-derevenetz , I've checked the code and looks like it is good to go unless we missed something.

oleg-derevenetz:
Hi @ihhub

> unless we missed something.

Yeah, this is encouraging :)

oleg-derevenetz (approved):

ihhub:
@oleg-derevenetz , thank a lot for your help in this feature!

oleg-derevenetz:
@ihhub

> thank a lot for your help in this feature!

Hope we are not missed anything :)

ihhub:
> Hi @ihhub just one more thing (C) lieutenant Columbo: currently there is an issue with sound from streams (at first I thought that that's my fault, but this is observed on the `master` branch too). How to reproduce:
> 
>     1. Start any map with streams nearby, for example, "Broken Alliance";
> 
>     2. Ensure that stream sound is loud and clear (you may even set the hero right on the stream);
> 
>     3. Open Castle screen for any castle and then exit from it;
> 
>     4. Select the hero standing on stream. You will not be able to hear stream sound anymore - any sounds (sawmill, magic shrine, campfire) will remain just fine except that one. Even if you move the hero over stream, sound of the stream won't be back.

Hi @oleg-derevenetz , do you still observe this issue on the latest commit?

oleg-derevenetz:
@ihhub

>  do you still observe this issue on the latest commit?

Yes, just tested this on master branch. The same issue with streams (and only streams).

ihhub:
> @ihhub
> fheroes2.engine.version_.0.9.15.2022-06-08.18-41-58.mp4
> 
> Turn the sound on while watching.

This issue appears only on SDL 2.

Issue #5249, opened 2022-04-10
Give option to make room for Alliance creatures


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

If you have a creature alliance bonus like Dwarven Alliance in the Good Campaign and you don't have room for dwarves, they will simply let you pass and disappear. This is how OG works but fheroes2 lets you choose and potentially dismiss creatures for neutral armies that want to join:
![image](https://user-images.githubusercontent.com/12501091/162645198-3459e268-d0da-418e-aa24-6b5030f966a2.png)

### Describe the possible solution

I suggest implementing an option in the dialog for alliance creatures. There would be some need for reworking of the logic and order of these dialogs.

Like this maybe:
1. They offer to join and you get to choose yes or no and to look at your army
2a. You deny and the message of letting you pass shows.
2b. You accept and they join.

This way you can avoid having them if you don't want them to begin with.

LeHerosInconnu:
@zenseii.

> I suggest implementing an option in the dialog for alliance creatures. There would be some need for reworking of the logic and order of these dialogs.
> 
> Like this maybe:
> 
> 1. They offer to join and you get to choose yes or no and to look at your army
>    2a. You deny and the message of letting you pass shows.
>    2b. You accept and they join.
> 
> This way you can avoid having them if you don't want them to begin with.

Yes, I think a text like this could be added to the current text:

' "My lord, we would also be honored to join your army."
Do you accept?'

In fheroes2:

![Wandering join fh2 02](https://user-images.githubusercontent.com/43583994/116571725-a00ef000-a90b-11eb-9032-8f99231461ec.png)

Similar discussion: https://github.com/ihhub/fheroes2/discussions/3309.

zenseii:
> Could we use common text for joining here? ![image](https://user-images.githubusercontent.com/55348946/163337701-860016ad-81ef-4204-bff9-a9dc15f18536.png) "The dwarves recognize their allies and gladly `want` to join your forces. "Do you accept?" In a case player refuses there would be nice to show the message, which is already implemented (or something similar, I'm concerning about "hail" here). ![image](https://user-images.githubusercontent.com/55348946/163339100-fc43b282-27fa-4c6e-8162-354683b862d9.png)

I agree with the points above. You could remove the part before the quotation marks maybe, not sure if that would look weird.

[...]

PR #5507, opened 2022-06-06
adding hero button to wagon and other recruit objects

shprotru:
fheroes2 brings enhancement to dismiss troops while someone is recruiting.

before:
Снимок экрана от 2022-06-06 10-59-34

after:
Снимок экрана от 2022-06-06 13-16-16

### DIFF
--- a/src/fheroes2/dialog/dialog.h
+++ b/src/fheroes2/dialog/dialog.h
@@ -99,7 +99,7 @@ namespace Dialog
     void DwellingInfo( const Monster &, uint32_t available );
     bool SetGuardian( Heroes &, Troop &, CapturedObject &, bool readonly );
     int ArmyInfo( const Troop & troop, int flags, bool isReflected = false );
-    int ArmyJoinFree( const Troop &, Heroes & );
+    int ArmyJoinFree( std::string, std::string, const Troop &, Heroes & );
     int ArmyJoinWithCost( const Troop &, uint32_t join, uint32_t gold, Heroes & );
     int ArmySplitTroop( uint32_t freeSlots, const uint32_t redistributeMax, uint32_t & redistributeCount, bool & useFastSplit );
     void Marketplace( Kingdom & kingdom, bool fromTradingPost );
--- a/src/fheroes2/dialog/dialog_armyinfo.cpp
+++ b/src/fheroes2/dialog/dialog_armyinfo.cpp
@@ -610,7 +610,7 @@ void DrawMonster( fheroes2::RandomMonsterAnimation & monsterAnimation, const Tro
         monsterAnimation.increment();
 }
 
-int Dialog::ArmyJoinFree( const Troop & troop, Heroes & hero )
+int Dialog::ArmyJoinFree( std::string dialog_title, std::string message, const Troop & troop, Heroes & hero )
 {
     fheroes2::Display & display = fheroes2::Display::instance();
     const bool isEvilInterface = Settings::Get().ExtGameEvilInterface();
@@ -618,10 +618,7 @@ int Dialog::ArmyJoinFree( const Troop & troop, Heroes & hero )
     // setup cursor
     const CursorRestorer cursorRestorer( true, Cursor::POINTER );
 
-    const Text title( _( "Followers" ), Font::YELLOW_BIG );
-
-    std::string message = _( "A group of %{monster} with a desire for greater glory wish to join you.\nDo you accept?" );
-    StringReplace( message, "%{monster}", Translation::StringLower( troop.GetMultiName() ) );
+    const Text title( dialog_title, Font::YELLOW_BIG );
 
     TextBox textbox( message, Font::BIG, BOXAREA_WIDTH );
     const int buttons = Dialog::YES | Dialog::NO;
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -612,7 +612,11 @@ void ActionToMonster( Heroes & hero, int32_t dst_index )
     else if ( join.reason == NeutralMonsterJoiningCondition::Reason::Free ) {
         DEBUG_LOG( DBG_GAME, DBG_INFO, hero.GetName() << " join monster " << troop.GetName() )
 
-        if ( Dialog::YES == Dialog::ArmyJoinFree( troop, hero ) ) {
+        std::string title = _( "Followers" );
+        std::string message = _( "A group of %{monster} with a desire for greater glory wish to join you.\nDo you accept?" );
+        StringReplace( message, "%{monster}", Translation::StringLower( troop.GetMultiName() ) );
+
+        if ( Dialog::YES == Dialog::ArmyJoinFree( title, message, troop, hero ) ) {
             hero.GetArmy().JoinTroop( troop );
 
             I.GetStatusWindow().SetRedraw();
@@ -2303,7 +2307,7 @@ void ActionToDwellingRecruitMonster( Heroes & hero, const MP2::MapObjectType obj
 
     if ( !troop.isValid() )
         Dialog::Message( title, msg_void, Font::BIG, Dialog::OK );
-    else if ( Dialog::YES == Dialog::Message( title, msg_full, Font::BIG, Dialog::YES | Dialog::NO ) )
+    else if ( Dialog::YES == Dialog::ArmyJoinFree(title, msg_full, troop, hero) )
         RecruitMonsterFromTile( hero, tile, title, troop, false );
 
     hero.SetVisited( dst_index, Visit::GLOBAL );


src/fheroes2/dialog/dialog_armyinfo.cpp
```diff
@@ -610,18 +610,15 @@ void DrawMonster( fheroes2::RandomMonsterAnimation & monsterAnimation, const Tro
...
     const bool isEvilInterface = Settings::Get().ExtGameEvilInterface();
 
     // setup cursor
     const CursorRestorer cursorRestorer( true, Cursor::POINTER );
 
-    const Text title( _( "Followers" ), Font::YELLOW_BIG );
-
-    std::string message = _( "A group of %{monster} with a desire for greater glory wish to join you.\nDo you accept?" );
-    StringReplace( message, "%{monster}", Translation::StringLower( troop.GetMultiName() ) );
+    const Text title( dialog_title, Font::YELLOW_BIG );
```
zenseii, line 621: Hi @shprotru !
Since we are touching this part of the code, could we replace the use of `Text` with the newer `fheroes2::Text` found in `ui_text.h`?

shprotru: I make attempt to use `fheroes2::Text`, but don't know how to solve build problem:
```
../fheroes2/gui/skill_bar.cpp:289:14: error: no matching constructor for initialization of 'fheroes2::Text'
        Text text( Skill::Primary::String( Skill::Primary::ATTACK ), Font::SMALL );
             ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../fheroes2/gui/ui_text.h:125:9: note: candidate constructor not viable: no known conversion from 'Font::(anonymous enum at ../fheroes2/gui/text.h:35:5)' to 'const fheroes2::FontType' for 2nd argument
        Text( std::string && text, const FontType fontType );
        ^
../fheroes2/gui/ui_text.h:124:9: note: candidate constructor not viable: no known conversion from 'Font::(anonymous enum at ../fheroes2/gui/text.h:35:5)' to 'const fheroes2::FontType' for 2nd argument
        Text( const std::string & text, const FontType fontType );
        ^
../fheroes2/gui/ui_text.h:126:9: note: candidate constructor not viable: requires single argument 'text', but 2 arguments were provided
        Text( const Text & text ) = default;
        ^
../fheroes2/gui/ui_text.h:127:9: note: candidate constructor not viable: requires single argument 'text', but 2 arguments were provided
        Text( Text && text ) = default;
        ^
../fheroes2/gui/ui_text.h:123:9: note: candidate constructor not viable: requires 0 arguments, but 2 were provided
        Text() = default;
```

src/fheroes2/dialog/dialog.h
```diff
@@ -99,7 +100,7 @@ namespace Dialog
     void DwellingInfo( const Monster &, uint32_t available );
     bool SetGuardian( Heroes &, Troop &, CapturedObject &, bool readonly );
     int ArmyInfo( const Troop & troop, int flags, bool isReflected = false );
-    int ArmyJoinFree( const Troop &, Heroes & );
+    int ArmyJoinFree( const fheroes2::Text &, const fheroes2::Text &, const Troop &, Heroes & );
```
zenseii, line 103: Hi @shprotru!
We do not have to change from `std::string` to `fheroes2::Text` since `fheroes2::Text` can take `std::string` just fine. Also to keep in line with current projects standards we should put the same identifier in these parameters in the declaration as in the definition. I also suggest to change the name of `dialog_title` to something like `dialogTitle` to follow the coding standards. Like this:
```cpp
int ArmyJoinFree( const std::string & dialog_title, const std::string & message, const Troop & troop, Heroes & hero );
```

This way we avoid having to include `ui_text.h` in the header file `dialog.h` in which we usually want to have as few inclusions as possible. 

Don't hesitate to ask if something is unclear :)

src/fheroes2/dialog/dialog_armyinfo.cpp
```diff
@@ -610,20 +610,17 @@ void DrawMonster( fheroes2::RandomMonsterAnimation & monsterAnimation, const Tro
...
+int Dialog::ArmyJoinFree( const fheroes2::Text & dialog_title, const fheroes2::Text & message, const Troop & troop, Heroes & hero )
 {
     fheroes2::Display & display = fheroes2::Display::instance();
     const bool isEvilInterface = Settings::Get().ExtGameEvilInterface();
 
     // setup cursor
     const CursorRestorer cursorRestorer( true, Cursor::POINTER );
 
-    const Text title( _( "Followers" ), Font::YELLOW_BIG );
+    const Text title( dialog_title.text(), Font::YELLOW_BIG );
```
zenseii, line 621: I did a bad job explaining but this is where I wanted to use `fheroes2::Text` instead of the old `Text` class. It would be something like this:
```suggestion
    const fheroes2::Text title( dialogTitle, fheroes2::FontType::normalYellow() );
```

zenseii:
Hi @shprotru!

The warnings have to do with the fact that you need to use the `fheroes2::Text` specific functions/methods now that you are using that instead of `class Text`.

Take this example from the warnings:
```cpp
text.Blit( pos.x + 320 - text.w() / 2, pos.y + 64 );
```

`fheroes2::Text` doesn't have a `Blit` member function so you need to call its respective `draw` function. In addition that function takes the additional parameter `Image &`. You'll also notice that `fheroes2::Text` has its width function named as `width` instead of just `w`. The same goes for `h` and `height`. Last but not least, the new `fheroes2::Text` shifts the text up by 2 pixels compared to the old `Text` which means we need to move down two pixels and therefore add 2 pixels in the Y position parameter. In the above example this gives us this change:

 ```cpp
text.draw( pos.x + 320 - text.width() / 2, pos.y + 66, display );
```

File: src/fheroes2/heroes/heroes_action.cpp [hunk 4]
```diff
@@ -612,7 +612,12 @@ void ActionToMonster( Heroes & hero, int32_t dst_index )
     else if ( join.reason == NeutralMonsterJoiningCondition::Reason::Free ) {
         DEBUG_LOG( DBG_GAME, DBG_INFO, hero.GetName() << " join monster " << troop.GetName() )
 
-        if ( Dialog::YES == Dialog::ArmyJoinFree( troop, hero ) ) {
+        const fheroes2::Text header( _( "Followers" ), fheroes2::FontType::normalYellow() );
```
zenseii, line 615: I'm not sure I understand why we have to move these string declarations from `Dialog::ArmyJoinFree` to here? Is there any instance where this header and body will be different? If not I suggest to just keep them where they were before.

zenseii: Ah wait I see now. You can disregard this comment.

zenseii (changes requested):
Hi @shprotru !

I just tested this and would it be possible to add this for dwellings that give you creatures for free too? Like the Peasant hut, fearie trees etc. ?
image


Also we have a slightly similar issue regarding creatures that join you due to alliances like dwarves and dragons. This would require a bit more work though since some extra dialog logic needs to be added if you read the discussion in that issue #5249 .

If you want to go ahead and add that too but need some clarification then don't hesitate to ask either me or the other members that contributed to that issue's discussion.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2214,7 +2219,7 @@ void ActionToDwellingJoinMonster( Heroes & hero, const MP2::MapObjectType object
 
         AudioManager::PlaySound( M82::EXPERNCE );
 
-        if ( Dialog::YES == Dialog::Message( title, message, Font::BIG, Dialog::YES | Dialog::NO ) ) {
+        if ( Dialog::YES == Dialog::ArmyJoinFree( title, message, troop, hero ) ) {
             if ( !hero.GetArmy().CanJoinTroop( troop ) )
                 Dialog::Message( troop.GetName(), _( "You are unable to recruit at this time, your ranks are full." ), Font::BIG, Dialog::OK );
```
zenseii, line 2224: Right now there is no way to make this message about ranks being full appear since you cannot press "Yes":
image.

I suggest to remove this message and make the else statement further down the default action for this dialog. What do you think?

zenseii (changes requested):
Hi @shprotru ! I left one comment. Please take a look when you should have the time 😊

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2219,7 +2219,7 @@ void ActionToDwellingJoinMonster( Heroes & hero, const MP2::MapObjectType object
 
         AudioManager::PlaySound( M82::EXPERNCE );
 
-        if ( Dialog::YES == Dialog::ArmyJoinFree( title, message, troop, hero ) ) {
+        if ( Dialog::YES == Dialog::Message( title, message, Font::BIG, Dialog::YES | Dialog::NO ) ) {
```
zenseii, line 2222: I didn't ask if you could revert this change, but the logic needs to reflect that the message will not appear anymore.
The code I'm asking for would be something like this.
```cpp
if ( Dialog::YES == Dialog::ArmyJoinFree( title, message, troop, hero ) ) {
                tile.MonsterSetCount( 0 );
                hero.GetArmy().JoinTroop( troop );
                hero.MovePointsScaleFixed();
                Interface::Basic::Get().GetStatusWindow().SetRedraw();
}
```

What do you think?

ihhub:
Hi @shprotru , before proceeding with the review could you please explain why we need this feature if a player can easily hit a Space button one more time? I think we have similar discussion [here](https://github.com/ihhub/fheroes2/discussions/5508).

Enhancements are good but we need to be cautious about them and not overuse them. I would say that in cases when a hero **cannot** repeat the same action they might help to improve user experience.

Let's do this way: we listen the opinion of @oleg-derevenetz , @Branikolog , @oleg-derevenetz , @zenseii and @idshibanov about this enhancement and then decide what we want to do.

**Remember**: look at this enhancement with closed eyes on similar buttons in other dialogs and question whether it is useful to have it.

zenseii:
Hi @ everyone in this PR 😅

I've been thinking a bit about this PR after @oleg-derevenetz's point that he made in another issue about adding the marketplace button. To summarize I believe we should only introduce such buttons to the dialogs if these dialogs cannot be exited and reentered again. Existing examples of these are level up, creature joining/"followers"  etc.

That's why when it comes to the dialogs covered by this PR, which are the dwelling dialogs, I do not think we need them since the player can simply exit and reenter with the space bar after having made necessary changes (or just make them before entering).

oleg-derevenetz:
> Let's do this way: we listen the opinion of @oleg-derevenetz , @Branikolog , @oleg-derevenetz , @zenseii and @idshibanov about this enhancement and then decide what we want to do.

I believe that if the same action can be repeated later using spacebar, then there is no need to add a shortcut for it - it just clutters up the interface.

Branikolog:
Hi, everyone.

In my personal point of view, we can add here this button, since this window 99% is the same to common joining window:
image
We can modify text here, as @zenseii proposed.
We can use a part of text from the next window, appearing after player accepts. Something like: "Your ranks are full, so you cannot join new troops"` and add it under the question.
So, the next window (`You are unable to recruit at this time, your ranks are full.`) will never appear anymore.

AAS:
Guys, could you please just merge it as is? 
It is 3+ years old now :(

PR #5518, opened 2022-06-08
Minor fixes for the music continuation feature

oleg-derevenetz:
Follow-up to #5505

* If `playbackMode` is `CONTINUE_TO_PLAY_INFINITE`, but track type doesn't support the playback continuation, convert playbackMode to the `REWIND_AND_PLAY_INFINITE` just in case - to avoid the temptation of `replayCurrentMusic()` to restart it even if the track type is not restartable. Currently it shouldn't happen though, but `replayCurrentMusic()` doesn't check the track type, just the playback mode.

* Remove music fadeout support: first of all, it is unused, and also it deadlocks with the current approach by the following scheme:

1. `Music::Stop()` acquires `audioMutex`;
2. `Music::Stop()` calls `Mix_FadeOutMusic()`, which eventually (when the music is finally stopped) leads to the call of the `replayCurrentMusic()` from the different thread (thread started by the `SDL_Mixer` for the music playback);
3. `replayCurrentMusic()` waits on the `audioMutex` in a different thread;
4. Either `Mix_FadeOutMusic()` or `Mix_PlayingMusic()` are not able to return until callback is not finished;
5. Deadlock!

* If music, which was started in the `PLAY_ONCE` mode, finishes without being forcibly stopped, it should be able to start the same music track right away. Before this PR this wasn't the case.

* It should be able to play the same track using the `RESUME_AND_PLAY_INFINITE` -> any other mode -> `RESUME_AND_PLAY_INFINITE` sequence of modes. Before this PR the track was restarted from an improper position during the second playback in the `RESUME_AND_PLAY_INFINITE` mode.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -292,7 +292,8 @@ namespace
 
         // We are here because a track in the PLAY_ONCE mode has finished playing
         if ( musicSettings.currentTrackPlaybackMode == Music::PlaybackMode::PLAY_ONCE ) {
-            // Since this track has finished playing, we will not be able to resume it with all our desire
+            // We cannot resume this track, but in future we should be able to play it in the
+            // RESUME_AND_PLAY_INFINITE mode if necessary, so reset its position to zero
             musicSettings.currentTrack.position = 0;
             musicSettings.trackManager.update( musicSettings.currentTrackUID, musicSettings.currentTrack );
             musicSettings.resetCurrentTrack();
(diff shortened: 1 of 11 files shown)


ihhub:
@oleg-derevenetz , great thanks for these fixes!

oleg-derevenetz:
Hi @ihhub I am aware that it's time for a release, but could we wait with this a bit? The latest changes should be tested properly.

ihhub:
@oleg-derevenetz , we will release this Saturday then.

PR #5519, opened 2022-06-08
Reset the interface when changing the UI theme, just redraw it is not enough

oleg-derevenetz:
Regression after #5490

I removed this `Reset()` call in that PR and it turns out that it was in vain.


ihhub:

PR #5522, opened 2022-06-10
Rework the music resume logic, replace std::lock_guard by std::scoped_lock

oleg-derevenetz:
I believe that this is much more stable (and also a bit simpler I think) implementation, current implementation in `master` branch is  deadlock-prone because of acquiring audio mutexes in `SDL_Mixer` callbacks. Implementation from this PR doesn't acquire any mutexes in callbacks and, as far as I can tell, works well and is much more suitable for the new release - we don't want our users to suffer. But it still needs to be extensively tested, if possible.

Also I heeded to persistent recommendations from SonarCloud and replaced `std::lock_guard` by `std::scoped_lock` from C++17. Honestly, if there is no need to safely acquire multiple locks at the same time, I'd prefer old `std::lock_guard`, because due to the variadic nature of `scoped_lock`'s constructor you can easily make a mistake and write like this:

```cpp
std::scoped_lock lock;
```

which is no-op. `lock_guard` will not allow this.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -258,8 +258,8 @@ namespace
 
     void playMusic( const uint64_t musicUID, Music::PlaybackMode playbackMode )
     {
-        // Unregister this callback to ensure that it will not be called while we are
-        // modifying the current track information
+        // Unregister this callback to ensure that it will not be called while we are modifying the
+        // current track information
         Mix_HookMusicFinished( nullptr );
 
         MusicInfo musicInfo = musicSettings.trackManager.getMusicInfoByUID( musicUID );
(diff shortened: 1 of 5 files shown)


src/engine/audio.cpp
```diff
@@ -551,6 +450,9 @@ void Audio::Quit()
     Music::Stop();
     Mixer::Stop();
 
+    Mix_ChannelFinished( nullptr );
+    Mix_HookMusicFinished( nullptr );
```
ihhub, line 454: We do this call within `Music::Stop()` function. I think we don't need it.

ihhub (changes requested):
Hi @oleg-derevenetz , I left one comment here. Could you please take a look at it?

src/engine/audio.cpp
oleg-derevenetz: Ah, yes, really. I'll remove it.

ihhub (approved):

ihhub:
@oleg-derevenetz , great work! Thank you!

PR #5525, opened 2022-06-11
Nintendo Switch: protect the logfile with mutex

oleg-derevenetz:
While it is safe to concurrently [call](https://www.man7.org/linux/man-pages/man3/syslog.3.html#ATTRIBUTES) `syslog()` and [access](https://en.cppreference.com/w/cpp/io/cerr) `std::cout`/`std::cerr` from multiple threads simultaneously, it is still not safe to access Switch-specific `logFile` object this way.


ihhub:

PR #5526, opened 2022-06-11
Rework settings and introduce 3D audio UI option

ihhub:
All Audio settings are moved to a separate window because we cannot have more than 9 options in a single window. Since we replaced 3 options by 1 I added Interface type and Battle Resolve options into Main Menu Settings and Cursor Type and HotKeys into Adventure System Options.
image
image


zenseii (commented):
Hi @ihhub!

I've currently only found an issue which might not even be related to this but if you turn off volume for music in external mode and then switch to midi the sound will not reactivate for external or just play midi music. 

https://user-images.githubusercontent.com/12501091/173192395-95cc9752-384b-415e-9923-ed2fa62aa6da.mp4

I can open a separate issue for this if necessary.

oleg-derevenetz (approved):

zenseii (approved):

PR #5527, opened 2022-06-11
Multiple fixes for the AsyncManager, make it ready for use to restart the music in the RESUME_AND_PLAY_INFINITE mode

oleg-derevenetz:
The current approach with detached thread to restart the music is simple and works well enough, but it has one drawback - there is a possibility (although minimal) that eventually the detached thread will outlive the `main()`, which may lead to various issues, because when `main()` is exited, all global objects will be destroyed, and only then, during the process termination, other threads will be terminated, that is, there is always a possibility for a detached thread to access already destroyed object(s). BTW, `AsyncSoundManager` from the `AudioManager` module could suffer from this too, because its worker thread can outlive `main()` as well and can access global objects in other modules that are already destroyed while `g_asyncSoundManager` global object is not destroyed yet - since order of destruction of global objects in different modules is indeterminate. This is not good, we need to perform the proper cleanup. I adapted the `AsyncManager` for this with several changes, because in its current form it doesn't work reliable enough due to various race issues in it. See details below.

### DIFF
--- a/src/engine/thread.cpp
+++ b/src/engine/thread.cpp
@@ -25,6 +25,20 @@
 
 namespace MultiThreading
 {
+    void AsyncManager::createWorker()
+    {
+        if ( !_worker ) {
+            _runFlag = true;
+            _worker = std::make_unique<std::thread>( AsyncManager::_workerThread, this );
+
+            {
+                std::unique_lock<std::mutex> lock( _mutex );
+
+                _masterNotification.wait( lock, [this] { return !_runFlag; } );
+            }
+        }
+    }
+
     void AsyncManager::stopWorker()
     {
         if ( _worker ) {
@@ -42,18 +56,11 @@ namespace MultiThreading
         }
     }
 
-    void AsyncManager::createWorker()
+    void AsyncManager::notifyWorker()
     {
-        if ( !_worker ) {
-            _runFlag = true;
-            _worker = std::make_unique<std::thread>( AsyncManager::_workerThread, this );
+        _runFlag = true;
 
-            {
-                std::unique_lock<std::mutex> lock( _mutex );
-
-                _masterNotification.wait( lock, [this] { return !_runFlag; } );
-            }
-        }
+        _workerNotification.notify_all();
     }
 
     void AsyncManager::_workerThread( AsyncManager * manager )
@@ -90,12 +97,5 @@ namespace MultiThreading
 
             manager->executeTask();
         }
-    }
-
-    void AsyncManager::notifyWorker()
-    {
-        _runFlag = true;
-
-        _workerNotification.notify_all();
     }
 }
--- a/src/engine/thread.cpp
+++ b/src/engine/thread.cpp
@@ -25,30 +25,34 @@
 
 namespace MultiThreading
 {
-    AsyncManager::~AsyncManager()
+    void AsyncManager::stopWorker()
     {
         if ( _worker ) {
             {
                 std::scoped_lock<std::mutex> lock( _mutex );
 
-                _exitFlag = 1;
-                _runFlag = 1;
-                _workerNotification.notify_all();
+                _exitFlag = true;
+                _runFlag = true;
             }
+
+            _workerNotification.notify_all();
 
             _worker->join();
             _worker.reset();
         }
     }
 
-    void AsyncManager::createThreadIfNeeded()
+    void AsyncManager::createWorker()
     {
         if ( !_worker ) {
-            _runFlag = 1;
+            _runFlag = true;
             _worker = std::make_unique<std::thread>( AsyncManager::_workerThread, this );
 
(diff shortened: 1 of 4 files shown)


src/engine/thread.h
```diff
@@ -59,8 +61,8 @@ namespace MultiThreading
         std::condition_variable _masterNotification;
         std::condition_variable _workerNotification;
 
-        uint8_t _exitFlag{ 0 };
-        uint8_t _runFlag{ 1 };
+        std::atomic<bool> _exitFlag{ false };
```
oleg-derevenetz, line 64: This flag should be atomic, because `_workerThread` doesn't hold any locks when checking it.

src/engine/thread.h
```diff
@@ -45,6 +46,7 @@ namespace MultiThreading
 
         void createThreadIfNeeded();
 
+        // The _mutex should be acquired while calling this method
```
oleg-derevenetz, line 49: This method changes `_runFlag` which should be protected by the `_mutex`.

src/engine/thread.cpp
```diff
@@ -25,30 +25,34 @@
...
                 std::scoped_lock<std::mutex> lock( _mutex );
 
-                _exitFlag = 1;
-                _runFlag = 1;
-                _workerNotification.notify_all();
+                _exitFlag = true;
+                _runFlag = true;
             }
 
+            _workerNotification.notify_all();
```
oleg-derevenetz, line 38: There is no need to hold a lock while calling any of the `notify*()` methods of `std::condition_variable` here and in other similar places.

src/engine/thread.h
```diff
@@ -59,8 +64,8 @@ namespace MultiThreading
         std::condition_variable _masterNotification;
         std::condition_variable _workerNotification;
 
-        uint8_t _exitFlag{ 0 };
-        uint8_t _runFlag{ 1 };
+        std::atomic<bool> _exitFlag{ false };
```
oleg-derevenetz, line 67: This flag should be atomic, because `_workerThread` doesn't hold any locks when checking it.

src/engine/thread.h
```diff
@@ -31,26 +32,30 @@ namespace MultiThreading
...
 
-        virtual ~AsyncManager();
+        virtual ~AsyncManager() = default;
 
         AsyncManager & operator=( const AsyncManager & ) = delete;
-        AsyncManager & operator=( AsyncManager && ) = delete;
+
+        // Stop and join the worker thread. This cannot be done in the destructor (directly or indirectly) due
+        // to the potential race on the vptr since this class has virtual methods.
+        void stopWorker();
```
oleg-derevenetz, line 43: We can't reliably join the worker thread in the destructor because this class has virtual methods that could be called by the worker thread. This leads to so-called [vptr race](https://github.com/google/sanitizers/wiki/ThreadSanitizerPopularDataRaces#data-race-on-vptr) (which I personally observed during testing today), so we need to do this in the separate method.

src/engine/thread.h
```diff
@@ -31,26 +32,30 @@ namespace MultiThreading
...
     protected:
         std::mutex _mutex;
 
-        void createThreadIfNeeded();
+        // Create the worker thread if it doesn't exist yet.
+        void createWorker();
 
-        void notifyThread();
+        // Notify the worker thread about a new task. The _mutex should be acquired while calling this method.
+        void notifyWorker();
```
oleg-derevenetz, line 52: This method modifies the `_runFlag` while not holding a lock by itself, so we need to acquire the `_mutex` beforehand.

ihhub:
Hi @oleg-derevenetz , is this pull request ready for review?

oleg-derevenetz:
> Hi @oleg-derevenetz , is this pull request ready for review?

Hi @ihhub yes, I believe so.

src/engine/thread.h
```diff
@@ -31,26 +32,32 @@ namespace MultiThreading
     {
     public:
         AsyncManager() = default;
-
         AsyncManager( const AsyncManager & ) = delete;
-        AsyncManager( AsyncManager && ) = delete;
 
-        virtual ~AsyncManager();
+        virtual ~AsyncManager() = default;
```
ihhub, line 37: I recommend to at least assert that `_worker` is empty in the destructor.

ihhub (changes requested):
Hi @oleg-derevenetz , I left one suggestion here. Would you mind to take a look at it?

src/engine/thread.h
oleg-derevenetz: @ihhub If joinable `std::thread` is still running while being destructed, `std::terminate()` will be called. There is no way for joinable thread to outlive this destructor.

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks again for polishing this code!

PR #5531, opened 2022-06-12
Fix the inter-thread race with SDL timer in Interface::StatusWindow

oleg-derevenetz:
Regularly reported by the ThreadSanitizer. Fortunately, as far as I can see, there are no other places where the `fheroes2::Timer` is currently used.

### DIFF
--- a/src/engine/timing.cpp
+++ b/src/engine/timing.cpp
@@ -18,11 +18,12 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-#include "timing.h"
-
 #include <thread>
 
 #include <SDL.h>
+
+#include "timing.h"
+#include "logging.h"
 
 namespace fheroes2
 {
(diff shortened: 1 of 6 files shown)


src/engine/timing.cpp
```diff
@@ -40,12 +41,16 @@ namespace fheroes2
             remove();
 
             id = SDL_AddTimer( interval, fn, param );
+            if ( id == 0 ) {
```
ihhub, line 44: Hi @oleg-derevenetz ,
I was actually thinking to replace SDL timer with a proper C++ implementation as there is no solid reason to use it with SDL. This will also reduce the dependency from SDL. What do you think?

oleg-derevenetz: Hi @ihhub If you want to do it, I won't mind at all of course. This is the only place where async timer is used, and it can be replaced just by measuring the time intervals in the event loop.

ihhub: Let's just replace it without SDL and extra threads.

oleg-derevenetz: We can just remove it really, because for event loops `fheroes2::TimeDelay` is more than enough.

ihhub:

PR #5534, opened 2022-06-12
appdata.xml should be updated like this

Laserlicht:
Otherwise, the stores may not recognize the new version cleanly.

https://flathub.org/apps/details/io.github.ihhub.Fheroes2

Flathub still shows the old version in the info, although the new one is actually uploaded.

### DIFF
--- a/script/packaging/common/fheroes2.appdata.xml
+++ b/script/packaging/common/fheroes2.appdata.xml
@@ -16,8 +16,11 @@
     <content_attribute id="violence-fantasy">mild</content_attribute>
   </content_rating>
   <releases>
-    <release date="2022-05-07" version="v0.9.16">
+    <release date="2022-06-12" version="v0.9.16">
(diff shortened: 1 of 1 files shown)


Laserlicht:
It would even be possible to insert the change log there as well.

Then it will also be displayed in the stores (like Flathub).

https://github.com/flathub/com.heroicgameslauncher.hgl/blob/cf13653857c2fdab8749bc5b2c73e5725391c65a/com.heroicgameslauncher.hgl.metainfo.xml#L195

https://flathub.org/apps/details/com.heroicgameslauncher.hgl

ihhub:
@Laserlicht , thanks for the correction! As of now we do not add any change logs anywhere else except changelog.txt as it is hard to maintain too many sources.

Laserlicht:
Thank you. 😁

Yes, I understand. All good. 😊

PR #5535, opened 2022-06-13
Fix a delay in game while crossing a terrain and playing MIDI music

### LINKED ISSUE #1979: Adventure map, change of terrain type by the hero, there is a delay for the music when the hero changes terrain (in MIDI)
**Edit 2021/02/21.**
**Edit 2022/06/07.**

In the original game, there is no delay for the music when the hero changes terrain and the music resumes where it left off (in MIDI):

https://user-images.githubusercontent.com/43583994/108638297-cc7f3980-748e-11eb-852c-01e49283dfd6.mp4

In fheroes2, there is a delay for the music when the hero changes terrain (in MIDI)(and the music starts over from the beginning (will be corrected with another issue)):

https://user-images.githubusercontent.com/43583994/108638312-e587ea80-748e-11eb-8e48-7e2121ebb528.mp4

In fheroes2:

Hero change terrain fh2 01

Move the hero Roxana between the two types of terrain.

Here is the save file:
[Hero change terrain 01.zip](https://github.com/ihhub/fheroes2/files/5365185/Hero.change.terrain.01.zip)

ihhub:
The original issue comes from the fact that SDL needs some time to process MIDI track before playing it. At the same time we were trying to get the number of allocated mixer channels but we had to wait as the same mutex were used by another thread to play the track.

The solution is to just cache the number of mixer channels and never call SDL function to get the same info. In this case we don't need to hold any mutex lock and it doesn't block environmental sound processing.

close #1979

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -565,7 +565,8 @@ void Mixer::SetChannels( const int num )
 
     audioChannelCount = Mix_AllocateChannels( num );
     if ( num != audioChannelCount ) {
-        ERROR_LOG( "Failed to allocate the required amount of channels for sound. The required number of channels " << num << " but allocated only " << audioChannelCount )
+        ERROR_LOG( "Failed to allocate the required amount of channels for sound. The required number of channels " << num << " but allocated only "
+                                                                                                                    << audioChannelCount )
     }
 
     if ( audioChannelCount > 0 ) {
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -551,12 +551,12 @@ bool Audio::isValid()
     return isInitialized;
 }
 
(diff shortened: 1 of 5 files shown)


src/engine/audio.cpp
```diff
@@ -496,6 +498,8 @@ void Audio::Quit()
         Mix_Quit();
 #endif
 
+        audioChannelCount = 0;
+
```
oleg-derevenetz, line 502: Hi @ihhub The `Mix_OpenAudio()` allocates `MIX_CHANNELS` channels on call, so in the `Audio::Init()` we probably need to either assign `MIX_CHANNELS` to `audioChannelCount` or assign the result of `Mix_AllocateChannels( -1 )` to it - to not out of sync.

oleg-derevenetz (approved):

PR #5539, opened 2022-06-14
Don't waste time playing the pre-battle sound if the game sounds are turned off

### LINKED ISSUE #5536: Combat screen, starting theme shouldn't be played, while sounds are off for the whole game
### Preliminary checks

While we start any battle, a short music theme is always played before the combat music. While we switch all sounds off, game still waits few seconds, before you can move your troops over battlefield.

Engine should skip the first few seconds and start battle immediately in a case sounds are off.

### Additional info

_No response_

oleg-derevenetz:
close #5536


ihhub:

PR #5549, opened 2022-06-17
Rework extended game options a bit

oleg-derevenetz:
Currently extended options that may affect the overall game balance are divided into `opt_battle`, `opt_world` and `opt_addons`. This division is somewhat incomprehensible, especially the last two options (I have no idea what the "addons" are and what is their difference from other options). In addition, many of the former extended options are no longer available. I propose to divide the extended options into just two categories - those that affect the overall game balance, and those that do not affect it.


ihhub:

PR #5552, opened 2022-06-18
Add extra information for XMI to MID conversion

ihhub:
This will help us to compose new songs as well. Also refer to [this wiki page](https://github.com/ihhub/fheroes2/wiki/MIDI-music-explanation).


src/engine/xmi2mid.cpp
```diff
@@ -357,17 +509,19 @@ struct MidiEvents : public std::vector<MidiChunk>
                 }
 
                 switch ( *ptr >> 4 ) {
-                // metadata
                 case 0x0F: {
+                    // Meta-Event, always starts from 0xFF.
```
oleg-derevenetz, line 513: I am aware that this code was written by the previous team long time ago, but I should note that this parser is error-prone in relation to out-of-bounds read, because in many cases `ptr` is incremented and then dereferenced without any bounds checking (bounds checking is performed just once per cycle in the `while` condition). Just an observation :)

P.S. Interesting, it is assumed here that if the upper half of the byte is `F`, then the lower half is also `F`, it is not even checked.

src/engine/xmi2mid.cpp
```diff
@@ -168,88 +315,92 @@ struct XMITrack
...
-{};
+using XMITracks = std::list<XMITrack>;
 
 struct XMIData
 {
     XMITracks tracks;
 
     explicit XMIData( const std::vector<uint8_t> & buf )
     {
+        // Please to https://moddingwiki.shikadi.net/wiki/XMI_Format#File_format
```
oleg-derevenetz, line 326: I suppose there should be "Please refer to ...".

oleg-derevenetz (approved):

src/engine/xmi2mid.cpp
```diff
@@ -497,99 +497,105 @@ struct MidiEvents : public std::vector<MidiChunk>
...
-                    emplace_back( delta + quantity.value, *ptr - 0x10, *( ptr + 1 ), 0x7F );
-                    ptr += 3 + quantity.lengthInBytes;
+            switch ( *ptr >> 4 ) {
+            // key pressure
+            case 0x0A:
+            // control change
+            case 0x0B:
+            // pitch bend
+            case 0x0E:
+                emplace_back( delta, *ptr, *( ptr + 1 ), *( ptr + 2 ) );
```
oleg-derevenetz, line 542: It is also possible to read outside the buffer here.

src/engine/xmi2mid.cpp
```diff
@@ -497,99 +497,105 @@ struct MidiEvents : public std::vector<MidiChunk>
...
+                ptr += 3;
+                break;
+
+            // XMI events do not have note off events.
+            // note on
+            case 0x09: {
+                emplace_back( delta, *ptr, *( ptr + 1 ), *( ptr + 2 ) );
```
oleg-derevenetz, line 549: And here.

src/engine/xmi2mid.cpp
```diff
@@ -497,99 +497,105 @@ struct MidiEvents : public std::vector<MidiChunk>
...
-                        }
+                // note off
+                emplace_back( delta + quantity.value, static_cast<uint8_t>( *ptr - 0x10 ), *( ptr + 1 ), static_cast<uint8_t>( 0x7F ) );
+                ptr += 3 + quantity.lengthInBytes;
+                break;
+            }
+
+            // Program Change: in other words which instrument is going to be played.
+            case 0x0C:
+                emplace_back( delta, *ptr, *( ptr + 1 ) );
```
oleg-derevenetz, line 563: And here.

oleg-derevenetz (commented):
There is still a lot of places where out-of-bounds read is possible if XMI is truncated/damaged (I pointed out a few). If there is no desire to rework this now, then we can at least leave a TODO comment.

src/engine/xmi2mid.cpp
```diff
@@ -587,11 +615,15 @@ struct MidiEvents : public std::vector<MidiChunk>
...
+                if ( !checkDataPresence( ptr, end, 2 ) ) {
+                    break;
+                }
+
                 emplace_back( delta, *ptr, *( ptr + 1 ) );
                 ptr += 2;
                 break;
 
-            // unused command
+            // Uknown command.
```
oleg-derevenetz, line 626: It seems that there is a typo: "Uknown" -> "Unknown" :)

oleg-derevenetz (approved):
Hi @ihhub, everything looks good, except one typo in the comment :)

PR #5559, opened 2022-06-20
Multiple AudioManager fixes

oleg-derevenetz:
First of all, this PR aims to fix the potential failure of `AudioManager` to stop the sound/music. In addition to making the `_taskToExecute` atomic (so it can be safely accessed by different threads without holding the same mutex), `executeTask()` now acquires `_resourceMutex` before checking the `_taskToExecute` (and for the entire duration of its execution) to avoid the scenario when `_taskToExecute` was checked and task is planned, then `_resourceMutex` was acquired by the main thread, which stops the music, but then the worker thread starts the music again.

Possible scenarios (at least essential ones):

* 1\. `stopSounds()`/`ResetAudio()` is called from the main thread (MT) when `prepareTask()` is executing on the worker thread (WT). `prepareTask()` holds the `_mutex`, `_taskToExecute` has been set to something, then `_mutex` has been released. Then `executeTask()` will be eventually called, and it will holds the `_resourceMutex` until its task will be completed. Here we have the following options:

  * 1\.1\. If `_taskToExecute` will be cleared by `removeSoundTasks()`/`removeAllTasks()` on the MT before it will be checked by WT, then we are all set, no task will be performed.

  * 1\.2\. If it will be cleared later, then WT performs the full cycle, while MT is waiting on the `_resourceMutex`, and then, when WT finishes, MT acquires the `_resourceMutex` and calls `Music::Stop()`/`Mixer::Stop()`. All actions just performed by WT has been reset.

* 2\. `stopSounds()`/`ResetAudio()` is called from the MT when `executeTask()` is executing on the WT. The same as 1.1/1.2.

* 3\. `stopSounds()`/`ResetAudio()` is called from the MT between `prepareTask()` and `executeTask()` on the WT. We have the following options:

  * 3\.1\. Only `removeSoundTasks()`/`removeAllTasks()` is managed to be executed on MT before/during the `executeTask()` acquires the `_resourceMutex` on WT. The same as 1.1/1.2.

  * 3\.2\. `stopSounds()`/`ResetAudio()` is fully completed on MT before `executeTask()` acquires the `_resourceMutex` on WT. The `_taskToExecute` has already been cleared, so `executeTask()` has nothing to do.

* 4\. `stopSounds()`/`ResetAudio()` is called from the MT when no task is executed at all on WT. `removeSoundTasks()`/`removeAllTasks()` is called from the MT, it holds the `_mutex`, `_taskToExecute` has been cleared as well as all task queues. Even if WT has been already kicked and `prepareTask()` is a little late and already waiting on the `_mutex`, there will be no tasks when it acquires the `_mutex`.

Hi @ihhub can you come up with a scenario where the sound won't be stopped properly?

Also, I've always wondered why all the task queues are cleared when a certain type of sound should be played in non-asynchronous mode. In my opinion, this is wrong and we need to clear the task queue for sounds only of the type that we want to play in the non-async mode (for example, when we want to play music, then just the `musicTasks` should be cleared).

Also this PR fixes the regression after #5545 - terrain music and environment sounds wasn't updated properly after casting any teleport spells (Dimension Door, Town Gate, Town Portal).

### DIFF
--- a/src/fheroes2/audio/audio_manager.cpp
+++ b/src/fheroes2/audio/audio_manager.cpp
@@ -35,7 +35,6 @@
 #include <cassert>
 #include <deque>
 #include <optional>
-#include <tuple>
 #include <utility>
 
 namespace
@@ -233,11 +232,6 @@ namespace
 
             std::scoped_lock<std::mutex> lock( _mutex );
 
-            // Do not add sound to the queue if the queue already contains exactly the same sound
-            if ( std::find( _soundTasks.begin(), _soundTasks.end(), SoundTask( m82Sound, soundVolume ) ) != _soundTasks.end() ) {
-                return;
-            }
-
             _soundTasks.emplace_back( m82Sound, soundVolume );
 
             notifyWorker();
@@ -323,11 +317,6 @@ namespace
                 , soundVolume( soundVolume_ )
             {
                 // Do nothing.
-            }
-
-            bool operator==( const SoundTask & other ) const
-            {
-                return std::tie( m82Sound, soundVolume ) == std::tie( other.m82Sound, other.soundVolume );
             }
 
             int m82Sound{ 0 };
--- a/src/fheroes2/audio/audio_manager.cpp
+++ b/src/fheroes2/audio/audio_manager.cpp
@@ -33,8 +33,9 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
+#include <deque>
 #include <optional>
-#include <queue>
+#include <tuple>
 #include <utility>
 
 namespace
@@ -232,7 +233,12 @@ namespace
 
             std::scoped_lock<std::mutex> lock( _mutex );
 
-            _soundTasks.emplace( m82Sound, soundVolume );
+            // Do not add sound to the queue if the queue already contains exactly the same sound
+            if ( std::find( _soundTasks.begin(), _soundTasks.end(), SoundTask( m82Sound, soundVolume ) ) != _soundTasks.end() ) {
+                return;
+            }
+
+            _soundTasks.emplace_back( m82Sound, soundVolume );
 
             notifyWorker();
         }
@@ -263,9 +269,7 @@ namespace
         {
             std::scoped_lock<std::mutex> lock( _mutex );
 
-            while ( !_soundTasks.empty() ) {
-                _soundTasks.pop();
-            }
+            _soundTasks.clear();
 
             if ( _taskToExecute == TaskType::PlaySound ) {
                 _taskToExecute = TaskType::None;
@@ -287,10 +291,7 @@ namespace
         {
             std::scoped_lock<std::mutex> lock( _mutex );
 
-            while ( !_soundTasks.empty() ) {
-                _soundTasks.pop();
-            }
-
+            _soundTasks.clear();
             _loopSoundTask.reset();
 
             switch ( _taskToExecute ) {
@@ -308,11 +309,7 @@ namespace
             std::scoped_lock<std::mutex> lock( _mutex );
 
             _musicTask.reset();
-
-            while ( !_soundTasks.empty() ) {
-                _soundTasks.pop();
-            }
-
+            _soundTasks.clear();
             _loopSoundTask.reset();
 
             _taskToExecute = TaskType::None;
@@ -361,6 +358,11 @@ namespace
                 // Do nothing.
             }
 
+            bool operator==( const SoundTask & other ) const
+            {
+                return std::tie( m82Sound, soundVolume ) == std::tie( other.m82Sound, other.soundVolume );
+            }
+
             int m82Sound{ 0 };
             int soundVolume{ 0 };
         };
@@ -383,7 +385,7 @@ namespace
         };
 
         std::optional<MusicTask> _musicTask;
-        std::queue<SoundTask> _soundTasks;
+        std::deque<SoundTask> _soundTasks;
         std::optional<LoopSoundTask> _loopSoundTask;
 
         MusicTask _currentMusicTask;
@@ -409,7 +411,7 @@ namespace
 
             if ( !_soundTasks.empty() ) {
                 std::swap( _currentSoundTask, _soundTasks.front() );
-                _soundTasks.pop();
+                _soundTasks.pop_front();
 
                 _taskToExecute = TaskType::PlaySound;
 
--- a/src/fheroes2/audio/audio_manager.cpp
+++ b/src/fheroes2/audio/audio_manager.cpp
@@ -251,7 +251,7 @@ namespace
             notifyWorker();
         }
 
-        void removeSounds()
+        void removeSoundTasks()
         {
             std::scoped_lock<std::mutex> lock( _mutex );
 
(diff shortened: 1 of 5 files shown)


ihhub:
Hi @oleg-derevenetz , my apologies for the delay in reviews, I wasn't physically capable to do these days.

I had a castle with a surrounding sounds (the best is a lake, beach or river as they are loudest), I switched to any hero first and then fast double clicked on a castle in the list of castles. My previous changes fixed it but logically not for 100% according to the code. I hope this might help to verify your changes.

We have only one `AudioManager::PlaySound( wav, true );` call which by right should be asynchronous as it is for hero's movement steps. You're right about clearing music and other sounds for this case, we shouldn't do this.

`AudioManager::playLoopSounds` has only 1 call and it's asynchronous. We could make this function to be always asynchronous.

As for music: we can clear only music queue but then for each `AudioManager::PlayMusic()` we have to explicitly call `AudioManager::stopSounds` if needed.

oleg-derevenetz:
Hi @ihhub

> my apologies for the delay in reviews, I wasn't physically capable to do these days.

Nevermind, there is no hurry :) 

> I had a castle with a surrounding sounds (the best is a lake, beach or river as they are loudest), I switched to any hero first and then fast double clicked on a castle in the list of castles. My previous changes fixed it but logically not for 100% according to the code. I hope this might help to verify your changes.

Well yes, I did a few tests this way and everything looks fine (however it would be good to perform checks in the exact situation in which you had problems earlier). I mean, can you come up with some theoretical sequence of events when the approach of this PR (with atomic and additional lock) will not stop the sounds properly? 

> `AudioManager::playLoopSounds` has only 1 call and it's asynchronous. We could make this function to be always asynchronous.

We can also leave the synchronous option for balance, since other functions of this kind have it :) But I can remove it if you want. 

> As for music: we can clear only music queue but then for each `AudioManager::PlayMusic()` we have to explicitly call `AudioManager::stopSounds` if needed.

Hmm, but this is not related - clearing queues stops FUTURE sounds, while calling `stopSounds()` stops EXISTING sounds. In places where we want to stop sounds, we are already calling `stopSounds()`, and in places where we don't need to stop sounds, we don't need to clear the sound queue because it is mostly useless (if not even harmful) by itself.

ihhub:
Hi @oleg-derevenetz , the scenario is: music type is MIDI. A hero walks fast (speed 9) from a tile of one terrain type to an object on another terrain type. This makes all sounds of hero's walk to be queued till loading of a new music track is completed (which takes ~200 ms). So all sounds which are in the asynchronous queue will be played **after** the music track starts. But by that time the hero would be on the tile with the object which might have its own sound. So there is a chance that while trying to play the sound from the object and even clearing the queue of sounds (and/or loop sounds) the previous sound(s) would still be played.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -251,7 +251,46 @@ namespace
             notifyWorker();
         }
 
-        void removeSounds()
```
ihhub, line 254: I cannot comment in the above code so I will do here: `pushSound()` method has a drawback: we do not check the same sounds in the queue. We shouldn't push the same sounds as there's no reason to even play the same sound in multiple channels simultaneously. For example, in case of hero's walking on a high speed we might have multiple exact sounds in the queue and they will be played at the same time but in different channels.

Since there's no proper way to traverse through a queue we at least need to check the front element in the queue (the code can be better written):
```cpp
if ( !_soundTasks.empty() && _soundTasks.front().m82Sound == m82Sound && _soundTasks.front().soundVolume == soundVolume ) {
    return;
}
```
This scenario can happen when a hero moves from one terrain type to another with MIDI music type.

ihhub (changes requested):
Hi @oleg-derevenetz , I put several comments here. Would you mind to take a look at them?

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -251,7 +251,46 @@ namespace
             notifyWorker();
         }
```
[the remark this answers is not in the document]
oleg-derevenetz: @ihhub Why use queue then? We then can have just one variable of type `std::optional<LoopSoundTask>` then and always set it to the latest task.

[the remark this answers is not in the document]
ihhub: Yes. It makes much more sense. Something like this would be good enough:
```cpp
LoopSoundTask _queuedLoopSoundTask; // naming is a subject of change
LoopSoundTask _currentLoopSoundTask;
```

src/fheroes2/audio/audio_manager.cpp
oleg-derevenetz: Hi @ihhub we can replace `std::queue` by `std::deque` and we will have all the same benefits and also we will be able to traverse the queue (and even clear it in one action!). I pushed the corresponding commit.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -849,7 +900,8 @@ namespace AudioManager
             return;
         }
 
-        g_asyncSoundManager.sync();
+        g_asyncSoundManager.removeMusicTasks();
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub Well, I can just revert the change concerning clearing only one type of sounds and do as before - when adding a new sound of any type, all queues are cleared. Anyway, it's worked so far. What do you think?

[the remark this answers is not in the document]
ihhub: I think we can address it separately. It is indeed an issue and we shouldn't clear everything at once but to minimize changes in this pull request and focus on synchronization issues only maybe it's good to leave it as it is.

src/fheroes2/audio/audio_manager.cpp
ihhub: Great!

ihhub (approved):

ihhub:
Hi @oleg-derevenetz , the changes look good to me. If you don't have anything more to push then we can merge this PR.

oleg-derevenetz:
Hi @ihhub

> the changes look good to me. If you don't have anything more to push then we can merge this PR.

I just added a few TODO comments regarding clearing extra queues, I believe it's ready to merge now.

src/fheroes2/audio/audio_manager.cpp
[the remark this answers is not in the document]
oleg-derevenetz: I just added TODO comments regarding this.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -235,7 +233,12 @@ namespace
 
             std::scoped_lock<std::mutex> lock( _mutex );
 
-            _soundTasks.emplace( m82Sound, soundVolume );
+            // Do not add sound to the queue if the queue already contains exactly the same sound
```
oleg-derevenetz, line 236: Hi @ihhub I'm still not sure about this part. What scenario will it protect against? It will not help much against repetitive sounds (for example, horse steps on faster speeds), they quickly leave the queue to some free mixer channel, but already played sounds just don't have enough time to play to the end until the next one is started on the another mixer channel. Unit step sounds during battle are played in sync mode, but they are fast enough anyway to start playing the next sound before the previous ones end.

ihhub: Hi @oleg-derevenetz , I completely agree with you. This fix is only for case of hero's steps with MIDI music on. I don't see any other cases when it would be triggered. We could remove it if you wish, it won't change user experience but more effects resource usage as we're going to play the same sound in multiple channels.

oleg-derevenetz: Let's remove this check for now.

ihhub:
@oleg-derevenetz , great job! We have just few more places with audio...

Issue #5447, opened 2022-05-23
Remove dependency on SDL for tools on Visual Studio


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

We removed a lot of dependencies from SDL in the past. Most of tools do not use SDL related code at all so we need to build them without such dependency. This is required for further steps towards simplification of artists work setup.

### Describe the possible solution

Remove SDL dependency on Visual Studio projects for tools.

oleg-derevenetz:
Hi @ihhub If the problem is only to give artists access to tools, then we can just pack them additionally into a ZIP file, because we currently build them anyway.

ihhub:
Closing this issue as we fixed it in a different manner.

PR #5565, opened 2022-06-23
Rework tools

undef21:
* make tools independent from SDL;
* add usage info in output;

related issue: #5447


undef21:
@oleg-derevenetz after merging #5628 all windows external dependencies will be built and placed in [fheroes2-prebuilt-deps ](https://github.com/fheroes2/fheroes2-prebuilt-deps), right? If someone add a new windows library he needs to create PR in that repo?

oleg-derevenetz:
@undef21 they are already placed in that repo (more precisely, that repo contains scripts that build these dependencies, and dependencies itself are placed to the Releases section of that repo). If someone wants to add, remove or upgrade the dependencies, then a PR in that repo should be opened, yes. This only applies to Windows dependencies of course.

oleg-derevenetz:
@undef21 I suppose you will need `libpng` dependency for SDL1? For SDL2 it is already present.

undef21:
@oleg-derevenetz , yes I'll need `libnpg` later. `libpng` comes to project with `sdl2_image`. This pull request replaces `sdl_image`'s `IMG_SavePNG` function with `libpng` so we'll should delete `sdl_image` and add `libpng` for both SDL versions later.

oleg-derevenetz:
Hi @undef21 I just added `libpng` to the Windows pre-built dependency archive, so you may use it in this PR if you want, but `SDL_image` is not removed from this archive yet - I'll remove it later as soon as it is no longer needed.

src/engine/image_tool.cpp
```diff
@@ -58,68 +51,116 @@ namespace
...
+        sf.putRaw( reinterpret_cast<const char *>( pixels.data() ), pixels.size() );
+        sf.close();
+        return true;
+    }
 
-        memcpy( surface->pixels, image.image(), width * height );
+#if defined( ENABLE_PNG )
+    bool SavePNG( const fheroes2::Image & image, const std::string & path )
+    {
+        StreamFile sf;
```
ihhub, line 125: Can we please add some styling to this function as right now it is a monolith text which is really hard to read?

src/engine/image_tool.cpp
```diff
@@ -58,68 +51,116 @@ namespace
...
-        SDL_SetPaletteColors( surface->format->palette, paletteSDL.data(), 0, 256 );
-#else
-        SDL_SetPalette( surface, SDL_LOGPAL | SDL_PHYSPAL, paletteSDL.data(), 0, 256 );
-#endif
-
-        const uint32_t width = image.width();
-        const uint32_t height = image.height();
+    bool SaveBMP( const fheroes2::Image & image, const std::string & path )
+    {
+        const size_t HEADER_SIZE = 124;
```
ihhub, line 85: We need to save it in 8-bit format, 32-bit is too much. For reference you can look [here](https://github.com/ihhub/penguinV/blob/master/src/file/bmp_image.cpp).

ihhub (commented):
Hi @undef21 , I took a quick look at this pull request and left several comment here. Please check them when you have free time.

src/engine/image_tool.cpp
ihhub: Also that repository contains PNG related code as well. If we are going to copy it we need to make a reference where this code was taken. I think it would be just cop-paste code as penguinV works with 8-bit images mostly.

src/engine/image_tool.h
```diff
@@ -31,10 +32,10 @@ namespace fheroes2
     // Save an image into file. 'background' represents palette index from the original palette. Recommended value is 23.
     bool Save( const Image & image, const std::string & path, const uint8_t background );
 
-    bool Load( const std::string & path, Image & image );
```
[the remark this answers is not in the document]
undef21: @ihhub, this method is only one depended on SDL in this file. After moving we can build for example icn2img `icn2img: image.c serialize.c logging.c system.c` without SDL.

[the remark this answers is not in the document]
ihhub: Please check [this comment](https://github.com/ihhub/fheroes2/pull/5565/files#r928207655) which will remove any dependency on SDL for image operations.

ihhub:
Closing this pull request as we handled most of work in recent commits.

PR #5570, opened 2022-06-25
Fix the castle well build logic

### LINKED ISSUE #5162: AI missed to build the well in a castle
### Preliminary checks

### Platform

Windows

### Describe the bug

The castle has some other structures, which are not in priority (for instance, the Rainbow), but doesn't has cheap and important well.
image

### Save file

[Fail week of Druids.zip](https://github.com/ihhub/fheroes2/files/8340527/Fail.week.of.Druids.zip)

### Additional info

_No response_

idshibanov:
Fixes #5162 .

Building wells only on last day of the week is unreliable. Do that only during the first week.


ihhub:
@idshibanov , thank you very much for the fix!

PR #5571, opened 2022-06-25
AI should keep some creatures in reserve for castle defense

### LINKED ISSUE #5304: Adventure map, AI behavior, AI should keep some creatures in reserve for the possible defense of a castle (0.9.14 5285)
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the AI should keep some creatures in reserve for the possible defense of a castle.

In fheroes2:

Day 7 of the week.
The blue player's hero (human player) approaches the orange player's castle.
Castle no reserve defense troops 001

Day 1 of next week.
The orange player's castle is not defended.
Castle no reserve defense troops 002

The number of creatures available corresponds to the generation of creatures at the beginning of the week.
So the AI did not have the possibility to recruit creatures to put in defense.
Castle no reserve defense troops 003

### Save file

Here is the save file:
[Before no town defense 01.zip](https://github.com/ihhub/fheroes2/files/8515119/Before.no.town.defense.01.zip)

### Additional info

_No response_

idshibanov:
Fixes #5304 .

Now AI could leave some garrison (preferably slow units) behind when reinforcing the heroes if region is not completely safe.

Fighter heroes should be less likely to leave troops in town than regular ones to avoid players intentionally exploiting the feature by splitting the army on castle recaptures (common OG flaw).

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -155,8 +155,9 @@ namespace AI
             bool onlyHalf = false;
             Troop * unitToSwap = heroArmy.GetSlowestTroop();
             if ( unitToSwap ) {
-                double significanceRatio = isFigtherHero ? 20.0 : 10.0;
                 // if it's slow and doesn't contribute much to army strength we leave it all in town
+                // 5% of total army for Figthers 10% otherwise
+                const double significanceRatio = isFigtherHero ? 20.0 : 10.0;
                 if ( unitToSwap->GetStrength() > armyStrength / significanceRatio ) {
                     Troop * weakest = heroArmy.GetWeakestTroop();
 
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -148,6 +148,7 @@ namespace AI
 
         const uint32_t regionID = world.GetTiles( castle.GetIndex() ).GetRegion();
         // check if we should leave some troops in the garrison
+        // TODO: amount of troops left could depend on region's safetyFactor
         if ( castle.isCastle() && _regions[regionID].safetyFactor <= 100 && !garrison.isValid() ) {
             const Heroes::Role heroRole = hero.getAIRole();
             const bool isFigtherHero = ( heroRole == Heroes::Role::FIGHTER || heroRole == Heroes::Role::CHAMPION );
@@ -155,8 +156,6 @@ namespace AI
             bool onlyHalf = false;
             Troop * unitToSwap = heroArmy.GetSlowestTroop();
             if ( unitToSwap ) {
-                // if it's slow and doesn't contribute much to army strength we leave it all in town
-                // 5% of total army for Figthers 10% otherwise
                 const double significanceRatio = isFigtherHero ? 20.0 : 10.0;
                 if ( unitToSwap->GetStrength() > armyStrength / significanceRatio ) {
                     Troop * weakest = heroArmy.GetWeakestTroop();
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -125,10 +125,74 @@ namespace AI
 
         if ( recruit && buyArmy ) {
             CastleTurn( castle, underThreat );
-            ReinforceHeroInCastle( *recruit, castle, kingdom.GetFunds() );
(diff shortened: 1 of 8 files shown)


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -125,12 +125,77 @@ namespace AI
...
+        // check if we should leave some troops in the garrison
+        if ( castle.isCastle() && _regions[regionID].safetyFactor <= 100 && !garrison.isValid() ) {
+            const Heroes::Role heroRole = hero.getAIRole();
+            const bool isFigtherHero = ( heroRole == Heroes::Role::FIGHTER || heroRole == Heroes::Role::CHAMPION );
+
+            bool onlyHalf = false;
+            Troop * unitToSwap = heroArmy.GetSlowestTroop();
+            if ( unitToSwap ) {
+                // if it's slow and doesn't contribute much to army strength we leave it all in town
+                // 5% of total army for Figthers 10% otherwise
```
ihhub, line 159: This comment does not correct the line below.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -125,12 +125,77 @@ namespace AI
...
+        Army & garrison = castle.GetArmy();
+        const double armyStrength = heroArmy.GetStrength();
+
+        heroArmy.UpgradeTroops( castle );
+        castle.recruitBestAvailable( budget );
+        heroArmy.JoinStrongestFromArmy( garrison );
+
+        const uint32_t regionID = world.GetTiles( castle.GetIndex() ).GetRegion();
+        // check if we should leave some troops in the garrison
+        if ( castle.isCastle() && _regions[regionID].safetyFactor <= 100 && !garrison.isValid() ) {
```
ihhub, line 151: Shouldn't it be `_regions[regionID].safetyFactor < 0`? For example I started "Great War II" map. All regions have safety factor >= 0 where no enemy heroes exist.

ihhub (changes requested):
Hi @idshibanov , I put 2 comments here. Could you please in your free time check them?

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
idshibanov: Negative safety is when there's enemy heroes/castles in the region present. I actually thought of excluding those as well.

Safety 0 to 100 should be a neutral region that can have unexplored on dangerous regions nearby. If AI is too lazy to explore it they can leave some questionable empty castles. Only 100 to 200 (max) safety can be considered really safe (they have to be deep into friendly territory when having more than 1 castle).

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
idshibanov: This comment is justifying magic values for `significanceRatio`. If you think it's clear as is I can remove.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
ihhub: What if we make percentage of the army based on safety factor? Firstly, it will prevent players to abuse AI behavior as it will be unknown how much army AI hero would leave. Secondly, it would make more stronger AI heroes in regions which have close to 100 safety factor. What do you think?

ihhub: As an alternative we can leave a TODO comment to the future adjustments.

idshibanov: Yes, let's review it later after all changes will be merged and tested together.

ihhub (approved):

src/fheroes2/heroes/heroes.cpp
```diff
@@ -538,7 +538,7 @@ uint32_t Heroes::GetMaxMovePoints() const
         point += 500 * world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, GetColor() );
     }
     else {
-        const Troop * troop = army.GetSlowestTroop();
+        Troop * troop = army.GetSlowestTroop();
```
oleg-derevenetz, line 541: Is there a reason to remove `const` here? Usually we can't remove the `const` qualifier safely, but we [always can add ](https://godbolt.org/z/59nnv4Yje)it.

oleg-derevenetz (changes requested):
Hi @idshibanov I just left one question, could you please take a look when you have time?

oleg-derevenetz (approved):

ihhub:
@idshibanov , excellent work on AI!

PR #5574, opened 2022-06-25
Fix AI hero recruiment limit

idshibanov:
`availableHeroCount` wasn't incremented so AI could regularly buy 5 heroes on the first turn on large maps.


ihhub:

PR #5040, opened 2022-02-19
Allow different translations of "Red Tower"

a1exsh:
Both Sorceress and Warlock have a building named "Red Tower".  In some
translations, e.g. Russian "Buka" version, they are translated differently,
which makes it a bit nicer.

In order to support the same we have to use distinct translatable strings.

PR #5577, opened 2022-06-26
Update Polish translation - "Red Tower" #5040

Arthusppp:
Allow different translations of "Red Tower" #5040


ihhub:
Hi @Arthusppp , is it correct that this pull request just removes fuzzy comments?

Arthusppp:
Hi @ihhub. Depends on the master. At the end if the day, the translation should contain both red towers translated as "Czerwona Wieża".

ihhub:
@Arthusppp , thanks a lot for keeping Polish translation up to date!

PR #5582, opened 2022-06-28
Fix multiple audio issues

### LINKED ISSUE #5580: Main menu, gameplay sounds are playing during adjustment of sound volume after returning to main menu
### Preliminary checks

### Platform

Windows

### Describe the bug

When we return to main menu and start adjusting sound effects volume, game starts playing some of sounds, from the last played map.
image

### Save file

Way to reproduce:
Load map "Arrax the Jerk -> return to main menu, enter sound options and press sound effect volume button. There we can hear, I suppose, some volcano sounds.

### Additional info

_No response_

### LINKED ISSUE #5581: Main Menu, while fast switching to external music game continues playing midi
### Preliminary checks

### Platform

Windows

### Describe the bug

When we have selected MIDI music and then make fast double click to set External, game still continues playing MIDI in main menu.
image

### Save file

Use mouse left click as fast as you can to select External.

### Additional info

_No response_

oleg-derevenetz:
fix #5580
fix #5581

Scenario of the #5581 is as follows:

* The music type is changed for the first time, the current music is set to `MUS::UNKNOWN`, `MusicRestorer` started the preparation for the background music playback in the worker thread, BUT actual music playback is not started yet.
* The music type is quickly changed for the second time, BUT, since the actual music playback is still not started, the current music is still the `MUS::UNKNOWN`, so `MusicRestorer` has nothing to restore, and no music playback is planned for the worker thread.
* The previously planned music playback is actually started by the worker thread.

The proposed solution is to always start music playback in sync mode when changing the music type. There may be some minor delay if we are playing MIDI on Windows (~200ms?), but we will always be sure that the music playback is really started before the music type can be changed once again.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -76,8 +76,8 @@ namespace
     // a SDL_Mixer internal thread.
     //
     // TODO: according to SDL_Mixer manual, calls of any SDL_Mixer functions are not allowed in
-    // callbacks. The current code works, but it would be good to find a reliable way to perform
-    // channel cleanup without calling these functions.
+    // TODO: callbacks. The current code works, but it would be good to find a reliable way to
+    // TODO: perform channel cleanup without calling these functions.
     void channelFinished( const int channelId )
     {
         // This callback function should never be called if audio is not initialized
--- a/src/fheroes2/audio/audio_manager.cpp
+++ b/src/fheroes2/audio/audio_manager.cpp
@@ -746,10 +746,10 @@ namespace
                 // Adjust channel based on given parameters.
 
                 // TODO: this is very hacky way. We should not do this. For example in 3D audio mode when a hero moves alongside beach it is noticeable that ocean sounds
-                //       are 'jumping' in volume. Instead of such approach we need to get free channel ID which will be used for playback. Set volume for it and then
-                //       start playing. Such logic must be implemented within Audio related code.
-                //       As an alternative solution: we can use channel IDs which we freed in the previous step. However, be careful with synchronization for audio
-                //       access.
+                // TODO: are 'jumping' in volume. Instead of such approach we need to get free channel ID which will be used for playback. Set volume for it and then
+                // TODO: start playing. Such logic must be implemented within Audio related code.
+                // TODO: As an alternative solution: we can use channel IDs which we freed in the previous step. However, be careful with synchronization for audio
+                // TODO: access.
                 Mixer::Pause( channelId );
                 Mixer::setVolume( channelId, info.volumePercentage * soundVolume / 10 );
                 Mixer::Resume( channelId );
--- a/src/fheroes2/audio/audio_manager.cpp
+++ b/src/fheroes2/audio/audio_manager.cpp
@@ -483,7 +483,7 @@ namespace
         // Check if the music track is cached.
         if ( Music::Play( musicUID, playbackMode ) ) {
             DEBUG_LOG( DBG_ENGINE, DBG_TRACE, "Play cached music track " << trackId )
-            Game::SetCurrentMusicTrack( trackId );
+            Game::SetCurrentMusicTrackId( trackId );
             return;
         }
 
@@ -514,7 +514,7 @@ namespace
             else {
                 Music::Play( musicUID, filename, playbackMode );
 
-                Game::SetCurrentMusicTrack( trackId );
+                Game::SetCurrentMusicTrackId( trackId );
 
                 DEBUG_LOG( DBG_ENGINE, DBG_TRACE, "Play music track " << MUS::getFileName( trackId, MUS::EXTERNAL_MUSIC_TYPE::MAPPED, " " ) )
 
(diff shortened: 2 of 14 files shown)


oleg-derevenetz:
BTW, this PR solves the #5581, but for some reason on Windows the external music sounds quieter after a quick switch (when I select the MIDI music and then quickly press twice on the music type icon to switch to the External music) than after a gradual switch (when I'm waiting for the actual launch of the MIDI Expansion music playback before switching to the External music). And it sounds quieter even if I then quit the game and completely restart it from scratch, and even if I then change the music volume in the game settings! WTF. Another one funny issue with MIDI on Windows? This issue is not observed on macOS.

ihhub (changes requested):
Hi @oleg-derevenetz , I left 2 comments in this pull request. Could you please take a look at them?

Regarding #5581 : we can use a combination of `Mix_FadingMusic` function and checking the type of music before stopping the current music track.

oleg-derevenetz:
Hi @ihhub

> Regarding https://github.com/ihhub/fheroes2/issues/5581 : we can use a combination of Mix_FadingMusic function and checking the type of music before stopping the current music track.

I have an impression that this might add an extra delay when stopping the music in cases when it might be undesirable, because, if we are using MIDI music on Windows, then we have that famous extra 200ms delay, the completion of which `Mix_FadingMusic()` will also wait for.

ihhub (approved):

ihhub:

I was thinking even further: what if we try to play MIDI as sounds? In this case we won't care about overall volume on Windows. This is of course just in theory and I'm not sure that this would work.

oleg-derevenetz:
Hi @ihhub

> what if we try to play MIDI as sounds?

As far as I understand, `Mix_Chunk` cannot be created from `MIDI`. It is created by `Mix_LoadWAV()`/`Mix_LoadWAV_RW()` and these functions are able to create the `Mix_Chunk` from (naturally) `WAV`, `AIFF`, `RIFF`, `OGG` and `VOC`, but not from `MIDI`.

LeHerosInconnu:
Hello @oleg-derevenetz and @ihhub,

I thought of this to add the Wasteland and Beach/Ocean terrain type music, among other things, which do not exist in MIDI.
But why not save **all** MIDI music as audio files (.ogg or .wav) and when the player selects "MIDI" or "MIDI expansion" it is the audio file versions of the MIDI music that are used?

Unless the developers have fun solving problems by keeping the MIDI mode. :smile:

ihhub:
Hi @LeHerosInconnu , MIDI is not an ordinary audio file. It's a collection of instructions for instruments to be played opposed to recorder sounds in WAV or MP3 which are limited by frequencies.

LeHerosInconnu:
@ihhub.

> Hi @LeHerosInconnu , MIDI is not an ordinary audio file. It's a collection of instructions for instruments to be played opposed to recorder sounds in WAV or MP3 which are limited by frequencies.

Yes, I know how MIDI files work. :)
At first my idea was to play the music in MIDI format and record it while it is being played to be transformed into an audio file format, but this can be done by using a MIDI to OGG converter.

Isn't WAV hi-res, almost similar to FLAC?

Related link: https://videoconverter.wondershare.com/ogg/ogg-to-midi.html

oleg-derevenetz:
Hi @LeHerosInconnu

> At first my idea was to play the music in MIDI format and record it while it is being played to be transformed into an audio file format

If you mean doing it on the fly on the user's machine, then it's not trivial to do it in a portable way on all supported platforms, because the game itself may not even have an access to the resulting audio stream (to record it in a file) in case when external synthesizers are used for the MIDI playback. If you mean doing it in advance and then distribute the recorded audio files as part of the game, then there are licensing issues arise.

LeHerosInconnu:
@oleg-derevenetz.

> Hi @LeHerosInconnu
> 
> > At first my idea was to play the music in MIDI format and record it while it is being played to be transformed into an audio file format
> 
> If you mean doing it on the fly on the user's machine, then it's not trivial to do it in a portable way on all supported platforms, because the game itself may not even have an access to the resulting audio stream (to record it to a file) in case when external synthesizers are used for the MIDI playback (I even believe it's most of the cases - MIDI files are just passed to the external software or hardware synthesizer and it plays them directly). If you mean doing it in advance and then distribute the recorded audio files as part of the game, then there are licensing issues arise.

My idea is to make it "ready to use" for the player, so yes, the files would have to be converted first.

What if the program converts the MIDI files into memory and uses them that way without generating audio files (to avoid licensing issues)?

oleg-derevenetz:
@LeHerosInconnu

> What if the program converts the MIDI files into memory and uses them that way without generating audio files (to avoid licensing issues)?

Then we have to bundle some software synthesizer with the game to convert MIDI to WAV or some other audio format on the fly. Technically this may be possible (for example, there is [FluidSynth](https://www.fluidsynth.org/) in active development). We may even use it on Windows only, because issues with MIDI are usually Windows-specific.

PR #5585, opened 2022-07-01
Fix build order regression for AI towns

idshibanov:
Caused by oversight in #5570. AI currently wouldn't upgrade towns to castles because of the focus on the well.


ihhub:

PR #5586, opened 2022-07-01
Update debug mode logging to include hero role

idshibanov:
Added AI hero role to the debug output and reformatted the rest of it to make it a bit more concise.

Also the "has no parent tile" warning is my personal annoyance since it happens frequently and pollutes the log.

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -107,6 +107,26 @@ namespace
 
         return result;
     }
+
+    std::string GetHeroRoleString( const Heroes & hero )
+    {
+        switch ( hero.getAIRole() ) {
+        case Heroes::Role::SCOUT:
+            return _( "Scout" );
+        case Heroes::Role::COURIER:
+            return _( "Courier" );
+        case Heroes::Role::HUNTER:
+            return _( "Hunter" );
+        case Heroes::Role::FIGHTER:
+            return _( "Fighter" );
+        case Heroes::Role::CHAMPION:
+            return _( "Champion" );
+        default:
+            break;
+        }
+
+        return _( "Unknown" );
+    }
 }
 
 const char * Heroes::GetName( int heroid )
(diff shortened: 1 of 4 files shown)


src/fheroes2/heroes/heroes.cpp
```diff
@@ -107,6 +107,26 @@ namespace
...
+        case Heroes::Role::COURIER:
+            return _( "Courier" );
+        case Heroes::Role::HUNTER:
+            return _( "Hunter" );
+        case Heroes::Role::FIGHTER:
+            return _( "Fighter" );
+        case Heroes::Role::CHAMPION:
+            return _( "Champion" );
+        default:
+            break;
```
ihhub, line 125: I suggest to add
```cpp
// Did you add a new AI hero role? Add the appropriate logic for it!
assert( 0 );
```
here to notify a developer that they miss something.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -107,6 +107,26 @@ namespace
 
         return result;
     }
+
+    std::string GetHeroRoleString( const Heroes & hero )
+    {
+        switch ( hero.getAIRole() ) {
+        case Heroes::Role::SCOUT:
+            return _( "Scout" );
```
ihhub, line 115: We display these strings not for users but for developers. In this case do we need to have translation for them? If the answer is still yes I would suggest to add a prefix for them. For example, `_( "ai-hero-role|Scout" );`.

ihhub (changes requested):
Hi @idshibanov , I put several comment here. Could you please check them?

src/fheroes2/heroes/heroes.cpp
idshibanov: Good catch. I don't think we need a translation here, I'll get rid of it.

ihhub (approved):

ihhub:
@idshibanov , thanks a lot for these changes!

PR #5587, opened 2022-07-02
Remove unused field move_point_scale from the Heroes class

oleg-derevenetz:
Apparently this field was once used for some experimental functionality, but it's not the case anymore.


ihhub:
@oleg-derevenetz , good job for cleaning up the code!

PR #5588, opened 2022-07-04
Mixer::SetChannels(): don't call the Mix_ReserveChannels()

oleg-derevenetz:
I have no idea why we need to reserve the channel 0.


ihhub:
@oleg-derevenetz , well done! Neither I had any idea about this :)

PR #5592, opened 2022-07-05
Fix Campaign Window button flickering

### LINKED ISSUE #4499: Campaign info window, flickering buttons when pressing mouse on a button's shadow
### Preliminary checks

### Platform

Linux

### Describe the bug

When pressing the left mouse button on the shadow of any of the buttons in the Campaign info window the button switches rapidly between pressed and released state, until the mouse press is released or the mouse pointer is moved away from the button.  (I've tried to record the video, but unfortunately the flickering is not visible on the recording.)

The reason for the issue seems to be that the (pre-rendered) shadow is part of the released button sprite, but the pressed sprite doesn't include the shadow and therefore has smaller area (e.g. `CAMPXTRE` sprites 0 and 1 — the two states of the "View Intro" button):

image image

It doesn't hurt that clicking the button shadow activates the button, IMO, but the flickering is not nice.  We could, if we like it, extend the clickable area of the pressed button to match that of the released one: that would eliminate the flickering.

### Save file

No save file is needed.

### Additional info

_No response_

ihhub:
close #4499

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -2194,6 +2194,25 @@ namespace fheroes2
 
                 return true;
             }
+            case ICN::CAMPXTRG:
+            case ICN::CAMPXTRE: {
+                LoadOriginalICN( id );
+
+                auto & image = _icnVsSprite[id];
+
+                if ( image.size() >= 8 ) {
+                    for ( size_t i = 0; i < 4; ++i ) {
+                        Sprite resized( image[2 * i].width(), image[2 * i].height() );
+                        resized.reset();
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

oleg-derevenetz:
Hi @ihhub there is a serious regression: now game crashes on an assertion when trying to generate various buttons based on `ICN::CAMPXTRG`/`ICN::CAMPXTRE`, such as "Restart" button in the battle result dialog, "OKAY" button in the "Battle Only" window, etc, etc.

ihhub:
Oh, I'll fix it.

PR #5597, opened 2022-07-06
Show a window stating about missing video files for campaigns

ihhub:
This will help to avoid any questions from players regarding missing campaigns.

image

### DIFF
--- a/src/fheroes2/game/game_newgame.cpp
+++ b/src/fheroes2/game/game_newgame.cpp
@@ -138,7 +138,7 @@ namespace
     {
         fheroes2::showMessage( fheroes2::Text{ _( "Warning!" ), fheroes2::FontType::normalYellow() },
                                fheroes2::Text{ _( "Required video files are missing. Please make sure that all necessary files are present in the system." ),
-                                               fheroes2::FontType::normalYellow() },
+                                               fheroes2::FontType::normalWhite() },
                                Dialog::OK );
     }
 }
--- a/src/fheroes2/game/game_newgame.cpp
+++ b/src/fheroes2/game/game_newgame.cpp
@@ -138,7 +138,8 @@ namespace
     {
         fheroes2::showMessage( fheroes2::Text{ _( "Warning!" ), fheroes2::FontType::normalYellow() },
                                fheroes2::Text{ _( "Required video files are missing. Please make sure that all necessary files are present in the system." ),
-                               fheroes2::FontType::normalYellow() }, Dialog::OK );
+                                               fheroes2::FontType::normalYellow() },
+                               Dialog::OK );
     }
 }
 
--- a/src/fheroes2/game/game_newgame.cpp
+++ b/src/fheroes2/game/game_newgame.cpp
@@ -41,6 +41,8 @@
 #include "text.h"
 #include "translations.h"
 #include "ui_button.h"
+#include "ui_dialog.h"
+#include "ui_text.h"
 #include "ui_tool.h"
 #include "world.h"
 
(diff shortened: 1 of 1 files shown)


zenseii (commented):
Hi @ihhub!

This currently works well for SW campaigns, but for the expansion there is one common situation where the dialog will not appear, which is when you have the .smk files in the data directory, but you are lacking the anim directory.

I believe this PR wants to address the case when a user doesn't have the .smk files in what for windows is the anim directory, which is what happens relatively often.

I believe that in most of these cases a user will still have the .smk files in the data directory, in other words the ones that show the small clips for expansion campaign selection.

ihhub:
Hi @zenseii , this pull request addresses complains from some users saying that no campaign selection exists. A popup window does not contain any concrete information but points that some files are missing. This immediately highlights to a user that they need to copy all files or purchase the original game if they try to use a pirated version.

zenseii:
I see, @ihhub, and this means there are two issues that can be addressed by this PR.

What I'm trying to say is that the only way for the dialog to appear right now when you click the expansion campaign is if you are both missing the .smk files in the `anim` directory AND in the `data` directory.

However, I've found that typically users who miss video files usually have the video files that are originally present in the `data` directory like in the screenshot underneath showing the `data` directory. This is because often they have an old gog install with different folder setup etc that made the script fail in copying the `anim` resources. As I said, in my installation if I don't have the `anim` directory, the message will anyways not appear - I would have to remove `IVYDES.SMK` , `IVYPOL.SMK`, `IVYVOY.SMK` and `IVYWIZ.SMK`
to have the dialog appear for the expansion campaign selection, and it is very unlikely that the problem of a user is that they are missing these four .smk files, but very likely that they are missing the .smk files in the `anim` folder. Yet they will not be given feedback about this in this case.
DATA:
image
fheroes2:
image

src/fheroes2/game/game_newgame.cpp
```diff
@@ -323,6 +334,7 @@ fheroes2::GameMode Game::NewPriceOfLoyaltyCampaign()
 
     if ( !videos[0] ) {
         // File doesn't exist. Fallback to PoL campaign.
+        showMissingVideoFilesWindow();
```
zenseii, line 337: Basically what I wrote in my comment is that this dialog is very unlikely to appear and the user might be given a false positive thinking they have all files since a dialog for any price of loyalty campaign won't show up. At worst they might think that fheroes2 doesn't have support for these video files of the expansion yet, which I've heard players say.
Here's a vid which hopefully explains what I mean better:

https://user-images.githubusercontent.com/12501091/177584701-ad70197b-8b35-4546-b056-cdd43b484282.mp4

Maybe we should add code that displays this dialog after you've selected a campaign in the PoL menu and no .smk videos are available?

src/fheroes2/game/game_newgame.cpp
```diff
@@ -131,6 +133,14 @@ namespace
         COUT( "Press " << Game::getHotKeyNameByEventId( Game::HotKeyEvent::NEW_WIZARDS_ISLE_CAMPAIGN ) << " to choose Wizard's Isle Campaign." )
         COUT( "Press " << Game::getHotKeyNameByEventId( Game::HotKeyEvent::NEW_DESCENDANTS_CAMPAIGN ) << " to choose Descendants Campaign." )
     }
+
+    void showMissingVideoFilesWindow()
+    {
+        fheroes2::showMessage( fheroes2::Text{ _( "Warning!" ), fheroes2::FontType::normalYellow() },
+                               fheroes2::Text{ _( "Required video files are missing. Please make sure that all necessary files are present in the system." ),
```
ihhub, line 140: How about we show this message:
```cpp
Required video files for campaign selection window are missing. Please make sure that all necessary files are present in the system.
```
?

zenseii: Yes, I see now what you mean. This sentence you proposed is better.

zenseii (approved):

PR #5605, opened 2022-07-07
When scouting, use a straight line distance instead of an approximate distance, rework the neutral monsters counting logic for the extended scouting

### LINKED ISSUE #1466: Artifact, Crystal Ball, Crystal Ball effect distance
In fheroes2, the effective distance for the Crystal Ball artifact is not always the same as in the original game.
For a strictly horizontal and strictly vertical distance, the Crystal Ball artifact works correctly.
For other distances, there are differences.
I have not tested all the possibilities, but it must be related to how the distance is calculated according to the radius from the position of the hero.

In original game:

Crystal Ball effect distance 05

Crystal Ball effect distance 06

In fheroes2:

Crystal Ball effect distance 07

Crystal Ball effect distance 08

In original game:

Crystal Ball effect distance 01

Crystal Ball effect distance 02

In fheroes2:

Crystal Ball effect distance 03

Crystal Ball effect distance 04

In original game:

Crystal Ball effect distance 09

Crystal Ball effect distance 10

In fheroes2:

Crystal Ball effect distance 11

Crystal Ball effect distance 12

### LINKED ISSUE #2570: Make "extended scouting" random number not that random
Currently when the option "extended scouting" is on and when the hero has either basic or advance scouting, it will display approximate numbers for monsters and army
However this is randomly generated everytime you right-click, so you get a new number each time which makes it easy to guess what is the correct number.
Also the formula used is made such that even the random interval allow you to deduce the initial amount.
What I propose:
- seed the RNG for scouting based on tile position for monster
- seed the RNG for scouting based on monster id for hero's armies
- change the formula used to compute the random interval such that it is harder to figure out the initial troup size

oleg-derevenetz:
fix #1466

In the original game:

https://user-images.githubusercontent.com/32623900/177887700-aa85ac47-4400-4e59-9620-740ed7260736.mp4

With this PR:

https://user-images.githubusercontent.com/32623900/177887727-5a151fc8-63d4-45d7-bdba-bba2248a3451.mp4

---

close #2570

Rework the "extended scouting" algorithm of neutral monster counting by excluding the random from it (a situation where the numbers are constantly changing looks weird). The proposed algorithm is as follows:

1. The range into which a given number of monsters falls (one of standard ranges, i.e. 1 - 5, 5 - 10, 10 - 20, 20 - 50, etc) is taken;
2. With basic scouting level, the range is divided in half and the part of the range into which the monster count falls is returned. For example, for 1 wolf and 2 wolves "1-3 wolves" will be shown, for 3 and 4 wolves "3-5 wolves" will be shown, for 5 wolves "5-7 wolves" will be shown, and for 30 goblins "20-35 goblins" will be shown;
3. With advanced scouting level, the range is divided into four parts and the part of the range into which the monster count falls is returned. For example, for 20 and 25 goblins "20-27 goblins" will be shown, and for 30 goblins "27-35 goblins" will be shown;
4. With expert scouting level, the exact monster count is returned.

This algorithm is a subject to discussion of course.

### DIFF
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -133,7 +133,7 @@ namespace
         }
         else if ( isAbandonnedMine ) {
             const uint8_t spriteIndex = tile.GetObjectSpriteIndex();
-            if ( spriteIndex == 5 ) { // TODO: remove this hardocded value for real abandoned mine.
+            if ( spriteIndex == 5 ) { // TODO: remove this hardcoded value for real abandoned mine.
                 str = MP2::StringObject( objectType );
             }
             else {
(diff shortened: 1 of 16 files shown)


oleg-derevenetz:
Hi @idshibanov this PR adds another method for measuring the distance between tiles (I have made an effort to explain the difference in the comments to them, I don't know how clear it turned out). I replaced the `GetApproximateDistance()` by `GetStraightLineDistance()` in those places where I found it suitable (mostly in places where the distance between tiles is not tied to the route that the hero would have to overcome), but I'm in doubt regarding this place:

https://github.com/ihhub/fheroes2/blob/1781dd48209bdd2c4d9ee5109174fb8c1c82eb8a/src/fheroes2/world/world_regions.cpp#L69-L79

idshibanov (approved):
I think regions code can use the straight line distance. Otherwise looks good.

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -456,33 +456,23 @@ namespace
...
             }
-        }
-        else if ( Settings::Get().ExtWorldScouteExtended() ) {
-            uint32_t dist = static_cast<uint32_t>( hero->GetScoute() );
-            if ( hero->Modes( Heroes::VISIONS ) && dist < hero->GetVisionsDistance() )
-                dist = hero->GetVisionsDistance();
 
-            if ( Maps::GetApproximateDistance( hero->GetIndex(), dst ) <= dist )
-                return scoutingLevel;
             return Skill::Level::NONE;
```
ihhub, line 476: We can remove this line as the return below is the same.

ihhub (approved):

src/fheroes2/dialog/dialog_quickinfo.cpp
ihhub: To speed up work I removed this line myself.

ihhub:
@oleg-derevenetz , thanks a lot for fixing this logic!

Issue #5569, opened 2022-06-24
fheroes2 crashed after moving any hero to the castle


darkhorror1984:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Game crashed after moving any hero to the castle. Map name's - ARRAX.

### Save file

[AUTOSAVE.zip](https://github.com/ihhub/fheroes2/files/8981696/AUTOSAVE.zip)

idshibanov:
@darkhorror1984 what's the game version? Can't reproduce on latest build.

oleg-derevenetz:
Hi @idshibanov this savefile was produced by 0.9.16, which is not even the latest release (which is 0.9.17), not to mention the latest build. However even 0.9.16 should not crash when visiting a castle.

Hi @darkhorror1984 could you please grab the output from the console window at the moment of the crash? If this crash is due to the assertion failure (and most likely it is) there may be some useful information - which assertion was failed and where.

oleg-derevenetz:
Hmm, I still not able to reproduce the crash even with Russian resource files from Buka.

darkhorror1984:
The console window is clear.
![crash_windows8](https://user-images.githubusercontent.com/108156692/175761786-a24f900e-3d3d-4cae-93fd-1a1e329d7b3a.png)

oleg-derevenetz:
Hi @darkhorror1984 Could you please run the Event Viewer (by pressing `Win+R` and entering `eventvwr`), then open Windows Logs -> Application and look for related entries with source "Application Error" and "Windows Error Reporting"? If there are any, please copy their contents and paste it here. If there are links to some files with additional information, it would be good to get these files as a ZIP archive here as well.

darkhorror1984:
[error_report.zip](https://github.com/ihhub/fheroes2/files/8984815/error_report.zip)

darkhorror1984:
[error_report.zip](https://github.com/ihhub/fheroes2/files/8984867/error_report.zip)
[full_dump.zip](https://github.com/ihhub/fheroes2/files/8984868/full_dump.zip)

oleg-derevenetz:
[...]

PR #5608, opened 2022-07-09
Do not call SDL_Mixer functions from the callback set by the Mix_ChannelFinished()

oleg-derevenetz:
Related to #5569


ihhub:
Hi @oleg-derevenetz , an idea occured to me: what if we cache all sounds like we do for music? We're talking about kilobytes of data, not even megabytes.

oleg-derevenetz:
Hi @ihhub

> an idea occured to me: what if we cache all sounds like we do for music? We're talking about kilobytes of data, not even megabytes.

I've already thought about it, but... in this case, you need some kind of ID, which you can bind the cached `Mix_Chunk` to. While with ordinary in-game sounds this is relatively easy (just use their `M82` id), sound channels from videos are different story - they do not have an ID by themselves. So I decided not to cache sounds so far.

src/engine/audio.cpp
```diff
@@ -72,25 +73,91 @@ namespace
...
+        {
+            // Make sure that all sound samples have been eventually freed
+            assert( std::all_of( _channelSamples.begin(), _channelSamples.end(), []( const auto & item ) { return item.second.empty(); } ) );
+        }
+
+        SoundSampleManager & operator=( const SoundSampleManager & ) = delete;
+
+        void channelStarted( const int channelId, Mix_Chunk * sample )
+        {
+            const auto iter = _channelSamples.find( channelId );
```
ihhub, line 92: Maybe we should assert **sample** here,

src/engine/audio.cpp
```diff
@@ -72,25 +73,91 @@ namespace
...
+                sampleDeque.push_back( sample );
+
+                return;
+            }
+
+            const auto res = _channelSamples.try_emplace( channelId, std::deque<Mix_Chunk *>{ sample } );
+
+            if ( !res.second ) {
+                assert( 0 );
+            }
```
ihhub, line 106: `assert( !res.second );` would work too :)

oleg-derevenetz: Not quite - in this case we will get "unused variable" warning when we build with `NDEBUG`.

src/engine/audio.cpp
oleg-derevenetz: We can indeed, but since we are not dereferencing `sample` here, I decided not to add an assertion for it.

ihhub: Got it.

src/engine/audio.cpp
```diff
@@ -106,7 +173,14 @@ namespace
...
         if ( channel < 0 ) {
             ERROR_LOG( "Failed to play an audio chunk for channel " << channel << ". The error: " << Mix_GetError() )
+
+            Mix_FreeChunk( sample );
         }
+        else {
+            soundSampleManager.channelStarted( channel, sample );
+        }
+
+        soundSampleManager.clearFinishedSamples();
```
ihhub, line 183: According to [this page](https://wiki.libsdl.org/SDL_mixer/Mix_FreeChunk) `SDL_mixer will stop any channels this chunk is currently playing on.` which is very risky as the same channel might be used to play the current sound. Should we call this function before **Mix_PlayChannel**?

oleg-derevenetz: @ihhub `clearFinishedSamples()` will free just one sample for each occurence of the same channel. For example, if `_channelsToCleanup` will contain `{ 1, 1 }`, then `_channelSamples` should contain exactly two samples for this channel, and they will be freed in turn, one by one. AFAIK there is no way that currently played sound will be freed. Let's consider the following options:

1. Just launched channel finished playing before `channelStarted()` call. It will be added to the `_channelsToCleanup`, then `channelStarted()` is called and it will be added to the `_channelSamples` and then freed by `clearFinishedSamples()`. There is no issue.
2. Just launched channel finished playing after `channelStarted()` call, but before `clearFinishedSamples()`. It will be added both to `_channelsToCleanup` and `_channelSamples`, no issue again.
3. Just launched channel was not finished playing during execution of this function, it will be added only to `_channelSamples` and therefore will not be freed. No issue again.

> Should we call this function before `Mix_PlayChannel`?

There is no difference in this regard.

oleg-derevenetz: @ihhub in fact using the `std::deque` might be even somewhat overkill, because with current logic there can be maximum two samples in each queue (for each channel) at the same time, so even `std::pair` might be used instead. But this will raise the cognitive complexity of the code.

ihhub: My point is that according to the documentation `Mix_FreeChunk` stops playback of the channel. So in case of your second scenario shouldn't it stop the playback of the channel when the second sound is playing?

oleg-derevenetz: @ihhub It stops playback on the channel only if the sample that is being freed is currently played on that channel. In the second scenario it is not played on that channel anymore.

ihhub: Got it. Then in this case the solution looks good to me.

oleg-derevenetz: @ihhub I decided to replace `std::deque` with `std::pair`, because deque is really an overkill here.

ihhub:
Hi @oleg-derevenetz , how confident are we to upgrade the library just before release?

src/engine/audio.cpp
```diff
@@ -72,25 +73,107 @@ namespace
...
+
+        SoundSampleManager & operator=( const SoundSampleManager & ) = delete;
+
+        void channelStarted( const int channelId, Mix_Chunk * sample )
+        {
+            assert( channelId >= 0 && sample != nullptr );
+
+            const auto iter = _channelSamples.find( channelId );
+
+            if ( iter != _channelSamples.end() ) {
```
ihhub, line 100: From code readability std::deque was better :) Now it is hard to read it due to first and second presence.

oleg-derevenetz:
Hi @ihhub

> how confident are we to upgrade the library just before release?

This is optional and could be reverted. It makes sense to upgrade now only if it solves the MIDI issue on Windows 8.

src/engine/audio.cpp
oleg-derevenetz: `std::deque` means extra heap allocation, while `std::pair` lives entirely in the memory block allocated by the `std::map` for its element.

undef21 (commented):
With PowerShell 4.0 and .NET Framework 4.8 (WIn8) WebClient doesn't follow to redirect and return error instead file

src/engine/audio.cpp
oleg-derevenetz: @ihhub BTW haha funny, Microsoft STL has the block size for `std::deque` exactly 16 bytes (x2 64-bit pointers) :)

https://github.com/microsoft/STL/blob/062002644137095c606f0a5a11d04f0278e8632e/stl/inc/deque#L550-L558

This means that for for everything larger than 16 bytes deque becomes "vector of pointers" instead of "vector of values", which means another level of heap allocations. LOL, why so small??? GCC' libstdc++ allocates 512 bytes for the same memory block.

oleg-derevenetz:
> With PowerShell 4.0 and .NET Framework 4.8 (WIn8) WebClient doesn't follow to redirect and return error instead file

That's sad :( AFAIK `System.Net.WebClient` has the `AllowAutoRedirect` property set to `true` by default. AFAIK GitHub always redirects to a slurred URL when downloading.

ihhub:
> Hi @ihhub
> 
> > how confident are we to upgrade the library just before release?
> 
> This is optional and could be reverted. It makes sense to upgrade now only if it solves the MIDI issue on Windows 8.

Is it confirmed that the newer version solves Win 8 issue?

ihhub:
Let's revert library upgrade for now and create another pull request for it. We can merge this one after this manipulation.

oleg-derevenetz:
@ihhub

> Is it confirmed that the newer version solves Win 8 issue?

[Partially](https://github.com/ihhub/fheroes2/issues/5569#issuecomment-1180087914). Game doesn't crash anymore, but the MIDI music is not played.

ihhub:
@oleg-derevenetz , then it's very risky to roll the update for this release. Let's proceed with your changes🙂

ihhub:

undef21:
> That's sad :( AFAIK `System.Net.WebClient` has the `AllowAutoRedirect` property set to `true` by default. AFAIK GitHub always redirects to a slurred URL from `githubusercontent.com` when downloading. Any ideas?

The problem was in TLS 1.2. Fixed line is:
`powershell -Command " [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; (New-Object System.Net.WebClient).DownloadFile('https://github.com/libsdl-org/SDL_mixer/releases/download/release-2.6.0/SDL2_mixer-devel-2.6.0-VC.zip', '%tempPath%\sdl_mixer\sdl_mixer2.zip')"`

Interesting, is there someone who build the game with win7 or win8?

PR #5609, opened 2022-07-09
Preparations for Adventure Map rendering

ihhub:
Add extra comments for the future Adventure Map rendering to be done in a correct way.

Also fixed invalid logic when neutral monsters whose number was set in the Editor did not grow at all.

### DIFF
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -961,7 +961,7 @@ void Maps::Tiles::PlaceMonsterOnTile( Tiles & tile, const Monster & mons, const 
         tile.MonsterSetJoinCondition( Monster::JOIN_CONDITION_SKIP );
     }
     else if ( setDefinedCount || ( world.GetWeekType().GetType() == WeekName::MONSTERS && world.GetWeekType().GetMonster() == mons.GetID() ) ) {
-        // TODO: why there are neutral monsters whose number was set in the Editor always join for money? Check the logic in the original game.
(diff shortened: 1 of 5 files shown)


src/fheroes2/maps/maps_tiles_quantity.cpp
```diff
@@ -966,7 +960,9 @@ void Maps::Tiles::PlaceMonsterOnTile( Tiles & tile, const Monster & mons, const
         // Ghosts and elementals never join hero's army.
         tile.MonsterSetJoinCondition( Monster::JOIN_CONDITION_SKIP );
     }
-    else if ( tile.MonsterFixedCount() || ( world.GetWeekType().GetType() == WeekName::MONSTERS && world.GetWeekType().GetMonster() == mons.GetID() ) ) {
+    else if ( setDefinedCount || ( world.GetWeekType().GetType() == WeekName::MONSTERS && world.GetWeekType().GetMonster() == mons.GetID() ) ) {
+        // TODO: why there are neutral monsters whose number was set in the Editor always join for money? Check the logic in the original game.
```
oleg-derevenetz, line 964: Hi @ihhub yeah, wandering monsters with the number of units specified by the map designer are always considered as "hostile" and always join only for money:

https://handbookhmm.ru/2-diplomacy

oleg-derevenetz (approved):

src/fheroes2/maps/maps_tiles_quantity.cpp
ihhub: Great! I updated the comment.

PR #5618, opened 2022-07-11
Make best AI figther a champion

idshibanov:
Minor change to enable Champion role in few cases. Will act as a Fighter but this increases the chance best hero will get the army reinforcements from other heroes, including courier.


idshibanov:
@ihhub Up to you if you want to include it in outgoing release. Scope is relatively minor, but I like the behavior.

ihhub:
Hi @idshibanov , let's merge it right after the release. We delayed it for a bit and usually new changes require more testing.

ihhub (approved):

ihhub:
@idshibanov , thank you so much for the improvement!

PR #5621, opened 2022-07-11
MSVC: build with the latest SDL2_Mixer 2.6.0

oleg-derevenetz:
Related to #5569, should be merged after #5608


oleg-derevenetz:
@ihhub we can wait with it and later try to use mixer with FluidSynth (we can try to install extra libs using `vcpkg` BTW instead of direct download using batch files, its [portfile](https://github.com/microsoft/vcpkg/blob/master/ports/sdl2-mixer/portfile.cmake) has support for FluidSynth).

ihhub:
I would recommend to merge this just after the release to have enough time to test.

ihhub:
Hi @oleg-derevenetz , I assume to support the same library everywhere we need to update it on Windows since Unix systems soon or later will receive SDL Mixer 2.6 update.

This is still safe to do without introducing any (I hope) big issues.

What do you think?

ihhub:
Also they released SDL 2.6.1 yesterday :)

oleg-derevenetz:
Hi @ihhub I thought we were going to move in the direction of our own assembly of this library (with FluidSynth) :) If so, there is no point to update it now.

ihhub:
Haha, this is true but I am not convinced yet due to some unsolved problems pointed by you. We could take incremental steps towards the goal if possible like we did for soundfont loading code.

oleg-derevenetz:
@ihhub

> Haha, this is true but I am not convinced yet due to some unsolved problems pointed by you.

But I suppose we should at least try :) The most disturbing issue is the hang on exit if we switch between MIDI and external music during the game, but this issue may already be solved in the latest SDL_mixer (or the latest FluidSynth).

> We could take incremental steps towards the goal if possible like we did for soundfont loading code.

We need to address the issue from @undef21 first:

https://github.com/ihhub/fheroes2/pull/5608#issuecomment-1180608120

oleg-derevenetz:
@ihhub Currently I'm pushing this PR for `vcpkg` that will update the `SDL2_mixer` to 2.6.1:

https://github.com/microsoft/vcpkg/pull/25763

So I hope that we will be able to check it soon with `fheroes2`.

oleg-derevenetz:
@undef21 I still hope that we will not need this at all :) But if we decide not to go with 3rd party MIDI synthesizer, I'll do as you suggest.

oleg-derevenetz:
Closing this one because #5628 looks more promising.

PR #5623, opened 2022-07-11
Update main page screenshots

ihhub:
To address all new features of the game. The main page will be loading faster due to smaller size images (reduction by 350 KB). All images are saved as 8-bit images in comparison with 24-bit before.


noomorph (approved):

zenseii (commented):
Hi @ihhub!
Finally we can see modern scrollbars on the main page.

My only concern is how a wider picture will look on the https://ihhub.github.io/fheroes2/ page?

zenseii (approved):

Issue #5141, opened 2022-03-19
Popping sounds when using MP3 tracks


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

This issue was first reported in this PR containing a discussion on it:
https://github.com/ihhub/fheroes2/pull/5137
I'm not getting this with neither ogg nor flac audio files.
It sounds like fire crackling.

https://user-images.githubusercontent.com/12501091/159126265-ca274f09-3df6-4f36-bd8d-312b69b451f5.mp4

### Save file

Seems to be more prevalent in castle themes, although you can hear it other places too.

### Additional info

Might need to put the volume a bit higher. As stated before these popping sounds don't appear if the files are played outside of fheroes2.

I'm using the non-expansion tracks.

zenseii:
Actually listening to this with my speakers I don't hear the noise so I'm going to have to check on another PC to rule out that it's not specific to this PC.

zenseii:
Just checked another PC with the same MP3 files and the same sounds can be heard there.

ihhub:
Hi @zenseii , could you please also specify your OS? It is important to know as every OS has its own SDL implementation.

zenseii:
Right sorry I forgot. My second PC has Windows 7. Used SDL2 latest snapshot release build. Still GOG supplied MP3 tracks.

ihhub:
Hi @zenseii , what's the frequency of MP3 tracks?

zenseii:
Hi @ihhub. It's 320 kbps and 48000 Hz.

ihhub:
Hi @zenseii , did it help?

zenseii:
Hi @ihhub. I just tested it and it does help. I can't hear these crackling noises.
Here's a recording with 22050 Hz.
https://user-images.githubusercontent.com/12501091/174477862-6ee7b248-e769-4fa8-80f5-368333894687.mp4

ihhub:
I am closing this issue. The engine uses 22050 Hz frequency, setting it higher than 44100 Hz will result in such issues. Recommended frequency is 22050 Hz. We could state this somewhere in docs or in a wiki page.

Issue #4232, opened 2021-09-20
Extra sounds while playing MIDI music on Xubuntu 20.04


lua-rocks:
### Your question

When I start the game, I hear two tunes, one on top of the other - the usual background midi music and some awful pipes on top. Is it possible to turn off these nightmarish pipes? Thank you.

ihhub:
Hi @lua-rocks , could you please provide more information about your issue? It's not a question but rather a bug.
At minimum we need your operating system information, which exact version of fheroes2 you're running, your hardware. What version of Homm2 do you use? Does external music work?

lua-rocks:
Hi @ihhub , here is my system info:
```
                          ./+o+-       luarocks@MAINFRAME
                  yyyyy- -yyyyyy+      OS: Xubuntu 20.04 focal
               ://+//////-yyyyyyo      Kernel: x86_64 Linux 5.4.0-84-generic
           .++ .:/++++++/-.+sss/`      Uptime: 7h 50m
         .:++o:  /++++++++/:--:/-      Packages: 2409
        o:+o+:++.`..```.-/oo+++++/     Shell: fish 3.1.0
       .:+o:+o/.          `+sssoo+/    Resolution: 1280x720
  .++/+:+oo+o:`             /sssooo.   DE: Xfce
 /+++//+:`oo+o               /::--:.   WM: Xfwm4
 \+/+o+++`o++o               ++////.   WM Theme: Greybird
  .++.o+++oo+:`             /dddhhh.   GTK Theme: Greybird [GTK2]
       .+.o+oo:.          `oddhhhh+    Icon Theme: elementary-xfce-darker
        \+.++o+o``-````.:ohdhhhhh+     Font: Noto Sans 11
         `:o+++ `ohhhhhhhhyo++os:      Disk: 141G / 456G (33%)
           .o:`.syhhhhhhh/.oo++o`      CPU: AMD Phenom II X4 945 @ 4x 3GHz
               /osyyyyyyo++ooo+++/     GPU: GeForce GTX 465
                   ````` +oo+++o\:     RAM: 910MiB / 3935MiB
                          `oo++. 
```

fheroes2 version: **0.9.7 (05 September 2021)**

I hear these pipes using any files - from demo or from original homm2 sucession wars.

Also I've tried to switch option `music` in `~/.config/fheroes2/fheroes2.cfg` to all possible values (original, expansion, external) and it doesn't help.

ihhub:
[...]

PR #5628, opened 2022-07-12
Build SDL_Mixer with FluidSynth on Windows

### LINKED ISSUE #237: No sounds when music is off even when sounds are not
Set music level to off in settings and notice that no sounds are available after.
image

### LINKED ISSUE #2009: Combat screen, sometimes sound level of intro music of the combat is lower than sound level of music of the combat
In fheroes2, sometimes the sound level of the intro music of the combat is lower than the sound level of the music of the combat.
This is valid for Grass, Lava, Dirt, Desert, and Snow terrain types and sometimes also for Swamp terrain.

Here are the save and test scenario files:
[Test music start combat 01.zip](https://github.com/ihhub/fheroes2/files/5386889/Test.music.start.combat.01.zip)

### LINKED ISSUE #2901: Horse's steps are fading while crossing terrain border using MIDI sound

### LINKED ISSUE #5528: External audio not always returning to on after turning volume to zero
### Preliminary checks

### Platform

Windows

### Describe the bug

See relevant comment https://github.com/ihhub/fheroes2/pull/5526#pullrequestreview-1003469825

Audio doesn't restart for external audio or just plays midi music after having turned off volume, regardless of what audio type you currently have on. Sometimes midi music gets played instead of external, depending on in what audio type you turn the volume off or on.

It is not specific to the PR mentioned above. 

**EDIT: I've checked that it happens in 0.9.15 too.**

https://user-images.githubusercontent.com/12501091/173192395-95cc9752-384b-415e-9923-ed2fa62aa6da.mp4

### Save file

Just open config dialog in any build.

### Additional info

@oleg-derevenetz, it would be great if you have the time and opportunity to take a look at this.
I was asked by @ihhub to ask you if you could since he is unavailable to do so right now.

### LINKED ISSUE #5569: fheroes2 crashed after moving any hero to the castle

### LINKED ISSUE #5622: Main menu, fast switch external music causes it play too quiet.
### Preliminary checks

### Platform

Windows

### Describe the bug

The same actions described in #5581 cause now low music level. I believe there's some problem with fade.
image
We still need to press "Music type" button slow to avoid problems with music.

### Save file

None.

### Additional info

_No response_

oleg-derevenetz:
fix #237
fix #2009
fix #2901
fix #5528
fix #5569
fix #5622

Related to #5047 - it fixes the resumption of MIDI music (at least on Windows).

Current issues:

1. ~If music type is changed during the game at least once from MIDI to external (or vice versa), then the game waits for something on exit. Maybe this could be cured by using the latest `SDL_Mixer` or latest `FluidSynth` (both are not latest in the `vcpkg` port database), but maybe (although doubtful) it is somehow related to our code~ **FIXED** by building the app with the latest `SDL_mixer` (2.6.1);
2. ~Build times are HUGE, because we need to build libs by hand. I added the cache, but sometimes (when the cache is not used) build may take more than 1 hour~ **FIXED** by installing `vcpkg` tree from fork. This fork should be updated only episodically - when the new version of some library is released and we want to use it. In this circumstances the cache helps well - package installation takes just a few dozens of seconds;
3. MIDI sounds a bit different, but we can try different sound fonts (~BTW sound fonts are also HUGE - dozens or even hundreds of megabytes~ **FIXED** by adding support for SF3 (compressed) SoundFont format);
4. ~This will work only with SDL2. I even temporarily removed the Windows build with SDL1 because MSVC project structure was changed slightly and adaptation of scripts for SDL1 is still to be done~ **FIXED**: script adaptation is performed.


oleg-derevenetz:
Hi @ihhub what do you think about this (especially about using `vcpkg` and corresponding build times?). You can try Windows binary artifacts, they should work just fine (except that issue with waiting for something on exit, possibly on SDL audio deinitialization, for unknown reason).

ihhub:
Hi @oleg-derevenetz , I am just curious how it is possible that the original game plays MIDI with no issues but our project can't?

Build times are horrible for such small project. If we want to reduce timings we can build them in a separate repository and then just pull out.

I can see that this solution is far from perfect from now but I fell we are moving to the right direction.

oleg-derevenetz:
@ihhub

> I am just curious how it is possible that the original game plays MIDI with no issues but our project can't?

Well, original game was developed for DOS, so it obviously didn't use Windows MIDI subsystem, it has its own. More recent Windows version is based on that DOS version (apparently it's the same game with some sort of wrapper), so it uses the same home-grown MIDI synthesizer as the original (DOS) version. Why they should have issues that related to the Windows native MIDI synthesizer? :)

oleg-derevenetz:
@ihhub

> So why can't we use MIDI synthesizer for DOS as well? For example, [this one](https://github.com/Mindwerks/wildmidi) looks promising (I haven't tested).

That's what we currently do - we use the 3rd-party synthesizer (Fluidsynth). Why FluidSynth? Because it can be seamlessly integrated with SDL_Mixer and also it is in active development (which means that there are people that regularly fix buggz).

ihhub:
Let's do then step by step: does SDL Mixer 2.6 fix some of related issues? If then can we update it first without relying on a heavy packages and lengthy compilation times?

oleg-derevenetz:
@ihhub

Well, then I make a special PR today to add the support for sound font loading. It will not harm even if there are no any sound fonts (like it will be on all systems except Windows). 

> A 30 MB archive is a little overkill for the MIDI issue fix

Most of storage is occupied by the sound font file which is effectively the set of instruments for the synthesizer. I just get the recommended one on the Fluidsynth's website with appropriate license (BTW it sounds very well for my taste in the game, what do you think?). But we can always look for a smaller one.

oleg-derevenetz:
@ihhub

> A 30 MB archive is a little overkill for the MIDI issue fix😆

[Here](https://github.com/FluidSynth/fluidsynth/wiki/SoundFont) the 6Mb sound font is offered. I tried it with fheroes2 and it sounds different from the Windows MIDI, but with its own charm :) I believe there are many of them.

ihhub:
Hi @oleg-derevenetz , according to your description these changes fix MIDI resumption on Windows but **isMusicResumeSupported()** function doesn't have this format enabled. Is music track different using FluidSynth?

oleg-derevenetz:
Hi @ihhub

> Is music track different using FluidSynth?

Yes, it seems so. Apparently it changed from MIDI to the output music format of the FluidSynth (I don't know which one exactly, but this can be figured out using debugger or debug print).

> With recent changes in xmi2mid.cpp file we can extract all instruments used in xmi files and limit soundfonts to them.

Well yes, we can try to assemble our own sound font. In fact there is even no need to be a programmer for this, so maybe there are volunteer musicians that will be able to help in this regard.

oleg-derevenetz:
Hi @ihhub the issue when the app hung up when exiting is fixed with the latest `SDL_mixer`, you can try the build [here](https://github.com/ihhub/fheroes2/actions/runs/2673996797#artifacts).

oleg-derevenetz:
Hi @undef21 it would be great if you tried [this](https://github.com/ihhub/fheroes2/actions/runs/2673996797#artifacts) experimental build on WinXP (if you have free time for this) and shared your impressions.

oleg-derevenetz:
Hi @undef21

First of all, thank you very much for your help.

> hears good for me. I tested this build with win8. There is 5-10 seconds delay on starting of the game, then the midi sounds works well.

It's funny, on the Win10 machine that I got the startup process is instant. Maybe a Windows 8 machine just has a weaker hardware. Anyway, I have an idea that this is due to the need to parse and load the sound font file 30Mb in size :) I suppose that with smaller sound font file it would be better. If there is time and desire, try to replace the existing sound font file (`files/soundfonts/GeneralUser GS v1.471.sf2`) by the 6Mb sound font (`TimGM6mb`) from [here](https://github.com/FluidSynth/fluidsynth/wiki/SoundFont): just remove the existing sound font file and copy this to the same directory instead. There is even no need to rename it, just make sure that the file extension is `.sf2`. The sound will be slightly different, but there should be no pause on startup I believe.

> Also this build works with win xp too. But the game has many other issues with this system and make a player to solve additional quests for starting it.

We can try to build 32-bit version for WinXP, but this is not a priority task yet. And then when it becomes a priority, WinXP may already disappear :)

oleg-derevenetz:
Hi @Branikolog @LeHerosInconnu @zenseii you are welcome to try this experimental build as well (just get the latest one from this PR) and share your opinion.

Branikolog:
Hi, @oleg-derevenetz !
Wow, it sounds completely different. :)
I've experienced a few issues with this build:
image
And also, while Alt+Tab fullscreen application it prevents selecting of any system window because it wants to continue playing music.

undef21:
>> It's funny, on the Win10 machine that I got the startup process is instant.

It's more funny that build loads fonts on winxp instantly. Anyway the latest build doesn't have this problem on win8 too and loads the 30MB font instantly.

>> And then when it becomes a priority, WinXP may already disappear

:) Is winxp only reason to support SDL1?

oleg-derevenetz:
Hi @Branikolog

> Wow, it sounds completely different. :)

It depends on the sound font used, there are many of them, you can try different ones by replacing the `.sf2` file in `files/sounfdonts`. But most of them are big. @ihhub has an idea to assemble our own sound font only with the necessary instruments.

> I've experienced a few issues with this build:

Do you mean these "Instrument not found" messages? They are harmless, because FluidSynth performs instrument substitution if there is no such instrument in the currently used sound fonts. Perhaps native Windows MIDI also does not find some instruments and performs a similar substitution, it just does not report it :)

> And also, while Alt+Tab fullscreen application it prevents selecting of any system window because it wants to continue playing music.

Yeah, here is an issue, but it is not entirely clear why it happens. This build uses more recent SDL (2.0.22 vs 2.0.12 in the master branch). I just tried the build without FluidSynth, but with the more recent SDL, and faced the same issue, so more likely it is related either to SDL or to our code. We will need to debug this on Windows, since only Windows use this "true full screen" mode.

oleg-derevenetz:
@undef21

> It's more funny that build loads fonts on winxp instantly. Anyway the latest build doesn't have this problem on win8 too and loads the 30MB font instantly.

The latest build is a bit "lighter" compared to what you've tried before, because since latest SDL_mixer supports many music formats natively (without 3rd-party libs), I threw out a lot of extra libraries, so maybe it wasn't the SF2 load that was slowing down at all :) Anyway, it's good that the problem is no more.

> :) Is winxp only reason to support SDL1?

To be honest, I don't know exactly. I believe we still support the SDL1 just because we can :)

Branikolog:
@oleg-derevenetz 
>Do you mean these "Instrument not found" messages? They are harmless, because FluidSynth performs instrument substitution if there is no such instrument in the currently used sound fonts. Perhaps native Windows MIDI also does not find some instruments and performs a similar substitution, it just does not report it :)

Okay. It just made me suspicious, so I asked you about that. :)

>Yeah, here is an issue, but it is not entirely clear why it happens. This build uses more recent SDL (2.0.22 vs 2.0.12 in the master branch). I just tried the build without FluidSynth, but with the more recent SDL, and faced the same issue, so more likely it is related either to SDL or to our code. We will need to debug this on Windows, since only Windows use this "true full screen" mode.

I hope you'll find the solution. Minimized fheroes2 from the above build makes it impossible to run or use any application in system.

oleg-derevenetz:
@Branikolog

> I hope you'll find the solution. Minimized fheroes2 from the above build makes it impossible to run or use any application in system.

I already find it (sort of) - just replace the `SDL2.dll` by the same file from the latest release and issue will disappear :) So at least we just can bundle the earlier version of SDL2 on Windows. Anyway, it's still not entirely clear is this related to the SDL itself or just we are doing something wrong and this "wrong way" is not compatible with newer SDL anymore.

Hi @ihhub what do you think regarding this issue?

ihhub:
> @Branikolog
> 
> > I hope you'll find the solution. Minimized fheroes2 from the above build makes it impossible to run or use any application in system.
> 
> I already found it (sort of) - just replace the `SDL2.dll` by the same file from the ZIP archive of the latest release and issue will disappear :) So at least we just can bundle the earlier version of SDL2 on Windows. Anyway, it's still not entirely clear is this related to the SDL itself or just we are doing something wrong and this "wrong way" is not compatible with newer SDL anymore.
> 
> Hi @ihhub what do you think regarding this issue?

I remember we rolled back SDL 2 update on Windows due to some cursor related issues. Looks like it has more. I will take a look at it and possibly changed API which we don't use.

ihhub:
The problem is that the application in fullscreen mode immediately regains focus when a user clicks on another window. This issue started from 2.0.14 by introduction of `SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS` which developers claim to be set to false. However, in reality we have to manually set this hint to fix the issue.

I will test other parts of the game before rising a pull request to update SDL to 2.0.22.

oleg-derevenetz:
Hi @ihhub we will eventually need our own fork of `vcpkg` for better caching and dependency control. Currently I use my own [fork](https://github.com/oleg-derevenetz/vcpkg), but then (when my PR regarding `sdl2-mixer` will be eventually merged by Micro$oft people) we will need to make the project-wide mirror under [fheroes2](https://github.com/fheroes2) organization I suppose.

The idea is as follows: we make a fork and create a "snapshot" (by creating a tag or a branch) and use this snapshot in our own builds until we need to move to a different version of some library - then we create another "snapshot" and move to it. This approach provides stability, good cache hits and good build speed.

There is always a possibility that the compiler itself will change to a new version (in this case the cache will be rebuilt), but I suppose this should happen relatively rarely.

LeHerosInconnu:
Hello @oleg-derevenetz,

> Hi @Branikolog @LeHerosInconnu @zenseii you are welcome to try this experimental build as well (just get the latest one from this PR) and share your opinion.

Yes, it sounds different.
For me, this sort of "vibrato" in some notes is not welcome, it gives an impression of wobbly music. :)

I also have some errors:

Error MIDI test 01

The first time I launched the game, there was a long waiting time (several seconds) with the impression of a game that freezes.
This long time disappeared with the following launches of the game (x64 SDL2 Windows 7).
I didn't try to shut down the PC completely and launch the game again to see if the waiting time is still present when the game is launched for the first time.

oleg-derevenetz:
@LeHerosInconnu

> The first time I launched the game, there was a long waiting time (several seconds) with the impression of a game that freezes. This long time disappeared with the following launches of the game (x64 SDL2 Windows 7).

Can it be due to the antivirus software? The symptoms are very similar to that - a long start for the first time, fast in the following.

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu
> 
> > The first time I launched the game, there was a long waiting time (several seconds) with the impression of a game that freezes. This long time disappeared with the following launches of the game (x64 SDL2 Windows 7).
> 
> Can it be due to the antivirus software? The symptoms are very similar to that - a long start for the first time, fast in subsequent attempts.

Maybe, I will try to look at this next time I will restart the PC. :)

oleg-derevenetz:
@LeHerosInconnu

> For me, this sort of "vibrato" in some notes is not welcome, it gives an impression of wobbly music. :)

Fortunately, now you can choose the instruments yourself if you want, replacing the `.sf2` file in `files/soundfonts` with another one :) By the way, are you by any chance a musician? :)

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu
> 
> > For me, this sort of "vibrato" in some notes is not welcome, it gives an impression of wobbly music. :)
> 
> Fortunately, now you can choose the instruments yourself if you want, replacing the `.sf2` file in `files/soundfonts` with another one :)

Nice.

> By the way, are you by any chance a musician? :)

No, but I do whistle. :)
But I doubt I can reproduce all musics with high fidelity. :rofl:

oleg-derevenetz:
BTW I joust found the original Windows MIDI sound font (licensed my MS from Roland at the time). It has just ~3Mb in size. Unfortunately there is no way to bundle it with our game due to licensing issues.

LeHerosInconnu:
@oleg-derevenetz.

> BTW I just found the original Windows MIDI sound font, licensed by MS from Roland at the time (1990s). It has just ~3Mb in size. Unfortunately there is no way to bundle it with our game due to licensing issues.

There are warnings with this sound font:

FluidSynth Warning 001

FluidSynth Warning 002

FluidSynth Warning 003

oleg-derevenetz:
@LeHerosInconnu yes, I know - possibly because originally instruments database was in another format, and conversion to `.sf2` was performed not ideally, who knows. But at least it sounds the same (or very similar). But we can't use it anyway in our project.

oleg-derevenetz:
Hi @zenseii you once reported an issue (#5141) with popping sounds while playing MP3 tracks. When you have the opportunity and desire, could you check if you hear the popping sounds with build from this PR? Related to [this](https://github.com/ihhub/fheroes2/pull/5661#issuecomment-1186282888).

zenseii:
> Hi @zenseii you once reported an issue (#5141) related to the popping sounds while playing MP3 tracks.

Hi @oleg-derevenetz ! You'll be happy to know that none of that noise is there anymore with this build. I cross-checked with latest commit to master and there the noise is present as if someone lit up a bonfire in the background, so this PR is definitely better. Any noise left now is from my brain then.

As for the midi part, I've tested this and in general it sounds much better, but it has some quirks left.
Mostly it's mixing issues, like the mini organ being too loud in the main menu music, although when it's silent the music is very pleasant.
Then there's an issue with distorted noise/crackling sounds (in midi). This is very evident in the wizard castle music, where it seems like it's hitting some threshold for the bass. I don't think this is due to the soundfont. Not sure what. This is not noticeable on latest master commit with midi.

oleg-derevenetz:
Hi @zenseii, thank you very much!

> You'll be happy to know that none of that noise is there anymore with this build. I cross-checked with latest commit to master and there the noise is present as if someone lit up a bonfire in the background, so this PR is definitely better. Any noise left now is from my brain then. (In other words I can't hear any from the game).

Yeah, that's a good news. This means that SDL resampler really doesn't introduce distortions, so we can forget about that sampling rate mismatch.

> Then there's an issue with distorted noise/crackling sounds (in midi). This is very evident in the wizard's castle music, where it seems like it's hitting some threshold for the bass. I don't think this is due to the soundfont.

You can try the original Windows MIDI soundfont to compare. I don't want to post a direct link due to copyright concerns, but it can easily googled using the keywords "default windows midi soundfont". To my taste, if it is not 100% true to the original, it is very close.

zenseii:
@oleg-derevenetz 
> You can try the original Windows MIDI soundfont to compare. I don't want to post a direct link due to copyright concerns, but it can easily googled using the keywords "default windows midi soundfont". To my taste, if it is not 100% true to the original, it is very close.

I've tested this soundfont and the sound still pops/crackles a lot when the wizard castle track starts to become more "hectic" especially noticeable when the wind instruments do a sort of sforzando while the piano is accompanying with a very marked bass. Anyways it always happens at the same spots.

Yeah, it sounds very similar to the original midi. I have to admit that I'm not too familiar with the midi tracks, but, crackling sound, aside it is very hard to tell the difference.

oleg-derevenetz:
@zenseii

> I much prefer the 30 MB soundfont

Yeah, but 30MB is 30MB :)

ihhub:
> Hi @ihhub we will eventually need our own fork of `vcpkg` for better caching and dependency control. Currently I use my own [fork](https://github.com/oleg-derevenetz/vcpkg), but then (when my PR regarding `sdl2-mixer` will be eventually merged by Micro$oft people) we will need to make the project-wide mirror under [fheroes2](https://github.com/fheroes2) organization I suppose.
> 
> The idea is as follows: we make a fork and create a "snapshot" (by creating a tag or a branch) and use this snapshot in our own builds until we need to move to a different version of some library - then we create another "snapshot" and move to it. This approach provides stability, good cache hits and good build speed.
> 
> There is always a possibility that the compiler itself will change to a new version (in this case the cache will be rebuilt), but I suppose this should happen relatively rarely.

Looks like a good approach. Will it work only for builds or for developer's setup as well?

ihhub:
> @zenseii
> 
> > I much prefer the 30 MB soundfont
> 
> Yeah, but 30MB is 30MB :)
> 
> P.S. BTW the FluidSynth version in the vcpkg database is also not the latest one. Maybe with more recent version sound will become even better.

We could pack the soundfont into `resurrection.h2d` in a compressed format by zlib once we identify which soundfont we need. In this case the distribution of the game would be flawless as `resurrection.h2d` is required to be present.

oleg-derevenetz:
Hi @ihhub

> Looks like a good approach. Will it work only for builds or for developer's setup as well?

Currently this "vendor lock-in" works only for CI builds. Script that installs the dependencies uses the already existing vcpkg database, where the developer will get it is up to the developer.

> We could pack the soundfont into resurrection.h2d in a compressed format by zlib once we identify which soundfont we need.

I'd install the soundfont as a separate file because:

1. First and foremost, in this case users will have an opportunity to easily replace it by a new one if they want;
2. The soundfont will be compressed during the distribution anyway - both Inno Setup and ZIP files are compressed;
3. If we still need to compress the soundfont itself, we can use the SF3 format, which is compressed and also supported by FluidSynth.

We also can try to reduce its size by removing instruments that are unused in the game.

undef21:
Also various distributions already have these soundfonts. We need to provide soundfonts for windows only, right?

oleg-derevenetz:
Hi @undef21

> Also various distributions already have these soundfonts. We need to provide soundfonts for windows only, right?

Yes, that's true. On the rest of the OS there are not so many problems with native MIDI :)

LeHerosInconnu:
If I understood correctly, the problem is more related to Windows.
Why not use the .dls file already present in the "C:\Windows\System32\drivers\" folder then?
Apparently FluidSynth can load this type of file directly: https://github.com/FluidSynth/fluidsynth/pull/493.

undef21:
@oleg-derevenetz If we put a free soundfont to resurrection.h2d it can introduces the policy violation for various distros which require build all from source. In this case we'll have to build resurrection.h2d from sources.

oleg-derevenetz:
@undef21 Yes, I agree that it would be good to avoid this.

oleg-derevenetz:
Hi @ihhub I believe this PR is ready for review. The only remaining moments are:

1. Uncertainty about the specific soundfont file to be used. We can address this later separately;
2. Absence of our own vcpkg fork/mirror. This can be addressed later separately as well when my PR regarding `sdl2-mixer` port will be eventually merged into main vcpkg repo, but you can wait for it, it's up to you.

oleg-derevenetz:
@LeHerosInconnu

> BUT `libinstpatch` is explicitly [disabled](https://github.com/microsoft/vcpkg/blob/master/ports/fluidsynth/portfile.cmake) in vcpkg's portfile and not enabled back due to unknown reason.

That's probably because there is just no port for `libinstpatch` in the vcpkg database :) At least I was not able to find it.

oleg-derevenetz:
I have added support for SF3 and converted GeneralUser GS SoundFont to the SF3 format. Now it only occupies ~6MB instead of ~30MB.

oleg-derevenetz:
@ihhub

> complication of development process: we add another requirement to install extra software. Can it be done in another way or it is the only way? I believe an extra repository will fix it.

The only additional software required is `vcpkg`. We can of course add another one repo just to build the required libraries, but the dependency on prebuit binary blobs doesn't looks like open source way :) In other aspects, the process of installing required libs is fully automated.

> we will essentially break correctness of MIDI playback (I have to hear again as I got not time to run it). What if we couldn't fix it before the release giving users incorrect music? This could be bad. Can we set nativemidi by default till we find proper soundfont?

I wouldn't say that the sound is wrong, it's a little different, because the sound samples of each instrument are a little different, since these are not original instruments from Roland. BUT I should say that they are different on all these Linux systems, because they usually use some 3rd party sound fonts. Currently all that you should do to use native midi is to remove the `.sf2`/`.sf3` file, that's all.

ihhub:
@oleg-derevenetz , thanks for explanation! I will think about this overnight and we will do something tomorrow 🙂

oleg-derevenetz:
@ihhub

To the question of MIDI music on Linux:

https://github.com/ihhub/fheroes2/issues/4232#issuecomment-922626976

The first one is music on DOSBOX on Linux that uses one of these 3rd party sound fonts :)

ihhub:
Hi @oleg-derevenetz , I am thinking we need to have a separate repo to store all artifacts for Windows build: SDL 1 and SDL 2. We shouldn't compile them every time.
Right now in this pull request we use caching which is a great way but caching works per branch leading to long compilation times.

> JFYI: my PR, which updates the `sdl-mixer`, has been merged to the main [vcpkg repo](https://github.com/microsoft/vcpkg), so we can now use the main repo (or a fork from the main repo).

Is it correct that we don't need to compile SDL 2 anymore? If yes this is a step forward. However, should we just store needed artifacts on GitHub making less dependency on **vcpkg**?

Also for Windows users (which is the majority of our player base) I would recommend not to enable fluidsync by default.

files/soundfonts/README.txt
```diff
@@ -0,0 +1,5 @@
+This directory contains SoundFont files for MIDI synthesizer.
+======
+
+fheroes2.sf3 - SoundFont for MIDI synthesizer based on the GeneralUser GS
```
ihhub, line 4: How did we get `fheroes2.sf3`? Is it a copy-pasted file or we had to do some conversions / manipulations? Where it originates from? A small description how we got it would be useful.

src/engine/audio.cpp
```diff
@@ -983,7 +983,7 @@ bool Music::isPlaying()
 void Music::SetMidiSoundFonts( const ListFiles & files )
 {
     if ( files.empty() ) {
-        // No fonts to set
```
ihhub, line 986: To please our players we might introduce another audio option to switch between default system MIDI and provided soundfonts. It might solve the problem with sounds being close to the original. What do you think?

ihhub:
> `libinstpatch`

As I proposed earlier we can build everything by ourselves (using GitHub actions) and then store results in a separate info. It should be possible to do to cover everything. What do you think?

oleg-derevenetz:
@ihhub

> > `libinstpatch`
> 
> As I proposed earlier we can build everything by ourselves (using GitHub actions) and then store results in a separate info. It should be possible to do to cover everything. What do you think?

First we have to check whether it can be built for Windows at all and whether it will work there. What if it will not? This is just an additional possibility which still needs to be checked.

files/soundfonts/README.txt
oleg-derevenetz: @ihhub It is the same file but converted to the SF3 format. I can add a remark regarding this.

oleg-derevenetz:
@ihhub I can try to add a port for `libinstpatch` in the `vcpkg` repo and build `fluidsynth` with it, but it's not a fact that it will work or that it will help.

ihhub:
> @ihhub I can try to add a port for `libinstpatch` in the `vcpkg` repo and build `fluidsynth` with it, but it's not a fact that it will work or that it will help.

According to [this discussion](https://github.com/FluidSynth/fluidsynth/issues/535) it should be possible.

My apologies for asking too much. I just want to have close to a perfect solution for players and developers :(

ihhub:
@oleg-derevenetz , okay. Let's focus on what we have without `libinstpatch` as of now. The only question is left what to do with builds and artifacts. I am in favor compile them once and store somewhere from where any developer can download it with no need to compile on their end.

oleg-derevenetz:
@ihhub

> Aren't we doing the same by downloading demo version? :) After all we will be pointing to fheroes2 organization on GitHub so it's kinda safe I think... Or we can give an option to compile themselves or download from GitHub. This will solve all possible questions.

Well, then we need a repo for this.

oleg-derevenetz:
@ihhub

> I've sent you a request to `fheroes2` organization. Shall we name it as `fheroes2-artifacts`? Any other names are fine.

Maybe something like `fheroes2-prebuilt-deps`?

ihhub:
> @ihhub
> 
> > I've sent you a request to `fheroes2` organization. Shall we name it as `fheroes2-artifacts`? Any other names are fine.
> 
> Maybe something like `fheroes2-prebuilt-deps`?

Looks good to me.

oleg-derevenetz:
@ihhub

> to be clear: should I add the repo or you will do?

Already added :) I'll try to make it work later today.

ihhub:
> @ihhub
> 
> > to be clear: should I add the repo or you will do?
> 
> Already added :) I'll try to make it work later today.

Wonderful! My today is almost over :)

ihhub:
> @ihhub What we should do with the soundfonts? Should we add the setting to the InnoSetup to not install them (and therefore use the system MIDI)?

Yes, it would solve some issues for players who don't want to hear non-original sounds. We also need to specify some issues with this approach if possible.

Besides Windows we have another systems. Does MIDI work fine on MacOS?

ihhub:
@oleg-derevenetz , actually we have to install them along with the game. Please ignore my previous suggestion. We need to add an in-game option to switch between MIDI types.

oleg-derevenetz:
@ihhub

> actually we have to install them along with the game. Please ignore my previous suggestion. We need to add an in-game option to switch between MIDI types.

I suppose this can be done in a separate PR. Also please note that native MIDI is not available on all supported systems (at least Linux doesn't have it).

script/windows/install_packages.bat
```diff
@@ -1,67 +1,34 @@
 @echo off
 
-set tempPath="..\..\VisualStudio\packages\installed\temp"
+set DST_DIR=%~dp0\..\..\VisualStudio\packages
```
ihhub, line 3: We need to update **.gitignore** file as right now it filters `VisualStudio/packages/installed` path.

script/windows/fheroes2.iss
```diff
@@ -33,6 +31,7 @@ Source: "..\..\changelog.txt"; DestDir: "{app}"
 Source: "..\..\LICENSE"; DestDir: "{app}"
 Source: "..\..\files\lang\*.mo"; DestDir: "{app}\files\lang"
 Source: "..\..\files\data\*.h2d"; DestDir: "{app}\files\data"
+Source: "..\..\files\soundfonts\*.*"; DestDir: "{app}\files\soundfonts"
```
ihhub, line 34: We do not need to supply soundfonts for SDL1 as it doesn't support it. The same applies for normal zip archive.

script/windows/install_packages.bat
```diff
@@ -1,67 +1,34 @@
...
-powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/projects/SDL_mixer/release/SDL_mixer-devel-1.2.12-VC.zip', '%tempPath%\sdl_mixer\sdl_mixer.zip')"
-echo [5/6] Downloading packages
-powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-devel-2.0.4-VC.zip', '%tempPath%\sdl_mixer\sdl_mixer2.zip')"
-echo [6/6] Downloading packages
-powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/projects/SDL_image/release/SDL2_image-devel-2.0.5-VC.zip', '%tempPath%\sdl_image\sdl_image2.zip')"
+if not exist "%DST_DIR%" ( mkdir "%DST_DIR%" || exit /B 1 )
 
-xcopy /Y /S /Q "setup_packages.bat" "..\..\VisualStudio\packages\installed"
+echo Downloading %PKG_URL%                                                                                        && ^
+powershell -Command "%PKG_TLS%; (New-Object System.Net.WebClient).DownloadFile('%PKG_URL%', '%TEMP%\%PKG_FILE%')" || ^
```
ihhub, line 12: Can we do any CRC verification if possible? We can output SHA256 in release description within [fheroes2-prebuilt-deps](https://github.com/fheroes2/fheroes2-prebuilt-deps). This is to make such approach more trusted for developers.

ihhub (changes requested):
Hi @oleg-derevenetz , I put few more comments here. Would you mind to take a look at them?

.github/workflows/cmake.yml
```diff
@@ -63,6 +63,13 @@ jobs:
         shell: bash
     steps:
     - uses: actions/checkout@v2
+    - name: Prepare vcpkg cache
```
ihhub, line 66: Since we do not build these packages here do we need these changes?

oleg-derevenetz: @ihhub We do build them literally on the next line :)

script/windows/install_packages.bat
oleg-derevenetz: @ihhub The whole point of the CRC verification is to have the hash on the downloader's side, not on the side of a file being downloaded. There is no point to download hash with the file itself from the same source, because if this source is compromised, then both the file and its hash will be compromised, and CRC check will be useless.

ihhub: No-no. I meant to calculate CRC and compare it with a hardcoded value.

oleg-derevenetz: Honestly I would prefer to use the `vcpkg` directly while installing dependencies, because in this case all we would to when migrating to a new version of some library is to update the ref of the `vcpkg` snapshot. Now we should update the ref in our other repo, rebuild the binary package, then update the SHA256 hash here... A lot of extra actions. Meanwhile, most Open Source developers for Windows already have `vcpkg` installed on their machines.

Branikolog:
Hi, @oleg-derevenetz 
While using the above build I've experienced an application crash. Twice. Couldn't tell you right now, how exactly. The only thing I now, it crashed during composition switch. One crash I've experienced after crossing of the terrain. Another - pressing "New Game". Both MIDI.

oleg-derevenetz:
Hi @Branikolog

> While using the above build I've experienced an application crash. Twice. Couldn't tell you right now, how exactly. The only thing I now, it crashed during composition switch. One crash I've experienced after crossing of the terrain. Another - pressing "New Game". Both MIDI.

Any messages in the console? Also could you please try to remove the SF3 file and replace it by the original SF2 file from here:

http://www.schristiancollins.com/generaluser.php

(the latest file from the Download section) and see are the crashes still happen.

Branikolog:
@oleg-derevenetz 
It doesn't help.
There's nothing essential in console:
`fluidsynth: warning: Instrument not found on channel 9 [bank=128 prog=49], substituted [bank=128 prog=0]`
I've noticed, that the issue happens, when game switches between multiple compositions. Especially, when I enter various castles and select heroes on different terrains. Then, I select one hero and move him swiftly from one terrain to another and game freezes.
Funny thing: the game doesn't respond at all. But, when I try to resize the console - freeze is gone. o_O

oleg-derevenetz:
@Branikolog

> Funny thing: the game doesn't respond at all. But, when I try to resize the console - freeze is gone. o_O

This means that if you resize the console then you can just continue to play as usual?

undef21:
From the last build of this branch:

https://user-images.githubusercontent.com/71143197/180014203-f5681b48-c569-422d-896c-76041c7c0639.mp4
I move mouse before all clicks.

OS is Debian 10 amd_64.
When we are switching music type first time we have delay. Also midi music doens't continue playing after changing music types.

oleg-derevenetz:
@undef21

> OS is Debian 10 amd_64.

This branch changes nothing for Linux systems :)

Branikolog:
> This means that if you resize the console then you can just continue to play as usual without restarting the game?

Yes. Just changing the window of the console for at least 1px and game continues.

oleg-derevenetz:
@Branikolog

> Yes. Just changing the window of the console for at least 1px and game continues.

Then it looks like more an SDL-related issue. Anyway I was not able to reproduce it yet.

ihhub:
> > This means that if you resize the console then you can just continue to play as usual without restarting the game?
> 
> Yes. Just changing the window of the console for at least 1px and game continues.

Are you sure that you do not click on the console in meanwhile? Try to click on the console while game freezes and press Enter without resizing it.

Branikolog:
@ihhub 
No, clicking or minimizing console doesn't affect anything.
Oh!
Also forgot to mention! Freeze happens not in a moment. I move hero to a tile with other terrain and all sounds in game disappear But I still can interact with the game. But when I initiate any action, that produces sound: move hero, collect item, enter castle - game freezes.

ihhub:
> @ihhub No, clicking or minimizing console doesn't affect anything. Oh! Also forgot to mention! Freeze happens not in a moment. I move hero to a tile with other terrain and all sounds in game disappear But I still can interact with the game. But when I initiate any action, that produces sound: move hero, collect item, enter castle - game freezes.

Can you provide a save file to reproduce it?

undef21:
@oleg-derevenetz, of course. This bug is presented in the master branch. My mistake

script/windows/install_packages.bat
ihhub: I completely understand your point and it is a very solid one. I am thinking more from inexperienced / unmotivated developers perspective. The lower entry level / easier setup the higher chance more people would join development.

We can still leave the option to build using `vcpkg` on developer's machine. I wasn't against it.

oleg-derevenetz:
@undef21 on Linux It all depends on how exactly sdl_mixer is built to support the MIDI playback. The devil himself will break his leg there. Since Linux does not have native MIDI support, sdl_mixer can be built with timidity, maybe with fluidsynth, maybe with something else, all this of unknown versions and with unknown instrument sets.

script/windows/install_packages.bat
oleg-derevenetz: @ihhub Let's stick with pre-built binaries for now since we now have the infrastructure.

Branikolog:
[Sound prblms.zip](https://github.com/ihhub/fheroes2/files/9151639/Sound.prblms.zip)
Two more freezes occured when I pressed Save Game button and another case I even was out of PC... Not sure now it is related to current PR...
I am testing in window mode, if this helps....

oleg-derevenetz:
@Branikolog and in all cases resizing of the console window helps?

Branikolog:
> @Branikolog and in all cases resizing of the console window helps?

Yes. In 100% cases resize our maximizing console window helps.

oleg-derevenetz:
@Branikolog

> Yes. In 100% cases resizing or maximizing console window helps. 

LOL. I have the increasing impression that this is some SDL-related issue.

Branikolog:
> @Branikolog
> 
> > Yes. In 100% cases resizing or maximizing console window helps.
> 
> LOL. I have the increasing impression that this is some SDL-related issue.

Could be...

oleg-derevenetz:
@Branikolog since you are able to reproduce issue more or less reliably, could you please try to replace the `SDL2.dll` by the same file from the latest release and try this Frankenstein? When and if you have a free time OFC.

Branikolog:
> @Branikolog since you are able to reproduce issue more or less reliably, could you please try to replace the `SDL2.dll` by the same file from the latest release and try this Frankenstein creature? When and if you have a free time OFC.

I'll try tomorrow then...  Out of my PC now...

Branikolog:
@oleg-derevenetz 
While running the latest build above I see next message each time game tries to play new midi composition:
image

undef21:
@Branikolog try one of [this](https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170) libs

oleg-derevenetz:
@Branikolog

> While running the latest build above I see next message each time game tries to play new midi composition:

I suppose we will eventually need to add an option to install MSVC redistributables by the Inno Setup. But not in this PR :)

oleg-derevenetz:
`VCRUNTIME140_1.DLL` should not be needed anymore.

ihhub:
> > @Branikolog
> > > Yes. In 100% cases resizing or maximizing console window helps.
> > 
> > 
> > LOL. I have the increasing impression that this is some SDL-related issue.
> 
> Could be...

In this case could you try using latest snapshot?

oleg-derevenetz:
@ihhub

> In this case could you try using latest snapshot?

Did you manage to reproduce this issue? I tried a few dozen times yesterday using the provided savefile, no luck so far.

Branikolog:
Hi, guys.
I've downloaded the latest snapshot but couldn't reproduce freeze.
So, I've downloaded the current build once more and managed to somehow reproduce it again... But I still can't find the conditions to 100% reproduce the issue. I'm still trying to reproduce....
Btw. While gnawing the build I've experienced one more issue.
When we turn music off, we can hear the last sound of composition at the moment it stopped.

https://user-images.githubusercontent.com/55348946/180190739-bda32930-6561-4817-b565-361768198c89.mp4

oleg-derevenetz:
@Branikolog

> When we turn music off, we can hear the last sound of composition at the moment it stopped.

Yes, this issue is reproducible, but it's a relatively minor one. It's hard to say is this related to the FluidSynth itself or it is related to the SDL_Mixer (I mean, to the way that SDL_Mixer uses the FluidSynth's music output). I can try to build the game with the latest FluidSynth (it is currently not the latest version in the vcpkg portfile).

Branikolog:
@oleg-derevenetz 
I really don't know how it happens... It turns to be some magic trick or moon influence.
The game freezes even if I don't touch it. Bug can happen in a few minutes, or within an hour...
Here's a video of how I fix freeze...

https://user-images.githubusercontent.com/55348946/180233990-9d4d2356-dc07-49b1-bc0e-2a55f11967b3.mp4

oleg-derevenetz:
@Branikolog Then it looks more like some memory-related issue with some of these new libraries. Something gets corrupted somewhere and eventually it freeze  and then it can be "fixed" by a magical way by resizing the console. I can try to reduce the number of libraries by throwing out the SF3 support. Do you remember if there were similar problems with more early builds from this PR?

ihhub:
> @Branikolog Then it looks more like some memory-related issue with some of these new libraries. Something gets corrupted somewhere and eventually it freeze and then it can be "fixed" by a magical way by resizing the console. I can try to reduce the number of libraries by throwing out the SF3 support. Do you remember if there were similar problems with more early builds from this PR?

My theory is that the focus of the application switches to the console. You can replicate this my simplify selecting some text inside the console and then try to do something in UI.

oleg-derevenetz:
@ihhub Then it should be visible - window headers of both console window and game window should change colors.

ihhub:
@oleg-derevenetz , aha! I found why it freezes at least for my case: it tries to access DEBUG_LOG which is busy at the moment...

Branikolog:
Unfortunately, I didn't use the most early builds.
But I've tried all the stuff you and @ihhub proposed me. Switching SDL@.dll didn't helped, freeze occurred.
Switching to different .sf file didn't helped.
Although, when deleting .sf file or running master branch bug doesn't occur.

ihhub:
Hi @Branikolog , during the freeze please try to just make focus on the console window and press Enter without resizing it. Does it help with the freeze fix?

oleg-derevenetz:
@ihhub Interesting. But `ERROR_LOG` is just a print to `cerr` which should be safe. How it can be busy?

oleg-derevenetz:
@ihhub Yes of course.

ihhub:
> @ihhub Interesting. But `ERROR_LOG` is just a print to `cerr` which should be safe. How it can be busy?

I have never seen such behavior... maybe to be safe we need to save all logs in a separate file. It also prevents from logs being lost.

ihhub:
> @oleg-derevenetz I really don't know how it happens... It turns to be some magic trick or moon influence. The game freezes even if I don't touch it. Bug can happen in a few minutes, or within an hour... Here's a video of how I fix freeze...
> 
>  2022-07-21.17-01-27.mp4

Hi @oleg-derevenetz and @Branikolog , from the video you can see once the window is resized one more error log comes from fluidsynth.

ihhub:
> @ihhub @Branikolog Well, the new build without `cerr` is ready. But what if `fluidsynth` will conflict with itself?

It has logging options, we could try to set them if possible to log into a separate file with a mutex lock if it uses multithreading (which I doubt).

Anyway, @Branikolog , could you please test the newest build for freezes?

oleg-derevenetz:
To be honest, all these freezes on `cout`/`cerr` doesn't look normal at all. This should not happen at all, it should be safe to concurrently access these objects from multiple threads, this is guaranteed by the standard.

oleg-derevenetz:
Interesting, who is waiting on `stdin` at all? It looks like SDL messages are not pumped while someone is waiting on `stdin` or sort of (BTW this may explain why `ERROR_LOG` waits on `cerr` - it may be because someone else holds some stdio lock at the time waiting for console input). If we find the one waiting on `stdin`, then we will find the killer (C). The matter is complicated by the fact that we have a lot of additional libraries, for example, now we have `GLIB` as a dependency.

oleg-derevenetz:
I have added PDB files to the pre-built package to make debugging easier.

Branikolog:
@ihhub 
I'll check it in a few hours. 
One more thing... 
On my home PC with win8 this fluidsynth seem to be not working...  I hear standard midi sounds, not from new library...

Branikolog:
> It means that either some of required libraries (including MSVC redistributable libraries) was not loaded, or the soundfont was not found. Is it the same system that complained about the missing DLL?

Yes.

oleg-derevenetz:
@Branikolog There is a handy open source [tool](https://github.com/lucasg/Dependencies/releases) able to traverse the DLL dependencies for a given file and display where the corresponding DLL will be taken from (as well as missing DLLs). Here, for example, I renamed the `libmp3lame.dll` and loaded the `libfluidsynth-3.dll` by this tool:

dependencies

You can try to use this tool to sort out why the FluidSynth DLL isn't loaded on your home Win8 PC.

Branikolog:
@ihhub 
>Please confirm that after a freeze and unfreezing a new long message from fluidsynth is being added into the console?

image

Unfortunately no.  The game now freezes like a day before: on a world map.

But good news... I found a way to reproduce the bug in 80% of cases. So I can at least test changes you make....

oleg-derevenetz:
@Branikolog

> But good news... I found a way to reproduce the bug in 80% of cases.

How this can be done?

Branikolog:
@oleg-derevenetz 
I don't know, what actions are not necessary here... Please, try not to laugh, as it looks looks a shaman gestures.)
It seems to me, that freeze occurs when you load new map, step hero on a desert terrain and then click on the black field of the console window itself. Wait a few minutes and when you'll return to main window it would be freezed.

Branikolog:
@oleg-derevenetz 
Bingo! The problem is in console window! All previous builds of the last few days freeze, when we change focus on console in windowed mode on a tracks, where fluidsynth shows error message.

oleg-derevenetz:
@Branikolog Yes, I now was able to reproduce it. Please tell me, do you have "Select by mouse" turned on in the console window properties?

console

If yes, please try to disable this. If this is turned on, then the console will not update until the selection will cancelled, and by clicking to the console window, you are starting the selection process.

Branikolog:
@oleg-derevenetz 
Btw, Game freezes on master branch too in a case there appears any message.

oleg-derevenetz:
@Branikolog

> Btw, Game freezes on master branch too in a case there appears any message.

Yes, if I understood the mechanism correctly, then this is related purely to the console window, and it is unrelated to the game itself.

oleg-derevenetz:
So, basically, the following happens:

1. When you click on the console window, the text selection process is started and Windows pauses all output to the console;
2. When the game prints some message to the paused console, it pauses as well, because it is waiting for the console being able to print the message;
3. By pressing Enter or resizing the console window the selection process is cancelled, console output is resumed and the game resumes as well.

Branikolog:
@oleg-derevenetz It definitely works.
I think we should do something with console window: hide it or disable mouse selection.) As it can confuse players, as I was. Actually, I've experienced rare accidental freezes through the long period of time on running the game and opening save file list, when some messages regarding old save files appear in console. I've only understood it now.

oleg-derevenetz:
@Branikolog the console version will be definitely disabled closer to release I believe (I suppose we can just redirect `stdout`/`stderr` to some logfile to catch messages originated from a code we cannot control - such as different 3rd-party libs). Regarding disabling the mouse selection... I don't know really. Even if we turn if off, the user can always turn it back on.

oleg-derevenetz:
JFYI: I just reverted that temporary commit that disables logging to `cerr`.

Branikolog:
@oleg-derevenetz 
Thank you for the patience with this issue!

oleg-derevenetz:
@Branikolog

> Thank you for the patience with this issue!

That was funny :)

ihhub:
Hi @oleg-derevenetz , I assume that the bug with music off is still present and this is the only bug left?

oleg-derevenetz:
Hi @ihhub do you mean that bug when resumed music plays for a split second despite the zero volume? Yes, it is still present. I believe it's a bug in SDL_mixer and if we will be able to make a minimum working PoC, then we can create an issue in their repo.

ihhub:
> Hi @ihhub do you mean that bug when resumed music plays for a split second despite the zero volume? Yes, it is still present. I believe it's a bug in SDL_mixer and if we will be able to make a minimum working PoC, then we can create an issue in their repo.

Yes. If besides this we covered everything else I suggest to proceed with merge as a wider audience of players can test this huge change before the release.

ihhub (approved):

ihhub:
@oleg-derevenetz . amazing work!

PR #5635, opened 2022-07-12
Add support for loading of MIDI sound fonts

oleg-derevenetz:
Prerequisite for #5628

Also minor style nits from the same #5628.


ihhub:
@oleg-derevenetz , a bunch of thanks for polishing audio related code!

PR #5637, opened 2022-07-12
When adding a custom spellbook while loading a hero from MP2, check that the hero does not have it yet

### LINKED ISSUE #5633: Scenario selection window, spell book purchase bug
### Preliminary checks

### Platform

Windows

### Describe the bug

Sandro can't purchase a spell book due to full inventory... But this happens before game starts. :D
image
Bug occurs not each start. Just roll 5-10 times new game on this map.

### Save file

[Map_0248 bug.zip](https://github.com/ihhub/fheroes2/files/9091597/Map_0248.bug.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5633

The issue scenario is as follows:

1. The game loads a hero with a specified portrait (Sir Gallant);
2. If this hero is already busy (already hired or jailed), the game tries to get a freeman hero and "remake" it into the right race and set a custom portrait;
3. If this freeman hero already have a spellbook because of his former race, and also he has a "custom" spellbook as an artifact, then the error from #5633 will be observed, because there will be an attempt to add the spellbook twice.

Also this PR fixes an issue when old default spell may remain after changing the hero's race for the custom hero, like here:

Снимок экрана 2022-07-13 в 00 35 58

Custom Knight has the Stoneskin spell, "inherited" from the Wizard.

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -254,24 +254,23 @@ Heroes::Heroes( int heroid, int rc )
 
 void Heroes::LoadFromMP2( int32_t map_index, int cl, int rc, StreamBuf st )
 {
-    // reset modes
     modes = 0;
 
     SetIndex( map_index );
     SetColor( cl );
 
-    // unknown
+    // Unknown
     st.skip( 1 );
 
-    // custom troops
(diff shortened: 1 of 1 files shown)


ihhub:

PR #5640, opened 2022-07-13
Update image sizes on the published web page

ihhub:
related to #5623

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,11 +7,11 @@ improvements in the gameplay, graphics and logic (including support for high-res
 and UI improvements), breathing new life into one of the most addictive turn-based strategies.
 
 <p align="center">
-    <img src="docs/images/screenshots/screenshot_world_map.png?raw=true" width="820">
+    <img src="images/screenshots/screenshot_world_map.png?raw=true" width="820">
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -7,10 +7,11 @@ improvements in the gameplay, graphics and logic (including support for high-res
 and UI improvements), breathing new life into one of the most addictive turn-based strategies.
 
 <p align="center">
-    <img src="images/screenshots/screenshot_world_map.png?raw=true" width="410"> <img src="images/screenshots/screenshot_castle.png?raw=true" width="410">
+    <img src="docs/images/screenshots/screenshot_world_map.png?raw=true" width="820">
```
oleg-derevenetz, line 10: Since this `README.md` is in `docs`, we don't need to add `docs/` here and on the next similar line.

oleg-derevenetz (changes requested):
Hi @ihhub it seems that file paths are now wrong.

docs/README.md
ihhub: Yes, you were right. I fixed it.

zenseii (commented):
Hi, @ihhub.
I feel that the picture showing the adventure map is too small, at least when viewed from my PC:

<details>

image

</details>

Also the third and last picture is larger than the other two above. Do we scale these images at integers so that the pixels are as sharp as possible?

ihhub:
Hi @zenseii , this pull request addresses these issues. The third screenshot is bigger because of fixed size of the image set in html format.

zenseii (approved):

Issue #1211, opened 2020-08-13
Selecting Game Saves


LeHerosInconnu:
Hello to all the fheroes2 team,
Let me briefly introduce myself, "LeHerosInconnu" passionate about Heroes 2.
(Unknown Hero on Forums of Enroth https://heroes2.forumactif.com/f52-free-heroes-of-might-and-magic-ii-fheroes2).
sirDranik advised me to post some of my suggestions and ideas there.
So here is the first one.

I suggest that in the list of save files the most recent save be displayed at the top of the list, currently it's the other way around.

![Selecting Game Saves 01](https://user-images.githubusercontent.com/43583994/90146801-cc84db00-dd81-11ea-8fec-1af506c3cda8.png)
You can read the discussion on this here:
https://heroes2.forumactif.com/t934-suggestions-selecting-game-saves#8444

ihhub:
Hi @LeHerosInconnu , we sort all saves in alphabetic order. Since this window has limited space for save file names it feels like they're sorted by date. Shall we close this issue?

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu , we sort all saves in alphabetic order. Since this window has limited space for save file names it feels like they're sorted by date. Shall we close this issue?

It would be much more convenient if the last save was on top of the list and selected by default.
So, the user only needs to press the "OKAY" button to load the last save.

As for the limited space for the name of the save file, it's been on my list for a long time and it will be necessary to increase the size of this window one day or another because it needs it. :) (https://github.com/ihhub/fheroes2/discussions/4921)

In fheroes2:

![Load window size 001](https://user-images.githubusercontent.com/43583994/149001003-d07fa89b-afd5-49aa-94ac-fcb129fc40bf.png)

![Load window size 002](https://user-images.githubusercontent.com/43583994/149001029-74e8d880-acd0-4bcc-a427-26a6514287ef.png)

ihhub:
This issue was addressed in #9793

PR #5641, opened 2022-07-13
Always select the newest save in load menu

ghost:
Currently, if the player hasn't saved anything yet, the file at the top is selected. This change will make the newest save game be selected

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -293,7 +293,8 @@ std::string SelectFileListSimple( const std::string & header, const std::string 
             }
             listbox.Unselect();
         }
-    } else {
+    }
+    else {
         MapsFileInfoList sorted = lists;
         std::sort( sorted.begin(), sorted.end(), []( const Maps::FileInfo & a, const Maps::FileInfo & b ) -> bool { return a.localtime > b.localtime; } );
         listbox.SetCurrent( std::distance( lists.begin(), std::find( lists.begin(), lists.end(), sorted.front() ) ) );


ihhub:
Hi @Verlf-Cerf , I believe this change should be discussed first.
Assume that I named a file zzz which would be at the end of 100 files list. Then I want to load the last autosave file which would be at the too of the list. These are two edge cases. It also rises a question: why not autosave?

Both ideas have pros and cons. In every situation like this we usually create a discussion to find the best solution for everyone.

Branikolog:
Hi, @ihhub , @Verlf-Cerf 
The last save often goes close to the autosave regarding game progress. So players tend to check autosave first, before they want to load last made save file. Showing the save file list from not the beginning is not the most accurate solution, while running the engine for the first time. 
I think, changing the sorting of the list (regarding autosave) would be compromise here. We can place AUTOSAVE file in the beginning of the list and make it selected as default for the first application run. This will show player the date, the game was played the last time, so he can browse the whole list to find another save file with similar date-time.

LeHerosInconnu:
Hello @Verlf-Cerf, @ihhub and @Branikolog,

I'm also in favor of sorting the save files according to the last save in time (so the last save is always at the top of the list and selected by default), including the Autosave file.
The player may possibly make one or more final saves and quit the game before ending the current turn.

It would also be nice to allow the user to choose the method of sorting the files.
For example, with a setting in the fheroes2.cfg file:

"# sorting order for save files: 0 - by time, 1 - by alphabetical order
sorting order = 0"

It is also possible to add one or more buttons to change the sorting directly in the load/save file list window. :)

This is also one of my proposals here: https://github.com/ihhub/fheroes2/issues/1211.
With a discussion about it here: https://heroes2.forumactif.com/t934-suggestions-selecting-game-saves.

PR #5649, opened 2022-07-14
Revert resolution changes and add checks for rendering

ihhub:
Also fixed missing resizing ability.

Apparently, the previous changes for resolutions are invalid as they do not add a truthful full screen mode.

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -197,7 +197,7 @@ namespace
             if ( SDL_MUSTLOCK( surface ) ) {
                 const int returnCode = SDL_LockSurface( surface );
                 if ( returnCode < 0 ) {
-                    ERROR_LOG( "Failed to set lock surface. The error value: " << returnCode << ", description: " << SDL_GetError() )
+                    ERROR_LOG( "Failed to lock surface. The error value: " << returnCode << ", description: " << SDL_GetError() )
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/engine/screen.cpp
```diff
@@ -773,16 +729,24 @@ namespace
...
 
+            bool fullScreen = true;
             uint32_t flags = SDL_GetWindowFlags( _window );
             if ( ( flags & SDL_WINDOW_FULLSCREEN ) == SDL_WINDOW_FULLSCREEN || ( flags & SDL_WINDOW_FULLSCREEN_DESKTOP ) == SDL_WINDOW_FULLSCREEN_DESKTOP ) {
-                flags = 0;
+#if defined( __WIN32__ )
+                flags ^= SDL_WINDOW_FULLSCREEN;
+#else
+                flags ^= SDL_WINDOW_FULLSCREEN_DESKTOP;
+#endif
```
oleg-derevenetz, line 739: Why not the usual `&= ~` operation? The potential issue with XOR is that it is able to SET the flag if it is not set yet, while AND NOT is never able to set it.

oleg-derevenetz (commented):
Hi @ihhub I just left one question, please take a look when it's convenient for you.

src/engine/screen.cpp
ihhub: I have no idea why I decided to do this way. Good catch!

oleg-derevenetz (approved):

PR #5655, opened 2022-07-14
Remove virtual from Troop::GetCount()

oleg-derevenetz:
It is not overridden anywhere and only brings confusion.


idshibanov (approved):

ihhub:
@oleg-derevenetz , thanks for polishing the code!

Issue #5653, opened 2022-07-14
ARM (Raspberry Pi4) build fails: link error undefined reference to `__atomic_load_8'


imiron13:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Linux

### Describe the bug

I tried to build the game (0.9.17) for Raspberry pi 4 (Raspberry Pi OS, Debian based) following the [this instruction](https://github.com/ihhub/fheroes2/blob/master/docs/DEVELOPMENT.md).
It fails on a linking stage with the following error:
```
/usr/bin/ld: ../engine/libengine.a(audio.o): in function `(anonymous namespace)::musicFinished()':
audio.cpp:(.text+0x160): undefined reference to `__atomic_load_8'
/usr/bin/ld: ../engine/libengine.a(audio.o): in function `(anonymous namespace)::playMusic(unsigned long long, Music::PlaybackMode)':
audio.cpp:(.text+0x129c): undefined reference to `__atomic_fetch_add_8'
/usr/bin/ld: audio.cpp:(.text+0x14c0): undefined reference to `__atomic_fetch_add_8'
/usr/bin/ld: ../engine/libengine.a(audio.o): in function `(anonymous namespace)::MusicRestartManager::executeTask()':
audio.cpp:(.text+0x18dc): undefined reference to `__atomic_load_8'
/usr/bin/ld: ../engine/libengine.a(audio.o): in function `Music::Stop()':
audio.cpp:(.text+0x2e64): undefined reference to `__atomic_fetch_add_8'
collect2: error: ld returned 1 exit status
```
Compiler: g++ (Raspbian 10.2.1-6+rpi1) 10.2.1 20210110

I managed to finish the build by adding '-latomic' linker flag. I am not familiar with CMake so I just restarted linking with extra flag using a shell command.
Need to add this flag in the build system in order to support ARM/Raspberry Pi build.

See also:
https://github.com/ihhub/fheroes2/issues/4353

BTW, it works smoothly on Raspberry Pi 4. Haven't noticed any issues playing for ~30 mins. Quite enjoyable.

### Save file

N/A

imiron13:
> 

[...]

PR #5659, opened 2022-07-15
Yet another Makefile improvement

undef21:
1. Local system options should be passed via {CPP,CXX,LD}FLAGS;
2. According ld documentation the order of object files and libraries are matter, so now obj files are placed before libs.

linked issue #5653


src/Makefile
```diff
@@ -166,7 +168,7 @@ endif
 CCFLAGS := $(CCFLAGS) $(SDL_FLAGS)
 LIBS := $(SDL_LIBS) $(LIBS)
 
-export CC CXX AR WINDRES CCFLAGS CFLAGS CXXFLAGS CPPFLAGS LDFLAGS LIBS PLATFORM
+export CC CXX AR WINDRES CCFLAGS CFLAGS CXXFLAGS CPPFLAGS LDFLAGS LDLIBS LIBS PLATFORM
```
oleg-derevenetz, line 171: Hi @undef21

Since we already have `LIBS` variable:

https://github.com/ihhub/fheroes2/blob/dc5ff9ce45c7bc6fb37212033f45dc28fa961eae/src/Makefile#L105

we can just do like this:

```
LIBS := $(LIBS) -lz $(LDLIBS)
```

and that's all. There will be no need to propagate `LDLIBS` further. What do you think?

src/dist/Makefile
```diff
@@ -54,7 +46,7 @@ all: $(TARGET)
 
 $(TARGET): $(notdir $(patsubst %.cpp, %.o, $(SEARCH))) $(LIBENGINE) $(RES)
 	@echo "lnk: $@"
-	$(CXX) $(LDFLAGS) $^ $(LIBS) -o $@
+	$(CXX) -o $@ $^ $(LIBS) $(LDFLAGS)
```
oleg-derevenetz, line 49: @undef21

Is this change because some people like to specify additional libs in `LDFLAGS`? :) I believe that currently there will be no harm, but I remember some linkers in the past that has position-dependent `-L` option, i.e. in this case:

```
-L/some/path/1 -llib1 -L/some/path/2 -llib2
```

they were looking for `lib2` in both `/some/path/1` and `some/path/2`, but they were looking for `lib1` just in `/some/path/1`. I hope they are already extinct :)

undef21: > Is this change because some people like to specify additional libs in `LDFLAGS`?

Yes, it's this case.

ihhub:
@undef21 , thank you very much for this improvement!

PR #5661, opened 2022-07-16
Update SDL to 2.0.22 version for Windows

ihhub:
- update SDL library from 2.0.12 to 2.0.22 for Windows
- add more error checks for rendering
- add TODO comments for the future rendering enhancement

### DIFF
--- a/script/windows/install_packages.bat
+++ b/script/windows/install_packages.bat
@@ -12,7 +12,7 @@ xcopy /Y /Q "..\..\VisualStudio\packages\zlib1.2.11.zip" "%tempPath%\zlib\"
 echo [2/6] Downloading packages
 powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/release/SDL-devel-1.2.15-VC.zip', '%tempPath%\sdl\sdl.zip')"
 echo [3/6] Downloading packages
-powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/release/SDL2-devel-2.0.12-VC.zip', '%tempPath%\sdl\sdl2.zip')"
+powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/release/SDL2-devel-2.0.22-VC.zip', '%tempPath%\sdl\sdl2.zip')"
 echo [4/6] Downloading packages
 powershell -Command "(New-Object System.Net.WebClient).DownloadFile('https://www.libsdl.org/projects/SDL_mixer/release/SDL_mixer-devel-1.2.12-VC.zip', '%tempPath%\sdl_mixer\sdl_mixer.zip')"
 echo [5/6] Downloading packages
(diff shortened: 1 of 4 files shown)


ihhub:
Hi @oleg-derevenetz , do you observe the error with audio frequency on the last SDL Mixer version?

oleg-derevenetz:
@ihhub yes, on Windows the frequency is always 48000.

ihhub:
This is strange as this error doesn't appear on the older version of SDL.

oleg-derevenetz:
@ihhub

> This is strange as this error doesn't appear on the older version of SDL.

Maybe the older version of SDL just don't try to get the "real" frequency after an attempt to set a new one? :)

oleg-derevenetz:
@ihhub

> This is strange as this error doesn't appear on the older version of SDL.

This error is caused by [this code](https://github.com/libsdl-org/SDL/blob/cdf9010a19512dc84f6770d3617b848f5848876f/src/audio/wasapi/SDL_wasapi.c#L461-L470) in SDL:

```cpp
#if 1  /* we're getting reports that WASAPI's resampler introduces distortions, so it's disabled for now. --ryan. */
    this->spec.freq = waveformat->nSamplesPerSec;  /* force sampling rate so our resampler kicks in, if necessary. */
#else
    /* favor WASAPI's resampler over our own */
    if (this->spec.freq != waveformat->nSamplesPerSec) {
        streamflags |= (AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM | AUDCLNT_STREAMFLAGS_SRC_DEFAULT_QUALITY);
        waveformat->nSamplesPerSec = this->spec.freq;
        waveformat->nAvgBytesPerSec = waveformat->nSamplesPerSec * waveformat->nChannels * (waveformat->wBitsPerSample / 8);
    }
#endif
```

If I [replace](https://github.com/oleg-derevenetz/vcpkg/compare/sdl2-mixer-2.6.1...oleg-derevenetz:vcpkg:test) `#if 1` by `#if 0`, everything works as before:

https://github.com/oleg-derevenetz/fheroes2/actions/runs/2682976958#artifacts

SDL people claim that their resampler don't introduce distortions (unlike the WASAPI resampler). Are you sure about these distortions with rate 48000 and the latest SDL?

oleg-derevenetz:
> SDL people claim that their resampler doesn't introduce distortions (unlike the WASAPI's resampler).

@zenseii tells that there is no noise even with mismatched rates, so it is possible to just turn the corresponding `ERROR_LOG` into the `DEBUG_LOG( DBG_ENGINE, DBG_WARN )` and left it as is.

ihhub:
I found the source of audio issue for me: I set 3D audio on and when I open this save file water sounds are sometimes played in a way like the sound is distorted:
[-music.zip](https://github.com/ihhub/fheroes2/files/9127032/-music.zip)

I will make this ERROR as LOG to avoid player's confusion.

oleg-derevenetz (approved):

oleg-derevenetz:
Hi @ihhub what about that issue with cursor?

ihhub:
> Hi @ihhub what about that issue with cursor?

It's miraculously fixed. Probably they fixed it in the subsequent releases.

PR #5672, opened 2022-07-19
Reveal castle info to AI if it's partially in fog

idshibanov:
Right now AI is at disadvantage when it can ignore a castle if it's entrance isn't in view. Added a check on new tile revealed to them.


ihhub:
I have two questions regarding this approach:
- why don't we reveal all tiles where castle stands as AI cannot calculate a path to one tile surrounded by fog?
- aren't we making AI cheat by revealing the entrance of the castle? AI could see whether a player stands there.

ihhub:
Hi @idshibanov , could you please answer the question above when you have free time?

idshibanov:
I think you're correct, we'll need another approach here. Closing for now.

PR #5678, opened 2022-07-20
Show correct hero's level and experience on level up

### LINKED ISSUE #5676: Hero screen, hero's level and experience points are not updated when the hero gains levels and the player accesses the hero screen
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the hero's level and experience points are not updated when the hero gains levels and the player accesses the hero screen.

In fheroes2:

https://user-images.githubusercontent.com/43583994/179813117-f01072fa-2d91-4b93-a9b6-84832d1f8f97.mp4

### Save file

Not needed.

### Additional info

_No response_

ihhub:
close #5676

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -1007,10 +1007,18 @@ void Heroes::IncreaseExperience( const uint32_t amount, const bool autoselect )
     int oldLevel = GetLevelFromExperience( experience );
     int newLevel = GetLevelFromExperience( experience + amount );
 
-    for ( int level = oldLevel; level < newLevel; ++level )
-        LevelUp( false, autoselect );
+    const uint32_t updatedExperience = experience + amount;
 
-    experience += amount;
+    for ( int level = oldLevel; level < newLevel - 1; ++level ) {
+        experience = GetExperienceFromLevel( level );
+        LevelUp( false, autoselect );
+    }
+
+    experience = updatedExperience;
+
+    if ( newLevel > oldLevel ) {
+        LevelUp( false, autoselect );
+    }
 }
 
 /* calc level from exp */


src/fheroes2/heroes/heroes.cpp
```diff
@@ -1007,10 +1007,18 @@ void Heroes::IncreaseExperience( const uint32_t amount, const bool autoselect )
     int oldLevel = GetLevelFromExperience( experience );
     int newLevel = GetLevelFromExperience( experience + amount );
 
-    for ( int level = oldLevel; level < newLevel; ++level )
+    const uint32_t updatedExperience = experience + amount;
+
+    for ( int level = oldLevel; level < newLevel - 1; ++level ) {
+        experience = GetExperienceFromLevel( level );
```
oleg-derevenetz, line 1013: Hi @ihhub

Maybe I'm wrong here, but suppose that `experience` is 1800 now, `amount` is 2030 and the levels go like this:

```
1 ( 0 - 999 ) -> 2 ( 1000 - 1999 ) -> 3 ( 2000 - 2999 ) -> 4 ( 3000 - 3999 )
```

So currently hero's level is 2. Thus, when we calling `GetExperienceFromLevel()` for the first time, we will get 1000 experience instead of 1800, right?

ihhub: No, it will be 2030 because difference between levels is 1 so the condition `level < newLevel - 1` is **false**. However, if the new experience, let's say 3500, then for the first call of **LevelUp()** we will show 2000, then for the next we will show 3500.

ihhub: Hmm... Let me check why the logic is wrong but we display correctly.

oleg-derevenetz: Ah OK then. But of course name of `GetExperienceFromLevel()` looks misleading then.

PR #5684, opened 2022-07-22
fix build for VS2017 and Win8.1 PSDK

undef21:


src/thirdparty/libsmacker/smacker.h
```diff
@@ -60,11 +60,11 @@ extern "C" {
 
 /* OPEN OPERATIONS */
 /** open an smk (from a file) */
-smk smk_open_file(const char * filename, unsigned char mode);
+smk smk_open_file(const char * filename, const unsigned char mode);
```
oleg-derevenetz, line 63: Hi @undef21 are these changes in `smacker.h` essential for the older compilers? If no, please revert them, because it's a 3rd-party code and we should avoid make changes in this code unless absolutely necessary. If they are essential, please make a copy of the original file with the name `smacker.h.orig` (just like it done with `smacker.c`) so that in the future we will be able to immediately determine what changes have been made compared to the original version.

undef21: @oleg-derevenetz these changes fix the next errors:
error

oleg-derevenetz: @undef21 Ah OK, then please make a copy of the original file, like it's done with `smacker.c`.

oleg-derevenetz (approved):

ihhub:
@undef21 , thank you so much for this fix!

PR #5697, opened 2022-07-25
Fix experience from Tree of Knowledge

### LINKED ISSUE #5695: Adventure map, Tree of Knowledge, wrong amount of experience points given to the hero
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, there is a wrong amount of experience points given to the hero from the Tree of Knowledge.

In the original game, the Tree of Knowledge gives the number of experience points needed to progress from one level to the next level.
For example, if the hero is at level 1, he will receive 1000 experience points.
In the example, the hero is at level 1, he already has 999 experience points, and he receives 1000 experience points from the Tree of Knowledge.

In fheroes2, the hero receives only the number of experience points missing to progress to the next level.
In the example, the hero is at level 1, he already has 999 experience points, and he receives only 1 experience point from the Tree of Knowledge.

In original game:

Hero experience 002

Hero experience 005

Hero experience 007

In fheroes2:

Hero experience 009

Hero experience 010

Hero experience 012

### Save file

[0 9992a.zip](https://github.com/ihhub/fheroes2/files/9182786/0.9992a.zip)

### Additional info

Version 0.9.17 5674

Experience Points needed to reach the Hero Levels:
https://mightandmagic.fandom.com/wiki/Experience_(Heroes)#Heroes_of_Might_and_Magic_II

undef21:
fix #5695

### DIFF
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -2733,6 +2733,9 @@ void ActionToTreeKnowledge( Heroes & hero, const MP2::MapObjectType objectType, 
         bool conditions = 0 == funds.GetValidItemsCount();
         std::string msg;
 
+        const int level = hero.GetLevel();
+        const uint32_t possibleExperience = Heroes::GetExperienceFromLevel( level + 1 ) - Heroes::GetExperienceFromLevel( level );
+
         // free
         if ( conditions ) {
             msg = _(
@@ -2753,8 +2756,6 @@ void ActionToTreeKnowledge( Heroes & hero, const MP2::MapObjectType objectType, 
                 msg.append( _( "(Just bury it around my roots.)" ) );
                 StringReplace( msg, "%{res}", Resource::String( rc.first ) );
                 StringReplace( msg, "%{count}", rc.second );
-
-                const uint32_t possibleExperience = Heroes::GetExperienceFromLevel( hero.GetLevel() ) - hero.GetExperience();
 
                 const fheroes2::ExperienceDialogElement experienceUI( static_cast<int32_t>( possibleExperience ) );
                 const fheroes2::Text titleUI( title, fheroes2::FontType::normalYellow() );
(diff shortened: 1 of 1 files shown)


src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2733,6 +2733,9 @@ void ActionToTreeKnowledge( Heroes & hero, const MP2::MapObjectType objectType,
         bool conditions = 0 == funds.GetValidItemsCount();
         std::string msg;
 
+        const int level = hero.GetLevel();
+        const uint32_t possibleExperience = Heroes::GetExperienceFromLevel( level + 1 ) - Heroes::GetExperienceFromLevel( level );
```
oleg-derevenetz, line 2737: Hi @undef21 as it [turned out recently](https://github.com/ihhub/fheroes2/pull/5678#discussion_r926491549) the name of `GetExperienceFromLevel ()` is misleading - it returns not the value needed to reach the specified level, but the value that is needed to reach the NEXT level after the specified, so for example `GetExperienceFromLevel( 1 )` returns 1000, and with 1000XP the hero will become a level 2 hero. So I suppose you need to set the `possibleExperience` just to `Heroes::GetExperienceFromLevel( hero.GetLevel() )`.

undef21: @oleg-derevenetz  we need difference between levels. GetLevel(10) returns 15500 exp :)

oleg-derevenetz: @undef21 But not between `level + 1` and `level`. For example if our hero is of level 2, `Heroes::GetExperienceFromLevel( level + 1 )` will return 3200 points and `Heroes::GetExperienceFromLevel( level )` will return 2000 points, so the difference will be 1200 points, while he will need just 1000 points (2000-1000) to progress to level 3, right?

undef21: Fixed:)

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2733,6 +2733,9 @@ void ActionToTreeKnowledge( Heroes & hero, const MP2::MapObjectType objectType,
         bool conditions = 0 == funds.GetValidItemsCount();
         std::string msg;
 
+        const int level = hero.GetLevel();
```
oleg-derevenetz, line 2736: I would add the following assertion here:

```
assert( level > 0 );
```

I may be wrong, but `GetExperienceFromLevel()` may go to the infinite recursion if level is negative.

oleg-derevenetz (changes requested):

ihhub:
@undef21 , thanks a lot for this fix!

PR #5704, opened 2022-07-27
MSVC workflow: use 7zip instead of Info-ZIP

oleg-derevenetz:
Fix persistent ZIP installation issues.


ihhub:

PR #5705, opened 2022-07-29
Fix build with earlier Windows SDKs, copy soundfonts and *.mo files (if they exist) to the $(OutDir)

oleg-derevenetz:
With earlier Windows 8.1 SDK, `combaseapi.h` failed the build because it missed the `typename` in one place, so I used `WIN32_LEAN_AND_MEAN` to exclude COM from the `windows.h`.

Also I removed the MOD playback support. Are we really needed it for something? AFAIK there was no MOD music in any distributions of the original game.

Also I added an explicit initialization of `MID` module (MIDI) using the `MIX_INIT_MID`. Previously, it seemed to work without this, but this is necessary according to the documentation.

### DIFF
--- a/VisualStudio/SDL1.props
+++ b/VisualStudio/SDL1.props
@@ -12,8 +12,8 @@
       <AdditionalDependencies>SDL.lib;SDL_mixer.lib;zlib.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
     <PostBuildEvent>
-      <Command>xcopy /Y "$(MSBuildThisFileDirectory)packages\sdl1\lib\$(PlatformTarget)\*.dll" "$(OutDir)"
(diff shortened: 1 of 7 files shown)


undef21:
PSDK8.1 is pain.
Another way to avoid errors in `combaseapi.h` is disabling `/pemissive-`.
Could we use something like
```
<ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug-SDL1|x64'">
    <ClCompile>
      <ConformanceMode Condition="$(WindowsTargetPlatformVersion.StartsWith('8.1'))">false</ConformanceMode>
    </ClCompile>
</ItemDefinitionGroup>
```
in the `.vcproj` files instead defining macro in many files?

oleg-derevenetz:
Hi @undef21 there are just 3 files that include `windows.h`, and it seems to me a lesser evil than disabling the `/permissive-`.

ihhub:
@oleg-derevenetz , a bunch of thanks for these changes!

PR #5694, opened 2022-07-25
Rework music cache, don't cache the Mix_Music objects directly

oleg-derevenetz:
fix #5688

PR #5708, opened 2022-08-01
Fix issues with SDL_Mixer 1.x, as well as MIDI instrument and drum descriptions

oleg-derevenetz:
Fix the regressions after #5694

* It turns out that with ancient SDL_Mixer 1.x (used for SDL1 builds) `Mix_HookMusicFinished()`'s callback is not called if music is stopped using `Mix_HaltMusic()`, which causes the music cleanup process to fail;
* SDL_Mixer 1.x (at least 1.2.12) crashes somewhere inside its own music thread if `Mix_FreeMusic()` is called for an already inactive `Mix_Music` while playing another music track;
* Also this PR fixes MIDI instrument and drum descriptions which are currently messed up.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -1069,8 +1069,8 @@ void Music::Stop()
     // callback will not be called while we are modifying the current track information.
     Mix_HaltMusic();
 #if !SDL_VERSION_ATLEAST( 2, 0, 0 )
-    // Ancient SDL_Mixer 1.x doesn't call the Mix_HookMusicFinished()'s callback on Mix_HaltMusic(),
-    // so we need to call it manually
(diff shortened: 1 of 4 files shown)


src/engine/audio.cpp
```diff
@@ -183,6 +183,8 @@ namespace
     {
         assert( ptr != nullptr && size != 0 );
 
+        soundSampleManager.clearFinishedSamples();
```
oleg-derevenetz, line 186: This change is only to match the style of `musicTrackManager.clearFinishedMusic()` call, it is not essential for the logic.

ihhub:
@oleg-derevenetz , great job! Thanks a lot for polishing audio related logic!

PR #5709, opened 2022-08-03
Remove unused samples, instruments and presets from the fheroes2.sf3

oleg-derevenetz:
Reduced the SoundFont size from ~6.5Mb to ~2.5Mb and significantly accelerated its loading. The list of instruments is as follows:

instr

I.e. 25 instruments, Standard and Orchestral drum kits.

Missing instruments in the original SoundFont:

* `MIDI0013` (Desert terrain) theme needs the `128:49` Drums preset (so-called "Fix Room Drum Kit"), and fallbacks to the `128:0` (Standard drum kit);
* `MIDI_ORIGINAL_WIZARD` theme needs the `128:53` Drums preset (I have no idea what this preset is), and fallbacks to the `128:0`;
* `MIDI_ORIGINAL_KNIGHT` theme needs the `128:52` Drums preset (I have no idea what this preset is), and fallbacks to the `128:0`.


ihhub:
Hi @oleg-derevenetz , could you please confirm that all these sounds come from `GeneralUser GS`? Do you also think there is a way to get a full / missing Drum sounds?

oleg-derevenetz:
Hi @ihhub

> could you please confirm that all these sounds come from GeneralUser GS

Yes, they are just selectively exported from the original file to the SF3 format.

> Do you also think there is a way to get a full / missing Drum sounds?

They are not sounds, but presets. Standard preset `128:0` contains the full set of drum sounds. You see, unlike the usual music instruments, where each key of the virtual synthesizer corresponds to a musical note, in drum kits, each key of the virtual synthesizer corresponds to a separate drum sound, since there is no way to make a drum emit different musical notes. There are different presets for a different drum kits (`128:X`) and in each kit, the drums may sound a little different, but in general they are interchangeable with `128:0`.

I know where to get the `128:49` preset, but there are license issues arise. Regarding others, I don't know :) I even don't know what are these presets so far.

oleg-derevenetz:
@ihhub P.S. it looks like the drum is currently logged in a wrong way in `xmi2mid.cpp`, presets are logged as individual sounds, which is wrong.

ihhub:
Hi @oleg-derevenetz , let's not violate licenses :) We might find a solution in the future.

ihhub:

Issue #5716, opened 2022-08-08
Polish rendering on Adventure Map


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

After merging #5692 we still have multiple tasks to be done in order to polish rendering on Adventure Map:
- [ ] standartise object fading animation including heroes
- [ ] remove old save formats (before 0.9.16? release)
- [ ] do not store image parts but rather image regions and flip flag and use them while rendering
- [ ] rendering of monsters over Dragon City (added a TODO comment for it)
- [ ] rendering heroes / boats outside map while they are in fog
- [ ] overlapping of boats during movement (froth should be rendered under another boat for some cases; perspective of boats)
- [ ] path icon should be removed after a hero makes a turn, not before
- [ ] shadows of heroes must be drawn over other objects (tricky one. The original version does this but it breaks a lot of other places. For now we stick with common logic for all objects)

All issues related to rendering on Adventure Map should be reported here to make them easier to track.

### Describe the possible solution

Fix everything.

Branikolog:
An issue with wide tree:
![image](https://user-images.githubusercontent.com/55348946/185618608-eeee042b-9dee-4e4a-baca-89ac8cb4e4a0.png)
![image](https://user-images.githubusercontent.com/55348946/185619222-1f691ccd-1e5d-4b6f-936c-f156d0c07c26.png)
[Tree.zip](https://github.com/ihhub/fheroes2/files/9382170/Tree.zip)

ihhub:
Hi @Branikolog , I think it works exactly the same way as in the original game due to object info.

oleg-derevenetz:
It seems that Mine guards are no longer drawn (map "Skirmish"):

<img width="638" alt="Screenshot1" src="https://user-images.githubusercontent.com/32623900/185935699-de1c7ba1-ff78-4b25-ab94-11a6ba52e54d.png">

<img width="640" alt="Screenshot2" src="https://user-images.githubusercontent.com/32623900/185935720-113fb1a5-bb4d-4da5-a515-ff0a3c3254d4.png">

ihhub:
Oh, no. I will fix it.

Branikolog:
> Hi @Branikolog , I think it works exactly the same way as in the original game due to object info.

Hi, @ihhub !
[...]

PR #5719, opened 2022-08-08
Do not draw objects over the map edge if they are in fog

ihhub:
relates to #5716

### DIFF
--- a/src/fheroes2/gui/interface_gamearea.cpp
+++ b/src/fheroes2/gui/interface_gamearea.cpp
@@ -436,6 +436,8 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
     const bool drawFog = ( flag & LEVEL_FOG ) == LEVEL_FOG;
 #endif
 
+    const int friendColors = Players::FriendColors();
+
     TileUnfitRenderObjectInfo tileUnfit;
 
     const Heroes * currentHero = drawHeroes ? GetFocusHeroes() : nullptr;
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -493,9 +495,14 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
         }
     }
 
-    // Tile unfit objects should be rendered over the edge of the map, except the bottom. We shouldn't render their shadows.
+    // Tile unfit objects should be rendered over the edge of the map, except the bottom. We also shouldn't render their shadows over the edge.
+    // If a tile contains fog we shouldn't draw anything over the edge.
     if ( minY == 0 ) {
         for ( int32_t x = minX - 1; x < maxX + 1; ++x ) {
+            if ( drawFog && world.GetTiles( std::clamp( x, 0, maxX - 1 ), minY ).isFog( friendColors ) ) {
```
oleg-derevenetz, line 502: Hi @ihhub Can the `maxX` here be 0? If no, I'd still add an assertion nearby to verify that it doesn't happen, because if the third argument of `std::clamp()` is less than the second one, then the behavior is undefined.

ihhub: Hi @oleg-derevenetz , there is a check at line 406 to prevent this situation.

oleg-derevenetz (approved):

PR #5703, opened 2022-07-26
Fix the pre-battle arrangement of neutral troops

oleg-derevenetz:
fix #5700

PR #5727, opened 2022-08-08
Fix typo in Army::ArrangeForBattle()

### LINKED ISSUE #5726: Application crashes on AI turn
### Preliminary checks

### Platform

Windows

### Describe the bug

Assertion during AI turn.

### Save file

[bugg.zip](https://github.com/ihhub/fheroes2/files/9282776/bugg.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5726

Regression after #5703.

The crash due to an assertion failure was in the AI pathfinder because, due to this typo, the same AI-controlled hero with the same army on the same tile with the same guardians sometimes was able to defeat those guardians and sometimes (when guardians get the upgraded stack due to the non-deterministic random) he couldn't do it. This breaks the AI pathfinder logic because it always expects the same results of the `Army::GetStrength()` for the same army on the same tile.


ihhub:
@oleg-derevenetz wow and thanks for the fix!

PR #5732, opened 2022-08-09
AIPathfinder: make the cache of the tile army a static local variable, not a member of the class

oleg-derevenetz:
Minor optimization.


ihhub:
@oleg-derevenetz , thanks for the improvement!

PR #5747, opened 2022-08-12
Fix the battle order after bad morale events

### LINKED ISSUE #5745: Battle screen, triggered low morale should make other creature with the same speed to act
### Preliminary checks

### Platform

Windows

### Describe the bug

If a creature with triggered low morale skips its turn and hero has one more creature with the same speed, then turn should be transferred to the creature of the same player. If next creature also skips its turn due to low morale, the next one with the same speed should be able to make its action. Until one creature of the same speed don't have low morale triggered.
image
image

Currently fheroes2 works like Heroes 3: if creature has low morale triggered then next creature is going to act according to battle que.

If a creature skips its turn due to low morale and its speed is higher, than speeds of the rest, then opponent creature should make its action (in a case opponent's creatures have higher or the same speed comparing the rest of the current player).
image
(Second wolves skipped their turn due to low morale, so Pikemen are able to act now.)

Also, I want to mention that in a case the last creature has low morale triggered, then turn is passed to opponent creature in the beginning of the next turn. (In a case both sides have same speed first-acting creatures.)

### Save file

[Low morale.zip](https://github.com/ihhub/fheroes2/files/9315976/Low.morale.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5745

Also minor style nits (curly braces added, some comments paraphrased), some checks are replaced by assertions, because what is being checked for should never happen.

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -332,8 +332,8 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
     const int32_t dst = cmd.GetValue();
     const int32_t dir = cmd.GetValue();
 
-    Battle::Unit * attacker = GetTroopUID( attackerUID );
-    Battle::Unit * defender = GetTroopUID( defenderUID );
+    Unit * attacker = GetTroopUID( attackerUID );
+    Unit * defender = GetTroopUID( defenderUID );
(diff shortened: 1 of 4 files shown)


ihhub:
@oleg-derevenetz , thanks a lot for these fixes!

PR #5751, opened 2022-08-13
Fix some compilation warnings

ihhub:
Less warnings is always better.

### DIFF
--- a/src/engine/tinyconfig.cpp
+++ b/src/engine/tinyconfig.cpp
@@ -41,7 +41,7 @@ std::string ModifyKey( const std::string & str )
     key.erase( std::unique( key.begin(), key.end(), SpaceCompare ), key.end() );
 
     // change space
-    std::replace_if( key.begin(), key.end(), ::isspace, static_cast<char>( 0x20 ) );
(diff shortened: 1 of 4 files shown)


src/engine/tinyconfig.cpp
```diff
@@ -41,7 +41,7 @@ std::string ModifyKey( const std::string & str )
     key.erase( std::unique( key.begin(), key.end(), SpaceCompare ), key.end() );
 
     // change space
-    std::replace_if( key.begin(), key.end(), ::isspace, 0x20 );
+    std::replace_if( key.begin(), key.end(), ::isspace, static_cast<char>( 0x20 ) );
```
oleg-derevenetz, line 44: `'\x20'` is a valid `char` literal too and looks a bit shorter :)

oleg-derevenetz (commented):
Hi @ihhub I just left one minor optional suggestion, LGTM.

src/engine/tinyconfig.cpp
ihhub: Fixed!

oleg-derevenetz (approved):

PR #5753, opened 2022-08-13
src/engine/system.cpp: speed up resolution of unmodified paths on unix

trofi:
Without the change `fheroes2` can start on NixOS for a few minutes.
It happens due to suboptimal interaction of NixOS flat directory layout:

    /nix/store/...-fheroes2-${ver}/share/fheroes2/files/lang
    /nix/store/...-other-package-1/...
    /nix/store/...-other-package-2/...
    ...
    /nix/store/...-other-package-10000/...

and of `bool System::GetCaseInsensitivePath(path, correctedPath)` function.
When `System::GetCaseInsensitivePath("/nix/store/...-fheroes2-${ver}/share/fheroes2/files/lang", out);`
is called `/nix/store` is fully expanded first to the list of file to look
up `...-fheroes2-${ver}` entry in a case insensitive manner.

My system happens to have 400_000 `/nix/store` entries and it takes tens
of seconds to match a file there.

The change adds an optimisitic direct entry lookup to avoid linear
directory scan.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -401,6 +401,31 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
 
         correctedPath.append( delimiter );
 
+        // Avoid directory traversal and try to probe directory name directly.
+        // Speeds up file lookup when intermediate directories have a lot of
+        // files. Example is NixOS where file layout is:
+        //     /nix/store/...-fheroes2-${ver}/share/fheroes2/files/lang
+        //     /nix/store/...-other-package-1/...
+        //     /nix/store/...-other-package-2/...
+        // It's not uncommon for /nix/store to have tens of thousands files.
+        // GetCaseInsensitivePath() calls become very expensive on such systems.
+        //
+        // The idea is to try to open current subpath as a directory and avoid
+        // directory traversal altogether. Otherwise fall back to linear
+        // case-insensitive search.
+
+        std::string absSubpath = correctedPath + *subPathIter;
+        DIR * de = opendir( absSubpath.c_str() );
+        if ( de ) {
+            correctedPath = std::move( absSubpath );
+
+            closedir( d );
+            d = opendir( correctedPath.c_str() );
+
+            closedir( de );
+            continue;
+        }
+
         const struct dirent * e = readdir( d );
         while ( e ) {
             if ( strcasecmp( ( *subPathIter ).c_str(), e->d_name ) == 0 ) {


src/engine/system.cpp
```diff
@@ -401,6 +401,33 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
 
         correctedPath.append( delimiter );
 
+        {
```
ihhub, line 404: Please remove extra parenthesis here.

src/engine/system.cpp
```diff
@@ -401,6 +401,33 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
...
+            // GetCaseInsensitivePath() calls become very expensive on such systems.
+            //
+            // The idea is to try to open current subpath as a directory an avoid
+            // directory traversal altogether. Otherwise fall back to linear
+            // case-insensitive search.
+
+            std::string absSubpath = correctedPath + *subPathIter;
+            DIR * de = opendir( absSubpath.c_str() );
+            if ( de ) {
+                correctedPath += *subPathIter;
```
ihhub, line 421: Please use `std::move()` call to avoid extra memory operations.

ihhub (changes requested):
Hi @trofi , I left 2 comments in this pull request. Could you please take a look at them?

src/engine/system.cpp
trofi: Switched to `std::move( absSubpath )`.

ihhub (approved):

src/engine/system.cpp
```diff
@@ -401,6 +401,31 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
...
+        // Avoid directory traversal and try to probe directory name directly.
+        // Speeds up file lookup when intermediate directories have a lot of
+        // files. Example is NixOS where file layout is:
+        //     /nix/store/...-fheroes2-${ver}/share/fheroes2/files/lang
+        //     /nix/store/...-other-package-1/...
+        //     /nix/store/...-other-package-2/...
+        // It's not uncommon for /nix/store to have tens of thousands files.
+        // GetCaseInsensitivePath() calls become very expensive on such systems.
+        //
+        // The idea is to try to open current subpath as a directory an avoid
```
oleg-derevenetz, line 413: I believe there is a typo: `an avoid` -> `and avoid`.

oleg-derevenetz (changes requested):
Hi @trofi Found just one typo in comment, LGTM in general.

src/engine/system.cpp
trofi: Good catch! Updated.

oleg-derevenetz (approved):

ihhub:
@trofi , thank you so much for the fix!

PR #5754, opened 2022-08-13
Makefile: build translations by default

trofi:
Building translations by default allows running inplace fheroes2 with
languages enabled:

    $ make
    $ ./fheroes2

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -75,9 +75,6 @@ jobs:
       run: |
         make -j 2
       env: ${{ matrix.config.env }}
-    - name: Generate translations
-      run: |
-        make -C files/lang -j 2
     - name: Create app bundle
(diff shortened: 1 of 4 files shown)


oleg-derevenetz:
Hi @trofi then it turns out that the `Generate translations` step for Linux/macOS in the `make.yml` workflow becomes redundant, it would be good to remove it.

trofi:
> Hi @trofi then it turns out that the `Generate translations` step for Linux/macOS in the `make.yml` workflow becomes redundant, it would be good to remove it.

Sounds good! Removed redundant calls from `make.yml` and from `Debian`s `control` files.

oleg-derevenetz (approved):

ihhub:
@trofi , thank you so much for this improvement!

PR #5756, opened 2022-08-14
Fix possible crash while calculating points for a straight line

### LINKED ISSUE #5755: NaN in fheroes2::GetEuclideanLine()
### Preliminary checks

### Platform

macOS

### Describe the bug

Sometimes during the siege, when the tower tries to shoot at a unit located close to it, UB sanitizer may report the following attempt to convert NaN to int:

```
14:12:41: [DBG_BATTLE]	ApplyActionTower:  tower: 1, attack to Unit: [ 7 Paladins, Blue, pos: 7, -1 ]
tools.cpp:267:50: runtime error: nan is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior tools.cpp:267:50 in 
tools.cpp:267:89: runtime error: nan is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior tools.cpp:267:89 in 
../fheroes2/battle/battle_interface.cpp:2954:124: runtime error: signed integer overflow: -2147483648 - 18 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../fheroes2/battle/battle_interface.cpp:2954:124 in 
image.cpp:264:37: runtime error: negation of -2147483648 cannot be represented in type 'int32_t' (aka 'int'); cast to an unsigned type to negate this value to itself
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior image.cpp:264:37 in 
image.cpp:270:19: runtime error: signed integer overflow: 18 - -2147483648 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior image.cpp:270:19 in 
14:12:41: [DBG_BATTLE]	ApplyDamage:  14 to Unit: [ 7 Paladins, Blue, pos: 7, -1 ] and killed: 0
```

It seems that sometimes both `length` and `dx`/`dy` may become zero, and in this case here:

https://github.com/ihhub/fheroes2/blob/c1002c71e0c8a897c07e7f4da0781a0aa75c9d08/src/engine/tools.cpp#L260-L261

either `moveX` or `moveY` (or both) may become NaN. But I'm not sure how to better fix it in this particular case. Hi @ihhub what do you think?

### Save file

None available.

### Additional info

_No response_

ihhub:
close #5755

The problem was in a situation when `step` was greater than distance leading to length to be 0.


src/engine/tools.cpp
```diff
@@ -255,8 +255,13 @@ namespace fheroes2
...
         const int dy = pt2.y - pt1.y;
         const uint32_t dist = static_cast<uint32_t>( std::hypot( std::abs( dx ), std::abs( dy ) ) );
+
         // round up the integer division
-        const uint32_t length = ( step > 0 && dist >= step / 2 ) ? ( dist + step / 2 ) / step : 1;
+        uint32_t length = 1;
+        if ( step > 0 ) {
+            length = std::max( 1U, ( dist + step / 2 ) / step );
+        }
+
```
oleg-derevenetz, line 264: Thus, if `dx`, `dy` and `dist` are zero, the `line` below will consist of two identical points. Is this OK for rendering purposes? If yes, then I'm fine with this solution.

oleg-derevenetz (commented):

src/engine/tools.cpp
ihhub: Actually it is not correct. We shouldn't show an object on the same position during double time.

ihhub:
This PR must be merged for 1.0 release.

src/engine/tools.cpp
Districh-ru: Hi @ihhub and @oleg-derevenetz, the issue, described in #5755 happens for Paladin because `dist` is equal to 12 and the given `step` is equal to 25. So it passes the check `dist >= step / 2` and the result of `( dist + step / 2 ) / step` is equal to 0,98 and after converting to integer it is 0.
To avoid this I propose to slightly change this method, if you don't mind (including remove of redundant `std::abs(  )`):
```cpp
    std::vector<Point> GetEuclideanLine( const Point & pt1, const Point & pt2, const uint32_t step )
    {
        const int dx = pt2.x - pt1.x;
        const int dy = pt2.y - pt1.y;
        const uint32_t dist = static_cast<uint32_t>( std::hypot( dx, dy ) );
        // Round up the integer division and avoid the division by zero in calculation of total line points.
        const uint32_t length = ( step > 0 ) ? ( dist + step / 2 ) / step : 0;

        std::vector<Point> line;

        if ( length == 0 ) {
            // If the length is equal to 0 'pt2' could be closer to 'pt1' than 'step'.
            // In this case we put 'pt1' as the start of the line.
            line.emplace_back( pt1 );
            // And put 'pt2' as the end of the line only if 'pt1' is not equal to 'pt2'.
            if ( pt1 != pt2 ) {
                line.emplace_back( pt2 );
            }
        }
        else {
            // Otherwise we calculate the equclidean line, using the dermined parameters.
            const double moveX = dx / static_cast<double>( length );
            const double moveY = dy / static_cast<double>( length );

            line.reserve( length );

            for ( uint32_t i = 0; i <= length; ++i ) {
                line.emplace_back( static_cast<int>( pt1.x + i * moveX ), static_cast<int>( pt1.y + i * moveY ) );
            }
        }
        return line;
    }
```

And then for attack Paladin by the upper tower, we will have 2 points:

https://user-images.githubusercontent.com/113276641/208161713-bcbeed46-f65b-4d51-a466-a5d6ea0461fa.mp4

ihhub: Hi @Districh-ru , this function is used in calculating a projection for a missile of shooters. Can you please verify that it does break any? I might think if side effects as the missile appearing at the shooter sprite.

Districh-ru: > Hi @Districh-ru , this function is used in calculating a projection for a missile of shooters. Can you please verify that it does break any? I might think if side effects as the missile appearing at the shooter sprite.

This function is also used in `RedrawActionFly` and `RedrawRaySpell`.
The issue with

Districh-ru: You could see, that for the archer the arrow has lost its first position and at the end is rendered outside the targer:

https://user-images.githubusercontent.com/113276641/208225777-6ba23d66-6e60-41de-8e25-963de46c04e0.mp4

But for the Halfling its projectile is rendered without issues:

https://user-images.githubusercontent.com/113276641/208225807-fae581a9-70a2-4c7d-ba85-3f979b78ecde.mp4

ihhub: Hi @Districh-ru , I updated the code based on your suggestion.

src/engine/tools.cpp
```diff
@@ -255,19 +255,32 @@ namespace fheroes2
...
+        // Round up the integer division and avoid the division by zero in calculation of total line points.
+        const uint32_t length = ( step > 0 ) ? ( dist + step / 2 ) / step : 0;
 
         std::vector<Point> line;
-        line.reserve( length );
 
-        for ( uint32_t i = 0; i <= length; ++i ) {
-            line.emplace_back( static_cast<int>( pt1.x + i * moveX ), static_cast<int>( pt1.y + i * moveY ) );
+        if ( length == 0 ) {
+            // If the length is equal to 0 than 'pt2' could be closer to 'pt1' than 'step'.
```
Districh-ru, line 265: I can suggest to slightly optimize the function: if the length is 1, we also don't need to do the rest calculations as we already have the needed points: 'pt1' and 'pt2'.
```suggestion
        if ( length <= 1 ) {
            // If the length is equal to 1 than we have only two points: 'pt1' and 'pt2'.
            // If the length is equal to 0 than 'pt2' could be closer to 'pt1' than 'step'.
```

src/engine/tools.cpp
```diff
@@ -255,19 +255,32 @@ namespace fheroes2
...
+            if ( pt1 != pt2 ) {
+                line.emplace_back( pt2 );
+            }
         }
+        else {
+            // Otherwise we calculate the equclidean line, using the dermined parameters.
+            const double moveX = dx / static_cast<double>( length );
+            const double moveY = dy / static_cast<double>( length );
+
+            line.reserve( length );
```
Districh-ru, line 278: I didn't look carefully this last time, but in the next loop we get `( length + 1 )` points because `i` changes from `0` to `length`. :)
```suggestion
            line.reserve( length + 1 );
```

src/engine/tools.cpp
Districh-ru: Hi, @ihhub, I left two comments with minor optimizations for this code. Please check them, when you have time.

Branikolog (approved):

Districh-ru (approved):

oleg-derevenetz (approved):

PR #5757, opened 2022-08-15
Always use the current color of a unit when toggling the auto battle (take the Hypnotize spell into account)

oleg-derevenetz:
At the moment, when trying to toggle the auto battle when the hypnotized unit is active, the auto battle can be toggled for the wrong color.


ihhub:

PR #5760, opened 2022-08-17
Fix constantly moving AI heroes when they have nothing to do

### LINKED ISSUE #5650: Adventure map, AI being stuck on a small area is running non-stop
### Preliminary checks

### Platform

Windows

### Describe the bug

A couple AI heroes being stuck on a small island run in panic on a few squares wasting their movement points.
image
I think these heroes should stay waiting other hero visit their island. Lookin at them running purposelessly is a waste of player's time.

### Save file

[Pirate_Utopia_0054.zip](https://github.com/ihhub/fheroes2/files/9112974/Pirate_Utopia_0054.zip)
Multiplayer save.

### Additional info

_No response_

ihhub:
It is also much faster as we do not need to recalculate the map for a hero.

close #5650

### DIFF
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -677,7 +677,7 @@ int AIWorldPathfinder::getNearestTileToMove( const Heroes & hero )
     return -1;
 }
 
-bool AIWorldPathfinder::isHeroPossiblyBlockingWay( const Heroes & hero ) const
+bool AIWorldPathfinder::isHeroPossiblyBlockingWay( const Heroes & hero )
 {
     const int heroIndex = hero.GetIndex();
     const int heroColor = hero.GetColor();
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -679,7 +679,7 @@ int AIWorldPathfinder::getNearestTileToMove( const Heroes & hero )
 
 bool AIWorldPathfinder::isHeroPossiblyBlockingWay( const Heroes & hero )
 {
-    const int heroIndex = hero.GetIndex();
+    const int32_t heroIndex = hero.GetIndex();
     const int heroColor = hero.GetColor();
 
     auto isReachableDirection = [heroIndex, heroColor]( const int direction ) {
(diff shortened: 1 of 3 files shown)


src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -679,61 +679,108 @@ int AIWorldPathfinder::getNearestTileToMove( const Heroes & hero )
 
 bool AIWorldPathfinder::isHeroPossiblyBlockingWay( const Heroes & hero )
 {
-    // paths have to be pre-calculated to find a spot where we're able to move
-    reEvaluateIfNeeded( hero );
+    const int heroIndex = hero.GetIndex();
```
oleg-derevenetz, line 682: Strictly speaking, the map index is `int32_t`, not `int`. While they are usually the same, this may not always be the case.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -679,61 +679,108 @@ int AIWorldPathfinder::getNearestTileToMove( const Heroes & hero )
...
+    // |   | o |   |
+    // | o | H | o |
+    // | x | r | x |
+    if ( bottomReachable && ( topReachable || leftReachable || rightReachable ) && !bottomLeftReachable && !bottomRightReachable ) {
         return true;
     }
 
     // Is the hero standing on Stoneliths?
-    return world.GetTiles( start ).GetObject( false ) == MP2::OBJ_STONELITHS;
+    return world.GetTiles( heroIndex ).GetObject( false ) == MP2::OBJ_STONELITHS;
```
oleg-derevenetz, line 783: How about the following scenario:

```
|   | r | x |
| x | H | o |
|   | o |   |
```

? Is it possible that the hero will block the way in this case? Maybe just count the number of passable and impassable cells around the hero and assume that he can block the way if there are at least 2 impassable cells and 2 passable cells nearby at the same time? This is of course a bit pessimistic, but this should cover all the cases.

oleg-derevenetz (commented):
Hi @ihhub I just left a few notes and questions, please take a look when you have time.

src/fheroes2/world/world_pathfinding.cpp
ihhub: Sure. I will fix it.

src/fheroes2/world/world_pathfinding.cpp
ihhub: We cannot just count the number of tiles. Consider this example:
```
    // | x | x | x |
    // | r | H | r |
    // | r | r | r |
```
A hero is just standing near a "wall" like he doesn't block any other way.

oleg-derevenetz: That's why I call it pessimistic - due to possible false positives.

oleg-derevenetz: We can also try to count passable/impassable cells on each side separately and consider the hero as possibly blocking if there are at least 2 impassable cells on 2 different sides and at least 2 passable cells on 2 different sides.

ihhub: This pull request is aimed to avoid false positives which we have right now. I will check your idea.

ihhub: Hi @oleg-derevenetz , I couldn't find a proper dependency between cell positions without introducing false positives. So far if a non-passable cell is left/right/bottom/top then it has 3 combinations with another non-passable cells. If a cell is a corner then it has 4 combinations. I checked everything and the logic covers all possible cases. We can simplify the logic if needed.

ihhub: The complexity of this solution is much lower in comparison with previous solution.

oleg-derevenetz (approved):

PR #5762, opened 2022-08-19
Make "Show terrain penalty" option by default

ihhub:
Since we always show movement points for hero's dialog there is no reason to do not show penalty points from a terrain in this case.

relates to #1272


zenseii (approved):
LGTM.

Tested with and without expert pathfinding on all terrains, sea included.

Only question I have is if Lava terrain should only have 100 base penalty , i.e. when you have no pathfinding?
image

oleg-derevenetz (approved):

oleg-derevenetz:
> Only question I have is if Lava terrain should only have 100 base penalty , i.e. when you have no pathfinding?

Yes, Lava, just like Grass and Dirt, is not considered rough terrain.

PR #5765, opened 2022-08-19
Fix the distribution of Mutant Zombies in the Graveyard

### LINKED ISSUE #5763: Wrong graveyard Zombie distribution
### Preliminary checks

### Platform

Windows

### Describe the bug

There's wrong troop distribution in graveyard on the latest build:
image
It was always 5 stacks of 20 Zombies.

### Save file

[Graveyard.zip](https://github.com/ihhub/fheroes2/files/9382589/Graveyard.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5763

Regression after #5703 - before that PR, `Troops::ArrangeForBattle()` always divided 100 units of any kind into 5 stacks, so it worked in the Graveyard by coincidence. Now the number of stacks is pseudo-random, so `ArrangeForBattle()` cannot be used for the Graveyard (where there should always be 5 stacks) anymore.

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -791,7 +791,11 @@ void Army::setFromTile( const Maps::Tiles & tile )
         break;
 
     case MP2::OBJ_GRAVEYARD:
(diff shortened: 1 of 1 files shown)


ihhub:

PR #5767, opened 2022-08-19
Fix build for PSV

oleg-derevenetz:
This patch [is already in the SDL upstream](https://github.com/libsdl-org/SDL/pull/6086), but we need to apply it temporarily in our CI until the fixed version becomes available in the PSV SDK.


ihhub:
@oleg-derevenetz , thank you for the fix!

PR #5769, opened 2022-08-20
Remove "Remember last focus" option

### LINKED ISSUE #3147: Adventure map screen, reset of the castle list with the "game: remember last focus" option enabled
When a castle at the bottom of the list is selected and the option "game: remember last focus" is activated, and the next turn a hero is selected and moved, the castle list is reset to the top of the list.

In fheroes2:

https://user-images.githubusercontent.com/43583994/114273792-d13a8580-9a1b-11eb-84f2-2980cb8b21f8.mp4

ihhub:
This option might actually lead to thinking that the engine is buggy for players who enable this option without checking it. We need to make the logic of focusing identical for every player.

As argument towards removing this option:
- focusing on a castle on a next turn is usually illogical as most of players start doing actions by heroes
- we allow player to put heroes asleep

close #3147

### DIFF
--- a/files/lang/be.po
+++ b/files/lang/be.po
@@ -8354,9 +8354,6 @@ msgid ""
 "%{hp} HP."
 msgstr ""
 
-msgid "game: remember last focus"
-msgstr ""
-
 msgid "battle: show damage info"
 msgstr ""
 
--- a/files/lang/bg.po
+++ b/files/lang/bg.po
@@ -8420,9 +8420,6 @@ msgid ""
 "%{hp} HP."
 msgstr ""
 
-msgid "game: remember last focus"
-msgstr ""
-
 msgid "battle: show damage info"
 msgstr ""
 
--- a/files/lang/cs.po
+++ b/files/lang/cs.po
@@ -8557,9 +8557,6 @@ msgid ""
 "%{hp} HP."
 msgstr ""
 
-msgid "game: remember last focus"
-msgstr ""
-
 msgid "battle: show damage info"
 msgstr ""
 
--- a/files/lang/de.po
+++ b/files/lang/de.po
@@ -9790,9 +9790,6 @@ msgstr ""
 "Dieser Zauberspruch kontrolliert bis zu\n"
 "%{hp} TP."
 
-msgid "game: remember last focus"
-msgstr "Spiel: den letzten Fokus merken"
-
 msgid "battle: show damage info"
 msgstr "Kampf: Schadensinformationen anzeigen"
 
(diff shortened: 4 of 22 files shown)


zenseii (approved):
I agree with removing this. Most Heroes players are used to using the sleep/awake option for managing where to get the focus the next turn.

oleg-derevenetz (approved):

felix642:
After merging this PR I think we can close this issue : #3147

LeHerosInconnu:
Hello @ihhub, @zenseii and @felix642,

I use this option all the time and find it very convenient.
Two examples:
- To automatically select the hero who should perform his action first the next turn.
- To automatically select the castle to be built in priority the next turn, without having to find it in a long list of castles.

> This option might actually lead to thinking that the engine is buggy for players who enable this option without checking it.

Rather than removing this option, wouldn't it be more appropriate to make it so that the player can access a description of what each option does.
In the current list, before moving to a more complete interface https://github.com/ihhub/fheroes2/discussions/2472, this could be a message window that is displayed when the player right-clicks on an option line.

> * we allow player to put heroes asleep

This feature is currently "hidden" (only accessible by the "z" key on the keyboard, and it's not mentioned anywhere), and I think most users don't even know it exists.

> close #3147

One solution to this is to not reset the castle list in this case.
So that, in the example described, the player can continue to access the last castle in the list without having to scroll down the list.

ihhub:
Hi @LeHerosInconnu , you have valid points about this. However, keeping the list of experimental features is not a proper way to deliver information for end users. I absolutely agree that some features like making heroes to sleep must be shown for players so they know about them. This is planned to be a part of an interactive tutorial embedded into the game.

As for options, the best way is to follow [this guide](https://github.com/ihhub/fheroes2/wiki/Adding-a-new-option-into-the-game) for any of them. Some features presented in the game are extremely hard to adapt for these requirements. Also as we all noticed many of these features break the game or AI doesn't use them (for example, No interface mode or wait button). Combining your proposal in [this disccussion](https://github.com/ihhub/fheroes2/discussions/2472) we would find the most optimal solution to present options. As of now for 1.0 release we need to get rid of experimental options dialog at all and make some options as a part of the game like it is done for other features.

After 1.0 release we are going to add options to the game knowing that the game basis is stable enough and correct.

Lastly, regarding the deleted option: I agree that the original game focusing logic is not perfect so we can come up with logic to simplify player's life and at the same time to be easy to understand.

I think the best solution for all this is to allow players to drag and drop heroes and castles within lists like it is done for Heroes 4. This will cover any situation without introduction of extra options.

PR #5770, opened 2022-08-20
Apply PSV paches only if the cached version has not been used

oleg-derevenetz:
Cached version already contains all necessary fixes, so there is no need to apply them once again.


ihhub:

PR #5772, opened 2022-08-20
Added small offset to price of loyalty icons

### LINKED ISSUE #5620: Campaign selection window, some scenarios have shifted red frame
### Preliminary checks

### Platform

Windows

### Describe the bug

image
6, 7, 8 scenarios have shifted red ring.

### Save file

The Price of Loyalty campaign.

### Additional info

_No response_

felix642:
closes #5620 

For some reasons the price of loyalty's icons do not follow the standard offset for the last 3 items. 
When comparing with any other menu we can clearly see that the icons are at the wrong location.

The workaround I found is to add a few pixels to the offset of those specific scenarios. 

image

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -164,6 +164,21 @@ namespace
         offset.x *= deltaX;
         offset.y *= deltaY;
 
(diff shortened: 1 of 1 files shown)


ihhub:
@felix642 , thank you very much for the fix!

PR #5779, opened 2022-08-22
Implement the instant finish of the battle

### LINKED ISSUE #5626: Add an option to complete battle instantly
### Preliminary checks

Sometimes players want to finish battle as soon as possible, but they don't want to run away or surrender. The only way to do this currently is to set maximum speed and initiate auto battle. Or continue playing manually.

I think we should implement at least hot key for now to jump to the battle outcome, as if player pressed AUTO.

### Additional info

_No response_

oleg-derevenetz:
close #5626

Currently, only the hotkey can be used (`Q` by default - yes, just like in HoTA). There is no graphical button for this.

https://user-images.githubusercontent.com/32623900/185990658-f3099b04-7db1-4e0b-af83-760fa9d26b35.mp4

Currently, this works both in single-player and in hotseat mode, because, since now both human players are able to use the instant battle and auto combat, I do not see a reason to disable the instant finish in hotseat (and yes, it is convenient for debugging as well).

Also this PR fixes potential issues with non-deterministic random generator for Genie attack, Chain Lighting troop resistance and Earth Quake damage.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -2408,11 +2408,11 @@ void Battle::Interface::HumanBattleTurn( const Unit & b, Actions & a, std::strin
             EventShowOptions();
         }
         // Switch the auto battle mode on/off
-        else if ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTOSWITCH ) ) {
+        else if ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_SWITCH ) ) {
             EventAutoSwitch( b, a );
         }
         // Finish the battle in auto mode
-        else if ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTOFINISH ) ) {
+        else if ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_FINISH ) ) {
             EventAutoFinish( a );
         }
         // Cast the spell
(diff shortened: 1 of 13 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2708,10 +2716,24 @@ void Battle::Interface::EventShowOptions()
...
 
+    a.emplace_back( CommandType::MSG_BATTLE_AUTO_SWITCH, b.GetCurrentOrArmyColor() );
+
+    humanturn_redraw = true;
+    humanturn_exit = true;
+}
+
+void Battle::Interface::EventAutoFinish( Actions & a )
+{
+    if ( Dialog::Message( "", _( "Are you sure you want to finish the battle in auto mode?" ), Font::BIG, Dialog::YES | Dialog::NO ) != Dialog::YES ) {
```
ihhub, line 2731: Please use the newest version of UI Dialog:
```cpp
if ( fheroes2::showMessage( fheroes2::Text( "", {} ),
     fheroes2::Text( _( "Are you sure you want to finish the battle in auto mode?" ), fheroes2::FontType::normalWhite() ), Dialog::YES | Dialog::NO ) != Dialog::YES ) {
    return;
}
```

I know it's much longer but we might come up with some shorter solution in the future.

ihhub (commented):
Hi @oleg-derevenetz . everything looks good except one place where we should use newer version of UI calls. Please take a look.

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: Done in both places.

ihhub:
@oleg-derevenetz , a bunch of thanks for implementing this feature!

Issue #5780, opened 2022-08-24
View Puzzle screen, wandering creatures and campfires should not be displayed


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, wandering creatures and campfires should not be displayed in the View Puzzle screen.

In original game:

![View puzzle creatures 001](https://user-images.githubusercontent.com/43583994/186418589-97f606b3-dc26-4b4a-9d92-46f383f34abb.png)

In fheroes2:

![View puzzle creatures 002](https://user-images.githubusercontent.com/43583994/186418628-21c472f7-8141-4277-9820-84491778418a.png)

### Save file

Here is the test scenario file:
[0 9991b.zip](https://github.com/ihhub/fheroes2/files/9416019/0.9991b.zip)

Branikolog:
Hello, @LeHerosInconnu !
I've noticed, that both original game and fheroes2 show gatherable objects on the sea, like chests etc. 
I'm concerning whether they should be shown on a puzzle map, since they could disappear.

LeHerosInconnu:
Hello @Branikolog,

> Hello, @LeHerosInconnu ! I've noticed, that both original game and fheroes2 show gatherable objects on the sea, like chests etc. I'm concerning whether they should be shown on a puzzle map, since they could disappear.

I think they left them visible because the sea is completely similar and that these objects can therefore give a clue when the puzzle is only revealed in part and that part is only sea.
I personally agree with this. :)

Branikolog:
@LeHerosInconnu 
Okay. But these objects could disappear. Do original and fheroes2 puzzle maps update if someone collect all sea items on your screenshot?
I think, it's a bit weird, that campfires, resources and land chests shouldn't be drawn on the map, but sea objects are.
To my personal point, a large water part of the puzzle map makes it easier to dig, once player spot the area, since water is the place, where no one manages to dig. ;) Also, sea on a puzzle map shows, that you should seek for UA somewhere near beach, not in the middle of the continent.
So there're some pros and cons for both our theories. :)

LeHerosInconnu:
@Branikolog.

[...]

PR #5781, opened 2022-08-24
View Puzzle screen, wandering creatures and campfires should not be displayed

shprotru:
fix for #5780

@oleg-derevenetz @ihhub @idshibanov, how do you see, guys, the proper way to move this condition:
```
if ( !isPuzzleDraw && mp2_object == MP2::OBJ_CAMPFIRE ) {
            renderMainObject( dst, area, mp );
}
```
inside `isHiddenForPuzzle` function?

### DIFF
--- a/src/fheroes2/gui/interface_gamearea.cpp
+++ b/src/fheroes2/gui/interface_gamearea.cpp
@@ -464,7 +464,7 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
             }
 
             case MP2::OBJ_MONSTER: {
-                if ( isPuzzleDraw && !drawHeroes ) {
+                if ( isPuzzleDraw ) {
                     continue;
                 }
 
--- a/src/fheroes2/gui/interface_gamearea.cpp
+++ b/src/fheroes2/gui/interface_gamearea.cpp
@@ -464,6 +464,10 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
             }
 
             case MP2::OBJ_MONSTER: {
+                if ( !drawHeroes ) {
+                    continue;
+                }
+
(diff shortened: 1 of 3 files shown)


ihhub:
Hi @shprotru , this is wrong way to do as campfire sprite is more than 1 tile in size and not every tile has this object type.

ihhub:
The solution for campfire should be:
```cpp
bool MP2::isHiddenForPuzzle( const int terrainType, uint8_t tileset, uint8_t index )
{
    const int icnID = tileset >> 2;
    if ( icnID < 22 || icnID == 46 || ( icnID == 56 && index == 140 ) || ( icnID == 59 && index >= 14 && index <= 137 ) ) {
        return true;
    }

    return isDiggingHoleSprite( terrainType, tileset, index );
}
```

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -464,6 +464,10 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
             }
 
             case MP2::OBJ_MONSTER: {
+                if ( !drawHeroes ) {
```
ihhub, line 467: Are we certainly sure that this is correct for all cases? What's about View spells and View World?

shprotru: we can add here additional check for `isPuzzleDraw`:

```
if ( isPuzzleDraw && !drawHeroes ) {
                    continue;
}
```

will be this better?

ihhub: Personally, it feels like a hack which we must avoid. To do it properly we need to add another level of drawings like `LEVEL_MONSTERS` and use it properly without any mentioning of heroes or puzzle.

ihhub: Let's do an easy way:
```cpp
if ( isPuzzleDraw ) {
    // Monsters should not appear on puzzle map.
    continue;
}
```

ihhub:
@shprotru , thank you very much for the fix!

PR #5783, opened 2022-08-26
gcc-13: add missing '#include <cstdint>'

trofi:
Without the change build fails on weekly `gcc-13` snapshot as:

    ../fheroes2/spell/spell_info.h:33:5: error: 'uint32_t' does not name a type
       33 |     uint32_t getSpellDamage( const Spell & spell, const uint32_t spellPower, const HeroBase * hero );
          |     ^~~~~~~~
    ../fheroes2/spell/spell_info.h:25:1: note: 'uint32_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
       24 | #include <string>
      +++ |+#include <cstdint>

### DIFF
--- a/src/fheroes2/game/game_over.h
+++ b/src/fheroes2/game/game_over.h
@@ -24,6 +24,7 @@
 #ifndef H2GAMEOVER_H
 #define H2GAMEOVER_H
 
+#include <cstdint>
 #include <string>
 
 #include "game_mode.h"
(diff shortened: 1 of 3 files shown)


ihhub:
Hi @trofi , are you using a beta version of GCC 13 as this version is not even released?

ihhub:
@trofi , thank you very much for this fix!

trofi:
Yeah, I'm using weekly gcc snapshots to catch problems a bit before the release: https://mirrorservice.org/sites/sourceware.org/pub/gcc/snapshots/LATEST-13/. Minimizing headers includes is a frequent change in gcc releases: https://gcc.gnu.org/gcc-13/porting_to.html (thus it's not an accidental omission).

PR #5785, opened 2022-08-27
Implement some improvements when exchanging an army between heroes

oleg-derevenetz:
While watching [this](https://youtu.be/XckUhRMi3V0), I decided to propose this (debatable of course) PR, which is intended to make it easier for a player to exchange troops between heroes when performing typical PROgamer's 😎 tasks, such as:

* Creating hero chains in which you need to leave fast but relatively weak units to intermediate heroes:

https://user-images.githubusercontent.com/32623900/187012272-a2c47abd-2238-4454-af12-bf60f633462f.mp4

* Speeding heroes up, for which you need to leave just one fastest unit in hero's army:

https://user-images.githubusercontent.com/32623900/187012426-5e686413-232b-4fa3-8a02-a2f5f5c4115b.mp4

https://user-images.githubusercontent.com/32623900/187012432-f910fce7-eb39-4dc9-a9f2-ef3d74f27491.mp4

Hi @ihhub @Branikolog @LeHerosInconnu @zenseii thoughts are welcome. Please note that this solution doesn't assume any modification of original hero meeting screen (there is much less controls on it compared with the similar screen from HoMM3).

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -415,7 +415,7 @@ void Troops::JoinTroops( Troops & troops2 )
         }
 }
 
-void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UNKNOWN */, const bool keepInSrc /* = true */ )
+void Troops::MoveTroops( Troops & from, const int monsterToKeep )
 {
     assert( this != &from );
 
@@ -425,15 +425,13 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UN
 
     assert( isValid() && from.isValid() );
 
-    // The monster to keep is chosen from the destination stack set
-    if ( monsterToKeep != Monster::UNKNOWN && !keepInSrc ) {
+    if ( monsterToKeep != Monster::UNKNOWN ) {
         // Find a troop stack in the destination stack set consisting of monsters we need to keep
         const auto keepIter = std::find_if( begin(), end(), [monsterToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
 
             return troop->isValid() && troop->GetMonster() == monsterToKeep;
         } );
-        assert( keepIter != end() );
 
         // Find a troop stack in the source stack set consisting of monsters of a different type than the monster we need to keep
         const auto destIter = std::find_if( from.begin(), from.end(), [monsterToKeep]( const Troop * troop ) {
@@ -442,10 +440,9 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UN
             return troop->isValid() && troop->GetMonster() != monsterToKeep;
         } );
 
-        // If we found one, then exchange the monster to keep in the destination stack set with this troop stack. If we didn't find
-        // one, then this means that the source stack set after optimization consists of a single troop stack of monsters to keep
-        // and no additional actions are needed.
-        if ( destIter != from.end() ) {
+        // If we found both, then exchange the monster to keep in the destination stack set with the found troop stack in the
+        // source stack set to concentrate all the monsters to keep in the source stack set
+        if ( keepIter != end() && destIter != from.end() ) {
             Troop * keep = *keepIter;
             Troop * dest = *destIter;
 
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -490,14 +490,13 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UN
 
             break;
         }
-        else {
-            assert( remaining > 1 );
 
-            if ( JoinTroop( *troop ) ) {
-                troop->Reset();
+        assert( remaining > 1 );
 
-                --remaining;
-            }
+        if ( JoinTroop( *troop ) ) {
+            troop->Reset();
+
+            --remaining;
         }
     }
 
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -415,7 +415,7 @@ void Troops::JoinTroops( Troops & troops2 )
         }
 }
 
-void Troops::MoveTroops( Troops & from, const int monsterToKeep )
+void Troops::MoveTroops( Troops & from, const int monsterIdToKeep )
 {
     assert( this != &from );
 
@@ -425,19 +425,19 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep )
 
     assert( isValid() && from.isValid() );
 
-    if ( monsterToKeep != Monster::UNKNOWN ) {
+    if ( monsterIdToKeep != Monster::UNKNOWN ) {
         // Find a troop stack in the destination stack set consisting of monsters we need to keep
-        const auto keepIter = std::find_if( begin(), end(), [monsterToKeep]( const Troop * troop ) {
+        const auto keepIter = std::find_if( begin(), end(), [monsterIdToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
 
-            return troop->isValid() && troop->GetID() == monsterToKeep;
+            return troop->isValid() && troop->GetID() == monsterIdToKeep;
         } );
 
         // Find a troop stack in the source stack set consisting of monsters of a different type than the monster we need to keep
-        const auto destIter = std::find_if( from.begin(), from.end(), [monsterToKeep]( const Troop * troop ) {
+        const auto destIter = std::find_if( from.begin(), from.end(), [monsterIdToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
 
-            return troop->isValid() && troop->GetID() != monsterToKeep;
+            return troop->isValid() && troop->GetID() != monsterIdToKeep;
         } );
 
         // If we found both, then exchange the monster to keep in the destination stack set with the found troop stack in the
@@ -456,7 +456,7 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep )
 
             dest->Reset();
 
-            if ( !from.JoinTroop( Monster( monsterToKeep ), count ) ) {
+            if ( !from.JoinTroop( Monster( monsterIdToKeep ), count ) ) {
                 assert( 0 );
             }
         }
@@ -472,7 +472,7 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep )
             continue;
         }
 
-        if ( troop->GetID() == monsterToKeep ) {
+        if ( troop->GetID() == monsterIdToKeep ) {
             keep = troop;
 
             continue;
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -430,14 +430,14 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep )
         const auto keepIter = std::find_if( begin(), end(), [monsterToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
 
-            return troop->isValid() && troop->GetMonster() == monsterToKeep;
+            return troop->isValid() && troop->GetID() == monsterToKeep;
         } );
 
         // Find a troop stack in the source stack set consisting of monsters of a different type than the monster we need to keep
         const auto destIter = std::find_if( from.begin(), from.end(), [monsterToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
 
-            return troop->isValid() && troop->GetMonster() != monsterToKeep;
+            return troop->isValid() && troop->GetID() != monsterToKeep;
         } );
 
         // If we found both, then exchange the monster to keep in the destination stack set with the found troop stack in the
@@ -456,7 +456,7 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep )
 
             dest->Reset();
 
-            if ( !from.JoinTroop( monsterToKeep, count ) ) {
+            if ( !from.JoinTroop( Monster( monsterToKeep ), count ) ) {
                 assert( 0 );
             }
         }
@@ -472,7 +472,7 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep )
             continue;
         }
 
-        if ( troop->GetMonster() == monsterToKeep ) {
+        if ( troop->GetID() == monsterToKeep ) {
             keep = troop;
 
             continue;
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -415,7 +415,7 @@ void Troops::JoinTroops( Troops & troops2 )
         }
 }
 
-void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UNKNOWN */, const bool keepFrom /* = true */ )
+void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UNKNOWN */, const bool keepInSrc /* = true */ )
 {
     assert( this != &from );
 
@@ -426,7 +426,7 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UN
     assert( isValid() && from.isValid() );
 
     // The monster to keep is chosen from the destination stack set
-    if ( monsterToKeep != Monster::UNKNOWN && !keepFrom ) {
+    if ( monsterToKeep != Monster::UNKNOWN && !keepInSrc ) {
         // Find a troop stack in the destination stack set consisting of monsters we need to keep
         const auto keepIter = std::find_if( begin(), end(), [monsterToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -427,7 +427,7 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UN
 
     // The monster to keep is chosen from the destination stack set
     if ( monsterToKeep != Monster::UNKNOWN && !keepFrom ) {
-        // Find a troop stack in the source stack set consisting of monsters we need to keep
+        // Find a troop stack in the destination stack set consisting of monsters we need to keep
         const auto keepIter = std::find_if( begin(), end(), [monsterToKeep]( const Troop * troop ) {
             assert( troop != nullptr );
 
@@ -442,9 +442,9 @@ void Troops::MoveTroops( Troops & from, const int monsterToKeep /* = Monster::UN
             return troop->isValid() && troop->GetMonster() != monsterToKeep;
         } );
 
-        // If we found one, then exchange the monster to keep in the source stack set with this troop stack. If we didn't find one,
-        // then this means that the source stack set after optimization consists of a single troop stack of monsters to keep and no
-        // additional actions are needed.
+        // If we found one, then exchange the monster to keep in the destination stack set with this troop stack. If we didn't find
+        // one, then this means that the source stack set after optimization consists of a single troop stack of monsters to keep
+        // and no additional actions are needed.
         if ( destIter != from.end() ) {
             Troop * keep = *keepIter;
             Troop * dest = *destIter;
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -419,9 +419,8 @@ void Troops::MoveTroops( Troops & from, const int monsterIdToKeep )
 {
     assert( this != &from );
 
-    // Make an "optimized" version of each stack set (each unit type occupies just one slot)
-    Assign( GetOptimized() );
-    from.Assign( from.GetOptimized() );
+    // Combine troops of the same type in one slot to leave more room for new troops to join
+    MergeTroops();
 
     assert( isValid() && from.isValid() );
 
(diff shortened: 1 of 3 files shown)


zenseii:
Great, @oleg-derevenetz ! 👍 I see you picked up from where I was unable to.

I've quickly looked over the code, which is a lot more elegant than mine, and I'll spend time scrutinizing this with the experience I got during my own testing to make sure any outliers don't cause trouble.

LeHerosInconnu:
Hello @oleg-derevenetz,

> Hi @ihhub @Branikolog @LeHerosInconnu @zenseii thoughts are welcome. Please note that this solution doesn't assume any modification of original hero meeting screen (there is much less controls on it compared with the similar screen from HoMM3).

I've always been a fan of keeping the fastest (and weakest among them) troop when using those big yellow arrows.
So this is fine with me.
Also the pre-selection which allows to choose the troop to keep in priority in hero's army is also one of my proposals: https://github.com/ihhub/fheroes2/discussions/3742.
So, again, this seems to me to be a good way to operate. :)

oleg-derevenetz:
BTW in HoTA this works in a similar way:

https://user-images.githubusercontent.com/32623900/187031561-48af39aa-b550-4dcd-abcd-9f7d7dcf3249.mp4

Except that units selected in the destination army are not always taken into account:

https://user-images.githubusercontent.com/32623900/187031591-c4d3c91e-ab08-412b-ade6-ac5f0927c6c9.mp4

But sometimes they are:

https://user-images.githubusercontent.com/32623900/187031972-7f0aec7c-2091-4806-9097-b69596d4fa16.mp4

https://user-images.githubusercontent.com/32623900/187031976-ef27cc74-2741-4f9f-924c-3f9c97e5eced.mp4

https://user-images.githubusercontent.com/32623900/187031983-10fcd351-b1a7-4eef-acbf-dda8a353b2c2.mp4

Have no idea why. Anyway, HoTA have the similar mechanics, so experienced players should already get used to this.

Branikolog:
Hi, @oleg-derevenetz !
I think the selected troop and order of troops in destination and initial hero should be respected. In the PR build troop order is constantly changing.

oleg-derevenetz:
@Branikolog could you please elaborate a bit? In this PR troops are "squeezed" first - if there are stacks of the same troops, they will be joined in both armies to leave more place for different troops. Also the selected monster will be joined last (if it possible at all), so it will be joined in a last free slot, which naturally will change its order in the destination army.

Branikolog:
Few examples:
1)
image
image
I expect elf in the left hero army remain in the same slot.
2)
image
image
Destination hero army order was changed, when sprites should take the only free slot.
3)
image
image
While pressing transfer arrow hobbits shouldn't change their position, as there's no free space in the left army.
4)
image
image
Hobbits should also remain in the right slot and all transfered troops should take the remaining.
5)
image
image
Here the army order is respected (boars are placed in the free slot and armies were switched.
But when we have reversed inital order, boars are placed in the first slot:
image
image

oleg-derevenetz:
Hi @Branikolog now the algorithm should try harder to keep the original order of troops. However, it will still pre-merge troops in the destination army (although trying to maintain their original order), for example:

Screenshot 2022-08-29 в 00 26 31
Screenshot 2022-08-29 в 00 26 46

Branikolog:
@oleg-derevenetz 
Nice work!
I only have one more concern.
When we select one troop at destination army and press transfer button the troop is transferred. This behaviour is useful, but not apparent, since in the examples above the selected troop is tend to be the remaining or the one you split lastly. When you press an arrow it on contrary switches. I can get used to it, actually, but the logic is not obvious here though.
image
image
I think here we're substantially lacking switch button here. (<->) which will cover the above behaviour and would be extremely clear for players.
On the other hand, when we have a more complex situation, when destination army has more than 1 troop, the behaviour seems completely unclear:
image
image
The selected troop is transferred, while in the previous cases selection meant that troop is on contrary has a priority on leaving in army. 
If you ask me, how should it operate then, I'd rather perform no action here, as the current logic confuses me a lot.

Branikolog:
Just a few thought about the problem...
There were lots of proposals and even completed PRs regarding creature transferring.
The more we work, the more comfortable and predictable transferring logic becomes. What if we go even farther from the original mechanics and implement common OS behaviour here, when we can choose multiple creatures (by holding some key + left mouse click or pressing central mouse button (don't know why here it works as splitting single troop?) ), making them selected (with red glowing frame) and then by pressing arrow buttons transfer them to the hero we want? We have nice-looking frame animation and multiple items selection is the behaviour users are got used for many years.

oleg-derevenetz:
@Branikolog

> This behaviour is useful, but not apparent, since in the examples above the selected troop is tend to be the remaining or the one you split lastly. When you press an arrow it on contrary switches. I can get used to it, actually, but the logic is not obvious here though.

The behavior of "select monster and click transfer button" is always the same - "transfer all the monsters of this kind to the source army and join it last (if it possible at all)". It doesn't matter on which side this monster is selected, it always works in the same way. This covers both "chain" cases (because chained heroes typically have just one stack of fast, but weak units) and "speed up" cases (unit selection is performed in the source army in this case).

> I think here we're substantially lacking switch button here. (<->) which will cover the above behaviour and would be extremely clear for players.

Yes, but we don't have this button, so we need to give players at least some way to perform typical tasks.

> We have nice-looking frame animation and multiple items selection is the behaviour users are got used to for many years.

AFAIK none of really popular games in HoMM series (HoMM3, HoMM5) have an ability to select multiple stacks at once, so it should be, on the contrary, counterintuitive. Also all the "holding some key" combinations are already busy: Ctrl is to separate 1 unit in a free slot, Shift is to split in half, Alt to combine all splitted units of this kind to this slot - and these combinations are selected because they are already work this way in HoTA (and maybe HoMM5, I don't remember this for sure), which is the most popular game of the HoMM series now, therefore, it will be easier for those who have played HoTA to get used to it.

Branikolog:
@oleg-derevenetz 

> HoTA (and maybe HoMM5, I don't remember this for sure), which is the most popular game of the HoMM series now, therefore, it will be easier for those who have played HoTA to get used to it.

I'm dreaming that some day our project could become as (or even more) popular, since we're rushing for ultimate quality. :)

>Also all the "holding some key" combinations are already busy: Ctrl is to separate 1 unit in a free slot, Shift is to split in half, Alt to combine all splitted units of this kind to this slot 

Currently, Pressing left mouse button with Ctrl/Alt are not busy (in a case one troop is already selected): we can select one troop and then select more troops with holding Ctrl.
Also, what about central mouse button to select more troops?
Anyway, it's just an idea out of the current PR for sure. ;)

>This covers both "chain" cases (because chained heroes typically have just one stack of fast, but weak units) and "speed up" cases (unit selection is performed in the source army in this case).

True, but still I'm severely confused by the logic, where I select 1 of 2/3/4 troops, press transfer arrow aimed to the selected army and the only troop, who is transferred back is the selected one, while others are left in the initial army combined with transferred. o_O My hands are shaking here to select more troops to be able to switch them. )

zenseii:
Hi @oleg-derevenetz,

I've started testing this and the first thing I found was that same-creature stacks are joined into one stack. In general I would say that these shortcuts should be catered towards how a experienced player would use it, so that if someone chose to have this setup of their army then we should assume that there is a good reason for it, and therefore not break it by merging them all into a same stack.

I know this is due to how JoinTroop works, which is also why I didn't really like that function for this use case.

Here's a vid showing what I mean:

<details>

https://user-images.githubusercontent.com/12501091/187855093-f6bedf33-44ab-4e75-a3fc-3a9a03f0d034.mp4

<\details>

oleg-derevenetz:
Hi @zenseii this is for a reason - since this mechanics is intended for hero chains, it is always handy to transfer as much army as possible to the receiving hero, so his stacks of the same type are joined together. Experienced players do separate re-arrangement before battle anyway, because this rearrangement depends on specific attacked troops.

zenseii:
> Hi @zenseii this is for a reason - since this mechanics is intended for hero chains, it is always handy to transfer as much army as possible to the receiving hero, so his stacks of the same type are joined together. Experienced players do separate re-arrangement before battle anyway, because this rearrangement depends on specific attacked troops.

Yes, what you are saying is right and is the expected result for a player pressing that move button. However, I'm still worried that we will annoy players by ruining their army arrangement in cases where a transfer isn't done between a "troop-delivery"-hero and a main hero.

Not to mention that you should always keep your troops in a balanced arrangement when you end your turn. Like, for example, when you end the turn and an enemy suddenly attacks your out of nowhere with dimension door, or when you get ambushed when picking up an artifact.

In my humble opinion experienced players should always keep their troops arranged in an overall good formation, in which keeping troops in single stacks is often essential.

The single unit stacks/similar stacks can be preserved while still transferring all possible troops.

oleg-derevenetz:
@zenseii honestly I don't like an idea to keep splitted units as is when using the transfer button. The whole point of this button is to get as many reinforcements as possible from another hero, therefore, it makes sense to sacrifice the order of units in order to get more reinforcements. Hi @ihhub @Branikolog @LeHerosInconnu what do you think about pre-merging troops in the destination army in order to get more space for new units from the reinforcement?

zenseii:
> @zenseii honestly I don't like an idea to keep splitted units as is when using the transfer button. The whole point of this button is to get as many reinforcements as possible from another hero, therefore, it makes sense to sacrifice the order of units in order to get more reinforcements. Hi @ihhub @Branikolog @LeHerosInconnu what do you think about pre-merging troops in the destination army in order to get more space for new units from the reinforcement?

Yes, I agree that it makes sense to sacrefice the order of the units if there are any more units that could join, but I don't see why we have to sacrifice the order when there is no need to do it, only because they are the same troops.

oleg-derevenetz:
@zenseii Suppose that we have 3 groups of Sprites in the destination army, and full set of some other units in the source army (including Sprites). I select one group of Sprites in the destination army and then click on the Transfer button. With the current logic, all sprites in the destination army will be joined, then they will be exchanged with the first non-Sprites stack in the source army and then they all of them except one will be transferred to the destination army (after all other units are transferred). What would be the logic without the join? What if there is no Sprites in the source army? What if there are 3 groups of Sprites and 3 groups of, say, Peasants in the destination army? Are you going to propose sometimes to join stacks and sometimes not to join them? I believe that the logic should be simple and clear to the gamer, there should not be some quirky and non-intuitive logic "sometimes  it happens this way, and sometimes that way".

LeHerosInconnu:
Hello @oleg-derevenetz, @zenseii and @Branikolog,

> Hi @ihhub @Branikolog @LeHerosInconnu what do you think about pre-merging troops in the destination army in order to get more space for new units from the reinforcement?

I think the original layout should be kept as much as possible to avoid the player having to rearrange everything.

Here are some examples:

The three troops in the right army are transferred in the same order to the left army.
The similar troops of the left army are combined by concerving the initial position of the two troops on the right of the army.
The troops of gargoyles remain in the same slots.
Troop transfert 001
Troop transfert 002

Here the troop of one orc is combined with the troop with the largest number of orcs, the troop of four orcs completely to the right of the army, to make room for the troop of centaurs.
The troop of one gargoyle remains in the same slot.
Troop transfert 003
Troop transfert 004

Here the troop of one orc is combined with the troop with the largest number of orcs, the troop of four orcs completely to the right of the army, to make room for the troop of centaurs.
And the troop of goblins is combined with the other troop of goblins to make room for the troop of sprites.
The troop of one gargoyle remains in the same slot.
Troop transfert 005
Troop transfert 006

oleg-derevenetz:
@LeHerosInconnu this is a typical example of "sometimes we join troops and sometimes we do not join them", which I'd prefer to avoid :)

ihhub:
Hi everyone!

Looking at lengthy conversations here I believe everyone has their own opinion which is good and bad. I am okay with this way of doing things (others might have other opinion). To avoid further delays and unnecessary discussions I would like to know whether this change makes the logic better. Is the answer is yes in all aspect then we should proceed as it is.

oleg-derevenetz:
Hi @ihhub I should say that currently (without this PR) the order of joined units is NOT preserved, even if there is a way to preserve it (they are always joined):

https://user-images.githubusercontent.com/32623900/187936209-04f2fddc-f12a-40c7-aee1-9ba475e866bc.mp4

And I don't hear a dissatisfied chorus of players regarding this behavior :)

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu this is a typical example of "sometimes we join troops and sometimes we do not join them", which I'd prefer to avoid :) Why troops always joined to the right, and not to the left, because...? How about the situation like described above, with 3 stacks of Sprites? Should they be moved to the separate slots of the source army or not, and if yes, then why? If no, then why? So many questions, not just from me, but also from the ordinary players :)

I think that the selection with the red frame of a troop should be taken into account only in the origin army of the transfer, to indicate that it is this troop which will have to be kept in priority (so selecting a troop in the destination army has no effect at all).
As explained here: https://github.com/ihhub/fheroes2/discussions/3742.

In my examples above, the logic is simple.
For the first two screens:
- The first troop on the left in the left army, the troop of 1 orc, was added to the troop of orcs with the most orcs, the one furthest right in the left army, to free up a slot for the troop of centaurs.
- Then, the first troop on the left of the right army, the troop of 31 centaurs, was transferred to the freed slot of the left army.
- Then, the troop in the second slot of the left army, the troop of 16 goblins, was added to the troop of 18 goblins to free up a slot for the troop of sprites.
- Then, the next troop in the right army, the troop of 15 sprites, was transferred to the freed up slot of the left army.
- Then, the troop in the third slot of the left army, the troop of 3 orcs, was added to the troop of 5 orcs to free up a slot for the troop of medusas.
- Then, the next troop in the right army, the troop of 7 medusas, was transferred to the freed slot of the left army.

For the other screens, it's the same logic, except that there are fewer troops to transfer from the second army.

LeHerosInconnu:
@oleg-derevenetz.

> Hi @ihhub I should say that currently (without this PR) the order of joined units is NOT preserved, even if there is a way to preserve it (they are always joined):
> 
>  [fheroes2.engine.version_.0.9.18.2022-09-01.17-13-41.mp4 ](https://user-images.githubusercontent.com/32623900/187936209-04f2fddc-f12a-40c7-aee1-9ba475e866bc.mp4)

As the phoenix is faster, it's 1 phoenix that should have been kept in the army of origin. :smile: :eyeglasses: 

> And I don't hear a dissatisfied chorus of players regarding this behavior :)

They tried and they gave up: https://github.com/ihhub/fheroes2/discussions/5478. :rofl:

oleg-derevenetz:
@LeHerosInconnu

> I think that the selection with the red frame of a troop should be taken into account only in the origin army of the transfer, to indicate that it is this troop which will have to be kept in priority (so selecting a troop in the destination army has no effect at all).

That's a step back - it doesn't take into account chains of heroes, and doesn't implement the logic "move all army from the courier hero and leave him just this unit". That's why selection with red frame of a troop should be taken into account both in the source and in the destination army, just like in this case:

https://user-images.githubusercontent.com/32623900/187012272-a2c47abd-2238-4454-af12-bf60f633462f.mp4

> They tried and they gave up: https://github.com/ihhub/fheroes2/discussions/5478. 🤣

I don't see any dissatisfaction with the fact that the same units are combined when moving from army to army. What does that streamer really wants is to have an ability to perform the typical tasks - to speed up heroes and to leave the relatively fast, but weak unit(s) to the previous hero in chain. My proposal covers both cases, your proposal covers just the first case.

LeHerosInconnu:
Hello @oleg-derevenetz,

> @LeHerosInconnu
> 
> > I think that the selection with the red frame of a troop should be taken into account only in the origin army of the transfer, to indicate that it is this troop which will have to be kept in priority (so selecting a troop in the destination army has no effect at all).
> 
> That's a step back - it doesn't take into account chains of heroes, and doesn't implement the logic "move all army from the courier hero and leave him just this unit". That's why selection with red frame of a troop should be taken into account both in the source and in the destination army, just like in this case:
> 
>  https://user-images.githubusercontent.com/32623900/187012272-a2c47abd-2238-4454-af12-bf60f633462f.mp4 

The case your video shows is an army switch.
As it is demonstrated, the way to do this is not very intuitive or practical.
To swap armies, there should be a specific button dedicated to this task, simple and intuitive in its operation, as discussed in this thread: https://github.com/ihhub/fheroes2/discussions/4327.

The transfer button from the left army to the right army should be used only for this, to transfer troops from one army to the other (and not to swap troops from both armies), with the possibility to select a troop to indicate the player's preference for the troop or part of the troop (one creature) to keep in the origin army.
The same goes for the transfer button from the right army to the left army.

> > They tried and they gave up: #5478. :rofl: 
> 
> I don't see any dissatisfaction with the fact that the same units are combined when moving from army to army. What does that streamer really wants is to have an ability to perform the typical tasks - to speed up heroes and to leave the relatively fast, but weak unit(s) to the previous hero in chain **_even they are currently in the destination army_**. My proposal covers both cases, your proposal covers just the first case.

What this streamer wants is a quick and easy way to switch his armies: ;)
https://youtu.be/eJq7L3kO2Ng?t=1924
https://youtu.be/eJq7L3kO2Ng?t=2365
https://youtu.be/eJq7L3kO2Ng?t=15068

oleg-derevenetz:
Hi @LeHerosInconnu

> To swap armies, there should be a specific button dedicated to this task, simple and intuitive in its operation

This PR dispenses with the introduction of additional buttons. BTW these arrows are not separate buttons, they are even not the separate images - they are part of the original big image of the whole meeting screen (and in the original game they are not even clickable).

> What this streamer wants is a quick and easy way to switch his armies (like a one-click button): ;)
https://youtu.be/eJq7L3kO2Ng?t=1924
https://youtu.be/eJq7L3kO2Ng?t=2365
https://youtu.be/eJq7L3kO2Ng?t=15068

What this streamer wants in all these cases is a way to transfer an army from/to the courier hero, and this PR covers all the quoted cases.

LeHerosInconnu:
Hello @oleg-derevenetz, @zenseii, @Branikolog and @ihhub,

> Hi @LeHerosInconnu
> 
> > To swap armies, there should be a specific button dedicated to this task, simple and intuitive in its operation
> 
> This PR dispenses with the introduction of additional buttons. BTW these arrows are not separate buttons, they are even not the separate images - they are part of the original big image of the whole meeting screen (and in the original game they are not even clickable).
> 
> > What this streamer wants is a quick and easy way to switch his armies (like a one-click button): ;)
> > https://youtu.be/eJq7L3kO2Ng?t=1924
> > https://youtu.be/eJq7L3kO2Ng?t=2365
> > https://youtu.be/eJq7L3kO2Ng?t=15068
> 
> What this streamer wants in all these cases is a way to transfer an army from/to the courier hero, and this PR covers all the quoted cases.

The problem we have here is that we are trying to do with two buttons what requires three buttons to be done.
This is the same problem as "windows that are too small", where we are trying to put more elements (because they are needed) in a window that is too small, when the only "obvious" way to do it is to increase the size of the window so that all the elements can be fitted in smoothly in a suitable window.
The size of the window should be adapted to the content of the window, not the other way around (the content of the window should be adapted to the small size of the window).
If we impose constraints on ourselves, which I don't think are necessary, we shouldn't be surprised that it is more complicated to achieve what is necessary.

We have to overcome this "religion" of "we don't touch screens, windows, graphics etc. already present in the original game".

So, the window is too small, let's make the window bigger, we need an extra button, let's add an extra button, we need this, we need that, just do it.

Branikolog:
Hello everyone.
I'm totally agreed with @LeHerosInconnu and he says almost the same, that I posted several days ago in this PR. 

>I think the original layout should be kept as much as possible to avoid the player having to rearrange everything.

>I think that the selection with the red frame of a troop should be taken into account only in the origin army of the transfer, to indicate that it is this troop which will have to be kept in priority (so selecting a troop in the destination army has no effect at all).

>The problem we have here is that we are trying to do with two buttons what requires three buttons to be done.

In my thoughts army swap is a thing, that should be implemented through a separate button. So, @oleg-derevenetz , if you wish, you can leave current destination-army-single-stack-selection as it is now (as temporary solution), but in future I believe we'd add new button (which will cover even a cases, when we have both heroes with full armies 5x5).

Regarding troop merging. In my personal point of view, army arrangement in the destination hero should be kept as much possible as it could. But I don't care if transferred troops are re-arranged and even merged into one stack. The main thing is: If we can put more army into destination hero - lets merge his troops, if there's enough space - no need to merge similar stacks into one.

ihhub:
I would ask to hold any discussions on this pull request for a week till everyone's thoughts are settled because it has become a very lengthy dispute about the changes.

ihhub (approved):

ihhub:
Hi @zenseii , @LeHerosInconnu  and @Branikolog , do you have better ideas about these buttons' behavior except using for swap as this function must be delegated to a separate button? If we cannot give a better than @oleg-derevenetz solution I would suggest to go with it for now.

zenseii:
Hi everyone, 

I think this PR should be merged. In my opinion it only brings improvements to the existing troop moving. The highlighting of troops works as it should and is a clear improvement.

LeHerosInconnu:
Hello @ihhub and @zenseii,

> Hi @zenseii , @LeHerosInconnu and @Branikolog , do you have better ideas about these buttons' behavior except using for swap as this function must be delegated to a separate button? If we cannot give a better than @oleg-derevenetz solution I would suggest to go with it for now.

This should be kept simple and understandable, and it should be remembered that similar arrows will have to be used for artifacts.
For me the one way arrow means: Transfer as many troops (or artifacts) as possible from one hero to the other.

> If we cannot give a better than @oleg-derevenetz solution I would suggest to go with it for now.

> Hi everyone,
> 
> I think this PR should be merged. In my opinion it only brings improvements to the existing troop moving. The highlighting of troops works as it should and is a clear improvement.

We can go with the current version for now.

Branikolog:
Hi all.
@oleg-derevenetz made a huge work with the exchanging logic. Lets merge this PR. For the further enhancements we'll create a new issue then.

ihhub:
@oleg-derevenetz , thank you very much for these improvements!

PR #5798, opened 2022-08-31
Show intro videos like in the original game

ihhub:
Also fix 2 issues with video playback:
- the number of audio channels was always set to 1
- add support for scaled video playback

### DIFF
--- a/src/engine/smk_decoder.cpp
+++ b/src/engine/smk_decoder.cpp
@@ -36,6 +36,7 @@ namespace
 SMKVideoSequence::SMKVideoSequence( const std::string & filePath )
     : _width( 0 )
     , _height( 0 )
+    , _heightScaleFactor( 1 )
     , _fps( 0 )
     , _frameCount( 0 )
     , _currentFrameId( 0 )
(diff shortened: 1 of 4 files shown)


ihhub:
Hi @oleg-derevenetz , this pull request changes are very interesting. clang analyzer complains about potential leaks when I move 1-line method bodes from source file to header which makes no sense I can say...

oleg-derevenetz:
Hi @ihhub there is a visual representation in the `scan-build-result` file in the Artifacts section of the build, but it doesn't make sense to me yet.

ihhub:
@oleg-derevenetz , I ran clang-analysis on my Ubuntu machine and I got this error when I moved methods to the header and that is why it is very interesting. It indeed makes no sense.

oleg-derevenetz:
@ihhub There are reports of false positives of this kind in C++17 mode when `unique_ptr` is returned from a function, for example, this [one](https://github.com/llvm/llvm-project/issues/37524#issuecomment-1148739404). I tried more recent Clang Analyzer version (14), but this false positive is observed in this version too. It looks like it will have to somehow turn off this check using the `__clang_analyzer__` [macro](https://clang-analyzer.llvm.org/faq.html#exclude_code) (or maybe just not to transfer methods to the header for a while) :)

ihhub:
I shall just revert method transfer to header, this is the easiest way to do.

oleg-derevenetz (approved):

PR #5799, opened 2022-08-31
UI cleanup (part 2)

ihhub:
~~- remove hotseat selection for Battle Only mode (it is always in hotseat) as we aren't going to expand this feature~~
- add hotkey for difficulty selection within campaign
- optimize UI buttons and avoid extra rendering calls when they aren't active


oleg-derevenetz:
Hi @ihhub

> remove hotseat selection for Battle Only mode (it is always in hotseat) as we aren't going to expand this feature

Are you sure about this part? In future Battle Only over network may in theory  be possible.

ihhub:
> Hi @ihhub
> 
> > remove hotseat selection for Battle Only mode (it is always in hotseat) as we aren't going to expand this feature
> 
> Are you sure about this part? In future Battle Only over network may in theory be possible.

I can revert these changes if you think it is useful to keep as it is.

oleg-derevenetz:
@ihhub

> I can revert these changes if you think it is useful to keep as it is.

I suggest keeping the current functionality for now. It doesn't seem to bother anyone, right?

ihhub:
> @ihhub
> 
> > I can revert these changes if you think it is useful to keep as it is.
> 
> I suggest keeping the current functionality for now. It doesn't seem to bother anyone, right?

Sure. I will revert the changes.

oleg-derevenetz (approved):

PR #5836, opened 2022-09-15
Generate new default Gift button, move Difficulty button generation, fix pressed evil buttons

zenseii:
This generates a default Gift button that is available for all languages and makes the letter G available for button generation.

Close https://github.com/ihhub/fheroes2/issues/5237

image

image

The generation of the Difficulty button was moved to image_agg.cpp.

An issue with two rows of missing pixels of the right side of the evil campaign screen's pressed buttons was addressed also:

image

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -364,7 +364,7 @@ namespace fheroes2
 
                 break;
             }
-            case ICN::BTNGIFT_EVIL: 
+            case ICN::BTNGIFT_EVIL:
                 _icnVsSprite[id].resize( 2 );
                 for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
                     Sprite & out = _icnVsSprite[id][i];
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -415,19 +415,19 @@ namespace fheroes2
                 for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
                     Sprite & out = _icnVsSprite[id][i];
                     out = GetICN( ICN::EVIL_CAMPAIGN_BUTTONS, 0 + i );
-                    // clean the button
-                    Fill( out, 13 - i, 3 + i, 125, 16, getButtonFillingColor( i == 0, false ) );
+                    // clean the button. Needs to use different color to fill pressed state.
+                    Fill( out, 13 - 2 * i, 3 + 2 * i, 129 + 2 * i, 16, getButtonFillingColor( i == 0, false ) );
                 }
-                //const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fheroes2::FontColor::GRAY };
-                //const fheroes2::FontType pressedFont{ fheroes2::FontSize::BUTTON_PRESSED, fheroes2::FontColor::GRAY };
+                const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fheroes2::FontColor::GRAY };
+                const fheroes2::FontType pressedFont{ fheroes2::FontSize::BUTTON_PRESSED, fheroes2::FontColor::GRAY };
 
-                //const char * text = getSupportedText( gettext_noop( "DIFFICULTY" ), releasedFont );
+                const char * text = getSupportedText( gettext_noop( "DIFFICULTY" ), releasedFont );
 
-                //fheroes2::Text releasedText( text, releasedFont );
-                //fheroes2::Text pressedText( text, pressedFont );
+                fheroes2::Text releasedText( text, releasedFont );
+                fheroes2::Text pressedText( text, pressedFont );
 
-                //releasedText.draw( 10 /*+ ( 50 - releasedText.width() ) / 2*/, 5, _icnVsSprite[id][0] );
-                //pressedText.draw( 11 /*+ ( 50 - pressedText.width() ) / 2*/, 6, _icnVsSprite[id][1] );
+                releasedText.draw( 10 + ( 132 - releasedText.width() ) / 2, 5, _icnVsSprite[id][0] );
+                pressedText.draw( 10 + ( 132 - pressedText.width() ) / 2, 6, _icnVsSprite[id][1] );
 
                 break;
             }
@@ -436,19 +436,19 @@ namespace fheroes2
                 for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
                     Sprite & out = _icnVsSprite[id][i];
                     out = GetICN( ICN::X_CMPBTN, 0 + i );
-                    // clean the button
-                    Fill( out, 13 - i, 3 + i, 125, 16, getButtonFillingColor( i == 0, false ) );
+                    // clean the button. Needs to use different color to fill pressed state.
+                    Fill( out, 4 - i, 3 + i, 132, 16, getButtonFillingColor( i == 0, false ) );
                 }
-                // const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fheroes2::FontColor::GRAY };
-                // const fheroes2::FontType pressedFont{ fheroes2::FontSize::BUTTON_PRESSED, fheroes2::FontColor::GRAY };
+                 const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fheroes2::FontColor::GRAY };
+                 const fheroes2::FontType pressedFont{ fheroes2::FontSize::BUTTON_PRESSED, fheroes2::FontColor::GRAY };
 
-                // const char * text = getSupportedText( gettext_noop( "DIFFICULTY" ), releasedFont );
+                 const char * text = getSupportedText( gettext_noop( "DIFFICULTY" ), releasedFont );
 
-                // fheroes2::Text releasedText( text, releasedFont );
-                // fheroes2::Text pressedText( text, pressedFont );
+                 fheroes2::Text releasedText( text, releasedFont );
+                 fheroes2::Text pressedText( text, pressedFont );
 
-                // releasedText.draw( 10 /*+ ( 50 - releasedText.width() ) / 2*/, 5, _icnVsSprite[id][0] );
-                // pressedText.draw( 11 /*+ ( 50 - pressedText.width() ) / 2*/, 6, _icnVsSprite[id][1] );
+                 releasedText.draw( 4 + ( 132 - releasedText.width() ) / 2, 5, _icnVsSprite[id][0] );
+                 pressedText.draw( 4 + ( 132 - pressedText.width() ) / 2, 5, _icnVsSprite[id][1] );
 
                 break;
             }
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -458,18 +458,18 @@ namespace fheroes2
                     Sprite & out = _icnVsSprite[id][i];
                     out = GetICN( ICN::GOOD_CAMPAIGN_BUTTONS, 0 + i );
                     // clean the button
-                    Fill( out, 13 - i, 3 + i, 125, 16, getButtonFillingColor( i == 0, false ) );
+                    Fill( out, 13, 4 + i, 127, 15, getButtonFillingColor( i == 0 ) );
                 }
-                // const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fheroes2::FontColor::GRAY };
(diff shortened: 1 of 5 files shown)


ihhub:
Hi @zenseii , how about Evil interface? The code should be almost the same as for Good interface.

zenseii:
> Hi @zenseii , how about Evil interface? The code should be almost the same as for Good interface.

Yes, I thought about this and the only difference between them is whether to use `ICN::SYSTEM` or `ICN::SYSTEME`.

I don't think it should be necessary to repeat so much code so I'm thinking of a way to avoid this.

ihhub:
> > Hi @zenseii , how about Evil interface? The code should be almost the same as for Good interface.
> 
> Yes, I thought about this and the only difference between them is whether to use `ICN::SYSTEM` or `ICN::SYSTEME`.
> 
> I don't think it should be necessary to repeat so much code so I'm thinking of a way to avoid this.

And font colors :)

zenseii:
> > > Hi @zenseii , how about Evil interface? The code should be almost the same as for Good interface.
> > 
> > 
> > Yes, I thought about this and the only difference between them is whether to use `ICN::SYSTEM` or `ICN::SYSTEME`.
> > I don't think it should be necessary to repeat so much code so I'm thinking of a way to avoid this.
> 
> And font colors :)

Yes, my thought was that the font color would be adjusted automatically when you have chosen the evil interface but I haven't looked much into that part yet.

I was looking for code code where Evil button font generation is. Still trying to figure out this new approach.

LeHerosInconnu:
Hello @zenseii and @ihhub,

I find the "oblique strokes" a little too present in this version of the letter "G".

In fheroes2:
Compare with the letter "G" already present:

G letter on button 01

ihhub:
Hi @zenseii and @LeHerosInconnu , the algorithm to generate letters is not ideal and as highlighted by @zenseii make some differences in one outer color of a letter. We will fix it later.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -380,6 +410,69 @@ namespace fheroes2
 
                 break;
             }
+            case ICN::BUTTON_DIFFICULTY_ARCHIBALD: {
```
ihhub, line 413: I think we can reduce code duplication, especially with your changes:
```cpp
case ICN::BUTTON_DIFFICULTY_ARCHIBALD:
case ICN::BUTTON_DIFFICULTY_ROLAND:
case ICN::BUTTON_DIFFICULTY_POL: {
    std::array<fheroes2::Rect, 2> clearArea;
    int baseIcnId = ICN::UNKNOWN;
    fheroes2::Point textOffset;

    switch ( id ) {
    case ICN::BUTTON_DIFFICULTY_ARCHIBALD:
        ....
        break;
    case ICN::BUTTON_DIFFICULTY_ROLAND:
        ....
        break;
    case ICN::BUTTON_DIFFICULTY_POL:
        ....
        break;
    default:
        assert( 0 );
        break;
    }
....
```

However, it might be a bit complex to read. So why don't we expand `getSupportedText()` function usage by adding a new function:
```cpp
void renderTextOnButton( fheroes2::Image & releasedState, fheroes2::Image & pressedState, const char * text, const fheroes2::Point & releasedTextOffset, const fheroes2::Point & pressedTextOffset, const int32_t maxTextWidth, const fheroes2::FontColor fontColor )
{
    const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fontColor };
    const fheroes2::FontType pressedFont{ fheroes2::FontSize::BUTTON_PRESSED, fontColor };

    const char * text = getSupportedText( gettext_noop( "DIFFICULTY" ), releasedFont );

    fheroes2::Text releasedText( text, releasedFont );
    fheroes2::Text pressedText( text, pressedFont );

    releasedText.draw( releasedTextOffset.x + ( maxTextWidth - releasedText.width() ) / 2, releasedTextOffset.y, releasedState );
    pressedText.draw( pressedTextOffset.x + ( maxTextWidth - pressedText.width() ) / 2, pressedTextOffset.y, pressedState );
}
```

Then we can use this function to make such calls:
```cpp
renderTextOnButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "DIFFICULTY" ),  { 10, 5 }, { 10, 6 }, 132, fheroes2::FontColor::GRAY );
```
This will be applied to all new buttons.

What do you think?

ihhub (changes requested):
Hi @zenseii , good job! I left few suggestions in this pull request. Would you mind to take a look at them?

src/fheroes2/agg/agg_image.cpp
zenseii: Thanks! It should be implemented the way you suggested now. 

I attempted writing something similar last evening but your way is much tidier and will be easier to adjust later on, not to mention this structure can probably be used in other future similar cases.

src/fheroes2/game/game_campaign.cpp
```diff
@@ -723,62 +723,33 @@ namespace
...
 
-    fheroes2::ButtonSprite getDifficultyButton( const int campaignId, const int buttonIconID, const fheroes2::Point & offset )
+    fheroes2::ButtonSprite getDifficultyButton( const int campaignId, const fheroes2::Point & offset )
     {
-        fheroes2::FontColor fontColor{ fheroes2::FontColor::WHITE };
-        fheroes2::Rect fillingReleasedArea;
-        fheroes2::Rect fillingPressedArea;
-        fheroes2::Point releasedTextPosition;
-        fheroes2::Point pressedTextPosition;
+        fheroes2::Sprite releasedImage;
```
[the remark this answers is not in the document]
zenseii: Done!

src/fheroes2/agg/agg_image.cpp
```diff
@@ -343,20 +349,45 @@ namespace fheroes2
                     Fill( out, 13, 11, 113, 31, getButtonFillingColor( i == 0 ) );
                 }
                 break;
-            case ICN::BTNGIFT_GOOD:
+            case ICN::BTNGIFT_GOOD: {
```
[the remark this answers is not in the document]
zenseii: Done!

src/fheroes2/agg/agg_image.cpp
```diff
@@ -344,19 +363,31 @@ namespace fheroes2
                 }
                 break;
             case ICN::BTNGIFT_GOOD:
+            case ICN::BTNGIFT_EVIL: {
                 _icnVsSprite[id].resize( 2 );
+
+                const bool isGoodInterface = ( id == ICN::BTNGIFT_GOOD );
+                const int baseIcnId = isGoodInterface ? ICN::SYSTEM : ICN::SYSTEME;
+                const fheroes2::FontColor releasedFontColor = ( isGoodInterface ) ? fheroes2::FontColor::WHITE : fheroes2::FontColor::GRAY;
```
ihhub, line 371: Please remove parenthesis for `isGoodInterface` variable.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -344,19 +363,31 @@ namespace fheroes2
...
-                break;
-            case ICN::BTNGIFT_EVIL:
-                _icnVsSprite[id].resize( 2 );
-                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
-                    Sprite & out = _icnVsSprite[id][i];
-                    out = GetICN( ICN::SYSTEME, 11 + i );
+                    out = GetICN( baseIcnId, 11 + i );
                 }
+
+                const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, releasedFontColor };
```
ihhub, line 378: This code should use `renderTextOnButton()` function.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -369,8 +400,7 @@ namespace fheroes2
                 const fheroes2::FontType releasedFont{ fheroes2::FontSize::BUTTON_RELEASED, fheroes2::FontColor::WHITE };
```
ihhub, line 400: This code should use `renderTextOnButton()` function.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -380,6 +410,59 @@ namespace fheroes2
 
                 break;
             }
+            case ICN::BUTTON_DIFFICULTY_POL:
```
ihhub, line 413: Actually your previous implementation was cleaner. I was suggesting at the end to use `renderTextOnButton()` function for the later code to make it cleaner :)

ihhub (changes requested):
Hi @zenseii , I left few more comments here. Could you please check them?

src/fheroes2/agg/agg_image.cpp
zenseii: I see now. Then I'll revert back to the structure where every campaign button is its own case.

ihhub: Great!

zenseii: BTW it seems that with the current `renderTextOnButton()` the translated string is never shown, even if the letters exist?
I'll look into it, but would it have anything to do with using `gettext_noop()`?

ihhub: `gettext_noop()` function literally returns the same string. You forgot to call `getSupportedText` within `renderTextOnButton`.

zenseii: I've added `getSupportedText`, yet the text is still not translated:

As an example, if I translate MIN to DIF:
image
and have  use `_gettext_noop( "MIN" )` in `renderTextOnButton()` then the button will render as `MIN`:
image
If I set `_("MIN")` then it will render as `DIF` :
image

Setting breakpoints I can see that with `gettext_noop()` the parameter `const char * text` of `renderTextOnButton()` has a value of `MIN`, so it would seem that only MIN gets sent from the call to `renderTextOnButton()` when generating the MIN button, no matter what the current translated file in use says.

ihhub: I'll take a look tomorrow.

zenseii: Thanks. I'll keep looking at it to see if I can figure something out.

It seems that the text does get translated in `getSupportedText()`, since the value of `translatedText` changes to `DIF` but still the untranslated MIN is what is ultimately used when rendering...

```cpp
const char * getSupportedText( const char * text, const fheroes2::FontType & font )
    {
        const char * translatedText = _( text );
        return fheroes2::isFontAvailable( translatedText, font ) ? translatedText : text;
    }
```

It seems like something isn't going right at `return fheroes2::isFontAvailable(...)` since `text` aka MIN is the outcome of that statement rather than DIF.

ihhub: For me even with this check everything works fine. I removed it. Could you please check again? Also please do a clean build.

zenseii: I've tested it with a clean build while the comma character check is still there and it still returns false:
image

image
Maybe I've misunderstood but shouldn't the comma character indeed be missing from the `BUTTON_RELEASED` font since we haven't generated that character?

And yes, removing that check makes the translated string show up.

zenseii: Changing it to check for a D makes it not return false so the string gets translated
```cpp
const Sprite & commaCharacter = AGG::getChar( 'D', fontType );
        if ( commaCharacter.empty() ) {
            return false;
        }
```

ihhub: A comma doesn't exist for buttons. I haven't seen any button using it. D is the same random character as well as comma. Removing this check just makes to execute the loop below so no harm except few extra CPU cycles.

ihhub (approved):

src/fheroes2/agg/agg_image.cpp
```diff
@@ -344,39 +365,71 @@ namespace fheroes2
...
-                presesedText.draw( 10 + ( 50 - presesedText.width() ) / 2, 6, _icnVsSprite[id][1] );
+                break;
+            }
+            case ICN::BUTTON_DIFFICULTY_POL: {
+                _icnVsSprite[id].resize( 2 );
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out = GetICN( ICN::X_CMPBTN, 0 + i );
+                    // clean the button.
+                    Fill( out, 4, 3 + i, 132 - i, 16, out.image()[5 * 132] );
```
ihhub, line 429: This is not a part of this pull request but `out.image()[...]` is very dangerous without checks. We need to come up with an extra function which checks the value and throws an exception when we are trying to access out of boundaries array element.

zenseii: If `Blit()` is safer then we can use that instead. Sadly `getButtonFillingColor()` does not work because the Evil campaign buttons have a different color in their pressed state and the Expansion buttons have different colors for both pressed and released.

ihhub: We can just create an issue do not to forget to fix this possible application crash :)

ihhub:

Issue #1250, opened 2020-08-16
Scenario window, elements position


LeHerosInconnu:
- In fheroes2, the text "Rating 100%" is not centered horizontally in the window.
This text is not centered horizontally in the original game either.
For my part, I would prefer that it be centered horizontally in the window.
- In fheroes 2, the edge of the "OKAY" and "CANCEL" buttons are not the same distance from the edge of the window.
These buttons are not the same distance from the edge of the window in the original game either.
For my part, I would prefer that they be at the same distance from the edge of the window.
- The shadows of the difficulty level buttons and the scenario name bar do not have the same offset as in the original game.
4 pixels to the left, 3 pixels to the bottom in the original game.
4 pixels to the left, 4 pixels to the bottom in fheroes2.
For my part, I prefer the evenly distributed shadow of fheroes2.
So for me, no need to change that. :)

- The words "Hard", "Expert" and "Impossible" are not centered in relation to the difficulty level buttons, 1 pixel offset to the left of the word.

- Opponent buttons, hero class buttons and double arrow graphics for changing player order have no shadows.

**Some adjustments to make to improve the visual appearance of this window:**

By taking the title of the window "Scenario:" as a reference and which must remain in its current position.

- Move the position of the difficulty level selection buttons 2 pixels down to prevent the red frame indicating the selected difficulty from covering the text shadow of the letter "y" of the word "Difficulty".

- Put the same vertical space between the median of the line of the text and the element just above (11 pixels).
Example, the bottom of the scenario name graphic and the median of the line of the text "Game Difficulty:".

- Put the same vertical space between the baseline of the line of the text and the element just below (8 pixels).
Example, the baseline of the text "Game Difficulty:" and the top of the graphic of the difficulty level selection button.

- Adjust the vertical spacing between the median of the line of the text "Random" and the bottom of the hero class button graphic (5 pixels).

- The vertical spacing between the bottom of the scenario difficulty level graphic and the median of the line of the text of the difficulty level must remain at 7 pixels.

![Scenario window, improve the visual appearance 01](https://user-images.githubusercontent.com/43583994/92324670-48eaa100-f044-11ea-9251-4dd7bc82dc94.png)

In the original game:

![Scenario window 03](https://user-images.githubusercontent.com/43583994/90333832-4955c600-dfc9-11ea-856d-dfd9cc2e3c48.png)

In fheroes2:

[...]

PR #5839, opened 2022-09-17
Add Handicap feature present in the OG

### LINKED ISSUE #1255: Multiplayer mode, enter the players' handicap
In the original game, in the multiplayer mode, it is possible to enter a personalized name for each player.
In fheroes2 this is not possible.

Multiplayer mode 01

Multiplayer mode 02

In the original game, it is also possible to apply a handicap for each player.

Multiplayer mode 06

**Edit 2022/04/19.**

The operation for the names should remain the same as now (see post below).
Only the handicap choice for the players is missing.

### LINKED ISSUE #4721: Game, multiplayer, difference in functioning compared to the original game
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2 currently, for the multiplayer, it doesn't work like in the original game.

After switching the position of the Blue player and the Orange player.
In the original game, the order of play of the player colors has been changed and the Orange player plays first and at the Orange player position on the adventure map.

https://user-images.githubusercontent.com/43583994/143768348-09041b73-18d7-4f97-a930-b7938ba295b3.mp4

Note that the player color graphic is not updated in the original game when the player swaps the Blue and Orange players.
I think it should be in fheroes2.

In fheroes2, this is not the case.

https://user-images.githubusercontent.com/43583994/143768416-1d24e872-d62c-439c-8f5f-d42dbfd5d950.mp4

### Save file

Not needed.

### Additional info

_No response_

ihhub:
- add Handicap feature selection for human players for Single-player games and for Multi-player games (the original game supports only multi-player game) (close #1255 and close #4721) :
image
- display handicap information in scenario information window:
image
- scenario information window allows a player to right-click on handicap icon to see what it mean (as it is not obvious for a normal player)
- fix Multi class icon display in scenario information window
- add missing shadows for opponent, class and handicap icons (relates to #1250)


zenseii (commented):
Hi @ihhub ,

Since this PR removes the swap arrows in the single player scenario window, then it is highly related to this issue, and might even close it:
https://github.com/ihhub/fheroes2/issues/4721

It is also relevant to this discussion about these arrows:
https://github.com/ihhub/fheroes2/discussions/4701

To summarize these two discussions, these arrows should be removed, however we shouldn't forget about the functionality to change order of players in single player mode, and what we should do with that at a later point, whether to keep it or just not have it like the OG.

ihhub:
Hi @zenseii , these are valid points. We should get rid of these confusing arrows.

src/fheroes2/gui/player_info.cpp
```diff
@@ -199,244 +197,362 @@ bool Interface::PlayersInfo::SwapPlayers( Player & player1, Player & player2 ) c
...
-        const Player & player = *( ( *it ).player );
-        const fheroes2::Rect & rect1 = ( *it ).rect1;
-        const fheroes2::Rect & rect2 = ( *it ).rect2;
-        const fheroes2::Rect & rect3 = ( *it ).rect3;
+    const uint32_t humanColors = conf.GetPlayers().GetColors( CONTROL_HUMAN, true );
 
-        // 1. redraw opponents
+    // We need to render icon shadows and since shadows are drawn on left side from images we have to render images from right to left.
+    for ( auto iter = crbegin(); iter != crend(); ++iter ) {
+        const PlayerInfo & info = *iter;
```
oleg-derevenetz, line 252: What do you think about adding an assertion somewhere here to verify that `info.player` is not `nullptr`?

src/fheroes2/gui/player_info.h
```diff
@@ -31,35 +31,35 @@ namespace Interface
...
+        void UpdateInfo( Players &, const fheroes2::Point & playerTypeOffset, const fheroes2::Point & classOffset );
         bool SwapPlayers( Player & player1, Player & player2 ) const;
 
         Player * GetFromOpponentClick( const fheroes2::Point & pt );
         Player * GetFromOpponentNameClick( const fheroes2::Point & pt );
-        Player * GetFromOpponentChangeClick( const fheroes2::Point & pt );
         Player * GetFromClassClick( const fheroes2::Point & pt );
+        Player * getPlayerFromHandicapRoi( const fheroes2::Point & point );
+
+        /// show_play_info: show game info with color status (play/not play)
```
oleg-derevenetz, line 51: I believe this comment is a bit outdated :) How about this: "displayInGameInfo: use colored and grayscale race icons to distinguish between players who are still in the game and vanquished players respectively"?

src/fheroes2/system/players.h
```diff
@@ -144,8 +168,25 @@ class Player : public BitModes, public Control
...
+
+    void setHandicapStatus( const uint8_t status )
+    {
+        if ( control & CONTROL_HUMAN ) {
+            _handicapStatus = status;
+        }
+        else {
+            _handicapStatus = HandicapStatus::NONE;
+        }
+    }
```
oleg-derevenetz, line 189: How about something like this:

```cpp
if ( status == HandicapStatus::NONE ) {
  _handicapStatus = status;
  return;
}

assert( control & CONTROL_HUMAN );

_handicapStatus = status;
```

? This should help catch potential errors in handicap logic instead of masking them.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions, please take a look when you have time :)

src/fheroes2/gui/player_info.cpp
ihhub: An assertion was added in `void Interface::PlayersInfo::UpdateInfo()` method at the time of inserting these pointers (line 121).

src/fheroes2/gui/player_info.h
ihhub: Yes. It makes sense.

src/fheroes2/system/players.h
ihhub: This is a good point.

ihhub:
Phew. I committed an experimental code by mistake which generated a lot of clang-tidy warnings...

src/fheroes2/gui/player_info.cpp
oleg-derevenetz: Ah yes, they are checked in another function while populating the `PlayersInfo`. If you believe that's enough, I'm fine with that.

oleg-derevenetz (approved):

PR #5854, opened 2022-09-18
Before setting a handicap for a player, make sure that this player cannot be controlled by AI

oleg-derevenetz:
Fix crash due to failed assertion when clicking on the handicap field for a player that can be controlled both by AI and a human player:

handicap

Such players have both the `CONTROL_HUMAN` and `CONTROL_AI` flags at the same time.

Also, this PR adds functionality to remember handicap settings when restarting the same map in the same circumstances, as it has already been done for other player settings.


ihhub:

PR #5866, opened 2022-09-21
Add enhancements link to frontpage

zenseii:
Current text:
```
You can find a complete list of all of our changes and enhancements in its own wiki page (with link to it).
```

The text will currently show up here since talk about enhancements are made there and we want users to see the enhancements so it's better to have on the top rather than further down in my opinion.
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ If you would like to help translating the project, please read the [**translatio
 
 ## Donation
 
-We accept donations via **Patreon** or **PayPal**. All donations will be used only for the future project development as we do not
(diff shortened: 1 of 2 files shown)


oleg-derevenetz (approved):

ihhub:
Hi @zenseii , since we are changing readme files should we make `We accept donations via Patreon or PayPal.` as links? Not everyone know that small banners are clickable.

zenseii:
> Hi @zenseii , since we are changing readme files should we make `We accept donations via Patreon or PayPal.` as links? Not everyone know that small banners are clickable.

Sure! I remember there being some typos that I can fix too.

ihhub (approved):

ihhub:

PR #5867, opened 2022-09-21
fixed CreateDirectoryA for win and msc & icn ICNHeader signed offsets

tarboss:


src/tools/icn2img.cpp
```diff
@@ -105,7 +105,7 @@ int main( int argc, char ** argv )
                 dstfile += ".bmp";
             }
 
-            std::cout << "Image " << ii + 1 << " has offset of [" << static_cast<int32_t>( head.offsetX ) << ", " << static_cast<int32_t>( head.offsetY ) << "]"
+            std::cout << "Image " << ii + 1 << " has offset of [" << static_cast<int16_t>( head.offsetX ) << ", " << static_cast<int16_t>( head.offsetY ) << "]"
```
ihhub, line 108: There is not reason then to have this casting as it is redundant. Please remove it.

src/engine/system.cpp
```diff
@@ -114,7 +114,8 @@ namespace
 int System::MakeDirectory( const std::string & path )
 {
 #if defined( __WIN32__ ) && defined( _MSC_VER )
-    return CreateDirectoryA( path.c_str(), nullptr );
+    bool createDir = CreateDirectoryA( path.c_str(), nullptr );
```
ihhub, line 117: The return value of this function is `BOOL`, not `bool`. Also this variable should be `const`.

src/engine/system.cpp
```diff
@@ -114,7 +114,8 @@ namespace
...
+    return !createDir; // makedir compatible now!
```
ihhub, line 118: We are casting here from bool to integer. While it is still valid I think for better understanding it is good to do like this:
```cpp
// CreateDirectoryA returns 0 in case of failure and nonzero in case of success.
if ( returnValue == 0 ) {
    ERROR_LOG( "Unable to create a directory. Error: " << getLastErrorStr() )
   return -1;
}

return 0;
````

So we need to create a separate function in anonymous namespace at the top of this file with a body which is present in `getLastErrorStr` lambda function to be used in multiple functions.

File: src/engine/agg_file.h [hunk 4]
```diff
@@ -63,8 +63,8 @@ namespace fheroes2
             , offsetData( 0 )
         {}
 
-        uint16_t offsetX;
-        uint16_t offsetY;
+        int16_t offsetX;
```
ihhub, line 66: There is another place where we need to remove casting after changing the type:
```cpp
= decodeICNSprite( data, sizeData, header1.width, header1.height, static_cast<int16_t>( header1.offsetX ), static_cast<int16_t>( header1.offsetY ) );
```

ihhub (changes requested):
Hi @tarboss , I put several comment in this pull request. Could you please take a look?

src/engine/system.cpp
tarboss: yes, but read this:
https://stackoverflow.com/questions/6420526/what-is-the-difference-between-bool-and-bool
(11 years old)
- bool is a built-in C++ type 
- BOOL is a Microsoft specific type int
- const make the line even longer

src/tools/icn2img.cpp
tarboss: i remove the cast complete

src/engine/system.cpp
ihhub: Hi @tarboss , I'm fully aware about this difference and I'm pointing it out that this variable must be `BOOL` aka `int` as this is the value what the function returns. It is always important to avoid implicit conversion of types.

`const` marks that the variable is not modifiable. This is present in most of our code base.

src/engine/system.cpp
tarboss: no fan of lambda, i wana go small steps

src/engine/system.cpp
tarboss: okay chief

src/engine/system.cpp
ihhub: I did **not** ask to create a lambda function, I asked transform existing one into a normal function.

src/engine/system.cpp
tarboss: const BOOL createDir = CreateDirectoryA( path.c_str(), nullptr );

src/engine/agg_file.h
tarboss: u can do it, its in agg_image.cpp, i dont touched that file!

ihhub:
Closing this pull request as most of work was completed in a separate pull request recently.

PR #5868, opened 2022-09-22
UI improvements of the Settings dialog

zenseii:
This changes the settings dialog to allow for double-lined headers and subtexts.
Several strings were adjusted too.

Relates to https://github.com/ihhub/fheroes2/issues/5244

image

image

Whatever issues there are with overlapping texts now should be no problem to solve for the translators.


src/fheroes2/dialog/dialog_game_settings.cpp
```diff
@@ -48,8 +48,8 @@ namespace
 {
     const fheroes2::Size offsetBetweenOptions{ 92, 110 };
 
-    const int32_t titleOffset = 10;
-    const int32_t nameOffset = 10;
+    const fheroes2::Point textOffset = { 11, 12 };
```
ihhub, line 51: `const fheroes2::Point textOffset{ 11, 12 };` will do even better.

src/fheroes2/dialog/dialog_game_settings.cpp
```diff
@@ -73,8 +73,10 @@ namespace
         const fheroes2::Text title( titleText, fheroes2::FontType::smallWhite() );
         const fheroes2::Text name( nameText, fheroes2::FontType::smallWhite() );
 
-        title.draw( optionRoi.x + ( languageRoi.width - title.width() ) / 2, optionRoi.y - titleOffset, display );
-        name.draw( optionRoi.x + ( languageRoi.width - name.width() ) / 2, optionRoi.y + languageRoi.height + nameOffset, display );
+        const int16_t textMaxWidth = 87;
```
ihhub, line 76: Is there a reason why we need to use `int16_t` over `int32_t`?

src/fheroes2/dialog/dialog_game_settings.cpp
```diff
@@ -312,8 +314,9 @@ namespace
             }
             else if ( le.MousePressRight( windowCursorTypeRoi ) ) {
                 fheroes2::Text header( _( "Mouse Cursor" ), fheroes2::FontType::normalYellow() );
-                fheroes2::Text body( _( "Toggle color cursors on/off. Color cursors look nicer, but sometimes don't move as smoothly as black and white ones." ),
-                                     fheroes2::FontType::normalWhite() );
+                fheroes2::Text
+                    body( _( "Toggle colored cursor on or off. The colored cursor looks nicer, but sometimes it does not move as smoothly as a black and white one." ),
```
ihhub, line 318: Actually this statement is not true at all. I think we need to remove the last sentence as from performance point of view both cursor types behave the same.

ihhub (changes requested):
Hi @zenseii , I left several comment here. Would you mind to take a look at them?

src/fheroes2/dialog/dialog_game_settings.cpp
zenseii: Haha, I always thought this last statement was meant as a joke, but yeah we can easily remove it.

src/fheroes2/dialog/dialog_game_settings.cpp
zenseii: Thanks!

src/fheroes2/dialog/dialog_game_settings.cpp
```diff
@@ -270,19 +272,19 @@ namespace
 
             if ( le.MousePressRight( windowLanguageRoi ) ) {
                 fheroes2::Text header( _( "Select Game Language" ), fheroes2::FontType::normalYellow() );
-                fheroes2::Text body( _( "Change language of the game." ), fheroes2::FontType::normalWhite() );
+                fheroes2::Text body( _( "Change the language of the game." ), fheroes2::FontType::normalWhite() );
```
ihhub, line 275: I think for all sentences at lines 108, 275, 281, 287 and in `dialog_resolution.cpp` file, line 107 we have to manually update translation files, otherwise every translation will be lost. These changes are specific only to English so we shouldn't redo translations for other languages.

zenseii: Sure! They shouldn't be many languages that have this translated anyways.

ihhub:
Hi @zenseii , is this pull request ready for re-review?

zenseii:
Hi, @ihhub .
Yes, sorry I completely forgot about this one.

ihhub (approved):

ihhub:

PR #5870, opened 2022-09-23
Polish translation - update

Pulfiarz:
Things that were added/fixed:

- correcting a typo in "Telescope" artifact story;
- changing "DIFFICULTY" button in the campaigns to "TRUDNOŚĆ" (now better than ever, with some funny lines over S and C :D);
- proposing "WYŚLIJ" as "GIFT" button (same goes here, now with Polish letters :) ), it's quite long, I know, but it can fit;
- correcting some pop-ups' sentences;
- changing the names of all of spells, now each word in the name starts with a capital letter;
- translating remaining lines in .po file (mostly crew);
- changing Town Gate spell translation to "Miejska Brama", so it matches the Town Portal translation,
- proposing TYLKO WALKA instead of BITWA on BATTLE ONLY button.

After some discussions also:
- unifying Spell Book name as Księga Zaklęć (previously it was "Księga Magii", "Księga Zaklęć" or "Księga Czarów" - written with small or capital letters);
- changing all Elementals names into both-parts-with-capital-letter;
- treating Petrification like every other spell's name.

And that is probably all.

Waiting for your review :)


files/lang/pl.po
```diff
@@ -26,15 +26,17 @@ msgid ""
 "BATTLE\n"
 "ONLY"
 msgstr ""
+"TYLKO\n"
```
zenseii, line 29: This was Translated to "BITWA" ealier using Polish assets. This is just for information.

files/lang/pl.po
```diff
@@ -26,15 +26,17 @@ msgid ""
...
-msgstr ""
+msgstr "WYSLIJ"
 
 msgid "MIN"
-msgstr ""
+msgstr "MIN"
 
 msgid "DIFFICULTY"
-msgstr "TRUDNOŚĆ"
+msgstr "TRUDNOSC"
```
zenseii, line 39: In other languages I translated this with "level/degree" as in level of difficulty, since it sounded better. This is just a suggestion.

zenseii (commented):
Hi, @Pulfiarz.

I left a couple of comments. Please have a look if you should have the time.

I would also like to ask @Arthusppp to review if you want and have the time to, and leave an approval or a message afterwards.

files/lang/pl.po
Pulfiarz: Well I know, but i have proposed this translation because I started to wonder. Is this "a battle" or "to battle"? I will be 100% fine if the previous translation will stay, though

zenseii: I can't really contribute much to this since it's down to choice of words. I'm sure @Arthusppp can help you figure out what the best wording for this main menu button is.

Also note that if you use Polish assets (videos, menu etc.) then this change will not be seen, Only if you use assets other than Polish will it be seen.

Pulfiarz: But, if I understand correctly, it will be changed in the future, right?

files/lang/pl.po
Pulfiarz: Yes, I was thinking about that kind of solution. But, unfortunately, the Polish word for "level" is rather misleading :-(also degree won't do the trick , but TRUDNOSC isn't too long to fit the button, that is why I give it a shot. But, of course, it's just my opinion. :-)

files/lang/pl.po
zenseii: The biggest issue right now is that the Polish assets have a different button font than the one we're generating, and in my opinion there is reason to try to keep the original Polish font for players who prefer that one. I think we should avoid having a mishmash of button fonts.

I don't know what approach is planned or will be taken in this regard, but of course the aim would be to keep the experience coherent.

Considering that only really the Polish and Russian assets have a different button font than the English original (the Italian and French are mostly equal to the English), maybe one solution could be to generate such Polish button font on-the-fly.

This is all just speculations on my end, and at the end of the day the best translation is the goal.

files/lang/pl.po
zenseii: I see. Well, hopefully it will be possible to allow for a better translation on this button.

zenseii (approved):
Hi, @Pulfiarz.
I decided to review this with deepl so this PR does at least fill our minimum requirement of being approved. However, it is still preferable that someone with understanding of Polish looks over it.

Should you have more changes then you can always commit them to this PR.

files/lang/pl.po
```diff
@@ -9610,7 +9620,7 @@ msgstr ""
 "połowę obrażeń z ataku wręcz."
 
 msgid "You have no Magic Book, so you cannot cast a spell."
-msgstr ""
+msgstr "Nie posiadasz Księgi Magii, więc nie możesz rzucać zaklęć."
```
Arthusppp, line 9623: W innych miejscach używamy "Księgi Zaklęć". Uspójnimy?

Pulfiarz: Uspójniamy w całym pliku :)

Arthusppp (approved):

zenseii (approved):
Hi, @ihhub.

This PR is ready for a merge when you have the time to look at it.

ihhub:
@Pulfiarz , huge thanks for improving Polish translation!

PR #5871, opened 2022-09-23
Generate J, Q, V, W, X, Z, -

zenseii:
This generates all the remaining letters of the latin alphabet plus the hyphen.

J and Z had to be invented since no buttons have them.
image


ihhub:
@zenseii , thanks a lot for improving this!

Issue #1372, opened 2020-08-27
Combat screen, logs list, expression used for the position of hexagons


LeHerosInconnu:
In the logs list, the expressions used for the position of the hexagons are not very indicative.
"Moved Champions: 9, 4", "Moved Champions: 31, 24".

![Expression position of hexagons 01](https://user-images.githubusercontent.com/43583994/91484186-28368480-e8a9-11ea-9157-f697edf5873a.png)

I suggest improving this a bit to make it more understandable.

The horizontal lines of hexagons, starting from the top of the screen, are referenced with the letters of the alphabet.
The first horizontal line of hexagons is referenced with the letter "A".
The second horizontal line of hexagons with the letter "B".
The third horizontal line of hexagons with the letter "C".
Etc.

The hexagons in each horizontal line, starting from the left of the screen, are referenced with numbers.
The first hexagon of each line is referenced with the number "1".
The second hexagon of each line is referenced with the number "2".
The third hexagon of each line is referenced with the number "3".
Etc.

The first hexagon, starting from the left of the screen, of the first line, starting from the top of the screen, is therefore referenced "A1".

A troop moving from this hexagon into the fifth hexagon in the same line would be indicated by the expression "A1, A5".
"Moved Goblins: A1, A5" instead of "Moved Goblins: 0, 4".

A troop moving from the first hexagon into the fifth hexagon of the last line of hexagons would be indicated by the expression "I1, I5".
"Moved Goblins: I1, I5" instead of "Moved Goblins: 88, 92".

For creatures occupying two hexagons, both hexagons are indicated.
"Moved Champions: A10-11, A7-8" instead of "Moved Champions: 9, 6".

And also, I suggest putting the name of the creature first in each sentence.
So to write:
"Goblins move: A1, A5"
"Goblins move: I1, I5"
"Champions move:" A10-11, A7-8"

And write "Champions skip turn" instead of "Champion skipping turn".

Branikolog:
Not all players want to watch this numbers and letters in battle log.
I suggest simply type something like that: "Goblins take a new position"

LeHerosInconnu:
@Branikolog.

> Not all players want to watch this numbers and letters in battle log.
> I suggest simply type something like that: "Goblins take a new position"

If the position information is to be removed, it would be preferable to leave the notion of movement anyway:
"Goblins move to a new position."

Branikolog:
@LeHerosInconnu 

> "Goblins move to a new position."

[...]

PR #5876, opened 2022-09-24
Make logging for monster movement in battle more user-friendly

ihhub:
We still show coordinates in the battlefield but these values are useless.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3318,14 +3318,9 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
 
     Cursor::Get().SetThemes( Cursor::WAR_POINTER );
 
-#ifdef DEBUG_LOG
-    std::string msg = _( "Moved %{monster}: %{src}, %{dst}" );
+    std::string msg = _( "Moved %{monster}: from [%{src}] to [%{dst}]." );
     StringReplace( msg, "%{monster}", unit.GetName() );
-    StringReplace( msg, "%{src}", unit.GetHeadIndex() );
-#else
-    std::string msg = _( "Moved %{monster}" );
-    StringReplace( msg, "%{monster}", unit.GetName() );
-#endif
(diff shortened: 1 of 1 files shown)


zenseii (approved):

zenseii:
Relates to https://github.com/ihhub/fheroes2/issues/1372

oleg-derevenetz:
Hi @ihhub please resolve merge conflicts :)

> We still show coordinates in the battlefield but these values are useless.

Not quite useless, I've already gotten used to translating indexes into coordinates, but some skill is needed :)

zenseii:
I should add that while testing this I noticed that 2-hex creatures like Dragons will report the x-/horizontal coordinate according to where they first land with their head cell. This means for the same pair of cells it might report `[1,1]` or `[1,2]` depending on whether it had to turn around before flying to this cell or not.

It took some time before I understood why this was happening, but this doesn't necessarily have to change.

ihhub:
Hi @oleg-derevenetz,
what if we display extra value in debug mode? I think it should help to preserve your adopted style of conversion.

Hi @zenseii ,

Actually for wide creatures we can display 2 sets of coordinates. Will it help?

oleg-derevenetz:
Hi @ihhub

> what if we display extra value in debug mode? I think it should help to preserve your adopted style of conversion.

I'm fine with new logic :)

zenseii:
> Hi @oleg-derevenetz, what if we display extra value in debug mode? I think it should help to preserve your adopted style of conversion.
> 
> Hi @zenseii ,
> 
> Actually for wide creatures we can display 2 sets of coordinates. Will it help?

This would at least remove some ambiguity, since right now if you are playing "blind" and only looking at the coordinates, you  will have to remember that when moving sideways or forward the x-coordinate is where the head-end is, whereas if you move backwards the x-coordinate is where the tail-end will end up being.

I'm just thinking should we ever want to have a system where you can input the coordinates, like maybe for text support mode, then this might be a minor inconvenience.

oleg-derevenetz (approved):
Changes are fine, but merge conflicts should be resolved.

PR #5878, opened 2022-09-24
Map scroll with left button

Laserlicht:
This allows the map to be moved by holding down the left mouse button.

Based on this pull request: #5435

But with a little different concept.

There is a similar PR with the middle key: #5660

I added also an fix fo: #2332 (the error should now only occur in an unlikely event: When the mouse is pressed on the map, at the minimap the mouse is released and pressed, and moved back to the map)

Seems to work with touch devices as well. Can anyone test this?


zenseii:
> Seems to work with touch devices as well. Can anyone test this?

I've tested and it works well with a Windows 10 touch device 👍

src/engine/localevent.h
```diff
@@ -219,13 +219,23 @@ class LocalEvent
         return MousePressLeft() && ( rt & mouse_pl );
     }
 
+    bool MouseDownLeft() const
```
ihhub, line 222: What is the difference between down and pressed? Logically there are the same.

src/fheroes2/gui/interface_gamearea.h
```diff
@@ -272,6 +277,10 @@ namespace Interface
         void _setCenterToTile( const fheroes2::Point & tile ); // set center to the middle of tile (input is tile ID)
 
         void updateObjectAnimationInfo() const;
+
+        bool _dragScroll;
+        bool _dragScrollHysteresis;
+        fheroes2::Point _dragScrollPos;
```
ihhub, line 283: Shouldn't this member be called as `_startMouseDragPosition`?

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -886,33 +886,53 @@ void Interface::GameArea::QueueEventProcessing()
...
+    }
 
-    // change cusor if need
-    if ( updateCursor || index != _prevIndexPos ) {
-        Cursor::Get().SetThemes( Interface::Basic::GetCursorTileIndex( index ) );
-        _prevIndexPos = index;
-        updateCursor = false;
+    if ( _dragScrollHysteresis ) {
+        SetCenterInPixels( getCurrentCenterInPixels() + _dragScrollPos - mp );
+        _dragScrollPos = mp;
```
ihhub, line 905: In order to keep changes minimal let's just return from this function after this line. Then we won't need to add else block.

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -886,33 +886,53 @@ void Interface::GameArea::QueueEventProcessing()
     LocalEvent & le = LocalEvent::Get();
     const fheroes2::Point & mp = le.GetMouseCursor();
 
-    int32_t index = GetValidTileIdFromPoint( mp );
+    if ( le.MouseDownLeft() ) {
+        _dragScroll = true;
+        _dragScrollPos = mp;
+    }
+    if ( _dragScroll ) {
+        if ( abs( _dragScrollPos.x - mp.x ) > 10 || abs( _dragScrollPos.y - mp.y ) > 10 ) {
```
ihhub, line 894: Please create a const int32_t variable in unnamed namespace within this file and assign to 10 to avoid hardcoded values.

src/fheroes2/gui/interface_gamearea.h
```diff
@@ -272,6 +277,10 @@ namespace Interface
         void _setCenterToTile( const fheroes2::Point & tile ); // set center to the middle of tile (input is tile ID)
 
         void updateObjectAnimationInfo() const;
+
+        bool _dragScroll;
+        bool _dragScrollHysteresis;
```
ihhub, line 282: This is confusing: we have a method `isDragScroll()` which actually returns **_dragScrollHysteresis** instead of **_dragScroll**.
 Was the intention to create 3 members like:
- `_startMouseDragPosition`
- `_mouseDraggingInitiated`
- `_mouseDraggingMovement`
?

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -426,7 +426,15 @@ void Interface::Radar::QueueEventProcessing()
     }
     else if ( le.MouseCursor( rect ) ) {
         // move cursor
-        if ( le.MouseClickLeft() || le.MousePressLeft() ) {
+
+        if ( le.MouseDownLeft() ) {
+            _dragScroll = true;
+        }
+        if ( !le.MousePressLeft() ) {
```
ihhub, line 433: Are these 2 conditions opposite? If yes then we need to add else for the second condition.

src/fheroes2/gui/interface_radar.h
```diff
@@ -84,6 +84,8 @@ namespace Interface
         fheroes2::MovableSprite cursorArea;
         fheroes2::Point offset;
         bool hide;
+
+        bool _dragScroll;
```
ihhub, line 88: This member is not initialized anywhere.

src/fheroes2/gui/interface_gamearea.h
ihhub: Also these variables are not initialized anywhere.

ihhub (changes requested):
Hi @Laserlicht, I put several comments in this pull request. Would you mind to take a look at them?

src/engine/localevent.h
```diff
@@ -359,6 +369,9 @@ class LocalEvent
 
     fheroes2::Point mouse_wm; // wheel movement
 
+    bool mouse_button_ll; // last left
+    bool mouse_button_lr; // last right
```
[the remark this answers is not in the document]
Laserlicht: recommendation adopted

src/engine/localevent.cpp
```diff
@@ -1101,6 +1101,9 @@ LocalEvent & LocalEvent::GetClean()
 
 bool LocalEvent::HandleEvents( bool delay, bool allowExit )
 {
+    mouse_button_ll = MousePressLeft();
+    mouse_button_lr = MousePressRight();
```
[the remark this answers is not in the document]
Laserlicht: recommendation adopted

src/fheroes2/gui/interface_gamearea.h
Laserlicht: recommendation adopted

src/fheroes2/gui/interface_gamearea.cpp
Laserlicht: recommendation adopted

src/fheroes2/gui/interface_gamearea.cpp
Laserlicht: recommendation adopted - is it okay like this?

src/fheroes2/gui/interface_gamearea.h
Laserlicht: recommendation of variable names adopted

initialisation added

src/fheroes2/gui/interface_radar.h
Laserlicht: done

src/engine/localevent.h
Laserlicht: No - not really.

MouseDownLeft acts like an event. The value is only true for one frame. MousePressedLeft is true as long as the mouse is pressed. But exactly this behavior is required to run code once a button is pressed, not every frame.

src/fheroes2/gui/interface_radar.cpp
Laserlicht: See above about the difference between "Down" and "Press"

Laserlicht:
> Hi @Laserlicht, I put several comments in this pull request. Would you mind to take a look at them?

Thank you for your review! Have a look on the comments. :)

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -881,14 +881,36 @@ void Interface::GameArea::SetScroll( int direct )
     scrollTime.reset();
 }
 
+const int32_t minimalRequiredDraggingMovement = 10;
```
ihhub, line 884: Please put it in unnamed namespace at the beginning of the file (line 45). Right now this variable is global variable which should be avoided.

src/fheroes2/gui/interface_gamearea.h
```diff
@@ -272,6 +277,10 @@ namespace Interface
         void _setCenterToTile( const fheroes2::Point & tile ); // set center to the middle of tile (input is tile ID)
 
         void updateObjectAnimationInfo() const;
+
+        bool _mouseDraggingInitiated = false;
```
ihhub, line 281: Please put these members before methods (just before `_middlePoint()`  to avoid mixing methods and members in one section.

ihhub: Also despite of ability to use the in-class initialization of members we should put initialization where all other members have to keep the code consistent.

src/fheroes2/gui/interface_radar.h
```diff
@@ -84,6 +84,8 @@ namespace Interface
         fheroes2::MovableSprite cursorArea;
         fheroes2::Point offset;
         bool hide;
+
+        bool _mouseDragging = false;
```
ihhub, line 88: Also despite of ability to use the in-class initialization of members we should put initialization where all other members have to keep the code consistent.

src/engine/localevent.h
```diff
@@ -359,6 +369,9 @@ class LocalEvent
 
     fheroes2::Point mouse_wm; // wheel movement
 
+    bool isLeftMouseButtonPressed;
+    bool isRightMouseButtonPressed;
```
ihhub, line 373: These members are not initialized in the constructor.

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -426,7 +426,15 @@ void Interface::Radar::QueueEventProcessing()
...
-        if ( le.MouseClickLeft() || le.MousePressLeft() ) {
+
+        if ( le.MouseDownLeft() ) {
+            _mouseDragging = true;
+        }
+        if ( !le.MousePressLeft() ) {
+            _mouseDragging = false;
+        }
+
+        if ( _mouseDragging ) {
```
ihhub, line 437: Shouldn't this code be just
```cpp
if ( le.MousePressLeft() ) {
    const fheroes2::Point & pt = le.GetMouseCursor();
    ...
```
?

According to the logic it does exactly the same.

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -881,14 +881,36 @@ void Interface::GameArea::SetScroll( int direct )
...
+    if ( !le.MousePressLeft() ) {
+        _mouseDraggingInitiated = false;
+        _mouseDraggingMovement = false;
+    }
+
+    if ( _mouseDraggingMovement ) {
+        SetCenterInPixels( getCurrentCenterInPixels() + _startMouseDragPosition - mp );
+        _startMouseDragPosition = mp;
+        return;
+    }
```
ihhub, line 909: What if we have this code:
```cpp
if ( !le.MousePressLeft() ) {
    _mouseDraggingInitiated = false;
    _mouseDraggingMovement = false;
}
else if ( !_mouseDraggingInitiated ) {
    _mouseDraggingInitiated = true;
    _startMouseDragPosition = mp;
}
else if ( abs( _startMouseDragPosition.x - mp.x ) > minimalRequiredDraggingMovement || abs( _startMouseDragPosition.y - mp.y ) > minimalRequiredDraggingMovement ) {
    _mouseDraggingMovement = true;
}
````
?
Then there is no need to have a separate method in **LocalEvent**.

File: src/fheroes2/gui/interface_radar.h (hunk 12 above)
Comment (ihhub) [reply src/fheroes2/gui/interface_radar.h:88]: Also this member might not be needed based on the comment above.

File: src/engine/localevent.h (hunk 1 above)
Comment (ihhub) [reply src/engine/localevent.h:222]: Based on the comments below we might need to revert changes in this class.

File: src/engine/localevent.h [hunk 16]
```diff
@@ -219,13 +219,23 @@ class LocalEvent
...
+    }
+
     bool MousePressRight() const;
 
     bool MousePressRight( const fheroes2::Rect & rt ) const
     {
         return MousePressRight() && ( rt & mouse_pr );
     }
 
+    bool MouseDownRight() const
```
ihhub, line 234: This method as well as `isRightMouseButtonPressed` member are not used anywhere. Please remove them.

ihhub (changes requested):
Hi @Laserlicht , I left several more comments here. Could you please take a look at them?

src/fheroes2/gui/interface_gamearea.cpp
Laserlicht: Yes, basically - it is working. And would be the much more beautiful solution. But it dont fixes #2332 which is really annoing in combination with this functionality.
The current solution fix this problem.

The main problem of the current codebase is, that the radar and the adventure map have a complete seperate input handling/logic. They don't know each other. Because of this they don't know if the drag initiated in the own "module" or in the other. The "MouseDown" is a solution to detect this.

src/fheroes2/gui/interface_radar.cpp
Laserlicht: No - it don't. See below. :)

Laserlicht:
@ihhub I've added an comment. The other things will be fixed later. :)

src/fheroes2/gui/interface_gamearea.cpp
ihhub: I was thinking about this in the past and I have some prototype code for such cases. Give me please some time to come up with a conclusion.

Laserlicht:
@ihhub I think I found a better solution. Have a look on this. 😊

ihhub:
@Laserlicht , can you please rebase your branch from the master branch so we can test it even for Android?

ihhub (approved):

Laserlicht:
> @Laserlicht , can you please rebase your branch from the master branch so we can test it even for Android?

@ihhub Rebase done :)

Laserlicht:
> Hi, @Laserlicht ! There's a cursor flickering, when we reach frames of the adventure map.
> 2022-11-11.11-19-30.mp4

Hi @Branikolog,

Thanks,
Should be gone now. :)

Branikolog:
Hi, @Laserlicht 
There's another issue with cursor while dragging minimap.

https://user-images.githubusercontent.com/55348946/201333417-7ce5a10a-e075-4287-9c52-c59f7f8b4c1e.mp4

oleg-derevenetz:
Hi all,

Please note that this implementation should be verified in the "no interface" mode as well.

Laserlicht:
Hi @Branikolog 
Thank you! You find every inconsistency. 😉
Should be fixed now.

Hi @oleg-derevenetz 
Thank you very much. I hadn't thought of that at all. There was also a bug in it. But should be ok now.

oleg-derevenetz:
Hi @Laserlicht please add either `#include <cstdlib>` or `#include <cmath>` for `abs()` as suggested by the IWYU. [Any of them will do](https://en.cppreference.com/w/cpp/numeric/math/abs).

oleg-derevenetz:
Hi @Laserlicht please do the following:

1. Sync your `master` branch with the upstream;
2. Merge the `master` branch to this PR's branch to incorporate the latest changes;
3. Use `std::abs()`;
4. Use `#include <cstdlib>`. It seems that I need to add both `cstdlib` and `cmath` as aliases in the `iwyu.map`, currently there is only `cstdlib` :(

Laserlicht:
> Hi @Laserlicht please do the following:
@oleg-derevenetz Thanks! 😉

Branikolog:
Hi, @Laserlicht !
Found one more tiny issue. :)

https://user-images.githubusercontent.com/55348946/201467806-ccdc5806-0c93-4c02-97ac-33896003c3a5.mp4

I've also found cursor flickering for "no interface" mode, but not sure it's related to current PR (it happens in master branch too). There's no flickering while scrolling implemented in your PR. 

https://user-images.githubusercontent.com/55348946/201467876-73114404-436a-4414-b87b-1a295e6a298f.mp4

Laserlicht:
Hi @Branikolog! Thank you for testing. :)

> Found one more tiny issue. :)

I think I would call that a feature and not a bug. :)

Could certainly be fixed as well. But I think that doesn't really bother. Otherwise we would have to build in a detection whether the dragging is initiated in the menu area. Which is a bit stupid, because there is nothing that is done with it.

I would leave it in place as it is. :)

Or other opinion?

> I've also found cursor flickering for "no interface" mode, but not sure it's related to current PR (it happens in master branch too). There's no flickering while scrolling implemented in your PR.

Yes, I also think that this is a problem independent of this PR. We should record this in a separate issue. So that we do not mix too much.

zenseii:
Hi @Laserlicht,

I've tested it on Android now and the only issue I found was when dragging outside of the adventure map area, then letting go, and then pressing anywhere else than where you left the adventure map area makes the map view jump somewhere else. It might be similar to what @Branikolog noted above.

I think the best functioning of this would be similar to the View World screen which can be accessed anytime from the scenario information options (by pressing `V`). In that window you can scroll and drag out side of the adventure map view and then press again somewhere else and the viewing area remains the same.

Laserlicht:
Hi @zenseii 

Should now be an acceptable solution for android. optimally, you should not have a border anyway due to a suitable resolution.

The jumps when you release the mouse in the black area should now be gone. If you don't let go of the mouse, you can continue scrolling as soon as you enter the game area again.

By the way I deactivated that the black area can be used for classic scrolling (screen borders). That would have come into conflict here. I did not want to completely disable classic scrolling on android. Since you can also operate android with a mouse.

oleg-derevenetz:
@Laserlicht

> By the way I deactivated that the black area can be used for classic scrolling (screen borders). That would have come into conflict here.

This may be somewhat inconvenient for classic PC users who are used to playing in full-screen mode. I'm not against disabling this mode for Android (this might me even useful for #6092), BUT this should be disabled ONLY for Android. I suggest adding the auxiliary function somewhere in `engine` (maybe @ihhub can suggest where exactly) with the name like `isOutOfScreenScrollingAllowed()` or something like that (maybe with a different name), that will basically return `false` if `ANDROID` is defined and `true` otherwise and check this in the `game_startgame.cpp`.

Laserlicht:
@oleg-derevenetz You are absolutely right. I have completely overlooked the full screen mode on the desktop... Had a look at it in windowed mode....
I will change this.

Laserlicht:
@oleg-derevenetz @ihhub

I've reconsidered. Actually, we shouldn't hardcode this either.

Otherwise we don't have the problem with android, but with all other touch capable devices (e.g. steam deck).

Couldn't we introduce a function whether the last event was a touch event or a classic mouse event? Does something like this work with sdl? And if so, how?

Current games switch e.g. from gamepad to mouse/keyboard (and can adjust the on-screen-helps), if a corresponding event takes place.

oleg-derevenetz:
@Laserlicht

> Couldn't we introduce a function whether the last event was a touch event or a classic mouse event? Does something like this work with sdl? And if so, how?

I suppose, you can add a separate mode for that, say, `MOUSE_TOUCH`. See `enum LocalEvent::flag_t` and logic in `localevent.cpp` for details. But don't forget to clear this new mode in appropriate places.

Branikolog:
> Could certainly be fixed as well. But I think that doesn't really bother. Otherwise we would have to build in a detection whether the dragging is initiated in the menu area. Which is a bit stupid, because there is nothing that is done with it.
> 
> I would leave it in place as it is. :)

I think such tiny GUI flaws spoil overall impression of the product. This bug definitely shouldn't appear in the final release, so if it's possible, could you please fix it as well? The less issues we leave, the better project looks in players' eyes. ;)

Laserlicht:
> > Could certainly be fixed as well. But I think that doesn't really bother. Otherwise we would have to build in a detection whether the dragging is initiated in the menu area. Which is a bit stupid, because there is nothing that is done with it.
> > I would leave it in place as it is. :)
> 
> I think such tiny GUI flaws spoil overall impression of the product. This bug definitely shouldn't appear in the final release, so if it's possible, could you please fix it as well? The less issues we leave, the better project looks in players' eyes. ;)

Sure, I understand your point.

Yeah, I'm just having a little trouble calling that a bug. :)

If a player moves towards the map with the mouse button (or touch) pressed, doesn't he have the intention to trigger map dragging anyway? Or why else he keeps the button pressed?

Laserlicht:
> @Laserlicht The bug happens only if a player accidentally moves cursor out of adventure borders while trying to make a few "steps" scrolling a map. Also, if cursor is over system buttons, everything works ok. The issue is valid only for scrolling arrows.

Ah, now I understand the problem. Thank you! :)

Should also be a bug before this change, right?

I'll see if there is a simple solution for this.

Laserlicht:
> @Laserlicht The bug happens only if a player accidentally moves cursor out of adventure borders while trying to make a few "steps" scrolling a map. Also, if cursor is over system buttons, everything works ok. The issue is valid only for scrolling arrows.



> @Laserlicht The bug happens only if a player accidentally moves cursor out of adventure borders while trying to make a few "steps" scrolling a map. Also, if cursor is over system buttons, everything works ok. The issue is valid only for scrolling arrows.

I think we should deal with that in a separate issue. The PR here is already extensive anyway with everything that somehow belongs together with it.

Since the bug already existed before.

@oleg-derevenetz 
I once took a slightly different approach. Do you know why it's not working right now? I'm not that familiar with SDL right now. But logically it should now forbid scrolling over the edges of the screen (when using touch), right? But it is currently also treated like mouse, for whatever reason.

Branikolog:
Hi, @Laserlicht 
Well, the issue with heroes/castles arrows was before your PR indeed.

One more question about current implementation.
In a case cursor reaches edges of the adventure map, scrolling is no more happening, until we move it back, over the adventure map. In a case, we move cursor over the frame and then move it over the adventure map back, the focus changes instantly to the place, as if we moved map without being freezed for a while. I think these jumps make scrolling not really smooth.
Is it possible to continue updating adventure map focus while moving cursor over a frame? I mean showing single axis movement, depending on the frame part (bottom/upper or left/right frames).

https://user-images.githubusercontent.com/55348946/201950998-41135a64-133e-4c61-8e18-22dbc8af4e5a.mp4

Laserlicht:
> One more question about current implementation. In a case cursor reaches edges of the adventure map, scrolling is no more happening, until we move it back, over the adventure map.

Hi @Branikolog 
Yes, you are right. I have changed the approach a bit. I find now also immediately much better. Especially you should try it in the simplified interface. :)

Laserlicht:
I'm not sure why right now, but the SDL touch events are not fired on my device (win11 with touch). Therefore the detection does not work. :(

Branikolog:
Hi, @Laserlicht 
Right now the scrolling works well on Win + mouse. I have no other issues detected.

zenseii:
@Laserlicht, the problem I described is now solved. Good job! I think this PR has achieved what it set out to do.

Currently on Android there is a conflict with how this works and how there is an "invisible" cursor that is placed wherever you touch. This leads to the screen scrolling automatically in the opposite direction of where you dragged whenever you drag-scroll outside of the screen border because the game will think that you have left the "invisible" cursor to reside near the edge of the screen and it will stay there until you touch somewhere else.

As I wrote, this is more due to the current design of the Android port and how the cursor is always present even when not touching anywhere on the screen.

Laserlicht:
@zenseii 
Yes, I actually wanted to find a solution for the black area. but it didn't really work out that way (see the second last commit).

If an input is made via touch, then I think the complete scrolling over the margins can be skipped. However, the possibility to recognize which input method is currently being used is still missing (not only android uses touch; there are also vita, switch and desktop). Furthermore, Android can also be operated with a mouse.

And I am now also of the opinion that we should fix this not in this PR (but in a separate, following). Because it doesn't really have anything to do with the topic.

@Branikolog
I found another small bug that should be gone now. 😉

But now the PR should actually fit.

Branikolog:
Hi, @Laserlicht 
I'm only concerning about initiating scrolling, while holding cursor over interface buttons, portraits etc.

https://user-images.githubusercontent.com/55348946/202694591-23b250cc-8c75-4038-819a-88a31411999b.mp4

If I'm not mistaken, we were planning to add dragging for hero portraits and castles, to change the order. 
In this case, current implementation can interfere with future logic. Or am I wrong, @ihhub ?

ihhub:
Hi @Branikolog , yes, we have such plans and drag and drop ability will interfere these changes. However, we might completely change the underlying logic for mouse events.

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -432,6 +434,11 @@ void Interface::Radar::QueueEventProcessing()
     const Settings & conf = Settings::Get();
     LocalEvent & le = LocalEvent::Get();
     const fheroes2::Rect & rect = GetArea();
+    const fheroes2::Rect & borderrect = GetRect();
```
ihhub, line 437: Let's call it `borderArea` or `borderRoi` or at least `borderRect` because we use camelCase variable naming :)

Laserlicht: yes, of course :)

Laserlicht:
> If I'm not mistaken, we were planning to add dragging for hero portraits and castles, to change the order. In this case, current implementation can interfere with future logic. Or am I wrong, @ihhub ?

I have changed the implementation a bit. Now it is so that the dragging is started when you hold down the mouse over the UI/buttons. However, the dragging process is only executed when you drag onto the map with the mouse pressed. Should actually be a pretty good compromise, right? :)

Branikolog (approved):
Now, everything works nice!
I'm 100% sure, players would love this new feature.
Good work, @Laserlicht !

ihhub:
@Laserlicht , great work! Thanks a lot for this improvement!

PR #5879, opened 2022-09-25
Add a feature to give human player control to AI (debug build only)

ihhub:
This feature is very useful for checking AI behavior and testing maps. All suggestions are welcome. including those making it as a debug build only option.

It is proven that this is a very useful feature since I found multiple AI bugs (#5880, #5881 and #5882) by just playing one map.

Bonus:
image


ihhub:
Hi @oleg-derevenetz , @idshibanov , @zenseii and @Branikolog , I would like to know your opinion about this feature.

src/fheroes2/system/players.h
```diff
@@ -48,9 +48,9 @@ class Heroes;
 enum
 {
     CONTROL_NONE = 0,
+    CONTROL_REMOTE = 2, /*, CONTROL_LOCAL = CONTROL_AI | CONTROL_HUMAN */
     CONTROL_HUMAN = 1,
-    CONTROL_AI = 4,
-    CONTROL_REMOTE = 2 /*, CONTROL_LOCAL = CONTROL_AI | CONTROL_HUMAN */
+    CONTROL_AI = 4
```
oleg-derevenetz, line 53: Why not sort them here by value, just like in other places?

oleg-derevenetz (changes requested):
Hi @ihhub I just left one question, please take a look when you have time. This feature looks useful indeed, but I suggest to enable it only on debug builds (maybe also check for `IS_DEVEL()`).

src/fheroes2/system/players.h
ihhub: Oops, I missed. The intention was exactly to sort.

ihhub: Fixed!

zenseii:
Hi, @ihhub and @oleg-derevenetz,

I think this is a huge addition and will help speed up the development of the AI greatly. I also believe that in the current state of the AI it should only be available in debug mode.

In the future I think this could be a very good accessibility option for players having motorial limitations, and we've already heard a couple of fans request such a feature, however as long as there are issues like the ones @ihhub discovered, there is no reason to make this available in release builds.

One thing I noticed is that when you want to end AI mode it only gets into effect the next turn and doesn't immediately hand control over to you.

Here's a vid showcasing it although after asking to regain control the human player is in control whenever a castle is visited:
<details>

https://user-images.githubusercontent.com/12501091/192147772-c1846532-764a-482d-871a-47f3621b3fe0.mp4

</details>


Another thing is that any battle will be shown when the AI is in control, and this might be nice to be able to auto-resolve.

ihhub:
Hi @zenseii , I intentionally enable all battles to see how AI behaves. Since at least 2 people express concerns not to include this feature into release build I shall do it.

zenseii:
> Hi @zenseii , I intentionally enable all battles to see how AI behaves. Since at least 2 people express concerns not to include this feature into release build I shall do it.

Yes I can see it being very useful. My idea was just that when working on improving the AI you might be focusing on just the adventure map logic of the AI and not both battle and adventure map at the same time. There are cases where you could want this and cases where you don't want it.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -789,8 +794,22 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
...
             // if the hero is currently moving, pressing any key should stop him
-            if ( isMovingHero )
+            if ( isMovingHero ) {
                 stopHero = true;
+            }
+
+#if defined( WITH_DEBUG )
+            else if ( HotKeyPressEvent( Game::HotKeyEvent::TRANSFER_CONTROL_TO_AI ) ) {
+                if ( fheroes2::showMessage( fheroes2::Text( _( "Warning" ), fheroes2::FontType::normalYellow() ),
+                                            fheroes2::Text( _( "Do you want to transfer control from you into AI?" ), fheroes2::FontType::normalWhite() ),
```
zenseii, line 804:
```suggestion
                                            fheroes2::Text( _( "Do you want to transfer control from you to the AI?" ), fheroes2::FontType::normalWhite() ),
```

zenseii (changes requested):
I also left this small text adjustment. Please have a look.

ihhub:
Hi @zenseii and @oleg-derevenetz , I made this for debug build as of now.

Branikolog:
Hi!

> this might be nice to be able to auto-resolve.

True. Could it work according to the selected battle mode in options? Or would it be better to initiate battle skip by "Q" button right during AI fight?

ihhub:
Hi @zenseii , I updated and expanded the  description. Please take a look.

ihhub:
Hi @zenseii and @Branikolog , as this is only for debug mode we can leave battles as it is. We still have some issues with battles so it is better to show them for this mode.

Branikolog:
@ihhub 
I've noticed, that after cancelling the AI mode, game shows all dialog windows and you need to click them all to proceed.

ihhub:
Hi @Branikolog , I am aware of this and that's why I accepted to move it only for debug build.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -789,8 +794,24 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
...
-            if ( isMovingHero )
+            if ( isMovingHero ) {
                 stopHero = true;
+            }
+
+#if defined( WITH_DEBUG )
+            else if ( HotKeyPressEvent( Game::HotKeyEvent::TRANSFER_CONTROL_TO_AI ) ) {
+                if ( fheroes2::
+                         showMessage( fheroes2::Text( _( "Warning" ), fheroes2::FontType::normalYellow() ),
+                                      fheroes2::Text( _( "Do you want to transfer control from you into the AI? The effect will take place only on the next turn." ),
```
zenseii, line 805: The use of "into" is not correct here.
```suggestion
                                      fheroes2::Text( _( "Do you want to transfer control from you to the AI? The effect will take place only on the next turn." ),
```

zenseii (changes requested):

src/fheroes2/game/game_startgame.cpp
ihhub: Ah, yes. I missed it.

zenseii (approved):

oleg-derevenetz (approved):

idshibanov (approved):
Great change, will help out a ton

zenseii:
Another great way of using this is to make a multiplayer game with a some human players and then give AI control of all of them. This way you can take control of different AI players on the same map at any point.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1651,6 +1655,18 @@ namespace AI
             const std::vector<Game::DelayType> delayTypes = { Game::CURRENT_AI_DELAY };
 
             while ( LocalEvent::Get().HandleEvents( !hideAIMovements && Game::isDelayNeeded( delayTypes ) ) ) {
+#if defined( WITH_DEBUG )
+                if ( HotKeyPressEvent( Game::HotKeyEvent::TRANSFER_CONTROL_TO_AI ) && Players::Get( hero.GetColor() )->isAIAutoControlMode() ) {
+                    if ( fheroes2::showMessage( fheroes2::Text( _( "Warning" ), fheroes2::FontType::normalYellow() ),
+                                                fheroes2::Text( _( "Do you want to regain control from AI?" ), fheroes2::FontType::normalWhite() ),
```
zenseii, line 1661: I believe the text below was intended to be in this dialog since it's the regaining of the control which happens on the next turn.
`The effect will take place only on the next turn.`
```suggestion
                                                fheroes2::Text( _( "Do you want to regain control from AI? The effect will take place only on the next turn." ), fheroes2::FontType::normalWhite() ),
```

src/fheroes2/game/game_startgame.cpp
```diff
@@ -789,8 +794,23 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
...
             // if the hero is currently moving, pressing any key should stop him
-            if ( isMovingHero )
+            if ( isMovingHero ) {
                 stopHero = true;
+            }
+
+#if defined( WITH_DEBUG )
+            else if ( HotKeyPressEvent( Game::HotKeyEvent::TRANSFER_CONTROL_TO_AI ) ) {
+                if ( fheroes2::showMessage( fheroes2::Text( _( "Warning" ), fheroes2::FontType::normalYellow() ),
+                                            fheroes2::Text( _( "Do you want to transfer control from you to the AI? The effect will take place only on the next turn." ),
```
zenseii, line 804: Related to the above comment: the AI takes control immediately, not the next turn.
```suggestion
                                            fheroes2::Text( _( "Do you want to transfer control from you to the AI?" ),
```

zenseii (commented):
Hi, @ihhub.

I believe some text was placed in the wrong dialog. Please have a look when you have the time to.

src/fheroes2/game/game_startgame.cpp
ihhub: It is on the next turn. Please check this video:

https://user-images.githubusercontent.com/19829520/192236982-848c8148-005a-493c-a3c2-ec748f75c78c.mp4

zenseii: Ah you're right. My bad then. It looked like it happened on the same day for me.

ihhub: Sure. I will add an extra sentence.

zenseii (approved):

PR #5886, opened 2022-09-25
Generate button fonts for different code pages

zenseii:
Changes:

1. Added functions that generates extra needed fonts if necessary.
2. A comment was changed to make it not sound like we have to manually write code for letters of the pressed state too:
`// The downside of this font is that code is necessary for the generation of released and pressed states of each letter.`
3. Minor spelling corrections in comments I discovered.
4. Generated all currently needed button letters needed for Norwegian currrently.
5. Gave a basic solution for how letters with diacritics need to be offset in y-axis, like É or in this case Å.

This relates to:
https://github.com/ihhub/fheroes2/pull/5875

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -2453,7 +2453,7 @@ namespace fheroes2
                 return errorImage;
             }
 
-            // TODO: correct naming and standartise the code
+            // TODO: correct naming and standardize the code
             switch ( fontType ) {
             case Font::GRAY_SMALL:
                 return GetICN( ICN::GRAY_SMALL_FONT, character - 0x20 );
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -375,7 +375,7 @@ namespace fheroes2
                     Fill( out, 13, 11, 113, 31, getButtonFillingColor( i == 0 ) );
                 }
 
-                renderTextOnButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "BATTLE\nONLY" ), { 12, 5 }, { 10, 5 }, { 117, 47 },
+                renderTextOnButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "BATTLE\nONLY" ), { 12, 5 }, { 11, 6 }, { 117, 47 },
                                     fheroes2::FontColor::WHITE );
 
                 break;
@@ -419,7 +419,7 @@ namespace fheroes2
                     Fill( out, 13 + 2 * i, 3 + 2 * i, 129 - 2 * i, 16, out.image()[13 - 7 * i + ( 5 + i ) * ( 145 - ( 4 * i ) )] );
                 }
 
-                renderTextOnButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "DIFFICULTY" ), { 11, 5 }, { 11, 6 }, { 131, 17 },
+                renderTextOnButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "DIFFICULTY" ), { 11, 5 }, { 12, 6 }, { 131, 17 },
                                     fheroes2::FontColor::GRAY );
 
                 break;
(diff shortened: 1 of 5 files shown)


ihhub (changes requested):
Hi @zenseii , I put few comments in this pull request. Could you please take a look?

zenseii:
Hi, @ihhub.

I've written it in the proper approach now. Note that when doing this I initially thought I could get away with just preserving `ICN::BUTTON_GOOD_FONT_RELEASED` but what happened was that all pressed and the evil button fonts would get cleared as soon as a another language than English was set. 

Reapplying effects to the entire set of letters after changing alphabet, so letters 0-255, wasn't something I found a way to do since at that point the original good released button font, which the others would be generated from, would already have the effects applied to it.

Anyways this is ready for your review when you should have the time and would feel like looking at it.

src/fheroes2/gui/ui_font.cpp
```diff
@@ -2571,4 +2619,53 @@ namespace fheroes2
...
+            // Add new language generation code!
+            assert( 0 );
+            break;
+        }
+
+        icnVsSprite[ICN::BUTTON_GOOD_FONT_PRESSED].resize( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED].size() );
+        icnVsSprite[ICN::BUTTON_EVIL_FONT_RELEASED].resize( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED].size() );
+        icnVsSprite[ICN::BUTTON_EVIL_FONT_PRESSED].resize( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED].size() );
+        // Apply effects to new letters that are added on top of the base font's 96 letters.
+        for ( size_t i = 96; i < icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED].size(); ++i ) {
```
zenseii, line 2657: The initial thought behind using the magic number 96 here was that thr base 96 letters (0-95) in any of the code pages are always the English/Latin ones.

However, I'd prefer for the Polish font generation to be able to regenerate these letters as well to match the Polish button font. This would probably also require a different set of effects applied to the letters however.

ihhub: Hi @zenseii , why don't we just call `generateBaseButtonFont()` function to run through all letters? This will avoid any hardcoded values as well as corner cases as Polish?

ihhub: Ah, my bad. We're generating the code there as well... I think we can do something else...

ihhub: Hi @zenseii , I pushed one commit to simplify your work. Please take a look.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2615,6 +2631,7 @@ namespace fheroes2
                 // Restore original letters when changing language to avoid changes to them being carried over.
                 alphabetPreserver.restore();
                 generateAlphabet( language, _icnVsSprite );
+                generateExtraButtonFont( language, _icnVsSprite );
```
ihhub, line 2634: To keep the similar naming I suggest to use `generateButtonAlphabet()`.

zenseii: Yep I agree.

src/fheroes2/gui/ui_font.cpp
zenseii: Maybe it's better to preserve the released font before it has the effects applied to it?

ihhub: How about calling `generateGoodButtonFontBaseShape();` at the start of `generateExtraButtonFont()` function? I think it solves the problem.

zenseii: Ah, of course. That's better.

ihhub (approved):

zenseii:
@ihhub, should we add the empty Code Page functions for the other languages while working on them? 
There is a clang-tidy comment about this above.

ihhub:
> @ihhub, should we add the empty Code Page functions for the other languages while working on them, or should I do it now in this PR? There is a clang-tidy comment about this above.

You can add them during your work. No need to add empty functions.

ihhub:
@zenseii , great job! Thanks a lot for doing this!

PR #5900, opened 2022-09-29
Decrease monster growth bonus for AI

ihhub:
The full explanation is in comments within the code to make sure that everyone in the future will understand magical numbers within the code.

This is subject to discussion and rejection so I'm marking this pull request without milestone for now.

relates to #5797

### DIFF
--- a/src/fheroes2/game/difficulty.cpp
+++ b/src/fheroes2/game/difficulty.cpp
@@ -24,6 +24,8 @@
 #include "difficulty.h"
 #include "translations.h"
 
+#include <cassert>
+
 std::string Difficulty::String( int difficulty )
 {
     switch ( difficulty ) {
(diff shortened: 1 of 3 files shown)


src/fheroes2/game/difficulty.cpp
```diff
@@ -78,16 +80,35 @@ double Difficulty::GetGoldIncomeBonus( int difficulty )
...
+        return 1.0;
     case Difficulty::HARD:
-        return 1.2;
+        return 1.15;
     case Difficulty::EXPERT:
-        return 1.32;
+        return 1.26;
     case Difficulty::IMPOSSIBLE:
-        return 1.44;
+        return 1.37;
```
ihhub, line 108: For comparison: Peasants have 22 unit per week growth. With old 1.44 multiplier we had 22 * 1.44 = 31.68 (rounded to 31) peasants per week. With new multiplier it is 22 * 1.37 = 30.14 (rounded to 30). So 5% difference results in 1 peasant :)

zenseii:
Hi @ihhub.

I've looked at the code and the linked to discussion. I'll add my general thoughts in the discussion since this is likely not to be the last PR on this topic and therefore the discussion will continue on.

About the particular change of this PR, I think it is fine to go ahead with.

Maybe the only question is if the AI is ready for it to be reduced even more? I haven't played any proper maps against the AI in recent days, so I can't say.

zenseii (approved):

Branikolog (approved):
I have nothing against reducing AI perks. We're constantly improving the code, so the day, when AI becomes strong will happen anyway.

ihhub:
Hi @zenseii , I chose 5% as this number is usually a level of noise and many systems / processes / statistical distribution which means that it **almost** doesn't affect the outcome but it still lower than the original. In other words. yes, it is smaller but the difference in 1 peasant per castle is neglectable.

oleg-derevenetz (approved):

PR #5903, opened 2022-09-29
Generate initial CP1250 and CP1251 button fonts

zenseii:
This generates letters for the CP1252 encoding, which includes fonts for Polish, Hungarian and Czech.
Only Ś and Ć are made, the rest are replaced with their closest Latin couterparts since only those two letters are currently required by the translations we have.

Initial CP1251 is made by replacing with closest Latin alphabet counterparts and doing small edits to some of them.

CP1250, Polish:
image

CP1251, Cyrillic:
image

image

PR #5906, opened 2022-09-30
Add an overloaded Flip() function

ihhub:
This function is needed to avoid extra memory copies in few places, mostly font generation. It also improves performance of Adventure Map rendering.

relates to #5903
relates to #5716


zenseii (approved):
This looks good, @ihhub! We will have to remember to replace the previous used approach with this new one in the future.

oleg-derevenetz (approved):

PR #5907, opened 2022-09-30
Add proper icons for Petrification spell

### LINKED ISSUE #4336: Battle screen, Medusa's "Petrification" requires a new icon
### Preliminary checks

Currently we have a original icon for this effect. But its design is poor and ugly
image
We also need this icon to be shown in spell description window:
image

A nice starting issue for an artist.
An icon should be created using Heroes 2 palettes. It can look, for example, like some sort of stone object or similar to current implementation figure made of stone. Also, it can be a picture of medusa's face, as the effect is inspired by the medusa's ability to turn creatures into a stone.

### Additional info

_No response_

ihhub:
Petrification spell had a very ugly mini icon and no icon for the spell itself. This is how it looks in monster info dialog:
image

and spell information window:
image

The whole art work was done by @1GodRage

close #4336


src/fheroes2/agg/agg_image.cpp
```diff
@@ -893,6 +893,10 @@ namespace fheroes2
 
                     AddTransparency( image, 1 );
                 }
+
+                // Petrification spell does not have its icon in the original game.
```
zenseii, line 897:
```suggestion
                // The Petrification spell does not have its own icon in the original game.
```

zenseii (changes requested):
The rest of the code looks great so I only have this small comment to add.

src/fheroes2/agg/agg_image.cpp
ihhub: Fixed.

oleg-derevenetz (approved):

LeHerosInconnu:
Hello @ihhub,

Why choose icons with different postures for the same spell?
All the other spells have the same icons at different sizes.

There is also a smaller icon, but it is not yet used in fheroes2:
In original game (with the sprite corrected):
Medusa stone effect 02

ihhub:
Hi @LeHerosInconnu , the problem is that the original size icon is too wide so we had to use slightly different posture to keep a square proportion of the icon.

Yes, we don't use micro-icons anywhere in fheroes2.

ihhub:
@LeHerosInconnu , I believe we can improve it later. I see no issues not doing it.

LeHerosInconnu:
@ihhub.

> Hi @LeHerosInconnu , the problem is that the original size icon is too wide so we had to use slightly different posture to keep a square proportion of the icon.

Why not also adjust the sprite position of the largest icon to be similar to the medium sized icon then? :)

> Yes, we don't use micro-icons anywhere in fheroes2.

Is this not planned in the future?

> @LeHerosInconnu , I believe we can improve it later. I see no issues not doing it.

Okay.

ihhub:
@LeHerosInconnu , after looking at both sprites I realized that we can adjust sword's position a bit on the large image and leg position and sword on the smaller icon. I am sure we can do something even better than we have right now!

Branikolog:
Hi, @LeHerosInconnu 
There're some original icons, which are different:
image
image
Look at the sword of "Haste" spell.

LeHerosInconnu:
Hello @Branikolog.

> Hi, @LeHerosInconnu There're some original icons, which are different: image image
Look at the sword of "Haste" spell.

Yes, it's true.
But the general allure remains the same, only the sword has been shortened. :)

PR #5909, opened 2022-09-30
Add Romanian translation

zghp:
Translated until line 725


files/lang/ro.po
```diff
@@ -23,646 +23,706 @@ msgid ""
...
 msgid ""
 "A legion of\n"
 "%{monster}"
 msgstr ""
+"O legiune de\n"
+"%{monster}"
 
 msgid "army|Few"
-msgstr ""
+msgstr "armată|Câțiva"
```
zenseii, line 124: The part before the `|` needs to be left in its untranslated English because this is only there to give context and distinguish it from similar text elsewhere in the code, and it will actually be removed before shown in the game.

In other words the correct way to translate this would be:
```suggestion
msgstr "Câțiva"
```

Could you also correct this for the other lines below please?

zenseii (changes requested):
Hi, @zghp !

Thank you so much for this contribution. This translation is very well made and I only found one small mistake that if you have the time and want to it would be great if you could take a look at.

Don't hesitate to ask if you have any doubts.

files/lang/ro.po
```diff
@@ -23,793 +23,892 @@ msgid ""
 "BATTLE\n"
 "ONLY"
 msgstr ""
+"DOAR\n"
+"BĂTĂLIE"
 
 msgid "GIFT"
-msgstr ""
+msgstr "CADOU"
```
zenseii, line 30: This translation seems to be "a gift" or "a present". The button this text appears on refers to "to gift" or "to give". You use it to open the dialog where you can send/give/gift resources to other players.

files/lang/ro.po
```diff
@@ -23,793 +23,892 @@ msgid ""
...
-msgstr ""
+msgstr "Câteva"
 
 msgid "army|Several"
-msgstr ""
+msgstr "Mai multe"
 
 msgid "army|Pack"
-msgstr ""
+msgstr "O haită de"
```
zenseii, line 130: The reason we have two instances of "Pack", "Lots" etc. is because they appear in different places. In this case in this window which has limited space between creatures and that's why in English it is written as "Pack" and not "A pack of".

Would "Haită" work?

Also this is relevant for the other strings of this `army|` context too.

image

zenseii (changes requested):
Hi @zghp.

I left some more comments. It'd be great if you could have a look at them. Thanks!

zenseii:
Also, @zghp, could you please change the PR title to something more intuitive like "Add Romanian translation" or similar?
Thanks!

files/lang/ro.po
zghp: Made the requested changes -- one thing to notice, for "mai multe" and "extrem de multe" I couldn't find a better and shorter translation yet, everything that I could think of so far would change the meaning. What I can do is make  'army|zounds' as "O, Doamne!" maybe but for monsters it should be left as "extrem de multi". 

Please let me know what you think:)

zenseii: Hi @zghp,

I think these pictures describe best what the problem is:
image
image


Currently we have no way of enlarging these dialogs so preferrably these strings should be max 8 characters long.

zenseii: As for having different for army versus wandering creatures, I'm not sure if it is good from a player perspective since they would have to remember that those terms describe the same quantity.

In English they are:

Few - 1-4
Several - 4-9
Band  - 10-19
Lots - 20-49
Horde - 50-99
Throng - 100-249 
Legion - 250-499
Zounds - 500-999
Swarm - 1000++

zenseii: In the worst case we can do like in the French translation and just have quantities written out in numbers like `(1 - 4)` for armies and use text for wandering creatures. This had to be done because no short enough words were found.

This would therefore remain like this (ignore the fact that I haven't implemented all the special characters yet):
image

zenseii: I've looked over your entire PR now and the only part that remains are these two strings that are too long. Leave a comment or re-request a review once you want me to take a look at it again.

It is up to you to decide which approach we take for this problem.

zenseii: I saw that finding an expression that fits is tough for this one. One example of a more modern English variant is "Yikes!" which conveys the same idea as "Zounds" should do, as you can see in the picture below. Yes in English "Zounds of peasants" makes just as much sense as "Yikes of peasants", but that's part of the underlying joke. Like "OMGs of peasants". I guess "Oof!" would be even more modern. It doesn't have to be a literal translation as long as the meaning is sort of preserved. In other words, I'd be surprised if there's a euphemism in Romanian that stems from "God's wounds" just like Zounds originally does.

image

And yes, any experienced Heroes player would go "Yikes!" if they saw such a high number of those troops.

ihhub:
Hi @zghp , I fixed merge conflict for you.

ihhub (approved):
Hi @zghp , the changes look good to me.

zghp:
replaced the army[zounds, swarm, legion, etc.. ] with the equivalent numerals for now as i couldn't find appropriate translations that would also be sort of equal in length (otherwise they would look bad displayed on a player's screen)

zenseii:
> Hi, replaced the army[zounds, swarm, legion, etc.. ] with the equivalent numerals for now as i couldn't find appropriate translations that would also be sort of equal in length (otherwise they would look bad displayed on a player's screen)

Don't worry about it! Could you please add parentheses around the numbers? Ex. (1-4) . This will look better in game because the numbers will be properly separated.

zghp:
fixed it:)

zenseii (approved):
Hi @ihhub.

This PR is ready for your review.

Note that numbers were chosen for the same reason as the official French translation, i.e. no word short enough could be found at this moment.

ihhub:
@zghp , thank you so much for your contribution to the project!

PR #5912, opened 2022-10-01
Do not ask Player's name if the score is too low for High score board

### LINKED ISSUE #5892: “Your name” shouldn’t be asked every time you win a game
### Preliminary checks

“Your name” will be requested even if the score cannot get into the top 10 high scores when you win a game.

“You name” should not be asked when the score is not high enough, and directly return to the start screen.

### Additional info

_No response_

ihhub:
close #5892

### DIFF
--- a/src/fheroes2/game/game_highscores.cpp
+++ b/src/fheroes2/game/game_highscores.cpp
@@ -189,6 +189,7 @@ fheroes2::GameMode Game::DisplayHighScores( const bool isCampaign )
             assert( !campaignHighscoreData.empty() );
 
             if ( campaignHighscoreData.back().rating > rating ) {
+                gameResult.ResetResult();
                 return fheroes2::GameMode::MAIN_MENU;
             }
         }
@@ -198,6 +199,7 @@ fheroes2::GameMode Game::DisplayHighScores( const bool isCampaign )
             assert( !standardHighscoreData.empty() );
 
             if ( standardHighscoreData.back().rating > rating ) {
+                gameResult.ResetResult();
                 return fheroes2::GameMode::MAIN_MENU;
             }
         }
--- a/src/fheroes2/game/game_highscores.cpp
+++ b/src/fheroes2/game/game_highscores.cpp
@@ -132,7 +132,6 @@ namespace
         switch ( difficulty ) {
         case Campaign::CampaignDifficulty::Easy:
             return daysPassed * 125 / 100;
-            break;
         case Campaign::CampaignDifficulty::Normal:
             // Nothing we need to do here.
             return daysPassed;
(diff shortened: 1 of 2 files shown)


src/fheroes2/game/game_highscores.cpp
```diff
@@ -159,10 +179,38 @@ fheroes2::GameMode Game::DisplayHighScores( const bool isCampaign )
...
 
+    const bool isAfterGameCompletion = ( ( gameResult.GetResult() & GameOver::WINS ) != 0 );
+    bool askForPlayerName = true;
+    if ( isAfterGameCompletion ) {
+        // Check whether the game result is good enough to be put on high score board. If not then just skip showing the player name dialog.
+        if ( isCampaign ) {
+            const Campaign::CampaignSaveData & campaignSaveData = Campaign::CampaignSaveData::Get();
+            const uint32_t rating = getCampaignRating( campaignSaveData );
+            const auto & campaignHighscoreData = highScoreDataContainer.getHighScoresCampaign();
+            assert( !campaignHighscoreData.empty() );
```
oleg-derevenetz, line 190: I hope that there are no players left with `hgs` files created long ago and therefore empty :)

oleg-derevenetz (approved):

src/fheroes2/game/game_highscores.cpp
ihhub: Hi @oleg-derevenetz , I think we need to expand `HighScoreDataContainer::load()` method to exclude this case. I can do it in this pull request or separate one.

oleg-derevenetz: Hi @ihhub it's up to you :) I suppose that the probability of this scenario is very low.

zenseii (commented):
Hi @ihhub.

This looks good. I tested a game and got a low enough score. However, the high score board is still shown, which I noticed doesn't happen in the OG. You just go straight back to the main menu if you have a low score.

Also with the current implementation the player will not even be told what score they got, since the end of the dragon animation doesn't show this yet. This was already a problem before this PR anyways.

src/fheroes2/game/game_highscores.cpp
ihhub: Done.

zenseii (changes requested):
Hi, @ihhub.

I tested getting a low score and when you end up on the main menu it is impossible to open the high score from there.
Here's a save right before winning with a low score (not sure how much).
[LowScoreTest.zip](https://github.com/ihhub/fheroes2/files/9707803/LowScoreTest.zip)


<details>

https://user-images.githubusercontent.com/12501091/193847785-48e5e539-b321-4fa9-8273-c6aa6a827733.mp4
</details>

EDIT: All other dialogs can be opened. It is just the high score.

ihhub:
Hi @zenseii , thanks a lot for pointing the problem. It is fixed now.

zenseii (approved):
I've tested it and it works as it should.

PR #5917, opened 2022-10-01
Add Romanian font and language option

zenseii:
This adds ISO8859-16 font used by Romanian and adds the language as a choice in the game.

Button fonts will be added in a later PR when we know what characters are needed.
image

image

### DIFF
--- a/src/fheroes2/gui/ui_font.cpp
+++ b/src/fheroes2/gui/ui_font.cpp
@@ -2521,17 +2521,41 @@ namespace
         {
             std::vector<fheroes2::Sprite> & font = icnVsSprite[ICN::SMALFONT];
 
-             // A with grave accent ` and generate the grave accent for further use.
-            font[192 - 32].resize( font[33].width(), font[33].height() + 3 );
-            font[192 - 32].reset();
-            fheroes2::Copy( font[33], 0, 0, font[192 - 32], 0, 3, font[33].width(), font[33].height() );
-            font[192 - 32].setPosition( font[33].x(), font[33].y() - 3 );
-            fheroes2::Copy( font[192 - 32], 3, 4, font[192 - 32], 7, 0, 1, 1 );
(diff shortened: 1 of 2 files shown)


src/fheroes2/gui/ui_font.cpp
```diff
@@ -3267,6 +3455,9 @@ namespace fheroes2
         case SupportedLanguage::Turkish:
             // generateGoodCP1254ButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED] );
             break;
+        case SupportedLanguage::Romanian:
+            // generateGoodISO8859-16ButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED] );
```
zenseii, line 3459:
```suggestion
            // generateGoodISO8859_16ButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED] );
```

src/fheroes2/gui/ui_font.cpp
```diff
@@ -2427,6 +2427,190 @@ namespace
         }
     }
 
+    void generateISO8859_16Alphabet( std::vector<std::vector<fheroes2::Sprite>> & icnVsSprite )
```
zenseii, line 2430: I'm not sure this is the correct naming according to our standards, but the font name is ISO8859-16 and hyphen cannot be used. Having the underscore would make it easier should we decide to add more ISO8859 fonts other than nr.16

zenseii:
Hi, @ihhub.

This PR is ready for you to review when you have the possibility. I left a comment regarding doubts I had about the function name. Thanks!

src/fheroes2/gui/ui_font.cpp
ihhub: It's fine to use such name.

ihhub:
@zenseii , thanks a bunch for adding this change!

PR #5924, opened 2022-10-05
Implement Army Swap

zenseii:
A new attempt at the army swapping in this PR https://github.com/ihhub/fheroes2/pull/5456 after help from @ihhub and @oleg-derevenetz's explanations in that PR.
The move operator has been implemented, but the previous team's assignment/copy operator is still left as is. I've attempted to fix it several times up until now, but to no avail.

I'm marking it as draft since there's still one issue I need to figure out with the army bars not visually updating after the swap.
It can be seen here:
<details>
<summary>Click to show.</summary>

https://user-images.githubusercontent.com/12501091/194065756-eec9778a-6eef-4752-a175-931f715fd9cf.mp4
</details>

Also I'll implement a better UI solution for how to do the swap then the current right-click as commented in the previous PR.

Relates to https://github.com/ihhub/fheroes2/issues/4084.

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -207,6 +207,13 @@ Troops & Troops::operator=( const Troops & rhs )
     return *this;
 }
 
+Troops & Troops::operator=( Troops && troops )
+{
+    static_cast<std::vector<Troop *> &>( *this ) = std::move( troops );
+
+    return *this;
+}
+
 Troops::~Troops()
 {
     for ( iterator it = begin(); it != end(); ++it )
@@ -500,6 +507,15 @@ void Troops::MoveTroops( Troops & from, const int monsterIdToKeep )
 
     // Then, try to move as much of monsters to keep as we can (except the last one), if there is still a place for them
     moveTroops( false );
+}
+
+void Troops::swap( Troops & troops )
+{
+    if ( !isValid() || !troops.isValid() ) {
+        return;
+    }
+
+    std::swap( static_cast<std::vector<Troop *> &>( *this ), static_cast<std::vector<Troop *> &>( troops ) );
 }
 
 bool Troops::AllTroopsAreTheSame() const
--- a/src/fheroes2/army/army.h
+++ b/src/fheroes2/army/army.h
@@ -48,6 +48,7 @@ public:
     Troops( const Troops & troops );
     virtual ~Troops();
     Troops & operator=( const Troops & rhs );
+    Troops & operator=( Troops && troops );
 
     void Assign( const Troop *, const Troop * );
     void Assign( const Troops & );
@@ -86,6 +87,7 @@ public:
     // Implements the necessary logic to move unit stacks from army to army using the arrow buttons in the
     // hero's meeting dialog
     void MoveTroops( Troops & from, const int monsterIdToKeep );
+    void swap( Troops & troops );
 
     void MergeTroops();
     Troops GetOptimized() const;
--- a/src/fheroes2/castle/castle_dialog.cpp
+++ b/src/fheroes2/castle/castle_dialog.cpp
@@ -446,6 +446,19 @@ Castle::CastleDialogReturnValue Castle::OpenDialog( const bool readOnly, const b
                 need_redraw = true;
             }
 
+            if ( HotKeyPressEvent( Game::HotKeyEvent::SWAP_ARMIES ) || ( !heroes.Guard() && le.MouseClickLeft( rectSign1 ) )
+                 || ( heroes.Guard() && le.MouseClickLeft( buttonSwap.area() ) ) ) {
+                GetArmy().swap( heroes.Guest()->GetArmy() );
+
+                if ( topArmyBar.isSelected() )
+                    topArmyBar.ResetSelected();
+                if ( bottomArmyBar.isSelected() )
+                    bottomArmyBar.ResetSelected();
+                topArmyBar.Redraw(display);
+                bottomArmyBar.Redraw(display);
+                need_redraw = true;
+            }
+
             if ( conf.ExtCastleAllowGuardians() && !readOnly ) {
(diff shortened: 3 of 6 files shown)


src/fheroes2/army/army.cpp
```diff
@@ -207,6 +207,13 @@ Troops & Troops::operator=( const Troops & rhs )
     return *this;
 }
 
+Troops & Troops::operator=( Troops && troops ) noexcept
+{
+    static_cast<std::vector<Troop *> &>( *this ) = std::move( troops );
```
oleg-derevenetz, line 212: Hi @zenseii please note that `Army` in fact contains not pointers to `Troops` but pointers to `ArmyTroops`:

https://github.com/ihhub/fheroes2/blob/54cf329d3686351f0be3be88cbe15280b8d5443a/src/fheroes2/army/army.cpp#L800-L808

`ArmyTroop` is a descendant of `Troop`, but with extra field:

```cpp
const Army * army;
```

By simply swapping the contents, you will move units to the different `Army` instance, but these units will still reference to the old `Army` instance, which may lead to various issues, including (in theory) dangling pointers dereference and memory access violation. In either case, the invariant will be violated.

zenseii: Thanks @oleg-derevenetz! This clears up some doubts I had, and I believe it also explains why the troops aren't visually updated in the army bar because it takes `ArmyTroops` too.

I will try to come up with a solution to this.

zenseii:
After having used @oleg-derevenetz's implementation for moving troops between heroes I'm not sure we even need a way to swap all troops between two heroes. I've played several maps lately and have never missed it, even when playing on Android where such quick ways are more important.

My opinion is that it isn't worth it to meddle with the original heroes meeting window only to add an option that might almost never be used anyways.

However, in a castle there might still be a reason to have this. My main focus will be to add moving of all troops in a castle.

Branikolog:
Hi, @zenseii !
I think we need fast swap solution anyway, as current implementation helps in a case one of heroes has 1-2 troops. In a vase we have two full armies swap using pre-selction and pressing arrows doesn't work.
Also, pre-selecting troops and pressing arrow button is not really clear for players, in a case target hero has not enough free space to get all troops from other hero.
image
image

At last, it simply requires two steps, while single button would need only one. :)

oleg-derevenetz:
Hi @Branikolog how often you need to swap large armies between heroes in real life (I mean not the situation with hero chains when a big army should be transferred from one courier hero to the another courier, but the situation when both heroes have large armies)?

Branikolog:
Hello, @oleg-derevenetz .

> how often you need to swap large armies between heroes in real life

It's about the same to the case which current logic covers.
But army swap is much more obvious mechanic, while current implementation seems complex. Also, current implementation requires one extra action (I mean pre-selecting a troop before we press an arrow to swap), and it works predictable only for swapping with a single troop in hero army.
At last, implementing swap army function allows us to add a hotkey for this function. How can we handle army swap using keyboard with current implementation? I'm not telling, we should get rid of current logic. It's nice and helps a lot in some situations. But it doesn't cover all possible cases and it looks a bit complex.

zenseii:
> Hi @Branikolog how often you need to swap large armies between heroes in real life[...]

This is the same conclusion I arrived at. There are very few/rare real life occurrences in which it would save you time to be able to swap armies.

You would essentially need a main/champion hero that has 3 or more troops that are different from what a courier hero is bringing, which means that your champion has to have started moving around with 3 different types of troops than what the courier is bringing, and these troops that the champion has have to all be worse than what the courier is bringing. These are a lot of criterias for it to be useful. It also seems like this would need a very specific layout of a map to actually happen.

In the picture above I can imagine many players wanting to dismiss the centaurs and goblins anyways because they slow down this potential carrier hero, but sometimes not and your point is valid, @Branikolog  - I'm just pointing out that what you would do in the example is not always clear cut.

Another case in which army swap could be useful is when you have one champion army with troops that focus in adventure map speed, and another that focuses on castle sieging. I could imagine that one might want to just exchange these armies to swap the role of these heroes, i.e. strong hunter and castle sieger hero. Sure. But how often would you need to do this during a game and in how many maps would this be convenient?

Branikolog:
Hi, @zenseii !
All your points are valid.
Except chains I can also mention hero speeding, when we exchange army of champion with faster single troop of escort hero. And still, if we want to expand our list of hotkeys for army rearrangement between heroes, army swap is the function which is absolutely obvious and predictable for players and easy to implement without extra mouse clicks (which are required for current implementation).

zenseii:
> Hi, @zenseii ! All your points are valid. Except chains I can also mention hero speeding, when we exchange army of champion with faster single troop of escort hero.

If I'm not misunderstanding, this is exactly a case when you could use the current move all feature, even the preselecting. Yes it would cost you clicking one more time compared to army swapping, but I see we essentially disagree whether that is worth changing the appearance of the original dialog for 😅

> And still, if we want to expand our list of hotkeys for army rearrangement between heroes, army swap is the function which is absolutely obvious and predictable for players and easy to implement without extra mouse clicks (which are required for current implementation).

I see. This PR's implementation works as it should, but it might not be the best code for it. @ihhub was talking about a way to exchange the content of the heroes' armies rather than what is currently done where each creature has the new hero appointed to it. It seems more complicated than necessary.

There is already a hot key for it impmemented and this is the easiest part for this PR, but no visual button in the UI. The button should be easy enough to implement as we have already decided what solution to try first.

I'll make the necessary changes and then this PR can be used for testing and to see how many times one actually end up using it, before deciding if it should be implemented or not.

nogov:
Hi @zenseii
Have you thought about possibility to swap artifacts too? This is rather a common use case (at least, for H3 players with HD mod) to quickly swap armies AND artifacts.

ihhub:
Hi @zenseii , could you please tell the state of this pull request? Do you need any help with it or is it depended on some other changes?

PR #5934, opened 2022-10-07
Don't let the archers perform a double attack if they are out of ammo

### LINKED ISSUE #5931: Battle, double attacking shooters proceed one attack without arrows
### Preliminary checks

### Platform

Windows

### Describe the bug

In a case a shooter has odd number of arrows left he can come up to an attack, when he has only one arrow left.
image
In the OG he's allowed to proceed the second shot, even if he has only one arrow. I think it's a debatable logic. To my mind if shooter has only one arrow, he could make only one shot.
But currently in fheroes2 shooter is able to make a second hit with his knife. It looks like some sort of telekinetic swing. :D
image

### Save file

[Really distant swing.zip](https://github.com/ihhub/fheroes2/files/9733784/Really.distant.swing.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5931

### DIFF
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -1169,11 +1169,6 @@ int Battle::Unit::GetControl() const
     return !GetArmy() ? CONTROL_AI : GetArmy()->GetControl();
 }
 
-bool Battle::Unit::isArchers() const
-{
-    return ArmyTroop::isArchers() && shots;
-}
-
 void Battle::Unit::SpellModesAction( const Spell & spell, uint32_t duration, const HeroBase * hero )
 {
     if ( hero ) {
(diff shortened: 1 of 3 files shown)


src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1534,17 +1529,16 @@ void Battle::Unit::SpellRestoreAction( const Spell & spell, uint32_t spoint, con
...
-        return !isHandFighting();
+    if ( isHandFighting() ) {
+        return isAbilityPresent( fheroes2::MonsterAbilityType::DOUBLE_MELEE_ATTACK );
+    }
 
-    default:
-        break;
+    // Archers with double shooting ability can only fire a second shot if they still have ammo
+    if ( isAbilityPresent( fheroes2::MonsterAbilityType::DOUBLE_SHOOTING ) ) {
+        return GetShots() > 0;
```
ihhub, line 1538: Hi @oleg-derevenetz , should it be > 1 instead of > 0?

oleg-derevenetz: Hi @ihhub it should be > 0, because this function is called only after the first shot was already fired.

oleg-derevenetz: Hi @ihhub After some thought, I decided to rework the logic a bit to make it more obvious.

ihhub (approved):

ihhub:

PR #5830, opened 2022-09-12
Update Russian translation

dmitry-pshenitsin:

PR #5947, opened 2022-10-09
Update Russian translation

a1exsh:
Add proper translation for the "Defender" campaign scenario.
Follow-up: #5049
Cherry-picked some more translations from: #5830


zenseii:
Hi, @a1exsh.

I see that this relates to the previous PR where this context was added but would it be possible to at least add some more translations than just this? There is another Russian PR https://github.com/ihhub/fheroes2/pull/5830 that has been stale for a while so maybe you could grab some translations from there that you think are good and add them to this. Like 10 strings or something should be fine. We haven't set a minimum amount of strings for translations PRs.

a1exsh:
@zenseii added some of the translations from #5830 that didn't raise questions.

zenseii (approved):
@a1exsh, thank you for making this contribution and adding more than what I proposed.

I've looked over this with deepl for now so that at least the minimum amount of reviewing has been done for it. 

I've also asked for help from Russian speakers if they can have a look at it, but if no one does within reasonable time then this PR can be moved ahead.

zenseii:
Hi, @undef21 !

I noticed that you reviewed the other Russian PR that now has become stale, and I was wondering if you would have the time to have a look at this PR, if you want to of course.

files/lang/ru.po
```diff
@@ -1449,173 +1447,131 @@ msgstr "Сапоги Путника"
...
 
-#, fuzzy
 msgid "campaignBonus|Advanced Ballistics"
-msgstr "Баллиста"
+msgstr "Прод. баллистика"
 
-#, fuzzy
 msgid "campaignBonus|Expert Ballistics"
-msgstr "Баллиста"
+msgstr "Эксп. стрельба"
```
Branikolog, line 1466: "Эксп. баллистика" I suppose.

files/lang/ru.po
```diff
@@ -3554,21 +3500,17 @@ msgstr ""
...
 
-#, fuzzy
 msgid "QA Managers"
-msgstr "Менеджер по контролю качества"
+msgstr "Менеджеры по контролю качества"
 
-#, fuzzy
 msgid "Sound Design"
-msgstr "Композиторы и звукорежиссёры"
+msgstr "Звуковой дизайн"
```
undef21, line 3510: How about "Звуковое оформление"?

files/lang/ru.po
```diff
@@ -2711,24 +2667,20 @@ msgstr ""
...
+msgstr "Курсор мыши"
 
 msgid "Color"
-msgstr ""
+msgstr "Цветной"
 
-#, fuzzy
 msgid "Text Support"
-msgstr "Контроль качества и поддержка"
+msgstr "Текстовая поддержка"
```
undef21, line 2683: How about "Текстовое сопровождение"?

files/lang/ru.po
```diff
@@ -7957,7 +7901,7 @@ msgid ""
 msgstr ""
 
 msgid "Endless Sack of Gold"
-msgstr "Бездонный Мешок"
+msgstr "Бездонный Мешок Золота"
```
undef21, line 7904: In Russian only the first letter should be capital.

ihhub:
I believe we should close #5830 after merging this as the author of that pull request is gone.

files/lang/ru.po
a1exsh:
```suggestion
msgstr "Текстовое сопровождение"
```

files/lang/ru.po
a1exsh:
```suggestion
msgstr "Звуковое оформление"
```

files/lang/ru.po
a1exsh: Didn't we settle on first capitals when this point was last discussed about a year ago? @Branikolog

Branikolog: @a1exsh @undef21 , hello!
While searching through russian grammar I haven't discovered the exact rule regarding artifacts' names. The closest rule is regarding objects and works of art, where only the first letter is capital. Also such items as "Красная книга" are written with the first capital letter. So the majority of cases only the first letter should be capital one... Also, if there's any proper name it also should be written witch capital, like "Меч Геракла", but "Меч из камня" or "Лунный клинок".
At the same time, if we look through rules regarding medals, only the first letter after the word "medal" should be capital, so correct is "медаль/орден Отваги". In a case we have word "medal" standing on the first place of sentence (which is the case for the window header) we should make at least two capital letters: "Медаль Почетного летописца". 
I'm not sure, we should do so complex translation for each case individually, so I'm ok with the first capital for all artifacts, since all of them, medals or sacks are belong to the same item class (items blessing or cursing player).
Correct me, if I was wrong somewhere.
@a1exsh I'm sorry, it seems that I was wrong last year. It would be nice to have a Russian teacher in our team to consult on such questions. :)

a1exsh: @Branikolog did I mention last time that my mother is a philologist by her study?.. ;)

a1exsh: Seriously, though, whatever style we end up using, we should do that consistently.

Branikolog: > Seriously, though, whatever style we end up using, we should do that consistently.

Definitely! 
Both implementations we're talking about have pros and cons. Grammatically, I think, the first capital letter is more correct. "Нескончаемая Вязанка Дров"... I think it's too much privilege for just a cord of wood to be written with capital letter. :) But here we come up to a few exceptions. I'm talking about medals I've mentioned above. Also, the name "Anduran" for those rare artifacts would have to be written from capital, since it's the name.
On the other hand, while applying capital letters for all words in artifacts names the implementation looks maximum consistent. Also, this would help to distinguish artifacts' names within text item much easier.
So, @undef21 , what do you think? Can we ignore this controversial rules for the sake of consistency?

a1exsh: My point is that we can decide to change the style and then update all affected translations consistently (I myself was arguing for not using all capitals, but lost the argument last time).

What doesn't make sense is to start using a style different from the existing one, so I suggest we leave it like it is until a future decision.

files/lang/ru.po
```diff
@@ -1314,17 +1312,17 @@ msgid "The army will be carried over the scenario."
 msgstr ""
 
 msgid "The kingdom will have +%{count} %{resource} each day."
-msgstr ""
+msgstr "Королевство ежедневно будет получать +%{count} %{resource}."
 
 msgid "\"%{spell}\" spell will be carried over the scenario."
-msgstr ""
+msgstr "Заклинание \"%{spell}\" будет перенесено в этот сценарий."
```
Branikolog, line 1318: I think ` Заклинание \"%{spell}\" будет доступно в этом сценарии. `  would look better here.

files/lang/ru.po
```diff
@@ -1314,17 +1312,17 @@ msgid "The army will be carried over the scenario."
...
 
 msgid "%{hero} can be hired in the scenario."
-msgstr ""
+msgstr "%{hero} может быть нанят в этом сценарии."
 
 msgid ""
 "%{hero} has been defeated and will not appear in the subsequent scenarios."
-msgstr ""
+msgstr "%{hero} был повержен, и впоследстии Вы не сможете его нанять."
```
Branikolog, line 1325: Bonus relates to enemy hero, you've defeated. 
` %{hero} был повержен и больше не будет досаждать вам в последующих сценариях. `

Branikolog: Edit:
`%{hero} был повержен и больше не будет досаждать вам в этом последующих сценариях.`

ihhub:
@a1exsh , thank you very much for updating the translation!

PR #5957, opened 2022-10-10
Some corrections of typos and some wordchanges.

brandneetjan:


files/lang/nl.po
```diff
@@ -797,7 +797,7 @@ msgid "The Crown"
 msgstr "De Kroon"
 
 msgid "The Gauntlet"
-msgstr ""
+msgstr "De Spitsroede"
```
zenseii, line 800: I believe this translation refers to some kind of military running punishment? The way I understand it the use of "Gauntlet" here refers to the basic meaning of a running race.[ It seems to derive](https://www.dictionary.com/browse/gantlope) from the much more Germanic similar "Gantlope/Gatlop/Street race" and I believe you also have the word "lopen" in Dutch?

I'm emphasizing this because the scenario this title refers to is indeed a race against time where you have a limited time to complete it, but there is no concept of punishment as the "De spitsroede lopen" seems to suggest (at least wikipedia tells me this), but as I said I don't have any knowledge of Dutch so please ignore this if it is completely wrong.

Here is the title and the description of the map. It is much more of a race against time with no real enemy encounter.
image

zenseii (commented):
Hi, @brandneetjan.

Thank you for this contribution! I have looked over it with deepl and it looks good. As I said I don't know Dutch so please ignore my comments if my assumptions are wrong. I left only one comment and hopefully it can help you.

I would also like to kindly ask @piano1029's help in looking over these suggested changes if they should have the time to.

Don't hesitate to ask if you have any doubts!

zenseii:
Also, @brandneetjan, could you please change the title of this PR to something like "Update the Dutch translation" so that it is easy to tell what it is about without opening the PR? Thank you in advance!

files/lang/nl.po
zenseii: To add some more points I see how this map could be very closely resembling a narrow passageway alluding to such a military gauntlet. However reading further I see that (still from wikipedia) that in American English the spelling Gantlet is often preferred, and Heroes 2 leans more towards American English, especially considering that's where it was developed:
"For the punishment, the spelling gantlet is preferred in American English usage guides." [Under etymology and spelling](https://en.wikipedia.org/wiki/Running_the_gauntlet).

zenseii:
Should this PR be marked as draft, @ihhub ?

ihhub:
> Should this PR be marked as draft, @ihhub ?

Yes.

ihhub:
Hi @brandneetjan , please address given comments in this pull request so we can proceed with it.

PR #5961, opened 2022-10-11
Select Scenario Window Elements Position Fix (#1235)

### LINKED ISSUE #1235: Scenario selection window, elements position
In fheroes2:
- [ ] The position of the icons "Players Icon", "Size Icon", "Victory Condition Icon" and "Loss Condition Icon" is not the same as in the original game.
- [ ] The names of the scenarios in the scenarios list are not centered, the names of the scenarios are centered in the original game.
- [ ] The vertical spacing between the names of scenarios in the scenarios list is not regular, this spacing is regular in the original game.
- [ ] The text line of the scenario name "Broken Alliance" is not located at the same vertical position as in the original game.
- [ ] The scenario name "Broken Alliance" is not centered horizontally, the scenario name is not centered horizontally in the original game either, but it is less visible.
- [ ] The difficulty text line of the "Map difficulty: Normal" scenario is not located at the same vertical position as in the original game (the word "Normal" should be centered vertically in the frame and the text "Map difficulty:" should be aligned vertically to the word "Normal").
- [ ] The word used for the difficulty of the scenario "Normal" is not centered horizontally, the word used for the difficulty of the scenario is not centered horizontally either in the original game but it is less visible.
- [ ] The first line of the scenario description text is not located at the same vertical position as in the original game.

In the original game:

Scenario selection 01

In fheroes2:

Scenario selection 02

**Edit 2022/10/10.**

Positions with the number of pixels from the original game screen, with the corrections to be applied.

In original game:

Scenario selection window pixels 001

Since a supplementary icon for "The Succession Wars" and "The Price of Loyalty" versions has been added for the scenarios in fheroes2, the scenario name must be centered horizontally in the space between the icons on the left and the icons on the right in the list of scenarios.

In fheroes2:

Scenario selection window 002

HaveFun91:
Close #1235


HaveFun91:
Close #1235

zenseii:
Hi @HaveFun91.

I've linked the issue to this PR. The clang-tidy comments are valid so please have a look at them when you can and feel like it. Also it would be great if you could have a look at the Code style check that failed when you have the time.

Don't hesitate to ask if you have any questions.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -330,21 +329,24 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & all )
...
+    const fheroes2::Rect curCountPlayer( rt.x + 45, rt.y + 264, 18, 18 );
+    const fheroes2::Rect curMapSize( rt.x + 64, rt.y + 264, 18, 18 );
+    const fheroes2::Rect curMapType( rt.x + 83, rt.y + 264, 18, 18 );
+    const fheroes2::Rect curMapName( rt.x + 107, rt.y + 264, 160, 18 );
+    const fheroes2::Rect curVictoryCond( rt.x + 276, rt.y + 264, 18, 18 );
+    const fheroes2::Rect curLossCond( rt.x + 294, rt.y + 264, 18, 18 );
     const fheroes2::Rect curDifficulty( rt.x + 220, rt.y + 292, 114, 20 );
     const fheroes2::Rect curDescription( rt.x + 42, rt.y + 316, 292, 90 );
+    SelectedScenarioCoordinates selectedCoords( curCountPlayer.x, curMapSize.x, curMapType.x, curMapName.width, curMapName.x, curVictoryCond.x, curLossCond.x,
+                                                curDifficulty.x, curDescription.x, curDifficulty.width );
```
ihhub, line 349: Here we are storing 2 sets of the same coordinates and usage of structures is not exactly beneficial.

What if we create an enumeration in unnamed namespace which will contain all these magic numbers and then pass only x and y offset into ScenarioListBox? The biggest problem is not how we pass or process data but how we set this data and right now is this place. For example.
```cpp
const fheroes2::Rect curCountPlayer( rt.x + 45, rt.y + 264, 18, 18 );
```
could be (need to play with enumeration entry names):
```cpp
const fheroes2::Rect curCountPlayer( rt.x + CURRENT_PLAYER_OFFSET_X, rt.y + ICONS_OFFSET_Y, ICON_SIZE, ICON_SIZE );
```

What do you think?

src/fheroes2/dialog/dialog_selectscenario.h
```diff
@@ -65,6 +125,20 @@ class ScenarioListBox : public Interface::ListBox<Maps::FileInfo>
     }
 
     bool selectOk;
+
+private:
+    const ScenarioListItemCoordinates & _listItemCoords;
+    const SelectedScenarioCoordinates & _selectedCoords;
+    const short _offsetY = 4;
```
ihhub, line 132: Please do not use `short` variable type in the code as this is implementation dependent type.

src/fheroes2/dialog/dialog_selectscenario.h
```diff
@@ -65,6 +125,20 @@ class ScenarioListBox : public Interface::ListBox<Maps::FileInfo>
...
+
+    void _renderMapListItem( const Maps::FileInfo & info, fheroes2::Display & display, int32_t & dsty, bool current );
+    void _renderSelectedMapInfo( fheroes2::Display & display, const fheroes2::Point & dst );
+    void _renderMapName( const Maps::FileInfo & info, bool selected, const int32_t & baseYOffset, fheroes2::Display & display ) const;
+    const fheroes2::Sprite & _getPlayersCountIcon( uint8_t playersCount );
+    const fheroes2::Image & _getMapSizeIcon( uint16_t size );
+    const fheroes2::Sprite & _getMapTypeIcon( GameVersion version );
+    const fheroes2::Sprite & _getWinConditionsIcon( uint8_t condition );
+    const fheroes2::Sprite & _getLossConditionsIcon( uint8_t condition );
```
ihhub, line 141: All these methods must be static and their input arguments should be `const`.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -139,6 +139,8 @@ namespace
         return icon;
     }
 
+    fheroes2::Image undefinedMapSizeIcon;
```
ihhub, line 142: We should avoid using global variable unless it's the only way to implement things.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,137 +156,134 @@ namespace
...
+const fheroes2::Sprite & ScenarioListBox::_getLossConditionsIcon( uint8_t condition )
+{
+    int iconIndex = 36 + condition;
+    return fheroes2::AGG::GetICN( ICN::REQUESTS, iconIndex );
+}
 
-        index = 36 + info.conditions_loss;
-        const fheroes2::Sprite & spriteLoss = fheroes2::AGG::GetICN( ICN::REQUESTS, index );
-        fheroes2::Blit( spriteLoss, display, dst.x + 275 + spriteWins.width() + 2, dst.y + 265 );
+const fheroes2::Image & ScenarioListBox::_getMapSizeIcon( uint16_t size )
```
ihhub, line 252: Instead of returning an image and have a global variable to handle a corner case why don't we just create methods to draw stuff within them?

ihhub (changes requested):
Hi @HaveFun91, I put multiple comments in this pull request. Could you please check them?

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +154,130 @@
...
+    case Maps::XLARGE:
+        mapIconIndex = 29;
+        break;
+    default:
+        return GetNonStandardSizeIcon();
+    }
 
-        const fheroes2::Sprite & mapType = fheroes2::AGG::GetICN( ICN::MAP_TYPE_ICON, info._version == GameVersion::PRICE_OF_LOYALTY ? 1 : 0 );
-        fheroes2::Blit( mapType, display, dst.x + 46 + spriteCount.width() + 2 * 2 + spriteSize.width(), dst.y + 265 );
+    return fheroes2::AGG::GetICN( ICN::REQUESTS, mapIconIndex );
```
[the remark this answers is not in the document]
HaveFun91: I believe this is no longer a case.

src/fheroes2/dialog/dialog_selectscenario.cpp
HaveFun91: Yes, moving that values to enum makes sense. Commited fix on this, but couldn't get rid of magic calculations in total.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +178,141 @@ namespace
...
+    _renderMapName( info, current, dsty, display );
+    fheroes2::Blit( _getWinConditionsIcon( info.conditions_wins ), display, _x + SCENARIO_LIST_VICTORY_CONDITION_OFFSET_X, dsty );
+    fheroes2::Blit( _getLossConditionsIcon( info.conditions_loss ), display, _x + SCENARIO_LIST_LOSS_CONDITION_OFFSET_X, dsty );
+}
 
-    if ( info.size_w != info.size_h || info.size_w < Maps::SMALL || info.size_w > Maps::XLARGE ) {
-        const fheroes2::Image & nonStandardIcon = GetNonStandardSizeIcon();
+void ScenarioListBox::_renderSelectedScenarioInfo( fheroes2::Display & display, const fheroes2::Point & dst )
+{
+    Text text;
```
zenseii, line 219: Hi @HaveFun91 since we are touching this part of the code, `Text` is an old method that we are replacing with `fheroes2::Text` which is found in `ui_text.cpp`.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +178,141 @@ namespace
...
-        }
+    fheroes2::Blit( _getPlayersCountIcon( info.kingdom_colors ), display, dst.x + SELECTED_SCENARIO_COUNT_PLAYERS_OFFSET_X, dst.y + SELECTED_SCENARIO_GENERAL_OFFSET_Y );
+    _renderMapIcon( info.size_w, display, dst.x + SELECTED_SCENARIO_MAP_SIZE_OFFSET_X, dst.y + SELECTED_SCENARIO_GENERAL_OFFSET_Y );
+    fheroes2::Blit( _getMapTypeIcon( info._version ), display, dst.x + SELECTED_SCENARIO_MAP_TYPE_OFFSET_X, dst.y + SELECTED_SCENARIO_GENERAL_OFFSET_Y );
 
-        const fheroes2::Sprite & spriteSize = fheroes2::AGG::GetICN( ICN::REQUESTS, index );
-        fheroes2::Blit( spriteSize, display, dstx + offsetX, dsty );
-        offsetX += spriteSize.width() + 2;
-    }
+    text.Set( info.name, Font::BIG );
```
zenseii, line 226: We usually initialize and declare at once in the project.
Something like:
```cpp
fheroes2::Text text( info.name, fheroes2::FontType::normalWhite() );
```

Note that the old `Font::BIG` is called `normalWhite()` now.


Also note that `fheroes2::Text` has the methods `set()` and `draw()` instead of `Set` and `Blit`.
Another point is that the y-coordinate needs to be offset with the new draw compared to the old Blit. There should be a comment about this in the beginning of `ui_text.cpp`.

zenseii (changes requested):
Hi @HaveFun91.

I left a couple of small comments. Please have a look when you have the time to.

.gitignore
```diff
@@ -82,4 +83,4 @@ script/demo/demo
 VisualStudio/packages
 
 # Mac magic folders
-.DS_Store
+.DS_Store
```
zenseii, line 86: Could you please add a newline at the end of this file?

zenseii (changes requested):

oleg-derevenetz:
Hi all! Please note that some issues still present in this PR were not reported by our Clang-Tidy machinery due to recent regression in the upstream [clang-tidy-pr-comments](https://github.com/platisd/clang-tidy-pr-comments) workflow. The fix [is already on the way](https://github.com/platisd/clang-tidy-pr-comments/pull/31) and (I hope) will be merged shortly.

src/fheroes2/dialog/dialog_selectscenario.cpp
HaveFun91: I removed all `Text.h` dependecies.

src/fheroes2/dialog/dialog_selectscenario.cpp
HaveFun91: I have applied some changes. Not sure whether i got what you mean in terms of `draw()` method - was it about including offset within y-coordinate?

src/fheroes2/dialog/dialog_selectscenario.cpp
zenseii: Great! Thanks!

src/fheroes2/dialog/dialog_selectscenario.cpp
zenseii: > I have applied some changes. Not sure whether i got what you mean in terms of `draw()` method - was it about including offset within y-coordinate?

To be honest it's easier to figure out by looking at the game. IIRC when replacing Blit with draw you have to add an extra 2 in value to the y-coordinate.

Again it's easier to verify in the game so if it has looked right when you've checked it then it's nothing to worry about.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +183,143 @@ namespace
 
         return 0;
     }
+
+    int32_t GetCenteredTextXCoordinate( const int32_t startCoordX, const int32_t textBoxWidth, const int32_t textWidth )
```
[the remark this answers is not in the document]
HaveFun91: I don't get it - this warning makes no sense to me. For example why then
`virtual void draw( const int32_t x, const int32_t y, Image & output ) const = 0;` this is allowed - same considerations here.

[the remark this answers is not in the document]
HaveFun91: Is it about similar name of `textBoxWidth` and `textWidth`?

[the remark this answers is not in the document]
oleg-derevenetz: Hi @HaveFun91 it's a new warning from Clang-Tidy 14, it means that two (or more) parameters of the same (or convertible) types follow each other, so it is possible for a caller to swap/reorder them by mistake - for example, call the function `f(int x, int y)` as `f(y, x)`. You can ignore this warning for now, maybe we will turn it off later completely.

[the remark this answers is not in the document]
zenseii: Looking at the definition it seems it's suggesting you to change it into a struct. https://clang.llvm.org/extra/clang-tidy/checks/bugprone/easily-swappable-parameters.html

Not sure this is desirable or the right interpretation, @oleg-derevenetz ?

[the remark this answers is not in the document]
oleg-derevenetz: Hi @zenseii Yes, but it may be redundant to make a separate structure for each function that have several arguments of convertible types, so we need to think about the usefulness of this warning.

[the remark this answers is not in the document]
HaveFun91: There are some configuration possibilities here - maybe this setting: `MinimumLength` mentioned in doc would make sense to somekind of force user to create more complex object storing e.g. 5 parameters. Inventing brand new type for each of situation seems to be a mistake from my perspective.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +183,143 @@ namespace
 
         return 0;
     }
+
+    int32_t GetCenteredTextXCoordinate( const int32_t startCoordX, const int32_t areaWidth, const int32_t textWidth )
```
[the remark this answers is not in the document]
HaveFun91: As agreed - to be ignored for now.

src/fheroes2/dialog/dialog_selectscenario.cpp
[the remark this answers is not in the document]
oleg-derevenetz: @HaveFun91 you can ignore this warning for now (unless you really want to introduce structs in some cases).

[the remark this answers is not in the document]
HaveFun91: I will ignore this warning in that case.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -44,6 +43,45 @@
...
+        SELECTED_SCENARIO_DESCRIPTION_WIDTH = 292,
+        SELECTED_SCENARIO_DESCRIPTION_HEIGHT = 90,
+        SELECTED_SCENARIO_GENERAL_OFFSET_Y = 265,
+        // COMMON
+        ICON_SIZE = 18
+    };
+
+    void ShowToolTip( const std::string & header, const std::string & body )
+    {
+        fheroes2::showMessage( fheroes2::Text( header, fheroes2::FontType::normalYellow() ), fheroes2::Text( body, fheroes2::FontType::normalWhite() ), Dialog::ZERO );
```
ihhub, line 82: Please use `fheroes2::showStandardTextMessage()` here and in other places because some of them use **white** color instead of **yellow** for the header.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +183,143 @@ namespace
...
+        fheroes2::Fill( icon, 1, 1, 15, 15, fheroes2::GetColorId( 0x8D, 0x73, 0xFF ) );
+        fheroes2::Text text( _( "N" ), fheroes2::FontType::smallWhite() );
+        text.draw( ( 17 - text.width() ) / 2, ( 17 - text.height() ) / 2, icon );
 
-        text.Set( info.name, Font::BIG );
-        text.Blit( dst.x + 190 - text.w() / 2, dst.y + 265 );
+        fheroes2::Blit( icon, display, coordX, coordY );
+    }
+    else {
+        fheroes2::Blit( fheroes2::AGG::GetICN( ICN::REQUESTS, mapIconIndex ), display, coordX, coordY );
```
ihhub, line 293: I would suggest to add an assertion before this line to make that `mapIconIndex` is really >=0 because the second parameter of `GetICN()` function expects an unsigned int.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +183,143 @@ namespace
...
+const fheroes2::Sprite & ScenarioListBox::_getMapTypeIcon( const GameVersion version )
+{
+    return fheroes2::AGG::GetICN( ICN::MAP_TYPE_ICON, version == GameVersion::PRICE_OF_LOYALTY ? 1 : 0 );
+}
 
-        text.Set( _( "Map difficulty:" ), Font::BIG );
-        text.Blit( dst.x + 210 - text.w(), dst.y + 290 );
+const fheroes2::Sprite & ScenarioListBox::_getWinConditionsIcon( const uint8_t condition )
+{
+    int iconIndex = 30 + condition;
```
ihhub, line 310: This variable should be `uint32_t`.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -154,130 +183,143 @@ namespace
...
+    return fheroes2::AGG::GetICN( ICN::REQUESTS, iconIndex );
+}
 
-        text.Set( Difficulty::String( info.difficulty ) );
-        text.Blit( dst.x + 275 - text.w() / 2, dst.y + 290 );
+const fheroes2::Sprite & ScenarioListBox::_getLossConditionsIcon( const uint8_t condition )
+{
+    int iconIndex = 36 + condition;
```
ihhub, line 316: This variable should be `uint32_t`.

ihhub (changes requested):
Hi @HaveFun91, we are almost there. I put few comments related to code styling in general. Could you please take a look? Also please rebase your branch from master branch of this repo to use `fheroes2::showStandardTextMessage()` function.

src/fheroes2/dialog/dialog_selectscenario.cpp
HaveFun91: Assertion here doesn't make sense in my opinion - we have switch/case statement that sets specific values and on init it's -1. I mean - I can add it, but i don't see any benefit of putting assertion here.

ihhub (approved):

ihhub:
Hi @LeHerosInconnu, could you please test these changes?

zenseii (approved):
Hi @HaveFun91.

This looks good to me. I suggest to add a screenshot or some description showing the new changed dialog of this PR to quicker grasp the changes made from a visual perspective.

ihhub:
Before changes:
image
After changes:
image

@Branikolog and @LeHerosInconnu, I will need your opinion about this.

HaveFun91:
@Branikolog Tbh it compared two images above and i don't see any difference between S, M, XL, L buttons. Could you provide some more info about this? Or what should be done to correct placement?

HaveFun91:
I tried to move buttons... but seems like there are more underneath:
image

HaveFun91:
@Branikolog is it ok now?
image

Branikolog:
@HaveFun91 
OG menu
fheroes2 menu
There's some horizontal shifting left.

HaveFun91:
@Branikolog I have moved buttons to the left:
image

LeHerosInconnu:
Hello @HaveFun91 and @Branikolog,

> @Branikolog I have moved buttons to the left:
image

Everything looks good, except the spacing between the "L" and "XL" buttons is 8 pixels instead of 7 pixels.

Broken AllFlags in box 01

Also, in the example, the scenario name "Broken AllFlags" is not centered horizontally in the box (15 pixels on the left and 17 pixels on the right).

What do you think about moving the text line of the scenario file name up 1 pixel so that the lettering with jamb, in the example the "g" in the word "AllFlags", is not too close to the bottom of the text box (the text line would then have the same vertical position as the word "Normal" in the box for the scenario difficulty)?

HaveFun91:
@LeHerosInconnu  i uploaded fix.
See attached screenshot:
image

LeHerosInconnu:
Hello @HaveFun91,

> @LeHerosInconnu i uploaded fix. See attached screenshot:
image

Everything looks correct according to the screenshot provided.

ihhub:
@HaveFun91 , huge thanks for fixing this issue!

PR #5962, opened 2022-10-11
Update be.po

srzkptrsk:
Little bit changes

### DIFF
--- a/files/lang/be.po
+++ b/files/lang/be.po
@@ -34,19 +34,19 @@ msgid "DIFFICULTY"
 msgstr ""
 
 msgid "Warrior"
-msgstr ""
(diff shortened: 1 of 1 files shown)


zenseii:
Hi @srzkptrsk and thank you for this contribution!

I believe @Branikolog is most fit to review this, also because deepl doesn't have Belarusian support... It seems google translate does though so if worse comes to worst I'll check with that.

Branikolog (approved):
@srzkptrsk 
Brilliant! Cannot imagine this be translated better.
Thanks a lot for your contribution, @srzkptrsk !

ihhub:
@srzkptrsk , thank you very much for your contribution to the project!

PR #5963, opened 2022-10-11
Fix a few code smells detected by SonarCloud

oleg-derevenetz:


ihhub:
@oleg-derevenetz , great thanks for cleaning up the code!

PR #5964, opened 2022-10-11
Update workflows with latest versions of the actions which runs on Node 16

oleg-derevenetz:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

~The latest version of `microsoft/setup-msbuild` action still uses Node 12, hopefully Micro$oft will fix it in the foreseeable future.~ Already fixed.


ihhub:
@oleg-derevenetz , a bunch of thanks for update CI/CD pipeline!

PR #5985, opened 2022-10-16
Add Android specific code

ihhub:
This is minimum required code (besides commenting the whole audio-related code as I haven't figured out how to compile SDL Mixer) to run fheroes2 on an Android device.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -187,6 +187,19 @@ std::string System::GetDataDirectory( const std::string & prog )
     }
 
     return {};
+#elif defined( ANDROID )
+    const char * internalDir = SDL_AndroidGetInternalStoragePath();
+    if ( internalDir ) {
+        return System::ConcatePath( internalDir, prog );
+    }
+
+    if ( SDL_ANDROID_EXTERNAL_STORAGE_READ && SDL_AndroidGetExternalStorageState() ) {
+        const char * externalDir = SDL_AndroidGetExternalStoragePath();
+        if ( externalDir ) {
+            return System::ConcatePath( externalDir, prog );
+        }
+    }
+    return {};
 #else
     return GetHomeDirectory( prog );
 #endif
@@ -267,7 +280,7 @@ bool System::IsFile( const std::string & name, bool writable )
     }
 
     return writable ? ( 0 == _access( name.c_str(), 06 ) ) : ( 0 == _access( name.c_str(), 04 ) );
-#elif defined( TARGET_PS_VITA )
+#elif defined( TARGET_PS_VITA ) || defined( ANDROID )
     // TODO: check if it is really a file.
     return writable ? 0 == access( name.c_str(), W_OK ) : 0 == access( name.c_str(), R_OK );
 #else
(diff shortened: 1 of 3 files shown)


src/engine/system.cpp
```diff
@@ -186,6 +186,19 @@ std::string System::GetDataDirectory( const std::string & prog )
...
     }
 
+    return {};
+#elif defined( ANDROID )
+    const char * internalDir = SDL_AndroidGetInternalStoragePath();
+    if ( internalDir ) {
+        return System::ConcatePath( internalDir, prog );
+    }
+
+    if ( SDL_ANDROID_EXTERNAL_STORAGE_READ && SDL_AndroidGetExternalStorageState() ) {
```
oleg-derevenetz, line 196: AFAIK on recent Android versions you will need to explicitly ask the user for permission to read files from external storage. Simply specifying the appropriate permission in the app's manifest is no longer enough. Also IIRC on even more recent Android versions there is an even bigger mess with that - you can't just open an arbitrary file on the external storage anymore even if you declared and acknowledged the appropriate permission: you can open only those files that were created by this application itself. If the file in question wasn't created by this app, you can only use either special APIs (like MediaStore API) or the system file picker to open it.

src/fheroes2/h2d/h2d.cpp
```diff
@@ -33,7 +34,14 @@ namespace
     {
 #if defined( MACOS_APP_BUNDLE )
         return Settings::findFile( "h2d", fileName, path );
-        const std::string internalDirectory( "h2d" );
+#elif defined( ANDROID )
```
oleg-derevenetz, line 37: Why do we need special logic for Android in this case? Is the `Settings::findFile()` function not doing its job?

oleg-derevenetz (commented):
Hi @ihhub I just left one comment and one question regarding these changes, could you please take a look?

src/engine/system.cpp
oleg-derevenetz: P.S. There should be `&`, not `&&`, because `SDL_AndroidGetExternalStorageState()` returns the bitmask.

ihhub: This is a copy-paste mistake from the very old code. I will fix it.

Are you suggesting to remove external storage check or keep it as it is for the current state of code?

oleg-derevenetz (approved):

src/engine/system.cpp
oleg-derevenetz: Hi @ihhub

> Are you suggesting to remove external storage check or keep it as it is for the current state of code?

Let's leave it as it is, but I foresee problems in the future. Most likely with Android 10+ target API you won't be able to simply copy the OG data files to the SD card and open them by the fheroes2 app "as usual" because of these absolutely insane restrictions with ugly "best practices" from Google. There is the [MANAGE_EXTERNAL_STORAGE](https://support.google.com/googleplay/android-developer/answer/10467955?hl=en) permission that makes the situation somewhat easier, but this permission is primarily intended for apps like file managers, and the use of it by the game may be questioned if you want to publish the fheroes2 on Google Play.

ihhub: As of now the goal is to at least make it work in one way or another. We will polish the code over time.

PR #5988, opened 2022-10-16
Fixed issue where soundfonts are not found in the files directory

### LINKED ISSUE #5986: Music is not working on ubuntu
### Preliminary checks

### Platform

Linux

### Describe the bug

The issue is created based on a feedback from a user.

No music during application run. Please note that sound effects are OK.
This has started since: 0.9.18

OS: Ubuntu 20.04.3 LTS
fheroes2 engine, version: 0.9.18 - 0.9.20

$ ./fheroes2
fheroes2 engine, version: 0.9.20
15.10.2022 09:41:06: [ERROR] createMusic:  Failed to create a music track from memory. The error: No SoundFonts have been requested
15.10.2022 09:41:18: [ERROR] createMusic:  Failed to create a music track from memory. The error: No SoundFonts have been requested
15.10.2022 09:41:19: [ERROR] createMusic:  Failed to create a music track from memory. The error: No SoundFonts have been requested

The issue is actual for both MIDI and External music setting.

Version 0.9.17 is the last one where I dont get this problem.
No music at all but in game SFX work OK.

### Save file

None.

### Additional info

_No response_

felix642:
closes #5986

If for some reason someone deleted the files/soundfont directory (or does not have it). The music will fail to play because the SDL_Mixer won't be able to find any sound fonts. 

The SDL_Mixer libs seem to be using default paths that we don't want to overwrite when no sound fonts have been found. 

This PR adds a condition before calling the Mix_SetSoundFonts function to make sure that we don't overwrite the paths with an empty string.

### DIFF
--- a/src/fheroes2/audio/audio_manager.cpp
+++ b/src/fheroes2/audio/audio_manager.cpp
@@ -772,7 +772,11 @@ namespace AudioManager
             // Set the volume for all channels to 0. This is required to avoid random volume spikes at the beginning of the game.
             Mixer::setVolume( -1, 0 );
 
-            Music::SetMidiSoundFonts( midiSoundFonts );
+            // Some platforms (e.g. Linux) may have their own predefined soundfonts, don't overwrite them if we don't have our own
+            if ( !midiSoundFonts.empty() ) {
+                Music::SetMidiSoundFonts( midiSoundFonts );
+            }
+
             Music::setVolume( 100 * Settings::Get().MusicVolume() / 10 );
             Music::SetFadeInMs( 900 );
         }


felix642:
@oleg-derevenetz I would appreciate if you could review this PR.

oleg-derevenetz:
Hi @felix642 the current behavior was made for a reason - to allow the use of system-specific fallback mechanisms if needed. For example, currently there is an opportunity to turn off the usage of soundfonts on Windows and fallback to the native Windows MIDI by passing the empty list to the `Music::SetMidiSoundFonts()`. With this PR, there will be no such opportunity anymore - you will not be able to turn the Fluindsynth backend off and switch to the native Windows MIDI backend once you called the `Music::SetMidiSoundFonts()` at least once with some set of soundfonts. SDL_mixer implementation has its own [fallback mechanisms](https://github.com/ihhub/fheroes2/issues/5986#issuecomment-1279944333) that should be used when the user-provided list of soundfonts is empty.

src/engine/audio.cpp
```diff
@@ -1097,7 +1097,12 @@ void Music::SetMidiSoundFonts( const ListFiles & files )
...
 
-    if ( Mix_SetSoundFonts( System::FileNameToUTF8( filePaths ).c_str() ) == 0 ) {
-        ERROR_LOG( "Failed to set MIDI SoundFonts using paths " << filePaths << ". The error: " << Mix_GetError() )
+    if ( !filePaths.empty() ) {
+        if ( Mix_SetSoundFonts( System::FileNameToUTF8( filePaths ).c_str() ) == 0 ) {
+            ERROR_LOG( "Failed to set MIDI SoundFonts using paths " << filePaths << ". The error: " << Mix_GetError() )
+        }
+    }
+    else {
+        ERROR_LOG( "No MIDI SoundFonts found in the files directory." );
```
oleg-derevenetz, line 1106: Hi @felix642 Let's not make these changes here, because the engine is relatively low-level layer, it should not be aware about the game directory layout (especially about the `files` directory). I suggest to move this check to the `AudioManager::AudioInitializer::AudioInitializer()` (just skip the `Music::SetMidiSoundFonts()` call if the `midiSoundFonts` parameter is empty) and do not print the error in this case, because for most systems (I'd say - for all except Windows) custom soundfonts are not used at all and are not shipped with the game distribution, so this error message will be completely bogus for all these systems.

felix642:
> Hi @felix642 the current behavior was made for a reason - to allow the use of system-specific fallback mechanisms if needed. For example, currently there is an opportunity to turn off the usage of soundfonts on Windows and fallback to the native Windows MIDI "on the fly" by passing the empty list to the `Music::SetMidiSoundFonts()`. With this PR, there will be no such opportunity anymore - you will not be able to turn the Fluindsynth backend off and switch to the native Windows MIDI backend once you called the `Music::SetMidiSoundFonts()` at least once with some set of soundfonts. SDL_mixer implementation has its own [fallback mechanisms](https://github.com/ihhub/fheroes2/issues/5986#issuecomment-1279944333) that should be used when the user-provided list of soundfonts is empty.
> 
> P.S. However we can use the `SDL_NATIVE_MUSIC` SDL hint instead if necessary.

@oleg-derevenetz what do you think we should do to fix that issue then? 
I see that SDL tries to read an environment variable if the soundfont_paths is empty..

Maybe we could export in the env variable for the current sound fonts' paths before overwriting them so that it can at least fallback on those ? 

```
void Music::SetMidiSoundFonts( const ListFiles & files )

...

SDL_setenv("SDL_SOUNDFONTS", Mix_GetSoundFonts(), false);

if ( Mix_SetSoundFonts( System::FileNameToUTF8( filePaths ).c_str() ) == 0 ) {
    ERROR_LOG( "Failed to set MIDI SoundFonts using paths " << filePaths << ". The error: " << Mix_GetError() )
}
```

oleg-derevenetz:
@felix642

> what do you think we should do to fix that issue then?

For now, I suggest to leave the `Music::SetMidiSoundFonts()` as is, move this check to the more high-level [`AudioManager::AudioInitializer::AudioInitializer()`](https://github.com/oleg-derevenetz/fheroes2/blob/5f7d19f785d765d046dd0252f79caeeaf0b9f438/src/fheroes2/audio/audio_manager.cpp#L769) (just don't call the Music::SetMidiSoundFonts() if the midiSoundFonts parameter is empty) and not print the error message if there are no soundfonts, because they shipped only for Windows, all other platforms (Linux, OSX, Nintendo Switch, PS Vita, Android in the future) are using their own MIDI backends. Windows is the only platform with ugly native MIDI backend with numerous issues, that's why we are using Fluidsynth backend with custom soundfonts on it. All other native backends are fine.

felix642:
> @felix642
> 
> > what do you think we should do to fix that issue then?
> 
> For now, I suggest to leave the `Music::SetMidiSoundFonts()` as is, move this check to the more high-level [`AudioManager::AudioInitializer::AudioInitializer()`](https://github.com/oleg-derevenetz/fheroes2/blob/5f7d19f785d765d046dd0252f79caeeaf0b9f438/src/fheroes2/audio/audio_manager.cpp#L769) (just don't call the `Music::SetMidiSoundFonts()` if the `midiSoundFonts` parameter is empty) and not print the error message if there are no soundfonts, because they shipped only for Windows, all other platforms (Linux, OSX, Nintendo Switch, PS Vita) are using their own MIDI backends which doesn't require any custom soundfonts. Windows is the only platform with ugly native MIDI backend with numerous issues, that's why we are using Fluidsynth backend with custom soundfonts on it. Native MIDI backends on all other supported platforms are fine, there is no need to scare users by error that makes no sense (and even on Windows there will be no issue if there are no soundfonts - the app will use that ugly native MIDI backend, that's all).

@oleg-derevenetz, I have updated my PR and implemented the proposed changes.

Do you think we should add an assert to the setMidiSoundFonts function to make sure that the fileList is not empty?

oleg-derevenetz:
Hi @felix642

> Do you think we should add an assert to the setMidiSoundFonts function to make sure that the fileList is not empty?

Let's leave it as is. The thing is, this function is low-level and therefore it should just obey the orders, and we still can (in theory) implement the setting in the game UI to switch between FluidSynth and the system MIDI synthesizer, so empty list of soundfonts may be legitimate.

oleg-derevenetz (approved):
Hi @felix642 I just paraphrased the comment a bit to better describe why exactly we are doing this. LGTM.

felix642:
> Hi @felix642 I just paraphrased the comment a bit to better describe why exactly we are doing this. LGTM.

Thank you :)

I have also rebased my branch since most of the commits were not relevant to each other.

ihhub:
@felix642 and @oleg-derevenetz , a bunch if thanks for fixing this issue!

PR #5990, opened 2022-10-16
Fixed issue with kingdom overview scrollbar not remembering position

### LINKED ISSUE #1497: Hero, combat, necromancy skill, practicing the dark arts of necromancy, wrong window displayed
In fheroes2, the wrong window is displayed for practicing the dark arts of necromancy.

In original game:

Minimum 1 skeleton 01

Minimum 1 skeleton 02

In fheroes2:

Wrong window combat skeleton raised 01

Wrong window combat skeleton raised 02

### LINKED ISSUE #4197: Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, position of lists in Heroes Overview and Towns/Castles Overview screens are not completely memorized
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, position of lists in Heroes Overview and Towns/Castles Overview screens are not completely memorized.

In fheroes2:

https://user-images.githubusercontent.com/43583994/132983911-8f80a5bb-a60d-4da9-ab1f-e10b5b294956.mp4

### Save file

Here is the save file (0.9.7 4180):
[Kingdom overview positions 01.zip](https://github.com/ihhub/fheroes2/files/7149411/Kingdom.overview.positions.01.zip)

### Additional info

Related Pull Request merged: Remember UI position in Kingdom Overview dialog #4035

felix642:
closes #4197

To preserve compatibility with the old save files. I have decided to store both scroll bars position in the same variable (_topItemInKingdomView). 16 first bits are used for the castle index and 16 last bits are used for the hero index.

This means that we can have a maximum of 32 000 heroes before something start to break which looks safe to me.


oleg-derevenetz:
Hi @felix642 there is apparently wrong issue linked to this PR.

felix642:
> Hi @felix642 there is apparently a wrong issue linked to this PR.

You are right! I have update the description.

oleg-derevenetz:
> To preserve compatibility with the old save files. I have decided to store both scroll bars position in the same variable (_topItemInKingdomView). 16 first bits are used for the castle index and 16 last bits are used for the hero index.

Although this approach is undoubtedly a working one, it looks unnecessary quirky IMHO. Initially, a similar approach was used when storing adventure map data, but we are now gradually moving away from it by using separate variables for each kind of data instead of packing many different fields into the one variable and using bitmasks to load and store these fields. So I suggest to consider another approach: use different variables and make the backwards-compatible changes to the savefile structure in a usual way (bump the savefile version and use temporary code to handle the previous savefile versions).

felix642:
HI @oleg-derevenetz, I have updated the implementation and both values are stored in different variables now.

I'm not super familiar with the versioning system of saving files, so I would appreciate it if you could double-check to make sure that I correctly implemented everything. I ran some tests and everything seems to be working smoothly but I just want to make sure.

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -966,14 +967,27 @@ cost_t Kingdom::_getKingdomStartingResources( const int difficulty ) const
 
 StreamBase & operator<<( StreamBase & msg, const Kingdom & kingdom )
 {
-    return msg << kingdom.modes << kingdom.color << kingdom.resource << kingdom.lost_town_days << kingdom.castles << kingdom.heroes << kingdom.recruits
-               << kingdom.visit_object << kingdom.puzzle_maps << kingdom.visited_tents_colors << kingdom._lastBattleWinHeroID << kingdom._topItemInKingdomView;
+    msg << kingdom.modes << kingdom.color << kingdom.resource << kingdom.lost_town_days << kingdom.castles << kingdom.heroes << kingdom.recruits << kingdom.visit_object
+        << kingdom.puzzle_maps << kingdom.visited_tents_colors << kingdom._lastBattleWinHeroID << kingdom._topCastleInKingdomView;
+
+    if ( Game::GetLoadVersion() >= FORMAT_VERSION_0921_RELEASE ) {
```
oleg-derevenetz, line 973: This check is not needed here, just write the new field to the stream as usual, because we always create the save file of the latest possible version.

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -966,14 +967,27 @@ cost_t Kingdom::_getKingdomStartingResources( const int difficulty ) const
...
+    return msg;
 }
 
 StreamBase & operator>>( StreamBase & msg, Kingdom & kingdom )
 {
     msg >> kingdom.modes >> kingdom.color >> kingdom.resource >> kingdom.lost_town_days >> kingdom.castles >> kingdom.heroes >> kingdom.recruits >> kingdom.visit_object
-        >> kingdom.puzzle_maps >> kingdom.visited_tents_colors >> kingdom._lastBattleWinHeroID >> kingdom._topItemInKingdomView;
+        >> kingdom.puzzle_maps >> kingdom.visited_tents_colors >> kingdom._lastBattleWinHeroID >> kingdom._topCastleInKingdomView;
+
+    if ( Game::GetLoadVersion() >= FORMAT_VERSION_0921_RELEASE ) {
```
oleg-derevenetz, line 985: It would be good to add the `static_assert()` before this line:

```cpp
static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_0921_RELEASE, "Remove the check below." );
```

This will automatically break the build when the `LAST_SUPPORTED_FORMAT_VERSION` become equal to `FORMAT_VERSION_0921_RELEASE` and we will not forget to remove this check, that will become obsolete.

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -966,14 +967,27 @@ cost_t Kingdom::_getKingdomStartingResources( const int difficulty ) const
...
+        msg >> kingdom._topHeroInKingdomView;
+    }
+    else {
+        kingdom._topHeroInKingdomView = kingdom._topCastleInKingdomView;
```
oleg-derevenetz, line 989: Are you sure they should be the same if there is no `_topHeroInKingdomView` in the savefile? Maybe set it to -1 by default?

src/fheroes2/system/save_format_version.h
```diff
@@ -29,6 +29,7 @@ enum SaveFileFormat : uint16_t
     // If you're removing an old version you must assign the oldest available to LAST_SUPPORTED_FORMAT_VERSION located at the bottom.
 
     // 10000 value must be used for 1.0 release so all version before it should have lower than this value.
+    FORMAT_VERSION_0921_RELEASE = 9950,
```
oleg-derevenetz, line 32: Please set the `CURRENT_FORMAT_VERSION` below to the `FORMAT_VERSION_0921_RELEASE`. The idea is that we always create save files of the latest possible version, but we have special fallbacks if we are trying to load an older save file.

oleg-derevenetz (changes requested):
Hi @felix642 I just left a few comments, please take a look when you have time.

src/fheroes2/kingdom/kingdom.cpp
felix642: You are right! Thank you.

src/fheroes2/kingdom/kingdom.cpp
felix642: Done, thank you.

src/fheroes2/kingdom/kingdom.cpp
felix642: The default constructor already sets the value to -1. So I have simply removed this branch.

src/fheroes2/kingdom/kingdom_overview.cpp
```diff
@@ -760,20 +759,23 @@ void Kingdom::openOverviewDialog()
...
-        if ( le.MouseClickLeft( buttonExit.area() ) || Game::HotKeyCloseWindow() )
+        if ( le.MouseClickLeft( buttonExit.area() ) || Game::HotKeyCloseWindow() ) {
             break;
+        }
 
         // switch view: heroes/castle
         if ( buttonHeroes.isReleased() && le.MouseClickLeft( buttonHeroes.area() ) ) {
             buttonHeroes.drawOnPress();
             buttonCastle.drawOnRelease();
+            _topHeroInKingdomView = listStats->getTopId();
```
oleg-derevenetz, line 770: Why we are remembering the top hero before switching to the hero screen (while in fact we are currently on the castle screen)?

src/fheroes2/kingdom/kingdom_overview.cpp
```diff
@@ -760,20 +759,23 @@ void Kingdom::openOverviewDialog()
...
             listStats = &listHeroes;
             ResetModes( KINGDOM_OVERVIEW_CASTLE_SELECTION );
             redraw = true;
         }
         else if ( buttonCastle.isReleased() && le.MouseClickLeft( buttonCastle.area() ) ) {
             buttonCastle.drawOnPress();
             buttonHeroes.drawOnRelease();
+            _topCastleInKingdomView = listStats->getTopId();
```
oleg-derevenetz, line 778: Why we are remembering the top castle before switching to the castle screen (while in fact we are currently on the hero screen)? Shouldn't it be vice-versa?

oleg-derevenetz (commented):
Hi @felix642 I just left two questions (in fact one :), please take a look when possible.

src/fheroes2/kingdom/kingdom_overview.cpp
felix642: See previous comment.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
Hi @Branikolog , can you please verify this pull request?

Branikolog (approved):
Everything works fine!

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -967,13 +968,19 @@ cost_t Kingdom::_getKingdomStartingResources( const int difficulty ) const
...
 StreamBase & operator>>( StreamBase & msg, Kingdom & kingdom )
 {
     msg >> kingdom.modes >> kingdom.color >> kingdom.resource >> kingdom.lost_town_days >> kingdom.castles >> kingdom.heroes >> kingdom.recruits >> kingdom.visit_object
-        >> kingdom.puzzle_maps >> kingdom.visited_tents_colors >> kingdom._lastBattleWinHeroID >> kingdom._topItemInKingdomView;
+        >> kingdom.puzzle_maps >> kingdom.visited_tents_colors >> kingdom._lastBattleWinHeroID >> kingdom._topCastleInKingdomView;
+
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_0921_RELEASE, "Remove the check below." );
+    if ( Game::GetLoadVersion() >= FORMAT_VERSION_0921_RELEASE ) {
+        msg >> kingdom._topHeroInKingdomView;
+    }
```
ihhub, line 983: Hi @felix642 , I forgot to check this but for else condition we need to reset the newly added variable to `-1`. Could you please add this?

ihhub:
@felix642 , a bunch of thanks for fixing this problem!

PR #5993, opened 2022-10-19
Use _n() instead of Translation::ngettext()

oleg-derevenetz:
I have no idea why I used `Translation::ngettext()` here while there is a much shorter widely used alternative.


ihhub:

PR #5998, opened 2022-10-19
Fix terrain colors on the minimap

### LINKED ISSUE #5994: Minimap color differences
### Preliminary checks

### Platform

Windows

### Describe the bug

In the original game some of colors for minimap are different from current. Right now, within #5980 colors of obstacles on grass are fixed, but there some other differences left:
Mmap mini
The grass terrain color is different.
Wasteland terrain and obstacles have wrong colors.

### Save file

[Terrains.zip](https://github.com/ihhub/fheroes2/files/9819744/Terrains.zip)

### Additional info

_No response_

a1exsh:
Close #5994

### DIFF
--- a/src/fheroes2/gui/interface_radar.cpp
+++ b/src/fheroes2/gui/interface_radar.cpp
@@ -188,7 +188,7 @@ void Interface::Radar::Generate()
 
                 const MP2::MapObjectType objectType = tile.GetObject();
                 if ( objectType == MP2::OBJ_MOUNTS || objectType == MP2::OBJ_TREES )
-                    color += 2;
+                    color += 3;
             }
 
             fheroes2::SetPixel( spriteArea, x, y, color );
--- a/src/fheroes2/gui/interface_radar.cpp
+++ b/src/fheroes2/gui/interface_radar.cpp
@@ -56,11 +56,11 @@ namespace
         COLOR_DESERT = 0x76,
         COLOR_SNOW = 0x0D,
         COLOR_SWAMP = 0x68,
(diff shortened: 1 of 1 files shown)


a1exsh:
There seems to be a problem now with an orange hero staying on the wastelands (same color is used):
Screenshot from 2022-10-20 06-29-06

And, as I've mentioned already, both swamp and lava makes it really hard to see a green hero on them.  I think we should fix this by making the green player's color slightly brighter on the minimap.

a1exsh:
> same color is used

OK, it's _almost_ the same — on the above screenshot I forgot to compile from this branch, but it's very hard to see any difference anyway, as the colors are just one index apart.

Branikolog:
Hi, @a1exsh 
In the OG, actually, other players are not visible on minimap. :)
Btw, if we talking about players colours, fheroes2 has also differences:
PL colours
Currently, red, blue, yellow colours are different.

a1exsh:
> Currently, red, blue, yellow colours are different.

Do we want a separate issue for discussing player colors, or should we extend the scope of this one? :)

Since they are different from the OG anyway and there are low contrast problems with some combinations of player/terrains, we might want to adjust the player colors at some point as well.

Branikolog:
Hi, @a1exsh 
I see no problem making an issue for players colours as well. 
I think, within this PR you should focus on reproducing terrain colours first.

Branikolog (approved):
All terrain colors are the same to the OG. Good work.

ihhub:
@a1exsh , thanks a lot for this fix!

PR #5999, opened 2022-10-19
Update extract_homm2_resources.ps1

GeorgeK1ng:
Added support for extracting resources from orignal game released in Gold Edition by CDPROJEKT in Czech Republic in 2002. All needed files are automatically grabbed from installed game path + drive path too. In my original release is needed CD to play as it also have some of game files.

With this modification is not needed manual input and everything is done automatically. In my tests was mounted game ISO, then installed game, ISO keep mounted and then launch grabbing script. No issues detected. Test was done under Windows XP SP3

### DIFF
--- a/script/homm2/extract_homm2_resources.ps1
+++ b/script/homm2/extract_homm2_resources.ps1
@@ -63,9 +63,11 @@ try {
     Write-Host "[2/3] determining the HoMM2 directory"
 
     $homm2Path = $null
-
+    $homm2DrivePath = $null
+	
     foreach ($key in @("HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall",
                        "HKEY_CURRENT_USER\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall",
+                       "HKEY_LOCAL_MACHINE\SOFTWARE\New World Computing\Heroes of Might and Magic 2",
                        "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall",
                        "HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall")) {
         if (-Not (Test-Path -Path "Microsoft.PowerShell.Core\Registry::$key" -PathType Container)) {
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @GeorgeK1ng please put the indentation in order and keep changes minimal.

GeorgeK1ng:
I have fixed spacing and minimized changes in last code modification

script/homm2/extract_homm2_resources.ps1
```diff
@@ -63,23 +63,33 @@ try {
...
+	    # Legacy installed path detection
+            if ($null -Eq $path) {
+            	$path = $subkey.GetValue("AppPath")
+            }
+	    
             if ($null -Ne $path) {
                 $path = $path.TrimEnd("\")
-
+            	# Legacy installed drive path detection
+            	$DrivePath = $subkey.GetValue("CDDrive")
```
oleg-derevenetz, line 88: I have an impression that this temporary variable is not needed, because it is not used anywhere except the assignment below, so we just can assign the `$subkey.GetValue("CDDrive")` to the `$homm2DrivePath` below directly.

script/homm2/extract_homm2_resources.ps1
```diff
@@ -108,21 +118,25 @@ try {
     } else {
         $shell = New-Object -ComObject "Shell.Application"
 
-        foreach ($srcDir in @("HEROES2\ANIM", "ANIM", "DATA", "MAPS", "MUSIC")) {
-            if (-Not (Test-Path -Path "$homm2Path\$srcDir" -PathType Container)) {
-                continue
-            }
+        foreach ($homm2Dir in @($homm2Path, $homm2DrivePath )) {
+
```
oleg-derevenetz, line 122: Now `$homm2Dir` may become `$null` (because `$homm2DrivePath` may become `$null`, moreover, this is its "usual value" in most cases), we need to check this before entering the inner loop:

```
if ($null -Eq $homm2Dir) {
    continue
}
```

oleg-derevenetz (changes requested):
Hi @GeorgeK1ng I just left a few comments, please take a look when you have time.

It is also interesting what will happen with this script if there will be no disk in the CD drive when it is launched by the user?

GeorgeK1ng (commented):
As this is my first attempt in PowerShell script, I'm happy you improved my mod

GeorgeK1ng (commented):
OK

script/homm2/extract_homm2_resources.ps1
GeorgeK1ng: Thanks, I haven't expected this behaviour with $null

oleg-derevenetz:
Hi @GeorgeK1ng I have an impression that you forgot to push the requested changes.

GeorgeK1ng:
@oleg-derevenetz I commited it now

oleg-derevenetz:
Hi @GeorgeK1ng I just made a few minor modifications, could you please verify whether this script still works for your case (for a legacy installation with CD)?

GeorgeK1ng:
Yes it works

image

oleg-derevenetz (approved):

ihhub:
@GeorgeK1ng , thank you very much for your contribution to the project!

oleg-derevenetz:
> [offtopic] Windows XP with fheroes2. Wow!

It won't be for long 🙋‍♂️

PR #6005, opened 2022-10-20
Add basic CI for Android

oleg-derevenetz:
Since we are now supporting Android (sort of), we need to add CI for it. Just build the debug APK and run lint.


ihhub:
Hi @oleg-derevenetz , could you please explain the purpose of having the artifact from Android instead of let's say debug build?

oleg-derevenetz:
Hi @ihhub because now the debug build is unusable for a wide audience anyway :) Android app can be run only through a debugger because of issues with data files - a person will not be able even to upload them to the device without specific skills :) But the lint is a different thing, if lint fails because of some error in the Android code, we should have an ability to see why.

I'll add APK upload as soon as it will become more or less usable.

ihhub:
> Hi @ihhub because now the debug build is unusable for a wide audience anyway :) Android app can be run only through a debugger because of issues with data files - a person will not be able even to upload them to the device without specific skills :) But the lint is a different thing, if lint fails because of some error in the Android code, we should have an ability to see why.
> 
> I'll add APK upload as soon as it will become more or less usable.

Sure, It makes sense. I am creating an issue containing all possible tasks we need to do to make a proper Android release.

ihhub:

PR #6013, opened 2022-10-20
PS Vita: don't apply temporary patch to SDL2

oleg-derevenetz:
PSV SDL2 package was apparently updated to the latest version, we don't need this patch anymore.


ihhub:

PR #6014, opened 2022-10-21
Clang-Tidy: disable the readability-identifier-length check

oleg-derevenetz:
There are lots of valid and clear identifier names like `x`/`y`, `px`/`py`, `pt`, etc, etc.


ihhub:
@oleg-derevenetz , thanks for the fix!

Issue #3718, opened 2021-06-19
AI: improve the logic for army and artifact exchange between 2 AI heroes


ihhub:
We have a logic for meeting two AI heroes in `void AIMeeting( Heroes & left, Heroes & right )` function.

The logic for army exchange could be described shortly as "Move the strongest monsters from one hero to another". It is noticeable that during this exchange if one of the heroes for example has only one stack of Dwarves while another hero has Peasants and Dwarves the first hero will keep 1 Dwarf instead of having 1 Peasant. This should be improved.

Artifact exchange won't take into account "bad" artifacts or artifacts that generate resources. For example, Endless Purse of Gold has the same value as Power Axe. For the stronger hero Power Axe makes more point in battles.

ihhub:
Hi @oleg-derevenetz , would you mind to take a look at this issue?

oleg-derevenetz:
Hi @ihhub I've shifted my interests a bit for now, so to speak, from gamedev 😎 to developing firmware for various Keenetic routers, there are also interesting tasks there, but I'll try to somehow allocate time.

ihhub:
@oleg-derevenetz , no problem at all! You should do whatever is in your interest 🙂

PR #6016, opened 2022-10-21
Avoid using armies with 1 monster for AI heroes

ihhub:
Having just one monster in the army is very risky. Plus sometimes the weakest troop is too weak to even been present in the army.

We use 5% threshold to determine how many monsters we need to give back to another hero.

relates to #3718

Also renamed one method because it wasn't obvious what exactly it returns.


src/fheroes2/army/army.cpp
```diff
@@ -720,15 +721,81 @@ void Troops::JoinStrongest( Troops & troops2, bool saveLast )
...
+                if ( acceptableCount < 1 ) {
+                    acceptableCount = 1;
+                }
+
+                troops2.JoinTroop( *weakest, acceptableCount );
+                weakest->SetCount( weakest->GetCount() - acceptableCount );
+            }
+
+            // Make sure that this hero can survive an attack by splitting a single stack of monsters into multiple.
+            // The slower the monster the higher chance it would be killed being in one stack so more stacks are advisable.
```
oleg-derevenetz, line 752: Why not always split them into the maximum possible number of stacks?

oleg-derevenetz (commented):
Hi @ihhub I just left one optional suggestion and one question, please take a look when you have time.

src/fheroes2/army/army.h
```diff
@@ -68,7 +68,7 @@ class Troops : protected std::vector<Troop *>
 
     double getReinforcementValue( const Troops & reinforcement ) const;
 
-    uint32_t GetCount() const;
+    uint32_t GetOccupiedMonsterSlotCount() const;
```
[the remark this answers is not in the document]
ihhub: Sure!

src/fheroes2/army/army.cpp
```diff
@@ -720,20 +720,80 @@ void Troops::JoinStrongest( Troops & troops2, bool saveLast )
...
-            weakest->SetCount( weakest->GetCount() - 1 );
+            // First check if the weakest troop is actually worth to keep.
+            const double weakestStrength = weakest->GetStrength();
+            const double totalArmyStrength = GetStrength();
+            // The weakest army should not be more than 5% from the overall army strength.
+            const double strengthLimit = totalArmyStrength / 20;
+
+            if ( weakestStrength < strengthLimit ) {
+                // The weakest troop is less than limit. Just kick this weakling out.
+                troops2.JoinTroop( *weakest, weakest->GetCount() );
```
oleg-derevenetz, line 731: Hi @ihhub It would be good also to check the return value of such `JoinTroop()` calls, because their failure is apparently not a normal behavior here:

```cpp
if ( [...].JoinTroop( ... ) ) {
  assert( 0 );
}
```

What do you think?

ihhub: I am always for adding more checks to have a stable code.

ihhub:
Hi @oleg-derevenetz , I went a bit beyond what was requested because I couldn't look at such a horror in the code.

src/fheroes2/army/army.cpp
```diff
@@ -710,30 +715,106 @@ void Troops::JoinStrongest( Troops & troops2, bool saveLast )
...
+    }
 
-    const uint32_t chunk = troop.GetCount() / slots;
-    uint32_t remainingCount = troop.GetCount() % slots;
-    uint32_t remainingSlots = slots;
+    const uint32_t chunk = troop.GetCount() / maxSlots;
+    uint32_t remainingCount = troop.GetCount() % maxSlots;
+    uint32_t remainingSlots = maxSlots;
 
     auto TryCreateTroopChunk = [&remainingSlots, &remainingCount, chunk, troop]( Troop & newTroop ) {
```
oleg-derevenetz, line 819: Despite the `troop` is itself a reference, it [will be captured by value](https://godbolt.org/z/hca6b9n44) here instead of reference. I believe this is not desirable here and capture should be replaced by `&troop`.

oleg-derevenetz (approved):
Hi @ihhub LGTM, I just left one optional suggestion.

src/fheroes2/army/army.cpp
ihhub: Done.

PR #6017, opened 2022-10-21
Fix View World Map Symbols positioning

### LINKED ISSUE #4753: View World screen, all symbols (Mine, Artifact, etc.) are shifted
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, all symbols (Mine, Artifact, etc.) are shifted down.
Compare with the mines in the center completely south of the adventure map.
In the original game, there is a space between the mines and the bottom of the map, whereas in fheroes2 the mines are completely at the bottom of the map.
Also compare the artifacts that are indicated outside the map, as well as the resource piles in fheroes2.
The symbols are also not correctly centered horizontally on the objects.
Even the hero symbols are not centered correctly.

In original game:

View World icons position 001

In fheroes2:

View World icons position 002

In original game:

View World icons position 003

View World icons position 004

In fheroes2:

View World icons position 005

View World icons position 006

### Save file

Here is the save file:
[Broken_Alliance_0061 b.zip](https://github.com/ihhub/fheroes2/files/7655845/Broken_Alliance_0061.b.zip)

### Additional info

Related issue: https://github.com/ihhub/fheroes2/issues/4200.

HaveFun91:
Close #4753


HaveFun91:
@LeHerosInconnu i've made some improvements also based on Zoom Level of the map. Could you check whether it is ok now?

src/fheroes2/kingdom/view_world.h
```diff
@@ -41,6 +41,12 @@ enum class ViewWorldMode : int
     ViewAll = 6,
 };
 
+struct IconOffset
```
oleg-derevenetz, line 44: Hi @HaveFun91 if this struct is used just in the `view_world.cpp`, please move it to the anonymous namespace in this file. There is no reason to have it in the header file.

LeHerosInconnu:
@HaveFun91.

I will take a look at it tomorow. :)

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -132,6 +154,16 @@ namespace
         }
     }
 
+    IconOffset GetIconOffset( MP2::MapObjectType objectType, ViewWorld::ZoomLevel zoomLevel )
+    {
+        std::pair<MP2::MapObjectType, ViewWorld::ZoomLevel> key = std::make_pair( objectType, zoomLevel );
```
ihhub, line 159: This variable can be `const`.

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -132,6 +154,16 @@ namespace
         }
     }
 
+    IconOffset GetIconOffset( MP2::MapObjectType objectType, ViewWorld::ZoomLevel zoomLevel )
```
ihhub, line 157: Since this function is only called from one place we can accept `const std::pair<MP2::MapObjectType, ViewWorld::ZoomLevel> & key` instead of copying parameters 2 times. And call this function with making a pair.

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -132,6 +154,16 @@ namespace
...
+    {
+        std::pair<MP2::MapObjectType, ViewWorld::ZoomLevel> key = std::make_pair( objectType, zoomLevel );
+
+        if ( iconsOffsetMap.count( key ) > 0 )
+            return iconsOffsetMap.at( key );
```
ihhub, line 162: Here we are doing 2 search calls. We need to use an iterator:
```cpp
auto iter = iconsOffsetMap.find( key );
if ( iter != iconsOffsetMap.end() ) {
    return iter->second;
}
```

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -50,10 +51,31 @@ namespace
...
 {
-    const int tileSizePerZoomLevel[4] = {4, 6, 12, 32};
-    const int icnPerZoomLevel[4] = {ICN::MISC4, ICN::MISC6, ICN::MISC12, ICN::MISC12};
-    const int icnLetterPerZoomLevel[4] = { ICN::LETTER4, ICN::LETTER6, ICN::LETTER12, ICN::LETTER12 };
-    const int icnPerZoomLevelFlags[4] = {ICN::VWFLAG4, ICN::VWFLAG6, ICN::VWFLAG12, ICN::VWFLAG12};
+    struct IconOffset
+    {
+        int offsetX;
+        int offsetY;
+    };
```
ihhub, line 58: Please remove this structure and use `fheroes2::Point` structure.

ihhub (changes requested):
Hi @HaveFun91 , I left several comments in this pull request. Would you mind to take a look at them?

src/fheroes2/kingdom/view_world.cpp
HaveFun91: Well... but wouldn't it be better to leave implementation details to inside function? I know we use it once, but I think it's nice encapsulation and maybe it's better to leave it as it is? Since anyway we have to create a pair I don't see any performance overhead here.

HaveFun91: I made some change to pass value by reference so no copy happens if i get it right.

LeHerosInconnu:
@HaveFun91.

> @HaveFun91.
> 
> I will take a look at it tomorow. :)

I get this window when I try to run the executable ([fheroes2_windows_x64_SDL2.zip](https://github.com/ihhub/fheroes2/suites/8908943617/artifacts/407960487)) :

Crash 22 10 2022 01

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -132,6 +148,18 @@ namespace
         }
     }
 
+    fheroes2::Point GetIconOffset( const MP2::MapObjectType & objectType, const ViewWorld::ZoomLevel & zoomLevel )
```
ihhub, line 151: Passing parameters by const reference is not advisable for POD (plain old data) such as int. Please remove `&` from both of variables.

ihhub (changes requested):

ihhub:
Hi @HaveFun91 , could you please rebase this branch from master so @LeHerosInconnu will be able to test it?

HaveFun91:
@ihhub, @LeHerosInconnu rebased, should work now.

ihhub (approved):

ihhub:
Hi @Districh-ru , did we fix it recently?

ihhub:
I am closing this pull request as the fix was implemented in another pull request.

Districh-ru:
> Hi @Districh-ru , did we fix it recently?

Hi, @ihhub. Yes, the fix was made in #6562. But @LeHerosInconnu has some questions which I think were solved by #6624.

Issue #6008, opened 2022-10-20
Android release preparations


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Even having a very basic Android build we are far away from a proper release for this platform. We need to make multiple changes and adjustments in order to make it user-friendly and ready to be published in App Store.

### Describe the possible solution

- [x] lock the application in landscape mode. Right now it is possible to rotate the view.
- [x] add compilation flags which are present in other builds, especially for warnings.
- [x] make `files/data` directory and compiled translations to be a part of the package to avoid extra resource copying even by developers.
- [x] Android filesystem is case-sensitive. Right now the code does not handle it properly. The logic present for case-sensitive filesystems is not applicable for Android due to a fact that an Android application has access (even for reading) only to certain directories.
- [x] add proper support of internal and external storages.
- [x] evaluate the need of permissions needed for the application.
- [x] add Java code to verify the presence of files and possibly download a demo version and unzip the downloaded archive into the specific location.
- [x] add a script to setup Android build for developer. Include needed paths to `.gitignore` file (looks like it is done).
- [x] fix MIDI music for Android. Android has native support of it.
- [x] fix resolution detection. In the Android devices the height is bigger than width.
- [x] add package signing as a part of CI/CD pipeline.
- [x] finalize minimum Android version requirements for the application.
- [x] right mouse click simulation
- [x] add a UI element to toggle between full-screen and windowed modes
- [x] connect system audio level to game's level of music or audio or both
- [x] add code to open keyboard for typing words like it is done for Switch and Vita

oleg-derevenetz:
> Android filesystem is case-sensitive. Right now the code does not handle it properly. The logic present for case-sensitive filesystems is not applicable for Android due to a fact that an Android application has access (even for reading) only to certain directories.

If we are going to distribute APK via direct download, we can just use the [MANAGE_EXTERNAL_STORAGE](https://developer.android.com/training/data-storage/manage-all-files) permission and screw the Google "best practices".

> Android has native support of it.

I'm not sure about that. `SDL_Mixer` [has support](https://github.com/libsdl-org/SDL_mixer/tree/main/src/codecs/native_midi) for native MIDI on MacOS X and Windows, but not on Android. It can be built with TiMidity though (just like I did for our Android package), but in this case it requires additional files (GUS patches).

[...]

PR #6018, opened 2022-10-21
Add scripts for Windows, Linux and macOS to install Android pre-built packages

oleg-derevenetz:
Related to #6008

Also make `curl` call in the `download_demo_version.sh` a bit more sane.


ihhub:
@oleg-derevenetz , great job!

PR #6023, opened 2022-10-22
Android: lock the screen orientation to landscape

oleg-derevenetz:
Related to #6008

Hi @ihhub is this place "good enough" to set this hint, or is there some better place for this?

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -974,7 +974,9 @@ namespace
 
 #if defined( ANDROID )
             // Same as ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
-            SDL_SetHint( SDL_HINT_ORIENTATIONS, "LandscapeLeft LandscapeRight" );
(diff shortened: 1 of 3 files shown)


src/engine/screen.cpp
```diff
@@ -972,6 +972,11 @@ namespace
                 height_ = correctResolution.height;
             }
 
+#if defined( ANDROID )
+            // Same as ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
+            SDL_SetHint( SDL_HINT_ORIENTATIONS, "LandscapeLeft LandscapeRight" );
```
ihhub, line 977: This is a "good enough" place for this hint but we should add error check:
```cpp
if ( SDL_SetHint( SDL_HINT_ORIENTATIONS, "LandscapeLeft LandscapeRight" ) == SDL_FALSE ) {
    ERROR_LOG( "Failed to set a landscape orientations." )
}
```

ihhub (changes requested):

ihhub:
@oleg-derevenetz , great job! Thanks for this improvement!

PR #6026, opened 2022-10-23
Fix Campaign score sorting issue

### LINKED ISSUE #6011: Campaign High scores are sorted in a wrong order
### Preliminary checks

### Platform

Other

### Describe the bug

If the campaign high score is generated from anew in latest build it will be in the opposite order or wrong in some other way:
image

Here is the correct order:
(The original game doesn't show score)
image

I had to delete my fheroes2.hgs to test this using the latest version. Standard maps have the correct order.

Also, this means that the longer you take the higher up you will get on the high score. Currently if your high score does not end up on the table it means that you used less time than the lowest one. I guess it works more as a high score of shame :P

### Save file

Just follow the instructions above.

### Additional info

_No response_

ihhub:
close #6011

### DIFF
--- a/src/fheroes2/game/game_highscores.cpp
+++ b/src/fheroes2/game/game_highscores.cpp
@@ -188,7 +188,7 @@ fheroes2::GameMode Game::DisplayHighScores( const bool isCampaign )
             const auto & campaignHighscoreData = highScoreDataContainer.getHighScoresCampaign();
             assert( !campaignHighscoreData.empty() );
 
-            if ( campaignHighscoreData.back().rating > rating ) {
+            if ( campaignHighscoreData.back().rating < rating ) {
                 gameResult.ResetResult();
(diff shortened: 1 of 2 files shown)


zenseii (commented):
The scores are sorted correctly. I've built and tested this after deleting the `fheroes2.hgs`.

One issue I'm seeing is that even though I'm completing the Descendants campaign in 8 days, I'm still not prompted to input my name and my score is not displayed on the high scores.

I've attached a save to right before the final battle:
[descendants before complete.zip](https://github.com/ihhub/fheroes2/files/9845770/descendants.before.complete.zip)

oleg-derevenetz (approved):

ihhub:
Hi @zenseii , thank you for spotting another issue! I fixed this case.

zenseii (approved):
Tested and it works perfectly now.

PR #6031, opened 2022-10-23
Generate uniform MAX/MIN buttons with new method

### LINKED ISSUE #6006: Incorrect display of MIN button in 'move troops' dialog window
### Preliminary checks

### Platform

Windows

### Describe the bug

Hi,
'MIN' button has incorrect drawing in the 'move troops' dialog window.
In recruit dialog this button generates correctly.
I am using russian resources and this incorrect drawing is noticeable.

изображение  изображение

### Save file

No save file is needed, just start a new game and try to split the troops in hero's army.

### Additional info

There is a function ('generateLanguageSpecificImages'), that generates 'MIN' button for recruit dialog (NOT_UNIFORM_GOOD_MIN_BUTTON).
For 'move troops' dialog it is generated from parts of other buttons (UNIFORM_GOOD_MIN_BUTTON).
Maybe it should be generated like in recruit dialog.

zenseii:
Close #6006 

This generates the uniform (i.e. without shadows?) MAX/MIN buttons using the new methods so that they can be translated.

This is an initial step towards making the game's UI translatable.

image


src/fheroes2/agg/agg_image.cpp
```diff
@@ -481,6 +485,33 @@ namespace fheroes2
...
+            case ICN::UNIFORM_GOOD_MIN_BUTTON: {
+                _icnVsSprite[id].resize( 2 );
+
+                const bool isGoodInterface = ( id == ICN::UNIFORM_GOOD_MAX_BUTTON || id == ICN::UNIFORM_GOOD_MIN_BUTTON );
+                const int baseIcnId = isGoodInterface ? ICN::SYSTEM : ICN::SYSTEME;
+                const fheroes2::FontColor buttonFontColor = isGoodInterface ? fheroes2::FontColor::WHITE : fheroes2::FontColor::GRAY;
+
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    out.resize( GetICN( ICN::RECRUIT, 4 + i ).width() - 6, GetICN( ICN::RECRUIT, 4 + i ).height() - 5 + i );
```
ihhub, line 500: Since `GetICN()` is not an inline-function it should be good to cache it:
```cpp
Sprite & originalButton = GetICN( ICN::RECRUIT, 4 + i );
```

ihhub (changes requested):

Branikolog:
> Hi @ihhub
> 
> I believe I might need to make the button wider to allow for languages that use 4 letters? image
> 
> This would also become a problem in the creature recruit dialog when that MAX button will be generated, so I'm not sure how we want to deal with this exactly. The MIN and MAX buttons need to be the same width too.
> 
> Maybe we could find the width of the translated text and use that to determine how wide the buttons should be? This should probably be a generally available function.

Hi, @zenseii !

In Russian resources developers managed to fit the text into this button by making letters more tight and reducing the spaces between letters where possible.
MAX

zenseii:
> In Russian resources developers managed to fit the text into this button by making letters more tight and reducing the spaces between letters where possible.

Hi @Branikolog.

I've seen that the Russian original has very tight space between the letters. I know how to simply make all letters tighter just for Cyrillic, but I'm not sure this is the best way to go either... Also, note that these Russian original buttons were made in an image editing program and were not generated by code like we do, which is why they were able to adjust the text on every single button. Ours will need to be a general solution.

Branikolog:
@zenseii 

>Also, note that these Russian original buttons were made in an image editing program and were not generated by code like we do, which is why they were able to adjust the text on every single button. Ours will need to be a general solution.

Yeah, it's obvious. 
On the other side we can redraw some letters so they become more tight and the needed text would fit without individual adjustment of the text itself. I also believe, there could be even implemented a special law that could adjust spacing depending on edge pixels of both nearby letters. But I'm not sure it would be a trivial task... :)

ihhub:
Hi @zenseii and @Branikolog , we can make the button wider but I would suggest to keep the width the same for all languages.

I can check how to do it nicely.

zenseii:
I believe we can easily adjust the button length of any of these buttons using the template buttons we are currently using anyways:
image

It's just a matter of extending the middle part, and if we want to we can use them for larger buttons too by extending upwards.

ihhub:
> I believe we can easily adjust the button length of any of these buttons using the template buttons we are currently using anyways: image
> 
> It's just a matter of extending the middle part, and if we want to we can use them for larger buttons too by extending upwards.

Yes, we can. I suggest to do everything in stages. We can proceed with changes in this pull request as of now and later add a separate chunk of code to extend the button width.

src/fheroes2/agg/agg_image.cpp
zenseii: I've done this in the latest commit.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -481,6 +485,33 @@ namespace fheroes2
...
+                const int baseIcnId = isGoodInterface ? ICN::SYSTEM : ICN::SYSTEME;
+                const fheroes2::FontColor buttonFontColor = isGoodInterface ? fheroes2::FontColor::WHITE : fheroes2::FontColor::GRAY;
+
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    Sprite & out = _icnVsSprite[id][i];
+                    const Sprite & originalButton = GetICN( ICN::RECRUIT, 4 + i );
+                    out.resize( originalButton.width() - 6, originalButton.height() - 5 + i );
+                    out.reset();
+                    // Copy left main body of button.
+                    fheroes2::Copy( GetICN( baseIcnId, 11 + i ), 0, 0, out, 0, 0, originalButton.width() - 13 + i, originalButton.height() - 5 + i );
```
ihhub, line 504: A small optimization: we can also cache `GetICN( baseIcnId, 11 + i )` call and it will slightly improve code readability.

ihhub:
@zenseii , huge thanks for making this change!

Issue #5984, opened 2022-10-15
Trouble with sound while playing video


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

There's a terrible noise appearing occasionally during "Roland" campaign introduction video.
Also tone is changing randomly during video playback.
The issue appeared somewhere at between 11 and 21 of september (on 0.9.19 version).

### Save file

None.

zenseii:
The commits about audio in that timeframe are these two:
https://github.com/ihhub/fheroes2/pull/5829
https://github.com/ihhub/fheroes2/pull/5846

oleg-derevenetz:
About a month ago we migrated the Windows build to SDL 2.24.0, and this build has the following regression:

https://github.com/libsdl-org/SDL/issues/6391

This regression sometimes causes crashes (I got a few during the video playback BTW), but also it may also cause sound distortion/noise, because, since SDL audio resampling implementation now reads outside of its buffers, it may read random garbage, and this may produce wrong resampled output.

zenseii:
Hi, everyone.

I watched a fheroes2 video on twitch by streamer Lexiav and the issues with sound were very bad as you have described them.

There doesn't seem to have been made a PR on SDL's repo for this issue yet and since there is really no way of knowing when this will be fixed, what are our alternatives for solving this issue?

Is it possible to revert to an older SDL that did not have this regression? I'm just throwing ideas in the air here.

ihhub:
Ideally we need to roll back to SDL version which does not have such a bug. We did not jump a lot in versions. Plus we need to make sure SDL Mixer will still work with an older SDL.

oleg-derevenetz:
I rolled back SDL in the Windows package to the previous version (2.24.0 -> 2.0.22).

ihhub:
Hi @Branikolog , could you please test this issue and if it's fixed in the latest build close it?

Branikolog:
Hi, @ihhub and @oleg-derevenetz  !
The issue is gone!
Thank you for fixing it so much!

PR #6032, opened 2022-10-23
Android: implement basic filesystem access

oleg-derevenetz:
Now it is done as follows:

* OG data files should be placed to the `Android/data/org.fheroes2/files` directory relative to the phone's sdcard path (including `resurrection.h2d` and `*.mo` files). No debugger needed, just copy files as usual. The directory layout should be as follows:

```
Android/data/org.fheroes2/files/DATA
Android/data/org.fheroes2/files/MAPS
Android/data/org.fheroes2/files/MUSIC
Android/data/org.fheroes2/files/files/data/resurrection.h2d
Android/data/org.fheroes2/files/files/lang/*.mo
```

BE ADVISED THAT THIS DIRECTORY WILL BE REMOVED WHEN YOU UNINSTALL THE APPLICATION.

* Case insensitivity for data file names is fully supported;
* Save files will be placed to the `Android/data/org.fheroes2/files/files/save` (relative to the sdcard path).

Now APK is uploaded as PR build artifact, so literally anyone can try it, no special Android developer skills are required for installation.

PLEASE NOTE THAT THE APK IS BUILT IN DEBUG MODE, SO DON'T EXPECT AMAZING PERFORMANCE.


oleg-derevenetz:
BTW, don't you think that the list of languages has practically ceased to fit in 640x480?

ihhub:
Hi @oleg-derevenetz , I'm aware of the size problem with language dialog. I will fix it for the upcoming release.

ihhub:
@oleg-derevenetz , what's about phones without SD card support?

ihhub:
One more question: are we planning to make resurrection.h2d and translation files as a part of APK? Ideally, we should do this but the changes in this pull request do not support such.

oleg-derevenetz:
Hi @ihhub

> what's about phones without SD card support?

Despite its name, `/sdcard` refers to the built-in flash memory. In this regard, there cannot be phones without SD card at all :) But, while the `/sdcard` is the commonly used path, and even mentioned in some Android documentation, it still may be not 100% portable. There is the `Environment.getExternalStorageDirectory()` to determine the exact location, but it is deprecated in Android 10+. Another headache.

oleg-derevenetz:
@ihhub

> are we planning to make resurrection.h2d and translation files as a part of APK? Ideally, we should do this but the changes in this pull request do not support such.

This PR is just a part of a whole. To make these files a part of APK, we will probably need to extract them from so-called "app assets" before starting the game itself, because it will be not able to access these assets as a regular files, there is a need to use special API to access them.

oleg-derevenetz:
I suppose we can use the result of the `SDL_AndroidGetExternalStoragePath()` instead of the hardcoded `/sdcard/fheroes2`.

oleg-derevenetz:
> I suppose we can use the result of the `SDL_AndroidGetExternalStoragePath()` instead of the hardcoded `/sdcard/fheroes2`.

Yeah, it works just fine (I updated the PR description). But there is one pitfall: this directory will be removed when app is uninstalled.

ihhub:
> > I suppose we can use the result of the `SDL_AndroidGetExternalStoragePath()` instead of the hardcoded `/sdcard/fheroes2`.
> 
> Yeah, it works just fine (I updated the PR description). But there is one pitfall: this directory will be removed with all its contents when app is uninstalled. It is necessary to keep this in mind.

Will it be removed in case of application update? I assume it works like uninstall the old version and install a new one?

oleg-derevenetz:
@ihhub

> Will it be removed in case of application update? I assume it works like uninstall the old version and install a new one?

If you install the new APK while the old one is still installed - no, you will not lose anything. BUT, if you UNINSTALL the previous version first and then install a new one - then yes, it will be removed on uninstall and re-created (empty) when the new version is installed.

ihhub:
> @ihhub
> 
> > Will it be removed in case of application update? I assume it works like uninstall the old version and install a new one?
> 
> If you install the new APK while the old one is still installed - no, you won't lose anything. BUT, if you UNINSTALL the previous version first and then install a new one - then yes, it will be removed on uninstall and re-created (empty) when the new version is installed.

It's absolutely expected behavior but we'll highlight this in documentation.

oleg-derevenetz:
> It's absolutely expected behavior but we'll highlight this in documentation.

Yeah, but it's not handy when you always forced to remove the app to check the process of obtaining necessary permissions :) However, this is manageable. Also, users may be may be unhappy when they see that their data and save files are gone when they accidentally uninstalled the app. But, after all, this is a mainstream now.

oleg-derevenetz:
It seems that we don't need the `MANAGE_EXTERNAL_STORAGE` permission to access files that were put to the app-specific data directory on `/sdcard`, which is good. It is not entirely clear however are `READ_EXTERNAL_STORAGE` and `WRITE_EXTERNAL_STORAGE` are required to access this directory on Android 9 and earlier (they are 100% not required on Android 10+ because of so-called "scoped storage" feature). This needs to be checked separately.

oleg-derevenetz:
> It is not entirely clear however are READ_EXTERNAL_STORAGE and WRITE_EXTERNAL_STORAGE required to access this directory on Android 9 and earlier

It seems that they are [NOT required](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE) to access app-specific directory even on earlier Android versions, which simplifies things even more.

oleg-derevenetz:
Hi @ihhub I suppose this PR is ready for review. Placement of h2d and translations into APK will be addressed separately.

oleg-derevenetz:
BTW, it looks like I hear sound distortion like in #5984. Maybe we need to build Android app with an earlier SDL version as well (now it uses the latest release that has known issues with the sound). This needs additional verification and will be addressed separately.

ihhub:
1C064655-3F4C-4F28-8EC0-D10D039A7F76

ihhub:
> BTW, it looks like I hear sound distortion like in #5984. Maybe we need to build Android app with an earlier SDL version as well (now it uses the latest release that has known issues with sound). This needs additional verification and will be addressed separately.

I can confirm this distortion as well. Let's build with earlier version if it helps.

ihhub:
@oleg-derevenetz , wonderful job! Thanks a lot for taking care of Android build!

oleg-derevenetz:
> I can confirm this distortion as well. Let's build with earlier version if it helps.

It looks like it doesn't help. BTW this is not observed in the x86_64 simulator - sound in the simulator seems to work just fine. I also found that this fire sound in the introductory animation sounds too early on device (and just in time on the simulator).

ihhub:
@oleg-derevenetz , the problem might be with SDL mixer itself🤔

oleg-derevenetz:
@ihhub Ahaha, LOL. I had an impression that this issue is related to buffering, and tried to increase `samples` in the `SDL_AudioSpec` (more precisely, our own `struct Spec` inherited from it) from 2048 to 4096, but things just get noticeably worse. So I DECREASED `samples` to 1024, and it looks like all distortions are gone! Sound in animations now played just in time, and music is played just fine (at least MP3).

ihhub:
@oleg-derevenetz , the problem has solved by itself! I think we can adjust the value with no issues.

oleg-derevenetz:
@ihhub Did this resolve the distortions for you as well?

ihhub:
@oleg-derevenetz , I need to check this. Give me few minutes...

PR #6033, opened 2022-10-23
Add proper CP1250 font

zenseii:
This adds a proper font with special letters to support CP1250 encoding. Full font support for Polish and Hungarian and some Czech characters (all except the ones with caron and the u with circle).

image
image


ihhub:

PR #6040, opened 2022-10-24
Android: distribute *.h2d and *.mo files as part of the APK, bundle the TiMidity GUS patches to make MIDI work

oleg-derevenetz:
Related to #6008

Also I believe it's time to add APK to the pre-releases/snapshots. Please note that it is still built in debug mode because we don't have a proper signature.

Also minor style nits regarding package & tag names for releases.


oleg-derevenetz:
Hi @ihhub I believe this PR is ready for review. You can also try MIDI music, it should work just fine. I added GUS patches only for necessary instruments (just like it's done with SoundFont for Windows builds), so APK size is much less than it might be :)

android/app/src/main/java/org/fheroes2/MainActivity.java
```diff
@@ -20,10 +20,32 @@
...
+import org.apache.commons.io.IOUtils;
 import org.libsdl.app.SDLActivity;
 
 public class MainActivity extends SDLActivity
 {
+    @Override
+    protected void onCreate( Bundle savedInstanceState )
+    {
+        // Extract H2D and translations to the external app-specific storage (sdcard)
+        extractAssets( "files", getExternalFilesDir( null ) );
```
ihhub, line 40: Copying files (few MB) every time we startup an application might not be the best approach from performance and storage perspective.

Is this the only way to handle assets? Would [symbolic links](https://stackoverflow.com/questions/27726428/symbolic-link-creation-in-android-within-an-applications-asset-directory) help?

ihhub (changes requested):
Hi @oleg-derevenetz , I put 2 comments in this pull request. Could you please take a look at them?

MIDI works for me.

android/app/src/main/java/org/fheroes2/MainActivity.java
oleg-derevenetz: @ihhub This is the only documented & portable way. The specific mechanism of storing assets is undocumented and and is an implementation detail, which may change from Android version to Android version (and maybe even from device to device).

We can add some checks (for example, if there is already a file, do not copy it), but this may cause potential issues when this file should be updated in a new version of the app. Copying a few MB takes just a moment on the modern devices (about the same time that takes a load from flash card to memory due to cache), so I believe there is nothing to worry about. If there really will be any problems with this, then we will think about what to do with it.

ihhub: Fair point. I didn't notice any slowdowns as well so let's keep it as it is.

oleg-derevenetz: @ihhub BTW the assets are usually stored in some compressed form, so they need to be decompressed anyway before they can be used by the app. As I already said, the exact mechanism is an implementation detail.

ihhub (approved):

PR #6042, opened 2022-10-25
Translate language names into their native languages

zenseii:
This PR translates the language names so that they display correctly in the language selection dialog.


ihhub:
@zenseii , huge thanks for improving translations!

PR #6043, opened 2022-10-25
Modify and fix base font characters

### LINKED ISSUE #4799: Game, English language font, there is a horizontal bar "_" at the top with the "%" character
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, there is a horizontal bar "_" at the top with the "%" character.

In fheroes2:

Barre pourcentage english 01

This is also the case in the French language font.
French translation 017

### Save file

Not needed.

### Additional info

_No response_

### LINKED ISSUE #5514: "k" letter should be the same for all languages
### Preliminary checks

Even at language selection menu we can watch two different styles of  "k". (Compare "Polski" and "Norsk")
k

I think, we should stick to smaller one, since original "k" looks like a capital one.

### Additional info

_No response_

zenseii:
This adds logic that will modify the base fonts on the fly.

The modifying logic was placed in `ui_font.cpp` to maintain current structure and because the function updating the font shadow is local to that file.

Lowercase k for both normal and small font have been edited as an example and these were also requested by many users.
This PR fixes other font issues like the `%` and `-`.

New `k`:
image
image

Fixed `%` sign:
image
image

Fixed hyphen `-` (only normal font has issues):
image

The Cyrillic `к` in CP1251 has been readjusted because it uses lowercase k for its generation.

Close #5514 
Close #4799

### DIFF
--- a/src/fheroes2/gui/ui_font.cpp
+++ b/src/fheroes2/gui/ui_font.cpp
@@ -4203,7 +4203,7 @@ namespace fheroes2
         updateButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED], icnVsSprite[ICN::BUTTON_GOOD_FONT_PRESSED], icnVsSprite[ICN::BUTTON_EVIL_FONT_RELEASED],
                           icnVsSprite[ICN::BUTTON_EVIL_FONT_PRESSED] );
     }
-    void editOriginalFont( std::vector<fheroes2::Sprite> & icnVsSprite, const bool normalFont ) 
+    void editOriginalFont( std::vector<fheroes2::Sprite> & icnVsSprite, const bool normalFont )
     {
         if ( normalFont ) {
             // Proper lowercase k. Kept at end in case any letters use it for generation.
--- a/src/fheroes2/gui/ui_font.cpp
+++ b/src/fheroes2/gui/ui_font.cpp
@@ -4219,7 +4219,7 @@ namespace fheroes2
         updateNormalFontLetterShadow( icnVsSprite[75] );
     }
 
-    void modifyBaseSmallFont(std::vector<fheroes2::Sprite>& icnVsSprite)
+    void modifyBaseSmallFont( std::vector<fheroes2::Sprite> & icnVsSprite )
     {
         assert( icnVsSprite.size() > 0 );
 
--- a/src/fheroes2/gui/ui_font.cpp
+++ b/src/fheroes2/gui/ui_font.cpp
@@ -4148,6 +4148,11 @@ namespace fheroes2
     {
         assert( !icnVsSprite.empty() );
 
+        // Remove white line from % symbol
+        fheroes2::FillTransform( icnVsSprite[5], 3, 0, 4, 1, 1 );
+        fheroes2::FillTransform( icnVsSprite[5], 4, 1, 2, 1, 1 );
+        updateNormalFontLetterShadow( icnVsSprite[5] );
+
         // Proper lowercase k.
         icnVsSprite[75].resize( 6, 8 );
         icnVsSprite[75].reset();
--- a/src/fheroes2/gui/ui_font.cpp
+++ b/src/fheroes2/gui/ui_font.cpp
@@ -4124,6 +4124,11 @@ namespace fheroes2
     {
         assert( !icnVsSprite.empty() );
 
+        // Remove white line from % symbol
+        fheroes2::FillTransform( icnVsSprite[5], 5, 0, 5, 1, 1 );
+        fheroes2::FillTransform( icnVsSprite[5], 6, 2, 2, 1, 1 );
+        updateNormalFontLetterShadow( icnVsSprite[5] );
+
         // Proper lowercase k.
         fheroes2::FillTransform( icnVsSprite[75], 4, 1, 5, 8, 1 );
         fheroes2::Copy( icnVsSprite[43], 6, 5, icnVsSprite[75], 4, 7, 3, 1 );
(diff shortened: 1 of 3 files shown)


src/fheroes2/gui/ui_font.cpp
```diff
@@ -4203,4 +4203,31 @@ namespace fheroes2
         updateButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED], icnVsSprite[ICN::BUTTON_GOOD_FONT_PRESSED], icnVsSprite[ICN::BUTTON_EVIL_FONT_RELEASED],
                           icnVsSprite[ICN::BUTTON_EVIL_FONT_PRESSED] );
     }
+
+    void editOriginalFont( std::vector<fheroes2::Sprite> & icnVsSprite, const bool normalFont )
+    {
+        if ( normalFont ) {
```
ihhub, line 4209: Are we going to expand this function in the future? If not then I suggest to just put the code within `LoadModifiedICN()` function itself.
Also to not have any crashes it is good to check the size of `_icnVsSprite[id]` array to avoid out of boundaries element access.

ihhub (changes requested):

src/fheroes2/gui/ui_font.cpp
ihhub: Ah, I got the point why we put this code to a separate function.

src/fheroes2/gui/ui_font.cpp
```diff
@@ -4203,4 +4203,31 @@ namespace fheroes2
         updateButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED], icnVsSprite[ICN::BUTTON_GOOD_FONT_PRESSED], icnVsSprite[ICN::BUTTON_EVIL_FONT_RELEASED],
                           icnVsSprite[ICN::BUTTON_EVIL_FONT_PRESSED] );
     }
+
+    void editOriginalFont( std::vector<fheroes2::Sprite> & icnVsSprite, const bool normalFont )
```
ihhub, line 4207: Let's name it as `modifyBaseFont` as editing is usually done by humans.

ihhub: Also does it make more sense to just have 2 separate functions? It would improve code readability.

src/fheroes2/gui/ui_font.cpp
```diff
@@ -4203,4 +4203,31 @@ namespace fheroes2
...
+    {
+        if ( normalFont ) {
+            // Proper lowercase k. Kept at end in case any letters use it for generation.
+            fheroes2::FillTransform( icnVsSprite[75], 4, 1, 5, 8, 1 );
```
ihhub, line 4211: We must check the size of `icnVsSprite` before doing something.

ihhub (changes requested):

src/fheroes2/gui/ui_font.cpp
zenseii: Sure! I will make `modifyBaseNormalFont` and `modifyBaseSmallFont` in `fheroes2` namespace and call these within `LoadModifiedICN`.

src/fheroes2/gui/ui_font.cpp
zenseii: The functions now assert that `icnVsSprite` is not empty before doing anything.

src/fheroes2/gui/ui_font.cpp
```diff
@@ -4203,4 +4117,49 @@ namespace fheroes2
...
+        fheroes2::Copy( icnVsSprite[43], 7, 4, icnVsSprite[75], 6, 5, 3, 1 );
+        fheroes2::Copy( icnVsSprite[43], 7, 4, icnVsSprite[75], 7, 4, 2, 1 );
+        fheroes2::Copy( icnVsSprite[43], 6, 6, icnVsSprite[75], 4, 8, 4, 1 );
+        icnVsSprite[75].setPosition( icnVsSprite[75].x(), icnVsSprite[75].y() );
+        updateNormalFontLetterShadow( icnVsSprite[75] );
+    }
+
+    void modifyBaseSmallFont( std::vector<fheroes2::Sprite> & icnVsSprite )
+    {
+        assert( !icnVsSprite.empty() );
```
ihhub, line 4147: I would highly suggest to add this instead of the assertion:
```cpp
if ( icnVsSprite.size() < 96 ) {
    return;
}
```
as here we might try to access elements outside the container.

src/fheroes2/gui/ui_font.cpp
```diff
@@ -4203,4 +4117,49 @@ namespace fheroes2
         updateButtonFont( icnVsSprite[ICN::BUTTON_GOOD_FONT_RELEASED], icnVsSprite[ICN::BUTTON_GOOD_FONT_PRESSED], icnVsSprite[ICN::BUTTON_EVIL_FONT_RELEASED],
                           icnVsSprite[ICN::BUTTON_EVIL_FONT_PRESSED] );
     }
+
+    void modifyBaseNormalFont( std::vector<fheroes2::Sprite> & icnVsSprite )
+    {
+        assert( !icnVsSprite.empty() );
```
ihhub, line 4123: I would highly suggest to add this instead of the assertion:
```cpp
if ( icnVsSprite.size() < 96 ) {
    return;
}
```
as here we might try to access elements outside the container.

ihhub (changes requested):
Hi @zenseii , I put 2 more comments here. Could you please take a look at them?

ihhub (approved):

ihhub:
@zenseii , thank you so much for polishing fonts!

PR #6044, opened 2022-10-25
Android: setup the backup rules

oleg-derevenetz:
Related to #6008

Without special configuration, Android will attempt to backup almost all files both from private and external app data directories (except caches), which is not desirable, because per-app backup storage is very limited (up to 25MB per app) and we shouldn't backup OG data files, maps, music, etc, etc. This PR configures backup rules for Android 11 and lower devices, as well as for Android 12+ devices.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -32,6 +32,7 @@
 
     <application
         android:allowBackup="true"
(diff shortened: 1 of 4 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for this addition!

PR #6045, opened 2022-10-26
Add Czech language option in game and add fonts

zenseii:
This adds Czech language in the game and adds all normal and small fonts.

Button fonts should be made once we know what the buttons will have on them.


ihhub:

PR #6049, opened 2022-10-27
Fix the troop splitting logic

oleg-derevenetz:
Regression after #6016

See https://github.com/ihhub/fheroes2/issues/6008#issuecomment-1293007132 for issue details.


ihhub:
@oleg-derevenetz , thank for the quick fix!

PR #6055, opened 2022-10-28
Android: extract versionName and versionCode from global text configuration files

oleg-derevenetz:
Related to #6008


ihhub:

PR #6057, opened 2022-10-29
Android: build & sign the release APK on push

oleg-derevenetz:
Related to #6008

Steps to configure the release signing:

1. Create the keystore file in the JKS or PKCS12 format using the following command:

```
keytool -genkey -v -keystore fheroes2.keystore -alias release -keyalg RSA -keysize 2048 -validity 36500
```

The validity period of the certificate here is 10 years, maybe even more is better (50 years?).

2. Encode the resulting file as BASE64:

```
openssl base64 < fheroes2.keystore | tr -d '\n'
```

3. Create the following GitHub Actions Secrets:

* `ANDROID_KEYSTORE` with the value of the contents of the file `fheroes2.keystore`, encoded as BASE64;
* `ANDROID_KEYSTORE_PASSWORD` which should contain the password to the keystore file;
* `ANDROID_KEY_PASSWORD` which should contain the password to the `release` key alias.

That's all. The push workflow should work even if these secrets are absent, but it will produce the debug APK file instead of its release version.


src/fheroes2/agg/agg_image.cpp
```diff
@@ -65,10 +65,12 @@ namespace
                                                 ICN::BUTTON_DIFFICULTY_ROLAND,
                                                 ICN::BUTTON_DIFFICULTY_POL };
 
+#ifndef NDEBUG
```
ihhub, line 68: Could you please explain the purpose of  this change?

ihhub (changes requested):

src/fheroes2/agg/agg_image.cpp
oleg-derevenetz: Hi @ihhub this function is currently used just inside one `assert()` call, and clang reports the "unused function" warning when app is built with `NDEBUG`.

ihhub: Got it. Thank you very much for the explanation!

ihhub:
@oleg-derevenetz , I assume that I need to add GitHub Action secrets before proceeding with this pull request merge. I even think that 10 years is absolutely fine as we are going to update the game in the store often (hopefully).

oleg-derevenetz:
Hi @ihhub

> I assume that I need to add GitHub Action secrets before proceeding with this pull request merge.

This is not necessary, everything will work even without secrets (debug version will be created instead of release), they can be added later.

oleg-derevenetz:
Hi @ihhub

> I even think that 10 years is absolutely fine as we are going to update the game in the store often (hopefully).

I'm not sure how the app will be updated when it certificate will be changed. At least, If we will distribute the APK via GitHub release as well, there definitely will be issues with updating it. Try to install the app, then re-sign it with different certificate and update it using the resulting APK.

ihhub:
Hi @oleg-derevenetz , I tried to generate needed information. Encoded keystore and the password for it is possible to generate but I wasn't asked for a password to the release key alias.

oleg-derevenetz:
Hi @ihhub

> I tried to generate needed information. Encoded keystore and the password for it is possible to generate but I wasn't asked for a password to the release key alias.

That's how it looks like in my case:

```
$ keytool -genkey -v -keystore test.keystore -alias release -keyalg RSA -keysize 2048
Enter keystore password:
Re-enter new password:
What is your first and last name?
[Unknown]:  aaa
What is the name of your organizational unit?
[Unknown]:  bbb
What is the name of your organization?
[Unknown]:  ccc
What is the name of your City or Locality?
[Unknown]:  ddd
What is the name of your State or Province?
[Unknown]:  eee
What is the two-letter country code for this unit?
[Unknown]:  US
Is CN=aaa, OU=bbb, O=ccc, L=ddd, ST=eee, C=US correct?
[no]:  yes
Generating 2 048 bit RSA key pair and self-signed certificate (SHA256withRSA) with a validity of 90 days
for: CN=aaa, OU=bbb, O=ccc, L=ddd, ST=eee, C=US
Enter key password for <release>
(RETURN if same as keystore password):               <=== THIS
[Storing test.keystore]
```

ihhub:
This is how it works for me (Ubuntu 20):
```
keytool -genkey -v -keystore test.keystore -alias release -keyalg RSA -keysize 2048
Enter keystore password:  
Re-enter new password: 
What is your first and last name?
  [Unknown]:  aaa
What is the name of your organizational unit?
  [Unknown]:  bbb
What is the name of your organization?
  [Unknown]:  ccc
What is the name of your City or Locality?
  [Unknown]:  ddd
What is the name of your State or Province?
  [Unknown]:  eee
What is the two-letter country code for this unit?
  [Unknown]:  fff
Is CN=aaa, OU=bbb, O=ccc, L=ddd, ST=eee, C=fff correct?
  [no]:  yes

Generating 2,048 bit RSA key pair and self-signed certificate (SHA256withRSA) with a validity of 90 days
	for: CN=aaa, OU=bbb, O=ccc, L=ddd, ST=eee, C=fff
[Storing test.keystore]
```

oleg-derevenetz:
@ihhub maybe this version is a bit different and it doesn't ask this password by default. Apparently, it's the same password as for keystore itself.

ihhub:
Hi @oleg-derevenetz , assuming that the password are the same. The last question remains is population of this information. As far as I see we need to enter the correct one since this keystore will be used to sign the APK and its information will be available to public. Is it correct?

oleg-derevenetz:
@ihhub in perspective yes, but for testing purposes you can enter anything you want :)

ihhub:
Hi @oleg-derevenetz , I added required secrets. We will see if they are correct after merging this pull request.

ihhub:
@oleg-derevenetz , thank you so much for handling Android stuff!

ihhub:
Looks like it works :)

PR #6062, opened 2022-10-29
Android: update Gradle, build tools and Apache Commons IO dependency to the latest version

oleg-derevenetz:


android/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.3'
+        classpath 'com.android.tools.build:gradle:7.3.1'
```
ihhub, line 10: Here we specify 7.3.1 version while we download 7.4. Is it intentional?

android/app/build.gradle
```diff
@@ -40,7 +42,7 @@ android {
 
 dependencies {
     implementation project(':sdl2')
-    implementation 'org.apache.directory.studio:org.apache.commons.io:2.4'
+    implementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
```
ihhub, line 45: Could you please explain where this magic number 2.11.0 belongs to?

ihhub (commented):
Hi @oleg-derevenetz , I left 2 questions in this pull request. Could you please check them?

android/build.gradle
oleg-derevenetz: Hi @ihhub These are different things. This number is related to the version of Android build tools package from Google. 7.4 is related to the version of the Gradle itself.

android/app/build.gradle
oleg-derevenetz: Hi @ihhub it is related to the version of the corresponding build artifact in the Maven repository:

https://mvnrepository.com/artifact/commons-io/commons-io

android/build.gradle
ihhub: Thank you for the explanation!

android/app/build.gradle
ihhub: Thank you for the explanation!

ihhub:
@oleg-derevenetz , thank you very much for this improvement!

android/build.gradle
oleg-derevenetz: @ihhub More precisely this is called "Android plugin":

https://developer.android.com/studio/releases/gradle-plugin#7-3-0

:)

PR #6063, opened 2022-10-29
Update AI logic for monster distribution and recruitment

ihhub:
- while hiring monsters in a castle put monsters to the slots with the same type of monsters
- while attending dwellings on map try to redistribute army before joining monsters
- add logic to hire some monsters from dwellings if they are payable

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -900,7 +900,7 @@ bool Troops::mergeWeakestTroopsIfNeeded()
     assert( !sortedMultiSlotMonsterIdVsStrength.empty() );
 
     std::sort( sortedMultiSlotMonsterIdVsStrength.begin(), sortedMultiSlotMonsterIdVsStrength.end(),
-               []( const auto & left, const auto & right ){ return left.second < right.second; } );
+               []( const auto & left, const auto & right ) { return left.second < right.second; } );
 
     const int weakestMonsterToMerge = sortedMultiSlotMonsterIdVsStrength.front().first;
 
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -332,18 +332,19 @@ bool Troops::areAllTroopsUnique() const
 {
     std::set<int> monsterId;
 
-    size_t counter = 0;
-
     for ( const Troop * troop : *this ) {
-        if ( troop == nullptr || !troop->isValid() ) {
+        assert( troop != nullptr );
+        if ( !troop->isValid() ) {
             continue;
         }
 
-        monsterId.emplace( troop->GetID() );
-        ++counter;
+        auto [it, inserted] = monsterId.emplace( troop->GetID() );
+        if ( !inserted ) {
+          return false;
+        }
     }
 
-    return counter == monsterId.size();
+    return true;
 }
 
 bool Troops::HasMonster( const Monster & mons ) const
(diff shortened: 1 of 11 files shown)


src/fheroes2/army/army.h
```diff
@@ -72,12 +72,14 @@ class Troops : protected std::vector<Troop *>
...
+    bool areAllTroopsUnique() const;
+
     bool AllTroopsAreUndead() const;
     // Returns true if all valid troops have the same ID or if there are no troops, otherwise returns false
     bool AllTroopsAreTheSame() const;
 
-    bool JoinTroop( const Troop & );
-    bool JoinTroop( const Monster & mons, uint32_t count, bool emptySlotFirst = false );
+    bool JoinTroop( const Troop & troop );
+    bool JoinTroop( const Monster & mons, uint32_t count, bool emptySlotFirst );
```
oleg-derevenetz, line 82: Are you sure there is a need to remove the default value for this argument? Default arguments are handy sometimes :)

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few proposals, please take a look when possible.

src/fheroes2/army/army.cpp
```diff
@@ -328,6 +328,24 @@ uint32_t Troops::GetOccupiedSlotCount() const
...
 }
 
+bool Troops::areAllTroopsUnique() const
+{
+    std::set<int> monsterId;
+
+    size_t counter = 0;
+
+    for ( const Troop * troop : *this ) {
+        if ( troop == nullptr || !troop->isValid() ) {
```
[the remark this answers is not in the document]
ihhub: Sure. We also need to remove this weird way to allocate memory for individual members instead of storing values / objects.

src/fheroes2/army/army.cpp
```diff
@@ -840,6 +858,71 @@ void Troops::addNewTroopsToFreeSlots( const Troop & troop, uint32_t maxSlots )
...
+    }
+
+    if ( monsterIdVsStrength.size() == size() ) {
+        return false;
+    }
+
+    assert( monsterIdVsSlotCount.size() == monsterIdVsStrength.size() );
+
+    std::vector<std::pair<int, double>> sortedMultiSlotMonsterIdVsStrength;
+    for ( auto & [id, slotCount] : monsterIdVsSlotCount ) {
```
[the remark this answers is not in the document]
ihhub: Not, no reason.

src/fheroes2/army/army.h
ihhub: I removed the default value during the development to make sure that the logic is correct everywhere. The biggest problem with default values in cases like this that it is hard to understand what's going on in the code before you check these default arguments. So once I removed the value I found the place which was causing the incorrect logic.

If you think that it is good to put the default value back I am okay to do this :)

oleg-derevenetz: Hi @ihhub Well, as far as I can see, in the vast majority of places this argument is `false`, so it might be handy to have it `false` by default. But it's up to you to decide :) LGTM.

oleg-derevenetz (approved):

src/fheroes2/army/army.h
ihhub: Let's stick to no default parameter then since it would be obvious in the future for the new code.

PR #6067, opened 2022-10-30
Fix the translation generation for MSVC workflow

oleg-derevenetz:
Recent version of `gettext-devel` (0.21.1-1) is linked against `cygunistring-5.dll`, while Cygwin has just `cygunistring-2.dll` (?), so let's ~~stick with older version of this package (0.21-1) for a while~~ install the `libunistring5` package to fix the broken dependency.


ihhub (approved):

ihhub:
@oleg-derevenetz , that's a lot for this hot fix!

PR #6077, opened 2022-11-01
Fix catapult's boulder animation

### LINKED ISSUE #6075: Castle siege, catapult projectile misses spinning animation
### Preliminary checks

### Platform

Windows

### Describe the bug

image
image

Currently rock sprite moves over the battlefield being static.
In the original game projectile spins during the whole flight.

### Save file

Could be reproduced at any castle siege.

### Additional info

_No response_

ihhub:
close #6075

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3964,7 +3964,6 @@ void Battle::Interface::RedrawActionCatapult( int target, bool hit )
 {
     LocalEvent & le = LocalEvent::Get();
 
-    const fheroes2::Sprite & missile = fheroes2::AGG::GetICN( ICN::BOULDER, 0 );
     const fheroes2::Rect & area = GetArea();
 
     AudioManager::PlaySound( M82::CATSND00 );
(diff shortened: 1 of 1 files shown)


Branikolog (approved):

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3991,9 +3990,13 @@ void Battle::Interface::RedrawActionCatapult( int target, bool hit )
     pt2.y += area.y;
     max.y += area.y;
 
-    const std::vector<fheroes2::Point> points = GetArcPoints( pt1, pt2, max, missile.width() );
+    const fheroes2::Sprite & boulderFirstFrame = fheroes2::AGG::GetICN( ICN::BOULDER, 0 );
+    const std::vector<fheroes2::Point> points = GetArcPoints( pt1, pt2, max, boulderFirstFrame.width() );
     std::vector<fheroes2::Point>::const_iterator pnt = points.begin();
 
+    uint32_t boulderFrameId = 0;
+    const uint32_t boulderMaxFrame = fheroes2::AGG::GetICNCount( ICN::BOULDER );
```
oleg-derevenetz, line 3998: I would like just to point out that `...MaxFrame` might give an impression that this frame is included in the set of allowed frame id values, although from the rest of the code looks like this is not really the case :) Maybe `boulderFramesCount` is better? LGTM in general.

oleg-derevenetz (approved):

src/fheroes2/battle/battle_interface.cpp
ihhub: I can't argue with this. I will update the pull request.

ihhub: Fixed.

PR #6080, opened 2022-11-01
Rework the troops arrangement mechanics for the castle defense

### LINKED ISSUE #6076: Castle siege, army arrangement containing several stacks of the same type is not retained
### Preliminary checks

### Platform

Windows

### Describe the bug

In a case defending hero has several stacks of the same type creature and a few other creatures in castle garrison, those same type creatures are combined into one stack and creatures from castle garrison are merged into hero army.
image
image

This is very annoying, when you want some of your creatures to avoid siege, but they are nevertheless join battle.
So, I propose to transfer strongest creatures from garrison to hero army only in cases, when hero has some spare slots in army. In a case hero has all 5 slots occupied with creatures (even if it would be 5 stacks of 1 goblin) nothing should be merged from garrison to hero army, unless there're goblins left in garrison.

### Save file

[No_Name_3_0001.zip](https://github.com/ihhub/fheroes2/files/9910670/No_Name_3_0001.zip)

### Additional info

_No response_

oleg-derevenetz:
close #6076

Also fix the `Army::ArrangeForWhirlpool()`: it seems that "weakest stack" for whirlpool should be selected based on the unit's battle value, not HP.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1676,7 +1676,7 @@ namespace AI
 
         // TODO: do not transfer the whole army from one hero to another. Add logic to leave a fast unit for Scout and Courier. Also 3-5 monsters are better than
         // having 1 Peasant in one stack which leads to an instant death if the hero is attacked by an opponent.
-        taker.GetArmy().JoinStrongestFromArmy( giver.GetArmy(), false );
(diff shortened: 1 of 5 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for improving this part of the logic!

PR #6087, opened 2022-11-04
Draw the correct arrows for the paths on the adventure map

### LINKED ISSUE #6085: Adventure map, wrong arrows for the hero's path
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, there are wrong arrows for the hero's path.

The arrow used should be representative of the terrain penalty (for the targeted square) taking into account the hero's Pathfinding secondary skill, without taking into account the hero's oblique movement penalty.

In fact, the arrow should correspond to the penalty value displayed in the right-click information window for each square of the adventure map (example: "penalty: 150").

In original game:

It seems that the second arrow from the hero is not correct in the original game:
Arrow path 001

In fheroes2:

The third and sixth arrows from the hero are not correct:
Arrow path 002

In original game:

Arrow path 003

In fheroes2:

The fourth arrow from the hero is not correct:
Arrow path 004

Note: I noticed other problems with the arrows of the hero's path especially on the sea, I will complete this later.

### Save file

Here are the save files:
[Arrow path 50.zip](https://github.com/ihhub/fheroes2/files/9937243/Arrow.path.50.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #6085

Screenshot 1

Screenshot 2

This was the regression after the #5692 - while reviewing, I forgot that when leaving the road, we still need to show a short "road arrow" even though the road penalty is not actually applied (it is applied only when moving from a road tile to a road tile), and I didn't notice that `tileTo` should be replaced by `tileFrom` in one more place (in fact, due to the "road issue" mentioned earlier, we don't need to use the `tileTo` at all).

### DIFF
--- a/src/fheroes2/gui/interface_gamearea.cpp
+++ b/src/fheroes2/gui/interface_gamearea.cpp
@@ -705,16 +705,8 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
 
             uint32_t routeSpriteIndex = 0;
             if ( nextStep != path.end() ) {
-                const Maps::Tiles & tileFrom = world.GetTiles( from );
-                const Maps::Tiles & tileTo = world.GetTiles( nextStep->GetIndex() );
-                uint32_t cost = 0;
-
-                if ( tileFrom.isRoad() && tileTo.isRoad() ) {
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:

Issue #5466, opened 2022-05-26
Right-click via Touchscreen input not working on devices with Windows 10


admmlk:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

It should be possible to 'simulate' mouse right-click event with touchscreen input by touch-and-hold / double-hold / hold-and-drag - this is the way it works in other applications - however neither of it works in case using touchscreen input on Windows 10.

It makes the game "unplayable" (not able to inspect adventure map objects, hero/unit details etc.) on mobile devices with Windows.

Details:
fHeroes2 version: any latest (0.9.14 or lower), SDL2
System: Windows 10
Device: Asus Transformer T101HA

### Save file

Save file not needed for that issue.

### Additional info

Not sure if that's a bug or improvement, please change the type accordingly.

oleg-derevenetz:
Marking this as an improvement, because in its current state this game is designed to use either the mouse or game controller. Changes in the game control design may be required.

zenseii:
I should be able to look into this as I have a Win10 device with touch input support.

oleg-derevenetz:
Hi @zenseii BTW with initiatives like #5435 it may become even harder.

zenseii:
> Hi @zenseii BTW with initiatives like #5435 it may become even harder.

Yes. I believe they are talking about having a threshold for the amount of movement needed for the camera to move. Maybe @Laserlicht could share some thoughts about this?

Nevertheless, this issue is quite relevant for the future planned port to smartphones.

oleg-derevenetz:
> Nevertheless, this issue is quite relevant for the future planned port to smartphones, and would need to be addressed in time for that regard.

I have the impression that playing such strategies on smartphones will be inconvenient in any case - for example, I can't imagine how to simulate the adventure cursor on the touchscreen with the display of the number of days on the road when hovering over the tile. To be comfortable on smartphones, the game must be specially designed for them - with a seriously simplified gameplay, and HoMM series of the game (at least HoMM2/3) definitely aren't.

admmlk:
[...]

PR #6092, opened 2022-11-05
Add mouse right click simulation on touchscreen devices

### LINKED ISSUE #5778: Mouse position in a resized window is incorrect for SDL 2
### Preliminary checks

### Platform

Windows

### Describe the bug

SDL2 supports window resizing which works perfectly fine. However, some players pointed out that mouse cursor is too small which is true since we set it on an OS level. However, when a game is running in a windowed mode with a small resolution set and being resized on a big OS resolution, the mouse cursor is really small as it doesn't scale.

A solution for this is to use software mouse rendering which is slower but works fine. Besides this we have support of gamepads which is now enabled only for portable devices. So software rendering makes sense.

However, when a window is resized with software rendering mode enabled mouse position is not reflected properly.

### Save file

None.

### Additional info

_No response_

ihhub:
- add mouse right click simulation on touchscreen devices (second finger) (relates to #6008, relates to #5466)
- display mouse cursor on Android devices


oleg-derevenetz:
I got the impression that it would be difficult to distinguish "left press/click" from "right press/click". If first touch event arrives a bit earlier, then we already have a "left press/click", and if both events arrive at the same time, then where should be the cursor - under the first finger or under the second finger?

ihhub:
> 

Hi @oleg-derevenetz , yes, if a user presses two fingers at the same time then it's impossible to know which one. The idea is to press one finger first and then another. I haven't found any better way to do this as of now.

oleg-derevenetz:
@ihhub I was not able to get it to work this way yet :) For example, when I put my first finger on the "NEW GAME" item in the main menu, and then put my second finger somewhere else, I don't see the help dialog. When I release both fingers, I usually get the "New game" menu. But, when I put both fingers on screen at the same time, I usually get that hep dialog, but sometimes it is related to a thing under my first finger and sometimes - to a thing under my second finger :)

ihhub:
Hi @oleg-derevenetz , the idea is different. I put my first finger outside the rendering area (black area). Then the second finger on the item I want to left click. If my second finger blocks the view I still hold it and move it away keeping the popup dialog active.

oleg-derevenetz:
@ihhub Ah yes, this works :) But this definitely needs some guide, because this mechanics is not obvious (although undoubtedly witty).  Also we need to always maintain that black area somewhere.

src/engine/localevent.cpp
```diff
@@ -1272,34 +1277,46 @@ void LocalEvent::OnActiveEvent( const SDL_Event & event )
...
             = static_cast<double>( screenResolution.height * event.y - windowRect.y ) * ( static_cast<double>( gameSurfaceRes.height ) / windowRect.height );
 
         mouse_cu.x = static_cast<int32_t>( _emulatedPointerPosX );
         mouse_cu.y = static_cast<int32_t>( _emulatedPointerPosY );
+    }
+
+    if ( isFirstFinger ) {
+        SetModes( MOUSE_MOTION );
 
         if ( ( modes & MOUSE_MOTION ) && redraw_cursor_func ) {
```
oleg-derevenetz, line 1321: The `modes & MOUSE_MOTION` check seems redundant here, because we just set this mode a few lines above.

src/engine/localevent.cpp
```diff
@@ -1320,6 +1337,31 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
 
         mouse_button = SDL_BUTTON_LEFT;
     }
+    else if ( isSecondFinger ) {
+        if ( _numTouches < 2 ) {
+            // Only the second finger is pressing.
+            SetModes( MOUSE_MOTION );
+
+            if ( ( modes & MOUSE_MOTION ) && redraw_cursor_func ) {
```
oleg-derevenetz, line 1345: The `modes & MOUSE_MOTION` check seems redundant here, because we just set this mode a few lines above.

It seems that the same extra check is in the `LocalEvent::ProcessControllerAxisMotion()` as well - we set the `MOUSE_MOTION` and then check for it a few lines below, although it does not seem that this flag could have been reset between these lines.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions, please take a look when possible.

oleg-derevenetz:
Also I have an impression that cursor is shifted in regard to the touch point, and the farther the touch point is from the upper left edge of the screen (0:0), the greater the shift.

Laserlicht:
> Hi @oleg-derevenetz , the idea is different. I put my first finger outside the rendering area (black area). Then the second finger on the item I want to left click. If my second finger blocks the view I still hold it and move it away keeping the popup dialog active.

I would find the other solution better. So touch the relevant element and then touch anything with the second finger to open the window. This is how it is implemented in VCMI. Works quite well.

Because the "outside the rendering area" does not always exist. If you take a suitable resolution (aspect ratio), there is no black area. Besides, this is not intiutive.

But can imagine that this is more difficult to implement.

ihhub:
As I see multiple concerns with the proposed approach I will create another pull request to address mouse positioning with software rendering mode as touchscreen is a separate topic.

oleg-derevenetz:
Hi @ihhub I believe we can use the "right mouse button" part of this PR in the new release (but not the part with software-rendered mouse cursor, because it has objective issues). While "right mouse button" logic here is definitely not obvious, it at least works and can be used until a new logic is invented.

ihhub:
> Hi @ihhub I believe we can use the "right mouse button" part of this PR in the new release (but not the part with software-rendered mouse cursor, because it has objective issues). While "right mouse button" logic here is definitely not obvious, it at least works and can be used until a new logic is invented.

We can do this. I will update this PR today.

ihhub:
@oleg-derevenetz , now I spot another bug: the application crashes when I try to switch into fullscreen mode on an Android phone :(

oleg-derevenetz:
@ihhub

> now I spot another bug: the application crashes when I try to switch into fullscreen mode on an Android phone :(

I never tried to switch fullscreen mode on Android dynamically, only by changing the config file. Did this work at all at least once?

zenseii:
> @ihhub
> 
> > now I spot another bug: the application crashes when I try to switch into fullscreen mode on an Android phone :(
> 
> I never tried to switch fullscreen mode on Android dynamically, only by changing the config file. Did this work at all at least once?

I did it today with a snapshot build from master from today and did not experience this. I'll find out the exact commit I grabbed.

ihhub:
@oleg-derevenetz and @zenseii , the crash happens with my changes. I tested master branch and it works fine. I shall create a separate PR for `__linux__` macro fix.

ihhub:
> @ihhub
> 
> > now I spot another bug: the application crashes when I try to switch into fullscreen mode on an Android phone :(
> 
> I never tried to switch fullscreen mode on Android dynamically, only by changing the config file. Did this work at all at least once?

Don't change config file. Use our new fancy UI option within the game!😀

oleg-derevenetz:
@ihhub

> Don't change config file. Use our new fancy UI option within the game!😀

When I tried the fullscreen mode on Android, there wasn't any fancy option yet :)

oleg-derevenetz:
To be honest, after a cursory glance, I have no idea how this code can cause a crash after switching to the fullscreen.

DmitrySokolov:
> I would find the other solution better. So touch the relevant element and then touch anything with the second finger to open the window. This is how it is implemented in VCMI. Works quite well.

Have you considered just a long tap as rmb?

zenseii:
> > I would find the other solution better. So touch the relevant element and then touch anything with the second finger to open the window. This is how it is implemented in VCMI. Works quite well.
> 
> Have you considered just a long tap as rmb?

We have considered this and it would be very annoying considering how often you sometimes do this in this game.

zenseii (approved):
Tested on my Android phone and it works very well. Great job, @ihhub !

src/engine/localevent.cpp
```diff
@@ -41,7 +41,8 @@
 #include <SDL_gamecontroller.h>
 #include <SDL_keycode.h>
 
-#if defined( TARGET_PS_VITA ) || defined( TARGET_NINTENDO_SWITCH )
+#if defined( TARGET_PS_VITA ) || defined( TARGET_NINTENDO_SWITCH ) || defined( ANDROID )
+#define TOUCHSCREEN_SUPPORT
```
oleg-derevenetz, line 45: Hi @ihhib maybe call it something like `TOUCH_SUPPORT`, because the relevant code relates not just to devices with touchscreens, but also to devices with touchpads?

ihhub: Touchpads is actually a mouse but in a different form. Touchpads control mouse position. Please correct me in this if I am wrong.

oleg-derevenetz: @ihhub Not quite. For example, touchpad on my macbook has a concept of gestures using multiple fingers: slide by two fingers is scrolling event, pinching by two fingers is zoom in/out, [and so on](https://support.apple.com/en-us/HT204895).

ihhub: Okay, in this case we can just rename it to `TOUCH_SUPPORT` to avoid confusion.

oleg-derevenetz (approved):

PR #6094, opened 2022-11-05
Fix possible assertion hit for campaign scenario info dialog

### LINKED ISSUE #6091: Assertion on campaign scenario loading
### Preliminary checks

### Platform

Windows

### Describe the bug

image
After pressing "ignore" button briefing page appears, but campaign couldn't be started.
image

### Save file

[Good_9_Complete.zip](https://github.com/ihhub/fheroes2/files/9943789/Good_9_Complete.zip)

### Additional info

_No response_

ihhub:
close #6091

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -1182,10 +1182,6 @@ fheroes2::GameMode Game::SelectCampaignScenario( const fheroes2::GameMode prevMo
 
     const bool isDifficultySelectionAllowed = campaignSaveData.isStarting() && !allowToRestart;
 
-    if ( !scenario.isMapFilePresent() ) {
-        buttonOk.disable();
-    }
-
     if ( allowToRestart ) {
         buttonOk.disable();
         buttonOk.hide();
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

Branikolog:
Hi, @ihhub !
After skipping of the assertion, previously we could see inactive OKAY button, which indicated that you cannot start mission.
image

Within current PR the assertion is no more occurs, but OKAY button looks as normal. Nothing happens, when we press it, so it looks like a bug.
image

Branikolog:
@ihhub 
Now only one assertion is left, when we enter info window, initiate restart and then return back to map.

Branikolog:
@ihhub 
Still have an assertion after trying to restart scenario and returning back to the adventure map. :(

ihhub:
Hi @Branikolog , I reverted changes to the original logic for disabling buttons. The assertion happens because we are trying to load a map, clearing all information about the current game. If it fails we do not restore the state of the game back. So far there is no proper solution for this so disabling buttons is the only way to prevent crashes.

Branikolog (approved):

PR #6099, opened 2022-11-05
Remove redundant PS Vita-specific code from the Settings::Save()

oleg-derevenetz:


ihhub:

PR #6100, opened 2022-11-05
Reset the MOUSE_WHEEL in LocalEvent::GetClean() and LocalEvent::HandleEvents(), add a few comments regarding event modes

oleg-derevenetz:


ihhub:

PR #6102, opened 2022-11-06
Fix cursor software rendering mode on SDL 2 builds

### LINKED ISSUE #5419: Cursor too small
### Preliminary checks

I checked for issues related to cursors and didn't find this one.  I am so glad to be playing HOMM2 again.  However, I can barely see the cursor.  It doesn't matter if I fullscreen or minimize the screen, the cursor stays the same small size.  Is there a way to increase the size for more enjoyable play?  Here are two pictures.  It doesn't matter the resolution, the cursor is the same size.  I placed the cursor over a dark part to allow it to show up, but it is lost if anywhere else.  It is on the "High Score" sign.  The first is windowed, the next is maximized.

I would like to have an option to scale the cursor size.

### Additional info

_No response_

### LINKED ISSUE #5778: Mouse position in a resized window is incorrect for SDL 2

ihhub:
close #5419
close #5778


oleg-derevenetz:
Hi @ihhub there is something definitely wrong with this software cursor (see my [previous comment](https://github.com/ihhub/fheroes2/pull/6092#issuecomment-1304614069)). Here is a video from the Android simulator:

https://user-images.githubusercontent.com/32623900/200170778-a3f61fa7-2b1a-42cc-b56d-54c911f25ee9.mp4

I suppose you should be able to easily reproduce this in simulator too.

ihhub:
Hi @oleg-derevenetz , I actually forgot that in the previous pull request I set Android to use touchscreen. This fix works perfectly fine on Windows. I assume that it works as well on MacOS. We should add extra conditions for this position calculation.

oleg-derevenetz:
Hi @ihhub in the previous pull request the position of the cursor was shifted as well in regard to the touch position. I tried this on real device and there was a shift, but not as large as on video.

ihhub:
> Hi @ihhub in the previous pull request the position of the cursor was shifted as well in regard to the touch position. I tried this on real device and there was a shift, but not as large as on video.

No-no, I meant the fix is for non-Android devices, normal desktops and laptops with Windows OS, Linux OS or macOS. Can you please on your macOS?

oleg-derevenetz:
@ihhub

> No-no, I meant the fix is for non-Android devices, normal desktops and laptops with Windows OS, Linux OS or macOS. Can you please on your macOS?

Yes, sure, I'll check this once I finish cleaning my house :) But why Android is "special" in this regard?

ihhub:
> @ihhub
> 
> > No-no, I meant the fix is for non-Android devices, normal desktops and laptops with Windows OS, Linux OS or macOS. Can you please on your macOS?
> 
> Yes, sure, I'll check this once I finish cleaning my house :) But why Android is "special" in this regard?

I tried and it looks like Android handles it properly or I miss something in the implementation. I will take a look later.

oleg-derevenetz:
@ihhub

> I tried and it looks like Android handles it properly or I miss something in the implementation. I will take a look later.

I just added this line:

```cpp
fheroes2::cursor().enableSoftwareEmulation( true );
```

in the `Settings::Read()` before the `BinaryLoad()` call and got this weird cursor behavior. Without this line, the touch events work just fine in right places (but there is no cursor of course).

oleg-derevenetz:
Hi @ihhub

> Can you please on your macOS?

I checked the software cursor on macOS and there are issues as well. For example, both in fullscreen and in windowed mode with game resolution set to 2560x1600 software cursor is not able to move to the right or lower than some point approximately in the center of the screen, while "hardware cursor" has no issues with this at all. And it looks like in 640x480 fullscreen software cursor is able to move much lower than the lower edge of the screen (or much to the right than the right edge of the screen), while it is not able to move to the left or upper than 0:0.

oleg-derevenetz:
BTW that's how game screen looks on macOS when built with `sdl1-compat` and attempting to dynamically switch between windowed and fullscreen mode:

sdl1-compat

No so much "compat". Also when switching back to windowed mode the window contents is corrupted as well and game window still doesn't want to release the mouse cursor (I mean, the cursor still cannot go beyond the game window).

ihhub:
@oleg-derevenetz , that's a progress! However, SDL 1 support is useless on macOS :(

oleg-derevenetz:
@ihhub

> However, SDL 1 support is useless on macOS :(

That's not the "true" SDL1 anyway :) That's just the compatibility layer.

ihhub:
Hi @oleg-derevenetz and @zenseii , I overcomplicated things in the first attempt. Once we set rendering logical size all mouse events are relative to the renderer size so we don't need to rescale mouse position.

ihhub:
@oleg-derevenetz and @zenseii , it is ready for your review.

oleg-derevenetz (approved):
Hi @ihhub I just tested this PR on macOS in 2880x1800 and now it works just fine. The only thing I noticed is that the mouse cursor is much smaller than the same mouse cursor in the hardware mode, is this intentional? If yes, then LGTM.

ihhub:
> Hi @ihhub I just tested this PR on macOS in 2880x1800 and now it works just fine. The only thing I noticed is that the mouse cursor is much smaller than the same mouse cursor in the hardware mode, is this intentional? If yes, then LGTM.

The cursor is proportionally resized based on resolution. If you make 640x480 resolution in full screen you will be a huge cursor.

oleg-derevenetz:
@ihhub

> The cursor is proportionally resized based on resolution. If you make 640x480 resolution in full screen you will have a huge cursor.

OK, then I believe we can merge this PR.

zenseii (approved):

PR #6111, opened 2022-11-06
Add a separate graphics options window

ihhub:
The new window includes resolution, fullscreen / windowed modes (very useful for devices with no keyboard support, relates to #6008), v-sync option and system information. Besides this this pull request does:
- update fresh start window text: the settings can be changed by pressing a button and remove F4 mentioning as it is not true for all devices
- standartize all options within every dialog: we were rendering text differently for some dialogs
- simplify code for popup windows

**Note**: names and sentences are subject of changes so feel free to proposed better ones.

image

### DIFF
--- a/src/fheroes2/dialog/dialog_graphics_settings.cpp
+++ b/src/fheroes2/dialog/dialog_graphics_settings.cpp
@@ -172,7 +172,7 @@ namespace
                 fheroes2::showStandardTextMessage( _( "V-Sync" ), _( "Enabled V-Sync option is useful for some monitors to resolve flickering issues." ), 0 );
             }
             if ( le.MousePressRight( windowSystemInfoRoi ) ) {
-                fheroes2::showStandardTextMessage( _( "System Info" ), _( "Show extra information such as FPS and current date and time." ), 0 );
+                fheroes2::showStandardTextMessage( _( "System Info" ), _( "Show extra information such as FPS and current time." ), 0 );
             }
         }
 
--- a/src/fheroes2/dialog/dialog_graphics_settings.cpp
+++ b/src/fheroes2/dialog/dialog_graphics_settings.cpp
@@ -18,8 +18,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-#include "agg_image.h"
 #include "dialog_graphics_settings.h"
+#include "agg_image.h"
 #include "dialog_resolution.h"
 #include "game_hotkeys.h"
 #include "game_mainmenu_ui.h"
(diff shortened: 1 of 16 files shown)


oleg-derevenetz (approved):

Branikolog:
@ihhub , hello!
Everythng seems to work fine. Great improvement!
I can only notice, that new graphics window and sound settings are missing yellow header (name of the option you're browsing currently).
image
image
Like we have for other windows for resolutions or hotkeys:
image
Also, I'm concerning about removing the mentioning of F4 button as a default button for switching windowed modes. I'm afraid people would never know it's possible to switch the mode outside of this window... We can mention at least on right click description that default hotkey for this option is F4. But it's up to you.

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            if ( le.MousePressRight( windowResolutionRoi ) ) {
+                fheroes2::showStandardTextMessage( _( "Select Game Resolution" ), _( "Change the resolution of the game." ), 0 );
+            }
+            else if ( le.MousePressRight( windowModeRoi ) ) {
+                fheroes2::showStandardTextMessage( _( "window|Mode" ), _( "Toggle between fullscreen and windowed modes." ), 0 );
+            }
+            else if ( le.MousePressRight( windowVSyncRoi ) ) {
+                fheroes2::showStandardTextMessage( _( "V-Sync" ), _( "Enabled V-Sync option is useful for some monitors to resolve flickering issues." ), 0 );
```
zenseii, line 172: What do you think about phrasing it like this? I'm not sure about the line length of the actual code here and if the code style check will fail.

```suggestion
                fheroes2::showStandardTextMessage( _( "V-Sync" ), _( "The V-Sync option can be enabled  to resolve flickering issues on some monitors." ), 0 );
```

Branikolog:
@ihhub 
We currently have resolution, hot keys and experimental settings with yellow headers within this menu, which is more than 50%. :)
Graphics and audio are missing text within their windows.
I feel these both windows should have titles, as they look out of game overall style. But we can do this within a separate PR.

zenseii (commented):
Hi, @ihhub.

I only found one comment about the wording. As I said I would also like to test one of the snapshot builds, but the code looks good to me.

Also are we enabling full screen mode by default?

ihhub:
Hi @zenseii , no, we do not enabling fullscreen mode by default. However, on some devices we should.

src/fheroes2/dialog/dialog_graphics_settings.cpp
ihhub: Yup, looks better.

Branikolog:
@ihhub 
There's missing description for OKAY button in this window, which we have for other similar windows:
image

ihhub:
Hi @Branikolog and @zenseii , I addressed your comments. Please have a look again.

zenseii (approved):
LGTM. Looked at it in game too and no problem there.

PR #6126, opened 2022-11-08
Fix incorrect evaluation of army strength where hero's abilities should not be taken into account

### LINKED ISSUE #6122: Hero strength shouldn't affect the aggressiveness of neutral creatures
### Preliminary checks

### Platform

Windows

### Describe the bug

Attacking a Cyclops using high level hero causes him to run away:
image
While using a 1 lvl hero in the same conditions leads to a battle:
image
I believe, neutral creatures couldn't know if a hero has powerful spells or high stats. So they should be more aggressive.
Current logic of fheroes2 causes such ridiculous cases to happen, when 44 Cyclops could join a strong hero with only 3 dragons in his army.

### Save file

[Strong-weak hero.zip](https://github.com/ihhub/fheroes2/files/9960899/Strong-weak.hero.zip)

### Additional info

_No response_

ihhub:
close #6122

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -1504,7 +1504,7 @@ NeutralMonsterJoiningCondition Army::GetJoinSolution( const Heroes & hero, const
     }
 
     // Neutral monsters don't care about hero's stats. Ignoring hero's stats makes hero's army strength be smaller in eyes of neutrals and they won't join so often.
-    const double armyStrengthRatio = hero.GetArmy().getTroops().GetStrength() / troop.GetStrength();
+    const double armyStrengthRatio = hero.GetArmy().Troops::GetStrength() / troop.GetStrength();
 
     // The ability to accept monsters (a free slot or a stack of monsters of the same type) is a
     // mandatory condition for their joining in accordance with the mechanics of the original game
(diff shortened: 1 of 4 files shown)


src/fheroes2/army/army.cpp
```diff
@@ -1504,7 +1504,7 @@ NeutralMonsterJoiningCondition Army::GetJoinSolution( const Heroes & hero, const
     }
 
     // Neutral monsters don't care about hero's stats. Ignoring hero's stats makes hero's army strength be smaller in eyes of neutrals and they won't join so often.
-    const double armyStrengthRatio = static_cast<const Troops &>( hero.GetArmy() ).GetStrength() / troop.GetStrength();
+    const double armyStrengthRatio = hero.GetArmy().getTroops().GetStrength() / troop.GetStrength();
```
oleg-derevenetz, line 1507: Hi @ihhub are you sure that this will work? `Army::getTroops()` returns a reference, and `Troops::GetStrength()` is `virtual`, so `Army::GetStrength()` [will be called anyway](https://godbolt.org/z/rTqrcWqGY). The proper fix should be:

```cpp
hero.GetArmy().Troops::GetStrength()
```

ihhub: You're right! We also recently made a similar change for army estimation to be left in a castle and I think it was wrong.

Branikolog:
Hi, @ihhub !
There's still an issue. Neutral creature attacks the weaker hero of two with the same army.

ihhub:
> Hi, @ihhub ! There's still an issue. Neutral creature attacks the weaker hero of two with the same army.

Hi @Branikolog , how is it an issue?

Branikolog:
> Hi @Branikolog , how is it an issue?

I probably not clearly described the issue.
We have two heroes with the same army but with different levels/stats. We attack the same neutral creature with both of them. In one case creature flees, in second - attacks hero.)

Branikolog (approved):

oleg-derevenetz (approved):

PR #6130, opened 2022-11-08
Adventure map pathfinding: minor optimizations

oleg-derevenetz:
~~The idea is to calculate costs using the method in which the previous steps have more weight than the subsequent ones. However, this method has limitations, since the order of magnitude of costs increases significantly (at least when working with integers, because in this case the minimum coefficient for the previous steps (without rounding errors) is 2, so we have the 2^N coefficient for the first step after the Nth step). Using the `uint64_t` to represent the progressive costs, this method may be applied to at least a few dozens of initial steps in the path (for subsequent steps the regular summing algorithm will be applied to avoid the unsigned overflow), which should be enough for real applications, because on the next day hero paths will be recalculated anyway.~~

This approach doesn't work properly on larger maps, less optimal paths are generated. So I just leave various minor optimizations and added comments.

### DIFF
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -195,6 +195,20 @@ namespace
 
         return false;
     }
+
+    uint64_t addToProgressiveCost( uint64_t progressiveCost, const uint32_t additionalCost )
+    {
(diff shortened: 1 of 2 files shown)


ihhub:
@oleg-derevenetz , thank you very much for these optimizations!

PR #6134, opened 2022-11-09
Fix monster upgrade cost formula

### LINKED ISSUE #1322: Creature upgrade cost
In fheroes2, the cost to upgrade a creature is once the price difference between the two creatures.

I love this change from the original version of the game that sometimes makes the upgrade cost incoherent.
The best example is the cost of upgrading the giant to titan which is higher than the cost of purchasing the titan, a game design mistake for me.
Also this is necessary in fheroes2, as it is possible to buy unimproved creatures when the upgraded creature building is built, which is a very good addition to the original game and allows for new strategies.

However, in the original game, the cost to upgrade a creature is twice the price difference between the two creatures.
To enable the original game's operating mode, it is therefore necessary to add an option to the game:
"Game: The cost to upgrade a creature is only once the price difference between the two creatures."
And this option should be unchecked by default.

ihhub:
close #1322

I foresee some debates about this but this is how it works in the OG.

### DIFF
--- a/src/fheroes2/monster/monster.cpp
+++ b/src/fheroes2/monster/monster.cpp
@@ -823,7 +823,7 @@ int Monster::ICNMonh() const
 payment_t Monster::GetUpgradeCost() const
 {
     const Monster upgr = GetUpgrade();
-    const payment_t pay = id != upgr.id ? upgr.GetCost() - GetCost() : GetCost();
+    const payment_t pay = ( id != upgr.id ) ? ( upgr.GetCost() - GetCost() ) * 2 : GetCost();
 
     return pay;
 }


Alucard648:
This should be uptional thing as some factions, like Knight and Wizard auffer from this a lot.

zenseii:
> This should be uptional thing as some factions, like Knight and Wizard auffer from this a lot.

I believe that Wizard and Warlock suffering from this is no problem. In my opinion, to decide if we should deviate from the original intent we need lots of documented statistics to back this up. Until then my recommendation is we stick to the original, but keep the doors open for improvement.

a1exsh:
@ihhub can we plee-a-a-ase add an experimental option to keep the way it works currently (a-la HoMM III) as well? :)

ihhub:
> @ihhub can we plee-a-a-ase add an experimental option to keep the way it works currently (a-la HoMM III) as well? :)

Hi @a1exsh , to add a new option we need to follow this guide. Also experimental options dialog will cease to exist by 1.0 release.

Branikolog (approved):
Everything works as intended.
We should as much stick to original behaviour, as we can. 
It's a simple rule for upgrading so everyone know, how it is working. This definitely makes game more tough, but who said, that Heroes 2 is an easy game? Let's leave such gameplay-breaking options to be adjusted in future updates.

oleg-derevenetz (approved):

PR #6143, opened 2022-11-10
Version 0.9.21

ihhub:
Final steps for the major release

PR #6133, opened 2022-11-09
Drop support for save file formats older than 0.9.19

ihhub:
In preparations for 1.0 release.

PR #6144, opened 2022-11-10
Use IWYU to tidy up the inclusion of header files and forward declarations

oleg-derevenetz:
Currently, there is a big mess with this, because header files are included only when build fails and never removed, even if they are not needed anymore. Also in many cases where a header file is included, it would be more appropriate to just use a forward declaration. In this PR I propose to make our CI a bit more industrial-grade and use the [IWYU](https://github.com/include-what-you-use/include-what-you-use) that will care of all this automatically. It follows a good habits to explicitly include header files for all constructs used. It also has a good diagnostics that explains why a specific header file should be included, like this:

```
Warning: include-what-you-use reported diagnostics:

(/home/runner/work/fheroes2/fheroes2/src/fheroes2/campaign/campaign_data.h has correct #includes/fwd-decls)

/home/runner/work/fheroes2/fheroes2/src/fheroes2/campaign/campaign_data.cpp should add these lines:
#include "color.h"            // for Colors

/home/runner/work/fheroes2/fheroes2/src/fheroes2/campaign/campaign_data.cpp should remove these lines:

The full include-list for /home/runner/work/fheroes2/fheroes2/src/fheroes2/campaign/campaign_data.cpp:
#include "campaign_data.h"
#include <algorithm>          // for max, find
#include <array>              // for array
#include <cassert>            // for assert
#include <cstddef>            // for size_t
#include <memory>             // for allocator_traits<>::value_type
#include <utility>            // for move
#include "artifact.h"         // for Artifact, Artifact::ULTIMATE_CROWN, Art...
#include "color.h"            // for Colors
#include "game_video_type.h"  // for VideoAction, VideoAction::IGNORE_VIDEO
#include "heroes.h"           // for Heroes, Heroes::BAX, Heroes::CORLAGON
#include "maps_fileinfo.h"    // for FileInfo
#include "monster.h"          // for Monster, Monster::DWARF, Monster::ELF
#include "resource.h"         // for String, WOOD
#include "spell.h"            // for Spell, Spell::SETEGUARDIAN
#include "tools.h"            // for StringReplace
#include "translations.h"     // for gettext_noop, _
```

Sometimes there are false positives, but in general it works just fine.


oleg-derevenetz:
Hi @ihhub you can see a profit right now - unused headers are detected after #6133 :)

oleg-derevenetz:
Hi @ihhub if this PR is OK for you, it would be good to merge it before the others (perhaps with the exception of #6143, if you don't want this PR to go to the new release for some reason), because it adds a new check.

ihhub:
@oleg-derevenetz , oh... I saw your comment after merging other PRs :( I will check this PR now and then merge it.

oleg-derevenetz:
Hi @ihhub we will need to resolve conflicts first :) This is on the way.

ihhub:
> Hi @ihhub we will need to resolve conflicts first :) This is on the way.

My apologies for this.

oleg-derevenetz:
@ihhub

> My apologies for this.

Nevermind :) IWYU will do all the job.

.github/workflows/clang_tidy.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   tidy:
     name: Clang-Tidy
     runs-on: ubuntu-22.04
-    timeout-minutes: 30
+    timeout-minutes: 60
```
ihhub, line 16: Does clang-tidy need that long now?

src/engine/logging.h
```diff
@@ -21,8 +21,11 @@
 #ifndef H2LOGGING_H
 #define H2LOGGING_H
 
+// IWYU pragma: no_include <bits/types/struct_tm.h>
```
ihhub, line 24: What is the reason IWYU tries to include this header?

ihhub (commented):
Hi @oleg-derevenetz , I partially review this pull request. My browser keeps crashing while viewing this page so I will check the rest of changes later today in a proper machine.

.github/workflows/cmake.yml
```diff
@@ -10,27 +10,27 @@ jobs:
         config:
         - name: Linux SDL1
           os: ubuntu-latest
-          dependencies: libsdl1.2-dev libsdl-mixer1.2-dev libsdl-image1.2-dev gettext
+          dependencies: libsdl1.2-dev libsdl-mixer1.2-dev libsdl-image1.2-dev gettext ninja-build
```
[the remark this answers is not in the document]
oleg-derevenetz: Yes, of course. Ninja just faster than make.

.github/workflows/clang_tidy.yml
oleg-derevenetz: For this PR - yes :) Because too many files are changed. But not in general.

src/engine/logging.h
oleg-derevenetz: Sometimes in rare cases it tries to propose "internal" headers where corresponding symbols are in fact defined instead of "standard" ones. So I just added pragmas to ignore this so far. I hope this will be fixed in newer versions.

src/engine/logging.h
```diff
@@ -169,12 +172,10 @@ namespace Logging
...
     {
-        TextSupportLogger()
-        {
-            COUT( "----------" )
-        }
+        TextSupportLogger(){ COUT( "----------" ) }
 
-        TextSupportLogger( const TextSupportLogger & ) = delete;
+        TextSupportLogger( const TextSupportLogger & )
+            = delete;
```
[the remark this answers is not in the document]
oleg-derevenetz: Oddly enough, but yes. I just ran `clang-format` on this file by hand and the result was the same.

src/engine/logging.h
oleg-derevenetz: @ihhub I believe that this can be fixed by the mapping file (`iwyu.map`), but this still needs to be sorted out.

oleg-derevenetz: @ihhub I just tried to use the mapping file to solve this issue with "internal" headers and it seems that it works.

src/engine/system.cpp
```diff
@@ -24,19 +24,32 @@
...
+#endif
+
+#include <SDL_version.h>
+
+#if SDL_VERSION_ATLEAST( 2, 0, 0 ) && defined( ANDROID )
+#include <SDL_error.h>
+#include <SDL_system.h>
+#endif
+
+#if SDL_VERSION_ATLEAST( 2, 0, 1 ) && !defined( __linux__ )
```
ihhub, line 49: Is 2.0.1 version proposed by IWYU?

src/fheroes2/game/fheroes2.cpp
```diff
@@ -21,18 +21,30 @@
...
+#include <exception>
 #include <iostream>
+#include <list>
+#include <memory>
+#include <set>
 #include <string>
 
-#include <SDL.h>
+#include <SDL_events.h>
+#include <SDL_main.h> // IWYU pragma: keep
```
ihhub, line 33: I assume that this comment is to keep this header, isn't?

src/engine/system.cpp
oleg-derevenetz: No, it's because this header contains only functions that are available since 2.0.1.

src/fheroes2/game/fheroes2.cpp
oleg-derevenetz: Yes. This header redefines `main` using the preprocessor directive, IWYU apparently doesn't understand this.

ihhub:
@oleg-derevenetz , huge thanks for another automation step of our development!

ihhub:
Hi @oleg-derevenetz , it looks like `__linux__` broke Android build.

oleg-derevenetz:
@ihhub

>  it looks like `__linux__` broke Android build.

Hmm, it's probably because `__linux__` is set for Android too. Try to use something like this:

```
#if defined( __linux__ ) && !defined( ANDROID ) 
```

I suppose `__LINUX__` was set by some 3rd party header (maybe SDL?)

oleg-derevenetz:
@ihhub

Or maybe I'll take a look later today myself.

ihhub:
@oleg-derevenetz , yup, this, obviously, works. I will include the fix in my touchscreen related PR.

PR #6152, opened 2022-11-11
IWYU: add <cmath> as another alias for std::abs()'s internal header

oleg-derevenetz:
Both `#include <cmath>` and `#include <cstdlib>` [should be fine](https://en.cppreference.com/w/cpp/numeric/math/abs) for the `std::abs()`. Before this PR, there was only alias for the `<cstdlib>`.


ihhub:
@oleg-derevenetz , thank you very much for this small improvement!

PR #6158, opened 2022-11-12
Fix download links after 0.9.21

oleg-derevenetz:
File naming scheme was changed after #6040, now they are in release, so we need to update the download links.


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for the quick fix! I merged it since none of our pipeline jobs depend on documentation files.

oleg-derevenetz:
Hi @ihhub BTW, should we add release links for Android version or will it be somewhat premature?

ihhub:
@oleg-derevenetz , we should together with an instruction how to install the apk on a device. We want people to help with testing but we can specify that this is a POC version for now.

PR #6159, opened 2022-11-12
Allows to translate dialog text

srzkptrsk:
Minor update - allows to translate dialog text. It will be good to have this feature available

### DIFF
--- a/src/fheroes2/dialog/dialog_message.cpp
+++ b/src/fheroes2/dialog/dialog_message.cpp
@@ -34,6 +34,7 @@
 #include "screen.h"
 #include "text.h"
 #include "ui_button.h"
+#include "translations.h"
 
 namespace
 {
@@ -72,7 +73,7 @@ int Dialog::Message( const std::string & header, const std::string & message, in
     const CursorRestorer cursorRestorer( buttons != 0, Cursor::POINTER );
 
     TextBox textbox1( header, Font::YELLOW_BIG, BOXAREA_WIDTH );
-    TextBox textbox2( message, ft, BOXAREA_WIDTH );
+    TextBox textbox2( _(message), ft, BOXAREA_WIDTH );
 
     const int32_t headerHeight = !header.empty() ? textbox1.h() + 10 : 0;
     FrameBox box( 10 + headerHeight + textbox2.h(), buttons != 0 );


ihhub:
Hi @srzkptrsk , this change won't do anything as the engine has no idea what exactly to translate. What is your goal by doing this?

oleg-derevenetz:
Hi @srzkptrsk the GNU gettext engine works as follows: `xgettext` tool scans strings that are used as arguments for `_()`, `gettext()` and `ngettext()` calls and builds the table (basically `original text` -> translated text). If there is no original text to translate, there will be no entry in this table and `_()` will have no idea what to translate.

oleg-derevenetz:
@srzkptrsk Strings from map files cannot be translated using the `_()`, because there is no translation for them in the game's `*.mo` files, where they should come from?

ihhub:
I am closing this pull request based on explanation in [this comment](https://github.com/ihhub/fheroes2/issues/6188#issuecomment-1315261898).

dood-apo:
@oleg-derevenetz 

> If there is no original text to translate, there will be no entry in this table and `_()` will have no idea what to translate.

If there is no entry, `_()` will just return the original string (the missing key). I think that adding a possibility to put more translations into `*.po` files does not hurt? Unless we are violating some license by putting **only** quotes from the dialogs in the resource bundles?

> Strings from map files cannot be translated using the _(), because there is no translation for them in the game's *.mo files, where they should come from?

They would come from `*.mo` files after someone puts the effort to translate some messages in the `*.po` files :).
This does not require to translate everything at once, someone might want to translate only a particular map for a particular language, and this enables it a possibility.

oleg-derevenetz:
Hi @dood-apo this PR by itself doesn't offer the ability to "magically" translate something. We need to get all these string to translate in our .po files first, but currently they are auto-generated, so we need the infrastructure that will generate .po files from maps (licensing issues arise) and merge them with files generated from engine sources. In other words, we need some infrastructure, not just a single `_()` call, which is useless by itself.

dood-apo:
> this PR by itself doesn't offer the ability to "magically" translate something

Sure, but without `_()` call translating some dialogs is not even possible. So it enables further broadening translations, which shouldn't hurt I think?

> currently they are auto-generated

But `*.po` files are present in the repository? Or do you mean that a `.po` file for a newly added language is autogenerated (filled with real msgids but with empty msgstrs) to provide a starting point and later on gets committed to the repo once someone fills msgstrs?

oleg-derevenetz:
> So it enables further broadening translations, which shouldn't hurt I think?

Code that does nothing is not very useful.

> But `*.po` files are present in the repository?

Yes and no.

> Or do you mean that a `.po` file for a newly added language is autogenerated (filled with real msgids but with empty msgstrs) to provide a starting point and later on gets committed to the repo once someone fills msgstrs?

Not just for a newly added language. They are auto-generated after each commit into `master` that affects translated strings in the code. You've probably seen PR like #8320, right? There is a special [workflow](https://github.com/ihhub/fheroes2/blob/master/.github/workflows/translation_update.yml) that takes care of this.

dood-apo:
> Not just for a newly added language. They are auto-generated after each commit into master that affects translated strings in the code. You've probably seen PRs like https://github.com/ihhub/fheroes2/pull/8320, right? There is a special [workflow](https://github.com/ihhub/fheroes2/blob/master/.github/workflows/translation_update.yml) that takes care of this.

Gotcha. I guess I performed that workflow's job here then: https://github.com/ihhub/fheroes2/pull/8328

So the msgids are updated in the `*.po` files when someone changes them in the code. That's nice. But still for erratas and for translating maps people could add manually the entries (msgid+msgstr pairs) to `*.po` files and commit them, right? Their msgids would not be affected by the "mr fixit workflow" but would still work in runtime.

oleg-derevenetz:
> But still for erratas and for translating maps people could add manually the entries (msgid+msgstr pairs) to *.po files and commit them, right? Their msgids would not be affected by the "mr fixit workflow" but would still work in runtime.

Currently strings in .po files that are absent in the engine code are marked as _fuzzy_, so no, currently people can't just add arbitrary strings to these files, these strings will be immediately marked as fuzzy. As I said, this requires an infrastructure - some separate .po files that will be combined with engine .po files at the .mo files generation stage, something like this, I suppose.

dood-apo:
Thanks, now it is much more clear for me. We don't want manual entries and consciously protect against them. And since this is the case, adding `_()` won't enable any translators.

Issue #6161, opened 2022-11-12
Rework of experimental options


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Since there are always ongoing debates about whether to keep or remove certain experimental options we need to make a clear way how to deal with them.

### Describe the possible solution

I suggest to do these steps:
- [ ] add a configuration option to enable and disable all experimental options in the game. With the disabled state there is no way to change them or even load them with save files (controversial)
- [ ] save them in a separate file format which supports extension in the future
- [ ] handle issues with options being loaded from save files which update global state of these options within the game
- [ ] add a clear indicator that the game is being played with these options being enabled (an icon perhaps on the Adventure Map)
- [ ] propose a new dialog for these options to keep the same style of the game. Each option should follow the guide from [the wiki page](https://github.com/ihhub/fheroes2/wiki/Adding-a-new-option-into-the-game). We might stick to one set of sprites to turning on or off. Alternatively, we can use checkboxes but popup windows with a full description must be present. The page must contain a message telling that these options change the core gameplay and can lead to unexpected consequences.
- [ ] add an indicator that a save file contains enabled experimental options
- [ ] add an indicator that a high score was achieved with the enabled options
- [ ] add a question dialog for users which do not have these options enabled (global configuration flag) while they are loading a save file with the options
- [ ] add a way to show the options for campaign scenarios
- [ ] rename experimental options into something more game-style friendly

ihhub:
Hi @zenseii , @Branikolog and @oleg-derevenetz , I jotted down my notes about experimental options rework. If you have even better ideas please feel free to explain.

zenseii:
For reference here are all the current experimental options:
```
1. battle: show damage info
2. battle: deterministic events
3. battle: allow soft wait for troops

4. world: Scouting skill shows extended content info
5. world: allow to set guardian to objects
6. world: Windmills, Water Wheels and Magic Gardens can be captured

7. castle: allow guardians

8. heroes: allow buy a spellbook from Shrines
9. heroes: remember movement points when retreating or surrendering
10. heroes: allow to choose any primary skill in Arena

11. game: autosave will be made at the beginning of the day
12. game: offer to continue the game after victory condition
````

A bit off-topic, but I noticed these strings alongside the experimental ones when going through these in the PO file:
```
game: use evil interface
[...]

PR #6172, opened 2022-11-13
Make battle: deterministic events option by default

ihhub:
relates to #6161


src/fheroes2/battle/battle_main.cpp
```diff
@@ -233,11 +232,9 @@ Battle::Result Battle::Loader( Army & army1, Army & army2, int32_t mapsindex )
         }
     }
 
-    const uint32_t battleSeed = Settings::Get().ExtBattleDeterministicResult() ? computeBattleSeed( mapsindex, world.GetMapSeed(), army1, army2 )
-                                                                               : Rand::Get( std::numeric_limits<uint32_t>::max() );
-
     bool isBattleOver = false;
     while ( !isBattleOver ) {
+        const uint32_t battleSeed = computeBattleSeed( mapsindex, world.GetMapSeed(), army1, army2 );
```
oleg-derevenetz, line 237: Hi @ihhub there is really no need to re-compute the battle seed both for instant battle and manual battle, because none of the `computeBattleSeed()`'s arguments are changed. We can move it out of this loop, just like the `battleSeed` was computed before.

oleg-derevenetz (commented):
Hi @ihhub I just left one optional suggestion. LGTM otherwise.

src/fheroes2/battle/battle_main.cpp
ihhub: You're right. I misread as `if` body.

oleg-derevenetz (approved):

zenseii (approved):

PR #6176, opened 2022-11-13
Introduce AI Scout role

ihhub:
Also increase fog discovery value as it was very low and AI prefers to just stay in one place.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -131,6 +131,12 @@ namespace
             heroList.pop_back();
         }
 
+        if ( heroList.size() > 2 ) {
+            // Assign the role and remove them so they aren't counted towards the median strength
+            heroList.back().hero->setAIRole( Heroes::Role::SCOUT );
+            heroList.pop_back();
+        }
+
         assert( !heroList.empty() );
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -131,6 +131,12 @@ namespace
             heroList.pop_back();
         }
 
+        if ( heroList.size() > 2 ) {
+            // Assign the role and remove them so they aren't counted towards the median strength
+            heroList.back().hero->setAIRole( Heroes::Role::SCOUT );
```
oleg-derevenetz, line 136: So if we have exactly 3 heroes, we just assign scout role to the first one that came across without taking into account his strength?

oleg-derevenetz (commented):

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
ihhub: I think this is a bug. We shouldn't do this. I will fix.

idshibanov: You can probably merge this logic with the previous one (for Courier) that sorts the heroes before assigning them.

oleg-derevenetz (approved):

PR #6173, opened 2022-11-13
Tower::GetInfo(): correctly reflect the fact that the tower was destroyed during the battle

oleg-derevenetz:
fix #6160

PR #6177, opened 2022-11-13
Tower::GetInfo(): add const

oleg-derevenetz:
Follow-up to the #6173. Overlooked the opportunity to add `const` when refactoring old code.


ihhub:
@oleg-derevenetz , thanks for this tiny improvement!

PR #6185, opened 2022-11-14
Make the castle: allow guardians option by default

oleg-derevenetz:
Alternative approach to #6190
Related to #6161

This option is supported well enough both in the game logic and game UI, some efforts have been made to polish it and it would be a pity to throw it all away. AI doesn't know how to perceive its benefits for now, but at least it doesn't conflict with this feature.

After all, we already have changes in the core that affect the overall game balance even more - for example, the ability to hire non-upgraded units in the upgraded dwellings.

PR #6190, opened 2022-11-14
Remove the castle: allow guardians option

oleg-derevenetz:
Alternative approach to #6185
Related to #6161


ihhub:
@oleg-derevenetz , thank you so much for managing these options!

PR #6192, opened 2022-11-15
Integrate the game: autosave will be made at the beginning of the day option

oleg-derevenetz:
Alternative approach to #6195
Related to #6161

~~Both autosave files will be created, with names "AUTOTDAY" and "AUTOYDAY" (subject to discussion, but that's the best I've come up with so far - shorter names are better, and the "SAVE" part is redundant IMHO - obviously, these are save files because they are displayed in the list of save files).~~

Whether to perform the auto save at the beginning of the turn or at the end will be determined by the setting in the `fheroes2.cfg`.


zenseii:
For the record I'm fine with this implementation of it. Due to the points provided in https://github.com/ihhub/fheroes2/issues/6161#issuecomment-1315259784 I see that we won't come to an agreement on something that can possibly be implemented before 1.0, so I suggest leaving this PR open until after 1.0 and then looking at it.

This is my current point of view here. As always I'm always open to listening to any points made otherwise.

LeHerosInconnu:
Hello @oleg-derevenetz and @zenseii,

I suggest to keep the name "AUTOSAVE" in uppercase for the file corresponding to the same autosave file of the original game and to name the file "AUTOTURN" in uppercase for the file saved after the actions of the other players.
The name "AUTOTURN" refers to the save made automatically at the beginning of the player's turn.

zenseii (approved):

zenseii:
> How is it implemented in Heroes 4?

@ihhub, they had a bigger window so the names were like this:
image

ihhub:
In my opinion we should keep `autosave` as it to avoid confusion from users and add a separate UI option to toggle autosave at the start of the day so users won't be surprised seeing 2 save files at the same time.

zenseii:
I'd like to add in the the times when this feature comes in handy there will inevitably be a big time difference separating these two saves, because a lot will have happened during the opponent's turn. Maybe it's only me but even now I always look at the time and date of a save when I need to load because the system is simply alphabetical.

The times when there's not a huge time gap between these two autosaves it wouldn't even matter whether you pick one or the other since you would just be a quick "end turn" away from the other.

ihhub:
Hi @oleg-derevenetz , @zenseii and @LeHerosInconnu , I am fine with keeping it as an option in the configuration file like we do for some other options. This is not a logic changing option. If you are okay with this then we can solve this last experimental option issue. What do you think?

oleg-derevenetz:
Hi @ihhub then we need to squeeze it somehow in the GUI. Is there a place for that?

ihhub:
> Hi @ihhub then we need to squeeze it somehow in the GUI. Is there a place for that?

The problem is that I'm not sure how to name this option and what icons to give. Squeezing GUI is another headache😐 Having the option to be changed through the configuration file is not a new approach as we do for some existing features.

oleg-derevenetz:
@ihhub Well yes, config-only setting will go too, why not.

ihhub:
@oleg-derevenetz , thank you very much for this change! I merged it since basically we just moved this option from one place to another.

LeHerosInconnu:
Hello @ihhub, @oleg-derevenetz and @zenseii,

> Hi @oleg-derevenetz , @zenseii and @LeHerosInconnu , I am fine with keeping it as an option in the configuration file like we do for some other options. This is not a logic changing option. If you are okay with this then we can solve this last experimental option issue. What do you think?

I see I'm a little late. :)
This is not a type of option that the player will or would need to change often.
He will change it once and forget about it.
So it is acceptable to do it in an external file outside the game GUI.

It would also be convenient to have both saves.
So why not make the option decide only if the second save file ('TURNSAVE') is generated or not?
So we would have the "best of both worlds" ("AUTOSAVE" file only, or "AUTOSAVE" AND "TURNSAVE" files).

oleg-derevenetz:
> So we would have the "best of both worlds" ("AUTOSAVE" file only, or "AUTOSAVE" AND "TURNSAVE" files).

There were objections to having two autosave files with similar names - supposedly this would confuse users.

LeHerosInconnu:
Hello @oleg-derevenetz,

> > So we would have the "best of both worlds" ("AUTOSAVE" file only, or "AUTOSAVE" AND "TURNSAVE" files).
> 
> There were objections to having two autosave files with similar names - supposedly this would confuse users.

In this case it would be a deliberate choice of the player and not the fact of automatically having both saves. :)

oleg-derevenetz:
Hi @ihhub

> So we would have the "best of both worlds" ("AUTOSAVE" file only, or "AUTOSAVE" AND "TURNSAVE" files).

What is your opinion regarding this? It is possible to create two files with the option instead of just one.

Branikolog:
Hi everyone! 
My 50cent here... 
I think giving an ability to a player to have both autosaves (at the end and the beginning of a day)  is a nice improvement over the original heroes mechanics.  I'd love to use it myself too. 
But the only thing I'm concerning about is to have them both in save file list. Named similar they would definitely confuse player. Named differently can also confuse players, as this is completely atypical for the game and at the same time it takes space in that tiny short save list we have. 
Right now I cannot purpose any nice solution for them both, but in future, when we add an easy way to expand game windows, we can add into save file list a separate section with autosaves, where we could choose between both autosaves named in proper and obvious to a player way or switch between them  by clicking a single button nearby, so player could any time choose a needed file to load. Normal autosaves could be shown on a default, since players got used to it expect this file load as default. On clicking a button nearby player can switch to new autosave we implemented. 
Otherwise, we can offer two options of loading game, while selecting autosave: player would be introduced to a new dialog, where he can choose between two buttons representing both autosaves each time he wants to load game.

LeHerosInconnu:
@Branikolog.

> Hi everyone! My 50cent here... I think giving an ability to a player to have both autosaves (at the end and the beginning of a day) is a nice improvement over the original heroes mechanics. I'd love to use it myself too. But the only thing I'm concerning about is to have them both in save file list. Named similar they would definitely confuse player. Named differently can also confuse players, as this is completely atypical for the game and at the same time it takes space in that tiny short save list we have. Right now I cannot purpose any nice solution for them both, but in future, when we add an easy way to expand game windows, we can add into save file list a separate section with autosaves, where we could choose between both autosaves named in proper and obvious to a player way or switch between them by clicking a single button nearby, so player could any time choose a needed file to load. Normal autosaves could be shown on a default, since players got used to it expect this file load as default. On clicking a button nearby player can switch to new autosave we implemented. Otherwise, we can offer two options of loading game, while selecting autosave: player would be introduced to a new dialog, where he can choose between two buttons representing both autosaves each time he wants to load game.

I think the list of saves should be kept simple with no sub-windows, but there could be "selection filters" a bit like for the size of the scenario maps.

PR #6193, opened 2022-11-15
Implement all troops transfer to castle dialog

zenseii:
This adds quick troops transferring to the castle dialog using the implementation from this PR https://github.com/ihhub/fheroes2/pull/5785 by @oleg-derevenetz.

This will only add the hotkey for this as that was @Branikolog's requirement for 1.0, unless I misunderstood.

The mouse command for doing this depends on the drag-and-drop ability that @ihhub is working on - I believe this current PR can implement only the keyboard/button hotkeys and then the drag-and-drop can come later on when ready. This is because button hot keys will very much benefit the supported systems that use controller input.

The buttons used are up and down. This is visually logical.

Relates to issue https://github.com/ihhub/fheroes2/issues/4084

Relates to PRs https://github.com/ihhub/fheroes2/pull/5924 and https://github.com/ihhub/fheroes2/pull/5456


zenseii:
I need to find a better solution than passing the parameter `castleMove` which only purpose is to adjust for an assertion....
If only there was a way to check if troops are garrisoned/in a castle or not...

src/fheroes2/castle/castle_dialog.cpp
```diff
@@ -395,6 +395,7 @@ Castle::CastleDialogReturnValue Castle::OpenDialog( const bool readOnly, const b
 
     CastleDialogReturnValue result = CastleDialogReturnValue::DoNothing;
     bool need_redraw = false;
+    bool redrawAfterArmyAction = false;
```
oleg-derevenetz, line 398: Maybe rename it something like `isArmyActionPerformed`? While redraw is needed after this, this is not only thing that needs to be done.

oleg-derevenetz (commented):
Hi @zenseii I just performed some simplification of the logic and left one optional suggestion.

src/fheroes2/castle/castle_dialog.cpp
zenseii: Thanks! Yes this name is better.

zenseii (commented):
I've looked through your changes and this is much better.
Thanks for taking the time to make these changes, @oleg-derevenetz.

For me this PR is now good to go, @ihhub.

zenseii:
> Hi, @zenseii Castle army transfer works nice with arrows, but I wish we also have special buttons for "mouse-only-players". ;)

Thank you for testing it and I'm glad you found it useful, @Branikolog . As I stated in the opening message, for mouse and touch users we will come back to it once we have drag-and-drop interaction available. This should allow us to have an intuitive way of doing it while retaining the original look and not have to squeeze in buttons.

Branikolog:
@zenseii 
I apologize for my skepticism, but how drag-n-drop interaction would help us in this case? Will it be like dragging with both mouse buttons being pressed? I'm convinced, that separate buttons is the best solution... But as you stated, it requires squeezing and definitely new buttons. :(

zenseii:
> @zenseii I apologize for my skepticism, but how drag-n-drop interaction would help us in this case? Will it be like dragging with both mouse buttons being pressed? I'm convinced, that separate buttons is the best solution... But as you stated, it requires squeezing and definitely new buttons. :(

There are several ways, but I had a very simple and quick solution in my first PR for this involving a simple downwards swipe movement or a upwards swipe movement implemented. The only missing part from that command was the visual drag and drop.

Anyways I think this feature of moving all troops implemented here with just a hotkey is a must for the players using gamepads like Switch, PS Vita, Steam Deck etc. I really feel sorry for them having to do such a slow and manual process, which now will be 1 button press.

Branikolog:
@zenseii 

> Anyways I think this feature of moving all troops implemented here with just a hotkey is a must for the players using gamepads like Switch, PS Vita, Steam Deck etc.

Absolutely yes! Thanks for your answer!

ihhub (approved):

oleg-derevenetz:
Hi @zenseii I reduced the variable scopes a bit.

oleg-derevenetz (approved):

Issue #6105, opened 2022-11-06
Translate all game buttons


Effektus:
### Your question

And what hapend with menu translation i found only "Only battle" working?
![image](https://user-images.githubusercontent.com/12934196/200168512-0122230e-5fb5-48b5-9d08-85d390387ad8.png)

### Additional info

![image](https://user-images.githubusercontent.com/12934196/200168437-332079c3-5e81-473e-ae29-f9f5369d910c.png)

oleg-derevenetz:
@Effektus These buttons are images taken from the resources of the original game. If you have a localized variant of the original game, these buttons will have localized inscriptions, otherwise they will have English inscriptions. The "Battle Only" button is absent in the original resources, it is "hand-made" (generated by engine), so it's localized (sort of).

Effektus:
I did not understand? My question is, is anyone making these translated pictures now, or is the engine generating them?

oleg-derevenetz:
@Effektus These translated pictures are done by the publisher of the original game a long time ago for localized distributions intended for the respective countries. There is only one (AFAIK) button image generated by the engine - for the "Battle Only" button.

Effektus:
Yes my question is why is only one generated and where is this engine? It is build in "fheroes2" or it is outside project?

oleg-derevenetz:
@Effektus Because such generation is a difficult task (I suppose @sensei would be able to describe this process in detail), and this is just for the one button. We can not just supply a "localized copies" of these images with fheroes2 due to the licensing issues.

zenseii:
Hi, @Effektus.

As @oleg-derevenetz described thoroughly we are writing code that in runtime generates these buttons, because simply editing them in a graphics editing tool like photoshop and then distributing them with fheroes2 would lead to both licensing and copyright issues because we do not own these assets.

As for why the rest have not been translated it is because this is work in progress. I was intending to do this but right now I'm lost without a PC to develop on for an indefinite time due to unforseen events. Anyone is more than welcome to do this work and also to ask for help in achieving it. The foundaments have already been put in place to make it possible.

Also note that we have actually generated 5 buttons: Battle Only, Difficulty, Min (Recruit), Min (Fast split), Gift.

Also, @oleg-derevenetz, please note a minor typo 😅

> @sensei

oleg-derevenetz:
@zenseii

> Also, @oleg-derevenetz, please note a minor typo 😅

Yes, sorry :)

Effektus:
[...]

PR #6194, opened 2022-11-15
Make the Main Menu's dialog buttons translatable

zenseii:
This makes it possible to translate the main menu buttons. More need to be added still and there should be some discussion on structure.

Relates to https://github.com/ihhub/fheroes2/issues/6105


zenseii:
Hi, @ihhub.

I've added some buttons just to have example code that can be discussed. My biggest concern is how to preserve the original buttons from current asset and restore them when changing back from another language. Something similar to `OriginalAlphabetPreserver` could work.

ihhub:
> Hi, @ihhub.
> 
> I've added some buttons just to have example code that can be discussed. My biggest concern is how to preserve the original buttons from current asset and restore them when changing back from another language. Something similar to `OriginalAlphabetPreserver` could work.
> 
> Please have a look at this when it suits you.

Hi @zenseii , the main question is do we want to show the original buttons at all? All original buttons are a part of ICN assets so they are present within the game. Based on the code we aren't using them at all.

ihhub:
Hi @zenseii , then for these buttons if a selected language is the same as the assets we can add the code to retrieve images from the original ICNs. Something like this:
case ICN::NEW_GAME_BUTTON:
    if ( language == Language::Polish && crc32 == polishcrc32 ) {
        return fheroes2::AGG::GetICN( ICN::OriginalICN, id );
    }
...

ihhub:
Hi @zenseii , to modify the size of space character please refer to getSpaceCharWidth() function in ui_text.cpp file.

src/fheroes2/gui/ui_language.cpp
```diff
@@ -259,4 +259,12 @@ namespace fheroes2
 
         AGG::updateLanguageDependentResources( language, isOriginalResourceLanguage );
     }
+
+    SupportedLanguage getCurrentLanguage()
+    {
+        const Settings & conf = Settings::Get();
+
+        const fheroes2::SupportedLanguage currentLanguage = fheroes2::getLanguageFromAbbreviation( conf.getGameLanguage() );
```
ihhub, line 267: We can directly return the value of this function call without the need of a temporary variable.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -890,30 +1148,34 @@ namespace fheroes2
...
+                    if ( generateFrenchSpecificImages( id ) ) {
+                        return;
+                    }
+                    break;
+                case fheroes2::SupportedLanguage::Polish:
+                    if ( generatePolishSpecificImages( id ) ) {
+                        return;
+                    }
+                    break;
+                case fheroes2::SupportedLanguage::Italian:
```
ihhub, line 1171: What's about Russian? Is @Branikolog not happy to use the original resources?

ihhub (commented):
Hi @zenseii , I left few comments here. Can you please check them?

src/fheroes2/agg/agg_image.cpp
ihhub: Ideally, we should remove all crappy buttons generated from other original buttons but sometimes they are too different from what we generate.

ihhub:
@zenseii , huge thanks for these changes!

PR #6166, opened 2022-11-12
Reference project affiliated donation and social network links

zenseii:
Since we have created this Boosty account as per request of some users we should add the link to the relevant places and add other relevant links.

PR #6199, opened 2022-11-16
Rework credits page

ihhub:
To support multiple links for donation and social networks as well as to expand the list of participants.

relates to #6166


src/fheroes2/game/game_credits.cpp
```diff
@@ -275,107 +285,86 @@ namespace
...
+                                "Aimi Lindschouw\n"
+                                "Kiril Lipatov\n"
+                                "Matt Taylor\n"
+                                "Andrew Szucs\n"
+                                "Kresimir Condic\n"
+                                "slvclw\n" );
+
+        name.Set( supporters, Font::BIG, textWidth );
+        const int32_t supportersHeight = name.h();
+        name.Blit( textInitialOffsetX + ( textInitialOffsetX - name.w() ) / 2, offsetY + ( constributorsHeight - supportersHeight ) / 2, output );
```
zenseii, line 305: Should we line these two columns equally?
image

Also maybe add the titles to each colum "Contributors" "Supporters" ?

zenseii (changes requested):
This looks good to me I just left two suggestions, @ihhub. Please have a look when you have the time to.

src/fheroes2/game/game_credits.cpp
oleg-derevenetz: I suppose they are just vertically centered, and since they have a different height, there is a shift between lines. IMHO they looks good as is.

ihhub: It is fixed.

oleg-derevenetz (approved):

PR #6201, opened 2022-11-16
Remove the world: Scouting skill shows extended content info option

oleg-derevenetz:
Related to #6161

The previous logic was somewhat overcomplicated and confusing due to the need to focus on the scouting skill level in lots of places, I hope I didn't miss anything. It would be good to additionally test the new logic in different situations with artifacts that affect object info visibility (Crystal Ball?) to ensure that I didn't miss something in the mechanics.

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -30,6 +30,7 @@
 #include <ostream>
 #include <random>
 #include <set>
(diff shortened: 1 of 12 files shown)


Branikolog (approved):

ihhub:
@oleg-derevenetz , thank a lot for this change!

PR #6206, opened 2022-11-18
Add SDLCALL to callbacks used by SDL, remove the SDL_SetEventFilter()

oleg-derevenetz:
Technically `SDLCALL` usually expands either to `__cdecl` or to an empty string, but not always. Also SDL_mixer 1.x doesn't require its callbacks to be `SDLCALL`, while SDL_mixer 2.x does.

Also I noticed that `SDL_SetEventFilter()` is used. Please do not ever use it - as rightly [stated in the documentation](https://wiki.libsdl.org/SDL_SetEventFilter), the callback function set by `SDL_SetEventFilter()` has the same issues as any other SDL callbacks - it might be called from an internal SDL thread! (remember SDL_mixer callbacks). So we can't reliably use it to redraw software cursor or switch the fullscreen mode. The only thing you can do safely in it is filter an event out from the event queue by returning 0, that's what this function is designed for. You can't access or alter any external data in the callback without the proper synchronization.

P.S. Hi @ihhub do we use any SDL callbacks anywhere else?

### DIFF
--- a/iwyu.map
+++ b/iwyu.map
@@ -1,7 +1,9 @@
 [
-    { include: [ "<bits/getopt_core.h>",     private, "<unistd.h>", public ] },
-    { include: [ "<bits/std_abs.h>",         private, "<cmath>",    public ] },
-    { include: [ "<bits/std_abs.h>",         private, "<cstdlib>",  public ] },
-    { include: [ "<bits/types/struct_tm.h>", private, "<ctime>",    public ] },
-    { include: [ "<ext/alloc_traits.h>",     private, "<memory>",   public ] }
+    { include: [ "<bits/getopt_core.h>",     private, "<unistd.h>",     public ] },
+    { include: [ "<bits/std_abs.h>",         private, "<cmath>",        public ] },
+    { include: [ "<bits/std_abs.h>",         private, "<cstdlib>",      public ] },
+    { include: [ "<bits/types/struct_tm.h>", private, "<ctime>",        public ] },
+    { include: [ "<ext/alloc_traits.h>",     private, "<memory>",       public ] },
+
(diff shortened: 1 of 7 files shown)


oleg-derevenetz:
BTW, the removal of `SDL_SetEventFilter()` cured weird crashes on macOS with SDL1 while dynamically switching between fullscreen and windowed modes described [here](https://github.com/ihhub/fheroes2/pull/6102#issuecomment-1304873278). Previously ASAN reported out-of-bounds reads and writes by the `Image` class methods during this switch somewhere in the region of the previously allocated SDL screen buffer (it is reallocated when switching between fullscreen and windowed mode). I suppose that, due to the lack of the proper synchronization, "software mouse emulation" code continued to read or write to the old memory area of the screen buffer, although it had already been released. Now this doesn't happen anymore.

ihhub:
Hi @oleg-derevenetz , we use SDL only in 3 places:
- screen (display) related code in screen.cpp file. It doesn't have any callbacks
- audio
- localevent

So you didn't miss any place with SDL callback.

ihhub:
> Maybe this will also cure [this](https://github.com/ihhub/fheroes2/pull/6092#issuecomment-1311938983) issue?

Yes, it fixes the crash. I've tested and no more crash (however, mouse position issue is still there).

ihhub:
Hi @oleg-derevenetz , are you going to make any other changes in this PR?

oleg-derevenetz:
@ihhub

> are you going to make any other changes in this PR?

Since this PR fixes crashes and no other SDL callbacks are involved, I assume not.

ihhub:
@oleg-derevenetz , huge thanks for fixing these crashes!

PR #6209, opened 2022-11-18

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thank you very much for maintaining good code quality!

Issue #6036, opened 2022-10-24
Program screen goes completely black after the program is minimized in taskbar and then brought to the foreground


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In windowed mode (when a View spell is active or the View World screen is used), when the game window is minimized in the taskbar, then the game is brought to the foreground (by clicking on the game icon in the taskbar), the game window is completely black (or sometimes completely white).
It is necessary to switch to another program (by clicking its icon in the taskbar) and then return to the game window (again by clicking its icon in the taskbar) so that the content of the window is displayed correctly.

In fheroes2:

![View spells 001](https://user-images.githubusercontent.com/43583994/197519285-1ff52e69-4a4e-46f0-aa17-c77394b153d6.png)

![View spells 002](https://user-images.githubusercontent.com/43583994/197519290-6e5ff32e-e48c-47f2-a2e3-a2a821278d47.png)

### Save file

Not needed.

### Additional info

Related closed issue: https://github.com/ihhub/fheroes2/issues/3392.

oleg-derevenetz:
Interesting, I was not able to reproduce this issue so far using the "View World" screen. Is this observed with SDL1 or SDL2? What resolution is set in the game?

LeHerosInconnu:
Hello @oleg-derevenetz,

> Interesting, I was not able to reproduce this issue so far using the "View World" screen. Is this observed with SDL1 or SDL2? What resolution is set in the game?

Windows 7 - 64 bits - SDL2 - 640 x 480

zenseii:
I tested this on my 32-bit Win 7 SDL2 and I was able to replicate it. I had version 0.9.14 on the same PC and it did not show this behavior when restoring from minimized.
It did not happen on my Win 10 device.

ihhub:
These symptoms indicate that we do not refresh frame for this dialog. I think we can come up with something simple and efficient at the same time by calling an internal sdl function to refresh screen.

ihhub:
Hi @LeHerosInconnu , can you please test the latest build to se if the issue is gone?

zenseii:
[...]

Issue #4770, opened 2021-12-10
View world: restoring from windowed to fullscreen mode


a1exsh:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Linux

### Describe the bug

When the View World is open, going from windowed mode to full screen produces the following result (captured just the top-left corner, the rest of the screen is black):
![image](https://user-images.githubusercontent.com/489601/145544895-38a2a3a1-06c2-46a2-909c-27d5c15ea045.png)

Everything is rendered properly again as soon as I use mouse wheel to zoom or exit the View World mode.

I wasn't able to trigger this issue while switching between windowed and fullscreen modes in other places in the game, so I suppose some specific handling is missing in View World.

### Save file

No save file needed: just start a new game.

ihhub:
Hi @a1exsh , I can't reproduce this issue. Can you please tell what resolution do you use?

a1exsh:
@ihhub The actual display resolution is 3840 x 2160, but the game is set to use 800 x 600.

a1exsh:
I've just tried using a lower display resolution and I can reproduce it on 1920 x 1080 and 1024 x 768 as well (the game is at 800 x 600 in all cases).

Actually, I notice a similar rendering issue when switching between windowed and full screen modes in other places (like adventure map or castle screen), but there it is quickly recovering, as the next frame is rendered.

a1exsh:
Probably related is the fact that on View World screen we don't have the system info (wall time and FPS).

Compare adventure map:
![image](https://user-images.githubusercontent.com/489601/145628691-4126739d-0c85-4eb6-8bad-d78a2f23bd4e.png)

And View World:
![image](https://user-images.githubusercontent.com/489601/145628709-9bd5b31c-6aae-4da3-b7fd-813e37518260.png)

a1exsh:
> Probably related is the fact that on View World screen we don't have the system info (wall time and FPS).

[...]

PR #6213, opened 2022-11-19
Rework SDL events handling

ihhub:
The current implementation of SDL events is a complete mess. We do not even handle most of events and these should be disabled to avoid extra processing. Proper separation of SDL1 and SDL 2 allows us to support the code without mind-bending hacks. The place where initialization of event types is initiated should be within core initialization code as it belongs there.

Moreover, I fixed the place where we update cursor before we even initialize the system.

### DIFF
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -25,6 +25,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <map>
+#include <set>
 #include <utility>
 #include <vector>
 
@@ -678,19 +679,19 @@ namespace
 #endif
 
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
-    std::map<uint32_t, bool> eventTypeStatus;
+    std::set<uint32_t> eventTypeStatus;
 
     void setEventProcessingState( const uint32_t eventType, const bool enable )
     {
-        eventTypeStatus[eventType] = enable;
+        eventTypeStatus.emplace( eventType );
         SDL_EventState( eventType, ( enable ? SDL_ENABLE : SDL_IGNORE ) );
     }
 #else
-    std::map<uint8_t, bool> eventTypeStatus;
+    std::set<uint8_t> eventTypeStatus;
 
     void setEventProcessingState( const uint8_t eventType, const bool enable )
     {
-        eventTypeStatus[eventType] = enable;
+        eventTypeStatus.emplace( eventType );
         SDL_EventState( eventType, ( enable ? SDL_ENABLE : SDL_IGNORE ) );
     }
 #endif
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -1289,14 +1289,9 @@ bool LocalEvent::HandleEvents( bool delay, bool allowExit )
             }
             break;
         default:
-            if ( allowedEventTypes.count( event.type ) > 0 ) {
-                // If this assertion blows up then we included an event type but we didn't add logic for it.
-                assert( 0 );
-            }
-            else {
-                // If this assertion blows up then SDL might be broken as we disabled this event type.
-                assert( disabledEventTypes.count( event.type ) == 0 );
-            }
+            // If this assertion blows up then we included an event type but we didn't add logic for it.
+            assert( eventTypeStatus.count( event.type ) == 0 );
+
             // This is a new event type which we do not handle. It might have been added in a newer version of SDL.
             break;
         }
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -51,7 +51,6 @@
 
 #include "audio.h"
 #include "localevent.h"
-#include "logging.h"
 #include "pal.h"
 #include "screen.h"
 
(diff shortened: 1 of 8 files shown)


ihhub:
This PR might fix #6036 and #4770 because of `SDL_RENDER_TARGETS_RESET` event but this should be verified.

src/engine/localevent.cpp
```diff
@@ -674,6 +677,38 @@ namespace
...
+        else {
+            disabledEventTypes.emplace( eventType );
+            SDL_EventState( eventType, SDL_IGNORE );
+        }
+    }
+#else
+    std::set<uint8_t> allowedEventTypes;
+    std::set<uint8_t> disabledEventTypes;
+
+    void setEventProcessingState( const uint8_t eventType, const bool enable )
```
oleg-derevenetz, line 700: What if this function will be called twice with the same value of `eventType`, but different value of `enable`? It seems that this event type will be both in `allowedEventTypes` and `disabledEventTypes` at the same time. Is this OK?

src/engine/localevent.cpp
```diff
@@ -1214,31 +1249,94 @@ bool LocalEvent::HandleEvents( bool delay, bool allowExit )
...
+            fheroes2::Display & display = fheroes2::Display::instance();
+            fheroes2::Image temp( display.width(), display.height() );
+            if ( display.singleLayer() ) {
+                temp._disableTransformLayer();
+            }
+
+            fheroes2::Copy( display, temp );
+            display.release();
+            display.resize( temp.width(), temp.height() );
+            fheroes2::Copy( temp, display );
```
oleg-derevenetz, line 1271: How about moving this code to a separate function `HandleXXXEvent()`, like it's done with other events? This is optional, of course, I leave it at your discretion.

src/engine/screen.cpp
```diff
@@ -434,6 +440,9 @@ namespace
             }
 
             SDL_Cursor * tempCursor = SDL_CreateColorCursor( surface, offsetX, offsetY );
+            if ( tempCursor == nullptr ) {
+                ERROR_LOG( "Failed to create a cursor. The error description: " << SDL_GetError() )
+            }
             SDL_SetCursor( tempCursor );
```
oleg-derevenetz, line 446: Should we call `SDL_SetCursor()` with null cursor here? I know that there will be no harm, but is this needed for some reason?

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions, could you please take a look?

src/engine/screen.cpp
ihhub: It is done.

src/engine/localevent.cpp
ihhub: It is done. I also moved some methods into the same place under one macro. There is still a lot of mess...

src/engine/localevent.cpp
ihhub: I moved to use `std::map`. The previous implementation was overengineering.

src/engine/localevent.cpp
```diff
@@ -674,6 +676,24 @@ namespace
         currentCharIndex = 0;
     }
 #endif
+
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+    std::map<uint32_t, bool> eventTypeStatus;
+
+    void setEventProcessingState( const uint32_t eventType, const bool enable )
+    {
+        eventTypeStatus[eventType] = enable;
```
oleg-derevenetz, line 685: I believe we can use the `std::set` here. The options are:

1. We disabled the event, therefore it should never arrive and the `default` clause, that checks the assertion, will never be reached;
2. We enabled the event and added the `case` clause for it, so the `default` clause will never be reached as well;
3. We enabled the event but forgot to add the `case` clause for it, so the `default` clause will be reached and assertion will fail;
4. There is some unknown event, the `default` clause will be reached, but assertion will not fail, because the `eventTypeStatus` (maybe rename it as `knownEventTypes`?) set will have no entry for it.

What do you think? This is optional, of course, because `std::map` should work too, but it looks a bit redundant.

src/engine/screen.cpp
```diff
@@ -434,7 +440,12 @@ namespace
             }
 
             SDL_Cursor * tempCursor = SDL_CreateColorCursor( surface, offsetX, offsetY );
-            SDL_SetCursor( tempCursor );
+            if ( tempCursor == nullptr ) {
+                ERROR_LOG( "Failed to create a cursor. The error description: " << SDL_GetError() )
+            }
```
oleg-derevenetz, line 445: Should we call `clear()` below (that will release the current cursor) and swap the `_cursor` and `tempCursor` if we failed to create the `tempCursor`?

oleg-derevenetz (commented):
Hi @ihhub I just left an optional proposal and a question, please take a look when you have time.

src/engine/localevent.cpp
ihhub: You're right. I shall change it.

src/engine/screen.cpp
ihhub: I think we shouldn't since we don't call `SDL_SetCursor` for this case.

oleg-derevenetz (approved):

PR #6216, opened 2022-11-19
Change battlefield hexagon generation to match the original

### LINKED ISSUE #1492: Combat screen, grid, vertical dimension and irregularity of the lines of the hexagons
In fheroes2, the position of the grid at the top of the screen is offset by 1 pixel from the original game.
At the bottom of the screen, the grid is positioned correctly.
Also, the oblique lines that make up a hexagon are not drawn regularly, the segments that make up these lines are not all the same length.
In the original game, all of these segments are the same length.

In original game:

Battlefield H2 01

In fheroes2:

Battlefield fh2 02

Open the two images in different tabs in your browser, enlarge the images and switch between the two tabs alternately.

Battlefield H2 01 left grid creature indicator 01 Battlefield fh2 02 left grid creature indicator 01

Districh-ru:
close  #1492

This PR:
fheroes2

Original game:
Original

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -485,9 +485,9 @@ fheroes2::Image DrawHexagon( const uint8_t colorId )
     fheroes2::Image sf( w + 1, h + 1 );
     sf.reset();
 
-    fheroes2::DrawLine( sf, { r - 1, 1 }, { 0, l + 1}, colorId );
+    fheroes2::DrawLine( sf, { r - 1, 1 }, { 0, l + 1 }, colorId );
     fheroes2::SetPixel( sf, r, 1, colorId );
-    fheroes2::DrawLine( sf, { r + 1, 1 }, { w, l + 1}, colorId );
+    fheroes2::DrawLine( sf, { r + 1, 1 }, { w, l + 1 }, colorId );
 
     fheroes2::DrawLine( sf, { 0, l + 1 }, { 0, h - l }, colorId );
     fheroes2::DrawLine( sf, { w, l + 1 }, { w, h - l }, colorId );
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -507,7 +507,7 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
 
     fheroes2::Image sf( w, h );
     sf.reset();
-    fheroes2::Rect rt( 1, l - 1, w + 1, 2 * l + 4 );
+    fheroes2::Rect rt( 2, l - 1, w - 3, 2 * l + 4 );
     for ( int i = 0; i < w / 2; i += 2 ) {
         for ( int x = 0; x < rt.width; ++x ) {
             for ( int y = 0; y < rt.height; ++y ) {
@@ -517,7 +517,7 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
         --rt.y;
         rt.height += 2;
         rt.x += ( i == 0 ) ? 1 : 2;
-        rt.width -= 4;
+        rt.width -= ( i == 0 ) ? 2 : 4;
     }
 
     return sf;
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -485,14 +485,16 @@ fheroes2::Image DrawHexagon( const uint8_t colorId )
     fheroes2::Image sf( w + 1, h + 1 );
     sf.reset();
 
-    fheroes2::DrawLine( sf, { r, 0 }, { 0, l }, colorId );
-    fheroes2::DrawLine( sf, { r, 0 }, { w, l }, colorId );
+    fheroes2::DrawLine( sf, { r - 1, 1 }, { 0, l + 1}, colorId );
+    fheroes2::SetPixel( sf, r, 1, colorId );
+    fheroes2::DrawLine( sf, { r + 1, 1 }, { w, l + 1}, colorId );
 
     fheroes2::DrawLine( sf, { 0, l + 1 }, { 0, h - l }, colorId );
     fheroes2::DrawLine( sf, { w, l + 1 }, { w, h - l }, colorId );
 
-    fheroes2::DrawLine( sf, { r, h }, { 0, h - l }, colorId );
-    fheroes2::DrawLine( sf, { r, h }, { w, h - l }, colorId );
+    fheroes2::DrawLine( sf, { r - 1, h }, { 0, h - l }, colorId );
+    fheroes2::SetPixel( sf, r, h, colorId );
+    fheroes2::DrawLine( sf, { r + 1, h }, { w, h - l }, colorId );
 
     return sf;
 }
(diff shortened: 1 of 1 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1095,7 +1097,7 @@ Battle::Interface::Interface( Arena & a, int32_t center )
 
     // hexagon
     sf_hexagon = DrawHexagon( fheroes2::GetColorId( 0x68, 0x8C, 0x04 ) );
-    sf_cursor = DrawHexagonShadow( 2 );
```
ihhub, line 1098: Could you please explain this change?

Districh-ru: Hi @ihhub, this change was made tо set the strength of shadow under the cursor close to original game.
In fheroes2 it is darker, than in original.
Latest build of fheroes2:
изображение   изображение
This PR:
изображение  изображение
Original game:
изображение   изображение

zenseii: Personally I think the color difference is too small in the original so darker is better.

ihhub: Yes. We did this intentionally to have better visuals.

Districh-ru: I also more like the darker cursor shadow. I can revert this in the next commit to this PR.
With the next commit I want to try to rewrite shadow generation code to make shadows like in original game. In fheroes2, if you disable hex grid, you will see local shadows inside nonshadowed hex grid.
изображение

ihhub: Hi @Districh-ru , let revert this change but to do not make changes in shadow generation. The space in shadows were also made deliberately to easier understand how many hexes your monster can move.

zenseii: I'm not sure I understand what the current issue with shadows inside nonshadowed hex grid is, @Districh-ru ?

zenseii: In fact it is so deliberate that you can read about it in our wiki on enhancements and features:
https://github.com/ihhub/fheroes2/wiki/Features-and-enhancements-of-the-project

Districh-ru: Ok, @ihhub.
In the original game, the movement field shadow is solid. That's where I got the idea :)

ihhub: Thank you!

ihhub (approved):

src/fheroes2/battle/battle_interface.cpp
Districh-ru: > In fact it is so deliberate that you can read about it in our wiki on enhancements and features: https://github.com/ihhub/fheroes2/wiki/Features-and-enhancements-of-the-project

Thank you @ihhub  and @zenseii for stopping me from deleting the feature of this project! :)

Districh-ru:
@zenseii, @ihhub 
should I revert the space between hex shadows (to 3 pixels from one)?
It will be:
изображение   изображение

zenseii:
> @zenseii, @ihhub should I revert the space between hex shadows (to 3 pixels from one)? It will be:

@Districh-ru, I think if the pixel experts, @LeHerosInconnu and @Branikolog, didn't have anything to say about this then there's no reason to make such a change unless you notice something obviously ugly with how it works right now.

Districh-ru:
@zenseii, I reverted the space between the hexagonal shadows. It is now the same as the current master build, except for changes to fit the new hexagon shape.

zenseii (approved):

ihhub:
@Districh-ru , thanks you so much for this improvement!

LeHerosInconnu:
Hello everyone,

> > @zenseii, @ihhub should I revert the space between hex shadows (to 3 pixels from one)? It will be:
> 
> @Districh-ru, I think if the pixel experts, @LeHerosInconnu and @Branikolog, didn't have anything to say about this in the current master build then there's no reason to make such a change unless you notice something obviously ugly with how it works right now.

In fact, the shadow area of the hexagons for creature movement, for the shadow of the cursor position, for the effect area of attacks, etc. should be changed to fill the entire hexagon as it is in the original game.
Also, the shadow is too dark compared to the original game.
I will open an issue for this.

zenseii:
> Hello everyone,
> 
> > > @zenseii, @ihhub should I revert the space between hex shadows (to 3 pixels from one)? It will be:
> > 
> > 
> > @Districh-ru, I think if the pixel experts, @LeHerosInconnu and @Branikolog, didn't have anything to say about this in the current master build then there's no reason to make such a change unless you notice something obviously ugly with how it works right now.
> 
> In fact, the shadow area of the hexagons for creature movement, for the shadow of the cursor position, for the effect area of attacks, etc. should be changed to fill the entire hexagon as it is in the original game. Also, the shadow is too dark compared to the original game. I will open an issue for this.

Hi, @LeHerosInconnu.

I don't think we should make the shadow brighter. Please try to open the ship battle arena from the Battle Only menu and see that even now the shadow could be darker because it is almost indistinguishable. If we used the original shade it would be even worse.

LeHerosInconnu:
Hello @zenseii,

> > In fact, the shadow area of the hexagons for creature movement, for the shadow of the cursor position, for the effect area of attacks, etc. should be changed to fill the entire hexagon as it is in the original game. Also, the shadow is too dark compared to the original game. I will open an issue for this.
> 
> Hi, @LeHerosInconnu.
> 
> I don't think we should make the shadow brighter. Please try to open the ship battle arena from the Battle Only menu and see that even now the shadow for the movement range could be darker because it is almost indistinguishable. If we used the original shade it would be even worse. The snow terrain battle field is also a good example of why the original is too bright.

Effectively, I see what you mean.
Would it be possible to make the level of the shadow (more or less dark) be done automatically according to the background so that the shadow is more visible according to the case?

Districh-ru:
> Hello @zenseii,
> 
> > > In fact, the shadow area of the hexagons for creature movement, for the shadow of the cursor position, for the effect area of attacks, etc. should be changed to fill the entire hexagon as it is in the original game. Also, the shadow is too dark compared to the original game. I will open an issue for this.
> > 
> > 
> > Hi, @LeHerosInconnu.
> > I don't think we should make the shadow brighter. Please try to open the ship battle arena from the Battle Only menu and see that even now the shadow for the movement range could be darker because it is almost indistinguishable. If we used the original shade it would be even worse. The snow terrain battle field is also a good example of why the original is too bright.
> 
> Effectively, I see what you mean. Would it be possible to make the level of the shadow (more or less dark) be done automatically according to the background so that the shadow is more visible according to the case?

Hi @LeHerosInconnu, yes it is possible - right before setting the shadow strength there is a `switch ( groundType )`, with which we can set the shadow strength parameters for each ground type.
It is also possible to make two types of shadows - one to fit the hexagon grid and the other to use without the grid, with more space between the shading zones.

Branikolog:
Hi, everyone!
I personally OK with current shadow level.

>In fact, the shadow area of the hexagons for creature movement, for the shadow of the cursor position, for the effect area of attacks, etc. should be changed to fill the entire hexagon as it is in the original game.

Speaking about size of hexagons, I really love the way they are separated between each other, and I don't want to squeeze distances between them to be able to clearly understand the distances while playing without green grid.
Probably, we can make 1px wider (to fill the entire grid hexagon) the selected hexagon (under the cursor) to make it more noticeable to a player, but not sure, it's really necessary.
In a case green grid is activated, I'm OK with both implementations, but would love to pick the solution, which would make grid more distinguishable on battlefield.

Districh-ru:
@Branikolog, @LeHerosInconnu, If you think it is ok, I can open a PR with two types of hexagonal shadow:
      shadow that fits into the green grid and the same shadow size for the cursor;
      smaller shadow for tha case when green grid is off.

https://user-images.githubusercontent.com/113276641/204127583-4a402184-c0cb-4f22-8fe7-b8d2f2579b78.mp4

LeHerosInconnu:
Hello @Districh-ru,

I tested this.
Well done, it works like a charm. :)

PR #6217, opened 2022-11-19
Make fullscreen mode as a default state for Android devices

ihhub:
To have better initial experience with the game.

relates to #6008

### DIFF
--- a/src/fheroes2/system/settings.cpp
+++ b/src/fheroes2/system/settings.cpp
@@ -114,6 +114,11 @@ Settings::Settings()
     _optGlobal.SetModes( GLOBAL_BATTLE_SHOW_MOVE_SHADOW );
     _optGlobal.SetModes( GLOBAL_BATTLE_AUTO_SPELLCAST );
 
+#if defined( ANDROID )
+    // Due to the nature of Android devices having small screens in general it is good to make fullscreen option by default.
+    _optGlobal.SetModes( GLOBAL_FULLSCREEN );
+#endif
+
     // The Price of Loyalty is not supported by default.
     EnablePriceOfLoyaltySupport( false );
 }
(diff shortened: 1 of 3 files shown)


src/fheroes2/system/settings.cpp
```diff
@@ -114,6 +114,11 @@ Settings::Settings()
     _optGlobal.SetModes( GLOBAL_BATTLE_SHOW_MOVE_SHADOW );
     _optGlobal.SetModes( GLOBAL_BATTLE_AUTO_SPELLCAST );
 
+#if defined( ANDROID )
+    // Due to the nature of Android devices having small screens in general it is good to make fullscreen option by default.
+    _optGlobal.SetModes( GLOBAL_FULLSCREEN );
+#endif
```
oleg-derevenetz, line 120: Hi @ihhub what do you think about moving the device-dependent logic to the `engine`? Say, add something like `isHandheldDevice()` to the `engine` and use it here?

ihhub: Yup, we can do it.

src/fheroes2/system/settings.cpp
```diff
@@ -114,10 +115,10 @@ Settings::Settings()
     _optGlobal.SetModes( GLOBAL_BATTLE_SHOW_MOVE_SHADOW );
     _optGlobal.SetModes( GLOBAL_BATTLE_AUTO_SPELLCAST );
 
-#if defined( ANDROID )
-    // Due to the nature of Android devices having small screens in general it is good to make fullscreen option by default.
-    _optGlobal.SetModes( GLOBAL_FULLSCREEN );
-#endif
+    if ( fheroes2::isHandheldDevice() ) {
+        // Due to the nature of Android devices having small screens in general it is good to make fullscreen option by default.
```
oleg-derevenetz:
```suggestion
        // Due to the nature of handheld devices having small screens in general it is good to make fullscreen option by default.
```

ihhub: It is in the code🙂

oleg-derevenetz (approved):

PR #6219, opened 2022-11-19
Remove the world: Windmills, Water Wheels and Magic Gardens can be captured option

oleg-derevenetz:
Related to #6161


ihhub:

PR #6223, opened 2022-11-20
Rework key modifiers logic

ihhub:
to avoid SDL dependency and make the code readable. Also reduced the scope of some functions.

### DIFF
--- a/src/engine/localevent.h
+++ b/src/engine/localevent.h
@@ -26,6 +26,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <string>
 
 #include <SDL_events.h>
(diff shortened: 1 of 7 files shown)


src/engine/localevent.h
```diff
@@ -178,9 +188,9 @@ class LocalEvent
         mouse_motion_hook_func = pf;
     }
 
-    void SetKeyDownGlobalHook( void ( *pf )( int, int ) )
+    void setGlobalKeyboardEventHook( std::function<void( const fheroes2::Key, const int32_t )> hook )
```
oleg-derevenetz, line 191: In #6206 I named it as `...KeyDown...` to emphasize the fact that it processes only key down events, but ignores key up events, so it's not the "keyboard event hook" in the full sense of the word :) But this is a nitpick, so you are free to ignore it.

oleg-derevenetz (changes requested):
Hi @ihhub LGTM, just one matter-of-taste nitpick. Also could you please add the header requested by IWYU (maybe set this check as required)?

src/engine/localevent.h
ihhub: Sure, I can rename it.

oleg-derevenetz (approved):

PR #6227, opened 2022-11-20
Do not show resolution dialog on handheld devices on the first startup

ihhub:
Because users tend to choose the highest leading to disappointment while seeing tiny objects on Adventure Map.

relates to #6008

### DIFF
--- a/src/fheroes2/game/game_mainmenu.cpp
+++ b/src/fheroes2/game/game_mainmenu.cpp
@@ -198,15 +198,14 @@ fheroes2::GameMode Game::MainMenu( bool isFirstGameRun )
     if ( isFirstGameRun ) {
         fheroes2::selectLanguage( fheroes2::getSupportedLanguages(), fheroes2::getLanguageFromAbbreviation( conf.getGameLanguage() ) );
 
-        std::string resolutionText = _( " Before starting the game please choose game resolution." );
-        std::string greetingsText = _( "Welcome to Heroes of Might and Magic II powered by fheroes2 engine!" );
-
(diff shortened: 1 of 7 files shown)


src/fheroes2/game/game_mainmenu.cpp
```diff
@@ -197,13 +198,19 @@ fheroes2::GameMode Game::MainMenu( bool isFirstGameRun )
     if ( isFirstGameRun ) {
         fheroes2::selectLanguage( fheroes2::getSupportedLanguages(), fheroes2::getLanguageFromAbbreviation( conf.getGameLanguage() ) );
 
-        Dialog::Message( _( "Greetings!" ),
-                         _( "Welcome to Heroes of Might and Magic II powered by fheroes2 engine! Before starting the game please choose game resolution." ), Font::BIG,
-                         Dialog::OK );
+        std::string resolutionText = _( " Before starting the game please choose game resolution." );
```
oleg-derevenetz, line 201: Hi @ihhub this looks dangerous - I bet that translators will regularly overlook this extra space in the beginning. I propose to make two whole strings - one for each case.

ihhub: It makes sense. I shall do.

oleg-derevenetz: It is also better from the point of view that it works better with right-to-left alphabets (although I doubt that there will be such translations in the near future, who knows).

oleg-derevenetz (approved):

PR #6229, opened 2022-11-20
Fix multiple issues with fullscreen mode on Windows

### LINKED ISSUE #5250: Windows: Embedding/locking fheroes2 to a side with DPI higher than 100% breaks fullscreen
### Preliminary checks

When locking the game window to either side by using Windows Key + Left or Right Arrow, and then enter fullscreen breaks the window and it's not able to decide where to go and sometimes the window goes black. I'm on Windows 10. Resolution 1920x1080.

<details>
<summary>Locked to left half:</summary>

before
</details>

<details>
<summary>After pressing F4:</summary>

after

</details>

This is really low priority since there's not much reason to do this, but it's slightly related to the recent works on fullscreen mode.

### Additional info

Relevant discussion here: https://github.com/ihhub/fheroes2/pull/5209#issuecomment-1093628659

### LINKED ISSUE #6120: Fullscreen game mode: issues running on two screens
### Preliminary checks

### Platform

Windows

### Describe the bug

Fheroes2 works really bad using 2 monitors currently.
After initiating the fullscreen mode of fheroes2 with dual monitor set up all applications from the main screen are shifted for ~0.5 screen width to the second screen.
Second screen after switching fheroes2 to fullscreen (on screen 1) with explorer stretched on the whole screen 1:
Screen1

Second screen after switching fheroes2 to fullscreen(on the screen 1) with explorer stretched on the whole screen 2:
Screen2

Note, that we can even see a recycle bin (and all shortcuts) from the right part of the main screen here.

In a case we run fheroes2 on the screen 2 using fullscreen mode, the second screen changes resolution for lower one...:
Screen3
... and the game runs on the main screen in the upper left corner with cursour locked on that area:
Screen4

### Save file

None.

### Additional info

_No response_

ihhub:
close #5250
close #6120
relates to #4770


src/fheroes2/system/settings.cpp
```diff
@@ -123,6 +123,12 @@ Settings::Settings()
         scroll_speed = SCROLL_SPEED_NONE;
     }
 
+    // SDL 1 uses cursor software emulation by default as it does not support anything else. SDL 2 supports 'hardware' mode and software emulation.
+    // It is observed that on small resolutions with a resized window the cursor remains the same in size within the game.
+    // In order to avoid this we set cursor software emulation by default.
+    _optGlobal.SetModes( GLOBAL_CURSOR_SOFT_EMULATION );
+    fheroes2::cursor().enableSoftwareEmulation( true );
```
oleg-derevenetz, line 130: Hi @ihhub hmm, as far as I can see, software emulation only turns on everywhere, but it doesn't turn off anywhere - even if `cursor soft rendering` in `fheroes2.cfg` is `off`.

ihhub: You're right. I need to add the logic to turn it off.

ihhub: Fixed! :)

ihhub: In the future this will be only mode because we want to add support of gamepads so we will have to make it as software emulation.

oleg-derevenetz: @ihhub hardware cursor may move smoother, especially on slow systems, because its movement doesn't depend on SDL event loop processing.

zenseii: Yes, I've just experienced that the software one is much slower and makes the game slow down frequently, on a slower PC.
It is up to you if we want to take this into account or not.

ihhub: Okay, I will revert this change but keep fullscreen changes.

ihhub: In this case we are back to situation when in fullscreen mode the cursor is too small :)

oleg-derevenetz: @ihhub BTW I noticed that on Windows FPS never exceeds 60, while on macOS it could be 200+, and on macOS software cursor works much better.

ihhub: @oleg-derevenetz , did you enable V-sync on Windows?

ihhub: Ah, you're right. It's very strange considering the fact that on SDL 1 it works absolutely fine.

ihhub: This is V-Sync related issue. You need to turn it off and restart the game.

ihhub: I found out that while toggling V-Sync option we do not set these flags in SDL itself.

oleg-derevenetz: Well, there is nothing surprising that 60 fps software cursor works worse. It seems that SDL1 just ignores `vsync` option and FPS may achieve 200+ even if it is turned on, which shouldn't be.

ihhub: SDL 1 doesn't support V-Sync :)

zenseii: > In this case we are back to situation when in fullscreen mode the cursor is too small :)

Yes, as I said I'm personally fine with this. I'm not sure how much of an impact it might have on weaker systems like even an Android phone. Then there's also PS Vita which I've seen players have concerns about the performance, but I have no way to test this not owning the system.

oleg-derevenetz: @zenseii Touch devices (such as Android and Vita) have software emulated mouse enabled in any case. Also there is no much of "impact" with touch because mouse cursor instantly moved to the touch point, while with "real" mouse there is a ton of movement events on the way from the previous mouse point to a new one.

zenseii: This is a good point, @oleg-derevenetz . I was under the impression that we wanted to enable software cursor to enable manual control of the cursor like for Android with an onscreen "joystick", which could also be done for systems like PS Vita with a physical joystick. In this case it might affect performance, right?

ihhub: Hi @zenseii and @oleg-derevenetz , how is the performance with V-Sync being off? I rebased this branch from master so it has V-Sync fix.

zenseii: With V-sync off there are no performance issues with software cursor. With V-sync on I occasionally see some slowdown when a hero is turning at a sharp angle on the Adventure map.

zenseii (approved):

src/fheroes2/system/settings.cpp
oleg-derevenetz: @zenseii

> I was under the impression that we wanted to enable software cursor to enable manual control of the cursor like for Android with an onscreen "joystick".

I hope not. According to feedback here in comments and discussions, people like the direct pointing by finger much more than virtual on-screen joystick. Maybe it's a little less accurate, but much more fun. Software cursor is needed for better control - for example, when you touch a tile on the adventure map to pave a path, the form of the software cursor (that will remain on this tile once you remove your finger) will suggest you how many days will the road take and whether there will be a battle.

@ihhub

> how is the performance with V-Sync being off

There was no performance impact when V-Sync was off before and there is no such impact now.

ihhub: I think we can bring back the original change for cursor software emulation. What do you think @oleg-derevenetz ?

oleg-derevenetz: @ihhub I'm not against it. Hopefully this won't cause problems on slow machines.

ihhub: Interesting, but what's the size of cursor in macOS without this mode while playing 640x480 in full screen?

ihhub:
One more approach I need to explore is resizing of cursor. I remember that for newer SDL version on Windows cursor was too big. I think it's somehow might be possible to change.

src/fheroes2/system/settings.cpp
oleg-derevenetz: @ihhub Unfortunately, I can't make a screenshot that contains the hardware cursor, at least by regular means (Shift+Cmd+...), but the cursor in the form of a horse (on the adventure map) is about 1/4 of hero's size on the screen. Not so bad.

ihhub:
Hi @oleg-derevenetz and @zenseii , I made an alternative approach which should cover most of cases without damaging performance on other systems. Please take a look.

src/engine/screen.cpp
```diff
@@ -755,13 +755,29 @@ namespace
 
             bool fullScreen = true;
             uint32_t flags = SDL_GetWindowFlags( _window );
+            static_assert( ( SDL_WINDOW_FULLSCREEN_DESKTOP & SDL_WINDOW_FULLSCREEN ) == SDL_WINDOW_FULLSCREEN && SDL_WINDOW_FULLSCREEN_DESKTOP > SDL_WINDOW_FULLSCREEN,
```
oleg-derevenetz, line 758: Hi @ihhub why just not check if either one of these flags is set, and if yes, then just clear both flags at once? In this case we shouldn't depend on the order of flag values as well as whether they intersect or not. Or is it not working for some reason?

ihhub: That's a goo point. I'll change the code.

oleg-derevenetz (approved):

Branikolog:
Hi, @ihhub !
This PR changes nothing regarding #6120 . :(
Also, I want to note that on my home PC on Win 8 fheroes2 couldn't be minimized (Alt+Tab) from fullscreen mode long time ago. Application just blinks, trying to show desktop, and then recovers the game again. I was hoping, this PR would fix the issue, but it's still on. My other PC with Win 10 minimizes ok though.

oleg-derevenetz:
Hi @Branikolog for the record: do you have software emulated cursor turned on (`cursor soft rendering = off` in `fheroes2.cfg`)? If not, then nothing was supposed to change.

Branikolog:
@oleg-derevenetz , hello.
Oh, I completely forgot about this tiny detail! ))
I've tested once again properly and now the issue with fullscreen applications is gone. (#6120)
But I've found one flaw in current implementation.
While switching to windowed mode and back to fullscreen game sometimes doesn't stretch on the whole screen:
image
To reproduce the bug you need to switch game resolution in fullscreen mode (It works for me, in a case I set fullHD, which is maximum my notebook supports) and then switch to windowed mode.
To be sure, the "bug mode" is enabled, try to resize window with mouse: in normal mode, while resizing, new window area appears as black:
image
In a "bug mode" this area reminds me old win 98 windows lags:
изображение_viber_2022-11-24_23-34-09-146
So, after pressing F4 a few times game stratches in a way as I showed on the first screenshot.

ihhub:
Hi @Branikolog , can you please be more specific what settings you set and how exactly you reproduce it? I don't understand the meaning of bug mode.

ihhub:
I am proceeding to merge this pull request as the mentioned by @Branikolog issue is not directly related to changes. Also the changes in this pull request do not add any regression since on Windows software cursor rendering is disabled by default.

ihhub:
@Branikolog , please file another issue describing in details the problem for 2 monitors using needed settings for the game.

Branikolog:
@ihhub , hello!

>Hi @Branikolog , can you please be more specific what settings you set and how exactly you reproduce it? I don't understand the meaning of bug mode.

Start engine - enter fullscreen mode - set fHD resolution (I suppose, it also could be maximum available one) - set windowed mode - try to resize window several times dragging window corner - try to switch between fullscreen and window modes several times (issue occurs not each time we switch modes).

PR #6233, opened 2022-11-20
Style nits

oleg-derevenetz:
* `LocalEvent`: make method naming more uniform;
* Move `isHandheldDevice()` from `core` to `system`: I think it's more suitable there (optional change).

### DIFF
--- a/src/engine/system.h
+++ b/src/engine/system.h
@@ -30,6 +30,8 @@
 
 namespace System
 {
+    bool isHandheldDevice();
+
     int MakeDirectory( const std::string & );
     std::string ConcatePath( const std::string &, const std::string & );
 
--- a/src/engine/core.cpp
+++ b/src/engine/core.cpp
@@ -181,13 +181,4 @@ namespace fheroes2
     {
         return isComponentInitializedInternally( component );
     }
-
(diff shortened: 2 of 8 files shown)


src/engine/system.h
```diff
@@ -30,6 +30,8 @@
 
 namespace System
 {
+    bool isHandheldDevice();
```
ihhub, line 33: Hi @oleg-derevenetz , this was my first thought but it looks like this file contains mostly filesystem specific code.

ihhub (commented):
Hi @oleg-derevenetz , I left just one comment. Could you please check it?

src/engine/system.h
oleg-derevenetz: @ihhub That's because this code is mostly system-specific (and also screen and localevent, but these are fits even worse :)). Core is mostly about initialization+deinitialization. However, as I already said, this is optional, I can revert it.

ihhub: I am fine keeping it as it is. Both places are not ideal.

ihhub:
@oleg-derevenetz , thank you so much for code cleanup!

PR #6234, opened 2022-11-20
Fix inability to toggle V-Sync dynamically

ihhub:
relates to #6229

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -887,7 +887,7 @@ namespace
                 }
 
                 const fheroes2::Display & display = fheroes2::Display::instance();
-                 _createRenderer( display.width(), display.height() );
+                _createRenderer( display.width(), display.height() );
             }
         }
 
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -873,6 +873,22 @@ namespace
         void setVSync( const bool enable ) override
         {
             _isVSyncEnabled = enable;
+
+            if ( _window != nullptr ) {
+                // We do not need to rebuild window but renderer only.
+                if ( _texture != nullptr ) {
+                    SDL_DestroyTexture( _texture );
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz ,

Outside the scope of this PR: I am thinking that for V-Sync mode we might need to issue `SDL_RenderPresent` calls in a separate thread or event skip if queue is too big. It should solve the problem with lagging cursor. What do you think?

oleg-derevenetz:
> I am thinking that for V-Sync mode we might need to issue SDL_RenderPresent calls in a separate thread or event skip if a frame queue is too big. It should solve the problem with lagging cursor. What do you think?

I don't know is it safe to call `SDL_RenderPresent` from the separate thread. There is nothing said in the documentation regarding this.

oleg-derevenetz (approved):

src/engine/screen.cpp
```diff
@@ -1194,6 +1169,59 @@ namespace
...
+                ERROR_LOG( "Failed to create a window renderer of " << width_ << " x " << height_ << " size. The error: " << SDL_GetError() )
+                clear();
+                return false;
+            }
+
+            if ( SDL_SetHint( SDL_HINT_RENDER_SCALE_QUALITY, "linear" ) == SDL_FALSE ) {
+                ERROR_LOG( "Failed to set a linear scale hint for rendering." )
+            }
+
+            // Setting this hint prevents the window to regain focus after loosing it in fullscreen mode.
```
zenseii, line 1198:
```suggestion
            // Setting this hint prevents the window to regain focus after losing it in fullscreen mode.
```

zenseii (changes requested):
Hi, @ihhub . LGTM. I only noticed one very minor typo.

zenseii (approved):

PR #6235, opened 2022-11-20
Fix multiple issues with auxiliary functions from the engine

oleg-derevenetz:
While I'm here:

1. `System::MakeDirectory()` returned 1 (`true`) on Windows, but 0 (`false`) on other systems upon successful directory creation;
2. `System::Unlink()` returned 0 upon successful file/directory removal, which was counterintuitive;
3. Renamed `ConcatePath()` to `concatPath()` (I've been dreaming of doing this for a long time);
4. Fixed an issue that caused `fheroes2.log` not to be created when the game was started for the first time;
5. `System::GetConfigDirectory()`, `System::GetDataDirectory()`: use the current directory as a fallback instead of an empty string. The previous behavior leads to attempts to create a log file, config file, etc. in the root directory, which is hopeless.
6. `GetDirname()` was not quite POSIX-compliant:

| Argument | Old `GetDirname()` | POSIX `dirname()` / new `GetDirname()` |
| - | - | - |
| (empty) | (empty) | `.` |
| `/usr/` | `./` | `/` |
| `/` | `./` | `/` |

7. `GetBasename()` was not quite POSIX-compliant:

| Argument | Old `GetBasename()` | POSIX `basename()` / new `GetBasename()` |
| - | - | - |
| (empty) | (empty) | `.` |
| `/usr/` | `/usr` | `usr` |

8. Removed command-line switches;
9. Fixed a bug on *NIX when game was unable to locate data files if it was started using the relative path (say, `game/fheroes2`). That's because previous version of `splitUnitPath()` was not able to handle the relative paths properly:

| Argument | Before | Now |
| - | - | - |
| `/foo/bar` | `foo`, `bar` | `foo`, `bar` |
| `foo/bar` | `bar` | `foo`, `bar` |

10. Various style nits.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -30,9 +30,21 @@
 #if defined( _WIN32 )
 #include <clocale>
 #include <map>
-#endif
 
-#include "system.h"
+#include <direct.h>
+#include <io.h>
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#else
+#include <dirent.h>
+#include <unistd.h>
+
+#if defined( TARGET_PS_VITA )
+#include <psp2/io/stat.h>
+#else
+#include <sys/stat.h>
+#endif
+#endif
 
 #if defined( _WIN32 ) || defined( ANDROID )
 #include "logging.h"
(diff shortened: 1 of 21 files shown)


ihhub:
Hi @oleg-derevenetz ,
> While I'm here:

this is disturbing...

oleg-derevenetz:
@ihhub

> this is disturbing...

Only absolutely necessary changes! :)

src/engine/system.cpp
```diff
@@ -354,53 +372,19 @@ bool System::IsDirectory( const std::string & name, bool writable )
...
+    return _unlink( path.c_str() ) == 0;
 #else
-    return unlink( file.c_str() );
+    return unlink( path.c_str() ) == 0;
 #endif
 }
 
 #if !defined( _WIN32 ) && !defined( ANDROID )
 // TODO: Android filesystem is case-sensitive so it should use the code below.
 //       However, in Android an application has access only to a specific path on the system.
```
oleg-derevenetz, line 386: BTW, Android filesystem is in fact case-insensitive, at least in those places that are accessible by users (`/sdcard`) and where game data files are placed. We don't need to use the workarounds for this.

ihhub: I noticed that it's case-sensitive within an internal storage.

oleg-derevenetz: But we do not place any files here whose case we do not control. The only files placed to the internal storage are timidity files and `resurrection.h2d`.

ihhub: Android is... a unique OS...

ihhub:
@oleg-derevenetz , thanks a lot for cleaning up this mess!

PR #6238, opened 2022-11-21
Remove the heroes: allow to choose any primary skill in Arena option

oleg-derevenetz:
Related to #6161


ihhub:

PR #6242, opened 2022-11-22
Use logarithmic scale of sound volume

### LINKED ISSUE #6230: Audio volume changes are not linear
### Preliminary checks

If you try to change audio volume (mostly music) from level 5 to 10 the difference is minimal but from 0 to 5 is huge. It looks like SDL mixer uses `1-exp(-x)` formula for sound. This is understandable since it is decibels.

We need to convert our 0..10 range appropriately to cover 0, 10, 20..100% volume levels based on the formula.

### Additional info

_No response_

undef21:
close #6230

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -626,7 +626,7 @@ namespace
             return MIX_MAX_VOLUME;
         }
 
-        return ( std::exp( std::log( 10 + 1 ) * volumePercentage / 100 ) - 1 ) / 10 * MIX_MAX_VOLUME;
+        return ( static_cast<int>( std::exp( std::log( 10 + 1 ) * volumePercentage / 100 ) - 1 ) / 10 * MIX_MAX_VOLUME );
     }
 }
 
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -22,7 +22,6 @@
  ***************************************************************************/
 
 #include <algorithm>
-#include <array>
 #include <atomic>
 #include <cassert>
 #include <cmath>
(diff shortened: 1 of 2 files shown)


undef21:
@ihhub, @Branikolog could you check sound levels in this PR? I changed music level in the options dialog in the main menu. Please use external music.

Branikolog:
Hi, @undef21 Actually, I cannot hear much difference comparing with current version.

undef21:
@Branikolog could you check this PR again?

Branikolog:
Hi, @undef21 
Now's much better!
I only hear sometimes a short click sound (something similar to pluging headphones into 3.5 jack) , while increasing volume to the higher values, but I'm not sure, it's not my local problem of OS or notebook speakers. I'll check tomorow on my other PC.

undef21:
> I'm not sure, it's not my local problem of OS or notebook speakers.

Also check this in the master branch please.

Branikolog:
Hi, @undef21 .
On my other PC everything seems to be OK. Probably my home PC speakerphones seem to be a bit overloaded on high volume setting, that's why I can hear sometimes a little noise while rising it up.

undef21:
Hello @Branikolog. Could you listen this PR again?

Branikolog:
Hi, @undef21 !
Sorry, for late response!
Everything looks fine!

src/engine/audio.cpp
```diff
@@ -625,12 +625,7 @@ namespace
             return MIX_MAX_VOLUME;
         }
 
-        return volumePercentage * MIX_MAX_VOLUME / 100;
-    }
-
-    int normalizeFromSDLVolume( const int volume )
-    {
-        return volume * 100 / MIX_MAX_VOLUME;
+        return static_cast<int>( ( std::exp( std::log( 10 + 1 ) * volumePercentage / 100 ) - 1 ) / 10 * MIX_MAX_VOLUME );
```
ihhub, line 628: Could you please explain why we use a logarithm of 11 instead of 10? Also this formula needs a bit of explanation.

ihhub (changes requested):

src/engine/audio.cpp
undef21: Hello, @ihhub. The good explanation is [here](https://www.dr-lex.be/info-stuff/volumecontrols.html#ideal2). I used that article.

Resulting curve is:
```
y = ae^(bx) + c
a = 1/10
c = -1/10
b = ln(1 + 1/a) = ln 11
```

ihhub: I would suggest to at least add a comment with this link for any possible future questions.

undef21: Done.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@undef21 , thank you so much for this great improvement!

Issue #6174, opened 2022-11-13
Describe the installation process on Android


oleg-derevenetz:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Since now we have Android release (at least PoC),  we need to encourage users to test it.

### Describe the possible solution

It would be good to describe the installation process on Android in the `docs/ANDROID.md`, put it to the Android distribution ZIP file and add the corresponding entry to the `INSTALL.md`, like it's already done for PSV and Switch.

Mr-Bajs:
Great, I'm about trying to try it out on a Samsung Tablet A8 running Android 12. The readme file from the Android.zip, really don't explain where to put the DATA, MAPS, MUSIC etc folders.

Getting fheroes2 on F-droid would be great to.

oleg-derevenetz:
@Mr-Bajs long story short, they should be put into the `Android/data/org.fheroes2/files` on your flash card. If you have already tried running fheroes2, then this directory should even already exist.

Mr-Bajs:
I figured it out, thanks anyway.
I'll give it a try. I'm pretty new to all this GitHub.

zenseii:
> I figured it out, thanks anyway. I'll give it a try. I'm pretty new to all this GitHub.

BTW if you have any questions about how to address the comments I made in the Swedish translation update don't hesitate to ask.

zenseii:
I'll post PR with this in a moment.

PR #6243, opened 2022-11-22
Add Android installation and usage documentation

zenseii:
This adds an Android installation guide to the repo, steps in the installation wiki and bundles the guide in the Android zip archive.

Close https://github.com/ihhub/fheroes2/issues/6174

### DIFF
--- a/docs/README_android.md
+++ b/docs/README_android.md
@@ -6,14 +6,15 @@ do this.
 
 fheroes2 requires the data files from the original Heroes of Might and Magic II.
 
-Copy the the contents of the original game's "DATA" folder into the "internalStorage:Android/data/org.fheroes2/files/data/"
-folder and everything from the "MAPS" folder into "internalStorage:Android/data/org.fheroes2/files/maps/".
+From your original Heroes 2 installation copy the directories named "DATA", "MAPS", "ANIM" and "MUSIC" into the
+"internalStorage:Android/data/org.fheroes2/files/" folder.
 
-The music files should be placed in the "internalStorage:Android/data/org.fheroes2/files/music/" folder.
+Depending on your version of Heroes 2, the music files that need to be placed in the "MUSIC" directory can either be on your
+original CD or in the installation folder (GOG and Ubisoft).
 
-fheroes2 supports the ingame cinematics. This is necessary for full campaign support. To play the intro and the other
-videos, make sure that all of the "*.SMK" files are placed in the "internalStorage:Android/data/org.fheroes2/files/anim/"
-folder.
+Ingame cinematics are supported by fheroes2 and for these to be supported it is necessary to have them in the "ANIM" folder.
+Depending on your original version of Heroes 2 they can either be found on your CD, disc image or in a folder named "anim" in
+your Heroes 2 installation directory.
 
 ## Controls
 To simulate a right-click to get info on various items, you need to first touch and keep touching anywhere on the screen
--- a/docs/README_android.md
+++ b/docs/README_android.md
@@ -16,8 +16,10 @@ videos, make sure that all of the "*.SMK" files are placed in the "internalStora
 folder.
 
 ## Controls
-To simulate a right click to get info on various items, you need to first touch and keep touching anywhere on the screen
-and then touch on the item of interest. You can slide away while maintaining the touch to keep viewing the info on the item.
+To simulate a right-click to get info on various items, you need to first touch and keep touching anywhere on the screen
+and then touch on the item of interest. You can slide the touch away while maintaining it to keep viewing the info on the
+item. It is recommended to place the first touch outside the rendered game area to avoid performing a left-click after
+the simulated right-click is ended.
 
 By default normal adventure map scrolling on the borders of the screen is disabled. To pane the viewing area around you
 need to press anywhere on the adventure map and slide around to change where you are viewing.
--- a/docs/README_android.md
+++ b/docs/README_android.md
@@ -6,11 +6,10 @@ do this.
 
 fheroes2 requires the data files from the original Heroes of Might and Magic II.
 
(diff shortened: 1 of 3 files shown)


docs/README_android.md
```diff
@@ -0,0 +1,33 @@
...
+
+By default normal adventure map scrolling on the borders of the screen is disabled. To pane the viewing area around you
+need to press anywhere on the adventure map and slide around to change where you are viewing.
+
+Note that text input is not yet implemented.
+
+## Android specific notes
+
+Pointer movement speed can be changed with 'gamepad_pointer_speed' parameter in "ux0:data/fheroes2/fheroes2.cfg".
+
```
zenseii, line 31: Not sure how that managed to sneak in there.
```suggestion
```

docs/README_android.md
```diff
@@ -0,0 +1,30 @@
+# Android port of [**fheroes2**](README.md) project
+
+## Installation
+Install the fheroes2.apk to your device running Android. You need to have set the necessary permissions on your device to
+do this.
+
+fheroes2 requires the data files from the original Heroes of Might and Magic II.
+
+Copy HEROES2.AGG and HEROES2X.AGG (if you own the Price of Loyalty expansion) from the original game's "DATA" folder into
```
oleg-derevenetz, line 9: AFAIK, these are not the only files that need to be copied. There are also some SMK files here that are needed for campaign, for example, `IVYDES.SMK`. IMHO it is better to recommend copying the entire contents of the `DATA` folder.

docs/README_android.md
```diff
@@ -0,0 +1,30 @@
...
+"internalStorage:Android/data/org.fheroes2/files/data/" and everything from "MAPS" folder into
+"internalStorage:Android/data/org.fheroes2/files/maps/".
+
+The music files should be placed in the "internalStorage:Android/data/org.fheroes2/files/maps/" folder.
```
oleg-derevenetz, line 13: Do you mean `.../files/music/` folder?

docs/README_android.md
```diff
@@ -0,0 +1,30 @@
...
+
+fheroes2 supports the ingame cinematics. This is necessary for full campaign support. To play the intro and the other
+videos, make sure that all of the "*.SMK" files are placed in the "internalStorage:Android/data/org.fheroes2/files/anim/"
+folder.
+
+## Controls
+To simulate a right click to get info on various items, you need to first touch and keep touching anywhere on the screen
```
oleg-derevenetz, line 20: "anywhere on the screen outside the game window"... or something like that. I admit that "game window" doesn't fit well.

oleg-derevenetz (changes requested):
Hi @zenseii I just left a few comments, please take a look when it's convenient.

docs/README_android.md
zenseii: I agree that my original phrasing could do with improvement. By saying "anywhere on the screen" I'm trying to include the area of the screen that has the game rendered to it, because this works too, as well as a potential border for lower-than-native resolutions like 640x480.

zenseii: You are right that this could happen. It can be avoided if you first raise the initial finger you pressed down. We should also remember if someone uses the device's full resolution then there is no area outside game window.

oleg-derevenetz: > We should also remember if someone uses the device's full resolution then there is no area outside game window.

When using full resolution on a handheld device, all objects will be tiny. Also aspect ratio on phones is non-standard, so usually there are black bars on the sides.

zenseii: Full native resolution on phones is mostly useless right now, unless you mirror/cast or connect your phone to a separate monitor for playing fheroes2 (which is a good idea btw), Aside from this there are also Android tablets with larger screens, and native resolution would be relevant for them.

oleg-derevenetz: It would be desirable to at least recommend to place the first finger outside the game area, if possible, to avoid false "clicks".

zenseii: Yes, this is a good point. I'll try to come up with something that fits the various scenarios.

docs/README_android.md
```diff
@@ -0,0 +1,31 @@
...
+The music files should be placed in the "internalStorage:Android/data/org.fheroes2/files/music/" folder.
+
+fheroes2 supports the ingame cinematics. This is necessary for full campaign support. To play the intro and the other
+videos, make sure that all of the "*.SMK" files are placed in the "internalStorage:Android/data/org.fheroes2/files/anim/"
+folder.
+
+## Controls
+To simulate a right-click to get info on various items, you need to first touch and keep touching anywhere on the screen
+and then touch on the item of interest. You can slide the touch away while maintaining it to keep viewing the info on the
+item. It is recommended to place the first touch outside the rendered game area to avoid performing a left-click after
```
oleg-derevenetz, line 21: How about "... to avoid accidental left-click ..."?

oleg-derevenetz (approved):

docs/README_android.md
```diff
@@ -0,0 +1,31 @@
+# Android port of [**fheroes2**](README.md) project
+
+## Installation
+Install the fheroes2.apk to your device running Android. You need to have set the necessary permissions on your device to
+do this.
+
+fheroes2 requires the data files from the original Heroes of Might and Magic II.
+
+Copy the the contents of the original game's "DATA" folder into the "internalStorage:Android/data/org.fheroes2/files/data/"
```
ihhub, line 9: How about "Copy data, music, maps and anim directories into ..."? I think this is much easier to do then creating folders and copying files. What do you think?

zenseii: Yes, this is better.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:

PR #5983, opened 2022-10-15
Scenario Window UI Items position fix

HaveFun91:
Close #1250

PR #6244, opened 2022-11-22
iwyu.map: add new private header mappings for newer Ubuntu image version, speed up IWYU workflow a bit

oleg-derevenetz:
Fix migration of Ubuntu 22.04 image from 20221027.1 to 20221119.2 with the newer compiler.

See [this run](https://github.com/ihhub/fheroes2/actions/runs/3525606072/jobs/5912534513) and [this comment](https://github.com/ihhub/fheroes2/pull/5983#issuecomment-1324094001) for details.

### DIFF
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -10,16 +10,18 @@ jobs:
     timeout-minutes: 30
     steps:
     - uses: actions/checkout@v3
-    - name: Install dependencies
+    - name: Install dependencies and iwyu
       run: |
         sudo apt-get -y update
-        sudo apt-get -y install libsdl2-dev libsdl2-mixer-dev libsdl2-image-dev gettext ninja-build iwyu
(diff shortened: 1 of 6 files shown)


ihhub:
Hi @oleg-derevenetz , mostly likely the answer is no but is there a way to cache IWYU to improve performance?

oleg-derevenetz:
@ihhub you see, IWYU doesn't have any intermediate database. It just parses the sources and included headers using LLVM backend. But I need to think about that.

oleg-derevenetz:
@ihhub I reworked the workflow a bit to use the `iwyu_tool` script bundled with IWYU instead of `cmake`. The difference is `cmake` also runs compiler along with `include-what-you-use` to perform all build stages (if it didn't do this, the project would only be partially built, because there would be no object files and therefore nothing to link), while `iwyu_tool` runs just IWYU itself using `compile_commands.json` generated by CMake on the project generation stage. It became faster but not much - about 1/3 maybe.

ihhub:
@oleg-derevenetz , thank you very much for this improvement!

PR #6248, opened 2022-11-23
Add more Cyrillic button letters

zenseii:
This PR adds all Cyrillic Russian letters that were missing.

In other words:
Cyrillic CHE : Ч
Cyrillic PE : П
Cyrillic EF : Ф
Cyrillic SHA : Ш
Cyrillic SHCHA : Щ
Cyrillic E : Э

In addition 3 was fixed and A got proper spacing to nearby letters.

I used a pangram to show all letters at the same time in a more natural way than just listing them.
image


Branikolog:
Hi, @zenseii !
I cannot test the PR because there're no such letters in UI. :)
Could you, please, add here screenshots of all fixes, you've made after the first post?
I think "Ф" looks a little too small. I believe Can we make it a bit rounder? Like that:
Ef
(If you haven't already made that, for sure. :) )

Branikolog:
@zenseii 
I also think, that П could be 1 pixel thinner, since now it's even wider than Ч.
My suggestions are based not only on correct visual perception of letters, but on our future integration them into UI. If both variants are OK, I'd prefer the one, that takes less space, since there are going to appear cases, where we need to add more symbols on existing buttons.

zenseii:
@Branikolog, thanks for the input! The image you see in the starting post is up to date with my latest changes. I'll adjust the letters according to what you wrote.

As for checking in the game I basically change the Russian translation file's "BATTLE ONLY" string to whatever letters I want to see and then save it and overwrite the ru.mo.

I can upload such a change to this PR (and remove before merging) and then you can grab an artifact from here to test with.

Districh-ru:
Hi @zenseii!
Maybe the bottom part of letter 'З' should be one pixel lower:
изображение :   изображение   -->>  изображение

zenseii:
> Hi @zenseii! Maybe the bottom part of letter 'З' should be one pixel lower: изображение : изображение -->> изображение

Thanks for reminding me about this, @Districh-ru. I had noticed it myself but completely forgotten about it.

ihhub:
Hi @zenseii , can you please resolve the merge conflict and also tell me what's the status of this pull request?

Branikolog:
Hi, @zenseii !
Everything regarding Cyrillic letters is ok for me. 
I'm only concerning about Latin newly added A letter. Most of Latin letters have flat decorations size of 2px to each side of the symbol leg, while now it has 1px wide. I'm not sure it suits the style of other letters now.
A a

ihhub:
Hi @zenseii , I prefer the closest to the original which is the second one from the top. There's slight difference in shading for our letters which I know so after fixing it we might achieve better results.

Branikolog:
Hi, @ihhub , @zenseii !
I personally prefer the bottom one, as it has a little more space inside of upper part of the letter, which is, in my personal point, closer to the original, than the second one. Regarding legs, I'm ok with both these variants, which are wider, that 3rd. But anyway, it's all up to you.

Branikolog:
Hi, @zenseii and @ihhub !
Anyway, it's much better, than the original and is definitely a step forward.
You know, the original A letter is totally a mess and and even current implementation is much better:
ezgif com-gif-maker (93)
I believe, with future corrections (something I mentioned at https://github.com/ihhub/fheroes2/pull/6194#issuecomment-1337224782) it could even become much better and more pleasant to eye.

Branikolog (approved):

ihhub (approved):

ihhub:
@zenseii , thank you so much for this great improvement!

PR #6056, opened 2022-10-28
Consolidate French and CP1252 font generation

zenseii:
This PR makes the French font generation use the CP1252 font generation and maintains compatibility with the original French custom encoding.

All French specific letters have been added as well as all characters with the tilde `~` accent for Portuguese support.

The MO makefile has been changed to use CP1252 encoding for French.

French has been set to use CP1252 for button font.

This PR will also add French quotation marks since they are part of CP1252.
Close #4882

PR #6249, opened 2022-11-23
Fix French and Italian characters not being shown with their assets

zenseii:
Due to changes from this PR https://github.com/ihhub/fheroes2/pull/6056 when you have the French original assets you will not get the French special characters.

In addition while I was at it I added support for Italian characters too.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -881,7 +881,7 @@ namespace fheroes2
                 }
 
                 // Polish version uses CP1251
-                if ( crc32 == ( 0xE9EC7A63 || 0x88774771 ) ) {
+                if ( crc32 == 0xE9EC7A63 || crc32 == 0x88774771 ) {
                     // Engine expects that letter indexes correspond to charcode - 0x20.
                     // In case CP1251 font.icn contains sprites for chars 0x20-0x7F, 0xC0-0xDF, 0xA8, 0xE0-0xFF, 0xB8 (in that order).
                     // We rearrange sprites array for corresponding sprite indexes to charcode - 0x20.
@@ -892,7 +892,7 @@ namespace fheroes2
                     imageArray.erase( imageArray.begin() + 192 );
                 }
                 // German version uses CP1252
-                if ( crc32 == ( 0x04745D1D || 0xD0F0D852 ) ) {
+                if ( crc32 == 0x04745D1D || crc32 == 0xD0F0D852 ) {
                     imageArray.insert( imageArray.begin() + 96, 124, imageArray[0] );
                     std::swap( imageArray[164], imageArray[224] );
                     std::swap( imageArray[182], imageArray[225] );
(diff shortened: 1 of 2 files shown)


zenseii:
Hi, @ihhub.

I'm marking this as ready for review just to start the discussion. I've shown one way of solving this, but the problem is that the French original assets miss many special characters that we on the other hand generate. Optimally we'd have access to these characters when the French assets are used too, but that isn't possible right now.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -904,6 +904,26 @@ namespace fheroes2
                     std::swap( imageArray[220], imageArray[222] );
                     imageArray.erase( imageArray.begin() + 221, imageArray.end() );
                 }
+                // French version has its own special encoding but should conform to CP1252 too
+                verifiedFontWidth = ( id == ICN::FONT ) ? 9 : 5;
```
ihhub, line 908: Should we use CRC32 value instead of some tricks like this? I think we shouldn't support something which was modified outside the original assets.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -904,6 +904,26 @@ namespace fheroes2
...
+                if ( imageArray.size() == 96 && imageArray[3].width() == verifiedFontWidth ) {
+                    imageArray.insert( imageArray.begin() + 96, 160 - 32, imageArray[0] );
+                    imageArray[244 - 32] = imageArray[3];
```
ihhub, line 911: I remember that one of characters which is used for a full stop somehow contains another character. We should fix it and use the full stop character.

ihhub (commented):
Hi @zenseii , the proposed approach is fine with me. I left 2 questions here. Could you please take a look?

src/fheroes2/agg/agg_image.cpp
zenseii: I could not find this full stop being replaced.

Here are my notes:
```
These are present in `font` icn `àâçéèêîîîïïôùû`

ô = empty = comes after " == # in ascii
û = $
ù = &
â = empty = comes after ) == * in ascii
ï = empty = comes after ; == < in ascii
î = empty = comes after = == > in ascii
à = empty = comes after ? == @ in ascii
ç = empty = comes after ] == ^ in ascii
è = empty = comes after - before a == ` in ascii
ï (91) = empty = { in ascii
ê (92) = empty = | in ascii
î (93) = empty = } in ascii
é (94) = empty = ~ in ascii
î (95) = empty = € in ascii

Need to map these to nearest Latin too:
ÇÉÊÀ
```

In other words the only symbols that are replaced compared to English font file, or any other language, are the # and & signs. These aren't found anywhere else so for French they should just be considered missing. It's not a big problem.

ihhub: Hi @zenseii , this is the description in English version of the game:
image

and this is in French (using the assets):
image

Symbol **&** is displayed inproperly.

ihhub: I think we can address it separately.

ihhub (approved):

ihhub:
@zenseii , thanks a lot for the fix!

We might also think about a common storage for crc32 as we use them in several places.

PR #6250, opened 2022-11-24
Java: use final where appropriate

oleg-derevenetz:
Not as useful as `const` in C++, but won't hurt. Plus make `MainActivity` the `final class` as well.


ihhub:
@oleg-derevenetz , thanks you very much for this small improvement!

PR #6256, opened 2022-11-26
Add shadowing for fog in debug build

### LINKED ISSUE #5945: Debug mode movement pointer range
### Preliminary checks

### Platform

Linux

### Describe the bug

In debug mode I can see the whole map, but the mouse pointer only turns to movement (the horse) within the short range, supposedly in the area what should be seen by the hero w/o debugging.  There are no obstacles that could prevent movement.

Within the range:
image

Outside of the range:
image

### Save file

[Good_1_0001.zip](https://github.com/ihhub/fheroes2/files/9741084/Good_1_0001.zip)

### Additional info

`debug = 65535`

ihhub:
close #5945

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -342,11 +342,11 @@ namespace
 
     const fheroes2::Image & getDebugFogImage()
     {
-        static fheroes2::Image fog;
-        if ( fog.empty() ) {
-            fog.resize( 32, 32 );
-            fheroes2::FillTransform( fog, 0, 0, fog.width(), fog.height(), 2 );
-        }
+        static const fheroes2::Image fog = []() {
+            fheroes2::Image temp( 32, 32 );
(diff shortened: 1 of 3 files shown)


src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -339,6 +339,17 @@ namespace
 
         return imageMap.try_emplace( passable, std::move( sf ) ).first->second;
     }
+
+    const fheroes2::Image & getDebugFogImage()
+    {
+        static fheroes2::Image fog;
```
oleg-derevenetz, line 345: Hi @ihhub I propose to initialize the static variable in such cases by using a lambda. First of all, such initializations are truly atomic, which can be useful (maybe not in this particular case, but in general). Second, you will not need to check a flag (such as `empty()` each time). Third, you will be able to declare `fog` as `const`.

ihhub: You're right! I will do it.

oleg-derevenetz (approved):

PR #6257, opened 2022-11-26
An attempt to recreate the original rainbow animation (battlefield)

Districh-ru:
This PR changes the way the rainbow is drawn for the "good luck" event on the battlefield:

https://user-images.githubusercontent.com/113276641/204090850-53f04cc6-e391-43a2-bb5f-36271947d441.mp4

(In this video "good luck" is force set to 'true')

изображение

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -556,7 +556,7 @@ fheroes2::Image DrawRainbow( const std::vector<int32_t> & rainbowArc, const int3
     std::vector<int32_t>::const_iterator pnt = rainbowArc.begin();
 
     // Get the original good luck sprite, since it has a rainbow image which will be used to get line.
-    fheroes2::Sprite luckSprite = fheroes2::AGG::GetICN( ICN::EXPMRL, 0 );
+    fheroes2::Image luckSprite = fheroes2::AGG::GetICN( ICN::EXPMRL, 0 );
     
     // Declare an image for a single rainbow line from the center of the luckSprite.
     fheroes2::Image rainbowLine;
@@ -3938,7 +3938,8 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
             }
             rainbowDescend = std::max( 1, static_cast<int32_t>( 0.0342 * rainbowLength - 4.868 ) );
             rainbowTop = static_cast<int32_t>( 0.8524 * rainbowLength + 17.7 );
-            drawOffset = isRainbowFromRight ? ( rainbowDescendPoint.x - rainbowDescend - rainbowThickness / 2 ) : ( rainbowDescendPoint.x - rainbowDescend - rainbowAscend );
+            drawOffset
+                = isRainbowFromRight ? ( rainbowDescendPoint.x - rainbowDescend - rainbowThickness / 2 ) : ( rainbowDescendPoint.x - rainbowDescend - rainbowAscend );
         }
         else {
             pow2 = 0;
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -523,30 +523,32 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
     return sf;
 }
 
-void GetHalfArc( std::vector<double> & arc, const int32_t width, const int32_t height, const int32_t pow1, const int32_t pow2, const double pow2Ratio )
+void GetHalfArc( std::vector<double> & arc, int32_t width, const int32_t height, const int32_t pow1, const int32_t pow2, const double pow2Ratio )
 {
-    // for positive width arc will start from y = 0, for negative it will end at y = 0
+    // For positive width the arc will start from y = 0, for negative it will end at y = 0.
     const int32_t x0 = ( width < 0 ) ? width : 0;
 
-    // for positive height arc top will be y = 0, for negative - arc will be flipped for y coordinate (todo: consider height offset from previous arc part)
+    // For positive height the arc top will be y = 0, for negative - arc will be flipped for 'y' coordinate. (todo: consider height offset from previous arc part)
     const int32_t y0 = ( height < 0 ) ? height : 0;
 
-    // coefficients for multipliers of 'pow1' and 'pow2' degrees
+    // Coefficients for multipliers of 'pow1' and 'pow2' degrees.
     const double k1 = ( height * ( 1 - pow2Ratio ) ) / std::pow( width, pow1 );
     const double k2 = ( height * pow2Ratio ) / std::pow( width, pow2 );
 
-    // calculate 'y' coordinates for arc: y = k1*(x-x0)^pow1+k2*(x-x0)^pow2 and push it to 'arc' vector
-    for ( int x = 0; x < std::abs( width ); ++x ) {
-        double y = k1 * std::pow( ( x + x0 ), pow1 ) + k2 * std::pow( ( x + x0 ), pow2 ) - y0;
-        arc.push_back( y );
+    width = std::abs( width );
+    // Calculate 'y' coordinates for the arc: y = k1*(x-x0)^pow1+k2*(x-x0)^pow2 and push it to the 'arc' vector.
+    for ( int32_t x = 0; x < width; ++x ) {
+        arc.push_back( k1 * std::pow( ( x + x0 ), pow1 ) + k2 * std::pow( ( x + x0 ), pow2 ) - y0 );
     }
 }
 
-fheroes2::Image DrawRainbow( const std::vector<double> & rainbowArc, const int32_t & rainbowThickness, bool isVertical = false )
+fheroes2::Image DrawRainbow( const std::vector<double> & rainbowArc, const int32_t rainbowThickness, const bool isVertical, const bool flipHorizontally )
 {
-    // rainbow image size should include arc size plus thickness of the rainbow
+    // Rainbow image size should include the arc size plus the thickness of the rainbow.
     const int32_t rainbowWidth = static_cast<int32_t>( rainbowArc.size() );
     const int32_t rainbowHeight = static_cast<int32_t>( *std::max_element( rainbowArc.begin(), rainbowArc.end() ) ) + rainbowThickness;
+
+    // If the rainbow is vertical - swap width and height.
     const int32_t rainbowImgWidth = isVertical ? rainbowHeight : rainbowWidth;
     const int32_t rainbowImgHeight = isVertical ? rainbowWidth : rainbowHeight;
     fheroes2::Image rainbow( rainbowImgWidth, rainbowImgHeight );
@@ -557,20 +559,23 @@ fheroes2::Image DrawRainbow( const std::vector<double> & rainbowArc, const int32
     uint8_t colorGreen = 0;
     uint8_t colorBlue = 0;
 
-    // draw rainbow image for each 'x' coordinate
+    // Color normalization factor.
+    const double colorNorm = 256 * 4.6 / rainbowThickness;
+
+    // Draw a rainbow image for each 'x' coordinate.
     for ( int32_t x = 0; pnt != rainbowArc.end(); ++x, ++pnt ) {
         // fractional part of 'y' arc coordinate for better color reproduction
         const double frac = *pnt - std::floor( *pnt ) - 0.5;
 
-        // rainbow colors generation from: https://stackoverflow.com/questions/40629345/fill-array-dynamicly-with-gradient-color-c
+        // Rainbow colors generation from: https://stackoverflow.com/questions/40629345/fill-array-dynamicly-with-gradient-color-c
         for ( int i = 0; i < rainbowThickness; ++i ) {
-            // normalize 'x' coordinate to 4.6 (for 'fheroes2' rainbow colors) regions, each region is 256 units long
-            const int32_t normalized = static_cast<int32_t>( ( i - frac ) * 256 * 4.65 / rainbowThickness );
+            // Normalize 'x' coordinate to 4.6 (for 'fheroes2' rainbow colors) regions, each region is 256 units long.
+            const int32_t normalized = static_cast<int32_t>( ( i - frac ) * colorNorm );
 
-            // modificator for the color inside the closest region
+            // Modificator for the color inside the closest region.
             const uint8_t colorMod = ( normalized < 0 ) ? 0 : normalized % 256;
 
-            // find the region and set its colors
+            // Find the color region and set its colors.
             switch ( normalized / 256 ) {
             case 0:
                 colorRed = 255;
@@ -597,15 +602,21 @@ fheroes2::Image DrawRainbow( const std::vector<double> & rainbowArc, const int32
                 colorGreen = 0;
                 colorBlue = 255 - colorMod;
                 break;
+            default:
+                colorRed = 0;
+                colorGreen = 0;
+                colorBlue = 0;
+                break;
             }
 
-            if ( isVertical ) {
-                // for a vertical rainbow swap the 'x' and 'y' coordinates and flip horizontally
-                fheroes2::SetPixel( rainbow, rainbowImgWidth - i - static_cast<int32_t>( *pnt ), x, fheroes2::GetColorId( colorRed, colorGreen, colorBlue ) );
-            }
-            else {
-                fheroes2::SetPixel( rainbow, x, i + static_cast<int32_t>( *pnt ), fheroes2::GetColorId( colorRed, colorGreen, colorBlue ) );
-            }
+            // Calculate the 'y' coordinate for the current rainbow color.
+            const int32_t y = i + static_cast<int32_t>( *pnt );
+
+            // Set the 'x' and 'y' coordinates of the current rainbow pixel in the resulting rainbow image according to the rainbow direction.
+            const int32_t imgX = isVertical ? ( flipHorizontally ? y : rainbowImgWidth - y ) : ( flipHorizontally ? rainbowImgWidth - x : x );
+            const int32_t imgY = isVertical ? x : y;
+
+            fheroes2::SetPixel( rainbow, imgX, imgY, fheroes2::GetColorId( colorRed, colorGreen, colorBlue ) );
         }
     }
     return rainbow;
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
I've always been a little concerned about the duration of this animation. If the unit is located unsuccessfully, then the rainbow can be drawn for a long time :)

Districh-ru:
> I've always been a little concerned about the duration of this animation. If the unit is poorly positioned, then the rainbow can be drawn for a long time :)

Hi @oleg-derevenetz, I was also concerned about it and in the beginning the animation step was `const int32_t drawStep = borderDistance / 30;`. But later I changed it to a constant to reproduce the original game behavior.

@Branikolog, @LeHerosInconnu, what do you think: rainbow should have constant animation step or constant animation time?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+    uint8_t colorBlue = 0;
+
+    // draw rainbow image for each 'x' coordinate
+    for ( int32_t x = 0; pnt != rainbowArc.end(); ++x, ++pnt ) {
+        // fractional part of 'y' arc coordinate for better color reproduction
+        const double frac = *pnt - std::floor( *pnt ) - 0.5;
+
+        // rainbow colors generation from: https://stackoverflow.com/questions/40629345/fill-array-dynamicly-with-gradient-color-c
+        for ( int i = 0; i < rainbowThickness; ++i ) {
+            // normalize 'x' coordinate to 4.6 (for 'fheroes2' rainbow colors) regions, each region is 256 units long
```
ihhub, line 567: The comment says 4.6 but in the code it is 4.65. Please correct.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
     return sf;
 }
 
+void GetHalfArc( std::vector<double> & arc, const int32_t width, const int32_t height, const int32_t pow1, const int32_t pow2, const double pow2Ratio )
+{
+    // for positive width arc will start from y = 0, for negative it will end at y = 0
```
ihhub, line 528: It is optional but having a capital letter at the beginning of the comments and a full stop make them slightly better to read :)

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+
+    // for positive height arc top will be y = 0, for negative - arc will be flipped for y coordinate (todo: consider height offset from previous arc part)
+    const int32_t y0 = ( height < 0 ) ? height : 0;
+
+    // coefficients for multipliers of 'pow1' and 'pow2' degrees
+    const double k1 = ( height * ( 1 - pow2Ratio ) ) / std::pow( width, pow1 );
+    const double k2 = ( height * pow2Ratio ) / std::pow( width, pow2 );
+
+    // calculate 'y' coordinates for arc: y = k1*(x-x0)^pow1+k2*(x-x0)^pow2 and push it to 'arc' vector
+    for ( int x = 0; x < std::abs( width ); ++x ) {
```
ihhub, line 539: To avoid unnecessary computations by calling `std::abs()` please accept this variable without const and then reassign it using this function.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+        double y = k1 * std::pow( ( x + x0 ), pow1 ) + k2 * std::pow( ( x + x0 ), pow2 ) - y0;
+        arc.push_back( y );
+    }
+}
+
+fheroes2::Image DrawRainbow( const std::vector<double> & rainbowArc, const int32_t & rainbowThickness, bool isVertical = false )
```
ihhub, line 545: Please remove the default parameter since all usages of this function provide 3 parameters.

src/fheroes2/battle/battle_interface.cpp
ihhub, line 539: Please use `int32_t` instead of `int` to be consistent with types.

src/fheroes2/battle/battle_interface.cpp
ihhub, line 545: Please remove reference symbol from `rainbowThickness` as passing POD is always recommended by value.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3825,32 +3913,102 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
+        // the coordinate where the rainbow arc changes its direction
+        int32_t rainbowTop;
+        // offset from zero coordinates of battlefield of rainbow image to 'fall' onto the 'lucky' creature
+        int32_t drawOffset;
+        double pow2ratio;
+        double pow4ratio;
+
+        // set rainbow generation parameters
+        if ( isVerticalRainbow ) {
+            pow1 = 2, pow2 = 10, pow3 = 2, pow4 = 2;
```
ihhub, line 3958: Please do not use multiple operations at single line. Always put one operation on a separate line.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3825,32 +3913,102 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
+        GetHalfArc( rainbowArc, -rainbowArcBegin.width, rainbowArcBegin.height, pow1, pow2, pow2ratio );
+        GetHalfArc( rainbowArc, rainbowArcEnd.width, rainbowArcEnd.height, pow3, pow4, pow4ratio );
 
-        int width = 2;
-        fheroes2::Rect src( 0, 0, width, luckSprite.height() );
-        src.x = ( luckSprite.width() - src.width ) / 2;
-        int y = pos.y + pos.height - unitSprite.height() - src.height;
-        if ( y < 0 )
-            y = 0;
+        fheroes2::Image luckSprite = fheroes2::Flip( DrawRainbow( rainbowArc, rainbowThickness, isVerticalRainbow ), isRainbowFromRight, false );
```
ihhub, line 3983: Here we create 2 images. I suggest to do this way:
```cpp
fheroes2::Image luckSprite;
if ( isRainbowFromRight ) {
    luckSprite = fheroes2::Flip( DrawRainbow( rainbowArc, rainbowThickness, isVerticalRainbow ), true, false );
}
else {
    luckSprite = DrawRainbow( rainbowArc, rainbowThickness, isVerticalRainbow );
}
```
This code is bit longer but for cases when `isRainbowFromRight` is **false** it works much faster.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+
+    // draw rainbow image for each 'x' coordinate
+    for ( int32_t x = 0; pnt != rainbowArc.end(); ++x, ++pnt ) {
+        // fractional part of 'y' arc coordinate for better color reproduction
+        const double frac = *pnt - std::floor( *pnt ) - 0.5;
+
+        // rainbow colors generation from: https://stackoverflow.com/questions/40629345/fill-array-dynamicly-with-gradient-color-c
+        for ( int i = 0; i < rainbowThickness; ++i ) {
+            // normalize 'x' coordinate to 4.6 (for 'fheroes2' rainbow colors) regions, each region is 256 units long
+            const int32_t normalized = static_cast<int32_t>( ( i - frac ) * 256 * 4.65 / rainbowThickness );
```
ihhub, line 568: `256 * 4.65 / rainbowThickness` could be cached outside the double loop.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+    const int32_t y0 = ( height < 0 ) ? height : 0;
+
+    // coefficients for multipliers of 'pow1' and 'pow2' degrees
+    const double k1 = ( height * ( 1 - pow2Ratio ) ) / std::pow( width, pow1 );
+    const double k2 = ( height * pow2Ratio ) / std::pow( width, pow2 );
+
+    // calculate 'y' coordinates for arc: y = k1*(x-x0)^pow1+k2*(x-x0)^pow2 and push it to 'arc' vector
+    for ( int x = 0; x < std::abs( width ); ++x ) {
+        double y = k1 * std::pow( ( x + x0 ), pow1 ) + k2 * std::pow( ( x + x0 ), pow2 ) - y0;
+        arc.push_back( y );
```
ihhub, line 541: We can combine these 2 lines and directly pass the calculated value to `push_back()` call.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,94 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+            case 3:
+                colorRed = 0;
+                colorGreen = 255 - colorMod;
+                colorBlue = 255;
+                break;
+            case 4:
+                colorRed = colorMod;
+                colorGreen = 0;
+                colorBlue = 255 - colorMod;
+                break;
```
ihhub, line 599: We are missing here the default case.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3825,32 +3913,102 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
+        const fheroes2::Rect & battleArea = border.GetArea();
         const fheroes2::Sprite & unitSprite = fheroes2::AGG::GetICN( unit.GetMonsterSprite(), unit.GetFrame() );
+        const int32_t unitCenter = ( unitSprite.width() / CELLW + 1 ) * CELLW / 2;
+        const fheroes2::Point rainbowDescendPoint( pos.x + unitCenter, pos.y - unitSprite.height() / 2 );
+
+        // if the creature is low on the battleboard - the rainbow will be from the top (in the original game the threshold is about 140 pixels)
+        const bool isVerticalRainbow = ( rainbowDescendPoint.y > 140 );
+
+        // for the right army the rainbow will be from the right
+        const bool isRainbowFromRight = ( unit.GetCurrentColor() == arena.GetArmy2Color() );
```
ihhub, line 3925: Please use `isReflect()` method instead of this approach.

ihhub (changes requested):
Hi @Districh-ru , I left multiple comments here. Can you please check them?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Thank you @ihhub, it should be 4.6 in the code.
I was seeking the theshold where purple turns to red and forgot to correct here. :(

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Part of the game code uses 'int' and the other 'int32_t' so I'm confused what to use. Initially, I wrote 'int' everywhere, later I changed it to 'int32_t' and, of course, forgot to change it in this place.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Made this by integrating the flip operation into the `DrawRainbow` function, so the `Flip` function is no longer needed.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: @ihhub, `isReflect()` checks if the unit sprite is reflected and, for example, if the right army creature (with `isReflect()=1`) attacks from the left if turns and then we check the `isReflect()` and it returns `0`.
I wanted the rainbow to ascend from the side of the creature's hero.

Districh-ru: @ihhub, thanks a lot for reviewing my code and for optimization tips! I will take them into account.

Branikolog:
Hi, @Districh-ru 

> what do you think: rainbow should have constant animation step or constant animation time?

I cannot tell precisely right now. It would be easier to tell, if I'd test both versions. :)
I think, I would prefer constant animation time more, since on high speeds I can't even notice that short rainbow.
Btw, the work you've done looks really nice! 👍 
I noticed only a small difference in colours. 
Don't blame me much, but I like your previous implementation of rainbow for higher sells you've shown at https://github.com/ihhub/fheroes2/pull/6179#issuecomment-1325462651 . I personally would love to have in game something in between current (original) and the one I mentioned above. Something a little bit more parabolic-like. Just my thoughts, anyway. :)

Districh-ru:
> Hi, @Districh-ru
> 
> > what do you think: rainbow should have constant animation step or constant animation time?
> 
> I cannot tell precisely right now. It would be easier to tell, if I'd test both versions. :) I think, I would prefer constant animation time more, since on high speeds I can't even notice that short rainbow. Btw, the work you've done looks really nice! 👍 I noticed only a small difference in colours. Don't blame me much, but I like your previous implementation of rainbow for higher sells you've shown at [#6179 (comment)](https://github.com/ihhub/fheroes2/pull/6179#issuecomment-1325462651) . I personally would love to have in game something in between current (original) and the one I mentioned above. Something a little bit more parabolic-like. Just my thoughts, anyway. :)

Hi, @Branikolog 
On the video you mentioned, the animation time is constant.
I can set minimum draw parts to avoid super fast rainbow rendering for small rainbows, and set maximum draw parts for long ones. What do you think?
All rainbow parameters are calculated by linear equations and depend on the rainbow length. We can set a minimum rainbow arc height for small rainbows. (In my first video, the arc height was constant.) 
For example we can change it to start from about 10px:

https://user-images.githubusercontent.com/113276641/204142849-57e75059-951e-4576-9aa9-180120486380.mp4


But it won't be the same as original. Both options (orginal or not) will suit me as well.
...maybe in the future these parameters will be set by some game presets and maybe by the players themselves...

Districh-ru:
@Branikolog, what do you think of this:

https://user-images.githubusercontent.com/113276641/204143877-bd8857a9-735c-4819-9ffb-47d339142fc3.mp4



In this video, the maximum number of animation steps is set to 40 and the minimum to 20. So we don't have very slow and very fast animations.
If it is ok, I'll add it to the next commit.

Branikolog:
Hi, @Districh-ru 
For me personally, rainbows from the last video have acceptable generation speed and nice-looking ark height. But lets listen to other members, probably, they would be satisfied with more original-like behaviour.

zenseii:
Hi, guys. I'm fine with whichever.
My only concern is that it should speed up according to the battle speed that is set, but if that doesn't happen now in this PR then I wouldn't worry and just open that can in the future.

LeHerosInconnu:
Hello everyone,

I think that the duration of the rainbow animation should be the same whatever the length of the rainbow.

The rainbow animation should be finished before the end of the rainbow sound.

In the original game, the rainbow does not necessarily originate on the same side of the screen as the troop (coming from the right side of the screen for troops positioned on the right side of the screen at the beginning of the combat).

In the original game, the rainbow does not necessarily originate on the right or left side of the screen, it can also originate at the top of the screen.

In original game:

https://user-images.githubusercontent.com/43583994/204157298-8f2056c3-a40c-4b42-b034-25ca98add5ab.mp4

vlcsnap-2022-11-27-21h02m27s027

vlcsnap-2022-11-27-21h03m57s136

vlcsnap-2022-11-27-21h04m15s302

vlcsnap-2022-11-27-21h04m32s732

[0 9926.zip](https://github.com/ihhub/fheroes2/files/10099124/0.9926.zip)

src/fheroes2/battle/battle_interface.cpp
ihhub: Yes, it's a mix of here and there. In such cases use the majority rule 🙂

src/fheroes2/battle/battle_interface.cpp
ihhub: @Districh-ru , in this case make sure that such logic covers cases like Berserk spell. I'll check later this again.

Districh-ru: @ihhub , for Berserk spell (color == -1) and in other similar cases the rainbow will always be from the left. In this case (color < 0) I cuold use unit.GetColor. I'll do this in next commit.

Districh-ru:
Hello @zenseii , @LeHerosInconnu ,
The animation speed depends on the 'battle speed' setting.
I can make a constant count of drawing steps so on zero speed the animation time will be the same as rainbow sound time.
As I noticed, the original game draws a horisontal rainbow for the cretures in the upper 3 hexes and starting from the side of creatures army.
For the lower creatures the rainbow always comes from the top of the screen ("vertical rainbow"). And in original game its direction is not clear.
For such vertical rainbows I can set its direction according to the position of the creature. So it will always have full size and its trajectory will not be squeezed like in my last videos. (I'll try to implement this in the evening.)
What do you think?

Districh-ru:
Implementation of the vertical rainbow orientation approach mentioned above with corrected draw speed:

https://user-images.githubusercontent.com/113276641/204320997-d367d153-88c9-422f-b071-fc26f62a107c.mp4

Districh-ru:
What do you think about making the rainbow colors completely look like the original ones (from the center of the luck sprite)?

https://user-images.githubusercontent.com/113276641/204354329-a241cc36-ffe4-4dee-ab18-759ef8aa5d37.mp4

It is simpler and requires less computing resources. (Why did I choose the hardest way in the first place??? - Because I wanted to make a smoother transition of rainbow colors in the curve.)

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -523,6 +523,67 @@ fheroes2::Image DrawHexagonShadow( const uint8_t alphaValue )
...
+    if ( isVertical ) {
+        rainbowLine = fheroes2::Image( rainbowThickness, 1 );
+        // For a vertical rainbow orientation the line needs to be transposed. 
+        fheroes2::Transpose( fheroes2::Crop( luckSprite, luckSprite.width() / 2, 0, 1, rainbowThickness ), rainbowLine );
+        if ( !flipHorizontally ) {
+            rainbowLine = fheroes2::Flip( rainbowLine, true, false );
+        }
+    }
+    else {
+        rainbowLine = fheroes2::Crop( luckSprite, luckSprite.width() / 2, 0, 1, rainbowThickness );
```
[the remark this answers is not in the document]
Districh-ru: Changed luckSprite type to fheroes2::Image

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3825,32 +3886,117 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
+            pow2 = 0;
+            pow4 = 5;
+            pow2ratio = 0.0;
+            pow4ratio = 0.5;
+            rainbowLength = borderDistance;
+            // Formula for rainbowAscend corresponding to the original game was: 82.05 - 0.014 * rainbowLength.
+            rainbowAscend = 75;
+            // Formula for rainbowDescend corresponding to the original game was: 0.158 * rainbowLength - 18.25.
+            rainbowDescend = std::max( 1, static_cast<int32_t>( 0.1233 * rainbowLength + 0.7555 ) );
+            // Formula for rainbowTop corresponding to the original game was: 0.616 * rainbowLength + 31.5.
```
ihhub, line 3954: I don't think that we need to highlight what was in the original game as it rises questions where this data came from.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3825,32 +3886,117 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
-        int width = 2;
-        fheroes2::Rect src( 0, 0, width, luckSprite.height() );
-        src.x = ( luckSprite.width() - src.width ) / 2;
-        int y = pos.y + pos.height - unitSprite.height() - src.height;
-        if ( y < 0 )
-            y = 0;
+        GetHalfArc( rainbowArc, -rainbowArcBegin.width, rainbowArcBegin.height, pow1, pow2, pow2ratio );
+        GetHalfArc( rainbowArc, rainbowArcEnd.width, rainbowArcEnd.height, pow3, pow4, pow4ratio );
+
+        const fheroes2::Sprite luckSprite = DrawRainbow( rainbowArc, rainbowThickness, isVerticalRainbow, isRainbowFromRight );
```
ihhub, line 3967: This variable should be `fheroes2::Image` type.

ihhub (changes requested):
Hi @Districh-ru , I put several more comments here. Could you please take a look at them?

src/fheroes2/battle/battle_interface.cpp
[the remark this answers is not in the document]
Districh-ru: I have changed `fheroes2::Crop` to `fheroes2::Copy` as 'Crop' returns type `Sprite`.

Districh-ru:
Hi @ihhub, thanks for the comments. I have updated the code. Could you please check it out when you have time.

src/fheroes2/battle/battle_interface.cpp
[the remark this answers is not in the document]
ihhub: It is absolutely valid. Crop is used in cases when we are unknown about the size we want to copy.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3825,32 +3889,114 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
+        // The distance from the right or left battlefield border to the 'lucky' creature in the direction from the beginning of the animation to its end.
+        const int32_t borderDistance = isRainbowFromRight ? battleArea.width - rainbowDescendPoint.x : rainbowDescendPoint.x;
+
+        // The rainbow thickness in pixels: it must be 15 pixels to match the rainbow thickness from original sprite 'ICN::EXPMRL'.
+        const int32_t rainbowThickness = 15;
+
+        // Declare rainbow generation parameters:
+        // Rainbow arc parameters for: y = (1-pow2ratio)*k1*(x-x0)^pow1+pow2ratio*k2*(x-x0)^pow2.
+        // Parametars pow3, pow4, pow4ratio are the same as pow1, pow2, pow2ratio, but for the second part of the arc.
+        const int32_t pow1 = 2;
```
ihhub, line 3913: My only concern is about this place: we declare a lot of variables without default values. By human nature we might forget to set a value to one of them. Can we please add initialisation values to them like 0?

[Optional] you can use C++11 initialization style to avoid any implicit conversions. For example:
const int32_t pow1{ 2 };
This approach makes sure that use the correct types.

ihhub (changes requested):

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Made appropriate changes to the code.
@ihhub, thank you for tips and explanations!

ihhub (approved):

ihhub:

LeHerosInconnu:
Hello everyone,

I made some tests on this subject, for now it seems fine to me. :)

PR #6261, opened 2022-11-27
Set the minimum supported version to Android 6.0

oleg-derevenetz:
Related to #6008

* Apache commons-io 2.11.0 did not work with Android 6 due to the following error:

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.fheroes2, PID: 8710
    java.lang.NoClassDefFoundError: org.apache.commons.io.IOUtils$$ExternalSyntheticLambda2
        at org.apache.commons.io.IOUtils.<clinit>(IOUtils.java:183)
        at org.fheroes2.MainActivity.extractAssets(MainActivity.java:86)
        at org.fheroes2.MainActivity.onCreate(MainActivity.java:40)
        at android.app.Activity.performCreate(Activity.java:6309)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1113)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2530)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2665)
        at android.app.ActivityThread.-wrap11(ActivityThread.java)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1499)
        at android.os.Handler.dispatchMessage(Handler.java:111)
        at android.os.Looper.loop(Looper.java:207)
        at android.app.ActivityThread.main(ActivityThread.java:5765)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:789)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:679)
```

The [proper desugaring](https://developer.android.com/studio/write/java8-support) was implemented to fix this issue.

* Set `minSdkVersion` to the lowest API level on which the application has been tested and 100% works - 23 (Android 6.0). Maybe it will work even on older versions, but I don't have such devices.

Native libraries are still built with minimum API level set to 19 (Android 4.4), it's the minimum API level supported by the latest NDK at the moment.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -23,6 +23,13 @@ android {
         }
     }
 
+    compileOptions {
+        coreLibraryDesugaringEnabled true
+
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+
     externalNativeBuild {
         ndkBuild {
             path 'jni/Android.mk'
@@ -73,9 +80,10 @@ android {
 }
 
 dependencies {
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.2.0'
+
     implementation project(':sdl2')
-    // Versions 2.9.0+ do not work on devices with API level 23 (Android 6.0)
-    implementation group: 'commons-io', name: 'commons-io', version: '2.8.0'
+    implementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
 }
 
 task copyH2D(type: Copy) {
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -8,7 +8,7 @@ android {
     defaultConfig {
         applicationId 'org.fheroes2'
 
-        minSdkVersion 19
+        minSdkVersion 23
         targetSdkVersion 31
 
         // These paths are set relative to the android project directory
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -74,6 +74,7 @@ android {
 
 dependencies {
     implementation project(':sdl2')
+    // Versions 2.9.0+ do not work on devices with API level 23 (Android 6.0)
     implementation group: 'commons-io', name: 'commons-io', version: '2.8.0'
 }
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -74,7 +74,7 @@ android {
 
 dependencies {
     implementation project(':sdl2')
-    implementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
+    implementation group: 'commons-io', name: 'commons-io', version: '2.8.0'
 }
 
 task copyH2D(type: Copy) {


oleg-derevenetz:
Hi @Branikolog could you please verify that you are able to run APK from this PR on your Android 7 device? I was able to run this APK on my Android 6 tablet just fine.

ihhub (approved):

ihhub:
Hi @oleg-derevenetz , I suggest to explicitly state minimum recommended version of Android in `docs/README_android.md` file to avoid any future questions form users. What do you think?

oleg-derevenetz:
@ihhub it's not quite the minimum recommended version, if device's API level is less than `minSdkVersion`, then installation will be canceled and `INSTALL_FAILED_OLDER_SDK` will be returned (in Android GUI this should be accompanied with the corresponding error message for the user), so you just will not be able to install the app at all. We can add the mention of the minimum Android version in the readme, but this seems a bit redundant.

Branikolog:
Hi, @oleg-derevenetz and @ihhub !
I've tested fheroes2 on my Android 7.0 device and it works brilliant!
~60 fps on adventure map.

ihhub:

oleg-derevenetz:
@Branikolog

> ~60 fps on adventure map.

It is relatively slow. If you install the release snapshot build from [here](https://github.com/ihhub/fheroes2/releases/tag/fheroes2-android), it should be much faster. Please note that release builds are signed by the different certificate, so you won't be able just to update the debug build to release, most likely you will need to uninstall the debug build first.

Branikolog:
> @Branikolog
> 
> > ~60 fps on adventure map.
> 
> It is relatively slow. If you install the release snapshot build from [here](https://github.com/ihhub/fheroes2/releases/tag/fheroes2-android), it should be much faster. Please note that release builds are signed by the different certificate, so you won't be able just to update the debug build to release, most likely you will need to uninstall the debug build first.

Considering how old my phone is and its performance capabilities, fheroes2 runs relatively smooth. But, in a case I start moving cursor during hero galloping game becomes really slow, (The same issue I mentioned few days ago on Win.)

oleg-derevenetz:
@Branikolog

> But, in a case I start moving cursor during hero galloping game becomes really slow

Even on the release build with V-Sync turned off?

Branikolog:
@oleg-derevenetz 

> > But, in a case I start moving cursor during hero galloping game becomes really slow
> 
> Even on the release build with V-Sync turned off?

I mean android build  on my device from current PR had same lags as we have in master branch with V-sync on and software cursor rendering,

oleg-derevenetz:
@Branikolog

> I mean android build on my device from current PR had same lags as we have in master branch with V-sync on and software cursor rendering,

Yes, maybe, but for different reasons. Rendering on debug builds is very slow (by several times) by itself because compiler optimizations are off, and since cursor rendering on Android is done in software (cursors on touchscreens cannot be hardware) cursor movement "takes" the rendering time from the adventure map. So I suggest to try this with the release build on your device, it is already available since this PR is already merged.

Branikolog:
@oleg-derevenetz 
Good! I'll try it in a couple hours. 
Thanks for help and your inestimable contribution to the project. :)

zenseii:
> @Branikolog
> 
> > ~60 fps on adventure map.
> 
> It is relatively slow. If you install the release snapshot build from [here](https://github.com/ihhub/fheroes2/releases/tag/fheroes2-android), it should be much faster. Please note that release builds are signed by the different certificate, so you won't be able just to update the debug build to release, most likely you will need to uninstall the debug build first.

Wasn't the whole issue for @Branikolog that he couldn't use any Android build up until this one? Would he be able to try that build from Oct 25th then?

Branikolog:
@oleg-derevenetz 
FPS is still ~60 on a XL map. I'm not sure FPS works correctly here. When I try to move cursor during hero movement FPS shows ~100, but hero movement is slowed down for about 20%. Without touching anything hero moves quite smoothly.

oleg-derevenetz:
@Branikolog

> FPS is still ~60 on a XL map. I'm not sure FPS works correctly here.

It depends on the frequency of redrawing, and it, in turn, depends on the hero's speed of movement. For example, on my Win10 machine with hardware cursor I have ~20FPS when the hero just stands (because of map object animations), ~60FPS when it moves with the hero speed set to 1, and ~120+ FPS when it moves with the hero speed set to 9 (because the time interval between hero's frames is less than when the hero speed set to 1 and more frequent redrawing is needed).

> When I try to move cursor during hero movement FPS shows ~100, but hero movement is slowed down for about 20%.

Well, that FPS spike is because we need to draw additional frames (at least partially) to redraw the software cursor, and this causes the slowdown of hero's movement. It is not noticeable on powerful PC (if there is no V-Sync of course), but this still can be noticeable on weaker handheld devices. Perhaps, on more modern handheld devices, this would also be unnoticeable. In any case, no one is expected to drag their finger across the screen during the hero's movement :)

PR #6264, opened 2022-11-27
Remove some leftovers of support for visiting the allied castles

oleg-derevenetz:
We don't support this feature anymore, but some leftovers are still there.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you for cleaning up the code!

PR #6266, opened 2022-11-29
Use explicit shell: bash in GitHub Actions scripts

oleg-derevenetz:
The reason for this change is as follows: while `bash` is in fact used on *nix OS images if shell is not specified, it is [used with relaxed parameters ](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell)that doesn't include `-o pipefail`. I was forced to add `set -o pipefail` in place, which is actually unreliable and may be overlooked or forgotten.


ihhub:

PR #6269, opened 2022-11-30
Fix position of left barbarian's captain quarters part

### LINKED ISSUE #6263: Castle screen, barbarian captain's quarters has shifted column
### Preliminary checks

### Platform

Windows

### Describe the bug

image
Recently added column ran away.

### Save file

None.

### Additional info

_No response_

undef21:
fix #6263


ihhub:
I think we have to fix assets instead of ignoring an invalid value from them. What do you think?

undef21:
This bug is applying offsets twice. The first applying was in this file and the second one is [here](https://github.com/ihhub/fheroes2/blob/c33f389aaefcb0f1090caa08c50887ccf8321e55/src/fheroes2/gui/ui_castle.cpp#L289). The resources are correct.

Also, this regression was introduced in #5407.

ihhub (approved):

ihhub:
@Branikolog , please test this pull request for every castle.

Branikolog (approved):

ihhub:
@undef21 , thank you so much for fixing this issue!

PR #6267, opened 2022-11-30
Cleanup & simplify MSVC project and props files

oleg-derevenetz:
1. Removed obsolete and non-working project files for `82m2wav`, `til2img` and `xmi2mid` tools. They haven't been working for a long time and it doesn't seem like anyone cares;
2. Performed major deduplication & cleanup of MSVC project files, common definitions are moved to `*.props`;
3. Fixed multiple link-time warnings like `LIBCMT.lib(initializers.obj) : warning LNK4098: defaultlib 'msvcrt.lib' conflicts with use of other libs; use /NODEFAULTLIB:library` and `LINK : /LTCG specified but no code generation required; remove /LTCG from the link command line to improve linker performance` and enabled the handling of linker warnings as errors.

PR #6275, opened 2022-12-01
MSVC: do not define _DEBUG by hand

oleg-derevenetz:
Follow-up to #6267

There is no need to define `_DEBUG` by hand because it's [automatically defined](https://learn.microsoft.com/en-us/cpp/c-runtime-library/debug?view=msvc-170) when we set the `RuntimeLibrary` to any of the debug versions, which is what we are doing one line above.


ihhub:
Hi @oleg-derevenetz , can you please take a look at #6276 ? It might be a false alarm but it was reported just few hours ago.

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks for this code cleanup!

PR #6277, opened 2022-12-02
Update the dependency package hash for Android

oleg-derevenetz:
Updates SDL to 2.26.1


ihhub:

Issue #6276, opened 2022-12-01
Remove VCRUNTIME140.dll dependency


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Based on the report from user this dependency was introduced recently (needs verification). We need to make sure that these files either not needed or shipped with the archive.

### Save file

None.

oleg-derevenetz:
Hi @ihhub how is this a regression? This DLL was always needed, it's a part of the Visual C++ Redistributable Package.

ihhub:
Then it's a false alarm :)

oleg-derevenetz:
@ihhub some apps offer the installation of MSVC redistributable package in their installers.

ihhub:
@oleg-derevenetz , if it is easy to do then why not? If it takes days to implement let's scrap the idea for now.

PR #6278, opened 2022-12-02
Install the latest Microsoft Visual C++ Redistributable package using Inno Setup

oleg-derevenetz:
Related to #6276

I also switched this workflow from Cygwin to the pre-installed MSYS2, this somewhat accelerated the build - from 6-8 minutes to 4-6 minutes.

I have a few notes though:

1. ~We download the redistributable package installer [from the MS website](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170) via a permalink without any verification. Hash verification is useless here, because we don't control "the other side" and I suspect that there will be silent regular updates for that package, but maybe we can use at least the digital signature verification. We can try to figure this out.~ Implemented the mandatory signature verification using the `signtool.exe`.
2. ~When `vcredist.exe` updates libraries in `SYSTEM32`, Inno Setup detects this and suggests to reboot the system. I can turn it off, but I'm not sure that I need to do this. @ihhub what do you think?~ Turned it off. 
3. Currently Inno Setup always runs the `vcredist.exe` in the quiet mode (no GUI, no questions), and this is on purpose - if we will ask the average user "Do you want to install blah-blah-blah package?", he will say "what package?...", so this question is mostly useless. But we can offer `vcredist` installation as an option, if someone brings arguments that it's better this way.

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -44,18 +44,21 @@ jobs:
         shell: bash
     steps:
     - uses: actions/checkout@v3
-    - name: Install dependencies and cygwin
+    - name: Install dependencies
       run: |
         script/windows/install_packages.bat
-        choco.exe install cygwin cyg-get -y
-        cyg-get.bat make libiconv libunistring5 gettext-devel
+    - name: Install MSYS2 tools
+      run: |
+        C:\msys64\usr\bin\pacman.exe -S --noconfirm --noprogressbar gettext-devel
+      shell: cmd
     - name: Download MSVC Redistributable package
       run: |
-        $ErrorActionPreference = "Stop"
-        [void](New-Item -Path ".vcredist" -ItemType "directory")
-        # This should be a permanent link
-        Invoke-WebRequest -Uri "https://aka.ms/vs/17/release/vc_redist.${{ matrix.config.platform }}.exe" -OutFile ".vcredist\vcredist.exe"
-      shell: pwsh
+        mkdir .vcredist && ^
+        C:\msys64\usr\bin\wget.exe -O .vcredist/vcredist.exe "%VCREDIST_URL%"
+      shell: cmd
+      env:
+        # Permalink URL from https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170
+        VCREDIST_URL: https://aka.ms/vs/17/release/vc_redist.${{ matrix.config.platform }}.exe
     - uses: microsoft/setup-msbuild@v1
     - name: Generate version information
       run: |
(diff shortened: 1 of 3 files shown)


ihhub:
> When vcredist.exe updates libraries in SYSTEM32, Inno Setup detects this and suggests to reboot the system. I can turn it off, but I'm not sure that I need to do this. @ihhub what do you think?
Logically, if it's just a DLL Windows doesn't need to restart as we are installing files which are unused by any other application (in theory).

My biggest concern is the size of the installer which is now weights a wooping 30 MB. Plus we don't provide the same files with the ZIP archive. I am thinking what if we ship DLLs together with the game, wouldn't it be better for users?

I also noticed that with the latest changes we made for Visual Studio even debug build requires VC Redistributable package.

oleg-derevenetz:
@ihhub

> Logically, if it's just a DLL Windows doesn't need to restart as we are installing files which are unused by any other application (in theory).

These files can be used by any Windows app built with MSVC. And if there was earlier version of DLL with the same name, there is a high probability that it is already being used by some app. But that's not a reason for restart anyway, because installation of such DLL will be delayed, and, since there is already this DLL present, the older version is also good.

> I am thinking what if we ship DLLs together with the game, wouldn't it be better for users?

I'm not sure that we need to make a special infrastructure to track the changes in DLL usage when we can just install the standard MSVC package and forget about that, like everyone does :) That's what this package is for, after all.

> I also noticed that with the latest changes we made for Visual Studio even debug build requires VC Redistributable package.

Earlier builds required it as well, because SDL2 DLLs that are built with MSVC require it:

dlls

It's just that, as a rule, these libraries have already been installed earlier by some program that has already been installed, so most Windows PCs already have it.

oleg-derevenetz:
@zenseii

> I ran the installer. I noticed it installed the redistributable package even though I did that myself yesterday and my dlls were already up to date.

When you see the message "Installing MSVC blah blah package" it just means that Inno Setup runs the package installer EXE. Inno Setup has no idea whether you have all libs in place or not, it just runs the installer and that's all. Installer performs necessary checks and may even not install anything at all.

ihhub:
I've added the biggest complainer about Windows builds - @Branikolog . Lets see what he says🤣

oleg-derevenetz:
I can turn off the reboot request when Inno Setup detects the "delayed" DLL updates, there is no problem with that, and we can at least mention the URL of the `vcredist` in the ZIP archive's README.

ihhub:
> I can turn off the reboot request when Inno Setup detects the "delayed" DLL updates, there is no problem with that, and we can at least mention the URL of the `vcredist` in the ZIP archive's README, if you don't want to just add it to the ZIP.

Let's do this. We always meed to think from user's perspective, especially if they have no idea how to handle apps :)

ihhub (approved):

Branikolog:
Hi, @ihhub !
Installation file is so huge now! I could hardly find enough space on my PC to install fheroes2! :)
Everything installs, works etc. The only thing I missed in installation program is that it haven't offered me to specify a path on my storage.

oleg-derevenetz:
Hi @Branikolog

> The only thing I missed in installation program is that it haven't offered me to specify a path on my storage.

What do you mean? This:

path

? This is asked only when you perform a fresh install. It is not asked when there is an existing installation to update.

Branikolog:
Hi, @oleg-derevenetz !
Oh, this time I wasn't asked for it. Probably, I had already installed one copy of the engine, which I completely forgot to exist.
Clean install works as intended now. 👍

Branikolog (approved):

ihhub:

PR #6281, opened 2022-12-02
Do not generate Dragons for specific scenario of SW campaign

### LINKED ISSUE #6136: Campaign "Carry over forces" scenario shouldn't be abused with alliance creatures
### Preliminary checks

During campaign 8(9) missions walkthrough (the ones to gather forces to carry over to the last scenario) there could happen such case, that map occasionally generates Dragons, Ogres or Dwarves. In a case player has an alliance award they 100% join his army which significantly increases his chances to easily finish the final scenario. This rare cases are abused by players, waiting till the last week of the scenario to gather unfair number of free creatures.

I think we should exclude generation of these creatures. Or at least those particular, which 100% join player due to alliance.

### Additional info

_No response_

ihhub:
Only for Hard difficulty.

close #6136

### DIFF
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -397,7 +397,7 @@ namespace Maps
 
         void SetTerrain( uint32_t sprite_index, uint32_t shape /* 0: none, 1 : vert, 2: horz, 3: both */ );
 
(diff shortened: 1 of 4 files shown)


ihhub:
I am closing this pull request for now.

Issue #5373, opened 2022-05-05
Independent Source Size and Rendering Resolution with Multiple Scaling Options (à la HOMM3 HD)


Professor-CDawg:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

I prefer to play at resolutions lower than my monitor's native 1920x1080 because I find the game objects too small at 1080p.

This is currently possible in fheroes2's settings menu, but only at resolutions reported as supported from the OS. Furthermore, the image is scaled directly by the GPU's built-in scaling, which is quite blurry in my case (and I suspect the vast majority of cases).

I would like to be able to set a custom resolution not inherently supported by my OS (i.e., 853x480) so I can have a wider variety of widescreen options at resolutions lower than 1080p. Additionally, I would like to be able to select from multiple scaling options other than linear filtering, such as Cubic and Lanczos, as these produce an image that is much more pleasing to my eye.

The features I have described and requested are present in the HOMM3 HD mod for Heroes 3 (hence why I missed them when I tried fheroes2). In HOMM3 HD, I play at a "source size" of 1180x664 (a 16:9 resolution not supported by my OS), and use the Cubic "stretch filter" option.

### Describe the possible solution

I may be phrasing this wrongly, but I believe what I'm asking for could be summarized as an implementation of "software upscaling" (as opposed to how fheroes2 currently only allows for "hardware upscaling" via the system's GPU).

Put another way, add a "source size" option that is independent of the final rendering resolution; the final rendering resolution is arrived at via a selectable "stretch filter" (e.g. Cubic, Lanzcos).

### Additional info

Examples of HOMM3 HD's implementation of the feature:

source size: 1180x664; stretch filter: Linear
![image](https://user-images.githubusercontent.com/68561018/166863068-aa8d5723-d881-4cf1-a1e4-fe0ebea139db.png)

source size: 1180x664; stretch filter: Cubic
![image](https://user-images.githubusercontent.com/68561018/166862721-7b77730f-42b8-478a-8310-58cee9292034.png)

source size: 1066x600; stretch filter: Cubic
![image](https://user-images.githubusercontent.com/68561018/166862973-7e039a36-57ac-4095-b912-0c721852204a.png)

oleg-derevenetz:
Duplicate of #4946

tarboss:
[...]

Issue #4946, opened 2022-01-22
Screen Filter (e.g. xBRZ)


Laserlicht:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

When playing on a high-resolution monitor in original resolution, everything is very small.

If you play it in a lower resolution, then it can get very blurry.

### Describe the possible solution

In the HoMM3 HD mod it is possible to choose a lower resolution and increase it with a filter. xBRZ x4 +Hermite (at 1870x1052 on a 1440p monitor) for example gives a very good, sharp result and is a good compromise. Of course, it costs computing power, that's clear.

Unfortunately I don't have enough experience to implement something like this myself. But perhaps others will also find the idea good.

ihhub:
Hi @Laserlicht , what do you mean by blurry graphics? The game has nothing related to filtering. Did you try 800x600 or 1024x768 resolutions? Do you play in fullscreen? If not why don't you try to resize the window?

Also please specify your OS and what exact resolution and settings you have for your monitor.

Laserlicht:
Hi @ihhub 

Sorry, I was not precise enough.

Yes, full screen. But with a lower resolution than that the monitor. So that the gui elements are larger (and therefore also blurrier).

Laserlicht:
Example: If I use 1600x900 on my 2560x1440 monitor it would look like this (view it in 100%):
![unsharp](https://user-images.githubusercontent.com/13953785/150644678-50190d78-6de6-4709-9432-6ac054f23eda.png)

If I apply an xBrz filter on it ([e.g. this](https://www.maxlaumeister.com/pixel-art-upscaler/)) then it would look like this (view it in 100%):
![xbrz](https://user-images.githubusercontent.com/13953785/150644752-753d9c64-ba1f-43af-a6fa-68d3cb8f6d5b.png)

In my opinion much better.

afrikansviridovic:
@Laserlicht Have you tried 1280x720 with integer scaling option (AFAIK it is available for AMD and nVidia graphic cards)?

Laserlicht:
@afrikansviridovic:
The difference is very small.

We could e.g. use the [xbrz library from Zenju](https://sourceforge.net/projects/xbrz/).

For me, however, this is too much. Using the filter is not that hard, but you have to understand the rendering of fheroes2 first.

[...]

PR #2238, opened 2020-11-17
Fix rendering and mouse behavior for SDL2

undef21:
This pr was tested for SDL2 Linux build (with fake fullscreen).
relates to #1091
relates to #900

PR #6282, opened 2022-12-02
Make an option to toggle screen scaling type

ihhub:
With introduction of "fake" full screen mode for Windows it is visible that the picture is blurred with linear mode. To cover needs of everyone we are adding a new configuration option to choose pixelated image or blurred. By default it is linear.

relates to #5373
relates to #2238
relates to #4946


oleg-derevenetz (approved):

ihhub:
By the way, linear scaling is slower than nearest neighbour approach due to its complexity so users want to improve performance they need to switch to nearest mode. PS Vita should use this by default.

These are general thoughts.

Branikolog (approved):

PR #6289, opened 2022-12-03
Add the ability to set the maximum text width for each icon in settings dialogs

Districh-ru:
The settings dialogs have different spacing between elements: in audio and graphics settings there is more space than in other configuration dialogs. So for these two dialogs the maximum text width can be set higher to allow use of longer phrases (especially for translations).
This PR adds the ability to set per element (per icon) the maximum text width for configuration dialogs. In this PR maximum width is set to `( distance_between_elements - 5px )`.

### DIFF
--- a/src/fheroes2/gui/ui_option_item.h
+++ b/src/fheroes2/gui/ui_option_item.h
@@ -20,6 +20,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 
 #include "math_base.h"
--- a/src/fheroes2/gui/ui_option_item.h
+++ b/src/fheroes2/gui/ui_option_item.h
@@ -25,6 +25,12 @@
 
 #include "math_base.h"
 
+enum : int32_t
+{
+    TWO_ELEMENT_ROW_TEXT_WIDTH = 113,
+    THREE_ELEMENT_ROW_TEXT_WIDTH = 87
+};
+
 namespace fheroes2
 {
     class Sprite;
--- a/src/fheroes2/gui/ui_option_item.h
+++ b/src/fheroes2/gui/ui_option_item.h
@@ -25,14 +25,14 @@
 
 #include "math_base.h"
 
-enum : int32_t
-{
-    TWO_ELEMENT_ROW_TEXT_WIDTH = 113,
-    THREE_ELEMENT_ROW_TEXT_WIDTH = 87
-};
-
 namespace fheroes2
 {
+    enum UiOptionTextWidth : int32_t
+    {
+        TWO_ELEMENTS_ROW = 113,
+        THREE_ELEMENTS_ROW = 87
+    };
+
     class Sprite;
 
     void drawOption( const Rect & optionRoi, const Sprite & icon, std::string titleText, std::string valueText, const int32_t textMaxWidth );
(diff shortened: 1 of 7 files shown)


ihhub:
Hi @Districh-ru , we have 2 types of options setups: a row with 2 icons and a row with 3 icons. Should we stick to only 2 options while setting text width? My point comes from a perspective of standartization.

What do you think?

Districh-ru:
Hi @ihhub, you are right - for now we have only the current 2 types of option settings. I made this feature to allow flexible width adjustment for future possible changes and new dialogs (perhaps a dialog with 2 items on one line and 3 items on another). The width is calculated according to the spacing between icons for the current dialog code (near to the spacing setting).

Do you suggest sending the number of horizontal elements to the `drawOption` function and using `switch` to set the value for `textMaxWidth` in this function?
In this case, we also have to pass a parameter to the `drawOption` function, because it does not know what dialog called it.

Or maybe we have to rewrite the dialog generation function and store in it standardized parameters not only for `textMaxWidth`, but also other parameters (window sizes, spacing between elements) for the main types of element arrangement?  :)
For now there are 2 different configuration dialog code styles:
- battle, audio, system settings;
- game, graphics settings.

ihhub:
Hi @Districh-ru , do we still need this change after fixing the real root cause?

Districh-ru:
Hi @ihhub , this improvement is still needed if we want to let the translators to use more space in graphics and audio settings for translated texts.
For now I cannot fully translate "V-Sync", because it needs more thab 87px width:
image

_Originally posted by @Branikolog in https://github.com/ihhub/fheroes2/pull/6165#discussion_r1025158327_

ihhub:
@Districh-ru , in order to avoid values being stored and calculated in multiple places is to create an enumeration in `src/fheroes2/gui/ui_option_item.h` file where we have 2 entries: one for 2 options row window and another for 3 options row and use them in files where we call `drawOption()` function. What do you think?

Districh-ru:
> @Districh-ru , in order to avoid values being stored and calculated in multiple places is to create an enumeration in `src/fheroes2/gui/ui_option_item.h` file where we have 2 entries: one for 2 options row window and another for 3 options row and use them in files where we call `drawOption()` function. What do you think?

Hi, @ihhub, yes, for those 2 entries, this is more optimal implementation. I made the suggested changes.

src/fheroes2/gui/ui_option_item.h
```diff
@@ -20,13 +20,20 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 
 #include "math_base.h"
 
+enum : int32_t
```
ihhub, line 28: It is a good practice to reduce a scope of elements. Right now this enumeration is in the global scope. I suggest put it into the namespace below. On top of this it would be great to have a name of the enumeration, something like UiOptionTextWidth.


If you have another view please speak out🙂

Districh-ru: Hi @ihhub, yes it is not good to do "garbage" in the global namespace. :)
Thanks for the tip, I've corrected the code.
The enum name you suggested looks good, so I've shortened the enumerators names.

ihhub (approved):

ihhub:

PR #6271, opened 2022-11-30
Add Danish language option and font

zenseii:
This adds Danish as an option and includes it in the CP1252 font.

Relates to https://github.com/ihhub/fheroes2/pull/6237

PR #6292, opened 2022-12-03
Change Danish abbreviation to dk from da

zenseii:
Da is an older abbreviation and dk is the standard now.

Follow up to #6271 
Related to #6237


ihhub:
@zenseii , thank you for the fix!

PR #6293, opened 2022-12-03
Fix multiline word splitting for short phrases with a long first word

### LINKED ISSUE #6288: Multi-line text splits the long first word, if it is greater than the second word, on the next line
### Preliminary checks

### Platform

Windows

### Describe the bug

If there is long text that needs to be split into two or more lines to fit in the given width, then the width of each line is calculated from simply putting all the characters in the given number of lines. In this case, if the first line has a long line that would fit in the given width, it will be cut to fit the new width. IMHO it is incorect.
In the screenshot below, you can see that the first word fits the given width, but if we add a small word "splitts", then it will be splitt.

long word

It was found while checking the translation. :)
translation 1 translation 2

### Save file

It's an interface bug.

### Additional info

The function `Text::draw` in `ui_text.cpp` contains a code that calculates the multi-line text width for every line.
I'll try to analyze this algorithm.

Districh-ru:
close #6288
relates to #6165 

before: translation 1   after: translation 2

### DIFF
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -419,6 +419,43 @@ namespace
             offset->x += lineWidth;
         }
     }
+
+    int32_t getMaxWordWidth( const uint8_t * data, const int32_t size, const fheroes2::FontType & fontType )
+    {
+        assert( data != nullptr && size > 0 );
+
+        int32_t maxWidth = 1;
+
+        // If there are more than 50 characters in 'data', then we have a long sentence with a large number of words
+        // and there is no need to analyze the word width - there should already be enough space for such a phrase.
+        if ( size < 51 ) {
+            const CharValidator validator( fontType.size );
+
+            int32_t width = 0;
+
+            const uint8_t * dataEnd = data + size;
+            while ( data != dataEnd ) {
+                if ( *data == lineSeparator || isSpaceChar( *data ) ) {
+                    // If it is the end of line ("\n") or a space (""), then the word has ended.
+                    if ( width > maxWidth )
+                        maxWidth = width;
+                    width = 0;
+                }
+                else if ( validator.isValid( *data ) ) {
+                    width += getCharWidth( *data, fontType );
+                }
+                else {
+                    width += getCharWidth( invalidChar, fontType );
+                }
+                ++data;
+            }
+
+            if ( width > maxWidth )
+                maxWidth = width;
+        }
+
+        return maxWidth;
+    }
 }
 
 namespace fheroes2
@@ -530,7 +567,7 @@ namespace fheroes2
         int32_t correctedWidth = maxWidth;
         if ( offsets.size() > 1 ) {
             // This is a multi-line message. Optimize it to fit the text evenly.
-            int32_t startWidth = 1;
+            int32_t startWidth = getMaxWordWidth( reinterpret_cast<const uint8_t *>( _text.data() ), static_cast<int32_t>( _text.size() ), _fontType );
             int32_t endWidth = maxWidth;
             while ( startWidth + 1 < endWidth ) {
                 const int32_t currentWidth = ( endWidth + startWidth ) / 2;
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -426,32 +426,30 @@ namespace
 
         int32_t maxWidth = 1;
 
-        // If there are more than 50 characters in 'data', then we have a long sentence with a large number of words
-        // and there is no need to analyze the word width - there should already be enough space for such a phrase.
-        if ( size < 51 ) {
-            const CharValidator validator( fontType.size );
+        const CharValidator validator( fontType.size );
 
-            int32_t width = 0;
+        int32_t width = 0;
 
-            const uint8_t * dataEnd = data + size;
-            while ( data != dataEnd ) {
-                if ( *data == lineSeparator || isSpaceChar( *data ) ) {
-                    // If it is the end of line ("\n") or a space (""), then the word has ended.
-                    if ( width > maxWidth )
-                        maxWidth = width;
-                    width = 0;
+        const uint8_t * dataEnd = data + size;
+        while ( data != dataEnd ) {
+            if ( *data == lineSeparator || isSpaceChar( *data ) ) {
+                // If it is the end of line ("\n") or a space (""), then the word has ended.
+                if ( maxWidth < width ) {
+                    maxWidth = width;
                 }
-                else if ( validator.isValid( *data ) ) {
-                    width += getCharWidth( *data, fontType );
-                }
-                else {
-                    width += getCharWidth( invalidChar, fontType );
-                }
-                ++data;
+                width = 0;
             }
+            else if ( validator.isValid( *data ) ) {
+                width += getCharWidth( *data, fontType );
+            }
+            else {
+                width += getCharWidth( invalidChar, fontType );
+            }
+            ++data;
+        }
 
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @Districh-ru , what's the original width of the area where we render text and what's the width of the first word?

Districh-ru:
Hi @ihhub, the original width of the area is set to 87px and the width of the first word is 60px.
In #6289, I tried to solve a similar problem by changing the original max width, but with no success.
The current logic wants to avoid drawing phrases with very long one line and very short other (the last line). This is good. The algorithm starts at (1+87)/2px and but if the first word is too long, thr algorithm cannot reach the end of this word: the word is automatically split. the phrase fits into the resulting number of lines, and the algorithm considers that everything is in order.
In this PR, I change the initial "corrected width" value from 1 to the maximum word width. So the corrected width will be no less than the width of the longest word. If the first word is longer than the given maximum 87px, the algorithm uses this value (87px) for the the "corrected width".

Districh-ru:
> Hi @Districh-ru , please take a look at getMultiRowInfo() function which calculates row information. See if this function behaves properly as it must handle words separated by a space.

@ihhub, this function works properly. :)
To draw text, the function is first called with the `maxWidth` parameter (for the current example it is 87) and as a result we get the number of lines needed to draw the text (`offsets.size()`). If it is greater than 1 - we have a multiline text and another algorithm is launched here: it recursively tries to set a new `currentWidth[ i ] = ( maxWidth+ currentWidth[ i - 1 ] ) / 2` (where currentWidth[0]=startWidth=1) and checks if the number of resulting lines is no more than needed.
While doing this `getMultiRowInfo()` function is called with the `currentWidth` parameter, which may be less than required for the word to fit. And the function should put the text into the given `currentWidth` and automatically breaks the long word into several lines (it must do this for these conditions).
I don't think that we have to change the `getMultiRowInfo()` function.

Maybe we have to change the width correction algorithm (this PR only changes its `startWidth` value) ... or maybe get rid of it, since it uses the computing resources? :)

For comparison, text with this algorithm we get:
изображение   изображение

And without:
изображение  изображение

The original game:
изображение

ihhub:
Hi @Districh-ru , the current logic for multi-line text (which is applicable for `draw()` method of **Text** and **MultiFontText** classes is indeed imperfect. It uses a binary search algorithm to detect the minimum number of characters which can evenly fit on all lines. However, this algorithm does not account words which are longer that this limit. How we should do:
- scan the whole text and find the longest word in pixels (don't forget about MultiiFontText class)
- set the found value into `startWidth`
- run binary search algorithm code

So your idea is correct but it should be flexible for every case. Right now it is based on some hardcoded values which is incorrect. Also add implementation for **MultiFontText**.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -419,6 +419,43 @@ namespace
...
+
+    int32_t getMaxWordWidth( const uint8_t * data, const int32_t size, const fheroes2::FontType & fontType )
+    {
+        assert( data != nullptr && size > 0 );
+
+        int32_t maxWidth = 1;
+
+        // If there are more than 50 characters in 'data', then we have a long sentence with a large number of words
+        // and there is no need to analyze the word width - there should already be enough space for such a phrase.
+        if ( size < 51 ) {
```
ihhub, line 431: Please remove this hardcoded value.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -419,6 +419,43 @@ namespace
             offset->x += lineWidth;
         }
     }
+
+    int32_t getMaxWordWidth( const uint8_t * data, const int32_t size, const fheroes2::FontType & fontType )
```
ihhub, line 423: `size` should be **int32_t**.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -419,6 +419,43 @@ namespace
...
+            const CharValidator validator( fontType.size );
+
+            int32_t width = 0;
+
+            const uint8_t * dataEnd = data + size;
+            while ( data != dataEnd ) {
+                if ( *data == lineSeparator || isSpaceChar( *data ) ) {
+                    // If it is the end of line ("\n") or a space (""), then the word has ended.
+                    if ( width > maxWidth )
+                        maxWidth = width;
```
ihhub, line 441: Please add parenthesis for this block. Also swapping values will keep the code more "consistent":
```cpp
if ( maxWidth < width ) {
    maxWidth = width;
}

width = 0;
```

src/fheroes2/gui/ui_text.cpp
```diff
@@ -419,6 +419,43 @@ namespace
...
+                    width += getCharWidth( *data, fontType );
+                }
+                else {
+                    width += getCharWidth( invalidChar, fontType );
+                }
+                ++data;
+            }
+
+            if ( width > maxWidth )
+                maxWidth = width;
```
ihhub, line 454: Just a small correction:
```cpp
if ( maxWidth < width ) {
    maxWidth = width;
}
```
It would be easier to read :)

src/fheroes2/gui/ui_text.cpp
```diff
@@ -530,7 +567,7 @@ namespace fheroes2
         int32_t correctedWidth = maxWidth;
         if ( offsets.size() > 1 ) {
             // This is a multi-line message. Optimize it to fit the text evenly.
-            int32_t startWidth = 1;
+            int32_t startWidth = getMaxWordWidth( reinterpret_cast<const uint8_t *>( _text.data() ), static_cast<int32_t>( _text.size() ), _fontType );
```
ihhub, line 570: Please add implementation for **MultiFontText** class as well. There you will need to get max of results.

ihhub (changes requested):

src/fheroes2/gui/ui_text.cpp
Districh-ru: Ok. 👍 
In the beginning there was no condition with hardcoded value. I did this to reduce the processor usage in thoughts that for the long phrases the first iteration or `currentWidth` will be grater than the word.

ihhub: This implementation will be faster than the current one due to reduce number of iteration for binary search.

src/fheroes2/gui/ui_text.cpp
Districh-ru: Here we are dealing with different types of fonts, so I changed this part to be the same as calling `getMulti RowInfo`.

src/fheroes2/gui/ui_text.cpp
Districh-ru: Oh yes, you're right!
@ihhub, thanks a lot for helping with this PR!

ihhub (approved):

src/fheroes2/gui/ui_text.cpp
```diff
@@ -419,6 +419,41 @@ namespace
...
+        int32_t maxWidth = 1;
+
+        const CharValidator validator( fontType.size );
+
+        int32_t width = 0;
+
+        const uint8_t * dataEnd = data + size;
+        while ( data != dataEnd ) {
+            if ( *data == lineSeparator || isSpaceChar( *data ) ) {
+                // If it is the end of line ("\n") or a space (""), then the word has ended.
```
ihhub, line 436: A tiny comment correction: `space ("")` should be `space (" ")` with a space :)

Districh-ru: You have Expert Eagle Eye skill 👍

ihhub:
@Districh-ru , thank a lot for this important fix!

PR #6297, opened 2022-12-04
Optimize defensive strategy for AI castles

ihhub:
- add extended logic for castle's defense actions
- make proper sorting castles for AI for further actions
- do not build a boat if no heroes exist
- remove `Troops::operator=()` method
- code cleanup to reduce the scope of functions


src/fheroes2/army/army.cpp
```diff
@@ -182,6 +182,7 @@ Troops::Troops( const Troops & troops )
 
 Troops & Troops::operator=( const Troops & rhs )
 {
+    clear();
```
oleg-derevenetz, line 185: Hi @ihhub this is not quite right - since raw pointers are stored here, and not smart pointers, we should call `delete` for them first, and only after that we can call `clear()`, otherwise we will get a memory leak.

oleg-derevenetz (changes requested):

src/fheroes2/army/army.cpp
ihhub: Oh, yes. I forgot about this. We should call `Clear()` which does the correct logic.

ihhub: Actually, we need to add extra logic to deallocate pointers :(

ihhub: OMG-OMG-OMG:
```cpp
Troops troops;
troops.Assign( otherTroops );
```
and it does nothing. I will skip fixing this class for now...

oleg-derevenetz: `Assign()` is good for `Army`, which pre-allocates the slots :) And `Troops` doesn't pre-allocate them because while it stores just `Troop` objects, `Army` stores `ArmyTroop` objects. The legacy logic :)

ihhub: I just deleted the method as used another method to avoid any weird logic.

oleg-derevenetz: @ihhub In fact, the current logic is also weird, because if we have an `Army` object upcasted to `Troops` and then we want to make a copy of it, then we will get the copies of `Troop` instead of `ArmyTroop` :) Maybe it even right, but maybe not - it depends.

ihhub: We need to rewrite it but for 1.0 let's keep as it is.

src/fheroes2/army/army.h
```diff
@@ -51,7 +51,7 @@ class Troops : protected std::vector<Troop *>
     Troops() = default;
     Troops( const Troops & troops );
     virtual ~Troops();
-    Troops & operator=( const Troops & rhs );
+    Troops & operator=( const Troops & rhs ) = delete;
```
oleg-derevenetz, line 54: We don't need to name a parameter here then.

src/fheroes2/army/army.h
```diff
@@ -51,7 +51,7 @@ class Troops : protected std::vector<Troop *>
     Troops() = default;
     Troops( const Troops & troops );
```
oleg-derevenetz, line 52: Maybe we don't need this copy constructor at all?

oleg-derevenetz: Eh, no, we still need it.

src/fheroes2/army/army.h
ihhub: Fixed.

src/fheroes2/ai/normal/ai_normal_castle.cpp
```diff
@@ -176,20 +179,47 @@ namespace AI
             }
         }
 
-        // Call internally checks if it's valid (space/resources) to buy one
-        if ( castle.GetKingdom().GetFunds() >= PaymentConditions::BuyBoat() * ( islandOrPeninsula ? 2 : 4 ) )
+        // Call internally checks if it's valid (space/resources) to buy one.
```
oleg-derevenetz, line 182: Could you please elaborate, which call checks what? The current wording gives the impression that `GetKingdom()` does this.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -164,6 +164,8 @@ namespace
         const MP2::MapObjectType objectType = tile.GetObject();
 
         if ( !MP2::isActionObject( objectType ) ) {
+            // TODO: add logic to verify if all parts of puzzle are opened and the location is known.
+            // TODO: once it is done, check if the tile does not have a hole. If it does not mark it as a valid object.
```
oleg-derevenetz, line 168: Do you want to give the AI an ability to dig for the Ultimate Artifact in future? I think it will be a little too much, because there will be no way for a human player to capture this artifact in battle. Most likely it will be lost.

oleg-derevenetz (commented):

src/fheroes2/ai/normal/ai_normal_hero.cpp
oleg-derevenetz: The original Heroes2 AI is dumb, so it's OK for it. fheroes2 AI is too smart IMHO to allow it to dig.

idshibanov: I think we should still implement it in the future. We can feature-gate it behind game difficulty so AI would only dig on higher ones.

ihhub: But people are screaming to give them challenge :)

I am okay to put it for higher difficulties only.

src/fheroes2/ai/normal/ai_normal_castle.cpp
ihhub: Sure. It's an old comment but I'll change it.

oleg-derevenetz:
Hi @ihhub could you please resolve the merge conflicts here?

src/fheroes2/ai/normal/ai_normal_castle.cpp
```diff
@@ -176,20 +179,47 @@ namespace AI
...
+            const Battle::Tower tower( castle, Battle::TWR_RIGHT, Rand::DeterministicRandomGenerator( 0 ), 0 );
+            const Troop towerMonster( Monster::ARCHER, tower.GetCount() );
+            const double towerStrength = towerMonster.GetStrength();
+            if ( possibleReinforcementStrength > towerStrength ) {
+                castle.recruitBestAvailable( kingdom.GetFunds() );
+                OptimizeTroopsOrder( castle.GetArmy() );
+            }
 
-            castle.recruitBestAvailable( castle.GetKingdom().GetFunds() );
+            if ( castle.GetActualArmy().getTotalCount() > 0 ) {
```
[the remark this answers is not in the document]
ihhub: Nope. AI doesn't always leave troops as there could be no hero and no monsters to be hired.

ihhub:
> Hi @ihhub could you please resolve the merge conflicts here?

It is done.

oleg-derevenetz (approved):

PR #6305, opened 2022-12-05
Replace Focus class by std::variant

vasilenkoalexey:
Replace custom `Focus` class in favor of `std::variant` which is more efficient extensible and safe


ihhub:
Hi @vasilenkoalexey , can you please explain how is the proposed solution better than old way of using classes? Should it better to just cleanup the existing class as we aren't going to expand it with new data.

vasilenkoalexey:
Hi @ihhub 
From my point of view, `std::variant` looks exactly as cleaned current `Focus` class but template based and from `std` library (so i think its mean well designed, tested, optimized etc) by one line of code. Also with further refactoring we replace `switch` and multiple `if` statements by `std::visit` pattern which i hope should more efficient, compact and robust (as std::visit must support cases for all alternatives) and maybe we can remove some wrappers - `GetFocusType`, `GetFocus` etc.
For example 


```
void Game::restoreSoundsForCurrentFocus()
{
    AudioManager::ResetAudio();

    switch ( Interface::GetFocusType() ) {
    case GameFocus::HEROES: {
        const Heroes * focusedHero = Interface::GetFocus<Heroes>();
        assert( focusedHero != nullptr );

        const int heroIndexPos = focusedHero->GetIndex();
        if ( heroIndexPos >= 0 ) {
            Game::EnvironmentSoundMixer();
            AudioManager::PlayMusicAsync( MUS::FromGround( world.GetTiles( heroIndexPos ).GetGround() ), Music::PlaybackMode::RESUME_AND_PLAY_INFINITE );
        }
        break;
    }

    case GameFocus::CASTLE: {
        const Castle * focusedCastle = Interface::GetFocus<Castle>();
        assert( focusedCastle != nullptr );

        Game::EnvironmentSoundMixer();
        AudioManager::PlayMusicAsync( MUS::FromGround( world.GetTiles( focusedCastle->GetIndex() ).GetGround() ), Music::PlaybackMode::RESUME_AND_PLAY_INFINITE );
        break;
    }

    default:
        break;
    }
}

```

can be replaced in this manner ([https://en.cppreference.com/w/cpp/utility/variant/visit](https://en.cppreference.com/w/cpp/utility/variant/visit))

```
std::visit(
    []( auto && arg ) {
        using T = std::decay_t<decltype( arg )>;
        if constexpr ( std::is_same_v<T, int> )
            std::cout << "int with value " << arg << '\n';
        else if constexpr ( std::is_same_v<T, long> )
            std::cout << "long with value " << arg << '\n';
        else if constexpr ( std::is_same_v<T, double> )
            std::cout << "double with value " << arg << '\n';
        else if constexpr ( std::is_same_v<T, std::string> )
            std::cout << "std::string with value " << std::quoted( arg ) << '\n';
        else
            static_assert( always_false_v<T>, "non-exhaustive visitor!" );
    },
    w );

```

or even
```
std::visit(overloaded {
            [](auto arg) { std::cout << arg << ' '; },
            [](double arg) { std::cout << std::fixed << arg << ' '; },
            [](const std::string& arg) { std::cout << std::quoted(arg) << ' '; }
        }, v);
```

ihhub:
Hi @vasilenkoalexey ,
we always need to keep in mind code readability. Imagine that another engineer is going to improve the code or fix an issue there. What would be easier to read?

I believe the old fashioned C++ code is much easier to understand than fancy C++17 keywords, especially for people who aren't that familiar with C++17 standard.

On top of this which version would be easier to debug? A template-based code is hard to maintain in such a case.

I am not against code cleanup but I am concerned that this change would add unnecessary complexity in code readability.

I would like to hear @oleg-derevenetz opinion as well.

vasilenkoalexey:
> Imagine that another engineer is going to improve the code or fix an issue there.

I think that another engineer knows the standard library well and shouldn't waste time trying to figure out custom classes that do the same thing. That's why I think using the standard library is better - everyone knows it, everyone knows how it works, everyone knows how to use it. Its portable, designed by professionals, tested, efficient, solve typical problems in the best way.

> On top of this which version would be easier to debug? A template-based code is hard to maintain in such a case.

But... std::pair (which used in Focus class) already template-based, even std::vector template-based. There is no fundamental difference between `std::pair` `std::vector` `std::variant`

In any case, this is a matter of taste, personal experience and subjective perception of the programming language :)

oleg-derevenetz:
Hi @ihhub In general I don't mind modernizing the code. However, since this PR is essentially "modernization for the sake of modernization", and does not solve any problems that may be observable by end users, I suggest postponing this issue after the 1.0 release.

oleg-derevenetz:
> I have a little question - how to distinguish changes which solve not solve any problems that may be observable by end users?

Could you please explain how exactly you change can be observable by the end users (unless breaking something unintentionally)? I don't think we should commit potentially breaking changes if these are just "modernizations for the sake of modernizations" right before the release. We have more "real" work to do (and to test).

> For example, definitely does not solve any problems that may be observable by end users too but it committed.

This change has zero chance to break anything.

ihhub:
The change which made by @oleg-derevenetz was to address SonarQube code smells.

vasilenkoalexey:
Maybe I'm wrong, but looks like any PR can breaking something unintentionally its will be not "real" work which should to do (and to test)

PR #6306, opened 2022-12-05
Revert back regression changes in Blit function

ihhub:
relates to #6303

image

I will need some time to rewrite the logic of the functions properly.


oleg-derevenetz (approved):
Yes, that was a suspicious change :)

PR #6307, opened 2022-12-05
Rework the right mouse button simulation with touch events

oleg-derevenetz:
The idea of the new simulation is as follows:

1. Place the first finger on the target object;
2. Place the second finger anywhere on the screen (not necessarily on the black bar);
3. Now you can remove the first finger from the screen.

This implementation looks more familiar and doesn't require to use black bars for the correct operation. ~It has one flaw though - it doesn't currently support the simulation of moving the mouse with the right button pressed, but, AFAIK, it's not used in the game anyway~ no longer true.

Also this implementation fixes the flaw in the logic of the current implementation I encountered. This logic:

https://github.com/ihhub/fheroes2/blob/1aa3c433b78eedf7c54d89aabb32c14e9019911b/src/engine/localevent.cpp#L1368-L1387

may not work in case of three (or more) fingers and `_numTouches` may become negative. Consider the following situation:

1. I touch the screen by the first finger. The initial check passes because `_numTouches` is `0`, now `_numTouches` is set to `1` and `_firstFingerId` is set to the first finger's id;
2. I touch the screen by the second finger. The initial check passes because `_numTouches` is `1`, now `_numTouches` is set to `2` and `_secondFingerId` is set to the second finger's id;
3. I touch the screen by the third finger. The initial check fails because `_numTouches` is `2` and third finger's id is neither `_firstFingerId` nor `_secondFingerId` and `return` is called. So far so good.
4. I remove the first finger from the screen. The initial check passes because  `_numTouches` is `2`, but first finger's id equals the `_firstFingerId`, now `_numTouches` is set to `1`;
5. I remove the second finger from the screen. The initial check passes because  `_numTouches` is `1`, now `_numTouches` is set to `0`;
6. Now things getting interesting. I remove the third finger from the screen. The initial check passes because  `_numTouches` is `0`, now `_numTouches` is set to `-1` and we are in a complete mess.

### DIFF
--- a/src/fheroes2/gui/interface_buttons.cpp
+++ b/src/fheroes2/gui/interface_buttons.cpp
@@ -143,8 +143,6 @@ void Interface::ButtonsArea::ResetButtons()
     buttonAdventure.drawOnRelease();
     buttonFile.drawOnRelease();
     buttonSystem.drawOnRelease();
-
-    LocalEvent::Get().ResetPressLeft();
 }
 
 fheroes2::GameMode Interface::ButtonsArea::QueueEventProcessing()
--- a/docs/README_android.md
+++ b/docs/README_android.md
@@ -17,10 +17,9 @@ Depending on your original version of Heroes 2 they can either be found on your 
 your Heroes 2 installation directory.
 
 ## Controls
-To simulate a right-click to get info on various items, you need to first touch and keep touching anywhere on the screen
-and then touch on the item of interest. You can slide the touch away while maintaining it to keep viewing the info on the
-item. It is recommended that you place the first touch outside the rendered game area to avoid an accidental left-click
-when you end the simulated right-click.
+To simulate a right-click to get info on various items, you need to first touch and keep touching on the item of interest
+and then touch anywhere else on the screen. You can then remove your first finger from the screen and keep viewing the info
+on the item.
 
 By default normal adventure map scrolling on the borders of the screen is disabled. To pane the viewing area around you
 need to press anywhere on the adventure map and slide around to change where you are viewing.
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -1371,18 +1371,10 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
     switch ( event.type ) {
     case SDL_FINGERDOWN:
         if ( !_fingerIds.first ) {
-            // As soon as the two-finger gesture processing has started, ignore all new touches, even if only
-            // the second finger touches the screen
-            if ( _fingerIds.second ) {
-                return;
-            }
-
             _fingerIds.first = event.fingerId;
         }
         else if ( !_fingerIds.second ) {
             _fingerIds.second = event.fingerId;
-            // As soon as the two-finger gesture processing has started, ignore all events related to the first finger
-            _fingerIds.first.reset();
         }
         else {
             // Gestures of more than two fingers are not supported, ignore
@@ -1404,10 +1396,6 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
         return;
     }
 
-    // These states should be mutually exclusive
-    assert( !_fingerIds.first != !_fingerIds.second );
-
-    // Single-finger gesture, simulate the mouse movement and the left mouse button operation
     if ( event.fingerId == _fingerIds.first ) {
         const fheroes2::Display & display = fheroes2::Display::instance();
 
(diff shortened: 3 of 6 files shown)


ihhub:
I have an idea how to make the black areas active. I'll see if it's easy to implement.

oleg-derevenetz:
@ihhub

> I have an idea how to make the black areas active.

Do you mean extending the adventure map to the entire screen?

ihhub:
> @ihhub
> 
> > I have an idea how to make the black areas active.
> 
> Do you mean extending the adventure map to the entire screen?

Either this or still keep black area but you can click on it.

oleg-derevenetz:
@zenseii

> My only concern is about a left-click happening after you are finished with the right click and you then let go of finger 2 first before finger 1.

The whole idea here that you can just remove the first finger off the screen right after you touch the screen by the second finger.

oleg-derevenetz:
@zenseii Usually on the adventure map the quick info panel appears near the object, almost under the first finger, so users will have to remove the first finger willy-nilly just to be able to read the info from that panel. I think they will get used to it quickly :)

oleg-derevenetz:
@zenseii After all, we may describe this as a feature - the user can immediately click on an object or a button as soon as he looks at the description and makes sure that this is the one :)

oleg-derevenetz:
Hi @zenseii could you please take a look at the updated implementation? Now no left-click events should happen after the two-finger gesture.

Branikolog:
Hi, @oleg-derevenetz !
New right click implementation is quite intuitive and works well. The only issue I've experienced is when I accidentally remove second finger (or intentionally to move the first finger to the different location) and try to use second finger once again - nothing happens. I believe, that would be more convenient, if you can drag the first finger over the adventure map (after you've cancelled first right click without removing the first finger) pointing several objects without taking it off display and initiating multiple right clicks with the second finger. 
I'm not sure, this wouldn't conflict with our adventure map dragging and the previous changes you've made right before though.

oleg-derevenetz:
@Branikolog The whole point of this gesture is that you can freely remove the first finger from the screen after you touch with the second one. Events from the first finger are deliberately ignored after the touch of the second finger precisely in order to avoid accidental "clicks" or "drags" as requested by @zenseii. You should complete the right-click gesture by moving both fingers off the screen before you can start another single-finger or two-finger gesture.

Branikolog:
@oleg-derevenetz 

> @Branikolog The whole point of this gesture is that you can freely remove the first finger from the screen after you touch with the second one. Events from the first finger are deliberately ignored after the touch of the second finger precisely in order to avoid accidental "clicks" or "drags" if the second finger is removed before the first finger as requested by @zenseii. You should complete the right-click gesture by moving both fingers off the screen before you can start another single-finger or two-finger gesture.

Yeah. I got the point. I only proposed to not cancel the right click logic until we move **both** fingers off screen and at the same time don't initiate any other logic like map dragging or object selecting. So this can make possible to initiate several right clicks in a row without moving the first finger off the screen in a case player miskicked using his fingers on a tiny screens. 
I understand, that this is out of scope of the current PR. :)

Branikolog (approved):

oleg-derevenetz:
@Branikolog later today I can try to turn all events related to the first finger to the right-mouse-button events instead of just ignoring them if there is already the second finger on the screen. I hope this won't break anything.

ihhub (approved):

oleg-derevenetz:
@Branikolog

> So this can make possible to initiate several right clicks in a row without moving the first finger off the screen in a case player miskicked using his fingers on a tiny screens.

In fact, this will not work - for the same reason as moving the mouse to the new object while holding the right mouse button will not show info for that objects in the desktop version. Other UI elements will just not get events until the info window will be closed (because it has its own event loop, so these mouse events just will not reach other UI elements), and it will be closed once the right button will be released, and it will be released until the right mouse button will be released (or, which is effectively the same, the second finger will be removed from screen).

Branikolog:
Hello, @oleg-derevenetz !
I'm not sure we understand clearly each other. :) My idea was to allow to move cursor while holding the first finger on the screen and then initiate right click by touching the screen with second finger one more time.
Currently how it works: we press on the target with the first finger - touch screen with the other finger (info window appears) - in a case we release the other finger info window disappears (we're still holding the first finger) - we're trying to move cursor a bit to the side moving the first finger over the screen, as we supposedly missed target with our first finger - nothing changes and we cannot even initiate right click wit the second finger. 
It would be nice if in the situation above we, after releasing the other finger, can move cursor over a visible area of the adventure map with our first finger (to change position of the cursor) and touch screen once again with other finger to initiate the info window of the other object we pointed.
The point of this approach is just to allow player select target without taking his first finger off the screen. I find it a bit too confusing when I need to take both fingers off the screen and then put them back. I think it's more intuitive, when we use one finger to point any object on the screen without taking it off, and the other one just to initiate right click on this object (pressing and releasing it). Using the implementation from this PR makes me sometimes unsure, what finger do I have to put first.
You can answer, this could be a problem of a habit, and you could be probably right. This is just my observations. :)

oleg-derevenetz:
@Branikolog @zenseii you are welcome to try the latest APK from this PR.

Branikolog:
Hi, @oleg-derevenetz !
I've tried the latest version of this PR and everything works nice! Now we have both right click implementation for single info window and at the same time can promptly click several times in a row without removing pointer finger from the screen. 
I agree with @zenseii , I can't imagine how it could be implemented better considering  current UI. Nice work!

oleg-derevenetz:
@Branikolog do you think that this should be somehow described in `README_android.md`?

Branikolog:
@oleg-derevenetz 

> @Branikolog do you think that this should be somehow additionally described in `README_android.md`?

I think it's enough to just mention, that player needs to use two finger touch to initiate right click. It currently works quite intuitive and players could easily find the way, they feel more comfortable: by removing the first finger to uncover info window, or just to slide with it to another place on the screen to be able to re-initiate right click instantly on the other object.

src/fheroes2/gui/interface_buttons.cpp
```diff
@@ -143,8 +143,6 @@ void Interface::ButtonsArea::ResetButtons()
     buttonAdventure.drawOnRelease();
     buttonFile.drawOnRelease();
     buttonSystem.drawOnRelease();
-
-    LocalEvent::Get().ResetPressLeft();
```
oleg-derevenetz, line 147: Hi @ihhub this was an interesting issue :) Sometimes, say, when I exited the hero window, the first of the further clicks did not work on touchscreen. It turned out that it did not work when the cursor was then on the adventure map's button panel. The thing is, click works only when BOTH press and release events are happens on the same element, and this call reset the "press" part. With the mouse, this remained unnoticed because the mouse cursor takes time to get from one place to another, while with touch this happens instantly, so reset was performed right when the click event should be handled.

ihhub: Hi @oleg-derevenetz , I tested the changes and looks like that code was useless.

ihhub (approved):

ihhub:
@oleg-derevenetz , great thanks for improving this logic!

PR #6308, opened 2022-12-06
Android: do not backup fheroes2.bin

oleg-derevenetz:
This file is no longer in use and should not normally exist.


ihhub (approved):

ihhub:

PR #6319, opened 2022-12-07
Do not stop object animation during AI turn

ihhub:
If AI movement is not shown or AI heroes are hidden by fog Adventure Map is not updated so all objects become frozen. The changes allow to continue animation of objects for such cases.

Progress bar for AI turn is updated based on heroes' turns. It also renders when Adventure Map objects require an update so we do not waste time for rendering a frame just to indicate some progress and then few nanoseconds later we do the same.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -37,11 +37,13 @@
 #include "artifact.h"
 #include "castle.h"
 #include "color.h"
+#include "game_interface.h"
 #include "game_over.h"
 #include "game_static.h"
 #include "gamedefs.h"
 #include "ground.h"
 #include "heroes.h"
+#include "interface_status.h"
 #include "kingdom.h"
 #include "logging.h"
 #include "luck.h"
@@ -1742,7 +1744,7 @@ namespace AI
         }
     }
 
-    bool Normal::HeroesTurn( VecHeroes & heroes )
+    bool Normal::HeroesTurn( VecHeroes & heroes, const uint32_t startProgressValue, const uint32_t endProgressValue )
     {
         if ( heroes.empty() ) {
             // No heroes so we indicate that all heroes moved.
@@ -1759,6 +1761,8 @@ namespace AI
 
         const int monsterStrengthMultiplierCount = 2;
         const double monsterStrengthMultipliers[monsterStrengthMultiplierCount] = { ARMY_ADVANTAGE_MEDIUM, ARMY_ADVANTAGE_SMALL };
+
+        Interface::StatusWindow & status = Interface::Basic::Get().GetStatusWindow();
 
         while ( !availableHeroes.empty() ) {
             Heroes * bestHero = availableHeroes.front().hero;
@@ -1860,12 +1864,20 @@ namespace AI
 
             _pathfinder.setArmyStrengthMultiplier( originalMonsterStrengthMultiplier );
             _pathfinder.setSpellPointReserve( 0.5 );
+
+            // The size of heroes can be increased if a new hero is released from Jail.
+            const size_t maxHeroCount = std::max( heroes.size(), availableHeroes.size() );
+            const size_t progressValue = ( endProgressValue - startProgressValue ) * ( maxHeroCount - availableHeroes.size() ) / maxHeroCount + startProgressValue;
+
+            status.RedrawStatusIfNeeded( static_cast<uint32_t>( progressValue ) );
         }
 
         const bool allHeroesMoved = availableHeroes.empty();
 
         _pathfinder.setArmyStrengthMultiplier( originalMonsterStrengthMultiplier );
         _pathfinder.setSpellPointReserve( 0.5 );
+
+        status.RedrawStatusIfNeeded( endProgressValue );
 
         return allHeroesMoved;
     }
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1867,9 +1867,13 @@ namespace AI
 
             // The size of heroes can be increased if a new hero is released from Jail.
             const size_t maxHeroCount = std::max( heroes.size(), availableHeroes.size() );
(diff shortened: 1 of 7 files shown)


ihhub:
I realized that during AI we render up to 200 FPS... I need to fix it.

LeHerosInconnu:
Hello @ihhub,

A question about this.
Since the movement of AI heroes is not shown to the human player when they are under the fog of war, there is no need to make their animation under the fog of war.
How is this done currently, is the move generated or the hero sprite only "moved" to its final destination?

LeHerosInconnu:
Hello @zenseii,

> Hi, @LeHerosInconnu.
> 
> > How is this done currently, is the move generated or the hero sprite only "moved" to its final destination?
> 
> The AI hero is just moved. The reason I know this is because that's what happens when you have fog "disabled" in debug mode. So you don't see the actual movement only the result after their turn even though they are visible on the adventure map

Okay, thanks for the clarification. :)

Branikolog (approved):

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1860,13 +1864,21 @@ namespace AI
 
             _pathfinder.setArmyStrengthMultiplier( originalMonsterStrengthMultiplier );
             _pathfinder.setSpellPointReserve( 0.5 );
+
+            // The size of heroes can be increased if a new hero is released from Jail.
+            const size_t maxHeroCount = std::max( heroes.size(), availableHeroes.size() );
+            const size_t progressValue = ( endProgressValue - startProgressValue ) * ( maxHeroCount - availableHeroes.size() ) / maxHeroCount + startProgressValue;
```
oleg-derevenetz, line 1870: Hi @ihhub Could there in theory be division by zero if kingdom has no heroes?

src/fheroes2/gui/interface_status.cpp
```diff
@@ -470,10 +473,22 @@ void Interface::StatusWindow::TimerEventProcessing()
...
 {
-    turn_progress = v;
+    turn_progress = progressValue;
 
     interface.Redraw( REDRAW_STATUS );
-    fheroes2::Display::instance().render( GetArea() );
+
+    if ( Game::validateAnimationDelay( Game::MAPS_DELAY ) ) {
+        // Process events if any before rendering a frame. For instance, updating a mouse cursor position.
+        LocalEvent::Get().HandleEvents( false );
```
oleg-derevenetz, line 484: I'm not sure about that, but, since we handle the events here, why wait for `MAPS_DELAY` for this? We can just handle events each time when this function is called.

src/fheroes2/gui/interface_status.cpp
```diff
@@ -470,10 +473,22 @@ void Interface::StatusWindow::TimerEventProcessing()
...
+
+        uint32_t & frame = Game::MapsAnimationFrame();
+        ++frame;
+
+        interface.GetGameArea().SetRedraw();
+
+        interface.Redraw();
```
oleg-derevenetz, line 491: This can be simplified to `interface.Redraw( REDRAW_GAMEAREA );`, there is not need for separate `SetRedraw()` call. Or we can do the following:

```cpp
interface.SetRedraw( REDRAW_STATUS );

if ( Game::validateAnimationDelay( Game::MAPS_DELAY ) ) {
[...]
  interface.SetRedraw( REDRAW_GAMEAREA ); // Or interface.GetGameArea().SetRedraw(), which is effectively the same
}

interface.Redraw();
fheroes2::Display::instance().render();
```

What do you think?

oleg-derevenetz (commented):
Hi @ihhub I just left a few more questions, could you please take a look?

src/fheroes2/gui/interface_status.cpp
ihhub: The first correction is fine but the second one will do unnecessary rendering which this PR is trying to avoid.

src/fheroes2/gui/interface_status.cpp
ihhub: Sure. I put the call at the beginning of the function.

src/fheroes2/gui/interface_status.cpp
oleg-derevenetz: The current logic leads to the situation when the updated hourglass will not be rendered even if `interface.Redraw( REDRAW_STATUS )` is called, but `MAPS_DELAY` hasn't passed yet. Is this OK?

ihhub: Yes, this is the intention. If AI takes less than 250 ms per turn so we don't even draw it as it's instant.

ihhub: How the previous logic worked: we rendered minimum 4 frames during each AI turn even if the whole processing speed was less than a millisecond.

However, if we want to render at least one frame per each AI player I can add rendering after the first call of this method.

oleg-derevenetz: > However, if we want to render at least one frame per each AI player I can add rendering after the first call of this method.

I don't know is this necessary if the player's eye will not be able to catch this.

ihhub: Human reaction is about 250 ms in average :)

src/fheroes2/gui/interface_status.h
```diff
@@ -63,7 +63,7 @@ namespace Interface
 
         void SetState( const StatusType status );
         void SetResource( int, uint32_t );
-        void RedrawTurnProgress( uint32_t );
+        void RedrawStatusIfNeeded( const uint32_t progressValue );
```
oleg-derevenetz, line 66: Just one optional suggestion - maybe name it something like `[Re]drawAITurnProgress[IfNeeded]`, because current name is not very clear - if this is the first time you see this code, will you be able to tell right off the bat what the difference is between `RedrawStatusIfNeeded()`, `Redraw()` or `SetRedraw()`? :) LGTM otherwise.

oleg-derevenetz (approved):

src/fheroes2/gui/interface_status.h
ihhub: Done.

PR #6331, opened 2022-12-08
Make File Dialog and Adventure Dialog buttons translatable

zenseii:
Both Evil and Good interfaces.

The cancel button will be handled later.

Relates to #6105 

image
image

PR #6335, opened 2022-12-09
Update Russian translation (main menu and dialog buttons)

Districh-ru:
Translated the newly introduced translatable main menu buttons.
I used the translation of "Buka" for these buttons, but the text does not fit some buttons.

The result in fheroes2:
изображение изображение изображение изображение

изображение изображение изображение

изображение

<details><summary>Russian version of Heroes 2</summary>

heroes2

</details>
Hello @Branikolog, @undef21, @a1exsh. Please write, when you have time, if you have any suggestions for this translation.


Branikolog:
Hi, @Districh-ru !
I definitely have some thoughts regarding translations, but nevertheless we also need to squeeze font as much as we can.  There're lots of buttons we can make thinner as well as space between them. We've met difficulties with space in main menu, where buttons are large. Other in game buttons will definitely have more problems.)

ihhub:
> Hi, @Districh-ru !
> I definitely have some thoughts regarding translations, but nevertheless we also need to squeeze font as much as we can. There're lots of buttons we can make thinner as well as space between them. We've met difficulties with space in main menu, where buttons are large. Other in game buttons will definitely have more problems.)

Hi @Branikolog , we won't squeeze font for a particular language. Every language should use the same font for consistency.

Recently @zenseii fixed spacing for button's font reducing it by 4 pixels.

ihhub:
Strictly speaking our normal fonts aren't the same :) We can reduce spacing between every letter for button's font.

Districh-ru:
Hi, @Branikolog, @ihhub, @zenseii!
Problems with lack of space in translated phrases often arise. Russian phrases and words are longer than English ones, and we have to shorten them to fit in the given place.
For example, we have this issue in the battle log, which could be solved by introducing multiline messages:
изображение
And if we try to shorten the long phase, we have room to think, which we don't have for buttons:
изображение

We could shorten some letters in the main menu font, for example: 
изображение -> изображение
изображение  -> изображение
изображение -> изображение
изображение -> изображение
изображение -> изображение
изображение -> изображение

Now in Russian font we have narrow 'П' and 'Л': 
изображение   изображение :)

But we could also introduce a narrow font (for each font style) for such cases. I don’t remember in which game, but I saw such an implementation where a narrow font was used for long phases.

Districh-ru:
> Hi, @Districh-ru.
> 
> For the MAKC button we have a solution, but there will be another much more interesting button in the form of the castle exit button. BTW could you please sync this branch up against master as soon as #6331 ? I believe that dialog will be much more interesting to translate if I remember correctly from me and @Branikolog's previous discussions on the topic.
> 
> Also thank you for providing the shortened letters. It is indeed a mess right now with how wide they are.

@zenseii, I gave an example here with the MAX button just to illustrate possible problems, implying that in the future we will have to change size of many buttons like this. :)
Once #6331 is merged into master, I'll sync this branch and add the relevant translations for discussion.

And talking about letters width, IMHO the letters with "two" vertical lines (У, П, Н, К, И, Ц, Д, А ...) sould have the same width and the letters with "three" vertical lines (Ы, Ш, Щ, М, Ю, Ж) can be wider, but not much.

Districh-ru:
> Yep this is true. I'm just pointing out that we have a function that will adjust the length of the button according to the translated text, but currently this is only for single-lined buttons.
And here we have to check for case if the text is wide and the button beomes wider, will it fit the dialog box or the screen.

> > And talking about letters width, IMHO the letters with "two" vertical lines (У, П, Н, К, И, Ц, Д, А ...) sould have the same width and the letters with "three" vertical lines (Ы, Ш, Щ, М, Ю, Ж) can be wider, but not much.
> 
> Yes, I believe this is in fact be the standard rule for such fonts. The problem is that they were made by someone who is not familiar with this font at all, aka me, so I was hoping someone more familiar would come and point out how they need to be adjusted.
@zenseii, you have done a great work by adding cp-1251 letters to the font. 👍 In your PR it looks great.
And now when we tried to use the font in translation, we ran into some width issues. Therefore, I wrote only my point of view on the possible width of letters in this font. The letters "M", "H", "O", "P" are taken from the original English font, but for me they are wide and could be smaller.

Districh-ru:
If we want to keep the fonts consistent, we can only adjust the width of new cp-1251 characters:
изображение

Branikolog:
Hi, @Districh-ru and @zenseii !
I believe, that we should make all letters a bit thinner, than the original font, to make overall text style consistent and to avoid future problems with rest buttons translations.. Right now, such letters as П, Л, А have even smaller size _ _ decorations, which would save ~2px wideness for almost each symbol. I propose to adjust other letters in the same manner as well. Besides making some of them thinner as well, like @Districh-ru showed on screenshots.

ihhub:
Hi @Districh-ru , @Branikolog and @zenseii , what's the conclusion here? Are we going with these changes which will be applied for non-Russian assets or we delay it till we fix letters?

Districh-ru:
Hi, @ihhub. The better way is to fix the letters and then edit and apply this PR. But editing the letters is not a fast process.
So we could discuss the translation of buttons for a day or two and apply this PR. Maybe we will confirm only the part of translation, and other part will be moved from this PR to a new one.  Later, after fixing letters, we could return to the "problem" buttons.

Branikolog:
Hi, @ihhub 
I agree with @Districh-ru . Merging this PR will make buttons look not well, until we fix some letters and tighten words.
I believe, before merging this we should open a separate PR regarding Cyrillic letters, making them thinner and uniform-style.
We can proceed with discussing translation here for some buttons.
We can also accept some translations, that fit buttons, but I'm not sure, it would look well, if we leave 50% buttons of translated, and 50% - not.

ihhub:
Let's merge this pull request after fixing letters for the Russian font.

ihhub:
I'm moving this pull request to 1.0.1 release as it requires additional changes in font which we aren't going to make before the upcoming release.

ihhub:
Hi @Branikolog , could you please re-review this pull request?

Also @a1exsh and @undef21 would you mind to check the result of this pull request?

files/lang/ru.po
```diff
@@ -3699,13 +3721,12 @@ msgstr "Посмотреть очки рейтинга за кампании."
...
 msgstr "Этот файл содержит сохранение с некорректным типом игры."
 
-#, fuzzy
 msgid ""
 "This file was saved for a \"The Price of Loyalty\" map, but the "
 "corresponding game assets have not been provided to the engine."
 msgstr ""
-"Этот файл сохранен в версии \"Цена верности\".\n"
-"Некоторые элементы игры будут недоступны."
+"Этот файл сохранён для карты дополнения \"Цена верности\", но движку "
```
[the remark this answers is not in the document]
Districh-ru: Hello, @undef21. Yes, "движок" is a jargon and it could be changed to "программе" here.
But it is named like this in many places, i.e.: https://www.opennet.ru/opennews/art.shtml?num=58630
I used it because it is already used in phrase: `"Добро пожаловать в Heroes of Might and Magic II на движке fheroes2!"`
So if we want to change this jargon we also have to think about the change of "движке" in this phase to some other word? ... but "на программе fheroes2!" does not sound nice. :(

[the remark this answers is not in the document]
undef21: >but "на программе fheroes2!" does not sound nice. 

You are right. How about "Этот файл сохранён для карты из дополнения \"Цена верности\", но игре не предоставлены соответствущие ресурсы"? Or "...дополнения \"Цена верности\", но соответствующие игровые ресурсы не обнаружены"?

files/lang/ru.po
```diff
@@ -3699,13 +3721,12 @@ msgstr "Посмотреть очки рейтинга за кампании."
...
+"fheroes2 не предоставлены соответствующие игровые ресурсы."
```
Districh-ru, line 3729:
```suggestion
"Этот файл сохранён для карты дополнения \"Цена верности\", но "
"соответствующие игровые ресурсы не обнаружены."
```

files/lang/ru.po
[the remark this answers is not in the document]
Districh-ru: @undef21, thanks! The last variant is good, I'll change the phrase with the next commit.

Branikolog:
Hi, @Districh-ru !
I have a few proposals regarding the translation of buttons. I'd like to hear your opinion.
"Сведения о карте"
image
The description mentions scenario, not map. Can we simply call this button "СЦЕНАРИЙ"{I hope it would fit)? To my mind it suits well, since "сценарий" commonly means the plot of the thing.

image
Since we're going to add new campaign, these translations would make a mess.
I propose add the names of campaigns:
"ВОЙНЫ ЗА ПРЕСТОЛ"
"ЦЕНА ВЕРНОСТИ"

image
I propose show only "КАМПАНИЯ" without "играть". There're no more other verbs on these buttons. "Кампания", "Обычная игра" would look better together being nouns.

Regarding multiplayer game.... I think, "Много Игроков" doesn't suit well, looking on two buttons above.... Ideally, name of each button should introduce a complete and separate definition... So with this logic it should be like "Игра на много игроков" which is definitely too long and not grammatically pleasing...
What do you think about naming this mode "СОВМЕСТНАЯ ИГРА"?
In my personal opinion the best translation could be "многопользовательская игра" but it's way too long. "Сетевая игра" doesn't suit well, since we can play not only through network, but on the same PC without any connection.

A little offtop:
image
I think, if we align the transferred line without consideration of hyphen, the overall text on button would look more balanced and centered. I mean center all word symbols and then add a hyphen to the right side of transferred word.
How do you think? Is it possible to implement?

files/lang/ru.po
```diff
@@ -31,120 +31,142 @@ msgid ""
...
 "GAME"
 msgstr ""
+"НОВАЯ\n"
+"ИГРА"
 
 msgid ""
 "SAVE\n"
 "GAME"
 msgstr ""
+"СОХРА-\n"
```
Districh-ru, line 41:
```suggestion
" СОХРА-\n"
```

files/lang/ru.po
```diff
@@ -31,120 +31,142 @@ msgid ""
...
+"НИТЬ\n"
+"ИГРУ"
 
 msgid ""
 "LOAD\n"
 "GAME"
 msgstr ""
+"ЗАГРУ-\n"
```
Districh-ru, line 49:
```suggestion
" ЗАГРУ-\n"
```

files/lang/ru.po
```diff
@@ -31,120 +31,142 @@ msgid ""
...
 msgstr ""
+"ОБЫЧНАЯ\n"
+"ИГРА"
 
 msgid ""
 "CAMPAIGN\n"
 "GAME"
 msgstr ""
+"ИГРАТЬ\n"
+"КАМПАНИЮ"
```
Districh-ru, line 124:
```suggestion
msgstr "КАМПАНИЯ"
```

files/lang/ru.po
```diff
@@ -31,120 +31,142 @@ msgid ""
...
 
 msgid ""
 "MULTI-\n"
 "PLAYER\n"
 "GAME"
 msgstr ""
+"МНОГО\n"
+"ИГРОКОВ"
```
Districh-ru, line 132:
```suggestion
"СОВМЕСТНАЯ\n"
"ИГРА"
```

files/lang/ru.po
```diff
@@ -31,120 +31,142 @@ msgid ""
...
 msgid "CONFIG"
-msgstr ""
+msgstr "НАСТРОЙКИ"
 
 msgid ""
 "ORIGINAL\n"
 "CAMPAIGN"
 msgstr ""
+"СТАРАЯ\n"
+"КАМПАНИЯ"
```
Districh-ru, line 142:
```suggestion
"ВОЙНЫ ЗА\n"
"ПРЕСТОЛ"
```

files/lang/ru.po
```diff
@@ -31,120 +31,142 @@ msgid ""
...
 
 msgid ""
 "EXPANSION\n"
 "CAMPAIGN"
 msgstr ""
+"НОВАЯ\n"
+"КАМПАНИЯ"
```
Districh-ru, line 149:
```suggestion
"ЦЕНА\n"
"ВЕРНОСТИ"
```

Districh-ru:
Hi, @Branikolog, thanks for the proposals! I like them.

изображение изображение изображение

But "СЦЕНАРИЙ" does not fit:
изображение   **->**  изображение

It could be "ИНФОР-\nМАЦИЯ" - but it also does not look good.

Lets hope that in near future the size of these buttons will be changed to fit this as well as "ЗАГРУЗИТЬ" and "СОХРАНИТЬ" :)

Branikolog (approved):
Looks pretty well!
I hope I won't forget to create an issue to update these buttons' descriptions.

ihhub:
@Districh-ru , thank you so much for improving the translation!

PR #6336, opened 2022-12-09

oleg-derevenetz:
Make SonarCloud happier.

### DIFF
--- a/src/fheroes2/gui/interface_itemsbar.h
+++ b/src/fheroes2/gui/interface_itemsbar.h
@@ -438,7 +438,7 @@ namespace Interface
 
         bool ActionCrossItemBarDrag( const fheroes2::Point & cursor, ItemsActionBar<Item> & other )
         {
-            LocalEvent & le = LocalEvent::Get();
+            const LocalEvent & le = LocalEvent::Get();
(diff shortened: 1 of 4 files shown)


src/fheroes2/gui/interface_itemsbar.h
```diff
@@ -438,13 +438,10 @@ namespace Interface
...
-            LocalEvent & le = LocalEvent::Get();
-            Item * otherItemPress = other.GetItem( le.GetMousePressLeft() );
+            const LocalEvent & le = LocalEvent::Get();
 
-            // already did check for this before we go here, maybe not necessary?
-            if ( !otherItemPress ) {
-                return false;
-            }
+            Item * otherItemPress = other.GetItem( le.GetMousePressLeft() );
+            assert( otherItemPress != nullptr );
```
oleg-derevenetz, line 444: The only place where `ActionCrossItemBarDrag()` is called is located in the `QueueEventProcessing()` and there is already check that `other.GetItem( le.GetMousePressLeft() )` is not `nullptr`:

```cpp
if ( other.GetItem( le.GetMousePressLeft() ) && ActionCrossItemBarDrag( cursor, other ) ) {
    return true;
}
```

ihhub:

PR #6342, opened 2022-12-10
Tidy up hotkeys constants a bit, make separate hotkeys to cast battle & adventure spells, add the "Quit game" hotkey, handle hotkeys in the Adventure Map's File dialog

### LINKED ISSUE #6313: [Controls] Hotkey Inconsistency Issues
* _Split from #6310_

### Preliminary checks

### Platform

Windows

### Describe the bug

There's a lot of hotkeys in the original _Heroes II_ that do various things.  For example, hitting "F" will open the "File Menu" where one can load or save their game, quit, or start a new game.  In any case, it seems like _most_ of the hotkeys work as expected.  However, there are two small problems I've noticed:

1. If the player opens the File sub-menu and tries to use any of the expected hotkeys (N, Q, L, S) from within, they won't work.
2. There is no "Quit Game" hotkey.  At all. (Default "Q" in _Heroes II_.)
2b. However, in _fheroes2_, hitting Escape while on the Adventure Map _does_ open the Quit Game prompt.  This doesn't work in _Heroes II_.
3. In the original version of the game, both H and Tab would select the next Hero with usable movement points.  By default, _fheroes2_ only uses "H" for this.

Nothing earth-shattering.  Just some observations.

oleg-derevenetz:
fix #6313


Branikolog (approved):

ihhub (approved):

ihhub:

PR #6344, opened 2022-12-10
Death Wave spell animation fix (Battlefield)

### LINKED ISSUE #1041: "Holy Word" and "Holy Shout" spells should have darkening in a middle phase of animation
While using these spells screen darkens a bit in the middle of animation. (Or, probably, increasing contrast?)
Holy shout
While our version has none:

Holy shout2

### LINKED ISSUE #5701: Combat screen, Death Wave spell animation, differences from the original game
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, there are differences from the original game for Death Wave spell animation,
Note: I haven't tested with Death Ripple, but it's certainly the same.

In the original game, the damage animation (red sprites) is played at the same time as the animation of the creature taking damage.

In fheroes2, the damage animation is done first and then the animation of the creatures that take damage.

In original game:

https://user-images.githubusercontent.com/43583994/180996575-0759e2e7-e78d-43e3-9738-d82204c57dbc.mp4

In fheroes2:

https://user-images.githubusercontent.com/43583994/180996628-791c53a0-ea80-4141-92c8-de84ff894b55.mp4

https://user-images.githubusercontent.com/43583994/180996843-6838d397-ed7e-41e0-a370-f368096c535a.mp4

In fheroes2, the animation of the damage (red sprites) is not at the same position and is reversed horizontally.
In fheroes2, the animation of the troll taking damage is less pronounced  (the posture of the troll is not as inclined) than in Heroes 2.

In original game:

https://user-images.githubusercontent.com/43583994/180997160-67d770f0-1ae3-4f89-bd92-03d9a6142d01.mp4

In fheroes2:

https://user-images.githubusercontent.com/43583994/180997127-365f0054-619f-428b-8196-36517456de22.mp4

In original game:
vlcsnap-2022-07-26-13h12m30s667

In fheroes2:
Damage red sprites reversed horizontally.
vlcsnap-2022-07-26-13h12m23s766

In original game:
vlcsnap-2022-07-26-13h11m05s545

In fheroes2:
Red sprites position.
vlcsnap-2022-07-26-13h09m59s257

In original game:
vlcsnap-2022-07-26-13h10m36s831

In fheroes2:
The posture of the troll is not as inclined.
vlcsnap-2022-07-26-13h09m47s883
vlcsnap-2022-07-26-13h10m17s731

The wave effect that runs across the screen is also much less pronounced in the original game.
In fheroes2, the artwork of the battlefield is completely distorted, which is not very aesthetic.

In original game:
vlcsnap-2022-07-26-14h05m16s077

In fheroes2:
vlcsnap-2022-07-26-14h05m22s623

### Save file

[0 9928.zip](https://github.com/ihhub/fheroes2/files/9189078/0.9928.zip)

### Additional info

Version 0.9.17 5674

### LINKED ISSUE #5710: Mac installation but GOG only offers Windows HoMM2 install?
### Your question

Hello, 
I was able to install `fheroes2` on my M1 macbook pro via `homebrew` without a problem. My next step is to get the original HoMM2 game installed. The documentation recommends I install the [GOG game version](https://www.gog.com/game/heroes_of_might_and_magic_2_gold_edition) but it is only installable on Windows. How does this work? Thank you!

### Additional info

_No response_

Districh-ru:
Close #5701 and close #1041.
Relates to #1229.

Redone the animation sequence for Death Wave, Death Ripple, Holy Word and Holy Shout shout.
Now the flame animation (after the spell animation) is made only after applying damage from the spell. And during this flame animation the death animation and the wince animation is applied to the corresponding creatures.
To do this I had to make a copy of part of the code from `RedrawActionWincesKills` method. :(
To reduce the size of this duplication a new (small) method was made.

fheroes2 Death Death Ripple and Death Wave animations:

https://user-images.githubusercontent.com/113276641/206888610-90a2e068-487d-49cc-9e8a-469f80a587c5.mp4

Original game Death Ripple and Death Wave animations:

https://user-images.githubusercontent.com/113276641/206887219-44e3a4c3-8a93-4f9e-b54c-9c91eb0d64a4.mp4

<details><summary>fheroes2 Holy Word and Holy Shout animation:</summary>

https://user-images.githubusercontent.com/113276641/207246675-ed7064b1-1e7a-4606-82ae-378b12baf6a2.mp4

https://user-images.githubusercontent.com/113276641/207246687-81fa1b97-fc39-40ce-99d9-09cfb551d7ad.mp4

</details>

<details><summary>Original game Holy Word and Holy Shout animation:</summary>

https://user-images.githubusercontent.com/113276641/207246918-7dad4d13-04c6-4487-916b-559976be864d.mp4

https://user-images.githubusercontent.com/113276641/207246934-e32ef987-9709-4a4a-b963-47db28756617.mp4

</details>

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -5225,7 +5225,7 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
     if ( wnce ) {
         int32_t deathColor = Color::UNUSED;
 
-        for ( const auto & target : targets ) {
+        for ( const TargetInfo & target : targets ) {
             Unit * defender = target.defender;
             if ( defender == nullptr ) {
                 continue;
@@ -5250,7 +5250,7 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
             }
             // If the creature is damaged but is still alive set its wince animation.
             else if ( target.damage ) {
-                defender->SwitchAnimation( Monster_Info::WNCE );
+                defender->SwitchAnimation( Monster_Info::WNCE_UP );
                 AudioManager::PlaySound( defender->M82Wnce() );
             }
 
@@ -5290,7 +5290,7 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
             isDefenderAnimatimg = false;
 
             if ( wnce ) {
-                for ( const auto & target : targets ) {
+                for ( const TargetInfo & target : targets ) {
                     if ( target.defender == nullptr ) {
                         continue;
                     }
@@ -5300,12 +5300,21 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
                         target.defender->IncreaseAnimFrame( false );
                     }
                     else if ( target.damage ) {
-                        // Check if the animation will not be finished with the next frame, then set the defender wince animation state.
-                        isDefenderAnimatimg |= !target.defender->isFinishAnimFrame();
+                        // Check if the current animation is finished.
+                        const bool isFinishAnim = target.defender->isFinishAnimFrame();
+                        if ( target.defender->GetAnimationState() == Monster_Info::WNCE_DOWN && isFinishAnim ) {
+                            target.defender->SwitchAnimation( Monster_Info::STATIC );
+                        }
 
-                        // Animate only 2 wince frames at the start of the spell animation to get the maximum wince effect
-                        // and continue it only after spell animation ends.
-                        if ( frame < 2 || frame >= maxFrame ) {
+                        // If the main spell sprite animation and WNCE_UP are finised then switch unit animation to WNCE_DOWN.
+                        if ( frame >= maxFrame && isFinishAnim && target.defender->GetAnimationState() == Monster_Info::WNCE_UP ) {
+                            target.defender->SwitchAnimation( Monster_Info::WNCE_DOWN );
+                        }
+
+                        // If not all damaged units are set to STATIC animation then set isDefenderAnimatimg.
+                        isDefenderAnimatimg |= !( target.defender->GetAnimationState() == Monster_Info::STATIC );
+
+                        if ( !target.defender->isFinishAnimFrame() ) {
                             target.defender->IncreaseAnimFrame( false );
                         }
                     }
@@ -5317,18 +5326,9 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
         }
     }
 
-    if ( wnce ) {
-        for ( const auto & target : targets ) {
-            if ( target.defender && target.defender->GetAnimationState() == Monster_Info::WNCE ) {
-                target.defender->SwitchAnimation( Monster_Info::STATIC );
-                _currentUnit = nullptr;
-            }
-        }
-
+    if ( wnce && !mirrorImages.empty() ) {
         // Fade away animation for destroyed mirror images
-        if ( !mirrorImages.empty() ) {
-            RedrawActionRemoveMirrorImage( mirrorImages );
-        }
+        RedrawActionRemoveMirrorImage( mirrorImages );
     }
 }
 
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3363,18 +3363,17 @@ void Battle::Interface::SetHeroAnimationReactionToTroopDeath( const int32_t deat
     if ( deathColor == Color::UNUSED ) {
         return;
     }
-    else {
-        const bool attackersTurn = ( deathColor == arena.GetArmy2Color() );
-        OpponentSprite * attackingHero = attackersTurn ? opponent1 : opponent2;
-        OpponentSprite * defendingHero = attackersTurn ? opponent2 : opponent1;
-        // 60% of joyful animation
-        if ( attackingHero && Rand::Get( 1, 5 ) < 4 ) {
-            attackingHero->SetAnimation( OP_JOY );
-        }
-        // 80% of sorrow animation otherwise
-        else if ( defendingHero && Rand::Get( 1, 5 ) < 5 ) {
-            defendingHero->SetAnimation( OP_SORROW );
-        }
+
+    const bool attackersTurn = ( deathColor == arena.GetArmy2Color() );
+    OpponentSprite * attackingHero = attackersTurn ? opponent1 : opponent2;
+    OpponentSprite * defendingHero = attackersTurn ? opponent2 : opponent1;
+    // 60% of joyful animation
+    if ( attackingHero && Rand::Get( 1, 5 ) < 4 ) {
+        attackingHero->SetAnimation( OP_JOY );
+    }
+    // 80% of sorrow animation otherwise
+    else if ( defendingHero && Rand::Get( 1, 5 ) < 5 ) {
+        defendingHero->SetAnimation( OP_SORROW );
     }
 }
 
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -4756,6 +4756,9 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
     std::vector<int32_t> deathWaveCurve;
     deathWaveCurve.reserve( waveLength );
 
+    // Calculate the "Death Wave" curve as one period of cosine, which starts from 0 with an amlutude of 1/2 and shifted down by 0.5.
+    // So we get a smooth hill, which is the multiplied with 'strength' and shifted after that by -1px.
+    // (The "Death Wave" curve has to shift the image and cosine starts from 0 so we add extra 1px).
     for ( int32_t posX = 0; posX < waveLength; ++posX ) {
         deathWaveCurve.push_back( static_cast<int32_t>( std::round( strength * ( cos( posX / waveLimit ) / 2 - 0.5 ) ) ) - 1 );
     }
@@ -4822,7 +4825,7 @@ void Battle::Interface::RedrawActionColdRingSpell( int32_t dst, const TargetsInf
         }
 }
 
-void Battle::Interface::RedrawActionHolyShoutSpell( const int strength )
+void Battle::Interface::RedrawActionHolyShoutSpell( const uint8_t strength )
 {
     Cursor & cursor = Cursor::Get();
     LocalEvent & le = LocalEvent::Get();
@@ -4839,7 +4842,7 @@ void Battle::Interface::RedrawActionHolyShoutSpell( const int strength )
     // Make the spell effect more dark-red.
     fheroes2::Image blurredRed( blurred );
     fheroes2::ApplyPalette( blurredRed, PAL::GetPalette( PAL::PaletteType::RED ) );
-    fheroes2::AlphaBlit( blurredRed, blurred, static_cast<uint8_t>( 10 * strength ) );
+    fheroes2::AlphaBlit( blurredRed, blurred, ( 10 * strength ) );
 
     _currentUnit = nullptr;
     AudioManager::PlaySound( M82::MASSCURS );
@@ -5188,7 +5191,7 @@ fheroes2::Point CalculateSpellPosition( const Battle::Unit & target, int spellIC
         break;
     case ICN::REDDEATH:
         // Shift spell sprite position for wide ceature to its head.
-        result.x += pos.width / 2 + ( target.isReflect() ? 1 - spellSprite.width() - 2 * spellSprite.x() - pos.width / 8 : pos.width / 8 );
+        result.x += pos.width / 2 + ( target.isReflect() ? ( 1 - spellSprite.width() - 2 * spellSprite.x() - pos.width / 8 ) : ( pos.width / 8 ) );
         result.y -= pos.height - 4;
         break;
     case ICN::MAGIC08:
@@ -5261,7 +5264,7 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
     // For certain spells reflect the spell sprite if the creature is reflected.
     const bool isReflectICN = ( icn == ICN::SHIELD || icn == ICN::REDDEATH || icn == ICN::MAGIC08 );
     // Set the defender wince animation state.
-    bool isDefenderAnimatimg = wnce;
+    bool isDefenderAnimating = wnce;
     const uint32_t maxFrame = fheroes2::AGG::GetICNCount( icn );
     uint32_t frame = 0;
 
@@ -5269,7 +5272,7 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
 
     Game::passAnimationDelay( Game::BATTLE_SPELL_DELAY );
 
-    while ( le.HandleEvents() && ( frame < maxFrame || isDefenderAnimatimg ) ) {
+    while ( le.HandleEvents() && ( frame < maxFrame || isDefenderAnimating ) ) {
         CheckGlobalEvents( le );
 
         if ( Game::validateAnimationDelay( Game::BATTLE_SPELL_DELAY ) ) {
@@ -5286,8 +5289,10 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
                 }
             }
 
+            RedrawPartialFinish();
+
             // Reset the defender wince animation state.
-            isDefenderAnimatimg = false;
+            isDefenderAnimating = false;
 
             if ( wnce ) {
                 for ( const TargetInfo & target : targets ) {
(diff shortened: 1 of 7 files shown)


oleg-derevenetz:
> close https://github.com/ihhub/fheroes2/issues/5710

#5701 :)

Districh-ru:
Hello, everyone.
After opening this PR I noticed that after the creatures death the flame animation shifts down. It is because the spell position is calculated for every frame based on creatures sprite. I'll try to fix this tomorrow.

Districh-ru:
> > close #5710
> 
> #5701 :)

Hi, @oleg-derevenetz.  thanks for finding the typo. :)

Districh-ru:
> I think, Death Wave (level 3) should have larger and faster wave effect than smaller Death Ripple(level 2) spell.

Hi, @Alucard648.
In original game both Death Wave and Death Ripple have the same speed (5px/frame), Death Ripple height is about 7px and Death Wave height is about 14px.
I used this parameters in the current implementation.

Here is a video from original game for comparison:


https://user-images.githubusercontent.com/113276641/206887219-44e3a4c3-8a93-4f9e-b54c-9c91eb0d64a4.mp4

Districh-ru:
Hello, @Branikolog and @LeHerosInconnu!
Could you please review this PR when you have time.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +223,45 @@ namespace fheroes2
...
-            uint8_t * outImageY = outImageX + offsetOut;
-            uint8_t * outTransformY = outTransformX + offsetOut;
-            const uint8_t * outImageYEnd = outImageX + offsetOutEnd;
+        // Set pointers to the start and the end of the death wave curve.
+        std::vector<int32_t>::const_iterator pntX = deathWaveCurve.begin() + ( x < waveWidth ? waveWidth - x : 0 );
+        const std::vector<int32_t>::const_iterator endX = deathWaveCurve.end() - ( x > width ? x - width : 0 );
 
-            const int32_t offsetIn = offsetY >= 0 ? 0 : -offsetY * width;
+        for ( ; pntX != endX; ++pntX, ++outImageX, ++inImageX ) {
+            const uint8_t * outImageYEnd = outImageX + static_cast<int64_t>( height + *pntX ) * width;
```
oleg-derevenetz, line 250: It's highly unlikely that the result of this multiplication will overflow `int`. However if you want to silence the clang-tidy, it would probably be better to convert to `std::ptrdiff_t` instead (here and on the following line), because currently on 32-bit machines there is one extra conversion performed - first to `int64_t` and then, when adding the multiplication result to pointer, it will be converted once again to `std::ptrdiff_t` which is 32-bit on these machines.

Districh-ru: Hi @oleg-derevenetz, thanks for the explanations. I'll replace <int64_t> with <ptrdiff_t> to have less type conversions on 32 systems.
But, please, help me to make these lines more clean (without possible overflow).
If I convert the left part of the multiplication to <ptrdiff_t> I thought that the right part will be automatically converted. Or I should do the same conversion for the right part (`static_cast<ptrdiff_t>( width )`)?

oleg-derevenetz: > If I convert the left part of the multiplication to <ptrdiff_t> I thought that the right part will be automatically converted.

Yes, both operands of arithmetic operation are promoted/converted to the same type first, so there is no need to manually convert each operand. See the "Conversions" section here:

https://en.cppreference.com/w/cpp/language/operator_arithmetic

and the link regarding integral promotion in that section.

Branikolog:
Hello, @Districh-ru !
I've already made few tests.
Here's an issue I've found:
Red fog is drawn over troops standing in front of the target.
Wave

In the OG:
image

Districh-ru:
> Hello, @Districh-ru ! I've already made few tests. Here's an issue I've found: Red fog is drawn over troops standing in front of the target. 

It is because the `RedrawArmies();` method , which draws the whole army on the battlefield is called first and then the red fog for is drawn.
To fix this issue we need to redesign `RedrawArmies();` method to add ability to draw each unit one by one and spells one by one from the farest to the closest. 
But we have spells with more than one cell blast radius, for example the Cold Ring spell:

https://user-images.githubusercontent.com/113276641/206905837-62154df6-3be4-4e35-a583-330cb0c20526.mp4

It should be drawn under the second peasant, but if the peasant was one cell upper, it should be drawn over him.
I think this issue will happen to every spell animation.

ihhub:
Hi @Districh-ru , the best strategy to avoid heavy redevelopment is to do in steps. I believe that restructuring of `RedrawArmies()` method will require a lot of code changes. If yes it would be wiser to do it in a separate pull request.

Branikolog:
@Districh-ru 
I've also noticed differences with splash animations position:
ezgif com-gif-maker (99)
image

Districh-ru:
> @Districh-ru I've also noticed differences with splash animations position

@Branikolog, I tried to fix it, but I think OG has some kind of coordinate table for such sprite positions for each creature. So I set the horizontal position of the sprites close to OG. For the vertical position, I used the unit's cell position because I don't want to rely on the sprite's parameters. Therefore, the vertical position is not as close to OG. But, with such sprites (fog or explosions), it's probably better to snap to grid coordinates, not the parameters of the creature.

Districh-ru:
> Hi @Districh-ru , the best strategy to avoid heavy redevelopment is to do in steps. I believe that restructuring of `RedrawArmies()` method will require a lot of code changes. If yes it would be wiser to do it in a separate pull request.

Hi @ihhub, I think there will be changes that will affect the displaying creatures on the battlefield in different situations. We also have to decide which spell and under what conditions should be drawn under or over each creature.

ihhub:
> > Hi @Districh-ru , the best strategy to avoid heavy redevelopment is to do in steps. I believe that restructuring of `RedrawArmies()` method will require a lot of code changes. If yes it would be wiser to do it in a separate pull request.
> 
> Hi @ihhub, I think there will be changes that will affect the displaying creatures on the battlefield in different situations. We also have to decide which spell and under what conditions should be drawn under or over each creature.

I think we should do this in a separate PR.

Branikolog:
Hi, @Districh-ru !
Now spell effect has consistent position for both sides.
image
Currently, spell effect appears on the back side of the creature sprite. I find it a bit shifted, since it appears not over the creature, but a little behind it. How do you think, is it intended animation, or can we shift those red demons appear closer to the middle of creature hex (for 1-hex creatures) and to the border of adjoining hexes (for 2-hex creatures)?

Branikolog:
Also, I've noticed some differences in creatures sprites animations.
While red splash appears, all creatures perform "taking damage" animation. They freeze for a while being leaned back and then return to standing state.
Original Game:
https://user-images.githubusercontent.com/55348946/207298697-6b73cf64-4aa0-45df-ab5d-e942f02cc3c5.mp4
fheroes2:
https://user-images.githubusercontent.com/55348946/207298791-1e38f33b-53f0-4a6f-9354-bfa8d0e6a8d7.mp4
Take a look at halflings and champions in the current implementation.
Halflings freeze in a state which is not the most leaned back, and after freezing they continue leaning back and then returning to standing state.
Champions are on a contrary, freeze in a state after maximum leaning back. It looks like animation stops on the last "taking damage" frame.

Districh-ru:
Hi, @Branikolog!
I'll shift the red fog sprite as you suggested and try to place it in the middle of two-hex creatures.
Also I'll will do a more thorough analysis of the frame number with maximum wince for each creature.

Branikolog:
Hi, @Districh-ru 
Now wincing seems to happen in correct timings, but It looks like there's last frame missing. Or it happens too promptly, so I couldn't even notice it.

https://user-images.githubusercontent.com/55348946/207544248-ea7226e5-7410-4f84-99da-f3e68ee8cbbd.mp4

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +224,45 @@ namespace fheroes2
...
             return Image();
 
         Image out = in;
 
         const int32_t width = in.width();
-        const int32_t height = in.height();
-
-        if ( x + waveWidth < 0 || x - waveWidth >= width )
+        // If the death wave curve is outside of the battlefield - return the original image.
+        if ( x < 0 || ( x - waveWidth ) >= width )
```
ihhub, line 236: If `deathWaveCurve` is empty we don't need to do anything but just return from the function.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +224,45 @@ namespace fheroes2
...
+        const int32_t offsetX = x < waveWidth ? 0 : x - waveWidth;
+        uint8_t * outImageX = out.image() + offsetX;
+        const uint8_t * inImageX = in.image() + offsetX;
 
-            uint8_t * outImageY = outImageX + offsetOut;
-            uint8_t * outTransformY = outTransformX + offsetOut;
-            const uint8_t * outImageYEnd = outImageX + offsetOutEnd;
+        // Set pointers to the start and the end of the death wave curve.
+        std::vector<int32_t>::const_iterator pntX = deathWaveCurve.begin() + ( x < waveWidth ? waveWidth - x : 0 );
+        const std::vector<int32_t>::const_iterator endX = deathWaveCurve.end() - ( x > width ? x - width : 0 );
```
ihhub, line 248: Here we don't check the boundaries of our container. We have to check that `( x < waveWidth ? waveWidth - x : 0 )` is smaller than `deathWaveCurve.size()` (or clamp to its size) and the same for `( x > width ? x - width : 0 )`.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +224,45 @@ namespace fheroes2
...
+            }
 
-            for ( ; outImageY != outImageYEnd; outImageY += width, outTransformY += width, inImageY += width, inTransformY += width ) {
+            // Flip the image under the death wave to create a distortion effect.
+            for ( int32_t i = 0; i > *pntX; --i ) {
                 *outImageY = *inImageY;
-                *outTransformY = *inTransformY;
+                outImageY += width;
+                inImageY -= width;
             }
```
ihhub, line 266: Would we have out of boundaries read if `*pntX` value is bigger than `height`?

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +224,45 @@ namespace fheroes2
...
-            uint8_t * outTransformY = outTransformX + offsetOut;
-            const uint8_t * outImageYEnd = outImageX + offsetOutEnd;
+        // Set pointers to the start and the end of the death wave curve.
+        std::vector<int32_t>::const_iterator pntX = deathWaveCurve.begin() + ( x < waveWidth ? waveWidth - x : 0 );
+        const std::vector<int32_t>::const_iterator endX = deathWaveCurve.end() - ( x > width ? x - width : 0 );
 
-            const int32_t offsetIn = offsetY >= 0 ? 0 : -offsetY * width;
+        for ( ; pntX != endX; ++pntX, ++outImageX, ++inImageX ) {
+            const uint8_t * outImageYEnd = outImageX + static_cast<ptrdiff_t>( height + *pntX ) * width;
+            const uint8_t * inImageY = inImageX - static_cast<ptrdiff_t>( *pntX + 1 ) * width;
```
ihhub, line 252: I think we should have a loop to check that all values in `deathWaveCurve` are within accepted range of values. At least in debug mode.

ihhub (changes requested):
Hi @Districh-ru, I left several comments in this pull request. Could you please take a look at them in your spare time?

Districh-ru:
> Hi, @Districh-ru Now wincing seems to happen in correct timings, but It looks like there's last frame missing. Or it happens too promptly, so I couldn't even notice it.

Hi, @Branikolog. Yes, there is a small logic problem: I change the type of animation and then I increase the frame number. It will be fixed in next commit.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5165,50 +5215,121 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
...
+            }
+
+            SetHeroAnimationReactionToTroopDeath( deathColor );
+        }
+    }
+
+    // For certain spells reflect the spell sprite if the creature is reflected.
+    const bool isReflectICN = ( icn == ICN::SHIELD || icn == ICN::REDDEATH || icn == ICN::MAGIC08 );
+    // Set the defender wince animation state.
+    bool isDefenderAnimatimg = wnce;
```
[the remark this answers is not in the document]
Districh-ru: I left a misprint here. It has to be `isDefenderAnimating`.
I named this variable so because it is true if for all units their animation sequence is not finished.

src/fheroes2/gui/ui_tool.cpp
Districh-ru: If it is so and if it is outside of the battlefield we return the input image unchanged, as the caller function waits for an image.
I'll add the check for the size of deathWaveCurve.

src/fheroes2/gui/ui_tool.cpp
Districh-ru: I'll add a check for each *pntX if it is in the boundaries of the input image at the start of the loop. After this *pntX could not et higher, than the height.

Branikolog (approved):
Duration of the wave and red fog is still a bit different, from the original game, but it could be polished later. Current fix is a huge improvement!
Great work, @Districh-ru !

Districh-ru:
Thanks, @Branikolog!
I set the step of Death Wave equal to the original game. The original game has different speed settings and, I think, different frame delays when drawing spell effects. Yes, later we can more carefully set this delays.
And did you check the corrected spell effect for Holy Word and Holy Shout? Does this PR fix the #1041 issue? :)

Branikolog:
@Districh-ru 

> And did you check the corrected spell effect for Holy Word and Holy Shout? Does this PR fix the #1041 issue? :)

Yes. Spells look 95% identical.
I've only noticed that in the OG there's a slight difference in colour between the Word and the Shout spells. The Shout spell has a little bit darker (or adding red colur) in the middle of animation. I've tried to make several screenshots of the OG and every time middle frame has slight colour difference, but I'm not sure, if I simply missed needed frame and spells are 100% identical.
ezgif com-gif-maker (100)

Anyway, this difference is barely noticeable. (Like ripple and wave for me. :) )

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3370,6 +3358,23 @@ void Battle::Interface::RedrawActionWincesKills( const TargetsInfo & targets, Un
         RedrawActionRemoveMirrorImage( mirrorImages );
 }
 
+void Battle::Interface::SetHeroAnimationReactionToTroopDeath( const int32_t deathColor )
+{
+    if ( deathColor != Color::UNUSED ) {
```
ihhub, line 3363: This is optional but to reduce the indentation we can do this:
```cpp
if ( deathColor == Color::UNUSED ) {
    return;
}
....
The rest of code
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3370,6 +3358,23 @@ void Battle::Interface::RedrawActionWincesKills( const TargetsInfo & targets, Un
...
+        const bool attackersTurn = deathColor == arena.GetArmy2Color();
```
ihhub, line 3364: Please add parenthesis for this line: `const bool attackersTurn = ( deathColor == arena.GetArmy2Color() );` to be easier to read.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3773,7 +3778,21 @@ void Battle::Interface::RedrawActionSpellCastPart2( const Spell & spell, const T
...
 
         // targets damage animation
-        RedrawActionWincesKills( targets );
+        switch ( spell.GetID() ) {
+            // For some spells the damage animation is done during the flame animation after the spell animation.
+        case Spell::DEATHRIPPLE:
+        case Spell::DEATHWAVE:
+            RedrawTargetsWithFrameAnimation( targets, ICN::REDDEATH, M82::UNKNOWN, true );
+            break;
+
```
ihhub, line 3787: To be consistent please either add an empty line before `default` or remove empty line here.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4714,34 +4733,49 @@ void Battle::Interface::RedrawActionDisruptingRaySpell( const Unit & target )
...
     AudioManager::PlaySound( M82::MNRDEATH );
 
-    int position = 10;
+    int32_t position = 0;
     while ( le.HandleEvents() && position < area.width + waveLength ) {
         CheckGlobalEvents( le );
 
         if ( Game::validateAnimationDelay( Game::BATTLE_DISRUPTING_DELAY ) ) {
-            fheroes2::Blit( fheroes2::CreateDeathWaveEffect( copy, position, waveLength, strength ), _mainSurface );
+            fheroes2::Blit( fheroes2::CreateDeathWaveEffect( copy, position, waveLength, deathWaveCurve ), _mainSurface );
```
ihhub, line 4773: Here we are wasting resources for copying a full frame for no reason. It is not related to your chances as it is an old code. Can you please add a TODO comment here for the future optimization:
```cpp
// TODO: instead of rendering the whole frame for the wave effect we should render only the area where the effect is active.
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5165,50 +5218,124 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
...
+                for ( const TargetInfo & target : targets ) {
+                    if ( target.defender == nullptr ) {
+                        continue;
+                    }
+
+                    // Fully animate creature death.
+                    if ( !target.defender->isValid() ) {
+                        target.defender->IncreaseAnimFrame( false );
+
+                        // If the death animation is still in process then set isDefenderAnimatimg to false.
```
ihhub, line 5307: Please correct this comment: `isDefenderAnimatimg` --> `isDefenderAnimating`.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5165,50 +5218,124 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
...
         }
     }
 
-    if ( wnce )
-        for ( TargetsInfo::const_iterator it = targets.begin(); it != targets.end(); ++it )
-            if ( ( *it ).defender ) {
-                ( *it ).defender->SwitchAnimation( Monster_Info::STATIC );
-                _currentUnit = nullptr;
-            }
+    if ( wnce && !mirrorImages.empty() ) {
```
ihhub, line 5335: We can remove `wnce` check since we populate `mirrorImages` only when `wnce` is **true**.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +226,51 @@ namespace fheroes2
         }
     }
 
-    Image CreateDeathWaveEffect( const Image & in, int32_t x, int32_t waveWidth, int32_t waveHeight )
+    Image CreateDeathWaveEffect( const Image & in, const int32_t x, const int32_t waveWidth, const std::vector<int32_t> & deathWaveCurve )
```
ihhub, line 229: According to the code we create `deathWaveCurve` of the size of `waveWidth`. In this case I suggest to remove `waveWidth` as an input argument and declare this variable within the function:
```cpp
const int32_t waveWidth = static_cast<int32_t>( deathWaveCurve.size() );
```

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -223,51 +226,51 @@ namespace fheroes2
...
+        std::vector<int32_t>::const_iterator pntX = deathWaveCurve.begin() + ( x < waveWidth ? waveWidth - x : 0 );
+        const std::vector<int32_t>::const_iterator endX = deathWaveCurve.end() - ( x > width ? x - width : 0 );
 
-            uint8_t * outImageY = outImageX + offsetOut;
-            uint8_t * outTransformY = outTransformX + offsetOut;
-            const uint8_t * outImageYEnd = outImageX + offsetOutEnd;
+        for ( ; pntX != endX; ++pntX, ++outImageX, ++inImageX ) {
+            // The death curve should have only negative values and should not be higher, than the height of 'in' image.
+            if ( ( *pntX >= 0 ) || ( *pntX <= -height ) ) {
+                DEBUG_LOG( DBG_BATTLE, DBG_WARN, "Death Wave curve is out of range (" << -height << ", 0): " << *pntX )
```
ihhub, line 255: Based on the past problems with the code I suggest to just add `assert( 0 );` instead of logging the error so the issue will be immediately detected and fixed. If such a case happens with logging we would be a lot of log messages as this function is called many times as well as many values could be present.

ihhub (changes requested):
Hi @Districh-ru , I put few more comments in this pull request. Can you please take a look at them when you have time?

ihhub:
Hi @Districh-ru , I updated your initial message to link #1041 and to be closed upon merging this pull request.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: I'll try to do this in a new PR.

Districh-ru:
> Hi @Districh-ru , I updated your initial message to link #1041 and to be closed upon merging this pull request.

Hi. @ihhub. Thanks!

src/fheroes2/battle/battle_interface.cpp
ihhub: Yes, we don't need to do it here.

ihhub (approved):

ihhub:
@Districh-ru , huge thanks for upgrading these animations!

Issue #6124, opened 2022-11-08
Lord Corlagon is missing in the last Roland scenario


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In the original game 9th scenario in a case we don't defeat Lord Corlagon (he should run away from the battle) and win the scenario, he should appear in the last campaign scenario.
OG:
![image](https://user-images.githubusercontent.com/55348946/200588298-f31b4c2c-8f82-4f72-8de0-b02efa232705.png)
![image](https://user-images.githubusercontent.com/55348946/200588377-6c1f4628-0a23-498b-8ae0-92843a6d6de7.png)
![image](https://user-images.githubusercontent.com/55348946/200588574-135ca29f-9b7b-4b5c-9f53-c697686b2f51.png)
![image](https://user-images.githubusercontent.com/55348946/200591343-e4a39dac-9726-403b-8d46-e19566b73e74.png)

Currently, in fheroes2 he's always absent in the last scenario.

### Save file

[Good_9 final no corlagon.zip](https://github.com/ihhub/fheroes2/files/9962056/Good_9.final.no.corlagon.zip)

PR #6352, opened 2022-12-12
Implement the infrastructure to apply campaign bonuses and awards to a specific hero, multiple campaign-related fixes

### LINKED ISSUE #6027: Incorrect bonus skill application in campaign (and also different from OG order of starting heroes)
### Preliminary checks

### Platform

Windows

### Describe the bug

While playing the last scenario "Blood is Thicker" of expansion campaign "Voyage Home" I noticed that the selected secondary skill bonus is applied not to the main hero, but to the second hero (his sister). She is a Necromancer, and if I choose Base Necromancy bonus, it applies to a Necromancer who already has the Necromancy skill.
изображение
изображение

In the original game the order of the heroes is different: the first is the sister and the second is the main hero (in fheroes2 the main hero is the first).
изображение
And in the original game secondary skill bonus applies to the main hero:
изображение

### Save file

[vh4_0001.zip](https://github.com/ihhub/fheroes2/files/9845976/vh4_0001.zip)

### Additional info

The original game adds heroes to their kingdom in the order they are loaded from the map file. In fheroes2, heroes are always added in a predefined order (`vec_heroes`).
I tried changing the load order of the heroes (commented `vec_kingdoms.AddHeroes( vec_heroes );` and added `GetKingdom(hero->GetColor()).AddHeroes(hero);` after `hero->LoadFromMP2`.
As a result, the hero order became like in the original game (on this map and on the others), but the problem with incorrect skill application has not gone away.

oleg-derevenetz:
fix #6027

Currently (before this PR) campaign bonuses and awards are applied to the "best hero of the kingdom", which usually means "hero with the most powerful army". This PR introduces some sort of "infrastructure" to set the specific hero to get these bonuses & awards (with the fallback to the old behavior if no hero specified or found). After a quick run through the campaigns I found the following places (and fixed these places in this PR):

* Roland's campaign, mission "Defender" - there are two heroes, and Sorceress should get the bonus spell;
* Roland's campaign, mission "Final Justice" - there are multiple heroes, and Roland should get the bonus artifact, as well as army from "The Gauntlet";
* Archibald's campaign, mission "Apocalypse" - there are multiple heroes, and Archibald should get the bonus artifact, as well as army from the previous mission;
* Descendants campaign, mission "The Wayward Son" - there are multiple heroes, and Sorceress should get the bonus;
* Descendants campaign, mission "The Epic Battle" - there are multiple heroes, and Sorceress should get the bonus;
* Voyage Home campaign, mission "Blood is Thicker" - there are two heroes, and Gallavant should get the bonus skill.

Maybe there are another places, and maybe I missed something regarding the missions listed above. Hi @LeHerosInconnu, @Branikolog what do you think?

### DIFF
--- a/src/fheroes2/campaign/campaign_scenariodata.h
+++ b/src/fheroes2/campaign/campaign_scenariodata.h
@@ -103,7 +103,7 @@ namespace Campaign
         int32_t _type;
         int32_t _subType;
         int32_t _amount;
-        int32_t _spellId; // Spell ID of a spell scroll
+        int32_t _artifactSpellId; // Spell ID of a spell scroll
 
         ScenarioBonusData();
         ScenarioBonusData( const int32_t type, const int32_t subType, const int32_t amount );
--- a/src/fheroes2/campaign/campaign_scenariodata.h
+++ b/src/fheroes2/campaign/campaign_scenariodata.h
@@ -103,9 +103,11 @@ namespace Campaign
         int32_t _type;
         int32_t _subType;
         int32_t _amount;
+        int32_t _spellId; // Spell ID of a spell scroll
 
         ScenarioBonusData();
         ScenarioBonusData( const int32_t type, const int32_t subType, const int32_t amount );
+        ScenarioBonusData( const int32_t type, const int32_t subType, const int32_t amount, const int32_t spellId );
 
         friend StreamBase & operator<<( StreamBase & msg, const ScenarioBonusData & data );
         friend StreamBase & operator>>( StreamBase & msg, ScenarioBonusData & data );
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -630,7 +630,6 @@ namespace
                 assert( hero != nullptr );
 
                 if ( hero != nullptr ) {
-                    hero->SpellBookActivate();
                     hero->AppendSpellToBook( scenarioBonus._subType, true );
                 }
 
@@ -706,7 +705,6 @@ namespace
                 assert( hero != nullptr );
 
                 if ( hero != nullptr ) {
-                    hero->SpellBookActivate();
                     hero->AppendSpellToBook( awards[i]._subType, true );
                 }
 
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -530,7 +530,7 @@ namespace
         }
     }
 
-    Heroes * getHeroForScenarioBonusOrCampaignAwards( const Campaign::ScenarioInfoId & scenarioInfoId, const Kingdom & kingdom )
+    Heroes * getHeroToApplyBonusOrAwards( const Campaign::ScenarioInfoId & scenarioInfoId, const Kingdom & kingdom )
     {
         static const std::map<std::pair<int, int>, int> targetHeroes = { // Defender
                                                                          { { Campaign::ROLAND_CAMPAIGN, 5 }, Heroes::HALTON },
@@ -588,7 +588,7 @@ namespace
 
                 break;
             case Campaign::ScenarioBonusData::ARTIFACT: {
-                Heroes * hero = getHeroForScenarioBonusOrCampaignAwards( scenarioInfoId, kingdom );
+                Heroes * hero = getHeroToApplyBonusOrAwards( scenarioInfoId, kingdom );
                 assert( hero != nullptr );
 
                 if ( hero != nullptr ) {
@@ -598,7 +598,7 @@ namespace
                 break;
             }
             case Campaign::ScenarioBonusData::TROOP: {
-                Heroes * hero = getHeroForScenarioBonusOrCampaignAwards( scenarioInfoId, kingdom );
+                Heroes * hero = getHeroToApplyBonusOrAwards( scenarioInfoId, kingdom );
                 assert( hero != nullptr );
 
                 if ( hero != nullptr ) {
@@ -608,7 +608,7 @@ namespace
                 break;
             }
             case Campaign::ScenarioBonusData::SPELL: {
-                Heroes * hero = getHeroForScenarioBonusOrCampaignAwards( scenarioInfoId, kingdom );
+                Heroes * hero = getHeroToApplyBonusOrAwards( scenarioInfoId, kingdom );
                 assert( hero != nullptr );
 
                 if ( hero != nullptr ) {
@@ -623,7 +623,7 @@ namespace
 
                 break;
             case Campaign::ScenarioBonusData::STARTING_RACE_AND_ARMY: {
-                Heroes * hero = getHeroForScenarioBonusOrCampaignAwards( scenarioInfoId, kingdom );
+                Heroes * hero = getHeroToApplyBonusOrAwards( scenarioInfoId, kingdom );
                 assert( hero != nullptr );
 
                 if ( hero != nullptr ) {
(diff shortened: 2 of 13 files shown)


oleg-derevenetz:
@zenseii

> is this PR in any way relatable to the issue about the Corlagon award always being active in the final scenario of the Roland campaign? https://github.com/ihhub/fheroes2/issues/6124

No. One issue at a time :)

oleg-derevenetz:
> Is there any place we can reference to find out what scenarios should have what? I guess walkthroughs or just cheating through the original works.

I just did the quick run by modifying the code so I always win at the first turn. However I had to use the Debug Hero to test the army transfer between scenarios.

Districh-ru:
Hi, @zenseii, @oleg-derevenetz.
My issue is fixed. 👍 
<details><summary>Now I'm able to give Gallavant the Necromancy skill :)</summary>

изображение

</details>

Branikolog (approved):

src/fheroes2/game/game_campaign.cpp
```diff
@@ -530,6 +530,62 @@ namespace
...
+                                                                             { { Campaign::ARCHIBALD_CAMPAIGN, 10 }, Heroes::ARCHIBALD },
+                                                                             // Blood is Thicker
+                                                                             { { Campaign::VOYAGE_HOME_CAMPAIGN, 3 }, Heroes::GALLAVANT } };
+
+            const auto iter = targetHeroes.find( { scenarioInfoId.campaignId, scenarioInfoId.scenarioId } );
+            if ( iter != targetHeroes.end() ) {
+                // The "special" kingdom heroes may have the ID of another hero, but a custom name and portrait,
+                // so the search should be performed by the portrait ID
+                for ( Heroes * hero : kingdom.GetHeroes() ) {
+                    assert( hero != nullptr );
```
ihhub, line 548: In the modified methods below we use an assertion and a check while we do not use a check. Nothing wrong with this behavior but I would suggest to have the same logic everywhere.

ihhub (approved):
Hi @oleg-derevenetz , the code looks good to me. I left one optional comment about consistency but it is up to you if you want to address it.

src/fheroes2/game/game_campaign.cpp
oleg-derevenetz: Hi @ihhub `SetScenarioBonus()` used both the assertion and the check, while `applyObtainedCampaignAwards()` used neither assertion nor check, just the raw result of `kingdom.GetBestHero()`, which was suspicious, because `GetBestHero()` will return `nullptr` when there is no heroes in the kingdom. I decided to make everything uniform and choose the logic from `SetScenarioBonus()`, because, while there shouldn't be a situation when a campaign map has no heroes and `assert()` is fine for it, maps are still sort of an external data, so we need to handle the situation when map is wrong even on release build and do not crash on `nullptr` access.

ihhub: Thanks for the explanation!

src/fheroes2/heroes/heroes.cpp
```diff
@@ -192,13 +192,12 @@ Heroes::Heroes()
     , _aiRole( Role::HUNTER )
 {}
 
-Heroes::Heroes( int heroID, int race, int initialLevel )
+Heroes::Heroes( const int heroID, const int race, const int initialExperience )
     : Heroes( heroID, race )
```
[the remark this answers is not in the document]
oleg-derevenetz: Haha LOL Clang-Tidy has no idea how delegating constructors [work](https://en.cppreference.com/w/cpp/language/constructor#Explanation).

oleg-derevenetz:
Hi @zenseii

> I know stats and things change for every reload, but I couldn't find one reload in which he began with 4500 exp which he always seems to do in fheroes2.

This is because in fheroes2 such heroes were created with the level (5) specified instead of XP (5000), and the lower bound for level 5 is 4500XP. Now they are created by specifying XP instead of level. Still need to check if they have 5000XP in other scenarios (such as "Necromancer Guild" reward in Archibald campaign, etc) - they always have level 5, but I don't remember the specific amount of XP.

> Also notice in fheroes2 the Ivan hero starts with 11/20 mana.

This should be fixed as well. This is due to the spell point auto-replenishment (all heroes always replenish 1 spell point per turn, even if they are not hired, plus additional spell points depending on their Mysticism skill), and this auto-replenishment worked also at the day one, which is technically incorrect. Now it works only starting from day 2.

Branikolog:
Hi, @oleg-derevenetz 
I've also found some inconsistencies with campaign bonuses as well.
Roland/Archibald in the last original campaign mission starts with not sufficient mana points, while starting with the Crown.

image

PoL 7 mission bonuses include "Animate dead" spell. In the OG hero is provided with spell scroll, while in fheroes2 it's in spell book.
fheroes2:
image
OG:
image

Wizards Isle last mission provides "Earth Guardian" spell for starting hero. In the OG starting hero has no spell book, but when he buys one, the spell will be there.
image

fheroes2 currently provides a spell book to a hero with the spell inside.
image

oleg-derevenetz:
Hi @Branikolog

> Roland/Archibald in the last original campaign mission starts with not sufficient mana points, while starting with the Crown.

Exactly the same as in the original game:

crown

as well as in other similar cases in OG, like that case with the Legendary Scepter described by @zenseii in his review above. It looks like in campaign heroes never get the full spell points immediately if they receive additional Knowledge due to an artifact. Probably because they get this artifact just at the day one according to the scenario, so they have no time to replenish full spell points.

> In the OG starting hero has no spell book, but when he buys one, the spell will be there.
> fheroes2 currently provides a spell book to a hero with the spell inside.

Do you really think that this behavior should be altered?

oleg-derevenetz:
@Branikolog

> True. But it's inconsistent with current fheroes2 behaviour, which gives hero maximum possible spell points at the start of single scenario. I am not sure, actually, that maximum points doesn't ruin campaign walkthrough, but full spell points reserve looks more correct, and it's consistent with single scenarios.

Then it's not a bug. At best, this could be considered as an improvement.

Branikolog:
@oleg-derevenetz 

>Then it's not a bug. At best, this could be considered as an improvement.

Agreed. :)
Since we improved starting mana points for single scenarios, I think, we should improve game logic for campaigns as well.

oleg-derevenetz:
@Branikolog

> PoL 7 mission bonuses include "Animate dead" spell. In the OG hero is provided with spell scroll, while in fheroes2 it's in spell book.

This inconsistency should be fixed now.

> Wizards Isle last mission provides "Earth Guardian" spell for starting hero. In the OG starting hero has no spell book, but when he buys one, the spell will be there.

This inconsistency should be fixed now as well.

> Since we improved starting mana points for single scenarios, I think, we should improve game logic for campaigns as well.

I propose to postpone improvements of such kind after the 1.0 release. Let's fix real bugs first :)

Branikolog (approved):

src/fheroes2/campaign/campaign_scenariodata.h
```diff
@@ -103,9 +103,11 @@ namespace Campaign
         int32_t _type;
         int32_t _subType;
         int32_t _amount;
+        int32_t _spellId; // Spell ID of a spell scroll
```
ihhub, line 106: How about we name it `_artifactSpellId` to distiquish from normal spells?

ihhub (commented):
Hi @oleg-derevenetz , I left just one suggestion. Can you please take a look?

ihhub (approved):

ihhub:
@oleg-derevenetz , great thanks for these improvements!

PR #6353, opened 2022-12-12
Small hero movement code cleanup

ihhub:
Add small optimizations by caching values from functions and change some comments. The original intention was to reduce the amount of rendering calls during hero's movement but some more important tasks appeared.


src/fheroes2/heroes/heroes_move.cpp
```diff
@@ -970,29 +971,33 @@ void Heroes::FadeIn( const fheroes2::Point & offset ) const
     _alphaValue = 255;
 }
 
-bool Heroes::Move( bool fast )
+bool Heroes::Move( const bool jumpToNextTile )
```
oleg-derevenetz, line 974:
```suggestion
bool Heroes::Move( const bool jumpToNextTile /* = false */ )
```

oleg-derevenetz (approved):
Hi @ihhub I just left one optional suggestion, LGTM in general.

PR #6346, opened 2022-12-11
Fix idle animation stops (Battlefield)

Districh-ru:
Close #1229
Continue idle troop animation with every 'RedrawPartialStart'.
<details><summary>The creatures continue using idle animation while catapult attacks. ▼</summary>

https://user-images.githubusercontent.com/113276641/206894036-04656850-5f91-4f1f-8c44-b15d89704816.mp4

</details>
Also some methods do a screen copy before performing an animation and uses it during the whole animation.
This methods should change the animation from 'IDLE' to 'STATIC' themselves.
For the Death Wave spell I'll try to fix it in #6344

PR #6355, opened 2022-12-12
Fix animation of slowed flying creatures (Battlefield)

### LINKED ISSUE #2314: Slowed flying creature animation bug
Vampires don't move smoothly over battlefield. They turn into bat, bat starts waving with its wings, then bat appears the next hex. Then jumps the last one (Vampires have speed 2) and there we can see transformation.
[Vampires issue.zip](https://github.com/ihhub/fheroes2/files/5635963/Vampires.issue.zip)

Districh-ru:
close #2314

This PR:

https://user-images.githubusercontent.com/113276641/207112267-00b0f38f-7662-4bed-8820-8d42774e986a.mp4

<details><summary>For comparison</summary>
Original game:

https://user-images.githubusercontent.com/113276641/206899380-adbcfa89-4c82-4428-96a3-6d091241db26.mp4

fheroes2:

https://user-images.githubusercontent.com/113276641/206899463-7753d88e-6232-483a-9234-c3f3a84f67ac.mp4

</details>


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3392,6 +3398,20 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
     _movingUnit = &unit;
 
+    // If it is a flying creature that acts like walking one when it is under the Slow spell.
+    const bool canFly = unit.isAbilityPresent( fheroes2::MonsterAbilityType::FLYING );
+
+    // Slowed flying creature has to fly off.
+    if ( canFly && dst == path.begin() ) {
+        // The destination of fly off is same as the current creature position.
+        const fheroes2::Rect & pos1 = unit.GetRectPosition();
+        _movingPos = fheroes2::Point( pos1.x, pos1.y );
```
ihhub, line 3408: Please use `getPosition()` method here.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3392,6 +3398,20 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
     _currentUnit = nullptr;
     _movingUnit = &unit;
 
+    // If it is a flying creature that acts like walking one when it is under the Slow spell.
+    const bool canFly = unit.isAbilityPresent( fheroes2::MonsterAbilityType::FLYING );
+
+    // Slowed flying creature has to fly off.
+    if ( canFly && dst == path.begin() ) {
+        // The destination of fly off is same as the current creature position.
+        const fheroes2::Rect & pos1 = unit.GetRectPosition();
```
ihhub, line 3407: Please try to avoid using cryptic names. We can name is as `unitArea`.

src/fheroes2/battle/battle_interface.cpp
ihhub: Even better:
```cpp
_movingPos = unit.GetRectPosition().getPosition();
```
then we don't need an extra variable.

ihhub (changes requested):
Hi @Districh-ru , I left very minor few comments. Could you please check them?

ihhub (approved):

Branikolog:
Hello, @Districh-ru !
I've found a few issues in current implementation. Could you, please, look at them, when you have time?
Flying back issue:

https://user-images.githubusercontent.com/55348946/207599081-1f330da5-404f-43e5-adc5-b78453130772.mp4

Seems to happen only for 2-hex troops.

Also, taking off sprites sequence should be made after sprite is mirrored for moving back. Currently, game shows taking off animation, and after that sprite turns back.

Also, I've noticed, that sometimes landing animation is missing. Couldn't find out the exact conditions for that yet...

https://user-images.githubusercontent.com/55348946/207601490-bf4a8c87-ebef-4929-a57c-4d6540294d8e.mp4

Districh-ru:
Hello, @Branikolog!
The two-hex creatures are hard to deal with...
I was unable to reproduce the bug with non-landing sprites (or other slow flying ones).
I also did not like the behavior of the bridge: the flyer changes its animation to standing without  landing and then continues to fly without flying off. After opening or closing the bridge, the creature acts the same way.
So I changed this behavior - now the bridge opens before the creature takes off and closes after it lands.

In video I forced the game to assume that all creatures are flightlessб so they have more movement points then in slow mode.

https://user-images.githubusercontent.com/113276641/207692994-3f4df457-29ac-4d5c-8149-281b571ca25c.mp4

Branikolog:
@Districh-ru 

>So I changed this behavior - now the bridge opens before the creature takes off and closes after it lands.

This is absolutely expected and correct behaviour. Great, that you've handled this issue before I found it. :)

Talking about missing landing animation... It happens to be absolutely random.
I also noticed, that not slowed creatures are sometimes missing landing animation too.
Could it be a regression of our recent idle animation PR? I mean, landing animation could be interrupted by idle animation being initiated at the moment creature lands? I've checked version, which is ~ week old and creatures don't seem to have such issue...

Districh-ru:
@Branikolog 

> Talking about missing landing animation... It happens to be absolutely random. I also noticed, that not slowed creatures are sometimes missing landing animation too. Could it be a regression of our recent idle animation PR? I mean, landing animation could be interrupted by idle animation being initiated at the moment creature lands? I've checked version, which is ~ week old and creatures don't seem to have such issue...

An interesting observation, and according to your video, it is: the Sprite starts the idle animation instead of landing.
But what is more interesting - this PR does not contain the idle animation fix #6346 :))
I merged the 'master' to include #6346.
I'll analyze the behavior of idle animation.
@Branikolog, does this occur only when speed is set to 1?

Districh-ru:
I think I found the cause of this issue.
The IDLE animation starts only if creature is set to STATIC animation.
The landing animation is done by making a vector of two animations: FLY_LAND and STATIC. And the game sets the sate of the current animation according to the last animation in the vector. So I separated this two animations.
@Branikolog, could you please check this fix when you have time.

Districh-ru:
@Branikolog, its an issue with checking - if the animation is over. The function which checks if the animation is over analyzes if the current frame is the last frame. The Gargoyle landing has only 1 frame in animation sequence and it is the last frame. So this animation is always finished and it does not render its single frame.
I'll try to fix this, but this change will affect all animations.

Branikolog:
> @Branikolog, its an issue with checking - if the animation is over. The function which checks if the animation is over analyzes if the current frame is the last frame. The Gargoyle landing has only 1 frame in animation sequence and it is the last frame. So this animation is always finished and it does not render its single frame. I'll try to fix this, but this change will affect all animations.

That would be great, if you can fix it!
I've compared with old master branch (~week ago) and all landing animations seem to work correct for not slowed creatures there.
If you fix last frame rendering, everything else seem to be done.

Districh-ru:
@Branikolog, I've fixed the last frame issue and also noticed, that the animation delay is different for take off and landing for different creatures, so I fixed it too.
After the PR #6346 the game randomly switches the STATIC animation to IDLE, so I added the STAND_STILL animation, which will not be switched to IDLE for case when the creature waits for bridge and for after landing 1 frame standing. And there will be no more units counter rendered wen creature waits for bridge action.

Branikolog (approved):
@Districh-ru 
Now everything seems to work fine! Huge thanks for your patience! :)

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3394,22 +3405,85 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+                if ( bridge->NeedDown( unit, *dst ) ) {
+                    // Restore the initial creature position before rendering the animation.
+                    unit.SetPosition( startPosition );
+                    bridge->ForceAction( true );
+                    break;
+                }
+
+                // Fix for wide flyers - go the whole path with reflections to check the bridge.
+                if ( isWide ) {
+                    if ( unit.GetTailIndex() == *dst ) {
```
ihhub, line 3433: We can combine these 2 ifs into one.

ihhub (changes requested):
Hi @Districh-ru , I left 2 very minor comments here. Would you mind to take a look at them?

ihhub (approved):

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3394,48 +3406,131 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+            }
             AudioManager::PlaySound( unit.M82Move() );
             unit.SwitchAnimation( Monster_Info::MOVING );
             AnimateUnitWithDelay( unit, frameDelay );
             unit.SetPosition( *dst );
         }
 
-        // check for possible bridge close action, after unit's end of movement
-        if ( bridge && bridge->AllowUp() ) {
+        // Check for possible bridge close action, after land unit's end of movement.
```
oleg-derevenetz, line 3506: It would be good to add an explanation here **why** an exception has been made for flying units. Most people read code from top to bottom. In the comments, it is advisable to explain not so much **what** is happening here (in many cases it is obvious from the code), but **why**.

src/fheroes2/monster/monster_anim.h
```diff
@@ -41,6 +41,7 @@ namespace Monster_Info
     enum AnimationType : int
     {
         NONE,
+        STAND_STILL,
```
oleg-derevenetz, line 44: It would be good to add the comment here to describe the difference between `STAND_STILL` and `STATIC`, because it is not obvious for the people that see this enumeration for the first time.

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left a few suggestions, mostly concerning the style and additional comments.

src/fheroes2/battle/battle_bridge.cpp
```diff
@@ -118,11 +118,11 @@ void Battle::Bridge::SetPassable( const Unit & b ) const
...
 void Battle::Bridge::Action( const Unit & b, int32_t dst )
 {
-    bool action_down = false;
-
-    if ( NeedDown( b, dst ) )
-        action_down = true;
+    ForceAction( NeedDown( b, dst ) );
+}
 
+void Battle::Bridge::ForceAction( const bool action_down )
```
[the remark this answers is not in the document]
Districh-ru: Hi, @oleg-derevenetz, thanks for the detailed comments, I've made the suggested changes.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3394,48 +3406,135 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+            }
+
+            // Restore the initial path pointer state.
+            dst = path.begin();
+        }
+
+        // The destination of fly off is same as the current creature position.
+        _movingPos = unit.GetRectPosition().getPosition();
+        const bool isFromRightArmy = unit.isReflect();
+        const bool isFlyToRight = ( _movingPos.x < Board::GetCell( *dst )->GetPos().x );
```
oleg-derevenetz, line 3455: BTW what if the `path` is empty? In this case `dst` here refers to the `path.end()` (which is the same as `path.begin()` in this case), so we will dereference the "past-the-end" iterator, which is undefined behavior. In other places dereference and increment of `dst` is protected by `while ( dst != pathEnd )` check, but not here. Normally this function should not be called with an empty `path`, but I still believe that we need some sort of protection here. The same consideration applies to the `++dst` below, when we skip one path position while turning wide flying unit back.

oleg-derevenetz: P.S. We can just return if `path` is empty at the very beginning of this function and also add an assertion somewhere here to check that `dst` is not equal to the `pathEnd`.

Districh-ru: If the path is empty then we don't need to execute the whole method as there is no movement :)
I applied the suggested change. It should keep the method from unexpected situations with path.

Districh-ru: @oleg-derevenetz , if it is better to add an assertion `assert( dst != pathEnd );` instead of checking `dst == pathEnd` in 'if' part, please tell, I'll change the code.

oleg-derevenetz: @Districh-ru Honestly looking at this part I believe both assertion and check are redundant here, since we are checking for `empty()` at the very beginning of this function. When we are here, `dst` is always equal to `path.begin()` which can never be equal to `path.end()` because of that check for `empty()`, so we can just leave that check at the very beginning and revert the `( dst != pathEnd )` check below. It looks like that will be enough :)

oleg-derevenetz (approved):
LGTM.

ihhub:
@Districh-ru , big thanks for fixing these issues!

PR #6356, opened 2022-12-12
Do not mirror corpses of defeated troops (Battlefield)

### LINKED ISSUE #6321: Combat screen, corpses of defeated troops are mirroring, keep the initial orientation
### Preliminary checks

### Platform

Windows

### Describe the bug

After defeating battle dwarf corpse turns to keep the initial orientation of enemy troops.

https://user-images.githubusercontent.com/55348946/206187522-49d9a326-53e2-4711-ad1a-824057675c11.mp4

In the original game sprite always keeps the orientation it was before suffering the final blow.

### Save file

None.

### Additional info

_No response_

Districh-ru:
close #6321

https://user-images.githubusercontent.com/113276641/207124866-dcd3954e-e4b7-43a1-8821-b2c97c878272.mp4

https://user-images.githubusercontent.com/113276641/207167267-f30f739b-f16f-4827-9a02-287b73d43d54.mp4

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -382,11 +382,7 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
             }
 
             attacker->UpdateDirection();
-
-            // Restore direction only for the living defender and don't flip the corpse of a dead defender horizontally.
-            if ( !defender->GetDead() ) {
-                defender->UpdateDirection();
-            }
+            defender->UpdateDirection();
         }
         else {
             DEBUG_LOG( DBG_BATTLE, DBG_WARN, "incorrect param: " << attacker->String( true ) << " and " << defender->String( true ) )
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -379,10 +379,15 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
                     DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "double attack" )
                     BattleProcess( *attacker, *defender, dst, dir );
                 }
+
+                // Reflect defender only if he is alive.
+                defender->UpdateDirection();
             }
 
-            attacker->UpdateDirection();
-            defender->UpdateDirection();
+            // Reflect attacker only if he is alive.
+            if ( attacker->isValid() ) {
+                attacker->UpdateDirection();
+            }
         }
         else {
             DEBUG_LOG( DBG_BATTLE, DBG_WARN, "incorrect param: " << attacker->String( true ) << " and " << defender->String( true ) )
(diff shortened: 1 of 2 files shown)


src/fheroes2/battle/battle_action.cpp
```diff
@@ -382,7 +382,11 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
             }
 
             attacker->UpdateDirection();
-            defender->UpdateDirection();
+
+            // Restore direction only for the living defender and don't flip the corpse of a dead defender horizontally.
+            if ( !defender->GetDead() ) {
```
oleg-derevenetz, line 387: This is wrong, `GetDead()` returns the number of dead creatures in the stack (for the purpose of such spells as Animate Dead or Resurrect), so with current logic you will not restore direction for any stack that has any dead creatures, even if there are still alive creatures in the same stack. What you apparently really need here is the `isValid()`.

Also please note that `attacker` may also be dead at this point due to the `defender`'s retaliation strike. Please test your logic carefully in all possible situations, and not in just one that seems "the most suitable", this is especially important before the upcoming release.

Districh-ru: Hello, @oleg-derevenetz.
Thank you for pointing out this big error in my proposed logic!
I'll rewrite the introduced code and do more checks.

src/fheroes2/battle/battle_action.cpp
```diff
@@ -379,10 +379,15 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
                     DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "double attack" )
                     BattleProcess( *attacker, *defender, dst, dir );
                 }
+
+                // Reflect defender only if he is alive.
+                defender->UpdateDirection();
```
oleg-derevenetz, line 384: The check for `defender->isValid()` is not sufficient because `attacker` (for example, Crusader) may apply a double attack since then, and while `defender` was valid before the double attack, it may not be valid anymore here, after the double attack was performed. It's best to leave the previous `UpdateDirection()` calls, but wrap them into `if ( ...->isValid() )` checks.

src/fheroes2/battle/battle_action.cpp
Districh-ru: I'm sorry for opening this PR without testing the possible situations for proposed logic. )
I have updated the code. After the battle between the two creatures each of them will be reflected back only if it is alive. Also I added the 'UpdateDirection()' call at the beginning of the resurrection animation to restore the direction of dead creatures before resurrection.

src/fheroes2/battle/battle_action.cpp
Districh-ru: @oleg-derevenetz, many thanks for finding my mistakes and for helping me to understand the subtleties of battle logic!
I made the proposed change.

oleg-derevenetz (approved):

ihhub (approved):

Branikolog (approved):

ihhub:
@Districh-ru , huge thanks for fixing this issue!

PR #6361, opened 2022-12-14
Bulgarian Translation - Last Buttons translations

Effektus:
Last Buttons translations


Effektus:
It is better to delete old one?

Issue #3350, opened 2021-05-05
Recruit panel: Click on the button, drag outside the button = both texts appear


1GodRage:
When I push (left click) the button and release the mouse outside the button, it looks bad. :p
It seems that only the "Recruit panel" is concerned by this behaviour.

On this screenshot, I did it to all 3 buttons:
![image](https://user-images.githubusercontent.com/63362446/117180976-0ea0f180-add5-11eb-8c26-e1e50cdc65bb.png)![image](https://user-images.githubusercontent.com/63362446/117180577-9afee480-add4-11eb-97e9-df0b5e2dd895.png)

_(Or maybe it's like because I used my French PoL files?.. 🤔)_

Branikolog:
@1GodRage Hi! :)

> _(Or maybe it's like because I used my French PoL files?.. thinking)_

Seems you're right. I cannot reproduce it on my english version.
Texts in some buttons can be generated in special way... So, here you've found a case.
Could you, please, pay extended attention for other buttons while gameplay and add here, if you find something?

ihhub:
Hi @1GodRage , it looks like the problem of the resources themselves. Could you please the same UI in the original version?

ihhub:
For example, Exit button in the hero meeting dialog:
![image](https://user-images.githubusercontent.com/19829520/117561598-c0d50380-b0ca-11eb-9da0-36cf21584a3e.png)

ihhub:
The same happens for German GOG version. Looks like translators were lazy workers!

1GodRage:
> Could you, please, pay extended attention for other buttons while gameplay and add here, if you find something?

The "FIN" button in hero_meeting_screen (that you already noticed)
End battle screen : Same goes for the OK button.
![image](https://user-images.githubusercontent.com/63362446/117664208-cfdab500-b1a1-11eb-83a0-7435bdc89696.png)
Highscores: exit button. (maybe Campaig/scenario button also, but I didn't finished any game yet so I can't check it now. xD)
![image](https://user-images.githubusercontent.com/63362446/117664524-28aa4d80-b1a2-11eb-8b59-f607420b99bf.png)


But I found that in battle, the "Next" button remain pushed if I drag_outside it.
![image](https://user-images.githubusercontent.com/63362446/117663507-0401a600-b1a1-11eb-9153-462aa9c0a14d.png)
Same with maps size buttons when creating a new game:
[...]

PR #6368, opened 2022-12-16
Update the French translation

zenseii:
Translate all buttons and new strings.

close https://github.com/ihhub/fheroes2/issues/3350


ihhub:

PR #6372, opened 2022-12-17
Fix bridge logic after morale event, simplify Bridge class, minor code cleanup

oleg-derevenetz:
Before this PR:

https://user-images.githubusercontent.com/32623900/208215297-a937516d-89cb-41c5-8564-9367dc27cce2.mp4

Now:

https://user-images.githubusercontent.com/32623900/208215374-7a51bcb7-23df-49be-b8fb-c547999a9134.mp4

Plus minor cleanup and simplification of overcomplicated castle bridge logic.

### DIFF
--- a/src/fheroes2/battle/battle_bridge.cpp
+++ b/src/fheroes2/battle/battle_bridge.cpp
@@ -60,7 +60,7 @@ bool Battle::Bridge::AllowUp() const
     return isValid() && isDown() && !isOccupied();
 }
 
-bool Battle::Bridge::isOccupied() const
+bool Battle::Bridge::isOccupied()
 {
     const Battle::Graveyard * graveyard = Arena::GetGraveyard();
 
--- a/src/fheroes2/battle/battle_bridge.cpp
+++ b/src/fheroes2/battle/battle_bridge.cpp
@@ -21,9 +21,11 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-#include "battle_bridge.h"
+#include <cassert>
+
 #include "battle_arena.h"
 #include "battle_board.h"
+#include "battle_bridge.h"
 #include "battle_cell.h"
 #include "battle_grave.h"
 #include "battle_interface.h"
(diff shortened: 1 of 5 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3533,24 +3538,22 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
     // Slowed flying creature has to land.
     if ( canFly ) {
         // IMPORTANT: do not combine into vector animations with the STATIC at the end: the game could randomly switch it to IDLE this way.
-        std::vector<int> landAnim;
-        landAnim.push_back( Monster_Info::FLY_LAND );
-        landAnim.push_back( Monster_Info::STAND_STILL );
-        unit.SwitchAnimation( landAnim );
+        unit.SwitchAnimation( { Monster_Info::FLY_LAND, Monster_Info::STAND_STILL } );
```
oleg-derevenetz, line 3541: Hi @Districh-ru I simplified your code a bit here and in another similar place.

oleg-derevenetz:
Hi @ihhub there is no hurry with this PR - this bug is minor, so you can leave this PR for 1.01 if you wish.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi, @oleg-derevenetz, thanks, now this code is easier to read and easier to link `SwitchAnimation` with the comment! And also no 'push_back' calls.

src/fheroes2/battle/battle_bridge.cpp
```diff
@@ -21,110 +21,119 @@
...
-        const int32_t prevPos = b.GetHeadIndex();
-
-        if ( dstPos == GATES_CELL && ( prevPos == CELL_AFTER_GATES || prevPos == BELOW_BRIDGE_CELL || prevPos == ABOVE_BRIDGE_CELL ) ) {
-            return true;
-        }
-        else if ( dstPos == MOAT_CELL && prevPos != GATES_CELL ) {
-            return true;
-        }
+
+    const int32_t headIdx = unit.GetHeadIndex();
```
Districh-ru, line 83: I also read the code changes, now the bridge logic is more clear. :)

But here `unit.GetHeadIndex()` is used to check the position from where the walking unit will walk onto the bridge. So maybe it would be better to call variable as `curIdx` as we have `dstIdx` as the destination: 
```cpp
const int32_t curIdx = unit.GetHeadIndex();
```

oleg-derevenetz: I would agree with this if the unit somehow moved right in this function, and there would be some sort of `curHeadIdx` and `thenHeadIdx`, in other words, if the result of `GetHeadIndex()` changed during the call of this function, but this is not really the case. There is no value for it other than "current" inside this function, so adding a prefix just lengthens the variable name without adding any value IMHO.

Districh-ru: From this point of view, it is true. I looked from the point of the caller of this function.
It is OK to keep `headIdx` as it is always 'current' inside the function.

src/fheroes2/battle/battle_bridge.cpp
```diff
@@ -21,110 +21,119 @@
...
 #include "castle.h"
 
 Battle::Bridge::Bridge()
-    : destroy( false )
-    , down( false )
+    : _isDestroyed( false )
+    , _isDown( false )
 {}
 
 bool Battle::Bridge::isValid() const
```
ihhub, line 40: Since we modify this code I suggest to make this method **inlined**. We can also make `isValid()`, `isDown()` and `AllowUp()` **inlined** as well.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3416,8 +3417,12 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
     StringReplace( msg, "%{monster}", Translation::StringLower( unit.GetName() ) );
     StringReplace( msg, "%{src}", std::to_string( ( unit.GetHeadIndex() / ARENAW ) + 1 ) + ", " + std::to_string( ( unit.GetHeadIndex() % ARENAW ) + 1 ) );
 
+    // TODO: remove this temporary assertion
```
ihhub, line 3420: Do we still need this comment and assertion?

src/fheroes2/battle/battle_bridge.h
```diff
@@ -38,24 +38,25 @@ namespace Battle
 
         Bridge & operator=( const Bridge & ) = delete;
 
-        void Action( const Unit &, int32_t );
```
ihhub, line 41: Just a side correction: shall we make this class `final`? :)

ihhub (changes requested):
Hi @oleg-derevenetz , I put few comments here. Would you mind to take a look at them?

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: I converted some temporary assertions to the permanent ones and added a few more temporary assertions just to check that previous assignments really were redundant :)

src/fheroes2/battle/battle_bridge.h
oleg-derevenetz: I made it `final`, but I should note though that there is not too much sense in making `final` classes without virtual methods. `final` is mainly useful to regulate the use of virtual methods.

ihhub: Oops... completely forgot about virtual.

ihhub (approved):

src/fheroes2/battle/battle_bridge.h
oleg-derevenetz: At least, there will be no harm anyway :) No one is going to inherit from `Bridge` anyway.

Districh-ru (approved):

Branikolog (approved):

ihhub:

PR #6374, opened 2022-12-17
Fix extra frame rendering for status update during AI turn

ihhub:
This is a leftover code and regression from #6319 which was forgotten to be addressed:
- do not update progress status for AI if it isn't changed. Every movement for hero was updating the status bar
- remove one extra frame rendering upon calling `DrawAITurnProgress()` method
- code cleanup to avoid weird reference increment code

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1796,6 +1796,14 @@ namespace AI
                         }
                     }
 
+                    if ( Game::validateAnimationDelay( Game::MAPS_DELAY ) ) {
+                        // Update Adventure Map objects' animation.
+                        uint32_t & frame = Game::MapsAnimationFrame();
+                        ++frame;
+
+                        gameArea.SetRedraw();
+                    }
+
                     basicInterface.Redraw( Interface::REDRAW_GAMEAREA );
 
                     // If this assertion blows up it means that we are holding a RedrawLocker lock for rendering which should not happen.
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1701,6 +1701,8 @@ namespace AI
 
             const std::vector<Game::DelayType> delayTypes = { Game::CURRENT_AI_DELAY, Game::MAPS_DELAY };
 
+            fheroes2::Display & display = fheroes2::Display::instance();
+
             LocalEvent & le = LocalEvent::Get();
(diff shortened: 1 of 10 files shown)


src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1780,19 +1795,19 @@ namespace AI
...
+                        // Update Adventure Map objects' animation.
+                        Game::updateAdventureMapAnimationIndex();
 
-                if ( Game::validateAnimationDelay( Game::MAPS_DELAY ) ) {
-                    // Update Adventure Map objects' animation.
-                    uint32_t & frame = Game::MapsAnimationFrame();
-                    ++frame;
-                    gameArea.SetRedraw();
-                }
+                        gameArea.SetRedraw();
```
oleg-derevenetz, line 1802: This call is redundant IMHO because we are forcibly redrawing the game area three lines below.

Branikolog (approved):

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1890,8 +1892,10 @@ namespace AI
             if ( maxHeroCount > 0 ) {
                 // At least one hero still exist in the kingdom.
                 const size_t progressValue = ( endProgressValue - startProgressValue ) * ( maxHeroCount - availableHeroes.size() ) / maxHeroCount + startProgressValue;
-
-                status.DrawAITurnProgress( static_cast<uint32_t>( progressValue ) );
+                if ( currentProgressValue < progressValue ) {
```
oleg-derevenetz, line 1895: So you don't want to allow the progress go backwards if additional heroes appear during the heroes' turn (for example, from jail)? That was funny :) But I don't mind in general.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few comments, could you please take a look at them?

src/fheroes2/ai/ai_hero_action.cpp
ihhub: Yes, you're right. It's an old code which I forgot to delete.

src/fheroes2/ai/normal/ai_normal_hero.cpp
ihhub: I can just make `if ( currentProgressValue != progressValue ) {` condition for this if you insist :)

oleg-derevenetz: This is not necessary :)

oleg-derevenetz (approved):

PR #6375, opened 2022-12-17
Optimize rendering in multiple places

ihhub:
- in Mage Guild screen we were rendering one frame for no reason
- after closing a popup window and restoring a part of the screen we should render only that part, not the whole screen
- small optimizations to avoid non-inlined method calls
- after an update Status Bar should render only the part which was updated
- visiting a Hut of Magi we were not displaying certain frames


ihhub:
Hi @oleg-derevenetz and @Branikolog , if you no issues we can merge it even for 1.0 release.

src/fheroes2/dialog/dialog_box.cpp
```diff
@@ -173,9 +173,10 @@ void Dialog::NonFixedFrameBox::redraw()
 
 Dialog::NonFixedFrameBox::~NonFixedFrameBox()
 {
+    const fheroes2::Rect restoredArea{ _restorer->x(), _restorer->y(), _restorer->width(), _restorer->height() };
     _restorer->restore();
 
-    fheroes2::Display::instance().render();
+    fheroes2::Display::instance().render( restoredArea );
```
oleg-derevenetz, line 179: JFYI: after #6382 you will be able to use just `_restorer->rect()` here.

oleg-derevenetz (approved):

Branikolog (approved):

PR #6384, opened 2022-12-17
Fix wrong usage of std::isspace() in TinyConfig

oleg-derevenetz:
Fix the undefined behavior.


ihhub:
@oleg-derevenetz , thanks for fixing this issue!

PR #6394, opened 2022-12-18
Add French and German to use original assets for buttons

zenseii:
Italian is kept outside because the button font is bad and we don't have any players showing interest in this translation anyways.


ihhub:
@zenseii , thank you very much for this change!

Issue #5537, opened 2022-06-13
Adventure map, game crashes on checking info of "hidden" mine.


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Sometimes original maps contain mines, that are not rendered on a world map. This editor flaw is quite known.
While checking the info of such captured map application crashes.
![image](https://user-images.githubusercontent.com/55348946/173439314-cb6b0629-d578-4ed5-8857-840716261741.png)
![image](https://user-images.githubusercontent.com/55348946/173439422-406c0434-5f4e-4547-92ea-bfade0c05c33.png)

### Save file

[Artifact_Isle_bug.zip](https://github.com/ihhub/fheroes2/files/8894255/Artifact_Isle_bug.zip)

oleg-derevenetz:
Can't reproduce the crash so far on the latest snapshot - neither on macOS nor on Windows:

https://user-images.githubusercontent.com/32623900/173588886-195c63fe-2738-4f8f-94d1-43e9cb77ebd5.mp4

Maybe something else needs to be done?

oleg-derevenetz:
Hi @Branikolog could you please attach the map file itself, because it seems that my map pack doesn't include this map? Or maybe some classic map also has this bug?

Branikolog:
@oleg-derevenetz 
Sure:
[Artifact Isle.zip](https://github.com/ihhub/fheroes2/files/8901323/Artifact.Isle.zip)

Districh-ru:
A couple of months ago I found a bug in "Voyage Home" expansion company, map 2. It has an unknown mine. But OG also has the same bug and it does not lead to a game crush, so I forgot about it :)

![изображение](https://user-images.githubusercontent.com/113276641/208304218-ab548fd9-aeea-4558-b44e-798586348666.png)

Save file: [vh2_bad_mine.zip](https://github.com/ihhub/fheroes2/files/10254008/vh2_bad_mine.zip)

If I check the information about this mine, the game does not crush:
![изображение](https://user-images.githubusercontent.com/113276641/208304866-0289dc8b-a554-46ab-b3a9-fdd95fa5ef8f.png)

[...]

PR #6395, opened 2022-12-18
Fix incorrect resource placements on certain maps

### LINKED ISSUE #6393: Assertion on resource collecting
### Preliminary checks

### Platform

Windows

### Describe the bug

image
image

Original campaign, 4 Roland scenario has incorrectly placed resource, which causes assertion on collecting.

### Save file

[X_CRASH_ON_TAKE_RES.zip](https://github.com/ihhub/fheroes2/files/10253728/X_CRASH_ON_TAKE_RES.zip)

### Additional info

_No response_

ihhub:
close #6393

This change must be thoughtfully tested on many maps to make sure that we don't break anything. The test scope should be done on maps with "bad" resources as well as any normal resources.

### DIFF
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -23,6 +23,7 @@
 
 #include <cstdint>
 #include <list>
+#include <ostream>
 #include <utility>
 
 #include "army_troop.h"
@@ -598,7 +599,7 @@ void Maps::Tiles::QuantityUpdate( bool isFirstLoad )
             break;
         default:
             // Some maps have broken resources being put which ideally we need to correct. Let's make them 0 Wood.
-            ERROR_LOG( "Tile " << _index << " contains unknown resource type. Tileset " << objectTileset << ", object index " << objectIndex )
+            DEBUG_LOG( DBG_GAME, DBG_WARN, "Tile " << _index << " contains unknown resource type. Tileset " << objectTileset << ", object index " << objectIndex )
             resourceType = Resource::WOOD;
             count = 0;
             break;
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -22,6 +22,8 @@
  ***************************************************************************/
 
 #include <cstdint>
+#include <list>
+#include <utility>
 
 #include "army_troop.h"
 #include "artifact.h"
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -554,10 +554,32 @@ void Maps::Tiles::QuantityUpdate( bool isFirstLoad )
     }
 
     case MP2::OBJ_RESOURCE: {
-        const int res = Resource::FromIndexSprite( objectIndex );
+        int resourceType = Resource::UNKNOWN;
+        // TODO: add a function opposite to MP2::GetICNObject() to return tileset ID.
+        const int resourceTileSet = 46;
+
+        if ( ( objectTileset >> 2 ) == resourceTileSet ) {
+            // The resource is located at the top.
+            resourceType = Resource::FromIndexSprite( objectIndex );
+        }
+        else {
+            for ( TilesAddon & addon : addons_level1 ) {
+                if ( ( addon.object >> 2 ) == resourceTileSet ) {
+                    resourceType = Resource::FromIndexSprite( addon.index );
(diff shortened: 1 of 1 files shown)


Branikolog (approved):
A miracle is here! :)
Resource is weirdly looking, being rendered over mountains, If it's not possible, to render it under mountains, like it was in the OG, I cannot imagine any better solution here.

ihhub:
Hi @Branikolog , please test this issue on normal resources being generate on multiple maps as well as the underlying logic for placing them was modified.

ihhub:
@Branikolog , I also added an error log in case if incorrect resource type would appear on any map. Try to load dozens of maps to see if you get the error message.

Districh-ru:
Hi, @ihhub, @Branikolog, a couple of months ago I found a bug in "Voyage Home" expansion company, map 2. It has an unknown mine. But OG also has the same bug and it does not lead to a game crush, so I forgot about it :)

изображение

Save file: [vh2_bad_mine.zip](https://github.com/ihhub/fheroes2/files/10254008/vh2_bad_mine.zip)

src/fheroes2/maps/maps_tiles_quantity.cpp
```diff
@@ -554,23 +557,54 @@ void Maps::Tiles::QuantityUpdate( bool isFirstLoad )
...
+
+        if ( ( objectTileset >> 2 ) == resourceTileSet ) {
+            // The resource is located at the top.
+            resourceType = Resource::FromIndexSprite( objectIndex );
+        }
+        else {
+            for ( TilesAddon & addon : addons_level1 ) {
+                if ( ( addon.object >> 2 ) == resourceTileSet ) {
+                    resourceType = Resource::FromIndexSprite( addon.index );
+                    // If this happens we are in trouble. It looks like that map maker put the resource under an object which is impossible to do.
```
oleg-derevenetz, line 572: Hi @ihhub What if we just remove this invalid addon?

ihhub: Then we need to add logic to remove the data from other tiles to make sure that the shadow is removed. By the way, is the shadow present for this case? I'm off my PC for now.

oleg-derevenetz:
@Districh-ru see #5537 :)

src/fheroes2/maps/maps_tiles_quantity.cpp
ihhub: On top of this we will have to update object type and this is where we have no logic in the code. This should be added as a part of Editor phase.

Districh-ru:
> @Districh-ru see #5537 :)

Thanks, I'll add my post with the save file to this issue.

src/fheroes2/maps/maps_tiles_quantity.cpp
```diff
@@ -554,23 +557,54 @@ void Maps::Tiles::QuantityUpdate( bool isFirstLoad )
...
-        default:
+        case Resource::MERCURY:
+        case Resource::SULFUR:
+        case Resource::CRYSTAL:
+        case Resource::GEMS:
             count = Rand::Get( 3, 6 );
             break;
+        default:
+            // Some maps have broken resources being put which ideally we need to correct. Let's make them 0 Wood.
+            ERROR_LOG( "Tile " << _index << " contains unknown resource type. Tileset " << objectTileset << ", object index " << objectIndex )
```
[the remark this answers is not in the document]
ihhub: The error log would happen only if the new logic is wrong. This code affects only new maps. Old saves are untouched.

src/fheroes2/maps/maps_tiles_quantity.cpp
oleg-derevenetz: @ihhub there is a shadow on the nearby tile, but it seems that it's a mountain's shadow, not a resource shadow. Anyway, the OG doesn't remove these resources, so this solution is OK.

Districh-ru:
@ihhub, I also have checked all maps I have (it's a standard set from Buka CD) and did not find any unknown resource type.
Also I checked on what maps works your fix: maps 'peasants.mx2' and 'whoam.mp2'  has 1 resource under an object. And map 'lostreli.mp2' has 5.

oleg-derevenetz (approved):
LGTM. Could you please apply the IWYU suggestion? Also do you want to leave the `ERROR_LOG`? I used the `DEBUG_LOG` with `DBG_GAME` and `DBG_WARN` in similar cases.

ihhub:
> LGTM. Could you please apply the IWYU suggestion? Also do you want to leave the `ERROR_LOG`? I used the `DEBUG_LOG` with `DBG_GAME` and `DBG_WARN` in similar cases.

Oh, I missed `IWYU` suggestion again :( I changed logging to `DEBUG_LOG`.

PR #6398, opened 2022-12-19
Modification of walking units movement animation algorithm. Battlefield.

### LINKED ISSUE #1094: Barbarian castle well animation problems
I've noticed a bunch of problem with animations in barbarian well:
Goblins shift back a little after moving animation stops;
The same seemed to do Orc Chiefs.
Ogres have little shifting to left, right before walking animation starts.
This also happens, while walking animation stops.
Walking animation seem to be shifted for lots of creatures in well.

### LINKED ISSUE #1160: Swordsmen moving frames issue
Seems like walking swordsmen are missing one of the moving frames. Swordsmen sprite jumps a bit during movement... Especially noticeable at the start of moving. They lack smoothness during walking.

### LINKED ISSUE #1302: Golem movement speed issue
In the OG Iron and Steel Golems have their movement speed inconstant.
They start movement fast, but to the end of the movement they become slower.
Currently, their movement speed is persistent.

### LINKED ISSUE #1700: Golem sprite is shifted during diagonal movement
While diagonal moving golem's sprite shifts. We can notice that, when golems stop, they jump a bit, right after the movement had finished.

Districh-ru:
close #1700
close #1302
close #1160
close #1094 

The walking units movement loop now includes, besides main `MOVING` animation sequence:
- `MOVE_QUICK` - when creature moves for 1 cell,
- `MOVE_START` - movement from beginning till the first cell,
- `MOVE_END` - movement to the last cell in the path.

Fixed Golem movement animation.
Added ability to use frame 'x' (horizontal) offset from original BIN file in diagonal creature movement (should fix unit sprite shifting issues).

Golem movement: (now it has 4 frames of moving start)

https://user-images.githubusercontent.com/113276641/208481067-4c40839e-bd6d-4244-8fe7-68f58ede09fe.mp4

Slowed GIF with Golem and Skeleton (now it has 2 extra frames at start and 1 at the end) movement:

fheroes2 2022-12-19 19-45-29-933_

Slowed GIF with Swordsmen animation (now it has 2 extra frames at start):
fheroes2 2022-12-19 20-59-05-251

Added 02.01.2023: Also fixed walk animation in castle well (unit info dialog) and during the diagonal movement.


Districh-ru:
Hi, @Branikolog!
Could you, please, check the units animations with new animation implementation and also their interaction with the bridge.
This is a huge work, so, please, do it only in you spare time. :)

Branikolog:
Hi, @Districh-ru .

>The walking units movement loop now includes, besides main MOVING animation sequence:

That's great!

>Could you, please, check the units animations with new animation implementation and also their interaction with the bridge.
This is a huge work, so, please, do it only in your spare time. :)

Sure, I'll check all creatures but I'll do it a bit later. I'll be busy next few days.

Districh-ru:
> Sure, I'll check all creatures but I'll do it a bit later. I'll be busy next few days.

Don't rush, you have as much time as you need. :)
I want to optimize this code a bit and I'll also be busy until the weekend.

src/fheroes2/battle/battle_animation.cpp
```diff
@@ -92,6 +92,11 @@ int AnimationSequence::getFrame() const
     return isValid() ? _seq[_currentFrame] : 0;
 }
 
+size_t AnimationSequence::getFrameNumber() const
```
ihhub, line 95: Please make this method inlined: put inside the header.

ihhub: I also suggest to rename it to `getCurrentFrameId()` to be more specific.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1671,50 +1671,57 @@ void Battle::Interface::RedrawTroopSprite( const Unit & unit )
...
+                offset.x -= Sign( moveX ) * offsetX / 2;
             }
-            else if ( 0 != Sign( cy ) ) {
-                sp.x -= Sign( cx ) * ox / 2;
+
+            // Special movement animation fix for Iron and Steel Golem: its 'MOVING' and 'MOVE_END' animation is missing 1/4 of animation start.
+            else if ( ( _movingUnit->animation.getCurrentState() == Monster_Info::MOVING || _movingUnit->animation.getCurrentState() == Monster_Info::MOVE_END )
+                      && ( _movingUnit->GetID() == Monster::IRON_GOLEM || _movingUnit->GetID() == Monster::STEEL_GOLEM ) ) {
+                // Expand the 'x' sprite offset range from [0.25,1] to [0,1] to smoothly concatenate with other animations.
+                offset.x -= static_cast<int32_t>( ( 0.75 - movementProgress ) * moveX / 4 );
```
ihhub, line 1706: I really suggest to modify the code in `bin_info.cpp` file for these creatures rather than to have a hack like this. This should be avoided unless it's impossible to set animation info within `bin_info.cpp` file.

You can get all offsets from your current calculations and then apply changes in the file.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3480,55 +3488,98 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
-            }
-            AudioManager::PlaySound( unit.M82Move() );
+        // Render the unit movement with the movement sound.
+        AudioManager::PlaySound( unit.M82Move() );
+        AnimateUnitWithDelay( unit, frameDelay );
+        unit.SetPosition( *dst );
+
+        // Do a post-move check for the bridge action and set the animation the movement to the next cell in the path.
+        if ( canFly ) {
+            // The animation for the next step in the path of slow flying creatures is always "MOVING".
```
ihhub, line 3555: A small correction **slow** --> **slowed** as it was slowed down.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3480,55 +3488,98 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+        else {
+            // Check for possible bridge close action, after walking unit's end of movement to the next cell.
+            // This check should exclude the flying creature because it can't 'hang' here to wait
+            // for bridge to close. For this creature, the bridge should close after it lands.
+            if ( bridge && bridge->AllowUp() ) {
+                _movingUnit = nullptr;
+                unit.SwitchAnimation( Monster_Info::STAND_STILL );
+                bridge->ActionUp();
+                _movingUnit = &unit;
+                if ( !isOneStepPath && dst != ( pathEnd - 2 ) ) {
```
ihhub, line 3567: Are we 100% certain that `path` contains more than 2 cells? If not then this check is dangerous. The same for line 3576.

ihhub (changes requested):
Hi @Districh-ru , I left here several comments. Would you mind to check them in your spare time?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: We are certain that the path contains 2 or mode cells (`!isOneStepPath`).
If it contains two cells, then we can have 3 vector iterators: "first cell", "second cell" and the last "past-the-end" iterator (`endPath`).
So if we have only 2 cells, then `( path - 2 )` will point to the first `path` vector element.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi, @ihhub!
I've tried to implement this, but found out that the "offset" data from BIN file is already corrected. And this data contains only the information how to horizontally center the creature sprite in unit info dialog and I also use it to correct horizontal movement range when moving diagonally. But it centers the body of the creature and is not connected to its legs position. Using this data in horizontal movement will result in "sliding" of creature's feet and will result in worse animation for most creatures.
The creature position for its horizontal movement is set in its Sprite coordinate 'x' for the corresponding frame. So if we want to correct this coordinate not in the `battle_interface.cpp` file, we have to modify `agg_image.cpp`...
In next days I'll try to implement this fix in this file.

ihhub: As discussed in Discord please expand the comment explaining why we do this in the code so it would be obvious for any developer to understand this edge case.

Districh-ru: I think I found a way make a fix in `agg_image.cpp` and `bin_info.cpp`, I'm working on it...

Districh-ru: Hi, @ihhub. I've made some changes in 'agg_image.cpp' and 'bin_info.cpp' to fix Golem animation issues as well as changes for some other creatures in 'bin_info.cpp'. Could you please check the implementation of this fix when you have time. :)

Branikolog:
Hi, @Districh-ru 
I've started making tests. I've noticed, that slowed fling creatures have issue:

https://user-images.githubusercontent.com/55348946/210411870-74a177e0-c019-4802-bdc3-967ec67efb09.mp4

Districh-ru:
> Hi, @Districh-ru I've started making tests. I've noticed, that slowed fling creatures have issue:

Hi, @Branikolog! Thanks.
I'll fix this.

Branikolog:
Hi, @Districh-ru 
Thanks for flyers fix!
I hope, I've checked every creature in game.
I found, that master swordsmen animation is a bit twitchy, but the original game has the same issue. Is it simply the way how original artists made this animation? I suppose, we cannot fix it until we add some new frames, am I right?
Also I've noticed that goblin animation in well still shifts a bit, while walking animation stops. Is it possible to be fixed? I've noticed, you've managed to fix orcs and ogres animation same issue.
One more thing in this well:
War trolls while moving have one frame shifted. I haven't noticed this issue during battle, but here it's quite obvious. I suppose it's out of the current PR, but I just ask you to be sure, I'm not paranoic. :)

Districh-ru:
> I found, that master swordsmen animation is a bit twitchy, but the original game has the same issue. Is it simply the way how original artists made this animation? I suppose, we cannot fix it until we add some new frames, am I right?

Yes, it is how it was made for OG. I think they missed 1 frame. Fixing this like I did to Golem will bring a more ridiculous walk to Swordsmen. :) We have to draw this frame in future.

> Also I've noticed that goblin animation in well still shifts a bit, while walking animation stops. Is it possible to be fixed? I've noticed, you've managed to fix orcs and ogres animation same issue.

I've applied the shift fix for 3 pixels for: Orc, Ogre, Troll, Dwarf, Unicorn, Centaur, Boar, Rogue and for Goblin. 
But, as we see it is not enough for Goblin. Not its sprite is shifted by 6 pixels.

> War trolls while moving have one frame shifted. I haven't noticed this issue during battle, but here it's quite obvious. I suppose it's out of the current PR, but I just ask you to be sure, I'm not paranoic. :)

Fixed this too. :)

@Branikolog, thank you for the thorough check!

Branikolog:
Hi, @Districh-ru !
Nice! Now goblins and trolls move smoothly!
Could you, please, look at archers/rangers and Liches/Power Liches? I think they have same issue, shifting a bit, after moving animation stops.
Also, I've noticed, That Titans/Giants walking animation has one frame (supposedly starting frame, which I believe is included in walking animation too) shifted. The issue looks similar to troll issue above.
Could you, please, look at these issues? I believe everything else is fine now.

Districh-ru:
Hi, @Branikolog, I've fixed these creatures, please check it when you have time.

Branikolog:
Hi, @Districh-ru !
Good work! Sorry, but I have to ask your help one more time. The more I test, the more issues I notice. :)
It looks like Crusaders/Paladins animation is not OK in well too. The issue is similar to Titans. I think several frames have shifting during animation.
Also, talking about Crusaders. I've noticed they finish walking animation before reaching the target cell and make short jump in the end. Could you, please, look, if it is possible to be fixed?

https://user-images.githubusercontent.com/55348946/210771545-360ee931-99fe-4229-959a-9f27fa149327.mp4

Branikolog:
@Districh-ru 
The same issue with Cyclops.

https://user-images.githubusercontent.com/55348946/210782574-9d13fd9a-beb3-424e-a9e7-8792df576f97.mp4

Seems the last walking frame is missing right before standing frame.

Districh-ru:
> @Districh-ru The same issue with Cyclops.
> Seems the last walking frame is missing right before standing frame.

@Branikolog, I could not reproduce this bug with either Crusader or Cyclop...

https://user-images.githubusercontent.com/113276641/210784024-b87f01c2-25bb-41d5-a245-5f21bc0045ea.mp4

Branikolog:
Hi, @Districh-ru !
On your video everything works correct.
I've checked the build from current PR and the issue with last walking frame missing still occurs. I haven't checked every creature yet, but it seems, there're some other creatures with the same issue. Like elementals, for instance.

Districh-ru:
> Hi, @Districh-ru ! On your video everything works correct. I've checked the build from current PR and the issue with last walking frame missing still occurs. I haven't checked every creature yet, but it seems, there're some other creatures with the same issue. Like elementals, for instance.

@Branikolog, I've changed the logic of rendering of the movement animation to avoid possible low display time of the last frame in animation.
This is a temporary fix as it differs from the render logic in all other places. If it fixes these issues, I'll try to fix the use of this delay: now the movement animation (and shooting animation) delay is "CUSTOM" delay and it is not used as all other delays, including that there is no ability to check this delay in other functions than the function for movement animation for one cell.
The main problem is that `HumanTurn` function starts the player's turn and inside this function there is no mechanism to wait for this "CUSTOM" delay, as there is no ability to wait for this delay at the end of movement rendering function. So if the next creature is controlled by player its last movement frame could be shown for a very small time and then `HumanTurn` function immediately displays this creature standing and renders a contour around active creature.

Districh-ru:
I've added new delay type for unit movement animation and added the check at the start of player's turn for pass of any possible delay after previous battlefield action to reduce loss of last frames of these actions animations.
@Branikolog, when you have time could you please check the animation of walking and flying creatures and shooting animation of the creature (not the projectile, I'll try to fix projectile position soon in other PR, this PR is big enough).

Branikolog:
Hi, @Districh-ru !
I've started making some tests...
Now the engine doesn't skip the last frame while moving. But when creatures has high morale triggered, there's still an issue:

https://user-images.githubusercontent.com/55348946/210957039-9aa223dc-8162-4722-80b4-54528fe6ded3.mp4

Districh-ru:
> Hi, @Districh-ru ! I've started making some tests... Now the engine doesn't skip the last frame while moving. But when creatures has high morale triggered, there's still an issue:

Thanks, @Branikolog. The function for spell effects also, like `HumanTurn` function does not wait for previously set delays. :(
I'll fix this soon.
If you find some more frame synchronization issues, please write.

Branikolog:
@Districh-ru 
Elves missing animation after an arrow was launched:

https://user-images.githubusercontent.com/55348946/210969354-a5dd4c3d-0731-4863-bdd8-2b9b6c2bb853.mp4

OG:
https://user-images.githubusercontent.com/55348946/210969496-5c2fd77e-2683-4b14-8b42-658b96f64811.mp4

Districh-ru:
> @Districh-ru Elves missing animation after an arrow was launched:

And maybe not only Elves :)
Also in the Elves/Archers attack animation the arrow sometimes disappears for 1 frame.  I'll try to fix it too.

Districh-ru:
> And not only Elves. All distant attackers does not have this animation. :( 

And not only archers, for melee attacks there is also the ending animation, which is not currently used.
So this is a new issue that, I think, should be done in https://github.com/ihhub/fheroes2/pull/6414 PR or in a new PR, but after merging #6414.

> Also in the Elves/Archers attack animation the arrow sometimes disappears for 1 frame. I'll try to fix it too.

@Branikolog, I've fixed this issue and, I hope, the missing last frame in animations, please check it when you have time.

Branikolog (approved):

Districh-ru:
> Hi, @Districh-ru ! Everything else seem to be ok now. I have noticed several issues with walking animations for creature info window (elementals and medusas). Also Ghosts starting movement seem to make unnecessary idle... But I think it's out of current PR.

I can try to fix Medusa and Elemental shifts.
And about Ghosts - they have no "fly up" animation, so the game cannot change it and uses the previously set animation. :))
I'll fix it.

> @Branikolog, the absence of after attack animation for distant attacks I'll fix in other PR (#6414) is this fix has to be done in functions, that are modified in that PR. As this after attack animation should be rendered during wince/death of target.

This animation does not start as it has only 1 frame and the game logic, while checking if the animation is finished, considers that this animation is already finished :)

Branikolog:
@Districh-ru 

> I can try to fix Medusa and Elemental shifts.

Then also look at nomad walking animation in nomad info window. :)

> This animation does not start as it has only 1 frame and the game logic, while checking if the animation is finished, considers that this animation is already finished :)

We shouldn't miss that one frame. :)
Btw, Troll after attack animation seems to work properly.

Districh-ru:
> @Districh-ru
> Then also look at nomad walking animation in nomad info window. :)

OK. :)

> We shouldn't miss that one frame. :) Btw, Troll after attack animation seems to work properly.

Troll has 2 animation frames.

Districh-ru:
@Branikolog, I've fixed ghosts fly off issue and well animations for Nomad, Medusa and Elementals. Could you check it once more when you have time.

Branikolog:
@Districh-ru 

> @Branikolog, I've fixed ghosts fly off issue and well animations for Nomad, Medusa and Elementals. Could you check it once more when you have time.

Well done!

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3496,63 +3537,107 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
         }
+
+        // Switch animation to MOVING before going through the path.
+        unit.SwitchAnimation( Monster_Info::MOVING );
+    }
+    else {
+        // Every ground unit should start its movement from the special 'MOVE_START' animation
+        // or if it moves only for one cell its animation must be 'MOVE_QUICK'. So a check for 1 cell path is made.
+        // If a wide unit moves backwards for 1 cell, it turns twice, so it has 3 path points. And its first 'dst' point is equal to the last.
+        // TODO: try to rewrite path generation and movement of wide creatures to get more clear and unique path with non-wide (to get rid of of 'wide creature patches').
```
oleg-derevenetz, line 3557: Please note that all these turns back & forth specific for wide units are important not so much for visualization (it's actually secondary), but for the correct marking of cells that are reachable for the head and tail of wide units, which is vital for the pathfinding (see the description of #4018 for details). If you just throw out these turns, the cells will be marked incorrectly, and similar bugs will return.

Districh-ru: Hi @oleg-derevenetz, thanks for the advice!
I read the PR you linked to.
I understand that these turns were done for a reason. And I won't make changes until I'm sure that these changes won't ruin the game logic.
But now pathfinding is not always perfect, for example moving a wide creature for 1 cell diagonally and back makes it to do 2 steps:

https://user-images.githubusercontent.com/113276641/211160337-9ef2f276-0ff1-456c-849d-b982ab22bf81.mp4

I am interested in finding a better solution, but not right now. Maybe after a month. :(
So I left this TODO. I can remove it.

oleg-derevenetz: Currently for the battle pathfinding the recursive greedy (in a sense that it prefers to consider the cells with the minimum distance to the destination) depth-limited algorithm is used. While it does not always offer the shortest path, it always finds the path if it exists. There were attempts from the previous team to adapt the Dijkstra-based algorithms (including A-star) but these implementations have numerous issues with wide units so the game was literally non-playable during siege for example. With recursive approach we at least have a playable game (but maybe with some cosmetic issues) :) But if you have some ideas, you are welcome to try them, of course. I just wanted to point out that this area is quite sensitive :)

Regarding this video, both candidate cells (to the left and to the upper right) have the same distance to the destination cell (namely 1) so the pathfinder is free to choose any of them. It chooses upper right one, and it has enough movement points left to reach the destination, so it does not try other paths. However I agree that this path is not the best one. This is a tradeoff of greedy algorithm :)

src/fheroes2/battle/battle_animation.cpp
```diff
@@ -104,8 +104,13 @@ int AnimationSequence::firstFrame() const
...
-        return static_cast<double>( _currentFrame ) / ( static_cast<double>( animationLength() ) );
+    if ( isValid() ) {
+        // We return the progress coefficient for moving creature sprite from one position to another.
+        // At the start and the end of the path we also have creature's static position,
+        // but there is no such position on the cells which creature path through.
+        // So to make the movement start and end more smoothly we add 0.5 to the frame number (since it starts from zero).
+        return ( static_cast<double>( _currentFrame ) + 0.5 ) / static_cast<double>( animationLength() );
+    }
 
     return 0;
```
ihhub, line 115: What if we change a bit of logic like this:
```cpp
if ( !isValid() ) {
    return 0;
}

    // We return the progress coefficient for moving creature sprite from one position to another.
    // At the start and the end of the path we also have creature's static position,
    // but there is no such position on the cells which creature path through.
    // So to make the movement start and end more smoothly we add 0.5 to the frame number (since it starts from zero).
    return ( static_cast<double>( _currentFrame ) + 0.5 ) / static_cast<double>( animationLength() );
```
?

This is a bit better for code readability.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1666,50 +1666,50 @@ void Battle::Interface::RedrawTroopSprite( const Unit & unit )
...
+    fheroes2::Point offset = GetTroopPosition( unit, troopSprite );
 
     if ( _movingUnit == &unit ) {
         // Monster is moving.
-        // Here we're getting the first frame and then based on the offset from the first frame we calculate the position of the current frame.
-        // TODO: verify if it's the correct way as we have issues for monster movement animation.
-        const int monsterIcnId = unit.GetMonsterSprite();
-        const fheroes2::Sprite & firstMonsterFrame = fheroes2::AGG::GetICN( monsterIcnId, _movingUnit->animation.firstFrame() );
-        const int32_t ox = troopSprite.x() - firstMonsterFrame.x();
+        // Unit coordinates relevant to its position are set in Sprite coordinates 'x'  and 'y' (with bugs for some units).
```
ihhub, line 1675: I suggest to rephrase `(with bugs for some units)` into `(some values are set incorrectly)`. Because a coordinate cannot have a bug :)

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1666,50 +1666,50 @@ void Battle::Interface::RedrawTroopSprite( const Unit & unit )
...
+        // The offset data from BIN file is used in unit info dialog and to correct horizontal movement range when moving diagonally.
```
ihhub, line 1676: Let's change `unit info dialog` into `Monster info dialog`. This is how the window called within the code and in the UI.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2983,28 +2985,53 @@ void Battle::Interface::MouseLeftClickBoardAction( int themes, const Cell & cell
...
     if ( unit.isFinishAnimFrame() && unit.animation.animationLength() != 1 ) {
         // If it is the last frame in the animation sequence whith more than one frame or if we have no frames.
         return;
     }
 
     LocalEvent & le = LocalEvent::Get();
-    const uint64_t frameDelay = ( unit.animation.animationLength() > 0 ) ? delay / unit.animation.animationLength() : 0;
 
-    while ( le.HandleEvents( Game::isCustomDelayNeeded( frameDelay ) ) ) {
+    // In a loop we wait for the delay and then display the next frame.
```
ihhub, line 3017: A slight correction of the comment:
```cpp
// In the loop below we wait for the delay and then display the next frame.
```

src/fheroes2/game/game_delays.h
```diff
@@ -55,6 +55,7 @@ namespace Game
         BATTLE_POPUP_DELAY,
         BATTLE_COLOR_CYCLE_DELAY,
         BATTLE_SELECTED_UNIT_DELAY,
+        BATTLE_UNIT_MOVEMENT_DELAY,
```
ihhub, line 58: I suggest to name it as `CUSTOM_BATTLE_UNIT_MOVEMENT_DELAY` to indicate that it is not a fixed delay. Then we should rename `setUnitMovementDelay` into `setCustomUnitMovementDelay` accordingly.

src/fheroes2/game/game_delays.h
```diff
@@ -84,6 +88,9 @@ namespace Game
     int HumanHeroAnimSkip();
     int AIHeroAnimSkip();
 
+    // Returns true if every of delay type has passed.
+    bool isDelayPassed( const std::vector<Game::DelayType> & delayTypes );
```
ihhub, line 92: I realized that the name `isDelayNeeded` is very confusing (we can change it in a separate pull request) so I am proposing to rename `isDelayPassed()` into `hasEveryDelayPassed()`. In this case it is clear what this function does. What do you think?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3496,63 +3537,107 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
 
-        if ( show_anim ) {
-            // If a wide flyer is flying to the left its horizontal position should be shifted to the left by one cell.
-            if ( canFly && isWide && !unit.isReflect() ) {
-                _movingPos.x -= CELLW;
-            }
-            AudioManager::PlaySound( unit.M82Move() );
+        // Render the unit movement with the movement sound.
+        AudioManager::PlaySound( unit.M82Move() );
+        AnimateUnitWithDelay( unit );
```
ihhub, line 3608: I think in the future we must adjust sounds calls and synchronize them with frames. For example, in this particular situation there a chance that inside `AnimateUnitWithDelay()` we won't render a frame immediately since the delay hasn't been passed. This leads to a situation that the sound is being played but the monster is still standing. This might not be visible for an ordinary player and it is more of the perfection of the effects in the game.

You can ignore this comment or if you think it is a valid one add a TODO comment for such type of places for the future improvement.

src/fheroes2/battle/battle_interface.h
```diff
@@ -343,8 +343,12 @@ namespace Battle
         void RedrawLightningOnTargets( const std::vector<fheroes2::Point> & points, const fheroes2::Rect & drawRoi ); // helper function
         void RedrawRaySpell( const Unit & target, int spellICN, int spellSound, int32_t size );
 
+        // Wait for all possible battlefield action delays that could be set in previous functions to pass.
+        // Use this if your function may be called from other functions with different render delay types.
+        void WaitForActionsDelays();
```
ihhub, line 348: What do you think about renaming this method into `WaitForAllActionDelays()`?

ihhub (changes requested):
Hi @Districh-ru , I left here several comments mostly about the small corrections. Would you mind please to take at them when you have time?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3496,63 +3537,107 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+
+        // Switch animation to MOVING before going through the path.
+        unit.SwitchAnimation( Monster_Info::MOVING );
+    }
+    else {
+        // Every ground unit should start its movement from the special 'MOVE_START' animation
+        // or if it moves only for one cell its animation must be 'MOVE_QUICK'. So a check for 1 cell path is made.
+        // If a wide unit moves backwards for 1 cell, it turns twice, so it has 3 path points. And its first 'dst' point is equal to the last.
+        // TODO: try to rewrite path generation and movement of wide creatures to get more clear and unique path with non-wide (to get rid of of 'wide creature patches').
+        if ( isOneStepPath || ( isWide && path.size() == 3 && ( *dst == *( pathEnd - 1 ) ) ) ) {
```
oleg-derevenetz, line 3558: This condition looks rather arbitrary. What if wide unit moves like this:

https://user-images.githubusercontent.com/32623900/211188908-3dbfe2cd-a030-4050-987e-60cc36a352ea.mp4

? `path.size()` here is 3, but the first and last point of path do not match.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi @ihhub, I've applied you suggestions. Thanks for the review and sorry for my English. :)
I left this TODO, because I don't like how it sounds now: The Cavalry has a moving sound for 2 steps and when moving to the second cell the new sound overlaps the previous step sound even on the lowest battle speed.

ihhub: No problem at all!

ihhub (approved):

src/fheroes2/battle/battle_interface.cpp
Districh-ru: And here, maybe, we have also to change the sound speed according to the battle speed.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi, @oleg-derevenetz, thanks, I've missed this move.
That's also the reason I was thinking of changing the path generation for wide creatures. :)
I was thinking about this move and then the moving back, like in video I posted earlier (https://github.com/ihhub/fheroes2/pull/6398#discussion_r1064022397) confused me. :)
Yes, the wide creature can move one back step diagonally up and diagonally down. And we have to check that the end of path is in the cell next to the fist step in the path.
I could not find an easy logic to check this. :)
Please, check my last commit when you have time.

oleg-derevenetz: Movement from that video was possible because there are 2 candidate cells with equal "cost" (distance to target) for the first step. If one of these cells is occupied by an obstacle or some other unit (like in my video) the movement may become different.

ihhub:
Hi @Districh-ru , can you please fix conflict here?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3496,63 +3540,112 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+    else {
+        // Every ground unit should start its movement from the special 'MOVE_START' animation
+        // or if it moves only for one cell its animation must be 'MOVE_QUICK'. So a check for 1 cell path is made.
+        // If a wide unit moves backwards for 1 cell, it turns twice, so it has 3 path points. And its first 'dst' point is equal to the last.
+        // If it moves back diagonally up or down, we check if the end of the path is the cell next to the first cell in the path.
+        // TODO: try to rewrite path generation and movement of wide creatures to get more clear and unique path with non-wide (to get rid of of 'wide creature patches').
+        if ( isOneStepPath
+             || ( isWide && path.size() == 3
+                  && ( ( *( pathEnd - 1 ) == *dst ) || ( *( pathEnd - 1 ) == ( *dst - ARENAW ) ) || ( *( pathEnd - 1 ) == ( *dst - ARENAW + 1 ) )
+                       || ( *( pathEnd - 1 ) == ( *dst + ARENAW ) ) || ( *( pathEnd - 1 ) == ( *dst + ARENAW + 1 ) ) ) ) ) {
```
oleg-derevenetz, line 3565: Looks a bit scary :) And overcomplicated. How about this:

```cpp
const bool isOneStepPath = [&unit, &path, isWide]() {
    if ( path.size() == 1 ) {
        return true;
    }

    if ( !isWide ) {
        return false;
    }

    // If wide unit performs 3 movements and the first movement is a turn back, then it's
    // path consists of only one "real" movement, because...
    if ( path.size() == 3 && path[0] == unit.GetTailIndex() ) {
        // ... its last movement should be a return to the normal position
        assert( Board::GetDirection( path[1], path[2] ) == ( unit.isReflect() ? LEFT : RIGHT ) );

        return true;
    }

    return false;
}();
```

?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3496,63 +3540,112 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
...
+        else {
+            // Check for possible bridge close action, after walking unit's end of movement to the next cell.
+            // This check should exclude the flying creature because it can't 'hang' here to wait
+            // for bridge to close. For this creature, the bridge should close after it lands.
+            if ( bridge && bridge->AllowUp() ) {
+                _movingUnit = nullptr;
+                unit.SwitchAnimation( Monster_Info::STAND_STILL );
+                bridge->ActionUp();
+                _movingUnit = &unit;
+                if ( !isOneStepPath && dst != ( pathEnd - 2 ) ) {
```
oleg-derevenetz, line 3633: I suppose that here and below at line 3642 `isOneStepPath` is used as a replacement of `path.size() > 1`? Then it's better to use the latter because it's much more understandable and will remove quite logical questions, such as the one that @ihhub has.

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left a few proposals, please take a look when possible.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Many thanks, @oleg-derevenetz for this elegant solution!
I also want to make a copy of this lambda-function for `isOneStepLeft` to use it after the bridge actions. And while testing I found a bug with bridge:
If a wide creature wants to enter the bridge gate from the moat, walking back, the bridge does not open. And the next turn this creature walks though the bridge.
This bug can be also reproduced in current master build.

https://user-images.githubusercontent.com/113276641/211203212-822e75d9-a815-407d-835f-142021ec7274.mp4

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: > And while testing I found a bug with bridge

In original HoMM2, bridge should be lowered if wide unit steps on any part of it, even by its tail:

https://user-images.githubusercontent.com/32623900/211206227-ccc6476f-8b8b-4951-b4ff-f39f70f59af0.mp4

I believe we should do the same. This issue can be addressed in a separate PR.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: I couldn't quickly make a function to check if wide creatures had one step left after the bridge action, so I left `TODO`.
I think we can do it later in a separate PR.
Of the wide walking creatures, only Cavalry and Wolves have different `MOVING` and `QUICK_MOVE` animations, and the difference is only 1 frame.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: If you take a step back in fhefore2, as in the video, the bridge normally lowers.
The bug only occurs if a wide creature starts its turn in a moat.
I'll open an issue.

oleg-derevenetz: Did you noticed that in OG Unicorn turns back while the bridge is lowering? :)

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: I believe in the cycle we should just measure the distance between `dst` and `pathEnd` instead of `path.size()` and use relative (to `dst`) indexes instead of hardcoded ones.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Yes, we return back to the pathfinding algorithm. :)
If it prefers 1 cell path (down-right) over 2 cell (down-left and then right):

https://user-images.githubusercontent.com/113276641/211210214-9158ec73-6dc5-444e-9968-9a051b7a7d80.mp4

But there could be a case when a user wants to move Unicorn only for one cell down-left. It will be a correct path, its tail will occupy the bridge area.

oleg-derevenetz: > Yes, we return back to the pathfinding algorithm. :)

These movements are perfectly valid:

https://user-images.githubusercontent.com/32623900/211209379-c29388e7-209c-42e8-ab93-79211d30dfc8.mp4

and they should lower the bridge, but they didn't.

src/fheroes2/game/game_delays.cpp
```diff
@@ -205,6 +211,21 @@ uint32_t Game::ApplyBattleSpeed( uint32_t delay )
     return static_cast<uint32_t>( battleSpeedAdjustment * ( 10 - Settings::Get().BattleSpeed() ) * delay );
 }
 
+bool Game::hasEveryDelayPassed( const std::vector<Game::DelayType> & delayTypes )
+{
+    if ( delayTypes.empty() ) {
```
oleg-derevenetz, line 216: This check is not really necessary, because if `delayTypes` is empty, then the `for` loop below will be skipped immediately and `true` will be returned.

oleg-derevenetz (approved):
Since the implementation of the check for the "last step" in the movement cycle for wide units has been postponed, LGTM. Just one optional minor remark.

src/fheroes2/game/game_delays.cpp
ihhub: Hi @oleg-derevenetz , I remove the check here and in the function below.

ihhub:
@Districh-ru , huge thanks for these very useful fixes!

PR #6400, opened 2022-12-20
Fix forced IDLE animation after switching to STATIC animation. (Battlefield)

### LINKED ISSUE #6399: Combat screen, idle animation before retaliation
### Preliminary checks

### Platform

Windows

### Describe the bug

In rare cases I've noticed, there's a noticeable gap after creature was attacked and before it retaliated. If I'm not mistaken, right before creature retaliates, it shows some idle activity.

### Save file

None.

### Additional info

_No response_

Districh-ru:
close #6399

It seems that the problem was that while creature does its non `STATIC`/`IDLE` animation (attack, wince, moving) the delay till the next `IDLE` animation passes and after switching to `STATIC` animation the game automatically switches the creature to `IDLE` animation. So the creature could start its `IDLE` animation after wince, instead if `STATIC`, which changes to attack animation in the middle of `IDLE`.

### DIFF
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -1615,6 +1615,12 @@ bool Battle::Unit::isHaveDamage() const
 bool Battle::Unit::SwitchAnimation( int rule, bool reverse )
 {
     animation.switchAnimation( rule, reverse );
+
+    if ( rule == Monster_Info::STATIC ) {
+        // Reset the delay before switching to the 'IDLE' animation from 'STATIC'.
+        checkIdleDelay();
+    }
+
     return animation.isValid();
 }
 
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -1614,12 +1614,12 @@ bool Battle::Unit::isHaveDamage() const
 
 bool Battle::Unit::SwitchAnimation( int rule, bool reverse )
 {
-    animation.switchAnimation( rule, reverse );
-
-    if ( rule == Monster_Info::STATIC ) {
+    if ( rule == Monster_Info::STATIC && GetAnimationState() != Monster_Info::IDLE ) {
         // Reset the delay before switching to the 'IDLE' animation from 'STATIC'.
         checkIdleDelay();
     }
+
+    animation.switchAnimation( rule, reverse );
 
     return animation.isValid();
 }


Branikolog:
Hi, @Districh-ru 
After a short testing the issue is no more occurred for me. 👍

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1614,7 +1614,13 @@ bool Battle::Unit::isHaveDamage() const
 
 bool Battle::Unit::SwitchAnimation( int rule, bool reverse )
 {
+    if ( rule == Monster_Info::STATIC && GetAnimationState() != Monster_Info::IDLE ) {
```
ihhub, line 1617: Hi @Districh-ru , the fix looks pretty straightforward and I see no issues in the code. However, for me it looks like a hack rather than fixing the original problem. Essentially we have 3 places where we update the animation of monsters using `Game::validateAnimationDelay( Game::BATTLE_FRAME_DELAY )`. In one of them we made a fix a delay after wince animation (#5073). The root of the problem was that we were not waiting for the last frame after completing rendering it leading to a fact that the next frame was rendered immediately.

However, I might be wrong in this case. What do you think about this?

Districh-ru: Hi, @ihhub, you are certainly right about the issue, fixed in #5073.
IMHO the issue in most our rendering functions is that we use different asynchronous delays and first wait for some delay, in example `Game::BATTLE_FRAME_DELAY`, and then do rendering. And after the last loop and rendering the last frame we do not wait for this delay to show the user the results of render. After this we can switch to another animation with its delay and its time counter, which was not reset by checking the `Game::BATTLE_FRAME_DELAY`. So if we check in the next animation function for, in example, `Game::validateAnimationDelay( Game::BATTLE_OPPONENTS_DELAY )` it immediately returns true, as the counter for this type of delay was reset long-long before our previous renderings with `Game::BATTLE_FRAME_DELAY`. And we immediately do render a frame from this animation and user does not have time to see the previous frame as there was no delay before it (and here we waste our computation powers for 1 lost frame). So maybe it will be good in future to set what delays must sync their `reset()` and optimize all rendering functions. Or change animation functions to first render the frame and then - wait for delay (I did this in `AnimateUnitWithDelay()`).
But this are rendering delays, which are not the part of current PR. :)
In current PR I've changed the behavior of `idleTimer` delay. This delay is another delay which is (and has to be) asynchronous with animation delays. It sets the delay (and resets the time counter) before automatic switching unit animation state from `STATIC` to `IDLE`. The comparison  of this `idleTimer` counter with its delay does in `IdleTroopsAnimation()`, called by `RedrawArmies()`, and the frame renders with its animation delay, but the animation state changes according to its own `idleTimer` delay. And `idleTimer` delay does not affect on frame change in current animation state and on frame render delays. So this things are not connected.
But while doing (with waiting for rendering) of animations, other than `IDLE`, the `idleTimer` counter does not reset. So when we try to switch animation state to `STATIC` the game checks the `idleTimer` counter and compares it with its delay, which has passed long time ago, as the unit was doing some other animation for some time.
And this fix is only the issue, described in last paragraph. :)

ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for fixing this issue!

PR #6401, opened 2022-12-21
Add links to download the ARM64 Linux build, distinguish the archive names between x86-64 and ARM64

oleg-derevenetz:
Because now we have ARM64 builds.

Hi @ihhub could you please postpone the release creation until this PR will be merged since the names of x86-64 linux ZIP archives have been changed?

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -14,7 +14,7 @@ jobs:
           env: { FHEROES2_STRICT_COMPILATION: ON, FHEROES2_WITH_TOOLS: ON, FHEROES2_WITH_SDL1: ON }
           package_name: fheroes2_ubuntu_x86-64_SDL1.zip
           package_files: fheroes2 LICENSE script/linux/install_sdl_1.sh download_demo_version.sh extract_homm2_resources.sh changelog.txt README.txt files/lang/*.mo files/data/*.h2d
-          release_name: Ubuntu (Linux) build with SDL1 (latest commit)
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @oleg-derevenetz , sure, it can wait.

ihhub (approved):

oleg-derevenetz:
Hi @ihhub it seems that the list of required checks needs to be updated after merging of this PR, because names of some jobs have been changed.

ihhub:
Hi @oleg-derevenetz , I will update them after the release.

If you have no other things to add in this pull request I will proceed with merging it and the release.

oleg-derevenetz:
Hi @ihhub I believe I'm done with this PR :)

ihhub:

oleg-derevenetz:
Hi @ihhub since you apparently decided not to offer SDL1 binaries in the 1.0.0, should we remove the corresponding download links from the `INSTALL.md`?

ihhub:
Hi @oleg-derevenetz , the move was to avoid confusion from users having cryptic (SDL1 and SDL2) words in file names and also to reduce the amount of choices to pick from (Windows users had wooping 8 links). Yes, we should remove SDL1 from links.

If you think that we should add SDL 1 into the release please feel to express this opinion :)

oleg-derevenetz:
@ihhub No, I just think that we should remove the corresponding download links from the `INSTALL.md` (see #6402) :) Otherwise we still have non-working links to SDL1 downloads, which may confuse users.

PR #6402, opened 2022-12-21
INSTALL.md: remove SDL1 links

oleg-derevenetz:
SDL1 builds were removed from the 1.0.0 release.


oleg-derevenetz:
@zenseii

> We should remember to remove the links from the snapshot part on the bottom of the file too:

Only when we remove SDL1 support entirely. For now it's not the case, there are still workflows that produce these files.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this quick fix!

PR #6405, opened 2022-12-21
GitHub Actions: use ubuntu-20.04 image to build Linux x86-64 images using the Make workflow

### LINKED ISSUE #6403: 1.0.0 seems to require GLIBC 2.35, is that intentional?
### Your question

The 1.0.0 release (fheroes2_ubuntu_x86-64_SDL2.zip) seems to require GLIBC 2.35 or newer, was that change intentional or is my system broken?

0.9.21 was working fine with GLIBC 2.31.

### Additional info

_No response_

oleg-derevenetz:
close #6403

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -41,12 +41,12 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Install dependencies (Linux)
-      if: ${{ matrix.config.os == 'ubuntu-latest' }}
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @oleg-derevenetz , I think we should change `ubuntu-latest` for the rest of cases to `ubuntu-22.04` to avoid the same issue we faced with x86-64. What do you think?

ihhub (approved):

ihhub:

PR #6411, opened 2022-12-23
Optimize multiple places for rendering

ihhub:
- all event loops during battle animations did not check whether it is allowed to do rendering within handle event loop
- custom delays were not reset properly causing slight delays in first frame to be rendered. Video playback was affected the most
- avoid excessive rendering for multiple windows. Render only the area which was changed
- add todo comments for the future changes
- some animations during battle did not play the last frame like monster animation, hero's animation and hero's flag animation
- small code cleanup

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -312,16 +312,6 @@ namespace Battle
     int GetDirectionFromCursorSword( uint32_t sword );
     int GetCursorFromSpell( int );
 
-    struct CursorPosition
-    {
-        CursorPosition()
-            : index( -1 )
-        {}
-
-        fheroes2::Point coord;
-        int32_t index;
-    };
-
     class StatusListBox : public ::Interface::ListBox<std::string>
     {
     public:
@@ -856,20 +846,25 @@ void Battle::OpponentSprite::Redraw( fheroes2::Image & dst ) const
         fheroes2::Blit( hero, dst, offset.x + LEFT_HERO_X_OFFSET + hero.x(), offset.y + LEFT_HERO_Y_OFFSET + hero.y() );
 }
 
-void Battle::OpponentSprite::Update()
+bool Battle::OpponentSprite::updateAnimationState()
 {
     if ( _currentAnim.isLastFrame() ) {
         if ( _animationType != OP_STATIC ) {
-            if ( _animationType != OP_CAST_MASS && _animationType != OP_CAST_UP && _animationType != OP_CAST_DOWN )
+            if ( _animationType != OP_CAST_MASS && _animationType != OP_CAST_UP && _animationType != OP_CAST_DOWN ) {
                 SetAnimation( OP_STATIC );
+                return true;
+            }
         }
         else if ( _idleTimer.checkDelay() ) {
             SetAnimation( ( Rand::Get( 1, 3 ) < 2 ) ? OP_IDLE2 : OP_IDLE );
+            return true;
         }
+
+        return false;
     }
-    else {
-        _currentAnim.playAnimation();
-    }
+
+    IncreaseAnimFrame();
+    return true;
 }
 
 Battle::Status::Status()
@@ -2430,13 +2425,12 @@ void Battle::Interface::getPendingActions( Actions & actions )
 
 void Battle::Interface::HumanTurn( const Unit & b, Actions & a )
 {
-    Cursor & cursor = Cursor::Get();
+    Cursor::Get().SetThemes( Cursor::WAR_POINTER );
     LocalEvent & le = LocalEvent::Get();
 
     // Reset the cursor position to avoid forcing the cursor shadow to be drawn at the last position of the previous turn.
     index_pos = -1;
 
-    cursor.SetThemes( Cursor::WAR_POINTER );
     _currentUnit = &b;
     humanturn_redraw = false;
     humanturn_exit = false;
@@ -2452,9 +2446,6 @@ void Battle::Interface::HumanTurn( const Unit & b, Actions & a )
 
     popup.Reset();
 
-    // safe position coord
-    CursorPosition cursorPosition;
-
     Redraw();
 
     std::string msg;
@@ -2462,7 +2453,10 @@ void Battle::Interface::HumanTurn( const Unit & b, Actions & a )
 
     ResetIdleTroopAnimation();
 
-    while ( !humanturn_exit && le.HandleEvents() ) {
+    // TODO: update delay types within the loop to avoid rendering slowdown.
+    const std::vector<Game::DelayType> delayTypes{ Game::BATTLE_FLAGS_DELAY };
+
+    while ( !humanturn_exit && le.HandleEvents( Game::isDelayNeeded( delayTypes ) ) ) {
         // move cursor
         int32_t indexNew = -1;
         if ( le.MouseCursor( { _interfacePosition.x, _interfacePosition.y, _interfacePosition.width, _interfacePosition.height - status.height } ) ) {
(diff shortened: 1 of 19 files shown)


ihhub:
Hi @Districh-ru , I would like to see your opinion about these changes, especially about animation of monsters, flags and heroes. Some monsters have become more smooth in animation but some like wolves not. It would be great if you can verify that I didn't break anything in your past efforts for animation.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2996,34 +2990,51 @@ void Battle::Interface::AnimateUnitWithDelay( Unit & unit, uint32_t delay )
...
+    while ( le.HandleEvents( Game::isCustomDelayNeeded( frameDelay ) ) ) {
         CheckGlobalEvents( le );
 
         if ( Game::validateCustomAnimationDelay( frameDelay ) ) {
-            if ( unit.isFinishAnimFrame() )
+            Redraw();
+
+            if ( unit.isFinishAnimFrame() ) {
+                // We have reached the end of animation.
                 break;
```
Districh-ru, line 3009: In this place after we render the last frame we exit the render loop immediately and don't wait for the `frameDelay` from this frame till the next one. And if we render the first frame of the next animation sequence immediately, as it is done here, we can loose this last frame of current animation or do a shorter (random) delay after it. So this could be the reason for non-smooth animation of creatures.
I got the non-smooth movement animation for Sprites and when I put the hack `while ( !Game::validateCustomAnimationDelay( frameDelay ) ) ;` here before the `break;` the animation becomes smooth.

The same I can say for the other similar functions (`AnimateOpponents` , `RedrawTroopDefaultDelay` and others).

I think there should always be a pair: a rendered frame and its delay. We can wait for the delay and then render the frame, or first (and IMHO it's better way) render the frame and then always wait for the delay.

Districh-ru (changes requested):
Hi, @ihhub!
I left one general comment that can be applied to multiple places in the PR.
Please take a look when you have time.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: To wait the delay after rendering the last frame we can add a `bool` variable:

```cpp
    bool finishRender = false;
    while ( le.HandleEvents( Game::isCustomDelayNeeded( frameDelay ) ) ) {
        CheckGlobalEvents( le );

        if ( Game::validateCustomAnimationDelay( frameDelay ) ) {
            if ( finishRender ) {
                // There is nothing left to render in this loop.
                break;
            }
            
            Redraw();

            if ( unit.isFinishAnimFrame() ) {
                // We have reached the end of animation. But we have to wait for frameDelay.
                finishRender = true;
            }
            else {
                unit.IncreaseAnimFrame();
            }
        }
    }
```
Or act here as it was before:
```cpp
    // Reset the delay.
    Game::validateCustomAnimationDelay( frameDelay );
    // Make sure that the delay was reset.
    assert( Game::isCustomDelayNeeded( frameDelay ) );
    // Render the first frame immediately.
    Redraw();

    while ( le.HandleEvents( Game::isCustomDelayNeeded( frameDelay ) ) ) {
        CheckGlobalEvents( le );

        if ( Game::validateCustomAnimationDelay( frameDelay ) ) {
            if ( unit.isFinishAnimFrame() ) {
                // We have reached the end of animation.
                break;
            }

            unit.IncreaseAnimFrame();

            Redraw();
        }
    }
```

ihhub: I the second approach is better with slight corrections.

ihhub: I've updated the code. Could you please take a look at it again?

Districh-ru: In updated code there is no render of the last animation frame, as we increase the frame number and do not `Redraw();` after it. And after we reach the last frame by `unit.IncreaseAnimFrame();` the next is that after the delay we do `break;`.
To fix it you can remove `unit.IncreaseAnimFrame();` before the loop and change places of `unit.IncreaseAnimFrame();` and `Redraw();` inside the loop.
And also to be consistent with the in-loop code, where we first reset the delay and then do `Redraw();` maybe it might be better to do it in the same order before the loop.
```cpp
    // Reset the delay to wait till the next frame.
    Game::AnimateResetDelay( Game::DelayType::CUSTOM_DELAY );

    // Render the first frame before waiting any delay.
    Redraw();

    const uint64_t frameDelay = ( unit.animation.animationLength() > 0 ) ? delay / unit.animation.animationLength() : 0;

    while ( le.HandleEvents( Game::isCustomDelayNeeded( frameDelay ) ) ) {
        CheckGlobalEvents( le );

        if ( Game::validateCustomAnimationDelay( frameDelay ) ) {
            if ( unit.isFinishAnimFrame() ) {
                // We have reached the end of animation.
                break;
            }

            unit.IncreaseAnimFrame();
            Redraw();
        }
    }
```

And ... an idea came to me:
Maybe we have to reset all the delays at the final render in current function and in the next function wait for its delay which was reset in previous function. So we will keep the previous code (before this PR) and only add a reset for all rendering delays with some new method `.AnimateResetAllDelays()`.
Doing this we can make all animation to preserve its delays no matter how many calculations have to be done between them. In current PR code the delays between the function calls can be bigger than intended as many calculations are made after previous render and the reset of the new delay.

ihhub: Hi @Districh-ru , the previous implementation will not work with this as for the first Render() call we should have to wait. I will modify the code to make it work, hopefully, without breaking anything.

ihhub: @Districh-ru , I implemented your idea with small modifications. Could you please take a look?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3220,6 +3212,10 @@ void Battle::Interface::RedrawActionAttackPart2( Unit & attacker, const TargetsI
 
     // targets damage animation
     RedrawActionWincesKills( targets, &attacker );
+
+    // Reset the delay to wait till the next frame.
+    Game::AnimateResetDelay( Game::DelayType::BATTLE_FRAME_DELAY );
+
     RedrawTroopDefaultDelay( attacker );
```
Districh-ru, line 3219: `BATTLE_FRAME_DELAY` is reset in `RedrawActionWincesKills` function so there is no need to reset it here again. We can reset it at the beginning of function `RedrawActionAttackPart2`, so it will also include all calculations, which are made in this function before rendering a frame in `RedrawActionWincesKills`. And thу time needed for  calculations will not be added to the delay before frame render. (IMHO all delay resets has to be done as close as possible to the beginning of the function).
Also this type of delay is used in previous function `RedrawActionAttackPart1` for melee attacks, so in this case it was already reset there. And resetting it here for second time could make a longer delay from the previous frame render. To avoid this we can check that the delay was set and is already running:
```cpp
    // Reset the delay to wait till the next frame.
    if ( !Game::isDelayNeeded( { Game::DelayType::BATTLE_FRAME_DELAY } ) ) {
        Game::AnimateResetDelay( Game::DelayType::BATTLE_FRAME_DELAY );
    }
```

Also with the new rendering approach it is better to move
https://github.com/ihhub/fheroes2/blob/77acd650973673c13a51add23c633c335ff2f502/src/fheroes2/battle/battle_interface.cpp#L3320-L3323
in `RedrawActionWincesKills` function to the end of the `if` statement to avoid waiting for `BATTLE_FRAME_DELAY` twice.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3451,6 +3447,9 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
     unit.SwitchAnimation( Monster_Info::MOVING );
     const uint32_t movementFrames = static_cast<uint32_t>( unit.animation.animationLength() );
 
+    // Reset the delay to wait till the next frame.
+    Game::AnimateResetDelay( Game::DelayType::CUSTOM_DELAY );
+
```
Districh-ru, line 3452: We can also move this to the beginning of this function: 
```cpp
    if ( path.empty() ) {
        return;
    }

    // Reset the delay to wait till the next frame.
    Game::AnimateResetDelay( Game::DelayType::CUSTOM_DELAY );
...
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3635,6 +3634,10 @@ void Battle::Interface::RedrawActionFly( Unit & unit, const Position & pos )
 
     unit.SwitchAnimation( Monster_Info::FLY_UP );
     AudioManager::PlaySound( unit.M82Tkof() );
+
+    // Reset the delay to wait till the next frame.
+    Game::AnimateResetDelay( Game::DelayType::CUSTOM_DELAY );
+
```
Districh-ru, line 3640: And here too :)
```cpp
    // check if we're already there
    if ( unit.GetPosition().contains( destIndex ) ) {
        return;
    }

    // Reset the delay to wait till the next frame.
    Game::AnimateResetDelay( Game::DelayType::CUSTOM_DELAY );
```

Districh-ru (changes requested):
Hi, @ihhub! I left 3 comments with suggestions to slightly improve the smoothness of animation.
Please take a look when you have time.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: @ihhub, well done!
Now the movement animation looks smooth. 👍

Districh-ru (changes requested):
Hi, @ihhub, now the animations are smooth and the zombie flies do not freeze for a while after the attack animation. 👍 
I have only one suggestion left:
Could you, please, add the delay reset to the beginning of `RedrawActionWincesKills` function (line 3271):
```cpp
    // Reset the delay to wait till the next frame.
    if ( !Game::isDelayNeeded( { Game::DelayType::BATTLE_FRAME_DELAY } ) ) {
        Game::AnimateResetDelay( Game::DelayType::BATTLE_FRAME_DELAY );
    }
```
This function is called from different places and the delay may be not reset in all cases.

Districh-ru (approved):
👍

oleg-derevenetz (approved):

PR #6412, opened 2022-12-24
Fix controller support for PS Vita

### LINKED ISSUE #6410: Buttons no longer work in version 1.0.0
### Preliminary checks

### Platform

PlayStation Vita

### Describe the bug

After updating to version 1.0.0, the buttons and analog sticks other than the touchscreen no longer work (The binding of the keys is invalid.). Scrolling the map on the main screen is now easier (In previous version, needed to move the cursor to the edge of the screen to scroll.), but it's very difficult to accurately select a destination on the map or target enemies during combat (When you want to attack from a specified direction, it is more accurate to use the joystick to control the cursor.). Also, the hotkey settings are for the pc, so you can't change the controls.

### Save file

*

### Additional info

_No response_

ihhub:
close #6410

### DIFF
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -25,6 +25,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <map>
+#include <ostream>
 #include <set>
 #include <utility>
 #include <vector>
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -50,7 +50,6 @@
 #include "audio.h"
 #include "image.h"
 #include "localevent.h"
-#include "logging.h"
 #include "pal.h"
 #include "screen.h"
 #include "tools.h"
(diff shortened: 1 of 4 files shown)


src/engine/localevent.cpp
```diff
@@ -1232,6 +1232,16 @@ bool LocalEvent::HandleEvents( const bool sleepAfterEventProcessing, const bool
...
             break;
+        case SDL_JOYAXISMOTION:
+        case SDL_JOYBALLMOTION:
+        case SDL_JOYHATMOTION:
+        case SDL_JOYBUTTONDOWN:
+        case SDL_JOYBUTTONUP:
+        case SDL_JOYDEVICEADDED:
+        case SDL_JOYDEVICEREMOVED:
+        case SDL_CONTROLLERDEVICEREMAPPED:
+            // All these joystick and controller events aren't handled.
```
oleg-derevenetz, line 1243: Hi @ihhub So we should enable them only to not handle them? :) Are they handled somewhere inside the SDL itself?

ihhub: Funny thing is that controllers are derived from joysticks. If we check **SDL_ControllerAxisEvent**, **SDL_ControllerButtonEvent** and **SDL_ControllerDeviceEvent**, all of them point to `Joystick index`. Before my changes in #6213 we had enabled joystick events. I didn't know that they are correlated so I explicitly disabled them. Now it looks like PS Vita works only if joystick events are enabled.

Digging into the SDL code :
- SDL_gamecontroller.c, function `SDL_GameControllerEventWatcher` for **SDL_JOYBUTTONDOWN** and **SDL_JOYBUTTONUP** events calls `HandleJoystickButton` where internally it calls `SDL_PrivateGameControllerButton` and sets events **SDL_CONTROLLERBUTTONDOWN** and **SDL_CONTROLLERBUTTONUP**. Magic!

oleg-derevenetz: I suggest to add a comment describing why they shouldn't be disabled despite they are not handled by our code.

ihhub: Done.

oleg-derevenetz (approved):

PR #6413, opened 2022-12-24
Optimize Death Wave spell effect rendering. Battlefield.

Districh-ru:
Reworked the rendering of the Death Wave spell effect. Now the game only renders the part of the screen where the effect is located.
Also in this PR before the spell effect, all creatures are now forced to stop their `IDLE` animation and their unit counter is now turned off, as it is not so aesthetic when the creature freezes in some `IDLE` pose during the spell animation and when the unit counters, that are part of the user interface , are distorted by the spell. The same applies to the Holy Shout spell.

<details><summary>A video of the Death Wave spell effect</summary>

https://user-images.githubusercontent.com/113276641/209446836-30a4f65d-3acd-4739-ba2d-0040ca8fde31.mp4

</details>

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -4857,9 +4857,13 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
 
     fheroes2::Rect area = GetArea();
     // Cut out the battle log image so we don't use it in the death wave effect.
-    area.height -= 36;
+    area.height -= status.height;
+    // And if listlog is open, then cut off it too.
+    if ( listlog && listlog->isOpenLog() ) {
+        area.height -= listlog->GetArea().height;
+    }
 
-    const fheroes2::Sprite & copy = fheroes2::Crop( _mainSurface, area.x, area.y, area.width, area.height );
+    const fheroes2::Sprite & battleFieldCopy = fheroes2::Crop( _mainSurface, area.x, area.y, area.width, area.height );
     // The death wave horizontal length in pixels.
     const int32_t waveLength = 38;
     const int32_t waveStep = 5;
@@ -4875,36 +4879,35 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
         deathWaveCurve.push_back( static_cast<int32_t>( std::round( strength * ( cos( posX / waveLimit ) / 2 - 0.5 ) ) ) - 1 );
     }
 
-    AudioManager::PlaySound( M82::MNRDEATH );
-
     // Take into account that the Death Wave starts outside the battle screen.
     area.x -= waveLength;
     // The first frame of spell effect must have a part of the spell, so we start from its first position.
     int32_t position = waveStep;
     fheroes2::Display & display = fheroes2::Display::instance();
 
+    // Prepare the blank image for the Death Wave spell effect with the transform layer equal to "0"
+    fheroes2::Image spellEffect( waveLength, area.height );
+    std::fill( spellEffect.transform(), spellEffect.transform() + static_cast<size_t>( waveLength * area.height ), static_cast<uint8_t>( 0 ) );
+
+    AudioManager::PlaySound( M82::MNRDEATH );
+
     while ( le.HandleEvents() && position < area.width + waveLength ) {
         CheckGlobalEvents( le );
 
         if ( Game::validateAnimationDelay( Game::BATTLE_DISRUPTING_DELAY ) ) {
             const int32_t wavePositionX = ( area.x + position < 0 ) ? 0 : ( area.x + position );
+            const int32_t waveWidth = position > waveLength ? ( position > area.width ? ( waveLength - position + area.width ) : waveLength ) : position;
             const int32_t restorePositionX = ( wavePositionX < waveStep ) ? 0 : ( wavePositionX - waveStep );
             const int32_t restoreWidth = wavePositionX < waveStep ? wavePositionX : waveStep;
 
-            const fheroes2::Image spellEffect = fheroes2::CreateDeathWaveEffect( copy, position, deathWaveCurve );
-
-            const fheroes2::Rect renderArea( _interfacePosition.x + restorePositionX, _interfacePosition.y + area.y, spellEffect.width() + restoreWidth, area.height );
+            const fheroes2::Rect renderArea( _interfacePosition.x + restorePositionX, _interfacePosition.y + area.y, waveWidth + restoreWidth, area.height );
 
             // Place a copy of the original image where the Death Wave effect was on the previous frame.
-            fheroes2::Blit( copy, restorePositionX, area.y, display, renderArea.x, renderArea.y, restoreWidth, renderArea.height );
+            fheroes2::Blit( battleFieldCopy, restorePositionX, area.y, display, renderArea.x, renderArea.y, restoreWidth, renderArea.height );
 
             // Place the Death Wave effect to its new position.
-            fheroes2::Blit( spellEffect, display, renderArea.x + restoreWidth, renderArea.y );
-
-            // If the battle log was open we redraw it.
-            if ( listlog && listlog->isOpenLog() ) {
-                listlog->Redraw();
-            }
+            fheroes2::CreateDeathWaveEffect( spellEffect, battleFieldCopy, position, deathWaveCurve );
+            fheroes2::Blit( spellEffect, 0, 0, display, renderArea.x + restoreWidth, renderArea.y, waveWidth, area.height );
 
             // Render only the changed screen area.
             display.render( renderArea );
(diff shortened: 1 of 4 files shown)


src/fheroes2/gui/ui_tool.cpp
```diff
@@ -227,28 +227,33 @@ namespace fheroes2
 
     Image CreateDeathWaveEffect( const Image & in, const int32_t x, const std::vector<int32_t> & deathWaveCurve )
```
ihhub, line 228: We can accept another input parameter `Image & out` and then call resize lower in the code. For cases when we render the death wave effect in outside image borders the size remains the same which means that `out.resize()` call will do nothing. It will save resources for memory allocation and deallocation.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4876,18 +4877,43 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
...
 
         if ( Game::validateAnimationDelay( Game::BATTLE_DISRUPTING_DELAY ) ) {
-            // TODO: instead of rendering the whole frame for the wave effect we should render only the area where the effect is active.
-            fheroes2::Blit( fheroes2::CreateDeathWaveEffect( copy, position, deathWaveCurve ), _mainSurface );
-            RedrawPartialFinish();
+            const int32_t wavePositionX = ( area.x + position < 0 ) ? 0 : ( area.x + position );
+            const int32_t restorePositionX = ( wavePositionX < waveStep ) ? 0 : ( wavePositionX - waveStep );
+            const int32_t restoreWidth = wavePositionX < waveStep ? wavePositionX : waveStep;
+
+            const fheroes2::Image spellEffect = fheroes2::CreateDeathWaveEffect( copy, position, deathWaveCurve );
```
ihhub, line 4894: If we declare an image outside the loop and pass it into the function we can save resources for memory management operations.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4862,6 +4862,7 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
     const fheroes2::Sprite & copy = fheroes2::Crop( _mainSurface, area.x, area.y, area.width, area.height );
```
ihhub, line 4862: I just realized that the name of this variable is very tricky. Please rename it to avoid confusion from `copy()` calls for memory management.

ihhub, line 4862: `_mainSurface` should be the same size as `area`. If this is the case then simple call like this would save time:
```cpp
const fheroes2::Image mainSurfaceCopy( _mainSurface );
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4876,18 +4877,43 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
 
     AudioManager::PlaySound( M82::MNRDEATH );
```
ihhub, line 4878: To synchronize the audio with visual effect even better (we have some calculations before we render the next frame) please put this line just before the **while** loop. Of course we're talking here about few CPU ticks but still it's better to avoid them.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4876,18 +4877,43 @@ void Battle::Interface::RedrawActionDeathWaveSpell( const int32_t strength )
...
+            // Place a copy of the original image where the Death Wave effect was on the previous frame.
+            fheroes2::Blit( copy, restorePositionX, area.y, display, renderArea.x, renderArea.y, restoreWidth, renderArea.height );
+
+            // Place the Death Wave effect to its new position.
+            fheroes2::Blit( spellEffect, display, renderArea.x + restoreWidth, renderArea.y );
+
+            // If the battle log was open we redraw it.
+            if ( listlog && listlog->isOpenLog() ) {
+                listlog->Redraw();
+            }
```
ihhub, line 4907: This is actually tricky we draw the whole log but render only a part of it. In this case we expect a frame with the rest of log to be present on screen. We have 2 options here:
- render the whole area of the battle (just the battlefield itself). Slower but no issues.
- in case of open listlog we don't need to copy the whole `_mainSurface` but only a part:
```cpp
fheroes2::Image mainSurfaceCopy;

if ( listlog && listlog->isOpenLog() ) {
    const int32_ width = _mainSurface.width();
    const int32_t height = _mainSurface.height() - listlog->GetArea().height();
    mainSurfaceCopy.resize( width, height );
    fheroes2::Copy( _mainSurface, 0, 0, mainSurfaceCopy, 0, 0, width, height );

    area.height -= listlog->GetArea().height();
}
else {
    mainSurfaceCopy = _mainSurface
}
```
as you can see from the screenshot we don't need to care about the log part:
image

then we don't need to care about log list in the loop.

ihhub (changes requested):
Hi @Districh-ru , I put several comments for further optimization. Could you please take a look when you have time?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi, @ihhub, the `_mainSurface` also contains the bottom panel with buttons (Auto, Skip) and battle status log. So we always crop 36 pixels from the bottom.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: As we always have the status bar and we always have to crop the `_mainSurface` I've only added this:
```cpp
    area.height -= status.height;
    // And if listlog is open, then cut off it too.
    if ( listlog && listlog->isOpenLog() ) {
        area.height -= listlog->GetArea().height;
    }
```

PS
Maybe we have to close the listlog before any human action on the battlefield?
@Branikolog, @zenseii, what do you think? :)

Districh-ru: Yes, @zenseii, I mean auto-hide of the log list.

src/fheroes2/battle/battle_interface.cpp
ihhub: Oh, yes. Could you please create a variable in anonymous namespace at the beginning of this file to store 36 value? Something like `const int32_t statusBarHeight{ 36 };`

ihhub: We don't need to crop but to copy. Crop is slightly slower operation than Copy.

ihhub (approved):

src/fheroes2/battle/battle_interface.cpp
ihhub: I suggest to keep it as it is for now and create an issue or discussion topic to find the best solution.

Districh-ru: I only asked this question here because I ran into this situation here (I rarely open battle log). :)
Indeed, this issue needs to be discussed separately, and if we decide to implement auto-hide of the battle log, then this should be done in a separate PR. I've opened a discussion: #6420

oleg-derevenetz (approved):

ihhub:

Branikolog:
Sorry for being out of project for several days, guys. 
@Districh-ru 
It's a nice improvement, but I'm afraid turning off creature counters makes it impossible to understand, how many troops demolished, when you're focused on the most important stack at the moment. :)
Everything  else is fine, but I'd like to see how counters have changed after spell applied.

Districh-ru:
Hello, @Branikolog!
I've turned the counters off only for the wave effect.
I can revert this change.
But the counters are also turned off for wince animation. After wince animation they are turned on. (The counters are displayed only for `STATIC` and `IDLE` animations).

Branikolog:
@Districh-ru 
In the original game, creature counter updates after wincing and after appearing of counter itself: сreature winces -> counter appears -> number is changed. I suppose it's a separate issue, right? 
But I personally prefer looking at counters during animation, so I can prepare myself for a drastic losses and try to save in my memory each single face of my troop, that faced that terrifying spell. :D

ihhub:
@Branikolog , I think this is beyond the scope of this particular pull request. The changes here were targeted only to improve the performance of the algorithm.

Branikolog:
Hi, @ihhub !
The PR made creature counters hidden which resulted in not clear losses after spell applied. I personally find it a bit uncomfortable to track what creatures I've lost.

Districh-ru:
@ihhub, I also made the change that counter now is not displayed during the render of wave effect.
That's why I asked @Branikolog for review. :)
I can easily revert it in a new PR by changing `SwitchAllUnitsAnimation( Monster_Info::STAND_STILL );` to `SwitchAllUnitsAnimation( Monster_Info::STATIC );`
The counter update algorithm is beyond this PR, but currently in out engine: calculation of units kills (without updating the counter) -> disable counter + render wince animation -> change animation to 'static' with enabling the counters that displays the new value of live creatures.

PR #6418, opened 2022-12-26
Improvements to the French translation

hlabrand:
Hi! I read the whole .po file and found several improvements to be made: typos, inconsistencies, sentences that didn't sound great, sentences that were gendered but shouldn't be (wouldn't work for female heroes for instance), etc.

There is only one change I am not sure of: the translation of "UPGRADE". It used to be "ÉVOLUER", which works in some contexts (like Pokémon, where creatures grow into other ones) but didn't feel right and was inconsistent with other game messages, which used "améliorer" (improve, which is much better). However "AMÉLIORER" is 9 letters, 2 letters longer than the original; from the uppercase I assume this is used in a button, and so is the text going to overflow the button?


files/lang/fr.po
```diff
@@ -477,10 +477,10 @@ msgid "For valor in combat, %{name} receives %{exp} experience."
 msgstr "Pour sa bravoure au combat, %{name} reçoit %{exp} points d'expérience."
 
 msgid "The cowardly %{name} flees from battle."
-msgstr "Le lâche %{name} fuit la bataille."
+msgstr "%{name} fuit la bataille avec lâcheté."
 
 msgid "%{name} surrenders to the enemy, and departs in shame."
-msgstr "%{name} se rend à l'ennemi et part couvert de honte."
+msgstr "%{name} se rend à l'ennemi et part, son nom couvert de honte."
```
felix642, line 483: I don't agree with this change. I feel like the original translation was fine, why did we add "son nom" ?

files/lang/fr.po
```diff
@@ -711,10 +711,10 @@ msgstr ""
 "Affiche le déroulement du combat et indique toutes les actions individuelles."
 
 msgid "Are you sure you want to finish the battle in auto mode?"
-msgstr "Êtes vous sûr de vouloir finir la bataille en mode combat auto ?"
+msgstr "Voulez-vous vraiment finir la bataille en mode combat auto ?"
```
felix642, line 714: I'm assuming that those changes were done to improve consistency? 

We were using "Voulez-vous" at some places and "Êtes-vous sûr" at other ?

files/lang/fr.po
```diff
@@ -981,7 +981,7 @@ msgid ""
 "enemy castles to win."
 msgstr ""
 "Les seigneurs locaux refusent de prêter allégeance à Roland et doivent être\n"
-"soumis. Ils sont riches et puissants, alors préparez-vous à un combat "
+"matés. Ils sont riches et puissants, alors préparez-vous à un combat "
```
felix642, line 984: Not quite sure to know what "maté" is ? 

From the larousse it seems to be a drink, are you sure you're using the right word here ?

files/lang/fr.po
```diff
@@ -1352,7 +1353,7 @@ msgid ""
 msgstr ""
 "Gagnez la faveur des elfes. Ils ne permettront pas aux arbres d'être coupés, "
 "ils vous enverront donc du bois toutes les 2 semaines. Vous devez terminer "
-"votre mission avant le premier jour du septième mois, sinon le royaume va "
+"votre mission avant le premier jour du septième mois, sinon le royaume "
```
felix642, line 1356: You removed a word here, but now the sentence does not make sense anymore "sinon le royaume sûrement s'effondrer"

files/lang/fr.po
```diff
@@ -1521,13 +1522,13 @@ msgid ""
 "The dwarves hail you, \"Any friend of Roland is a friend of ours.  You may "
 "pass.\""
 msgstr ""
-"Les nains vous saluent, \"Tout ami de Roland est un ami à nous.  Vous pouvez "
+"Les nains vous saluent, \"Tous les ami des Roland sont nos amis. Vous pouvez "
```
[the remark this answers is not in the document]
felix642:
```suggestion
"Les nains vous saluent, \"Tous les amis de Roland sont nos amis. Vous pouvez "
```

felix642 (changes requested):
Hi @hlabrand, I've left a few comments. Would you mind taking a look ? :)

files/lang/fr.po
hlabrand: This is because if the hero is a female, the gender doesn't work - "Gwenneth part couvert de honte" doesn't work. If we imply her name is covered in shame, instead, the "son nom" works for both genders and it works. I thought it was an elegant way to avoid the gender issue but what do you think?

files/lang/fr.po
hlabrand: There was a few "voulez-vous" yes, and "êtes-vous sûr de vouloir" is longer and (this is minor but) if the player is not a man that doesn't quite work. I thought it was more elegant to use the compact formulation that works either way.

files/lang/fr.po
hlabrand: https://dictionnaire.reverso.net/francais-synonymes/mater+la+r%C3%A9bellion
It's a verb used specifically in a military context, "squashing the rebellion", "suppressing the uprising" kind of thing. I think it's the preferred word instead of "you must submit them".

files/lang/fr.po
hlabrand: Oh oops totally an oversight, thank you!

hlabrand:
Thank you so much for your comments! :D

files/lang/fr.po
felix642: Oh, that makes sense! Perfect !

files/lang/fr.po
felix642: Sounds good to me !

files/lang/fr.po
felix642: You are right, for some reasons I didn't realize it was a verb and searched "maté" instead of "mater". You can resolve this comment, thank you :)

felix642 (approved):
LGTM !

files/lang/fr.po
```diff
@@ -84,7 +84,7 @@ msgid "DISMISS"
 msgstr "RENVOYER"
 
 msgid "UPGRADE"
-msgstr "EVOLUER"
+msgstr "AMÉLIORER"
```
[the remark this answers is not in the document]
hlabrand: Sorry, is the question addressed to me? Do I need to try to build from source to see how this looks? Not sure I'll have time, maybe we can forego this change for now..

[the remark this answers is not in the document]
hlabrand: (It took me over an hour to install this correctly... Go figure)
After replacing with my mo here is what it looks like
ameliorer

I did not touch "renvoyer" so the issue was already there before... This makes me think the French version wasn't thoroughly tested for such UI artefacts? But i dont know if we want to shove it in this PR; I'm running out of vacation time :)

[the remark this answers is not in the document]
hlabrand: That makes total sense, thanks for explaining! Resolving this now :)

ihhub:
@hlabrand , thank you so much for your contribution to the project!

PR #6419, opened 2022-12-26
[1228] Persisting window position between sessions (SDL2)

VestinVestin:
Works under SDL2, ignores (but doesn't break) SDL1. From what I've seen, it's not worth it to get into the black magic of per-platform hackery necessary to make the SDL1 version work.


ihhub:
Hi @VestinVestin ,

You are trying to include headers which don't belong there. `settings.h` is a part of `fheroes2` code part while `screen.h` is a part of engine. Engine part shouldn't have any implementation related information.

VestinVestin:
> Hi @VestinVestin ,
> 
> You are trying to include headers which don't belong there. `settings.h` is a part of `fheroes2` code part while `screen.h` is a part of engine. Engine part shouldn't have any implementation related information.

Yeah, I knew I was doing something wrong there. I got too excited by getting any solution to work locally.
I've added a callback / hook instead. The only thing I don't like is that it persists the settings every time someone moves the window, but I guess logic like "Save on exit if modified" would belong in another PR.

src/fheroes2/system/settings.h
```diff
@@ -210,6 +211,13 @@ class Settings
         return video_mode;
     }
 
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+    fheroes2::Point WindowPosition() const
+    {
+        return window_position;
+    }
```
ihhub, line 218: Please name this method as per our coding rules: `getWindowPosition()` and it should return const reference.

src/fheroes2/system/settings.h
```diff
@@ -250,6 +258,10 @@ class Settings
     void SetSoundVolume( int v );
     void SetMusicVolume( int v );
 
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+    void SetWindowPosition( fheroes2::Point position );
```
ihhub, line 262: This method should be `inlined`.

src/fheroes2/system/settings.h
```diff
@@ -429,6 +441,11 @@ class Settings
     BitModes _optGlobal;
 
     fheroes2::Size video_mode;
+
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+    fheroes2::Point window_position{ 0, 0 };
```
ihhub, line 446: Please use correct `_camelCase` naming.

ihhub (changes requested):
Hi @VestinVestin , I left several comments in this pull request. Could you please take a look?

VestinVestin:
> Hi @VestinVestin , could you please address given comments in this pull request as well as fix merge conflicts here?

Yeah, sorry, life got in the way ^^". I'll try to sort this out relatively soon.

src/engine/localevent.h
```diff
@@ -195,6 +195,13 @@ class LocalEvent
         _globalKeyDownEventHook = std::move( hook );
     }
 
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
```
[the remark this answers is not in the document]
VestinVestin: Alright, as far as I can tell, I've removed it where it was unnecessary.

src/engine/localevent.h
```diff
@@ -303,6 +308,8 @@ class LocalEvent
         }
     }
 
+    static fheroes2::Point getCenterPosition();
```
ihhub, line 311: I suggest to name this method to be more accurate: `getWindowCenteredPosition()` as it is not obvious what this position for.

src/engine/localevent.h
```diff
@@ -375,6 +382,8 @@ class LocalEvent
     std::function<fheroes2::Rect( const int32_t, const int32_t )> _globalMouseMotionEventHook;
     std::function<void( const fheroes2::Key, const int32_t )> _globalKeyDownEventHook;
 
+    inline static std::function<void( const int, const int )> _onWindowMoved;
```
ihhub, line 385: Could you please explain why we need to use **inline** and **static** on a member of this class if it is a singleton?

src/fheroes2/game/fheroes2.cpp
```diff
@@ -337,6 +338,7 @@ int main( int argc, char ** argv )
         ERROR_LOG( "An unknown exception occurred during application runtime." )
         return EXIT_FAILURE;
     }
-
+    const Settings & conf = Settings::Get();
+    conf.SaveIfNecessary();
```
ihhub, line 342: Please do not put any code outside the exception catching block.

src/fheroes2/system/settings.cpp
```diff
@@ -255,6 +256,10 @@ bool Settings::Read( const std::string & filePath )
         setFullScreen( config.StrParams( "fullscreen" ) == "on" );
     }
 
+    // window position
+    fheroes2::Point center_position = LocalEvent::getCenterPosition();
```
ihhub, line 260: Please use camelCase variable naming, not C style.

ihhub: Is there a reason why we need to declare this variable? It is not used anywhere except one line below.

src/fheroes2/system/settings.h
```diff
@@ -66,6 +66,7 @@ class Settings
 {
 public:
     static constexpr const char * configFileName = "fheroes2.cfg";
+    bool config_modified = false;
```
ihhub, line 69: Do not declare public variables in a class. Put it in a private section and use camelCase naming with an underscore symbol for members.

src/fheroes2/system/settings.h
```diff
@@ -252,6 +259,11 @@ class Settings
     void SetSoundVolume( int v );
     void SetMusicVolume( int v );
 
+    inline void SetWindowPosition( const fheroes2::Point & position )
```
ihhub, line 262: Please remove **inline** keyword. Modern compilers do not need this anymore.

src/fheroes2/system/settings.h
```diff
@@ -441,6 +453,10 @@ class Settings
     // Global game options (GLOBAL_)
     BitModes _optGlobal;
 
+    fheroes2::Size video_mode;
```
ihhub, line 456: This member is not used anywhere.

src/fheroes2/game/game.cpp
```diff
@@ -30,6 +30,8 @@
 #include <utility>
 #include <vector>
 
+#include <SDL_version.h>
```
ihhub, line 33: Please do not put SDL dependency on files within fheroes2 directory. All SDL dependencies must be only **engine** directory.

src/engine/localevent.h
```diff
@@ -195,6 +195,11 @@ class LocalEvent
         _globalKeyDownEventHook = std::move( hook );
     }
 
+    static void setWindowMovedEventHook( std::function<void( const int, const int )> hook )
```
ihhub, line 198: Please make this method non-static as **_onWindowMoved** is going to be non-static as well.

src/fheroes2/game/fheroes2.cpp
ihhub: If we are saving everything just at the time of closing application would be it better to just query the current window position and avoid all these function calls, variable settings and etc.? I think this would be just few lines of code solution.

ihhub (changes requested):
Hi @VestinVestin , I put multiple comments in this pull request. Could you please take a look at them?

ihhub:
Marking this pull request as WIPas the author doesn't respond.

PR #6423, opened 2022-12-28
Fix screen flickering in SDL2

### LINKED ISSUE #6422: Screen flickering on tablet outside viewable area
### Preliminary checks

### Platform

Other

### Describe the bug

There is annoying screen flickering on Android tablet when the game is in 4:3 screen mode (any resolution). 
OS: Android 8.1.0
Device: Lenovo TB-X304L
Native screen resolution: 1280x800
Steps to reproduce:
1. Set game resolution to 640x480 or 800x600. 
2. Tap anywhere on the screen.
Result: everytime when the screen is being tapped there is flickering outside image borders (in the space where normally should be blank screen because image is 4:3 and the screen itself is 16:10).

### Save file

Not necessary - it can be reproduced even in main menu.

### Additional info

https://user-images.githubusercontent.com/10282847/209815240-abc9c24a-740f-40f8-9293-aa9af78422a9.mp4

https://user-images.githubusercontent.com/10282847/209815017-487b8536-12ef-4476-af97-839619668f16.mp4

ihhub:
close #6422

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -981,7 +981,7 @@ namespace
 
             int returnCode = SDL_RenderClear( _renderer );
             if ( returnCode < 0 ) {
-                ERROR_LOG( "Failed to clear render. The error value: " << returnCode << ", description: " << SDL_GetError() )
+                ERROR_LOG( "Failed to clear renderer. The error value: " << returnCode << ", description: " << SDL_GetError() )
                 return;
             }
 
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -1207,7 +1207,7 @@ namespace
 
             int returnCode = SDL_SetRenderDrawColor( _renderer, 0, 0, 0, SDL_ALPHA_OPAQUE );
             if ( returnCode < 0 ) {
-                ERROR_LOG( "Failed to set default color for rendered. The error value: " << returnCode << ", description: " << SDL_GetError() )
+                ERROR_LOG( "Failed to set default color for renderer. The error value: " << returnCode << ", description: " << SDL_GetError() )
             }
 
             returnCode = SDL_SetRenderTarget( _renderer, nullptr );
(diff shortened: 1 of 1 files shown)


src/engine/screen.cpp
```diff
@@ -1205,7 +1205,12 @@ namespace
                 return false;
             }
 
-            int returnCode = SDL_SetRenderTarget( _renderer, nullptr );
+            int returnCode = SDL_SetRenderDrawColor( _renderer, 0, 0, 0, SDL_ALPHA_OPAQUE );
+            if ( returnCode < 0 ) {
+                ERROR_LOG( "Failed to set default color for rendered. The error value: " << returnCode << ", description: " << SDL_GetError() )
```
oleg-derevenetz, line 1210: Hi @ihhub it seems that there is a typo: `default color for rendered` -> `default color for renderer` (?)

ihhub: Yup, it was a typo.

oleg-derevenetz:
Hi @ihhub I noticed that you don't call `SDL_RenderClear()` when you don't update a full frame. Could this be related to this issue?

ihhub:
> Hi @ihhub I noticed that you don't call `SDL_RenderClear()` when you don't update a full frame. Could this be related to this issue?

Yes. `SDL_RenderClear()` should be used for every call. Somehow on other systems is not needed but looks like for Android (only for this specific case) is.

ihhub:
@oleg-derevenetz , during Adventure Map we render a full frame so not exactly related to this.

oleg-derevenetz:
@ihhub I have an impression that flickering occurs when we move the software cursor (which does not require a full frame rendering by itself) AND we don't do the full frame rendering at the same time. Take s look on the second video - there is no flickering when adventure map is scrolled.

oleg-derevenetz (approved):

PR #6428, opened 2022-12-29
Better support of commandline arguments to check_copyright_headers

drevoborod:
Improved support of command line arguments. It allows to simplify future changes and adds auto formatting of usage messages.
Now error message will look like this:

```
usage: check_copyright_headers.py [-h] full_header_file header_template_file source_file [source_file ...]
check_copyright_headers.py: error: the following arguments are required: full_header_file, header_template_file, source_file
```

### DIFF
--- a/script/tools/check_copyright_headers.py
+++ b/script/tools/check_copyright_headers.py
@@ -13,8 +13,8 @@ CURRENT_YEAR = datetime.datetime.now().date().year
 class CustomArgumentParser(argparse.ArgumentParser):
     def error(self, message):
         self.print_usage(sys.stderr)
-        self.exit(1, '%(prog)s: error: %(message)s\n' %
-                  {'prog': self.prog, 'message': message})
+        self.exit(1, "%(prog)s: error: %(message)s\n" %
+                  {"prog": self.prog, "message": message})
 
 
 def parse_arguments():
--- a/script/tools/check_copyright_headers.py
+++ b/script/tools/check_copyright_headers.py
@@ -2,7 +2,6 @@
 
 import argparse
 import datetime
-from gettext import gettext
 import os
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @drevoborod This script in general is not intended to be run by hand, it is internal script for our CI, so "user friendliness" is not an essential requirement for it :)

script/tools/check_copyright_headers.py
```diff
@@ -1,24 +1,40 @@
...
 import subprocess
 import sys
 
 CURRENT_YEAR = datetime.datetime.now().date().year
 
+
+class CustomArgumentParser(argparse.ArgumentParser):
+    def error(self, message):
+        self.print_usage(sys.stderr)
+        self.exit(1, gettext('%(prog)s: error: %(message)s\n') %
```
oleg-derevenetz, line 17: LGTM in general, but `gettext` usage is really redundant here: no one will do localization/internationalization of this purely service script, so I believe that we can remove this dependency and use English-only messages with a calm soul.

script/tools/check_copyright_headers.py
```diff
@@ -1,24 +1,40 @@
...
+                  {'prog': self.prog, 'message': message})
```
oleg-derevenetz, line 18: LGTM in general, but `gettext` usage is really redundant here: no one will do localization/internationalization of this purely service script, so I believe that we can remove this dependency and use English-only messages with a calm soul.

oleg-derevenetz (changes requested):

script/tools/check_copyright_headers.py
drevoborod: > LGTM in general, but `gettext` usage is really redundant here: no one will do localization/internationalization of this purely service script, so I believe that we can remove this dependency and use English-only messages with a calm soul.

I agree. It was just a copy-paste from original .error() method. Ok, I removed it.

drevoborod:
> Hi @drevoborod This script in general is not intended to be run by hand, it is internal script for our CI, so "user friendliness" is not an essential requirement for it :)

I know :) Actually more important part is that I gave human-readable names to all commandline params variables in code. For example, not sys.argv[1] but args.full_header_file. It helps to remember what it is and not to make mistakes in case there will be some changes needed. Also, it's much more flexible solution - for example, one can add not position but named arguments or even optional arguments and so on. But it's up to you, I don't insist :)

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@drevoborod , thank you very much for your contribution to the project!

PR #6432, opened 2022-12-30
Revert the unit counter display while Death Wave or Holy Shout spell effect animation. Battlefield.

Districh-ru:
Reverted the hiding the unit counters before Death Wave or Holy Shout spells animation made in #6413.


ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for this change!

PR #6433, opened 2022-12-30
Battle log text overflowing

### LINKED ISSUE #6357: Long phrases in the battle log
### Preliminary checks

Some phrases do not fit in the width of the combat log. (especially in translations)
long_phrase

It could be reproduced in "Battle only" mode: Attack many master swordsmen with other many master swordsmen.
изображение

_Originally posted by @zenseii in https://github.com/ihhub/fheroes2/issues/6165#issuecomment-1339686868_ :
>This happens in many other languages and I honestly think we should just make it so that these strings wrap to the next line, or make them cut with a [...] part at the end like we do for campaign awards and bonuses.

### Additional info

_No response_

felix642:
When messages in the battle log are too long, we split it in 2 messages using the first space character we see.

closes #6357

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -369,41 +369,10 @@ namespace Battle
             return border.GetRect();
         }
 
-        // Split message at spaces when it's too long to fit in the box
-        static std::vector<std::string> splitMessage( const std::string & message, size_t maxLength )
-        {
-            std::vector<std::string> result;
-            std::string::size_type start = 0;
-            std::string::size_type currentEnd = 0;
-            std::string::size_type nextEnd = 0;
-
-            if ( message.length() <= maxLength ) {
-                result.push_back( message );
-                return result;
-            }
-
-            do {
-                currentEnd = nextEnd;
-                nextEnd = message.find( ' ', currentEnd + 1 );
-                if ( nextEnd == std::string::npos )
-                    nextEnd = message.length();
-
-                if ( nextEnd - start > maxLength ) {
-                    result.push_back( message.substr( start, currentEnd - start ) );
-                    start = currentEnd + 1;
-                }
-            } while ( nextEnd != message.length() );
-
-            if ( start != message.length() )
-                result.push_back( message.substr( start ) );
-
-            return result;
-        }
-
         void AddMessage( const std::string & str )
         {
             const int maxBattleLogSize = 72;
-            std::vector<std::string> splitMessages = splitMessage( str, maxBattleLogSize );
+            std::vector<std::string> splitMessages = stringSplit( str, maxBattleLogSize );
             messages.insert( messages.end(), std::make_move_iterator( splitMessages.begin() ), std::make_move_iterator( splitMessages.end() ) );
             if ( !openlog ) {
                 _scrollbar.hide();
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -368,9 +368,42 @@ namespace Battle
             return border.GetRect();
         }
 
-        void AddMessage( std::string str )
+        // Split message at spaces when it's too long to fit in the box
+        static std::vector<std::string> splitMessage( const std::string & message, size_t maxLength )
         {
-            messages.push_back( std::move( str ) );
+            std::vector<std::string> result;
+            std::string::size_type start = 0;
+            std::string::size_type currentEnd = 0;
+            std::string::size_type nextEnd = 0;
+
+            if ( message.length() <= maxLength ) {
+                result.push_back( message );
+                return result;
+            }
+
+            do {
+                currentEnd = nextEnd;
+                nextEnd = message.find( ' ', currentEnd + 1 );
+                if ( nextEnd == std::string::npos )
+                    nextEnd = message.length();
+
+                if ( nextEnd - start > maxLength ) {
+                    result.push_back( message.substr( start, currentEnd - start ) );
+                    start = currentEnd + 1;
+                }
+            } while ( nextEnd != message.length() );
+
+            if ( start != message.length() )
+                result.push_back( message.substr( start ) );
+
+            return result;
+        }
+
+        void AddMessage( const std::string & str )
+        {
+            const int maxBattleLogSize = 72;
+            std::vector<std::string> splitMessages = splitMessage( str, maxBattleLogSize );
+            messages.insert( messages.end(), std::make_move_iterator( splitMessages.begin() ), std::make_move_iterator( splitMessages.end() ) );
             if ( !openlog ) {
                 _scrollbar.hide();
             }
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -30,6 +30,7 @@
 #include <cstdlib>
 #include <functional>
 #include <initializer_list>
+#include <iterator>
 #include <ostream>
 #include <set>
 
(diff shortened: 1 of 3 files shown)


felix642:
> Hi, @felix642. Could you please address the IWYU check's failure? Either there's either a header that needs to be included or excluded.

Oh, didn't realize a check had failed. Thank you for letting me know. I will fix that as soon as possible and let you know once this PR is ready for review.

felix642:
Hi @zenseii,

Thank you for reminding me. I did notice that the Style check failed and even flagged the email I received from GitHub. But for some reason, I still forgot to fix it. I'll commit my fix later tonight.

Branikolog:
Hi, @zenseii !
I see at least three solutions to the problem with insufficient space for a string in battle log. 
First of all, I believe cutting line into two is not the best solution, as we simply won't see both lines at the same time below the battlefield. Unless we change the rules for showing info on hovering with the mouse over the battlefield (which could be a 4th solution to the problem, actually 😄). For instance, make a pause for 3-5 seconds, before, the first line comes up and we watch the info as it is on your screenshot.
As a solution I propose such things:
- shorten the entire string, so it'd fit info box. Rephrasing or cutting off something not essential. Like showing `Swordsmen do xxx damage` instead of `Master Swordsmen do xxx damage` (I hope player won't be confused with missing the first word, since he clearly understands, what troop he's currently sending into battle ). This is applied only for bottom info box with a separate rule. The entire battle log can show non-shortened string, as it has more space. But this approach could have some problems for other languages. For instance, there're "Цари минотавров" in Russian and cannot just cut off the first word without changing the second one. :)
- Make text line smoothly shifting, similar to advertising boards in the city. This could be a single movement and stop on reaching the end of the line, or looped shifting of the text from left to right.
- As 99% of modern devices have widescreens, we can change the default game width (I mean change the lowest resolution to become wider, from 640x480 to 720x480 or so) so we would be able move "skip" "auto", "settings" buttons somewhere else and expand bottom info box to fit the entire battle screen field. We have an issue with tiny auto&settings buttons for small mobile screens. To make them bigger we have to get some space and move them to the sides. In future, there could appear even some new buttons, and we'd definitely require more space for them, so we can start rearranging game UI from expanding battle info box and moving buttons.

Districh-ru:
Hi @zenseii , @felix642 , @Branikolog ,
We can cut every long string in the battle status bar ("Someone did sonething to another one and ...") and show the full string in the battle log separated into 2 lines if needed.
Or we can make a narrow font for the long phrases in the status bar to fit them.
Or ... we can use creatures icons instead of their names in status bar. :)

Districh-ru:
Hi, @felix642, IMHO currently we can use approach suggested by @zenseii, but split phrase in battle log if it does not fit (the Battle log, not the status bar). And later when we will (and if we will) have special narrow fonts for such cases we can change the display of phrases in status bar.

felix642:
@Districh-ru Sounds good to me.

Branikolog:
Splitting line and showing only the first part of the line could be a solution for now.
Between, I've noticed, that if we point mouse on a place, where no action could be done, we watch "Turn xx" line. Can we show second line (which was previously cut off) in such situations? I think turn number is not such an essential information, and hovering on some inactive places for showing full line is quite predictable behaviour for battle log. How do you think?
crop log

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -368,9 +369,42 @@ namespace Battle
             return border.GetRect();
         }
 
-        void AddMessage( std::string str )
+        // Split message at spaces when it's too long to fit in the box
+        static std::vector<std::string> splitMessage( const std::string & message, size_t maxLength )
```
ihhub, line 373: I suggest to put this function into **tools.h|cpp** files as it belongs there.

ihhub, line 373: `maxLength` can be **const** as we don't modify it within the function.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -368,9 +369,42 @@ namespace Battle
...
+            std::string::size_type nextEnd = 0;
+
+            if ( message.length() <= maxLength ) {
+                result.push_back( message );
+                return result;
+            }
+
+            do {
+                currentEnd = nextEnd;
+                nextEnd = message.find( ' ', currentEnd + 1 );
```
ihhub, line 387: Should we add a third parameter which accept **char** as a delimiter to be more generic of handling things?

src/fheroes2/battle/battle_interface.cpp
ihhub, line 373: Would it be even more efficient to pass `std::vector<string>` as an output parameter to add elements in place instead of creating a temporary array to later move elements from there?

ihhub (changes requested):
Hi @felix642 , I put few comments here. Would you mind to take a look when you have free time?

src/fheroes2/battle/battle_interface.cpp
felix642: Hi @ihhub,

It would in fact be more efficient to add the elements in place rather than copy them from one vector to another. The issue is that the method will have 2 responsibilities (split the string and insert it into another vector) and the code might lose readability. I'm not against the idea, but I think we should find a way to name the method properly and I can't find any good idea right now.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -368,9 +369,11 @@ namespace Battle
             return border.GetRect();
         }
 
-        void AddMessage( std::string str )
+        void AddMessage( const std::string & str )
         {
-            messages.push_back( std::move( str ) );
+            const int maxBattleLogSize = 72;
```
ihhub, line 374: This number really depends on a font and letter sizes in general. For instance:
72 times letter W is not equal 72 times letter i by width. Of course such text won't appear on the screen but this is a bright example of size differences.
I suggest to switch to fheroes2::Text within this class instead of old Text class. fheroes2::Text class has an embedded functionality to divide a string based on maximum width. Such a solution would be much more flexible.

Note: fheroes2::Text class renders text with 2 pixel offset by Y axis so please remember this.

What do you think?

ihhub (changes requested):
Hi @felix642 , I left one suggestion about a different approach to tackle this issue. Could you please take a look at it?

src/fheroes2/battle/battle_interface.cpp
felix642: Hi @ihhub, I did not know such class existed and I think this is a great idea. I'll take a look at the class and how to implement it properly within the battle log. Thank you for the suggestion :)

Districh-ru: Hi @felix642 and @ihhub. `fheroes2::Text` could be used here like (it is a quick example):
```cpp
        void RedrawItem( const std::string & str, int32_t px, int32_t py, bool ) override
        {
            const fheroes2::Text text( str, fheroes2::FontType::normalWhite() );
            const int32_t logWidth{ 600 };
            text.draw( px, py, logWidth, display );
        }
```
But we will have issues when the text is split into two lines, as the game makes a vector `messages` and considers each message as a single line text. It will result in overlapping of the bottom part of split text with the next message string.
So we will have to use something like `text.rows( logWidth);` and put the result in a new vector to keep row info about each message and use this vector in loop that calls `RedrawItem()` in `interface_list.h`. (we could make `RedrawItem()` to return the number of rendered rows).
So we will have to do changes in `ListBox` class (in `interface_list.h`) and these changes should not affect the other places where we use this class. 
And also one notice: currently all battle log massages are left aligned and  `.draw()` makes the text center aligned.

felix642:
Hi @ihhub, I'll do my best to address those comments as soon as possible. I'll mark this PR as a draft while I make the required changes and let you know once they are ready.

ihhub:
Closing this pull request in favor of #7847.

PR #6435, opened 2022-12-31
System::GetCaseInsensitivePath(): remove the addition of an extra "./" for relative paths

### LINKED ISSUE #6434: External music does not work anymore begins from 1.0.0 but works on 0.9.21
### Preliminary checks

### Platform

Linux

### Describe the bug

Good day, bad news.

External music does not work anymore begins from 1.0.0. I have Ubuntu 20.04.5 LTS (Focal Fossa). Build from source:

FOPT="-march=amdfam10 -O2"
cmake -B build -DUSE_SDL_VERSION=SDL2 -DENABLE_IMAGE=ON -DCMAKE_CXX_FLAGS="$FOPT"
cmake --build build

but in latest snapshot the same.

If I roll back git to 0.9.21 and build then external music works fine. Music in ./music/homm2_XX.ogg.
In 1.0.0 always play midi even if set external music.

In output I have:
[ERROR]	playSound:  Failed to create an audio chunk from memory. The error: Unsupported block alignment
but this error occurs in both 1.0.0 and 0.9.21. Only in 1.0.0 no external music.

Thanks.

### Save file

Save file not needed

### Additional info

_No response_

oleg-derevenetz:
fix #6434

Previous implementation always added `./` to the beginning of relative paths, even if they already begin with `./`, which was TECHICALLY correct (because `MUSIC`, `./MUSIC` and `././MUSIC` are in fact the same paths), but the logic of `findMusicFile()` in the `audio_manager.cpp` compares paths literally, so it was not able to find the music file.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -390,8 +390,9 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
 {
     correctedPath.clear();
 
-    if ( path.empty() )
+    if ( path.empty() ) {
         return false;
+    }
 
     DIR * d;
     bool last = false;
@@ -399,10 +400,11 @@ bool System::GetCaseInsensitivePath( const std::string & path, std::string & cor
     const char * delimiter = "/";
 
     if ( path[0] == delimiter[0] ) {
+        correctedPath.append( delimiter );
+
         d = opendir( delimiter );
     }
     else {
-        correctedPath = curDir[0];
         d = opendir( curDir );
     }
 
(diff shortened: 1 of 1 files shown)


asm64:
> fix #6434
> 
> Previous implementation always added `./` to the beginning of relative paths, even if they already begin with `./`, which was TECHICALLY correct (because `MUSIC`, `./MUSIC` and `././MUSIC` are in fact the same paths), but the logic of `findMusicFile()` in the `audio_manager.cpp` compares paths literally, so it was not able to find the music file.

Thank you very much. Happy New Year!

ihhub (approved):

ihhub:

Issue #6415, opened 2022-12-25
Maps creating collaboration


viT-1:
### Your question

In [one](https://youtu.be/NFy380A-6OE?t=7984) of youtube streams I had dialog about new maps with your team member.
It's a pitty that for licensing reasons, there is no possibility to add new fan map to your release pack,
because they was created by original Map Editor.
I should recreate my map with fheroes map editor?
Are there any recommendations regarding this? Cloud storing, copyrighting, etc...

oleg-derevenetz:
To be honest, I'm not sure why distribution of new maps created by 3rd parties using the original map editor should violate the 3DO or GoG license. This is separate work and all rights belong to the author of the map, these maps doesn't include any copyrighted material (such as images from the original game, for example). So if any 3rd party person will create a new map and release it with [appropriate GPL-compatible license](https://www.gnu.org/licenses/license-list.en.html#OtherLicenses) (such as CC BY) we should be able to distribute it.

ihhub:
GoG release EULA EULA UBISOFT UK--Jan09.doc file:
```
It is not permitted:
-    To make copies of the Multimedia Product,
-    To operate the Multimedia Product commercially,
-    To use it contrary to morality or the laws in force,
-    To modify the Multimedia Product or create any derived work,
-    To transmit the Multimedia Product via a telephone network or any other electronic means, except during multi-player games on authorised networks,
-    To create or distribute unauthorised levels and/or scenarios,
-    To decompile, reverse engineer or disassemble the Multimedia Product.
```

oleg-derevenetz:
> To create or distribute unauthorised levels and/or scenarios

So basically you can't use the original map editor at all because by doing this you create "unathorized scenarios" lol. This is kind of absurd.

Laserlicht:
I can't really imagine that the EULA applies here either.

In the files from the CD (dos/windows) I can not find anything. I think a EULA from 2009 can not be applied to backdated releases. If you create the MAPs with the original release, then it's okay, but not with the GOG version?

It's almost like creating/editing images in Photoshop, but then you're not allowed to distribute them...

In fact, the EULA states that you are not allowed to use the map editor. Sounds to me like a copied standard EULA...

ihhub:
I am moving this topic into the discussion section as it is definitely not related to the engine.

PR #6437, opened 2022-12-31
Prepare code for the future Editor

ihhub:
This is a preparation step towards building the future Editor. Right now the code related to object loading and handling on Adventure Map is a complete mess. This pull request is targeted to cleanup the code and make it more or less readable. The list of changes:
- rename `uniq, level, object and index` of **TilesAddon** structure and `uniq, objectTileset, objectIndex and _level` of class **Tiles** into `_uid, _layerType, _objectType and _imageIndex` to have the same member names as these variables are exactly the same for both classes. In the future class **Tiles** will have only one stack of objects.
- add **ObjectIcnType** enumeration and replace many hardcoded values by entries from the enumeration
- add multiple TODO comments for the future changes (I wanted to keep this pull request as small as possible)

This pull request touches the underlying logic for Adventure Map objects. While it does not change the logic at all there is a chance that these changes might break anything for loaded from save files maps or new maps. I tested multiple maps and everything look correct.

relates to #4330 and #6415

### DIFF
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -66,43 +66,6 @@ namespace Maps
         TERRAIN_LAYER = 3 // roads, water flaws and cracks. Essentially everything what is a part of terrain.
     };
 
-    struct TileObjectPartInfo
-    {
-        TileObjectPartInfo() = default;
-
-        TileObjectPartInfo( const uint32_t uid_, const uint8_t layerType_, const uint8_t objectType_, const uint8_t imageIndex_, const uint8_t extraInfo_ )
-            : uid( uid_ )
-            , layerType( layerType_ )
-            , objectType( objectType_ )
-            , imageIndex( imageIndex_ )
-            , extraInfo( extraInfo_ )
-        {
-            // Do nothing.
-        }
-
-        TileObjectPartInfo( const TileObjectPartInfo & ) = default;
-
-        ~TileObjectPartInfo() = default;
-
-        TileObjectPartInfo & operator=( const TileObjectPartInfo & ) = delete;
-
-        // Unique identifier of an object. UID can be shared among multiple object parts if an object is bigger than 1 tile.
-        uint32_t uid{ 0 };
-
-        // Layer type shows how the object is rendered on Adventure Map. See ObjectLayerType enumeration.
-        uint8_t layerType{ OBJECT_LAYER };
-
-        // The type of object which correlates to ICN id. See MP2::GetICNObject() function for mor details.
-        uint8_t objectType{ 0 };
-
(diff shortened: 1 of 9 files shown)


ihhub:
Hi @oleg-derevenetz , I deactivated ["std::byte" should be used when you need byte-oriented memory access](https://sonarcloud.io/organizations/ihhub/rules?open=cpp%3AS6022&rule_key=cpp%3AS6022) rule in SonarQube because these warnings make no sense on our codebase.

src/fheroes2/maps/maps_tiles.h
```diff
@@ -98,10 +98,18 @@ namespace Maps
...
 
-        uint32_t uniq;
-        uint8_t level;
-        uint8_t object;
-        uint8_t index;
+        // Unique identifier of an object. UID can be shared among multiple object parts if an object is bigger than 1 tile.
+        uint32_t _uid;
+
+        // Layer type shows how the object is rendered on Adventure Map. See ObjectLayerType enumeration.
+        uint8_t _layerType;
```
oleg-derevenetz, line 105: Is it possible to declare the `ObjectLayerType` as `enum class` and change the type of this field to `ObjectLayerType` instead of `uint8_t` here and in the `Tiles` class below? Maybe in future?

src/fheroes2/maps/maps_tiles.h
```diff
@@ -98,10 +98,18 @@ namespace Maps
...
+
+        // The type of object which correlates to ICN id. See MP2::GetICNObject() function for more details.
+        // First 2 bits are used for marking a tile as a road and indicating that the object has animation.
```
oleg-derevenetz, line 108: Is it possible to elaborate here, which bit is responsible for what exactly - here and in the `Tiles` class below?

oleg-derevenetz (commented):
Hi @ihhub I just left two optional suggestions, please take a look when possible.

src/fheroes2/maps/maps_tiles.h
ihhub: Hi @oleg-derevenetz , I will update it in the future. The value is actually obtained from **quantity1** member of `mp2tile_t` structure but according to
```cpp
uint8_t quantity1; // Bitfield, first 3 bits are flags, rest is used as quantity
```
there are 3 bits which are in use and I have no idea what the third bit for (as of now).

src/fheroes2/maps/maps_tiles.h
ihhub: It is done. The code is so convoluted :(

Districh-ru (approved):
I've tested maps from original CD and didn't notice problems at first glance.

oleg-derevenetz (approved):

PR #6438, opened 2022-12-31
[WIP] Integer display scale factor

a1exsh:
Hi!

As announced in #5016, here's a draft for supporting integer factor resource scaling.  E.g., `1920x1080` can fit up to 3 OG screen by width, but only 2 by height, so the factor is 2 in this case.  A resolution of `2560x1440` is `4x:3x` of the OG's `640x480`, so a useful scale factor is either 2 or 3 (for now the maximum is chosen automatically, but we may let user choose the factor freely as well).  Higher resolutions with lower-than-maximum scale factors are likely the most useful combinations.

The change is focused on the rendering engine and tries to avoid touching the application code as much as possible — the vast majority of pixel offsets are hard-coded values in the game code, obtained by measuring the original game's resources, which would be infeasible to touch. ;)

The way this was made possible is by associating an integer "scale factor" value to each `Image` instance (including `Display`) and making the images "lie" about their dimensions: regardless of the scale factor, an image instance will always report the same `width` and `height` — its "logical dimensions".  The actual physical dimensions are only used by the image manipulation and rendering routines, for the most part contained in a single file `image.cpp`.

The resources are loaded with the default scale factor of `1` and on-demand can be resized and cached for higher scale factors (up to `4`).  Resizing is done using our existing nearest neighbour algorithm, but this is where other algorithms may be plugged in later, if desired.

Once this is given, the next step is to always create `Image` instances with the display's current scale factor, and, because the display is also reporting its `width`/`height` in the logical units, the calculations like centering a box on the screen continue to work "magically".

---
Edit: added example.

Consider the code for rendering a window of arbitrary size in the middle of the screen:
https://github.com/ihhub/fheroes2/blob/master/src/fheroes2/gui/ui_window.cpp#L33-L36

```c++
    StandardWindow::StandardWindow( const int32_t width, const int32_t height, Image & output )
        : _output( output )
        , _activeArea( ( output.width() - width ) / 2, ( output.height() - height ) / 2, width, height )
        , _windowArea( _activeArea.x - borderSize, _activeArea.y - borderSize, _activeArea.width + 2 * borderSize, _activeArea.height + 2 * borderSize )
...
```

It takes the dimensions of the inner active area `width`×`height` and adds a border of the hard-coded size of 16 logical pixels around it in all directions.  Let's work out the coordinates for rendering such a window border on the display at resolution `1920x1080` with scale factor 2, for an inner area of `640x480` logical pixels.

As this is application code (not the rendering engine), my priority was to leave it untouched.  The `width()` and `height()` methods used here return the logical dimensions, so for our display these values are `1920/2 = 960` and `1080/2 = 540`.  The `width`  and `height` parameters may either be hard-coded in the caller, or they may come from measuring the dimensions of a game resource, such as result of calling `fheroes2::AGG::GetICN( ICN::SWAPWIN, 0 )`.  In the original resources, this the window displayed when two heroes of the same player meet each other for exchange of armies and artifacts.  It has the dimensions of `640x480` pixels.  Because in this example our display's current scale factor is 2, the `Image` instance returned by the `GetICN()` function is pre-scaled 2×2, so its physical dimensions are actually `1280x960`.  The logical dimensions, visible to the application code, are the same _regardless_ of the current scale factor.

So the `activeArea` works out, in logical pixels, to `{ (960 - 640)/2 = 160, (540 - 480)/2 = 30, 640, 480 }`. The `broderSize` is hard-coded as 16, so  `windowArea` works out, again in logical pixels, to `{ 160 - 16 = 144, 30 - 16 = 14, 640 + 2*16 = 672, 480 + 2*16 = 512}`.

When the window border image is rendered on the display, the logical pixels are translated to physical ones, by simply multiplying them by the scale factor.  This, again, happens in the rendering engine's functions such as `Blit()`, so the application code doesn't need any changes for rendering to work correctly.  The border area becomes, in physical pixels: `{ 144*2 = 288, 14*2 = 28, 1344, 512*2 = 1024 }`.  The inner active window area — `{ 160*2 = 320, 30*2 = 60, 640*2 = 1280, 480*2 = 960 }`.

image

---
Finally, for correct controller operation, the mouse position should be also tracked taking the current scale factor into account.

The rendering code is always working with static pre-scaled resources (there is no real-time upscaling and it's not needed).

As I've mentioned in the original discussion, I observe no performance impact of the adjusted rendering procedure. :mx_claus: :champagne: 

---
Edit: some notes on the implementation.  It's not without hacks, some of which are quite ugly.

This has mostly to do with the fact that `Display::instance()` is used as default parameter in a couple of places, such as default target for rendering buttons, etc.  It would be better to add an abstraction layer of a "rendering context", so that direct rendering on the display is prohibited.  That would also help to reduce code duplication where coordinates of the top-left corner are constantly need to be added manually to each X,Y screen offsets.

Not directly related to this, I think it may make sense to eventually rework the UI controls system to properly encapsulate the state/rendering and event handling.  Right now it's all over the place and is simply copied over from one dialog code to the other, with slight changes.  I don't know what are you exact plans for the Editor, but it feels like this is where even more dialogs and controls will be introduced. ;)

---
TODO:
- [ ] Make changing resolutions / scale factor 100% safe (there are some corner cases right now)
- [x] Figure out the crash in the multi-language font generation
- [ ] Check how it works (or doesn't) on non-default build targets
- [ ] Explore advanced upscaling filters

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -551,7 +551,6 @@ namespace fheroes2
 
         if ( !image._data ) {
             clear();
-
             return;
         }
 
@@ -1131,7 +1130,6 @@ namespace fheroes2
 
     void Copy( const Image & in, Image & out )
     {
-        out.resize( in.width(), in.height() );
         Copy( in, 0, 0, out, 0, 0, in.width(), in.height() );
     }
 
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1203,16 +1203,17 @@ namespace fheroes2
 
     Sprite CreateContour( const Image & image, uint8_t value )
     {
-        const int32_t width = image._w();
-        const int32_t height = image._h();
-
-        Sprite contour( width, height );
+        Image contour( image.width(), image.height(), image.scaleFactor() );
         contour.reset();
-        if ( width < 2 || height < 2 ) {
+
+        if ( contour.width() < 2 || contour.height() < 2 ) {
             return contour;
         }
 
         assert( !contour.empty() );
+
+        const int32_t width = image._w();
+        const int32_t height = image._h();
 
         const uint8_t * inY = image.transform();
         uint8_t * outImageY = contour.image();
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -435,7 +435,7 @@ namespace fheroes2
     }
 
     Image::Image( int32_t width_, int32_t height_ )
-        : Image( width_, height_, Display::scaleFactor() )
+        : Image( width_, height_, Display::currentScaleFactor() )
     {
     }
 
@@ -452,7 +452,7 @@ namespace fheroes2
         : _width( 0 )
         , _height( 0 )
         , _scaleFactor( 1 )
-        , _singleLayer( false )
+        , _singleLayer( image_._singleLayer )
     {
         copyFrom( image_ );
     }
@@ -510,7 +510,7 @@ namespace fheroes2
 
         _width = 0;
         _height = 0;
-        _scaleFactor = Display::scaleFactor();
+        _scaleFactor = Display::currentScaleFactor();
     }
 
     void Image::fill( uint8_t value )
@@ -529,21 +529,21 @@ namespace fheroes2
 
     void Image::_resize( int32_t physicalWidth_, int32_t physicalHeight_, int32_t scaleFactor_ )
     {
-        if ( _width == physicalWidth_ && _height == physicalHeight_ && _scaleFactor == scaleFactor_ ) {
-            return;
-        }
+        _scaleFactor = scaleFactor_;
 
         if ( physicalWidth_ <= 0 || physicalHeight_ <= 0 ) {
             clear();
             return;
         }
 
+        // don't reset if only changing scale factor, for example
+        if ( _width != physicalWidth_ || _height != physicalHeight_ ) {
+            const size_t totalBytes = static_cast<size_t>( 2 * physicalWidth_ * physicalHeight_ );
+            _data.reset( new uint8_t[totalBytes] );
+        }
+
         _width = physicalWidth_;
         _height = physicalHeight_;
-        _scaleFactor = scaleFactor_;
-
-        const size_t totalBytes = static_cast<size_t>( 2 * _width * _height );
-        _data.reset( new uint8_t[totalBytes] );
     }
 
     void Image::reset()
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -389,8 +389,8 @@ namespace
             return;
         }
 
-        const int32_t widthIn = in.width();
-        const int32_t widthOut = out.width();
+        const int32_t widthIn = in._w();
+        const int32_t widthOut = out._w();
 
         const uint8_t * imageInY = in.image() + inY * widthIn + inX;
         uint8_t * imageOutY = out.image() + outY * widthOut + outX;
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -251,11 +251,16 @@ namespace
             height -= offsetY;
         }
 
+        x *= image.scaleFactor();
+        y *= image.scaleFactor();
+        width *= image.scaleFactor();
+        height *= image.scaleFactor();
+
         return true;
     }
 
     bool Verify( int32_t & inX, int32_t & inY, int32_t & outX, int32_t & outY, int32_t & width, int32_t & height, int32_t widthIn, int32_t heightIn, int32_t widthOut,
-                 int32_t heightOut )
+                 int32_t heightOut, int32_t scaleFactor )
     {
         if ( widthIn <= 0 || heightIn <= 0 || widthOut <= 0 || heightOut <= 0 || width <= 0 || height <= 0 ) // what's the reason to work with empty images?
             return false;
@@ -314,13 +319,21 @@ namespace
             height -= offsetY;
         }
 
+        inX *= scaleFactor;
+        inY *= scaleFactor;
+        outX *= scaleFactor;
+        outY *= scaleFactor;
+        width *= scaleFactor;
+        height *= scaleFactor;
+
         return true;
     }
 
     bool Verify( const fheroes2::Image & in, int32_t & inX, int32_t & inY, const fheroes2::Image & out, int32_t & outX, int32_t & outY, int32_t & width,
                  int32_t & height )
     {
-        return Verify( inX, inY, outX, outY, width, height, in.width(), in.height(), out.width(), out.height() );
+        assert( in.scaleFactor() == out.scaleFactor() );
+        return Verify( inX, inY, outX, outY, width, height, in.width(), in.height(), out.width(), out.height(), in.scaleFactor() );
     }
 
     uint8_t GetPALColorId( uint8_t red, uint8_t green, uint8_t blue )
@@ -427,27 +440,27 @@ namespace fheroes2
     }
 
     Image::Image( int32_t width_, int32_t height_, int32_t scaleFactor_ )
-        : _scaleFactor( scaleFactor_ )
-        , _width( 0 )
+        : _width( 0 )
         , _height( 0 )
+        , _scaleFactor( scaleFactor_ )
         , _singleLayer( false )
     {
         Image::resize( width_, height_ );
     }
 
     Image::Image( const Image & image_ )
-        : _scaleFactor( 1 )
-        , _width( 0 )
+        : _width( 0 )
         , _height( 0 )
+        , _scaleFactor( 1 )
         , _singleLayer( false )
     {
         copyFrom( image_ );
     }
 
     Image::Image( Image && image_ ) noexcept
-        : _scaleFactor( 1 )
-        , _width( 0 )
+        : _width( 0 )
         , _height( 0 )
+        , _scaleFactor( 1 )
         , _data( std::move( image_._data ) )
         , _singleLayer( false )
     {
@@ -622,8 +635,8 @@ namespace fheroes2
 
     void Sprite::setPosition( int32_t x_, int32_t y_ )
     {
-        _x = x_ * _scaleFactor;
-        _y = y_ * _scaleFactor;
+        _x = x_ * scaleFactor();
+        _y = y_ * scaleFactor();
     }
 
     ImageRestorer::ImageRestorer( Image & image_ )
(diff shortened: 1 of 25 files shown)


ihhub:
Hi @a1exsh , wouldn't it cleaner to just modify screen.cpp file to set effective for the rendering engine and "scaled" for SDL resolutions there? SDL 2 will handle all the stuff with cursor position.

a1exsh:
> Hi @a1exsh , wouldn't it cleaner to just modify screen.cpp file to set effective for the rendering engine and "scaled" for SDL resolutions there? SDL 2 will handle all the stuff with cursor position.

@ihhub that may well be the easiest way, considering just the nearest neighbour upscaling.  Maybe even more advanced filter like xBRz can be plugged in directly before rendering on the screen with tolerable performance impact.

My end goal here is different, however, as I want to see how externally upscaled resources can be put into the game, as I've mentioned in the linked discussion: https://github.com/ihhub/fheroes2/discussions/5016#discussioncomment-4554424

Here's a preview of what kind of difference it makes:
Screenshot from 2023-01-03 08-47-12

Screenshot from 2023-01-03 08-46-38

ihhub:
Hi @a1exsh , I'm afraid you have to change a LOT of places related to rendering and logic if you want to change every single sprite in the game. We have many hardcoded and semi-hardcoded places in the code which are bound to existing graphics and it's not easy to change them in one shot. Would the result he worth these changes which I estimate might take months to adjust?

a1exsh:
> Hi @a1exsh , I'm afraid you have to change a LOT of places related to rendering and logic if you want to change every single sprite in the game. We have many hardcoded and semi-hardcoded places in the code which are bound to existing graphics and it's not easy to change them in one shot. Would the result he worth these changes which I estimate might take months to adjust?

@ihhub but this is exactly what I managed to avoid with this patch — not a single hard-coded resource dimension was changed, and it still works!  I can record a longer demo later today, or, better yet, you can now just run the code from this branch to see it for yourself. :)

It might be hard to believe, but please check the updated description for this PR where I added an explanation on how/why this hack works.

ihhub:
@a1exsh , please try with the most obvious sprite: Air Elemental and see if you technique works.

a1exsh:
> @a1exsh , please try with the most obvious sprite: Air Elemental and see if you technique works.
>  also please make sure that cycling animation still works with your changes.

@ihhub sure, here you are:

https://user-images.githubusercontent.com/489601/210412585-9889148a-132e-4434-b270-3460d92f27ef.mp4

The reason it works is because I'm using the `Resize()` function (without sub-pixel precision) from the rendering engine, which is aware of transforms and works with color indexes as opposed to RGB values.

Making this work with xBRz maybe a bit challenging, but interesting at that.  Here I count on your help with explaining how advanced transforms of the Air Elemental actually work. ;)

ihhub:
Hi @a1exsh , Air Elemental is one of examples where any upscaling algorithm won't work unless you write a special code to handle this like we partially do in `fheroes::Resize()`. Therefore I highlighted that the only valid way to do this is to apply a scaling algorithm just before rendering a frame. For example, it is visible the Main Menu screenshot has some artefacts like a smooth trail above the door on the left side of the image. This is because multiple images upscaled separately which is wrong for such frame.

Transform layer is a collection of functions (up to 255), each of them uses its own LUT table. There is no such thing as alpha channel for images, it is replaced by LUT tables. The original game uses up to 15 different LUT tables. The Air Elemental has as far as I remember 5-7 LUT table usages. There is no formula how a set of LUT table IDs can generate a new table, maximum what we can do is approximation.

The resulted image in Display class does not have transform layer so it is absolutely valid to apply any scaling algorithms there.

a1exsh:
> Hi @a1exsh , Air Elemental is one of examples where any upscaling algorithm won't work unless you write a special code to handle this like we partially do in `fheroes::Resize()`.

My overall plan includes writing such special code.

> Therefore I highlighted that the only valid way to do this is to apply a scaling algorithm just before rendering a frame. 

That would be the easiest and cleanest way to do it, indeed, assuming the algorithm doesn't hurt performance too much.

> For example, it is visible the Main Menu screenshot has some artefacts like a smooth trail above the door on the left side of the image. This is because multiple images upscaled separately which is wrong for such frame.

Please don't look at that door on the screenshot — the lantern animation frames are not upscaled, as I haven't automated upscaling of 100% game resources just yet. ;)

> Transform layer is a collection of functions (up to 255), each of them uses its own LUT table. There is no such thing as alpha channel for images, it is replaced by LUT tables. The original game uses up to 15 different LUT tables. The Air Elemental has as far as I remember 5-7 LUT table usages. There is no formula how a set of LUT table IDs can generate a new table, maximum what we can do is approximation.

Thanks, that confirms my understanding.  Given these transforms are reversible for some background color X, it should be possible to still upscale in an elaborate manner.  I will give it a try, even if just for fun. :)

ihhub:
Hi @a1exsh , I would highly recommend to implement on the level of screen.cpp to immediately see the results and adopt the algorithm per needs. Right now doing everything manually is no way approach to make things in any reasonable time.

a1exsh:
As promised, I've tried with upscaling all the game resources using ESRGAN+ model, and here is the result (some sensitive parts are still upscaled using nearest neighbour, e.g. fonts and hero mini-portraits):

https://user-images.githubusercontent.com/489601/211197705-8778bc84-2fda-403d-864b-361edce93055.mp4

Monster animations in the well:
https://user-images.githubusercontent.com/489601/211197719-87038c7e-6dde-4748-bb65-e6826c9e770c.mp4

The game is almost fully functional with that and it even seems that upscaling may be done at load time, given a powerful enough videocard.

ihhub:
Hi @a1exsh , is it correct that you upscale the resources of the game and then load them by the engine? If yes this is not possible to use this approach for users as it's a direct violation of licensing. Nobody has rights to distribute game's resources, even modified, without Ubisoft approval.

a1exsh:
> Hi @a1exsh , is it correct that you upscale the resources of the game and then load them by the engine? If yes this is not possible to use this approach for users as it's a direct violation of licensing. Nobody has rights to distribute game's resources, even modified, without Ubisoft approval.

@ihhub yes, for this demo I'm currently upscaling the resources externally and then load them into game, as they are ready to use.

Of course, I've thought about the licensing problem and I'm not suggesting to distribute the resources in any form.  My original plan involves providing a script for the users to run on their machine that will do the upscaling (once) by extracting the resources present on the user's machine.

After a bit more testing I can see that potentially this might not even be needed, as a modern video card can process a lot of images very fast, so it might be possible to upscale resources on-the-fly, without extra preparatory steps.  I want to do a benchmark on that to be sure.

Do you see the potential of this approach now?  I need to solve the Air Elemental and cycling (and the animations, e.g. see the Skull Pile which is not looking that great), but do you like how it looks (or do you think users may like it)?  I can tell you that on a big screen (3840x2160) the difference is striking.

Laserlicht:
@a1exsh 

Most emulators and games that are enhanced with algorithms like hqx use live upscaling. Shouldn't be that much of a problem.

I think both variants have their advantages.

### Spontaneously I can think of
**Textures from the hard disk:**
- faster
- more complex scaling algorithms possible 
- desired algorithm does not necessarily have to be implemented in fheroes2
- possibility for own "texture packs" (e.g. new interface or free resources)

**Live Upscaling:**
- much easier to handle for the user
- no extra files on the hard disk
- easier on non pc devices

And such an upscaling also makes sense on a FullHD or 1440p monitor. Not to mention all the HiDPI devices.

I would be very happy if we could find a good solution. It is visually for many people (not everyone likes the blurry/pixel optics) a good improvement on new devices.

a1exsh:
> @a1exsh
> 
> Most emulators and games that are enhanced with algorithms like hqx use live upscaling. Shouldn't be that much of a problem.
> 
> I think both variants have their advantages.
> ### Spontaneously I can think of
> 
> **Textures from the hard disk:**
> 
>     * faster
> 
>     * more complex scaling algorithms possible
> 
>     * desired algorithm does not necessarily have to be implemented in fheroes2
> 
>     * possibility for own "texture packs" (e.g. new interface or free resources)

While technically possible and maybe exciting, that's still problematic from the legal point of view for anyone who decides to distribute such texture packs, so after thinking a bit longer, I believe we should not enable or encourage that.  So live upscaling it has to be.

> **Live Upscaling:**
> 
>     * much easier to handle for the user
> 
>     * no extra files on the hard disk
> 
>     * easier on non pc devices

> Here I have upscaled fheroes2 (video) with xbr4. Even the fonts are okay, I think. But not comparable to approaches based on neural networks.

Nice!

Laserlicht:
> While technically possible and maybe exciting, that's still problematic from the legal point of view for anyone who decides to distribute such texture packs, so after thinking a bit longer, I believe we should not enable or encourage that. So live upscaling it has to be.

I didn't mean to offer modified original textures as a texture pack. 

I was thinking more about something like this: https://www.youtube.com/watch?v=BMz2CFtGuEw
An customization of the existing design. A design theme.

And even if original textures are used. The legal risk then clearly lies with the provider of the texture package.

Theoretically it would be also possible to replace all original textures to create a really free fheroes. Even if I don't think it's that likely that a designer will be found who can invest so much time to create everything from scratch. And that it then also remains close to the original.

ihhub:
Hi @a1exsh , could you please tell the status of this pull request? Are we stuck here and need another approach or we are going to continue within the same direction?

a1exsh:
> Are we stuck here and need another approach or we are going to continue within the same direction?

@ihhub for now I am stuck on the color cycling problem.  The upscaling algorithms I have tried work in RGB colorspace, which is understandable, but the cycling only works with the indexed colors and a palette.  In the current approach, the cycling information is lost already when saving an original image resource into a PNG file for upscaling later.  When loading an upscaled image, the reverse mapping from RGB to indexes is done and some more information is getting lost.

It may be still possible, by marking the regions of the original image where cycling should be applied, to map the corresponding regions of an upscaled image to cycled indexs.  I haven't tried that yet, mainly because some of the images are severely affected by mapping to the game palette in any case.  E.g. look at the rainbow of the sorceress town:

Upscaled 2x, RGB:
image

In-game 1920x1080 2x, game palette:
image

That is a pity, because the rest of the buildings in most of the towns are actually quite okay.  Making that rainbow look nice probably means switching the in-game representation of images to RGB, but that's a bigger change than I hoped for.  And it still doesn't help to solve the cycling problem.

What else may be worth noticing, for the future work:
- smaller images with sensitive details (such as fonts or hero mini-portraits) are easily destroyed by NN upscaling and should rather be done using xBRz or nearest neighbour
- animated sequences should be extracted differently, so that upscaling is done on the complete frame vs. partial changed parts
- adventure map tiles should be left alone, as they are extremely sensitive for any misplaced pixels on the edges — not even xBRz did an acceptable job of upscaling those

For now I stopped working on this, though I think it was worth exploring what is possible.  I may return to it one day, if I find any new way around the cycling issue. :)

ihhub:
Hi @a1exsh , I am closing this pull request as the work on this feature has stopped due to multiple reasons. Please feel to reopen the pull request once you are able to address all the issues.

PR #6439, opened 2022-12-31
Changelog in appdata.xml

Laserlicht:
Script for parsing the `changelog.txt` into `*.appdata.xml`. Copies the whole changlog (except the oldest entry) in the appdata.xml.

Appdata can be updated after changelog update by executing `python3 script/tools/changelog_update_appdata.py changelog.txt script/packaging/common/fheroes2.appdata.xml`.

@ihhub Do you think that's a good idea, or should we rather discard it?

Furthermore I have also added the donate link.

Changelogs looks like this in the software center (sorry for german):
grafik

### DIFF
--- a/script/tools/changelog_update_appdata.py
+++ b/script/tools/changelog_update_appdata.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 
 import argparse
-import re
 import datetime
+import re
+import sys
 from xml.sax.saxutils import escape
 
 #
--- a/script/tools/changelog_update_appdata.py
+++ b/script/tools/changelog_update_appdata.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+
+import argparse
+import re
+import datetime
+import html
+
+#
+# Appstream file could be validated with 'appstream-util validate io.github.ihhub.Fheroes2.appdata.xml'
+#
+
+class CustomArgumentParser(argparse.ArgumentParser):
+    def error(self, message):
+        self.print_usage(sys.stderr)
+        self.exit(1, "%(prog)s: error: %(message)s\n" %
+                  {"prog": self.prog, "message": message})
+
+def parse_arguments():
+    parser = CustomArgumentParser()
+    parser.add_argument("changelog_file", help="changelog.txt file")
+    parser.add_argument("appdata_file", help="*.appdata.xml file")
+    return parser.parse_args()
+
+def main():
+    args = parse_arguments()
+    
+    changelog = open(args.changelog_file, encoding='utf-8').read()
+    appdata = open(args.appdata_file, encoding='utf-8').read()
+    
+    tpl = '''
+    <release date="{}" version="v{}">
+      <url>https://github.com/ihhub/fheroes2/releases/tag/{}</url>
+      <description>
+        <p>Changes in v{} ({}):</p>
+        <ul>
+{}
+        </ul>
+      </description>
+    </release>\n'''[1:]
+    tmp = ''
+    
+    regex = r"version ([\d.]+) \(([\w ]+)\)\n(.*?)[\n]{2}"
+    for match in re.findall(regex, changelog, re.MULTILINE | re.DOTALL):
+        tmp += tpl.format(
+            datetime.datetime.strptime(match[1], '%d %B %Y').strftime('%Y-%m-%d'),
+            match[0],
+            match[0],
+            match[0],
+            match[1],
+            re.sub(r'- (.*)', r'          <li>\1</li>', html.escape(match[2]))
+        )
+        
+    new_appdata = re.sub(r'<releases>(.*?)</releases>', r'<releases>\n' + tmp + '  </releases>',
+        appdata,
+        flags=re.MULTILINE | re.DOTALL
+    )
+    
+    open(args.appdata_file, 'w', encoding='utf-8').write(new_appdata)
+
+if __name__ == "__main__":
+    main()
--- a/script/tools/changelog_update_appdata.py
+++ b/script/tools/changelog_update_appdata.py
@@ -37,7 +37,7 @@ def main():
 {}
         </ul>
       </description>
-    </release>\n'''[1:]
+    </release>\n'''.lstrip("\r\n")
     tmp = ''
     
     regex = r"version ([\d.]+) \(([\w ]+)\)\n(.*?)[\n]{2}"
--- a/script/tools/changelog_update_appdata.py
+++ b/script/tools/changelog_update_appdata.py
@@ -24,10 +24,10 @@ def parse_arguments():
 
 def main():
     args = parse_arguments()
-    
+
     changelog = open(args.changelog_file, encoding='utf-8').read()
     appdata = open(args.appdata_file, encoding='utf-8').read()
-    
+
     tpl = '''
     <release date="{}" version="v{}">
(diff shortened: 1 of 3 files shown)


script/tools/changelog_update_appdata.py
```diff
@@ -0,0 +1,61 @@
...
+import datetime
+from xml.sax.saxutils import escape
+
+#
+# Appstream file could be validated with 'appstream-util validate io.github.ihhub.Fheroes2.appdata.xml'
+#
+
+class CustomArgumentParser(argparse.ArgumentParser):
+    def error(self, message):
+        self.print_usage(sys.stderr)
```
oleg-derevenetz, line 14: Please add `import sys` at the imports section, otherwise this will not work:

```
$ ./changelog_update_appdata.py
Traceback (most recent call last):
  File "/home/oleg/Temp/./changelog_update_appdata.py", line 61, in <module>
    main()
  File "/home/oleg/Temp/./changelog_update_appdata.py", line 25, in main
    args = parse_arguments()
  File "/home/oleg/Temp/./changelog_update_appdata.py", line 22, in parse_arguments
    return parser.parse_args()
  File "/usr/lib/python3.10/argparse.py", line 1826, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/usr/lib/python3.10/argparse.py", line 1859, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/usr/lib/python3.10/argparse.py", line 2100, in _parse_known_args
    self.error(_('the following arguments are required: %s') %
  File "/home/oleg/Temp/./changelog_update_appdata.py", line 14, in error
    self.print_usage(sys.stderr)
NameError: name 'sys' is not defined
```

script/tools/changelog_update_appdata.py
```diff
@@ -0,0 +1,61 @@
...
+def parse_arguments():
+    parser = CustomArgumentParser()
+    parser.add_argument("changelog_file", help="changelog.txt file")
+    parser.add_argument("appdata_file", help="*.appdata.xml file")
+    return parser.parse_args()
+
+def main():
+    args = parse_arguments()
+    
+    changelog = open(args.changelog_file, encoding='utf-8').read()
```
oleg-derevenetz, line 27: Nice looking one-liner, but `changelog_file` will remain opened until the whole script exits :) However, in script like this it's OK.

oleg-derevenetz (changes requested):
Hi @Laserlicht I found one missing import in your script and left one comment, please take a look when you have time.

script/tools/changelog_update_appdata.py
Laserlicht: Oh! Yes of course. :)

script/tools/changelog_update_appdata.py
Laserlicht: Yes, you're right. :)

ihhub:
Hi @Laserlicht , please rebase your branch from `master`. @oleg-derevenetz fixed the issue with code style.

ihhub:
@Laserlicht , I think displaying the changes is always good so players can see that the project is evolving.

script/tools/changelog_update_appdata.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
ihhub, line 1: Is there a reason we want to use Python 3 over 2?

ihhub, line 1: This file is missing copyright header.

script/tools/changelog_update_appdata.py
```diff
@@ -0,0 +1,62 @@
...
+    tpl = '''
+    <release date="{}" version="v{}">
+      <url>https://github.com/ihhub/fheroes2/releases/tag/{}</url>
+      <description>
+        <p>Changes in v{} ({}):</p>
+        <ul>
+{}
+        </ul>
+      </description>
+    </release>\n'''[1:]
```
ihhub, line 40: Here we take all symbols except the first one. Should it be better to just remove **<** from the beginning of the line instead of making a copy of the string?

ihhub (changes requested):
Hi @Laserlicht , I left multiple comments in this pull request. Could you please take a look? On top of this I have a question: where and when this script should be executed?

oleg-derevenetz:
> On top of this I have a question: where and when this script should be executed?

I believe it's a purely service script that should be executed by hand when the next release being prepared, after the `changelog.txt` was updated.

script/tools/changelog_update_appdata.py
oleg-derevenetz: Hi @ihhub the first symbol here is a newline, not `<` :) I believe this is done for the sake of code formatting. However I have an impression that if this script will be used on Windows, there will be in fact two symbols at the beginning - `\r\n` :) Probably it's better to use `strip()` instead.

script/tools/changelog_update_appdata.py
Laserlicht: The script is only running under python3 right now. The question is whether it is really worth the effort to support python2 (it has been out of support for three years now).

oleg-derevenetz: As far as I understand, this script is not intended to be run by end user in any way. It's just an utility script for internal use. I believe that there is no need to support the python 2, because it will be run mostly by @ihhub in the controlled environment, right?

ihhub: This is correct.

Laserlicht: Yes, it is currently missing from all scripts. Maybe this should be corrected in a separate commit for all?

script/tools/changelog_update_appdata.py
```diff
@@ -37,7 +37,7 @@ def main():
 {}
         </ul>
       </description>
-    </release>\n'''[1:]
+    </release>\n'''.lstrip("\r\n")
```
oleg-derevenetz, line 40: Just `lstrip()` (without parameters) will be enough, because it will strip any whitespace characters, including newlines, carriage returns, spaces, tabs, etc, etc.

script/tools/changelog_update_appdata.py
Laserlicht: Right. Is only for styling. Have now converted it to `lstrip`.

script/tools/changelog_update_appdata.py
Laserlicht: @oleg-derevenetz Not really. It breaks the format. The indentation before `<release>` is lost.

oleg-derevenetz: @Laserlicht Ah yes, the indentation, right.

Laserlicht:
> I believe it's a purely service script that should be executed by hand when the next release being prepared, after the `changelog.txt` was updated.

Exactly. I don't think it's really worth building that into the CI/CD here.

script/tools/changelog_update_appdata.py
oleg-derevenetz: If we want to have copyright headers in all scripts, we should prepare the infrastructure first to verify their presence. The difference of scripts is that they should have a shebang in the first line. I propose to address this later in separate PR.

oleg-derevenetz (approved):

ihhub:
@Laserlicht , a bunch of thanks for this change!

PR #6441, opened 2023-01-01
extract_homm2_resources.sh: detect multiple Python versions

oleg-derevenetz:
On certain systems (including the modern macOS versions) there may be only `python3` binary available.


ihhub (approved):

ihhub:

PR #6442, opened 2023-01-01
Render spell effect animation so as not to cover the front troops. Battlefield.

Districh-ru:
This PR makes the spell that effect is only rendered above the creature it is applied and creatures on the same row or higher (farther), while not overlapping the front (closer) creatures.
fheroes2 2023-01-01 19-40-05-667 fheroes2 2023-01-01 19-36-55-469

To implement this the 'overlaySprite' structure was added to 'Unit' class. And this structure is used in `RedrawArmies()` function.

<details><summary>The description of the issue.</summary>
Here's an issue I've found:
Red fog is drawn over troops standing in front of the target.

Wave

In the OG:
image

</details>

_Issue originally posted by @Branikolog in https://github.com/ihhub/fheroes2/issues/6344#issuecomment-1345546298_

This PR affects these spell effect animations:
(Mass) Shield, (Mass) Cure, (Mass) Haste, (Mass) Slow, Mass Bless, (Mass) Curse, (Mass) Dispel,
Bless, Blind, Anti Magic, Stone Skin, Steel Skin, Paralyze, Hypnotize, Dragon Slayer, Berserker,
Death Wave, Death Ripple, Holy Word, Holy Shout, Cold Ray, Resurrection.
And also Morale effect animation.


Districh-ru:
Hi, @ihhub! I don't think that my implementation of this idea is good enough. Could you please check it in your spare time.
Also the `RedrawArmies()` function has duplicate code (and I added new ones). I'll want to rewrite it soon to reduce duplication and to fix a couple of issues.

src/fheroes2/battle/battle_troop.h
```diff
@@ -34,6 +34,7 @@
 #include "battle_animation.h"
 #include "battle_cell.h"
 #include "bitmodes.h"
+#include "image.h"
```
ihhub, line 37: Mixing logic and UI is not the best approach. I would restrain from doing such. As an alternative we can have a map of monsters IDs vs their overlay images which is stored either in `Battle::Interface` class. What do you think?

src/fheroes2/battle/battle_troop.h
```diff
@@ -71,6 +72,14 @@ namespace Battle
         uint32_t FindZeroDuration() const;
     };
 
+    // Sprite to render over the unit (i.e. spell effect animation)
+    struct OverlaySprite
+    {
+        const fheroes2::Sprite * overlaySprite = nullptr;
+        fheroes2::Point position = fheroes2::Point( 0, 0 );
+        bool reflect = false;
```
ihhub, line 80: Please use C++11 initialization list:
```cpp
const fheroes2::Sprite * overlaySprite{ nullptr };
fheroes2::Point position; <-- no need as default constructor sets zeroes
bool reflect{ false };
```

src/fheroes2/battle/battle_troop.h
```diff
@@ -100,6 +109,30 @@ namespace Battle
...
+        }
+
+        bool isOverlaySpriteReflect() const
+        {
+            return overlaySprite.reflect;
+        }
+
+        void SetOverlaySprite( const fheroes2::Sprite & sprite, fheroes2::Point pos, bool flip );
+
+        void ResetOverlaySprite();
```
ihhub, line 134: If we move to usage of a map then most of these methods disappear as during rendering we will have the direct access to all structures.

src/fheroes2/battle/battle_troop.h
Districh-ru: Hi, @ihhub, thanks for the advice, I'll try this approach soon.

Districh-ru: Hi @ihhub, I've tried to implement the overlay sprite for spells in `Battle::Interface`.
Could you please check my implementation when you have time.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5744,12 +5823,17 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
...
                     }
                 }
             }
 
             ++frame;
+
+            // Remove all overlay sprites when the animation is finished.
+            if ( frame == maxFrame ) {
+                _unitOverlaySprite.erase( overlaySpriteBegin, overlaySpriteEnd );
+            }
```
ihhub, line 5836: Let's add this part just in case:
```cpp
if ( frame == maxFrame ) {
    _unitOverlaySprite.erase( overlaySpriteBegin, overlaySpriteEnd );
}
else {
    assert( frame < maxFrame );
}
```
This is to make sure that we don't screw up logic above.

The same comment for lines 5893-5896.

src/fheroes2/battle/battle_interface.h
```diff
@@ -94,6 +94,16 @@ namespace Battle
...
     };
 
+    // Sprite to render over the unit (i.e. spell effect animation)
+    struct OverlaySprite
+    {
+        OverlaySprite( const uint32_t unitId, const bool reflect );
+        uint32_t unitId{ 0 };
+        const fheroes2::Sprite * overlaySprite{ nullptr };
+        fheroes2::Point position;
+        bool reflect{ false };
```
ihhub, line 104: Can we please name it like `isReflectedImage`? In this case it would be obvious what this member is for.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5694,20 +5773,20 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
         CheckGlobalEvents( le );
 
         if ( Game::validateAnimationDelay( Game::BATTLE_SPELL_DELAY ) ) {
-            RedrawPartialStart();
-
             if ( frame < maxFrame ) {
-                for ( const auto & target : targets ) {
+                std::vector<Battle::OverlaySprite>::iterator overlaySprite = overlaySpriteBegin;
```
ihhub, line 5777: Can we please name this variable differently to differ it from **overlaySprite** of `OverlaySprite` structure? Something like `overlaySpriteIter` would work.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5788,14 +5874,11 @@ void Battle::Interface::RedrawTroopWithFrameAnimation( Unit & unit, int icn, int
         CheckGlobalEvents( le );
 
         if ( Game::validateAnimationDelay( Game::BATTLE_SPELL_DELAY ) ) {
-            RedrawPartialStart();
-
             if ( frame < maxICNFrame ) {
-                const fheroes2::Sprite & spellSprite = fheroes2::AGG::GetICN( icn, frame );
-                const fheroes2::Point & pos = CalculateSpellPosition( unit, icn, spellSprite );
-                fheroes2::Blit( spellSprite, _mainSurface, pos.x, pos.y, reflect );
+                _unitOverlaySprite.back().overlaySprite = &fheroes2::AGG::GetICN( icn, frame );
```
ihhub, line 5878: This is very dangerous to do. We have a code to even clear some images during runtime. What if we store **icnId** and **icnIndex** in the structure and then call `fheroes2::AGG::GetICN()` at the time we need to blit an image?

ihhub (changes requested):
Hi @Districh-ru , I put several comments in this pull request. Could you please take a look when you have time?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1439,9 +1449,20 @@ void Battle::Interface::RedrawArmies()
...
+                                if ( overlaySprite.unitId == deadUnits[i]->GetUID() && overlaySprite.overlaySprite != nullptr ) {
+                                    troopOverlaySpriteBeforeWall.emplace_back( &overlaySprite );
+                                }
+                            }
                         }
                         else {
                             deadTroopAfterWall.emplace_back( deadUnits[i] );
+                            for ( const Battle::OverlaySprite & overlaySprite : _unitOverlaySprite ) {
+                                if ( overlaySprite.unitId == deadUnits[i]->GetUID() && overlaySprite.overlaySprite != nullptr ) {
+                                    troopOverlaySpriteBeforeWall.emplace_back( &overlaySprite );
```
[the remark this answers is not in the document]
Districh-ru: Hi @ihhub, many thanks for the review and suggestions.
Here I made a mistake (forgot to rename after the copy-paste), you are right it should be `troopOverlaySpriteAfterWall`.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1494,6 +1527,11 @@ void Battle::Interface::RedrawArmies()
                 RedrawTroopSprite( *movingTroopBeforeWall[i] );
             }
 
+            // Render the overlay srite for units in current cell row above all units in this and upper rows.
+            for ( const Battle::OverlaySprite * overlaySprite : troopOverlaySpriteBeforeWall ) {
+                fheroes2::Blit( *overlaySprite->overlaySprite, _mainSurface, overlaySprite->position.x, overlaySprite->position.y, overlaySprite->reflect );
```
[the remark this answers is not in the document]
Districh-ru: I've added an assertion here and in the similar code.

src/fheroes2/battle/battle_interface.h
```diff
@@ -94,6 +94,16 @@ namespace Battle
         CAPTAIN
     };
 
+    // Sprite to render over the unit (i.e. spell effect animation)
+    struct OverlaySprite
```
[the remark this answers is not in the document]
Districh-ru: For now it is used only for rendering spell effects for units, so the name `UnitSpellEffectImage` will be OK. If we plan to use it for some other effects, maybe in expansion, we will rename it then.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: The loop can go further than `maxFrame` when we have a long wince or death animation, like for Bone Dragons or Phoenixes. So if I put an assertion here, we will get it in such 'legal' animations. :)

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi @ihhub, I've tried to implement like you suggested. 
Could you please check the code in your spare time.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1439,9 +1444,23 @@ void Battle::Interface::RedrawArmies()
                     if ( deadUnits[i] && cellId != deadUnits[i]->GetTailIndex() ) {
                         if ( isCellBefore ) {
                             deadTroopBeforeWall.emplace_back( deadUnits[i] );
+
+                            // Check for overlay sprites of dead units (i.e. Resurrect spell).
+                            for ( const Battle::UnitSpellEffectImage & overlaySprite : _unitOverlaySprite ) {
+                                if ( overlaySprite.unitId == deadUnits[i]->GetUID() && overlaySprite.icnId != ICN::UNKNOWN ) {
```
ihhub, line 1450: Is there a case when **icnId** is going to be **ICN::UNKNOWN**? I see no place where we possibly set this value or create an instance of `UnitSpellEffectImage` structure using default constructor. If there is no such a case please remove all checks but add an assertion before calling **fheroes2::AGG::GetICN()** to make sure that **icnId** is not **ICN::UNKNOWN**.

src/fheroes2/battle/battle_interface.h
```diff
@@ -443,6 +460,8 @@ namespace Battle
         };
 
         BridgeMovementAnimation _bridgeAnimation;
+
+        std::vector<UnitSpellEffectImage> _unitOverlaySprite;
```
ihhub, line 464: Since this is an array of `UnitSpellEffectImage` should we name this member as `_unitSpellEffectImages` / `_unitSpellEffectInfos` ?

src/fheroes2/battle/battle_interface.h
```diff
@@ -94,6 +95,22 @@ namespace Battle
         CAPTAIN
     };
 
+    // Sprite to render over the unit (spell effect animation)
+    struct UnitSpellEffectImage
```
ihhub, line 99: How about name it as **UnitSpellEffectInfo**? This structure is not exactly an image but a bucket of data for rendering an image. What do you think?

ihhub (changes requested):
Hi @Districh-ru , I put few more comments here. Would you mind check them and see if they make sense?

src/fheroes2/battle/battle_interface.h
Districh-ru: Hi @ihhub, yes, with the current structure contents this name is better. 👍 It will be like `PopupDamageInfo`. But near to `UnitSpellEffectInfo` we have `OpponentSprite` class, which does not contain a sprite image, but has info. :) So maybe it should be also renamed to something like **OpponentSpriteInfo** or **OpponentImageInfo**? If you think so, I can rename this class in this PR.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5694,20 +5774,21 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
...
+                const fheroes2::Sprite & spellSprite = fheroes2::AGG::GetICN( icn, frame );
+
+                for ( const Battle::TargetInfo & target : targets ) {
                     if ( target.defender ) {
-                        const bool reflect = ( isReflectICN && target.defender->isReflect() );
-                        const fheroes2::Sprite & spellSprite = fheroes2::AGG::GetICN( icn, frame );
-                        const fheroes2::Point & pos = CalculateSpellPosition( *target.defender, icn, spellSprite );
-                        fheroes2::Blit( spellSprite, _mainSurface, pos.x, pos.y, reflect );
+                        ( *overlaySpriteIter ).position = CalculateSpellPosition( *target.defender, icn, spellSprite );
+                        ( *overlaySpriteIter ).icnIndex = frame;
```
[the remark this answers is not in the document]
Districh-ru: Yeah, I'm not looking for easy ways :)))

src/fheroes2/battle/battle_interface.h
ihhub: Let's do old code cleanup in a separate pull request. I have no objections of doing so but it's better to keep this pull request as small as possible for the future track of changes.

ihhub (approved):

src/fheroes2/battle/battle_interface.h
Districh-ru: Sure!

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -5680,6 +5749,20 @@ void Battle::Interface::RedrawTargetsWithFrameAnimation( const TargetsInfo & tar
 
     // For certain spells reflect the spell sprite if the creature is reflected.
     const bool isReflectICN = ( icn == ICN::SHIELD || icn == ICN::REDDEATH || icn == ICN::MAGIC08 );
+
+    _unitSpellEffectInfos.reserve( targets.size() );
```
oleg-derevenetz, line 5753: According to the logic below, there can already be some elements in the `_unitSpellEffectInfos`, right? (BTW, where will they come from there?) If yes, then we should call `_unitSpellEffectInfos.reserve( _unitSpellEffectInfos.size() + targets.size() );` because we need to take into account the already existing elements.

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left one suggestion (accompanied by a question), please take a look when you have time.

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: > I made this implementation for possible future use if we decide (for example in the expansion) to display some sprites above the troops for some time.

I have no objections against this, but then please add the corresponding TODO comment to the declaration of `_unitSpellEffectInfos` in the `battle_interface.h` describing that while we currently don't need to persist the `UnitSpellEffectInfo`s, this may be needed in the future to [... describe why ...]. Also maybe we can add assertions at the beginning or at the end of methods where `_unitSpellEffectInfos` is currently used (`RedrawTargetsWithFrameAnimation()` and `RedrawTroopWithFrameAnimation()`?) to verify that `_unitSpellEffectInfos` is empty at the beginning and remains empty at the end (or maybe just at the beginning or at the end - at your discretion).  When there is a need to display some permanent effects, we will remove them, but in the meantime it's good to make sure that the logic works correctly.

Branikolog (approved):
Excellent!

oleg-derevenetz (approved):

ihhub:
@Districh-ru , huge thanks for improving this part of the game!

PR #6447, opened 2023-01-02
Reduce CPU usage in Select Language dialog

### LINKED ISSUE #6445: High CPU usage in Select Language dialog
### Preliminary checks

### Platform

Linux

### Describe the bug

When the "Select Language" dialog is open, a single CPU core is used to ~90%, which is not the case for normal game operations or other dialogs.

### Save file

No save file needed to reproduce.

### Additional info

_No response_

Districh-ru:
close #6445
Do not call `setGameLanguage` on every loop, do it only if `chosenLanguage` was changed.


ihhub (approved):

ihhub:

PR #6450, opened 2023-01-02
Streamlining MacOS installation instructions

kfishster:
After going through the installation process for MacOS, I wanted to contribute some installation pointers back to the docs. This PR addresses

* ensuring that every Mac installation path includes the last step of gathering game resources
* adding a link to a helpful library to open a self-extracting exe without having to install Windows emulation software on a unix-based machine
  * can also add this to the linux instructions, should I also create a similar section there?
* adding the path to the scripts when using the homebrew installation for Mac

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -139,18 +139,18 @@ Once you obtain the fheroes2 executable using any of the options above, you shou
 * If you have a legally purchased copy of the original game, run the extract resources script which will prompt you to enter 
   the location of the original game, and will extract all the necessary resource files. Script can be run from the following paths depending on how you installed `fheroes2`
   * `fheroes2-extract-resources` if you used a package manager (MacPorts or Homebrew)
-  * `script/homm2/extract_homm2_resources_for_app_bundle.sh` if you built from source using the[**macOS native app**](#macos-native-app) method
+  * `script/homm2/extract_homm2_resources_for_app_bundle.sh` if you built from source using the [**macOS native app**](#macos-native-app) method
   * `script/homm2/extract_homm2_resources.sh` for all other cases
 
 * As an alternative to the previous step, you can manually copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` (some of them may
   be missing depending on the version of the original game) from the original game directory to 
-  * `~/Library/Application Support/fheroes2` if you built from source using the[**macOS native app**](#macos-native-app) method
+  * `~/Library/Application Support/fheroes2` if you built from source using the [**macOS native app**](#macos-native-app) method
   * `~/.fheroes2` for all other cases
 
 * If you don't have a legally purchased copy of the original game, you can download and install the demo version of the original game
   by running the download demo script. Script can be run from the following paths depending on how you installed `fheroes2`
   * `fheroes2-install-demo` if you used a package manager (MacPorts or Homebrew)
-  * `script/homm2/download_demo_version_for_app_bundles.sh` if you built from source using the[**macOS native app**](#macos-native-app) method
+  * `script/homm2/download_demo_version_for_app_bundles.sh` if you built from source using the [**macOS native app**](#macos-native-app) method
   * `script/homm2/download_demo_version.sh` for all other cases
 
 <a name="linux"></a>
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -89,6 +89,8 @@ port install fheroes2
 
 Then follow the instructions on the screen.
 
+Follow the [instructions below](#macos-resources) to gather resources necessary for `fheroes2` to function as expected.
+
 <a name="homebrew-mac"></a>
 ### Homebrew
 
@@ -97,6 +99,10 @@ If you are using [**Homebrew**](https://brew.sh/), you can install the game by r
 ```sh
 brew install fheroes2
 ```
+
+You will find the extraction and demo download scripts in the following path: `/opt/homebrew/Cellar/fheroes2/<version>/share/doc/fheroes2/homm2/`
+
+Follow the [instructions below](#macos-resources) to gather resources necessary for `fheroes2` to function as expected.
 
 <a name="macos-native-app"></a>
 ### macOS native app
@@ -108,15 +114,7 @@ make FHEROES2_MACOS_APP_BUNDLE=ON
 make FHEROES2_MACOS_APP_BUNDLE=ON bundle
 ```
 
-* If you have a legally purchased copy of the original game, run the `extract_homm2_resources_for_app_bundle.sh` script supplied in the
-  `script/homm2` directory. You will be prompted to enter the location of the original game, and the script will extract all the
-  necessary resource files.
-
-* As an alternative to the previous step, you can manually copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` (some of them may
-  be missing depending on the version of the original game) from the original game directory to `~/Library/Application Support/fheroes2`.
-
-* If you don't have a legally purchased copy of the original game, you can download and install the demo version of the original game
-  by running the `download_demo_version_for_app_bundle.sh` script supplied in `script/demo` directory.
+Follow the [instructions below](#macos-resources) to gather resources necessary for `fheroes2` to function as expected.
 
 <a name="macos-zip-archive"></a>
 ### macOS ZIP archive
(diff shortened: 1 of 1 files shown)


docs/INSTALL.md
```diff
@@ -129,14 +127,27 @@ make FHEROES2_MACOS_APP_BUNDLE=ON bundle
...
+### Gathering game resources
+
+Once you obtain the fheroes2 executable using any of the options above, you should follow these steps to load in the correct resources
+
+* If you have a legally purchased copy of the original game in a self-extracting Windows executable (such as from GOG), you can utilize 
+  [`innoextract`](https://constexpr.org/innoextract/#use) to extract files out of the exe without having to use wine/windows emulation 
+  software on your Unix-based machine
+
+* If you have a legally purchased copy of the original game, run the `extract_homm2_resources.sh` script (`extract_homm2_resources_for_app_bundle.sh` 
+  if built from source) supplied in the `script/homm2` directory (or directory of the pre-built binary installation). You will be prompted to enter
```
oleg-derevenetz, line 142: fheroes2 for macOS can be built from source in two ways:

1. As regular *NIX app. In this case the `extract_homm2_resources.sh` should be used;
2. As macOS-specific app bundle (so-called ["macOS native app"](https://github.com/ihhub/fheroes2/blob/master/docs/INSTALL.md#macos-native-app)). In this case the directory layout is different and `extract_homm2_resources_for_app_bundle.sh` script should be used. I suggest to clarify that we are talking about the app bundle here.

docs/INSTALL.md
```diff
@@ -129,14 +127,27 @@ make FHEROES2_MACOS_APP_BUNDLE=ON bundle
...
+  [`innoextract`](https://constexpr.org/innoextract/#use) to extract files out of the exe without having to use wine/windows emulation 
+  software on your Unix-based machine
+
+* If you have a legally purchased copy of the original game, run the `extract_homm2_resources.sh` script (`extract_homm2_resources_for_app_bundle.sh` 
+  if built from source) supplied in the `script/homm2` directory (or directory of the pre-built binary installation). You will be prompted to enter 
+  the location of the original game, and the script will extract all the necessary resource files.
 
 * As an alternative to the previous step, you can manually copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` (some of them may
-  be missing depending on the version of the original game) from the original game directory to the fheroes2 installation directory.
+  be missing depending on the version of the original game) from the original game directory to `~/Library/Application Support/fheroes2`.
```
oleg-derevenetz, line 146: This path is valid only if you build the app bundle. If not (MacPorts doesn't build it, Homebrew doesn't build it as well, correct me if I'm wrong), the correct path is `~/.fheroes2`.

docs/INSTALL.md
```diff
@@ -129,14 +127,27 @@ make FHEROES2_MACOS_APP_BUNDLE=ON bundle
...
 
 * If you don't have a legally purchased copy of the original game, you can download and install the demo version of the original game
-  by running the `download_demo_version.sh` script supplied in the ZIP archive.
+  by running the `download_demo_version.sh` (`download_demo_version_for_app_bundle.sh` if built from source) script supplied in `script/demo` directory (or directory of the pre-built binary installation).
```
oleg-derevenetz, line 149: The same suggestion about app bundle.

oleg-derevenetz (commented):
Hi @kfishster I left a few suggestions, please take a look when you have time.

docs/INSTALL.md
```diff
@@ -98,6 +100,10 @@ If you are using [**Homebrew**](https://brew.sh/), you can install the game by r
 brew install fheroes2
 ```
 
+You will find the extraction and demo download scripts in the following path: `/opt/homebrew/Cellar/fheroes2/<version>/share/doc/fheroes2/homm2/`
```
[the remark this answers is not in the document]
kfishster: You're right, homebrew does, indeed, copy over the utility bins, but it looks like the 1.0.0 version did not bundle the `extract-resources`  bin
 
image

Looks like the homebrew formula only lists the demo bin: https://github.com/Homebrew/homebrew-core/blob/master/Formula/fheroes2.rb 

I can follow up on fixing that next, and I'll re-structure the instructions to reflect these bins.

docs/INSTALL.md
kfishster: Corrected this in the latest iteration, explicitly listing path for app-bundle build

docs/INSTALL.md
kfishster: Corrected this in the latest iteration, mistake comes from my own mistake when reading the installation guide, hopefully these edits makes things clearer!

kfishster:
The motivation for this change was initially combining some of the repeated instructions, but as you pointed out, there are slight differences depending on the way you get the binary, which fully justifies the previous pattern. The latest iteration just moves some of that down to a different section. 

Ideally, this might be best done via some sort of an installation script - replacing the demo/extraction script with a single script that walks the user through the options - demo/full - checks whether sdl2 dependencies are installed, and, depending on whether it was installed as a mac native app, putting the resources in the right place. I guess that would be just putting another layer of abstraction over this, but wanted to see what you think about that idea :)

docs/INSTALL.md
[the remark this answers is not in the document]
kfishster: Opened PR to correct Homebrew formula: https://github.com/Homebrew/homebrew-core/pull/119615

oleg-derevenetz:
> Ideally, this might be best done via some sort of an installation script - replacing the demo/extraction script with a single script that walks the user through the options - demo/full - checks whether sdl2 dependencies are installed, and, depending on whether it was installed as a mac native app, putting the resources in the right place. I guess that would be just putting another layer of abstraction over this, but wanted to see what you think about that idea :)

On Windows, we have a similar function in the installer:

screenshot

But I don't think that we should have this function if app is installed using the package manager. I believe it's a package maintainer's responsibility to provide scripts of such kind, because in various systems/distributions various [unexpected paths](https://www.reddit.com/r/HoMM/comments/zy152v/comment/j23muwp/?utm_source=reddit&utm_medium=web2x&context=3) can be used. What we can do is provide some basic "building blocks" that can be combined and/or patched by the package maintainer.

For example, MacPorts [package](https://github.com/macports/macports-ports/blob/master/games/fheroes2/Portfile) configures the auxiliary scripts and provides a quite detailed instructions as a result of `port install fheroes2` command like this:

```
notes "
You will need to have a demo version or the full version of Heroes of Might and
Magic II game to be able to play.

If you do not have the original Heroes of Might and Magic II game, type the
following command:

fheroes2-install-demo

This script will download and install all the necessary files from the demo
version of the original Heroes of Might and Magic II game.

If you have a legally purchased copy of the original game, type the following
command:

fheroes2-extract-resources

This script will extract all the necessary resource files from the original
Heroes of Might and Magic II game.

As an alternative to the previous step, you can manually copy the subdirectories
'ANIM', 'DATA', 'MAPS' and 'MUSIC' (some of them may be missing depending on the
version of the original game) from the original game directory to the '.fheroes2'
subdirectory of your home directory.

When all prerequisites are in place, you can start the game by typing:

fheroes2

Please visit project website:

${homepage}

and read ${prefix}/share/doc/fheroes2/README.txt for more information.
```

That's why current installation instructions for [MacPorts](https://github.com/ihhub/fheroes2/blob/master/docs/INSTALL.md#macports) are so laconic - the user should literally "follow the instructions on the screen" and that's all. If something like this were for Homebrew, it would be good.

However, of course, `innoextract` tool should be mentioned in `INSTALL.md` because it's a 3rd-party tool (which BTW can be installed through MacPorts, as well as, I suppose, via Homebrew, which could be mentioned in the `INSTALL.md`). I can mention the `innoextract` in this MacPorts instruction during the next port update.

oleg-derevenetz:
Please note the #6451 (however, naturally, it will become generally available only in the next release).

kfishster:
Got it, that makes sense. I believe you can add messaging to Homebrew, so I can work on that change so as to make both of the MacOS package manager experiences consistent. Your change to add `innoextract` to the extraction script is a great solution. Should I still merge this documentation in and we can amend it once #6451 is merged as well?

oleg-derevenetz:
> I believe you can add messaging to Homebrew, so I can work on that change so as to make both of the MacOS package manager experiences consistent.

Could you please do this as well (as you fixed the `fheroes2-extract-resources` case)? I don't use Homebrew and will not be able to properly test the changes.

> Should I still merge this documentation in and we can amend it once https://github.com/ihhub/fheroes2/pull/6451 is merged as well?

We can merge this PR as is and amend them later once 1.01 will be released. LGTM, I will request a review from a few more community members.

oleg-derevenetz (approved):

kfishster:
> Could you please do this as well (as you fixed the `fheroes2-extract-resources` case)? I don't use Homebrew and will not be able to properly test the changes.

Yep, I'll work on that change! I apologize for the wording, when I said "you could change" I meant that globally, not you specifically. I'll make that PR.

kfishster:
(sorry, closed PR, not merged, will wait for code maintainers to merge)

ihhub (approved):

ihhub:
@kfishster , thanks a lot for the contribution to the project!

PR #6451, opened 2023-01-03
extract_homm2_resources.sh: add the innoextract support

oleg-derevenetz:
Ease of use for package maintainers and end users.

### DIFF
--- a/script/homm2/extract_homm2_resources.sh
+++ b/script/homm2/extract_homm2_resources.sh
@@ -69,11 +69,11 @@ if [[ -f "$HOMM2_PATH" ]]; then
 
     innoextract -e -s -d "$EXTRACT_DIR" -- "$HOMM2_PATH"
 
-    for SUBDIR in . app; do
-        if verify_homm2_path "$EXTRACT_DIR/$SUBDIR"; then
+    for ITEM in "$EXTRACT_DIR" "$EXTRACT_DIR"/*; do
+        if verify_homm2_path "$ITEM"; then
             echo_green "HoMM2 installer package: $HOMM2_PATH"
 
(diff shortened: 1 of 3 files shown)


script/homm2/extract_homm2_resources.sh
```diff
@@ -45,21 +49,47 @@ fi
...
 echo_green "Destination directory: $DEST_PATH"
 
-echo_stage "[2/3] determining the HoMM2 directory"
+echo_stage "[2/3] determining the HoMM2 directory or installer package"
 
 if [[ "$#" -gt "0" ]]; then
     HOMM2_PATH="$1"
 else
-    read -e -p "Please enter the full path to the HoMM2 directory (e.g. /home/user/GOG Games/HoMM 2 Gold): " HOMM2_PATH
+    read -e -p "Please enter the full path to the HoMM2 directory or installer package (e.g. /home/user/GOG Games/HoMM 2 Gold): " HOMM2_PATH
```
kfishster, line 57: [nit] you can provide another example in the "e.g" to indicate that installer package could be "exe"

```
Please enter the full path to the HoMM2 directory or installer package (e.g. /home/user/GOG Games/HoMM 2 Gold or /home/user/games/installer_package.exe)
```

It's not absolutely clear that the downloaded .exe from GoG could be unpacked, so maybe that adds some helpful context

ihhub (approved):

ihhub:

kailight:
Not sure if its related, but when trying to extract_homm2_resources.sh using .exe from GoG

error reason: could not open slice 0: "homm2-1.bin": iostream error

The error is not helpful, I do not know what to do next

PR #6453, opened 2023-01-04
Don't show magic guild if hero in castle does not have a spellbook

kfishster:
## Current behavior
1. Hero opens magic guild
2. Hero prompted to purchase a spellbook
3. Hero clicks "No"
=> Magic guild opens

3. Hero clicks "Yes"
=> Magic guild opens

## Expected behavior (what is in the original game and this fix)
1. Hero opens magic guild
2. Hero prompted to purchase a spellbook
3. Hero clicks "No"
=> Prompt closes, magic guild does not open

4. Hero clicks "Yes"
=> Spellbook is purchased and magic guild opens

You can argue for both fixing this and not fixing this:
* **Fix**: make it follow the OG game behavior, from the perspective of the hero, if they don't purchase the book, they should not be shown the spells
* **No Fix**: in the OG game it is annoying that having a hero in the castle without a book blocks you from seeing the spells available

This PR is to take the **Fix** path, but I'll defer to the community to weight in on this decision.

### DIFF
--- a/src/fheroes2/castle/castle_dialog.cpp
+++ b/src/fheroes2/castle/castle_dialog.cpp
@@ -494,7 +494,9 @@ Castle::CastleDialogReturnValue Castle::OpenDialog( const bool openConstructionW
                             need_redraw = true;
                         }
 
-                        OpenMageGuild( hero );
+                        if ( !( hero != nullptr && !hero->HaveSpellBook() ) ) {
+                            OpenMageGuild( hero );
+                        }
                     }
                     else if ( isMonsterDwelling ) {
                         fheroes2::ButtonRestorer exitRestorer( buttonExit );


Branikolog:
Hello, @kfishster !
I believe we had an issue long time ago regarding this difference. Current behaviour was implemented intentionally. 
In my personal point of view, not being able to check mage guild spells with the above conditions is annoying and is an example of bad UI logic. We can still check it, when there's no hero in town. Why should we forbid player to see which spells are available in mage guild, when hero without book stays in town garrison?
The only thing I can imagine we can improve here, is when player clicks "No" on buying book, we can show a message, that all spells in mage guild will not be available for visiting hero.

kfishster:
@Branikolog I appreciate the explanation, I suspected this might have been done by design. Just to know for next time - should I have created an issue first and then created the fix, or maybe pitched it in Discord? What's the best procedure? I might have gotten a bit too excited to find a bug here.

oleg-derevenetz:
> The only thing I can imagine we can improve here, is when player clicks "No" on buying book, we can show a message, that all spells in mage guild will not be available for visiting hero.

IIRC there is already an indication of the fact that the spells in this mage guild are not available for a visiting hero if he does not have a spellbook - the different message will be shown at the bottom of the mage guild screen (it will not mention that these spells are added to the hero's spellbook).

Branikolog:
@kfishster 
It's ok with both opening issue here or asking at discord bug section. You could also create a discussion here, if you want to be sure, the issue is valid and you want more deeply discuss your possible fix.
I believe, these steps should go before you're going to create a fix. Heroes2 game has really complex and diverse logic. Even old players tend to forget something and the issue they've noticed was just false. There were dozens of cases, when I've created a new thread, and closed it in a while,after realized, everything works as intended. :)

oleg-derevenetz:
> What's the best procedure?

Usually it's better to open an issue or discussion first. However if there is an obvious issue (e.g. crash) and you have a fix, it may be acceptable to describe this issue directly in the PR.

Branikolog:
> > The only thing I can imagine we can improve here, is when player clicks "No" on buying book, we can show a message, that all spells in mage guild will not be available for visiting hero.
> 
> IIRC there is already an indication of the fact that the spells in this mage guild are not available for a visiting hero if he does not have a spellbook - the different message will be shown at the bottom of the mage guild screen (it will not mention that these spells are added to the hero's spellbook).

I was thinking the same. :) But I've checked once more, the message in status bar says: "The above spells are available here." 
Could he have just opened issue but haven't implemented the improvement yet? Couldn't find it...

oleg-derevenetz:
> Could he have just opened issue but haven't implemented the improvement yet?

It was implemented. Here is a screen without the spellbook:



And here is with spellbook:

Branikolog:
@oleg-derevenetz 
Ah, right! Forgot to check the case with spell book. I thought, there was a phrase saying, hero cannot learn spells. :)
Thanks!

kfishster:
Got it, ok, I'll close this PR and take the other paths if I find more things to fix!

PR #6455, opened 2023-01-04
Set audio frequency to 44100 on Windows

### LINKED ISSUE #6429: Music crackling/popping
### Preliminary checks

### Platform

Windows

### Describe the bug

Most of the music in this version of the game has a very noticeable crackle, which can be noticed only few seconds into the main menu music for example, but happens during all the music in game I heard so far.

Of course I've tried first if the issue is on my side, but was not able to resolve it: My audio drivers are up to date. I have no audio enhancements on my audio device enabled. My format is set to 48000Hz 24Bits and I've tried almost every possible setting too (reducing channels to 16Bits, changing sample rate to 44100Hz or even 96000Hz).

**However**, switching from my headphones to my speakers, I didn't hear any crackle, so it is most likely on my side. I am still putting this here in case of a suggestion and also because the same music does not have crackle in the other versions of the game (GOG version for DOS/Windows are both fine), so it still possibly has something to do with the way the playback works in this project.

### Save file

Save file not required to reproduce the issue.

### Additional info

Playing the track (homm2_41.ogg) itself via a media player in Windows, I do not hear the crackle either, even with volume set to much higher than in game.

oleg-derevenetz:
fix #6429

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -68,7 +68,7 @@ namespace
             channels = 2; // Support stereo audio.
             silence = 0;
 #if defined( ANDROID )
-            // TODO: a value greater than 1024 causes audio distortion on Android
+            // Value greater than 1024 causes audio distortion on Android
             samples = 1024;
 #else
             samples = 2048;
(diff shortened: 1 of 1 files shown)


ihhub:
@oleg-derevenetz , thank you so much for fixing it!

PR #6458, opened 2023-01-05
Android Toolset Translation (German)

Laserlicht:
German Translation of Toolset App.

If anyone wants to contribute other languages: feel free to attach xml file, then I'll expand the PR.

@oleg-derevenetz It should be enough to add the file, or is there more to be done?

### DIFF
--- a/android/app/src/main/res/values-de/strings.xml
+++ b/android/app/src/main/res/values-de/strings.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+    <string name="app_name">fheroes2</string>
+    <string name="activity_toolset_label">fh2 Werkzeugkasten</string>
+    <string name="activity_toolset_game_logo_img_content_description">fheroes2 Logo</string>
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
> It should be enough to add the file, or is there more to be done?

No, just adding the file should be enough.

android/app/src/main/res/values-de/strings.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+    <string name="app_name">fheroes2</string>
+    <string name="activity_toolset_label">fh2 Werkzeugkasten</string>
```
oleg-derevenetz, line 5: How this label looks on the launcher screen? Isn't a bit long?

Laserlicht: Yes, it is shortened. But you can still see what is meant.

The question is, is that so bad? For me, about a third of the app labels are shortened. Also official ones from google.

grafik

oleg-derevenetz: Yes, that's true. For some reason, Google doesn't even allow to break icon labels into two lines instead shortening.

ihhub:
@Laserlicht , thank you so much for this addition to the project!

PR #6459, opened 2023-01-05
Exclude Java files from SonarCloud analysis

oleg-derevenetz:
Currently SonarCloud [fails the analysis](https://github.com/ihhub/fheroes2/actions/runs/3842572586/jobs/6544128440) because project now includes multiple Java files (for Android) and they are not compiled in the SonarCloud workflow. While in theory we can try to build Android project for SonarCloud, this means that we will not build some other parts of the project (for example, tools), and also this will be slower than just build the Linux project, so I decided to just exclude Java files from SonarCloud analysis.


ihhub:

PR #6464, opened 2023-01-05
Combat: allow wide units to take a step back

### LINKED ISSUE #6462: cannot move a large creature one cell backwards
### Preliminary checks

### Platform

Windows

### Describe the bug

I cannot move a large creature one cell backwards. For this, on the rear cell of a large creature, the mouse pointer should offer to move rather than to open the creature detail window.

### Save file

This is a general problem, it can reproduced in every buttle with every large creature.

### Additional info

_No response_

oleg-derevenetz:
fix #6462


ihhub:

PR #6467, opened 2023-01-07
Battle::Interface::GetBattleCursor(): add the extra assertion

oleg-derevenetz:
Follow-up to #6464 - just in case.


ihhub:

PR #6483, opened 2023-01-14
Fix an infinite loop in the RedrawActionWincesKills() on Lich attack

### LINKED ISSUE #6482: Infinite loop when redrawing the wince animation after a lich shot
### Preliminary checks

### Platform

Windows

### Describe the bug

* Load the attached savefile;
* Attack Roxana by Myra;
* Arrange the units as shown on the following screenshot:

lich

* Make the Lich shoot at the Golems;
* Observe an infinite loop in the `Battle::Interface::RedrawActionWincesKills()`.

### Save file

[Lich_Test.zip](https://github.com/ihhub/fheroes2/files/10416162/Lich_Test.zip)

### Additional info

Looks like a regression after #6414. On commit b4383ecb there is no infinite loop while on commit b200ee96 it is observed.

Districh-ru:
close #6482

I changed the logic for checking that Lich cloud animation is finished.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3460,87 +3460,93 @@ void Battle::Interface::RedrawActionWincesKills( const TargetsInfo & targets, Un
     while ( le.HandleEvents( Game::isDelayNeeded( { Game::BATTLE_FRAME_DELAY } ) ) ) {
         CheckGlobalEvents( le );
 
-        if ( Game::validateAnimationDelay( Game::BATTLE_FRAME_DELAY ) ) {
-            bool redrawBattleField = false;
+        if ( !Game::validateAnimationDelay( Game::BATTLE_FRAME_DELAY ) ) {
+            continue;
+        }
 
-            if ( attacker != nullptr ) {
-                if ( attacker->isFinishAnimFrame() ) {
-                    attacker->SwitchAnimation( Monster_Info::STATIC );
-                }
-                else {
-                    attacker->IncreaseAnimFrame();
-                }
+        bool redrawBattleField = false;
 
-                redrawBattleField = true;
+        if ( attacker != nullptr ) {
+            if ( attacker->isFinishAnimFrame() ) {
+                attacker->SwitchAnimation( Monster_Info::STATIC );
             }
             else {
-                for ( const Battle::TargetInfo & target : targets ) {
-                    if ( target.defender ) {
-                        redrawBattleField = true;
-                        break;
-                    }
-                }
+                attacker->IncreaseAnimFrame();
             }
 
-            if ( redrawBattleField ) {
-                RedrawPartialStart();
-
-                if ( drawLichCloud ) {
-                    // Draw a Lich cloud above the target unit.
-                    const fheroes2::Sprite & spellSprite = fheroes2::AGG::GetICN( ICN::LICHCLOD, lichCloudFrame );
-                    const fheroes2::Point & pos = CalculateSpellPosition( *defender, ICN::LICHCLOD, spellSprite );
-                    fheroes2::Blit( spellSprite, _mainSurface, pos.x, pos.y, false );
-                    ++lichCloudFrame;
-                }
-
-                RedrawPartialFinish();
-            }
-
+            redrawBattleField = true;
+        }
+        else {
             for ( const Battle::TargetInfo & target : targets ) {
                 if ( target.defender ) {
-                    if ( target.defender->isFinishAnimFrame()
-                         && ( target.defender->GetAnimationState() == Monster_Info::WNCE || target.defender->GetAnimationState() == Monster_Info::WNCE_DOWN ) ) {
-                        target.defender->SwitchAnimation( Monster_Info::STATIC );
-                    }
-                    else if ( drawLichCloud && lichCloudFrame == wnceUpStartFrame && ( target.defender->GetAnimationState() == Monster_Info::STAND_STILL ) ) {
-                        target.defender->SwitchAnimation( Monster_Info::WNCE_UP );
-                        AudioManager::PlaySound( target.defender->M82Wnce() );
-                    }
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:
Hi @Branikolog , please priotirize this pull request over others as this is a critical bug.

Branikolog (approved):

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3518,29 +3537,8 @@ void Battle::Interface::RedrawActionWincesKills( const TargetsInfo & targets, Un
...
-                    }
-                    else {
-                        target.defender->IncreaseAnimFrame();
-                    }
-                }
-            }
-
-            if ( finishedAnimation && ( !drawLichCloud || ( lichCloudFrame == lichCloudMaxFrame ) ) ) {
-                // All unit frames are rendered and if it was a Lich attack also its cloud frames are rendered too.
+            if ( ( animatingTargets == finishedAnimationCount ) && ( !drawLichCloud || ( lichCloudFrame >= lichCloudMaxFrame ) ) ) {
```
oleg-derevenetz, line 3540: I believe that `lichCloudFrame` should not under any circumstances become more than `lichCloudMaxFrame`, right? It doesn't make any sense and this condition looks suspicious.

I suggest adding an assertion somewhere at the beginning (or maybe even better at the end) of the `HandleEvents()` loop that will check that `animatingTargets` is still less than `finishedAnimationCount` and `lichCloudFrame` is still less than `lichCloudMaxFrame`.

Also, although I could not add a comment to the corresponding line, I suggest turning

```cpp
while (...) {
  if ( Game::validateAnimationDelay( Game::BATTLE_FRAME_DELAY ) ) {
    [... long loop body ...]
  }
}
```

into

```cpp
while (...) {
  if ( !Game::validateAnimationDelay( Game::BATTLE_FRAME_DELAY ) ) {
    continue;
  }
  [... long loop body ...]
}
```

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left a few comments, could you please take a look?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi @oleg-derevenetz.
The bug https://github.com/ihhub/fheroes2/issues/6482 was because in logic we first checked the `finishedAnimationCount` and after that changed some animations state including setting `WNCE_DOWN` animation to `STATIC` (finish) state.
Also `lichCloudFrame` could be more than `lichCloudMaxFrame` for some long death animations (like Phoenix and Bone Dragon). This was another my mistake in previous logic.

Districh-ru:
Hi @Branikolog, could you, please check once more the recent changes also checking the Lich attack for a couple of creatures with long wince animations (Centaur, Roc), long death animation (Bone Dragon, Medusa), and with short animations (Peasant, Minotaur).
I did some test in Battle Only mode and now the animations must be correct and there should not be infinite loops.

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: > Also `lichCloudFrame` could be more than `lichCloudMaxFrame` for some long death animations (like Phoenix and Bone Dragon). This was another my mistake in previous logic.

Under what conditions exactly can this happen? As far as I can see, `lichCloudFrame` is incremented only in one place - at line 3496, but there is a condition `drawLichCloud && lichCloudFrame < lichCloudMaxFrame`, so if `lichCloudFrame` is already equal to the `lichCloudMaxFrame`, there should be no increment, and therefore `lichCloudFrame ` could not become more than `lichCloudMaxFrame`, isn't it?

Districh-ru: > > Also `lichCloudFrame` could be more than `lichCloudMaxFrame` for some long death animations (like Phoenix and Bone Dragon). This was another my mistake in previous logic.
> 
> Under what conditions exactly can this happen? As far as I can see, `lichCloudFrame` is incremented only in one place - at line 3496, but there is a condition `drawLichCloud && lichCloudFrame < lichCloudMaxFrame`, so if `lichCloudFrame` is already equal to the `lichCloudMaxFrame`, there should be no increment, and therefore `lichCloudFrame ` could not become more than `lichCloudMaxFrame`, isn't it?

Yes, you are right, I've added this new condition so now `lichCloudFrame`  should not be higher than `lichCloudMaxFrame`.
I've changed the condition to `( !drawLichCloud || ( lichCloudFrame == lichCloudMaxFrame )` and added an assertion to check that `lichCloudFrame` is less or equal to `lichCloudMaxFrame`.

Branikolog (approved):
Checked all creatures (choose only 1from upgraded and regular). All wincing and dying animations doesn't have issues with lich cloud.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3460,89 +3460,96 @@ void Battle::Interface::RedrawActionWincesKills( const TargetsInfo & targets, Un
...
+            return TargetInfo::isFinishAnimFrame( info );
+        } );
+
+        // There sould not be more finished animations than we started.
+        assert( finishedAnimationCount <= animatingTargets );
+
+        // There sould not be more Lich cloud animation frames than in corresponding ICN.
+        assert( lichCloudFrame <= lichCloudMaxFrame );
+
+        // Important: 'lichCloudFrame' can be more than 'lichCloudMaxFrame' when performing some long 'KILL' animations.
```
oleg-derevenetz, line 3549: Hi @Districh-ru please remove this comment.

oleg-derevenetz (changes requested):

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi @oleg-derevenetz, did it.
Thanks, forgot to remove it in previous commit.

oleg-derevenetz (approved):

ihhub:
@Districh-ru , thank you so much for this quick fix!

Issue #1625, opened 2020-09-10
Combat, add the "Game: Battle show spell info" option


LeHerosInconnu:
In fheroes2, there is an option to display creature damage during combat in the information box.
It would be nice to add an option that does the same for spells "Game: Battle show spell info".

Additional information is also displayed in the information box.

To indicate if the troop has already played for the current turn:
"Turn done: Yes" or "Turn done: No" or "Turn done: Wait" (displayed only when the option "battle: soft wait troop" is activated).

To indicate if the troop has already used its retaliation for the current turn:
"Retaliation: Yes" or "Retaliation: No".

Here are some screens I used for the previous fheroes2 project to explain how it works.
Not all information is shown in the information box.
These screens are for explanatory purposes only.

For damage spells:

Damage: 400
Perish: 10
(spacing line)
Turn done: Yes
(spacing line)
Retaliation: Yes

![Heroes 2 Screen 75](https://user-images.githubusercontent.com/43583994/92737151-2fff2a00-f37b-11ea-920b-c834d52a50c0.png)

This also works for damage spells with an area of effect.
The information box is displayed for all affected creatures, allied and enemies:

For enemy rogues:

Damage: 30
Perish: 7
(spacing line)
Turn done: Yes
(spacing line)
Retaliation: Yes

For allied sprites:

Damage: 30
Perish: 15
(spacing line)
Turn done: Yes
(spacing line)
Retaliation: Yes

![Heroes 2 Screen 77](https://user-images.githubusercontent.com/43583994/92738199-08f52800-f37c-11ea-8a30-22e7155640e1.png)

![Heroes 2 Screen 78](https://user-images.githubusercontent.com/43583994/92738402-32ae4f00-f37c-11ea-9ba8-f92efe792b7c.png)

For resurrection spells:
The information box is also displayed on the corpses of troops.
Creatures raised: 12/12: 12 rangers will be resurrected out of a maximum of 12 possible 

Troop Hit Points: 120
(spacing line)
Spell Hit Points: 600
Creatures raised: 12/12
(spacing line)
Turn done: No
(spacing line)
Retaliation: Yes

![Heroes 2 Screen 79](https://user-images.githubusercontent.com/43583994/92738447-3b9f2080-f37c-11ea-8ae4-3dd2cf7baadb.png)

Branikolog:
> The information box is displayed for all affected creatures, allied and enemies:

In this case battlefield looks like a mess.. Too much info windows appear. 
[...]

PR #6493, opened 2023-01-14
Combat, Adding tooltip damage info for single target attack spells

kfishster:
As mentioned in #1625, the super helpful damage info tooltip that shows the range of damage when attacking the enemy troop does not show anything when using spells. When using a single-target damage spell such as a Lightning Bolt, Arrow, or Coldray, the spell in the magic book shows you the exact damage it would deal, but when hovering over an enemy troop, you have to cancel the spell, check the troop's health to calculate how many troops will perish, and then go back to using the spell again. Additionally, the spellbook does not show the spell adjustments you might encounter when targeting special troops, for example when using a cold ray on an Air Elemental (cold spells have 2x damage on air type monsters).

This damage tooltip, however, will ignore defending hero's artifacts - if the defending hero has an artifact reducing the damage of certain spells (such as the Ice Heart to reduce cold spell damage by half), the tooltip should not show that, as that is not known to the attacking hero.

_Before:_
image

_After:_
image

Interesting open question - **should magic resistance be mentioned in the tooltip**? Spell will not do any damage 25% of the time on dwarves.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -6040,8 +6040,8 @@ Battle::PopupDamageInfo::PopupDamageInfo()
     : Dialog::FrameBorder( 5 )
     , _cell( nullptr )
     , _defender( nullptr )
-    , _minDamage ( 0 )
-    , _maxDamage ( 0 )
+    , _minDamage( 0 )
+    , _maxDamage( 0 )
     , _redraw( false )
 {}
 
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -6086,12 +6086,15 @@ void Battle::PopupDamageInfo::SetAttackInfo( const Cell * cell, const Unit * att
 
 void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const HeroBase * hero, const Unit * defender, const Spell spell )
 {
-    if ( hero == nullptr || !SetDamageInfoBase( cell, defender ) ) {
+    assert( hero != nullptr );
+
+    // TODO: Currently, this functionality only supports a simple single-target spell case
+    // We should refactor this to apply to all cases
+    if ( !spell.isSingleTarget() || !spell.isDamage() ) {
         return;
     }
 
-    // Only show popup for single target damage spells such as bolt, arrow, or cold ray
-    if ( !spell.isSingleTarget() || !spell.isDamage() ) {
+    if ( !SetDamageInfoBase( cell, defender ) ) {
         return;
     }
 
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -6101,7 +6101,7 @@ void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const Unit 
     }
 
     const int spellPoints = hero ? hero->GetPower() : DEFAULT_SPELL_DURATION;
-    const uint32_t spellDamage = defender->CalculateDamage( spell, (uint32_t)spellPoints, hero, 0 /* targetInfo damage */, true /* ignore defending hero */ );
+    const uint32_t spellDamage = defender->CalculateDamage( spell, (uint32_t)spellPoints, hero, (uint32_t)0 /* targetInfo damage */, true /* ignore defending hero */ );
 
     _redraw = true;
     _minDamage = spellDamage;
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -2830,7 +2830,7 @@ void Battle::Interface::HumanCastSpellTurn( const Unit & /*b*/, Actions & a, std
 
         const Cell * cell = Board::GetCell( index_pos );
         if ( cell && _currentUnit && cell->GetUnit() ) {
-            popup.SetSpellAttackInfo( cell, _currentUnit, cell->GetUnit(), humanturn_spell );
+            popup.SetSpellAttackInfo( cell, _currentUnit->GetCurrentOrArmyCommander(), cell->GetUnit(), humanturn_spell );
         }
         else {
             popup.Reset();
@@ -6040,6 +6040,8 @@ Battle::PopupDamageInfo::PopupDamageInfo()
     : Dialog::FrameBorder( 5 )
     , _cell( nullptr )
     , _defender( nullptr )
+    , _minDamage ( 0 )
+    , _maxDamage ( 0 )
     , _redraw( false )
 {}
 
@@ -6048,9 +6050,9 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
     _battleUIRect = battleUIRect;
 }
 
-bool Battle::PopupDamageInfo::SetDamageInfoBase( const Cell * cell, const Unit * attacker, const Unit * defender )
+bool Battle::PopupDamageInfo::SetDamageInfoBase( const Cell * cell, const Unit * defender )
 {
-    if ( cell == nullptr || attacker == nullptr || defender == nullptr ) {
+    if ( cell == nullptr || defender == nullptr ) {
         return false;
     }
 
@@ -6066,7 +6068,7 @@ bool Battle::PopupDamageInfo::SetDamageInfoBase( const Cell * cell, const Unit *
 
 void Battle::PopupDamageInfo::SetAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender )
 {
-    if ( !SetDamageInfoBase( cell, attacker, defender ) ) {
+    if ( attacker == nullptr || !SetDamageInfoBase( cell, defender ) ) {
         return;
     }
 
@@ -6082,26 +6084,24 @@ void Battle::PopupDamageInfo::SetAttackInfo( const Cell * cell, const Unit * att
     }
 }
 
-void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender, const Spell spell )
+void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const HeroBase * hero, const Unit * defender, const Spell spell )
 {
-    if ( !SetDamageInfoBase( cell, attacker, defender ) ) {
+    if ( hero == nullptr || !SetDamageInfoBase( cell, defender ) ) {
         return;
     }
 
     // Only show popup for single target damage spells such as bolt, arrow, or cold ray
-    if ( !( spell.isSingleTarget() && spell.isDamage() ) ) {
+    if ( !spell.isSingleTarget() || !spell.isDamage() ) {
         return;
     }
 
-    const HeroBase * hero = attacker->GetCurrentOrArmyCommander();
-
-    // if defender unit immune to magic, do not show the tooltip
+    // If defender unit immune to magic, do not show the tooltip
     if ( !defender->AllowApplySpell( spell, hero ) ) {
         return;
     }
 
     const int spellPoints = hero ? hero->GetPower() : DEFAULT_SPELL_DURATION;
-    const uint32_t spellDamage = defender->CalculateDamage( spell, spellPoints, hero, 0 /* targetInfo damage */, true /* ignore defending hero */ );
+    const uint32_t spellDamage = defender->CalculateSpellDamage( spell, spellPoints, hero, 0 /* targetInfo damage */, true /* ignore defending hero */ );
 
     _redraw = true;
     _minDamage = spellDamage;
@@ -6131,8 +6131,8 @@ void Battle::PopupDamageInfo::Redraw() const
 
     std::string str = _minDamage == _maxDamage ? _( "Damage: %{max}" ) : _( "Damage: %{min} - %{max}" );
 
-    StringReplace( str, "%{min}", (int)_minDamage );
-    StringReplace( str, "%{max}", (int)_maxDamage );
+    StringReplace( str, "%{min}", _minDamage );
+    StringReplace( str, "%{max}", _maxDamage );
 
     Text damageText( str, Font::SMALL );
 
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -2777,7 +2777,7 @@ void Battle::Interface::HumanBattleTurn( const Unit & b, Actions & a, std::strin
         if ( cell ) {
             if ( CursorAttack( themes ) ) {
                 const Unit * b_enemy = cell->GetUnit();
-                popup.SetInfo( cell, _currentUnit, b_enemy );
+                popup.SetInfo( cell, _currentUnit, b_enemy, nullptr );
             }
             else
                 popup.Reset();
@@ -2827,6 +2827,20 @@ void Battle::Interface::HumanCastSpellTurn( const Unit & /*b*/, Actions & a, std
 
         if ( cursor.Themes() != themes )
             cursor.SetThemes( themes );
+
+        bool resetPopup = true;
+        const Cell * cell = Board::GetCell( index_pos );
+        if (cell && _currentUnit && humanturn_spell.isSingleTarget() && humanturn_spell.isDamage()) {
+            const Unit * b_enemy = cell->GetUnit();
+            if (b_enemy && b_enemy->AllowApplySpell( humanturn_spell, _currentUnit->GetCurrentOrArmyCommander() ))
+            {
+                popup.SetInfo( cell, _currentUnit, b_enemy, &humanturn_spell );
+                resetPopup = false;
+            }
+        }
+
+        if (resetPopup)
+            popup.Reset();
 
         if ( le.MouseClickLeft() && Cursor::WAR_NONE != cursor.Themes() ) {
             if ( !Board::isValidIndex( index_pos ) ) {
@@ -6041,9 +6055,9 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
     _battleUIRect = battleUIRect;
 }
 
-void Battle::PopupDamageInfo::SetInfo( const Cell * cell, const Unit * attacker, const Unit * defender )
+void Battle::PopupDamageInfo::SetInfo( const Cell * cell, const Unit * attacker, const Unit * defender, const Spell * spell )
 {
-    if ( cell == nullptr || attacker == nullptr || defender == nullptr ) {
+    if ( (cell == nullptr || attacker == nullptr || defender == nullptr) && (cell == nullptr || spell == nullptr || defender == nullptr) ) {
         return;
     }
 
@@ -6055,6 +6069,7 @@ void Battle::PopupDamageInfo::SetInfo( const Cell * cell, const Unit * attacker,
     _cell = cell;
     _attacker = attacker;
     _defender = defender;
+    _spell = spell;
 }
 
 void Battle::PopupDamageInfo::Reset()
@@ -6064,6 +6079,7 @@ void Battle::PopupDamageInfo::Reset()
         _cell = nullptr;
         _attacker = nullptr;
         _defender = nullptr;
+        _spell = nullptr;
     }
 
     Game::AnimateResetDelay( Game::BATTLE_POPUP_DELAY );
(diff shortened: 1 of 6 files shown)


LeHerosInconnu:
Hello @kfishster,

> This damage tooltip, however, will ignore defending hero's artifacts - if the defending hero has an artifact reducing the damage of certain spells (such as the Ice Heart to reduce cold spell damage by half), the tooltip should not show that, as that is not known to the attacking hero.

If this is applied it should only be for the first attack with the spell, because by the second attack with the spell (or the same type of spell, e.g. Lightning Bolt and Chain Lightning) the player has noticed that the damage of the spell is reduced.
And how to deal with this for another combat against the same opposing hero knowing that he has artifacts that reduce the damage of the same spells?
Maybe it would be easier to display the damage taking into account the artifacts, or whatever, of the opposing hero.
In my case, it would not bother me.

> Interesting open question - should magic resistance be mentioned in the tooltip? Spell will not do any damage 25% of the time on dwarves.

Both ways are fine for me.
The player will quickly learn that dwarves have a percentage chance of resisting bad spells (and this is indicated in the creature's info window).

kfishster:
@LeHerosInconnu thanks for the feedback. The way I think about this is that there is _public info_ and _private info_ when it comes to any combat scenario. The _public info_ is the kind of thing you can see at any point in combat, Attack/Defense skill, troop size, troop skills, etc. All these things are used to calculate your potential attack damage from a troop, and in the case of this PR, the total spell damage. The _private info_ is the "face down cards" that you or the opposing player will have, this is meant to be hidden, and if you make some sort of move that will uncover this information, you have to keep track of it yourself, it shouldn't be available in the same way as the _public info_. This also means that if you missed any of the clues or the _private info_ changed, you should not be basing any of the tooltips on that info. For example, if the opposing hero battles you the first time with an artifact that reduces damage, and then you face them again, there is no way for you to know whether they still have that artifact or not, so making that assumption on the basis of prior information would be incorrect.

So in this case, I would prefer to keep the behavior as implemented, codifying the _private info_ into the damage tooltip would take away a part of the Heroes experience.

LeHerosInconnu:
Hello @kfishster,

> For example, if the opposing hero battles you the first time with an artifact that reduces damage, and then you face them again, there is no way for you to know whether they still have that artifact or not, so making that assumption on the basis of prior information would be incorrect.

If the two combats are consecutive and take place in the same player's turn, or if the opposing hero is visible on the adventure map and has not encountered any other hero, it results that the opposing hero still possesses the same artifact, thus with the same results.

> So in this case, I would prefer to keep the behavior as implemented, codifying the _private info_ into the damage tooltip would take away a part of the Heroes experience.

I see your point, the problem is that incorrect information is displayed when it is known, and players will report this as a bug.
I am okay with doing it this way for now, and let's see how it's received. :)

kfishster:
@LeHerosInconnu I see your point, but I'd still argue that information should not be codified. In the same example, if I am about to attack an opposing hero and I see their army as "pack of wolves", "horde of peasants", if I attack, lose the battle, and am about to attack with a different hero, the game will still show me the "pack of wolves", "horde of peasants" display, not the exact numbers, although, as a player, I know the exact troop numbers.

> I see your point, the problem is that incorrect information is displayed when it is known, and players will report this as a bug.

Yea, this is a valid point. I was trying to find a similar example with attacking, where an enemy hero might have an artifact that reduces the damage they receive, but I don't believe it exists in HoMM2. Might be worth adding some documentation to the wiki on what this tooltip takes into account, or something of the sort.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6041,9 +6054,9 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
     _battleUIRect = battleUIRect;
 }
 
-void Battle::PopupDamageInfo::SetInfo( const Cell * cell, const Unit * attacker, const Unit * defender )
+void Battle::PopupDamageInfo::SetInfo( const Cell * cell, const Unit * attacker, const Unit * defender, const Spell * spell )
```
ihhub, line 6057: Would it be cleaner to just create a separate method for setting up spell information? I can see that the existing method is complex and we are adding even more logic on top of this. The existing method then could be renamed into `SetPhysicalDamageInfo()`.

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1317,14 +1317,25 @@ void Battle::Unit::SpellModesAction( const Spell & spell, uint32_t duration, con
...
+    uint32_t dmg = CalculateDamage( &spell, spoint, hero, target.damage, false /* ignore defending hero */ );
+
+    // apply damage
+    if ( dmg ) {
+        target.damage = dmg;
+        target.killed = ApplyDamage( dmg );
+    }
+}
+
+uint32_t Battle::Unit::CalculateDamage( const Spell * spell, uint32_t spoint, const HeroBase * hero, uint32_t target_damage, bool ignoreDefendingHero ) const
```
ihhub, line 1330: There is no reason to change from a reference to pointer. Please revert this change.

src/fheroes2/battle/battle_troop.h
```diff
@@ -194,6 +194,7 @@ namespace Battle
         void SetBlindAnswer( bool value );
         void SpellModesAction( const Spell &, uint32_t, const HeroBase * );
         void SpellApplyDamage( const Spell &, uint32_t, const HeroBase *, TargetInfo & );
+        uint32_t CalculateDamage( const Spell *, uint32_t, const HeroBase *, uint32_t, bool ) const;
```
ihhub, line 197: As per our coding style please add all input argument variables. The old code was written in not a very logical way.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6075,16 +6090,30 @@ void Battle::PopupDamageInfo::Redraw() const
         return;
     }
 
-    assert( _cell != nullptr && _attacker != nullptr && _defender != nullptr );
+    assert( ( _cell != nullptr && _defender != nullptr && _attacker != nullptr ) );
```
ihhub, line 6093: Could you please explain the reason why we need to change this line? It seems to me exactly the same as well as extra parenthesis was added.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6075,16 +6090,30 @@ void Battle::PopupDamageInfo::Redraw() const
...
+
+    uint32_t minDamage = 0;
+    uint32_t maxDamage = 0;
 
-    uint32_t minDamage = _attacker->CalculateMinDamage( *_defender );
-    uint32_t maxDamage = _attacker->CalculateMaxDamage( *_defender );
+    if ( _spell != nullptr && _spell->isSingleTarget() && _spell->isDamage() ) {
```
ihhub, line 6098: We should check spell conditions before setting it as in the proposed new method like `SetSpellDamageInfo()`.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6075,16 +6090,30 @@ void Battle::PopupDamageInfo::Redraw() const
...
+        const HeroBase * hero = _attacker->GetCurrentOrArmyCommander();
+        const uint32_t spoint = hero ? hero->GetPower() : DEFAULT_SPELL_DURATION;
```
ihhub, line 6100: Please name the variable according to our coding standard using `camelCase`. A name like **spellCost** or **spellPoints** would be much easier to read.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6075,16 +6090,30 @@ void Battle::PopupDamageInfo::Redraw() const
...
 
-    if ( _attacker->Modes( SP_BLESS ) ) {
-        minDamage = maxDamage;
+        uint32_t spellDmg = _defender->CalculateDamage( _spell, spoint, hero, 0 /* targetInfo damage */, true /* ignore defending hero */ );
```
ihhub, line 6102: This variable can be **const**.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2828,6 +2828,19 @@ void Battle::Interface::HumanCastSpellTurn( const Unit & /*b*/, Actions & a, std
         if ( cursor.Themes() != themes )
             cursor.SetThemes( themes );
 
+        bool resetPopup = true;
+        const Cell * cell = Board::GetCell( index_pos );
+        if ( cell && _currentUnit && humanturn_spell.isSingleTarget() && humanturn_spell.isDamage() ) {
```
ihhub, line 2833: I suggest to add spell checks in the proposed `SetSpellDamageInfo()` method to make the code cleaner.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2828,6 +2828,19 @@ void Battle::Interface::HumanCastSpellTurn( const Unit & /*b*/, Actions & a, std
...
+            const Unit * b_enemy = cell->GetUnit();
```
ihhub, line 2834: Please use camelCase variable naming. A name like **defenderUnit** is much better to read and understand.

ihhub (changes requested):
Hi @kfishster , I put several comments in this pull request. Could you please take a look at them?

Branikolog:
Hi, @kfishster !
It's a useful addition to the game UI! Showing how much creatures would be decimated after spell cast is definitely the information everyone would like to see during battles.
I'm ok with proposed implementation (without taking into consideration opposite hero's spell reducing artifacts), but what about golems, which has native -50% damage ability for some spells? I think, for golems in particular, info tooltip should display reduced damage, since this is permanent reduction.

src/fheroes2/battle/battle_interface.cpp
kfishster: Artifact of refactoring, reverting this line change!

src/fheroes2/battle/battle_interface.cpp
kfishster: Refactored this logic to set the damage via separate methods for normal attack and for spell attack

src/fheroes2/battle/battle_interface.cpp
kfishster: Refactored this logic to have a separate `SetInfo()` method for the two attack cases

kfishster:
> I'm ok with proposed implementation (without taking into consideration opposite hero's spell reducing artifacts), but what about golems, which has native -50% damage ability for some spells? I think, for golems in particular, info tooltip should display reduced damage, since this is permanent reduction.

@Branikolog, the current implementation accounts for the known unit modifiers such as golems taking half the damage or air elementals taking twice as much damage when hit with a cold spell. This reuses the same logic that calculates spell damage, but the only part that it ignores is the defending hero's artifact effects on spell damage.

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1317,6 +1317,17 @@ void Battle::Unit::SpellModesAction( const Spell & spell, uint32_t duration, con
...
+    uint32_t dmg = CalculateDamage( spell, spoint, hero, target.damage, false /* ignore defending hero */ );
+
+    // apply damage
+    if ( dmg ) {
+        target.damage = dmg;
+        target.killed = ApplyDamage( dmg );
+    }
+}
+
+uint32_t Battle::Unit::CalculateDamage( const Spell & spell, uint32_t spoint, const HeroBase * hero, uint32_t target_damage, bool ignoreDefendingHero ) const
```
ihhub, line 1330: Please use **camelCase** style of variable naming.

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1385,6 +1396,7 @@ void Battle::Unit::SpellApplyDamage( const Spell & spell, uint32_t spoint, const
     // check artifact
     if ( hero ) {
         const HeroBase * defendingHero = GetCommander();
+        bool useDefendingHeroArts = defendingHero && !ignoreDefendingHero;
```
ihhub, line 1399: This variable can be **const**.

src/fheroes2/battle/battle_troop.cpp
ihhub, line 1330: **CalculateDamage** is a very broad name to be used. For any reader it means that this method calculates damage while attacking which is not the case here. Something like `calculateSpellDamage` will be an accurate name for this method. What do you think?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6041,29 +6048,74 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
...
+}
+
+void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender, const Spell spell )
+{
+    if ( !SetDamageInfoBase( cell, attacker, defender ) ) {
+        return;
+    }
+
+    // Only show popup for single target damage spells such as bolt, arrow, or cold ray
+    if ( !( spell.isSingleTarget() && spell.isDamage() ) ) {
```
ihhub, line 6092: The opposite condition would be easier to read because **!( ... )** is usually ignored while fast reading:
```cpp
if ( !spell.isSingleTarget() || !spell.isDamage() ) {
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6041,29 +6048,74 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
...
+        return;
+    }
+
+    const HeroBase * hero = attacker->GetCurrentOrArmyCommander();
+
+    // if defender unit immune to magic, do not show the tooltip
```
ihhub, line 6098: Can we please start this comment from a capital letter while we do in other comments? This is to keep the consistency in the code.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6075,27 +6127,17 @@ void Battle::PopupDamageInfo::Redraw() const
...
-    }
+    assert( _cell != nullptr && _defender != nullptr );
 
-    std::string str = minDamage == maxDamage ? _( "Damage: %{max}" ) : _( "Damage: %{min} - %{max}" );
+    std::string str = _minDamage == _maxDamage ? _( "Damage: %{max}" ) : _( "Damage: %{min} - %{max}" );
 
-    StringReplace( str, "%{min}", minDamage );
-    StringReplace( str, "%{max}", maxDamage );
+    StringReplace( str, "%{min}", (int)_minDamage );
+    StringReplace( str, "%{max}", (int)_maxDamage );
```
ihhub, line 6135: Please do not use C-style casting in C++ code. Is there a need to even do casting here? If yes please use **static_cast**.

src/fheroes2/battle/battle_interface.h
```diff
@@ -232,15 +232,18 @@ namespace Battle
...
 
         void setBattleUIRect( const fheroes2::Rect & battleUIRect );
-        void SetInfo( const Cell * cell, const Unit * attacker, const Unit * defender );
+        void SetAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender );
+        void SetSpellAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender, const Spell spell );
         void Reset();
         void Redraw() const;
 
     private:
+        bool SetDamageInfoBase( const Cell * cell, const Unit * attacker, const Unit * defender );
```
ihhub, line 241: Please add an empty line after the declaration of this method to have visual separation of members and methods.

src/fheroes2/battle/battle_interface.h
```diff
@@ -232,15 +232,18 @@ namespace Battle
...
         fheroes2::Rect _battleUIRect;
         const Cell * _cell;
-        const Unit * _attacker;
         const Unit * _defender;
+        uint32_t _minDamage;
+        uint32_t _maxDamage;
```
ihhub, line 246: These members must be initialized in the constructor.

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1317,6 +1317,17 @@ void Battle::Unit::SpellModesAction( const Spell & spell, uint32_t duration, con
 }
 
 void Battle::Unit::SpellApplyDamage( const Spell & spell, uint32_t spoint, const HeroBase * hero, TargetInfo & target )
+{
+    uint32_t dmg = CalculateDamage( spell, spoint, hero, target.damage, false /* ignore defending hero */ );
```
ihhub, line 1321: This variable can be **const**.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6041,29 +6048,74 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
...
 
-    _redraw = true;
     _cell = cell;
-    _attacker = attacker;
     _defender = defender;
+
+    return true;
+}
+
+void Battle::PopupDamageInfo::SetAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender )
```
ihhub, line 6067: We don't use **attacker** anywhere. I believe we can simply remove this variable usage everywhere in **PopupDamageInfo** subsequently simplifying the code.

ihhub (changes requested):
Hi @kfishster , I left several more comments in this pull request. Could you please check them in your spare time?

src/fheroes2/battle/battle_interface.cpp
kfishster: This change was due to `clang-tidy` marking this as a narrow conversion (`StringReplace` requires an int, the variable is a `uint32_t`), although this could also be accomplished with an overloaded `StringReplace` that takes in a uint32_t, so I'll probably just do that

src/fheroes2/battle/battle_interface.cpp
kfishster: I removed the attacker from the shared logic, but it still needed for determining damage in the regular combat damage calculation. Let me know if this is simpler, I can certainly refactor even more by passing in only the attacker properties we need (such as whether attacker has a bless/curse spell and damage ranges), but I'd argue that makes things a bit more complex rather than simple.

src/fheroes2/battle/battle_troop.cpp
kfishster: Great point!

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6075,27 +6127,17 @@ void Battle::PopupDamageInfo::Redraw() const
...
-    }
+    assert( _cell != nullptr && _defender != nullptr );
 
-    std::string str = minDamage == maxDamage ? _( "Damage: %{max}" ) : _( "Damage: %{min} - %{max}" );
+    std::string str = _minDamage == _maxDamage ? _( "Damage: %{max}" ) : _( "Damage: %{min} - %{max}" );
 
-    StringReplace( str, "%{min}", minDamage );
-    StringReplace( str, "%{max}", maxDamage );
+    StringReplace( str, "%{min}", _minDamage );
+    StringReplace( str, "%{max}", _maxDamage );
```
ihhub, line 6135: If we want to avoid creating another overloaded version of the same function would this code work better:
```cpp
StringReplace( str, "%{min}", std::to_string( _minDamage ) );
StringReplace( str, "%{max}", std::to_string( _maxDamage ) );
```
?

We can create multiple copies of the same function but I think it might be a mess and harder to understand the code. What do you think?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6041,29 +6050,72 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
...
+}
+
+void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const HeroBase * hero, const Unit * defender, const Spell spell )
+{
+    if ( hero == nullptr || !SetDamageInfoBase( cell, defender ) ) {
+        return;
+    }
+
+    // Only show popup for single target damage spells such as bolt, arrow, or cold ray
+    if ( !spell.isSingleTarget() || !spell.isDamage() ) {
```
ihhub, line 6094: I see no big reason not to show for mass spells like Chain Lighting or Cold Ring. We should add such support in the future. Therefore I suggest to at least add a TODO comment explaining what is missing.

ihhub, line 6094: These checks should go before the first if where we call **!SetDamageInfoBase()** method. This method is computational heavy.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6041,29 +6050,72 @@ void Battle::PopupDamageInfo::setBattleUIRect( const fheroes2::Rect & battleUIRe
...
+        _minDamage = _maxDamage;
+    }
+    else if ( attacker->Modes( SP_CURSE ) ) {
+        _maxDamage = _minDamage;
+    }
+}
+
+void Battle::PopupDamageInfo::SetSpellAttackInfo( const Cell * cell, const HeroBase * hero, const Unit * defender, const Spell spell )
+{
+    if ( hero == nullptr || !SetDamageInfoBase( cell, defender ) ) {
```
ihhub, line 6089: If there a possibility that a hero who suppose to cast a spell is absent? If not we should just add an assertion for nullptr check for **hero** input argument.

ihhub (changes requested):
Hi @kfishster , I added just few more comments. We are almost there. Could you please take a look at them?

src/fheroes2/battle/battle_interface.cpp
kfishster: Overloading the `StringReplace` might be best for consistency. There is already an `int` overloaded method in `tools.cpp`, so we would have a situation where some classes are calling into the overload (although they could be just using `to_string` to use the main method). I would suggest keeping the `uint32_t` overload for consistency. Does that align with your thinking?

src/fheroes2/battle/battle_interface.cpp
kfishster: Will do. The multi-target spell problem is outlined in the initial issue #1625, it's mostly a UI issue - how do we display it effectively.

src/fheroes2/battle/battle_interface.cpp
kfishster: I'll add an assertion - I don't believe we can have a case where a spell is cast by a neutral troop without a hero/captain. Could you think of a scenario where that would happen?

ihhub: There is no such a scenario.

src/fheroes2/battle/battle_interface.cpp
ihhub: Overloading a function while changing **int32_t** to **uint32_t** is doing almost the same as just doing static_cast. The functions are identical and we know that we don't handle values over 2 billion so it is safe to do just casting.

But let's hear a third person's opinion to handle this properly. @oleg-derevenetz , what do you think?

oleg-derevenetz: Hi @ihhub I think that it is worth to just remove the overload with `int` completely and use the `std::to_string()` instead.

kfishster: Thanks for the feedback. I've removed the `uint32_t` overload for `StringReplace` and just used `std::to_string` for the string interpolation. Removing the `int` overload would be much trickier - there are about 191 invocations throughout code that use it, so it is best to keep it intact.

ihhub (approved):

oleg-derevenetz (approved):

ihhub:
@kfishster , thank you very much for this improvement!

PR #6494, opened 2023-01-15
Fix the AI resurrect spell logic

### LINKED ISSUE #6492: Combat screen, AI fails to cast Resurrect spell
### Preliminary checks

### Platform

Windows

### Describe the bug

It was really hard to catch and reproduce the bug.
Here's a guide to reproduce the bug. I recommend to repeat all steps, since depending on hero stats and creatures  the result may vary.

https://user-images.githubusercontent.com/55348946/212499987-bdf1c635-1159-4e37-9346-ad0a3b47caee.mp4

https://user-images.githubusercontent.com/55348946/212499994-9cf9715f-a918-4fdc-8d05-268d66ce3ce3.mp4

It was really hard to catch and reproduce the bug. Sorry for such long video.

AI it strying to use Resurrect spell but nothing happens.

### Save file

None.

### Additional info

_No response_

oleg-derevenetz:
fix #6492

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_spell.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_spell.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2021 - 2022                                             *
+ *   Copyright (C) 2021 - 2023                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
--- a/src/fheroes2/ai/normal/ai_normal_spell.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_spell.cpp
@@ -540,10 +540,6 @@ namespace AI
             const Unit * unit = arena.GraveyardLastTroop( idx );
             assert( unit != nullptr && !unit->isValid() );
 
-            if ( !unit->AllowApplySpell( spell, _commander ) ) {
-                continue;
-            }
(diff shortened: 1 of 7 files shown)


src/fheroes2/ai/normal/ai_normal_spell.cpp
```diff
@@ -528,6 +518,33 @@ namespace AI
...
+            if ( !unit->AllowApplySpell( spell, _commander ) ) {
+                continue;
+            }
+
+            updateBestOutcome( unit );
+        }
+
+        // Then consider the stacks from the graveyard
+        for ( const int32_t idx : arena.GraveyardOccupiedCells() ) {
+            const Unit * unit = arena.GraveyardLastTroop( idx );
```
ihhub, line 540: Should we allow to resurrect any unit on a cell, not only the last one?

ihhub (commented):
Hi @oleg-derevenetz , I left just one question. Could you please take a look at it?

src/fheroes2/ai/normal/ai_normal_spell.cpp
oleg-derevenetz: Hi @ihhub then behavior will be different from the behavior that is allowed for a human player, because human player is allowed to resurrect only the topmost unit, right?

ihhub: You're correct. Let's leave it as it is.

ihhub (approved):

Branikolog (approved):

ihhub:
@oleg-derevenetz , great thanks for this fix!

PR #6499, opened 2023-01-17
Don't use DBG_ENGINE in fheroes2 code

oleg-derevenetz:
Leave it for the engine itself and use the `DBG_GAME` instead.


ihhub:
@oleg-derevenetz , a bunch of thanks for this code correction!

PR #6503, opened 2023-01-18
SonarCloud: use the compile_commands.json instead of build wrapper

oleg-derevenetz:
Speed up the analysis, because now we do not need to perform a full build beforehand.


ihhub:
@oleg-derevenetz , great job! Now SonarQube analysis should come earlier for all builds.

PR #6505, opened 2023-01-19
Android toolset: add the ability to browse & delete save files

oleg-derevenetz:


oleg-derevenetz:
The icons used belong to the [Google Material Icons](https://developers.google.com/fonts/docs/material_icons?hl=en) set, which are licensed under the Apache 2.0 license. I'm not sure where to mention it better. Attribution is desirable but not mandatory.

ihhub:
> The icons used belong to the [Google Material Icons](https://developers.google.com/fonts/docs/material_icons?hl=en) set, which are licensed under the Apache 2.0 license. I'm not sure where to mention it better. Attribution is desirable but not mandatory.

Hi @oleg-derevenetz , we should mention the license like we do for all 3rd party libraries. We can put either a readme file under **android/app/src/main/res/** directory and specify while directories (`drawable-*`) belong to this license, or we could create a directory under the root called **third-lib-attribution** and put the file there.

Laserlicht:
@oleg-derevenetz Updated the translation for the new strings and created a pull request for you in your branch. Hope it works like this. 🙂

oleg-derevenetz:
Hi @Laserlicht I think it would be better to wait a bit with this, because something may still change. It would be better to create a separate PR after merging this one.

Laserlicht:
> Hi @Laserlicht I think it would be better to wait a bit with this, because something may still change. It would be better to create a separate PR after merging this one.

Okay for me. :)

ihhub (approved):

ihhub:
I suggest that we need at least an opinion of @Branikolog as he is usually the biggest critic of UX changes :)

Branikolog:
Hi, @oleg-derevenetz 
Sorry, for the late response. I couldn't test the PR due to some technical issues with my device. 
Save file manager works like a charm! Files could be deleted in two clicks. :+1: 
If it's possible, it would be nice to place of single and campaign files into two separate folders. Once we play a couple of games there would be a mess of files. So distribution them at least into two separate folders would help a bit to control files.
I'd also would like to be able to sort files according to file name or date created, but it seems to me, that I'm asking too much. :) Anyway, current implementation is definitely a step forwards! :+1: 
Between, during the installation fheroes2 on 13 Android device I've met a certain problem with copying Music folder. Toolset seems to ignore music folder and the only way for me to copy this folder manually was to connect phone to my PC.

ihhub:
Hi @oleg-derevenetz and @Branikolog , I suggest to make the proposed improvement in a separate PR to unlock this pull request being merged. What do you think?

oleg-derevenetz:
Hi @ihhub

> I suggest to make the proposed improvement in a separate PR to unlock this pull request being merged. What do you think?

I believe that further improvements could be made incrementally in separate PR(s).

Branikolog (approved):

ihhub:

Issue #6047, opened 2022-10-26
The game crashed when trying to pick up an artifact


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Other

### Describe the bug

A streamer found a crash when trying to pick up what I presume is an artifact on the map, but it looks to be hidden beneath some stones.

The stream was in Russian so maybe someone who speaks it will be able to understand better what might have caused it.
It happened shortly after 2h21m15s:
https://www.twitch.tv/videos/1634826358?t=02h21m15s

They are using the demo and the map is a downloaded one. It is possible to see the name of the map later in the video.

I'm not sure what OS they are using and I'm not entirely sure what fheroes2 version this is.

### Save file

I asked the streamer if they could share a save file here or on Discord.

zenseii:
Looking at 02:22:55. The name seems to be "01 Dраконы".

ihhub:
@zenseii , it looks like not the original map coming with the game.

zenseii:
> @zenseii , it looks like not the original map coming with the game.

I see. In the video it is also noticeable how there is a hero with no portrait. The streamer had many experimental options enabled - castle guardian, mine guardian, but I'm not sure this would explain such a behavior. My biggest guess is that there is some kind of bug carried over from the map editor in this map.

ihhub:
@zenseii , the hero picked a Magic Book artifact which must not be present at the map. I assume this is a modified by hex editor map. The crash is due to an assertion. I think we can reproduce this.

zenseii:
Ah well that is probably the reason for the crash. I thought maybe it was the Book of Elements when the streamer said book, but he did right click it so that would show what it was exactly, because he luckily had the scouting experimental option enabled.

ihhub:
@zenseii , my bad, it is `Ultimate Book of Knowledge` actually.

ihhub:
@zenseii , ultimate artifacts do not have any sprites so fading operation leads to a crash.

zenseii:
It should not be possible to place ultimate artifacts on the map with the editor so presumably the map creator hacked the map to make it possible.

I'm curious about how the original game would behave in this situation, but I'm not able to find this map.

[...]

PR #6507, opened 2023-01-21
Add checks to identify MX2 being loaded as MP2

ihhub:
Players should not rename maps as they want. If people want to play PoL maps they must own this version of the game.

This is a very basic check but it works on most of maps.

relates to #6047

### DIFF
--- a/src/fheroes2/world/world_loadmap.cpp
+++ b/src/fheroes2/world/world_loadmap.cpp
@@ -236,7 +236,7 @@ bool World::LoadMapMP2( const std::string & filename, const bool isOriginalMp2Fi
             case MP2::OBJ_EXPANSION_DWELLING:
             case MP2::OBJ_EXPANSION_OBJECT:
             case MP2::OBJ_JAIL:
-                DEBUG_LOG( DBG_GAME, DBG_INFO, "Failed to load The Price of Loyalty '" << filename << "' map which is not supported by this version of the game." )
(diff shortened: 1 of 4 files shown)


src/fheroes2/world/world_loadmap.cpp
```diff
@@ -226,6 +229,21 @@ bool World::LoadMapMP2( const std::string & filename )
...
         }
 
+        if ( checkPoLObjects ) {
+            switch ( mp2tile.mapObjectType ) {
+            case MP2::OBJ_BARRIER:
+            case MP2::OBJ_TRAVELLER_TENT:
+            case MP2::OBJ_EXPANSION_DWELLING:
+            case MP2::OBJ_EXPANSION_OBJECT:
+            case MP2::OBJ_JAIL:
+                DEBUG_LOG( DBG_GAME, DBG_INFO, "Failed to load The Price of Loyalty '" << filename << "' map which is not supported by this version of the game." )
```
oleg-derevenetz, line 239: Don't you think that "Failed to load The Price of Loyalty map MAP.MP2 which is not supported by this version of the game" looks better than "Failed to load The Price of Loyalty MAP.MP2 map which is not supported by this version of the game"?

oleg-derevenetz (commented):
Hi @ihhub I just left one suggestion related to aesthetics, could you please take a look?

src/fheroes2/world/world_loadmap.cpp
ihhub: It is corrected.

oleg-derevenetz (approved):

PR #6508, opened 2023-01-21
Fix loss of the first frame in the after attack unit animation. Battlefield.

### LINKED ISSUE #6465: Battlefield. Some units miss after attack animation.
### Preliminary checks

### Platform

All

### Describe the bug

Some units, which have only one frame of after attack animation miss it. 
Example:

Elves missing animation after an arrow was launched:

https://user-images.githubusercontent.com/55348946/210969354-a5dd4c3d-0731-4863-bdd8-2b9b6c2bb853.mp4

OG:

https://user-images.githubusercontent.com/55348946/210969496-5c2fd77e-2683-4b14-8b42-658b96f64811.mp4

_Originally posted by @Branikolog in https://github.com/ihhub/fheroes2/issues/6398#issuecomment-1373379579_
      

### Save file

Not needed, can be reproduced in "Battle only" mode.

### Additional info

It is because for 1 frame animation sequences current frame is always equal to the last frame and the engine decides the the animation is finished (always).
We may do a patch here or change the engine behavior with one frame animation sequences: so for the first time the engine will consider this frame as a current and the animation is not finished and after an attempt to increase the frame number the engine will continue using this "current" and only frame, but the animation will be considered as finished.

Districh-ru:
close #6465 
 
This PR fixes the after attack (melee or distance) animation.
Before this PR the engine did the change of the current frame to the next one (or changed the animation if it is finished) and then made a render. Now it first renders and then changes the frame/animation.
Also in this PR the redundant check for need of render at the start of the loop was removed as it is checked at the end of the loop.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3502,7 +3502,7 @@ void Battle::Interface::RedrawActionWincesKills( const TargetsInfo & targets, Un
     const uint32_t lichCloudMaxFrame = fheroes2::AGG::GetICNCount( ICN::LICHCLOD );
     // Wince animation under the Lich cloud, second part: the frame number after which the target animation will be switched to 'WNCE_UP'.
     const uint32_t wnceUpStartFrame = 1;
-    // Wince animation under the Lich cloud, third part: the frame number after which the target animation will be switched to 'WNCE_SOWN'.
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

ihhub (approved):

ihhub:

PR #6510, opened 2023-01-22
BattlePathfinder::evaluateForUnit(): minor polishing

oleg-derevenetz:
* Don't call redundant `operator=()`;
* In C++17 `operator[]()` is [guaranteed](https://en.cppreference.com/w/cpp/container/unordered_map/operator_at) to be the same as the `try_emplace().first->second`, but the former looks shorter.


ihhub:

PR #6514, opened 2023-01-23
Optimize mini-map rendering

Districh-ru:
An attempt to reduce CPU usage on mini-map (radar) rendering:

- [X] remove the use of "transform" image layer since mini-map is always non-transparent,
- [X] remove code for not square maps,
- [X] use direct mini-map image fill without many calls for `SetPixel` or `Fill` functions,
- [X] update the mini-map image only if it has to be changed,
- [x] update mini-map image only in Hero's scout area on Hero movement.

This PR changes the code of mini-map (radar) rendering so please test it with different Hero speeds, different movement directions when interaction with teleports and other active map elements (i.e. capture of castles, mines). These changes affects also the World View mini-map.

This PR also introduces:
- Reveal the area in Hero scout range after any change of scout range immediately after this change.

To test PR you can use this save: [221.zip](https://github.com/ihhub/fheroes2/files/10607734/221.zip)
It contains objects to test radar update: (Jail, Hut of Magi, Magelan Maps, Stone Lyth ...)


ihhub:
Hi @Districh-ru , if the last stage takes too much time we can do this is a separate pull request.

Districh-ru:
> Hi @Districh-ru , if the last stage takes too much time we can do this is a separate pull request.

Hi @ihhub , I plan to push the commit with the last stage today evening. I've done setting ROI for radar using hero's `GetScoute()` and `GetCenter()` methods. I need to test it.
I don't like redrawing the whole radar on hero move. On big maps it takes up to 12-15 ms (much CPU time is used for drawing castles). Usung ROI might speed up radar rendering up to 10 times on hero move.
And this will be a first step, which might be also optimized in future PRs (redraw not ROI, but only the scouted points and captured objects).
Sorry, but on working days I don't have much spare time. :(

ihhub:
Hi @Districh-ru , no pressure. I just proposed a way to have some progress on your work but since you are going to update it soon it is fine to wait.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -1075,6 +1079,18 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
 
                             RedrawFocus();
 
+                            const int32_t & heroDirection = hero->GetDirection();
```
oleg-derevenetz, line 1082: Hi @Districh-ru this trick is called "temporary lifetime extension": `Heroes::GetDirection()` returns a temporary object, and then lifetime of this temporary object is extended by binding of a `const` lvalue or rvalue reference to this temporary. Usually lifetime of a temporary ends at the end of an expression (in this case at the terminating `;`), but by binding a reference this lifetime is "extended" and becomes approximately equal to lifetime of the reference - once the reference is out of scope, temporary object will be destroyed as well.

This technique was once popular, especially somewhere near C++98, because it performs a some sort of guaranteed copy elision: no copy constructor is called when a reference is created. But, since C++11 and especially C++17 we have [guaranteed copy elision](https://en.cppreference.com/w/cpp/language/copy_elision) in typical use cases.

Also this technique should be used with caution, because it is [very easy to use it incorrectly](https://godbolt.org/z/GheKPxjan). Uncomment the commented fragments and see what happens. I recommend paying special attention to the last two examples with the [range-based for](https://en.cppreference.com/w/cpp/language/range-for), because, as you can see in the range-based for's "Explanation" section, it uses the temporary lifetime extension under the hood.

While this technique is absolutely legal, I recommend NOT to use it (unless absolutely necessary) and count on copy elision instead. Also there is absolutely no need to use it for basic types, like in this case and other similar cases in this PR.

Districh-ru: Hi @oleg-derevenetz, many thanks for warning and the detailed explanation with an example!
I'll take this into account in future PRs.
And I want to clarify: in current code we have
``` CPP
const fheroes2::Point & destPos = Maps::GetPoint( dst_index );

...

fheroes2::Point Maps::GetPoint( const int32_t index )
{
    return fheroes2::Point( index % world.w(), index / world.w() );
}

```
Is it safer (better) to declare `const fheroes2::Point destPos = ...` ?

oleg-derevenetz: > Is it safer (better) to declare `const fheroes2::Point destPos = ...` ?

I would say this way: with modern C++ standards and compilers there is usually no gain in using temporary lifetime extension (except, probably, for some exotic cases). In this case `destPos` is guaranteed to be created in-place and no copies will be made. However, this is still a valid mechanism, and there is no need to rewrite all existing places where it is used (a lot of the fheroes2 code was written in pre-C++11 manner and there's nothing particularly wrong with that). I recommend simply not using this mechanism in new code unless absolutely necessary.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -1031,6 +1033,8 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
             break;
         }
 
+        const bool isJumpHeroSpeed = ( 10 == conf.HeroesMoveSpeed() );
```
ihhub, line 1036: Is there a reason why we created this variable since it is used only in one place?

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -182,67 +178,44 @@ void Interface::Radar::SetPos( int32_t ox, int32_t oy )
...
-            new_sz.width = area.width;
-            new_sz.height = ( worldHeight * area.width ) / worldWidth;
-            offset.x = 0;
-            offset.y = ( area.height - new_sz.height ) / 2;
-        }
-        else {
-            new_sz.width = area.width;
-            new_sz.height = area.height;
-        }
+void Interface::Radar::SetRedraw() const
```
ihhub, line 201: This method can be inlined.

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -182,67 +178,44 @@ void Interface::Radar::SetPos( int32_t ox, int32_t oy )
...
+}
 
-        fheroes2::Image resized( new_sz.width, new_sz.height );
-        fheroes2::Resize( spriteArea, resized );
-        spriteArea = std::move( resized );
-    }
+void Interface::Radar::SetMapRedraw()
+{
+    _needMapRedraw = true;
+    _roi = fheroes2::Rect( 0, 0, world.w(), world.h() );
```
ihhub, line 209: We can use C++11 initialization list which checks for implicit conversions:
```cpp
_roi = { 0, 0, world.w(), world.h() };
```

src/fheroes2/game/game_startgame.cpp
```diff
@@ -686,6 +686,7 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     if ( player->isAIAutoControlMode() ) {
                         radar.SetHide( false );
+                        radar.SetMapRedraw();
```
ihhub, line 689: This is actually a problem: **_roi** in `Interface::Radar` class is being updated only after an explicit call of **SetMapRedraw()** methods. This might lead to some issues in the future when somebody might add a code to redraw the mini-map but forgets to call **SetMapRedraw()**.

What if we set `_roi = { 0, 0, world.w(), world.h() };` at the end of `Interface::Radar::RedrawObjects()` method default to full image every time we render the mini-map. Yes, it might be a little bit slower (few more CPU cycles) but this might prevent from the described before mistakes.

What do you think?

src/fheroes2/gui/interface_radar.h
```diff
@@ -82,22 +85,24 @@ namespace Interface
...
-        fheroes2::Image spriteArea;
-        fheroes2::MovableSprite cursorArea;
-        fheroes2::Point offset;
-        bool hide;
-        bool _mouseDraggingMovement;
+        fheroes2::Image _map{ RADARWIDTH, RADARWIDTH };
+        fheroes2::MovableSprite _cursorArea;
+        fheroes2::Rect _roi;
+        double _zoom{ 1.0 };
+        bool _hide;
```
ihhub, line 103: It is better to be consistent. Please add a default value for this member as well.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -383,7 +400,17 @@ namespace
 
         hero.Move2Dest( dst );
 
-        I.GetGameArea().SetCenter( hero.GetCenter() );
+        // Clear previous hero position on radar.
+        I.GetRadar().SetMapRedraw( fromRoi );
+
+        I.Redraw( Interface::REDRAW_RADAR );
+
+        const fheroes2::Point & toPosition = hero.GetCenter();
```
ihhub, line 408: Let's be consistent and make this variable as const variable only (without &).

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -383,7 +400,17 @@ namespace
...
+        I.GetGameArea().SetCenter( toPosition );
+        const int32_t & scoutRange = hero.GetScoute();
```
ihhub, line 410: This is definitely not needed. All POD (plain old data) types must not be as const reference. Please make it as a normal variable.

src/fheroes2/gui/interface_radar.h
```diff
@@ -56,12 +57,14 @@ namespace Interface
 
         void SetPos( int32_t ox, int32_t oy ) override;
         void SetRedraw() const;
+        void SetMapRedraw();
+        void SetMapRedraw( const fheroes2::Rect & roi );
```
ihhub, line 61: Shall we rename it into something like `setRenderArea` or `setRenderRoi` or use a word refresh? Right now it is not exactly cleat what these methods do.

ihhub (changes requested):
Hi @Districh-ru , I put several comments in this pull request. Would you mind to take a look at them?

src/fheroes2/game/game_startgame.cpp
Districh-ru: Hi @ihhub, this is a leftover from one of optimization approaches. :)
I'll revert this.

src/fheroes2/game/game_startgame.cpp
Districh-ru: Sure, it'll be safer. And as we set the whole `_roi` here, I'll remove this from `SetMapRedraw()` and make this method in-lined.

src/fheroes2/gui/interface_radar.h
Districh-ru: OK, there will be `SetRenderWholeMap()` to render objects for entire mini-map in and `SetRenderArea(...)`.

src/fheroes2/gui/interface_radar.cpp
Districh-ru: In this method we call `SetRedraw(...)` method  from `Interface::Basic` and we cannot make it inlined.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2680,6 +2701,7 @@ void ActionToMagellanMaps( Heroes & hero, const MP2::MapObjectType objectType, i
             hero.setVisitedForAllies( dst_index );
 
             Interface::Basic & I = Interface::Basic::Get();
+            I.GetRadar().SetRenderWholeMap();
```
ihhub, line 2704: Do we need this call to be present since we reset ROI after each render?

Districh-ru: Hi @ihhub, this method besides setting the `_roi` to whole map also sets `_needMapRedraw` to `true`. So we need to call it if we want to update the map image part of radar. Without this call only the cursor position will be updated.
I've renamed this method to `SetRenderMap()`, removed setting of `_roi` as it is now set after every `RedrawObjects()` call. Also I made it inline and changed its comment a bit.

src/fheroes2/game/game_startgame.cpp
Districh-ru: `SetRenderWholeMap()` is called at the start of `HumanTurn()`.
I've added `_roi = { 0, 0, world.w(), world.h() };` to `Build()` method which is called at the start of scenario.
Now it works OK with map change.

ihhub (approved):

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -380,31 +346,55 @@ void Interface::Radar::RedrawObjects( int color, ViewWorldMode flags ) const
                 // Castles and Towns can be partially covered by other non-action objects so we need to rely on special storage of castle's tiles.
                 if ( visibleTile ) {
                     if ( !getCastleColor( fillColor, tile.GetCenter() ) ) {
-                        continue;
+                        // This is a visible tile and not covered by other objets, so fill it with the ground tile data.
```
oleg-derevenetz, line 349:
```suggestion
                        // This is a visible tile and not covered by other objects, so fill it with the ground tile data.
```

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1257,6 +1257,19 @@ int Heroes::GetScoute() const
                              + GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::HEROES ) + GetSecondaryValues( Skill::Secondary::SCOUTING ) );
 }
 
+fheroes2::Rect Heroes::GetScoutRoi( const bool noDirection /* = false */ ) const
```
oleg-derevenetz, line 1260: I suggest calling this parameter `ignoreDirection`, but this is optional.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -923,6 +924,12 @@ void ActionToBoat( Heroes & hero, int32_t dst_index )
     hero.FadeOut( offset );
     hero.ResetMovePoints();
     hero.Move2Dest( dst_index );
+
+    // Update radar image without checking for the direction as the boat may have different direction.
```
oleg-derevenetz, line 928: It seems that the direction of the hero is not ignored here, what does this comment mean?

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2078,11 +2085,18 @@ void ActionToTeleports( Heroes & hero, int32_t index_from )
     hero.GetPath().Hide();
     hero.FadeOut();
 
+    Interface::Basic & I = Interface::Basic::Get();
+    // Before entering a Teleport the Hero makes a move into it, so set the scout area of this move to update radar image.
```
oleg-derevenetz, line 2089: This is not quite true, because player may just hit the space button and re-use the Teleport (the same logic applies to the Whirlpool BTW). JFYI.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1257,6 +1257,19 @@ int Heroes::GetScoute() const
...
+{
+    const int32_t scoutRange = GetScoute();
+    const fheroes2::Point heroPosition = GetCenter();
+
+    if ( noDirection ) {
+        return { heroPosition.x - scoutRange, heroPosition.y - scoutRange, heroPosition.x + scoutRange + 1, heroPosition.y + scoutRange + 1 };
```
oleg-derevenetz, line 1266: How exactly does it work? `Rect` is `{ x, y, width, height }` and according to this logic (and to similar logic below) calculations of `width` & `height` depend on `heroPosition`, although they should depend only on `scoutRange`. Am I missing something?

oleg-derevenetz (commented):
Hi @Districh-ru I just left a few suggestions and questions, could you please take a look when you have some free time?

src/fheroes2/heroes/heroes.cpp
Districh-ru: Hi @oleg-derevenetz!
Making a new check of possible scout radius change I found that it might be drastically changed without clearing the fog (rescue from jail, capture artifact). And the fog is cleared only on the next move, so the mini-map cannot be correctly rendered using hero direction.
So I removed the code using direction data.

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: Thanks for pointing this out! This comment is left over from one of the previous approaches.
I've changed it.

src/fheroes2/heroes/heroes.cpp
Districh-ru: I used a not clear `roi` implementation. I was using `width` and `height` as the max coordinates of `roi`. It was a bit easier to use.
I've changed this to a regular `fheroes2::Rect` implementation. Now width is `width`.

Districh-ru:
> Hi, @Districh-ru.
> I ran a test of your changes and found that the mini map is not updated if you visit an observation tower. I'm blue player and visited the tower down to the left. 
> I don't know a map that has the cartographer but I wanted to check that too. I tested the Eye of the Magi and it worked as it should.

Hi, @zenseii!
Thanks for finding this bug!
I've made a fix for Observation Tower and also for Castle capture.

For cartographer (Magelan Maps) I render the whole map. I use this map for checks, it contains a cartographer: [221.zip](https://github.com/ihhub/fheroes2/files/10559289/221.zip)

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1257,6 +1263,20 @@ int Heroes::GetScoute() const
...
+    const int32_t scoutRange = GetScoute();
+    const fheroes2::Point heroPosition = GetCenter();
+
+    if ( ignoreDirection ) {
+        return { heroPosition.x - scoutRange, heroPosition.y - scoutRange, 2 * scoutRange + 1, 2 * scoutRange + 1 };
+    }
+
+    return { heroPosition.x - ( ( direction == Direction::RIGHT ) ? 1 : scoutRange ), heroPosition.y - ( ( direction == Direction::BOTTOM ) ? 1 : scoutRange ),
+             ( ( direction == Direction::LEFT || direction == Direction::RIGHT ) ? 0 : scoutRange ) + scoutRange + 1,
+             ( ( direction == Direction::TOP || direction == Direction::BOTTOM ) ? 0 : scoutRange ) + scoutRange + 1 };
```
oleg-derevenetz, line 1277: Hmm, suppose that `scoutRange` is 2 and hero is moving to the right:

```
v previous hero position
*>**
   ^ a newly discovered tile
0123
```

So we need to redraw 4 tiles beginning from tile 0, right? With this logic:

```
return { 1 - 1, ..., 0 + 2 + 1, ... };
```

So just 3 tiles will be redrawn instead of 4, right?

src/fheroes2/heroes/heroes.h
```diff
@@ -493,6 +496,11 @@ class Heroes final : public HeroBase, public ColorBase
     void FadeIn( const fheroes2::Point & offset = fheroes2::Point() ) const;
     void Scoute( const int tileIndex ) const;
     int GetScoute() const;
+
+    // Returns the area to render the minimap image around Hero's position in his scout range.
+    // For non-diagonal move the area is set only in move direction and one pixel behind (to clear Hero's previous position).
+    fheroes2::Rect GetScoutRoi( const bool ignoreDirection = false ) const;
```
oleg-derevenetz, line 502: Maybe call it `GetRadarScoutRoi()` to emphasize that this ROI is related to radar? Or maybe somehow emphasize that this ROI is measured in tiles (maybe in the describing comment) and therefore make this method more "generic" and not related to the radar only?

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -895,7 +900,10 @@ void ActionToCastle( Heroes & hero, int32_t dst_index )
             hero.GetKingdom().AddCastle( castle );
             world.CaptureObject( dst_index, hero.GetColor() );
             castle->Scoute();
-            Interface::Basic::Get().SetRedraw( Interface::REDRAW_CASTLES );
+            const int32_t scoutRange = static_cast<int32_t>( GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::CASTLE ) );
+            Interface::Basic & I = Interface::Basic::Get();
+            I.GetRadar().SetRenderArea( { castlePosition.x - scoutRange, castlePosition.y - scoutRange, 2 * scoutRange + 1, 2 * scoutRange + 1 } );
+            I.SetRedraw( Interface::REDRAW_CASTLES | Interface::REDRAW_RADAR );
```
oleg-derevenetz, line 906: What do you think about putting this code to a local lambda to avoid the code duplication?

src/fheroes2/heroes/heroes_meeting.cpp
```diff
@@ -650,5 +654,16 @@ void Heroes::MeetingDialog( Heroes & otherHero )
...
     armyCountBackgroundRestorerRight.reset();
     restorer.restore();
+
+    // If the scout area bonus is increased with the new artifact we reveal the fog and update the radar.
+    if ( hero1ScoutAreaBonus < bag_artifacts.getTotalArtifactEffectValue( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) {
+        Scoute( this->GetIndex() );
+        ScoutRadar();
+    }
+    if ( hero2ScoutAreaBonus < otherHero.GetBagArtifacts().getTotalArtifactEffectValue( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) {
+        Scoute( otherHero.GetIndex() );
```
oleg-derevenetz, line 664: IMHO this looks wrong. This call will scout tiles **_as-if_** the current hero is scouting, but while standing on the other hero's position, which is wrong, because the other hero may have different scouting radius.

oleg-derevenetz (commented):
Hi @Districh-ru I just left a few notes and questions, could you please take a look?

In general, of course, the complication of the radar API is somewhat alarming. Where before it was enough to call `iface.Redraw( Interface::REDRAW_RADAR )`, now we need a lot more calls, including proper ROI calculations, what can be very easily done wrong or completely forgotten. Maybe somehow it is possible to encapsulate this (maybe in the future PRs) with simpler API.

src/fheroes2/heroes/heroes.cpp
Districh-ru: Hi @oleg-derevenetz, yes, now radar calls are complicated as now radar API does 3 functions:
1. Render a new cursor position on every `iface.Redraw( Interface::REDRAW_RADAR )` without rendering the map
2. Set ROI and set bit to render the map in ROI on the next `iface.Redraw( Interface::REDRAW_RADAR )`
3. Render the whole map on the next `iface.Redraw( Interface::REDRAW_RADAR )`
We can separate `Interface::REDRAW_RADAR` into `Interface::REDRAW_RADAR_MAP` and `Interface::REDRAW_RADAR_CURSOR` to call them both from `iface.Redraw()`. And if ROI was not set before `iface.Redraw()` then the whole map will be rendered.

And about `scoutRange`: if it is 2, then: `return { 1 - 2, ..., 0 + 2 + 1, ...} and there will be 4 tiles : `-1, 0, 1, 2`
Currently the last tile number in ROI is the tile number after the last rendered tile (like vector.end() ). If it is complicated I can change it and remove extra `+ 1` in ROI logic.

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: Thanks for the suggestion!
I'll do it in the evening.

src/fheroes2/heroes/heroes_meeting.cpp
Districh-ru: Yes, it is wrong, it should be `otherHero.Scoute( otherHero.GetIndex() );`

src/fheroes2/heroes/heroes.cpp
Districh-ru: Yes, there is a bug in setting ROI. I'll fix this bug today in the evening.

Districh-ru, line 1277:
```suggestion
             ( ( direction == Direction::LEFT || direction == Direction::RIGHT ) ? 1 : scoutRange ) + scoutRange + 1,
             ( ( direction == Direction::TOP || direction == Direction::BOTTOM ) ? 1 : scoutRange ) + scoutRange + 1 };
```

Districh-ru:
Hi, @oleg-derevenetz, I've tried to simplify the radar API and also made a lambda function for Castle capture. This is my first lambda function. :)
Please check it when you have time.

Districh-ru:
Hi @oleg-derevenetz, I've tried to change the radar API to get rid of many sets. Now `REDRAW_RADAR` redraws mini-map fully or in a ROI if it is set with a cursor over it and `REDRAW_MAP_CURSOR` renders the previously generated mini-map with a cursor over it.
Could you please check the committed changes when you have time.

src/fheroes2/gui/interface_radar.h
```diff
@@ -55,13 +56,19 @@ namespace Interface
         Radar & operator=( const Radar & ) = delete;
 
         void SetPos( int32_t ox, int32_t oy ) override;
+
+        // Set the render the cursor over the previously generated radar map image.
+        // To redraw the radar map image and render a cursor over it call "Interface::Basic::Get().SetRedraw( Interface::REDRAW_RADAR );".
```
oleg-derevenetz, line 61: What do you think about passing an argument for this `SetRedraw()`? Something like this:

```cpp
void SetRedraw( const uint32_t redrawMode ) const
{
    // Only these values are allowed here
    assert( ( redrawMode & ~( REDRAW_RADAR_CURSOR | REDRAW_RADAR ) ) == 0 );

    _interface.SetRedraw( redrawMode  );
}
```

This should help to simplify many places where `Radar::SetRedraw()` is used because there will be no need to use the `Interface::Basic::Get()` boilerplate anymore.

src/fheroes2/gui/interface_events.cpp
```diff
@@ -498,7 +498,10 @@ void Interface::Basic::EventSwitchShowRadar() const
         }
         else {
             conf.SetShowRadar( true );
-            radar.SetRedraw();
+            Basic & I = Basic::Get();
+            // We have to force set ROI to full map as it could be previously set to hero scout area on the last move.
+            I.GetRadar().SetRenderArea( { 0, 0, world.w(), world.h() } );
+            I.SetRedraw( REDRAW_RADAR );
```
oleg-derevenetz, line 504: We are already inside the `Interface::Basic` here. We can either remove the `const` qualifier here and do the following:

```cpp
radar.SetRenderArea( { 0, 0, world.w(), world.h() } );
SetRedraw( REDRAW_RADAR );
```

or, if we change the `Radar::SetRedraw()`, we can keep the `const` and do the following:

```cpp
radar.SetRenderArea( { 0, 0, world.w(), world.h() } );
radar.SetRedraw( REDRAW_RADAR );
```

I suggest looking at the rest of the places where this can potentially be simplified.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -686,7 +686,7 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     if ( player->isAIAutoControlMode() ) {
                         radar.SetHide( false );
-                        radar.SetRedraw();
+                        Redraw( REDRAW_RADAR );
```
oleg-derevenetz, line 689: There we can delay the redraw, because we perform another `Redraw()` call two lines below. We can use either the `SetRedraw( REDRAW_RADAR )` or `radar.SetRedraw( REDRAW_RADAR )` if we rewrite the `Radar::SetRedraw()`.

oleg-derevenetz (commented):
Hi @Districh-ru Yes, it seems like many boilerplate were removed. Let's see if we can simplify the code even more. I left a few suggestions, could you please take a look when you have time?

src/fheroes2/gui/interface_events.cpp
Districh-ru: Hi, @oleg-derevenetz, many thanks for the suggestions!
I used the second variant here, but I could not keep `const` as `SetRenderArea()` is not a `const` method.

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -182,70 +177,41 @@ void Interface::Radar::SetPos( int32_t ox, int32_t oy )
...
-            offset.y = ( area.height - new_sz.height ) / 2;
-        }
-        else {
-            new_sz.width = area.width;
-            new_sz.height = area.height;
-        }
+void Interface::Radar::SetRedraw( const uint32_t redrawMode ) const
+{
+    // Only radar redraws are allowed here.
+    assert( redrawMode & ( REDRAW_RADAR_CURSOR | REDRAW_RADAR ) );
```
oleg-derevenetz, line 201: Hi @Districh-ru

The whole point of this check is to verify that ONLY `REDRAW_RADAR_CURSOR` and/or `REDRAW_RADAR` are passed as `redrawMode`. What if `redrawMode` will be equal to, say, `REDRAW_RADAR_CURSOR | REDRAW_BUTTONS`? In this case this `assert()` will not fail, while this one:

```cpp
assert( ( redrawMode & ~( REDRAW_RADAR_CURSOR | REDRAW_RADAR ) ) == 0 );
```

will fail, right?

Districh-ru: Hi @oleg-derevenetz, yes, you are absolutely right.
Thanks for explanation!

oleg-derevenetz (approved):
LGTM, but I suggest postponing the merge after the 1.0.1 release, because this release will be literally tomorrow(?), and this PR contains a lot of various changes, so we will have an additional month for their thorough testing until the next release. @ihhub what do you think?

oleg-derevenetz:
> because this release will be literally tomorrow(?)

Ah no, it's not tomorrow.

ihhub:
Hi @oleg-derevenetz , do you still suggest to postpone?

oleg-derevenetz:
@ihhub Well, there are still many reviewers left. Let's make a decision after the full review.

ihhub:
I am going to review every critical pull request today and tomorrow.

src/fheroes2/gui/interface_radar.h
```diff
@@ -55,13 +56,19 @@ namespace Interface
         Radar & operator=( const Radar & ) = delete;
 
         void SetPos( int32_t ox, int32_t oy ) override;
-        void SetRedraw() const;
+
+        // Set the render redraw flag: 'REDRAW_RADAR' - to redraw the radar map image fully or in ROI and render the cursor over it,
+        // 'REDRAW_RADAR_CURSOR' - to render the previously generated radar map image and the cursor over it.
+        void SetRedraw( const uint32_t redrawMode ) const;
+
+        // Set the "need" of render the radar map only in geven 'roi' on next radar Redraw call.
```
ihhub, line 64: Misspelled **geven** --> **given**. Also please add **the** before this word.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -980,6 +980,12 @@ bool Heroes::PickupArtifact( const Artifact & art )
 
     if ( isControlHuman() ) {
         std::for_each( assembledArtifacts.begin(), assembledArtifacts.end(), Dialog::ArtifactSetAssembled );
+        // If the scout area bonus is increased with the new artifact we update the radar.
+        const std::vector<fheroes2::ArtifactBonus> bonuses = fheroes2::getArtifactData( art.GetID() ).bonuses;
+        if ( std::find( bonuses.begin(), bonuses.end(), fheroes2::ArtifactBonus( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) != bonuses.end() ) {
```
ihhub, line 985: Please use `art.isArtifactBonusPresent( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE )` instead of this code.

Also note that there is a chance that assembled artifact could appear so we need to check for it as well. I suggest to go this way:
```cpp
bool isScoutingUpdateRequired = false;
if ( art.isArtifactBonusPresent( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) {
    isScoutingUpdateRequired = true;
}

for ( const auto & assembledArtifact : assembledArtifacts ) {
    if ( Artifact( assembledArtifact ._assembledArtifactID).isArtifactBonusPresent( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) {
        isScoutingUpdateRequired = true;
        break;
    }
}

if ( isScoutingUpdateRequired  ) {
    Scoute( this->GetIndex() );
    ScoutRadar();
}
```

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1287,6 +1295,12 @@ void ActionToWitchsHut( Heroes & hero, const MP2::MapObjectType objectType, int3
         else {
             hero.LearnSkill( skill );
 
+            // When Scouting skill is learned we reveal the fog and redraw the radar map image in a new sout area of the hero.
```
ihhub, line 1298: Misspelled **sout** --> **scout**.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1300,6 +1314,13 @@ void ActionToWitchsHut( Heroes & hero, const MP2::MapObjectType objectType, int3
     DEBUG_LOG( DBG_GAME, DBG_INFO, hero.GetName() )
 }
 
+void Heroes::ScoutRadar() const
+{
+    Interface::Basic & I = Interface::Basic::Get();
+    I.GetRadar().SetRenderArea( this->GetScoutRoi( true ) );
```
ihhub, line 1320: Please remove `this->`:
```cpp
I.GetRadar().SetRenderArea( GetScoutRoi( true ) );
```
is shorter.

src/fheroes2/heroes/heroes_meeting.cpp
```diff
@@ -650,5 +654,16 @@ void Heroes::MeetingDialog( Heroes & otherHero )
     armyCountBackgroundRestorerLeft.reset();
     armyCountBackgroundRestorerRight.reset();
     restorer.restore();
+
+    // If the scout area bonus is increased with the new artifact we reveal the fog and update the radar.
+    if ( hero1ScoutAreaBonus < bag_artifacts.getTotalArtifactEffectValue( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) {
+        Scoute( this->GetIndex() );
```
ihhub, line 660: Please remove `this->`:
```cpp
Scoute( GetIndex() );
```

ihhub (changes requested):
Hi @Districh-ru , I left multiple comments in this pull request. Could you please take a look when you have time?

src/fheroes2/dialog/dialog_system_options.cpp
```diff
@@ -431,7 +431,7 @@ namespace fheroes2
                 basicInterface.Reset();
 
                 // We need to redraw radar first due to the nature of restorers. Only then we can redraw everything.
-                basicInterface.Redraw( Interface::REDRAW_RADAR );
+                basicInterface.Redraw( Interface::REDRAW_RADAR_CURSOR );
                 basicInterface.Redraw( Interface::REDRAW_ALL );
```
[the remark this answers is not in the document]
oleg-derevenetz: @ihhub There was some issue with rendering if we did not render the radar first separately. There is a comment regarding this.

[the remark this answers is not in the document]
ihhub: No-no. What I suggest is this:
```cpp
basicInterface.Redraw( Interface::REDRAW_RADAR_CURSOR );
basicInterface.Redraw( Interface::REDRAW_ALL & ~( Interface::REDRAW_RADAR | Interface::REDRAW_RADAR_CURSOR ) );
```
so we don't render the radar twice.

src/fheroes2/gui/interface_radar.cpp
```diff
@@ -257,89 +223,89 @@ void Interface::Radar::Redraw()
     if ( !hideInterface || conf.ShowRadar() ) {
```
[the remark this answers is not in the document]
Districh-ru: Hi, @ihhub, the next condition is not the opposite here.
In the first condition we render the borders for mini-map only in "Hide Interface" mode and only if it is visible (`conf.ShowRadar()`).
The code in the second condition renders the mini-map if we are not in "Hide Interface" mode and mini-map is visible.
I'll change the code to this:
``` cpp
    if ( conf.isHideInterfaceEnabled() ) {
        if ( conf.ShowRadar() ) {
            BorderWindow::Redraw();
        }
        else {
            // We are in "Hide Interface" mode and radar is turned off so we have nothing to render.
            return;
        }
    }

... the rest of the code without if condition
```

src/fheroes2/dialog/dialog_system_options.cpp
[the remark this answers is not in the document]
Districh-ru: OK, I've tested this suggestion and slightly changed it:
```cpp
// We need to redraw radar first due to the nature of restorers. Only then we can redraw everything.
// And we do a full radar redraw as it could be hidden in "Hide Interface" mode so it was not updated.
basicInterface.Redraw( Interface::REDRAW_RADAR );
basicInterface.Redraw( Interface::REDRAW_ALL & ~( Interface::REDRAW_RADAR_CURSOR | Interface::REDRAW_RADAR ) );
```

src/fheroes2/gui/interface_radar.cpp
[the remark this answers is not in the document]
ihhub: Hi @Districh-ru , in this case please just cache **conf.ShowRadar()** and use in both places.

src/fheroes2/dialog/dialog_system_options.cpp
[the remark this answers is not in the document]
ihhub: Nice! A faster solution :)

src/fheroes2/game/game_interface.h
```diff
@@ -61,13 +61,16 @@ namespace Interface
...
     {
-        REDRAW_RADAR = 0x01,
-        REDRAW_HEROES = 0x02,
-        REDRAW_CASTLES = 0x04,
-        REDRAW_BUTTONS = 0x08,
-        REDRAW_STATUS = 0x10,
-        REDRAW_BORDER = 0x20,
-        REDRAW_GAMEAREA = 0x40,
+        // To render the cursor over the previously generated radar map image.
+        REDRAW_RADAR_CURSOR = 0x01,
```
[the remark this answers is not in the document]
Districh-ru: Both of these flags calls radar `Radraw()` method, with only difference:
`REDRAW_RADAR` redraws the mini-map pixel-by-pixel in a ROI (if it was set) or fully and renders a cursor over it
and `REDRAW_RADAR_CURSOR` restores the image under the cursor position and render the new position of cursor without redrawing the mini-map.
Without these 2 enumerations we have to make a method to set some `bool isMapRedrawNeeded` in radar class before `SetRedraw(  REDRAW_RADAR );` or `Redraw(  REDRAW_RADAR );` if we want to update the radar map. <- this was implemented earlier and it was more complicated radar API.

src/fheroes2/heroes/heroes.cpp
Districh-ru: `isArtifactBonusPresent` is the member of `BagArtifacts` class and we have `const Artifact & art`. So we can't use this method to `art`. That's why I used this code.
I'll add the same check if any artifact was assembled.

ihhub: Oh, yes, You're right.

Districh-ru: @ihhub, I've made a lambda-function to do checks for `art` and `assembledArtifact `. Could you please check it.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -980,6 +980,31 @@ bool Heroes::PickupArtifact( const Artifact & art )
...
+            return false;
+        };
+
+        // If the scout area bonus is increased with the new artifact we update the radar.
+        if ( scout( art.GetID() ) ) {
+            return true;
+        }
+
+        // If there were artifacts assembled we check them for scout area bonus.
+        if ( !assembledArtifacts.empty() ) {
```
ihhub, line 1001: This check is no needed as the for-loop will handle it. Please remove it.

ihhub (commented):
Hi @Districh-ru , I left one more suggestion. Could you please check it?

src/fheroes2/gui/interface_radar.cpp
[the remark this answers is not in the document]
Districh-ru: I removed `conf.ShowRadar()` from the second place as now there is no need to check it. We do all checks in shown above double `if` condition. In "Show Interface" mode the check of `conf.ShowRadar()` is not needed.

src/fheroes2/heroes/heroes.cpp
Districh-ru: Oh, yes, you are right. :)

ihhub (approved):

ihhub:
Hi @Branikolog and @zenseii , could you please re-review this pull request when you have spare time?

Districh-ru:
> Hi, @Districh-ru.
> 
> I got an assertion failure when visiting the observation tower on the Slugfest map (Small), but not on Broken Alliance (Medium) `at line 277 in interface_radar.cpp`:
> 
> ```c++
> assert( _roi.x >= 0 && _roi.y >= 0 && ( _roi.width + _roi.x ) <= world.w() && ( _roi.height + _roi.y ) <= world.h() );
> ```

Hi, @zenseii!
Thanks for finding this bug! As the scout area is too big (41 tiles width and height) it originally starts from `roi.x`=-1 and `roi.y`=-3. The code changes these values to 0, but when checking `roi.width` and `roi.height` the value correction uses non corrected `x` and `y`, so we get `roi.width`=36-(-1)=37 and `roi.heigh`=36-(-3)=39 and then we get an assertion.
I've fixed this issue and checked this map and the test map for scouting near the map edges.

Branikolog:
Hi, @Districh-ru 
I've noticed, when you buy a hero with improved scouting skill, the area around castle is not revealed.
Also, When you step out with this hero, fog is extra revealed only on a lower part of castle area.
image

Districh-ru:
> Hi, @Districh-ru I've noticed, when you buy a hero with improved scouting skill, the area around castle is not revealed. Also, When you step out with this hero, fog is extra revealed only on a lower part of castle area. 

Hi, @Branikolog. Thanks for bug-tracking!
I've updated the code. Now the map will be revealed (with the update of mini-map) with every possible hero recruitment.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -717,6 +718,10 @@ bool Heroes::Recruit( const int col, const fheroes2::Point & pt )
     // Update the set of recruits in the kingdom
     kingdom.GetRecruits();
 
+    // After recruiting a hero we reveal map and radar image in hero scout area.
+    Scoute( GetIndex() );
+    ScoutRadar();
```
oleg-derevenetz, line 723: Hi @Districh-ru what if this hero is not controlled by local human player (for example, this hero is hired by AI)?

Districh-ru: Hi, @oleg-derevenetz!
`Scoute()` calls `ClearFog()` and as we reveal the map for human player we should do it also for AI for equality. :)
But `ScoutRadar()` will only waste some resources, thanks for the note. So lets call it only for human players.

oleg-derevenetz: @Districh-ru yes, I meant the `ScoutRadar()` part, because there is no need to redraw the radar in this case.

Districh-ru:
> Hi, @Districh-ru.
> I've tested the changes and everything works. Please resolve the merge conflicts when you have the time to.

Hi, @zenseii.
I've resolved the merge conflicts and also did minor optimization available after this merge: before this merge the View world map was render on the whole screen so we had to render radar image over it every time. Now we can render radar image only once and then only redraw the cursor.

Branikolog (approved):

ihhub:
@Districh-ru , great thanks for this improvement!

PR #6516, opened 2023-01-24
Add support for scaled resolutions

ihhub:
Based on the demand from users with high resolution displays. This feature is available only on SDL 2.

Also fixed text alignment for resolution selection dialog.

relates to #5373
relates to #4946


src/fheroes2/dialog/dialog_resolution.cpp
```diff
@@ -225,13 +253,14 @@ namespace Dialog
...
-            RedrawInfo( roi.getPosition(), selectedResolution );
+            RedrawInfo( roi.getPosition(), selectedResolution, display );
             display.render();
         }
 
-        if ( selectedResolution.width > 0 && selectedResolution.height > 0
-             && ( selectedResolution.width != currentResolution.width || selectedResolution.height != currentResolution.height ) ) {
-            display.resize( selectedResolution.width, selectedResolution.height );
+        if ( selectedResolution.width > 0 && selectedResolution.height > 0 && selectedResolution.scale > 0
+             && ( selectedResolution.width != currentResolution.width || selectedResolution.height != currentResolution.height
```
oleg-derevenetz, line 261: Since `ResolutionInfo` already has `operator=()`, we can just add `operator!=()` and use it here and in other similar places.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions, please take a look when you have time. In addition, this `scale` is actually not really used anywhere yet, as far as I understand?

src/engine/screen.cpp
```diff
@@ -124,6 +128,42 @@ namespace
...
+        // Scaling is available only on SDL 2.
+        if ( resolutions.size() < 2 ) {
+            return resolutions;
+        }
+
+        // Add resolutions with scale factor. No need to run through the newly added elements so we remember the size of the array.
+        const size_t resolutionCountBefore = resolutions.size();
+
+        // Since all resolutions are sorted then the first resolution (which is the highest) cannot have any scale factor.
+        for ( size_t currentId = resolutionCountBefore - 1; currentId > 1; --currentId ) {
```
oleg-derevenetz, line 141: Suppose that we have two resolutions:

2x2 (1x1 with scale factor 2)
1x1

We will not be able to offer the first resolution as a scaled version of the second one because at the first iteration of this `for` loop `currentId` will be 1 and condition `currentId > 1` will be immediately `false`, right?

src/engine/screen.cpp
```diff
@@ -124,6 +128,42 @@ namespace
...
+            int32_t scaleFactor = 1;
+            bool newScaleFactorFound = true;
+            while ( newScaleFactorFound ) {
+                newScaleFactorFound = false;
+                ++scaleFactor;
```
oleg-derevenetz, line 146: Maybe alter the logic a bit here and just check is the another "bigger" resolution is divisible by the current "smaller" resolution (i.e.

```cpp
biggerResolution.width / smallerResolution.width == biggerResolution.height / smallerResolution.height && biggerResolution.width % smallerResolution.width == 0 && biggerResolution.height && smallerResolution.height == 0
```

? Also of course check that both `biggerResolution` and `smallerResolution` doesn't have zero width or height.

src/fheroes2/dialog/dialog_resolution.cpp
```diff
@@ -225,13 +253,12 @@ namespace Dialog
...
             buttonCancel.draw();
-            RedrawInfo( roi.getPosition(), selectedResolution );
+            RedrawInfo( roi.getPosition(), selectedResolution, display );
             display.render();
         }
 
-        if ( selectedResolution.width > 0 && selectedResolution.height > 0
-             && ( selectedResolution.width != currentResolution.width || selectedResolution.height != currentResolution.height ) ) {
-            display.resize( selectedResolution.width, selectedResolution.height );
+        if ( selectedResolution.width > 0 && selectedResolution.height > 0 && selectedResolution.scale > 0 && ( selectedResolution != currentResolution ) ) {
```
oleg-derevenetz, line 260: The inner parentheses looks redundant here.

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions and suggestions, could you please take a look when you have time?

Districh-ru (changes requested):
Hi @ihhub, I found one bug, could you please address it when you have time:
If I set the 2x resolution in windowed mode (i.e. `960x540(x2)`) and then switch to fullscreen mode by clicking the `Mode` icon (not F4), the engine switches to 960x540 without x2.
If I press F4 in windowed mode for the first time the engine correctly switches to 960x540(x2). But if I press F4 two more times the engine first switches to 960x540(x2) windowed and then to 960x540 fullscreen without x2.
Changing resolution without switching fullscreen/windowed mode works correctly.
I use Windows OS.

src/engine/screen.cpp
ihhub: Actually, you're right. This is bug. It should be **currentId > 0**.

src/engine/screen.cpp
```diff
@@ -124,6 +128,38 @@ namespace
...
+        }
+
+        // Add resolutions with scale factor. No need to run through the newly added elements so we remember the size of the array.
+        const size_t resolutionCountBefore = resolutions.size();
+
+        // Since all resolutions are sorted then the first resolution (which is the highest) cannot have any scale factor.
+        for ( size_t currentId = resolutionCountBefore - 1; currentId > 0; --currentId ) {
+            assert( resolutions[currentId].width > 0 && resolutions[currentId].height > 0 );
+
+            for ( size_t biggerId = currentId - 1;; ) {
```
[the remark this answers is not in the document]
ihhub: **size_t** is unsigned int. currentId is 1, biggerId can be and will be 0. Then after the iteration we decrement biggerId leading to overflow.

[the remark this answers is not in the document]
oleg-derevenetz: Oh, maybe not because of that case of just 2 resolutions.

src/engine/screen.cpp
```diff
@@ -124,6 +128,38 @@ namespace
...
+                assert( resolutions[biggerId].width > 0 && resolutions[biggerId].height > 0 );
+
+                if ( ( resolutions[biggerId].width % resolutions[currentId].width ) == 0 && ( resolutions[biggerId].height % resolutions[currentId].height ) == 0
+                     && ( resolutions[biggerId].width / resolutions[currentId].width ) == ( resolutions[biggerId].height / resolutions[currentId].height ) ) {
```
oleg-derevenetz, line 148: Hi @ihhub BTW is there a possibility that there will be 2 equal resolutions in the list? If yes, then we should handle this case somehow to avoid additional resolutions such as "640x480x1" :)

ihhub: It is not possible. This function receives a set of resolutions. Then we add resolutions which are smaller than the smallest in the set. And then we run through the whole list of resolutions adding the scaled ones. However, to be more accurate I will change **possibleResolutions** into a set.

ihhub: I think no harm to actually return std::set from this function to make sure that all resolutions are unique. What do you think?

oleg-derevenetz: Yes, I believe that this will look more logical.

ihhub: I pushed the change but it looks a bit ugly...

ihhub:
> Hi @ihhub, I found one bug, could you please address it when you have time: If I set the 2x resolution in windowed mode (i.e. `960x540(x2)`) and then switch to fullscreen mode by clicking the `Mode` icon (not F4), the engine switches to 960x540 without x2. If I press F4 in windowed mode for the first time the engine correctly switches to 960x540(x2). But if I press F4 two more times the engine first switches to 960x540(x2) windowed and then to 960x540 fullscreen without x2. Changing resolution without switching fullscreen/windowed mode works correctly. I use Windows OS.

Hi @Districh-ru , could you please test the latest changes in this pull request and see if this issue is gone?

Districh-ru:
> Hi @Districh-ru , could you please test the latest changes in this pull request and see if this issue is gone?

Hi, @ihhub, I've tested the last commit. The issue is gone, now the Fullscreen mode is correctly set for scaled resolutions.

Districh-ru (approved):

ihhub:
Hi @oleg-derevenetz , I addressed your concerns about uniqueness of elements as well I made extra optimization of the code. Please take a look again when you have time.

ihhub:
Hi @Branikolog , could you please test this pull request as the highest priority as we need this change for the upcoming release?

src/engine/screen.cpp
```diff
@@ -573,16 +601,18 @@ namespace
             return { VITA_FULLSCREEN_WIDTH, VITA_FULLSCREEN_HEIGHT };
         }
 
-        std::vector<fheroes2::Size> getAvailableResolutions() const override
+        std::vector<fheroes2::ResolutionInfo> getAvailableResolutions() const override
         {
-            static std::vector<fheroes2::Size> filteredResolutions;
+            static std::vector<fheroes2::ResolutionInfo> filteredResolutions;
```
oleg-derevenetz, line 606: We can make the `filteredResolutions` `static const` and use a lambda to initialize it once and for all. This initialization will be also guaranteed thread-safe (although this may not be so important here, but why not, if we can have the thread safety for free).

src/engine/screen.cpp
```diff
@@ -821,12 +850,12 @@ namespace
             return ( flags & SDL_WINDOW_FULLSCREEN ) != 0 || ( flags & SDL_WINDOW_FULLSCREEN_DESKTOP ) != 0;
         }
 
-        std::vector<fheroes2::Size> getAvailableResolutions() const override
+        std::vector<fheroes2::ResolutionInfo> getAvailableResolutions() const override
         {
-            static std::vector<fheroes2::Size> filteredResolutions;
+            static std::vector<fheroes2::ResolutionInfo> filteredResolutions;
```
oleg-derevenetz, line 855: Here we also can make the `filteredResolutions` `static const` and initialize it using a lambda.

src/engine/screen.cpp
```diff
@@ -1321,20 +1352,22 @@ namespace
             return ( ( _surface->flags & SDL_FULLSCREEN ) != 0 );
         }
 
-        std::vector<fheroes2::Size> getAvailableResolutions() const override
+        std::vector<fheroes2::ResolutionInfo> getAvailableResolutions() const override
         {
-            static std::vector<fheroes2::Size> filteredResolutions;
+            static std::vector<fheroes2::ResolutionInfo> filteredResolutions;
```
oleg-derevenetz, line 1357: Here we also can make the `filteredResolutions` `static const` and initialize it using a lambda.

src/fheroes2/system/settings.cpp
```diff
@@ -380,8 +384,13 @@ std::string Settings::String() const
...
     os << "# fheroes2 configuration file (saved by version " << GetVersion() << ")" << std::endl;
 
+    const fheroes2::Display & display = fheroes2::Display::instance();
+
     os << std::endl << "# video mode (game resolution)" << std::endl;
-    os << "videomode = " << fheroes2::Display::instance().width() << "x" << fheroes2::Display::instance().height() << std::endl;
+    os << "videomode = " << display.width() << "x" << display.height() << std::endl;
+
+    os << std::endl << "# screen scale ratio" << std::endl;
+    os << "screen scale ratio = " << display.scale() << std::endl;
```
oleg-derevenetz, line 393: What do you think about integrating this into `videomode`? Say, there will be two formats of this setting: good old `videomode = 640x480` and the new one `videomode = 640x480x2`?

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions and suggestions, could you please take a look?

Branikolog:
Hi, @ihhub 
Everything works stable.
A few things I don't get...
This PR adds an ability to double only one resolution for my device:
image
My monitor is 1080p, but for some reason none of custom resolutions are available for being x2.
Also, while choosing x2 resolution in window mode or even full screen I have blurred image:
image
I suppose, this feature was made to create a precise x2 picture increase, but currently it works the same to software cursor mode with linear filtering as if I manually resize game window....
Sorry, if I understood somethig wrong...

oleg-derevenetz:
> My monitor is 1080p, but for some reason none of custom resolutions are available for being x2.

640x480x2 resolution is 1280x960, but with in-game picture linearly scaled x2, so GUI elements doesn't look small, like with just 1280x960. To have, say, 800x600x2, you should have 1600x1200 in the list of supported resolutions, and to have 928x520x2 you should have 1856x1040 in the list of supported resolutions. Do you have them in this list?

> Also, while choosing x2 resolution in window mode or even full screen I have blurred image

Yes, it was to be expected because of upscaling.

> I suppose, this feature was made to create a precise x2 picture increase, but currently it works the same to software cursor mode with linear filtering as if I manually resize game window....

There is still one of SDL upscaling algorithms is applied. It's in fact really as-if you resized your window x2 linearly in both dimensions.

Districh-ru:
> Hi, @ihhub Everything works stable. A few things I don't get... This PR adds an ability to double only one resolution for my device: My monitor is 1080p, but for some reason none of custom resolutions are available for being x2. Also, while choosing x2 resolution in window mode or even full screen I have blurred image: I suppose, this feature was made to create a precise x2 picture increase, but currently it works the same to software cursor mode with linear filtering as if I manually resize game window.... Sorry, if I understood somethig wrong...

Hi, @Branikolog.
Interesting, I have also 1080p display, but fheroes2 allowed me to use two 2x resolutions:
изображение
As I know, this depends on resolutions supported by graphics card. Some drivers allow to set custom resolutions. After I added new resolutions fheroes2 allowed me to use them:
изображение

(There would be good in future to have non integer scales, for example set in-game resolution to 1366x768 and upscale it to fullscreen 1920x1080.)

Branikolog:
Hi, @oleg-derevenetz !

> Do you have them in this list?

That's a valid point. I've added this custom resolution and one more x2 option appeared for me. :+1: 

>There is still one of SDL upscaling algorithms is applied. It's in fact really as-if you resized your window x2 linearly in both dimensions.

Well, I've expected, if I set 640x480 x2 (window more) I would get a clear image with each game pixel include 2 pixels of my monitor. But it works the same (I mean being blurred while custom stretched) as if I stretched the game window by hand for some unknown, non-integer size.
That looks a bit weird, applying this x2 resolution in fullscreen mode in hardware mode looks blurry, while choosing 640x480 x1 which looks completely the same in fullscreen mode, except of scaled cursor, and it works crispy. 

Anyway, Thanks for clarification, @oleg-derevenetz , very much. :)

Hi, @Districh-ru !

>Interesting, I have also 1080p display, but fheroes2 allowed me to use two 2x resolutions:

My monitor is rather old and supports quite a poor set of resolutions. I had to add manually some low widescreen resolutions, as the lowest default available 16:9 for me was  1280x768.

>As I know, this depends on resolutions supported by graphics card. Some drivers allow to set custom resolutions. After I added new resolutions fheroes2 allowed me to use them:

Worked for me as well. :)

>(There would be good in future to have non integer scales, for example set in-game resolution to 1366x768 and upscale it to fullscreen 1920x1080.)

Isn't it upscaled if we run the application fullscreen? I mean each of the available resolutions are stretching the image on running application fullscreen.
For me, this particular PR changes almost nothing, since I mostly play fullscreen so any resolution I choose shows the image upscaled to fullscreen. Even more, while playing with hardware cursor mode I enjoy crisp and clear image, no matter what resolution I've chosen. However playing with software emulated cursor mode turned on, makes image blurred. If I want to play in a window mode I simply stretch game window by hand to any size I want, which brings the same quality as software emulated cursor mode (I mean scaled and blurred image). So those fixed x2 options could be useful only for the cases where user runs application fullscreen, but without stretching on default (in the middle of the screen with black borders). But I'm not sure anyone uses such video driver setting for any application in OS.
I've expected these x2 resolutions scaling the game for an integer x2 value in window mode at least, keeping crisp image while playing in stretched window mode, but unfortunately it doesn't work for me.

More anticipated feature for me is to allow to set a custom resolutions, as I have to manually set them for my OS, before it appears in resolution selection window.
Many players (same as me) are missing widescreen resolutions, lower than default availabe (which is ~1280x768... I've noticed the majority of streamers choose this resolution, as it's the lowest available widescreen one, otherwise they set lower resolutions, but 4:3 ratio). As you can see on my screenshot I've had to set a custom 928x520 resolution within my OS to be able to play 16:9 with satisfying size of battle screen. 
I wish an option to be able to paste a custom resolution at least for software emulated cursor mode (since in hardware mode OS itself switches to the resolution we select, which couldn't be possible, if it's not in the OS list, I suppose), but I'm not sure it's possible though.

Branikolog (approved):

src/engine/screen.cpp
```diff
@@ -57,74 +57,102 @@
...
-        const std::vector<fheroes2::Size> possibleResolutions
-            = { { 640, 480 },   { 800, 600 },  { 1024, 768 },  { 1152, 864 }, { 1280, 600 }, { 1280, 720 },  { 1280, 768 }, { 1280, 960 },
-                { 1280, 1024 }, { 1360, 768 }, { 1400, 1050 }, { 1440, 900 }, { 1600, 900 }, { 1680, 1050 }, { 1920, 1080 } };
+        const std::set<fheroes2::ResolutionInfo> possibleResolutions
+            = { { 640, 480, 1 },   { 800, 600, 1 },  { 1024, 768, 1 },  { 1152, 864, 1 }, { 1280, 600, 1 }, { 1280, 720, 1 },  { 1280, 768, 1 }, { 1280, 960, 1 },
+                { 1280, 1024, 1 }, { 1360, 768, 1 }, { 1400, 1050, 1 }, { 1440, 900, 1 }, { 1600, 900, 1 }, { 1680, 1050, 1 }, { 1920, 1080, 1 } };
 
-        const fheroes2::Size lowestResolution = resolutions.back();
-        for ( const fheroes2::Size & resolution : possibleResolutions ) {
+        const fheroes2::ResolutionInfo lowestResolution = *std::min_element( resolutions.begin(), resolutions.end() );
```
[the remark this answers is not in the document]
ihhub: Yes, we should do this.

src/engine/screen.cpp
```diff
@@ -57,74 +57,102 @@
...
+        for ( const fheroes2::ResolutionInfo & resolution : possibleResolutions ) {
             if ( lowestResolution.width < resolution.width || lowestResolution.height < resolution.height || resolution == lowestResolution ) {
                 continue;
             }
```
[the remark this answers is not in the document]
ihhub: There were users who had literally one 4K resolution in the list. This hack fixes the issues for them.

src/engine/screen.cpp
```diff
@@ -57,74 +57,102 @@
...
 
-        std::sort( resolutions.begin(), resolutions.end(), SortResolutions );
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+        // Scaling is available only on SDL 2.
+        if ( resolutions.size() < 2 ) {
+            return { resolutions.begin(), resolutions.end() };
+        }
+
+        std::sort( resolutions.begin(), resolutions.end(),
+                   []( const fheroes2::ResolutionInfo & first, const fheroes2::ResolutionInfo & second ) { return first < second; } );
```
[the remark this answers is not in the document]
ihhub: Sure'!

src/fheroes2/system/settings.cpp
ihhub: This can be done as well.

src/engine/screen.cpp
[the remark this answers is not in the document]
oleg-derevenetz: Aha, that's why. OK then :)

Districh-ru:
> > (There would be good in future to have non integer scales, for example set in-game resolution to 1366x768 and upscale it to fullscreen 1920x1080.)
> 
> Isn't it upscaled if we run the application fullscreen? I mean each of the available resolutions are stretching the image on running application fullscreen. 

Not exact. Any resolution changes the Desktop resolution ans the upscale is done by monitor or graphic card driver. This causes a problem in Windows OS: Opened windows change their size to match the set resolution, changing their inner arrangement. And after exiting the game some of opened windows do not restore its original size and inner arrangement. Also some windows are force dropped to my second monitor, which is TV and is mostly turned off, so I have to get them back from the switched off TV "to the touch". :)  This is the Windows OS issue and it causes with many applications. But if fheroes2 would set the in-game resolutuion to chosen and in fullscreen do not change the Desktop resolution it would be very nice for me. :)
Setting `x2` resolution the upscale is done by fheroes2 engine (actually by SDL2) and not by monitor or graphics driver.

Branikolog:
> Not exact. Any resolution changes the Desktop resolution ans the upscale is done by monitor or graphic card driver. This causes a problem in Windows OS: Opened windows change their size to match the set resolution, changing their inner arrangement. And after exiting the game some of opened windows do not restore its original size and inner arrangement. Also some windows are force dropped to my second monitor, which is TV and is mostly turned off, so I have to get them back from the switched off TV "to the touch". :) This is the Windows OS issue and it causes with many applications. But if fheroes2 would set the in-game resolutuion to chosen and in fullscreen do not change the Desktop resolution it would be very nice for me. :)

I experience the same problems, but I prefer image quality over my OS windows arrangement. :)

Isn't this problem was fixed while running on software emulated cursor mode? (which lowers quality as well, but makes it possible to run engine on any resolution plus scaling the cursor)

> Setting `x2` resolution the upscale is done by fheroes2 engine (actually by SDL2) and not by monitor or graphics driver.

I see.

oleg-derevenetz (approved):

Districh-ru:
> I experience the same problems, but I prefer image quality over my OS windows arrangement. :)

Then in would be great if the next improvements could give the ability to set the scaler, i.e. like in Homm3HD :)
изображение

oleg-derevenetz:
@Districh-ru you are already able to choose the scaling method (among supported by SDL) via `fheroes2.cfg`.

Districh-ru:
Thanks, @oleg-derevenetz, I forgot about this. I prefer Bilinear scaling, so I did not search the сап file for it.
Setting `screen scaling type = nearest` disables blurring. @Branikolog is that what you expected from the "x2" resolutions?

Branikolog:
@Districh-ru 
I'll check it later once more to be sure... I've checked that mode yesterday, but with nearest scaling the image on x2 resolutions looked like scaled not for integer value.

Branikolog:
> Setting `screen scaling type = nearest` disables blurring. @Branikolog is that what you expected from the "x2" resolutions?

I believe, nearest scaling should be turned on as a default while selecting x2 resolutions in window mode.
Is it time to add software/hardware modes and scaling toggles in game, so player could easily adjust the image? Right now I have to edit .cfg file each time I want to switch to that x2 resolutions

PR #6520, opened 2023-01-24
Do not allow to load saves from a PoL map if PoL expansion isn't present

ihhub:
We should not allow to do that as some objects will be absent on the Adventure Map.

relates to #6047

### DIFF
--- a/src/fheroes2/game/game_io.cpp
+++ b/src/fheroes2/game/game_io.cpp
@@ -57,7 +57,7 @@ namespace
     {
         enum
         {
-            IS_PICE_OF_LOYALTY_MAP = 0x4000
+            IS_PRICE_OF_LOYALTY_MAP = 0x4000
         };
 
         HeaderSAV()
@@ -75,7 +75,7 @@ namespace
             info.localtime = static_cast<uint32_t>( rawtime );
 
             if ( fi._version == GameVersion::PRICE_OF_LOYALTY )
-                status |= IS_PICE_OF_LOYALTY_MAP;
+                status |= IS_PRICE_OF_LOYALTY_MAP;
         }
 
         uint16_t status;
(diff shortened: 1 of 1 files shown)


src/fheroes2/game/game_io.cpp
```diff
@@ -57,7 +57,7 @@ namespace
     {
         enum
         {
-            IS_LOYALTY = 0x4000
+            IS_PICE_OF_LOYALTY_MAP = 0x4000
```
oleg-derevenetz, line 60: It seems that there is a typo: `IS_PICE_OF_LOYALTY_MAP` -> `IS_PRICE_OF_LOYALTY_MAP`.

src/fheroes2/game/game_io.cpp
```diff
@@ -193,11 +193,11 @@ fheroes2::GameMode Game::Load( const std::string & filePath )
...
     }
 
-    if ( ( header.status & HeaderSAV::IS_LOYALTY ) && !conf.isPriceOfLoyaltySupported() ) {
-        fheroes2::showMessage( fheroes2::Text( _( "Warning" ), fheroes2::FontType::normalYellow() ),
-                               fheroes2::Text( _( "This file was saved by the \"The Price of Loyalty\" version of the game.\nSome items may not be available." ),
-                                               fheroes2::FontType::normalWhite() ),
-                               Dialog::OK );
+    if ( ( header.status & HeaderSAV::IS_PICE_OF_LOYALTY_MAP ) && !conf.isPriceOfLoyaltySupported() ) {
+        fheroes2::showStandardTextMessage( _( "Warning" ),
+                                           _( "This file was saved for the \"The Price of Loyalty\" map which is not supported by this version of the game." ),
```
oleg-derevenetz, line 198: It looks like fheroes2 itself doesn't support PoL :) How about this:

```
This file was saved for the "The Price of Loyalty" map, but the corresponding game assets are not available.
```

?

oleg-derevenetz (changes requested):
Hi @ihhub I just left two suggestions, what do you think?

oleg-derevenetz (approved):

Issue #6527, opened 2023-01-26
Show the enemy creature movement zone


DM-VLK:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Difficulties with quick determining the zone of movement of an enemy

### Describe the possible solution

I propose to make a viewing of the enemy’s movement in the battle, when hovering a cursor with a pressed Shift or Ctrl, as it done in Homm3 HD.

ihhub:
Hi @DM-VLK , there is an option in the battle called **"Shadow movement"** which does exactly what you need.

DM-VLK:
> there is an option in the battle called **"Shadow movement"** which does exactly what you need.

But this is for friendly creatures, not enemy ones. I meant showing the enemy's creatures turn zone like it's done in HOMM3 HD:

![5](https://user-images.githubusercontent.com/112470261/214890712-d9b21bc6-d462-44ce-a85b-f1d8452f5e8e.png)

Branikolog:
> I think we need to come up with a solution suitable for all platforms as for example Android devices do not have keyboard :)

That's a valid point!

Hi, @DM-VLK !
Are you sure, you really have difficulties with determining enemy movement zone? Unlike H3, Heroes 2 has not such a large battlefield and the majority of creatures have 3-6 movement speed, which is not so hard to determine, based on the size of the battlefield. If you remember, H1 even didn't have battle grid, as the battlefield itself was so tiny and after a few minutes of gameplay players were got used to creatures movement range.

DM-VLK:
> Are you sure, you really have difficulties with determining enemy movement zone?

I would not say so, but it would be more convenient with such an opportunity, besides, I play quite rarely. In any case, there are more important tasks, and this is just a proposition for the future.

Branikolog:
@DM-VLK 

> I would not say so, but it would be more convenient with such an opportunity, besides, I play quite rarely. In any case, there are more important tasks, and this is just a proposition for the future.

We have discussed before this and many other issues a lot with the development team and contributors. We consider fheroes2 engine as good quality product and came to conlcusion, that each feature enabled in game should be available for any player.  Fheroes2 engine, by supporting various platforms, doesnt help much regarding the solution we came. :) But we have to consider every platform, resolution and game mode while implementing something.

[...]

PR #6529, opened 2023-01-26
Rebuild the battle pathfinding cache dynamically if necessary

oleg-derevenetz:
Related to #6527

Currently battle pathfinding cache is built once per turn (well, not quite, but close enough) for the particular "current" unit (the one whose turn has come). This is inconvenient in cases where you need to quickly recalculate the cache for another arbitrary unit, so with this PR cache will be rebuilt on demand.


ihhub:

PR #6531, opened 2023-01-28
Fix scroll bar bottom position when dragging and fix high FPS rate when dragging scroll bar by mouse

### LINKED ISSUE #5840: Scroll bar shifts in the bottom edge while click
### Preliminary checks

### Platform

Windows

### Describe the bug

Scroll bar shifts a bit when we move it to a bottom edge of a short list and left click on it.

https://user-images.githubusercontent.com/55348946/190853391-833ad05c-2e76-4014-84d9-71bb27b27f66.mp4

### Save file

None.

### Additional info

_No response_

Districh-ru:
close #5840

This PR also makes the render of scrollbar only if scrollbar position was changed by mouse dragging to reduce FPS and CPU usage.

### DIFF
--- a/src/fheroes2/gui/interface_list.h
+++ b/src/fheroes2/gui/interface_list.h
@@ -388,24 +388,22 @@ namespace Interface
             }
             if ( le.MousePressLeft( _scrollbar.getArea() ) && ( _size() > maxItems ) ) {
                 const fheroes2::Point mousePosition = le.GetMouseCursor();
-                const fheroes2::Rect scrollbarArea = _scrollbar.getArea();
-                const bool noScrollNeeded = scrollbarArea.width < scrollbarArea.height
-                                                ? ( ( ( scrollbarArea.y > mousePosition.y ) || ( ( scrollbarArea.y + scrollbarArea.height ) < mousePosition.y ) )
-                                                    || ( mousePosition.y == _mousePosition.y ) )
-                                                : ( ( ( scrollbarArea.x > mousePosition.x ) || ( ( scrollbarArea.x + scrollbarArea.width ) < mousePosition.x ) )
-                                                    || ( mousePosition.x == _mousePosition.x ) );
-                if ( noScrollNeeded ) {
-                    return false;
-                }
 
-                _mousePosition = mousePosition;
-                needRedraw = true;
+                int32_t prevX = _scrollbar.x();
+                int32_t prevY = _scrollbar.y();
 
                 UpdateScrollbarRange();
 
                 _scrollbar.moveToPos( mousePosition );
+
+                // We don't need to render the scrollbar if it's position is not changed.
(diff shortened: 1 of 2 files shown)


src/fheroes2/gui/ui_scrollbar.cpp
```diff
@@ -125,7 +125,7 @@ namespace fheroes2
         if ( _isVertical() ) {
             const int32_t scrollbarImageMiddle = height() / 2;
             const int32_t minYPos = _area.y + scrollbarImageMiddle;
-            const int32_t maxYPos = _area.y + roiHeight + height() - scrollbarImageMiddle - 1;
```
ihhub, line 128: Could you please verify that all scrollbars in the game behave properly after this change?

src/fheroes2/gui/interface_list.h
```diff
@@ -468,6 +479,7 @@ namespace Interface
         int maxItems;
 
         fheroes2::Point ptRedraw;
+        fheroes2::Point _mousePosition;
```
ihhub, line 482: I suggest to name this member as **_prevMousePosition** to be even more clear about its purpose.

src/fheroes2/gui/interface_list.h
```diff
@@ -387,12 +387,23 @@ namespace Interface
                 return true;
             }
             if ( le.MousePressLeft( _scrollbar.getArea() ) && ( _size() > maxItems ) ) {
+                const fheroes2::Point mousePosition = le.GetMouseCursor();
```
ihhub, line 390: What if we remember the position of the scrollbar before and then check where it should be after. If there are no change we don't draw. What do you think?

ihhub (changes requested):
Hi @Districh-ru , I put few comments and a suggestion. Could you please take a look?

src/fheroes2/gui/ui_scrollbar.cpp
Districh-ru: Hi @ihhub, I've checked the scrollbars in Language/Video Mode selection, Load, Save, choose map, Kingdom view, Hero and Town scroll, Battle log, Town Portal dialogs. It works OK.
The problem was in rounding the result of division by 2. I've just simplified the expression: 
_area.y + roiHeight + height() - scrollbarImageMiddle - 1 = _area.y + roiHeight + height() - height() / 2 - 1 = _area.y + roiHeight + height() / 2 -1 = _area.y + roiHeight + scrollbarImageMiddle - 1.
And removed extra -1, which was added to patch this division rounding problem, which was leading to render the scrollbar one pixel down than its lowest position in some conditions.

src/fheroes2/gui/interface_list.h
Districh-ru: As we don't check the mouse position now this part of code is removed.

ihhub (approved):

Branikolog (approved):
👍

ihhub:

PR #6539, opened 2023-01-30
Android Save File Manager: implement filtering of save files by type

oleg-derevenetz:
Requested by @Branikolog in #6505

Screenshot_20230130-183300

It is possible to filter standard, campaign and multiplayer save files in various combinations as desired.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -33,7 +33,7 @@
         android:fullBackupContent="@xml/full_backup_content"
         android:hardwareAccelerated="true"
         android:icon="@mipmap/ic_launcher"
-        android:label="@string/app_name"
+        android:label="@string/app_label"
         android:supportsRtl="false">
 
         <activity
@@ -66,7 +66,7 @@
             android:label="@string/activity_toolset_label"
             android:launchMode="singleTask"
             android:taskAffinity="org.fheroes2.ToolsetTask"
-            android:theme="@style/Theme.Material3.DayNight">
+            android:theme="@style/Theme.MaterialComponents.DayNight">
 
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -80,6 +80,6 @@
             android:icon="@mipmap/ic_launcher_toolset"
             android:label="@string/activity_save_file_manager_label"
             android:taskAffinity="org.fheroes2.ToolsetTask"
-            android:theme="@style/Theme.Material3.DayNight" />
+            android:theme="@style/Theme.MaterialComponents.DayNight" />
     </application>
 </manifest>
--- a/android/app/src/main/java/org/fheroes2/SaveFileManagerActivity.java
+++ b/android/app/src/main/java/org/fheroes2/SaveFileManagerActivity.java
@@ -30,15 +30,18 @@ import android.view.View;
 import android.widget.ArrayAdapter;
 import android.widget.ImageButton;
 import android.widget.ListView;
+import android.widget.ToggleButton;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 
 public final class SaveFileManagerActivity extends Activity
 {
-    private static final String[] SAVE_FILE_EXTENSIONS = { ".sav", ".savc", ".savh" };
-
     private File saveFileDir = null;
+
+    private ToggleButton filterStandardToggleButton = null;
+    private ToggleButton filterCampaignToggleButton = null;
+    private ToggleButton filterMultiplayerToggleButton = null;
 
     private ListView saveFileListView = null;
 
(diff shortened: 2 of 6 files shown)


ihhub (approved):

Branikolog:
Hi, @oleg-derevenetz !
I've just tested new toolset feature.
While selecting a single category, it is shown in a gray colour (like it's unavailable), but two or three categories chosen are underlined with blue colour. This is quite confusing.
I hope it's not the problem of my device...

Screenshot_2023-01-31-16-49-19-17_591ba0aa0c25b75f4ad8b373a1ebed7f

Screenshot_2023-01-31-16-49-27-83_591ba0aa0c25b75f4ad8b373a1ebed7f

My multiplayer folder is empty. I think, it should be grayed in upper menu no matter what folder is selected, untill some new folders appear there. 

Screenshot_2023-01-31-16-50-08-86_591ba0aa0c25b75f4ad8b373a1ebed7f

oleg-derevenetz:
Hi @Branikolog you are doing it wrong (C). These are not folders, but filters, and you can have more than one filter active at the same time - for example, you are able to choose to show both campaign and multiplayer save files at the same time. By default all filters are turned on, so when you are pressing any of the buttons, this filter is removed and corresponding files are not shown. Colored filter means "enabled" and grayed means "disabled".

Branikolog:
@oleg-derevenetz 
Just try to select only 1 category.
It is "grayed" instead of being underlined with blue colour, like on the first screenshot I've attached above (sorry for them being so large, my phone has resolution even higher than my notebook :smile:  )
While selecting 2 or 3 filters work as intended: they underlined with blue.

oleg-derevenetz:
@Branikolog when only one category left, this button is intentionally disabled, and that's how the disabled control looks like in Android. I can give a user an ability to turn off all filters and see an empty list instead.

Branikolog:
@oleg-derevenetz 
The current behaviour is rather confusing...
On default we have 3 filter enabled (with blue lines). Then I firstly switch off multiplyer, it loses its blue line. Standard and campaign filters are left enbled with blue line. (The second screenshot I've posted above)
Then I press campaign filter button. The blue line there disappears...
So I expect the only filter applied (Standard one) to have the blue line.
But current implementation looks like on the first screenshot: there's no blue line and "Standard" filter. (it even has changed colour of text, like it's disabled) So we literally watch two same style folders (Multi and campaign) with white line, which are not selected, and one folder (standard) without any line and with grayed text, which looks, like this folder is unavailable.
There's no need to allow user to turn off the only filter, but I think it should be clear, which of them is selected (by underlining with blue line).
I also expect empty folder to have grayed text too. It could be even not selectable at all.
Anyway, everything works well, I'm only trying to suggest visually more clearer design.
The logic with filters is quite complex although. I'd personally make a simple folder selection without making it possible to show all files together (campaign+standard+multi messed up).

oleg-derevenetz:
> I also expect empty folder to have grayed text too.

There is no concept of "folders" here. And I'm strongly against graying out an "empty" filter, because it could become a "non-empty" at any time in background - for example, if new save files were copied to device over USB while toolset is running. As well as "non-empty" filter can become "empty" at any time if some save files were deleted over USB while toolset is running.

oleg-derevenetz:
> There's no need to allow user to turn off the only filter

This is usually done by disabling the corresponding control :)

Branikolog:
> > I also expect empty folder to have grayed text too.
> 
> There is no concept of "folders" here. And I'm strongly against graying out an "empty" filter, because it could become a "non-empty" at any time in background - for example, if new save files were copied to device over USB while toolset is running. As well as "non-empty" filter can become "empty" at any time if some save files were deleted over USB while toolset is running.

I'm not sure, it's a correct files interaction. I mean, adding or deleting some files on your own, while the application is running... Is it safe enough for any other programm? :)

How do you think of representing the only applied filter, like this?

1 filter

oleg-derevenetz:
> I'm not sure, it's a correct files interaction. I mean, adding or deleting some files on your own, while the application is running... Is it safe enough for any other programm?

I believe that blocking some actions based on some "snapshot" of a dataset that could be changed in background at any time is anti-pattern for any program.

oleg-derevenetz:
> How do you think of representing the only applied filter, like this?

I can try to alter the "disabled" button style to make it the same as non-disabled button, but Android styles are a bit messy and it may take some time. I propose to not disable it then for a while and allow to remove all filters.

oleg-derevenetz:
@Branikolog

> > How do you think of representing the only applied filter, like this?
> 
> I can try to alter the "disabled" button style to make it the same as non-disabled button, but Android styles are a bit messy and it may take some time. I propose to not disable it then for a while and allow to remove all filters.

Unfortunately after the research I didn't found a proper way to make a disabled `ToggleButton` look like a non-disabled (by "proper way" I mean some portable method that works with any 3rd-party styles), so I used a hack - when you press on a last remaining active filter button, it will perform a "snap back" and check itself back again. Maybe someone who is more familiar with the Android UI than me will offer the better option.

Branikolog (approved):
Thanks, @oleg-derevenetz !
Now it's much clearer, what's going on in the file manager. :)

ihhub:
@oleg-derevenetz , a bunch of thanks for this improvement!

PR #6533, opened 2023-01-29
Update German in Android-Toolset

Laserlicht:

PR #6544, opened 2023-02-01
Android Toolset: add Russian translation

oleg-derevenetz:
#6533 should be merged first.


Branikolog (approved):
Everything looks correct.
I'd personally would change "Стандарт" to "Обычная", but it's not critical.
Also, "fh2 инструментарий" is too long for icon name, but I think it's a local problem of my  device OS, which doesn't support line break for russian font.

oleg-derevenetz:
Hi @Branikolog

> Also, "fh2 инструментарий" is too long for icon name, but I think it's a local problem of my device OS, which doesn't support line break for russian font.

What would you suggest instead?

Branikolog:
@oleg-derevenetz 

> > Also, "fh2 инструментарий" is too long for icon name, but I think it's a local problem of my device OS, which doesn't support line break for russian font.
> 
> What would you suggest instead?

Actually, I don't have any good ideas regarding this. I even didn't manage to find a way to rename app icons on Android 13...
Change word order, probably? (Инструментарий fh2) 
Anyway, It's not so critical for me.

oleg-derevenetz:
> Change word order, probably? (Инструментарий fh2)

I prefer to place "fh2" first because many devices sort shortcuts alphabetically by default and if I put "Инструментарий" first, then these shortcuts will be too far apart.

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for handling this translation!

PR #6559, opened 2023-02-04
Fix invalid Kingdom income calculation

ihhub:
If one of resources has negative value per day then the original formula was just increasing it. This is how it was before the fix:
image
and this is how it should be (after the fix):
image


oleg-derevenetz (approved):
I believe that this negative income is from artifacts?

ihhub:
> I believe that this negative income is from artifacts?

Yes.

Issue #6553, opened 2023-02-04
Make pixel artist work more efficient


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Based on the latest feedback from people who want to contribute into the project's pixel art there are multiple obstacles for them to get started with work. We have to improve and simplify the onboarding stage for any potential artists.

### Describe the possible solution

This is the list of things we need to do in order to simplify artists' work
- [x]  add a separate compilation stage to compile tools such as **icn2img** and **extractor**. We don't need to cover every single OS. Windows and (optional for now) macOS are enough. Most of artists use only these OSes. Only SDL 2.
- [x] add extra bash and bat scripts to automate the work of tools or modify them. For instance, **icn2img** accepts only a single image.
- [x] add documentation as we did for developers and translators which tools to download, how to extract images and which issues to look in GitHub as a start point.
- [x] add a tool to load an image from a disk, convert it internally and save it on disk as well. This will allow artists to see how the final image would look like in the game and to verify that they use correct palette.
- [x] add a tool to generate an image of colors present in KB.PAL file. 16x16 tiles of 16 pixels each.
- [ ] add a list of color indexes used for cycling into the documentation.
- [x] add support of loading of 8-bit BMP images

oleg-derevenetz:
> add a separate compilation stage to compile tools such as icn2img and extractor. We don't need to cover every single OS. Windows and (optional for now) macOS are enough. Most of artists use only these OSes. Only SDL 2.

Tools are already being built for Windows, macOS and Linux, they are just not packed into the ZIP file.

oleg-derevenetz:
Hi @ihhub Could you please tell me, which tools are still worth to build and put into the ZIP file? As far as I can see, not all tools even have a project file for MSVC. Are there any superfluous/outdated ones among them? Maybe some need to be modernized, like `icn2img`? I am very little familiar with the artists' needs.

ihhub:
Hi @oleg-derevenetz , `extractor` and `icn2img`. The rest of tools are no in use by anyone.

An artist should download these 2 tools, ideally run a script which will unpack files and create all necessary images. We can also add extra generated files which will contain image offsets present in some ICN images.

ihhub:
[...]

PR #6564, opened 2023-02-05
icn2img tool: handle multiple ICN files

oleg-derevenetz:
Related to #6553

On Windows, only globs in the filename part of a path are supported (`dir/arch*.?cn`); globs in the intermediate directory names (`*/filename.icn`) are not supported. On *NIX there are no such limitations.


src/tools/icn2img.cpp
```diff
@@ -21,103 +21,207 @@
...
+                dstFileName += ".bmp";
             }
 
-            std::cout << "Image " << ii + 1 << " has offset of [" << static_cast<int32_t>( head.offsetX ) << ", " << static_cast<int32_t>( head.offsetY ) << "]"
-                      << std::endl;
+            static_assert( std::is_same_v<decltype( header.offsetX ), uint16_t> && std::is_same_v<decltype( header.offsetY ), uint16_t>,
+                           "Offset types have been changed, check the casts below" );
 
-            fheroes2::Save( image, dstfile, 23 );
+            std::cout << "Image " << spriteIdx + 1 << " has offset of [" << static_cast<int16_t>( header.offsetX ) << ", " << static_cast<int16_t>( header.offsetY )
```
oleg-derevenetz, line 219: These offsets are in fact signed, right?

ihhub:
Hi @oleg-derevenetz , as a separate solution I am using a batch file on Windows🙂

oleg-derevenetz:
> as a separate solution I am using a batch file on Windows🙂

This is slow, because a separate instance of `icn2img` is running for each ICN file. Also on *NIX globs will not work either, because, although on *NIX shell does all the hard work with globs, this still requires the application to be able to accept multiple arguments with file names, which `icn2img` is not able to do in its current form.

oleg-derevenetz:
@ihhub BTW is this normal that extracted PNG images are absolutely transparent? They looks fine in Explorer preview (it seems that it doesn't handle transparency very well), but completely transparent in most viewers.

ihhub:
Hi @oleg-derevenetz , performance is not an issue here as this tool is not a part of the main project. It is good to have but it is optional.

Some PNG files are totally transparent if they are 1x1 pixel size but 99% of them have some visible information.

ihhub:
@oleg-derevenetz , I use **XnView** and even normal Window Viewer works fine to me.

src/tools/icn2img.cpp
ihhub: Actually, yes. We have to fix it.

oleg-derevenetz:
@ihhub

> I use XnView and even normal Window Viewer works fine to me.

I have a weird issue with PNG images (BMP images are fine). PNG files are of the right size, but entirely transparent. What if you open PNG file in Paint for example? I tried the files from `archer.icn`.

ihhub:
Hi @oleg-derevenetz , I compiled the latest version of these tools from master branch for SDL 2 and everything works. I can see PNG file via XnView. This is a batch file I used to extract all images (yeah, it's an absolute mess but it works):
```bat
@echo off

for /R %%f in (*.AGG) do (
 mkdir "%%~nf"
 extractor.exe "%%~nf.AGG" "%%~nf"
 
 for /R %%i in (%%~nf\*.icn) do (
  mkdir "%%~nf\%%~ni"
  icn2img.exe "%%~nf\%%~ni.icn" . "%%~nf\KB.PAL"
 )
)
```

ihhub:
Paint shows proper images as well:
image

oleg-derevenetz:
@ihhub

>  I compiled the latest version of these tools from master branch for SDL 2 and everything works.

Well, in my case extracted PNG files look strange. In Windows Explorer:

explorer

But for example in MS Paint they are completely transparent :) I attached the archive with PNG images. Are they OK on your system?

[archer.zip](https://github.com/ihhub/fheroes2/files/10611417/archer.zip)

ihhub:
Your images are very weird:
image

I also noticed that your images are bigger. Do you use SDL 2 tools? SDL 1 doesn't support PNG and these images by size look like Bitmap.

ihhub:
Also please try a full rebuild of the tool. Just in case...

oleg-derevenetz:
@ihhub

> Do you use SDL 2 tools?

Yes, and these files are really PNG - at least they have proper PNG header.

oleg-derevenetz:
And here are BMP files (just replaced ".png" with ".bmp" in `icn2img.cpp`):

[archer.zip](https://github.com/ihhub/fheroes2/files/10611480/archer.zip)

It seems that they have no issues.

oleg-derevenetz:
@ihhub Huh, I just removed the `libpng16.dll` from build dir and now these PNGs are fine. It seems that SDL tries to use `libpng` if present (it seems that otherwise it uses some internal PNG library implementation) and this somehow messes things up.

ihhub:
Wow! This is really bad. That is why I created an issue to remove dependencies from SDL about loading and saving images but this is another story. We can not to include unused libraries into the final package for tools.

ihhub:
Let's remove this library to avoid messing things.

oleg-derevenetz:
> Let's remove this library to avoid messing things.

I patched `vcpkg` port to force the `sdl2-image` to use the built-in PNG library and removed the `libpng` completely. I tested `icn2img` and can confirm that issue is gone.

ihhub (approved):

PR #6502, opened 2023-01-18
Rework the way how object ICN types are used

ihhub:
- remove **mp2riddle_t** structure and add appropriate comments into `MapSphinx::LoadFromMP2()` method
- rename **_objectType** member into **_objectIcnType** and make it as `MP2::ObjectIcnType` to avoid confusion from `MapObjectType` enumeration
- move multiple static methods from `Tiles` class into anonymous namespace to reduce the complexity of the class
- rename **mp2_object** member of `Tiles` class into **_mainObjectType** to make it clear what it is
- rename all **tileset** usages into **objectIcnType**
- remove all usages of bitwise shift operations on objectIcnTypes (except where we read data from an MP2 file). This was an ultimate goal of the changes as the code was a complete mess and not sustainable to be supported
- fix multiple compilation warnings
~~- add **_hasObjectAnimation** and **_isMarkedAsRoad** members into `TilesAddon` and `Tiles` classes. For now we don't save them as we do not use them anywhere (at least for now)~~

This pull request changes the very core of Adventure Map rendering and passability logic so please test these changes on old saves and new maps very carefully. Pay attention to passability in builds from the master branch and in this branch, as well as Adventure Map object rendering.

relates to #6415

PR #6567, opened 2023-02-05
Fix possibly corrupted save format

ihhub:
This is a possible for some corrupted save files introduced by #6502. We need to explicitly cast enumeration entries to prevent possible corruption.

This is a save file which might be created because of latest changes:
[Artifact_Isle_gazebo.zip](https://github.com/ihhub/fheroes2/files/10610242/Artifact_Isle_gazebo.zip)


ihhub:
Hi @Branikolog , could you please try to replicate the scenario leading to this corrupted save file? I don't think it is possible to fix it without a lot of effort but it is important to know that the problem is fixed for the future save files.

ihhub:
As discussed with @Branikolog in Discord this is due to a fact that this save file was created during testing #6502 but before [this commit](https://github.com/ihhub/fheroes2/pull/6502/commits/92f97c8f3d3fd3ee7603213ca75eebd89bf7d0e8). However, adding extra checks would not harm anything.

oleg-derevenetz (approved):

PR #6577, opened 2023-02-05
Update team members in credits page

ihhub:
To appreciate @Districh-ru work.


oleg-derevenetz (approved):
AFAIK in HoMM2 it's not the case, but in HoMM3 Minotaurs have minimum morale +1.

Districh-ru (approved):
Many thanks for the appreciation!

PR #6585, opened 2023-02-06
Fix saving of Maps::Tiles and Maps::TilesAddon enums

oleg-derevenetz:
Follow-up to #6567 which fixed not all cases.

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <map>
 #include <set>
+#include <type_traits>
 
 #include "agg_image.h"
 #include "army.h"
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz , I added static cast for a reason that in the future we are going to extend enumerations, at least for main object type.

oleg-derevenetz:
Hi @ihhub

> I added static cast for a reason that in the future we are going to extend enumerations, at least for main object type.

All static casts are still in place, but the new ones were added for `TilesAddon` to help choosing a proper `operator<<()` overload.

ihhub:
Oops, I meant static assert :)

oleg-derevenetz:
> Oops, I meant static assert :)

I left `static_assert()` on reading, because that's where we should add compatibility shims if underlying type of `enum` will be changed. What would be the purpose of `static_assert()` on writing?

ihhub:
You're right, no purpose.

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for this hot fix!

PR #6590, opened 2023-02-07
feature: Add lighthouse information popup in kingdom overview

### LINKED ISSUE #6558: Add a popup dialog when right-clicking on Lighthouses icon and its number in Kingdom Overview dialog
### Preliminary checks

There is no way to know for new players what this tiny icon stands for:
image

We need to add a popup window showing the number of captured Lighthouses and a short description about this.

### Additional info

_No response_

Osmodium:
resolves #6558

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -18,7 +18,6 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-#include "image.h"
 
 #include <algorithm>
 #include <cassert>
@@ -26,6 +25,7 @@
 #include <cstring>
 #include <type_traits>
 
+#include "image.h"
 #include "image_palette.h"
 
 namespace
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -18,7 +18,6 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
-
 #include <algorithm>
 #include <cassert>
 #include <cstdlib>
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -18,13 +18,14 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "image.h"
+
 #include <algorithm>
 #include <cassert>
 #include <cstdlib>
 #include <cstring>
 #include <type_traits>
 
-#include "image.h"
 #include "image_palette.h"
 
 namespace
@@ -1116,6 +1117,11 @@ namespace fheroes2
     {
         out.resize( in.width(), in.height() );
         Copy( in, 0, 0, out, 0, 0, in.width(), in.height() );
+    }
+
+    void Copy( const Image & in, Image & out, int32_t outX, int32_t outY )
+    {
+        Copy( in, 0, 0, out, outX, outY, in.width(), in.height() );
     }
 
     void Copy( const Image & in, int32_t inX, int32_t inY, Image & out, int32_t outX, int32_t outY, int32_t width, int32_t height )
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1119,11 +1119,6 @@ namespace fheroes2
         Copy( in, 0, 0, out, 0, 0, in.width(), in.height() );
     }
 
-    void Copy( const Image & in, Image & out, int32_t outX, int32_t outY )
-    {
-        Copy( in, 0, 0, out, outX, outY, in.width(), in.height() );
-    }
-
     void Copy( const Image & in, int32_t inX, int32_t inY, Image & out, int32_t outX, int32_t outY, int32_t width, int32_t height )
     {
         if ( !Verify( in, inX, inY, out, outX, outY, width, height ) ) {
--- a/src/fheroes2/gui/ui_kingdom.cpp
+++ b/src/fheroes2/gui/ui_kingdom.cpp
@@ -19,6 +19,12 @@
  ***************************************************************************/
 
 #include "ui_kingdom.h"
+
+#include <algorithm>
+#include <cstdint>
+#include <memory>
+#include <string>
+
 #include "agg_image.h"
 #include "icn.h"
 #include "image.h"
@@ -28,10 +34,6 @@
 #include "ui_dialog.h"
 #include "ui_text.h"
 #include "world.h"
-#include <algorithm>
-#include <cstdint>
-#include <memory>
-#include <string>
 
 namespace fheroes2
 {
--- a/src/fheroes2/gui/ui_kingdom.cpp
+++ b/src/fheroes2/gui/ui_kingdom.cpp
@@ -30,6 +30,7 @@
 #include "image.h"
 #include "kingdom.h"
 #include "mp2.h"
+#include "tools.h"
 #include "translations.h"
 #include "ui_dialog.h"
 #include "ui_text.h"
@@ -52,8 +53,12 @@ namespace fheroes2
         const std::string body( _( "For every lighthouse controlled, your heroes can move further on water." ) );
 
         const CustomImageDialogElement lighthouseImageElement( AGG::GetICN( ICN::OBJNMUL2, 61 ) );
-        const Text lighthouseControlledText( _( "Controlled: " ) + std::to_string( lighthouseCount ), FontType::smallWhite() );
+
+        std::string lighthouseControlledString = _( "Controlled: %{count}" );
+        StringReplace( lighthouseControlledString, "%{count}", std::to_string( lighthouseCount ) );
+        const Text lighthouseControlledText( lighthouseControlledString, FontType::smallWhite() );
         const TextDialogElement lighthouseControlledElement( std::make_shared<Text>( lighthouseControlledText ) );
+
         showMessage( Text( header, FontType::normalYellow() ), Text( body, FontType::normalWhite() ), buttons,
                      { &lighthouseImageElement, &lighthouseControlledElement } );
     }
--- a/src/fheroes2/gui/ui_kingdom.cpp
+++ b/src/fheroes2/gui/ui_kingdom.cpp
@@ -50,13 +50,32 @@ namespace fheroes2
     {
         const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
         const std::string header( _( "Lighthouses" ) );
-        const std::string body( _( "For every lighthouse controlled, your heroes can move further on water." ) );
+        const std::string body( _( "For every lighthouse controlled, your ships will move further each day." ) );
 
-        const CustomImageDialogElement lighthouseImageElement( AGG::GetICN( ICN::OBJNMUL2, 61 ) );
+        Image combined( 128, 69 );
+        combined.reset();
 
-        std::string lighthouseControlledString = _( "Controlled: %{count}" );
+        // shadow top
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 60 ), combined, 2, 33 );
+        // shadow middle
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 71 ), combined, 3, 37 );
+        // shadow bottom
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 72 ), combined, 17, 39 );
+
+        // lighthouse top
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 59 ), combined, 63, 0 );
+        // lighthouse middle
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 61 ), combined, 57, 5 );
+        // lighthouse bottom
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 73 ), combined, 49, 37 );
+        // lighthouse light
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 62 ), combined, 61, 10 );
+
+        const CustomImageDialogElement lighthouseImageElement( combined );
+
+        std::string lighthouseControlledString = _( "%{count}" );
         StringReplace( lighthouseControlledString, "%{count}", std::to_string( lighthouseCount ) );
-        const Text lighthouseControlledText( lighthouseControlledString, FontType::smallWhite() );
+        const Text lighthouseControlledText( lighthouseControlledString, FontType::normalWhite() );
         const TextDialogElement lighthouseControlledElement( std::make_shared<Text>( lighthouseControlledText ) );
 
         showMessage( Text( header, FontType::normalYellow() ), Text( body, FontType::normalWhite() ), buttons,
--- a/src/fheroes2/gui/ui_kingdom.cpp
+++ b/src/fheroes2/gui/ui_kingdom.cpp
@@ -21,7 +21,6 @@
 #include "ui_kingdom.h"
 #include "agg_image.h"
 #include "icn.h"
-#include "kingdom.h"
 #include "translations.h"
 #include "ui_dialog.h"
 #include "ui_text.h"
@@ -41,7 +40,7 @@ namespace fheroes2
     {
         const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
         const std::string header( _( "Lighthouses" ) );
-        const std::string body( _( "For every lighthouse controlled, your heroes moves further on water." ) );
+        const std::string body( _( "For every lighthouse controlled, your heroes can move further on water." ) );
 
         const CustomImageDialogElement lighthouseImageElement( AGG::GetICN( ICN::OBJNMUL2, 61 ) );
         const Text lighthouseControlledText( _( "Controlled: " + std::to_string( lighthouseCount ) ), FontType::smallWhite() );
--- a/src/fheroes2/gui/ui_kingdom.cpp
+++ b/src/fheroes2/gui/ui_kingdom.cpp
@@ -49,7 +49,6 @@ namespace fheroes2
     void showLighthouseInfo( const Kingdom & kingdom, const int buttons )
     {
         const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
-        const std::string header( StringObject( MP2::OBJ_LIGHTHOUSE, 2 ) );
         const std::string body( _( "For every lighthouse controlled, your ships will move further each day." ) );
 
         Image combined( 128, 69 );
@@ -76,7 +75,7 @@ namespace fheroes2
         const Text lighthouseControlledText( std::to_string( lighthouseCount ), FontType::normalWhite() );
         const TextDialogElement lighthouseControlledElement( std::make_shared<Text>( lighthouseControlledText ) );
 
-        showMessage( Text( header, FontType::normalYellow() ), Text( body, FontType::normalWhite() ), buttons,
+        showMessage( Text( StringObject( MP2::OBJ_LIGHTHOUSE, 2 ), FontType::normalYellow() ), Text( body, FontType::normalWhite() ), buttons,
                      { &lighthouseImageElement, &lighthouseControlledElement } );
     }
 }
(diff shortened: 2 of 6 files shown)


oleg-derevenetz:
Hi @Osmodium you can use special keywords in the PR description to automatically link issue(s) to it as described here:

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

oleg-derevenetz:
Also if you are using MSVC for development, there is a handy plugin that will properly format the code literally in two clicks:

https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.ClangFormat

I use it myself on Windows. If you use Linux or macOS for development, I recommend to install and use the `clang-format` tool.

Osmodium:
@oleg-derevenetz Thank you, I wasn't familiar with these, but I will use these going forward 👍

Osmodium:
I don't know why it fails in files I haven't touched 🤔

oleg-derevenetz:
> I don't know why it fails in files I haven't touched 🤔

You touched 3 files: `ui_kingdom.cpp`, `ui_kingdom.h` and `kingdom_overview.cpp`, and in all of them copyright date remains ending in 2022. Since you touched them, you should as well update the copyright date to end in 2033, as suggested by code style check.

Osmodium:
Ahh I see. I was cheated by the fact that it referenced a `.tmp` file 😅

oleg-derevenetz:
@Osmodium According to IWYU diagnostics:

https://github.com/ihhub/fheroes2/pull/6590/commits/e4368c9b450e51c4f3a17b10503c0ceb1db83374

you should have added a few headers to `ui_kingdom.cpp` (please note that you should use `cstdint` instead of `stdint.h`). Why did you removed so many headers from `kingdom_overview.cpp`?

oleg-derevenetz:
> I got the extension you suggested, but that does nothing to add or cleanup headers in the file, so I manually cleaned them up to the bare minimum, or that was the idea.

Just let the IWYU manage this and follow its suggestions. "Bare minimum" is objectionable because it depends on how headers include each other internally, and it's dangerous to count on it.

Osmodium:
> > I got the extension you suggested, but that does nothing to add or cleanup headers in the file, so I manually cleaned them up to the bare minimum, or that was the idea.
> 
> Just let the IWYU manage this and follow its suggestions. "Bare minimum" is objectionable because it depends on how headers include each other internally, and it's dangerous to count on it.

Is there a way to do this locally so I don't have to rely on the report in github?

oleg-derevenetz:
@Osmodium regarding style: normally we using the following inclusion style in some hypothetical `file.cpp`:

```cpp
#include "file.h" // this module's header

#include <some_std_header>
#include <another_std_header>

#include "some_another_local_module_header"
#include "another_one_header_of_another_local_module"
```

Headers in each group are sorted alphabetically (`clang-format` checks this).

Osmodium:
> @Osmodium regarding style: usually we use the following inclusion style in some hypothetical `file.cpp`:
> 
> ```c++
> #include "file.h" // this module's header
> 
> #include <some_std_header>
> #include <another_std_header>
> 
> #include "some_another_local_module_header.h"
> #include "another_one_header_of_another_local_module.h"
> ```
> 
> Headers in each group are sorted alphabetically (`clang-format` checks this).

Okay thanks :).
It seems like it this time wanted to move the standard libraries includes down under the local includes?
https://github.com/ihhub/fheroes2/actions/runs/4115736768/jobs/7104876677

oleg-derevenetz:
> It seems like it this time wanted to move the standard libraries includes down under the local includes?

Each group of headers (this module's header, library headers, other module's headers) should be separated by an empty line.

Osmodium:
> > It seems like it this time wanted to move the standard libraries includes down under the local includes?
> 
> Each group of headers (this module's header, library headers, other module's headers) should be separated by an empty line.

Yeah, I just couldn't see that from the report.
Btw. thank you for being patient with me :)

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +44,17 @@ namespace fheroes2
...
     }
+
+    void showLighthouseInfo( const Kingdom & kingdom, const int buttons )
+    {
+        const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
+        const std::string header( _( "Lighthouses" ) );
+        const std::string body( _( "For every lighthouse controlled, your heroes can move further on water." ) );
+
+        const CustomImageDialogElement lighthouseImageElement( AGG::GetICN( ICN::OBJNMUL2, 61 ) );
+        const Text lighthouseControlledText( _( "Controlled: " ) + std::to_string( lighthouseCount ), FontType::smallWhite() );
```
oleg-derevenetz, line 55: Plain string concatenations are in general not very good for localizations - for example, it is assumed here that localized version of this string will always be left-to-right and the number of controlled lighthouses will always be at the end of the string :) I'd rather prefer something like this instead:

```cpp
std::string lighthouseControlledString = _( "Controlled: %{count}" );
StringReplace( lighthouseControlledString, "%{count}", std::to_string( lighthouseCount ) );

const Text lighthouseControlledText( lighthouseControlledString, ... );
[...]
```

I believe that this version offers a better control to the translator. What do you think?

Osmodium: I thought about how to best go about that and yes I agree that your approach is more flexible and thus gives more control to the translator. I will look at implementing it like that, thanks! :)

Osmodium: Added in [b048aac](https://github.com/ihhub/fheroes2/pull/6590/commits/b048aac8de36a5ac58fd47f23080b9ee15acd359)

ihhub:
Hi @Osmodium , can you please attach a screenshot for the dialog to see how it looks like?

Osmodium:
> Hi @Osmodium , can you please attach a screenshot for the dialog to see how it looks like?

Hi @ihhub.
I have attached 2 screenshots in the issue here: https://github.com/ihhub/fheroes2/issues/6558#issuecomment-1420704617

ihhub:
Hi @Osmodium , based on your screenshots the projector in the lighthouse is missing as well as the lower part of it.

Osmodium:
> Hi @Osmodium , based on your screenshots the projector in the lighthouse is missing as well as the lower part of it.
Is it possible to achieve with the `CustomImageDialogElement` in the dialog? Also is there a way to browse icons?

Osmodium:
> Hi @Osmodium , please use multiple images from ICN:OBJNMUL2, ranging from 59 to 73. I believe you need to create a combined image using 59, 60, 61, 62, 71, 72, 73 indexes with proper shifts. Then you create a separate image which will be a combination of all these sprites and feed it into `CustomImageDialogElement`. Moreover, to be really cool we can use `DynamicImageDialogElement` class to provide the animation of Lighthouse.

HI @ihhub, okay I was wondering if this was what you meant :) I will have a look at how to do this.
Do you have any example or know where this is already done in the game so I have something to go by regarding constructing the lighthouse?

ihhub:
Hi @Osmodium , it is done as a part of Adventure Map rendering inside interface_area.cpp but there is no code specific to Lighthouse at all.

Osmodium:
@ihhub Cool! I'll have a look at it, thanks for the pointer :)

Osmodium:
Hi @ihhub 
I have looked at the ICN's of the lighthouse and can see that the base of the lighthouse includes a part of the shadow.
I'm not sure it would make sense to include the shadow in the dialog and if not that it might look wierd with the cutoff on the base.

image

ihhub:
Hi @Osmodium , I suggest to include shadow as well. We have the similar case when we display Mage Guild building as well. Also please don't forget to include the inner mirror image🙂

Now it looks much better!

ihhub:
@Osmodium , I also think that we should just display a number over "Controlled: N" as the description above states it. Please use normal font for it.

Osmodium:
Hi @ihhub, Ah I wasn't aware that it was used for the mage guild, will look at that for a bit of inspiration.
Yeah the image was just to show a bit of the progress along with the cut-off shadow 😄. I think, however, that in order to make it animate a constructed image, I'd have to extend the `DynamicImageDialogElement` to support this. Or do you have another idea?

ihhub:
Hi @Osmodium , we can do this in 2 stages (and 2 pull requests):
- complete a full static image of Lighthouse with a shadow
- add animation later (you can create a new issue for later work)

In this case you won't be overwhelmed with work and it's easier to review small chunks of code.

Osmodium:
Hi @ihhub!
I think to split it into 2 different PR's sounds like a good idea.
So, the current code has the dialog looking like this:
image
I changed the text to resemble the dialog that pops up when visiting a lighthouse:
image
Hope this is closer to what you had in mind :)

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +45,40 @@ namespace fheroes2
...
+    {
+        const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
+        const std::string header( _( "Lighthouses" ) );
+        const std::string body( _( "For every lighthouse controlled, your ships will move further each day." ) );
+
+        Image combined( 128, 69 );
+        combined.reset();
+
+        // shadow top
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 60 ), combined, 2, 33 );
```
ihhub, line 59: Please use **Copy** over **Blit** as `combined` image is fully empty.

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +45,40 @@ namespace fheroes2
 
         showResourceMessage( header, body, buttons, kingdom.GetIncome( Kingdom::INCOME_ALL ) );
     }
+
+    void showLighthouseInfo( const Kingdom & kingdom, const int buttons )
+    {
+        const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
+        const std::string header( _( "Lighthouses" ) );
```
ihhub, line 52: Please modify `MP2::StringObject()` function to handle multiple Lighthouses like it is done for few other objects and call it within **showMessage()** function call.

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +45,40 @@ namespace fheroes2
...
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 61 ), combined, 57, 5 );
+        // lighthouse bottom
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 73 ), combined, 49, 37 );
+        // lighthouse light
+        Blit( AGG::GetICN( ICN::OBJNMUL2, 62 ), combined, 61, 10 );
+
+        const CustomImageDialogElement lighthouseImageElement( combined );
+
+        std::string lighthouseControlledString = _( "%{count}" );
+        StringReplace( lighthouseControlledString, "%{count}", std::to_string( lighthouseCount ) );
```
ihhub, line 77: There is no need to translate a number. **std::to_string()** would be sufficient.

src/fheroes2/gui/ui_kingdom.cpp
ihhub, line 59: To avoid hardcoded values please consider 2 factors:
- a size of tile on Adventure Map is always fixed and equal to **TILEWIDTH**
- each fheroes2::Sprite image returned from **AGG::GetICN()** call has an offset

A combination of these 2 ideas will lead to much smaller usage of hardcoded values.

src/fheroes2/kingdom/kingdom_overview.cpp
```diff
@@ -785,6 +788,9 @@ void Kingdom::openOverviewDialog()
         else if ( le.MousePressRight( rectIncome ) ) {
             fheroes2::showKingdomIncome( *this, 0 );
         }
+        else if ( le.MousePressRight( rectLighthouse ) ) {
+            fheroes2::showLighthouseInfo( *this, 0 );
```
ihhub, line 792: Please use **Dialog::ZERO** instead of **0**.

ihhub (changes requested):
Hi @Osmodium , I put several comments in this pull request. Could you please take a look at them?

src/fheroes2/gui/ui_kingdom.cpp
Osmodium: Fixed in [c11025b](https://github.com/ihhub/fheroes2/pull/6590/commits/c11025b2d4d988d812516cc8a780ede1c9ff8d90)

src/fheroes2/kingdom/kingdom_overview.cpp
Osmodium: Changed in [a55eba7](https://github.com/ihhub/fheroes2/pull/6590/commits/a55eba7b7c09c1a7516f460729a72c92041b2c8d)

Osmodium: Fixed in [a55eba7](https://github.com/ihhub/fheroes2/pull/6590/commits/a55eba7b7c09c1a7516f460729a72c92041b2c8d)

src/fheroes2/gui/ui_kingdom.cpp
Osmodium: [7d2d744](https://github.com/ihhub/fheroes2/pull/6590/commits/7d2d7444487252b25b84da2c8d220bd912ae682d) & [6a38e74](https://github.com/ihhub/fheroes2/pull/6590/commits/6a38e7404ddf4a93d9a16ade3822763c0ac12161)

src/fheroes2/gui/ui_kingdom.cpp
Osmodium: Use **TILEWIDTH** and sprite offsets to place parts in [d93891f](https://github.com/ihhub/fheroes2/pull/6590/commits/d93891f863bdd28936d7b4aa10a8e21abb85485c)

Osmodium: Use **Copy** over **Blit** [eb7ef61](https://github.com/ihhub/fheroes2/pull/6590/commits/eb7ef612763380db144741b5146d0b873c901b26)

src/engine/image.cpp
```diff
@@ -1118,6 +1119,11 @@ namespace fheroes2
         Copy( in, 0, 0, out, 0, 0, in.width(), in.height() );
     }
 
+    void Copy( const Image & in, Image & out, int32_t outX, int32_t outY )
+    {
+        Copy( in, 0, 0, out, outX, outY, in.width(), in.height() );
+    }
+
```
ihhub, line 1126: Please remove this new function. I understand the intention of adding it but we should avoid it because the signature of the function is not the same as others. It contains only 1 input image with no coordinates and an output image with coordinates. We should explicitly state either for both or for nobody.

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +45,39 @@ namespace fheroes2
...
+        const uint32_t lighthouseCount = world.CountCapturedObject( MP2::OBJ_LIGHTHOUSE, kingdom.GetColor() );
+        const std::string body( _( "For every lighthouse controlled, your ships will move further each day." ) );
+
+        const Sprite & shadow_top = AGG::GetICN( ICN::OBJNMUL2, 60 );
+        const Sprite & shadow_middle = AGG::GetICN( ICN::OBJNMUL2, 71 );
+        const Sprite & shadow_bottom = AGG::GetICN( ICN::OBJNMUL2, 72 );
+        const Sprite & lighthouse_top = AGG::GetICN( ICN::OBJNMUL2, 59 );
+        const Sprite & lighthouse_middle = AGG::GetICN( ICN::OBJNMUL2, 61 );
+        const Sprite & lighthouse_bottom = AGG::GetICN( ICN::OBJNMUL2, 73 );
+        const Sprite & lighthouse_light = AGG::GetICN( ICN::OBJNMUL2, 62 );
```
ihhub, line 60: Please name the variable with **camelCase**. The current naming is not suitable.

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +45,39 @@ namespace fheroes2
...
+        Copy( lighthouse_middle, combined, TILEWIDTH * 2 + lighthouse_middle.x(), TILEWIDTH + lighthouse_middle.y() - top_offset );
+        Copy( lighthouse_bottom, combined, TILEWIDTH * 2 + lighthouse_bottom.x(), TILEWIDTH * 2 + lighthouse_bottom.y() - top_offset );
+        Copy( lighthouse_light, combined, TILEWIDTH * 2 + lighthouse_light.x(), TILEWIDTH + lighthouse_light.y() - top_offset );
+
+        const CustomImageDialogElement lighthouseImageElement( combined );
+
+        const Text lighthouseControlledText( std::to_string( lighthouseCount ), FontType::normalWhite() );
+        const TextDialogElement lighthouseControlledElement( std::make_shared<Text>( lighthouseControlledText ) );
+
+        showMessage( Text( StringObject( MP2::OBJ_LIGHTHOUSE, 2 ), FontType::normalYellow() ), Text( body, FontType::normalWhite() ), buttons,
```
ihhub, line 80: I suggest to add a comment above explaining why we set number 2 here.

ihhub (changes requested):
Hi @Osmodium , I added just few more comments. Could you please take a look at them?

src/fheroes2/gui/ui_kingdom.cpp
Osmodium: Added in [1ba0809](https://github.com/ihhub/fheroes2/pull/6590/commits/1ba080900cb0ee1a213c60a241441ead5aa365ac)

src/fheroes2/gui/ui_kingdom.cpp
Osmodium: Renamed in [1ba0809](https://github.com/ihhub/fheroes2/pull/6590/commits/1ba080900cb0ee1a213c60a241441ead5aa365ac)

src/engine/image.cpp
Osmodium: Removed in [7da9544](https://github.com/ihhub/fheroes2/pull/6590/commits/7da9544856332647b6ff254f0fb468dcb9fed2ec)

src/fheroes2/gui/ui_kingdom.cpp
```diff
@@ -33,4 +45,40 @@ namespace fheroes2
...
+
+        const Sprite & shadowTop = AGG::GetICN( ICN::OBJNMUL2, 60 );
+        const Sprite & shadowMiddle = AGG::GetICN( ICN::OBJNMUL2, 71 );
+        const Sprite & shadowBottom = AGG::GetICN( ICN::OBJNMUL2, 72 );
+        const Sprite & lighthouseTop = AGG::GetICN( ICN::OBJNMUL2, 59 );
+        const Sprite & lighthouseMiddle = AGG::GetICN( ICN::OBJNMUL2, 61 );
+        const Sprite & lighthouseBottom = AGG::GetICN( ICN::OBJNMUL2, 73 );
+        const Sprite & lighthouseLight = AGG::GetICN( ICN::OBJNMUL2, 62 );
+
+        const int32_t top_offset = TILEWIDTH - lighthouseTop.height();
```
ihhub, line 62: Please rename this variable too :)

Osmodium: Haha omg 😂

Osmodium: @ihhub there we go :) [26bee86](https://github.com/ihhub/fheroes2/pull/6590/commits/26bee86aeab5b929a52921f6199695e6f530c0d6)

ihhub (approved):
@Osmodium , well done!

oleg-derevenetz (approved):

ihhub:
@Osmodium , thank you so much for your contribution to the project!

Osmodium:
> @Osmodium , thank you so much for your contribution to the project!

@ihhub , thank you for being patient with me :) It's a great project for one of the games I played alot, I hope to be able to contribute even more :)

PR #6596, opened 2023-02-08
Fix the "unused parameter" warnings in release builds, remove unused variable

oleg-derevenetz:


ihhub:
@oleg-derevenetz , thanks a lot for cleaning up the code!

PR #6601, opened 2023-02-09
Improve & modernize AGG extraction tools

oleg-derevenetz:
Related to #6553

Also revived & modernized all original tools. I have only some doubts about the correctness of the images that `til2img` extracts.


.github/workflows/make.yml
```diff
@@ -11,75 +11,204 @@ jobs:
...
+            fheroes2
+            ./docs/README.txt
+            files/data/*.h2d
+            files/lang/*.mo
+            ./script/agg/extract_agg.sh
+            ./script/demo/download_demo_version.sh
+            ./script/homm2/extract_homm2_resources.sh
+            script/linux/install_sdl_1.sh
+            ./src/tools/extractor
+            ./src/tools/icn2img
```
ihhub, line 31: I recommend not to mix the main application with these tools. Knowing the reaction of users they will start asking questions and run these for no reason.

VisualStudio/tools/bin2txt/sources.props
```diff
@@ -1,6 +1,57 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
+    <ClCompile Include="..\engine\agg_file.cpp" />
```
ihhub, line 4: I think this tool doesn't have so many dependencies on files. At least it shouldn't.

src/tools/bin2txt.cpp
```diff
@@ -1,6 +1,9 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2022                                                    *
+ *   Copyright (C) 2022 - 2023                                             *
+ *                                                                         *
+ *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
+ *   Copyright (C) 2009 by Andrey Afletdinov <fheroes2@gmail.com>          *
```
ihhub, line 6: Actually I wrote this tool completely few years ago (yeah, it's ugly as the purpose was to study bin files) so this copyright header is not needed :)

ihhub (changes requested):
Hi @oleg-derevenetz , I left few comments here. My biggest concern here is that we must separate tools for artists and put them into a separate package. My point is that if we ship them together with the application it might rise even more questions like why and what these tool do, are they needed to run the game, why they don't do anything and etc. What do you think?

src/tools/bin2txt.cpp
oleg-derevenetz: Yeah, that's what happening when using one specific file as a template :)

oleg-derevenetz:
Hi @ihhub

> My biggest concern here is that we must separate tools for artists and put them into a separate package.

I addressed your concerns in the latest commits. Also could you please take a look at images generated by the `til2img` tool? Are they fine?

ihhub:
Hi @oleg-derevenetz , I think we need to move TIL image reading to a separate place to be used by the engine and the tool. I can do this tomorrow.

ihhub:
Hi @oleg-derevenetz , the only question besides til2img tool I have why we removed the scripts? I thought they were very useful :(

oleg-derevenetz:
Hi @ihhub

> Hi @oleg-derevenetz , the only question besides til2img tool I have why we removed the scripts? I thought they were very useful :(

I can bring them back, but all they did was run `extractor` and then `icn2img` with certain parameters :) There is no complex logic or directory traversal. Why they are useful? :)

ihhub:
> Hi @ihhub
> 
> > Hi @oleg-derevenetz , the only question besides til2img tool I have why we removed the scripts? I thought they were very useful :(
> 
> I can bring them back, but all they did was run `extractor` and then `icn2img` with certain parameters :) There is no complex logic or directory traversal. Why they are useful? :)

They are very useful. I faced many people in my life who can't work in command line but they are experts in their field. So having a one-click solution is much better for them.

oleg-derevenetz:
Hi @ihhub

> They are very useful. I faced many people in my life who can't work in command line but they are experts in their field. So having a one-click solution is much better for them.

OK, I'll bring them back.

Regarding `til2img`, I see no difference with the new API :) Hope these images are fine. Here's what they look like (`ground32.til`):

screen1
screen2

ihhub:
Hi @oleg-derevenetz , I think I know why: please load KB.PAL palette within the application.

oleg-derevenetz:
@ihhub

> I think I know why: please load KB.PAL palette within the application.

Yeah, now they look much better :) Original `til2img` implementation didn't use a palette for some reason.

ihhub:
> @ihhub
> 
> > I think I know why: please load KB.PAL palette within the application.
> 
> Yeah, now they look much better :) Original `til2img` implementation didn't use a palette for some reason.

Hi @oleg-derevenetz , the original code had hardcoded palette from the game. We removed it and only few colors are present used in a message about missing resources.

ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for your tremendous work!

PR #6391, opened 2022-12-18
Add missing generated small buttons and adjust some others

zenseii:
Add UNIFORM exit button for market place, add okay button for necromancy dialog at end of battle. Fix size of many buttons to fit smallest possible words for French.

PR #6605, opened 2023-02-11
Rewrite standard window generation code

### LINKED ISSUE #6547: Correct standard window generation code
### Preliminary checks

### Platform

Windows

### Describe the bug

Standard window generation code has multiple defects and we need to fix them:
- vertical bars at the left and right sides of the window are too long
- the inner part of the window has repetitive pattern which is easily visible by eyes
- [optimization] for some windows we don't need to generated the middle part of the window and as of now we waste recourses by doing such
image

### Save file

No need.

### Additional info

_No response_

Districh-ru:
fix #6547
relates to #6391

Made rendering of a Standard window using game resources and connecting them by using simple transition pattern: 
pattern

Also added ability not to render the window background if only the borders are needed (in example for Castle, Battle dialogs).

castle dialog

<details><summary>Standard Window for Evil interface with and without background</summary>

evil with background

evil without background

</details>

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -18,13 +18,14 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "image.h"
+
 #include <algorithm>
 #include <cassert>
 #include <cstdlib>
 #include <cstring>
 #include <type_traits>
 
-#include "image.h"
 #include "image_palette.h"
 
 namespace
@@ -2285,7 +2286,8 @@ namespace fheroes2
             const int32_t halfWidth = width / 2;
 
             for ( int32_t x = 0; x < halfWidth; ++x ) {
-                const int32_t stepY = 1 << ( ( halfWidth - x ) / 2 + 1 ); //static_cast<int32_t>( std::pow( 2, ( halfWidth - x ) / 2 + 1 ) );
+                // The step is 2 to the power, which decreases by 1 every second line and is 1 in the center.
+                const int32_t stepY = 1 << ( ( halfWidth - x ) / 2 + 1 );
                 const int32_t offsetY = stepY / 2 * ( ( x + halfWidth ) % 2 );
 
                 for ( int32_t y = 0; y < height; ++y ) {
@@ -2334,7 +2336,8 @@ namespace fheroes2
                 transformIn += widthIn;
             }
             for ( int32_t y = 0; y < halfHeight; ++y ) {
-                const int32_t stepX = 1 << ( ( halfHeight - y ) / 2 + 1 ); // static_cast<int32_t>( std::pow( 2, ( halfHeight - y ) / 2 + 1 ) );
+                // The step is 2 to the power, which decreases by 1 every second line and is 1 in the center.
+                const int32_t stepX = 1 << ( ( halfHeight - y ) / 2 + 1 );
                 const int32_t offsetX = stepX / 2 * ( ( y + halfHeight ) % 2 );
 
                 for ( int32_t x = 0; x < width; ++x ) {
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1447,6 +1447,211 @@ namespace fheroes2
         DrawLine( image, { roi.x, roi.y + roi.height - 1 }, { roi.x + roi.width, roi.y + roi.height - 1 }, value, roi );
     }
 
+    void DitheringTransition( const Image & in, int32_t inX, int32_t inY, Image & out, int32_t outX, int32_t outY, int32_t width, int32_t height, const bool isVertical,
+                              const bool isReverse )
+    {
+        if ( !Verify( in, inX, inY, out, outX, outY, width, height ) ) {
+            return;
+        }
+
+        const int32_t widthIn = in.width();
+        const int32_t offsetIn = inY * widthIn + inX;
+        const uint8_t * imageIn = in.image() + offsetIn;
+        const uint8_t * transformIn = in.transform() + offsetIn;
+
+        const int32_t widthOut = out.width();
+        const int32_t offsetOut = outY * widthOut + outX;
+        uint8_t * imageOut = out.image() + offsetOut;
+        uint8_t * transformOut = out.transform() + offsetOut;
+
+        const bool isInNonSinglelayerToOutSinglelayer = !in.singleLayer() && out.singleLayer();
+
+        if ( isVertical ) {
+            // We also go in a loop from the right part of the image to its center.
+            const uint8_t * imageInRightPoint = imageIn + width - 1;
+            const uint8_t * transformInRightPoint = transformIn + width - 1;
+
+            uint8_t * imageOutRightPoint = imageOut + width - 1;
+            uint8_t * transformOutRightPoint = transformOut + width - 1;
+
+            const int32_t halfWidth = width / 2;
+
+            // We make a symmetric transition and if width is odd we shift one line right.
+            if ( ( width % 2 ) == 1 ) {
+                if ( isReverse ) {
+                    --imageOutRightPoint;
+                    --transformOutRightPoint;
+                    --imageInRightPoint;
+                    --transformInRightPoint;
+                }
+                else {
+                    ++imageOut;
+                    ++transformOut;
+                    ++imageIn;
+                    ++transformIn;
+                }
+            }
+
+            for ( int32_t x = 0; x < halfWidth; ++x ) {
+                // The step is 2 to the power, which decreases by 1 every second line and is 1 in the center.
+                // We limit the step power to 30 to not overflow the 32 bit 'stepY'.
+                const int32_t stepPower = std::min( 30, ( halfWidth - x ) / 2 + 1 );
+                const int32_t stepY = 1 << stepPower;
+
+                // The point position in dithered pattern.
+                const int32_t patternPoint = stepY / 2 * ( ( x + halfWidth ) % 2 );
+
+                for ( int32_t y = 0; y < height; ++y ) {
+                    const int32_t offsetOutX = y * widthOut;
+                    const int32_t offsetInX = y * widthIn;
+                    const int32_t offsetY = y % stepY;
+
+                    if ( isReverse == ( patternPoint != offsetY ) ) {
+                        if ( isInNonSinglelayerToOutSinglelayer && ( *( transformIn + offsetInX ) == 1 ) ) {
+                            // Skip pixel.
+                            continue;
+                        }
+
+                        // First part of transition: we copy single pixels.
+                        *( imageOut + offsetOutX ) = *( imageIn + offsetInX );
+
+                        if ( !out.singleLayer() ) {
+                            if ( in.singleLayer() ) {
+                                // Set the copied pixel visible.
+                                *( transformOut + offsetOutX ) = 0;
+                            }
+                            else {
+                                *( transformOut + offsetOutX ) = *( transformIn + offsetInX );
+                            }
+                        }
+                    }
+                    else {
+                        if ( isInNonSinglelayerToOutSinglelayer && ( *( transformInRightPoint + offsetInX ) == 1 ) ) {
+                            // Skip pixel.
+                            continue;
+                        }
+
+                        // Second part of transition: we copy image excluding single pixels.
+                        *( imageOutRightPoint + offsetOutX ) = *( imageInRightPoint + offsetInX );
+
+                        if ( !out.singleLayer() ) {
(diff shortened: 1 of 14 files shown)


src/engine/image.cpp
```diff
@@ -2250,6 +2251,179 @@ namespace fheroes2
...
+            const uint8_t * transformIn2 = transformIn + width - 1;
+
+            uint8_t * imageOut2 = imageOut + width - 1;
+            uint8_t * transformOut2 = transformOut + width - 1;
+
+            const int32_t halfWidth = width / 2;
+
+            for ( int32_t x = 0; x < halfWidth; ++x ) {
+                // The step is 2 to the power, which decreases by 1 every second line and is 1 in the center.
+                const int32_t stepY = 1 << ( ( halfWidth - x ) / 2 + 1 );
```
oleg-derevenetz, line 2290: What if `halfWidth` is more than 31?

src/engine/image.cpp
```diff
@@ -2250,6 +2251,179 @@ namespace fheroes2
...
+                const int32_t offsetY = stepY / 2 * ( ( x + halfWidth ) % 2 );
+
+                for ( int32_t y = 0; y < height; ++y ) {
+                    const int32_t offsetOutX = y * widthOut;
+                    const int32_t offsetInX = y * widthIn;
+
+                    if ( isReverse ? ( ( y % stepY ) != offsetY ) : ( ( y % stepY ) == offsetY ) ) {
```
ihhub, line 2297: Could you please create a separate variable for **( y % stepY )** expression as it is difficult to read this condition?

Also we can optimize the condition:
```cpp
if ( isReverse == ( offset != offsetY ) )
```

src/engine/image.cpp
```diff
@@ -2250,6 +2251,179 @@ namespace fheroes2
...
+                        if ( !in.singleLayer() && out.singleLayer() && ( *( transformIn + offsetInX ) == 1 ) ) {
```
ihhub, line 2298: We can cache **!in.singleLayer() && out.singleLayer()** expression outside the loops.

src/engine/image.cpp
```diff
@@ -2250,6 +2251,179 @@ namespace fheroes2
...
+                ++transformOut;
+                ++imageIn;
+                ++transformIn;
+            }
+
+            const uint8_t * imageIn2 = imageIn + width - 1;
+            const uint8_t * transformIn2 = transformIn + width - 1;
+
+            uint8_t * imageOut2 = imageOut + width - 1;
+            uint8_t * transformOut2 = transformOut + width - 1;
```
ihhub, line 2284: We should avoid cryptic variable naming involving numbers. It is not easy to understand what number 2 stands for.

src/fheroes2/gui/ui_window.cpp
```diff
@@ -19,42 +19,284 @@
...
     }
 
     void StandardWindow::render()
     {
-        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ( Settings::Get().isEvilInterfaceEnabled() ? ICN::SURDRBKE : ICN::SURDRBKG ), 0 );
-        const fheroes2::Image renderedImage
-            = fheroes2::Stretch( sprite, borderSize, 0, sprite.width() - borderSize, sprite.height() - borderSize, _windowArea.width, _windowArea.height );
-        fheroes2::Blit( renderedImage, _output, _windowArea.x, _windowArea.y );
+        // Notice: ICN::SURDRBKE and ICN::SURDRBKG has 16 (equals to BORDERWIDTH) pixels shadow from the left and the bottom sides.
+        const fheroes2::Sprite & horizontalSprite = fheroes2::AGG::GetICN( ( Settings::Get().isEvilInterfaceEnabled() ? ICN::SURDRBKE : ICN::SURDRBKG ), 0 );
```
ihhub, line 69: Please cache **Settings::Get().isEvilInterfaceEnabled()** value in a separate variable.

src/fheroes2/gui/ui_window.cpp
```diff
@@ -19,42 +19,284 @@
...
+                    fheroes2::Copy( _output, _windowArea.x + fromOffsetX, backgroundOffsetY, _output, _windowArea.x + toOffsetX, backgroundOffsetY,
+                                    std::min( backgroundSpriteWidth, _windowArea.width - cornerSize - toOffsetX ), backgroundCopyHeight );
+                    toOffsetX += stepX;
+                }
+            }
+
+            // If we need more copies to fill background vertically we make a transition and copy existing image in full background width.
+            if ( backgroundVerticalCopies > 0 ) {
+                int32_t toOffsetY = cornerSize + backgroundSpriteHeight;
+                fheroes2::DitheringTransition( _output, backgroundOffsetX, backgroundOffsetY, _output, backgroundOffsetX, _windowArea.y + toOffsetY - transitionSize,
```
ihhub, line 147: I can see that this is the only place (within this if-body where we use output image as input. Is it possible not to use it? If yes then we can simplify logic in **fheroes2::DitheringTransition()** in relation to **singleLayer()** calls.

src/fheroes2/gui/ui_window.cpp
ihhub, line 69: This class is in **fheroes2** so all calls within this method should be without **fheroes2**. This will improve readability of the code.

ihhub (changes requested):
Hi @Districh-ru , I left several comments in this pull request. Would you mind to take a look at them in your spare time?

src/engine/image.cpp
Districh-ru: Hi @oleg-derevenetz, you are right it should not be more than 30.
I have not thought about so big transition length...
I've added a limit (`min(30,...)`).
But in practice so big transition length woun't be used. :)

oleg-derevenetz: > But in practice so big transition length woun't be used. :)

Well, anyway it's better be safe than sorry :) There is a (wrong) opinion exists that shifting the variable to the left by the number of bits which is more than the variable size in bits will left the variable with zero value, although in fact it is an undefined behavior - not only the variable may in theory have a literally ANY value after this, but also the optimizer counts on the fact that such a shift will never ever happen.

src/fheroes2/gui/ui_window.cpp
Districh-ru: Hi @ihhub, here we use output as input because previously we rendered there copies of background image horizontally with transitions and now we use this duplicated image to make its copies vertically. I use `_output` to avoid making a temporary image.
All cases of `singleLayer()` are present in `DitheringTransition()` for possible future use in other places.
(We could also add there a scale factor for pattern: currently it changes step every 2 line, but it could bу not 2 ... but let's leave this to another PR when this feature will be needed.)

src/engine/image.cpp
Districh-ru: I'll try to avoid them in future. But sometimes it's hard to come up with names for variables. ;)

src/fheroes2/gui/ui_window.cpp
Districh-ru: Sure!

src/engine/image.cpp
Districh-ru: > Well, anyway it's better be safe than sorry :)

Absolutely agree!

src/engine/image.cpp
```diff
@@ -1446,6 +1447,211 @@ namespace fheroes2
...
+                    if ( isReverse == ( patternPoint != offsetY ) ) {
+                        if ( isInNonSinglelayerToOutSinglelayer && ( *( transformIn + offsetInX ) == 1 ) ) {
+                            // Skip pixel.
+                            continue;
+                        }
+
+                        // First part of transition: we copy single pixels.
+                        *( imageOut + offsetOutX ) = *( imageIn + offsetInX );
+
+                        if ( !out.singleLayer() ) {
```
ihhub, line 1518: Let's please simplify reading of this function as it is hard to read so many incapsulated blocks:
```cpp
if ( out.singleLayer() ) {
    continue;
}
... // the rest of the code.
```

The same for lines 1537, 1609 and 1630.

src/engine/image.cpp
```diff
@@ -1446,6 +1447,211 @@ namespace fheroes2
...
+                            if ( in.singleLayer() ) {
+                                // Set the copied pixel visible.
+                                *( transformOut + offsetOutX ) = 0;
+                            }
+                            else {
+                                *( transformOut + offsetOutX ) = *( transformIn + offsetInX );
+                            }
```
ihhub, line 1525: If the input image is a single-layer image would be more efficient to just memset before the loops and later in the code only use `if ( !in.singleLayer() ) {` check?

Moreover, combining this comment and the comment above we can have:
```cpp
memset(...);
const bool copyTransform = !in.singleLayer() && !out.singleLayer();
...
// inside the loop
if ( copyTransform ) {
    *( transformOut + offsetOutX ) = *( transformIn + offsetInX );
}
```
What do you think?

src/fheroes2/gui/ui_window.cpp
```diff
@@ -19,47 +19,285 @@
...
+        Blit( verticalSprite, 0, verticalSpriteHeight - borderEdgeOffset, _output, _windowArea.x, bottomCornerOffsetY - extraCornerSize, cornerSize, extraCornerSize );
+
+        Blit( verticalSprite, verticalSpriteWidth - borderEdgeOffset, bottomCornerSpriteOffsetY, _output, rightCornerOffsetX - extraCornerSize, bottomCornerOffsetY,
+              extraCornerSize, cornerSize );
+        Blit( verticalSprite, rightCornerSpriteOffsetX, verticalSpriteHeight - borderEdgeOffset, _output, rightCornerOffsetX, bottomCornerOffsetY - extraCornerSize,
+              cornerSize, extraCornerSize );
+
+        // Render the background.
+        if ( _hasBackground ) {
+            const Sprite & backgroundSprite = AGG::GetICN( ( isEvilInterface ? ICN::STONEBAK_EVIL : ICN::STONEBAK ), 0 );
```
ihhub, line 111: If it is not difficult can we please move this if-body into a separate private method to simplify reading of the code?

ihhub (changes requested):
Hi @Districh-ru , I put several suggestions here. Would you mind to take a look at them in your free time?

src/engine/image.h
```diff
@@ -241,6 +241,10 @@ namespace fheroes2
 
     void DrawRect( Image & image, const Rect & roi, uint8_t value );
 
+    // Make a transition to "in" image from left to right or vertically - from top to bottom. The direction of transition can be reversed.
```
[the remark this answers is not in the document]
Districh-ru: Hi @ihhub, I'll add a link to https://en.wikipedia.org/wiki/Dither, but I didn't find this algorithm, I looked at dithered gradient images and decided to make our function as it is. :)

src/engine/image.cpp
Districh-ru: We copy image and transform layer only for some particular pixels and if we set the whole transform layer to 0 before the loops we will also set the pixels, that are not rewritten in the loop and can have transform layer other than 0.
So I think it is not suitable here. :(

src/fheroes2/gui/ui_window.cpp
Districh-ru: I've moved not the whole if-body but only the render of background sprite. I left here the transitions from borders to background as to move them to a new method we have to transfer many variables or compute them once again in this method.

src/engine/image.cpp
ihhub: Ah, you're right. This is sad.

ihhub:
I included @zenseii into the list of reviewers as his button-related work depends on this pull request.

ihhub (approved):
Hi @Districh-ru , I approve this pull request but I will test it later for other window sizes to see how the algorithm behaves.

Districh-ru:
> Look at the shadow of the dialog overlaying the ocean between the town portal dialog and the file options dialog: image
> 
> In one it looks dithered in the other it is completely uniform.

Hi, @zenseii.
I'll try to fix this tomorrow: the original windows use the shadow from ICN, while the Standard Windows code makes shadow with `ApplyTransform` function. We can try to use shadow from ICN for this window also.

ihhub:
Hi @zenseii and @Districh-ru , if the original windows do not have proper shadows with effects let's fix them separately. This pull request is big enough 🙂

Districh-ru:
Hi @ihhub, @zenseii!
I found a simple solution to fix the window shadow and couldn't resist implementing it:
изображение

ihhub:
@Districh-ru , thank you so much for this fix! I tested it on my side and the solution looks great! If we find something buggy we can fix it later.

PR #6608, opened 2023-02-12
Move TIL image decoding to a common place

ihhub:
as discussed in #6601

### DIFF
--- a/src/engine/image_tool.cpp
+++ b/src/engine/image_tool.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2020 - 2022                                             *
+ *   Copyright (C) 2020 - 2023                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
@@ -235,7 +235,7 @@ namespace fheroes2
         return true;
     }
 
-    Sprite decodeICNSprite( const uint8_t * data, uint32_t sizeData, const int32_t width, const int32_t height, const int16_t offsetX, const int16_t offsetY )
+    Sprite decodeICNSprite( const uint8_t * data, const uint32_t sizeData, const int32_t width, const int32_t height, const int16_t offsetX, const int16_t offsetY )
     {
         Sprite sprite( width, height, offsetX, offsetY );
         sprite.reset();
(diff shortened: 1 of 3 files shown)


src/engine/image_tool.cpp
```diff
@@ -325,6 +325,24 @@ namespace fheroes2
...
     }
 
+    void decodeTILImages( const uint8_t * data, const size_t imageCount, const int32_t width, const int32_t height, std::vector<Image> & output )
+    {
+        assert( data != nullptr && imageCount > 0 && width > 0 && height > 0 );
+
+        output.clear();
+        output.resize( imageCount );
+
+        const size_t imageSize = static_cast<size_t>( width * height );
```
oleg-derevenetz, line 335: We can avoid this warning this way:

```suggestion
        const size_t imageSize = static_cast<size_t>( width ) * height;
```

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3421,24 +3421,15 @@ namespace fheroes2
...
 
-                const uint32_t count = buffer.getLE16();
-                const uint32_t width = buffer.getLE16();
-                const uint32_t height = buffer.getLE16();
-                const uint32_t size = width * height;
-                if ( headerSize + count * size != data.size() ) {
+                const int32_t count = buffer.getLE16();
+                const int32_t width = buffer.getLE16();
+                const int32_t height = buffer.getLE16();
+                if ( count < 1 || width < 1 || height < 0 || ( headerSize + static_cast<size_t>( count * width * height ) ) != data.size() ) {
```
oleg-derevenetz, line 3427: Why `width < 1` but `height < 0`? Are there valid images with `width == 1` and `height == 0`? Also `getLE16()` returns `uint16_t`, so it is impossible for `uint16_t` to become negative on widening conversion to `int32_t`.

Also I propose to replace `static_cast<size_t>( count * width * height )` by `static_cast<size_t>( count ) * width * height` here as well.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3447,7 +3438,7 @@ namespace fheroes2
                     const bool horizontalFlip = ( shapeId & 2 ) != 0;
                     const bool verticalFlip = ( shapeId & 1 ) != 0;
 
-                    for ( uint32_t i = 0; i < count; ++i ) {
+                    for ( int32_t i = 0; i < count; ++i ) {
```
oleg-derevenetz, line 3441: Since `i` is used just as an index of `vector` here, I propose to make it `size_t`.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few proposals and questions, could you please take a look when you have time?

src/fheroes2/agg/agg_image.cpp
ihhub: But count is **int32_t**.

ihhub: Actually, I will reduce it to normal uint16_t as we read it from the stream.

oleg-derevenetz: > But count is int32_t.

Ah yes, it is signed now.

oleg-derevenetz (approved):
LGTM. As soon as IWYU become happy, let's try to use it in the new `til2img` tool.

PR #6621, opened 2023-02-13
Add a tool to generate an image from a palette file

ihhub:
relates to #6553


src/tools/pal2img.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include "image.h"
+#include "image_palette.h"
+#include "image_tool.h"
+#include "serialize.h"
+#include "system.h"
+
+int main( int argc, char ** argv )
+{
+    if ( argc != 3 ) {
```
oleg-derevenetz, line 35: Since this tool did not accept variable number of arguments, I replaced `<` by `!=` here.

src/tools/pal2img.cpp
```diff
@@ -0,0 +1,76 @@
...
+        std::string baseName = System::GetBasename( argv[0] );
+
+        std::cerr << baseName << " generates an image with colors based on a provided palette file." << std::endl
```
oleg-derevenetz, line 38: I replaced imperative mood by the indicative one to match other tools.

src/tools/pal2img.cpp
```diff
@@ -0,0 +1,76 @@
...
+                  << "Syntax: " << baseName << " palette_file.pal output.bmp" << std::endl;
```
oleg-derevenetz, line 39: This tool may (and will, if built with `FHEROES2_WITH_IMAGE` and I pass the `output.png` as the second parameter) also generate PNG files. You can add this to the description, for example:

```suggestion
                  << "Syntax: " << baseName << " palette_file.pal output.[bmp|png]" << std::endl;
```

or you can not add it - it's up to you.

oleg-derevenetz (approved):
LGTM, just one optional suggestion regarding support of PNG file generation.

PR #6626, opened 2023-02-13
extractor tool: extract each AGG file to a separate subdir

oleg-derevenetz:
Related to #6553

It [turns out](https://github.com/ihhub/fheroes2/issues/6553#issuecomment-1428380378) that `HEROES2.AGG` and `HEROES2X.AGG` may contain items with the same names, thus, extracting both of these files to the same directory is undesirable.


ihhub (approved):

ihhub:

PR #6627, opened 2023-02-13
Make ICNHeader::offsetX and ICNHeader::offsetY signed

### LINKED ISSUE #5857: tools icn2img CreateDirectoryA
### Preliminary checks

### Platform

Windows

### Describe the bug

**system.cpp:**

#if defined( __MINGW32__ ) || defined( _MSC_VER )
// clang-format off
// shellapi.h must be included after windows.h
#include <windows.h>
#include <shellapi.h>
#include <direct.h>   // added
// clang-format on
#else
#include <dirent.h>
#endif

int System::MakeDirectory( const std::string & path )
{
#if defined( __WIN32__ ) && defined( _MSC_VER )
    //If the function succeeds, the return value is nonzero.
    //bool ret = CreateDirectoryA( path.c_str(), nullptr );
    //GetLastError();
    // _mkdir zero if dir created (opposite CreateDirA)
    return _mkdir( path.c_str());

i changed it to _mkdir() on my install (add #include <direct.h> ), because CreateDirectoryA() returns zero if fail!!!
----------------------------------------------------------------------
**icn2img.cpp**
    if ( System::MakeDirectory( prefix ) != 0) {
        std::cout << "Error mkdir: " << prefix << std::endl;
        std::cerr << "Error: " << strerror( errno ) << std::endl;
        std::cerr << "(hint: u can only create one directory under windows!)\n";
        return EXIT_FAILURE;
    }

### Save file

not avail

### Additional info

_No response_

oleg-derevenetz:
close #5857

### DIFF
--- a/src/engine/agg_file.h
+++ b/src/engine/agg_file.h
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2020 - 2022                                             *
+ *   Copyright (C) 2020 - 2023                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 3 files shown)


ihhub (approved):

ihhub:

PR #6629, opened 2023-02-13
add a frame for a hero portrait on the quickinfo dialog

### LINKED ISSUE #6599: Add missing frame over small hero portrait in hero info popup dialog
### Preliminary checks

If you right click on a castle you can see that the castle icon has a frame around it:
image

But the hero icon does not have it:
image

which is odd.

Add a similar frame for the small hero icon.

### Additional info

_No response_

undef21:
fix #6599

### DIFF
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -796,12 +796,13 @@ void Dialog::QuickInfo( const HeroBase & hero, const fheroes2::Point & position 
     dst_pt.y = cur_rt.y + 2;
     text.draw( dst_pt.x, dst_pt.y, display );
 
+    const fheroes2::Sprite & frame = fheroes2::AGG::GetICN( conf.isEvilInterfaceEnabled() ? ICN::LOCATORE : ICN::LOCATORS, 22 );
+
     // mini port heroes
     const fheroes2::Sprite & port = isActiveHero ? activeHero->GetPortrait( PORT_SMALL ) : activeCaptain->GetPortrait( PORT_SMALL );
     if ( !port.empty() ) {
-        const fheroes2::Sprite & frame = fheroes2::AGG::GetICN( conf.isEvilInterfaceEnabled() ? ICN::LOCATORE : ICN::LOCATORS, 22 );
-        fheroes2::Blit( frame, display, cur_rt.x + ( cur_rt.width - frame.width() ) / 2, cur_rt.y + 13 );
-        fheroes2::Blit( port, display, cur_rt.x + ( cur_rt.width - port.width() ) / 2, cur_rt.y + 17 );
+        fheroes2::Blit( frame, display, cur_rt.x + ( cur_rt.width - frame.width() ) / 2, cur_rt.y + 12 );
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @undef21 , I think we need to move up the frame as right now it covers a part of text for some translations:
image

Let's hear @Branikolog  and @LeHerosInconnu about this.

undef21:
Hello, @ihhub. There are too many info in this dialog. We can't move the hero name string upper and army sprites lower . We could move the frame up to 1px, move texts down, reduce vertical intervals between strings. The tallest sprites of monsters anyway will a little overlap last string.

Branikolog:
Hi, @ihhub and @undef21 
Actually, we have a few pixels above hero name and few below creature counters. 
In the OG name is placed higher, than in fheroes2.
image

Together with shifting the frame a bit to upper side and moving primary skills down I believe we can +/- fit all elements into this window. The space between skills could also be reduced (1px), but I'm not sure, whether this would look well.
And yes, the frame itself could be placed 1-2 px higher.

undef21:
Hello, @Branikolog. If we shift the hero name and monster counters, they could overlap decorative elements of the dialog frame.

Branikolog:
@undef21 
Frame 2

Here I moved hero frame 1 px higher.
Creature counters 1 px lower. Creatures - 2 pxs lower. And reduced space between primary skills for 1px between each line. 
More or less everything fits. We can also play with creatures position, since Titans and Dragons are always overlap something (in other places). For me it looks acceptable...

undef21:
new2

@Branikolog, fixed.

LeHerosInconnu:
Hello everyone,

> Hi @undef21 , I think we need to move up the frame as right now it covers a part of text for some translations:
image
> 
> Let's hear @Branikolog and @LeHerosInconnu about this.

This seems okay, although too tight for my taste, in the last proposal.
I think it's time to get serious about this window, "nibbling" pixels is only good for a short time. :)

Related discussions: https://github.com/ihhub/fheroes2/discussions/3330, https://github.com/ihhub/fheroes2/discussions/4687.

Branikolog:
Hi, @LeHerosInconnu 

> although too tight for my taste, in the last proposal.

Agreed, but we have no choice currently. And it's definitely better, than for overlapping and no frame cases.

Branikolog (approved):

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -796,12 +796,13 @@ void Dialog::QuickInfo( const HeroBase & hero, const fheroes2::Point & position
     dst_pt.y = cur_rt.y + 2;
     text.draw( dst_pt.x, dst_pt.y, display );
 
+    const fheroes2::Sprite & frame = fheroes2::AGG::GetICN( conf.isEvilInterfaceEnabled() ? ICN::LOCATORE : ICN::LOCATORS, 22 );
```
ihhub, line 799: I suggest to name this variable **heroPortraitFrame** to avoid misleading name which might be interpreted as dialog frame.

ihhub (commented):
Hi @undef21 , I put one suggestion here. Could you please check it?

ihhub (approved):

ihhub:
@undef21 , thank you very much for the improvement!

PR #6636, opened 2023-02-14
Add an ability to load 8-bit images

ihhub:
relates to #6553

### DIFF
--- a/src/engine/image_tool.cpp
+++ b/src/engine/image_tool.cpp
@@ -195,6 +195,7 @@ namespace fheroes2
                 for ( ; inX != inXEnd; ++inX, ++outX, ++transformX ) {
                     assert( *inX < palette->ncolors );
                     const SDL_Color * color = palette->colors + *inX;
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
                     if ( color->a < 255 ) {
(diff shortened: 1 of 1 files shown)


src/engine/image_tool.cpp
```diff
@@ -165,12 +165,62 @@ namespace fheroes2
...
+                    assert( *inX < palette->ncolors );
+                    const SDL_Color * color = palette->colors + *inX;
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+                    if ( color->a < 255 ) {
+                        if ( color->a == 0 ) {
+                            *transformX = 1;
+                        }
+                        else if ( color->r == 0 && color->g == 0 && color->b == 0 ) {
+                            *transformX = 2;
+                        }
```
oleg-derevenetz, line 205: Is it fine that `*outX` remains uninitialized in both cases above?

oleg-derevenetz (commented):
Hi @ihhub I just left one question, could you please take a look?

src/engine/image_tool.cpp
ihhub: It is fixed. It wasn't strictly necessary but it's better to prevent any issues.

oleg-derevenetz (approved):

PR #6640, opened 2023-02-14
Memorize zoom level of the View World screen

### LINKED ISSUE #6619: View World screen, zoom level of the View World screen is not memorized
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the zoom level of the View World screen is not memorized unlike in the original game.

In original game:

https://user-images.githubusercontent.com/43583994/218465811-9786658d-c718-4aeb-bc77-52d8d5e16c9a.mp4

In fheroes2:

https://user-images.githubusercontent.com/43583994/218465897-797af0e5-c3e4-458a-8f69-0052cff95b7c.mp4

### Save file

Not needed.

### Additional info

_No response_

Districh-ru:
close #6619

### DIFF
--- a/src/fheroes2/kingdom/view_world.h
+++ b/src/fheroes2/kingdom/view_world.h
@@ -25,6 +25,7 @@
 #include <cstdint>
 
 #include "math_base.h"
+#include "settings.h"
 
 namespace Interface
 {
@@ -47,14 +48,6 @@ enum class ViewWorldMode : int32_t
 class ViewWorld
 {
 public:
-    enum ZoomLevel : int32_t
-    {
-        ZoomLevel0 = 0,
-        ZoomLevel1 = 1,
-        ZoomLevel2 = 2,
-        ZoomLevel3 = 3, // Max zoom, but should only exists for debug builds
-    };
-
     static void ViewWorldWindow( const int32_t color, const ViewWorldMode mode, Interface::Basic & interface );
 
     struct ZoomROIs
@@ -68,7 +61,7 @@ public:
         const fheroes2::Rect & GetROIinPixels() const;
         fheroes2::Rect GetROIinTiles() const;
 
-        ZoomLevel _zoomLevel;
+        ZoomLevel _zoomLevel{ ZoomLevel::ZoomLevel1 };
         fheroes2::Point _center;
         std::array<fheroes2::Rect, 4> _roiForZoomLevels;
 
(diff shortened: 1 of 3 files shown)


Districh-ru:
Hello @LeHerosInconnu!
Could you please check this PR in your spare time.

src/fheroes2/kingdom/view_world.h
```diff
@@ -68,7 +68,7 @@ class ViewWorld
         const fheroes2::Rect & GetROIinPixels() const;
         fheroes2::Rect GetROIinTiles() const;
 
-        ZoomLevel _zoomLevel{ ZoomLevel2 };
```
ihhub, line 71: I suggest to keep the previous code to avoid having an uninitialized member.

src/fheroes2/system/settings.h
```diff
@@ -451,6 +461,7 @@ class Settings
 
     int game_type;
     int preferably_count_players;
+    int32_t _viewWorldZoomLevel{ 2 };
```
ihhub, line 464: It is good to avoid using hardcoded values. To minimize file dependencies I suggest to create a separate header file which will contain only `ViewWorldZoomLevel` enumeration (by the way it can be uint8_t as it doesn't have too many entries. Use this enumeration for `ViewWorld` and `Settings` classes. What do you think?

ihhub (changes requested):
Hi @Districh-ru , I put few comments here. Would you mind to check them?

src/fheroes2/system/settings.h
Districh-ru: Hi @ihhub!
In `settings.h` we already have enumerations for `AdventureMapScrollSpeed` and `MusicSource`. What if we put `enum class ViewWorldMode : uint8_t` there, as `zoomLevel` becomes a part of global game configuration?

ihhub: I think it will be better than just hardcoded values.

src/fheroes2/kingdom/view_world.h
Districh-ru: I'll also change it to ZoomLevel1, as it is the default value in original game.

ihhub (approved):

Branikolog (approved):

ihhub:
@Districh-ru , many thanks for adding this improvement!

LeHerosInconnu:
Hello @Districh-ru,

> Hello @LeHerosInconnu! Could you please check this PR in your spare time.

I did some tests, it seems to work correctly.

PR #6654, opened 2023-02-15
Fix Mini-map reset of enemy object colors after enemy vanquish

Districh-ru:
Fix an issue, found by **Dmytr01** (on Discord):
The color of enemy objects (mines) was not reset on Mini-map after enemy vanquish.

This archive contains two saves I used to check the fix (capture an enemy castle, defeat an enemy hero): [saves.zip](https://github.com/ihhub/fheroes2/files/10746348/saves.zip)

### DIFF
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -188,6 +188,11 @@ namespace
 
         hero.FadeOut();
         hero.SetFreeman( reason );
+
+        // If the enemy is vanquished we fully redraw the radar map image as there might be color reset of enemy's objects.
+        if ( hero.GetKingdom().isLoss() ) {
+            Interface::Basic::Get().SetRedraw( Interface::REDRAW_RADAR );
+        }
     }
 
     void RecruitMonsterFromTile( Heroes & hero, Maps::Tiles & tile, const std::string & msg, const Troop & troop, bool remove )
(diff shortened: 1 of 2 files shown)


oleg-derevenetz (commented):
What bothers me somewhat about this solution is that we are probably mixing logic and visualization too much here. Yes, I know that the old code likes to sin with this, but it's better to separate these two things somehow in a new code if possible.

Districh-ru:
> What bothers me somewhat about this solution is that we are probably mixing logic and visualization too much here. Yes, I know that the old code likes to sin with this, but it's better to separate these two things somehow in a new code if possible.

Hi, @oleg-derevenetz. Tomorrow I'll think where to move this code ... maybe to `heroes_action.cpp`.
I put it here as this is the place where loss conditions are checked and we don't need to do any more checks, just call the `Redraw` function.

oleg-derevenetz:
@Districh-ru Thinking logically, the enemy kingdom could be vanquished on the human player's turn only after some hero's action, right? So probably we could move this logic somewhere inside that file. Code in `kingdom.cpp` is more "UI-agnostic" because it could be executed for AI-controlled kingdoms, etc.

ihhub (approved):

Districh-ru:
> @Districh-ru Thinking logically, the enemy kingdom could be vanquished on the human player's turn only after some hero's action, right? So probably we could move this logic somewhere inside that file. Code in `kingdom.cpp` is more "UI-agnostic" because it could be executed for AI-controlled kingdoms, etc.

@oleg-derevenetz, you are right. From the first sight Ш did not want to complicate logic in `heroes_action.cpp` and check if the enemy is vanquished. But in fact it is not so complicated.
Thanks for pointing the right way!
I've made the full radar update for the cases of castle capture and hero lose in battle. If the enemy is vanquished after 7 days without a castle the map already fully redraws at the start of day.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:

PR #6661, opened 2023-02-16
Fix a few code smells, remove unused code

oleg-derevenetz:
I have no idea what the `map_actions.*` infrastructure was used for (some "new format map only" ?), but apparently it is not used anymore.


ihhub:
Hi @oleg-derevenetz , it was used in the new map XML format which we scrapped over time.

src/fheroes2/world/world.cpp
```diff
@@ -1408,7 +1390,16 @@ StreamBase & operator>>( StreamBase & msg, World & w )
...
 
     msg >> w.vec_tiles >> w.vec_heroes >> w.vec_castles >> w.vec_kingdoms >> w._rumors >> w.vec_eventsday >> w.map_captureobj >> w.ultimate_artifact >> w.day >> w.week
-        >> w.month >> w.heroes_cond_wins >> w.heroes_cond_loss >> w.map_actions >> w.map_objects >> w._seed;
+        >> w.month >> w.heroes_cond_wins >> w.heroes_cond_loss;
+
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_1002_RELEASE, "Remove the logic below." );
+    if ( Game::GetLoadVersion() < FORMAT_VERSION_1002_RELEASE ) {
+        uint32_t dummy;
+
+        msg >> dummy;
```
ihhub, line 1399: This will work only if `dummy` is 0 but what if it is not? Should we at least add an assertion?

ihhub (changes requested):
Hi @oleg-derevenetz , I left one question here. Could you please check it?

src/fheroes2/world/world.cpp
oleg-derevenetz: Hi @ihhub If `dummy` is no 0, then it will be handled lust like any other corrupted save file. If we add an assertion here, app will crash on the failed assertion instead.

ihhub: In this case can we at least error log this case?

ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for maintaining code quality!

PR #6664, opened 2023-02-17
Fix remaining MSVC warnings

### LINKED ISSUE #6566: Fix remaining warnings for Visual Studio
### Preliminary checks

We have walked a long path starting from more than 1500 compilation warnings for Microsoft Visual Studio builds till the current 34 warnings. To improve the code quality and to avoid any further compilation warnings we should take the final step and fix the remaining warnings. A Half of them are from **smacker.c** file which is not a part of the project's source code but we have no choice since we compile it as a part of the project. We can just silence them by adding extra casting if required.

Fix all the remaining compilation warnings on VS and set finally treat all warnings as errors.

### Additional info

_No response_

oleg-derevenetz:
close #6566

### DIFF
--- a/VisualStudio/common.props
+++ b/VisualStudio/common.props
@@ -9,9 +9,9 @@
       <ConformanceMode>true</ConformanceMode>
       <LanguageStandard>stdcpp17</LanguageStandard>
       <WarningLevel>Level4</WarningLevel>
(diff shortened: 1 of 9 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , awesome job! Now the number of warnings won't grow anymore.

PR #6666, opened 2023-02-17
Fix volcano sounds

### LINKED ISSUE #6650: Adventure map, wrong ambient sound for lava objects
### Preliminary checks

### Platform

Windows

### Describe the bug

Currently two lava objects have wrong sounds.
Volcano 2
Volcano 1
05. wav composition should be played while hero standing near these two objects.

### Save file

[Lava.zip](https://github.com/ihhub/fheroes2/files/10743138/Lava.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #6650

### DIFF
--- a/src/fheroes2/agg/m82.cpp
+++ b/src/fheroes2/agg/m82.cpp
@@ -202,9 +202,10 @@ int M82::FromSpell( const int spellID )
     return UNKNOWN;
 }
 
-// TODO: this function returns a sound track based on a provided tile. It works fine for most of objects as they have only one "main" tile.
-// However, some objects like Oracle or Volcano can be bigger than 1 tile leading to multiple sounds coming from the same object and these
-// sounds might not be synchronized. This is mostly noticeable with 3D Audio mode on.
+// TODO: This function works fine for most of objects as they have only one "main" tile. However,
+// TODO: some objects like Oracle or Volcano can be bigger than 1 tile leading to multiple sounds
+// TODO: coming from the same object and these sounds might not be synchronized. This is mostly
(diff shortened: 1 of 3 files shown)


Branikolog (approved):

src/fheroes2/agg/m82.cpp
```diff
@@ -214,6 +224,19 @@ M82::SoundType M82::getAdventureMapObjectSound( const MP2::MapObjectType objectT
...
+        case MP2::OBJ_ICN_TYPE_UNKNOWN:
+            return UNKNOWN;
+        // Small volcanoes
+        case MP2::OBJ_ICN_TYPE_OBJNLAVA:
+            return LOOP0005;
+        default:
+            break;
+        }
+        // Other volcanoes
+        return LOOP0027;
```
ihhub, line 239: I assume that we are not fixing the problem with multiple tiles having the same object ICN type and object ID as written in the TODO comments above but only separating sounds by object ICN types?

ihhub (commented):
Hi @oleg-derevenetz , I left just one question here. Could you please take a look?

src/fheroes2/agg/m82.cpp
oleg-derevenetz: @ihhub No, that's why I left the corresponding TODO comment in place. Only proper sound depending on the volcano type is implemented.

ihhub: Thank you for the confirmation!

ihhub (approved):

ihhub:

PR #6672, opened 2023-02-18
Fix multiple issues in the Dialog::ArmyInfo()

oleg-derevenetz:
This PR fixes the following issues:

1. CPU hog when viewing quick info on a unit using the right mouse button. How to reproduce:
1.1. Open hero window or any other window where you can view quick info on the unit;
1.2. Hover the mouse over the unit, press and hold the right mouse button;
1.3. Observe the 100% CPU load in Task Manager.

2. Hero is able to upgrade troops during the battle if he is defending a castle, this castle has an upgraded dwelling and there is non-upgraded unit from that dwelling in hero's army. How to reproduce:

https://user-images.githubusercontent.com/32623900/219902090-cf8cb47f-c721-47ef-9681-30083451ede8.mp4

Also I moved all auxiliary functions in the `dialog_armyinfo.cpp` to the anonymous namespace.

### DIFF
--- a/src/fheroes2/dialog/dialog_recruit.cpp
+++ b/src/fheroes2/dialog/dialog_recruit.cpp
@@ -437,13 +437,13 @@ Troop Dialog::RecruitMonster( const Monster & monster0, uint32_t available, cons
         bool skipHotKeyCheck = false;
 
         if ( le.MousePressRight( monsterArea ) ) {
-            ArmyInfo( Troop( monster, available ), READONLY );
+            ArmyInfo( Troop( monster, available ), ZERO );
 
             // Perform a full rendering to properly restore the parts of the screen outside of this dialog
             display.render();
         }
         else if ( le.MouseClickLeft( monsterArea ) ) {
-            ArmyInfo( Troop( monster, available ), READONLY | BUTTONS );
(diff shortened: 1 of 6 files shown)


src/fheroes2/dialog/dialog_recruit.cpp
```diff
@@ -437,13 +437,13 @@ Troop Dialog::RecruitMonster( const Monster & monster0, uint32_t available, cons
...
 
         if ( le.MousePressRight( monsterArea ) ) {
-            ArmyInfo( Troop( monster, available ), READONLY );
+            ArmyInfo( Troop( monster, available ), ZERO );
 
             // Perform a full rendering to properly restore the parts of the screen outside of this dialog
             display.render();
         }
         else if ( le.MouseClickLeft( monsterArea ) ) {
-            ArmyInfo( Troop( monster, available ), READONLY | BUTTONS );
```
ihhub, line 446: `Dialog::READONLY` entry is not used anywhere after these changes. Could you please remove it?

ihhub (changes requested):
Hi @oleg-derevenetz , I left here on comment. Would you mind to take a look at it?

ihhub (approved):

ihhub:

PR #6673, opened 2023-02-19
MSVC: enable link-time optimization in release mode

oleg-derevenetz:
This slows down the build somewhat in release mode, but performs various useful [optimizations](https://learn.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=msvc-160), including cross-module inlining.

### DIFF
--- a/VisualStudio/Release.props
+++ b/VisualStudio/Release.props
@@ -7,8 +7,11 @@
     <ClCompile>
       <Optimization>Full</Optimization>
       <FavorSizeOrSpeed>Speed</FavorSizeOrSpeed>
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz , a little background history: compilation of Windows build 3 years ago was very slow (I think around 15 minutes) so we disabled this optimization in order to make it faster. After multiple years we made a lot of code cleanup so now it is acceptable time even with this option enabled.

ihhub (approved):

ihhub:

PR #6678, opened 2023-02-20
Fix the spell id calculation for a spell scroll

### LINKED ISSUE #6676: Cannot load a specific map
### Preliminary checks

### Platform

Windows

### Describe the bug

With the latest changes (1-2 days ago) I lost an ability to load my test map using latest commit version.
image
The map is still could be loaded using older versions of the engine.

### Save file

[No Name 3.zip](https://github.com/ihhub/fheroes2/files/10781736/No.Name.3.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #6676

Previously, the spell id of a spell scroll was not truncated to its lower 8 bits while loading it from MP2 and therefore may have invalid values, which is wrong. This now has been caught by an assertion.

### DIFF
--- a/src/fheroes2/maps/maps_tiles_quantity.cpp
+++ b/src/fheroes2/maps/maps_tiles_quantity.cpp
@@ -21,6 +21,7 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include <algorithm>
 #include <cassert>
 #include <cstdint>
 #include <limits>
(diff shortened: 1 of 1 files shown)


Branikolog (approved):
Thanks for the prompt fix, @oleg-derevenetz !

ihhub:
Hi @oleg-derevenetz , this pull request is marked as `a work in progress` so I suspect that I should halt my review.

oleg-derevenetz:
Hi @ihhub I just decided to add more sanity checks just in case. Now I marked it as ready for review.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this quick fix!

PR #6680, opened 2023-02-20
Avoid overflow on manipulations with number of monsters on tile, remove static_cast

oleg-derevenetz:
Avoid & log potential overflows when manipulating the number of monsters on the tile (for example, during weekly/monthly growth).


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for these corrections!

PR #6682, opened 2023-02-20
Speed up the creation of a file list in Windows

oleg-derevenetz:
Use `FindFirstFileEx()` with accelerating flags instead of old `FindFirstFile()`. However, there is unlikely to be a strong accelerating effect in the case of map files (especially at the first run), because we need to open each map file separately to read its header. This operation is instant on my Windows (on SSD and even on 7200RPM HDD) as well as on macOS and Ubuntu (running in VM), but on configurations with a slower HDD, it still can take time.


ihhub:
@oleg-derevenetz , thank you very much for this speed up!

PR #6694, opened 2023-02-22
Fix adding an element of a vector to the same vector (golem animation frames copy issue)

### LINKED ISSUE #6689: Iron & Steel Golem animation issues
### Preliminary checks

### Platform

Linux

### Describe the bug

https://user-images.githubusercontent.com/32623900/220469527-f52c9d2b-1339-4b14-9963-5b01abf1aed8.mp4

Looks like they missing some frames. This is reproducible with the following settings:

settings

### Save file

No save file needed.

### Additional info

_No response_

Districh-ru:
fix #6689

https://user-images.githubusercontent.com/113276641/220747514-306bae7c-4054-4a5a-b7aa-2a02dff5b2fd.mp4

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -147,7 +147,11 @@ namespace
 
                 if ( ( resolutions[biggerId].width % resolutions[currentId].width ) == 0 && ( resolutions[biggerId].height % resolutions[currentId].height ) == 0
                      && ( resolutions[biggerId].width / resolutions[currentId].width ) == ( resolutions[biggerId].height / resolutions[currentId].height ) ) {
-                    resolutions.emplace_back( resolutions[currentId].width, resolutions[currentId].height, resolutions[biggerId].width / resolutions[currentId].width );
+                    // IMPORTANT: we MUST do a copy of a vector element if we want to emplace it to the same vector.
+                    const fheroes2::ResolutionInfo currentResolution = resolutions[currentId];
+                    const int32_t biggerResolutionWidth = resolutions[biggerId].width;
+
+                    resolutions.emplace_back( currentResolution.width, currentResolution.height, biggerResolutionWidth / currentResolution.width );
                 }
             }
         }
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -149,9 +149,9 @@ namespace
                      && ( resolutions[biggerId].width / resolutions[currentId].width ) == ( resolutions[biggerId].height / resolutions[currentId].height ) ) {
                     // IMPORTANT: we MUST do a copy of a vector element if we want to emplace it to the same vector.
                     const fheroes2::ResolutionInfo currentResolution = resolutions[currentId];
-                    const int32_t biggerResolutionWidth = resolutions[biggerId].width;
+                    const int32_t scaleFactor = resolutions[biggerId].width / currentResolution.width;
 
-                    resolutions.emplace_back( currentResolution.width, currentResolution.height, biggerResolutionWidth / currentResolution.width );
+                    resolutions.emplace_back( currentResolution.width, currentResolution.height, scaleFactor );
                 }
             }
         }
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -1869,13 +1869,10 @@ namespace fheroes2
                 // Some checks that we really have CP1251 font
                 const int32_t verifiedFontWidth = ( id == ICN::FONT ) ? 19 : 12;
                 if ( imageArray.size() == 162 && imageArray[121].width() == verifiedFontWidth ) {
-                    // IMPORTANT: we MUST do a copy of a vector element if we want to insert it to the same vector.
-                    const fheroes2::Sprite firstImage = imageArray[0];
-
                     // Engine expects that letter indexes correspond to charcode - 0x20.
                     // In case CP1251 font.icn contains sprites for chars 0x20-0x7F, 0xC0-0xDF, 0xA8, 0xE0-0xFF, 0xB8 (in that order).
                     // We rearrange sprites array for corresponding sprite indexes to charcode - 0x20.
-                    imageArray.insert( imageArray.begin() + 96, 64, firstImage );
+                    imageArray.insert( imageArray.begin() + 96, 64, imageArray[0] );
                     std::swap( imageArray[136], imageArray[192] ); // Move sprites for chars 0xA8
                     std::swap( imageArray[152], imageArray[225] ); // and 0xB8 to it's places.
                     imageArray.pop_back();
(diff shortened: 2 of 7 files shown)


oleg-derevenetz:
Hi @ihhub BTW we have a lot of the code like this:

https://github.com/ihhub/fheroes2/blob/3af8445a243f645e5ae6a06c27e61d3040644dc0/src/fheroes2/agg/agg_image.cpp#L1875

or this:

https://github.com/ihhub/fheroes2/blob/3af8445a243f645e5ae6a06c27e61d3040644dc0/src/fheroes2/gui/ui_font.cpp#L142

The problem here in both cases is that `insert()` takes the last parameter by const reference, and if reallocation occurs during an insertion, then this reference will be invalidated.

Districh-ru:
> Hi @Districh-ru it seems that situation is a bit worse then I thought:
> 
> https://en.cppreference.com/w/cpp/container/vector/insert
> 
> Using `insert()` with `first` and `last` iterators pointing into `*this` is undefined behavior in any case (see description of overload N 4), even if no reallocation occurs, so `reserve()` is not enough. It seems that there is no other way than using a temporary vector to be guaranteed to be safe.

Hi @oleg-derevenetz, and what do you think of using `push_back` (or `emplace_back`) in a loop for this case?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2230,9 +2230,11 @@ namespace fheroes2
 
                     _icnVsSprite[id].reserve( golemICNSize + copyFramesNum );
                     // 'MOVE_MAIN' frames starts from the 6th frame in Golem ICN sprites.
-                    const std::vector<fheroes2::Sprite>::const_iterator firstFrameToCopy = _icnVsSprite[id].begin() + 6;
-                    _icnVsSprite[id].insert( _icnVsSprite[id].end(), firstFrameToCopy, firstFrameToCopy + copyFramesNum );
-                    for ( int32_t i = 0; i < copyFramesNum; ++i ) {
+                    size_t copyFrame = 6;
+
+                    for ( int32_t i = 0; i < copyFramesNum; ++i, ++copyFrame ) {
+                        _icnVsSprite[id].push_back( _icnVsSprite[id][copyFrame] );
```
oleg-derevenetz, line 2236: Hi @Districh-ru this solution suffers from the same issue. `push_back()` gets the value to insert by const reference. Suppose that there is no room left and vector is forced to reallocate. Suppose it will reallocate first and only then it will access the reference that is passed as `push_back()` argument.

ihhub:
Hi @Districh-ru and @oleg-derevenetz , this is indeed a huge problem. We must fix it immediately.

src/fheroes2/agg/agg_image.cpp
Districh-ru: Hi @oleg-derevenetz, I changed the code to make a copy of a vector element and then `std::move` it in in `emplace_back` to a vector.

Districh-ru:
Hi @ihhub and @oleg-derevenetz, I also made some changes to other parts of the code where I found `push_back` or `emplace_back` of a vector element to the same vector.

src/fheroes2/campaign/campaign_data.cpp
```diff
@@ -842,7 +843,10 @@ namespace Campaign
             std::string description( monsters.back().GetMultiName() );
 
             while ( monsters.back() != monsters.back().GetUpgrade() ) {
-                monsters.emplace_back( monsters.back().GetUpgrade() );
```
oleg-derevenetz, line 845: There is a different situation: `Monster::GetUpgrade()` returns a temporary, which is then will be added to the vector. So the temporary will be created first, and then inside `emplace_back()` only this temporary will be accessed through the rvalue reference (i.e. moved to the vector), not any of already existing vector elements, so this operation is safe.

src/fheroes2/campaign/campaign_data.cpp
```diff
@@ -856,7 +860,10 @@ namespace Campaign
...
```
oleg-derevenetz, line 859: The same situation as above: `emplace_back()` deals with temporary, not with existing element of `monsters`.

src/fheroes2/dialog/dialog_recruit.cpp
```diff
@@ -368,7 +369,10 @@ Troop Dialog::RecruitMonster( const Monster & monster0, uint32_t available, cons
 
     std::vector<Monster> upgrades = { monster0 };
     while ( upgrades.back().GetDowngrade() != upgrades.back() ) {
-        upgrades.emplace_back( upgrades.back().GetDowngrade() );
```
oleg-derevenetz, line 371: `Monster::GetDowngrade()` returns a temporary, not a reference to some already existing element of `upgrades`, so this operation is safe as well.

src/fheroes2/gui/ui_campaign.cpp
```diff
@@ -155,7 +156,10 @@ namespace fheroes2
             uiElements.emplace_back( monsterUI.back().get() );
 
             while ( monsters.back() != monsters.back().GetUpgrade() ) {
-                monsters.emplace_back( monsters.back().GetUpgrade() );
```
oleg-derevenetz, line 158: The same logic as above - the temporary object is added to the vector, not some element of the same vector.

oleg-derevenetz (changes requested):
Hi @Districh-ru It's really not THAT bad :) I just left a few comments, please take a look when you have time.

src/fheroes2/campaign/campaign_data.cpp
Districh-ru: Hi @oleg-derevenetz, thanks for the explanations.
I reverted the changes in these files.

oleg-derevenetz (approved):

src/engine/screen.cpp
```diff
@@ -147,7 +147,11 @@ namespace
 
                 if ( ( resolutions[biggerId].width % resolutions[currentId].width ) == 0 && ( resolutions[biggerId].height % resolutions[currentId].height ) == 0
                      && ( resolutions[biggerId].width / resolutions[currentId].width ) == ( resolutions[biggerId].height / resolutions[currentId].height ) ) {
-                    resolutions.emplace_back( resolutions[currentId].width, resolutions[currentId].height, resolutions[biggerId].width / resolutions[currentId].width );
+                    // IMPORTANT: we MUST do a copy of a vector element if we want to emplace it to the same vector.
+                    const fheroes2::ResolutionInfo currentResolution = resolutions[currentId];
+                    const int32_t biggerResolutionWidth = resolutions[biggerId].width;
```
ihhub, line 152: Would it makes sense to calculate a scale factor instead of width?

ihhub (changes requested):
Hi @Districh-ru , I left one comment here. Could you please take a look?

src/engine/screen.cpp
Districh-ru: Hi @ihhub, certainly, we can cache the scale factor here.

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for these fixes!

Issue #6600, opened 2023-02-09
Right clicking on empty hero or castle shows Adventure Map information


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Right click on any empty hero or castle slot or just on a scrollbar and observe that info dialog is present for Adventure Map:
![image](https://user-images.githubusercontent.com/19829520/217867343-20abf92a-0fd7-4042-8d6e-6a1a3a516aa0.png)
![image](https://user-images.githubusercontent.com/19829520/217867388-b19b2a9e-0ca2-48f3-88dd-b1a8f0a2a98a.png)

We shouldn't show this at first place as a user is not clicking on Adventure Map itself.

### Save file

Any save file works.

PR #6704, opened 2023-02-23
Fix Right clicking on empty hero or castle shows Adventure Map inform…

slawomir-witkowski:
…ation (#6600)


Branikolog (approved):

ihhub:
@slawomir-witkowski , thank you so much for the fix!

PR #6714, opened 2023-02-25
Display the world date for save files, refactor the game I/O a bit

oleg-derevenetz:
Implement the proposal from #6707

screenshot

There were two main problems because of which it was impossible to make a "naive" implementation and which caused refactoring:

1. The world date is stored in the zipped part of the save file, so we can't just seek & read a few bytes, we should unzip this part first, which is relatively slow;
2. There was no infrastructure to use the save file version in the `struct Maps::FileInfo`, so it was impossible to just add a couple of fields there.

### DIFF
--- a/src/engine/zzlib.cpp
+++ b/src/engine/zzlib.cpp
@@ -90,7 +90,7 @@ namespace
     }
 }
 
-bool ZStreamFile::read( const std::string & fn, const size_t offset /* = 0 */ )
+bool ZStreamBuf::read( const std::string & fn, const size_t offset /* = 0 */ )
 {
     StreamFile sf;
     sf.setbigendian( true );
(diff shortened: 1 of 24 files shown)


ihhub (approved):

Branikolog (approved):
Hi, @oleg-derevenetz 
Great work! I believe, this improvement is very useful.
I noticed, that you've used non white colour for file path string. Was it intended?

oleg-derevenetz:
Hi @Branikolog

> I noticed, that you've used non white colour for file path string.

What do you mean? It is white (see the screenshot in the PR description), but with smaller font size than before, because now there is less space for it, and I doubt that this information is really needed so often that it is worth writing it in large letters.

Branikolog:
@oleg-derevenetz 

> What do you mean? It is white (see the screenshot in the PR description), but with smaller font size than before, because now there is less space for it, and I doubt that this information is really needed so often that it is worth writing it in large letters.

Oh, sorry, it's white indeed. This font together with the background are not so commonly met in the game, so it made an impression for me, it was not white, but beige. Sorry, for bothering you.

I also agreed with your solution to make file path in a small font. This information is helpful only for the first time player knows about it :)

oleg-derevenetz:
@Branikolog BTW I changed the font size for the map path in the map quick info window too - for consistency. The map name and type (SW/PoL) now goes first in large letters and then goes file path in small letters.

ihhub:
@oleg-derevenetz , well done!

PR #6716, opened 2023-02-25
Added numeric speed of monsters to hero dialog fix #6526

### LINKED ISSUE #6526: The speed of creatures in the information window
### Preliminary checks

The speed of creatures in the information window is indicated only by words (Average, Fast, etc.)

I propose to indicate the speed of creatures in the window of information with a number in brackets from 1 to 6

### Additional info

_No response_

slawomir-witkowski:
fix #6526

### DIFF
--- a/src/fheroes2/army/army_troop.h
+++ b/src/fheroes2/army/army_troop.h
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2022                                             *
+ *   Copyright (C) 2019 - 2023                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2009 by Andrey Afletdinov <fheroes2@gmail.com>          *
@@ -77,6 +77,8 @@ public:
 protected:
     friend StreamBase & operator<<( StreamBase &, const Troop & );
     friend StreamBase & operator>>( StreamBase &, Troop & );
+
+    static std::string GetSpeedString( uint32_t speed );
 
     uint32_t count;
 };
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -232,14 +232,8 @@ std::string Battle::Unit::GetShotString() const
 
 std::string Battle::Unit::GetSpeedString() const
 {
-    const uint32_t speedValue = GetSpeed( true, false );
-
-    std::string output( Speed::String( speedValue ) );
-    output += " (";
-    output += std::to_string( speedValue );
-    output += ')';
-
-    return output;
+    const uint32_t speed = GetSpeed( true, false );
+    return Troop::GetSpeedString( speed );
 }
 
 uint32_t Battle::Unit::GetInitialCount() const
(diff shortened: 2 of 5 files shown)


ihhub:
Hi @slawomir-witkowski , in order to connect a corresponding issue please use one of the forms in your pull request description:
fix #<issue ID>
close #<issue ID>

ihhub (approved):

Branikolog:
Hello, @ihhub and @slawomir-witkowski !
Creature info looks well.
I'm concerning about well info, actually. Long phrases look little tight.
image
Also, it's the only place, where brackets are used in this window now.... I don't have any suggestions, how to make it without brackets... Maybe, we can remove numeric speed description at all from this window?

slawomir-witkowski:
One idea would be to do it like this:
speed: 2
Very Slow

Branikolog:
> speed: 2
> Very Slow

Well, I think it's better, that previous implementation. And has same style to other attributes.
So, for me removing the number or placing it in such manner both acceptable.
What do you think, @ihhub ? Shall we leave numeric values for well dialog in such way?

ihhub:
Hi @slawomir-witkowski and @Branikolog , I believe we should remove `(value)` representation for Well to keep esthetical view of the window. In monster's dialog it makes sense to add extra values as hero stats affect monster stats but not in Well. We also need to remember not to make the information cramped.

src/fheroes2/army/army_troop.h
```diff
@@ -68,6 +68,7 @@ class Troop : public Monster
     virtual std::string GetDefenseString() const;
     virtual std::string GetShotString() const;
     virtual std::string GetSpeedString() const;
+    virtual std::string GetSpeedString( uint32_t speed ) const;
```
oleg-derevenetz, line 71: Hi @slawomir-witkowski I don't see any reason for this function to be `virtual`, since it is not designed to be overridden in descendants. It can even be made `static` because it doesn't access any fields, and `protected` instead of `public`, because it is not used anywhere outside the `Troop`->`Battle::Unit` class hierarchy.

src/fheroes2/kingdom/speed.h
```diff
@@ -41,7 +42,7 @@ namespace Speed
         INSTANT = 9
     };
 
-    std::string String( const int speed );
+    std::string String( const uint32_t speed );
```
oleg-derevenetz, line 45: I suggest to revert this change and ignore the corresponding warning about the conversions. This may be addressed separately later... someday.

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -232,14 +232,8 @@ std::string Battle::Unit::GetShotString() const
...
-    const uint32_t speedValue = GetSpeed( true, false );
-
-    std::string output( Speed::String( speedValue ) );
-    output += " (";
-    output += std::to_string( speedValue );
-    output += ')';
-
-    return output;
+    const uint32_t speed = GetSpeed( true, false );
+    return Troop::GetSpeedString( speed );
```
oleg-derevenetz, line 236: To get rid of the issues with hiding methods with the same names (but different arguments) in descendants, you may:

1. Either rename this method;
2. Or add the following:

```cpp
using Troop::GetSpeedString;
```

in the `Battle::Unit` class.

src/fheroes2/kingdom/speed.h
slawomir-witkowski: Ok, I didn't want to touch this file, but my PR fails because of this.

oleg-derevenetz: > Ok, I didn't want to touch this file, but my PR fails because of this.

You can use `static_cast` for now if needed.

ihhub:
@slawomir-witkowski , a bunch of thanks for this update!

PR #6723, opened 2023-02-26
CMake to copy executable to project root

Kazhuu:
Copy fheroes2 executable to project root when done building.

### DIFF
--- a/src/fheroes2/CMakeLists.txt
+++ b/src/fheroes2/CMakeLists.txt
@@ -1,6 +1,6 @@
 ###########################################################################
 #   fheroes2: https://github.com/ihhub/fheroes2                           #
-#   Copyright (C) 2022                                                    #
+#   Copyright (C) 2022 - 2023                                             #
 #                                                                         #
 #   This program is free software; you can redistribute it and/or modify  #
 #   it under the terms of the GNU General Public License as published by  #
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @Kazhuu what do you think about the inverted approach - copy executable to the root of the source directory, like it's done for `make`?

Kazhuu:
That would also work @oleg-derevenetz. I have always preferred to run the executable from the build folder and have seen projects other projects to work like this as well. When for example debugging with GDB. In GDB I can just run `make -j` and then `run` to run the executable again. But yeah it could be done from the root directory as well if you want that instead :+1:

Kazhuu:
I change the the behavior now to copy the executable to project root now. I think it would be good to add that file to `.gitignore` as well. Now the executable is showing up as an untracked file.

oleg-derevenetz:
Hi @Kazhuu,

I just slightly changed your PR. Resulting binary file should not be copied to the `PROJECT_SOURCE_DIR` if it was built for macOS app bundle, because this app bundle has a totally different directory layout. `Makefile` doesn't copy such binary as well:

https://github.com/ihhub/fheroes2/blob/60772c724315724ce395d2198dbb6fed06da1970/Makefile#L38-L43

Kazhuu:
> Hi @Kazhuu,
> 
> I just slightly changed your PR. Resulting binary file should not be copied to the `PROJECT_SOURCE_DIR` if it was built for macOS app bundle, because this app bundle has a totally different directory layout, so you probably won't even be able to run such a binary separately. `Makefile` doesn't copy such binary as well:
> 
> https://github.com/ihhub/fheroes2/blob/60772c724315724ce395d2198dbb6fed06da1970/Makefile#L38-L43

Okay I changed the code so that the executable is only copied when on non-apple platforms.

ihhub (approved):

ihhub:
@Kazhuu , huge thanks for this improvement!

Issue #5822, opened 2022-09-11
Adventure map, Town Portal and Town Gate spells should be displayed in the magic book when the hero is in a boat


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the Town Portal and Town Gate spells should be displayed in the magic book when the hero is in a boat.
This is the case in the original game.

In original game:

![Town spells on boat 001](https://user-images.githubusercontent.com/43583994/189531919-54dc26f8-7702-4e14-ad29-16f1a27c0386.png)

In fheroes2:

![Town spells on boat 006](https://user-images.githubusercontent.com/43583994/189531927-3d621418-c85a-445f-8a6c-f219e5404940.png)

### Save file

Here is the test scenario file:
[0 9956e.zip](https://github.com/ihhub/fheroes2/files/9542700/0.9956e.zip)

### Additional info

Version 0.9.19

felix642:
And how should those spells behave when you're in a boat @LeHerosInconnu? I think that we are currently hiding those spells because the hero is not allowed to use then while in a boat.

LeHerosInconnu:
Hello @felix642,

> And how should those spells behave when you're in a boat @LeHerosInconnu? I think that we are currently hiding those spells because the hero is not allowed to use then while in a boat.

In the original game, when the hero tries to cast one of the Town Portal or Town Gate spells while in a boat, this message is displayed:
![Town spells on boat 002](https://user-images.githubusercontent.com/43583994/189534970-70bf6349-6356-4e20-ad70-e16be95054ac.png)

In fheroes2, when the hero at land tries to cast the Summon Boat spell while is he not adjacent to sea, this message is displayed:
![Town spells on boat 008](https://user-images.githubusercontent.com/43583994/189534975-dece44fe-316a-4f5f-bddb-0e19ed37c601.png)

In fheroes2, when the hero tries to cast the Summon Boat spell when he is already in a boat, this message is displayed:
![Town spells on boat 015](https://user-images.githubusercontent.com/43583994/189535344-f3bfb29f-ed29-447e-8ef6-13459675ba14.png)

[...]

PR #6729, opened 2023-02-27
Gray out summon boat, town gate and town portal spells

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/5822 by graying Town Gate and Town Portal spells instead of not showing them in the spell book while onboard. Also changes Summon Boat spell to gray out when not near water.

### DIFF
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -21,6 +21,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "heroes_base.h"
+
 #include <algorithm>
 #include <cassert>
 #include <vector>
@@ -30,14 +32,13 @@
 #include "castle.h"
 #include "gamedefs.h"
 #include "heroes.h"
-#include "heroes_base.h"
 #include "kingdom.h"
+#include "maps.h"
 #include "race.h"
 #include "serialize.h"
 #include "spell_info.h"
 #include "tools.h"
 #include "translations.h"
-#include "maps.h"
 
 HeroBase::HeroBase( const int type, const int race )
     : magic_point( 0 )
@@ -401,7 +402,7 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
         }
 
         if ( spell == Spell::SUMMONBOAT ) {
-            int32_t boatDestination = fheroes2::getPossibleBoatPosition(hero);
+            int32_t boatDestination = fheroes2::getPossibleBoatPosition( hero );
             if ( !Maps::isValidAbsIndex( boatDestination ) ) {
                 if ( res != nullptr ) {
                     *res = _( "This spell can be casted only nearby water." );
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -402,7 +402,7 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
         }
 
         if ( spell == Spell::SUMMONBOAT ) {
-            int32_t boatDestination = fheroes2::getPossibleBoatPosition( hero );
+            const int32_t boatDestination = fheroes2::getPossibleBoatPosition( *hero );
             if ( !Maps::isValidAbsIndex( boatDestination ) ) {
                 if ( res != nullptr ) {
                     *res = _( "This spell can be casted only nearby water." );
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2022                                             *
+ *   Copyright (C) 2019 - 2023                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2010 by Andrey Afletdinov <fheroes2@gmail.com>          *
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -392,6 +392,13 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
             return false;
         }
 
+        if ( ( spell == Spell::SUMMONBOAT || spell == Spell::TOWNGATE || spell == Spell::TOWNPORTAL ) && hero->isShipMaster() ) {
+            if ( res != nullptr ) {
+                *res = _( "This spell cannot be used on a boat." );
+            }
+            return false;
+        }
+
         if ( spell == Spell::TOWNGATE || spell == Spell::TOWNPORTAL ) {
             const KingdomCastles & castles = hero->GetKingdom().GetCastles();
             bool hasCastles = std::any_of( castles.begin(), castles.end(), []( const Castle * castle ) { return castle && castle->GetHero() == nullptr; } );
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -37,6 +37,7 @@
 #include "spell_info.h"
 #include "tools.h"
 #include "translations.h"
+#include "maps.h"
 
 HeroBase::HeroBase( const int type, const int race )
     : magic_point( 0 )
@@ -397,6 +398,16 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
                 *res = _( "This spell cannot be used on a boat." );
             }
             return false;
+        }
+
+        if ( spell == Spell::SUMMONBOAT ) {
+            int32_t boatDestination = fheroes2::getPossibleBoatPosition(hero);
+            if ( !Maps::isValidAbsIndex( boatDestination ) ) {
+                if ( res != nullptr ) {
+                    *res = _( "This spell can be casted only nearby water." );
+                }
+                return false;
+            }
         }
 
         if ( spell == Spell::TOWNGATE || spell == Spell::TOWNPORTAL ) {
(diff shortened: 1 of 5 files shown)


LeHerosInconnu:
Hello @Kazhuu,

There is a related discussion on this subject here: https://github.com/ihhub/fheroes2/discussions/4845. :)

Kazhuu:
I'll fix the town gate, town portal and summon boat spells in this PR to become grayed out. Then on the issue https://github.com/ihhub/fheroes2/discussions/4845 we fix the rest of the spells.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -301,53 +301,8 @@ namespace
...
-                boatDestination = tileId;
-                break;
-            }
-        }
-
-        if ( !Maps::isValidAbsIndex( boatDestination ) ) {
-            Dialog::Message( "", _( "This spell can be casted only nearby water." ), Font::BIG, Dialog::OK );
-            return false;
-        }
+        int32_t boatDestination = fheroes2::getPossibleBoatPosition( &hero );
```
oleg-derevenetz, line 305: Hi @Kazhuu I recommend to add an assertion here like this:

```cpp
assert( Maps::isValidAbsIndex( boatDestination ) );
```

just in case to catch the potentially invalid use of Summon Boat spell due to some logic error now or in future.

Kazhuu: Good point! I added the assert here.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -301,53 +301,9 @@ namespace
...
```
oleg-derevenetz, line 305: BTW, I believe that this variable can now be `const`?

src/fheroes2/heroes/heroes_base.cpp
```diff
@@ -392,6 +394,23 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
...
 
+        if ( ( spell == Spell::SUMMONBOAT || spell == Spell::TOWNGATE || spell == Spell::TOWNPORTAL ) && hero->isShipMaster() ) {
+            if ( res != nullptr ) {
+                *res = _( "This spell cannot be used on a boat." );
+            }
+            return false;
+        }
+
+        if ( spell == Spell::SUMMONBOAT ) {
+            int32_t boatDestination = fheroes2::getPossibleBoatPosition( hero );
```
oleg-derevenetz, line 405: This variable can be `const`.

oleg-derevenetz (changes requested):
LGTM except a few minor style issues.

src/fheroes2/spell/spell_info.cpp
```diff
@@ -311,4 +316,42 @@ namespace fheroes2
 
         return description;
     }
+
+    int32_t getPossibleBoatPosition( const HeroBase * hero )
+    {
```
oleg-derevenetz, line 321: Also I recommend here to return `-1` immediately if `hero` is `nullptr` or is not a hero, because castle captains are not able to cast adventure spells. Something like this maybe:

```cpp
if ( hero == nullptr || !hero->isHeroes() ) {
    return -1;
}
```

src/fheroes2/heroes/heroes_spell.cpp
Kazhuu: :+1:

src/fheroes2/heroes/heroes_base.cpp
Kazhuu: :+1:

src/fheroes2/spell/spell_info.cpp
Kazhuu: Yes this makes sense. I'll add these. However I was thinking would it better to take HeroBase as a const reference instead of a pointer. This way the first nullptr check can be skipped. However, now the `dynamic_cast` can throw `bad_cast` error. What is the policy of raised errors in this project. I saw few try ... catch blocks used. Or would it better we stick to the pointer here?

oleg-derevenetz: Hi @Kazhuu I just checked this and it seems that we can just get rid of `HeroBase` in this function completely and replace it by `Heroes`, because it is never called for a `HeroBase` instance. This way we get rid of `dynamic_cast` as well. I added the corresponding commit.

oleg-derevenetz (approved):

src/fheroes2/spell/spell_info.cpp
Kazhuu: Yeah much better. Thank you!

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -672,10 +628,10 @@ void Heroes::ActionSpellCast( const Spell & spell )
         apply = ActionSpellDimensionDoor( *this );
         break;
     case Spell::TOWNGATE:
-        apply = isShipMaster() ? false : ActionSpellTownGate( *this );
```
ihhub, line 675: Just in case I would recommend to add assertions in **ActionSpellTownGate() and ActionSpellTownPortal()** that a hero is not on water and the opposite assertion inside **ActionSpellSummonBoat()**. This is to prevent any bad coding in the future.

ihhub (commented):
Hi @Kazhuu , I left just one small suggestion. Could you please take a look when you have free time?

src/fheroes2/heroes/heroes_spell.cpp
Kazhuu: Yes good idea! I added not on water assertions on town gate, town portal and summon boat spells. Not sure what you mean the opposite of hero is not on water to summon boat? Because summon boat can only be used form the land, not on water.

ihhub (approved):

ihhub:

LeHerosInconnu:
Hello @Kazhuu,

I did some tests with the summon boat spell.
It turns out that if there are no more neutral boats (not belonging to any player) or player-built boats available and there is at least one free square adjacent to the hero, the spell is not grayed out, as it should be, and if the hero casts the summon boat spell the message "Summon Boat failed!!!" is displayed.
This is what we want to remove with these improvements.


- Save file: Before boats 02.savh (multiplyer game)(video below).
There are two neutral boats (not belonging to any player) at the beginning of the scenario.
Troyan casts the summon boat spell twice, the two neutral boats are summoned near Troyan.
After that, the summon boat spell is not grayed out in Troyan's magic book, although it should be because there is no other boat to summon.
There is one free square adjacent to Natasha, but there is no additional boat available.
If Troyan casts the non-greyed spell, the message "Summon Boat failed!!!" is displayed. This is what we want to remove with these improvements.

https://user-images.githubusercontent.com/43583994/230159377-38156be6-52ef-4447-becd-e4026053ea6c.mp4

- Save file: Before boats 01.savh (multiplyer game)(video below).
There are four boats available, two neutral boats from the beginning of the scenario and two boats built at the Weed Patch shipyard.
Natasha casts the summon boat spell four times, the four boats are summoned near Natasha.
After that, the summon boat spell is grayed out in Natasha's magic book.
There is no free square adjacent to Natasha (and furthermore, there is no additional boat available).
In this case, everything works as intended.

https://user-images.githubusercontent.com/43583994/230160100-e0c07105-cdc0-4b7b-bfac-7f8352a0c9f4.mp4

- Save file: Before boats 02.savh (multiplyer game)(video below).
There are two neutral boats (not belonging to any player) at the beginning of the scenario.
Troyan casts the summon boat spell twice, the two neutral boats are summoned near Troyan.
After that, the summon boat spell is not grayed out in Troyan's magic book, although it should be because there is no other boat to summon.
There is one free square adjacent to Natasha, but there is no additional boat available.
If Troyan casts the non-greyed spell, the message "Summon Boat failed!!!" is displayed.
A boat is built at the Weed Patch shipyard.
Troyan casts the summon boat spell, the just built boat is summoned near Troyan.
After that, the summon boat spell is grayed out in Troyan's magic book.

https://user-images.githubusercontent.com/43583994/230160243-bf68e704-7641-47fc-ac65-f42bbb8adfad.mp4

- Save file: Before boats 03.savh (multiplyer game)(video below).
There are two neutral boats (not belonging to any player) at the beginning of the scenario.
A boat is built at the Bloodreign shipyard.
Troyan casts the summon boat spell twice, the two neutral boats are summoned near Troyan.
After that, the summon boat spell is not grayed out in Troyan's magic book, although it should be because there is no other boat to summon.
There is one free square adjacent to Troyan, but there is no additional boat available.
If Troyan casts the non-greyed spell, the message "Summon Boat failed!!!" is displayed.

https://user-images.githubusercontent.com/43583994/230160448-89c3153c-aee5-4896-af32-50b6afecd3c0.mp4

- Save file: Before boats 04.savh (multiplyer game)(video below).
There are two neutral boats (not belonging to any player) at the beginning of the scenario.
Troyan casts the summon boat spell twice, the two neutral boats are summoned near Troyan.
After that, the summon boat spell is not grayed out in Troyan's magic book, although it should be because there is no other boat to summon.
There is at least one free square adjacent to Natasha, but there is no additional boat available.
If Troyan casts the non-greyed spell, the message "Summon Boat failed!!!" is displayed.
A boat is built at the Bloodreign shipyard.
Troyan casts the summon boat spell, the just built boat is summoned near Troyan.
Then Troyan casts the summon boat spell and boats are built at the Bloodreign shipyard until all free squares adjacent to Troyan are occupied by boats.
After that, the summon boat spell is grayed out in Troyan's magic book.

https://user-images.githubusercontent.com/43583994/230160582-fa3030b1-8401-4b9e-9914-9f5bac8a97c4.mp4

Here are the save files (multiplyer game):
[Before boats 01 04.zip](https://github.com/ihhub/fheroes2/files/11161581/Before.boats.01.04.zip)

If the hero tries to cast the spell while it is grayed out, a message is displayed depending on the case.
Something like this:
Yellow title for all message windows:
"Summon Boat"

Window Texts;
If there is no boat available:
"There is no boat available to cast this spell."

If there is no free square near the hero:
"There is no free location at sea near the hero to cast this spell."

If there is no boat available and no free square near the hero:
"There is no boat available and no free location at sea near the hero to cast this spell."

Kazhuu:
hello @LeHerosInconnu could you make another issue with the same description? I could fix this separately. Commented here it gets lost easily because this PR is already merged.

PR #6737, opened 2023-02-28
Allow the player to refuse to level up in the Tree of Knowledge even if this level up is for free

oleg-derevenetz:
Implement the proposal from #6725

Also fix the experience calculation for AI-controlled heroes: #5697 fixed the experience calculation only for human-controlled heroes, but not for AI-controlled ones.


ihhub (approved):

ihhub:
I merged it since the majority decided to go with this solution.

@oleg-derevenetz , thank you very much handling this change!

PR #6742, opened 2023-03-01
Tune sounds for guarded creature dwellings and streamline the logic

### LINKED ISSUE #6735: Adventure map, guarded creature dwellings, activation sound for the battle dialog window
### Preliminary checks

Currently, activation sound for Dragon city, City of the dead and Troll bridge plays only when hiring dialog window appears.
image

When we visit these objects for the first time (or when these objects were previously captured by other players) no sounds are played.
image

I suggest in a case these objects belong to other players or you visit them for the first time, play the current activation sound only on battle dialog window. This behaviour is consistent with other objects, where player should fight to guardians.
image
After winning the battle there could be no activation sound, since you're in object currently.

### Additional info

_No response_

oleg-derevenetz:
close #6735

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1223,30 +1223,34 @@ namespace
     void AIToDwellingBattleMonster( Heroes & hero, const MP2::MapObjectType objectType, const int32_t tileIndex )
     {
         Maps::Tiles & tile = world.GetTiles( tileIndex );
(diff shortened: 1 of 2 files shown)


Branikolog (approved):
Works like a charm!
Thanks, @oleg-derevenetz !

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these valuable changes!

PR #6743, opened 2023-03-02
Fix battle AI tactics estimation

idshibanov:
Currently AI sitting in a castle defensively when enemy's archer ratio is too low. Aggressive behaviour should be the default stance and during castle siege AI should stay within the wall only if archers are stronger rather than looking at the ratios in the first place.

Also reducing walls archer penalty because strength isn't all about the damage dealt.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -524,13 +524,13 @@ namespace AI
                 _defendingCastle = true;
                 _myShooterStr += towerStr;
                 if ( !attackerIgnoresCover )
-                    _enemyShooterStr /= 2;
+                    _enemyShooterStr /= 1.5;
             }
             else {
                 _attackingCastle = true;
                 _enemyShooterStr += towerStr;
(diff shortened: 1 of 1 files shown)


src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -548,21 +548,13 @@ namespace AI
...
-        }
 
         // When we have in X times stronger army than the enemy we could consider it as an overpowered and we most likely will win.
         const bool myOverpoweredArmy = _myArmyStrength > _enemyArmyStrength * overPowerRatio;
         const double enemyArcherRatio = _enemyShooterStr / _enemyArmyStrength;
 
-        double enemyArcherThreshold = 0.75;
-        if ( _defendingCastle ) {
-            // Don't make shooters to kill us while we are standing in the castle.
-            enemyArcherThreshold /= 2;
```
idshibanov, line 562: Since the threshold is ignored during the castle defence this isn't needed anymore.

oleg-derevenetz (approved):

Branikolog (approved):

PR #6744, opened 2023-03-03
AI using town portal

idshibanov:
This one is difficult to test on a regular build since those spells are high level and rare. One suggestion is to give all AI heroes the portal spell using `hero->AppendSpellToBook( Spell::TOWNPORTAL, true );` snippet as a test build change.


Alucard648:
As for testing, you could edit a map to give enemy heroes spell scroll with Town Portal apell and Town Gate spell.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1939,6 +1950,29 @@ namespace AI
...
+            hero.FadeOut();
+        }
+
+        hero.Move2Dest( targetIndex );
+        hero.SpellCasted( townPortal );
+        hero.GetPath().Reset();
+
+        if ( AIHeroesShowAnimation( hero, AIGetAllianceColors() ) ) {
+            Interface::Basic::Get().GetGameArea().SetCenter( hero.GetCenter() );
+            hero.FadeIn();
```
oleg-derevenetz, line 1970: I suppose that we need to update the radar here as well if player transferred his control to AI, just like it done for the Dimension Door spell in the `HeroesCastDimensionDoor()`, right, @Districh-ru?

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -458,14 +460,15 @@ void AIWorldPathfinder::reset()
...
         _isArtifactBagFull = false;
     }
 }
 
 void AIWorldPathfinder::reEvaluateIfNeeded( const Heroes & hero )
 {
-    auto currentSettings = std::tie( _pathStart, _pathfindingSkill, _currentColor, _remainingMovePoints, _maxMovePoints, _armyStrength, _isArtifactBagFull );
-    const auto newSettings = std::make_tuple( hero.GetIndex(), static_cast<uint8_t>( hero.GetLevelSkill( Skill::Secondary::PATHFINDING ) ), hero.GetColor(),
+    auto currentSettings = std::tie( _hero, _pathStart, _pathfindingSkill, _currentColor, _remainingMovePoints, _maxMovePoints, _armyStrength, _isArtifactBagFull );
+    const auto newSettings = std::make_tuple( &hero, hero.GetIndex(), static_cast<uint8_t>( hero.GetLevelSkill( Skill::Secondary::PATHFINDING ) ), hero.GetColor(),
```
oleg-derevenetz, line 471: Since `_hero` is added only in the `AIWorldPathfinder` itself before the `_armyStrength`, I propose to move the `_hero` field here at the position before the `_armyStrength` as well to match the order of declarations.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -475,10 +478,10 @@ void AIWorldPathfinder::reEvaluateIfNeeded( const Heroes & hero )
     }
 }
 
-void AIWorldPathfinder::reEvaluateIfNeeded( const int start, const int color, const double armyStrength, const uint8_t skill, const bool isArtifactBagFull )
+void AIWorldPathfinder::reEvaluateIfNeeded( const int start, const int color, const double armyStrength, const uint8_t skill )
 {
-    auto currentSettings = std::tie( _pathStart, _pathfindingSkill, _currentColor, _remainingMovePoints, _maxMovePoints, _armyStrength, _isArtifactBagFull );
-    const auto newSettings = std::make_tuple( start, skill, color, 0U, 0U, armyStrength, isArtifactBagFull );
+    auto currentSettings = std::tie( _hero, _pathStart, _pathfindingSkill, _currentColor, _remainingMovePoints, _maxMovePoints, _armyStrength, _isArtifactBagFull );
+    const auto newSettings = std::make_tuple( nullptr, start, skill, color, 0U, 0U, armyStrength, false );
```
oleg-derevenetz, line 484: I propose to move the `_hero` field here at the position before the `_armyStrength` as well.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1905,15 +1905,26 @@ namespace AI
...
-            if ( MP2::isActionObject( hero.GetMapsObject(), hero.isShipMaster() ) )
+            const Route::Step & step = path.front();
+            const int32_t targetIndex = step.GetIndex();
+
+            if ( step.GetFrom() != targetIndex && world.GetTiles( targetIndex ).GetObject() == MP2::OBJ_CASTLE ) {
+                HeroesCastTownPortal( hero, targetIndex );
+            }
+            else if ( MP2::isActionObject( hero.GetMapsObject(), hero.isShipMaster() ) ) {
                 hero.Action( hero.GetIndex(), true );
+            }
```
oleg-derevenetz, line 1916: How exactly this works (I mean, the `else if` branch)? If `path` is not empty, but its `GetFrontDirection()` is `UNKNOWN`, then the first step of that path is not near the current hero's position, but we react to this by using the action object at the hero's position. Is this logic intended for Stone Liths?

idshibanov: Yes, it's the original logic for Stone Liths. I could put a comment there.

oleg-derevenetz: Yes, please, it would be nice to describe the logic in a comment.

src/fheroes2/ai/ai_hero_action.cpp
Districh-ru: Hi @oleg-derevenetz  and @idshibanov.
The control could be transferred to AI only in debug versions, so this part of radar rendering was not optimized and radar map is always fully updated in:
https://github.com/ihhub/fheroes2/blob/7129e9cf12b3885d96820534a4de3ee6034e5eaa/src/fheroes2/game/game_startgame.cpp#L686-L689
I think there is no need of placing the radar update calls in AI actions as normal AI actions are not shown on radar (mini-map) and in release version transferring control to AI is not available.

oleg-derevenetz: @Districh-ru I'm asking this because there is the corresponding code in the `HeroesCastDimensionDoor()`:

https://github.com/ihhub/fheroes2/blob/7129e9cf12b3885d96820534a4de3ee6034e5eaa/src/fheroes2/ai/ai_hero_action.cpp#L1928-L1937

Is this code within `WITH_DEBUG` not needed anymore? Maybe it should be removed then?

Districh-ru: @oleg-derevenetz, thanks for the recall, I forgot about this radar redraw call in `HeroesCastDimensionDoor()` :)
I analyzed the behavior of AI and can propose to keep the existing `SetRedraw( Interface::REDRAW_RADAR )` in `ai_hero_action.cpp`:
 - on line 1934 it update the radar after hero move
 - and on line 1866 the code is needed for case if AI uses the Dimention door consequently.
 
And the line 688 in `game_startgame.cpp` could be removed as it updates the map at the start of the turn and the AI turn begins with a move. (I was wrong thinking that it updates the map on every move, sorry).

For the current function `HeroesCastTownPortal()` we can skip adding this code as AI will make a move by this (or other) hero after teleporting to a town. 

There is no code for radar update in Teleport/Whirlpool functions and the game correctly redraws the radar by ai_hero_action.cpp line 1934 because the AI makes a move after using a teleport. For human player we need to redraw the radar after every such action as the player can cast a spell or use the teleport and then take a look on map without making a move.

Districh-ru: But if there can be a case when AI will consequently use portal to several towns it would be good to add radar update code here (like in lines 1863-1868).

oleg-derevenetz: @Districh-ru talking about simplicity, is there a way to update the radar just in one place instead of multiple ones for AI-auto-controlled heroes as well? As you can see, updating it in multiple places looks a little difficult because after a while you start to doubt whether you need to add a radar update in a certain place or not.

Districh-ru: I'll try to find such a place and open a new PR for AI controlled heroes radar update.

oleg-derevenetz (approved):

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1905,15 +1906,27 @@ namespace AI
...
         }
     }
 
     void HeroesCastDimensionDoor( Heroes & hero, const int32_t targetIndex )
     {
         const Spell dimensionDoor( Spell::DIMENSIONDOOR );
-        if ( !Maps::isValidAbsIndex( targetIndex ) || !hero.CanCastSpell( dimensionDoor ) ) {
+        if ( !Maps::isValidAbsIndex( targetIndex ) || hero.isShipMaster() != world.GetTiles( targetIndex ).isWater() ) {
```
ihhub, line 1925: Please move this if-condition at the very beginning of the method to avoid extra variable creation if case of early return of the function.

src/fheroes2/world/world_pathfinding.h
```diff
@@ -162,6 +162,7 @@ class AIWorldPathfinder : public WorldPathfinder
     void setSpellPointReserve( const double reserve );
 
 private:
+    void processWorldMap() override;
```
ihhub, line 165: Please add an empty line after this one to have better visual separation  between method's declarations.

ihhub (commented):
Hi @idshibanov , I left 2 small suggestions here. Would you mind to check them in your spare time?

ihhub (approved):

ihhub:
@idshibanov , huge thanks for this improvement in AI!

PR #6745, opened 2023-03-03
Clean up AI priority tasks if attacker was killed

idshibanov:
Currently AI could rush to castle defence even if enemy hero threatening it was taken out. Adding more information to track & finally taking advantage of attack tasks.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal.h
+++ b/src/fheroes2/ai/normal/ai_normal.h
@@ -103,6 +103,26 @@ namespace AI
         uint32_t patrolDistance = 0;
     };
 
+    struct PriorityTask
+    {
+        PriorityTaskType type = PriorityTaskType::ATTACK;
+        double threatLevel = 0.0;
+        std::vector<int> secondary;
+
+        PriorityTask() = default;
+        PriorityTask( PriorityTaskType t, double threat )
+            : type( t )
+            , threatLevel( threat )
+        {}
+
+        PriorityTask( PriorityTaskType t, double threat, int secondaryTask )
+            : type( t )
+            , threatLevel( threat )
+        {
+            secondary.push_back( secondaryTask );
+        }
+    };
+
     struct BattleTargetPair
     {
         int cell = -1;
@@ -258,6 +278,8 @@ namespace AI
         double getFighterObjectValue( const Heroes & hero, const int index, const double valueToIgnore, const uint32_t distanceToObject ) const;
         double getCourierObjectValue( const Heroes & hero, const int index, const double valueToIgnore, const uint32_t distanceToObject ) const;
         int getCourierMainTarget( const Heroes & hero, double lowestPossibleValue ) const;
+
+        void updatePriorityTargets( Heroes & hero, int32_t tileIndex, const MP2::MapObjectType objectType );
 
         bool purchaseNewHeroes( const std::vector<AICastle> & sortedCastleList, const std::set<int> & castlesInDanger, int32_t availableHeroCount,
                                 bool moreTasksForHeroes );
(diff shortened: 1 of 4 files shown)


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1694,6 +1694,33 @@ namespace AI
...
+                _priorityTargets.erase( tileIndex );
+            }
+            else if ( task.type == PriorityTaskType::ATTACK ) {
+                // check if battle was actually won or attacker still there
+                const Heroes * attackHero = world.GetTiles( tileIndex ).GetHeroes();
+                const Castle * attackCastle = world.getCastleEntrance( Maps::GetPoint( tileIndex ) );
+
+                if ( !attackHero && ( !attackCastle || attackCastle->GetColor() == hero.GetColor() ) ) {
+                    for ( const int secondaryTask : task.secondary ) {
+                        _priorityTargets.erase( secondaryTask );
```
oleg-derevenetz, line 1716: Is there a possibility that one of the secondary tasks will have the same tile index as the primary task? In this case `task` reference will be invalidated right during iteration over the `task.secondary`. As far as I understand, this should not happen now, but I propose to add `assert( secondaryTask != tileIndex )` before this line just in case.

oleg-derevenetz (commented):
Hi @idshibanov I just left an optional proposal, could you please take a look?

src/fheroes2/ai/normal/ai_normal.h
```diff
@@ -103,6 +104,26 @@ namespace AI
         uint32_t patrolDistance = 0;
     };
 
+    struct PriorityTask
+    {
+        PriorityTaskType type = PriorityTaskType::ATTACK;
+        double threatLevel = 0.0;
+        std::vector<int> secondary;
```
ihhub, line 111: Are we expecting to have more than 1 task on a single tile? Should we use **std::set**?

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1694,6 +1694,34 @@ namespace AI
...
+            }
+            else if ( task.type == PriorityTaskType::ATTACK ) {
+                // check if battle was actually won or attacker still there
+                const Heroes * attackHero = world.GetTiles( tileIndex ).GetHeroes();
+                const Castle * attackCastle = world.getCastleEntrance( Maps::GetPoint( tileIndex ) );
+
+                if ( !attackHero && ( !attackCastle || attackCastle->GetColor() == hero.GetColor() ) ) {
+                    for ( const int secondaryTask : task.secondary ) {
+                        assert( secondaryTask != tileIndex );
+                        _priorityTargets.erase( secondaryTask );
```
ihhub, line 1717: Since **_priorityTargets** is a map we can break from this loop afterwards.

ihhub (changes requested):
Hi @idshibanov , I put several comments in this pull request. Could you please take a look?

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: Although _priorityTargets is a map, there could be multiple castles threatened by the same hero, so we need to clean it up anyway. At the same time thinking about this I'll have to cover another edge case (multiple heroes beside 1 castle) so I'll update this logic shortly.

src/fheroes2/ai/normal/ai_normal.h
idshibanov: I could switch to a set to make it more explicit for the future.

ihhub (approved):

Districh-ru (approved):
I played one test map and one original map making threats for AI player and allowing to kill my hero and the AI heroes returned to their deals after defeating my hero.
👍

oleg-derevenetz (approved):

PR #6746, opened 2023-03-03
Summon boat spell will only summon friendly boats fix #4109

### LINKED ISSUE #4109: Adventure map, player can summon a boat belonging to opponent
### Preliminary checks

### Platform

Windows

### Describe the bug

While playing multiplayer game one player disembarks, and the other player could summon the boat, that belongs to the first player.
image
image

### Save file

2 player hotseat save file.
[Summon Opponents boat.zip](https://github.com/ihhub/fheroes2/files/7072490/Summon.Opponents.boat.zip)

### Additional info

_No response_

slawomir-witkowski:


oleg-derevenetz:
This approach is problematic. First of all, `World::CaptureObject()` et cetera are designed to work with a limited number of static objects that are amenable to capture, such as castles, mines and so on. In case of boats, literally every water tile near a coast could be "captured" this way. Besides that, the proposed logic only captures these tiles and never releases them back, even if there is no boat on this tile anymore.

slawomir-witkowski:
I get your point.  But we need a place to store boat color for a tile index. How about adding:
 std::map<int32_t, int> _boatColors; 
to World class?

oleg-derevenetz:
> I get your point. But we need a place to store boat color for a tile index. How about adding: std::map<int32_t, int> _boatColors; to World class?

Either this way or maybe store the boat color in the `Maps::Tiles` class itself. It is also necessary to think over a mechanism for clearing this value when there is no boat anymore.

src/fheroes2/castle/castle.cpp
```diff
@@ -2492,16 +2492,19 @@ bool Castle::BuyBoat() const
         kingdom.OddFundsResource( PaymentConditions::BuyBoat() );
 
         left.setBoat( Direction::RIGHT );
+        left.setEmptyBoatColor( kingdom.GetColor() );
```
oleg-derevenetz, line 2495: I have an impression that we could combine the `setBoat()` and `setEmptyBoatColor()` in one function by adding the `color` parameter to the `setBoat()`.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -576,6 +576,9 @@ namespace
         hero.SetShipMaster( true );
         hero.GetPath().Reset();
 
+        // Boat is no longer empty so we reset color to default
+        world.GetTiles( dst_index ).resetEmptyBoatColor();
```
oleg-derevenetz, line 580: This should be also performed in case of AI-controlled hero in the `ai/ai_hero_action.cpp`.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -590,6 +593,8 @@ namespace
         hero.ResetMovePoints();
         hero.Move2Dest( dst_index );
         world.GetTiles( fromIndex ).setBoat( Maps::GetDirection( fromIndex, dst_index ) );
+        world.GetTiles( fromIndex ).setEmptyBoatColor( hero.GetColor() );
```
oleg-derevenetz, line 596: This should be also performed in case of AI-controlled hero in the `ai/ai_hero_action.cpp` (once you change the `setBoat()` signature, you'll find these additional places automatically).

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -347,21 +347,28 @@ namespace
         if ( !Maps::isValidAbsIndex( boatDestination ) ) {
             Dialog::Message( "", _( "This spell can be casted only nearby water." ), Font::BIG, Dialog::OK );
             return false;
+            return false;
```
oleg-derevenetz, line 350: Why is there an additional return here?

src/fheroes2/maps/maps_tiles.h
```diff
@@ -468,6 +483,9 @@ namespace Maps
 
         // This field does not persist in savegame.
         uint32_t _region = REGION_NODE_BLOCKED;
+
+        // Empty boats can be summoned, we need to remember which player they belong to
+        int _emptyBoatColor = Color::NONE;
```
oleg-derevenetz, line 488: Please note that this field should be saved into the save file, therefore, you need to add the necessary functionality. Also I recommend to put it before the `_region` to outline that `_emptyBoatColor` should be saved.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -347,21 +347,28 @@ namespace
...
         }
 
         for ( const int32_t boatSource : Maps::GetObjectPositions( center, MP2::OBJ_BOAT, false ) ) {
             assert( Maps::isValidAbsIndex( boatSource ) );
 
+            Maps::Tiles & tileSource = world.GetTiles( boatSource );
+            int boatColor = tileSource.getEmptyBoatColor();
+            bool isFriendlyBoat = boatColor == hero.GetColor() || boatColor == Color::NONE;
```
oleg-derevenetz, line 358: `boatColor` and `isFriendlyBoat` can be `const`.

oleg-derevenetz (changes requested):
Hi @slawomir-witkowski I just left a few suggestions, please take a look when you have time.

oleg-derevenetz:
Hi @slawomir-witkowski there are merge conflicts. Please merge the latest changes from the `master` branch to this PR and resolve these conflicts.

oleg-derevenetz (approved):
Hi @slawomir-witkowski I added a few changes, mostly minor ones, except that I added the new save file version to properly support the intermediate save files that were created between 1.0.1 and this PR. LGTM.

oleg-derevenetz:
BTW I'm not 100% sure about this, but probably in the original game boats that belong to the vanquished kingdom should become neutral again. This should be verified.

slawomir-witkowski:
Hi @oleg-derevenetz I didn't think of vanquished kingdom case. But we should be able to reset _emptyBoatColor for affected ships.

oleg-derevenetz:
@slawomir-witkowski

> I didn't think of vanquished kingdom case.

It seems that even vanquished kingdom retains its boats. In the following video boat belongs to the green player and green hero is able to summon it, but the blue hero cannot summon this boat even after the green player is already vanquished:

https://user-images.githubusercontent.com/32623900/222926703-55198c14-6274-45fb-aca0-ccffc361b963.mp4

I even waited another day on purpose, but it didn't help.

slawomir-witkowski:
Yes, I verified this as well. In OG no boat stealing even from dead players :)

LeHerosInconnu:
Hello @slawomir-witkowski and @oleg-derevenetz,

I never tested this in the original game.
And yes, I think the boats of an eliminated player should become neutral and be able to be called with the summon boat spell from that point on.

Resource generators become neutral again and can be captured by visiting the mine, as in the beginning of the game.
Boats become neutral again and can be captured by boarding the boat, and using the summon boat spell, as in the beginning of the game.

oleg-derevenetz:
> And yes, I think the boats of an eliminated player should become neutral and be able to be called with the summon boat spell from that point on.

We tested the OG behavior and in OG boats not become neutral when the kingdom is vanquished. Maybe that's for a reason. I suggest to replicate the OG behavior in this PR (I.e. left everything as is), consider the behavior proposed by you as an improvement, discuss it and address in a separate PR.

LeHerosInconnu:
Hello @oleg-derevenetz,

> > And yes, I think the boats of an eliminated player should become neutral and be able to be called with the summon boat spell from that point on.
> 
> We tested the OG behavior and in OG boats do not become neutral when the kingdom is vanquished. Maybe that's for a reason.

I think I know the reason. :D
It is called:
"Forgetfulness"

> I suggest to replicate the OG behavior in this PR (I.e. left everything as is), consider the behavior proposed by you as an improvement, discuss it and address in a separate PR.

Okay, I'll open an issue for that.

oleg-derevenetz:
@LeHerosInconnu

> I think I know the reason. :D
> It is called:
> "Forgetfulness"

This is very doubtful, because the same behavior has been preserved at least in HoMM3:

https://user-images.githubusercontent.com/32623900/222963899-ad8a1a65-ce96-4117-80ad-549fbf396e47.mp4

and maybe in later games of HoMM series as well (can't test this right now).

LeHerosInconnu:
@oleg-derevenetz.

> @LeHerosInconnu
> 
> > I think I know the reason. :D
> > It is called:
> > "Forgetfulness"
> 
> This is very doubtful, because the same behavior has been preserved at least in HoMM3:

What could be a coherent reason for this functioning?
Personally, I don't see any.

The spell also works differently in Heroes 3.
From level advanced the spell can create a boat (which does not exist on the adventure map).

> and maybe in later games of HoMM series as well (can't test this right now).

I can't say anything about that either.
I haven't played Heroes 4.
I only tried the Heroes 5 demo briefly.

ihhub:
Hi @slawomir-witkowski , @LeHerosInconnu and @oleg-derevenetz , let's not disturb vanished players here and change if needed this behavior in a separate pull request keeping changes here minimal for easier review :)

src/fheroes2/maps/maps_tiles.h
```diff
@@ -462,6 +472,9 @@ namespace Maps
 
         bool tileIsRoad = false;
 
+        // Heroes can only summon neutral empty boats or empty boats belonging to their kingdom.
+        int _emptyBoatColor = Color::NONE;
```
ihhub, line 476: Please make this member as **uint8_t** as **int** takes 4 bytes and we don't use all of them. This might be small but on big maps we can have 60K difference in memory usage.

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1649,7 +1651,8 @@ std::string Maps::Tiles::String() const
     os << std::endl
        << "quantity 1      : " << static_cast<int>( quantity1 ) << std::endl
        << "quantity 2      : " << static_cast<int>( quantity2 ) << std::endl
-       << "add. metadata   : " << additionalMetadata << std::endl;
+       << "add. metadata   : " << additionalMetadata << std::endl
+       << "empty boat color: " << Color::String( _emptyBoatColor ) << std::endl;
```
ihhub, line 1655: We should show this only for **MP2::OBJ_BOAT** as this information makes no sense for most of tiles. Moreover all elements in the printed data are aligned with **:** symbol while this entry is by 1 character longer.

src/fheroes2/maps/maps_tiles.h
ihhub, line 476: Logically speaking this is not the color of an empty boat because we don't show a flag of the owner. I suggest to rename logic according to something like **boatOwnerColor**.

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -352,15 +352,19 @@ namespace
         for ( const int32_t boatSource : Maps::GetObjectPositions( center, MP2::OBJ_BOAT, false ) ) {
             assert( Maps::isValidAbsIndex( boatSource ) );
 
+            Maps::Tiles & tileSource = world.GetTiles( boatSource );
+            const int boatColor = tileSource.getEmptyBoatColor();
+            const bool isFriendlyBoat = boatColor == hero.GetColor() || boatColor == Color::NONE;
```
ihhub, line 357: Please add parenthesis for both conditions for easier code reading. I would also suggest to swap them as the second condition is much faster in execution.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -587,6 +587,9 @@ namespace
         hero.SetShipMaster( true );
         hero.GetPath().Reset();
 
+        // Boat is no longer empty so we reset color to default
+        world.GetTiles( dst_index ).resetEmptyBoatColor();
```
ihhub, line 591: Should we move this logic inside **Maps::Tiles::setAsEmpty()** so that we don't forget to update this is we cleanup a tile for boats?

ihhub (changes requested):
Hi @slawomir-witkowski , I added few comments in this pull request. Would you mind check them in your spare time?

src/fheroes2/heroes/heroes_action.cpp
oleg-derevenetz: @ihhub `setAsEmpty()` is not used when a hero gets into the boat because this tile does not become empty.

src/fheroes2/maps/maps_tiles.cpp
oleg-derevenetz: @ihhub

> Moreover all elements in the printed data are aligned with : symbol while this entry is by 1 character longer.

How it is longer? The colon is in the same place in all cases.

ihhub: Oh, you're right! It's time for me to sleep then😐

src/fheroes2/heroes/heroes_action.cpp
ihhub: You're absolutely right. I'm curious: the number of boats is unlimited, isn't it?

oleg-derevenetz: > You're absolutely right. I'm curious: the number of boats is unlimited, isn't it?

I don't see any reason for them to be limited. Technically they could be placed on every water tile.

ihhub: In this case we have no choice but to keep boat information in every tile.

oleg-derevenetz: @ihhub We can use those existing quantities to not allocate a special storage for this info. Can you tell how exactly to store this color in this case?

ihhub: Hi @oleg-derevenetz , this actually is my next step of code preparation for the Editor. Right now we use quantity1 and quantity2 in some quirky way that nobody understands. We can't simply use them as there could be a case that a boat can stand on an object which requires this additional information like Derelict Ship.

src/fheroes2/heroes/heroes_spell.cpp
slawomir-witkowski: > Please add parenthesis for both conditions for easier code reading. I would also suggest to swap them as the second condition is much faster in execution.

We always calculate distance. Do you think we should only calculate it for friendly boats?
I was wondering, if performance gain is worth nesting if statements.

src/fheroes2/heroes/heroes_action.cpp
oleg-derevenetz: @ihhub We still can add this reset into the `Tiles::setAsEmpty()` and do not call the `resetEmptyBoatColor()` when "Summon Boat" spell is used. At least it won't hurt.

src/fheroes2/heroes/heroes_spell.cpp
oleg-derevenetz: I suppose we can add the `if ( !isFriendlyBoat ) { continue; }` before than `GetStraightLineDistance()` will be called and then check only `distance`.

oleg-derevenetz:
Hi @slawomir-witkowski I just added a commit with minor changes which reduce the number of `static_cast`s and impose a range checking on the color value during the conversion `int`->`uint8_t` just in case.

Branikolog:
> > @LeHerosInconnu
> > > I think I know the reason. :D
> > > It is called:
> > > "Forgetfulness"
> > 
> > 
> > This is very doubtful, because the same behavior has been preserved at least in HoMM3:
> 
> What could be a coherent reason for this functioning? Personally, I don't see any.

I agree with @LeHerosInconnu. Initially boats have two states: _neutral_ and _belonging to one player_.
After player has vanquished it seems like the new boat state is appearing: abandoned one, which couldn't be summoned by anyone, but could be captured. But do we need that to exist? There could be rised a question by players in a case they defeated all players on a map, but still can't summon the boat they watch on map. After player is gone it looks quite natural, when all his mines, lighthouses and boats become neutral. 
I think we don't need making things so complicated having that type of "tied netutral" boats.

oleg-derevenetz:
> There could be rised a question by players in a case they defeated all players on a map, but still can't summon the boat they watch on map.

That's how it works in the original game, and not in just one, but in at least multiple games of HoMM series (if not all), which a sort of suggests that it can be made for a reason and changing this may break some maps for example.

LeHerosInconnu:
Hello @oleg-derevenetz,

> > There could be rised a question by players in a case they defeated all players on a map, but still can't summon the boat they watch on map.
> 
> That's how it works in the original game, and not in just one, but in at least multiple games of HoMM series (if not all), which a sort of suggests that it can be made for a reason and changing this may break some maps for example.

I don't see how this can alter a scenario as all ships on the adventure map are neutral at the beginning of the game.

oleg-derevenetz:
> I don't see how this can alter a scenario as all boats on the adventure map are neutral at the beginning of the game.

But new boats that are built in the Shipyard and used by a particular player are not.

Branikolog:
Hi, @oleg-derevenetz 

> But new boats that are built in the Shipyard and used by a particular player are not.

That's a valid point. However I find the existing logic not really intuitive. But if you insist on keeping it closer to the original (as this could bring some benefits or penalties during gameplay) I don't have any explicit objections despite of my personal desire in just doing the logic simpler and more obvious to a common player.

>By the way, in the scenario editor of fheroes2, it should be possible to assign the ownership of a boat at the creation of the scenario.

👍

ihhub (approved):

ihhub:
@slawomir-witkowski , thank you so much for this change!

PR #6750, opened 2023-03-03
Add virtual keyboard support for touch devices

ihhub:
This is a base code to support multi-language virtual keyboard for mobile devices such as Android. Right now only English is supported, LANG button is disabled for now. ~~Also we will need to add missing 7, 8 and 9 symbols as button font is used for buttons.~~

This feature is applicable for Android, PS Vita and Switch devices.

relates to #6008 

image

image

image


oleg-derevenetz:
I suggest to remove the "LANG" button for now, because I anticipate questions from users like "why can't I switch the language?"

ihhub:
> I suggest to remove the "LANG" button for now, because I anticipate questions from users like "why can't I switch the language?" and add it later when this feature will be fully supported.

Hi @oleg-derevenetz , I disabled the button for now. We need it to at least align the last row properly.

ihhub:
Hi @zenseii , could you please add 7, 8 and 9 letters to button's font in a separate pull request if you have spare time?

ihhub:
I will try to add 7, 8 and 9 letters tomorrow.

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -288,6 +290,8 @@ bool Dialog::InputString( const std::string & header, std::string & res, const s
     LocalEvent & le = LocalEvent::Get();
     le.OpenVirtualKeyboard();
```
oleg-derevenetz, line 291: I suppose we need to do something with external virtual keyboard here and in other similar places - I mean, there should be no situation when we show both the system virtual keyboard and the our own at the same time. Currently `System::isHandheldDevice()` does not intersect with `LocalEvent::OpenVirtualKeyboard()`, but It would be good to make the logic more explicit. Maybe add something like `System::isNativeVirtualKeyboard()`...?

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -0,0 +1,544 @@
...
+
+        std::function<DialogAction( KeyboardRenderer & )> action;
+
+        fheroes2::ButtonSprite button;
+
+        // This is used only for buttons which should have pressed state for some layouts.
+        bool isInvertedRenderingLogic{ false };
+    };
+
+    std::vector<std::vector<uint8_t>> getNumericCharacterLayout( const fheroes2::SupportedLanguage language )
```
oleg-derevenetz, line 194: Why `uint8_t` here and in other similar places, and not just `char`?

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -0,0 +1,544 @@
...
+        return {};
+    }
+
+    std::vector<std::vector<uint8_t>> getNonCapitalCharacterLayout( const fheroes2::SupportedLanguage language )
+    {
+        auto layout = getCapitalCharacterLayout( language );
+        for ( auto & letters : layout ) {
+            for ( uint8_t & letter : letters ) {
+                std::string singleCharString( 1, static_cast<char>( letter ) );
+                singleCharString = StringLower( singleCharString );
```
oleg-derevenetz, line 225: Since we have the single-character buttons (I mean, they are not multibyte), we can just `StringLower()` them all at once: convert `letters` to the `std::string()`, then apply `StringLower()` to the whole string and then replace individual letters using the new characters from that string instead of creating a special `std::string` for every single character.

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -0,0 +1,544 @@
...
+        }
+
+        return buttons;
+    }
+
+    void addExtraEnglishButtons( std::vector<std::vector<KeyboardButton>> & buttons, const LayoutType layoutType, const bool isEvilInterface )
+    {
+        switch ( layoutType ) {
+        case LayoutType::LowerCase:
+            buttons.back().emplace_back( _( "Keyboard|BACK" ), 60, isEvilInterface, []( KeyboardRenderer & renderer ) {
```
oleg-derevenetz, line 276: It would be good to verify somewhere here that target vector (`buttons` in this particular case) is not empty before calling `back()`, because if it is empty, then it is undefined behavior. And maybe use intermediate variables like `lastButtonRow` or something to not call `back()` everywhere.

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -0,0 +1,544 @@
...
+                                    []( const KeyboardRenderer & ) { return DialogAction::UpperCase; } );
+
+            buttons.emplace_back();
+            buttons.back().emplace_back( _( "Keyboard|123" ), 60, isEvilInterface, []( const KeyboardRenderer & ) { return DialogAction::Numeric; } );
+            buttons.back().emplace_back( _( "Keyboard|SPACE" ), 160, isEvilInterface, []( KeyboardRenderer & renderer ) {
+                renderer.appendCharacter( ' ' );
+                return DialogAction::AddLetter;
+            } );
+            buttons.back().emplace_back( _( "Keyboard|LANG" ), 60, isEvilInterface, []( const KeyboardRenderer & ) { return DialogAction::DoNothing; } );
+            buttons.back().back().button.disable();
```
oleg-derevenetz, line 290: In C++17, `emplace_back()` returns a reference to the inserted element (and `emplace()` returns an iterator pointing to the emplaced element) which might be handy in cases like this (to not write these cryptic `back().back()` and `back().front()`) :)

oleg-derevenetz (commented):
Hi @ihhub I just left a few proposals, please take a look when you have time :)

Districh-ru (commented):
Hi @ihhub!
For the first glance I have a couple of small suggestions:
1. For me the buttons are small enough (on my cell phone) and we have enough space (640x480 px) for the bigger keyboard window. So I suggest to enlarge the buttons or to add extra space between them.
<details><summary>fheroes2 on a cell phone</summary>


</details><details><summary>Keyboard that I use on a cell phone (for comparison)</summary>


</details>

2. Maybe besides of adding "7", "8", and "9" we could add the symbols for Shift key ("▲") and Backspace ("◄"). Or rename them to something like "CAPS L" and "BKSP" - this is discussible.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: Ah, this is a leftover from the initial implementation. I will fix it.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: Moreover, we can return `std::string` from layouts since we work with single-char strings.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: Yup. I just wanted to not overcomplicate code but I will try to redo it a bit.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: That's a good point. I will add checks and assertions.

src/fheroes2/dialog/dialog_selectcount.cpp
ihhub: I was thinking about this. Ideally we should have only one virtual keyboard. I will ask players who use Vita to show how the keyboard looks like.

Districh-ru (changes requested):
Hi @ihhub, I found a small bug. Could you address it please when you have time. :)

The "UPP" button flashes twice when pressed:

https://user-images.githubusercontent.com/113276641/222890586-3fa2cdf4-b156-4a00-b934-3e816a1dcb53.mp4

The same behavior is on my cell phone, but not noticeable as at this time the button is covered with my finger. :)

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+namespace
+{
+    const int32_t buttonOffset{ 5 };
+    const int32_t defaultButtonWidth{ 20 };
+    const int32_t defaultButtonHeight{ 25 };
+    const fheroes2::Size windowSize{ 420, 250 };
+    const fheroes2::Point offsetFromWindowBorders{ 25, 50 };
+    const fheroes2::Size inputAreaSize{ 268, 21 };
+    const int32_t inputAreaOffset{ 2 };
```
Districh-ru, line 56: When we are talking about handheld devices we have to take into account that they have higher density of pixels, so the resulting image is smaller than on PC monitor.
I ran fheroes2 on my cell phone in 640x480 and found that the keyboard buttons are located to close to each other IMHO.
I changed the keyboard parameters and got this variant, which is better for me. (This variont was tested only on PC).
изображение
<details><summary>The changed parameters</summary>

```cpp
    const int32_t buttonOffset{ 10 };
    const int32_t defaultButtonWidth{ 35 };
    const int32_t defaultButtonHeight{ 35 };
    const fheroes2::Size windowSize{ 560, 330 };
    const fheroes2::Point offsetFromWindowBorders{ 25, 70 };
```

</details>

But I can set the native phone resolution 1440x720 and will get a small keyboard:
изображение

So I can also propose not to use the constants here, but calculate the variables according to the current resolution. This could be done in a new PR. :)

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -0,0 +1,565 @@
...
+            lastButtonRow->emplace_back( _( "Keyboard|LANG" ), 60, isEvilInterface, []( const KeyboardRenderer & ) { return DialogAction::DoNothing; } );
+            lastButtonRow->back().button.disable();
+            break;
+        case LayoutType::UpperCase:
+            lastButtonRow->emplace_back( _( "Keyboard|BACK" ), 60, isEvilInterface, []( KeyboardRenderer & renderer ) {
+                renderer.removeLastCharacter();
+                return DialogAction::Backspace;
+            } );
+            lastButtonRow->emplace( lastButtonRow->begin(), _( "Keyboard|UPP" ), 60, isEvilInterface,
+                                    []( const KeyboardRenderer & ) { return DialogAction::LowerCase; } );
```
Districh-ru, line 309: I suggest placing the Caps Lock and Backspace buttons on the bottom row next to the space (123 and LANG can be placed to the left and right of OKAY), as other languages can have more than 7 letters in the bottom row, for example in Russian in each row of the keyboard there are 2 letters more than in English.

Districh-ru (commented):
Hi @ihhub. I have a couple of suggestions for the keyboard appearance. Could you take a look in your spare time.

LeHerosInconnu:
Hello @ihhub,

> This feature is applicable for Android, PS Vita and Switch devices.

This could possibly be useful on a PC too. :)
This could be configurable, at least, in the game settings file.
What do you think?

ihhub:
> Hello @ihhub,
> 
> > This feature is applicable for Android, PS Vita and Switch devices.
> 
> This could possibly be useful on a PC too. :) This could be configurable, at least, in the game settings file. What do you think?

This can be done in a separate pull request. This PR is a basis for the future work so we need to keep the changes minimally required here.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: I will update the parameters to have bigger buttons but we need to keep in mind that the number of letters per row can be much more as you pointed out in a separate comment. We have to preserve the size of buttons for any layout.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: I took an idea from my phone's layout :) I think these nitpicks we can do in a separate pull request as the code structure allows to easily change any layout.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: I enlarged the window keeping some space for longer letter rows. This is just a bit shorter than your proposal. Ideally for big resolutions we have to resize the whole window, not only buttons as edit box should be visible as well.

ihhub:
Hi @Districh-ru , what do you think about LANG button? Should we create an icon for it as well? On my phone it is a globe for example.

Districh-ru:
> Hi @Districh-ru , what do you think about LANG button? Should we create an icon for it as well? On my phone it is a globe for example.

Hi @ihhub, yes we can make something like globe.
In the next two days I can try to make something looking like a globe in 10x10 (or 12x12) pixels and put it instead of another unused letter. It will look like this: изображение   
изображение

But in current PR we can fully hide this button as currently we support only English. And also disabled LANG button may produce questions from users - "how to enable it?", or "maybe something is wrong with my game files", so I propose to hide it if there is no other language to be switched.

Districh-ru (changes requested):
Hi @ihhub, I have one comment. Could you please address it when you have time:
According to [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) and the [Linux naming rules](https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/) the ":" symbol should be removed from our keyboard for naming files or there would be a save error.

oleg-derevenetz:
@Districh-ru

> According to [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) and the [Linux naming rules](https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/) the ":" symbol should be removed from our keyboard for naming files or there would be a save error.

There is no problem with colon on *NIX:

```
oleg@gentoo ~/Temp/xx $ ls -l
total 0
-rw-r--r-- 1 oleg oleg 0 Mar  5 20:41 :aaa:
```

And even though there will be problems with it on Windows, this will simply lead to the appearance of the "There was an issue during saving" dialog. Nothing terrible will happen.

Districh-ru:
> There is no problem with colon on *NIX:
> And even though there will be problems with it on Windows, this will simply lead to the appearance of the "There was an issue during saving" dialog. Nothing terrible will happen.

@oleg-derevenetz, thanks for clarifying this!
So as the keyboard will be used only on handheld devices (not on Windows) we can leave the colon.
If in future the keyboard will be used on Windows we could remove the colon symbol only for this OS.

Districh-ru (approved):

ihhub:
Hi @oleg-derevenetz and @Districh-ru , I am okay to remove LANG button but the space button might be centered. What are you thoughts?

ihhub:
Hi @Branikolog please review this pull request as the highest priority as it is must have for the release.

oleg-derevenetz:
> I am okay to remove LANG button but the space button might be centered. What are you thoughts?

Can't this button be hidden/not drawn? And as a last resort we can make two shift buttons :)

Districh-ru:
I agree with @oleg-derevenetz proposal.
If we hide LANG button there will be a blank space between Space and Backspace - I'm OK with it.

Branikolog:
@ihhub 
Also, backspace and shift arrows for evil interface are brown. Is it possible to make the more "grayish"?

ihhub:
Hi @oleg-derevenetz and @Districh-ru , I hid LANG button as well as UP button for Numeric layout. Please take a look at the screenshots.

ihhub:
> @ihhub Also, backspace and shift arrows for evil interface are brown. Is it possible to make the more "grayish"?

I believe we need to modify palette used for this operation.

ihhub:
> @ihhub Also, backspace and shift arrows for evil interface are brown. Is it possible to make the more "grayish"?

It is fixed as well.

oleg-derevenetz:
@ihhub

> I hid LANG button as well as UP button for Numeric layout. Please take a look at the screenshots.

IMHO LGTM. I'm OK with the blank space too.

Districh-ru (approved):
👍

Branikolog (approved):

PR #6752, opened 2023-03-04
Rework the radar (mini-map) update calls for AI auto controlled player

Districh-ru:
relates to #6744

The "AI auto control mode" is currently available only in debug version of fheroes2 for AI debugging purposes.
For human player we made special mini-map updates after some actions, which does not include hero movement as the player can do this action and then wants to overview the world and think about the next action. AI don't do this. So mini-map update is needed only after hero move.
The `.SetRedraw( Interface::REDRAW_RADAR )` for "AI auto control mode" player is called from `Heroes::Scoute(()` method.
It could possibly be placed in `Maps::ClearFog()`, for example, but to do this we have to transfer there the player/hero ID and to mix different logic.


oleg-derevenetz (approved):

ihhub (approved):

ihhub:

PR #6754, opened 2023-03-04
Fix zero delays for Battlefield animations to reduce CPU usage at 10th battle speed

Districh-ru:
If you set the Battle speed to 10 and start a battle with Zombies or Genies or Ghosts you can notice the high CPU usage on master build.
This PR makes the maximum Battle speed (10th) to use the half of 9th speed delays.
The custom unit movement delay is capped at 1 ms from the bottom.
 Visually the battle movement preserves the same as before the PR.

### DIFF
--- a/src/fheroes2/game/game_delays.cpp
+++ b/src/fheroes2/game/game_delays.cpp
@@ -34,6 +34,8 @@ namespace
 {
     std::vector<fheroes2::TimeDelay> delays( Game::LAST_DELAY + 1, fheroes2::TimeDelay( 0 ) );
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):
LGTM. Just one thing - I recommend to add here:

https://github.com/ihhub/fheroes2/blob/7129e9cf12b3885d96820534a4de3ee6034e5eaa/src/fheroes2/game/game_delays.cpp#L32-L36

a static assertion that `DEFAULT_BATTLE_SPEED` is more than 0 and less than 10.

ihhub (approved):

ihhub:

PR #6758, opened 2023-03-04
Fix AI being stuck on watering hole

idshibanov:
We missed this on a recent refactor. Currently AI hero gets stuck in a loop if such object is on the map.


ihhub (approved):

ihhub:

PR #6764, opened 2023-03-05
zzlib.cpp: avoid extra reinterpret_cast conversions, check the results of static_cast conversions

oleg-derevenetz:
Code smell fixes.

PR #6767, opened 2023-03-06
zzlib.cpp: avoid potential unsigned overflow when calculating the initial buffer size

oleg-derevenetz:
Follow-up to #6764


ihhub:

PR #6770, opened 2023-03-06
Rename "scoute"

oleg-derevenetz:
Don't use this wrong word, replace it with "scout" as an action, use "scouting distance" instead of "scoute value", etc, etc.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for this typo fix!

PR #6775, opened 2023-03-07
Add missing part of Captain's Quarter in Sorceress Town

### LINKED ISSUE #3843: Add missing Captain's Quarters part in Sorceress town as a separate image
If a Sorceress town does not have built castle but contains Captain's Quarters it's visible that a part of the Quarters is missing. This is the issue in the original game assets. We would like to add missing portion of Quarters but creating an extra image which will be a part of fheroes2 assets. In other words we need to draw an image which will be attached to existing Quarters sprite making it as a completed building rather than truncated in the middle.

This is the task for artists but anyone is welcome to try. Once you're done with image creation either post it here or approach ihhub in [Discord](https://discord.com/invite/xF85vbZ) for assistance in converting image into readable by game format.

image

ihhub:
close #3843

all credits go to @golden-lining

image


oleg-derevenetz (approved):

Districh-ru (commented):
Great work, @golden-lining!
I have one suggestion/question. Cold you please address it when you have time.

The original image (the right part of the Captain's Quarter) as I think has a shadow on the roof from the castle. So maybe it'll be good to make it a bit brighter.
изображение -> изображение
(I edited the image by copy-paste, so it only represents the idea, not the real result)

What do you think? :)

src/fheroes2/castle/castle_building_info.cpp
```diff
@@ -188,7 +188,7 @@ namespace
         case BUILD_CASTLE:
             return { 0, 0, 201, 179 };
         case BUILD_CAPTAIN:
-            return { 223, 122, 37, 52 };
+            return { 204, 122, 37, 52 };
```
Districh-ru, line 191: If we move the `x` coordinate to еду left by 19 px we also have to increase the `width` by the same value as the right part of the sprite retains on the same place. :)
```suggestion
            return { 204, 122, 37, 71 };
```

As I checked the changed ROI does not overlap with others, but IMHO it should be different for Town and Castle as for the Castle case.
In this PR a click on the right part of the Castle wall gives the info about the Captain Quarters.
Maybe we have to adjust all ROIs for all Castle buildings in a separate PR.

Districh-ru (changes requested):
Hi @ihhub! I left one suggestion for building ROI. Could you have a look when you have time.

The updated sprite for Captain's Quarter looks great. Well done @golden-lining! 👍

src/fheroes2/castle/castle_building_info.cpp
ihhub: Sure. I will revert this change.

Districh-ru (approved):

PR #6779, opened 2023-03-07
Remove the MSVC 2017 support

### LINKED ISSUE #6578: Discontinue Visual Studio 2017 project files
### Preliminary checks

I think we reached the point where nobody is using Visual Studio 2017. I asked in Discord whether anyone is still using it. If not then we will proceed with removing it. However, it is not required right now as no harm to be done if we keep it.

Just remove VS 2017 files.

### Additional info

_No response_

oleg-derevenetz:
close #6578


ihhub:
@oleg-derevenetz , but how about your dev setup?

undef21:
Hello, @oleg-derevenetz. Does recent MSVC (>= 2019) support Platform SDK 8.1? If no, do we need WIN32_LEAN_AND_MEAN macro in source code? AFAIK, you added this macro to support Win 8.1 + VS2017

oleg-derevenetz:
@undef21

> Does recent MSVC (>= 2019) support Platform SDK 8.1? If no, do we need WIN32_LEAN_AND_MEAN macro in source code? AFAIK, you added this macro to support Win 8.1 + VS2017

While we maybe don't need it anymore (strictly speaking, i.e. most likely there will be no build issues even without this macro), it may be still useful because defining it [excludes](https://learn.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers#faster-builds-with-smaller-header-files) a bunch of useless (in our case) APIs and header files inside the `windows.h`, so I personally don't see a reason to remove it right now.

ihhub:
@oleg-derevenetz , thank you very much for this change! If there would be a need to revert it back we can do it with ease.

PR #6780, opened 2023-03-08
Fix mini-map update when interacting with map objects (Teleport, Artifact, Mine, Resource, Hero)

### LINKED ISSUE #6810: Adventure map, Haunt spell, incorrect right-click information window for mine squares other than the active mine square (version 1.0.1 6774)
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, there is incorrect right-click information window for mine squares other than the active mine square.

In fheroes2:

https://user-images.githubusercontent.com/43583994/224418699-e2202650-6972-4ea1-abd6-83be1ac50526.mp4

### Save file

[0 9930.zip](https://github.com/ihhub/fheroes2/files/10945393/0.9930.zip)

### Additional info

_No response_

Districh-ru:
This PR fixes the absence of mini-map update before entering a Teleport or Whirlpool.

The condition used to check if Hero has moved to a Teleport does not work properly so I moved the mini-map redraw after every hero move to `heroes_move.cpp`.

fheroes2:

https://user-images.githubusercontent.com/113276641/223682822-7c3bd128-9f65-442b-825d-48ab74be2f2f.mp4

This PR:

https://user-images.githubusercontent.com/113276641/223691078-1787beca-26b3-4073-bbe3-bf4c622e1a0b.mp4

UPD:
This PR also makes mini-map to update right after hero move in hero scout area.
If hero action leads to mini-map objects change then mini-map is updated only in the area of changed objects (capture of mine, resource or artifact, defeat a hero).
Previously the mini-map update was made only after hero move and action assuming that the action object will be in hero scout area.

UPD2:
This PR also relates to #6808 
And fixes #6810

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -1076,10 +1076,6 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
                             gameArea.SetCenter( hero->GetCenter() );
                             ResetFocus( GameFocus::HEROES );
 
-                            // Update the radar map image in the area that is visible to the hero after his movement.
-                            radar.SetRenderArea( hero->GetScoutRoi() );
-                            radar.SetRedraw( REDRAW_RADAR );
-
                             RedrawFocus();
 
                             if ( stopHero ) {
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -763,8 +763,11 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
     }
 
     radar.SetHide( false );
+    // Force set radar ROI for the whole world as there could unused ROI from the last allied AI hero move in Hot Seat mode.
+    radar.SetRenderArea( { 0, 0, world.w(), world.h() } );
     statusWindow.Reset();
     gameArea.SetUpdateCursor();
+
     Redraw( REDRAW_GAMEAREA | REDRAW_RADAR | REDRAW_ICONS | REDRAW_BUTTONS | REDRAW_STATUS | REDRAW_BORDER );
 
     fheroes2::Display & display = fheroes2::Display::instance();
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -763,8 +763,6 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
     }
 
     radar.SetHide( false );
-    // Force set radar ROI for the whole world as there could unused ROI from the last allied AI hero move in Hot Seat mode.
-    radar.SetRenderArea( { 0, 0, world.w(), world.h() } );
     statusWindow.Reset();
     gameArea.SetUpdateCursor();
 
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -1787,25 +1787,22 @@ namespace
 
         assert( world.GetTiles( index_to ).GetObject() != MP2::OBJ_HEROES );
 
+        Interface::Basic & I = Interface::Basic::Get();
+        // Update the radar map image after possible hero move before entering a Teleport.
+        I.GetRadar().SetRenderArea( hero.GetScoutRoi() );
+        I.SetRedraw( Interface::REDRAW_RADAR );
+
         AudioManager::PlaySound( M82::KILLFADE );
         hero.GetPath().Hide();
         hero.FadeOut();
 
-        Interface::Basic & I = Interface::Basic::Get();
         const fheroes2::Point fromPoint = Maps::GetPoint( index_from );
-        if ( hero.GetCenter() == fromPoint ) {
-            // If hero was already in Teleport and player hit the space bar.
-            I.GetRadar().SetRenderArea( { fromPoint.x, fromPoint.y, 1, 1 } );
-        }
-        else {
-            // Before entering a Teleport the hero may make a move into it, so we update the radar map image of this move.
-            I.GetRadar().SetRenderArea( hero.GetScoutRoi() );
-        }
 
         // No action and no penalty
         hero.Move2Dest( index_to );
 
         // Clear the previous hero position
+        I.GetRadar().SetRenderArea( { fromPoint.x, fromPoint.y, 1, 1 } );
         I.Redraw( Interface::REDRAW_RADAR );
 
         I.GetGameArea().SetCenter( hero.GetCenter() );
@@ -1834,25 +1831,22 @@ namespace
             return;
         }
 
+        Interface::Basic & I = Interface::Basic::Get();
+        // Update the radar map image after possible hero move before entering a Whirlpool.
+        I.GetRadar().SetRenderArea( hero.GetScoutRoi() );
+        I.SetRedraw( Interface::REDRAW_RADAR );
+
         AudioManager::PlaySound( M82::KILLFADE );
         hero.GetPath().Hide();
         hero.FadeOut();
 
-        Interface::Basic & I = Interface::Basic::Get();
         const fheroes2::Point fromPoint = Maps::GetPoint( index_from );
-        if ( hero.GetCenter() == fromPoint ) {
-            // If hero was already in Whirlpool and player hit the space bar.
-            I.GetRadar().SetRenderArea( { fromPoint.x, fromPoint.y, 1, 1 } );
-        }
-        else {
-            // Before entering a Whirlpool the hero may make a move into it, so we update the radar map image of this move.
-            I.GetRadar().SetRenderArea( hero.GetScoutRoi() );
-        }
 
         // No action and no penalty
         hero.Move2Dest( index_to );
 
         // Clear the previous hero position
+        I.GetRadar().SetRenderArea( { fromPoint.x, fromPoint.y, 1, 1 } );
         I.Redraw( Interface::REDRAW_RADAR );
 
         I.GetGameArea().SetCenter( hero.GetCenter() );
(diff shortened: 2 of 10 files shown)


Branikolog:
Hello, @Districh-ru !
Minimap doesn't updates after disembarking. Only after another hero moves, the dot corresponding disembarked hero moves.

Districh-ru:
> Hello, @Districh-ru !
> Minimap doesn't updates after disembarking. Only after another hero moves, the dot corresponding disembarked hero moves.

Hello, @Branikolog !
Thanks for checking. I'll fix this bug a little later today.

ihhub:
Hi @Districh-ru , was the issue pointed by @Branikolog addressed?

src/fheroes2/game/game_startgame.cpp
```diff
@@ -763,8 +763,11 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
     }
 
     radar.SetHide( false );
+    // Force set radar ROI for the whole world as there could unused ROI from the last allied AI hero move in Hot Seat mode.
+    radar.SetRenderArea( { 0, 0, world.w(), world.h() } );
```
ihhub, line 767: I notice one possible issue with radar code:
```cpp
void Interface::Radar::Redraw( const bool redrawMapObjects )
{
...
if ( _hide ) {
   ...
}
else {
   if ( redrawMapObjects ) {
       // Only here we update ROI.
       RedrawObjects( Players::FriendColors(), ViewWorldMode::OnlyVisible );
   }
}
...
}
```
It looks like we have 2 places where ROI is not reset properly:
- when radar is hidden
- when we don't want to render map objects

I think we should update the ROI for them. What do you think?

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -605,6 +610,12 @@ namespace
         AudioManager::PlaySound( M82::KILLFADE );
         hero.GetPath().Hide();
         hero.FadeIn( fheroes2::Point( offset.x * Game::HumanHeroAnimSkip(), offset.y * Game::HumanHeroAnimSkip() ) );
+
+        // Clear hero position marker from the boat and scout the area on radar after disembarking.
+        Interface::Basic & I = Interface::Basic::Get();
+        I.GetRadar().SetRenderArea( hero.GetScoutRoi( false ) );
```
ihhub, line 616: I am just curious: **ignoreDirection** is used to expand radar ROI to be update. While I see the benefit not dong extra calculations the code is quite complex to maintain. What do you think if we always return enlarged by 1 ROI for hero's scouting?

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1634,10 +1652,17 @@ namespace
...
+                Interface::Basic & I = Interface::Basic::Get();
+
+                I.GetGameArea().runSingleObjectAnimation( std::make_shared<Interface::ObjectFadingOutInfo>( tile.GetObjectUID(), tile.GetIndex(), tile.GetObject() ) );
 
                 tile.QuantityReset();
+
+                const fheroes2::Point artifactPosition = Maps::GetPoint( dst_index );
+
+                // Update the position of picked up artifact on radar to remove its mark.
+                I.GetRadar().SetRenderArea( { artifactPosition.x, artifactPosition.y, 1, 1 } );
```
ihhub, line 1664: In the future we need to use another approach in order to make the code flexible enough and to avoid explicit calls of radar. For instance, a callback function when a hero starts a movement, changes their position or have changes in their skills or inventory. We don't need to address it right now but it might be a long term solution for us.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1972,7 +1983,37 @@ namespace
 
                 tile.QuantitySetColor( hero.GetColor() );
 
-                Interface::Basic::Get().Redraw( Interface::REDRAW_GAMEAREA );
+                // Set the radar update ROI according to captured object size and position.
+                fheroes2::Rect radarRoi( Maps::GetPoint( dst_index ), { 1, 1 } );
```
ihhub, line 1987: I understand the idea to minimize the size of radar's ROI to be updated but we shouldn't do such micro optimization as we could have a chance that a new capturing object would be added and somebody forget to update this code. I suggest to just update ROI which is equivalent of hero's scout area. The same applies for Haunt spell.
Moreover, we can just update radar ROI on a completion of any action, avoiding corner cases like for pickup resources, whirlpools and etc. What do you think?

src/fheroes2/maps/maps_tiles.h
```diff
@@ -377,6 +377,7 @@ namespace Maps
         static std::pair<uint32_t, uint32_t> GetMonsterSpriteIndices( const Tiles & tile, const uint32_t monsterIndex );
         static void PlaceMonsterOnTile( Tiles & tile, const Monster & mons, const uint32_t count );
         static void UpdateAbandonedMineSprite( Tiles & tile );
+        void setAbandonedMineObjectType() const;
```
ihhub, line 380: Please make this function static. We would like to move out all specific to objects functions out of the class.

ihhub (changes requested):
Hi @Districh-ru , I put multiple suggestions in this pull request. I am not sure that all of them are valid so analyze them carefully. Could you please check them in your free time?

Districh-ru:
> Hi @Districh-ru , was the issue pointed by @Branikolog addressed?

Hi, @ihhub. Yes, @Branikolog's issue was addressed in https://github.com/ihhub/fheroes2/pull/6780/commits/a75edf20d3adc6354f640fa335f4d03250ecae27. :)

src/fheroes2/game/game_startgame.cpp
Districh-ru: Hi @ihhub, you are right. I've protected from one possible case. I suddenly got it yesterday, and reproduced it only once.
I'll apply your suggestion as it will protect us from all possible cases including possible asynchronous setting of ROI and calling Redraw.

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: You mean always act like `ignoreDirections = true` ?
We can do this, the direction is only used for hero moving to the next tile. It is most often action, but enlarging the ROI will not drastically reduce the game performance. As after making proper radar updates for every hero action I thought to make (in a separate PR) a check "wasFogCleared" and if not - update only hero position (2x2 rect).

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: Such function, which will give us tile numbers of updated tiles will be very useful not only for radar, but for adventure map also.
Currently resources and artifacts are located in one tile so we update the radar only for that tile. If we do not plan to bring the bigger objects to the game this will work for every resource/artifact. But in that case we will have the "big" object position and size ad we will than use something like: ` I.GetRadar().SetRenderArea( { artifact.x, artifact.y, artifact.width, artifact.height } );
And IMHO there is no need to update the whole hero scout ROI if we have to change only one tile and we know where it is .:)

Putting radar call into `tile.QuantityReset()` is also a not a good idea, but `QuantityReset()` might call some other UI function...

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: Than we have to decide what we want:
- Have an unified function witch is launched for about 90 cases of action objects in hero scout area and is needed only in about 10 cases in mostly in 1x1 or 2x2 area.
- Have explicit calls for radar but only when it is needed but with more complicated logic.
Before this PR the Scout ROI was updated every time after on the hero move after hero actions. But we have tracked all actions  and added all needed radar calls.
Also if we had 80 of 90 cases were the radar update is needed than the better solution will be to call it every time rather than calling it in each of 80 calls.
Talking about spells - it is other action and redrawing radar after every possible spell is also is not a solution.

And if we will add some new action object with a completely new action function and if we will display it on radar, yes we will have to make it properly render including removal after pickup or changing color after capture.

But a good Idea is to make a function that will automatically get the object size and tell it to radar like a reverse function to `Maps::Tiles::getIndexOfMainTile()`
I think such function might be introduced but in a separate PR. I'll ad a `TODO`.

Branikolog (approved):
Good work, @Districh-ru !
Thanks for your valuable contribution. :)

src/fheroes2/heroes/heroes_action.cpp
ihhub: Yes, I meant to just enlarge the whole area by 1 tile no matter if hero stands or walks. Seem like you did this so it is great.

src/fheroes2/heroes/heroes_action.cpp
ihhub: Hi @Districh-ru , evaluating the size of an object is very easy:
- get object ID via **Maps::Tiles::GetObjectUID()** call (or `getObjectIdByObjectIcnType()`)
- put into a stack all nearby tiles and run
- for each tile pushed in the stack run **doesObjectExist()** method
- if the method returns true push all neighbours and update ROI
- repeat till the stack is empty

ihhub: If you think that implementation of this function would take time then please leave a TODO comment in the code for the future change.

ihhub: Actually, you've done this :)

Districh-ru: Thanks for the approach, @ihhub!
Something similar came into my mind. I'll do it in a separate PR.

ihhub (approved):

ihhub:
Hi @oleg-derevenetz and @zenseii could you please review this pull request today if you have time?

oleg-derevenetz (approved):

ihhub:

PR #6791, opened 2023-03-08
Add max lines of translations for new languages

zenseii:
This should hopefully relieve us of some headaches.

Max 30 lines for translations of new languages that don't already exist in the project.


ihhub (approved):

ihhub:
@zenseii , thanks a lot for this documentation change!

Issue #4387, opened 2021-10-11
Combat screen, sometimes it is difficult to switch back to manual mode when Auto combat mode has been activated


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, sometimes it is difficult to switch back to manual mode when Auto combat mode has been activated, the click does not always seem to be taken into account.
I also made a suggestion about this: https://github.com/ihhub/fheroes2/discussions/4385.

In fheroes2:

![Auto combat button 02](https://user-images.githubusercontent.com/43583994/136807758-7dbf5b73-1281-48c9-82ee-289abd238c15.png)

![Auto combat button 01](https://user-images.githubusercontent.com/43583994/136801667-5633b6ed-eb18-4c88-b2aa-9b37c19a897d.png)

Try to switch to manual mode towards the end of the combat.

### Save file

Here is the save file:
[Before last combat 01.zip](https://github.com/ihhub/fheroes2/files/7323060/Before.last.combat.01.zip)

### Additional info

Version 0.9.8 4311

Branikolog:
Hi, @LeHerosInconnu 
Small precision:
It seems to me that turning off is initiated only by clicking back on "AUTO" button, while in the OG we can click wherever  we like.
Also want to note, that both right and left clicks should work for that.

slawomir-witkowski:
You cannot stop auto-combat when oponnent's units are in control. 
You can toggle auto-combat by pressing A. But when you press ESC, you get a confirmation dialog.

PR #6835, opened 2023-03-14
Interrupt auto combat on mouse click anywhere.

slawomir-witkowski:
fix: #4387

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -6011,7 +6011,7 @@ void Battle::Interface::CheckGlobalEvents( LocalEvent & le )
 
     // Interrupting auto battle
     if ( arena.AutoBattleInProgress() && arena.CanToggleAutoBattle()
-         && ( le.MouseClickLeft( btn_auto.area() )
+         && ( ( le.MouseClickLeft() || le.MouseClickRight() )
               || ( le.KeyPress()
                    && ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_SWITCH )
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Currently this PR works like this:

https://user-images.githubusercontent.com/32623900/224969252-3bdabadd-872e-4301-84bc-f47fb814b843.mp4

There is no any auto battle in progress, but it still asks to interrupt the auto-battle if I press `Esc` even during the human player turn. Also we shouldn't even ask to interrupt the auto battle on AI vs AI battles (e.g. in debug mode). Also the case "human vs human" should be verified for sanity, i.e. the question should be asked (and auto-battle is cancelled) only for a current human player (that is, the one whose unit is having its turn at the moment), just like it happens now in the `master` branch.

slawomir-witkowski:
Thank you for your feedback. I wanted to add the ability to cancel auto-combat when it's AI's turn. For players it's not obvious that you can only cancel during your turn. But for now, I only did a small change allowing you to click anywhere to cancel.

ihhub:
Hi @Branikolog , could you please express your opinion about this pull request?

Branikolog:
Hi, @slawomir-witkowski and @ihhub 
The idea to make mouse click to interrupt auto battle is nice, but in the current PR it happens without dialog window with message. I think for both cases (we press Esc button or we press mouse buttons) the same message should appear.
Also, auto battle interruption with Esc button or mouse click should be possible while enemy AI turn. Now you can do it only during your own turn which makes you unable to stop the battle up to the moment the creature you want to control manually starts makes an action.

slawomir-witkowski:
Ok guys, I will close this PR not to block others from picking up this issue. @Branikolog's requirements are perfectly reasonable. It will require some effort to do this right.  I may revisit this in the future.

PR #6844, opened 2023-03-15
SP now calculated with SPR after mouse click on icon

### LINKED ISSUE #6838: Magic Book, cost in spell points is not correct in the spell information window when the hero has the artifact that reduces the cost of elemental spells
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the cost in spell points is not correct in the spell information window when the hero has the artifact that reduces the cost of elemental spells.
The cost should be adjusted to "15".

In fheroes2:

Cost summon spell 001

Cost summon spell 002

Cost summon spell 003

### Save file

Here is the save file:
[Cost summon elemental 01.zip](https://github.com/ihhub/fheroes2/files/10970672/Cost.summon.elemental.01.zip)

### Additional info

_No response_

autoexecbatman:
fix #6838 
Now the '_hero' is passed in the 'draw()' 
* Have not checked if it has any impact on the overall code.

### DIFF
--- a/src/fheroes2/gui/ui_dialog.cpp
+++ b/src/fheroes2/gui/ui_dialog.cpp
@@ -432,7 +432,7 @@ namespace fheroes2
     {
         assert( spell.isValid() );
 
-        const Text spellNameText( std::string( _spell.GetName() ) + " [" + std::to_string( _spell.spellPoints( _hero ) ) + ']', FontType::smallWhite() );
+        const Text spellNameText( std::string( _spell.GetName() ) + " [" + std::to_string( _spell.spellPoints( nullptr ) ) + ']', FontType::smallWhite() );
 
         const Sprite & icn = AGG::GetICN( ICN::SPELLS, _spell.IndexSprite() );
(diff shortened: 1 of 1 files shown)


thestahan (commented):
I think you should also pass the hero in line 435, which is used to calculate the width of the area.

autoexecbatman:
@thestahan honestly I would only add one change at a time . What you are referring to makes sense, but I am also not sure what this function does in the ctor and it seems that `nullptr` is passed explicitly to that function in many cases.
I can't see any visible effect currently either if I pass a nullptr or the _hero. 
might need to test with a 2 digit value becoming a 1 digit value

autoexecbatman:
@thestahan I've tested that change and it changes the spell points in the mage guild to get calculated with reduction . 
I apologize if I was too hasty about posting the second commit. I prefer to not add it .

thestahan:
> Any advice on how to delete this change ? I don't want to make a mess in git .

Yes, you can just revert the commit - it'll create a new commit that reverts the changes.

ihhub (approved):

Branikolog (approved):

ihhub:
@autoexecbatman , thank you so much for your contribution to the project!

PR #6846, opened 2023-03-16
Clang Analyzer & Clang-Tidy: migrate to Clang 15

oleg-derevenetz:


ihhub:
Hi @oleg-derevenetz , could you please tell extra benefits of the upgrade? Every version of clang-format slightly modifies code style so it is good to do this if there is a benefit🙂

oleg-derevenetz:
Hi @ihhub `clang-format` is not affected by this change (yet), because code style checking workflow should be as fast as possible and I prefer to use only the pre-installed tools (LLVM 15 with its tools [is not pre-installed](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#language-and-runtime)).

But in general all LLVM 15 tools got a new features, for example, `clang-format` now has the `InsertBraces` directive that forces braces after control statements, which may be useful (I'm not saying that it should be turned on immediately :).

Clang-Tidy got [new checks](https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/ReleaseNotes.html#improvements-to-clang-tidy), especially you may like [this one](https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc/const-correctness.html).

Clang Analyzer also [got](https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#static-analyzer) a few optimizations and new checkers.

PR #6854, opened 2023-03-17
Adjusted number of resources subtracted in event messages.

### LINKED ISSUE #6795: Adventure map, the number of resources subtracted in the event messages should be adjusted
### Preliminary checks

### Platform

Windows

### Describe the bug

In the original game, when an amount of resources is taken away from the player in an event, global or local, the amount of resources taken away from the player is adjusted to the player's actual treasury in the message displayed.
This is not the case in fheroes2.
In fheroes2, it is always the quantity indicated in the scenario editor which is displayed, without taking into account the treasury of the player.
This should be corrected.

In original game:

Gold event true value 001

Gold event true value 003
Gold event true value 004

Gold event true value 006
Gold event true value 007

Gold event true value 008
Gold event true value 009

In fheroes2:

Gold event true value 002

In original game:

Event true value 001

Event true value 002

In fheroes2:

Event true value 003

### Save file

Here is the test scenario file (for map event):
[0 9925f.zip](https://github.com/ihhub/fheroes2/files/10932461/0.9925f.zip)

### Additional info

_No response_

slawomir-witkowski:
fix #6795

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -699,8 +699,6 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     conf.SetCurrentColor( player->GetColor() );
 
-                    world.ClearFog( player->GetColor() );
-
                     kingdom.ActionBeforeTurn();
 
                     iconsPanel.ShowIcons( ICON_ANY );
@@ -746,8 +744,9 @@ fheroes2::GameMode Interface::Basic::StartGame()
                         Maps::Tiles::updateFogDirectionsInArea( { 0, 0 }, { world.w(), world.h() }, hotSeatAIFogColors( player ) );
                     }
 
-                    if ( !loadedFromSave )
+                    if ( !loadedFromSave ) {
                         kingdom.ActionNewDayResourceUpdate( nullptr );
+                    }
 
                     kingdom.ActionBeforeTurn();
 
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -27,7 +27,6 @@
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
-#include <list>
 #include <ostream>
 #include <string>
 #include <type_traits>
@@ -369,22 +368,6 @@ void ShowNewWeekDialog()
     fheroes2::showStandardTextMessage( "", message, Dialog::OK );
 }
 
-void ShowEventDayDialog()
-{
-    const Kingdom & myKingdom = world.GetKingdom( Settings::Get().CurrentColor() );
-    const EventsDate events = world.GetEventsDate( myKingdom.GetColor() );
-
-    for ( const EventDate & event : events ) {
-        if ( event.resource.GetValidItemsCount() ) {
-            fheroes2::showResourceMessage( fheroes2::Text( event.title, fheroes2::FontType::normalYellow() ),
-                                           fheroes2::Text( event.message, fheroes2::FontType::normalWhite() ), Dialog::OK, event.resource );
-        }
-        else if ( !event.message.empty() ) {
-            fheroes2::showStandardTextMessage( event.title, event.message, Dialog::OK );
-        }
-    }
-}
-
 void ShowWarningLostTownsDialog()
 {
     const Kingdom & myKingdom = world.GetKingdom( Settings::Get().CurrentColor() );
@@ -716,6 +699,8 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     conf.SetCurrentColor( player->GetColor() );
 
+                    world.ClearFog( player->GetColor() );
+
                     kingdom.ActionBeforeTurn();
 
                     iconsPanel.ShowIcons( ICON_ANY );
@@ -760,6 +745,9 @@ fheroes2::GameMode Interface::Basic::StartGame()
                     if ( isHotSeatGame ) {
                         Maps::Tiles::updateFogDirectionsInArea( { 0, 0 }, { world.w(), world.h() }, hotSeatAIFogColors( player ) );
                     }
+
+                    if ( !loadedFromSave )
+                        kingdom.ActionNewDayResourceUpdate( nullptr );
 
                     kingdom.ActionBeforeTurn();
 
@@ -848,7 +836,15 @@ fheroes2::GameMode Interface::Basic::HumanTurn( bool isload )
             ShowNewWeekDialog();
         }
 
-        ShowEventDayDialog();
+        myKingdom.ActionNewDayResourceUpdate( []( const EventDate & event, const Funds & funds ) {
+            if ( funds.GetValidItemsCount() ) {
+                fheroes2::showResourceMessage( fheroes2::Text( event.title, fheroes2::FontType::normalYellow() ),
+                                               fheroes2::Text( event.message, fheroes2::FontType::normalWhite() ), Dialog::OK, funds );
+            }
+            else if ( !event.message.empty() ) {
+                fheroes2::showStandardTextMessage( event.title, event.message, Dialog::OK );
+            }
+        } );
 
         if ( conf.isAutoSaveAtBeginningOfTurnEnabled() ) {
             Game::AutoSave();
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -2667,7 +2667,7 @@ namespace
         if ( event_maps && event_maps->isAllow( hero.GetColor() ) ) {
             hero.SetMove( false );
 
-            Funds fundsToUpdate = hero.GetKingdom().GetFunds().CalculateEventResourceUpdate( event_maps->resources );
+            const Funds fundsToUpdate = hero.GetKingdom().GetFunds().CalculateEventResourceUpdate( event_maps->resources );
 
             std::vector<fheroes2::ResourceDialogElement> resourceUI = fheroes2::getResourceDialogElements( fundsToUpdate );
             std::unique_ptr<fheroes2::ArtifactDialogElement> artifactUI;
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -2667,7 +2667,9 @@ namespace
         if ( event_maps && event_maps->isAllow( hero.GetColor() ) ) {
             hero.SetMove( false );
 
-            std::vector<fheroes2::ResourceDialogElement> resourceUI = fheroes2::getResourceDialogElements( event_maps->resources );
+            Funds fundsToUpdate = hero.GetKingdom().GetFunds().CalculateEventResourceUpdate( event_maps->resources );
+
+            std::vector<fheroes2::ResourceDialogElement> resourceUI = fheroes2::getResourceDialogElements( fundsToUpdate );
             std::unique_ptr<fheroes2::ArtifactDialogElement> artifactUI;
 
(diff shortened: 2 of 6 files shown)


src/fheroes2/game/game_startgame.cpp
```diff
@@ -333,8 +333,8 @@ void ShowNewWeekDialog()
 
 void ShowEventDayDialog()
 {
-    const Kingdom & myKingdom = world.GetKingdom( Settings::Get().CurrentColor() );
-    const EventsDate events = world.GetEventsDate( myKingdom.GetColor() );
+    Kingdom & myKingdom = world.GetKingdom( Settings::Get().CurrentColor() );
+    const EventsDate & events = myKingdom.GetAppliedEvents();
 
     for ( const EventDate & event : events ) {
         if ( event.resource.GetValidItemsCount() ) {
```
ihhub, line 340: I suggest to do a simplified version of this fix:
```cpp
void ShowEventDayDialog()
{
    const Kingdom & myKingdom = world.GetKingdom( Settings::Get().CurrentColor() );
    const EventsDate events = world.GetEventsDate( myKingdom.GetColor() );

    Funds funds = myKingdom.GetFunds();

    for ( const EventDate & event : events ) {
        Funds fundsToUpdate;
        if ( event.resource.GetValidItemsCount() ) {
            const Funds fundsBefore = funds;
            funds = funds + event.resource;
            funds.Trim();
            fundsToUpdate = funds - fundsBefore;
        }

        if ( fundsToUpdate ) {
            fheroes2::showResourceMessage( fheroes2::Text( event.title, fheroes2::FontType::normalYellow() ),
                                           fheroes2::Text( event.message, fheroes2::FontType::normalWhite() ), Dialog::OK, fundsToUpdate );
        }
        else if ( !event.message.empty() ) {
            fheroes2::showStandardTextMessage( event.title, event.message, Dialog::OK );
        }
    }
}
```
This should be enough to fix the initial problem. What do you think?

ihhub (changes requested):
Hi @slawomir-witkowski , I proposed an alternative approach to fix the problem. Could you please take a look at it and see if it is a valid one?

src/fheroes2/game/game_startgame.cpp
slawomir-witkowski: Hi @ihhub, this doesn't work. The problem is that at this point myKingdom.GetFunds(); returns funds that have already been reduced by a call to AddFundsResource(event.resource) in kingdom.cpp:262.  
We need a way to store information about how much each event subtracted for each kingdom in the current day.

Branikolog:
Hello, @slawomir-witkowski 
After scenario restart I still got messages, which I got on previous game:
image
image
The second one is which I've established in map editor.
But the first (-1000) was showed on previous game, when a player didn't have resources anymore. Now It appears before the main message.

slawomir-witkowski:
Initially I focused on daily events. Now I added support for Map Events as well. @Branikolog Castle generates 1000gold/day, so a daily event can display -1000 every day. If you still see problems, can you attach your map to the issue?

src/fheroes2/game/game_startgame.cpp
ihhub: Ah, you are right. Actually we have a much bigger problem here:
**World::NewDay()** is called for the whole day which updates all statuses for players, especially updating resources. Let's imagine this situation:
- we play in hot-seat mode 1 vs 1
- my turn first, yours second
- you completed the turn and you know that you have 5000 gold in your kingdom
- a new day starts deducting 10,000 gold from each player
- I attack your hero and you need to surrender to save Dragon but you couldn't as your gold now is zero

In this case the logic is wrong as you haven't started your turn so you must have that 5,000 gold.

Hi @LeHerosInconnu, could you please verify this scenario?

LeHerosInconnu: It works correctly in the original game.

slawomir-witkowski: Hi guys, IMO this problem deserves a separate issue. We have to consider all resource updates and do them in the right moment.

Consider an opposite situation when players get +10k bonus. Player first to act can use the gold to buy an army and kill another player before he has a chance to use his bonus.

ihhub: Hi @slawomir-witkowski , as confirmed the logic for new day is not correct so your changes and the logic are actually correlate to each other. I foresee that even after your fix other would find that the number of resources in the bottom-right corner is not updated (@LeHerosInconnu missed that for now). Your proposed solution is indeed fixing the problem but since the original code is wrong the solution looks more like a temporary patch (any other solution would look the same to be honest).

If we move events out of NewDay() method and execute them once with and without windows then we could solve 3 existing problems at once. Right now even if we go with the proposed solution we would need to redesign it due to other 2 highlighted issues. What do you think?

slawomir-witkowski: Ok, I added a change that will delay updating resources until your turn. Now you can save the dragon. However, I don't see the problem with resources in bottom-right corner not being updated.

Branikolog:
Hi, @slawomir-witkowski !
I've tested on daily events exactly. The problem is that there appear several messages from previously played games (which showed the substracted value on previous game). So, starting new game for several times comes up with several messages.
Your last fix solved the problem with restart.
But after we load the game on the next turn we'd see two messages for the player loaded the game. The same happens if we load the game, skip 1 turn and then restart the scenario.

https://user-images.githubusercontent.com/55348946/226594880-cc3ee58b-e782-4cdd-9899-612063abe245.mp4

slawomir-witkowski:
@Branikolog Thank you for this video. It allowed me to find and fix the problem.

src/fheroes2/resource/resource.h
```diff
@@ -96,6 +96,7 @@ class Funds
     int GetValidItems() const;
     uint32_t GetValidItemsCount() const;
 
+    Funds GetUpdate( const Funds & ) const;
```
ihhub, line 99: It is not clear what exactly this method does at first place. I suggest to just remove it or rename in a way what it reflects what it actually does. Also please add the name of the input argument.

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -236,7 +236,10 @@ void Kingdom::ActionNewDay()
 
     // Reset the effect of the "Identify Hero" spell
     ResetModes( IDENTIFYHERO );
+}
 
+const EventsDate Kingdom::ActionNewDayResourceUpdate()
```
ihhub, line 241: In order to avoid unnecessary double event loops and also not to mix UI and logic this method can accept an `std::function`. For AI it is going to be nullptr while for humans it is going to be passed as a lambda function:
```cpp
[]( const EventDate & event, const Funds & funds ) {
    if ( funds ) {
        fheroes2::showResourceMessage( fheroes2::Text( event.title, fheroes2::FontType::normalYellow() ), fheroes2::Text( event.message, fheroes2::FontType::normalWhite() ), Dialog::OK, funds );
    }
    else if ( !event.message.empty() ) {
        fheroes2::showStandardTextMessage( event.title, event.message, Dialog::OK );
    }
};
```
so then we don't need to have extra functions to run events for the second time. After this we:
- remove **ShowEventDayDialog()** function
- within **Interface::Basic::HumanTurn()** method
```cpp
...
    if ( !isload ) {
        if ( 1 < world.CountWeek() && world.BeginWeek() ) {
            ShowNewWeekDialog();
        }

       auto lambdaFunction = ...;
        ActionNewDayResourceUpdate( lambdaFunction ); // or directly define here.
        ...
```
- inside **Interface::Basic::StartGame()** method after line `world.ClearFog( player->GetColor() );` for AI we call **ActionNewDayResourceUpdate( nullptr )**.

At the end the code becomes smaller and we still don't mix logic and UI. What do you think?

ihhub (changes requested):
Hi @slawomir-witkowski . I put few more suggestions in this pull request. Would you mind to take a look once you have spare time?

src/fheroes2/kingdom/kingdom.cpp
slawomir-witkowski: Hi, I made the suggested changes.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -716,6 +699,8 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     conf.SetCurrentColor( player->GetColor() );
 
+                    world.ClearFog( player->GetColor() );
```
ihhub, line 702: Could you please explain why we need to clear fog here?

src/fheroes2/game/game_startgame.cpp
```diff
@@ -761,6 +746,9 @@ fheroes2::GameMode Interface::Basic::StartGame()
                         Maps::Tiles::updateFogDirectionsInArea( { 0, 0 }, { world.w(), world.h() }, hotSeatAIFogColors( player ) );
                     }
 
+                    if ( !loadedFromSave )
```
ihhub, line 749: Please add parenthesis for this if-body. It would be easier to read.

src/fheroes2/resource/resource.h
```diff
@@ -96,6 +96,7 @@ class Funds
     int GetValidItems() const;
     uint32_t GetValidItemsCount() const;
 
+    Funds CalculateEventResourceUpdate( const Funds & eventFunds ) const;
```
ihhub, line 99: I think this method shouldn't belong to this class as it's for a very specific case. Usually developers just create a separate helper function outside the class to address specific cases. For instance you can take on a mess which still exist in Maps::Tiles class where many non-generic methods were added in the past.

What do you think, @oleg-derevenetz ?

src/fheroes2/kingdom/kingdom.cpp
```diff
@@ -256,9 +259,12 @@ void Kingdom::ActionNewDay()
     }
 
     // Resources from events
-    const EventsDate events = world.GetEventsDate( GetColor() );
+    EventsDate events = world.GetEventsDate( GetColor() );
     for ( const EventDate & event : events ) {
-        AddFundsResource( event.resource );
+        Funds fundsUpdate = GetFunds().CalculateEventResourceUpdate( event.resource );
```
ihhub, line 264: Please make this variable const.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2667,7 +2667,9 @@ namespace
         if ( event_maps && event_maps->isAllow( hero.GetColor() ) ) {
             hero.SetMove( false );
 
-            std::vector<fheroes2::ResourceDialogElement> resourceUI = fheroes2::getResourceDialogElements( event_maps->resources );
+            Funds fundsToUpdate = hero.GetKingdom().GetFunds().CalculateEventResourceUpdate( event_maps->resources );
```
ihhub, line 2670: Please make this variable const.

ihhub (changes requested):
Hi @slawomir-witkowski , I added few more comments and questions here. We've almost completed it. Would you mind to check them?

src/fheroes2/game/game_startgame.cpp
slawomir-witkowski: This shouldn't be here. My mistake.

src/fheroes2/resource/resource.h
oleg-derevenetz: @ihhub I'm fine with the current solution where this auxiliary function is moved to the namespace.

ihhub (approved):

Branikolog (approved):

ihhub:
@slawomir-witkowski , huge thanks for these changes!

PR #6855, opened 2023-03-17
Fix quick info for non-action objects when their action tile is under the fog. Adventure map.

Districh-ru:
In fheroes2 when you try to get quick info for an object whose action (main) tile is under the fog you get the "Uncharted Territory" instead of the object info:
fheroes2 2023-03-17 20-41-09-980

In original game you are able to get object quick info if its action (main) tile is under the fog.

This PR makes the fog check to be done only for the tile under the cursor and if it is the non-action tile then get info from the action tile of this object:
fheroes2 2023-03-17 20-09-03-104

This PR also slightly reworks the code for all quick info text strings.

### DIFF
--- a/src/fheroes2/dialog/dialog.h
+++ b/src/fheroes2/dialog/dialog.h
@@ -85,7 +85,7 @@ namespace Dialog
     std::string SelectFileLoad();
     std::string SelectFileSave();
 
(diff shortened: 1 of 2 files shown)


ihhub (approved):

ihhub:

PR #6896, opened 2023-03-25
Fix a few more code smells

oleg-derevenetz:
Follow-up to #6894.


ihhub (approved):

ihhub:

PR #6899, opened 2023-03-26
Focus on the AI hero before he disappears due to boarding a boat or using a Dimension Door or Town Portal spell

### LINKED ISSUE #6897: AI hero with a strong army dissapears (Town gate/portal?)
### Preliminary checks

### Platform

Windows

### Describe the bug

Orange AI Myra hero (visible close to the center, near a Xanadu) with a strong army dissapears after the next AI turn, the saved game is attached.  It's hard to beleive that AI can disband such a strong army. If town gate / portal spell is applied, may be it is better to implement "fade out" animation.

### Save file

[MARAN_0073.zip](https://github.com/ihhub/fheroes2/files/11070523/MARAN_0073.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #6897

Unlike using Stone Liths or Whirlpool, AI hero can cast Dimension Door or Town Portal (or get on board the boat) right at the beginning of his turn without any previous movement, so we need to focus on the corresponding tile first to display the fade-out animation.

### DIFF
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -625,7 +625,7 @@ namespace
         hero.SetShipMaster( false );
         AudioManager::PlaySound( M82::KILLFADE );
         hero.GetPath().Hide();
-        hero.FadeIn( fheroes2::Point( offset.x * Game::HumanHeroAnimSkip(), offset.y * Game::HumanHeroAnimSkip() ) );
+        hero.FadeIn( offset );
 
         // Clear hero position marker from the boat and scout the area on radar after disembarking.
         Interface::Basic & I = Interface::Basic::Get();
--- a/src/fheroes2/maps/mp2.cpp
+++ b/src/fheroes2/maps/mp2.cpp
@@ -531,11 +531,6 @@ bool MP2::isWeekLife( const MapObjectType objectType )
     case OBJ_DRAGON_CITY:
     case OBJ_CITY_OF_DEAD:
     case OBJ_TROLL_BRIDGE:
-    // TODO: former AI-related logic, OBJ_COAST should never be set
-    // TODO: as a visited object, remove this temporary check later
-    case OBJ_COAST:
-        assert( 0 );
-        return true;
     // for AI
     case OBJ_HEROES:
         return true;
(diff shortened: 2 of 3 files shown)


src/fheroes2/maps/mp2.cpp
```diff
@@ -532,7 +532,6 @@ bool MP2::isWeekLife( const MapObjectType objectType )
     case OBJ_CITY_OF_DEAD:
     case OBJ_TROLL_BRIDGE:
     // for AI
-    case OBJ_COAST:
     case OBJ_HEROES:
```
ihhub, line 535: I see no logic in the code why Heroes belong to week-life objects. Should we remove this case?

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -625,7 +625,7 @@ namespace
         hero.SetShipMaster( false );
         AudioManager::PlaySound( M82::KILLFADE );
         hero.GetPath().Hide();
-        hero.FadeIn( fheroes2::Point( offset.x * Game::HumanHeroAnimSkip(), offset.y * Game::HumanHeroAnimSkip() ) );
```
ihhub, line 628: Special fading animation depending on animation speed from and to boats was added on purpose. I don't remember exactly why we did this. Perhaps @Branikolog shed a light about this? I just want to understand what was the logic behind and why we are changing it right now.

ihhub (changes requested):
Hi @oleg-derevenetz , I left few comments here. Could you please check them?

src/fheroes2/heroes/heroes_action.cpp
oleg-derevenetz: @ihhub This behavior was added in #1899, BUT then it was removed in #4164, and only for boarding, but not for disembarkation(???). Now the boarding and disembarkation is performed with the different speed, which looks wrong. What's more, this is the only fading animation that depends on animation speed, which looks suspicious. Other fading animations (hero elimination, Dimension Door, Stone Liths, etc, etc) does not depend on it. Looks like this is just a fossil :)

src/fheroes2/maps/mp2.cpp
oleg-derevenetz: I'm not sure about that because this logic is somehow used by AI and out of scope of this PR. This needs to be investigated separately. The `OBJ_COAST` case looks clear, because I implemented the logic for paving path through the water myself :)

ihhub (approved):

ihhub:

src/fheroes2/maps/mp2.cpp
oleg-derevenetz: @ihhub This logic is used in the `Heroes::markHeroMeeting()`, `Heroes::unmarkHeroMeeting()` and `Heroes::hasMetWithHero()`, which is used by the AI code to decide whether one AI hero should meet another AI hero to get reinforcements. I will not touch this.

PR #6902, opened 2023-03-26
Fix clicks in Android MIDI music, use 44100 Hz sample frequency for all systems.

Districh-ru:
Relates to #6429 

This PR:
1) Fixes MIDI music clicks on Android which uses Timidity backend by replacing some GUS patches with fixed ones, downloaded from https://www.doomworld.com/idgames/sounds/eawpats

2) Makes the engine to use 44100 Hz sample rate as music files (*ogg*, *mp3*) and midi samples (in *pat* and *sf3* files) has sample rate up to 48000 Hz which is more than 22050 Hz, currently used in the engine.
IMHO it is not good to reduce the music quality, taking into account that sound cards mostly has 48000 Hz or higher sampling rate.
To preserve quality of the game audio effects which has 22050 Hz sample rate we should set the sample rate that is a multiple of 22050, so I choose 22050 * 2 = 44100 Hz. (Setting 48000 Hz results in multiple sound distortions under Windows).

3) Reduces the game volume by 0.5 dB to leave a room for possible sound overload in SDL_mixer or in OS mixer and/or resampler.
This is not noticeable volume change. It is done to overprotect from possible overloads.
When sound is resampled it is possible to get time samples more than allowed amplitude. On example picture if we will resample the sound for 2x sample rate we will get a value slightly over 0 dB:
изображение

The comparison of changes, made by this PR:

MIDI music recorded from Android device before this PR (the vertical lines in spectrum - are clicks):

https://user-images.githubusercontent.com/113276641/227791047-997d0f30-d39f-4e35-b6ee-79faa9a0893d.mp4

MIDI music recorded from Android device after this PR:

https://user-images.githubusercontent.com/113276641/227795245-db941236-4828-43d8-93e5-d774512f3dad.mp4

On video it can be seen (and heard) that music has frequencies above 11 kHz.


oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for these useful changes!

PR #6911, opened 2023-03-29
Create small and large obstacles on the battlefield at the same time.

### LINKED ISSUE #4715: Combat screen, it should be possible to have small and large obstacles on the battlefield at the same time
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, it should be possible to have small and large obstacles on the battlefield at the same time.
This happens in the original game.
I don't remember seeing this in fheroes2.

In original game:

Both large and small obstacles H2 01

Both large and small obstacles H2 02

Both large and small obstacles H2 03

### Save file

[0 9947.zip](https://github.com/ihhub/fheroes2/files/7611995/0.9947.zip)

### Additional info

Related closed issue: (https://github.com/ihhub/fheroes2/issues/1613).

slawomir-witkowski:
fix #4715

### DIFF
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -28,7 +28,6 @@
 #include <cassert>
 #include <cstdint>
 #include <cstdlib>
-#include <iterator>
 #include <memory>
 #include <ostream>
 #include <set>
@@ -595,20 +594,21 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
 
     // Limit the number of two-hex obstacles to maxTwoHexObstacleCount
     uint8_t twoHexCount = 0;
-    for ( size_t i = 0; i < objs.size(); ) {
-        if ( isTwoHexObject( objs[i] ) ) {
+    for ( auto iter = objs.begin(); iter != objs.end(); ) {
+        if ( isTwoHexObject( *iter ) ) {
             ++twoHexCount;
             if ( twoHexCount > maxTwoHexObstacleCount ) {
-                objs.erase( std::next( objs.begin(), i ) );
+                iter = objs.erase( iter );
                 continue;
             }
         }
-        ++i;
+        ++iter;
     }
 
     const size_t objectsToPlace = std::min( objs.size(), static_cast<size_t>( Rand::GetWithGen( 0, maxSmallObstacleCount, gen ) ) );
 
     for ( size_t i = 0; i < objectsToPlace; ++i ) {
+        // two-hex obstacles are not allowed on column 8 as they would cover column 9 which is reserved for units
         const bool checkRightCell = isTwoHexObject( objs[i] );
 
         // tall obstacles like trees should not be placed on top 2 rows
@@ -617,7 +617,8 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
         uint32_t dest;
         do {
             dest = GetRandomObstaclePosition( gen );
-        } while ( at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || ( dest % 11 ) == 8 ) ) || ( isTallObstacle && dest <= 21 ) );
+        } while ( at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || ( dest % ARENAW ) == 8 ) )
+                  || ( isTallObstacle && dest < ( ARENAW * 2 ) ) );
 
         SetCobjObject( objs[i], dest );
     }
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -597,13 +597,13 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
     uint8_t twoHexCount = 0;
     for ( size_t i = 0; i < objs.size(); ) {
         if ( isTwoHexObject( objs[i] ) ) {
-            twoHexCount++;
+            ++twoHexCount;
             if ( twoHexCount > maxTwoHexObstacleCount ) {
                 objs.erase( std::next( objs.begin(), i ) );
                 continue;
             }
         }
-        i++;
+        ++i;
     }
 
     const size_t objectsToPlace = std::min( objs.size(), static_cast<size_t>( Rand::GetWithGen( 0, maxSmallObstacleCount, gen ) ) );
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -573,8 +573,8 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
 
     const auto largeObstacleHexCount = std::count_if( begin(), end(), []( const Cell & cell ) { return cell.GetObject() == 0x40; } );
 
-    uint8_t maxSmallObstacleCount;
-    uint8_t maxTwoHexObstacleCount;
+    uint8_t maxSmallObstacleCount = 0;
+    uint8_t maxTwoHexObstacleCount = 0;
 
     if ( largeObstacleHexCount == 0 ) {
         maxSmallObstacleCount = 6;
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -21,11 +21,14 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "battle_board.h"
+
 #include <algorithm>
 #include <array>
 #include <cassert>
 #include <cstdint>
 #include <cstdlib>
+#include <iterator>
 #include <memory>
 #include <ostream>
 #include <set>
@@ -34,7 +37,6 @@
 #include "battle.h"
 #include "battle_arena.h"
 #include "battle_army.h"
-#include "battle_board.h"
 #include "battle_bridge.h"
 #include "battle_troop.h"
 #include "castle.h"
@@ -552,6 +554,21 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
 
     const auto largeObstacleSize = std::count_if( begin(), end(), []( const Cell & cell ) { return cell.GetObject() == 0x40; } );
     const uint8_t maxSmallObstacleCount = largeObstacleSize > 7 ? 3 : 4;
+
+    // Ensure obstacles vector will only have maximum 2 TwoHexObjects
+    if ( largeObstacleSize > 0 ) {
+        uint8_t twoHexCount = 0;
+        for ( size_t i = 0; i < objs.size(); ) {
+            if ( isTwoHexObject( objs[i] ) ) {
+                twoHexCount++;
+                if ( twoHexCount > 2 ) {
+                    objs.erase( std::next( objs.begin(), i ) );
+                    continue;
+                }
+            }
+            i++;
+        }
+    }
 
     const size_t objectsToPlace = std::min( objs.size(), static_cast<size_t>( Rand::GetWithGen( 0, maxSmallObstacleCount, gen ) ) );
 
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -550,7 +550,10 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
 
     Rand::ShuffleWithGen( objs, gen );
 
-    const size_t objectsToPlace = std::min( objs.size(), static_cast<size_t>( Rand::GetWithGen( 0, 4, gen ) ) );
+    const auto largeObstacleSize = std::count_if( begin(), end(), []( const Cell & cell ) { return cell.GetObject() == 0x40; } );
+    const uint8_t maxSmallObstacleCount = largeObstacleSize > 7 ? 3 : 4;
+
+    const size_t objectsToPlace = std::min( objs.size(), static_cast<size_t>( Rand::GetWithGen( 0, maxSmallObstacleCount, gen ) ) );
 
     for ( size_t i = 0; i < objectsToPlace; ++i ) {
         const bool checkRightCell = isTwoHexObject( objs[i] );
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -52,9 +52,9 @@
 
 namespace
 {
-    int GetRandomObstaclePosition( std::mt19937 & gen )
+    uint32_t GetRandomObstaclePosition( std::mt19937 & gen )
     {
-        return Rand::GetWithGen( 3, 7, gen ) + ( 11 * Rand::GetWithGen( 1, 7, gen ) );
+        return Rand::GetWithGen( 3, 7, gen ) + ( 11 * Rand::GetWithGen( 0, 8, gen ) );
     }
 
     bool isTwoHexObject( const int icnId )
@@ -594,20 +594,16 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
     for ( size_t i = 0; i < objectsToPlace; ++i ) {
         const bool checkRightCell = isTwoHexObject( objs[i] );
 
-        int32_t dest;
-        Battle::Indexes around;
+        uint32_t dest;
         do {
             dest = GetRandomObstaclePosition( gen );
-            around = Board::GetAroundIndexes( dest );
-        } while (
-            at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || dest % 11 == 7 ) )
-            || ( std::any_of( around.begin(), around.end(), [this]( int index ) { return ( at( index ).GetObject() != 0 && at( index ).GetObject() != 0x40 ); } ) ) );
+        } while ( at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || dest % 11 == 7 ) ) );
 
         SetCobjObject( objs[i], dest );
     }
 }
 
-void Battle::Board::SetCobjObject( const int icn, const int32_t dst )
+void Battle::Board::SetCobjObject( const int icn, const uint32_t dst )
 {
     at( dst ).SetObject( 0x80 + ( icn - ICN::COBJ0000 ) );
 
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -594,10 +594,14 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
     for ( size_t i = 0; i < objectsToPlace; ++i ) {
         const bool checkRightCell = isTwoHexObject( objs[i] );
 
-        int32_t dest = GetRandomObstaclePosition( gen );
-        while ( at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || dest % 11 == 7 ) ) ) {
+        int32_t dest;
+        Battle::Indexes around;
+        do {
             dest = GetRandomObstaclePosition( gen );
-        }
+            around = Board::GetAroundIndexes( dest );
+        } while (
+            at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || dest % 11 == 7 ) )
+            || ( std::any_of( around.begin(), around.end(), [this]( int index ) { return ( at( index ).GetObject() != 0 && at( index ).GetObject() != 0x40 ); } ) ) );
 
         SetCobjObject( objs[i], dest );
     }
--- a/src/fheroes2/battle/battle_board.cpp
+++ b/src/fheroes2/battle/battle_board.cpp
@@ -606,9 +606,7 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
         i++;
     }
 
-    // TODO for testing only
-    //  const size_t objectsToPlace = std::min( objs.size(), static_cast<size_t>( Rand::GetWithGen( 0, maxSmallObstacleCount, gen ) ) );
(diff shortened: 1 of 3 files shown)


ihhub (approved):
Hi @slawomir-witkowski , code wise the change looks good but we might need to check how it behaves in the original game in terms of statistics. I think if a big obstacle is present then the small ones appear less frequent.

Hi @LeHerosInconnu could you help us with this?

LeHerosInconnu:
Hello @slawomir-witkowski and @ihhub,

> Hi @slawomir-witkowski , code wise the change looks good but we might need to check how it behaves in the original game in terms of statistics. I think if a big obstacle is present then the small ones appear less frequent.
> 
> Hi @LeHerosInconnu could you help us with this?

So I did this boring test :) and this is what I observed:
- There are no large obstacles for the Swamp, Lava, Desert and Beach terrains.
- There does not seem to be a specific rule about the number of small obstacles when large obstacles are present on the battlefield.

In original game, some examples:
Grass, Dirt, Snow and Wasteland 003
Grass, Dirt, Snow and Wasteland 015
Grass, Dirt, Snow and Wasteland 025

Here are screenshots of all test cases (twenty combats on each type of terrain) containing at least one large obstacle on the battlefield:
[Combat obstacles 01.zip](https://github.com/ihhub/fheroes2/files/11133073/Combat.obstacles.01.zip)

Note: In the original game, the number of obstacles and their positions will always be the same even if the scenario is restarted from the beginning.
[0 9923b.zip](https://github.com/ihhub/fheroes2/files/11133075/0.9923b.zip)

Branikolog:
Hi, @ihhub !
I've made several dozens of tests and obstacle generation seem to be sometimes too excessive.
I've caught a few cases, where up to 6 objects were generated on a battlefield:
6 obj2
6 obj1
@LeHerosInconnu , do you know if in the original game so many obstacles could be placed on a battlefield? I even couldn't manage to reproduce 5 obstacles as you did on the screenshots above.
I also feel, that small obstacles appear quite rarely for the cases with 1-2 large objects on a battlefield in the original game. While I met the generation with 2 large and 2-3 small obstacles quite often. Anyway, I couldn't even closely reproduce the same generation from my screenshots in the original game.

One more observation I have, but I'm not sure it's related to the current issue.
In the original game some objects could be generated on the 3rd column of hexes, right in front of 2hex creature, while in fheroes2 objects seem to be generated with a gap of 1 hex between object and 2-hex creature at least (on the 4th column).
OG:
close obst

LeHerosInconnu:
Hello @Branikolog,

> @LeHerosInconnu , do you know if in the original game so many obstacles could be placed on a battlefield? I even couldn't manage to reproduce 5 obstacles as you did on the screenshots above.

I can't be sure that this doesn't happen in the original game, but if it does it must be pretty rare.

> I also feel, that small obstacles appear quite rarely for the cases with 1-2 large objects on a battlefield in the original game. While I met the generation with 2 large and 2-3 small obstacles quite often. Anyway, I couldn't even closely reproduce the same generation from my screenshots in the original game.

Maybe we can implement an interim rule until we get more information on the subject.
Something like this:
- 1 large obstacle, up to 4 small obstacles (one and two hexes, maximum 2 two-hex obstacles)
- 2 large obstacles, up to 3 small obstacles (one and two hexes, maximum 2 two-hex obstacles)

> One more observation I have, but I'm not sure it's related to the current issue. In the original game some objects could be generated on the 3rd column of hexes, right in front of 2hex creature, while in fheroes2 objects seem to be generated with a gap of 1 hex between object and 2-hex creature at least (on the 4th column).

Yes, it is that way in the original game, and it can be a disadvantage as well as an advantage. :)

slawomir-witkowski:
Hi Guys,
1. Large obstacles have varying number of hexes. How many hexes do we need to consider it large?
2. We need a decision if we should place obstacles right in front of 2hex creature?

LeHerosInconnu:
Hello @slawomir-witkowski,

> Hi Guys,
> 
> 1. Large obstacles have varying number of hexes. How many hexes do we need to consider it large?

Obstacles are considered large when they occupy at least 3 hexes on the battlefield.

Example on the Snow terrain, the 3 hex obstacle in the upper part of the battlefield:
Grass, Dirt, Snow and Wasteland 015

Example on the Wasteland terrain, the 3 hex obstacle in the upper part of the battlefield.
Grass, Dirt, Snow and Wasteland 025

> 2. We need a decision if we should place obstacles right in front of 2hex creature?

For my part I am okay with both ways, but my preference goes to the version of fheroes2 (3 hexes, with 1 free hex in front of two-hex creatures) which is less "frustrating" for the player and which allows not to have too many troops "blocked", and which on the other hand requires more troops to block an access if necessary (3 hexes to be blocked in fheroes2, against 2 hexes to be blocked in Heroes 2), which balances well the difference of functioning.

slawomir-witkowski:
My thinking is that the number of small obstacles we add should depend on the number of hexes blocked by large obstacles. 
For example: 
If large blocked 8+ hexes, add max 3 small.
If large blocked 12+ hexes, add max 1 small.

slawomir-witkowski:
I propose the following solution which I implemented:

1. If large obstacle size is over 7 hexes, up to 3 small obstacles are allowed.
2. If large obstacle size is 7 or less, up to 4 small obstacles are allowed.
3. In both cases above, only 2 small TwoHexObstacles are allowed.
4. If no large obstacle exists, up to 4 small obstacles are allowed even if all are 2hex.

Branikolog:
Hi, @LeHerosInconnu 

> > One more observation I have, but I'm not sure it's related to the current issue. In the original game some objects could be generated on the 3rd column of hexes, right in front of 2hex creature, while in fheroes2 objects seem to be generated with a gap of 1 hex between object and 2-hex creature at least (on the 4th column).
> 
> Yes, it is that way in the original game, and it can be a disadvantage as well as an advantage. :)

I find the original obstacles positioning more diverse because of this rule.  We're losing 18 hexes of  potential obstacles on the field. :)

Hi, @slawomir-witkowski 

> * If large obstacle size is over 7 hexes, up to 3 small obstacles are allowed.

What if there're two large (7+ hexes) obstacles?

> * If large obstacle size is 7 or less, up to 4 small obstacles are allowed.
> * In both cases above, only 2 small TwoHexObstacles are allowed.
> * If no large obstacle exists, up to 4 small obstacles are allowed even if all are 2hex.

How do you think, @LeHerosInconnu , can we rise small obstacles generation up to 5 for the last case? Can the original battlefield contain 5 1-2hex obstacles?

LeHerosInconnu:
Hello @slawomir-witkowski, @ihhub and @Branikolog,

> I propose the following solution which I implemented:
> 
> 1. If large obstacle size is over 7 hexes, up to 3 small obstacles are allowed.
> 2. If large obstacle size is 7 or less, up to 4 small obstacles are allowed.
> 3. In both cases above, only 2 small TwoHexObstacles are allowed.
> 4. If no large obstacle exists, up to 4 small obstacles are allowed even if all are 2hex.

> Hi, @LeHerosInconnu
> 
> > > One more observation I have, but I'm not sure it's related to the current issue. In the original game some objects could be generated on the 3rd column of hexes, right in front of 2hex creature, while in fheroes2 objects seem to be generated with a gap of 1 hex between object and 2-hex creature at least (on the 4th column).
> > 
> > 
> > Yes, it is that way in the original game, and it can be a disadvantage as well as an advantage. :)
> 
> I find the original obstacles positioning more diverse because of this rule. We're losing 18 hexes of potential obstacles on the field. :)

Even without those 18 hexes, there's plenty of room for lots of obstacles. :)
In Heroes 2, it can be very annoying when there are two troops of two-hex creatures in the hero's army and they end up with an obstacle right in front of their position.
And it is also very penalizing for the AI when this happens.

> Hi, @slawomir-witkowski
> 
> > * If large obstacle size is over 7 hexes, up to 3 small obstacles are allowed.
> 
> What if there're two large (7+ hexes) obstacles?
> 
> > * If large obstacle size is 7 or less, up to 4 small obstacles are allowed.
> > * In both cases above, only 2 small TwoHexObstacles are allowed.
> > * If no large obstacle exists, up to 4 small obstacles are allowed even if all are 2hex.
> 
> How do you think, @LeHerosInconnu , can we rise small obstacles generation up to 5 for the last case? Can the original battlefield contain 5 1-2hex obstacles?

I can't answer that with certainty.
I think that for no large obstacle, we can place up to 5 small obstacles with 3 two-hex obstacles maximum.

So what do you think about this:
- No large obstacle, up to 5 small obstacles (one and two hexes, maximum 3 two-hex obstacles)
- If the total number of hexes of the large obstacle(s) is less than or equal to 7 hexes, up to 4 small obstacles (one and two hexes, maximum 2 two-hex obstacles)
- If the total number of hexes of the large obstacle(s) is between 8 and 13 hexes, up to 3 small obstacles (one and two hexes, maximum 2 two-hex obstacles)
- If the total number of hexes of the large obstacle(s) is over 13 hexes, up to 2 small obstacles (one and two hexes, maximum 1 two-hex obstacle)
- In all cases, there can be no more than 2 large obstacles, regardless of their size.

slawomir-witkowski:
Hi @LeHerosInconnu,

Your idea sounds reasonable to me. How do you guys make decisions like this? These rules are somewhat arbitrary, if we don't have OG source. @ihhub Can you approve that these are the rules we want to implement?

LeHerosInconnu:
Hello @slawomir-witkowski,

> Hi @LeHerosInconnu,
> 
> Your idea sounds reasonable to me. How do you guys make decisions like this? These rules are somewhat arbitrary, if we don't have OG source.

Thousands of hours playing the game over decades, and also watching hundreds of let's play videos. :)

slawomir-witkowski:
Hi @LeHerosInconnu, 
I implemented rules you suggested. For testing purposes we always generate maximum number of small obstacles. So you can get a feel if battle area is not too clutered. I also added a change that allows to place single hex obstacles on column 8.
image

LeHerosInconnu:
Hello @Kazhuu,

> @LeHerosInconnu I now have implemented all the cases mentioned in this PR. Including the case when opponent only have a village that cannot be upgraded to a castle. In this situation you cannot cast identify hero spell and the text you wanted is shown.
> 
> Another thing. I still cannot load you save file for some reason. I tried both multiplayer -> hot seat and single player. In both it doesn't show up. I tested this on both Windows and Linux and still the same issue. Probably have to dig source code to see what is the issue. In issue #6952 your saves show up and are loaded fine.
> 
> Also I still cannot reproduce the issue you have. It would help if I could get the save file working first. Are you sure you checkout my PR and pulled the latest files? Could you try again now? :)

I have downloaded this file: "fheroes2_windows_x64_SDL2.zip" here: https://github.com/ihhub/fheroes2/suites/12125020088/artifacts/639202812.
Procedure I do for each PR test.

I did some new tests. :)

In fheroes2:

For the Haunt spell, it does not work properly.
The spell can still be cast:

https://user-images.githubusercontent.com/43583994/230914850-d26975f0-afd5-49b5-b739-bd80ec974dd0.mp4

The message for Set Guardian has been successfully modified:
Set Guardian modified text 01

For the Identify Hero spell, it doesn't work properly.
The green opponent still has heroes under his command, but the spell cannot be cast and indicates that no opponent has the possibility to have heroes under his command (multiplayer game and singleplayer game):

https://user-images.githubusercontent.com/43583994/230915320-23e05b99-091d-4e32-9a32-c6d6ab789ddb.mp4

https://user-images.githubusercontent.com/43583994/230915361-fbca1abc-493c-4e62-83c8-f70ebed5ed01.mp4

Here are the save files ("Before casting haunt 03.sav" (singleplayer game), "Identify Hero 01.savh" (multiplayer game), "Identify Hero 02.sav" (singleplayer game)):
[Before casting haunt Identify 03.zip](https://github.com/ihhub/fheroes2/files/11190909/Before.casting.haunt.Identify.03.zip)

Here are the test scenario files ("0 9922.MX2" (Haunt, Set Guardian), "0 9922b.MX2" (Identify Hero)):
[0 9922, 0 9922b.zip](https://github.com/ihhub/fheroes2/files/11190913/0.9922.0.9922b.zip)

LeHerosInconnu:
Hello @slawomir-witkowski, @ihhub and @Branikolog,

> Hi @LeHerosInconnu,
> I implemented rules you suggested. For testing purposes we always generate maximum number of small obstacles. So you can get a feel if battle area is not too clutered. I also added a change that allows to place single hex obstacles on column 8.

So I did a quick test on 20 combats. :)

I think we should avoid having too many small obstacles adjacent to each other or adjacent to a large obstacle.
Indeed what is the point of having small obstacles if they are all clumped together. :D

There could be a rule that allows only two small obstacles adjacent to each other among all the small obstacles.
And another one that allows only two small obstacles adjacent to a large obstacle.
What do you think about this?

Configurations with 3 two-hex obstacles and 2 one-hex obstacles seem to be more frequent than configurations with 4 one-hex obstacles and 1 two-hex obstacle, or 5 one-hex obstacles.
But this may be just a coincidence due to the number of combats (20).

In fheroes2:
Some examples of clumped obstacles:
Quick test obstacle 01 001
Quick test obstacle 01 004
Quick test obstacle 01 005
Quick test obstacle 01 010
Quick test obstacle 01 014
Quick test obstacle 01 017
Quick test obstacle 01 020

Here are the screenshots all the obstacle configurations during the quick test:
[Quick test obstacle 01.zip](https://github.com/ihhub/fheroes2/files/11191476/Quick.test.obstacle.01.zip)

slawomir-witkowski:
I added a change that prevents adjacent small obstacles. With the exception of second hex of a 2-hex obstacle.  Hex 83 is an example of this:
image

LeHerosInconnu:
Hello @slawomir-witkowski, @ihhub and @Branikolog,

> I added a change that prevents adjacent small obstacles. With the exception of second hex of a 2-hex obstacle. Hex 83 is an example of this:
> image

As can be seen in the following screeenshoots, one hexe obstacles can be adjacent in the original game (and also one hexe and two hexes obstacles).

While continuing my investigations on the positioning of obstacles on the battlefield, I observed these configurations:
In original game (let's play video):
Configuration of 5 obstacles with a group of 3 adjacent obstacles:
Obstacles adjacency 001
Obstacles adjacency 002

Configuration of 5 obstacles with two groups of 2 adjacent obstacles:
Obstacles adjacency 003

For now, we can say that only a group of 5 adjacent obstacles, and a group of 4 adjacent obstacles is not allowed, or has not yet been observed, or nobody remembers to have observed it, in the original game. :)

LeHerosInconnu:
Hello @slawomir-witkowski,

> Currently we don't place obstacles on top and bottom row. Don't you guys think we should?

Yes, small obstacles should be allowed to be placed in the top and bottom rows.

slawomir-witkowski:
Ok, I'll do that. It should reduce clumping. Another reason for clumping compared to OG is that we are not placing obstacles on first 3 and last 3 columns. So we have 2 columns less available to put obstacles on compared to OG.

slawomir-witkowski:
I pushed my changes. Now obstacles can also be on top and bottom row. There is no additional "anti-clumping" logic.  Can you take a look? Do we need to add some more rules?

LeHerosInconnu:
Hello @slawomir-witkowski and @Branikolog,

> Ok, I'll do that. It should reduce clumping. Another reason for clumping compared to OG is that we are not placing obstacles on first 3 and last 3 columns. So we have 2 columns less available to put obstacles on compared to OG.

I am not against allowing small obstacles in the third first and last columns.
@Branikolog what do you think?

> I pushed my changes. Now obstacles can also be on top and bottom row. There is no additional "anti-clumping" logic. Can you take a look? Do we need to add some more rules?

I did some new tests with the latest changes (80 combats on Grass, Dirt, Snow and Wasteland).

In fheroes2:
I observed only 1 configuration with 4 small obstacles grouped together and 1 configuration with 5 small obstacles grouped together:
GDSW 02 003
GDSW 02 065

@Branikolog, how do you feel about this?
For my part it is acceptable taking into account that it will be effective only in some rare configurations.

Here are the screenshots of all the obstacle configurations during this test (Grass, Dirt, Snow and Wasteland):
[D 02.zip](https://github.com/ihhub/fheroes2/files/11239819/D.02.zip)
[G 02.zip](https://github.com/ihhub/fheroes2/files/11239825/G.02.zip)
[SW 02.zip](https://github.com/ihhub/fheroes2/files/11239827/SW.02.zip)

For the sake of clarity, therefore, for the time being, these rules are taken into account:
- No large obstacle, up to 5 small obstacles (one and two hexes, maximum 3 two-hex obstacles)
- If the total number of hexes of the large obstacle(s) is less than or equal to 7 hexes, up to 4 small obstacles (one and two hexes, maximum 2 two-hex obstacles)
- If the total number of hexes of the large obstacle(s) is between 8 and 13 hexes, up to 3 small obstacles (one and two hexes, maximum 2 two-hex obstacles)
- If the total number of hexes of the large obstacle(s) is over 13 hexes, up to 2 small obstacles (one and two hexes, maximum 1 two-hex obstacle)
- In all cases, there can be no more than 2 large obstacles, regardless of their size.

And:
- No small obstacles allowed in the third first and last columns

@slawomir-witkowski, can you confirm this?

slawomir-witkowski:
@LeHerosInconnu I confirm, these are the implemented rules.

LeHerosInconnu:
Hello @slawomir-witkowski and @Branikolog,

After some more viewing of let's play Heroes 2 videos, it turns out that it is possible to have 6 small obstacles simultaneously on the battlefield. :)

In original game:

Six obstacles Heroes 2 01

We should therefore change the rule of "no large obstacle" to:
- No large obstacle, up to 6 small obstacles (one and two hexes, maximum 3 two-hex obstacles)

Note: There can be a battlefield without any obstacles, a battlefield with only 1 large obstacle, and a battlefield with only 2 large obstacles in the original game.

Branikolog:
Hi, @LeHerosInconnu and @slawomir-witkowski 

> > Currently we don't place obstacles on top and bottom row. Don't you guys think we should?
> 
> Yes, small obstacles should be allowed to be placed in the top and bottom rows.

I've made several dozens of tests of the OG and found out, that high objects are never placed on upper two rows:
fheroes2:
image

@LeHerosInconnu , could you confirm this? Those trees being cut doesn't look well for me.

> > Ok, I'll do that. It should reduce clumping. Another reason for clumping compared to OG is that we are not placing obstacles on first 3 and last 3 columns. So we have 2 columns less available to put obstacles on compared to OG.
> 
> I am not against allowing small obstacles in the third first and last columns. @Branikolog what do you think?

I can tell, that I even insist on placing objects there. :) The overall battlefield is not so large, so object being placed on first and last 3 colums would make bettlefields more diverse.

>I observed only 1 configuration with 4 small obstacles grouped together and 1 configuration with 5 small obstacles grouped
>For my part it is acceptable taking into account that it will be effective only in some rare configurations.

I agree. Probably, placing small (1 and 2 hex) objects close to each other could occur with some reduced chance. (something like reducing the chance of appearing of the next small object close to target by few % for each new object) 
Testing current PR build with maximum objects generated doesn't looks bad for me. Adding first and last 3 columns, I suppose, even make such groups occur even more rare. Am I right?

slawomir-witkowski:
I added: 
1. placement of small obstacles in the third first and last columns
2. no tall obstacles like trees on top 2 rows

After you guys take a look at how it works and accept, I can bring back random number of obstacles and we will have a final solution ready.

LeHerosInconnu:
Hello @slawomir-witkowski and @Branikolog,

> I added:
> 
> 1. placement of small obstacles in the third first and last columns
> 2. no tall obstacles like trees on top 2 rows
> 
> After you guys take a look at how it works and accept, I can bring back random number of obstacles and we will have a final solution ready.

So I did a test for that, works fine for me. :)

slawomir-witkowski:
I added randomization of small obstacles. This is the final solution.

LeHerosInconnu:
Hello @slawomir-witkowski,

> I added randomization of small obstacles. This is the final solution.

I've tested this, works fine. :)

Branikolog (approved):
Obstacle generation seems OK for me.
I feel, that 4-6 objects appear quite frequently, but I'm not sure it's much different from the original game..

src/fheroes2/battle/battle_board.cpp
```diff
@@ -550,21 +571,59 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
...
-        int32_t dest = GetRandomObstaclePosition( gen );
-        while ( at( dest ).GetObject() != 0 || ( checkRightCell && at( dest + 1 ).GetObject() != 0 ) ) {
+        // tall obstacles like trees should not be placed on top 2 rows
+        const bool isTallObstacle = isTallObject( objs[i] );
+
+        uint32_t dest;
+        do {
             dest = GetRandomObstaclePosition( gen );
-        }
+        } while ( at( dest ).GetObject() != 0 || ( checkRightCell && ( at( dest + 1 ).GetObject() != 0 || ( dest % 11 ) == 8 ) ) || ( isTallObstacle && dest <= 21 ) );
```
ihhub, line 620: Please replace the hardcoded value 11 by **ARENAW**.

Also could you please add a comment in the code why we check for 8.

ihhub, line 620: `dest < ( ARENAW * 2 )` would be better without using hardcoded values.

src/fheroes2/battle/battle_board.cpp
```diff
@@ -550,21 +571,59 @@ void Battle::Board::SetCobjObjects( const Maps::Tiles & tile, std::mt19937 & gen
...
+        maxTwoHexObstacleCount = 2;
+    }
+    else {
+        maxSmallObstacleCount = 2;
+        maxTwoHexObstacleCount = 1;
+    }
+
+    // Limit the number of two-hex obstacles to maxTwoHexObstacleCount
+    uint8_t twoHexCount = 0;
+    for ( size_t i = 0; i < objs.size(); ) {
```
ihhub, line 598: We can use pure iterators here as we don't use std::next() anywhere in the code:
```cpp
for ( auto iter = obj.begin() iter != obj.end(); ) {
    if ( isTwoHexObject( *iter ) ) {
        ++twoHexCount;
        if ( twoHexCount > maxTwoHexObstacleCount ) {
            iter = objs.erase( iter );
            continue;
        }
    }
    ++iter;
}
```
What do you think?

ihhub (changes requested):
Hi @slawomir-witkowski , I left few comments in this pull request. Would you mind to take a look at them when you have time?

src/fheroes2/battle/battle_board.cpp
slawomir-witkowski: I like your idea. It is much cleaner.

ihhub (approved):

ihhub:
@slawomir-witkowski , good job! Thank you very much for improving this logic!

PR #6912, opened 2023-03-29
Fix multiple typos

Moiman:


ihhub:
@Moiman , thank you so much for these fixes!

PR #6913, opened 2023-03-30
Improve AI fog discovery logic

ihhub:
All tiles covered by fog can be divided into 2 categories:
- tiles which are neighboring passable tiles. These tiles most likely be passable and might have some valuable objects
- tiles which are not reachable. These tiles are usually useless unless they are on another part of land / sea separated by impassable objects

In the majority of situations uncovering passable tiles is the best strategy and this is what actually most of human players do. With the proposed changes AI heroes tend to uncover "proper" tiles by expanding the territory which they can access.

### DIFF
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -630,8 +630,10 @@ uint32_t AIWorldPathfinder::getMovementPenalty( int src, int dst, int direction 
     return defaultPenalty;
 }
 
-int AIWorldPathfinder::getFogDiscoveryTile( const Heroes & hero )
+int AIWorldPathfinder::getFogDiscoveryTile( const Heroes & hero, bool & isTerritoryExpansion )
 {
+    isTerritoryExpansion = false;
+
     // paths have to be pre-calculated to find a spot where we're able to move
     reEvaluateIfNeeded( hero );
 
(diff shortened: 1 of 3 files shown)


src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -646,25 +648,25 @@ int AIWorldPathfinder::getFogDiscoveryTile( const Heroes & hero )
...
-        if ( start != currentNodeIdx ) {
+        if ( bestIndex == -1 && start != currentNodeIdx ) {
             int32_t maxTilesToReveal = Maps::getFogTileCountToBeRevealed( currentNodeIdx, scoutingDistance, _currentColor );
             if ( maxTilesToReveal > 0 ) {
                 // Found a tile where we can reveal fog. Check for other tiles in the queue to find the one with the highest value.
-                int bestIndex = currentNodeIdx;
-                for ( ; lastProcessedNode < nodesToExplore.size(); ++lastProcessedNode ) {
-                    const int nodeIdx = nodesToExplore[lastProcessedNode];
+                bestIndex = currentNodeIdx;
+                for ( size_t currentNodeId = lastProcessedNode + 1; currentNodeId < nodesToExplore.size(); ++currentNodeId ) {
```
oleg-derevenetz, line 661: Is there any way to rename this variable somehow? We already have `currentNodeIdx` and now we have also `currentNodeId` which is very similar and can be easily confused with the former one.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion, please take a look when possible.

idshibanov (approved):

src/fheroes2/world/world_pathfinding.cpp
ihhub: That's a tough task. Is **unprocessedNode** good in relation to **lastProcessedNode**?

oleg-derevenetz (approved):

Issue #5307, opened 2022-04-20
AI: 


BlkSwnNoir:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Saw mention that artifacts are placed into four tiers, insignificance of artifact? 
Could have unseen hero in area that is intended for use?

Through a couple different scenarios, the hero here seems fixated on heading towards second town of Sheltemburg 

![fheroes2_KiOuz616mL](https://user-images.githubusercontent.com/103940477/164152461-d382366a-e400-40fe-9277-4826cc50d186.gif)

### Save file

[Skips Artifact.zip](https://github.com/ihhub/fheroes2/files/8518289/Skips.Artifact.zip)

Branikolog:
Hi, @BlkSwnNoir , there could happen such thing, that artifact has special requirements as secondary skill or sum of gold. Could it happen here, that AI is just not able to pick it up at the moment?

zenseii:
Hi @Branikolog and @BlkSwnNoir . I just checked the save and the Artifact is not guarded nor needs a special skill nor resources.

My guess is that the AI sees the undefended castle of the human player and prioritizes that over getting the mine and the item. I wouldn't say it's a good choice.

![image](https://user-images.githubusercontent.com/12501091/164701177-742c0a22-3b75-481d-b840-2e76ecc915a6.png)

I also checked that the AI hero did not have a full inventory by defeating him.

idshibanov:
There's a special AI hero role called "Fighter" and they tend to focus on killing monsters/chasing enemy heroes rather than picking up things. I believe this is another example of unbalanced role priorities.

Ideally need to collect saves with similar examples to try to find a better balance there.

ihhub:
Reevaluating this as an improvement.

Issue #3589, opened 2021-06-01
Adventure map, AI behavior, hero Luna of the yellow AI player does not capture mines located near his movement path (0.9.3 3574)


LeHerosInconnu:
The hero Luna of the yellow AI player does not capture mines located near his movement path.
Later Luna attacks a green AI hero but does not capture the mine and moves away, to capture the mine the next turn.

In fheroes2:

https://user-images.githubusercontent.com/43583994/120367568-53a33f80-c311-11eb-91a6-c9c217a4cf6d.mp4

End the turn several times.

Here is the save file:
[Luna not capture mine move_.zip](https://github.com/ihhub/fheroes2/files/6578402/Luna.not.capture.mine.move_.zip)

**Edit 2021/11/25.**

Another save for this (0.9.9 4628).

Hero Halon of the green AI player does not pick up the Crystal resource pile or capture the Crystal mine.
I don't think he has anything really urgent to do with his army of few Zombies. :D
The next turn, Halon comes to his senses and turns back to capture the mine and pick up the resource pile.

In fheroes2:

![Halon Crystal mine 01](https://user-images.githubusercontent.com/43583994/143434902-585a9a94-3943-4959-b743-f326ec9455eb.png)

End the turn twice.

Here is the save file:
[Before Halon no take resourc.zip](https://github.com/ihhub/fheroes2/files/7602640/Before.Halon.no.take.resourc.zip)

ihhub:
Hi @LeHerosInconnu , is it still a valid issue?

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu , is it still a valid issue?

Another I'm putting on my "to be tested soon" list. :)

PR #6934, opened 2023-04-02
Consider object type while calculating object value over distances for AI heroes

ihhub:
Right now all objects decrease their value over distance to a hero. Logically it works fine but not always accurate. There are objects on maps which still should have high enough value over distance. For instance, a mine has a value of 2000. When an AI hero stands just 7 tiles away (700 move points) this value drops to ~0. For comparison, a resource located 3 tiles away (300 move points) will still have value 100. The formula should be adjusted for some objects so they don't loose their value drastically. On the other hand, some objects should not be worth visiting if they are far away like Luck modifiers.

relates to #3589, #5307

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1604,41 +1604,41 @@ namespace AI
         ObjectValidator objectValidator( hero, _pathfinder, *this );
         ObjectValueStorage valueStorage( hero, *this, lowestPossibleValue );
 
-        auto getObjectValue = [&objectValidator, &valueStorage, this, heroStrength, &hero, leftMovePoints]( const int destination, uint32_t & distance, double & value,
-                                                                                                            const MP2::MapObjectType objectType,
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @Districh-ru , I included you into the list of reviewers to get your opinion about the formula of object values over distances. It can be easily represented in a 3D space where a top of a hill is an object and the formula is used to calculate the slope of the hill. I think for certain objects the slope should be steeper if an object is valuable only on short distances and flatter slope if object's value should be in focus even over big distances.

Districh-ru (approved):
LGTM 👍 
In future we may want to adjust the coefficients for some objects and maybe to use different non-linear scales for some objects.

oleg-derevenetz (approved):

idshibanov (commented):
Good start. I thought you're going to adjust the slope more drastically - i.e. choosing something else than nlog(n) based on type.

ihhub:
Hi @idshibanov , we will adjust it later. Right now we made a ground for the future fine tuning.

Issue #6845, opened 2023-03-16
Editor preparations


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

The next major goal of the project is deliver an in-game editor. This issue serves as a collection of tasks related to the main task.

### Describe the possible solution

This is a list of tasks we have to complete in order to deliver a base version of the Editor:
- [x] remove usage of quantity1 and quantity2 in the code. Make them more readable without the calculation magic present in the code.
- [x] Tiles class should just have a stack of all TilesAddon. Move the same members of the Tiles class away from the class. This will make the code much cleaner and easier to maintain
- [x] Move all auxiliary methods outside of the Tiles class. Tiles class is designed just to store information about a tile in general, not about specific objects
- [x] Create an information table for each object in the game: which images are used to determine the object, its properties, passabilities
- [x] Remove unused structures in mp2.h file and expand the code with necessary explanations within comments where we actually read data for these structures
- [x] add a shortcut key (E) to open an Editor main menu and add there needed buttons like it is done in the original Editor (for now only for debug mode). The main menu image must be resizable
- [x] add a new mode within interface_area to enable Editor mode (something similar what we did for view world
- [x] add UI elements to handle terrain types and change cursor type based on a chosen type
- [x] change terrain type using terrain brush (enable support for keeping objects on terrain as well as selecting an area for terrain)
- [x] add historical changes support: undo and redo functionality
- [x] propose a map format which stores objects as whole and their data. For example, a castle must have only 1 main tile while in the editor and the game it will populate the needed number of tiles.
- [x] add required helper functions into the code to load new map format. These functions should be shared among the Editor and the game
- [x] add a UI toolset to handle objects within the Editor. It should be a different way than in the original game
- [x] display passabilities for objects on the map
- [x] figure out text encoding support for maps
- [x] investigate whether it is required to have a separate format for campaign scenarios
- [x] change `MapObjectType` type to uint16_t to support new objects
- [x] add a window to change Hero properties
- [x] add a window to change Castle / Town properties
- [x] add a window to change Sphinx properties
- [x] add a window to change Bottle / Sign properties
- [x] add a window to change Jail properties
- [x] add a window to change Random Hero properties
- [x] add a window to change Random Town properties
- [x] add a window to change Event properties
- [x] add a window to change Monster properties
[...]

PR #6940, opened 2023-04-04
Cleanup MP2 format and Maps::Tiles class

ihhub:
- remove the remaining unused structures in `mp2.h` file
- move several methods out of Maps::Tiles class as they don't belong there being too specific to certain object types. Not all methods were moved to keep this pull request relatively small.

relates to #6845


ihhub:
Hi @Branikolog , could you please test general map gameplay for new games and from saves since this pull request touching important parts of the code responsible for retrieving information from objects regarding monster types, gold, artifacts and spells.

oleg-derevenetz (approved):

PR #6946, opened 2023-04-04
Improve the destination indication for Teleport spell

### LINKED ISSUE #6941: Combat, Teleport spell, both destination hexagons should be indicated for two-hexagon creatures
### Preliminary checks

In fheroes2, for Teleport spell, both destination hexagons should be indicated for two-hexagon creatures.

In fheroes2:

https://user-images.githubusercontent.com/43583994/229849517-510d566a-d74e-4e0b-b92b-4eaecae9d35d.mp4

Both destination hexagons should be indicated for two-hexagon creatures.

### Additional info

[0 9922.zip](https://github.com/ihhub/fheroes2/files/11150486/0.9922.zip)

oleg-derevenetz:
close #6941

Also, this PR fixes the issue when cells are highlighted, even if the current spell does not apply to units on these cells. Currently on the `master` branch:

https://user-images.githubusercontent.com/32623900/229931766-7bfca010-d673-42ae-8d60-fa9bcf6f06e4.mp4

With this PR:

https://user-images.githubusercontent.com/32623900/229931819-4463921c-0218-46c8-96b6-364beba7429d.mp4

This change may be controversial, but I believe that since we do not highlight the cells with dragons in this video, we should not also highlight the cells with other units on which this spell does not work. Empty cells will continue to be highlighted as before.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1145,7 +1145,7 @@ Battle::Interface::Interface( Arena & battleArena, const int32_t tileIndex )
     , _flyingUnit( nullptr )
     , b_current_sprite( nullptr )
     , index_pos( -1 )
-    , teleport_src( -1 )
+    , _teleportSpellSrcIdx( -1 )
     , listlog( nullptr )
     , _cursorRestorer( true, Cursor::WAR_POINTER )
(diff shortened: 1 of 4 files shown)


ihhub (approved):

ihhub:

PR #6959, opened 2023-04-08
Expand AI control debugging functionalities

ihhub:
- ~~AI heroes use Dimension Door spell to get close to objects which require to stand in front of them. Once a hero stands in front of it nothing happens and the hero just uses the spell more and more times. This is due to the condition of movement cost which we assumed to be greater than 0 for normal movement.~~
- added an ability to regain control from AI during Dimension Door usage by AI (not a very elegant solution by it works)
- add autosave logic when a human player is controlled by AI. It is helpful if you want to repeat the same actions again
- object types were not being updated during AI turn so we could see object Heroes instead of Castle when a hero tries to come back to the same castle where it came from within a single turn (this is more like a cosmetic fix, no logic changes)
- small code cleanup

~~You can try to use this save file to test against master branch:  [AI_jumps_around.zip](https://github.com/ihhub/fheroes2/files/11182747/AI_jumps_around.zip)~~

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1993,11 +1993,11 @@ namespace AI
         if ( Players::Get( hero.GetColor() )->isAIAutoControlMode() ) {
             LocalEvent::Get().HandleEvents( false );
             if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_TRANSFER_CONTROL_TO_AI )
-                && fheroes2::showStandardTextMessage( _( "Warning" ), _( "Do you want to regain control from AI? The effect will take place only on the next turn." ),
-                                                      Dialog::YES | Dialog::NO )
-                       == Dialog::YES ) {
-                    Players::Get( hero.GetColor() )->setAIAutoControlMode( false );
-                    return;
+                 && fheroes2::showStandardTextMessage( _( "Warning" ), _( "Do you want to regain control from AI? The effect will take place only on the next turn." ),
+                                                       Dialog::YES | Dialog::NO )
+                        == Dialog::YES ) {
+                Players::Get( hero.GetColor() )->setAIAutoControlMode( false );
+                return;
             }
         }
 #endif
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1990,13 +1990,14 @@ namespace AI
         }
 
 #if defined( WITH_DEBUG )
-        LocalEvent::Get().HandleEvents( false );
-        if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_TRANSFER_CONTROL_TO_AI ) && Players::Get( hero.GetColor() )->isAIAutoControlMode() ) {
-            if ( fheroes2::showStandardTextMessage( _( "Warning" ), _( "Do you want to regain control from AI? The effect will take place only on the next turn." ),
-                                                    Dialog::YES | Dialog::NO )
-                 == Dialog::YES ) {
-                Players::Get( hero.GetColor() )->setAIAutoControlMode( false );
-                return;
+        if ( Players::Get( hero.GetColor() )->isAIAutoControlMode() ) {
+            LocalEvent::Get().HandleEvents( false );
+            if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_TRANSFER_CONTROL_TO_AI )
+                && fheroes2::showStandardTextMessage( _( "Warning" ), _( "Do you want to regain control from AI? The effect will take place only on the next turn." ),
+                                                      Dialog::YES | Dialog::NO )
+                       == Dialog::YES ) {
+                    Players::Get( hero.GetColor() )->setAIAutoControlMode( false );
+                    return;
             }
         }
 #endif
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1862,7 +1862,8 @@ namespace AI
                 if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_TRANSFER_CONTROL_TO_AI ) && Players::Get( hero.GetColor() )->isAIAutoControlMode() ) {
                     if ( fheroes2::showStandardTextMessage( _( "Warning" ),
                                                             _( "Do you want to regain control from AI? The effect will take place only on the next turn." ),
-                                                            Dialog::YES | Dialog::NO ) == Dialog::YES ) {
+                                                            Dialog::YES | Dialog::NO )
+                         == Dialog::YES ) {
                         Players::Get( hero.GetColor() )->setAIAutoControlMode( false );
                         continue;
                     }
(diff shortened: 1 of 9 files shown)


src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -1092,14 +1092,33 @@ std::list<Route::Step> AIWorldPathfinder::getDimensionDoorPath( const Heroes & h
...
+
 std::list<Route::Step> AIWorldPathfinder::buildPath( const int targetIndex, const bool isPlanningMode /* = false */ ) const
 {
     assert( _pathStart != -1 && targetIndex != -1 );
 
     std::list<Route::Step> path;
 
-    // Destination is not reachable
     if ( _cache[targetIndex]._cost == 0 ) {
+        // It could happen in 3 situations:
```
oleg-derevenetz, line 1112: Are you sure that pathfinder is not able to handle these situations on its own? I have an impression that it should. Also, the movement cost in situations 2 and 3 is not zero.

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions, please take a look when you have time.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1986,9 +1984,24 @@ namespace AI
...
 
+#if defined( WITH_DEBUG )
+        if ( Players::Get( hero.GetColor() )->isAIAutoControlMode() ) {
+            LocalEvent::Get().HandleEvents( false );
+            if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_TRANSFER_CONTROL_TO_AI )
+                 && fheroes2::showStandardTextMessage( _( "Warning" ), _( "Do you want to regain control from AI? The effect will take place only on the next turn." ),
+                                                       Dialog::YES | Dialog::NO )
+                        == Dialog::YES ) {
+                Players::Get( hero.GetColor() )->setAIAutoControlMode( false );
+                return;
```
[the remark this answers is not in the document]
ihhub: Nope. If you stop in the middle of AI turn you have to wait till the end. This is the same behavior if you stop in the middle of normal hero walking process.

src/fheroes2/world/world_pathfinding.cpp
ihhub: Based on the provided save file AI is not able to get a path to a monster standing next to it. Hence, I added this fix. Something is wrong with pathfinder for sure.

src/fheroes2/ai/ai_hero_action.cpp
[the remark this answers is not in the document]
oleg-derevenetz: Can we at least let the current spell cast to finish? Mana and movement points are already spent anyway :)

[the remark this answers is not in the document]
ihhub: I meant that you can't stop AI in the middle of the turn. It will still continue to do the stuff till the end of the turn.

src/fheroes2/world/world_pathfinding.cpp
oleg-derevenetz: > Based on the provided save file AI is not able to get a path to a monster standing next to it.

If this is really the case, then it's the AI pathfinder issue and we need a proper fix. It should be able to handle these cases "naturally", just like the human player's pathfinder does it.

src/fheroes2/ai/ai_hero_action.cpp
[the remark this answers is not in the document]
ihhub: Ah, I got your point. I need to remove **return** :)

src/fheroes2/world/world_pathfinding.cpp
oleg-derevenetz: I'd say that AI pathfinder is fine:

https://user-images.githubusercontent.com/32623900/230716429-e0d1b7b9-6914-4c84-8d1c-96dfb258743f.mp4

AI is perfectly able to get path to a monster standing next to it and attack it.

ihhub: Yes, you are right. Somehow AI isn't able to use Dimension Door pathfinding properly.

ihhub: So the problem lies in `AIWorldPathfinder::getDimensionDoorPath` method...

oleg-derevenetz: Are you sure that your fixes in `AI::Normal::HeroesTurn()` are not enough by themselves?

ihhub: I still feel it like a hack, to be honest.

ihhub: The root of the problem is that pathfinding depends on army ratio parameters which stands currently at 1.8. If a hero army strength is lesser than (monster's strength * the coefficient) then the tile is blocked. It is useful when you want to create a path through monsters and heroes but it fails for cases when you want to attack a monster.

ihhub: if you are fine with the fix as of now then I am okay. Frankly speaking I have no idea how to fix it in an elegant way.

oleg-derevenetz: I need to think a bit more.

Branikolog:
Hi, @ihhub 
I've found some weird behaviour of AI controlled hero: he teleports towards treasure chest, but then teleports to capture sulfur mine without gathering the chest:

https://user-images.githubusercontent.com/55348946/230720407-b964c6d4-592c-4752-aa9e-65838095e6af.mp4

src/fheroes2/world/world_pathfinding.cpp
ihhub: I am reviewing pathfinding logic, it looks wrong. When a hero reaches position 755 he should be able to attack monsters but they are unreachable.

ihhub: Hi @idshibanov , could you please help here if you have time? I think it is an important fix we need to do for AI in order to work correctly.

ihhub:
Hi @Branikolog , this is not strictly speaking "invalid" behavior. Capturing a mine is more valuable that collecting a chest. Multi-target tasking is not implemented.

ihhub:
Hi @oleg-derevenetz and @idshibanov , I reverted back changes and partially fixed the original problem. We were not processing tiles under monster protection at all. The goal of the original logic is not to allow heroes go through monsters and their protected territory which is fixable by just not processing other tiles for the protected one.

As the result AI behaves the same as my original attempt but normal pathfinding still doesn't work as intended and a hero cannot reach a monster by any means expect Dimension Door.

ihhub:
And now AI hero is trying to killing himself while capturing a mine :(

Branikolog:
Hi, @ihhub 

> Hi @Branikolog , this is not strictly speaking "invalid" behavior. Capturing a mine is more valuable that collecting a chest. Multi-target tasking is not implemented.

I think collecting a resource that will cost only one movement is not such a loss of time for a hero, even if it's a champion one. This behaviour is more correct, than teleporting to a resource and then leaving it because there appeared several new points of interest. Indeed, if there was discovered really some valuable point, like weak castle/hero and the teleporting hero has really limited mov. points to reach the new point, so he can leave the resource. But such situations appear in much rare cases and the current behaviour will confuse players much more often ("why this hero didn't collect that treasure chest and run away?"). I think the situation, when AI hero collected a resource and because of that he couldn't reach a castle to capture looks more natural, than the case when he's constantly leaving resources right in front of him. Even human player players can sometimes spend mov. points not effectively.
I think, while teleporting AI hero should rise the priority of the target objects as leaving them seems like a waste of mana to nowhere, like in the video I attached above. And on contrary, when AI hero is moving "on horse" he should update priority on revealing new points of interest. Actually, it currently does not and I'm going to create an issue the nearest time.

ihhub:
Hi @Branikolog , I stripped off changes related to AI fixes in this pull request as @oleg-derevenetz fixed them temporary in #6966 .

oleg-derevenetz (approved):

PR #6966, opened 2023-04-08
Do not allow to access overprotected tiles using the Dimension Door spell

oleg-derevenetz:
Related to #6959

Should fix an inability to access some tiles that require to stand in front of them (i.e. overprotected tiles that has too strong army of wandering monsters on them). "Too strong army" is detected using the same rules as `AIWorldPathfinder` uses for the "normal" movement. This change still allows to "steal" pickupable objects from under the nose of guardians:

https://user-images.githubusercontent.com/32623900/230740334-0f082bff-f1e7-4319-b39b-a3227ed05f80.mp4

Hi @ihhub could you please verify this change in regard to #6959? Does it resolve the corresponding issues?

### DIFF
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -1010,8 +1010,7 @@ std::list<Route::Step> AIWorldPathfinder::getDimensionDoorPath( const Heroes & h
         return {};
     }
 
-    const MapsIndexes & monsters = Maps::getMonstersProtectingTile( targetIndex );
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for the fix! It solves the initial problem. However, the pathfinding is incorrect in general but we can fix it for the next release.

oleg-derevenetz:
@ihhub what do you mean by "incorrect in general"? Do you mean the entire principle that tiles with a too strong protection is inaccessible by AI is incorrect?

PR #6972, opened 2023-04-09
Version 1.0.3

ihhub:
Polishing the game.


ihhub:
Sadly but we have to go with the release as Google Play Store release might take some time to be approved.

Issue #6822, opened 2023-03-12
Google Play Store release preparations


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

We would like to release fheroes2 in Google Play Store and we need to make sure that this is going to be flawless.

### Describe the possible solution

 I still think that there might be issues with the release and this is a list of needed steps to make we should do:
- [ ]  add more translations for Java related code. Right now only English and Russian are supported
- [ ] possibly detect OS language and set it in the configuration file for the first run
- [ ] add support of other languages in virtual keyboard (at least the main languages in which the OG was released
- [ ] create a set of screenshots to be used in the store (are these in the main page good?)
- [ ] create a description of the application, stating clearly that this is an engine

Feel free to express your opinion about the steps.

oleg-derevenetz:
> Right now only English and German are supported

And Russian.

> possibly detect OS language and set it in the configuration file for the first run

I recommend to let the user select the in-game language at the first run, just like it works now, because Android supports much more languages than supported by the fheroes2 (and original game resources supports even less languages).

> add support of other languages in virtual keyboard (at least the main languages in which the OG was released

I believe this is optional and not vital really :)

Alucard648:
What about also porting to RuStore? For russian auditory.

oleg-derevenetz:
> What about also porting to RuStore? For russian auditory.

In fact, there will be no need for any special porting, already released APK could be used (APK signature should ideally be the same to allow installation from one store and upgrade from another store (or manually)). But AFAIK RuStore has some limitations regarding open source projects driven by a team (they don't have an idea who is personally responsible for a project in such cases), and they were going to make a special procedure for the publishing of such projects. But I'm not aware of the details and timeline.

ihhub:
Hi @oleg-derevenetz , thanks for the input! I corrected the list.

Mr-Bajs:
Android is getting more restrictive. After the android release of fheroes2 Google has as far as I can tell, trying to restrict users access to the android/data folders. There seams to be a fair amount of annoyance on the web when searching for a solution about this. 

I assume publishing on Google Play will add alot of new users, and probably alot issues and problems with people that can't load custom maps or delete save files from a bloated save files folder.
 
[...]

PR #6973, opened 2023-04-09
Add a document for privacy policy for the project

ihhub:
It is required for Google Play Store. It is simple as we don't collect any personal information.

relates to #6822

### DIFF
--- a/docs/PRIVACY_POLICY.md
+++ b/docs/PRIVACY_POLICY.md
@@ -1,0 +1,17 @@
+# [**fheroes2**](README.md) privacy policy
+
+## General
+
+**fheroes2** is an open source project and it does not collect any personal
+information in any of its official GitHub and Google Play Store releases.
+The project does not contain any library or programming module from
+third-party suppliers which collects sensitive personal information from its users.
+
+## Contact information
+
+If you have any questions or concerns about any of the matters in relation to
+this Privacy Policy then there are several ways of contacting us:
+- via email to fhomm2@gmail.com
+- via Discord at https://discord.gg/xF85vbZ by contacting any of the collaborators
+
+This Privacy Policy was last updated 04th April 2023.
--- a/docs/PRIVACY_POLICY.md
+++ b/docs/PRIVACY_POLICY.md
@@ -1,5 +1,10 @@
 # [**fheroes2**](README.md) privacy policy
 
+**fheroes2** is provided at no cost and is intended for use as is.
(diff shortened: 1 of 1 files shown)


ihhub:
> This looks good to me, even if I have no prior experience with writing such policies.
> 
> It's a bit thought-provoking how **not** collecting any data seems to be a rare thing for apps on the play store.

That's a reality we live in. Most of apps collect personal information in one or another form.

Districh-ru (approved):
I have no experience in writing such documents, but its look good to me.

Maybe we can add a phrase (it's a copy-past from some other policy):
"fheroes2 is provided at no cost and is intended for use as is.
 This document is used to inform regarding our policies with the collection, use, and disclosure of Personal Information if anyone decided to use fheroes2.
If you choose to use fheroes2, then you agree to this policy. 
We will not collect, use or share your information with anyone. We do not collect any personally identifiable information from children."

Also I don't know how it is organized, but Android possibly may create the error report (I use modded firmware with disabled error reporting on my cell phone, so I don't know how it works and if it works). If there may be such error report, then we can add something like:
"We want to inform you that whenever you use fheroes2, in a case of an error in the app the data and information can be collected (through third-party products) on your phone called Log Data. This Log Data may include information such as your device name, operating system version, the time and date of your use of fheroes2, and other statistics."

oleg-derevenetz:
> Also I don't know how it is organized, but Android possibly may create the error report

Unlike iOS that [uploads crash reports to the AppStore automatically](https://developer.apple.com/documentation/xcode/acquiring-crash-reports-and-diagnostic-logs), Android doesn't do this by default and requires explicit integration of some crash reporting framework, for example, [Firebase Crashlytics](https://firebase.google.com/products/crashlytics) (Crashlytics is the Google's product, but nonetheless it requires explicit integration and doesn't work "silently").

ihhub:
Hi @Districh-ru and @zenseii , I updated the document by adding @Districh-ru suggestion (not full but most of it). Please take a look when you have time.

idshibanov (approved):

ihhub:
Hi @oleg-derevenetz , if you have time please review this pull request by today so I can prepare everything for the Google Play Store release?

oleg-derevenetz (approved):

ihhub:
Hi @zenseii , if you think that the last change needs correction please comment and I can create another pull request for this.

PR #6975, opened 2023-04-09
Android: build AAB in addition to APK

oleg-derevenetz:
Related to #6822


ihhub:

PR #6976, opened 2023-04-09
Android: remove non-translatable string from Polish localization

oleg-derevenetz:
Demo version URL is marked as non-translatable and is not need to be translated.


ihhub:

PR #6979, opened 2023-04-09
Norwegian translation update

zenseii:
Bring it up to date and make several corrections. Changed use of board to map.


ihhub:
@zenseii , huge thanks for this translation update!

PR #6993, opened 2023-04-12
Update Slovak translation

Roxterat:
Slovak translation updates


Roxterat:
Hey @zenseii ! What is "disable build." line (about line 1700, just before "Cannot afford %{name}." line) used for? If for anything at all..

Roxterat:
> > Hey @zenseii ! What is "disable build." line (about line 1700, just before "Cannot afford %{name}." line) used for? If for anything at all..
> 
> Hi, @Roxterat.
> 
> I checked and as I remembered this is a string that refers to the status bar message when you hover over a building icon in the castle options screen and this particular building has been disabled. AFAIK it is not possible to disable buildings in the original map editor so there's no easy way nor original way to make this message appear in the game.
> 
> The fact that the message should be spelled "Building disabled." instead of the current incorrect English also supports this assumption.
> 
> In Heroes 2 only the shipyard and the castle are ever disabled, and both of these have unique descriptions in those cases.
> 
> Please mark this PR as ready for review when you want me to review it.

I see, thank you for cheking it out!
Should be ready for a review now.

ihhub (approved):

ihhub:
@Roxterat , thank you so much for this update!

PR #7004, opened 2023-04-13
Redraw the status window after the amount of kingdom resources has changed at the beginning of the day, fix recurring events

### LINKED ISSUE #7002: Adventure map, resource counters are not updated on the beginnig of a new day
### Preliminary checks

### Platform

Windows

### Describe the bug

While playing on high resolutions resources are always visible in right bottom corner. In the begginning of a new day resource counters are not updating until player performs any action:

https://user-images.githubusercontent.com/55348946/231872983-33d4c94d-998f-4241-bea4-c3a332289cab.mov

### Save file

None.

### Additional info

_No response_

oleg-derevenetz:
fix #7002

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -857,6 +857,10 @@ fheroes2::GameMode Interface::Basic::HumanTurn( const bool isload )
             }
         } );
 
+        // The amount of the kingdom resources has changed, the status window needs to be updated
+        Redraw( REDRAW_STATUS );
+        display.render();
+
         if ( conf.isAutoSaveAtBeginningOfTurnEnabled() ) {
             Game::AutoSave();
         }
--- a/src/fheroes2/world/world.cpp
+++ b/src/fheroes2/world/world.cpp
@@ -1578,7 +1578,7 @@ bool EventDate::isAllow( const int col, const uint32_t date ) const
         return false;
     }
 
-    if ( firstOccurrenceDay < date ) {
+    if ( firstOccurrenceDay > date ) {
         // The date has not come.
         return false;
     }


Branikolog:
Hi, @oleg-derevenetz !
Counters are fixed now. The only thing I've noticed, is on the first day of the week, they are refreshing only after we press ok in "new week" window. It's not critical, though. How do you think, should it happen before new week window appears?

oleg-derevenetz:
> How do you think, should it happen before new week window appears?

Resource updates now occur simultaneously with the processing of date-bound events (they can also add and take away kingdom resources), so it would be strange if such events were triggered before the new week message.

Branikolog:
Hi, @oleg-derevenetz !

> so it would be strange if such events were triggered before the new week message.

Cannot say it's just "before"... For me it looks "simultaneously". Together with the message on the screen we watch updated resource stats.
For some reason, daily events are not working for me now. They're triggered only on the first day. Could you, please, check if something went wrong the latest releases? I cannot even check, how things in this PR are happening with the resource withdraw event on a new week.
In my personal point, the couters should be updated simultaneously with the "new week" message. Current implementation of resource withdrawal works in such manner, where resources are withdrawing with the message appeared on the screen. So accepting "new week" message you won't even see, how res counters are changing (I suppose). Or should we better update res counters for withdrawing event after we accept the message?

oleg-derevenetz:
@Branikolog

> Cannot say it's just "before"... For me it looks "simultaneously". Together with the message on the screen we watch updated resource stats.

I mean the following events:

https://user-images.githubusercontent.com/32623900/231984756-fb704b9d-885d-4978-ad2c-c064e53f35f8.mp4

Currently on 1/2/1 the "new week" message is shown first, then the event message is shown and then the amount of resources is updated according both to everyday income and income from event. You cannot show these windows "simultaneously", either the "new week" message should be shown first or the event message. And if the event message will be shown first (before the "new week message") it would look strange IMHO, because from the logical point of view the new day/week comes _first_, and only _then_ you get the event (along with resources).

oleg-derevenetz:
> For some reason, daily events are not working for me now.

Recurring events should be fixed now as well.

ihhub (approved):

ihhub:

PR #7006, opened 2023-04-14
Fix a crash on deleting last save file

undef21:
Also replace `System::Unlink` with `std::filesystem::remove`

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -440,15 +440,6 @@ bool System::IsDirectory( const std::string & path, bool writable )
 #endif
 }
 
-bool System::Unlink( const std::string & path )
-{
-#if defined( _WIN32 )
-    return _unlink( path.c_str() ) == 0;
-#else
-    return unlink( path.c_str() ) == 0;
-#endif
-}
-
 #if !defined( _WIN32 ) && !defined( ANDROID )
 // based on: https://github.com/OneSadCookie/fcaseopen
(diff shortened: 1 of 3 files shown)


oleg-derevenetz:
Hi @undef21 while using `std::filesystem` is fine of course, I still recommend to use it in the `engine` part and export some function (like `System::Unlink`, but maybe with different name, it's up to you) to avoid exposing implementation details when they are not really needed - for example, including the `<system_error>` header and using the `ec` variable.

What is usually needed is to remove some file and (optionally) find out if the deletion was successful, with such special function this can be done in one line.

Also this reduces the risk that someone will forget about the non-throwing overload and will use the overload without the `std::error_code` parameter.

What do you think?

undef21:
Hi @oleg-derevenetz 
OK, I'll move std::filesystem-related changes to another PR.

oleg-derevenetz:
BTW in this old code the result of `Unlink()` is not checked at all and save file is removed from the list even if it wasn't in fact deleted. This is of course is not directly related to the topic of this PR, but still looks not good :)

undef21:
> BTW in this old code the result of `Unlink()` is not checked at all and save file is removed from the list even if it wasn't in fact deleted. This is of course is not directly related to the topic of this PR, but still looks not good :)

fixed in #7009

ihhub:
@undef21 , thank you very much for the fix!

PR #7009, opened 2023-04-14
Use std::filesystem instead OS-specific funtions

undef21:


undef21:
Hi @ihhub . This pr turned out to be more difficult than I initially thought. The main problem is the game uses single-byte encoding while modern filesystem uses multi-byte encoding. So I'm closing this pr until a solution for multi-byte encodings is found.

PR #7015, opened 2023-04-15
Modernize the Android Toolset

oleg-derevenetz:
While current version in `master` branch is quite workable, it is not written in a completely canonical way - it doesn't use a `ViewModel` architecture. As a result, there are issues with UI when `Activity` is recreated during any background operation (such as ZIP archive extraction or reading a list of save files). Such recreation usually occurs when the screen is rotated (in `master` branch screen rotation is disabled during a background operation as a workaround), when UI theme is changed between day and night, when system language is changed, and other similar operations. With this PR recreation of `Activity` is handled in the right way even during a background operation.

Most likely you will need latest Android Studio (2022.2.1 Flamingo) to build this PR.


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for this update!

PR #7019, opened 2023-04-16
Reset battlefield grid cursor position after using the Spell Book

Districh-ru:
The bug was found by **drevoborod** on Discord.
If your cursor pointer is not _WAR_NONE_ (X) and you call the Spell Book by hotkey ("C") and try to use the **Teleport** spell you will encounter an assertion that checks that the cursor type is not _WAR_NONE_ or _SP_TELEPORT_.
You can just simple point your creature and call the Spell Book:
cursor

This happens as the cursor position and type were not updated from the state before opening the Spell Book.

This PR resets the battlefield cursor position on battlefield after using the Spell Book.


ihhub (approved):

ihhub:

Issue #6952, opened 2023-04-06
Adventure map, adjustment for Summon Boat spell


LeHerosInconnu:
Hello @Kazhuu,

I did some tests with the summon boat spell.
It turns out that if there are no more neutral boats (not belonging to any player) or player-built boats available and there is at least one free square adjacent to the hero, the spell is not grayed out, as it should be, and if the hero casts the summon boat spell the message "Summon Boat failed!!!" is displayed.
This is what we want to remove with these improvements.

- Save file: Before boats 02.savh (multiplyer game)(video below).
There are two neutral boats (not belonging to any player) at the beginning of the scenario.
Troyan casts the summon boat spell twice, the two neutral boats are summoned near Troyan.
After that, the summon boat spell is not grayed out in Troyan's magic book, although it should be because there is no other boat to summon.
There is one free square adjacent to Natasha, but there is no additional boat available.
If Troyan casts the non-greyed spell, the message "Summon Boat failed!!!" is displayed. This is what we want to remove with these improvements.

https://user-images.githubusercontent.com/43583994/230159377-38156be6-52ef-4447-becd-e4026053ea6c.mp4

- Save file: Before boats 01.savh (multiplyer game)(video below).
There are four boats available, two neutral boats from the beginning of the scenario and two boats built at the Weed Patch shipyard.
Natasha casts the summon boat spell four times, the four boats are summoned near Natasha.
After that, the summon boat spell is grayed out in Natasha's magic book.
There is no free square adjacent to Natasha (and furthermore, there is no additional boat available).
In this case, everything works as intended.

https://user-images.githubusercontent.com/43583994/230160100-e0c07105-cdc0-4b7b-bfac-7f8352a0c9f4.mp4

- Save file: Before boats 02.savh (multiplyer game)(video below).
There are two neutral boats (not belonging to any player) at the beginning of the scenario.
Troyan casts the summon boat spell twice, the two neutral boats are summoned near Troyan.
After that, the summon boat spell is not grayed out in Troyan's magic book, although it should be because there is no other boat to summon.
There is one free square adjacent to Natasha, but there is no additional boat available.
If Troyan casts the non-greyed spell, the message "Summon Boat failed!!!" is displayed.
A boat is built at the Weed Patch shipyard.
Troyan casts the summon boat spell, the just built boat is summoned near Troyan.
After that, the summon boat spell is grayed out in Troyan's magic book.

[...]

Issue #4849, opened 2021-12-26
Adventure map, Summon Boat spell, replace the word "water" by the word "ocean" in the message window that appears when the hero casting the spell is not positioned near the ocean


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

In fheroes 2, as a square of the sea is designated by the term "Ocean" in the right click information window, I think it would be more appropriate to use this word in the texts referring to it.

In fheroes 2:

![Nearby ocean 01](https://user-images.githubusercontent.com/43583994/147409179-0d57fc5c-2667-40fb-9b74-20455b52eca5.png)

![Nearby ocean 02](https://user-images.githubusercontent.com/43583994/147409186-7c3369fc-1830-46fe-8c50-9a2ad9a48db6.png)

### Describe the possible solution

I suggest to replace the word "water" by the word "ocean" in the message window that appears when the hero casting the spell is not positioned near the ocean.

This text:

"...nearby ocean."

Instead of:

"...nearby water."

zenseii:
Since this could be even for lakes and not just oceans, I would say a better wording is "near a body of water".

LeHerosInconnu:
Hello @zenseii,

> Since this could be even for lakes and not just oceans, I would say a better wording is "near a body of water".

The purpose here is to refer to the type of terrain "Ocean" and not to the area covered by the water. :)
The word "Ocean" is displayed in the information window of the terrain square with the right click.

PR #7020, opened 2023-04-16
6952: Improve Summon Boat spell grayed out messages

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/6952 and https://github.com/ihhub/fheroes2/issues/4849 by adding more logic to Summon Boat adventure spell when it is grayed out.


src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -303,38 +296,28 @@ namespace
...
+        // Player should have summonable boats before calling this function.
+        assert( !summonableBoats.empty() );
 
-            const uint32_t distance = Maps::GetStraightLineDistance( boatSource, hero.GetIndex() );
+        const int heroColor = hero.GetColor();
 
-            if ( distance > 1 ) {
-                Interface::GameArea & gameArea = Interface::Basic::Get().GetGameArea();
-                gameArea.runSingleObjectAnimation( std::make_shared<Interface::ObjectFadingOutInfo>( tileSource.GetObjectUID(), boatSource, MP2::OBJ_BOAT ) );
+        const int32_t boatSource = summonableBoats.at( 0 );
```
ihhub, line 309: **summonableBoats.front()** will do the same.

src/fheroes2/spell/spell_info.cpp
```diff
@@ -358,4 +360,32 @@ namespace fheroes2
         }
         return boatDestination;
     }
+
+    Maps::Indexes getSummonableBoats( const Heroes & hero )
```
ihhub, line 364: The return array is never used and only first value from it is being used. I suggest to just return an index of a boat. If it is less than 0 then it is an invalid position. In this case the code will be faster.

In this case can we just use **getPossibleBoatPosition()** function for our needs?

ihhub (changes requested):
Hi @Kazhuu , I left multiple comments here. Could you please check them in your spare time?

src/fheroes2/spell/spell_info.cpp
Kazhuu: I changed the function to return possible boat position as int32_t and if it's not found simply return -1 as a result. Reduces code coupling and makes the function simpler to use as well.

Kazhuu:
Hello @ihhub I made the changes. Very good comments. The code should now be simpler as well. Can you review again?

src/fheroes2/heroes/heroes_spell.cpp
```diff
@@ -303,38 +296,27 @@ namespace
...
         const int32_t boatDestination = fheroes2::getPossibleBoatPosition( hero );
         assert( Maps::isValidAbsIndex( boatDestination ) );
 
-        for ( const int32_t boatSource : Maps::GetObjectPositions( center, MP2::OBJ_BOAT, false ) ) {
-            assert( Maps::isValidAbsIndex( boatSource ) );
-
-            Maps::Tiles & tileSource = world.GetTiles( boatSource );
-            const int boatColor = tileSource.getBoatOwnerColor();
-            const int heroColor = hero.GetColor();
+        int32_t boatSource = fheroes2::getSummonableBoat( hero );
```
ihhub, line 302: This variable can be **const** as we don't modify it after. SonarCloud actually detects this but only for PRs from this main repo.

src/fheroes2/spell/spell_info.cpp
```diff
@@ -348,14 +350,42 @@ namespace fheroes2
...
+    int32_t getSummonableBoat( const Heroes & hero )
+    {
+        const int32_t center = hero.GetIndex();
+
+        for ( const int32_t boatSource : Maps::GetObjectPositions( center, MP2::OBJ_BOAT, false ) ) {
+            assert( Maps::isValidAbsIndex( boatSource ) );
+
+            Maps::Tiles & tileSource = world.GetTiles( boatSource );
+            const int boatColor = tileSource.getBoatOwnerColor();
+            const int heroColor = hero.GetColor();
```
ihhub, line 372: This variable can be declared outside the loop. Since it is not inlined we will save some time doing this.

Kazhuu:
Hello @ihhub. Fixed the issues. Can you look again?

Branikolog:
I've made a couple of tests and graying seems to work correctly.
The only thing I'm concerning is wording in a case I summoned a few boats and there're no more available:

https://github.com/ihhub/fheroes2/assets/55348946/5d537cab-0a7b-44fa-9265-560e250875f0

I'm a little bit confused, as the message says, that no boat is available. Actually there're two boats at least available (which I summoned already) but this phrase makes an impression, that no boats can be summoned at all. I believe, this message is actual only for the case, when there're actually no free boats at all which could be summoned.
Maybe, if it would be spelled like: "There's no _more_ boat available to cast this spell.", it would make more clear, that I have available boats, but I simply can't summon more to the current location. How do you think?
Sorry, if I just understood this message incorrectly. :)

Kazhuu:
Good point @Branikolog. I changed to text that you suggested. What you think @LeHerosInconnu?

Kazhuu:
Sorry for the late answer. So what is the text that should be displayed and in what conditions @LeHerosInconnu and @Branikolog? It would be easier for me to fix this if stated clearly :)

Branikolog:
Hi, @Kazhuu !

For the case when there's no any boats available to summon, game should show the message: "There's no boat available to cast this spell."
For the case, when hero has already summoned 1 or more boats, which are standing empty nearby: "There's no more boat available to cast this spell."

Kazhuu:
Okay sounds good @Branikolog but how the text should be reflected on this message "There is no boat available and no free ocean location near the hero to cast this spell."? Should this text also have two variants? To be honest if so I think logic wise this gets quite complicated to implement and the benefit is only little bit better wording for the text.

LeHerosInconnu:
Hello @Kazhuu and @Branikolog,

> Hi, @Kazhuu !
> 
> For the case when there's no any boats available to summon, game should show the message: "There's no boat available to cast this spell." For the case, when hero has already summoned 1 or more boats, which are standing empty nearby: "There's no more boat available to cast this spell."

> Okay sounds good @Branikolog but how the new change should be reflected on this message "There is no boat available and no free ocean location near the hero to cast this spell."? Should this text also have two variants? To be honest if so I think logic wise this gets quite complicated to implement and the benefit is only little bit better wording for the text.

If this becomes too complicated to manage, for me it's okay to have only the sentence "There is no boat available to cast this spell." as it can encompass both cases and is "acceptable" for the description.
@Branikolog, what do you think?

Kazhuu:
I'd say it's not possible to implement, but I myself would like to keep the text more simple :+1:

Branikolog:
@Kazhuu 

Okay then. If it really doesn't worth overcomplicating then I'm OK with initial implementation without adding "more".

I'll probably create an issue for future improvement.

Branikolog (approved):

Kazhuu:
Thanks @Branikolog! Yeah another issue is good as well. I might work on it after a while, if nobody else takes it :+1:

ihhub:
We will merge this pull request for 1.0.6.

ihhub:
@Kazhuu , thanks a lot for these improvements!

PR #7022, opened 2023-04-16
Expand wide screen support

ihhub:
Add a special resolution for devices to support wide screens which are prevalent in modern devices. Only for SDL 2. The newly added resolution is extremely useful for mobile phones where we have to keep resolution as low as possible but avoid the game being rendered only on half of the phone's screen. This resolution is set by default for the first application run.

image

relates to #5373

I am open for suggestions, corrections how to improve user experience in this field.

### DIFF
--- a/src/fheroes2/game/fheroes2.cpp
+++ b/src/fheroes2/game/fheroes2.cpp
@@ -28,6 +28,7 @@
 #include <memory>
 #include <set>
 #include <string>
+#include <vector>
 
 #include <SDL_events.h>
 #include <SDL_main.h> // IWYU pragma: keep
--- a/src/fheroes2/game/fheroes2.cpp
+++ b/src/fheroes2/game/fheroes2.cpp
@@ -129,8 +129,23 @@ namespace
             const Settings & conf = Settings::Get();
 
             fheroes2::Display & display = fheroes2::Display::instance();
+            if ( conf.isFirstGameRun() && System::isHandheldDevice() ) {
+                // We do not show resolution dialog for first run on handheld devices. In this case it is wise to set 'widest' resolution by default.
+                const std::vector<fheroes2::ResolutionInfo> resolutions = fheroes2::engine().getAvailableResolutions();
+                fheroes2::ResolutionInfo bestResolution{ conf.currentResolutionInfo() };
 
-            display.setResolution( conf.currentResolutionInfo() );
+                for ( const fheroes2::ResolutionInfo & info : resolutions ) {
+                    if ( info.gameWidth > bestResolution.gameWidth && info.gameHeight == bestResolution.gameHeight ) {
+                        bestResolution = info;
+                    }
+                }
+
+                display.setResolution( bestResolution );
+            }
+            else {
+                display.setResolution( conf.currentResolutionInfo() );
+            }
+
             display.fill( 0 ); // start from a black screen
 
             fheroes2::engine().setTitle( GetCaption() );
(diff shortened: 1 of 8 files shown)


src/fheroes2/game/fheroes2.cpp
```diff
@@ -129,8 +130,23 @@ namespace
...
+                const std::vector<fheroes2::ResolutionInfo> resolutions = fheroes2::engine().getAvailableResolutions();
+                fheroes2::ResolutionInfo bestResolution{ conf.currentResolutionInfo() };
+
+                for ( const fheroes2::ResolutionInfo & info : resolutions ) {
+                    if ( info.gameWidth > bestResolution.gameWidth && info.gameHeight == bestResolution.gameHeight ) {
+                        bestResolution = info;
+                    }
+                }
+
+                display.setResolution( bestResolution );
```
oleg-derevenetz, line 144: I have an impression that `bestResolution` will not be saved in the `fheroes2.cfg` after the first run (because `ReadConfigs()` which creates a default config is called earlier) and you will get a different resolution on the second run.

src/fheroes2/system/settings.cpp
```diff
@@ -255,24 +255,33 @@ bool Settings::Read( const std::string & filePath )
         size_t pos = value.find( 'x' );
 
         if ( pos != std::string::npos ) {
-            _resolutionInfo.width = GetInt( value.substr( 0, pos ) );
+            _resolutionInfo.gameWidth = GetInt( value.substr( 0, pos ) );
```
oleg-derevenetz, line 258: What do you think about using a `regex` here? Something like [this](https://github.com/ihhub/fheroes2/blob/ca2920e152bdaf4c6ee1feefdbd53698a4cb8d5f/src/engine/tinyconfig.cpp#L103-L139).

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions, could you please take a look?

Districh-ru (commented):
Hi @ihhub, I tested this PR on Redmi 6 (Android 8) and Samsung Galaxy Tab 4 (Android 10) and all is OK. The best resolution is set on the first run.
I have only one suggestion (possibly for the next PR), could you please take a look:
Maybe we can make other widescreen resolutions for the given array of 4x3 resolution heights.
For example we can make 960x600, 1152x720 in addition to 768x480.

 
 And if we are talking not only about handheld devices I would be very glad to have 1366x768 (x1.4) for my desktop (1092x1080). :)

src/fheroes2/system/settings.cpp
oleg-derevenetz: Implemented this in commit [#2398233](https://github.com/ihhub/fheroes2/pull/7022/commits/2398233a4ed90cd120da8b5552e1a4936d841ec7), please take a look.

src/fheroes2/game/fheroes2.cpp
ihhub: Hi @oleg-derevenetz , **FirstGameRun** flag of Settings is reset in `fheroes2::GameMode Game::MainMenu( bool isFirstGameRun )` function. This function is called after the highlighted by you code which means that we save the settings.

ihhub:
Hi @Districh-ru , adding new non-standard resolutions is very tricky area. I suggest to do this after testing changes from this pull request and see if it works fine for other devices.

Districh-ru (approved):

oleg-derevenetz (approved):

PR #7048, opened 2023-04-22
implementing docker image for linux builder

juchem:
> from the included README.md:

A self-contained, sandboxed docker image that will build and install the
[fheroes2](https://github.com/ihhub/fheroes2) game engine, without
the need to setup a developer environment, install unwanted packages in
your machine or clutter it with build artifacts.

**TL;DR**: build and install latest version under `/usr/local`:
```
docker run -it --rm -v /usr/local:/out bstlang/fheroes2-linux-builder
```

The source code for this image can be found at
[ihhub/fheroes2](https://github.com/ihhub/fheroes2/tree/master/docker/builder/linux).

A pre-built image can be found in [docker
hub](https://hub.docker.com/r/bstlang/fheroes2-linux-builder). Note that when using
the pre-built image there's **no need** to clone this repository. The
repository is only needed when building the image itself.

Binaries will be installed into guest directory `/out`. Mount that directory
with `-v host_dir:/out` to install it into some host directory.

Choose the version to build by setting environment variable `VERSION` to the
appropriate [revision tag](https://github.com/ihhub/fheroes2/releases)
(defaults to `HEAD` for bleeding edge).

Example: install fheroes2 `1.0.3` under `~/opt` with (change `OUT_DIR` accordingly):
```
OUT_DIR="$HOME/opt"
VERSION=1.0.3
mkdir -p "${OUT_DIR}"
docker run -it --rm -v "${OUT_DIR}:/out" -e "VERSION=${VERSION}" bstlang/fheroes2-linux-builder
unset OUT_DIR
```


ihhub:
Hi @juchem , could you please explain the real benefit of doing this? I am fully aware what Docker is and what its purpose is but I would like to understand possible advantages of using it over the native application compilation except installing extra packages as make and g++ (cmake is not needed at all)? Also why debian was chosen? There are other flavours of Linux which have much smaller size like Alpine Linux. What's the size of the final Docker image? Will binaries work for all flavors of Linux?

juchem:
Hi there @ihhub. I generally create images like these to build software from the internet, so I don't have to install unwanted packages in my system. `fheroes2` requires more than just `g++` and `make` if built with all dependencies. For instance, `SDL2`, and `libpng` is included in the aforementioned image. A self-contained image keeps things clean



































. I understand that for regular maintainers and contributors of the engine, this doesn't matter at all.

The benefits are not universal. In the very least, one would need `docker` installed, so it becomes yet another build option. If one does have `docker`, though, building and installing from source becomes rather trivial: it's a one liner `docker run` command that's listed in the `README.md` (a pre-built image is currently hosted on my `hub.docker.com` repo, but if this is accepted I'd expect the `fheroes2` community to create its own repo (free on `hub.docker.com`) for authenticity/security reasons).

Some potential benefits include:
- the build process becomes 100% reproducible regardless of the user's environment;
- a self-contained image for simplified building/testing, on or off CI;
- the basis for a reproducible development environment: if one wants to start hacking away at `fheroes2`, it lowers the upfront investment and time to first contribution by not having to set up a build environment (a linux devenv can be set up this way in other OSes as well) - in my experience docker devenvs have proven priceless in many o the professional projects I've worked on so far - the source dir can either be mounted on the host or on a volume for persistence across container instances;
- taking inspiration from this, similar images can be created for other OSes like Mac and Windows.

As far as the image size, I don't believe it matters as much given that building, in most cases, is a one and done deal. Note that the `slim` image flavor is being used, which optimizes for image size. If minor gains in image size ends up mattering, changing the image base is also something that can be easily achieved by modifying a couple of lines in the `Dockerfile`.

As for portability, as long as the dynamic libraries are compatible (specially `glibc` version), it should work on other linux distros just fine.

Given that I already took the time to create it, I decided to offer it back to the community. If there's no interest in this PR, I'd understand it completely.

Other than that, thank you for all the hard work put on this amazing project.

ihhub:
Hi @juchem ,

I still don't see any good value of doing this. The compilation of the project is very simple and straightforward and all developers just install or already have installed packages (gcc and make) on their machines. While Docker is a powerful tool, using it for compilation of an application with almost no dependencies is an overengineering solution by adding an extra step just for the sake of step.

juchem:
that's very reasonable. thank you for the consideration

PR #7050, opened 2023-04-22
Slovak translation string pointing change

Roxterat:
Correct pointing to array positions - credits to Granfalegion


Roxterat:
@zenseii this is the reason why I think it doesnt work.

ihhub (approved):

Roxterat:
> > > Thanks for this fix. As a side note I suggest putting credits in the PR opening statement and not in the PR title.
> > 
> > 
> > Can I edit PR's opening statement from here? Tried to google for solution but could only find how to change state, files and other PR's characteristics.
> 
> Yes, you just have to click the three dots in the upper right corner of the statement.

If the opening statement is the very first comment in the PR then I have done that..

ihhub (approved):

ihhub:
@Roxterat , thanks a lot for this fix!

PR #7051, opened 2023-04-22
Rework the search algorithm for available scaled resolutions

oleg-derevenetz:
Related to #7046

Use the reverse logic to search the available scaled resolutions: do not search for bigger resolutions for existing smaller ones (640x480 -> 1280x960) but add lower "pseudo-resolutions" for existing bigger resolutions instead (1280x1024 -> 640x512x2).

resolutions

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -134,7 +134,7 @@ namespace
             const fheroes2::ResolutionInfo lowestResolution = *( resolutions.begin() );
 
             for ( const fheroes2::ResolutionInfo & resolution : possibleResolutions ) {
-                if ( lowestResolution < resolution || lowestResolution == resolution ) {
+                if ( lowestResolution.gameWidth < resolution.gameWidth || lowestResolution.gameHeight < resolution.gameHeight || lowestResolution == resolution ) {
                     continue;
                 }
 
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -108,86 +108,80 @@ namespace
             return { { fheroes2::Display::DEFAULT_WIDTH, fheroes2::Display::DEFAULT_HEIGHT } };
         }
 
-        std::vector<fheroes2::ResolutionInfo> resolutions;
-        resolutions.reserve( resolutionSet.size() );
-        for ( const auto & resolution : resolutionSet ) {
+        std::set<fheroes2::ResolutionInfo> resolutions;
+
+        for ( const fheroes2::ResolutionInfo & resolution : resolutionSet ) {
             if ( IsLowerThanDefaultRes( resolution ) ) {
                 continue;
             }
 
-            resolutions.emplace_back( resolution );
+            resolutions.emplace( resolution );
         }
 
         if ( resolutions.empty() ) {
             return { { fheroes2::Display::DEFAULT_WIDTH, fheroes2::Display::DEFAULT_HEIGHT } };
         }
 
-        // TODO: add resolutions which are close to the current screen aspect ratio.
-        // Some operating systems do not work well with SDL so they return very limited number of high resolutions.
-        // Populate missing resolutions into the list.
-        const std::set<fheroes2::ResolutionInfo> possibleResolutions
-            = { { 640, 480 },   { 800, 600 },  { 1024, 768 },  { 1152, 864 }, { 1280, 600 }, { 1280, 720 },  { 1280, 768 }, { 1280, 960 },
-                { 1280, 1024 }, { 1360, 768 }, { 1400, 1050 }, { 1440, 900 }, { 1600, 900 }, { 1680, 1050 }, { 1920, 1080 } };
+        {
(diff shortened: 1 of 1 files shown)


Districh-ru (changes requested):
Hi, @oleg-derevenetz.
I left one comment. Could you check it when you have time.

src/engine/screen.cpp
```diff
@@ -108,86 +108,80 @@ namespace
...
-        const fheroes2::ResolutionInfo lowestResolution = resolutions.front();
-        assert( *std::min_element( resolutions.begin(), resolutions.end() ) == resolutions.front() );
+            assert( !resolutions.empty() );
+            const fheroes2::ResolutionInfo lowestResolution = *( resolutions.begin() );
 
-        for ( const fheroes2::ResolutionInfo & resolution : possibleResolutions ) {
-            if ( lowestResolution.gameWidth < resolution.gameWidth || lowestResolution.gameHeight < resolution.gameHeight || resolution == lowestResolution ) {
-                continue;
+            for ( const fheroes2::ResolutionInfo & resolution : possibleResolutions ) {
+                if ( lowestResolution < resolution || lowestResolution == resolution ) {
```
Districh-ru, line 137: `lowestResolution < resolution` returns true only if all elements of `resolution` are higher then ones of `lowestResolution`.
If only one parameter (i.e. height) is higher and the other (width) is not higher it will return `false`.
Maybe this condition can be replaced with `( !( lowestResolution > resolution ) )`.

So on my handheld devices I've got resolutions with initially unsupported height. For example my phone's screen resolution is 1440x720 and I get:



<details><summary>On master build I don't have these resolutions:</summary>



</details>

Districh-ru (changes requested):

src/engine/screen.cpp
oleg-derevenetz: Yes, these resolutions are clearly superfluous here. Now I use the logic from `master` here.

oleg-derevenetz: > `lowestResolution` < `resolution` returns true only if all elements of `resolution` are higher then ones of `lowestResolution`.

That's not quite true, `width` and `height` are compared in order using the short-circuit logic, so if `res1.width` is less than `res2.width`, then `res1.height` is not compared with `res2.height` at all.

ihhub:
Hi @oleg-derevenetz , my biggest fear not implementing this in the past is the amount of resolutions available for end users. Having tons of them seems like an overkill.

The biggest problem is the way how we add resolutions in case of having a single resolution returned by SDL. Logically we should add resolutions which are close to the aspect ration of the original one. For instance, having 1920x1080 and adding 1280x960 doesn't look right for some people. I don't have a strong opinion about the best solution we should choose at the moment.

oleg-derevenetz:
@ihhub

> my biggest fear not implementing this in the past is the amount of resolutions available for end users. Having tons of them seems like an overkill.

Well, it's up to you :) But IMHO having more scaled resolutions is fine, because existing resolutions usually doesn't fit the scaling well.

> The biggest problem is the way how we add resolutions in case of having a single resolution returned by SDL. Logically we should add resolutions which are close to the aspect ration of the original one. For instance, having 1920x1080 and adding 1280x960 doesn't look right for some people. I don't have a strong opinion about the best solution we should choose at the moment.

This logic left intact in this PR (apart from incorrect optimization of resolution comparison which is reverted).

src/engine/screen.cpp
Districh-ru: Thanks for the explanation.

Districh-ru (approved):
Now it works OK. 👍
On my cell phone I have only one scaled resolution (x1.5) as before this PR. On PC I have some new (x2.0) resolutions.

IMHO it will be better in future to set only the game resolution and have an option to use desktop resolution as screen resolution with scaling.
Talking about PC for me there is no difference - run the game in fullscreen using `640x480` or `640x480(x2.0)` - desktop resolution will be changed, all windows and icons will be rearranged. The only difference is that in first place there only will be used the video card or monitor built-in scaler. And running the game in a window I can always change the window size (i.e. maximize it) so (x1.0) and (x2.0) will give the same picture.
And, as I can see, on android the game always use native screen resolution and stretches the picture to fit it, keeping the aspect ratio. So there will be no difference between (x1.0) and other scales for the same resolution.

oleg-derevenetz:
> IMHO it will be better in future to set only the game resolution and have an option to use desktop resolution as screen resolution with scaling.

Technically there is no problem with this - all we need is to always use the `SDL_WINDOW_FULLSCREEN_DESKTOP` fullscreen mode (and always use the software cursor to properly scale it). Currently this mode is used almost everywhere except Windows. This mode is better than the "hardware" fullscreen mode (`SDL_WINDOW_FULLSCREEN`) in many aspects - for example, it properly works with multiple monitors and it doesn't cause negative effects such as rearranging shortcut icons on the desktop.

ihhub (approved):

ihhub:

Issue #6837, opened 2023-03-14
Adventure screen, list of castles should not be adjusted when a hero moves


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the list of castles should not be adjusted when a hero moves.

In fheroes2:

https://user-images.githubusercontent.com/43583994/225036557-73a1173b-65de-43ed-bd4e-80333bc290bf.mp4

**Edit 2023/03/24.**

There is the same problem with the list of heroes.

In fheroes2:

https://user-images.githubusercontent.com/43583994/227537275-750af327-17fe-4e8d-8407-0de6b22cace3.mp4

I can open another issue for this if necessary.

### Save file

Not needed.
Here is the save file (list of heroes):
[Hero list 01.zip](https://github.com/ihhub/fheroes2/files/11062696/Hero.list.01.zip)

autoexecbatman:
https://github.com/ihhub/fheroes2/blob/master/src/fheroes2/gui/interface_list.h

It seems this issue is happening in the template class ListBox
Line 263 is affecting the hero list scrollbar.
Could not yet find a solution. Maybe someone will find this useful

ihhub:
Hi @autoexecbatman , I think shouldn't even refresh the castle's list in first place so no scrollbars would be updated.

autoexecbatman:
Inside the kingdom overview window .

LeHerosInconnu:
Hello @Kazhuu,

> I could take a stab at this. I tested on original game and the scrollbars doesn't update at all. I guess the point to update the heroes scrollbar is to bring the currently moving hero to the view. The game seems to scroll the currently moving hero to the bottom of the scrollbar view.
> 
> And yes the castle scrollbar shouldn't be updated when hero moves, but what about the hero scrollbar? I think it's nice feature that it automatically scrolls to the hero that is moving. Now it seems to always scroll the hero to the bottom of the view. I guess one improvement would be not to auto scroll if the moving hero is already in the view?

It is better to let the user set the position of the heroes list at his convenience, for whatever reason, and not to "automatically" change the position of the heroes list.
[...]

PR #7053, opened 2023-04-23
Do not reset castle or heroes scrollbars when hero moves

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/6837. Now both castles and heroes scrollbars positions are retained when hero moves. Previously both of them were modified. Castle scrollbar always goes to the top and heroes scrollbar is either placed to the top or bottom of the scrollbar view depending on the hero's position on the list.

### DIFF
--- a/src/fheroes2/game/game_interface.h
+++ b/src/fheroes2/game/game_interface.h
@@ -179,9 +179,10 @@ namespace Interface
             return controlPanel;
         }
 
+        void SetFocus( Heroes *, const bool retainScrollBarPosition );
         void SetFocus( Heroes * );
         void SetFocus( Castle * );
-        void ResetFocus( const int priority, const bool resetScrollBarPosition );
+        void ResetFocus( const int priority, const bool retainScrollBarPosition );
         void RedrawFocus();
         void updateFocus();
 
--- a/src/fheroes2/game/game_interface.h
+++ b/src/fheroes2/game/game_interface.h
@@ -181,7 +181,7 @@ namespace Interface
 
         void SetFocus( Heroes * );
         void SetFocus( Castle * );
-        void ResetFocus( int priority, const bool resetScrollBarPosition = true );
+        void ResetFocus( const int priority, const bool resetScrollBarPosition );
         void RedrawFocus();
         void updateFocus();
 
--- a/src/fheroes2/game/game_interface.h
+++ b/src/fheroes2/game/game_interface.h
@@ -181,7 +181,7 @@ namespace Interface
 
         void SetFocus( Heroes * );
         void SetFocus( Castle * );
-        void ResetFocus( int );
+        void ResetFocus( int priority, const bool resetScrollBarPosition = true );
         void RedrawFocus();
         void updateFocus();
 
--- a/src/fheroes2/game/game_interface.h
+++ b/src/fheroes2/game/game_interface.h
@@ -180,7 +180,6 @@ namespace Interface
         }
 
         void SetFocus( Heroes *, const bool retainScrollBarPosition );
-        void SetFocus( Heroes * );
         void SetFocus( Castle * );
         void ResetFocus( const int priority, const bool retainScrollBarPosition );
         void RedrawFocus();
--- a/src/fheroes2/gui/interface_focus.cpp
+++ b/src/fheroes2/gui/interface_focus.cpp
@@ -155,8 +155,12 @@ void Interface::Basic::ResetFocus( const int priority, const bool retainScrollBa
     case GameFocus::HEROES:
         if ( focus.GetHeroes() && focus.GetHeroes()->GetColor() == player->GetColor() )
             SetFocus( focus.GetHeroes(), retainScrollBarPosition );
-        else if ( !myKingdom.GetHeroes().empty() )
+        else if ( !myKingdom.GetHeroes().empty() ) {
+            // Reset scrollbar here because the current focused hero might have
+            // lost a battle and is not in the heroes list anymore.
+            iconsPanel.ResetIcons( ICON_HEROES );
             SetFocus( myKingdom.GetHeroes().front(), false );
+        }
         else if ( !myKingdom.GetCastles().empty() ) {
             iconsPanel.SetRedraw( ICON_HEROES );
             SetFocus( myKingdom.GetCastles().front() );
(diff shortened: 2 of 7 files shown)


src/fheroes2/game/game_interface.h
```diff
@@ -181,7 +181,7 @@ namespace Interface
 
         void SetFocus( Heroes * );
         void SetFocus( Castle * );
-        void ResetFocus( int );
+        void ResetFocus( int priority, const bool resetScrollBarPosition = true );
```
ihhub, line 184: Could you please make **priority** input argument as **const** and also remove the default value for the second argument? I know that this way is easier to make changes but default parameters usually lead to difficulties to debug code.

src/fheroes2/gui/interface_focus.cpp
```diff
@@ -122,15 +121,17 @@ void Interface::Basic::updateFocus()
     }
 }
 
-void Interface::Basic::ResetFocus( int priority )
+void Interface::Basic::ResetFocus( int priority, const bool resetScrollBarPosition /* = true */ )
 {
     Player * player = Settings::Get().GetPlayers().GetCurrent();
 
     if ( player ) {
```
ihhub, line 128: Since we modify his function could we please do the following:
```cpp
if ( player == nullptr ) {
    return;
}
...
// the rest of the code.
```
?

This would definitely increase code readability as having many if-s and switch cases is hard to read.

ihhub (changes requested):
Hi @Kazhuu , I put multiple suggestions here. Could you please check them and see if they make sense?

src/fheroes2/game/game_interface.h
Kazhuu: Indeed!

ihhub (approved):

Branikolog:
Hi, @Kazhuu 
I've found that once we change focus from castle to a hero, castle list updates:

https://user-images.githubusercontent.com/55348946/234865836-2145b877-d4ff-430e-a71a-2d259de4e68f.mp4

After battle, castle list is also updated in the current PR build.
Castle list shouldn't be updated during single turn, no matter what hero we choose or what action we make.

Kazhuu:
Hello @Branikolog I fixed the issue. The function that needed to be changed is also used elsewhere and what I tested manually everything seems to function as it should. Can you test again?

src/fheroes2/gui/interface_focus.cpp
```diff
@@ -221,11 +223,7 @@ void Interface::Basic::RedrawFocus()
         iconsPanel.SetRedraw();
     }
 
-    if ( type != FOCUS_CASTLE && iconsPanel.IsSelected( ICON_CASTLES ) ) {
-        iconsPanel.ResetIcons( ICON_CASTLES );
-        iconsPanel.SetRedraw();
-    }
-    else if ( type == FOCUS_CASTLE && !iconsPanel.IsSelected( ICON_CASTLES ) ) {
```
Kazhuu, line 228: This code was responsible for updating castle scrollbar back to the top. There is similar code for the heroes scrollbar but after debugging for a while it seems that code block is not executed like this one is when changing focus between castle and heroes lists. Now sure why. But removing these lines fix the issue and what I tested everything seems to function like it should.

LeHerosInconnu:
Hello @Kazhuu,

> Hello @LeHerosInconnu I did some changes. Also maybe it would be better to post this in the PR where the discussion is now happening to make it easier for people to follow :)
> 
> I noticed one issue that seems to exits before this change as well. When you double click the hero and press next hero in the down left or right corner and then exit. The focus is retained on the hero you double clicked before, not the one you press next to. This however works with castles. Maybe if this need to be fixed, create another issue about this?

_Originally posted by @Kazhuu in https://github.com/ihhub/fheroes2/issues/6837#issuecomment-1528664542_

In the original game, switching heroes or castles by accessing the hero or castle screen, or by accessing the hero or castle screen through one of the Kingdom Overview screens, does not change the hero or castle selected in the list of heroes or castles when exiting those screens.
The user may simply want to check something without changing his initial selection.
It should work the same way in fheroes2.

I tested this, everything seems to work as it should, except for the point described above where the castle should always remain selected in the castle list.
I can open a new issue for that if necessary.

Branikolog:
Hello, @Kazhuu !
I've found another issue. When hero uses Dimension Door spell both lists reset.

https://user-images.githubusercontent.com/55348946/235302267-c049b946-06b5-4c7d-905f-7c7e1c78ab9b.mp4

The same happens with Town Portal and Town Gate spells.

@LeHerosInconnu 

>In the original game, switching heroes or castles by accessing the hero or castle screen, or by accessing the hero or castle screen through one of the Kingdom Overview screens, does not change the hero or castle selected in the list of heroes or castles when exiting those screens.
>The user may simply want to check something without changing his initial selection.

At the same time user may want to scroll through castles and when he finds the one, he can exit and check its surroundings. Or scroll through castles till he finds the one with hero, to exit and instantly move the hero from this castle? But anyway, the logic should be consistent for all cases at least. Either not refocus on heroes/towns scrolling by small corner buttons, or focus on the town/hero that was the last shown on the screen.

Kazhuu:
Good find @Branikolog. Could you also provide me a save file so I can test this more easily :+1:

Branikolog:
@Kazhuu 
Sure.
[Scrollbar test DD+TP.zip](https://github.com/ihhub/fheroes2/files/11359242/Scrollbar.test.DD%2BTP.zip)
Check hero Wrathmont, he has DD and TP spells in his book.

Branikolog:
@Kazhuu 
Here's one more issue:

https://user-images.githubusercontent.com/55348946/235304779-8e3670de-2e39-4729-90fd-8c684a621537.mp4

Kazhuu:
@Branikolog I fixed the issues. Adventure spells like town portal now do not reset scrollbars and when hero loses a battle, the scrollbar is positioned correctly. Can you test one more time? And thanks for testing!

Branikolog:
Hi, @Kazhuu !
Sorry for the late response!
I've found out that when you enter a castle with a hero by pressing spacebar, scrollbars are reset:

https://user-images.githubusercontent.com/55348946/236293972-d9a95c6b-2c14-4217-b6ac-4acb02d60f48.mov

The same thing happens, when we press spacebar to visit some objects on the adventure map. Like creature dwelling.

Also, when you dismiss hero both scrollbars a reset to upper position. 

https://user-images.githubusercontent.com/55348946/236294568-e7ef19b0-9e58-46f8-ae54-9a5fec98b148.mov

For heroes list it's ok to be reset, since the first hero is selected, but I think castle list should keep its position unchanged.
I think, the last issue is a bit out of current PR, since the PR is supposed to fix issues with scrollbars for hero movement. If you wish, I can create a separate issue, which could be linked to some future PRs.

Kazhuu:
Hi @Branikolog I fixed the issue you described. Can you now test again?

Branikolog (approved):
I can't find any move-related issues here more.
Good work, @Kazhuu !
I'll create a separate issues relating hero dismiss and castle capturing later, where scrollbars reset too.

Kazhuu:
Thanks @Branikolog for testing. Could @ihhub or someone review this PR now?

oleg-derevenetz:
@Kazhuu thank you very much!

Issue #5259, opened 2022-04-11
Spell info window, text should be adjusted when the hero has zero ("0") mana


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

In fheroes2, the text should be adjusted when the hero has zero ("0") mana.

In fheroes2 (with "Town Portal" spell):

![0 mana 01](https://user-images.githubusercontent.com/43583994/162739256-ed912df4-1c4b-44d1-88f4-515bfa6ffbeb.png)

### Describe the possible solution

It could be:

"You have 0 mana, ..."

Or:

"You have no mana, ..."

Instead of:

"You only have 0 mana, ..."

### Additional info

Version 0.9.13 5171

PR #7054, opened 2023-04-23
Adust spell info message when hero has zero (0) mana

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/5259 by changing the message to "You have no mana..." when hero has 0 mana points.

image

### DIFF
--- a/src/fheroes2/heroes/heroes_base.cpp
+++ b/src/fheroes2/heroes/heroes_base.cpp
@@ -361,8 +361,17 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
     }
 
     if ( !HaveSpellPoints( spell ) ) {
-        if ( res ) {
+        const Heroes * hero = dynamic_cast<const Heroes *>( this );
+        assert( hero != nullptr );
+
+        if ( hero->GetSpellPoints() == 0 && res ) {
+            *res = _( "That spell costs %{mana} mana. You have no mana, so you can't cast the spell." );
+            StringReplace( *res, "%{mana}", spell.spellPoints( hero ) );
+        }
+        else if ( res ) {
             *res = _( "That spell costs %{mana} mana. You only have %{point} mana, so you can't cast the spell." );
+            StringReplace( *res, "%{mana}", spell.spellPoints( hero ) );
+            StringReplace( *res, "%{point}", hero->GetSpellPoints() );
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


src/fheroes2/heroes/heroes_base.cpp
```diff
@@ -361,8 +361,17 @@ bool HeroBase::CanCastSpell( const Spell & spell, std::string * res /* = nullptr
     }
 
     if ( !HaveSpellPoints( spell ) ) {
-        if ( res ) {
+        const Heroes * hero = dynamic_cast<const Heroes *>( this );
```
ihhub, line 364: **dynamic_cast** is usually considered a very last resort to be use so it is good to avoid using it (I know that we have few places where we use it as well). I am thinking that this method should accept one more parameter **availableSpellPoints** and then use it to update the message. What do you think?

Kazhuu: We already have one more `dynamic_cast` also beside this one in this function. The other cast would not be solved by adding another parameter here. How about we refactor this whole function out of `HeroBase` class and put it to `Heroes` class or somekind of support/helper class instead. Then just take `Heroes` instance as a parameter. Would this make more sense?

ihhub: If this method is not used by anything else except **Hero** then yes, we should move it there. Moreover, we can just make as a const method so no need to pass Heroes parameter.

Kazhuu: Now I checked why we even have this it seems both `Heroes` and `Captain` are inherited from this class. And it seems `Captain` instance can also call `CanCastSpell` function. So moving this to `Heroes` class is only feasible if we make this virtual in the base class and implement copy of this method to he `Captain` class. Or maybe something else, not sure. Could we have this one `dynamic_cast` here for now and think about the design of this on some other issue perhaps?

Kazhuu: @ihhub I actually tested and the dynamic cast here is not needed, so I removed it from these changes. There is still one dynamic cast in this function which would require more refactoring to get rid of. Hopefully the PR is now okay. Can you review again?

ihhub (approved):

ihhub:
Hi @LeHerosInconnu, could you please review this pull request?

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu, could you please review this pull request?

It works correctly.

ihhub:

PR #7060, opened 2023-04-25
Handle events during fade animations, sync rendering by "CustomAnimationDelay"

Districh-ru:
This PR adds
- events handling during fade animations (i.e. at the end of battle). This fixes soft rendered cursor freezes.
- fade animations synchronization by "CustomAnimationDelay" instead of `delayforMs()`.
Also a small code clean-up was made in this PR.

### DIFF
--- a/src/fheroes2/gui/ui_tool.cpp
+++ b/src/fheroes2/gui/ui_tool.cpp
@@ -81,7 +81,7 @@ namespace
             }
 
             double averageFps = 0;
-            for ( const double & value : _fps ) {
+            for ( const double value : _fps ) {
                 averageFps += value;
             }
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/gui/ui_tool.cpp
```diff
@@ -54,42 +54,45 @@ namespace
...
 
             _fps.push_front( fps );
-            while ( _fps.size() > 10 )
+            while ( _fps.size() > 10 ) {
                 _fps.pop_back();
+            }
 
             double averageFps = 0;
-            for ( const double value : _fps )
+            for ( const double & value : _fps ) {
```
ihhub, line 84: Please do **not** const references on POD (plain old data) types because their size is either equal or even small than a reference. For example, sizeof(uint8_t) is 1 while sizeof(uint8_t&) is 8 on 64-bit systems.

ihhub (changes requested):
Hi @Districh-ru , I left one correction here. Would you mind to check it?

ihhub:
@Districh-ru , huge thanks for these enhancements!

PR #7061, opened 2023-04-25
Fix Battlefield animations freeze while waiting for Good Luck sound to end

Districh-ru:
This PR fixes the freeze of Battlefield animations after finishing rainbow animation and until Good Lock sound is over.
This PR also removes waiting for Good Luck sound if sound volume is set to 0.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -4510,15 +4510,19 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
         const double drawStep = rainbowLength / 30.0;
 
         // Don't waste time waiting for Good Luck sound if the game sounds are turned off
-        if ( Settings::Get().SoundVolume() > 0 ) {
-            AudioManager::PlaySound( M82::GOODLUCK );
-        }
+        const bool soundOn = Settings::Get().SoundVolume() > 0;
+
+        AudioManager::PlaySound( M82::GOODLUCK );
 
         double x = 0;
-        while ( le.HandleEvents( Game::isDelayNeeded( { Game::BATTLE_MISSILE_DELAY } ) ) && ( Mixer::isPlaying( -1 ) || x < rainbowLength ) ) {
+        while ( le.HandleEvents( Game::isDelayNeeded( { Game::BATTLE_MISSILE_DELAY } ) ) && ( ( soundOn && Mixer::isPlaying( -1 ) ) || x < rainbowLength ) ) {
             CheckGlobalEvents( le );
 
             if ( Game::validateAnimationDelay( Game::BATTLE_MISSILE_DELAY ) ) {
+                // Reset all units idle animation delay during rainbow animation not to start new idle animations.
+                // This does not affect Zombies, Genies, Medusas and Ghosts as they have permanent idle animations.
+                ResetIdleTroopAnimation();
+
                 RedrawPartialStart();
 
                 if ( x < rainbowLength ) {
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @Districh-ru , could you please explain why not playing sound is applicable only for Luck? We have a lot of other sounds during battle which have the same issue.

Branikolog:
Hi, @Districh-ru 
I've tested this PR. 
Correct me, if I'm wrong. This PR adds idle animations after rainbow animation is finished, but still stays above the creature, till sound is over, right? During my midnight tests I've created a discussion, thinking I was testing master branch which has animations stopped. :D
I personally love, how animations are freezed for a while, as I described in #7107. So I'm not sure in this PR implementation of idle animations.
Speaking about no sound case - I think it's a valid improvement. But I actually can't see much difference in final rainbow stage lasting on 10 speed with sound off. On the lower speeds it is more noticeable. Did you set some minimal timings so there's a certain pause for the highest speeds?

Districh-ru:
Hi, @Branikolog!
Thanks for the review. 

> Correct me, if I'm wrong. This PR adds idle animations after rainbow animation is finished, but still stays above the creature, till sound is over, right?

Yes, you are right. Currently in master build while the rainbow is "going" from the edge of the screen to the creature all animations, including IDLE, are going and when the rainbow animation is finished the whole screen (hero's flag, Genies, Zombie's flies, ...) freezes.
This PR fixes this freeze.
I'll try to make all creatures except Genies, Ghosts and Zombies to stop their idle animation. :)


> Did you set some minimal timings so there's a certain pause for the highest speeds?

Currently we have only the time of Luck sound and if the Rainbow animation finishes earlier we freeze the screen and wait until the sound is over.
As the rainbow animation lasts for some time (also on the highest speed) so I don't think we need some extra time to make player notice that the Good Luck was applied.
Current master build:

https://user-images.githubusercontent.com/113276641/236485888-0fbfa012-0171-4b3c-9800-8022aff4c55a.mp4

This PR:

https://user-images.githubusercontent.com/113276641/236485949-794bbfbd-64cb-415d-9c07-353715acc2fd.mp4

Branikolog (approved):
Looks well. Works faster for no sound battle. No more freezes at all during battle.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -4505,20 +4505,35 @@ void Battle::Interface::RedrawActionLuck( const Unit & unit )
...
+        const int32_t rainbowDrawSteps = 30;
         // Rainbow animation draw step (in original game it is random and about 7-11 pixels).
         // We set the constant animation time for all rainbows: rainbowLength/30 fits the rainbow sound duration on '1' speed.
-        const double drawStep = rainbowLength / 30.0;
+        const double drawStep = static_cast<double>( rainbowLength ) / rainbowDrawSteps;
+
+        // Don't waste time waiting for Good Luck sound if the game sounds are turned off
+        const bool soundOn = Settings::Get().SoundVolume() > 0;
 
         AudioManager::PlaySound( M82::GOODLUCK );
```
ihhub, line 4516: Would it make sense not to play the sound if sound volume is off?

ihhub (commented):
Hi @Districh-ru , I left one question. Could you please take a look when you have time?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi, @ihhub.
I tried it first, but there may be (and the test say that there are) other sounds in the game and `Mixer::isPlaying( -1 )` will wait for them to finish.
Yes, we can not play this sound and it will not affect this function.

ihhub (approved):

ihhub:

PR #7062, opened 2023-04-25
Add screen fade effect when changing display image

Districh-ru:
This PR enables screen fade effect when:
- starting/loading a game;
- going to main menu;
- showing credits;
- showing the High Scores;
- opening/closing the World View.
- opening/closing dialogs (full fade effect is only in 640x480 resolution):
  - Castle,
  -  Hero,
  - Heroes Meeting
  -  Battle,
  -  Kingdom Summary,
  - Puzzle Map;
- quitting the game.

This PR also fixes twice rendered window borders and window shadow when opening Castle dialog from Kingdom Summary dialog window.


Districh-ru:
> Hi, @Districh-ru I've tested it for a few moments already. Fading animation needs to be corrected. Currently, it just looks like fade in, and at the darkest point next window is showing already rendered. Fading out animation is definitely missing. I think, current implementation suits exiting the application only. Fading in anmation seems to be lasting for too long. Even if it's going to be cut in a half and added fading out animation (as I proposed above) I'm still not sure, if it is not too long. This needs to be tested though. Also, there's no fading animation at all, while exiting castle, battle and hero window. Battle music starts too early: together with fading in animation. I suggest battle music start in the middle of fading (when fading out starts)

Hello, @Branikolog.
Many thanks for the proposals!
In this PR I initially just enabled already present in game fade-out animation.
The last commit introduces fade-in animation. So now the screen is turned to black with the current image and fades-in with the new image from the opened dialog.
The fade time was reduced to 100 ms for each fade: full fade-out -> fade-in takes 200 ms.
When resolution is not 640x480 the fade effect is also used, but only half of it as we do not need to darken the whole game screen. @Branikolog, what do you think about this? If this idea is not good, it can be reverted. :)
I have not added fade after battle and fade-in after castle and hero if yoг changed castle/hero in opened dialog as it is a little complicated.
So maybe it would be better polish the fade animation in this PR and add it later to the needed places in the next PR.

Branikolog:
Hi, @Districh-ru 
Okay, I will hold on my suggestions on the places, where fading could be applied. :)
Current implementation for 640x480 looks nice to me. It's a bit faster, than the original game. But in the OG the timings seem to be different for hero and castle windows. This could even be a performance issue. So I I'm ok with the current fade speed. I think, too slow fade can annoy players. 
Speaking about higher resolutions... It's acceptable, but I'm concerning about the moment, when frame is rendered with a black screen inside. Fade out only is a proper decision here and I like it, but the moment, when frame with a black screen inside appears for milliseconds looks weird for me. It's not a critical issue, actually. Probably, if the fade out will be a little faster... Like 70-80ms for this particular case this would not bother me so much. :)
Anyway, fading improves the aesthetic of the game for all resolutions. 👍

Districh-ru:
> Hi, @Districh-ru Okay, I will hold on my suggestions on the places, where fading could be applied. :) 

Hi, @Branikolog, you are welcome for suggestions on where we need to add the fade effect. The fade out effect is easy to apply, while for fade in we have to prepare the proper image. So in some places it is a little complicated that's why I suggest to add fade effects in separate PR:) 

> Speaking about higher resolutions... It's acceptable, but I'm concerning about the moment, when frame is rendered with a black screen inside. Fade out only is a proper decision here and I like it, but the moment, when frame with a black screen inside appears for milliseconds looks weird for me. It's not a critical issue, actually. Probably, if the fade out will be a little faster... Like 70-80ms for this particular case this would not bother me so much. :)

I made the fade effect when opening a hero/castle/battle in a window to be only half (to have no black screen inside) and it lasts for 50 ms.

Districh-ru:
> Hi, @Districh-ru. I've tested this on various resolutions and from what I can tell it is only in effect for higher than 640x480 resolutions when you are loading up the adventure map? It is much easier to notice when using 640x480 for me at least. It is as @Branikolog said quite a bit faster than the original effect, but I'm not sure this is a problem for players, so IMO we can move ahead with it according to @Branikolog's suggestions and see what feedback we get.

Hi, @zenseii.
Thanks for testing!
The fade effect for the resolutions, higher than 640x480 will be only when the whole screen image is changing.
When opening the dialog windows for castle/hero/battle there will be only half of fade effect: it is a compromise as we cannot fade-out the whole game image (it is bigger) and the fade-in effect starts from the half as rendering the black rectangle inside a dialog window is not so good. When closing castle/hero dialog there will be only half of fade-out effect.

I also increased the fade time from 100 to 150 ms. It is also fast and IMHO will not annoy the players. :)

I added the fade effects when going in/out the main menu. @Branikolog, @zenseii, could you please recheck the fade effect behavior in your spare time. ;)

Branikolog:
Hi, @Districh-ru !

> you are welcome for suggestions on where we need to add the fade effect.

Well, You've already added fade for all places I was going to point. :)
The only one is missing is for checking heroes in tavern. On exiting hero info in tavern it somehow works as intended. Fade is missing for opening hero info only.

First I loaded this build I've experienced a few slow downs in fade while opening hero or castle screens. But later couldn't reproduce the issue. Probably it's an isue of my PC...

Speaking about fade for higher resolutions. I like that faster fade. Cannot tell, that 100ms was bad, but 50 I feel a little better. But it's just my personal point. Probably if other team and players would vote for slower fade, I can agree on 75ms. And even 100 is acceptable. It's better than lack of fade anyway. 👍

Districh-ru:
Hi @Branikolog.
I finished adding fade effects to all the places where I think it would be good.

Branikolog:
Hi, @Districh-ru 
I've noticed that hero info window fade is missing during battle. The same as it was in tavern. Could you, please, check it?

Districh-ru:
> Hi, @Districh-ru I've noticed that hero info window fade is missing during battle. The same as it was in tavern. Could you, please, check it?

Hi, @Branikolog.
I fixed this in current commit.
...
And then opened the Kingdom view dialog. :) I think it also should have fade effects when opening hero/castle dialog.
Also the heroes meeting dialog does not have fade effect.
I'll try to fix this later today.

Districh-ru:
I found a little bug: If I change the troops in hero dialog opened from castle dialog then fade-in will be with unchanged troops and after that they will be changed. I need to do fade-in only after first pre-render of new (changed) image.
I'll fix it tomorrow.

Districh-ru:
Hi @Branikolog.
I made the fixes. I want to believe that it was the last noticeable rendering bug. :)
Could you please check the fade effect in 640x480 and in the other (higher) resolution when you have time.

Districh-ru:
> I want to believe that it was the last noticeable rendering bug. :)

Not, it saw not. ;)
I slightly polished the fade animation to avoid rendering frames with already rendered fade effect. This should smooth out the fade animation a bit.

Branikolog:
Hello, @Districh-ru 
I've tested PR with the latest changes.
I'm still concerning on adding so slow fade-in and fade-out for such windows as hero info in battle and castle (for visiting hero) where we can watch for a few moment completely black window (over the adventure map background) while switching between hero and castle or hero and battle. But it's my subjective point... I personally expect higher fade speeds, same to battle, hero and castle windows appearing on the adventure map screen.

But the second place I want to point is hero meeting dialog, where we watch fade-in, then new larger window appears blacked and then fade-out proceeds. 

https://user-images.githubusercontent.com/55348946/236319467-1fc677a7-7105-450b-8d7b-0ce2c4d2c8df.mov

It looks quite raw for me. How do you think, if we won't make fade-in for the hero meeting dialog and just draw hero info with fade-out only. Both windows have different size so for me it looks like hero info appears over hero meeting window (fade-in should be applied for the hero info window too, in a case we close it)?

Districh-ru:
Hello, @Branikolog, thanks tor testing.
>I'm still concerning on adding so slow fade-in and fade-out for such windows as hero info in battle and castle (for visiting hero) where we can watch for a few moment completely black window (over the adventure map background) while switching between hero and castle or hero and battle. But it's my subjective point... I personally expect higher fade speeds, same to battle, hero and castle windows appearing on the adventure map screen.

Sure, I'll revert to 2x100 ms for full fade effects.

> But the second place I want to point is hero meeting dialog, where we watch fade-in, then new larger window appears blacked and then fade-out proceeds.
> It looks quite raw for me. How do you think, if we won't make fade-in for the hero meeting dialog and just draw hero info with fade-out only. Both windows have different size so for me it looks like hero info appears over hero meeting window (fade-in should be applied for the hero info window too, in a case we close it)?

I'll try to make the half-fade effects in this case or remove fade al all.
The hero meeting dialog needs to be redone in the future as it has window borders but these are not the game theme borders (Good/Evil). So it needs a special approach. :)

Districh-ru:
Hi, @Branikolog.
I applied the fade effect improvements, discussed in Discord: use half fade when changing dialog in a window, add fade effects to Campaign scenario info and to View World. Could you check it in your spare time.

ihhub:
Hi @Districh-ru , could you please fix merge conflicts? I believe we need to merge logic from your separate PR related to cancelling pre-battle sound.

src/fheroes2/kingdom/kingdom_overview.cpp
```diff
@@ -771,7 +796,22 @@ void Kingdom::openOverviewDialog()
...
+        if ( !isDefaultScreenSize ) {
+            // We need to expand the ROI for the next render to properly render window borders and shadow.
+            fheroes2::Rect roi( background.windowArea() );
+            roi.x -= BORDERWIDTH;
+            roi.width += BORDERWIDTH;
+            roi.height += BORDERWIDTH;
+            display.updateNextRenderRoi( roi );
+        }
+
+        fheroes2::fadeInDisplay( background.activeArea(), !isDefaultScreenSize );
```
ihhub, line 810: This chunk of code is duplicated 2 more times. I suggest to create a special function for it.

src/fheroes2/system/settings.h
```diff
@@ -190,9 +190,8 @@ class Settings
 
     static bool isFadeEffectEnabled()
     {
-        // TODO: fix fading effect for the original resolution (640 x 480) and enable back this option.
-        // return video_mode == fheroes2::Size( fheroes2::Display::DEFAULT_WIDTH, fheroes2::Display::DEFAULT_HEIGHT );
-        return false;
+        // TODO: Store fade effect setting in configuration file or remove this function.
```
ihhub, line 193: This should be addressed within the current pull request. If we are unsure about the proposed changes then we can store it in a configuration file. However, I prefer to not to introduce any new configuration settings.

ihhub (changes requested):
Hi @Districh-ru , I quickly checked this pull request and put few comments here. Could you please check them? I will do a comprehensive review again once I have more free time.

src/fheroes2/battle/battle_interface.h
```diff
@@ -289,6 +289,12 @@ namespace Battle
             return _surfaceInnerArea;
         }
 
+        // Battlefield interface ROI.
+        const fheroes2::Rect & GetIntrfaceRoi() const
```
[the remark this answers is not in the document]
Districh-ru: Hi, @ihhub.
Thanks for finding this typo!

src/fheroes2/castle/castle.h
```diff
@@ -229,7 +229,7 @@ class Castle : public MapPosition, public BitModes, public ColorBase, public Con
 
     void DrawImageCastle( const fheroes2::Point & pt ) const;
 
-    CastleDialogReturnValue OpenDialog( const bool openConstructionWindow );
+    CastleDialogReturnValue OpenDialog( const bool openConstructionWindow, const bool fade = false, const bool renderBackgroundDialog = true );
```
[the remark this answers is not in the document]
Districh-ru: Sure!
I removed default values from this function and did the same in heroes.h for `OpenDialog()`

src/fheroes2/kingdom/kingdom_overview.cpp
Districh-ru: I made a `fheroes2::Rect _windowWithShadowArea` in `StandardWindow` class.
It can be used in other places of the code where we use `StandardWindow` as a render ROI.

src/fheroes2/system/settings.h
Districh-ru: Initially I left this setting as it was already in the code and wanted to discuss what we will do with it.
I already discussed the ability to set fade on/off in configuration file with @Branikolog and we decided that there will be no need in disabling fade effect after we polish it in this PR.
Now, after I got your comment I removed `isFadeEffectEnabled()` from the code.

Branikolog (approved):
👍

ihhub:
Hi @Districh-ru , I tested on my machine: why the fading animation is so fast?

ihhub:
SDL 2, debug build:

https://github.com/ihhub/fheroes2/assets/19829520/c534efbf-e33c-4c32-821a-1e6da6ae3e6b

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -44,6 +44,13 @@
 
 namespace
 {
+    // The parameters of display fade start and end alpha.
+    const uint8_t fullDarkAlpha = 0;
+    const uint8_t fullBrightAlpha = 255;
+    const uint32_t screenFadeTimeMs = 100;
+    const uint32_t screenFadeFrameCount = 4;
```
ihhub, line 51: Here we are doing 4 frame of animation within 100 ms. This equal to 40 FPS.  We can easily increase it to 60 FPS so the animation might look much smoother.

ihhub (commented):
Hi @Districh-ru , I left one suggestion to improve user experience for animation. Please take a look.

src/fheroes2/gui/ui_tool.cpp
Districh-ru: Hi, @ihhub.
Sure.
Initially it was 60 fps but later I decided to slightly reduce the computational load.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -114,6 +123,54 @@ namespace
...
+                    break;
+                }
+
+                const uint8_t fadeAlpha = static_cast<uint8_t>( std::round( alpha ) );
+
+                if ( fadeAlpha == 255 ) {
+                    // This alpha is for fully bright image so there is no need to apply alpha.
+                    Copy( temp, 0, 0, display, roi.x, roi.y, roi.width, roi.height );
+                }
+                else if ( fadeAlpha == 0 ) {
```
ihhub, line 159: Here is a chance that **alpha** might be bigger than 255 or smaller than 0 due to a chosen step. I suggest to change conditions to **>= 255** and **<= 0** and convert `alpha` to **int32_t** first.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -114,6 +123,54 @@ namespace
...
 
     SystemInfoRenderer systemInfoRenderer;
+
+    void fadeDisplay( const uint8_t startAlpha, const uint8_t endAlpha, const fheroes2::Rect & roi, const uint32_t fadeTimeMs, const uint32_t frameCount )
+    {
+        if ( ( frameCount < 2 ) || roi.height == 0 || roi.width == 0 ) {
+            return;
+        }
+
+        fheroes2::Display & display = fheroes2::Display::instance();
```
ihhub, line 133: I suggest to run **operator^** or fheroes2::Rect to correct **roi** as it might be out of display's ROI. If it should not happen then it is better to assert.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -838,7 +864,16 @@ fheroes2::GameMode Interface::Basic::HumanTurn( const bool isload )
...
+    if ( _needFadeIn ) {
+        _needFadeIn = false;
+
+        fheroes2::fadeInDisplay();
+
+        SetRedraw( REDRAW_GAMEAREA );
+    }
+    else {
+        display.render();
+    }
```
ihhub, line 876: Can we move this code into a separate function as we have an exact code at lines 1293-1302?

src/fheroes2/gui/interface_events.cpp
```diff
@@ -355,10 +356,16 @@ fheroes2::GameMode Interface::Basic::EventScenarioInformation()
...
 
+        // We are opening campaign scenario info. It is a full screen image change. So do th Adventure map screen fade-out.
+        fheroes2::fadeOutDisplay();
+
         AudioManager::ResetAudio();
 
         const fheroes2::GameMode returnMode = Game::SelectCampaignScenario( fheroes2::GameMode::CANCEL, true );
         if ( returnMode == fheroes2::GameMode::CANCEL ) {
+            // We are going back to the Adventure map with fade-in.
+            Get().setNeedFadeIn();
```
ihhub, line 367: Since we are using an instance of this class shall we convert this method into non-static to avoid calling **Get()**?

src/fheroes2/game/game_loadgame.cpp
```diff
@@ -167,7 +168,12 @@ fheroes2::GameMode Game::LoadGame()
     buttons.back().setPosition( buttonPos.x, buttonPos.y + buttonYStep * 5 );
     buttons.back().draw();
 
-    display.render();
+    if ( isNeedFadeIn() ) {
+        fheroes2::fadeInDisplay();
+    }
+    else {
+        display.render();
+    }
```
ihhub, line 176: I suggest to move this code into a separate function and replace 5 places with a call of the newly created function. This will also help with **isNeedFadeIn()** as from the name a developer thinks that nothing changes which is not true.

ihhub (changes requested):
Hi @Districh-ru , I left few more comments here. Could you please take a look when you have time?

src/fheroes2/gui/ui_tool.cpp
Districh-ru: Hi @ihhub, the first and the last value of **alpha** is determined by `uint8_t startAlpha` and `uint8_t endAlpha` which cannot be < 0 or > 255. The **alphaStep** is calculated to get all **alpha** values in given limits (from `startAlpha` to `endAlpha`). *alphaStep** and **alpha** has to be the floating point type to get the correct values in `[startAlpha - endAlpha]` range. And there is no chance of alpha to be > 255 or < 0.
If I change **alpha** (or `alphaStep`) to any integer type then it is possible to get the **alpha** values out of the given range.
So I left **alpha** as double, but add an assertion: `assert( alpha >= 0 && alpha <= 255 );` to overprotect this code. :)

src/fheroes2/gui/ui_tool.cpp
Districh-ru: Sure, I used the **^** operator with display ROI. Currently it will not happen, but in future it could be possible if we will allow to place (the movable_ windows partly out of screen.

src/fheroes2/gui/interface_events.cpp
Districh-ru: Thanks for the suggestion! I removed this method and used already made functions in Game namespace.

src/fheroes2/game/game_loadgame.cpp
Districh-ru: I renamed this method to 'validateDisplayFadeIn()' (like the name for validate game delays functions, which checks the delay and resets it).

ihhub:
Hi @Districh-ru , please rebase this pull request as it contains merge conflicts.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1341,8 +1370,13 @@ void Battle::Interface::RedrawPartialStart()
 
 void Battle::Interface::RedrawPartialFinish()
 {
-    fheroes2::Display & display = fheroes2::Display::instance();
+    redrawPreRender();
+
+    fheroes2::Display::instance().render();
```
ihhub, line 1375: Do we need to redraw the whole display?

src/fheroes2/game/game_newgame.cpp
```diff
@@ -239,9 +243,12 @@ fheroes2::GameMode Game::NewSuccessionWarsCampaign()
     fheroes2::Display & display = fheroes2::Display::instance();
     const fheroes2::Point roiOffset( ( display.width() - display.DEFAULT_WIDTH ) / 2, ( display.height() - display.DEFAULT_HEIGHT ) / 2 );
 
+    // Fade-out screen before playing video.
+    fheroes2::fadeOutDisplay();
+
     display.fill( 0 );
```
ihhub, line 249: Do we need to call this method since **fheroes2::fadeOutDisplay();** does fading to the black frame?

src/fheroes2/game/game_newgame.cpp
```diff
@@ -341,6 +348,9 @@ fheroes2::GameMode Game::NewPriceOfLoyaltyCampaign()
         return fheroes2::GameMode::SELECT_CAMPAIGN_SCENARIO;
     }
 
+    // Fade-out display before playing video.
+    fheroes2::fadeOutDisplay();
```
ihhub, line 352: The same comment for line 368 as for line 249.

ihhub (changes requested):
Hi @Districh-ru , I left few more comments here. We are almost there. Please take a look once you have time.

src/fheroes2/game/game_newgame.cpp
Districh-ru: Currently we cannot guarantee that the black color index, used in `fadeOutDisplay()` is the same as black in video palette.
I'll change the fade function to always use **0** for the full black color index as in all game palettes it is used for the black color. And then we can remove `display.fill( 0 );`.

ihhub:
@Districh-ru , thank you very much for these improvements!

PR #7068, opened 2023-04-26
Fix calculation of castle towers strength for AI

### LINKED ISSUE #5523: Ai defending a castle will just sit in there with his flying units taking ranged hits until they die
### Preliminary checks

I tried fheroes2 out in it's early development, and had this issue back then too, which is main reason i decided to give it a break and come back and try it later on when you've had some time fixing all the ai problems. Basically, when you attack an enemy that is defending his base with flying units, they will just sit in there until you go inside. So i can basically just sit there with my titans killing off the black dragons inside his base since he won't move. 

I think the ai is acting this way cause it thinks the arrow towers will deal with the enemy and then it just breaks. When i encountered this problem again yesterday he also had range unit which i killed off, and maybe ai doesn't recognize the ranged units being dead and try to defend them?

### Additional info

_No response_

### LINKED ISSUE #7063: Castle siege, AI stays in castle with only 1 shooting tower with a decent melee army
### Preliminary checks

### Platform

Windows

### Describe the bug

There's only 1 tower shooting of the defending AI. Hero has no attacking spells but he still stays in castle until all walls are broken:
image

### Save file

[Dragon_Island_0002.zip](https://github.com/ihhub/fheroes2/files/11326953/Dragon_Island_0002.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #5523
fix #7063

Plus minor code modernization.

Case from #5523:

https://user-images.githubusercontent.com/32623900/234799576-f344d4e6-29cd-43fb-ae91-ede338204295.mp4

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -509,7 +509,7 @@ namespace AI
             const bool attackerIgnoresCover
                 = arena.GetForce1().GetCommander()->GetBagArtifacts().isArtifactBonusPresent( fheroes2::ArtifactBonusType::NO_SHOOTING_PENALTY );
 
-            auto getTowerStrength = [&currentUnit]( const Tower * tower ) { return ( tower && tower->isValid() ) ? tower->GetScoreQuality( currentUnit ) : 0; };
+            auto getTowerStrength = []( const Tower * tower ) { return ( tower && tower->isValid() ) ? tower->GetStrength() : 0; };
 
             double towerStr = getTowerStrength( Arena::GetTower( TowerType::TWR_CENTER ) );
             towerStr += getTowerStrength( Arena::GetTower( TowerType::TWR_LEFT ) );
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -882,8 +882,8 @@ void Troops::JoinAllTroopsOfType( const Troop & targetTroop ) const
     }
 }
 
-Army::Army( HeroBase * hero /* = nullptr */ )
-    : commander( hero )
+Army::Army( HeroBase * cmdr /* = nullptr */ )
+    : commander( cmdr )
     , _isSpreadCombatFormation( true )
     , color( Color::NONE )
 {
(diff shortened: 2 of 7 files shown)


src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -509,7 +509,7 @@ namespace AI
             const bool attackerIgnoresCover
                 = arena.GetForce1().GetCommander()->GetBagArtifacts().isArtifactBonusPresent( fheroes2::ArtifactBonusType::NO_SHOOTING_PENALTY );
 
-            auto getTowerStrength = [&currentUnit]( const Tower * tower ) { return ( tower && tower->isValid() ) ? tower->GetScoreQuality( currentUnit ) : 0; };
```
oleg-derevenetz, line 512: Original logic here looks somewhat surreal - as if the tower is going to shoot at the current unit.

idshibanov: Tower is an odd thing to estimate. I think it the future it needs to override the `GetStrength()` method with its own version looping through enemies on the board.

oleg-derevenetz: Hi @idshibanov currently `GetStrength()` is used to estimate the strength of all regular units in this function, so I suppose that estimation of towers should follow the same pattern, right?

idshibanov: Whole thing needs cleaning up to be honest, but it's a quite an undertaking.

GetStrength gives you a strategic value of a ArmyTroop without taking into account current battle situation/spells/statuses. It's fine to use for a tower, but ideally it needs its own code to determine value (i.e. tower has no health and can't be targeted). Most of the battle code uses score quality instead of strength but it requires to know who's attacking and defending.

oleg-derevenetz: @idshibanov I propose to make improvements of such kind later in separate PR (they still need to be thought through, e.g. how exactly the potential `Battle::Unit::GetStrength()` should perform its estimation), and so far just use the same logic for all battle units in this method, in order not to introduce distortions in the strength evaluation mechanics. With the current mechanics (`GetStrength()` for regular units, `GetScoreQuality()` against the `currentUnit` for towers), the distortions are too great.

Branikolog (approved):

ihhub (approved):

ihhub:
Hi @idshibanov , could you please review this pull request when you have time?

idshibanov (approved):

ihhub:

PR #7079, opened 2023-04-28
Print more detailed information about the hero patrol settings in debug mode

oleg-derevenetz:
Currently in `master` branch only patrol distance ("square") is printed, but not the patrol center. A number of clumsily named variables have also been renamed.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for cleaning up the code!

PR #7080, opened 2023-04-29
Fix adventure map fog reveal on campaign scenario restart and when loading a Hot Seat save made in the first game day.

Districh-ru:
The issue was found by **Stisen11** on Discord channel: "When I restart Good Campaign level 2, the adventure map is ALL black! That is until I click and move a hero. Then the fog of war renders wierdly..."

This PR does these fixes:
- do not force set player color to -1 when you end current player turn not by "ending" it (i.e. restarting a campaign scenario);
- set the the only human player color before updating adventure map fog directions for Single player games;
- rework of `Players::FriendColors()` to return return current friends player colors, if player does not exist - return 0.

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -636,7 +636,9 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
     const bool isHotSeatGame = conf.IsGameType( Game::TYPE_HOTSEAT );
     if ( !isHotSeatGame ) {
-        assert( currentColor != -1 );
+        // It is not a Hot Seat (multiplayer) game so we set current color to the only human player.
+        conf.SetCurrentColor( Players::HumanColors() );
+
         // Fully update fog directions if there will be only one human player.
         Interface::GameArea::updateMapFogDirections();
     }
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -706,8 +706,8 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
                     conf.SetCurrentColor( playerColor );
 
-                    // Reveal fog to be compatible with the old pre 1.0.2 saves.
-                    if ( isHotSeatGame ) {
+                    // Reveal fog on the first day to be compatible with the old pre 1.0.2 saves.
+                    if ( world.CountDay() == 1 ) {
                         world.ClearFog( playerColor );
                     }
 
@@ -749,11 +749,13 @@ fheroes2::GameMode Interface::Basic::StartGame()
                     Redraw();
                     display.render();
 
+                    // Reveal fog on the first day to be compatible with the old pre 1.0.2 saves.
+                    if ( world.CountDay() == 1 ) {
+                        world.ClearFog( playerColor );
+                    }
+
                     // In Hot Seat mode there could be different alliances so we have to update fog directions for some cases.
                     if ( isHotSeatGame ) {
-                        // Reveal fog to be compatible with the old pre 1.0.2 saves.
-                        world.ClearFog( playerColor );
-
                         Maps::Tiles::updateFogDirectionsInArea( { 0, 0 }, { world.w(), world.h() }, hotSeatAIFogColors( player ) );
                     }
 
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -606,10 +606,17 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
     radar.Build();
     radar.SetHide( true );
+
+    // Hide the world map at the first drawing
+    const int currentColor = conf.CurrentColor();
+    conf.SetCurrentColor( -1 );
+
     iconsPanel.HideIcons( ICON_ANY );
     statusWindow.Reset();
 
-    SetRedraw( REDRAW_GAMEAREA | REDRAW_RADAR | REDRAW_ICONS | REDRAW_BUTTONS | REDRAW_STATUS | REDRAW_BORDER );
+    Redraw( REDRAW_GAMEAREA | REDRAW_RADAR | REDRAW_ICONS | REDRAW_BUTTONS | REDRAW_STATUS | REDRAW_BORDER );
+
+    conf.SetCurrentColor( currentColor );
 
     bool loadedFromSave = conf.LoadedGameVersion();
     bool skipTurns = loadedFromSave;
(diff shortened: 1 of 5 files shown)


Districh-ru:
Hello @Branikolog, could you please test this fix when you have time. It affect the start/load of maps, and also the function that gets "friends colors" for mini-map and for reveal functions.
I tested Campaign restart, Campaign/Standard/Hot-Seat game start and load, played several turns on the map and did not find new bugs.

Branikolog:
Hi, @Districh-ru !
Could you please look at this save file (multiplayer game)?

https://user-images.githubusercontent.com/55348946/235303679-636ffd98-1ff9-4fbc-b180-fb852995a54b.mp4

[wide test.zip](https://github.com/ihhub/fheroes2/files/11359220/wide.test.zip)

oleg-derevenetz:
Hi @Districh-ru all these changes are superfluous and only mask the real problem. The real problem is that `Players::current_color` is not cleared on campaign map restart probably because `Players::SetStartGame()` is not called.

Districh-ru:
> Hi @Districh-ru all these changes are superfluous and only mask the real problem. The real problem is that `Players::current_color` is not cleared on campaign map restart probably because `Players::SetStartGame()` is not called.

Hi @oleg-derevenetz.
Thanks for the proposal, but starting a new campaign scenario and restarting it executes the same code:
https://github.com/ihhub/fheroes2/blob/529f226f464bfd097fdfdf5c869051374f3ac795/src/fheroes2/game/game_campaign.cpp#L1527
But is reset here and retains '-1' in `Interface::Basic::StartGame()` ...

src/fheroes2/game/game_startgame.cpp
```diff
@@ -606,17 +606,10 @@ fheroes2::GameMode Interface::Basic::StartGame()
 
     radar.Build();
     radar.SetHide( true );
-
-    // Hide the world map at the first drawing
-    const int currentColor = conf.CurrentColor();
-    conf.SetCurrentColor( -1 );
-
```
oleg-derevenetz, line 613: The purpose of this piece of code (and the following one removed by you) was to properly draw the interface in case when loaded save file is "incorrect" in some way (e.g. it was created from already won/lost game). In this case player should see the UI with adventure map totally in fog first and then he gets a message from `gameResult.LocalCheckGameOver()` call (and no player will even get his turn).

Districh-ru: @oleg-derevenetz, thanks for the comment.
I reverted this change.

oleg-derevenetz: > The only concern is why we need to show the UI and the Adventure map if we will not give the player a turn. :)

Well, we should show something anyway (not some partially-drawn UI).

oleg-derevenetz (changes requested):
Hi @Districh-ru there is still an issue with multiplayer save file from @Branikolog.

Districh-ru:
> Hi @Districh-ru there is still an issue with multiplayer save file from @Branikolog.

I discussed this with @Branikolog at Discord and the issue is in old saves and it is a different issue: this PR fixes the "fog directions update" not working correctly - it is only the UI issue (the mini-map is rendered correctly, you can move your heroes, only the adventure map is covered with fog).
In @Branikolog's save the whole fog data for the red player is set to unrevealed: the mini-map is black and you cannot move the heroes.
I'm analyzing this issue.

Districh-ru:
Hi, @Branikolog.
I made the changes to properly reveal fog in old multiplayer save files. Could you check it when you have time.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -627,15 +627,18 @@ fheroes2::GameMode Interface::Basic::StartGame()
...
+        // Clear fog around heroes, castles and mines for all players when starting a new map or if the save was done at the first day.
         for ( const Player * player : sortedPlayers ) {
             world.ClearFog( player->GetColor() );
         }
     }
 
     const bool isHotSeatGame = conf.IsGameType( Game::TYPE_HOTSEAT );
     if ( !isHotSeatGame ) {
+        // It is not a Hot Seat (multiplayer) game so we set current color to the only human player.
+        conf.SetCurrentColor( Players::HumanColors() );
```
oleg-derevenetz, line 640: I recommend to check that `Players::HumanColors()` really contains just one color here, something like this:

```cpp
assert( Color::Count( Players::HumanColors() ) == 1 );
```

or maybe even perform this check directly in `Settings::SetCurrentColor()`, because you can't have multiple "current colors" anyway.

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left one suggestion, please take a look when you have time.

src/fheroes2/system/players.h
```diff
@@ -251,7 +251,16 @@ class Players : public std::vector<Player *>
...
     static int FriendColors();
 
-    int current_color;
+    int getCurrentColor() const
+    {
+        return _currentColor;
+    }
+
+    // The color should belong to one player or be NONE (neutral player).
+    void setCurrentColor( int color );
```
oleg-derevenetz, line 260: `color` can be `const` in this method.

oleg-derevenetz (approved):
Hi @Districh-ru I just left one minor suggestion. Hope that using `Color::NONE` instead of `-1` will not break anything.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -803,7 +799,9 @@ fheroes2::GameMode Interface::Basic::StartGame()
         }
 
         // don't carry the current color from the last player to the next turn
-        conf.SetCurrentColor( -1 );
+        if ( res == fheroes2::GameMode::END_TURN ) {
```
oleg-derevenetz, line 802: Hi @Districh-ru BTW it seems that now we can remove this check because current color will become `Color::NONE` in the `Players::SetStartGame()` anyway.

src/fheroes2/system/players.h
Districh-ru: Hi @oleg-derevenetz, I quickly checked that using `Color::NONE` does not break the game. `-1` instead was the cause of this PR :)

src/fheroes2/game/game_startgame.cpp
Districh-ru: Sure!

src/fheroes2/game/game_startgame.cpp
```diff
@@ -627,15 +621,18 @@ fheroes2::GameMode Interface::Basic::StartGame()
     std::vector<Player *> sortedPlayers = conf.GetPlayers().getVector();
     std::sort( sortedPlayers.begin(), sortedPlayers.end(), SortPlayers );
 
-    if ( !loadedFromSave ) {
-        // Clear fog around heroes, castles and mines for all players when starting a new map.
+    if ( !loadedFromSave || world.CountDay() == 1 ) {
```
ihhub, line 624: I assume that the second condition is only for old saves. Is it correct?

ihhub (commented):
Hi @Districh-ru , I left just one question here. Could you please take a look?

src/fheroes2/game/game_startgame.cpp
Districh-ru: Hi, @ihhub.
Yes, you are right. It will not harm the new save files.

ihhub (approved):

ihhub:

PR #7084, opened 2023-04-30
Standartize metadata for all objects on Adventure Map

ihhub:
- remove cryptic quantity1 and quantity2 values everywhere in the code and replace them with human-readable **metadata** array
- avoid hardcoded values for many object conditions like Artifact, Daemon Cave or Shipwreck
- fix AI pathfinding when Spell Scroll was considered as an artifact which requires to defeat an army
- fix Shipwreck Survivor bonuses. It always contained artifact of level 1
- add many checks in the code to prevent calling functions for unsupported objects
- add specialized functions to get information about objects. We have to avoid generic functions

relates to #6845 

**ATTENTION**
This pull request changes almost every object interaction on Adventure Map in relation to conditions and resources. Also I added a conversion code from old save format to a new one. As a result, it is important to test all objects for new maps as well as objects on maps loaded from old saves.

The best strategy for testing by developers is to enable debug mode and see metadata values on objects. Now it is easily readable.

### DIFF
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -404,7 +404,7 @@ namespace Maps
 
         static uint8_t convertOldMainObjectType( const uint8_t mainObjectType );
 
-        // The old code was using a weird quantity based values which were very hard to understand.
+        // The old code was using weird quantity based values which were very hard to understand.
         // Since we must have backwards compatibility we need to do the conversion.
         void quantityIntoMetadata( const uint8_t quantityValue1, const uint8_t quantityValue2, const uint32_t additionalMetadata );
 
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -399,11 +399,6 @@ namespace Maps
         friend StreamBase & operator<<( StreamBase &, const Tiles & );
         friend StreamBase & operator>>( StreamBase &, Tiles & );
 
-        friend bool operator<( const Tiles & l, const Tiles & r )
-        {
-            return l.GetIndex() < r.GetIndex();
-        }
-
         static void renderAddonObject( fheroes2::Image & output, const Interface::GameArea & area, const fheroes2::Point & offset, const TilesAddon & addon );
         void renderMainObject( fheroes2::Image & output, const Interface::GameArea & area, const fheroes2::Point & offset ) const;
 
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -23,6 +23,7 @@
 #ifndef H2TILES_H
 #define H2TILES_H
 
+#include <array>
 #include <cstdint>
 #include <list>
 #include <string>
@@ -332,67 +333,16 @@ namespace Maps
         // (castle has a hero or garrison, dwelling has creatures, etc)
         bool isCaptureObjectProtected() const;
 
-        // Get Tile metadata field #1 (used for things like monster count or resource amount)
-        uint8_t GetQuantity1() const
-        {
-            return quantity1;
-        }
-
-        void setQuantity1( const uint8_t value )
-        {
-            quantity1 = value;
-        }
-
-        void setQuantity2( const uint8_t value )
-        {
-            quantity2 = value;
-        }
-
-        // Get Tile metadata field #2 (used for things like animations or resource type )
-        uint8_t GetQuantity2() const
-        {
-            return quantity2;
-        }
-
-        int QuantityVariant() const
-        {
-            return quantity2 >> 4;
-        }
-
-        int QuantityExt() const
-        {
-            return 0x0f & quantity2;
-        }
-
-        void QuantitySetVariant( const int variant )
-        {
-            quantity2 &= 0x0f;
-            quantity2 |= variant << 4;
-        }
-
-        void QuantitySetExt( int ext )
-        {
-            quantity2 &= 0xf0;
-            quantity2 |= ( 0x0f & ext );
-        }
-
         void SetObjectPassable( bool );
 
-        // Get additional metadata.
-        int32_t getAdditionalMetadata() const
+        const std::array<uint32_t, 3> & metadata() const
         {
-            return additionalMetadata;
+            return _metadata;
         }
 
-        // Set Tile additional metadata field.
-        void setAdditionalMetadata( const uint32_t value )
+        std::array<uint32_t, 3> & metadata()
         {
-            additionalMetadata = value;
-        }
-
-        void clearAdditionalMetadata()
-        {
-            additionalMetadata = 0;
+            return _metadata;
         }
 
         Heroes * GetHeroes() const;
@@ -457,6 +407,12 @@ namespace Maps
         static void renderAddonObject( fheroes2::Image & output, const Interface::GameArea & area, const fheroes2::Point & offset, const TilesAddon & addon );
         void renderMainObject( fheroes2::Image & output, const Interface::GameArea & area, const fheroes2::Point & offset ) const;
 
+        static uint8_t convertOldMainObjectType( const uint8_t mainObjectType );
+
+        // The old code was using a weird quantity based values which were very hard to understand.
+        // Since we must have backwards compatibility we need to do the conversion.
+        void quantityIntoMetadata( const uint8_t quantityValue1, const uint8_t quantityValue2, const uint32_t additionalMetadata );
+
         Addons addons_level1; // bottom layer
         Addons addons_level2; // top layer
 
@@ -493,12 +449,7 @@ namespace Maps
 
         uint8_t heroID = 0;
 
-        // TODO: Combine quantity1 and quantity2 into a single 16/32-bit variable except first 2 bits of quantity1 which are used for level type of an object.
-        uint8_t quantity1 = 0;
-        uint8_t quantity2 = 0;
-
-        // Additional metadata is not set from map's information but during runtime like spells or monster joining conditions.
-        int32_t additionalMetadata = 0;
+        std::array<uint32_t, 3> _metadata{ 0 };
 
         bool tileIsRoad = false;
 
--- a/src/fheroes2/maps/maps_tiles_helper.cpp
+++ b/src/fheroes2/maps/maps_tiles_helper.cpp
@@ -509,7 +509,7 @@ namespace Maps
         return { Resource::GOLD, tile.metadata()[1] };
     }
 
-    ShipwreckCaptureCondition getShipwrechCaptureCondition( const Tiles & tile )
+    ShipwreckCaptureCondition getShipwreckCaptureCondition( const Tiles & tile )
     {
         if ( tile.GetObject( false ) != MP2::OBJ_SHIPWRECK ) {
             // Why are you calling this for an unsupported object type?
@@ -922,7 +922,7 @@ namespace Maps
             if ( art == Artifact::SPELL_SCROLL ) {
                 static_assert( Spell::FIREBALL < Spell::SETWGUARDIAN, "The order of spell IDs has been changed, check the logic below" );
 
-                // Spell ID is by 1 bigger than in the original game.
+                // Spell ID has a value of 1 bigger than in the original game.
                 const uint32_t spell = std::clamp( tile.metadata()[0] + 1, static_cast<uint32_t>( Spell::FIREBALL ), static_cast<uint32_t>( Spell::SETWGUARDIAN ) );
 
                 tile.metadata()[1] = spell;
@@ -1024,7 +1024,7 @@ namespace Maps
                 res = Resource::Rand( false );
             }
 
-            // 2 pieces of random resource.
+            // 2 pieces of random resources.
             setResourceOnTile( tile, res, 2 );
             break;
         }
@@ -1032,7 +1032,7 @@ namespace Maps
         case MP2::OBJ_LEAN_TO:
             assert( isFirstLoad );
 
-            // 1-4 pieces of random resource.
+            // 1-4 pieces of random resources.
             setResourceOnTile( tile, Resource::Rand( false ), Rand::Get( 1, 4 ) );
             break;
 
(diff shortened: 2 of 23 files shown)


ihhub:
Hi @oleg-derevenetz , @idshibanov and @Districh-ru , could you please review this pull request very carefully as there is a chance that I messed up somewhere. Once you are done with the review I will ask more people to review from the perspective of users.

ihhub:
So far I've tested most of stuff and it works fine for me. However, I don't have enough old save files to test everything.

Districh-ru (changes requested):
Hi, @ihhub.
I did a short test and got some bugs. Could you check it when you have time.
An assertion occurs when I try to open View World on map GRWAR2.MX2 (Great War 2). It happens in `Funds getDailyIncomeObjectResources( const Tiles & tile )` (file "maps_tiles_helper.cpp"):
`tile.GetObject( false )` returns **OBJ_RESOURCE (155)** for the tile #18415 and the tile actually has resource:
изображение
<details><summary>Tile info:</summary>
******* Tile info *******
Tile index      : 18415, point: (127, 127)
UID             : 8764
MP2 object type : 155 (Ресурс)
ICN object type : 46 (OBJNRSRC.ICN)
image index     : 5 (animated: 0)
layer type      : 0 - Object layer
region          : 8
ground          : Пустошь (isRoad: 0)
shadow          : false
passable from   : center, top, top-right, right, bottom-right, bottom, bottom-left, left
metadata value 1: 4
metadata value 2: 9
metadata value 3: 0
--------- Level 1 --------
UID             : 2559
ICN object type : 37 (OBJNTWSH.ICN)
image index     : 128
layer type      : 2 (2) - Shadow layer
is shadow       : yes
--------- Level 2 --------
UID             : 3025
ICN object type : 31 (MTNCRCK.ICN)
image index     : 25
layer type      : 0 (0) - Object layer
is shadow       : no
--- Extra information ---
*************************
</details>

This assertion occurs also on other started/loaded maps.

The same bug occurs when I try to load this save: [221_ghost_legion.zip](https://github.com/ihhub/fheroes2/files/11366462/221_ghost_legion.zip)

Districh-ru:
I checked View World fix and now it works correctly and the attached save file loads correctly.
I found another bug:
When I try to load [vh2_bad_mine.zip](https://github.com/ihhub/fheroes2/files/11366481/vh2_bad_mine.zip) I also get an assertion in `Maps::Tiles::quantityIntoMetadata()`.

ihhub:
Hi @Districh-ru , thank you very much for catching these bugs! I addressed all of them.

Districh-ru (changes requested):
Loading my save files I got assertions. could you check them also when you have time.
**1.**  [vh2_bad_mine.zip](https://github.com/ihhub/fheroes2/files/11366481/vh2_bad_mine.zip) I also get an assertion in `Maps::Tiles::quantityIntoMetadata()` (map_tiles.cpp)  - > `MP2::OBJ_SAWMILL`.
**2.** [A_Battle_of_four_0001.zip](https://github.com/ihhub/fheroes2/files/11366557/A_Battle_of_four_0001.zip) - in `quantityIntoMetadata()` - > `MP2::OBJ_DERELICT_SHIP`
**3.** [Biznaf_1_0065.zip](https://github.com/ihhub/fheroes2/files/11366562/Biznaf_1_0065.zip)  - in `quantityIntoMetadata()` - > `MP2::OBJ_SAWMILL`

Districh-ru:
After https://github.com/ihhub/fheroes2/pull/7084/commits/8482cdcbbee6f6111ec8279ed9bb02c0c09822cb I get only this bug:
[Biznaf_1_0065.zip](https://github.com/ihhub/fheroes2/files/11366562/Biznaf_1_0065.zip) and [3_Continents_0005.zip](https://github.com/ihhub/fheroes2/files/11366603/3_Continents_0005.zip) in "profit.cpp" `ProfitConditions::FromMine()`
All the rest of my save file load without an assertion. :)

ihhub:
Hi @Districh-ru , I fixed the last bug as well. Please check again.

Districh-ru:
> Hi @Districh-ru , I fixed the last bug as well. Please check again.

Well done!
Now all my saves load without assertions.

src/fheroes2/maps/maps_tiles.h
```diff
@@ -493,12 +449,7 @@ namespace Maps
 
         uint8_t heroID = 0;
 
-        // TODO: Combine quantity1 and quantity2 into a single 16/32-bit variable except first 2 bits of quantity1 which are used for level type of an object.
-        uint8_t quantity1 = 0;
-        uint8_t quantity2 = 0;
-
-        // Additional metadata is not set from map's information but during runtime like spells or monster joining conditions.
-        int32_t additionalMetadata = 0;
+        std::array<uint32_t, 3> _metadata{ 0 };
```
oleg-derevenetz, line 452: Why `std:array` and not `std::tuple`? As far as I can see, there are no such situations when index is not known at compile time. Am I wrong?

oleg-derevenetz (approved):
Hi @ihhub I checked savefiles that I have (mostly those from bug reports) and tried some new games with big maps, everything seems fine so far, but this of course requires more extensive testing. As for the code itself, I left one (optional) question, could you please take a look when possible?

src/fheroes2/maps/maps_tiles.h
ihhub: Hi @oleg-derevenetz , I honestly did not think about tuple :) I can change it to tuple.

oleg-derevenetz: It's up to you :) There is `std::get<N>( T )` function that extracts N'th element from T, N is verified at compile time. It works also with arrays, but with arrays you can forget and write `T[N]`, while with tuple you cannot.

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -186,233 +381,192 @@ namespace Maps
...
+            assert( 0 );
+            return {};
+        }
 
-        assert( artifactId >= std::numeric_limits<Quantity1Type>::min() && artifactId <= std::numeric_limits<Quantity1Type>::max() );
+        switch ( static_cast<ArtifactCaptureCondition>( tile.metadata()[2] ) ) {
+        case ArtifactCaptureCondition::HAVE_WISDOM_SKILL:
+            return { Skill::Secondary::LEADERSHIP, Skill::Level::BASIC };
+        case ArtifactCaptureCondition::HAVE_LEADERSHIP_SKILL:
+            return { Skill::Secondary::WISDOM, Skill::Level::BASIC };
```
Districh-ru, line 424: Maybe there is a misprint: the return values are swapped

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -524,44 +671,78 @@ namespace Maps
...
+        case MP2::OBJ_TREE_HOUSE:
+        case MP2::OBJ_TROLL_BRIDGE:
+        case MP2::OBJ_WAGON_CAMP:
+        case MP2::OBJ_WATCH_TOWER:
+        case MP2::OBJ_WATER_ALTAR:
+            tile.metadata()[0] = count;
+            return;
+        default:
+            // Why are you calling this function for an unsupported object type?
+            assert( 0 );
```
Districh-ru, line 740: After defeating a monster in a New game I get this assertion: **tile.GetObject( false )** returns **OBJ_NONE (0)**

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -186,233 +381,192 @@ namespace Maps
...
-            }
-            break;
-
         case MP2::OBJ_CAMPFIRE:
-            return Funds( Resource::GOLD, getGoldAmountFromTile( tile ) ) + Funds( rc );
+            return Funds{ static_cast<int>( tile.metadata()[0] ), tile.metadata()[1] } + Funds{ Resource::GOLD, tile.metadata()[1] * 100 };
 
         case MP2::OBJ_FLOTSAM:
-            return Funds( Resource::GOLD, getGoldAmountFromTile( tile ) ) + Funds( Resource::WOOD, tile.GetQuantity1() );
+            return Funds{ Resource::WOOD, tile.metadata()[0] } + Funds{ Resource::GOLD, tile.metadata()[1] };
```
Districh-ru, line 543: Should we multiply **tile.metadata()[1]** by 100 here?
As I understand we store gold in hundreds. It would be good to have a comment about maybe in Funds class.
Or could we do multiply the gold amount by 100 when we are putting it to `tile.metadata()[1]` and do not use `* 100` in every place we want to get gold amount from metadata.

Districh-ru (changes requested):
Hi, @ihhub.
I have some bug reports and suggestions. Could you please take a look when you have time.

**1.** Picking up a any resource results in displaying that I picked **0** of the resource. But it is actually picked up as the amount of kingdom resource is increased:
изображение

**2.** I cannot pick-up a bottle: [221_bottle_bug.zip](https://github.com/ihhub/fheroes2/files/11374566/221_bottle_bug.zip)
изображение

src/fheroes2/maps/maps_tiles_helper.cpp
ihhub: Nope. We don't store Gold in hundreds anymore. Old format did but for simplification of debugging and development I remove this logic so Flotsam actually contains correct gold.

src/fheroes2/maps/maps_tiles_helper.cpp
ihhub: Yes, it was a bug. It is fixed now.

src/fheroes2/maps/maps_tiles_helper.cpp
ihhub: It is fixed now. We were deleting monster object from the map before setting metadata. This was a bug from the old code.

ihhub:
Hi @Districh-ru , I addressed all your comments except the bottle one. I don't know how it happens but it seems that the bottle actually doesn't exist there as it is showing OBJ_NONE. Did you start a new game for this?

Districh-ru:
> Hi @Districh-ru , I addressed all your comments except the bottle one. I don't know how it happens but it seems that the bottle actually doesn't exist there as it is showing OBJ_NONE. Did you start a new game for this?

Yes, I started a new game. In master built this bottle can be picked up. This is my test map: [221.zip](https://github.com/ihhub/fheroes2/files/11374857/221.zip)

src/fheroes2/maps/maps_tiles_helper.cpp
Districh-ru: The previous case for MP2::OBJ_CAMPFIRE has `Funds{ Resource::GOLD, tile.metadata()[1] * 100' so I thought that we missed it here. :)

ihhub:
Hi @Districh-ru , I fixed the bug and also some objects were missing on Adventure Map. Please take a look.

src/fheroes2/maps/maps_tiles_helper.cpp
ihhub: Campfire is the only object which has a connection between resource and gold. I checked the original game and if resource is N then the given gold is always N * 100.

Districh-ru: So maybe It'll be good to leave this comment in code to have no questions in future. :)

ihhub: It is done.

ihhub:
Hi @LeHerosInconnu, could you please help to test changes in this pull request? We need your comprehensive testing skills for this pull request.

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu, could you please help to test changes in this pull request? We need your comprehensive testing skills for this pull request.

Are there any other changes planned before I start the test?

ihhub:
> Hello @ihhub,
> 
> > Hi @LeHerosInconnu, could you please help to test changes in this pull request? We need your comprehensive testing skills for this pull request.
> 
> Are there any other changes planned before I start the test?

Nope. I've recently fixed some issues found by @Branikolog and reported through Discord. As of now only bug fixing. You can download the newer version once the compilation is done.

oleg-derevenetz:
Hi @ihhub is this fine that when Tree of Knowledge has no resource requirement, there is "unknown resource" warning printed in the `Funds` constructor? It doesn't interfere with the game, but it looks weird and litters the log with messages about what is not really a problem.

ihhub:
> Tree of Knowledge

Hi @oleg-derevenetz , actually it is wrong. This message is added to avoid bad written code. I will have a look at it. Thank you for spotting it!

ihhub:
@oleg-derevenetz , fixed!

ihhub:
> What do you think about the following fixing method instead: allow to pass the `Resource::UNKNOWN` value to the `Funds` constructor (add the special `switch` case with just `break` and maybe also a check that `count` is zero as well and log a warning if it is not, because there cannot be non-zero amount of unknown resource)? Or this is undesirable?

I think we should prevent even passing unknown types into the constructor. This might help to find some future issues with wrong code or wrong data read from files. Of course no harm by doing this but in my view we should log such undesired behavior.

oleg-derevenetz:
Hi @ihhub there is the same issue in the `Maps::getFundsFromTile()`: non-pickable objects may contain no funds.

ihhub:
> Hi @ihhub there is the same issue in the `Maps::getFundsFromTile()`: non-pickable objects may contain no funds.

... I shall add the case with zero unknown resources :(

LeHerosInconnu:
Hello @ihhub,

I did a test on a test scenario and did not encounter any problems.
I will do another test playing a standard scenario.

ihhub:
Thank you @LeHerosInconnu !

Branikolog (approved):
Haven't found more bugs here. Great job, fheroes2 team! :D

Districh-ru (approved):
I played fully one new standard game and for a couple of week a campaign scenario and a standard game from an old save file. I did not notice any bugs.
Well done!

LeHerosInconnu:
Hello @ihhub,

> Hello @ihhub,
> 
> I did a test on a test scenario and did not encounter any problems.
> I will do another test playing a standard scenario.

I finished a standard scenario started with the PR version and continued with the merged version without any problems.

PR #6957, opened 2023-04-07
Add proper icons for screen scrolling speed

ihhub:
close #6552

all credits go to @golden-lining

PR #7094, opened 2023-05-01
WIP: Support for 32bit images

Aurelain:
The type of `image()` has been converted from `uint8_t` to `uint32_t`.
With this change, normal RGBA images can be rendered:
image
This should usher a bright new future for fheroes2 :).

Some notes:
- My strategy was as follows: change the type and then fix every new error that Visual Studio throws. In 80% of the case the fix was straightforward (just adapt the types). The most troublesome were the `memcpy()`, but were eventually fixed.
- Some extensive testing is needed. I just entered a few games and clicked some menus to see if anything crashes.
- This change is a milestone to another Pull Request that I'll submit in the coming days, where we allow the user to place unpacked png files next to the agg archives and they'll be dynamically loaded in game, overriding the old assets.
- SDL1 no longer works due to this change. From my understanding, you were already thinking of deleting it from the source code.
- I have close to zero experience with C++, so please be kind.

### DIFF
--- a/src/engine/h2d_file.cpp
+++ b/src/engine/h2d_file.cpp
@@ -190,7 +190,13 @@ namespace fheroes2
 
         const size_t size = static_cast<size_t>( width * height );
         image.resize( width, height );
-        memcpy( image.image(), data.data() + 4 + 4 + 4 + 4, size );
+
+        //memcpy( image.image(), data.data() + 4 + 4 + 4 + 4, size * 4 );
+        uint32_t * imageData = image.image();
+        const uint8_t * pixelData = data.data() + 4 + 4 + 4 + 4;
+        for ( size_t j = 0; j < size; ++j, ++imageData, ++pixelData ) {
+            *imageData = *pixelData;
+        }
         memcpy( image.transform(), data.data() + 4 + 4 + 4 + 4 + size, size );
 
         image.setPosition( x, y );
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1019,7 +1019,7 @@ namespace fheroes2
                     for ( ; imageOutX != imageOutXEnd; --imageInX, --transformInX, ++imageOutX ) {
                         if ( *transformInX > 0 ) { // apply a transformation
                             if ( *transformInX != 1 ) { // skip pixel
-                                //*imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
+                                *imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
                             }
                         }
                         else { // copy a pixel
@@ -1044,7 +1044,7 @@ namespace fheroes2
                         }
 
                         if ( *transformInX > 0 && *transformOutX == 0 ) { // apply a transformation
-                            //*imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
+                            *imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
                         }
                         else { // copy a pixel
                             *transformOutX = *transformInX;
@@ -1074,7 +1074,7 @@ namespace fheroes2
                     for ( ; imageInX != imageInXEnd; ++imageInX, ++transformInX, ++imageOutX ) {
                         if ( *transformInX > 0 ) { // apply a transformation
                             if ( *transformInX != 1 ) { // skip pixel
-                                //*imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
+                                *imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
                             }
                         }
                         else { // copy a pixel
@@ -1099,7 +1099,7 @@ namespace fheroes2
                         }
 
                         if ( *transformInX > 0 && *transformOutX == 0 ) { // apply a transformation
-                            //*imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
+                            *imageOutX = *( transformTable + ( *transformInX ) * 256 + *imageOutX );
                         }
                         else { // copy a pixel
                             *transformOutX = *transformInX;
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -378,16 +378,16 @@ namespace
         const int32_t widthIn = in.width();
         const int32_t widthOut = out.width();
 
-        const uint8_t * imageInY = in.image() + inY * widthIn + inX;
-        uint8_t * imageOutY = out.image() + outY * widthOut + outX;
-        const uint8_t * imageInYEnd = imageInY + height * widthIn;
+        const uint32_t * imageInY = in.image() + inY * widthIn + inX;
+        uint32_t * imageOutY = out.image() + outY * widthOut + outX;
+        const uint32_t * imageInYEnd = imageInY + height * widthIn;
 
         if ( in.singleLayer() ) {
             // All pixels in a single layer image do not have any transform values so there is no need to check for them.
             for ( ; imageInY != imageInYEnd; imageInY += widthIn, imageOutY += widthOut ) {
-                const uint8_t * imageInX = imageInY;
-                uint8_t * imageOutX = imageOutY;
-                const uint8_t * imageInXEnd = imageInX + width;
+                const uint32_t * imageInX = imageInY;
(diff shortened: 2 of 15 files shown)


ihhub:
Hi @Aurelain , could you please explain why this change is needed? It actually breaks a lot of things, not only compilation.

Aurelain:
Hi, @ihhub.
I agree it has the potential to break things. However, I feel we need to be courageous about this.
The main benefits would be:
- It allows us to improve the old assets (see https://github.com/ihhub/fheroes2/pull/6438#issuecomment-1484097281)
- It allows the users to use custom assets, which work without any special knowledge of palette or tools to inject the assets into the agg files. This is related to the other PR I'm working on.

ihhub:
@Aurelain ,

We already provide **h2dmgr** tool to convert images into game's image format so anyone can add an image into the engine. The game uses 2-layer images, not a single layer as 32-bit. You replaced only the first layer responsible for storing image data but you broke the second layer functionality which is responsible for image transformation. Image transformation includes not only alpha channel logic but much more than this. Essentially, these changes killed all shadows, color cycling and many special effects in the game. On top of this, the game will be slower and eat more RAM.

It is not possible to simply use a 16- / 24- / 32-bit image format as transformation layer logic is designed only to handle 8-bit images. This is why #6438 pull request is stuck. This is not about courage or lack of C++ knowledge, it is how the game assets are designed and we have to work with them. Unless someone rewrites the whole logic from 8-bit images to 32-bit images including special effects, cycling, multiple types of shadows and etc.

ihhub:
> Could you please take some screenshots with examples of broken things (shadows, special effects)? From what I imagined, the change would have no downsides for the old assets, as any value lower than 256 should be considered as being paletted (8bit). My intention was to enhance the possibilities for new assets. As for performance, I didn't feel a change in FPS (still above 125 when traversing the map). Of course the change must have a cost, but my feeling (coupled with the already achieved transition to 32 bit surfaces) was that the current hardware should completely hide this cost.

- open any monster dialog
- see phoenix animation for cycling animation
- see Blood Lust spell animation 
- see water on Adventure Map for cycling animation
- no objects have shadows on Adventure Map
- font has no shadows
- no windows has shadows

Run the game on Raspberry Pi to see the difference. You might use too powerful PC so you can't see the difference in performance. You can measure RAM usage in Task Manager.

For now cycling animation works because all values are still stored within 8-bit range but once you start adding any 32-bit images everything will break.

ihhub:
@Aurelain , the brightest example of broken rendering is Air Elemental. If you still want to go with this direction this is a start point to make everything work.

Aurelain:
Hi, @ihhub.

It turns out the shadows and the Air Elemental are working properly, but I had forgotten them commented out in Blit. At one point they were throwing errors and I had temporarily suspended them, but I restored them now.

I checked the Pheonix and it seems fine. I don't know how to test the Blood Lust easily.
Please point me to other susceptible issues, if you know any.

And, if I may, please ponder a bit about the strategic implications of this direction. I know it feels it came out of the blue (without first discussing it with the core-team), but let's ignore these circumstances and study the PR on technical merits. We should gather some pros and cons, but besides the facts (increased memory usage) there is also a hard-to-quantify argument: fheroes2 gets **modernized**. This could open up some unexpected opportunities.

ihhub:
@Aurelain , please do proof-of-concept of 32-bit images which require special transformations like shadows since you insist on the word **modernize**. Take Air Elemental as an example. At best you would get a pixelated Air Elemental when you do resizing. Next look at standard window generation code: we generate many elements using the engine so to modernize the game you should change many places where do the such, including the whole font generation.

Yes, your solution works as of now and what you did is to expand 8-bit storage into 32-bit while all remaining operations use 8-bit. However, once you start adding 32 bit images especially those which replace the existing images with transformations then all defects would come out. Blood Lust effect appears when you use the same spell in battle. Try it on 32-bit image which contains real 32-bit values.

More importantly, what exactly you mean by "modernize" word? Upscaling existing images and distribute them? Then it is illegal and we have no rights for game's resources.

ihhub:
Hi @Aurelain , also please do not think that I am being strict towards you or your changes. It is important to understand the scope of the problem and all downsides. Hence, I asked to do POC on your side using 32-bit images to see how it works with transformation layer.

Aurelain:
> Hi @Aurelain , also please do not think that I am being strict towards you or your changes. It is important to understand the scope of the problem and all downsides. Hence, I asked to do POC on your side using 32-bit images to see how it works with transformation layer.

Hi, @ihhub. Sure, no problem, it's reasonable to require more research on such a sensitive issue.
I will investigate a possible remastering of the Air Elemental for the POC, but I have some notes:
1. the normal use-case for 32bit images would be to provide a completely new asset, which would come with all the necessary transforms (except those for flag-colors, I guess). Therefore, it would seem reasonable to just enclose all custom (surgical) edits to an asset inside a `if (is8Bit) {...}`.
2. even if this POC would fail for some use-cases, it most certainly works for others (as can be seen in the top image).
3. even if there are no immediate benefits for the 32bit images (in fact, there are some disadvantages), it feels like a stepping stone to other things. 

Some ideas for use-cases:
- Mechanic pre-processing of the assets to clean-up some pixels (as mentioned in #6438)
- Enhancement by AI
- Importing of assets from other games :)

None of the above seem illegal...

Aurelain:
Hi, @ihhub.

After some research, I have some notes:

**1. The ability to store 32bit assets inside Image could be achieved in one of 2 ways:**

a) By changing the type of image() from `uint8_t` to `uint32_t`. This is the approach I used in this PR. It has the advantages of extending the current property in-place, keeping it as the source-of-truth. Unfortunately, it also comes with some (minor?) disadvantages:
- it always requires more memory (because, of course, 4 bytes occupy more memory than 1 byte), even when not using 32bit assets (100% of current assets).
- it breaks SDL1. I don't think this is a problem, but I don't know enough about its use-cases.
- it requires changes in quite a few places in the code (as can be seen by the current clang-tidy report)

b) By adding an extra property (e.g. `image32()`). I haven't yet tried this approach. It has the disadvantage of being separate from `image()`, and thus developers might forget about it and cause desyncs. However, it might be more attractive for you, increasing the chances of accepting the PR, as it would have some interesting advantages:
- nothing would break
- it would require fewer code changes (not sure about this, but I think only the functions in `image.cpp` should be adapted)
- it could be governed by a flag in cfg and by `if` clauses in every function that changes the `image()` in `image.cpp`.
- you could theoretically switch between the 8bit/32bit assets dynamically, at the press of a keyboard shortcut. This would have no real benefits, just bragging rights (similar to how Blizzard allowed the toggle in Diablo 2 Resurrected).

**2. When using approach 1a, the color information could be stored in one of 2 ways:**
a) As a paletted number from 0 to 255 for the 8 bit assets and as an RGBA number from 256 to 4228250625 for the 32 bit assets. This is the current approach. It has the advantage that all the existing assets work, with only a few carefully placed conditions.
b) Always as an RGBA number. During this research I also tried this technique, and I found that it works properly, but it would require some work to translate the transform table for the existing assets (indeed, Air Elemental is a good example). I think the color-cycling can also be made to work. The only advantage to this approach is that future developers are not surprised by the contents of each pixel (it would always be RGBA).

**3. The transition to 32bit is just a stepping stone to higher quality.**
As can be seen from the image below, it would directly allow level 3 of quality (in-game 32bit asset). Additionally, it would open the way to level 4 of quality, where some future PR could introduce resampling of high-resolution assets (although I don't see a horizon for that - it seems difficult).
image


**4. The h2dmgr tool produces unexpected color conversions (see level 2 of quality above).**
This might be a bug, as I expect the game palette is able to display the bluish colors of the H3 air elemental...

I'm inclined to stop working on the 32bit direction until we have a Proof-of-Concept for level 4 of quality...

oleg-derevenetz:
> The h2dmgr tool produces unexpected color conversions (see level 2 of quality above).

This should be fixed by [this](https://github.com/ihhub/fheroes2/pull/6957/commits/0fe2c290eb81fd60154d29b5e5b43c4c93b53e3c) commit in different PR. The thing is that JPG/BMP image loading functions in SDL/SDL_image can theoretically return surface in any supported color format (and there are [many of them](https://wiki.libsdl.org/SDL2/SDL_PixelFormatEnum) with different orders of colors), but `fheroes2::Load()` currently considers just one color order.

Issue #6923, opened 2023-04-01
Reduce AI obsession with Wells


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Ai is too obsessed visiting wells, especially having the Dimension Door spell. This results in an infinite loop of actions by an AI hero.

### Describe the possible solution

We have to change well value calculations to avoid an abusive behavior by AI visiting it.

ihhub:
I am closing this issue as we fixed multiple aspects of AI behavior in the past.

Issue #5884, opened 2022-09-25
AI repeats the same actions inifinitely


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

There are cases when AI repeats the same set of actions infinitely leading to broken gameplay:

https://user-images.githubusercontent.com/19829520/192142585-eef47a35-2542-4ffd-a856-d3b97be8496c.mp4

### Save file

[infinite_action_loop.zip](https://github.com/ihhub/fheroes2/files/9640696/infinite_action_loop.zip)

LeHerosInconnu:
@ihhub.

AI does not have enough memory.
Please, allocate it more memory. :)

ihhub:
This is another example:


https://user-images.githubusercontent.com/19829520/192146945-b3b36e5a-5f0f-4221-a844-d6765f9725b5.mp4

[infinite_loop_actions.zip](https://github.com/ihhub/fheroes2/files/9640894/infinite_loop_actions.zip)

ihhub:
I am closing this issue as this scenario was partially addressed by recent changes for the AI.

PR #7108, opened 2023-05-05
Do not stop hero's movement after the first Dimension Door jump

ihhub:
relates to #5884
relates to #6923

Before the changes Dimension Door jump was always done once and then the target was re-evaluated. With these changes the behavior of an AI hero is the same as a normal walk: complete the full path without interruptions and changing mind.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1927,7 +1927,6 @@ namespace AI
             uint32_t dimensionDoorDistance = AIWorldPathfinder::calculatePathPenalty( dimensionPath );
             uint32_t moveDistance = _pathfinder.getDistance( bestTargetIndex );
             if ( dimensionDoorDistance && ( !moveDistance || dimensionDoorDistance < moveDistance / 2 ) ) {
-
                 while ( dimensionDoorDistance < moveDistance / 2 && !dimensionPath.empty() && bestHero->MayStillMove( false, false )
                         && bestHero->CanCastSpell( Spell::DIMENSIONDOOR ) ) {
                     HeroesCastDimensionDoor( *bestHero, dimensionPath.front().GetIndex() );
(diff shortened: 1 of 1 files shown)


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1923,11 +1923,27 @@ namespace AI
...
             // check if we want to use Dimension Door spell or move regularly
-            const std::list<Route::Step> & dimensionPath = _pathfinder.getDimensionDoorPath( *bestHero, bestTargetIndex );
-            const uint32_t dimensionDoorDistance = AIWorldPathfinder::calculatePathPenalty( dimensionPath );
-            const uint32_t moveDistance = _pathfinder.getDistance( bestTargetIndex );
+            std::list<Route::Step> dimensionPath = _pathfinder.getDimensionDoorPath( *bestHero, bestTargetIndex );
+            uint32_t dimensionDoorDistance = AIWorldPathfinder::calculatePathPenalty( dimensionPath );
+            uint32_t moveDistance = _pathfinder.getDistance( bestTargetIndex );
             if ( dimensionDoorDistance && ( !moveDistance || dimensionDoorDistance < moveDistance / 2 ) ) {
-                HeroesCastDimensionDoor( *bestHero, dimensionPath.front().GetIndex() );
+                while ( dimensionDoorDistance < moveDistance / 2 && !dimensionPath.empty() && bestHero->MayStillMove( false, false )
```
oleg-derevenetz, line 1930: Hi @ihhub do you know why DD is used only when its distance is less than half of the total path distance?

ihhub: Hi @oleg-derevenetz , it was introduced by @idshibanov but I believe this is done to avoid an excessive usage of the spell.

oleg-derevenetz (approved):

ihhub:
Hi @idshibanov , can you please review this pull request so we can merge it for the upcoming release?

PR #7109, opened 2023-05-05
Fix not returning to load menu while switching between campaign scenarios

### LINKED ISSUE #6037: Campaign load menu doesn't return while switching between two scenarios
### Preliminary checks

### Platform

Windows

### Describe the bug

While cancelling campaign briefing the game doesn't return to load menu in a case we switch between two scenarios.
In a case we doesn't switch missions, game returns back to loading screen.

https://user-images.githubusercontent.com/55348946/197528306-0b1b5106-bdc8-4718-8c84-e59d7e42a06f.mp4

### Save file

[d2_Complete.zip](https://github.com/ihhub/fheroes2/files/9851724/d2_Complete.zip)

### Additional info

_No response_

ihhub:
close #6037


Districh-ru (approved):
I tested this PR, all works good. 👍
But to be able to load old saves that are made at the end of campaign scenario which are the point of this PR we need a fix: #7111

PR #7111, opened 2023-05-05
Fix loading campaign save made at the end of scenario

Districh-ru:
relates to #7109
This PR fixes loading a save made at the end of a campaign scenario.
Such old saves store '-1' as the current player color.
Recently we decided not to use negative numbers for player colors as in some cases it produces bugs.
All places in the engine that were setting player color to '-1' were changed to 'Color::NONE' in https://github.com/ihhub/fheroes2/pull/7080

### DIFF
--- a/src/fheroes2/system/players.cpp
+++ b/src/fheroes2/system/players.cpp
@@ -39,6 +39,7 @@
 #include "normal/ai_normal.h"
 #include "race.h"
 #include "rand.h"
+#include "save_format_version.h"
 #include "serialize.h"
 #include "settings.h"
 #include "world.h"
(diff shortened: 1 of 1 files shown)


src/fheroes2/system/players.cpp
```diff
@@ -530,6 +531,13 @@ StreamBase & operator>>( StreamBase & msg, Players & players )
     int current;
     msg >> colors >> current;
 
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_1004_RELEASE, "Remove the logic below." );
+    // The old save files made at the end of a campaign scenario has player color set to '-1' which is not supported now.
+    // So we change the incorrect color '-1' to 'Color::NONE'.
+    if ( current == -1 ) {
```
ihhub, line 537: Should we add this logic only for old saves?

Districh-ru: Yes, in new saves there are no negative values for player color. I checked it on saves made by the current master build.
In #7080 we a made the engine not to use '-1' as the player color so new saves will not contain this value.

Districh-ru: Sure, I'll add it.
But it is not necessary as `( current == -1 )` for new saves will always return `false`. It does nothing to the new saves.

ihhub: Fair point. Then let's keep it as it is.

ihhub (approved):

ihhub:

src/fheroes2/system/players.cpp
Districh-ru: OK.
In future this fix will do no harm but will be redundant so I left the save version check to remove it. :)

PR #7113, opened 2023-05-05
Remove duplicate line in translations.cpp

Roxterat:
SK & CS grammar logic merge


Roxterat:
@zenseii just a little fix because why use the same thing twice..

ihhub:
@Roxterat , thank you for this correction!

PR #7130, opened 2023-05-07
Add popup windows for monster dialog buttons

ihhub:
image

To be consistent with the rest of the game.


Branikolog (approved):
New info window works well.
If there exists better wording I believe @zenseii could propose it. I'm OK with current implementation though.

PR #7136, opened 2023-05-07
Android: limit visibility of inner class members

oleg-derevenetz:
They should not have any visibility other than `private`.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for the code cleanup!

PR #7143, opened 2023-05-08
Fix an issue that caused an AI-controlled hero standing next to Stone Liths to block the path of another AI-controlled hero standing on these Stone Liths

### LINKED ISSUE #7138: Adventure map, AI stuck in front of stone liths
### Preliminary checks

### Platform

Windows

### Describe the bug

Two yellow heroes are blocking each other:
image

### Save file

[Stuck in Liths.zip](https://github.com/ihhub/fheroes2/files/11420759/Stuck.in.Liths.zip)

### Additional info

_No response_

### LINKED ISSUE #7140: Adventure map, to re-visit stone liths AI choses too distant and not optimal cell
### Preliminary checks

### Platform

Windows

### Describe the bug

Sometimes to re-visit stone liths AI hero choses a cell to step and return, which requires him to waste not 1 movement, but 2 or more. Sometimes he starts to walk around some abject and then returns back, wasting all his movement points.

https://user-images.githubusercontent.com/55348946/236819476-18333faf-374b-4341-9ff4-4cd5405561fd.mov

One of red heroes here (0:14) choses a cell right above stone liths.

### Save file

[Stuck in Liths.zip](https://github.com/ihhub/fheroes2/files/11420934/Stuck.in.Liths.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #7138
fix #7140

At first I thought to limit the case to checking only friendly heroes nearby, but then I decided that all AI-controlled heroes should behave "like a gentleman" with enemy ones and give way to them, if it turns out that neither of them has an advantage and does not dare to attack the other.

### DIFF
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -912,6 +912,13 @@ bool AIWorldPathfinder::isHeroPossiblyBlockingWay( const Heroes & hero )
         return true;
     }
 
+    // Is the hero standing near the Stone Liths that are occupied by another hero?
+    for ( const int32_t idx : Maps::ScanAroundObject( heroIndex, MP2::OBJ_STONE_LITHS ) ) {
+        if ( world.GetTiles( idx ).GetObject() == MP2::OBJ_HEROES ) {
+            return true;
+        }
+    }
+
     // Is the hero standing on Stone Liths?
     return world.GetTiles( heroIndex ).GetObject( false ) == MP2::OBJ_STONE_LITHS;
 }
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -774,8 +774,10 @@ int AIWorldPathfinder::getNearestTileToMove( const Heroes & hero )
             continue;
         }
 
-        // Tile is reachable and the hero has enough army to defeat potential guards
-        if ( _cache[newIndex]._cost > 0 ) {
+        const WorldNode & node = _cache[newIndex];
+
+        // Tile is directly reachable (in one move) and the hero has enough army to defeat potential guards
+        if ( node._cost > 0 && node._from == start ) {
             return newIndex;
         }
     }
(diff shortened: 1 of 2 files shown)


src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -912,6 +912,13 @@ bool AIWorldPathfinder::isHeroPossiblyBlockingWay( const Heroes & hero )
         return true;
     }
 
+    // Is the hero standing near the Stone Liths that are occupied by another hero?
+    for ( const int32_t idx : Maps::ScanAroundObject( heroIndex, MP2::OBJ_STONE_LITHS ) ) {
+        if ( world.GetTiles( idx ).GetObject() == MP2::OBJ_HEROES ) {
```
ihhub, line 917: Should we check that another hero is a friendly one? Blocking a way for an enemy hero seems to be a good tactic. Also do we need to check that the blocked hero cannot move anywhere else except to the tile where the current hero stands?

ihhub (commented):
Hi @oleg-derevenetz , I left 2 questions regarding the proposed solution. Could you please check them?

src/fheroes2/world/world_pathfinding.cpp
oleg-derevenetz: @ihhub

> Should we check that another hero is a friendly one?

If neither hero has an advantage to attack the other, then they can get stuck there forever, which IMHO is undesirable.

> Also do we need to check that the blocked hero cannot move anywhere else except to the tile where the current hero stands?

If it's a hero of the same color, then he can't do it 100%, otherwise we wouldn't be here (we only end up here if AI is not able to find tasks for any of the heroes). With heroes of other colors, the situation is more complicated, but I think it would be an extra complication - after all, in other cases this function does not check whether the heroes being moved are really blocking someone, it is enough that they stand in some "narrow place".

ihhub: Yes, this would be more reasonable. Also to avoid hardcoded directions we can use passability values.

Branikolog (approved):

Branikolog:
Many thanks for the prompt fix, @oleg-derevenetz !

ihhub (approved):

ihhub:

PR #7145, opened 2023-05-08
Slovak translation update

Roxterat:
Update of Slovak translation file. Mostly hotkeys, some menu texts and most creatures. Also string order for plurals is back due to translations.cpp being fixed now.


Roxterat:
@zenseii I just can't live without this. 😄

PR #7154, opened 2023-05-09
Remove obsolete parameter of Heroes::Action()

oleg-derevenetz:
Apparently this parameter was used in the past when it was possible to pave the way through the whirlpool, but with the current passability rules it is not needed (and even harmful: if `AIAutoControlMode` is turned off on the current turn, semi-AI-controlled heroes are not able to use Whirlpool until the next turn).

This change needs to be properly tested (with different whirlpool usage, movement in jump and normal mode, etc) and it should not be merged before the upcoming 1.0.4 release.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1978,7 +1978,7 @@ namespace AI
             }
             else if ( MP2::isActionObject( hero.GetMapsObject(), hero.isShipMaster() ) ) {
                 // use the action object hero is standing on (Stone Liths)
-                hero.Action( hero.GetIndex(), true );
+                hero.Action( hero.GetIndex() );
             }
         }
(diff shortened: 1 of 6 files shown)


ihhub (approved):

ihhub:
Hi @oleg-derevenetz , could you please take care of pull request for some time as I need to solve some personal issues in meanwhile?

oleg-derevenetz:
Hi @ihhub yes, of course. But I'm not very good at some areas, e.g. rendering, so if I will be in a doubt, I'll wait for your review anyway.

ihhub:
@oleg-derevenetz , thank you very much! No worries about rendering, once I back I will handle these pull requests.

Branikolog (approved):

ihhub:
@oleg-derevenetz , thank you very much for this cleanup!

Issue #6924, opened 2023-04-01
Add support for other languages in virtual keyboard


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

The recently added virtual keyboard only supports English. While this is good for many cases it doesn't solve issues with Sphinx for instance.

### Describe the possible solution

Add the most major languages support for the virtual keyboard.

ihhub:
Hi @Branikolog , so based on this conversation lowering down the difficulty level does not make things worse :)

ihhub:
Rising the priority for this issue as it is a game breaking problem on some maps.

ihhub:
I'm closing this issue as the base work has been completed.

PR #7156, opened 2023-05-10
Extra virtual keyboard language support for Russian

ihhub:
This was the easiest to do.

image

I might not have the latest translation files for Russian so buttons might be translated for users.

relates to #6924

### DIFF
--- a/src/fheroes2/dialog/dialog_selectcount.cpp
+++ b/src/fheroes2/dialog/dialog_selectcount.cpp
@@ -367,7 +367,7 @@ bool Dialog::InputString( const std::string & header, std::string & res, const s
 
             fheroes2::Blit( inputArea, display, text_rt.x, text_rt.y );
             text.draw( text_rt.x + ( text_rt.width - text.width() ) / 2, text_rt.y + 3, display );
-            display.render();
+            display.render( box.GetArea() );
         }
     }
 
--- a/src/fheroes2/dialog/dialog_selectcount.cpp
+++ b/src/fheroes2/dialog/dialog_selectcount.cpp
@@ -21,6 +21,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "dialog.h"
+
 #include <algorithm>
 #include <cassert>
 #include <cstddef>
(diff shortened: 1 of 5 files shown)


Districh-ru (commented):
Hi, @ihhub.
Well done!
I have two suggestions which can be discussed and one issue. Could you take a look when you have time.

1. Possibly we don't need a Language Change button in a numeric keyboard as id does no change to the input characters.


https://github.com/ihhub/fheroes2/assets/113276641/a943983e-1f58-4409-9c76-469d920ce8b1



2. IMHO the SPACE button should have a translated text according to the game selected language, not the current keyboard layout.

3. We miss the "_" sign at the end of the input phrase. This is not the issue of Change Language implementation and can be addressed in a new PR.
изображение
изображение

Branikolog:
Hi, @ihhub !
I think it would be better to remember selected language, how do you think? In a case we want to enter 6 player's names we need to press each time [...] button to switch the language.
Also, I've noticed than when the text exceeds the available amount of symbols for some places it is not showed, until we delete enough symbols...

https://github.com/ihhub/fheroes2/assets/55348946/2b165fe6-4235-47f2-9874-3a389f97c1f6

And I agree with all points of @Districh-ru .

ihhub:
Hi @Branikolog , @Districh-ru and @oleg-derevenetz , I addressed all given comments. Please take a look.

Districh-ru (approved):
Works great!

Let's postpone polishing the virtual keyboard for the next release:
- fix the miss of the cursor "_" sign: изображение ;
- add the ability to place a cursor at any place of the string to add/remove characters there;
- implement a long press of a button (for example, to delete a phrase by holding the "<=]" button instead of pressing it multiple times);
- implement button double-click to separate Shift (only the next character entered will be uppercase) and Caps Lock;
- and make cursor blinking in all input fields (not only in Virtual Keyboard).

*edited*

oleg-derevenetz:
> fix the miss of the cursor "_" sign

What is the purpose of this cursor sign would be? Currently you can only add characters to the end of the string or remove them from the end, you cannot add or remove characters in the middle of the string for example.

Districh-ru:
> What would be the purpose of this cursor sign? Currently you can only add characters to the end of the string or remove them from the end, you cannot add or remove characters in the middle of the string for example.

If I press the Space button I don't see a feedback where the next character will be placed and I cannot check if I accidentally pressed Space many times before entering the next word.
изображение изображение

The cursor character do not always mean that it can be placed not only at the end of the string. :)
And it'll be good to implement the ability to place the cursor in any place to add/remove characters there.

Branikolog (approved):

Issue #1954, opened 2020-10-10
Combat, castle siege, water sprite not displayed for troops in the moat


LeHerosInconnu:
In the original game, when a troop is in the moat during siege, there is a sprite displayed on the bottom of the creature sprite to simulate the troop is in the water.
In fheroes2, the water sprite is not displayed for troops in the moat.

It's not perfect in the original game and it could be improved.

In original game:

![Moat H2 01](https://user-images.githubusercontent.com/43583994/95660321-5bfefe00-0b27-11eb-9842-efb2843338a4.png)

In fheroes2:

![Moat fh2 01](https://user-images.githubusercontent.com/43583994/95660323-60c3b200-0b27-11eb-88ae-26fd5614e981.png)

oleg-derevenetz:
I'm curious why this sprite was needed at all? Isn't it obvious from the context that unit is in the moat when it is in the moat?

LeHerosInconnu:
Hello @oleg-derevenetz and @zenseii,

This is a nice aesthetic touch in the original game that could be improved in fheroes2. ;)

In original game:

![In the moat 01](https://user-images.githubusercontent.com/43583994/206261599-9eba5d96-8899-46d7-80bd-c44ad58176be.png)

ihhub:
We should solve this issue algorithmically as the original game does this way imperfect.

PR #7157, opened 2023-05-10
Improved troops in moat drawing.

slawomir-witkowski:
fix: #1954 
image

### DIFF
--- a/src/fheroes2/battle/battle_arena.h
+++ b/src/fheroes2/battle/battle_arena.h
@@ -216,7 +216,8 @@ namespace Battle
             CASTLE_TOP_ARCHER_TOWER_POS = 19,
             CASTLE_BOTTOM_ARCHER_TOWER_POS = 85,
             CASTLE_TOP_GATE_TOWER_POS = 40,
-            CASTLE_BOTTOM_GATE_TOWER_POS = 62
+            CASTLE_BOTTOM_GATE_TOWER_POS = 62,
+            CASTLE_BRIDGE_POS = 49
         };
 
     private:
--- a/src/fheroes2/battle/battle_arena.h
+++ b/src/fheroes2/battle/battle_arena.h
@@ -208,6 +208,7 @@ namespace Battle
         enum
         {
             CATAPULT_POS = 77,
+            CASTLE_BRIDGE_POS = 49,
             CASTLE_GATE_POS = 50,
             CASTLE_FIRST_TOP_WALL_POS = 8,
             CASTLE_SECOND_TOP_WALL_POS = 29,
@@ -216,8 +217,7 @@ namespace Battle
             CASTLE_TOP_ARCHER_TOWER_POS = 19,
             CASTLE_BOTTOM_ARCHER_TOWER_POS = 85,
             CASTLE_TOP_GATE_TOWER_POS = 40,
-            CASTLE_BOTTOM_GATE_TOWER_POS = 62,
-            CASTLE_BRIDGE_POS = 49
+            CASTLE_BOTTOM_GATE_TOWER_POS = 62
         };
 
     private:
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1125,6 +1125,49 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
     }
 }
 
+void Battle::Interface::RedrawTroopMoat( const uint8_t moatCellId, const Unit & unit )
+{
+    if ( moatCellId ) {
+        // Do not draw moat fragment on bridge cell when bridge is down
+        if ( moatCellId == Arena::CASTLE_BRIDGE_POS ) {
+            const Bridge * bridge = Arena::GetBridge();
+            if ( bridge->isDown() ) {
+                return;
+            }
+        }
+
+        const Cell * moatCell = Board::GetCell( moatCellId );
+        assert( moatCell != nullptr );
+
+        const Unit * unitOnCell = moatCell->GetUnit();
+        if ( unitOnCell != &unit )
+            return;
+
+        // Draw small moat fragment at the bottom of the cell
+        const fheroes2::Rect & rect = moatCell->GetPos();
+        const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::MOATWHOL, 0 );
+        const int32_t inX = std::max( rect.x - sprite.x(), 0 );
+        const int32_t inY = rect.y - sprite.y() + 35;
+        const int32_t outX = rect.x + 1;
+        const int32_t outY = rect.y + 35;
+        const int32_t width = rect.width - 1;
+        const int32_t height = rect.height - 43;
+        fheroes2::Blit( sprite, inX, inY, _mainSurface, outX, outY, width, height );
+
+        const Settings & conf = Settings::Get();
+
+        // Apply shadow to moat fragment
+        if ( conf.BattleShowMoveShadow() && _currentUnit == &unit ) {
+            fheroes2::Blit( _hexagonCursorShadow, 0, 10, _mainSurface, outX, outY, width, height );
+        }
+
+        if ( conf.BattleShowMouseShadow() && _currentUnit && !( _currentUnit->GetCurrentControl() & CONTROL_AI )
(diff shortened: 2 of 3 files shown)


src/fheroes2/battle/battle_arena.h
```diff
@@ -216,7 +216,8 @@ namespace Battle
             CASTLE_TOP_ARCHER_TOWER_POS = 19,
             CASTLE_BOTTOM_ARCHER_TOWER_POS = 85,
             CASTLE_TOP_GATE_TOWER_POS = 40,
-            CASTLE_BOTTOM_GATE_TOWER_POS = 62
+            CASTLE_BOTTOM_GATE_TOWER_POS = 62,
+            CASTLE_BRIDGE_POS = 49
```
oleg-derevenetz, line 220: I recommend to place this value before the `CASTLE_GATE_POS` which is 50.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1125,6 +1125,49 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
     }
 }
 
+void Battle::Interface::RedrawTroopMoat( const uint8_t moatCellId, const Unit & unit )
+{
+    if ( moatCellId ) {
```
oleg-derevenetz, line 1130: I recommend to replace this by

```cpp
if ( moatCellId == 0 ) {
    return;
}
[...]
```

to reduce code nesting.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1376,6 +1419,8 @@ void Battle::Interface::RedrawArmies()
             Arena::CASTLE_GATE_POS,           Arena::CASTLE_BOTTOM_GATE_TOWER_POS, Arena::CASTLE_THIRD_TOP_WALL_POS,  Arena::CASTLE_BOTTOM_ARCHER_TOWER_POS,
             Arena::CASTLE_FOURTH_TOP_WALL_POS };
 
+    const uint8_t moatCellIds[ARENAH] = { 7, 19, 28, 39, 49, 61, 72, 84, 95 };
```
oleg-derevenetz, line 1422: In all other places, `int32_t` is used to represent the cell id and `-1` is used to represent the invalid/nonexistent/whatever cell id (because there is a valid cell with id `0`). While this logic is perfectly valid, it is somewhat out of the general series, so if there are no serious arguments for keeping it, I recommend doing "as everywhere". What do you think?

oleg-derevenetz (changes requested):
Hi @slawomir-witkowski I just left a few suggestions, please take a look when you have time.

oleg-derevenetz (approved):
LGTM

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1125,6 +1125,50 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
...
+    // Draw small moat fragment at the bottom of the cell
+    const fheroes2::Rect & rect = moatCell->GetPos();
+    const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::MOATWHOL, 0 );
+    const int32_t inX = std::max( rect.x - sprite.x(), 0 );
+    const int32_t inY = rect.y - sprite.y() + 35;
+    const int32_t outX = rect.x + 1;
+    const int32_t outY = rect.y + 35;
+    const int32_t width = rect.width - 1;
+    const int32_t height = rect.height - 43;
+    fheroes2::Blit( sprite, inX, inY, _mainSurface, outX, outY, width, height );
```
Districh-ru, line 1157: The grid is a little overlapped by the moat fragment:
изображение
The same if for the Bridge.
изображение

 It is not a regression of current PR, but now it becomes noticeable.
 master build: 
изображение

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1125,6 +1125,50 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
...
+    const Unit * unitOnCell = moatCell->GetUnit();
+    if ( unitOnCell != &unit )
+        return;
+
+    // Draw small moat fragment at the bottom of the cell
+    const fheroes2::Rect & rect = moatCell->GetPos();
+    const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::MOATWHOL, 0 );
+    const int32_t inX = std::max( rect.x - sprite.x(), 0 );
+    const int32_t inY = rect.y - sprite.y() + 35;
+    const int32_t outX = rect.x + 1;
```
Districh-ru, line 1153: The moat fragment is shifted by one pixel to the right.
fheroes2 2023-05-15 19-28-00-642

Also when troop is moving in the moat it moves on the water, not in it. :)

Districh-ru (changes requested):
Hi, @slawomir-witkowski.
Good work. I left 3 comments and one suggestion, could you take a look in your spare time.

Maybe we can cache the image from `_mainSurface` before rendering the troop by **RedrawTroopSprite()** with grid and shadow and use the cached image in **RedrawTroopMoat()** where we can get the troop sprite position and overlap only its legs in the moat: e.g. call `RedrawTroopMoat()` from `RedrawTroopSprite()` where we have the `drawnPosition` of troop sprite ... or return this Point from `RedrawTroopSprite()` function to use it in `RedrawTroopMoat()`.
What do you think?)

src/fheroes2/battle/battle_arena.h
```diff
@@ -208,6 +208,7 @@ namespace Battle
         enum
         {
             CATAPULT_POS = 77,
+            CASTLE_BRIDGE_POS = 49,
```
ihhub, line 211: Can we reuse an existing enumeration entry **CELL_MOAT = 49,**?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1125,6 +1125,50 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
...
+            return;
+        }
+    }
+
+    const Cell * moatCell = Board::GetCell( moatCellId );
+    assert( moatCell != nullptr );
+
+    const Unit * unitOnCell = moatCell->GetUnit();
+    if ( unitOnCell != &unit )
+        return;
```
ihhub, line 1146: Just to be consistent with the rest code: please add parenthesis for this if-body as it is done for lines 1136-1138.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1376,6 +1420,8 @@ void Battle::Interface::RedrawArmies()
             Arena::CASTLE_GATE_POS,           Arena::CASTLE_BOTTOM_GATE_TOWER_POS, Arena::CASTLE_THIRD_TOP_WALL_POS,  Arena::CASTLE_BOTTOM_ARCHER_TOWER_POS,
             Arena::CASTLE_FOURTH_TOP_WALL_POS };
 
+    const int32_t moatCellIds[ARENAH] = { 7, 19, 28, 39, 49, 61, 72, 84, 95 };
```
ihhub, line 1423: Could we please use std::array<int32_t> here as it does check the size of the array at initialization stage?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1125,6 +1125,50 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
...
+    const Cell * moatCell = Board::GetCell( moatCellId );
+    assert( moatCell != nullptr );
+
+    const Unit * unitOnCell = moatCell->GetUnit();
+    if ( unitOnCell != &unit )
+        return;
+
+    // Draw small moat fragment at the bottom of the cell
+    const fheroes2::Rect & rect = moatCell->GetPos();
+    const fheroes2::Sprite & sprite = fheroes2::AGG::GetICN( ICN::MOATWHOL, 0 );
```
ihhub, line 1150: The original image is made in a clumsy way as it is visible that this is a separate image. How about we change the logic of rendering the troops to be even better than in the original game? A simple algorithm to calculate the position of the unit and how much not to draw would solve the problem. What do you think?

ihhub (changes requested):
Hi @slawomir-witkowski , I left multiple comments and proposals. Would you mind to take a look at them in your spare time?

slawomir-witkowski:
Removed PR.

PR #7160, opened 2023-05-11
Android Toolset: minor cleanup

oleg-derevenetz:
* Enforce check for ongoing background operation in `ViewModel` methods to better prevent potential race conditions. Currently, this prevention is performed using only the UI logic (by disabling/enabling UI controls), so it's better to be safe than sorry;
* Removed some single-line methods and short private methods that are used only in one place.


ihhub:

PR #7166, opened 2023-05-13
Add Polish to Virtual Keyboard

ihhub:
Polish has extra 9 characters on top of English alphabet.
image

relates to #6924

### DIFF
--- a/src/fheroes2/gui/ui_keyboard.cpp
+++ b/src/fheroes2/gui/ui_keyboard.cpp
@@ -105,8 +105,14 @@ namespace
             resize( size );
         }
 
-        const fheroes2::Rect & getWindowRoi() const
+        fheroes2::Rect getWindowRoi() const
         {
+            if ( !_window ) {
+                // You are calling this method for an empty renderer!
+                assert( 0 );
+                return {};
+            }
+
             return _window->activeArea();
         }
 
@@ -131,7 +137,7 @@ namespace
 
             renderInputArea();
 
-            _output.render( getWindowRoi() );
+            _output.render( _window->activeArea() );
         }
 
         void appendCharacter( const char character )
--- a/src/fheroes2/gui/ui_keyboard.cpp
+++ b/src/fheroes2/gui/ui_keyboard.cpp
@@ -152,7 +152,7 @@ namespace
 
             _info += character;
 
-            renderInputArea();
+            _output.render( renderInputArea() );
         }
 
         void removeLastCharacter()
@@ -163,7 +163,7 @@ namespace
 
             _info.pop_back();
 
-            renderInputArea();
+            _output.render( renderInputArea() );
         }
 
     private:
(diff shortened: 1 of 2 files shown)


src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -79,27 +97,52 @@ namespace
...
+            _window.reset();
+
+            _window = std::make_unique<fheroes2::StandardWindow>( offset.x, offset.y, size.width, size.height, true, _output );
+
+            _window->render();
 
             renderInputArea();
 
-            _output.render( getWindowRoi() );
+            _output.render( _window->activeArea() );
```
Districh-ru, line 145: While resizing the keyboard window we do **render()** in `renderInputArea()` then here and then we add buttons and do a third render in `processVirtualKeyboardEvent()`.
So we get an extra frame without buttons and sometimes partially rendered:

https://github.com/ihhub/fheroes2/assets/113276641/634be282-2e56-4eea-9ebd-c13346018930

Districh-ru (changes requested):
Hi, @ihhub. I noticed one bug. Could you take a look when you have time.

src/fheroes2/gui/ui_keyboard.cpp
ihhub: Hi @Districh-ru , could you please check the latest commit? I think it should be fixed. I cannot reproduce it properly as my machine is to powerful for this :(

ihhub: @Districh-ru , on top of this I fixed another place where we were calling rendering for no reason.

Districh-ru: @ihhub
I checked the fix. Now there are no rendering bugs. 👍

Districh-ru (approved):
The Polish Keyboard works fine.

I have an idea to discuss:
Since our Polish virtual keyboard has the same 3 rows as the English one, it might be good to show it by default. And also there is no need to switch back to English as all English letters are present in the Polish keyboard.

oleg-derevenetz (approved):

Districh-ru:
Added diacritics to CP1250 characters used in Polish keyboard.
изображение

Branikolog:
Hi, @ihhub 

> The Polish Keyboard works fine.
> 
> I have an idea to discuss: Since our Polish virtual keyboard has the same 3 rows as the English one, it might be good to show it by default. And also there is no need to switch back to English as all English letters are present in the Polish keyboard.

That's a reasonable point from @Districh-ru 

This would help to avoid texture floating behind the buttons, when we switch between languages and the window becomes larger.
Also, in this case the keyboard itself could be aligned in the center of the screen, since now it is not and I'm not sure we should do this with two keyboards available.

ihhub:
@Districh-ru  and @Branikolog , we can do this outside this PR. Also don't forget about consistency which is important if we handle all other languages.

Districh-ru:
> @Districh-ru and @Branikolog , we can do this outside this PR. Also don't forget about consistency which is important if we handle all other languages.

Yes, we can do this in other PR when we have more suppurted keyboard languages and we can decide how to behave with the ones that have all English letters in their alphabet. I suggested out this idea as it is noticeable here that switching to Polish keyboard simply adds the upper row without changing others.

Branikolog (approved):

PR #7177, opened 2023-05-14
Add shadow to the hero meeting dialog

jasonbtiller:
The current strategy doesn't work, but I was hoping that I could get feedback on what I'm doing wrong. The changes I've made follow the pattern in Dialog::RecruitMonster but no shadow is drawn.

### DIFF
--- a/src/fheroes2/heroes/heroes_meeting.cpp
+++ b/src/fheroes2/heroes/heroes_meeting.cpp
@@ -29,6 +29,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <iostream>
 
 #include "agg_image.h"
 #include "army.h"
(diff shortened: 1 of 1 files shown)


Districh-ru:
Hi @jasonbtiller.
Thanks for the attempt to fix this issue.
Some ICN resources have many images and for dialog windows, like for the Recruit dialog, it has 2 images: '0' is the dialog background image and '1' is the shadow tor this dialog. But ICN::SWAPWIN has only one image and no shadow in game resources. You might want to analyze the game resources using the fheroes2 tools. These tools can be downloaded from any pre-release build with "(latest commit)" text in the header. For Windows x64 you can download fheroes2 tools here: https://github.com/ihhub/fheroes2/releases/tag/fheroes2-windows-x64-SDL2

You tried the approach from a different solution which cannot be applied here as there is no prepared image with shadows for this dialog in game resources.
To fix this issue you may generate window shadows like it is done for **StandardWindow()**:
https://github.com/ihhub/fheroes2/blob/166e38180b8eefbfd7ac150a127ed25c50803de7/src/fheroes2/gui/ui_window.cpp#L250-L266
or take shadows from some other game resources image.

And you are welcome to suggest your own solution. :)

jasonbtiller:
Thank you for your suggestion, @Districh-ru! I implemented your suggestion to draw the shadow "manually," although I admit I don't entirely understand what the code is doing. I've tested it manually under Linux and it appears to draw the shadow as expected.

jasonbtiller:
I see my submission failed a coding style check. A few were obvious, but I'm not sure that I agree with these and I would appreciate people's thoughts:

```
          // "left side" shadow
-        ApplyTransform( i, x-b  , y+b  , b  , 1    , 5 );
-        ApplyTransform( i, x-b  , y+b+1, 1  , h-2  , 5 );
-        ApplyTransform( i, x-b+1, y+b+1, b-1, 1    , 4 );
-        ApplyTransform( i, x-b+1, y+b+2, 1  , h-4  , 4 );
-        ApplyTransform( i, x-b+2, y+b+2, b-2, 1    , 3 );
-        ApplyTransform( i, x-b+2, y+b+3, 1  , h-6  , 3 );
-        ApplyTransform( i, x-b+3, y+b+3, b-3, h-b-3, 2 );
+        ApplyTransform( i, x - b, y + b, b, 1, 5 );
+        ApplyTransform( i, x - b, y + b + 1, 1, h - 2, 5 );
+        ApplyTransform( i, x - b + 1, y + b + 1, b - 1, 1, 4 );
+        ApplyTransform( i, x - b + 1, y + b + 2, 1, h - 4, 4 );
+        ApplyTransform( i, x - b + 2, y + b + 2, b - 2, 1, 3 );
+        ApplyTransform( i, x - b + 2, y + b + 3, 1, h - 6, 3 );
+        ApplyTransform( i, x - b + 3, y + b + 3, b - 3, h - b - 3, 2 );
```

I think the extra whitespace is helpful, especially when trying to understand how the various arguments are being transformed through the multiple calls. I want to make this section obvious with what's happening and how the arguments relate to each other. Is this a no-no?

Thanks for your feedback!

oleg-derevenetz:
Hi @jasonbtiller please follow the code style rules and fix the corresponding check. If you want to keep the alignment of function arguments in a "tabular" style, please use the `// clang-format off` / `// clang-format on` pairs (although it is not customary to use such formatting in this project), and in any case, please add spaces to arithmetic expressions (`x-b+1` -> `x - b + 1`). I'll request an additional review from @ihhub if you prefer to use custom formatting.

jasonbtiller:
I made the requested changes but I did except the ApplyTransform blocks from clang-format. I hope that's OK!

src/fheroes2/heroes/heroes_meeting.cpp
```diff
@@ -90,6 +91,47 @@ namespace
...
     }
+
+    /// Draw shadow around an area
+    ///
+    /// @note This is a good candidate for becoming a generic drawing
+    /// function available system-wide.
+    ///
+    /// @param i Image to draw on
+    /// @param box Draw shadow on the left and bottom of this area
+    void DrawShadow( fheroes2::Image & i, fheroes2::Rect box )
```
ihhub, line 102: Please create a separate function in **ui_tool.h|cpp** called **void DrawWindowShadow( fheroes2::Image & image, const fheroes2::Rect & roi, const int32_t borderSize )**. You can move the code present in `void StandardWindow::render()`, lines 251-266. There is not reason to create a set of separate variable just to improve code readability.

ihhub (changes requested):
Hi @jasonbtiller , I left a comment in regards of code structure. Could you please take a look?

ihhub:
Closing this PR as the original issue has been fixed in #9666

Issue #1893, opened 2020-10-03
Interface in "EVIL" mode, have a real "EVIL" interface


LeHerosInconnu:
In the original game, in "EVIL" interface mode, many screens and windows are not changed from the "GOOD" interface mode.
This makes the "EVIL" interface a bit "half-made".

I suggest fixing this in fheroes2 in order to have a real "EVIL" interface.
I will complete this post on this later.

Branikolog:
Hi, @LeHerosInconnu 
I think the background for Mage guild should also be gray:
![image](https://user-images.githubusercontent.com/55348946/141102908-90267152-dd90-4d63-9690-f61523659370.png)
Btw, could you, please, update issue, as we've fixed "MAX" button for split dialog?

LeHerosInconnu:
Hello @Branikolog,

> Hi, @LeHerosInconnu I think the background for Mage guild should also be gray: ![image](https://user-images.githubusercontent.com/55348946/141102908-90267152-dd90-4d63-9690-f61523659370.png) Btw, could you, please, update issue, as we've fixed "MAX" button for split dialog?

Absolutely. :)

Branikolog:
Also, I think we should make grayscale battle UI too! :)
![image](https://user-images.githubusercontent.com/55348946/141745989-ddd50d8d-a832-48da-a695-7bbeb4e4bce5.png)

LeHerosInconnu:
Hello @Branikolog,

> Also, I think we should make grayscale battle UI too! :)
> ![image](https://user-images.githubusercontent.com/55348946/141745989-ddd50d8d-a832-48da-a695-7bbeb4e4bce5.png)

Yes, everything that reminds the "GOOD" interface should be adapted. :)

ihhub:
Kingdom Overview dialog is rendered only in Good interface:
![image](https://github.com/ihhub/fheroes2/assets/19829520/746b30c3-a79a-46d8-ab54-6fb42e79085a)

LeHerosInconnu:
Hello @everyone,

There are other screens concerned in the [Evil interface.zip](https://github.com/ihhub/fheroes2/files/5322999/Evil.interface.zip) file from the original post. :)

Mr-Bajs:
Done
- [x] Mage Guild background
- [x] Theives Guild background
- [x] Scenario Information dialog
- [x] Creature recruit dialog
- [x] New Game menu & sub menus
- [x] Load Game menu  & sub menus
- [x] High score buttons

[...]

PR #7192, opened 2023-05-16
Add button shadow generation, rework Recruit dialog, add Evil interface to Recruit dialog.

### LINKED ISSUE #6670: Rework button shadow generation code
### Preliminary checks

We now generate buttons and these can be longer or shorter than the original ones. The problem is that shadows for buttons are often embedded in the background and not generated according to the size of the button.

The prime example is this below. Look at the shadow on the lower right part of the button that is missing:
image

We have code for generating buttons with shadows, but iirc this code assumes that there isn't already a shadow present, which in many cases throughout the assets is the case:
https://github.com/ihhub/fheroes2/blob/aeb57f0b546caffa171423475cd408b88c51d522/src/fheroes2/gui/ui_button.cpp#L454-L479

We need to solve two things:
1. How to add a shadow when there is already a partial shadow present. Clean the background or use a standard background for all dialogs?
2. Where to put the code for the buttons that need shadows.
3. Optional: Should we generate shadows for buttons that are on uniform backgrounds like this?:
image

### Additional info

This issue is necessary to solve before we can continue with adding new buttons as pointed out by @Branikolog here https://github.com/ihhub/fheroes2/issues/6391#issuecomment-1367933390

Districh-ru:
Close #6670, relates to https://github.com/ihhub/fheroes2/issues/1893
This PR adds button (and not only button) shadows generation code and recreates the Recruit dialog to add Evil interface to Recruit dialog and give the ability to rearrange the UI elements.
Recruit dialog

Recruit info


Branikolog:
Hi, @Districh-ru !
I've checked shadows for buttons (hope I haven't missed any).
Here's an issue I've found with upgrade arrows in recruit window:

https://github.com/ihhub/fheroes2/assets/55348946/42607560-5ea8-4a19-99c0-8ec4c2ee9e9f

Could you, please, look at this place?
Could you also point me which button shadow generation touches this PR exactly?

Districh-ru:
> Hi, @Districh-ru ! I've checked shadows for buttons (hope I haven't missed any). Here's an issue I've found with upgrade arrows in recruit window:

Hi, @Branikolog! Thanks for testing, I'll fix this issue soon.

> Could you, please, look at this place? Could you also point me which button shadow generation touches this PR exactly? I've looked through all possible buttons rendering and even found an issue (which is not related to current PR). :)

Only in the Recruit dialog all buttons has shadows, generated by a new algorithm.
When we will polish this PR and merge it, we can update button shadows for all other dialogs.

Districh-ru:
@Branikolog, the issue is fixed. Could you please take a look when you have time.

Also I want to mention that these shadows are also generated by the new algorithm: 
изображение изображение

изображение изображение

Branikolog (approved):
Now everything relating shadows looks great!
The only issue I've noticed is left yellow arrow being pressed has yellow line on the north-west part.
image

But this issue is valid for master branch too and I suppose it's related to this particular button generation.
I think a new issue should be opened, since these buttons require some work to show proper pressing animation. Now it's a bit incorrect.

src/fheroes2/gui/ui_button.cpp
```diff
@@ -156,6 +156,19 @@ namespace fheroes2
         return true;
     }
 
+    bool ButtonBase::drawShadow( Image & output ) const
+    {
+        if ( !isVisible() ) {
+            return false;
```
ihhub, line 162: We never check the return value. We should remove it then.

src/fheroes2/gui/ui_button.h
```diff
@@ -92,6 +92,9 @@ namespace fheroes2
 
         bool draw( Image & output = Display::instance() ) const; // will draw on screen by default
 
+        // Render a button shadow on output image. 'Display::instance()' is used by default.
+        bool drawShadow( Image & output = Display::instance() ) const;
```
ihhub, line 96: I don't think that we need to put this render specific method in the base class. I can see that we call this method only in one particular file where we know the input images for buttons. Shall we use this data and create just a separate helper function to handle the cases?

ihhub (changes requested):

src/fheroes2/dialog/dialog_recruit.cpp
```diff
@@ -43,185 +43,171 @@
...
+        const fheroes2::Sprite & originalBackground = fheroes2::AGG::GetICN( ICN::RECRBKG, 0 );
+        const fheroes2::Sprite & recruitWindowTitle = fheroes2::AGG::GetICN( ICN::BLDGXTRA, 0 );
+
+        fheroes2::Image recruitWindow( 132, 67 );
+        recruitWindow.reset();
+        fheroes2::Copy( recruitWindowTitle, 3, 58, recruitWindow, 3, 0, 63, 14 );
+        fheroes2::Copy( recruitWindowTitle, recruitWindowTitle.width() - 66, 58, recruitWindow, 66, 0, 63, 14 );
+        fheroes2::Copy( originalBackground, 138, 54, recruitWindow, 0, 0, 3, 63 );
+        fheroes2::Copy( originalBackground, 267, 54, recruitWindow, 129, 0, 3, 63 );
+        fheroes2::Copy( originalBackground, 138, 117, recruitWindow, 0, 63, 132, 4 );
```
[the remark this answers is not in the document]
Districh-ru: We Copy only the header and the borders of "Cost per troop" window part. The other part of the image needs to be transparent to make shadows and to Blit it to the recruit dialog, so we need to reset the transform layer to "1".

src/fheroes2/dialog/dialog_recruit.cpp
```diff
@@ -269,46 +259,48 @@ Troop Dialog::RecruitMonster( const Monster & monster0, uint32_t available, cons
...
-    const fheroes2::Rect pos( dialogOffset.x, dialogOffset.y, box.width(), box.height() );
+    const fheroes2::Rect windowActiveArea( window.activeArea() );
 
-    fheroes2::Blit( boxShadow, display, pos.x - BORDERWIDTH, pos.y + BORDERWIDTH );
+    const fheroes2::Sprite & originalBackground = fheroes2::AGG::GetICN( ICN::RECRBKG, 0 );
 
-    drawRecruitWindow( display, { pos.x, pos.y }, windowIcnId );
+    // Render the recruit count background from original recruit dialog ICN.
+    fheroes2::Image background( 68, 19 );
+    fheroes2::Copy( originalBackground, 134, 159, background, 0, 0, 68, 19 );
```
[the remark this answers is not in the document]
Districh-ru: 👍

src/fheroes2/gui/ui_button.cpp
Districh-ru: Sure)
I used the `bool` type for this method to match the other `draw...` methods of the class.

src/fheroes2/gui/ui_button.h
Districh-ru: Sure, we can make a helper to render button shadows.
I only wanted to make it easy to use in any other file for dialog button shadows render (for the initialized button just call a method).
Currently this method was called in one file, but in future we may want to make shadows for buttons in other files, e.g. for the Virtual Keyboard (it is not done in current PR not to complicate it):
изображение

If in future if it will be needed we may simply restore this method in a Button class.

src/engine/image.cpp
```diff
@@ -752,6 +754,97 @@ namespace fheroes2
         return out;
     }
 
+    void addSoftShadow( const Sprite & in, Image & out, const Point & outPos, const Point & shadowOffset )
+    {
+        const int32_t outWidth = out.width();
```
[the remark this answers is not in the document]
Districh-ru: Sure)

src/engine/image.cpp
```diff
@@ -752,6 +754,97 @@ namespace fheroes2
...
+        uint8_t * transformOut = out.transform();
+        uint8_t * imageOut = out.image() + outPos.x + shadowOffsetX + in.x() + static_cast<ptrdiff_t>( outPos.y + shadowOffsetY + in.y() ) * outWidth;
+
+        for ( int32_t y = 0; y < maxY; ++y ) {
+            const int32_t offsetY = y + shadowOffsetY;
+            for ( int32_t x = 0; x < maxX; ++x ) {
+                const int32_t offsetX = x + shadowOffsetX;
+                const int32_t inOffset = offsetX + offsetY * inWidth;
+                const bool isTransparent = ( ( offsetX < 0 ) || ( offsetY < 0 ) || ( offsetX >= inWidth ) || ( offsetY >= inHeight ) )
+                                           || ( isInNonSingleLayer && ( *( transformIn + inOffset ) == 1 ) );
```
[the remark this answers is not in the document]
Districh-ru: Done, now it looks better!

src/engine/image.h
```diff
@@ -192,6 +192,9 @@ namespace fheroes2
     // Generates a new image with a shadow of the shape of existing image. Shadow must have only (-x, +y) offset.
     Sprite addShadow( const Sprite & in, const Point & shadowOffset, const uint8_t transformId );
 
+    // Apply shadow that softly reduces strength to edges on the 'out' image using 'in' image shape.
+    void addSoftShadow( const Sprite & in, Image & out, const Point & outPos, const Point & shadowOffset );
```
ihhub, line 196: `Softly reduces` might not be understandable for many. How about **addGradientShadow**?

src/engine/image.cpp
```diff
@@ -752,6 +754,126 @@ namespace fheroes2
...
+
+                        if ( transformTableId == 2 ) {
+                            // We reached the strongest shadow table ID.
+                            break;
+                        }
+                    }
+                }
+
+                // If the shadow has to be applied.
+                if ( transformTableId < 6 ) {
```
ihhub, line 848: Shall we add the opposite check to reduce code complexity:
```cpp
if ( transformTableId == 6 ) {
    continue;
}
```
What do you think?

src/fheroes2/dialog/dialog_recruit.cpp
```diff
@@ -349,22 +387,35 @@ Troop Dialog::RecruitMonster( const Monster & monster0, uint32_t available, cons
...
-    monsterSwitchLeft.draw();
-    monsterSwitchRight.draw();
 
-    display.render( back.rect() );
+    display.render( roi );
+
+    const fheroes2::Rect monsterArea( dialogOffset.x + 24, dialogOffset.y + 19, 75, 95 );
+
+    auto buttonReleaseRestore = [&display, &background, &dialogOffset]( fheroes2::ButtonBase & button ) {
+        if ( !button.isPressed() ) {
```
ihhub, line 407: **isReleased()** does the same :)

src/fheroes2/dialog/dialog_recruit.cpp
```diff
@@ -43,185 +43,181 @@
...
 #include "ui_dialog.h"
 #include "ui_text.h"
 #include "ui_tool.h"
+#include "ui_window.h"
 #include "world.h"
 
 namespace
 {
-    void drawRecruitWindow( fheroes2::Image & output, const fheroes2::Point & offset, const int icnId )
+    void drawButtonShadow( fheroes2::Image & output, const int buttonIcn, const uint32_t icnIndex, const fheroes2::Point offset )
```
ihhub, line 62: **offset** input argument should be const reference instead of const value.

src/fheroes2/gui/ui_window.h
```diff
@@ -45,12 +45,18 @@ namespace fheroes2
             return _windowArea;
         }
 
+        const Rect & windowWithShadowArea() const
+        {
+            return _windowWithShadowArea;
+        }
```
ihhub, line 51: Shall we call it more generic like **totalArea** (not the great name)? It doesn't matter whether it is with or without shadow. What do you think?

ihhub (changes requested):
Hi @Districh-ru , I left few more tiny suggestions. Please take a look once you have time.

src/engine/image.h
Districh-ru: Hi, @ihhub.
:))
Yes, it would be better to name it Gradient (no Soft) shadow.

src/fheroes2/dialog/dialog_recruit.cpp
Districh-ru: Oops, missed it.
Thanks for your Eagle Eye skill!

Districh-ru:
> The only issue I've noticed is left yellow arrow being pressed has yellow line on the north-west part. upg arrow

Hi, @Branikolog, if forgot to say that it is fixed now. :)
fheroes2 2023-05-26 20-06-11-575
But these left/right arrow needs to be slightly redone in future as they are simply rotated up/down arrows and after the rotation they have incorrect shading.

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for the changes!

Branikolog:
@Districh-ru 

> Hi, @Branikolog, if forgot to say that it is fixed now. :)

Glad to hear that! :)

> But these left/right arrow needs to be slightly redone in future as they are simply rotated up/down arrows and after the rotation they have incorrect shading.

That I was talking about here:

> I think a new issue should be opened, since these buttons require some work to show proper pressing animation. Now it's a bit incorrect.

PR #7200, opened 2023-05-18
Fix some critical issues detected by SonarQube

ihhub:
My eyes cannot accept bugs detected by SonarQube.

### DIFF
--- a/src/engine/tinyconfig.cpp
+++ b/src/engine/tinyconfig.cpp
@@ -131,7 +131,8 @@ fheroes2::Point TinyConfig::PointParams( const std::string & key, const fheroes2
 
     fheroes2::Point result;
 
-    static_assert( std::is_integral<decltype( result.x )>::value, "The type of result fields is not integer, check the logic of this method" );
+    static_assert( std::is_integral<decltype( result.x )>::value && std::is_integral<decltype( result.y )>::value,
+                   "The type of result fields is not integer, check the logic of this method" );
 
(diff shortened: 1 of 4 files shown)


src/engine/tinyconfig.cpp
```diff
@@ -131,8 +131,7 @@ fheroes2::Point TinyConfig::PointParams( const std::string & key, const fheroes2
 
     fheroes2::Point result;
 
-    static_assert( std::is_integral<decltype( result.x )>::value && std::is_integral<decltype( result.y )>::value,
-                   "The type of result fields is not integer, check the logic of this method" );
+    static_assert( std::is_integral<decltype( result.x )>::value, "The type of result fields is not integer, check the logic of this method" );
```
oleg-derevenetz, line 134: Hi @ihhub the whole purpose of this `static_assert()` is to detect the situation when either `x` or `y` is not of integral type :) This SonarCloud issue is plain wrong, and I believe that it should be closed as a false positive and not "fixed" this way - by reducing the protection provided by the `static_assert()`. Now we know that they are of the same integral type, but this logic is designed to protect against potential errors in the future if the programmer accidentally makes one or both of them not integer types, because then he will need to change the regular expression.

ihhub: I marked it as false positive. I'll revert this change.

oleg-derevenetz (approved):

PR #7204, opened 2023-05-20
Some refactorings in border drawing code.

KOLANICH:


src/fheroes2/gui/interface_border.cpp
```diff
@@ -119,301 +119,240 @@ void Interface::GameBorderRedraw( const bool viewWorldMode )
...
+        fheroes2::Blit( icnadv, srcrt.x, srcrt.y, display, dstpt.x, dstpt.y, srcrt.width, srcrt.height );
+
+        if ( !viewWorldMode ) {
+            // bottom border of heroes and towns area, top border of resources area
+            dstpt.y = srcrt.y + BORDERWIDTH + iconsCount * 32;
+            srcrt.y = srcrt.y + BORDERWIDTH + 4 * 32;
+            fheroes2::Blit( icnadv, srcrt.x, srcrt.y, display, dstpt.x, dstpt.y, srcrt.width, srcrt.height );
+        }
+    };
+    drawHeroesAndTownsHorizontalBorders();
```
ihhub, line 355: Could you please explain the benefit of having a lambda-function and a single call of it just right its declaration? Wouldn't it be faster to just have normal code?

src/fheroes2/gui/interface_border.cpp
```diff
@@ -119,301 +119,240 @@ void Interface::GameBorderRedraw( const bool viewWorldMode )
...
-
-    dstpt.y += srcrt.height;
-    srcrt.y += srcrt.height;
-
-    if ( isEvilInterface ) {
-        srcrt.height = 6;
-        repeatPattern( icnadv, srcrt.x, srcrt.y, srcrt.width, srcrt.height, display, dstpt.x, dstpt.y, BORDERWIDTH, srcrt.height + vertPadHeight );
-        dstpt.y += srcrt.height + vertPadHeight;
-        srcrt.y += srcrt.height;
+    auto drawLeftBorder = [&]() {
```
ihhub, line 175: The same comment as for **drawHeroesAndTownsHorizontalBorders**: is there a benefit of having a lambda function and calling it only once? It does not solve code duplication issues.

src/fheroes2/gui/interface_border.cpp
```diff
@@ -119,301 +119,240 @@ void Interface::GameBorderRedraw( const bool viewWorldMode )
...
-
-    if ( isEvilInterface ) {
-        srcrt.height = TILEWIDTH;
-        repeatPattern( icnadv, srcrt.x, srcrt.y, srcrt.width, srcrt.height, display, dstpt.x, dstpt.y, BORDERWIDTH, vertRepeatHeightTop );
-        dstpt.y += vertRepeatHeightTop;
-        srcrt.y += TILEWIDTH;
-
-        srcrt.width = BORDERWIDTH;
-        srcrt.height = 35;
+    auto drawHorizontalBorder = [&]( bool isTop ) {
```
ihhub, line 122: **isTop** can be **const** as we don't modify it within the function.

src/fheroes2/gui/interface_border.cpp
```diff
@@ -119,301 +119,240 @@ void Interface::GameBorderRedraw( const bool viewWorldMode )
...
-        dstpt.y += srcrt.height + vertPadHeight;
-        srcrt.y += srcrt.height;
-
-        srcrt.height = 103;
+        dstpt.x += srcrt.width;
+        srcrt.x += srcrt.width;
+
+        // transitioning line to the left block
+        srcrt.width = 6;
+        auto PadWidthLeft = isTop ? topPadWidthLeft : bottomPadWidthLeft;
```
ihhub, line 136: Please use camelCase variable naming as well as avoid using **auto** for plain old data types. Moreover, this variable can be const.

ihhub (changes requested):
Hi @KOLANICH , I put several comments in this pull request. Could you please take a look at them? Also as I understood correctly it is just code cleanup, isn't it?

src/fheroes2/gui/interface_border.cpp
KOLANICH: Fixed.

src/fheroes2/gui/interface_border.cpp
KOLANICH: Fixed.

src/fheroes2/gui/interface_border.cpp
oleg-derevenetz: Then there is no need to give the lambda a name at all. You can just do this:

```cpp
  [](){
      [... do something ...]
  }();
```

Also, if you want to avoid name conflicts / reuse the same variable names in different blocks / manage variable scopes for RAII, then you can just use code blocks:

```cpp
{
  std::scoped_lock lock{m1};
  [... do something ...]
}

{
  std::scoped_lock lock{m1, m2};
  [... do something else...]
}
```

Lambdas for the purposes of code organization alone look redundant. Lambdas are useful mainly for reusable code, or maybe some complex code that needs to initialize a `static` or `const`/`constexpr` variable.

KOLANICH: >Then there is no need to give the lambda a name

Of course there is a need. First it serves the purpose of documenting. So called "self-documenting code". Second, it can later be promoted to a non-lambda, leaving just a sequence of function calls within that method. If I wanted to have it anonymous + some wrapping, just a code block of `{}` without any lambda should have been OK.

oleg-derevenetz: > Of course there is a need. First it serves the purpose of documenting. So called "self-documenting code". Second, it can later be promoted to a non-lambda, leaving just a sequence of function calls within that method.

It looks like over-engineering for a "future" that can never come.

ihhub:
I am closing this pull request as it brings no value to the code.

PR #7205, opened 2023-05-20
Fix Russian translation: a boar is not a barrow.

KOLANICH:


Districh-ru (changes requested):
Hi @KOLANICH.
In English-Russian Dictionary (by V.D. Arakin, Z.S. Vigodskaya, N.N. Iljina) Boar is "боров" or "кабан".
изображение

Google thinks the same: https://translate.google.ru/?sl=en&tl=ru&text=boar&op=translate

Lingvo suggests to translate as "хряк". :)

In original Buka version it is translated as "Боров".
изображение

So lets keep "Боров" without changes.

KOLANICH:
"Buka" translation is not authoritative. Almost universally translators do translation in a poor way called "надмозг" when they not do their best to keep the original meaning, but instead try to be creative and destroy the original meaning by replacing it with own one. It doesn't mean we should do that too.

I guess in the case of Buka they saw that the word `boar` is phonetically similar to the Russian translation of the word `barrow` and so used it. It doesn't mean that their translation is correct. A `barrow` is not just a `boar`, but a **castrated** `boar`, the meaning that originally clearly wasn't present.

My position is that a translation should be as "accurate" (keeping the original meaning) as possible, even if it limits the ways for a translator to enjoy his "creativity".

Districh-ru (commented):
> "Buka" translation is not authoritative. Almost universally translators do translation in a poor way called "надмозг" when they not do their best to keep the original meaning, but instead try to be creative and destroy the original meaning by replacing it with own one. It doesn't mean we should do that too.

Yes, Fargus and others made many poor/creepy translations.

> I guess in the case of Buka they saw that the word `boar` is phonetically similar to the Russian translation of the word `barrow` and so used it. It doesn't mean that their translation is correct. A `barrow` is not just a `boar`, but a **castrated** `boar`, the meaning that originally clearly wasn't present.

And talking about translation of the word `barrow` - it has nothing common with boar. :)
изображение

> My position is that a translation should be as "accurate" (keeping the original meaning) as possible, even if it limits the ways for a translator to enjoy his "creativity".

@KOLANICH, I agree with your position.👍 And thanks for polishing the engine code and the translation!

We can rename "Боров" to "Вепрь" to avoid connection with the castrated boar.
@Branikolog what is your opinion about this?

KOLANICH:
>Yes, Fargus and others made many poor/creepy translations.

I didn't mean Fargus or GTA, I meant, for example, the translations of works by Tolkien by Murav'ev and Kist'akovsky.

>And talking about translation of the word barrow - it has nothing common with boar. :)

Paper dictionary references mean nothing for me, I don't use them for many years :)

https://en.wiktionary.org/wiki/barrow#Etymology_3
>Noun barrow (plural barrows) (obsolete except in scientific use and in some dialects) A castrated [boar](https://en.wiktionary.org/wiki/boar).

https://bigtex.com/livestock-terminology/
>Barrow – a young, castrated male swine

https://ru.wiktionary.org/wiki/боров
>кастрированный самец свиньи

KOLANICH:
Since the official policy of this project is to match the "translation" by Buka, [here is the precompiled translation file for the ones who want it](https://github.com/ihhub/fheroes2/files/11554817/ru.zip), where boars are not castrated.

PR #7206, opened 2023-05-20
Translation ru: Correct the translation of `halfling` as `полурослик`, not `hobbit` (`хоббит`).

KOLANICH:
These are distinct words and they shouldn't be used one instead of another for the translation being accurate.

### DIFF
--- a/files/lang/ru.po
+++ b/files/lang/ru.po
@@ -2345,7 +2345,7 @@ msgstr ""
 "заклинание для каждого уровня."
 
 msgid "The Orchard increases production of Halflings by %{count} per week."
-msgstr "Фруктовый сад увеличивает прирост Хоббитов на %{count} в неделю."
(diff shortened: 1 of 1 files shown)


Districh-ru (approved):

Branikolog:
Hi, @KOLANICH !
Definitely "полурослик" is a correct translation for halfling. Official "Buka" localization also had the correct name for this creature.
Since we're changing creature name, could you, please, check, if there are other places where "хоббиты" could appear in game translation? Like artifact descriptions or campaign briefing texts?

KOLANICH:
```bash
rg -i хобб ./files/lang/
```

- nothing found.

Branikolog (approved):
Thanks, @KOLANICH !
This is definitely the place that required a fix!

ihhub:
@KOLANICH , thank you so much for this translation correction!

Issue #7176, opened 2023-05-14
During battle, the moving animation is jerky


kazenoyouni:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

PlayStation Vita

### Describe the bug

During battle, the moving animation is jerky, especially flying creatures. It was smooth until 1.0.2, but from 1.0.3 it became like this. And the latest 1.0.4 also has this problem.

### Save file

This happens in all battles.

oleg-derevenetz:
Hi @kazenoyouni could you please attach a video of these animations, if possible? Also, does this depend on the animation speed, and, if yes, on which speed it happens?

kazenoyouni:
1.0.2
https://github.com/ihhub/fheroes2/assets/111102725/125b0479-0dab-499a-a6a6-ac658d2366d6

1.0.4 a
https://github.com/ihhub/fheroes2/assets/111102725/6ae0dde6-781a-4ac6-9118-92265a565ad4

1.0.4 b
https://github.com/ihhub/fheroes2/assets/111102725/7f65eb89-b145-4d9e-9f0b-ca6a1aefed53

Hi @oleg-derevenetz ,
It is more noticeable at speed 8-10. I usually play the game at speed 10.
Speed 7 or less is difficult to understand because the animation itself is slow.

oleg-derevenetz:
Hi @Districh-ru what do you think about this issue? You're more of a combat animation expert than I am. At least Cyclops in the third video really moves jerkily, it's quite noticeable.

Districh-ru:
Hi @oleg-derevenetz and @kazenoyouni. Yes, the movement of creatures looks jerky on v. 1.0.4 videos.
I tried to reproduce it on my Android devices, but could not. Also I could not find any changes in creatures moving animation after v. 1.0.2.
Only just before the v. 1.0.2 release we made a change in animation delays for the "10" battle speed.
I'll analyze the code of creatures movement animation for possible slowdowns.

[...]

Issue #3112, opened 2021-04-06
What are minimal platform-independent system requirements of this?


KOLANICH:
I mean size of RAM.

ihhub:
Hi @KOLANICH , it really depends on a resolution. What's your resolution?

KOLANICH:
480x272

ihhub:
The minimal resolution for the game is 640 x 480. Right now it requires around 60 MB RAM.

KOLANICH:
>The minimal resolution for the game is 640 x 480. Right now it requires around 60 MB RAM.

So it won't fit PSP-1xxx ("Thick") in its current state neither by memory nor by screen.

I wonder if there is any model allowing one to predict memory consumption based on screen size. I guess it may be something like `const not dependent on resolution + video buffer + f(sprites fitting the screen)`. Can it be built by compiling the game with dufferent resolutions, running against a benchmark file and fitting the coefficients?

ihhub:
Hi @KOLANICH , we do not have any formula for this as we don't have need to do it.

ihhub:
I'm closing this issue as the primary question was answered.

PR #7208, opened 2023-05-20
Optimize CPU load while Battlefield rendering

Districh-ru:
This PR relates to #7176 issue and optimizes the Battlefield render:
- render Battlefield background only when it is changed (grid on/off) and cache it;
- use `fheroes2::Copy()` instead of `fheroes2::Blit()` where possible:
  - when preparing the Battlefield background;
  - on every render when drawing the rectangle images without transparency (count background, status bar background, ...);
  - during DeathWave and EarthQuake spell effects.

Full render time for the Castle battlefield: 1.5 -> 0.85 ms
Full render time for the "Ship" battlefield, used in Battle only" mode: 1.0 -> 0.6 ms. 
<details><summary>Render time for optimized parts for the "Ship" battlefield</summary> 

```
RedrawCoverStatic render time:   0.2916 ms. -> 0.0443 ms.
RedrawPartialFinish render time: 0.1684 ms. -> 0.0142 ms.
RedrawInterface render time:     0.0159 ms. -> 0.0052 ms.
```

</details>

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1261,7 +1261,7 @@ Battle::Interface::Interface( Arena & battleArena, const int32_t tileIndex )
     _mainSurface.resize( area.width, battlefieldHeight );
     _battleGround.resize( area.width, battlefieldHeight );
 
-    // As `_battleGround` and '_mainSurface' are used to prepare battlefield screen to render on display they does not need to have a transform layer.
+    // As `_battleGround` and '_mainSurface' are used to prepare battlefield screen to render on display they do not need to have a transform layer.
     _battleGround._disableTransformLayer();
     _mainSurface._disableTransformLayer();
 
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1131,7 +1131,6 @@ void Battle::ArmiesOrder::Redraw( const Unit * current, const uint8_t currentUni
 Battle::Interface::Interface( Arena & battleArena, const int32_t tileIndex )
     : arena( battleArena )
     , _surfaceInnerArea( 0, 0, fheroes2::Display::DEFAULT_WIDTH, fheroes2::Display::DEFAULT_HEIGHT )
-    , _mainSurface( fheroes2::Display::DEFAULT_WIDTH, fheroes2::Display::DEFAULT_HEIGHT )
     , icn_cbkg( ICN::UNKNOWN )
     , icn_frng( ICN::UNKNOWN )
     , humanturn_spell( Spell::NONE )
@@ -1226,9 +1225,6 @@ Battle::Interface::Interface( Arena & battleArena, const int32_t tileIndex )
     // Shadow that fits the hexagon grid.
     _hexagonGridShadow = DrawHexagonShadow( 4, 1 );
 
-    // As '_mainSurface' is used to prepare battlefield screen to render on display it does not need to have a transform layer.
-    _mainSurface._disableTransformLayer();
-
     btn_auto.setICNInfo( ICN::TEXTBAR, 4, 5 );
     btn_settings.setICNInfo( ICN::TEXTBAR, 6, 7 );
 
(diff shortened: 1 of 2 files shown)


Districh-ru:
Hi @kazenoyouni.
I made some minor optimizations in battle rendering. If you have time could you please check is the CPU load reduced in this PR on PlayStation Vita hardware. Here is the [compiled PS Vita binary.](https://github.com/ihhub/fheroes2/suites/13040703670/artifacts/706814327)

kazenoyouni:
Hi @Districh-ru.
It works perfectly. CPU load reduced to 30%~40%!

https://github.com/ihhub/fheroes2/assets/111102725/b5246171-9b3d-46bf-ae5f-6198989e06d8

Districh-ru:
> Hi @Districh-ru. It works perfectly. CPU load reduced to 30%~40%!

Great! @kazenoyouni, thank you for testing. If you find some other places where fheroes2 is heavily CPU intensive, please write.

kazenoyouni:
Hi, @Districh-ru.
Duly noted. I will do that.
Thank you very much.

KOLANICH:
It increases memory footprint (while, I guess, the goal should be to decrease memory footprint, maybe even at cost of frame rendering time, it is a turn-based strategy, not a FPS), doesn't it? I wonder if it can be rerendered on each frames, but not by using software raster, but using OpenGL. Hexagons seem to be simple enough shapes to generate triangulation for them on fly and simple enough not to need a texture.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1230,27 +1232,42 @@ Battle::Interface::Interface( Arena & battleArena, const int32_t tileIndex )
...
+    if ( listlog ) {
+        listlog->SetPosition( area.x, area.y + battlefieldHeight );
+    }
     status.SetLogs( listlog.get() );
 
+    // Battlefield area excludes the lower part where the status log is located.
+    _mainSurface.resize( area.width, battlefieldHeight );
+    _battleGround.resize( area.width, battlefieldHeight );
+
+    // As `_battleGround` and '_mainSurface' are used to prepare battlefield screen to render on display they does not need to have a transform layer.
```
ihhub, line 1264: Spelling issue: they **do** not (plurar form).

ihhub (commented):
Hi @Districh-ru , I left one spelling correction here. Could you please check it?

src/fheroes2/battle/battle_interface.cpp
Districh-ru:
```suggestion
    // As `_battleGround` and '_mainSurface' are used to prepare battlefield screen to render on display they do not need to have a transform layer.
```

Districh-ru: Hi @ihhub, thanks for correcting!)

ihhub:
Hi @Districh-ru , please resolve merging conflicts for this pull request.

Districh-ru:
> It increases memory footprint (while, I guess, the goal should be to decrease memory footprint, maybe even at cost of frame rendering time, it is a turn-based strategy, not a FPS), doesn't it? I wonder if it can be rerendered on each frames, but not by using software raster, but using OpenGL and rendering it on GPU. Hexagons seem to be simple enough shapes to generate triangulation for them on fly and simple enough not to need a texture.

Hi, @KOLANICH, yes this PR increases the memory footprint by 553.75 Kbytes. This is a cost for about 2 times less CPU usage. It is important for the handheld devices as it reduces the battery usage as long as for devices with low CPU (some old devices).
This PR does not only cache the hexagon grig. It aldo caches: ground, objects at the right and left parts of battlefield, big and small obstacles. These parts are never changed (except the rare switch of grid render).

ihhub:
Hi @KOLANICH , we can't simply use any rendering technology out of the box as fheroes2 contains its own rendering engine and we call SDL (which is essentially a wrapper over DirectX or OpenGL) just to render already prepared frame. It is possible to use GPU memory and update data there but it requires a lot of code restructure and also data transfer between CPU and GPU is a bottleneck for most of operations on GPU especially when GPU doesn't support more than one DMA engines.

Talking about memory footprint: 0.5 MB is nothing in terms of memory. If we need we can just deallocate certain ICN images if we are hitting memory limit.

KOLANICH:
>0.5 MB is nothing in terms of memory

It is a lot in terms of memory for such a device as Sony PSP. Which has 32 MiB of RAM, while for the original game in minimal system requirements it is written 8 MiB (and I guess it is not for the game alone, but for the typical system whole, a game developer cannot task consumers with measuring free memory on their systems, but everyone used to know total physical memory on their systems) and 16 MiB as recommended. According to #3112 this game alone requires 60 MiB of RAM, I guess on a 64-bit machine, then on 32-bit one it should require ~30% less. Also originally it didn't require a 3D-capable GPU, but PSP has a 3D-capable GPU with 2 MiB of RAM. I guess that PSP is more powerful than the computers the original game was intended to run on ... and still we cannot even think about porting this game to PSP before memory issue is resolved.

This is where the mindset "memory is not scarce, let's waste it!" (instead of the mindset "memory is always scarce", which is more aligned to reality) has lead us and the whole software industry (to the point that a modern browser cannot handle 10 tabs on a 2 GiB RAM 64-bit system and Xorg session is OOM-crashes (with taking all the active programs with itself) even on 8 GiB system, let alone 2 GiB one where it happens even more frequently (and happenned during I was writing this message) ) to.

ihhub:
@KOLANICH and @Districh-ru , if memory usage is a concern we can modify the whole rendering engine to properly handle single-layers images and do not store double memory for them as we do. In this case we can release a lot of memory, not only 0.5 MB for this case. The proposed approach uses a single layer image but only half of the allocated memory is utilized.

ihhub:
Another area where memory optimization and CPU is easy to do is **ImageRestorer** especially when it is called for a single-layered image. We can store just half of the image if possible.

KOLANICH:
Thanks for the info.

ihhub:
@Districh-ru and @KOLANICH , this pull request increases memory usage not by 553.75 Kbytes but by 519,680 bytes which is 507.5 KB. The main surface before the changes was 640 x 480 pixels while right now it is 640 x 443 pixels. However, we have 2 surfaces with double layer memory allocation.

ihhub (approved):

oleg-derevenetz (commented):
Hi @Districh-ru I just left one question, could you please take a look when you have time?

Districh-ru:
> @KOLANICH and @Districh-ru , if memory usage is a concern we can modify the whole rendering engine to properly handle single-layers images and do not store double memory for them as we do. In this case we can release a lot of memory, not only 0.5 MB for this case. The proposed approach uses a single layer image but only half of the allocated memory is utilized.

I already started this modification and will open a PR soon.

src/fheroes2/battle/battle_interface.h
```diff
@@ -332,7 +332,8 @@ namespace Battle
         void HumanCastSpellTurn( const Unit & /* unused */, Actions & actions, std::string & msg );
 
         void RedrawCover();
-        void RedrawCoverStatic( const Settings & conf, const Board & board );
+        void _redrawBattleGround( const Settings & conf );
+        void _redrawCoverStatic( const Settings & conf );
```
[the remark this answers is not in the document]
Districh-ru: Hi @oleg-derevenetz, it was left from the previous realization of `RedrawCoverStatic()` method. You are right that we can simply access Settings in these methods.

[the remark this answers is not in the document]
Districh-ru: I made the appropriate changes. Thanks for pointing this out!

oleg-derevenetz (approved):

ihhub:
@Districh-ru , a bunch of thanks for this optimization!

PR #7212, opened 2023-05-21
Fix cases when AI ignores some objects during path evaluation

ihhub:
std::binary_search works only if an array is partially sorted in respect to value - https://en.cppreference.com/w/cpp/algorithm/binary_search

~~Since **getObjectsOnTheWay()** method returns unsorted array we have to sort it in order to fulfill binary search requirements.~~
**_mapObjects** wasn't been completely sorted during fog revealing events.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1673,6 +1673,8 @@ namespace AI
             if ( !isDimensionDoor ) {
                 // Dimension door path does not include any objects on the way.
                 std::vector<IndexObject> list = _pathfinder.getObjectsOnTheWay( destination );
+                std::sort( list.begin(), list.end() );
+
                 for ( IndexObject & pair : list ) {
                     if ( objectValidator.isValid( pair.first ) && std::binary_search( _mapObjects.begin(), _mapObjects.end(), pair ) ) {
                         const double extraValue = valueStorage.value( pair, 0 ); // object is on the way, we don't loose any movement points.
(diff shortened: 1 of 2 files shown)


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1673,6 +1673,8 @@ namespace AI
             if ( !isDimensionDoor ) {
                 // Dimension door path does not include any objects on the way.
                 std::vector<IndexObject> list = _pathfinder.getObjectsOnTheWay( destination );
+                std::sort( list.begin(), list.end() );
```
oleg-derevenetz, line 1676: Hi @ihhub but there is `_mapObjects` should be sorted, and not the `list`.

ihhub: Woops... I am completed blind.

ihhub: Boom! Even better. **_mapObjects** can be unsorted during fog revealing.

oleg-derevenetz: > Boom! Even better. _mapObjects can be unsorted during fog revealing.

Yes.

oleg-derevenetz (changes requested):
Hi @ihhub I recommend also to add the assertion that `_mapObjects` is `std::is_sorted` somewhere near that `std::binary_search` (maybe only for the debug build?) and also to add the comment why we are using the `upper_bound` here.

oleg-derevenetz (approved):

Districh-ru (approved):
I tested AI on two XL maps, one time with debug hero (with Dimension Door spell) for a half an hour and found no bugs. Also I did not notice strange AI behavior.

PR #7216, opened 2023-05-22
Use an older version of the Android Gradle Plugin

### LINKED ISSUE #7187: (Android) 1.0.4 crashes at start. 1.0.3 works fine.
### Preliminary checks

### Platform

Android

### Describe the bug

(Android) 1.0.4 crashes at start. 1.0.3 works fine. Xperia T2 Ultra (Android 5.1.1).

### Save file

Just crashes on start.

### Additional info

_No response_

oleg-derevenetz:
fix #7187

https://issuetracker.google.com/issues/279034662
https://issuetracker.google.com/issues/282544786

This issue was introduced in the Android Gradle Plugin 8.0 and should be fixed in 8.1 (currently in beta, will be available along with the new version of Android Studio). As a workaround, I rolled back to an older AGP version (7.4.2). With older AGP version, there are warnings during the desugaring stage:

```
Warning: Desugared library: Cannot amend library reference java.nio.file.Path java.nio.file.Path.of(java.net.URI) because the holder is not a library class.
Warning: Desugared library: Cannot amend library reference java.nio.file.Path java.nio.file.Path.of(java.lang.String, java.lang.String[]) because the holder is not a library class.
```

but it seems that everything works fine both in debug and release builds on Android 6 tablet that I temporarily borrowed.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -80,8 +80,7 @@ android {
 }
 
 dependencies {
-    // TODO: version 2.0.3 does not emulate java.nio properly on Android 7 and below
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs_nio:2.0.2'
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs_nio:2.0.3'
 
     implementation project(':isotools')
     implementation project(':sdl2')
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,7 +7,9 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.0.1'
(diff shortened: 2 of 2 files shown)


ihhub (approved):

ihhub:
Hi @oleg-derevenetz , I approve it and I saw the reply from Google. Before merging we need to verify whether it truly fixes the problem for users.

oleg-derevenetz:
@ihhub

> Before merging we need to verify whether it truly fixes the problem for users.

The original problem manifests itself only on release builds. So either someone will build the release (with his own signature) and run it on some pre-Android 8 device (I did it on Android 6 device and it works fine), or we merge it and people (e.g. @raidoraido) will be able to try the release build from the latest snapshot signed by us.

ihhub:
@oleg-derevenetz , then we shall merge :) Thank you very much!

PR #7183, opened 2023-05-14
MusicalEffectPlayer: do not change music volume if there is no music file for the corresponding effect, reduce I/O when searching for music files

oleg-derevenetz:
fix #7181

close #7182

PR #7223, opened 2023-05-24
Use try_emplace() from C++17 instead of emplace()

oleg-derevenetz:
Fix minor code smells recently introduced by #7183.


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks for keeping the code the highest quality!

PR #7233, opened 2023-05-26
Reduce memory usage: do not allocate memory for the transform layer for single layer images

Districh-ru:
This PR disables the `transform` layer for images that are set as `singleLayer` and reworks the image processing functions not to work with this layer for such images.

For 1366x768 resolution with opened View World for XL map the engine takes 72 MB, in master build - 82 MB.


src/engine/h2d_file.cpp
```diff
@@ -191,7 +191,9 @@ namespace fheroes2
         const size_t size = static_cast<size_t>( width * height );
         image.resize( width, height );
         memcpy( image.image(), data.data() + 4 + 4 + 4 + 4, size );
-        memcpy( image.transform(), data.data() + 4 + 4 + 4 + 4 + size, size );
+        if ( !image.singleLayer() ) {
+            memcpy( image.transform(), data.data() + 4 + 4 + 4 + 4 + size, size );
+        }
```
ihhub, line 196: As of now all images being stored in h2d file are dual-layer images. I suggest to revert changes as of now.

src/engine/image.h
```diff
@@ -57,16 +58,29 @@ namespace fheroes2
...
-        virtual const uint8_t * image() const;
+        virtual uint8_t * image()
+        {
+            return _data.get();
+        }
+
+        virtual const uint8_t * image() const
+        {
+            return _data.get();
+        }
```
ihhub, line 69: Please revert these changes. Virtual functions cannot be inlined.

src/engine/image.h
```diff
@@ -34,8 +35,8 @@ namespace fheroes2
     class Image
     {
     public:
-        Image();
-        Image( int32_t width_, int32_t height_ );
+        Image() = default;
+        Image( const int32_t width_, const int32_t height_, const bool singleLayer_ = false );
```
ihhub, line 39: We must avoid using default parameters for arguments, especially for widely used classes and structures. I recommend to keep the same number of argument for the constructor and for player where we need to create a single-layer image just call the required method first and only then resize.

src/engine/image.h
```diff
@@ -89,28 +107,28 @@ namespace fheroes2
...
 
     class Sprite : public Image
     {
     public:
-        Sprite();
-        Sprite( int32_t width_, int32_t height_, int32_t x_ = 0, int32_t y_ = 0 );
-        Sprite( const Image & image, int32_t x_ = 0, int32_t y_ = 0 );
-        Sprite( const Sprite & sprite );
+        Sprite() = default;
+        Sprite( const int32_t width_, const int32_t height_, const int32_t x_ = 0, const int32_t y_ = 0, const bool singleLayer_ = false );
```
ihhub, line 129: The same as for Image: please do not use default parameters for an argument.

src/engine/image.h
```diff
@@ -173,20 +191,23 @@ namespace fheroes2
...
     private:
         Image & _image;
         Image _copy;
 
-        int32_t _x;
-        int32_t _y;
+        int32_t _x{ 0 };
+        int32_t _y{ 0 };
         int32_t _width;
         int32_t _height;
```
ihhub, line 206: We should provide default values for everything or for nothing to be consistent with the code.

src/engine/image.cpp
```diff
@@ -464,27 +468,18 @@ namespace fheroes2
     Image & Image::operator=( Image && image_ ) noexcept
     {
         if ( this != &image_ ) {
-            // We shouldn't copy or move different types of images.
-            assert( _singleLayer == image_._singleLayer );
+            // We shouldn't move different types of images.
+            // assert( _singleLayer == image_._singleLayer );
```
ihhub, line 472: Please either remove the code and uncomment it.

src/engine/image.cpp
```diff
@@ -514,9 +513,10 @@ namespace fheroes2
             return;
         }
 
-        const size_t size = static_cast<size_t>( width_ * height_ );
+        // For single layer images do not allocate memory for the second (transform) layer.
+        const size_t size = _singleLayer ? static_cast<size_t>( width_ * height_ ) : static_cast<size_t>( width_ * height_ ) * 2;
```
ihhub, line 517: Interesting that here we put width and height under a single static cast while for other places we put only width.

src/engine/image.cpp
```diff
@@ -1292,8 +1321,14 @@ namespace fheroes2
         memcpy( out.transform(), in.transform(), in.width() * in.height() );
     }
 
-    Sprite CreateContour( const Image & image, uint8_t value )
+    Sprite CreateContour( const Image & image, const uint8_t value )
     {
+        if ( image.empty() || image.singleLayer() ) {
+            // Contour can be created only for non zero images with transform layer.
```
ihhub, line 1327: How about `A contour can be created only for non-empty images with the transform layer.` ?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -465,3482 +467,3649 @@ namespace
     }
 }
 
-namespace fheroes2
+namespace fheroes2::AGG
 {
-    namespace AGG
```
ihhub, line 470: Can we please revert the changes related to namespace text offset as almost the whole file was modified for this?

src/engine/image_tool.cpp
```diff
@@ -291,90 +293,142 @@ namespace fheroes2
         return true;
     }
 
-    Sprite decodeICNSprite( const uint8_t * data, const uint32_t sizeData, const int32_t width, const int32_t height, const int16_t offsetX, const int16_t offsetY )
+    Sprite decodeICNSprite( const uint8_t * data, const uint32_t sizeData, const int32_t width, const int32_t height, const int16_t offsetX, const int16_t offsetY,
```
ihhub, line 296: The information related to the number of layers should come from ICN itself, not the opposite way around. Once we read an image we will know whether there are any transform pixels present. If they are not then we can convert the image into a single-layer. Yes, it is not the best approach in terms of performance but it is much more flexible. What do you think?

ihhub (changes requested):
Hi @Districh-ru , I left several comments in this pull request. Would you mind to take a look at them?

src/engine/h2d_file.cpp
Districh-ru: Hi @ihhub , yes, in h2d all images are double-layer, but if we want to get the single-layer image not to waste memory for storing transform layer with zeros (e.g. for "hotkeys_icon.image" and "graphics_icon.image") we can make `image` single-layer before calling `h2d::readImage()` and so the loader will not waste resources for reading the transform layer.
We can revert these changes and convert the image to single-layer after calling `h2d::readImage()` but we will waste resources for reading transform layer, and then for converting image data from double-layer to single-layer.

Can we change the h2d images format to store the `singleLayer` bit?

src/engine/image.h
Districh-ru: Sure!

src/engine/image.h
Districh-ru: Or can we revert changes in this constructor and make a new constructor `Image( const int32_t width_, const int32_t height_, const bool singleLayer_ );` ?

src/engine/image.h
Districh-ru: Sure!)

src/engine/image.cpp
Districh-ru: I forgot to remove the comments after testing..

src/engine/image.cpp
Districh-ru: I'll fix this line.

src/engine/image.cpp
Districh-ru: Yes, it sounds better!

src/fheroes2/agg/agg_image.cpp
Districh-ru: Sure)

src/engine/h2d_file.cpp
ihhub: Yes, we can change it as well as we should modify some existing images stored there as well. Therefore, I am proposing not to touch this part and address this in a separate pull request.

src/engine/image.h
ihhub: I think it is not intuitive for the developers to have such a constructor as it rises questions:
- so what type of image a normal constructor creates?
- why we can't immediately create a single-layer image without resizing?

One way is to add to every constructor such an option or just have a single method to do this.

src/engine/image_tool.cpp
Districh-ru: Sure, I'll try this approach a little later.

src/engine/h2d_file.cpp
Districh-ru: OK, lets leave it to a separate PR, I'll revert the changes.

src/engine/image.h
Districh-ru: OK, lets revert the changes for now and if ti will be needed in future we can make a method to create and resize the image or add the `const bool singleLayer_` to every constructor.

ihhub:
Hi @Districh-ru , something is wrong with `src/fheroes2/battle/battle_interface.cpp` file as it has 11k changed lines.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -513,6 +534,11 @@ namespace fheroes2
                 const uint8_t * data = body.data() + headerSize + header1.offsetData;
 
                 _icnVsSprite[id][i] = decodeICNSprite( data, sizeData, header1.width, header1.height, header1.offsetX, header1.offsetY );
+
+                // Disable transform layer if it is not needed (contains only zeros).
+                if ( isTransformLayerNotUsed( _icnVsSprite[id][i] ) ) {
```
ihhub, line 539: Here we are running through all data twice. We should do such a check within `decodeICNSprite()` function. Otherwise, this code will slow down resource loading in twice.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2681,10 +2737,23 @@ namespace fheroes2
...
+                        original.image()[2314] = 70;
+                        original.image()[5160] = 71;
+                        original.image()[5827] = 18;
+                        original.image()[7474] = 167;
+                    }
+                }
+                return true;
+            case ICN::PORT0092:
+            case ICN::PORT0095:
+                // Sorceress and Necromancer captains have incorrect transparent pixels.
```
ihhub, line 2750: What values do these pixels have?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3615,11 +3687,43 @@ namespace fheroes2
 
                 break;
             }
+            case ICN::BRCREST: {
+                LoadOriginalICN( id );
+                if ( _icnVsSprite[id].size() == 7 ) {
+                    // All sprites except the last in this ICN are non-transparent.
+                    for ( int32_t i = 0; i < 6; ++i ) {
+                        _icnVsSprite[id][i]._disableTransformLayer();
```
ihhub, line 3695: Aren't we doing the same in `LoadOriginalICN()` function?

ihhub (changes requested):
Hi @Districh-ru , I left several more comments here. Could you please check them?

src/engine/h2d_file.cpp
```diff
@@ -173,6 +173,7 @@ namespace fheroes2
 
     bool readImageFromH2D( H2DReader & reader, const std::string & name, Sprite & image )
     {
+        assert( !image.singleLayer() );
```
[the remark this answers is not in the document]
Districh-ru: Hi, @ihhub, sure. I added a comment and a TODO to implement the single-layer images in H2D format.

src/engine/image.h
```diff
@@ -75,42 +83,45 @@ namespace fheroes2
...
-        void reset(); // makes image fully transparent (transform layer is set to 1)
-        void clear(); // makes the image empty
+        // Set all data int the image layer to 0 and make non single layer images fully transparent (transform layer is set to 1).
+        void reset();
+
+        // Make the image empty.
+        void clear();
 
-        void fill( uint8_t value ); // fill 'image' layer with given value, setting 'transform' layer set to 0
+        // Fill 'image' layer with given value, setting the non single layer images 'transform' layer to 0.
```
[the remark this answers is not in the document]
Districh-ru: Sure)

src/engine/image.cpp
```diff
@@ -514,82 +509,95 @@ namespace fheroes2
             return;
         }
 
-        const size_t size = static_cast<size_t>( width_ * height_ );
+        // For single layer images do not allocate memory for the second (transform) layer.
+        const size_t size = _singleLayer ? static_cast<size_t>( width_ ) * height_ : static_cast<size_t>( width_ ) * height_ * 2;
```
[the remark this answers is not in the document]
Districh-ru: Thanks for the suggestion! It looks better and removes the duplication of `width_ * height_` in the code.

src/engine/image.cpp
```diff
@@ -619,42 +627,36 @@ namespace fheroes2
...
     ImageRestorer::ImageRestorer( Image & image )
         : _image( image )
-        , _x( 0 )
-        , _y( 0 )
         , _width( image.width() )
         , _height( image.height() )
-        , _isRestored( false )
     {
         _updateRoi();
+        // Since the restorer is used to restore the screen part, the transform layer is not needed.
```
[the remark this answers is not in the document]
Districh-ru: Currently this class is used to restore the `fheroes2::Display::instance()` which is single-layer. But maybe in future we will want to restore the double-layer images. So I left the class name unchanged and changed the logic to:
``` cpp
        if ( image.singleLayer() ) {
            _copy._disableTransformLayer();
        }
```

src/fheroes2/agg/agg_image.cpp
Districh-ru: You are right. I made a check for the need of a transform layer while decoding ICN.

src/fheroes2/agg/agg_image.cpp
Districh-ru: I checked the pixel numbers that needs a fix and picked up their color to fix these and some other images,

src/fheroes2/agg/agg_image.cpp
Districh-ru: Yes, this is a leftover after the previous code implementation.
I removed this loop.
 :)

Districh-ru:
This PR became very big and is now marked as draft until #7316 (which contains some fixes made in this PR) is merged.
Also tomorrow I'll try to open a PR with code clean-up made here in `image.cpp` to make this PR focus only on making single-layer images not to use memory for the transform layer.

Districh-ru:
Closing this PR in favor to #8241

PR #7238, opened 2023-05-27
Improve AI retreat logic during battles

ihhub:
AI should sometimes should take into an account the future.

The previous conditions work only for the current AI turn but not for the next one which is not always the case.

### DIFF
--- a/src/fheroes2/ai/ai_common.cpp
+++ b/src/fheroes2/ai/ai_common.cpp
@@ -32,6 +32,7 @@
 #include "payment.h"
 #include "rand.h"
 #include "resource.h"
+#include "settings.h"
 
 namespace AI
 {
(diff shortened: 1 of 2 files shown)


src/fheroes2/ai/ai_common.cpp
```diff
@@ -139,6 +140,21 @@ namespace AI
...
     bool CanPurchaseHero( const Kingdom & kingdom )
     {
-        return kingdom.GetCountCastle() > 0 && kingdom.AllowRecruitHero( true );
+        if ( kingdom.GetCountCastle() == 0 ) {
+            return false;
+        }
+
+        if ( kingdom.GetColor() == Settings::Get().CurrentColor() ) {
+            // This is the AI's current turn.
+            return kingdom.AllowRecruitHero( true );
```
oleg-derevenetz, line 149: Hi @ihhub I don't really think that `Kingdom::AllowRecruitHero()` is applicable here at all, because it checks the number of already hired heroes in a kingdom, and if there is already a full house, it will always return `false`, despite the fact that in this particular case it is necessary to substract the fighting hero from the number of already hired heroes, because he will be automatically "dismissed" in case of retreat, so there will always be an opportunity to redeem him, if only there is enough money. What do you think?

ihhub: This is a good point! I will expand the logic.

ihhub: Fixed.

oleg-derevenetz (approved):
LGTM. I suppose that currently AI never retreats if it has a full house of heroes, and this PR should fix this.

PR #7251, opened 2023-05-29
Make the text input cursor blink and add the ability to set the cursor position anywhere in the edited string

Districh-ru:
This PR:
- adds the text input cursor to the Virtual Keyboard;
- makes the text input cursor blink not to confuse the cursor with the underscore;
- adds ability to set cursor position in Virtual Keyboard and in the name input dialog.

https://github.com/ihhub/fheroes2/assets/113276641/33786f54-0cbc-4c29-8c20-8454f788b783


src/fheroes2/gui/ui_tool.cpp
```diff
@@ -641,6 +640,32 @@ namespace fheroes2
...
+    {
+        if ( text.empty() || pointerCursorXPosition <= textStartXPosition ) {
+            // The text is empty or mouse cursor position is to the left of input field.
+            return 0;
+        }
+
+        const int32_t maxOffset = pointerCursorXPosition - textStartXPosition;
+        int32_t positionOffset = 0;
+        for ( size_t i = 0; i < text.size(); ++i ) {
+            positionOffset += Text::getCharacterWidth( static_cast<uint8_t>( text[i] ), Font::BIG );
```
ihhub, line 654: Let's avoid using legacy code. We can use `fheroes2::AGG::getChar().width()` call here.

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -361,7 +377,7 @@ bool Dialog::InputString( const std::string & header, std::string & res, const s
                 buttonOk.draw();
             }
 
-            text.set( InsertString( res, charInsertPos, "_" ), fheroes2::FontType::normalWhite() );
+            text.set( InsertString( res, charInsertPos, isCursorVisible ? "_" : "\x7F" ), fheroes2::FontType::normalWhite() );
```
ihhub, line 380: Is there a reason we need to use `\x7F` character? Can we use a space character instead?

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -391,14 +374,15 @@ std::string SelectFileListSimple( const std::string & header, const std::string
...
-                charInsertPos = GetInsertPosition( filename, le.GetMouseCursor().x, enter_field.x );
+                charInsertPos = fheroes2::getTextInputCursorPosition( filename, charInsertPos, le.GetMouseCursor().x, enter_field.x );
                 if ( filename.empty() ) {
                     buttonOk.disable();
                 }
 
                 needRedraw = true;
             }
-            else if ( le.KeyPress() && ( !is_limit || fheroes2::Key::KEY_BACKSPACE == le.KeyValue() || fheroes2::Key::KEY_DELETE == le.KeyValue() ) ) {
+            else if ( le.KeyPress() && ( fheroes2::Key::KEY_UP != le.KeyValue() ) && ( fheroes2::Key::KEY_DOWN != le.KeyValue() )
```
ihhub, line 384: Is there any specific logic why we ignore Key UP and Key DOWN events? What's about Page UP and Page DOWN?

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -150,34 +155,60 @@ namespace
...
 
-            _info += character;
+            _info.insert( _cursorPosition, 1, character );
+
+            ++_cursorPosition;
 
             _output.render( renderInputArea() );
         }
 
         void removeLastCharacter()
```
ihhub, line 165: In this case we should rename this method as it doesn't remove the last character anymore.

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -150,34 +155,60 @@ namespace
                 return;
             }
 
-            _info += character;
```
ihhub, line 153: I suggest to rename the method into **insertCharacter** as we can insert it in the middle of a string.

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -150,34 +155,60 @@ namespace
...
+        void changeCursorState()
+        {
+            _isCursorVisible = !_isCursorVisible;
+
+            _cursor = _isCursorVisible ? '_' : '\x7F';
+
+            renderInputArea();
+        }
+
+        void setCursortPosition( const int32_t clickXPosition, const int32_t startXPosition )
```
ihhub, line 192: Misspelled Cursor**t**.

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -579,9 +610,12 @@ namespace
 
         const fheroes2::Rect windowRoi{ renderer.getWindowRoi() };
         const fheroes2::Rect buttonsRoi = getButtonsRoi( buttons, windowRoi.getPosition() + offsetFromWindowBorders );
+        const fheroes2::Rect textRoi{ windowRoi.x + ( windowRoi.width - inputAreaSize.width ) / 2 + inputAreaBorders,
```
ihhub, line 613: I suggest to create a method in **KeyboardRenderer** class to return this Rect to keep the calculations in one place.

ihhub (changes requested):
Hi @Districh-ru , I left several comments and questions. Would you mind to take a look at them?

src/fheroes2/gui/ui_tool.cpp
Districh-ru: Hi @ihhub. Thanks for the suggestion!
Sure, lets use `getChar()` function.
But after making changes here I had to make some changes in `skill_bar.cpp`.

src/fheroes2/dialog/dialog_selectcount.cpp
Districh-ru: The space character is 6 pixels wide, and '_' is 9. So we need to have a transparent character with the width of '_'.
Using Space we cannot avoid text shift:
fheroes2 2023-06-03 13-34-20-587

src/fheroes2/dialog/dialog_selectfile.cpp
Districh-ru: We ignore Key UP/DOWN to allow user to switch between the existing saves using UP/DOWN keys.
And you are right, we need to check for all listbox changes not to overwrite its results.
I changed logic a bit, now if `const bool listboxEvent = listbox.QueueEventProcessing()` is true this `if` block is skipped and user can properly use UP/DOWN/Page_UP/Page_Down keys in save dialog.

src/fheroes2/gui/ui_keyboard.cpp
Districh-ru: OK)

src/fheroes2/gui/ui_keyboard.cpp
Districh-ru: Sure)

src/fheroes2/gui/ui_keyboard.cpp
Districh-ru: Thanks.

src/fheroes2/gui/ui_keyboard.cpp
Districh-ru: OK)

src/fheroes2/dialog/dialog_selectcount.cpp
ihhub: Got it. Thanks for the explanation!

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -289,25 +290,29 @@ namespace fheroes2
 {
     void RedrawPrimarySkillInfo( const fheroes2::Point & pos, PrimarySkillsBar * bar1, PrimarySkillsBar * bar2 )
     {
+        fheroes2::Display & display = fheroes2::Display::instance();
```
ihhub, line 293: **fheroes2::** is not needed since it is within fheroes2 namespace.

src/fheroes2/gui/ui_tool.h
```diff
@@ -133,6 +137,10 @@ namespace fheroes2
 
     void FadeDisplayWithPalette( const Image & top, const Point & pos, const uint8_t paletteId, const int32_t fadeTimeMs, const int32_t frameCount );
 
+    // Returns the character position number in the 'text' string.
+    size_t getTextInputCursorPosition( const std::string & text, const FontType & fontType, const size_t currentTextCursorPosition, const int32_t pointerCursorXPosition,
```
ihhub, line 141: I suggest to name **pointerCursorXPosition** as **pointerCursorXOffset** and **textStartXPosition** as **textStartXOffset** to avoid confusion between symbol position in text and coordinate position.

ihhub (changes requested):
Hi @Districh-ru , I put just 2 small suggestions. Would you mind to check them?

src/fheroes2/gui/skill_bar.cpp
Districh-ru: Oh, yes)

src/fheroes2/gui/ui_tool.h
Districh-ru: Sure)

ihhub (approved):

oleg-derevenetz (approved):

PR #7254, opened 2023-05-31
Fix missing logic for AI heroes to meet each other in the middle of a turn

ihhub:
- we did not update map object cache when a hero moves so other heroes were using the old cached data and failed to find the hero if there is a need to meet each other. Such behavior is really hard to catch so you can use this save to test the master branch and this branch (green AI hero moves to a castle and another hero ignores him in master branch):
[Great_War_II_0018_hero_inte.zip](https://github.com/ihhub/fheroes2/files/11612281/Great_War_II_0018_hero_inte.zip)
- fix Magellan's Maps visit by AI heroes when discovered objects were not being added for exploration
- fix duplication of object information from the fog discovery event when AI uses "View All" spell.
- reduce the amount of objects being processed for a possible movement. Only action objects should be considered as useful for any movement.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1586,6 +1586,9 @@ namespace AI
             break;
         case MP2::OBJ_COAST:
             AIToCoast( hero, dst_index );
+
+            // Coast is not an action object by definition.
+            isAction = false;
             break;
 
         case MP2::OBJ_MONSTER:
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1574,7 +1574,6 @@ namespace AI
 
         const Maps::Tiles & tile = world.GetTiles( dst_index );
         const MP2::MapObjectType objectType = tile.GetObject( dst_index != hero.GetIndex() );
-        bool isAction = true;
 
         const bool isActionObject = MP2::isActionObject( objectType, hero.isShipMaster() );
         if ( isActionObject )
@@ -1585,10 +1584,8 @@ namespace AI
             AIToBoat( hero, dst_index );
             break;
         case MP2::OBJ_COAST:
+            // Coast is not an action object by definition but we need to do hero's animation.
             AIToCoast( hero, dst_index );
-
-            // Coast is not an action object by definition.
-            isAction = false;
             break;
 
         case MP2::OBJ_MONSTER:
@@ -1827,8 +1824,8 @@ namespace AI
             AIToSirens( hero, objectType, dst_index );
             break;
         default:
-            assert( !isActionObject ); // AI should know what to do with this type of action object! Please add logic for it.
-            isAction = false;
+            // AI should know what to do with this type of action object! Please add logic for it.
+            assert( !isActionObject );
             break;
         }
 
@@ -1836,7 +1833,7 @@ namespace AI
             hero.GetPath().Reset();
 
         // ignore empty tiles
-        if ( isAction )
+        if ( isActionObject )
             AI::Get().HeroesActionComplete( hero, dst_index, objectType );
     }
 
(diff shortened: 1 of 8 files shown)


src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1586,6 +1586,9 @@ namespace AI
             break;
         case MP2::OBJ_COAST:
             AIToCoast( hero, dst_index );
+
+            // Coast is not an action object by definition.
+            isAction = false;
```
oleg-derevenetz, line 1591: What is the purpose of this separate `isAction` variable? Why we can't just use `isActionObject` here instead?

src/fheroes2/ai/normal/ai_normal.cpp
```diff
@@ -46,12 +47,20 @@ namespace AI
...
+            // std::lower_bound is used in order to find the correct spot for insertion in a sorted array.
+            auto iter = std::lower_bound( _mapActionObjects.begin(), _mapActionObjects.end(), indexObject,
+                                          []( const IndexObject & left, const IndexObject & right ) { return left.first < right.first; } );
+            if ( iter != _mapActionObjects.end() && iter->first == indexObject.first ) {
+                // The object exist! Most likely because of View All spell.
+                iter->second = indexObject.second;
+            }
+            else {
+                _mapActionObjects.emplace( iter, indexObject );
+            }
```
oleg-derevenetz, line 63: Is it possible to use the `updateMapActionObjectCache()` here somehow instead of duplicating the code?

oleg-derevenetz (commented):

src/fheroes2/ai/normal/ai_normal.cpp
```diff
@@ -46,12 +47,20 @@ namespace AI
     void Normal::revealFog( const Maps::Tiles & tile )
     {
         const MP2::MapObjectType object = tile.GetObject();
-        if ( object != MP2::OBJ_NONE ) {
+        if ( MP2::isActionObject( object ) ) {
```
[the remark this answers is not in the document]
ihhub: Sure.

src/fheroes2/ai/normal/ai_normal.cpp
ihhub: Yup. We can do this.

src/fheroes2/ai/ai_hero_action.cpp
ihhub: It's leftover before we changed pathfinding and removed Coast as an action object for AI. I will change it.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1660,7 +1648,20 @@ namespace AI
...
-        assert( std::is_sorted( _mapObjects.begin(), _mapObjects.end() ) );
+        assert( std::is_sorted( _mapActionObjects.begin(), _mapActionObjects.end() ) );
+
+        std::set<int> objectIndexes;
+
+        for ( const auto & actionObject : _mapActionObjects ) {
+            if ( actionObject.second == MP2::OBJ_HEROES ) {
+                const Maps::Tiles & tile = world.GetTiles( actionObject.first );
+                const Heroes * tileHero = tile.GetHeroes();
+                assert( tileHero != nullptr );
```
oleg-derevenetz, line 1659: If fheroes2 would be built with `WITH_DEBUG` and also with `NDEBUG` (unlikely but nevertheless possible), there will be "unused variable" warnings. To avoid this we can either do this:

```cpp
assert( world.GetTiles( actionObject.first ).GetHeroes() != nullptr );
```

or this:

```cpp
if ( tileHero == nullptr ) {
  assert( 0 );
}
```

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1660,7 +1648,20 @@ namespace AI
...
+            }
+
+            assert( objectIndexes.count( actionObject.first ) == 0 );
+            objectIndexes.emplace( actionObject.first );
```
oleg-derevenetz, line 1663: Here the same thing is done twice. We can do something like this instead:

```cpp
const auto [dummy, inserted] = objectIndexes.emplace( actionObject.first );
if ( !inserted ) {
  assert( 0 );
}
```

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few suggestions, please take a look when you have time :)

oleg-derevenetz (approved):

Districh-ru (approved):

PR #7228, opened 2023-05-25
Always recalculate the hero's path when setting the focus on this hero, if the hero is not moving at the moment

oleg-derevenetz:
fix #7226

Recalculate the hero's path in only one place instead of multiple.

Also I simplified the logic of `MoveHeroFromArrowKeys()`: the old logic was overcomplicated with "special" logic for various object types and even individual objects, none of this is necessary.

PR #7259, opened 2023-06-01
Do not recalculate the hero path on focus change if player is controlled by AI

oleg-derevenetz:
Fix issue observed after #7228 on debug builds when a developer grants AI temporary control over a human-controlled player. Here is a save file:

[AI_Crash.zip](https://github.com/ihhub/fheroes2/files/11625209/AI_Crash.zip)

On this video game crashes on failed assertion in the `AIToTeleports()` because the hero's path is not empty on the teleport tile, which should not be under normal conditions:

https://github.com/ihhub/fheroes2/assets/32623900/a1c23b54-c261-41e3-810b-433945d90ac9

Usually, `SetFocus()` should be called for a human-controlled player only, but in debug builds we have `AIAutoControlMode` and `SetFocus()` may be called for a hero in this mode - for example, it is called on the teleport tile before performing the action, which leads to invalid hero's path due to its recalculation on focus change after #7228 using the wrong pathfinder.

With this PR:

https://github.com/ihhub/fheroes2/assets/32623900/74d57688-2a34-40f8-9d28-a3030a698de1

### DIFF
--- a/src/fheroes2/gui/interface_focus.cpp
+++ b/src/fheroes2/gui/interface_focus.cpp
@@ -22,6 +22,7 @@
  ***************************************************************************/
 
 #include <algorithm>
+#include <cassert>
 #include <vector>
 
 #include "audio.h"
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:

PR #7262, opened 2023-06-01
Add shadows for the Virtual Keyboard buttons

Districh-ru:
This PR adds shadows for the Virtual keyboard buttons

virtKB

### DIFF
--- a/src/fheroes2/gui/ui_keyboard.cpp
+++ b/src/fheroes2/gui/ui_keyboard.cpp
@@ -22,7 +22,6 @@
 
 #include <algorithm>
 #include <cassert>
-#include <cstddef>
 #include <cstdint>
 #include <cstdlib>
 #include <functional>
--- a/src/fheroes2/gui/ui_keyboard.cpp
+++ b/src/fheroes2/gui/ui_keyboard.cpp
@@ -600,14 +600,14 @@ namespace
         const fheroes2::Sprite & okayButtonReleasedImage = fheroes2::AGG::GetICN( buttonIcnId, 0 );
         const fheroes2::Sprite & okayButtonPressedImage = fheroes2::AGG::GetICN( buttonIcnId, 1 );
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -591,7 +602,13 @@ namespace
 
         fheroes2::ButtonSprite okayButton( windowRoi.x + ( windowRoi.width - okayButtonReleasedImage.width() ) / 2, windowRoi.y + windowRoi.height - 35,
```
ihhub, line 603: I suggest to move `windowRoi.x + ( windowRoi.width - okayButtonReleasedImage.width() ) / 2, windowRoi.y + windowRoi.height - 35` to a separate Point variable as we use it in 2 places.

src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -477,8 +479,9 @@ namespace
 
         fheroes2::Rect roi{ offset.x + offsets.front(), offset.y, 1, 1 };
 
-        int32_t yOffset = offset.y;
-        for ( size_t i = 0; i < buttonLayout.size(); ++i ) {
+        const size_t buttonRows = buttonLayout.size();
+
+        for ( size_t i = 0; i < buttonRows; ++i ) {
```
ihhub, line 484: I am not sure that this change will make the code faster as `size()` method is inlined and marked as const expression.

ihhub (changes requested):
Hi @Districh-ru , I left 2 comments in this pull request. Would you mind to check them if you have spare time?

src/fheroes2/gui/ui_keyboard.cpp
Districh-ru: Hi, @ihhub, `buttonRows` is also used 2 times in the 497 line to I cached it as it it used here in the loop and in the line 497.

src/fheroes2/gui/ui_keyboard.cpp
Districh-ru: Sure)

ihhub (approved):

Branikolog (approved):
Well done!

ihhub:

PR #7264, opened 2023-06-02
Russian translation: "Succession Wars" as "Войны Престолонаследия", not "войны за престол"

KOLANICH:
Extracted a part of #7207 into a separate PR.

As discussed in #7207, all first letters in this case are capitalized [according to the rules of Russian language](https://orfogrammka.ru/OGL03/70091346.html#id-3Названияупотреблениепрописныхбуквкавычекслитноедефисноераздельноенаписание-3234Собственныеназваниявойн), because it is [the name of the war](https://mightandmagic.fandom.com/wiki/Second_War_of_Enrothian_Succession).

### DIFF
--- a/files/lang/ru.po
+++ b/files/lang/ru.po
@@ -3209,7 +3209,7 @@ msgid "Map Type:\n"
 msgstr "Тип карты:\n"
 
 msgid "The Succession Wars"
-msgstr "Война за престол"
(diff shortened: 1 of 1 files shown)


ihhub:
I am closing this pull request as it is better to stick to the original translation, even if it's wrong, as many players are used to this translation.

PR #7268, opened 2023-06-03
Adjust battlefield Idle and Flag animation speed, avoid setting 0 animation delay

### LINKED ISSUE #7125: Battle screen, creature idle animation shouldn't be dependent on speed settings
### Preliminary checks

Currently, On highest speeds idle animation looks awful.

https://user-images.githubusercontent.com/55348946/236633307-fa0d6400-e8fd-43cd-b5fe-3f64afcff3b5.mp4

For the slow speed setting it also looks not really good.

I propose to choose optimal speed for idle animations and play it the same for the whole 0-9 range of battle speeds.
The same should be applied for the hero flag.

### Additional info

_No response_

Districh-ru:
Close #7125

This PR is a compromise between current idle animation speed setting implementation and constant idle animation speed.
The player has to visually understand the battle speed so the flag and idle animation IMHO should depend on the battle speed setting but not as much as movement animation.
The battle animation delay coefficient changes in the interval from 1.5 (for speed 1) to 0.05 (for speed 10).
This PR makes Idle/Flag animation delay coefficient to change in the interval from 1.0 (speed 1) to 0.5 (speed 10). With these delays animations does not look so awful.

Also this PR adds a fix to avoid adjusting the battle animation speed to 0 ms which can result in high CPU load. 1 ms delay is set instead.

### DIFF
--- a/src/fheroes2/game/game_delays.cpp
+++ b/src/fheroes2/game/game_delays.cpp
@@ -187,8 +187,8 @@ void Game::UpdateGameSpeed()
     const int32_t battleSpeed = conf.BattleSpeed();
     // For the battle speed = 10 avoid the zero delay and set animation speed to the 1/3 of battleSpeedAdjustment step.
     const double adjustedBattleSpeed = ( battleSpeed < 10 ) ? ( ( 10 - battleSpeed ) * battleSpeedAdjustment ) : ( battleSpeedAdjustment / 3 );
-    // For the battle speed =  5 .. 10 adjust the Idle animation not to have insane high animation speed.
-    const double adjustedIdleAnimationSpeed = ( adjustedBattleSpeed < 1 ) ? ( adjustedBattleSpeed + 1 ) / 2 : adjustedBattleSpeed;
+    // Reduce the Idle animation adjustment interval to (0.5 .. 1).
+    const double adjustedIdleAnimationSpeed = ( 19 - battleSpeed ) / 18.0;
 
     delays[BATTLE_FRAME_DELAY].setDelay( static_cast<uint64_t>( 120 * adjustedBattleSpeed ) );
     delays[BATTLE_MISSILE_DELAY].setDelay( static_cast<uint64_t>( 40 * adjustedBattleSpeed ) );
--- a/src/fheroes2/game/game_delays.cpp
+++ b/src/fheroes2/game/game_delays.cpp
@@ -36,7 +36,7 @@ namespace
 
     static_assert( ( DEFAULT_BATTLE_SPEED >= 0 ) && ( DEFAULT_BATTLE_SPEED < 10 ) );
 
-    const double battleSpeedAdjustment = 1.0 / static_cast<double>( 10 - DEFAULT_BATTLE_SPEED );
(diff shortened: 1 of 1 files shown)


Branikolog:
Hello, @Districh-ru !
I've tested your PR and I really appreciate this changes. Idle animation of Genies, ghosts etc looks so much better!
For me, personally, I would prefer constant idle animation for any speed selected. I think there's no obvious visual correlation between animation speed selected and idle speed. Indeed, if you choose 1 and then choose 10 you would definitely notice the difference in idle animation. But when you run the game for the first time today, for instance, you could hardly distinguish which speed setting is selected based on idle speed now... Previously you could understand that easily, just because 9-10 speed setting looked tremendously. 😅 
I don't know, if such fine adjustment is possible and relevant for this PR but...
Speaking about the idle speed, I personally prefer 1.0 interval (1 speed in current PR) which looks the most pleasant to my eye. All idle speeds above 5 look acceptable, but too fast in my opinion.  So, since the average player uses 5-8 speed while playing I suggest to set 1.0 interval not as 1 speed, but as 5-8 to bring the maximum pleasant experience for the majority of players. Or, lets say, at least 5 speed.
I don't know how intervals are correlated with actual speed, but I can suggest an interval based on my personal feelings like 1.2-0.8 to spread over 1-10 speeds in battle. How do you think?

Districh-ru:
> Hello, @Districh-ru ! I've tested your PR and I really appreciate this changes. Idle animation of Genies, ghosts etc looks so much better! For me, personally, I would prefer constant idle animation for any speed selected. I think there's no obvious visual correlation between animation speed selected and idle speed. Indeed, if you choose 1 and then choose 10 you would definitely notice the difference in idle animation. But when you run the game for the first time today, for instance, you could hardly distinguish which speed setting is selected based on idle speed now... Previously you could understand that easily, just because 9-10 speed setting looked tremendously. 😅 I don't know, if such fine adjustment is possible and relevant for this PR but... Speaking about the idle speed, I personally prefer 1.0 interval (1 speed in current PR) which looks the most pleasant to my eye. All idle speeds above 5 look acceptable, but too fast in my opinion. So, since the average player uses 5-8 speed while playing I suggest to set 1.0 interval not as 1 speed, but as 5-8 to bring the maximum pleasant experience for the majority of players. Or, lets say, at least 5 speed. I don't know how intervals are correlated with actual speed, but I can suggest an interval based on my personal feelings like 1.2-0.8 to spread over 1-10 speeds in battle. How do you think?

Hello, @Branikolog!

1.2-0.8 is OK for me) The difference is barely noticeable, but it is there and it's good.
Also I noticed that heroes idle animation was not changed in this PR (was ultra-fast at 10 speed) and fixed it.

Branikolog:
@Districh-ru 
Looks good! I'll try to test this PR more deeply up tomorrow and review as soon as possible!

Branikolog (approved):

ihhub (approved):

ihhub:

drevoborod:
Tried to understand the difference between old and fixed behaviour by watching Phoenixes animation. It's constant animation but it's not affected by battle speed at all :) Just FYI, it doesn't look like an issue to me.

PR #7277, opened 2023-06-05
Campaign scenario UI: show the currently selected bonus if the campaign scenario is already in progress

### LINKED ISSUE #7274: Info shows wrong choice in campaign
### Preliminary checks

### Platform

Linux

### Describe the bug

Start a campaign, and you'll be presented with 3 choices. Select second or third choice, and start the map. Then click Info button. It will claim that you selected the first choice.

Рабочий стол 1_365

I chose barb here, not necro

### Save file

[Evil_4_0026.savc.gz](https://github.com/ihhub/fheroes2/files/11654879/Evil_4_0026.savc.gz)

### Additional info

_No response_

oleg-derevenetz:
fix #7274

### DIFF
--- a/src/fheroes2/campaign/campaign_savedata.cpp
+++ b/src/fheroes2/campaign/campaign_savedata.cpp
@@ -95,6 +95,7 @@ namespace Campaign
         _currentScenarioInfoId = { -1, -1 };
         _daysPassed = 0;
         _difficulty = CampaignDifficulty::Normal;
+        _currentScenarioBonus = {};
     }
 
     void CampaignSaveData::setCarryOverTroops( const Troops & troops )
--- a/src/fheroes2/campaign/campaign_savedata.cpp
+++ b/src/fheroes2/campaign/campaign_savedata.cpp
@@ -27,6 +27,8 @@
 
 #include "army.h"
 #include "campaign_data.h"
+#include "game_io.h"
+#include "save_format_version.h"
 #include "serialize.h"
 
 namespace Campaign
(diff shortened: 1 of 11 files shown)


ihhub:
Hi @oleg-derevenetz , I think we need to change the way how we store campaign data. Yes, this change works fine but in the future we need to fix the problem with browsing previous completed scenarios and there we will have problem as we don't save any of them. The best approach would be to store bonus ID for each scenario instead of currently selected bonus.

What do you think?

oleg-derevenetz:
@ihhub are we able to browse the previously completed scenarios using the original UI? I thought that there are auto-saves for this at the completion of each scenario. I propose to implement this as soon as we directly approach the implementation of viewing the campaign scenario history, and do not engage in overengineering in advance, because it is not a fact that some kind of implementation "from the head" will then fit.

ihhub:
@oleg-derevenetz , yes, it is possible to browse previously completed scenarios in the original game.

oleg-derevenetz:
@ihhub ah, OK then, I'll try to rework the bonus storage in this PR later.

src/fheroes2/campaign/campaign_savedata.cpp
```diff
@@ -153,14 +160,42 @@ namespace Campaign
...
+        static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_1005_RELEASE, "Remove the logic below." );
+        if ( Game::GetVersionOfCurrentSaveFile() < FORMAT_VERSION_1005_RELEASE ) {
+            // There is always a bonus in all the original HoMM2 campaign missions
+            data._bonusesForFinishedMaps.resize( data._finishedMaps.size(), 0 );
+        }
+        else {
+            msg >> data._bonusesForFinishedMaps;
+
+            // Make sure that the number of elements in the vector of map bonuses matches the number of elements in the vector of finished maps
+            data._bonusesForFinishedMaps.resize( data._finishedMaps.size(), -1 );
```
ihhub, line 195: Could you please explain why we are setting -1 here instead of 0?

ihhub (commented):
Hi @oleg-derevenetz , I left one comment here. Would you mind to take a look?

src/fheroes2/campaign/campaign_savedata.cpp
oleg-derevenetz: Hi @ihhub We can set 0 as well, but:

1. Normally this should be a no-op;
2. If this is not a no-op, then I prefer to emphasize here that there is no information about selected bonus, and value 0 means that the first bonus was selected. Also in future we May have a campaign where there will be no bonuses for certain missions.

ihhub: Good points. Thanks for the explanation!

ihhub:
@oleg-derevenetz , huge thanks for these improvements!

PR #7278, opened 2023-06-06
Improve army distribution in castles for AI players

ihhub:
The current AI behavior does not cover possible scenarios:
- army is not distributed evenly after visiting by a hero. In most cases a hero grabs the most of the army and leaves just one stack of weak monsters. There is no perfect rule how to distribute an army of weak monsters but it is proven that having 5 stacks of monsters is better than 1 stack as there is a higher chance of shooting towers
- monsters were not upgraded if they were initially in a castle (use [AI_does_not_merge_army.zip](https://github.com/ihhub/fheroes2/files/11666182/AI_does_not_merge_army.zip) save, wait for the 7th day when a Yellow hero moves to the castle)
- if monsters are present in the castle no army shuffling was done (use [ai_castle_army_distribution.zip](https://github.com/ihhub/fheroes2/files/11666209/ai_castle_army_distribution.zip) save and just end the turn. Red hero will move to the castle)

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -217,7 +217,15 @@ namespace AI
         Army & heroArmy = hero.GetArmy();
         Army & garrison = castle.GetArmy();
 
+        // Merge all troops in the castle to have the best army.
+        heroArmy.JoinStrongestFromArmy( garrison );
+
+        // Upgrade troops and try to merge them again.
         heroArmy.UpgradeTroops( castle );
+        garrison.UpgradeTroops( castle );
+        heroArmy.JoinStrongestFromArmy( garrison );
+
+        // Recruit more troops and also merge them.
         castle.recruitBestAvailable( budget );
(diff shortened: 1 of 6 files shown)


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -217,7 +217,15 @@ namespace AI
         Army & heroArmy = hero.GetArmy();
         Army & garrison = castle.GetArmy();
 
+        // Merge all troops in the castle to have the best army.
+        heroArmy.JoinStrongestFromArmy( garrison );
+
+        // Upgrade troops and try to merge them again.
         heroArmy.UpgradeTroops( castle );
+        garrison.UpgradeTroops( castle );
+        heroArmy.JoinStrongestFromArmy( garrison );
```
oleg-derevenetz, line 226: Is there a reason for double-merge? Why we can't upgrade everything first and then merge already upgraded troops?

oleg-derevenetz (commented):
Hi @ihhub I just left one question, could you please take a look when possible?

oleg-derevenetz (approved):

PR #7282, opened 2023-06-07
Speedup Homebrew on macOS

oleg-derevenetz:
At the moment, installing Homebrew packages on GitHub macOS VMs is very slow (sometimes this process does not even fit into the 30-minute limit). This PR aims to fix this with the following changes:

* Do not run a separate `brew update` step. `brew install` performs "lightweight" update (using `brew update --auto-update`).
* Do not update outdated dependencies of installed packages (using the `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK` environment variable). This can be a bit dangerous if some installed package expects a newer version of its dependency, but this leads to a good acceleration of the package installation process.


ihhub:

PR #7286, opened 2023-06-09
Fix rating in campaign congratulations text in win video

Districh-ru:
This PR fixes wrong display of Rating in congratulations text shown in WIN video. In 'master' build it was calculated without taking into account of campaign difficulty.


ihhub (approved):

ihhub:

PR #7299, opened 2023-06-12
Explicitly call static function 'GetStrength()' to resolve the ambiguity between 'Army::GetStrength()' and 'Troops::GetStrength()'

### LINKED ISSUE #7298: Assertion while AI trying to split troops.
### Preliminary checks

### Platform

Windows

### Describe the bug

An assertion fails while AI turn:
`Assertion failed: weakestStack != nullptr, file G:\Games\heroes2\git\src\fheroes2\army\army.cpp, line 858`

### Save file

[troopSplitAssertion.zip](https://github.com/ihhub/fheroes2/files/11719315/troopSplitAssertion.zip)

### Additional info

This assertion occurs as for some reason AI hero has no troops in his army...

Districh-ru:
fix #7298 

This PR fixes the army troops strength calculation while AI performing the army split by explicitly calling the static function 'GetStrength()' from the corresponding class.

The incorrect army strength calculation sometimes led to the dismissal of the only unit in the hero's army: [troopSplitAssertion.zip](https://github.com/ihhub/fheroes2/files/11719315/troopSplitAssertion.zip)

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -698,7 +698,10 @@ void Troops::JoinStrongest( Troops & giverArmy, const bool keepAtLeastOneSlotFor
 
     // First check if the weakest troop is actually worth to keep.
     const double weakestStrength = weakest->GetStrength();
-    const double totalArmyStrength = GetStrength();
+    const double totalArmyStrength = Troops::GetStrength();
+
+    assert( totalArmyStrength >= weakestStrength );
+
     // The weakest army should not be more than 5% from the overall army strength.
     const double strengthLimit = totalArmyStrength / 20;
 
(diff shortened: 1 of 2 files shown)


oleg-derevenetz (approved):

src/fheroes2/army/army.cpp
```diff
@@ -698,7 +698,10 @@ void Troops::JoinStrongest( Troops & giverArmy, const bool keepAtLeastOneSlotFor
 
     // First check if the weakest troop is actually worth to keep.
     const double weakestStrength = weakest->GetStrength();
-    const double totalArmyStrength = GetStrength();
+    const double totalArmyStrength = Troops::GetStrength();
+
+    assert( totalArmyStrength >= weakestStrength );
+
     // The weakest army should not be more than 5% from the overall army strength.
     const double strengthLimit = totalArmyStrength / 20;
```
ihhub, line 706: I can't comment below so I do here: at line 734 we have the code:
```
cpp
weakest->SetCount( weakest->GetCount() - acceptableCount );
```
I think there could be the case when **acceptableCount** is equal to **weakest->GetCount()**. This is the case when the count is 1. Could you please modify the code to be:
```cpp
if ( acceptableCount == weakest->GetCount() ) {
    weakest->Reset();
}
else {
    weakest->SetCount( weakest->GetCount() - acceptableCount );
}

ihhub (changes requested):

src/fheroes2/army/army.cpp
oleg-derevenetz: @ihhub in fact, count 0 is fine, it is considered as an empty slot anyway.

ihhub: Then it is fine. We should come up with a better code design as the current one is very bad.

ihhub:

PR #7313, opened 2023-06-14
Rework game credits to keep aspect ratio

### LINKED ISSUE #7303: Make credits non-stretchable
### Preliminary checks

Credits are now stretched according to a chosen resolution and it doesn't look good at all for widescreen resolutions:
image

Render the same background as we do for the Main Menu where we might probably add some action items in the future.

### Additional info

_No response_

Districh-ru:
Close #7303

This PR keeps the aspect ratio if game credits and removes redundant resize on every render loop and render of an image when it is in NO_ACTION state (showing page for some time without any effects).

creditsPage

This PR also reworks the code to reduce code repeating and not to use legacy 'Text' class.

### DIFF
--- a/src/fheroes2/game/game_credits.cpp
+++ b/src/fheroes2/game/game_credits.cpp
@@ -32,6 +32,7 @@
 #include "audio_manager.h"
 #include "cursor.h"
 #include "game_delays.h"
+#include "game_mainmenu_ui.h"
 #include "icn.h"
 #include "image.h"
 #include "localevent.h"
@@ -39,7 +40,6 @@
 #include "pal.h"
 #include "screen.h"
 #include "settings.h"
-#include "text.h"
 #include "tools.h"
 #include "translations.h"
 #include "ui_text.h"
@@ -49,8 +49,9 @@ namespace
 {
     void transformToBlack( fheroes2::Image & out )
     {
-        if ( out.empty() )
+        if ( out.empty() ) {
             return;
+        }
 
         assert( !out.singleLayer() );
 
@@ -100,7 +101,7 @@ namespace
             switch ( _animationState ) {
             case AnimationState::FADING_IN:
                 _alphaValue += ALPHA_VALUE_STEP;
-                if ( _alphaValue >= 255 ) {
+                if ( _alphaValue > 255 ) {
                     _alphaValue = 255;
                     _animationState = AnimationState::NO_ACTION;
                     _noActionCounter = 0;
@@ -119,8 +120,9 @@ namespace
                     _animationState = AnimationState::FADING_IN;
 
                     ++_pageId;
-                    if ( _pageId >= _imageCount )
+                    if ( _pageId >= _imageCount ) {
                         _pageId = 0;
+                    }
                 }
 
                 break;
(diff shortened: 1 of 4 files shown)


src/fheroes2/game/game_credits.cpp
```diff
@@ -719,50 +708,83 @@ void Game::ShowCredits()
         pages.emplace_back( generatePriceOfLoyaltyCreditsSecondPage() );
         pages.emplace_back( generatePriceOfLoyaltyCreditsThirdPage() );
     }
+
     pages.emplace_back( generateSuccessionWarsCreditsFirstPage() );
     pages.emplace_back( generateSuccessionWarsCreditsSecondPage() );
 
-    const fheroes2::Sprite header = generateHeader();
+    // Resize the credits pages.
+    const int32_t resizedPageHeight = creditsRoi.width * pages.front().height() / pages.front().width();
```
ihhub, line 716: This is not correct for devices with vertical orientation. Please take a look at `GetScaledICN()` function to handle this case as well.

src/fheroes2/game/game_credits.cpp
```diff
@@ -345,61 +340,49 @@ namespace
...
-        const int32_t miniMonsterXOffset = textInitialOffsetX + ( output.width() - textInitialOffsetX ) / 2;
-
-        title.Set( _( "Original project before 0.7" ), Font::YELLOW_SMALL, textWidth );
-        title.Blit( textInitialOffsetX + ( output.width() - textInitialOffsetX - title.w() ) / 2, offsetY, output );
-        offsetY += title.h();
+        const int32_t miniMonsterXOffset = textInitialOffsetX * 3 / 2;
 
-        name.Set( "Andrey Afletdinov\nhttps://sourceforge.net/\nprojects/fheroes2/", Font::SMALL, textWidth );
-        name.Blit( textInitialOffsetX + ( output.width() - textInitialOffsetX - name.w() ) / 2, offsetY, output );
+        renderText( output, offsetX, offsetY, textWidth - 10, _( "Original project before 0.7" ), "Andrey Afletdinov\nhttps://sourceforge.net/\nprojects/fheroes2/" );
```
ihhub, line 377: The font for this particular text is different from others.

ihhub (changes requested):
Hi @Districh-ru , I left 2 comments in this pull request. Would you mind to take a look at them?

src/fheroes2/game/game_credits.cpp
Districh-ru: Hi, @ihhub.
It would not be correct if here we were calculating the resizedwidth and height by display size.
As `creditsRoi` is set by using properly scaled Main Menu background image parameters (x, y, width, height) we can use width here to calculate the height if resized credits page.
All is correct for high height resolutions:
изображение

src/fheroes2/game/game_credits.cpp
Districh-ru: Thanks for noticing this.
Fixed)

ihhub (approved):

src/fheroes2/game/game_credits.cpp
```diff
@@ -719,50 +713,83 @@ void Game::ShowCredits()
         pages.emplace_back( generatePriceOfLoyaltyCreditsSecondPage() );
         pages.emplace_back( generatePriceOfLoyaltyCreditsThirdPage() );
     }
+
     pages.emplace_back( generateSuccessionWarsCreditsFirstPage() );
     pages.emplace_back( generateSuccessionWarsCreditsSecondPage() );
 
-    const fheroes2::Sprite header = generateHeader();
+    // Resize the credits pages. 'creditsRoi' is made using MAin Menu background parameters that were already properly calculated for the current resolution.
```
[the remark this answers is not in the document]
Districh-ru: Hi @zenseii. Fixed, thanks!

ihhub:

PR #7317, opened 2023-06-15
Improve image processing logic for single-layer images

Districh-ru:
This PR contains changes made in https://github.com/ihhub/fheroes2/pull/7233
While working on https://github.com/ihhub/fheroes2/pull/7233 there were made some necessary for that PR changes in the code, but PR became very big. To reduce the size of that PR and to properly test and merge some changes this PR is made. It contains:
- improved image processing logic for single-layer images: currently (in master build) some functions analyze and change the 'transform' layer of single-layer images which is not correct (wastes resources and in some places may give unpredictable result);
- clean up `image.h` and `image.cpp` code to make Clang-Tidy a little happier.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -198,29 +198,35 @@ namespace
         69,  69,  69,  69,  69,  69,  69,  69,  69,  69,  242, 243, 244, 245, 246, 247, 248, 249, 250, 251, 252, 253, 254, 255 // No cycle
     };
 
-    bool Validate( const fheroes2::Image & image, int32_t x, int32_t y, int32_t width, int32_t height )
+    bool Validate( const fheroes2::Image & image, const int32_t x, const int32_t y, const int32_t width, const int32_t height )
     {
-        if ( image.empty() || width <= 0 || height <= 0 ) // what's the reason to work with empty images?
+        if ( image.empty() || width <= 0 || height <= 0 ) {
+            // What's the reason to work with empty images?
             return false;
+        }
 
-        if ( x < 0 || y < 0 || x + width > image.width() || y + height > image.height() )
+        if ( x < 0 || y < 0 || x + width > image.width() || y + height > image.height() ) {
             return false;
+        }
 
         return true;
     }
 
     bool Verify( const fheroes2::Image & image, int32_t & x, int32_t & y, int32_t & width, int32_t & height )
     {
-        if ( image.empty() || width <= 0 || height <= 0 ) // what's the reason to work with empty images?
+        if ( image.empty() || width <= 0 || height <= 0 ) {
+            // What's the reason to work with empty images?
             return false;
+        }
 
         const int32_t widthOut = image.width();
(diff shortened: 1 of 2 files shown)


src/engine/image.h
```diff
@@ -97,20 +99,21 @@ namespace fheroes2
...
         void copy( const Image & image );
 
-        int32_t _width;
-        int32_t _height;
+        int32_t _width{ 0 };
+        int32_t _height{ 0 };
         std::unique_ptr<uint8_t[]> _data; // holds 2 image layers
 
-        bool _singleLayer; // only for images which are not used for any other operations except displaying on screen. Non-copyable member.
+        // Jnly for images which are not used for any other operations except displaying on screen.
```
ihhub, line 106: **Jnly** is misspelled :)

src/engine/image.cpp
```diff
@@ -345,10 +365,11 @@ namespace
                 int32_t minDistance = 3 * 255 * 255;
                 uint32_t bestPos = 0;
 
-                const uint8_t * correctorX = transformTable + 256 * 15;
+                // 3840 = 256 * 15. This is the offset of "No cycle" transform table.
+                const uint8_t * correctorX = transformTable + 3840;
```
ihhub, line 369: A compiler will always optimize **256 * 15** into **3840** so it is fine to keep the expression as it is.

src/engine/image.cpp
```diff
@@ -1871,52 +2051,64 @@ namespace fheroes2
             isEqual = true;
 
             for ( size_t i = 1; i < input.size(); ++i ) {
-                if ( *imageIn[0] != *imageIn[i] || *transformIn[0] != *transformIn[i] ) {
+                if ( *imageIn[0] != *imageIn[i] || ( transformIn[0] && transformIn[i] && ( *transformIn[0] != *transformIn[i] ) ) ) {
```
ihhub, line 2054: Right now `transform()` method always returns a valid pointer.

src/engine/image.h
Districh-ru: Hi, @ihhub, thanks)

src/engine/image.cpp
Districh-ru: It is good, I'll revert this change.

src/engine/image.cpp
Districh-ru: A little earlier in this function we make this pointer `nullprt` if image is `singleLayer()`. :)
https://github.com/ihhub/fheroes2/blob/d423af40860866fc95a45ba34fe08edda03a4f85/src/engine/image.cpp#L2035-L2039
So we do this check not to compare the transform layers if one or both images are single-layer.
`_disableTransformLayer()` currently sets `_singleLayer = true;`, some data may be left in transform layer and we should not analyze it if image is marked as single-layer.

ihhub: Thanks for the explanation!

ihhub (approved):

PR #7322, opened 2023-06-16
Do not play more than 2 simultaneous creature walk sounds. Battlefield

Districh-ru:
Currently in master build on high battle speeds when creature is moving the walk sound is played for move to every next cell. It results in many simultaneous move sounds playback.

This PR makes limit to play maximum 2 simultaneous creature walk sounds. The sound in one channel will play fully until it ends. The simultaneous playing sound will be interrupted on every cell with the new walk sound. This is made to hear all creature steps to the next cell.

Also the SDL2 ability to fade out sound is used in this PR to avoid clicks when stopping the sound playback.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -917,6 +917,20 @@ void Mixer::setVolume( const int channelId, const int volumePercentage )
     }
 }
 
+void Mixer::fadeOutChannel( const int channelId, const int timeMs )
+{
+    const std::scoped_lock<std::recursive_mutex> lock( audioMutex );
+
+    if ( isInitialized ) {
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+        Mix_FadeOutChannel( channelId, timeMs );
+#else
+        // SDL1 does not have sound fade functions so we stop the sound in this channel.
+        Mix_HaltChannel( channelId );
+#endif
+    }
+}
+
 void Mixer::Pause( const int channelId /* = -1 */ )
 {
     const std::scoped_lock<std::recursive_mutex> lock( audioMutex );
--- a/src/engine/audio.h
+++ b/src/engine/audio.h
@@ -57,6 +57,9 @@ namespace Mixer
 
     void setVolume( const int channelId, const int volumePercentage );
 
+    // Stop playing sound in a channel after fading it out for a specified time in milliseconds.
+    void fadeOutChannel( const int channelId, const int timeMs );
+
     void Pause( const int channelId = -1 );
     void Resume( const int channelId = -1 );
     void Stop( const int channelId = -1 );
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3954,6 +3954,8 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
 
     // We allow to play maximum 2 simultaneous walk sounds. '-1' means that the channel is not set.
     std::array<int, 2> walkSounds{ -1, -1 };
+    // Set sound fade out time: 325 ms for battle speed 1 - 100 ms for battle speed 10.
+    const int soundFateTimeMs = 1300 / ( Settings::Get().BattleSpeed() + 3 );
 
     while ( dst != pathEnd ) {
         // Check if a wide unit changes its horizontal direction.
@@ -4002,8 +4004,8 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
         if ( ( walkSounds[0] != -1 ) && Mixer ::isPlaying( walkSounds[0] ) ) {
             // The walk sound is playing in the first channel.
             if ( walkSounds[1] != -1 && Mixer ::isPlaying( walkSounds[1] ) ) {
-                // The walk sound is also playing in the second channel. Stop it.
-                Mixer::Stop( walkSounds[1] );
+                // The walk sound is also playing in the second channel. Fade it out and stop.
+                Mixer::fadeOutChannel( walkSounds[1], soundFateTimeMs );
             }
 
             // Start a new walk sound in the second channel.
@@ -4060,8 +4062,8 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
     }
 
     if ( Mixer::isPlaying( walkSounds[0] ) && Mixer::isPlaying( walkSounds[1] ) ) {
-        // Both sound channels are playing walk sound. Stop sound in the second channel.
-        Mixer::Stop( walkSounds[1] );
+        // Both sound channels are playing walk sound. Stop sound in the second channel with fade.
+        Mixer::fadeOutChannel( walkSounds[1], soundFateTimeMs );
     }
 
     // Slowed flying creature has to land.
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3952,6 +3952,9 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
         }
     }
 
+    // We allow to play maximum 2 simultaneous walk sounds. '-1' means that the channel is not set.
+    std::array<int, 2> walkSounds{ -1, -1 };
+
     while ( dst != pathEnd ) {
         // Check if a wide unit changes its horizontal direction.
         if ( isWide && unit.GetTailIndex() == *dst ) {
@@ -3994,7 +3997,30 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
 
         // Render the unit movement with the movement sound.
         // TODO: adjust sounds calls and synchronize them with frames. Take into account that some sounds (like for Cavalry) consists of a sequence of steps.
-        AudioManager::PlaySound( unit.M82Move() );
+
+        // The first audio channel plays sound until it ends, sound in the second channel is interrupted at every step.
+        if ( ( walkSounds[0] != -1 ) && Mixer ::isPlaying( walkSounds[0] ) ) {
+            // The walk sound is playing in the first channel.
+            if ( walkSounds[1] != -1 && Mixer ::isPlaying( walkSounds[1] ) ) {
+                // The walk sound is also playing in the second channel. Stop it.
+                Mixer::Stop( walkSounds[1] );
+            }
+
+            // Start a new walk sound in the second channel.
+            walkSounds[1] = AudioManager::PlaySound( unit.M82Move() );
+        }
+        else {
+            if ( ( walkSounds[1] != -1 ) || Mixer::isPlaying( walkSounds[1] ) ) {
+                // The walk sound playing is playing only in the second channel. Make it the first channel and start walk sound in the second one.
+                walkSounds[0] = walkSounds[1];
+                walkSounds[1] = AudioManager::PlaySound( unit.M82Move() );
+            }
+            else {
+                // There is no walk sound playing in both channels. Start walk sound in the first channel.
+                walkSounds[0] = AudioManager::PlaySound( unit.M82Move() );
+            }
+        }
+
         AnimateUnitWithDelay( unit );
         unit.SetPosition( *dst );
 
@@ -4031,6 +4057,11 @@ void Battle::Interface::RedrawActionMove( Unit & unit, const Indexes & path )
                 unit.SwitchAnimation( Monster_Info::MOVING );
             }
         }
+    }
+
+    if ( Mixer::isPlaying( walkSounds[0] ) && Mixer::isPlaying( walkSounds[1] ) ) {
+        // Both sound channels are playing walk sound. Stop sound in the second channel.
+        Mixer::Stop( walkSounds[1] );
     }
 
     // Slowed flying creature has to land.


Branikolog:
Hi, @Districh-ru !

https://github.com/ihhub/fheroes2/assets/55348946/d40fee1f-4256-411e-86e4-542f96957768

I hope my ears don't lie to me. It seems that the last sound played in a sequence is cut. It is noticeable for walking creatures, especially making 1 single step. Is this behaviour intended? While making two or more steps sounds seem to be played fully (or closer to that). Curious thing is that flying sound of Dragons and Phoenixes is not cut.
Also, I can hear light clicks, while Air Elementals moving. I believe this is step sound being cut in the middle of sequence...

Districh-ru:
> Hi, @Districh-ru !
> I hope my ears don't lie to me. It seems that the last sound played in a sequence is cut. It is noticeable for walking creatures, especially making 1 single step. Is this behaviour intended? While making two or more steps sounds seem to be played fully (or closer to that). Curious thing is that flying sound of Dragons and Phoenixes is not cut. Also, I can hear light clicks, while Air Elementals moving. I believe this is step sound being cut in the middle of sequence...

Hi, @Branikolog!
I fixed the cut of single step move sounds and increased the sound fade time - it should reduce the clicks.
Sometimes at the end of movement there may be a case when the walk sound in the first channel is ending and then the sound in the second channel will be stopped (with fade). This is made not to interrupt the previously started walk sound (which is big enough as it has not still ended). If you think that we always have to play fully the last step sound I can make a change in the logic.

Also I forgot to mention that in this PR only walk sounds were changed (as they have step sound and sometimes a sequence of steps). Now the fly sounds use the same logic.

Could you please check how now the move sounds are played when you have time.

Branikolog:
Hi, @Districh-ru !
For the flying creatures the first sound is played entirely, and all others are cut. I suppose it should work the opposite, right?

https://github.com/ihhub/fheroes2/assets/55348946/18bf1ba0-3afd-47cd-938c-5d5ef9b527b5

Increased fade has fixed clicks for Air elementals. Except of the last sound: I can still hear a click when Elemental stops.

https://github.com/ihhub/fheroes2/assets/55348946/981cafa2-cb3e-46f9-84c7-c5ef947e591a

Districh-ru:
> Hi, @Districh-ru ! For the flying creatures the first sound is played entirely, and all others are cut. I suppose it should work the opposite, right?

It works this way:
- The first sound always plays entirely in the 1st channel.
- If sound in the 1st channel is plying the next sound starts in the 2nd channel halting (with fade) the sound in the 2nd channel (if there is one already playing).
- If sound in the 1st channel is finished then the already playing sound in the 2nd channel is moved to the 1st and is played entirely.

While walking steps have their maximum volume at the beginning of sound (the sound of step), the flying sound has its maximum around the middle of the sound (before flapping the wing must be raised), so for some big birds on very high battle speeds the sound might be cut just before the sound of wing flap.
We can try to adjust it and set the minimum play time, or (I think it'll be better) just revert the changes for the flying creatures as flying sounds do not produce such mess as the walk sound: the fly sound contains only 1 wing flap, while the walk sounds mostly contains two or more steps.

> Increased fade has fixed clicks for Air elementals. Except of the last sound: I can still hear a click when Elemental stops.

And again we are faced with the imperfection of SDL_mixer. :(
I thought that the sound fade-out will be a linear function of time like:
изображение
But in SDL_mixer it has the form of a step function of time:
изображение

And at every step, when the amplitude is significantly changed, there might be clicks in sound.
We need to find a way to make SDL_mixer to use less steps in fade effect or do a fade effect by our engine "on the fly" before performing the creature move, but it will increase the CPU load.

This issue needs some time to find a proper solution, so I'm marking this PR as draft.

oleg-derevenetz:
> But in SDL_mixer it has the form of a step function of time:

I believe, that's because SDL_mixer works this way: when SDL audio driver needs more data, the mixing function is called to produce another audio chunk, and the whole chunk has the same volume. We can try to reduce the `chunkSize` in the `engine/audio.cpp` from 2048 (current value) to 1024 or even 512, and this should help to make the curve "smoother". But this change should be properly tested on all supported platforms (e.g. on Android there were audio distortions when the chunk size was "too big").

Districh-ru:
> I believe, that's because SDL_mixer works this way: when SDL audio driver needs more data, the mixing function is called to produce another audio chunk, and the whole chunk has the same volume (I mean, the specific fading channel has the same volume during the whole chunk). We can try to reduce the `chunkSize` in the `engine/audio.cpp` from 2048 (current value) to 1024 or even 512, and this should help to make the curve "smoother". But this change should be properly tested on all supported platforms (e.g. on Android there were audio distortions when the chunk size was "too big").

@oleg-derevenetz, thanks for the suggestion. Yes, SDL_mixer works this way: for each chunk it adjusts the volume: `Mix_Volume(i, (mix_channel[i].fade_volume * (mix_channel[i].fade_length-ticks)) / mix_channel[i].fade_length);`.

I reduced the `chunkSize` to 1024 and increased the fade time a little and now the clicks are almost not noticeable (The spectrum analyzer shows them, but their level is low to hear). @Branikolog could you please check it when you have time.

Also I reverted the simultaneous sound limit.

I'll also check the changed `chunkSize` on Vita3K emulator.

Branikolog:
Hi, @Districh-ru !
I'm sorry for such a late response.

I've tested the build and found a few issues. Could you check, please, if you can fix it?

I've found that sometimes the last sound is faded too obviously. I've set battle speed as 5 and I can clearly hear fade in some cases.

https://github.com/ihhub/fheroes2/assets/55348946/d13cdde9-a564-4477-89fd-3f5ef4326963

There's definitely a fade after the first crusader moves.
As the second crusader moves I can't hear fade clearly.
I've also found, that sometimes the last sound plays entirely, but sometimes it's cut (or at least it looks like):

https://github.com/ihhub/fheroes2/assets/55348946/af3ed671-5611-4886-9ddc-c740673a82e2

Here the first crusader walking sound looks like been interrupted. The second crusader has walking sound played completely after the troop has stopped.

When I set 10 speed some sounds are actually not really pleasant to ear. Like champion or crusader sounds.

https://github.com/ihhub/fheroes2/assets/55348946/81f8e665-3801-4f18-9fef-8adaf1867453

It sounds predictable according to the proposed logic in this PR, but it simply sounds not pleasant. I'm not sure, if it's possible to do something here... Maybe skip several simultaneous sounds or simply play only a single sound, since creature can cover the whole battlefield almost instantly...

Districh-ru:
> There's definitely a fade after the first crusader moves. As the second crusader moves I can't hear fade clearly. 

Hi @Branikolog!
On the 1st video for the 1st crusader we hear the first sound fully and the second is faded because the first sound has not ended (the sound file has non ended. it is also possible that there was playing very silent sound from the end of the walk sound at that moment) when the crusader came to his destination. When the 2nd crusader walk for 3 steps we hear the first sound fully played, the second step sound is faded when the third step sound is started and this sound is not faded as the first sound has ended, so we can hear it fully.

> Here the first crusader walking sound looks like been interrupted. The second crusader has walking sound played completely after the troop has stopped.

On the 2nd video we hear the fully played first sound. the walk sound for all other steps is interrupted. And the 1st crusader ends his movement when the first sound is near its and and we can clearly hear the fade of the second channel sound. The 2nd crusader has one step less movement and the first sound has more volume at this time than the faded sound in second channel.

Should we fade all sounds when the unit ends his movement (e.g. cavalry moves for 1 hexagon, but move sound is made for many steps, should we stop the sound when the unit stops)?

> When I set 10 speed some sounds are actually not really pleasant to ear. Like champion or crusader sounds.
> It sounds predictable according to the proposed logic in this PR, but it simply sounds not pleasant. I'm not sure, if it's possible to do something here... Maybe skip several simultaneous sounds or simply play only a single sound, since creature can cover the whole battlefield almost instantly...

Yes, in this case we have many-many sounds, but a little less than in master build as all second channel sounds are faded.
We cannot predict the movement time as it depends not only on the set battle speed, but on setting "VSYNC" (if it is off the engine can skip rendering frames if monitor frame rate is less the engine render rate), it also depends on CPU (slow devices compute moves slower and they take more time while sound is played on its default sample rate).
We can try to make playing only one sound for 10th speed, but some units move sounds consists of multiple steps, some of two and some of one. How it will look if creature moves for 6 hexagons and sounds like one step ...
I can not clearly decide how to behave here. And if we make this setting for 10th speed, should we do it fo 9th or 8th? :)

Districh-ru:
Hi, @Branikolog!
I made these changes:
1. The last step walk sound is not faded, the previous already playing sound is faded instead.
2. For high speed moves sound is played only in one channel without any fades. High speed move is a move when creature should do more than 100 animation frames per second (this takes into account the battle speed and creature move speed).
This logic works only for walking (not flying) creatures.

Could you please check these changes when you have time. :)

Branikolog:
Hi, @Districh-ru !
I've started testing the changes. While testing on Crusaders and Champions assertion occurs quite often. Seems like it's random and doesn't depend on distance. It happens on battle speed higher than 3 and below 8.
For 9 and 10 battle speed you've made a single sound playing for walking troops. (Some of creatures, like golems still play multiple sounds though on 9 speed) I think it sounds acceptable. But what about flyers? I find it weird, when walking and flying creatures have different sound logic. I think for flyers we should also play a single flying sound too, to be consistent. 
Also, with haste some creatures on 8 battle speed start playing a single sound while moving, but some not. 

https://github.com/ihhub/fheroes2/assets/55348946/65626f0a-72f3-4158-85ae-906abea228cb

Anyway, this solution is quite different from the previous implementation and I'd like to hear an opinion from other members of our community and devs too.

On 6 and 7 battle speed movement sound plays a little weird at the end of troop path:

https://github.com/ihhub/fheroes2/assets/55348946/d95e5a6d-ef56-409a-95c9-cc39a48d9a95

Sounds like echo. While on 8 speed it feel more pleasant.

Districh-ru:
Hello, @Branikolog,

> I've started testing the changes. While testing on Crusaders and Champions assertion occurs quite often. Seems like it's random and doesn't depend on distance. It happens on battle speed higher than 3 and below 8.

This bug is overcomed: performing sound fade (or possibly very short sounds) sometimes caused to set the "clean-up" (after the playback) of the same channel twice. Now it will not store duplicate channels in the `_channelsToCleanup` set.
@oleg-derevenetz, could you please check if solution for `channelFinished()` is correct when you have time.



> Some of creatures, like golems still play multiple sounds though on 9 speed

Currently I made this rule: If movement animation delay is less than 10 ms we consider this as a High Battle Speed and do not play any simultaneous sound while unit is moving.
It was made as different creatures have cell-to-cell walk time. The slowed creature will need more time to go to the next cell so his "battle speed" is not high. :)
This rule was made for test and we can change it if we want.

> But what about flyers?

From now the same rules are applied to the fliers. The previous solution was bad for fliers so it was removed.

> On 6 and 7 battle speed movement sound plays a little weird at the end of troop path

Now there should be no echo at the end of movement.

@Branikolog, could you please test these changes in youк spare time.

oleg-derevenetz:
Hi @Districh-ru

> This bug is overcomed: performing sound fade (or possibly very short sounds) sometimes caused to set the "clean-up" (after the playback) of the same channel twice. Now it will not store duplicate channels in the _channelsToCleanup set.
@oleg-derevenetz, could you please check if solution for channelFinished() is correct when you have time.

There is nothing wrong to perform the clean-up on the same channel twice. There may be at most two samples in the channel's sample queue, and this is just fine to have two instances of the same channel in the `_channelsToCleanup` in this case. The real problem here may be a race, when `SoundSampleManager::channelFinished()` is called before the `SoundSampleManager::channelStarted()` if the sound is very-very short. Unfortunately, we only get the "real" channel number only after the `Mix_PlayChannel()` is called, and if the sound effect is very short, we may not have time to register the chunk on this channel before it finishes its playback. Ideally, we should start the playback in "paused" state, properly register it and then resume the playback, but I'm not sure yet how to do this with SDL_mixer.

oleg-derevenetz:
Hi @Districh-ru as soon as I reverted your `vector`->`map` commit and made the `Mixer::fadeOutChannel()` no-op (by commenting out the `Mix_FadeOutChannel()` call) assertions are gone (please try it yourself). Apparently there is a race between `Mix_Playing()` check and `Mix_FadeOutChannel()` call: on `Mix_Playing()` check the channel is still playing, but at the moment of calling `Mix_FadeOutChannel()` it is already not playing, hence the additional "not accounted" playback.

Districh-ru:
> Hi @Districh-ru as soon as I reverted your `vector`->`map` commit and made the `Mixer::fadeOutChannel()` no-op (by commenting out the `Mix_FadeOutChannel()` call) assertions are gone (please try it yourself). Apparently there is a race between `Mix_Playing()` check and `Mix_FadeOutChannel()` call: on `Mix_Playing()` check the channel is still playing, but at the moment of calling `Mix_FadeOutChannel()` it is already not playing, hence the additional "not accounted" playback.

@oleg-derevenetz, thanks for the suggestion. I've also reverted to `vector`. Yes, the issue is with the `Mix_FadeOutChannel()`. Without it (commenting it or replacing with `Mix_HaltChannel()`) all works fine.

Sometimes SDL_mixer calls `channelFinished()` twice for the same once started channel:
```
channelFinished(): channel to clean # 1
channelFinished(): channel to clean # 2
channelFinished(): channel to clean # 0
channelFinished(): channel to clean # 3
clearFinishedSamples(): clearing channel # 1
clearFinishedSamples(): clearing channel # 2
clearFinishedSamples(): clearing channel # 0
clearFinishedSamples(): clearing channel # 3
channelStarted(): set first queue channel # 0
channelStarted(): set first queue channel # 1
channelStarted(): set first queue channel # 2
channelStarted(): set first queue channel # 3
channelFinished(): channel to clean # 1
clearFinishedSamples(): clearing channel # 1
channelStarted(): set first queue channel # 1
channelFinished(): channel to clean # 0
channelFinished(): channel to clean # 2
channelFinished(): channel to clean # 0
channelFinished(): channel to clean # 3
channelFinished(): channel to clean # 1
clearFinishedSamples(): clearing channel # 0
clearFinishedSamples(): clearing channel # 2
clearFinishedSamples(): !!! ERROR !!! nullptr in channel # 0
```

oleg-derevenetz:
> Sometimes SDL_mixer calls channelFinished() twice for the same once started channel:

I have an impression that if we call `Mix_FadeOutChannel()` on already stopped channel, it just calls `channelFinished()` handler once again for the same chunk.

oleg-derevenetz:
The "quick fix" may be pausing the channel in the beginning of the `Mixer::fadeOutChannel()` (before the `Mix_Playing()` check) and unpausing it at the end (after the `Mix_FadeOutChannel()` call).

Districh-ru:
> I have an impression that if we call `Mix_FadeOutChannel()` on already stopped channel, it just calls `channelFinished()` handler once again for the same chunk.

Yes, it looks like this is the cause of the problem.

> The "quick fix" may be pausing the channel at the beginning of the `Mixer::fadeOutChannel()` (before the `Mix_Playing()` check) and unpausing it at the end (after the `Mix_FadeOutChannel()` call). Hope that is pauses immediately upon return from `Mix_Pause()` call and there will be no another race... :)

`Mix_Pause( channelId );` did not help :(
<details>
```
Play(): channel # 0
channelStarted(): set first queue channel # 1
Play(): channel # 1
...
clearFinishedSamples(): clearing channel # 1
channelStarted(): set first queue channel # 1
Play(): channel # 1
fadeOutChannel(): channel # 0
channelFinished(): channel to clean # 0
channelFinished(): channel to clean # 2
channelFinished(): channel to clean # 0
```
</details>

And what if we replace `assert( sampleQueue.first != nullptr );` with `break;`?
```cpp
if ( sampleQueue.first == nullptr ) {
    // This channel is already cleaned.
    break;
}

**Comment by oleg-derevenetz** [comment]
> `Mix_Pause( channelId );` did not help :(

Probably because the pausing process is not "instant".

> And what if we replace `assert( sampleQueue.first != nullptr );` with `break;`?

Not `break`, but `continue`, because there may be other channels in the queue. I'd say it would just mask the real problem. We need to think how to properly avoid the race that actually causes this issue.

**Comment by Districh-ru** [comment]
Interesting, but SDL_mixer also calls `Mix_Playing()` inside `Mix_FadeOutChannel()`.
<details><summary>mixer.c</summary>

```cpp
/* Fade out a channel and then stop it automatically */
int Mix_FadeOutChannel(int which, int ms)
{
    int status;

    status = 0;
    if (audio_opened) {
        if (which == -1) {
            int i;

            for (i=0; i<num_channels; ++i) {
                status += Mix_FadeOutChannel(i, ms);
            }
        } else if (which < num_channels) {
            Mix_LockAudio();
            if (Mix_Playing(which) &&
                (mix_channel[which].volume > 0) &&
                (mix_channel[which].fading != MIX_FADING_OUT)) {
                mix_channel[which].fade_volume = mix_channel[which].volume;
                mix_channel[which].fade_length = (Uint32)ms;
                mix_channel[which].ticks_fade = SDL_GetTicks();

                /* only change fade_volume_reset if we're not fading. */
                if (mix_channel[which].fading == MIX_NO_FADING) {
                    mix_channel[which].fade_volume_reset = mix_channel[which].volume;
                }

                mix_channel[which].fading = MIX_FADING_OUT;

                ++status;
            }
            Mix_UnlockAudio();
        }
    }
    return(status);
}
```
<details>

oleg-derevenetz:
> Interesting, but SDL_mixer also calls `Mix_Playing()` inside `Mix_FadeOutChannel()`.

Yeah, it's hard to say how exactly `Mix_FadeOutChannel()` can cause an extra call of the "channel finished" handler, and in which case exactly.

Districh-ru:
> Not `break`, but `continue`, because there may be other channels in the cleanup queue. I'd say it would just mask the real problem.

Sure, `continue`, thanks! 👍 

> We need to think how to properly avoid the race that actually causes this issue.

We can try to calculate the sound duration time and start a time counter when sound is started by `playSound()`. Before calling `Mix_FadeOutChannel()` - check is there enough time left for sound to fade. If not - skip `Mix_FadeOutChannel()` call.

For now I'll make this mask with a TODO comment in this PR to test the battlefields move sounds approach.

oleg-derevenetz:
> We can try to calculate the sound duration time and start a time counter when sound is started by `playSound()`. Before calling `Mix_FadeOutChannel()` - check is there enough time left for sound to fade. If not - skip `Mix_FadeOutChannel()` call.

This approach is vulnerable to the same TOCTOU problem as the `Mix_Playing()` check.

Districh-ru:
> This approach is vulnerable to the same TOCTOU problem as the `Mix_Playing()` check.

You are right. And subtracting a special delay (to compensate the time between check and use) will be a patch and will not work for all (slow or fast) systems. :(
So we need to prepare a sound with fade: calculate the time for 1 cell move, copy a sound for this time into a new sound vector and add to the end the fade. ... But it will take CPU and RAM ... :(

Or change the approach not to use sound fade at all. In example, calculate the `cellSkip = soundDuration / oneCellMoveTime` and skip this number of cells before playing the nest move sound.

oleg-derevenetz:
> Or change the approach not to use sound fade at all.

I vote for not using sound fade at all - at least for now. I don't know yet exactly how it should be used in a safe manner.

Districh-ru:
> I vote for not using sound fade at all - at least for now. I don't know yet how exactly it should be used in a safe manner.

OK, tomorrow I'll implement another approach with `cellSkip`.

oleg-derevenetz:
Hi @Districh-ru, applying the following patch (to the SDL2 git branch of SDL_mixer) fixes this issue:

```patch
diff --git a/src/mixer.c b/src/mixer.c
index 03ea759f..fd2111be 100644
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -403,6 +403,8 @@ mix_channels(void *udata, Uint8 *stream, int len)
 
                     /* rcg06072001 Alert app if channel is done playing. */
                     if (!mix_channel[i].playing && !mix_channel[i].looping) {
+                        mix_channel[i].fading = MIX_NO_FADING;
+                        mix_channel[i].expire = 0;
                         _Mix_channel_done_playing(i);
 
                         /* Update the volume after the application callback */
```

You can try it yourself: checkout & prepare the vcpkg repo, add the patchfile with this contents into the `ports/sdl2-mixer`, add this patchfile to the `PATCHES` section of `portfile.cmake`, build the patched sdl2_mixer libs and replace our pre-built libs with the patched ones.

Districh-ru:
> Hi @Districh-ru, applying the following patch (to the SDL2 git branch of SDL_mixer) fixes this issue:
> You can try it yourself: checkout & prepare the vcpkg repo, add the patchfile with this contents into the `ports/sdl2-mixer`, add this patchfile to the `PATCHES` section of `portfile.cmake`, build the patched sdl2_mixer libs and replace our pre-built libs with the patched ones.

Hi, @oleg-derevenetz, many thanks for help with SDL_mixer!

I made a local brunch to investigate later the cause of `Mix_FadeOutChannel()` later. I'll try your approach in that brunch. And if we good a good result we can start a PR in SDL_mixer repo.

In this PR I've changed the approach and I like the result better than the previous one: there are no stuck record effect and no extra echoes in sound. And we avoid fade bugs, noticed in the beginning of this PR.

src/fheroes2/audio/audio_manager.cpp
```diff
@@ -906,6 +906,14 @@ namespace AudioManager
         g_asyncSoundManager.pushSound( m82, Settings::Get().SoundVolume() );
     }
 
+    uint32_t getSoundDurationMs( const int m82 )
+    {
+        const std::vector<uint8_t> & v = GetWAV( m82 );
```
oleg-derevenetz, line 911: We need to acquire the `AsyncSoundManager`'s resource mutex before this line:

```cpp
std::scoped_lock<std::recursive_mutex> lock( g_asyncSoundManager.resourceMutex() );
```

because `GetWAV()` operates with `wavDataCache` that could be used/changed from a background thread.

Districh-ru: Thanks for the explanation.

Districh-ru:
Hi, @Branikolog, the approach to limit simultaneous walk/fly sounds has changed as we encountered several SLD_mixer framework related issues. So not to use questionable features now we play only "about" one sound at a time. "About" is because on low battle speeds we need to avoid possible sound playback delays in case the sound has ended at the beginning of the next turn and we have to listen silence until the next one.

The algorithm is simple: we now the theoretical time for cell to cell move (from game assets) and we know the move sound duration time. So we can calculate how many cells to wait until the next sound playback.
But on high speeds the real move speed may be less than theoretical due to CPU limit or display v-sync limit. So for this case (if creature movement FPS is more than 50) we start the next sound after a fact that a previous one has ended. A possible pause after the sound ends and the next cell step will be not noticeable as the move speed is very high and I doubt that on these speeds the whole path time will be more than the sound play time. :) But we keep this logic for a case of slow system.

Could you please test this approach when you have time.

Branikolog:
Hi, @Districh-ru 

I've made a couple of tests and here my thoughts.

But firstly I should mention, that application crashes, when I haste some creature and try to move on 10 battle speed.

From the coding point this approach looks nice, but in reality it feels that sound speed doesn't correlate with game speed.
I'm not against playing a single walking sound for faster game speed (10, 9): creatures move so fast, that it's not possible to count the number of hexes it walked, so a single sound fits quite well for the case. Most of creatures stop before a single walking track finishes. Not sure about 8 battle speed with this approach though.
But on some lower speeds it feels like something is wrong and it feels like creature moves faster (because some sounds are played more frequent).

https://github.com/ihhub/fheroes2/assets/55348946/0d301842-92ab-4bd3-9c05-22d88905d80a

6 (and even 7) speed is felt quite OK.
On 5 battle speed walking sounds seem to be more frequent, which makes an impression, that crusader walks faster.
4 speed performs the same sound playback as on 6 speed.
On 3 speed engine seem to play a walking sound for each cell walked. This also makes an impression, that crusader walks faster, than on a previous speed.
The same thing on 2 speed. Also, I've noticed that crusaders felt out of rhythm a bit on this walk. :)

So as a result this implementation sounds really different on different battle speed selected. 
I expect more linear difference in sounds regarding battle speed selected.
Regarding these issues I mostly concerned about the difference in each battle speed: why do they sound so different?
Nevertheless, players not really often change battle speed, being stick to their favourite one, so the difference is not much matter until the next issue: 

https://github.com/ihhub/fheroes2/assets/55348946/be488a78-79ca-4229-baef-33fb8ade2193

Under the haste crusader has less frequent moving sounds (the different approach is used). I believe within a single selected battle speed there should be the similar sound approach: faster troop with haste applied should sound a bit faster (with a little more frequent steps)

Districh-ru:
@Branikolog, thanks for thorough testing!

> But firstly I should mention, that application crashes, when I haste some creature and try to move on 10 battle speed.

On 10th speed with Haste spell the movement time for one cell became 0 and our sound logic tried to divide by 0. Now it is fixed.

> Not sure about 8 battle speed with this approach though.
> But on some lower speeds it feels like something is wrong and it feels like creature moves faster (because some sounds are played more frequent).

Now one sound is played only on 9 and 10 battle speeds.
I slightly reduced the needed delay before the next sound playback so now creatures steps sounds more frequent, but not very frequent as in master build. :)

> Under the haste crusader has less frequent moving sounds (the different approach is used). I believe within a single selected battle speed there should be the similar sound approach: faster troop with haste applied should sound a bit faster (with a little more frequent steps)

Good point!
I made logic based on technical characteristics (display refresh rate, cells to move ...) and forgot that Haste/Slow should sound like Haste and Slow. :))
Now the delay (in cells) before next sound playback is calculated before applying Haste/Slow coefficients. So the creature would souтв faster/slower.

Districh-ru:
> I'm not against playing a single walking sound for faster game speed (10, 9): creatures move so fast, that it's not possible to count the number of hexes it walked, so a single sound fits quite well for the case. Most of creatures stop before a single walking track finishes.

I made a separate PR as probably we have found a solution for this case. While other cases still need some time to find the best solution...

oleg-derevenetz:
@Districh-ru FYI: https://github.com/libsdl-org/SDL_mixer/pull/539

drevoborod:
Issues I noticed:
- For creatures with 4 legs (Cavarly, Wolves, Unicorns) on both low and high speeds the last steps sound plays after a creature has been already stopped.
- On very fast flying creature (Phoenix under Haste spell) on maximum speed sound of the flight is being played after the flight had been finished.

ihhub:
Hi @Districh-ru , could you please explain the status of this pull request and what should be done? Can we implement a part of logic / code which partially fixes the problem but doesn't introduce any issues?

Districh-ru:
Hi, @ihhub,
a part of the logic was merged (https://github.com/ihhub/fheroes2/pull/7471). On the weekend I'll check if we can quickly apply some logic from here.
But probably we should update each sound manually to allow their speed-up - some sounds have 1 step, others two or more; one steps are at the beginning of the sound, others - at the middle.

PR #7342, opened 2023-06-19
Add instructions for installing Gentoo packages

DarthGandalf:


docs/INSTALL.md
```diff
@@ -191,6 +192,43 @@ The recommended option requires the Heroes of Might and Magic II installer file
 For the manual installation you have to copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` from the original game or demo directory to the
 `~/.var/app/io.github.ihhub.Fheroes2/data/fheroes2` directory. The destination folder will be opened when this option is selected.
 
+<a name="gentoo-package"></a>
```
ihhub, line 195: Hi @DarthGandalf , is it correct that installing homm2 package installs fheroes2 package? If yes then this is not legally correct. HoMM2 is a legal trademark owned by Ubisoft, naming a package like this which essentially installs a different engine is not valid. A user must install only fheroes2 package separately.

DarthGandalf: Would something like `fheroes2-data-homm2-gog` be ok?

chewi: I think we're okay here. `games-strategy/homm2-gold-gog` does not install fheroes2, it only installs the data files. It installs them under `/usr/share/fheroes2`, but that is not significant. The package also depends on `games-engines/fheroes2`, but they are sufficiently separated.

ihhub: Let me think a bit about the best solution.

@oleg-derevenetz and @zenseii , if you have some ideas please feel free to express them.

oleg-derevenetz: Hi @ihhub my point of view is as follows: this is the Gentoo's responsibility, not ours. If Gentoo portage committers accepted this scheme, then they don't see any problems with it, and we are not connected with Gentoo Linux distribution in any way and we are not in position to control them. We can only describe the existing installation process on this platform.

chewi: I am the Gentoo Games lead and I suggested it. :blush:

chewi: That's one reason. You could play it with Wine/DOSBox, but we wouldn't package it for that approach.

More importantly, it's where fheroes2 looks for the data in the first place. It's the default value of `FHEROES2_DATA`.

ihhub: Then we are okay.

docs/INSTALL.md
```diff
@@ -191,6 +192,43 @@ The recommended option requires the Heroes of Might and Magic II installer file
...
+```
+
+It will ask you to put the .exe installer of the game and optionally the zip file with FLAC music to your distfiles directory,
+and if it's there, it will unpack everything to the correct places and pull `games-engines/fheroes2` as a dependency.
+
+USE-flag `flac` determines whether the game will use FLAC music or OGG music.
+
+#### Demo version
+
+You can install the demo version for free by adding `games-strategy/homm2-demo HoMM2-Demo` line to your `/etc/portage/package.license`, then using the following command:
```
ihhub, line 215: Is there a reason why we need to type homm2-demo twice?

docs/INSTALL.md
```diff
@@ -191,6 +192,43 @@ The recommended option requires the Heroes of Might and Magic II installer file
...
+<a name="gentoo-package"></a>
+### Gentoo package
+
+If you are using [**Gentoo**](https://gentoo.org), you can install the game using package manager. There are several options:
+
+#### GOG version
+
+If you purchased the game from GOG, add `games-strategy/homm2-gold-gog GOG-EULA` line to your `/etc/portage/package.license`, then use the following command:
+
+```sh
```
ihhub, line 204: Since GoG provides only one type of homm2 version shall we drop **gold** in the name of the command?

ihhub (commented):
Hi @DarthGandalf and @chewi  I left 2 more questions in this pull request. Could you please take a look?

docs/INSTALL.md
DarthGandalf: That's the syntax of the `package.license` file: first word is the package name, including the category (the category is before `/`), then to which license(s) the user agrees for this package, list of licenses is space separated. The text of licenses is available in [portage tree](https://gitweb.gentoo.org/repo/gentoo.git/tree/licenses). I named this file HoMM2-Demo because the license text is quite specific to the demo version of Heroes 2

It could as well be `*/* HoMM2-Demo` if the user agrees with this license for all packages. This would make more sense for the gog package, since the `GOG-EULA` license is actually used by more than 1 package.

And FOSS licenses are accepted by default, which is why there's no need to explicitly accept GPL for games-engines/fheroes2.

docs/INSTALL.md
DarthGandalf: I just tried to be specific, for case if they ever decide to start selling Succession Wars separately. But can drop "gold" from it... @chewi ?

chewi: I'd rather be specific, avoids any confusion.

ihhub (approved):

ihhub:
@DarthGandalf , thank you very much fir these changes!

PR #7347, opened 2023-06-20
Make an AI hero to stop moving when a new action object is discovered

### LINKED ISSUE #7290: Stop a moving AI hero if a new action object has been revealed during movement
### Preliminary checks

At the moment we update object cache while unveiling fog during AI hero movement. However, there is a chance that the newly uncovered object could be much more useful that the current set of available objects to visit.

We need to modify functions to unveil fog to indicate that the moving hero has to stop to re-evaluate the priority upon discovering a new object. This should improve the efficiency of map exploration and also we should be able to update priorities for castles and heroes in case of finding a new enemy hero or castle.

### Additional info

_No response_

ihhub:
close #7290

You can test this save file (see the difference in a Red AI hero):
[AI_fog_discvery.zip](https://github.com/ihhub/fheroes2/files/11800995/AI_fog_discvery.zip)

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal.cpp
+++ b/src/fheroes2/ai/normal/ai_normal.cpp
@@ -24,11 +24,14 @@
 #include <cstdint>
 
 #include "army.h"
+#include "heroes.h"
+#include "kingdom.h"
 #include "maps_tiles.h"
 #include "pairs.h"
 #include "payment.h"
 #include "profit.h"
 #include "rand.h"
+#include "route.h"
 #include "world.h"
 
 namespace AI
--- a/src/fheroes2/ai/normal/ai_normal.cpp
+++ b/src/fheroes2/ai/normal/ai_normal.cpp
@@ -67,6 +67,7 @@ namespace AI
             }
 
             if ( hero->isMoveEnabled() ) {
+                hero->SetMove( false );
                 hero->GetPath().Reset();
                 break;
             }
--- a/src/fheroes2/ai/normal/ai_normal.cpp
+++ b/src/fheroes2/ai/normal/ai_normal.cpp
@@ -44,7 +44,7 @@ namespace AI
         _pathfinder.reset();
     }
 
-    void Normal::revealFog( const Maps::Tiles & tile )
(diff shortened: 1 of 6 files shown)


src/fheroes2/heroes/heroes.cpp
```diff
@@ -1486,8 +1486,6 @@ int Heroes::getNumOfTravelDays( int32_t dstIdx ) const
     const std::list<Route::Step> routePath = world.getPath( *this, dstIdx );
 
     if ( routePath.empty() ) {
-        DEBUG_LOG( DBG_GAME, DBG_TRACE, "unreachable point: " << dstIdx )
```
ihhub, line 1489: This line just pollutes the log while trying to test AI behavior.

src/fheroes2/ai/normal/ai_normal.cpp
```diff
@@ -44,14 +47,30 @@ namespace AI
...
+        const KingdomHeroes & heroes = kingdom.GetHeroes();
+        for ( Heroes * hero : heroes ) {
+            if ( hero == nullptr ) {
+                // How is it even possible?
+                assert( 0 );
+                continue;
+            }
+
+            if ( hero->isMoveEnabled() ) {
+                hero->GetPath().Reset();
```
oleg-derevenetz, line 70: Shouldn't we call `hero->SetMove( false )` here before resetting the hero's path, or it will be handled elsewhere?

oleg-derevenetz (commented):
Hi @ihhub I just left one question here, would you mind to take a look when possible?

src/fheroes2/ai/normal/ai_normal.cpp
ihhub: You're right. Without this line the hero animation "jumps".

oleg-derevenetz (approved):

ihhub:
Hi @drevoborod , could you please help to verify this change?

drevoborod:
> Hi @drevoborod , could you please help to verify this change?

Check completed. The behaviour definitely changed: when the red player discovered new resource in the ocean, it stopped movement and turned to the discovery.

PR #7348, opened 2023-06-20
Reduce the number of calls to get a singleton instance

ihhub:
Just a simple code cleanup to reduce the amount of useless work by CPU.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -163,7 +163,6 @@ public:
         const fheroes2::Sprite & port = Heroes::GetPortrait( index, PORT_SMALL );
         fheroes2::Display & display = fheroes2::Display::instance();
 
-
         if ( !port.empty() ) {
             fheroes2::Blit( port, display, dstx + 5, dsty + 3 );
         }
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -161,13 +161,15 @@ public:
     void RedrawItem( const int & index, int32_t dstx, int32_t dsty, bool current ) override
     {
         const fheroes2::Sprite & port = Heroes::GetPortrait( index, PORT_SMALL );
+        fheroes2::Display & display = fheroes2::Display::instance();
+
 
         if ( !port.empty() ) {
-            fheroes2::Blit( port, fheroes2::Display::instance(), dstx + 5, dsty + 3 );
+            fheroes2::Blit( port, display, dstx + 5, dsty + 3 );
         }
 
         const fheroes2::Text text( Heroes::GetName( index ), current ? fheroes2::FontType::normalYellow() : fheroes2::FontType::normalWhite() );
-        text.draw( dstx + 50, dsty + 5, fheroes2::Display::instance() );
+        text.draw( dstx + 50, dsty + 5, display );
     }
 
     void RedrawBackground( const fheroes2::Point & dst ) override
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -26,7 +26,6 @@
 #include <algorithm>
 #include <cstddef>
 #include <cstdint>
-#include <string>
 #include <vector>
 
 #include "agg_image.h"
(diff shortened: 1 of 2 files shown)


src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -126,16 +126,18 @@ class SelectEnumMonster : public SelectEnum
...
         if ( current ) {
             fheroes2::Text text( mons.GetName(), fheroes2::FontType::normalYellow() );
-            text.draw( dstx + 50, dsty + 10, fheroes2::Display::instance() );
+            text.draw( dstx + 50, dsty + 10, display );
         }
         else {
             fheroes2::Text text( mons.GetName(), fheroes2::FontType() );
-            text.draw( dstx + 50, dsty + 10, fheroes2::Display::instance() );
+            text.draw( dstx + 50, dsty + 10, display );
         }
```
Districh-ru, line 141: We can reduce code duplication a bit here. (And explicitly indicate the `normalWhite()` font type.) What do you think?
```cpp
        const fheroes2::Text text( mons.GetName(), current ? fheroes2::FontType::normalYellow() : fheroes2::FontType::normalWhite() );
        text.draw( dstx + 50, dsty + 10, display );

```

And to finally clean-up this function we can make **const** line 129: `const Monster mons( index );`. :)

Also the same clean-up, including introducing `fheroes2::Display & display = fheroes2::Display::instance();`, can be done to `SelectEnumHeroes::RedrawItem()` (lines 166-181) and `SelectEnumArtifact::RedrawItem()` (lines 197-214).

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -233,16 +235,18 @@ class SelectEnumSpell : public SelectEnum
...
         if ( current ) {
             fheroes2::Text text( spell.GetName(), fheroes2::FontType::normalYellow() );
-            text.draw( dstx + 80, dsty + 10, fheroes2::Display::instance() );
+            text.draw( dstx + 80, dsty + 10, display );
         }
         else {
             fheroes2::Text text( spell.GetName(), fheroes2::FontType() );
-            text.draw( dstx + 80, dsty + 10, fheroes2::Display::instance() );
+            text.draw( dstx + 80, dsty + 10, display );
         }
```
Districh-ru, line 250: The same comment as above.
```cpp
        const fheroes2::Text text( spell.GetName(), current ? fheroes2::FontType::normalYellow() : fheroes2::FontType::normalWhite() );
        text.draw( dstx + 80, dsty + 10, display );
```

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -266,17 +270,19 @@ class SelectEnumSecSkill : public SelectEnum
...
         if ( current ) {
             fheroes2::Text text( str, fheroes2::FontType::normalYellow() );
-            text.draw( dstx + 50, dsty + 10, fheroes2::Display::instance() );
+            text.draw( dstx + 50, dsty + 10, display );
         }
         else {
             fheroes2::Text text( str, fheroes2::FontType() );
-            text.draw( dstx + 50, dsty + 10, fheroes2::Display::instance() );
+            text.draw( dstx + 50, dsty + 10, display );
         }
```
Districh-ru, line 286: The same comment as above.
```cpp
        const fheroes2::Text text( skill.GetName(), current ? fheroes2::FontType::normalYellow() : fheroes2::FontType::normalWhite() );
        text.draw( dstx + 50, dsty + 10, display );
```

Districh-ru (changes requested):
Hi @ihhub. I left a code clean-up suggestion. Could you take a look in your spare time.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -167,17 +163,12 @@ class SelectEnumHeroes : public SelectEnum
     {
         const fheroes2::Sprite & port = Heroes::GetPortrait( index, PORT_SMALL );
 
-        if ( !port.empty() )
+        if ( !port.empty() ) {
             fheroes2::Blit( port, fheroes2::Display::instance(), dstx + 5, dsty + 3 );
```
Districh-ru, line 166: As every hero in non-corrupted original assets has a portrait this condition will be true and `fheroes2::Display::instance()` is called twice: here and in line 171.
So we can reduce the number of these system calls.

Districh-ru (changes requested):
Hi @ihhub. I left one suggestion. Could you take a look when you have time.

src/fheroes2/dialog/dialog_selectitems.cpp
ihhub: Fixed.

Districh-ru (approved):

PR #7351, opened 2023-06-22
Fix Abandoned Mine right sprite update

Districh-ru:
This PR fixes this issue: https://github.com/ihhub/fheroes2/issues/5716#issuecomment-1601480162
It makes engine to update the Abandoned Mine right sprite even if its tile object type is not `MP2::OBJ_NON_ACTION_ABANDONED_MINE`.

Master build:
bug

The save file with the issue: [ghostmine_capture_sprit_bug.zip](https://github.com/ihhub/fheroes2/files/11823903/ghostmine_capture_sprit_bug.zip)

This PR: 
after_fix

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1712,8 +1712,8 @@ void Maps::Tiles::RestoreAbandonedMine( Tiles & tile, const int resource )
     auto restoreMineObjectType = [&tile]( int directionVector ) {
         if ( Maps::isValidDirection( tile._index, directionVector ) ) {
             Tiles & mineTile = world.GetTiles( Maps::GetDirectionIndex( tile._index, directionVector ) );
-            if ( ( mineTile._uid == tile._uid || mineTile.FindAddonLevel1( tile._uid ) || mineTile.FindAddonLevel2( tile._uid ) )
-                 && ( mineTile.GetObject() == MP2::OBJ_NON_ACTION_ABANDONED_MINE ) ) {
+            if ( ( mineTile.GetObject() == MP2::OBJ_NON_ACTION_ABANDONED_MINE )
+                 && ( mineTile._uid == tile._uid || mineTile.FindAddonLevel1( tile._uid ) || mineTile.FindAddonLevel2( tile._uid ) ) ) {
                 mineTile.SetObject( MP2::OBJ_NON_ACTION_MINES );
             }
         }
@@ -1728,11 +1728,12 @@ void Maps::Tiles::RestoreAbandonedMine( Tiles & tile, const int resource )
 
     if ( Maps::isValidDirection( tile._index, Direction::RIGHT ) ) {
         Tiles & rightTile = world.GetTiles( Maps::GetDirectionIndex( tile._index, Direction::RIGHT ) );
-        TilesAddon * addon = rightTile.FindAddonLevel1( tile._uid );
 
         if ( rightTile._uid == tile._uid ) {
             restoreRightSprite( rightTile._objectIcnType, rightTile._imageIndex );
         }
+
+        TilesAddon * addon = rightTile.FindAddonLevel1( tile._uid );
 
         if ( addon ) {
             restoreRightSprite( addon->_objectIcnType, addon->_imageIndex );
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1709,52 +1709,41 @@ void Maps::Tiles::RestoreAbandonedMine( Tiles & tile, const int resource )
         }
     };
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
I'm not sure how to better fix this because of addons. I suppose that's the OG editor's problem and there is no valid non-action object on this tile right at the start of this map?

oleg-derevenetz:
@Districh-ru Usually when one sprite is placed on top of another, one of them should go to the addons, so it would be good for @ihhub to take a look on this situation to ensure that there is no issues with addon mechanics (it may be a bit sophisticated).

Districh-ru:
> @Districh-ru Usually when one sprite is placed on top of another, one of them should go to the addons, so it would be good for @ihhub to take a look on this situation to ensure that there is no issues with addon mechanics (it may be a bit sophisticated).

Sure.)
The sprite of the Abandoned mine here is placed as the main sprite (ICN type 51, image index 7) and other sprites are in the addons and the tile object type is set according to the level 2 addon...

<details><summary>Tile info</summary>

```
******* Tile info *******
Tile index      : 4035, point: (3, 28)
UID             : 5166
MP2 object type : 100 (Mountains)
ICN object type : 51 (OBJNGRAS.ICN)
image index     : 7 (animated: 0)
layer type      : 0 - Object layer
region          : 0
ground          : Swamp (isRoad: 0)
shadow          : false
passable from   : nowhere
metadata value 1: 0
metadata value 2: 0
metadata value 3: 0
--------- Level 1 --------
UID             : 5132
ICN object type : 33 (TREJNGL.ICN)
image index     : 17
layer type      : 2 (2) - Shadow layer
is shadow       : yes
--------- Level 2 --------
UID             : 5169
ICN object type : 23 (MTNSWMP.ICN)
image index     : 53
layer type      : 2 (2) - Shadow layer
is shadow       : no
--- Extra information ---
*************************
```

</details>

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1718,13 +1718,13 @@ void Maps::Tiles::RestoreAbandonedMine( Tiles & tile, const int resource )
         Tiles & rightTile = world.GetTiles( Maps::GetDirectionIndex( tile._index, Direction::RIGHT ) );
         TilesAddon * addon = rightTile.FindAddonLevel1( tile._uid );
 
+        restoreRightSprite( rightTile._objectIcnType, rightTile._imageIndex );
```
ihhub, line 1721: The whole logic of Abandoned Mine restoration is not entirely correct. Instead of relying on object type and object index we should rely on object ID. We do this by searching an addon by UID but we don't check the same for non-addon object (top object). I assume that if there are multiple Mines covering each other this logic might produce invalid results.

ihhub (approved):

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1709,52 +1709,41 @@ void Maps::Tiles::RestoreAbandonedMine( Tiles & tile, const int resource )
         }
     };
 
+    auto restoreMineObjectType = [&tile]( int directionVector ) {
+        if ( Maps::isValidDirection( tile._index, directionVector ) ) {
+            Tiles & mineTile = world.GetTiles( Maps::GetDirectionIndex( tile._index, directionVector ) );
+            if ( ( mineTile._uid == tile._uid || mineTile.FindAddonLevel1( tile._uid ) || mineTile.FindAddonLevel2( tile._uid ) )
+                 && ( mineTile.GetObject() == MP2::OBJ_NON_ACTION_ABANDONED_MINE ) ) {
```
ihhub, line 1716: I suggest to swap both conditions in order to follow the principle of fast fail.

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1709,52 +1709,41 @@ void Maps::Tiles::RestoreAbandonedMine( Tiles & tile, const int resource )
...
     for ( TilesAddon & addon : tile.addons_level1 ) {
-        restoreLeftSprite( addon._objectIcnType, addon._imageIndex );
+        if ( addon._uid == tile._uid ) {
+            restoreLeftSprite( addon._objectIcnType, addon._imageIndex );
+        }
     }
 
     if ( Maps::isValidDirection( tile._index, Direction::RIGHT ) ) {
         Tiles & rightTile = world.GetTiles( Maps::GetDirectionIndex( tile._index, Direction::RIGHT ) );
         TilesAddon * addon = rightTile.FindAddonLevel1( tile._uid );
```
ihhub, line 1731: Please move this variable after the if-condition below to keep the code logically separate.

ihhub (commented):
Hi @Districh-ru , I left 2 little optimization comments. Could you please take a look at them?

ihhub (approved):

oleg-derevenetz (approved):

ihhub:

Issue #7364, opened 2023-06-28
Linux version of fheroes2_tools extracts fully transparent .png from .icn files


Districh-ru:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Linux

### Describe the bug

After extraction of HEROES2.agg file, I have a directory "icn/HEROES2" with lot of subdirectories inside it. Every subdirectory contains several .png files and one "offsets.txt". But none of *.png contains useful data - everything I looked at consists just of several transparent pixels. A couple of examples of such directories are attached. Maybe something went wrong on extraction? My system is Linux, so I use fheroes2_tools_ubuntu_x86-64_SDL2. There were no errors during extraction process, actually:
```
Total extracted sprites: 14444, failed sprites: 0
```

[curse.zip](https://github.com/ihhub/fheroes2/files/11891214/curse.zip)
[zombie.zip](https://github.com/ihhub/fheroes2/files/11891278/zombie.zip)

_Originally posted by @drevoborod in https://github.com/ihhub/fheroes2/issues/7316#issuecomment-1610882739_

PR #7365, opened 2023-06-28
Remove PNG support from *NIX tools packages

oleg-derevenetz:
Related to #7364

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -43,7 +43,6 @@ jobs:
           env:
             FHEROES2_STRICT_COMPILATION: ON
             FHEROES2_WITH_TOOLS: ON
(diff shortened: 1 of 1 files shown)


Districh-ru (approved):
👍 
BMP images are saved correctly by tools, checked on Ubuntu (Kubuntu).
This is a good workaround until we find how to fix the save of PNG files (by `fheroes2::Save()`) on *NIX systems.

oleg-derevenetz:
> This is a good workaround until we find how to fix the save of PNG files (by fheroes2::Save()) on *NIX systems.

I have an impression that SDL_image/libpng interaction is broken when 8-bit indexed surfaces are used. STB backend instead of libpng works just fine, but I suppose SDL_image is mostly built with libpng on various Linux distributions (at least on Ubuntu it is).

PR #7373, opened 2023-07-02
Expand gamepad / controller support for other platforms

### LINKED ISSUE #4435: Generic sdl2 game controller support on compatible devices
### Preliminary checks

This #ifdef restriction means that sdl2 game controller is only initialised on Vita or Switch, and not on other systems with compatible sdl2 game controllers. Perhaps game controller support can be made more generic.

https://github.com/ihhub/fheroes2/blob/c79c9866f821a92b73622aadf2cff45d2d98e2ae/src/fheroes2/game/fheroes2.cpp#L150

By removing this #ifdef I am able to initialise game controller support on linux handheld devices. However, I'm not sure if that's the best solution for all use cases.

### Additional info

https://github.com/ihhub/fheroes2/blob/c79c9866f821a92b73622aadf2cff45d2d98e2ae/src/fheroes2/game/fheroes2.cpp#L150

ihhub:
close #4435

We were facing issues with controllers when software cursor mode was not set. In the past we were toggling between hardware and software cursor modes. However, now if software cursor mode is set we never set if off. In this case it is safe to enable controller support for such cases.

This pull request requires extensive testing on multiple platforms before being merged.

### DIFF
--- a/src/engine/core.cpp
+++ b/src/engine/core.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2021 - 2022                                             *
+ *   Copyright (C) 2021 - 2023                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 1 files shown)


Sparticuz:
Is there an easy way to test this besides compiling? It looks like the github action expired so I can't get the build.

ihhub:
Closing this pull request as it doesn't solve the original problem.

PR #7376, opened 2023-07-03
Fix the object cache update when AI-controlled hero uses the Dimension Door spell

oleg-derevenetz:
Fix the assertion failure in the `master` branch on this line:

https://github.com/ihhub/fheroes2/blob/96a1ef0d61cad88760bb48dd93cfc6086f4d31e5/src/fheroes2/ai/normal/ai_normal_hero.cpp#L1673

How to reproduce:

1. Load the attached save file using the debug build;
2. Press F8 to transfer the control to AI;
3. Wait until the blue Wizard Caerlyn begins to attack the Necromancer's castle Chandler.

[Crash.zip](https://github.com/ihhub/fheroes2/files/11931896/Crash.zip)

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -2091,7 +2091,7 @@ namespace AI
             const size_t heroesBefore = heroes.size();
             _pathfinder.reEvaluateIfNeeded( *bestHero );
 
-            const int prevHeroPosition = bestHero->GetIndex();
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
PS Vita build currently fails because SDL_Mixer for PSV is built with `libxmp` support, but the `libxmp` itself is not installed by the Vitadev Package manager (VDPM). There's nothing I can do about it yet.

ihhub (approved):

ihhub:

PR #7378, opened 2023-07-03
Pass --no-convert to msgfmt

### LINKED ISSUE #7353: Build broken with gettext-0.22 because of ISO-8859-16
### Preliminary checks

### Platform

Linux

### Describe the bug

Here's what happens:

```
...
sed -e '1,20 s/UTF-8/ISO-8859-16/' ro.po | iconv -f utf-8 -t ISO-8859-16 | msgfmt - -o ro.mo
<stdin>: warning: Charset "ISO-8859-16" is not a portable encoding name.
                  Message conversion to user's charset might not work.
msgfmt: present charset "ISO-8859-16" is not a portable encoding name
make: *** [Makefile:55: ro.mo] Error 1
```

But it works with gettext-0.21.1

https://908863.bugs.gentoo.org/attachment.cgi?id=864273 has the full build log.

Calling it like `msgfmt - -o ro.mo --no-convert` fixes the build, but: a) I cannot verify that the result is correct due to lack of non-English data files, b) unconditionally adding that option breaks build with gettext-0.21.1: `msgfmt: unrecognized option '--no-convert'`

Ref https://bugs.gentoo.org/908863

### Save file

N/A

### Additional info

_No response_

DarthGandalf:
Close #7353

### DIFF
--- a/files/lang/Makefile
+++ b/files/lang/Makefile
@@ -19,7 +19,8 @@
 ###########################################################################
 
 ICONV  = iconv
-MSGFMT = sed -e '1,20 s/UTF-8/$(1)/' $< | $(ICONV) -f utf-8 -t $(1) | msgfmt - -o $@
+# TODO: consider converting game fonts and texts to UTF-8 in the engine instead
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

ihhub:
@DarthGandalf , a bunch of thanks for this fix!

PR #7316, opened 2023-06-15
Redo ICN decode code and agg_image.cpp code clean-up

Districh-ru:
This PR contains changes made in #7233
While working on #7233 there were made some necessary for that PR changes in the code, but PR became very big. To reduce the size of that PR and to properly test and merge some changes this PR is made. It contains:
- rework of **decodeICNSprite()** function to gain 1-30% speed-up while decoding ICN images;
- clean-up of `agg_image.cpp` to make Clang-Tidy a little happier;
- fix of  some images which were containing invalid transparent pixels (pixel count starts from 0x0 - upper left corner):
  - mini portraits of PoL heroes (ICN::**MINIPORT**):
    - Solmyr (ICN id 60, pixels 5x0, 10x10, 28x11,4x14),
    -  Danwin (ICN id 61, pixels 21x1, 20x2, 21x2, 23x2, 25x14),
    -  Gallavant (ICN id 65, pixels 19x16, 2x20, 1x21),
    -  Ceallach (ICN id 67, pixel 12x1);
  - Barbarian captain mini icon (ICN::**MINICAPT**, id 1, pixel 22x2);
  - Sorceress captain (ICN::**PORT0092**, id 0, pixels 8x20 and 8x66)
  - Necromancer (ICN::**PORT0095**, id 0, pixel 8x22) captain (shown in Castle dialog);
  - Sphere of Negation artifact (ICN::**ARTIFACT**, id 99, black pixels of the shadow);
  - ship battlefield background (ICN::**CBKGWATR**, id 0, pixel 125x36);
  - first image in "ICH::**BRCREST**" (id 0, pixel 30x5)  - icon of the blue player, shown during enemy turn;
  - Main Menu background (ICN::**HEROES**, id0, pixel at 260x305);
- use `fheroes2::Copy()` instead of `fheroes2::Blit()` while rendering the Armageddon spell.

PR #7383, opened 2023-07-04
Fix army button for evil interface

Districh-ru:
Fix regression after #7316
regression

This PR: 
fixed


drevoborod:
Tested 7316, but didn't know how to check all buttons if they have unexpected changes or not :( Looks like we need some kind of automation testing by screenshots comparison. Maybe screenshots of every graphical element should be taken once as reference and compared on every merge with same elements in the PR.

ihhub (approved):

ihhub:
@Districh-ru , thank you very much for this quick fix!

PR #7388, opened 2023-07-06
Add Dismiss hotkey to hero screen and fix last hotkey not showing up in the hotkey dialog

zenseii:
This makes it possible to press the Dismiss hotkey in the hero screen, by default "D".

Additionally it was discovered that the last hotkey didn't show up in the hotkey dialog due to both the call to `reserve()` and the for loop in `Game::getAllHotKeyEvents()`.

### DIFF
--- a/src/fheroes2/heroes/heroes_dialog.cpp
+++ b/src/fheroes2/heroes/heroes_dialog.cpp
@@ -316,8 +316,11 @@ int Heroes::OpenDialog( const bool readonly, const bool fade, const bool disable
 
         // button click
         le.MousePressLeft( buttonExit.area() ) ? buttonExit.drawOnPress() : buttonExit.drawOnRelease();
-        if ( buttonDismiss.isEnabled() )
-            le.MousePressLeft( buttonDismiss.area() ) ? buttonDismiss.drawOnPress() : buttonDismiss.drawOnRelease();
+        if ( buttonDismiss.isEnabled() && ( le.MousePressLeft( buttonDismiss.area() ) || HotKeyPressEvent( Game::HotKeyEvent::ARMY_DISMISS_TROOP ) ) )
+            buttonDismiss.drawOnPress();
+        else {
+            buttonDismiss.drawOnRelease();
+        }
         if ( buttonPrevHero.isEnabled() )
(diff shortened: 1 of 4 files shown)


src/fheroes2/heroes/heroes_dialog.cpp
```diff
@@ -316,8 +316,11 @@ int Heroes::OpenDialog( const bool readonly, const bool fade, const bool disable
 
         // button click
         le.MousePressLeft( buttonExit.area() ) ? buttonExit.drawOnPress() : buttonExit.drawOnRelease();
-        if ( buttonDismiss.isEnabled() )
-            le.MousePressLeft( buttonDismiss.area() ) ? buttonDismiss.drawOnPress() : buttonDismiss.drawOnRelease();
+        if ( buttonDismiss.isEnabled() && ( le.MousePressLeft( buttonDismiss.area() ) || HotKeyPressEvent( Game::HotKeyEvent::ARMY_DISMISS_TROOP ) ) )
```
ihhub, line 319: Should we rename **ARMY_DISMISS_TROOP** into more generic name and also update the description?

ihhub (commented):
Hi @zenseii , I left one question here. Could you please check it?

src/fheroes2/heroes/heroes_dialog.cpp
```diff
@@ -316,8 +316,11 @@ int Heroes::OpenDialog( const bool readonly, const bool fade, const bool disable
...
+            buttonDismiss.drawOnPress();
+        else {
+            buttonDismiss.drawOnRelease();
+        }
```
oleg-derevenetz, line 323: Hi @zenseii According to the logic before this change, we won't draw the Dismiss button at all if it is not enabled. So I suppose the logic should be like this:

```cpp
if ( buttonDismiss.isEnabled() ) {
  if ( le.MousePressLeft( buttonDismiss.area() ) || HotKeyPressEvent( Game::HotKeyEvent::ARMY_DISMISS_TROOP ) ) {
    buttonDismiss.drawOnPress();
  }
  else {
    buttonDismiss.drawOnRelease();
  }
}
```

What do you think?

oleg-derevenetz (approved):

ihhub (approved):

ihhub:

PR #7395, opened 2023-07-07
Minor ICN decode speed-up

Districh-ru:
This PR is a part of #7233 and it makes ICN decode a bit faster (about 2-3 %).

### DIFF
--- a/src/engine/image_tool.cpp
+++ b/src/engine/image_tool.cpp
@@ -355,8 +355,15 @@ namespace fheroes2
                     break;
                 }
 
-                memcpy( imageData + posX, data, pixelCount );
-                memset( imageTransform + posX, static_cast<uint8_t>( 0 ), pixelCount );
+                if ( pixelCount > 1 ) {
+                    memcpy( imageData + posX, data, pixelCount );
+                    memset( imageTransform + posX, static_cast<uint8_t>( 0 ), pixelCount );
+                }
+                else {
+                    *( imageData + posX ) = *data;
+                    *( imageTransform + posX ) = static_cast<uint8_t>( 0 );
+                }
+
                 data += pixelCount;
                 posX += pixelCount;
             }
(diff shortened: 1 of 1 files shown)


src/engine/image_tool.cpp
```diff
@@ -355,8 +355,15 @@ namespace fheroes2
...
-                memcpy( imageData + posX, data, pixelCount );
-                memset( imageTransform + posX, static_cast<uint8_t>( 0 ), pixelCount );
+                if ( pixelCount > 1 ) {
+                    memcpy( imageData + posX, data, pixelCount );
+                    memset( imageTransform + posX, static_cast<uint8_t>( 0 ), pixelCount );
+                }
+                else {
+                    *( imageData + posX ) = *data;
+                    *( imageTransform + posX ) = static_cast<uint8_t>( 0 );
+                }
```
ihhub, line 365: Is it really faster to have a distinguish code for the case of 1 byte? Wouldn't it be slower as usually this command (repeat pixels N times) is usually executed for the number of pixels more than 1.

src/engine/image_tool.cpp
```diff
@@ -371,12 +378,18 @@ namespace fheroes2
...
+                const uint32_t pixelCount = ( countValue != 0 ) ? countValue : *( ++data );
 
                 if ( ( transformValue & 0x40 ) && ( transformType <= 15 ) ) {
-                    memset( imageTransform + posX, transformType, pixelCount );
+                    if ( pixelCount > 1 ) {
+                        memset( imageTransform + posX, transformType, pixelCount );
+                    }
+                    else {
+                        *( imageTransform + posX ) = transformType;
+                    }
```
ihhub, line 392: The same comment as above. I am not entirely sure that we would have any benefit in performance by doing this.

ihhub (changes requested):
Hi @Districh-ru , I left few comments in this pull request. Could you please take a look when you have time?

src/engine/image_tool.cpp
Districh-ru: Hi @ihhub.
In this part only in 15% cases `pixelCount` is equal to 1.
I did test more thorough test by doing many measures of `decodeICNSprite()` computation time doing these actions: start fheroes2, load a save of the "Grand War II" map and cast View All spell (to load all map objects).
And computation speed with these changes and without them is almost the same.
So not to overcomplicate the code I revert these changes.

src/engine/image_tool.cpp
Districh-ru: In this part in 45% cases `pixelCount` is equal to 1.
But here, like in the case above, the computation speed with these changes and without them is also almost the same.
I also revert these changes.

ihhub (approved):

ihhub:

PR #7396, opened 2023-07-08
Avoid redrawing all Castle Well screen elements while performing monsters animation to reduce CPU usage

Districh-ru:
This PR optimizes the CPU use while rendering the Castle Well: cache the background image not to perform full render of all icons and texts for every creature animation frame.


src/fheroes2/castle/castle_well.cpp
```diff
@@ -303,248 +300,288 @@ void Castle::OpenWell()
     }
 }
 
-void Castle::WellRedrawInfoArea( const fheroes2::Point & cur_pt, const std::vector<fheroes2::RandomMonsterAnimation> & monsterAnimInfo ) const
+void Castle::_wellRedrawAvailableMonsters( const uint32_t dwellingType, const bool restoreBackground, fheroes2::Image & background ) const
 {
-    fheroes2::Display & display = fheroes2::Display::instance();
-    const bool isEvilInterface = Settings::Get().isEvilInterfaceEnabled();
+    if ( !( building & dwellingType ) ) {
+        // This building is not build.
```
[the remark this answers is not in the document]
Districh-ru: Hi, @zenseii!
Thank you for testing and for spell checking. :)

The vertical line to the right of EXIT button is fixed. In original assets this line is transparent. To speed-up rendering I did not clear the made image before copying Well background assuming that all transparent parts will be covered with building's images and buttons. Now this line is filled with black color.

src/fheroes2/castle/castle_well.cpp
```diff
@@ -254,41 +254,38 @@ void Castle::OpenWell()
...
+        }
+
+        for ( size_t dwellingId = 0; dwellingId < CASTLEMAXMONSTER; ++dwellingId ) {
+            if ( building & castleDwellings[dwellingId] ) {
+                if ( le.MouseClickLeft( rectMonster[dwellingId] ) || pressedHotkeyBuildingID == castleDwellings[dwellingId] ) {
+                    if ( RecruitMonster( Dialog::RecruitMonster( castleMonster[dwellingId], dwelling[dwellingId], true, 2 ) ) ) {
+                        // Update available monster count on background.
+
+                        _wellRedrawAvailableMonsters( castleDwellings[dwellingId], true, background );
+                    }
```
ihhub, line 273: I think we need to put a break here to avoid exit the loop.

src/fheroes2/castle/castle_well.cpp
```diff
@@ -254,41 +254,38 @@ void Castle::OpenWell()
...
+                }
+                else if ( le.MousePressRight( rectMonster[dwellingId] ) ) {
+                    Dialog::DwellingInfo( castleMonster[dwellingId], dwelling[dwellingId] );
```
ihhub, line 276: I think we need to put a break here to avoid exit the loop.

src/fheroes2/castle/castle_well.cpp
```diff
@@ -303,248 +300,291 @@ void Castle::OpenWell()
...
+
+            text.set( textString, statsFontType );
+            text.draw( renderPoint.x - text.width() / 2, renderPoint.y, background );
+
+            _wellRedrawAvailableMonsters( dwellingType, false, background );
         }
+    }
+}
+
+void Castle::_wellRedrawMonsterAnimation( const fheroes2::Point & cur_pt, std::array<fheroes2::RandomMonsterAnimation, CASTLEMAXMONSTER> & monsterAnimInfo ) const
```
ihhub, line 532: Since this is a new method please name variables using camelCase naming.

ihhub (changes requested):
Hi @Districh-ru , I left few comments in this pull request. Would you mind to check them?

src/fheroes2/castle/castle_well.cpp
Districh-ru: Sure, as we cannot do left click on more than one dwelling at one time. :)

src/fheroes2/castle/castle_well.cpp
Districh-ru: Yes, we also cannot right click more than one dwelling at once.

src/fheroes2/castle/castle_well.cpp
Districh-ru: Thanks for noticing this!

ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for polishing the engine!

PR #7399, opened 2023-07-09
Add ability to select instrument brush and brush size in Editor UI

Districh-ru:
This PR relates to https://github.com/ihhub/fheroes2/issues/6845 and adds Map Editor UI ability to:
- [x] select Brush size;
- [x] add right click tips for Brush size;
- [x] select terrain/object type;
- [x] add right click tips for terrain/object type;
- [x] add text for selected terrain/object type.

fheroes2 2023-07-12 18-46-28-118 изображение


src/fheroes2/editor/editor_interface_panel.h
```diff
@@ -59,11 +107,17 @@ namespace Interface
...
 
         // Index of selected Map Editor instrument.
-        uint8_t _selectedInstrument{ 0 };
-        const static uint8_t instrumentsCount = 6;
+        uint8_t _selectedInstrument{ Instrument::TERRAIN };
+
+        std::array<fheroes2::Button, Instrument::INSTRUMENTS_COUNT> _instrumentButtons;
+        std::array<fheroes2::Rect, Instrument::INSTRUMENTS_COUNT> _instrumentButtonsRect;
+
+        uint8_t _selectedBrush{ Brush::WATER };
```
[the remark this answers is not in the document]
Districh-ru: It will be used in the next commit. :)

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +301,141 @@ namespace Interface
...
+                fheroes2::showStandardTextMessage( _( "Monsters" ), _( "Used to place a monster group." ), Dialog::ZERO );
+            }
+            else if ( le.MousePressRight( _objectButtonsRect[Brush::HEROES] ) ) {
+                fheroes2::showStandardTextMessage( _( "Heroes" ), _( "Used to place a hero." ), Dialog::ZERO );
+            }
+            else if ( le.MousePressRight( _objectButtonsRect[Brush::ARTIFACTS] ) ) {
+                fheroes2::showStandardTextMessage( _( "Artifacts" ), _( "Used to place an artifact." ), Dialog::ZERO );
+            }
+            else if ( le.MousePressRight( _objectButtonsRect[Brush::TREASURES] ) ) {
+                fheroes2::showStandardTextMessage( _( "Treasures" ), _( "Used to place a resource or treasure." ), Dialog::ZERO );
```
ihhub, line 435: Should we extra the text from these lines and from lines 222-269 into a container and use it?

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +301,141 @@ namespace Interface
...
+            }
+            else if ( le.MousePressRight( _terrainButtonsRect[Brush::DIRT] ) ) {
+                fheroes2::showStandardTextMessage( _( "Dirt" ), _( "No special modifiers." ), Dialog::ZERO );
+            }
+            else if ( le.MousePressRight( _terrainButtonsRect[Brush::WASTELAND] ) ) {
+                fheroes2::showStandardTextMessage( _( "Wasteland" ), movePenaltyText( "1.25" ), Dialog::ZERO );
+            }
+            else if ( le.MousePressRight( _terrainButtonsRect[Brush::BEACH] ) ) {
+                fheroes2::showStandardTextMessage( _( "Beach" ), movePenaltyText( "1.25" ), Dialog::ZERO );
+            }
```
ihhub, line 378: We are repeating the same text as lines 177-210. Should we store it in a separate container?

ihhub (changes requested):
Hi @Districh-ru , I left multiple comments here. Could you please check them?

ihhub:
Hi @Districh-ru , I think creating separate functions to return text instead of the proposed container would be even better solution.

Districh-ru:
> Hi @Districh-ru , I think creating separate functions to return text instead of the proposed container would be even better solution.

Hi, @ihhub, sure. Initially I was thinking about making such function.. :)
I'll address your comments a little later.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +157,129 @@ namespace Interface
...
+        }
+
+        if ( _selectedInstrument == Instrument::TERRAIN ) {
+            const fheroes2::Sprite & selection = fheroes2::AGG::GetICN( ICN::TERRAINS, 9 );
+            fheroes2::Blit( selection, 0, 0, display, _terrainButtonsRect[_selectedTerrain].x - 2, _terrainButtonsRect[_selectedTerrain].y - 2, selection.width(),
+                            selection.height() );
+
+            std::string text;
+            switch ( _selectedTerrain ) {
```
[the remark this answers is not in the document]
Districh-ru: Sure, let's reduce the repetitions.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +283,116 @@ namespace Interface
             }
         }
 
+        if ( _selectedInstrument == Instrument::TERRAIN || _selectedInstrument == Instrument::ERASE ) {
+            for ( uint8_t i = 0; i < BrushSize::BRUSH_SIZE_COUNT; ++i ) {
```
ihhub, line 287: I suggest to avoid **BrushSize::BRUSH_SIZE_COUNT** here and use **_brushSizeButtonsRect.size()** as we handle std::array so the method is inlined. This will make the code more flexible for any future (if any) changes.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -54,14 +58,22 @@ namespace Interface
...
+        _instrumentButtons[_selectedInstrument].press();
+        _brushSizeButtons[_selectedBrushSize].press();
     }
 
     void EditorPanel::setPos( const int32_t displayX, int32_t displayY )
     {
         int32_t offsetX = displayX;
 
-        for ( uint8_t i = 0; i < instrumentsCount; ++i ) {
+        for ( uint8_t i = 0; i < Instrument::INSTRUMENTS_COUNT; ++i ) {
```
ihhub, line 76: I suggest to avoid **Instrument::INSTRUMENTS_COUNT** here and use **_instrumentButtons.size()** as we handle std::array so the method is inlined. This will make the code more flexible for any future (if any) changes.

This is also applicable for any other cases where we use loops in the code within this file. This way we would also reduce the risk of mistakes in the code.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +156,112 @@ namespace Interface
...
         _buttonUndo.draw();
         _buttonNew.draw();
         _buttonSpecs.draw();
         _buttonFile.draw();
         _buttonSystem.draw();
+
+        display.render( _rectInstrumentPanel );
+    }
+
+    int EditorPanel::_getGroundId( uint8_t brushId )
```
ihhub, line 199: This variable can be **const**.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +156,112 @@ namespace Interface
...
+        case Brush::BEACH:
+            return Maps::Ground::BEACH;
+        default:
+            break;
+        }
+
+        return Maps::Ground::UNKNOWN;
+    }
+
+    std::string EditorPanel::_getObjectTypeName( uint8_t brushId )
```
ihhub, line 227: This variable can be **const**.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +156,112 @@ namespace Interface
...
+        case Brush::DESERT:
+            return Maps::Ground::DESERT;
+        case Brush::DIRT:
+            return Maps::Ground::DIRT;
+        case Brush::WASTELAND:
+            return Maps::Ground::WASTELAND;
+        case Brush::BEACH:
+            return Maps::Ground::BEACH;
+        default:
+            break;
```
ihhub, line 221: Should we assert or it is okay to handle other cases?

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +156,112 @@ namespace Interface
...
+        if ( _selectedInstrument == Instrument::TERRAIN ) {
+            const fheroes2::Sprite & selection = fheroes2::AGG::GetICN( ICN::TERRAINS, 9 );
+            fheroes2::Blit( selection, 0, 0, display, _terrainButtonsRect[_selectedTerrain].x - 2, _terrainButtonsRect[_selectedTerrain].y - 2, selection.width(),
+                            selection.height() );
+
+            const fheroes2::Text terrainText( Maps::Ground::String( _getGroundId( _selectedTerrain ) ), fheroes2::FontType::smallWhite() );
+            terrainText.draw( _rectInstrumentPanel.x + 72 - terrainText.width( 118 ) / 2, _rectInstrumentPanel.y + 107, display );
+        }
+
+        if ( _selectedInstrument == Instrument::OBJECT ) {
```
ihhub, line 180: Since we are not changing **_selectedInstrument** variable inside the above if-body then we should use **else** condition:
```cpp
if ( _selectedInstrument == Instrument::TERRAIN ) {
...
}
else if ( _selectedInstrument == Instrument::OBJECT ) {
...
}
```

This way it would be slightly faster :)

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +283,116 @@ namespace Interface
...
 
+        if ( _selectedInstrument == Instrument::TERRAIN || _selectedInstrument == Instrument::ERASE ) {
+            for ( uint8_t i = 0; i < BrushSize::BRUSH_SIZE_COUNT; ++i ) {
+                if ( le.MousePressLeft( _brushSizeButtonsRect[i] ) ) {
+                    if ( _brushSizeButtons[i].drawOnPress() ) {
+                        _selectedBrushSize = i;
+                    }
+                }
+                else {
+                    if ( i != _selectedBrushSize ) {
```
ihhub, line 294: Please combine else with if. SonarQube would be happy about this.

src/fheroes2/editor/editor_interface_panel.cpp
ihhub, line 199: I suggest to return `const char*` (or std::string) by this function and call **Maps::Ground::String()** internally. In this case we don't need to call **Maps::Ground::String** anywhere else in this code within this file, this function will return similar value as **_getObjectTypeName()** and we can avoid hardcoding values in line 334-360.

ihhub (changes requested):
Hi @Districh-ru , I added several comments here. Could you please check them when you have time and see if them make sense?

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Hi @ihhub, sure, let's make the code more flexible.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: OK)

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Sure, let's leave an assertion that if a new area is added, we must remember to give it a name.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Sure. :)

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: OK)

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -54,14 +59,22 @@ namespace Interface
...
+        _instrumentButtons[_selectedInstrument].press();
+        _brushSizeButtons[_selectedBrushSize].press();
     }
 
     void EditorPanel::setPos( const int32_t displayX, int32_t displayY )
     {
         int32_t offsetX = displayX;
 
-        for ( uint8_t i = 0; i < instrumentsCount; ++i ) {
+        for ( uint8_t i = 0; i < static_cast<uint8_t>( _instrumentButtonsRect.size() ); ++i ) {
```
ihhub, line 77: We can just use **size_t** without the need of casting.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -83,8 +96,31 @@ namespace Interface
         const fheroes2::Sprite & instrumentPanel = fheroes2::AGG::GetICN( ICN::EDITPANL, _selectedInstrument );
         _rectInstrumentPanel = { displayX, displayY, instrumentPanel.width(), instrumentPanel.height() };
 
-        // System buttons top row.
+        offsetX = displayX + 14;
+        int32_t offsetY = displayY + 128;
+        for ( uint8_t i = 0; i < static_cast<uint8_t>( _brushSizeButtonsRect.size() ); ++i ) {
```
ihhub, line 101: We can just use **size_t** without the need of casting.

src/fheroes2/editor/editor_interface_panel.h
```diff
@@ -58,12 +111,24 @@ namespace Interface
...
         Editor & _interface;
 
+        static const char * _getTerrainTypeName( const uint8_t brushId );
+
+        static const char * _getObjectTypeName( const uint8_t brushId );
+
         // Index of selected Map Editor instrument.
-        uint8_t _selectedInstrument{ 0 };
-        const static uint8_t instrumentsCount = 6;
+        uint8_t _selectedInstrument{ Instrument::TERRAIN };
```
ihhub, line 119: I suggest to move all new member at the end of the class and put all **uint8_t** members at the end. In this case the class will be smaller by size due to padding rules.

src/fheroes2/editor/editor_interface_panel.h
```diff
@@ -31,6 +31,59 @@ namespace Interface
 {
     class Editor;
 
+    enum Instrument : uint8_t
```
ihhub, line 34: Are we going to use these enumerations somewhere else outside this class? If not let's declare then within the class in **private** section (just before any member) to reduce their scope and accessibility.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +296,114 @@ namespace Interface
...
+            else if ( le.MousePressRight( _brushSizeButtonsRect[BrushSize::LARGE] ) ) {
+                fheroes2::showStandardTextMessage( _( "Large Brush" ), brushSizeText( 4 ), Dialog::ZERO );
+            }
+            else if ( le.MousePressRight( _brushSizeButtonsRect[BrushSize::AREA] ) ) {
+                fheroes2::showStandardTextMessage( _( "Area Fill" ), _( "Used to click and drag for clearing large areas." ), Dialog::ZERO );
+            }
+        }
+
+        if ( _selectedInstrument == Instrument::TERRAIN ) {
+            for ( uint8_t i = 0; i < static_cast<uint8_t>( _terrainButtonsRect.size() ); ++i ) {
```
ihhub, line 332: I suggest to keep **size_t** here and only later at the time of assignment to **_selectedTerrain** convert it. The same applies for other similar places.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +296,114 @@ namespace Interface
...
+                    _selectedObject = i;
+                    _redraw();
+                }
+            }
+
+            for ( uint8_t objectId = Brush::WATER; objectId < Brush::TOWNS; ++objectId ) {
+                if ( le.MousePressRight( _objectButtonsRect[objectId] ) ) {
+                    std::string text = _( "Used to select objects most appropriate for use on %{terrain}." );
+                    StringReplaceWithLowercase( text, "%{terrain}", _getTerrainTypeName( objectId ) );
+                    fheroes2::showStandardTextMessage( _getObjectTypeName( objectId ), text, Dialog::ZERO );
```
ihhub, line 386: Should we add **break** here since only one element can active mouse right clicking?

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +157,124 @@ namespace Interface
...
+            return _( "Artifacts" );
+        case Brush::TREASURES:
+            return _( "Treasures" );
+        default:
+            // Have you added a new object type?
+            assert( 0 );
+            break;
+        }
+
+        return {};
```
ihhub, line 277: fheroes2::Text class does not accept **nullptr** as it handle it as an error which is valid since we don't want to have empty texts. This line initiates default constructor for `const char*` which is in fact 0 or `nullptr`. To keep the code aligned with **_getTerrainTypeName** method I recommend to return something like **"Unknown object type"**. It would be easier to catch such an error.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +296,114 @@ namespace Interface
...
+            else if ( le.MousePressRight( _brushSizeButtonsRect[BrushSize::AREA] ) ) {
+                fheroes2::showStandardTextMessage( _( "Area Fill" ), _( "Used to click and drag for clearing large areas." ), Dialog::ZERO );
+            }
+        }
+
+        if ( _selectedInstrument == Instrument::TERRAIN ) {
+            for ( uint8_t i = 0; i < static_cast<uint8_t>( _terrainButtonsRect.size() ); ++i ) {
+                if ( ( _selectedTerrain != i ) && le.MousePressLeft( _terrainButtonsRect[i] ) ) {
+                    _selectedTerrain = i;
+                    _redraw();
```
ihhub, line 335: Should we add **break** here since we expect only one button to be pressed?

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -152,6 +296,114 @@ namespace Interface
...
+            else if ( le.MousePressRight( _terrainButtonsRect[Brush::BEACH] ) ) {
+                fheroes2::showStandardTextMessage( _getTerrainTypeName( Brush::BEACH ), movePenaltyText( "1.25" ), Dialog::ZERO );
+            }
+        }
+
+        if ( _selectedInstrument == Instrument::OBJECT ) {
+            for ( uint8_t i = 0; i < static_cast<uint8_t>( _objectButtonsRect.size() ); ++i ) {
+                if ( ( _selectedObject != i ) && le.MousePressLeft( _objectButtonsRect[i] ) ) {
+                    _selectedObject = i;
+                    _redraw();
```
ihhub, line 378: Should we add **break** here since we expect only one button to be pressed?

ihhub (changes requested):
Hi @Districh-ru , I left few more comments in this pull request. My apologies if it takes too much time. Could you please take a look at the comment again and validate if they make sense?

src/fheroes2/editor/editor_interface_panel.h
Districh-ru: Hi @ihhub, for now we use them only in `EditorPanel` class. Later if we may want to use them outside of class we can move the needed enumeration outside of class and possibly to other H file (e.g. editor.h).
So let's put them now in the private section of class.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Sure)

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Sure)

src/fheroes2/editor/editor_interface_panel.h
Districh-ru: Sure, it will also arrange class members by type to make it easier to maintain the class in future.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: OK)

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Yes, it will reduce unneeded computations.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Thanks for the explanation!
Let's return "Unknown object type".

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Yes)

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Yes)

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -121,16 +158,124 @@ namespace Interface
...
+
+        if ( _selectedInstrument == Instrument::TERRAIN ) {
+            const fheroes2::Sprite & selection = fheroes2::AGG::GetICN( ICN::TERRAINS, 9 );
+            fheroes2::Blit( selection, 0, 0, display, _terrainButtonsRect[_selectedTerrain].x - 2, _terrainButtonsRect[_selectedTerrain].y - 2, selection.width(),
+                            selection.height() );
+
+            const fheroes2::Text terrainText( _getTerrainTypeName( _selectedTerrain ), fheroes2::FontType::smallWhite() );
+            terrainText.draw( _rectInstrumentPanel.x + 72 - terrainText.width() / 2, _rectInstrumentPanel.y + 107, display );
+        }
+
```
ihhub, line 181: Please remove this empty line.

ihhub (approved):

ihhub:
@drevoborod , could you please test these changes? You will need a debug version of the game. Press E in the Main Menu to open the Editor.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Sure)

drevoborod:
Checked the following: 

- [x] select terrain/object type;
- [x] select Brush size;
- [x] add right click tips for Brush size;
- [x] add right click tips for terrain/object type;
- [x] add text for selected terrain/object type.

@Districh-ru , no issues, but:
1. is it OK that no type of selected brush actually allows to draw anything? And same to object selection - actual objects selection window does not appear on clicking to corresponding buttons?
2. Is this text correct? Actually this unlimited size brush can be used not for cleaning but for creation of large size areas. 
изображение
And in formal editor2.exe there is other text:
изображение
Looks like you'we taken this text from the case when eraser is selected. In formal editor2.exe these two texts differ depending on what is selected: some terrain or the eraser, and in current editor's revision there is no separate text.
3. By the way, should all these texts be same as in formal editor? Some of them differ.
4. Seems like in these messages there should be a non-breaking space between an article and a noun: 
изображение
изображение
изображение
изображение
изображение

Districh-ru:
Hi @ihhub and @drevoborod, I'll address points 2 and 4 today a little later.

Districh-ru:
Hi @drevoborod, the brush size texts are now separated to draw/erase and a/an article in object type descriptions is moved to a new line.
Could you please check these changes when you have time.

drevoborod:
> the brush size texts are now separated to draw/erase and a/an article in object type descriptions is moved to a new line.

Checked it. Now it seems OK for me. But I noticed another minor issue. It's up to you to fix it or not:
изображение
I suggest to place "4 by 4" and similar texts on the same line, so possibly it could be better to leave them on the first line.

ihhub (approved):

ihhub:
@Districh-ru , well done!

PR #7401, opened 2023-07-11
Allow to lower difficulty down of campaign during gameplay

### LINKED ISSUE #6922: Allow to lower difficulty of campaign during gameplay
### Preliminary checks

Some players faced a situation when a chosen difficulty level for a campaign is too high for them. Right now it is not possible to change campaign's difficulty unless you restart the campaign.

We should give an ability to lower campaign difficulty through the gameplay. Lowering it down reduces the final score and also gives an  opportunity to finish the campaign without restarting it, especially if you are on the last stages of a campaign.

### Additional info

_No response_

ihhub:
close #6922

This would help some players to succeed with their campaign.

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -1534,15 +1534,17 @@ fheroes2::GameMode Game::SelectCampaignScenario( const fheroes2::GameMode prevMo
                                    fheroes2::Text( _( "Restart the current scenario." ), fheroes2::FontType::normalWhite() ), Dialog::ZERO );
         }
         else if ( ( buttonOk.isEnabled() && ( le.MouseClickLeft( buttonOk.area() ) || HotKeyPressEvent( HotKeyEvent::DEFAULT_OKAY ) ) ) || restartButtonClicked ) {
-            if ( restartButtonClicked ) {
-                if ( fheroes2::showStandardTextMessage( _( "Restart" ), _( "Are you sure you want to restart this scenario?" ), Dialog::YES | Dialog::NO ) == Dialog::NO ) {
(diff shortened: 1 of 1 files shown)


Branikolog:
Hi, @ihhub !
I've made several checks and here're my thoughts:
Once you've entered scenario briefing menu in the middle of scenario gameplay and changed difficulty level you cannot roll it back. I mean for the cases when you accept lower difficulty but didn't restart the scenario and just cancelled the briefing menu. The game continues showing the selected lower level in that menu. Only reloading of the save game helps to bring back the difficulty level you're currently playing on.
Also for the briefing **before** starting the particular scenario (when loading the _complete_ save file or when you've just finished previous scenario) I think it should be possible to switch between current difficulty level and the lower one until I start the current scenario. Before starting a scenario player could select a lower difficulty level, but then change his mind and decide to play it on the current one.
Also there's definitely lacking a warning message, that once you've restarted the scenario with lower difficulty selected, you won't be able to set a higher difficulty later and the campaign results will correspond the new difficulty level you've selected no matter how many scenarios you've finished previously on a higher difficulty level..

Branikolog:
Hi, @ihhub 
While starting a new campaign from the 1st scenario I can change difficulty level back to hard. Also I don't see a warning message if I'm going to select a lower difficulty. Was it intended to behave like that?

ihhub:
> Hi, @ihhub While starting a new campaign from the 1st scenario I can change difficulty level back to hard. Also I don't see a warning message if I'm going to select a lower difficulty. Was it intended to behave like that?

Hi @Branikolog , yes it is. Since this is a first scenario of the campaign you don't need such a confirmation.

Branikolog:
@ihhub 
While selecting lower difficulty and restarting without warning a player could be confused, that he can do the same in any other scenario. I think we shouldn't allow player to do this, since he can always return to main menu and restart campaign choosing the difficulty for the first scenario any time.

Branikolog (approved):

PR #7408, opened 2023-07-12
Improve AI pathfinding logic regarding protected and guarded tiles

### LINKED ISSUE #6969: Fix invalid pathfinding for AI heroes
### Preliminary checks

### Platform

Windows

### Describe the bug

Based on comment [here](https://github.com/ihhub/fheroes2/pull/6959#discussion_r1161099317).

The current implementation of pathfinding contains an idea that AI heroes can go "through" neutral monsters creating complex paths like capturing a mine and defeating monsters on the way. If neutral monsters are too powerful for a hero than tiles reachable for monsters are marked as inaccessible. This brings problem that heroes might attack these more powerful monsters based on logic in **ai_normal_hero.cpp**. However, at the end they aren't able to reach them.

We need to rework pathfinding logic for AI to avoid these cases.

### Save file

No need.

### Additional info

_No response_

### LINKED ISSUE #7406: Adventure map, AI behavior, AI hero attacks the castle if there is no defending hero, but does not attack the castle if there is a defending hero (1.0.5 7355)
### Preliminary checks

In fheroes2, the AI hero attacks the castle if there is no defending hero, but does not attack the castle if there is a defending hero.
The defending creatures are the same (6 bone dragons), and the AI hero has no chance of winning in either case.

Here are the save files (end the turn):
[A glorious victory text 01.zip](https://github.com/ihhub/fheroes2/files/12028382/A.glorious.victory.text.01.zip)

The AI's assessment of enemy strength needs to be improved.

### Additional info

_No response_

oleg-derevenetz:
fix #6969
close #7406

In `AIWorldPathfinder` there are two types of checks for tiles:

1. Whether the tile is accessible in principle (`isTileAccessibleForAIWithArmy()`). In `master` branch, any tile with "strong enough" hostile army on it is considered as "non-accessible". This includes both monsters and guarded objects (such as Abandoned Mine, Graveyard and even artifacts with `FIGHT_*` capture conditions). In this PR, only tiles that are guarded by wandering monsters (i.e. tiles next to these monsters), which are strong enough in comparison with the hero army, are considered as non-accessible. Tiles with wandering monsters itself, as well as tiles with objects which has their own guards (such as Graveyard), are always considered as accessible from the pathfinder's point of view. It gives a lot more freedom to the "high-level" AI logic.
2. Whether the tile can be "walked through" (`isTileAvailableForWalkThroughForAIWithArmy()`) - i.e. AI pathfinder continues to pave the path through this tile as if it doesn't contain the obstructing object (such as enemy hero, monster or some pickupable object). This logic is improved as well in this PR: tiles with artifacts that have "special" capture conditions are not considered as "transparent" from the AI pathfinder's point of view anymore, almost any object that could be picked up (not just artifacts) is now considered as "transparent", and a few more small changes.

Also this PR fixes some potential issues with Town Gate and Town Portal spells - for example, in `master` branch, if AI calculates the pathfinder cache for some hero with Town Portal spell and then hires a new hero in some castle, this castle is still considered as available as the Town Portal target even after the `AIWorldPathfinder::reEvaluateIfNeeded()` call for the same hero, because the cache is not updated in this case. Assertions have also been added to detect the issues of such kind.


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -605,7 +605,7 @@ namespace
         }
 
         case MP2::OBJ_CASTLE:
-            return AIShouldVisitCastle( hero, index, heroArmyStrength ) || ai.isPriorityTask( index );
+            return AIShouldVisitCastle( hero, index, heroArmyStrength );
```
oleg-derevenetz, line 608: This change closes the #7406 - `AIShouldVisitCastle()` always returns `true` for non-occupied castles that belong to this hero's kingdom, so it is redundant for the purposes of #7265, but if the castle in question is hostile for a hero, then the hero will try to capture it even if his army is obviously weak.

ihhub: Are we making AI heroes kamikaze towards enemy's castles?

oleg-derevenetz: They were kamikazes before this change in this PR, now they are not.

oleg-derevenetz: @ihhub the thing is that before this change AI heroes tend to attack the castles designated as priority targets even when heroes were too weak, hence the behavior from #7406.

ihhub: Ah, got it. I misunderstood.

ihhub:
Hi @drevoborod , could you please help to test this pull request?

oleg-derevenetz:
Hi @drevoborod this PR addresses AI behavior, so attention should be paid to various suspicious anomalies in the AI behavior, such as "AI heroes constantly attack too strong guarded object and being killed" or "AI does not attack a poorly guarded, but valuable object", that sort of thing.

idshibanov (approved):
Haven't tested it extensively, but approach looks correct.

drevoborod:
Just to mention: testing in progress :) Nothing suspicious found yet but I want to examine it much deeper.

drevoborod:
Just to mention: testing in progress :) Nothing suspicious found yet but I want to examine it much deeper.

drevoborod:
Game crashes after trying to get back control from AI. Fails after a hero collects any sea object.

```
----------
Flotsam


You search through the flotsam, but find nothing.
Press RETURN to choose OK.
----------
fheroes2: ../fheroes2/ai/normal/ai_normal_hero.cpp:648: bool {anonymous}::HeroesValidObject(const Heroes&, double, int32_t, const AIWorldPathfinder&, AI::Normal&, double): Assertion `0' failed.
zsh: IOT instruction (core dumped)  .

```

[AUTOSAVE_132.zip](https://github.com/ihhub/fheroes2/files/12087859/AUTOSAVE_132.zip)

**Steps to reproduce:**
1. Load provided savegame.
2. Provide control to AI (by pressing F8).
3. Wait for next turn and take control back (F8 again).
AI hero will move to the object on the water and collect it, so corresponding dialogue will appear. After pressing OK in the dialogue, the game will crash.

oleg-derevenetz:
> Game crashes after trying to get back control from AI.

This is a long-known issue not related to this PR (it can be reproduced on `master` branch as well). It can be fixed later in a separate PR.

drevoborod:
Testing finished. Haven't noticed any serious issues.
- allowed AI to play for ~170 game days and recorded a video (5 hours). Watched it partly, haven't noticed anything interesting. Also watched how the AI used "Town portal" spell - seems not optimal but OK.
- created several simple maps with different obstacles and with different enemy player's armies - stronger, weaker, equal. The AI behaviour seems to be adequate.

oleg-derevenetz:
@drevoborod thank you very much!

ihhub:

PR #7417, opened 2023-07-16
Disable Campaign button if no campaign resources are present

### LINKED ISSUE #7414: Disable Campaign option for loading saves if campaigns aren't supported
### Preliminary checks

At the moment a user can play campaigns by loading save files even if campaigns are not supported. This should not happen.

Disable campaign button in Load Game menu if campaigns are not supported.

### Additional info

_No response_

ihhub:
close #7414

This is easily verified on the demo version of the game.

### DIFF
--- a/src/fheroes2/game/game_loadgame.cpp
+++ b/src/fheroes2/game/game_loadgame.cpp
@@ -156,7 +156,7 @@ fheroes2::GameMode Game::LoadGame()
     fheroes2::Button buttonMultiplayerGame( 0, 0, ICN::BUTTON_MULTIPLAYER_GAME, 0, 1 );
     fheroes2::Button buttonCancel( 0, 0, ICN::BUTTON_LARGE_CANCEL, 0, 1 );
 
-    std::array<fheroes2::ButtonBase*, 4> buttons{ &buttonStandardGame, &buttonCampaignGame, &buttonMultiplayerGame, &buttonCancel };
+    std::array<fheroes2::ButtonBase *, 4> buttons{ &buttonStandardGame, &buttonCampaignGame, &buttonMultiplayerGame, &buttonCancel };
 
     if ( !isSuccessionWarsCampaignPresent() ) {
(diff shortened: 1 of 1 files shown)


src/fheroes2/game/game_loadgame.cpp
```diff
@@ -157,6 +157,10 @@ fheroes2::GameMode Game::LoadGame()
     buttons[2].setICNInfo( ICN::BUTTON_MULTIPLAYER_GAME, 0, 1 );
     buttons[3].setICNInfo( ICN::BUTTON_LARGE_CANCEL, 0, 1 );
 
+    if ( !isSuccessionWarsCampaignPresent() ) {
+        buttons[1].disable();
+    }
```
Districh-ru, line 162: The disabled button of this type has doubled shadow strength:
изображение

So we need to act like we do for main menu and use `fheroes2::ButtonSprite` for this button and use  `ICN::NEW_CAMPAIGN_DISABLED_BUTTON` as ICN for disabled button:
```cpp
fheroes2::ButtonSprite buttonCampaignGame( buttonPos.x, buttonPos.y + buttonYStep * 1, fheroes2::AGG::GetICN( ICN::BUTTON_CAMPAIGN_GAME, 0 ),
                                               fheroes2::AGG::GetICN( ICN::BUTTON_CAMPAIGN_GAME, 1 ), fheroes2::AGG::GetICN( ICN::NEW_CAMPAIGN_DISABLED_BUTTON, 0 ) );
```

But this ICN also needs to be improved (in a separate PR) as its text is not translated: :)
изображение

Districh-ru (changes requested):
Hi @ihhub, I left one comment. Could you take a look in your spare time.

src/fheroes2/game/game_loadgame.cpp
ihhub: Good point. I will address it.

PR #7418, opened 2023-07-16
Maps Tiles related code cleanup

ihhub:
To make things easier for the Editor changes.
- move 2 functions out of the Tiles::Maps class as they don't belong there
- function and variable renaming for easier code support

relates to #6845


src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -1190,7 +1190,7 @@ void AIWorldPathfinder::setArmyStrengthMultiplier( const double multiplier )
     }
 }
 
-void AIWorldPathfinder::setSpellPointReserve( const double reserve )
+void AIWorldPathfinder::setSpellPointReserveRatio( const double reserve )
 {
-    _spellPointsReserved = reserve;
+    _spellPointsReservedRatio = reserve;
 }
```
Districh-ru, line 1196: This method could be inlined, what do you think?

Districh-ru (changes requested):
Hi @ihhub, I left one suggestion. Could you check it when you have time.

src/fheroes2/world/world_pathfinding.cpp
oleg-derevenetz: In #7408 I made this variable `constexpr` because it does not change anywhere in reality.

Districh-ru: It changes value only in this line in this code:
https://github.com/ihhub/fheroes2/blob/44f0f6b9e06e397f131ade88755cd7c7a5924fba/src/fheroes2/ai/normal/ai_normal_hero.cpp#L2116
In all other places its value is set to 0.5.
In #7408 this line is removed so there it can be `constexpr`.

oleg-derevenetz: @Districh-ru Yes, in any case, this PR is still in some way "in flux".

oleg-derevenetz (approved):

src/fheroes2/world/world_pathfinding.cpp
ihhub: Hi @Districh-ru , I revert these changes not to conflict with #7408

PR #7430, opened 2023-07-19

oleg-derevenetz:
Follow-up to the latest PRs.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -714,14 +714,8 @@ namespace AI
 
     void Normal::removeEnemyArmies( const int32_t tileIndex )
     {
-        for ( auto iter = _enemyArmies.begin(); iter != _enemyArmies.end(); ) {
-            if ( iter->index == tileIndex ) {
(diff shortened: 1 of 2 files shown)


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -714,14 +714,8 @@ namespace AI
...
-        for ( auto iter = _enemyArmies.begin(); iter != _enemyArmies.end(); ) {
-            if ( iter->index == tileIndex ) {
-                iter = _enemyArmies.erase( iter );
-            }
-            else {
-                ++iter;
-            }
-        }
+        _enemyArmies.erase( std::remove_if( _enemyArmies.begin(), _enemyArmies.end(), [tileIndex]( const EnemyArmy & item ) { return item.index == tileIndex; } ),
+                            _enemyArmies.end() );
```
oleg-derevenetz, line 718: Please note that `std::remove_if()` is much more effective because vector elements are shifted exactly once regardless of number of elements removed, while with multiple `erase()` calls there may be multiple shifts of the same elements.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for cleaning the code!

PR #7432, opened 2023-07-19
Swedish translation update

Mr-Bajs:


ihhub:
@Mr-Bajs , thanks a lot for this update!

PR #7435, opened 2023-07-20
Really turn off the AIAutoControlMode only on the next turn

oleg-derevenetz:
Currently, when `AIAutoControlMode` is turned off during the turn of the AI-controlled human player, it turns off only "half" until the end of the turn (i.e. AI still makes decisions, but all actions are performed as if this player is already controlled by a human). This leads to various bugs, e.g. [this one](https://github.com/ihhub/fheroes2/pull/7408#issuecomment-1640717228). This PR fixes this by really fully turning off the `AIAutoControlMode` only on the next turn.

Also this PR facilitates `AIAutoControlMode` switching, e.g. when AI-controlled hero actively uses DD/TG/TP spells or Whirlpools.

Also all `AIAutoControlMode`-related code has been moved under the `WITH_DEBUG` macro.


src/fheroes2/gui/interface_focus.cpp
```diff
@@ -51,7 +51,15 @@ void Interface::AdventureMap::SetFocus( Heroes * hero, const bool retainScrollBa
         return;
     }
 
-    assert( player->GetColor() == hero->GetColor() && ( player->isControlHuman() || ( player->isControlAI() && player->isAIAutoControlMode() ) ) );
+#ifndef NDEBUG
+#if defined( WITH_DEBUG )
```
ihhub, line 55: Would it be better to combine these two macro into one
```cpp
#if !defined( NDEBUG ) && defined( WITH_DEBUG )
```
?

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -482,4 +486,57 @@ void Game::globalKeyDownEvent( const fheroes2::Key key, const int32_t modifier )
         conf.setTextSupportMode( !conf.isTextSupportModeEnabled() );
         conf.Save( Settings::configFileName );
     }
+#if defined( WITH_DEBUG )
+    else if ( key == hotKeyEventInfo[hotKeyEventToInt( HotKeyEvent::WORLD_TRANSFER_CONTROL_TO_AI )].key ) {
+        static bool recursiveCall = false;
+
+        if ( !recursiveCall ) {
+            class RecursionGuard
```
ihhub, line 494: I do not entirely understand the purpose of this class over normal code:
```cpp
recursiveCall = true;
... // do something
recursiveCall = false;
```
except RAII principle. Could you please explain it?

ihhub (changes requested):
Hi @oleg-derevenetz , I put few comments here. Would you mind to take a look at them?

src/fheroes2/game/game_hotkeys.cpp
oleg-derevenetz: @ihhub Except RAII, there is none :) I can change this if you want. I like RAII because it lets you not to care about additional `return`-like calls in future. It's very easy to forget to add another `recursiveCall = false` before such a `return`.

src/fheroes2/gui/interface_focus.cpp
ihhub: Yeah, now I get it. Frankly speaking this code doesn't look nice but I see no other way to make it better :)

src/fheroes2/game/game_hotkeys.cpp
ihhub: Let's do this way.

ihhub (approved):

ihhub:
Hi @drevoborod , would you mind to test these changes?

Districh-ru (approved):
Well done!
Now after switching back to human control AI correctly finishes the turn. And during this turn you can change your mind. :)
I let AI to play and win a map and found no issues.

drevoborod:
I agree - no issues. Testing completed :)

ihhub:

PR #7442, opened 2023-07-22

Ben80-hero:
Fix a game analytics bug in Monster::GetMonsterStrength.

Here in this line:

More correctly like this:
const double attackDefense = 1.0 + attack * 0.1 + defense * 0.1;

Rationale:

1) In practice, damage in battle is determined by the difference between attack parameter (of the 1st monster) and defense parameter (2nd monster), and not by some formula that would include parameters only one of the monsters. Therefore, in practice, as a result, one unit of defense parameter mutually
Example:
Consider a battle (exchange of blows) of 2 hypothetical monsters. The 1st monster has attack parameter is 15 and defense parameter is 5.
 2nd monster has attack is 5, and the defense is 15. In practice, these 2 monsters in the battle with each other are equal. However according to the formula above, for the 1st monster the function will return 2.75, and for the 2nd monster it will return 2.25.


src/fheroes2/monster/monster.cpp
```diff
@@ -179,7 +179,7 @@ double Monster::GetMonsterStrength( int attack, int defense ) const
     if ( defense == -1 )
         defense = GetDefense();
 
-    const double attackDefense = 1.0 + attack * 0.1 + defense * 0.05;
+    const double attackDefense = 1.0 + attack * 0.1 + defense * 0.1;
```
ihhub, line 182: Hi @Ben80-hero , please optimize the formula as we have 2 identical multipliers.

ihhub (changes requested):
Hi @Ben80-hero , I left one optimization comment and also how did you test the behavior of AI for this change? Has it become better? If yes for which cases?

src/fheroes2/monster/monster.cpp
Ben80-hero: I did not conduct testing, and I doubt that tests are possible or necessary in this case. That is, in this case, we simply refine one of the estimates on which the actions of the AI ​​are based. The more accurate the AI's estimates, the better its overall decisions.

I think that these changes will have the greatest effect on high-level creatures - their power was clearly underestimated.

oleg-derevenetz:
This formula (with exactly the same coefficients) was initially added by @idshibanov in #692 and it probably has some grounds. Let's ask him for a review.

src/fheroes2/monster/monster.cpp
Ben80-hero: Have optimized formula.

PR #7449, opened 2023-07-22
Disable transform layer for images where it is not needed (to speed-up image processing)

Districh-ru:
This PR is a part of #7233

This PR disables `transform()` layer for images that have no transparent (or shadow/bright transform) pixels:
- during ICN load;
- for battle arena fade effect (after the battle);
- for spell effects:
  - Death Wave;
  - Holy Shout;
  - Earthquake;
  - Armageddon;
- for Puzzle map.

Doing this will save some computation resources while processing these images: no operations with the transform layer would be performed.

This PR changes properties of many loaded images so to test it is good to check rendering of maps, in castles, battlefield, dialogs for possible transparency bugs (e.g. blacked image parts in place of transparency of sprites).
As the changes are the same for the same type of images the check may include viewing each type of render, e.g.:
- check the render of the castle of one race with and without buildings, the animation of building, check the well, tavern and other objects;
- check the battle with shooters, flyers, wide and non-wide creatures;
- check render of mentioned above spells;
- check the puzzle map, the View World;
- check load, new game, campaign info, high score dialogs.

### DIFF
--- a/src/engine/image_tool.cpp
+++ b/src/engine/image_tool.cpp
@@ -350,9 +350,7 @@ namespace fheroes2
                 ++data;
             }
             else if ( 0x80 > *data ) { // 0x01-0x7F - repeat a pixel N times
-                if ( noTransformLayer && ( static_cast<int32_t>( posX ) < width ) ) {
-                    noTransformLayer = false;
-                }
+                noTransformLayer = noTransformLayer && ( static_cast<int32_t>( posX ) >= width );
 
                 const uint8_t pixelCount = *data;
                 ++data;
@@ -369,9 +367,7 @@ namespace fheroes2
                 posX += pixelCount;
             }
             else if ( 0x80 == *data ) { // 0x80 - end of image
-                if ( noTransformLayer && ( static_cast<int32_t>( posX ) < width ) ) {
-                    noTransformLayer = false;
-                }
+                noTransformLayer = noTransformLayer && ( static_cast<int32_t>( posX ) >= width );
 
                 break;
             }
--- a/src/engine/image_tool.cpp
+++ b/src/engine/image_tool.cpp
@@ -339,6 +339,9 @@ namespace fheroes2
 
         const uint8_t * dataEnd = data + sizeData;
 
+        // The need for a transform layer can only be determined during ICN decoding.
+        bool noTransformLayer = true;
+
         while ( true ) {
             if ( 0 == *data ) { // 0x00 - end of row
                 imageData += width;
@@ -347,6 +350,10 @@ namespace fheroes2
                 ++data;
             }
             else if ( 0x80 > *data ) { // 0x01-0x7F - repeat a pixel N times
+                if ( noTransformLayer && ( static_cast<int32_t>( posX ) < width ) ) {
+                    noTransformLayer = false;
+                }
+
                 const uint8_t pixelCount = *data;
                 ++data;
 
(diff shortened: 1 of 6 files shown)


src/engine/image_tool.cpp
```diff
@@ -347,6 +350,10 @@ namespace fheroes2
                 ++data;
             }
             else if ( 0x80 > *data ) { // 0x01-0x7F - repeat a pixel N times
+                if ( noTransformLayer && ( static_cast<int32_t>( posX ) < width ) ) {
+                    noTransformLayer = false;
+                }
```
ihhub, line 355: I think this would be faster:
```cpp
noTransformLayer = noTransformLayer && ( static_cast<int32_t>( posX ) >= width );
```

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2828,10 +2845,11 @@ namespace fheroes2
                 if ( !_icnVsSprite[id].empty() ) {
                     Sprite & original = _icnVsSprite[id][0];
                     if ( original.width() == 101 && original.height() == 93 ) {
-                        replaceTransformPixel( original, 2314, 70 );
-                        replaceTransformPixel( original, 5160, 71 );
-                        replaceTransformPixel( original, 5827, 18 );
-                        replaceTransformPixel( original, 7474, 167 );
+                        original._disableTransformLayer();
+                        original.image()[2314] = 70;
```
ihhub, line 2849: If speed is a concern then we need to cache **original.image()** is a virtual function and cannot be inlined.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3907,16 +3932,33 @@ namespace fheroes2
...
+                }
+                return true;
+            }
             case ICN::GAME_OPTION_ICON: {
                 _icnVsSprite[id].resize( 2 );
 
                 h2d::readImage( "hotkeys_icon.image", _icnVsSprite[id][0] );
                 h2d::readImage( "graphics_icon.image", _icnVsSprite[id][1] );
+
+                for ( fheroes2::Sprite & sprite : _icnVsSprite[id] ) {
```
ihhub, line 3956: I suggest to leave a TODO comment to modify the resources instead of adding a hack here. h2d file should contain the correct resources in order to speed up loading and avoid unneccessary computations.

ihhub (changes requested):
Hi @Districh-ru , I left few comments here. Could you please take a look at them?

src/engine/image_tool.cpp
Districh-ru: Hi @ihhub, thanks for the suggestion!

src/fheroes2/agg/agg_image.cpp
Districh-ru: Sure!)

src/fheroes2/agg/agg_image.cpp
Districh-ru: I left a TODO in `h2d_file.cpp`

ihhub (approved):

ihhub:
Hi @Districh-ru , could you please update the description to specify the scope of testing? At least to verify that we don't break any rendering by these changes.

Districh-ru:
> Hi @Districh-ru , could you please update the description to specify the scope of testing? At least to verify that we don't break any rendering by these changes.

Hi, @ihhub, a short description is added.

Districh-ru:
Hi, @drevoborod, could you please test this PR when you have time.

drevoborod:
Testing completed, didn't find any issues. All mentioned cases were checked plus several extra.

ihhub:

PR #7455, opened 2023-07-22
Apply AI growth bonus only on first AI player turn

### LINKED ISSUE #6939: Add monster growth bonus for AI in castles at first AI player turn
### Preliminary checks

Some players abuse the AI unit growth bonus in castles by waiting and capturing AI castles on a first day of a week. We reduce this bonus in the past but it is still present.

In order to minimize the abuse from player's side we need to apply the bonus only at the first AI player being played per current turn. In single-player map a player who goes first in a turn won't get extra units in captured castles / towns. During multiplayer mode AI still goes after human players so AI would have a chance to use this bonus properly.

### Additional info

An alternative solution is to completely remove the bonus but as of now AI is not strong enough for this.

ihhub:
This pull request fixes multiple issues:
- apply weekly monster growth bonuses only when the first AI player starts its turn (close #6939). This eliminates a possibility to abuse this bonus by human players when they capture a castle on the first day of week.
- global events which were marked as not applicable for AI players were still used for them. For example, AI players should not have events deducting their resources in Great War II map as human players do.

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -708,6 +708,12 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
             Kingdom & kingdom = world.GetKingdom( playerColor );
 
             if ( skipTurns && !player->isColor( conf.CurrentColor() ) ) {
+                if ( kingdom.isPlay() && kingdom.GetControl() == CONTROL_AI ) {
+                    // Only the first AI player can trigger AI bonuses.
+                    // Since this player is the one the bonuses have been applied.
+                    isFirstAIPlayer = false;
+                }
+
                 continue;
             }
 
@@ -767,7 +773,7 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
                     // TODO: remove this temporary assertion
                     assert( res == fheroes2::GameMode::END_TURN );
 
-                    if ( isFirstAIPlayer && !loadedFromSave ) {
+                    if ( isFirstAIPlayer ) {
                         isFirstAIPlayer = false;
                         // All bonuses for AI must be applied on the first AI player turn, not the first player in general.
                         // This prevents human players to abuse AI bonuses.
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -698,6 +698,8 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
 
         res = fheroes2::GameMode::END_TURN;
 
+        bool isFirstAIPlayer = true;
+
         for ( const Player * player : sortedPlayers ) {
             assert( player != nullptr );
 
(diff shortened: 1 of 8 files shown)


src/fheroes2/game/game_startgame.cpp
```diff
@@ -765,6 +767,13 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
                     // TODO: remove this temporary assertion
                     assert( res == fheroes2::GameMode::END_TURN );
 
+                    if ( isFirstAIPlayer && !loadedFromSave ) {
+                        isFirstAIPlayer = false;
+                        // All bonuses for AI must be applied on the first AI player turn, not the first player in general.
+                        // This prevents human players to abuse AI bonuses.
+                        world.NewDayAI();
+                    }
```
oleg-derevenetz, line 775: Let's assume that there are two players: human-controlled player and AI-controlled player. Suppose that human player performs the save on day 1/1/1 and then loads this save before the AI player will get the chance to start its turn. Will AI bonuses be applied in this case? I suggest to integrate this logic into the `World::NewWeek()` instead (just call the corresponding methods strictly on day 1). What do you think?

oleg-derevenetz (commented):
Hi @ihhub I just left one question (with the related suggestion), could you please take a look when you have time?

src/fheroes2/game/game_startgame.cpp
ihhub: Hi @oleg-derevenetz , we cannot put the new logic inside ActionNewWeek() as this is called when the first player in a turn starts while the aim of this change is to apply bonuses when first AI player starts its turn.

However, you are right about this place and this is a bug. I will fix it.

oleg-derevenetz (approved):

ihhub:
Hi @drevoborod , could you please help to test this change?

drevoborod:
How am I supposed to know what bonuses has the AI got? When I provide control to the AI player with F8, it doesn't count as if it's real AI player, right?

ihhub:
Hi @drevoborod , what I did for testing: I put my hero in front of a castle on a first day of week. I captured the cate and checked how many Centaurs were there for hire. There should be 10. The army defending the castle was 24 centaurs. Then I loaded a save on the same day before the capture but didn't attack the castle and waited one more day to do it. The number of centaurs in this case jumped to 36 that means that the AI got 12 centaurs to hire. It was a game on Expert difficulty where the growth bonus is 25.4%. If you play on Impossible difficulty the growth bonus for the AI must be 36.8%.

Yes, there is no way to verify the growth directly, even by passing the control to the AI. However, knowing what army in a castle before and after (assuming that the AI has enough money to hire them) serves as a verification.

Once I'm back home I can attach a save file for the aforementioned case of mine.

ihhub:
@drevoborod , another more accurate test would be to give the AI a fully built castle and create daily global event to remove all resources. Capture a castle on the first and on second days and check how many troops for hire.

drevoborod:
Checked growth bonus on all difficulty levels. Here are saved games for the levels where are differencies between release and fixed versions: 
[Test_Growth_.zip](https://github.com/ihhub/fheroes2/files/12144323/Test_Growth_.zip)
Everything seems to be OK:
- [x] Human player does not get bonuses.
- [x] AI player gets bonuses.

ihhub:
Thank you @drevoborod !

PR #7457, opened 2023-07-23
Fix the misplaced check for the need of transform layer while loading ICN images

Districh-ru:
A quick fix for #7449

This PR puts the check for need of transform layer into the right place.
In previous PR it was placed to the part where non-transparent images were loaded. So the transform layer was considered as needed for all images.
There is no noticeable bug in #7449, but it does not work as intended without this fix.

### DIFF
--- a/src/engine/image_tool.cpp
+++ b/src/engine/image_tool.cpp
@@ -344,6 +344,8 @@ namespace fheroes2
 
         while ( true ) {
             if ( 0 == *data ) { // 0x00 - end of row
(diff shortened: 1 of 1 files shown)


Districh-ru:
Hi @drevoborod, could you please test this PR (once more). Just play the game for a while and check that there is no render bugs (absence of transparency where needed).

drevoborod:
Testing completed, no issues found.

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for this quick fix!

PR #7458, opened 2023-07-23
Make std::scoped_locks const

oleg-derevenetz:


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for cleaning up the technical debt!

PR #7466, opened 2023-07-23
The Shield spell should not affect the damage of the castle towers

oleg-derevenetz:
See https://handbookhmm.ru/2-spells for details.

Also renamed variables and functions with names correlated to "blind answer" to "blind retaliation" and added a few comments.

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -365,7 +365,7 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
         const bool doubleAttack = attacker->isDoubleAttack();
 
         if ( attacker->isArchers() || handfighting ) {
-            defender->SetBlindAnswer( defender->Modes( SP_BLIND ) );
+            defender->SetBlindRetaliation( defender->Modes( SP_BLIND ) );
 
             BattleProcess( *attacker, *defender, dst, dir );
 
@@ -375,13 +375,13 @@ void Battle::Arena::ApplyActionAttack( Command & cmd )
                     defender->SetResponse();
                 }
 
-                defender->SetBlindAnswer( false );
-
                 if ( doubleAttack && attacker->isValid() && !attacker->Modes( SP_BLIND | IS_PARALYZE_MAGIC ) ) {
                     DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "double attack" )
                     BattleProcess( *attacker, *defender, dst, dir );
                 }
             }
+
+            defender->SetBlindRetaliation( false );
 
             // Berserk and Hypnotize spells should only be removed after an attack initiated by this unit, but not after its retaliatory attack
             attacker->removeAffection( SP_BERSERKER | SP_HYPNOTIZE );
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -508,7 +508,7 @@ uint32_t Battle::Unit::CalculateDamageUnit( const Unit & enemy, double dmg ) con
 {
     if ( isArchers() ) {
         if ( !isHandFighting() ) {
-            // check skill archery +%10, +%25, +%50
+            // Hero's Archery skill may increase damage
             if ( GetCommander() ) {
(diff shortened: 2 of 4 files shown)


agiat86:
That website says it shouldn't work against towers. But is that the case really?

Alucard648:
I think, Shield`s protection vs towers should work in Ressurection expansion.

oleg-derevenetz:
> That website says it shouldn't work against towers. But is that the case really?

If you don't believe it, then you can check for yourself.

Ben80-hero:
Yes, this is correct. The Shield spell should not affect the damage of the castle towers. This spell is already quite effective (even if not affect the damage of the castle towers). In addition, there are no signs indicating that the original game has a bug in this respect.

Branikolog:
I personally expect, that this spell should affect towers, since their characteristics are correlated with archers. 
But replicating of the original operation for such core spell behaviour seems like more proper way to go.

Branikolog (approved):
I'm no a pro in creating descriptions. Placing exceptions in the description looks unnatural for me. Maybe we should just point, that the spell halves damage from ranged enemy troops (ballista and turrets are not troops, right? 😄)
Everything else works as intended.

oleg-derevenetz:
> spell halves damage from ranged enemy troops

Also not good because ranged enemy units may as well fight in melee, and in this case Shield will not work.

Ben80-hero:
> the spell halves damage from ranged enemy troops

the spell halves distance attack damage from ranged enemy troops
??

Branikolog:
@oleg-derevenetz 
>Also not good because ranged enemy units may as well fight in melee, and in this case Shield will not work.

Absolutely true.

Maybe something similar to proposed by @Ben80-hero ?
"the spell halves ranged attack damage from enemy troops"
How do you think?

oleg-derevenetz:
I suggest leaving the current wording of the description of spells, because it is quite clear, and consider other ways to improve it in a separate PR.

ihhub (approved):

ihhub:

LeHerosInconnu:
Hello everyone,

Too bad that this has been reverted back to work as in the original game.
In the original game, the mass shield spell is used very rarely, mainly when the hero doesn't know the shield spell, because the most dangerous ranged troop is always targeted first by AI controlled shooters in the original game.

In fheroes2, the AI no longer systematically targets the most dangerous ranged troop.
The mass shield spell is a little more usable in fheroes2.
But where it really became a spell to consider in fheroes2 was during a siege, when ballista and turrets can inflict substantial damage, at least when the spell also protected against ballista and turrets.
Now that we've returned to the way things worked in the original game, the shield and mass shield spells are back to being spells to be considered only when the hero doesn't know that much other spells.

For me, it would be an improvement over the original game if the shield and mass shield spells reduced ballista and turrets damage during a siege.
Actually, I had planned this for my mod even before testing how it worked in fheroes2: https://github.com/ihhub/fheroes2/discussions/2966. :)
And also a change for the archery secondary skill: https://github.com/ihhub/fheroes2/discussions/2968.

oleg-derevenetz:
@LeHerosInconnu so you recommend to revert this change? Let's vote then. @LeHerosInconnu is voting for reverting. @ihhub @Branikolog @zenseii @Districh-ru?

Districh-ru:
I think we should act like in the original game (do not revert this PR).
We may plan to add a new shield spell in Resurrection expansion that will also work against ballista and turrets: they are more powerful then bows and can break through the original Shield spell.

ihhub:
I suggest to keep the logic as it is.

oleg-derevenetz:
OK, so 1 vote for reverting this change and 4 votes against this (including my vote). Bets are made, there are no more bets.

Branikolog:
I personally liked the previous logic, as towers are shooting with simple arrows and it looks quite natural, that "shield" spell reduces the damage. Nevertheless I respect the original logic. So I'm OK with any decision we come to.

crypt47:
> I personally liked the previous logic, as towers are shooting with simple arrows and it looks quite natural, that "shield" spell reduces the damage.

yeh, me too. Visited the "bug" only to say that.

PR #7472, opened 2023-07-25
Restoring original Summon Boat spell

Ben80-hero:
In OG, summoning a boat by a hero is done as follows:
first, the boat belonging to the given hero is searched (each boat
belonging to a certain player, but frequently also belonging to certain hero),
If such boat is not found, then a free boat belonging to this player is searched for.
A free boat on the map that does not belong to anyone cannot be summoned.
Summoning a boat in fheroes2 works in a completely different way. BTW, this
may break a number of maps.
It should be noted that summoning a boat in OG is also quite convenient - usually it is most profitable to summon the boat that this particular hero used.
(not another boat, which another hero might need soon).
This PR restores the original algorithm. The only thing which I did not realized is adaptation for Save / Load (I ask for help).

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1499,6 +1499,7 @@ namespace
         AI::Get().HeroesClearTask( hero );
 
         world.GetTiles( dst_index ).resetBoatOwnerColor();
+        world.GetTiles( dst_index ).resetBoatOwnerHeroID();
 
         DEBUG_LOG( DBG_AI, DBG_INFO, hero.GetName() )
     }
@@ -1517,7 +1518,7 @@ namespace
 
         hero.ResetMovePoints();
         hero.Move2Dest( dst_index );
-        from.setBoat( Maps::GetDirection( fromIndex, dst_index ), hero.GetColor() );
+        from.setBoat( Maps::GetDirection( fromIndex, dst_index ), hero.GetColor(), hero.GetID() );
         hero.SetShipMaster( false );
         hero.GetPath().Reset();
 
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -2539,15 +2539,15 @@ bool Castle::BuyBoat() const
 
     if ( MP2::OBJ_NONE == left.GetObject() && left.isWater() ) {
         kingdom.OddFundsResource( PaymentConditions::BuyBoat() );
-        left.setBoat( Direction::RIGHT, kingdom.GetColor() );
+        left.setBoat( Direction::RIGHT, kingdom.GetColor(), -1 );
     }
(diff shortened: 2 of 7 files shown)


Ben80-hero:
I also ask to help correct inaccuracies and code style so that I don’t rack my brains about what’s wrong (and next time I’ll do it right).

oleg-derevenetz:
> In OG, summoning a boat by a hero is done as follows:
first, the boat belonging to the given hero is searched (each boat
belonging to a certain player, but frequently also belonging to certain hero),
If such boat is not found, then a free boat belonging to this player is searched for.
A free boat on the map that does not belong to anyone cannot be summoned.

Where did you get all this nonsense? This video from the original game demonstrates that your understanding is just plain wrong:

https://github.com/ihhub/fheroes2/assets/32623900/30ed4220-7a10-425a-acb5-b8a92eeb9c6c

In this video:

1. Neutral boat is summoned just fine;
2. The nearest boat (the neutral one) is summoned, not the boat "belonging to this hero" (or even to this player).

In the future, please, double-check the facts and don't waste your time (your own and the rest of the project participants) on "corrections" that are not correct (pun intended).

Ben80-hero:
> Where did you get all this nonsense? This video from the original game demonstrates that your understanding is just plain wrong:
> 

This follows from an analysis of the OG code. I'll post the decompiled listing later. Why this does not work in practice is not entirely clear.

oleg-derevenetz:
> This follows from an analysis of the OG code. I'll post the decompiled listing later.

Please don't do this. This may be illegal in certain jurisdictions, so we can't use decompiled code, just an observable behavior (and it's not what you say it is).

Ben80-hero:
> This may be illegal in certain jurisdictions, so we can't use decompiled code, just an observable behavior (and it's not what you say it is).

I'm not going to show anything publicly. However, this engine would not have been made if there were no reverse data used at all. I suspect that yet you will have to use some of such data (even if you think otherwise).

oleg-derevenetz:
> I'm not going to show anything publicly.

This may be still illegal. @ihhub what do you think?

> However, this engine would not have been made if there were no reverse data used at all.

At least, this team is not engaged in reverse engineering of the original source code. Only observable behavior from the original game is used.

Ben80-hero:
> We mustn't share nor use any source code obtained from reverse engineering. Please refer to EULA documents.

But you will not deny that we discussed the decompiled code together ? I don't specify which game.

Ben80-hero:
> > I'm not going to show anything publicly.
> 
> This may be still illegal. @ihhub what do you think?
> 
> > However, this engine would not have been made if there were no reverse data used at all.
> 
> At least, this team is not engaged in reverse engineering of the original source code. Only observable behavior from the original game is used.

How do you think, where did the first author of fheroes (Afletdinov) get the basic structures, for example, the structures of the cells of the map? If your answer is "from observations", then you are unlikely to be right.

ihhub:
I'm closing this pull request and every subsequent messaging as the conversation has gone beyond the original topic.

PR #7476, opened 2023-07-25

Ben80-hero:
I would like to fix a game analytics in Monster::GetMonsterStrength.

Here in this line:

More correctly like this:
const double attackDefense = 1.0 + 0.1 * ( attack + defense );

Rationale:

1) In practice, damage in battle is determined by the difference between attack parameter (of the 1st monster) and defense parameter (2nd monster), and not by some formula that would include parameters only one of the monsters. Therefore, in practice, as a result, one unit of defense parameter mutually
Example:
Consider a battle (exchange of blows) of 2 hypothetical monsters. The 1st monster has attack parameter is 15 and defense parameter is 5.  2nd monster has attack is 5, and the defense is 15. In practice, these 2 monsters in the battle with each other are equal, and according to the formula above, for the 1st monster the function will return 2.75, and for the 2nd monster it will return 2.25.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1499,7 +1499,6 @@ namespace
         AI::Get().HeroesClearTask( hero );
 
         world.GetTiles( dst_index ).resetBoatOwnerColor();
-        world.GetTiles( dst_index ).resetBoatOwnerHeroID();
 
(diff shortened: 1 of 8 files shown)


oleg-derevenetz:
Hi @Ben80-hero please don't make pull requests from your `master` branch. In this case you will be able to make just one pull request at a time, and properly synchronizing it with upstream changes becomes problematic. Create a separate "feature" branch for each pull request, it's much more convenient.

PR #7477, opened 2023-07-25

Ben80-hero:
I would like to fix a game analytics in Monster::GetMonsterStrength.

Here in this line:

More correctly like this:
const double attackDefense = 1.0 + 0.1 * ( attack + defense );

Rationale:

1) In practice, damage in battle is determined by the difference between attack parameter (of the 1st monster)  and defense parameter (2nd monster), and not by some formula that would include parameters only one of the monsters. Therefore, in practice, as a result, one unit of defense parameter mutually
Example:
Consider a battle (exchange of blows) of 2 hypothetical monsters. The 1st monster has attack parameter is 15 and defense parameter is 5.  2nd monster has attack is 5, and the defense is 15. In practice, these 2 monsters in the battle with each other are equal, and according to the formula above, for the 1st monster the function will return 2.75, and for the 2nd monster it will return 2.25.

### DIFF
--- a/src/fheroes2/monster/monster.cpp
+++ b/src/fheroes2/monster/monster.cpp
@@ -179,7 +179,7 @@ double Monster::GetMonsterStrength( int attack, int defense ) const
     if ( defense == -1 )
         defense = GetDefense();
 
-    const double attackDefense = 1.0 + attack * 0.1 + defense * 0.05;
+    const double attackDefense = 1.0 + 0.1 * ( attack + defense );
     return attackDefense * fheroes2::getMonsterData( id ).battleStats.monsterBaseStrength;
 }
 


ihhub:
Hi @Ben80-hero , as I requested in the previous PR could you please provide some results why this formula is better than the old? The formula used in the original game is not always suitable for fheroes2 as fheroes2's AI is absolutely different and more sophisticated.

Please find situations of overestimation or underestimation of enemy's army strength by AI heroes leading them to attack or not attack. And if this change indeed better then the behavior of the AI heroes would improve.

Ben80-hero:
Hi @ihhub ,

I've run tests that confirm that the proposed change improves the AI's assessment of the effects of combat.

Test1:
AI army is 25 ogres, AI hero (RED) skills are starting for Knight (Attack 2, Defense 2).
My army is 20 ogres, my hero (BLUE) skills are starting for Knight (Attack 2, Defense 2) + Artifact,
which gives Defense +12.
The AI thinks he is stronger, attacks and loses.


https://github.com/ihhub/fheroes2/assets/139842209/042d32ad-0e37-4ee9-bf0a-e9f9ea008ab1

Ben80-hero:
Test2:
AI army is 20 ogres, AI hero (RED) skills are initial for Knight (Attack 2, Defense 2) + Artifact,
which gives Defense +12.
My army is 25 ogres, my hero's (BLUE) skills are starting for Knight (Attack 2, Defense 2).
The AI thinks he is weaker and doesn't attack. And he could attack and win. I show it by attacking him and losing.



https://github.com/ihhub/fheroes2/assets/139842209/f6c6d224-8777-4c54-a79c-6d242a8464c1

Ben80-hero:
Test3:
The AI army is 20 ogres, 10 trolls, 3 cyclops, 25 orc chieftains, 125 goblins.
AI Hero (RED) skills - initial for the Knight (Attack 2, Defense 2).
My army is 16 ogres, 8 trolls, 2 cyclops, 20 orc chieftains, 100 goblins.
My Hero's (BLUE) skills - initial for Knight (Attack 2, Defense 2) + Artifact,
which gives Defense +12.
The AI thinks it's stronger, attacks and loses.


https://github.com/ihhub/fheroes2/assets/139842209/6bcdf1be-7c1f-4f47-a4be-83587c882ab1

Ben80-hero:
**After proposed correction:**
AI army is 25 ogres, AI hero (RED) skills are starting for Knight (Attack 2, Defense 2).
My army is 20 ogres, my hero (BLUE) skills are starting for Knight (Attack 2, Defense 2) + Artifact,
which gives Defense +12.
AI understands that he is weaker and does not attack. Then I catch up with him and win.


https://github.com/ihhub/fheroes2/assets/139842209/d433900d-18ee-4019-9877-9fdbb7ab5fda

Ben80-hero:
Hi @ihhub ,

Should the analysis be extended by trying to classify different cases ? By the way, today I will re-upload the video so that the 1st frame looks better.

PR #7486, opened 2023-07-28
Fix build on PS Vita

oleg-derevenetz:


ihhub (approved):

ihhub:

Issue #7448, opened 2023-07-22
Adventure map, allow instant exit from the puzzle map window


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Player has to wait until the fade animation finishes to close this window. Even if there're no revealed puzzle pieces yet.

https://github.com/ihhub/fheroes2/assets/55348946/0f089a27-a02a-4f49-880d-eb7f8644bc19

### Describe the possible solution

I propose to allow pressing "Exit" button before fade animation finishes. Players can sometimes enter this window accidentally. Or check, whether any pieces are revealed. This waiting for a second is a bit annoying.

PR #7487, opened 2023-07-28
Allow early exit from puzzle map during reveal animation

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/7448 by allowing exiting puzzle map during reveal animation.

### DIFF
--- a/src/fheroes2/kingdom/puzzle.cpp
+++ b/src/fheroes2/kingdom/puzzle.cpp
@@ -125,6 +125,7 @@ namespace
             if ( le.MouseClickLeft( buttonExit.area() ) || Game::HotKeyCloseWindow() ) {
                 return true;
             }
+
             if ( Game::validateAnimationDelay( Game::PUZZLE_FADE_DELAY ) ) {
                 fheroes2::Blit( sf, display, dstx, dsty );
 
@@ -152,6 +153,7 @@ namespace
                 assert( alpha >= 0 );
             }
         }
+
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


Kazhuu:
It seems Vita build is broken at the moment?

oleg-derevenetz:
> It seems Vita build is broken at the moment?

Please see #7486.

src/fheroes2/kingdom/puzzle.cpp
```diff
@@ -103,7 +103,8 @@ namespace
         }
     }
 
-    void revealPuzzle( const Puzzle & pzl, const fheroes2::Image & sf, int32_t dstx, int32_t dsty, const std::function<fheroes2::Rect()> * drawControlPanel = nullptr )
+    void revealPuzzle( const Puzzle & pzl, const fheroes2::Image & sf, int32_t dstx, int32_t dsty, fheroes2::Button & buttonExit,
```
ihhub, line 106: Would it be better to just return true or false from this function and then check the return value from the place where we call the function? Passing information from a caller of a function to the function must be the last resort.

ihhub (changes requested):
Hi @Kazhuu , I left a suggestion regarding the proposed approach. Would you mind to take a look?

Kazhuu:
> Hi @Kazhuu , I left a suggestion regarding the proposed approach. Would you mind to take a look?

Okay I returned the first implementation with boolean return variable. Can you @ihhub take a look?

src/fheroes2/kingdom/puzzle.cpp
```diff
@@ -146,6 +152,7 @@ namespace
                 assert( alpha >= 0 );
             }
         }
+        return false;
```
ihhub, line 155: I recommend to add an empty line before this return to have a visual separation of the code parts.

Kazhuu:
Hello @ihhub I was thinking because this version also need to pass exit button as a parameter, like you mentioned is not good. But do you think actually returning boolean value adds more code on the caller side and makes the code more complicated? Would you reconsider the implementation where the `revealPuzzle` function takes care of the exit button handling. Then I can remove those two while loops from the caller side. What do you think?

Kazhuu:
Hello @ihhub I made the changes you requested. Can you have a look again, thanks!

ihhub (approved):

ihhub:
@Kazhuu , big thanks for this change!

Issue #6888, opened 2023-03-23
Map selection menu, memorizing map size filter


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In the original game once player selected map size, it will be applied the next time he opens map list, even after loading and restarting the game.
fheroes2:

https://user-images.githubusercontent.com/55348946/227226159-3ecfc213-32aa-4248-b325-93d64d3682f8.mp4

The selection is preserved until player restarts the application.

### Save file

None.

ihhub:
Hi @Branikolog , could you please expand the description of this issue? I believe that we shouldn't do this for all cases. For instance if a player chooses a standard map and then a campaign the selection wouldn't be preserved. At least state all possible cases and when the selection should be memorized. Also it is good to note that this is a runtime feature and we shouldn't remember it in the configuration file.

I'm assigning this issue for your update. Once you're done unassign yourself.

Branikolog:
Hi, @ihhub .
Actually, in the original game the state of map filtering is always saved, until you restart the application. No matter you start campaign or multiplayer game, win a scenario or select expansion/original maps, map size filtering is always preserved.

PR #7490, opened 2023-07-29
Remember map filter selection until you quit the game

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/6888 by storing current map filter selection filter and applying it again when selection a scenario. Remember until you quit the game.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectscenario.cpp
+++ b/src/fheroes2/dialog/dialog_selectscenario.cpp
@@ -85,6 +85,8 @@ namespace
         MAP_SIZE_BUTTON_OFFSET_Y = 23
     };
 
+    Maps::mapsize_t currentMapFilter = Maps::ZERO;
+
     void ShowToolTip( const std::string & header, const std::string & body )
     {
         fheroes2::showStandardTextMessage( header, body, Dialog::ZERO );
@@ -198,12 +200,6 @@ namespace
     {
         const int32_t centerTransform = areaWidth > textWidth ? ( areaWidth - textWidth ) / 2 : 0;
         return startCoordX + centerTransform;
-    }
-
-    fheroes2::ButtonBase * SwitchPressedMapSizeButtons( fheroes2::Button & newlyPressedButton )
-    {
-        newlyPressedButton.press();
-        return &newlyPressedButton;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -502,6 +526,7 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps
                 currentPressedButton->drawOnPress();
             }
             else {
+                settings.setCurrentMapFilter( Maps::SMALL );
```
ihhub, line 529: Since this is only UI related issue please keep the change only within UI space. Settings class are designed to store permanent information. As a solution just declare a separate variable within this file in anonymous namespace and use it as storage to memorize the selection.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -444,9 +444,33 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps
...
+    case Maps::XLARGE:
+        currentPressedButton = &buttonSelectXLarge;
+        listbox.SelectMapSize( xlarge, Maps::XLARGE );
+        break;
+    default:
+        currentPressedButton = &buttonSelectAll;
+        listbox.SelectMapSize( all, Maps::ZERO );
+        break;
+    }
+    currentPressedButton->press();
```
ihhub, line 473: We don't need **currentPressedButton** variable here. We can directly call **press();** method for each case.

ihhub (changes requested):
Hi @Kazhuu , I put 2 comments here. Could you please check them?

src/fheroes2/dialog/dialog_selectscenario.cpp
Kazhuu: Not sure did you mean that you call `press()` in each case or I remove the whole variable. I did the latter now.

src/fheroes2/dialog/dialog_selectscenario.cpp
Kazhuu: Yeah good idea :+1:

Kazhuu:
> Hi @Kazhuu , I put 2 comments here. Could you please check them?

I did the fixes and also I found a crash which I also fixed. When you select M size map for example in single player. Then go to multiplayer with 6 players, the game crashed. Now if the map list is empty, then default to all filter instead.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -444,9 +440,52 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps
...
+            buttonSelectXLarge.press();
+            listbox.SelectMapSize( xlarge, Maps::XLARGE );
+        }
+        break;
+    default:
+        buttonSelectAll.press();
+        listbox.SelectMapSize( all, Maps::ZERO );
+        break;
+    }
     listbox.SetCurrent( GetInitialMapId( all ) );
```
ihhub, line 489: I suggest to add an empty line here for slightly better readability.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -444,9 +440,52 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps
...
     listbox.SetAreaItems( { rt.x + 55, rt.y + 55, 270, 175 } );
 
-    fheroes2::ButtonBase * currentPressedButton = &buttonSelectAll;
-    buttonSelectAll.press();
-    listbox.SelectMapSize( all, Maps::ZERO );
+    switch ( currentMapFilter ) {
+    case Maps::SMALL:
+        if ( small.empty() ) {
+            buttonSelectAll.press();
+            listbox.SelectMapSize( all, Maps::ZERO );
```
ihhub, line 447: Should we change **currentMapFilter** value to `Maps::ZERO` here as well as in other places?

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -499,11 +538,12 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps
         if ( le.MouseClickLeft( buttonSelectSmall.area() ) || HotKeyPressEvent( Game::HotKeyEvent::MAIN_MENU_MAP_SIZE_SMALL ) /*&& buttonSelectSmall.isEnabled()*/ ) {
             if ( small.empty() ) {
                 fheroes2::showStandardTextMessage( "", _( "No maps exist at that size" ), Dialog::OK );
-                currentPressedButton->drawOnPress();
+                buttonSelectSmall.drawOnPress();
```
ihhub, line 541: Please verify this situation:
1. You have no S size maps
2. You select XL maps
3. You press S size button

I suspect that S size button would be in a pressed state which is wrong. If this is the case please revert changes for **SwitchPressedMapSizeButtons** usage which was actually introduced in order to remember the last successful selection.

ihhub (changes requested):
Hi @Kazhuu , I put few more comments here. Could you please take a look at them when you have time?

src/fheroes2/dialog/dialog_selectscenario.cpp
Kazhuu: Yes that was indeed the case. I reverted this.

Kazhuu:
Hello @ihhub! I did the requested changes. Can you check again?

ihhub (approved):

ihhub:

Issue #7071, opened 2023-04-27
Add a confirmation dialog for Auto battle mode


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Even when playing on a PC there is a chance when a player can misclick on Auto button rather than the settings battle due to a bad UX design.

### Describe the possible solution

Since we haven't figured out what to do with button layout the safest temporary way is ti add a confirmation dialog asking to confirm auto battle mode.

PR #7492, opened 2023-07-30
Add dialog to confirm auto combat

Kazhuu:
Solves https://github.com/ihhub/fheroes2/issues/7071.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -3112,6 +3112,12 @@ void Battle::Interface::EventAutoSwitch( const Unit & unit, Actions & actions )
         return;
     }
 
+    if ( fheroes2::showMessage( fheroes2::Text( "", {} ), fheroes2::Text( _( "Are you sure you want to enable auto combat?" ), fheroes2::FontType::normalWhite() ),
+                                Dialog::YES | Dialog::NO )
+         != Dialog::YES ) {
+        return;
+    }
+
     actions.emplace_back( CommandType::MSG_BATTLE_AUTO_SWITCH, unit.GetCurrentOrArmyColor() );
 
     humanturn_redraw = true;


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3112,6 +3112,12 @@ void Battle::Interface::EventAutoSwitch( const Unit & unit, Actions & actions )
         return;
     }
 
+    if ( fheroes2::showMessage( fheroes2::Text( "", {} ), fheroes2::Text( _( "Are you sure you want to enable auto combat?" ), fheroes2::FontType::normalWhite() ),
+                                Dialog::YES | Dialog::NO )
+         != Dialog::YES ) {
+        return;
+    }
+
```
oleg-derevenetz, line 3120: Hi @Kazhuu this event can be raised both from the UI button and the hotkey (by default it's 'A'). I believe there is no need to show this dialog for the hotkey.

Kazhuu: I totally forgot about the hotkey. I guess for consistency I can also add this for the hotkey I think.

Kazhuu: Done. Can you take look again?

oleg-derevenetz: Hi @Kazhuu Is there any particular reason to add the function parameter instead of just showing this dialog in the `Battle::Interface::ButtonAutoAction()`?

Kazhuu: Yeah true, silly me. I moved the dialog now.

ihhub (approved):

ihhub:
Hi @zenseii , could you please review the proposed text for this feature?

ihhub:
@Kazhuu , thank you so much for this change!

PR #7500, opened 2023-08-02
Fix estimation of the castle garrison strength for priority tasks

### LINKED ISSUE #7488: AI hero sits in a castle forever despite having the stongest army
### Preliminary checks

### Platform

Windows

### Describe the bug

The green hero sits in the castle for many turns. I noticed that sometimes heros with strong armies enter "defend castle" mode and that makes them absolutely passive. That is something that didn't exist in the previous versions. I think 'defend castle' regime is not well balanced. In this case the paralyzed green hero can wipe out (I bet) all the enemies in 2 weeks, instead of sitting in the castle.

### Save file

[Uploading vh1_0016.savc…]()
[save.zip](https://github.com/ihhub/fheroes2/files/12202589/save.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #7488

Also a few improvements:

1. More correct and optimal enemy army strength estimation - if the hero is in the castle, do not count the hero's army and the castle's army separately, but instead count their combined strength. This may be somewhat rough, but IMHO is closer to the truth, because the castle cannot attack by itself, but the hero can take troops from the castle garrison. It also allows to optimize the code a bit (use map with relatively fast access instead of vector), because now we have just one instance of `EnemyArmy` for a particular tile.
2. A more correct assessment of a castle under threat. In this case, we should evaluate it not from the point of view of momentary benefits for the hero (for example, the presence of reinforcements), but from the point of view of the castle, which then will have to be taken back from the enemy, that is, according to approximately the same logic as enemy castles. Otherwise, problems arise, and even distant enemy castles in certain cases look much more attractive than the most well-built own castle.


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -604,7 +655,7 @@ namespace AI
             --daysToReach;
         }
 
-        const double defenders = castle.GetArmy().GetStrength();
+        const double defenders = castle.GetGarrisonStrength( enemyArmy.hero );
```
oleg-derevenetz, line 658: This was actually the main reason for the behavior when even a very powerful hero is constantly sitting in a castle, in combination with this logic:

https://github.com/ihhub/fheroes2/blob/b3d3ad14dd1e3272050c81c45b33981a660f2031/src/fheroes2/ai/normal/ai_normal_kingdom.cpp#L899-L908

In fact, the guest hero's army was not taken into account at all when assessing the security of the castle.

Although the current behavior may not be ideal either - for example, the hero may run out of the castle in certain circumstances and then return - it is quite difficult to come up with the "right" behavior in circumstances when there is a strong enough opponent near your castle that you do not dare to attack.

src/fheroes2/castle/castle.h
```diff
@@ -198,10 +198,11 @@ class Castle : public MapPosition, public BitModes, public ColorBase, public Con
...
-    // Returns the garrison strength estimation calculated as if the attacking hero really attacked this
-    // castle - including an estimate of the strength of the combined army consisting of the garrison and
-    // the hero's troops (if present), castle-specific bonuses from moat, towers and so on, relative to
-    // the attacking hero's abilities. See the implementation for details.
+    // Returns the garrison strength estimation calculated as if this castle had really been attacked, including
+    // an estimate of the strength of the combined army consisting of the garrison and the guest hero's troops
+    // (if present), castle-specific bonuses from moat, towers and so on, relative to the attacking hero's abilities
+    // (if the 'attackingHero' is not 'nullptr'). See the implementation for details.
+    double GetGarrisonStrength( const Heroes * attackingHero ) const;
     double GetGarrisonStrength( const Heroes & attackingHero ) const;
```
ihhub, line 206: Is there a reason why can't we just use the existing method by getting a pointer? Before the changes the method was used just in 3 places.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1981,7 +2025,7 @@ namespace AI
                 const Heroes * anotherHero = tile.GetHeroes();
                 if ( anotherHero == nullptr ) {
                     // The hero died.
-                    removeEnemyArmies( tileIndex );
+                    _enemyArmies.erase( tileIndex );
```
ihhub, line 2028: There could be a situation when an AI hero defends a castle and the attacker and the defender die (Armageddon spell, for example). In this case the hero dies but the castle still hasn't been captured.

ihhub (changes requested):
Hi @oleg-derevenetz , I put 2 comments here. Could you please check them in your spare time?

src/fheroes2/castle/castle.h
oleg-derevenetz: Hi @ihhub no, there is no other reason other than not having to add an & everywhere. But I just made the intermediate function inline :)

src/fheroes2/ai/normal/ai_normal_hero.cpp
oleg-derevenetz: @ihhub Yes, that's a good point. BTW we need to handle a castle here regardless of the battle outcome.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2013,11 +2068,12 @@ namespace AI
         switch ( task.type ) {
         case PriorityTaskType::DEFEND:
         case PriorityTaskType::REINFORCE: {
-            if ( objectType == MP2::OBJ_HEROES ) {
+            if ( hero.GetIndex() != tileIndex ) {
```
oleg-derevenetz, line 2071: This change should fix the following invalid behavior:

https://github.com/ihhub/fheroes2/assets/32623900/957df321-6c46-49fb-be1c-1836f54cb268

After this change:

https://github.com/ihhub/fheroes2/assets/32623900/786b7fb7-8ef0-4834-a046-aacc0af90328

kitovyj:
It seems that in the first attached game of https://github.com/ihhub/fheroes2/issues/7488 the green hero demonstrates another wierd behaviour now. Instead of sitting in the castle still, he starts moving towards the enemy sitting in the castle nearby (the threat that forced him sitting still before). But in the next turn the green returns back instead of attacking the enemy, repeating that evolution many times

oleg-derevenetz:
@kitovyj

> It seems that in the first attached game of https://github.com/ihhub/fheroes2/issues/7488 the green hero demonstrates another wierd behaviour now. Instead of sitting in the castle still, he starts moving towards the enemy sitting in the castle nearby (the threat that forced him sitting still before). But in the next turn the green returns back instead of attacking the enemy, repeating that evolution many times

That's why I wrote:

> Although the current behavior may not be ideal either - for example, the hero may run out of the castle in certain circumstances and then return - it is quite difficult to come up with the "right" behavior in circumstances when there is a strong enough opponent near your castle that you do not dare to attack.

In that save, Ruby want to occupy poorly defended enemy castle far away, but when she moves out of the green castle, AI realizes that this castle is now poorly protected against nearby Fineous hero, so Ruby went back on the next turn. On the other hand, Ruby does not dare to attack Fineous because the outcome of the battle in this case is uncertain.

kitovyj:
> @kitovyj
> 
> > It seems that in the first attached game of #7488 the green hero demonstrates another wierd behaviour now. Instead of sitting in the castle still, he starts moving towards the enemy sitting in the castle nearby (the threat that forced him sitting still before). But in the next turn the green returns back instead of attacking the enemy, repeating that evolution many times
> 
> That's why I wrote:
> 
> > Although the current behavior may not be ideal either - for example, the hero may run out of the castle in certain circumstances and then return - it is quite difficult to come up with the "right" behavior in circumstances when there is a strong enough opponent near your castle that you do not dare to attack.
> 
> In that save, Ruby want to occupy poorly defended enemy castle far away, but when she moves out of the green castle, AI realizes that this castle is now poorly protected against nearby hero Fineous, so Ruby went back on the next turn. On the other hand, Ruby does not dare to attack Fineous because the outcome of the battle in this case is uncertain.

Ahh, I see! But is it really uncertain? I think Ruby can easily beat him. Aren't we dealing here with some flaws of battle outcome evaluation function then?

kitovyj:
> @kitovyj
> 
> > It seems that in the first attached game of #7488 the green hero demonstrates another wierd behaviour now. Instead of sitting in the castle still, he starts moving towards the enemy sitting in the castle nearby (the threat that forced him sitting still before). But in the next turn the green returns back instead of attacking the enemy, repeating that evolution many times
> 
> That's why I wrote:
> 
> > Although the current behavior may not be ideal either - for example, the hero may run out of the castle in certain circumstances and then return - it is quite difficult to come up with the "right" behavior in circumstances when there is a strong enough opponent near your castle that you do not dare to attack.
> 
> In that save, Ruby want to occupy poorly defended enemy castle far away, but when she moves out of the green castle, AI realizes that this castle is now poorly protected against nearby hero Fineous, so Ruby went back on the next turn. On the other hand, Ruby does not dare to attack Fineous because the outcome of the battle in this case is uncertain.

PS: I'm also curious if a distance penalty is considered when calculating priorities? I think it's often better to attack closer castle even at the expence of some losses rather than run through the entire map towards presently undefended castle, losing time and leaving your own castles undefended.

oleg-derevenetz:
@kitovyj

> Ahh, I see! But is it really uncertain? I think Ruby can easily beat him. Aren't we dealing here with some flaws of battle outcome evaluation function then?

I have made efforts to fix the evaluation mechanism of threatening heroes and castles, but this needs to be tested further.

> PS: I'm also curious if a distance penalty is considered when calculating priorities? I think it's often better to attack closer castle even at the expense of some losses rather than run through the entire map towards presently undefended castle, losing time and leaving your own castles undefended.

Yes, there is a distance penalty. But castles are always more attractive than heroes in this regard, so a more distant castle may be more attractive than a nearer hero.

oleg-derevenetz:
Hi @ihhub what do you think about this approach? AI heroes just sitting in their castles looks weird indeed and it would be nice to have a fix for this in the next release.

kitovyj:
> @kitovyj
> 
> > Ahh, I see! But is it really uncertain? I think Ruby can easily beat him. Aren't we dealing here with some flaws of battle outcome evaluation function then?
> 
> I have made efforts to fix the evaluation mechanism of threatening heroes and castles, but this needs to be tested further.
> 

Thank you! I'm looking forward to testing the improvements!

ihhub (approved):
Hi @oleg-derevenetz , everything looks fine to me and the behavior of the AI has become better so it is a step forward. We don't need to make a perfection in one shot so even a small step forward is a progress.

ihhub:
Hi @drevoborod , could you please help us to test this pull request? It is needed to be merged for the upcoming release.

drevoborod:
My  first impressions after testing: AI becomes maybe too brave, because it leaves castle when enemy definitely can attack it.
Here are some examples:
1. Here you can attack nearest castle with Luna because AI hero got out: 
[Broken_Alliance_take_castle.zip](https://github.com/ihhub/fheroes2/files/12316426/Broken_Alliance_take_castle.zip)

2. Here yellow AI also leaves its single castle and is being eliminated by red player. At the same time, red player is not so brave: by some reason, he leaves the castle but after couple of steps returns back. Although later easily beats the yellow player (wait for several turns): 
[Test_bravery_0001.zip](https://github.com/ihhub/fheroes2/files/12316455/Test_bravery_0001.zip)

(btw, I finally got the answer to the old question: who is stronger - titans or black dragons? At least here we can see that 10 dragons are definitely stronger than 10 titans :)

So, for me it seems that it needs a little more balancing maybe.

oleg-derevenetz:
Hi @drevoborod

> Here you can attack nearest castle with Luna because AI hero got out:

In this case hero got out to eliminate the threat. Eliminating the threat (is it is possible, and in this case it's possible) has a higher priority than defending a castle (but it depends on distance, because on larger distances defending a castle may become more preferable).

> Here yellow AI also leaves its single castle and is being eliminated by red player. At the same time, red player is not so brave: by some reason, he leaves the castle but after couple of steps returns back.

If you are able to see both players, it doesn't mean they are able to see each other. I want to remind you that AI does not see the entire map, his vision is limited by the fog of war.  We must not forget about this when evaluating his actions. For example, on this map yellow AI just don't see the red castle entrance with the red hero:

fog

Similarly, the red AI does not see the yellow hero when he runs out of the castle.

drevoborod:
> Eliminating the threat (is it is possible, and in this case it's possible) has a higher priority than defending a castle (but it depends on distance

Exactly - it depends on distance, but here we see that AI can't reach my hero in one turn, but my hero can reach its castle. So maybe it's more preferable to wait in such case? Of course, if AI can predict that enemy hero has more movepoints..

> Similarly, the red AI does not see the yellow hero when he runs out of the castle.

Ok, I see. But why red player returns? It's stronger than yellow.. But if you say that it's normal, so I must agree :)
Btw, what tool did you use to show grid and view area of AI players? :) Such tool definitely can help me in testing :)

oleg-derevenetz:
@drevoborod

> Exactly - it depends on distance, but here we see that AI can't reach my hero in one turn, but my hero can reach its castle. So maybe it's more preferable to wait in such case? Of course, if AI can predict that enemy hero has more movepoints.

By "depends on distance" I mean that if there is approximately equal distance between the AI hero and his castle under threat and between the same AI hero and the enemy hero (who threatens the castle), on shorter distances the hero is more preferable and on larger distances the castle is more preferable. If AI hero is already sitting in the castle and there is a threat nearby (less than 30 tiles away) and AI hero is strong enough, he will try to attack.

> Ok, I see. But why red player returns? It's stronger than yellow.

Yes, it's stronger. But according to the current AI rules, if AI player doesn't lose the game, his hero should have at least x1.5 advantage in his strength to attack. In this case advantage is not even close to x1.5.

> Btw, what tool did you use to show grid and view area of AI players? :) Such tool definitely can help me in testing :)

I patch the code to load this save file as a multiplayer save file :)

kitovyj:
With this commit AI behaves much more natural. I don't see major flaws.

oleg-derevenetz:
Hi @ihhub I made a couple of new changes:

* I removed the logic of reinforcing a castle in danger. Currently this logic doesn't fit well, because hero either attack the intruder or sits in the castle, so there is currently no reason to divide hero's forces. Perhaps in the future it will make sense to return it, if we think carefully about the mechanism and conditions of such reinforcement.
* While I'm here, I reworked the mechanics of hero & garrison optimization. Arrangement of troops should be always optimized now (using the `AI::OptimizeTroopsOrder()`) and should always have the weakest units split into free slots (1 per slot).

drevoborod:
Looks like the branch is in active development state so time for testing has not come yet? @oleg-derevenetz

oleg-derevenetz:
@drevoborod

> why red player still returns back to its castle although it is way more powerful now? It definitely could beat yellow player in the first day.

Yes. But when red hero stops upon exiting his castle and re-assesses his targets, castle seems more attractive due to the distance scaling:

Yellow hero: value 6562 becomes 3903 after scaling;
Red castle: value 5250 becomes 4789 after scaling.

So, although initial value of the hero is higher, the castle still wins. Yes, it doesn't look good, although formally of course it may be true - the castle is much closer. Maybe current x1,25-x1.5 handicap is not enough.

> But I'm not sure it's good decision to force such tactics.. I mean that it looks more natural when different players or even heroes use different tactics.

Pro players always use splitting. They split weaker stacks by 1 unit before battles and at the end of the turn (when they can be attacked by an opponent). The point here is that, firstly, they can take the retaliation from powerful enemy stacks (but our AI can't do this yet), and, secondly, they take on extra damage - the whole stack of 10 Titans should make 5 shots to eliminate 5 stacks of 1 peasant, while it make just one shot to eliminate 1 stack of 5 peasants. So it is almost always profitable and there is no reason not to use it.

drevoborod:
Ok, with this new build red player eliminater yellow one in one turn :) But here is another save and question: stronger purple player's hero appears near Luna but does not attack. Is it expected? Why?
[Broken_Alliance_0073.zip](https://github.com/ihhub/fheroes2/files/12320498/Broken_Alliance_0073.zip)

oleg-derevenetz:
@drevoborod

> But here is another save and question: stronger purple player's hero appears near Luna but does not attack. Is it expected? Why?

Because it's not stronger. This hero's strength is estimated at 5674 points and Luna's strength is estimated as 5979 points, not due to the army mostly, but to the Luna's stats - Jaclyn is level 1 hero while Luna is level 14 hero with lot of spell power. Plus attacking hero should have at least x1.3 advantage there, which of course is not the case. In fact, Luna wins here with mass slow and bless, although having lost most of the Phoenixes, so AI estimation is close to reality. Rebecca is a different matter, she has a significant advantage and runs to attack Luna.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -604,33 +575,26 @@ namespace AI
...
             attackTask->second.secondaryTaskTileId.insert( castleIndex );
         }
 
         auto defenseTask = _priorityTargets.find( castleIndex );
         if ( defenseTask == _priorityTargets.end() ) {
-            _priorityTargets[castleIndex] = { PriorityTaskType::DEFEND, enemyStrength, enemyArmy.index };
-        }
-        else if ( defenseTask->second.threatLevel < enemyStrength ) {
-            defenseTask->second.secondaryTaskTileId.insert( defenseTask->first );
-            defenseTask->second.threatLevel = enemyStrength;
```
oleg-derevenetz, line 627: Hi @ihhub do you happen to remember why we added a castle here as a secondary task to itself? The logic is not quite clear to me now.

oleg-derevenetz:
With the latest changes AI should become less brave, especially when defending hero in castle is able to defend his castle, but not able to defeat enemy in the open field.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
ihhub: Hi @oleg-derevenetz , it is a bug. Line 626 should be **defenseTask->second.secondaryTaskTileId.insert( enemyArmy.index );**

drevoborod:
And also it seems that AI became too cautious: just dissmiss 6 swordsmen from the castle where Luna is, and it will be attacked, but with them the AI doesn't dare to attack.
[Broken_Alliance_guard.zip](https://github.com/ihhub/fheroes2/files/12322624/Broken_Alliance_guard.zip)

drevoborod:
> He doesn't lose in my case, but just 2 dragons survive.

I  tried 3 times with same result..

> This behavior is related solely to the army strength estimation,

That's what i'm talking about - if we teach the AI to split its troops, we need to teach it to calculate threats differently.

oleg-derevenetz:
@drevoborod

> That's what i'm talking about - if we teach the AI to split its troops, we need to teach it to calculate threats differently.

What exactly does "differently" mean? :) Also, if human player will split his troops right now in the `master` branch, AI will may fail with their strength assessment as well. This still needs to be gradually improved, if some working ideas will be available.

drevoborod:
> What exactly does "differently" mean?

Have no idea unfortunately :( I also don't understand how threat is being calculated currently, just ask to pay attention on potentially problematic behaviour.

> But this reproducibility is currently limited to a particular machine - more specifically, to the compiler.

It's interesting, didn't know.. Yes, I compiled the engine by myself, so it's defenitely the explanation.

Ok, in my opinion, this PR has to be tested in more different situations - maybe played several dozens of hours on different maps.. But unfortunately I don't have so much time :) In other hand, it works, and what could we loose if something goes wrong? :)

kitovyj:
While testing the artifact of this PR(which I like very much, it seems to be great improvement!), I collected some saves illustrating another issue. In principle the issue may be even not in the scope of the branch, but since I tested the game with this artifact I want to post it here...

Briefly, the problem is: while pursuing distanced targets AI ignores minor targets it passes by. 

The most prominent example is "refused attack the castle" save. AI went a long way toward distanced enemy castle. I stopped one step away from the gate since its move points ended. Then, enemy AI suddenly approached another valuable castle of our protagonist, which is far away. When its turn comes, AI immediately runs back ignoring its original target standing just one move away. The endangered castle is distant : several "turns" away, so why not take the castle first and then run back to defend "the capital"?

Another example is "don't attack nearby" - AI hero pursuits its target far away and passes by another enemy hero. Normally, we would expect it to attack the hero since it would cost almost nothi

ng in terms of game time but the strategic/tactical benefits are great (as in the example).

Other saves illustrates similar situation. 

In general I think it would be great to implement something like "dont ignore castles/heroes/resources standing close to you path while moving to your main target" strategy. It could be a significant amount of code/new structures though(? or maybe it is possible to find relatively simple elegant solution). Some cases may be can be solved with just "distance penalty" tweaking?

[ai-ignores-nearby-targets.zip](https://github.com/ihhub/fheroes2/files/12325793/ai-ignores-nearby-targets.zip)

ihhub:
Hi @oleg-derevenetz , is it correct that we are still in the middle of fine-tuning the behavior of the AI? If yes I am not surprised as with the current implementation we cannot cover all the cases. We don't have a true kingdom planner to define what heroes and castles should do in a collaboration.

If there are things which should be patched or tuned would it be better to extract parts of the code which are definitely beneficial and then merge them into the upcoming release?

oleg-derevenetz:
Hi @ihhub the main goal of this PR is to fix an issue when AI heroes are sitting in their castles like owls when any threat is nearby (even a weaker one), sometimes even blocking each other. I'd say that behavior is much better now - AI now attacks if it can and defends the castle if it have to. There are reports that sometimes AI overestimate/underestimate the army strength (just like in that case with a few Swordsmen described by @drevoborod, that prevent the attack of a castle by more powerful hero while they are in the castle garrison), but issues of such kind are related to army strength estimation logic and are observed in the `master` branch as well.

This PR also forces AI to always have its army arranged for a battle (not just when it attacks, but also after battles/reinforcements/hiring from dwellings on the map, etc) and introduces splitting of the weakest stack into free slots in all cases. @drevoborod has an opinion that army strength estimation should be somehow adapted to such a splitting, but there are still no ideas how exactly this should be done. BTW such a splitting was performed in the `master` branch for castle garrisons and for weak heroes anyway. In general, I think it can be improved later, if there are any ideas. I have none yet.

ihhub:
@oleg-derevenetz , I am okay with the changes as I see benefits from this. Are you okay with merging this pull request?

ihhub:
Hi @kitovyj , described by you scenarios are documented in GitHub issues. We are aware of them and this pull request is not targeted to fix them.

oleg-derevenetz:
@ihhub

> I am okay with the changes as I see benefits from this. Are you okay with merging this pull request?

Yes. If there are new ideas to improve the AI behavior, they can be addressed separately. I don't have any concrete ideas at the moment.

ihhub:
@oleg-derevenetz , awesome work!

PR #7509, opened 2023-08-03
Update the Norwegian Translation

zenseii:
This translates new strings, improves many and translates a few more town names.


ihhub (approved):

ihhub:

Issue #7501, opened 2023-08-02
Implement report issue and attach save file from Android Helper tool


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Right now a user that experiences a problem while playing on Android and wants to attach the save file has to connect the phone to a PC and transfer the save there, and then make the issue from their PC.

I realize this might be problematic because a user would have to have a Github account to actually upload and create an issue.

In addition the privacy policy would have to be updated to reflect how the app would allow users to share save data.

EDIT: We made it possible to extract saves from the Android save folder using the fh2 helper app. There is still no way to attach save files from just the phone so it will have to be transfered to a PC for uploading to github.

### Describe the possible solution

Make save sharing from Android more convenient.

ihhub:
Hi @zenseii , it is doable but I believe we need to have more information how we could implement this. I suggest to focus only on **fh2Toolset** as this is the application which was written for Android devices only. Adding a functionality such as sharing selected save files could be what users want.

oleg-derevenetz:
> I realize this might be problematic because a user would have to have a Github account to actually upload and create an issue.

The user will still need a GitHub account to create an issue here, whether it will happen automatically or manually.

oleg-derevenetz:
> There is still no way to attach save files from just the phone

There is a way - in the GitHub web UI (in its mobile form) you should click on the text edit field and then press the "attach file" button in the button row at the top of the text edit field (with the icon representing some sort of a scenery with a mountain and the sun) and then choose the File app to select the ZIP archive to upload and attach.

oleg-derevenetz:
![fheroes2_save_files.zip](https://github.com/ihhub/fheroes2/files/12389325/fheroes2_save_files.zip)

I just attached this ZIP file via GitHub mobile app :) Sony Xperia with Android 13.

PR #7510, opened 2023-08-03
Android Toolset: add the ability to export and import save files (as a ZIP archive)

oleg-derevenetz:
Related to #7501

https://github.com/ihhub/fheroes2/assets/32623900/ebf186af-4559-4099-8fd3-67211a4db681

The types of saved files imported from the ZIP archive (single player, campaign, multiplayer) can be additionally adjusted by a switch at the top of the screen.


oleg-derevenetz:
Hi @zenseii

> However, perhaps in this case we should show a dialog saying something like "No fheroes2-compatible save files were found in the selected zip archive"?

To be honest, I don't like modal dialogs very much, especially when they are the not the result of a user's immediate actions, but a background processes. We can show the status of the latest action instead, like it's done with OG assets extraction, but IMHO this will be redundant as well, because the user will still see that the no newly imported save files were added to the list.

oleg-derevenetz:
> Also I suggest mentioning #7501 since this is very relevant to it.

It is already mentioned in the first line of the PR description :)

oleg-derevenetz:
Hi @zenseii I have implemented something like a status message that should be shown when something goes wrong:

Screenshot_20230805-181008

oleg-derevenetz:
@zenseii

> The import and export icons are, to me at least, very easy to confuse with upload and download symbols.

I use icons from the Google Material Icons set (Apache license):

https://fonts.google.com/icons?icon.set=Material+Icons

You csn browse it and suggest icons that are more suitable in your opinion.

> I realized the save management button is enabled even when original assets aren't found right after installing the app.

I just see no reason to impose artificial restrictions on what user is able to do if there are no technical reasons for these restrictions. Save file management works just fine even without OG assets (exactly the same as on the desktop BTW), so why we should impose any restrictions? The user will figure out when and how best to use this opportunity.

oleg-derevenetz:
@zenseii

> They are basically the same icons turned 90 degrees.

Yes, and thus they turned into doors :) IMHO it will be difficult for the user to understand how these doors are related to the import/export operations. "Upload/download" is a closer association IMHO.

> Alternatively:

ZIP folder icon - maybe, there is something in this, but I doubt that just a folder icon is a good association with the import operation :)

oleg-derevenetz:
@zenseii

> I guess we could use a mix of the download icon and the zipped icon

So what the zipped icon should be used for? Import operation OPENS the ZIP file and export operation CREATES the ZIP file, so there is some uncertainty as well.

oleg-derevenetz:
@zenseii

> I suppose we could use the zipped archive icon if we had a disabled state for the extract button, in case no file is selected, then enable it once a file is. This would give some feedback as to what button does what.

Currently there is dialog appears when you press the Export button and there are no files selected. I thought it would be better than just a disabled or doing nothing button, because the dialog explains exactly what needs to be done to make it work.

oleg-derevenetz:
@zenseii OK, what do you think about these icons:

Screenshot_20230808-123924

?

ihhub (approved):

ihhub:
@oleg-derevenetz , great thanks for adding this feature!

PR #7519, opened 2023-08-06
Fix Army info button in level up dialog

### LINKED ISSUE #7514: The Hero info button in level up dialog has black background
### Preliminary checks

### Platform

Windows

### Describe the bug

Seen here:
image

It is the same background as it has in the adventure map buttons
image

### Save file

No need.

### Additional info

_No response_

zenseii:
Credits to @Districh-ru for the solution.

We have to use the `Copy()` function to avoid the `_singleLayer` attribute to be set to true, because this happens when you just use the `=-operator` since the regular Army button is rightly set to be single layered after the PR #7449

Fix #7514

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -3152,15 +3152,8 @@ namespace fheroes2
 
                 const int releasedIndex = ( id == ICN::GOOD_ARMY_BUTTON ) ? 0 : 4;
 
-                const Sprite & originalReleased = GetICN( ICN::ADVBTNS, releasedIndex );
-                _icnVsSprite[id][0].resize( originalReleased.width(), originalReleased.height() );
-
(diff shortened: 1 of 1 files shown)


LeHerosInconnu:
Hello @zenseii,

> Hi, @LeHerosInconnu. Could you please verify if this fixes the problem with the button?

I did a quick test with this save file: [Hero button 01.zip](https://github.com/ihhub/fheroes2/files/12267534/Hero.button.01.zip), and it works fine.

Districh-ru (changes requested):
Hi, @zenseii, I left a couple of comments. Could you take a look when you have time.
And we also have to fix this button for the Evil interface the same way at these lines:
https://github.com/ihhub/fheroes2/blob/13c5cf9bd6ab784a8eaac208b8040cb095786985/src/fheroes2/agg/agg_image.cpp#L3180-L3181

Districh-ru (approved):
Well done!
изображение изображение

ihhub (approved):

ihhub:

PR #7521, opened 2023-08-06
Implement basic terrain draw in Map Editor (currently without terrain connections)

Districh-ru:
Relates to #6845

This PR adds Editor ability to place (draw) specified terrain on a map by 1x1, 2x2 or 4x4 brush or by selecting a rectangle area on map.
Terrain interconnections are not implemented in this PR and they will be done in one of the next Editor related PRs.
Currently you cannot hold mouse button and draw withe selected brush by moving the cursor as it is done in original Editor, jus one draw in the selected area by one mouse click. This function will be implemented after the implementation of terrain interconnection.

https://github.com/ihhub/fheroes2/assets/113276641/4cb8e4e7-d937-46d4-94fd-f41c8041ee53

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -91,17 +91,34 @@ namespace Interface
 
     void Editor::redraw( const uint32_t force /* = 0 */ )
     {
-        const fheroes2::Display & display = fheroes2::Display::instance();
+        fheroes2::Display & display = fheroes2::Display::instance();
 
         const uint32_t combinedRedraw = _redraw | force;
 
         if ( combinedRedraw & REDRAW_GAMEAREA ) {
             // Render all except the fog.
-            _gameArea.Redraw( fheroes2::Display::instance(), LEVEL_OBJECTS | LEVEL_HEROES | LEVEL_ROUTES );
+            _gameArea.Redraw( display, LEVEL_OBJECTS | LEVEL_HEROES | LEVEL_ROUTES );
 
             // TODO:: Render horizontal and vertical map tiles scale and highlight with yellow text cursor position.
 
             // TODO: Render square cursor around tiles according to Brush Size.
+            if ( _editorPanel.isTerrainEdit() ) {
+                LocalEvent & le = LocalEvent::Get();
+                const fheroes2::Point & mousePosition = le.GetMouseCursor();
+                const int32_t index = _gameArea.GetValidTileIdFromPoint( mousePosition );
+
+                if ( index > -1 ) {
+                    const int32_t brushSize = _editorPanel.getBrushSize();
+
+                    if ( brushSize > 0 ) {
+                        const int32_t worldWidth = world.w();
+                        const int32_t cursorSizeX = std::min( brushSize, worldWidth - index % worldWidth ) - 1;
+                        const int32_t cursorSizeY = std::min( brushSize, world.h() - index / worldWidth ) - 1;
+                        const int32_t endIndex = index + cursorSizeX + worldWidth * cursorSizeY;
+                        _gameArea.renderTileCursor( display, index, endIndex );
+                    }
+                }
+            }
         }
 
         if ( combinedRedraw & ( REDRAW_RADAR_CURSOR | REDRAW_RADAR ) ) {
@@ -494,7 +511,7 @@ namespace Interface
                         break;
                     }
 
-                    brushTile = &world.GetTiles( Maps::GetDirectionIndex( brushTile->GetIndex(), Direction::BOTTOM ) );
+                    brushTile += world.w();
                 }
             }
             else {
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -58,7 +58,7 @@ class Castle;
 
 namespace
 {
-    int32_t brushAreaEndIndex( const int32_t brushSize, const int32_t startIndex )
+    int32_t getBrushAreaEndIndex( const int32_t brushSize, const int32_t startIndex )
     {
         const int32_t worldWidth = world.w();
         const int32_t cursorSizeX = std::min( brushSize, worldWidth - startIndex % worldWidth ) - 1;
@@ -113,16 +113,15 @@ namespace Interface
 
             // TODO:: Render horizontal and vertical map tiles scale and highlight with yellow text cursor position.
 
-            if ( _editorPanel.isTerrainEdit() ) {
-                if ( _tileUnderCursor > -1 ) {
-                    const int32_t brushSize = _editorPanel.getBrushSize();
+            if ( _editorPanel.isTerrainEdit() && ( _tileUnderCursor > -1 ) ) {
+                const int32_t brushSize = _editorPanel.getBrushSize();
 
-                    if ( brushSize > 0 ) {
-                        _gameArea.renderTileCursor( display, _tileUnderCursor, brushAreaEndIndex( brushSize, _tileUnderCursor ) );
-                    }
-                    else if ( _editorPanel.isAreaSelect() ) {
-                        _gameArea.renderTileCursor( display, _selectedTile, _tileUnderCursor );
-                    }
+                if ( brushSize > 0 ) {
+                    _gameArea.renderTileAreaSelect( display, _tileUnderCursor, getBrushAreaEndIndex( brushSize, _tileUnderCursor ) );
+                }
+                else if ( brushSize == 0 ) {
+                    // Render area selection from the tile where the left mouse button was pressed till the tile under the cursor.
+                    _gameArea.renderTileAreaSelect( display, _selectedTile, _tileUnderCursor );
                 }
             }
         }
@@ -286,7 +285,9 @@ namespace Interface
                 if ( Cursor::POINTER != cursor.Themes() ) {
                     cursor.SetThemes( Cursor::POINTER );
                 }
-                if ( !_gameArea.isDragScroll() && ( !_editorPanel.isAreaSelect() || _selectedTile == -1 ) ) {
+
+                // TODO: Add checks for object placing/moving, and other Editor functions that uses mouse dragging.
+                if ( !_gameArea.isDragScroll() && ( _editorPanel.getBrushSize() > 0 || _selectedTile == -1 ) ) {
                     _radar.QueueEventProcessing();
                 }
             }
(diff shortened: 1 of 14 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -54,6 +56,17 @@
 
 class Castle;
 
+namespace
+{
+    int32_t brushAreaEndIndex( const int32_t brushSize, const int32_t startIndex )
```
ihhub, line 61: Let's name it as **getBrushAreaEndIndex** to be more accurate about the function.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -90,15 +103,28 @@ namespace Interface
...
+            if ( _editorPanel.isTerrainEdit() ) {
+                if ( _tileUnderCursor > -1 ) {
+                    const int32_t brushSize = _editorPanel.getBrushSize();
+
+                    if ( brushSize > 0 ) {
+                        _gameArea.renderTileCursor( display, _tileUnderCursor, brushAreaEndIndex( brushSize, _tileUnderCursor ) );
+                    }
+                    else if ( _editorPanel.isAreaSelect() ) {
+                        _gameArea.renderTileCursor( display, _selectedTile, _tileUnderCursor );
+                    }
```
ihhub, line 125: Is there a valid scenario when **brushSize == 0** and **!_editorPanel.isAreaSelect()**? If not then we shouldn't check for **_editorPanel.isAreaSelect()**. According to the logic no harm to check **brushSize >= 0** as `brushAreaEndIndex` will always return the same tile id.

The same for line 543.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -260,7 +286,7 @@ namespace Interface
                 if ( Cursor::POINTER != cursor.Themes() ) {
                     cursor.SetThemes( Cursor::POINTER );
                 }
-                if ( !_gameArea.isDragScroll() ) {
+                if ( !_gameArea.isDragScroll() && ( !_editorPanel.isAreaSelect() || _selectedTile == -1 ) ) {
```
ihhub, line 289: I would suggest to at least add a TODO as later we are going to add objects and other things. This would help to remember to update this place.

src/fheroes2/editor/editor_interface.h
```diff
@@ -46,12 +46,20 @@ namespace Interface
         static fheroes2::GameMode eventFileDialog();
         void eventViewWorld();
 
+        bool useMouseDragMovment() override
```
ihhub, line 49: **Movment** is misspelled.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -71,6 +71,25 @@ namespace Interface
...
+            return 1;
+        case BrushSize::MEDIUM:
+            return 2;
+        case BrushSize::LARGE:
+            return 4;
+        case BrushSize::AREA:
+            return -1;
+        default:
+            // Have you added a new Brush size? Update the logic above!
+            assert( 0 );
```
ihhub, line 87: Please add break here just for the sake of good style practices.

src/fheroes2/editor/editor_interface_panel.h
```diff
@@ -43,6 +44,28 @@ namespace Interface
...
+        {
+            return _getGroundId( _selectedTerrain );
+        }
+
+        bool isTerrainEdit() const
+        {
+            return _selectedInstrument == Instrument::TERRAIN;
+        }
+
+        bool isAreaSelect() const
```
ihhub, line 59: I recommend to be more specific and include terrain word into the name as it is not obvious what kind of area it is.

src/fheroes2/editor/editor_interface_panel.h
```diff
@@ -58,7 +81,11 @@ namespace Interface
     private:
         Editor & _interface;
 
-        static const char * _getTerrainTypeName( const uint8_t brushId );
+        static int _getGroundId( const uint8_t brushId );
+        static const char * _getTerrainTypeName( const uint8_t brushId )
+        {
+            return Maps::Ground::String( _getGroundId( brushId ) );
+        }
```
ihhub, line 88: Please add empty lines in between method for visual separation.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -71,6 +71,25 @@ namespace Interface
...
+    {
+        switch ( _selectedBrushSize ) {
+        case BrushSize::SMALL:
+            return 1;
+        case BrushSize::MEDIUM:
+            return 2;
+        case BrushSize::LARGE:
+            return 4;
+        case BrushSize::AREA:
+            return -1;
```
ihhub, line 84: Would it be better to just return 0 together with default so the logic for (if any in the future) any new brush size would be still area rather than trying to handle 0-size areas?

Then the logic for line 120-125 would be much simplier.

src/fheroes2/maps/maps_tiles.h
```diff
@@ -333,6 +333,12 @@ namespace Maps
             return _terrainImageIndex;
         }
 
+        void setTerrainImage( const uint16_t terrainImageIndex, const bool horizontalFlip, const bool verticalFlip )
+        {
+            _terrainImageIndex = terrainImageIndex;
+            _terrainFlags = static_cast<uint8_t>( verticalFlip ) + 2 * static_cast<uint8_t>( horizontalFlip );
```
ihhub, line 339: Please do not use mathematical operations on boolean variables. While in theory a boolean variable should contain 0 or 1, in reality it depends :) I suggest to do this way:
```cpp
_terrainFlags = ( verticalFlip ? : 1 : 0 ) + ( horizontalFlip ? 2 : 0 );
```

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -231,6 +233,33 @@ namespace
 
 namespace Maps
 {
+    void setTerrainImageOnTiles( const int32_t startTileId, const int32_t endTileId, const int groundId )
+    {
+        if ( startTileId < 0 || endTileId < 0 ) {
+            return;
+        }
```
ihhub, line 240: Let's add a check for map size as well. It would be a case when **startTileId** or **endTileId** are beyond the map size.

src/fheroes2/maps/maps_tiles.h
```diff
@@ -333,6 +333,12 @@ namespace Maps
             return _terrainImageIndex;
         }
 
+        void setTerrainImage( const uint16_t terrainImageIndex, const bool horizontalFlip, const bool verticalFlip )
```
ihhub, line 336: Essential we don't set a terrain image here but terrain type / information. Should we just name it as **setTerrain**?

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -770,6 +770,29 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
     updateObjectAnimationInfo();
 }
 
+void Interface::GameArea::renderTileCursor( fheroes2::Image & dst, const int32_t startTile, const int32_t endTile ) const
```
ihhub, line 773: It is not a cursor but highlighted area / active area / selected area. **renderTileCursor** might be interpret as `call this function to render a cursor of a tile`.

ihhub (changes requested):
Hi @Districh-ru , I left few comments in this pull request. Would you mind to take a look when you have free time?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Hi, @ihhub, sure!

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Yes, it is not a valid scenario right now and possibly ever.
I removed the `isAreaSelect()` method and now `brushSize == 0` will be for the tile area select.

src/fheroes2/editor/editor_interface.cpp
Districh-ru: OK.

src/fheroes2/editor/editor_interface.h
Districh-ru: Thanks!

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Sure. :)

src/fheroes2/editor/editor_interface_panel.h
Districh-ru: This method is not used anymore, but in future we will select an area to clear the objects using this logic for area selection. So if we will restore it in the future we will call it `isTileAreaSelect()`.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: For Area Select we will return 0, but for the default I'll leave the assertion as if we will add a new brush size we will have to make aтв test the logic for it.

src/fheroes2/maps/maps_tiles.h
Districh-ru: Thanks!
I will take this into account in the future.

src/fheroes2/maps/maps_tiles.h
Districh-ru: OK.

src/fheroes2/gui/interface_gamearea.cpp
Districh-ru: Yes, lets name it `renderTileAreaSelect()` as we render the selection of map tiles (in this PR for filling the terrain).

ihhub (approved):

ihhub:
Hi @drevoborod , could you please also test this pull request as well?

Branikolog:
Hi, @Districh-ru !
Terrain generation works fine!
I've only caught some minor UI bugs.
Adventure map frame is not rendered correctly for my selected custom 976x550 resolution:
frame corner

144x144 button has 2px text shift on pressed state:
ezgif com-resize (2)

ihhub:
Hi @Branikolog and @Districh-ru , font related things we should fix in a separate pull request.

Districh-ru:
> Hi, @Districh-ru ! Terrain generation works fine! I've only caught some minor UI bugs. Adventure map frame is not rendered correctly for my selected custom 976x550 resolution: 

Hi, @Branikolog, this bug will be addressed later. There is a TODO in `editor_interface.cpp` near the Status window redraw call.
First we will modernize this Status window to display Editor related info. And then we will adopt it to the interface.

> 144x144 button has 2px horizontal text shift on pressed state:

Currently we use the resources form original game assets. Later we may make these buttons to have generated text or use a dialog instead of the menu to set the map size (as in future we may plan to have more than four original map sizes, e.g. like it is done in HoMM3: HotA map editor). So let's address it later after we make the new map format.

Branikolog (approved):
Glad to hear that, @Districh-ru !

ihhub:

Issue #4352, opened 2021-10-10
Battle-only dialog: multiple issues


a1exsh:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Linux

### Describe the bug

There are a number of issues with the Battle-only dialog:
![image](https://user-images.githubusercontent.com/489601/136689028-2bf09948-4358-48e4-b923-d174c82e40ea.png)

1. The Evil and Good styles are mixed: the outer border may be Evil, but the inner border is always Good.
2. The `OKAY` / `CANCEL` buttons in selection dialogs may use style different from the main `OKAY` button.
3. The pressed states of the buttons have opaque corners that stay after moving the mouse pointer away from the button and releasing the press.
   This problem exists in both Good and Evil styles, it also exists for the main dialog OKAY button.

### Save file

No save file needed.

Branikolog:
Hello, @a1exsh.
Here I can also add that we window background looks grainy here. Is it original texture applied?

ihhub:
Hi @a1exsh , please take a note that this battle mode is not a part of the original game and it has the lowest priority.

Branikolog:
I also propose to remove yellow arrows from this dialog not to confuse players.
![Battle only](https://user-images.githubusercontent.com/55348946/230593428-b6f929fe-f06e-4c32-bccd-2d6c3f9b0f59.png)
Human/AI checkboxes should be moved a bit so translated strings not go out of the window.
I also think instant battle shouldn't be initiated for the battle only mode. I find it a bit weird when I select heroes and armies, press OK to start a battle only mode (which is a mode for battles) and watch the result instantly. 
I couldn't find any opened issue for the thing, but I remember someone has proposed it already:
this dialog should remember all elements you've set before. I personally often forgot something while setting battle settings so I have to restart and set everything once again. 
There could also be a button to reset all values to default.

ihhub:
Hi @a1exsh , please update your original message as most of issues have been fixed.

a1exsh:
@ihhub updated — thanks for coming around to fix those :)  The only thing I see now is the mixing of evil/good styles.
@Branikolog the arrows are still there and are inactive, so probably makes sense to paint them over.

PR #7522, opened 2023-08-06
Rework the "select from list" dialogs used in Battle only mode to use them in Editor in future

Districh-ru:
Relates to https://github.com/ihhub/fheroes2/issues/6845 as these "select from list" dialogs are proposed to be used in Map Editor (to select map objects, edit a hero, ...).

This PR implements generation of the text background for `StandardWindow` class. This will allow to make a shaded text background of any size for use in dialog boxes (for example, redesign the Save/Load dialog to allow more files in the list and display longer filenames).

This PR fixes some issues of https://github.com/ihhub/fheroes2/issues/4352:
- use background, scrollbar and buttons for the selected interface type ( Good / Evil );
- cut long phrases with "..." at the end not to overlap the scrollbar;
- add shadows to OKAY and CANCEL buttons (and the scrollbar).

This PR also fixes:
- add the right click info window for all current list types;
- redraw delay after pressing CANCEL in Select Secondary Skill dialog;
- possible rendering bug of scrollbar image restorer when closing dialog (which has its own image restorer, like "Standard Window") with the scroll bar on it. (This issue cloud be observed before the previous redraw issue was fixed.)
- the vertical size and the count of shown list elements are calculated from the current screen height.

<details><summary>Screenshots:</summary>

</details>


src/fheroes2/gui/ui_scrollbar.h
```diff
@@ -32,8 +32,14 @@ namespace fheroes2
     {
     public:
         Scrollbar();
+        Scrollbar( const Scrollbar & ) = delete;
+        Scrollbar & operator=( const Scrollbar & ) = delete;
 
-        ~Scrollbar() override = default;
+        ~Scrollbar() override
+        {
+            // We need to reset the ImageRestorer for the scroll bar to avoid its restoration by destructor and possible race with th dialog window restorer.
```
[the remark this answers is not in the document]
Districh-ru: Hi @zenseii, thanks for the spell check!

https://github.com/ihhub/fheroes2/issues/7534 issue will be addressed later - we should exclude already used skills (and heroes) from the list in list constructor, e.g. by passing it a vector of already used skills.

Districh-ru:
> Well done! I see that the builds have failed. Not sure why.

This PR https://github.com/ihhub/fheroes2/pull/7532 should fix Linux build issues.

Branikolog:
Hi, @Districh-ru !
Everything is fine except text alignment in lists.
text2
text1

Districh-ru:
Hi, @Branikolog, thanks for noticing this! Done.
изображение изображение

Branikolog (approved):

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -52,37 +59,92 @@ class SelectEnum : public Interface::ListBox<int>
...
+        fheroes2::addGradientShadow( buttonCancelSprite, display, buttonCancel.area().getPosition(), { -5, 5 } );
+        buttonCancel.draw();
+    }
 
-        fheroes2::Blit( fheroes2::AGG::GetICN( ICN::LISTBOX, 9 ), display, dst.x + area.width - 25, dst.y + area.height - 74 );
+    void RedrawBackground( const fheroes2::Point & /* unused */ ) override
+    {
+        if ( listBackground != nullptr ) {
+            listBackground->restore();
+        }
```
ihhub, line 144: Here we have a check for an empty unique pointer while we don't have such check for **background** in the code below. I suggest to delete default constructor forcing to use the defined constructor. In this case it is guaranteed that both unique pointers are initialized (std::make_unique will throw an exception if it is not the case).

What do you think?

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -102,268 +164,265 @@ class SelectEnum : public Interface::ListBox<int>
...
 };
 
 class SelectEnumArtifact : public SelectEnum
 {
 public:
-    using SelectEnum::SelectEnum;
+    explicit SelectEnumArtifact( const fheroes2::Size & rt )
+        : SelectEnum( rt )
+    {
+        constexpr int offset = 42;
```
ihhub, line 314: **constexpr** is not useful for variables like this. Compiler will not optimize them. It is useful for expressions / formulas to be calculated during compilation.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -102,268 +164,265 @@ class SelectEnum : public Interface::ListBox<int>
...
 
-    // setup cursor
-    const CursorRestorer cursorRestorer( true, Cursor::POINTER );
-
-    std::vector<int> skills( MAXSECONDARYSKILL * 3, 0 );
-
-    for ( int i = 0; i < MAXSECONDARYSKILL * 3; ++i )
+    for ( int i = 0; i < MAXSECONDARYSKILL * 3; ++i ) {
         skills[i] = i;
+    }
```
ihhub, line 392: I think you would like to use **std::iota** for the same purpose.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -375,121 +434,55 @@ Artifact Dialog::SelectArtifact()
...
 
-    for ( size_t i = 0; i < monsters.size(); ++i )
-        monsters[i] = static_cast<int>( i + 1 ); // skip Monster::UNKNOWN, safe to do this as the monsters of spells can't be more than 2 billion
-
-    Dialog::FrameBorder frameborder( { 260, 280 }, fheroes2::AGG::GetICN( ICN::TEXTBAK2, 0 ) );
-    const fheroes2::Rect & area = frameborder.GetArea();
+    for ( size_t i = 0; i < monsters.size(); ++i ) {
+        // Skip Monster::UNKNOWN, safe to do casting as the monsters can't be more than 2 billion
+        monsters[i] = static_cast<int>( i + 1 );
+    }
```
ihhub, line 455: In the code above we used **for ( int i = 0; i < MAXSECONDARYSKILL * 3; ++i ) {** while here we rely on vector size. Let's be consistent :)

src/fheroes2/dialog/dialog_selectitems.h
```diff
@@ -32,11 +32,11 @@
...
 namespace Dialog
 {
-    Monster SelectMonster( int id = Monster::UNKNOWN );
-    int SelectHeroes( int cur = Heroes::UNKNOWN );
-    Artifact SelectArtifact();
-    Spell SelectSpell( int id = Spell::NONE );
-    Skill::Secondary SelectSecondarySkill();
+    Monster selectMonster( const int monsterId = Monster::UNKNOWN );
+    int selectHeroes( const int heroId = Heroes::UNKNOWN );
+    Artifact selectArtifact();
```
ihhub, line 37: Nothing to do here: it is interesting that this function doesn't have a default parameter.

ihhub (changes requested):
Hi @Districh-ru , I left few small suggestions here. Would you mind to take a look at them?

src/fheroes2/gui/ui_window.cpp
```diff
@@ -268,6 +268,38 @@ namespace fheroes2
         ApplyTransform( _output, _windowArea.x + _windowArea.width - borderSize - 1, _windowArea.y + _windowArea.height, 1, borderSize - 1, 5 );
     }
 
+    void StandardWindow::applyTextBackgroundShading( const Rect roi )
+    {
+        fheroes2::Rect shadingRoi = roi ^ _activeArea;
```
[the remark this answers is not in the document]
Districh-ru: Hi @ihhub, sure.

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Yes, let's disable the default constructor so we don't need to check `listBackground` and `background` for `nullptr`.

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Thanks for the explanation!

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Sure, thanks for the suggestion
Currently we may use it, but in future to address https://github.com/ihhub/fheroes2/issues/7534 we should pass a vector of skills to exclude and skip them (using `continue;`) in such loop. (Or we may find a better solution.)

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: OK :)
Now all such loops are replaced with `std::iota()`.

src/fheroes2/dialog/dialog_selectitems.h
Districh-ru: Initially this functions and `selectSecondarySkill()` had no default parameters, but lets be consistent for future use (e.g. for selecting the existing artifact/skill and changing it in Editor).

ihhub (approved):

ihhub:
@Districh-ru , great job!

PR #7532, opened 2023-08-09
Fix a few recently introduced code smells, fix GitHub Actions build with ASAN & TSAN

oleg-derevenetz:
Fixes issues like this one:

https://github.com/ihhub/fheroes2/actions/runs/5809318729/job/15748134142

Apparently, `libasan` and `libtsan` are broken or not installed on the current revision of Ubuntu 20.04 image from GitHub.

### DIFF
--- a/.github/workflows/clang_analyzer.yml
+++ b/.github/workflows/clang_analyzer.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   clang:
     name: Clang Analyzer
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 30
     defaults:
       run:
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   tidy:
     name: Clang-Tidy
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


ihhub:
Hi @oleg-derevenetz , could you please explain why we are changing label of docker images to latest? According to [this page](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) we can still use Ubuntu 20.

oleg-derevenetz:
Hi @ihhub

> could you please explain why we are changing label of docker images to latest? According to [this page](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) we can still use Ubuntu 20.

Yes, we can, and we still will use it for builds where we produce build artifacts (such as Linux release builds). But 20.04 is currently broken:

https://github.com/ihhub/fheroes2/actions/runs/5809318729/job/15748134142
https://github.com/ihhub/fheroes2/actions/runs/5819911892/job/15779434571

There are problems with `libasan` and `libtsan`, they seems to be installed, but they are broken somehow, so builds with `FHEROES2_WITH_ASAN` and `FHEROES2_WITH_TSAN` fail on 20.04.

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for keeping the code in the best shape!

PR #7542, opened 2023-08-12
AI heroes shouldn't know what exactly a Treasure or Sea Chest contains

ihhub:
AI heroes "knew" what is inside Treasure and Sea chests and it was visible when they ignore some chests while moving to more distant ones. This is wrong and breaks user experience of "fair" AI.

You can check this save file on master branch and this branch: 
[Chests.zip](https://github.com/ihhub/fheroes2/files/12327199/Chests.zip)

What is done in this pull request:
- make Treasure and Sea chests to have the "same" averaged value, removing the knowledge what it inside (except the fact that there is an important for the human player artifact)
- Treasure Chest should have the same distance modifier as Sea Chest
- artifact check was missing for the chests for Courier heroes
- an assertion was triggered for Sea Chests when an AI hero picks it up on the way

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1066,12 +1066,10 @@ namespace AI
         case MP2::OBJ_SEA_CHEST:
         case MP2::OBJ_TREASURE_CHEST: {
             const Artifact art = getArtifactFromTile( tile );
-            if ( art.isValid() ) {
+            if ( art.isValid() && isFindArtifactVictoryConditionForHuman( art ) ) {
                 // WINS_ARTIFACT victory condition does not apply to AI-controlled players, we should leave this object untouched for the human player.
-                if ( isFindArtifactVictoryConditionForHuman( art ) ) {
-                    assert( 0 );
-                    return -dangerousTaskPenalty;
-                }
+                assert( 0 );
+                return -dangerousTaskPenalty;
             }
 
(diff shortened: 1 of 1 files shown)


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1730,8 +1728,21 @@ namespace AI
...
         case MP2::OBJ_GENIE_LAMP:
-        case MP2::OBJ_RESOURCE:
-        case MP2::OBJ_SEA_CHEST: {
+        case MP2::OBJ_RESOURCE: {
+            return twoTiles;
+        }
+        case MP2::OBJ_SEA_CHEST:
+        case MP2::OBJ_TREASURE_CHEST: {
+            if ( getArtifactFromTile( tile ).isValid() ) {
+                const Artifact art = getArtifactFromTile( tile );
```
oleg-derevenetz, line 1737: We can do some optimization here (just like in the above case I believe):

```cpp
const Artifact art = getArtifactFromTile( tile );
if ( art.isValid() ) {
[...]
```

What do you think?

In general, AI should not be able to take these artifacts anyway because `HeroesValidObject()` should not allow an access to these objects, so these checks exist just in case. But I think they won't hurt anyway.

oleg-derevenetz (commented):
Hi @ihhub I just left one optional suggestion, could you please take a look when possible?

oleg-derevenetz (approved):

PR #7550, opened 2023-08-12
Fix marketplace assertion if one of the resources is not selected

### LINKED ISSUE #7547: Assertion when trading resources
### Preliminary checks

### Platform

Windows

### Describe the bug

Assertion triggers when clicking on the resource to sell in the trading dialog.
trade-assertion

### Save file

[assertion.zip](https://github.com/ihhub/fheroes2/files/12327265/assertion.zip)

### Additional info

_No response_

Districh-ru:
Fix #7547

If one (or both) of the resources is not selected (this will always happen at the beginning because we cannot select both resources at the same time) their type is `Resource::Unknown` - this means that resource is not selected.
This case is now handled and `getTradeCost()` returns 0 as we may get only nothing for nothing.


ihhub (approved):

ihhub:
@Districh-ru , thank you so much for the quick fix!

PR #7552, opened 2023-08-12
Fix issue when AI hero "jumps back" during the fog discovery, various fixes & improvements

### LINKED ISSUE #7546: AI hero jumps during movement
### Preliminary checks

### Platform

Windows

### Describe the bug

AI heroes jump while walking. It should be the case. Hit F8 for AI mode and see what happens to the hero.

### Save file

[Chests.zip](https://github.com/ihhub/fheroes2/files/12327191/Chests.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #7546

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -208,9 +208,7 @@ namespace
 
         if ( object == MP2::OBJ_HEROES ) {
             const Heroes * hero = tile.GetHeroes();
-            if ( hero == nullptr ) {
-                return {};
-            }
+            assert( hero != nullptr );
 
             if ( hero->isFriends( kingdomColor ) ) {
                 return {};
@@ -230,9 +228,7 @@ namespace
 
         if ( object == MP2::OBJ_CASTLE ) {
             const Castle * castle = world.getCastleEntrance( Maps::GetPoint( tileIndex ) );
-            if ( castle == nullptr ) {
-                return {};
-            }
+            assert( castle != nullptr );
 
             // Neutral castles don't pose a threat because they can't hire heroes
             if ( castle->GetColor() == Color::NONE || castle->isFriends( kingdomColor ) ) {
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -208,7 +208,10 @@ namespace
 
         if ( object == MP2::OBJ_HEROES ) {
             const Heroes * hero = tile.GetHeroes();
-            assert( hero != nullptr );
+            // TODO: this function can be called when the game world is not fully initialized yet
+            if ( hero == nullptr ) {
+                return {};
+            }
 
             if ( hero->isFriends( kingdomColor ) ) {
                 return {};
@@ -228,7 +231,10 @@ namespace
 
         if ( object == MP2::OBJ_CASTLE ) {
             const Castle * castle = world.getCastleEntrance( Maps::GetPoint( tileIndex ) );
-            assert( castle != nullptr );
+            // TODO: this function can be called when the game world is not fully initialized yet
+            if ( castle == nullptr ) {
+                return {};
+            }
 
             // Neutral castles don't pose a threat because they can't hire heroes
             if ( castle->GetColor() == Color::NONE || castle->isFriends( kingdomColor ) ) {
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -966,7 +966,7 @@ void Heroes::calculatePath( int32_t dstIdx )
         return;
     }
 
-    path.setPath( world.getPath( *this, dstIdx ), dstIdx );
+    path.setPath( world.getPath( *this, dstIdx ) );
 
     if ( !path.isValidForMovement() ) {
         path.Reset();
@@ -1657,6 +1657,7 @@ void Heroes::Dismiss( int reason )
 
     modes = 0;
 
+    path.Hide();
     path.Reset();
 
     SetMove( false );
@@ -1705,7 +1706,7 @@ void Heroes::ActionNewPosition( const bool allowMonsterAttack )
 
         if ( !targets.empty() ) {
             SetMove( false );
-            GetPath().Hide();
+            ShowPath( false );
 
             // first fight the monsters on the destination tile (if any)
             MapsIndexes::const_iterator it = std::find( targets.begin(), targets.end(), GetPath().GetDestinationIndex() );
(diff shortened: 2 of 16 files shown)


src/fheroes2/heroes/route.h
```diff
@@ -141,18 +183,17 @@ namespace Route
...
-        int GetAllowedSteps() const;
-        static int GetIndexSprite( int from, int to, int mod );
+        static int GetIndexSprite( const int from, const int to, const int cost );
 
     private:
         friend StreamBase & operator<<( StreamBase &, const Path & );
         friend StreamBase & operator>>( StreamBase &, Path & );
 
         const Heroes * _hero;
-        int32_t _dst;
```
oleg-derevenetz, line 155: The original logic with a separate destination field has always seemed superfluous to me, so I decided to simplify it a bit.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1735,17 +1736,17 @@ void Heroes::ActionNewPosition( const bool allowMonsterAttack )
...
 
-// Move hero to a new position. This function applies no action and no penalty
 void Heroes::Move2Dest( const int32_t dstIndex )
 {
     const int32_t currentIndex = GetIndex();
 
-    if ( dstIndex != currentIndex ) {
-        world.GetTiles( currentIndex ).SetHeroes( nullptr );
-        SetIndex( dstIndex );
-        Scout( dstIndex );
```
oleg-derevenetz, line 1746: Calling `Scout()` in this method turned out not to be the best idea, because this method is quite low-level and is called, for example, when loading an MP2 map file and not all data structures are properly initialized yet. Calling `Scout()` in such conditions is quite risky. In addition, it turns out that with normal walking, this call is superfluous. So I decided to remove this call from here and call it explicitly where needed (casting movement spells, Stone Liths and so on).

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -208,9 +208,7 @@ namespace
 
         if ( object == MP2::OBJ_HEROES ) {
             const Heroes * hero = tile.GetHeroes();
-            if ( hero == nullptr ) {
-                return {};
-            }
+            assert( hero != nullptr );
```
oleg-derevenetz, line 211: ... but now we can replace this check by an assertion and survive the start of a new map!

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -230,9 +228,7 @@ namespace
 
         if ( object == MP2::OBJ_CASTLE ) {
             const Castle * castle = world.getCastleEntrance( Maps::GetPoint( tileIndex ) );
-            if ( castle == nullptr ) {
-                return {};
-            }
+            assert( castle != nullptr );
```
oleg-derevenetz, line 231: ... and replace this check by an assertion as well.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -864,10 +864,7 @@ bool Heroes::Recruit( const int col, const fheroes2::Point & pt )
 
     // After recruiting a hero we reveal map in hero scout area.
     Scout( GetIndex() );
-    if ( isControlHuman() ) {
-        // And the radar image map for human player.
-        ScoutRadar();
```
oleg-derevenetz, line 869: Why `ScoutRadar()` is sometimes called only for a human player and sometimes it is always called - for example here:

https://github.com/ihhub/fheroes2/blob/cbb8594829cb7c302218342152c6230dd19689c7/src/fheroes2/heroes/heroes.cpp#L1584-L1587

?

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1116,35 +1113,37 @@ bool Heroes::PickupArtifact( const Artifact & art )
...
-                return true;
-            }
+    const auto scout = [this]( const int32_t artifactID ) {
+        const std::vector<fheroes2::ArtifactBonus> & bonuses = fheroes2::getArtifactData( artifactID ).bonuses;
+        if ( std::find( bonuses.begin(), bonuses.end(), fheroes2::ArtifactBonus( fheroes2::ArtifactBonusType::AREA_REVEAL_DISTANCE ) ) == bonuses.end() ) {
             return false;
-        };
-
-        // If the scout area bonus is increased with the new artifact we update the radar.
-        if ( scout( art.GetID() ) ) {
```
oleg-derevenetz, line 1139: Why this scouting was originally performed only for a human player? AI should have benefits from artifacts with `AREA_REVEAL_DISTANCE` bonus as well. This is video from the `master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/4b97f6d1-b653-4229-bb6c-11952419a284

Hero gets a bonus only when he already starts moving in the wrong direction. This is video from this PR:

https://github.com/ihhub/fheroes2/assets/32623900/d51e77b1-d944-4393-8c0c-e592e96fcd7a

Hero sees the chest immediately after picking up the artifact.

How about `ScoutRadar()` call in this case?

oleg-derevenetz (commented):
Hi @Districh-ru could you please answer a couple of questions as an author of PR #6514?

src/fheroes2/heroes/heroes.cpp
Districh-ru: Hi, @oleg-derevenetz.
`ScoutRadar()` is useful only for human player because in other case the radar image is hidden and `ScoutRadar()` call is redundant. So it is better to always check for `isControlHuman()` before calling `ScoutRadar()`. Or simply add ```     if ( !isControlHuman() ) {        return;    }``` to `ScoutRadar()`.
For AI-controlled player radar is always updated fully here:
https://github.com/ihhub/fheroes2/blob/708ed576430f6563a0465e6f962001fc9d6b4cd5/src/fheroes2/game/game_startgame.cpp#L790-L796

When making https://github.com/ihhub/fheroes2/pull/6514 first I placed `ScoutRadar()` in some (mostly called) places the check for `isControlHuman()` was added to optimize CPU usage. But later some of the remaining places were forgotten. :(

src/fheroes2/heroes/heroes.cpp
Districh-ru: Yes, the AI was forgotten here because all tests were made only for human player to check the radar render. :(
`Scout( this->GetIndex() );` should be called also for AI player, while `ScoutRadar();` is needed only for human player.

Thanks for finding and fixing this issue!

src/fheroes2/heroes/route.h
```diff
@@ -124,8 +143,31 @@ namespace Route
...
         }
 
-        void Reset();
-        void PopFront();
+        void Reset()
+        {
+            clear();
+        }
+
+        // Truncates the path after the current step (the first in the queue). Is the queue is empty, then does nothing.
```
Districh-ru, line 151:
```suggestion
        // Truncates the path after the current step (the first in the queue). If the queue is empty, then does nothing.
```

Districh-ru (approved):
I played a bit and tested objects interaction, increase of hero's scout area (Scouting skill upgrade, capturing Telescope), board boat and use of Dimension Door and Town Gate spells; let AI to play using these actions and found no issues.
Well Done! 👍

src/fheroes2/heroes/heroes.cpp
ihhub: These are unfair labor conditions!

ihhub (approved):

ihhub:
Hi @Branikolog , please test it today.

We are going to delay the release due to these changes.

ihhub:

PR #7553, opened 2023-08-12
AI should decline battles if passing by

### LINKED ISSUE #7503: The AI creates path through pyramid, attacks it and dies.
### Preliminary checks

### Platform

Windows

### Describe the bug

The AI green player avoids the blue player by creating path through the pyramid and attacks it, but it is obviously too weak and dies.

It has a decent enough army:
image

https://github.com/ihhub/fheroes2/assets/12501091/c141cf27-1eca-430c-80aa-3ac373c2087a

There are several issues here, but the core one is that the AI's evaluation of the pyramid needs improvement.

### Save file

[ai suicide pyramid.zip](https://github.com/ihhub/fheroes2/files/12240464/ai.suicide.pyramid.zip)

### Additional info

_No response_

idshibanov:
Fixes #7503 .

Had to expose HeroesValidObject call otherwise AI might choose to go to the Pyramid willingly and then decline the battle.


oleg-derevenetz:
Hi @idshibanov what do you think about making a minimal wrapper method in the `AI::Base`/`AI::Normal` classes that will take only necessary parameters and call the existing `HeroesValidObject()` method that will remain in the anonymous namespace? This would keep the necessary changes minimal.

oleg-derevenetz:
Hi @idshibanov I just integrated my `AIWorldPathfinder`-related changes in this PR, so we can get the best of the two approaches.

idshibanov:
> Hi @idshibanov I just integrated my `AIWorldPathfinder`-related changes into this PR, so we can get the best of the both approaches.

Updated and reverted the ValidObject changes. Should be good to go now.

oleg-derevenetz:
> Updated and reverted the ValidObject changes. Should be good to go now.

Yes, thank you! I just added a bit of minor changes:

* Call the `isValidHeroObject()` using the shortcut logic, i.e. `if (!isValidHeroObject(...)) return;`. Since `HeroesValidObject()` validates not just the army advantage, but also the utility of the object (e.g. whether it contains something useful), I don't see a reason not to use this validation;
* Added the ability to avoid visiting several more types of objects - objects that can offer recruits after fight (e.g. Troll Bridge) and Abandoned Mine;
* I slightly reworked the logging of the actions of the heroes (I wanted to do it for a long time). Previously, logging of an action for some reason was performed at its end, which looked weird, especially if something else was logged during the same action. Now logging of an action is performed at its beginning.

I'll do a couple more tests tomorrow.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1132,7 +1132,7 @@ namespace AI
 
             // This object could have already been visited
             if ( value < 1 ) {
-                return -valueToIgnore;
```
oleg-derevenetz, line 1135: Because of this typo, AI was a big fun of empty water chests, flotsams, etc - it might even prefer them to an empty enemy castle :)

src/fheroes2/ai/ai.h
```diff
@@ -111,6 +111,7 @@ namespace AI
 
         virtual void Reset();
         virtual void resetPathfinder() = 0;
+        virtual bool isValidHeroObject( const Heroes & hero, const int32_t index, const bool underHero = false ) = 0;
```
ihhub, line 114: Please do not use default parameters for virtual functions. You can take a look [here](https://stackoverflow.com/questions/3533589/can-virtual-functions-have-default-parameters) for an explanation.

ihhub (changes requested):
Hi @idshibanov , I left one suggestion here. Could you please take a look?

src/fheroes2/ai/normal/ai_normal_hero.cpp
ihhub: Oops...

ihhub (approved):

ihhub:
@idshibanov and @oleg-derevenetz , thank you so much for the change!

PR #7570, opened 2023-08-19
Limit dimension door usage on easier difficulties

idshibanov:
Could tag it with "AI downgrade" instead. It's time to make "easy" difficulty actually easier.

Still debating if it's worth putting `_dimensionDoorsUsed` in the savegame.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -720,6 +720,7 @@ namespace AI
 
         for ( Heroes * hero : heroes ) {
             hero->ResetModes( Heroes::SLEEPER );
+            hero->setDimensionDoorUsage( 0 );
 
             const double strength = hero->GetArmy().GetStrength();
             _combinedHeroStrength += strength;
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -2135,6 +2135,7 @@ namespace AI
         hero.Scout( targetIndex );
         hero.Move2Dest( targetIndex );
         hero.SpellCasted( dimensionDoor );
+        hero.setDimensionDoorUsage( hero.getDimensionDoorUses() + 1 );
         hero.GetPath().Reset();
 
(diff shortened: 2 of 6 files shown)


src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -720,6 +720,7 @@ namespace AI
 
         for ( Heroes * hero : heroes ) {
             hero->ResetModes( Heroes::SLEEPER );
+            hero->setDimensionDoorUsage( 0 );
```
idshibanov, line 723: If you wish to test this on existing savegames you can add `hero->AppendSpellToBook( Spell::DIMENSIONDOOR, true );` line right here to gift the spell to every AI hero.

oleg-derevenetz:
Is it really that hard to play against an AI that uses DD? :) This spell appears quite late (unless, of course, the author of the map has already set up maximum-level mage guilds in the castles of the AI players).

Alucard648:
Some maps has AI starting with fully built castle and hero with Expert Wisdom to cast lv5 spells. If he gets DD in castle, have fun getting rushed on week 1 with T6 units.

ihhub:
I agree with @oleg-derevenetz that the Dimension Door spell appears very often and this change would affect literally just few maps but won't make the AI easier in general. We can still proceed with such a change but we have to change something else to make AI dumber🙂

idshibanov:
> I agree with @oleg-derevenetz that the Dimension Door spell appears very often and this change would affect literally just few maps but won't make the AI easier in general. We can still proceed with such a change but we have to change something else to make AI dumber🙂

I agree that DD is rare but we gotta start somewhere. That's not the only change that's coming to differentiate between Easy and Expert levels.

oleg-derevenetz (approved):

oleg-derevenetz:
> Still debating if it's worth putting _dimensionDoorsUsed in the savegame.

In general, AI is not able to make save files during its turn, BUT there is "AI auto control mode" in debug builds that can make autosaves, BUT if we load such an autosave, we are able to transfer control to AI only on next turn, so I don't see a reason to put this field to the save file. What are the reasons for putting it?

ihhub (approved):

Branikolog (approved):

ihhub:

PR #7572, opened 2023-08-20
Fix some of the recently introduced code smells

oleg-derevenetz:


ihhub:
Hi @oleg-derevenetz , can you please point to the warning in release build within GitHub Actions? We interpret all warnings as errors by compilers.

oleg-derevenetz:
Hi @ihhub

> We interpret all warnings as errors by compilers.

We do this only in debug builds. In release builds, they are still warnings:

https://github.com/ihhub/fheroes2/actions/runs/5915668448/job/16042143538#step:8:132

Turning warnings into errors in release builds is considered harmful in principle, because release builds are often performed in an environment imposed by package managers, and it is completely unknown what build settings may be there. For example, there may be some additional warnings turned on at the system level that we do not currently have.

ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for code cleanup!

PR #7598, opened 2023-08-21
Add A and O with diaereses to CP1252 button font

zenseii:
Needed for Swedish.
image
image


ihhub (approved):

src/fheroes2/gui/ui_font.cpp
```diff
@@ -5581,9 +5581,18 @@ namespace
         // Change spacing relative to other characters.
         released[192 - 32].setPosition( buttonFontOffset.x - 1, -3 );
         released[193 - 32].setPosition( buttonFontOffset.x - 1, -3 );
+        released[196 - 32].setPosition( buttonFontOffset.x - 2, -2 );
```
Districh-ru, line 5584: The original "A" letter has '-1' horizontal offset, while here "Ä" has '-2'.
https://github.com/ihhub/fheroes2/blob/3a0474c6f70401855e5069c6ff2287545973aba1/src/fheroes2/gui/ui_font.cpp#L4301

If '-2' better fits the 'A' letter then we should change offsets for all "A" letters.

Also for CP1250 there is no horizontal offset for all "A" letters so later we should put there '-1' or '-2'.

Districh-ru (changes requested):
Hi, @zenseii. I left one question. Could you take a look when you have time.

Districh-ru (approved):
👍 
OK, @zenseii, if it will be needed we will update "A" offsets for other code pages.

ihhub:

PR #7616, opened 2023-08-23
Remove duplicated lines to be translated

ihhub:
All these lines exist in the code but without ":" symbol. We are forcing translators to do double work.


ihhub:
I am closing this pull request as #7617 should be fixed first.

PR #7619, opened 2023-08-23
Drop SDL1 support

### LINKED ISSUE #6002: SDL1 Installer and executable doesn't work under Windows XP
### Preliminary checks

### Platform

Windows

### Describe the bug

SDL1 Installer doesn't work under XP, even with modified MajorSubsystemVersion to 5 under Optional Header using CFF Explorer and recalculated checksum. Another error apper then.

When I downloaded SDL1 ZIP package and tried running **fheroes2.exe** it can't be launched. Because it's not valid Win32 executable. But when I mod MajorSubsystemVersion to 5, I can finaly play under Windows XP SP3.

Can you change needed MajorSubsystemVersion in SDL1 compilation to be able to run under Windows XP? No issues detected, game works. I tried it only few minutes with success.

SDL2 ZIP version is not suitable to apply same hack. 

I have no idea where should I look for changing MajorSubsystemVersion from Vista (6.0) to XP (5.1) in published code.

Extracted SDL1 ZIP
image

Extracted SDL1 ZIP with fixed fheroes2.exe 
image

image

Testing was done on Windows XP SP3+ (+ means all updates and additions until 04-2019 when MS released last updated). Of course testing image was build using my XP2ESD project which help users to run XP on modern systems. 

I hope you can change SDL1 to be fully supported under XP from next version. Otherwise I can provide small HEX patch for already compiled EXE.

### Save file

No save file

### Additional info

_No response_

### LINKED ISSUE #6007: SMPEG.DLL in SDL1 have depency to missing LIBSTDC++-6.DLL
### Preliminary checks

### Platform

Windows

### Describe the bug

I have checked all files with Depency Walker and smpeg.dll have missing imports and can cause problems. I found this as real issue when testing fheroes2 under Windows XP SP3 here, but this bug have nothing to do with host as exist all Windows version and builds. This is from Windows 10
https://github.com/ihhub/fheroes2/issues/6002#issue-1415624307

image

image

If I understand correctly, here is solution how to compile it without needs of LIBSTDC++-+.DLL
https://stackoverflow.com/questions/6404636/libstdc-6-dll-not-found

### Save file

No save file

### Additional info

_No response_

### LINKED ISSUE #6084: How to deal with sdl12-compat in Mac app bundle builds?
### Preliminary checks

### Platform

macOS

### Describe the bug

The macOS Homebrew project removed SDL 1.2 not too long ago in favor of sdl12-compat, which makes sense since it's actively maintained, and recommended as a drop-in replacement for software which uses SDL 1.2.  However, this introduces some unique challenges for SDL 1.2 projects which use dylibbundler, which I did when adding Mac app bundle support to this project in order to allow for distribution of the app bundle without requiring users have libraries installed in the same locations on their systems.

In particular, the sdl12-compat library is built rather strangely due to how it operates, and how it's built with some of the original SDL 1.2 header files.  The end result here is that it doesn't list the SDL 2.0 library as a dependency, so dylibbundler never bundles the SDL 2.0 library into the app bundle, and it fails to launch with an error saying that SDL 2.0 can't be found (as that library is a dependency; it's loaded via internal C code in the library rather than marking the SDL 2.0 library as a dependency and letting the OS load it).

So, the end result here is that if somebody tries to run an SDL 1.2 Mac app bundle build that was built with sdl12-compat instead of true SDL 1.2, the app bundle will fail to run, even though it will appear to build correctly.

What I'm hoping to do here is open some discussion around what would be the best way of resolving this with the fheroes2 project.  I've tried talking to the SDL folks, but it's gone nowhere, so while I have solutions myself, there doesn't appear to be an officially endorsed solution.

As I see it, there are really two options:
1. Hard fail on SDL 1.2 Mac app bundle builds if they have sdl12-compat instead of true SDL 1.2.  This would be the case for any SDL 1.2 library with a patch version of 50 or above (so, 1.2.50 or above).
2. Detect sdl12-compat and copy the SDL 2.0 library to the same location within the Mac app bundle as the rest of the libraries as a separate (non-dylibbundler) step.

Option 2 would be more user-friendly while also being very doable in CMake (I already have a working implementation in another project at https://github.com/Kreeblah/MazeOfGalious, which relies on getting the path to the SDL 2.0 library from a modified FindSDL2.cmake file at https://github.com/Kreeblah/MazeOfGalious/blob/main/cmake/FindSDL2.cmake), though the Makefile-based builds could be challenging.  I'm not sure how to get the SDL 1.2 patch version out of that process, or how to determine the path to the SDL 2.0 library on the filesystem there.

Would the maintainers here have any preferences on how to go about solving this?  I'm happy to start work on it, but I was hoping to get thoughts on it first.

### Save file

N/A

### Additional info

_No response_

oleg-derevenetz:
close #6002
close #6007
close #6084

SDL1 is outdated and has not been developed for ~10 years, and now SDL3 is on the way. It's time to finally abandon SDL1.


CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ set(USE_SDL_VERSION SDL2 CACHE STRING "Version of the SDL library used")
 set(FHEROES2_DATA ${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME} CACHE STRING "Path to the fheroes2 data directory (relative to CMAKE_INSTALL_PREFIX)")
 
 # For ComboBox in CMake GUI
-set_property(CACHE USE_SDL_VERSION PROPERTY STRINGS SDL SDL2)
+set_property(CACHE USE_SDL_VERSION PROPERTY STRINGS SDL2)
```
oleg-derevenetz, line 53: Didn't delete this completely in anticipation of the upcoming SDL3.

ihhub (approved):

ihhub:
@oleg-derevenetz , well done! I will update runner requirements later today once I'm back home.

PR #7624, opened 2023-08-24
Show ... in Load/Save File dialog for long names

### LINKED ISSUE #7497: load saved game, saved game selection dialog: not enough space for displaying game names
### Preliminary checks

Some saved game names are shortened so that they appear identical.
shortened

I would suggest widening the game selection dialog window 1.5 times.

### Additional info

_No response_

ihhub:
relates to #7497

Also small rendering optimization for text in multiple places to avoid extra memory copies.

image

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -30,6 +30,7 @@
 #include <memory>
 #include <sstream>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "agg_image.h"
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -139,38 +139,40 @@ void FileInfoListBox::RedrawItem( const Maps::FileInfo & info, int32_t dstx, int
 {
     char shortDate[20];
     char shortHours[20];
-    char shortTime[20];
+    char shortMinutes[20];
 
     const tm tmi = System::GetTM( info.timestamp );
 
     std::fill( shortDate, std::end( shortDate ), static_cast<char>( 0 ) );
     std::fill( shortHours, std::end( shortHours ), static_cast<char>( 0 ) );
-    std::fill( shortTime, std::end( shortTime ), static_cast<char>( 0 ) );
+    std::fill( shortMinutes, std::end( shortMinutes ), static_cast<char>( 0 ) );
     std::strftime( shortDate, ARRAY_COUNT( shortDate ) - 1, "%b %d,", &tmi );
     std::strftime( shortHours, ARRAY_COUNT( shortHours ) - 1, "%H", &tmi );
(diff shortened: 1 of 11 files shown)


src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -139,38 +140,40 @@ void FileInfoListBox::RedrawItem( const Maps::FileInfo & info, int32_t dstx, int
...
             savname.erase( dotPos );
 
-        text.Set( savname, ( current ? Font::YELLOW_BIG : Font::BIG ) );
-        text.Blit( dstx + 5, dsty, 150 );
+        const fheroes2::FontType font = current ? fheroes2::FontType::normalYellow() : fheroes2::FontType::normalWhite();
+        fheroes2::Display & display = fheroes2::Display::instance();
+
+        fheroes2::Text text{ std::move( savname ), font };
+        text.fitToOneRow( 150 );
+        text.draw( dstx + 5, dsty + 2, display );
```
Districh-ru, line 167: We can slightly optimize the code.
`dsty + 2` is used 4 times. We can cache it or simply do `dst += 2;` because `dsty` is not `const`.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -139,38 +140,40 @@ void FileInfoListBox::RedrawItem( const Maps::FileInfo & info, int32_t dstx, int
...
 
-        text.Set( shortDate, ( current ? Font::YELLOW_BIG : Font::BIG ) );
-        text.Blit( dstx + 225 - text.w(), dsty );
+        text.set( shortDate, font );
+        text.draw( dstx + 225 - text.width(), dsty + 2, display );
 
-        text.Set( shortHours, ( current ? Font::YELLOW_BIG : Font::BIG ) );
-        text.Blit( dstx + 245 - text.w(), dsty );
+        text.set( shortHours, font );
+        text.draw( dstx + 245 - text.width(), dsty + 2, display );
```
Districh-ru, line 173: Here we can also cache `dstx + 245` or do `dstx += 245;` to use it in the next `text.draw( dstx, ... )`.

Districh-ru (changes requested):
Hi, @ihhub, I left two comments regarding minor computational optimizations. Could you please take a look when you have time.

Also, IMHO, this PR will not close the mentioned in #7497 issue: "_Some saved game names are shortened so that they appear identical._"
изображение

A solution would be to use **StandardWindow** class and make a new, wider (and possibly higher) dialog. We could redesign this dialog in one of the next PRs, e.g like we did for "Select ..." dialog in Battle Only. What do you think? :)

Branikolog (approved):

src/fheroes2/dialog/dialog_selectfile.cpp
ihhub: While we usually do this I think one CPU tick is faster than caching a whole variable :)

Districh-ru (approved):

PR #7639, opened 2023-08-25
Cancel Hut of Magi reveal area animation when a player clicks any button

ihhub:
If a map contains a lot of eyes of magi it would take ages to show all of them. This is a solution to avoid such.

Try this save file on this branch and on master:
[Hut_of_Magi.zip](https://github.com/ihhub/fheroes2/files/12441063/Hut_of_Magi.zip)

### DIFF
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -3291,15 +3291,14 @@ namespace
                     Maps::ClearFog( eyeIndex, scoutRange, hero.GetColor() );
 
                     const fheroes2::Point eyePosition = Maps::GetPoint( eyeIndex );
-
-                    I.getGameArea().SetCenter( eyePosition );
-
                     const fheroes2::Rect eyeRoi( eyePosition.x - scoutRange, eyePosition.y - scoutRange, 2 * scoutRange + 1, 2 * scoutRange + 1 );
 
                     if ( skipAnimation ) {
                         clearRadarArea = fheroes2::getBoundaryRect( clearRadarArea, eyeRoi );
                         continue;
                     }
+
+                    I.getGameArea().SetCenter( eyePosition );
 
                     I.getRadar().SetRenderArea( eyeRoi );
                     I.redraw( Interface::REDRAW_GAMEAREA | Interface::REDRAW_RADAR );
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -3281,13 +3281,12 @@ namespace
 
                 fheroes2::Display & display = fheroes2::Display::instance();
 
-                const size_t fullTimeViewCount = 3;
-                size_t viewCount = 0;
                 size_t maxDelay = 7;
 
-                for ( const int32_t eyeIndex : eyeMagiIndexes ) {
-                    const int32_t scoutRange = static_cast<int32_t>( GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES ) );
+                const int32_t scoutRange = static_cast<int32_t>( GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES ) );
+                bool skipAnimation = false;
 
+                for ( const int32_t eyeIndex : eyeMagiIndexes ) {
                     Maps::ClearFog( eyeIndex, scoutRange, hero.GetColor() );
 
                     const fheroes2::Point eyePosition = Maps::GetPoint( eyeIndex );
@@ -3299,12 +3298,21 @@ namespace
                     I.getRadar().SetRenderArea( eyeRoi );
                     I.redraw( Interface::REDRAW_GAMEAREA | Interface::REDRAW_RADAR );
 
+                    if ( skipAnimation ) {
+                        continue;
+                    }
+
                     display.render();
 
                     LocalEvent & le = LocalEvent::Get();
                     size_t delay = 0;
 
                     while ( delay < maxDelay && le.HandleEvents( Game::isDelayNeeded( { Game::MAPS_DELAY } ) ) ) {
+                        if ( le.KeyPress() || le.MouseClickLeft() || le.MouseClickMiddle() || le.MouseClickRight() ) {
+                            skipAnimation = true;
+                            break;
+                        }
+
                         if ( Game::validateAnimationDelay( Game::MAPS_DELAY ) ) {
                             ++delay;
                             Game::updateAdventureMapAnimationIndex();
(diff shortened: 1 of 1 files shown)


src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -3281,6 +3281,10 @@ namespace
 
                 fheroes2::Display & display = fheroes2::Display::instance();
 
+                const size_t fullTimeViewCount = 3;
+                size_t viewCount = 0;
+                size_t maxDelay = 7;
+
                 for ( const int32_t eyeIndex : eyeMagiIndexes ) {
                     const int32_t scoutRange = static_cast<int32_t>( GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES ) );
```
oleg-derevenetz, line 3289: I suppose we can move this out of the loop since scouting range doesn't depend on eye position?

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -3309,6 +3313,11 @@ namespace
                             display.render();
                         }
                     }
+
+                    ++viewCount;
+                    if ( viewCount > fullTimeViewCount && maxDelay > 0 ) {
+                        --maxDelay;
```
oleg-derevenetz, line 3319: It seems that after 7 eyes max delay becomes zero and player will not see the animation for remaining eyes at all, is this intentional?

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion and one question, could you please take a look when you have time?

LeHerosInconnu:
Related discussion: https://github.com/ihhub/fheroes2/discussions/7105.

Branikolog:
Hi, @ihhub 
I personally don't like the way how it's working in this PR. After 6 or 7 eye the revealing speed rises exponentially and the last areas are revealing too fast, appearing for milliseconds. Implementation looks not smooth and raw. I'd rather prefer to speed up for 0.5-1sec the whole revealing (optionally, as I'm ok with current speed) and allow player to skip by pressing any game button. This could be a click for each eye or even skipping the whole process at once.

ihhub:
Hi @Branikolog , please be very specific how exactly you see the implementation to be done. Also take a note about touchscreen devices.

oleg-derevenetz:
Hi @ihhub I recommend not to change the animation speed, but just to cancel the whole animation on key press or mouse button press.

Branikolog:
@ihhub 

I'm personally think that both implementations are fine. Probably the one, where we just need to make a single left click or press any button on keyboard to skip all animations at once, brings much more convenience, as @zenseii noticed, we can always check minimap for new areas to explore them more thoroughly.

LeHerosInconnu:
Hello @everyone,

> Hi @ihhub I recommend not to change the animation speed, but just to cancel the whole animation on key press or mouse button press.

Yes.
I stand by my proposal (https://github.com/ihhub/fheroes2/discussions/7105) to completely stop all animation(s) after a mouse click or press of any keyboard key.

The solution of stopping only one animation at a time may seem attractive, but would quickly become annoying, as the player would have to perform several actions in succession to stop all animations, which beyond three actions would be criminal for mouse survival or keyboard life, as the player would then frantically click the mouse button or hit the keyboard wildly to skip animations. :) :D

ihhub:
Based on the majority of people's opinion I am going to implement animation interruption and immediate reveal of the rest of eyes if a user clicks any button.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -3295,12 +3298,21 @@ namespace
                     I.getRadar().SetRenderArea( eyeRoi );
                     I.redraw( Interface::REDRAW_GAMEAREA | Interface::REDRAW_RADAR );
 
+                    if ( skipAnimation ) {
+                        continue;
+                    }
```
Districh-ru, line 3303: If we skip viewing the rest "Eyes" then we need only to redraw the radar part without gamearea redraw. Gamearea will be rendered after we finish this loop at line 3327.
So we can move `I.getGameArea().SetCenter( eyePosition );` (from line 3294) after the check for `skipAnimation` and if `skipAnimation` is `true` do only `REDRAW_RADAR`:

```suggestion
                    if ( skipAnimation ) {
                        I.redraw( Interface::REDRAW_RADAR );
                        continue;
                    }

                    I.getGameArea().SetCenter( eyePosition );
                    I.redraw( Interface::REDRAW_GAMEAREA | Interface::REDRAW_RADAR );
```

Districh-ru (changes requested):
Hi, @ihhub, I like the implementation, but I noticed a delay after pressing a key to skip the rest of the "Eyes". I left one suggestion for rendering optimization in this case. Could you please take a look when you have time.

ihhub:
Hi @Districh-ru , I made an extra optimization. Please take a look.

Districh-ru (changes requested):
Hi, @ihhub, The optimization with `clearRadarArea` is a good idea. 👍 

Also line 3295 can be moved to the line 3303 becauser if we skip veiwing the rest Eyes we don't need to center the view on them.
https://github.com/ihhub/fheroes2/blob/fd4b357aa500e7ce73a3ef9ba55136bf8809d565/src/fheroes2/heroes/heroes_action.cpp#L3295

Districh-ru (approved):

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -3281,26 +3281,39 @@ namespace
 
                 fheroes2::Display & display = fheroes2::Display::instance();
 
-                for ( const int32_t eyeIndex : eyeMagiIndexes ) {
-                    const int32_t scoutRange = static_cast<int32_t>( GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES ) );
+                const size_t maxDelay = 7;
+
+                const int32_t scoutRange = static_cast<int32_t>( GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES ) );
+                bool skipAnimation = false;
+                fheroes2::Rect clearRadarArea;
```
oleg-derevenetz, line 3288: Let me join this nitpickers club and notice that `clearRadarArea` name is a bit misleading, because this is not exactly the area that will be cleared on the radar (in the sense that the fog will be removed in this area). I would call it something like `radarRenderArea`.

oleg-derevenetz (approved):
Hi @ihhub I just left one optional suggestion regarding variable naming. LGTM otherwise.

Branikolog (approved):

src/fheroes2/heroes/heroes_action.cpp
ihhub: Haha, I will rename it.

PR #7640, opened 2023-08-26
Add small skill icons for Xanadu visit dialog

### LINKED ISSUE #5242: Show the 4 primary skill icons +1 for xanadu visit
### Preliminary checks

Currently when visiting the xanadu and you are granted the primary skill bonus you only get a text which is not in line with the style of the rest of the game because there's no visual queue showing what you got.

image

Maybe a stacked primary skill icon sort of like mass haste and a +1 next to it?

### Additional info

Visit the Xanadu and have a hero at least level 10 minus level of diplomacy.

ihhub:
close #5242

The icons are right clickable.
image

You can test the change using this save file:
[xanadu.zip](https://github.com/ihhub/fheroes2/files/12444369/xanadu.zip)

### DIFF
--- a/src/fheroes2/gui/ui_dialog.cpp
+++ b/src/fheroes2/gui/ui_dialog.cpp
@@ -635,7 +635,7 @@ namespace fheroes2
 
     void SmallPrimarySkillDialogElement::draw( Image & output, const Point & offset ) const
     {
-        const fheroes2::Sprite & originalImage = fheroes2::AGG::GetICN( ICN::SWAPWIN, 0 );
+        const Sprite & originalImage = AGG::GetICN( ICN::SWAPWIN, 0 );
 
         switch ( _skillType ) {
         case Skill::Primary::ATTACK:
--- a/src/fheroes2/gui/ui_dialog.cpp
+++ b/src/fheroes2/gui/ui_dialog.cpp
@@ -271,13 +271,7 @@ namespace fheroes2
         assert( 0 );
     }
 
-    CustomImageDialogElement::CustomImageDialogElement( const Image & image )
-        : _image( image )
-    {
-        _area = { _image.width(), _image.height() };
-    }
-
-    CustomImageDialogElement::CustomImageDialogElement( Image && image )
+    CustomImageDialogElement::CustomImageDialogElement( Image image )
         : _image( std::move( image ) )
     {
         _area = { _image.width(), _image.height() };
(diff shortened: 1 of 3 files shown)


Branikolog:
Hi, @ihhub 

There's too much free space in this window.
Could you, please, explain, why have you placed icons in a square? Why not in a row?

xanadu

Branikolog:
Hi, @zenseii !

It's not a problems in my head. :)

>It might confuse players into thinking that they can be selected. The arena and the Xanadu function differently and should be distinguished more, in my opinion.

Once player click on these icons and nothing happens - I believe everyone understands the way this objects operates.
Also, here we can see "+1", while in the arena window icons have just name of the skill below, which is quite easy to distinguish in my opinion. 

>It looks like we are just copying and reusing the look of the arena, which could leave an impression of lacking creativity and just copy pasting the same thing several places in the game.

For me it on contrary looks like a consistency in design and respecting the style of the game UI. 

I'm not strictly against current implementation and I just got curious, once I've seen this screenshot. The window seemed a little unusual to my eye.
Thanks for your answer, @zenseii .

If everyone is ok with this implementation I cannot insist on changing anything. :)

LeHerosInconnu:
Hello @everyone,

I stand by my position (https://github.com/ihhub/fheroes2/discussions/4877) and continue to think that this arrangement of elements best suits the game:

> xanadu

The large font could eventually be used instead of the small font as there's enough room for it (we'd have to see the visual with the "+1" in large font), especially when screen resolutions larger than the original game's are used.

ihhub:
Okay. I will put all items in one row.

ihhub:
Hi @LeHerosInconnu , using normal font feels unnatural:
image

So I kept small font for icons.

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -654,6 +623,45 @@ namespace fheroes2
...
+            fheroes2::Copy( originalImage, 216, 51, output, offset.x, offset.y, _iconSize.width, _iconSize.height );
+            break;
+        case Skill::Primary::DEFENSE:
+            fheroes2::Copy( originalImage, 216, 84, output, offset.x, offset.y, _iconSize.width, _iconSize.height );
+            break;
+        case Skill::Primary::POWER:
+            fheroes2::Copy( originalImage, 216, 117, output, offset.x, offset.y, _iconSize.width, _iconSize.height );
+            break;
+        case Skill::Primary::KNOWLEDGE:
+            fheroes2::Copy( originalImage, 216, 150, output, offset.x, offset.y, _iconSize.width, _iconSize.height );
```
Districh-ru, line 651: We can remove `fheroes2::` because we are in fheroes2 namespace. :)

Districh-ru (changes requested):
Hi, @ihhub. I like how Xanadu dialog looks now, well done!
I left one one minor comment. Could you check it when you have time.
Also it seems that currently `FrameDialogElement` class is never used. :(

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu , using normal font feels unnatural: image
> 
> So I kept small font for icons.

Effectively, thanks for the visual with the big font.

oleg-derevenetz:
Hi @ihhub this PR introduced (a kind of) regression - there are new object slicing warnings from the SonarCloud, because `Sprite` instances are passed not by reference now, but by value as `Image` instances. I believe that slicing is harmless in this case, but there are still "errors" from the SonarCloud's point of view.

ihhub:
@oleg-derevenetz , oh. I should revert changes then😞

Issue #6681, opened 2023-02-20
Make lags more smooth


shprotru:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Currently game menu have few places with hangs for some seconds:

- [ ] if you put a lot of maps in maps directory, then click "standard game" from the "new game" menu
- [x] if you click "Language" in "options" menu

### Describe the possible solution

I suggest showing the new window, on top of all windows, with an `ICN::HOURGLAS` animated icon without any progress bar. This will be more common to any user expirience.

Branikolog:
What about just changing game cursor to sundial in such cases? It's a common feature for PC OSes, when cursor changes to "loading" state while heavy process.

shprotru:
Yes, may be that will be even better for tiny interfaces, but you will still see an interface on background, it should be faded, to make accent that something going. Preloading window could be a base for future network actions in engine like "online authorization", how to deal just with one cursor if happened an network error? The user will already be prepared for where to view the information if he see that preloading window in other situations like I described above. For now you can't see in UI: have you broken map files or not, just only in system logs(stdout/stderror streams) you can see that information. In a preloading window we will have possibility to show some additional information for user, or give him an cancelling control. For now it can be hard to make in one PR, because I don't hear anything about "programming contexts" in c++, while I know how it should be about that from golang.

oleg-derevenetz:
> For now it can be hard to make in one PR, because I don't hear anything about "programming contexts" in c++, while I know how it should be about that from golang.

In C++, this kind of functionality is usually provided by frameworks/libraries that are used to implement asynchronous networking operations, such as Asio.

shprotru:
that's the huge problem of c++ to me: there are always too many ways to do the same thing. By other words: lots of frameworks implement too much of same things. Time passed and accepted new standarts without standard features needed every developer. Found useful [example](https://stackoverflow.com/questions/68051489/context-select-like-features-in-c).

How the cancel button should behave:

1. it should be always visible, with possibility to player click it in any time
2. it should be disabled for timeout, then enabled and give player possibility to click it

it should be regular rectangle button, or something like small square buttin with [x] sign in right upper corner?

ihhub:
Hi @shprotru , I don't think we should discuss C++ in general here. Every language has its own purpose.

oleg-derevenetz:
[...]

PR #7642, opened 2023-08-26
Speed up map, translation and save files loading

ihhub:
relates to #6681

The slowest part - file reading from a storage, remains unchanged.

### DIFF
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -338,15 +338,17 @@ void StreamBuf::reallocbuf( size_t size )
 
 void StreamBuf::put8( const uint8_t v )
 {
-    if ( sizep() == 0 ) {
+    if ( sizep() < 1 ) {
         reallocbuf( capacity() + capacity() / 2 );
-        assert( itput != nullptr );
     }
 
-    if ( sizep() > 0 ) {
-        *itput = v;
-        ++itput;
+    if ( sizep() < 1 ) {
+        assert( 0 );
+        return;
     }
+
+    *itput = v;
+    ++itput;
 }
 
 uint8_t StreamBuf::get8()
@@ -450,8 +452,10 @@ void StreamBuf::putRaw( const char * ptr, size_t sz )
         }
     }
 
-    // Make sure that the possible previous memory reallocation was correct.
-    assert( sizep() >= sz );
+    if ( sizep() < sz ) {
+        assert( 0 );
+        return;
+    }
 
     memcpy( itput, ptr, sz );
     itput = itput + sz;
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -455,6 +455,15 @@ std::string StreamBuf::toString( size_t sz )
     return std::string( it1, it2 );
 }
 
+std::string_view StreamBuf::toStringView( size_t sz )
+{
+    uint8_t * it1 = itget;
+    uint8_t * it2 = itget + ( sz ? sz : sizeg() );
+    it2 = std::find( it1, it2, 0 );
+    itget = it1 + ( sz ? sz : sizeg() );
+    return std::string_view( reinterpret_cast<const char *>( it1 ), it2 - it1 );
+}
+
 void StreamBuf::skip( size_t sz )
 {
     itget += sz <= sizeg() ? sz : sizeg();
@@ -632,7 +641,7 @@ StreamBuf StreamFile::toStreamBuf( const size_t size )
     return buffer;
 }
 
-std::string StreamFile::toString( size_t sz )
+std::string StreamFile::toString( const size_t sz )
 {
     const std::vector<uint8_t> buf = getRaw( sz );
     std::vector<uint8_t>::const_iterator itend = std::find( buf.begin(), buf.end(), 0 );
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -446,22 +446,16 @@ void StreamBuf::putRaw( const char * ptr, size_t sz )
     itput = itput + sz;
 }
 
-std::string StreamBuf::toString( size_t sz )
+std::string StreamBuf::toString( const size_t size )
 {
-    uint8_t * it1 = itget;
-    uint8_t * it2 = itget + ( sz ? sz : sizeg() );
-    it2 = std::find( it1, it2, 0 );
-    itget = it1 + ( sz ? sz : sizeg() );
-    return std::string( it1, it2 );
-}
+    const size_t length = ( size > 0 && size < sizeg() ) ? size : sizeg();
 
-std::string_view StreamBuf::toStringView( const size_t sz )
-{
     uint8_t * it1 = itget;
-    uint8_t * it2 = itget + ( sz ? sz : sizeg() );
+    uint8_t * it2 = itget + length;
     it2 = std::find( it1, it2, 0 );
-    itget = it1 + ( sz ? sz : sizeg() );
-    return { reinterpret_cast<const char *>( it1 ), static_cast<size_t>( it2 - it1 ) };
+    itget = it1 + length;
+
+    return { it1, it2 };
 }
 
 void StreamBuf::skip( size_t sz )
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -21,6 +21,8 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "serialize.h"
+
 #include <algorithm>
 #include <cassert>
 #include <cstring>
@@ -28,7 +30,6 @@
 
 #include "endian_h2.h"
 #include "logging.h"
-#include "serialize.h"
 
 namespace
 {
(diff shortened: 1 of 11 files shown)


ihhub:
Hi @shprotru, could you please test this pull request for performance improvement? I believe it will be slightly faster than master branch.

shprotru:
Hi @ihhub ! I tested from the provided branch `file_reading_speedup`, and there is how it looks like. As fo me as user, there is no changes and technically there are couldn't be, because of speed of the hdd. And my own hdd is enough modern and advanced. This problem as I have been described could be only prettified with preloading window with animation.

https://github.com/ihhub/fheroes2/assets/53114202/11e6dcc5-f85b-4850-8346-b56c37b95788

ihhub:
Hi @shprotru , is Language selection screen remains the same in terms of loading?

shprotru:
the language selection window is opening with acceptable speed, but it's laggy a little bit

https://github.com/ihhub/fheroes2/assets/53114202/9185c231-55c4-4fb3-8545-b6be9cb1e8a3

ihhub:
> the language selection window is opening with acceptable speed, but it's laggy a little bit
> 
Is it faster than on master branch? Also is your storage HDD or SSD? For my laptop with SSD it's almost instant.

shprotru:
There are no noticeable differences on the eye between master branch and this one. My system is installed on SSD, and my `/home` directory is mounted to HDD, where are fheroes2 and maps installed, so answer is HDD.

src/engine/translations.cpp
```diff
@@ -146,15 +147,8 @@ namespace
...
-        uint32_t length;
-
-        chunk()
-            : offset( 0 )
-            , length( 0 )
-        {
-            // Do nothing.
-        }
+        uint32_t offset{ 0 };
+        uint32_t length{ 0 };
```
oleg-derevenetz, line 151: These initializations are redundant here, because there is no default constructor for this `struct` anymore, and existing constructor initializes both fields:

https://godbolt.org/z/7nc1EEz1K

But in general it's harmless, they just will not be used :)

src/engine/translations.cpp
```diff
@@ -183,7 +177,7 @@ namespace
         return ~crc;
     }
 
-    std::string getTag( const std::string & str, const std::string & tag, const std::string & sep )
+    std::string getTag( const std::string_view & str, const std::string & tag, const std::string & sep )
```
oleg-derevenetz, line 180: There is no need to pass the `string_view` by reference.

src/engine/translations.cpp
```diff
@@ -255,71 +233,73 @@ namespace
...
                 }
 
                 const uint32_t offset1 = buf.get32();
                 buf.seek( offset1 );
-                const std::string msg1 = buf.toString( length1 );
+                const std::string_view msg1 = buf.toStringView( length1 );
 
-                const uint32_t crc = crc32b( msg1.c_str() );
-                buf.seek( offset_strings2 + index * 8 /* length, offset */ );
+                const uint32_t crc = crc32b( msg1.data() );
```
oleg-derevenetz, line 301: Please note that `std::string_view` is not designed to be used this way, because it doesn't have to be null-terminated. For example, if `buf` contains something like "AABB", and `length1` is 1, you will get the full "AABB" string here, not just "A":

https://godbolt.org/z/Yvj9rePj4

Need to be careful when transforming `std::string_view` to a C string, you can't just do it "head-on".

oleg-derevenetz (changes requested):
Hi @ihhub I just left some questions and suggestions here. In general, the main reason for the pause is I/O, the impact of these changes is negligible. We can only "prettify" this pause by showing some "please wait" dialog or maybe changing the mouse cursor to the "wait" variant, but I'm still not sure if it's really necessary.

src/engine/serialize.cpp
```diff
@@ -470,14 +455,18 @@ std::string StreamBuf::toString( size_t sz )
     return std::string( it1, it2 );
 }
 
-void StreamBuf::skip( size_t sz )
+std::string_view StreamBuf::toStringView( const size_t sz )
 {
-    itget += sz <= sizeg() ? sz : sizeg();
+    uint8_t * it1 = itget;
+    uint8_t * it2 = itget + ( sz ? sz : sizeg() );
```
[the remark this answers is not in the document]
ihhub: This is valid point and I copy-pasted the code.

src/engine/serialize.cpp
```diff
@@ -470,14 +455,18 @@ std::string StreamBuf::toString( size_t sz )
...
+    it2 = std::find( it1, it2, 0 );
+    itget = it1 + ( sz ? sz : sizeg() );
```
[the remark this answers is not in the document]
ihhub: Mmmm... another copy-pasted code :(

src/engine/translations.cpp
ihhub: Oh, I think I should revert changes related to string_view (or some dangerous ones). Thank you so much for the explanation!

oleg-derevenetz: Also `crc32()` can be modified to accept the `string_view` instead of `char*`.

src/engine/serialize.cpp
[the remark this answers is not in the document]
ihhub: There are places where we read from MP2 files and they have fixed sizes for elements but symbol '\0' can appear earlier. I will keep the same logic as before but beautified it a bit.

Districh-ru (changes requested):
Hi, @ihhub, I left two suggestions. Could you please check if they are valuable in your spare time

src/engine/serialize.h
```diff
@@ -248,12 +248,38 @@ class StreamBuf : public StreamBase
...
+    }
+
+    void advance( const size_t size )
+    {
+        itput += size;
+    }
+
+    size_t size() const
+    {
+        return sizeg();
```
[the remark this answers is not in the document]
ihhub: These are virtual functions. They cannot be inlined.

Districh-ru (approved):
👍 
I loaded several saves and started new maps, changed translations and found no issues.
On my hardware load of translations file is now about 7% faster.
<details>
Load of MO files in microseconds:

new | old | %
-- | -- | --
1916.2 | 2081.3 | 8.616011
1617 | 1843.2 | 13.98887
1842.9 | 1890 | 2.555755
1794.3 | 1889.5 | 5.30569
605.1 | 666.9 | 10.21319
920.1 | 997.3 | 8.390392
983.1 | 1098.1 | 11.69769
312.3 | 321.2 | 2.849824
1818.4 | 1894.2 | 4.1685
519.1 | 535.8 | 3.217107
339.3 | 351.8 | 3.684055
598.1 | 625 | 4.497576
1718.8 | 1883 | 9.553177
332.7 | 327.4 | -1.59303
185.6 | 197.2 | 6.25
1050 | 1179.5 | 12.33333
1353.4 | 1450.8 | 7.19669
167.5 | 182.4 | 8.895522
197 | 207.8 | 5.482234
594.5 | 650.4 | 9.40286

</details>

src/engine/translations.cpp
```diff
@@ -185,13 +178,14 @@ namespace
 
     std::string getTag( const std::string & str, const std::string & tag, const std::string & sep )
     {
-        std::string res;
-        if ( str.size() > tag.size() && tag == str.substr( 0, tag.size() ) ) {
-            size_t pos = str.find( sep );
-            if ( pos != std::string::npos )
-                res = str.substr( pos + sep.size() );
+        if ( str.size() > tag.size() && str.rfind( tag, 0 ) == 0 ) {
```
oleg-derevenetz, line 181: In the `master` branch, there is a check that `str` starts with `tag`. Here the original check is replaced by the check that the `tag` could only be at the beginning of `str`. These checks are not equivalent, because if `str` is `tagtag`, then the check in `master` branch will return `true`, while this check will return `false`. Is this intentional?

oleg-derevenetz (commented):
Hi @ihhub I just left one notice and one question, and also added my own commit with changes related to detection of an error-prone behavior (`StreamBuf::reallocbuf()` should never be called for StreamBufs created from read-only memory blocks such as `const char *` or `const std::vector`) and minor optimizations (there is no need to `std::fill()` with zeroes when we just may [value-initialize](https://en.cppreference.com/w/cpp/language/value_initialization) the array). Could you please take a look when you have time?

src/engine/translations.cpp
ihhub: Hi @oleg-derevenetz , [rfind()](https://en.cppreference.com/w/cpp/string/basic_string/rfind) searches from right to left (from end to start) from the given position `(thus, the found substring, if any, cannot begin in a position following pos)`. If the position is 0 then it will search only at the start of the string and never proceeds further (see [this](https://stackoverflow.com/questions/1878001/how-do-i-check-if-a-c-stdstring-starts-with-a-certain-string-and-convert-a)).

However, I can just revert the change to avoid any confusion. It is not worth big debates :)

src/engine/serialize.cpp
```diff
@@ -461,25 +446,22 @@ void StreamBuf::putRaw( const char * ptr, size_t sz )
...
+
     uint8_t * it1 = itget;
-    uint8_t * it2 = itget + ( sz ? sz : sizeg() );
-    it2 = std::find( it1, it2, 0 );
-    itget = it1 + ( sz ? sz : sizeg() );
-    return std::string( it1, it2 );
+    itget += length;
+    uint8_t * it2 = std::find( it1, itget, 0 );
+
+    return { it1, it2 };
```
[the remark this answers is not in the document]
ihhub: See these lines:
```cpp
name = dataStream.toString( 13 );
...
std::string answer = dataStream.toString( 13 );
```
We expect to have 13 bytes no matter what but the string can be shorter. I agree that the logic is not obvious and I can create 2 overloads of this function with and without input argument. What do you think?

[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub I think that we can leave it as is so far.

src/engine/translations.cpp
oleg-derevenetz: @ihhub Ah, yes, we are searching from the position 0. Let's leave it as is.

PR #7647, opened 2023-08-27
Fix high FPS while right clicking on items in lists

### LINKED ISSUE #6518: Mouse right click on any item which does not have an overloaded ActionListPressRight() method leads to high FPS rate
### Preliminary checks

### Platform

Windows

### Describe the bug

- enable system info
- go to language or resolution dialog
- right click on any of the element
- see how FPS skyrocketing while you don't do any actions

### Save file

No need

### Additional info

_No response_

ihhub:
close #6518

We shouldn't consider a right mouse click as an event for full list redraw as this event serves as an information action.

**IMPORTANT**
This change requires heavy testing of all lists in the game as it modifies the behavior for all of them.


Districh-ru (commented):
Hi, @ihhub, I tested save/load, new game, town portal, kingdom overview, heroes, castles and select items lists. All works great, in Select game resolution/language dialogs right clicking does not provoke high FPS.
But I found one minor issue: in Battle Only open Select Monster dialog and right-click on any monster - you can notice that the info dialog is not closed immediately after releasing the right mouse button. This can be fixed in `dialog_armyinfo.cpp` at line 649 by placing this code (taken from the Quick View dialog code):
```cpp
    // Restore background.
    restorer.restore();
    display.render( restorer.rect() );
```

This is not the issue of current PR because you can observe the same behavior when right-clicking any unit on the battlefield. But if you want, you can fix it here. :)

Also in one of the next PRs we should apply the same fix for holding the right mouse button on any empty monster/artifact/captain/secondary skill in castle, hero, kingdom overview or hero meting dialogs as it currently causes high FPS.

src/fheroes2/gui/interface_list.h
```diff
@@ -438,15 +438,15 @@ namespace Interface
                             _currentId = id;
                             ActionListSingleClick( item, mousePos, rtAreaItems.x, rtAreaItems.y + offsetY );
                         }
+                        needRedraw = true;
+
                         return true;
                     }
                     else if ( le.MousePressRight( rtAreaItems ) ) {
                         ActionListPressRight( item, mousePos, rtAreaItems.x, rtAreaItems.y + offsetY );
                         return true;
```
Districh-ru, line 447: After some checks for the cause of high FPS on every right mouse button hold on empty field (castle captain, secondary skill) in Kingdom Overview I found that this return value (`true`) is used to set `redraw` to `true`. In other places it seems that this state is used for the same purposes: to redraw screen or update data after changes.
But holding the right mouse button above the list item should not cause the change of any data. Therefore possibly we should return `false` here. What do you think, @ihhub?

PS High FPS on right mouse press on any empty artifact/unit in heroes meeting, hero and castle screens can be fixed
<details><summary> by changing: </summary>

https://github.com/ihhub/fheroes2/blob/e335ab273dd3e20a3c8551ad9ebdfa2aee9cf417/src/fheroes2/army/army_bar.cpp#L627-L629

and
https://github.com/ihhub/fheroes2/blob/e335ab273dd3e20a3c8551ad9ebdfa2aee9cf417/src/fheroes2/resource/artifact.cpp#L1241-L1243

to:
```cpp

        return true;
    }

    return false;
```

</summary>

ihhub: Hi @Districh-ru , I think we should also return a boolean value from **ActionListPressRight** as well to be consistent with the code.

Districh-ru: Hi, @ihhub, you are right, then `ActionListPressRight()` will return true if redraw is needed and we will return it from `QueueEventProcessing()`.

ihhub:
Closing this pull request as the issue is no longer valid.

PR #7649, opened 2023-08-27
Fix the weird logic of MP2::isWaterActionObject()

### LINKED ISSUE #7648: Вылетает игра
### Preliminary checks

### Platform

Windows

### Describe the bug

Файл с сохранением игры прикрепил.
Нужно загрузить игру и сделать конец хода.
Безымянный

### Save file

[save.zip](https://github.com/ihhub/fheroes2/files/12448380/save.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #7648

Although I'm perfectly sure about the `OBJ_HEROES` case, I'm not so sure about removed cases (barriers, monsters, resources and even other object types in PoL). Are there really any known maps where similar objects are located on water or should be accessed directly from water?

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -501,10 +501,6 @@ int Interface::AdventureMap::GetCursorFocusShipmaster( const Heroes & hero, cons
         break;
     }
 
-    case MP2::OBJ_MONSTER:
-        // TODO: Can monsters be placed on water?
-        return isWater ? Cursor::DistanceThemes( Cursor::CURSOR_HERO_FIGHT, hero.getNumOfTravelDays( tile.GetIndex() ) ) : Cursor::POINTER;
-
     case MP2::OBJ_COAST:
         return Cursor::DistanceThemes( Cursor::CURSOR_HERO_ANCHOR, hero.getNumOfTravelDays( tile.GetIndex() ) );
 
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -501,6 +501,10 @@ int Interface::AdventureMap::GetCursorFocusShipmaster( const Heroes & hero, cons
         break;
     }
 
+    // Some map editors allow to place monsters on water tiles
+    case MP2::OBJ_MONSTER:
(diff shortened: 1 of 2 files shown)


oleg-derevenetz:
Hi @Branikolog thank you! Do you know any well-known maps that exploit OG editor bugs and have Barriers, Monsters, Resources, etc located on water? This PR will break such maps.

Branikolog:
Hi, @oleg-derevenetz !
I've managed to create a map with such objects and assertion occurs indeed.
[Water objects.zip](https://github.com/ihhub/fheroes2/files/12455126/Water.objects.zip)

oleg-derevenetz:
> I've managed to create a map with such objects and assertion occurs indeed.

That's why I'm asking, are we need to support invalid maps of this kind? :)

Branikolog:
Actually, I've seen only several such maps with the "issues".
I think it would be better to support such maps as people will continue sending such assertion save files to us, as the way assertion occurs in the fog is not evident until we open the file in debug build.

oleg-derevenetz:
> I think it would be better to support such maps as people will continue sending such assertion save files to us, as the way assertion occurs in the fog is not evident until we open the file in debug build.

On the other hand, these maps are invalid. For instance, if there is a map editor that produces maps with invalid objects (e.g. a multi-tile object where not all tiles are valid) the game will work incorrectly and some assertions will undoubtedly fail. But this doesn't mean that we need to somehow "fix" everything in engine. Therefore, the question is how widespread it is and whether there are any really popular maps that exploit this editor bug somehow.

Branikolog:
@oleg-derevenetz 

I've seen a streamer playing such map right yesterday (with several stacks of Hydras placed in the ocean). So I'm afraid of constant complains from players regarding this issue. I think we should support such maps (as the original game somehow handles such cases). If you and other members are against we should at least forbid loading such maps at all showing a message of corrupted file.

Districh-ru (approved):
I tested some maps (including second scenario of Voyage Home campaign) with many water and found no issues.

oleg-derevenetz:
Hi @zenseii

> I know there are fan made maps which have hex-edited objects on the sea that shouldn't be there, but I don't think we should worry to much about making adjustments for them to work.

Currently they will work with this PR. At least map from @Branikolog with resources, monsters and objects on water definitely work, both AI and human player are able to interact with these objects.

ihhub (approved):

ihhub:

PR #7661, opened 2023-08-29
Fix a few minor code issues

oleg-derevenetz:
Don't include `<locale>` twice and use `std::from_chars()` instead of fancy `GetInt()` (with UB).


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for the code cleanup!

Issue #1892, opened 2020-10-03
File to Load window, position of elements


LeHerosInconnu:
- The title "File to Load:" is not yellow.
- The title is not centered horizontally with respect to the scenario frame.
- The scenario names are not centered horizontally with respect to the scenario frame.
- The vertical spacing between lines is not regular (19 pixels and 20 pixels from baseline of text to baseline of text).
- The scenario list text block is not centered vertically with respect to the scenario frame.
- The selected scenario is not in yellow within the frame of the selected scenario.
- The selected scenario is not centered vertically within the frame of the selected scenario.
- The selected scenario inside the selected scenario frame should be the same distance from the left edge of the frame as the scenario names in the scenario list.
- The "OKAY" and "CANCEL" buttons are not the same horizontal distance from the edge of the window closest to the button.

In fheroes2:

![File to load window evil fh2 01](https://user-images.githubusercontent.com/43583994/95000108-30779300-05be-11eb-9a9a-027d4c085d71.png)

Issue #4053, opened 2021-08-21
Allow sorting saved games list by date


a1exsh:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Currently we only sort the list of saved games in the "Save Game" and "Load Game" dialogs by file name.  Often it is useful to be able to sort by date (file timestamp) instead.

### Describe the possible solution

Add clickable headers to the file list dialogs: `Name` and `Date`.  Clicking on either of the headers should toggle sorting by that file attribute.

LeHerosInconnu:
Related issue: https://github.com/ihhub/fheroes2/issues/1211.

ihhub:
It was fixed in #9793

PR #7662, opened 2023-08-29
Rework the Save/Load game dialog to allow variable dialog size and support Evil interface

### LINKED ISSUE #7497: load saved game, saved game selection dialog: not enough space for displaying game names

Districh-ru:
Close #7497, relates to #1892

This PR implements variable size for Save/Load dialog, using StandardWindow class.

- The horizontal size of window is extended not to cut the save file names if a name was given in fheroes2 (you can rename save file in OS and if the name is very long it will be cut with '...' at the end).
- The vertical size is set automatically according to the selected screen height and available save files: dialog is not higher than (screen height - 100) and not lower than 4 rows in the list.
- Fixed rendering of slider when item count in the list is equal or less than rows in dialog (when there is nowhere to scroll in the list).
- Added setting the `needRedraw` to `true` in cases when list was changed (list item removed, list item unselected).
- Applied most proposals of #1892;
- ~Added in-game date near the name of selected file to load.~

изображение

изображение

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -75,23 +75,7 @@ namespace
         return res;
     }
 
-    void redrawShortGameDate( fheroes2::Image & output, const Maps::FileInfo & info, const int32_t dstx, const int32_t dsty, const int32_t dateTimeWidth,
-                              const fheroes2::FontType font )
-    {
-        if ( info.worldDay == 0 && info.worldWeek == 0 && info.worldMonth == 0 ) {
-            return;
-        }
-
-        std::string statusMsg = _( "M:%{month}, W:%{week}, D:%{day}" );
-        StringReplace( statusMsg, "%{month}", info.worldMonth );
-        StringReplace( statusMsg, "%{week}", info.worldWeek );
-        StringReplace( statusMsg, "%{day}", info.worldDay );
-
-        const fheroes2::Text text( statusMsg, font );
-        text.draw( dstx + ( dateTimeWidth - text.width() ) / 2, dsty, output );
-    }
-
-    void redrawDateTime( fheroes2::Image & output, const uint32_t timestamp, const int32_t dstx, const int32_t dsty, const fheroes2::FontType font )
+    void redrawDateTime( fheroes2::Image & output, const time_t timestamp, const int32_t dstx, const int32_t dsty, const fheroes2::FontType font )
     {
         const size_t arraySize = 5;
 
@@ -321,18 +305,10 @@ namespace
         background.applyTextBackgroundShading( { listRoi.x, listRoi.y, textInputRoi.width, listRoi.height } );
         background.applyTextBackgroundShading( { listRoi.x + textInputRoi.width, listRoi.y, dateTimeWidth, listRoi.height } );
         background.applyTextBackgroundShading( textInputRoi );
+        // Make background for the selected file date and time.
+        background.applyTextBackgroundShading( { dateTimeoffsetX, textInputRoi.y, dateTimeWidth, textInputRoi.height } );
 
-        std::unique_ptr<fheroes2::ImageRestorer> textInputBackground;
-        if ( isEditing ) {
-            textInputBackground
-                = std::make_unique<fheroes2::ImageRestorer>( fheroes2::Display::instance(), textInputRoi.x, textInputRoi.y, textInputRoi.width, textInputRoi.height );
-        }
-        else {
-            // Make background for the selected file date and time.
-            background.applyTextBackgroundShading( { dateTimeoffsetX, textInputRoi.y, dateTimeWidth, textInputRoi.height } );
-            textInputBackground
-                = std::make_unique<fheroes2::ImageRestorer>( fheroes2::Display::instance(), textInputRoi.x, textInputRoi.y, listRoi.width, textInputRoi.height );
-        }
+        fheroes2::ImageRestorer textInputAndDateBackground( fheroes2::Display::instance(), textInputRoi.x, textInputRoi.y, listRoi.width, textInputRoi.height );
 
         const bool isEvilInterface = Settings::Get().isEvilInterfaceEnabled();
 
@@ -447,8 +423,10 @@ namespace
             fheroes2::Sprite released;
             fheroes2::Sprite pressed;
 
-            makeButtonSprites( released, pressed, "...", 15, isEvilInterface, false );
-            buttonVirtualKB = std::make_unique<fheroes2::ButtonSprite>( dateTimeoffsetX + 5, textInputRoi.y - 3, released, pressed );
+            const int32_t buttonWidth = 17;
+
+            makeButtonSprites( released, pressed, "...", buttonWidth, isEvilInterface, false );
+            buttonVirtualKB = std::make_unique<fheroes2::ButtonSprite>( area.x + ( area.width - buttonWidth ) / 2, buttonOffsetY, released, pressed );
 
             fheroes2::addGradientShadow( released, display, buttonVirtualKB->area().getPosition(), { -5, 5 } );
             buttonVirtualKB->draw();
@@ -456,8 +434,8 @@ namespace
             Game::passAnimationDelay( Game::DelayType::CURSOR_BLINK_DELAY );
         }
         else if ( listbox.isSelected() ) {
-            // Render the game date from file info.
-            redrawShortGameDate( display, listbox.GetCurrent(), dateTimeoffsetX, textInputRoi.y + 4, dateTimeWidth, fheroes2::FontType::normalYellow() );
+            // Render the saved file date and time.
+            redrawDateTime( display, listbox.GetCurrent().timestamp, dateTimeoffsetX, textInputRoi.y + 4, fheroes2::FontType::normalYellow() );
         }
 
         display.render( background.totalArea() );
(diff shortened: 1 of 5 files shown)


src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -72,430 +72,524 @@ namespace
...
+    bool redrawTextInputField( const std::string & filename, const fheroes2::Rect & field, const bool whiteFont )
     {
         fheroes2::Display & display = fheroes2::Display::instance();
 
-        const fheroes2::Text title( header, fheroes2::FontType::normalYellow() );
-        title.draw( windowArea.x + ( windowArea.width - title.width() ) / 2, windowArea.y + 32, display );
-
         if ( filename.empty() ) {
             return false;
         }
```
ihhub, line 81: Shall we put this check before getting display to avoid extra operations?

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -72,430 +72,524 @@ namespace
...
-    char shortMinutes[20];
+        // TODO: Use system locale date format (e.g. 9 Mar instead of Mar 09).
+        std::strftime( shortMonth, arraySize, "%b", &tmi );
+        std::strftime( shortDate, arraySize, "%d", &tmi );
+        std::strftime( shortHours, arraySize, "%H", &tmi );
+        std::strftime( shortMinutes, arraySize, "%M", &tmi );
+        std::string savname( System::GetBasename( info.file ) );
 
-    const tm tmi = System::GetTM( info.timestamp );
+        if ( !savname.empty() ) {
```
ihhub, line 190: Since we are moving this code can we please use the opposite logic for fast return?
```cpp
if ( savname.empty() ) {
    return;
}
```

Then the number of changes would be less in this pull request and it would be easier to read in the future.

Moreover, we should put this code at the beginning of the function to avoid unnecessary operations:
```cpp
std::string savname( System::GetBasename( info.file ) );
if ( savname.empty() ) {
    return;
}
```

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -72,430 +72,524 @@ namespace
...
 
-    return list2;
-}
+    MapsFileInfoList getSortedMapsFileInfoList()
+    {
+        ListFiles list1;
+        list1.ReadDir( Game::GetSaveDir(), Game::GetSaveFileExtension(), false );
+
+        MapsFileInfoList list2( list1.size() );
+        int32_t saveFileCount = 0;
```
ihhub, line 261: Can we please use **size_t** here and drop static_cast call at line 267?

ihhub (changes requested):
Hi @Districh-ru , I put few comments here. My main notes are related to visual interpretation of the dialog:
- despite of the fact that the original game centered titles for save files I think having them shifted to right is visually better. This allows for faster search of specific files.
- I am not sure about the separator in the middle of the text. This separator works only for English date representation and also it differs from the rest of dialogs. Should we remove it?
- current selection text box should be the same length as the above text area. It seems unnatural having a truncated size text box. We accept that a user might enter a lengthy name but we show only that much in the above text box
- the screenshot for evil interface is missing virtual keyboard button. Is it intentional?

What do you think @Branikolog and @zenseii ?

Districh-ru:
Hi, @ihhub.
> - I am not sure about the separator in the middle of the text. This separator works only for English date representation and also it differs from the rest of dialogs. Should we remove it?
> - current selection text box should be the same length as the above text area. It seems unnatural having a truncated size text box. We accept that a user might enter a lengthy name but we show only that much in the above text box

The separator could be removed, it was made to visually separate file name from date/time like in tables. In future we will add the ability to sort files by date.
Maybe we should make the headers **Save Name** and **Date** above the list?
Currently we have only English date representation. I left a TODO to implement date representation in system locale. And if we swap month and date the date text will have almost the same length and will fit the separator: "Sep 01,..." <-> "01 Сен,...".
The text input box should have the same width as the maximum displayedfile name width not to allow to input filenames longer than can be displayed. Or we will never fix #7497 (until we reach the OS file name length limit) :)
The part of the list is date/time and we do not input time and date. We can also introduce a separator in text input and display some extra info in the right part.


> - the screenshot for evil interface is missing virtual keyboard button. Is it intentional?

The screenshot for Evil interface is done for Load Game dialog, in this dialog we only select file by click and do not input its name. :)

ihhub:
@Districh-ru , also would be it be better to have the edit box the same length as the 2 text boxes above and for load mode show a date but not for the save mode?

ihhub:
@Branikolog and @zenseii , could you please express your opinion about this UI change? This is a must have feature for the release.

ihhub:
@Districh-ru , in meanwhile could you please address my comments as they don't affect the look of the window? :)

Districh-ru:
> Having read through the comments again, I believe we should have the extra box near the selected save show the in game time of the save, since we can see the real-life one above.

Done. This is how it looks now:
изображение

Districh-ru:
> This looks very good to me. Out of curiosity and not having looked at the code. What happens when there are double digit months w/r to fitting in the boxed are? (Weeks and days only range from 1-7 and 1-4 and can't be double digit of course.)

There is a space for one digit:
изображение
We can always expand the "dateTime" area width if this space is not enough for the translated text (if it will have three wide letters).

ihhub:
Hi @Districh-ru and @zenseii , displaying just a date is not a great idea as players usually finish a map within 1-2 days. They pay more attention to time. Also we must be consistent and display exactly like in the box above. For save files we can display the datetime of opening the dialog. What do you think?

Districh-ru:
> Hi, @Districh-ru. I realized that we already have some UI design for headers present in the original game in the scenario selection dialog.
> 
> My proposal is that we can add buttons with the text "NAME" and "DATE" (and other options?), which pressing them would also sort the list accordingly. As a bonus, this would solve completly or partially the issues #1211 (@LeHerosInconnu's first issue so maybe it's about time😅) and #4053. 

Hi, @zenseii. Like this?

изображение

But IMHO it'll be better to implement date sorting in a new PR because this one is already big (and I'm busy today :) ).

I'm marking this PR as ready for review.

Branikolog:
Hi, @Districh-ru !

I've checked your PR and I love it really much! 😇 

There're only a few minor issues I want to mention and hear your opinion as well as our team for sure.
image
Save file with WWW fits the list well. But the bottom line shows "..." instead of the last symbol. Is it intended behaviour here? Since these boxes have the same width, I'm surprised to see that text doesn't fit into bottom line.

Since you've moved virtual keyboard button from the right bottom cornet to the central bottom line I believe the button with "..." could be a little wider, since there's enough space between "Okay" and "Cancel" buttons. This could be done in separate PR for sure.

And my biggest concern is about the window size itself. 
While on x480 resolutions it is comparable to other game windows and looks acceptable:
image

On larger resolutions the list looks terrifying (It's too large and too many strings in it. I believe, that our programmers from the Team are got used to a large number of text strings 😅, but within the game UI it looks quite overwhelming.).
image
I agree, that using higher resolutions we should employ all advantages. But I nevertheless suggest making it vertically smaller. Not the same as for x480 resolutions, but I prefer it being not larger, than system option menu (preferably in my opinion for smaller resolutions) or scenario info window (could be better for fHD etc).

Branikolog:
Hi, @zenseii and @ihhub 

> I believe we should left-side align the save names instead of right-side align them as @ihhub suggested. I don't feel too strongly about changing this from what it is originally.

I'm also got used to left-side text aligning, but after we've stretched the box with file names I'm afraid it would be difficult to correlate name and date (especially if the name is really short). So I think placing them in the middle is a way to go...

ihhub:
@Districh-ru , @Branikolog , @zenseii , let's leave any new features for this dialog outside this pull request and purely focus only on enhancing the existing dialog.

ihhub (approved):

Branikolog:
@zenseii 

> I believe one problem in your image is the jarring difference in resolution between the stretched out main menu background, and the non-stretched file select dialog.

Not exactly. The main thing bothers me is that it is the largest window in the game now. Even much bigger, than battle or castle screens. And whole internals of this window is full of text, like an excel file. The rest windows in the game are much smaller and more balanced, than this one, which looks like a bill. 😅 
From a usability point this window is great, as I don't have to scoll list to reach the lowest files. But from a design point, it stands out being too massive and monotonous containing just files, files and only files. 
I suggest making it not as short, as for 640x480, but to make it a little more uniform to existing UI elements, reducing its length up to ~ 22-25 strings.

Districh-ru:
> Save file with WWW fits the list well. But the bottom line shows "..." instead of the last symbol. Is it intended behaviour here? Since these boxes have the same width, I'm surprised to see that text doesn't fit into bottom line.

Hi, @Branikolog, this is intentional behavior because in the text input field we have "WWWWWWWWWW" with a '_' cursor at the end.
In the text input field, we limit the maximum length of the save file name so that it is displayed without truncation in the list.
Currently anyone can enter a very-very long phrase by using the Virtual Keyboard, we should implement such limit there in future PRs. Also everybody can rename the save file in OS explorer. In these cases we will have the file name larger than we can display.

We can change the parameters of this dialog at any time (minimum, maximum lines; dialog width, text aligning) if we get reasoned feedback.

Branikolog (approved):
Amazing work!

ihhub:

petar-stamenkov:
Hello, while testing if everything's okay after merging the changes introduced in this PR to the branch of #7680, I noticed there is a crash if you click at the end of the save filename input and add a character to the filename. It looks like it's caused by the addition of underscore to the filename here:
https://github.com/ihhub/fheroes2/blob/117839baa6554684d360088f1e80442fc6561084/src/fheroes2/dialog/dialog_selectfile.cpp#L494

Changing it to pass the filename without underscore seems to fix the crash.

Districh-ru:
> Hello, while testing if everything's okay after merging the changes introduced in this PR to the branch of #7680, I noticed there is a crash if you click at the end of the save filename input and add a character to the filename. It looks like it's caused by the addition of underscore to the filename here:
> https://github.com/ihhub/fheroes2/blob/117839baa6554684d360088f1e80442fc6561084/src/fheroes2/dialog/dialog_selectfile.cpp#L494
> 
> Changing it to pass the filename without underscore seems to fix the crash.

Hi, @petar-stamenkov. Thanks for the bug report! I'll fix this regression later today.

PR #7667, opened 2023-08-29
Do not try to scout the area if the hero is in jail and gets an artifact assigned using the map editor

### LINKED ISSUE #7663: Core dump on opening the map
### Preliminary checks

### Platform

Linux

### Describe the bug

When trying to start single player game on attached map, the game crashes with following errors:

```
../fheroes2/heroes/heroes.cpp:1428: void Heroes::Scout(int) const: Assertion `GetColor() != Color::NONE' failed.
```

### Save file

[THEIMMOR.MX2.zip](https://github.com/ihhub/fheroes2/files/12468996/THEIMMOR.MX2.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #7663

Also fixed a typo: `Heroes::BAX` -> `Heroes::BRAX`.


src/fheroes2/heroes/heroes.cpp
```diff
@@ -1125,6 +1125,11 @@ bool Heroes::PickupArtifact( const Artifact & art )
         std::for_each( assembledArtifacts.begin(), assembledArtifacts.end(), Dialog::ArtifactSetAssembled );
     }
 
+    // If the hero is in jail and gets an artifact assigned using the map editor, then there is no need to scout the area
+    if ( GetColor() == Color::NONE ) {
```
ihhub, line 1129: Should we check hero flag **JAIL** instead? I am just worrying what if someone tries to add a neutral hero on map but not in Jail.

ihhub (changes requested):
Hi @oleg-derevenetz , I left one suggestion here. Could you please check it?

src/fheroes2/heroes/heroes.cpp
oleg-derevenetz: Hi @ihhub We can't just swap them :) We need to rework the corresponding loading method a bit because it clears all flags (and it's right about that). OK, I'll try to rework this.

ihhub: mmm... then we can leave it as it is. If it takes too much time it is okay to use your approach.

oleg-derevenetz: @ihhub I'll try to do this later today.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these chnages!

PR #7671, opened 2023-08-31
Do not show second brackets when hero can learn secondary skill from witch hut

### LINKED ISSUE #7670: Adventure map, Witch's Hut, right-click information window, a line with brackets and no text is displayed below the secondary skill line when the selected hero does not have the secondary skill of the witch's hut
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, a line with brackets and no text is displayed below the secondary skill line when the selected hero does not have the secondary skill of the witch's hut.

In fheroes2:

Archery hut bug 01

### Save file

Here is the save file:
[Archery hut bug 01.zip](https://github.com/ihhub/fheroes2/files/12476937/Archery.hut.bug.01.zip)

### Additional info

Version 1.0.7 7640

lwollnikowski:
Fix #7670

- hero does not know skill, has all skill slots occupied = 'already has max skills'
- hero does know skill, has all skill slots occupied = 'already knows this skill'
- hero does not know skill, can learn new skills = no second brackets
- hero does know skill, can learn new skills = 'already knows this skill'
Screenshot 2023-08-31 10-10-56
Screenshot 2023-08-31 10-11-10
Screenshot 2023-08-31 10-11-14

Screenshot 2023-08-31 10-11-23
Screenshot 2023-08-31 10-11-25
Screenshot 2023-08-31 10-11-40

### DIFF
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -276,16 +276,20 @@ namespace
             const Heroes * hero = Interface::GetFocusHeroes();
 
             if ( hero ) {
-                str.append( "\n(" );
+                std::string heroStr;
 
                 if ( hero->HasSecondarySkill( skill.Skill() ) ) {
-                    str.append( _( "already knows this skill" ) );
+                    heroStr = ( _( "already knows this skill" ) );
                 }
                 else if ( hero->HasMaxSecondarySkill() ) {
-                    str.append( _( "already has max skills" ) );
+                    heroStr = ( _( "already has max skills" ) );
                 }
 
-                str += ')';
+                if ( !heroStr.empty() ) {
+                    str.append( "\n(" );
+                    str.append( heroStr );
+                    str += ')';
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -276,16 +276,20 @@ namespace
             const Heroes * hero = Interface::GetFocusHeroes();
 
             if ( hero ) {
-                str.append( "\n(" );
+                std::string heroStr;
```
ihhub, line 279: Let's use C-string here with nullptr by default. Then we won't have an extra std::string. What do you think?

ihhub (changes requested):
Hi @lwollnikowski , I left one suggestion here. Could you please check it when you have time?

src/fheroes2/dialog/dialog_quickinfo.cpp
lwollnikowski: I can use it. I know std::string has little resources overhead over C-string, are there other reasons?

ihhub (approved):

ihhub:
@lwollnikowski , thank you so much for this fix!

src/fheroes2/dialog/dialog_quickinfo.cpp
ihhub: This is the only reason. We pay a lot of attention to performance so if it's possible to avoid extra computations it's better to do it. After all running a well optimized application on a mobile phone means longer battery life :)

PR #7676, opened 2023-09-02
Change Army Order to Turn Order

zenseii:
This has been bothering me for long enough now.

Army Order sounds like and means something more similar to this
image

If Turn Order isn't accepted, then alternatives are "Battle Queue" or "Turn Queue".

I've changed the relevant identifiers I could find that were using some variant of "army order".

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1295,7 +1295,7 @@ Battle::Interface::~Interface()
 
 void Battle::Interface::SetOrderOfUnits( const std::shared_ptr<const Units> & units )
 {
-    armies_order.Set( GetArea(), units, arena.GetArmy2Color() );
+    _turnOrder.Set( GetArea(), units, arena.GetArmy2Color() );
 }
 
 fheroes2::Point Battle::Interface::GetMouseCursor() const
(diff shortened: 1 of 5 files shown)


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1008,7 +1008,7 @@ void Battle::ArmiesOrder::RedrawUnit( const fheroes2::Rect & pos, const Battle::
 {
     // Render background.
     const fheroes2::Sprite & backgroundOriginal = fheroes2::AGG::GetICN( ICN::SWAPWIN, 0 );
-    fheroes2::Copy( backgroundOriginal, 37, 268, output, pos.x + 1, pos.y + 1, armyOrderMonsterIconSize - 2, armyOrderMonsterIconSize - 2 );
```
ihhub, line 1011: Shall we rename **ArmiesOrder** class as well into **TurnOrder**?

ihhub (commented):
Hi @zenseii , I am absolutely fine with the change. I left one suggestion in the code. Could you please take a look?

ihhub:
This PR tells me that our AI doesn't use Army order option at all :)

oleg-derevenetz:
> This PR tells me that our AI doesn't use Army order option at all :)

Human players rarely use this options too. I can't even remember when I used it last time.

ihhub:
> > This PR tells me that our AI doesn't use Army order option at all :)
> 
> Human players rarely use this option too. I can't even remember when I used it last time or have seen how HoMM2/HoMM3 streamers use it.

True. However, in some cases army formation could help to win a battle.

src/fheroes2/battle/battle_interface.h
```diff
@@ -458,7 +458,7 @@ namespace Battle
         std::unique_ptr<StatusListBox> listlog;
 
         PopupDamageInfo popup;
-        ArmiesOrder armies_order;
+        TurnOrder armies_order;
```
ihhub, line 461: Should we rename this variable as well into **_turnOrder**?

ihhub (approved):

ihhub:
@zenseii , thank you so much for the change!

PR #7677, opened 2023-09-02
Fix SonarQube bugs and blocker issues

ihhub:
plus small code cleanup.

regression from #7640

### DIFF
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -27,6 +27,7 @@
 #include <array>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <map>
 #include <ostream>
 #include <string>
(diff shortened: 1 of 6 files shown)


src/engine/translations.cpp
```diff
@@ -191,8 +191,15 @@ namespace
     const char * stripContext( const char * str )
     {
         const char * pos = str;
-        while ( *pos && *pos++ != contextSeparator )
-            ;
+        while ( *pos ) {
```
oleg-derevenetz, line 194: We can just use [`std::strchr()`](https://en.cppreference.com/w/cpp/string/byte/strchr) here:

https://godbolt.org/z/9Y1E7rhec

It will probably even work faster.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion, could you please take a look when possible?

oleg-derevenetz (approved):

PR #7680, opened 2023-09-02
Show latest save files first

### LINKED ISSUE #1211: Selecting Game Saves

petar-stamenkov:
fix #1211

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -236,7 +236,7 @@ MapsFileInfoList GetSortedMapsFileInfoList()
     if ( static_cast<size_t>( saveFileCount ) != list2.size() ) {
         list2.resize( saveFileCount );
     }
-    std::sort( list2.begin(), list2.end(), Maps::FileInfo::FileSorting );
+    std::sort( list2.begin(), list2.end(), Maps::FileInfo::TimestampSorting );
 
     return list2;
 }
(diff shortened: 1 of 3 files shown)


ihhub:
We shouldn't blindly change the way how save files are sorted. The original issue was created due to small width of the window which was fixed recently. Moreover, the correct approach is to add an option to choose sorting style ( #4053).

When a player has hundreds of save files which are sorted by time this would be a nightmare to browse within. I am talking about the majority of players.

petar-stamenkov:
@ihhub I see, that sounds reasonable. I guess this PR can be closed then.

Do you have any advice for new contributors on how to pick issues to work on - for example some issue label to look for or something like that?

petar-stamenkov:
Hi @zenseii, thanks for the help!

joelostblom:
I think having files sorted in chronological order by default would be a helpful feature to have even when https://github.com/ihhub/fheroes2/issues/4053 is implemented (and it seems like that might take a while since it is a bigger feature). What do you think about having the chronological order suggested here as the default for now and keep it that way even after that PR is implemented? 

> When a player has hundreds of save files which are sorted by time this would be a nightmare to browse within. I am talking about the majority of players.

Regarding this, I would expect that the most common user scenario is to want to load the latest saved file and not to browse through many saved files looking for a specific one, but we might have different experience with how we use the game here (I'm a more casual player myself who just found this fantastic project after not playing HOMM2 for years, so thank you for all your work on this project)

PR #7698, opened 2023-09-04
Translate disabled buttons and reset them when changing languages

zenseii:
Relates to #6105 

Example of when you have only demo files (Note the difference between disabled and pressed state in the cancel button):
image

The MAX disabled button was removed because it wasn't used anywhere and we have code for its generation in the recruit 
creatures dialog afaik, and that already gets translated correctly.

One question is whether we should generate the disabled campaign and vertical dismiss buttons in the code where they are used, or keep it the way it is in this PR.

The main argument for keeping it the way it is in this PR is if we will use these buttons in more than one place.


ihhub (approved):

Districh-ru (approved):
👍

ihhub:
@zenseii , thank you so much for handling this part!

PR #7700, opened 2023-09-04
Fix regressions after #7662: cursor position setting in text input field and display restoration after closing the virtual keyboard

Districh-ru:
Fix regressions after #7662:
- fix cursor position setting in text input field in Save dialog:
    - clicking to the right of a file name and trying to type a letter caused the game crash;
    - if the filename was truncated with '...', clicking on it resulted in an unpredictable cursor position;
- fix display redraw after closing the Virtual Keyboard: the part outside of the Save dialog was updated with a delay.


ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for these quick fixes!

Issue #7548, opened 2023-08-12
German translation in upgrade creature dialog: button label misaligment and too long creature trait names


kitovyj:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

German translation in upgrade creature dialog: misaligment and too long creature trait names, please see the screenshot. Note also untranslated "exit" in the city screen.
![german tr](https://github.com/ihhub/fheroes2/assets/15792725/da01bb47-7549-4c0a-82d1-3b4395e29c57)

### Save file

not needed

ihhub:
Hi @kitovyj , can you please come up with another word or a shorter version of it for German translation? I believe Exit in castle window is a separate issue.

kitovyj:
> Hi @kitovyj , can you please come up with another word or a shorter version of it for German translation? I believe Exit in castle window is a separate issue.

Maybe @Laserlicht can help here, I don't have german version of the game and I'm not a native speaker. I have the version from GOG, maybe somebody knows how to switch it to german version, then I can just look there.

Laserlicht:
The official buttons:
![grafik](https://github.com/ihhub/fheroes2/assets/13953785/1f9ab925-20e8-4d49-a21c-fffa5eec9469)

In the original resources the font space is reduced to fit into button...

Perhaps an abbreviation would also have to be used. E.g. `ENTLAS.` or `ENTL.`

kitovyj:
> 

Thanks for the suggestion. Hah, I see that in the original version they also have long "Geschwindigkeit" not fitting the dialog space.

kitovyj:
> > The official buttons:
> > In the original resources the font space is reduced to fit into button...
> > Perhaps an abbreviation would also have to be used. E.g. `ENTLAS.` or `ENTL.`
> 
> Hi, everyone. I will make these buttons expand or shrink dynamically according to the text on them in a PR so we can avoid abbreviations (hopefully). Same thing I did in the campaign dialog.

Well, in this case it will not help, the buttons would overlay, or ENDE will become small and not convdnient to use.

kitovyj:
[...]

PR #7706, opened 2023-09-05
Make battle buttons translatable

zenseii:
Relates to and final part of #6105 

This has been implemented:

1. Variable width for the SKIP and AUTO buttons according to their texts.
2. Variable width of the Status Bar according to the widths of the buttons.

For the Cyrillic font it looks quite good even though the buttons are slightly wider than the originals which means we have 15 pixels less for text:

image

Meanwhile, for the Latin font it is obvious that we have to cram the letters closer together. I found the sweet spot to be 2 pixels closer in the x-direction.

This is what it looks like for the Latin font without cramming:

image

Implementing a feature to cram letters would need a new function for calculating text width because it needs to take into account that the letters are rendered on top of each other, and we need a new function for actually rendering the text with letters on top of each other.

This change would give us 12 pixels more in width in the case when we have two words of 4 letters (ex. AUTO + SKIP),  ( 2 pixels * ( 2 words * ( 4 letters - 1 first letter )  ) ). We would then only lose 3 pixels for text.

Such big changes might be better to do in a separate PR.

This change could be used to solve button problems like those described here https://github.com/ihhub/fheroes2/issues/7548

This is what the SKIP button would look like with 2 pixels closer for letters.
image

Related to #9261

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -191,7 +191,9 @@ namespace
                                                 ICN::BUTTON_QUICK_COMBAT_GOOD,
                                                 ICN::BUTTON_QUICK_COMBAT_EVIL,
                                                 ICN::BUTTON_SKIP,
-                                                ICN::BUTTON_AUTO };
+                                                ICN::BUTTON_AUTO,
+                                                ICN::BUTTON_BATTLE_SETTINGS,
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @zenseii , can you please explain what is missing to be implemented in this pull request? Can we help to move forward with the changes or the idea?

PR #7711, opened 2023-09-07
Show highscores after multiplayer game

### LINKED ISSUE #4470: No highscore for multiplayer
### Preliminary checks

### Platform

Windows

### Describe the bug

Win multiplayer game by eliminating red player:
image
After that we watch victory animation with dragon and then return to main menu.

### Save file

[Multiplayer score.zip](https://github.com/ihhub/fheroes2/files/7397740/Multiplayer.score.zip)

### Additional info

_No response_

petar-stamenkov:
Fixes #4470

### DIFF
--- a/src/fheroes2/game/game_over.cpp
+++ b/src/fheroes2/game/game_over.cpp
@@ -490,8 +490,7 @@ fheroes2::GameMode GameOver::Result::checkGameOver()
         }
     }
     else {
-        // The result of the multiplayer game shouldn't be stored by this class
-        const uint32_t multiplayerResult = [currentColor, activeHumanColors]() -> uint32_t {
+        result = [currentColor, activeHumanColors]() -> uint32_t {
             // If all human-controlled players have been vanquished, then the game is over
             if ( activeHumanColors == 0 ) {
                 return GameOver::LOSS_ALL;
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Are the multiplayer results just supposed to be among standard/single player game results? Shouldn't there be a separate list for them (just like for campaign)?

oleg-derevenetz:
> As pointed out in the issue thread, in the OG the multiplayer high score is shared with the non-multiplayer high score.

Funny, I always thought they should be separate. Well, OK then :)

src/fheroes2/game/game_over.cpp
```diff
@@ -539,16 +538,16 @@ fheroes2::GameMode GameOver::Result::checkGameOver()
...
             AudioManager::ResetAudio();
             Video::ShowVideo( "WIN.SMK", Video::VideoAction::WAIT_FOR_USER_INPUT, { standardGameResults() }, true );
 
-            res = fheroes2::GameMode::MAIN_MENU;
+            res = fheroes2::GameMode::HIGHSCORES_STANDARD;
         }
-        else if ( multiplayerResult & GameOver::LOSS ) {
-            const bool showLossDialog = [currentColor, isCurrentPlayerWasActive, multiplayerResult]() {
+        else if ( result & GameOver::LOSS ) {
+            const bool showLossDialog = [currentColor, isCurrentPlayerWasActive, currentResult = result]() {
```
oleg-derevenetz, line 550: We can just add `this` here to the capture list and access the `result` directly instead of `currentResult = result`, but it's a matter of taste.

oleg-derevenetz (approved):
LGTM. BTW I just checked HoMM3 and in this game multiplayer results are mixed with single player results as well.

src/fheroes2/game/game_over.cpp
petar-stamenkov: Thanks for the suggestion, I've updated it.

oleg-derevenetz: Hi @petar-stamenkov both variants are fine. The advantage of the previous variant is that `currentResult` was `const`, and when you are not going to change the variable, `const` is good. But the new variant is also good.

ihhub (approved):

ihhub:
@petar-stamenkov , huge thanks for this improvement!

PR #7713, opened 2023-09-07
Fix number placeholders not being replaced in campaign spell bonus texts

zenseii:
Fixes this %{count} not being replaced:
image

### DIFF
--- a/src/fheroes2/gui/ui_campaign.cpp
+++ b/src/fheroes2/gui/ui_campaign.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2022                                                    *
+ *   Copyright (C) 2022 - 2023                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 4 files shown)


src/fheroes2/spell/spell.cpp
```diff
@@ -147,9 +148,23 @@ const char * Spell::GetName() const
     return _( spells[id].name );
 }
 
-const char * Spell::GetDescription() const
```
ihhub, line 150: I recommend to revert these changes and in `ui_campaign.cpp` use **getSpellDescription()** function call. `GetDescription()` in the future will be gone as we will move **spellstats_t spells[]** array to `spell_info.cpp` class to allow adding new entries like we've done for monsters and artifacts.

ihhub (changes requested):
Hi @zenseii , I proposed an alternative approach. Could you please take a look?

src/fheroes2/gui/ui_campaign.cpp
```diff
@@ -103,7 +105,7 @@ namespace fheroes2
         case Campaign::ScenarioBonusData::SPELL: {
             const Spell spell( bonusData._subType );
             const SpellDialogElement spellUI( spell, nullptr );
-            const TextDialogElement spellDescriptionUI( std::make_shared<Text>( spell.GetDescription(), FontType::normalWhite() ) );
```
ihhub, line 106: Could you please also fix the same issue at line 203?

ihhub (approved):

ihhub:
@zenseii , thank you so much for this fix!

PR #7582, opened 2023-08-20
Make vertical buttons translatable

zenseii:
This makes the vertical DISMISS and EXIT buttons in the hero screen translatable.

image

PR #7714, opened 2023-09-07
Generate vertical high scores buttons

zenseii:
Part of #6105 

The horizontal centering of the N looks off in Standard probably because its width is an even number (+1 for x offset), like described in #7582 
image

I used the original buttons due to the background which I'm not sure how to make a template background of.


Districh-ru (changes requested):
Hi, @zenseii. I left one suggestion regarding computation optimization.

And what do you think about fixing the horizontal position of single-letter text lines?
This can be done in a new PR.

We can modify this line: 
https://github.com/ihhub/fheroes2/blob/3e6efc003eaeffc9fb051359baf0107d6145ac8c/src/fheroes2/gui/ui_text.cpp#L309

In example, by this code:
```cpp
// For button font single letters in a row we add extra 1 to the width to more properly center odd-length letters.
const int32_t esxtraOffsetX
    = ( size == 1 && ( fontType.size == fheroes2::FontSize::BUTTON_RELEASED || fontType.size == fheroes2::FontSize::BUTTON_PRESSED ) ) ? 1 : 0;
render( data, size, x + ( maxWidth - correctedLineWidth + esxtraOffsetX ) / 2, y, output, fontType );
```

изображение изображение

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1907,6 +1910,53 @@ namespace fheroes2
...
+                _icnVsSprite[id].resize( 2 );
+
+                const int32_t originalID = ICN::HISCORE;
+                int32_t originalICNIndex = 0;
+                if ( id == ICN::BUTTON_HSCORES_VERTICAL_STANDARD ) {
+                    originalICNIndex = 2;
+                }
+                else if ( id == ICN::BUTTON_HSCORES_VERTICAL_EXIT ) {
+                    originalICNIndex = 4;
+                }
```
ihhub, line 1925: Since we are handling only 3 cases I suggest to add another if-else branch with an assertion:
```cpp
else {
    assert( id == ICN::BUTTON_HSCORES_VERTICAL_CAMPAIGN );
}
```

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1907,6 +1910,53 @@ namespace fheroes2
 
                 break;
             }
+            case ICN::BUTTON_HSCORES_VERTICAL_CAMPAIGN:
+            case ICN::BUTTON_HSCORES_VERTICAL_STANDARD:
+            case ICN::BUTTON_HSCORES_VERTICAL_EXIT: {
```
ihhub, line 1915: A very minor nit-picking: ideally it is good to sort cases in alphabetic order to easier reading. It's up to you if you want to keep this way.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1907,6 +1910,53 @@ namespace fheroes2
...
+                }
+
+                if ( useOriginalResources() ) {
+                    _icnVsSprite[id][0] = GetICN( originalID, originalICNIndex );
+                    _icnVsSprite[id][1] = GetICN( originalID, originalICNIndex + 1 );
+                    break;
+                }
+
+                for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
+                    const Sprite & originalButton = GetICN( originalID, originalICNIndex + i );
```
ihhub, line 1934: We can avoid static_cast for the loop and do this:
```cpp
for ( size_t i = 0; i < _icnVsSprite[id].size(); ++i ) {
    const Sprite & originalButton = GetICN( originalID, originalICNIndex + static_cast<uint32_t>( i ) );
```
Please note that I changed from `int32_t` to `uint32_t`.

ihhub (changes requested):
Hi @zenseii , I left just few minor comments here. Would you mind to check them when you have time?

Districh-ru:
> Hi, @Districh-ru. I've addressed your comments. There's still a problem with words with "M" that are aligned incorrectly. Look at the C and N which are aligned wrongly in CAMPAIGN, especially when compared to how N is aligned in STANDARD. The same issue with alignment happens in vertical DISMISS so I believe it has to do with the width of M. It does not happen for a wider letter like W.

Hi, @zenseii.
In these cases the "M" letter is 18 pixels wide and it gives the text widths = 19 (odd again).
We can try to fix it by expanding the conditions for the extraOffsetX:
```cpp
// For button font single letters in a row we add 1 extra pixel to the width to more properly center odd-width letters only if 'maxWidth' is not odd.
const int32_t extraOffsetX
    = ( size == 1 && ( maxWidth % 2 == 0 ) && ( fontType.size == fheroes2::FontSize::BUTTON_RELEASED || fontType.size == fheroes2::FontSize::BUTTON_PRESSED ) )
        ? 1
        : 0;
```

изображение изображение
Comparing to original ones:
изображение изображение



After this the only one misplaced letter is "G"... Aтв currently I don't see a solution except making a special condition tor this letter, like `if (data[0]=='G') {extraOffsetX+=1;}`. But let's leave for the future when we will get a better solution. You can leave a TODO, if you want (like: Implement pixel perfect letters placement.). I believe we will be able to fix it after we implement adaptive spacing between letters (including spacing/width setting for single letters), like for WA, when '/' line of W is overlapped with '/' part of A.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1907,6 +1910,56 @@ namespace fheroes2
...
+                    break;
+                }
+
+                for ( size_t i = 0; i < _icnVsSprite[id].size(); ++i ) {
+                    const Sprite & originalButton = GetICN( originalID, originalICNIndex + static_cast<uint32_t>( i ) );
+                    Sprite & out = _icnVsSprite[id][i];
+
+                    out = originalButton;
+                    // Clean the button
+                    Fill( out, 4 - static_cast<uint32_t>( i ), 4 + static_cast<uint32_t>( i ), 19, 123, getButtonFillingColor( i == 0 ) );
```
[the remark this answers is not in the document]
Districh-ru: Hi, @zenseii, you can do `static_cast<int32_t>` here in both cases because `i` will not reach the maximum value for `int32_t` and `Fill()` accepts `int32_t` here.
It will make Clang-Tidy happy and avoid triple conversion (size_t -> uint32_t -> int32_t).

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1907,6 +1910,56 @@ namespace fheroes2
 
                 break;
             }
+            case ICN::BUTTON_HSCORES_VERTICAL_CAMPAIGN:
+            case ICN::BUTTON_HSCORES_VERTICAL_EXIT:
+            case ICN::BUTTON_HSCORES_VERTICAL_STANDARD: {
+                _icnVsSprite[id].resize( 2 );
+
+                const int32_t originalID = ICN::HISCORE;
+                int32_t originalICNIndex = 0;
```
Districh-ru, line 1919: These two variables are used in `GetICN( int icnId, uint32_t index )`. So to avoid type conversions we can declare these variables using the corresponding types:
```suggestion
                const int originalID = ICN::HISCORE;
                uint32_t originalICNIndex = 0;
```

Districh-ru (changes requested):
Hi, @zenseii, I left one comment. Could you please take a look when you have time.

Districh-ru (approved):
Great, most of the buttons in the game can now be translated!

ihhub (approved):

ihhub:
@zenseii , huge thanks for this nice improvement!

PR #7715, opened 2023-09-07
Adjust the description of the Eagle Eye skill

### LINKED ISSUE #7702: Eagle Eye skill description misunderstanding
### Preliminary checks

Screenshot_20230904-204143
Screenshot_20230904-204250

The Eagle Eye skill description says that there's a chance for my hero to learn spells, that an enemy has "used AGAINST him during battle". 

Spells like bless, anti-magic, cure, stoneskin etc. isn't technically used AGAINST my hero. They're used on the enemy's own units... 

Rephrase the skill description as so:

"... used in battle against him"

That way, it can't be misunderstood. 

I have always thought that only curse or damage spells were meant to be learned, and that being able to learn bless spells was a bug in the original game

### Additional info

_No response_

oleg-derevenetz:
close #7702

But I'm not sure about the wording. Why not "... that was cast by an enemy during combat"?

### DIFF
--- a/src/fheroes2/heroes/skill.cpp
+++ b/src/fheroes2/heroes/skill.cpp
@@ -508,13 +508,13 @@ std::string Skill::Secondary::GetDescription( const Heroes & hero ) const
     case EAGLEEYE:
         switch ( Level() ) {
         case Level::BASIC:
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for the update!

PR #7717, opened 2023-09-07
Berserker spell: always use the nearest cell to attack the nearby unit

oleg-derevenetz:
Related to #7716

~~Not exactly a bug, but for now I'm marking it as a bug.~~
Fixes issue [reported](https://github.com/ihhub/fheroes2/pull/7717#issuecomment-1710900329) by @zenseii.

~~Based on some new code from #7690, so should be merged after that PR.~~

This PR changes the logic of determining the movement distance for the flying units from the `BattlePathfinder`'s point of view. Previously, the movement distance for flying units was always equal to 1 for any position. Now it is equal to the straight line distance to this position. This distance is used only for AI estimations and in theory this change should not break anything, but for the sake of confidence, it may not be worth including this change in an upcoming release until it has been properly tested.


oleg-derevenetz:
> Actually I restested it and was able to reproduce it with this PR too. It seems to be connected to those specific cells or the fact that the wolves were more distant in this video?

No, that's unrelated issue. This is because the distances between units were determined as the distances between their head cells, and when wolves are near cyclops, the straight line distance from the cyclops to the wolf's head is 2 cells and the straight line distance to the troll is also 2 cells, therefore, they were evaluated as being at an equal distance, although this is not really the case. In #7690, I have added a special function that determines distances of this kind correctly, and now I use it in the method `Battle::Board::GetNearestTroops()` (that returns a sorted list of nearest troops) too.

Why it sometimes worked correctly and sometimes not - because the distances were sorted using `std::sort()`, and it has the right to rearrange the "identical" elements as it pleases, so sometimes Wolves go first, and sometimes Trolls go first, depending on their exact positions on the board.

ihhub (approved):

ihhub:
@zenseii , a lot of thanks for this improvement!

ihhub:
@zenseii , oops... so yeah, @oleg-derevenetz did good as well🙂

PR #7719, opened 2023-09-08
Do not allow AI to use the Summon Boat spell if there is already a reachable tile with an empty boat nearby

### LINKED ISSUE #7709: Adventure map, AI summons a boat, right near the one he previously landed from
### Preliminary checks

### Platform

Windows

### Describe the bug

image

https://github.com/ihhub/fheroes2/assets/55348946/7f496ae3-4b12-4963-912e-66fb76f3e2e0

AI landed on a small island, but for some reason decided to summon a new boat while trying to escape from the island.

### Save file

[Useless summon.zip](https://github.com/ihhub/fheroes2/files/12539137/Useless.summon.zip)

### Additional info

_No response_

oleg-derevenetz:
close #7709

I'm still not sure if it's worth making this change, because it fixes a purely cosmetic "issue" at the expense of some additional computational complexity.

The case when there is a boat next to the AI-controlled hero and it is reachable:

https://github.com/ihhub/fheroes2/assets/32623900/ac44c004-867d-4e60-a612-c72011b9183f

Hero uses the existing boat, although it is not profitable for him, because he could summon the boat directly on the tile next to the Shipwreck Survivor, which would save him some movement points on the next turn.

The case when there is a boat next to the AI-controlled hero, but it is NOT reachable:

https://github.com/ihhub/fheroes2/assets/32623900/e6831db2-2aa9-4c81-9e43-40619a5cf5ba

The hero acts effectively by summoning the boat to the tile closest to the Shipwreck Survivor.

### DIFF
--- a/src/fheroes2/world/world_pathfinding.cpp
+++ b/src/fheroes2/world/world_pathfinding.cpp
@@ -315,7 +315,57 @@ namespace
         }
 
         // ... and this tile should be reachable from the shore (as if this shore tile were a water tile)
-        return toTile.isPassableFrom( Direction::Reflect( direction ), true, false, color );
+        if ( !toTile.isPassableFrom( Direction::Reflect( direction ), true, false, color ) ) {
+            return false;
+        }
+
+        // Even if it is possible to use the Summon Boat spell to get to this tile, this should not be done
+        // if there is already a boat on an adjacent reachable tile, because such a behavior looks ugly.
+        // Boat from the adjacent tile should be used instead.
+        const bool isEmptyBoatNearby = [from, color, &fromTile, fromWater]() {
+            const int32_t fromX = from % world.w();
+            const int32_t fromY = from / world.w();
+
+            for ( int32_t dy = -1; dy <= 1; ++dy ) {
+                for ( int32_t dx = -1; dx <= 1; ++dx ) {
+                    if ( dx == 0 && dy == 0 ) {
+                        continue;
+                    }
+
+                    const int32_t tileX = fromX + dx;
+                    const int32_t tileY = fromY + dy;
+
+                    if ( !Maps::isValidAbsPoint( tileX, tileY ) ) {
+                        continue;
+                    }
+
+                    const int32_t tileIdx = tileY * world.w() + tileX;
+                    const int tileDirection = Maps::GetDirection( from, tileIdx );
+
+                    assert( tileDirection != Direction::UNKNOWN && tileDirection != Direction::CENTER );
+
+                    if ( !fromTile.isPassableTo( tileDirection ) ) {
+                        continue;
+                    }
+
+                    const Maps::Tiles & tile = world.GetTiles( tileIdx );
+
+                    if ( tile.GetObject() != MP2::OBJ_BOAT ) {
+                        continue;
+                    }
+
+                    if ( !tile.isPassableFrom( Direction::Reflect( tileDirection ), fromWater, false, color ) ) {
+                        continue;
+                    }
+
+                    return true;
+                }
+            }
+
+            return false;
+        }();
+
+        return !isEmptyBoatNearby;
     }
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub @zenseii any thoughts? Is it worth it?

oleg-derevenetz:
> I suppose if the AI can see that it can cross the "sea part" of a voyage just as quickly as it could without using summon boat, then it would be better to save those spell points.

In general case, this doesn't work this way, because, just like I explained to @Branikolog, AI doesn't consider the sea part of a voyage as a some separate thing, it's just a part of a complex path that can be paved through wandering monsters, enemy heroes, objects with guardians, guarded artifacts, etc, etc, and it is impossible to say in general case "well, this path will save us 5 mana, and it's better because of this" while in fact this "better" path will be paved through some guarded object where hero may spend extra 50 mana instead. Such attempts of "savings" are useless.

ihhub:
Hi @oleg-derevenetz , I don't the described behavior as a bug. It's a perception of human players (@Branikolog ) thinking that boarding an existing boat is more beneficial. We shouldn't replicate what a human can do but we should implement the most optimal way of doing things.

Before proceeding with this change I would like to hear @Branikolog explanation how boarding a boat is more beneficial than summoning a new one if the last case would lead to a shorter path.

Branikolog:
Hello, my dear teammates. 
The main thing that bothers me with the issue, I initially opened, is the fact, that AI hero summons **the second boat** (without even a guaranty, that the new boat will appear at the tile, which is more beneficial regarding his planned path, since boat appears randomly (according to the logic, which AI, I suppose, doesn't consider)). 
So, while summoning one more boat AI doesn't consider, where comes this boat from. In some cases, it could be a boat, that another hero of the current player used to disembark from. I'm afraid of the situation, when the current hero steals the boat of some random hero so that hero potentially can't escape from the island he previously disembarked on. In some cases this can ruin the entire strategy of AI. So I think sacrificing a few cells from summoning another boat, is quite a reasonable solution to prevent other heroes from being stuck on island without boats.

oleg-derevenetz:
> AI hero summons the second boat (without even a guaranty, that the new boat will appear at the tile, which is more beneficial regarding his planned path, since boat appears randomly (according to the logic, which AI, I suppose, doesn't consider)).

No, there **is** such a guarantee. AI is cheating in this regard - the boat is summoned to exactly the right tile.

> In some cases, it could be a boat, that another hero of the current player used to disembark from. I'm afraid of the situation, when the current hero steals the boat of some random hero so that hero potentially can't escape from the island he previously disembarked on.

In the same way, AI can take away a boat from this hero even if there is no boat near to the hero that casts the Summon Boat spell.

Branikolog:
@oleg-derevenetz 

>No, there is such a guarantee. AI is cheating in this regard - the boat is summoned to exactly the right tile.

That's great to hear! Such micro-cheat helps balancing potential abuse of human players who can summon several boats standing on a single tile for allied heroes. 👍 

>In the same way, AI can take away a boat from this hero even if there is no boat near to the hero that casts the Summon Boat spell.

Absolutely true. But with the fix you introduced, for this particular case the hero won't steal potentially necessary boat of other hero.

I know, this is a controversial solution, but it reduces the chance of boat-stealing. At the same time it doesn't bring any guarantees, that other heroes couldn't steal the boat, which is out of the scope of this issue. Allowing AI to summon boats in general is a risky thing, as one day a strong hero could be stuck on island without a boat (and without a summoning spell) because other hero stole the boat. I believe fixing the https://github.com/ihhub/fheroes2/issues/7709 helps avoiding that. The case, when AI disembarks on a tiny island or peninsula, gathers some items and boards back is not that really a rare case, actually.

The best solution is to teach AI considering which boat he summons and make a decision whether or not should he use the spell, but I believe this is a complex task, which barely would be implemented this year, since we have lots of more important issues to fix.

oleg-derevenetz:
I recommend against using this palliative solution because its effect is largely questionable, but the computational complexity increases.

PR #7720, opened 2023-09-08
Show only skills that the hero doesn't already have in battle only skill selection list;

### LINKED ISSUE #7534: Battle only mode and maybe future editor; possibility to select same skill more than once
### Preliminary checks

### Platform

Windows

### Describe the bug

In the battle only mode it is possible to select the same skill, at the same level or different levels, more than once in the skill selection dialog. The future editor uses these dialogs so this issue will need to be solved for that too.

image

### Save file

Prohibit a skill from being applied more than once.

### Additional info

_No response_

petar-stamenkov:
Fixes #7534

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -24,6 +24,7 @@
 #include "dialog_selectitems.h"
 
 #include <algorithm>
+#include <cstddef>
 #include <cstdint>
 #include <memory>
 #include <numeric>
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -414,7 +414,7 @@ Skill::Secondary Dialog::selectSecondarySkill( const Heroes & hero, const int sk
 
     if ( result == Dialog::OK || listbox.ok ) {
         const int skillIndex = listbox.GetCurrent();
-        return { 1 + skillIndex / 3, 1 + ( skillIndex % 3 ) };
+        return { SelectEnumSecSkill::getSkillFromListIndex( skillIndex ), SelectEnumSecSkill::getLevelFromListIndex( skillIndex ) };
     }
 
     return {};
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -361,7 +361,7 @@ public:
 class SelectEnumSecSkill : public SelectEnum
 {
 public:
-    static int getSkillFromListIndex(int index)
+    static int getSkillFromListIndex( int index )
     {
         return 1 + index / 3;
     }
@@ -400,8 +400,8 @@ Skill::Secondary Dialog::selectSecondarySkill( const Heroes & hero, const int sk
 
     std::iota( skills.begin(), skills.end(), 0 );
     skills.erase( std::remove_if( skills.begin(), skills.end(),
-                                  [&hero]( const int listIndex ) { return hero.HasSecondarySkill( SelectEnumSecSkill::getSkillFromListIndex( listIndex ) ); } ), 
-                  skills.end());
+                                  [&hero]( const int listIndex ) { return hero.HasSecondarySkill( SelectEnumSecSkill::getSkillFromListIndex( listIndex ) ); } ),
+                  skills.end() );
 
     SelectEnumSecSkill listbox( { 350, fheroes2::Display::instance().height() - 200 } );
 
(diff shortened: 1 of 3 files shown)


src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -372,7 +382,7 @@ class SelectEnumSecSkill : public SelectEnum
 
     void RedrawItem( const int & index, int32_t dstx, int32_t dsty, bool current ) override
     {
-        const Skill::Secondary skill( 1 + index / 3, 1 + ( index % 3 ) );
+        const Skill::Secondary skill( getSkillFromListIndex( index ), getLevelFromListIndex( index ) );
```
Districh-ru, line 385: Hi, @petar-stamenkov, we can do the same change here:
https://github.com/ihhub/fheroes2/blob/1145386e98b2da5490995c6a4bcede9b304ec371/src/fheroes2/dialog/dialog_selectitems.cpp#L417

petar-stamenkov: Hi @Districh-ru, thanks for the suggestion! I've updated it.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -372,23 +382,26 @@ class SelectEnumSecSkill : public SelectEnum
...
 
-Skill::Secondary Dialog::selectSecondarySkill( const int skillId /* = Skill::Secondary::UNKNOWN */ )
+Skill::Secondary Dialog::selectSecondarySkill( const Heroes & hero, const int skillId /* = Skill::Secondary::UNKNOWN */ )
 {
     std::vector<int> skills( static_cast<size_t>( MAXSECONDARYSKILL * 3 ), 0 );
 
     std::iota( skills.begin(), skills.end(), 0 );
+    skills.erase( std::remove_if( skills.begin(), skills.end(),
+                                  [&hero]( const int listIndex ) { return hero.HasSecondarySkill( SelectEnumSecSkill::getSkillFromListIndex( listIndex ) ); } ),
+                  skills.end() );
```
Districh-ru, line 404: @petar-stamenkov, what do you think about slightly simplifying this logic?
Currently we make a vector filled with `0`. Then we fill it with integers from 0 to (MAXSECONDARYSKILL * 3 -1). And then we erase some integers, checking if hero has this skill.
We can fill the vector with the needed integers, doing the suggested check when filling, in example:
```cpp
    std::vector<int> skills;
    skills.reserve( MAXSECONDARYSKILL * 3 );

    for ( int i = 0; i < MAXSECONDARYSKILL * 3; ++i ) {
        if ( !hero.HasSecondarySkill( SelectEnumSecSkill::getSkillFromListIndex( i ) ) ) {
            skills.push_back( i );
        }
    }
```

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: > Hi @Districh-ru, thanks for the suggestion! I've updated it.

@petar-stamenkov, thanks for updating the `ActionListPressRight()` with `getSkillFromListIndex()` and `getLevelFromListIndex()`! I missed it.
Could you please also update the `return` in `Dialog::selectSecondarySkill()`, line 417. :)

petar-stamenkov: sorry, fixed now :)

src/fheroes2/dialog/dialog_selectitems.cpp
petar-stamenkov: That looks cleaner, thanks. I've updated it.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -372,23 +382,28 @@ class SelectEnumSecSkill : public SelectEnum
...
+        fheroes2::SecondarySkillDialogElement( Skill::Secondary( getSkillFromListIndex( index ), getLevelFromListIndex( index ) ), Heroes() ).showPopup( Dialog::ZERO );
     }
 };
 
-Skill::Secondary Dialog::selectSecondarySkill( const int skillId /* = Skill::Secondary::UNKNOWN */ )
+Skill::Secondary Dialog::selectSecondarySkill( const Heroes & hero, const int skillId /* = Skill::Secondary::UNKNOWN */ )
 {
-    std::vector<int> skills( static_cast<size_t>( MAXSECONDARYSKILL * 3 ), 0 );
+    std::vector<int> skills;
+    skills.reserve( MAXSECONDARYSKILL * 3 );
```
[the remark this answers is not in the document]
Districh-ru: We can do a static casting here to please ClangTidy: `static_cast<size_t>( MAXSECONDARYSKILL * 3 )`.
Sorry, I forgot it in my example.

Districh-ru (approved):
Well done, @petar-stamenkov!

ihhub (approved):

ihhub:
@petar-stamenkov , huge thanks for handling this issue!

PR #7724, opened 2023-09-09
Fix ? symbols appearing in single-line texts

### LINKED ISSUE #6885: Word breaks in spell names shows question marks "?"
### Preliminary checks

### Platform

Linux

### Describe the bug

When adding a line break when translating a spell name, when right clicking to view the spell description the line break will show up as a question mark (?). The Spellbook and Mage Guilds having very lite space for the spells name, so this is a problem i guess will occur for several translations. 

Another problem is that the hyphen (-) should only be shown when word is broken with the line break but not being shown on a seperate line. An added hyphen and makes sence in the Guild and the Spellbook, but instead will always be wrong in the description and the description title and vice versa if you remove the hyphen.

Should the the spells have two naming strings, one for the Guild and Spellbook and one for every other case? Or is there something else rather than **/n** to use when breaking a word in this case? 

You could ofcourse not break the words manually but then you instead have a problem that the words breaks without a hyphen and in unnatural places that just looks bad.

### Save file

Manual word break
Skärmdump_Spells

Computer word break
Skärmdump_Spells2

### Additional info

Related issue: _Summoning spells' names are too big for spell book_ https://github.com/ihhub/fheroes2/issues/556

ihhub:
relates #6885

### DIFF
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -288,6 +288,7 @@ namespace
                 continue;
             }
 
+            // TODO: remove this hack or expand it to cover more cases.
             if ( *data == lineSeparator ) {
                 // This should never happen as line cannot contain line separator in the middle.
                 // But due to some limitations in UI we have to deal with it.
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -283,15 +283,15 @@ namespace
         const uint8_t * dataEnd = data + size;
 
         for ( ; data != dataEnd; ++data ) {
+            if ( isSpaceChar( *data ) ) {
+                offsetX += spaceCharWidth;
+                continue;
+            }
+
             if ( *data == lineSeparator ) {
                 // This should never happen as line cannot contain line separator in the middle.
                 // But due to some limitations in UI we have to deal with it.
                 // The only way is to just ignore it here.
-                continue;
-            }
-
-            if ( isSpaceChar( *data ) ) {
-                offsetX += spaceCharWidth;
                 continue;
             }
 
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -283,6 +283,13 @@ namespace
         const uint8_t * dataEnd = data + size;
 
         for ( ; data != dataEnd; ++data ) {
+            if ( *data == lineSeparator ) {
+                // This should never happen as line cannot contain line separator in the middle.
+                // But due to some limitations in UI we have to deal with it.
+                // The only way is to just ignore it here.
+                continue;
+            }
+
             if ( isSpaceChar( *data ) ) {
                 offsetX += spaceCharWidth;
                 continue;


ihhub:
@zenseii , done.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -283,6 +283,13 @@ namespace
...
 
         for ( ; data != dataEnd; ++data ) {
+            if ( *data == lineSeparator ) {
+                // This should never happen as line cannot contain line separator in the middle.
+                // But due to some limitations in UI we have to deal with it.
+                // The only way is to just ignore it here.
+                continue;
+            }
+
             if ( isSpaceChar( *data ) ) {
```
Districh-ru, line 293: Space characters are more often used in phrases than line separators.
If we place this check before the check for the `lineSeparator` should it optimize the loop, what do you think?

Districh-ru (commented):
Hi, @ihhub, I left one question. Could you please check if it is valuable.

In future, if it is needed, we can expand the logic to check the previous and next characters if line separator is present to:
- remove '-' , e.g. in "Necro-\nmancer";
- add space where needed, e.g. in "New\nGame".
Maybe we should leave a TODO?

src/fheroes2/gui/ui_text.cpp
ihhub: Sure. I can do this.

ihhub:
> Hi, @ihhub, I left one question. Could you please check if it is valuable when you have time.
> 
> In future, if it is needed, we can expand the logic to check the previous and next characters if line separator is present to:
> 
> * remove '-' , e.g. in "Necro-\nmancer";
> * add space where needed, e.g. in "New\nGame".
>   Maybe we should leave a TODO?

I don't fully understand what exactly I need to check. Is it related to #7726?

ihhub:
@Districh-ru , I think we need to deal with usage of new classes first and also use multi-line text over single line where it is needed.

Districh-ru:
> I don't fully understand what exactly I need to check. Is it related to #7726?

I'll check #7726 next.
I only mentioned that if we remove the line separator it will not make the separated phrase look good in all cases. And we may address these cases in the future if we want to deal with multi-line texts in a single line, in example not to make "Necro-mancer" from "Necro-\nmancer" and "NewGame" from "New\nGame".
This is beyond this PR. :)

Districh-ru (approved):

ihhub:
@zenseii , this logic of line breaking is implemented in #7726.

Districh-ru:
Talking about the need of placing space in the place of line separator. After removing the '?' signs we will get not correct words here:
изображение

When previous to the separator character is not ( space or dash ) and next character is not space - we should put a space instead of the line separator.

ihhub:
@Districh-ru , I think this part should be fixed from the translation point of view. We should use a patchy code to fix translations. Translation text can be adjusted to accommodate such situations.

ihhub:
On the other hand, @Districh-ru idea might be valid.

Districh-ru:
> @Districh-ru , I think this part should be fixed from the translation point of view. We should use a patchy code to fix translations. Translation text can be adjusted to accommodate such situations.

You are right, but we started to fix it (line separators in one line text) in code and can do it a little better. :)
The suggested approach should be placed after the `*data == lineSeparator` check.
Tn this case we should also check that `data` is not the first and the last character in the phrase.

Districh-ru:
@ihhub, I tried this code:
```cpp
if ( ( data != dataEnd - size ) && ( data != dataEnd - 1 ) && *( data - 1 ) != 0x20 && *( data - 1 ) != '-' && *( data + 1 ) != 0x20 ) {
                    offsetX += spaceCharWidth;
                }
```

There are no spaces and dashes near the line separator:
изображение

There is a dash near the line separator:
изображение

Feel free to use it if you want. :)

ihhub:
Hi @Districh-ru , while the current PR brings almost no impact in the existing string with the proposed by you changes we are literally alternating some information in strings. To be honest even this PR is a hack since we don't display the text as it should be: multi-line text.

Districh-ru:
> Hi @Districh-ru , while the current PR brings almost no impact in the existing string with the proposed by you changes we are literally alternating some information in strings. To be honest even this PR is a hack since we don't display the text as it should be: multi-line text.

I used this code inside the `if ( *data == lineSeparator ) { ... }` condition and it has no impact to normal strings, without `\n`.
I just suggested a possible solution for this current, not ordinary case. :)
And I agree that we shouldn't do or do only minimal hacks.
Perhaps in the future we should split some phrases into single-line and multi-line, like we did for factions.

ihhub:
@Districh-ru , splitting text into single line and multi line is better solution.

PR #7725, opened 2023-09-09
Optimize `getAroundIndexes()` function

Districh-ru:
This PR optimizes the calculation of `getAroundIndexes()` function, which is used for many in-game actions (e.g. checking if we need to draw the attack cursor near the monster):
- cache some variables;
- remove the `isValidAbsPoint()` calls in the loop;
- do not do loops for tiles out of map boundaries (check it before starting the loops).

Now this function in future can be used with non-square maps, previously it used map width as map height.


ihhub (approved):

ihhub:
@Districh-ru , thank you very much for this optimization!

PR #7729, opened 2023-09-09
Code cleanup for SDL 2 cursor class

ihhub:
Leftover code from SDL 1 removal.

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -543,12 +543,7 @@ namespace
         RenderCursor()
             : _cursor( nullptr )
         {
-            _emulation = false;
-
-            const int returnCode = SDL_ShowCursor( _show ? SDL_ENABLE : SDL_DISABLE );
-            if ( returnCode < 0 ) {
-                ERROR_LOG( "Failed to set cursor state. The error value: " << returnCode << ", description: " << SDL_GetError() )
-            }
+            // Do nothing.
         }
 
     private:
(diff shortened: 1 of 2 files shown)


src/engine/screen.cpp
```diff
@@ -543,12 +543,7 @@ namespace
         RenderCursor()
             : _cursor( nullptr )
         {
-            _emulation = false;
-
-            const int returnCode = SDL_ShowCursor( _show ? SDL_ENABLE : SDL_DISABLE );
-            if ( returnCode < 0 ) {
-                ERROR_LOG( "Failed to set cursor state. The error value: " << returnCode << ", description: " << SDL_GetError() )
-            }
```
oleg-derevenetz, line 551: Hi @ihhub why we are removing this code? `_show` is `false` by default, shouldn't we immediately synchronize the SDL cursor state with this setting?

ihhub: We have SDL_ShowCursor() call in fheroes2.cpp which disables the cursor.

The code in this constructor was useful for SDL1 because cursor was independent on SDL window. In SDL2 we have to create a window first and then set cursor. However, at the time of calling this constructor a window doesn't exist but it does when we call SDL_ShowCursor() in fheroes2.cpp.

ihhub:
I'm marking it for 1.0.9 release as the change might break cursor behavior on some platforms.

ihhub:
Closing this pull request as it brings no real value at the moment.

PR #7731, opened 2023-09-09
Bump the timeout for SonarCloud Analyzer workflow to 60 minutes

oleg-derevenetz:
Our codebase is gradually growing (and Microsoft may be gradually getting carried away with saving money), and, when changing any popular header file that is included in many project files, or during the recurrent upgrade of GitHub Ubuntu runner images (with some newer compiler) the time for a full SonarCloud analysis is dangerously close to 30 minutes.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you for the change! Interesting is that we removed SDL1 and the pipeline in general should be faster as less runner are being used.

PR #7739, opened 2023-09-10
Improve path calculations by AIWorldPathfinder in planning mode

oleg-derevenetz:
In the `master` branch, when planning long-distance travel both by land and by water, `AIWorldPathfinder` uses the maximum number of movement points corresponding to the type of surface on which the hero is currently located. That is, if hero is currently located on land, pathfinder will use the maximum number of movement points corresponding to the hero on land all the way, even when hero is already moving on the water - and, accordingly, on the contrary: if the hero is on the water, pathfinder will be a bit too optimistic about his movements on the ground. This can lead to the laying of less optimal paths, which is intended to be corrected by this PR by introducing a separate account of the maximum number of movement points on land and on water for the AI in the planning mode.


ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for this nice improvements!

PR #7751, opened 2023-09-14
Rework game settings on Adventure Map

ihhub:
- avoid saving into configuration file within Main Menu settings (relates to #6681). A slight improvement in speed after closing the language window.
- move all Interface related settings into a separate window
- allow changing game's language on the Adventure Map. This is much more convenient to do rather going to the Main Menu.
- allow changing game's resolution on the Adventure Map. Changing resolution is the most important on the Adventure Map.

This is how System's Options window look like after the changes:
image

This is the new Interface window:
image


ihhub:
Hi @zenseii , could you please this window's text:
image

?

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -0,0 +1,310 @@
...
+            if ( le.MouseClickLeft( windowInterfaceTypeRoi ) ) {
+                return SelectedWindow::InterfaceType;
+            }
+            if ( le.MouseClickLeft( windowInterfacePresenceRoi ) ) {
+                return SelectedWindow::InterfacePresence;
+            }
+            if ( le.MouseClickLeft( windowCursorTypeRoi ) ) {
+                return SelectedWindow::CursorType;
+            }
+            if ( le.MouseClickLeft( windowScrollSpeedRoi ) ) {
```
Districh-ru, line 216: I'm a little concerned that to simply change the scroll speed or cursor color we close this dialog, restore the image below it, and then completely redraw it and initialize all the restorers. Here we could only restore `emptyDialogRestorer` and call `drawOptions();` with `display.render()`. It applies to other dialogs with the same options (hero speed, quick-battle, ...).
But since the settings are changed not often and not much we can leave it as is. :)
So this is an optional comment. We could leave a TODO to optimize this part in the future if this optimization is needed. What do you think?

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -221,9 +221,9 @@ namespace
 
 namespace fheroes2
 {
-    void openGraphicsSettingsDialog()
+    void openGraphicsSettingsDialog( const std::function<void()> & updateUI )
     {
-        drawMainMenuScreen();
+        updateUI();
```
Districh-ru, line 226: This UI update is probably redundant here: the previous settings dialog is closed after every action and its restorer restores the pure image without any dialogs.

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+        return SelectedWindow::Exit;
+    }
+}
+
+namespace fheroes2
+{
+    void openInterfaceSettingsDialog( const std::function<void()> & updateUI )
+    {
+        updateUI();
```
Districh-ru, line 264: The same comment about the redundant UI update: the action in the previous dialog closes it and restores the screen image.

Districh-ru (changes requested):
Hi, @ihhub.
I like the changes in settings dialogs! Now we can change game resolution and language without going back to the main menu.
I left some suggestions regarding rendering optimizations. Could you please take a look in your spare time.

And maybe we could add hero speed change by mouse wheel, like it is done for the scroll speed and sound volume. What do you think? :)

src/fheroes2/dialog/dialog_interface_settings.cpp
ihhub: It is fixed.

Branikolog (approved):

mihaimaruseac (approved):

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -0,0 +1,317 @@
...
+            drawCursorType( windowCursorTypeRoi );
+            drawScrollSpeed( windowScrollSpeedRoi );
+        };
+
+        drawOptions();
+
+        const auto refreshWindow = [&drawOptions, &emptyDialogRestorer, &display]() {
+            emptyDialogRestorer.restore();
+            drawOptions();
+            display.render();
```
Districh-ru, line 188: In this case we do not need to render the full screen. We may render only the changed area:
```suggestion
            display.render( emptyDialogRestorer.rect() );
```

src/fheroes2/dialog/dialog_system_options.cpp
```diff
@@ -214,129 +208,140 @@ namespace
...
+            drawTextSupportModeOptions( windowTextSupportModeRoi );
+            drawBattles( windowBattlesRoi );
+        };
+
+        drawOptions();
+
+        const auto refreshWindow = [&drawOptions, &emptyDialogRestorer, &display]() {
+            emptyDialogRestorer.restore();
+            drawOptions();
+            display.render();
```
Districh-ru, line 247: The same comment as above.
```suggestion
            display.render( emptyDialogRestorer.rect() );
```

Districh-ru (changes requested):
Hi, @ihhub, I left two similar comments regarding minimization of render area on dialog contents update. Could you take a look when you have time.

Districh-ru (approved):
Well done!
I really like the reworked game settings dialogs!

PR #7752, opened 2023-09-14
Use Home and End hotkeys to navigate among items within a list

ihhub:
To quickly jump to the beginning or end of a list.


Districh-ru (approved):
👍

PR #7753, opened 2023-09-15
Remove old Dialog::Message usage from the code

ihhub:
We need to standardize the code for the future changes as it is hard to do when we still use old and new code all over the place. This pull request replaces the usage of `Dialog::Message` by  `fheroes2::showStandardTextMessage`.

Additionally, there some places with `fheroes2::showStandardTextMessage` where Font::BIG was using causing unexpected appearance of buttons.

All changes are divided into 2 categories:
- if an old `Dialog::Message` call ended with Font::BIG, the the new call should end with Dialog::ZERO
- if an old `Dialog::Message` call ended with Dialog::SOMETHING, then the new call should remove Font::BIG before Dialog::SOMETHING


Districh-ru (approved):
I looked through the code changes, tested some of the changed messages and at first glance found no bugs.
The next will be the replacement of `Text` with `fheroes2::Text`? :)

Issue #7460, opened 2023-07-23
Stop background color cycling to improve performance


drevoborod:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

When any dialogue window is open, background colour cycling is not being stopped. It leads to unnecessary resources consumption. Also, in the areas which are occupied by the shadow the cycling is being stopped, so part of background water under the shadow is stopped, and outside of the shadow water is moving. It looks weird. 

https://github.com/ihhub/fheroes2/assets/10282847/82758218-50c9-4a6b-bd26-4c72093d6b14

### Describe the possible solution

I suggest to stop background colour cycling when any dialogue is open.

ihhub:
Hi @drevoborod , there could be monsters which require cycling animation such as Phoenix or Titans so we cannot simply stop animation for them.

drevoborod:
> On the other hand there is a disconnect between only colors cycling, like the sea, and no other objects animating.

And the shadow, as I mentioned before. It's a little strange when a water under the shadow is not animated but all other water moves.

> there could be monsters which require cycling animation such as Phoenix or Titans so we cannot simply stop animation for them.

That's why I'm talking about only **background** color cycling. Maybe there is a way to separate background and foreground cycling?

drevoborod:
https://github.com/ihhub/fheroes2/assets/10282847/46f4e05e-bbe7-4c1b-9434-87d0d0b41cf5

drevoborod:
https://github.com/ihhub/fheroes2/assets/10282847/5eaca21a-5eba-47ea-b8b0-aeb03bfcb923

ihhub:
I am closing this issue as performance for such cases is not a problem for the engine.

PR #7757, opened 2023-09-15
Rework pre- and post-rendering code and remove TextSprite class

### LINKED ISSUE #1689: Hero screen, info bar text at the bottom of the screen, text too long when changing the place of artifacts
In fheroes2, the text in the info bar at the bottom of the screen can sometimes be a bit long when the player wants to change the place of artifacts.
The sentence "Exchange the place of artifacts" would be sufficient to avoid a too long text with long names of artifacts.

In fheroes2:

Info bar Artifact text too long fh2 01

ihhub:
Here are few things this pull request handles:
- status bar does not allow to render beyond specified width (close #1689)
- SystemInfoRenderer is not a global variable anymore
- replace TextSprite class by MovableText
- replace ColorCycling class by RenderProcessor class as the first one did not belong to localevent scope
- speed up rendering related to color cycling by adding a time range (related to #7460)


src/engine/localevent.h
```diff
@@ -278,7 +278,7 @@ class LocalEvent
 
     static int32_t getCurrentKeyModifiers();
 
-    static void RegisterCycling( void ( *preRenderDrawing )() = nullptr, void ( *postRenderDrawing )() = nullptr );
+    static void RegisterCycling( const std::function<void()> & preRenderDrawing = nullptr, const std::function<void()> & postRenderDrawing = nullptr );
```
oleg-derevenetz, line 281: We can use `= {}` here instead of `= nullptr`, it's the same: both create an empty function.

src/fheroes2/game/fheroes2.cpp
```diff
@@ -178,7 +178,9 @@ namespace
             SDL_ShowCursor( SDL_DISABLE ); // hide system cursor
 
             // Initialize local event processing.
-            LocalEvent::RegisterCycling( fheroes2::PreRenderSystemInfo, fheroes2::PostRenderSystemInfo );
+            _systemInfoRenderer = std::make_unique<fheroes2::SystemInfoRenderer>();
+
+            LocalEvent::RegisterCycling( [this]() { _systemInfoRenderer->preRender(); }, [this]() { _systemInfoRenderer->postRender(); } );
```
oleg-derevenetz, line 183: I recommend to perform this registration in the `SystemInfoRenderer`'s constructor and also unregister these methods in the `SystemInfoRenderer`'s destructor to ensure that these methods will not be called when it is destroyed. Also you can use the pointer to member function here instead of lambda using `std::bind()`, but that's optional.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a couple of suggestions, please take a look when you have time.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -243,6 +169,96 @@ namespace fheroes2
...
+    SystemInfoRenderer::SystemInfoRenderer()
+        : _startTime( std::chrono::steady_clock::now() )
+        , _text( fheroes2::Display::instance() )
+    {
+        LocalEvent::RegisterCycling( [this]() { preRender(); }, [this]() { postRender(); } );
+    }
+
+    SystemInfoRenderer::~SystemInfoRenderer()
+    {
+        LocalEvent::RegisterCycling( {}, {} );
```
oleg-derevenetz, line 212: `ColorCycling::registerDrawing()` does nothing in this case, it never unregisters the handlers.

src/engine/localevent.h
```diff
@@ -278,14 +278,14 @@ class LocalEvent
...
-    static void RegisterCycling( void ( *preRenderDrawing )() = nullptr, void ( *postRenderDrawing )() = nullptr );
+    static void RegisterCycling( const std::function<void()> & preRenderDrawing, const std::function<void()> & postRenderDrawing );
 
     // These two methods are useful for video playback
     static void PauseCycling();
 
     static void ResumeCycling()
     {
-        RegisterCycling();
+        RegisterCycling( {}, {} );
```
oleg-derevenetz, line 288: Since `ColorCycling::registerDrawing()` should unregister the handlers in this case, I suppose this method should do something different.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few comments here, please take a look when you have time.

src/fheroes2/gui/ui_tool.cpp
ihhub: Yes, you are right. The code is a mess here.

src/fheroes2/gui/statusbar.cpp
```diff
@@ -22,18 +22,40 @@
...
+        // No updates.
+        return;
+    }
+
+    _prevMessage = msg;
+
+    auto text = std::make_unique<fheroes2::Text>( std::move( msg ), fheroes2::FontType::normalWhite() );
+    if ( text->width() > _roi.width ) {
+        text->fitToOneRow( _roi.width );
     }
```
Districh-ru, line 51: `fitToOneRow(()` already compares text width and `_roi.width` and returns from the function if `originalTextWidth <= maxWidth`.
IMHO this condition is redundant and we can leave only `text->fitToOneRow( _roi.width );`. If the text fits the `_roi.width` the `fitToOneRow()` will do nothing to it.

src/fheroes2/gui/statusbar.cpp
```diff
@@ -22,18 +22,40 @@
...
+
+    const fheroes2::Rect messageRoi{ _roi.x + ( _roi.width - text->width() ) / 2, _roi.y, text->width(), text->height() };
```
Districh-ru, line 53: We can cache the `text->width()` because it is used two times here.

Districh-ru (changes requested):
Hi @ihhub. I left two comments. Could you please take a look when you have time.

src/engine/screen.cpp
```diff
@@ -1385,18 +1385,14 @@ namespace fheroes2
             // Previous position of cursor must be updated as well to avoid ghost effect.
             _renderFrame( getBoundaryRect( temp, _prevRoi ) );
 
-            if ( _postprocessing != nullptr ) {
-                _postprocessing();
-            }
+            _postprocessing();
```
oleg-derevenetz, line 1388: The same question - if `_postprocessing` is not set, we will get the `std::bad_function_call` here.

src/engine/screen.cpp
```diff
@@ -1385,18 +1385,14 @@ namespace fheroes2
...
 
             Copy( backup, 0, 0, *this, backup.x(), backup.y(), backup.width(), backup.height() );
         }
         else {
             _renderFrame( getBoundaryRect( temp, _prevRoi ) );
 
-            if ( _postprocessing != nullptr ) {
-                _postprocessing();
-            }
+            _postprocessing();
```
oleg-derevenetz, line 1395: The same question - if `_postprocessing` is not set, we will get the `std::bad_function_call` here.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -243,6 +170,87 @@ namespace fheroes2
...
+            _restorer.restore();
+            _isHidden = true;
+        }
+    }
+
+    SystemInfoRenderer::SystemInfoRenderer()
+        : _startTime( std::chrono::steady_clock::now() )
+        , _text( fheroes2::Display::instance() )
+    {
+        RenderProcessor::instance().registerRenderers( [this]() { preRender(); }, [this]() { postRender(); } );
```
oleg-derevenetz, line 208: Shouldn't we unregister them in the destructor?

oleg-derevenetz (commented):

Districh-ru (approved):

oleg-derevenetz (approved):
Hi @ihhub what bothered me the most was that the pre/post-renderers should be registered and unregistered in one place, otherwise, later the logic will be unclear and it will be difficult to maintain, so I made the appropriate changes. Plus small changes that limit the ability to create and copy certain classes. Please take a look when you have time.

ihhub:
@oleg-derevenetz , your changes look great. Thank you!

PR #7760, opened 2023-09-15
Update Russian translation

Districh-ru:
- Removed `#, fuzzy` marks added after  #7755 because the translated text does not have double spaces.
- Corrected the translation for the Steelskin spell description.
- Updated the copyright year ... after 9 months. :)

### DIFF
--- a/files/lang/ru.po
+++ b/files/lang/ru.po
@@ -1,5 +1,5 @@
 # Russian translation of fheroes2
-# Copyright (C) 2021 - 2022 fheroes2 team <fhomm2@gmail.com>
+# Copyright (C) 2021 - 2023 fheroes2 team <fhomm2@gmail.com>
 # Copyright (C) 2009 Igor Orlov <igororlov@inbox.ru>
 # Copyright (C) 2009 Andrey Afletdinov <afletdinov@gmail.com>
 # This file is distributed under the same license as the fheroes2 package.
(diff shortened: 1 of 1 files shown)


ihhub (approved):

Branikolog:
Hi, @Districh-ru !
Why have you decided to take away amount of defense increased by these spells? I find this information useful. The same as the exact damage dealed by attacking spells or the number of HP restored by Cure or Resurrection spells which were added in the descriptions over the original.

Districh-ru:
> Hi, @Districh-ru ! Why have you decided to take away amount of defense increased by these spells? I find this information useful. The same as the exact damage dealed by attacking spells or the number of HP restored by Cure or Resurrection spells which were added in the descriptions over the original.

Hello, @Branikolog! 
I did this to make the translation more consistent with the original text. :)
But we may keep the amount info. I reverted the these changes in the translation.
If we think that this information is useful in spell descriptions than probably we should add it to the original English text. What do you think? Was it discussed with the team?

Branikolog:
Hi, @Districh-ru 

>If we think that this information is useful in spell descriptions than probably we should add it to the original English text. What do you think? Was it discussed with the team?

I don't remember we had such discussion.
English version doesn't have such description indeed. I find it really useful to know, what is the exact values of these spells.

Branikolog (approved):

ihhub:
@Districh-ru , thank you very much for the update!

PR #7763, opened 2023-09-16
Hero related code cleanup

ihhub:
- do not use empty structures **KingdomCastles** and **KingdomHeroes** as they were doing literally nothing. It is hard to handle different variable types.
- change gold income logic for AI. It is better to avoid any floating point calculation errors
- do some code clean up in Heroes class to make the code human readable

### DIFF
--- a/src/fheroes2/game/difficulty.h
+++ b/src/fheroes2/game/difficulty.h
@@ -40,7 +40,9 @@ namespace Difficulty
     std::string String( int );
 
     int GetScoutingBonus( int difficulty );
-    double GetGoldIncomeBonus( int difficulty );
+
+    // Returns an extra gold bonus modifier for AI based on difficulty level.
+    double GetGoldIncomeBonusForAI( const int difficulty );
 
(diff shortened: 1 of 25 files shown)


oleg-derevenetz (approved):

src/fheroes2/game/difficulty.h
```diff
@@ -40,7 +40,9 @@ namespace Difficulty
     std::string String( int );
 
     int GetScoutingBonus( int difficulty );
-    double GetGoldIncomeBonus( int difficulty );
+
+    // Returns an extra gold bonus modifier for AI based on difficulty level.
+    double GetGoldIncomeBonusForAI( const int difficulty );
```
Districh-ru, line 45: Since we are renaming this function we can name it from the lower character.
```suggestion
    double getGoldIncomeBonusForAI( const int difficulty );
```

Districh-ru (changes requested):
Hi, @ihhub. I left one minor comment about function naming. Could you please take a look in your spare time.

Districh-ru (approved):

PR #7766, opened 2023-09-16
Improve AI behavior for Ghosts during battle

### LINKED ISSUE #7685: Improve AI ghosts' target priorities
### Preliminary checks

This issue was reported by several users of fheroes2.

There are still obvious cases when the ghosts do not choose to attack a stack which would grant them immediate higher numbers (aka peasants).

One such reproducible example is if you pit 1 wolf and 4 peasants against 1 ghost. The ghosts will attack the wolves first for whatever reason I can't check right now:
Screenshot_20230903_101409_fheroes2

In general, make sure ghosts prioritize more increasing their numbers.

Although there could be cases when purely basing their decision on this isn't the best choice, for example against shooters with peasants in their army.

### Additional info

_No response_

ihhub:
Since Ghosts have Soul Eater ability it is important to calculate the possible growth of their numbers after the attack.

close #7685

### DIFF
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -1104,7 +1104,8 @@ int32_t Battle::Unit::GetScoreQuality( const Unit & defender ) const
     const double attackerPowerLost = ( attacker.Modes( CAP_MIRRORIMAGE ) || defender.Modes( CAP_TOWER ) || defendersDamage >= hp ) ? 1.0 : defendersDamage / hp;
     const bool attackerIsArchers = isArchers();
 
-    double attackerThreat = CalculateDamageUnit( defender, ( static_cast<double>( GetDamageMin() ) + GetDamageMax() ) / 2.0 );
+    const uint32_t initialDamage = CalculateDamageUnit( defender, ( static_cast<double>( GetDamageMin() ) + GetDamageMax() ) / 2.0 );
+    double attackerThreat = initialDamage;
 
     if ( !canReach( defender ) && !defender.Modes( CAP_TOWER ) && !attackerIsArchers ) {
         // Can't reach, so unit is not dangerous to defender at the moment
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):
While this approach seems to work just fine, it looks quite funny from the logical point of view :) As far as I'm familiar with the logic of AI target selection, AI-controlled unit tends to attack the enemy unit that he considers the most dangerous for himself. So in this case the more units in the enemy stack are killed during the attack, the more "dangerous" this enemy stack is considered :)

ihhub:
Hi @oleg-derevenetz , yeah. I couldn't find a better place to implement it plus we have to do the same for Lord Vampires.

Branikolog:
Hi, @ihhub 
I'm really enjoying such improvements, when AI becomes smarter in battles. 

I've noticed one issue here. It seems that AI is too much obsessed with "eating" weaker creatures, while he could be much more effective targeting strong creatures.

https://github.com/ihhub/fheroes2/assets/55348946/8c43c3da-12e2-45e9-b8b8-7df3dbc21b2c

400 ghosts could easily beat the titans here, no matter, who is the first target.
But in more equal circumstances the result of the battle can change because of this obsession:

https://github.com/ihhub/fheroes2/assets/55348946/4b72fb14-e2cf-4b33-8299-54c0b0b974c1

If ghosts focus Titans in this situation, they could win the battle.

https://github.com/ihhub/fheroes2/assets/55348946/51d69e12-5768-4079-836d-05c6e7056769

Even with one additional attack made by Titans because of high morale triggered.

ihhub:
Hi @Branikolog , what you are asking is to have a prediction ability for the AI. At the moment such logic has not been implemented and the AI makes decisions only based on the best outcome for the current turn.

Branikolog:
@ihhub 

Implemented behaviour could be easily abused: I can simply take stack of Titans and 4 stacks of peasants. While ghosts would be eating peasants, titans can shoot several times.

I'm proposing to not only teach the AI to eat as much creatures, as he can (10 peasants vs 1 Titan), but to consider also the amount of HP ghosts reduce with their attack (10HP of peasants is **much** lower, than 300+ HP of demolished titan).

Attacking 10 peasants with 400 ghosts is not really a good decision, while ghosts can hit for more than 600HP other stack of creatures (titans).

Issue #7765, opened 2023-09-16
Touch screen controls


Novicek:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

While playing on my tablet I get mostly comfortable experience, there are some issues I'd like being looked at:

1. "Right click" functionality - current implementation is very clunky in having to do a two finger gesture and then moving remaining finger to be able to read a tooltip window.
2. "Spacebar" functionality - the ability to interact with an object a hero is standing on has been added in fheroes2 and is doable on keyboard by pressing spacebar. However, on touch devices, I haven't found a way to achieve this.
3. Seeing cursor in combat - This one is probably the biggest change. Currently, it is very difficult to check a cursor for the angle of attack or ranged damage penalty, since when you remove the finger from the screen, the action will be executed.

### Describe the possible solution

1. It's much more common and comfortable for tooltips and such to be shown after long pressing on a touch device rather than doing the two finger gesture. This would be applicable on most screens. The projected window should also not appear directly on the tile/space you touched, but should render towards the center of the screen with only the corner of the window touching where you pressed with your finger. This prevents the need to move your finger away to actually read the tooltip.
2. Said two finger gesture could be used for interacting with objects your hero is standing on instead.
3. My proposed change would be to make actions need confirmation by double clicking. First click would only set the cursor where you pressed with you being able to remove finger and check for angle of attack/ranged penalty (broken arrow cursor) but not executing the action. After confirming, if you press on the space which would result in the same action, it would get executed. In my mind this still isn't an ideal solution but it would work together with keeping long press on showing tooltip window and being able to see your own cursor.

oleg-derevenetz:
> "Right click" functionality - current implementation is very clunky in having to do a two finger gesture and then moving remaining finger to be able to read a tooltip window.

"You are doing it wrong" (C). You don't have to move any finger anywhere. That's how it is described in the README for Android:

> To simulate a right-click to get info on various items, you need to first touch and keep touching on the item of interest
and then touch anywhere else on the screen. You can then remove your first finger from the screen and keep viewing the info
on the item.

It's quite convenient.

Novicek:
> > "Right click" functionality - current implementation is very clunky in having to do a two finger gesture and then moving remaining finger to be able to read a tooltip window.
> 
[...]

PR #7767, opened 2023-09-16
Add the ability to confirm actions in battle for touch devices

oleg-derevenetz:
Related to #7765

The ability to confirm actions is supported for moves, melee attacks, shots, and spell casts.

It would be good to check its work on desktop devices (with mouse), on touch-only devices AND on touch devices with the mouse (it should behave adequately and not require unnecessary clicks when performing operations with the mouse, as well as in the case when pointing at the cell is performed with the mouse, but triggering the action is performed with the touch). Also it would be good to check its work on PSV and Switch.

### DIFF
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -509,9 +509,8 @@ namespace Battle
 
             ~BoardActionIntentUpdater()
             {
-                // Do not remember intermediate touch gestures (such as simulated mouse button pressing) as intents. When using the touchpad,
-                // only a complete simulated click is considered an intent.
-                if ( _isFromTouchpad && !_isClick ) {
+                // Do not remember intermediate touch gestures as intents
+                if ( _isFromTouchpad ) {
                     return;
                 }
 
@@ -519,11 +518,6 @@ namespace Battle
             }
 
             BoardActionIntentUpdater & operator=( const BoardActionIntentUpdater & ) = delete;
-
-            void setClick()
-            {
-                _isClick = true;
-            }
 
             void setIntent( const BoardActionIntent & intent )
             {
@@ -540,7 +534,6 @@ namespace Battle
         private:
             Interface & _interface;
             const bool _isFromTouchpad;
-            bool _isClick{ false };
             std::optional<BoardActionIntent> _intent;
         };
     };
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -398,8 +398,8 @@ namespace Battle
         void EventShowOptions();
         void ButtonAutoAction( const Unit & unit, Actions & actions );
         void ButtonSettingsAction();
-        void ButtonSkipAction( Actions & acrions );
-        void MouseLeftClickBoardAction( const int themes, const Cell & cell, Actions & actions );
+        void ButtonSkipAction( Actions & actions );
+        void MouseLeftClickBoardAction( const int themes, const Cell & cell, const bool isConfirmed, Actions & actions );
         void MousePressRightBoardAction( const Cell & cell ) const;
 
         int GetBattleCursor( std::string & statusMsg ) const;
@@ -483,6 +483,20 @@ namespace Battle
         // this may be needed in the future (for example, in expansion) to display some sprites over
         // troops for some time (e.g. long duration spell effects or other permanent effects).
         std::vector<UnitSpellEffectInfo> _unitSpellEffectInfos;
+
+        struct BoardActionIntent
+        {
+            int cursorTheme = Cursor::NONE;
+            int32_t cellIndex = -1;
+
+            bool operator==( const BoardActionIntent & other ) const
+            {
+                return cursorTheme == other.cursorTheme && cellIndex == other.cellIndex;
+            }
+        };
+
+        // Intents are used to confirm actions in combat performed using touch gestures
+        BoardActionIntent _boardActionIntent;
     };
 }
 
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -496,6 +496,53 @@ namespace Battle
 
         // Intents are used to confirm actions in combat performed using touch gestures
         BoardActionIntent _boardActionIntent;
+
+        class BoardActionIntentUpdater
+        {
+        public:
+            BoardActionIntentUpdater( Interface & interface, const bool isFromTouchpad )
+                : _interface( interface )
+                , _isFromTouchpad( isFromTouchpad )
+            {}
+
+            BoardActionIntentUpdater( const BoardActionIntentUpdater & ) = delete;
+
+            ~BoardActionIntentUpdater()
+            {
+                // Do not remember intermediate touch gestures (such as simulated mouse button pressing) as intents. When using the touchpad,
+                // only a complete simulated click is considered an intent.
+                if ( _isFromTouchpad && !_isClick ) {
+                    return;
+                }
+
+                _interface._boardActionIntent = _intent.value_or( BoardActionIntent{} );
+            }
+
+            BoardActionIntentUpdater & operator=( const BoardActionIntentUpdater & ) = delete;
+
+            void setClick()
+            {
+                _isClick = true;
+            }
+
+            void setIntent( const BoardActionIntent & intent )
+            {
+                _intent = intent;
+            }
+
+            bool isConfirmed()
+            {
+                // If the mouse event has been triggered by the touchpad, it should be considered confirmed only if this event orders to
+                // perform the same action that is already indicated on the battle board with the mouse cursor.
+                return ( !_isFromTouchpad || _interface._boardActionIntent == _intent );
+            }
+
+        private:
+            Interface & _interface;
+            const bool _isFromTouchpad;
+            bool _isClick{ false };
+            std::optional<BoardActionIntent> _intent;
+        };
     };
 }
 
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -33,6 +33,7 @@
 
 #include "battle_animation.h"
 #include "battle_board.h"
+#include "cursor.h"
 #include "dialog.h"
 #include "icn.h"
 #include "image.h"
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -500,8 +500,8 @@ namespace Battle
         class BoardActionIntentUpdater
         {
         public:
-            BoardActionIntentUpdater( Interface & interface, const bool isFromTouchpad )
-                : _interface( interface )
+            BoardActionIntentUpdater( BoardActionIntent & storedIntent, const bool isFromTouchpad )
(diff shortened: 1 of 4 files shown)


Branikolog:
Hi, @oleg-derevenetz 

I've made a couple of tests on my phone with touchscreen only.
If we press on any cell, it stayed selected waiting for you to confirm an action by touching it once again. But if we (accidentally or on demand) touch other place out of battle cells range and then return to the first cell - the action proceeds at once without confirmation touch.

https://github.com/ihhub/fheroes2/assets/55348946/3b46e1bf-d538-415e-a6fb-68bcee2d7a3e

Actually, after playing a couple of battles I can say, that this implementation makes experience a little bit better. Just a little bit. But not worse for sure!
I had to press the same cell one more time to initiate the action I want. So I can still miss the attacking direction. I press once -> take away finger to make sure, I chose the right direction -> and then touch this place again. This implementation will work much better if there was some sort of confirmation button, dialog or touchscreen gesture, that won't require covering the target with my finger.

_Just to share my experience and my thoughts on the problem:_

In some other mobile games I've seen another solution for such precise control cases, when player should press on the screen to make an action menu appear around the selected cell. Player needs to choose the action (attack angle in our case) and swing to the action he wants and then release the finger. But I'm afraid it requires an enormous amount of work. :(

Also, if the engine had an option of interacting with some sort of virtual joystick and buttons, battles wouldn't be such an issue at all. With one finger we can move cursor over the game (by touching the area of interest or moving cursor with the help of joystick) and the other finger is used press right or left virtual mouse button (placed somewhere near the edge of the screen.

oleg-derevenetz:
@Branikolog

> But if we (accidentally or on demand) touch other place out of battle cells range and then return to the first cell - the action proceeds at once without confirmation touch.

Yes, this needs to be fixed. 

> So I can still miss the attacking direction.

No, you cannot. If attack direction changes, you should confirm it once again. After a new touch, both the cell and the cursor should be the same as after the first touch to proceed.

Districh-ru (commented):
Hi, @oleg-derevenetz, I like this new ability, but but I ran into the issue found by @Branikolog, along with a slight modification of it:

https://github.com/ihhub/fheroes2/assets/113276641/c14ed31a-9c9a-4039-8791-9fffed516f1e

If you start pointing any cell and move the cursor to the cursor to some other cell the action will be performed to the last pointed cell.
This is because `MouseLeftClickBoardAction()` is called for the cell where the click was ended.

This issue was already present in the engine and can be reproduced by mouse: Press right mouse button on any cell and move to the other, release the right mouse button and an action will be applied.
It can be addressed in the new PR.

<details><summary>A possible approach to solve it might be:</summary>

```cpp
            if ( !_mousePressed && le.MousePressLeft() ) {
                _mousePressIndex = index_pos;
                _mousePressed = true;
            }
            else if ( le.MouseClickLeft() ) {
                _mousePressed = false;
                if ( _mousePressIndex == index_pos ) {
                    MouseLeftClickBoardAction( themes, *cell, isConfirmed, actions );
                }
            }
```

</details>

oleg-derevenetz:
@Districh-ru

> If you start pointing any cell and move the cursor to the cursor to some other cell the action will be performed to the last pointed cell.

This is well-known behavior and it is observed in many other places IIRC. Let's not focus on it for now. And this in any case does not interfere with the logic of the mechanism from this PR:


https://github.com/ihhub/fheroes2/assets/32623900/42f716b0-84a1-429c-a17c-0e7952496e47

No matter how much we move the cursor with the mouse button pressed, we will not be able to force the unit to perform an "unconfirmed" action (sorry for the vertical video).

oleg-derevenetz:
> But if we (accidentally or on demand) touch other place out of battle cells range and then return to the first cell - the action proceeds at once without confirmation touch.

This issue should not be observed anymore:

https://github.com/ihhub/fheroes2/assets/32623900/612a913a-9eb2-45a0-82db-b0b179f2c28e

Branikolog:
Hi, @oleg-derevenetz 

> > So I can still miss the attacking direction.
> 
> No, you cannot. That's the whole trick. If attack direction changes, you should confirm it once again. After a new touch, both the cell and the cursor should be the same as after the first touch to proceed.

True, but If I accidentally touch the wrong cell once again (since my finger covers the entire sprite and its surroundings) the game will confirm the action, which I was intending to change.

The implementation from current PR looks good: you can click on any troop, take away finger and clearly see, what action was selected. 👍 The main missing feature here is to **confirm** the selected action **without pressing the same** place and potentially change action (which I described above).
I don't have any good ideas though, how to implement this. Standard dialog window is a 100% no... Maybe show confirmation button somewhere over battle journal?  Or use triple touch? 🤔

> All these menus are always bad IMHO. While with current implementation you are able to see the highlight of affected cells, these menus are not able to show anything like this.

I was talking about the implementation, when player clicks on a certain cell and it stays selected (like it is made in your PR), but you can interact with a small menu, sliding on a larger area around the cell (the nearest cells around the target) while the cell continues being selected. Then choose the attack direction you want. and release the finger.
It should look something like this:
Attack selection

The only question is how to cancel the selection. Probably touching any distant cell or initiate the right click...

oleg-derevenetz:
> Maybe show confirmation button somewhere over battle journal? Or use triple touch? 🤔

The confirmation button will most likely look "foreign", and triple touch... I believe that the user should not first read the instructions in three volumes in order to be able to perform the simplest actions in the game, and all sorts of special gestures are definitely not intuitive. In fact, those who have already played fheroes2 on mobile devices can also be confused by this PR and they may decide that battle actions are "not working".

> I was talking about the implementation, when player clicks on a certain cell and it stays selected (like it is made in your PR), but you can interact with a small menu, sliding on a larger area around the cell (the nearest cells around the target) while the cell continues being selected. Then choose the attack direction you want. and release the finger.

The problem is that the need for careful control is not limited to melee attacks, it is also needed when casting certain AOE spells (it would be good to raise the finger and see the area of effect before using them) or when shooting performed by stack of liches with their cloud damage.

Branikolog:
Totally agreed with you, @oleg-derevenetz .

> The confirmation button will most likely look "foreign", and triple touch... I believe that the user should not first read the instructions in three volumes in order to be able to perform the simplest actions in the game, and all sorts of special gestures are definitely not intuitive. In fact, those who have already played fheroes2 on mobile devices can also be confused by this PR and they may decide that battle actions are "not working".

All these proposals are questionable, for sure. That's why I mentioned, that had no really good ideas. :)
Your implementation is good for watching the cell, you've selected. The only difficulty is to touch exactly the same place once again. :) Nevertheless it is better, than current implementation from master branch. 

>The problem is that the need for careful control is not limited to melee attacks, it is also needed when casting certain AOE spells (it would be good to raise the finger and see the area of effect before using them) or when shooting is performed by stack of liches with their cloud damage.

For attacking interaction after we've selected the target we can visually highlight the target itself and the cells representing attacking positions with special cell shadows (flickering or some custom filling). Allowing player to slide near those cells to be able to see, what attacking angle is going to be activated on finger release.

Attack selection2

For single or AOE spells we can simply darken the area of the spell (or the target cell of single spell) and show the spell icon (which proceeds the cast) and the red cross (which cancels the spell) to the sides of the area, so player can simply confirm the action by touching the icon and not covering the target cell with his finger.
For shooters the same approach could be applied. 
So my idea is to highlight the target cell but player should touch the outer cells (which can have even small action icons) and release the finger in proper position to initiate the action.

Sorry, for wasting so much of your time. 😅 I just want the best solution for this issue to be invented.

oleg-derevenetz:
@Branikolog so in fact some sort of "menu" with "confirm" button should be sufficient. Player touches the target cell, removes the finger from the screen and then either clicks the "confirm" button or choose a different cell/attack direction. This should be enough for all cases. The problem is that this confirmation button will most likely look ugly and alien.

Branikolog:
@oleg-derevenetz 

> @Branikolog so in fact some sort of "menu" with "confirm" button should be sufficient. Player touches the target cell, removes the finger from the screen and then either clicks the "confirm" button or choose a different cell/attack direction. This should be enough for all cases - melee attack, shot and spellcast. 

Indeed.

>The problem is that this confirmation button will most likely look ugly and alien.

Everything new in the engine looks alien... For the first time. ;) As a reference I can mention damage box popping up over the battlefield.

I believe that players will forgive us some new UI elements for the sake of comfortable gameplay if they would be added with a proper visual design. 
We can even move new UI elements out of battle screen. But this is valid only for phones and PCs. Tablets or some rare foldable phones could still have close to 4:3 aspect ratio though. :(

Between, I've tested the PR build and found 0 issues or weird operation scenarios. 👍

oleg-derevenetz:
> I think in this case, a press and hold function that opens up a submenu with enlarged view of the angle attacks options then that requires an additional click on the option you choose.

This submenu works only for direct attacks and doesn't work for shots and spell casts, it is not enough. A lot of changes that solve only part of the problems. A separate confirmation button would solve all the problems, but I'm not a great designer, so if someone offers a working and not shabby solution with a confirmation button, it will be good.

> Is this confirmation step optional? Wouldn't it be best to add it as an opt in solution in option to see how it is recieved by the players testing it?

No, it is not optional (but it distincts actions performed by mouse and by touch - the former do not require confirmation). To make it optional we need to rework the battle option window, get the corresponding setting icon somewhere, etc.

Districh-ru (approved):
Well done, @oleg-derevenetz!
It works as intended, I like it.
I connected the mouse via OTA to my tablet and tested the battle using touchpad and mouse and found no issues. The mouse clicks works without changes (action by single click) and touch taps require a confirmation. If the cursor was positioned by mouse over a target tapping it works as a confirmation - as I expected.
I also checked this PR on my desktop (win 10 x64) and on PS Vita emulator (Vita3K) and found no issues. 👍 The Vita3K emulates touchscreen and I had to click twice to confirm the action as expected.

Currently this solution is the best to implement battle actions confirmation for touchscreen.

For adding a button or menu we need to find a proper place for it - it should not be foreign on the screen (fit into the screen image, do not overlap important screen parts) and should be comfortable to use it. If we will find the place and design a button for action confirmation we can add it later in a new PR in addition to this solution.

Branikolog (approved):
As a first step to implement a comfortable touch experience this PR is OK. 
It does not bring much convenience for now, though, without the confirmation UI.

src/fheroes2/battle/battle_interface.h
```diff
@@ -481,6 +482,60 @@ namespace Battle
...
+        };
+
+        // Intents are used to confirm actions in combat performed using touch gestures
+        BoardActionIntent _boardActionIntent;
+
+        class BoardActionIntentUpdater
+        {
+        public:
+            BoardActionIntentUpdater( Interface & interface, const bool isFromTouchpad )
+                : _interface( interface )
```
ihhub, line 504: Should we pass **_boardActionIntent** instead of **interface**? We use only this member in this class.

ihhub (commented):
Hi @oleg-derevenetz , I left one suggestion here. Could you please check it in your free time?

ihhub (approved):

oleg-derevenetz:
I hope that this will not be some "surprise" for Android players who are used to the old behavior...

ihhub:
@Branikolog please take a note to announce this feature in the next release.

PR #7779, opened 2023-09-20
Add historical changes support for the Editor

ihhub:
relates to #6845
- implement historical changes support through undo and redo methods (**U** and **R** keys)
- rename Editor class into EditorInterface to avoid the clash of names with a namespace
- fix application quitting event when pressing Save Map button in the Editor
- rename multiple Tiles class' members to be more human readable
- add a check in the main loop to detect invalid logic and to avoid infinite loop execution

https://github.com/ihhub/fheroes2/assets/19829520/b4a9dee2-6eae-404f-8d53-e5a2f1639421

This PR also fixes rare flickering on a first load of a new map in the Editor:

https://github.com/ihhub/fheroes2/assets/19829520/bff88c01-3a19-4ac7-a776-82911c065f5b


Districh-ru (changes requested):
Hi, @ihhub, I tested the undo/redo feature and I like how it works. It could be easily used for all new editor actions.
I only found one issue with the radar update on undo/redo and left a couple of comments about it. Could you please check them when you have time.

Also currently we can make the unlimited number of undo history records. Maybe we should limit the history manager maximum size to, in example, 100 records. What do you think?

src/fheroes2/editor/history_manager.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+            return foundDifference;
+        }
+
+        bool redo() override
+        {
+            for ( const Maps::Tiles & tile : _after ) {
+                Maps::Tiles temp{ tile };
+
+                world.GetTiles( tile.GetIndex() ) = std::move( temp );
```
oleg-derevenetz, line 75: It looks a bit odd - both here and in the `undo()` method below temporary tile is created first and then it is moved to the `world`. Why not just copy the `tile` to the `world` directly?

src/fheroes2/editor/history_manager.cpp
```diff
@@ -0,0 +1,125 @@
...
+            return;
+        }
+
+        try {
+            if ( action->prepare() ) {
+                _manager.add( std::move( _action ) );
+            }
+        }
+        catch ( ... ) {
+            // If an exemption happens here then something is very wrong with the code.
```
oleg-derevenetz, line 121:
```suggestion
            // If an exception happens here then something is very wrong with the code.
```

src/fheroes2/editor/history_manager.h
```diff
@@ -0,0 +1,118 @@
...
+            return ( _lastActionId > 0 );
+        }
+
+        bool isRedoAvailable() const
+        {
+            return ( _lastActionId < _actions.size() );
+        }
+
+    private:
+        std::vector<std::unique_ptr<Action>> _actions;
```
oleg-derevenetz, line 114: What do you think about imposing some kind of sane limit on the action history size? For example, ~500 or ~1000 actions? Or do you think that there is no need for this?

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -189,6 +189,10 @@ namespace
             = { HotKeyCategory::WORLD_MAP, gettext_noop( "hotkey|new map from scratch" ), fheroes2::Key::KEY_S };
         hotKeyEventInfo[hotKeyEventToInt( Game::HotKeyEvent::EDITOR_RANDOM_MAP_MENU )]
             = { HotKeyCategory::WORLD_MAP, gettext_noop( "hotkey|new random map" ), fheroes2::Key::KEY_R };
+        hotKeyEventInfo[hotKeyEventToInt( Game::HotKeyEvent::EDITOR_UNDO_LAST_ACTION )]
+            = { HotKeyCategory::WORLD_MAP, gettext_noop( "hotkey|undo last action" ), fheroes2::Key::KEY_U };
+        hotKeyEventInfo[hotKeyEventToInt( Game::HotKeyEvent::EDITOR_REDO_LAST_ACTION )]
+            = { HotKeyCategory::WORLD_MAP, gettext_noop( "hotkey|redo last action" ), fheroes2::Key::KEY_R };
```
oleg-derevenetz, line 195: BTW, why all editor's hotkeys are going to the `WORLD_MAP` category? Is this some kind of temporary solution?

src/fheroes2/maps/maps_tiles.h
```diff
@@ -109,6 +115,21 @@ namespace Maps
...
 
+        bool operator==( const Tiles & tile ) const
+        {
+            return ( _addonBottomLayer == tile._addonBottomLayer ) && ( _addonTopLayer == tile._addonTopLayer ) && ( _index == tile._index )
+                   && ( _terrainImageIndex == tile._terrainImageIndex ) && ( _terrainFlags == tile._terrainFlags ) && ( _uid == tile._uid )
+                   && ( _layerType == tile._layerType ) && ( _objectIcnType == tile._objectIcnType ) && ( _imageIndex == tile._imageIndex )
+                   && ( _hasObjectAnimation == tile._hasObjectAnimation ) && ( _isMarkedAsRoad == tile._isMarkedAsRoad ) && ( _mainObjectType == tile._mainObjectType )
+                   && ( _metadata == tile._metadata ) && ( _tilePassabilityDirections == tile._tilePassabilityDirections )
+                   && ( _isTileMarkedAsRoad == tile._isTileMarkedAsRoad ) && ( _occupantHeroId == tile._occupantHeroId );
+        }
```
oleg-derevenetz, line 126: And in C++20 we can just use the [default comparisons](https://en.cppreference.com/w/cpp/language/default_comparisons) :)

oleg-derevenetz (commented):

src/fheroes2/game/game_hotkeys.cpp
ihhub: Actually, no. I will move it to a separate category.

src/fheroes2/maps/maps_tiles.h
ihhub: The biggest problem as we don't compare every member in this class :(

src/fheroes2/editor/history_manager.cpp
ihhub: **operator=** was not specified in order to avoid any tiles to be duplicated in the game by any dev mistake. However, the Editor requires it. I was thinking to keep it this wat but I am fine to allow the copy.

oleg-derevenetz: I'm OK with this. Will you find it appropriate to add some comment regarding this (to minimize the surprise), but it's up to you.

ihhub: It is done.

Districh-ru (approved):

src/fheroes2/editor/history_manager.h
```diff
@@ -0,0 +1,126 @@
...
+
+            _actions.push_back( std::move( action ) );
+
+            ++_lastActionId;
+
+            if ( _actions.size() > maxActions ) {
+                --_lastActionId;
+                _actions.pop_front();
+            }
+        }
```
oleg-derevenetz, line 82: I recommend to add

```cpp
assert( _actions.size() <= maxAction );
```

after that `if` clause just in case.

oleg-derevenetz (approved):

PR #7796, opened 2023-09-22
Clean up the code for Maps::Tiles

ihhub:
- remove **_hasObjectAnimation** and **_isMarkedAsRoad** members from Maps::Tiles and Maps::TilesAddon since they were not used anywhere in the code
- remove the usage of bit flags for the above variables. We shouldn't read them
- rename multiple variable for structures for better understanding what they are for
- make the logic of loading data from MP2 format clear
- simplify the logic for **_isTileMarkedAsRoad** (relates to #7772)
- if an object is set we shouldn't check image index as it is not necessary
- remove weird logic for **layerType**

This pull request should be merged before #7772 in order to have much cleaner code for the future maintenance. There are more changes to be done but it is better to do them separately to avoid too complex changes in one shot.

This pull request request needs thoughtful testing to make sure that it does not break passability rules and does not introduce any rendering issues. How to test this pull request:
- load multiple maps and save them in debug mode within View World (set VIEWWORLD_DEBUG_ZOOM_LEVEL)
- compare them with master branch for rendering and passability rules
- load multiple save files and check that rendering and passabilities are correct

PR #7806, opened 2023-09-23
Rework logic related to object UIDs

ihhub:
Object UIDs are heavily used in the Editor and previously we had no proper mechanism to reset it and maintain. Also the history manager was not updating it as well.

This pull request should be merged before #7772 but after #7796.


oleg-derevenetz (approved):

src/fheroes2/system/save_format_version.h
```diff
@@ -44,5 +45,5 @@ enum SaveFileFormat : uint16_t
 
     LAST_SUPPORTED_FORMAT_VERSION = FORMAT_VERSION_1000_RELEASE,
 
-    CURRENT_FORMAT_VERSION = FORMAT_VERSION_1009_RELEASE
+    CURRENT_FORMAT_VERSION = FORMAT_VERSION_PRE1_1009_RELEASE
```
Districh-ru, line 48: As I understand here we renamed version 10012 from `FORMAT_VERSION_1009_RELEASE` to `FORMAT_VERSION_PRE1_1009_RELEASE` and made a new 10014 `FORMAT_VERSION_1009_RELEASE`.
So we should not change the current version to `FORMAT_VERSION_PRE1_1009_RELEASE` (10012), it should be 10014 `FORMAT_VERSION_1009_RELEASE`.
Currently every saved game by this PR hangs the engine if you try to load it. :(

Districh-ru (changes requested):
Hi, @ihhub, I left one comment regarding the issue with save version setting. Could you please address it when you have time.

src/fheroes2/system/save_format_version.h
ihhub: You're absolutely right!

Districh-ru (approved):

Issue #7272, opened 2023-06-05
Program, closing some screens may cause a different mouse cursor to appear briefly


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, closing some screens may cause a different mouse cursor to appear briefly.

In fheroes2:

https://github.com/ihhub/fheroes2/assets/43583994/1f361a2d-b500-46dc-8888-f7d41d4a973f

![Shotcut_00_00_08_280](https://github.com/ihhub/fheroes2/assets/43583994/c7821afd-e80d-4cef-9668-1bb535811ca2)

### Save file

Here is the save file:
[Dominion_0064.zip](https://github.com/ihhub/fheroes2/files/11615881/Dominion_0064.zip)

### Additional info

Version 1.0.4 7265

Related issues: https://github.com/ihhub/fheroes2/issues/6937, https://github.com/ihhub/fheroes2/issues/1266.

PR #7808, opened 2023-09-23
Fix missing cursor for 1 frame when it was being hidden by another function

ihhub:
Before the changes:

https://github.com/ihhub/fheroes2/assets/19829520/e4a90ac9-0f89-451a-8931-d88f577c42ec

After the changes:

https://github.com/ihhub/fheroes2/assets/19829520/401dbe59-c80e-445d-92a3-5a43c11531c2

I didn't touch the Editor as it needs more changes than this.


src/fheroes2/dialog/dialog_adventure.cpp
```diff
@@ -118,8 +118,10 @@ int Dialog::AdventureOptions( bool enabledig )
             fheroes2::showStandardTextMessage( _( "Cancel" ), _( "Exit this menu without doing anything." ), Dialog::ZERO );
     }
 
-    // restore background
+    // Restore background and cursor.
     back.restore();
+    cursorRestorer.reset();
```
oleg-derevenetz, line 123: Hi @ihhub this can be achieved by just using the proper scopes. Instead of

```cpp
const CursorRestorer cursorRestorer( ... );
[...]
display.render();

return ...;
```

we can just use

```cpp
{
    const CursorRestorer cursorRestorer( ... );
    [...]
}

display.render();

return ...;
```

No extra dynamic memory allocations, fully automatic.

ihhub: Hi @oleg-derevenetz , yes, your approach is absolutely valid. Why I did with unique pointers is to avoid heavy code changes since these are just few lines of code. However, if you think it's worth doing the correct way I can change few hundreds lines of code :)

ihhub: @oleg-derevenetz and @Districh-ru , by the way we should use ROI for the `display.render();` call not to render the whole screen. What do you think?

Districh-ru: @ihhub, i think - yes.
`display.render( back.rect() );` for this case.

Districh-ru (commented):
Hi, @ihhub, unfortunately these changes returns us back to https://github.com/ihhub/fheroes2/issues/7272


https://github.com/ihhub/fheroes2/assets/113276641/3cf774da-8646-453d-85c4-6160adc155f7


https://github.com/ihhub/fheroes2/assets/113276641/d3358fb3-8caf-454f-8271-632569b87888

ihhub:
@Districh-ru , 😕

PR #7811, opened 2023-09-24
Update followers window to support text mode

ihhub:
relates to #5270

You can test the change using this save file:
[Broken_Alliance_0001.zip](https://github.com/ihhub/fheroes2/files/12707889/Broken_Alliance_0001.zip)

### DIFF
--- a/src/fheroes2/dialog/dialog_armyinfo.cpp
+++ b/src/fheroes2/dialog/dialog_armyinfo.cpp
@@ -28,7 +28,6 @@
 #include <cstdint>
 #include <memory>
 #include <string>
-#include <type_traits>
 #include <utility>
 #include <vector>
 
--- a/src/fheroes2/dialog/dialog_armyinfo.cpp
+++ b/src/fheroes2/dialog/dialog_armyinfo.cpp
@@ -28,6 +28,7 @@
 #include <cstdint>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


Districh-ru (approved):
Well done!
However, I can suggest an improvement for one of the next PRs: Allow the right click on monster image to see dialog with monster characteristics.

Branikolog:
Hi, @ihhub 

Looks nice. But now I'm missing the same icon for Diplomacy and monsters retreating dialogs too. :)

image
image

ihhub:
Hi @Branikolog , the diplomacy window is too big. Adding a new icon would make it over the default resolution. I didn't touch other windows as the goal actually was to make this window to support text mode. I can remove the icon for the consistency.

Branikolog:
@ihhub 

> the diplomacy window is too big. Adding a new icon would make it over the default resolution.

Totally agreed. Maybe we need to widen windows for such cases? (for the distant future) There're also a few artifact descriptions, that make window fall out of the edges. Also, the majority of screens are wide nowadays, so making such dialogs wider for 50-100px wouldn't overflow the game UI, I believe.

ihhub:
@Branikolog , what's you conclusion then?

Branikolog:
@ihhub 

I've heard an opinion from players, that with text only in this window it's hard sometimes to even find the creature name.
I think adding **highlighted creature name** could solve the problem too here. While having an icon is much more visually pleasant and informative 👍 . Player can click to check creature characteristics too. 
So, If reworking diplomacy window could be proceeded in some day - we can leave current implementation as it is. If we will make just highlighting creature name without any icons, so I think adding such and then removing it will look.. weird. You know, such options that appear and then are gone could make players asking how to revert it back etc. And stick to 0.9.20 version, for instance, to play with beloved, but controversial feature. 😅 
Right now I love, how this window looks. 👍  But not sure, that we can rework diplomacy window in similar manner too.

ihhub:
@Branikolog , I am then removing the icon.

PR #7812, opened 2023-09-24
WIP: avoid making text copies when possible

ihhub:
A small optimization for text rendering.


ihhub:
I am closing this pull request for now as it has a very low priority but requires significant amount of testing. This kind of optimization can be done in parallel while we are changing the code.

PR #7813, opened 2023-09-24
Properly display player name in the New Game window

ihhub:
also remove the usage of the old class.

Before the changes (master branch) to display AAAAAAAAAA name:
image

After the changes:
image

### DIFF
--- a/src/fheroes2/gui/player_info.cpp
+++ b/src/fheroes2/gui/player_info.cpp
@@ -36,7 +36,6 @@
 #include "race.h"
 #include "screen.h"
 #include "settings.h"
-#include "text.h"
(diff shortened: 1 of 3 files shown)


Districh-ru (commented):
Hi, @ihhub. 
Now the user can understand that player name is longer than displayed in the dialog. This PR completes its goal.

My only suggestion is to show full name by pressing right mouse button on player's name. What do you think?

ihhub:
Hi @Districh-ru , let's do it in a separate pull request as an improvement. I am sure that we might need to display more than just a name.

Districh-ru (approved):

Branikolog (approved):

PR #7815, opened 2023-09-24
Fix memorizing the position in the map list when changing the map size filter

### LINKED ISSUE #7797: Bug with map choosing menu
### Preliminary checks

### Platform

Windows

### Describe the bug

I found a bug while choosing a map. After I pick the map and click "choose" i get to list of the maps, but not to position I was on, but somewhere deeper.
image

PS this bug works only S, M and XL map size. 

### Save file

https://github.com/ihhub/fheroes2/assets/145407420/4cc86500-32dc-4385-8164-e958991b456d

### Additional info

_No response_

oleg-derevenetz:
fix #7797


ihhub:

PR #7817, opened 2023-09-25
Fix battle AI protecting paralyzed archers

idshibanov:
AI was counting disabled shooters skewing battle estimations the wrong way.

### DIFF
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -763,6 +763,11 @@ uint32_t Battle::Unit::Resurrect( const uint32_t points, const bool allow_overfl
     return resurrect;
 }
 
+bool Battle::Unit::canShoot() const
+{
+    return isArchers() && !Modes( SP_BLIND | SP_PARALYZE | SP_STONE ) && !isHandFighting();
+}
+
 uint32_t Battle::Unit::ApplyDamage( Unit & enemy, const uint32_t dmg, uint32_t & killed, uint32_t * ptrResurrected )
 {
     killed = ApplyDamage( dmg );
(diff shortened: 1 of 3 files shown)


src/fheroes2/battle/battle_troop.cpp
```diff
@@ -763,6 +763,11 @@ uint32_t Battle::Unit::Resurrect( const uint32_t points, const bool allow_overfl
     return resurrect;
 }
 
+bool Battle::Unit::canShoot() const
+{
+    return isArchers() && !Modes( SP_BLIND | SP_PARALYZE | SP_STONE ) && !isHandFighting();
+}
```
ihhub, line 769: We can make this method inlined by moving the implementation into the header file.

ihhub (commented):
Hi @idshibanov , I left one comment here. Would you mind to take a look when you have free time?

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -763,6 +763,11 @@ uint32_t Battle::Unit::Resurrect( const uint32_t points, const bool allow_overfl
     return resurrect;
 }
 
+bool Battle::Unit::canShoot() const
+{
+    return isArchers() && !Modes( SP_BLIND | SP_PARALYZE | SP_STONE ) && !isHandFighting();
```
oleg-derevenetz, line 768: There is a "unifying" constant:

```suggestion
    return isArchers() && !Modes( SP_BLIND | IS_PARALYZE_MAGIC ) && !isHandFighting();
```

src/fheroes2/battle/battle_troop.cpp
idshibanov: I thought about it but it would require add additional include to battle_troop.h file. What is worse?

ihhub: Let's leave it as it is then.

ihhub (approved):

oleg-derevenetz (approved):

ihhub:

PR #7845, opened 2023-09-30
Teach battle AI to use shield spells

idshibanov:


ihhub (approved):

Branikolog:
Hi, @idshibanov !
I've made a couple of tests.
image
AI has default bless spell and Shield/mass shield spells in his book. 
In this case AI prefers using bless over the shield spell (maybe because he is more frightened by phoenixes). He uses shield only after he blessed 3 his key stacks. When the opposite hero had just 3 SP, he continued using bless all the time, as it lasts only 3 rounds. :)
If I have more ranged troops (or less phoenixes), AI uses shield at the first moments in the battle. 👍 
I only suggest rising the priority of the spell a little bit, especially the mass one.
I believe single shield brings more benefits for the most vulnerable stack. Does new logic counts this aspect?
[AI test shield.zip](https://github.com/ihhub/fheroes2/files/12792081/AI.test.shield.zip)

idshibanov:
> In this case AI prefers using bless over the shield spell (maybe because he is more frightened by phoenixes). He uses shield only after he blessed 3 his key stacks. When the opposite hero had just 3 SP, he continued using bless all the time, as it lasts only 3 rounds. :)
If I have more ranged troops (or less phoenixes), AI uses shield at the first moments in the battle. 👍

This is somewhat intentional and due to 40 phoenixes. Right now shield value is proportional to strength of enemy's archers. If 50% of enemy army are archers then shield is slightly more preferable than Stoneskin but worse than Bless. If enemy has archers only then shield is almost always cast. I'll bump up the priority slightly.

PR #7846, opened 2023-09-30
AI should avoid splitting troops on easy difficulty

idshibanov:
Might be up for discussion but removing the optimization should make battles easier. It will be a noticeable difference in behaviour between difficulties.

### DIFF
--- a/src/fheroes2/game/difficulty.cpp
+++ b/src/fheroes2/game/difficulty.cpp
@@ -159,7 +159,14 @@ uint32_t Difficulty::GetDimensionDoorLimit( int difficulty )
     return UINT32_MAX;
 }
 
-bool Difficulty::AISplittingWeakStacks( const int difficulty )
+bool Difficulty::allowAIToSplitWeakStacks( const int difficulty )
 {
-    return difficulty > Difficulty::NORMAL;
+    switch ( difficulty ) {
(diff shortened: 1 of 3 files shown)


oleg-derevenetz (approved):
Technically this change is OK. However, I do not know how much this will make the task easier for inexperienced players on easy difficulties :)

src/fheroes2/game/difficulty.cpp
```diff
@@ -158,3 +158,8 @@ uint32_t Difficulty::GetDimensionDoorLimit( int difficulty )
     }
     return UINT32_MAX;
 }
+
+bool Difficulty::AISplittingWeakStacks( const int difficulty )
+{
+    return difficulty > Difficulty::NORMAL;
```
ihhub, line 164: I recommend to still use switch case so it is easier to read the code and it handles all possible difficulties (if any new in the future as well).

ihhub (changes requested):
Hi @idshibanov , I left 2 minor comments here. Could you please take a look?

Also I don't believe that it makes any significant difference in the gameplay but let's hear others.

Branikolog (approved):

ihhub (approved):

ihhub:
@idshibanov , huge thanks for this change!

PR #7849, opened 2023-09-30
Use the "continue movement" button to interact with an object a hero is standing on

### LINKED ISSUE #7765: Touch screen controls

oleg-derevenetz:
close #7765

I know it looks somewhat controversial, but this mechanism allows the hero to interact with the object on which he stands when playing on touch-screen devices, without additional user interface changes and various cryptic finger gestures that no one will know about.

### DIFF
--- a/src/fheroes2/game/game_interface.h
+++ b/src/fheroes2/game/game_interface.h
@@ -110,7 +110,7 @@ namespace Interface
         void updateFocus();
 
         void EventSwitchHeroSleeping();
-        fheroes2::GameMode EventDefaultAction( const fheroes2::GameMode gameMode );
+        fheroes2::GameMode EventDefaultAction();
         void EventOpenFocus() const;
         fheroes2::GameMode EventSaveGame() const;
         void EventPuzzleMaps() const;
@@ -118,7 +118,7 @@ namespace Interface
         void EventSystemDialog() const;
         void EventNextHero();
         void EventNextTown();
-        void EventContinueMovement() const;
+        fheroes2::GameMode EventContinueMovement();
         void EventKingdomInfo() const;
         void EventCastSpell();
         void EventSwitchShowRadar() const;
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -1035,7 +1035,7 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_TOGGLE_ICONS ) )
                 EventSwitchShowIcons();
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_CONTINUE_HERO_MOVEMENT ) )
-                EventContinueMovement();
+                res = EventContinueMovement();
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_DIG_ARTIFACT ) )
                 res = EventDigArtifact();
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_SLEEP_HERO ) )
@@ -1068,7 +1068,7 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
                 _gameArea.SetScroll( SCROLL_BOTTOM );
             // default action
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_DEFAULT_ACTION ) )
-                res = EventDefaultAction( res );
+                res = EventDefaultAction();
             // open focus
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_OPEN_FOCUS ) )
                 EventOpenFocus();
--- a/src/fheroes2/gui/interface_buttons.cpp
+++ b/src/fheroes2/gui/interface_buttons.cpp
@@ -151,12 +151,10 @@ fheroes2::GameMode Interface::ButtonsArea::QueueEventProcessing()
     LocalEvent & le = LocalEvent::Get();
     fheroes2::GameMode res = fheroes2::GameMode::CANCEL;
 
-    if ( buttonNextHero.isEnabled() )
-        le.MousePressLeft( nextHeroRect ) ? buttonNextHero.drawOnPress() : buttonNextHero.drawOnRelease();
+    le.MousePressLeft( nextHeroRect ) ? buttonNextHero.drawOnPress() : buttonNextHero.drawOnRelease();
     le.MousePressLeft( movementRect ) ? buttonMovement.drawOnPress() : buttonMovement.drawOnRelease();
     le.MousePressLeft( kingdomRect ) ? buttonKingdom.drawOnPress() : buttonKingdom.drawOnRelease();
-    if ( buttonSpell.isEnabled() )
-        le.MousePressLeft( spellRect ) ? buttonSpell.drawOnPress() : buttonSpell.drawOnRelease();
+    le.MousePressLeft( spellRect ) ? buttonSpell.drawOnPress() : buttonSpell.drawOnRelease();
     le.MousePressLeft( endTurnRect ) ? buttonEndTurn.drawOnPress() : buttonEndTurn.drawOnRelease();
     le.MousePressLeft( adventureRect ) ? buttonAdventure.drawOnPress() : buttonAdventure.drawOnRelease();
     le.MousePressLeft( fileRect ) ? buttonFile.drawOnPress() : buttonFile.drawOnRelease();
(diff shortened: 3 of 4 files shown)


ihhub:
Hi @oleg-derevenetz , the idea is valid and I think we can use it. The only suggestion I have in order to make it intuitive is to change the icon to something else. @Districh-ru might help in this case.

ihhub:
@oleg-derevenetz , also we will need to change the description of the button.

@zenseii , do you have any suggestions how we should name the button and what description should it have?

Novicek:
Hi @oleg-derevenetz I think this is an elegant solution, but there is one issue I can see. When you stand on a tile, you can interact with it by pressing this button, but as soon as you click on a map to set a destination, I don't think there is a way to unset it again and so you have to move at that point, am I correct? I'm not sure what the solution to this would be without additional interface changes though.

oleg-derevenetz:
> I don't think there is a way to unset it again and so you have to move at that point, am I correct?

Yes, this solution will no be able to exactly replace the spacebar key, but IMHO it is not possible without some UI overhaul, what I would like to avoid.

oleg-derevenetz:
> also we will need to change the description of the button.

I made an update in this regard, but maybe someone will suggest something better.

ihhub (approved):

Branikolog:
Hi, @oleg-derevenetz !
Nice one!
Everything is good. 👍 


> > I don't think there is a way to unset it again and so you have to move at that point, am I correct?
> 
> Yes, this solution will no be able to exactly replace the spacebar key, but IMHO it is not possible without some UI overhaul, what I would like to avoid.

What do you think about the ability to reset hero path just after a few seconds holding this button pressed? So player can unset the path if he occasionally pressed somewhere and revisit object again?

oleg-derevenetz:
> What do you think about the ability to reset hero path just after a few seconds holding this button pressed? So player can unset the path if he occasionally pressed somewhere and revisit object again?

Eh, I'm afraid that this will not be the most obvious way for the player :) But if there are no more ideas, we can go this way in the future.

Branikolog (approved):
@oleg-derevenetz 
We can add a description like: "Hold to reset the path." Or something similar.
Anyway, it's a good idea of using this button. I was about to agree to widen the adventure map interface, to fit a pack of new buttons and interaction elements, but your proposal allows to skip that huge work. :)

oleg-derevenetz:
> We can add a description like: "Hold to reset the path." Or something similar.

"No one ever reads the instruction" (C) :)

oleg-derevenetz:
> What do you think about the ability to reset hero path just after a few seconds holding this button pressed? So player can unset the path if he occasionally pressed somewhere and revisit object again?

OK, I just implemented such a behavior. Please note that I implemented it, so to speak, "globally", i.e. there is now full support for "long press" event for mouse buttons (well, only for the left button for now, but adding other buttons should be now pretty easy) in the fheroes2 event system. I hope it doesn't break anything (including editor) - at least I tried not to let that happen :) The thing is that before this PR, the click could go on indefinitely - any amount of time could pass between pressing and releasing the mouse button, and this was still considered a click. This is not the case now: if a "click" lasts longer than 750ms, then the click event is not triggered at all, but the long-press event is triggered instead.

Also (during testing) I found and fixed the following bug:

https://github.com/ihhub/fheroes2/assets/32623900/7a79a87b-99c9-4aa9-a50e-c42f9e00aa1f

Branikolog:
Hi, @oleg-derevenetz !
I haven't found any issues with holding LMB. 
Sad, that holding more than 750ms cancels the click event at all. I would like to be able to hold LMB infinitely, and in a case I release it over the button, and in a case there's no "hold event" for that area, left click proceeds.
If it's not possible, I suggest to slightly increase the time from 750ms to ~850-900ms, As sometimes I hold the button a little longer, while hesitating during game process. It's just my personal feels, though.

oleg-derevenetz:
Hi @Branikolog

> I would like to be able to hold LMB infinitely, and in a case I release it over the button, and in a case there's no "hold event" for that area, left click proceeds.

This doesn't looks good, because it means that whether it's a click or hold event depends on the order of checks during event handling, and if events are handled in the multiple places of the event handling hierarchy, the "correct order" may be difficult to keep track of. We already had similar issues with "click vs release", so I decided not to replicate that design issue once again.

Branikolog:
@oleg-derevenetz 
Yeah, I can only imagine the complexity of the logic here but I expected such answer from you.. :)
Increased timeout works better for me. 👍 
It is going to become a splendid feature when the new icon for spacebar action appears here!

ihhub (approved):

oleg-derevenetz:
Hi @Districh-ru

> it'll be good to improve it in future, for example, display the released button after the long press delay is passed (the button is now visually pressed after the long press event, but there is no action when the mouse button is released)

I suppose there's no need to mix it up. The player already knows that the long-press event has come because the corresponding action has been performed. But he is still "pressing" this button, so it should be drawn pressed until he releases it. If it was a real physical button, it would behave the same as it does now.

> if there is no actions for the long press event - do not reset the button pressed state and when the mouse button is released - act as if it was clicked (as is done now in master build)

I don't like this logic, because `LocalEvent` has no idea about what you expect. There are no "subscriptions" or something like this. The logic of the work is determined purely by the order of checks for various events, and, as I mentioned above in my discussion with @Branikolog, when the triggering of one check depends on whether another check has already been performed, especially if it's all scattered across different classes, then it becomes difficult to keep track of the "correct order" of checks. I believe that such a logic should be avoided whenever possible.

Districh-ru:
Hi, @oleg-derevenetz, 
I was talking about the case when there is no long-press event at all for the button (currently for all buttons except the "move" button, and not only the buttons) and the player does not know that the click action is cancelled after a long press.
I have mostly the same concern as @Branikolog described: I was pressing button a little longer than 850 ms (maybe I was thinking about do I really want to do this action) ant then I release the mouse button an find that the pressed button does nothing (a surprise. why? a bug? ). Yes, it is a rare case, but sometimes it happens and I was thinking that we may somehow let the player know that the click action will be cancelled and that this is not a bug of the game or hardware. :)
Open any dialog in OS, press the any button for a long time, release the mouse button and the click action will be performed without any canceling. I'm used to such behavior, sorry.
So I'm only talking that we can try to improve the click/long-press behavior in future only if we find a good solution for it. I agree that we should not do a race of handlers and bring a mess to the event checks.
And maybe I'm worrying for nothing.. :)

And the introduction of the long-press event is a good think and I'm not against it - for example we can expand the behavior of the virtual keyboard's shift/caps-lock key: single press - shift, long press - caps lock.

oleg-derevenetz:
@Districh-ru

> Open any dialog in OS, press the any button for a long time, release the mouse button and the click action will be performed without any canceling. I'm used to such behavior, sorry.

And at the same time these OS do not have the built-in support for the long-press event:

https://learn.microsoft.com/en-us/dotnet/desktop/winforms/input-mouse/events?view=netdesktop-7.0

And each developer is forced to implement this event somehow in his own way in each specific case. Doesn't look too convenient. Initially, I implemented this event purely for this button, but then decided to do it more globally. But there is a problem - I could make these events somewhat dependent, as Click and Release events work now (if we checked for the Click event and it already happened, then, even if we do not perform any action for it, the Release event is cleared and will not be triggered). This at one time brought certain problems if the Click and Release events were processed in different places (for example, the state of the button/element was drawn in one place according to the Press/Release events, and the click event was processed in another, unrelated place).

oleg-derevenetz:
@Branikolog @Districh-ru OK, if you think it will be more convenient, then I did it the way you want - that is, the click event will be canceled only if the long-click event not only occurred, but was also checked/triggered. Otherwise, the click event still will be fired regardless of the time passed between the press and release events.

Districh-ru (approved):
@oleg-derevenetz, thanks!
I tested the change and now the left click behavior is more convenient for me.

ihhub:
@oleg-derevenetz , amazing work!

PR #7852, opened 2023-10-01
Add functionality to place monsters in the Editor

ihhub:
relates to #6845

https://github.com/ihhub/fheroes2/assets/19829520/6185e4ee-d468-495e-9af2-1a95c6744956

There are 2 problems which should be addressed later:
- while changing the terrain we should remove all **main** objects from a tile
- while using the eraser we should remove all **main** objects from a tile

Both issues will be addressed in a separate pull request.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -586,6 +586,20 @@ namespace Interface
                 }
             }
         }
+        else if ( _editorPanel.isMonsterSettingMode() ) {
+            if ( tile.isWater() ) {
+                fheroes2::showStandardTextMessage( _( "Monster" ), _( "Monsters cannot be set on water." ), Dialog::OK );
+            }
+            else if ( !Maps::isClearGround( tile ) ) {
+                fheroes2::showStandardTextMessage( _( "Monster" ), _( "Choose a tile which does not contain any objects." ), Dialog::OK );
+            }
+            else {
+                const fheroes2::ActionCreator action( _historyManager );
+
+                Maps::setMonsterOnTile( tile, _editorPanel.getMonsterId(), 0 );
+                _redraw |= mapUpdateFlags;
+            }
+        }
     }
 
     void EditorInterface::mouseCursorAreaPressRight( const int32_t tileIndex ) const
@@ -623,8 +637,13 @@ namespace Interface
         }
     }
 
-    void EditorInterface::updateCursor( const int32_t /*tileIndex*/ )
+    void EditorInterface::updateCursor( const int32_t tileIndex )
     {
-        Cursor::Get().SetThemes( Cursor::POINTER );
+        if ( _cursorUpdater ) {
+            _cursorUpdater( tileIndex );
+        }
+        else {
+            Cursor::Get().SetThemes( Cursor::POINTER );
+        }
     }
 }
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -593,7 +593,7 @@ namespace Interface
             else if ( !Maps::isClearGround( tile ) ) {
                 fheroes2::showStandardTextMessage( _( "Monster" ), _( "Choose a tile which does not contain any objects." ), Dialog::OK );
             }
-            else {
+            else if ( Monster{ _editorPanel.getMonsterId() }.isValid() ) {
                 const fheroes2::ActionCreator action( _historyManager );
 
                 Maps::setMonsterOnTile( tile, _editorPanel.getMonsterId(), 0 );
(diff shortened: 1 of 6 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -586,6 +586,20 @@ namespace Interface
...
+                fheroes2::showStandardTextMessage( _( "Monster" ), _( "Monsters cannot be set on water." ), Dialog::OK );
+            }
+            else if ( !Maps::isClearGround( tile ) ) {
+                fheroes2::showStandardTextMessage( _( "Monster" ), _( "Choose a tile which does not contain any objects." ), Dialog::OK );
+            }
+            else {
+                const fheroes2::ActionCreator action( _historyManager );
+
+                Maps::setMonsterOnTile( tile, _editorPanel.getMonsterId(), 0 );
+                _redraw |= mapUpdateFlags;
```
Districh-ru, line 600: We should not place a monster if `_editorPanel.getMonsterId() == Monster::UNKNOWN`. Currently the Editor places a 0 (zero) count of peasants. :)

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -586,6 +586,20 @@ namespace Interface
...
+            if ( tile.isWater() ) {
+                fheroes2::showStandardTextMessage( _( "Monster" ), _( "Monsters cannot be set on water." ), Dialog::OK );
+            }
+            else if ( !Maps::isClearGround( tile ) ) {
+                fheroes2::showStandardTextMessage( _( "Monster" ), _( "Choose a tile which does not contain any objects." ), Dialog::OK );
+            }
+            else {
+                const fheroes2::ActionCreator action( _historyManager );
+
+                Maps::setMonsterOnTile( tile, _editorPanel.getMonsterId(), 0 );
```
Districh-ru, line 599: An added monster always has zero uid and ICN object type.
```
******* Tile info *******
Tile index      : 260, point: (8, 7)
UID             : 0
MP2 object type : 152 (Отряд)
ICN object type : 0 (UNKNOWN)
image index     : 2
```

Districh-ru (changes requested):
Hi, @ihhub, I left a couple comments after a quick check. Could you please check them in your spare time.

Also there is a mouse cursor issue. We may address it in a new PR: if you open the quick info or any dialog (in example quit dialog) the cursor is set to the POINTER after the dialog is closed and returned back to the monster icon after you move it to the nearby tile. (In future we could show a POINTER cursor above the places were the monster can not be placed).

https://github.com/ihhub/fheroes2/assets/113276641/747050ea-c69d-459c-b91e-238f9a1a160f


And a question to discuss - maybe after choosing a monster we should open the `Dialog::SelectCount()` with the default monster count set by default?

src/fheroes2/editor/editor_interface.cpp
ihhub: This is bug which is present in the game as well.

ihhub:
Hi @Districh-ru , it is better to set monster count separately while you click on the monster. Doing it right after choosing a monster will limit your ability to quickly set them. I know that 0 doesn't mean 0 monsters but we can find a solution later.

ihhub:
@Districh-ru , I am aware of the cursor issue. This is due to cursor updater class. I am still thinking how to address it properly in a separate PR.

ihhub:
Hi @zenseii , could you please review the new text dialogs here?

Districh-ru (approved):
Well done!
изображение

In future we should also add random monster placing.

Branikolog:
Hi, @ihhub 
I've made a couple of tests. 👍 
The only tiny thing I would like to see is turning monster cursor into pointer while hovering over edges of the adventure map. But I suppose it is related to the issue you've discussed above with @Districh-ru . It's not urgent for sure. Everything else works well.

ihhub:
Hi @Branikolog , I addressed your concern regarding the cursor.

Branikolog:
Thanks, @ihhub !
There's an issue with positioning of the cursor after right click now:

https://github.com/ihhub/fheroes2/assets/55348946/1c4ec6b7-612f-41fc-9be0-0e74da73108f

ihhub:
Hi @Branikolog , this is the same problem which @Districh-ru spotted. This will be addressed in the separate pull request.

Branikolog:
@ihhub 
I mostly concern about the **position** of the cursor (creature sprite) and the highlighted cell of the map:
image
Creature sometimes appears on the map really far from the cursor.

The issue spotted by @Districh-ru , relates to showing the pointer cursor instead of creature sprite. Correct me if I'm wrong.

But I want to apologize, the issue I'm talking about is valid for a previous implementation too.

Districh-ru:
> Creature sometimes appears on the map really far from the cursor.

I also observe this issue: it appears only with disabled software cursor - the upper left corner of the monster sprite is set as the cursor center.
With software cursor turned on this issue does not appear - the center of monster sprite is set as the cursor center.

ihhub:
Hi @Branikolog , I fixed the issue with the cursor. Please take a look.

PR #7856, opened 2023-10-02
Restrict heroes meetings and army merging on Easy difficulty

ihhub:
To address the complains of players about the harshness of the AI.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_kingdom.cpp
@@ -40,6 +40,8 @@
 #include "audio_manager.h"
 #include "castle.h"
 #include "color.h"
+#include "difficulty.h"
+#include "game.h"
 #include "game_interface.h"
 #include "ground.h"
 #include "heroes.h"
@@ -122,10 +124,19 @@ namespace
         std::vector<Heroes *> _heroes;
     };
 
-    void setHeroRoles( VecHeroes & heroes )
+    void setHeroRoles( VecHeroes & heroes, const int difficulty )
     {
         if ( heroes.empty() ) {
             // No heroes exist.
+            return;
+        }
+
+        if ( !Difficulty::areAIHeroRolesAllowed( difficulty ) ) {
+            // All heroes are equal.
+            for ( Heroes * hero : heroes ) {
+                hero->setAIRole( Heroes::Role::HUNTER );
+            }
+
             return;
         }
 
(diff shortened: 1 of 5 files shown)


idshibanov:
I was thinking if we should leave Scout role as is. Easy difficulty is already the one with the worst vision.

src/fheroes2/ai/normal/ai_normal_kingdom.cpp
```diff
@@ -691,6 +702,9 @@ namespace AI
         const AIAutoControlModeCommitter aiAutoControlModeCommitter( kingdom );
 #endif
 
+        // TODO: each AI player should have its own difficulty level.
+        _difficulty = Game::getDifficulty();
```
oleg-derevenetz, line 706: Hi @ihhub AFAIK game difficulty is set just once for the entire map, and there is no separate difficulties for each player. Why the support for separate difficulties for each AI player is needed?

ihhub: Hi @oleg-derevenetz , my intention is to give AI players different difficulties for 2 reasons:
- map making to make weaker or stronger AIs (debatable but useful for the future campaign)
- human's AI allies should not have all benefits as opposing AIs

Of course if you think that none of these make sense I am absolutely fine to remove the comment and not to touch the code in the future.

idshibanov: We discussed an ability to set difficulty per player in the future. I see two potential uses: difficulty override using map editor to fine-tune the balance and reducing AI perks for player allies on high difficulties. If you play on Impossible allied AI players should stay Normal.

oleg-derevenetz: Hi @ihhub I'm just a bit concerned about this change because now in some places this value is used, and in other places (in the AI battle code, when calculating the bonus to the creature growth for the AI-controlled town/castle, when calculating the movement bonus for the AI-controlled heroes, when calculating the income bonus for the AI-controlled kingdom, etc, etc) the result of the `Game::getDifficulty()` call is used. I'm not at all against the idea of a separate difficulty for each AI player, I just think that it should be well thought out separately and implemented as a "one piece" as a separate pull request in future. Otherwise, confusion may arise.

idshibanov: @ihhub lets remove the variable for now and we will cache it later as a separate PR. I want this change to go in.

oleg-derevenetz (approved):

idshibanov (approved):

Districh-ru (approved):

PR #7859, opened 2023-10-03
Combat victory window creature position

### LINKED ISSUE #1611: Combat, Victory window, the position of the creature sprites is not centered horizontally in the window
The spacing is not the same between the edges of the window and the sprites of the boars and the sprites of the elementals.

In fheroes2:

Full elementals victory 01

paprocki:
Fix #1611
Margins fixed and displaying starts from the right of the window so overlapping is more appealing. 
To test with different army sizes you can setup a hardcoded army in battle_dialogs.cpp, l. 500:
```
    Troops killedcustom = Troops(); 
    Troop troop = Troop(Monster::AIR_ELEMENT, 10);
    Troop troopBoar = Troop( Monster::BOAR, 10 );

    killedcustom.PushBack( troop, troop.GetCount() );
    killedcustom.PushBack( troopBoar, 5 );

    const Troops killed1 = killedcustom; //_army1->GetKilledTroops();
```
    
I'm new to c++ and memory management on this level, so if you could be extra alert on that while reviewing, then that would be great.

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -539,21 +539,6 @@ void Troops::MergeSameMonsterTroops()
     }
 }
 
-Troops Troops::GetReversed() const
-{
-    Troops result;
-    result.reserve( size() );
-
-    for ( auto iter = rbegin(); iter != rend(); ++iter ) {
-
-        Troop * newTroop = new Troop( **iter );
-        result.push_back( newTroop );
-    }
-
-    return result;
-}
-
-
 bool Troops::MergeSameMonsterOnce()
 {
     for ( size_t slot = 0; slot < size(); ++slot ) {
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -539,6 +539,21 @@ void Troops::MergeSameMonsterTroops()
     }
 }
 
+Troops Troops::GetReversed() const
+{
+    Troops result;
+    result.reserve( size() );
+
+    for ( auto iter = rbegin(); iter != rend(); ++iter ) {
+
+        Troop * newTroop = new Troop( **iter );
+        result.push_back( newTroop );
+    }
+
+    return result;
+}
+
+
 bool Troops::MergeSameMonsterOnce()
 {
     for ( size_t slot = 0; slot < size(); ++slot ) {
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -1502,12 +1502,12 @@ bool Army::isMeleeDominantArmy() const
     return meleeInfantry > other;
 }
 
-void Army::drawSingleDetailedMonsterLine( const Troops & troops, int32_t cx, int32_t cy, uint32_t width )
+void Army::drawSingleDetailedMonsterLine( const Troops & troops, int32_t cx, int32_t cy, int32_t width )
 {
     fheroes2::drawMiniMonsters( troops, cx, cy, width, 0, 0, false, true, false, 0, fheroes2::Display::instance() );
 }
 
-void Army::drawMultipleMonsterLines( const Troops & troops, int32_t posX, int32_t posY, uint32_t lineWidth, bool isCompact, const bool isDetailedView,
+void Army::drawMultipleMonsterLines( const Troops & troops, int32_t posX, int32_t posY, int32_t lineWidth, bool isCompact, const bool isDetailedView,
                                      const bool isGarrisonView /* = false */, const uint32_t thievesGuildsCount /* = 0 */ )
 {
     const uint32_t count = troops.GetOccupiedSlotCount();
--- a/src/fheroes2/army/army_ui_helper.cpp
+++ b/src/fheroes2/army/army_ui_helper.cpp
@@ -59,7 +59,7 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
         }
 
         int marginLeft = 18;
-        chunk =  ( width - marginLeft ) / ( count + ( slotsToSkip * 2 ) ) ;
+        chunk = ( width - marginLeft ) / ( count + ( slotsToSkip * 2 ) );
 
         slots = troops.Size() + slotsToSkip;
 
@@ -81,7 +81,7 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
             troop = troops.GetTroop( slot );
         }
         else {
-            troop = troops.GetTroop( ( troops.Size() - 1 ) - slot + slotOffset ); 
+            troop = troops.GetTroop( ( troops.Size() - 1 ) - slot + slotOffset );
         }
 
         if ( troop == nullptr || !troop->isValid() ) {
--- a/src/fheroes2/army/army_ui_helper.cpp
+++ b/src/fheroes2/army/army_ui_helper.cpp
@@ -45,8 +45,8 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
     }
 
     int slotsToSkip = 1;
-    if ( count < 5 ) {
-        slotsToSkip = 1;
+    if ( count <= 2 ) {
+        slotsToSkip = 2;
     }
     else if ( count >= 7 ) {
         slotsToSkip = 0;
@@ -65,7 +65,7 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
     const int slotOffset = slotsToSkip;
 
     for ( size_t slot = 0; slot < slots; ++slot ) {
-
+        //Skip the monster handling when a slot is empty
         if ( slotsToSkip > 0 ) {
             slotsToSkip--;
             cx -= chunk;
@@ -113,7 +113,13 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
         else {
             if ( slotsToSkip == 0 ) {
                 const int offsetY = 28 - monster.height();
-                const int OffsetX = -15; //(monster.width() / 2) -12;
+                int OffsetX = -14;    
+
+                //Center the monster if there is only one
+                if ( count == 1 && slot == 1 ) {
+                    OffsetX = -10;
+                }
+
                 int x = ( cx - ( monster.width() / 2 ) ) + OffsetX;
                 int y = cy + offsetY + monster.y(); 
                 fheroes2::Blit( monster, output, x, y);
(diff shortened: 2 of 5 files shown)


ihhub:
Hi @paprocki , please fix code style issues spotted by GitHub Actions.

src/fheroes2/army/army.cpp
```diff
@@ -539,6 +539,21 @@ void Troops::MergeSameMonsterTroops()
     }
 }
 
+Troops Troops::GetReversed() const
```
ihhub, line 542: We should avoid create a new method just to cover a single case in the code. We can archive the same by doing like this at line 57 of **army_ui_helper.cpp** file:
```cpp
Troops reversedTroops;

if ( troops.Size() > 0 ) {
    for ( size_t i = troops.Size() - 1; ; ) {
        const Troop * monster = troops.Get( i );
        reversedTroops.PushBack( *monster, monster->GetCount() );

        if ( i > 0 ) {
            --i;
        }
        else {
            break;
        }
    }
}
```

Troops class is very... bad written at the moment but this is better to complicate it even more.

ihhub (changes requested):
Hi @paprocki , I left one comment about one optimization. Could you please take a look?

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,36 @@
...
+
+    int marginLeft = 18;
+    const int chunk = ( width - marginLeft ) / (count + ( slotsToSkip * 2 ) );
+    Troops reversedTroops = troops.GetReversed();
+    const size_t slots = reversedTroops.Size() + slotsToSkip;
 
     if ( !isCompact ) {
-        cx += chunk / 2;
+        cx -= chunk / 2;
+        cx += width;
```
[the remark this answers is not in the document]
paprocki: Fixed

src/fheroes2/army/army.cpp
paprocki: I did as requested.

paprocki:
I changed a bit more code than intended, because i saw an error with the compact army. Please review the code again

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,57 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
+            slotsToSkip = 2;
+        }
+        else if ( count >= 7 ) {
+            slotsToSkip = 0;
+        }
+
+        int marginLeft = 18;
+        chunk = static_cast<int32_t>( ( width - marginLeft ) / ( count + ( slotsToSkip * 2 ) ) );
+
+        if ( troops.Size() > 0 ) {
```
ihhub, line 64: We can remove this check since at line 39 we have a similar check. To be safe we can replace this if-condition by an assertion:
```cpp
assert( troops.Size() > 0 );
for ( size_t i = troops.Size() - 1;; ) {
...
```

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,57 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
+        if ( slotsToSkip > 0 && !isCompact ) {
+            slotsToSkip--;
+            cx -= chunk;
+            continue;
+        }
+
         const Troop * troop = troops.GetTroop( slot );
+        if ( !isCompact ) {
+            troop = reversedTroops.GetTroop( slot - slotOffset );
+        }
```
ihhub, line 96: We should do this way to avoid extra calculations:
```cpp
const Troop * troop = nullptr;
if ( !isCompact ) {
    troop = troops.GetTroop( slot );
}
else {
    troop = reversedTroops.GetTroop( slot - slotOffset );
}
```

However, this pounders a question: do we need **reversedTroops** at first place?

Essentially, **reversedTroops** is just a reversed version of **troops** so accessing the first element of **reversedTroops** is accessing the last element of **troops**. In this case will this work
```cpp
troop = troops.GetTroop( troops.Size() - slot + slotOffset );
```
?

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -89,11 +132,27 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
             text.draw( cx + chunk - text.width() - offset, cy + 23, output );
         }
         else {
-            const int offsetY = 28 - monster.height();
-            fheroes2::Blit( monster, output, cx - monster.width() / 2 + monster.x() + 2, cy + offsetY + monster.y() );
-            text.draw( cx - text.width() / 2, cy + 29, output );
+            if ( slotsToSkip == 0 ) {
+                const int offsetY = 28 - monster.height();
+                int OffsetX = -14;
```
ihhub, line 137: Please name the variables from a non-capital letter as per our coding standard.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -89,11 +132,27 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
             text.draw( cx + chunk - text.width() - offset, cy + 23, output );
         }
         else {
-            const int offsetY = 28 - monster.height();
-            fheroes2::Blit( monster, output, cx - monster.width() / 2 + monster.x() + 2, cy + offsetY + monster.y() );
-            text.draw( cx - text.width() / 2, cy + 29, output );
+            if ( slotsToSkip == 0 ) {
```
ihhub, line 135: Please merge else with if.

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,57 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
-        cx += chunk / 2;
+        if ( count <= 2 ) {
+            slotsToSkip = 2;
+        }
+        else if ( count >= 7 ) {
+            slotsToSkip = 0;
+        }
+
+        int marginLeft = 18;
+        chunk = static_cast<int32_t>( ( width - marginLeft ) / ( count + ( slotsToSkip * 2 ) ) );
```
ihhub, line 62: Could you please change the input argument type `const uint32_t width` into `const int32_t width` for this function as well as for **drawMultipleMonsterLines**? After this we don't need to do the static casting here and at line 81.

ihhub (changes requested):
Hi @paprocki , I left few more comments here. Would you mind to check them?

ihhub (approved):

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,46 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
         count = troops.GetOccupiedSlotCount();
     }
 
-    const int chunk = width / count;
+    int chunk = static_cast<int32_t>( width / count );
+    size_t slots = troops.Size();
+    int slotsToSkip = 1;
+
+    int slotOffset = 1;
```
Districh-ru, line 51: The same comment as above :)

```suggestion
    int32_t slotsToSkip = 1;

    int32_t slotOffset = 1;
```

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,46 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
-        cx += chunk / 2;
+        if ( count <= 2 ) {
+            slotsToSkip = 2;
+        }
+        else if ( count >= 7 ) {
+            slotsToSkip = 0;
+        }
+
+        int marginLeft = 18;
+        chunk = static_cast<int32_t>( ( width - marginLeft ) / ( count + ( slotsToSkip * 2 ) ) );
```
Districh-ru, line 62: Here we can also make variable types consistent and cast only `count` which is `uint32_t`:
```suggestion
        int32_t marginLeft = 18;
        chunk = ( width - marginLeft ) / ( static_cast<int32_t>( count ) + ( slotsToSkip * 2 ) );
```

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -44,14 +44,46 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
+
+        slots = troops.Size() + slotsToSkip;
```
Districh-ru, line 64: `slots` is already made equal to `troops.Size()` in 48 line, so here we may speed-up the computations and do not call `.Size()` once more:
```suggestion
        slots += static_cast<size_t>( slotsToSkip );
```

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -88,12 +120,26 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
-            text.draw( cx - text.width() / 2, cy + 29, output );
+            int offsetX = -14;
+
+            // Center the monster if there is only one
+            if ( count == 1 && slot == 1 ) {
+                offsetX = -10;
+            }
+
+            int x = ( cx - ( monster.width() / 2 ) ) + offsetX;
+            int y = cy + offsetY + monster.y();
```
Districh-ru, line 133: `fheroes2::Blit()` accepts `int32_t` so to avoid unnecessary type conversions lets set these variables type to `int32_t`.
```suggestion
            int32_t x = ( cx - ( monster.width() / 2 ) ) + offsetX;
            int32_t y = cy + offsetY + monster.y();
```

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -88,12 +120,26 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
             fheroes2::Blit( monster, output, cx + offset, cy + offsetY + monster.y() );
             text.draw( cx + chunk - text.width() - offset, cy + 23, output );
         }
-        else {
+        else if ( slotsToSkip == 0 ) {
             const int offsetY = 28 - monster.height();
```
Districh-ru, line 124: To be consistent with `fheroes2::Blit()` input variables types lets set the ёште32_её type for this variable:
```suggestion
            const int32_t offsetY = 28 - monster.height();
```

src/fheroes2/army/army_ui_helper.cpp
```diff
@@ -88,12 +120,26 @@ void fheroes2::drawMiniMonsters( const Troops & troops, int32_t cx, const int32_
...
-            fheroes2::Blit( monster, output, cx - monster.width() / 2 + monster.x() + 2, cy + offsetY + monster.y() );
-            text.draw( cx - text.width() / 2, cy + 29, output );
+            int offsetX = -14;
```
Districh-ru, line 125: The same comment as previous:
```suggestion
            int32_t offsetX = -14;
```

Districh-ru (changes requested):
Hi, @paprocki, good work!
изображение

I left several comments regarding variables type consistency. Could you please take a look in your spare time.

Branikolog (approved):

Districh-ru (approved):

ihhub:
@paprocki , huge thanks for this fix!

Issue #7853, opened 2023-10-01
Campaign briefing window, insufficient space for scenario name 


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

![image](https://github.com/ihhub/fheroes2/assets/55348946/16c7fcc0-ae26-4fa3-a1f8-9da18c5875c5)

### Describe the possible solution

I propose using the same approach, that was used for battle log etc. (#7847)

Branikolog:
It would be nice to be able to right click on this name to see it full.

ihhub:
I am marking as Beyond OG scope as this is a part of the OG resources.

PR #7864, opened 2023-10-04
Completely remove old Text classes

ihhub:
This pull request completely remove the old Text classes which were limited in functionality and slower. This is the list of dialogs where changes were done:
- battle outcome summary
- battle outcome artifact dialog
- battle outcome necromancy dialog
- hero dialog info during battle
- building requirement dialog
- building info in castle dialog
- building info in Kingdom overview
- recruit hero dialog
- captain info inside castle dialog
- arena dialog
- creature info dialog
- diplomacy window for monster joining
- treasure chest dialog
- hero level up dialog
- marketplace dialog
- setting up values dialog for only battle mode (like setting skill value)
- monster separation
- name entering dialog (name of a player for a multiplayer game)
- Thieves' guild dialog
- campaign info screen (relates to #7853)
- interface status (resource pickup event, date window and kingdom window)
- hero dialog
- hero meeting dialog
- town portal spell dialog
- spell bool mana value

All dialogs were compared to the ones before the changes, Some multi-line texts changed due to the way we render it plus in 2 places text was shifted by 1 pixel.

On top of this, the new Text classes were not handling special cases with spaces. You can revert the change and test Arena dialog to see what's happening.

Please note that I didn't do a lot of optimizations, the goal of this pull request is to get rid of the old classes. We should optimize the code separately.


src/fheroes2/dialog/dialog_levelup.cpp
```diff
@@ -77,9 +77,7 @@ int DialogOneSecondary( const Heroes & hero, const std::string & name, const int
 
 int DialogSelectSecondary( const std::string & name, const int primarySkillType, const Skill::Secondary & sec1, const Skill::Secondary & sec2, Heroes & hero )
 {
-    std::string header = _( "%{name} has gained a level." );
-    header.append( "\n \n" );
-    header.append( _( "%{skill} +1" ) );
+    std::string header = _( "%{name} has gained a level.\n%{skill} +1" );
```
Districh-ru, line 80: Was the empty text string removed for some reason? :)
```suggestion
    std::string header = _( "%{name} has gained a level.\n\n%{skill} +1" );
```

Districh-ru (changes requested):
Hi @ihhub, I checked most of the changed places and found one issue with the lost empty line (`\n`). Could you check it when you have time.

One pixel text shift in some places is OK for me.
I also found some places to make minor optimizations, but this PR is big enough. Its better to make optimizations later in separate PRs and not only for the text render.

src/fheroes2/dialog/dialog_levelup.cpp
ihhub: Nope. I think my rendering was off with the old text. I fixed it.

Districh-ru (approved):
Well done!
Later I'll do some rendering optimizations for dialogs. For example we can optimize `dialog_armyinfo` like we did for the castle well: do not fully render all unchanged texts for every monster animation frame.

Branikolog (approved):

Issue #7860, opened 2023-10-03
Fix the German translation


erazortt:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

I have gone through the whole German translation file comparing every entry with the strings inside my original HEROES2.EXE file. I have not taken the strings from the original version blindly however, since at times your current translation was clearly much better. I chose for each string the "best" representation. "Best" can of corse be subjective, however sometimes, the translation was quite bad, and at times even almost unintelligible. In these cases I took the english version as a guide.

Most notable change besides that is that the player is now consistently addressed by the 2nd-person-plural form. There was a weird mix of this form and the 3rd-person-plural form, which was throwing the player out of the lore atmosphere. 
I decided to go with the 2nd-person-plural form since this is typlical in German for anything with a medieval theme and for fairy tales. Corresponding movies are also always translated like that into German (Game of Thrones, Lord of the Rings,...).

PS: this issue supersedes https://github.com/ihhub/fheroes2/issues/7848

### Describe the possible solution

~Use this file to update the existing in the repository: [de.zip](https://github.com/ihhub/fheroes2/files/12797205/de.zip)~
Merge the pull request #7865.

### Additional info

The changes were based on the most current revision of the file (https://github.com/ihhub/fheroes2/commit/70a635a16c8b0812a48748c0b54a42a3c35cf30e)

erazortt:
I created a pull request. So the zip file should not be needed anymore.

PR #7865, opened 2023-10-05
German translation fixes for issue #7860

erazortt:
Fix the German translation.
See issue #7860 for details


erazortt:
So do you have a way to cut the changes into pieces?

PR #7888, opened 2023-10-06
Make Grass as a default terrain for new maps in the Editor

ihhub:
Because a new map is always filled with Water so a user needs to make another selection to select another terrain type.


Districh-ru (approved):
I was also thinking about this change because selecting the grass instead of water is mostly my first action in editor. :)

PR #7890, opened 2023-10-06
Prioritize when hiring those heroes whose portraits are not used by customized heroes

### LINKED ISSUE #7878: Game, Adventure map, prioritize the use of hero portraits not already used in the jail for player heroes
### Preliminary checks

In fheroes2, prioritize the use of hero portraits not already used in the jail for player heroes.

In the example, the portrait of the hero Sir Gallant has been selected in the scenario editor to be assigned to the hero Anduran, who is being held prisoner in a jail in the northern center of the adventure map in the DOMINION.MX2 scenario (PoL scenario).
In this scenario, players do not have heroes placed on the adventure map by the scenario creator.
Players automatically get a random hero at the start of the scenario, by enabling the "Start with hero in each player's main castle" option in the scenario editor.

In scenario editor:

Sir Gallant twin editor 001
Sir Gallant twin editor 002
Sir Gallant twin editor 003

In fheroes2:
The blue and red players have a hero with the same portrait of Sir Gallant.
Sir Gallant twin 01

Here are the save file and the scenario file:
[Sir Gallant twin 01.zip](https://github.com/ihhub/fheroes2/files/12828827/Sir.Gallant.twin.01.zip)

In fheroes2, prioritize the use of hero portraits not already used in the jail for player heroes.
At least as long as there are other portraits available.

### Additional info

Version 1.0.8 7767

oleg-derevenetz:
close #7878

However, I should note that although this PR helps in the case of hiring a hero from the castle and assigning a starting hero (which is the case in #7878), this may not work for heroes placed on the map by the map creator. When loading the map, blocks are processed sequentially, and if the block with the hero goes first, and the block with the jail goes next, then at the time of processing the block with the hero (and assigning a specific hero to the corresponding tile), it is still unknown that there is also a jail on the map with a potentially customized hero. An improvement that also solves this problem would be somewhat more complicated and can be implemented sometime later if desired.

Also this PR optimizes the use of heroes with custom portraits by jails - instead of choosing a random hero and "remaking" it into the right one, the desired hero is used immediately (if possible). I've wanted to do this for a long time, but I kept forgetting.

There is no rush with this PR, it can potentially break something (although it shouldn't), so it is desirable to merge it after the 1.0.9 release.


oleg-derevenetz:
Hi @ihhub I would like to get your opinion on the following issue. In the process of testing this PR, it turned out that the PoL map editor can save maps as SW maps (with `.MP2` extension) even if this map uses the portraits of heroes from the PoL campaign (such as Jarkonas). If the player has the PoL resources, then everything works fine. However, if the player has just the SW resources, then certain problems arise - nothing crashes, of course, but the portrait of such a hero is not displayed:

polhero

Do we need to fix this somehow (for example, reassign another portrait) or not, or are there any other suggestions?

ihhub:
Hi @oleg-derevenetz , we shouldn't allow to even load such maps. I believe we need to add a check during map loading as well as save file loading.

ihhub:
@oleg-derevenetz , we need to add a check in **world_loadmap.cpp** file, **updateTileMetadata()** function, `case MP2::OBJ_HEROES`, under **if ( checkPoLObjects ) {** condition:
```cpp
        if ( checkPoLObjects ) {
            Heroes * hero = tile.getHero();
            assert( hero );
            if ( hero->isPoLPortrait() ) {
                return false;
            }
            ...
        }
```

ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for this update!

PR #7892, opened 2023-10-07
Allow to open the Editor only if The PoL version is used

ihhub:
Since the Editor is going to use extended features of all objects we need to use only the PoL. Plus this bans demo version to be exploited.


Districh-ru (commented):
Maybe here we should show a dialog, like we do for save files. Something like this:
"The editor uses extended features of all objects including the \"The Price of Loyalty\" expansion. But the corresponding game assets have not been provided to the engine."

ihhub:
Hi @Districh-ru , we actually disable the PoL campaign buttons is no PoL content present. Saves are different as we need to load them fully and verify the whole content to make sure that they aren't made for a PoL map.

Districh-ru:
> Hi @Districh-ru , we actually disable the PoL campaign buttons is no PoL content present. Saves are different as we need to load them fully and verify the whole content to make sure that they aren't made for a PoL map.

Hi, @ihhub, I just suggested to warn the user why he can't open the editor. The inability to play PoL campaigns and maps without PoL assets is obvious, but the inability to open the editor without PoL assets is not so obvious (also user may plan to design a map without any PoL objects).

ihhub:
Hi @zenseii , could you please review the added text content here?

Districh-ru (approved):

PR #7903, opened 2023-10-08
Remove extra space in dialog phrases

Districh-ru:
After we moved to a new Text rendering class there is no need to put a space on an empty line to force renderer no to skip it.
This PR removes extra space on empty lines and also does the same update in the translations.


ihhub (approved):

ihhub:
@Districh-ru , thank you very much for these fixes!

PR #7909, opened 2023-10-08
Rework Hot Keys dialog to support flexible size and Evil interface

### LINKED ISSUE #7829: hotkey menu is too narrow
### Preliminary checks

Hotkey menu is too narrow.

Make it wide, so everything could be seen.
PS I know than if you click on the line, appears a new window, where you can see full line.

### Additional info

_No response_

Districh-ru:
Fix #7829 
This PR makes Hot Keys dialog wider to fit most hot-key descriptions and adds support of evil interface.

evil-hotkeys

### DIFF
--- a/src/fheroes2/dialog/dialog_hotkeys.cpp
+++ b/src/fheroes2/dialog/dialog_hotkeys.cpp
@@ -18,49 +18,37 @@
  *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
  ***************************************************************************/
 
+#include "dialog_hotkeys.h"
+
+#include <algorithm>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:
Hi @Districh-ru , in the future we might need more of such windows so we should think how to reduce code duplication and come up with a single window being populated by different lists.

Districh-ru:
> Hi @Districh-ru , in the future we might need more of such windows so we should think how to reduce code duplication and come up with a single window being populated by different lists.

Hi @ihhub, sure!

Branikolog (approved):

ihhub:
@Districh-ru , huge thanks for this change!

PR #7916, opened 2023-10-09
Update Russian translation

Districh-ru:
Translate recently changed and added phrases

### DIFF
--- a/files/lang/ru.po
+++ b/files/lang/ru.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: fheroes2\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2023-10-09 14:08+0000\n"
-"PO-Revision-Date: 2023-10-08 13:59+0300\n"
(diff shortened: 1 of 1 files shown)


files/lang/ru.po
```diff
@@ -3317,9 +3317,8 @@ msgstr "Подтвердить Ваш выбор."
 msgid "Unknown Hero"
 msgstr "Неизвестный герой"
 
-#, fuzzy
 msgid "%{color} %{race} hero"
-msgstr "%{color} игрок"
+msgstr "%{race} - %{color} герой"
```
Branikolog, line 3321: Hi, @Districh-ru .
What do you think if we change the order here:
"%{color} герой - %{race}" or "%{color} герой ( %{race} )"
I think the colour type is more important information for map makers and it subjectively looks better.

Districh-ru: Hi, @Branikolog, sure, let the color be first.
I also translated one new phrase in the last commit: "(tent visited)"

Branikolog (approved):

ihhub:
@Districh-ru , thanks a lot for the update!

Issue #7801, opened 2023-09-23
Teach AI to use marketplace


fheroes2bugs:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Haven't found an issue, about marketplace.
AI can't use marketplace.

At map marooned AI (green) has a lot of strong neutral monsters surrounding him, and have town+marketplace. When I found him at 6 month, he had starting army and 2 heroes. He could trade gold for ore and build something, then buy units.

### Describe the possible solution

Teach AI to use marketplace

### Additional info

No save/video at this time.

ihhub:
I'm closing this issue as @oleg-derevenetz imemented basic logic for the AI to use marketplaces to build castles.

PR #7939, opened 2023-10-14
Teach AI to use the marketplace to accelerate the development of their castles

oleg-derevenetz:
Related to #7801

Trading on the market is currently used only for the development of castles (construction or upgrade of castle buildings and buying a boat), but not for hiring units or heroes. Trading Post objects on the map are also not supported yet.

This PR is mainly intended for the creation and testing a sort of infrastructure for the use of marketplaces by AI.

Also this PR gets rid of the separate `payment_t` type, which is nothing but an alias for `Funds`.


src/fheroes2/ai/normal/ai_normal_castle.cpp
```diff
@@ -45,73 +47,73 @@ namespace
...
     const std::vector<BuildOrder> defensiveStructures = { { BUILD_LEFTTURRET, 1 }, { BUILD_RIGHTTURRET, 1 }, { BUILD_MOAT, 1 }, { BUILD_CAPTAIN, 1 } };
-    const std::vector<BuildOrder> supportingDefensiveStructure = { { BUILD_MAGEGUILD1, 1 }, { BUILD_SPEC, 2 }, { BUILD_TAVERN, 1 } };
+    const std::vector<BuildOrder> supportingDefensiveStructures = { { BUILD_MAGEGUILD1, 1 }, { BUILD_SPEC, 2 }, { BUILD_TAVERN, 1 } };
 
     const std::vector<BuildOrder> & GetIncomeStructures( int type )
     {
-        static const std::vector<BuildOrder> standard = { { BUILD_CASTLE, 1 }, { BUILD_STATUE, 1 } };
-        static const std::vector<BuildOrder> warlock = { { BUILD_CASTLE, 1 }, { BUILD_STATUE, 1 }, { BUILD_SPEC, 1 } };
+        static const std::vector<BuildOrder> standard = { { BUILD_CASTLE, 1 }, { BUILD_MARKETPLACE, 1 }, { BUILD_STATUE, 1 } };
+        static const std::vector<BuildOrder> warlock = { { BUILD_CASTLE, 1 }, { BUILD_MARKETPLACE, 1 }, { BUILD_STATUE, 1 }, { BUILD_SPEC, 1 } };
```
ihhub, line 65: Shouldn't Statue be more valuable then marketplace? It generates Gold. The same applies for BuildOrder below.

src/fheroes2/castle/castle.h
```diff
@@ -254,13 +254,14 @@ class Castle : public MapPosition, public BitModes, public ColorBase, public Con
...
 
     bool BuyBuilding( uint32_t );
-    bool AllowBuyBoat() const;
-    bool BuyBoat() const;
     uint32_t GetBuildingRequirement( uint32_t ) const;
 
     int CheckBuyBuilding( const uint32_t build ) const;
     static int GetAllBuildingStatus( const Castle & );
 
+    bool AllowBuyBoat( const bool checkPayment = true ) const;
```
ihhub, line 262: Can we please remove the default value and change line 2540 of castle.cpp file to avoid having default values of input arguments? Since we literally use this function in 2 places I think it would be good to not to have the default values.

ihhub (commented):
Hi @oleg-derevenetz , I left 2 questions here. Could you please take a look?

src/fheroes2/ai/ai_common.cpp
```diff
@@ -19,53 +19,81 @@
...
+        // Perhaps the kingdom already has the necessary supply of resources
+        if ( !kingdom.AllowPayment( requiredFunds ) ) {
+            // Even if the kingdom does not have the necessary supply of these resources right now, there may be enough resources of another type available to get the
+            // missing resources as a result of resource exchange
+            if ( !calculateMarketplaceTransaction( kingdom, requiredFunds ) ) {
+                return false;
+            }
+        }
+
+        return BuildIfPossible( castle, building );
```
ihhub, line 96: Hi @oleg-derevenetz , I have a small concern about the approach: we allow the AI to change resources without any limits. It might be a dangerous approach in a long term. For example, the AI might trade gems in Wizard castle at the start of castle building but then it would need to trade gems back for bigger loss. I am sure how to tackle this issue elegantly.

However, another thing worries me is that we are trading resources with multiplier higher than 1. Essentially, we might have these resources but we still trade. Should we do trading only when the multiplier is 1?

oleg-derevenetz: Hi @ihhub

> It might be a dangerous approach in a long term. For example, the AI might trade gems in Wizard castle at the start of castle building but then it would need to trade gems back for bigger loss. I am sure how to tackle this issue elegantly.

In a long term there will be most likely multiple castles in the kingdom of a different types as well as multiple resource mines of a different type. It is impossible to calculate in advance. According to my observations, the use of markets accelerates the development of castles, not slows them down. This needs to be looked at based on the results of more extensive testing. If the castle is not able to build cheaper buildings, then what's the point of saving gems for more expensive ones if you build it much, much later?

> However, another thing worries me is that we are trading resources with multiplier higher than 1. Essentially, we might have these resources but we still trade. Should we do trading only when the multiplier is 1?

How is that? We never trade if we already have enough resources for this building.

src/fheroes2/ai/normal/ai_normal_castle.cpp
oleg-derevenetz: Hi @ihhub Yes, it is possible, given that the Market and the Statue require different types of resources for construction (Wood and Ore, respectively), it is possible to make the construction of the Statue a higher priority.

src/fheroes2/ai/ai_common.cpp
oleg-derevenetz: @ihhub the idea here is as follows: suppose we want to build a building with multiplier 2 that costs 5 Ore and 1000 Gold by having 5 Ore and 100000 Gold. It's obvious that we can buy additional 5 Ore for Gold if needed, so there is no point to refuse to build this building. That's why we make an adjustment using the `calculateMarketplaceTransaction()` here.

oleg-derevenetz: In fact, the `AllowPayment()` check is not required here really, `calculateMarketplaceTransaction()` will return the non-empty `std::optional` object with a no-op transaction (`Funds` with all zeroes) if there are already enough funds in the kingdom, `AllowPayment()` check is just much faster in this case, that's the only reason why it's here.

ihhub: These multipliers were added to not to spend last resources for buildings that aren't very important. For example, **BUILD_THIEVESGUILD** has multiplier 10. This is reasonable that Thieves' Guild is pretty useless for the AI. With the proposed logic the AI will trade resources to build this building which I don't think is so important. So in general we are dealing with 3 parameters:
1. order / importance of a building. Right now it is done by placing data into vectors
2. true cost of the building. This is constant value
3. multiplier or percentage of kingdom resources that can be taken to build the building

Taking all these parameters into an account we have 2 possible scenarios:
- try to build the first needed building. If cannot then trade resources and try again. This is what is currently implemented. Disadvantage: we might end up trading too many resources.
- the multiplier should affect the importance of the building. For example, BUILDING_1 has multiplier 5 and BUILDING_2 has multipler 1 and BUILDING_1 has higher priority. Should we try to build BUILDING_2 first since multiplier is 1 which means that the building is very valuable.

I am not asking to change anything for now. Just sharing my understanding.

What do you think?

oleg-derevenetz: Hi @ihhub

> With the proposed logic the AI will trade resources to build this building which I don't think is so important.

That's not how AI behaves with this PR - I mean, "not so important" buildings will remain "not so important", the required resources just will be, relatively speaking, "brought to the same denominator" using the marketplace(s). For instance, if the "not so important" building has the multiplier 2 and costs 5 Ore and 1000 Gold, then the kingdom should have either additional 5 ore or additional 12500 Gold (or a large supply of another resource).

This logic helps to develop the castles much faster. For instance, generic build order has the priority (or maybe "resource penalty") x2-x3 for `DWELLING_MONSTER2`, and x4 for `DWELLING_MONSTER1`, therefore, if the map does not have the necessary number of mines nearby, this slows down the castle development. I suggest you try to observe the AI castle development progress on the `master` branch and with this PR and compare them.

> the multiplier should affect the importance of the building. For example, BUILDING_1 has multiplier 5 and BUILDING_2 has multiplier 1 and BUILDING_1 has higher priority. Should we try to build BUILDING_2 first since multiplier is 1 which means that the building is very valuable.

Given the lack of resources, BUILDING_2 will have already been built by the time enough resources are finally accumulated for BUILDING_1 (taking into account the multiplier). And if trade is not allowed, then, if the kingdom simply does not have the necessary mines, then BUILDING_1 will never be built. This is not good at all.

ihhub: Thanks for the explanation!

ihhub (approved):

ihhub:
@oleg-derevenetz , awesome work! I tested the AI and now it's a beast :)

oleg-derevenetz:
Hi @ihhub just a single illustration. The "Seven Lakes" map, blue Warlock player. Both screenshots were made at 1-4-1 (that is, after three in-game weeks), the first one is with `master` branch, the second one is with this PR. Both games were played from the same save file made at 1-1-1.

<details>
<summary>master branch</summary>

master

No Crypt (remember x3 penalty for the `DWELLING_MONSTER2`, AI should have 30 Ore in cash to build it, which is unrealistic, at least on this map), therefore no Maze at all.
</details>

<details>
<summary>this PR</summary>

pr

Almost everything has been built, except for the dragons. And no Thieves' Guild - this is still the low-priority building, no resources were wasted :)
</details>

I have a suspicion that the players will soon cry and demand to disable AI trading for low difficulties of the game.

ihhub:
@oleg-derevenetz , we never claimed that we are making a weak AI...

PR #7958, opened 2023-10-18
Add support for the future new heroes

ihhub:
relates to #6845

- make Heroes::UNKNOWN entry first. Having this entry as non-zero was preventing to add new heroes in the future
- add much more checks in the code and rewrite certain functions to be human readable
- rearrange some code to have some sort of structure (it is still a huge mess)

This pull request needs very extensive testing to make sure that it doesn't break things:
- verify that all heroes from old saves are the same as loading the same files by master branch build.
- verify that all heroes from new saves are the same as before making a save.
- verify that heroes transferred between scenarios in campaigns are correct.
- verify that basic functionality with heroes for new and old saves is correct. This includes but not limited to checking heroes in castles, hiring and dismissing them.
- verify that winning and loosing conditions related to heroes are correct for new and old saves.
- verify that custom heroes set in maps are correct while starting a hero game for these maps.
- verify that the heroes order for Battle Only and the Editor is not broken.

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -2178,7 +2178,7 @@ Heroes * AllHeroes::GetHeroForHire( const int race, const int heroIDToIgnore ) c
 
 void AllHeroes::Scout( int colors ) const
 {
-    for ( Heroes * hero : *this ) {
+    for ( const Heroes * hero : *this ) {
         assert( hero != nullptr );
         if ( colors & hero->GetColor() ) {
             hero->Scout( hero->GetIndex() );
@@ -2336,7 +2336,7 @@ StreamBase & operator<<( StreamBase & msg, const AllHeroes & heroes )
 {
     msg << static_cast<uint32_t>( heroes.size() );
 
-    for ( const auto & hero : heroes ) {
+    for ( Heroes * const & hero : heroes ) {
         msg << *hero;
     }
 
@@ -2364,7 +2364,7 @@ StreamBase & operator>>( StreamBase & msg, AllHeroes & heroes )
         msg >> *heroes[0];
     }
     else {
-        for ( auto & hero : heroes ) {
+        for ( Heroes *& hero : heroes ) {
             hero = new Heroes();
             msg >> *hero;
         }
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -2263,8 +2263,7 @@ StreamBase & operator>>( StreamBase & msg, Heroes & hero )
             hero._id = Heroes::UNKNOWN;
             hero.portrait = Heroes::UNKNOWN;
         }
-        else
-        {
+        else {
             ++hero._id;
             ++hero.portrait;
         }
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -1967,8 +1967,7 @@ Heroes::HeroSeedsForLevelUp Heroes::GetSeedsForLevelUp() const
 
 double Heroes::getAIMinimumJoiningArmyStrength() const
 {
-    // Ideally we need to assert here that the hero is under AI control.
-    // But in cases when we regain a temporary control from the AI then the hero becomes non-AI.
+    assert( isControlAI() );
 
     double strengthThreshold = 0.05;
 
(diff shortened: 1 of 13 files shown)


src/fheroes2/heroes/heroes.cpp
```diff
@@ -1818,12 +1824,17 @@ const fheroes2::Sprite & Heroes::GetPortrait( int id, int type )
             return mediumSizePortrait.try_emplace( id, std::move( output ) ).first->second;
         }
         case PORT_SMALL:
-            return Heroes::DEBUG_HERO > id ? fheroes2::AGG::GetICN( ICN::MINIPORT, id ) : fheroes2::AGG::GetICN( ICN::MINIPORT, BRAX );
+            if ( id == Heroes::DEBUG_HERO ) {
+                return fheroes2::AGG::GetICN( ICN::MINIPORT, BRAX );
```
Districh-ru, line 1828:
```suggestion
                return fheroes2::AGG::GetICN( ICN::MINIPORT, BRAX - 1 );
```

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1921,21 +1932,83 @@ std::string Heroes::String() const
...
 void AllHeroes::Init()
 {
-    if ( !empty() ) {
-        AllHeroes::clear();
-    }
+    clear();
+
+    reserve( Heroes::HEROES_COUNT );
+
+    push_back( new Heroes( Heroes::UNKNOWN, Race::KNGT ) );
```
Districh-ru, line 2011: Why do we store data for the **UNKNOWN** hero? Is it ever used? :)
The `isValid()` method returns false for the UNKNOWN hero. So, IMHO, we can remove this data and get data from this vector this way (line 2078): `return ( *this )[hid - 1];`.
What do you think?

src/fheroes2/heroes/heroes.cpp
Districh-ru, line 1828: ICN index for the small portrait of Debug hero should also be deducted by 1.
```suggestion
                return fheroes2::AGG::GetICN( ICN::MINIPORT, BRAX - 1 );
```

Districh-ru (changes requested):
Hi, @ihhub, I quickly reviewed the PR and left two comments. Could you please check them in your spare time.
I'll do a more thorough test of this PR a little later.

Branikolog:
Hi, @ihhub 

I've started testing your PR recently.

Here's the issue I've found:

https://github.com/ihhub/fheroes2/assets/55348946/4101d504-c1aa-4f1f-9545-3ddf66be7640

[Heroes test.zip](https://github.com/ihhub/fheroes2/files/13069707/Heroes.test.zip)

Branikolog:
@ihhub 

Debug Hero from old save has wrong portrait now:

PR:

image

Master:

image

[AUTOSAVE.zip](https://github.com/ihhub/fheroes2/files/13069840/AUTOSAVE.zip)

src/fheroes2/heroes/heroes.cpp
ihhub: Hi @Districh-ru , I don't understand. Why we need to change the ICN value as this pull request does not touch images?

ihhub:
> Hi, @ihhub
> 
> I've started testing your PR recently.
> 
> Here's the issue I've found:
> 
> [Heroes test.zip](https://github.com/ihhub/fheroes2/files/13069707/Heroes.test.zip)
It is fixed. It was a bug in inverse logic.

src/fheroes2/heroes/heroes.cpp
ihhub: Ah, my bad. You are absolutely right!

Branikolog:
@ihhub 

> It is fixed. It was a bug in inverse logic.

I'll continue testing then, once it is compiled. 👍

oleg-derevenetz (changes requested):
Hi @ihhub I just left one minor and one optional suggestion, please take a look when you have time.

src/fheroes2/heroes/heroes.cpp
ihhub: Hi @Districh-ru , let's keep this first entry in the vector to have consistency with hero IDs as it is easier to debug the code. Later we should find another way how to avoid storing all this hero information as a more or less static array but it is beyond this pull request.

Districh-ru: Okay, let's do it this way.

Districh-ru (approved):

PR #7963, opened 2023-10-19
AI-controlled battle: increase the priority of those enemy units that have not yet got their turn

### LINKED ISSUE #7942: Battle: improve AI attack order
### Preliminary checks

AI range units and also melee attack order is up to down no matter if they have attacked yet or not.
In some situations, those centaurs got luck and killed my 1 archimag, but that would not happen, if centaurs, archers and grifons attacked the right troops.

https://github.com/ihhub/fheroes2/assets/145407420/db08cdb4-daf8-44f4-9c82-6b2da7331f4c

mistake cent

archers mistake

grif mistake

AI should attack those troops, that have not yet attacked.

### Additional info

version 1.09

save: 
[order of archers.zip](https://github.com/ihhub/fheroes2/files/12907480/order.of.archers.zip)

oleg-derevenetz:
close #7942

https://github.com/ihhub/fheroes2/assets/32623900/a0446f88-b23e-4470-8883-6582f9f4e194

However, now the defending Centaurs also choose attacking Centaurs (who didn't get their turn yet) as their target for the first shot instead of the Archmages (who already finished their turn), and as a result, the attacker suffers losses, which did not happen before.


src/fheroes2/battle/battle_troop.cpp
```diff
@@ -1182,30 +1171,35 @@ int32_t Battle::Unit::GetScoreQuality( const Unit & defender ) const
...
+        attackerThreat *= -2;
     }
-    // Finally ignore disabled units (if belong to the enemy)
+    // Ignore disabled enemy units
     else if ( attacker.isImmovable() ) {
         attackerThreat = 0;
     }
+    // Increase the priority of those enemy units that have not yet got their turn
+    else if ( !attacker.Modes( TR_MOVED ) ) {
+        attackerThreat *= 1.25;
```
oleg-derevenetz, line 1196: I'm not sure about the value of this coefficient.

ihhub (approved):

src/fheroes2/battle/battle_troop.cpp
ihhub: I don't have a better suggestion about this value as well. It should be more complex than just a multiplier but for now it is fine.

Branikolog:
Hi, @oleg-derevenetz 
That's really a nice battle improvement.

I've only have a little concerns about what happening in the battle now. Could you, please, explain, if this is expected behaviour, or something is going wrong?

>However, now the defending Centaurs also choose attacking Centaurs (who didn't get their turn yet) as their target for the first shot instead of the Archmages (who already finished their turn), and as a result, the attacker suffers losses, which did not happen before.

The defending centaurs attack player's centaurs in a case they haven't done a shot, that's good.
But why do they ignore archers in this case later?

https://github.com/ihhub/fheroes2/assets/55348946/961db845-a7eb-404f-96a4-1ed181146401

I've purposefully reduced the number of centaurs in hero army and even placed them lower than archers but defending centaurs still prefer attacking centaurs instead of archers.

ihhub:
@Branikolog , how the AI behaves on master branch?

oleg-derevenetz:
Hi @Branikolog I cannot reproduce this behavior. Both with this PR and on the `master` branch Centaurs prefer to attack neither Archers nor Centaurs but Archmages:

https://github.com/ihhub/fheroes2/assets/32623900/46394db1-49ec-4f82-b8c1-2c98bf000295

because Archmages have bigger value.

Branikolog:
Hi, @ihhub !
Since @oleg-derevenetz changed the logic of choosing the target by the AI, it beaves differently: AI focuses on Magi first. Then demolishes centaurs (until 4-6 are left in a stack) and only then starts shooting the rangers. So the problem of the priority of Centaurs over Rangers is actual in master brunch. But I suppose, this PR touches target selection too, since centaurs attack centaurs from time to time.

Branikolog:
@oleg-derevenetz 

Try this save file. (you can also switch Rangers and Centaurs, so defending AI attacks Centaurs only before the made their own attack). So they behave as you mentioned in the first message of the PR.

[Centaurs attack.zip](https://github.com/ihhub/fheroes2/files/13060532/Centaurs.attack.zip)

oleg-derevenetz:
@Branikolog

> Since @oleg-derevenetz changed the logic of choosing the target by the AI

What specific change do you have in mind? I don't recall any recent changes that could affect the choice of target when shooting.

Branikolog:
@oleg-derevenetz 

> @Branikolog
> 
> > Since @oleg-derevenetz changed the logic of choosing the target by the AI
> 
> What specific change do you have in mind? I don't recall any recent changes that could affect the choice of target when shooting.

I mean rising the priority of troops from this PR.

oleg-derevenetz:
@Branikolog

> I mean rising the priority of troops from this PR.

Ah, in that sense. Probably multiplication isn't the best way to raise a priority. Suggestions are welcome.

Branikolog:
@oleg-derevenetz 

I mostly concerned about the fact, that AI attacks Magi, Centaurs sometimes... But rangers are considered as a target only after he kills almost whole stacks of Magi and Centaurs. 24 Archers is definitely not a small threat, comparing to 2-3 Magi or 20 Centaurs. But he continues ignoring Rangers until the end of the battle (if it occurs)...

Multiplication definitely works. Maybe not 100% accurately, but it is. Maybe, coefficient shouldn't be not so high, as AI prefers shooting 18 centaurs (before their turn), than focusing 6 Magi (after their shot was made) in my save file.

Maybe there's also a problem with ranger's value, since AI doesn't want to attack them, while I personally consider them as the target +/- equal to a few Magi or lots of Centaurs.

oleg-derevenetz:
It seems that the skew in the archers value estimation arises due to this correction during the threat evaluation:

https://github.com/ihhub/fheroes2/blob/749507c3fde5c33850aac079ed575cf5d7c807c5/src/fheroes2/battle/battle_troop.cpp#L1205-L1207

I still don't really understand why this is done, I need to figure it out.

oleg-derevenetz:
Hi @Branikolog I have made a number of changes regarding the combat targets evaluation, and removed/rewritten some ancient position evaluation code. According to my tests, the problems with assessing the threat from the archers should be fixed now. It would be nice if you take a look at the current behavior when you have time and let me know if you notice any obvious errors.

Branikolog:
Hi, @oleg-derevenetz 
I've made a couple of tests.
Now the situation with Rangers is better. AI prefers Rangers before they made a shot and focuses other strong stack, after Rangers made a shot. I think this behaviour does not cover all cases, but makes shooting logic better in general and battles become more diverse. 👍 

I've only encountered strange case, when Centaurs attacked Rangers with the first Centaur stack and the last one (during one turn). It looks a little weird, since Magi are still the strongest target.

https://github.com/ihhub/fheroes2/assets/55348946/8eec7485-a393-465d-bc23-2422efa4371e

Cannot tell, it's bad, since in my opinion in the current case targeting Rangers is not really bad idea, since Centaurs have more chance to kill Archers, than Magi. But I'm just worried because of this randomness. 11 Magi is much stronger target, than 24 Rangers.

oleg-derevenetz:
Hi @Branikolog

> It looks a little weird, since Magi are still the strongest target.

Yes, they are a bit strongest (432 average damage vs 412 average damage of Rangers, because Rangers are shooting twice), BUT Archmagi already had their turn, while Rangers have yet to get their turn, so Archmagi have a penalty in the sense of priority. That's why Rangers are selected as a target first.

Branikolog:
@oleg-derevenetz 
But what about the cases, when the lowest Centaur stack targets Rangers? Both Magi and Rangers made their shot that turn.

oleg-derevenetz:
@Branikolog

> But what about the cases, when the lowest Centaur stack targets Rangers? Both Magi and Rangers made their shot that turn.

Yes, and since there are 23 Rangers and 11 Archmagi remain at that moment, the average damage of 23 Rangers is 396 here and average damage of 11 Archmagi is 395, and both have the same penalty, so Rangers are preferred.

Branikolog (approved):
@oleg-derevenetz 
I see. Thanks for clarifications!
Thank you for working on this improvement! The battles become really fun! :)

ihhub (approved):

ihhub:

Issue #7965, opened 2023-10-19
debug version and the original get the different result in battle


fheroes2bugs:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In autobattle I got different result in autobattle in debug version (debug = 5460) and the original one.

In original I don't lose unit(also my archers get morale):

https://github.com/ihhub/fheroes2/assets/145407420/cb987082-dcb6-47d0-bd79-6bbfe67539e1

In debug version I loose unit:

https://github.com/ihhub/fheroes2/assets/145407420/cef26d44-cdd9-46e8-86b7-8985bd5ac0d4

Sure I understand that lose of units is caused by random damage of units, but in original one I can play it 100 times and never get my unit killed, and in the second one (in debug version, when archers get luck) I always get my unit killed.

**The bug is in different luck in original fheroes and debug version.**
In debug version I don't get luck. 

### Save file

[autobattle - morale.zip](https://github.com/ihhub/fheroes2/files/13046268/autobattle.-.morale.zip)

### Additional info

version 1.09

oleg-derevenetz:
> The bug is in different luck in original fheroes and debug version.

What do you mean by "original fheroes"? If the "original fheroes" is the binary from our Releases section and "debug version" is the one that you built yourself, then it is inevitable for two reasons:

1. The logic of AI operation in newer sources may differ. AI makes different decisions during the battle, and it affects everything, including random events (including damage, luck and morale);
2. The operation of the random number generator may be different depending on the brand and version of the compiler. That is, MSVC 2019, MSVC 2022, Clang and GCC all may give different chains of pseudo-random numbers for the same seed.

ihhub:
Hi @oleg-derevenetz , I assume nothing we can do here as you correctly described the root of the problem, can't we?

oleg-derevenetz:
Hi @ihhub if debug and release versions are built from the different set of sources (or using different compilers) then this is inevitable due to reasons described above.

ihhub:
Closing this issue as nothing we can do.

PR #7967, opened 2023-10-20
Fix the random generator usage by the non-Arena code

### LINKED ISSUE #7964: Battle: morale in autobattle and player battle don't work the same way
### Preliminary checks

### Platform

Windows

### Describe the bug

Here is computer moving goblins and my opponent gets morale (double turn):

https://github.com/ihhub/fheroes2/assets/145407420/4891d303-b581-4b32-8a16-1343cdbf4ffc

Here is me moving goblins and my opponent don't get morale (doble turn):

https://github.com/ihhub/fheroes2/assets/145407420/96891d26-bc11-4148-9537-f6d96a8bfb39

PS also I have found, that in debug version (made from cmake) moral also works the different way.

### Save file

[autobattle - morale.zip](https://github.com/ihhub/fheroes2/files/13046094/autobattle.-.morale.zip)

### Additional info

version 1.09

oleg-derevenetz:
fix #7964

This PR prohibits the use of the `Battle::Arena`'s random generator by external code (mostly by AI decision-making code), which was the main reason for the non-reproducibility by a human of results of battles performed by AI.

It also redesignes the `Battle::Commands` in order to be able to perform additional checks at compile time and introduces a few minor improvements.

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -1027,7 +1027,9 @@ void Battle::Arena::ApplyActionCatapult( Command & cmd )
                     }
                 }
 
+#ifdef WITH_DEBUG
                 using TargetUnderlyingType = std::underlying_type_t<decltype( target )>;
+#endif
 
                 DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "target: " << static_cast<TargetUnderlyingType>( target ) << ", damage: " << damage << ", hit: " << hit )
             }
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -755,7 +755,7 @@ void Battle::Arena::TargetsApplySpell( const HeroBase * hero, const Spell & spel
     }
 }
 
-std::vector<Battle::Unit *> Battle::Arena::FindChainLightningTargetIndexes( const HeroBase * hero, Unit * firstUnit )
+std::vector<Battle::Unit *> Battle::Arena::FindChainLightningTargetIndexes( const HeroBase * hero, Unit * firstUnit, const bool applyRandomMagicResistance )
 {
     std::vector<Unit *> result = { firstUnit };
     std::vector<Unit *> ignoredTroops = { firstUnit };
@@ -778,9 +778,10 @@ std::vector<Battle::Unit *> Battle::Arena::FindChainLightningTargetIndexes( cons
     while ( result.size() != CHAIN_LIGHTNING_CREATURE_COUNT && !foundTroops.empty() ) {
         bool targetFound = false;
         for ( size_t i = 0; i < foundTroops.size(); ++i ) {
-            const int32_t resist = foundTroops[i]->GetMagicResist( Spell::CHAINLIGHTNING, heroSpellPower, hero );
-            assert( resist >= 0 );
-            if ( resist < static_cast<int32_t>( _randomGenerator.Get( 1, 100 ) ) ) {
+            const uint32_t resist = foundTroops[i]->GetMagicResist( Spell::CHAINLIGHTNING, heroSpellPower, hero );
+            assert( resist < 100 );
+
+            if ( !applyRandomMagicResistance || resist < _randomGenerator.Get( 1, 100 ) ) {
                 ignoredTroops.push_back( foundTroops[i] );
                 result.push_back( foundTroops[i] );
                 foundTroops.erase( foundTroops.begin() + i );
@@ -804,7 +805,7 @@ std::vector<Battle::Unit *> Battle::Arena::FindChainLightningTargetIndexes( cons
     return result;
 }
 
-Battle::TargetsInfo Battle::Arena::TargetsForChainLightning( const HeroBase * hero, int32_t attackedTroopIndex )
+Battle::TargetsInfo Battle::Arena::TargetsForChainLightning( const HeroBase * hero, const int32_t attackedTroopIndex, const bool applyRandomMagicResistance )
 {
     Unit * unit = GetTroopBoard( attackedTroopIndex );
     if ( unit == nullptr ) {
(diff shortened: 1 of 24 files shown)


oleg-derevenetz:
Hi @fheroes2bugs

Regarding #7964: I believe that in this PR I have eliminated the main reasons for the discrepancies that led to a different outcome of the battle when performing the same actions by the AI and the human player. According to my tests, now the outcome of the battle should be the same if the AI and the human player perform **exactly** the same actions (by "exactly" I mean maximum accuracy - for example, if you cast a spell on a wide unit, that spell must be cast on exactly the same cell of that unit).

Regarding #7965: according to my tests with this PR, the battle outcome is the same when performing absolutely the same actions in Debug and Release builds, if both Release and Debug builds were created from the same source code using the same compiler. Just in case, let me remind again that the same battle outcome in case of using different compilers is not currently guaranteed. And of course, this is not guaranteed in the case of using different versions of the game (built from different versions of the source code), because changes are regularly made to the logic of the combat AI.

src/fheroes2/battle/battle_action.cpp
```diff
@@ -835,7 +838,8 @@ Battle::TargetsInfo Battle::Arena::TargetsForChainLightning( const HeroBase * he
     return targets;
 }
 
-Battle::TargetsInfo Battle::Arena::GetTargetsForSpells( const HeroBase * hero, const Spell & spell, int32_t dest, bool * playResistSound /* = nullptr */ )
+Battle::TargetsInfo Battle::Arena::GetTargetsForSpell( const HeroBase * hero, const Spell & spell, const int32_t dst, bool applyRandomMagicResistance,
```
ihhub, line 841: A tiny code perfection: please add **const** to `bool applyRandomMagicResistance` as you did in other methods.

src/fheroes2/battle/battle_arena.h
```diff
@@ -173,7 +173,10 @@ namespace Battle
 
         void ApplyAction( Command & );
 
-        TargetsInfo GetTargetsForSpells( const HeroBase * hero, const Spell & spell, int32_t dest, bool * playResistSound = nullptr );
+        // Returns a list of targets that will be affected by the given spell applied to a cell with a given index
+        // (potentially by the given hero). This method can be used by external code to evaluate the applicability
+        // of a spell, and does not use probabilistic mechanisms to determine units resisting the given spell.
+        TargetsInfo GetTargetsForSpell( const HeroBase * hero, const Spell & spell, const int32_t dst );
```
ihhub, line 179: We can make this method **inlined** since it contains only one code line.

src/fheroes2/battle/battle_action.cpp
```diff
@@ -1026,7 +1036,11 @@ void Battle::Arena::ApplyActionCatapult( Command & cmd )
                     }
                 }
 
-                DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "target: " << target << ", damage: " << damage << ", hit: " << hit )
+#ifdef WITH_DEBUG
+                using TargetUnderlyingType = std::underlying_type_t<decltype( target )>;
+#endif
+
+                DEBUG_LOG( DBG_BATTLE, DBG_TRACE, "target: " << static_cast<TargetUnderlyingType>( target ) << ", damage: " << damage << ", hit: " << hit )
```
ihhub, line 1043: This line should be under ifdef block as well since it defines a type which is used here.

src/fheroes2/battle/battle_command.h
```diff
@@ -24,52 +24,137 @@
...
+        AUTO_FINISH
     };
 
-    class Command : public std::vector<int>
+    class Command final : public std::vector<int>
     {
     public:
-        explicit Command( const CommandType cmd );
-        Command( const CommandType cmd, const int param1, const int param2 = -1, const int param3 = -1, const int param4 = -1 );
+        static constexpr std::integral_constant<CommandType, CommandType::MOVE> MOVE{};
```
ihhub, line 58: I am just trying to understand this part of the code. We have an enumeration with items. We declare a set of structures (**struct integral_constant**) which is later used to interpret constructor in-place for **emplace_back()** calls (`Command( std::integral_constant<CommandType, cmd> /* tag */, const Types... params )`) and push necessary parameters into the **Command** class. Is my understanding correct?

src/fheroes2/battle/battle_command.h
```diff
@@ -24,52 +24,137 @@
...
+            else {
+                static_assert( sizeof...( params ) == 0 );
+            }
+
+            if constexpr ( sizeof...( params ) > 0 ) {
+                reserve( sizeof...( params ) );
+
+                // Put the elements of the parameter pack in reverse order using the right-to-left associativity of the assignment operator
+                int dummy = 0;
+                (void)( ( *this << params, dummy ) = ... );
```
ihhub, line 130: Could you please explain how this part of the code works? I understand that we are calling **Command & operator<<( const int val );** but I am confused with dummy variable.

ihhub (changes requested):
Hi @oleg-derevenetz , I left a set of comments and questions. Could you please take a look at them once you have free time?

src/fheroes2/battle/battle_action.cpp
oleg-derevenetz: Hi @ihhub this argument cannot be const because it will be changed inside this method in certain circumstances.

src/fheroes2/battle/battle_action.cpp
oleg-derevenetz: @ihhub DEBUG_LOG macro is already under such an ifdef internally.

src/fheroes2/battle/battle_command.h
oleg-derevenetz: @ihhub Yes. This is constructor template that allows to check the number of parameters at compile time depending on the command type. Since this check should be performed at compile time, the command type should be known at compile time, i.e. it should be a non-type template parameter. Since we cannot just set such a parameter for the constructor using an angle braces like `<CommandType::MOVE>foo()`, especially during various `emplace()` calls, the first parameter serves as a tag to let the template argument deduction figure out this non-type template parameter depending on type of the first argument.

src/fheroes2/battle/battle_action.cpp
ihhub: Ah, my bad.

src/fheroes2/battle/battle_command.h
ihhub: Thanks for the explanation!

src/fheroes2/battle/battle_command.h
oleg-derevenetz: @ihhub this is the fold expression that expands into `(*this << params[0], dummy) = ((*this << params[1], dummy) = ((*this << params[2], dummy) = ( ... = ((*this << params[n-1], dummy) = (*this << params[n], dummy)))))`. On each stage, the comma operator works as follows: it evaluates its left part first, then it evaluates its right part and returns the result of its evaluation - that is, it puts the Nth param to `this` first and then returns the value of `dummy` which will be reassigned to `dummy` on the next stage and immediately returned, and so on.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these useful changes!

PR #7969, opened 2023-10-20
Implement artifacts placing in Editor

Districh-ru:
This PR implements artifact placing and erasing in editor.
It also:
 - fixes information dialog for the Spell Scroll artifact and adds spell selection for this artifact;
 - puts the Spell Book to the top of the list in Battle Only mode;
 - hides the needed spell points for the Petrification (medusa) spell and random spells;
 - adds Random Spell icons:
277167504-5a4488db-52d6-414d-ba92-94ee63fa77ae;
 - adds Magic Book map sprite (discussible, it is a temporary sprite and can be changed in the other PR):
изображение;
- adds Treasures (resources) to the Erase tool.

https://github.com/ihhub/fheroes2/assets/113276641/15d593ed-cf82-4789-8ef8-33bad07fcb21


src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -485,10 +485,17 @@ Skill::Secondary Dialog::selectSecondarySkill( const Heroes & hero, const int sk
     return {};
 }
 
-Spell Dialog::selectSpell( const int spellId /* = Spell::NONE */ )
+Spell Dialog::selectSpell( const int spellId, const bool includeRandomSpells )
 {
     std::vector<int> spells = Spell::getAllSpellIdsSuitableForSpellBook();
 
+    if ( includeRandomSpells ) {
```
ihhub, line 492: Ultimate artifacts cannot be placed on the map as they do not have images (see ICN::OBJNARTI). We should exclude them from the Editor as well as Magic Book.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -521,9 +535,24 @@ Artifact Dialog::selectArtifact( const int artifactId /* = Artifact::UNKNOWN */
         listbox.SetCurrent( artifactId );
     }
 
-    const int32_t result = listbox.selectItemsEventProcessing( _( "Select Artifact:" ) );
+    if ( listbox.selectItemsEventProcessing( _( "Select Artifact:" ) ) == Dialog::OK || listbox.ok ) {
+        Artifact artifact( listbox.GetCurrent() );
+
+        if ( artifact.GetID() == Artifact::SPELL_SCROLL ) {
```
ihhub, line 541: Do we need to make a spell selection a scroll while placing it? Should we do this after putting it on the Adventure Map and changing its information?

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -515,6 +517,38 @@ namespace Interface
...
+            else if ( le.MouseClickLeft( _objectButtonsRect[Brush::ARTIFACTS] ) ) {
+                const Artifact artifact = Dialog::selectArtifact( _artifact.GetID(), true );
+                if ( artifact.GetID() != Artifact::UNKNOWN ) {
+                    _artifact = artifact;
+
+                    _interface.setCursorUpdater( [artifact]( const int32_t /*tileIndex*/ ) {
+                        uint32_t imageIndex = artifact.IndexSprite32() * 2 + 1;
+
+                        if ( imageIndex < 9 ) {
+                            // First four ultimate artifacts have no map images. Show "Ult ART" image instead.
```
ihhub, line 533: We should fix Ultimate artifact rendering. For now let's just put a TODO to do not forget.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -515,6 +517,38 @@ namespace Interface
...
+                        const fheroes2::Sprite & shadowImage = fheroes2::AGG::GetICN( ICN::OBJNARTI, imageIndex - 1 );
+                        const int32_t shadowImageWidth = shadowImage.width();
+                        const int32_t shadowImageHeight = shadowImage.height();
+                        const int32_t mainImageWidth = mainImage.width();
+                        const int32_t mainImageHeight = mainImage.height();
+
+                        fheroes2::Image image( mainImageWidth + shadowImageWidth, std::max( shadowImageHeight, mainImageHeight ) );
+                        image.reset();
+                        fheroes2::Copy( shadowImage, 0, 0, image, 0, shadowImage.y() - mainImage.y(), shadowImageWidth, shadowImageHeight );
+                        fheroes2::Copy( mainImage, 0, 0, image, shadowImageWidth, 0, mainImageWidth, mainImageHeight );
```
ihhub, line 547: In the future we need to come up with a function in this file which will glue given images into one for objects that occupy more than 1 tile. Either we can add a TODO comment here or we can create a function.

ihhub (changes requested):
Hi @Districh-ru , I left several comments here. Could you please take a look once you have time?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2414,39 +2414,72 @@ namespace fheroes2
...
-                    AlphaBlit( originalImage, image, 0, 0, 128 );
-                    AlphaBlit( originalImage, image, 4, 4, 192 );
-                    Blit( originalImage, image, 8, 8 );
-
-                    AddTransparency( image, 1 );
-                }
-
-                // The Petrification spell does not have its own icon in the original game.
-                h2d::readImage( "petrification_spell_icon.image", _icnVsSprite[id][66] );
+                if ( _icnVsSprite[id].size() == 60 ) {
```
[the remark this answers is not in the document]
Districh-ru: Sure!

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2414,39 +2414,72 @@ namespace fheroes2
...
+                    int32_t imageWidth = randomSpellImage.width();
+                    ApplyPalette( randomSpellImage, PAL::GetPalette( PAL::PaletteType::BROWN ) );
+                    Blit( randomSpellImage, randomSpellImage, true );
+
+                    // Add text on random spell images.
+                    for ( uint32_t i = 1; i < 6; ++i ) {
+                        Sprite & originalImage = _icnVsSprite[id][i + 67];
+                        Copy( randomSpellImage, originalImage );
+
+                        Text text( "spell", FontType::smallWhite() );
```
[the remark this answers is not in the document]
Districh-ru: Yes, I like this solution, thanks.
изображение

I left "lv." text and made it translatable. Later we will make the same texts for random items:
изображение

src/fheroes2/resource/artifact.h
```diff
@@ -171,6 +171,13 @@ class Artifact
         SWORD_ANDURAN,
         SPADE_NECROMANCY,
 
+        // Editor-related random artifacts.
```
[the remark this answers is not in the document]
Districh-ru: In original game it is a little mess with these artifacts. On map Random artifact (any level) has the same object ICN ID as the Magic Book. To avoid this types overlap it is better to add a new entry for `RANDOM_ALL_LEVELS` artifact. We cannot have two descriptions for the Magic Book and Random Artifact with the same ID.
We can replace all "Dummy #" artifacts with "Ultimate" and "Random level #" artifact entries. But there is no free place for Random (all levels) artifact. :(
What is you opinion how we could better separate Magic Book and Random artifact but use the same ID for them to avoid new entries creation?

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Done.
I made them placeable for test and also it can be a case when the map maker will want to place an ultimate artifact on the map protected by strong monsters and also place an other (diggable) Ultimate Artifact on the map.

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: It is done in the original map editor, but it takes more actions to place a Spell Scroll with a certain spell.
If a map maker wants to place a spell scroll then it's very likely he will also want to select a spell for it. If he does not want to make a selection right now he can simply press OKAY and a random spell will be set for the scroll.

Hello, @Branikolog, what is your opinion about when we should select a spell for the Spell Scroll artifact when we want to select and place it on map?

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: Done.
But if we do not plan to make them placeable on map like ordinary artifacts then these sprites are not necessary.

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: This function will be implemented in the next "Castles placing in Editor" PR and this code will be replaced with a function.
I left a TODO.

src/fheroes2/agg/agg_image.cpp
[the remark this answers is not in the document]
ihhub: Once the image is generated there is no way to translate the text. How about we use **normal yellow or white font** and just render a number as 1, 2, 3, 4 and 5?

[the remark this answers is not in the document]
ihhub: Something like this:

image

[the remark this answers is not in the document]
Districh-ru: Sure)
изображение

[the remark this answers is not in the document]
ihhub: Nice!

src/fheroes2/resource/artifact.h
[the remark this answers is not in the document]
Districh-ru: The Random and Ultimate artifacts set their own object IDs and I only added new ICN IDs for these artifacts so the logic should work fine because it is based on object Ids. :)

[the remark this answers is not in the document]
ihhub: The original Editor does not save random artifacts as **OBJ_ARTIFACT** but as **OBJ_RANDOM_ARTIFACT, OBJ_RANDOM_ARTIFACT_TREASURE, OBJ_RANDOM_ARTIFACT_MINOR and OBJ_RANDOM_ARTIFACT_MAJOR** so logically there is no overlap between Magic Book and Random Artifact.

I think we should rename Dummy into proper Random Artifacts. They aren't used anywhere else.

Magic Book does not have its own image for the Adventure Map (sadly). I believe it should but this is another topic for discussion. What we can do is to modify ICN::OBJNARTI array to add 2 empty images for Magic Book. Then IDs for random artifact will always correspond to their images. The only problem such a change would cause is PoL artifacts not having proper images on the Adventure Map. In this case we can add a fix for map loading and for saves. I can handle this change if you need my help.

What do you think about this?

src/fheroes2/spell/spell.cpp
```diff
@@ -132,12 +132,12 @@ spellstats_t spells[] = {
     { gettext_noop( "Set Air Guardian" ), 15, 0, 0, 53, 4, gettext_noop( "Sets Air Elementals to guard a mine against enemy armies." ) },
     { gettext_noop( "Set Fire Guardian" ), 15, 0, 0, 54, 4, gettext_noop( "Sets Fire Elementals to guard a mine against enemy armies." ) },
     { gettext_noop( "Set Water Guardian" ), 15, 0, 0, 55, 4, gettext_noop( "Sets Water Elementals to guard a mine against enemy armies." ) },
-    { "Random", 1, 0, 0, 0, 0, "Random" },
-    { "Random 1", 1, 0, 0, 0, 0, "Random 1" },
-    { "Random 2", 1, 0, 0, 0, 0, "Random 2" },
-    { "Random 3", 1, 0, 0, 0, 0, "Random 3" },
-    { "Random 4", 1, 0, 0, 0, 0, "Random 4" },
-    { "Random 5", 1, 0, 0, 0, 0, "Random 5" },
+    { gettext_noop( "Random spell" ), 0, 0, 0, 67, 0, gettext_noop( "Random selected any level spell." ) },
```
ihhub, line 135: Hi @zenseii , could you please review these lines?

src/fheroes2/resource/artifact.h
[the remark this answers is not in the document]
ihhub: For this matter should we use **OBJ_RANDOM_ARTIFACT_XXX** for random artifacts?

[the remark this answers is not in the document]
Districh-ru: > The original Editor does not save random artifacts as **OBJ_ARTIFACT** but as **OBJ_RANDOM_ARTIFACT, OBJ_RANDOM_ARTIFACT_TREASURE, OBJ_RANDOM_ARTIFACT_MINOR and OBJ_RANDOM_ARTIFACT_MAJOR** so logically there is no overlap between Magic Book and Random Artifact.

Yes, our editor does the same:
<details><summary>maps_tiles_helper.cpp, lines 3085-3104</summary>

```cpp
        switch ( artifactId ) {
        case Artifact::RANDOM_ALL_LEVELS:
            tile.SetObject( MP2::OBJ_RANDOM_ARTIFACT );
            break;
        case Artifact::RANDOM_1_LEVEL:
            tile.SetObject( MP2::OBJ_RANDOM_ARTIFACT_TREASURE );
            break;
        case Artifact::RANDOM_2_LEVEL:
            tile.SetObject( MP2::OBJ_RANDOM_ARTIFACT_MINOR );
            break;
        case Artifact::RANDOM_3_LEVEL:
            tile.SetObject( MP2::OBJ_RANDOM_ARTIFACT_MAJOR );
            break;
        case Artifact::RANDOM_ULTIMATE:
            tile.SetObject( MP2::OBJ_RANDOM_ULTIMATE_ARTIFACT );
            break;
        default:
            tile.SetObject( MP2::OBJ_ARTIFACT );
            break;
        }
```

</details> 

But the right mouse button artifact info and icon is not the same for the Magic Book and for the Random artifact. That is why I made a new entry for it instead of making hacks into right mouse button events handler. ... And placed all other random artifacts aside. :)

> I think we should rename Dummy into proper Random Artifacts. They aren't used anywhere else.

Sure, lets remove the Dummies and place Ult, Rand1, Rand2 and Rand3 instead. 

> Magic Book does not have its own image for the Adventure Map (sadly). I believe it should but this is another topic for discussion. What we can do is to modify ICN::OBJNARTI array to add 2 empty images for Magic Book. Then IDs for random artifact will always correspond to their images. The only problem such a change would cause is PoL artifacts not having proper images on the Adventure Map. In this case we can add a fix for map loading and for saves. I can handle this change if you need my help.

Yes, in future I was also planing to make ability to place the Magic Book on a map. So we should leave it.
I think we should place the `Artifact::RANDOM_ALL_LEVELS` after the PoL artifacts and do not change their indexes.
The new expansion artifacts will be added after this entry. Yes, this random artifact index will be not close to other random artifacts ... but is it so needed?
I just don't want to add new confusion that PoL artifact indexes will be shifted but SW will not. :)

src/fheroes2/dialog/dialog_selectitems.cpp
Branikolog: Hi, @Districh-ru !

I suggest selecting a spell right after we place an artifact on the adventure map.
Right now we select the spell and all scrolls we place have the same one, I suppose. Am I right?

https://github.com/ihhub/fheroes2/assets/55348946/bff22da7-d0e5-435c-a0fb-cc8b0498435d

It would be better to initiate spell selection dialog right after the scroll was placed on the adventure map, so we can place some different scrolls in a row.

Districh-ru: Hi, @Branikolog!
Yes, you are right. Your suggestion is reasonable. I'll do it in the next commit.

ihhub:
Hi @Districh-ru , could you please update this pull request with the latest changes for the Editor?

ihhub:
This pull request should be updated and merged after #8016.

src/fheroes2/editor/editor_interface_panel.cpp
ihhub: This was implemented recently :)

src/fheroes2/spell/spell.cpp
ihhub: We can revert these changes.

Districh-ru: > We can revert these changes.

The random spells will not be reverted. Only Random Artifact entries will be reverted.

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Hello, @Branikolog, this PR was updated: spells for the Spell Scroll are now selected when placing this artifact on map. Also a temporary sprite for the Magic Book was added.
Could you please review this PR when you have time.

Branikolog:
Hi, @Districh-ru 
I've found a few small issues. I'm not sure all of them are related to the current PR. I can create new topics for them in issues thread, if you wish.
So first of all I noticed that some artifacts are aligned differently in the list:
image
Scroll seemed to be closer to the right edge of artifact column.
While the book is placed aligned to the left:
image
UA artifact fell out of column at all:
image

I personally would love to have all artifacts aligned to the center of the column, since some of them have wide sprites, and some - thin (like scroll).


https://github.com/ihhub/fheroes2/assets/55348946/136da897-23c2-4b18-8275-d784587eca97


When cancelling scroll selection we can still place it on the adventure map. But the cursor does not switched to scroll. We should continue showing the artifact cursor of the last selected one.

Districh-ru:
Hi, @Branikolog!
> Hi, @Districh-ru I've found a few small issues. I'm not sure all of them are related to the current PR. I can create new topics for them in issues thread, if you wish. So first of all I noticed that some artifacts are aligned differently in the list: 

They are aligned the same way as on the map. Spell scroll is right-aligned in the tile, while the magic book currently is left-aligned. We can fix it in and discuss a better sprite for the magic book a separate issue/PR.
All artifact sprites contains of two tile-images: main and shadow (to the left). The Ultimate artifact does not has a shadow, so it is centered, but not aligned with all other two-tile artifacts. In the last commit I temporary added an empty image for the shadow to this artifact to align it.
изображение изображение

> When cancelling scroll selection we can still place it on the adventure map. But the cursor does not switched to scroll. We should continue showing the artifact cursor of the last selected one.

It is fixed in the last commit for Artifacts, Monsters, Treasures and Heroes.
Could you please check it in your spare time.

Branikolog:
@Districh-ru 

>We should return to sprites alignment after we implement placing more complex sprites, like mountains, Watchtowers, Castles and others to find a better way to display them.
Making a quick fix only for the artifacts might break rendering other objects, that are not implemented at the moment, but will use the seme sprite generation function.

Sure! A fix for just UA is enough as it looked the worst.

Cursor-artifact should be also shown in a case we return to artifact placing mode:

https://github.com/ihhub/fheroes2/assets/55348946/861ffe90-c0b4-4f03-b8c8-2b701382bc91

Districh-ru:
> Cursor-artifact should be also shown in a case we return to artifact placing mode:

@Branikolog, sure, fixed. :)

Branikolog (approved):
@Districh-ru 
Great!
Thanks for fixing this.

src/fheroes2/spell/spell.cpp
ihhub: My bad.

src/fheroes2/maps/map_object_info.cpp
```diff
@@ -41,76 +41,42 @@ namespace
...
-            objects.emplace_back( std::move( object ) );
+        for ( uint32_t artifactId = Artifact::SPELL_SCROLL; artifactId < Artifact::ARTIFACT_COUNT; ++artifactId ) {
+            addArtifactObject( artifactId, MP2::OBJ_ARTIFACT, artifactId * 2U - 1U );
         }
 
-        {
-            Maps::ObjectInfo object{ MP2::OBJ_RANDOM_ARTIFACT_MINOR };
-            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNARTI, 167, fheroes2::Point{ 0, 0 }, MP2::OBJ_RANDOM_ARTIFACT_MINOR, Maps::OBJECT_LAYER );
-            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNARTI, 166, fheroes2::Point{ -1, 0 }, MP2::OBJ_NONE, Maps::SHADOW_LAYER );
+        // Assign the Magic Book temporary made sprites.
```
ihhub, line 65: Should we move this line before SPELL_SCROLL to be in a logical order with the enumeration?
```cpp
for ( uint32_t artifactId = Artifact::ARCANE_NECKLACE; artifactId < Artifact::MAGIC_BOOK; ++artifactId ) {
    addArtifactObject( artifactId, MP2::OBJ_ARTIFACT, artifactId * 2U - 1U );
}

// Assign the Magic Book temporary made sprites.
addArtifactObject( Artifact::MAGIC_BOOK, MP2::OBJ_ARTIFACT, 207U );

for ( uint32_t artifactId = Artifact::SPELL_SCROLL; artifactId < Artifact::ARTIFACT_COUNT; ++artifactId ) {
    addArtifactObject( artifactId, MP2::OBJ_ARTIFACT, artifactId * 2U - 1U );
}
```

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -1168,6 +1168,36 @@ namespace
...
+            }
+
+            for ( const Maps::TilesAddon & addon : currentTile.getBottomLayerAddons() ) {
+                if ( addon._uid == uid ) {
+                    currentTile.Remove( uid );
+                    removeUidFromTilesAround( tileIndex, uid );
+                    return;
+                }
+            }
+            for ( const Maps::TilesAddon & addon : currentTile.getTopLayerAddons() ) {
```
ihhub, line 1191: Please add an empty line above for easier readability.

src/fheroes2/spell/spell.cpp
```diff
@@ -132,13 +132,13 @@ spellstats_t spells[] = {
...
     { gettext_noop( "Set Fire Guardian" ), 15, 0, 0, 54, 4, gettext_noop( "Sets Fire Elementals to guard a mine against enemy armies." ) },
     { gettext_noop( "Set Water Guardian" ), 15, 0, 0, 55, 4, gettext_noop( "Sets Water Elementals to guard a mine against enemy armies." ) },
-    { "Random", 1, 0, 0, 0, 0, "Random" },
-    { "Random 1", 1, 0, 0, 0, 0, "Random 1" },
-    { "Random 2", 1, 0, 0, 0, 0, "Random 2" },
-    { "Random 3", 1, 0, 0, 0, 0, "Random 3" },
-    { "Random 4", 1, 0, 0, 0, 0, "Random 4" },
-    { "Random 5", 1, 0, 0, 0, 0, "Random 5" },
-    { gettext_noop( "Petrification" ), 1, 0, 0, 66, 0,
+    { gettext_noop( "Random spell" ), 0, 0, 0, 67, 0, gettext_noop( "Random selected any level spell." ) },
```
ihhub, line 135: All spell names contains capital letters in their words. I suggest to use the same approach for Random Spells.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3662,6 +3692,46 @@ namespace fheroes2
                     }
                 }
                 return true;
+            case ICN::OBJNARTI:
+                LoadOriginalICN( id );
+                if ( _icnVsSprite[id].size() == 206 ) {
+                    // If we have the Price of Loyalty assets we nake a map sprite for the Magic Book artifact.
```
ihhub, line 3698: **nake** is misspelled.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -357,6 +357,57 @@ namespace Interface
...
+    void EditorPanel::_setCursor()
+    {
+        if ( _selectedInstrument != Instrument::OBJECT ) {
+            _interface.setCursorUpdater( {} );
+            return;
+        }
+
+        switch ( _selectedObject ) {
+        case Brush::MONSTERS:
+            if ( _monsterType < 0 ) {
```
ihhub, line 369: **setCustomCursor** function handles invalid cases. Is there a reason why we are doing the check here again as these cases aren't supposed to happen?

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -3058,6 +3099,24 @@ namespace Maps
...
+            return false;
+        }
+
+        tile.Remove( uid );
+
+        // An object may occupy several tiles, we recursively check all around tiles for its parts by UID.
+        removeUidFromTilesAround( tile.GetIndex(), uid );
+
+        resetObjectMetadata( tile );
+        tile.setAsEmpty();
```
ihhub, line 3115: Please add a TODO comment as it might not be correct to do this. Consider this situation:
- put an artifact on a map
- put a Mine on the map that its top layer is on the same level as the artifact
- erase the artifact

I believe the tile will be marked as OBJ_NONE while in reality it should be OBJ_NON_ACTION_MINES.

#8019 add a logic to set tile's object type which is correlated to this logic.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -627,6 +632,30 @@ namespace Interface
...
+                fheroes2::showStandardTextMessage( _( "Artifacts" ), _( "Artifacts cannot be placed on water." ), Dialog::OK );
+            }
+            else if ( !Maps::isClearGround( tile ) ) {
+                fheroes2::showStandardTextMessage( _( "Artifacts" ), _( "Choose a tile which does not contain any objects." ), Dialog::OK );
+            }
+            else {
+                const int32_t artifactType = _editorPanel.getArtifactType();
+
+                // For each Spell Scroll artifact we select a spell.
+                if ( Maps::getObjectsByGroup( Maps::ObjectGroup::Artifact )[artifactType].metadata[0] == Artifact::SPELL_SCROLL ) {
```
ihhub, line 646: This is risky. Please assert **artifactType** over the size of the returned array and logic check the type of the object.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -627,6 +632,30 @@ namespace Interface
...
+                    const int spellId = Dialog::selectSpell( Spell::RANDOM, true ).GetID();
+
+                    if ( spellId != Spell::NONE ) {
+                        setObjectOnTile( tile, Maps::ObjectGroup::Artifact, _editorPanel.getArtifactType() );
```
ihhub, line 650: This line of code can be moved outside the if-else-body and put before it since we don't have any rendering calls before setting the spell.

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -1679,6 +1709,17 @@ namespace Maps
         tile.metadata()[1] = value;
     }
 
+    void setSpellScrollSpellId( Tiles & tile, const uint32_t spellId )
```
ihhub, line 1712: We can modify setSpellOnTile() function to handle OBJ_ARTIFACT and assert if the artifact is not SPELL_SCROLL. In this case we don't need a separate function to do the same thing..

ihhub (changes requested):
Hi @Districh-ru , I left a couple of comments here. Could you please take a look at them when you have time?

src/fheroes2/maps/map_object_info.cpp
Districh-ru: Hi, @ihhub, in original editor there is no Magic Book and PoL artifacts go right after the SW artifacts. Also we already brake the order not including "Dummy" artifacts. So I added the Magic Book to the end as a new map artifact.
Also maybe we should place Random artifacts in the start of the list, like in original editor.
@Branikolog what are your thought about the place for the Magic Book and Random artifacts?

src/fheroes2/editor/editor_interface_panel.cpp
Districh-ru: I did this change to `setCustomCursor()`. It can be reverted if you think that it is not good.

src/fheroes2/maps/maps_tiles_helper.cpp
Districh-ru: Sure, we need to better handle these cases and use other (new) function here.

src/fheroes2/maps/maps_tiles_helper.cpp
Districh-ru: You are right, thanks!

src/fheroes2/maps/map_object_info.cpp
ihhub: Hi @Districh-ru , it doesn't matter how the list shows in the Editor as our artifact object list has no connection with the list there. My suggestion is to align with the enumeration for artifacts we have at the moment.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -50,8 +50,14 @@ namespace
...
     {
+        if ( type == -1 ) {
+            // The object type is not set. We show the POINTER cursor for this case.
+            Cursor::Get().SetThemes( Cursor::POINTER );
+            return;
+        }
+
         const auto & objectInfo = Maps::getObjectsByGroup( group );
-        if ( type < 0 || type >= static_cast<int32_t>( objectInfo.size() ) ) {
+        if ( type < -1 || type >= static_cast<int32_t>( objectInfo.size() ) ) {
```
ihhub, line 60: We should keep the original logic as it is more understandable even if we handle the **-1** case above.

src/fheroes2/maps/map_object_info.cpp
ihhub: I also suggest to rename DUMMY into **UNUSED_83..86** so it will be clear what they are for, and update the corresponding comment.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -627,6 +632,36 @@ namespace Interface
...
+                if ( artifactInfo[artifactType].objectType == MP2::OBJ_ARTIFACT && artifactInfo[artifactType].metadata[0] == Artifact::SPELL_SCROLL ) {
+                    const int spellId = Dialog::selectSpell( Spell::RANDOM, true ).GetID();
+
+                    if ( spellId == Spell::NONE ) {
+                        // We do not place the Spell Scroll artifact if the spell for it was not selected.
+                        return;
+                    }
+
+                    setObjectOnTile( tile, Maps::ObjectGroup::Artifact, _editorPanel.getArtifactType() );
+                    Maps::setSpellOnTile( tile, spellId );
```
ihhub, line 658: **setSpellOnTile()** need the corresponding code to handle **MP2::OBJ_ARTIFACT**. It is missing at the moment.

ihhub (changes requested):
Hi @Districh-ru , I put a few more comment here. Could you please take a look at them when you have spare time?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Hi, @ihhub. The code is already there (lines 1428-1432) :)

https://github.com/Districh-ru/fheroes2/blob/e9ff79691f942290ac2bfdcbcc72d2aa23d9b376/src/fheroes2/maps/maps_tiles_helper.cpp#L1419-L1438

ihhub: Oh, yes. I somehow totally missed this. My bad :(

ihhub (approved):

ihhub:
@Districh-ru , huge thanks for implementing this feature!

PR #7975, opened 2023-10-21
Fix the logic of purchasing a boat in shipyard

### LINKED ISSUE #7970: Cannot build a boat in a specific castle
### Preliminary checks

### Platform

Linux

### Describe the bug

In the castle Erliquin (the last of three castles owned by a player, a wizard's one) it is impossible to build a boat. If you try game takes your resources for a boat, game plays a sound of a boat materializing but it turns out it is a fraud, no boat created.

### Save file

[bug.zip](https://github.com/ihhub/fheroes2/files/13059558/bug.zip)

### Additional info

_No response_

ihhub:
close #7970

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -496,9 +496,10 @@ void Castle::PostLoad()
             JoinRNDArmy();
     }
 
-    // fix shipyard
-    if ( !HaveNearlySea() )
+    // Remove shipyard if no sea terrain present.
+    if ( !HasSeaAccess() ) {
         building &= ~BUILD_SHIPYARD;
+    }
 
     // remove tavern from necromancer castle
     if ( Race::NECR == race && ( building & BUILD_TAVERN ) ) {
@@ -1476,7 +1477,7 @@ int Castle::CheckBuyBuilding( const uint32_t build ) const
         }
         break;
     case BUILD_SHIPYARD:
-        if ( !HaveNearlySea() ) {
+        if ( !HasSeaAccess() ) {
             return BUILD_DISABLE;
         }
         break;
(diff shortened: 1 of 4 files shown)


src/fheroes2/castle/castle.cpp
```diff
@@ -2155,42 +2156,113 @@ Heroes * Castle::GetHero() const
...
+
+    if ( Maps::isValidAbsPoint( possibleSeaTile.x + 1, possibleSeaTile.y ) && doesTileHaveBoat( world.GetTiles( index + 1 ) ) ) {
+        return true;
+    }
+
+    return false;
 }
 
-bool Castle::PresentBoat() const
+int Castle::getTileIndexToPlaceBoat() const
```
oleg-derevenetz, line 2233: Hi @ihhub map indexes are usually `int32_t`, even in this function `index` is `int32_t`. It would be nice if this function followed the general logic and also returned `int32_t` instead of just `int`.

src/fheroes2/castle/castle.cpp
```diff
@@ -2541,32 +2609,21 @@ bool Castle::BuyBoat() const
         return false;
     }
 
-    if ( !Maps::isValidAbsPoint( center.x, center.y + 2 ) ) {
+    // If this assertion blows up then you didn't even check conditions to build a shipyard!
+    assert( HasSeaAccess() );
+
+    const int index = getTileIndexToPlaceBoat();
```
oleg-derevenetz, line 2615: And accordingly, this variable should also be of type `int32_t`.

src/fheroes2/castle/castle.cpp
ihhub: Thanks for spotting it!

oleg-derevenetz (approved):

Districh-ru (changes requested):
Hi, @ihhub, I left one comment regarding sea access check. Could you please take a look when you have time.

Districh-ru (approved):

drevoborod:
Now it's possible to buy a boat, but it's impossible to occupy it by a hero - it becomes unreacheable. And, of course, after that it's impossible to buy new boat, because there is one already. So, we have here some lock situation, which doesn't seem to be ok.

oleg-derevenetz:
Hi @drevoborod

> Now it's possible to buy a boat, but it's impossible to occupy it by a hero - it becomes unreacheable. And, of course, after that it's impossible to buy new boat, because there is one already. So, we have here some lock situation, which doesn't seem to be ok.

IIRC there is nothing we can do with this, because there is just one tile suitable for a boat near this castle and it is unreachable because of a tree. BUT this boat can be summoned once built, so it's still a progress.

drevoborod:
> P.S. In the original game there is funny behavior in this castle:
> DOSBox.0.74-2.1.Cpu.speed_.max.100.cycles.Frameskip.0.Program_.HEROES2.2023-10-24.22-50-20.mp4
> 
> There is no boat on the adventure map nearby, but there is a boat in the castle and Shipyard doesn't allow to buy another one.

Yes, I also checked the original game behaviuor and the result was same if there was already a shipyard on start; if there was no shipyard and I bought it myself, the message appeared (It's Knight castle but I don't think it matters): 
изображение

ihhub:
Thanks @drevoborod !

PR #7982, opened 2023-10-23
Remove special case for castle passability

ihhub:
This PR removes special passability for castles since the normal case passability works exactly the same.


Districh-ru (approved):
Castles passability is OK.
Master build:
изображение
This PR:
изображение

Branikolog (approved):

PR #7983, opened 2023-10-24
Move the barrier color constants to the anonymous namespace and fix the patrol zone for imprisoned heroes

oleg-derevenetz:
Fixing the patrol zone is merely a cosmetic action, but it helps to avoid some confusion.


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for these modifications!

PR #7986, opened 2023-10-24
Use the battle pathfinder's distance estimation when finding an optimal position for attack during the next turn

### LINKED ISSUE #7972: Combat, "safe" AI move doesn't consider obstacles on a battlefield
### Preliminary checks

### Platform

Windows

### Describe the bug

https://github.com/ihhub/fheroes2/assets/55348946/f11d2290-a840-4956-9667-dfb9eb13d8ed

As you can see on the video, wolves being afraid of enormous stacks of boars prefer to take a "safe" position, even in those areas, where boars couldn't reach because of obstacles.

I've also met similar issue, when during siege attacking champions ignore castle gates and make just 1 step forwards, instead of coming close to walls, to be able to step in a moat if some walls were destroyed.

### Save file

[Broken_Alliance master.zip](https://github.com/ihhub/fheroes2/files/13060336/Broken_Alliance.master.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #7972

https://github.com/ihhub/fheroes2/assets/32623900/42345880-f723-4529-ad3c-9565cf9431ad

Also this PR improves AI behavior in some other cases:

<details>
<summary>AI should not try to kite non-blocked archers</summary>

`master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/c73c097e-1468-4f18-87fe-8a754c7e2691

This PR:

https://github.com/ihhub/fheroes2/assets/32623900/162cbaa4-378a-4969-85bf-a0bb1f3293f4
</details>

<details>
<summary>Better damage estimation by AI</summary>

`master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/e976d752-09dd-4336-81b8-f6048d17b36e

This PR:

https://github.com/ihhub/fheroes2/assets/32623900/70ce364e-2b4a-472e-98c8-77fcdfecb341
</details>

as well as in a few others.


Branikolog:
Hi, @oleg-derevenetz 

I found out that with current PR defending ranged troops are not really fond of changing position, especially, if the speeds are equal.

https://github.com/ihhub/fheroes2/assets/55348946/46a01dda-6b89-46fa-9737-60da17424070

oleg-derevenetz:
Hi @Branikolog

> I found out that with current PR defending ranged troops are not really fond of changing position, especially, if the speeds are equal.

Yes, this is done intentionally. Shooters runs away only if there is a real chance to escape.

Branikolog:
@oleg-derevenetz 

> Otherwise it turns into just running around and shooter will still be overtaken, just a little later.

Just in case 1vs1.
But if there are multiple defending stacks they could run in opposing directions, so attacker won't cover both...
Master:

https://github.com/ihhub/fheroes2/assets/55348946/8540f43f-c304-4135-a941-11a0370a0abf

PR:

https://github.com/ihhub/fheroes2/assets/55348946/9421eadc-e1e7-4334-afae-78b514cf0b2b

oleg-derevenetz:
@Branikolog

> But if there are multiple defending stacks they could run in opposing directions, so attacker won't cover both...

In the current paradigm, there is no interaction between stacks in regard to planning, each stack makes a decision separately. When (and if) some operative level of planning is introduced, then it will be possible to implement some kind of purposeful run-up like that. But not within the scope of this PR :)

Branikolog:
@oleg-derevenetz 

>When (and if) some operative level of planning is introduced, then it will be possible to implement some kind of purposeful run-up like that. But not within the scope of this PR :)

Sure! I am not insisting on doing everything in just a single PR. Just want to share my ideas with you.

>In the current paradigm, there is no interaction between stacks in regard to planning, each stack makes a decision separately. 

What if we simply make shooters run (with equal or 1 speed slower) (the situation from my video above) for the cases when the number of defending stacks exceeds attacking stacks? Like 2vs1, 3vs2? There's rather high possibility to evade from attacker and make 1 extra shot, if you have more than 1 stack, which can bite attacker to come farther away from one of stacks allowing it to make additional shot.

oleg-derevenetz:
@Branikolog

> What if we simply make shooters run (with equal or 1 speed slower) (the situation from my video above) for the cases when the number of defending stacks exceeds attacking stacks?

Well, this will be a half-hearted solution because without a coordination mechanism, fleeing stacks of archers can run in one direction, and not in different directions (and therefore can be overtaken by one unit), and there is no such coordination now. It will look silly :) It is better to think over and implement some sort of more general planning first.

Branikolog:
@oleg-derevenetz 

>It is better to think over and implement some sort of more general planning first

I hope you or some our other members manage with that some day.
Thank you for working on this anyway!

oleg-derevenetz:
Hi @Branikolog BTW (unrelated to this PR) Have you tried to play around with the recently made changes regarding Marketplace support for AI? Has AI started to develop their castles and economy better according to your observations?

Branikolog:
@oleg-derevenetz 

> Hi @Branikolog BTW (unrelated to this PR) have you tried to play around with the recently made changes regarding Marketplace support for AI? Has AI started to develop their castles and economy better according to your observations?

Not yet. I was mostly focused on UI and battles. Haven't played the entire map for a while.

Branikolog (approved):
With this PR ranged troops kite only in a case they have a higher speed.
Obstacles are also included into position planning. 👍 
I hope soon we come up with an maneuvering logic, which helps rangers to evade or assist in killing blocking attacker.

ihhub:
@oleg-derevenetz , huge thanks for this update!

PR #7990, opened 2023-10-25
Fix several issues with loading ICN resources

### LINKED ISSUE #7987: Wrong behavior of the "Continue Movement" button
### Preliminary checks

### Platform

Windows

### Describe the bug

The first problem that is 100% reproducible for me:

https://github.com/ihhub/fheroes2/assets/32623900/e44bea7a-cbc0-44ad-b75c-d3f22375d062

The second problem that is 100% reproducible for me as well:

https://github.com/ihhub/fheroes2/assets/32623900/a4d4532d-ac12-4e9b-85fd-a96215ff33c8

Both save files are attached to this issue.

### Save file

[save.zip](https://github.com/ihhub/fheroes2/files/13127806/save.zip)

### Additional info

_No response_

ihhub:
fix #7987

We must use **LoadOriginalICN()** function only when dealing with the same ICN id.

As a side bonus: we won't load certain ICN resources twice as it was before the changes.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -2637,7 +2637,7 @@ namespace fheroes2
                 }
                 return true;
             case ICN::TROLL2MSL:
-                LoadOriginalICN( ICN::TROLLMSL );
+                LoadModifiedICN( ICN::TROLLMSL );
                 if ( _icnVsSprite[ICN::TROLLMSL].size() == 1 ) {
                     _icnVsSprite[id].resize( 1 );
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):
Hi @ihhub Thank you, this change seems to fix both issues I reported. I would just like to recommend adding some description in the code regarding cases when it is worth calling `LoadOriginalICN()` and cases in which it is not worth doing (and `loadICN()` should be used instead). Otherwise, I suspect that in the future we may face similar problems again.

ihhub:
@oleg-derevenetz , it is done. I also added an assertion to prevent possible recursions.

ihhub:
@Districh-ru and @oleg-derevenetz , I also found 2 more "bad" place with resources loading. Hopefully, newly added assertions can help to prevent invalid logic in the future.

PR #8002, opened 2023-10-29
Rename .appdata.xml file to .metainfo.xml and add <launchable/> tag

Alexander-Wilms:


ihhub:
Hi @Alexander-Wilms , could you please explain the purpose of this pull request?

Alexander-Wilms:
@ihhub The Free desktop spec recommends the more new and more universal .metainfo.xml extension:

https://freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location

Also, the launchable tag is needed to make the file pass validation:

`flatpak run org.freedesktop.appstream.cli validate --strict script/packaging/common/fheroes2.metainfo.xml`

script/tools/changelog_update_appdata.py
```diff
@@ -1,62 +0,0 @@
-#!/usr/bin/env python3
```
ihhub, line 1: Could you please revert this change and rename the original file using **git mv** command? In this case the file content will be preserved and only difference (if any) is going to be shown.

ihhub (changes requested):
Hi @Alexander-Wilms , I left one comment here. Could you please take a look at it?

script/tools/changelog_update_appdata.py
Alexander-Wilms: Done. But it's only shown as being renamed when viewing that specific commit:

https://github.com/ihhub/fheroes2/pull/8002/commits/6a6a56d6459d556334ae24ef22606b2e80583b12

ihhub (approved):

ihhub:
@Laserlicht , could you please take a look a these changes as well?

Laserlicht (approved):

ihhub:
@Alexander-Wilms , thank you so much for your contribution to the project!

PR #8003, opened 2023-10-30
Remove the END_TURN battle command, improve command argument checks, fix a few issues found thanks to improved checks

oleg-derevenetz:
* The `END_TURN` battle command has been removed. It was essentially a useless command that only made it difficult to control the correctness of the arguments of other commands, especially the attack command, because the attack was a composite action of moving and attacking, and when it came time to check the correctness of the arguments of the `ATTACK` command, the attacking unit was already in the wrong place from which it started its attack.
* The verification of the arguments of battle commands has become stricter. Also, now if there is an error in the command arguments or preconditions, the corresponding error is logged using `ERROR_LOG()` macro and, additionally, if executable was built with the `WITH_DEBUG` defined, assertion failure is raised. This has already helped to catch some issues (see below).
* Some AI-only evaluation code has been moved from `Battle::Board` to the AI battle planner.
* Improved logging of the AI battle decisions.

<details>
<summary>Issue with attack from the moat</summary>

With `master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/bc7fbe40-278e-4527-9262-050d87a7955f

There is nothing inherently wrong with this attack, except that it was performed from the wrong cell. Wide units are able to attack from the moat, but only from the one of their cells that is not in the moat - i.e. in this case Cavalry should attack from their tail cell.

With this PR:

https://github.com/ihhub/fheroes2/assets/32623900/bfe7fb05-9739-4abb-8774-1a4f5871d6c3

I should also note that, although this attack is absolutely legal in fheroes2, a human player is limited in carrying it out due to some imperfection of battle management, although in certain circumstances human player is also quite capable of carrying out such an attack:

https://github.com/ihhub/fheroes2/assets/32623900/e81c0fdd-f72a-471c-8c8f-2b2a7ed82db3
</details>


ihhub (approved):

ihhub:
Hi @Branikolog , can you please do some sanity tests for battles to make sure that the changes don't break anything?

ihhub:
@oleg-derevenetz , thank you very much for the changes!

PR #8010, opened 2023-11-02
Add support for right SHIFT, CTRL and ALT hot keys

ihhub:
As requested by players in Discord.


Districh-ru (commented):
Hi @ihhub, the right SHIFT and CTRL keys works good, but perhaps it will be good also to add the **right ALT** key. What do you think?

ihhub:
> Hi @ihhub, the right SHIFT and CTRL keys works good, but perhaps it will be good also to add the **right ALT** key. What do you think?

Hi @Districh-ru , sure, I can add. Somehow it doesn't work for my keyboard.

Districh-ru:
> Hi @Districh-ru , sure, I can add. Somehow it doesn't work for my keyboard.

Lets test it. It should work for English keyboard layout. Some other layouts, including Russian, interpret the right Alt key as **Alt GR**. Thanks to OS updates. :(
I hate such behavior and for many years use installer of special Russian keyboard layout with the normal right Alt from here (in Russian): https://tabun.everypony.ru/blog/computers/105151.html

IMHO we should add right Alt because it can be used on some layouts. I made a test on my system and right Alt works perfectly for me.

Districh-ru (approved):

PR #8016, opened 2023-11-03
Add treasure placement in the Editor

ihhub:
This pull request adds an ability to place Treasure type objects on the Adventure Map within the Editor.
image

The new and old item lists were rendering images and text incorrectly. This pull request fixes it. To verify the changes add these lines:
```cpp
        fheroes2::DrawRect( display, { destination.x + middleImageOffsetX - ( itemSprite.width() / 2 ),
                                       destination.y, itemSprite.width(), itemOffsetY * 2 }, 80 );
        fheroes2::DrawRect( display, { destination.x + textOffsetX, destination.y, background->activeArea().width - textOffsetX - 55, itemOffsetY * 2 }, 96 );
```
at line 159 in **dialog_selectitems.cpp** file (`void renderItem()` method) to observe how text and item rendering is done.

**Note**: object erasure is not handled in this pull request and requires a set of changes that should be done separately.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -153,8 +153,8 @@ public:
         fheroes2::Display & display = fheroes2::Display::instance();
 
         if ( !itemSprite.empty() ) {
-            fheroes2::Blit(
-                itemSprite, display, destination.x + middleImageOffsetX - ( itemSprite.width() / 2 ), destination.y + itemOffsetY - ( itemSprite.height() / 2 ) );
+            fheroes2::Blit( itemSprite, display, destination.x + middleImageOffsetX - ( itemSprite.width() / 2 ),
+                            destination.y + itemOffsetY - ( itemSprite.height() / 2 ) );
         }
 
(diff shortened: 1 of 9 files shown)


src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -263,13 +268,16 @@ class SelectEnumHeroes : public SelectEnum
...
+        renderItem( port, Heroes::GetName( index ), { dstx, dsty }, 45 / 2, 50, _offsetY / 2, current );
     }
 
     void ActionListPressRight( int & index ) override
     {
         Dialog::QuickInfo( *world.GetHeroes( index ) );
     }
+
+private:
+    const int32_t _offsetY{ 35 };
```
oleg-derevenetz, line 280: Since `_offsetY` is always 35 (and cannot be redefined via constructor) I suppose we can make it `static` and save a few bytes of heap or stack - here and in similar classes below.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion, please take a look when you have time.

Districh-ru (approved):

oleg-derevenetz (approved):

PR #8022, opened 2023-11-05
Town Portal dialog: use the same dialog code as for the select dialogs in Battle Only mode (dialog_selectitems).

### LINKED ISSUE #7906: Adventure map, town portal spell window extension
### Preliminary checks

image
The castle list is too small for such screen size.

Town portal window could be adjusted according resolution selected.
I suggest adding a few more lines into the Town Portal spell UI. The same as was implemented for save file list.
On the screen game is running on ...x600. The castle list could be extended for 2-3 lines here (but not making it too long).

### Additional info

_No response_

Districh-ru:
Fix #7906

Town Portal dialog will change its height according to the total towns in list, but maximum 12.

screenshot

This PR adds ability to add a text under the dialog title.


Branikolog (approved):

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -374,15 +389,49 @@ class SelectEnumSecSkill : public SelectEnum
...
+        renderItem( castleIcon, castle->GetName(), { dstx, dsty }, 35, 75, _offsetY / 2, current );
+    }
+
+    void ActionListPressRight( int & index ) override
+    {
+        Dialog::QuickInfoWithIndicationOnRadar( *world.getCastleEntrance( Maps::GetPoint( index ) ), background->activeArea() );
+    }
+
+private:
+    static const int32_t _offsetY{ 35 };
```
ihhub, line 422: Please make this variable **public** and use it at lines 610-611.

ihhub (changes requested):
Hi @Districh-ru , I left 2 comments here. Would you mind to take a look at them?

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Hi, @ihhub, it will not help because we make this list object only after the 610-611 lines using the values, calulated in these lines.
I'll redo the constructor of **SelectEnum** class to use `_offsetY` value and calculate the dialog height taking list items count into account. not to have such dialogs:
изображение

But maybe it'll be good to make this improvement in a separate PR, what do you think?

ihhub: This class is defined at lines 389-420. It is used at line 662. I am asking to make this `_offsetY` **static** variable to be public.

Districh-ru: Oh, yes, we can do this. :)

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for this change!

PR #8040, opened 2023-11-11
Shipyard: Treat water with an object's shadow as water free of objects

### LINKED ISSUE #8039: Seven Lakes, Red player cannot build a shipyard
### Preliminary checks

### Platform

Windows

### Describe the bug

Shipyard somehow is not available for red player. 
fheroes2:
image
image

OG:
image

### Save file

None.

### Additional info

_No response_

Districh-ru:
Fix #8039

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -2248,7 +2248,8 @@ int32_t Castle::getTileIndexToPlaceBoat() const
 
         // Mark the tile as worthy to a place a boat if the main addon does not exist on this tile.
         // This means that all objects on this tile are not primary objects (like shadows or some parts of objects).
-        return ( tile.getObjectIcnType() == MP2::OBJ_ICN_TYPE_UNKNOWN );
+        return ( tile.getObjectIcnType() == MP2::OBJ_ICN_TYPE_UNKNOWN
+                 || ( tile.GetObject() == MP2::OBJ_NONE && tile.getLayerType() == Maps::ObjectLayerType::SHADOW_LAYER ) );
     };
 
     const int32_t index = Maps::GetIndexFromAbsPoint( possibleSeaTile.x, possibleSeaTile.y );
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -2176,9 +2176,10 @@ bool Castle::HasSeaAccess() const
             return true;
         }
 
-        // If this is an action object and it can be removed then it is possible to put a boat here.
+        // If this is a shadow or an action object and it can be removed then it is possible to put a boat here.
         const MP2::MapObjectType objectType = tile.GetObject();
-        return MP2::isPickupObject( objectType ) || objectType == MP2::OBJ_BOAT;
+        return MP2::isPickupObject( objectType ) || objectType == MP2::OBJ_BOAT
+               || ( objectType == MP2::OBJ_NONE && tile.getLayerType() == Maps::ObjectLayerType::SHADOW_LAYER );
     };
 
     const int32_t index = Maps::GetIndexFromAbsPoint( possibleSeaTile.x, possibleSeaTile.y );
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -2178,8 +2178,7 @@ bool Castle::HasSeaAccess() const
 
         // If this is an object's shadow or this is an action object that can be removed then it is possible to put a boat here.
         const MP2::MapObjectType objectType = tile.GetObject();
-        return MP2::isPickupObject( objectType ) || objectType == MP2::OBJ_BOAT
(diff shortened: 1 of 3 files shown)


Branikolog (approved):
@Districh-ru 
Thanks for such a fast fix! I can now build a shipyard even in a previously generated game.

Branikolog:
@Districh-ru 
Oh, the only small issue here...
I'm not able to build a boat in this city.

Districh-ru:
> @Districh-ru 
> Oh, the only small issue here...
> I'm not able to build a boat in this city.

Sad. :)
I'll try to fix it tomorrow morning

Branikolog (approved):
Now shipyard is available and I can purchase a boat there. 👍

src/fheroes2/castle/castle.cpp
```diff
@@ -2176,9 +2176,10 @@ bool Castle::HasSeaAccess() const
             return true;
         }
 
-        // If this is an action object and it can be removed then it is possible to put a boat here.
+        // If this is an object's shadow or this is an action object that can be removed then it is possible to put a boat here.
         const MP2::MapObjectType objectType = tile.GetObject();
-        return MP2::isPickupObject( objectType ) || objectType == MP2::OBJ_BOAT;
+        return MP2::isPickupObject( objectType ) || objectType == MP2::OBJ_BOAT
+               || ( objectType == MP2::OBJ_NONE && tile.getLayerType() == Maps::ObjectLayerType::SHADOW_LAYER );
```
ihhub, line 2182: We are checking object layer only of main object on the tile, not the whole tile. I suggest to use `Maps::Tiles::isShadow()` that covers everything. However, this does no cover all possible cases. I think the logic should be like this:
```cpp
MP2::isPickupObject( objectType ) || // pickup objects can be removed later
objectType == MP2::OBJ_BOAT || // obviously, if boat exists then it can be removed later
!tile.isObjectOrBackgroundLayerType()
```
The last function does not exist but the idea is to check main tile addon and bottom layers addon that none of addons are on object and background layers.

By the way `isObjectOrBottomLayerType` is named incorrectly as it should be for OBJECT_LAYER and BACKGROUND_LAYER layers, not for bottom layer.

ihhub (changes requested):
Hi @Districh-ru , I left one suggestion here. Could you please take a look at it when you have time?

src/fheroes2/castle/castle.cpp
Districh-ru: Hi, @ihhub, sure.
The `isObjectOrBottomLayerType()` function name is fixed now and moved to the tile and addon methods.
I used the suggested `!tile.isObjectOrBackgroundLayerType()` logic.
Thanks!

src/fheroes2/maps/maps_tiles.h
```diff
@@ -174,6 +180,12 @@ namespace Maps
             return objectType == _mainObjectType;
         }
 
+        // Returns true if tile's main addon layer type is Object or Background and false if it is Terrain or Shadow.
+        bool isObjectOrBackgroundLayerType() const
+        {
+            return _mainAddon.isObjectOrBackgroundLayerType();
+        }
```
ihhub, line 187: This is not entirely correct for all cases. There could be a situation when **_addonBottomLayer** can have an object with this layer type but the main addon doesn't. We should run through all entries of **_addonBottomLayer** as well.

I also suggest to rename this function to avoid misleading reading. Including a word **tile** will solve the issue.

src/fheroes2/maps/maps_tiles.h
```diff
@@ -67,6 +67,12 @@ namespace Maps
 
         ~TilesAddon() = default;
 
+        // Returns true if addon's layer type is Object or Background and false if it is Terrain or Shadow.
+        bool isObjectOrBackgroundLayerType() const
```
ihhub, line 71: While addressing comments for Ocean Objects PR I was thinking how we properly name this function. How about we name it as `doesOccupyTile()`? Or even better `isPassabilityTransparent()` and use opposite logic as:
```cpp
bool isPassabilityTransparent() const
{
    return _layerType == SHADOW_LAYER || _layerType == TERRAIN_LAYER;
}
```

What do you think?

ihhub (changes requested):
Hi @Districh-ru , I added few more comments here. Could you please take a look at them?

ihhub:
This pull request should be merged for the upcoming release as it is a regression fix. Therefore, I am marking it as the high priority PR.

src/fheroes2/maps/maps_tiles.h
Districh-ru: Hi, @ihhub, yes we should also check this case. I changed the function.

src/fheroes2/maps/maps_tiles.h
Districh-ru: Yes, lets do so. I named the tile function the same way.
I'm not good in naming functions and variables. :(

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for these fixes!

Issue #7973, opened 2023-10-21
Combat, AI is moving too safe while opposing army with strong ranged troops


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

https://github.com/ihhub/fheroes2/assets/55348946/518ab599-033d-4a6e-b62e-db8f405493b7

Current behaviour could be easily abused by players.

AI shouldn't use such "safe" approach in battles against ranged troops, placing his melee troops too far away . I expect the current approach be used against melee only army (with minimum ranged potential (considering attacking spells too)).

### Describe the possible solution

In the example above Pikemen could reach halflings on the second round already, but they prefer to stop out of boars range, so none of them could cover strong stack of halflings on the second round. 
Continuing the third round the majority of Pikemen stacks made only a single step forward, that allows halflings to shoot one more time on the next round (there's only one stack of Pikemen rushing, which could be killed by halflings.

I'm also concerned about the threat AI is considering now. Just a few pairs of boars are not such a strong rivals. AI decided to stay away from their range, but more effective tactic in this case is rushing ranged troops. AI has more stacks and player will potentially try to hide halflings behind boars. So Pikemen using the maximum movement range will do more efficiently, if they ignore boars here and walk forward. 
Even if player's army has strong enough melee troops, AI placing his troops out of enemy range will give the player a few more turns for spell casting or shooting the defending AI.

Probably, if there were some other slow stacks (like ogres or golems) Pikemen could took the cells which are not reachable by slow stacks, but at the same close enough, to cover ranged troops the upcoming round.

[Shivery AI.zip](https://github.com/ihhub/fheroes2/files/13060420/Shivery.AI.zip)

Branikolog:
The similar problem with wolves is observed in https://github.com/ihhub/fheroes2/issues/7972#issue-1955404012 issue.

PR #8047, opened 2023-11-16
Improve the logic of AI defensive tactics

### LINKED ISSUE #8031: Combat, AI skips turns of several troops in a row
### Preliminary checks

I don't know why AI preferred to stay, while he could focus my Centaurs or come closer to fliers.

https://github.com/ihhub/fheroes2/assets/55348946/c4e61900-1321-41de-9c3c-292889167e4b

AI should proceed attacks.

[Skip attack.zip](https://github.com/ihhub/fheroes2/files/13308702/Skip.attack.zip)

### Additional info

_No response_

oleg-derevenetz:
close #8031

No rocket science, just a (hopefully) sane implementation of defensive tactics plus minor code deduplication.

`master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/6899c395-8e41-4596-b4d6-f7811e8468cf

Swordsmen do not actually cover the Rangers, they run out as soon as the Mummies come into range, allowing the Mummies to block the shooters and surround the infantry units, resulting in defeat.

This PR:

https://github.com/ihhub/fheroes2/assets/32623900/4963776a-361a-4ea0-8ad9-facfcd0d62cb

Here the AI acts in the style of a human player: Rangers are safe under the protection of Swordsmen who stand on the defensive and do not allow the Mummies to surround the infantry units and approach the shooters, resulting in a victory with moderate losses.

Behavior in the situation from the @Branikolog's issue:

https://github.com/ihhub/fheroes2/assets/32623900/17bfad44-fee3-4b42-a543-fd3f0374cfa5

There are no skipped turns, all units participate in the battle.


ihhub (approved):

Branikolog:
Hello, @oleg-derevenetz !
I've made a few tests here.
Here's what I've observed:

https://github.com/ihhub/fheroes2/assets/55348946/48ca0b4c-06c9-436f-b09e-d673a80a9d56

When placing griffins not really close to boars and mages AI decided to skip rocs. In this particular situation, I suppose, AI doesn't know how to block that free single cell near mages with 2-hex Roc. So why did he decide to skip?
But at the same time moved Golems and Mummies, which is good.

If I skip griffins - AI skips all of them: Rocs+Golems+Mummies.

https://github.com/ihhub/fheroes2/assets/55348946/eec9901d-947c-4a8c-8e09-b9cc03b54247

So I cannot say the PR fixes the issue entirely.

Speaking about your example from Battle mode.
I can see a few weak spots. (Probably, they are not related to your fix, but I just want to mention them.)

https://github.com/ihhub/fheroes2/assets/55348946/7d704e80-ea2b-433e-9766-f43ae6b7e088

Swordsmen prefer defending Archers, but there're only 3. I'm not sure it is useful defending so tiny stack.

Even If I set higher swordsmen quantity - they still behave the same. I expect them attacking Mummies instead of defending useless stack of archers. Note, that with such behaviour they give the first strike to mummies, which brings more losses.

https://github.com/ihhub/fheroes2/assets/55348946/cd0a2497-69a8-46b8-9867-4e24eeb62e81

I've also noticed that swordsmen don't attack mummies at all, while defending archers. I expect them fighting till death with nearest foes, while keeping the defensive position. Maybe they can skip attack, but only in a case they will be dead after mummies retaliate. Correct me, if I'm wrong, but currently AI cannot plan such things as not attacking to keep the line of defending troops unbroken. So when they are not attacking mummies standing right in front of them.

Branikolog:
Hi, @oleg-derevenetz 

>Griffins in your video are placed just on the border of two halves of the battlefield, so everything is determined by the nearest cell to the attacker. For Rocks it's the cell with index 26, which is "on the other side", while for Mummies it's the cell 38, which is "on our side".

I find it weird when melee 1-hex troops attack that griffins stack, but rocs preferred to skip. I think we should somehow make AI consistent in his decisions of attacking or not with all his troops.

>and on the other - so as not to incur unnecessary losses while fighting surrounded by enemy melee units on the enemy side of the battlefield.

In the situation above griffins are still remain flying troops, that can move over the whole battlefield and AI shouldn't be scared attacking them.
Once he blinded minotaurs he still didn't want to attack neither griffins nor centaurs and just skipped.
Even if there're no griffins AI still prefers skipping:

https://github.com/ihhub/fheroes2/assets/55348946/e988fbe1-d2d0-49ad-9c68-7a7c475acaf0

How do you think, should blinded melee troop be ignored in such situation and AI should behave as there're no melee troops on opponent's side?

>>If I skip griffins - AI skips all of them: Rocs+Golems+Mummies.

>And that's the right behavior.

From my point of view a player should move such slow troops as mummies and golems closer to the defended shooting stack. I know it's not really related to the current PR. How do you think, should I open a separate issue for such cases?

>>So I cannot say the PR fixes the issue entirely.

>I think the current behavior is more correct in the defensive tactics mode than just attacking enemy units headlong.

From my point of view AI could attack troops from opposite army in a case there're not much potential losses expected in melee combat on the opposite side of the battlefield. Or, as I mentioned above, he should make all his walking army a little more compact (gather all troops closer to the defended archers), so he can defend his archers more effectively. Maybe if there were fliers instead of mummies and golems, skipping could be the proper behaviour.
But anyway, AI attacking me, even if it's a little less optimal, looks much better, that just skipping for a unknown reason. If Minotaurs were not blinded and all bottom troops were fliers this behaviour would look proper and correct. Now it looks like something is wrong and rather large stack of rocs just doing nothing, while it can bring significant damage to griffins or centaurs.

>This is a question of when to choose an attacking tactic, and when to choose a defensive one. In general, this is out of scope of this PR, but for general understanding, currently defensive tactics is chosen when player's side has the superiority in archers, but it doesn't have an overwhelming superiority in general army strength ("overwhelming" is x6 - x10 depending on the composition of the army and the type of the current unit - offensive tactics is more readily chosen for flying units). "Tiny stack" is a rather vague concept. Need some numbers :) Anyway, it's out of scope of this PR.

It looks like I accidentally touched the edge values for this logic. :)
Unfortunately, since I don't know how exactly the code works, I cannot advise any numbers, except of saying: "Lower/higher."
I'm not sure anyone can propose the correct numbers and we just need to changing them with a little steps and testing it in different situations.
Should I create a separate issue for such defending of "tiny" archer stacks? :)

>It can do this, but this is not done deliberately for defensive tactics, so that the units protecting the archers will last as long as possible. With the current behavior, they will take damage just once per turn (considering the case when number of attackers is equal to the number of defending units), while with the more "agressive" behavior they will take damage at least twice per turn - normal damage from the attacker + retaliation damage. It is assumed that archers (and "free" units not covering the archers) will do all the job, because we have an advantage in them, and they will not be able to do their job if their guards are destroyed quickly. Even one or two additional turns with non-blocked archers can affect the outcome of the battle in this case, so the units covering the archers behave passively and do not get into trouble (not like Titus Pullo in the first scene of "Rome").

Ideally I expect AI checking the overall strength of archers in such situation: if they are quite weak, then defending rows of swordsmen shouldn't suffer standing without an action. But this is rather tricky logic to adjust.
How do you think about making such logic, so small swordsmen stack won't attack if they will be eliminated from the retaliation? I mean if swordsmen attack and die - then just skip, allowing archers 1 extra shot. If swordsmen stack can survive retaliation - they should attack nearby enemy without opening a path for attackers to cover the archers.

>if they will be eliminated from the retaliation? I mean if swordsmen attack and die - then just skip,
If you think it's just too risky, then what about making them skip their turn if they couldn't survive the potential retaliation AND the upcoming the attack of the troop? (Sure, there could be a situations, when swordsmen could be facing several stacks, but in this case they just die without dealing any damage suffering attacks from 2 and more enemy stacks.)
But if we have two and more stacks of swordsmen and only a single enemy stack, all swordsmen are going to be scared attacking, while 1 stack can attack, suffer retaliation, but the next stack will attack without retaliation. I believe than in average situation this brings more profit, than all 3 stacks of swordsmen would skip until enemy would break 1 and get the archers covered.

>while with the more "agressive" behavior they will take damage at least twice per turn - normal damage from the attacker + retaliation damage.

Don't forget that swordsmen deal some damage reducing the attacking stack, which makes future enemy attacks weaker. (In the example from the video above swordsmen even have an advantage in speed, which allows them to attack first > eliminate some mummies before they retaliate and attack back). 
Ideally, if we can create a formula, that considers archers strength (the speed they eliminate enemy stacks) +swordsmen strength (the damage swordsmen can deal on their own), and the potential losses of defending swordsmen it would be the perfect defending tactics, but for now I prefer AI to behave more aggressive, rather than skipping, as it looks like a bug in logic.

oleg-derevenetz:
Hi @Branikolog

> I think we should somehow make AI consistent in his decisions of attacking or not with all his troops.

It is consistent. There is a clear logic: if the point from which it is advantageous to attack (the closest of those that gives the maximum potential damage) is on our side of the battlefield, then the unit will attack, otherwise it will not.

> In the situation above griffins are still remain flying troops, that can move over the whole battlefield and AI shouldn't be scared attacking them.

If we have an advantage in shooting units, there is no point in putting flying units in a risky position, which will be outnumbered on the enemy side of the battlefield and will suffer unnecessary losses.

> How do you think, should blinded melee troop be ignored in such situation and should AI behave as there're no melee troops on opponent's side?

I would say that this is an unnecessary risk, because the opponent may have Cure or Dispel Magic spells. He can heal the Minotaurs and they will attack our Rocks on the next turn, as they have an advantage in speed.

> From my point of view a player should move such slow troops as mummies and golems closer to the defended shooting stack.

Currently they are moving closer to the shooters only if there is a real way to cover them. But in your video there is no way to do it because of obstacles. In theory, of course, they can try to snuggle up to other units covering the shooter, but in this case you can say that this means the risk of using mass damage spell by the attacker, therefore, this behavior does not apply. Free units will only attack the enemy as soon as he appears in their half of the battlefield.

> From my point of view AI could attack troops from opposite army in a case there're not much potential losses expected in melee combat on the opposite side of the battlefield.

Currently it will do it - if it has x6 - x10 advantage in the overall army power (regardless of the advantage in archers) AI will switch to the attacking tactics. Maybe these numbers can be reduced, but this PR does not relate to the choice of tactics, but to actions with the chosen tactics.

> Now it looks like something is wrong and rather large stack of rocs just doing nothing, while it can bring significant damage to griffins or centaurs.

... and then get ambushed by the Minotaurs on the next turn if enemy has dispelling magic. I personally don't care much that "something may look wrong/like a bug" from the player's point of view as long as the behavior is backed up by clear logic.

> Unfortunately, since I don't know how exactly the code works, I cannot advise any numbers, except of saying: "Lower/higher."
I'm not sure anyone can propose the correct numbers and we just need to changing them with a little steps and testing it in different situations.
Should I create a separate issue for such defending of "tiny" archer stacks? :)

This refers to the choice of tactics that is out of scope of this PR. We can set it up later if there are any specific ideas on this.

> Ideally I expect AI checking the overall strength of archers in such situation: if they are quite weak, then defending rows of swordsmen shouldn't suffer standing without an action. But this is rather tricky logic to adjust.

Again, it's a matter of choosing tactics.

> How do you think about making such logic, so small swordsmen stack won't attack if they will be eliminated from the retaliation?

This means looking just one step ahead. But in general if the goal is to hold out for as many turns as possible then it is more profitable NOT to attack. Consider the following case:

https://github.com/ihhub/fheroes2/assets/32623900/20ac0d71-375f-4a6d-83f4-2321f7e16598

Dwarves lasted 3 turns, attacking the Minotaurs uselessly, and did not kill a single one. With the "non-attacking" tactics:

https://github.com/ihhub/fheroes2/assets/32623900/d87dd40f-4390-4108-8cbd-992ec0c3f499

Dwarves lasted 6(!) turns. And every extra turn is an extra shot of archers covered by dwarves.

> but for now I prefer AI to behave more aggressive, rather than skipping, as it looks like a bug in logic.

Again, don't worry too much about "how it looks like". Most players have no idea about the game mechanics, how their actions may turn out and what kind of behavior may be more profitable in the general case, taking into account current priorities (priorities "kill more enemies" and "hold out more turns" generally require different behavior). You need to do what is right in the general case, taking into account the chosen tactics, and not what the player "expects".

We can add the behavior "while covering shooters, attack only those units that are guaranteed to be unable to respond," but this improvement can be made a separate PR, because this also needs to be thought through - for example, should you attack a paralyzed unit or not, and if so, in what cases? The thing is that from the "hold out more turns" point of view it is more profitable NOT to attack such a unit, and archers will not to attack it as well (unless there are no other targets).

Branikolog:
@oleg-derevenetz 

>Again, don't worry too much about "how it looks like". Most players have no idea about the game mechanics, how their actions may turn out and what kind of behavior may be more profitable in the general case, taking into account current priorities (priorities "kill more enemies" and "hold out more turns" generally require different behavior). We need to do what is right in the general case, taking into account the chosen tactics, and not what the player "expects". From the point of view of the average player, the natural behavior of AI is to stupidly run at the nearest enemy unit, because it's what most AI implementations do :)

I agree, that average implementations are quite weak. :) We're both here to make fheroes2 implementation outstanding. ;)
But at the same time I expect AI behave in such manner, so I (as an experienced player) always know, what is happening and why. Currently, I find it really weird, when Rocs and other troops stand and do nothing, while they can at least come closer, but this issue for sure, is out of the scope of this PR. (But not out the https://github.com/ihhub/fheroes2/issues/8031 ;) )

>We can add the behavior "while covering shooters, attack only those units that are guaranteed to be unable to respond," but this improvement can be made a separate PR, because this also needs to be thought through - for example, should you attack a paralyzed unit or not, and if yes, in what cases? The thing is that from the "hold out more turns" point of view it is more profitable NOT to attack such a unit, and archers will not to attack it as well (unless there are no other targets).

Not responded attacks... Do you mean hydra's non-retaliatiative ability? Or just the troops, that were already retaliated this turn? (This is not going to happen too often, since to cover the rangers you need almost all stacks from your army and they are standing still without attacking.)

Sure, I don't want to force you making too huge PR so it would be impossible to test properly. Let's make improvements step by step.
But we definitely need a defined rule for choosing the most profitable behaviour (attacking or holding more turns) in this case.

>>How do you think about making such logic, so small swordsmen stack won't attack if they will be eliminated from the retaliation?

>This means looking just one step ahead. But in general if the goal is to hold out for as many turns as possible then it is more profitable NOT to attack. Consider the following case:


>Again, it's a matter of choosing tactics.


Indeed, but we need a clear logic to define which case is more proper for living as much as possible, and which brings more benefits when attacking.

Right now I see this logic as not optimal:

PR:

https://github.com/ihhub/fheroes2/assets/55348946/2aeca19d-8deb-4da8-b4ba-c150933b2c43

AI lost 9 swordsmen holding their attacks.

Master:

https://github.com/ihhub/fheroes2/assets/55348946/991dce91-3f7c-4f92-afc4-188a394a2318

AI lost just 5 swordsmen. I believe I could personally reduce my losses even more targeting the most vulnerable mummies.

It goes without saying that covering 5 archers is not really worth it, considering how many swordsmen we have in army.

But I went even further and here's the next example:

https://github.com/ihhub/fheroes2/assets/55348946/990b2e2b-ff01-4b5c-a086-bfcc5f7561f8

Doubled swordsmen stacks covering those 5 archers look ridiculous skipping attacks.

Here's the outcome from the master:

image


>>I think we should somehow make AI consistent in his decisions of attacking or not with all his troops.

>It is consistent. There is a clear logic: if the point from which it is advantageous to attack (the closest of those that gives the maximum potential damage) is on our side of the battlefield, then the unit will attack, otherwise it will not.

I apologize if I understood you incorrectly. Griffins were placed on the edge of Rocs "safe" zone. But in the edge of the safe zone of Golems and Mummies, right? 

image

Is this why Golems and Mummies move towards griffins?

Attacking griffins

I assume they are willing to attack griffins. But Rocs are not, which is weird to me, as Rocs potentially appear on the same place as golems will and are going to be potentially attacked by minotaurs (potentially cured) (I'll come back later to this issue of the chances and risks with blind spell)

oleg-derevenetz:
Hi @Branikolog

> Currently, I find it really weird, when Rocs and other troops stand and do nothing, while they can at least come closer

This needs to be formalized. "Come closer" is how much "closer"? Just stand all in a pile closely? Then some will interfere with others because of the need to bypass others, and this behavior increases the likelihood of the enemy using mass damage spells.

> Do you mean hydra's non-retaliatiative ability? Or just the troops, that were already retaliated this turn? (This is not going to happen too often, since to cover the rangers you need almost all stacks from your army and they are standing still without attacking.)

Both. But, as you said, this is not going to happen too often, so I decided not to implement this.

> Doubled swordsmen stacks covering those 5 archers look ridiculous skipping attacks.

I agree that AI should not use defensive tactics in cases like this. There should be some minimum threshold in archer's superiority, not just "I have 3 archers and you have none". This needs to be thought out separately.

> Griffins were placed on the edge of Rocs "safe" zone. But in the edge of the safe zone of Golems and Mummies, right?

Not quite. Since battlefield has the hexagonal grid, grid cells on nearby rows are placed with offset. For Rocks, the closest cell near the Griffins is cell 26 (upper right one near the Griffins - on the same row where Rocks are placed) which is considered "enemy side", while for Mummies it's cell 38, to the right of the Griffins, which is "our side", since it has offset to the right.

oleg-derevenetz:
> I apologize if I understood you incorrectly. Griffins were placed on the edge of Rocs "safe" zone. But in the area of the safe zone of Golems and Mummies, right?

OK, I streamlined the logic a bit. Now an enemy unit will be attacked if there is any way to attack it from a defended area, even if it means attacking not from the cell that is closest to the attacker:

https://github.com/ihhub/fheroes2/assets/32623900/8e49df7a-3682-4f56-a24b-35b29bc61228

Defended area is the player's half of the battlefield (max 5 cells from the corresponding edge).

> I'll come back later to this issue of the chances and risks with blind spell

I suggest not wasting time discussing particular cases, but working out a general logic, if you believe that the current logic does not suit. Perhaps this should be done after re-evaluating the mechanism for choosing tactics - that is, when to choose defensive tactics, and when to choose offensive tactics, because choosing defensive tactics with a small number of archers is indeed not always justified. This can be done in a separate PR if you have any specific ideas. Perhaps after that, some of the questions about the current logic of the defensive tactics will disappear altogether, since this tactics just will not be applied in situations that are now considered controversial.

Please keep in mind that computers don't understand general ideas. Specific algorithms are needed, and preferences are given to simpler ones if they give results that are "good enough" in most cases, because the more complex the algorithm, the more difficult it is to implement, test and then maintain it (especially for people who came to the project recently and want to figure it out).

Branikolog:
Hi, @oleg-derevenetz 

>This needs to be formalized. "Come closer" is how much "closer"? Just stand all in a pile closely? Then some will interfere with others because some of them will need to perform extra steps to bypass others (and some of them may be blocked completely by surrounding friendly units), and this behavior increases the likelihood of the enemy using mass damage spells.

I personally would place my melee walking troops in area, where I can attack a potentially placed enemy troops. On the screenshots above it's the hexes right in front of boars and rocs: Golems should be able to attack enemies standing on those hexes. At the same time the most distant hex from rocs and boars should be preferably chosen to not overcrowd the area and be more safe in a case of mass attacking spell.

>>Griffins were placed on the edge of Rocs "safe" zone. But in the edge of the safe zone of Golems and Mummies, right?

>Not quite. Since battlefield has the hexagonal grid, grid cells on nearby rows are placed with offset. For Rocks, the closest cell near the Griffins is cell 26 (upper right one near the Griffins - on the same row where Rocks are placed) which is considered "enemy side", while for Mummies it's cell 38, to the right of the Griffins, which is "our side", since it has offset to the right.

This is the issue that bothers me: If AI wants to attack with mummies and golems, he should want to attack with rocs or any other 2-hex creature.

>OK, I streamlined the logic a bit. Now an enemy unit will be attacked if there is any way to attack it from a defended area, even if it means attacking not from the cell that is closest to the attacker:
>Defended area is the player's half of the battlefield (max 5 cells from the corresponding edge). After the Rocs moved to attack the Griffins, the passage to the Magi opened, and the Mummies with Golems began to approach them to protect them from all sides.

Now it looks much better. I'll make a few tests later to ensure everything  else goes well.

>Attempts to develop general behavior based on a particular case will not lead to anything good. Although the logic that with the chosen defensive tactics, units should not leave their half of the battlefield was made before me, I fully agree with it and understand why this was done. This is done in order to avoid unnecessary losses in the general case, because in the general case switching to this tactic means that the enemy has more melee units while you have more shooters. In the general case, trying to attack someone in the enemy half of the battlefield, you put your units under attack of enemy infantry, where they will be quickly outnumbered and surrounded (once again, I remind you that, logically, such tactics should be chosen if you have more shooters, but less melee units than the enemy, so sending melee units, which you already have less than the enemy, directly into a bunch of enemy infantry is not very smart). Please note once again that I am talking about actions with the chosen tactics, and not about the choice of tactics. The right choice of tactics is a separate, although of course related, problem, and there is a field for additional tuning.

I completely agree with you, @oleg-derevenetz . Firstly we should create a reasonable set of tactical patterns for general cases.
I was not aware, that current implementation choses one tactics and makes all actions within it according to the predefined logic. 
So we need to think later about adding more flexibility into the current tactics or teach AI change his choice more frequently according to the battle circumstances (or widen the choice of tactics, adding something like "defensive, but with the chance of doing +/- safe attacks on the other side of the battlefield" tactics).

The case with minotaurs seemed quite obvious for me, but I can be aware of not the whole logic in the code, to give you advices. I just propose some ideas, showing different situations, so we can come up with elegant and simple solution, which works in 90% of cases together.
Once minotaurs are blinded I personally expect them temporary off in 95% of cases. For instance: I definitely prefer coming closer and attack shooters not far away from them. In general case I don't expect enemy to dispel this effect and I personally rush the opposite side of the battlefield. Why do I blinded them previously? Just to look at them standing? :) 
If I know, that enemy has dispel (used previously in this or even other battles) then I for sure won't rush those archers (even considering the fact, that AI could use other spell). But there're lots of other thing to count on, such as the attacking order of troops. But anyway, this is really far out of the current PR...

>This can be done in a separate PR if you have any specific ideas. Perhaps after that, some of the questions about the current logic of the defensive tactics will disappear altogether, since this tactics just will not be applied in situations that are now considered controversial.

The only two things that bother me here is that case with those large stacks of swordsmen standing without action and the case, when they defend too weak stack of archers.
I'll try to test your PR tomorrow.

I want to thank you for your patience explaining things to me and for that hard work you're doing on the project and AI especially.

oleg-derevenetz:
@Branikolog

> I was not aware, that current implementation choses one tactics and makes all actions within it according to the predefined logic.

To be more precise, each unit chooses a certain tactic, not the army as a whole. That is, for example, flying units are currently more willing to switch from the defensive tactics (when their side has the simultaneous superiority in archers) to the attack tactics - for them, a x6 superiority in the power of the army over the enemy is enough, while infantry units are switching to the attack tactics only with a x10 superiority.

> Why do I blinded them previously? Just to look at them standing? :)

In the case of defensive tactics - to be able to focus the archers' fire on the remaining units, deal with them faster and then focus the fire on the Minotaurs. It's like Slow, but better. AI-controlled archers are aware of this BTW and will not attack blinded stacks while there are another enemy stacks on the battlefield.

> The only two things that bother me here is that case with those large stacks of swordsmen standing without action and the case, when they defend too weak stack of archers.

This should be tuned in terms of choosing tactics. It is necessary to take into account not just "superiority in archers", but also other factors that have yet to be considered. Perhaps some minimal superiority in archers is needed. Perhaps it is necessary to take into account the comparative proportion of archers in the army, for example, if it is small (how small exactly?) then AI should not go on the defensive.

> I want to thank you for your patience explaining things to me and for that hard work you're doing on the project and AI especially.

You are welcome :)

Branikolog:
@oleg-derevenetz 

>In the case of defensive tactics - to be able to focus the archers' fire on the remaining units, deal with them faster and then focus the fire on the Minotaurs. It's like Slow, but better. AI-controlled archers are aware of this BTW and will not attack blinded stacks while there are another enemy stacks on the battlefield.

Indeed.

> It is necessary to take into account not just "superiority in archers", but also other factors that have yet to be considered. Perhaps some minimal superiority in archers is needed. Perhaps it is necessary to take into account the comparative proportion of archers in the army, for example, if it is small (how small exactly?) then AI should not go on the defensive.

Maybe additionally to rangers superiority (over the enemy archers) we should compare the potential damage melee troops and shooters can deal? Like if archers cannot do more than ~20% of all melee troops damage - then no need to defend them at all. If shooters make >~50% damage of potential damage of melee troops, then "standing out" tactics should be chosen? for the cases when defending troops make ~20-50% of rangers' damage - then they can defend rangers and attack nearby enemies at the same time(holding their position).
I'm suggesting to consider not only archers' strength while comparing to enemy archers, but also consider the proportion of archers damage to melee defending troops (swordsmen). Also the potential damage defenders could take and inflict should also be considered while choosing the "standing out" tactics.

If defending swordsmen can hit attacking troops and demolish >~30-40% of enemy stack (no matter how rangers currently success in damage dealing)  - than "standing out" tactics should be canceled too and defenders should attack the nearest weak enemies but without moving. This would help to finish attacking troops, that were heavily hit by rangers. So rangers could choose another target.

oleg-derevenetz:
Hi @Branikolog

> Maybe additionally to rangers superiority (over the enemy archers) we should compare the potential damage melee troops and shooters can deal? Like if archers cannot do more than ~20% of all melee troops damage - then no need to defend them at all.

All calculations regarding "superiority" are already made in terms of "strength" which is calculated in a complex way, but it mostly depends on damage, so, for simplicity, we can assume that this is a damage.

OK, I added such a condition as part of this PR that in order for a unit to choose defensive tactics, the strength of archers must be at least 20% of the total strength of the army. Here is the general idea (with comments, but the logic is mostly self-explanatory) currently:

https://github.com/oleg-derevenetz/fheroes2/blob/9b8f026fc7408a7717cab5c936fd3fdeefbc456d/src/fheroes2/ai/normal/ai_normal_battle.cpp#L836-L874

Most of this logic already was here before. But I suggest doing an additional tuning (for example, attacks from the spot by units covering the archers) in the future in a separate PR.

Branikolog:
Hi, @oleg-derevenetz !

>Most of this logic already was here before. But I suggest doing all potential additional tuning (for example, regarding attacks from the spot by units covering the archers) in the future in a separate PR.

Sure. Be free to postpone any of the additional logic for the later PRs for the sake of simpler merging of the current PR.

If you are going to add attacking from the spot later, I think, 20% is a little too much. Maybe then we reduce the "magic number" to 15%, how do you think? Currently 15-16 archers do a decent damage but AI tends to run into the fight with swordsmen. Actually, this is not really bad decision, since mummies are not capable of reaching archers then, but in more general case, I think, Swordsmen covering even a bit weaker stack could bring nice battle results (if they can defend and attack at the same time).

Here's one video I shot with the PR's build.

https://github.com/ihhub/fheroes2/assets/55348946/4146a14b-f272-4f1c-af72-43740ffd3088

The fact that swordsmen tend to switch from defending to attacking tactics during the battle multiple times bothers me much. If I'm not mistaken, I met such problem on a master build too. As you can see on the video they just left mummies running back, but still remain under attack and suffer losses. Sometimes they decide to switch back and attack, so it looks weird during the battle.
On this video I managed to catch the case, where 1 stack of swordsmen decided to stay and fight till death. :D This sacrifice was  really courageous, but quite reckless. :) (By the way, why did that stack exactly stay with mummies?)
Could you, please, look if we can fix it with some minor changes here?

If rangers are not under the potential threat on the battlefield there's no need to switch tactics and defend them with swordsmen (especially for those swordsmen, that are too far away and when there're no ranged troops on the opponent side, unless they are succeeded in retreating without any losses).
In the video above swordsmen are too far away: most of them are on the opposite side of the battlefield. Considering the fact, that Mummies are still able to attack swordsmen, while they are coming back and shooters are not under attack - there's no need to retreat. If swordsmen are able to run away - then they can retreat to give shooters a few attacks. Such behaviour could take place, if all swordsmen stacks are able to cover and defend archers in a single movement.

The second thing that bothers me within current PR is the fact, that AI decided to protect Rangers, but somehow there're not enough stacks to do it. 1 brave stack of swordsmen left 1vs5 and was defeated. So there is a hole in protection wall in front of archers, where mummies managed to sneak and prevent them from shooting.

Could you, please, look, what can be fixed within the current PR and what we're going to leave for the further tasks?

oleg-derevenetz:
@Branikolog

> On this video I managed to catch the case, where 1 stack of swordsmen decided to stay and fight till death.

If a defensive tactic is chosen, but the stack is in the enemy half of the battlefield (or cannot reach its archers in 2 turns), then it will continue to fight.

> Could you, please, look if we can fix it with some minor changes here?

I don't see why we need to "fix" this.

> The fact that swordsmen tend to switch from defending to attacking tactics during the battle multiple times bothers me much.

If the swordsmen are in the minority, then it's better to let them defend archers than than to let the mummies break through to the archers. Swordsmen do everything right. I don't see a problem here.

> The second thing that bothers me within current PR is the fact, that AI decided to protect Rangers, but somehow there're not enough stacks to do it. 1 brave stack of swordsmen left 1vs5 and was defeated. So there is a hole in protection wall in front of archers, where mummies managed to sneak and prevent them from shooting.

Umm, so what? In this case defending units will stay close to friendly archers and attack enemy units blocking these archers. What is wrong with this behavior and what exactly are you offering instead?

oleg-derevenetz:
Hi @ihhub @Branikolog any decision regarding this pull request? I want to try to do something regarding #8070 (AKA #7973), but these changes will probably depend on changes made in this PR, and I would prefer to avoid duplicated changes and conflicts.

ihhub:
Hi @oleg-derevenetz , I see that these changes makes the AI decision better. We cannot to fix everything in one shot but we can iteratively improve the **overall** behaviour.

@Branikolog , the same question to you: does this PR make the AI logic better in general without going to corner cases? If yes then I suggest to move forward as we cannot endlessly discuss a perfect way which cannot be achieved with one bunch of changes.

Branikolog:
Hi, @ihhub !
I need 1-2 days more to thoroughly test the PR as I'm worried that it might lead to worse results in battles. Could you, please, hold merging for a little?

oleg-derevenetz:
Hi @Branikolog

> I need 1-2 days more to thoroughly test the PR as I'm worried that it might lead to worse results in battles.

When testing battles, please do not limit yourself to AI vs AI battles only. These battles are not indicative: the current offensive AI algorithms are such that it is easy to distract it from archers by individual units. It is necessary to test human vs AI battles, because defensive tactics from `master` practically does not protect the archers and allows the human player to break through to them and neutralize them too easily. In this regard, defensive tactics in the `master`branch is poorly implemented.

ihhub:
Hi @Branikolog is busy with some personal issues so his feedback is delayed at the moment.

ihhub:
Hi @Branikolog , please provide your feedback about this pull request.

oleg-derevenetz:
Hi @ihhub the state of this PR is still uncertain, I believe, and it will not go to 1.0.11?

ihhub:
> Hi @ihhub the state of this PR is still uncertain, I believe, and it will not go to 1.0.11?

I will chase @Branikolog in person within (my) tomorrow and come back to you. I would like to put it to 1.0.11.

Branikolog:
Hi, @oleg-derevenetz and @ihhub 

First of all I want to apologize for such a late response to this PR.
For me it's really important to be sure that battles controlled by AI worse, as this would lower the overall AI efficiency in the match.
So I spent lots of time analysing current behaviour.

Testing the PR with enhanced battle mode was really a pleasure, by the way. :)


>If a defensive tactic is chosen, but the stack cannot reach either friendly archers or any of the enemy units blocking them within the next 2 turns (and in your video these Swordsmen cannot reach their archers in 2 turns, they are missing one cell), BUT there is an enemy unit(s) within reach, then it will continue to use the attacking tactics (i.e. it will not try to retreat to cover the archers, because it's probably going to be late, but it will attack most dangerous reachable enemy stacks instead).

This logic is quite straightforward, but sometimes results of the battle really suffer from such approach:

image

image

image

While the common outcome is:
image

(in a case when all swordsmen defend archers)

I'm afraid the more variative scenarios we want AI to solve, the more we have to add into the existing AI logic.
I personally cannot describe whole effective behaviour in just a few logic blocks. 😞 There're so many things to consider so I think it's not possible to implement the whole logic of defending shooters in just 1 or 2 logic statemens.
But here're a few thoughts on the existing logic:

Defensive logic shines only in a case army has multiple melee stacks and all troops are close to the defended shooter. 👍  
But in a case 1 troop is too far away - he stays and suffers much from all attacks focused on him. ➖ 

https://github.com/ihhub/fheroes2/assets/55348946/f53d45fb-2a50-4dcc-8168-002c7676bd61

In a case the defending army has unsufficient stacks, they start moving backwards allowing enemy to attack. ➖ 

https://github.com/ihhub/fheroes2/assets/55348946/b6b7c595-5941-4986-821c-4d7f8bad34f3

Trying to defend shooters with insufficient troop stack number makes no sense. It even makes worse, since followed enemy troops can reach shooters. Is it possible to implement some sort of counter, so AI would know how many troops are needed and are ready to cover shooters to initiate defensive tactics? I'm not asking to do it urgently, but with this issue AI would loose more shooters during battles, which is not good.

>I don't see how and why we need to "fix" this. What exactly do you propose, with numbers? Increase the limit of turns from 2 to...? If you set the limit too high, then slow units, like Hydra or Iron Golems, will just crawl uselessly on the battlefield, trying to cover the archers, instead of attacking enemy units near them, and as a result, neither archers will be covered, nor damage to the enemy will be inflicted.

You're absolutely right. crawling back with a slow troop makes no sense.

Maybe defensive tactics should be initiated in the current way only if shooters could be covered within just 1 turn or all troops are faster than enemy, how do you think? Currently some troops are left alone during defending manuver and are surrounded by enemies (if they are too far away) which brings higher losses. I personally expect only faster troops to run back in a case troops are located not close enough (1+ turn to reach shooters).

I believe that if AI could consider troops' speed deciding to run back only if his troops are faster, than enemy, would reduce losses in average sutuation. If it's possible to implement for sure.
In a case there's at least one strong enough enemy troop that is faster than defending army, he can reach the shooter and spoil the whole maneuver. I believe turning to defensive tactics in the middle of the battle and rushing shooters while facing same or higher speed enemies is not a proper choice. (It's obvious on my video with golems) Ideally the mutual location of enemy and defending troops should be considered too, so defending troops could reach shooters faster, than enemies.

Uncontrolled turning into defensive tactics just allows enemy to attack defending troops on their way back to shooters, focus those 1-2 stacks, that decided not to protect shooters and also attack shooters in this case, as there'is not enough stacks to cover shooters.

So I propose to shorten the distance all troops can step back and defend shooters. I mean allow defensive tactics only if troops can reach shooters in 1 turn. This would bring benefit in 99% of cases. On longer distances the battle outcomes become not stable. For instance, if one stack gets one additional move due to morale triggered, it will die with the current logic. :)
Or we can think (now or in the future) about more complex logic. For instance: for fast-ultrafast troops allow 3 turns to come back, 2 turns for very slow-average speed stacks.  Or set distance individually for each speed.
Or allow defensive behaviour for those troops, that are located on a certain part of the battlefield (as an idea).

One more small remark:

I think turning into attacking state could be done when the army threat is not so overwhelming:

https://github.com/ihhub/fheroes2/assets/55348946/7db92e08-80fd-4e55-a0cd-df1d182c1f51

Opposite army is significantly stronger, I have just a few mummies but AI continues defending archers. I don't know the exact value, but I suggest rising it for a few %. But it's just my personal feeling. How do you think, @oleg-derevenetz ? Anyway, this could be done later.

>When testing battles, please do not limit yourself to AI vs AI battles only. 

Sure, but improving logic in AI vs AI battles would make the enemy heroes more effective on the adventure map. They would get less losses in simple battles against neutral monsters allowing to perform better against human and other players without wasting too much troops on early stages of the game. So this PR is supposed to make AI controlled heroes perform better agains neutral monsters -> make AI a better rival for other players on the map. :)


A couple of curious cases I found while testing your PR. Not sure what's going on here, but everthing goes according to the proposed logic - I apologise for bothering.

https://github.com/ihhub/fheroes2/assets/55348946/32f333da-92be-403d-a3cc-e8894b75a125

2 of 3 swordsmen decided to run back (to protect rangers probably from flier). I find it not really effective, since rangers don't have much chances to survive here, especially considering the fact, that swordsmen need 2 turns to reach them. But why only two stacks of swordsmen decided to run back? I suppose one of them later decided to proceed attacking, because rangers became too weak after the losses. And the second stack decided to attack flier just because Gargoyles were in attacking range and swordsmen considered them as a proper target. Correct me, if I'm wrong. 
Anyway, the whole situation with running back looks kind of strange.


https://github.com/ihhub/fheroes2/assets/55348946/06b9afb5-ba0a-4d30-9f24-854740af3c3b

There is no space around shooters for 1 stack of swordsmen. So they just stayed still until minotaurs came closer. But minotaurs were definitely in attack radius of swordsmen earlier. So swordsmen just skipped 1 turn, while they could attack. I expect them to continu attacking or coming closer to the rest of defending stacks.

oleg-derevenetz:
> This logic is quite straightforward, but sometimes results of the battle really suffer from such approach:

In your video, Swordsmen are moving forward to attack the Mummies because of that 15% limit in strength (proposed by you by the way :)). Archers' strength in your setup is 163.8 and total army strength is 1109, so archers' strength is 14.77% of total army strength which is less than 15%. If we add just one archer to the group of 12 archers (and turn it to the group of 13 archers) we will see the results like the following:

result

because in this case Swordsmen initially move to defend archers instead of trying to attack during the first few turns, and I cannot break to the archers even when controlling the Mummies manually because archers are properly defended. As you can see, there is another corner case appeared due to this 15% archer's strength threshold.

> Trying to defend shooters with insufficient troop stack number makes no sense. It even makes worse, since followed enemy troops can reach shooters.

They will always try to reach shooters - especially if they are properly controlled by the human player (AI-controlled units are too stupid for this at the moment, so it is not indicative). And it always makes sense to protect the shooters by placing the units protecting them close to the shooters, because at least this does not allow the enemy to completely surround the shooters, and slows down the elimination of the shooters. Going on the attack in this sense may help against a (currently) blunt enemy AI, that may be relatively easily distracted from shooters, but the human player will not be tricked by this, he will just bypass the nearby attacking units and surround the shooters.

> Currently some troops are left alone during defending manuver and are surrounded by enemies (if they are too far away) which brings higher losses.

In your video these units just should not go on the offense just so that after the loss of a pair of swordsmen, the strength of the archers would become more than 15% of the total strength of the army and they would have been forced to go into defensive mode. This is just another example of typical corner case. As I demonstrated above, if we add just one archer to our army, then Swordsmen just act correctly from the very beginning (by covering the archers) and suffer minimal losses. In the `master` branch, there is no that 15% threshold and Swordsmen will go on the defensive from the very beginning, even with 12 archers.

> I believe turning to defensive tactics in the middle of the battle and rushing shooters while facing same or higher speed enemies is not a proper choice.

OK, let's stick to the following logic for now: if melee unit is already in the enemy half of the battlefield, it will just never choose the defensive tactics and will never retreat to defend the archers. Such logic was already present, it was just that the protection of archers was a higher priority, now it is not the case.

> But why only two stacks of swordsmen decided to run back?

That's because during the turn of the first Swordsmen stack, one Swordsmen died from the retaliation attack. Relative strength of Archers is now above the 15% of total strength, so defending Archers is now a priority, so other two stacks are moved to protect them.

> There is no space around shooters for 1 stack of swordsmen. So they just stayed still until minotaurs came closer. But minotaurs were definitely in attack radius of swordsmen earlier. So swordsmen just skipped 1 turn, while they could attack. I expect them to continu attacking or coming closer to the rest of defending stacks.

When defensive tactics is chosen, units will attack only targets on their half of the battlefield. There is nothing new in this regard in comparison with the `master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/be2074bb-612c-4aed-89ea-c1133f5b6ed8

In this video from the `master` branch 27 Swordsmen are standing exactly in the middle of the battlefield, but do not attack the Minotaurs because they are on the "enemy half" of the battlefield.

oleg-derevenetz:
@Branikolog

> The problem here is not in the % :)

The problem here is that offensive tactics were initially chosen, when in fact defensive tactics should have been chosen initially. And the defensive tactics were chosen because the strength ratio between the archers and the entire army as a whole was slightly below 15%. When another coefficient of such kind is introduced, there is always a risk of some boundary value that will cause the logic to be unstable, like in this case: we lose just a single swordsman and the logic changes.

> The problem here is not in choosing right % of shooters strength to initiate this tactics, but in the fact, that AI tend to leave 1 stack sacrificing against the whole enemy army which is not optimal in most cases. I cannot tell it's a corner case. You can set up the same battle conditions (with leadership preferably) and just try to make 10 instant battles, where 2 would be with much higher losses.

Again, what exactly do you suggest? Whatever the limit on the number of turns for which it is necessary to reach the archers to cover them, there will always be someone who will not keep up with it - for example, due to the morale event. In the previous comment, for instance, you suggested limiting the number of turns to reach archers to one. You do understand that in this case there will be those who will have time to retreat and those who will not have time, right? And the same thing will happen in the case of two turns limit, three turns limit, and so on. A restriction that enables defensive tactics only for units in our half of the battlefield also does not solve this problem 100%, because it may happen that some units will be in our half of the battlefield (and they will retreat to cover the archers), and some will be in the enemy's half (and they will not retreat).

> Made some tests. Well... Sometimes the results still not really good...

They can't _always_ be good (see above).

> So, troops are not turning into defensive if they are on the enemy side, am I correct?

Yes.

> If I eliminate a significant part of swordsmen, they are still not defending nearby shooters.

In your video defending hero has attack spells which counts as "shooter strength" as well, so yes, enemy here is considered as having the superiority in shooters, and if enemy has superiority in shooters, then defensive tactics will not be chosen.

oleg-derevenetz:
> Maybe we should allow attack within movement range in such situations? I suppose it's out of the scope of this PR, right?

Yes, first of all, it's out of scope of this PR, and second, if we allow attack the enemy targets on the enemy half of the battlefield within their movement range, they may get too far away from their archers and fail to help them when the units covering them are attacked or even completely destroyed. Not to mention that they won't be able to choose defensive tactics while on the opposite side of the battlefield - with the latest changes in this PR. But they still can attack enemies on their own half of the battlefield, of course.

oleg-derevenetz:
Current implementation of the AI defensive tactics in the `master` branch does not defend archers at all. I have no idea how anything can be "worse" than this:

https://github.com/ihhub/fheroes2/assets/32623900/722a4679-3a41-46c7-be1d-6dffaf7ec68e

And I should note that this is not some corner case like that "special" arrangement "on the edge" in which the death of one unit changes tactics, no. In the `master` branch AI always behaves like that. The same army deals with these 50 Mummies just fine with proper management. This is AI management from this PR:

https://github.com/ihhub/fheroes2/assets/32623900/a6b7922b-e85f-4bf4-aab7-e8dbcb70dddb

and this is my manual management "in the style of this PR":

https://github.com/ihhub/fheroes2/assets/32623900/f2e5ddf2-9c49-4abc-a243-0157b65065cb

Manual management gave slightly better results, but it's just a pure luck (Mummies just did a little less damage overall due to the randomness of this process).

Branikolog:
@oleg-derevenetz 

>In your video defending hero has attack spell which counts as "shooter strength" as well, so yes, enemy here is considered as having the superiority in shooters, and if enemy has superiority in shooters, then defensive tactics will not be chosen.

Oh, forgot about this AI logic. My apologies!

>A restriction that enables defensive tactics only for units in our half of the battlefield also does not solve this problem 100%, because it may happen that some units will be in our half of the battlefield (and they will retreat to cover the archers), and some will be in the enemy's half (and they will not retreat).

You're right. But after some tests I came up with a conclusion, that it works better in average: such logic reduces the chance of initiating defensive tactics and pretty fast swordsmen tend to continue their attacks all together on the second half on a battlefield. I mean cases, when a single stack was left alone against the whole enemy army occur not so often now.

To reduce the chance of leaving a single troop on the second half of a battlefield some additional rules could be implemented later or some day. :)
For instance, if >=50% of melee troops turned into shooter-defensive state, all troops, that faster than enemy should also run away, even if they need more than 2 turns to reach shooters or if they are located on the second half of the battlefield.


From my observations:
Sometimes slow troops tend to stay on the first half of the battlefield guarding shooters, while there's completely no threat for them. (I think we still need to have some sort of logic "Attack all" and "Defend all" to prevent some troops from chilling near shooters or sacrificing 1 stack on the second part of the battlefield, while others are coming back).
Sometimes during a fight on that edge case of equal strength of shooters and melee troops, AI decides to protect shooters placing one stack right near them, while there're lots of enemies nearby wasting that stack for nothing. But this maneuver occurs not really often.

For now, as a starting point we can merge this PR.
Sorry for such a long response. I really appreciate your work on this improtant logic.

Branikolog (approved):

oleg-derevenetz:
> Sometimes slow troops tend to stay on the first half of the battlefield guarding shooters, while there's completely no threat for them.

It's probably not that they're "tend to stay", they simply do not have time to get anywhere until, due to the losses of faster units, the shooters' strength begins to exceed the same 15% of the total army strength, so they go on the defensive. Regarding "no threat", hm... since they are slow, it makes no sense for them to attack the other half of the battlefield, because the faster enemy units will simply bypass them (again, I'm talking about proper human-controlled behavior, not AI behavior that doesn't like "run without attacking" and often gets unnecessary losses because of this) and block/destroy the archers first, and then these slow units.

Branikolog:
@oleg-derevenetz 

>It's probably not that they're "tend to stay", they simply do not have time to get anywhere until, due to the losses of faster units, the shooters' strength begins to exceed those 15% of the total army strength, so they go on the defensive. Regarding "no threat", hm... since they are slow, it makes no sense for them to attack the other half of the battlefield, because the faster enemy units will simply bypass them (again, I'm talking about proper human-controlled behavior, not AI behavior that doesn't like "run without attacking" and often suffers unnecessary losses because of this) and block/destroy the archers first, and then these slow units.

Sure, it is definitely a corner case, which cannot be handled by simple logic. And yes, in a case enemy has faster troops, he can bypass slow golems standing near rangers. 
I hope some day we can work on such corner cases ad teach AI maneuvering to rush enemy from a certain angle not to be bypassed, or leave shooters and come to help the rest of army fighting far away from defended shooters, or step forward to meet the approaching enemies, in a case it's just  single stack which cannot cover shooters well.

oleg-derevenetz:
> leave shooters and come to help the rest of army fighting far away from defended shooters

They can do it now as well. If enemy melee units will suffer losses in such a way that enemy army will have an excess of archers and a lack of units to cover them (more specifically, more than 2/3 of the enemy army strength will be made up of archers) AI will go offensive. And of course it will go offensive with a large overall advantage of his army over the enemy.

> or step forward to meet the approaching enemies

That's what AI in `master` branch currently do :) In fact, I consider such behavior harmful. Maybe it looks "cool", but in fact such "step forward" brings nothing but additional problems, because it just leaves an extra cell near friendly archers free for an extra enemy unit, but gives no real benefit.

Branikolog:
@oleg-derevenetz 

>That's what AI in master branch currently do :) In fact, I consider such behavior harmful. Maybe it looks "cool", but in fact such "step forward" brings nothing but additional problems, because it just leaves an extra cell near friendly archers free for an extra enemy unit, but gives no real benefit.

Sure, if we have a sufficient number of stacks to cover shooters, then keeping the position is the best tactics.

But if there's only one stack to cover shooters, I don't see any reasons to stay near in such situation:

image

>They can do it now as well. If enemy melee units will suffer losses in such a way that enemy army will have an excess of archers and a lack of units to cover them (more specifically, more than 2/3 of the enemy army strength will be made up of archers) AI will go offensive, even if it has superiority in archers. And of course it will go offensive with a large overall advantage of his army over the enemy.

image

In this situation Golems just stayed for a 4-5 turns until swordsmen were eliminated. Mummies barely had a chance to reach rangers, so involving golems into a battle from a certain angle would help to minimize the losses.

But the first thing I wish to be added after this PR will be merged is attacking of defending troops while keeping the position, as currently they only can retaliate or switch to attacking state, when enemies are almost finished.

oleg-derevenetz:
> But if there's only one stack to cover shooters, I don't see any reasons to stay near in such situation:

But why actually run forward here? First, the golems occupy a cell near the shooters, which otherwise would simply be occupied by the enemy. Secondly, by occupying this particular cell, golems force enemy golems to go one cell further in order to reach the archers, which can result in an extra turn, during which the shooters can have time to make an extra shot.

> Mummies barely had a chance to reach rangers

:))) There is no problem for Mummies to reach Rangers in this case at all. All that prevents them is a dumb red AI that allows archers to freely shoot a crowd of mummies from afar. A human player would definitely try to block the shooters, especially if the blue AI would not cover them.

> But the first thing I wish to be added after this PR will be merged is attacking of defending troops while keeping the position, as currently they only can retaliate

Without attacking, they will last more turns (maybe even twice as long). I have already demonstrated this effect in one of the comments here before. This is the whole point of defensive tactics: do not get into trouble, do not incur unnecessary losses, let the shooters do all the work.

Here are two videos for comparison where the attacker is guaranteed to win. In the first video, I control the army manually and force the Swordsmen to attack:

https://github.com/ihhub/fheroes2/assets/32623900/02903601-5ce2-4d0e-99d2-7e289989d2cd

All Mummies are dead, 12 Swordsmen are lost. In the next video, I let the AI perform the defense (no attacks, just retaliation):

https://github.com/ihhub/fheroes2/assets/32623900/f3b9db13-d10d-4fde-879d-b01c33c82e4d

All Mummies are dead and... 12 Swordsmen are lost, just like in the first video. Then why introduce another coefficient that needs to be carefully selected and take into account various corner cases - to avoid a situation in which defenders simply won't be able to hold out long enough if they attack - just like in this case:

https://github.com/ihhub/fheroes2/assets/32623900/5bc2a8c2-d518-45d3-b5d1-c40b1acb8cb7

? BTW, here is another video when AI uses "non-attacking" defensive tactics, with the same initial setup:

https://github.com/ihhub/fheroes2/assets/32623900/eb2ec270-a258-458a-b283-1c33bbd54cd2

Why, Mr. Anderson? Why, why, why? So that people who don't know how to play don't complain about "unnatural behavior" or "bug"? Honestly, I don't care much about that. I don't think this is a top priority task or a task at all. IMHO what really needs to be thought about first regarding the defensive tactics is more fast (in fewer turns) and maybe more optimal (with fewer units) cover for archers.

ihhub:
@oleg-derevenetz , thank you so much for these changes and patience with @Branikolog !

PR #8052, opened 2023-11-19
Make the touch device ID part of the finger ID

oleg-derevenetz:
This should allow gestures to be handled correctly even when using different touchpads for different fingers. Although I have not met with this yet, in theory multiple touchpads can be used simultaneously, and in this case, without these changes, the game may not behave as the user expects.


ihhub (approved):

ihhub:

PR #8056, opened 2023-11-19
Reduce gold income and hero movement bonuses for AI

ihhub:
Since we added logic for marketplace usage by the AI and it has a very significant gold bonus, it is very hard to face the AI on highest difficulties. Having 60% of gold bonus is too huge. We are going to expand marketplace logic for monster purchase and upgrades so we should limit its bonuses.

This pull request also reduces movement bonus for the AI heroes on Expert difficulty to differentiate from Impossible difficulty.

These changes are subjective and you are welcome to suggest something different or even decline the proposal changes.

### DIFF
--- a/src/fheroes2/game/difficulty.cpp
+++ b/src/fheroes2/game/difficulty.cpp
@@ -66,16 +66,28 @@ int Difficulty::GetScoutingBonus( int difficulty )
 
 double Difficulty::getGoldIncomeBonusForAI( const int difficulty )
 {
+    // The original game has the following bonuses for AI:
+    // Easy - -0.25 (-25%)
+    // Normal - 0.0 (no bonus)
+    // Hard - 0.29 (29%)
+    // Expert - 0.4 (40%)
+    // Impossible - 0.6 (60%)
+    //
+    // However, the current AI is able to use marketplace to trade resources.
+    // This means that the AI has a lot of resources at the end which might break the balance.
+    //
+    // Therefore, gold income bonuses are reduced by about 17%.
+
     switch ( difficulty ) {
     case Difficulty::EASY:
         // It is deduction from the income.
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub although the logic for using markets by AI was added almost a month ago, a wide audience of players met with it at best yesterday, after the release of the next version. I propose to get some feedback from the players, and then make a decision.

Branikolog:
Hi, @ihhub 
I personally experienced that in the majority of cases AI manages to buy 90% of troops in the castles that is threatened by my forces (considering the fact that AI is constantly facing other players). Only on a later game stages, when AI is close to being defeated he starts experiencing gold shortage. It seems to me that he has more than enough money currently.
I believe we're ready to decrease a little (10-20%) money cheat for him.
But this thing needs to be tested thoroughly.

kitovyj:
AI also has a movement bonus? I beleive that is a bit too much. At least I would prefer it to be a separate option.

oleg-derevenetz:
> AI also has movement bonus? I believe that is a bit too much.

AI has movement bonus of 75 movement points on Expert and Impossible difficulties because it have the same bonus in the original game. Consider:

https://github.com/ihhub/fheroes2/assets/32623900/3f734f08-ccdd-48af-8948-97bddf78d439

kitovyj:
> > AI also has movement bonus? I believe that is a bit too much.
> 
> AI has movement bonus of 75 movement points on Expert and Impossible difficulties because it has the same bonus in the original game. Consider:
> 
>  DOSBox.0.74-2.1.Cpu.speed_.max.100.cycles.Frameskip.0.Program_.HEROES2.2024-01-13.03-10-01.mp4

I see. Thank you for the clarification.

Issue #7491, opened 2023-07-30
Hero screen, Necromancy icon name alignment


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

![ezgif com-resize (1)](https://github.com/ihhub/fheroes2/assets/55348946/156e80ec-57c1-4a0c-a6d3-916158bd07b0)

The skill name should be shifted for 1 px to the left.

### Save file

None.

ihhub:
Nothing we can do about it as the rendering of skill names is the same. English is not the only language which requires special handling for some skill names. Moreover, the font has shadow which is exactly 1 pixel on the left. If we shift the name then the shadow will appear on the golden frame.

Therefore, I am closing this issue as we aren't going to fix it.

Branikolog:
Hi, @ihhub !

In the original game the skill name is positioned the same way I was mentioned in the issue. 

![image](https://github.com/ihhub/fheroes2/assets/55348946/bd071cda-264e-4f5c-958e-0b79792e807c)

I think cutting shadow for such cases is a nice visual improvement of the game UI.
Even without fixing shadow of the first letter it looks better than the current implementation of fheroes2.

ihhub:
Fixed in #8058 .

a1exsh:
Hm, seems to be missing the handling in the pop-up dialog when checking the skill:
![Screenshot from 2024-01-04 15-11-22](https://github.com/ihhub/fheroes2/assets/489601/c79e243a-cd0c-4d56-8560-1336e89cea61)

The hero details itself is fixed:
![Screenshot from 2024-01-04 15-15-08](https://github.com/ihhub/fheroes2/assets/489601/2f419754-72b9-4343-9d2c-8eac09cae7d1)

JasonQorbin:
Hi. That's correct. We said that the pop-up would be fixed in a separate PR. Will try to have a look at this over the weekend.

PR #8058, opened 2023-11-19
Shifted the name of the Necromancy skill in the secondary skill bar l…

JasonQorbin:
This PR addresses issue #7491. The change makes th UI render more like the original game by rendering the name of the Necromancy skill one pixel to the left of center so that the long skill name doesn't overlap with the gold border on the righthand side of the secondary skill tile. The solution is not perfect because the text has a shadow that now touches the left side border but the original game looked like this also.

A few things to note about the code changes:

1. I opted to add a check to see if the current skill is the necromancy skill and handle the special case instead of rounding the result of the padding calucation down because I didn't want to affect all the exiting content (which no one has complained about and is probably happy with).
2. The problem pointed out by @ihhub is still relevant: This only fixes the visual bug for the english version of the game and for long strings in other languages we will need a better solution. Handling a special case for every long skill name in every language is not feasible.
3. I added a lambda for the centering logic which is repeated several times in this function. This may be beyond the scope of the issue. Happy to revert that back.

Happy to take guidance on this to arrive at a better more general solution


JasonQorbin:
Sorry. Updated formatting with clang-format.

ihhub:
Hi @JasonQorbin , the root of the problem is in presence of shadow within the font which is 1 pixel wide. The shadow is present at left and bottom sides. Therefore, in order to tackle the problem we need to get width of the text, width of the icon and calculate with 1 pixel shift:
```cpp
xPos = ( iconWidth - textWidth + 1 ) / 2;
```
Essentially, the whole fix is
```cpp
text.draw( pos.x + ( pos.width - text.width() + 1 ) / 2, pos.y + 53, dstsf );
```

Please give a try.

JasonQorbin:
I didn't realize I could squish the text. Thanks for that. I'll take another look this evening.

JasonQorbin:
Hi @ihhub. I've tried a few approaches.

The fix you recomended solves the problem of the name overlapping the righthand border, which was the original problem we were trying to solve. However, it still casts a shadow on the lefthand border like so:

Necromancy with 1px adjustment

I then attempted the suggestion from @Branikolog in the original issue of "cutting the shadow for such cases".
So in the current version of the PR, if the text is just wide enough that the shadow would fall on the border then
I repaint that section of the border to remove the shadow. This is the result:

Necromancy with 1px adjustment and border repainted

ihhub:
Hi @JasonQorbin , I am adding changes to limit rendering of text within a given ROI. See #8101. With this approach it will be easier to avoid any hacky code.

JasonQorbin:
Text, centered, no more shadow, and no more hacky redrawing of the border thanks to the new text drawing function.
Screenshot 2023-12-10 204357

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -238,14 +238,15 @@ void SecondarySkillsBar::RedrawItem( Skill::Secondary & skill, const fheroes2::R
 
     if ( use_mini_sprite ) {
         const fheroes2::Text text{ std::to_string( skill.Level() ), fheroes2::FontType::smallWhite() };
-        text.draw( pos.x + ( pos.width - text.width() ) - 3, pos.y + pos.height - 10, dstsf );
+        text.drawInRoi( pos.x + ( pos.width - text.width() ) - 3, pos.y + pos.height - 10, dstsf, pos );
     }
     else {
         fheroes2::Text text{ Skill::Secondary::String( skill.Skill() ), fheroes2::FontType::smallWhite() };
-        text.draw( pos.x + ( pos.width - text.width() ) / 2, pos.y + 5, dstsf );
+        int skillNamePaddingX = ( pos.width - text.width() ) / 2 - 1;
```
ihhub, line 245: This variable can be const.

ihhub (commented):
Now the code is very nice and neat :)

ihhub:
Hi @Branikolog , please test these changes since you requested them.

src/fheroes2/gui/skill_bar.cpp
JasonQorbin: Done. Also mirrored the same variable for the skill level text for consistency.

ihhub (approved):

Branikolog:
Hello, @JasonQorbin !
Thanks for working on this issue.
Necromancy skill name aligned properly in the hero info window now. 👍 
But lvlup window still has the issue.
image

ihhub:
Hi @Branikolog , the level up window is a separate issue.

ihhub:
Hi @Branikolog , I suggest to check other languages and how it looks for skill names.

Branikolog:
Hi, @ihhub 
Okay.
What about skill info window?

image

It is aligned properly, but there's a shadow on a frame, while in hero info window shadow was dismissed for the frame?

Branikolog:
@ihhub 

>Hi @Branikolog , I suggest to check other languages and how it looks for skill names.

Thanks to the ability to select language in the adventure map settings it was really easy and comfortable. :)

Symbols are cut for the names on other languages:

image

image

Such implementation is not OK in my opinion.

JasonQorbin:
I'm happy to apply the fix the other menus and windows. But I would appreciate some guidance on what sort of intervention would be considered OK for handling languages that, it would appear, the UI wasn't designed for. In that last screenshot, the text of the skill levels of adjacent skill would overlap if they aren't truncated. And those are single words so we can't render it out on two separate lines.

Maybe we can consider scaling the text down? Not sure that would look good and very long words would become illegible very quickly at some resolutions.

Maybe combined with scaling the UI up? Might be fine on higher resolutions and larger screens but that wouldn't work at the resolution shown in the screenshots above.

I'm just kicking around ideas to try to spark something. I'm not really sure what the answer is either.

JasonQorbin:
Hi everyone.

I've changed my local copy to truncate the text with ellipses ("...") if it is any longer than the max width of the block in the Hero Screen as suggested by @zenseii. 

Excuse the weird character encoding, but I think it gets the point across.

Screenshot 2023-12-14 001606

@Branikolog if you're happy with this then I'll update the pull request with this change and then we can create a new issue(s) to fix both the level-up window and the one that pops up when you right-click the skill.

Branikolog:
Hi, @JasonQorbin and @zenseii 
Yeah, that's a nice solution. We have some other names too long for these skill boxes. So showing "..." is definitely better than text overlapping the nearest icon text.
image

oleg-derevenetz:
Hi @JasonQorbin there is something weird about this PR, too many changed lines (4036 lines added, 2276 lines removed, 70 files affected). Please sync it with `master` properly (e.g. by merging `master` into this PR's branch).

JasonQorbin:
Apologies. Should be fixed now. Should only be 11 lines in one file.

I rebased my commits together and ended up overwriting other peoples commits. Terribly sorry!!

Branikolog (approved):

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -237,15 +237,23 @@ void SecondarySkillsBar::RedrawItem( Skill::Secondary & skill, const fheroes2::R
...
+        fheroes2::Text text{ std::to_string( skill.Level() ), fheroes2::FontType::smallWhite() };
+        if ( text.width() > 76 )
+            text.fitToOneRow( pos.width );
+        text.drawInRoi( pos.x + ( pos.width - text.width() ) - 3, pos.y + pos.height - 10, dstsf, pos );
     }
     else {
         fheroes2::Text text{ Skill::Secondary::String( skill.Skill() ), fheroes2::FontType::smallWhite() };
-        text.draw( pos.x + ( pos.width - text.width() ) / 2, pos.y + 5, dstsf );
+        if ( text.width() > 76 )
+            text.fitToOneRow( pos.width );
```
ihhub, line 248: Please add parenthesis for this if-body:
```cpp
if ( text.width() > 76 ) {
    text.fitToOneRow( pos.width );
}
```

The same applies for the code below.

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -237,15 +237,23 @@ void SecondarySkillsBar::RedrawItem( Skill::Secondary & skill, const fheroes2::R
     fheroes2::Blit( sprite, dstsf, pos.x + ( pos.width - sprite.width() ) / 2, pos.y + ( pos.height - sprite.height() ) / 2 );
 
     if ( use_mini_sprite ) {
-        const fheroes2::Text text{ std::to_string( skill.Level() ), fheroes2::FontType::smallWhite() };
-        text.draw( pos.x + ( pos.width - text.width() ) - 3, pos.y + pos.height - 10, dstsf );
+        fheroes2::Text text{ std::to_string( skill.Level() ), fheroes2::FontType::smallWhite() };
+        if ( text.width() > 76 )
+            text.fitToOneRow( pos.width );
+        text.drawInRoi( pos.x + ( pos.width - text.width() ) - 3, pos.y + pos.height - 10, dstsf, pos );
```
ihhub, line 243: I don't think that we need to modify this code. There is literally only 1 digit to render. I suggest to revert these changes. This part of code is executed in Kingdom Overview dialog.

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -237,15 +237,23 @@ void SecondarySkillsBar::RedrawItem( Skill::Secondary & skill, const fheroes2::R
...
     }
     else {
         fheroes2::Text text{ Skill::Secondary::String( skill.Skill() ), fheroes2::FontType::smallWhite() };
-        text.draw( pos.x + ( pos.width - text.width() ) / 2, pos.y + 5, dstsf );
+        if ( text.width() > 76 )
```
ihhub, line 247: We must avoid hardcoded values. This condition should be based on **sprite** width which is 75 pixels. In other words, the condition should be `if ( text.width() > sprite.width() + 1 )`.

ihhub (changes requested):
Hi @JasonQorbin , I left several comments in this pull request. Can you please take a look a them?

ihhub (approved):

ihhub:
@JasonQorbin , thank you so much for your contribution and patience with this issue!

PR #8059, opened 2023-11-19
Enable Dependabot for actions from the GitHub Marketplace

oleg-derevenetz:
To keep these actions up-to-date.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#github-actions

PR #8063, opened 2023-11-21
Do not run the SonarCloud Analyzer workflow for Dependabot PRs

oleg-derevenetz:
Related to #8059. Pull requests created by Dependabot have no access to the secrets, so this workflow will fail.


ihhub (approved):

ihhub:

PR #8066, opened 2023-11-22
Group objects for the Editor into logical buckets

ihhub:
relates to #6845

Also make top object parts do not have any layer as they should.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -816,7 +816,7 @@ int Dialog::selectHeroes( const int heroId /* = Heroes::UNKNOWN */ )
 
 int Dialog::selectHeroType( const int heroType )
 {
-    const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::Hero );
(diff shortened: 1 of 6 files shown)


Districh-ru (commented):
Hi, @ihhub.
I did a quick test and found that the Ocean objects placing is broken now:
- there are less objects in the list,
- when you select one object the cursor shows the other one,
- you can not place any Ocean object on the map.

https://github.com/ihhub/fheroes2/assets/113276641/d3c51d06-a70d-43f2-b095-4a758ab431e0

ihhub:
Hi @Districh-ru , thanks a lot for finding these issues. I will fix them today.

Districh-ru (approved):
Well done, @ihhub! But some water objects are now in the new category and currently inaccessible.
We should add buttons (maybe with the temporary images) to access this and other new categories in one of tрe next PRs.

ihhub:
Hi @Districh-ru , other water objects exist under a separate category.

Districh-ru:
> Hi @Districh-ru , other water objects exist under a separate category.

Yes, I see it, but currently there is no ability to select any of them in Editor. So I mentioned that we should implement new buttons before the release of Editor.

Issue #8025, opened 2023-11-06
Main menu, file list shouldn't be updated after file was deleted


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

https://github.com/ihhub/fheroes2/assets/55348946/fe5db019-9e34-4e4a-89a0-438d6224cb1a

### Describe the possible solution

File list shouldn't jump to the upper position after any file was deleted. It forces a user to scroll down the whole list each time he needs to delete a couple of files.

PR #8071, opened 2023-11-22
fix scroll location on file deletion

MaMadDl:
Fix for #8025

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -509,6 +509,9 @@ namespace
                     listbox.updateScrollBarImage();
 
                     listbox.SetListContent( lists );
+
+                    listbox.SetCurrent( std::max( listId - 1, 0 ) );
+
                 }
 
                 needRedraw = true;
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -509,9 +509,7 @@ namespace
                     listbox.updateScrollBarImage();
 
                     listbox.SetListContent( lists );
-
                     listbox.SetCurrent( std::max( listId - 1, 0 ) );
-
                 }
 
                 needRedraw = true;
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -494,6 +494,9 @@ namespace
             }
 
             if ( !isEditing && le.KeyPress( fheroes2::Key::KEY_DELETE ) && isListboxSelected ) {
+                listbox.SetCurrent( listId );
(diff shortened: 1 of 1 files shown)


src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -509,6 +509,7 @@ namespace
                     listbox.updateScrollBarImage();
 
                     listbox.SetListContent( lists );
```
Districh-ru, line 511: We can also remove this line to avoid list position reset: a reference to `lists` is already set for the contents of the `listbox` and the action of the `SetListContent()` method is the reset of selected item and list position. The other action is list scrollbar range update which is now done by the newly added `SetCurrent()` method. So we can safely remove this method call and preserve the list position.
What do you think?)

Districh-ru (changes requested):
Hi @MaMadDl.
Thanks for fixing this issue!
We can make this fix a little better. I left one comment regarding this. Could you please check it out in your spare time.

Districh-ru (approved):

Branikolog:
I haven't found any bugs in this implementation. 👍 
The only thing concerns me is that list jumps to the target list string **after** the file was deleted:

https://github.com/ihhub/fheroes2/assets/55348946/fb0098ce-4393-43cd-bc6a-1966f7079f33

Maybe the list should be updated right before delete dialog appears?
How do you think, @MaMadDl  @Districh-ru and @ihhub ? Is it a more common behaviour of lists in modern OSes?

Districh-ru:
> I haven't found any bugs in this implementation. 👍 The only thing concerns me is that list jumps to the target list string **after** the file was deleted:
> Maybe the list should be updated right before delete dialog appears? How do you think, @MaMadDl @Districh-ru and @ihhub ? Is it a more common behaviour of lists in modern OSes?

This can be implemented if we add between the lines 496 and 497 the code to show current file and redraw the list:
```cpp
                listbox.SetCurrent( listId );
                listbox.Redraw();
```

Then the list will act like it is done, in example, in Total Commander. :)

MaMadDl:
> I haven't found any bugs in this implementation. 👍 The only thing concerns me is that list jumps to the target list string **after** the file was deleted:
> 2023-11-25.12-04-52.mp4
> 
> Maybe the list should be updated right before delete dialog appears? How do you think, @MaMadDl @Districh-ru and @ihhub ? Is it a more common behaviour of lists in modern OSes?
> 
> The implementation from the PR works fine anyway.

I think this is expected behaviour because you are deleting a file that's not shown on the screen at the time. I have seen both behavoiours, but I think snapping to selected file should happen either way.

what @Districh-ru suggested can be a nice touch

Branikolog (approved):
No bugs detected. 👍

ihhub (approved):

ihhub:
@MaMadDl , thank you very much for your contribution to the project!

Issue #859, opened 2020-06-21
"Elemental Storm" spell animation doesn't cover whole battlefield


Branikolog:
We can see a small area near the right edge without animation.
![Storm](https://user-images.githubusercontent.com/55348946/85222756-6e152f00-b3c6-11ea-8d26-a867734be0ff.jpg)

idshibanov:
Fix here have to wait until #815 is merged to avoid the confict.

Issue #261, opened 2019-12-23
Monster count window has incorrect position in battle


ihhub:
Here is a screenshot of a battle. As you could see most of monster count windows are on the top of monsters (notice number 20 at Goblins):
![image](https://user-images.githubusercontent.com/19829520/71361239-5acb0780-25cd-11ea-804d-cf18f9228f64.png)

And here is a screenshot how it should look correctly:
![image](https://user-images.githubusercontent.com/19829520/71361268-71715e80-25cd-11ea-8edd-02ea7ae2c658.png)

ihhub:
The position of the count is defined in **MonsterAnimInfo** structure.

PR #8072, opened 2023-11-23
Add multi-display device support

MaMadDl:
Adding Multi Monitor Support
Tested on Windows (Win 11) 
Untitled


ihhub:
Hi @MaMadDl , can you please explain what problem this pull request solves?

MaMadDl:
when you have more than one monitor 
and you want the game to run on your second (or third) Monitor you can't really change it.
the Ctrl+Alt+WinKey doesnt always works
so i added the functionality to do so .

MaMadDl:
It should cause we are using SDL to handle monitor coordinates 
I have linux (ubuntu 20.04) as well but haven't got to test it yet

ihhub:
Hi @MaMadDl , what do you mean I can't change it if it is possible to drag a window in between monitors? I'm just trying to understand the problem before trying to analyze a solution for it.

Could you also elaborate "the Ctrl+Alt+WinKey doesnt always work"? In which cases it works and in which it doesn't?

MaMadDl:
when you are in fullscreeen (and you want to play in fullscreen not windowed)  game is pretty much stuck on the first monitor (primary monitor) connected. 
usually some engines allow moving between monitors using Crtrl+Alt+ ArrowKey to move game window between monitors. this engine doesn't ATM. 
so in order to play the game on fullscreen  on any other monitors than the first monitors you need to add functionality on engine level.  and that's what i did.

MaMadDl:
@Mr-Bajs 
Apparently X11 can correctly handle screen switch internally ( I could move game window between monitors using `WinKey+Shift+ArrowKey`)

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -24,6 +24,8 @@
 #include <string>
 #include <utility>
 
+#include <SDL_video.h>
```
oleg-derevenetz, line 27: SDL is not welcome outside of the `engine`. You should not directly use SDL in `fheroes2`, only through the medium of `engine`.

oleg-derevenetz, line 27: I will not touch on the general need for these changes yet, but I should note that SDL is not welcome outside of the `engine`. You should not directly use SDL in `fheroes2`, only through the medium of `engine`.

oleg-derevenetz (changes requested):

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: I didn't add it voluntarily `IWYU` workflow forced that. check [20f480e](https://github.com/ihhub/fheroes2/pull/8072/commits/20f480ed394de3ed354f1e635b4c468c115d968c) build status

as for the need, I needed this kind of functionality when I'm playing the game so I tried to add it.

oleg-derevenetz: I mean that only low-level engine (code in the `src/engine`) should work with SDL and it should offer an abstract interface for code in `src/fheroes2` to work with hardware. No code in `src/fheroes2` should work with the SDL directly. But this can be corrected later after a principled decision from @ihhub about taking this PR into work.

src/engine/screen.h
```diff
@@ -139,10 +139,21 @@ namespace fheroes2
             return _nearestScaling;
         }
 
+        int DisplayMonitor() const
+        {
+            return _displayMonitor;
+        }
+
+        void setDisplayMonitor( int monitor )
```
ihhub, line 147: The input argument can be const.

src/engine/screen.h
```diff
@@ -210,6 +223,16 @@ namespace fheroes2
         // Do not call this method. It serves as a patch over the basic class.
         void resize( int32_t width_, int32_t height_ ) override;
 
+        void setDisplayMonitor( int monitor )
```
ihhub, line 226: We have exactly the same methods and variable inside the class above. Why we repeat them here?

src/fheroes2/dialog/dialog_graphics_settings.cpp
ihhub: @oleg-derevenetz is right. We must not expose SDL code outside **engine**. However, the bigger concern of mine is these UI changes. The options are not aligned and the new option must have a new icon according to [this description](https://github.com/ihhub/fheroes2/wiki/Adding-a-new-option-into-the-game). Therefore, if you would like to proceed with this feature but do not have a solution for UI please revert all UI changes here.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -509,6 +509,7 @@ namespace
                     listbox.updateScrollBarImage();
 
                     listbox.SetListContent( lists );
+                    listbox.SetCurrent( std::max( listId - 1, 0 ) );
```
ihhub, line 512: Please revert this change. This is not related to the current PR.

src/fheroes2/game/fheroes2.cpp
```diff
@@ -166,10 +166,12 @@ namespace
                         bestResolution = info;
                     }
                 }
-
+                // set display monitor to first monitor available
+                display.setDisplayMonitor( 0 );
```
ihhub, line 170: There is no reason to set to 0 as this is a default value we set in the member of **BaseRenderEngine** class.

src/engine/screen.cpp
```diff
@@ -1071,8 +1071,10 @@ namespace
             }
 
             flags |= SDL_WINDOW_RESIZABLE;
-
-            _window = SDL_CreateWindow( _previousWindowTitle.data(), _prevWindowPos.x, _prevWindowPos.y, resolutionInfo.screenWidth, resolutionInfo.screenHeight, flags );
+            SDL_Rect monRect;
+            SDL_GetDisplayBounds( DisplayMonitor(), &monRect );
+            _window = SDL_CreateWindow( _previousWindowTitle.data(), monRect.x + _prevWindowPos.x, monRect.y + _prevWindowPos.y, resolutionInfo.screenWidth,
+                                        resolutionInfo.screenHeight, flags );
```
ihhub, line 1077: Could you please explain why we need this change?

src/engine/screen.cpp
```diff
@@ -1161,7 +1163,6 @@ namespace
         SDL_Renderer * _renderer;
         SDL_Texture * _texture;
         int _driverIndex;
-
```
ihhub, line 1164: Please revert this line.

src/engine/screen.cpp
```diff
@@ -1336,11 +1337,11 @@ namespace fheroes2
              && info.screenHeight == _screenSize.height ) // nothing to resize
             return;
 
+        _engine->setDisplayMonitor( _displayMonitor );
         const bool isFullScreen = _engine->isFullScreen();
 
         // deallocate engine resources
         _engine->clear();
-
```
ihhub, line 1343: Please revert this line.

src/engine/screen.h
```diff
@@ -178,6 +189,8 @@ namespace fheroes2
     private:
         bool _isFullScreen;
 
+        int _displayMonitor{ 0 };
+
         bool _nearestScaling;
```
ihhub, line 194: Either make every member to be initialized in place (and remove the default constructor) or initialize every member inside the constructor to be consistent with the code.

ihhub (changes requested):
Hi @MaMadDl , I left multiple comments in this pull request. Could you please take a look at them?

src/fheroes2/dialog/dialog_selectfile.cpp
MaMadDl: Yes. this is included here by mistake

src/engine/screen.cpp
MaMadDl: In order for SDL to show display on a monitor, it need it's global coordinate which is calculated between all connected monitors. 
in a One monitor setup, `x` and `y` coordinates of monitor is always `0` and can be ignored in calculations. but in Multi-monitor setup starting position of monitor (top-left side) is moved in negative or positive side by the resolution of the monitor
so the calculation must be included when we are creating window (`SDL_CreateWindow()` )

src/engine/screen.h
MaMadDl: That is my Thoughts exactly. this is caused by Analyzer checks forcing changes
check the [Bot Review](https://github.com/ihhub/fheroes2/pull/8072#pullrequestreview-1746977179) for build [d88c6fd](https://github.com/ihhub/fheroes2/pull/8072/commits/d88c6fd8c0f5b53fc2e5bbd47786fa27cb549176)

src/engine/screen.h
MaMadDl: I couldn't find a cleaner way to pass display number from `settings` to the `engine`. creating instance to setting inside engine was messier in my opinion.
and I thought in the future there might be a need to this in display class as well. so I kept a copy

src/engine/screen.h
```diff
@@ -139,10 +139,21 @@ namespace fheroes2
             return _nearestScaling;
         }
 
+        int DisplayMonitor() const
```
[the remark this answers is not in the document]
MaMadDl: this is acceptable, changes made so fast I didn't thought of a better naming for the property and method

I will change this after all other comments resolved.

ihhub:
Hi @MaMadDl , I marked this pull request as a draft as it still requires some work from your side. Once you address all comments please re-request review.

MaMadDl:
@ihhub @oleg-derevenetz 
changed most of the reviews required changes

As for the dialog itself, instead of creating a new 5 button layout
I've added a button missing in the setting (Interface Type which is accessible only from in-game)

src/engine/screen.cpp
```diff
@@ -1072,7 +1072,14 @@ namespace
 
             flags |= SDL_WINDOW_RESIZABLE;
 
+#if defined( _WIN32 )
+            SDL_Rect monRect;
```
ihhub, line 1076: As I explained previously we should not use a word **monitor** anywhere. Please use **display**. This is applicable for all newly added code and texts.

src/engine/screen.cpp
```diff
@@ -1072,7 +1072,14 @@ namespace
...
+            SDL_GetDisplayBounds( getDisplayId(), &monRect );
+            _window = SDL_CreateWindow( _previousWindowTitle.data(), monRect.x + _prevWindowPos.x, monRect.y + _prevWindowPos.y, resolutionInfo.screenWidth,
+                                        resolutionInfo.screenHeight, flags );
```
ihhub, line 1079: Could you please explain why we have this code only for windows? Is there an SDL documentation regarding this?

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -120,13 +132,33 @@ namespace
...
+        }
+        else {
+            value = _( "Good" );
+        }
+        goodInterface = fheroes2::Crop( goodInterface, 0, 0, goodInterface.width() / 2, goodInterface.height() );
+        fheroes2::Blit( goodInterface, evilInterface );
+
+        fheroes2::drawOption( optionRoi, evilInterface, _( "Interface Type" ), std::move( value ), fheroes2::UiOptionTextWidth::TWO_ELEMENTS_ROW );
+    }
+
```
ihhub, line 154: Interface does not below to **Graphics** group as it exists within **Interface Settings** group. Run any map and open System Options to see this.

@Branikolog what were your suggestions regarding extra Graphics options which we are missing?

src/fheroes2/game/fheroes2.cpp
```diff
@@ -166,10 +166,12 @@ namespace
                         bestResolution = info;
                     }
                 }
-
+                // set display monitor to first monitor available
+                display.setDisplayId( 0 );
```
ihhub, line 170: We don't need to set display ID for the first run as this is the default value.

src/fheroes2/system/settings.cpp
```diff
@@ -477,6 +482,9 @@ std::string Settings::String() const
         os << "editor = beta" << std::endl;
     }
 
+    os << std::endl << "# Display Monitor for Multi-Monitor Setups (defaults to first monitor)" << std::endl;
```
ihhub, line 485: The comment is not entirely accurate. This option will still be present even if a device has one display. Also I suggest to put it right after **videomode** option.

src/fheroes2/system/settings.cpp
```diff
@@ -326,6 +327,10 @@ bool Settings::Read( const std::string & filePath )
         _optGlobal.SetModes( GLOBAL_ENABLE_EDITOR );
     }
 
+    if ( config.Exists( "monitor" ) ) {
+        setDisplayId( std::max( config.IntParams( "monitor" ), 0 ) );
```
ihhub, line 331: We don't need a variable in this class. We can directly call **setDisplayId()** method of `BaseRenderEngine`.

src/fheroes2/system/settings.cpp
```diff
@@ -477,6 +482,9 @@ std::string Settings::String() const
         os << "editor = beta" << std::endl;
     }
 
+    os << std::endl << "# Display Monitor for Multi-Monitor Setups (defaults to first monitor)" << std::endl;
+    os << "monitor = " << getDisplayId() << std::endl;
```
ihhub, line 486: We can directly call a method of `BaseRenderEngine` class to retrieve the value.

src/engine/screen.h
```diff
@@ -139,10 +142,21 @@ namespace fheroes2
...
+        void setDisplayId( const int monitor )
+        {
+            _displayId = monitor;
+        }
+
     protected:
         BaseRenderEngine()
             : _isFullScreen( false )
             , _nearestScaling( false )
+
```
ihhub, line 159: Initialize **_displayId** member here.

ihhub (changes requested):
Hi @MaMadDl , I put few comments here. Could you please take a look at them?

@Branikolog please check one comment I left here.

MaMadDl (commented):
Hello @ihhub, I have made the requested changes. Could you please review them?

src/engine/screen.cpp
```diff
@@ -1055,8 +1055,13 @@ namespace
                 ERROR_LOG( "Failed to set a hint for screen orientation." )
             }
 #endif
-
```
ihhub, line 1058: Please revert this change.

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -253,6 +299,17 @@ namespace fheroes2
                 conf.Save( Settings::configFileName );
                 windowType = SelectedWindow::Configuration;
                 break;
+            case SelectedWindow::DisplayList:
+                fheroes2::engine().setDisplayId( ( fheroes2::engine().getDisplayId() + 1 ) % fheroes2::getNumberOfVideoDisplays() );
```
ihhub, line 303: I tested this as a user. What exactly do we expect from this option? It does nothing for my setup: I have 2 monitors and Windows 10. Also, according to logic it also does nothing expect setting a variable to a new state. The change must be visible immediately to a user.

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -49,17 +49,28 @@ namespace
...
     const fheroes2::Rect vSyncRoi{ optionOffset.x, optionOffset.y + offsetBetweenOptions.height, optionWindowSize, optionWindowSize };
     const fheroes2::Rect systemInfoRoi{ optionOffset.x + offsetBetweenOptions.width, optionOffset.y + offsetBetweenOptions.height, optionWindowSize, optionWindowSize };
+    const fheroes2::Rect monitorListRoi{ optionOffset.x + offsetBetweenOptions.width * 2, optionOffset.y, optionWindowSize, optionWindowSize };
+    const fheroes2::Rect interfaceTypeRoi{ optionOffset.x + offsetBetweenOptions.width * 2, optionOffset.y + offsetBetweenOptions.height, optionWindowSize,
+                                           optionWindowSize };
+
+    void drawDisplay( const fheroes2::Rect & optionRoi )
+    {
+        fheroes2::drawOption( optionRoi, fheroes2::AGG::GetICN( ICN::GAME_OPTION_ICON, 1 ), _( "Display ID" ),
+                              fheroes2::getDisplayName( fheroes2::engine().getDisplayId() ), fheroes2::UiOptionTextWidth::TWO_ELEMENTS_ROW );
```
ihhub, line 72: Displaying a name is a bit useless. This is my setup:
image

I am clicking on it with no changes. This is because Windows detects both monitors (laptop and a separate monitor) just generic ones.

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -49,17 +49,28 @@ namespace
         Mode,
         VSync,
         SystemInfo,
+        DisplayList,
```
ihhub, line 52: This is entirely correct: we do not show a display list but switching displays. So more accurate name should be `SwitchDisplay`.

src/fheroes2/game/fheroes2.cpp
```diff
@@ -166,7 +166,7 @@ namespace
                         bestResolution = info;
                     }
                 }
-
+                // set display monitor to first monitor available
```
ihhub, line 169: Please revert this change.

src/fheroes2/system/settings.h
```diff
@@ -443,6 +442,7 @@ class Settings
     BitModes _optGlobal;
 
     fheroes2::ResolutionInfo _resolutionInfo;
+
```
ihhub, line 445: Please revert these changes.

src/engine/screen.cpp
```diff
@@ -1472,4 +1483,14 @@ namespace fheroes2
     {
         return *( Display::instance()._cursor );
     }
+
+    int getNumberOfVideoDisplays()
+    {
+        return SDL_GetNumVideoDisplays();
+    }
```
ihhub, line 1490: This function must check return value of **SDL_GetNumVideoDisplays** according to https://wiki.libsdl.org/SDL2/SDL_GetNumVideoDisplays and log an error if any.

src/engine/screen.h
```diff
@@ -35,6 +35,9 @@ namespace fheroes2
     class Cursor;
     class Display;
 
+    int getNumberOfVideoDisplays();
+    const char * getDisplayName( const int display );
```
ihhub, line 39: These methods should belong to **BaseRenderEngine** class and be overridden in child classes. We do not want to stick only to SDL 2.

src/engine/screen.h
```diff
@@ -178,6 +193,8 @@ namespace fheroes2
     private:
         bool _isFullScreen;
 
+        int _displayId;
```
ihhub, line 196: Please move this member to the bottom. Right now the size of the class based on the variable declaration is 64 * 3 bits while moving this variable to the bottom makes it 64 * 2 bits.

src/fheroes2/dialog/dialog_graphics_settings.cpp
ihhub: I suggest to completely remove this option and make the second row to have just 2 options aligned appropriately. We can copy-paste a part of 2x2 window just for these 2 options.

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -197,9 +237,15 @@ namespace
             else if ( le.MousePressRight( windowVSyncRoi ) ) {
                 fheroes2::showStandardTextMessage( _( "V-Sync" ), _( "The V-Sync option can be enabled to resolve flickering issues on some monitors." ), 0 );
             }
-            if ( le.MousePressRight( windowSystemInfoRoi ) ) {
+            else if ( le.MousePressRight( windowDisplayListRoi ) ) {
+                fheroes2::showStandardTextMessage( _( "Monitor Selection" ), _( "Toggle Between available monitors, Restart Required to take Effect" ), 0 );
```
ihhub, line 241: As explained before we shouldn't use word **monitor**.

ihhub (changes requested):
Hi @MaMadDl , I left few more comments. Please take a look.

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: @ihhub  changes require restart cause i didn't want to recreate window
I have mentioned this on right click help

ihhub: No, they do not. See how the application changes the resolution. We recreate a window from scratch.

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: @ihhub if both monitor are the same name you need to click once then restart to see changes since you can't see them at runtime

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: @ihhub I saw that you are calling `updateUI()` after each change 
I was concerned about incompatibilities might occur if resolution is not supported by the new display. so I decided to not do it.
I will add it

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: @ihhub do you mean something like this ?


this can be done but new resource needs to be added to ICNs
(can be done in runtime but i think it would be messy)

src/fheroes2/dialog/dialog_graphics_settings.cpp
ihhub: As I wrote in another comment clicking on something should provide a feedback to user. We mustn't restart the app for this.

src/fheroes2/dialog/dialog_graphics_settings.cpp
ihhub: This **should** be done during runtime like as we do in many other place. Just use `fheroes2::Copy()` function for this matter.

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: @ihhub I Added `UpdateUI()` in the latest Review Update

MaMadDl (commented):
Hi @ihhub 
display change function now works in runtime (without the need to restart)
and i have changed the layout to 5 button layout like so :

Screenshot 2023-12-14 001522
Screenshot 2023-12-14 001450

Could you please review them?

ihhub:
Hi @MaMadDl , I've been thinking about the use case of this option. Usually, player just drag a window of any application between multiple monitors without going to specific settings to do some magic. I think saving display ID and the position of a window is good enough. Giving an option to somehow switch between monitors seems to me unnatural. At least I haven't seen such functionality in any software.

MaMadDl:
Hi @ihhub, I've seen this option in almost all of modern game engines (one example is COD Warzone or simpler game like Rimworld). of course this could be done using Windowed Mode. but I personally like trapping cursor to bounding of the game (which windowed mode can't provide). to me not having this option cause a lot of frustration for me (like in all _From Software_ Games) and feels unnatural. as I've said in before, I required this feature to play so I added and thought in case others also needed this feature should be present.

ihhub:
Hi @MaMadDl , that's a fair point. Thanks for the explanation!

ihhub (changes requested):
Hi @MaMadDl , I left few comments here. I will review deeper in the future. Could you please take a look at them when you have time?

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -224,6 +249,7 @@ namespace fheroes2
     void openGraphicsSettingsDialog( const std::function<void()> & updateUI )
     {
         Settings & conf = Settings::Get();
+        fheroes2::Display & display = fheroes2::Display::instance();
```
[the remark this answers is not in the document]
MaMadDl: Hello @ihhub 
this line is used in line 283 and it's required to have access to `changeDisplayEngine` function
I can make it inline while calling `changeDisplayEngine` .

MaMadDl (commented):
Hello @ihhub 
I have changed Reviewed  requested Please take a look.

src/engine/screen.cpp
```diff
@@ -1472,4 +1484,21 @@ namespace fheroes2
     {
         return *( Display::instance()._cursor );
     }
+
+    int BaseRenderEngine::getNumberOfVideoDisplays()
+    {
+        if ( int displayCount = SDL_GetNumVideoDisplays(); displayCount > 0 ) {
```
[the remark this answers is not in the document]
ihhub: clang-format is correct about this. if-condition must not have any variable declarations.

src/engine/screen.h
```diff
@@ -139,10 +139,25 @@ namespace fheroes2
...
+        {
+            return _displayId;
+        }
+
+        void setDisplayId( const int display )
+        {
+            _displayId = display;
+        }
+
+        int getNumberOfVideoDisplays();
```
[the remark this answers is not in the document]
ihhub: clang-tidy is correct about this as well.

src/engine/screen.cpp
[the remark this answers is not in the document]
MaMadDl: This is if-initializer Style introduced in C++17

src/engine/screen.h
```diff
@@ -139,10 +139,25 @@ namespace fheroes2
             return _nearestScaling;
         }
 
+        int getDisplayId() const
```
ihhub, line 142: After further thinking: this class should have only 5 methods, all are virtual methods (notice I changed ID to index as we handled indicies):
```cpp
virtual uint8_t getCurrentDisplayIndex() const
{
    return 0;
}

virtual void setDisplayIndex( const uint8_t )
{
    // Do nothing.
}

virtual uint8_t getMaximumDisplays() const
{
    return 1;
}

virtual const char * getDisplayName( const uin8_t )
{
    return "";
}

protected:
...
virtual bool isAllocated() const
{
    return false;
}
```

There is no need for an extra member of a class as it should not be used for other technologies expect SDL 2.

Then inside `Display::setResolution();` method add a check if `!_engine->isAllocated()` then skip resolution check. This will avoid code duplication present in **changeDisplayEngine()**.

Then for SDL2 Render Engine class add a member `_displayIndex` and within `setDisplayIndex()` method call **clear** method.

Upon changing the display index the code would look like this:
```cpp
fheroes2::BaseRenderEngine & engine = fheroes2::engine(); // don't forget to cache it!
engine.setDisplayIndex( ( engine.getCurrentDisplayIndex() + 1 ) % fheroes2::engine().getMaximumDisplays() );

const fheroes2::ResolutionInfo currentResolution{ display.width(), display.height(), display.screenSize().width, display.screenSize().height };

[display](fheroes2::Display::instance()).setResolution( selectedResolution );
```

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -126,9 +137,14 @@ namespace
...
         const Settings & conf = Settings::Get();
         const bool isEvilInterface = conf.isEvilInterfaceEnabled();
-        const fheroes2::Sprite & dialog = fheroes2::AGG::GetICN( ( isEvilInterface ? ICN::ESPANBKG_EVIL : ICN::ESPANBKG ), 0 );
+        const fheroes2::Sprite & dialog = fheroes2::AGG::GetICN( ( isEvilInterface ? ICN::CSPANBKE : ICN::CSPANBKG ), 0 );
         const fheroes2::Sprite & dialogShadow = fheroes2::AGG::GetICN( ( isEvilInterface ? ICN::CSPANBKE : ICN::CSPANBKG ), 1 );
 
+        fheroes2::Sprite placeholder = fheroes2::AGG::GetICN( Settings::Get().isEvilInterfaceEnabled() ? ICN::SURDRBKE : ICN::SURDRBKG, 0 );
+        // the image taken from other big ICN and cropping it's background with adequate size
+        // the offsets are staticly fixed based on original image
+        placeholder = fheroes2::Crop( placeholder, 50, 30, dialog.width() - BORDERWIDTH * 2 - 20, optionWindowSize + 20 );
```
ihhub, line 146: Right now what we do:
- get an image, create another image and copy it
- Crop a part of an image and replace newly created image
- use this cropped image to draw stuff once
- keep the cropped image for the duration of this dialog life

We should do everything by using the original image (`fheroes2::AGG::GetICN()`  returns const reference) and Blit / render it using the same coordinates you used in `Crop()` function call.

ihhub (changes requested):
Hi @MaMadDl , I proposed another solution in the comments. Could you please take a look?

MaMadDl (commented):
Hello @ihhub 
I have made Changes based on the reviews. could you please take look

src/fheroes2/dialog/dialog_graphics_settings.cpp
MaMadDl: You mean removing Crop and use another overload of Blit ?

src/engine/screen.h
MaMadDl: i have kept `engine ` and `display ` inside case (as I didn't know what you mean by caching it if not putting them before `while` loop)

also i have inverted the logic of `isAllocated` method since true should mean engine is allocated

ihhub (changes requested):
Hi @MaMadDl , I left several more comment here. Can you please take a look?

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -139,18 +153,26 @@ namespace
...
         const fheroes2::Rect windowSystemInfoRoi( systemInfoRoi + windowRoi.getPosition() );
+        const fheroes2::Rect windowSwitchDisplayRoi( switchDisplayRoi + windowRoi.getPosition() );
 
-        const auto drawOptions = [&windowResolutionRoi, &windowModeRoi, &windowVSyncRoi, &windowSystemInfoRoi]() {
+        const auto drawOptions = [&windowResolutionRoi, &windowModeRoi, &windowVSyncRoi, &windowSystemInfoRoi, &windowSwitchDisplayRoi]() {
             drawResolution( windowResolutionRoi );
             drawMode( windowModeRoi );
             drawVSync( windowVSyncRoi );
             drawSystemInfo( windowSystemInfoRoi );
+            drawDisplay( windowSwitchDisplayRoi );
```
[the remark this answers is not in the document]
MaMadDl: @ihhub I think it should remain even for Single Monitor. Clicking will have no effect

MaMadDl (commented):
Hello @ihhub. I have made changes requested. Could you please take a look ?

src/fheroes2/dialog/dialog_graphics_settings.cpp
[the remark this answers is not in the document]
ihhub: Yes, and this is why this option must be disabled if no changes appear. If an option is enabled users always expect to be able to click on it and something to happen.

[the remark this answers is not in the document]
MaMadDl: I think the name and meaning of option is intuitive enough that users will understand that. it's just like `effect` or `music` in audio settings, although the option itself won't change, but you can see the effects of your click as increase to audio levels. if it was something that would happen occasionally, I would agree

there are some instances like this ( changing fullscreen has no beneficial effect, should we disable that too while we are in android ? )

src/fheroes2/dialog/dialog_graphics_settings.cpp
```diff
@@ -139,18 +153,25 @@ namespace
         fheroes2::Blit( dialogShadow, display, windowRoi.x - BORDERWIDTH, windowRoi.y + BORDERWIDTH );
         fheroes2::Blit( dialog, display, windowRoi.x, windowRoi.y );
 
+        // The Image is taken from another big ICN and cropped its background with adequate size
+        // The Offsets are staticly fixed based on original image to Fill the empty spaces on the second row alongside with it's shadow
```
ihhub, line 157: There is no need to make Image or Offsets to start with a capital letter, as well as Fill word.

src/fheroes2/dialog/dialog_graphics_settings.cpp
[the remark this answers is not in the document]
ihhub: We have a separate issue to disable some options on some platforms like text support on Android devices.

ihhub (changes requested):
Hi @MaMadDl , I left few more minor comments here. We are almost there. Could you please take a look at them?

src/fheroes2/dialog/dialog_graphics_settings.cpp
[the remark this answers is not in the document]
MaMadDl: Overall I'm not against adding functionality when there's only 1 display. but disabling doesn't sounds like a good route since it will already is disabled and we don't have disabled ICN for `GAME_OPTION_ICON` ( like how VSync has)
Maybe we can Show a dialog if there's only 1 display connected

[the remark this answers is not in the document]
ihhub: We have a way to make any object disabled: please use `fheroes2::ApplyPalette( yourImage, PAL::GetPalette( PAL::PaletteType::DARKENING ) );` This is how all buttons or any other UI objects are rendered in a disabled state.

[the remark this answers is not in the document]
MaMadDl: Do you want this added now ?

MaMadDl:
Hi, @zenseii i have not been able to reproduce Problem 1 . but I suspect it has something to do when we check for engine allocation in line `#1202 of screen.cpp` 

as for the Problem 2, I think we should check if resolution is supported by the current display. 
in allocate method there's a way this is done using `GetNearestResolution` but the problem is the `resolutions` the method is checking against is calculated for first display (`getAvailableResolutions` method checking for display 0 `line #859`).  

@ihhub what do you think ?

MaMadDl:
Hi @zenseii 

could you get the higher resolution before the change ? sadly both my displays are 1920x1080 so I cannot test this myself. 

but as far as I could tell `getAvailableResolutions` was always checking for available display resolutions for first display. 
I have changed this to calculate list of resolutions based on the display the window is in.
After a quick google search it appears that SDL has issues with HIDPI on window for 4k displays but I'm not an expert in SDL.

the error is caused by calling to `SDL_GetNumVideoDisplays()` in `getMaximumDisplays()` before engine is allocated. it could be avoided by checking for engine allocation (still testing).

MaMadDl:
@zenseii 
changes are already commited in [0287e22](https://github.com/ihhub/fheroes2/pull/8072/commits/0287e22ecbc30188946a6c91b590743dc1972599)

MaMadDl:
Hi @zenseii 

the error on startup is fixed.

also for the 4K resolution I think I found the problem :
in `getAvailableResolutions()` the lambda function is calculated at the startup once (with default display index 0) and after that every time its called we get the result from startup 
`line #855`:
`static const std::vector<fheroes2::ResolutionInfo> filteredResolutions = [=]() {`
can you check for your resolutions after removing `static const` 
like `std::vector<fheroes2::ResolutionInfo> filteredResolutions = [=]() {`

MaMadDl:
Hi @zenseii 

the error happens because SDL trying to render on a display that does not exists. 

because this could happen in any stage of the game and not just in options
the only solution I could think of is using `SDL_AddEventWatch` on `SDL_DisplayEvent` and re-allocate the engine in case the display goes out of bound 

let me know what you think.

MaMadDl:
> Meanwhile, I found another bug. When you have your screens set up to extend the primary screen, and end fheroes2 while it is being displayed on what is the second screen, then if you change to single screen mode, which means you're only using the 2nd screen or the first screen, then the App will throw an exception in `xstring.cpp line 401` when clicking on the Graphics option in settings:

Hi @zenseii 
to fix this I have added `SDL_DISPLAYEVENT` to localevent polling. in any stage of the game a display disconnect this event is called and allocating last available display in case current display is out of bound.



> I'm also getting this message in the console upon booting the game:
> 
> ```
> fheroes2 engine, version: 1.0.11
> 13.02.2024 16:37:32: [ERROR]    `anonymous-namespace'::RenderEngine::getMaximumDisplays:  Failed to Get Number of Displays, description: Video subsystem has not been initialized
> ```

Also this error is handled.

MaMadDl:
Hi @zenseii 
I'm unable to reproduce this error. can you check the conf file to see if resolution is saved or changed?
I haven't changed how resolution is saved or even works. (it's in line `#261` in `dialog_graphics_settings.cpp`)

ihhub:
I am closing this pull request as it isn't ready for merging and it wasn't changed for more than a year. Also, #9853 will solve the original problem of having multiple monitors and remembering which one was used.

PR #8076, opened 2023-11-25
Allow application closure by close button if no resources available

ihhub:
Based on the feedback from social media.


Branikolog (approved):
I wish we can close the application on any game stage, not only on "no resources", "main menu" and "adventure map" screens. ;)

PR #8077, opened 2023-11-25
Rework Battle Only mode

### LINKED ISSUE #7876: Re-running Battle Only should give you the same setup
### Preliminary checks

A common use case for Battle Only is testing the same setup with minor variations or changes in tactics. Most of the time a player selects that option they will want to try the same thing again.

The default behavior should be to reload the last battle. Next to the "Okay" button could be "Reset" button to clear out all changes and revert to the default loadout.

### Additional info

_No response_

ihhub:
- make Battle Only battlefield with random terrain, not only Water
- save the state of Battle Only after a battle. It can be reset by RESET button
- add CANCEL button to exit the window

Also reduce the number of tiles being generated for this mode.

image

close #7876


Branikolog (approved):

Districh-ru (changes requested):
Hi, @ihhub. The changes to Battle Only mode works properly, well done!
I just left three comments to the code. Could you please take a look when you have time.

src/fheroes2/battle/battle_only.h
```diff
@@ -66,47 +74,67 @@ namespace Battle
...
+            void reset();
+        };
+
+        std::array<ArmyInfo, 2> armyInfo;
 
         std::unique_ptr<ControlInfo> cinfo2;
 
-        fheroes2::Rect rtPortrait1;
-        fheroes2::Rect rtPortrait2;
+        bool _backupCompleted;
```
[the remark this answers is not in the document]
ihhub: Yes! I missed it.

Districh-ru (approved):

oleg-derevenetz (approved):
LGTM. By the way, to diversify the background in battles, you can sometimes plant trees on free cells on those types of terrain on which it is possible. But this is just an optional cosmetic effect :)

src/fheroes2/game/game_startgame.cpp
```diff
@@ -132,10 +132,26 @@ namespace
...
+    world.generateBattleOnlyMap();
+
+    bool reset = false;
+    bool allowBackup = true;
+
+    while ( battleOnlySetup.setup( allowBackup, reset ) ) {
+        allowBackup = false;
+
+        if ( reset ) {
+            world.generateBattleOnlyMap();
```
oleg-derevenetz, line 146: Is there any reason to regenerate the map when the Reset button is pressed?

ihhub: Hi @oleg-derevenetz , we need to reset all stats for heroes which a user might change within this dialog.

ihhub:
Hi @oleg-derevenetz , yes, we can improve it later. Here are more things that we can do for this dialog.

PR #8081, opened 2023-11-25
Use small font for labels, increase dialog height in the Campaign Difficulty selection dialog

roman-yagodin:
Using small font for difficulty labels resolve width issues. Note that difficulty selection part of the scenario dialog also use small font for same labels.

game_campaign_en

Seems like Norwegian translation were updated and currently does not have mentioned issues with incufficent dialog height:

game_campaign_nb

But still I've increased dialog height to resolve similar issue with Bulgarian translation (easy diffuculty description):

game_campaign_bg

The only remaining issue is the Bulgarian translation of dialog title slightly overlaps internal dialog borders - but this one not looks that critical to me.

### DIFF
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -888,7 +888,7 @@ namespace
 
     int32_t setCampaignDifficulty( int32_t currentDifficulty, const int32_t maximumAllowedDifficulty )
     {
-        const fheroes2::StandardWindow frameborder( 234, 270, true );
+        const fheroes2::StandardWindow frameborder( 234, 312, true );
         const fheroes2::Rect & windowRoi = frameborder.activeArea();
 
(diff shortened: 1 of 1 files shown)


ihhub:
We should increase the width of this window and move these 3 difficulty options apart without reducing the font. These 3 options are copied as a single image from the original map description image. What we need to do is to each options as a separate image separately and put them further apart. As I can see even the title of some languages does not fit into.

roman-yagodin:
I've started from just that (splitting difficulty sprite in order to increase dialog width), but then falled back to simplier solution.

You want to preserve font size for difficulty labels for more consistency with the original game?

ihhub:
Hi @roman-yagodin , yes. Imagine having tiny dialog on phones. The dialog can be much bigger for sure. We should fall back to a smaller if we are limited in dialog size.

roman-yagodin:
Closing in favor of #8111

PR #8088, opened 2023-11-27
Fix update of the morale/luck indicator when changing the hero in the Battle Only mode

### LINKED ISSUE #8086: Battle Only Mode: Morale/Luck indicator area not cleared when changing hero
### Preliminary checks

### Platform

Other

### Describe the bug

As seen in the image the icon is drawn on top of the old one.

Reproduce it by changing hero. Same on both hero sides. Also happens with the luck icon.

Screenshot_20231127_054041_fheroes2

### Save file

Just open the battle only mode.

### Additional info

_No response_

oleg-derevenetz:
fix #8086

Perhaps it would be possible to do this using RAII, but luck/morale indicators usually do not have to restore their background when they are destroyed, so I decided to implement an ad-hoc restoration.

Also this PR fixes excessive FPS when hovering the mouse cursor over the primary/secondary skill panels on the Battle Only screen.

### DIFF
--- a/src/engine/image.h
+++ b/src/engine/image.h
@@ -147,7 +147,7 @@ namespace fheroes2
 
         ImageRestorer( const ImageRestorer & ) = delete;
 
-        // Restorers the original image if necessary, see the implementation for details
+        // Restores the original image if necessary, see the implementation for details
         ~ImageRestorer();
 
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:

PR #8096, opened 2023-12-01
Add basic logic to load and save maps through UI

ihhub:
- add a new map format entry - RESURRECTION and make the corresponding logic in the code
- rename some Settings class related methods
- add basic logic to load and save new maps
- expand new map format
- make code cleanup in `Maps::FileInfo` class

relates to #6845

### DIFF
--- a/src/fheroes2/dialog/dialog_selectscenario.cpp
+++ b/src/fheroes2/dialog/dialog_selectscenario.cpp
@@ -656,8 +656,8 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps 
         else if ( le.MousePressRight( sizeMaps ) || le.MousePressRight( curMapSize ) )
             ShowToolTip( _( "Size Icon" ), _( "Indicates whether the map\nis small (36 x 36), medium\n(72 x 72), large (108 x 108),\nor extra large (144 x 144)." ) );
         else if ( le.MousePressRight( mapTypes ) || le.MousePressRight( curMapType ) )
-            ShowToolTip( _( "Map Type" ),
-                         _( "Indicates whether the map is made for \"The Succession Wars\", \"The Price of Loyalty\" or \"Resurrection\" version of the game." ) );
+            // TODO: update this tooltip once the Editor is out for public.
+            ShowToolTip( _( "Map Type" ), _( "Indicates whether the map is made for \"The Succession Wars\" or \"The Price of Loyalty\" version of the game." ) );
         else if ( le.MousePressRight( mapNames ) ) {
             const Maps::FileInfo * item = listbox.GetFromPosition( le.GetMouseCursor() );
             if ( item )
--- a/src/fheroes2/dialog/dialog_selectscenario.cpp
+++ b/src/fheroes2/dialog/dialog_selectscenario.cpp
@@ -111,6 +111,9 @@ namespace
         case GameVersion::PRICE_OF_LOYALTY:
             body.add( { _( "The Price of Loyalty" ), fheroes2::FontType::normalWhite() } );
             break;
+        case GameVersion::RESURRECTION:
+            body.add( { _( "Resurrection" ), fheroes2::FontType::normalWhite() } );
+            break;
         default:
             // Did you add a new map version? Add the logic above!
             assert( 0 );
@@ -184,8 +187,8 @@ namespace
     {
         const Settings & conf = Settings::Get();
 
-        const std::string & mapName = conf.CurrentFileInfo().name;
-        const std::string & mapFileName = System::GetBasename( conf.CurrentFileInfo().file );
+        const std::string & mapName = conf.getCurrentMapInfo().name;
+        const std::string & mapFileName = System::GetBasename( conf.getCurrentMapInfo().file );
         size_t mapId = 0;
         for ( MapsFileInfoList::const_iterator mapIter = lists.begin(); mapIter != lists.end(); ++mapIter, ++mapId ) {
             if ( ( mapIter->name == mapName ) && ( System::GetBasename( mapIter->file ) == mapFileName ) ) {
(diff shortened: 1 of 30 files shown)


src/fheroes2/maps/map_format_info.h
```diff
@@ -47,25 +47,52 @@ namespace Maps::Map_Format
         std::vector<ObjectInfo> objects;
     };
 
-    struct MapFormat
+    struct BaseMapFormat
```
oleg-derevenetz, line 50: Hi @ihhub could you please explain why there are two map formats - "base" and "extended"(?) and why there is separate function to load the "base" map format? In which cases is it planned to use the basic format, and in which cases - the extended one?

ihhub: Hi @oleg-derevenetz , the separation was done to handle 2 cases:
- load basic map information when we display a list of maps in the Editor and when we are going to display them for normal New Game. We should avoid reading unnecessary data due to performance issues on some old machines with HDDs (#6681)
- Extended format is used to load the map fully for the Editor and for the future game loading

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -285,7 +288,36 @@ namespace Interface
                     res = eventNewMap();
                 }
                 else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_SAVE_GAME ) ) {
-                    fheroes2::showStandardTextMessage( _( "Warning!" ), "The Map Editor is still in development. Save function is not implemented yet.", Dialog::OK );
+                    const std::string dataPath = System::GetDataDirectory( "fheroes2" );
```
Districh-ru, line 291: Why do we want to save map files to a hidden folder (... "AppData\Roaming" ...) instead of normal Maps folder located near the fheroes2 executable?
I was surprised when I didn't find just saved map in the Maps folder along with the SW and PoL maps.
IMHO al least we should allow user to choose where to save a map.

Also if we have map files with the same file name in both folders (normal and "hidden") "save map" will save to the "hidden" folder and "load map" will show and load file from the normal folder. So if I change a map from the normal maps folder and save it with the same file name I'll not be able to load the changed file.
This issue can be addressed later when we will implement a map save dialog.

src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -574,29 +591,25 @@ MapsFileInfoList Maps::PrepareMapsFileInfoList( const bool multi )
...
-    std::sort( result.begin(), result.end(), Maps::FileInfo::NameSorting );
+    std::sort( result.begin(), result.end(), sortByMapNames );
 
     return result;
 }
 
-MapsFileInfoList Maps::prepareResurrectionMapsFileInfoList()
+MapsFileInfoList Maps::getResurrectionMapFileInfos()
 {
     // TODO: set the proper resurrection map extension.
```
Districh-ru, line 601: This comment can be removed now. :)

Districh-ru (changes requested):
Hi, @ihhub, I left some comments. Could you please take a look when you have time?

src/fheroes2/editor/editor_interface.cpp
oleg-derevenetz: > Why do we want to save map files to a hidden folder (... "AppData\Roaming" ...) instead of normal Maps folder located near the fheroes2 executable?

Because normal MAPS folder may not be writable for the regular user, when app is installed to the Program Files on Windows, /usr on Linux or /opt on macOS. This is more correct approach.

src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -640,7 +656,8 @@ const Maps::FileInfo * Dialog::SelectScenario( const MapsFileInfoList & allMaps
         else if ( le.MousePressRight( sizeMaps ) || le.MousePressRight( curMapSize ) )
             ShowToolTip( _( "Size Icon" ), _( "Indicates whether the map\nis small (36 x 36), medium\n(72 x 72), large (108 x 108),\nor extra large (144 x 144)." ) );
         else if ( le.MousePressRight( mapTypes ) || le.MousePressRight( curMapType ) )
-            ShowToolTip( _( "Map Type" ), _( "Indicates whether the map is made for \"The Succession Wars\" or \"The Price of Loyalty\" version of the game." ) );
+            ShowToolTip( _( "Map Type" ),
+                         _( "Indicates whether the map is made for \"The Succession Wars\", \"The Price of Loyalty\" or \"Resurrection\" version of the game." ) );
```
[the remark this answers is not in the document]
ihhub: Good point. I will revert this change and add a TODO comment.

src/fheroes2/editor/editor_mainmenu.cpp
```diff
@@ -307,7 +309,21 @@ namespace Editor
...
+                return fheroes2::GameMode::CANCEL;
+            }
+
+            fheroes2::fadeOutDisplay();
+            Game::setDisplayFadeIn();
+
+            return Interface::EditorInterface::Get().startEdit();
         }
         else {
             showWIPInfo();
```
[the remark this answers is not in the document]
ihhub: Yup.

Branikolog:
Hi, @ihhub .
Could you please, check if everything goes well in a few cases I found.

First of all I've noticed, that on map loading there's a short screen flickering:

https://github.com/ihhub/fheroes2/assets/55348946/089a3b21-19e5-40b4-8a5d-c36f338f2f40

As for the second, we can change capitalized symbols while saving a map.
Do we need some confirmation dialog or warning message, that the map name is already exist and the map is going to be overwritten?
I'm afraid, that in future there could be a mess with map names.

https://github.com/ihhub/fheroes2/assets/55348946/5b6b45de-4d6d-4f6c-865f-62cf509b6286

ihhub:
Hi @Branikolog , short flickering was even before. My changes do not introduce it.

 File overriding confirmation is not implemented in this pull request. We will add it later. I attempted to make minimal changes to unblock the rest of work.

Branikolog:
@ihhub 
Okay. I just wanted to be sure you're aware of these issues.

One more question:

image

I've noticed, that we have a "Resurrection" type map even in the description. But all maps are showing "Succession wars map". Is this description differentiation is not implemented yet?

Districh-ru (approved):

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Lets choose for now the easiest way that will work on all systems. And later, after Editor is released we may update map save/load to allow use of different paths.
Maybe we should not name the folder "maps" not to confuse with the folder "maps" with the original maps. Maybe something like "fh2maps". Then all users will know that this is a place for maps made by fheroes2 Editor.
What do you think, @ihhub?

oleg-derevenetz:
Hi @ihhub

> I noticed that clang-tidy complained about something different today. Looks like it was updated by its authors.

Yeah, by me - https://github.com/platisd/clang-tidy-pr-comments/pull/57 :)

ihhub:
> Hi @ihhub
> 
> > I noticed that clang-tidy complained about something different today. Looks like it was updated by its authors.
> 
> Yeah, by me - [platisd/clang-tidy-pr-comments#57](https://github.com/platisd/clang-tidy-pr-comments/pull/57) :)

But the new rule wasn't added by you :)

Issue #7933, opened 2023-10-12
support Chinese??


StrongZhu:
### Your question

hi, 
  does this version support Chinese??
  As we known that there are Chinese of Heroes2. no only for game GUI, but also for the video, 
  can find video in here  :   https://www.bilibili.com/video/BV14y4y1p7L6/
  
  I tried to copied some maps of Chinese version to the folder, then started the game,  it show "????" for all Chinese chars.

  checked the source code quickly, seems that it load font from resource, then run 'Blt' to generate bitmap for char....
  NOT sure SDL support Chinese char or not

ihhub:
Hi @StrongZhu , as we know there were no official translation of Heroes of Might and Magic 2 for Chinese (Mandarin) by modifying an English version of the game on Windows. Also the engine does not support any UTF-8 characters at the moment.

Also we don't use SDL for image generation.

ihhub:
@StrongZhu , on top of this to support Chinese there should be translation to Chinese. You cannot just grab a pirated version from the internet made for Windows and it would automagically work for fheroes2. Are you willing to provide full translation of the game for Chinese?

ihhub:
@StrongZhu , I had to remove your link. All piracy content is prohibited.

none2003:
> Hi [@StrongZhu](https://github.com/StrongZhu) , as we know there were no official translation of Heroes of Might and Magic 2 for Chinese (Mandarin) by modifying an English version of the game on Windows. Also the engine does not support any UTF-8 characters at the moment.
> 
> Also we don't use SDL for image generation.

I would like to kindly clarify that, based on my research, around 1997-1998, 3DO entrusted Ubisoft's Shanghai branch with distributing Heroes II (including its expansion) in Mainland China, [wiki](https://zh.wikipedia.org/wiki/%E9%AD%94%E6%B3%95%E9%97%A8%E4%B9%8B%E8%8B%B1%E9%9B%84%E6%97%A0%E6%95%8C%E7%B3%BB%E5%88%97). This version featured an exceptionally thorough localization by Ubisoft Shanghai, encompassing GUI adaptation, translated story texts, localized video files, and voiceovers.

I sincerely hope that fheroes2 will consider supporting non-Latin character systems, such as CJK languages.

ihhub:
Hi @none2003 

This is interesting. Is there an official store that sells a digital Mandarin version of the game?

none2003:
[...]

Issue #6207, opened 2022-11-18
Typing player name with supported languages


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

![image](https://user-images.githubusercontent.com/55348946/202706833-5fda754c-576c-4a21-8bab-f7e43d2efac8.png)
Only English is used while creating player's name.

### Describe the possible solution

Since fheroes2 supports multiple languages, it would be nice, if players could type their names using the language supported by the engine.

zenseii:
Just adding that this will probably only be possible once we add UTF-8 encoding support. There should be an issue for that.

undef21:
I think sphinx's riddle dialog has the same issue. Some localized maps can have non-ASCII answers.
AFAIK windows versions of OG has support of national encodings but dos versions (e.g. French) hasn't. In dos versions the game filtered user input and allowed only ASCII chars.

zenseii:
Basically this is relevant for any place where the user is asked to type, so even writing save game names.

667bdrm:
On this map sphinx dialog allow enter only digits. English or Russian letters typing does not change the field.

Ubuntu 20.04.5 LTS

![image](https://user-images.githubusercontent.com/416680/219967853-2a6b14bc-10dd-4e20-a5d3-e82ce0905087.png)

[Горы Альтерак.zip](https://github.com/ihhub/fheroes2/files/10777634/default.zip)

ihhub:
This issue was partially fixed by introduction of the Virtual Keyboard.

PR #8098, opened 2023-12-02
Add initial support for UTF-8 text input

ihhub:
relates to #7933, #6845 and #6207

### DIFF
--- a/VisualStudio/fheroes2/sources.props
+++ b/VisualStudio/fheroes2/sources.props
@@ -3,6 +3,7 @@
   <ItemGroup>
     <ClCompile Include="src\engine\agg_file.cpp" />
     <ClCompile Include="src\engine\audio.cpp" />
+    <ClCompile Include="src\engine\char_encoding.cpp" />
     <ClCompile Include="src\engine\core.cpp" />
     <ClCompile Include="src\engine\dir.cpp" />
     <ClCompile Include="src\engine\h2d_file.cpp" />
@@ -238,6 +239,7 @@
   <ItemGroup>
     <ClInclude Include="src\engine\agg_file.h" />
     <ClInclude Include="src\engine\audio.h" />
+    <ClInclude Include="src\engine\char_encoding.h" />
     <ClInclude Include="src\engine\core.h" />
     <ClInclude Include="src\engine\dir.h" />
     <ClInclude Include="src\engine\endian_h2.h" />
--- a/src/engine/char_encoding.cpp
+++ b/src/engine/char_encoding.cpp
@@ -1,0 +1,133 @@
+/***************************************************************************
+ *   fheroes2: https://github.com/ihhub/fheroes2                           *
+ *   Copyright (C) 2023                                                    *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ *   This program is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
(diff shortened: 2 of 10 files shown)


Branikolog:
Hi, @ihhub !
Russian and Belarussian works well.
But when I select English as game language and try to add some symbols, while my system input language is set as Russian or Belarussian - nothing happens. In master branch I can still type, no matter what language input was chosen. We should have an opened issue for MAC users only in issues section. With this PR the issue is valid for Win too now.

oleg-derevenetz:
Hi @Branikolog

> But when I select English as game language and try to add some symbols, while my system input language is set as Russian or Belarussian - nothing happens.

When you select English as game language, no Russian or Belarussian fonts exist, so when Russian or Belarussian symbols are coming from the keyboard, they are ignored.

Branikolog:
Hello, @oleg-derevenetz .

>In master branch UTF-8 does not come from the keyboard.

I'm just speaking as a common user. :)
For me personally current implementation for English is less convenient, than the one from master branch. 

If the game language is set as English, it is implied, that no other languages could be used to type texts. So I expect only Latin symbols to appear. In that case I feel there's no need to select exactly English within OS to type, since no other symbols are allowed. We've already had some complains on MAC, as I mentioned above, that game forces user to switch OS input language to be able to add text. 
To sum up, I suggest for "English" game language add always Latin symbols, no matter what text input was set within OS. If it's possible, for sure.
For such languages as Russian and Belarussian everything works fine for me in this PR.

oleg-derevenetz:
Hi @ihhub there are my observations:

1. This doesn't work on Linux. No "key press" event is generated for non-English languages, thus, `if ( le.KeyPress() ) {` checks doesn't work;
2. On Android, when `SDL_StartTextInput()` is called, the Android native virtual keyboard appears on the screen. But you still cannot enter characters using this virtual keyboard (even English ones!) because "key press" events are not generated either.

ihhub:
Hi @oleg-derevenetz , thanks for the notes! I will try to test on Android and on Linux. Most likely these OSes behave differently from Windows.

oleg-derevenetz:
Hi @ihhub

> I will try to test on Android and on Linux. Most likely these OSes behave differently from Windows.

Also please note that, although I couldn't verify it, there will be issues with national symbols in file names on Linux. Although Linux filesystems (such as EXT3) are encoding-agnostic (i.e. they treat file names just as an array of bytes), particular apps (e.g. shell) will treat file names according to their locale settings (set via `LC_*` environment variables) and most likely they won't understand CP1251 file names, because most likely their locales are set to UTF-8.

ihhub:
Hi @oleg-derevenetz , I believe we shouldn't allow to enter non-English names for files at the moment.

Branikolog:
@oleg-derevenetz and @ihhub 

> @Branikolog
> 
> Please see my notes above. If no key information comes for non-English symbols (only their UTF-8 symbol codes) there is no way to match these UTF-8 symbols to corresponding English symbols on the same keyboard keys (i.e. `й`->`q`, `ц`->`w`) unless you have an information about keyboard layout, and the ways to obtain this information are highly system-dependent (and maybe even flavor-dependent in case of Linux, because some flavors use X Window and some use Wayland), therefore, it's hardly worth it.

I actually don't know how exactly things happen in the code regarding to symbol codes. 
I just express my opinion as a potential user of fheroes2 engine.
If things I was talking above are not possible to implement by creating a proper logic or using a combination of new code for non-English languages selected and the old code related to text input for English game language only, then I can approve this PR, as It brings more benefits, than losses. 💯 
But I believe, there could be complains from Windows users on forcing them to switch OS language input, same as I've already had from MAC users.

Districh-ru:
Hi, @ihhub. The CP1251 letters works good for me (Win10), but I cannot input some symbols:
изображение

ihhub:
Hi @Branikolog , what was before doesn't mean it was correct. We will see what we could do about this.

@Districh-ru , most likely it is because out CP-1251 font does not contain all symbols.

ihhub:
Closing this pull request as a part of source code was integrated into the engine and another part is not a correct solution.

PR #8104, opened 2023-12-04
Fix a crash with castle overlapping inside a map

### LINKED ISSUE #5974: Map loading problem
### Preliminary checks

### Platform

Windows

### Describe the bug

There's an assertion while loading a particular map.
image
If we skip (ignore) three times the game starts, but in a such way:
image

I suppose, there's a problem with castles, placed in such manner. Original game load this map without a problem, but these castles are unreachable. Also, heroes hired there can't leave the sell in front of castle gates.
image

### Save file

[Map_0072.zip](https://github.com/ihhub/fheroes2/files/9784191/Map_0072.zip)

### Additional info

_No response_

ihhub:
close #5974

### DIFF
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -2796,8 +2796,8 @@ void AllCastles::AddCastle( Castle * castle )
     const size_t id = _castles.size() - 1;
     const fheroes2::Point & center = castle->GetCenter();
 
-    // We need to override any existing castle's ID that is why we use [] operator to access std::map.
-    // Castles are added from top to bottom, from left to right so a newer castle must override existing data for a tile if any.
+    // Castles are added from top to bottom, from left to right.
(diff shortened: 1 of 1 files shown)


src/fheroes2/castle/castle.cpp
```diff
@@ -2806,11 +2806,17 @@ void AllCastles::AddCastle( Castle * castle )
                 continue;
             }
 
-            _castleTiles[center + fheroes2::Point( x, y )] = id;
+            const auto [dummy, inserted] = _castleTiles.try_emplace( center + fheroes2::Point( x, y ), id );
```
oleg-derevenetz, line 2809: Hi @ihhub I believe that comment before this `for` loop is not valid anymore :) I suppose we need a new comment that applies to the current logic.

oleg-derevenetz (approved):

PR #8118, opened 2023-12-10
Battle pathfinder: replace std::bitset by std::array<bool>

oleg-derevenetz:
With MSVC, operations with `std::bitset::reference` are surprisingly slow, for example:

```
Function Name			Total CPU [unit, %]	Self CPU [unit, %]	Module
std::bitset<99>::_Set_unchecked	102 (1,83%)		88 (1,58%)		fheroes2.exe
```

88 milliseconds (total during the whole battle) is a bit too much for such an operation (it's the 3rd place in terms of the "Self CPU" indicator during the instant battle due to the MSVC profiler!), even though it was a relatively long battle. Unfortunately, MSVC doesn't show the timings for the built-in `operator=()` for `bool`, because it's a basic single-instruction operation, but I suppose that it shouldn't take so much time.

Of course, `std::bitset<99>` occupies much less space (16 bytes on 64-bit system instead of 99 bytes), but this difference is actually insignificant in this case.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this optimization!

PR #8127, opened 2023-12-12
Update the French translation, add missing offset for an A with accent

zenseii:
Small fixes here and there that should improve the experience.


ihhub (approved):

ihhub:
@zenseii , a lot of thanks for this update!

PR #8134, opened 2023-12-13
StreamFile: improve I/O error handling

### LINKED ISSUE #8133: Appropriate error message when saving a game but disk is full
### Preliminary checks

If you try to save your current game when there's no more space in the disk, then you'll be prompted that save was successful. At this point there are two options:

1. There's literally no space, then no save will be created, but
2. if there was just enough to create a file, then the file will be corrupted and nr unusable

Give an error message when saving the game unsuccessfully

### Additional info

_No response_

oleg-derevenetz:
close #8133


oleg-derevenetz:
Hi @Giovarco could you please test this pull request in your environment in regard to #8133? Now you should see the message "There was an issue during saving" (or a translated version into your language) if there is no space on disk. Just in case, here is an [instruction](https://github.com/ihhub/fheroes2/wiki/Download-Windows-latest-compiled-packages-of-a-pull-request) on how to download the pull request binaries for Windows created by our CI.

oleg-derevenetz:
Hi @ihhub although this PR is ready for review, I suggest to postpone its merge after the 1.0.11 release, because it seems that currently most of our team is busy cleaning up usual end-of-the-year mess at their jobs and there is a danger of missing some issue in the code that is used in so many places.

oleg-derevenetz:
Hi @Giovarco

> Unfortunately I cannot test this as I don't have access to my PC. Maybe someone else is willing to do it?

Well, I tested this myself at the time, but it would be good to have independent testing by someone else. Thank you anyway.

src/engine/serialize.h
```diff
@@ -219,18 +217,26 @@ class StreamBase
...
 
     void setconstbuf( bool );
     void setfail( bool );
+
+private:
+    enum : uint32_t
+    {
+        FailureFlag = 0x00000001,
+        ConstBufFlag = 0x00000002,
+        BigendianFlag = 0x00000004
```
ihhub, line 236: Should we name them using CAPITAL letters as you pointed out in the past? :)

ihhub (approved):

src/engine/serialize.h
oleg-derevenetz: Hi @ihhub currently we have two styles: `ALL_CAPITAL` like here (mostly in the old code):

https://github.com/ihhub/fheroes2/blob/7084547aac6546c859e60fb72592bb1ea3fec60b/src/fheroes2/heroes/heroes.h#L73-L87

and `PascalCase` like here (mostly in the new code):

https://github.com/ihhub/fheroes2/blob/7084547aac6546c859e60fb72592bb1ea3fec60b/src/fheroes2/army/army.h#L140-L148

So I used `PascalCase` here, because it's a new code (sort of). But I can change the style if necessary.

ihhub: We just need to agree which way to use and stick to it. From my experience enumeration items are usually written in `ALL_CAPITAL`. However, this is my personal experience based on several companies.

oleg-derevenetz: OK, let's stick to `ALL_CAPITAL` style here. At least I can now remove that `Flag` suffix :)

ihhub: Good job!

ihhub:
@oleg-derevenetz  thank you so much for these additions!

PR #8147, opened 2023-12-17
Add Tree of Knowledge claimed and cost info to quick info dialog

zenseii:
image

image

image

### DIFF
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -275,7 +275,7 @@ namespace
             }
             else {
                 const auto rc = payment.getFirstValidResource();
-                str.append( std::to_string( rc.second ) + " " );
+                str.append( std::to_string( rc.second ) );
                 str += ' ';
                 str.append( Translation::StringLower( Resource::String( rc.first ) ) );
             }
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -266,6 +266,7 @@ namespace
     {
         const MP2::MapObjectType objectType = tile.GetObject( false );
         std::string str = MP2::StringObject( objectType );
+        const Heroes * hero = Interface::GetFocusHeroes();
 
         if ( isVisited ) {
             const Funds & payment = getTreeOfKnowledgeRequirement( tile );
@@ -281,10 +282,17 @@ namespace
             }
             str += ')';
 
-            const Heroes * hero = Interface::GetFocusHeroes();
             if ( hero ) {
-                str += '\n';
-                str.append( hero->isVisited( tile ) ? _( "(already gained)" ) : _( "(not gained)" ) );
+                str.append( "\n(" );
+                str.append( hero->isVisited( tile ) ? _( "already claimed" ) : _( "not claimed" ) );
+                str += ')';
+            }
+        }
+        else {
+            if ( hero ) {
+                str.append( "\n\n(" );
+                str.append( _( "not claimed" ) );
+                str += ')';
             }
         }
 
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -262,6 +262,34 @@ namespace
         return str;
     }
 
+    std::string showTreeOfKnowledgeInfo( const Maps::Tiles & tile, const bool isVisited )
+    {
+        const MP2::MapObjectType objectType = tile.GetObject( false );
+        std::string str = MP2::StringObject( objectType );
+
+        if ( isVisited ) {
+            const Funds & payment = getTreeOfKnowledgeRequirement( tile );
+            str.append( "\n\n(" );
+            if ( payment.GetValidItemsCount() == 0 ) {
+                str.append( "free" );
+            }
+            else {
+                const auto rc = payment.getFirstValidResource();
+                str.append( std::to_string( rc.second ) + " " );
+                str.append( Translation::StringLower( Resource::String( rc.first ) ) );
+            }
+            str += ')';
+
+            const Heroes * hero = Interface::GetFocusHeroes();
+            if ( hero ) {
+                str.append( "\n" );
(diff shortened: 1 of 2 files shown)


src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -262,6 +262,34 @@ namespace
...
+                const auto rc = payment.getFirstValidResource();
+                str.append( std::to_string( rc.second ) + " " );
+                str.append( Translation::StringLower( Resource::String( rc.first ) ) );
+            }
+            str += ')';
+
+            const Heroes * hero = Interface::GetFocusHeroes();
+            if ( hero ) {
+                str.append( "\n" );
+                str.append( hero->isObjectTypeVisited( objectType ) ? _( "(already gained)" ) : _( "(not gained)" ) );
```
Districh-ru, line 286: We can not rely only on object type here: if you visit one Tree, then all other Trees will show "(already gained)" even if the level was not gained by that tree.
```suggestion
                str.append( hero->isVisited( tile ) ? _( "(already gained)" ) : _( "(not gained)" ) );
```

Districh-ru (changes requested):
Hi, @zenseii. I very like this improvement - currently when I have many heroes and there are many Trees of Knowledge on the map it is hard to remember what hero has visited what tree (I don't talk about remembering what resources each Tree demands).
I left one comment regarding getting "level gained" info. Could you please take a look when you have time.

Districh-ru (approved):

LeHerosInconnu:
Hello @zenseii,

I haven't tested this PR, but I think the expressions already used in other places, i.e. "(already visited)" and "(not visited)" should be used.
This would give continuity to the expressions and be more "natural" according to the hero selected.
Also, when no hero is selected, the right-click information window should also indicate the cost, but without the expression "(already visited)" or "(not visited)".
In addition, in the current state of a hero's experience point progression, it may happen that a hero gains several levels when visiting a tree of knowledge.
The use of the word "level" therefore becomes problematic in this case if it does not correspond to the result of the visit to the location, one or more levels gained.

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -262,6 +262,34 @@ namespace
...
+
+        if ( isVisited ) {
+            const Funds & payment = getTreeOfKnowledgeRequirement( tile );
+            str.append( "\n\n(" );
+            if ( payment.GetValidItemsCount() == 0 ) {
+                str.append( _( "treeOfKnowledge|free" ) );
+            }
+            else {
+                const auto rc = payment.getFirstValidResource();
+                str.append( std::to_string( rc.second ) + " " );
```
ihhub, line 278: Let's do this way:
```cpp
str.append( std::to_string( rc.second ) );
str += ' ';
```
This should be faster.

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -262,6 +262,34 @@ namespace
...
+                str.append( Translation::StringLower( Resource::String( rc.first ) ) );
+            }
+            str += ')';
+
+            const Heroes * hero = Interface::GetFocusHeroes();
+            if ( hero ) {
+                str.append( "\n" );
```
ihhub, line 285: Let's do this way:
```cpp
str += '\n';
```

ihhub (changes requested):
Hi @zenseii , I left 2 minor optimizations. Could you please take a look at them?

src/fheroes2/dialog/dialog_quickinfo.cpp
Districh-ru: Hi, @zenseii, in the last commit the `str += ' ';` was added, so we should remove this inserted space from the upper line :)
```suggestion
                str.append( std::to_string( rc.second ) );
```

Districh-ru (approved):

ihhub (approved):

ihhub:
@zenseii , thank you so much for the improvement!

LeHerosInconnu:
Hello @zenseii,

> @LeHerosInconnu
> 
> > I wouldn't call it laziness here, but a convention for places that give a bonus. The hero has not yet benefited from the location bonus: "(not visited)". The hero has already benefited from the location bonus: "(already visited)". This seems clear to me and in line with the other bonus locations. And that's how it works in the original game.
> 
> I don't see the benefit of such continuity, compared to the clear benefits of unambiguity. Also, we are already fine with **not** having 100 % continuity in for example the Shrines, which state `(already learned)` and **not** `(already visited)` : image
> 
> Addiontally, the Witch's huts' `(already knows this skill)` would have to be `(already visited)` then : image
> 
> If we seriously want to adopt this `(already visited)`, which **really** carries the meaning of "already visited and claimed whatever you can from", then we would have to change also the shrines, the witch's huts, etc., for continuity's sake.
> 
> To me having `(visited)` for the Tree of Knowledge makes as much sense as having `(talked to)` and `(not talked to)` for it.
> 
> On the contrary, I argue that variety in what words and expressions we use only enrichens the game and the experience, and self-imposed restrictions for the vocabulary of the game just so that a few words are used everywhere, is an unnecessary restriction with debatable benefits.

In the original game, for the shrine and witch's hut, the spell and secondary skill are not indicated in the right-click information window.
Also, these locations, unlike others such as the tree of knowledge, gazebo, mercenary camp, etc., do not systematically give a bonus when visited.
The hero can learn the spell or secondary skill in another way, making a visit to the place useless if he has already learned the spell or secondary skill.
A different, and added text compared to the original version of the game, can and should therefore be used.
For the tree of knowledge, the expressions "(not visited)" and "(already visited)" are already used in the original game, and I think they play their informative role well.

Also, for the witch's hut, I think the expression "(already learned)" could replace the expression "(already knows this skill)" to simplify the expression (we know it's a skill, and we don't indicate "(already knows this spell)" either).
And, for the shrine and witch's hut, when the hero doesn't know the spell or secondary skill, there's no expression to indicate this fact, only the name of the spell or secondary skill is indicated.
I think we should also display the expression "(not learned)" when a hero is selected and doesn't know the spell or secondary skill, as this is confusing with displaying the spell and secondary skill when no hero is selected.
I always tend to check that a hero is selected when I right-click on a shrine or witch's hut, which shouldn't be the case.
The interface in this case is not "uniform" in the way it indicates information to the player.
I can open a specific issue if necessary.

PR #8149, opened 2023-12-17
Don't interact with the transform layer if the image is a single layer

Districh-ru:
There is no need to set new data for the transform layer if the image is already in single layer mode - the transform layer will never be accessed.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -503,7 +503,10 @@ namespace fheroes2
         if ( !empty() ) {
             const size_t totalSize = static_cast<size_t>( _width ) * _height;
             memset( image(), value, totalSize );
-            memset( transform(), static_cast<uint8_t>( 0 ), totalSize );
+
+            if ( !_singleLayer ) {
+                memset( transform(), static_cast<uint8_t>( 0 ), totalSize );
+            }
         }
     }
 
@@ -532,8 +535,11 @@ namespace fheroes2
         if ( !empty() ) {
             const size_t totalSize = static_cast<size_t>( _width ) * _height;
             memset( image(), static_cast<uint8_t>( 0 ), totalSize );
-            // Set the transform layer to skip all data.
-            memset( transform(), static_cast<uint8_t>( 1 ), totalSize );
+
+            if ( !_singleLayer ) {
+                // Set the transform layer to skip all data.
+                memset( transform(), static_cast<uint8_t>( 1 ), totalSize );
+            }
         }
     }
 
@@ -556,7 +562,7 @@ namespace fheroes2
 
         _singleLayer = image._singleLayer;
 
-        memcpy( _data.get(), image._data.get(), size * 2 );
+        memcpy( _data.get(), image._data.get(), _singleLayer ? size : size * 2 );
     }
 
     Sprite::Sprite( const int32_t width_, const int32_t height_, const int32_t x_ /* = 0 */, const int32_t y_ /* = 0 */ )


ihhub:
Hi @Districh-ru , let's target this pull request for the next release. I think we might need to add assertions to verify this changes.

Districh-ru:
> Hi @Districh-ru , let's target this pull request for the next release. I think we might need to add assertions to verify this changes.

Hi, @ihhub, sure, let's postpone it to v. 1.1.0.
These changes in example should speed-up the radar rendering because it uses `fill()` to render the single layer image.

src/engine/image.cpp
```diff
@@ -503,7 +503,10 @@ namespace fheroes2
         if ( !empty() ) {
             const size_t totalSize = static_cast<size_t>( _width ) * _height;
             memset( image(), value, totalSize );
-            memset( transform(), static_cast<uint8_t>( 0 ), totalSize );
+
+            if ( !_singleLayer ) {
```
ihhub, line 507: This change is useful in terms of performance but we need to be sure that we are going to right thing. I suggest to add assertions into `transform()` method checking that an image is **not** a single layer image. After that we need to run the game to see that no code actually calls this function for a single layer image.

Districh-ru: Such assertions were added in #7233 and the only such access was found in `setResolution()` (setting transform layer to `1` to detect some UI artifacts by invalid code). I changed this code to `Image::reset();`.
In future PR I want to change images creation: if resize is called for single-layer image then there is no need to allocate memory for the unused transform layer (applies to image restorers and screen).

ihhub:
Hi @Districh-ru , I hit the assertion in Credits within `Resize()` function call.

Districh-ru:
> Hi @Districh-ru , I hit the assertion in Credits within `Resize()` function call.

Thanks, I'll check it. Previously I checked only the gameplay :)

ihhub:
> > Hi @Districh-ru , I hit the assertion in Credits within `Resize()` function call.
> 
> Thanks, I'll check it. Previously I checked only the gameplay :)

Hi @Districh-ru , I believe we will need to rewrite Resize() function for this matter. I recommend not to spend too much time at the moment doing this unless a fix is very trivial.

Districh-ru:
> > > Hi @Districh-ru , I hit the assertion in Credits within `Resize()` function call.
> > 
> > 
> > Thanks, I'll check it. Previously I checked only the gameplay :)
> 
> Hi @Districh-ru , I believe we will need to rewrite Resize() function for this matter. I recommend not to spend too much time at the moment doing this unless a fix is very trivial.

Yes a fix is to make `resizedOutput` single-layer in game_credits.cpp before calling `Resize()`.

But I plan to rewrite Resize today. :)

ihhub:
@Districh-ru , then cool :)

ihhub:
Hi @Districh-ru , it looks good but I will play around for a couple of days to make sure that no other places hit the assertion.

Districh-ru:
> Hi @Districh-ru , it looks good but I will play around for a couple of days to make sure that no other places hit the assertion.

Sure, I'll do the same.
I quickly checked **image.cpp** and all other functions there should work properly with single-layer images. `Resize()` is the only that needs a fix (and is fixed in this PR).

ihhub:
Hi @Districh-ru , I am worried that we might add crashes due to assertions and some undefined behavior on non-Windows platforms. Should we do these changes only for debug builds to be safe?

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for the changes!

PR #8154, opened 2023-12-20
Add text width-adapted CANCEL button, change to StandardWindow for Resolution, Language, Adventure, File and Scenario Info dialogs

zenseii:
This PR makes the single-lined CANCEL button adapt its width according to the text on it.

To make this possible it was necessary to switch to the StandardWindow class for the Language, Resolution, Adventure, File and Scenario Info dialogs. This also gives an evil interface mode for the first two mentioned dialogs plus the Scenario Info dialog.

The Uniform CANCEL button was adapted too and the Player name input dialog's keyboard button had to be moved to allow for that extra button width, since we don't have a way of widening that dialog yet.

The Scenario Info dialog's height was increased by a few pixels to avoid the overlapping of the handicap icons' shadow with the Okay and Cancel buttons.


src/fheroes2/dialog/dialog_language_selection.cpp
```diff
@@ -199,24 +250,29 @@ namespace
                 fheroes2::Text body( _( "Click to choose the selected language." ), fheroes2::FontType::normalWhite() );
                 fheroes2::showMessage( header, body, 0 );
             }
+            else if ( le.MousePressRight( listRoi ) ) {
+                continue;
+            }
```
Districh-ru, line 255: While holding the right mouse button (RMB) a user can press Up or Down key and select another language, but no dialog will be performed until RMB is released:

https://github.com/ihhub/fheroes2/assets/113276641/9b175395-ace1-4350-83f4-7b5ce23f4c5c

We can place this check before the 230 line and then no event processing will be done until the RMB is released.

src/fheroes2/dialog/dialog_resolution.cpp
```diff
@@ -245,22 +287,40 @@ namespace Dialog
                 fheroes2::Text body( _( "Click to apply the selected resolution." ), fheroes2::FontType::normalWhite() );
                 fheroes2::showMessage( header, body, 0 );
             }
-
-            if ( resList.isSelected() ) {
-                selectedResolution = resList.GetCurrent();
+            else if ( le.MousePressRight( listRoi ) ) {
+                continue;
```
Districh-ru, line 291: The same comment as previous: we can move this condition before the 266 line.

src/fheroes2/dialog/dialog_resolution.cpp
```diff
@@ -130,104 +135,141 @@ namespace
...
 
         LocalEvent & le = LocalEvent::Get();
         while ( le.HandleEvents() ) {
             le.MousePressLeft( buttonOk.area() ) && buttonOk.isEnabled() ? buttonOk.drawOnPress() : buttonOk.drawOnRelease();
             le.MousePressLeft( buttonCancel.area() ) ? buttonCancel.drawOnPress() : buttonCancel.drawOnRelease();
 
-            resList.QueueEventProcessing();
+            const int listId = listBox.getCurrentId();
+            listBox.QueueEventProcessing();
+            bool needRedraw = listId != listBox.getCurrentId();
```
Districh-ru, line 268: This variable also can be **const**. :)

Districh-ru (changes requested):
Hi, @zenseii. Thanks for updating these dialogs. In the future, I'll try to improve StandardWindow to reduce code duplication in such dialogs.
I left several comments. Could you please take a look in your spare time.

src/fheroes2/dialog/dialog_language_selection.cpp
```diff
@@ -200,23 +255,25 @@ namespace
                 fheroes2::showMessage( header, body, 0 );
             }
 
-            if ( items.isSelected() ) {
-                const fheroes2::SupportedLanguage newChosenLanguage = items.GetCurrent();
+            if ( !listBox.IsNeedRedraw() && !needRedraw ) {
```
Districh-ru, line 258: We can simplify this check: `needRedraw` can be `true` only after `_currentId` is changed during the `QueueEventProcessing();`. And if it is changed then `listBox.IsNeedRedraw()` will be true.
So there is no need to check `needRedraw` here.
```suggestion
            if ( !listBox.IsNeedRedraw() ) {
```

src/fheroes2/dialog/dialog_resolution.cpp
```diff
@@ -246,21 +292,36 @@ namespace Dialog
                 fheroes2::showMessage( header, body, 0 );
             }
 
-            if ( resList.isSelected() ) {
-                selectedResolution = resList.GetCurrent();
+            if ( !listBox.IsNeedRedraw() && !needRedraw ) {
```
Districh-ru, line 295: The same comment here :)
```suggestion
            if ( !listBox.IsNeedRedraw() ) {
```

Districh-ru (changes requested):
Hi, @zenseii. The new dialogs works okay. I just left a minor speed-up suggestion. Could you please consider it when you have time.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3709,6 +3692,96 @@ namespace fheroes2
...
+                    const int32_t evilOriginalBoxOffsetX = 26;
+                    const int32_t evilOriginalBoxOffsetY = 27;
+                    const int32_t upperPartHeight = 20;
+
+                    Copy( originalEvilBackground, evilOriginalBoxOffsetX, evilOriginalBoxOffsetY, _icnVsSprite[id][1], 0, 0, boxWidth, upperPartHeight );
+
+                    Copy( originalEvilBackground, evilOriginalBoxOffsetX, evilOriginalBoxOffsetY + upperPartHeight + 14, _icnVsSprite[id][1], 0, upperPartHeight,
+                          boxWidth, 10 );
+
+                    Copy( _icnVsSprite[id][0], 6, 5, _icnVsSprite[id][1], 6, 5, 359, 19 );
```
ihhub, line 3780: I am actually worried about this part of the code. Are we certain that we copy everything into the new image? If something is missing (like we forget to copy even a single pixel) then on release builds we are going to face some graphical artefacts. In this case we have 2 solutions:
1. Add `reset()` call after the creation of the image
2. Verify that we indeed populate the whole image completely

src/fheroes2/agg/agg_image.cpp
```diff
@@ -4095,6 +4168,20 @@ namespace fheroes2
...
+            case ICN::UNIFORMBAK_GOOD:
+            case ICN::UNIFORMBAK_EVIL: {
+                _icnVsSprite[id].resize( 1 );
+                const bool isEvilInterface = ( id == ICN::UNIFORMBAK_EVIL );
+                const fheroes2::Sprite & original = GetICN( isEvilInterface ? ICN::BUYBUILE : ICN::BUYBUILD, 1 );
+                if ( !original.empty() ) {
+                    _icnVsSprite[id][0].resize( 246, 45 );
+                    const fheroes2::Sprite & croppedImage = Crop( original, 0, 0, 125, 45 );
+                    Copy( croppedImage, 0, 0, _icnVsSprite[id][0], 0, 0, 123, 45 );
+                    Copy( croppedImage, 0, 0, _icnVsSprite[id][0], 123, 0, 123, 45 );
```
ihhub, line 4180: I think we can avoid having `Crop()` call here and directly call `Copy()` using the original image. It is going to be much faster.

ihhub (changes requested):
Hi @zenseii , I left several comments here. Would you mind to take a look at them in your spare time?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3709,6 +3692,97 @@ namespace fheroes2
...
+                    // Generate Evil Icons
+                    _icnVsSprite[id][1] = _icnVsSprite[id][0];
+
+                    const std::vector<uint8_t> & goodToEvilPalette = PAL::GetPalette( PAL::PaletteType::GOOD_TO_EVIL_INTERFACE );
+                    fheroes2::ApplyPalette( _icnVsSprite[id][0], _icnVsSprite[id][1], goodToEvilPalette );
+                }
+
+                return true;
+            }
+            case ICN::METALLIC_BORDERED_TEXTBOX: {
```
Districh-ru, line 3745: In many other cases the Evil sprite has a different ICN id while here we have the same ICN id and different image index for Good and Evil interfaces.
IMHO it should be good to leave a comment about this. :)
Or should we change to the `METALLIC_BORDERED_TEXTBOX_GOOD` and `METALLIC_BORDERED_TEXTBOX_EVIL` ICNs to correspond the all other cases with Good/Evil interface sprites?

src/fheroes2/dialog/dialog_adventure.cpp
```diff
@@ -21,106 +21,124 @@
...
+        background.renderButton( buttonWorld, apanel, 0, 1, { largeButtonsXOffset, largeButtonsYOffset }, fheroes2::StandardWindow::Padding::TOP_LEFT );
+        background.renderButton( buttonPuzzle, apanel, 2, 3, { largeButtonsXOffset, largeButtonsYOffset }, fheroes2::StandardWindow::Padding::TOP_RIGHT );
+        background.renderButton( buttonInfo, isEvilInterface ? ICN::BUTTON_INFO_EVIL : ICN::BUTTON_INFO_GOOD, 0, 1, { largeButtonsXOffset, largeButtonsYOffset + 2 },
+                                 fheroes2::StandardWindow::Padding::CENTER_LEFT );
+        background.renderButton( buttonDig, apanel, 6, 7, { largeButtonsXOffset, largeButtonsYOffset + 2 }, fheroes2::StandardWindow::Padding::CENTER_RIGHT );
+        background.renderButton( buttonCancel, isEvilInterface ? ICN::BUTTON_SMALL_CANCEL_EVIL : ICN::BUTTON_SMALL_CANCEL_GOOD, 0, 1, { 0, 11 },
+                                 fheroes2::StandardWindow::Padding::BOTTOM_CENTER );
+
+        if ( !enableDig ) {
+            buttonDig.disable();
```
Districh-ru, line 79: To render the disabled button we should also call `.draw()` method here:
```suggestion
            buttonDig.disable();
            buttonDig.draw();
```

src/fheroes2/dialog/dialog_adventure.cpp
```diff
@@ -21,106 +21,124 @@
...
-        if ( le.MousePressRight( buttonInfo.area() ) )
-            fheroes2::showStandardTextMessage( _( "Scenario Information" ), _( "View information on the scenario you are currently playing." ), Dialog::ZERO );
-        if ( le.MousePressRight( buttonDig.area() ) )
-            fheroes2::showStandardTextMessage( _( "Digging" ), _( "Dig for the Ultimate Artifact." ), Dialog::ZERO );
-        if ( le.MousePressRight( buttonCancel.area() ) )
-            fheroes2::showStandardTextMessage( _( "Cancel" ), _( "Exit this menu without doing anything." ), Dialog::ZERO );
-    }
+        // restore background
+        back.restore();
+        display.render( background.totalArea() );
```
Districh-ru, line 134: In the loop we can render only the `activeArea` (the inner dialog area where all dialog items can change). The `totalArea` should be rendered only when dialog is opened and closed to render also the dialod borders and its shadow.
```suggestion
        display.render( background.activeArea() );
```

This comment is also for other changed dialogs.

Districh-ru (changes requested):
Hi, @zenseii, I did a quick overview of PR and left some comments. Could you please check them when you have time.
I'll do a more thorough review tomorrow.

src/fheroes2/dialog/dialog_adventure.cpp
Districh-ru: @zenseii, you are absolutely right with these changes!

src/fheroes2/gui/player_info.cpp
```diff
@@ -427,9 +427,9 @@ bool Interface::PlayersInfo::QueueEventProcessing()
         if ( player != nullptr ) {
             if ( conf.getCurrentMapInfo().AllowChangeRace( player->GetColor() ) ) {
```
Districh-ru, line 428: To make Sonar a little happier we should combine these two conditions:
```suggestion
        if ( player != nullptr && conf.getCurrentMapInfo().AllowChangeRace( player->GetColor() ) ) {
```

src/fheroes2/gui/player_info.cpp
```diff
@@ -440,9 +440,9 @@ bool Interface::PlayersInfo::QueueEventProcessing()
...
```
Districh-ru, line 441: These two conditions can also be combined.

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -315,70 +354,66 @@ namespace
...
-                    ratingRoi = RedrawRatingInfo( rectPanel.getPosition(), rectPanel.width );
-                    buttonOk.draw();
-                    buttonCancel.draw();
-                    display.render();
+
+                    display.render( background.activeArea() );
                 }
             }
             else if ( le.MouseWheelUp() || le.MouseWheelDn() ) {
                 if ( playersInfo.QueueEventProcessing() ) {
```
Districh-ru, line 381: These two conditions can also be combined. :)

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -84,43 +85,33 @@ namespace
...
 
         // text class
         text.set( _( "Class:" ), normalWhiteFont );
-        text.draw( rt.x, rt.y + 264, rt.width, display );
+        text.draw( rt.x, rt.y + 248, rt.width, display );
+    }
+
+    void RedrawMapTitle( const fheroes2::Rect & roi )
+    {
+        fheroes2::Text text( Settings::Get().getCurrentMapInfo().name, fheroes2::FontType::normalWhite() );
```
Districh-ru, line 113: This variable can be `const`.

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -282,25 +315,31 @@ namespace
...
-                    levelCursor.setPosition( coordDifficulty[Game::getDifficulty()].x, coordDifficulty[Game::getDifficulty()].y ); // From 0 to 4, see: Difficulty enum
-                    buttonOk.draw();
-                    buttonCancel.draw();
-                }
 
+                    ratingRoi = RedrawRatingInfo( roi.getPosition(), roi.width );
+                    levelCursor.setPosition( coordDifficulty[Game::getDifficulty()].x - levelCursorOffset,
+                                             coordDifficulty[Game::getDifficulty()].y - levelCursorOffset ); // From 0 to 4, see: Difficulty enum
+                }
                 display.render();
```
Districh-ru, line 343: Here we can do render only in the `activeArea()`:
```suggestion
                display.render( roi );
```

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -282,25 +315,31 @@ namespace
...
-                    RedrawDifficultyInfo( pointDifficultyInfo );
                     playersInfo.resetSelection();
                     playersInfo.RedrawInfo( false );
-                    ratingRoi = RedrawRatingInfo( rectPanel.getPosition(), rectPanel.width );
-                    levelCursor.setPosition( coordDifficulty[Game::getDifficulty()].x, coordDifficulty[Game::getDifficulty()].y ); // From 0 to 4, see: Difficulty enum
-                    buttonOk.draw();
-                    buttonCancel.draw();
-                }
 
+                    ratingRoi = RedrawRatingInfo( roi.getPosition(), roi.width );
```
Districh-ru, line 339: Rating text possibly can be better vertically aligned with the OK/Cancel buttons (about 3 pixels lower), what do you think?
изображение

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -315,70 +354,66 @@ namespace
...
-                    RedrawDifficultyInfo( pointDifficultyInfo );
-                    playersInfo.RedrawInfo( false );
-                    ratingRoi = RedrawRatingInfo( rectPanel.getPosition(), rectPanel.width );
-                    buttonOk.draw();
-                    buttonCancel.draw();
-                    display.render();
+                    ratingArea.restore();
+                    ratingRoi = RedrawRatingInfo( roi.getPosition(), roi.width );
+
+                    display.render( background.activeArea() );
```
Districh-ru, line 368: We can simply use `roi` here not to call once again the `activeArea()` method:
```suggestion
                    display.render( roi );
```

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -315,70 +354,66 @@ namespace
...
+                    opponentsArea.restore();
+                    classArea.restore();
+                    handicapArea.restore();
                     playersInfo.RedrawInfo( false );
-                    ratingRoi = RedrawRatingInfo( rectPanel.getPosition(), rectPanel.width );
-                    buttonOk.draw();
-                    buttonCancel.draw();
-                    display.render();
+
+                    display.render( background.activeArea() );
```
Districh-ru, line 377: And here too. :)
```suggestion
                    display.render( roi );
```

src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -315,70 +354,66 @@ namespace
...
```
Districh-ru, line 388:
```suggestion
                    display.render( roi );
```

Districh-ru (changes requested):
Hi, @zenseii, I left several comment. Could you please take a look when you have time?

src/fheroes2/dialog/dialog_adventure.cpp
```diff
@@ -21,106 +21,121 @@
...
 
-    // restore background
-    back.restore();
-    display.render();
-
+int Dialog::AdventureOptions( const bool enableDig )
+{
+    const int result = selectAdventureOption( enableDig );
     return result;
 }
```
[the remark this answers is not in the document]
Districh-ru: OK, we can leave it if it is safer and fixes some crashes. But having a lot of functions that only calls one next function can result in a mess of calls. :)
I never encountered such bugs on resolution change.... If I do I'll check what is happening.

src/fheroes2/dialog/dialog_file.cpp
```diff
@@ -33,116 +34,128 @@
...
+        // setup cursor
+        const CursorRestorer cursorRestorer( true, Cursor::POINTER );
+
+        fheroes2::Display & display = fheroes2::Display::instance();
+
+        const int32_t totalDialogWidth = 337;
+        const int32_t totalDialogHeight = 252;
+
+        // Prepare restorer of the adventure map when save feedback dialog is shown.
+        fheroes2::ImageRestorer back( display, ( display.width() - totalDialogWidth - BORDERWIDTH ) / 2, ( display.height() - totalDialogHeight + BORDERWIDTH ) / 2,
```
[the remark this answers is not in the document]
Districh-ru: It was needed before we started to use the `StandardWindow` which has its own restorer that is called in destructor.
`return Interface::AdventureMap::Get().EventSaveGame();` finises the execution of the current function and the lifeline of created instance of `StandardWindow`. So the destructor with the restorer will be called and there is no need to restore the background just before the destructor which will do the same.
If the new dialog was called in this function then we would need to use `back.restore();`.

src/fheroes2/dialog/dialog_file.cpp
```diff
@@ -33,116 +34,128 @@
...
-    display.render();
+        return result;
+    }
+}
 
+fheroes2::GameMode Dialog::FileOptions()
+{
+    const fheroes2::GameMode result = selectFileOption();
     return result;
 }
```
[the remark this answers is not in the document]
Districh-ru: OK, lets leave the decision to @ihhub. :)

src/fheroes2/game/game_scenarioinfo.cpp
Districh-ru: Ow, then its OK. :)

src/fheroes2/dialog/dialog_file.cpp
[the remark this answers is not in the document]
Districh-ru: Sorry, I was wrong.

Districh-ru:
@zenseii, could you please resolve the merge conflicts.

src/fheroes2/dialog/dialog_file.cpp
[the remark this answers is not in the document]
Districh-ru: Yes, the method `EventSaveGame()` is called from the `return`, so the `StandardWindow` destructor will be executed only after this function.
We can use a `std::unique_ptr` for `StandardWindow` instance and "release" it here just before the `return`.
But lets do it later, this PR focuses on the other things and is big enough. Lets merge it. :)

Districh-ru (approved):
Well done @zenseii, the GUI gets better and better!

ihhub (approved):

ihhub:
@zenseii , huge thanks for these essential updates!

PR #8159, opened 2023-12-21
Radar rendering optimization

Districh-ru:
This PR polishes radar (mini-map) rendering code a bit.
These changes increase radar rendering speed by about 5% (when fully rendering the fully revealed map).


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -930,7 +932,11 @@ namespace Interface
 
     void EditorInterface::mouseCursorAreaPressRight( const int32_t tileIndex ) const
     {
-        Editor::showPopupWindow( world.GetTiles( tileIndex ) );
+        const Maps::Tiles & tile = world.GetTiles( tileIndex );
+
+        DEBUG_LOG( DBG_DEVEL, DBG_INFO, std::endl << tile.String() )
```
ihhub, line 937: I suggest not to use `std::endl` but "\n". Then we won't need extra dependency in the code.

Also, why do we need this change and if we do should we move it to `showPopupWindow()` function?

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -1737,15 +1737,17 @@ int32_t Maps::Tiles::getIndexOfMainTile( const Maps::Tiles & tile )
...
+            const int32_t index = offsetX + x;
             if ( Maps::isValidAbsIndex( index ) ) {
                 const Maps::Tiles & foundTile = world.GetTiles( index );
                 if ( foundTile.GetObject( false ) != correctedObjectType ) {
                     continue;
                 }
 
-                if ( foundTile.GetObjectUID() != 0 && uids.count( foundTile.GetObjectUID() ) > 0 ) {
+                const uint32_t objectUid = foundTile.GetObjectUID();
+                if ( objectUid != 0 && uids.count( objectUid ) > 0 ) {
```
ihhub, line 1750: This change won't increase performance as **GetObjectUID()** is an inlined function.

ihhub (changes requested):
Hi @Districh-ru , I put 2 comments here. Can you check them in your free time?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Hi, @ihhub.
This code is needed to check tiles in editor in debug mode when we implement new editor features. We have the same one on the game code to help in fixing issues.
When starting this PR I also wanted to fix some radar rendering issues in editor and used this code. But I found that we need firs to implement containers like `vec_castles` and `vec_heroes` only after we will fix the radar.
We can remove this code from this PR, but add it to the next editor-related PR.

src/fheroes2/maps/maps_tiles.cpp
Districh-ru: Ow, yes)
I reverted the change.

src/fheroes2/editor/editor_interface.cpp
ihhub: Let's then move the code into `showPopupWindow()` function.

Districh-ru: OK, I did it in the last commit. :)

ihhub (approved):

ihhub:
@Districh-ru , thanks a lot for this optimization!

PR #8160, opened 2023-12-22
Fix multiple spellbook-related issues in the Battle Only mode

oleg-derevenetz:

https://github.com/ihhub/fheroes2/assets/32623900/f789af4c-6daa-4825-a525-d897886a9d5f

With this PR:

https://github.com/ihhub/fheroes2/assets/32623900/fe237e04-77f9-40fe-8b5f-7cb39bf6a3b1

Another problem that this PR is intended to fix is that when removing the spell book through the UI, the list of spells that the hero knows has not actually been cleared. Also in the `master` branch, for the purposes of evaluating heroes and their armies by the AI, spells from scrolls were taken into account even if the hero did not have a spell book (and he could not actually use these scrolls).


Districh-ru (approved):

ihhub (approved):

ihhub:

PR #8173, opened 2023-12-24
Fix too wide text rendering on buttons

a1exsh:
In German, for example both "Neustart" (Restart) and "Ablehnen" (Decline) don't fit into the button width.

image ⇒ image

Ideally, the translation could be improved, but at least it looks less weird when it doesn't fit. :)

:santa: :gift:

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -444,8 +444,13 @@ namespace
 
         const char * textSupported = getSupportedText( text, releasedFont );
 
-        const fheroes2::Text releasedText( textSupported, releasedFont );
-        const fheroes2::Text pressedText( textSupported, pressedFont );
+        fheroes2::Text releasedText( textSupported, releasedFont );
+        fheroes2::Text pressedText( textSupported, pressedFont );
+
+        if ( releasedText.height( buttonSize.width ) > buttonSize.height || pressedText.height( buttonSize.width ) > buttonSize.height ) {
+            releasedText.fitToOneRow( buttonSize.width );
+            pressedText.fitToOneRow( buttonSize.width );
+        }
 
         const fheroes2::Size releasedTextSize( releasedText.width( buttonSize.width ), releasedText.height( buttonSize.width ) );
         const fheroes2::Size pressedTextSize( pressedText.width( buttonSize.width ), pressedText.height( buttonSize.width ) );


ihhub:
Hi @zenseii , does this change correlate with your pull request?

a1exsh:
@ihhub @zenseii ah, I wasn't aware of existing PRs in this area.

For example, there is at least this further instance in the German translation that doesn't fit nicely:
image

Do you have it in scope of your changes?

ihhub:
Hi @a1exsh , let's wait for @zenseii  reply. He is the one working on all sorts of buttons at the moment.

ihhub:
Hi @zenseii and @a1exsh , I am actually thinking that we should combine both approaches. We can expand buttons up to certain limits but not exceed them. What do you think?

a1exsh:
> Hi @zenseii and @a1exsh , I am actually thinking that we should combine both approaches. We can expand buttons up to certain limits but not exceed them. What do you think?

Makes sense. I'd also keep the code for truncation, just in case a translation is updated and doesn't fit once again, as a fallback to avoid the two-line rendering as on the screenshot.

Districh-ru (commented):
Hi, all!
For the tall buttons (e.g. main menu) if the translated text is taller than the button it will be truncated to one line.
I suggest to add a TODO: Implement text truncating to the specified number of lines.

a1exsh:
> For the tall buttons (e.g. main menu) if the translated text is taller than the button it will be truncated to one line.

That's true. At the same time, the tall buttons have more space and less likely to trigger this. It's a fallback mainly for single-line buttons.
Have you seen a translation that cuts it too short in the main menu or it's theoretical?

Districh-ru:
> That's true. At the same time, the tall buttons have more space and less likely to trigger this. It's a fallback mainly for single-line buttons. Have you seen a translation that cuts it too short in the main menu or it's theoretical?

I have not seen such translations, but theoretically it is possible, so I suggest adding only a TODO comment.

a1exsh:
> > > For the tall buttons (e.g. main menu) if the translated text is taller than the button it will be truncated to one line.
> > 
> > 
> > That's true. At the same time, the tall buttons have more space and less likely to trigger this. It's a fallback mainly for single-line buttons. Have you seen a translation that cuts it too short in the main menu or it's theoretical?
> 
> The Polish translation has many such issues in the main menu. A typical issue in such areas are the two campaign buttons.

Ah, indeed.  Didn't realize it can go up to 3 lines on the tall buttons, but even that isn't always enough:
image

Do we address it in your PR or that needs extra effort?

ihhub (approved):

zenseii (commented):
@a1exsh. I've tested how this currently works and as I said it could work as a final measure in case a text doesn't fit. However, I've since addressed all the buttons which currently had too wide texts in languages (that I could find), by updating them to the auto-adjustable buttons.

There is also an issue where this solution cuts text that is shown fully on other buttons, like Kampania here:
image

Still, as I mentioned before, these buttons will be auto-adjustable too. In fact I have that among my (several) goals for the next release.

One limitation with this PR's solution is that it naturally can't help with auto-adjustable buttons like the ones on the campaign screen:
image

To solve these my plan is as mentioned to make the letters go closer once the text's width exceeds a certain threshold. Of course, in the meantime it would have been better if a translator could make abbreviations for these strings, since ultimately they have to check if their translations fit within the UI in many other places too, like the creature info dialog.

XCOM-HUB:
The Swedish translation also has some of these issues.

Lämn
Mkt  snabb

Branikolog:
Hi, @XCOM-HUB!

I believe the text in the creature characteristics window is related to another issue (#7548), which is expected to be resolved by extending the window itself. Hopefully, this will be addressed in the near future.

XCOM-HUB:
@Branikolog – Ok. Even if it isn't addressed in the upcoming release of fheroes2, I have shortened the labels that are currently too long for the characteristics window. "HP kvar" instead of "Hälsopoäng kvar", "Kvick" instead of "Mkt. snabb", and so on.

Branikolog:
@XCOM-HUB 

Indeed, this is exactly the approach we usually follow. Even with dynamically resizing buttons, we still cannot fit unlimited amounts of text into them. This is why translators in fheroes2 always try to find the shortest possible wording for button labels.
Unfortunately, with support for so many languages, it is not always possible to keep translations equally short. In those cases, the ability to adapt button width becomes very helpful.

PR #8174, opened 2023-12-24
Hide swap army/artifact buttons in the Battle Only

a1exsh:
The buttons are not actually active, so it's confusing to have them there:

- Before:
   image
- After:
  image

Related: #4352

### DIFF
--- a/src/fheroes2/battle/battle_only.cpp
+++ b/src/fheroes2/battle/battle_only.cpp
@@ -465,8 +465,8 @@ void Battle::Only::RedrawBaseInfo( const fheroes2::Point & top ) const
 
     // hide the swap army/artifact arrows
     const fheroes2::Sprite & moveButtonBackground = fheroes2::AGG::GetICN( ICN::STONEBAK, 0 );
-    fheroes2::Blit( moveButtonBackground, 292, 270, display, top.x + 292, top.y + 270, 48, 44 );
-    fheroes2::Blit( moveButtonBackground, 292, 363, display, top.x + 292, top.y + 363, 48, 44 );
+    fheroes2::Copy( moveButtonBackground, 292, 270, display, top.x + 292, top.y + 270, 48, 44 );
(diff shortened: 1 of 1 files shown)


src/fheroes2/battle/battle_only.cpp
```diff
@@ -462,6 +462,11 @@ void Battle::Only::RedrawBaseInfo( const fheroes2::Point & top ) const
     }
 
     fheroes2::RedrawPrimarySkillInfo( top, armyInfo[0].ui.primarySkill.get(), armyInfo[1].ui.primarySkill.get() );
+
+    // hide the swap army/artifact arrows
+    const fheroes2::Sprite & moveButtonBackground = fheroes2::AGG::GetICN( ICN::STONEBAK, 0 );
+    fheroes2::Blit( moveButtonBackground, 292, 270, display, top.x + 292, top.y + 270, 48, 44 );
+    fheroes2::Blit( moveButtonBackground, 292, 363, display, top.x + 292, top.y + 363, 48, 44 );
```
Districh-ru, line 469: Since the original image (STONEBAK) has no shadows or transparency, we can use `Copy()` instead of `Blit()` to speed up rendering a bit.
```suggestion
    fheroes2::Copy( moveButtonBackground, 292, 270, display, top.x + 292, top.y + 270, 48, 44 );
    fheroes2::Copy( moveButtonBackground, 292, 363, display, top.x + 292, top.y + 363, 48, 44 );
```

Districh-ru (changes requested):
Hi, @a1exsh, thanks for your contribution!
I left one minor suggestion. Could you please take a look when you have time.

src/fheroes2/battle/battle_only.cpp
ihhub: No, `Blit()` behaves as `Copy()` only when handling single-layer images. By default images are double-layer images. Therefore, the complexity of the operation increases leading to slower execution.

ihhub (approved):

Districh-ru (approved):
👍

ihhub:
@a1exsh , thank you so much for the fix!

PR #8176, opened 2023-12-24
AI: use the "cautious offensive" tactics only against melee-only armies without the distant attack potential (including damage spells)

### LINKED ISSUE #7973: Combat, AI is moving too safe while opposing army with strong ranged troops

### LINKED ISSUE #8070: Combat, enemy advancement issue
### Preliminary checks

### Platform

Windows

### Describe the bug

Enemy is not using the whole movement range if I place my troops at the distance (move speed+2) in front of his troops.

https://github.com/ihhub/fheroes2/assets/55348946/afc90dd1-d9ef-4f9c-93bc-b64f192fab08

This could be highly abused by cheating humans allowing archers to shoot a few more turns.
I think there's no need to use such tactics if AI troops are slower than my.

Even if AI is going towards slower troops I can abuse this logic and slow down swordsmen for a few turns:

https://github.com/ihhub/fheroes2/assets/55348946/0499596d-5a72-4c35-9be2-6e58314ae5cd

I suggest swordsmen rush without slowing down in a case enemy army has strong shooters and/or weak troop standing in front of swordsmen.

### Save file

None.

### Additional info

_No response_

oleg-derevenetz:
fix #8070
close #7973

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal.h
+++ b/src/fheroes2/ai/normal/ai_normal.h
@@ -250,6 +250,7 @@ namespace AI
         bool _defendingCastle = false;
         bool _considerRetreat = false;
         bool _defensiveTactics = false;
+        bool _cautiousOffensive = false;
     };
 
     class Normal : public Base
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -834,7 +834,12 @@ namespace AI
             _enemyShootersStrength += commanderMaximumSpellDamageValue( *enemyCommander );
         }
 
-        _defensiveTactics = [this, &currentUnit]() {
+        assert( _myArmyStrength > 0.0 && _enemyArmyStrength > 0.0 );
+
+        const double myArcherRatio = _myShootersStrength / _myArmyStrength;
+        const double enemyArcherRatio = _enemyShootersStrength / _enemyArmyStrength;
+
+        _defensiveTactics = [this, &currentUnit, myArcherRatio, enemyArcherRatio]() {
             // Unit is already in the enemy half of the battlefield, just let it keep attacking
             if ( !isPositionLocatedInDefendedArea( currentUnit, currentUnit.GetPosition() ) ) {
                 return false;
@@ -858,21 +863,13 @@ namespace AI
                 return true;
             }
 
-            assert( _myArmyStrength > 0.0 && _enemyArmyStrength > 0.0 );
-
-            const double myArcherRatio = _myShootersStrength / _myArmyStrength;
-            const double enemyArcherRatio = _enemyShootersStrength / _enemyArmyStrength;
-
-            const double myArcherThreshold = 0.15;
-            const double enemyArcherThreshold = 0.66;
-
             // If we have an unfavorable ratio of infantry and shooters for defense, then it is better to choose an offensive
-            if ( myArcherRatio < myArcherThreshold ) {
+            if ( myArcherRatio < 0.15 ) {
                 return false;
             }
 
             // If the enemy has too many shooters, but not enough infantry to cover them, then it makes sense to choose an offensive
-            if ( enemyArcherRatio > enemyArcherThreshold ) {
+            if ( enemyArcherRatio > 0.66 ) {
                 return false;
             }
 
@@ -880,12 +877,12 @@ namespace AI
         }();
 
         // If an offensive tactic is chosen, then this means that, most likely, the enemy has more shooters, which means that we should try to attack the enemy and
-        // neutralize his shooters as quickly as possible. A cautious offensive tactics can be chosen only if our army is fighting an enemy army that cannot attack
-        // from a distance.
-        _cautiousOffensive = ( _enemyShootersStrength < 0.1 );
+        // neutralize his shooters as quickly as possible. A cautious offensive tactics can be chosen only if our army is fighting an enemy army that has limited
+        // distance attack capabilities.
+        _cautiousOffensive = ( enemyArcherRatio < 0.15 );
 
         DEBUG_LOG( DBG_BATTLE, DBG_TRACE,
-                   ( _defensiveTactics ? "Defensive" : "Offensive" )
+                   ( _defensiveTactics ? "Defensive" : ( _cautiousOffensive ? "Cautious offensive" : "Offensive" ) )
                        << " tactics have been chosen. Army strength: " << _myArmyStrength << ", shooters strength: " << _myShootersStrength
                        << ", enemy army strength: " << _enemyArmyStrength << ", enemy shooters strength: " << _enemyShootersStrength )
     }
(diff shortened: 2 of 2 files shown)


ihhub (approved):

Branikolog:
Hi, @oleg-derevenetz !
This problem bothered me so much time. I'm so glad, here's the PR fixing it! :)
I've already started testing battles and here's one thing that I start thinking after a few dozens of battles:

https://github.com/ihhub/fheroes2/assets/55348946/dbdf3198-4be1-4d4c-9c1f-4793bd183729

I think 1 ranger is not a huge threat for minotaurs to switch off "cautious offensive" tactics. How do you think? I suggest comparing the overall shooters strength with defending melee army, If shooters make significantly less damage, than melee troops could do to attacking troops (minotaurs) AI can be a little bit more cautious. :)

oleg-derevenetz:
Hi @Branikolog

> I suggest comparing the overall shooters strength with defending melee army, If shooters make significantly less damage

So what should be the factor of archers strength vs total strength? 5%? 10%? 15%? Do you remember about the corner cases that are "attached" to each constant of this kind?

Branikolog:
@oleg-derevenetz 
Corner cases... Sure! :D
What if we use the same value is used for defending archers? If their strength is not so high, comparing the melee defending troops, attackers shouldn't really bother much rushing them and be attacked by defenders.

Branikolog:
But anyway, the value is up to you.
I hope, you agree, that 1 ranger is not really a reason for minotaurs to appear under attack of brave swordsmen? :)

ihhub:
@Branikolog , as usual please be more specific. Give examples when N shooters matter but N - 1 don't. Also how about 1 Mage vs Minotaurs?

oleg-derevenetz:
@Branikolog OK, for now, let's make such an assessment that we can resort to cautious offensive if the share of archers in the total power of the enemy army is no more than 15%.

Branikolog:
@oleg-derevenetz 
Thanks! Now I feel it is working better. We can later adjust the value (maybe enemy can rush if shooters are 1-2% weaker, than the current value) , but now it's +/- adequate.

@ihhub 
Sorry, but I couldn't be more specific here, since we don't count on the number of troops, but on the strength. I'm not really good in calculations of troops threat. But @oleg-derevenetz made it well!

Branikolog:
@oleg-derevenetz 
One more related thing I want to mention. It is probably out of the PR, but I would like to share it and hear your opinion.

https://github.com/ihhub/fheroes2/assets/55348946/09dacf45-89d6-4e7c-87fb-4b0a124c09c4

Champions are definitely much faster, than golems and they really don't need to advance ahead so much. 
I think it would be nice if champions try to position themselves out of the range of defending slow golems, even in "non-cautious" state.

oleg-derevenetz:
@Branikolog

> Champions are definitely much faster, than golems and they really don't need to advance ahead so much.

We need to develop a common algorithm, and this cannot be based on just one situation. What if there is also a faster stack in addition to golems? What if golem's stacks are stronger than this stack? What if that faster stack is stronger than stacks of golems? Etc, etc. The more complicated the logic, the worse, because with each new condition, a lot of new boundary situations and gray areas arise.

Branikolog:
@oleg-derevenetz 
 
> > Champions are definitely much faster, than golems and they really don't need to advance ahead so much.
> 
> We need to develop a common algorithm, and this cannot be based on just one situation. What if there is also a faster stack in addition to golems? What if golem's stacks are stronger than this stack? What if that faster stack is stronger than stacks of golems? What if they are roughly equal in power? Etc, etc. The more complicated the logic, the worse, because with each new condition, a lot of new boundary situations and gray areas arise.

Is it within the scope of this PR, or should we discuss it in a separate thread/PR-draft?

oleg-derevenetz:
@Branikolog

> Is it within the scope of this PR, or should we discuss it in a separate thread/PR-draft?

Since there is not even an approximate clarity on how exactly this should work right now, this is definitely beyond the scope of this PR :)

Branikolog:
@oleg-derevenetz 

> Since there is not even an approximate clarity on how exactly this should work right now, this is definitely beyond the scope of this PR :)

I have some approximate ideas. :D But the logic is going to consist of multiple rules, so lets hold it back or discuss somewhere else. :)
For now I would like to test this PR for several hours more to be sure I'm approving a 100% working improvement. :)

oleg-derevenetz:
@Branikolog Just for your information, how exactly the "cautious offensive" mode works currently. The candidate path of the unit is taken and for each tile tile "threat level" is calculated, and the farthest cell of the path is chosen from those with the least threat level. For example, if such a path passes through tiles with indexes 1, 2, 3 and there is unit A with threat level 100 and unit B with threat level 200, but unit B can access tiles 2 and 3, while unit A can access tiles 1 and 2, then path should be cut to the tile 1. However, if unit B can access just tile 2, then the path does not need to be cut at all. Here is an example of a general solution (certainly not ideal, but it works well enough). Need to work out something like that.

oleg-derevenetz:
> I have some approximate ideas.

It is necessary to have a common mathematical/logical model. E.g. "for each our unit we take the metric A (or calculate it in this way), then compare it with coefficient X, if it more than X, then we do Y, else we take another metric B for each enemy unit, ..." etc. Here's how computers behave :)

Branikolog (approved):
@oleg-derevenetz 
>It is necessary to have a common mathematical/logical model. E.g. "for each our unit we take the metric A (or calculate it in this way), then compare it with coefficient X, if it more than X, then we do Y, else we take another metric B for each enemy unit, ..." etc. Here's how computers behave :) Excluding maybe neural networks, but that's a completely different story.
Thank you for this explanation! I understand it better how it works. If I can, I come up with the proposal how we can improve the logic. Considering the explained current implementation.
The PR is OK for me. Haven't found any issues.

ihhub:

Issue #13, opened 2019-09-02
No animation for resource collection


ihhub:
There is no animation for resource collection and disappearance on MacOS with SDL 2.

vasilenkoalexey:
I am not sure, but can be resolved by https://github.com/ihhub/fheroes2/pull/383

ihhub:
@vasilenkoalexey your solution fixes the issue.

PR #8183, opened 2023-12-25
Hungarian language update #13

big4billy:
* Some grammar corrections (especialy on artifact names' capitals);
* Changes on speed names because of short place in dialogue windows;
* Newly appeared fuzzy removals on Editor section;
* Some new translations


files/lang/hu.po
```diff
@@ -9571,10 +9565,10 @@ msgid ""
 msgstr ""
 
 msgid "Endless Pouch of Sulfur"
-msgstr "Végtelen Kénzacskó"
+msgstr "Végtelen kénzacskó"
 
 msgid "The %{name} provides %{count} unit of sulfur per day."
-msgstr "A(z) %{name} napi %{count} ként biztosít neked."
+msgstr "A %{name} napi %{count} ként biztosít neked."
```
ihhub, line 9571: Why are we modifying the same sentence again when we've just recently added it? This is true for many lines where we are removing **(z)**

ihhub (changes requested):
Hi @big4billy , I have one common question in this pull request. Could you please check it?

files/lang/hu.po
big4billy: That was a dilemma of us because of the Hungarian definite article grammar rules. There are two definite article in our language: "a" and "az". The meaning are the same "the" but "a" should stand before a consonant starting noun and "az" before a vowel starting word. In our case, it's important to know the starting letters of {name} parameters. There is azn universal solution using "a(z)" marking the possibility of both cases but it's not too aesthetic in look. I thought, if {name} parameter could be just one determined word, I try to fit the definite articles to the only possible parameter words. This round shows you these changes from "a(z)" to the fitted articles.

big4billy: I'm trying to be not just accurate in meaning but grammarly correct in our language as well.

big4billy: I hoped it's enough to explain our modifications. I didn't revert any sentence of changes with "a(z)", so yes, these is the current strings actually.

big4billy: I think these changes are right and nothing to to with them. They might be approved.

ihhub (approved):

ihhub:
@big4billy , thank you so much for the changes!

PR #8184, opened 2023-12-26
Fix inaccurate Jail removal

### LINKED ISSUE #8179: Assertion failure when opening a jail located next to another jail
### Preliminary checks

### Platform

Windows

### Describe the bug

https://github.com/ihhub/fheroes2/assets/32623900/b30f56f3-81ac-4422-9411-37cd367e68cf

In this case assertion failure happens due to the fact that the object type at tile 37 (the top left jail in a row) becomes `OBJ_NONE` after calling the `Maps::removeObjectSprite()` for the tile with the jail opened by the AI-controlled hero.

### Save file

[!HIRETST.zip](https://github.com/ihhub/fheroes2/files/13766175/HIRETST.zip)

### Additional info

_No response_

ihhub:
close #8179

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1443,6 +1443,7 @@ void Maps::Tiles::fixTileObjectType( Tiles & tile )
 
 void Maps::Tiles::Remove( uint32_t uniqID )
 {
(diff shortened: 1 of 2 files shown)


oleg-derevenetz (approved):
I have tested the behavior of this PR in a number of situations and it seems to fix the issue.

PR #8185, opened 2023-12-26
Expand the map format

ihhub:
- add supported language (see comments in the code)
- add metadata structures for the objects. Castle metadata is not complete as we need to define what we want to store

### DIFF
--- a/src/fheroes2/maps/map_format_info.cpp
+++ b/src/fheroes2/maps/map_format_info.cpp
@@ -81,6 +81,7 @@ namespace Maps::Map_Format
         if ( metadata.metadata.size() != temp.size() ) {
             // This is a corrupted file!
             assert( 0 );
+            metadata.metadata = { 0 };
         }
         else {
             std::copy_n( temp.begin(), metadata.metadata.size(), metadata.metadata.begin() );
@@ -104,6 +105,7 @@ namespace Maps::Map_Format
         if ( metadata.defenderMonsterType.size() != temp.size() ) {
             // This is a corrupted file!
             assert( 0 );
+            metadata.defenderMonsterType = { 0 };
         }
         else {
             std::copy_n( temp.begin(), metadata.defenderMonsterType.size(), metadata.defenderMonsterType.begin() );
@@ -115,6 +117,7 @@ namespace Maps::Map_Format
         if ( metadata.defenderMonsterCount.size() != temp.size() ) {
             // This is a corrupted file!
             assert( 0 );
+            metadata.defenderMonsterCount = { 0 };
         }
         else {
             std::copy_n( temp.begin(), metadata.defenderMonsterCount.size(), metadata.defenderMonsterCount.begin() );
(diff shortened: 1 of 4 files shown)


src/fheroes2/maps/map_format_info.h
```diff
@@ -72,6 +93,9 @@ namespace Maps::Map_Format
 
         std::string name;
         std::string description;
+
+        // The language in which the map was made. This is important in order to load map's texts and display them properly.
+        uint8_t supportedLanguage{ 0 };
```
oleg-derevenetz, line 98: I'm just wondering what we should do if we load a map that uses a language other than what is used in the game. The game does not use Unicode (and the fonts from the original game are not Unicode), so in any case, we most likely will not be able to show characters from another language from events, hero names, and so on.

ihhub: We can at least warn a user about this. Right now we have no such an ability. Are you suggesting that we should move to Unicode? Then we need to complete UTF-8 integration :(

oleg-derevenetz: > Are you suggesting that we should move to Unicode?

I don't know about this really, this can be done later (If we have to, because creating Unicode fonts for the whole game that will match the style of the original game is still a problem, and we need multiple fonts of this kind I suppose). Just wondering, because without Unicode fonts there is no way to show texts in multiple languages at the same time. But we can at least warn the user, yes.

ihhub: My idea was to at least to allow a creator of a map to choose a language they want (or multiple languages) to use while creating a map. Then the engine read this information, outputs properly (say hi to ??? ??? ??? in the list of maps) and displays these texts within the engine. I think this will require less development time than creation of fonts.

However, at first we can support only English and I can remove this member from the structure.

src/fheroes2/maps/map_format_info.cpp
```diff
@@ -66,28 +68,84 @@ namespace Maps::Map_Format
...
+    }
+
+    StreamBase & operator>>( StreamBase & msg, StandardObjectMetadata & metadata )
+    {
+        std::vector<uint32_t> temp;
+        msg >> temp;
+
+        if ( metadata.metadata.size() != temp.size() ) {
+            // This is a corrupted file!
+            assert( 0 );
```
oleg-derevenetz, line 83: Assertion may be fine for debug builds, but in release builds, if we can't read the field properly, then we should also set the data being loaded to some default sane values, I suppose (here and in other similar places).

src/fheroes2/maps/map_format_info.h
oleg-derevenetz: No, I'm fine with it. IIRC the map format is versioned, so there is no problem with changing it later.

oleg-derevenetz (changes requested):
Hi @ihhub I just left one suggestion, please take a look when you have time.

src/fheroes2/maps/map_format_info.cpp
ihhub: Good point.

src/fheroes2/maps/map_format_info.h
ihhub: I reverted this change for now since we don't even use it.

oleg-derevenetz (approved):

Districh-ru (approved):

PR #8188, opened 2023-12-26
Modify the logic to calculate owned mines

ihhub:
We shouldn't depend on some image indices but on metadata from a tile.


oleg-derevenetz (approved):
LGTM. I wonder why `OBJ_HEROES` were also taken into account earlier.

Districh-ru (approved):

ihhub:
> LGTM. I wonder why `OBJ_HEROES` were also taken into account earlier.

I think a very early version of the engine (before GitHub) had different way of handling heroes and therefore this logic was there. But this is just my theory.

PR #8197, opened 2023-12-27
Implement makeShadow() to work with single-layer images

Districh-ru:
This PR makes `makeShadow()` to make a shadow for the input single-layer image like for the fully non-transparent double-layer image.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -2568,14 +2568,14 @@ namespace fheroes2
 
         assert( !out.empty() );
 
-        const int32_t widthOut = out.width();
-
         if ( in.singleLayer() ) {
             // In this case we add a shadow of the fully non-transparent rectangular 'in' image.
             FillTransform( out, 0, shadowOffset.y, width, height, transformId );
 
             return out;
         }
+
+        const int32_t widthOut = out.width();
 
         const uint8_t * transformInY = in.transform();
         const uint8_t * transformInYEnd = transformInY + width * height;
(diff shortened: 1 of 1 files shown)


src/engine/image.cpp
```diff
@@ -2568,6 +2568,13 @@ namespace fheroes2
 
         assert( !out.empty() );
 
+        if ( in.singleLayer() ) {
+            // In this case we add a shadow of the fully non-transparent rectangular 'in' image.
+            FillTransform( out, 0, shadowOffset.y, width, height, transformId );
```
ihhub, line 2573: Shouldn't we fill only the area which is an offset from the original image?

Districh-ru: Hi, @ihhub, this function only makes a shadow of the whole image. For double-layer image it goes through all the image and makes sets `transformId` for all non-transparent pixels. We should do the same for the single-layer image to be consistent. This image is fully non-transparent so it is faster to use `FillTransofrm()` instead of setting every pixel in a loop, like we doo for double-layer images.
It is obvious that only the offset area from the original image will be needed when we add a shadow to the the original image.
But this function is named `makeShadow()` and should return the shadow for the whole image. The made shadow might be used without rendering the original sprite above it (curently it is not used, but such use is possible).
We can change the `addShadow()` function not to call this one and to add shadow only where it is needed, like we do in `addGradientShadow()`.
But let's do it in a new PR later (if it will be needed).

ihhub (approved):

ihhub:
@Districh-ru , thank you very much for the improvement!

PR #8204, opened 2023-12-29
Small memory optimizations

ihhub:
- restructure few structures / classes to reduce padding. Maps::Tiles size was reduced from 96 to 88 bytes so it saves 288 KB on XL maps for 64-bit builds
- use single write operation for writing a string into a stream

### DIFF
--- a/src/engine/serialize.cpp
+++ b/src/engine/serialize.cpp
@@ -191,7 +191,7 @@ StreamBase & StreamBase::operator<<( const std::string & v )
 {
     put32( static_cast<uint32_t>( v.size() ) );
     // A string is a container of bytes so it doesn't matter which endianess is being used.
-    putRaw( v.c_str(), v.size() );
+    putRaw( v.data(), v.size() );
 
     return *this;
 }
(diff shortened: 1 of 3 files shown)


src/engine/serialize.cpp
```diff
@@ -190,9 +190,8 @@ StreamBase & StreamBase::operator<<( const uint32_t v )
 StreamBase & StreamBase::operator<<( const std::string & v )
 {
     put32( static_cast<uint32_t>( v.size() ) );
-
-    for ( std::string::const_iterator it = v.begin(); it != v.end(); ++it )
-        put8( *it );
+    // A string is a container of bytes so it doesn't matter which endianess is being used.
+    putRaw( v.c_str(), v.size() );
```
oleg-derevenetz, line 194: Although this is technically correct, and `c_str()` and `data()` are guaranteed to be the same since C++11, I would use `data()` here instead, because using `c_str()` may give the (wrong) impression that the stored string cannot contain null characters inside itself. But actually it's more of a quibble.

oleg-derevenetz (approved):
Hi @ihhub I just left one optional comment, LGTM otherwise.

src/engine/serialize.cpp
ihhub: I've changed it.

PR #8210, opened 2023-12-29
Fix several SonarQube smells

ihhub:
The easiest ones to fix.

### DIFF
--- a/src/engine/translations.cpp
+++ b/src/engine/translations.cpp
@@ -196,7 +196,7 @@ namespace
         return pos ? ++pos : str;
     }
 
-    struct Mofile
+    struct MOFile
     {
         // TODO: plural forms are not in use: Plural-Forms.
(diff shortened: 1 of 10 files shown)


src/engine/translations.cpp
```diff
@@ -196,18 +196,18 @@ namespace
         return pos ? ++pos : str;
     }
 
-    struct mofile
+    struct MOFile
```
oleg-derevenetz, line 199: MO is an acronym for "Machine Object", in contrast to PO, which is "Portable Object", so `MOFile` is sort of an acronym too and not the proper noun and should not be written as `Mofile` IMHO :)

oleg-derevenetz (approved):

src/engine/translations.cpp
ihhub: Sure!

PR #8212, opened 2023-12-31
AI: implement the logic to transfer the slowest troops to the garrison at the end of the turn to try to get a movement bonus on the next turn

### LINKED ISSUE #8207: Teach AI to garrison slower troops in the castle at the end of turn
### Preliminary checks

Currently when the AI ends the turn and it has a hero visiting a castle, it will not garrison any troops, which means that the next day their movement points will be based on a potentially slow troop in that army. Instead, if it would garrison these troops, it would get a single-turn movement points boost that can be good for many reasons, like having a greater chance to reach nearby objects and get back to the castle in one turn, among many other things.

Additionally, since by default you get more detailed information about troops in a hero's army than a castle's garrison, not garrisoning is suboptimal. One exception being the times when you own 2 thieves guilds. Therefore, garrisoning troops would then improve this by hiding what troops the AI has to the human player.

This is an example of the hero visiting the castle and you can see that they have garrisoned 0 troops, keeping the very slow Golems in their army. The best way here would be to only keep the very fast boars in their army.

image
image

Teach the AI to garrison troops by leaving only the fastest troops in its army, whether that is a single troop type or several. The AI will need to garrison troops in the corresponding castle slot they would have wanted the troops to be in their army in case they get attacked.

I suggest to avoid making a very complex logic for it, since even just a simple one would bring big benefits. For example, there can be special cases where you could optimize more, like if your castle has ultra fast troops that you don't intend to keep in your army the next day, but you know that you are in no threat to be attacked, so you move them to your army just to get the movement points, and replace them with desirable troops the next day. However, I suggest not diving into such cases - They are few, and just doing the basic variant of this optimization will be much better than the average heroes 2 player from my experience (I've never seen a streamer do or be aware of this trick).

Ideally, the AI should keep the one, or the several troops, with the highest speed in its army, and "move up" in the garrison the troops that are slower, starting with the slowest one in case the castle has slots filled with creatures already.

Note that this should only be done at the end of the player turn for every hero the AI currently has in a castle, not for every visit to a castle. I suppose it would even be beneficial for heroes that are permanently stationed in a castle for the reason of concealing its actual army.

### Additional info

Relates to this https://github.com/ihhub/fheroes2/issues/6755

oleg-derevenetz:
close #8207

Demo video:

https://github.com/ihhub/fheroes2/assets/32623900/dc1d3998-a340-4eab-ab07-5c9943f98ff3

In the `master` branch AI hero has not enough movement points to attack my hero after dismissing the Dragons:

https://github.com/ihhub/fheroes2/assets/32623900/9d076999-c7c2-4f62-a0a9-f5ae4c1806c8

### DIFF
--- a/src/fheroes2/ai/ai.h
+++ b/src/fheroes2/ai/ai.h
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2023                                             *
+ *   Copyright (C) 2019 - 2024                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2010 by Andrey Afletdinov <fheroes2@gmail.com>          *
--- a/src/fheroes2/ai/ai.h
+++ b/src/fheroes2/ai/ai.h
@@ -86,6 +86,8 @@ namespace AI
     class Base
     {
     public:
+        virtual ~Base() = default;
+
         virtual void KingdomTurn( Kingdom & kingdom ) = 0;
         virtual void BattleTurn( Battle::Arena & arena, const Battle::Unit & unit, Battle::Actions & actions ) = 0;
 
@@ -118,8 +120,6 @@ namespace AI
         // Should be called at the beginning of the battle even if no AI-controlled players are
         // involved in the battle - because of the possibility of using instant or auto battle
         virtual void battleBegins() = 0;
-
-        virtual ~Base() = default;
 
         virtual void tradingPostVisitEvent( Kingdom & kingdom ) = 0;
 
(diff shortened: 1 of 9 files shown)


oleg-derevenetz:
Hi @zenseii

> This PR did disclose some behavior which some might consider "cheating" by the AI, and I'm not saying it has to be fixed here, hence the approval.

Please note that there are [different rules](https://github.com/ihhub/fheroes2/pull/6741#issuecomment-1454322499) for army merging in case of human-controlled and AI-controlled castles. It's been like this since the beginning.

oleg-derevenetz:
> Ah, yes I had a vague memory of discussing this before. In this case I have nothing more to add and it works as it should right now.

Maybe someone will call this "cheating" and perhaps different rules for AI and human player can even be called such to some extent, BUT I should note that this was the easiest way to strengthen AI during the castle defense.

oleg-derevenetz:
> This is luckily not the case in fheroes2 as far as I'm aware.

Yes, in fheroes2 AI army arrangement is not dependent on the composition of an enemy army, only on the composition on the AI army itself.

ihhub (approved):

ihhub:

PR #8214, opened 2023-12-31
Implement warning message in the Editor

ihhub:
relates to #6845

image

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -62,6 +62,7 @@
 #include "ui_button.h"
 #include "ui_dialog.h"
 #include "ui_map_object.h"
+#include "ui_text.h"
 #include "ui_tool.h"
 #include "view_world.h"
 #include "world.h"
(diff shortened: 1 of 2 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -187,6 +188,13 @@ namespace Interface
             // Render all except the fog.
             _gameArea.Redraw( display, LEVEL_OBJECTS | LEVEL_HEROES | LEVEL_ROUTES );
 
+            if ( _warningMessage.isValid() ) {
+                const fheroes2::Rect & roi = _gameArea.GetROI();
+
+                const fheroes2::Text text{ _warningMessage.message(), fheroes2::FontType::normalWhite() };
+                text.draw( roi.x + 4, roi.y + roi.height - text.height() - 4, display );
```
Districh-ru, line 195: The text might be longer than the width of `_gameArea` (especially the translated text). For this case we should use multi-line text, but currently multi-line text is always canter aligned. So we can crop the long messages and place "..." at the end. Or leave it for now and later implement different text alignment for multi-line texts and use multi-line text here with left alignment.

ihhub: Let's go with text truncation for now.

Branikolog (approved):
Hi, @ihhub 

New text string works nice! 👍 

There're two not related issues with UI:

1)
image

2)
image

Districh-ru (approved):

PR #8215, opened 2024-01-01
Small improvements over image loading

ihhub:
- speed up TIL images. Previously most of images were made as 2-layered images causing extra operations being executed. Also, rearranging loops improves CPU caching.
- change the ICN image validation conditions. ICN::UNKNOWN shouldn't be even called.
- `GetModesSprite()` function should return a const reference instead of making a copy of an image
- reduce the scope of ICN images


Districh-ru (approved):
👍 
Soon I'll update `Fill()` to create a new image with the number of layers equal to the input image. And also the other functions in **image.cpp** that creates a new image if it is needed.

PR #8216, opened 2024-01-01
Mark fheroes2 as DPI-aware on Windows

### LINKED ISSUE #8209: Incorrect scaling when Windows DPI is set to other than 100%
### Preliminary checks

### Platform

Windows

### Describe the bug

On Windows 11 (I haven't tested others), I have set the DPI setting to 150% and this leads to fheroes2 creating a window which is 50% larger than the selected resolution in the app. This leads to non-integer scaled pixels even in the integer scaled resolutions (i.e. 640x480 (2x) ).

In fact on my Intel iGPU PC there is no resolution available that gives integer scaled pixels with the DPI set to 150%, which is the default DPI set by the OS for this PC, and there is no simple way of adding custom resolutions.

This is true for both Full Screen and Windowed mode.

I have Software Cursor set to on.

Looking around I see that others have noticed the same issue with SDL/Windows, perhaps this one suggesting to use `SDL_GetDisplayDPI()` could be a solution:
https://www.reddit.com/r/sdl/comments/115p1ft/comment/j937j67/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

I'm aware that this is not directly a fheroes2 bug, but it might be worth investigating to make the user experience better.

### Save file

No save file needed.

### Additional info

This SDL repo issue talks about what approach to use for SDL3:
https://github.com/libsdl-org/SDL/issues/7134

This is a discussion related to SDL scaling and how MacOS uses points not pixels and potential problems with that, I believe we have no such issues right now for MacOS as none have reported it.
https://discourse.libsdl.org/t/high-dpi-mode/34411/2

Slightly related to https://github.com/ihhub/fheroes2/issues/5250 and the connected PRs

oleg-derevenetz:
close #8209

According to my observations, this fix works just fine in case of 1 monitor and _almost_ fine in case of multiple monitors with different scale - sometimes, when transferring the app window between monitors having a different scale, the window size suddenly changes (in the direction of the monitor to which the transfer is carried out - that is, if the window is transferred from a monitor with a larger scale to a monitor with a smaller scale, then the window becomes smaller in size, and vice versa). This happens _sometimes_ and I was not able to find some kind of pattern. Probably this is due to some of pitfalls described here:

https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows?redirectedfrom=MSDN#common-pitfalls-win32

but since fheroes2 do not handle Windows messages by itself (this is all done inside the SDL), to be honest, I don't want to try to make some potential SDL fixes yet (it's probably already fixed or has yet to be fixed in SDL3), so the possibility of merging this PR is up for discussion.

### DIFF
--- a/VisualStudio/fheroes2/common.props
+++ b/VisualStudio/fheroes2/common.props
@@ -5,6 +5,9 @@
       <AdditionalIncludeDirectories>src\engine;src\fheroes2\gui;src\fheroes2\maps;src\fheroes2\kingdom;src\fheroes2\campaign;src\fheroes2\game;src\fheroes2\editor;src\fheroes2\dialog;src\fheroes2\system;src\fheroes2\spell;src\fheroes2\monster;src\fheroes2\castle;src\fheroes2\agg;src\fheroes2\heroes;src\fheroes2\resource;src\fheroes2\ai;src\fheroes2\army;src\fheroes2\battle;src\fheroes2\h2d;src\fheroes2\objects;src\fheroes2\world;src\fheroes2\audio;src\fheroes2\image;src\thirdparty\libsmacker;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions Condition="'$(FHEROES2_BUILD_NUMBER)'!=''">BUILD_VERSION=$(FHEROES2_BUILD_NUMBER);%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
+    <Manifest>
+      <AdditionalManifestFiles>src\resources\fheroes2.manifest %(AdditionalManifestFiles)</AdditionalManifestFiles>
+    </Manifest>
     <ResourceCompile>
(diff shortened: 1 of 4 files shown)


oleg-derevenetz:
> The difference between my problem and the multi-DPI one is that in my situation an end-user wouldn't have gotten any feedback telling them that the game shouldn't be so blurry

Also if DPI awareness is not enabled, OS performs all the scaling, so even if user will use in-game video settings to change resolutions, say, to 800x600 and back to 640x480, this will not work - according to fheroes2 understanding, resolution is 640x480, while in fact with 1.5 multiplier it's 960x720. If DPI awareness is enabled, fheroes2 works not with "virtual", but with "real" pixels, so in this case in-game resolution change will work just fine. I don't really understand why the window size changes in rare cases during the app window transfer between monitors with different multipliers, but it's almost certainly due to some kind of malfunction inside SDL.

ihhub (approved):

ihhub:

PR #8218, opened 2024-01-01
Make the output of Flip() to have the same layers as the input image

Districh-ru:


ihhub (approved):

ihhub:
@Districh-ru , thanks a lot for the improvement!

Issue #8221, opened 2024-01-02
Add description of how much gold was paid to surrender in battle result dialog


zenseii:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Neither the original game nor fheroes2 tells you how much gold you got from your opponent surrendering. I suggest to add this in the battle result dialog. It is quite possible that the height of this dialog will need to be extended in this case. This can be achieved with the `StandardWindow` object.

Something like:

`%{heroName} accepts the %{goldAmount} gold and allows %{surrenderingHeroName} and their army to surrender.`

Initial description about teaching AI to surrender, which now has been implemented:

<details>

When defending a castle, the only way to escape with a hero is by surrendering, because retreat is disabled.

It is very often worth it to surrender if you only have a few troops left and it would only cost you a few tens or hundreds of gold. This would avoid granting a winning attacker 500 extra EXP for taking out a hero, giving them any artifacts, and you losing a good hero.

The logic shouldn't be very different from how the current retreat logic works, except it would need to take into account whether the gold payment is worth it. I suppose if the AI has an overabundance of gold compared to the cost of surrendering, then surrendering is fine.

Unless it doesn't require complex logic, I suggest not teaching the AI to use the fheroes2 specific feature of using the market place to get enough gold to surrender. I'm also not sure this feature is very good, especially since you get to use it when attacked outside of your own turn, but that is a discussion for another issue or discussion thread, hence why I suggest not looking much into letting the AI do this right now.

Note that I specifically restricted this to sieges, because for normal battles outside castles, the logic behind choosing to surrender compared to retreating is much more complex.

Secondly, note that the way surrendering in the original Heroes 2, and naturally the way it works in fheroes2 right now, is that the non-surrendering party does not have anything to say whether the opponent gets to surrender or not - it is essentially a retreat paid by gold (plus you get to keep the surviving creatures).

**I strongly suggest** that any discussions about the way surrendering works be done in a separate discussion, as this issue is only about giving the AI the ability to do what the human player does right now - the AI does not get to refuse your surrender so neither should you be allowed to refuse them.

 

### Describe the possible solution

Teach the AI surrender when defending in a castle siege like it would have retreated in a regular battle.

### Additional info

[...]

PR #8222, opened 2024-01-02
Allow AI heroes to surrender if the conditions of retreat have been met, but it is impossible to retreat

oleg-derevenetz:
Related to #8221

In this video the pumped-up AI-controlled hero surrenders (after a farewell spellcast), and is rehired by AI on the next turn:

https://github.com/ihhub/fheroes2/assets/32623900/93ab4a32-ad8f-423f-a090-de42539ed3e3

I did not introduce any special coefficients for the necessary gold reserve, because currently, when AI is going to retreat, it already has a little army, and the value of a pumped hero, and even with artifacts, in any case exceeds the value of those remnants of his army at the time of surrender.

This PR does not address a suggestion to improve the UI regarding the display of the amount of gold received due to surrender.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2020 - 2023                                             *
+ *   Copyright (C) 2020 - 2024                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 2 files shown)


oleg-derevenetz:
> I checked the original game and the 500 EXP bonus for winning a siege when the defending hero surrenders should in fact be there.

Yeah, it seems that attacker gains extra 500 XP if opposing hero surrenders, BUT only if it's a siege. Attacker doesn't get these 500 XP if the battle takes place in an open field.

oleg-derevenetz:
> EDIT: I've looked everywhere on the repository and can't find mention of this at all. I can only assume this is another missing feature of the OG/bug.

There was no bonus for a successful siege from the very beginning, there was only 500XP bonus for defeating the hero, and that's all. Anyway, I just added this independent bonus, so attacking hero will always get 500XP for a successful siege and additional 500XP for defeating an enemy hero (if this hero did not retreat or surrender).

oleg-derevenetz:
> I checked the case when there's a captain, and you shouldn't get 500XP for that:

Well, then, in this regard, the original logic from previous team was also wrong.

ihhub (approved):

ihhub:

PR #8235, opened 2024-01-05
Improve AI heroes behavior on the Adventure Map

ihhub:
- speed up decision making regarding Hut of Magi
- add correction coefficient for object value calculation that depends on time. The longer AI heroes play the lesser distance penalty.
- adjust fog discovery priorities for different AI hero roles
- add fog discovery value correction over time for AI heroes. In other words AI heroes will tend to uncover more fog over time.
- ~~adjust Observation Tower evaluation to exclude fog that can be uncover just by hero stepping on the object~~
- speed up Stables evaluation
- ~~remove hardcoded bonus values for Daemon Cave~~
- avoid visiting Oasis and Watering Hole if they bring no benefit
- improve picking correct tile index for fog discovery. Previously, AI heroes will choose any first tile that allows to uncover fog. Now the search continues to check all neighbouring tiles to find the best one.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2020 - 2023                                             *
+ *   Copyright (C) 2020 - 2024                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -334,11 +334,10 @@ namespace
         }
 
         case MP2::OBJ_OBSERVATION_TOWER: {
-            const int32_t fogCountToUncoverByHero = Maps::getFogTileCountToBeRevealed( index, hero.GetScoutingDistance(), hero.GetColor() );
             const int32_t fogCountToUncoverByTower
                 = Maps::getFogTileCountToBeRevealed( index, GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::OBSERVATION_TOWER ), hero.GetColor() );
 
-            return fogCountToUncoverByTower > fogCountToUncoverByHero;
+            return fogCountToUncoverByTower > 0;
         }
 
         case MP2::OBJ_OBELISK:
@@ -846,7 +845,7 @@ namespace
         return fogDiscoveryBaseValue;
     }
 
-    uint32_t getFogDiscoveryMaximizingPeriod( const Heroes & hero )
+    uint32_t getTimeoutBeforeFogDiscoveryIntensification( const Heroes & hero )
     {
         switch ( hero.getAIRole() ) {
         case Heroes::Role::SCOUT:
@@ -1245,13 +1244,11 @@ namespace AI
             const int32_t fogCountToUncoverByTower
                 = Maps::getFogTileCountToBeRevealed( index, GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::OBSERVATION_TOWER ), hero.GetColor() );
 
-            const int32_t fogCountToUncoverByHero = Maps::getFogTileCountToBeRevealed( index, hero.GetScoutingDistance(), hero.GetColor() );
-
-            if ( fogCountToUncoverByTower - fogCountToUncoverByHero <= 0 ) {
+            if ( fogCountToUncoverByTower == 0 ) {
                 // Nothing to uncover.
                 return -dangerousTaskPenalty;
             }
-            return fogCountToUncoverByTower - fogCountToUncoverByHero;
+            return fogCountToUncoverByTower;
         }
         case MP2::OBJ_MAGELLANS_MAPS: {
             // Very valuable object.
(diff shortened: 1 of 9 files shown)


oleg-derevenetz:
Hi @ihhub

> remove hardcoded bonus values for Daemon Cave

Daemon Cave works as follows: you get a choice between fighting the demon's servants and the demon himself. In the first case, you get a fixed reward, in the second you will get a random outcome (including the outcome when hero will be eaten by a demon) with a random reward (by "random" I mean randomly set at the start of the game on the map). This random reward can be retrieved using the `getFundsFromTile()`, but not always - sometimes it's an artifact and in this case there is no sense to call `getFundsFromTile()`, and sometimes it's just experience and so on. Anyway, AI always choose to fight with demon servants and to get a fixed reward (which is hardcoded), because It doesn't want its hero to be eaten by a demon randomly.

ihhub:
Hi @oleg-derevenetz , I reverted this change.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -1207,13 +1242,16 @@ namespace AI
...
         case MP2::OBJ_OBSERVATION_TOWER: {
-            const int fogCountToUncover
+            const int32_t fogCountToUncoverByTower
                 = Maps::getFogTileCountToBeRevealed( index, GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::OBSERVATION_TOWER ), hero.GetColor() );
-            if ( fogCountToUncover <= 0 ) {
+
+            const int32_t fogCountToUncoverByHero = Maps::getFogTileCountToBeRevealed( index, hero.GetScoutingDistance(), hero.GetColor() );
+
+            if ( fogCountToUncoverByTower - fogCountToUncoverByHero <= 0 ) {
                 // Nothing to uncover.
```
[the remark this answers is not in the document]
oleg-derevenetz: Actually, if we think about it, action objects are usually not used by a hero during the regular fog discovery, therefore, if the hero does not see more on this tower than without it, then it is still worth using this tower for the fog discovery.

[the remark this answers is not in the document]
ihhub: I revert the changes.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -916,6 +916,8 @@ int AIWorldPathfinder::getFogDiscoveryTile( const Heroes & hero, bool & isTerrit
             }
         }
 
+        int32_t maxTilesToReveal = 0;
```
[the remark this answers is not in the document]
ihhub: How it was previously: iteratively go through every tile, check all neighbours and if any neighbour has fog to uncover, stop there and return this neighbour tile index.

Now: iteratively go through every tile, check all neighbours and if any neighbour has fog to uncover, remember this neighbour, continue through all other neighbours and find the one with highest number of tiles to uncover. Return the best.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2095,10 +2143,28 @@ namespace AI
...
+                    // This is actually a very useful fog discovery action which might lead to finding of new objects.
+                    // Increase the value of this action.
+                    fogDiscoveryValue /= 2;
+
+                    if ( world.CountDay() > period ) {
+                        fogDiscoveryValue = 0;
+                    }
+                    else {
+                        fogDiscoveryValue = fogDiscoveryValue * ( period - world.CountDay() ) / period;
+                    }
```
Districh-ru, line 2161: For the case when `world.CountDay() > period` the `/= 2` is redundant because the variable will be set to 0.
So we can slightly increase the performance:

```suggestion
                    if ( world.CountDay() > period ) {
                        fogDiscoveryValue = 0;
                    }
                    else {
                        // Increase the value of this action.
                        fogDiscoveryValue /= 2;
                        
                        fogDiscoveryValue = fogDiscoveryValue * ( period - world.CountDay() ) / period;
                    }
```

Districh-ru (changes requested):
Hi, @ihhub, I let AI to play an XL map and fond the AI movement actions good.
I left one minor speed-up suggestion. Could you please take a look.

src/fheroes2/ai/normal/ai_normal_hero.cpp
oleg-derevenetz: @Districh-ru  If so, there is no need to divide it by 2 first and then perform separate calculations, we can just write like this:

```cpp
else {
    fogDiscoveryValue = fogDiscoveryValue / 2 * ( period - world.CountDay() ) / period;
}
```

Districh-ru: Sure, I just left the division with the comment why the division is done from the previous code.

src/fheroes2/world/world_pathfinding.cpp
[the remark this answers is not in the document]
oleg-derevenetz: I pushed commit with my vision of `AIWorldPathfinder::getFogDiscoveryTile()`. It is somewhat simpler, it calls `Maps::getFogTileCountToBeRevealed()` more times, but this is compensated by not calling many other things. Here is a profiler's call tree for the `master` branch:

master

and here is a call tree for this PR with my commit:

pr

Both are made from the same savefile made on XL map "Marooned" at 6-3-1 (that is, a sufficiently explored map by AI at that moment). I suggest testing this variant, but I don't insist :) This commit can be reverted.

idshibanov (approved):
Great change

Branikolog (approved):
Haven't found any issues.

src/fheroes2/world/world_pathfinding.cpp
```diff
@@ -876,114 +875,110 @@ uint32_t AIWorldPathfinder::getMovementPenalty( const int from, const int to, co
...
     const int scoutingDistance = hero.GetScoutingDistance();
-
     const Directions & directions = Direction::All();
-    std::vector<bool> tilesVisited( world.getSize(), false );
-    std::vector<int> nodesToExplore;
-
-    tilesVisited[start] = true;
 
-    nodesToExplore.push_back( start );
+    struct TileCharacteristics
```
ihhub, line 885: Hi @oleg-derevenetz , what was the reason to remove whirlpool and teleport tiles?

ihhub: Got it.

PR #8236, opened 2024-01-05
Fix a SonarQube smell

zenseii:


ihhub (approved):

ihhub:
@zenseii , thank you very much for the fix!

PR #8240, opened 2024-01-07
Adjust text in the skill popup dialog

JasonQorbin:
- Made the skill name and skill level text centre properly.
- Also truncate the text with '...' if the text is too long for the icon (happens in some languages). The full, unmodified skill name and skill level text can still be seen at the top of the popup dialog.
- Lastly limited the text to only draw within the icon boundary so that in edge cases the text shadow doesn't fall on the icon's gold border.

Related to issue #7491

Screenshot 2024-01-07 095034


src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -694,13 +694,23 @@ namespace fheroes2
         Blit( background, 0, 0, output, offset.x, offset.y, background.width(), background.height() );
 
         const Sprite & icn = AGG::GetICN( ICN::SECSKILL, _skill.GetIndexSprite1() );
-        Blit( icn, 0, 0, output, offset.x + ( background.width() - icn.width() ) / 2, offset.y + ( background.height() - icn.height() ) / 2, icn.width(), icn.height() );
+        const int32_t iconOffsetX = offset.x + ( background.width() - icn.width() ) / 2;
+        const int32_t iconOffsetY = offset.y + ( background.height() - icn.height() ) / 2;
+        Blit( icn, 0, 0, output, iconOffsetX, iconOffsetY, icn.width(), icn.height() );
+
+        const fheroes2::Rect textBoundary( iconOffsetX, iconOffsetY, icn.width(), icn.height() );
```
Districh-ru, line 701: Since we are rendering the skill `icn` image and want to render text only in its boundaries we can first set the `Rect` for `icn` (to cache all needed values) and then use it everywhere. Also the skill image has no transparency or shadows so we can use `Copy()` here instead of `Blit()`:
```cpp
const fheroes2::Rect icnRect( offset.x + ( background.width() - icn.width() ) / 2, offset.y + ( background.height() - icn.height() ) / 2, icn.width(), icn.height() );
Copy( icn, 0, 0, output, icnRect );
```

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -694,13 +694,23 @@ namespace fheroes2
...
 
-        const Text skillName( Skill::Secondary::String( _skill.Skill() ), FontType::smallWhite() );
-        skillName.draw( offset.x + ( background.width() - skillName.width() ) / 2, offset.y + 8, output );
+        Text skillName( Skill::Secondary::String( _skill.Skill() ), FontType::smallWhite() );
+        if ( skillName.width() > textBoundary.width ) {
+            skillName.fitToOneRow( textBoundary.width );
+        }
+        skillName.drawInRoi( offset.x + ( background.width() - skillName.width() ) / 2 - 1, offset.y + 8, output, textBoundary );
```
Districh-ru, line 707: We can skip the text `width()` check because it is done in `fitToOneRow()` method.
Also as we want to render text only in the given `textBoundary` (or we can name it `icnRect` because it is a rectangle where `icn` is rendered) we the more proper way is to do text offsets relative to this Rect:
```cpp
skillName.fitToOneRow( icnRect.width );
skillName.drawInRoi( icnRect.x + ( icnRect.width - skillName.width() ) / 2, icnRect.y + 5, output, icnRect );
```

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -694,13 +694,23 @@ namespace fheroes2
...
 
-        const Text skillDescription( Skill::Level::StringWithBonus( _hero, _skill ), FontType::smallWhite() );
-        skillDescription.draw( offset.x + ( background.width() - skillDescription.width() ) / 2, offset.y + 56, output );
+        Text skillDescription( Skill::Level::StringWithBonus( _hero, _skill ), FontType::smallWhite() );
+        if ( skillDescription.width() > textBoundary.width ) {
+            skillDescription.fitToOneRow( textBoundary.width );
+        }
+        skillDescription.drawInRoi( offset.x + ( background.width() - skillDescription.width() ) / 2 - 1, offset.y + 56, output, textBoundary );
```
Districh-ru, line 713: The same comment as previous:
```cpp
skillDescription.fitToOneRow( icnRect.width );
skillDescription.drawInRoi( icnRect.x + ( icnRect.width - skillDescription.width() ) / 2, icnRect.y + 53, output, icnRect );
```

Districh-ru (changes requested):
Hi, @JasonQorbin. Many thanks for your contribution to the project!
 I left three suggestions. Could you please check them when you have time?

Districh-ru (approved):
👍

ihhub (approved):

ihhub:
@JasonQorbin , thank you so much for this fix!

PR #8241, opened 2024-01-07
Do not allocate memory to transform layer if image is already set as single-layer

Districh-ru:
This PR contains changes from #8149 and should be merged after (or instead of) it.

This PR twice reduces the size of the allocated memory for single-layer images (if a single-layer state is set before the memory allocation for the image) during the `resize()` and `copy()`.

The fheroes2 GUI often uses image restores which are single-layer images, but for every such restorer the engine also allocates memory for the never used transform-layer. Such operations are also done when generating big View World image and even for the single-layer screen surface.

NOTE: if already existing double-layer image is made single-layer the transform layer stays untouched: memory reallocation for this case is not done not to waste CPU resources.


ihhub:
Hi @Districh-ru , I actually like the changes but I am still concerned of forcing single layer images to have only have of the array as we might still miss some places in the code where we use single-layered images.

What do you think if we merge all changes in the code except the ones made in Image class and possibly in screen.cpp?

Districh-ru:
> Hi @Districh-ru , I actually like the changes but I am still concerned of forcing single layer images to have only have of the array as we might still miss some places in the code where we use single-layered images.
> 
> What do you think if we merge all changes in the code except the ones made in Image class and possibly in screen.cpp?

Hi @ihhub, we have #8149 that only removes writes to the transform layer for the single-layer images, but keeps memory allocation for it.
Or you mean that we should first apply all other changes including new assertions? I'll prepare a PR a little later with such changes.

But single-layer images are single-layer and if somewhere we access the transform layer for such images then it is wrong. And we should fix that places instead. Keeping the transform layer for such images just as a patch for cases when some function accidentally access the transform layer IMHO is not good. :)

ihhub:
Hi @Districh-ru , do we need any changes from here after we merged #8149?

Districh-ru:
> Hi @Districh-ru , do we need any changes from here after we merged #8149?

Hi @ihhub, no.

I'm closing this PR.

PR #8250, opened 2024-01-09
Implement Mountains, Rocks and Trees placing in Editor

Districh-ru:
This PR adds Mountains, Rocks and Trees objects placing in Editor.
The object selection dialog will be improved later.

mountains

forests

rocks

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -1131,11 +1131,20 @@ int Dialog::selectDwellingType( const int dwellingType )
     return selectObjectType( dwellingType, objectInfo.size(), listbox );
 }
 
-int Dialog::selectMountainType( const int dwellingType )
+int Dialog::selectMountainType( const int mountainType )
 {
     const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::LANDSCAPE_MOUNTAINS );
 
     DwellingTypeSelection listbox( objectInfo, { 420, fheroes2::Display::instance().height() - 180 }, _( "Select Mountain Object:" ) );
 
-    return selectObjectType( dwellingType, objectInfo.size(), listbox );
+    return selectObjectType( mountainType, objectInfo.size(), listbox );
+}
+
+int Dialog::selectTreeType( const int treeType )
+{
+    const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::LANDSCAPE_TREES );
+
+    DwellingTypeSelection listbox( objectInfo, { 420, fheroes2::Display::instance().height() - 180 }, _( "Select Tree Object:" ) );
+
+    return selectObjectType( treeType, objectInfo.size(), listbox );
 }
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -1544,6 +1544,15 @@ int Dialog::selectMountainType( const int mountainType )
     return selectObjectType( mountainType, objectInfo.size(), listbox );
 }
 
+int Dialog::selectRockType( const int rockType )
+{
+    const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::LANDSCAPE_ROCKS );
+
+    DwellingTypeSelection listbox( objectInfo, { 420, fheroes2::Display::instance().height() - 180 }, _( "Select Rock Object:" ) );
+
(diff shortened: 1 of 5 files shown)


src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -1534,3 +1534,30 @@ void Dialog::selectMineType( int32_t & type, int32_t & color )
         display.render( area );
     }
 }
+
+int Dialog::selectMountainType( const int mountainType )
+{
+    const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::LANDSCAPE_MOUNTAINS );
+
+    DwellingTypeSelection listbox( objectInfo, { 420, fheroes2::Display::instance().height() - 180 }, _( "Select Mountain Object:" ) );
```
ihhub, line 1542: Since we use the same list for all these types should we rename it to `GenericObjectTypeSelection`?

ihhub (commented):
Hi @Districh-ru , in general everything looks good. I left one question here. Could you please check it?

Branikolog:
Hi, @Districh-ru 
Is it intended that stumps belong to the "Trees" category?
The "Miscellaneous" category has stumps drawn on the icon, so it's rather confusing, this object is included in different category. Or is this just a problem of not well fitting icon for miscellaneous?
image

src/fheroes2/dialog/dialog_selectitems.cpp
Districh-ru: Sure, lets do it.

Districh-ru:
> Hi, @Districh-ru Is it intended that stumps belong to the "Trees" category? The "Miscellaneous" category has stumps drawn on the icon, so it's rather confusing, this object is included in different category. Or is this just a problem of not well fitting icon for miscellaneous? 

Hi, @Branikolog, all these icons are temporary and we may change them while populating new objects.
The stumps are cut trees (and in original game and editor are called "Dead Tree") to IMHO they belong to Trees category.
For the Miscellaneous objects a new icon will be used in a new PR:

изображение

ihhub (approved):

Branikolog (approved):
Nice work!
I have only concerns about one tree object passability, but it fully replicates the OG. Maybe those tiny difference in trees height matters in the way hero can pass around it.
image

Anyway, have no other questions regarding anything in this PR. 👍

ihhub:
@Districh-ru , many thanks for this amazing work!

Districh-ru:
> Nice work! I have only concerns about one tree object passability, but it fully replicates the OG. Maybe those tiny difference in trees height matters in the way hero can pass around it. image
> 
> Anyway, have no other questions regarding anything in this PR. 👍

@Branikolog, as I noticed the 1 tile height objects have different passability vs 2 tile high objects. So sometime almost the same objects have a little different passability:
изображение

We can discuss the passability improvements later. But IMHO we should to keep all passabilities for the original maps. And it wouldn't be good if we have different passability for the same objects for different map types. I may confuse players.

Branikolog:
Hi, @Districh-ru 

>We can discuss the passability improvements later.

Sure!

>But IMHO we should to keep all passabilities for the original maps. And it wouldn't be good if we have different passability for the same objects for different map types. I may confuse players.

I agree, but me personally and many other players (I believe) expect the same passability for the objects, that are 95% identical visually. 

That orc hut height difference introduced in just a couple of pixels is barely noticeable, but passability is different. The same story with the trees from my screenshot above: they have identical size and looks, but the rules for passing are different. That bothers me more, than keeping the original rules.
I'm 99% sure, if you show those trees and huts even to an old and devoted H2 player - he will be sure, the passing rules are the same for these objects.
If I'm mistaken, and the original passing rules for these different objects are obvious and predictable for all players, I will gladly agree to keep it as it is. :) 
I understand, that rules work according to exact object height, but sometimes similar objects behave differently and that's not really user friendly, especially considering the amount of such similar objects. :)

Districh-ru:
I totally agree that the same objects should have the same passability.
But for upper the tile passability, fixed for the green hills, we can change the "layer" of the objects (is it on the ground or it is over the ground). For the trees and orc hut the passability can not be fixed by changing the object data in editor object presets. This passability is set by the engine: low (1 tile height) objects has different passability vs tall objects.
If we change such passabilities in the engine then it will also change the passability for such objects for all original maps. And doing this we need to be sure that we will not break any map made by the original editor.
So we we can not quickly do such change. :(

Branikolog:
@Districh-ru 

I'm not insisting on doing such changes at all or forcing them as fast as possible.
Just pointing that out for the future "friendliness" of the game and the editor. :)

PR #8253, opened 2024-01-10
Fixed accidental map scrolling after exiting castle/hero screen.

### LINKED ISSUE #2333: Adventure screen, left-clicking the "EXIT" button in the castle screen near the edge of the screen should not activate moving the adventure map with the "arrow" cursor
In fheroes2, left-clicking the "EXIT" button in the castle screen near the edge of the screen should not activate moving the adventure map with the "arrow" cursor.

Moving the adventure map with the "arrow" cursor should only be activated by moving the mouse to the edge of the screen from the adventure screen.

This flaw is also present in the original game.

Independently of this problem, it would also be nice to be able to set a delay before moving the adventure map with the "arrow" cursor (for example in the file "fheroes2.cfg").

In fheroes2:

Exit window near edge 01_0-96

dood-apo:
Added a flag to indicate that fast scroll is disabled, re-enabling it once the cursor leaves the borders.

Fix #2333


src/fheroes2/system/settings.h
```diff
@@ -220,6 +225,11 @@ class Settings
         game_difficulty = difficulty;
     }
 
+    void SetFastScroll( const bool enable )
```
ihhub, line 228: Settings class is not the place to store a temporary variable for runtime stuff. Please revert all changes in this file.

**GameArea** is the class responsible for scrolling. So this is the correct place to store some data. However, hero and castle screen windows can be open also from Kingdom Overview dialog so your changes most likely will break some logic in these cases.

We need to make sure that both screen are open only from Adventure Map and apply the conditions approapriately.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -1153,6 +1153,12 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
             fastScrollRepeatCount = 0;
         }
 
+        // re-enable fast scroll once the cursor leaves the border area
+        if ( !isScrollLeft( le.GetMouseCursor() ) && !isScrollRight( le.GetMouseCursor() ) && !isScrollTop( le.GetMouseCursor() )
+             && !isScrollBottom( le.GetMouseCursor() ) ) {
+            Settings::Get().SetFastScroll( true );
+        }
```
ihhub, line 1160: Please use a single Adventure Map ROI and check if the cursor within this area rather than doing 4 calls that are much more expensive to execute. Moreover, since the whole logic for stopping and allowing scrolling is done within this method why don't we store a needed variable here?

ihhub (changes requested):
Hi @dood-apo , I put 2 comments here. Could you please take a look at them?

src/fheroes2/system/settings.h
dood-apo: Migrated the fast scroll state handling to `GameArea` and reverted changes mage to `settings.h`.

Regarding the kingdom overview case - I don't think we need any additional logic. Fast scroll gets disabled upon clicking "exit" on any of the 3 screens: hero, castle, kingdom overview (that last case was missed in the initial PR). It gets reenabled once we are on the world map and the cursor is in any place on the screen except the fast scroll borders. So even if disabling fast scroll wasn't needed (we came to hero screen from overview) it doesn't hurt.

src/fheroes2/game/game_startgame.cpp
dood-apo: I have also limited the cursor checks to 2 from 4. Since the areas vulnerable to "exit into scroll" bug are right and bottom borders only, we don't need to worry about left and up.

dood-apo:
> Hi @dood-apo , I put 2 comments here. Could you please take a look at them?

Hey @ihhub. Addressed your comments and added some improvements. Waiting for the second round of the review.

src/fheroes2/castle/castle_dialog.cpp
```diff
@@ -404,6 +406,9 @@ Castle::CastleDialogReturnValue Castle::OpenDialog( const bool openConstructionW
             if ( le.MouseClickLeft( buttonExit.area() ) || Game::HotKeyCloseWindow() ) {
                 result = CastleDialogReturnValue::Close;
 
+                // disable fast scroll for resolutions where the exit button is directly above the border
```
ihhub, line 409: Please start all comments with a capital letter. It will be easier to read them :)

src/fheroes2/game/game_startgame.cpp
```diff
@@ -1153,6 +1153,11 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
             fastScrollRepeatCount = 0;
         }
 
+        // re-enable fast scroll once the cursor leaves the vulnerable parts of the border area
+        if ( !isScrollRight( le.GetMouseCursor() ) && !isScrollBottom( le.GetMouseCursor() ) ) {
```
ihhub, line 1157: I am not sure that we need to cover the whole right and bottom part of the Adventure Map. For instance, a player can just move the cursor to the middle of bottom part of the screen and they would expect scrolling to work. Should we take into an account the area of the exit button plus some extra area?

ihhub (changes requested):
Hi @dood-apo, I left several comment in this pull request. Can you please take a look at them?

src/fheroes2/game/game_startgame.cpp
dood-apo: Sure, I can take a look into it.

src/fheroes2/gui/interface_gamearea.h
```diff
@@ -290,6 +300,7 @@ namespace Interface
         bool _mouseDraggingInitiated;
         bool _mouseDraggingMovement;
         bool _needRedrawByMouseDragging;
+        bool _fastScroll;
```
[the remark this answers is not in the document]
dood-apo: `_isFastScrollEnabled` perhaps? We haven't used the form "scrolling" yet.

dood-apo:
Hey @ihhub , I'm ready for the next round.

src/fheroes2/gui/interface_base.h
```diff
@@ -110,6 +110,14 @@ namespace Interface
             return cursorPos.y >= display.height() - BORDERWIDTH;
         }
 
+        static bool isFastScrollVulnerableArea( const fheroes2::Point & cursorPos )
```
ihhub, line 113: What if we set the area within `setFastScrollStatus()` call (we should rename then the method) since we know the area for a button being pressed? In this case we will avoid having any hardcoded values.

ihhub: Expanding the idea: we can simply store `fheroes2::Rect` without the need of the boolean variable. Then instead of
```cpp
if ( scrollPosition != SCROLL_NONE && _gameArea.isFastScrollEnabled() ) {
```
we can do
```cpp
if ( scrollPosition != SCROLL_NONE && _gameArea.getNoScrollArea() != fheroes2::Rect() ) {
```
(a very bad name, we need to think about it).

Then the code:
```cpp
if ( !isFastScrollVulnerableArea( le.GetMouseCursor() ) ) {
    _gameArea.setFastScrollStatus( true );
}
```

will look like:
```cpp
if ( !( _gameArea.getNoScrollArea() & le.GetMouseCursor() ) ) {
    _gameArea.setNoScrollArea( {} );
}
```

dood-apo: Hey @ihhub , that's a very disciplined approach and is definitely quite elegant in defining the precise area. However by getting more precise we are shooting ourselves in the foot in my opinion.

For example the exit button on hero dialog is tall and thin and only barely overlaps with the right scroll area. If we implement it to care only about that overlapping area (which would be ~2 px wide!) then a minimal twitch to the right with the mouse will trigger scrolling. I think this is not what we want.

Now when I think more about it, I think I liked the most version with disabling bottom and right scroll borders completely. It is the most streamlined approach (easier to understand this behavior from the user perspective then the pixel perfect approach or the 0.6 and 0.8 multiplier approach). If you exit the dialog straight to the scroll area, then it does not trigger scrolling. You need to leave and re-enter the scroll area to enable it. This seems reasonable for me.

What do you think?

ihhub: Hi @dood-apo , hardcoding the solution to specific sides of a screen might not be the best solution. For instance, a player can have this resolution:
image

As you can see the EXIT button in castle window is very far from the edges. Then a player clicks on the button and tries to scroll using their Android device. The scrolling won't work. Therefore I proposed to use the button area as reference. We can simply expand it by 1 button height lower and by half of the button width to the right and to the left.

Please correct me if I am wrong.

dood-apo: Hey @ihhub , my understanding is that it wouldn't behave as you described, regardless if the user operates with the touchscreen or mouse.

In your scenario as soon as EXIT is clicked, this code in the main loop fires (because the EXIT is not above the borders, so is not in the area vulnerable to the fast scroll bug):
```
// Re-enable fast scroll once the cursor leaves the vulnerable parts of the border area.
if ( !isScrollRight( le.GetMouseCursor() ) && !isScrollBottom( le.GetMouseCursor() ) ) {
    _gameArea.setFastScrollStatus( true );
}
```

So the scrolling possibility is reenabled instantly in this case. I don't have an android device to test it, but I've verified on Windows using only touchscreen on my notebook.

> hardcoding the solution to specific sides of a screen might not be the best solution

I don't think we are hardcoding here. We're just saying:

1. If you've just left the dialog by clicking EXIT we are disabling the fast scroll feature (ability to hover over borders to trigger "classic" homm2 scroll). To be specific only disabling it on right and bottom borders, but that is just an optimization.
2. Back to the main loop: if the cursor position is **not** on those borders, we reenable the ability to use the fast scroll feature. In the example you have shown it will happen **immediately** so practically speaking the fast scroll would not be disabled at all.

ihhub:
Hi @Branikolog , can you please test this pull request and give your feedback as we have a debate which way is better :)

Branikolog:
Hi, @ihhub and @dood-apo 

From a usability point I personally (as a player) prefer scrolling being initiated after cursor just changed its position just a bit being already located on the scrolling edge, rather than coming back to the central area and returned to the edge again. There could be some cases, when player can close the castle window and just instantly wants to scroll to the right, but with the current implementation it doesn't work. So player needs to return cursor back and then slide to right once again.

https://github.com/ihhub/fheroes2/assets/55348946/76fc7b17-603c-4873-86c0-9ed086d6eaa3

If we allow to initiate scrolling just when cursor changes the position (for some predefined distance around 3-5pxs) then user can just slide to right and once cursor changes its position vertically (let's assume, that horizontal change is insufficient, as cursor could be too close to the edge) the scrolling proceeds. A human being could hardly perform a strict horizontal move with his arm, so scrolling would be inevitably initiated just after the cursor is going to be shifted vertically enough.
With the current implementation user slides to right, nothing happens, then he moves cursor out of the screen edge and slides to the edge again, which is uncomfortable in my opinion.

I'm also not sure, why do we limit such logic only for two borders?
The same approach I suggest implementing for all screen edges, as player can exit the castle with a hot key when cursor is too close to the one of all 4 edges:

https://github.com/ihhub/fheroes2/assets/55348946/d7500283-82f2-4eea-b450-13faace64ad6

dood-apo:
> There could be some cases, when player can close the castle window and just instantly wants to scroll to the right, but with the current implementation it doesn't work. So player needs to return cursor back and then slide to right once again

Correct, that was the idea. But I think it is a fair point, that we should be able to reenable the disabled scrolling also by going even closer to the screen edge with the cursor (without the need of reentering the game area). However I think we should not initiate it if the movement is done in the direction away from the screen edge. What do you think @Branikolog ?

> I'm also not sure, why do we limit such logic only for two borders?

At first it was implemented on all 4. @ihhub pointed out that 4 checks are costly, so I limited their number to borders I thought  are affected. I simply missed the hotkey scenario, thanks for bringing that up.

dood-apo:
Hey @Branikolog , please check if the current behavior is what you would expect from the user perspective.

Branikolog:
Hi, @dood-apo !

I've made a couple of tests.
The left and the upper edges work as expected: after closing castle window and moving cursor along the edges the scrolling is initiated and not initiated if we slowly move cursor out of the edges. 👍 
But for the right and bottom edges the scrolling is not initiated if castle window was closed while cursor was on the most right (or bottom) position.

https://github.com/ihhub/fheroes2/assets/55348946/8ece5090-4334-476c-b792-b88fed73c6ec

dood-apo:
Hi @Branikolog , thanks for such a thorough testing!

> for the right and bottom edges the scrolling is not initiated if castle window was closed while cursor was on the most right (or bottom) position

Fixed that one.

I've also realized that the case of dismissing the dialogs by hotkeys broaden the scope of this bug. With the latest commits I think I've addressed all the cases:
1. Hero dialog
2. Castle dialog
3. All dialogs brought up by the interface buttons
4. All dialogs brought up by hotkeys (eg. 'S' save or 'E' end turn when heroes can still move)

Branikolog:
@dood-apo 

If we're going to apply this logic to all game dialogs, I suggest adding diplomacy and battle results dialogs into the list too.
image

image

Branikolog (approved):
@ihhub 
Than I don't found any more issues as with the changes of this PR scrolling works fine.

@dood-apo 
Thanks for fixing the problem!

dood-apo:
> If we're going to apply this logic to all game dialogs, I suggest adding diplomacy and battle results dialogs into the list too.

While we are at it we should add to the list "week of the creature", countdown to capture a town, signs and story related popups :)
But yeah, this can be addressed in a separate PR.

dood-apo:
Actually I think all the other dialogs (diplomacy, windmills, signs, etc.) can be addressed quite easily. I can hook up to `HotKeyCloseWindow` (or maybe even `HotKeyPressEvent` - interrupting the ongoing scroll with any keypress seems reasonable to me) method and always disable scrolling there. This should cover all the cases when the cursor is on the border and we are dismissing a dialog (any dialog) with a hotkey.

@Branikolog , @ihhub , what do you think?

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -936,6 +939,75 @@ void Interface::GameArea::SetScroll( int direct )
...
+void Interface::GameArea::setFastScrollStatus( const bool enable )
+{
+    _isFastScrollEnabled = enable;
+    _resetMousePositionForFastScroll = true;
+}
+
+bool Interface::GameArea::mouseIndicatesFastScroll( const fheroes2::Point & mousePosition )
+{
+    const fheroes2::Display & display = fheroes2::Display::instance();
+    constexpr int DEADZONE = 3;
```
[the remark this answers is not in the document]
Districh-ru: Hi, @dood-apo.
And one other suggestion: we can use int32_t to fully match the type of `x` and `y` in `fheroes2::Point` structure:
```suggestion
    constexpr int32_t deadZone= 3;
```

dood-apo:
@Districh-ru addressed your comment.
@Branikolog can you please check if the current implementation covers all the dialog types, when you have time?

Districh-ru (approved):
I like the way it works. Well done, @dood-apo.
And thanks for your contribution to the project!

ihhub (approved):

ihhub:
@dood-apo , thank you so much for this fix!

PR #8255, opened 2024-01-11
Optimize the AIPathfinder::getFogDiscoveryTile()

oleg-derevenetz:
Related to https://github.com/ihhub/fheroes2/pull/8235#discussion_r1444103847

The difference with the `master` branch is that in the `master` branch, until at least one tile is found next to which there is fog, tiles next to which there is an obstacle are also considered, and there can be quite a lot of such tiles until a first tile with fog nearby is found (and then tiles without fog nearby will be discarded anyway). With this PR, there are two passes - during the first pass, only tiles are considered, next to which there is fog (and in the vast majority of cases this is the end of it), and during the second pass, the tiles next to which there is an obstacle will be considered. In a typical case, this speeds up the process, but in the worst case (when there is no fog near any of the accessible tiles), there will be two passes through the map tiles instead of one.

Profiler's call tree for the same XL map in the same circumstances as in the comment linked above:

optimized

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -407,31 +407,45 @@ bool Troops::CanJoinTroop( const Monster & mons ) const
            || std::any_of( begin(), end(), []( const Troop * troop ) { return !troop->isValid(); } );
 }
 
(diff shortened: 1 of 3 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for this optimization!

PR #8256, opened 2024-01-11
Make AI to attack human heroes with higher chance than AI heroes

ihhub:
This is to avoid waiting tactics from human players.

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_hero.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_hero.cpp
@@ -611,11 +611,7 @@ namespace
                 return AIShouldVisitCastle( hero, index, heroArmyStrength );
             }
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub why do you prefer to do this through an assessment of the _ability_ to attack, rather than through an assessment of the _priority_ of the attack (i.e. using `HeroesValidObject()` instead of `getXXXObjectValue()`)?

ihhub:
> Hi @ihhub why do you prefer to do this through an assessment of the _ability_ to attack, rather than through an assessment of the _priority_ of the attack (i.e. using `HeroesValidObject()` instead of `getXXXObjectValue()`)?

These are 2 approaches. I think yours one is better. I shall change the logic.

oleg-derevenetz (approved):

idshibanov (approved):

PR #8259, opened 2024-01-11
Rename OBJ_HERO object

ihhub:
This has been annoying me for a very long time. This is a type of object for a single hero, not many standing at one tile.


Districh-ru (approved):
Totally agree!

PR #8262, opened 2024-01-12
Add vertical gradient fonts

MaMadDl:
based on Discussion #8078
Adding colors bounds for all the colors in pal.h
Adding 3 point Linear vertical gradient creator
Adding new font color for NORMAL and LARGE fonts

Sample Output of Fonts:
Large Font:
Screenshot 2024-04-14 214121
Screenshot 2024-04-14 221955

Normal Font:
Screenshot 2024-04-14 214214
Screenshot 2024-04-14 222113


ihhub:
Hi @MaMadDl , before proceeding with this pull request review please update the original description to include all fonts that you are adding here so every reviewer will have visual interpretation of them.

MaMadDl (commented):
Hello @zenseii 
Requested Reviews have been changed
also for Large Font I've changed the outline pixel width to 2 

Please Take a look

MaMadDl:
@Mr-Bajs  you're right.
I'm overwriting the letter color by checking for value in its transform layer. 
where at both ends of S it starts get thinner, I suspect I'm overwriting that the transparent part.
will look into it ASAP.

src/engine/image.h
```diff
@@ -318,4 +318,7 @@ namespace fheroes2
     void Transpose( const Image & in, Image & out );
 
     void updateShadow( Image & image, const Point & shadowOffset, const uint8_t transformId, const bool connectCorners );
+
+    void ApplyVerticalGradient( fheroes2::Image & image, uint8_t darkColor, uint8_t brightColor, uint8_t offset, uint8_t borderWidth );
```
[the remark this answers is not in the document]
ihhub: clang-tidy is correct here.

src/engine/image.cpp
```diff
@@ -3106,4 +3106,55 @@
...
+        // first line is broken for unknown reason
+        uint8_t * data = inData;
+        uint8_t * transform = inTransform;
+        const uint8_t * dataEnd = data + width;
+        for ( ; data != dataEnd; ++data, ++transform ) {
+            *transform = 1;
+        }
+
+        for ( uint8_t i = 0; i < borderWidth; i++ ) {
+            fheroes2::Image cnt = CreateContour( image, 0 );
```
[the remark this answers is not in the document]
ihhub: clang-tidy is correct here.

MaMadDl:
Hello @zenseii 
Thanks for the tip on saving image

here's modified version of Fonts 
Normal
NormalR
Large
LargeR

I have changed the function to take in a Color as well 
and wrap all of them with one black contour

fixed Clang-Tidy issues mentioned by @ihhub 

Let me know what you think

src/engine/image.cpp
```diff
@@ -3106,4 +3106,67 @@ namespace fheroes2
...
+        }
+
+        for ( uint8_t i = 0; i < borderWidth; i++ ) {
+            fheroes2::Sprite cnt = CreateContour( image, borderColor );
+            Blit( cnt, image );
+        }
+
+        fheroes2::Sprite cnt = CreateContour( image, 0 );
+        Blit( cnt, image );
+    }
```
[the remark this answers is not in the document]
MaMadDl: @zenseii 
black contours should come last 
and for brown I think it's best to keep the method customizable so it could be reused.

MaMadDl (commented):
also I have moved the 2 static createContour into `agg_image.cpp` instead of `update_shadow` method

as for the matter of cutter 1 pixel from the middle, after removing 1 pixel from middle a lot of characters got deformed For both Large and Normal fonts so I don't think we should go that route
here's some of the deformations:
Normal `=` 
=normal
`Normal` with middle Cut
=cut
`Large` with middle Cut
=LargeCut



I noticed there's a method called `fheroes2::Stretch`. does that work as linear scaling on `x` and `y` and can we use it to handle the cutting '1 pixel line` is what I asked before.

src/engine/image.cpp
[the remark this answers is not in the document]
MaMadDl: @zenseii 
You were right  I mistook it for 2nd layer ( what is drawn before Black)

but overall I think we should go with dark brown for outside layer, in my opinion slightly faded outline looks better

MaMadDl:
Hello @zenseii

I have added a version of silver gradient however it's slightly thicker than those in campaign 
here's an example for letter H:
Original:
Screenshot 2024-01-16 043858
Recreated:
Screenshot 2024-01-16 044005

there's also a problem when I'm trying to separate original white font shade(pixels that's slightly darker) where we are checking by value (Lines `3133` and  `3153` in `image.cpp` ) and i think it's conflicting with Silver color creation and wasn't able to find good solution for it yet.

let me know what you think

ihhub:
Hi @MaMadDl , please use `Resize()` method for resizing things. `Stretch()` is for different purposes.

MaMadDl:
Hi @zenseii,

I've managed to fix the issues and made the method a lot simpler and shorter

the final output are:
Silver Gradient (Normal and Large):
silverNormalH
silverLargeH

Golden Gradient(Normal and Large):
GoldNormalH
GoldLargeH

src/engine/image.cpp
```diff
@@ -3106,4 +3106,51 @@ namespace fheroes2
...
+        uint8_t * outData = outImage.image();
+        uint8_t * outTransform = outImage.transform();
+
+        const uint8_t center_y = static_cast<uint8_t>( std::max( 1, ( height / 2 ) - height % 2 ) );
+        const float dColor = (float)( outsideColor - insideColor );
+
+        for ( uint8_t row = 0; row < height; row++ ) {
+            float heightScale = static_cast<float>( abs( center_y - row ) ) / center_y;
+
+            uint8_t val = static_cast<uint8_t>( abs( insideColor + (int)( heightScale * dColor ) ) );
```
Districh-ru, line 3129: We can slightly simplify this code:
```suggestion
        const uint8_t dColor = outsideColor - insideColor;

        for ( uint8_t row = 0; row < height; row++ ) {
            const uint8_t heightScale = ( dColor * static_cast<uint8_t>( std::abs( centerY - row ) ) ) / centerY;

            const uint8_t val = static_cast<uint8_t>( std::abs( insideColor + heightScale ) );
```

src/engine/image.cpp
```diff
@@ -3106,4 +3106,51 @@ namespace fheroes2
...
+            uint8_t * inRowStart = inData + static_cast<ptrdiff_t>( row * width );
```
Districh-ru, line 3130: This pointer can be a pointer to `const`.
It is also better (to avoid overflow) to cast not the result of the multiplication, but the multiplier:
```suggestion
            const uint8_t * inRowStart = inData + static_cast<ptrdiff_t>( row ) * width;
```
The `cast` suggestion is also for the rest lies in this loop. :)

src/engine/image.cpp
```diff
@@ -3106,4 +3106,51 @@ namespace fheroes2
...
+            uint8_t * outRowStart = outData + static_cast<ptrdiff_t>( row * width );
+
+            uint8_t * inRowEnd = inData + static_cast<ptrdiff_t>( ( row + 1 ) * width );
```
Districh-ru, line 3133: This pointer can be a pointer to const.

src/engine/image.cpp
```diff
@@ -3106,4 +3106,51 @@ namespace fheroes2
...
+            uint8_t * inTrans = inTransform + static_cast<ptrdiff_t>( row * width );
```
Districh-ru, line 3134: This pointer can be a pointer to const.

Districh-ru (changes requested):
Hi, @MaMadDl, thanks for the gradient fonts implementation, I like how it looks.
I left several comments, could you please take a look when you have time.

src/engine/image.cpp
```diff
@@ -3106,4 +3106,51 @@ namespace fheroes2
...
+                    }
+                }
+            }
+        }
+
+        for ( uint8_t i = 0; i < borderWidth; i++ ) {
+            fheroes2::Sprite cnt = CreateContour( outImage, borderColor );
+            Blit( cnt, outImage );
+        }
+        fheroes2::Copy( outImage, image );
```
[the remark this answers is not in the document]
MaMadDl: Actually we do. In place processing is what I went for in the first implementation and the results had a lot of issues specially on silver gradient.

MaMadDl (commented):
Hi @Districh-ru 
I have made the changes requested , would please take a look?
also on one of the changes requested I had some issues which is commented on.

src/engine/image.cpp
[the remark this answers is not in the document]
Districh-ru: @MaMadDl 
As for now the issued in the code are fixed and we can also avoid making a copy of image and do changes directly in the input image.
I removed the `outImage` from the code and got the same result without any issues:
изображение


<details>

```cpp
        assert( !image.singleLayer() );

        const int32_t height = image.height();
        const int32_t width = image.width();

        uint8_t * inData = image.image();
        uint8_t * inTransform = image.transform();

        const uint8_t centerY = static_cast<uint8_t>( std::max( 1, ( height / 2 ) - height % 2 ) );
        const uint8_t dColor = outsideColor - insideColor;

        for ( uint8_t row = 0; row < height; row++ ) {
            const uint8_t heightScale = ( dColor * static_cast<uint8_t>( std::abs( centerY - row ) ) ) / centerY;

            const uint8_t val = static_cast<uint8_t>( std::abs( insideColor + heightScale ) );
            uint8_t * inRowStart = inData + static_cast<ptrdiff_t>( row ) * width;

            const uint8_t * inRowEnd = inData + static_cast<ptrdiff_t>( row + 1 ) * width;
            uint8_t * inTrans = inTransform + static_cast<ptrdiff_t>( row ) * width;

            for ( ; inRowStart != inRowEnd; ++inRowStart, ++inTrans ) {
                if ( *inTrans == 0 ) {
                    // 21 is the pixel limit of shadows in Base white Font
                    if ( *inRowStart < 21 ) {
                        *inRowStart = val;
                    }
                    else {
                        *inTrans = 1;
                    }
                }
            }
        }

        for ( uint8_t i = 0; i < borderWidth; i++ ) {
            fheroes2::Sprite cnt = CreateContour( image, borderColor );
            Blit( cnt, image );
        }
```
</details>

src/engine/image.h
```diff
@@ -227,6 +227,8 @@ namespace fheroes2
 
     void ApplyTransform( Image & image, int32_t x, int32_t y, int32_t width, int32_t height, const uint8_t transformId );
 
+    void ApplyVerticalGradient( Image & image, const uint8_t outsideColor, const uint8_t insideColor, const uint8_t borderWidth, const uint8_t borderColor );
```
Districh-ru, line 230: Since this function is applicable only to fonts should we place it in `ui_font.h|cpp` files or name it `applyFontVerticalGradientAndContour` (yes, we apply contour here too).
Also please name new functions according to the [code style suggestions](https://github.com/ihhub/fheroes2/wiki/Code-style). :)

src/engine/image.cpp
```diff
@@ -1120,6 +1120,55 @@ namespace fheroes2
         }
     }
 
+    void ApplyVerticalGradient( fheroes2::Image & image, const uint8_t outsideColor, const uint8_t insideColor, const uint8_t borderWidth, const uint8_t borderColor )
```
Districh-ru, line 1123: We are already in `fheroes2` namespace so there is no need to type it for `fheroes2::Image`:
```suggestion
    void ApplyVerticalGradient( Image & image, const uint8_t outsideColor, const uint8_t insideColor, const uint8_t borderWidth, const uint8_t borderColor )
```

Districh-ru (commented):
@MaMadDl, I left some more comments. Could you please check them?

src/engine/image.h
MaMadDl: since most of font manipulator methods are still in `image.h|cpp`  and `ui_font.h` is mostly for multi-language stuff I think `image.h` is more suitable. as for name I tried to follow the conventions already existing in the header

MaMadDl (commented):
hi @Districh-ru 
I've made the changes requested, could you please check them ?

Districh-ru (approved):
Well done, @MaMadDl !

MaMadDl:
Hi @zenseii Conflicts resolved and merged with master

ihhub (approved):

ihhub:
@MaMadDl , thank you very much for this feature!

PR #8266, opened 2024-01-12
Offer the Wisdom skill to the heroes of "magic" classes on a mandatory basis at least once every three level-ups

### LINKED ISSUE #8261: Secondary skills, mandatory wisdom for magic heroes
### Preliminary checks

### Platform

Windows

### Describe the bug

There is mandatory wisdom proposal for magic(Necromancer, sorcerer, sorceress, warlock) heroes in OG 3 levels after last proposal. 

### Save file

Easy to check in OG, and fheroes2

### Additional info

F.e If wisdom wasnt proposed, it always appear on left slot on level 4.
If it was proposed on level 2, the latest it can be proposed is level 5. 
Always wisdom is mandatory 3 levels after last proposal.

oleg-derevenetz:
fix #8261

### DIFF
--- a/src/fheroes2/heroes/skill.cpp
+++ b/src/fheroes2/heroes/skill.cpp
@@ -96,12 +96,12 @@ namespace
         return 0;
     }
 
-    int SecondaryPriorityFromRace( const int race, const std::vector<int> & exclude, const uint32_t seed )
+    int SecondaryPriorityFromRace( const int race, const std::vector<int> & blacklist, const uint32_t seed )
     {
         Rand::Queue parts( MAXSECONDARYSKILL );
 
         for ( auto skill : allSecondarySkills ) {
-            if ( std::find( exclude.begin(), exclude.end(), skill ) != exclude.end() ) {
+            if ( std::find( blacklist.begin(), blacklist.end(), skill ) != blacklist.end() ) {
                 continue;
             }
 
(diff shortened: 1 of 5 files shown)


src/fheroes2/heroes/skill.cpp
```diff
@@ -669,83 +733,69 @@ void Skill::SecSkills::FillMax( const Skill::Secondary & skill )
         resize( HEROESMAXSKILL, skill );
 }
 
-int Skill::SecondaryGetWeightSkillFromRace( int race, int skill )
+std::pair<Skill::Secondary, Skill::Secondary> Skill::SecSkills::FindSkillsForLevelUp( const int race, const uint32_t firstSkillSeed,
+                                                                                      uint32_t const secondSkillSeed ) const
 {
-    const stats_t * ptr = GameStatic::GetSkillStats( race );
+    std::vector<int> blacklist;
```
ihhub, line 739: Since we are touching this function should we change the container type to `std::set` or `std::unsorted_set` which are designed for these purposes?

ihhub (changes requested):
Hi @oleg-derevenetz , I have only one suggestion here. Would you mind to take a look?

src/fheroes2/heroes/skill.cpp
oleg-derevenetz: @ihhub Eh, well, since we are rewriting this anyway, we can.

ihhub (approved):

ihhub:

PR #8272, opened 2024-01-13
Defensive tactics in battle: allow units covering archers with the AREA_SHOT capability to attack neighboring units on their own

oleg-derevenetz:
Such archers may not always attack enemy units in close proximity to friendly units due to fear of friendly fire, so covering friendly units should help by attacking enemy units.

https://github.com/ihhub/fheroes2/assets/32623900/c0c659fd-8c7a-4ed1-a8f6-d8b531bbcf65

With this PR:

https://github.com/ihhub/fheroes2/assets/32623900/13e23882-fd04-4d8b-821a-10d2dab9f92d

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -1637,8 +1637,10 @@ namespace AI
                 }
 
                 // It makes sense for a unit that ignores retaliation to attack neighboring enemy units, even if it is covering an archer, since in this case it will not
-                // receive unnecessary retaliatory damage (which could affect the duration of the cover)
-                if ( !currentUnit.isIgnoringRetaliation() ) {
+                // receive unnecessary retaliatory damage (which could affect the duration of the cover). Also, archers with the ability to shoot at area may not always
+                // attack enemy units in close proximity to friendly units due to fear of friendly fire, so covering friendly units should help by attacking enemy units
+                // next to them.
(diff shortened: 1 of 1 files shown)


ihhub:
The change is very straightforward. I foresee that @Branikolog would ask more complex stuff like what if there are units that a shooter can still hit.

oleg-derevenetz:
> what if there are units that a shooter can still hit

:) If there are units that a shooter can hit, he will just shoot at them as usual.

Branikolog:
I've already checked the issue this night and was ready to approve just in an hour of testing. :)
Suprisingly this works perfectly. If damage from liches is too high over allies, they skip and melee fight proceeds. If damage dealed to allied stacks is reasonable and enemy is suffering more - liches attack. Liches often prefer damaging a couple of enemy stacks + one allied and that is adequate. Liches can even demolish an allied stack if there's a considerable damage to enemy.
Well done, @oleg-derevenetz ! With this logic there's much more fun in battles now. :)

Branikolog (approved):

oleg-derevenetz:
> Will the genies still not attack the vampires?

No, Genies will attack the Vampires, and Liches will attack these distant archers. The point of "not attacking" is to not take a retaliatory damage to last more turns and let the archers do the job, but in this case archers cannot do the job, so there is no choice.

ihhub (approved):

ihhub:

PR #8276, opened 2024-01-14
Single layer image processing improvements

Districh-ru:
This PR is a part of #8241 but without changes to `transform()` layer memory allocation and `reset()` of transform layer.


src/engine/image.cpp
```diff
@@ -519,9 +519,9 @@ namespace fheroes2
             return;
         }
 
-        const size_t size = static_cast<size_t>( width_ ) * height_;
```
ihhub, line 522: Should we revert these change for memory allocation as per description?

src/engine/screen.cpp
```diff
@@ -1343,10 +1349,9 @@ namespace fheroes2
         }
 
         Image::resize( info.gameWidth, info.gameHeight );
-        _screenSize = { info.screenWidth, info.screenHeight };
+        Image::reset();
 
-        // To detect some UI artifacts by invalid code let's put all transform data into pixel skipping mode.
-        std::fill( transform(), transform() + width() * height(), static_cast<uint8_t>( 1 ) );
+        _screenSize = { info.screenWidth, info.screenHeight };
```
ihhub, line 1354: I suggest to revert these changes since we still preserve memory for the transform layer (for now).

ihhub (changes requested):
Hi @Districh-ru , I put 2 comments here. Would you mind to take a look at them when you have time?

src/engine/image.cpp
Districh-ru: Hi, @ihhub, the amount of allocated memory will be the same as in master build. I've just moved `* 2` from 524 line to 522 line because the `size` of allocated memory (in bytes) is `_width * _height * 2` and also to do all calculations in one line and allocation in another line (without any mathematical operations with the size of memory to allocate).

src/engine/screen.cpp
Districh-ru: I master build we just put `1` to the transform layer. I changed it to `reset()` method that will do the same and in addition will put `0` to the just resized screen image data layer.
I thought it would be a little safer this way:
- we will fix possible UI artifacts resulting from an undefined image data layer. This has never happened, but it is possible. And we use `setResolution()` rarely, so filling the image data with `0` won't have much impact on the engine's performance;
- if we remove the transform layer for the screen image, then no more changes will be required here - all the necessary changes will be made in the `reset()` method.

But I could be wrong. If you think so I'll revert the changes. :)

src/engine/image.cpp
ihhub: Oh, yes. You're right. My bad.

src/engine/screen.cpp
ihhub: It seems reasonable. Thanks for the explanation!

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for these improvements!

PR #8277, opened 2024-01-14
Use proper name for PoL campaign in High Scores

zenseii:
The OG never refers to it as "The Price of Loyalty" iirc:
image

This fixes this issue:
image


ihhub (approved):

ihhub:
@zenseii , huge thanks for the fix!

Issue #8237, opened 2024-01-06
Build fails: "end-of-line within string"


Stiivi:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

macOS

### Describe the bug

The build fails with:

```
sed -e 'y/äëïöőüűÄËÏŐÖÜŰ/aeioouuAEIOOUU/' lt.po | iconv -f utf-8 -t ascii//TRANSLIT | msgfmt - -o lt.mo
<stdin>:3865: end-of-line within string
iconv: warning: invalid characters: 702
msgfmt: found 1 fatal error
make[1]: *** [Makefile:61: lt.mo] Error 1
make[1]: Leaving directory '/Users/stefan/Developer/Sources/fheroes2/files/lang'
make: *** [Makefile:39: all] Error 2
```

Broken commit: 5dfd271c0fb41e00927252699abea36aba12aafb

Last working commit (no error): 4174e71dd640ccda6d808876c05c24f284efaa00

### Save file

(no save file)

Stiivi:
No changes on my side. Just did a clean git clone of 92b39d82d9fcd998a8775acdcc919a4973d03c5f into a new location, got the same error.

Attached is whole checout+build session with some git info at the end.

[fheroes2-build-session.txt](https://github.com/ihhub/fheroes2/files/13850808/fheroes2-build-session.txt)

oleg-derevenetz:
I cannot reproduce this issue on my macOS with neither make nor CMake. Our CI cannot reproduce this issue as well. What is your locale? Is it UTF-8 locale? What will be the output of `env | grep LANG` command in your console?

Stiivi:
`env | grep LANG` gives no results - no env variable set. There is only one env. locale-related variable set: `LC_CTYPE=UTF-8`.

ihhub:
Hi @Stiivi , do you use macOS 14?

ihhub:
I believe [this Homebrew pull request](https://github.com/Homebrew/homebrew-core/pull/158820) is facing the same problem.

Stiivi:
Using MacOS Sonoma 14.2.1.

The lined Homebrew issue seems to be exhibiting the same issue:

```
[...]

PR #8278, opened 2024-01-14
Use CP1257 for Lithaunian

zenseii:
Possible solution for #8237

PR #8279, opened 2024-01-14
Workaround for misbehaved iconv on MacOS 14

Stiivi:
Added few filler characters to the `lt.po` translation file in a form of a dummy comment to work around `iconv` issues on MacOS 14 (Sonoma).

Issue: #8237

Tested on MacOS 14.2.1


ihhub:
I am closing this pull request as we merged #8278 as an alternative solution.

Issue #4495, opened 2021-10-25
Adventure map, passability test


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Hello @ihhub,

So, I finished my big (and boring) test on passability without any object adjacency.
There are still many differences, here some examples:

Sawmills.
![Passability test 015](https://user-images.githubusercontent.com/43583994/138714812-55a54129-918b-420d-8870-b698d652948c.png)
![Passability test 016](https://user-images.githubusercontent.com/43583994/138714828-157dea7a-5162-483c-90c3-02967b5e33a7.png)
![Passability test 017](https://user-images.githubusercontent.com/43583994/138714843-bc5902dc-bb20-4ab0-b560-b2941a952d9c.png)
![Passability test 018](https://user-images.githubusercontent.com/43583994/138714853-8d35b789-5232-42a9-9a73-f699f8bcfb9c.png)
![Passability 02 013](https://user-images.githubusercontent.com/43583994/138714973-c34fc5c5-0e74-42af-b502-1c6985764c7c.png)
![Passability 02 014](https://user-images.githubusercontent.com/43583994/138714986-ad94bb5c-5a6b-4295-9e9a-112d52ce079f.png)
![Passability 02 015](https://user-images.githubusercontent.com/43583994/138715002-2bb6fb60-271e-4263-b808-73a731cf63ff.png)
![Passability 02 016](https://user-images.githubusercontent.com/43583994/138715009-60c08bb1-08fc-40cd-8957-d0541862c069.png)
![Passability 02 069](https://user-images.githubusercontent.com/43583994/138715194-45125f01-28c8-4f9a-b4fd-083745a8e6ae.png)
![Passability 02 070](https://user-images.githubusercontent.com/43583994/138715202-f5d12f7d-b8a0-4a93-89c8-b5fa5651cf97.png)

Water Lake.
![Passability 02 045](https://user-images.githubusercontent.com/43583994/138715245-955aa118-9e3f-4108-88ef-47327d9283b9.png)
[...]

PR #8280, opened 2024-01-14
Implement Landscape Miscellaneous (non action) objects placing in Editor

Districh-ru:
This PR adds Landscape (non action) miscellaneous objects to editor database.
There are lots of such objects, currently they are added (and shown) in the order of their appearance in game ICNs. I hope I didn't miss any non-action objects. :)
Later we well make a new dialog to show these objects by groups (lake, cactus, flowers, ... ). Suggestions for this dialog are accepted.

eidtor

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -3145,7 +3145,7 @@ namespace fheroes2
                     Blit( GetICN( ICN::OBJNWAT2, 2 ), 5, 0, _icnVsSprite[id][9], 1, 4, 24, 21 );
 
                     // Make Landscape Miscellaneous objects button.
-                    Blit( GetICN( ICN::OBJNMUL2, 16 ), 3, 0, _icnVsSprite[id][10], 1, 4, 24, 19 );
+                    Blit( GetICN( ICN::OBJNDIRT, 73 ), 8, 0, _icnVsSprite[id][10], 1, 1, 24, 24 );
 
                     // Make erase Dwellings button image.
                     Blit( GetICN( ICN::OBJNMULT, 114 ), 7, 0, _icnVsSprite[id][11], 1, 1, 24, 24 );
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -462,9 +462,9 @@ namespace
 
             const fheroes2::Sprite & image = fheroes2::generateMapObjectImage( _objectInfo[objectId] );
             const int32_t imageHeight = image.height();
-            if ( imageHeight > TILEWIDTH * 2 ) {
+            if ( imageHeight > TILEWIDTH * 3 ) {
                 // Reduce the size of very tall images to fit the list.
-                const double ratio = imageHeight / ( TILEWIDTH * 2. );
+                const double ratio = imageHeight / ( TILEWIDTH * 3. );
                 fheroes2::Image resized( static_cast<int32_t>( image.width() / ratio ), static_cast<int32_t>( imageHeight / ratio ) );
                 fheroes2::Resize( image, resized );
                 renderItem( resized, getObjectName( _objectInfo[objectId] ), { posX, posY }, _imageOffsetX, _textOffsetX, _offsetY / 2, isSelected );
@@ -1274,7 +1274,7 @@ int Dialog::selectLandscapeMiscellaneousObjectType( const int objectType )
 {
     const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::LANDSCAPE_MISCELLANEOUS );
 
-    DwellingTypeSelection listbox( objectInfo, { 420, fheroes2::Display::instance().height() - 180 }, _( "Select Landscape Object:" ) );
+    GenericObjectTypeSelection listbox( objectInfo, { 420, fheroes2::Display::instance().height() - 180 }, _( "Select Landscape Object:" ) );
 
     return selectObjectType( objectType, objectInfo.size(), listbox );
 }
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -462,10 +462,11 @@ namespace
 
             const fheroes2::Sprite & image = fheroes2::generateMapObjectImage( _objectInfo[objectId] );
             const int32_t imageHeight = image.height();
-            if ( imageHeight > TILEWIDTH * 3 ) {
+            const int32_t imageWidth = image.width();
+            if ( imageHeight > TILEWIDTH * 3 || imageWidth > TILEWIDTH * 5 ) {
                 // Reduce the size of very tall images to fit the list.
-                const double ratio = imageHeight / ( TILEWIDTH * 3. );
-                fheroes2::Image resized( static_cast<int32_t>( image.width() / ratio ), static_cast<int32_t>( imageHeight / ratio ) );
+                const double ratio = std::max( imageHeight / ( TILEWIDTH * 3. ), imageWidth / ( TILEWIDTH * 5. ) );
+                fheroes2::Image resized( static_cast<int32_t>( imageWidth / ratio ), static_cast<int32_t>( imageHeight / ratio ) );
                 fheroes2::Resize( image, resized );
                 renderItem( resized, getObjectName( _objectInfo[objectId] ), { posX, posY }, _imageOffsetX, _textOffsetX, _offsetY / 2, isSelected );
             }
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -1249,6 +1249,15 @@ int Dialog::selectDwellingType( const int dwellingType )
     return selectObjectType( dwellingType, objectInfo.size(), listbox );
 }
 
+int Dialog::selectLandscapeMiscellaneousObjectType( const int objectType )
+{
+    const auto & objectInfo = Maps::getObjectsByGroup( Maps::ObjectGroup::LANDSCAPE_MISCELLANEOUS );
+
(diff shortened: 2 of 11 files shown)


Districh-ru:
In original game there are some objects called "Nothing Special".
изображение
But we have no unused object types to give these object the new names. :(
(in example: Swampy Lake, Reed, Rotten Roots)

ihhub (approved):

ihhub:
Hi @Branikolog , please prioritize this pull request.

Branikolog:
Hi, @Districh-ru 

Here are few issues I've found. Could you, please, look if they are valid and could be fixed?

I believe, that we should always place "no objects" like ground cracks under roads, streams and all other objects.
image

The hole, though, looks well placed both under or above roads/streams, but while placing some streams over it we should preserve the overlay of the existing stream nearby:
image

Cracks and holes shouldn't be allowed to be placed on a water:
image

Branikolog:
@Districh-ru 
Not sure, it's relarted to the current PR, but after loading the map, that I was used for testing this PR, any info terrain except ocean is stated as "beach":
image

[22.zip](https://github.com/ihhub/fheroes2/files/13956570/22.zip)

Districh-ru:
> Not sure, it's related to the current PR, but after loading the map, that I was using for tests of this PR, any info terrain except ocean is stated as "beach"

While reading map from file we read each tile one by one and if tile contains an object then we also update passabilities on map, but at this time not all tile are loaded and have tile image index reset to 0 (zero). And this is an index of water tile, so the pass updater sees that near the tile there is a "water" and also sets it as Beach (OBJ_COAST).

Hi, @ihhub, should we first set all terrain image index data for all tiles and then load objects or change `setObjectOnTile()` to call `world.updatePassabilities();` only if a boolean parameter (e.g. `const bool doUpdatePassabilities`) is set to true and while loading a map set it to `false` and update passabilities only after the whole map is loaded only once (it will be also faster). What do you think?

I also suggest to fix it in a separate PR, because this issue occurs on any map with any objects.

ihhub:
Hi @Districh-ru , yes, let's fix map loading in a separate pull request. @Branikolog , if you don't have other comments please approve this pull request.

Branikolog:
Hi, @Districh-ru 

I've noticed some objects have differences in passabilities with the OG. Could you, please, check if I was wrong?

1. 
image

image

2. 
image

image

3.
image

image

4. 
image

image

5. 
image

image

6.
image

image

7. 
image

image

image

8.
image

image

Also, I couldn't find this cactus in the original editor: 
image

:)

Branikolog:
@Districh-ru 

>We have some situations to discuss, that is why I suggest to fix it separately.

Sure. But I had to mention them so we are aware of that problems. :)

Branikolog:
Hi, @Districh-ru 

> 1. Yes, originally this tile is not passable: изображение
>    But it looks like a bug: this tile is a part of the ground and a hero in this tile will look like standing on the ground so it looks more like a passable:
>    изображение
>    We have already changed some passabilities for mountains, so I decided to make passavilities here more correspond to the object look.

Maybe horse's legs are too close to the swamp lake? I cannot test it, but it seems to me that with the proposed passability for this tile hero can make a diagonal move to the left. While doing that he will partially step over a lake, which is not good. Probably a green tile with red corner would be more proper here?
But actually, the OG have its own issues that seem even more broken, and in my personal opinion they require even more attention from our side.
image


I think we can find even more such controversial passability issues (I mean in the original editor to be fixed). I've just found the differences with fheroes2 engine. I'm not against that fixes, but discussing it here could make this PR postponed for really long time.


> * Nice catch: this is a difference between our engine and the original game. An map made by the original editor:
>   изображение
>   You can open an issue, showing these differences between original game and our engine to discuss what passability should we use for the original maps.
>   This impassable tile part may break some maps done by the original editor. :(
> *

Well, hero actally steps too close to the water in this case, so current fheroes2 passability seems even more proper to me. If we forget original maps, I'd rather prefer current passing rules for that tile in our future editor.

> 4\. Another nice catch: this is also a difference between our engine and the original game:
>     изображение
>     It cannot be fixed in editor. But IMHO this passability better suits this object and should not be change.

Agreed.


> 5\. This passability was changed by the same reason as in 1.

Won't hero be standing too close to the hot lava then?



> 7\. The same as 4, this is a map created by the original engine:
>     изображение
>     But do we need to change this passability? This is a single blooming cactus and do not need extra walk around. :)

This one just stands out from the other cactuses, that's why I've noticed it. Actually, I was more intrigued, why do other cactuses have such wide walkaround area... Maybe cactuses are so spiky, so hero just doesn't want to come too close to them? :)
If we follow such logic, I'd prefer to have such excessive walkaround area around all cactuses. :)

> 8\. The same as 4, this is a map created by the original engine:
>     изображение
>     In this case for me the passability in our engine is more consistent with the object's image, but it may break some maps done by the original editor. :(

Agreed. With the original passability hero just steps over the crack, which looks bad.

> 9\. This cactus is the only cactus for the Wasteland terrain :)
>     изображение

Ah found it! Thanks. 
It works fine, btw. :)

Branikolog:
> 3\. Nice catch: this is a difference between our engine and the original game. An map made by the original editor:
>     изображение
>     You can open an issue, showing these differences between original game and our engine to discuss what passability should we use for the original maps.
>     This impassable tile part may break some maps done by the original editor. :(

Actually, this one was mentioned by @LeHerosInconnu in https://github.com/ihhub/fheroes2/issues/4495#issue-1035322069 , but was not fixed yet.

Hey, @ihhub , could you, please, tell us your opinion on these issues? Do we need to fix the passability in the original maps for objects I've mentioned above?

ihhub:
@Districh-ru and @Branikolog , passabilities of individual objects must be the same as in the original **Editor** which actually displays them. This involves 4 types of terrain layers and nothing more. For some objects we can fix obvious cases.

For the game passabilities we have to fix outside the scope of this pull request and the Editor. I was suspecting that we did not fix some cases and there they are. @Branikolog , please document them in #4495 issue.

ihhub:
> Hi @Districh-ru , yes, let's fix map loading in a separate pull request. @Branikolog , if you don't have other comments please approve this pull request.

Terrain problem has been fixed.

ihhub:
> Hi, @Districh-ru
> 
> Here are few issues I've found. Could you, please, look if they are valid and could be fixed?
> 
> I believe, that we should always place "no objects" like ground cracks under roads, streams and all other objects. image
> 
> The hole, though, looks well placed both under or above roads/streams, but while placing some streams over it we should preserve the overlay of the existing stream nearby: image
> 
> Cracks and holes shouldn't be allowed to be placed on a water: image

This should be fixed now.

ihhub:
@Districh-ru and @Branikolog , I also fixed very slow map loading. Now it should be instant.

ihhub:
Hi @Branikolog ,
> The hole, though, looks well placed both under or above roads/streams, but while placing some streams over it we should preserve the overlay of the existing stream nearby:

This is fixed as well.

Districh-ru:
Hi, @Branikolog, issues # 1 and 5 are fixed to match the original editor.
We will discuss the needed passability changes later and apply them in a new PR.

Districh-ru:
> This one just stands out from the other cactuses, that's why I've noticed it. Actually, I was more intrigued, why do other cactuses have such wide walkaround area... Maybe cactuses are so spiky, so hero just doesn't want to come too close to them? :)
> If we follow such logic, I'd prefer to have such excessive walkaround area around all cactuses. :)

The passability logic differs for tall objects (1 object tile and one tile above it in the top level passable by hero) and for low objects (without the upper tile). We can see the same difference in example for the Ork's Watch Tower on snow and other terrains. So currently it is defined by engine according to the rule: is there an upper passable object tile (not empty) image.

Branikolog:
Hi, @ihhub 

I've still experience an issue with streams and holes:

https://github.com/ihhub/fheroes2/assets/55348946/78d96960-4dc5-4e86-a019-fceb273e6e18

ihhub:
Hi @Branikolog , it works as expected. The patch was put first and the river second. I see no issues with this. What you are asking is to do unnecessary work for object placement conditions where we need to handle special cases like expanding an existing river. This must not be handled by the Editor.

Branikolog:
Hi, @ihhub 
It actually looks bad. In my personal opinion we shouldn't allow players such placing, that looks like a bug. We're not allowed to place road on the ocean, right?

Maybe it should better to make a straightforward logic then, that always places such objects under a road or a stream?

ihhub:
> Hi, @ihhub It actually looks bad. In my personal opinion we shouldn't allow players such placing, that looks like a bug. We're not allowed to place road on the ocean, right?
> 
> Maybe it should better to make a straightforward logic then, that always places such objects under a road or a stream?

@Branikolog , no, this is not true. As a map maker I might choose to put an object either under a stream or above. What you are suggesting is to force some objects appear **always** under a stream and to add extra unnecessary logic to the engine's Editor to facilitate this "special" case.

Here you go the same behavior in the original Editor:
image

So no, we are **not** going to do this.

Branikolog:
@ihhub 
I agree with you, that handling such edge cases is a huge pain and I am ready to approve this pr, as everything else works fine. But I personally consider such cases as bugs, when occuring in the released editor. Even the possibility to place any object into won't place or work visual flaws makes an impression of half-baked and not well-polished product.
To not to postpone this PR even more we can merge it, but I hope there could be something name with issue in the future.

ihhub:
> @ihhub I agree with you, that handling such edge cases is a huge pain and I am ready to approve this pr, as everything else works fine. But I personally consider such cases as bugs, when occuring in the released editor. Even the possibility to place any object into wrong place or work visual flaws makes an impression of half-baked and not well-polished product. To not to postpone this PR even more we can merge it, but I hope there could be something name with issue in the future.

You can add a comment in #6845 issue to possibly address this problem in the future.

Branikolog (approved):
Okay.

ihhub:
@Districh-ru , thank you so much for this gigantic work!

Branikolog:
@ihhub 

>You can add a comment in https://github.com/ihhub/fheroes2/issues/6845 issue to possibly address this problem in the future.

There's already a reposted message by @Districh-ru including a screenshots of broken streams.

PR #8282, opened 2024-01-14
Remove "the" from Price of Loyalty title

zenseii:
Missed this one.


src/fheroes2/dialog/dialog_selectscenario.cpp
```diff
@@ -109,7 +109,7 @@ namespace
             body.add( { _( "The Succession Wars" ), fheroes2::FontType::normalWhite() } );
             break;
         case GameVersion::PRICE_OF_LOYALTY:
-            body.add( { _( "The Price of Loyalty" ), fheroes2::FontType::normalWhite() } );
+            body.add( { _( "Price of Loyalty" ), fheroes2::FontType::normalWhite() } );
```
ihhub, line 112: I don't understand why we are changing here. The Succession Wars has **THE** as well.

PR #8286, opened 2024-01-15
Do not do operations with pointer to nullptr in image.cpp

### LINKED ISSUE #8285: Undefined behavior in fheroes2::Flip()
### Preliminary checks

### Platform

macOS

### Describe the bug

If output image is single-layer, `transformOutY` becomes `nullptr`:

https://github.com/ihhub/fheroes2/blob/618cfd9c3aa9c8a4748d0f625e4d274a5cd00f42/src/engine/image.cpp#L2263

But later `transformOutY` is subject to arithmetic operations, e.g. here:

https://github.com/ihhub/fheroes2/blob/618cfd9c3aa9c8a4748d0f625e4d274a5cd00f42/src/engine/image.cpp#L2295

which is [undefined behavior](https://en.cppreference.com/w/cpp/language/operator_arithmetic#Pointer_arithmetic) - you only can add or subtract zero to/from `nullptr`, and that's all.

The same problem is probably present in other places where single-layer images are processed.

### Save file

None needed.

### Additional info

Credits: Clang UB Sanitizer.

Districh-ru:
close #8285

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -2289,8 +2289,7 @@ namespace fheroes2
                     }
 
                     if ( transformOutY != nullptr ) {
-                        // 'in' image is single-layer, 'out' image is double-layer (with the transform layer).
-                        // Set 'in' image transform data not to skip image data.
+                        // Set 'out' image transform data not to skip image data.
                         memset( transformOutY, static_cast<uint8_t>( 0 ), width );
 
                         transformOutY += widthOut;
@@ -2332,12 +2331,11 @@ namespace fheroes2
             const uint8_t * imageInY = in.image() + offsetIn + ( height - 1 ) * widthIn;
 
             if ( in.singleLayer() ) {
-                // 'in' image is single-layer, 'out' image is double-layer (with the  transform layer).
                 for ( ; imageOutY != imageOutYEnd; imageOutY += widthOut, imageInY -= widthIn ) {
                     memcpy( imageOutY, imageInY, static_cast<size_t>( width ) );
 
                     if ( transformOutY != nullptr ) {
-                        // Set 'in' image transform data not to skip image data.
+                        // Set 'out' image transform data not to skip image data.
                         memset( transformOutY, static_cast<uint8_t>( 0 ), width );
 
                         transformOutY += widthOut;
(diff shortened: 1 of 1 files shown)


src/engine/image.cpp
```diff
@@ -1412,21 +1423,21 @@ namespace fheroes2
                     --imageOutRightPoint;
                     --imageInRightPoint;
 
-                    if ( transformOutRightPoint ) {
+                    if ( transformOutRightPoint != nullptr ) {
```
oleg-derevenetz, line 1426: `if ( transformOutRightPoint ) {` is also OK, [non-null pointers are truthy and null pointers are falsy](https://en.cppreference.com/w/cpp/language/implicit_conversion#Boolean_conversions), so previous syntax is also OK here and in other places.

oleg-derevenetz (approved):
Hi @Districh-ru I just did a test run of this PR's branch using the same sanitizer, no issues were found. There are some unnecessary (from my point of view) changes, but it's a matter of taste.

src/engine/image.cpp
Districh-ru: Hi @oleg-derevenetz, I made these changes to more clearly show that we are checking for `nullptr`. The compiled code will be the same.

Hi @ihhub, what syntax is better to leave in our code?

ihhub: We can the way you implemented. It's a matter of taste but your way is not explicit.

ihhub (approved):

ihhub:
@Districh-ru , thank you very much for these changes!

oleg-derevenetz:
Hi @Districh-ru probably SonarQube analyzer needs some checks or assertions to ensure that there will be no dereferences of null pointers:

https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=ihhub_fheroes2

You can create PR branch directly on this repo to run the SonarQube checks for that PR.

Districh-ru:
> Hi @Districh-ru probably SonarQube analyzer needs some checks or assertions to ensure that there will be no dereferences of null pointers:
> 
> https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=ihhub_fheroes2
> 
> You can create PR branch directly on this repo to run the SonarQube checks for that PR.

Hi @oleg-derevenetz , thanks for noticing this. I'll use SonarLint Visual Studio extension to check these bugs.

Issue #2260, opened 2020-11-26
Combat screen, Surrender window, wrong cost for the hero's surrender, and also when the hero owns the Statesman's Quill artifact


LeHerosInconnu:
In fheroes2, there is a wrong cost for the hero's surrender.

In the original game, the cost is 50 percent of the total gold price of the creatures.
In the example, with 1 Peasant (cost 20 gold) the price of surrender is 10 gold.

In original game:

![Surrender H2 01](https://user-images.githubusercontent.com/43583994/100388692-9ebf5b80-302b-11eb-8c3e-189e90c2f9c9.png)

In fheroes2:

![Surrender fh2 01](https://user-images.githubusercontent.com/43583994/100388704-a8e15a00-302b-11eb-900e-4a1a3f7fb214.png)

In the original game, the cost is 10 percent of the total gold price of the creatures if the hero owns the Statesman's Quill artifact.
In the example, with the Statesman's Quill artifact, with 1 Peasant (cost 20 gold) the price of surrender is 2 gold.

In original game:

![Surrender H2 02](https://user-images.githubusercontent.com/43583994/100390322-3a52cb00-3030-11eb-9504-c4e16ce70253.png)

In fheroes2:

![Surrender fh2 03](https://user-images.githubusercontent.com/43583994/100390338-48a0e700-3030-11eb-9ba1-4adcfc7fcecc.png)

![Surrender fh2 02](https://user-images.githubusercontent.com/43583994/100390339-4a6aaa80-3030-11eb-8b12-40893f9a6ce0.png)

Here is the test scenario file:
[Surrender fh2 01.zip](https://github.com/ihhub/fheroes2/files/5604907/Surrender.fh2.01.zip)

undef21:
@LeHerosInconnu , is cost depends on Diplomacy skill in original game?

LeHerosInconnu:
@undef21.

> @LeHerosInconnu , is cost depends on Diplomacy skill in original game? If yes, is Diplomacy effect summed up with Statesman's Quill effect?
> There is minimal cost 100 gold in fheroes2. Please, check surrending with army's cost > 200 gold

In the original game the cost is reduced with the Diplomacy skill and the effect of the Statesman's Quill further reduces the cost.

Surrender with 1000 Peasants (cost 20,000 gold).

In original game:

No Diplomacy, without Statesman's Quill.
[...]

PR #8290, opened 2024-01-16
Correct the Statesman's Quill's effect to 20%

zenseii:
According to @LeHerosInconnu's [tests of the OG](https://github.com/ihhub/fheroes2/issues/2260#issuecomment-735281489) and [this article from handbookhmm](https://handbookhmm.ru/2-diplomacy) (see part number 4), the Statesman's Quill should reduce the cost **to** 20%, not 10%.

Note that the handbook article lists the wrong values for diplomacy's effect, which is correct in fheroes2.

This PR also changes the artifact description to match the change, which in the OG was wrong since it said 10% when it was 20%.

### DIFF
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2023                                             *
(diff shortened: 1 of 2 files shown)


oleg-derevenetz:
Hi @zenseii `master` branch currently performs exactly like in that comment:

https://github.com/ihhub/fheroes2/assets/32623900/a81ddb96-da22-4b25-a03f-0413f1e882ab

PR #8296, opened 2024-01-17
Fix SonarQube issues in image.cpp

Districh-ru:


src/engine/image.cpp
```diff
@@ -597,10 +597,10 @@ namespace fheroes2
     Sprite & Sprite::operator=( Sprite && sprite ) noexcept
     {
         if ( this != &sprite ) {
-            Image::operator=( std::move( sprite ) );
-
             std::swap( _x, sprite._x );
             std::swap( _y, sprite._y );
+
+            Image::operator=( std::move( sprite ) );
```
oleg-derevenetz, line 603: Does this change need to convince SonarQube of the correctness of the logic, or is it just a change of some kind of aesthetic nature?

ihhub: I suggest to revert this change if it makes not difference in SonarQube analysis results.

ihhub (approved):

src/engine/image.cpp
Districh-ru: Hi, @oleg-derevenetz  and @ihhub, this change is done to fix Clang-Tidy issue: `'sprite' used after it was moved [bugprone-use-after-move]`

ihhub:
@Districh-ru , thank you very much for these fixes!

PR #8297, opened 2024-01-17
Make rules for AI surrender and retreat more unified

### LINKED ISSUE #8292: AI shouldn't surrender in the final fight
### Preliminary checks

### Platform

Windows

### Describe the bug

Currently AI can surrender while defending his last castle:
image

image

There's no need to surrender from the last battle for the AI player.
If I'm not mistaken, the engine has such rule for running away from the battle. 

Maybe that's why I sometimes find some AI heroes in tavern with a considerable army, that surrendered but the player has been vanished.

### Save file

[surrender.zip](https://github.com/ihhub/fheroes2/files/13956781/surrender.zip)

### Additional info

_No response_

oleg-derevenetz:
close #8292

The changes proposed in this PR are as follows:

1. Allow AI to retreat even if it's a "final fight", just like it currently happens with surrendering. Rationale:
1.1. If the AI is part of an alliance, then it helps its alliance by not handing over valuable artifacts to the enemy;
1.2. If the AI is not a part of an alliance, then it at least doesn't help its enemy. It's the enemy's problem to kill the AI so that it can't escape: just like in a battle vs human player;
1.3. If enemy will win the scenario by defeating this hero, it doesn't matter whether this hero will be killed, will surrender or retreat. There is no need for a special logic for this case.

2. Allow AI to retreat or surrender even if it cannot be re-hired by the AI directly on the current or next turn. Rationale:
2.1. It is impossible to calculate this accurately. Even if AI doesn't have necessary resources right now, it can find missing gold on the map and/or capture a town/castle that could bring the missing gold tomorrow;
2.2. The hero is very likely to be available in the tavern for more than one day, so allowing the enemy to kill your hero (and receive his artifacts), whom you will most likely be able to re-hire, relatively speaking, a couple of days later is not very smart.

3. Allow player to surrender even if his kingdom doesn't have any castles. Rationale:
3.1. This is how it works in the original game (see the video below);
3.2. Even if player doesn't have a castle right now, it doesn't mean that he can't get it later - e.g. by capturing it with another hero. In short, we do not need to make such decisions for the player, he must decide for himself.

https://github.com/ihhub/fheroes2/assets/32623900/17c8411e-9487-4374-9cba-7562360c0375


LeHerosInconnu:
@oleg-derevenetz.

>    1.3. If enemy will win the scenario by defeating this hero, it doesn't matter whether this hero will be killed, will surrender or retreat. There is no need for a special logic for this case.

Let the human player fight the last opposing hero to the death to feel the joy of total fulfillment.

Do AI heroes surrender to each other?

oleg-derevenetz:
> Let the human player fight the last opposing hero to the death to feel the joy of total fulfillment.

He has to earn it :)

> Do AI heroes surrender to each other?

Yes, of course.

Branikolog:
Hello, @oleg-derevenetz 

I'm agreed with most of proposed rules.

I'm only concerned about surrendering from the last owned castle.
There's no 100% guarantee that AI would manage to capture another town (I think implementing such long-term planning for AI is not going to happen the nearest time as it requires really much work). So I consider defending the last castle = the last fight. 
Leaving from the battle should bring greater losses to the attacking player. Defending hero should inflict maximum damage. Even if he kills just a few troops more, this would certainly increase the possibility for the rest heroes to capture the same castle back or feel stronger against weakened attacker in the future. 
Firstly, it's quite risky to allow attacker win with less losses as he can defend the castle with higher chance especially considering, that surrenered hero couldn't be hired to reinforce initial owner of the castle.
Secondly, surrendering gives money to the attacker, which helps him to hire someone in that castle or build a defending structure in that castle.
Thirdly, in the current implementation (from this PR) AI doesn't consider valuable artifacts (actually any artifacts at all), and surrenders with empty hero from the last castle. 

Similar arguments I'd apply for the battles, when AI has no castles. There's no guarantee, that AI would manage to capture any town/castle; do this before new week starts and surrendered hero disappears from the tavern. Surrendering also provides enemy more money increasing enemy's chances to win and decreasing homeless AI's chances to survive because of spending too much money.

So in the most uniform case surrendering from such "final" or "semi-final" battles brings more benefits to attacking side and decreases chances to survive for suffering side.

So, I suggest that AI should retreat only if hero really has **valuable** artifacts, so attacker won't get any. In the common battles in this case AI should prefer retreating. From the siege - surrender and leave attacker without **valuable** artifacts (if there're any).

oleg-derevenetz:
Hi @Branikolog

> There's no 100% guarantee that AI would manage to capture another town

It doesn't mean that it wouldn't.

> Firstly, it's quite risky to allow attacker win with less losses as he can defend the castle with higher chance especially considering, that surrenered hero couldn't be hired to reinforce initial owner of the castle.

Please see specific values of the corresponding ratio in #8321. AI will retreat or surrender when the strength of the enemy army is 10x or more higher than the strength of the AI army (please note that this takes into account not just strength of troops but also magical strength). It is very unlikely that AI can inflict at least some sensitive losses with this ratio, but the probability of losing a hero and giving the enemy artifacts (which will raise his stats even more) is high.

> Secondly, surrendering gives money to the attacker, which helps him to hire someone in that castle or build a defending structure in that castle.

The army at this point is already so small that half of its cost can be neglected. If this is the end of the game (since AI is going to surrender its last castle), there is usually no shortage of money.

> Thirdly, in the current implementation (from this PR) AI doesn't consider valuable artifacts (actually any artifacts at all), and surrenders with empty hero from the last castle.

This can't be happening. In order for the AI to consider the hero for retreat or surrender, either this hero should have level 2+ OR this hero should have non-empty artifacts bag.

> So in the most uniform case surrendering from such "final" or "semi-final" battles brings more benefits to attacking side and decreases chances to survive for suffering side.

I disagree. These few cents that the attacking player can get for the remnants of the AI army are about nothing for the late stage of the game, but the artifacts that the attacking hero can get if AI will fight till the end can be quite valuable.

> So, I suggest that AI should retreat only if hero really has valuable artifacts, so attacker won't get any. In the common battles in this case AI should prefer retreating. From the siege - surrender and leave attacker without valuable artifacts (if there're any).

In regular battles AI will surrender if ALL these conditions are met:

1. Hero is valuable (level 2+ OR some artifacts in the bag - because in HoMM2 there is no way to buy any artifacts for any money);
2. Remnants of his army are more powerful than the average default army he will get in case of retreat;
3. He can pay for the surrender with a margin (see #8321).

If only first condition is met, he will retreat. To surrender during siege, only first and third conditions are required. I believe that there is no need to artificially limit AI in terms of choosing between retreat and surrender.

Branikolog:
@oleg-derevenetz 

If we look at the situation from the save I've attached in https://github.com/ihhub/fheroes2/issues/8292#issue-2085152995 with current implementation.

We attack the castle. Enemy has much weaker army and retreats in the early stages of the battle. This is his last castle and hero. So he pays us 4200 gold, inflicts ~0 damage to our forces and disappears from this land. 
I want to note here, that he has 0 artifacts (the only reason in my opinion he potentially could consider surrendering in this situation). 
So in my opinion the decision to surrender was rather controvertial here.
AI hero could stand a few turns more inflicting more damage to our army. Surely, if he had some powerful artifact surrendering could have sence, but this hero was "empty" in this case. He anyway loses so bringing attacker more problems during the siege is a nice strategy rather then surrendering into nowhere and giving extra 4200. 

Surrendering from a siege in this case allows us to save more troops and get some money which plays against initial owner of the castle (if he even has more heroes that could potentially start besieging castle back) and his allies. It is definitely two huge cons of this decision: 2x" ➖ ". Surrendering although prevents us from getting artifacts from the defending hero. That's a " ➕ " for sure. But in a particular case he had 0.
Also, there's no guarantee, that initial owner of the castle would be able to hire his hero back till the end of the week. I'm sure, AI is not considering day of the week when retreating, so he can potentially loose his hero into nowhere surrendering on 5th or 6th day of the week and being too far away to reach any castle to capture and re-hire. Generally it's rather a " ➖ " for this decision. 

I'm not against of AI retreating with his last hero (which is a bit unusual for me as I got used to H3 logic of "fight till death" in a case AI has no castles and has no where to run), but in a case of defending a castle with his last hero I see no reason to surrender except being so greedy to give artifacts to the winner. :) As a human player I could initiate surrendering if I would be sure, I have other hero that has a chance to capture a town/castle and re-hire that defending hero. But we don't have such complex logic yet. In a common case in my opinion, retreating from the last owned castle brings quite a high chance for AI to loose that hero.
For the case there's a battle in a field (and it is not the last hero) the proposed rules suit well: hero can surrender/retreat and potentially be re-hired later. At least this looks logical. 
When fighting with the last hero there's no other reason to retreat than bury all artifacts belonging to the last hero with him in the deepest darksides of the tavern. :D

> In order for the AI to consider the hero for retreat or surrender, either this hero should have level 2+ 

Maybe lvl 2 is a little too low, but I haven't thought about it yet while just trying to find a consensus with you in the question. :)

oleg-derevenetz:
@Branikolog OK, now the rules are as follows:

Retreat:

1. If the hero has artifacts, he should in any case consider retreating;
2. Otherwise, if this hero is the last one, and the kingdom has no castles, then there is no point in retreating;
3. Otherwise, if this hero is relatively experienced (his level >= X where X can be difficulty-dependent), then he can consider retreating.

Surrender:

1. If the hero has artifacts, he should in any case consider surrendering;
2. Otherwise, if this hero is the last one, and either the kingdom has no castles, or this hero is defending the last castle, then there is no point in surrendering;
3. Otherwise, if this hero is relatively experienced (his level >= X where X can be difficulty-dependent), then he can consider surrendering.

oleg-derevenetz:
Hi @zenseii all artifacts are considered valuable, both in this PR and in `master` branch. Actually AI-controlled hero may have both cursed and useful artifacts at the same time, so If we consider the surrender in terms of the value of artifacts, then this is not such a simple question. Anyway, I propose to consider improvements of such kind separately.

Branikolog:
Hi, @zenseii 

Sure, if a hero has only cursed artifacts then he shouldn't consider them in this case.
What if AI hero has boots, for instance and a cursed artifact (- morale) at the same time? Or playing as necromant with the same artifact set? I think the logic in this case shoud be even more complex, considering the value of all artifacts and other special conditition.

But I believe it's out of the scope of this PR.

Hi, @oleg-derevenetz 

Could you, please, explain a bit 3rd point of surrendering/retreating? How is this implemented in the latest commit? 
Is this works like the higher difficulty we're playing, the weaker heroes (up to 2lvl) should retreat (I mean when all other conditions allow retreating)?
I personally expect AI consider retreating on "Expert" and "Impossible" when the hero is lvl2 or higher, "Hard" - lvl3 or higher, "Normal" and "Easy" - Lvl4 or higher.

oleg-derevenetz:
Hi @Branikolog

> Could you, please, explain a bit 3rd point of surrendering/retreating? How is this implemented in the latest commit?
Is this works like the higher difficulty we're playing, the weaker heroes (up to 2lvl) should retreat (I mean when all other conditions allow retreating)?
I personally expect AI consider retreating on "Expert" and "Impossible" when the hero is lvl2 or higher, "Hard" - lvl3 or higher, "Normal" and "Easy" - Lvl4 or higher.

So far, only an infrastructure has been created for such a separation, which can be configured separately in scope of #8321. Currently for all difficulties the minimum hero level for surrender and retreat is 3. The same is in the `master` branch, but in `master` branch it is hardcoded and independent of the game difficulty.

Branikolog (approved):
@oleg-derevenetz 
Thanks for the answer! 
With the latest changes AI is not surrendering or retreating with weak and artifact-poor heroes. AI is saving artifacts but fights till the end, when there's no reason to save his last hero!

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -620,14 +587,101 @@ namespace AI
...
+                const int gameDifficulty = Game::getDifficulty();
+                const bool isGameCampaign = Game::isCampaign();
+
+                // TODO: consider taking speed/turn order into account in the future
+                if ( _myArmyStrength * Difficulty::getArmyStrengthRatioForAIRetreat( gameDifficulty ) >= _enemyArmyStrength ) {
+                    return Outcome::ContinueBattle;
+                }
+
+                const bool hasArtifacts = [actualHero]() {
+                    const BagArtifacts & artifactsBag = actualHero->GetBagArtifacts();
```
ihhub, line 608: Here we need to be smarter: a hero can have purely bad artifacts so they bring no value and actually (in the future) we should consider them as a sign not to retreat :) As of now I suggest to check this:
```cpp
const auto & artifactData = fheroes2::getArtifactData( art.GetID() );
if ( !artifactData.bonuses.empty() || !artifactData.curses.empty() ) {
    return false;
}
```

ihhub (changes requested):
Hi @oleg-derevenetz , I left one suggestion. Would you mind to check it?

src/fheroes2/ai/normal/ai_normal_battle.cpp
oleg-derevenetz: Hi @ihhub

> a hero can have purely bad artifacts so they bring no value and actually (in the future) we should consider them as a sign not to retreat :)

Well, since we're talking about it, we can do it now. Now, if the hero has ONLY cursed artifacts (i.e. does not have a single artifact with non-empty bonuses), and is not relatively experienced, he will not retreat or surrender.

ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for these changes!

PR #8299, opened 2024-01-18
Add infos on the cost of surrender

zenseii:
This discloses how much the diplomacy skill affects the cost of surrender, and adds that the cost of surrender is under normal circumstances half the total cost of troops remaining in the army, in gold only.

image

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2023                                             *
+ *   Copyright (C) 2019 - 2024                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2010 by Andrey Afletdinov <fheroes2@gmail.com>          *
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -1000,7 +1000,7 @@ int Battle::Arena::DialogBattleHero( const HeroBase & hero, const bool buttons, 
         else if ( le.MousePressRight( btnSurrender.area() ) && _currentColor == hero.GetColor() && !hero.isCaptain() ) {
             fheroes2::showStandardTextMessage(
                 _( "Surrender" ),
-                _( "Surrendering costs gold. However if you pay the ransom, the hero and all of his or her surviving creatures will be available to recruit again." ),
+                _( "Surrendering costs gold. However if you pay the ransom, the hero and all of his or her surviving creatures will be available to recruit again. The cost of surrender is half of the total cost of the troops remaining in the army." ),
                 Dialog::ZERO );
         }
         else if ( le.MousePressRight( portraitArea ) && actionHero != nullptr ) {
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -1000,7 +1000,7 @@ int Battle::Arena::DialogBattleHero( const HeroBase & hero, const bool buttons, 
         else if ( le.MousePressRight( btnSurrender.area() ) && _currentColor == hero.GetColor() && !hero.isCaptain() ) {
             fheroes2::showStandardTextMessage(
                 _( "Surrender" ),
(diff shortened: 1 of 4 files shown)


src/fheroes2/heroes/skill.cpp
```diff
@@ -507,20 +507,30 @@ std::string Skill::Secondary::GetDescription( const Heroes & hero ) const
         str = _n( "%{skill} increases the hero's viewable area by one square.", "%{skill} increases the hero's viewable area by %{count} squares.", count );
         break;
     }
-    case DIPLOMACY:
-        str = _( "%{skill} allows the hero to negotiate with monsters who are weaker than their army." );
+    case DIPLOMACY: {
+        str = _( "%{skill} allows the hero to negotiate with monsters who are weaker than their army, and reduces the cost of surrender." );
+        str += "\n\n";
+        uint32_t percent = 40;
```
oleg-derevenetz, line 513: Hi @zenseii Eh, I'm aware of the fact that these values are currently hardcoded in the `Battle::Force::GetSurrenderCost()`, but since they are now hardcoded also here, it would be better to add at least a free function to the `Skill` namespace to return this percent and use it in both places. What do you think?

src/fheroes2/heroes/skill.cpp
```diff
@@ -855,6 +863,20 @@ uint32_t Skill::GetNecromancyPercent( const HeroBase & hero )
     return std::min( percent, 100u );
 }
 
+uint32_t Skill::GetDiplomacySurrenderPercent( const int level )
+{
+    switch ( level ) {
```
ihhub, line 868: Shouldn't we return 20, 40 and 60 values respectively? At line 526 Expert level will show 20% while Basic will show 40%.

Also to avoid hardcoded values in `battle_army.cpp` we have:
```cpp
const int diplomacyLevel = commander->GetLevelSkill( Skill::Secondary::DIPLOMACY );
if ( diplomacyLevel > Skill::Level::NONE ) {
    mod *= ( Skill::GetDiplomacySurrenderPercent( diplomacyLevel ) * 0.02 );
}
```

but we can do with the proposed changes:
```cpp
mod *= 1 - ( Skill::GetDiplomacySurrenderPercent( commander->GetLevelSkill( Skill::Secondary::DIPLOMACY ) / 100.0 );
```

ihhub (changes requested):
Hi @zenseii , I left one suggestion here. Would you mind to take a look?

src/fheroes2/heroes/skill.h
```diff
@@ -44,6 +44,7 @@ namespace Skill
 
     uint32_t GetNecromancyBonus( const HeroBase & hero );
     uint32_t GetNecromancyPercent( const HeroBase & hero );
+    uint32_t GetDiplomacySurrenderPercent( const int level );
```
oleg-derevenetz, line 47: Hi @zenseii I'm not insisting, but maybe call this function something like `GetDiplomacySurrenderCostDiscount`/`GetDiplomacySurrenderCostDiscountPercent`?

src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -1000,7 +1000,7 @@ int Battle::Arena::DialogBattleHero( const HeroBase & hero, const bool buttons,
         else if ( le.MousePressRight( btnSurrender.area() ) && _currentColor == hero.GetColor() && !hero.isCaptain() ) {
             fheroes2::showStandardTextMessage(
                 _( "Surrender" ),
-                _( "Surrendering costs gold. However if you pay the ransom, the hero and all of his or her surviving creatures will be available to recruit again." ),
+                _( "Surrendering costs gold. However if you pay the ransom, the hero and all of his or her surviving creatures will be available to recruit again. The cost of surrender is half of the total cost of the troops remaining in the army." ),
```
oleg-derevenetz, line 1003: I should note however that this comment is not 100% accurate. E.g. while non-true-resurrected troops "remain in the army", their cost is not accounted for surrender cost calculation purposes. The same applies to elementals summoned during the battle. But I understand that it is unrealistic to describe all this rather complex mechanics in a couple of sentences.

oleg-derevenetz (approved):

src/fheroes2/battle/battle_dialogs.cpp
oleg-derevenetz: > "only... against an opposing hero" : in fheroes2 you can surrender if you attack a castle with a captain, probably not if they don't even have that. I haven't checked this up against the original game.

In OG you can surrender to a captain as well:

https://github.com/ihhub/fheroes2/assets/32623900/d7a6f631-79a7-4f48-adda-143e96fea545

> "based on the power of your army." could suggest that paying for summons, mirror images and non-true resurrected creatures is intentional

At the time, this was considered a bug of the original game: #5168.

ihhub (approved):

ihhub:

PR #8301, opened 2024-01-18
Moved the warning about losing all the castles to the beginning of the turn.

dood-apo:
There is no reason for displaying it at the end of the turn, except for the fact that OG did it.
Let's be better!

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -433,8 +433,14 @@ void ShowWarningLostTownsDialog()
         Game::DialogPlayers( myKingdom.GetColor(), _( "Beware!" ),
                              _( "%{color} player, this is your last day to capture a town, or you will be banished from this land." ) );
     }
-    else if ( lostTownDays > 0 && lostTownDays <= Game::GetLostTownDays() ) {
-        std::string str = _( "%{color} player, you only have %{day} days left to capture a town, or you will be banished from this land." );
+    else if ( lostTownDays > 0 ) {
+        std::string str;
+        if ( lostTownDays <= Game::GetLostTownDays() ) {
+            str = _( "%{color} player, you only have %{day} days left to capture a town, or you will be banished from this land." );
+        }
+        else {
+            str = _( "%{color} player, you have lost your last town. If you do not conquer another town in next week, you will be eliminated." );
+        }
         StringReplace( str, "%{day}", lostTownDays );
         Game::DialogPlayers( myKingdom.GetColor(), _( "Beware!" ), str );
     }
@@ -1399,16 +1405,8 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
 
         if ( myKingdom.isPlay() ) {
             // these warnings should be shown at the end of the turn
-            if ( myCastles.empty() ) {
-                const uint32_t lostTownDays = myKingdom.GetLostTownDays();
-
-                if ( lostTownDays > Game::GetLostTownDays() ) {
-                    Game::DialogPlayers( conf.CurrentColor(), _( "Beware!" ),
-                                         _( "%{color} player, you have lost your last town. If you do not conquer another town in next week, you will be eliminated." ) );
-                }
-                else if ( lostTownDays == 1 ) {
-                    Game::DialogPlayers( conf.CurrentColor(), _( "Defeat!" ), _( "%{color} player, your heroes abandon you, and you are banished from this land." ) );
-                }
+            if ( myCastles.empty() && myKingdom.GetLostTownDays() == 1 ) {
+                Game::DialogPlayers( conf.CurrentColor(), _( "Defeat!" ), _( "%{color} player, your heroes abandon you, and you are banished from this land." ) );
             }
 
             if ( !conf.isAutoSaveAtBeginningOfTurnEnabled() ) {


oleg-derevenetz:
> There is no reason for displaying it at the end of the turn, except for the fact that OG did it.

In fact, this is not entirely true. The display of this message at the end of the turn is due to the fact that the first of the remaining 7 days ("next week") actually starts _tomorrow_ for the hero:

https://github.com/ihhub/fheroes2/assets/32623900/4150ed7e-b552-4f52-8fdb-aca0a2661183

Moving this message to the beginning of the turn looks misleading because it gives the impression that "next week" starts _today_, and then at the beginning of the next turn player gets a message that "you have 7 days", and he's like "what?".

> Let's be better!

This is at least controversial.

dood-apo:
Right now the warning is issued at the end of the homeless player's turn. Then on the beginning of the homeless player's next turn another one pops up with the countdown. What exactly the player can do with that first warning? He doesn't play in between of those two. In the current form it is superfluous.

The message says "If you do not conquer another town in next week", it doesn't say that the current turn is part of "next week". Btw it should say "the next week", but that is a topic for another PR since it impacts all the translation files.

dood-apo:
> If shown at the beginning of the turn, it's not entirely clear whether the current day is counted, because it is just begun (for this player), so it is also sort of "upcoming". At least I suspect that many people may think like this.

Sure. But does it matter that much? If anything a player will be happy to see that he still has 7 days left to find home instead of 6 when he sees the next popup. At least start of turn message can be acted upon. End of turn message is redundant with the upcoming countdown popup.

oleg-derevenetz:
Hi @ihhub @zenseii @Branikolog could you please share your opinion on this proposal?

Branikolog:
Hi, @oleg-derevenetz 

I personally have been always confused by this case. :D I mean that fact, that "in week" includes the 8th day. 
I remember the days, when we discussed this unique case, when player starts without castles and has actually 8 days to capture any castle. :)
I think it is better to warn a player on the first seconds of the game, that he has no castles and has to hurry. Just explain him a rules on the early beginning. 
Imagine you're novice player. You start the game. You have a few heroes and make some actions on the first day. But at the end of the day you have a message, that something is wrong with your kingdom and you have to solve this problem. If people are confused by both of these messages (saying in a week and then saying in 7 days) we're currently discussing, we can exclude the wording "in the next week" from the first message and show it in the beginning of the game. 
Also, I'm not happy with the wording of the message itself, as player didn't lose castles, he simply don't have any. He has 0 castles by the idea of mapmaker in the beginning of the scenario, right? So in my opinion, the message should better say something like: "... player, you have no town under your control currently. If you don't conquer any you will be eliminated soon!"

On the other hand what prevents us from telling the player the exact day he would loose? That would exclude the mess as player would know the exact day he is going to be eliminated. 
"... player, you have no town under your control currently. If you don't conquer any town you will be eliminated on 1 month, 2 week and 1 day! "

So I prefer showing a message on the start of the first day of the scenario, but I find current message not suited well for that purpose.

Other thing, that bothers me is the fact, that we don't show similar message, that player lost his last castle if enemy captured the last one. Currently, it says, that "you have 7 days to...." I think it would be nice to point a player, the reason he has to rush. 
It could be a combination of both current messages:
"... player, you have lost your last town and you have only 7 days to capture a town, or you will be banished from this land."
On the first day he starts his turn without a castle only. Current wording is OK for all other days.
But it is definitely out of the scope of this PR for sure.

dood-apo:
> Other thing, that bothers me is the fact, that we don't show similar message, that player lost his last castle if enemy captured the last one.

We actually do if the enemy precedes that player in the turn order. At least this is the case in hot seat, I didn't take time to reproduce it in "standard" with AI playing before me in the turn order, but looking at the code it should make no difference. It's one of those turn/day things. First player has a great advantage of ability to attack on day 1 with his fresh growth (before player 2 has a chance to buy his new troops), player 2 on the other hand has one more turn to look for a new home if he has lost one.

> You only have today/this turn and the next 7 days

I like this one! But I wouldn't use word "turn". Everything is reported in moths/weeks/days and this has its own charm.

dood-apo:
> > has one more turn to look for a new home if he has lost one.
> 
> This is incorrect. They have the same amount of days and therefore turns.

How so? If player 1 takes all the castles from player 2 and passes the turn, then player 2 has the turn in the current day and also subsequent 7 days to find a town. However if player 2 takes all the castles from player one, his passing of the turn triggers the new day. Player one will not have a turn in the day he lost his castle, because he has already took one (when he still had a castle). He is met on his following turn with a direct info that he has 7 days left. So in my book player 2 has **8 turns** to act, while player 1 has **7 turns**.

dood-apo:
> > So in my book player 2 has **8 turns** to act, while player 1 has **7 turns**.
> 
> I see what you mean now and yes this is right if we specifically talk about turns "after you're displayed a dialog about having lost the last town", however they've all had the same amount of turns to act, it's just that the preceding player didn't have a turn they had lost all castles.

Let's forget all the dialogs for a moment. What I mean is that there is a difference in turns (not days) left to find home depending on your placement in the turn order.

> What this also means is that the first player will only ever see the message about "today" if they start the map without a castle.
As for other players they will see the message about "today" if a preceding player takes their castle and if they start the map without a castle, but not otherwise, e.g. the 4th player takes the 3rd player's last castle.

Correct, but this is the current behavior (for OG and fheroes2) and is orthogonal to moving the message to the start of the turn. The message `%{color} player, you have lost your last town. If you do not conquer another town in next week, you will be eliminated` has dual purpose - game start with no towns or warning about loosing all towns **in this current day** (because some other player **taking turn earlier this day** took all your castles). However both reasons play well with warning on the start of the turn in my opinion.

dood-apo:
> the OG gives 1 day less to every player except for the first player

Just verified with the OG. You are right that there are differences. In my opinion the OG is fair in this regard, and fheroes2 is not. As I explained earlier, in hotseat fheroes2 the second player has 8 turns to act, the first 7. In the OG as soon as your very first turn starts without a castle (no matter if you are P1 or P2) you are prompted with the countdown from 7. It is fheroes2 which gives additional day to players following the aggressor in the turn order.

If the OG is bugged in this regard then I would say it is in cases where you start without a town (you've got 8 days to act).

oleg-derevenetz:
> In my opinion the OG is fair in this regard, and fheroes2 is not.

In OG, the "next week" sometimes starts tomorrow (if second player in the turn queue captures the last castle of the first player in that queue) and sometimes it starts today (if first player captures the last castle of the second player), which looks a bit contradictory. In fheroes2, the "next week" always starts tomorrow. Yes, in this case, the second player has one more turn, that's correct, but this is compensated by the fact that the second player in turn queue always loses (in a way) in terms of initiative. Which option is more "fair" here is a debatable question.

dood-apo:
> In OG, the "next week" sometimes starts tomorrow (if second player in the turn queue captures the last castle of the first player in that queue) and sometimes it starts today (if first player captures the last castle of the second player), which looks a bit contradictory

Right, but the OG does not display `%{color} player, you have lost your last town. If you do not conquer another town in next week, you will be eliminated` at all in this case. Just starts with the countdown. So no contradictions there.

> Yes, in this case, the second player has one more turn in reserve, that's correct, but this is compensated by the fact that the second player in turn queue always loses (in a way) in terms of initiative. Which option is more "fair" here is a debatable question.

I've only said that "the OG is fair in this regard", I didn't say anything about the general balance of multiplayer games. Now you're saying that indeed there is a difference with time left to capture the castle, but it is intentional to provide a better balance. I don't buy this explanation, as this is a super fringe case. When you're without castles your most likely done anyway. 8 days of wandering around instead of 7 can't cover for any handicaps that the preceding player has over you.

oleg-derevenetz:
> I don't buy this explanation, as this is a super fringe case. When you're without castles your most likely done anyway. 8 days of wandering around instead of 7 can't cover for any handicaps that the preceding player has over you.

Well, it's your opinion. I just read all this thread, and funny thing that you made similar arguments in the following comment:

https://github.com/ihhub/fheroes2/pull/8301#issuecomment-1899906649

:) Anyway, IIRC there were similar discussions in the past somewhere, and I propose not to change the game logic back and forth endlesdly.

dood-apo:
> I just read all this thread, and funny thing that you made similar arguments in the following comment

Yeah, seems funny indeed. I just wanted to explain how the game currently behaves, because it seemed to me that it was misunderstood. Later down the thread when @zenseii brought to my attention that the OG is different in this matter, it struck me that the OG behaves naturally and this behavior doesn't even need that darn dialog from the PR :).

> Anyway, IIRC there were similar discussions in the past somewhere, and I propose not to change the game logic back and forth endlesdly.

Sure. I'm kind of new here, apologies if I haven't read all the past discussions.

oleg-derevenetz:
> Sure. I'm kind of new here, apologies if I haven't read all the past discussions.

That's OK. fheroes2 really has many differences from OG, and these "improvements" were somehow justified by their "lobbyists" at the time. For instance, in OG summoned elementals and not non-true-resurrected units are taken into account when surrender cost is calculated, but in fheroes2 it's not the case, you pay for what you get. Also, OG remember just one surrendered (or retreated) hero, and if hero retreated on day 7, he will be lost at day 1, so there can be the case when you will not be able to hire it. In fheroes2, that's not the case, its behavior is more like HoMM3 in this regard (see #8304 for details). At the time, there were even HoMM3-like "wait" and "castle guardian" (two heroes in a castle) mechanics, but they were cleaned up before the 1.0 release.

PR #8310, opened 2024-01-20
Revert changes for Resize() function

ihhub:
regression from #8276

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -2823,6 +2823,7 @@ namespace fheroes2
         else {
             const uint8_t * imageOutYEnd = imageOutY + widthOut * heightRoiOut;
             int32_t idY = 0;
+
             // Pre-calculation of X position
             std::vector<int32_t> positionX( widthRoiOut );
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @Districh-ru , I have to revert this change for now as big font is completely broken.

Districh-ru:
Hi, @ihhub, sad. I'll check it and fix my code in another PR.

src/engine/image.cpp
```diff
@@ -2781,63 +2770,52 @@ namespace fheroes2
...
-
-                        if ( isOutNotSingleLayer ) {
-                            *transformOutX = *transformInX;
-                            ++transformOutX;
+                        else {
+                            *imageOutX = *imageInX;
                         }
-                    }
 
-                    if ( !isOutNotSingleLayer ) {
```
Districh-ru, line 2839: Here was my introduced bug: if ( **!** isOutNotSingleLayer ) {
And is should be ``` if ( isOutNotSingleLayer ) { ```

PR #8012, opened 2023-11-02
Modernize cursor update for the game area

Districh-ru:
Based on #7808.

This PR does the next changes:
 - fixes #8026;
 - pass `gameArea` to the `QuickInfo()` functions;
 - restore cursor immediately after the Quick Info dialog is closed if no cursor change is predicted;
 - schedule immediate cursor update after Quick Info dialog is closed;
 - set `POINTER` cursor when map dragging is started (it is made to show the player that no action will be performed after the mouse button is released, discussible, can be reverted);
 - add forced software cursor render on the screen after cursor icon is updated;
 - remove the forced cursor update setting on every `CURRENT_HERO_DELAY` pass: this part needs testing, the assertions are temporary put instead;
 - schedule the cursor update after Puzzle or View World dialog is closed;
 - make View World use reference to the `gameArea` of given `interface` (AdventureMab/EditorInterface);
 - do not make `gameArea` copies for the View World, use references to the original `gameArea`.

PR #8316, opened 2024-01-20
View World rendering improvement

Districh-ru:
This PR is a part of #8012 and includes:
- fix of some SonarQube and Clang-Tidy issues;
- make View World use reference to the gameArea of given interface (AdventureMap / EditorInterface);
- do not make gameArea copies for the View World (use reference).
- schedule the cursor update after View World is closed.

### DIFF
--- a/src/fheroes2/kingdom/view_world.cpp
+++ b/src/fheroes2/kingdom/view_world.cpp
@@ -25,6 +25,7 @@
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <utility>
 
 #include "agg_image.h"
@@ -32,7 +33,6 @@
 #include "color.h"
 #include "cursor.h"
 #include "game_hotkeys.h"
-#include "game_interface.h"
 #include "gamedefs.h"
 #include "heroes.h"
 #include "icn.h"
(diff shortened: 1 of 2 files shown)


src/fheroes2/kingdom/view_world.cpp
```diff
@@ -218,9 +216,11 @@ namespace
             temporaryImg._disableTransformLayer();
             temporaryImg.resize( redrawAreaWidth, redrawAreaHeight );
 
-            Interface::GameArea gamearea = Interface::AdventureMap::Get().getGameArea();
-            gamearea.SetAreaPosition( 0, 0, redrawAreaWidth, redrawAreaHeight );
+            // Remember the original game area ROI and center of the view.
+            const fheroes2::Rect gameAreaRoi( gameArea.GetROI() );
```
ihhub, line 220: This variable can be const reference.

src/fheroes2/kingdom/view_world.h
```diff
@@ -50,25 +50,42 @@ class ViewWorld
...
 
-        ZoomLevel _zoomLevel{ ZoomLevel::ZoomLevel1 };
-        fheroes2::Point _center;
-        std::array<fheroes2::Rect, 4> _roiForZoomLevels;
+        ZoomLevel getZoomLevel() const
+        {
+            return _zoomLevel;
+        }
+
+        fheroes2::Point getCenter() const
```
ihhub, line 74: We should return const reference from this method.

src/fheroes2/kingdom/view_world.cpp
```diff
@@ -612,9 +610,9 @@ void ViewWorld::ViewWorldWindow( const int32_t color, const ViewWorldMode mode,
     // Creates fixed radar on top-right, suitable for the View World window
     Interface::Radar radar( interface.getRadar(), fheroes2::Display::instance() );
 
-    const Interface::GameArea gameArea = interface.getGameArea();
-    const fheroes2::Rect worldMapROI = gameArea.GetVisibleTileROI();
-    const fheroes2::Rect visibleScreenInPixels = gameArea.GetROI();
+    Interface::GameArea & gameArea = interface.getGameArea();
```
ihhub, line 613: I remember there was a reason why we made a copy of the game area so please check that we don't mess with Adventure Map while using the existing game area. However, logically this change should not break any functionality.

ihhub (changes requested):
Hi @Districh-ru , I put few comments in this pull request. Can you please check them once you have time?

src/fheroes2/kingdom/view_world.cpp
Districh-ru: Hi, @ihhub, just after these lines we change the game area ROI using `gameArea.SetAreaPosition()`. I thought that in this case it is better to have a copy, not a cached const reference to the previous state of this changed Rect.

src/fheroes2/kingdom/view_world.cpp
Districh-ru: We made a copy because we changed game area ROI and Center in `CacheForMapWithResources` constructor.
Now these data is cached and restored so we can use the original gameArea.

src/fheroes2/kingdom/view_world.cpp
Districh-ru: Using const reference here results in crash. :(

ihhub: Oh, yeah. We are changing the size.

ihhub (approved):

ihhub:
@Districh-ru , thanks a lot for these enhancements!

PR #8317, opened 2024-01-20
Fix OKAY button translation for Interface Settings dialog after any setting change

### LINKED ISSUE #8315: Okey button in the interface settings menu changes itself

When using any other language other then default english and going into the interface settings and changing the scroll speed, the button changes to the english "Okey" button that that also looks different. If you click around on the other settings then it changes back.
Also the smaller buttons for the other languages seems to be missing the "shine" effect, but that's maybe something you are aware of, but i'll guess that's another issue.

Districh-ru:
Fix #8315


ihhub (approved):

ihhub:

PR #8322, opened 2024-01-21
Cursor update code improvements

Districh-ru:
This PR is a part of #8012. It contains:
- fixes for some Clang-Tidy and SonarQube issues;
- minor optimization of CursorRestorer:
    - update cursor theme only if cursor visibility is true. There is no difference what theme the invisible cursor uses so theme update is redundant;
    - do not do checks for theme and visibility in destructor because it is done in `SetThemes()` for themes and `show()` method for visibility is inlined.

### DIFF
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2023                                             *
+ *   Copyright (C) 2019 - 2024                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2009 by Andrey Afletdinov <fheroes2@gmail.com>          *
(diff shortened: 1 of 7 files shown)


oleg-derevenetz:
> update cursor theme only if cursor visibility is true. There is no difference what theme the invisible cursor uses so theme update is redundant

Can't it be like that somewhere:

1. `CursorRestorer` has been used to set the cursor theme and temporary hide it;
2. Cursor then has been shown "by hand"

? At least this looks like a valid scenario, and cursor theme optimization may become a hidden "gotcha", what do you think?

Districh-ru:
> Can't it be like that somewhere:
> 
>     1. `CursorRestorer` has been used to set the cursor theme and temporary hide it;
> 
>     2. Cursor then has been shown "by hand"
> 
> 
> ? At least this looks like a valid scenario, and cursor theme optimization may become a hidden "gotcha", what do you think?

Hi, @oleg-derevenetz, yes it can be. And in this case we will need to set cursor theme by hand.
Currently restorer is used to do one change for an opened dialog and restore the cursor after dialog is closed.

This optimization was made not do do many changes in our code where cursor is hidden by restorer to `const CursorRestorer cursorRestorer( false, Cursor::Get().Themes() );`

I think, I'll add a new constructor `CursorRestorer( const bool visible );` without theme.

PR #8338, opened 2024-01-23
Clang-Tidy comments workflow: junk the paths while unzipping the user-provided ZIP archive

oleg-derevenetz:
Normally, Info-ZIP's `unzip` command should remove "parent dir" path components (''../'') from the names of extracted files. However, since we don't need to recreate some directory structure here at all, It's better to be safe than sorry.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so this precaution step!

PR #8342, opened 2024-01-25
Change the logic to handle the future map loading

ihhub:
- add checks for objects placement on the Adventure Map
- add abort logic for history manager to revert changes back to the previous state if there is an error
- add logic to load objects on the Adventure Map for the future map loading

relates to #6845

### DIFF
--- a/src/fheroes2/editor/history_manager.cpp
+++ b/src/fheroes2/editor/history_manager.cpp
@@ -112,7 +112,6 @@ namespace fheroes2
 
     ActionCreator::~ActionCreator()
     {
-
         auto * action = dynamic_cast<MapAction *>( _action.get() );
         if ( action == nullptr ) {
             // How is it even possible?
--- a/src/fheroes2/editor/history_manager.cpp
+++ b/src/fheroes2/editor/history_manager.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2023                                                    *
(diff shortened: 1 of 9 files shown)


src/fheroes2/editor/history_manager.h
```diff
@@ -58,10 +58,17 @@ namespace fheroes2
 
         ActionCreator & operator=( const ActionCreator & ) = delete;
 
+        void abort()
```
oleg-derevenetz, line 61: Hi @ihhub what do you think about the reverse approach: add `commit()` method which should be called when everything is done properly, and abort in destructor if `commit()` wasn't called?

ihhub: Yup, it makes more sense. I shall change it.

Districh-ru (approved):

src/fheroes2/editor/history_manager.cpp
```diff
@@ -110,11 +110,11 @@ namespace fheroes2
...
-    ActionCreator::~ActionCreator()
+    void ActionCreator::commit()
     {
         auto * action = dynamic_cast<MapAction *>( _action.get() );
         if ( action == nullptr ) {
-            // How is it even possible?
+            // How is it even possible? Did you call this method twice?
             assert( 0 );
             return;
         }
```
oleg-derevenetz, line 120: Why do we need the `try`/`catch` block below with just an `assert()` call in `catch`?

ihhub: I've removed it.

PR #8343, opened 2024-01-26
Add quick hero switch feature to the adventure map interface

### LINKED ISSUE #8260: Adventure map, Allow to select a hero by clicking on him (+ holding a key) on the worldmap
### Preliminary checks

Currently, finding the right hero to select is a tedious task. It can be very time-consuming when you're managing hero chains and switching the army from one hero to another, for example.

A solution would be to be able to select the hero directly by clicking on it on the map. On heroes 3 Horn of the Abyss, they chose to hold down the shift button while clicking on the hero to select it, and it's really great to use. It's a real time-saver

### Additional info

_No response_

zenseii:
By default hold shift and left-click on a hero to change focus to that hero from the currently focused one, or you can hold down the left click for some time to switch (for Android).

Fix #8260 

https://github.com/ihhub/fheroes2/assets/12501091/54eb84be-15ea-4614-8917-9d8bc4894884


src/fheroes2/game/game_startgame.cpp
```diff
@@ -1121,6 +1124,22 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
                 EventOpenFocus();
         }
 
+        if ( HotKeyHoldEvent( Game::HotKeyEvent::WORLD_QUICK_SELECT_HERO ) && GetFocusHeroes() ) {
```
ihhub, line 1127: Should we move this condition just 2 lines above inside `if ( le.KeyPress() ) {` body? It also this code:
```cpp
            if ( isMovingHero ) {
                stopHero = true;
            }
```
so I don't think we need the same code here.

ihhub, line 1127: We check `GetFocusHeroes()` twice: at lines 1127 and 1135 (aka `selectedHero`). I recommend to remove the one at line 1127.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -1121,6 +1124,22 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
...
+                stopHero = true;
+            }
+            const fheroes2::Point & mousePosition = le.GetMouseCursor();
+            const int32_t index = _gameArea.GetValidTileIdFromPoint( mousePosition );
+            Heroes * selectedHero = world.GetTiles( index ).getHero();
+            const Heroes * currentHero = GetFocusHeroes();
+            if ( selectedHero != nullptr && selectedHero != currentHero && selectedHero->GetColor() == currentHero->GetColor() ) {
+                cursor.SetThemes( GetCursorTileIndex( index ) );
+                if ( le.MouseClickLeft() ) {
+                    EventSwitchFocusedHero( selectedHero );
```
ihhub, line 1138: Should we pass `const int32_t tileIndex` to **EventSwitchFocusedHero()** method and do all the logic with current and selected hero there? In this case we can reduce code duplication from here and from method **mouseCursorAreaLongPressLeft()**.

ihhub (changes requested):
Hi @zenseii , I put several comments in this pull request. Could you please check them?

src/fheroes2/game/game_startgame.cpp
```diff
@@ -1119,6 +1122,17 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
             // open focus
             else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_OPEN_FOCUS ) )
                 EventOpenFocus();
+            else if ( HotKeyHoldEvent( Game::HotKeyEvent::WORLD_QUICK_SELECT_HERO ) ) {
+                const fheroes2::Point & mousePosition = le.GetMouseCursor();
+                const int32_t index = _gameArea.GetValidTileIdFromPoint( mousePosition );
```
ihhub, line 1127: Since **mousePosition** is used only in one place we can combine these 2 lines together:
```cpp
const int32_t index = _gameArea.GetValidTileIdFromPoint( le.GetMouseCursor() );
```

ihhub (changes requested):
Hi @zenseii , I added 2 more tiny suggestions. Can you please check them as well?

ihhub (approved):

ihhub:

PR #8348, opened 2024-01-27
Add basic map loading for gameplay

ihhub:
- add logic to get available kingdom colors present on a map
- add logic to get races for each kingdom
- add logic to allow full map loading (random castles and heroes don't work yet and I'm sure many other things)
- fix town passability

relates to #6845

### DIFF
--- a/src/fheroes2/maps/map_format_helper.cpp
+++ b/src/fheroes2/maps/map_format_helper.cpp
@@ -283,6 +283,26 @@ namespace Maps
 
                 map.playerRace[i] = 0;
             }
+
+            // Only one race can be present.
+            if ( ( map.playerRace[i] & Race::RAND ) != 0 ) {
+                map.playerRace[i] = Race::RAND;
+            }
+            if ( ( map.playerRace[i] & Race::MULT ) != 0 ) {
+                map.playerRace[i] = Race::MULT;
+            }
+            else {
+                size_t raceCount = 0;
+                for ( uint8_t raceIdx = 0; raceIdx < 6; ++raceIdx ) {
+                    if ( ( map.playerRace[i] & ( 1 << raceIdx ) ) != 0 ) {
+                        ++raceCount;
+                    }
+                }
+
+                if ( raceCount > 1 ) {
+                    map.playerRace[i] = Race::MULT;
+                }
+            }
         }
 
         map.computerPlayerColors = map.computerPlayerColors & map.availablePlayerColors;
--- a/src/fheroes2/maps/map_format_helper.cpp
+++ b/src/fheroes2/maps/map_format_helper.cpp
@@ -241,8 +241,6 @@ namespace Maps
 
                     const uint32_t race = metadata[1];
                     heroRacesPresent[color] |= ( 1 << race );
-
-                    break;
                 }
                 else if ( object.group == ObjectGroup::KINGDOM_TOWNS ) {
                     assert( object.index < townObjects.size() );
@@ -258,8 +256,6 @@ namespace Maps
 
                     const uint32_t race = townObjects[object.index].metadata[0];
                     townRacesPresent[color] |= ( 1 << race );
-
-                    break;
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -190,4 +193,235 @@ namespace Maps
...
+        static_assert( Race::KNGT == 1 << 0, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::BARB == 1 << 1, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::SORC == 1 << 2, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::WRLK == 1 << 3, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::WZRD == 1 << 4, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::NECR == 1 << 5, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::MULT == 1 << 6, "The race values have changed. You are going to break map format!" );
+        static_assert( Race::RAND == 1 << 7, "The race values have changed. You are going to break map format!" );
+
+        const size_t mainColors{ 6 };
```
oleg-derevenetz, line 216: I suggest to make it `constexpr` instead of `const`. While `const` variables of basic types may be initialized at compile time, this is not guaranteed, because formally it's just an optional compiler optimization.

```suggestion
        constexpr size_t mainColors{ 6 };
```

oleg-derevenetz: Also, there is `KINGDOMMAX` define in `gamedefs.h` that is used to define maximum number of kingdoms/players in the game.

src/fheroes2/maps/map_format_info.h
```diff
@@ -145,6 +145,9 @@ namespace Maps::Map_Format
         uint8_t computerPlayerColors{ 0 };
         std::vector<uint8_t> alliances;
 
+        // Only 6 players are allowed per map.
+        std::array<uint8_t, 6> playerRace{ 0 };
```
oleg-derevenetz, line 149: You may want to use `KINGDOMMAX` here as well.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -190,4 +193,235 @@ namespace Maps
...
+            const auto & mapTile = map.tiles[tileIndex];
+
+            for ( const auto & object : mapTile.objects ) {
+                if ( object.group == ObjectGroup::KINGDOM_HEROES ) {
+                    assert( object.index < heroObjects.size() );
+
+                    const auto & metadata = heroObjects[object.index].metadata;
+
+                    const uint32_t color = metadata[0];
+                    assert( color < heroColorsPresent.size() );
```
oleg-derevenetz, line 240: Can this function be called for an untrusted input (e.g. user-supplied map)? If so, `assert()` is good, but not enough if game was built without assertions (i.e. with `NDEBUG` defined) here and in other similar places.

oleg-derevenetz (commented):
Hi @ihhub I left a few suggestions, please take a look when you have time.

src/fheroes2/maps/map_format_info.h
ihhub: Actually, I want to keep it as hardcoded value to be able to catch any changes at line `assert( map.playerRace.size() == mainColors );`.

src/fheroes2/maps/map_format_helper.cpp
ihhub: It makes sense to call this function even during map loading to prevent any misuse of the game format. I will return bool from this function to verify that everything is fine.

oleg-derevenetz (approved):

ihhub:
Hi @Branikolog , can you please check this pull request ? Please note that random heroes and towns aren't supported yet.

Branikolog:
Hi, @ihhub !
I've done several tests and haven't found any serious problems.
Castle entrance is rendered under the road:
image

PR #8350, opened 2024-01-28
GitHub Actions: Request only the minimum necessary permissions for workflows

oleg-derevenetz:
Most workflows don't need any permissions at all.

* Workflows that use `ncipollo/release-action` need the `contents: write` to be able to upload releases;
* Workflows that create PRs (like `translation_update.yml`) and PR comments (like `clang_tidy_comments.yml`) need the `pull-requests: write` permission;
* `translation_update.yml` workflow also needs the `contents: write` permission to be able to create the PR branch;
* `pr_author_auto_assign.yml` workflow needs the `pull-requests: write` permission to be able to assign PR authors to their PRs.

I also recommend setting the minimum default permissions for the `GITHUB_TOKEN` in repo settings -> Actions -> General:

permissions

This PR also uses `github-script` instead of `bash` to parse untrusted input and set environment variables and imposes additional checks on corresponding values.

### DIFF
--- a/.github/workflows/clang_tidy_comments.yml
+++ b/.github/workflows/clang_tidy_comments.yml
@@ -22,29 +22,42 @@ jobs:
       uses: actions/github-script@v7
       with:
         script: |
-          let artifacts = await github.rest.actions.listWorkflowRunArtifacts({
+          const artifacts = await github.rest.actions.listWorkflowRunArtifacts({
               owner: context.repo.owner,
               repo: context.repo.repo,
               run_id: ${{ github.event.workflow_run.id }},
           });
-          let matchArtifact = artifacts.data.artifacts.filter((artifact) => {
+          const matchArtifact = artifacts.data.artifacts.filter((artifact) => {
               return artifact.name == "clang-tidy-result"
           })[0];
-          let download = await github.rest.actions.downloadArtifact({
+          const download = await github.rest.actions.downloadArtifact({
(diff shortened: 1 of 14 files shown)


.github/workflows/clang_tidy_comments.yml
```diff
@@ -19,29 +22,42 @@ jobs:
...
+          const fs = require("fs");
+          function exportVar(varName, fileName, regEx) {
+              const val = fs.readFileSync("${{ github.workspace }}/clang-tidy-result/" + fileName, {
+                  encoding: "ascii"
+              }).trimEnd();
+              assert.ok(regEx.test(val), "Invalid value format for " + varName);
+              core.exportVariable(varName, val);
+          }
+          exportVar("PR_ID", "pr-id.txt", /^[0-9]+$/);
+          exportVar("PR_HEAD_REPO", "pr-head-repo.txt", /^[-./0-9A-Z_a-z]+$/);
```
oleg-derevenetz, line 59: Unfortunately, I was unable to find any "official" info regarding allowed characters except this:

user-name

and this:

repo-name

ihhub: Thanks for the explanation! I believe we should go with this approach.

ihhub (approved):

ihhub:
Hi @oleg-derevenetz , thank you very much for the changes! I've updated the workflow permissions.

PR #8359, opened 2024-01-31
Auto combat toggling improvements.

### LINKED ISSUE #4387: Combat screen, sometimes it is difficult to switch back to manual mode when Auto combat mode has been activated

dood-apo:
Three improvements in this PR:

1. Conditions for showing interrupt dialog streamlined - we always show the confirmation dialog, regardless which kind of input has triggered the interrupt.
2. It is now possible to interrupt your own auto battle mode on AI's turn.
3. It is now possible to interrupt with left/right mouse click anywhere on the screen.

AI vs AI debug mode battles are unaffected - we are making sure that the color requesting the interrupt belongs to a human.
Human vs human behaves as it suppose to, meaning you can only interrupt your own auto battle status.

Fix #4387

### DIFF
--- a/src/fheroes2/battle/battle_arena.cpp
+++ b/src/fheroes2/battle/battle_arena.cpp
@@ -1249,8 +1249,17 @@ bool Battle::Arena::EnemyOfAIHasAutoBattleInProgress() const
         return false;
     }
 
-    return ( GetCurrentForce().GetControl() & CONTROL_AI ) && !( getEnemyForce( GetCurrentColor() ).GetControl() & CONTROL_AI )
-           && ( _autoBattleColors & getEnemyForce( GetCurrentColor() ).GetColor() );
+    if ( !( GetCurrentForce().GetControl() & CONTROL_AI ) ) {
+        return false;
+    }
+
+    const auto & enemyForce = getEnemyForce( GetCurrentColor() );
+
+    if ( enemyForce.GetControl() & CONTROL_AI ) {
+        return false;
+    }
+
+    return ( _autoBattleColors & enemyForce.GetColor() );
 }
 
 bool Battle::Arena::CanToggleAutoBattle() const
--- a/src/fheroes2/battle/battle_arena.cpp
+++ b/src/fheroes2/battle/battle_arena.cpp
@@ -1253,7 +1253,7 @@ bool Battle::Arena::EnemyOfAIHasAutoBattleInProgress() const
         return false;
     }
 
-    const auto & enemyForce = getEnemyForce( GetCurrentColor() );
+    const Force & enemyForce = getEnemyForce( GetCurrentColor() );
 
     if ( enemyForce.GetControl() & CONTROL_AI ) {
         return false;
--- a/src/fheroes2/battle/battle_arena.cpp
+++ b/src/fheroes2/battle/battle_arena.cpp
@@ -1243,6 +1243,20 @@ bool Battle::Arena::AutoBattleInProgress() const
     return false;
 }
 
+bool Battle::Arena::EnemyOfAIHasAutoBattleInProgress() const
+{
+    if ( _currentUnit == nullptr ) {
+        return false;
+    }
+
+    if ( ( GetCurrentForce().GetControl() & CONTROL_AI ) && !( getEnemyForce( GetCurrentColor() ).GetControl() & CONTROL_AI )
+         && ( _autoBattleColors & getEnemyForce( GetCurrentColor() ).GetColor() ) ) {
+        return true;
+    }
+
+    return false;
+}
+
 bool Battle::Arena::CanToggleAutoBattle() const
 {
     if ( _currentUnit == nullptr ) {
--- a/src/fheroes2/battle/battle_arena.cpp
+++ b/src/fheroes2/battle/battle_arena.cpp
@@ -1249,12 +1249,8 @@ bool Battle::Arena::EnemyOfAIHasAutoBattleInProgress() const
         return false;
     }
 
-    if ( ( GetCurrentForce().GetControl() & CONTROL_AI ) && !( getEnemyForce( GetCurrentColor() ).GetControl() & CONTROL_AI )
-         && ( _autoBattleColors & getEnemyForce( GetCurrentColor() ).GetColor() ) ) {
-        return true;
-    }
-
-    return false;
+    return ( GetCurrentForce().GetControl() & CONTROL_AI ) && !( getEnemyForce( GetCurrentColor() ).GetControl() & CONTROL_AI )
+           && ( _autoBattleColors & getEnemyForce( GetCurrentColor() ).GetColor() );
 }
 
 bool Battle::Arena::CanToggleAutoBattle() const
--- a/src/fheroes2/battle/battle_arena.h
+++ b/src/fheroes2/battle/battle_arena.h
@@ -98,6 +98,7 @@ namespace Battle
         bool BattleValid() const;
 
         bool AutoBattleInProgress() const;
+        bool EnemyOfAIHasAutoBattleInProgress() const;
         bool CanToggleAutoBattle() const;
 
         uint32_t GetTurnNumber() const
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -6318,15 +6318,26 @@ void Battle::Interface::CheckGlobalEvents( LocalEvent & le )
         }
     }
 
-    // Interrupting auto battle
-    if ( ( arena.AutoBattleInProgress() || arena.EnemyOfAIHasAutoBattleInProgress() )
-         && ( ( le.MouseClickLeft() || le.MouseClickRight()
-                || ( le.KeyPress()
-                     && ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_SWITCH ) || Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) ) )
-              && fheroes2::showMessage( fheroes2::Text( "", {} ),
-                                        fheroes2::Text( _( "Are you sure you want to interrupt the auto battle?" ), fheroes2::FontType::normalWhite() ),
-                                        Dialog::YES | Dialog::NO )
-                     == Dialog::YES ) ) {
+    // Check if auto battle interruption was requested.
+    InterruptAutoBattle( le );
+}
+
+void Battle::Interface::InterruptAutoBattle( LocalEvent & le )
+{
+    // Interrupt only if automation is currently on.
+    if ( !arena.AutoBattleInProgress() && !arena.EnemyOfAIHasAutoBattleInProgress() ) {
+        return;
+    }
+
+    if ( !le.MouseClickLeft() && !le.MouseClickRight() && !Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_SWITCH )
+         && !Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) {
+        return;
+    }
+
+    auto interrupt = fheroes2::showMessage( fheroes2::Text( "", {} ),
+                                            fheroes2::Text( _( "Are you sure you want to interrupt the auto battle?" ), fheroes2::FontType::normalWhite() ),
+                                            Dialog::YES | Dialog::NO );
+    if ( interrupt == Dialog::YES ) {
         // Identify which color requested the auto battle interrupt.
         auto color = arena.GetCurrentColor();
         if ( arena.GetCurrentForce().GetControl() & CONTROL_AI ) {
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -6320,7 +6320,7 @@ void Battle::Interface::CheckGlobalEvents( LocalEvent & le )
 
     // Interrupting auto battle
     if ( ( arena.AutoBattleInProgress() || arena.EnemyOfAIHasAutoBattleInProgress() )
-         && ( ( le.MouseClickLeft( btn_auto.area() )
+         && ( ( le.MouseClickLeft() || le.MouseClickRight()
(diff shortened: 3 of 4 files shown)


src/fheroes2/battle/battle_arena.cpp
```diff
@@ -1243,6 +1243,16 @@ bool Battle::Arena::AutoBattleInProgress() const
...
 }
 
+bool Battle::Arena::EnemyOfAIHasAutoBattleInProgress() const
+{
+    if ( _currentUnit == nullptr ) {
+        return false;
+    }
+
+    return ( GetCurrentForce().GetControl() & CONTROL_AI ) && !( getEnemyForce( GetCurrentColor() ).GetControl() & CONTROL_AI )
+           && ( _autoBattleColors & getEnemyForce( GetCurrentColor() ).GetColor() );
```
oleg-derevenetz, line 1253: I suggest to rewrite this according to the "early return" pattern, i.e. like this:

```cpp
if (!( GetCurrentForce().GetControl() & CONTROL_AI )) {
  return false;
}

const Force & enemyForce = getEnemyForce( GetCurrentColor() );

if ( enemyForce.GetControl() & CONTROL_AI ) {
  return false;
}

return (_autoBattleColors & enemyForce.GetColor());
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6319,16 +6319,24 @@ void Battle::Interface::CheckGlobalEvents( LocalEvent & le )
...
-                                    == Dialog::YES ) ) ) ) ) {
-        _interruptAutoBattleForColor = arena.GetCurrentColor();
+    if ( ( arena.AutoBattleInProgress() || arena.EnemyOfAIHasAutoBattleInProgress() )
+         && ( ( le.MouseClickLeft() || le.MouseClickRight()
+                || ( le.KeyPress()
+                     && ( Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_SWITCH ) || Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) ) )
+              && fheroes2::showMessage( fheroes2::Text( "", {} ),
+                                        fheroes2::Text( _( "Are you sure you want to interrupt the auto battle?" ), fheroes2::FontType::normalWhite() ),
+                                        Dialog::YES | Dialog::NO )
+                     == Dialog::YES ) ) {
```
oleg-derevenetz, line 6329: Since we're here, I suggest to rewrite this complicated-looking mess with lots of parentheses as well using the same "early return" pattern, perhaps using an auxiliary lambda.

src/fheroes2/battle/battle_arena.h
```diff
@@ -98,6 +98,7 @@ namespace Battle
         bool BattleValid() const;
 
         bool AutoBattleInProgress() const;
+        bool EnemyOfAIHasAutoBattleInProgress() const;
         bool CanToggleAutoBattle() const;
```
oleg-derevenetz, line 102: There is a need to investigate whether this method is needed now. Perhaps it should now be used only for assertions (or it should not be used at all).

oleg-derevenetz (changes requested):
Hi @dood-apo I just left a few suggestions, could you please take a look when you have time?

src/fheroes2/battle/battle_arena.h
dood-apo: I believe the new method is needed. The existing `AutoBattleInProgress` method checks the state only for the active side, which means it will always say `false` during AI's turn.

oleg-derevenetz: Please note that `AutoBattleInProgress()` as well as `CanToggleAutoBattle()` are used not just for battle UI. They are also used to decide what to do with "unlimited battles" (either turn the auto battle off or retreat) when AI is in charge, and their logic is exactly what is needed. What I meant was whether they are needed elsewhere in the combat UI code.

oleg-derevenetz: For instance, there is a check for `arena.CanToggleAutoBattle()` in the `Battle::Interface::EventAutoSwitch()`. Is this check really needed in this place, or can it be turned into an assertion? Is there a legal option when it might turn out to be `false` in this place?

dood-apo: > Please note that `AutoBattleInProgress()` as well as `CanToggleAutoBattle()` are used not just for combat UI. They are also used to decide what to do with "unlimited battles" (either turn the auto battle off or retreat) when AI is in charge, and their logic is exactly what is needed.

Right. This was exactly why I've created a new `EnemyOfAIHasAutoBattleInProgress` method for usage with the interruption code instead of modifying `AutoBattleInProgress` and `CanToggleAutoBattle`.

> For instance, there is a check for arena.CanToggleAutoBattle() in the Battle::Interface::EventAutoSwitch(). Is this check really needed in this place, or can it be turned into an assertion?

Haven't looked into it to be honest. Does it seem like it should be in the scope of this PR?

oleg-derevenetz: > Does it seem like it should be in the scope of this PR?

Well, technically it is not :) It just looks like this check is unnecessary from my point of view, because this event can be raised only during the human battle turn, and if we have a human battle turn, then it's not an AI turn, and there is a currently active unit, right? :) Anyway, this can be addressed later separately.

dood-apo: > It just looks like this check is unnecessary from my point of view, because this event can be raised only during the human battle turn, and if we have a human battle turn, then it's not an AI turn, and there is a currently active unit, right?

Correct. Seems like `EventAutoSwitch` can and should only be used during a human battle turn. Btw, at first we've had a disconnect, because the diff above your comments shows more than just `CanToggleAutoBattle` and I thought you were referring to `EnemyOfAIHasAutoBattleInProgress` instead.

dood-apo: > Anyway, this can be addressed later separately by replacing this check by an assertion.

Let's do it right away. Would you prefer to assert on `CanToggleAutoBattle` or completely remove that check from `EventAutoSwitch`?

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6318,17 +6318,36 @@ void Battle::Interface::CheckGlobalEvents( LocalEvent & le )
...
+    if ( !arena.AutoBattleInProgress() && !arena.EnemyOfAIHasAutoBattleInProgress() ) {
+        return;
+    }
+
+    if ( !le.MouseClickLeft() && !le.MouseClickRight() && !Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_AUTO_SWITCH )
+         && !Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) {
+        return;
+    }
+
+    auto interrupt = fheroes2::showMessage( fheroes2::Text( "", {} ),
```
oleg-derevenetz, line 6337: This variable can be `const`.

Regarding `auto`: in this particular project we prefer not to use `auto` everywhere. This can certainly be useful somewhere, for example, when using complex types like `std::map<...>::const_iterator` or lambda types (which are unique), when type is obvious from context (`const auto x = static_cast<X>( Y );`) or when iterating using the range-based for loop (speaking of std::map, people often confuse `for (const std::pair<const Key, Value> & item : map) {...}` with `for (const std::pair<Key, Value> & item : map) {...}`, so `auto` can be useful here as well), but in simple cases we prefer to specify the type explicitly, because it makes it easier to review without an IDE.

src/fheroes2/battle/battle_interface.h
```diff
@@ -386,6 +386,7 @@ namespace Battle
         void SwitchAllUnitsAnimation( const int32_t animationState ) const;
         void UpdateContourColor();
         void CheckGlobalEvents( LocalEvent & );
+        void InterruptAutoBattle( LocalEvent & );
```
oleg-derevenetz, line 389: This is an old-style declaration from the previous team. In the new code, we prefer to include parameter names into declarations, not just types. Also I suggest to name it somewhat less affirmative, say, `InterruptAutoBattleIfRequested` or something like this (maybe something better will come to mind), because currently it looks like it ALWAYS interrupts the auto battle.

oleg-derevenetz (changes requested):
Hi @dood-apo I just left a few more suggestions, please take a look when you have time.

src/fheroes2/battle/battle_arena.h
oleg-derevenetz: > Would you prefer to assert on `CanToggleAutoBattle` or completely remove that check from `EventAutoSwitch`?

When I'm in doubt, I usually turn the doubtful check into an assertion with comment like this: `TODO: remove this temporary assertion`.

ihhub (approved):

Branikolog:
Hello, @dood-apo 
Thanks for working on this useful improvement of the engine UI!
Now it's much easier to cancel auto battle. 👍 
I've noticed that interruption dialog can be initiated several times when a creature is moving:

https://github.com/ihhub/fheroes2/assets/55348946/0c52579a-cd92-4547-af93-6ec94c918a7a

I think it's not a proper behaviour, when we can confirm cancelling auto combat, then initiate dialog again, confirm one more time or cancel (but battle was already cancelled). Maybe it would be more correct not to show the dialog again, once player accepted cancellation.

Also, I really don't know, why do we have confirmation dialog on turning auto mode only for clicking on "auto" button in the left corner, but not for pressing the hotkey. I believe, player can accidentally press "A" button and initiate auto battle when he's not ready or was about to surrender.
image

Not sure it is in the scope of this PR, but  I believe it is kind of related. :)

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6333,16 +6334,21 @@ void Battle::Interface::InterruptAutoBattleIfRequested( LocalEvent & le )
...
+    int color = arena.GetCurrentColor();
+    if ( arena.GetCurrentForce().GetControl() & CONTROL_AI ) {
+        color = arena.GetOppositeColor( color );
+    }
+
+    // The battle interruption is already scheduled, no need for the dialog.
+    if (color == _interruptAutoBattleForColor) {
+        return;
+    }
+
```
oleg-derevenetz, line 6347: Hi @dood-apo I would also add something like the following code after this check, because, in theory, there can be either 0 or the current color, the other options are unacceptable, because they should have already been processed before switching to this unit. If this is not the case, then there is a mistake in the logic somewhere.

```suggestion

    assert( _interruptAutoBattleForColor == 0 );

```

dood-apo:
Hi @Branikolog ,

> I think it's not a proper behaviour, when we can confirm cancelling auto combat, then initiate dialog again, confirm one more time or cancel (but battle was already cancelled). Maybe it would be more correct not to show the dialog again, once player accepted cancellation.

Fixed.

> Also, I really don't know, why do we have confirmation dialog on turning auto mode only for clicking on "auto" button in the left corner, but not for pressing the hotkey.

Addressed that one also. Renamed the `EventAutoSwitch` method to `EventStartAutoBattle` while I was there, since the interruption code is actually separate from the auto battle start code and `EventAutoSwitch` could not be reached in switching off scenario.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6318,17 +6318,44 @@ void Battle::Interface::CheckGlobalEvents( LocalEvent & le )
...
+    // Right now there should be no pending auto battle interruptions.
+    assert( _interruptAutoBattleForColor == 0 );
+
+    const int interrupt = fheroes2::showMessage( fheroes2::Text( "", {} ),
+                                                 fheroes2::Text( _( "Are you sure you want to interrupt the auto battle?" ), fheroes2::FontType::normalWhite() ),
+                                                 Dialog::YES | Dialog::NO );
+    if ( interrupt == Dialog::YES ) {
+        // Interrupt, but only if the color identified belongs to a human player.
+        if ( !( arena.getForce( color ).GetControl() & CONTROL_AI ) ) {
```
oleg-derevenetz, line 6357: Hi @dood-apo I have an impression that this check could be turned into an assertion and moved somewhere after we figured out the color (at line ~6342), because, if that wasn't the case, then we shouldn't be here, because we would not have passed the check at the very beginning of this function at line 6328, right? What do you think?

dood-apo: Spot on.

But this time maybe it is fine to simply remove this check without converting to an assertion. The first check `if ( !arena.AutoBattleInProgress() && !arena.EnemyOfAIHasAutoBattleInProgress() ) { return; }` makes sure that there is at least one human player at the battlefield, while `if ( arena.GetCurrentForce().GetControl() & CONTROL_AI ) { color = arena.GetOppositeColor( color ); }` does not allow AI to be the selected color.

oleg-derevenetz: > But this time maybe it is fine to simply remove this check without converting to an assertion.

Yes, sure. BTW, when `AUTO_SWITCH` command is processed, there are additional sanity checks in this regard:

https://github.com/ihhub/fheroes2/blob/3e093673745924c292574364da7d1f8ff8e7a875/src/fheroes2/battle/battle_action.cpp#L1318-L1344

so it will be checked anyway.

Branikolog (approved):
@dood-apo , everything works fine.
Thanks for your contribution to the project!

ihhub:
We will merge this pull request after the release which is tomorrow. This is to minimize introduction of issues just before the release.

ihhub:
@dood-apo , thank you so much for this improvement!

PR #8360, opened 2024-01-31
Add a method to not align all text uniformly

### LINKED ISSUE #8295: Add text function that places as many words on the first line as possible before breaking to the next
### Preliminary checks

This is to avoid cases like these for "Their cowardice ..." where more words could have been put on the first line of that text:

image

Fit as many words as possible in the width of the maxWidth specified.

### Additional info

Me and @Districh-ru were also talking about having a text function that can do other alignments than center, but I'm not sure where exactly we would need it currently?

ihhub:
close #8295


ihhub:
Hi @zenseii , can you please test the newly added function and if it works use it in the code by adding commits in this branch?

Districh-ru (approved):

ihhub:

PR #8362, opened 2024-02-02
Fix the Nintendo Switch build

oleg-derevenetz:
For some reason, in the latest `devkita64` Docker image `sdl2-config --libs` uses not the "canonical" way of linking the library (`-lSDL2`), but the "full path" way (`/opt/devkitpro/portlibs/switch/lib/libSDL2.a`), what our `SDL2_mixer` linkage heuristics was not designed for, and the linkage was made in the wrong sequence.

Before:

https://github.com/ihhub/fheroes2/actions/runs/7757283009/job/21156607398#step:5:282

Now:

https://github.com/ihhub/fheroes2/actions/runs/7760886470/job/21168314572#step:5:282

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -142,9 +142,7 @@ endif
 
 # SDL_LIBS can already be defined in a platform-specific Makefile
 ifeq ($(origin SDL_LIBS),undefined)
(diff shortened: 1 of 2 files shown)


ihhub (approved):

ihhub:

PR #8364, opened 2024-02-03
Battle::Arena: remove basic private method, replace condition by assertion

oleg-derevenetz:


src/fheroes2/battle/battle_arena.cpp
```diff
@@ -431,7 +431,9 @@ void Battle::Arena::UnitTurn( const Units & orderHistory )
                 AI::Get().BattleTurn( *this, *_currentUnit, actions );
             }
             else {
-                HumanTurn( *_currentUnit, actions );
+                assert( _interface != nullptr );
```
oleg-derevenetz, line 434: If we are here, then there must be the battle interface, otherwise, there is some kind of logic error.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for keeping the code clean!

PR #8366, opened 2024-02-04
Remove MinGW support from Makefiles, reformat YAML files uniformly

oleg-derevenetz:
MinGW platform for fheroes2 is currently not officially supported, the build for it doesn't actually work, there is no CI for it and and there are no people willing to fix/support it (and it seems that there is no target audience).


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for keeping the code clean!

PR #8385, opened 2024-02-09
Do not do display zeroing twice during application startup

ihhub:
Faster loading of the engine.

Plus small speed up for bin info file processing.

### DIFF
--- a/src/fheroes2/game/fheroes2.cpp
+++ b/src/fheroes2/game/fheroes2.cpp
@@ -155,24 +155,20 @@ namespace
             const Settings & conf = Settings::Get();
 
             fheroes2::Display & display = fheroes2::Display::instance();
+            fheroes2::ResolutionInfo bestResolution{ conf.currentResolutionInfo() };
+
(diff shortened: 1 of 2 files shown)


src/fheroes2/game/fheroes2.cpp
```diff
@@ -155,24 +155,20 @@ namespace
...
                     }
                 }
-
-                display.setResolution( bestResolution );
-            }
-            else {
-                display.setResolution( conf.currentResolutionInfo() );
             }
 
-            display.fill( 0 ); // start from a black screen
```
oleg-derevenetz, line 175: Hi @ihhub could you please point out where is proper zeroing performed now?

ihhub: Hi @oleg-derevenetz , `void Display::setResolution( ResolutionInfo info )` method calls **Image::reset();** function that does zeroing of the image.

oleg-derevenetz (approved):

PR #8386, opened 2024-02-09
Fix a crash while placing random castles in the Editor

ihhub:
relates to #6845

### DIFF
--- a/src/fheroes2/maps/map_format_helper.cpp
+++ b/src/fheroes2/maps/map_format_helper.cpp
@@ -227,7 +227,7 @@ namespace Maps
         // Towns can be neutral so 1 more color for them.
         std::array<bool, mainColors + 1> townColorsPresent{ false };
         std::array<uint8_t, mainColors> heroRacesPresent{ 0 };
-        std::array<uint8_t, mainColors> townRacesPresent{ 0 };
+        std::array<uint8_t, mainColors + 1> townRacesPresent{ 0 };
 
         const auto & heroObjects = getObjectsByGroup( ObjectGroup::KINGDOM_HEROES );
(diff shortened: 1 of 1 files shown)


src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -227,7 +227,7 @@ namespace Maps
         // Towns can be neutral so 1 more color for them.
         std::array<bool, mainColors + 1> townColorsPresent{ false };
         std::array<uint8_t, mainColors> heroRacesPresent{ 0 };
-        std::array<uint8_t, mainColors> townRacesPresent{ 0 };
+        std::array<uint8_t, mainColors + 1> townRacesPresent{ 0 };
```
Districh-ru, line 230: Hi, @ihhub , should we do the same for random race heroes (`heroRacesPresent`)?

ihhub: Hi @Districh-ru , we don't have neutral heroes :) So there is no need of doing this.

Districh-ru: Sure, heroes can not be neutral. But Im talking about random ones. :)
изображение

ihhub: Random heroes have color. The crash was for neutral castles that have no color.

Districh-ru: Ow, sorry. :(
I was a little confused: the PR name has "Random", the variable name has "Races" and nothing about color. And the number of main colors and races are the same in the game and can be used for the length of array for races and for colors...
So I treated the `townRacesPresent` as count of towns for race, but should as `townRacesPresentForColor`.

Districh-ru (approved):

PR #8392, opened 2024-02-10
Adjust the difficulties of campaign scenarios that were hard-coded in the original game

### LINKED ISSUE #8351: Wrong difficulty/starting resources for many Campaign Scenarios
### Preliminary checks

### Platform

Windows

### Describe the bug

I've checked the starting resources of PoL scenarios in the OG up against fheroes2, and we can see that fheroes2 gives the wrong resources presumably due to setting a different difficulty for these missions than the OG does.

Here are my findings, reported as: (current fheroes2 resource level) -> (correct resource level) :

<details>

Starting Resource difficulty level that needs to be set
Roland:
Rol 1 checked
Rol 2 Normal -> Easy
Rol 3 checked
Rol 4 checked
Rol 5 checked
Rol 6 checked
Rol 7 checked
Rol 8 checked
Rol 9 checked
Rol 10 checked

Archibald:
Arch 1 checked
Arch 2 Normal -> Easy
Arch 3 checked
Arch 4 checked
Arch 5 checked
Arch 6 checked
Arch 7 checked
Arch 8 checked
Arch 9 checked
Arch 10 checked
Arch 11 checked

Descendants:
Des 1 Normal -> Easy
Des 2 checked
Des 3 checked
Des 4 checked
Des 5 checked
Des 6 Expert -> Hard
Des 7 checked
Des 8 checked

Wizard's Isle:
Wiz 1 checked
Wiz 2 checked
Wiz 3 checked
Wiz 4 Normal -> Hard

Voyage Home:

Voy 1 Normal -> Easy
Voy 2 checked
Voy 3 checked
Voy 4 checked

Price of Loyalty:

Pol 1 Normal -> Easy
Pol 2 checked fine
Pol 3 checked fine
Pol 4 checked fine
Pol 5 checked fine
Pol 6 Normal -> Hard
Pol 7 Normal -> Hard
Pol 8 Normal -> Expert

</details>

Looking at the PoL scenarios in the OG editor I can see that they are set as Normal difficulty, so presumably the OG did some hardcoded changes to these maps.

The question is whether we should hardcode these maps to change their difficulties to match the OG, or if we should set the specific resources. I haven't checked if there are other hints that could tell us whether those maps are supposed to have an AI from those difficulties, and honestly I'm not sure we need to check that since our AI is different anyways and we can adjust it the way we want.

If we take the 2nd Roland and Archibald maps as examples, then if we decide to lower its difficulty to the OG's Easy, then that would mean that for fheroes2 campaign's Normal difficulty, this map would still be Easy difficulty since the inherent difficulty is kept, meanwhile for fheroes2 Easy campaign difficulty, it would still be Easy, and there wouldn't really be a difference from Normal campaign difficulty.

My suggestion for this case is to force Normal to never be lower than Normal. One alternative is that we can just have both Normal and Easy be the same there. Another alternative is to add even more resources to Easy difficulty, more than what you get for Easy difficulty missions, sort of like a Very Easy difficulty.

### Save file

Save games for fheroes2
[PoL Complete.zip](https://github.com/ihhub/fheroes2/files/14077256/PoL.Complete.zip)

For the OG just use the code "911" to automatically win a scenario.

Make sure not to choose any starting resource bonuses.

### Additional info

We have to check the other campaigns too.

oleg-derevenetz:
fix #8351

This PR also fixes a number of issues where the difficulty setting for the campaign was not taken into account. For example, in the `master` branch:

https://github.com/ihhub/fheroes2/assets/32623900/d6c2928f-17f8-4629-a79a-669f131a9e94

With this PR:

https://github.com/ihhub/fheroes2/assets/32623900/3ef52347-b339-4a2f-a01a-a180ee982cce

### DIFF
--- a/src/fheroes2/campaign/campaign_savedata.cpp
+++ b/src/fheroes2/campaign/campaign_savedata.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2021 - 2023                                             *
+ *   Copyright (C) 2021 - 2024                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
--- a/src/fheroes2/campaign/campaign_savedata.cpp
+++ b/src/fheroes2/campaign/campaign_savedata.cpp
@@ -27,6 +27,7 @@
 
 #include "army.h"
 #include "campaign_data.h"
+#include "difficulty.h"
 #include "serialize.h"
 
 namespace Campaign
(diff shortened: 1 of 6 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these useful changes!

PR #8395, opened 2024-02-11
Update editor's erasure tool

Districh-ru:
In this PR the Erasure tool is updated to allow erase of all editor supported objects. This update contains:
- new erase groups;
- support for undo/redo of erase operation.

https://github.com/ihhub/fheroes2/assets/113276641/12755f83-7cd8-474a-aed6-ca60208e8ba1


src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -3283,4 +3141,47 @@ namespace Maps
 
         return true;
     }
+    std::set<uint32_t> getObjectUidsInArea( const int32_t startTileId, const int32_t endTileId )
```
ihhub, line 3144: Please add an empty line before to separate 2 functions.

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -3283,4 +3141,47 @@ namespace Maps
...
+    {
+        const int32_t mapWidth = world.w();
+        const int32_t maxTileId = mapWidth * world.h() - 1;
+        if ( startTileId < 0 || startTileId > maxTileId || endTileId < 0 || endTileId > maxTileId ) {
+            return {};
```
ihhub, line 3149: Please add an assertion here as this is not an expected case.

ihhub (changes requested):

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -167,6 +169,94 @@ namespace
     {
         return isConditionValid( Maps::getGroundLevelUsedTileOffset( info ), mainTilePos, condition );
     }
+
+    bool removeObjects( Maps::Map_Format::MapFormat & mapFormat, fheroes2::Point tileAreaIndices, std::vector<Maps::ObjectGroup> objectGroups )
```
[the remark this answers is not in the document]
Districh-ru: Hi, @ihhub, thanks for your review and valuable suggestions!
I missed the `const &` here. :(  Done!

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -167,6 +169,94 @@ namespace
...
+        std::set<uint32_t> objectsUids = Maps::getObjectUidsInArea( tileAreaIndices.x, tileAreaIndices.y );
+
+        if ( objectsUids.empty() ) {
+            return false;
+        }
+
+        bool needRedraw = false;
+
+        // Filter objects by group and remove them from '_mapFormat'.
+        for ( std::vector<Maps::Map_Format::TileInfo>::iterator mapTile = mapFormat.tiles.begin(); mapTile != mapFormat.tiles.end(); ++mapTile ) {
```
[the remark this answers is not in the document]
Districh-ru: OK, lets not use iterators here. It will also simplify the streams/roads erase code.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -284,6 +284,59 @@ namespace Interface
         return {};
     }
 
+    std::vector<Maps::ObjectGroup> EditorPanel::getEraseObjectGroups() const
```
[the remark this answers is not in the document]
Districh-ru: Sure!

src/fheroes2/maps/maps_tiles_helper.cpp
Districh-ru: 👍

src/fheroes2/maps/maps_tiles_helper.cpp
Districh-ru: Done.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -167,6 +168,95 @@ namespace
...
+        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size(); ++mapTileIndex ) {
+            Maps::Map_Format::TileInfo & mapTile = mapFormat.tiles[mapTileIndex];
+
+            if ( mapTile.objects.empty() ) {
+                continue;
+            }
+
+            for ( auto object = mapTile.objects.begin(); object != mapTile.objects.end(); ) {
+                if ( objectsUids.find( object->id ) == objectsUids.end()
+                     || std::none_of( objectGroups.begin(), objectGroups.end(), [&object]( const Maps::ObjectGroup group ) { return group == object->group; } ) ) {
```
ihhub, line 193: Can you please explain when we could have the same object UID but different object group? Is it only for towns? If yes, then what happen if we remove this condition, erase object UID at the end of this loop and treat towns as normal objects?

Moreover, `updateRoadOnTile` and `updateStreamOnTile` calls do not modify mapFormat but these calls keep existing objects on Adventure Map.

ihhub (commented):
Hi @Districh-ru , I left one comment regarding the logic. Could you please take a look at it?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Hi @ihhub. We get object's UIDs from `world` and the main tile of some objects may be outside of the selected area to remove. So we go though all Map_Format tiles. (In future we should optimize it and limit the range of search).
These tow checks are made:
- first - to check if the object in the current `mapTile` is the objects that was selected by user;
- second - if object is in the group of objects that can be erased (user can unselect some groups).
There can be a case when there is more than one object on the `mapTile`. And different objects with one UID is only Town/Castle. And we treat them as normal objects and search only for the main group ( we don't search for flags and basements).

The found UID is removed from `objectsUids`. But if we currently remove any found UID without checking for group then we will not erase the Town: the Town Flag is located prior to the Town. ... We can improve this logic a bit here (erase town by finding its first flag) ... it'll need some tests.

For Roads and Streams we should remove them by `updateRoadOnTile` and `updateStreamOnTile` to properly update Roads and Streams on tiles around in `world`. These objects are updated in Map_Format by `action.commit();` using `writeTile()` function.

Districh-ru: The code is updated: now we remove any found UID except flags and basements because they are the part of town.

ihhub (approved):

ihhub:
Hi @Branikolog , please test this PR as the highest priority.

Branikolog:
Hi, @Districh-ru !
Nice work, as always!

Found only one tiny issue here:

https://github.com/ihhub/fheroes2/assets/55348946/0119a592-28a8-458f-ab97-0232586aa29e

Road under the castle is not erased somehow.

Districh-ru:
Hi, @Branikolog, the issue is fixed.
In future we will rework roads placing for the new fheroes2 map format. It should fix some road placing issues including this one without a need of placing extra "hacky" checks to the code.

Branikolog:
Hi, @Districh-ru 
Assertion is occurring, when I release erasure tool (area brush) over the edge of the world map.

Districh-ru:
> Hi, @Districh-ru Assertion is occurring, when I release erasure tool (area brush) over the edge of the world map.

@Branikolog, thanks for finding this issue!
Fixed.

ihhub:
@Districh-ru , thanks a lot for these changes!

PR #8403, opened 2024-02-13
Add info about ranged penalty elimination to archery skill description

### LINKED ISSUE #8398: Archery skill, missing description for shooting without obstacle penalty
### Preliminary checks

The Archery skill allows to shoot past castle walls. But the description does not describe this feature.
image

I suggest adding a description about shooting without obstacle penalty. Players should know about this useful feature.

### Additional info

_No response_

zenseii:
Add a description about ranged penalty being when having the archery skill.
Add ranged penalty Get function to `GameStatic` and use in relevant places in the code.

Fix #8398 

image

### DIFF
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -61,6 +61,7 @@
 #include "monster_info.h"
 #include "resource.h"
 #include "settings.h"
+#include "skill.h"
 #include "speed.h"
 #include "spell.h"
 
--- a/src/fheroes2/ai/normal/ai_normal_battle.cpp
+++ b/src/fheroes2/ai/normal/ai_normal_battle.cpp
@@ -933,8 +933,20 @@ namespace AI
 
         // Mark as castle siege only if any tower is present. If no towers present then nothing to defend and most likely all walls are destroyed as well.
         if ( castle && Arena::isAnyTowerPresent() ) {
-            const bool attackerIgnoresCover
-                = arena.GetForce1().GetCommander()->GetBagArtifacts().isArtifactBonusPresent( fheroes2::ArtifactBonusType::NO_SHOOTING_PENALTY );
+            const bool attackerIgnoresCover = [&arena]() {
+                const HeroBase * commander = arena.GetForce1().GetCommander();
+                assert( commander != nullptr );
+
+                if ( commander->GetBagArtifacts().isArtifactBonusPresent( fheroes2::ArtifactBonusType::NO_SHOOTING_PENALTY ) ) {
+                    return true;
+                }
+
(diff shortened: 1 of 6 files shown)


src/fheroes2/game/game_static.h
```diff
@@ -63,6 +63,7 @@ namespace GameStatic
     int32_t ObjectVisitedModifiers( const MP2::MapObjectType objectType );
 
     int GetBattleMoatReduceDefense();
+    uint32_t getCastleWallRangedPenalty();
```
oleg-derevenetz, line 66: Hi @zenseii I usually add a comment to new functions or a method with a brief (if possible) description of what this function does. For example:

```cpp
// Returns the approximate distance between two tiles with given indexes. This distance is calculated as the number of
// tiles (truncated to the nearest smaller integer value) that would need to be traversed in a straight direction to
// overcome the distance that corresponds to the distance that the hero would have to cover when moving between tiles
// with given indexes using the usual rules and penalties.
//
// For example, in order to navigate to a tile B, which is located seven tiles below and five to the right of tile A,
// the hero should move two tiles down and five tiles right down diagonally (diagonal movement costs 50% more), which
// corresponds to 2 + 5 * 1.5 = 9.5 (9 after truncation) tiles traversed in the straight direction.
//
// This function should be avoided unless high precision is not important.
uint32_t GetApproximateDistance( const int32_t pos1, const int32_t pos2 );
```

The advantage is that many IDEs understand such comments and show tooltips when hovering over this function. For example, this is how it looks in MS Visual Studio:

tooltip

The exception is if it is absolutely obvious from the name of the function what it does.

src/fheroes2/ai/normal/ai_normal_battle.cpp
```diff
@@ -951,15 +965,15 @@ namespace AI
                 _myShootersStrength += towerStr;
 
                 if ( !attackerIgnoresCover ) {
-                    _enemyShootersStrength /= 1.5;
+                    _enemyShootersStrength /= 1 + ( GameStatic::getCastleWallRangedPenalty() / 100.0 );
```
oleg-derevenetz, line 968: These `/ 1.5` here and below were introduced in #6743 by replacing `/ 2`. Perhaps the value 1.5 is not directly related to this penalty, but, on the other hand, if the penalty is reduced this coefficient should also be reduced (and if the penalty is increased, it should also be increased). OK, let's do it this way.

oleg-derevenetz (changes requested):
Hi @zenseii I just left one question and one reasoning. Could you please take a look when you have time?

Branikolog (approved):

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@zenseii , huge thanks for this change!

PR #8407, opened 2024-02-15
Add initial iOS Support

dbratcher:
Sets up an iOS project and experiments with getting resources to load in the iOS system


ihhub:
I've tested these changes. Without file I/O operation functions the engine doesn't run. I think we should do the changes in steps:
- add file I/O operations support for iOS
- add compilable version as this PR
- add a way to avoid the need of game resources while building


What do you think @oleg-derevenetz ?

oleg-derevenetz:
Hi @ihhub I don't have an iOS device (I usually use Android), but are there issues with I/O on iOS? I always thought that POSIX I/O API does work on iOS just fine. And `std::filesystem` should also work since it's cross-platform by design.

ihhub:
> Hi @ihhub I don't have an iOS device (I usually use Android), but are there issues with I/O on iOS? I always thought that POSIX I/O API does work on iOS just fine. And `std::filesystem` should also work since it's cross-platform by design.

I am able to compile the code on macOS for iOS without any changes in the source code from this PR but the game stops immediately with no data warning. I will check on my machine what exactly the case.

oleg-derevenetz:
> I am able to compile the code on macOS for iOS without any changes in the source code from this PR but the game stops immediately with no data warning. I will check on my machine what exactly the case.

I suppose we may need to tune some logic related to h2d file bundling and/or path discovery maybe, but I believe that POSIX I/O should work in general on iOS.

ihhub:
Hi @oleg-derevenetz ,

I've made many changes to bring this pull request to a suitable level. I confirm that I am able to run the engine on a real iPhone device. This is not the best solution so far and many things are missing but at least it's a start.

ihhub (approved):

ihhub:
> Hi @ihhub looks good to me. It would be nice to add a short instruction on how to build fheroes2 for iOS - how and where to install dependencies and so on. Even better would be to add workflow for automatic builds :) But this can be done later in a separate PR.

Absolutely. I will add instructions and proceed with the rest of changes. I wanted to make the base changes otherwise we would have stuck for a long time.

ihhub:
@dbratcher , thank you very much for your contribution to the project!

PR #8408, opened 2024-02-15
Convenient list scrolling for touch screens

### LINKED ISSUE #2420: Expand scrolling area with scroll wheel of Save/Load list
Currently, we can scroll only if cursor is over list itself or over a scrollbar.
image
image
Over the other areas scrolling using mouse scroll wheel doesn't work. It would be nice if we can use scroll wheel over any place here and savelist during game.

dood-apo:
On the smaller touch screens (mobile phones) it is quite difficult to navigate scenario list dialogs or load game dialogs if there are a lot of list items in them. Using a thin scrollbar or small arrows is the only possible way. Hero/castle lists are even harder to operate. Having more than 4 heroes/castles requires the user to fight with even smaller scrollbars.

This PR addresses this inconvenience, as presented on a video below.

https://github.com/ihhub/fheroes2/assets/84446398/da0b5d9c-1264-4230-b4a4-6d53bcf00fbc

close #2420

### DIFF
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -667,6 +667,7 @@ LocalEvent::LocalEvent()
     : modes( 0 )
     , key_value( fheroes2::Key::NONE )
     , mouse_button( 0 )
+    , _isDragInProgress( false )
     , _mouseButtonLongPressDelay( mouseButtonLongPressTimeout )
 {}
 
@@ -705,6 +706,9 @@ void LocalEvent::OpenTouchpad()
 LocalEvent & LocalEvent::Get()
 {
     static LocalEvent le;
+    if ( !le.MousePressLeft() ) {
+        le._isDragInProgress = false;
+    }
 
     return le;
 }
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -706,9 +706,6 @@ void LocalEvent::OpenTouchpad()
 LocalEvent & LocalEvent::Get()
 {
     static LocalEvent le;
-    if ( !le.MousePressLeft() ) {
-        le._isDragInProgress = false;
-    }
 
     return le;
 }
@@ -991,6 +988,7 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
 
                 ResetModes( MOUSE_PRESSED );
                 SetModes( MOUSE_RELEASED );
+                _isDragInProgress = false;
             }
 
             mouse_button = SDL_BUTTON_LEFT;
@@ -1016,6 +1014,7 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
             ResetModes( MOUSE_PRESSED );
             SetModes( MOUSE_RELEASED );
             SetModes( MOUSE_TOUCH );
+            _isDragInProgress = false;
         }
 
         mouse_button = SDL_BUTTON_RIGHT;
@@ -1085,6 +1084,7 @@ void LocalEvent::HandleControllerButtonEvent( const SDL_ControllerButtonEvent & 
         else {
             ResetModes( MOUSE_PRESSED );
             SetModes( MOUSE_RELEASED );
+            _isDragInProgress = false;
         }
 
         if ( button.button == SDL_CONTROLLER_BUTTON_A ) {
@@ -1314,6 +1314,7 @@ void LocalEvent::HandleMouseButtonEvent( const SDL_MouseButtonEvent & button )
     else {
         ResetModes( MOUSE_PRESSED );
         SetModes( MOUSE_RELEASED );
+        _isDragInProgress = false;
     }
 
     mouse_button = button.button;
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -667,7 +667,6 @@ LocalEvent::LocalEvent()
     : modes( 0 )
     , key_value( fheroes2::Key::NONE )
     , mouse_button( 0 )
-    , _isDragInProgress( false )
     , _mouseButtonLongPressDelay( mouseButtonLongPressTimeout )
 {}
 
@@ -988,7 +987,7 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
 
                 ResetModes( MOUSE_PRESSED );
                 SetModes( MOUSE_RELEASED );
-                _isDragInProgress = false;
+                ResetModes( DRAG_ONGOING );
             }
 
             mouse_button = SDL_BUTTON_LEFT;
@@ -1014,7 +1013,7 @@ void LocalEvent::HandleTouchEvent( const SDL_TouchFingerEvent & event )
             ResetModes( MOUSE_PRESSED );
             SetModes( MOUSE_RELEASED );
             SetModes( MOUSE_TOUCH );
-            _isDragInProgress = false;
+            ResetModes( DRAG_ONGOING );
         }
 
         mouse_button = SDL_BUTTON_RIGHT;
@@ -1084,7 +1083,7 @@ void LocalEvent::HandleControllerButtonEvent( const SDL_ControllerButtonEvent & 
         else {
             ResetModes( MOUSE_PRESSED );
             SetModes( MOUSE_RELEASED );
-            _isDragInProgress = false;
+            ResetModes( DRAG_ONGOING );
         }
 
         if ( button.button == SDL_CONTROLLER_BUTTON_A ) {
@@ -1314,7 +1313,7 @@ void LocalEvent::HandleMouseButtonEvent( const SDL_MouseButtonEvent & button )
     else {
         ResetModes( MOUSE_PRESSED );
         SetModes( MOUSE_RELEASED );
-        _isDragInProgress = false;
+        ResetModes( DRAG_ONGOING );
     }
 
     mouse_button = button.button;
--- a/src/engine/localevent.h
+++ b/src/engine/localevent.h
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2023                                             *
+ *   Copyright (C) 2019 - 2024                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2008 by Josh Matthews <josh@joshmatthews.net>           *
@@ -306,6 +306,16 @@ public:
         }
     }
 
+    bool isDragInProgress() const
+    {
+        return _isDragInProgress;
+    }
+
+    void setDragStatus( const bool inProgress )
+    {
+        _isDragInProgress = inProgress;
+    }
+
 private:
     LocalEvent();
 
@@ -364,6 +374,7 @@ private:
     uint32_t modes;
     fheroes2::Key key_value;
     int mouse_button;
+    bool _isDragInProgress;
 
     fheroes2::Point mouse_pl; // press left
     fheroes2::Point mouse_pm; // press middle
--- a/src/engine/localevent.h
+++ b/src/engine/localevent.h
@@ -353,7 +353,7 @@ private:
         MOUSE_TOUCH = 0x0020,
         // Key on the keyboard is currently being held down
         KEY_HOLD = 0x0040,
-        // Key on the keyboard is currently being held down
+        // Some UI component registered the start of drag motion
         DRAG_ONGOING = 0x0080,
     };
 
--- a/src/engine/localevent.h
+++ b/src/engine/localevent.h
@@ -308,12 +308,12 @@ public:
 
     bool isDragInProgress() const
     {
-        return _isDragInProgress;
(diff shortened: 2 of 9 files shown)


ihhub:
Hi @dood-apo , I linked #2420 issue to this pull request.

Branikolog:
Hi, @dood-apo !
I've made a several tests and experienced a few issues. Could you, please, look, if they can be fixed?

First of all, while scrolling hero and castle lists we shouldn't be able to interact and scroll other UI elements, like minimap or adventure map:

https://github.com/ihhub/fheroes2/assets/55348946/2be98229-beb0-491f-93c3-723d11e47185

The related problem occurs during battle, when I accidentally release mouse button while scrolling out of the list:

https://github.com/ihhub/fheroes2/assets/55348946/3d576da8-9794-41f8-b3e9-7f3a6259f29d

Secondly, hero and castle lists should be scrolled further, when we exceed the UI box containing them. There's not really much space to make short swipes exactly in the area with castles/heroes on small android screens.
I believe it would be more convenient, if we touch the list and can swipe much/higher/lower without taking away finger to reach upper or bottom items.

Also, when we scroll and move cursor out of the lists, hero/castle selection doesn't work for the first click. We need to do second click to select hero/castle.

https://github.com/ihhub/fheroes2/assets/55348946/15a7bccf-9362-4c0b-9aec-ca99d08123ed

On my Android device I sometimes experience a list jump, when starting to scroll. But I couldn't reproduce that on PC version.

https://github.com/ihhub/fheroes2/assets/55348946/ed934dcc-5d24-4312-bf16-3fd74021ebe0

While scrolling the save file list, which contains multiple small elements I feel no inconvenience, as scrolling seems to be applied when mouse reaches the next list element (which is just a text string and is quite small in this case). But using the implemented scrolling for lists with larger elements (hero/castle lists on the adventure map and especially kingdom overview list) I feel that I need to make too many swipes. Is it possible to make list scroll faster when I swipe over it? I mean I need to pass the distance equal to the height of the list element to make list going up/down for just one step. I'm not sure, we need to be able to scroll the entire list with just a single move within the list itself (like current scroll bars are working currently), But I definitely expect more list elements passed when I make a single swipe.


https://github.com/ihhub/fheroes2/assets/55348946/68d2e286-3a7a-4f26-a27b-dc2c3b4e9c90

dood-apo:
Hi @Branikolog ,

As always, thanks for such a thorough testing!

> First of all, while scrolling hero and castle lists we shouldn't be able to interact and scroll other UI elements, like minimap or adventure map

Yes, I was aware of this and it should be fixed. However this issue is already present on master, while dragging the srollbar itself. I believe it is out of scope of this PR, since it would require a lot of tinkering with how the events are being handled by the other UI components and it is not a regression. Maybe creating a separate issue to track it might possibly work?

> The related problem occurs during battle, when I accidentally release mouse button while scrolling out of the list

Same as above and still reproducible on master with the scrollbar dragging.

> Secondly, hero and castle lists should be scrolled further, when we exceed the UI box containing them. There's not really much space to make short swipes exactly in the area with castles/heroes on small android screens.
I believe it would be more convenient, if we touch the list and can swipe much/higher/lower without taking away finger to reach upper or bottom items.

Correct. Same reason as points above, and still reproducible on master with the scrollbar dragging.

> Also, when we scroll and move cursor out of the lists, hero/castle selection doesn't work for the first click. We need to do second click to select hero/castle.

That was a bug in my implementation. Fixed.

> On my Android device I sometimes experience a list jump, when starting to scroll. But I couldn't reproduce that on PC version.

This is a mystery to me to be honest. Changes should be completely portable. I'll try to repro on my phone, when I have some more time.

> While scrolling the save file list, which contains multiple small elements I feel no inconvenience, as scrolling seems to be applied when mouse reaches the next list element (which is just a text string and is quite small in this case). But using the implemented scrolling for lists with larger elements (hero/castle lists on the adventure map and especially kingdom overview list) I feel that I need to make too many swipes. Is it possible to make list scroll faster when I swipe over it? I mean I need to pass the distance equal to the height of the list element to make list going up/down for just one step. I'm not sure, we need to be able to scroll the entire list with just a single move within the list itself (like current scroll bars are working currently), But I definitely expect more list elements passed when I make a single swipe.

In my opinion when you are dragging the contents of a list to scroll it, it has to be somewhat in sync with your finger. Scrolling faster than you're dragging will feel really weird. For example if you open MS Word and set font to gigantic, your finger drags will also not scroll that much, right? And I think it's a good thing, because otherwise there would be this effect of things escaping from your finger. The concern of too many swipes needed for castles/heroes lists on the main UI would be addressed in the future by fixing the issue with other UI elements stealing mouse events from the lists (your first three observations - if we agree that it can be tracked as a separate issue), because if you're allowed to drag past the list area and still drag the content, one longer swipe will suffice.

dood-apo:
> > On my Android device I sometimes experience a list jump, when starting to scroll. But I couldn't reproduce that on PC version.
> 
> This is a mystery to me to be honest. Changes should be completely portable. I'll try to repro on my phone, when I have some more time.

Confirmed, on Android if you drag-scroll but finish that drag outside of the list area (eg. on the button panel), the subsequent touch on the list results in moving the list to a different index. Root-caused also - turns out that drag status reset is done when not-pressed-mouse enters list area, but when you enter it with the finger flying above the screen.... well you get the point ;). I'll try to think of a better way to reset drag state.

BTW also 100% reproducible on PC, you just need a touch screen ;).

oleg-derevenetz:
I am not very familiar with this part of the code, but it seems to me that the problem with random interactions with neighboring elements and the problem with jumps can be solved by organizing an internal event loop, just like it's done for example in `Interface::BorderWindow`. However, animations on the map, etc are likely to freeze while dragging in this case.

Branikolog:
Hi, @dood-apo 
Sorry for the late response.

>In my opinion when you are dragging the contents of a list to scroll it, it has to be somewhat in sync with your finger. Scrolling faster than you're dragging will feel really weird. For example if you open MS Word and set font to gigantic, your finger drags will also not scroll that much, right? And I think it's a good thing, because otherwise there would be this effect of things escaping from your finger. The concern of too many swipes needed for castles/heroes lists on the main UI would be addressed in the future by fixing the issue with other UI elements stealing mouse events from the lists (your first three observations - if we agree that it can be tracked as a separate issue), because if you're allowed to drag past the list area and still drag the content, one longer swipe will suffice.

Well, I just shared my opinion. Probably your solution is the way to go, let's see someone else's feedback on this implementation. 👍 

>Yes, I was aware of this and it should be fixed. However this issue is already present on master, while dragging the srollbar itself. I believe it is out of scope of this PR, since it would require a lot of tinkering with how the events are being handled by the other UI components and it is not a regression. 

Sure! With the issue using heroes/castles lists is a bit tricky, as they are just 4 items height...
Btw, using dragging of the adventure map and minimap works well in this regard. Only heroes/castles lists scrolling have that issue currently.

dood-apo:
> Btw, using dragging of the adventure map and minimap works well in this regard. Only heroes/castles lists scrolling have that issue currently.

Right, because those guys (adventure map and minimap) are the ones who "steal" the events and I'd rather not mess with such a core functionality at this time.

> There're no other issues except I've already mentioned.

Actually I've built this for my phone and played a couple of scenarios. Found one more bug. Let's say that in the selected resolution 4 heroes fit the whole list and the fifth makes the list scrollable. If you have five heroes and play around with dragging the list, then dismiss one hero, the list becomes unclickable :)

Fixed that with the latest commit, so we should be good to go.

dood-apo:
Hi @Branikolog ,

As it was bugging me, that the feature would look half baked without addressing map scrolling interactions, battlefield interactions and extending the swipe range, I took some time to look into the other components. Turns out that fixing the issues was much easier than I thought, so I gave it a try.

With the current PR state (09f2bf6) I think that all of the issues originally identified are addressed.

I am marking this as a draft for now, will open up for the review once I'll play a couple of scenarios on my phone using this code.

dood-apo:
Hi @Branikolog ,

Played one standard scenario and the first 4 missions of Archibald campaign. All seems to be working fine.
Pease check versus the original issues when you have time.

dood-apo:
Any updates @Branikolog , @ihhub , @Districh-ru ?
I've got another touchscreen improvement ready (mitigation of attack direction miss-clicks), but it is based on this one.

src/fheroes2/gui/interface_list.h
```diff
@@ -461,7 +488,8 @@ namespace Interface
                     if ( ActionListCursor( item, mousePos ) )
                         return true;
 
-                    if ( le.MouseClickLeft( rtAreaItems ) ) {
+                    if ( le.MouseClickLeft( rtAreaItems ) && !_lockClick ) {
```
ihhub, line 491: Please swap the conditions since checking a boolean variable is always faster than running a function call.

ihhub (changes requested):
Hi @dood-apo , I left few comments here. Would you mind to take a look at them?

src/engine/localevent.h
```diff
@@ -364,6 +374,7 @@ class LocalEvent
     uint32_t modes;
     fheroes2::Key key_value;
     int mouse_button;
+    bool _isDragInProgress;
```
oleg-derevenetz, line 377: Hi @dood-apo is there any reason to use separate bool flag instead of `enum` constant like for other similar cases?

src/engine/localevent.cpp
```diff
@@ -705,6 +706,9 @@ void LocalEvent::OpenTouchpad()
 LocalEvent & LocalEvent::Get()
 {
     static LocalEvent le;
+    if ( !le.MousePressLeft() ) {
```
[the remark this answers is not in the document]
dood-apo: `_isDragInProgress` member purpose is to be able to identify situations when a mouse press is the very beginning of the drag inside a particular component (in this case scrollable list). So we need to be able to set it to `true` from the list code. However, it is possible to drag out of the region of the list and release the button there. In this case the `_isDragInProgress` needs to be set to false, even though we are not inside the list component event processing anymore.

That was the reason I've placed it it `LocalEvent` singleton - so that `_isDragInProgress` can be set to `false` for all cases, whenever mouse button is released (even if it was released while hovering over a different ui component). So RAII idea does not apply here I think?

Having said that, I completely agree that the current place for the state reset is unfortunate. I have reworked it a bit, please check it that resetting approach is acceptable.

dood-apo:
Hi @Branikolog ,

> I'm still concerning about scrolling lists with tall elements, but I cannot suggest any other solution but making it not synced with the distance your finger moved. Although it's not perfect too. Scrolling of long lists with current implementation could become a bit annoying.

True, it takes a lot swipes to get through the unit list in battle only setup. I would advocate for keeping the swipe scroll content in sync with the finger though, mainly because this is industry standard. To get to the correct monster in a list faster, one can use the scrollbar to roughly set the region, then "fine tune" by swipe scrolling.

> You've extended scrolling range for all lists, so scrolling proceeds, when finger moves out of the list. It works fine everywhere, except the battle journal, where scrolling stops, when we move finger out of the journal area. Was it intentional?

Not at all :). With the latest version this should also be addressed.

src/engine/localevent.h
```diff
@@ -343,6 +353,8 @@ class LocalEvent
         MOUSE_TOUCH = 0x0020,
         // Key on the keyboard is currently being held down
         KEY_HOLD = 0x0040,
+        // Key on the keyboard is currently being held down
```
oleg-derevenetz, line 356: Hi @dood-apo I suppose there must be some other comment here :)

src/fheroes2/gui/interface_list.h
```diff
@@ -417,15 +417,42 @@ namespace Interface
...
+                        // Remember where has the drag started.
+                        _dragStartPos = mousePosition;
+                        le.RegisterDrag();
+
+                        // We have just started the drag, it might as well be a legitimate click.
+                        _lockClick = false;
+                    }
+
+                    const fheroes2::Point deltaPoint = _dragStartPos - mousePosition;
+                    const int delta = ( _scrollbar.isVertical() ? deltaPoint.y : deltaPoint.x );
```
Districh-ru, line 443: Since `deltaPoint` is used only here and only one part of it is used then we don't need to calculate both `x` and `y` at one time and calculate `delta` without creating the `deltaPoint`:
```suggestion
                    const int delta = _scrollbar.isVertical() ? ( _dragStartPos.y - mousePosition.y ) : ( _dragStartPos.x - mousePosition.x );
```

src/engine/localevent.h
```diff
@@ -306,6 +306,16 @@ class LocalEvent
         }
     }
 
+    bool isDragInProgress() const
+    {
+        return modes & DRAG_ONGOING;
+    }
+
+    void RegisterDrag()
```
Districh-ru, line 314: We can name the new method according to the [suggested code style](https://github.com/ihhub/fheroes2/wiki/Code-style) :)
```suggestion
    void registerDrag()
```

Districh-ru (changes requested):
Hi, @dood-apo, nice work!
I left two comments, could you please check them when you have time.

Districh-ru (approved):

Branikolog (approved):
Great work, @dood-apo !

ihhub:
@dood-apo , thank you so much for this improvement!

Issue #8409, opened 2024-02-15
Combat, functioning of genie ability and damage display


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In the original game, when the genie's ability comes into play, half the number of creatures, rounded up, are removed from the initial group, while the "first" creature's hit points are retained.
In the example, the troop of 13 giants is reduced to 6 giants (13 / 2 = 6,5 rounded up to 7, 13 - 7 = 6), and the 72 hit points of the first creature are retained.

In fheroes2, this is not the case.
The troop of 13 giants is reduced to 7 giants, and the 133 hit points of the first creature are reduced to 67 hit points.

Moreover, as the ability reduces the number of creatures, it is not necessary to display the damage supposedly suffered in the combat logs.
Only the number of creatures eliminated should be displayed, in the example with the correction applied:
"Genie destroys half the enemy troop!
7 giants perishs."
Eventually on the same line:
"Genie destroys half the enemy troop, 7 giants perish!"

In Heroes 2:

https://github.com/ihhub/fheroes2/assets/43583994/87cb9db0-be66-47ba-948d-9ef759032812

In fheroes2:

https://github.com/ihhub/fheroes2/assets/43583994/07febead-0802-4c0c-a895-defcef9abb44

### Save file

Here are the save file and the test scenario file:
[Genie ability bug 01.zip](https://github.com/ihhub/fheroes2/files/14296624/Genie.ability.bug.01.zip)

### Additional info

Related closed issue: https://github.com/ihhub/fheroes2/issues/3609.

Alucard648:
I think, triggering Genie`s ability should also play spell effect like "Creature bisected in half with massive blood splash ala Heroes 1", complete with minor quake, just like massive hits in Heroes 7.

LeHerosInconnu:
A more recent save for this (damage display) (1.1.9 9966):
[Before genie dam elemental app 01.zip](https://github.com/user-attachments/files/21194558/Before.genie.dam.elemental.app.01.zip)

In fheroes2:

[...]

PR #8411, opened 2024-02-15
Fix the mechanics of Genie special ability

oleg-derevenetz:
Fix part of #8409 regarding a bug in the game logic.

### DIFF
--- a/src/fheroes2/battle/battle_action.cpp
+++ b/src/fheroes2/battle/battle_action.cpp
@@ -916,17 +916,18 @@ Battle::TargetsInfo Battle::Arena::GetTargetsForDamage( const Unit & attacker, U
 
         const auto abilityIter = std::find( attackerAbilities.begin(), attackerAbilities.end(), fheroes2::MonsterAbility( fheroes2::MonsterAbilityType::ENEMY_HALFING ) );
         if ( abilityIter != attackerAbilities.end() && _randomGenerator.Get( 1, 100 ) <= abilityIter->percentage ) {
-            const uint32_t numToBeKilled = defender.GetCount() / 2 + defender.GetCount() % 2;
+            const uint32_t halvingDamage = ( defender.GetCount() / 2 + defender.GetCount() % 2 ) * defender.Monster::GetHitPoints();
+            if ( halvingDamage > res.damage ) {
+                // Replaces damage, not adds extra damage
+                res.damage = std::min( defender.GetHitPoints(), halvingDamage );
 
-            // Replaces damage, not adds extra damage
-            res.damage = std::min( defender.GetHitPoints(), numToBeKilled * defender.Monster::GetHitPoints() );
+                Interface * iface = Arena::GetInterface();
+                if ( iface ) {
+                    std::string str( _n( "%{name} destroys half the enemy troops!", "%{name} destroy half the enemy troops!", attacker.GetCount() ) );
+                    StringReplace( str, "%{name}", attacker.GetName() );
 
-            Interface * iface = Arena::GetInterface();
-            if ( iface ) {
-                std::string str( _n( "%{name} destroys half the enemy troops!", "%{name} destroy half the enemy troops!", attacker.GetCount() ) );
-                StringReplace( str, "%{name}", attacker.GetName() );
-
-                iface->SetStatus( str, true );
+                    iface->SetStatus( str, true );
+                }
             }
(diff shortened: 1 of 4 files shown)


Alucard648:
The string that logs Genies ability to insta-kill half of enemy stack should be all caps to empasize extreme nature of creature`s absolute destruction.

oleg-derevenetz:
> The string that logs Genies ability to insta-kill half of enemy stack should be all caps to empasize extreme nature of creature`s absolute destruction.

It contains an exclamation mark at the end :) In any case, this PR is intended to fix the logic of the Genie ability. I don't know for sure why is this message displayed in its current form, but I suppose that this can be corrected/improved later separately if there is such a need.

ihhub:
I am sure @Branikolog is going to fight against changing Genie's special ability chances :)

oleg-derevenetz:
Hi @ihhub

> I am sure @Branikolog is going to fight against changing Genie's special ability chances :)

But they were not changed. `_randomGenerator.Get( 1, 10 ) == 2` in the `master` branch is exactly 10% (one chance in 10). If only there was a `<=` comparison instead of `==`, then it would be a different matter.

ihhub (approved):

Branikolog (approved):
Everything works nice!
I hope chances for applying halving were left the same as before.

oleg-derevenetz:
Hi @Branikolog

> I hope chances for applying halving were left the same as before.

By the way, this is an ambiguous question :) According to the [game manual](https://handbookhmm.ru/2-monsters) and in fact it was 10%, but it was shown in unit's properties as 20%. Now it is in fact 10% and is shown as 10%.

Branikolog:
Hello, @oleg-derevenetz 

> Hi @Branikolog
> 
> > I hope chances for applying halving were left the same as before.
> 
> By the way, this is an ambiguous question :) According to the [game manual](https://handbookhmm.ru/2-monsters) and in fact it was 10%, but it was shown in unit's properties as 20%. Now it is in fact 10% and is shown as 10%.

There was a discussion 1+year ago regarding genies ability. Some players complained, that in fheroes2 "halving" occurred too rare, comparing to the original game. After making 150+ fights I came to conclusion, that in the original game the ability triggered in more than 15% attacks. So that's why it was set to 20% in fheroes2. 
I'm not sure we should roll this change back, since players got used to more frequent triggering and the overall weakness of this troop without this special ability.

oleg-derevenetz:
@Branikolog the funny thing is that change to 20% was obviously made in wrong way. Currently in the `master` branch the logic looks like this: `if ( _randomGenerator.Get( 1, 10 ) == 2) {... perform the having... }`, i.e. some random number in the range of 1...10 is taken and if this number is 2 (not less than or equal to 2!) then the halving is performed, so currently in the `master` branch the chance is still 10%. So in fact this PR changes nothing in regard to actual chance to trigger this ability, it just fixes the wrong chance displayed in the unit's properties. I can make the actual chance (as well as displayed chance) 20% if you wish.

Branikolog:
@oleg-derevenetz 

> @Branikolog the funny thing is that change to 20% was obviously made in wrong way. Currently in the `master` branch the logic looks like this: `if ( _randomGenerator.Get( 1, 10 ) == 2) {... perform the halving...}`, i.e. some random number in the range of 1...10 is taken and if this number is 2 (not less than or equal to 2!) then the halving is performed, so currently in the `master` branch the chance is still 10%. So in fact this PR changes nothing in regard to actual chance to trigger this ability, it just fixes the wrong chance displayed in the unit's properties. I can make the actual chance (as well as displayed chance) 20% if you wish.

To be honest, I don't remember, whether I've tested triggering chances in fheroes2. I was mainly focused on testing the original game leaving the implementation in fheroes2 to the development/contribution team.

I'm a bit busy this week, so I cannot perform deep tests. But I can try to make another set of Genies tests in the OG later to be assured, that 20% is an actual chance for triggering.

oleg-derevenetz:
Hi @Branikolog

> I'm a bit busy this week, so I cannot perform deep tests. But I can try to make another set of Genies tests in the OG later to be assured, that 20% is an actual chance for triggering.

Currently with this PR the chance of triggering is 10%, and it was 10% before merging of this PR too due to an implementation error. Judging by the fact that no one actually noticed this error, it seems that 10% is fine. I can change it to 20% if you want.

PR #8419, opened 2024-02-17
Fix the troops carry over mechanics for campaign

### LINKED ISSUE #8418: Crash when selecting Roland hero in final Roland campaign map and he has empty carry over army
### Preliminary checks

### Platform

Windows

### Describe the bug
EDIT: Please see the post right below this. 

Loading up the final map in Roland's campaign in the attached save and saw that he has an empty movement point bar. Selecting him triggers this assertion in `heroes.cpp`:
image

image

### Save file

[Good_9_Complete.zip](https://github.com/ihhub/fheroes2/files/14317040/Good_9_Complete.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #8418

According to my observations, in the original game, troops for carry over are always taken from the one of the heroes available in the kingdom who was hired first (including the hero originally present on the map at the beginning of the scenario). That is, if nothing happened to the hero initially present on the map, then his troops will be taken. If this hero was lost, the troops of the hero who was hired after him will be taken. If this hero was lost, but then rehired, then the troops of the hero who was hired after him will still be taken. It doesn't matter which hero won the last battle of the campaign scenario.

If the scenario was won, but there is not a single hero in the human player's kingdom, then no troops are carried over, and in the mission to which they were supposed to be carried, the corresponding hero starts with the minimum initial army.

### DIFF
--- a/src/fheroes2/battle/battle_main.cpp
+++ b/src/fheroes2/battle/battle_main.cpp
@@ -472,15 +472,6 @@ Battle::Result Battle::Loader( Army & army1, Army & army2, int32_t mapsindex )
             necromancySkillAction( *winnerHero, result.killed, winnerHero->isControlHuman() );
         }
 
-        if ( winnerHero ) {
-            const Heroes * kingdomHero = dynamic_cast<const Heroes *>( winnerHero );
-            if ( kingdomHero ) {
-                Kingdom & kingdom = kingdomHero->GetKingdom();
-
-                kingdom.SetLastBattleWinHero( *kingdomHero );
-            }
-        }
-
         break;
     }
 
--- a/src/fheroes2/game/game_campaign.cpp
+++ b/src/fheroes2/game/game_campaign.cpp
@@ -499,8 +499,17 @@ namespace
     void replaceArmy( Army & army, const std::vector<Troop> & troops )
     {
         army.Clean();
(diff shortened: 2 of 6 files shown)


oleg-derevenetz:
> I'm not saying we need to follow the OG here. I think this PR's logic is perfectly fine, and there is an infinitesimally small chance that any player will complain about this difference.

Normally this shouldn't happen at all because there is no way to NOT carry over the army, because you cannot capture the target town having no heroes.

> I think we should change the right-click info for the carry-over forces when the hero will start with minimal troops:

This right-click info is "general" info, it just explains the general meaning of this award and doesn't report the exact troops that will be carried over. I think this change would be redundant, especially for an option that is virtually impossible (without using special tricks like patching the sources or applying the (currently not even implemented) cheat codes).

oleg-derevenetz:
> Just to clarify, it does show the composition of the carry-over army, and the descriptive text isn't very good. It should probably say "This army..." . However, improving that isn't really in the scope of this PR.

Ah, apparently, I did not follow the campaign UI improvements. I specialize more in game logic :) Well, in theory, yes, then we can add some text for the case of an empty army. You can do it yourself as part of a separate PR if you want :)

ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for this fix!

PR #8420, opened 2024-02-17
Adding Dockerfile and CPack commands to build .deb package within a Docker container

dmikushin:
Dear all,

I'm not satisfied with the provided installation options. This PR offers building a binary Debian package in a Docker container. The package is created straight from CMake via CPack. Some info fields for the Debian package are placeholders, please feel free to fill them in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,3 +212,18 @@ if(NOT MACOS_APP_BUNDLE)
 			)
 	endif(WIN32)
 endif(NOT MACOS_APP_BUNDLE)
+
+set(CPACK_GENERATOR "DEB")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Your Name") # replace with your name
+set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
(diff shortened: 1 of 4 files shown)


ihhub:
Hi @dmikushin , can you please explain the need of Docker usage over normal CMake or even make approach? fheroes2 is available as a debian package as well via standard package managers.

dmikushin:
Hi Ihar, thank you for pointing this out. My goal was to get an installable package for the latest codebase, like a "nightly". So I thought why not to have it, if it takes only 5 minutes to put together?

Thank you for all your hard work, fheroes2 is a lot of joy for those like me who played the original game as a child.

ihhub:
Hi @dmikushin , my point is that Docker based solution is an extra complication to achieve the same goal. GitHub Actions run within Docker containers. A normal user won't have Docker installed on their machine. I also suggest to take a look at #7048 as we had similar discussion in the past.

Moreover, if a person has a way to compile the project of their machine then no need to generate a Debian package at all. Running `make` command will do the same magic and without extra dependencies.

PR #8429, opened 2024-02-24
Fix invalid calculation of "short" object passability

ihhub:
relates to #4495


ihhub:
Closing this pull request as it is very outdated and the solution doesn't work.

PR #8454, opened 2024-03-05
Update DK translation (v1)

Stisen1:
- Reaching for a 100% translation mark.
- Changes to some wordings for easier interpretation and scanning through tool tips. 
- Changed the skill class "Nekromantik" to "Nekromanti" as Danish dictionaries suggest that "nekromanti" covers necromancy as a performance or act rather than an ideology. We don't actually have a word for necromancy as a skill. ALSO... "nekromantik" doesn't fit inside the skill window on the Hero Screen Window.


ihhub (approved):

ihhub:
@Stisen1 , thank you so much for the update!

PR #8459, opened 2024-03-06
fixed double daily resources award

dood-apo:
`GetIncome()` already handles campaign awards!

### DIFF
--- a/src/fheroes2/kingdom/kingdom.cpp
+++ b/src/fheroes2/kingdom/kingdom.cpp
@@ -25,8 +25,6 @@
 
 #include <algorithm>
 #include <cassert>
-#include <cstddef>
-#include <memory>
 #include <ostream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @dood-apo what bothers me a bit here is that these awards should be available only to human-controlled players, but it seems that `Kingdom::GetIncome()` doesn't check this, so AI-controlled players can also get these bonuses, don't they?

dood-apo:
> Hi @dood-apo what bothers me a bit here is that these awards should be available only to human-controlled players, but it seems that `Kingdom::GetIncome()` doesn't check this, so AI-controlled players can also get these bonuses, don't they?

Good catch! Verified in the debugger and this was actually the case :)
Fixed.

ihhub (approved):

oleg-derevenetz (approved):

ihhub:
@dood-apo , thank you very much for the fix!

PR #8463, opened 2024-03-07
Fix rendering flaws of battle journal when using scroll bar

### LINKED ISSUE #8460: Combat screen, rendering flaws of battle journal scroll bar
### Preliminary checks

### Platform

Windows

### Describe the bug

There's a rendering issue while dragging scroll bar in batte journal.

https://github.com/ihhub/fheroes2/assets/55348946/f085e55e-6861-463f-83a3-28449d4a5acd

### Save file

None.

### Additional info

_No response_

Districh-ru:
Fix #8460 
This PR does `listlog` redraw and render it on the screen when it is needed.

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -2796,12 +2796,23 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
     // Add offsets to inner objects
     const fheroes2::Rect mainTowerRect = main_tower + _interfacePosition.getPosition();
     const fheroes2::Rect turnOrderRect = _turnOrder + _interfacePosition.getPosition();
-    const int32_t battleFieldHeight = _interfacePosition.height - status.height - ( listlog->isOpenLog() ? listlog->GetArea().height : 0 );
-    const fheroes2::Rect battleFieldRect{ _interfacePosition.x, _interfacePosition.y, _interfacePosition.width, battleFieldHeight };
-    if ( listlog && listlog->isOpenLog() && ( le.MouseCursor( listlog->GetArea() ) || le.MousePressLeft( listlog->GetArea() ) ) ) {
+    fheroes2::Rect battleFieldRect{ _interfacePosition.x, _interfacePosition.y, _interfacePosition.width, _interfacePosition.height - status.height };
+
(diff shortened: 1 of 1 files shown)


ihhub (approved):

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2796,9 +2800,17 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
...
+
+    fheroes2::Rect lislogRect;
+    const bool isListlogOpen = listlog && listlog->isOpenLog();
+
+    if ( isListlogOpen ) {
+        lislogRect = listlog->GetArea();
+        battleFieldRect.height -= lislogRect.height;
+    }
+
+    if ( isListlogOpen && ( le.MouseCursor( lislogRect ) || le.MousePressLeft( lislogRect ) ) ) {
```
dood-apo, line 2813: Is this a refactor? I can't spot any functional change in this part.
If the idea was to fix the missing check for `listlog` existence in line 2799, then why not do only this:
```
const int32_t battleFieldHeight = _interfacePosition.height - status.height - ( ( listlog && listlog->isOpenLog() ) ? listlog->GetArea().height : 0 );
```
?

Districh-ru: Hi @dood-apo, yes, it is a refactor. Initially I made the fix here, but later moved it to a better place and leaved the changes made here.
Yes, the first one thing it to add a check for `listlog` ( `!= nullptr` ).
And the second - is to avoid three `listlog->GetArea()` calls if listlog is opened: we now cache it to `lislogRect`.

dood-apo: Is it better though? That caching adds a substantial amount of code, when `listlog->GetArea()` is just a plain getter.
Also the flow is not atomic now, since `lislogRect` is initialized to zeros and populated with `listlog->GetArea()` only within a conditional. What do you think?

Districh-ru: @dood-apo, thanks for the suggestion!
Now `lislogRect` is a const reference and will be initialized only if battle log is opened.

ihhub:
@Districh-ru , a bunch of thanks for this fix!

PR #8464, opened 2024-03-07
added possibility to "draw" the attack angle between two hexes

### LINKED ISSUE #7070: Add a new UI element during battle to avoid touchscreen misclicks
### Preliminary checks

Playing a battle of a mobile phone with a small screen (even 5") there is always a chance to misclick which leads to frustration and disappointment from the gameplay. We should account for cases when it's not possible to click precisely by a finger.

When attaching a monster we might display a wheel of directions (actually a hexagon) which show all possible directions the player can attack the opponent. In this case it is not needed to do precise movement on a screen to make an attack from a specified angle.

### Additional info

_No response_

dood-apo:
On the smaller touch screens (mobile phones) it is quite difficult to input the correct attack angle. Currently the only option is to click on the appropriate sub-region of a target hex, which is an extremely small area on mobile phones.

This PR addresses this inconvenience, as presented on a video below. An additional input method is introduced, namely "drawing" an attack angle between the source cell and the target cell of attack. Once the attack direction sword is drawn by this method, it gets confirmed by tapping the target cell.

https://github.com/ihhub/fheroes2/assets/84446398/b4ec84f1-b297-43f2-924a-7deb1bb9f97b

Close #7070

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -2800,7 +2800,7 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
     const fheroes2::Rect battleFieldRect{ _interfacePosition.x, _interfacePosition.y, _interfacePosition.width, battleFieldHeight };
 
     // Swipe attack motion finished, but the destination was outside the arena. We need to clear the swipe attack state.
-    if ( !le.isDragInProgress() && Board::isValidIndex( _swipeAttack.swipeSrcCellIndex ) && !Board::isValidIndex( index_pos ) ) {
+    if ( !le.isDragInProgress() && Board::isValidIndex( _swipeAttack.srcCellIndex ) && !Board::isValidIndex( index_pos ) ) {
         _swipeAttack = {};
     }
 
@@ -2935,11 +2935,11 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
     else if ( le.MouseCursor( battleFieldRect ) ) {
         int themes = GetBattleCursor( msg );
 
-        if ( _swipeAttack.isValid ) {
+        if ( _swipeAttack.isValid() ) {
             // The swipe attack motion is either in progress or has finished.
-            if ( index_pos == _swipeAttack.swipeDstCellIndex ) {
+            if ( index_pos == _swipeAttack.dstCellIndex ) {
                 // The cursor is above the stored destination, we should display the stored attack theme.
-                themes = _swipeAttack.swipeDstTheme;
+                themes = _swipeAttack.dstTheme;
             }
             else {
                 // The cursor has left the destination. Abort the swipe attack.
@@ -2947,12 +2947,12 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
                 _boardActionIntent = {};
             }
         }
-        else if ( _swipeAttack.isValidSwipeAttackDestination( themes, index_pos, _currentUnit->GetHeadIndex(), _currentUnit->GetTailIndex() ) ) {
+        else if ( _swipeAttack.isValidDestination( themes, index_pos ) ) {
             // Valid swipe attack target cell. Calculate the attack angle based on destination and source cells.
-            themes = GetSwordCursorDirection( Board::GetDirection( index_pos, _swipeAttack.swipeSrcCellIndex ) );
+            themes = GetSwordCursorDirection( Board::GetDirection( index_pos, _swipeAttack.srcCellIndex ) );
 
             // Remember the swipe destination cell and theme.
-            _swipeAttack.storeDst( themes, index_pos );
+            _swipeAttack.setDst( themes, index_pos );
 
             // Clear any pending intents. We don't want to confirm previous actions by performing swipe attack motion.
             _boardActionIntent = {};
@@ -2985,12 +2985,12 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
                 MousePressRightBoardAction( *cell );
             }
             else if ( le.MousePressLeft( battleFieldRect ) ) {
-                if ( !le.isDragInProgress() && !_swipeAttack.isValid ) {
+                if ( !le.isDragInProgress() && !_swipeAttack.isValid() ) {
                     le.registerDrag();
 
                     // Remember the swipe source cell and theme.
                     _swipeAttack = {};
-                    _swipeAttack.storeSrc( themes, index_pos );
+                    _swipeAttack.setSrc( themes, index_pos, _currentUnit );
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/fheroes2/battle/battle_interface.h
```diff
@@ -477,6 +477,57 @@ namespace Battle
...
+                }
+
+                return true;
+            }
+
+            int32_t swipeSrcCellIndex;
+            int32_t swipeDstCellIndex;
+            bool isValid;
+            int swipeSrcTheme;
+            int swipeDstTheme;
```
oleg-derevenetz, line 526: I believe, it would be better to initialize the struct fields in-place:

```suggestion
            int32_t swipeSrcCellIndex{ -1 };
            int32_t swipeDstCellIndex{ -1 };
            bool isValid{ false };
            int swipeSrcTheme{ Cursor::NONE };
            int swipeDstTheme{ Cursor::NONE };
```

src/fheroes2/battle/battle_interface.h
```diff
@@ -477,6 +477,57 @@ namespace Battle
...
+        struct SwipeAttack
+        {
+            void clear()
+            {
+                swipeSrcCellIndex = -1;
+                swipeDstCellIndex = -1;
+                isValid = false;
+                swipeSrcTheme = 0;
+                swipeDstTheme = 0;
+            }
```
oleg-derevenetz, line 489: ... and this will allow to remove this method and use just:

```cpp
_swipeAttack = {};
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1171,6 +1171,7 @@ Battle::Interface::Interface( Arena & battleArena, const int32_t tileIndex )
     , _teleportSpellSrcIdx( -1 )
     , listlog( nullptr )
     , _bridgeAnimation( { false, BridgeMovementAnimation::UP_POSITION } )
+    , _swipeAttack( { -1, -1, false, 0, 0 } )
```
oleg-derevenetz, line 1174: ... and to remove this initialization.

src/fheroes2/battle/battle_interface.h
```diff
@@ -523,6 +574,11 @@ namespace Battle
                 _intent = intent;
             }
 
+            void clearStoredIntent()
+            {
+                _storedIntent = BoardActionIntent{};
```
oleg-derevenetz, line 579: You can as well clear the `Battle::Interface::_boardActionIntent` directly instead, because `_storedIntent` is a reference to this field. Since this method is called from the `Battle::Interface` methods, you can replace its call by this:

```cpp
_boardActionIntent = {};
```

oleg-derevenetz (commented):
Hi @dood-apo I just left a few suggestions, please take a look when you have time.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2927,7 +2933,32 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
...
+            else {
+                // The cursor has left the destination. Abort the swipe attack.
+                _swipeAttack = {};
+                _boardActionIntent = {};
+            }
+        }
+        else if ( _swipeAttack.isValidSwipeAttackDestination( themes, index_pos, _currentUnit->GetHeadIndex(), _currentUnit->GetTailIndex() ) ) {
+            // Valid swipe attack target cell. Calculate the attack angle based on destination and source cells.
+            int direction = Board::GetDirection( index_pos, _swipeAttack.swipeSrcCellIndex );
+            themes = GetSwordCursorDirection( direction );
```
oleg-derevenetz, line 2953: Since the `direction` it is not used anywhere except for this single line, we can avoid this temporary variable:

```suggestion
            themes = GetSwordCursorDirection( Board::GetDirection( index_pos, _swipeAttack.swipeSrcCellIndex ) );
```

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -2942,11 +2973,27 @@ void Battle::Interface::HumanBattleTurn( const Unit & unit, Actions & actions, s
             boardActionIntentUpdater.setIntent( { themes, index_pos } );
 
             if ( le.MouseClickLeft( battleFieldRect ) ) {
-                MouseLeftClickBoardAction( themes, *cell, boardActionIntentUpdater.isConfirmed(), actions );
+                bool isConfirmed = boardActionIntentUpdater.isConfirmed();
```
oleg-derevenetz, line 2976: This variable can be `const`.

```suggestion
                const bool isConfirmed = boardActionIntentUpdater.isConfirmed();
```

oleg-derevenetz (commented):
Hi @dood-apo I just left a couple of minor suggestions, please take a look when you have time.

oleg-derevenetz (approved):

src/fheroes2/battle/battle_interface.h
```diff
@@ -477,6 +477,47 @@ namespace Battle
...
+                if ( swipeSrcTheme != Cursor::WAR_MOVE && swipeSrcTheme != Cursor::WAR_FLY && swipeSrcCellIndex != head && swipeSrcCellIndex != tail ) {
+                    return false;
+                }
+
+                return true;
+            }
+
+            int32_t swipeSrcCellIndex{ -1 };
+            int32_t swipeDstCellIndex{ -1 };
+            bool isValid{ false };
```
ihhub, line 514: We can remover this member and have the corresponding function:
```cpp
bool isValid() const
{
    return ( swipeSrcCellIndex != -1 ) && ( swipeDstCellIndex != -1 );
}
```

because **isValid** is set only within `storeDst()` call which is risky.

src/fheroes2/battle/battle_interface.h
```diff
@@ -477,6 +477,47 @@ namespace Battle
...
+                swipeSrcTheme = theme;
+                swipeSrcCellIndex = index;
+            }
+
+            void storeDst( int theme, int32_t index )
+            {
+                isValid = true;
+                swipeDstTheme = theme;
+                swipeDstCellIndex = index;
+            }
```
ihhub, line 493: Should we call them **setSrc()** and **setDst()**? We barely use `store` verb in the code.

ihhub (changes requested):
Hi @dood-apo , I left several comments here. Could you please take a look at them?

src/fheroes2/battle/battle_interface.h
dood-apo: Being valid takes more than that. But you have a point, that it is more robust to inspect the structure and assess if it is valid than rely on a member set externally by client code calls to `setDst`. I've refactored a bit more to be able to reuse `isValidDestination` for `isValid` purposes.

dood-apo:
By the way, aren't we approaching the moment where `SwipeAttack` is warranted its own compilation unit?

ihhub (approved):

oleg-derevenetz (approved):

dood-apo:
> Hi, @dood-apo. Could you please solve the merge conflict if you have the time?

Done!

ihhub:
@dood-apo , huge thanks for this improvement!

PR #8467, opened 2024-03-09
Add a window telling about demo version being used

ihhub:
Based on people misunderstanding that we are providing a game engine and not a game. This window will appear one time when a player opens "New game" dialog.

### DIFF
--- a/src/fheroes2/game/fheroes2.cpp
+++ b/src/fheroes2/game/fheroes2.cpp
@@ -253,16 +253,15 @@ namespace
         std::unique_ptr<fheroes2::h2d::H2DInitializer> _h2dInitializer;
     };
 
-    // This function checks the possible situation when a user uses a demo version
-    // of the game. There is no 100% way to detect it so many things are made with
-    // assumptions.
+    // This function checks for a possible situation when a user uses a demo version
+    // of the game. There is no 100% certain way to detect this, so assumptions are made.
     bool isProbablyDemoVersion()
     {
         if ( Settings::Get().isPriceOfLoyaltySupported() ) {
             return false;
         }
 
-        // A demo version of the game only has 1 map.
+        // The demo version of the game only has 1 map.
         const ListFiles maps = Settings::FindFiles( "maps", ".mp2", false );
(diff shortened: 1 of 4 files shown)


src/fheroes2/game/fheroes2.cpp
```diff
@@ -252,6 +252,20 @@ namespace
...
+    // This function checks the possible situation when a user uses a demo version
+    // of the game. There is no 100% way to detect it so many things are made with
+    // assumptions.
+    bool isProbablyDemoVersion()
+    {
+        if ( Settings::Get().isPriceOfLoyaltySupported() ) {
+            return false;
+        }
+
+        // A demo version of the game only has 1 map.
```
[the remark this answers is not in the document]
dood-apo: Also "has only" sounds more natural I think.

src/fheroes2/game/game_newgame.cpp
```diff
@@ -548,6 +548,13 @@ fheroes2::GameMode Game::NewGame()
 
     LocalEvent & le = LocalEvent::Get();
 
+    if ( isProbablyDemoVersion ) {
+        fheroes2::showStandardTextMessage( _( "Warning!" ),
+                                           _( "You appear to be using the demo version of Heroes of Might and Magic II. "
+                                              "Please note that this version only allows one map to be played." ),
+                                           Dialog::OK );
```
dood-apo, line 555: Does it make clear that fheroes2 is a game engine and modernizes whatever version of the original homm2 that the user has? I think this message can still be misunderstood as fheroes2 being a some kind of demo. How about something like this:
```
fheroes2 needs data files from the original Heroes of Might and Magic II to operate.
You appear to be using the demo version of Heroes of Might and Magic II for this purpose.
Please note that only one scenario will be available in this setup.
```
?

ihhub: Sure. Your suggested message is better than mine. I will change it.

oleg-derevenetz (approved):

dood-apo (approved):

Issue #6921, opened 2023-04-01
Incorrect area threat evaluation when a Barrier is present


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

AI does not take into an account a situation when a barrier is present which is blocking the way and essentially dividing the area by 2 parts. Load the save, pass the control to AI and see that AI is actually prioritizing fog discovery over barrier passing.

### Save file

[Wizard_s_Land_3_0050.zip](https://github.com/ihhub/fheroes2/files/11128357/Wizard_s_Land_3_0050.zip)

Issue #8382, opened 2024-02-09
AI spell usage


crypt47:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

In current save file an enemy in the castle has pretty high Spell Power and a destructive spell Meteor shower (or something like that). I know this because on some other conditions the spell was used with devastating effect. But currently the enemy hero uses Berserk spell which I simply cancel by Dispel. The enemy is then locked withing it's castle walls and eliminated with almost zero loses.
[Beltway_0051.sav.zip](https://github.com/ihhub/fheroes2/files/14218172/Beltway_0051.sav.zip)

### Describe the possible solution

This is right, that Berserk is optimal for inverting damage. But it can be canceled too easily. The enemy should be more bitchy in presense of strong ranged attackers plus after castle walls. He should switch to "I take you with me" scenario.

ihhub:
Hi @oleg-derevenetz , I believe we need to adjust the priority for Berserk spells for situations when the opponent hasn't casted a spell and when the creature is not the next in a row to make an action. What do you think?

oleg-derevenetz:
Hi @ihhub AFAIK the AI trend to use the non-damage spells in such a way that they could be easily dispelled by an opponent is well known (this applies not only to Dispel, but to other control spells as well - e.g. Slow, Blind, etc). I'm afraid that trying to solve this problem head-on by simply changing priorities will lead to the fact that AI will almost always use damage spells instead, because it always casts its spell during its first turn. Ideally, we need to implement some kind of "waiting for the right moment" mechanism, but I don't have any good ideas yet on how to formalize this very concept of the "right moment". In the pro gamer's tournaments, the process of choosing a spell and the right moment to apply it is somewhat similar to poker :) Everyone is guessing what spells the opponent will have, if you cast a control spell too early, then it can be dispelled, and if it's too late, then you are risking that the opponent will cast Slow/Mass Slow.

crypt47:
> ... In the pro gamer's tournaments, the process of choosing a spell and the right moment to apply it is somewhat similar to poker :) Everyone is guessing what spells the opponent have, if you cast a control spell too early, then it can be dispelled, and if it's too late, then you are risking that the opponent will cast Slow/Mass Slow.

[...]

Issue #6259, opened 2022-11-26
Adventure AI: calculate influence areas around heroes and castles


idshibanov:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Currently enemy threat calculations rely on rough estimates using map regions logic. With the tools available AI can't be sure if hero moving to a map object (like a Mine) will be attacked next turn or not. Only a vague "this is a dangerous area" penalty applies.

AI needs to plan ahead and to know exact enemy strength for every map tile. Friendly heroes should be accounted for as well since they can counter and "defend" weaker heroes. Only max values should be picked. 

### Describe the possible solution

A possible solution would add a cache layer (vector for O1 access) tracking enemy threat levels (for next turn) and friendly heroes only for every map tile. At the beginning of a turn the cache is reset and visible threats are reevaluated reusing some of the pathfinder logic. We might need a new pathfinder class for it.

When calculating paths to the objects pathfinder should not only tell if object can be reached but also warn if path is dangerous. Do we care about threat in the middle of the path if the final tile is safe? Some things to test and explore here.

We will have to maintain the cache within a single AI turn. Every battle completed should be followed by a check if that was an existing threat. Every tile uncovered should check if we see a new castle/hero followed by an update to the cache.

### Additional info

This approach is closer to what we already do on Battle AI side, might be useful to review it.

fheroes2bugs:
found another example of computer is not calculating area near him: https://www.youtube.com/watch?v=nYjrHK_FFzI

fheroes2bugs:
An another example, version 1.1.9, computer is running near me with only 1 unit, he can see no danger:

https://github.com/user-attachments/assets/d77f5005-55e3-4d7c-9495-203d56b712ef

save file: [suicide.zip](https://github.com/user-attachments/files/20850774/suicide.zip)

PR #8468, opened 2024-03-09
Use the AI pathfinder to assess the threat of enemy heroes instead of relying on the region database

oleg-derevenetz:
To implement the long-standing TODO. This should make the threat assessment on the adventure map more precise for AI. However, AI heroes can still approach stronger enemy heroes, for example, in the following cases:

* When trying to defend their own castles;
* If the path to the "safe" object runs close to the enemy hero (currently, the AI pathfinder always paves the shortest path in terms of terrain). For example, in this case (save file is taken from https://github.com/ihhub/fheroes2/issues/8382#issuecomment-1984016598) AI hero now runs to the safe place (to his castle), but first he has to run past the enemy hero because of the terrain:

https://github.com/ihhub/fheroes2/assets/32623900/6ca25b8e-7c8a-4b9a-bf61-e583579bb9aa

* If the "safe" objects (or tiles in the case of the fog discovery) are located so far away that their priority is relatively low, even taking into account the threat penalty because of an enemy hero(es).

Also this PR simplifies the logic related to the AI personalities. We don't really need the whole `AI::Normal` instances for each player just to handle AI personalities (and nothing else, because currently the `AI::Normal` singleton instance does all the job). The save files are fully compatible with previous versions (before the changes were made).


src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -788,7 +788,7 @@ namespace
     // Multiply by this value if you are getting a FREE upgrade.
     const double freeMonsterUpgradeModifier = 3;
 
-    const double dangerousTaskPenalty = 20000.0;
+    const double dangerousTaskPenalty = 50000.0;
```
idshibanov, line 791: Why such a significant increase? I suspect it's to counter overvalued castles? To be honest I always treated these values as a placeholder since they are quite arbitrary. There should be a multiplication/scaling component rather than flat subtraction otherwise result is rather unpredictable.

Maybe not in this PR but I would add a TODO to take a look at it later.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2117,107 +2117,100 @@ namespace AI
...
-                        assert( 0 );
-                        break;
-                    }
-
-                    if ( castle->GetColor() == hero.GetColor() ) {
-                        // Friendly castles are always the priority so no penalty for them.
-                        break;
-                    }
+                // Pre-cache the pathfinder database for the enemy hero
+                _pathfinder.reEvaluateIfNeeded( *enemyArmy.hero );
```
idshibanov, line 2147: I wonder if there's a performance hit here? Especially on XL maps with fog uncovered. For every potential AI hero decision we're going to pathfind all visible enemy heroes over and over.

src/fheroes2/ai/normal/ai_normal_hero.cpp
oleg-derevenetz: Hi @idshibanov

> Why such a significant increase? I suspect it's to counter overvalued castles?

Well, not just castles, but yes, some objects seems overvalued in comparison with this constant.

> There should be a multiplication/scaling component rather than flat subtraction otherwise result is rather unpredictable.

I don't understand yet how exactly this should be scaled. What wording of the TODO comment do you suggest?

src/fheroes2/ai/normal/ai_normal_hero.cpp
oleg-derevenetz: Hi @idshibanov there will be a performance impact, but, judging by my tests, it is acceptable. On the other hand, I'm afraid if we want accurate calculations, not approximate ones, then we can't do without it, because in the interval between making decisions, the heroes will move over the map, puck up or capture objects, kill wandering monsters, etc.

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: By scaling I mean to take into account base object value when applying a penalty. Don't have exact formula in mind but I would start with knocking down 70-80% of the value first. We can call it normalization for dangerous objects.

That should help balance out the outliers. Currently it's 500 for resource pile vs 50000 for the best castle. With the flat -20k or -50k they could still suicide for a castle sometime. Arguably it's better to steal resources while running away than even capturing mines.

Another approach is to make a special `getDangerousObjectValue` but to be honest those switch cases aren't perfect either.

oleg-derevenetz: @idshibanov I added the appropriate TODO comment for now. BTW this constant is currently used not just as "relative penalty", but also as a specific value in `getXXXObjectValue()` meaning "don't go here".

src/fheroes2/ai/normal/ai_normal_hero.cpp
oleg-derevenetz: I have made efforts to address the concerns regarding the potential performance hit. Accurate estimate (which uses the AI pathfinder) is now only used for enemy heroes located relatively close, for the rest, a rougher (and faster) estimate is used.

src/fheroes2/ai/normal/ai_normal_hero.cpp
crypt47: > Why such a significant increase? I suspect it's to counter overvalued castles?

Please consider, in current version it's very easy to leave some castle empty with a strong hero waiting. All the enemy heroes will go straight to the mouse trap.

crypt47:
> Why such a significant increase? I suspect it's to counter overvalued castles?

Please consider, in current version it's very easy to leave some castle empty with a strong hero waiting. All the enemy heroes will go straight to the mouse trap.

oleg-derevenetz:
> Please consider, in current version it's very easy to leave some castle empty with a strong hero waiting. All the enemy heroes will go straight to the mouse trap.

Unfortunately, this PR alone will not completely solve such issues. This only increases the accuracy of assessing the threat from enemy heroes, and there is also a need to think about improving the system of evaluating object values/penalties, as well as possibly improving the "safety" of the fog detection process.

idshibanov (approved):

ihhub (approved):

ihhub:
Hi @oleg-derevenetz , I believe that the performance is impacted by this change but we can improve it by adding caching mechanism later.

Also, I think this pull request fixes #6921 (I tested it)  and #6259 issues.

oleg-derevenetz:
Hi @ihhub

> but we can improve it by adding caching mechanism later

This is not so easy to do because:

* During one pass (when choosing the hero with the "best target"), threat data for one hero is not suitable for others, because what is a threat to one hero may not be a threat to another;
* Between passes, the threat data should be invalidated and recalculated, because some heroes may move or disappear (including enemy heroes).

In any case, now for potentially threatening heroes who are relatively far away, threat assessment is performed in an approximate but fast way, accurate calculations using AI pathfinder are performed only for threats that our hero can potentially encounter during one turn.

> and https://github.com/ihhub/fheroes2/issues/6259 issues

About this, it's not quite like that, because pathfinder still can pave the path near some threating enemy heroes (although formally, the finishing point of the path may be "safe" in this regard). I was thinking about adding some kind of threat assessment functionality to AI pathfinder so that it avoids tiles that are under threat while paving the path, but here, first of all, there is also a performance problem, and also there may be different corner cases (for example, what if all the tiles around the hero are already considered "dangerous" because the enemy hero approached him during the previous turn?).

In any case, I think we need to somehow fix the target selection mechanism first, because now the hero can choose a target, even if it is dangerous for him, penalties do not save the situation too much. And then we'll see.

ihhub:
@oleg-derevenetz , very big thanks for these valuable improvements!

@Branikolog , please do AI tests on master branch. We needed to proceed because they are more things we need to do in regards to the AI.

crypt47:
> > Please consider, in current version it's very easy to leave some castle empty with a strong hero waiting. All the enemy heroes will go straight to the mouse trap.
> 
> Unfortunately, this PR alone will not completely solve such issues. This PR only increases the accuracy of assessing the threat from enemy heroes, and there is also a need to think about improving the system of evaluating object values/penalties, as well as possibly improving the "safety" of the fog detection process.

I'm sorry if this PR is already merged and addressed but just for the record. I've run 1.0.13 version to test and the improvement is not very noticeable. For example in the save attached, on the next move the hero Gem will face the weak hero Wrethmont, and in the move after that Wethmont is still making circles after more strong hero.
[Beltway_0083.sav.zip](https://github.com/ihhub/fheroes2/files/14641212/Beltway_0083.sav.zip)

oleg-derevenetz:
> I know that in Heroes3 he can't.

This is just not true, at least in HoTA:

https://github.com/ihhub/fheroes2/assets/32623900/80f8f303-07a9-4a79-aed5-69b7f67c71ce

https://github.com/ihhub/fheroes2/assets/32623900/b9c0a037-93d5-4255-9af7-2d9174f345a4

Naturally, hero can retreat to HoMM2 OG as well even if there are no castles in his kingdom:

https://github.com/ihhub/fheroes2/assets/32623900/11457747-05cc-4ee9-a8ff-e2e5cc7a97a7

crypt47:
> > I know that in Heroes3 he can't.
> 
> This not the case, at least in HoTA:

If it's not a bug it's ok. Doesn't really matter. I played native Linux version, it's the very first without any addons.

PR #8475, opened 2024-03-12
Make the default left and right hotkeys select skills in the level up dialog

zenseii:
This makes it possible to use the left and right hot keys to select between two skills in the level up dialog. 

Relates to https://github.com/ihhub/fheroes2/issues/5270


ihhub (approved):

ihhub:

PR #8402, opened 2024-02-13
Remove the AI unit growth bonuses and tune the income bonuses for different difficulty levels

oleg-derevenetz:
Related to #8321.

Also this PR tunes the AI surrender behavior.

PR #8482, opened 2024-03-15
Version 1.0.13

ihhub:
Making the engine better.


oleg-derevenetz:
Hi @ihhub do you have plans to include #8402 into upcoming release taking into account our with @LeHerosInconnu tests? There are still some things that can be improved (for example, to per-castle resource bonuses added), but this requires additional testing in a good way, so it will go to a separate PR.

ihhub:
@oleg-derevenetz , since we are delaying the release for more I think we can include it. I will review it soon.

Districh-ru (approved):

PR #8491, opened 2024-03-17
Hungarian language update #17

big4billy:
Last new adds and some corrections. We are now up-to-date and only the possible new additions and changes need to pay attention in the future ...


ihhub (approved):

ihhub:
@big4billy , huge thanks for this update!

PR #8502, opened 2024-03-20
Fix wrong name for Neutral race troops

zenseii:
If your army is only composed of Neutral troops, then this is what is displayed in current `Master`:
image

I've corrected it to Neutral in this PR as that is what those troops are called.

image

Note that in the OG this dialog would display "Multiple", but IMO that is wrong.


ihhub (approved):

ihhub:

PR #8508, opened 2024-03-21
Small change of phrasing of creature seperation

Mr-Bajs:
This changes _units_ into _creatures_ because in the rest of the game a unit is refered to as a stack of creatues not a single creature. Keeping the game more consistent.

Also changes _slot_ into _unit_ to be more inline of the rest of the terminology of the game. Though keeping instead of _unit_ are also fine. Keeping slots might be more clear, but the image selector makes it clear enough in my mind.

Example: Current on top and this suggestion down below

example


ihhub (approved):

Branikolog (approved):

ihhub:
@Mr-Bajs , many thanks for the correction!

PR #8513, opened 2024-03-22
Implement hero details dialog in Editor

Districh-ru:
https://github.com/ihhub/fheroes2/assets/113276641/28ec5585-fcdd-45dd-a95b-6de89edff6be

This PR also:
- reworks the original Hero Screen dialog;
- slightly improves text input dialog and hero portrait rendering.

The already made resurrection maps with heroes and castles will cause crashes. these heroes and castles should be erased from map in master build and added back in this build (in master build there was no metadata for these objects).


Districh-ru:
Hi, @zenseii, could you please check and fix (if needed) the new phrases in `heroes_dialog.cpp`.

Districh-ru:
Thank you, @zenseii!

Branikolog:
Hi, @Districh-ru 
I've found several issues, could you, please, look, if they are related to the changes of this PR?

There's an issue with updating mana points. I believe, if we set a custom number which exceeds the maximum number of points, the point box shoud show the exact number that was typed. Currently increasing hero knowledge we increase the number of point we entered in a field.

https://github.com/ihhub/fheroes2/assets/55348946/d484b98c-c700-4133-8522-16a6947246fe

There's also a graphic flaw around spell points box, when the number exceeds 999 for both sides.
After revisiting hero window mana points could be resetted:

https://github.com/ihhub/fheroes2/assets/55348946/4fd7256f-700b-4648-88ad-b4971ee84f48

If we enter magic book additional mana points are also resetted.

Do we allow 999+ mana points at all? Magic book bookmark shows 999 anyway here.
image

Also, I believe, if we place a default hero window should show the default primary attributes for the class. In the current build they are 0-0-1-0. (I even not sure, we should allow 0 knowledge)
image

Currently hero army slots are empty and it's not clear whether hero starts with the default army or some custom one. I think we need some sort of visual representation for a default army of the selected hero class.

Districh-ru:
Hi, @Branikolog, the issue with the spell points is fixed. If the a custom value was set then it will not change until it is reset by the right mouse click. (The exception is when the custom value is equal to the maximum value - this is considered as the default value)

> Do we allow 999+ mana points at all? Magic book bookmark shows 999 anyway here. 

Currently it is not allowed to render more than 999 spell points in the spell book. The comment says that more than 999 spell points is a case of the broken code. So lets keep this limit.

> Also, I believe, if we place a default hero window should show the default primary attributes for the class. In the current build they are 0-0-1-0. (I even not sure, we should allow 0 knowledge)

Now the default primary skills' values will be set. We can allow the map maker to set the minimum values "0-0-1-0" values. If he sets knowledge to 0 than he understands that the maximum spell points value will be 0. It may be done for a reason.

> Currently hero army slots are empty and it's not clear whether hero starts with the default army or some custom one. I think we need some sort of visual representation for a default army of the selected hero class (maybe generate some icon with text or blur/fade hero army slots).

Now when the custom army is not set there is a text saying that the default troops will be used for the hero's army:
изображение

Branikolog:
Hi, @Districh-ru !
Thanks for fixing those issues.
Adjusting mana points works well now.
I do also love default troop icon. 👍 

I've just realized, that we have same problem here with primary and secondary skills.
While adjusting exp value hero would have certain skills leveled. I think we need same "default skills" and "default secondary skills" representation to be able to distinguish whether hero has pre set 0-0-1-1 skills or would have some random values, based on a chance.
Also, when placing a random hero I suppose a spell book can appear in his inventory, depending on a class appeared. 
I hope for the cases of full inventory nothing will break in logic, when inventory is already full...
image

There's also an issue with Spell power points. The engine allows to set 0 value, but the actual SP is showed as 1. So in a case of placing artifacts adding SPs the value is decreased by 1 point.

https://github.com/ihhub/fheroes2/assets/55348946/99d4068b-66f7-4253-8d75-b77dad57c937

Branikolog:
@Districh-ru 
Sorry for asking for too much within this PR. If you think, my concerns become too overwhelming you're free to postpone fixes for some future PRs. :)

One more thing I've noticed.
The header of hero window shows an empty space on default. I think, It would be nice to show something like "Random Warlock hero (lvlxxx)" and "Random hero (lvlxxx)" (in a case we selected random hero).
image

Districh-ru:
HI @Branikolog!
Now the minimum Spell Power is limited to 1.
Heroes with no custom names will show "Random ... hero" title.
I made the "Default skill" text for the secondary skills, but it is not so easy to do the same for the primary skills. :( 
If I find a good solution I'll do it in the next two days. If no - we can fix it in a new PR later.
изображение

Branikolog:
@Districh-ru 
Thanks, the above changes work nice!

>I made the "Default skill" text for the secondary skills, but it is not so easy to do the same for the primary skills. :(
If I find a good solution I'll do it in the next two days. If no - we can fix it in a new PR later.

I think we need something to implement, as there's no way to set the default primary skill values once you've changed them. So it is impossible to distinguish whether hero has pre-set values or they will be generated according to the level of the hero:
image

Districh-ru:
Hi, @Branikolog,
I've added a display that the default Primary Skills' values are being used and also now we display the base value in parentheses to the left of the normal value (with artifacts' modifiers).
изображение

To do this I reworked the primary skills bar a little. I also made the same changes to the Primary Skills bar in Battle Only mode (without the reset to default). Could you please test it too when you have some time. :)

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -93,25 +97,25 @@ void PrimarySkillsBar::RedrawItem( int & skill, const fheroes2::Rect & pos, fher
 
         switch ( skill ) {
         case Skill::Primary::ATTACK:
-            fheroes2::Blit( backSprite, 217, 52, dstsf, dstpt.x, dstpt.y, ww, ww );
+            fheroes2::Copy( backSprite, 217, 52, dstsf, dstpt.x, dstpt.y, ww, ww );
```
ihhub, line 100: Just for the future: for these cases it is better to have some sort of verification call that the image is indeed a single-layer image. This should be done only on debug mode. This is outside the scope of this pull request and not a high priority.

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -130,63 +134,188 @@ void PrimarySkillsBar::RedrawItem( int & skill, const fheroes2::Rect & pos, fher
...
+            break;
+        default:
+            break;
+        }
+        return false;
+    }
+
+    fheroes2::showStandardTextMessage( Skill::Primary::String( skill ), Skill::Primary::StringDescription( skill, _hero ), Dialog::OK );
+
     return false;
```
ihhub, line 269: The logic before the changes was that we return **true** for this case. Now we return false.

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -130,63 +134,188 @@ void PrimarySkillsBar::RedrawItem( int & skill, const fheroes2::Rect & pos, fher
...
 bool PrimarySkillsBar::ActionBarLeftMouseSingleClick( int & skill )
 {
-    if ( Skill::Primary::UNKNOWN != skill ) {
-        fheroes2::showStandardTextMessage( Skill::Primary::String( skill ), Skill::Primary::StringDescription( skill, _hero ), Dialog::OK );
-        return true;
+    if ( skill == Skill::Primary::UNKNOWN ) {
+        return false;
     }
 
+    if ( _isEditMode ) {
```
ihhub, line 218: I suggest to have the opposite logic first to avoid a huge code being in a if-body.

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -130,63 +134,188 @@ void PrimarySkillsBar::RedrawItem( int & skill, const fheroes2::Rect & pos, fher
...
+            }
+            break;
+        case Skill::Primary::KNOWLEDGE:
+            if ( primarySkillEditHandler( value, _hero->getKnowledgeBaseValue() ) ) {
+                _hero->setKnowledgeBaseValue( static_cast<int>( value ) );
+                return true;
+            }
+            break;
+        default:
+            break;
```
ihhub, line 262: Please add an assertion here.

src/fheroes2/gui/skill_bar.cpp
```diff
@@ -130,63 +134,188 @@ void PrimarySkillsBar::RedrawItem( int & skill, const fheroes2::Rect & pos, fher
...
+            _hero->setPowerBaseValue( -1 );
+            return true;
+        case Skill::Primary::KNOWLEDGE:
+            if ( _hero->getKnowledgeBaseValue() == -1 ) {
+                return false;
+            }
+            _hero->setKnowledgeBaseValue( -1 );
+            return true;
+        default:
+            break;
```
ihhub, line 305: Please add an assertion here.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -606,6 +608,301 @@ void Heroes::LoadFromMP2( const int32_t mapIndex, const int colorType, const int
...
+    }
+
+    SetSpellPoints( heroMetadata.magicPoints > -1 ? static_cast<uint32_t>( heroMetadata.magicPoints ) : GetMaxSpellPoints() );
+
+    move_point = GetMaxMovePoints();
+}
+
+bool Heroes::updateHeroMetadata( Maps::Map_Format::HeroMetadata & heroMetadata ) const
+{
+    bool hasChanges = false;
```
ihhub, line 762: I am just curious: what if we create a function to compare values of this class before and after the changes? Would it make the logic more straightforward? Then we won't need this complex function.

src/fheroes2/heroes/skill.cpp
```diff
@@ -173,6 +173,26 @@ int Skill::Primary::GetInitialSpell( int race )
...
+        case ATTACK:
+            return ptr->initial_primary.attack;
+        case DEFENSE:
+            return ptr->initial_primary.defense;
+        case POWER:
+            return ptr->initial_primary.power;
+        case KNOWLEDGE:
+            return ptr->initial_primary.knowledge;
+        default:
+            break;
```
ihhub, line 189: Please add an assertion here.

ihhub (changes requested):

Branikolog:
Hi, @Districh-ru 
The effect applied by primary skills seems to be broken.
Look, 14 knowledge provides just 110 mana points:
image

Also, 13 hero defense skill results in overall peasant defense as just 12:
image
image

src/fheroes2/gui/skill_bar.cpp
Districh-ru: I changed in all bars `Blit()` to `Copy()` for the cases when we update the background. And in some places the background is taken from the part of a double-layer image. So not in all cases there should be a single-layer image as input. In example:
изображение

Also we are using the original game resources and we certain know what image part we are copying - with transparency or without it. :)

src/fheroes2/gui/skill_bar.cpp
Districh-ru: IMHO `ActionBarLeftMouseSingleClick()` should return **true** only when the update of the caller dialog is needed.
`showStandardTextMessage()` has the build-in `ImageRestorer` so there is no need to redraw the caller dialog after this action and we return **false**.
So the previous logic was not quite right - it had no harm but wasted some resources to redraw the unchanged dialog.

ihhub: Got it.

src/fheroes2/gui/skill_bar.cpp
ihhub: Not all the original resources are properly created. We fixed some of them manually. The suggested check doesn't need to exist. If the game resources are modified we shouldn't fix modding attempts.

src/fheroes2/heroes/heroes.cpp
Districh-ru: @ihhub, thanks for the suggestion. Now it might be a little slower to update hero metadata but the code looks much better.

Districh-ru:
> Hi, @Districh-ru The effect applied by primary skills seems to be broken. 

Hi, @Branikolog, thanks for noticing this bug. It is fixed now. Could you please test it when you have time.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -606,6 +608,230 @@ void Heroes::LoadFromMP2( const int32_t mapIndex, const int colorType, const int
...
+    }
+
+    // Hero's portrait.
+    heroMetadata.customPortrait = portrait;
+
+    // Hero's artifacts.
+    const size_t artifactCount = bag_artifacts.size();
+    assert( artifactCount == heroMetadata.artifactMetadata.size() );
+    for ( size_t i = 0; i < artifactCount; ++i ) {
+        const int artifactId = bag_artifacts[i].GetID();
```
ihhub, line 783: I think we can avoid having a temporary variable:
```cpp
heroMetadata.artifact[i] = bag_artifacts[i].GetID();

// The spell scroll may contain a spell.
if ( heroMetadata.artifact[i] == Artifact::SPELL_SCROLL ) {
...
```

src/fheroes2/heroes/heroes.h
```diff
@@ -57,6 +57,10 @@ namespace Battle
 namespace Maps
 {
     class Tiles;
+    namespace Map_Format
```
ihhub, line 60: Mostly a code style correction: please add an empty line before namespace to separate it from the class.

src/fheroes2/heroes/heroes_dialog.cpp
```diff
@@ -58,7 +63,56 @@
...
+        case Race::NECR:
+            preparePortrait( ICN::CMBTHRON, 9 );
+            break;
+        case Race::RAND:
+            // TODO: Make a portrait for the random hero.
+            preparePortrait( ICN::NOMAD, 10 );
+            break;
+        default:
+            // Have you added a new race? Correct the logic above!
+            assert( 0 );
```
ihhub, line 108: Please add a break here for consistency.

src/fheroes2/maps/map_format_info.cpp
```diff
@@ -208,4 +209,28 @@ namespace Maps::Map_Format
 
         return true;
     }
+    bool HeroMetadata::operator==( const HeroMetadata & anotherHeroMetadata ) const
+    {
+        {
```
ihhub, line 214: Please remove this extra code block parenthesis.

src/fheroes2/maps/map_format_info.cpp
```diff
@@ -208,4 +209,28 @@ namespace Maps::Map_Format
...
+                return false;
+            }
+
+            if ( !std::equal( armyMonsterType.begin(), armyMonsterType.end(), anotherHeroMetadata.armyMonsterType.begin() )
+                 || !std::equal( armyMonsterCount.begin(), armyMonsterCount.end(), anotherHeroMetadata.armyMonsterCount.begin() )
+                 || !std::equal( artifact.begin(), artifact.end(), anotherHeroMetadata.artifact.begin() )
+                 || !std::equal( artifactMetadata.begin(), artifactMetadata.end(), anotherHeroMetadata.artifactMetadata.begin() )
+                 || !std::equal( availableSpells.begin(), availableSpells.end(), anotherHeroMetadata.availableSpells.begin() )
+                 || !std::equal( secondarySkill.begin(), secondarySkill.end(), anotherHeroMetadata.secondarySkill.begin() )
+                 || !std::equal( secondarySkillLevel.begin(), secondarySkillLevel.end(), anotherHeroMetadata.secondarySkillLevel.begin() ) ) {
```
ihhub, line 229: Since all of these are containers we can easily use `==` operator (see [this](https://en.cppreference.com/w/cpp/container/vector/operator_cmp)). For example,
```cpp
armyMonsterType == anotherHeroMetadata.armyMonsterType
```

ihhub (changes requested):
Hi @Districh-ru , I left few more comments here. Can you please take a look at them once you have time?

src/fheroes2/heroes/heroes.cpp
Districh-ru: Sure, thanks!

src/fheroes2/maps/map_format_info.cpp
Districh-ru: I updated this method and moved it to H file

Branikolog:
@Districh-ru 

> > Hi, @Districh-ru The effect applied by primary skills seems to be broken.
> 
> Hi, @Branikolog, thanks for noticing this bug. It is fixed now. Could you please test it when you have time.

I'm afraid the problem is still there. Are you sure, that you've applied changes? Mana points and creature stats does not correspond the primary skills level chosen.

Also, I think, once you've decided to set any primary skill to custom - all skills should change their state to "custom". Now I can change one skill, but the rest are shown as default.
image

Districh-ru:
Hi @Branikolog, Mana Point and Primary Skills should be fixed now. Also now any spell can be removed from the Spell Book in Editor and in Battle Only mode.
Could you please check it when you have time?

Branikolog:
Hi, @Districh-ru 
I've made a couple of tests and I have found one issue.

Mana dialog text does not fit the window.
image

Also, if we set a custom experience value, there's no clear whether it was modified and is fixed or it is a default value.
image

I suggest showing empty field or "-" in it to indicate, that the value has not been modified yet.
Also it would be nice if default secondary skill icons were visible under "Default skill" . But this can be done in the separate PR, as this PR has already become really large in terms of new elements implemented.

Districh-ru:
Hi, @Branikolog 
> Mana dialog text does not fit the window.

It is fixed now.

> Also, if we set a custom experience value, it's no clear whether it was modified and is fixed or it is just a default value.
> I suggest showing empty field or "-" in it to indicate, that the value has not been modified yet.

Done. Could you please check it when you have some time.

> Also it would be nice if default secondary skill icons were visible under "Default skill" . But this can be done in the separate PR, as this PR has already become really large in terms of new elements implemented.

I suggest to do this in the other PR.

Branikolog (approved):
@Districh-ru 
Excellent!

ihhub (approved):

ihhub:
@Districh-ru . many-many thanks for these changes!

PR #8527, opened 2024-03-29
add option FHEROES2_WITH_SYSTEM_SMACKER

undef21:
to use system libsmacker instead bundled one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@
 # FHEROES2_WITH_ASAN: build with UB Sanitizer and Address Sanitizer (small runtime overhead, incompatible with FHEROES2_WITH_TSAN)
 # FHEROES2_WITH_TSAN: build with UB Sanitizer and Thread Sanitizer (large runtime overhead, incompatible with FHEROES2_WITH_ASAN)
 # FHEROES2_WITH_IMAGE: build with SDL_image (requires libpng)
+# FHEROES2_WITH_SYSTEM_SMACKER: build with external libsmacker instead bundled one
 # FHEROES2_WITH_TOOLS: build additional tools
 # FHEROES2_MACOS_APP_BUNDLE: create a Mac app bundle (only valid when building on macOS)
 # FHEROES2_DATA: set the built-in path to the fheroes2 data directory (e.g. /usr/share/fheroes2)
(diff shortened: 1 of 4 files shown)


ihhub:
Hi @undef21 , can you please explain the reason of using the system library? As far as I remember we ourselves fixed some bugs in this library which comes with the code.

undef21:
Hello, @ihhub, @oleg-derevenetz .

Many linux distributions have policies which require to use system libraries as much as possible instead bundled ones. So this PR is mainly for distribution maintainers.

> Yes, I upgraded it to the latest version some time ago, the original libsmacker (without our patches) was nearly unusable, so I removed the option to use the non-bundled one.

Now Debian and Ubuntu have `libsmacker` in their repositories ([Debian](https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=libsmacker-dev) and [Ubuntu](https://packages.ubuntu.com/search?keywords=libsmacker-dev&searchon=names)) with `tree-size-check` patch and `fheroes2` in that distributions already [uses](https://packages.debian.org/sid/fheroes2) system libsmacker. So I suggest restore building with external libraries instead bundled ones.

Makefile
```diff
@@ -25,6 +25,7 @@
 # FHEROES2_WITH_ASAN: build with UB Sanitizer and Address Sanitizer (small runtime overhead, incompatible with FHEROES2_WITH_TSAN)
 # FHEROES2_WITH_TSAN: build with UB Sanitizer and Thread Sanitizer (large runtime overhead, incompatible with FHEROES2_WITH_ASAN)
 # FHEROES2_WITH_IMAGE: build with SDL_image (requires libpng)
+# FHEROES2_WITH_SYSTEM_SMACKER: build with external libsmacker instead bundled one
```
oleg-derevenetz, line 28:
```suggestion
# FHEROES2_WITH_SYSTEM_SMACKER: build with an external libsmacker instead of the bundled one
```

oleg-derevenetz (changes requested):
Hi @undef21 looks good, the only recommendation is to slightly rephrase the comment. I suppose this has been tested on systems where there is a `libsmacker` from packages?

Makefile
undef21: Thanks. Fixed.

Yes, this has been tested on Debian and Ubuntu, also another Debian derivatives (Kali, Raspbian) use the patched library. Another systems should patch their libsmacker packages to use this options or use the bundled library.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@undef21 , huge thanks for this change!

PR #8531, opened 2024-03-31
Add compression step for the map format

ihhub:
relates to #6845

### DIFF
--- a/src/fheroes2/maps/map_format_info.cpp
+++ b/src/fheroes2/maps/map_format_info.cpp
@@ -193,7 +193,7 @@ namespace Maps::Map_Format
                 return msg;
             }
 
-            // Try to free some memory
+            // Let's try to free up some memory
             temp = std::vector<uint8_t>{};
 
             decompressed.putRaw( decompressedData.data(), decompressedData.size() );
--- a/src/fheroes2/maps/map_format_info.cpp
+++ b/src/fheroes2/maps/map_format_info.cpp
@@ -25,6 +25,7 @@
 #include <type_traits>
 
 #include "serialize.h"
+#include "zzlib.h"
 
 namespace
 {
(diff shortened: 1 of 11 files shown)


src/fheroes2/maps/map_format_info.cpp
```diff
@@ -152,15 +153,56 @@ namespace Maps::Map_Format
...
+
+            const std::vector<uint8_t> decompressedData = Compression::decompressData( temp.data(), temp.size() );
+            if ( decompressedData.empty() ) {
+                // This is a corrupted file.
+                map = {};
+                return msg;
+            }
+
+            // Try to free some memory
+            temp = std::vector<uint8_t>{};
```
oleg-derevenetz, line 197: Initially there was the code like this:

``` cpp
temp = {};
[...]
decompressedData = {};
```

As I understand it, this was done with the intention of freeing up memory. Unfortunately, it doesn't work that way because this syntax calls the `vector& operator=( std::initializer_list<T> )`, which in general is not required to shrink the buffer. There is also [shrink_to_fit()](https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit), but it is non-binding and it is also not required to actually shrink the buffer. In my opinion, the most reliable way is to use the `vector& operator=( vector&& )` overload, which is most likely to exchange states with a temporary vector and then the temporary vector (with the old buffer) [will be destroyed](https://godbolt.org/z/oYWTdv79G).

As for `decompressedData`, it is now `const` and will be destroyed right after the `putRaw()` call because it goes out of scope.

oleg-derevenetz (commented):
Hi @ihhub I've made some small changes, please take a look when you have time.

src/fheroes2/maps/map_format_info.cpp
ihhub: Thank you so much for the correction!

oleg-derevenetz (approved):

Districh-ru (approved):

PR #8532, opened 2024-03-31
Remove Editor's beta flag

ihhub:
relates to #6845


Districh-ru (commented):
Hi, @ihhub, should we in this PR also enable editor only for PoL assets?

Districh-ru (approved):

PR #8533, opened 2024-03-31
Fix cursor rendering within Editor's File Menu dialog

ihhub:
relates to #6845

To reproduce the issue disable Animation within the Editor and use software rendering mode.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -734,6 +734,8 @@ namespace Interface
             le.MousePressLeft( buttonQuit.area() ) ? buttonQuit.drawOnPress() : buttonQuit.drawOnRelease();
             le.MousePressLeft( buttonCancel.area() ) ? buttonCancel.drawOnPress() : buttonCancel.drawOnRelease();
 
+            bool redraw = false;
+
             if ( le.MouseClickLeft( buttonNew.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::EDITOR_NEW_MAP_MENU ) ) {
                 if ( eventNewMap() == fheroes2::GameMode::EDITOR_NEW_MAP ) {
                     result = fheroes2::GameMode::EDITOR_NEW_MAP;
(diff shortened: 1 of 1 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -764,18 +766,27 @@ namespace Interface
             }
             else if ( le.MousePressRight( buttonNew.area() ) ) {
                 fheroes2::showStandardTextMessage( _( "New Map" ), _( "Create a new map, either from scratch or using the random map generator." ), Dialog::ZERO );
+                redraw = true;
```
Districh-ru, line 769: Should we force `display.render()` for the hardware cursor?
```suggestion
                redraw = fheroes2::cursor().isSoftwareEmulation();
```

Districh-ru (commented):
Hi, @ihhub, I left one question that is applied to all `redraw = true;` lines.
The same bug is also in Save/Load menu. And many other places.
Possibly we should fix the other way.
 ...
 maybe: the `CursorRestorer` should set some `bool _needCursorRedraw` to `true` and if there is no rendering preparation (all sprites are copied to the `display`) then we call `display.redraw()` .
 ...
or we can introduce `bool _isRenderLocked` and set it to true when `display` preparation is in process. It should automatically reset to `false` by the main rendering function. And if it is `false` we can update cursor from the `CursorRestorer` destructor.
Initially such update was in that destructor but it caused a lot of graphic bugs in restorers "race" so we removed it.
In other words - we should stop rendering while a lot of restorers updates `display` or other functions do `redraw()` to display. Catch any redraw call and keep the `_needRedraw` state and render only when the `_isRenderLocked` is set to `false`.

This is only a quick concept so it may be wrong. :)

Stisen1:
I've tested the build that @Districh-ru suggests. The code changes doesn't apply debug errors.

Running the game's map editor: 

The animation can be switched on and off as I see fit. It looks nice and smooth and easy to use. 

I am using these fheroes2.cfg settings:

image

ihhub:
Closing this pull request as we accepted another fix.

PR #8539, opened 2024-04-03
Add a window to change Monster properties

idshibanov:
relates to #6845

Later we can make it's own dialog window to control monster join condition, but it outside of OG scope.

Screenshot 2024-04-03 032143

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -273,6 +273,13 @@ namespace
                     objectIter = mapTile.objects.erase( objectIter );
                     needRedraw = true;
                 }
+                else if ( objectIter->group == Maps::ObjectGroup::MONSTERS ) {
+                    assert( mapFormat.standardMetadata.find( objectIter->id ) != mapFormat.standardMetadata.end() );
+                    mapFormat.standardMetadata.erase( objectIter->id );
+
+                    objectIter = mapTile.objects.erase( objectIter );
+                    needRedraw = true;
+                }
                 else {
                     objectIter = mapTile.objects.erase( objectIter );
                     needRedraw = true;
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -856,7 +856,7 @@ namespace Interface
                     }
 
                     std::string str = _( "Set %{monster} Count" );
-                    StringReplace( str, "%{monster}", Monster( object.index + 1 ).GetName() );
+                    StringReplace( str, "%{monster}", Monster( static_cast<int>( object.index ) + 1 ).GetName() );
 
                     fheroes2::ActionCreator action( _historyManager, _mapFormat );
                     if ( Dialog::SelectCount( str, 0, 500000, monsterCount ) ) {
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -846,6 +846,20 @@ namespace Interface
                         action.commit();
                     }
                 }
+                else if ( objectType == MP2::OBJ_MONSTER ) {
+                    uint32_t monsterCount = 0;
+
+                    auto monsterMetadata = _mapFormat.standardMetadata.find( object.id );
+                    if ( monsterMetadata != _mapFormat.standardMetadata.end() ) {
+                        monsterCount = monsterMetadata->second.metadata[0];
+                    }
+
+                    fheroes2::ActionCreator action( _historyManager, _mapFormat );
+                    if ( Dialog::SelectCount( _( "Set Monster Count" ), 0, 500000, monsterCount ) ) {
+                        _mapFormat.standardMetadata[object.id] = { monsterCount, Monster::JOIN_CONDITION_UNSET, 0 };
+                        action.commit();
+                    }
+                }
             }
         }
         else if ( _editorPanel.isTerrainEdit() ) {
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -860,7 +860,7 @@ namespace Interface
 
                     fheroes2::ActionCreator action( _historyManager, _mapFormat );
                     if ( Dialog::SelectCount( str, 0, 500000, monsterCount ) ) {
-                        _mapFormat.standardMetadata[object.id] = { monsterCount, Monster::JOIN_CONDITION_UNSET, 0 };
+                        _mapFormat.standardMetadata[object.id] = { static_cast<int32_t>( monsterCount ), Monster::JOIN_CONDITION_UNSET, 0 };
                         action.commit();
                     }
                 }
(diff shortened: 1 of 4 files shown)


Stisen1:
The dialog-box works as intended. The minimum number 0 is a wierd one though...

https://github.com/ihhub/fheroes2/assets/118435605/60ef5eaa-5ca7-4949-a041-3b2711540d2b

https://github.com/ihhub/fheroes2/assets/118435605/ce3cc05b-def7-4d4d-8188-3c9a50ea35fd

1. The dialog saves the inserted number of paladins. It's there when you open the dialog box again. 
2. When I save the map and run it, the standard amount of paladins is what you get. 
3. I've tried different values: same result. 

Furthermore: The new code come up with new issues, that I'm sure you've already noticed and get fixed down the line. 
 
image

Stisen1:
@

> @Stisen1.
> 
> > Furthermore: The new code come up with new issues, that I'm sure you've already noticed and get fixed down the line.
> 
> These issues are on your side. As can be seen by the build checks, the changes compile just fine. I've also tested locally myself. Please make sure you are making a clean build and that you are checking out the correct branch locally with GitHub.

@zenseii

I refreshed my local directory with a new one. Now VS has no issue with the changed codes in the editor_interface.cpp file. I get the exact same results as the abovementioned numbered list.

You can delete your comment, because I erased the thing that you commented on.

src/fheroes2/monster/monster.h
```diff
@@ -38,7 +38,8 @@ class Monster
     {
         JOIN_CONDITION_SKIP = 0,
         JOIN_CONDITION_MONEY = 1,
-        JOIN_CONDITION_FREE = 2
+        JOIN_CONDITION_FREE = 2,
+        JOIN_CONDITION_UNSET = 3
```
ihhub, line 42: Let's make it as **-1** to make sure that all negative values represent unset states.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -846,6 +847,20 @@ namespace Interface
...
+                else if ( objectType == MP2::OBJ_MONSTER ) {
+                    uint32_t monsterCount = 0;
+
+                    auto monsterMetadata = _mapFormat.standardMetadata.find( object.id );
+                    if ( monsterMetadata != _mapFormat.standardMetadata.end() ) {
+                        monsterCount = monsterMetadata->second.metadata[0];
+                    }
+
+                    fheroes2::ActionCreator action( _historyManager, _mapFormat );
+                    if ( Dialog::SelectCount( _( "Set Monster Count" ), 0, 500000, monsterCount ) ) {
```
ihhub, line 859: I am suggesting to use monster name here as it is not easy to identify which monster is being modified.

ihhub (changes requested):
Hi @idshibanov , I left 2 comments here. Can you please take a look at them?

ihhub:
Hi @Branikolog , please review it as the highest priority.

src/fheroes2/monster/monster.h
idshibanov: We store this value as uint in metadata, might trigger overflow warnings.

src/fheroes2/editor/editor_interface.cpp
idshibanov: image
Would have to check translations if they go over the dialog box limit.

Districh-ru: Since it is a temporary solution it is OK.
Later we will make a new monster select dialog. And also we can update this "SelectCount" dialog to allow multiple header lines.

Districh-ru (approved):

src/fheroes2/monster/monster.h
ihhub: I suggest to still change it to -1 and modify metadata type from `uint32_t` to `int32_`. I don't think that there is a need to even go beyond 2 billion values.

Branikolog:
Hello!
I've made a couple of tests.
As mentioned by @Stisen1 , when starting a map in single player mode creature numbers don't correspond those I've set in editor.

https://github.com/ihhub/fheroes2/assets/55348946/1d98054b-ce1f-47ee-87f6-f92841bfcc5c

I've also experienced a case, when new monsters I place have non zero value.
This happens after I load the map, delete the existing monster and trying to place other:

https://github.com/ihhub/fheroes2/assets/55348946/f3bfa213-07a5-4f99-9c6f-97d294943006

Branikolog:
I find it a bit weird too, when I see monster number as 0. If we consider this a default monster number, maybe it's better to use empty field or any non-numeric symbols (like "-" or so).

I can't adjust random monster numbers:
image

src/fheroes2/editor/editor_interface.cpp
Branikolog: Russian translation has few:
image

idshibanov:
> I find it a bit weird too, when I see monster number as 0. If we consider this a default monster number, maybe it's better to use empty field or any non-numeric symbols (like "-" or so).

That's the OG behaviour that I kept. To set a random amount you put a 0 there as well. It also represents exactly what's in the data, but I can see how it can confuse users.

> I've also experienced a case, when new monsters I place have non zero value.
> I can't adjust random monster numbers
> Creature numbers don't correspond those I've set in editor

Should be fixed now.

Branikolog:
Hi, @idshibanov 

>That's the OG behaviour that I kept. To set a random amount you put a 0 there as well. It also represents exactly what's in the data, but I can see how it can confuse users.

Indeed, but in the original game there's a hint explaining what "0" means.

image

Branikolog (approved):
If we're going to come back with different dialog later, we can come up with another implementation for that "0" indicator.

Stisen1:
How about doing the same thing as how you have 1-4 monsters CountasEnum = "few" in the monster.cpp? Something like: if number value === "0"; return whatever enum that correspond the default enum value for that class of monsters on a map's startup. So if you leave the number untouched, the box says "few | several | etc ". And when you close the dialog box, depending on input or left untouched, the number of monsters displayed will be either default enum or the one corresponding to input size... But uhm... You can't inspect that, if right-click makes you open the box.

Branikolog:
Hi, @Stisen1 !

> How about doing the same thing as how you have 1-4 monsters CountasEnum = "few" in the monster.cpp? Something like: if number value === "0"; return whatever enum that correspond the default enum value for that class of monsters on a map's startup. So if you leave the number untouched, the box says "few | several | etc ". And when you close the dialog box, depending on input or left untouched, the number of monsters displayed will be either default enum or the one corresponding to input size... But uhm... You can't inspect that, if right-click makes you open the box.

But how "Random Monster" object is going to handled then? 

Showing the number of creatures on right inspecting click seems like a convenient improvement for me, though.

Stisen1:
Hi @Branikolog

> Showing the number of creatures on right inspecting click seems like a convenient improvement for me, though

Just make a bigger dialog box for random monsters, that show proportional enum values for each random monster tier.

ihhub:
@Stisen1 and @Branikolog , we will change this window later. Now we are targeting functionality.

ihhub:
@idshibanov thank you so much for this change!

PR #8540, opened 2024-04-04
Add the jailed hero details dialog in Editor

Districh-ru:
Relates to #6845
This dialog is based on Hero Details dialog and allows to set the jailed hero race.
hero

### DIFF
--- a/src/fheroes2/heroes/heroes.cpp
+++ b/src/fheroes2/heroes/heroes.cpp
@@ -608,8 +608,14 @@ void Heroes::LoadFromMP2( const int32_t mapIndex, const int colorType, const int
     DEBUG_LOG( DBG_GAME, DBG_INFO, name << ", color: " << Color::String( GetColor() ) << ", race: " << Race::String( _race ) )
 }
 
-void Heroes::applyHeroMetadata( const Maps::Map_Format::HeroMetadata & heroMetadata, const int raceType, const bool isEditor )
+void Heroes::applyHeroMetadata( const Maps::Map_Format::HeroMetadata & heroMetadata, const int raceType, const bool isInJail, const bool isEditor )
 {
+    modes = 0;
+
+    if ( isInJail ) {
+        SetModes( JAIL );
+    }
+
     if ( _race != raceType ) {
         SetModes( CUSTOM );
 
@@ -838,6 +844,10 @@ Maps::Map_Format::HeroMetadata Heroes::getHeroMetadata() const
 
     // Hero's spell points.
     heroMetadata.magicPoints = ( magic_point == UINT32_MAX ) ? static_cast<int16_t>( -1 ) : static_cast<int16_t>( magic_point );
+
+    if ( Modes( JAIL ) ) {
+        heroMetadata.jailedHeroRace = static_cast<uint8_t>( _race );
+    }
 
     return heroMetadata;
 }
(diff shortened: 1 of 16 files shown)


src/fheroes2/maps/map_object_info.h
```diff
@@ -158,6 +158,8 @@ namespace Maps
 
     const std::vector<ObjectInfo> & getObjectsByGroup( const ObjectGroup group );
 
+    const Maps::ObjectInfo & getObjectInfo( const Maps::ObjectGroup group, const int32_t objectType );
```
ihhub, line 161: **objectType** should be **objectIndex**.

ihhub, line 161: Please remove `Maps` enumeration usage in this line as this is already under the same enumeration.

src/fheroes2/maps/map_format_info.h
```diff
@@ -133,17 +134,20 @@ namespace Maps::Map_Format
         // The amount of magic points (mana). Negative value means it is not set.
         int16_t magicPoints{ -1 };
 
+        // Jailed hero race, used only for jailed heroes. Race for normal heroes is stored in the ObjectInfo metadata.
+        uint8_t jailedHeroRace{ Race::RAND };
```
ihhub, line 138: Let's assign this value to 0 and do not use race here. Value 0 corresponds to NONE as a race which is even better to have to detect invalid values.

Also, I suggest to name this member as **race**. Simple as it sounds. Then we won't need an extensive comment for this. Moreover, we can also use this member to check that the hero's race is indeed the same as the sprites for them.

src/fheroes2/heroes/heroes.cpp
```diff
@@ -839,6 +845,10 @@ Maps::Map_Format::HeroMetadata Heroes::getHeroMetadata() const
     // Hero's spell points.
     heroMetadata.magicPoints = ( magic_point == UINT32_MAX ) ? static_cast<int16_t>( -1 ) : static_cast<int16_t>( magic_point );
 
+    if ( Modes( JAIL ) ) {
+        heroMetadata.jailedHeroRace = static_cast<uint8_t>( _race );
+    }
```
ihhub, line 850: As per proposal below we can simply do this:
```cpp
heroMetadata.race = static_cast<uint8_t>( _race );
```

ihhub (changes requested):
Hi @Districh-ru , I put several suggestions here. Would you mind to take a look at them?

src/fheroes2/maps/map_format_info.h
Districh-ru: Hi, @ihhub, sure!
I also had thoughts about using this variable to store the normal hero's race but left it unchanged.

src/fheroes2/maps/map_object_info.h
Districh-ru: OK)
I just moved this function without reviewing and changing it. :(

ihhub (approved):

Branikolog (approved):
Hi, @Districh-ru 
Everything works fine!
Just have a tiny question regarding faction name.
image

Not sure it is going to look better, but have you tried to place faction name a little lower? Currently it is placed just above the bottom line of background crown image. For me personally, it looks like the text underlining, which is not pleasant to eye. Could it look better, of the name is shifted ~5px lower?

Districh-ru:
Hi, @Branikolog 
> Could it look better, of the name is shifted ~5px lower?

Yes, it looks a little better. We can replace the background image later if we find a suitable one. :)

изображение

ihhub:
@Districh-ru , many thanks for this update!

PR #8559, opened 2024-04-09
Update Portuguese translation

altiereslima:


ihhub:
@altiereslima , huge thanks for this update!

Issue #7925, opened 2023-10-11
Creature separation window, add more vertical spacing between the small number buttons and the "OKAY" and "CANCEL" buttons


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

I suggest adding a little more vertical spacing between the small number buttons and the "OKAY" and "CANCEL" buttons.
A spacing of 17 pixels would be nice.

In fheroes2 (1.0.9 7923):

![Separation window 01](https://github.com/ihhub/fheroes2/assets/43583994/81a71a5f-21ac-4db0-b1ab-0bbfc063efdf)

Spacing in an older version of fheroes2 (0.6.3180):

![Separation window 02](https://github.com/ihhub/fheroes2/assets/43583994/81c5da8c-4985-42ae-b77f-997d2557162b)

### Describe the possible solution

Add more vertical spacing between the small number buttons and the "OKAY" and "CANCEL" buttons.
A spacing of 17 pixels would be nice.

ihhub:
This main issue will be fixed by #2388.

PR #8568, opened 2024-04-11
Creature separation window, add more vertical spacing between the small number buttons and the "OKAY" and "CANCEL" buttons

Cheerfulbull:
Made to address issue [#7925](https://github.com/ihhub/fheroes2/issues/7925). This increases the spacing between the Okay / Cancel buttons and the small numbers.
Screenshot from 2024-04-11 22-33-30
I understand that the editor is currently priority, so feel free to disregard this, but I thought this was a fairly easy issue, so might as well fix it.


Branikolog:
Hi, @Cheerfulbull !
I've checked the PR and have a few comments.
First of all, fast split icons highlighting is grey now. Is this change intentional? Previously, it had a red glowing effect.

I also feel, that the space between fast split icons and buttons line is a bit too large. I suggest making it the same, as between those split icons and the text above them.

Not sure, if it's related to the PR, but frame decoration looks not well in this area.
image

Branikolog:
Hi, @Mr-Bajs 

> If the window would be improved, then some more space between the arrows (up/down) and the Min/Max buttonw would be welcome because it's rather tight on a touch device.

Indeed. But I believe it's out of the scope of this PR.
I believe there could be created a separate set of issues, dedicated increasing the convenience using touch screens. We have dozens of places, where the original UI appears to be too small for touch devices. Some windows require even a complete rework in my opinion.

Cheerfulbull:
Hello @Branikolog , I have fixed the padding to make it more even. I was unable to replicate the visual glitch you sent. As for they grey highlighting (It's gold now for some reason?) I'm not sure what's causing that.

Cheerfulbull:
Also I don’t know why the MacOS check failed, seems to be a bug

ihhub:
> Also I don’t know why the MacOS check failed, seems to be a bug

It is fixed.

ihhub (approved):

ihhub:
Hi @Branikolog , please check these changes.

ihhub:
@Branikolog , now you will be able to download a build.

Branikolog:
Hi, @ihhub !

Yellow selection frame is still the case. It is supposed to be red.

Also, there're still tiny defects on a frame which is automatically stretched according to the required size.
image

I suppose it is a flaw of our frame-generating algorithm which cuts decoration on the frame.

ihhub:
Hi @Branikolog , I reverted back color changes.

@Cheerfulbull , can you please explain why you changed colors?

Cheerfulbull:
That was a mistake, I thought that was one of the positioning / window size functions.

Branikolog (approved):

PR #8569, opened 2024-04-11
Expand map format to support daily events

ihhub:
relates to #6845


ihhub:
@Branikolog , please test basic functionality of the Editor and playing maps. Use only newly created maps. You might need to delete old ones.

Districh-ru (approved):

PR #8571, opened 2024-04-12
Update Portuguese Translation

altiereslima:


altiereslima:
> Hi, @altiereslima. Could you take a look at button texts in the game? I see that many of the words are too long for one line and need to be adjusted. I'm many talking about the New Game menu. The other buttons don't matter since we will adjust the button backgrounds.

I'll take a look at it.

PR #8576, opened 2024-04-17
Expand logic for artifact placement in the Editor

ihhub:
- only one Random Ultimate artifact can be placed
- fix Spell Scroll artifact metadata
- each artifact now has its own metadata

relates to #6845


Districh-ru (approved):

Branikolog (approved):

ihhub:
Hi @Districh-ru , I also fixed stream and road placement and their updates. Now we don't move objects within the same map format tile.

PR #8577, opened 2024-04-19
Fix radar rendering for heroes and castles

idshibanov:
Relates to #6845 .

Castles and heroes must be synced up with `vec_heroes` and `vec_castles` world objects to be properly rendered on the map.


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1062,6 +1063,12 @@ namespace Interface
 
             setObjectOnTileAsAction( tile, groupType, _editorPanel.getSelectedObjectType() );
 
+            Heroes * hero = world.GetHeroForHire( Race::IndexToRace( static_cast<int>( objectInfo.metadata[1] ) ) );
+            if ( hero ) {
+                hero->SetCenter( tilePos );
+                hero->SetColor( static_cast<int>( color ) );
```
Districh-ru, line 1069: We should convert color index here:
```suggestion
                hero->SetColor( static_cast<int>( Color::IndexToColor( static_cast<int>( color ) ) ) );
```

Districh-ru (commented):
Hi, @idshibanov, I left one comment, could you please take a look when you have time?

Possibly in the future we may improve radar "for Editor" to use `Map_Format::MapFormat & map` instead of `world` to get objects' data for render.

Districh-ru (approved):

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -125,12 +126,32 @@ namespace Maps
...
+                else if ( object.group == Maps::ObjectGroup::KINGDOM_HEROES ) {
+                    const auto & metadata = heroObjects[object.index].metadata;
+                    const uint8_t color = Color::IndexToColor( static_cast<int>( metadata[0] ) );
+
+                    Heroes * hero = world.GetHeroForHire( static_cast<int>( metadata[1] ) );
+                    if ( hero ) {
+                        hero->SetCenter( { static_cast<int32_t>( i ) % world.w(), static_cast<int32_t>( i ) / world.w() } );
+                        hero->SetColor( color );
+                    }
+                }
```
ihhub, line 154: This change breaks map loading for gameplay. Such logic shouldn't exist here as this logic is within world_loadmap.cpp file.

ihhub: `map_format_helper.cpp` file should not contain any logic related to World itself. It should only help with loading information from the map format. The rest of logic is set in `world_loadmap.cpp`. We can move the proposed logic for the Editor only.

idshibanov: @ihhub Is there a place for MapFormat to be stored during gameplay? Not within editor files or loadmap. I don't see a way for Radar to access `MapFormat & map` currently.

ihhub (approved):

ihhub:
@idshibanov , thank you so much for the fix!

PR #8582, opened 2024-04-21
Allow river deltas to be placed anywhere

ihhub:
relates to #6845


Branikolog (approved):

Districh-ru (approved):
We need to update river placing logic to properly connect with deltas and do not overlap them. (in the new PR)

PR #8588, opened 2024-04-24
Update object placement rules in the Editor

ihhub:
- non-action objects can be put anywhere except on tiles with active action object parts
- active action object parts can be put only on tiles with clear ground

relates to #6845


ihhub:
The build is failing so far because of worldwide GitHub Action issues.

Branikolog:
Hi, @ihhub !
I've started testing and I've found a couple of rendering problems, that occur for object placement. 
image
image
image

ihhub:
Hi @Branikolog , objects are being placed on the map in order. I am sure if you put them in a different order the rendering is going to be different. Please verify it.

Branikolog:
Hi, @ihhub !
Indeed, the order I place objects effects the overall result. But in my opinion, we should prevent some apparent cases from being shown to a mapmaker.
When placing one mine over another in different order we see that 1 case is absolutely ineligible.
image

Maybe for the case we place objects on the same horizontal line there could be some variations depending on the object size, its type and the idea of map-maker, so we can leave the existing logic as is. But for the cases we place one object on a higher line I believe we shouldn't show its sprite overlapping the lower one.

I agree, that this could be fixed by proper order of object placement, but this is really inconvenient and forces map-maker doing multiple undesired actions, like deleting the existing object to place one nearby and then re-placing the previously deleted object.

ihhub:
@Branikolog , mines have top layer parts which will always overlap bottom layer parts.

An order of object placement is the most understandable for everyone while having some special rules for some objects is not going to solve anything. This is a map maker responsibility to make proper object overlapping the way how they want, not the way how we think it should be done.

ihhub:
> @ihhub I believe in some obvious cases we should try to make rendering a little more polished. image I hardly believe anyone would ever decide to place swamp over the mountains in such a horrible way. In a case of multiple objects on the same tile the overall result is absolutely unpredictable and looks not well. To avoid such picture map-maker has to re-build the entire area, which is not user-friendly.
> 
> Perhaps some kind of function could be implemented to allow map-maker to select the exact layer for the object he is about to place...

Hi @Branikolog , this is a problem of incorrect layer types for some objects which is an open task for #6845 - `update object layers, mostly for background layer`. Lake is a background object.

Branikolog:
Hi, @ihhub 
What about the cases, we place one object over another and some parts from horizontal lines are rendered on different level? Like that:
image

ihhub:
> Hi, @ihhub What about the cases, we place one object over another and some parts from horizontal lines are rendered on different level? Like that: image

Do you have a clear set of rules that can cover all cases for all possible objects? Can you explain how rendering should be done with the "new" logic for your highlighted scenario? Please remember the original game doesn't allow such cases for obvious reasons including rendering issues.

Branikolog:
@ihhub 
Actually, the original game allows to place objects one over another:
image

The thing, that bothers me is those two parts in the middle:
image
Logically the right part should be also rendered over the mountains.
I suggest considering other parts of the forest: if at least 1 part of the object is rendered over the mountains - then the whole forest object should be rendered over the mountains. It's literally what you was talking about a few days ago: If we place forest over the mountain - let it be rendered over it. If we place mountain over the forest - the forest should be hidden under the mountain object.

ihhub:
@Branikolog , every object can have bottom and top parts. Mountain top parts will always be rendered over bottom tree parts. We also agreed to change bottom and top parts for mountains to be consistent. As a result, I see no issues in rendering.

ihhub:
I am proceeding with merging these changes as no major flaws were found. We need to proceed further as this pull request review is delayed for far too long.

PR #8590, opened 2024-04-24
Fix some unused variable cases

oleg-derevenetz:
`numberOfColorsPresent` variable was introduced in #8348, but then the validation of the value of this variable was removed in #8560, so now this variable is only written and never read, which is detected by (newer?) compilers as unused variable and breaks the build in strict mode.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for the

PR #8591, opened 2024-04-25
AI should determine best skill when leveling up heroes

idshibanov:
Currently AI just rolls randomly when picking a new secondary skills. This decision should be driven by a value estimation, same as everything else.


ihhub:
Hi @idshibanov , please rebase your branch from master to fix GitHub Actions.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,81 @@ namespace
...
+        }
+        case Skill::Secondary::MYSTICISM:
+            return hero.HaveSpellBook() ? 500.0 : 100.0;
+        case Skill::Secondary::EAGLE_EYE:
+            return hero.HaveSpellBook() ? 250.0 : 0.0;
+        case Skill::Secondary::DIPLOMACY:
+            // discourage AI picking it up, but if it's already there prioritise leveling to save gold
+            return level == Skill::Level::BASIC ? 100.0 : 1250.0;
+        default:
+            // UNKNOWN skill can be a valid check
```
ihhub, line 942: Please add an assertion here to make sure that we don't hit it or when we add a new skill.

src/fheroes2/world/world.cpp
```diff
@@ -1278,6 +1278,16 @@ uint32_t World::CheckKingdomLoss( const Kingdom & kingdom ) const
...
 
+double World::getWaterPercentage() const
+{
+    return waterPercentage;
+}
+
+double World::getLandRoughness() const
+{
+    return mapRoughness;
+}
```
ihhub, line 1289: We can move the implementation of these methods to the header file. A compiler will inline them.

src/fheroes2/world/world.cpp
```diff
@@ -1278,6 +1278,16 @@ uint32_t World::CheckKingdomLoss( const Kingdom & kingdom ) const
     return GameOver::COND_NONE;
 }
 
+double World::getWaterPercentage() const
+{
+    return waterPercentage;
```
ihhub, line 1283: If this member is named as **percentage** then its value should be roughly from 0 to 100. Right now it is a normalized value from 0 to 1.

src/fheroes2/world/world_regions.cpp
```diff
@@ -208,24 +214,31 @@ void World::ComputeStaticAnalysis()
...
                 // else then ground is an obstacle for water navigation
                 ++obstacles[0][x].second;
                 ++obstacles[1][y].second;
             }
         }
     }
 
+    const double passableTileCount = ( width * height ) - obstacleCount;
+    waterPercentage = waterCount / passableTileCount;
+    mapRoughness = terrainPenalty / ( ( passableTileCount - waterCount ) * Maps::Ground::defaultGroundPenalty );
```
ihhub, line 240: I would suggest to at least add an assertion that **passableTileCount** is more than 0. It is a very tiny chance of this but just in case.

ihhub, line 240: **( passableTileCount - waterCount )** can be zero on maps with only water.

src/fheroes2/world/world_regions.cpp
```diff
@@ -208,24 +214,31 @@ void World::ComputeStaticAnalysis()
...
+                terrainPenalty += Maps::Ground::GetPenalty( tile, 0 );
                 // else then ground is an obstacle for water navigation
                 ++obstacles[0][x].second;
                 ++obstacles[1][y].second;
             }
         }
     }
 
+    const double passableTileCount = ( width * height ) - obstacleCount;
+    waterPercentage = waterCount / passableTileCount;
```
ihhub, line 239: Since we want to have percentage and this value shouldn't be extremely precise we can have this variable as **uint8_t**.

src/fheroes2/world/world.cpp
```diff
@@ -1278,6 +1278,16 @@ uint32_t World::CheckKingdomLoss( const Kingdom & kingdom ) const
...
+}
+
+double World::getLandRoughness() const
+{
+    return mapRoughness;
```
ihhub, line 1288: The name of method is get**Land**Roughness while the member is called **map**Roughness. I suggest to be consistent.

ihhub (changes requested):
Hi @idshibanov , I put several comments here. Could you please take a look at them when you have time?

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
         return fogDiscoveryBaseValue;
     }
 
+    double getSecondarySkillValue( const Heroes & hero, const Skill::Secondary & skill )
```
ihhub, line 873: Since all values returned by this function positive as well as they are not actually floating values I suggest to return **uint32_t** value which simplifies calculations and comparison of values.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2485,6 +2563,29 @@ namespace AI
...
+        }
+        DEBUG_LOG( DBG_AI, DBG_TRACE,
+                   hero.GetName() << " picking a skill: " << getSecondarySkillValue( hero, left ) << " for " << left.String( left.first ) << " and "
+                                  << getSecondarySkillValue( hero, right ) << " for " << right.String( right.first ) )
+
+        const double leftValue = getSecondarySkillValue( hero, left );
+        const double rightValue = getSecondarySkillValue( hero, right );
+
+        if ( std::fabs( leftValue - rightValue ) < 0.001 ) {
+            if ( leftValue < 300.0 ) {
```
ihhub, line 2581: Magic number! :)

A small comment explaining why this  value was chosen would be great to have.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2485,6 +2563,29 @@ namespace AI
...
+        if ( !right.isValid() ) {
+            // if both left and right are invalid returning either is fine
+            return left;
+        }
+        DEBUG_LOG( DBG_AI, DBG_TRACE,
+                   hero.GetName() << " picking a skill: " << getSecondarySkillValue( hero, left ) << " for " << left.String( left.first ) << " and "
+                                  << getSecondarySkillValue( hero, right ) << " for " << right.String( right.first ) )
+
+        const double leftValue = getSecondarySkillValue( hero, left );
+        const double rightValue = getSecondarySkillValue( hero, right );
```
ihhub, line 2578: I suggest to put these variables before DEBUG_LOG and use them for logging.

ihhub (changes requested):
Hi @idshibanov , I left few more small comments here. Would you mind to take a look once you have spare time?

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: It has to stay double so we can reuse it later in `getObjectValue` calls for the witch hut/etc. Values aren't that arbitrary; they were based on other estimates there.

crypt47:
Hello, I've always wanted to ask if Mysticism has any value. It looks like it has pretty much scores in descision making code, but it only gives you something like 4 sp/day at max. This is pretty low, especially for magic heroes who may use 100 sp and more during one battle.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2485,6 +2563,30 @@ namespace AI
...
     }
 
+    Skill::Secondary Normal::pickSecondarySkill( const Heroes & hero, const Skill::Secondary & left, const Skill::Secondary & right )
+    {
+        // heroes can get 1 or 0 skill choices depending on a level
+        if ( !right.isValid() ) {
+            // if both left and right are invalid returning either is fine
+            return left;
+        }
+        const double leftValue = getSecondarySkillValue( hero, left );
```
ihhub, line 2573: Please add an empty line after if-body for better code readability.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
         return fogDiscoveryBaseValue;
     }
 
+    double getSecondarySkillValue( const Heroes & hero, const Skill::Secondary & skill )
+    {
+        const int type = skill.Skill();
```
ihhub, line 875: Please add an assertion to verify that **skill** is valid.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2485,6 +2563,30 @@ namespace AI
...
+                                  << right.String( right.first ) )
+
+        if ( std::fabs( leftValue - rightValue ) < 0.001 ) {
+            // If skill value is lower than the threshold then it's undesireable. Avoid learning it.
+            if ( leftValue < 300.0 ) {
+                return left.Level() == Skill::Level::BASIC ? right : left;
+            }
+            return left.Level() == Skill::Level::BASIC ? left : right;
+        }
+        return leftValue > rightValue ? left : right;
```
ihhub, line 2587: Please add an empty line after if-body for better code readability.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -2485,6 +2563,30 @@ namespace AI
...
+        DEBUG_LOG( DBG_AI, DBG_TRACE,
+                   hero.GetName() << " picking a skill: " << leftValue << " for " << left.String( left.first ) << " and " << rightValue << " for "
+                                  << right.String( right.first ) )
+
+        if ( std::fabs( leftValue - rightValue ) < 0.001 ) {
+            // If skill value is lower than the threshold then it's undesireable. Avoid learning it.
+            if ( leftValue < 300.0 ) {
+                return left.Level() == Skill::Level::BASIC ? right : left;
+            }
+            return left.Level() == Skill::Level::BASIC ? left : right;
```
ihhub, line 2585: This logic is not entirely accurate (assuming that we don't know how values are being calculated in **getSecondarySkillValue()**:
```cpp
if ( leftValue < 300.0 ) {
    return left.Level() == Skill::Level::BASIC ? right : left; // <--- here looks okay since right and left can be both of BASIC level
}

return left.Level() == Skill::Level::BASIC ? left : right; // <-- here it NOT okay since right can have higher level than left
```
Also, we should consider randomness of a choice. Right now we are fixed on one of them.

src/fheroes2/world/world.h
```diff
@@ -410,6 +420,8 @@ class World : protected fheroes2::Size
     std::map<uint8_t, Maps::Indexes> _allTeleports; // All indexes of tiles that contain stone liths of a certain type (sprite index)
     std::map<uint8_t, Maps::Indexes> _allWhirlpools; // All indexes of tiles that contain a certain part (sprite index) of the whirlpool
 
+    uint8_t waterPercentage = 0;
+    double landRoughness = 1.0;
```
ihhub, line 424: Let's follow new members naming:
```cpp
uint8_t _waterPercentage{ 0 };
double _landRoughness{ 1.0 };
```

src/fheroes2/world/world_regions.cpp
```diff
@@ -208,24 +215,35 @@ void World::ComputeStaticAnalysis()
...
             }
         }
     }
 
+    const int passableTileCount = ( width * height ) - obstacleCount;
+    assert( passableTileCount > 0 );
+
+    waterPercentage = static_cast<uint8_t>( waterCount * 100 / passableTileCount );
+
+    const int landTiles = passableTileCount - waterCount;
```
ihhub, line 244: This value must always be greater than 0. We can add an assertion to check this.

src/fheroes2/world/world_regions.cpp
```diff
@@ -200,6 +203,10 @@ void World::ComputeStaticAnalysis()
         obstacles[3].emplace_back( y, 0 ); // ground, rows
     }
 
+    int obstacleCount = 0;
+    int waterCount = 0;
+    double terrainPenalty = 0;
```
ihhub, line 208: Use **uint32_t** for this variable for faster calculation in the loop. Then at the very end we can convert it to double.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
         return fogDiscoveryBaseValue;
     }
 
+    double getSecondarySkillValue( const Heroes & hero, const Skill::Secondary & skill )
+    {
+        const int type = skill.Skill();
+        const int level = skill.Level();
+        if ( hero.GetLevelSkill( skill.first ) >= level ) {
```
ihhub, line 877: Please use **type** here.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
...
+        case Skill::Secondary::ESTATES: {
+            const Heroes::Role role = hero.getAIRole();
+            if ( role == Heroes::Role::CHAMPION || role == Heroes::Role::FIGHTER ) {
+                return 0.0;
+            }
+            return 1000.0;
+        }
+        case Skill::Secondary::PATHFINDING: {
+            const double roughness = world.getLandRoughness();
+            return ( roughness > 1.25 ) ? 1000.0 : ( roughness > 1.1 ) ? 250.0 : 100.0;
```
ihhub, line 921: What if the value is lower than 1?

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
...
+        }
+        case Skill::Secondary::MYSTICISM:
+            return hero.HaveSpellBook() ? 500.0 : 100.0;
+        case Skill::Secondary::EAGLE_EYE:
+            return hero.HaveSpellBook() ? 250.0 : 0.0;
+        case Skill::Secondary::DIPLOMACY:
+            // discourage AI picking it up, but if it's already there prioritise leveling to save gold
+            return level == Skill::Level::BASIC ? 100.0 : 1250.0;
+        case Skill::Secondary::UNKNOWN:
+            return 0;
```
ihhub, line 942: This should not happen, please remove such a case and it should be handled by **default** case.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
...
+        }
+        case Skill::Secondary::ESTATES: {
+            const Heroes::Role role = hero.getAIRole();
+            if ( role == Heroes::Role::CHAMPION || role == Heroes::Role::FIGHTER ) {
+                return 0.0;
+            }
+            return 1000.0;
+        }
+        case Skill::Secondary::PATHFINDING: {
+            const double roughness = world.getLandRoughness();
```
ihhub, line 920: This skill value should depend on AI role as well.

src/fheroes2/ai/normal/ai_normal_hero.cpp
```diff
@@ -870,6 +870,84 @@ namespace
...
+        const int type = skill.Skill();
+        const int level = skill.Level();
+        if ( hero.GetLevelSkill( skill.first ) >= level ) {
+            return 0;
+        }
+
+        switch ( type ) {
+        case Skill::Secondary::WISDOM:
+            // wouldn't check castles/spell availability since high wisdom drives building priority
+            return level == Skill::Level::BASIC ? 2500.0 : 1000.0;
```
ihhub, line 884: This skill value should depend on AI role as well. Couriers don't care about Wisdom so much. The same for many skills below.

ihhub (changes requested):
Hi @idshibanov , I left several comments here. Would you mind please to take a look at them?

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: No, this is a valid case in case a hero only get a single skill choice.

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: I don't see how AI role influences this. If map has a lot of swamp and desert ANY hero should have this skill.

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: For example, Town Portal/Dimension Door would help out a Courier a lot. Hero development is a long-term investment. Also, a courier might change role later on. Bigger spellbook always helps.

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: Then this skill has minimal value; a 100. In most cases it's undesirable, don't see an issue with that.

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: Don't think it's needed here. Skill choice might be invalid.

src/fheroes2/ai/normal/ai_normal_hero.cpp
idshibanov: In the second case it's a valuable skill. Basic level then is more desirable than Advanced or Expert in order to secure the choice. I think the logic is accurate.

oleg-derevenetz:
Hi @idshibanov I rebased this PR against `master` to resolve conflicts after #8986, could you please take a look, just in case?

oleg-derevenetz (approved):

ihhub (approved):

ihhub:

PR #8597, opened 2024-05-02
Update layer type for several objects

ihhub:
Background layer is designed for objects that are a part of background terrain and these objects must be rendered first.

relates to #6845

This pull request also fixes cases described [here](https://github.com/ihhub/fheroes2/pull/8588#issuecomment-2077260122).

### DIFF
--- a/src/fheroes2/maps/map_object_info.cpp
+++ b/src/fheroes2/maps/map_object_info.cpp
@@ -1908,8 +1908,8 @@ namespace
         for ( uint32_t count = 0; count < 2; ++count ) {
             const uint32_t offset = 188U + count * 2U;
             Maps::ObjectInfo object{ MP2::OBJ_NOTHING_SPECIAL };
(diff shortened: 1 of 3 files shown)


ihhub:
Hi @Branikolog , I made 3 objects on the left and lava pool as background objects. The lake cannot be a background object as it has top layer image parts. The green object seems not be just a background object as it definitely has some shape. The tiny magma like object is a volcano which is not a background object.

Branikolog:
Hello, @ihhub !
I've made such conclusion, because other similar objects behave like that. All other lakes in the editor belong to background object group. This was the only lake, that was placed over the mountain object.
Magma object is also looks like other terrain objects, that belong to the background.
If this just a matter of the original resources, then we can proceed with merging of this PR. But I think it should be fixed in future in a separate PR.
image

ihhub:
Hi @Branikolog , we have same behavior as in the OG.

Branikolog (approved):
Good!

PR #8599, opened 2024-05-02
Add Editor icon to the Main Menu

ihhub:
relates to #6845

image

this pull request is going to be merged after all major tasks being completed.


Branikolog:
Hi, @ihhub !
I've found several issues here. Could you, please, try to fix them?
image

When I click over the new icon a selection frame appears for a while. The lines of this frame have different width. Also, in the original game those frames are made with a variety of colours, while the new one looks like made with a single yellow colour:
image

Also, there appears a small shifting (those nail heads seem to be cut a little) when I point new icon and press it (or open some other menu with a hotkey) 

https://github.com/ihhub/fheroes2/assets/55348946/9a42d8d2-90ea-4f78-aad9-70ca5630298a

ihhub:
Hi @Branikolog , shifting is observable for all buttons, not only the new one. It is a known issue and we aren't going to fix it right now. This is visible because you are using resolution non divisible by 640 x 480.

ihhub:
@Branikolog , the same applies for thick and non-thick lines: these are artifacts of image scaling.

ihhub:
@Branikolog , the first issue is fixed. I missed old resources to be included.

Branikolog:
@ihhub 

> @Branikolog , the same applies for thick and non-thick lines: these are artifacts of image scaling.

I see! I think my eye caught it because lines are rendered with brighter colours, unlike in the OG.

Branikolog (approved):
If we're not going to work on the frame colour now, then I have no objections on merging this PR. 
Everything else looks nice! 👍

Districh-ru (approved):
I like the editor icon!
I only concern - should we show it for the Demo and SW game assets (without PoL)?

Branikolog:
I think in future we could make the frame look more rough, to match the style of other icons.
ezgif com-resize (7)

ihhub:
Hi @Districh-ru , this is a good point. We shouldn't show it for no POL versions. I will add a code to handle this.

oleg-derevenetz (approved):

PR #8600, opened 2024-05-03
Random terrain generation

idshibanov:
Debug feature only. Basic editor usage only to assist with mapmaking. More to come soon™.


src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+#include "resource.h"
+#include "ui_map_object.h"
+#include "world.h"
+#include "world_object_uid.h"
+
+namespace Maps::Generator
+{
+    // ObjectInfo ObjctGroup based indicies do not match old objects
+    const int NEUTRAL_COLOR = 6;
+    const int RANDOM_CASTLE_INDEX = 12;
```
ihhub, line 56: Please use camelCase for variable naming.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+    const std::vector<int> playerStartingTerrain = { Ground::GRASS, Ground::DIRT, Ground::SNOW, Ground::LAVA, Ground::WASTELAND };
+    const std::vector<int> neutralTerrain = { Ground::GRASS, Ground::DIRT, Ground::SNOW, Ground::LAVA, Ground::WASTELAND, Ground::BEACH, Ground::SWAMP, Ground::DESERT };
```
ihhub, line 58: We can use `std::array<>` over std::vector. `std::array` allocates memory on stack which is faster.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+    std::vector<fheroes2::Point> directionOffsets = { { 0, -1 }, { 1, 0 }, { 0, 1 }, { -1, 0 }, { -1, -1 }, { 1, -1 }, { 1, 1 }, { -1, 1 } };
```
ihhub, line 60: This can be `std::array<>` as well. Plus it can be **const**.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+        Node & getNode( int32_t index )
+        {
+            return getNode( { index % mapSize, index / mapSize } );
+        }
+    };
+
+    struct Region
+    {
+    public:
```
ihhub, line 126: Structures by default use **public** section so here is no need to declare it.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+#include "maps_tiles_helper.h"
+#include "math_base.h"
+#include "race.h"
+#include "rand.h"
+#include "resource.h"
+#include "ui_map_object.h"
+#include "world.h"
+#include "world_object_uid.h"
+
+namespace Maps::Generator
```
ihhub, line 52: All the variables and functions must be put in anonymous namespace, except **generateWorld()** function. This will reduce the scope of all of them.

src/fheroes2/maps/map_generator.h
```diff
@@ -0,0 +1,37 @@
...
+namespace Maps::Generator
+{
+    struct Configuration
+    {
+        uint32_t playerCount = 2;
+        uint32_t regionSizeLimit = 300;
+        bool basicOnly = true;
+    };
+
+    bool generateWorld( Map_Format::MapFormat & mapFormat, Configuration config );
```
ihhub, line 36: **config** parameter must be passed by const reference as it is not a trivial type.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+        // generate road based paths
+        // place objects avoiding the borders
+        //
+        // make sure objects accessible before
+        // make sure paths are accessible - delete obstacles
+        // place treasures
+        // place monsters
+        return true;
+    }
+}
```
ihhub, line 528: Please add an empty line at the end of this file.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+    bool placeMine( Map_Format::MapFormat & mapFormat, NodeCache & data, Region & region, const int resource )
+    {
+        const auto & node = Rand::Get( region._nodes );
+        Maps::Tiles & mineTile = world.GetTiles( node.index );
+        const int32_t mineType = fheroes2::getMineObjectInfoId( resource, mineTile.GetGround() );
+        return actionObjectPlacer( mapFormat, data, mineTile, Maps::ObjectGroup::ADVENTURE_MINES, mineType );
+    }
+
+    bool generateWorld( Map_Format::MapFormat & mapFormat, Configuration config )
```
ihhub, line 369: Playing with multiple map generators I noticed one big flaw in all of them: they try to generate everything at once or at one iteration. Map generation is usually an iterative process where objects are placed incrementally. The current implementation might work well but then we are going to face the similar issue of imperfection of results.

What I am suggesting is to create a class / structure which has individual methods to placing objects or changing terrain. As of now we can group the steps specified in this function as separate methods.

In theory we should have something like this:
```
initialization();

while ( requiresAdjustment ) {
    adjustTerrain(); // <--- sometimes we need to slightly adjust terrain based on objects for passability

    modifyObjects(); // <--- we might move objects or place new ones or remove some of them

    reevaluate(); // <--- here we have to see that balancing is right and if it is not set adjustment flag to true
}
```

This is just a mock-up code, no need to follow it. I wrote it to explain my idea behind the generator.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+    {
+        if ( config.playerCount < 2 || config.playerCount > 6 ) {
```
ihhub, line 371: We should assert `config.playerCount <= 6` case as it is definitely a logical problem with the code.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+        OPEN,
+        BORDER,
+        ACTION,
+        OBSTACLE,
+        PATH
+    };
+
+    struct Node
+    {
+        int index = -1;
```
ihhub, line 73: Let's be consistent with the code:
- in all structures and classes please put methods first and then members later
- initialization is good to do in C++11 way. For example, `int index{ -1 };
- some members of structures have empty initialization value: `uint32_t _id{};`. It should be initialized with 0: `uint32_t _id{ 0 };`

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+        NodeType type = NodeType::OPEN;
+        uint32_t region = 0;
+        uint16_t mapObject = 0;
+        uint16_t passable = DIRECTION_ALL;
+
+        Node() = default;
+        explicit Node( int index_ )
+            : index( index_ )
+        {}
```
ihhub, line 82: Please add a comment inside this method:
```cpp
// Do nothing.
```
to make SonarQube happy.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+    };
+
+    class NodeCache
+    {
+        int32_t mapSize = 0;
```
ihhub, line 87: As per comment above please move these variables at the bottom of class and also add **private** section before them.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+            : mapSize( width )
+            , outOfBounds( -1 )
+            , data( width * height )
+        {
+            outOfBounds.type = NodeType::BORDER;
+
+            for ( int y = 0; y < height; ++y ) {
+                const int rowIndex = y * width;
+                for ( int x = 0; x < width; ++x ) {
+                    const int index = rowIndex + x;
```
ihhub, line 102: Just a minor adjustment: **height** and **width** are `int32_t`, not `int`. While on most of platforms it is true, sometimes it is not.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+        // Process only "open" nodes that exist at the start of the loop and ignore what's added
+        const size_t nodesEnd = region._nodes.size();
+
+        while ( region._lastProcessedNode < nodesEnd ) {
+            checkAdjacentTiles( rawData, region );
+            ++region._lastProcessedNode;
+        }
+    }
+
+    bool canFitObject( NodeCache & data, const ObjectInfo & info, const fheroes2::Point & mainTilePos, const bool isAction = false )
```
ihhub, line 203: I think this is the same function as **isObjectPlacementAllowed()** and **isActionObjectAllowed()** functions in editor_interface.cpp file. Do you think we can avoid code duplication for this?

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+        if ( isAction ) {
+            for ( int x = objectRect.x - 1; x <= objectRect.width + 1; x++ ) {
+                Node & pathNode = data.getNode( mainTilePos + fheroes2::Point{ x, objectRect.height + 1 } );
+                pathNode.type = NodeType::PATH;
+            }
+            data.getNode( mainTilePos ).type = NodeType::ACTION;
+        }
+    }
+
+    bool putObjectOnMap( Map_Format::MapFormat & mapFormat, Tiles & tile, const ObjectGroup groupType, const int32_t objectIndex )
```
ihhub, line 278: It is the same as **EditorInterface::_setObjectOnTile()** method. We should reuse the existing code.

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+        const fheroes2::Point tilePos = tile.GetCenter();
+        const auto & objectInfo = Maps::getObjectInfo( groupType, type );
+        if ( canFitObject( data, objectInfo, tilePos, true ) && putObjectOnMap( mapFormat, tile, groupType, static_cast<uint32_t>( type ) ) ) {
+            markObjectPlacement( data, objectInfo, tilePos, true );
+            return true;
+        }
+        return false;
+    }
+
+    bool placeCastle( Map_Format::MapFormat & mapFormat, NodeCache & data, Region & region, int targetX, int targetY )
```
ihhub, line 308: **placeCastle** and **placeMine** have the same code as `editor_interface.cpp` file. In the future we might even change the code there. We need to avoid using code duplication.

We can move existing functions from `editor_interface.cpp` file to a new file. We can call it as `editor_map_helper.h|cpp` and use the functions here and in `editor_interface.cpp` file

src/fheroes2/maps/map_generator.cpp
```diff
@@ -0,0 +1,528 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    bool placeMine( Map_Format::MapFormat & mapFormat, NodeCache & data, Region & region, const int resource )
+    {
+        const auto & node = Rand::Get( region._nodes );
+        Maps::Tiles & mineTile = world.GetTiles( node.index );
+        const int32_t mineType = fheroes2::getMineObjectInfoId( resource, mineTile.GetGround() );
```
ihhub, line 365: We should use **ObjectInfo** from `mapFormat` which already contains mineType. See **EditorPanel::getMineObjectProperties()** for this matter.

src/fheroes2/editor/editor_interface.h
```diff
@@ -153,6 +153,11 @@ namespace Interface
         int32_t _selectedTile{ -1 };
         int32_t _tileUnderCursor{ -1 };
 
+#if defined( WITH_DEBUG )
+        uint32_t _playerCount = 2;
```
ihhub, line 157: Should we store **Configuration** object instead of these 2 members?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -770,6 +771,34 @@ namespace Interface
...
+                    else {
+                        _warningMessage.reset( _( "Not able to generate a map with given parameters." ) );
+                    }
+                }
+                else if ( HotKeyPressEvent( Game::HotKeyEvent::EDITOR_RANDOM_MAP_CONFIGURATION ) ) {
+                    uint32_t newCount = _playerCount;
+                    if ( Dialog::SelectCount( "Pick player count", 2, 6, newCount ) ) {
+                        _playerCount = newCount;
+                    }
+                    newCount = _regionSizeLimit;
```
ihhub, line 796: I am not sure that region size is an understandable entity for map makers. Can we come up with something more user friendly metric? :)

ihhub (changes requested):
Hi @idshibanov , I left multiple comments and suggestions here. Could you please take a look at them once you have time?

src/fheroes2/maps/map_generator.cpp
idshibanov: Later on I use it to select a random terrain type. Rand::Get takes a only a vector.

src/fheroes2/maps/map_generator.cpp
idshibanov: For some reason array doesn't play nicely with Point class allocation. I get `no suitable constructor exists to convert from "int" to "fheroes2::PointBase2D<int32_t>"`

src/fheroes2/maps/map_generator.cpp
idshibanov: Few commits back I did use isObjectPlacementAllowed() and isActionObjectAllowed() functions but now `canFitObject` operates on generator cache rather than map format struct. Many rules that are checked by the editor (like passability) can be ignored during the generation.

I think the only reuse possible here is to make a method to extract objectParts from ObjectInfo to iterate through, but we still need two separate calls for editor and generator.

src/fheroes2/maps/map_generator.cpp
idshibanov: Similar situation as in previous comment. For example this call doesn't recalculate passabilities after every call.

src/fheroes2/maps/map_generator.cpp
idshibanov: I don't see how `getMineObjectProperties` would help here. I'm going in opposite direction - not having ObjectInfo I need to find out which index contains what I need. ObjectInfo groups have arbitrary order and no enums to support it, so I had to make this function.

src/fheroes2/editor/editor_interface.cpp
idshibanov: I'm fine with renaming it but can't think of more suitable name.

src/fheroes2/maps/map_generator.cpp
idshibanov: I agree that this is the worst case of duplication, but can't think of a clean way to extract it. I'd have to pass 2-3 function callbacks into it for object placement check, action object check an additional marking call for the generator. Editor version also have to handle error messages for the user feedback but generator just tries another time.

src/fheroes2/maps/map_generator.cpp
idshibanov: I'm not sure if it's worth exposing things like adjustTerrain/modifyObjects/reevaluate. We'd never need to call those outside of generation process. They will probably stay as private methods with a single public call for generateWorld.

I agree with iteration principle overall but to nitpick I'd prefer to avoid terrain adjustments/moving objects as it might create more problems rather than solve. Potentially I was thinking to mark object most placements before "committing" them to actually place on the map, that's why a separate NodeCache exists. It will be easier to walk back on the decision rather than dealing with map format.

ihhub: My point was to still keep only one function **generateWorld** while the logic inside it can be any format.

ihhub (changes requested):
Hi @idshibanov , I left few more comments here. Also, some comments are still valid form the previous review. Could you please take a look at them?

src/fheroes2/maps/map_random_generator.h
```diff
@@ -0,0 +1,40 @@
...
+{
+    struct MapFormat;
+}
+
+namespace Maps::Generator
+{
+    struct Configuration
+    {
+        uint32_t playerCount = 2;
+        uint32_t regionSizeLimit = 300;
```
ihhub, line 35: Please make these 2 parameters as `int32_t` so we can remove a lot of `static_cast` code.

src/fheroes2/maps/map_random_generator.cpp
```diff
@@ -0,0 +1,541 @@
...
+            // Fix missing references
+            for ( const uint32_t adjacent : region._neighbours ) {
+                mapRegions[adjacent]._neighbours.insert( region._id );
+            }
+        }
+
+        // Step 5. Object placement
+        for ( Region & region : mapRegions ) {
+            if ( region._id == 0 )
+                continue;
```
ihhub, line 482: Please add parenthesis for this if-block like it is done for the rest of the code.

src/fheroes2/maps/map_random_generator.cpp
```diff
@@ -0,0 +1,541 @@
...
+        Maps::Tile & mineTile = world.getTile( node.index );
+        const int32_t mineType = fheroes2::getMineObjectInfoId( resource, mineTile.GetGround() );
+        return actionObjectPlacer( mapFormat, data, mineTile, Maps::ObjectGroup::ADVENTURE_MINES, mineType );
+    }
+}
+
+namespace Maps::Generator
+{
+
+    bool generateMap( Map_Format::MapFormat & mapFormat, const Configuration & config )
```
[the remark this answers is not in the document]
ihhub: We should return `Map_Format::MapFormat`. If it is empty then something went wrong.

ihhub (changes requested):
Hi @idshibanov , I left several comments here as well as open questions. Could you please take a look at them in your spare time?

src/fheroes2/world/world_regions.cpp
```diff
@@ -354,7 +354,7 @@ void World::ComputeStaticAnalysis()
     bool stillRoomToExpand = true;
     while ( stillRoomToExpand ) {
         stillRoomToExpand = false;
-        for ( size_t regionID = REGION_NODE_FOUND; regionID < regionCenters.size(); ++regionID ) {
```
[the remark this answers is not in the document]
idshibanov: Yes

src/fheroes2/maps/map_random_generator.cpp
```diff
@@ -0,0 +1,541 @@
...
+            for ( const int resource : resoures ) {
+                // TODO: do a gradual distribution instead of guesses
+                for ( int tries = 0; tries < 5; tries++ ) {
+                    if ( placeMine( mapFormat, data, region, resource ) ) {
+                        break;
+                    }
+                }
+            }
+
+            Maps::updateRoadOnTile( world.getTile( region._centerIndex ), true );
```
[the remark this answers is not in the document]
ihhub: The work for this change has started in #9590 . The changes there are a hard requirement to proceed with this pull request.

ihhub:
Hi @idshibanov , I haven't re-reviewed changes but from my memory I think we should do the following:
- remove the code from being only debug
- Add a window with all the settings for random map selection in the Editor. It is absolutely fine if it doesn't generate everything yet. We can state that this is Work in Progress

I would like us to have a very basic version which can be used by map makers and to have feedback from them once we make the changes.

ihhub:
@idshibanov , could you also please make IWYU happy?🙂

idshibanov:
> Hi @idshibanov , I haven't re-reviewed changes but from my memory I think we should do the following:
> 
>     * remove the code from being only debug
> 
>     * Add a window with all the settings for random map selection in the Editor. It is absolutely fine if it doesn't generate everything yet. We can state that this is Work in Progress
> 
> 
> I would like us to have a very basic version which can be used by map makers and to have feedback from them once we make the changes.

I suggest making UI changes in a separate PR as it might spawn another discussion on how it should look like. Also somebody else would be able to pick it up.

ihhub:
@idshibanov , let's do what we can do: remove debug only and move the logic to random map selection.

idshibanov:
Hi @Districh-ru, please check my last 5 commits and let me know if there a better way to do the following:
- I resurrected `world.generateMapForEditor ` to reset the map between generations (or if you had another map open in the editor)
- Renamed setTerrain to setTerrainWithTransition and exposed one which does not handle the transition in order to put 1x1 tiles during gen
- If there's a new way to place castles instead of my function in map_random_generator.cpp

src/fheroes2/maps/map_random_generator.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    bool generateMap( Map_Format::MapFormat & mapFormat, const Configuration & config )
+    {
+        if ( config.playerCount < 2 || config.playerCount > 6 ) {
+            assert( config.playerCount <= 6 );
+            return false;
+        }
+        if ( config.regionSizeLimit < 200 ) {
+            return false;
+        }
```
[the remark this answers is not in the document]
idshibanov: I think my ideas was that generation could unexpectedly fail to satisfy the parameters even if they look reasonable. The caller (map editor) should handle the `false` returned gracefully and give the map maker an opportunity to retry.

Districh-ru:
Hi @idshibanov, I've pushed 3 commits to:
- enable "new random map" selection in editor's main menu;
- use `interface.generateNewMap()` instead of `world.generateMapForEditor()` to fully re-initialize map (previously the water was not fully initialized in `mapFormat` and after save/load or undo/redo it had bugs);
- add `interface.placeCastle()` method that is now used in editor and map generator (in future we should update it to operate with race, town/castle instead of type id).

Currently editor has no external methods to place objects but we can make them for the map generator if it is needed.
Note: The town placing in editor works slower than in map generator in Debug version.

The new ` Maps::setTerrainOnTile()` is now used in `generateNewMap()` method for filling map with water.

src/fheroes2/editor/editor_mainmenu.cpp
```diff
@@ -212,33 +226,27 @@ namespace Editor
...
-
-                    continue;
+                    outputEditorNewMapMenuInTextSupportMode();
                 }
-                if ( le.MouseClickLeft( buttonLoadMap.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::EDITOR_LOAD_MAP_MENU ) ) {
+                else if ( le.MouseClickLeft( buttonLoadMap.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::EDITOR_LOAD_MAP_MENU ) ) {
                     return fheroes2::GameMode::EDITOR_LOAD_MAP;
                 }
-                if ( le.MouseClickLeft( buttonMainMenu.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) {
+                else if ( le.MouseClickLeft( buttonMainMenu.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) {
```
ihhub: Please revert these changes. The original code was correct: there is no need to use `else` after **return**.

src/fheroes2/editor/editor_mainmenu.cpp
```diff
@@ -212,33 +226,27 @@ namespace Editor
...
                     return fheroes2::GameMode::MAIN_MENU;
                 }
 
-                if ( le.isMouseRightButtonPressedInArea( buttonNewMap.area() ) ) {
-                    // TODO: update this text once random map generator is ready.
-                    //       The original text should be "Create a new map, either from scratch or using the random map generator."
-                    fheroes2::showStandardTextMessage( _( "New Map" ), _( "Create a new map from scratch." ), Dialog::ZERO );
+                else if ( le.isMouseRightButtonPressedInArea( buttonNewMap.area() ) ) {
```
ihhub: Please remove `else` here.

ihhub:
Hi @zenseii , could you please review changes in strings? We also added a popup dialog telling the state of the feature and I feel it should be written better.

ihhub:
@idshibanov , thank you very much for this feature!

PR #8604, opened 2024-05-03
Add a multi-line text editing feature to edit Sign and Bottle texts and make some fixes to text rendering

Districh-ru:
This PR adds ability to input long phases in the multi-line input dialog. It adds ability to input Sign and Bottle text in Editor. It will be also used to input map description in the next PR.

https://github.com/ihhub/fheroes2/assets/113276641/46c71bda-3a35-4ec7-b305-dd1dafd9e8b4

This PR also fixes:
- incorrect cursor position calculation for texts with spaces;
- incorrect multi-line text width calculation for `UniformedVerticalAlignment` texts;
- incorrect cutting of words for multi-font text cases:
изображение **=>>** изображение

This PR also affects single-line text input dialog.


src/fheroes2/gui/ui_text.h
```diff
@@ -88,6 +90,10 @@ namespace fheroes2
 
     int32_t getFontHeight( const FontSize fontSize );
 
+    // Returns the character position number in the 'text' string.
```
ihhub, line 93: This function does **not** belong to this file. This file should contain only common classes and structures and not anything specific. So, the original place `ui_tool.h|cpp` was a valid place for this function.

We can expose **CharHandler** class within **ui_text.h** file to avoid having this function here.

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -228,59 +230,79 @@ bool Dialog::SelectCount( const std::string & header, uint32_t min, uint32_t max
...
     const fheroes2::Sprite & inputArea = fheroes2::AGG::GetICN( ( isEvilInterface ? ICN::BUYBUILD : ICN::BUYBUILE ), 3 );
 
-    const int32_t titleHeight = title.empty() ? 0 : titlebox.height( BOXAREA_WIDTH ) + 10;
-    const int32_t keyBoardButtonExtraHeight = 20;
-    const FrameBox box( 10 + titleHeight + textbox.height( BOXAREA_WIDTH ) + 10 + inputArea.height() + keyBoardButtonExtraHeight, true );
-    const fheroes2::Rect & box_rt = box.GetArea();
+    const int32_t inputAreaWidth = isMultiLine ? 224 : inputArea.width();
+    const int32_t inputAreaHeight = isMultiLine ? 265 : inputArea.height();
+
+    const int32_t frameBoxHeight = 10 + titleHeight + textbox.height( BOXAREA_WIDTH ) + 10 + inputAreaHeight + keyBoardButtonExtraHeight;
```
ihhub, line 272: We can cache `textbox.height( BOXAREA_WIDTH )` value.

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -315,41 +337,40 @@ bool Dialog::InputString( const std::string & header, std::string & res, const s
...
         else if ( le.MouseClickLeft( textInputArea ) ) {
-            charInsertPos = fheroes2::getTextInputCursorPosition( res, fheroes2::FontType::normalWhite(), charInsertPos, le.GetMouseCursor().x,
-                                                                  textInputArea.x + ( textInputArea.width - text.width() ) / 2 );
+            charInsertPos = text.getTextInputCursorPosition( charInsertPos, le.GetMouseCursor(), textInputArea );
 
             redraw = true;
         }
 
-        if ( le.MousePressRight( buttonCancel.area() ) ) {
+        else if ( le.MousePressRight( buttonCancel.area() ) ) {
```
ihhub, line 387: If we added `else` here then please remove an empty line above.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -22,87 +22,109 @@
...
+    const uint8_t cursorChar{ '_' };
+
+    class CharHandler
     {
     public:
-        explicit CharValidator( const fheroes2::FontSize fontSize )
-            : _charLimit( fheroes2::AGG::getCharacterLimit( fontSize ) )
+        explicit CharHandler( const fheroes2::FontType fontType )
+            : _fontType( fontType )
+            , _charLimit( fheroes2::AGG::getCharacterLimit( _fontType.size ) )
```
ihhub, line 47: I suggest to use `fontType.size` over `_fontType.size` as using class members already in the initialization list is risky. This code is of course absolutely valid but it's better to be safe.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -22,87 +22,109 @@
...
+            case fheroes2::FontSize::SMALL:
+                return 4;
+            case fheroes2::FontSize::NORMAL:
+                return 6;
+            case fheroes2::FontSize::LARGE:
+            case fheroes2::FontSize::BUTTON_RELEASED:
+            case fheroes2::FontSize::BUTTON_PRESSED:
+                return 8;
+            default:
+                assert( 0 ); // Did you add a new font size? Please add implementation.
```
ihhub, line 100: Please add **break** here as well.

src/fheroes2/gui/ui_text.h
```diff
@@ -169,10 +182,17 @@ namespace fheroes2
...
+            _text = std::move( text );
+            _fontType = fontType;
+        }
 
         // This method modifies the underlying text and ends it with '...' if it is longer than the provided width.
-        void fitToOneRow( const int32_t maxWidth );
+        void fitToOneRow( const int32_t maxWidth, const bool ignoreSacesAtEnd = true );
+
+        // Returns the character position number in the 'text' string. NOTICE: the input text should contain the cursor ('_').
+        size_t getTextInputCursorPosition( const size_t currentTextCursorPosition, const Point & pointerCursorOffset, const Rect & textRoi );
```
ihhub, line 195: This method should **not** exist for this class. This is a special case which is not related to text class itself. As of now I see no easy way to remove it so let's skip this comment and come back to it once we address other comments.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -734,6 +741,72 @@ namespace fheroes2
         _text += truncatedEnding;
     }
 
+    size_t Text::getTextInputCursorPosition( const size_t currentTextCursorPosition, const Point & pointerCursorOffset, const Rect & textRoi )
+    {
+        {
```
ihhub, line 746: Please remove one level of parenthesis.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -22,87 +22,109 @@
...
-        default:
-            assert( 0 ); // Did you add a new font size? Please add implementation.
+            return 0;
         }
 
-        return 0;
-    }
+        fheroes2::FontType _fontType;
+        uint32_t _charLimit;
+        int32_t _spaceCharWidth;
```
ihhub, line 108: These members can be const.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -22,87 +22,109 @@
...
+            , _charLimit( fheroes2::AGG::getCharacterLimit( _fontType.size ) )
+            , _spaceCharWidth( _getSpaceCharWidth() )
         {}
 
-        bool isValid( const uint8_t character ) const
+        // Returns true if character is available to render, including space (' ') and new line ('\n').
+        bool isAvailable( const uint8_t character ) const
         {
-            return character >= 0x21 && character <= _charLimit;
+            return ( character == spaceChar || _isValid( character ) || character == lineSeparator );
```
ihhub, line 54: Let's move **isSpaceChar** function declaration above this class and use it here and at line 64.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -113,121 +135,94 @@ namespace
...
             ++data;
         }
 
         return width;
     }
 
-    // Returns text lines parameters (in pixels) in 'offsets': x - represents the line widths, y - represents vertical line shift.
+    // If 'countCharacters' is true - returns text lines parameters (in characters) in 'offsets': x - the number of characters on the line, y - vertical line number.
+    // If 'countCharacters' is false - returns text lines parameters (in pixels) in 'offsets': x - line widths, y - vertical line shift.
     void getMultiRowInfo( const uint8_t * data, const int32_t size, const int32_t maxWidth, const fheroes2::FontType & fontType, const int32_t rowHeight,
```
ihhub, line 184: While code duplication is bad I think we should have 2 functions for this case rather than handling 2 different cases in one function. One function should behave exactly as it should be before. The other one should accept only `std::deque<int32_t>` which will contain the number of characters per line. We would have code duplication but it would be much easier to read the code. What do you think?

ihhub (changes requested):
Hi @Districh-ru , I left several comments here. Would you mind to take a look at them when you have spare time?

src/fheroes2/gui/ui_text.h
Districh-ru: I left a TODO to fix it in the future.

src/fheroes2/gui/ui_text.cpp
Districh-ru: I've updated this function to return both vectors.

src/fheroes2/gui/ui_text.h
```diff
@@ -88,6 +91,60 @@ namespace fheroes2
...
+        bool isAvailable( const uint8_t character ) const
+        {
+            return ( isSpaceChar( character ) || _isValid( character ) || isLineSeparator( character ) );
+        }
+
+        const fheroes2::Sprite & getSprite( const uint8_t character ) const
+        {
+            // Display '?' in place of the invalid character.
+            return fheroes2::AGG::getChar( _isValid( character ) ? character : '?', _fontType );
+        }
```
ihhub, line 125: The implementation for these methods should be in **ui_text.cpp** file as we are exposing too much for the logic for this class.  We are also adding extra dependencies in header files by using logic for these methods here.

Once we move these method's implementation **isLineSeparator**, **isSpaceChar** can be normal functions as before (instead of current `fheroes2::CharHandler::isSpaceChar()`  calls).

src/fheroes2/gui/ui_text.h
```diff
@@ -88,6 +91,60 @@ namespace fheroes2
...
+        {
+            return _spaceCharWidth;
+        }
+
+    private:
+        // Returns true if character is valid for the current code page, excluding space (' ') and new line ('\n').
+        bool _isValid( const uint8_t character ) const
+        {
+            return character >= 0x21 && character <= _charLimit;
+        }
```
ihhub, line 139: Please move the implementation to the cpp file.

src/fheroes2/gui/ui_text.h
```diff
@@ -88,6 +91,60 @@ namespace fheroes2
 
     int32_t getFontHeight( const FontSize fontSize );
 
+    class CharHandler
```
ihhub, line 94: Please move this class at the bottom of the file just before `bool isFontAvailable()` function since this class is used occasionally outside this file.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -113,121 +52,95 @@ namespace
...
         }
 
         return width;
     }
 
-    // Returns text lines parameters (in pixels) in 'offsets': x - represents the line widths, y - represents vertical line shift.
+    // Returns text lines parameters (in pixels) in 'offsets': x - line widths, y - vertical line shift. And in 'charsInLine' - the number of characters on the line.
     void getMultiRowInfo( const uint8_t * data, const int32_t size, const int32_t maxWidth, const fheroes2::FontType & fontType, const int32_t rowHeight,
-                          std::deque<fheroes2::Point> & offsets )
+                          std::deque<fheroes2::Point> & offsets, std::vector<int32_t> & charsInLine )
```
ihhub, line 101: I suggest to create a structure:
```cpp
struct TextLineInfo
{
    fheroes2::Point offset;

    int32_t characterCount{ 0 };
};
```

Then we won't need to handle 2 input containers. Right now these 2 input containers must match by size but we don't verify it and they are different types.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -113,121 +52,95 @@ namespace
...
     {
         assert( data != nullptr && size > 0 && maxWidth > 0 );
 
         if ( offsets.empty() ) {
             offsets.emplace_back();
```
ihhub, line 106: If we use proposed structure we can still have only one call to emplace back a new element.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -113,121 +52,95 @@ namespace
...
+        int32_t lineLength = charsInLine.back();
+        int32_t lastWordLength = 0;
+        int32_t lineWidth = offset->x;
 
         while ( data != dataEnd ) {
-            if ( *data == lineSeparator ) {
+            if ( fheroes2::CharHandler::isLineSeparator( *data ) ) {
                 if ( lineLength > 0 ) {
-                    offset->x += lineWidth;
+                    offset->x = lineWidth;
```
ihhub, line 126: Could you please explain why we modified the logic here?

src/fheroes2/gui/ui_text.h
```diff
@@ -169,10 +233,17 @@ namespace fheroes2
...
+            _text = std::move( text );
+            _fontType = fontType;
+        }
 
         // This method modifies the underlying text and ends it with '...' if it is longer than the provided width.
-        void fitToOneRow( const int32_t maxWidth );
+        void fitToOneRow( const int32_t maxWidth, const bool ignoreSacesAtEnd = true );
+
+        // Returns the character position number in the 'text' string. NOTICE: the input text should contain the cursor ('_').
+        size_t getTextInputCursorPosition( const size_t currentTextCursorPosition, const Point & pointerCursorOffset, const Rect & textRoi );
```
ihhub, line 246: Please make it as a normal function as **isFontAvailable()**. Having this method automatically ponders a question: why other classes don't have it. Making it just a function removes these questions.

src/fheroes2/gui/ui_text.h
```diff
@@ -169,10 +233,17 @@ namespace fheroes2
...
-        void set( std::string text, const FontType fontType );
+        void set( std::string text, const FontType fontType )
+        {
+            _text = std::move( text );
+            _fontType = fontType;
+        }
 
         // This method modifies the underlying text and ends it with '...' if it is longer than the provided width.
-        void fitToOneRow( const int32_t maxWidth );
+        void fitToOneRow( const int32_t maxWidth, const bool ignoreSacesAtEnd = true );
```
ihhub, line 243: **ignoreSacesAtEnd** is misspelled. Also, it is not clear: at the end of line or text?

src/fheroes2/gui/ui_text.h
ihhub, line 94: I suggest to name it as `FontCharHandler` as seeing `fheroes2::CharHandler` in the code outside this file might confuse developers.

ihhub (changes requested):
Hi @Districh-ru , I added few more comments here. Can you please take a look at them?

src/fheroes2/gui/ui_text.cpp
Districh-ru: Initially for the Multi-Font texts we were counting the the `lineWidth` from **0** and we added it to the `offset->x` that contains the width from the previous font.
Now we count `lineWidth` from this previous `offset->x` value, so now we can use `=` instead `+=`. I thought that it will be a little faster)
For the non-Multi-Font texts we always count from **0** and here we were doing ` = 0 + lineWidth `.

src/fheroes2/gui/ui_text.h
Districh-ru: Other classes also do not have `fitToOneRow()` method. :)
In this method we use `_text`, `_fontType` fields and `.width( maxWidth )` method.
If we want to make it as a separate function we will need one again to make inside it the `fheroes2::Text` object and do not forget to insert the cursor ('_') to the original text (`_text` in the existing object already has '_' inserted).
So first we need to make `Text` class helper functions to get Info from the `Text` object to set cursor position...

Districh-ru: Done!)

Branikolog:
Hi, @Districh-ru !

I've found a problem with window size for all events, signs all other dialog windows:
image

Can you, please, fix it?

ihhub:
@Branikolog , the bug should be fixed now.

ihhub:
Hi @Districh-ru , I made the following changes in this pull request:
- removed extra `fheroes2::` namespace usage (it was in new and old code)
- do not pass **FontType** by reference by rather by value. The size of **FontType** is just 2 bytes while reference has size of 4 or 8 bytes
- fix rendering bug for multi-line text

ihhub (approved):

Branikolog (approved):
👍

ihhub:
@Districh-ru , huge thanks for these changes!

PR #8605, opened 2024-05-04
Add a window to enter Random Ultimate Artifact radius

ihhub:
also, make monster count window to contain an image.

relates to #6845

### DIFF
--- a/src/fheroes2/dialog/dialog_selectcount.cpp
+++ b/src/fheroes2/dialog/dialog_selectcount.cpp
@@ -28,6 +28,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "agg_image.h"
--- a/src/fheroes2/dialog/dialog_selectcount.cpp
+++ b/src/fheroes2/dialog/dialog_selectcount.cpp
@@ -70,10 +70,10 @@ namespace
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -70,10 +71,10 @@ namespace
     }
 }
 
-class SelectValue : public fheroes2::Rect
+class SelectValue final : public fheroes2::Rect
```
Districh-ru, line 74: Since this class is based on `Rect` and its fields (x, y, width and height) are not used we can remove `fheroes2::Rect pos;` (line 159) and use these fields instead of `pos`.
Then in line 229 we can reduce the render area: `display.render( selectionBox );`.

Or we can add a method:
```cpp
    fheroes2::Rect getArea() const
    {
        return pos;
    }
```
and in line 229 call it to set the render area: `display.render( selectionBox.getArea() );`.

And in this case there is no need to use `Rect` as a base class, since its methods and fields are never used for `SelectValue` class.

Districh-ru (commented):
Hi, @ihhub, I left a couple of suggestions. Could you please check them out if they are valuable when you have time?

Districh-ru (approved):

Branikolog (approved):
Works like a charm!

PR #8610, opened 2024-05-04
Properly connect roads and castles in Editor

Districh-ru:
Relates to #6845

img

### DIFF
--- a/src/fheroes2/maps/maps_tiles_helper.cpp
+++ b/src/fheroes2/maps/maps_tiles_helper.cpp
@@ -3147,8 +3147,8 @@ namespace Maps
             }
             return true;
         case MP2::OBJ_CASTLE:
-        case MP2::OBJ_RANDOM_TOWN:
         case MP2::OBJ_RANDOM_CASTLE:
+        case MP2::OBJ_RANDOM_TOWN:
             if ( !placeObjectOnTile( tile, info ) ) {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -3134,6 +3142,24 @@ namespace Maps
             tile.metadata()[0] = info.metadata[0];
             tile.metadata()[1] = info.metadata[1];
 
+            if ( updateMapPassabilities ) {
+                world.updatePassabilities();
+            }
+            return true;
+        case MP2::OBJ_CASTLE:
+        case MP2::OBJ_RANDOM_TOWN:
+        case MP2::OBJ_RANDOM_CASTLE:
```
ihhub, line 3151: Please swap these 2 cases to have an alphabetic order.

ihhub (commented):
Hi @Districh-ru , I left one small correction here. Could you please check it?

ihhub (approved):

Branikolog:
Haven't found any major issues here.

Not sure about such case only:

https://github.com/ihhub/fheroes2/assets/55348946/7bc46090-da4d-4727-b4d4-ef0b1a10247c

There's a small road defect that appears when trying connect castle with a horizontal road.

Districh-ru:
> There's a small road defect that appears when trying connect castle with a horizontal road.

Currently there is no road tile for such connection. In future after we move roads to `Map_Format` we can add new road types. In this case it will be a combination of two roads: horizontal one and the castle road above it.

Branikolog (approved):
Okay. 
So missing road for castle entrance works fine in the rest of cases.

ihhub:
@Districh-ru , many thanks for this fix!

PR #8613, opened 2024-05-05
Implement simple Save Map dialog for Editor

Districh-ru:
relates to https://github.com/ihhub/fheroes2/issues/6845
img

This dialog allows:
- setting file name;
- changing map name;
- deleting the existing map file.

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2023                                             *
+ *   Copyright (C) 2019 - 2024                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2009 by Andrey Afletdinov <fheroes2@gmail.com>          *
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -368,7 +368,7 @@ std::string System::GetBasename( std::string_view path )
 
 std::string System::truncateFileExtensionAndPath( std::string_view path )
 {
-    std::string res = System::GetBasename( path );
+    std::string res = GetBasename( path );
     const size_t it = res.rfind( '.' );
     if ( it != std::string::npos ) {
         res.resize( it );
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -369,10 +369,13 @@ std::string System::GetBasename( std::string_view path )
 std::string System::truncateFileExtensionAndPath( std::string_view path )
 {
     std::string res = GetBasename( path );
-    const size_t it = res.rfind( '.' );
-    if ( it != std::string::npos ) {
-        res.resize( it );
+
+    const size_t pos = res.rfind( '.' );
+
+    if ( pos != std::string::npos ) {
+        res.resize( pos );
     }
+
     return res;
 }
 
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -366,6 +366,16 @@ std::string System::GetBasename( std::string_view path )
     return std::string{ path.substr( pos + 1 ) };
 }
 
+std::string System::truncateFileExtensionAndPath( std::string_view path )
+{
+    std::string res = System::GetBasename( path );
+    const size_t it = res.rfind( '.' );
+    if ( it != std::string::npos ) {
+        res.resize( it );
+    }
+    return res;
+}
+
 bool System::IsFile( const std::string & path, bool writable )
 {
     if ( path.empty() ) {
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -566,13 +566,20 @@ namespace Interface
                         continue;
                     }
 
+                    std::string fullPath = System::concatPath( mapDirectory, fileName + ".fh2m" );
+
+                    if ( System::IsFile( fullPath )
+                         && Dialog::NO
+                                == fheroes2::showStandardTextMessage( "", _( "Are you sure you want to overwrite the existing map?" ), Dialog::YES | Dialog::NO ) ) {
+                        continue;
+                    }
+
                     _mapFormat.name = std::move( mapName );
+                    _loadedFileName = std::move( fileName );
 
                     if ( _mapFormat.description.empty() ) {
                         _mapFormat.description = "Put a real description here.";
                     }
-
-                    std::string fullPath = System::concatPath( mapDirectory, fileName + ".fh2m" );
 
                     if ( Maps::Map_Format::saveMap( fullPath, _mapFormat ) ) {
                         // On some OSes like Windows, the path may contain '\' symbols. This symbol doesn't exist in the resources.
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1567,16 +1567,19 @@ namespace Interface
 
         std::string fileName = _loadedFileName;
         std::string mapName = _mapFormat.name;
+        std::string fullPath;
 
-        if ( !Editor::mapSaveSelectFile( fileName, mapName ) ) {
-            return;
-        }
+        while ( true ) {
+            if ( !Editor::mapSaveSelectFile( fileName, mapName ) ) {
+                return;
+            }
 
-        std::string fullPath = System::concatPath( mapDirectory, fileName + ".fh2m" );
+            fullPath = System::concatPath( mapDirectory, fileName + ".fh2m" );
 
-        if ( System::IsFile( fullPath )
-             && Dialog::NO == fheroes2::showStandardTextMessage( "", _( "Are you sure you want to overwrite the existing map?" ), Dialog::YES | Dialog::NO ) ) {
-            return;
+            if ( !System::IsFile( fullPath )
+                 || fheroes2::showStandardTextMessage( "", _( "Are you sure you want to overwrite the existing map?" ), Dialog::YES | Dialog::NO ) == Dialog::YES ) {
+                break;
+            }
         }
 
         _mapFormat.name = std::move( mapName );
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1660,7 +1660,7 @@ namespace Interface
 
         if ( Maps::Map_Format::saveMap( fullPath, _mapFormat ) ) {
             // On some OSes like Windows, the path may contain '\' symbols. This symbol doesn't exist in the resources.
-            // To avoid this we have to replace all '\' symbols by '/' symbols.
+            // To avoid this we have to replace all '\' symbols with '/' symbols.
             StringReplace( fullPath, "\\", "/" );
 
             _warningMessage.reset( _( "Map saved to: " ) + fullPath );
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -541,7 +541,7 @@ namespace Interface
                     res = eventNewMap();
                 }
                 else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_SAVE_GAME ) ) {
-                    saveMapFile();
+                    saveMapToFile();
                 }
                 else if ( HotKeyPressEvent( Game::HotKeyEvent::MAIN_MENU_LOAD_GAME ) ) {
                     res = eventLoadMap();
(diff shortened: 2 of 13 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -539,35 +541,7 @@ namespace Interface
                     res = eventNewMap();
                 }
                 else if ( HotKeyPressEvent( Game::HotKeyEvent::WORLD_SAVE_GAME ) ) {
```
ihhub, line 543: What's about UI menu to save a map?

src/fheroes2/editor/editor_save_map_window.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+    std::string resizeToShortName( const std::string & str )
+    {
+        std::string res = System::GetBasename( str );
+        const size_t it = res.rfind( '.' );
+        if ( it != std::string::npos ) {
+            res.resize( it );
+        }
+        return res;
+    }
```
ihhub, line 68: This code is repeated at least 3 times. I suggest to move it to `system.h|cpp` file. We can name it as `truncateFileExtension()`.

src/fheroes2/editor/editor_save_map_window.cpp
```diff
@@ -0,0 +1,498 @@
...
+        listbox.SetScrollButtonUp( listIcnId, 0, 1, { scrollbarOffsetX, listRoi.y + 1 } );
+        listbox.SetScrollButtonDn( listIcnId, 2, 3, { scrollbarOffsetX, listRoi.y + listRoi.height - 15 } );
+        listbox.setScrollBarArea( { scrollbarOffsetX + 2, listRoi.y + topPartHeight, 10, listRoi.height - 2 * topPartHeight } );
+        listbox.setScrollBarImage( fheroes2::AGG::GetICN( listIcnId, 4 ) );
+        listbox.SetAreaMaxItems( ( listRoi.height - 7 ) / fheroes2::getFontHeight( fheroes2::FontSize::NORMAL ) );
+        listbox.SetListContent( lists );
+        listbox.updateScrollBarImage();
+
+        if ( fileName.empty() ) {
+            fileName = _( "My map" );
```
ihhub, line 292: File name as well as map name are not translatable. We don't support any other languages at the moment.

ihhub (changes requested):
Hi @Districh-ru , I put several comments here. Can you please check them?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Hi, @ihhub, the save map UI is opened in the function called in the next line. This function is also called from File dialog.

Branikolog:
Hello, @Districh-ru !
Thanks for working on such important improvement!

First of all, can you explain, please, why do we need to have separate file name and map name options? I've always found this really confusing in the original game editor. 

After a couple of tests I found a few issues. Could you, please, look?

After creating a new map (without any fh2m maps available in the list) and trying to save - "okay" button is in disabled mode and I cannot press the button and save it with the default name (But I can press Enter and save it). However, once I've added and deleted any symbol - I save map with a proposed name. But "Okay" button is still in a disabled mode.


https://github.com/ihhub/fheroes2/assets/55348946/40c969f5-f367-43dd-b675-9d893746409b


https://github.com/ihhub/fheroes2/assets/55348946/49fb93bf-73d0-4371-9b7d-f26aa3037608

After deleting all files "okay" button is still in active state:

https://github.com/ihhub/fheroes2/assets/55348946/386a440e-0814-4e2e-abb3-f2167b6874c7

When I try to enter any symbol  in this case - application crashes.

Districh-ru:
Hello, @Branikolog!
Thanks for testing!

> First of all, can you explain, please, why do we need to have separate file name and map name options? I've always found this really confusing in the original game editor.

The map name can be translated and it should not exceed 17 symbols to fit original dialogs. It can also contain special symbols( "?", "/" and others) that are not allowed in filename by OS.

In example map "WHOAM.MP2" on Buka's CD has name "Кто я?". Modern OS allows to have spaces but do not allow the '?' symbol. :(

> After a couple of tests I found a few issues. Could you, please, look?

These issues should be fixed now.
...
Probably we have them also in save/load dialog :(

Districh-ru:
> Probably we have them also in save/load dialog :(

And, yes, we have them.

ihhub:
Hi @oleg-derevenetz , could you please take a look at the changes mostly for system.h|cpp files?

ihhub:
Hi @zenseii , can you please check newly added strings?

src/engine/system.cpp
```diff
@@ -366,6 +366,16 @@ std::string System::GetBasename( std::string_view path )
     return std::string{ path.substr( pos + 1 ) };
 }
 
+std::string System::truncateFileExtensionAndPath( std::string_view path )
+{
+    std::string res = GetBasename( path );
+    const size_t it = res.rfind( '.' );
```
oleg-derevenetz, line 372: It's not really an iterator, but rather a position, so I'd call it `pos`, not `it`.

ihhub:
Hi @zenseii , I've addressed your comments.

oleg-derevenetz (commented):
Hi @Districh-ru I just left one suggestion (naming-related), LGTM otherwise.

src/engine/system.cpp
Districh-ru: Hi @oleg-derevenetz, I've addressed your comment.

oleg-derevenetz (approved):

Branikolog:
Hi, @Districh-ru 

Thank you for answering my questions.

For me personally current implementation looks quite confusing. I'm changing file name, but map name remains the same on default.
As a result we have two names in save dialog, but only one in load game dialog. So it is not really clear, what name we see in load game dialog.

https://github.com/ihhub/fheroes2/assets/55348946/fd90405e-5976-4b9e-8c44-5b0758b73076

When creating a new map and saving it I expect this name appear in map list. But currently the engine first of all offers me to enter a file name and allows to save even with the same map name. I agree, that it's up to a mapmaker to change both names in fields at once, but this behaviour seems too complex and inconvenient.

What if we mirror any file name typed into map name field on default? If player wants to add special symbols or change map name, he should click on the second field to proceed his ideas.
For now, when we select any file in the list a suggested map name remains the same. I think it should be updated, so we don't need to type new map name each time we want to save map in a new file.

Several issues found. Could you, please check them?

Application crashes when I try to check map info with right click. :(

Application crashes when I delete file name from the field and left click on it (empty field).

Also, we're introducing new map format but the game engine still can not recognize map name and shows "?????" instead.
image

I thought this shouldn't be an issue with our new maps.


When  we refuse confirmation dialog, I think save file list should be still on the screen. The most common case of refusing the dialog is when player accidentaly trying to overwrite the existing file. So, I expect the engine shows file list back, so I can select the proper file to be overwritten.

https://github.com/ihhub/fheroes2/assets/55348946/dca746a7-7c4f-4ce8-ab27-8515a7bc5746


When I delete the whole text "okay" button still remains in disabled state, after I select the existing file.

https://github.com/ihhub/fheroes2/assets/55348946/c6588798-1e1b-47fb-826c-d4b7b22202fa


One more question. Do we really need a separate text dialog here?

https://github.com/ihhub/fheroes2/assets/55348946/90691b61-ade9-4428-9c5f-6c015b96be66

Why cannot flashing line appear in the right field so we can type text there with a keyboard or virtual one?

ihhub:
Hi @Branikolog , file name and map name are absolutely different. Name of the map remains the same no matter how you name your file. It is the same as demanding to name an Excel file with a tab called "Finance" as `finance.xlsx`.

What we can do is to show **.f2m** at the end of each file name but we need to display it with different (gray?) font. What do you think @Districh-ru ?

Branikolog:
Hi, @ihhub 

> Hi @Branikolog , file name and map name are absolutely different. Name of the map remains the same no matter how you name your file. It is the same as demanding to name an Excel file with a tab called "Finance" as `finance.xlsx`.
> 
> What we can do is to show **.f2m** at the end of each file name but we need to display it with different (gray?) font. What do you think @Districh-ru ?

I agree with your statements, but the current implementation looks really confusing with those 2 separate customizable fields.
As I stated above, in the "load" game dialog we have just map names. Maybe we should render same window type there for the consistency?

tlsa:
I think the save dialog should only show filename.

Map name should be completely separate.

That would remove the confusion.

Branikolog:
Hi, @tlsa 

> I think the save dialog should only show filename.
> 
> Map name should be completely separate. It probably belongs in a separate map properties dialog.
> 
> That would remove the confusion.

Then, after saving a map it would be difficult to find it in maps list (as it shows map names only) when file and map names are different.

I like idea of @ihhub to show map format, so it easer to distinguish map and file names.

tlsa:
> Then, after saving a map it would be difficult to find it in maps list (as it shows map names only) when file and map names are different.

I'd change the title of the window from "Save Map" to "Save Map:\n[map title]" to avoid that problem.

Branikolog:
@Districh-ru 

image

Maybe file extensions could be some scale of grey, as @ihhub suggested. I believe, they shouldn't be so bright, even the less recognizable colour could be better.

All my notes are a matter of my personal preferences of the design. Dialog works fine after all. It's just has a little "toolish" design, which is not even consistent with "Load map" window, introducing two separate lists.

ihhub:
@Branikolog , new map format does not support any languages rather than English at the moment so seeing ??? is absolutely valid.

To solve the problem with map names I suggest 2 solutions:
- do not show map name at all. Users will know what they are dealing with in this UI window. We can still show map information by right mouse click
- show map name together with map type icon, size and player count like it is done in load map window. In this case it will be much more obvious what map name stands for

I will vote for the second solution as users would still need to have some information about maps.

Branikolog:
Hi, @ihhub 

I think, players should meet same dialog for save and load scenarios. Missing part with either file name or map name confuses users anyway. Can we come up with a design a little more uniform for these two dialogs?
(File extension helps, but not in the yellow colour in my opinion. Takes too much attention and also confuses with selection colour too.)

>show map name together with map type icon, size and player count like it is done in load map window. In this case it will be much more obvious what map name stands for

I suppose, you're proposing to take away map file name list and show same dialog to "load map" one.
But we still need a special field to edit map file name...

I think we're going to change it anyway in future. It would be nice to show map description, some additional map details and even minimap image, so player can browse map list with maximum comfort. If we're going to expand the window, we can show map name just in a separate field, after selecting the map, together with all other information.

So I'm +/- ok with current implementation (considering my remarks on colours) if we show same dialog when we're trying to load a map.

Districh-ru:
Everybody, thanks for your suggestions!

I decided to make a more common to other editors (Heroes 3, Heroes 4) save dialog: show only left aligned file name with extension. More file data including map name and description can be checked by the right mouse button press on the map file.
изображение

The name of the saved map can be changed by clicking on it. We may remove this feature once the Map Specifications dialog is implemented.

Districh-ru:
To match the Save Map dialog, the Load Map dialog will now show the map file names:
изображение

tlsa:
Save looks good now.

Load looks good if it's only the map loader for the editor. The map loader for the game should continue to show the map **name**.

Players don't care about the map filename.
Editors need to know what file they're actually editing.

Districh-ru:
> Load looks good if it's only the map loader for the editor. The map loader for the game should continue to show the map **name**.

The select scenario dialog shows map name as always. :) The changes were made for Editor's map loader.

ihhub (approved):

Branikolog:
Looks good to me.
@Districh-ru , did map name modifying was not implemented in this iteration?

Branikolog:
Sorry, didn't find where to click for the first time. :)
Actually, the spacing between map name and window title could be a little wider.
image

It would be nice to have a field for the map name, but since this solution is temporary, I don't insist on that...
image

Districh-ru:
Hi, @Branikolog.
I've applied your suggestions:
изображение

Branikolog:
Hi, @Districh-ru !

Maybe it's better to have field size as maximum possible string size (Not a full line, like file name in the bottom part, but a short one center-aligned), rather than adjusted to the name one... How do you think?
For short names it looks weird. :)
image

Districh-ru:
> Maybe it's better to have field size as maximum possible string size (Not a full line, like file name in the bottom part, but a short one center-aligned), rather than adjusted to the name one... How do you think? For short names it looks weird. :) 

@Branikolog 
Sure. I did the changes.
The widest string can be if it consists of the widest characters. :)
изображение

Branikolog:
@Districh-ru 
 I didn't expect it be so wide. 
 I suggest making it the same size and location as the bottom line.
 
image

Districh-ru:
@Branikolog 
Done)

Branikolog (approved):
Looks nice!

ihhub:
@Districh-ru , many thanks for this implementation!

PR #8615, opened 2024-05-05
Fix software cursor show delay

Districh-ru:

To reproduce the issue on 'master' branch disable Animation within the Editor and use software rendering mode (`cursor soft rendering = on` in **fheroes2.cfg**).

When cursor image is not changed and the cursor must be shown after closing info dialog the **CursorRestorer** destructor now will now force display render to show the cursor.

https://github.com/ihhub/fheroes2/assets/113276641/07bac4c4-3282-4af5-94ec-be0d19e5617a

The cases when cursor image is changed we should be handled in the corresponding code: the corresponding cursor handler must be called. This case is not fixed in this PR.


ihhub (approved):

ihhub:

PR #8616, opened 2024-05-05
Minor improvements to 'showMessage()' function

Districh-ru:
These changes are made to work better with https://github.com/ihhub/fheroes2/pull/8615

When the cursor should be hidden (showing message by the right click) now we do not change the cursor theme to help https://github.com/ihhub/fheroes2/pull/8615 to show it quicker.

Also this PR slightly optimizes the render loop. (This dialog with animation can be checked, in example, in Tavern).


ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for this improvement!

PR #8618, opened 2024-05-05
Fix some bugs in save/load game dialog

Districh-ru:
This PR fixes bugs observed in #8613 :
- deletion of the last file in the Load Game dialog results in game crash;
- deleting the save name in Save Game dialog does not show that the OK button is disabled (but it is);
- deleting the save name in Virtual Keyboard in Save Game dialog does not disable the OK button;
- if the Virtual Keyboard was opened for an empty save name and the non-empty name was given the OK button remains disabled after the keyboard dialog closes.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -381,6 +381,17 @@ namespace
             charInsertPos = filename.size();
         }
 
+        auto buttoOkdisabler = [&buttonOk, &filename]() {
+            if ( filename.empty() ) {
+                buttonOk.disable();
+                buttonOk.draw();
+            }
+            else if ( buttonOk.isDisabled() ) {
+                buttonOk.enable();
+                buttonOk.draw();
+            }
+        };
+
         listbox.Redraw();
         redrawTextInputField( filename, textInputRoi, isEditing );
 
@@ -481,13 +492,7 @@ namespace
                     isListboxSelected = false;
                     needRedraw = true;
 
-                    if ( filename.empty() ) {
-                        buttonOk.disable();
-                    }
-                    else {
-                        buttonOk.enable();
-                    }
-                    buttonOk.draw();
+                    buttoOkdisabler();
 
                     // Set the whole screen to redraw next time to properly restore image under the Virtual Keyboard dialog.
                     display.updateNextRenderRoi( { 0, 0, display.width(), display.height() } );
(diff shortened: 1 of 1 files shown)


ihhub (approved):

Branikolog:
Hi, @Districh-ru 
Sorry, I forgot, that you've opened a new PR with fix for the file list...
Could you, please, try to fix both issues I've mentioned in https://github.com/ihhub/fheroes2/pull/8613#issuecomment-2094942431 as they still valid in the current build?

Also I've noticed, that I cannot delete files in save file window. Deleting works only in Load game window...
I think we need to add a special button to be able to delete files. It will be useful for devices without physical keyboard and also allows to delete files in both load and save game windows. How do you think?

ihhub:
> Hi, @Districh-ru Sorry, I forgot, that you've opened a new PR with fix for the file list... Could you, please, try to fix both issues I've mentioned in [#8613 (comment)](https://github.com/ihhub/fheroes2/pull/8613#issuecomment-2094942431) as they still valid in the current build?
> 
> Also I've noticed, that I cannot delete files in save file window. Deleting works only in Load game window... I think we need to add a special button to be able to delete files. It will be useful for devices without physical keyboard and also allows to delete files in both load and save game windows. How do you think?

Hi @Branikolog , all new features should be done outside this pull request. We aren't going to add file deletion here.

Branikolog:
@ihhub 

> > Hi, @Districh-ru Sorry, I forgot, that you've opened a new PR with fix for the file list... Could you, please, try to fix both issues I've mentioned in [#8613 (comment)](https://github.com/ihhub/fheroes2/pull/8613#issuecomment-2094942431) as they still valid in the current build?
> > Also I've noticed, that I cannot delete files in save file window. Deleting works only in Load game window... I think we need to add a special button to be able to delete files. It will be useful for devices without physical keyboard and also allows to delete files in both load and save game windows. How do you think?
> 
> Hi @Branikolog , all new features should be done outside this pull request. We aren't going to add file deletion here.

I just asked for an opinion. :)
I remember, that some users asked for that feature. It can also help in managing save files without fh2 tools.

Districh-ru:
Hi, @Branikolog,  I've fixed the issues you found. The right click info crash is fixed in https://github.com/ihhub/fheroes2/pull/8624.
Could you check it please when you have time.
We will add the DELETE button in the other PR later.

Branikolog:
Hi, @Districh-ru 

Are we going to fix the issue with disabled "okay" button in a separate PR?

https://github.com/ihhub/fheroes2/assets/55348946/9d71c9ad-ad72-4088-beb7-c632ea04463d

Everything except this issue works fine! 👍

Districh-ru:
> Are we going to fix the issue with disabled "okay" button in a separate PR?

Hi, @Branikolog, I forgot this. :(
Thanks for reminding me! Now it should be fixed.

Could you please also recheck other cases for OK button enable/disable when you have time?

Branikolog (approved):
@Districh-ru 
Thanks for fixing this!

ihhub:
@Districh-ru , thank you so much for the fixes!

PR #8623, opened 2024-05-06
Properly handle objects while changing terrain

ihhub:
- remove objects that do not belong to the new terrain
- update castles' basements according to the new terrain

relates to #6845

Things to test:
- erasure tool works as intended
- placing water over objects that cannot be placed on water (and vice versa) results in their removal
- town basements should be updated according to new terrain
- general object placement on different terrain types

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -542,12 +542,7 @@ namespace
             const auto & townObjectInfo = Maps::getObjectInfo( groupType, objectType );
             const auto & basementObjectInfo = Maps::getObjectInfo( Maps::ObjectGroup::LANDSCAPE_TOWN_BASEMENTS, basementId );
 
-            if ( !isObjectPlacementAllowed( townObjectInfo, tilePos ) ) {
-                errorMessage = _( "Objects cannot be placed outside the map." );
-                return false;
-            }
-
-            if ( !isObjectPlacementAllowed( basementObjectInfo, tilePos ) ) {
+            if ( !isObjectPlacementAllowed( townObjectInfo, tilePos ) || !isObjectPlacementAllowed( basementObjectInfo, tilePos ) ) {
                 errorMessage = _( "Objects cannot be placed outside the map." );
                 return false;
             }
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1764,7 +1764,7 @@ namespace Interface
                     continue;
                 }
 
-                if ( !verifyTerrainPlacement( pos, object.group, object.index, errorMessage ) ) {
+                if ( !verifyTerrainPlacement( pos, object.group, static_cast<int32_t>( object.index ), errorMessage ) ) {
                     uids.emplace( object.id );
                 }
             }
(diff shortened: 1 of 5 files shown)


Branikolog:
Hi, @ihhub 

Multiple assertions occur, when I've tried to place any terrain over the existing, after I've placed several map objects. 
image

ihhub:
Hi @Branikolog , I've fixed the crashes.

Branikolog:
@ihhub 

One object does not seem to be affected by terrain change:

https://github.com/ihhub/fheroes2/assets/55348946/65727b8a-6e39-4776-9b17-1f6f3f483a18

ihhub:
Hi @Branikolog , river delta is terrain independent by design.

Branikolog (approved):
So no other problems occurred. 👍

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -395,6 +392,217 @@ namespace
...
+                return false;
+            }
+        }
+        else if ( groupType == Maps::ObjectGroup::LANDSCAPE_TREES ) {
+            const auto & objectInfo = Maps::getObjectInfo( groupType, objectType );
+            if ( !checkConditionForUsedTiles( objectInfo, tilePos, []( const Maps::Tiles & tileToCheck ) { return !tileToCheck.isWater(); } ) ) {
+                errorMessage = _( "Trees cannot be placed on water." );
+                return false;
+            }
+        }
```
Districh-ru, line 446: Can we combine these mostly identical code, in example:
```cpp
        if ( groupType != Maps::ObjectGroup::ADVENTURE_WATER && groupType != Maps::ObjectGroup::LANDSCAPE_WATER ) {
            const auto & objectInfo = Maps::getObjectInfo( groupType, objectType );
            if ( !checkConditionForUsedTiles( objectInfo, tilePos, []( const Maps::Tiles & tileToCheck ) { return !tileToCheck.isWater(); } ) ) {
                errorMessage = _( "%{objectGroup} cannot be placed on water." );
                StringReplace( text, "%{objectGroup}", getObjectGroupName( groupType ) );
                return false;
            }
        }
```

and introduce `getObjectGroupName()` function to return group name?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -395,6 +392,217 @@ namespace
...
+
+            if ( !isObjectPlacementAllowed( townObjectInfo, tilePos ) ) {
+                errorMessage = _( "Objects cannot be placed outside the map." );
+                return false;
+            }
+
+            if ( !isObjectPlacementAllowed( basementObjectInfo, tilePos ) ) {
+                errorMessage = _( "Objects cannot be placed outside the map." );
+                return false;
+            }
```
Districh-ru, line 589: Can we join these two conditions?
`if ( !isObjectPlacementAllowed( townObjectInfo, tilePos ) || !isObjectPlacementAllowed( basementObjectInfo, tilePos ) )` ...

Districh-ru (commented):
Hi, @ihhub, I left two comments. Could you please check them when you have time?

src/fheroes2/editor/editor_interface.cpp
ihhub: I believe it is going to be very lengthy and ugly to read :(

ihhub: Actually, you are right. I was looking at the different place.

PR #8624, opened 2024-05-06
Fix a crash when rendering a multi-font text with very long words

Districh-ru:
This PR fixes a regression after https://github.com/ihhub/fheroes2/pull/8604: If a multi-font text contains a very-long word (e.g. a path to file that contains no spaces) the render of such text will crash if this word is not left-aligned.
It is because previously to detect that the word really is the first in the sentence the condition `x == 0` (offset from the left in text ROI) was checked. But it is not always true.
Now the engine will check if the "first" word that not fits the line starts from the right part of the line than it will be moved to the next line.

To check the issue just press right mouse button on every save file in Load game dialog.

### DIFF
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -192,27 +192,25 @@ namespace
                             data -= lastWordLength;
                         }
                     }
+                    else if ( isSpaceChar( *data ) ) {
+                        // Current character could be a space character then current line is over.
+                        // For the characters count we take this space into the account.
+                        lineInfo->characterCount = lineLength + 1;
+                        lineInfo->offset.x = lineWidth;
+
+                        // We skip this space character.
+                        ++data;
+                    }
+                    else if ( lastWordLength > 0 ) {
+                        // Exclude last word from this line.
+                        data -= lastWordLength;
+
+                        lineInfo->characterCount = lineLength - lastWordLength;
+                        lineInfo->offset.x = lineWidth - getLineWidth( data, lastWordLength, fontType );
+                    }
                     else {
-                        if ( isSpaceChar( *data ) ) {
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/ui_text.cpp
```diff
@@ -381,11 +381,11 @@ namespace
                             ++data;
                         }
                         else {
-                            if ( lineInfo->offset.x == 0 ) {
+                            if ( lineInfo->offset.x < maxWidth / 2 ) {
```
ihhub, line 384: @Districh-ru , can you please explain why we need to divide by 2?

Districh-ru: Hi, @ihhub.
We have the almost same logic in `getMultiRowInfo()` and `renderMultiLine()`. But with that small difference:
- In `getMultiRowInfo()` we calculate the `offset.x` for each line. If the line was just started then `offset.x == 0`. Otherwise we have continued this line with the other font.
- In `renderMultiLine()` we have already calculated `offet.x`. In this case a way to determine that the word that does not fit the line is a word with a new font is to check thay its start `offset.x` is in the right part of text ROI. Currently we render only centered multi-line texts where every line starts in the left part of ROI (for this case `offset.x` is always `< maxWidth / 2`). 
If `offet.x >= maxWidth / 2` then this first word is started not from the begining of the line (line begins only with offsets that are lower tnan `maxwidth / 2`).


After we finish the Editor preparations I'll update the `Text` code.

Districh-ru: Just press right mouse button on any saved file in Load game dialog: the long file path that does not fit any line crashes the engine without a fix.

ihhub (approved):

ihhub:

PR #8632, opened 2024-05-11
Do not redraw save dialog fully when only editing the save name

Districh-ru:
This PR optimizes save game/map dialog not to render the whole dialog when only the save name was changed.


ihhub (approved):

ihhub:
@Districh-ru , thanks a lot for the fix!

PR #8633, opened 2024-05-11
Fix maps sorting in Editor

Districh-ru:
We show file names, not map names. So we should sort maps by file name.


ihhub (approved):

ihhub:

PR #8638, opened 2024-05-12
Fix Save map dialog list rendering

Districh-ru:
This PR fixes the issue on small resolutions when there are more maps to display (maximum 22) than can fit the height of the map list.
Master build:


ihhub (approved):

ihhub:
Hi @Branikolog , please test it as a priority.

ihhub:
@Districh-ru , thanks a lot for the fix!

PR #8652, opened 2024-05-14
Display object coordinates in the right mouse click popup window

ihhub:
relates to #6845

This is subject for discussions. We need these coordinates for the future winning and loss conditions where we might show coordinates of objects.

image


src/fheroes2/editor/editor_object_popup_window.cpp
```diff
@@ -80,6 +80,12 @@ namespace Editor
             infoString = getObjectInfoText( tile );
         }
 
+        infoString += "\n[";
+        infoString += std::to_string( tile.GetIndex() % world.w() );
+        infoString += "; ";
+        infoString += std::to_string( tile.GetIndex() / world.w() );
+        infoString += ']';
```
Districh-ru, line 87: Hi, @ihhub, I have one concern: Should we show `mainTileIndex` in the case when it is not `-1`?
In example, you may click on any tile of an unnamed castle , but you will get the info that it is a castle. So a map makes should know the coordinates of the castle (its "main" action tile), but not the coordinates of the current tile to use this coordinates to determine the castle in the list of many unnamed castles.

ihhub: Hi @Districh-ru , however, showing coordinates not of the current tile might be misleading as well. We are showing only the type of an object. If a user clicks on a Dragon City (which is big) they still expect to see "Dragon City" on every tile.

Districh-ru: Agree.
If it will be needed in future we may add showing the `mainTileIndex` only for castles to help users to find the appropriate castle in a list that contains castles with their main tile coordinates.

Districh-ru (approved):

Branikolog (approved):

PR #8660, opened 2024-05-16
Expand map format for the future changes

ihhub:
Also, add a check for versions.

We are bumping the minimum version to 2 to avoid all existing map being made in the past.

relates to #6845

### DIFF
--- a/src/fheroes2/maps/map_format_info.h
+++ b/src/fheroes2/maps/map_format_info.h
@@ -207,6 +207,11 @@ namespace Maps::Map_Format
         int32_t monsterCount{ 0 };
     };
 
+    struct ShrineMetadata
+    {
+        std::vector<int32_t> allowedSpells;
+    };
+
     struct DailyEvent
     {
         std::string message;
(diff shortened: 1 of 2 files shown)


src/fheroes2/maps/map_format_info.h
```diff
@@ -207,6 +207,11 @@ namespace Maps::Map_Format
         int32_t monsterCount{ 0 };
     };
 
+    struct ShrineMetadata
+    {
+        std::vector<int32_t> allowedSpells;
```
Districh-ru, line 212: Maybe we should use `bannedSpells` instead. because initially Shrines allows all spells and when the vector is empty then all spell will be available.
And for the `allowedSpells` we should fill this vector with all spells on the object creation? :)

What do you think, @ihhub?

ihhub: The idea behind is this:
- if the container is not empty randomly choose any from the provided list
- if the container is empty, then choose any (default case)

Having banned spells might not be scalable for the future changes if we add new spells.

Districh-ru (approved):

PR #8663, opened 2024-05-16
Add a general window for action objects with no metadata

ihhub:
relates to #6845

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1210,6 +1210,9 @@ namespace Interface
                         action.commit();
                     }
                 }
+                else {
+                    fheroes2::showStandardTextMessage( MP2::StringObject( objectType ), _( "This object has no properties to change." ), Dialog::OK );
+                }
             }
         }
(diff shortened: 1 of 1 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1210,6 +1210,9 @@ namespace Interface
                         action.commit();
                     }
                 }
+                else {
+                    fheroes2::showStandardTextMessage( MP2::StringObject( objectType ), _( "This object has no properties to change." ), Dialog::OK );
```
Districh-ru, line 1214: Hi, @ihhub, should we use `_warningMessage.reset();` for this case?
```cpp
std::string msg = "%{object} has no properties to change.";
StringReplace( str, "%{object}", MP2::StringObject( objectType ) );
_warningMessage.reset( std::move( msg ) );
```

ihhub: Hi @zenseii , this is for a singular object.

ihhub: However, some objects are plural as Ruins.

Districh-ru (approved):

PR #8669, opened 2024-05-17
Differentiate Random Castle and Town names

ihhub:
relates to #6845

image

image


Districh-ru:
We could also differentiate "Town Army" and "Castle Army" than. :)

Districh-ru (approved):
As I mentioned earlier we can also apply such change to change the text in the lower-right part of this dialog "Castle Army" (line 391) to show "Town Army" when we are editing Town properties.
It can be done in this or in a new PR.

ihhub:
Let's move it for 1.1.1 release. It is a very minor change and we need to go with translations.

ihhub:
Hi @Districh-ru and @zenseii , I am just thinking whether it's worth doing such granular differentiation. What do you think?

PR #8670, opened 2024-05-18
Revert changes related to fast scroll

ihhub:
This code is for general use. Here we are doing some weird magic for a specific case. Moreover, it is not correct for the Editor.

Relates to #8253. This change might break the previous logic.


ihhub:
Closing this pull request as it requires another approach.

PR #8676, opened 2024-05-18
Fix Editor interface items positions when Hide Interface mode is enabled in the engine

Districh-ru:
If Hide Interface mode is enabled in the engine: temporary disable it while organizing the Editor's interface and then restore the setting.


ihhub (approved):

ihhub:
@Districh-ru , great work!

PR #8680, opened 2024-05-18
Set the last edited map file as a default for the new Standard game

Districh-ru:
This PR sets the last loaded or saved map in Editor as a default selected map for the new Standard Game. It will be selected only if the map can be played.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1607,6 +1607,11 @@ namespace Interface
 
         _loadedFileName = System::truncateFileExtensionAndPath( filePath );
 
+        // Set the loaded map as a default map for the new Standard Game.
+        Maps::FileInfo & mapInfo = Settings::Get().getCurrentMapInfo();
+        mapInfo.filename = filePath;
+        mapInfo.name = _mapFormat.name;
+
         return true;
     }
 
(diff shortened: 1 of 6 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1607,6 +1608,11 @@ namespace Interface
 
         _loadedFileName = System::truncateFileExtensionAndPath( filePath );
 
+        // Set the loaded map as a default map for the new Standard Game.
+        Maps::FileInfo & mapInfo = Settings::Get().getCurrentMapInfo();
```
ihhub, line 1612: This is definitely a hack. We should use `SetCurrentFileInfo()` method for this. Create **FileInfo** object and use **readResurrectionMap()** method for this matter. Going forward: you can create a function in this file which accepts `Maps::Map_Format::BaseMapFormat`, create a new method in **Maps::FileInfo** called for example `loadResurrectionMap()` and move most of the logic there from `Maps::FileInfo::readResurrectionMap()`.

By the way, line 396 in **maps_fileinfo.cpp** should be `Maps::Map_Format::BaseMapFormat map;` as we don't need the full map to be loaded.

ihhub (changes requested):
Hi @Districh-ru , I left a suggestion in this pull request. Would you mind to take a look when you have spare time?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: @ihhub, many thanks for the suggestion!

ihhub (approved):

Branikolog (approved):

ihhub:
@Districh-ru , huge thanks for this update!

PR #8687, opened 2024-05-19
Change Detail Mode description to reflect the correct purpose

ihhub:
relates to #6845


ihhub:
Hi @zenseii , I am also concerned about "Cell Details" text as this is not the same as "Detail Mode":

image

PR #8688, opened 2024-05-19
Use golden color to highlight cells that are sometimes possible to edit

ihhub:
relates to #6845

image

image

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -656,8 +656,8 @@ namespace Interface
                     const fheroes2::Point indices = getBrushAreaIndicies( brushSize, _tileUnderCursor );
 
                     assert( Maps::isValidAbsIndex( indices.x ) );
-                    const bool isActionObject = ( _editorPanel.isDetailEdit() && brushSize.width == 1 && brushSize.height == 1 &&
-                                                    MP2::isOffGameActionObject( world.GetTiles( indices.x ).GetObject() ) );
+                    const bool isActionObject = ( _editorPanel.isDetailEdit() && brushSize.width == 1 && brushSize.height == 1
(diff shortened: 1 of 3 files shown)


Districh-ru (approved):
Hi, @ihhub, LGTM. Now it will be easier for the map maker to find the action tile if objects' passability is hidden.
I left one optional suggestion. Could you please take a look if it is valuable.

ihhub:
Hi @Districh-ru , I see no comments for your suggestion here :)

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -655,12 +655,16 @@ namespace Interface
                 if ( brushSize.width > 0 && brushSize.height > 0 ) {
                     const fheroes2::Point indices = getBrushAreaIndicies( brushSize, _tileUnderCursor );
 
-                    _gameArea.renderTileAreaSelect( display, indices.x, indices.y );
+                    assert( Maps::isValidAbsIndex( indices.x ) );
+                    const bool isActionObject = MP2::isOffGameActionObject( world.GetTiles( indices.x ).GetObject() );
+                    const bool isRenderForActionObject = ( isActionObject && brushSize.width == 1 && brushSize.height == 1 && _editorPanel.isDetailEdit() );
```
Districh-ru, line 660: @ihhub, ow it was lost ...
I suggested to check `_editorPanel.isDetailEdit()` first and only then call all other functions:
```cpp
const bool isRenderForActionObject = ( _editorPanel.isDetailEdit() && brushSize.width == 1 && brushSize.height == 1 && MP2::isOffGameActionObject( world.GetTiles( indices.x ).GetObject() ) );
```

ihhub: Sure, we can do this.

Branikolog (approved):

PR #8689, opened 2024-05-19
Properly connect Streams and Deltas while placing Streams

Districh-ru:
This PR fixes Streams and Deltas connections:
изображение

This PR fixes only new Streams connection with already places Deltas.

When placing a new Delta the connections are not updated and streams under the Delta are not removed. These fixes will be done after we move Streams logic from `world` vector to `Map_Format`.


Districh-ru:
In the master branch the connection of river deltas with streams is not properly implemented. All river delta tiles are considered by the engine as streams and streams are automatically placed above deltas during stream interconnection algorithm. 
It causes extra streams above the delta and after erasing deltas.

https://github.com/ihhub/fheroes2/assets/113276641/cd649b96-2f53-4d32-9792-730e2bb1fc9d

Branikolog:
Hi, @Districh-ru !
Nice work.
I have only one question.
Is it possible to apply stream connection for such case?
image

Districh-ru:
> Is it possible to apply stream connection for such case?

Hi, @Branikolog. 
No, the original assets do not have stream sprites for such connections. :(
Late we can design the new sprites.

Districh-ru:
> @Districh-ru Okay. Connection with the existing deltas work well for the rest available cases. However, if I place delta near any stream - connection does not occur. Is there a way to apply connection in such case? At least for the cases, when it is possible?

@Branikolog, now the existing streams should update when placing a delta nearby.

Branikolog (approved):

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1534,6 +1534,25 @@ namespace Interface
...
+            const Maps::ObjectInfo & objectInfo = Maps::getObjectInfo( groupType, objectType );
+            std::for_each( objectInfo.groundLevelParts.begin(), objectInfo.groundLevelParts.end(), [&tile]( const Maps::LayeredObjectPartInfo & info ) {
+                if ( info.icnType != MP2::OBJ_ICN_TYPE_OBJNMUL2 && info.icnIndex != 0 && info.icnIndex != 13 ) {
+                    return;
+                }
+
+                Maps::updateStreamsToDeltaConnection( tile, info.icnIndex == 13 );
+            } );
+
+            action.commit();
```
ihhub, line 1554: The above code might not change anything at all but we are still committing the change which will result in an empty action. We need to make sure that at least one updateStreamsToDeltaConnection() is being called.

ihhub (changes requested):
Hi @Districh-ru , I left one comment here. Could you please check it?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Hi, @ihhub , `commit()` is called here only when the line 1540 (`_setObjectOnTile()`) returnes true:
If object was set we commit the changes including the possible streams update.

ihhub (approved):

ihhub:

PR #8695, opened 2024-05-20
Swedish translation update (#17)

Mr-Bajs:


ihhub:
@Mr-Bajs , many thanks for the update!

ihhub:
@Mr-Bajs , the translation is completed only by 99%. There are few lines which were missed for the translation :)

Issue #8556, opened 2024-04-08
Process is frozen and use 100% CPU


dimonb:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

macOS

### Describe the bug

Game is frozen with 100% cpu usage when casting blood last (not reprodusable

process sample in attachment
[Sample of fheroes2.txt](https://github.com/ihhub/fheroes2/files/14903603/Sample.of.fheroes2.txt)

### Save file

Not reproducible, save file make no sense

oleg-derevenetz:
Hi @dimonb did you notice if the usual animations on the battle screen were working at that time (for example, were the flags of the heroes flying)?

ihhub:
Hi @dimonb , can you please reply to the original question?

dimonb:
Hi guys, sorry for delay. No other unusual animation. I got this one more time :( also blood last
It is the only case when game stuck with 100% CPU

ihhub:
@dimonb , I have an idea in mind how to fix or at least narrow down the problem so I will create a separate pull request from where you might check out the branch. I will do it after the upcoming release which is in a few days. If I miss creating a pull request please ping me again next week.

ihhub:
@dimonb , actually, I made a pull request #8698 for this matter. I set 20 seconds as a hard stop for playing Blood Lust sound. Could you please compile and test it on your side? 20 seconds would be enough to identify that this was a deadlock. To test I suggest to use Battle Only mode which is designed for battles.

dimonb:
@ihhub will try in several days, thx!

ihhub:
Hi @dimonb , have you tried the fix?

dimonb:
Yes, could not reproduce any more On 27 Jul 2024, at 13:21, Ihar Hubchyk ***@***.***> wrote:﻿
Hi @dimonb , have you tried the fix?

ihhub:
I think this might be the problem with emulation of x86 code on ARM64. This could be solved by proving ARM build for macOS.

What do you think @oleg-derevenetz ?

As of now I am closing this issue as the provided fix is a pure hack and shouldn't exist.

oleg-derevenetz:
[...]

PR #8698, opened 2024-05-20
Exit from a deadlock while playing Blood Lust sound on macOS

ihhub:
relates to #8556


oleg-derevenetz:
I should say that while this change is fine for test purposes to figure out what in fact happens, I really don't recommend to perform "patching" of this kind for 3rd-party bugs in our code (not mentioning the fact that this is not the only place that waits for some sound to finish to proceed).

ihhub:
Hi @oleg-derevenetz , yes, I wasn't planning to proceed with this hack as a solution.

ihhub:
Closing this pull request as we shouldn't use any hacks like this in the code.

PR #8702, opened 2024-05-20
AI: properly handle the Spell Book artifacts on the Adventure Map

### LINKED ISSUE #8696: AI heroes will try to pickup Magic Books on the adventure map even though they have one
### Preliminary checks

The AI will try to pick up the Magic Book placed on the adventure map even though it already has one. An easy way to test it is to place an AI hero in the middle of Magic Books.
Screenshot_20240520_101548_fheroes2

Although the above picture shows a case in which it doesn't matter for the AI, there could be other cases where it would be better for the AI to prioritize some other move.

Make the AI avoid trying to pick up Magic Books when it already has one.

### Additional info

_No response_

oleg-derevenetz:
close #8696

As far as I know, currently there is no way to place the Spell Book artifact inside an object other than `OBJ_ARTIFACT` (e.g. inside Shipwreck Survivor). If this is ever implemented in the Editor, it will be necessary to additionally redo the processing logic in a number of places: currently, if the hero's artifacts bag is full, when visiting such objects, hero receives money instead of the artifact, but he will not receive these money if his artifacts bag is not full, but he just already has a Spell Book.


ihhub:
I am moving this to 1.1.1 release as it is dangerous to push changes just a day before the release.

ihhub:
Hi @oleg-derevenetz , I've checked the code. Is it correct that now neither human nor AI hero can pickup Magic Book if they have one?

oleg-derevenetz:
Hi @ihhub

> Is it correct that now neither human nor AI hero can pickup Magic Book if they have one?

You see, in the `master` branch, both of them cannot actually pick it up, because `Heroes::PickupArtifact()` has corresponding checks and will not allow to do this. But nevertheless, in the `master` branch an AI-controlled hero will endlessly try to do this until no movement points left - with side effects: e.g. if the Spell Book is guarded, it also will endlessly fight the guards.

ihhub (approved):

ihhub:

PR #8703, opened 2024-05-21
Add missing Sphinx part

ihhub:
Now we can finally add new content with ease.
All credits for **PushPop** from Discord.

image


src/fheroes2/agg/icn.h
```diff
@@ -1116,6 +1116,8 @@ namespace ICN
         BUTTON_LOAD_MAP_EVIL,
         BUTTON_LOAD_MAP_GOOD,
 
+        OLD_ADVENTURE_OBJECT_PARTS,
```
Districh-ru, line 1119: Could you please explain what this is for? :)

Districh-ru (changes requested):
Hi, @ihhub.
The fixed Sphinx looks good! Thanks, **PushPop**!
I left one question regarding the need of adding an entry to the ICN enumeration. Could you please check it when you have time?

src/fheroes2/agg/icn.h
ihhub: Oops, it's a leftover from my initial implementation.

Districh-ru (approved):

PR #8742, opened 2024-05-25
Android: implement the file manager for FH2M maps

### LINKED ISSUE #8730: Add an ability to import and export fh2m maps on Android devices
### Preliminary checks

Many players express frustration than they cannot copy maps between devices if they use Android device.

We should expand functionality of **fh2Toolset** to make this happen.

### Additional info

_No response_

oleg-derevenetz:
close #8730

Nothing particularly complicated, Save File Manager-like UI. Only FH2M files are supported - maps in other formats could not be created by our Editor, they are "import-only", the "Extract HoMM2 assets" button in the Toolset itself successfully copes with this.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -85,9 +85,9 @@ dependencies {
     implementation project(':isotools')
     implementation project(':sdl2')
 
-    implementation group: 'commons-io', name: 'commons-io', version: '2.15.1'
+    implementation group: 'commons-io', name: 'commons-io', version: '2.16.1'
 
-    implementation 'com.google.android.material:material:1.11.0'
+    implementation 'com.google.android.material:material:1.12.0'
 }
 
 tasks.register('copyH2D', Copy) {
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     android:installLocation="auto">
 
     <!-- OpenGL ES 2.0 -->
@@ -46,7 +47,8 @@
             android:process=":GameActivityProcess"
             android:screenOrientation="sensorLandscape"
             android:taskAffinity="org.fheroes2.GameTask"
-            android:theme="@android:style/Theme.NoTitleBar.Fullscreen">
+            android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
+            tools:ignore="DiscouragedApi,UnusedAttribute">
 
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -81,5 +81,13 @@
             android:label="@string/activity_save_file_manager_label"
             android:taskAffinity="org.fheroes2.ToolsetTask"
             android:theme="@style/Theme.MaterialComponents.DayNight.NoActionBar" />
+
+        <activity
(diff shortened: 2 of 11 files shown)


ihhub:
Hi @oleg-derevenetz , I tested the build and I found several issues:
1. No mentioning that maps must be in a zip format. Since we export them in a zip format then the same format is used for import.
Screenshot_20240529-205222
2. Cannot import native .fh2m files:
Screenshot_20240529-205327
3. Some zip files are not detected (I assume it is a Java package issue):
Screenshot_20240529-210227
but these files are absolutely valid:
Screenshot_20240529-210532

oleg-derevenetz:
Hi @ihhub

> No mentioning that maps must be in a zip format. Since we export them in a zip format then the same format is used for import.

The logic is currently absolutely the same as for importing/exporting save files. They are also imported & exported as ZIP archives and you cannot import a save file as is. As far as I understand, users do not encounter difficulties with this.

> Some zip files are not detected (I assume it is a Java package issue):

This file selection window is a system Activity that is called by our app. It's hard to say what it doesn't like about these archives. Can they be opened by the Save File Manager or using the assets extraction button?

ihhub:
@oleg-derevenetz , something is strange as I used full-set.zip file to install a full version of the game. After that I am not able to even select this file anymore. I was downloading them from my own Google Drive.

One thing we definitely need to do is to disable save and map related buttons if we cannot start the game.

oleg-derevenetz:
@ihhub

> something is strange as I used full-set.zip file to install a full version of the game. After that I am not able to even select this file anymore.

So you cannot select this file anymore even using the "Extract HoMM2 Assets" button? It looks like your Android goes slightly mad :) It looks like your Android doesn't allow you to choose files that are NOT created by your app, but this is nonsense, since this system Activity is designed to provide the application with access to arbitrary files (and, as I understand, this initially worked).

> One thing we definitely need to do is to disable save and map related buttons if we cannot start the game.

Are there any specific reasons for it? Because technically these things are totally unrelated: you can import save files and F2HM maps even if no game assets are extracted yet.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for this feature!

oleg-derevenetz:
Hi @ihhub

> Cannot import native .fh2m files:

I tried to add such an opportunity for the sake of interest, but it doesn't work so far. The problem is that, when calling the file chooser activity (or, more precisely, just the "chooser activity", because not just files can be selected in general), you need to pass not a template(s) for file names, but a set of MIME types. That's a problem, because there is, obviously, no standard, well-known MIME type for our fh2m files. I tried to somehow "register" it in various ways (mostly utilizing the `intent-filter` in the app manifest), but none of the methods worked properly so far, those custom MIME types are not recognized by the system chooser activity (probably because it's a system one and it doesn't care about MIME types registered by the 3-rd party apps, like fheroes2).

ihhub:
Hi @oleg-derevenetz , thank you so much for this comprehensive research.

We can come up with another solution in the future. At the moment I don't have any good solution except the one to somehow make zip files within the Editor but this seems odd to have.

oleg-derevenetz:
Hi @ihhub

> We can come up with another solution in the future. At the moment I don't have any good solution except the one to somehow make zip files within the Editor but this seems odd to have.

In general, importing/exporting something as a ZIP file is commonly used on Android (mostly because of all this pain with MIME types, and difficulties with getting "real" file names when using choosers, I believe, because URIs issued by choosers may have nothing to do with the original file names, as well as `DISPLAY_NAME`s returned by `ContentResolver` for these URIs), so it shouldn't be something non-ordinary for a regular user. At least AFAIK no one complained about a similar mechanism with save files. I think creating a ZIP archive, if necessary, does not pose a problem for the user, widespread OS have built-in tools for this, there is no point in embedding this functionality into the editor.

PR #8768, opened 2024-05-30
Fix hero editing assertion

ihhub:
Hero color must be set at first.


Districh-ru (approved):
This fix works fine.
But in the future we should not call `Heroes::Scout()` while we are in editor: there is nothing to scout. :)

PR #8788, opened 2024-05-31
Put the Graveyard object to the correct category, "hide" the ugly versions of Graveyard sprites

### LINKED ISSUE #8784: [Scenario Editor] Graveyard is in the wrong category
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the graveyard is in the wrong category.
The graveyard is in the "Dwelling" list when it should be in the "Adventure Object" list.

In fheroes2:

Graveyard wrong list 001

Graveyard wrong list 002

### Save file

Not needed.

### Additional info

_No response_

oleg-derevenetz:
fix #8784

Also related to #8785.

However, I am not sure about the logic of ordering in the objects list (did I put the Graveyard on the list to the correct place?).

### DIFF
--- a/src/fheroes2/maps/map_format_info.cpp
+++ b/src/fheroes2/maps/map_format_info.cpp
@@ -34,10 +34,60 @@ namespace
     // This value is set to avoid any corrupted files to be processed.
     const size_t minFileSize{ 128 };
 
-    const uint16_t minimumSupportedVersion{ 2 };
+    constexpr uint16_t minimumSupportedVersion{ 2 };
 
     // Change the version when there is a need to expand map format functionality.
-    const uint16_t currentSupportedVersion{ 2 };
+    constexpr uint16_t currentSupportedVersion{ 3 };
+
+    void convertFromV2ToV3( Maps::Map_Format::MapFormat & map )
+    {
+        static_assert( minimumSupportedVersion <= 2, "Remove this function." );
+
+        if ( map.version > 2 ) {
+            return;
+        }
+
+        for ( Maps::Map_Format::TileInfo & tileInfo : map.tiles ) {
+            for ( Maps::Map_Format::TileObjectInfo & objInfo : tileInfo.objects ) {
+                if ( objInfo.group == Maps::ObjectGroup::ADVENTURE_DWELLINGS ) {
+                    switch ( objInfo.index ) {
+                    case 17: // Graveyard, grass terrain, ugly version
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @oleg-derevenetz , according to
https://github.com/ihhub/fheroes2/blob/829cb34ef4cea3f7f7ec5ea0920f47a07beb798f/src/fheroes2/maps/map_object_info.h#L120

we simply cannot move objects from one place to another because it breaks map format.

What we need to do is:
- bump the current map version:

https://github.com/ihhub/fheroes2/blob/829cb34ef4cea3f7f7ec5ea0920f47a07beb798f/src/fheroes2/maps/map_format_info.cpp#L40

- add migration code from version 2 to version 3 (`bool loadFromStream( StreamBase & msg, MapFormat & map )` method, after this line:
https://github.com/ihhub/fheroes2/blob/829cb34ef4cea3f7f7ec5ea0920f47a07beb798f/src/fheroes2/maps/map_format_info.cpp#L257

we can call a function simply `migrateV2toV3()` or something like this, and there we need to change **objects** for each tile and more specifically their group and their index.

- we **must** preserve all objects as they might appear in the existing maps created with version 2. I suggest to put the objects at the end of the list. If the alternative graveyards are not pleasant for looks we can move them after the normal ones.

Please let me know if you have any questions about this.

oleg-derevenetz:
> I suggest to put the objects at the end of the list.

I would not like to put Graveyards after Barriers and Traveller's Tents, so I left them at the beginning. However, as for "ugly" versions of Graveyards, I put them at the end to hide them in a way.

ihhub (approved):

Districh-ru (approved):
The Editor gives the ability to choose any Graveyard appearance and do not force the map maker to choose one variant instead of the other. The map maker may select any Graveyard that better fits his map. :)

oleg-derevenetz:
> The map maker may select any Graveyard that better fits his map. :)

I have a suspicion that these crude Graveyards are just mock-ups accidentally left from some alpha versions of the game. They are frankly poorly drawn (totally not in the style of the game as a whole), not used in the original editors and do not appear in any known map. I don't really think there's a need to encourage map makers to use them. Unfortunately, they are apparently no longer so easy to delete by converting them into "proper" Graveyards, because "proper" Graveyards are larger objects than those mock-ups.

ihhub:

PR #8794, opened 2024-06-01

oleg-derevenetz:
Close some technical debts.


ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for keeping the code in a good shape!

PR #8806, opened 2024-06-04
Move an existing object in front while trying to place an exact one on the same tile

ihhub:
If the same object is being placed then move the existing one on top of all other objects.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1847,9 +1847,7 @@ namespace Interface
             objectIndex = type;
         }
 
-        for ( size_t i = 0; i < objects.size(); ++i ) {
-            auto & object = objects[i];
-
+        for ( auto & object : objects ) {
             if ( object.group == groupType && object.index == static_cast<uint32_t>( objectIndex ) ) {
                 if ( object.id == Maps::getLastObjectUID() ) {
                     // Just do nothing since this is the last object.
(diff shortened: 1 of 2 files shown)


LeHerosInconnu:
Related discussions: https://github.com/ihhub/fheroes2/discussions/8749, https://github.com/ihhub/fheroes2/discussions/8746.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1794,4 +1815,129 @@ namespace Interface
...
+    }
+
+    void EditorInterface::_updateObjectUID( const uint32_t oldObjectUID, const uint32_t newObjectUID )
+    {
+        for ( auto & tile : _mapFormat.tiles ) {
+            for ( size_t i = 0; i < tile.objects.size(); ) {
+                if ( tile.objects[i].id == oldObjectUID ) {
+                    tile.objects[i].id = newObjectUID;
+
+                    if ( i != tile.objects.size() - 1 ) {
```
Districh-ru, line 1932: Should we cache `tile.objects.size() - 1`?

Districh-ru (approved):
Hi, @ihhub.
Maybe we should rename this PR to, in example: Allow a placed object to be moved to front by placing it a second time on its main tile.
What do you think? :)

I also left one optional suggestion.

src/fheroes2/editor/editor_interface.cpp
ihhub: In this case the benefit of caching is neglectable since we are going to make an extra one operation for the sake of saving one operation.

PR #8813, opened 2024-06-07
Make default knowledge value for random hero as 1

### LINKED ISSUE #8750: [Scenario Editor] The default Knowledge value should be "1", not "0"
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the default Knowledge value should be "1", not "0".

In fheroes2:

Default knowkedge value 1 01

### Save file

Not needed.

### Additional info

_No response_

ihhub:
Also, limit minimum knowledge value to 1 rather than 0.

close #8750


src/fheroes2/heroes/skill.h
```diff
@@ -203,10 +202,10 @@ namespace Skill
...
         friend StreamBase & operator>>( StreamBase &, Primary & );
 
-        int attack;
-        int defense;
-        int power;
-        int knowledge;
+        int attack{ 0 };
+        int defense{ 0 };
+        int power{ 0 };
+        int knowledge{ 0 };
```
Districh-ru, line 208: Should we also set `1` as the default value for power and knowledge skills then?

Districh-ru (changes requested):
Hi, @ihhub.
In this PR we only clamp the "modified" knowledge value, but the base `knowledge` can still be 0.
We should not allow to enter a value lower than 1 for knowledge in hero edit dialog and in battle only dialog.


https://github.com/ihhub/fheroes2/assets/113276641/3b76f6c2-ee2a-4741-8bd2-8d6c11884520

Branikolog (approved):
Fix works as intended. I'm not sure, whether it should be possible to set knowledge to zero... Maybe, mapmakers would like to make a hero without spell points though.
Att and Def are ok to be 0 as default since many magic heroes have zero values from start

Districh-ru (approved):

PR #8815, opened 2024-06-08
Code cleanup for LocalEvent (part 3)

ihhub:
Simple method renaming.

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -292,7 +292,7 @@ namespace
         display.render();
 
         while ( le.HandleEvents() ) {
-            le.MousePressLeft( buttonOkay.area() ) ? buttonOkay.drawOnPress() : buttonOkay.drawOnRelease();
+            le.isMouseLeftButtonPressedInArea( buttonOkay.area() ) ? buttonOkay.drawOnPress() : buttonOkay.drawOnRelease();
 
             bool redrawScreen = false;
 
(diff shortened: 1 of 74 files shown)


src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -292,7 +292,7 @@ namespace
         display.render();
 
         while ( le.HandleEvents() ) {
-            le.MousePressLeft( buttonOkay.area() ) ? buttonOkay.drawOnPress() : buttonOkay.drawOnRelease();
+            le.isMouseLeftButtonPressedInArea( buttonOkay.area() ) ? buttonOkay.drawOnPress() : buttonOkay.drawOnRelease();
```
Districh-ru, line 295: Hi, @ihhub.
While changing all such lines in this PR should we also update them to:
```cpp
buttonOkay.drawOnState( le.isMouseLeftButtonPressedInArea( buttonOkay.area() ) );
```
?

src/fheroes2/battle/battle_interface.h
```diff
@@ -292,7 +292,7 @@ namespace Battle
             return _interfacePosition;
         }
 
-        fheroes2::Point GetMouseCursor() const;
+        fheroes2::Point getMouseCursorPos() const;
```
Districh-ru, line 295: Should we rename this method to, in example, `getMouseCursorPosOnBattlefield()`?
Not to confuse with the same method in LocalEvent class that gives the absolute cursor position and the current one gives the position relatively to the battlefield interface area.

src/fheroes2/battle/battle_dialogs.cpp
ihhub: We can do this change in a separate pull request as this will require significant amount of work. The purpose of this change to make the code more readable for the future changes.

src/fheroes2/battle/battle_interface.h
ihhub: I've renamed it into `getRelativeMouseCursorPos()` since this method relates to the battlefield.

src/fheroes2/battle/battle_dialogs.cpp
ihhub: We have 454 places of this code :) It's better to do this separately.

Districh-ru: Sure!

Districh-ru (approved):

PR #8818, opened 2024-06-10
Add more checks for SDL functions

ihhub:
To make sure that we properly handle everything and to avoid infinite loops in cases of error.

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -426,7 +426,10 @@ namespace
             fheroes2::Cursor::show( enable );
 
             if ( !_emulation ) {
-                SDL_ShowCursor( _show ? SDL_ENABLE : SDL_DISABLE );
+                const int returnCode = SDL_ShowCursor( _show ? SDL_ENABLE : SDL_DISABLE );
+                if ( returnCode < 0 ) {
+                    ERROR_LOG( "Failed to set cursor. The error value: " << returnCode << ", description: " << SDL_GetError() )
+                }
             }
         }
 
@@ -669,6 +672,9 @@ namespace
 
             _window = SDL_CreateWindow( "", 0, 0, resolutionInfo.gameWidth, resolutionInfo.gameHeight, 0 );
             if ( _window == nullptr ) {
+                ERROR_LOG( "Failed to create an application window of " << resolutionInfo.gameWidth << " x " << resolutionInfo.gameHeight
+                                                                        << " size. The error: " << SDL_GetError() )
+
                 clear();
                 return false;
             }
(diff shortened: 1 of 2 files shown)


src/engine/screen.cpp
```diff
@@ -1091,34 +1107,39 @@ namespace
             const uint32_t renderingFlags = renderFlags();
 
             const int driverCount = SDL_GetNumRenderDrivers();
-            for ( int driverId = 0; driverId < driverCount; ++driverId ) {
-                int returnCode = SDL_GetRenderDriverInfo( driverId, &rendererInfo );
-                if ( returnCode < 0 ) {
-                    ERROR_LOG( "Failed to get renderer driver info. The error value: " << returnCode << ", description: " << SDL_GetError() )
-                    continue;
-                }
+            if ( driverCount >= 0 ) {
```
oleg-derevenetz, line 1110: Is the situation correct when there are zero renderer drivers? I'm asking this because at line 867 the similar situation, but with video displays, is regarded as incorrect.

src/engine/screen.cpp
```diff
@@ -1206,7 +1227,13 @@ namespace
 
         void _retrieveWindowInfo()
         {
-            const int32_t displayIndex = SDL_GetWindowDisplayIndex( _window );
+            assert( _window != nullptr );
+
+            const int displayIndex = SDL_GetWindowDisplayIndex( _window );
+            if ( displayIndex < 0 ) {
+                ERROR_LOG( "Failed to get window display index. The error value: " << displayIndex << ", description: " << SDL_GetError() )
+            }
```
oleg-derevenetz, line 1235: Shouldn't we `return` here and below (at line 1239)? Because if `displayIndex` is incorrect, then `SDL_GetCurrentDisplayMode()` call most likely should also fail, and if it fails, then `displayMode` most likely will remain uninitialized and will contain some garbage, so it makes no sense to use its values later (at least the `SDL_GetCurrentDisplayMode()` documentation contains no guarantees regarding `displayMode` initialization in case of failure, so it's not a part of the API contract).

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions, please take a look when you have time.

src/engine/screen.cpp
ihhub: I did this based on the documentation. `SDL_GetNumRenderDrivers` states:
```cpp
 * \returns a number >= 0 on success or a negative error code on failure; call
 *          SDL_GetError() for more information.
```
while `SDL_GetNumVideoDisplays`:
```cpp
 * \returns a number >= 1 or a negative error code on failure; call
 *          SDL_GetError() for more information.
```

src/engine/screen.cpp
ihhub: I've added return calls here.

src/engine/screen.cpp
oleg-derevenetz: So there really can be zero rendering drivers? Well, OK then.

oleg-derevenetz (approved):

PR #8821, opened 2024-06-10
Speed up save file loading and saving operations

ihhub:
We must avoid opening a file twice for reading and writing as it is unsafe and inefficient. Additionally, I added extra checks for stream operation failure.

### DIFF
--- a/src/engine/zzlib.h
+++ b/src/engine/zzlib.h
@@ -48,8 +48,9 @@ namespace Compression
     // it to the end of the buffer. Returns true on success or false on error.
     bool readFromFileStream( StreamFile & fileStream, StreamBuf & output );
 
-    // Zips the contents of the buffer from the current read position to the end of the buffer and
-    // writes it to the specified file stream. Returns true on success and false on error.
+    // Zips the contents of the buffer from the current read position to the end of the buffer and writes
+    // it to the specified file stream. The current read position of the buffer does not change. Returns
+    // true on success and false on error.
     bool writeIntoFileStream( StreamFile & fileStream, StreamBuf & data );
 
     fheroes2::Image CreateImageFromZlib( int32_t width, int32_t height, const uint8_t * imageData, size_t imageSize, bool doubleLayer );
--- a/src/engine/zzlib.h
+++ b/src/engine/zzlib.h
@@ -32,6 +32,7 @@
 #include "image.h"
 
 class StreamBuf;
+class StreamFile;
 
 namespace Compression
 {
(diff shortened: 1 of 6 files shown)


src/fheroes2/game/game_io.cpp
```diff
@@ -255,11 +264,13 @@ fheroes2::GameMode Game::Load( const std::string & filePath )
...
-
-    if ( sb.fail() || endOfDataMarker != SAV2ID3 ) {
-        DEBUG_LOG( DBG_GAME, DBG_WARN, "File " << filePath << " is corrupted" )
+    dataStream >> endOfDataMarker;
+    if ( dataStream.fail() ) {
+        showGenericErrorMessage();
+        return fheroes2::GameMode::CANCEL;
+    }
 
+    if ( endOfDataMarker != SAV2ID3 ) {
```
Districh-ru, line 273: Can we combine these two checks?

```suggestion
    if ( dataStream.fail() || ( endOfDataMarker != SAV2ID3 ) ) {
```

Districh-ru (approved):
Hi, @ihhub, all works OK: standard and campaign games are properly saved and loaded. The same is for the high scores.
I left one optional suggestion.

src/engine/zzlib.h
```diff
@@ -44,15 +44,14 @@ namespace Compression
...
+    // Reads & unzips the zipped chunk from the specified file stream and appends
+    // it to the end of the buffer. Returns true on success or false on error.
+    bool readFromFileStream( StreamFile & fileStream, StreamBuf & output );
 
-    // Zips the contents of the buffer from the current read position to the end of the buffer and
-    // writes (or appends) it to the specified file. The current read position of the buffer does
-    // not change. Returns true on success and false on error.
-    bool writeFile( StreamBuf & input, const std::string & fn, const bool append = false );
+    // Zips the contents of the buffer from the current read position to the end of the buffer and writes
+    // it to the specified file stream. The current read position of the buffer does not change. Returns
```
oleg-derevenetz, line 52: I would like to emphasize here that the current read position of `data` will not change after calling this method (even though the `data` buffer will be completely "read").

src/fheroes2/game/game_io.cpp
```diff
@@ -127,24 +127,31 @@ bool Game::Save( const std::string & filePath, const bool autoSave /* = false */
...
 
     // End-of-data marker
-    sb << SAV2ID3;
+    dataStream << SAV2ID3;
+    if ( dataStream.fail() ) {
+        return false;
+    }
 
-    if ( sb.fail() || !Compression::writeFile( sb, filePath, true ) ) {
+    if ( !Compression::writeIntoFileStream( fileStream, dataStream ) ) {
```
oleg-derevenetz, line 154: Why not:

```cpp
if ( dataStream.fail() || !Compression::writeIntoFileStream( fileStream, dataStream ) ) {
```

?

src/fheroes2/game/game_io.cpp
```diff
@@ -327,7 +338,7 @@ bool Game::LoadSAV2FileInfo( std::string filePath, Maps::FileInfo & fileInfo )
     fileInfo = std::move( header.info );
     fileInfo.filename = std::move( filePath );
 
-    return true;
+    return !fs.fail();
```
oleg-derevenetz, line 341: I believe we should check for I/O failure (and return `false`) a bit earlier, before checking the `header.gameType` and moving "read" data to the `fileInfo`, what do you think?

ihhub: Sure! It makes sense.

src/fheroes2/game/game_io.cpp
ihhub: It is done.

src/fheroes2/game/game_io.cpp
ihhub: It is done.

src/engine/zzlib.h
ihhub: I added back the sentence in the comment for this function.

oleg-derevenetz (approved):
I reverted the 44ccceb since I already introduced the same changes in 9fba1e0, so with your changes, the same phrase was repeated twice :)

PR #8822, opened 2024-06-10
Hungarian language small update 05

big4billy:
Some corrections and shrinking texts to fit


ihhub:
@big4billy , many thanks for the update!

PR #8823, opened 2024-06-12
Add an ability to edit Rumors for maps

### LINKED ISSUE #8714: Add a UI window to add rumors for map in the Editor
### Preliminary checks

There is no way at the moment to do this.

Add a window to be able to edit custom rumors.

### Additional info

_No response_

ihhub:
close #8714

image

image

### DIFF
--- a/src/fheroes2/editor/editor_rumor_window.cpp
+++ b/src/fheroes2/editor/editor_rumor_window.cpp
@@ -223,7 +223,7 @@ namespace Editor
 
             if ( le.MouseClickLeft( buttonAdd.area() ) ) {
                 std::string newRumor;
-                if ( Dialog::inputString( _( "Rumor:" ), newRumor, {}, longestRumor, false, true ) ) {
+                if ( Dialog::inputString( _( "Rumor:" ), newRumor, {}, longestRumor, true, true ) ) {
                     if ( std::any_of( rumors.begin(), rumors.end(), [&newRumor]( const auto & rumor ) { return rumor == newRumor; } ) ) {
                         fheroes2::showStandardTextMessage( _( "Rumor" ), _( "This rumor already exists in the list." ), Dialog::OK );
                         continue;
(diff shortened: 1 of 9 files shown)


src/fheroes2/editor/editor_rumor_window.cpp
```diff
@@ -0,0 +1,290 @@
...
+        std::unique_ptr<fheroes2::ImageRestorer> _listBackground;
+    };
+}
+
+namespace Editor
+{
+    bool openRumorWindow( std::vector<std::string> & rumors )
+    {
+        // Remove all empty rumors.
+        rumors.erase( std::remove_if( rumors.begin(), rumors.end(), []( const auto & rumor ) { return rumor.empty(); } ), rumors.end() );
```
Districh-ru, line 132: Is there possible to add an empty rumor? If not then there is no need to erase them.
Also should we place here an assertion instead? (and to the 228 and 248 lines)

Districh-ru (approved):
Hi @ihhub, rumors works fine!
I left one optional question/suggestion.

Branikolog:
Hi, @ihhub 
I've found out that after pressing OK or Cancel in the rumor creating dialog there's a short delay, before dialog disappears.
image

I do also believe, that tavern window is not really large to keep long texts. Currently, It seems that I can create really long (is there any restrictions at all?) message, that is going not to fit the tavern window.
image
In this case there's no way to even check, what is hidden after "..."

src/fheroes2/editor/editor_rumor_window.cpp
ihhub: We can add an assertion. This check is more a prevention from people trying to "hack" the format.

ihhub:
Hi @Branikolog , I've reduced the sentence limit to 200 characters and also expanded the window:
image

image

Branikolog:
Hi, @ihhub !

>and also expanded the window:

Making window larger does not fix the problem with text editing and looks not really nice in my opinion.
image

Can we use here a window with larger field to edit rumors? Currently it is impossible to read and edit long messages.
Something similar to map description for instance:
image

The problem with delayed window shutdown is still on.

ihhub:
Hi @Branikolog , I've addressed your comments.

PR #8827, opened 2024-06-12
Properly check the PoL resources in order to enable the Editor

ihhub:
Reported by some users who cannot access the Editor but they see the button in the Main Menu which is not interactive.


Districh-ru (approved):
👍 
Yes, for Editor we need original PoL AGG resources, but not the campaign maps which could be deleted by some users.

Issue #8834, opened 2024-06-14
An event is not triggering if a player is boarding a ship


Killbrum:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Greetings!
Recently, I've encountered a weird case, and mainly, I assume that it's a bug. On Pirates Utopia map there is the event on the bottom right corner (the last pixel). It has to be triggered when some one teleports to the closes tile, spawns a boat, and onboards the boat (tricky case). Still, in my case, nothing happened. I assume because this logic is not handled (like even can occur if someone boarding a boat). PC hasn't reach that place at all.

I will try to upload correct save file soon

![event](https://github.com/ihhub/fheroes2/assets/7841390/86e7e6b1-7d9f-45fc-9890-99844fb78d82)

### Save file

So far no save game, but I will try to add

oleg-derevenetz:
Currently in fheroes2, both event and boat are "map objects", and there cannot be more than one object on the same tile at the same time (excluding the hero objects, because they are "special"). AI will not try to summon a boat here, because there is already an object on that tile. Logic for human players is a bit different, and human players are allowed to summon a boat on that tile, but the result is that the boat object "overwrites" the event object, so there is no event object on that tile anymore after the boat summoning.

oleg-derevenetz:
The fact that only one object can be placed on a tile at a time (even in the case of "invisible" objects, such as events) is the part of the core game mechanics. In the OG editor, you cannot place an event on a tile with the boat, and vice versa:

https://github.com/ihhub/fheroes2/assets/32623900/2a5e0e6f-5a0a-4e8e-a160-2a56f1f57305

Moreover, in the original game, the hero cannot disembark from the boat in the place where an event object is located:

https://github.com/ihhub/fheroes2/assets/32623900/2e28492b-ab3e-4c78-8999-461bd22a8822

This is because the hero can only disembark on tiles where there is another "invisible" object called a "coast object", and you cannot have both the "event object" and "coast object" on the same tile at the same time. Therefore, I am inclined to believe that a human player simply should not be able to summon a boat to the tile on which the event object is located - just as AI cannot do it now.

zenseii:
[...]

PR #8835, opened 2024-06-14
Bring the logic of the Summon Boat spell for the human player in line with its logic of operation in the case of AI

oleg-derevenetz:
Related to #8834

This PR just makes sure that `OBJ_EVENT` objects will not be overwritten when summoning a boat. The problem (and the need) to replicate the behavior of the original game in this case is still questionable.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for the update!

PR #8858, opened 2024-06-22
Add support of editing daily events within the Editor

### LINKED ISSUE #8713: Add a window to edit daily events in the Editor
### Preliminary checks

At the moment there is no way to edit them.

Simple as stated - add a new window for the editing events.

### Additional info

_No response_

ihhub:
- add support for editing daily events within the Editor
- allow to enter negative values for resources (Adventure Map events and daily events). Use **-** (minus) on your keyboard to swap between negative and positive values
- fix cases when changing the value by mouse and then modifying it by keyboard resets the value
- fix incorrect UI element layout for entering values. Now it follows the same layout as in the monster hiring window
- fix text rendering in Adventure Map event window
- add checks to skip events that will never trigger during gameplay

image

image

image

close #8713


Branikolog:
Hi, @ihhub 
I'm testing this PR now.
I managed to initiate an application crash when trying to type "space"+"Enter"(new line)+confirm in the event text field.
The crash occurs for all other dialogs of the following type:
image

Should I create a separate issue for it?

Branikolog:
"-" mark is rendered out of the field range for large numbers:
image

ihhub:
Hi @Branikolog , please create a separate issue regarding the text editing. This PR is way too big.

Branikolog:
Hi, @ihhub 
Now it is better, but some digits are still over the field frames:
image
Maybe we can just place "-" sign to the left of the field, so it never overlaps UI elements? Maybe we can place it into a special checkbox or something like that?

ihhub:
@Branikolog , we can simply don't allow to enter less than `-99999` and no more than `999999` resources. This is UI issue and should be addressed separately.

Branikolog:
It would be nice to show the exact day (month, week, day) near this counter. 
image
Everything else seem to work properly. 👍

ihhub:
> It would be nice to show the exact day (month, week, day) near this counter. image Everything else seem to work properly. 👍

This is an improvement for the future.

Branikolog (approved):
Great job!

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -722,4 +722,43 @@ namespace fheroes2
...
+            // No key is pressed.
+            return false;
+        }
+
+        if ( le.isKeyPressed( fheroes2::Key::KEY_BACKSPACE ) ) {
+            value = value / 10;
+            return true;
+        }
+
+        if ( minimum < 0 && value != 0 && le.isKeyPressed( fheroes2::Key::KEY_MINUS ) ) {
```
Districh-ru, line 740: Should we also process the keypad minus?
```suggestion
        if ( minimum < 0 && value != 0 && ( le.isKeyPressed( fheroes2::Key::KEY_MINUS ) || le.isKeyPressed( fheroes2::Key::KEY_KP_MINUS ) ) ) {
```

src/fheroes2/editor/editor_daily_event_spec_window.cpp
```diff
@@ -0,0 +1,367 @@
...
+                    display.render( resourceRoi );
+                    break;
+                }
+            }
+
+            if ( le.MouseClickLeft( messageRoi ) ) {
+                std::string temp = eventMetadata.message;
+
+                if ( Dialog::inputString( _( "Message:" ), temp, {}, 200, true, true ) ) {
+                    eventMetadata.message = std::move( temp );
```
Districh-ru, line 330: Should we limit the maximum text lines?
изображение

Districh-ru (changes requested):
Hi, @ihhub.
I left two comments. Could you please take a look when you have time?

src/fheroes2/editor/editor_daily_event_spec_window.cpp
ihhub: I believe this is not a trivial problem since the window where we actually type has different width leading to different results. And it is also true that an event window which pops up in the game has also different width.

The shown example is an extreme case and I agree that this is a bug. However, this is the problem for many other places and we should fix it in a separate pull request.

src/fheroes2/gui/ui_tool.cpp
ihhub: It is fixed.

Districh-ru (approved):

PR #8859, opened 2024-06-22
Update development instructions

celliott113:
Add step in Windows to ensure the user clones the project before proceeding onto next steps.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -7,6 +7,7 @@ Please follow the instructions below to be able to build the project from source
 <a name="windows"></a>
 ### Windows
 
+* You will need to clone the `fheroes2` project.
 * Go to the directory `script/windows` and run the file `install_packages.bat`. This script will install all the
   required packages to build the project using Microsoft Visual Studio.
 * If you have the original game, then copy the subdirectories `ANIM`, `DATA`, `MAPS` and `MUSIC` (some of them may
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -4,10 +4,11 @@
 
 Please follow the instructions below to be able to build the project from source:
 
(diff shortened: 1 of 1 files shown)


docs/DEVELOPMENT.md
```diff
@@ -7,6 +7,7 @@ Please follow the instructions below to be able to build the project from source
 <a name="windows"></a>
 ### Windows
 
+* You will need to clone the `fheroes2` project.
```
ihhub, line 10: This is true not only on Windows but on all platforms.

ihhub (changes requested):

docs/DEVELOPMENT.md
celliott113: You are correct, I have adjusted by removing from Windows and adding to the top of the doc to encompass all platforms.

docs/DEVELOPMENT.md
```diff
@@ -4,6 +4,8 @@
 
 Please follow the instructions below to be able to build the project from source:
 
+<b>Note:</b> You will need to clone the `fheroes2` project before following the below instructions.
```
oleg-derevenetz, line 7: There is no need to use HTML tags if there are corresponding Markdown tags:

```suggestion
**Note:** You will need to clone the `fheroes2` project before following the instructions below.
```

oleg-derevenetz (changes requested):
Hi @celliott113 I just left one suggestion, please take a look when you have time.

celliott113 (commented):
Updated with @oleg-derevenetz recommendations. Necessary to maintain document consistency within the markdown language.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@celliott113 , thank you very much for your contribution to the project!

PR #8873, opened 2024-06-24
Remove CONFIG button in Main Menu's New Game group

ihhub:
I believe that CONFIG button is not suitable for New Game group within the Main Menu. It is simply not related to New Game at all. We already have the Door as a place for the settings as well as inside the Adventure Map, Editor and Battles. I am suggesting to remove it.

This is controversial and you might disagree with this.

image


oleg-derevenetz (commented):
Technically this PR looks OK, but I personally doubt that there is a real need to remove this button. Door as a button looks a bit obscure.

ihhub:
Closing this PR for now as it has no real value on at the moment.

PR #8877, opened 2024-06-25
Increase most of object selection windows by 50 pixels

ihhub:
This is extremely critical for low resolutions. Most of Android devices use `xxxx x 480` resolution. This is how it looks before the changes:
image
image

And this is after the changes:
image
image


Districh-ru (approved):
Hi, @ihhub, LGTM!
I have just one optional suggestion:
We can declare a `const int32_t dialogHeightDeduction = 150;` with a comment that this value is deducted from the screen height to get the dialog height. And use it in all changed lines.

Branikolog:
Hi, @ihhub !

I love this change! Most of windows became more informative now with this change.

I'd also would love to see more lines for some windows, like rocks too.
This is how they are looking currently:
image

I've noticed, that window size was increased for water objects list:
PR:
image

But it still shows the same number of objects at once:
Master:
image

If this was made on purpose, is it possible to align objects in the middle of the window at least?

ihhub:
Hi @Branikolog , the requested by you change is hard to achieve because of the uniqueness of objects. We can do this in a separate pull request.

Issue #8717, opened 2024-05-23
Implement missing victory and losing conditions in the Editor


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

The Editor has only normal winning and losing conditions at the moment while the original Editor has more than this.

### Describe the possible solution

Implement missing conditions.

PR #8879, opened 2024-06-25
Add first special map conditions within the Editor

ihhub:
relates to #8717

Implement Gold Accumulation Victory condition and Run of Time Loss condition.

Please take a note that this is a basic version at the moment. We are going to improve the UI later.

image

### DIFF
--- a/src/fheroes2/editor/editor_map_specs_window.cpp
+++ b/src/fheroes2/editor/editor_map_specs_window.cpp
@@ -373,6 +373,9 @@ namespace
 
                 _goldAccumulationValue.setOffset( uiOffset );
                 _goldAccumulationValue.draw( output );
+
+                const fheroes2::Text text( _( "Gold:" ), fheroes2::FontType::normalWhite() );
+                text.draw( uiOffset.x - text.width() - 5, roi.y + ( fheroes2::ValueSelectionDialogElement::getArea().height - text.height() ) / 2 + 2, output );
                 break;
             }
             default:
@@ -474,6 +477,9 @@ namespace
 
                 _outOfTimeValue.setOffset( uiOffset );
                 _outOfTimeValue.draw( output );
+
+                const fheroes2::Text text( _( "Days:" ), fheroes2::FontType::normalWhite() );
+                text.draw( uiOffset.x - text.width() - 5, roi.y + ( fheroes2::ValueSelectionDialogElement::getArea().height - text.height() ) / 2 + 2, output );
                 break;
             }
             default:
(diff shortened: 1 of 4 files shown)


src/fheroes2/editor/editor_map_specs_window.cpp
```diff
@@ -308,51 +307,254 @@ namespace
...
+        }
+
+        bool processEvents()
+        {
+            switch ( _conditionType ) {
+            case Maps::FileInfo::VICTORY_DEFEAT_EVERYONE:
+                // No events to process.
+                return false;
+            case Maps::FileInfo::VICTORY_COLLECT_ENOUGH_GOLD:
+                return _goldAccumulationValue.processEvents();
```
Districh-ru, line 398: An optional suggestion than can be done later.
Will we allow to set value from the keyboard when, in example, the cursor is over the value?

src/fheroes2/editor/editor_map_specs_window.cpp
```diff
@@ -308,51 +307,254 @@ namespace
...
+                break;
+            case Maps::FileInfo::LOSS_OUT_OF_TIME: {
+                const fheroes2::Rect roi{ _restorer.x(), _restorer.y(), _restorer.width(), _restorer.height() };
+                const fheroes2::Point uiOffset{ roi.x + ( roi.width - fheroes2::ValueSelectionDialogElement::getArea().width ) / 2, roi.y };
+
+                _outOfTimeValue.setOffset( uiOffset );
+                _outOfTimeValue.draw( output );
+
+                const fheroes2::Text text( _( "Days:" ), fheroes2::FontType::normalWhite() );
+                text.draw( uiOffset.x - text.width() - 5, roi.y + ( fheroes2::ValueSelectionDialogElement::getArea().height - text.height() ) / 2 + 2, output );
```
Districh-ru, line 485: When you have to set for some months, weeks and days you need to calculate how many days it should be.
We can try to make life easier for mapmakers. :)
We could separate Days (0-6), Weeks (0-3), Months (0-11), Years setting.
Or simply display Weeks, Months and Years calculated values according to the given Days value.
And it can be done later. :)

Districh-ru (approved):
Hi, @ihhub, All works great!
I left just two optional suggestions than could help users to interact with the Editor.

src/fheroes2/editor/editor_map_specs_window.cpp
ihhub: This is UI element specific change. Let's do this separately.

ihhub:
@Branikolog , I am merging this PR as we need to speed up covering all features for the upcoming release. If you find issues with this PR please file issues for them.

PR #8883, opened 2024-06-26
Allow normal victory for "Accumulate gold" condition

ihhub:
relates to #8717

- fix rendering while changing victory or loss condition
- fix obtain artifact incorrect metadata handling
- add an option to allow normal victory for "Accumulate Gold" victory condition

image

### DIFF
--- a/src/fheroes2/editor/editor_map_specs_window.cpp
+++ b/src/fheroes2/editor/editor_map_specs_window.cpp
@@ -37,6 +37,7 @@
 #include "difficulty.h"
 #include "editor_daily_events_window.h"
 #include "editor_rumor_window.h"
+#include "editor_ui_helper.h"
 #include "game_hotkeys.h"
 #include "game_over.h"
 #include "icn.h"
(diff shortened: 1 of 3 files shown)


src/fheroes2/editor/editor_map_specs_window.cpp
```diff
@@ -317,22 +318,31 @@ namespace
             // Do nothing.
         }
 
-        void setCondition( const uint8_t conditionType, std::vector<uint32_t> & conditionMetadata )
+        void setCondition( Maps::Map_Format::MapFormat & mapFormat )
         {
-            _conditionType = conditionType;
+            _conditionType = mapFormat.victoryConditionType;
+
+            _isNormalVictoryAllowed = false;
```
Districh-ru, line 325: Why do we force reset `_isNormalVictoryAllowed`?
In original Editor this state is stored if you change the victory condition type.

https://github.com/ihhub/fheroes2/assets/113276641/4feb7275-745a-40e4-9379-4c13aeddcd1e

src/fheroes2/editor/editor_map_specs_window.cpp
```diff
@@ -317,22 +318,31 @@ namespace
...
 
             switch ( _conditionType ) {
             case Maps::FileInfo::VICTORY_DEFEAT_EVERYONE:
                 // This condition has no metadata.
-                conditionMetadata.clear();
+                mapFormat.victoryConditionMetadata.clear();
+
+                mapFormat.allowNormalVictory = false;
+                mapFormat.isVictoryConditionApplicableForAI = false;
```
Districh-ru, line 333: Please correct me if I misunderstand. In the `setCondition()` method we get data from `mapFormat` and set the `VictoryConditionUI` members. Then why we do changes in `mapFormat`?
Should we update `mapFormat` only when the dialog is closed by OKAY button and do no changes otherwise?

Districh-ru (changes requested):
Hi, @ihhub, I left two questions. Could you please take a look in your spare time?

src/fheroes2/editor/editor_map_specs_window.cpp
ihhub: Oh, this is a leftover from development.

src/fheroes2/editor/editor_map_specs_window.cpp
ihhub: The values for other conditions are not set and being reset if you switch to other condition. I believe it is better to keep the same logic for this special variables / checkboxes. Moreover, with the current approach if I select Accumulate gold and click the checkbox, then select None and return back the checkbox is going to be unselected. This is because we don't update the map properties upon this event. Therefore, I put the code to ensure that the values are consistent.

Another reason is to verify that the values are correct for certain conditions as not all of them use these 2 values.

Districh-ru (approved):

src/fheroes2/editor/editor_map_specs_window.cpp
Districh-ru: OK)
I wanted to say that I needed some time to understand the purpose of `setCondition()` method. From the first place it uses `mapFormat` to set conditions for the UI ... but the next lines tells the this method is used to set conditions in `mapFormat` ... So is the next line ( `_isNormalVictoryAllowed = mapFormat.allowNormalVictory;` ) because we set UI using the `mapFormat` or is should be ( `mapFormat.allowNormalVictory = _isNormalVictoryAllowed ;` ) because we set conditions in `mapFormat`. :)
Maybe it is better to split this method in the future into two - the first will not modify `mapFormat` and the next one will update it.

PR #8894, opened 2024-06-29
Allow to disable any building in castles and towns

ihhub:
- allow to disable any building in castles and towns (UI work was done by @Districh-ru in the past)
- add an extra check to avoid a crash while reading a save file
- do initial code cleanup related to castles (it's still horror but this is a start)

image

image


oleg-derevenetz (approved):
Hi @ihhub I didn't find any obvious issues in this PR, there is only one question regarding the Shipyard.

src/fheroes2/castle/buildinginfo.cpp
```diff
@@ -564,84 +580,67 @@ bool BuildingInfo::DialogBuyBuilding( bool buttons ) const
...
-        if ( building == BUILD_SHIPYARD ) {
-            res = _( "Cannot build %{name}. The castle is too far away from an ocean." );
+    switch ( _status ) {
+    case BuildingStatus::NOT_TODAY:
+    case BuildingStatus::NEED_CASTLE:
+        return GetBuildConditionDescription( _status );
+
+    case BuildingStatus::BUILD_DISABLE:
+        if ( _buildingType == BUILD_SHIPYARD ) {
+            std::string res = _( "Cannot build %{name}. The castle is too far away from an ocean." );
```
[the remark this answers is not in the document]
ihhub: This is a good point. I should move the common code after this switch-case code block.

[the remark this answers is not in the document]
ihhub: I have an idea to solve this.

src/fheroes2/castle/buildinginfo.cpp
```diff
@@ -446,7 +462,7 @@ bool BuildingInfo::DialogBuyBuilding( bool buttons ) const
     const fheroes2::Text descriptionText( std::move( extendedDescription ), fheroes2::FontType::normalWhite() );
 
     // prepare requirement build string
-    std::string requirement = fheroes2::getBuildingRequirementString( castle.GetRace(), static_cast<building_t>( building ) );
+    std::string requirement = fheroes2::getBuildingRequirementString( castle.GetRace(), static_cast<building_t>( _buildingType ) );
```
Districh-ru, line 465: Hi, @ihhub. 
Maybe we shouldn't show building requirements and construction costs if the building is disabled?
изображение

ihhub: Hi @Districh-ru , I am not sure about this. On one hand, you are right, no reason to show this. On the other hand, other castles might have this building being enabled and consistency is important.

Districh-ru: Then in other castles the requirements will be shown.
I mentioned this because some player might think that if he meets all requirements then the building will be enabled. :)
In Heroes 3 the building description is left unchanged and the Requirements are replaced with with "This building is disabled"
Maybe we should do the same way?
изображение
изображение

LeHerosInconnu:
Hello everyone,

Here's a suggestion for the mixed color bar under the building name text:

In the editor, buildable but with a limitation.
damier seul 02

In the editor, fully built, taking into account the limitation (so, no upgrades possible).
damier seul 01

During the game, buildable but with a limitation.
damier seul 02

During the game, fully built, limitation reached (so, no upgrades possible).
damier seul 01

During the game, since the construction of buildings has limitations, it would be a good rule to indicate the maximum possible construction level in the right click information window, right from the base of the building.
For example, for the mage guild, even before building level one, indicate that the maximum building level is level two.

src/fheroes2/castle/buildinginfo.cpp
ihhub: Hi @Districh-ru , Shipyard still shows construction cost even being disabled due to no water around the castle. I am okay to hide requirements but keep construction cost to be consistent.

ihhub: This is how it looks after hiding the requirements:
image

Districh-ru (approved):

PR #8911, opened 2024-07-03
Allow skipping docs installation

lemenkov:
Application packaging sometimes requires to list docs explicitly in a packaging' setup file and sometimes even package some of them (namely license file) separately. In this case an extra installation step becomes unnecessary. This patch allows packager to skip it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,12 +135,14 @@ if(NOT MACOS_APP_BUNDLE)
 		FILES_MATCHING
 		PATTERN *.h2d
 		)
-if(INSTALL_DOCS)
-	install(
-		FILES docs/README.txt LICENSE changelog.txt
-		DESTINATION ${CMAKE_INSTALL_DOCDIR}
-		)
-endif(INSTALL_DOCS)
+
+	if(INSTALL_DOCS)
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @lemenkov , could you please point to well-known libraries that use such a step for CMake? As I see we are going to add an extra option just for the sake of saving space for 3 files.

lemenkov:
> Hi @lemenkov , could you please point to well-known libraries that use such a step for CMake? 

Nope, nothing comes to my mind.

ihhub:
> > Hi @lemenkov , could you please point to well-known libraries that use such a step for CMake?
> 
> Nope, nothing comes to my mind.

Then I don't understand the benefit of doing this step for CMake. I haven't seen any software that doesn't copy license files into an installation location. Please explain if something I am missing for this change.

lemenkov:
Hello, and sorry for the delayed answer. First I handled this a bit differently so I do not need this PR any more.

The issue we need to treat License data and other supplementary doc-files differently while packaging.  So I do not need them to be installed at all - I pick them up during the packaging stage. Anyway it's not a big deal and I am going to close it.

PR #8912, opened 2024-07-03
Handle the SDL events strictly in turn

### LINKED ISSUE #8910: Surface Duo - keyboard input is laggy
### Preliminary checks

### Platform

Android

### Describe the bug

On surface duo, keyboard input is slow / laggy.  I have to push 'h' or 't' , 5 or 6 times before a single click is registered.

### Save file

None

### Additional info

I suspect this is a niche issue considering it the application seems fast on my galaxy phone. I've tried multiple keyboards. It's slow with different keyboards.  And the keyboard input is fast in other applications on the surface duo.

Surface Duo is a great way to play the game portably otherwise.

oleg-derevenetz:
fix #8910


denierlexiese:
@oleg-derevenetz . I've tried it out.  First impressions, keyboard now works well. Strangely mouse has double click now.  I tried multiple mice, and they all do the same thing.  I tried a separate double detection click app, and it suggests the mouse doesn't double click.

oleg-derevenetz:
Hi @denierlexiese

> Strangely mouse has double click now. I tried multiple mice, and they all do the same thing.

That's because SDL emulates fake touch events from the mouse events, and vice versa. The latter has already been disabled, while the former has not, so each mouse press or release event generated the corresponding fake touch event. Now this behavior is disabled too, so there should be no more "double clicks", please try the latest Android artifact from this PR.

oleg-derevenetz:
Hi @denierlexiese do you still have any input issues (with keyboard, mouse or touch) with the Android binary from this PR after the latest changes (disabled generation of fake touch events from mouse events)?

denierlexiese:
@oleg-derevenetz - yes it works well now.Thank you!.  It works well even in usb-c mode with both touch, mouse, and keyboard.

oleg-derevenetz:
Hi @ihhub probably this PR should be postponed after the upcoming release - depending on your release schedule.

ihhub:
Hi @oleg-derevenetz , the release is scheduled in 3 days so it would be safer to merge this PR after the release.

ihhub:
Hi @oleg-derevenetz , the only worry of mine is that could cause delays of event processing on low end devices when software cursor mode is active. I will try to do some tests on my machine to see how it works.

denierlexiese:
@oleg-derevenetz ,  quick note, when I hook my surface duo up to a 4k monitor the fps drops to 30.  The mouse and keyboard are really slow, like lags behind 3 to 4 frames.  Not sure what's the cause.  The surface duo is annoying because it seems like it isn't able to downscale to 1080p only does the resolution of the output monitor.  On my 1080p monitors the game runs fine.   On my 4k, even with a resolution in the game of 800x600 it runs really slow on a 4k monitor.    Could be a surface duo thing. Just noting odd behavior.

oleg-derevenetz:
Hi @denierlexiese please note that the build from this PR is a debug build with all compiler optimizations turned off. It is not indicative in terms of the game performance, including the performance of the rendering, which is done in software. Real game performance needs to be evaluated on release builds.

oleg-derevenetz:
@denierlexiese

> when I hook my surface duo up to a 4k monitor the fps drops to 30. The mouse and keyboard are really slow, like lags behind 3 to 4 frames.

BTW, is this observed on the Google Play build as well (except the keyboard of course)?

denierlexiese:
Yes, I believe you are correct, that on the Google play build it will get slow on my surface duo as well at large resolutions for mouse lag. On my galaxy phone it doesn't, but I believe my galaxy phone forces 1080p no matter what TV it's running on.

ihhub (approved):

ihhub:

PR #8913, opened 2024-07-03
Fix the unused variable warning in release builds

oleg-derevenetz:
See https://github.com/ihhub/fheroes2/actions/runs/9771857348/job/26975277607#step:8:157 for details.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -1901,8 +1901,7 @@ namespace Interface
             return;
         }
 
-#ifndef NDEBUG
-        size_t objectsReplaced = 0;
+        [[maybe_unused]] size_t objectsReplaced = 0;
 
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz , the proposed solution would lead to invalid logic since replaceKey method is required to be used for both release and debug builds. The compiler is unhappy because we only set the variable value and increment it but never read. I believe the easiest solution would be to use (void)variableName;

oleg-derevenetz:
> the proposed solution would lead to invalid logic since replaceKey() function is required to be used for both release and debug builds

Eh, yes, for some reason I thought that `objectsReplaced` should "normally" always be zero (so no replacements should happen), but it can be 1 as well.

> I believe the easiest solution would be to use (void)variableName;

Let's try the `[[maybe_unused]]` attribute from C++17 then.

ihhub (approved):

ihhub:

PR #8929, opened 2024-07-07
Fix UIDs restoration after Redo operation in Editor

Districh-ru:
The value of `_latestObjectUIDAfter` was never set previously. Now it is set in `prepare()` method that is called from `ActionCreator::commit()` method.


ihhub (approved):

ihhub:

Issue #8505, opened 2024-03-20
Adventure map, faulty artifact being picked up causes forest to disappear temporary


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

The artifact my hero is picking does not correspond the sprite we watch on the screen. Picking it up also causes forest to disappear during fading animation.

https://github.com/ihhub/fheroes2/assets/55348946/8a5f9614-456b-4c74-a64f-3cee61d2bc3f

I've checked on two separate games - there's always a magic ring. But this could be a coincidence though.

### Save file

[Lost_Relic_artifact issue.zip](https://github.com/ihhub/fheroes2/files/14672894/Lost_Relic_artifact.issue.zip)

ihhub:
This is the problem of object removal logic.

Branikolog:
Actually, this map (Lost Relic) has multiple places, where artifacts were placed on improper places. If I'm not mistaken we were already fixed some issue on this map: an artifact was placed on an improper cell somewhere in swamps.

ihhub:
As expected this is due to object positioning on this tile:
```
******* Tile info *******
Tile index      : 1502, point: (62, 20)
MP2 object type : 169 (Artifact)
UID             : 1181
ICN object type : 43 (TREFIR.ICN)
image index     : 22
layer type      : 0 - Object layer
is shadow       : no
region          : 4
ground          : Grass (isRoad: 0)
ground img index: 72, image flags: 0
passable from   : center, top, top-right, right, bottom-right, bottom, bottom-left, left, top-left
metadata value 1: 11
metadata value 2: 0
metadata value 3: 0
--------- Level 1 --------
UID             : 2071
ICN object type : 11 (OBJNARTI.ICN)
image index     : 105
layer type      : 0 - Object layer
is shadow       : no
--- Extra information ---
*************************
```

PR #8936, opened 2024-07-07
Fix swap of the main tile addon if a top object already exists in that tile

Districh-ru:
The objects order on the same tile is determined by the current main addon object type (checked by `getObjectTypeByIcn()` function) and the layer type of the new tile being placed.

This PR fixes the `getObjectTypeByIcn()` function that was giving object type only for the main tile and giving `OBJ_NONE` for all other tiles. Now the function checks all object tiles.

It also forbids main addon swap if a new tile is the background layer one and the current tile already has an object with ID.

Master build:

https://github.com/ihhub/fheroes2/assets/113276641/140d8fac-6d5b-491a-af13-e3186577a294

This PR:

https://github.com/ihhub/fheroes2/assets/113276641/d3e15a10-bc20-41ca-b2ec-d9c4db49130f

### DIFF
--- a/src/fheroes2/maps/maps_tiles_helper.cpp
+++ b/src/fheroes2/maps/maps_tiles_helper.cpp
@@ -1196,17 +1196,21 @@ namespace
 
     bool placeObjectOnTile( const Maps::Tiles & tile, const Maps::ObjectInfo & info )
     {
+        // If this assertion blows up then what kind of object you are trying to place if it's empty?
         assert( !info.empty() );
 
         // Verify that the object is allowed to be placed.
         const fheroes2::Point mainTilePos = tile.GetCenter();
 
         for ( const auto & partInfo : info.groundLevelParts ) {
-            const fheroes2::Point pos = mainTilePos + partInfo.tileOffset;
-            const bool isMainObject = ( partInfo.layerType != Maps::SHADOW_LAYER && partInfo.layerType != Maps::TERRAIN_LAYER );
+            if ( partInfo.layerType == Maps::SHADOW_LAYER || partInfo.layerType == Maps::TERRAIN_LAYER ) {
+                // Shadows and terrain objects do not affect on passability so it is fine to ignore them not being rendered.
+                continue;
+            }
 
-            if ( !Maps::isValidAbsPoint( pos.x, pos.y ) && isMainObject ) {
-                // Only shadow and terrain layer object parts are allowed not to be placed.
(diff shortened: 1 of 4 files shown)


src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -1234,7 +1234,8 @@ namespace
 
             Maps::Tiles & currentTile = world.GetTiles( pos.x, pos.y );
 
-            if ( !isMainObject ) {
+            // Also do not move the main addon if a background layer object is placing over an object with ID.
+            if ( !isMainObject || ( partInfo.layerType == Maps::BACKGROUND_LAYER && currentTile.GetObjectUID() != 0 ) ) {
```
ihhub, line 1238: This is not a valid condition. We need to check whether the object have ground parts below. If it does then this object stands in front of the object we are adding.

ihhub (changes requested):
Hi @Districh-ru , I left one comment here. Could you please take a look when you have time?

src/fheroes2/maps/maps_tiles_helper.cpp
ihhub: How about this:
- check the type of the main tile
- check if any of top layer addons have the same type
- if they do then don't set the new object type since top layer objects are always drawn on top
- an exception applies for true action objects while heroes, castle entrances and etc since we need them to setup properties

What do you think? Will it fix the original problem?

ihhub: On top of this we need to check the order of layers on the tile. If a tile has an object of BACKGROUND_LAYER and we are setting an object of OBJECT_LAYER then we have to change the tile object type. While in the opposite case we shouldn't.

Districh-ru: Hi, @ihhub, thanks for the explained suggestion!
I've done the changes and they work good on a test map.

ihhub (approved):

Branikolog:
Hi!
If I place an artifact on a tile behind the castle - it still shown as artifact on right click on it:
image
After I pick it up the tile is shown as grass:
image
When the same cell without pre placed artifact belongs to castle area:
image

ihhub:
Hi @Branikolog , this is the same issue as #8505 as it is not related to this PR.

Branikolog (approved):
Okay. No more issues found then. :)

ihhub:

PR #8955, opened 2024-07-13
Use showStandardTextMessage() instead of showMessage() where possible and reduce string copying by using std::move()

Districh-ru:
This PR replaces the usage of `showMessage()` by `showStandardTextMessage()` where `fheroes2::Text` is used.
Also in such updated places `std::move()` is now mostly used to prevent sting copying.

This PR affects most pop-up messages. The most changes are done to the hero's actions dialogs.


ihhub (approved):

ihhub:
@Districh-ru , many thanks for these optimizations!

PR #8958, opened 2024-07-14
Fix several SonarQube code smells

ihhub:
To make the code cleaner.

### DIFF
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -1065,7 +1065,7 @@ namespace fheroes2
             }
             break;
         case fheroes2::Key::KEY_DELETE:
-            if ( !res.empty() && pos < res.size() ) {
+            if ( pos < res.size() ) {
                 res.erase( pos, 1 );
             }
             break;
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -1065,9 +1065,8 @@ namespace fheroes2
             }
             break;
         case fheroes2::Key::KEY_DELETE:
-            if ( !res.empty() ) {
-                if ( pos < res.size() )
-                    res.erase( pos, 1 );
+            if ( !res.empty() && pos < res.size() ) {
+                res.erase( pos, 1 );
             }
             break;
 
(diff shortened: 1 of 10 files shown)


src/fheroes2/castle/buildinginfo.cpp
```diff
@@ -94,14 +94,14 @@ namespace
     }
 }
 
-struct buildstats_t
+struct buildingStats
```
oleg-derevenetz, line 97: I always thought that the type names in this project should start with a capital letter, so this struct should be named something like `BuildingStats`.

src/fheroes2/maps/mp2.h
```diff
@@ -49,7 +49,7 @@ namespace MP2
     };
 
     // Tile structure from the original map format.
-    struct mp2tile_t
+    struct mp2TileInfo
```
oleg-derevenetz, line 52: The same note as above - probably `MP2TileInfo`.

src/fheroes2/maps/mp2.h
```diff
@@ -101,7 +101,7 @@ namespace MP2
     };
 
     // Addon structure from the original map format.
-    struct mp2addon_t
+    struct mp2AddonInfo
```
oleg-derevenetz, line 104: The same note as above - probably `MP2AddonInfo`.

src/engine/localevent.cpp
```diff
@@ -1065,9 +1065,8 @@ namespace fheroes2
             }
             break;
         case fheroes2::Key::KEY_DELETE:
-            if ( !res.empty() ) {
-                if ( pos < res.size() )
-                    res.erase( pos, 1 );
+            if ( !res.empty() && pos < res.size() ) {
```
oleg-derevenetz, line 1068: If `res` is empty here, then `pos` can never be less than `res.size()`, because size would be zero, and `pos` is of unsigned `size_t` type, so it can never be less than zero. The check for `!res.empty()` seems redundant.

oleg-derevenetz (changes requested):
Hi @ihhub I just left a few comments, could you please take a look when you have time?

src/fheroes2/castle/buildinginfo.cpp
ihhub: You're right! I completely forgot.

Issue #8780, opened 2024-05-31
Show creature info for the editor dialog


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

![image](https://github.com/ihhub/fheroes2/assets/55348946/7a441b35-64a1-4b4d-a668-c961f06be79b)

Some players could forget the precise creature stats when creating a map.

### Describe the possible solution

It would be nice to show the selected creature properties on right click for this dialog.

ihhub:
This can be done by introducing a new UI element and using the structure based on `DialogElement` class.

PR #8960, opened 2024-07-14
Add monster stats inside editor count dialog

orzm:
This resolves issue https://github.com/ihhub/fheroes2/issues/8780. It is done by changing SelectCount function to accept pointer to DialogElement object instead of a static background image. This allows to insert an interactive element into the dialog, such as the one that allows to view monster info when portrait is clicked.

New dialog window:
SelectCount_new

On left mouse button click on creature portrait:
SelectCount_new_leftClick

On right mouse button hold on creature portrait:
SelectCount_new_rightClick


ihhub:
Hi @orzm , please address IWYU complaints and also include a screenshot of the new dialog in the description of this pull request.

orzm:
> Hi @orzm , please address IWYU complaints and also include a screenshot of the new dialog in the description of this pull request.

Done

ihhub:
Hi @orzm , while the solution solve the problem my main concern is increase of code duplication. The new function is 90% identical than the old one. I suggest to do this way:
1. **Dialog::SelectCount()** function should accept `const DialogElement * uiElement = nullptr` instead of the current `const fheroes2::Image & backgroundImage`.
2. **uiElement** has needed functionality: **draw()** and **processEvents()** method.
3. Derive a new class from `CustomImageDialogElement` class and override the above methods. Let's assume we call it `MonsterDialogElement`.
4. Pass an instance of the newly created class in 2 places in the code where you call `Dialog::SelectMonsterCount()` at the moment.

Therefore, the updated code will support not only monsters but any other UI elements.

What do you think?

orzm:
> Hi @orzm , while the solution solve the problem my main concern is increase of code duplication. The new function is 90% identical than the old one. I suggest to do this way:
> 
> 1. **Dialog::SelectCount()** function should accept `const DialogElement * uiElement = nullptr` instead of the current `const fheroes2::Image & backgroundImage`.
> 2. **uiElement** has needed functionality: **draw()** and **processEvents()** method.
> 3. Derive a new class from `CustomImageDialogElement` class and override the above methods. Let's assume we call it `MonsterDialogElement`.
> 4. Pass an instance of the newly created class in 2 places in the code where you call `Dialog::SelectMonsterCount()` at the moment.
> 
> Therefore, the updated code will support not only monsters but any other UI elements.
> 
> What do you think?

This seems reasonable. I'll work on a solution based on Your suggestions.

Thanks.

ihhub:
Hi @orzm , I've converted this to draft for now. Please mark this pull request as ready to review once it's completed.

ihhub:
Hi @orzm , I rebased your branch from master branch as we updated dependencies and GitHub Actions were failing for your pull request.

src/fheroes2/army/army_bar.cpp
```diff
@@ -474,16 +475,10 @@ bool ArmyBar::ActionBarLeftMouseSingleClick( ArmyTroop & troop )
...
 
-                fheroes2::Sprite surface;
-
-                if ( mons.isValid() ) {
-                    surface = fheroes2::AGG::GetICN( ICN::STRIP, 12 );
-                    fheroes2::renderMonsterFrame( mons, surface, { 6, 6 } );
-                }
-
                 int32_t count = 1;
+                const fheroes2::MonsterDialogElement monsUi{ mons };
```
ihhub, line 479: This code will caused assertion. Try to set a random monster in the Editor. I think we can use `std::unique_ptr` to handle both cases when monster exists and it doesn't.

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -830,6 +832,38 @@ namespace fheroes2
         return false;
     }
 
+    MonsterDialogElement::MonsterDialogElement( const Monster & monster )
+        : _monster( monster )
+    {
+        assert( _monster.isValid() );
+        const fheroes2::Sprite sprite = fheroes2::AGG::GetICN( ICN::STRIP, 12 );
```
ihhub, line 839: This variable must be const reference to avoid memory copy.

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -830,6 +832,38 @@ namespace fheroes2
...
+    }
+
+    void MonsterDialogElement::processEvents( const Point & offset ) const
+    {
+        LocalEvent & le = LocalEvent::Get();
+        const Rect rect{ offset.x, offset.y, _area.width, _area.height };
+        if ( le.isMouseRightButtonPressedInArea( rect ) ) {
+            showPopup( defaultElementPopupButtons );
+        }
+        if ( le.MouseClickLeft( rect ) ) {
```
ihhub, line 857: Here we should use `else if` since right and left mouse clicking are mutually exclusive. This will save unnecessary checks in the code.

src/fheroes2/gui/ui_dialog.cpp
ihhub, line 839: Please remove `fheroes2::` from all places in this file since the code is being executed within the same namespace.

src/fheroes2/gui/ui_dialog.h
```diff
@@ -342,6 +343,23 @@ namespace fheroes2
...
+        ~MonsterDialogElement() override = default;
+
+        void draw( Image & output, const Point & offset ) const override;
+
+        void processEvents( const Point & offset ) const override;
+
+        void showPopup( const int buttons ) const override;
+
+    private:
+        Monster _monster;
```
ihhub, line 360: This member can be `const` as we don't modify it.

ihhub (changes requested):
Hi @orzm , I left multiple comments here. Would you mind to take a look at them when you have time?

src/fheroes2/army/army_bar.cpp
orzm: I get Your point. However:
1) I think You're pointing to the wrong place in code. I guess situation You've mentioned is handled [here](https://github.com/ihhub/fheroes2/pull/8960/files#diff-33704c51a8784d78313c0f993e7ae4a166b206daa982c990f001b893cac35695R1201-R1206).
2) In code pointed above assertion won't occur due to double-checking of isValid() value right before creation of MonsterDialogElement. That behaviour is wrong as well though, because the dialog a random monster won't show at all. I'll fix that.

Nevertheless, `std::unique_ptr` is a good idea and I'll apply that both in army_bar.cpp and editor_interface.cpp for sake of making the code as error-proof as possible.

orzm: Added `std::unique_ptr` instances in mentioned files. Also modified logic, so that count dialog for a random monster is again available. I had to modify `ModifyDialogElement` constructor to throw exception instead of assertion to make current logic work. Don't know if that'll be an acceptable solution.

src/fheroes2/gui/ui_dialog.cpp
orzm: Done

src/fheroes2/gui/ui_dialog.cpp
orzm: Done

src/fheroes2/gui/ui_dialog.cpp
orzm: Done

src/fheroes2/gui/ui_dialog.h
orzm: Done

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -806,6 +809,40 @@ namespace fheroes2
...
+        }
+        const Sprite & sprite = AGG::GetICN( ICN::STRIP, 12 );
+        _area = { sprite.width(), sprite.height() };
+    }
+
+    void MonsterDialogElement::draw( Image & output, const Point & offset ) const
+    {
+        Sprite sprite = AGG::GetICN( ICN::STRIP, 12 );
+        renderMonsterFrame( _monster, sprite, { 6, 6 } );
+        Blit( sprite, 0, 0, output, offset.x, offset.y, sprite.width(), sprite.height() );
```
Districh-ru, line 826: An optional suggestion.
We can use `Copy()` here because all monster images don't have any transparent pixels. We can even make the `sprite` to be single-layer:
```suggestion
        Sprite sprite = AGG::GetICN( ICN::STRIP, 12 );
        sprite._disableTransformLayer();
        renderMonsterFrame( _monster, sprite, { 6, 6 } );
        Copy( sprite, 0, 0, output, offset.x, offset.y, sprite.width(), sprite.height() );
```

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -135,6 +136,11 @@ bool Dialog::SelectCount( std::string header, const int32_t min, const int32_t m
             redraw_count = true;
         }
 
+        if ( uiElement ) {
+            uiElement->processEvents( uiOffset );
+            display.render();
+        }
```
Districh-ru, line 142: When `uiElement` is not `nullptr` the `display.render()` method is called on every loop. It results in very high render calls:
изображение
We can make `processEvents()` to return `true` if `render()` is needed but to do this we will need to update `DialogElement` class and all its inheritances.

So to solve only this issue we can check for mouse action in the UI element area and only then process the events and do `render()`.
Like this (a code without any optimizations):
```suggestion
        if ( uiElement
             && ( le.isMouseLeftButtonReleasedInArea( { uiOffset, uiElement->area() } ) || le.isMouseRightButtonPressedInArea( { uiOffset, uiElement->area() } ) ) ) {
            uiElement->processEvents( uiOffset );
            display.render();
        }
```

Districh-ru (changes requested):
Hi, @orzm, I left two comments. Could you please check them when you have time?

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -806,6 +809,40 @@ namespace fheroes2
         return false;
     }
 
+    MonsterDialogElement::MonsterDialogElement( const Monster & monster )
+        : _monster( monster )
+    {
+        if ( !_monster.isValid() ) {
+            throw std::logic_error( "Invalid monster provided" );
```
ihhub, line 816: We do not use exceptions in the code unless it is absolutely necessary like missing resources. Please use an assertion for this matter.

src/fheroes2/gui/ui_dialog.cpp
orzm: Done

src/fheroes2/dialog/dialog_selectcount.cpp
orzm: Done

src/fheroes2/gui/ui_dialog.cpp
orzm: Restored assertion and adjusted code in editor_interface.cpp, so that it won't cause failure for random monster

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -137,8 +137,11 @@ bool Dialog::SelectCount( std::string header, const int32_t min, const int32_t m
         }
 
         if ( uiElement ) {
-            uiElement->processEvents( uiOffset );
-            display.render();
+            const fheroes2::Rect uiRect{ uiOffset, uiElement->area() };
```
Districh-ru, line 140: We can cache `uiRect` outside of the `while` loop. In example, in line 99:
```cpp
const fheroes2::Rect uiRect = uiElement ? fheroes2::Rect( uiOffset, uiElement->area() ) : fheroes2::Rect();
```

Districh-ru (changes requested):
Well done, @orzm.
I left just one optimization suggestion.

src/fheroes2/dialog/dialog_selectcount.cpp
orzm: Done

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -135,6 +138,13 @@ bool Dialog::SelectCount( std::string header, const int32_t min, const int32_t m
             redraw_count = true;
         }
 
+        if ( uiElement ) {
+            if ( le.isMouseLeftButtonReleasedInArea( uiRect ) || le.isMouseRightButtonPressedInArea( uiRect ) ) {
```
Districh-ru, line 142: To make SonarQube happy it is better to fix the maintainability issue and merge these two `if` conditions.
>rule S1066: Nested code - blocks of code inside blocks of code - is eventually necessary, but increases complexity. This is why keeping the code as flat as possible, by avoiding unnecessary nesting, is considered a good practice.
Merging if statements when possible will decrease the nesting of the code and improve its readability.

Districh-ru (approved):
👍 
@orzm, I left just one more minor code improvement comment.

src/fheroes2/dialog/dialog_selectcount.cpp
orzm: Done

ihhub (approved):

PR #8964, opened 2024-07-14
Allow to type "/", "*", "-", "+" and "." from the keypad

Districh-ru:
This PR allows to type "/", "*", "-", "+" and "." symbols from the keypad then Num Lock is on.


ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for this improvement!

Issue #8826, opened 2024-06-12
Expand Virtual Keyboard to support more languages


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

At the moment the Virtual Keyboard supports only English, Polish and Russian languages. The map makers will need more languages to be supported for map creations.

### Describe the possible solution

Add more languages to be supported. The start point is **ui_keyboard.cpp** file.

ihhub:
I am closing this issue. If we need to add more languages we can create separate issues specific to these languages.

PR #8983, opened 2024-07-22
Add Ukrainian language virtual keyboard

ihhub:
relates #8826

image


Districh-ru:
Hmm, it looks that we miss these two letters for buttons font: 
изображение

Districh-ru:
Now we have them on the keyboard:
изображение

Districh-ru (approved):

ZaHaDumGoth:
Yes, there were 2 letters missing. Now is correct. Thank you.

PR #8986, opened 2024-07-23
AI code refactoring

oleg-derevenetz:
This PR contains no functional changes, just AI code simplification and refactoring.

In the `master` branch, for historical reasons, the AI code is organized in such a way that there is a rudimentary support for different AI implementations. I believe it looks redundant at the moment. All we have for now is a single AI implementation called "normal" and it does absolutely everything, therefore, in my opinion, there is no need for an extra layer, which just contains a bunch of methods that do literally nothing.

With this PR:

* There is no longer a bunch of code that was made "for the future", but now just doesn't do anything useful;
* There are no more virtual methods (`AI::Base` -> `AI::Normal`) which do nothing useful but prevent the compiler from performing optimizations (e.g. code inlining during LTO);
* The local code has been moved to an anonymous namespace as much as possible;
* The AI battle planner code is now independent from the general AI code and vice versa, so they can be used (and are used) independently of each other.
* Code structure is improved in general. The AI code is more clearly divided into separate independent modules, no more (well, almost) "god headers", which include a lot of unrelated things that are then defined in different CPP files.

If in the future we suddenly still need to support various AI engines, it will be possible to make corresponding changes. But for now, the current AI code architecture looks like a bit of overengineering, hence this PR.


idshibanov (approved):
I have no objections about this. 4 years ago I hoped for a competing AI to pop up but looks like it's never going to happen.

ihhub (approved):

ihhub:

PR #8991, opened 2024-07-26
Limit the use of the Summon Boat by AI

oleg-derevenetz:
In the original game, AI could not use the Summon Boat spell at all, and many of the original maps (including the maps of the original campaign) were created with this in mind. For example, on the famous "Broken Alliance" map, the powerful AI-only Purple hero sits on a separate island with a fully built castle and Titans, and, if this hero has the Summon Boat spell (which happens quite often), he may just summon a neutral boat at the beginning and kill everyone on the continent. Well, he can also get the Dimension Door spell and do the same, but the probability of this is still much lower (at least such a scenario was possible in the original game as well).

In fheroes2, however, the AI is able to use this spell. To mitigate the impact of this on the gameplay of the original maps, this PR at least prohibits the AI from summoning "neutral" boats (i.e. boats placed on the map by the map creator and not yet used by anyone).


oleg-derevenetz:
> However shouldn't the human player also be limited and not be able to summon neutral boats?

In the OG, the human player (at least the first "in line") is not limited in this regard:

https://github.com/user-attachments/assets/703e5a9e-ebb7-4363-98d9-ff617d04f63d

and this is what the developers of a number of maps were probably counting on.

> That approach is more sensible in my mind having the AI and human as equal options as possible.

Providing AI and humans with as much equal opportunities as possible is not the end in itself IMHO. The real goal is a hassle-free gameplay, including the original maps developed at the time for the original game.

oleg-derevenetz:
> To no doubt to make some OG maps unplayable but the gain is an better and simpler game overall. Anyhow it would be wise to have a set of replacement maps for the broken maps if going that route.

Many people play fheroes2 for nostalgic reasons. That's what they write in the Google Play reviews, something like "I was happy to play the maps and campaigns I've known since childhood". If we redo it too much, then it won't be HoMM2 anymore, but a completely different game, unfamiliar to these people. This, of course, does not mean that reasonable improvements are not needed, but IMHO OG maps and campaigns should still remain playable. At that time, the map creators widely practiced the approach of "locking" the "pre-pumped" AI opponents on the islands so that the human players would come to them when they were ready, but AI opponents could not come to human players too early (Broken Alliance, mission 8 of PoL campaign, etc). Allowing the AI players to summon neutral boats (often placed by the map creator for use by a human player) a kind of ruins that.

ihhub:
Hi @oleg-derevenetz , would it be better not limit such behavior only for the original maps? We can distinguish map types during gameplay.

oleg-derevenetz:
Hi @ihhub

> would it be better not limit such behavior only for the original maps? We can distinguish map types during gameplay.

Do you mean to ban summoning only on the original HoMM2 maps, but enable it on Resurrection maps? Okay, let's do it this way.

ihhub (approved):

ihhub:

PR #8994, opened 2024-07-27
Update Russian translation

Districh-ru:
Translate recently added/changed phrases.


files/lang/ru.po
```diff
@@ -4432,33 +4431,28 @@ msgid "None."
 msgstr "Нет."
 
 msgid "One side defeats another."
-msgstr "Одна сторона побеждает другую."
+msgstr "Один союз побеждает других."
```
oleg-derevenetz, line 4434: As far as I understand, this change was made due to the fact that there could only be 2 alliances in the original maps, but there may be multiple in the `fh2m` maps? Maybe then it is worth changing the corresponding English strings too? As well as constants like `WINS_SIDE` or `VICTORY_DEFEAT_OTHER_SIDE`, since now there may be multiple "other sides"?

Districh-ru: Hi, @oleg-derevenetz, our editor currently supports only 2 alliances. But the editor code was made to support more alliances in the future.
When we will add such ability we will update this phrase and the other alliances-related phrases.

oleg-derevenetz: Hi @Districh-ru Then perhaps it is not worth updating the translation to a phrase implying that there may be more than two alliances? At least until it is fully implemented everywhere.

Districh-ru: Sure, we will update the translation after the original phrases are changed.

oleg-derevenetz: Ah, that's the editor-specific msgid. I thought that it's common msgid from `game_over.cpp`, but that msgid is formulated in a different way ("Your side defeats the opposing side."). Then everything is OK.

ihhub: Hi @Districh-ru , let's revert this change for now to proceed with the rest of review.

oleg-derevenetz: Hi @ihhub this change is OK, it's editor-only.

ihhub (approved):

ihhub:
@Districh-ru , thank you very much for the update!

@oleg-derevenetz , if something is not correct we can still change it later. I wanted to unblock this pull request to avoid having it in a limbo.

oleg-derevenetz:
Hi @ihhub

> if something is not correct we can still change it later. I wanted to unblock this pull request to avoid having it in a limbo.

Everything is OK. At first I thought that the translated string relates to the game itself, but the I realized that it's the editor-only string.

PR #8996, opened 2024-07-27
Fix Cliff Nest glowing pixel

### LINKED ISSUE #7666: Castle construction screen, cliff nest has glowing pixel
### Preliminary checks

### Platform

Windows

### Describe the bug

nest

A small dot is shining in the Cliff Nest icon.

### Save file

None.

### Additional info

_No response_

Districh-ru:
Fix #7666 

Master build:
before

This PR:
after

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -578,6 +578,35 @@ namespace
         fheroes2::ReplaceTransformIdByColorId( newImage, 6U, 10U );
         fheroes2::Blit( newImage, releasedSprite, offsetX + 3, offsetY );
     }
+
+    void removeColorCycling( fheroes2::Image & image )
+    {
+        if ( image.empty() ) {
+            return;
+        }
+
+        uint8_t * data = image.image();
+        const uint8_t * imageEnd = data + static_cast<size_t>( image.width() ) * image.height();
+
+        for ( ; data < imageEnd; ++data ) {
+            if ( *data < 214 ) {
+                continue;
+            }
+
+            if ( *data < 218 ) {
+                *data = 188;
+            }
+            else if ( *data < 222 ) {
+                *data = 118;
+            }
+            else if ( *data < 231 ) {
+                continue;
+            }
+            else if ( *data < 241 ) {
+                *data = 69;
+            }
+        }
+    }
 }
 
 namespace fheroes2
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @Districh-ru , while this solution is absolutely correct I think we should avoid individual image fixes as it bloats our code base. My idea is to have a list of images and automatically apply non-cycling palette on them if they are requested for rendering. This approach is much more flexible and easier to maintain. What do you think?

Districh-ru:
> Hi @Districh-ru , while this solution is absolutely correct I think we should avoid individual image fixes as it bloats our code base. My idea is to have a list of images and automatically apply non-cycling palette on them if they are requested for rendering. This approach is much more flexible and easier to maintain. What do you think?

Hi, @ihhub, initially I was also thinking to apply the non-cycling palette. But here we need to update the only pixel while the `ApplyPalette()` will update every pixel in the image.
Also we are fixing the image from the original assets that have not changed for about 20 years and with high probability will not change in future. So there will be no case that this glowing pixel will be fixed or moved to the other pixel in the assets.

Or should we write a new function that scans the image and replaces only the "cycling animation" pixels (color ID 214-221 and 231-235 and 238-241) ?
A little later I can implement such function and use it for every such image.
And it is better to fix images on load and not on every render, or am I wrong... :)

Districh-ru:
@ihhub, I've added a function to quickly update images with glowing pixels. Is it OK or should I revert the changes?

ihhub:
Hi @Districh-ru , you are correct about the images not being changed in the past and they aren't going to be changed in the future. My idea is to balance the complexity of the code and the outcome as we already have a huge code base to maintain regarding corrections of images. Therefore, I proposed to use `fheroes2::ApplyPalette()` function to avoid rewriting a heavy code just for few pixels in each image.

Your newly proposed function does exactly the same as `ApplyPalette( image, PAL::GetPalette( PAL::PaletteType::NO_CYCLE ) );`. While the performance benefit is not obvious (accessing a table is O(1) while code branching is >O(1) and impacts CPU prediction capabilities), having an extra function is not necessary.

Districh-ru:
> Your newly proposed function does exactly the same as `ApplyPalette( image, PAL::GetPalette( PAL::PaletteType::NO_CYCLE ) );`. While the performance benefit is not obvious (accessing a table is O(1) while code branching is >O(1) and impacts CPU prediction capabilities), having an extra function is not necessary.

Hi, @ihhub.
I removed the new function and used `ApplyPalette()` to fix the glowing pixels.

ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for this fix!

PR #9000, opened 2024-07-27
Update hero spell points indicator after the battle and before performing defeated opponent fade animation

### LINKED ISSUE #7518: Adventure map, the hero spell points indicator bar should be updated before returning to the adventure map after a combat
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the hero spell points indicator bar should be updated before returning to the adventure map after a combat.
This is the case in the original game.

In original game:

https://github.com/ihhub/fheroes2/assets/43583994/2e32935b-c560-4256-ab58-086a3a3fe7ec

In fheroes2:

https://github.com/ihhub/fheroes2/assets/43583994/3cdae4eb-2f7d-42be-a6c0-35ab0b0b8cee

### Save file

Hera are the save files:
[Spell points indicator update 01.zip](https://github.com/ihhub/fheroes2/files/12269959/Spell.points.indicator.update.01.zip)

### Additional info

_No response_

Districh-ru:
Fix #7518

This PR:
- forces the heroes icons update after any battle during player actions;
- forces heroes and castles update during AI actions after the battle with the human player;
- also makes some minor code improvements to heroes action code including AI actions.

Master build (AI actions):

https://github.com/user-attachments/assets/e20f597f-f0d1-4128-bd04-a972e5c3be6d

This PR (AI actions): 

https://github.com/user-attachments/assets/74d0a483-6f66-4e4f-a903-90fa54d75d62

This PR (player actions):

https://github.com/user-attachments/assets/76df81b6-ab0c-4d71-ab55-42d67d1c4308

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -382,10 +382,10 @@ namespace
                 assert( 0 );
 
                 if ( isDefenderHuman ) {
-                    DEBUG_LOG( DBG_AI, DBG_INFO, defender->GetName() << " the AI hero is protecting the human castle " << castle->GetName() )
+                    DEBUG_LOG( DBG_AI, DBG_WARN, defender->GetName() << " the AI hero is protecting the human castle " << castle->GetName() )
                 }
                 else {
-                    DEBUG_LOG( DBG_AI, DBG_INFO, defender->GetName() << " the human hero is protecting the AI castle " << castle->GetName() )
+                    DEBUG_LOG( DBG_AI, DBG_WARN, defender->GetName() << " the human hero is protecting the AI castle " << castle->GetName() )
                 }
             }
 
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -328,6 +328,20 @@ namespace
         taker.GetBagArtifacts().exchangeArtifacts( giver.GetBagArtifacts(), taker, giver );
     }
 
+    // Call this function after the battle with human controlled hero or castle to update icons on the interface.
+    // This function also runs fade-in after the battle if it is planned.
+    void redrawPlayerIcons()
+    {
+        if ( Game::validateDisplayFadeIn() ) {
+            Interface::AdventureMap::Get().redraw( Interface::REDRAW_ICONS );
+
+            fheroes2::fadeInDisplay();
+        }
+        else {
+            Interface::AdventureMap::Get().setRedraw( Interface::REDRAW_ICONS );
+        }
+    }
+
     void AIToCastle( Heroes & hero, const int32_t dstIndex )
     {
         Castle * castle = world.getCastleEntrance( Maps::GetPoint( dstIndex ) );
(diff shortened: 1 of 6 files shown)


src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -376,22 +376,49 @@ namespace
...
 
             Heroes * defender = castle->GetHero();
-            const bool playVanishingHeroSound = defender != nullptr && defender->isControlHuman();
+            const bool isDefenderHuman = castle->isControlHuman();
+
+            if ( defender && ( defender->isControlHuman() != isDefenderHuman ) ) {
+                assert( 0 );
+
+                if ( isDefenderHuman ) {
+                    DEBUG_LOG( DBG_AI, DBG_INFO, defender->GetName() << " the AI hero is protecting the human castle " << castle->GetName() )
```
oleg-derevenetz, line 385: If something happens that shouldn't normally happen, it's better to use `DBG_WARN`:

```suggestion
                    DEBUG_LOG( DBG_AI, DBG_WARN, defender->GetName() << " the AI hero is protecting the human castle " << castle->GetName() )
```

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -376,22 +376,49 @@ namespace
...
+                }
+                else {
+                    DEBUG_LOG( DBG_AI, DBG_INFO, defender->GetName() << " the human hero is protecting the AI castle " << castle->GetName() )
```
oleg-derevenetz, line 388: Same as above.

```suggestion
                    DEBUG_LOG( DBG_AI, DBG_WARN, defender->GetName() << " the human hero is protecting the AI castle " << castle->GetName() )
```

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2336,7 +2364,7 @@ namespace
             return;
         }
 
-        enum class Outcome
+        enum class Outcome : uint8_t
```
oleg-derevenetz, line 2367: Are there any real reasons for this change (and similar changes below)? This `enum` is a local one, it is not stored anywhere to read it later on a different machine (and never will).

If we are talking about pinching pennies, then yes, `uint8_t` is usually smaller than `int`, but this may be compensated by various speed penalties. For instance, IIRC, ARM Cortex CPUs have 32-bit ECC granularity, so they cannot "just update" a single byte in some 32-bit aligned memory location from L1 cache. In this case, the L1 data memory system reads the existing data from the memory to the cache first, "merges" the updated bytes into those from-memory bytes, then calculates ECC from that merged value and only then stores it into memory. A lot of work compared to just storing a 32-bit or 64-bit value at once, right? :) Yes, CPU attempts to apply different tricks trying to "merge" multiple different store operations together, etc. What I'm trying to say is that in this particular case there will be no observable gain from this change, so why do this? :)

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left some suggestions and one question regarding this PR, could you please take a look when you have time?

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: Hi, @oleg-derevenetz, thanks for the clarification!
I did not know such details and just followed the clang-tidy suggestion `enum 'Outcome' uses a larger base type ('int', size: 4 bytes) than necessary for its value set, consider using 'std::uint8_t' (1 byte) as the base type to reduce its size [performance-enum-size]`.

But when talking about size of this one instance and CPU performance I choose the performance.

I've reverted the change.

oleg-derevenetz: > I did not know such details and just followed the clang-tidy suggestion `enum 'Outcome' uses a larger base type ('int', size: 4 bytes) than necessary for its value set, consider using 'std::uint8_t' (1 byte) as the base type to reduce its size [performance-enum-size]`.

Some new suggestion from clang-tidy > 15.0? I'm also waiting for `ubuntu-24.04` image out of beta (and become `ubuntu-latest`) to start using the Clang tools 18.x. Anyway, this may be a good suggestion if we are hitting memory limits somewhere (for example, if we have some hotspot in our code processing an array of structures with enums that needs to be fitted into CPU cache). In this case, there is nothing like that, there is a single variable of a local `enum` type, so all we have there is extra words in the source file and no real gain.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@Districh-ru , many thanks for these updates!

PR #9002, opened 2024-07-27
Minor code clean in battle_interface.cpp/.h

Districh-ru:
Optimized Lightning spell effect calculation, Rainbow image generation (by disabling transform layer), renamed some variables.

Fixed a bug: endless width of units turn order table (width was not reset to 0 at the beginning of calculations).


ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for this code cleanup!

PR #9013, opened 2024-07-29
Mark the appropriate headers with IWYU pragma: associated

oleg-derevenetz:
In fheroes2, a scheme is sometimes practiced (especially in old code) when, for one header file with declarations, there are multiple CPP files with implementations of functions or classes declared in that header file (or even vice versa, when one CPP file contains implementations for declarations from multiple header files). Relatively recent versions of IWYU have the [pragma "associated"](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-associated) for better tracking of such dependencies.

### DIFF
--- a/src/fheroes2/ai/ai_battle_spell.cpp
+++ b/src/fheroes2/ai/ai_battle_spell.cpp
@@ -25,7 +25,7 @@
 #include <ostream>
 #include <vector>
 
-#include "ai_battle.h"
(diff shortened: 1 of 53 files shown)


Districh-ru (approved):

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for making the code better!

PR #9014, opened 2024-07-29
Update of Ukrainian translation

ZaHaDumGoth:
Small fixes of spells name and map editor.


files/lang/uk.po
```diff
@@ -4456,13 +4450,11 @@ msgstr "Дозволити стандартні умови перемоги"
...
 
-#, fuzzy
 msgid "Set alliances:"
-msgstr " альянс"
+msgstr "Встановити альянс:"
 
-#, fuzzy
 msgid "{%number} alliance: "
-msgstr " альянс"
+msgstr "Номер альянсу:"
```
Districh-ru, line 4457: Hi, @ZaHaDumGoth.
The `{%number}` in this text is replaced with `1-й` or `2-й`, or ...
So this string should be translated as "{%number} альянс:" so in the editor it will be, in example, "1-й альянс:".

ZaHaDumGoth: Unfortunately {%number} was grey colored instead of blue as others with replacing function. Thats why I considered it meas "quantity". Ok I will fix it.

oleg-derevenetz (approved):

oleg-derevenetz:
@ZaHaDumGoth Thank you very much!

PR #8358, opened 2024-01-31
Update the resource panel immediately after a purchase.

dood-apo:
Fix #4732

PR #9026, opened 2024-08-04
Rework the castle buildings construction logic

### LINKED ISSUE #9020: Castle Construction screen, quantity of resources is not updated when construction takes place after changing castle with the right left arrows at the bottom of the screen
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, the quantity of resources is not updated when construction takes place after changing castle with the right left arrows at the bottom of the screen.

In fheroes2:

https://github.com/user-attachments/assets/e7312535-b1fa-4bf9-b3b1-26eff648f8fd

### Save file

Here is the save file:
[Before market no update 01.zip](https://github.com/user-attachments/files/16487516/Before.market.no.update.01.zip)

### Additional info

_No response_

oleg-derevenetz:
fix #9020

Currently in the `master` branch, buildings that are "under construction" have not actually been built yet, and the resources for their construction have not yet been written off from the kingdom's balance. #8358 tried to solve this issue "head-on" by pre-calculating the final kingdom balance and drawing it in advance, but this approach did not always work well due to the possibility of further redrawing the "actual" balance on top of the previously drawn "pre-calculated" one in certain cases. I decided not to try to "catch" these redraws because I believe that this approach is too error-prone, so I decided to implement the opposite logic - buy a building first, and only then build it.

With this PR:

https://github.com/user-attachments/assets/005f4d88-0d9d-4fe5-97d5-17ba3a6f49b9

Also renamed the `building_t` to the `BuildingType`, because `_t` suffix is reserved by POSIX for their own identifiers.

Since this PR changes the logic of redrawing the castle buildings, it should be well tested. Maybe it won't go to the next release (depends on the specific date of this release).

### DIFF
--- a/src/fheroes2/ai/ai_common.cpp
+++ b/src/fheroes2/ai/ai_common.cpp
@@ -44,7 +44,7 @@
 #include "resource.h"
 #include "resource_trading.h"
 
-bool AI::BuildIfPossible( Castle & castle, const building_t building )
+bool AI::BuildIfPossible( Castle & castle, const BuildingType building )
 {
     switch ( castle.CheckBuyBuilding( building ) ) {
     case BuildingStatus::LACK_RESOURCES: {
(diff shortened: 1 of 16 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for this important update!

PR #9027, opened 2024-08-04
Rename cost_t to Cost

oleg-derevenetz:
The `_t` suffix is reserved by POSIX for their own identifiers.


ihhub (approved):

ihhub:

Issue #9025, opened 2024-08-04
Adventure Map, Treasure Chest window, sprites for the gold pile and experience points star, and the text below the sprites, are not centered horizontally in relation to the position of the corresponding "YES" "NO" buttons


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the sprites for the gold pile and experience points star, and the text below the sprites, are not centered horizontally in relation to the position of the corresponding "YES" "NO" buttons.

In fheroes2:

![XP Need correction 01](https://github.com/user-attachments/assets/ea49bb7f-3997-4f6f-9f91-a0b966df50f1)

**Edit 2024/08/05.**

If necessary, the text can be arranged on several lines, for example:

"500
(XP Needed: 1000)"

Or:

"500

XP Needed:

1000"

With:

"1000
(Total: 6000)"

Or:

"1000

Total:

6000"

Or even, the size of the window can be enlarged horizontally.

### Save file

Not needed.

PR #9028, opened 2024-08-04
Minor chest dialog changes

### LINKED ISSUE #9024: Game, Information about XP
### Preliminary checks

I suggest replacing the text "Need:" with the text "XP Needed:" in all windows giving information about the experience points needed to advance to the next level.

In the example:

This:

"500 (XP Needed: 1000)"

Instead of this:

"500 (Need: 1000)"

In fheroes2:

XP Need correction 01

Replace the text "Need:" with the text "XP Needed:" in all windows giving information about the experience points needed to advance to the next level.

Example:

This:

"500 (XP Needed: 1000)"

Instead of this:

"500 (Need: 1000)"

### Additional info

_No response_

Cheerfulbull:
Close #9024 by centring the XP icon and text and changing the XP message to "XP Needed". One problem: the XP message, when centred, becomes very cramped (comes very close to the edge). What would be the best way to fix this? 
Thank you.


oleg-derevenetz:
Hi @Cheerfulbull I suggest we limit the scope of this PR to #9025 and create a separate PR for #9024. The idea with the phrase "XP needed:" is questionable in general, because:

* The string is getting longer without explicit necessity;
* It is in English that there is a convenient two-letter abbreviation "XP", meaning "experience". In other languages, there may not be such a convenient short abbreviation, that is, we get a localization issue out of the blue.

Cheerfulbull:
> Hi @Cheerfulbull I suggest we limit the scope of this PR to #9025 and create a separate PR for #9024. The idea with the phrase "XP needed:" is questionable in general, because:
> 
> * The string is getting longer without explicit necessity;
> * It is in English that there is a convenient two-letter abbreviation "XP", meaning "experience". In other languages, there may not be such a convenient short abbreviation, that is, we get a localization issue out of the blue.

Alright, thank you, I updated the code and removed the text change, the updated PR description would then be:
>Close #9025 by centring the XP icon and text.

oleg-derevenetz:
If I understand correctly what the issue is, I suppose @LeHerosInconnu will say they're still not centered. With this PR:

chest

@LeHerosInconnu what do you think?

LeHerosInconnu:
Hello @Cheerfulbull and @oleg-derevenetz,

> If I understand correctly what the issue is, I suppose @LeHerosInconnu will say they're still not centered. With this PR:
> 
> chest
> 
> @LeHerosInconnu what do you think?

Effectively, you can see even with the naked eye that graphics and texts are not centered horizontally in relation to the respective buttons. :)

Centering chest elements 01

ihhub:
Hi @LeHerosInconnu , does this change fixes the original problem?

LeHerosInconnu:
Hello @ihhub,

> Hi @LeHerosInconnu , does this change fixes the original problem?

Okay, I'll test it when the conflicts will be resolved.

ihhub:
@LeHerosInconnu , actually can you please check the latest snapshot? I believe we might have fixed it.

ihhub:
Closing this pull request as the original problem was fixed in https://github.com/ihhub/fheroes2/pull/9103

PR #9029, opened 2024-08-05
Fix multiple file-related bugs

oleg-derevenetz:
`std::filesystem`'s `stem()` doesn't return the empty string if filename begins with dot and has no other dots (`/foo/.bar` -> `.bar`), so I preferred to replicate this behavior in our home-grown support library.

Also this PR fixes the following issues:

* Issues with files whose names consist only of an extension (`.sav`, `.fh2m` and so on) . Currently in the `master` branch, when trying to display a list of save files, among which there is at least one file with a name consisting only of an extension (e.g. `.sav`), we will get the assertion failure inside the `getLineWidth()`. Since such filenames are absolutely legal, I decided to fix this. Savefiles with this PR:

  https://github.com/user-attachments/assets/121a4561-71fa-47ae-b9cd-6ba9e43ca64f

  Please note that users are able to load and overwrite the savefile with the name `.sav` if they choose this file from the list, but if they enter the name `.sav` to the input control, the final filename becomes `.sav.sav` (because of the extension that is automatically added to this name), what I consider acceptable.

  Maps with this PR:

  https://github.com/user-attachments/assets/bc125761-a0b2-48f8-8410-e1b53470b084

  Please note that users are able to to start the map with filename `.fh2m` and load it to the Editor, but are not able to overwrite it, because of the extension that is always added to this name, what I also consider acceptable.

* Currently in the `master` branch, users are not able to load a savefile from the Adventure map GUI (well, almost - see below), if there are no savefiles of the type of game that is currently being played:

  https://github.com/user-attachments/assets/edd760d0-82b6-48b0-96a5-cb609fb2835c

  In other words, user is able to load a campaign savefile using hotkey, but is not able to do it via GUI. With this PR:

  https://github.com/user-attachments/assets/197cc9f1-5c6a-4282-8dcd-4e239fd9c7cf

* Currently in the `master` branch, in the "load game" dialog, when the last savefile has been deleted using the corresponding hotkey, the following happens:

  https://github.com/user-attachments/assets/7cbaed52-de5d-47a4-95f8-408414cb3987

  This is due to a crash that previously occurred when the dialog box was redrawn with an empty file list. I find this behavior problematic because the list of savefiles may be empty here for natural reasons - for example, if all available savefiles of the appropriate type have an outdated version (no longer supported) or corrupted - because the logic of the main menu only checks for files with the appropriate extension, but does not check that they are really suitable for loading. It is better to fix the crash in another way, which I did (at least I hope so). With this PR:

  https://github.com/user-attachments/assets/458e9439-6ca8-48c8-aa67-983be27924bb

### DIFF
--- a/src/engine/system.cpp
+++ b/src/engine/system.cpp
@@ -366,13 +366,13 @@ std::string System::GetBasename( std::string_view path )
     return std::string{ path.substr( pos + 1 ) };
 }
 
-std::string System::truncateFileExtensionAndPath( std::string_view path )
+std::string System::GetStem( std::string_view path )
 {
     std::string res = GetBasename( path );
 
     const size_t pos = res.rfind( '.' );
 
-    if ( pos != std::string::npos ) {
+    if ( pos != 0 && pos != std::string::npos ) {
         res.resize( pos );
     }
 
--- a/src/engine/system.h
+++ b/src/engine/system.h
@@ -48,7 +48,7 @@ namespace System
 
     std::string GetDirname( std::string_view path );
     std::string GetBasename( std::string_view path );
-    std::string truncateFileExtensionAndPath( std::string_view path );
+    std::string GetStem( std::string_view path );
 
     bool IsFile( const std::string & path, bool writable = false );
     bool IsDirectory( const std::string & path, bool writable = false );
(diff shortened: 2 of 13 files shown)


ihhub:
Hi @oleg-derevenetz , I will review your pull requests and others within several days as I still need some time to put my full attention to the project.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these updates!

PR #9032, opened 2024-08-07
Tune the set of C/C++ compiler warnings

oleg-derevenetz:
* `-Winit-self` and `-Wuninitialized` are part of the `-Wall`;
* `-Wextra-semi` is C++ only;
* New warnings: `-Wcast-qual` and `-Wsuggest-override`, which are supported both by GCC and Clang.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for the changes!

PR #9033, opened 2024-08-08
Fix the audio playback and distance calculation in the 3D Audio mode

oleg-derevenetz:
This PR is intended to fix the following issues:

* Currently in the `master` branch, the volume adjustment for sounds (including environmental sounds) is performed as follows: sound playback is started, and, once channel ID is known (it may not be known in advance, because in most cases, playback starts on the "first available" channel, and SDL does all the internal bookkeeping itself), playback is paused, channel parameters (volume and sound effects) are adjusted, and then playback resumes. The problem here is that the sound may have time to play a little before it is paused, which can sometimes lead to sudden volume spikes (which is especially noticeable in the 3D Audio mode, see the removed `TODO` comment). With this PR, if there is a need to adjust any additional playback parameters, then the audio fragment itself is muted before playback begins (yes, there really is such an opportunity, oddly enough, because each audio chunk has its own volume, independent of the channel volume), and its volume is restored only after all channel parameters are settled. There should be no more volume spikes or "jumping" volume in 3D Audio mode. In addition, I have canceled many previous precautions (for example, pre-setting the channel volume to zero in a number of cases).
* Currently in the `master` branch, adjusting the environmental sound volume in the 3D Audio mode is implemented quite strangely. First of all, the channel volume is configured, which is ALREADY depends on the distance to the sound source object (just like in the non-3D mode). Then, additional channel configuration is performed using the `Mix_SetPosition`, where the sound volume is additionally adjusted via the `distance` parameter, which is calculated as follows: `255 * ( volumePercentage - 100 )`. Firstly, this formula looks a bit strange, because its result becomes negative if `volumePercentage` is less than 100, and secondly, the distance to the object is already taken into account in the channel volume. This PR uses a more canonical approach to volume control (as I imagine it): namely, all the mixer channels have the same volume, which is set once and for all at the game startup or through the settings, and `Mix_SetPosition` is always used to adjust the volume depending on the distance to the sound source, both in the 3D and non-3D audio mode.

### DIFF
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -880,9 +880,11 @@ void Mixer::Stop( const int channelId /* = -1 */ )
 {
     const std::scoped_lock<std::recursive_mutex> lock( audioMutex );
 
-    if ( isInitialized ) {
-        Mix_HaltChannel( channelId );
+    if ( !isInitialized ) {
+        return;
     }
+
+    Mix_HaltChannel( channelId );
 }
 
 bool Mixer::isPlaying( const int channelId )
--- a/src/engine/audio.cpp
+++ b/src/engine/audio.cpp
@@ -790,7 +790,7 @@ int Mixer::Play( const uint8_t * ptr, const uint32_t size, const int channelId, 
 
     soundSampleManager.clearFinishedSamples();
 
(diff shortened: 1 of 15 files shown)


ihhub (approved):

ihhub:

PR #9038, opened 2024-08-10
Fix the recently introduced code smell

oleg-derevenetz:
Also added the `google-explicit-constructor` check for Clang-Tidy. Unfortunately, there is no non-vendor-specific check for this.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you for taking care the quality of the code!

Issue #6799, opened 2023-03-09
Combat, sprites of the mirror troop info window animation should be different (brighter sprites)


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

In fheroes2, the sprites of the mirror troop info window animation should be different (brighter sprites).

In fheroes2:

![Mirror troop Mirror Image spell 01](https://user-images.githubusercontent.com/43583994/224078946-0329b047-aa11-4c98-b8e8-58d33a3bb4c4.png)

### Describe the possible solution

In fheroes2, the sprites of the mirror troop info window animation should be different (brighter sprites).

### Additional info

Related issue: https://github.com/ihhub/fheroes2/issues/6796.

PR #9040, opened 2024-08-10
Render mirrored troop in dialog properly

orzm:
Resolves issue https://github.com/ihhub/fheroes2/issues/6799

https://github.com/user-attachments/assets/ad51e59e-6f84-45e0-a0e2-1aa2106cc6c9

### DIFF
--- a/src/fheroes2/dialog/dialog_armyinfo.cpp
+++ b/src/fheroes2/dialog/dialog_armyinfo.cpp
@@ -47,6 +47,7 @@
 #include "monster_anim.h"
 #include "monster_info.h"
 #include "morale.h"
+#include "pal.h"
 #include "resource.h"
 #include "screen.h"
 #include "settings.h"
@@ -444,7 +445,10 @@ namespace
         fheroes2::Display & display = fheroes2::Display::instance();
 
         if ( fheroes2::FitToRoi( monsterSprite, inPos, display, outPos, inSize, roi ) ) {
-            fheroes2::Blit( monsterSprite, inPos, display, outPos, inSize, isReflected );
+            fheroes2::Sprite outMonsterSprite = monsterSprite;
+            if ( troop.IsMirror() )
+                fheroes2::ApplyPalette( outMonsterSprite, PAL::GetPalette( PAL::PaletteType::MIRROR_IMAGE ) );
+            fheroes2::Blit( outMonsterSprite, inPos, display, outPos, inSize, isReflected );
         }
 
         if ( isAnimated )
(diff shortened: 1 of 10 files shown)


oleg-derevenetz:
Hi @orzm instead of passing the mirror flag through the function parameter chains, I suggest to add a virtual method just like `Troop::GetAffectedDuration()` (which is already used for similar purposes).

orzm:
> Hi @orzm instead of passing the mirror flag through the function parameter chains, I suggest to add a virtual method just like `Troop::GetAffectedDuration()` (which is already used for similar purposes). What do you think?

Done. Thanks for Your suggestion.

oleg-derevenetz (approved):
Hi @orzm there is an existing virtual method `Troop::isModes()` which allows you to do exactly what you need in your case. It turns out that there is no need for a special virtual method for that.

src/fheroes2/dialog/dialog_armyinfo.cpp
```diff
@@ -444,7 +445,11 @@ namespace
         fheroes2::Display & display = fheroes2::Display::instance();
 
         if ( fheroes2::FitToRoi( monsterSprite, inPos, display, outPos, inSize, roi ) ) {
-            fheroes2::Blit( monsterSprite, inPos, display, outPos, inSize, isReflected );
+            fheroes2::Sprite outMonsterSprite = monsterSprite;
+            if ( troop.isModes( Battle::CAP_MIRRORIMAGE ) ) {
+                fheroes2::ApplyPalette( outMonsterSprite, PAL::GetPalette( PAL::PaletteType::MIRROR_IMAGE ) );
+            }
+            fheroes2::Blit( outMonsterSprite, inPos, display, outPos, inSize, isReflected );
```
ihhub, line 452: To avoid extra operations for cases of no mirror image we should do like this:
```cpp
if ( troop.isModes( Battle::CAP_MIRRORIMAGE ) ) {
    fheroes2::Sprite outMonsterSprite = monsterSprite;
    fheroes2::ApplyPalette( outMonsterSprite, PAL::GetPalette( PAL::PaletteType::MIRROR_IMAGE ) );
    fheroes2::Blit( outMonsterSprite, inPos, display, outPos, inSize, isReflected );
}
else {
    fheroes2::Blit( monsterSprite, inPos, display, outPos, inSize, isReflected );
}
```

orzm: Done

ihhub (approved):

ihhub:
@orzm , many thanks for this improvement!

PR #9047, opened 2024-08-15
Replace reinterpret_cast with static_cast

oleg-derevenetz:
`reinterpret_cast` is not needed here.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for the correction!

PR #9048, opened 2024-08-17
Turn on the -Wmissing-declarations warning and move all detected global functions to the anonymous namespace

oleg-derevenetz:
GCC has the [-Wmissing-declarations](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wno-missing-declarations) warning option, which can be used to detect global functions that are not declared in header files (Clang also has an option with the same name, but it has a different meaning there and is enabled by default).

Also this PR fixes the incorrect behavior of the "Gift" button in the Marketplace dialog. In the `master` branch:

https://github.com/user-attachments/assets/e95a20f2-ba5c-43b3-bb4d-8ef9110cfcb4

With this PR:

https://github.com/user-attachments/assets/5acbec32-1ec3-4338-9368-2b13c12a75cf

It's obvious from the old code that the authors wanted to do something like this, but in the end it didn't work.

### DIFF
--- a/android/app/jni/Android.mk
+++ b/android/app/jni/Android.mk
@@ -38,6 +38,7 @@ FHEROES2_C_WARN_OPTIONS := \
 FHEROES2_CPP_WARN_OPTIONS := \
     -Wctor-dtor-privacy \
     -Wextra-semi \
+    -Wmissing-declarations \
     -Woverloaded-virtual \
     -Wsuggest-override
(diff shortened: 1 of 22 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for making the code cleaner!

PR #9051, opened 2024-08-18
Enable a few more compiler warnings, add more RAII for C-style "alloc()/free(), open()/close()"-like function pairs

oleg-derevenetz:
Close the long-standing TODO by enabling a few more useful compiler warnings. Unfortunately, SDL headers still cannot be compiled with them, and they are not treated as system headers since `sdl2-config --cflags` uses `-I` flag and not `-isystem` (gcc and clang do not issue warnings for system headers), so I had to disable new warnings for them using pragmas. Fortunately, we have relatively few files that use SDL headers.

Also this PR migrates a few places in the code to RAII.

Also this PR replaces custom `std::function`-based deleters in `std::unique_ptr`s by pointer-to-function-based. The reasons are as follows:

1. `std::unique_ptr`'s destructor as well as various methods like `reset()` require that deleter should not throw. However, with `std::function`, deleter may throw if we accidentally pass an empty deleter, while with function pointer it's just impossible to accidentally forget to specify the deleter. See [this example](https://godbolt.org/z/j34dPnar9) for details;
2. `std::function` may allocate heap memory and needs the appropriate header file.

### DIFF
--- a/android/app/jni/Android.mk
+++ b/android/app/jni/Android.mk
@@ -27,10 +27,12 @@ FHEROES2_C_WARN_OPTIONS := \
     -Wextra \
     -Wcast-align \
     -Wcast-qual \
+    -Wdouble-promotion \
     -Wfloat-conversion \
     -Wfloat-equal \
     -Wredundant-decls \
     -Wshadow \
+    -Wswitch-default \
     -Wundef \
     -Wunused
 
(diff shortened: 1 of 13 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this constant code cleanup!

I moved 1.1.2 release date to 8th September. I will be absent again till 3rd September and after that we can make a release.

PR #9055, opened 2024-08-21
Fix some code smells

oleg-derevenetz:
* C++20 introduced the [addressing restriction](https://en.cppreference.com/w/cpp/language/extending_std#Addressing_restriction) for stdlib functions (perhaps in order to "bless" the potential presence of "stdlib functions" that are not in fact functions - e.g. macros), so it's better be safe than sorry, especially if it doesn't cost anything;
* Rename various `xxx_t` identifiers;
* Move some module-local types & code to the anonymous namespace;
* Make the `skill_static.h` - related code more clear.


ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for this code cleanup!

PR #9060, opened 2024-08-23
Hungarian language small update 06

big4billy:


oleg-derevenetz:
Hi @big4billy thank you!

PR #9061, opened 2024-08-24
Optimize XMI to MID decode code

Districh-ru:
This PR removes some redundant operations in "xmi2mid" decoder (creating a new vector just to have a temporary copy of other vector's data, calculate binaryTime to discard it and recalculate, ...)

Decode time in master build `->` this PR:
```
xmi2mid time for [16] track is 560.9 -> 360.7 micro seconds
xmi2mid time for [15] track is 408.4 -> 223.4 micro seconds
xmi2mid time for [13] track is 232.4 -> 152.5 micro seconds
xmi2mid time for [14] track is 349.2 -> 200.1 micro seconds
xmi2mid time for [10] track is 396.3 -> 247.8 micro seconds
xmi2mid time for [19] track is 859.8 -> 649.2 micro seconds
xmi2mid time for [23] track is 1450 -> 1095.9 micro seconds
xmi2mid time for [22] track is 1413.1 -> 1054.1 micro seconds
xmi2mid time for [21] track is 1577.5 -> 965.8 micro seconds
```


src/engine/audio_xmi2mid.cpp
```diff
@@ -680,14 +652,16 @@ namespace
...
+    const MidData mid( xmi.trackEvents );
+
+    // Create a buffer for the midi data.
+    StreamBuf sb( mid.size() );
 
-    const MidData mid( xmi.tracks );
     sb << mid;
 
-    return { sb.data(), sb.data() + sb.size() };
+    return { sb.data(), sb.data() + sb.capacity() };
```
oleg-derevenetz, line 666: Hi @Districh-ru please note that even if `mid.size()` is less than `minBufferCapacity` (currently it's 1024 bytes), the `StreamBuf`'s `capacity()` will be not less than `minBufferCapacity`, so the resulting vector may be larger than necessary. Please see the `StreamBuf::reallocbuf()` implementation for details.

Districh-ru: @oleg-derevenetz, thanks for clarifying this. I reverted the change.
In our case all midi file are bigger than 1024 bytes so the both variants work. :)

src/engine/audio_xmi2mid.cpp
```diff
@@ -295,95 +296,85 @@ namespace
...
+            trackEvents.data = buf.cbegin() + static_cast<ptrdiff_t>( sb.capacity() - sb.size() );
+            trackEvents.dataEnd = buf.cend();
 
-                if ( evnt.size() != iff.length ) {
-                    ERROR_LOG( "parse error: "
-                               << "out of range" )
-                    break;
-                }
+            // The single track XMI files does not have any data after the EVNT data.
+            if ( ( trackEvents.data + iff.length ) != trackEvents.dataEnd ) {
```
oleg-derevenetz, line 332: Please note that you cannot just add an arbitrary value to the pointer. Even if you don't dereferencing the resulting pointer, it still result as an [undefined behavior](https://en.cppreference.com/w/cpp/language/operator_arithmetic#Pointer_arithmetic).

Districh-ru: Thanks for noticing this!
Fixed.

src/engine/audio_xmi2mid.cpp
```diff
@@ -618,60 +612,38 @@ namespace
...
-        {
-            // MIDI format 0 can contain only one track
-            assert( tracks.count() == 1 );
-
-            // XMI files play MIDI at a fixed clock rate of 120 Hz
-            if ( !tracks.empty() && tracks.front().events.trackTempo > 0 ) {
-                ppqn = ( tracks.front().events.trackTempo * 3 / 25000 );
-            }
+            // Every header has 2*uint8_t
+            return 8U + mthd.length + 8U + track.mtrk.length;
```
oleg-derevenetz, line 635: To be honest, the logic here doesn't look particularly transparent, despite the comment above :) Is there a way to elaborate or rework it somehow? Otherwise, it is not entirely clear where these constants come from and what 2*uint8_t has to do with it.

Districh-ru: I've updated the logic and the comment.

oleg-derevenetz: Taking this opportunity, replaced the various `size()` methods with `sizeInBytes()`. For instance, `MidiEvents::size()` shadowed the inherited `std::vector::size()`, which honestly didn't look good.

oleg-derevenetz (approved):
Hi @Districh-ru I just made a few minor changes, please review them whenever possible. LGTM otherwise.

Districh-ru:
> Hi @Districh-ru I just made a few minor changes, please review them whenever possible. LGTM otherwise.

Hi, @oleg-derevenetz, thanks for making the code better!

ihhub (approved):

ihhub:
@Districh-ru , many thanks for these improvements!

PR #9062, opened 2024-08-24
Add filename hash check in AGG files

Districh-ru:
This PR adds the filename has check while reading the AGG files.

### DIFF
--- a/src/engine/agg_file.cpp
+++ b/src/engine/agg_file.cpp
@@ -20,6 +20,7 @@
 
 #include "agg_file.h"
 
+#include <cctype>
 #include <cstdint>
 #include <iterator>
 #include <string>
--- a/src/engine/agg_file.cpp
+++ b/src/engine/agg_file.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2020 - 2022                                             *
+ *   Copyright (C) 2020 - 2024                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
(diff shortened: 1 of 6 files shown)


src/engine/agg_file.cpp
```diff
@@ -18,24 +18,46 @@
...
 
+#include "agg_file.h"
+
 #include <cstdint>
 #include <string>
 
-#include "agg_file.h"
+namespace
+{
+    uint32_t aggFilenameHash( const std::string & fileName )
```
oleg-derevenetz, line 28: At the time, I wrote the [hash calculation function](https://github.com/ihhub/fheroes2/blob/5d8276ba5d244136e5f352046a4563966521abf3/src/tools/extractor.cpp#L58-L74) for the `extractor` tool. We can put it to the `tools.h` and use it in both places.

Districh-ru: Ow, I missed it. :(
Sure, @oleg-derevenetz, thanks for pointing this out!
I'll update the code.

oleg-derevenetz: To avoid all these issues with MSVC linking, we can move this function to the `agg_file.h` as an inline function. After all, logically it may even fit better here. `tools.h` is a bit too "generic".

Districh-ru: I moved this function to `agg_file` and added it to sources.props for Extractor.

Probably in future we can expand the `AGGFile` class and use in extractor.

Districh-ru:
Hi @ihhub and @oleg-derevenetz, I will be away until September 4th. If you find any issues in my opened PRs, you can directly fix them.

oleg-derevenetz (approved):

oleg-derevenetz:
@Districh-ru thank you!

PR #9064, opened 2024-08-25
Update of Ukrainian Translation

ZaHaDumGoth:
Upper cases, endings and dots at the end.


oleg-derevenetz:
Hi @ZaHaDumGoth thank you!

Issue #8924, opened 2024-07-06
Adventure map, allow to delete the hero's predefined path on the adventure map regardless the state of the "Hero Movement" button


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

When the selected hero has movement points left, it is possible to delete the hero's predefined path on the adventure map by doing a left click held on the "Hero movement" button.
However, when the hero has no more movement points (the "Hero Movement" button is "grayed out"), it is no longer possible to do so.
I suggest that it should be possible to delete the hero's predefined path on the adventure map regardless of whether the "Hero Movement" button is "grayed out" or "not grayed out".

In fheroes2:

https://github.com/ihhub/fheroes2/assets/43583994/15a22546-a551-4256-b661-a44a4b71c6ed

### Describe the possible solution

Allow to delete the hero's predefined path on the adventure map regardless the state of the "Hero Movement" button.

AlexSnowLeo:
Hi @LeHerosInconnu. This is the expected behavior, am I right?

https://github.com/user-attachments/assets/25c412e4-724d-49b5-9610-eaf09b345ec5

LeHerosInconnu:
Hello @AlexSnowLeo,

> Hi @LeHerosInconnu. This is the expected behavior, am I right?
> 
>  https://github.com/user-attachments/assets/25c412e4-724d-49b5-9610-eaf09b345ec5

Yes, exactly. :)

PR #7979, opened 2023-10-22
Add hero action button image

Districh-ru:
This PR adds an image for hero action button (relates to #7849)

изображение изображение
изображение изображение

https://github.com/ihhub/fheroes2/assets/113276641/fd60db58-985a-4c7d-b148-1e1aeabdb218

PR #9085, opened 2024-09-05
Update Hero Movement button

AlexSnowLeo:
Close this issue: #8924
Fixed check on Hero Movement button to delete hero path always regardless of the button status.


oleg-derevenetz:
Despite the fact that it technically works, it doesn't look very elegant. First of all, this button is disabled, that is, from the user's point of view, it cannot be pressed. Also, since it is disabled, there is no press and release animation.

AlexSnowLeo:
> Despite the fact that it technically works, it doesn't look very elegant. First of all, this button is disabled, that is, from the user's point of view, it cannot be pressed. Also, since it is disabled, there is no press and release animation.

Hi @oleg-derevenetz. In this case, I see two options:
- button can be pressed in the disabled state
- add another state for the hero movement button, which will show that there is a hero path, but there are no movement points;

Could you describe how you see the solution to this problem?

oleg-derevenetz:
I'd prefer to add a special state for this case, just like it's done for the "visit the object under the hero" action, but we need some artwork for this, which would somehow indicate the possibility of canceling the current hero route.

AlexSnowLeo:
> I'd prefer to add a special state for this case, just like it's done for the "visit the object under the hero" action, but we need some artwork for this, which would somehow graphically indicate the possibility of canceling the current hero route.

Do we need to wait for the icon design for this feature? Or is it possible to use some existing image?

oleg-derevenetz:
> Do we need to wait for the icon design for this feature? Or is it possible to use some existing image?

I am not aware of any existing image with the suitable style and meaning, so we need to come up with and draw it first.

AlexSnowLeo:
> > Do we need to wait for the icon design for this feature? Or is it possible to use some existing image?
> 
> I am not aware of any existing image with the suitable style and meaning, so we need to come up with and draw it first.

Do you have any instructions on how to do this? Or is it possible to ask for help from a certain person (interface designer)?

oleg-derevenetz:
> Do you have any instructions on how to do this? Or is it possible to ask for help from a certain person (interface designer)?
I could try to draw a new icon based on the current one, but how do I get the current one and add a new one to the game data?

I see two options here:

1. You can generate a new button image based on existing button image, like it's done in #7979;
2. You can draw a new button and add it to the `resurrection.h2d`, like it's done e.g. in #8599. You will need to use the `h2dmgr` tool (see the contents of `src/tools`) to add a JPG or PNG image to our data file. These tools are available for certain platforms as separate archives in the Releases section (e.g. [here](https://github.com/ihhub/fheroes2/releases/tag/fheroes2-windows-x64-SDL2) for Windows) or you can build them yourself.

I think the choice of option depends on the desired button image (it will also need to be thought up).

AlexSnowLeo:
> I think the choice of option depends on the desired button image (it will also need to be thought up).

The idea is something like this. Show a red destination at the bottom, as in the additional icon with a horse on end.

oleg-derevenetz:
> Maybe we should bring this up for discussion?

Hi @ihhub @zenseii @Districh-ru @Branikolog what do you think about the image? Any ideas?

AlexSnowLeo:
> @oleg-derevenetz and @AlexSnowLeo. The concept in the picture is pretty much what I had in mind. Since we are making a pressable replacement for the disabled state of the button, in other words when there are no movement points left, maybe something like a horse that has been tethered to a short pole, to symbolize having parked for the day.

Unfortunately, I am far from an artist. Maybe something like this? I will also attach the source code in the format paint.net.

[002.zip](https://github.com/user-attachments/files/16917054/002.zip)

ihhub:
@AlexSnowLeo and @oleg-derevenetz , let's wait for @Districh-ru being available. He generated the original icon and he could suggest how to do it for the new one.

oleg-derevenetz:
> Unfortunately, I am far from an artist.

We are also not talented artists, so making changes that require new art is usually problematic. But in this case, I'm afraid we can't do without it if we want everything to look nice.

AlexSnowLeo:
Hi @Districh-ru. Yes, I have already considered this place in the code and tried to draw something suitable from the existing pictures of the game. But unfortunately there is no suitable picture (perhaps the image of the hero on a horse is suitable, but it is too colorful). Maybe there is not only another way to go? Make a mask of the image of a standing horse. And then apply it in the code along with the shadow on the button. The problem for me so far is only in shadow rendering. And let's still choose a concept, is this the one that @zenseii suggested - a horse standing on a leash?

Districh-ru:
Hi, @AlexSnowLeo, sorry for the late reply.
Yes we can choose @zenseii's concept.
We can use the original button image, but redraw the legs and draw a pole.
If you have questions or issues making the image or its shadow - please write.

Hi, @Branikolog, what do you think about this image for the cancellation of move path?

FlorianBaba:
Hello there 👋 

I would indeed love to participate.
I just need to think about it a bit to come up with a suitable icon. 
But it shouldn't take me too long, I'll try to do it next week!

I'll keep you posted 👍

FlorianBaba:
Hi everyone 👋 

I've spent some time thinking about a suitable icon, trying to respect the original style. 
Here are 4 suggestions:

horse_icon_A A 
horse_icon_B B 
horse_icon_C C 
horse_icon_D D

Just to share you my approach, I first tried @zenseii idea, which I thought was a good one. But I came to the same conclusion : i.e. that the icon is visually overloaded and not easily understandable.

Then I went back to @AlexSnowLeo original idea, using the destination cross of the route taken by the heroes on the map. I tried to make the cross more prominent by using a horse's head icon.

FlorianBaba:
We talked it over with @zenseii and thought it would be interesting to add a new suggestion, the one he mentioned in his previous message.

horse_icon_E E

It's up to you to you now ^^

Branikolog:
Hello everyone! 
Thanks for making so many choices. :)

I personally prefer D from an implementation quality state. 

> horse_icon_D D

Not perfect one, though.

Logically, I expect something like @AlexSnowLeo proposed. Just a standing horse. We already have galloping and dancing icons, so new standing one feels appropriate. I'm only concerning about the horse design. Needs to be more polished and better proportioned. I also not sure about cross or ! mark. It looks too tiny. I prefer an icon being more simple: without multiple small details.


 

I personally thought about placing cross, which is used for path destination, on this button (
image
)

It can look simple and really easy to distinguish from other icons.

But the standing horse feels kind of uniform design. 👍  It just needs to be more proportional, and players should always be able to easily distinguish it from the other two button states, especially the galloping one.

FlorianBaba:
@zenseii I've tried to take your feedback into account for the E version.
It's not easy on very small icons, because each added pixel completely changes the perception.

So :
- The tail is lengthened and revised
- A few pixels have been removed to make the horse less plump.

horse_icon_Ev2 E v2

Yyour feedback on the ears made me smile, because I had asked myself the same question as you the first time.
But after checking, the ears point upwards when the horses are grazing.

horse1
horse2

Branikolog:
@zenseii 
>@Branikolog, I also considered having only the cross, but I believe it will look incoherent if the button switches states between Galloping horse, Prancing horse and a cross. This is why I'm in favor of the horse standing still.

I agree with this. The only thing I was concerning about is that two icons with a horse standing horizontally could be not distinguishable especially on tiny mobile screens. We should come up with a design which allow player to easily notice and understand the difference between these button states.



> * The tail is lengthened and revised
> * A few pixels have been removed to make the horse less plump.
> 
> horse_icon_Ev2 E v2

I have few suggestions on polishing this design even more, but I'm still hesitating about the concept. :)
The proposed icon brings me in mind a conception of resting, rather than cancelling. Such icon could fit hero "sleep" mode better in my opinion. :)

Since we work on hero path cancelling, just standing still horse logically fits more in my opinion. So, how do you guys think of placing just standing horse for this icon? Or we can go even further and draw horse during "braking" process :).
Something like that:

horse stop2

LeHerosInconnu:
Hello everyone,

Here's a proposal with just a few lines to symbolize a sense of movement (in the case of the predefined path displayed on the adventure map), without altering the original graphics too much.
In order of preference.

First:


Second:


Third:

FlorianBaba:
Hello again 👋 

@zenseii and I have had a lot of discussions on the subject, and we've come to the conclusion that since the button will have 3 versions, it's better that the three illustrations look the same and that you can see the continuity when moving from one to the other.

So here's a new proposal :
horse_icon_F7

And a gif version to show you the consistency with the other versions of the button
Version_3_revamp5

Branikolog:
@zenseii & @FlorianBaba 
The last proposal seems good for me. I was afraid, that standing horse would be hard to distinguish, but because of completely different legs design it is quite legible and the overall design looks uniform. 👍 
Indeed, standing still horse suits the idea of a hero without any selected objective. (And better, than grazing horse).

Districh-ru:
> @Districh-ru, I believe we should implement this new icon by generating it on the fly in the engine, rather than adding it to fheroes2 resources, because some parts are the same as the galloping horse. What do you think? Also feel free to give any feedback on the proposed icon.
> 
> We should push the changes adding the new icon to this PR.

I like the icon proposed by @FlorianBaba, thanks!

a) We can add only the changed parts to the engine resources and make a new icon by taking galloping horse and "Blitting" the changed part (the unchanged pixels will be transparent) on it during the ICN load.
b) Or we can also update the image on fly by erasing the changed part and drawing them using engine functions.

The second approach needs more careful icon "drawing" in the code. ... and it will make out `agg_image.cpp` bigger ...

Hi, @ihhub, which approach do you prefer?

ihhub:
Hi @zenseii and @Districh-ru , if the image difference is small and easy to generate then we can go with storing it in fheroes2 resources. If the image generation is easier to do then we can go with image generation approach. The size of the code is not an issue.

AlexSnowLeo:
> Hello again 👋
> 
> @zenseii and I have had a lot of discussions on the subject, and we've come to the conclusion that since the button will have 3 versions, it's better that the three illustrations look the same and that you can see the continuity when moving from one to the other.
> 
> So here's a new proposal 
> And a gif version to show you the consistency with the other versions of the button 

@FlorianBaba Can you send me just a clean mask of a standing horse, without unnecessary details? I will try to encode this and apply a shadow using the game engine on the button substrate.

AlexSnowLeo:
> @AlexSnowLeo, the reason we are talking about just including the image differentials is due to copyright. The standing horse is derived from the galloping one and therefore we can't just take the entire standing horse and supply it with the engine.

It is for this reason that I initially proposed to make only a modification of the current image. For example, add only the cross symbol of the path:
image

FlorianBaba:
> @FlorianBaba Can you send me just a clean mask of a standing horse, without unnecessary details? I will try to encode this and apply a shadow using the game engine on the button substrate. ([alex.v.timofeev@gmail.com](mailto:alex.v.timofeev@gmail.com))


Yes, no problem, I'll send you an e-mail later today 👍

AlexSnowLeo:
> @AlexSnowLeo, the reason we are talking about just including the image differentials is due to copyright. The standing horse is derived from the galloping one and therefore we can't just take the entire standing horse and supply it with the engine.

@zenseii, @ihhub and @Districh-ru 
I did a little demo in the image editor. How can the current prancing horse icon in the code be modified approximately? Please take a look. Nevertheless, the tail section requires more detailed rendering, although the legs can be copied :)

https://youtu.be/rfMVEBdVCu4

The resulting button image, without shadow:
изображение

Districh-ru:
Hi, @zenseii, I'll take a look in the next days.
I believe we can use `updateShadow()` instead of putting many "shade" pixels: make an image on a transparent background, apply `updateShadow()` and then `Blit()` this image to the empty button.

Districh-ru:
The logic to disable the path is to hold the left mouse button (LMB) over the Move button when hero has movement points.
When he has not I used the same logic, but made the button disabled showing the moving horse which means that the movement is planned. If you hold the LMB then the path will be cleared and the disabled button with the standing horse will be shown meaning than no movement is planned:

https://github.com/user-attachments/assets/c8166e51-9b77-4b9c-8811-724614a9f6a6

IMHO making the standing horse button enabled will make the player think that it can be clicked (not long holding) and making it so will be different with the logic for the case when there are movement points left. And making the button enabled but allowing only long hold will make player thing that this button is buggy.

This logic can be changed, what are your suggestions @Branikolog, @LeHerosInconnu, @oleg-derevenetz, @zenseii ?

Districh-ru:
@zenseii, do you mean this?

https://github.com/user-attachments/assets/93d05252-5b8a-4f7e-808f-4d98654c1305

The issue is that currently this button is also pressable my the "M" key...

But I'm voting for keeping the current master build logic: to reset the path you need to hold (not click) this button.
And the better solution would be to add this button and do no replacing, but we are out of space in ...x480 resolution.

Tomorrow I'll update some icon's shadows to more correspond the original one:
изображение изображение

Districh-ru:
@zenseii,
got it :)

https://github.com/user-attachments/assets/65de64a0-e9a5-4b5c-8594-efacc76228b5

Branikolog:
Hi all!
I agree with @Districh-ru , that the best implementation for re-visit button is to place it separately within the GUI.
I personally don't even understand the purpose of deleting the route, when a hero is not standing on any object. The only purpose for deleting the route currently is the case, when a hero has movement points and is standing on an object.
Regardless, the proposed implementation works well, but looks kind of complicated and redundant in some cases.

LeHerosInconnu:
Hello everyone,

> This logic can be changed, what are your suggestions @Branikolog, @LeHerosInconnu, @oleg-derevenetz, @zenseii ?

This mode of operation would be what would best correspond to what needs to be achieved.
> https://github.com/user-attachments/assets/c8166e51-9b77-4b9c-8811-724614a9f6a6

This operation is confusing, with a button that can be pressed just to clear the predefined path.
> https://github.com/user-attachments/assets/93d05252-5b8a-4f7e-808f-4d98654c1305

This operation is even more confusing than the previous one, with a pressable button that doesn't perform any action when pressed (all other buttons in their non-grayed state trigger an action with a single click), and which only works with a held click.
> https://github.com/user-attachments/assets/65de64a0-e9a5-4b5c-8594-efacc76228b5

The first operating mode would be the most compliant and appropriate.
With a single click or a click and hold to activate the action, or with only a click and hold to activate the action.
But for my part, I don't think this extra button state is really necessary; originally it was just a matter of being able to clear the hero's path with a prolonged click whether the button was grayed out or not: https://github.com/ihhub/fheroes2/issues/8924.
I think this extra button state is just complicating life for little benefit, as well as confusing the user.
The operation implemented by @AlexSnowLeo here https://github.com/ihhub/fheroes2/issues/8924 is sufficient.

Branikolog:
Hi, @LeHerosInconnu !

>originally it was just a matter of being able to clear the hero's path with a prolonged click whether the button was grayed out or not

I believe, that clicking and holding inactive button is confusing and is not like any button is supposed to operate. Inactive button = means no action could be performed.

Since you've opened the issue, could you, please, explain the purpose of deleting the route, when a hero is not standing on any object? :)

LeHerosInconnu:
@Branikolog,

> Hi, @LeHerosInconnu !
> 
> > originally it was just a matter of being able to clear the hero's path with a prolonged click whether the button was grayed out or not
> 
> I believe, that clicking and holding inactive button is confusing and is not like any button is supposed to operate. Inactive button = means no action could be performed.
> 
> Since you've opened the issue, could you, please, explain the purpose of deleting the route, when a hero is not standing on any object? :)

Let's imagine that a predefined path is displayed for my hero to go to a mine.
Now I change my mind and decide to wait until the next turn to confirm that the hero will go to the mine.
So I want to remove the predefined path displayed so that it's clear to me (no path displayed for the hero at the start of the next turn).
Now, if the hero has movement points left, I can remove the predefined path, but I can't do it if there are no movement points left, hence the opening of the issue.

Branikolog:
@zenseii @LeHerosInconnu 

Thanks for explanation! :)

The proposed implementation is OK for me. The only thing I'm concerned is the resetting the route, when a hero has no one:
https://github.com/user-attachments/assets/fcff733b-d54e-4eb9-97b8-0162e862def4

In this case I suggest proceeding the visit of the object, when player holds the button a little more longer than needed.

LeHerosInconnu:
@zenseii,

> So ideally, in the future, this standing horse would be a separate button in the UI and replace the press and hold interaction.

Better in this case not to confuse the user with a button status that will be separated later.
The button's operation with a click and hold is sufficient, and the user is informed of this in the button's information window.

In fheroes2:

Hero movement button info window 01

LeHerosInconnu:
@Branikolog,

> @zenseii @LeHerosInconnu
> 
> Thanks for explanation! :)

:)

> The proposed implementation is OK for me. The only thing I'm concerned is the resetting the route, when a hero has no one:
> 
>  2025-01-17.13-49-50.mp4 
> In this case I suggest proceeding the visit of the object, when player holds the button a little more longer than needed.

Indeed, since there's no predefined path displayed, the held click, regardless of duration, should also make the hero revisit the castle (same functioning as the other buttons).

Branikolog:
> Better in this case not to confuse the user with a button status that will be separated later.
> The button's operation with a click and hold is sufficient, and the user is informed of this in the button's information window.

At some point I agree with that. But as a result there would be 0 cases when we see inactive move button. Probably that is the reason I consider using a new button: not to confuse old players with active move button (old one) as they got used to the behaviour of the button, which implies some available movement points in a case it is in active state.

LeHerosInconnu:
@Branikolog,

> > Better in this case not to confuse the user with a button status that will be separated later.
> > The button's operation with a click and hold is sufficient, and the user is informed of this in the button's information window.
> 
> At some point I agree with that. But as a result there would be 0 cases when we see inactive move button. Probably that is the reason I consider using a new button: not to confuse old players with active move button (old one) as they got used to the behaviour of the button, which implies some available movement points in a case it is in active state.

Yes, I think the states of the movement button should remain as they are now.
The standard and grayed out states of the resting horse button will be used separately for another dedicated button.

Branikolog:
> Yes, I think the states of the movement button should remain as they are now.

As I mentioned before, we shouldn't allow to click and hold on disabled movement button, as it looks not that good. To initiate route reset the button should be always in active state, so player could confirm, that something is happening. So not to confuse old players with active state of the old movement button (running horse) I suggest using the new button with standing horse for the case hero has 0 movement points. It's not that ideal solution, but up to implementation of the expanded UI I think it a compromised solution.

Districh-ru:
The "inactive horse" button generation code was changed to reduce the code lines and to have shadows a little closer to the other UI buttons:
изображение

Districh-ru:
@zenseii, thanks for the detail comparison and explanation!
I'll try to update the shape and the shading tomorrow using the new code. And if it'll be hard or the code will be big enough I'll revert the changes.

Districh-ru:
Hi @zenseii!
I've reverted the icon generation code but updated it a little. :)
Also I've addressed the issue @Branikolog mentioned here: https://github.com/ihhub/fheroes2/pull/9085#issuecomment-2598042753

LeHerosInconnu:
Hello @Branikolog,

> > Yes, I think the states of the movement button should remain as they are now.
> 
> As I mentioned before, we shouldn't allow to click and hold on disabled movement button, as it looks not that good. To initiate route reset the button should be always in active state, so player could confirm, that something is happening. So not to confuse old players with active state of the old movement button (running horse) I suggest using the new button with standing horse for the case hero has 0 movement points. It's not that ideal solution, but up to implementation of the expanded UI I think it a compromised solution.

The state of the button also serves as a visual indicator to the player that the hero still has movement points or not (button in standard state, hero can move, button in grayed out state, hero can no longer move).
Now, if a different button is introduced and in its standard state the hero has no more movement points, this initial way of information is lost.

ihhub:
> @Districh-ru
> 
> > The code is big, but I have no idea how to make is smaller since we want to make a better than functionally generated shading for the horse icon than functional generation can give now. Or should we add the generated image parts to **resurrection.h2d**? (maybe in the next PR and not only for this image...)
> 
> This could be a possibility and honestly a better approach to take moving forwards since it took quite a lot of time to write every line for almost every pixel. Basically the only original part used was the horse's head and somewhat the legs. But I do remember @ihhub saying that the amount of lines didn't really matter much.

From maintenance point of view having an image difference stored in **resurrection.h2d** file would definitely be better. Yes, let's do this in a separate pull request to avoid extra work for this issue.

ihhub (approved):

src/fheroes2/gui/interface_buttons.cpp
```diff
@@ -178,7 +178,12 @@ fheroes2::GameMode Interface::ButtonsPanel::queueEventProcessing()
     else if ( _buttonHeroMovement.isEnabled() && le.MouseClickLeft( _heroMovementRect ) ) {
         res = _interface.EventHeroMovement();
     }
-    else if ( _buttonHeroMovement.isEnabled() && le.MouseLongPressLeft( _heroMovementRect ) ) {
+    // The 'MouseLongPressLeft()' for '_heroMovementRect' should be called only if current hero
+    // has valid path not to trigger the '_mouseButtonLongPressDelay'.
+    // The inlined 'isMouseLeftButtonPressed()' method is called not to call 'isValidForMovement()'
+    // method every time this button is enabled.
+    else if ( _buttonHeroMovement.isEnabled() && le.isMouseLeftButtonPressed() && GetFocusHeroes()->GetPath().isValidForMovement()
```
oleg-derevenetz, line 185: Hi @Districh-ru shouldn't the `isMouseLeftButtonPressedInArea()` be used here instead of just `isMouseLeftButtonPressed()`?

Districh-ru: Hi @oleg-derevenetz, you are absolutely right!
`isMouseLeftButtonPressedInArea()` is better: it is also inlined and `isValidForMovement()` will be called only if this button is pressed, not any other.
Thanks!

src/fheroes2/gui/interface_events.cpp
```diff
@@ -197,7 +197,7 @@ fheroes2::GameMode Interface::AdventureMap::EventHeroMovement()
 void Interface::AdventureMap::EventResetHeroPath()
 {
     Heroes * hero = GetFocusHeroes();
-    if ( hero == nullptr ) {
+    if ( hero == nullptr || !hero->GetPath().isValidForMovement() ) {
```
oleg-derevenetz, line 200: Hi @Districh-ru Could you please explain if this change is somehow justified from the point of view of game logic, or is it just some kind of optimization?

If it's an optimization, then I'd say that it's a "counter-optimization" :) Resetting an already empty path is no-op, while this additional check checks both that the path is not empty and that it is suitable for normal movement, which in this case looks pointless (because "non-normal" movement is available only for AI-controlled heroes).

If you just need to check that the path is empty, then it's better to entrust this work to the `Path::Reset()` below, which in turn just calls "std::list::clear()", which will make all the necessary checks faster and in a more efficient way.

Districh-ru: Hi @oleg-derevenetz,  yes, it is an optimization ant it is not only to bypass `Path::Reset()` but to also bypass `setRedraw( REDRAW_GAMEAREA | REDRAW_BUTTONS )` and do not force these redraws.

oleg-derevenetz: @Districh-ru But as I can see the `EventResetHeroPath()` is already called only if `isValidForMovement()` for the current hero is true (the only place where it is called is line 193 of `interface_buttons.cpp`), is there a point to re-check this once again?

Districh-ru: Ow, yes, currently it is used only in `interface_buttons.cpp` where this check already is done. 
I've reverted this redundant check.
But this function is not often called so this check should not slow the engine as well as an extra redraw of game area and buttons. :)

And should we check only `!currentHero->GetPath().empty()` instead of `currentHero->GetPath().isValidForMovement()` in `interface_butons.cpp`, line 190?

oleg-derevenetz: @Districh-ru

> And should we check only !currentHero->GetPath().empty() instead of currentHero->GetPath().isValidForMovement() in interface_butons.cpp, line 190?

It depends on your taste mostly :) We can leave it as is.

ihhub:
@AlexSnowLeo , @zenseii and @Districh-ru , many thanks for this feature!

Issue #8589, opened 2024-04-24
Editor post 1.1 release changes


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

After the release of the Editor we still have a long list of tasks to do. This issue serves as a collections of these tasks in order to keep them in one place.

### Describe the possible solution

- [ ] allow to place heroes over the passable parts of the objects
- [ ] show the selected object's passability while placing it (e.g. like in original editor)
- [ ] add a possible functionality for scripting support for objects and the overall map (just expand the map format for the future changes support)
- [ ] change Adventure Map event functionality to support extended format
- [ ] implement zoom option
- [ ] implement the alternative "Kingdom Overview" (maybe we can call it "Map Summary") for Editor to show all castles, heroes, mines (and maybe some other objects) on the map with ability to edit these object's details
- [ ] display some object's metadata in the quick info (by the right mouse button)

ihhub:
I am going to create a set of GitHub issues based on this list since the Editor went public so it would be easier to manage things separately.

LeHerosInconnu:
Hello @ihhub,

Definitely, a "REDO" button is needed. :)

big4billy:
Hello @ihhub,

I installed the new version 1.1 but the Editor button (the globe) doesn't appear. I uninstalled the old one and reinstalled the whole game. Os: Win10; Game: v1.1 x64

ihhub:
Hi @big4billy , are you sure that you use Price of Loyalty version of the game?

big4billy:
Hi! Hmm, I have the Gold version from GOG. Is that not proper for the Editor?

ihhub:
@1GodRage ,
- can you put lava pool the way you want in the original editor?
- not for 1.1.1
- what do you mean by that? All dialogs remember your last selection
- where to change?
- press R (see the hotkey list dialog)
- once you select an object you can place as many as you want
- the same way as in the original game
- the engine doesn't support text buffer at the moment

lwollnikowski:
> * generate unique new map names and file names to avoid accidental file rewriting

Is it fine to simply generate name like `My Map YYYYMMDD_hhmmss` on map creation? Or am I missing something?

PR #9089, opened 2024-09-07
Do not discard area selection while moving mouse out of map

ihhub:
After the changes:

https://github.com/user-attachments/assets/a15d5391-6eae-4cde-a87a-3357890f25b6

relates to #8589


ihhub:
Hi @zenseii , can you please verify the functionality as @Branikolog seems offline for many days?

ihhub:
Hi @zenseii , the problem is that a user can release mouse button once the cursor out of the Adventure Map area. This might lead to much worse scenarios if we allow to do that.

Districh-ru (approved):

PR #9092, opened 2024-09-07
Remove all remaining const_cast conversions

oleg-derevenetz:
Continuation of the #9067 (but much easier).

Although they are not harmful at the moment, they are potentially dangerous things that should be avoided whenever possible.


ihhub (approved):

ihhub:

PR #9096, opened 2024-09-08
Preserve the order of artifacts when adding the Spell Book to the artifacts bag

### LINKED ISSUE #9093: Hero screen, conserve the order of artifacts when the hero buys a magic book while already in possession of artifacts
### Preliminary checks

In fheroes2, the order of artifacts when the hero buys a magic book while already in possession of artifacts should be conserved.

In fheroes2:
Before buying the magic book:
Before magic book artifacts order 001

After buying the magic book:
Before magic book artifacts order 002

https://github.com/user-attachments/assets/806911ad-f24b-49ff-be0d-b6f5b4b0823c

Here is the save file:
[Before magic book artifacts order 01.zip](https://github.com/user-attachments/files/16921902/Before.magic.book.artifacts.order.01.zip)

The location of artifacts positioned by the player should be preserved as closely as possible.
Artifacts are moved from left to right and top to bottom.
If there is already an artifact in the slot that will receive the magic book, the artifact is moved one slot to the right.
If there is already an artifact in the slot where the artifact is to be moved, this artifact is moved one slot to the right, and so on.

### Additional info

Related closed issue: https://github.com/ihhub/fheroes2/issues/1644.

oleg-derevenetz:
close #9093

### DIFF
--- a/src/fheroes2/resource/artifact.cpp
+++ b/src/fheroes2/resource/artifact.cpp
@@ -833,10 +833,8 @@ bool BagArtifacts::PushArtifact( const Artifact & art )
         return false;
     }
 
+    // There should not be more than one Magic Book in the artifact bag at a time.
     if ( art.GetID() == Artifact::MAGIC_BOOK && isPresentArtifact( art ) ) {
-        // We add a magic book while adding a hero on the map. If the map creator also added a
-        // Magic Book to the hero's artifacts, then we are faced with a situation of two Magic
-        // Books, which should not be the case.
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub I'm not sure about the timing of the upcoming release, if it is close, then this PR can be postponed until the next release.

ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for this fix! I am planning to do a release in the next few days. It is being delayed as I am still facing a lot of personal issues at the moment.

PR #9068, opened 2024-08-27
endian_h2: fall back to POSIX standard endian.h

alarixnia:
Fixes build on NetBSD.

PR #9098, opened 2024-09-10
Fix some Windows-specific SDL-related encoding issues

oleg-derevenetz:
SDL accepts and returns strings (for example, file paths, window titles, and so on) in UTF-8 encoding. However, on Windows fheroes2 works with locale-specific ANSI encoding (CP_ACP), which in the vast majority of cases is not UTF-8, therefore, when calling SDL functions and processing the returned results, a conversion must be performed. This conversion has already been partially implemented before, but not in all cases. This PR:

* Adds CP_ACP -> UTF-8 conversions for SDL/SDL_image function calls in the `image_tool.cpp` module. Before these changes, tools using this module had problems with images and directories with non-ASCII names on Windows (e.g. `icn2img` tool was not able to extract ICN file to a directory with non-ASCII name);
* Adds UTF-8 -> CP_ACP conversion for the `SDL_GetPrefPath()` result. This function, for example, returns a valid path to `APPDATA` on Windows, but previously it could not be used in cases where the username contained non-ASCII characters;
* Adds the CP_ACP -> UTF-8 and UTF-8 -> CP_ACP conversions for window title where appropriate. Not that it solves anything, just for the sake of order;
* Adds caching of `System::GetConfigDirectory()` and `System::GetDataDirectory()` results. Locations of the app config and data directories, in theory, should not change while the app is running, so they can be cached just fine;
* Replaces various system-dependent I/O APIs by `std::filesystem`;
* Unloads `engine/tools.*` a bit, which has turned into a dump of everything in a row (as it usually happens with tools) - math tools has been moved to new `math_tools` module, locale aware `StringReplaceWithLowercase()` has been moved to `translations`;
* The contents of the `endian_h2.h` has been moved to `serialize.h`, since it is the only place where it was used, and the fallback to the new POSIX 1003.1-2024 from #9068 has been added.


oleg-derevenetz:
Hi @ihhub

> please test this PR as the first priority.

We can postpone it after the release of fheroes2 1.1.2, if necessary. Tools are used by developers, so there is no big issue if they will not work with non-ASCII file paths on Windows some more time, and caching directories is not a critical feature for playability. Everything else like moving to `std::filesystem` is just a simplification of the code.

ihhub:
@oleg-derevenetz , sure.

ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for these updates!

PR #9102, opened 2024-09-15
Prevent redistribution of hero's last unit

filipni:
Moving a hero's last unit by dragging and dropping caused the game to crash.
dialog-error
This PR adds a check that returns early if the last unit should be saved.

### DIFF
--- a/src/fheroes2/army/army_bar.cpp
+++ b/src/fheroes2/army/army_bar.cpp
@@ -62,23 +62,28 @@ namespace
         assert( overallCount > 0 );
 
         if ( overallCount == 1 ) {
-            // prevent cross-army split if we lose the last unit in the hero army
+            // Prevent units from moving between armies if this would result in no units remaining in the hero's army
             if ( saveLastTroop ) {
                 return;
             }
 
-            // it seems that we are just moving a single unit to a free cell
+            // It seems that we are just moving a single unit to a free cell
             assert( !troopTarget.isValid() );
 
             Army::SwapTroops( troopFrom, troopTarget );
         }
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (commented):
Hi @filipni, The logic you propose removes some of the functionality of the troop split window. If we think about it a little bit, then we may ask - why this functionality:

https://github.com/user-attachments/assets/e12e8277-12f0-49fd-8ed3-4ff11931478d

should work for a slot with 2 units, but shouldn't work for a slot with 1 unit? It doesn't seem logical. The real issue here is that we always pass 1 as the `redistributeCount`, even for the last unit, simultaneously passing 0 as the `maxCount`, which causes an assertion failure in the `ValueSelectionDialogElement`'s constructor, which (quite rightly) does not expect the current value to be greater than the maximum (apparently, this assertion simply did not exist before). All we have to do here is pass 0 instead of 1 as the `redistributeCount` value if we are trying to "redistribute" the last stack consisting of 1 unit. That's how this works now:

https://github.com/user-attachments/assets/34b3e876-e2f1-4dda-a659-6e10799a9dcf

What do you think about this approach?

P.S. I have also rephrased some comments and added a check for the case of "splitting" a slot with just one monster into a slot with just one monster of the same type.

filipni:
Thanks for the clarification @oleg-derevenetz. I should've spent some time getting familiar with this dialog before attempting a fix. Your approach seems reasonable and correct.

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@filipni , thank you very much for your contribution and @oleg-derevenetz for your help solving this issue!

PR #9116, opened 2024-09-21
Rework the text rendering code

Districh-ru:
This PR contains part of the code made for #8854 
It removes the duplicated code for text line properties calculation.
Now the renderer uses the same function (`getTextLineInfos()`) as all other `Text` class methods do. So now `width()`, `height()`, and the renderer uses the same text properties calculation algorithm.

The calculated text data currently is not stored and is recalculated for every Text class method like in the master build.
To do this calculation once the constructor and `set()` method should be changed including the change of all the code where the `Text` class is used.
This change will be done in the other PR after this PR is merged.


src/fheroes2/gui/ui_text.cpp
```diff
@@ -105,7 +110,7 @@ namespace
             if ( isSpaceChar( *data ) ) {
                 spaceWidth += spaceCharWidth;
             }
-            else {
+            else if ( !isLineSeparator( *data ) ) {
```
ihhub, line 113: Could you please explain this change?

ihhub (commented):
Hi @Districh-ru , I left one comment here. Could you please take a look?

ihhub:
Hi @Branikolog , please test multiple scenarios with text rendering including extreme cases and different languages.

src/fheroes2/gui/ui_text.cpp
Districh-ru: Hi, @ihhub,
the `getTruncatedLineWidth()` function is used for single-line texts rendering and in `fitToOneRow()` function. And in original texts there are no `\n` symbols in such lines. But in translations it might be specially used for some long texts: e.g. spell name is rendered for as multi-line in the spell book and as a single-line.
In text renderer `\n` is not rendered and has zero sized image (see the logic of `getTextLineInfos()`, line 153-163). :)
But here in master build it was checked by `.getWidth()` and this invalid character is replaced by `?` in this function so its width is not zero there.
So to correspond the renderer behavior I made this change.
Originally it was made about 4 months ago and, sorry, I don't remember what phrase in what translation was displayed incorrectly, but this change was made to fix such rare bug.

Branikolog:
Hi, @Districh-ru and @ihhub !
The app crashes, when I'm trying to enter save/load file list.

Districh-ru:
> Hi, @Districh-ru and @ihhub !
> The app crashes, when I'm trying to enter save/load file list.

 Hi, @Branikolog, if it is an assertion fail could you please provide a screenshot of the assettion failure dialog.
Sadly, I cannot reproduce this crash.

Branikolog:
@Districh-ru 

> > Hi, @Districh-ru and @ihhub !
> > The app crashes, when I'm trying to enter save/load file list.
> 
> Hi, @Branikolog, if it is an assertion fail could you please provide a screenshot of the assettion failure dialog. Sadly, I cannot reproduce this crash.

There's no assertion. Application is just disappearing without any message.
I've checked, that crash does not occur without any save file in the folder or even if I make several new files.

I have dozens of save files and I suppose some of them cause the problem. Maybe the number of my files itself could be the reason app crashes, how do you think?

I can attach the whole folder of my save files if you wish.

Branikolog (approved):
The issue with save list is gone and couldn't find any other issues with texts.

ihhub (approved):

ihhub:

PR #9118, opened 2024-09-21
Battle log: split long phrases into two lines

### LINKED ISSUE #6357: Long phrases in the battle log

Districh-ru:
Fix #6357 by splitting long phrases into two lines:
изображение

For more tests I temporarily reduced the allowed text width (not in this PR, only on my PC - only for test):
изображение

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -743,8 +743,28 @@ namespace Battle
                 // Find the maximum allowed phrase to fit the line. We assume that all words are separated by a space.
                 pos = str.rfind( ' ', pos - 1 );
 
-                // There should not be such a large word.
-                assert( pos != std::string::npos );
+                // Theoretically there could be such a large word that did not fit the log width.
+                if ( pos == std::string::npos ) {
+                    // Find the next word in this phrase.
+                    pos = str.find( ' ' );
+
+                    // Cut this long word with the trailing '...'.
+                    text.fitToOneRow( battleLogElementWidth );
+
+                    if ( pos == std::string::npos ) {
+                        // This is the only word in the phrase.
+                        str = text.text();
+                    }
+                    else {
+                        _messages.push_back( text.text() );
+                        // Put the rest of the phrase to the next line.
+                        str = str.substr( pos + 1 );
+                        // There is no need to split this phrase any more.
+                        pos = std::string::npos;
+                    }
+
+                    break;
+                }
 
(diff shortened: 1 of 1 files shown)


Branikolog (approved):

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -735,28 +735,47 @@ namespace Battle
...
+            // Check if message string fits the log width.
+            fheroes2::Text text( str, fheroes2::FontType::normalWhite() );
+            size_t pos = std::string::npos;
+
+            while ( text.width() > battleLogElementWidth ) {
+                // Find the maximum allowed phrase to fit the line. We assume that all words are separated by a space.
+                pos = str.rfind( ' ', pos - 1 );
+
+                // There should not be such a large word.
+                assert( pos != std::string::npos );
```
ihhub, line 747: I am not sure that this is always true. We could simply break from this loop.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -735,28 +735,47 @@ namespace Battle
...
+
+                text.set( str.substr( 0, pos ), fheroes2::FontType::normalWhite() );
+            }
+
+            if ( pos == std::string::npos ) {
+                _messages.push_back( std::move( str ) );
```
ihhub, line 753: How about the case when indeed a word can be longer than area width? Should we use `fitToOneRow()` and `text()` methods here to get the proper string?

ihhub (changes requested):
Hi @Districh-ru , I left 2 questions here. Could you please take a look at them?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Hi, @ihhub !
Theoretically there might be a very long word: https://en.wikipedia.org/wiki/Longest_word_in_English
I've added an easy logic for a such case.

In original game there are no such words and when adding new phrases and translating them we should avoid using extra-large words that do not fit even the battle log. :))

ihhub (approved):

ihhub:

PR #9119, opened 2024-09-21
getHeroAnimation() function minor code clean-up

Districh-ru:
Initialize vectors with the known constants instead of using `.push_back()` in a loop from a const array (and then never used) to a vector.


ihhub (approved):

ihhub:
@Districh-ru , thank you so much for this code cleanup!

PR #9124, opened 2024-09-22
Set UID counter to the maximum object's UID value after loading a saved game

### LINKED ISSUE #9123: Crash during AI turn due to unknown object
### Preliminary checks

### Platform

Windows

### Describe the bug

Reported on discord. Using public 1.1.2 version 

Load the save file and end the turn. An assertion will trigger.

Schermafbeelding_2024-09-21_225841

### Save file

[584f9408-1634-4945-94dd-c1c26d959b70.zip](https://github.com/user-attachments/files/17088790/584f9408-1634-4945-94dd-c1c26d959b70.zip)

### Additional info

_No response_

Districh-ru:
Fix #9123

When loading a saved game the `objectCounter` is left as it was before the game load. Oh fheroes2 start it is set to 0. So after loading a saved game from the main menu the `objectCounter` starts from 0 and it leads to duplication of map objects' UIDs.

This PR adds a loop though all loaded tiles to get the maximum UID value and set `objectCounter` equal to the found value.

**This PR fixes the cause of the #9123 issue but not the result of setting the incorrect UID that is present in the attached save file.** The attached save file will contain the explained bug which happens when `removeObjectFromMapByUID()` is called. In that save two boats with the same ID are positioned in the nearby tiles and are considered as one 2-tile object by this function.

### DIFF
--- a/src/fheroes2/world/world.cpp
+++ b/src/fheroes2/world/world.cpp
@@ -1304,7 +1304,7 @@ void World::updatePassabilities()
     }
 }
 
-void World::PostLoad( const bool setTilePassabilities )
+void World::PostLoad( const bool setTilePassabilities, const bool updateUidCounterToMaximum )
 {
     if ( setTilePassabilities ) {
         updatePassabilities();
@@ -1326,6 +1326,26 @@ void World::PostLoad( const bool setTilePassabilities )
 
     resetPathfinder();
     ComputeStaticAnalysis();
+
+    if ( updateUidCounterToMaximum ) {
+        // Find the maximum UID value.
+        uint32_t maxUid = 0;
+
+        for ( const Maps::Tiles & tile : vec_tiles ) {
+            maxUid = std::max( tile.getMainObjectPart()._uid, maxUid );
+
+            for ( const auto & addon : tile.getBottomLayerAddons() ) {
+                maxUid = std::max( addon._uid, maxUid );
+            }
+
+            for ( const auto & addon : tile.getTopLayerAddons() ) {
+                maxUid = std::max( addon._uid, maxUid );
+            }
+        }
+
+        // And set the UID counter value with the found maximum.
+        Maps::setLastObjectUID( maxUid );
+    }
 }
 
 uint32_t World::GetMapSeed() const
@@ -1472,7 +1492,7 @@ IStreamBase & operator>>( IStreamBase & stream, World & w )
 
     stream >> w.map_objects >> w._seed;
 
-    w.PostLoad( false );
+    w.PostLoad( false, true );
 
     return stream;
 }
--- a/src/fheroes2/world/world.cpp
+++ b/src/fheroes2/world/world.cpp
@@ -1327,24 +1327,28 @@ void World::PostLoad( const bool setTilePassabilities, const bool updateUidCount
     resetPathfinder();
     ComputeStaticAnalysis();
 
-    if ( updateUidCounterToMaximum ) {
-        // Find the maximum UID value.
-        uint32_t maxUid = 0;
+    // Find the maximum UID value.
+    uint32_t maxUid = 0;
 
-        for ( const Maps::Tiles & tile : vec_tiles ) {
-            maxUid = std::max( tile.getMainObjectPart()._uid, maxUid );
+    for ( const Maps::Tiles & tile : vec_tiles ) {
+        maxUid = std::max( tile.getMainObjectPart()._uid, maxUid );
 
-            for ( const auto & addon : tile.getBottomLayerAddons() ) {
-                maxUid = std::max( addon._uid, maxUid );
-            }
-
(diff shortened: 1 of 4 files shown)


src/fheroes2/world/world.cpp
```diff
@@ -1326,6 +1326,26 @@ void World::PostLoad( const bool setTilePassabilities )
 
     resetPathfinder();
     ComputeStaticAnalysis();
+
+    if ( updateUidCounterToMaximum ) {
```
oleg-derevenetz, line 1330: Hi @Districh-ru what do you think about doing this always, but initializing the `maxUid` by `Maps::getLastObjectUID()` result instead of zero? Yes, this may be an unnecessary action in certain cases, but it will not be possible to accidentally forget about it by specifying the wrong flag.

Also, since we set the `maxUid` to zero here, what if the current game-wide last object UID value is already larger than the last UID value used by the savefile? Won't this lead to any unexpected consequences?

Districh-ru: > Hi @Districh-ru what do you think about doing this always, but initializing the `maxUid` by `Maps::getLastObjectUID()` result instead of zero? Yes, this may be an unnecessary action in certain cases, but it will not be possible to accidentally forget about it by specifying the wrong flag.

Hi, @oleg-derevenetz, the `Maps::getLastObjectUID()` will return the `objectCounter` which is `0` if fheroes2 was just loaded or it will give the value from the previously played game which might be not the same as a loaded one. In example, you played a big map with 20000 UIDs and then loaded a save from a small map with 3000 UIDs - then you will jump to 20000 which is much more than the maximum UID for the loaded map. It is not critical and all will be OK, but IMHO it is not good - lets continue UIDs from the currently loaded save.

> Also, since we set the `maxUid` to zero here, what if the current game-wide last object UID value is already larger than the last UID value used by the savefile? Won't this lead to any unexpected consequences?

After we load a game all tiles are fully updated with the data from the save file and we don't need to care about what objects were in tiles before. But we should avoid getting from `objectCounter` values that are already used for objects' UIDs in the loaded save.
Previously we got this unexpected consequences because `objectCounter` returned values that were used on the map and we got more than one object with the same UID.

Funny that this issue did not happen for a long time. :)

A better solution for UID generation is to start `objectCounter` from `0` and in `getNewObjectUID()` analyze all tiles and return the next unused UID. But its more expensive and we have whole uint32 for `objectCounter` so we can simply start generating `objectCounter` from the maximum UID value used in the current game.

And yes, we can `updateUidCounterToMaximum` not only on save load but also when starting a new game from map file. It might help us to deal with hacked original maps that may have hacked UIDs (if such exists). Doing it will not harm good maps. What do you think @ihhub?

oleg-derevenetz: > And yes, we can updateUidCounterToMaximum not only on save load but also when starting a new game from map file. It might help us to deal with hacked original maps that may have hacked UIDs (if such exists). Doing it will not harm good maps.

I'm not very familiar with map internals, hence this question. Are these UIDs already assigned to objects in MP2/MX2 maps? If so, there should already be a code that will update the latest in-game UID upon such maps start, to let the spells like "Summon Boat" work, right?

Districh-ru: Yes, these UIDs are assigned to objects in MP2/MX2 by Editor and the maximum UID value is stored in the map file in last 4 bytes. And it is all OK when you start a map and play it, this value is set to the `objectCounter`. But `objectCounter` value is not stored in a save file so after loading a game the proper `objectCounter` value is not restored.
In our map format we do not store this and UIDs are set for every object beginning from 0 on map load using the `getNewObjectUID()` function.

ihhub: Hi @Districh-ru , I am fine to add an extra check. After all we are talking about few hundreds of operations which is nothing on what we do during any map or save file load.

Districh-ru: Funny, but the official maps may also have incorrect maxUID stored in last 4 bytes: map `DEATHG.MP2` has a value the is higher than the maxUID by 1 - it is OK, but it is probably the original Editor bug.
I made an assertion that initial counter value should be not less than the maximum UID on the map.

oleg-derevenetz: > Funny, but the official maps may also have incorrect maxUID stored in last 4 bytes: map DEATHG.MP2 has a value the is higher than the maxUID by 1 - it is OK, but it is probably the original Editor bug.

I suppose that the last added object simply was removed later by the map author for some reason, that's why stored UID is one more than `maxUid`. This can hardly be considered a bug :)

ihhub: Having the counter higher than the last object UID is okay while having it smaller is a problem

ihhub (approved):

ihhub:

PR #9125, opened 2024-09-22
Add language support for fh2m map format

ihhub:
This pull request contains major changes in multiple places in the code to add language support for maps:
- change Virtual Keyboard to support languages that are not currently set as active.
- modify language selection dialog to get selected language within the dialog.
- modify TextBase class to support languages. It is not a perfect solution and most importantly it is not performant enough as usage of `fheroes2::LanguageSwitcher` class has to be optimized.

Due to the complexity of changes to support translations for hero and castle names it was decided not to support translations for them as of now. Castle and hero names are used in many places, not only in their perspective dialogs so changing them require a lot of code re-structing which I would like to avoid as of now.

Places that support translation:
- map name
- map description
- daily and adventure rumors
- events
- sphinx

Unsolved problems for the future:
- performance of `fheroes2::LanguageSwitcher` class. At the moment the class always reloads resources.
- `Dialog::Input()` function accepts a language and changes its UI accordingly while it should maintain the same language for UI as the game's language.
- `Heroes::OpenDialog()` method must be rewritten as it's a mix of everything.
- as mentioned above hero and castle names should support translation.

relates to #8589


src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -669,9 +678,24 @@ IStreamBase & Maps::operator>>( IStreamBase & stream, FileInfo & fi )
...
+    if ( Game::GetVersionOfCurrentSaveFile() < FORMAT_VERSION_1103_RELEASE ) {
+        fi.mainLanguage = fheroes2::SupportedLanguage::English;
+    }
+    else {
+        using LanguageUnderlyingType = std::underlying_type_t<decltype( fi.mainLanguage )>;
+        static_assert( std::is_same_v<LanguageUnderlyingType, uint8_t>, "Type of language has been changed, check the logic below" );
+        LanguageUnderlyingType language;
+
+        stream >> language;
+        fi.mainLanguage = static_cast<fheroes2::SupportedLanguage>( language );
```
oleg-derevenetz: Hi @ihhub all this should no longer be necessary, now there are templates of `operator>>()` and `operator<<()` for enums (including enum classes), so you can just do:

```suggestion
        stream >> fi.mainLanguage;
```

ihhub: Hi @oleg-derevenetz , I've just fixed it. Thank you for spotting!

src/fheroes2/gui/ui_text.cpp
```diff
@@ -321,6 +321,15 @@ namespace
...
         return std::max( maxWidth, width );
     }
+
+    std::unique_ptr<fheroes2::LanguageSwitcher> getLanguageSwitcher( const fheroes2::TextBase & text )
+    {
+        if ( !text.getLanguage() ) {
+            return {};
+        }
+
+        return std::make_unique<fheroes2::LanguageSwitcher>( text.getLanguage().value() );
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub what Clang-Tidy is trying to say here (as I understand it) is that it is not sure that the first and the second calls of `text.getLanguage()` will yield the same result. I believe that a temporary variable will save the day.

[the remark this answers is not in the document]
ihhub: Hi @oleg-derevenetz , thanks for pointing out. I've changed this place.

src/fheroes2/maps/maps_fileinfo.h
```diff
@@ -144,6 +146,16 @@ namespace Maps
...
         static bool sortByMapName( const FileInfo & lhs, const FileInfo & rhs );
 
+        // Only Resurrection Maps contain supported language.
+        std::optional<fheroes2::SupportedLanguage> getSupportedLanguage() const
+        {
+            if ( version == GameVersion::RESURRECTION ) {
+                return mainLanguage;
+            }
+
+            return std::nullopt;
```
oleg-derevenetz, line 156: `std::nullopt` is perfectly fine here and in other similar places, but just `{}` is just shorter and means effectively the same. `std::nullopt` is intended mainly for situations where type inference is needed and there is no way to infer the actual type of `{}`. But, as I already said, `std::nullopt` is OK and if you want to use it for clarity, then no problem.

oleg-derevenetz (approved):
Hi @ihhub I just left a single style-related optional comment, LGTM.

Branikolog:
Hello, @ihhub !
I've started testing your PR and found some issue. Could you, please, look of this behaviour ok?

After switching languages Russian becomes partly unreadable:

https://github.com/user-attachments/assets/1f50532b-ab09-4211-9fdd-9fa431046541

Also, I'm concerned about some titles shown according to the selected map language, without considering interface language:

https://github.com/user-attachments/assets/a1a70658-fa6c-4a6a-acb5-bf926b334a75

src/fheroes2/maps/maps_fileinfo.h
ihhub: It is changed.

ihhub:
Hi @Branikolog , I've addressed your comments. Please review it again.

ihhub:
Hi @Branikolog and @zenseii , I realized that my solution for dialog' titles leads to showing ???. I will fix it soon.

ihhub:
Hi @Branikolog and @zenseii , this pull request is ready for your re-review.

Branikolog (approved):

PR #9126, opened 2024-09-22
Battlefield: optimize Damage Info pop-up dialog rendering

Districh-ru:
This PR optimizes the Damage Info pop-up dialog rendering and fixes some minor bugs:
- render dialog once to image (when it has to pop-up) and then just copy this image to `display` on every `redraw()` call;
- do not reset `Game::BATTLE_POPUP_DELAY` on every `HumanTurn` loop even if this dialog is turned off;
- properly show dialog near the two-cell wide units and do not update it if the cursor is moved to the next cell of the same unit;
- hide dialog when the cursor is moved out of the cells area (when the unit is at the edge of the battlefield).

Master build:

https://github.com/user-attachments/assets/e6a881d1-8c36-49d5-a6b1-cabbb7dadc26

This PR:

https://github.com/user-attachments/assets/98209ae5-07aa-487a-8482-385396cb529c

### DIFF
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -248,26 +248,33 @@ namespace Battle
     class PopupDamageInfo : public Dialog::FrameBorder
     {
     public:
-        PopupDamageInfo();
+        explicit PopupDamageInfo( fheroes2::Rect * battleUIRect )
+            : Dialog::FrameBorder( 5 )
+            , _battleUIRect( battleUIRect )
+        {
+            // Do nothing.
+        }
+
         PopupDamageInfo( const PopupDamageInfo & ) = delete;
 
         PopupDamageInfo & operator=( const PopupDamageInfo & ) = delete;
 
-        void setBattleUIRect( const fheroes2::Rect & battleUIRect );
-        void SetAttackInfo( const Cell * cell, const Unit * attacker, const Unit * defender );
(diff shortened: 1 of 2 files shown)


Branikolog (approved):
👍

src/fheroes2/battle/battle_interface.h
```diff
@@ -248,26 +248,33 @@ namespace Battle
     class PopupDamageInfo : public Dialog::FrameBorder
     {
     public:
-        PopupDamageInfo();
+        explicit PopupDamageInfo( fheroes2::Rect * battleUIRect )
```
ihhub, line 251: Could you please explain why we switch from storing a fheroes2::Rect value into a pointer? Working with pointers is way more dangerous. The size of a pointer is 8 bytes while the size of fheroes2::Rect is 16 bytes.

ihhub (commented):
Hi @Districh-ru , I left one comment here. Would you mind to take a look at it?

src/fheroes2/battle/battle_interface.h
Districh-ru: Hi, @ihhub , I did this change jus because I did not want to have a copy of data and tried to make this reference more safe and initialize it in constructor with another member of `Battle::Interface`.
But it is OK to store `fheroes2::Rect`, lets be more safe in here. :)
I've reverted this change.

ihhub (approved):

ihhub:

PR #9135, opened 2024-09-24
Fix a few code smells

oleg-derevenetz:
This PR mostly fixes the code smells related to the cases when non-final virtual methods are called from the constructor, as well as a few other minor code smells.


ihhub (approved):

ihhub:

PR #9144, opened 2024-09-25
Do not take into account top layer when selecting objects to erase.

### LINKED ISSUE #9132: The town disappears along with the erased monster next to it
### Preliminary checks

### Platform

Windows

### Describe the bug

If we try to erase any monster, which is placed above town - they both disappear.

https://github.com/user-attachments/assets/d095ba36-b155-42d7-b484-0e835a32aca4

### Save file

None.

### Additional info

_No response_

Districh-ru:
Fix #9132

In original Editor the Eraser does not erase objects by selecting only their top layer parts.
In the game there could be very small or even empty top layer parts like the upper tiles of Towns.
So not to confuse map maker and to correspond the original Editor behavior this PR skips checking the Top layer when getting objects to remove in the highlighted area.

### DIFF
--- a/src/fheroes2/maps/maps_tiles_helper.cpp
+++ b/src/fheroes2/maps/maps_tiles_helper.cpp
@@ -3296,12 +3296,9 @@ namespace Maps
                     }
                 }
 
-                for ( const auto & addon : currentTile.getTopLayerAddons() ) {
-                    // Top layer addons don't have layer type.
-                    if ( addon._uid != 0 ) {
-                        objectsUids.insert( addon._uid );
-                    }
-                }
+                // The top layer objects are not taken into account to correspond the original editor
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @Districh-ru , this change is a bit controversial. What's about shadow layer then?

Districh-ru:
> Hi @Districh-ru , this change is a bit controversial. What's about shadow layer then?

Hi, @ihhub, shadow layer is already ignored in this function:
https://github.com/ihhub/fheroes2/blob/d24070c4f16270bcbed4a06b7cf35dcfc0334cfc/src/fheroes2/maps/maps_tiles_helper.cpp#L3289-L3297

ihhub (approved):

ihhub:
@Districh-ru , many thanks for this change!

PR #9150, opened 2024-09-26
Some code optimizations

Districh-ru:
Use const reference to Point, Size and Rect in function arguments, use array instead of vector


ihhub:
@Districh-ru , thank you so much for this code cleanup!

PR #9151, opened 2024-09-27
Improve the I/O error handling

oleg-derevenetz:
The idea here is to allow external `operator<<()` and `operator>>()` to set (but not clear!) the `FAILURE` flag for a stream to signal errors during serialization (and especially deserialization) of complex objects.

PR #9153, opened 2024-09-28
Migrate AllCastles and AllHeroes to std::unique_ptr

oleg-derevenetz:
Partially based on #9151 and should be merged after it.


ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for this change!

PR #9156, opened 2024-10-01
Do not mark captured castle objects by Color::UNUSED

oleg-derevenetz:
In the `master` branch, when some player becomes vanquished, all his objects become neutral and get the `Color::NONE`, except towns or castles, which get the `Color::UNKNOWN` (you may see it in the "capture color:" section of the debug info printed to the console when you press the RMB on the corresponding object). The reason is that `Color::NONE` removes flags on objects completely, while towns and castles still should have "neutral" flags on both sides of the gate. This PR makes it so that "newly-made-neutral" towns & castles get the `Color::NONE` (just like the "neutral-from-the-very-beginning" ones), but flags remain in place.

Castle flags are displayed correctly:

https://github.com/user-attachments/assets/674cd63d-5e8b-4503-ac72-6846bc12d687

Also this PR removes the need to use `Maps::Tiles::removeOwnershipFlag()` on mines when using the Haunt spell on them, and removes this method completely.


ihhub:
Hi @oleg-derevenetz , is this the same as in the original game? I think the OG doesn't show flags for this case.

oleg-derevenetz:
Hi @ihhub

> is this the same as in the original game? I think the OG doesn't show flags for this case.

I don't know how to check this in the original game, since, if a player is vanquished and the game still continues, then this means that his castles belong to some other player :) I was forced to patch the fheroes2 sources a bit to at least verify that flags work at least like in the `master` branch.

ihhub (approved):

ihhub:
@Branikolog , can you please review this pull request?

Branikolog (approved):

ihhub:

PR #9159, opened 2024-10-02
Clean-up battle interface Status class

Districh-ru:
This PR unifies setting the status bar massage on the Battlefield in the most cases (where there is no need to set the top message without removing the bottom one).
It also updated this class methods' names to correspond the code style.


ihhub (approved):

ihhub:
@Districh-ru , thank you very much for this cleanup!

PR #9160, opened 2024-10-02
Clean battle interface TurnOrder class

Districh-ru:
This PR cleans the `TurnOrder` class.
In this PR the calculation of unit icons positions is done only when the units count is changed and on every redraw call.
The pointers to the units are still updated on every redraw call since they could be changed and currently there is no way to easy detect such change.


src/fheroes2/battle/battle_interface.cpp
```diff
@@ -1091,42 +1080,42 @@ void Battle::TurnOrder::QueueEventProcessing( std::string & msg, const fheroes2:
...
-    number.draw( pos.x + 2, pos.y + 4, output );
+    number.drawInRoi( pos.x + 2, pos.y + 4, output, pos );
 
     if ( isCurrentUnit ) {
-        fheroes2::DrawRect( output, { pos.x, pos.y, turnOrderMonsterIconSize, turnOrderMonsterIconSize }, currentUnitColor );
+        fheroes2::DrawRect( output, pos, currentUnitColor );
     }
     else {
-        uint8_t color = 0;
+        uint8_t color = ARMY_COLOR_BLUE;
```
ihhub, line 1112: Let's make the default color as **ARMY_COLOR_GRAY** which is a neutral army.

ihhub (commented):
Hi @Districh-ru , I left one suggestion here. Could you please check it?

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Sure!

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for this cleanup!

PR #9166, opened 2024-10-04
Small code clean in Castle class

Districh-ru:
This PR updates `Castle` class' private methods naming, makes methods' arguments `const` where needed and makes some methods inline (code moved from castle.cpp to castle.h).


ihhub (approved):

ihhub:
@Districh-ru , many thanks for the cleanup!

PR #9171, opened 2024-10-05
Add an additional assertion

oleg-derevenetz:
There should be no valid reason for `World::getCastleEntrance()` to return `nullptr` for a tile with `OBJ_CASTLE`.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for keeping the code in a better state!

PR #9173, opened 2024-10-06
Migrate the build from macOS 12 to macOS 13

oleg-derevenetz:
macOS 12 is now [deprecated](https://github.com/actions/runner-images/issues/10721) with all the usual consequences - less runners, intermittent failures and so on.


ihhub (approved):

ihhub:

PR #9176, opened 2024-10-07
Fix the order in which events are handled by various UI elements in the "no interface" mode

oleg-derevenetz:
Fix the regression introduced by #8408.

https://github.com/user-attachments/assets/5c9aed12-27e0-4170-896e-3761687a1338

With this PR:

https://github.com/user-attachments/assets/99698337-d834-46ef-be1d-60e6713c0010

### DIFF
--- a/src/fheroes2/game/game_startgame.cpp
+++ b/src/fheroes2/game/game_startgame.cpp
@@ -1227,7 +1227,6 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
             _gameArea.ResetCursorPosition();
         }
 
-        // gamearea
         if ( !_gameArea.NeedScroll() && !isMovingHero ) {
             if ( !_radar.isDragRadar() ) {
                 _gameArea.QueueEventProcessing( isCursorOverGamearea );
@@ -1245,7 +1244,7 @@ fheroes2::GameMode Interface::AdventureMap::HumanTurn( const bool isload )
             break;
         }
 
-        // animation of the hero's movement
+        // Animation of the hero's movement
         if ( Game::validateAnimationDelay( Game::CURRENT_HERO_DELAY ) ) {
             Heroes * hero = GetFocusHeroes();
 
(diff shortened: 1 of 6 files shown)


Districh-ru (commented):
Hi, @oleg-derevenetz, I found the next two issues:
1. If you press the left mouse button over the Icons Panel and move the cursor to the game area keeping the button pressed you will start dragging the game area:

https://github.com/user-attachments/assets/f2af4f48-d210-41c7-924e-9aa8ce1c2e73


2. The next issue is the rendering order related and is not directly connected to the Events PR and could be fixed in the other PR.
When the buttons are overlapped by the other window if you press the button it will be rendered over the overlapped window for a short time:

https://github.com/user-attachments/assets/a5beb04a-ca5a-4c0b-881e-f1691ffabd58

PS
I'll be on a business trip for about a week and probably I'll not be able to review this PR for this time.

oleg-derevenetz:
Hi @Districh-ru

> If you press the left mouse button over the Icons Panel and move the cursor to the game area keeping the button pressed you will start dragging the game area

This is not the Icons Panel (`iconsPanel`), this is `buttonsArea`, and this behavior is observed not just with this PR, but in the `master` branch as well. This PR doesn't alter this behavior and doesn't touch the `buttonsArea` at all. In your own video, when you press the left mouse button over the **real** Icons Panel (that one with the list of heroes & castles), no dragging of the game area happens. `iconsPanel` is designed to be scrollable by dragging, that's why capturing the mouse by it is important (and it works properly according to your video). `buttonsArea` is intended for a single-click or long-press events and does not capture the mouse. If this is an issue, then it does not apply to this PR. I suggest solving it separately at another time and in another PR.

Districh-ru:
Sorry, I confused Buttons panel with Icons panel.
And, yes, it is an issue because in "Interface mode" game area scrolling does not start in this case.
This and the second mentioned issue can be addressed in a new PR

Districh-ru (approved):

ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for this fix!

PR #9179, opened 2024-10-10
Fix map titles not being cleared when changing map

### LINKED ISSUE #9177: Map selection dialog, text leftowers for long map names
### Preliminary checks

### Platform

Windows

### Describe the bug

Some parts of the edge symbols are visible after switching maps. 
image

https://github.com/user-attachments/assets/46c7c686-dc59-4839-997e-2e1dd04f5a8f

### Save file

None.

### Additional info

_No response_

zenseii:
Fix #9177 

The total area being cleared/restored is now increased ~~because the new map editor allows for longer map names than the original.~~ <-This is wrong: 17 characters was the max amount for original maps and it's the same for fheroes2.

### DIFF
--- a/src/fheroes2/game/game_scenarioinfo.cpp
+++ b/src/fheroes2/game/game_scenarioinfo.cpp
@@ -245,7 +245,7 @@ namespace
         }
 
         // Set up restorers.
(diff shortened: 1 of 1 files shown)


ihhub (approved):

Branikolog (approved):

ihhub:
@zenseii , a bunch of thanks for this fix!

Issue #9094, opened 2024-09-08
Combat screen, Auto combat and Instant Combat windows, change the text for better understanding


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

When the user presses the "Q" key on the keyboard (to instantly resolve the combat and exit the combat screen), a window is displayed to allow him to validate his action or not.
However, the text is not clear.

### Describe the possible solution

In fheroes2:

![Auto and instant combat modes 001](https://github.com/user-attachments/assets/21a4b6db-6099-491a-bb10-666ad0910b3a)

I suggest displaying the following text:

~"Are you sure you want to resolve the combat in instant combat mode?"~
"Are you sure you want to end the combat in instant combat mode?"

Instead of:

"Are you sure you want to finish the battle in auto mode?"

I also suggest changing the text of the window displayed when the user presses the “AUTO” button to continue the combat in automatic mode.

In fheroes2:

![Auto and instant combat modes 002](https://github.com/user-attachments/assets/f28d4cc1-2d84-45c2-a729-2d4f3439241e)

I suggest displaying the following text:

"Are you sure you want to enable the auto combat mode?"

Instead of:

"Are you sure you want to enable auto combat?"

And to change the text of the window displayed when the user presses the “AUTO” button to interrupt the combat in automatic mode.

In fheroes2:

![Auto and instant combat modes 002a](https://github.com/user-attachments/assets/82d5e56d-c133-4589-9b69-ee88c6e5f775)

I suggest displaying the following text:

"Are you sure you want to interrupt the auto combat mode?"

Instead of:

"Are you sure you want to interrupt the auto combat?"

oleg-derevenetz:
I have to explain the difficulty with this proposal (and with #9095). We already have the "finish the battle in auto mode" hotkey definition in the `fheroes2.key`, and if the wording in the game and in the hotkey config file will differ, this may confuse the user. At the same time, the configuration files must remain backward compatible, so we can't just change the wording back and forth without further thought.

LeHerosInconnu:
I have opened a related discussion to this: https://github.com/ihhub/fheroes2/discussions/9100.

LeHerosInconnu:
Hello everyone,

[...]

Issue #9095, opened 2024-09-08
Adventure map, System Options window, change to texts for instant combat settings buttons


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

In fheroes2, the texts of the instant combat settings buttons are not clear.

### Describe the possible solution

First, I suggest changing the text “Battles” by the text “Combat” (without “s” at the end) for the three buttons.

For this one:

![Auto and instant combat modes 004](https://github.com/user-attachments/assets/08077123-8c39-4f3a-b834-3da9799d074a)

No change ("Manual" is okay).

For this other one:

![Auto and instant combat modes 003](https://github.com/user-attachments/assets/aa401fe7-482d-4bce-95ab-a9f7bd720ae5)

Change the text to:

"Instant,
No Spells"

Instead of:

"Auto, No
Spells"

For the last one:

![Auto and instant combat modes 005](https://github.com/user-attachments/assets/ccb58046-c5a5-4fae-979b-b820f7386364)

Change the text to:

"Instant,
With Spells"

Instead of:

"Auto Resolve"

ihhub:
Hi @zenseii , could you please provide your opinion about this?

zenseii:
@ihhub, this should be converted into a discussion due to no one on the team agreeing ho this change in previous discussions.

PR #9180, opened 2024-10-10
Add dialog to choose between Auto Combat and Quick Combat

zenseii:
This will let more users notice the Quick Combat feature and it will give touch devices access to Quick Combat.

This PR also adds basic shine effect to the generated buttons' edges.

image

Related to #9095 and #9100

Fix https://github.com/ihhub/fheroes2/issues/9094


Branikolog:
Hi, @zenseii !

Nice work! This dialog definitely makes gameplay more comfortable.
From interaction point this dialog is absolutely fine!
Here are my few notes on the implementation and the design.

- I suppose, large button is considered as a prototype yet? I feel there are some shiny areas missing on them:
image

- I feel two rows of buttons are rarely used in the game UI. In my personal opinion, the placement of three buttons so close to each other in two rows alongside looks a little out of place here. The large buttons increase this feeling even further. :) I can hardly mention any similar dialog with the same kind of design.

What if we take the other window type and make this dialog look like the existing adventure/file options?
image

I mean we can use that brown marble window type with the same buttons and texts you've implemented.

I noticed one text issue with descriptions but not sure it is relevant though:
Right click on the "AUTO" button states:
image

And the AUTO COMBAT description states:
image

Were the first words chosen to be different on purpose?

Branikolog:
Hi, @zenseii !

I personally love how this dialog is looking now. 👍 

image

Auto combat could be stopped by Esc button too. Should we mention it in the description?

image

Most dialogs have hotkeys. How do you think, if it is neccessary to add a hotkeys here to be able to select both of these features? I know, that both features are able to be triggered outside of this dialog and from my perspective it is just a matter of logic of controls consistency.

Branikolog (approved):
Wonderful!
Absolutely love this enhancement! Thanks!

Districh-ru (approved):
Well done, @zenseii!
Also I like that all engine-made buttons have the upper shine lines now.
изображение

oleg-derevenetz:
Hi @zenseii technically everything looks OK, but I have a naming-related question. Everywhere in the code and in the hotkey config file / settings the "auto resolve" action is called like "auto finish" or "finish the battle in auto mode" (in case of hotkey settings). Is there any way to unify this so that there is no confusion, because I have the impression that an end user may not always understand that this is the same thing? At the same time, it would be undesirable to touch the current name in the hotkey settings, because the previous versions of the engine will not understand these changes "right away".

Districh-ru:
@zenseii 
I vote for `Quick Combat` to be more close to the later HoMM games.

I think that the name of this feature should tell the player that the combat will be instantly finished without any player actions (automatically). And Quick can be understood as automatic. :)
So we will have `Auto Combat` to give control to the AI and watch how it goes (not sow quick), and `Quick Combat` to quickly (almost instantly) do the combat by giving control to the AI without watching the battle.
And ... Quick Resolve sounds for me more like quickly solving equations or other problems. :))

Branikolog:
@zenseii 

>I personally only care that the name we stick with fits, and all of the above do imo.

100%

I like the version @Districh-ru voted for. I think both phrasings shouldn't look idetical. Players should be able to distinguish both features as easy as possible. So differences in the first word in my opinion catch player attention more effective: Auto Combat / Quick Combat. 👍

Branikolog:
@Districh-ru 

> Well done, @zenseii! Also I like that all engine-made buttons have the upper shine lines now. изображение


But still not all of them have those shine lines 😢 :

image

image

image

image

oleg-derevenetz:
As soon as we agree on the name of this thing, I will add a commit to this PR renaming the symbols in the code and in the hotkey config file (in a backward-compatible manner).

LeHerosInconnu:
Hello @zenseii and everyone,

> @LeHerosInconnu, if you have any feedback regarding pixel-precise™ placement of the buttons and elements then that would be appreciated.

The word "CANCEL" is not centered horizontally in the button, 9 pixels to the right, 12 pixels to the left.

Cancel button 96 x 25 pixels 01

Also for the "half pixel" that can't be separated when an uneven number is used, e.g. 25 pixels divided by two gives 12.5, i.e. 12 pixels on one side and 13 pixels on the other, we should set up a convention so that the non-separable pixel is always added on the right side for a horizontal dimension, i.e. in the example 12 pixels on the left and 13 pixels on the right, and is always added at the bottom for a vertical dimension, i.e. in the example, 12 pixels at the top and 13 pixels at the bottom.
I can open a specific issue for this if necessary.

The "AUTO COMBAT" button is narrower than the "AUTO RESOLVE" button, 106 pixels for the former and 108 pixels for the latter.

AUTO COMBAT button 106 x 56 pixels 01
AUTO RESOLVE button 108 x 56 pixels 01

About the rest, here's how I see it.

For this one:
AUTO button info window 01

The window's yellow title should be:

"Auto Combat Modes".

The text should be:

"Choose whether to play the combat in auto combat mode or in instant combat mode."

In the info bar at the bottom of the screen the text should be:

"Auto combat modes"

For this other one:
Auto combat modes choice window 01

The window's yellow title should be:

"Auto Combat Modes".

In the info bar at the bottom of the screen the text should be:

"Auto combat modes"

For this other one:
AUTO COMBAT button info window 01

The window's yellow title should be:

"Auto Combat" (this is already the case).

The text should be:

"The computer plays the combat for you, while you watch the progression of the combat.

This mode can be interrupted at any time by pressing the Auto Combat hotkey or the default Cancel key, or by pressing the right or left mouse button anywhere on the combat screen."

For the last one:
INSTANT COMBAT button info window 01

The window's yellow title should be:

"Instant Combat".

The text should be:

"The combat ends instantly from the current state.

This action cannot be undone once validated."

I continue to think that the expressions “Auto Combat” and “Instant Combat” are more appropriate.
Indeed, both expressions are self-explanatory.
“Auto Combat” represents combat continued automatically, and “Instant Combat” represents combat ended instantly.

I think it's important that the word "combat" is present in both expressions.
It's to combat that the modes refer, and then the user need only remember the two significant words “auto” and “instant”.

oleg-derevenetz:
> Which means that we're only missing @ihhub and @oleg-derevenetz's choices.

I'm fine with "auto combat" and "instant combat".

oleg-derevenetz:
> Did we decide on `Instant Combat`? Me, @Districh-ru and @Branikolog voted `Quick Combat`.

These changes are just a preliminary, because there were many symbols like "auto battle" and similar. I want to check everything once again a day later, like I usually do (to ensure that I didn't forgot or miss something), but I don't want to delay it, so I made these preliminary changes. I will replace the "instant" by "quick" if this option turns out to be the final one :)

Districh-ru:
I prefer the more known **Quick Combat**, but I'm OK with **Instant Combat** - it also has noticeable differences in the first word to catch player attention more effectively and it tells the player that the combat will be performed almost instantly meaning that no combat actions will be shown and no manual interaction with the army can be done until the combat is finished.

oleg-derevenetz:
So, what is the final wording? :)

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2127,6 +2131,28 @@ namespace
 
             break;
         }
+        case ICN::BUTTON_AUTO_COMBAT_GOOD:
+        case ICN::BUTTON_AUTO_COMBAT_EVIL: {
+            _icnVsSprite[id].resize( 2 );
+
+            const bool isEvilInterface = ( id == ICN::BUTTON_AUTO_COMBAT_EVIL );
+
+            getTextAdaptedButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "AUTO\nCOMBAT" ),
```
ihhub, line 2140: Would it be better to include prefix for this and the below translations?
```cpp
"battle|AUTO\nCOMBAT"
"battle|AUTO\nRESOLVE"
```

I assume it might be easier to understand these lines for translators. What do you think?

src/fheroes2/agg/agg_image.cpp
```diff
@@ -2127,6 +2131,28 @@ namespace
 
             break;
         }
+        case ICN::BUTTON_AUTO_COMBAT_GOOD:
```
ihhub, line 2134: This is beyond the scope of this pull request: I think we should stop adding new entries for ICN and just provide a function to generate a button for a specific text and interface. This will ease the work of the future button addition in the work. We should address it in a completely separate pull request. As of now we can just create an issue for the change. What do you think?

ihhub:
Hi, everyone!

While the later games have similar buttons my concerns that having 2 buttons with at least one identical word would still confuse players. Ideally, we need to have 2 buttons with absolutely different names. Having `Auto Combat` and `Instant/Quick Combat` doesn't differentiate too much.

After thinking for some time we can name the button to give the control to AI as `Auto Control` and another button for finishing the battle as `Quick Combat` (`Instant Combat` is also fine with me).

What do you think about this?

ihhub (changes requested):
Hi @zenseii , I also left few comments here. Could you please take a look at them?

oleg-derevenetz:
> After thinking for some time we can name the button to give the control to AI as `Auto Control` and another button for finishing the battle as `Quick Combat` (`Instant Combat` is also fine with me).

> What do you think about this?

Auto Combat is a sort of a well-known/established term in HoMM game series:

auto-homm2

auto-homm3

I actually don't think that fheroes2 should be somehow original in this matter :)

ihhub:
Hi @oleg-derevenetz , in this case let's follow the original game name. With this logic in mind we should use `Quick Combat` for another button.

Branikolog:
After a few days I came to a conclusion, that "Auto" and "Quick" combat is the best for me. "Auto combat", as @oleg-derevenetz stated, is a well established term. "Quick" is simply shorter than "Instant". Both buttons have the first word different which is much more preferable in regard of visual recognition, than second word on the second line.

LeHerosInconnu:
Hello everyone,

@zenseii,

> > [...] we should set up a convention so that the non-separable pixel is always added on the right side for a horizontal dimension
> 
> If you can, then please make an issue for this.

I'll do it. :)

> > The "AUTO COMBAT" button is narrower than the "AUTO RESOLVE" button, 106 pixels for the former and 108 pixels for the latter.
> 
> This will be fixed once we change RESOLVE to COMBAT, since the width of the buttons depend on the width of the text on them to accommodate for translations. I've already thought about making a function that forces buttons in the same dialog to have the same size, but I don't want to implement that here.

Okay.

> > The window's yellow title should be:
> > "Auto Combat Modes".
> 
> To avoid confusion with Auto Combat, how about just Automatic Modes.

I think it's important to specify that this refers to combat.
"Automatic Combat Modes".

> > The text should be:
> > "Choose whether to play the combat in auto combat mode or in instant combat mode."
> 
> My impression is that there was a wish to use the word "fight" instead of "play", which is why it says "fight out". I also think fight sounds better than play, which sounds less serious and makes me more aware of this being a game than a battle with high stakes. How about just "Choose between auto combat mode and quick/instant combat mode."

The words "fight" and "combat" are synonyms, that makes it redundant.

The word "play" is used here more in the sense of playback, like playing back a film, rather than in the sense of playing, having fun.
But I'm not a native English speaker, so I don't know if it's seen more one way than the other.

"Choose between auto combat mode and quick/instant combat mode."
This sentence isn't informative enough about what's going to happen.

What about:
"Choose whether to proceed the combat in auto combat mode or in instant combat mode."

> > In the info bar at the bottom of the screen the text should be:
> > "Auto combat modes"
> 
> Same comment as above.

Yeah. :)
"Automatic combat modes".

> > The window's yellow title should be:
> > "Auto Combat" (this is already the case).
> > The text should be:
> > "The computer plays the combat for you, while you watch the progression of the combat.
> 
> Same comment here about "plays". Should we just replace plays with "fight"?

What about:
"The computer continues the combat for you, while you watch the progression of the combat.

> > This mode can be interrupted at any time by pressing the Auto Combat hotkey or the default Cancel key, or by pressing the right or left mouse button anywhere on the combat screen."
> 
> I specifically avoided the use of "mouse" because this is not true for touch devices, even though they have a simulated right-click. Also, you can right or left click anywhere on the game screen, even when you use higher than original resolutions.

What about:
"This mode can be interrupted at any time by pressing the Auto Combat hotkey or the default Cancel key, or by activating the left or right click anywhere on the game screen."

> > The window's yellow title should be:
> > "Instant Combat".
> 
> > The text should be:
> > "The combat ends instantly from the current state.
> > This action cannot be undone once validated."
> 
> These will depend on the wording we go for. We can change the description text.

"The combat ends instantly from the current state."
"The combat ends quickly from the current state."
My choice is made. :D

> > I continue to think that the expressions “Auto Combat” and “Instant Combat” are more appropriate. Indeed, both expressions are self-explanatory. “Auto Combat” represents combat continued automatically, and “Instant Combat” represents combat ended instantly.
> > I think it's important that the word "combat" is present in both expressions. It's to combat that the modes refer, and then the user need only remember the two significant words “auto” and “instant”.
> 
> I agree about not using battle and rather have combat in this case.

Ah. :)

> > The word "CANCEL" is not centered horizontally in the button, 9 pixels to the right, 12 pixels to the left.
> 
> Thanks for checking. I'll fix this.

> @LeHerosInconnu. I looked at the cancel button and the one you're showing is the original cancel button not the one we generate. However, the change you're requesting will affect our generated buttons too, and should be made an issue/discussion for first.

The "CANCEL" button comes directly from this window.
Auto combat modes choice window 01

But I made a mistake by taking into account the whole width of the button instead of what is considered to be the "top" of the button.
The text is well centered with eight pixels on each side. :)

> I also have a comment about this text for the quick combat feature:
> 
> > The combat ends instantly from the current state.
> 
> Using "instantly" is fine, even with quick combat being the word, but the use of "ends" makes it sound like the screen will be exited as if you're quitting, and this can be confusing. How about something like "The combat will be fought out in an instant from the current state."?

What about:
"The combat is instantly resolved from the current state."

@ihhub,

> While the later games have similar buttons my concerns that having 2 buttons with at least one identical word would still confuse players. Ideally, we need to have 2 buttons with absolutely different names. Having `Auto Combat` and `Instant/Quick Combat` doesn't differentiate too much.
> 
> After thinking for some time we can name the button to give the control to AI as `Auto Control` and another button for finishing the battle as `Quick Combat` (`Instant Combat` is also fine with me).
> 
> What do you think about this?

So, here are a few examples of menus with buttons containing the same word:

"STANDARD GAME"
"CAMPAIGN GAME"
"MULTI-PLAYER GAME"
Same names in menus 001

"STANDARD GAME"
"CAMPAIGN GAME"
"MULTI-PLAYER GAME"
Same names in menus 002

"NEW MAP"
"LOAD MAP"
Same names in menus 003

"NEW GAME"
"LOAD GAME"
"SAVE GAME"
Same names in menus 004

On the contrary, I think that having the same word on several buttons referring to the same thing is easier to apprehend and remember, and is also much clearer.

Consider this:

"NEW SCENARIO"
"LOAD PARTY"
"SAVE GAME"

And what do you think about it?

> Hi @oleg-derevenetz , in this case let's follow the original game name. With this logic in mind we should use `Quick Combat` for another button.

The word "quick" is vague.
The word "quick" can be interpreted as "fast" or "rapid", the user may think that he will see the combat in an accelerated way, etc.
With the word "instant", there's no ambiguity: instant is instantaneous, so there's no need to wonder.

LeHerosInconnu:
Hello @zenseii,

> @LeHerosInconnu,
> 
> > With the word "instant", there's no ambiguity: instant is **instantaneous**, so there's no need to wonder.
> 
> I'm not sure this falls under instantaneous :
> 
>  [fheroes2.engine.version_.1.1.5.2025-01-16.13-57-32.mp4 ](https://github.com/user-attachments/assets/31447534-2f49-4330-8b5b-371984ef18e2)
> 
> However, it is "quicker" than if I'd tried to win the battle. Also don't try this at home, especially not with a maxed stack unless you don't plan to use your PC for a while 😨

It's "quicker" than actually doing the combat, but it's not "quick" either; it just shows that there's still room for improvement in combat resolution in this mode.

You have to look at the meaning of the word itself.

Here's an instant combat:

https://github.com/user-attachments/assets/7b42bbc0-32fe-4fee-8d94-ab738e89b70c

And here's a quick combat:

https://github.com/user-attachments/assets/5a2f728e-f158-4362-b44f-ee2ae174c17d

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -440,8 +440,8 @@ void Game::HotKeysLoad( const std::string & filename )
                             continue;
                         }
                     }
-                    else if ( eventName == std::string_view( "instant combat" ) ) {
-                        value = config.StrParams( "finish the battle in auto mode" );
+                    else if ( eventName == std::string_view( "quick combat" ) ) {
+                        value = config.StrParams( "resolve the battle in quick combat mode" );
```
oleg-derevenetz, line 444: Hi @zenseii this change is wrong. The idea is to look for the old hotkey name (currently it is "finish the battle in auto mode") if the new hotkey name ("quick combat") was not found.

oleg-derevenetz:
> I'm done pushing changes to this branch now. If you have the chance to look over the changes from resolve/instant to quick then that would be greatly appreciated.

Thank you! I'll make an effort to take a look during tomorrow.

src/fheroes2/game/game_hotkeys.cpp
```diff
@@ -432,7 +433,22 @@ void Game::HotKeysLoad( const std::string & filename )
                 const char * eventName = _( hotKeyEventInfo[eventId].name );
                 std::string value = config.StrParams( eventName );
                 if ( value.empty() ) {
-                    continue;
+                    // TODO: remove this temporary workaround
```
ihhub, line 436: How temporary is this workaround? :)

oleg-derevenetz: @ihhub How temporary it should be? :)

LeHerosInconnu:
Hello @zenseii,

> @LeHerosInconnu, I've addressed your text suggestions. I shortened this text because `action` was redundant in this context and `undone` already implies that something has been done, and the word `validate` doesn't work in this context. image

The words "resolved" and "ended" mean the same thing, and the "and ended" part makes the sentence a bit cumbersome.

What about:
"The combat is resolved from the current state to the end."

> As for this description I changed the part about while you observe because it sounds like the game is forcing you to sit down and watch. `activating` was changed with performing since that is the correct expression, alternatively we could just say `right or left clicking`.
> 
> image

The "at the currently set combat speed" part of the sentence will raise a question in the user's mind: "What does that mean?"; whereas the text should be descriptive and clear.

> As for this description I changed the part about while you observe because it sounds like the game is forcing you to sit down and watch.

This indicates that the user will be watching the combat (it's simply a description of what happens in this mode), and that he will be able to observe the AI's actions, which may be beneficial to his understanding of how the AI works.

> alternatively we could just say `right or left clicking`.

Since "by pressing" is used for keys and is not really suitable for clicks (mouse or other, we press a button not a click), there is need to use a more appropriate term for clicks and in the same style as "by pressing", "by performing" is fine.

Why replace the word "time" with "moment"?

What about:
"The computer continues the combat for you, while you watch the progression of the combat.

This can be interrupted at any time by pressing the Auto Combat hotkey or the default Cancel key, or by performing a left or right click anywhere on the game screen."

Branikolog:
>"The combat is resolved from the current state to the end."

Looks nice.
However "The combat ends instantly from the current state." looked good for me too.

>The "at the currently set combat speed" part of the sentence will raise a question in the user's mind: "What does that mean?"; whereas the text should be descriptive and clear.

I agreed. Mentioning speed is useless in the description. 
Mentioning "the progression of the combat" also feels redundant, since we're mentioning in the following description, that the battle could be interrupted at any moment...

What about more short and simple? "The computer continues the combat for you in auto mode/automatically"? 
I'm even OK with "The computer continues the combat for you."

LeHerosInconnu:
@Branikolog and @zenseii,

So, I am fine to go with:

"The combat is resolved from the current state.

This cannot be undone."

and:

"The computer continues the combat for you.

This can be interrupted at any moment by pressing the Auto Combat hotkey or the default Cancel key, or by performing a left or right click anywhere on the game screen."

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3195,35 +3196,114 @@ void Battle::Interface::EventShowOptions()
...
+        if ( le.MouseClickLeft( buttonCancel.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) {
+            return;
+        }
+        if ( le.MouseClickLeft( buttonAutoCombat.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_TOGGLE_AUTO_COMBAT ) ) {
+            EventStartAutoCombat( unit, actions );
+            return;
+        }
+        if ( le.MouseClickLeft( buttonQuickCombat.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::BATTLE_QUICK_COMBAT ) ) {
+            EventQuickCombat( actions );
+            return;
```
oleg-derevenetz, line 3285: I changed the logic to immediately `return` here and in the similar case above instead of checking for `!actions.empty()`, because in theory technically `actions` can already be non-empty when calling this method.

src/fheroes2/dialog/dialog_system_options.cpp
```diff
@@ -184,14 +184,14 @@ namespace
         const Settings & conf = Settings::Get();
         if ( conf.BattleAutoResolve() ) {
             const bool spellcast = conf.BattleAutoSpellcast();
-            std::string value = spellcast ? _( "Auto Resolve" ) : _( "Auto, No Spells" );
+            std::string value = spellcast ? _( "Quick Combat" ) : _( "Quick, No Spells" );
```
oleg-derevenetz, line 187: I wonder if it's worth leaving the old wording here specifically, because technically "Auto Resolve" is better suited here, IMHO - the computer here really tries to conduct a quick combat at first, but this is not the only thing it does - it also offers to conduct the manual battle if you don't like the results.

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: OK, I changed the logic a bit by changing the return type of `EventStartAutoCombat()` and `EventQuickCombat()` methods, now it should work in a more convenient way.

src/fheroes2/dialog/dialog_system_options.cpp
oleg-derevenetz: OK, I reverted this place to the previous wording.

oleg-derevenetz (approved):

src/fheroes2/battle/battle_interface.cpp
oleg-derevenetz: > I also just realized that we need to setup a cursor restorer for this dialog because the cursor in this dialog currently lacks the yellow outline which it has in the confirmation dialogs and the combat settings dialog.

Do as you see fit :)

src/fheroes2/game/game_hotkeys.cpp
ihhub: Hi @oleg-derevenetz , as you mentioned yourself we should be reluctant to changing the hotkeys file. However, such cases might happen and it could be an improvement or literally a bug fix. We could actually introduce versioning through a separate parameter in this file. If users mess this parameters it would be on their shoulders.

We could do this in a separate pull request as @zenseii did a lot of work here and we would like to proceed with the changes.

ihhub (approved):

ihhub:
@zenseii , many thanks for this useful feature!

PR #9191, opened 2024-10-13
Cache translation file loading operations

ihhub:
Since we support multi-language texts (see #9125) we do language switching, which results in heavy operations. We need to optimize the language switching work and this pull request addresses only I/O operations that usually are the slowest.

Partially relates to #6681


oleg-derevenetz (approved):
Hi @ihhub I took the liberty to make a few changes to simplify the logic a bit, please take a look when you have time.

PR #9198, opened 2024-10-14
Fix a few code smells

oleg-derevenetz:
Fix the remaining code smells related to the cases when non-final virtual methods are called from the constructor.

This PR also formats our Python service scripts using `black` and fixes or disables all warnings found by `pylint`.

This PR also paraphrases a few lines from #9191.


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for these fixes!

PR #9200, opened 2024-10-14
Fix debug build crash with French and Italian assets

zenseii:
When using Italian and French assets the deletion call to the `imageArray` vector would go out of bounds. This did not happen in release builds.

The solution was to not use the subtraction operator `-` and instead write the result that gets added to the beginning position parameter of the delete() call. Using parentheses would have worked too.

Additionally this fixes the normal font î being empty in `master`.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -2541,7 +2541,7 @@ namespace
             }
             // The French version replaces several ASCII special characters with language-specific characters.
             // In the engine we use CP1252 for the French translation but we have to preserve the homegrown encoding
-            // of the original so that original French maps's text is displayed correctly.
+            // of the original so that original French maps' texts are displayed correctly.
             if ( crc32 == 0xD9556567 || crc32 == 0x406967B9 ) {
                 // The engine expects that letter indexes correspond to charcode - 0x20, but the original French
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
> When using Italian and French assets the deletion call to the imageArray vector would go out of bounds. This did not happen in release builds.

Yeah, nice catch. Need to be careful with the pointer and iterator arithmetic, because you cannot just add or substract an arbitrary number to/from pointer or iterator (often they are the same thing, at least internally), because if the resulting pointer points outside of the array/vector/... (and this is not the case of the past-the-end object - i.e. `end()`, which is allowed), then the behavior is undefined, even if this happens temporarily. This may also depend on compiler optimizations, e.g. in this hypothetical case:

```cpp
int arr[3] {};

int * ptr = arr + 10 - 9;
```

there can be UB if ptr is calculated as-is, but there will be no UB if compiler applies an optimization and pre-calculates the `10 - 9` part.

oleg-derevenetz:
> I suppose the reason this didn't trigger in release builds is due to compiler optimizations.

Some `std` implementations may also check the resulting value in their implementations of `operator+()` etc for iterators  in debug builds to verify that they are not out-of-bounds, but these checks are removed in release builds.

ihhub (approved):

ihhub:
@zenseii , manty thanks for fixing this issue!

PR #9203, opened 2024-10-15
Added Czech Virtual Keyboard

GeorgeK1ng:
Added Czech Virtual Keyboard

image

P.S. Special letters Ť, Ó, Ň, Ď that are not available on physical keyboard are not included.

### DIFF
--- a/src/fheroes2/gui/ui_keyboard.cpp
+++ b/src/fheroes2/gui/ui_keyboard.cpp
@@ -68,13 +68,13 @@ namespace
     bool isSupportedForLanguageSwitching( const fheroes2::SupportedLanguage language )
     {
         switch ( language ) {
+		case fheroes2::SupportedLanguage::Czech:
         case fheroes2::SupportedLanguage::English:
             // English is a default language so it is not considered as an extra language.
             return false;
         case fheroes2::SupportedLanguage::Polish:
         case fheroes2::SupportedLanguage::Russian:
         case fheroes2::SupportedLanguage::Slovak:
-        case fheroes2::SupportedLanguage::Czech:
         case fheroes2::SupportedLanguage::Ukrainian:
             return true;
         default:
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/ui_keyboard.cpp
```diff
@@ -74,6 +74,7 @@ namespace
         case fheroes2::SupportedLanguage::Polish:
         case fheroes2::SupportedLanguage::Russian:
         case fheroes2::SupportedLanguage::Slovak:
+        case fheroes2::SupportedLanguage::Czech:
```
ihhub, line 77: Please sort languages in an alphabetic order in all places we changed.

ihhub (changes requested):
Hi @GeorgeK1ng , I added one comment here. Can you please take a look at it?

src/fheroes2/gui/ui_keyboard.cpp
GeorgeK1ng: Done

GeorgeK1ng:
@ihhub Fixed as requested

GeorgeK1ng:
> @GeorgeK1ng
> 
> > P.S. Special letters Ť, Ó, Ň, Ď that are not available on physical keyboard are not included.
> 
> Could you please elaborate. To me it seems strange that if these letters are needed to write Czech but we don't make them available. This keyboard is the only way to insert special letters and so this would not be good for map makers or players who want to write Czech. Is this correct?

They are really rare to use. I have no idea where they are used normally. For example I don't rememeber when I used them, in job I write a lot o mails and tickets. 

Of course it's easy to add them, but they are not listed on physical keyboard. You need to use some kind of key kombinations.

I don't think someone will reguest them, they are realy rarely used.

ihhub (approved):

ihhub:
@GeorgeK1ng , many thanks for this update!

GeorgeK1ng:
@ihhub I hope the Czech community will appreciate it :)

PR #9204, opened 2024-10-16
Fix French and German ultimate staff sprite with PoL assets

zenseii:
The staff sprite is corrupted/wrong in at least the German and French PoL assets from GOG. Apparently this bug was introduced with patch 2.0 in those languages. In the Succession Wars asset the correct sprite is still present so we use that instead.

In `master`:
image

In original game:
image


src/fheroes2/agg/agg.h
```diff
@@ -56,7 +56,7 @@ namespace AGG
         std::string _expansionAGGFilePath;
     };
 
-    std::vector<uint8_t> getDataFromAggFile( const std::string & key );
+    std::vector<uint8_t> getDataFromAggFile( const std::string & key, const bool ignoreExpansion );
```
ihhub, line 59: What if we name the new argument as `loadSuccessionWarsResource`? Expansion might be considered by code readers as PoL expansion or as Resurrection expansion.

ihhub (changes requested):
Hi @zenseii , I put 2 comments here. Would you mind to take a look when you have spare time?

ihhub (approved):

ihhub:
@zenseii , huge thanks for this fix!

PR #9206, opened 2024-10-16
ui_button.cpp/.h code clean

Districh-ru:
This PR updates the `ui_button.cpp` and `ui_button.h` code:
- use `unique_prt` instead of `new`/`delete`;
- remove the never set `Display & output = Display::instance()` argument and call `Display::instance()` only when it is needed for render;
- make some methods inline (move from cpp to h);
- ...


src/fheroes2/gui/ui_button.h
```diff
@@ -87,29 +95,29 @@ namespace fheroes2
...
-            _offsetY = offsetY_;
+            _offsetX = offsetX;
+            _offsetY = offsetY;
         }
 
         // Will draw on screen by default
         bool draw( Image & output = Display::instance() ) const;
 
-        // Will draw on screen by default. Returns true in case of state change. This method calls render() internally.
-        bool drawOnPress( Display & output = Display::instance() );
```
ihhub, line 100: Having `Display & output` an input argument was deliberate decision. The only problem I see here is that it should be `Image` instead of `Display`. It was done to support rendering buttons on an image rather than on screen for the future image scaling support for Virtual Keyboard, battle and castle dialogs and etc. I suggest not to simplify things here and in other similar places.

ihhub (changes requested):
Hi @Districh-ru , I left one comment here. Can you please check it?

src/fheroes2/gui/ui_button.h
Districh-ru: Hi, @ihhub,
I've reverted this change.

We can not use `Image` instead of `Display` because we do `output.render();'.
But it looks really odd: we have the only instance oа `Display` and if I'm not wrong there will never be other argument than `Display::instance()`. That's why I removed `Display & output = Display::instance()`.

And as we can see this forced render is not needed (and is harmful) in some places.
Probably we need functions like "drawOnPressWithoutRender( Image & output )" :)

ihhub: We can do a trick to create another class which can accept in constructor either `Image` or `Display`. Based on given input it could behave one way or another. But, we need to think about a proper solution for this.

ihhub: However, I suggest to replace (in another pull request) many instances of
```cpp
button.isEnabled() && le.isMouseLeftButtonPressedInArea( button.area() ) ? button.drawOnPress() : button.drawOnRelease();
```
by
```cpp
button.drawOnState( le.isMouseLeftButtonPressedInArea( button.area() ) );
```
This will be a step forward for maintaining and improving the code. What do you think?

Districh-ru: I fully agree that the next (in a new PR) we should use `drawOnState()` in any place we can to have the same logic in all places.
And the next we should think about "forced render" issue and using `Image` instead of `Display`.
Maybe by using new method `drawAndRenderToScreen()` that will always use `Display` and render to it and another function with `Image` as input argument that will only `draw()` without `render()`.

src/fheroes2/gui/ui_button.cpp
```diff
@@ -520,60 +464,39 @@ namespace fheroes2
...
     int ButtonGroup::processEvents()
     {
         LocalEvent & le = LocalEvent::Get();
 
-        for ( size_t i = 0; i < _button.size(); ++i ) {
-            if ( _button[i]->isEnabled() ) {
-                le.isMouseLeftButtonPressedInArea( _button[i]->area() ) ? _button[i]->drawOnPress() : _button[i]->drawOnRelease();
+        for ( const auto & button : _button ) {
+            if ( button->isEnabled() ) {
+                le.isMouseLeftButtonPressedInArea( button->area() ) ? button->drawOnPress() : button->drawOnRelease();
```
ihhub, line 499: Since we are touching this code should we use `drawOnState()` call here?

src/fheroes2/gui/ui_button.h
```diff
@@ -131,32 +139,68 @@ namespace fheroes2
...
     public:
-        ButtonSprite( int32_t offsetX = 0, int32_t offsetY = 0 );
-        ButtonSprite( int32_t offsetX, int32_t offsetY, Sprite released, Sprite pressed, Sprite disabled = Sprite() );
+        explicit ButtonSprite( const int32_t offsetX = 0, const int32_t offsetY = 0 )
+            : ButtonBase( offsetX, offsetY )
+        {
+            // Do nothing.
+        }
+
+        ButtonSprite( const int32_t offsetX, const int32_t offsetY, Sprite released, Sprite pressed, Sprite disabled = Sprite() )
```
ihhub, line 195: We can use `Sprite disabled = {}` to have more flexibility but this is optional.

ihhub (changes requested):
Hi @Districh-ru , I left 2 more comments here. Can you please take a look at them?

src/fheroes2/gui/ui_button.cpp
Districh-ru: Hi, @ihhub, sure!
I've started to do such changes in the code and will open a PR soon.

src/fheroes2/gui/ui_button.h
Districh-ru: Done!

ihhub (approved):

ihhub:
@Districh-ru , thank you very much for this cleanup!

PR #9207, opened 2024-10-16
Shorten map title when too long in several places

zenseii:
Fixed in map info in editor, and map info and title in scenario setup.

The map title in the map setup dialog aka `game_scenarioinfo` will adjust the available length according to the width of the select button which can vary according to the language.

I chose to always center-align the text because making the text move to the left and make use of the available space there looked weird. The empty space on the left will always be equal to the width of the select button, this will ensure that if we implement a second button there with a new feature, there will be space for this new button to be of equal length to the select button so that we keep symmetry.

This PR:
image

image


ihhub (approved):

ihhub:
@zenseii , thank you so much for this fix! Actually, I had almost the same code when I was trying to allow longer map names :)

PR #9213, opened 2024-10-18
Fix incorrect Scrollbar position after it is upscaled

### LINKED ISSUE #9212: Game, problem with scroll bar in "File to Load:" and "File to Save:" windows
### Preliminary checks

### Platform

Windows

### Describe the bug

In fheroes2, there is a problem with scroll bar in "File to Load:" and "File to Save:" windows.

In fheroes2:

Load save windows 001

Load save windows 002

### Save file

Here is the save file (1.1.2 9210):
[Problem scroll bar 01.zip](https://github.com/user-attachments/files/17434653/Problem.scroll.bar.01.zip)

### Additional info

_No response_

Districh-ru:
Fix #9212

Currently to update the scrollbar image position after its image is changed we call `moveToIndex( _currentIndex )` just to update image and not to move the slider.

In the future we should do position update in `setImage()` function for this case and use `moveToIndex()` only when we want to move the scrollbar (by mouse or by keyboard).
This change needs many changes in the code and tests so in this PR I've only reverted a change in `moveToIndex()` done in #9201.

### DIFF
--- a/src/fheroes2/gui/ui_scrollbar.cpp
+++ b/src/fheroes2/gui/ui_scrollbar.cpp
@@ -68,10 +68,8 @@ namespace fheroes2
 
     bool Scrollbar::moveToIndex( const int indexId )
     {
-        if ( _currentIndex == indexId ) {
-            // Nothing to change.
-            return false;
-        }
+        // TODO: Make a function to update the Scrollbar position after its image is changed
+        // and call it in the code instead of this one with 'indexId' that is equal to '_currentId'.
 
         const int32_t roiWidth = _area.width - width();
         const int32_t roiHeight = _area.height - height();


Districh-ru:
Hi, @LeHerosInconnu, thanks for finding this bug!

I could reproduce it only after I moved most of the save files out from Save folder to have less files than the dialog can display.
Thу change in this PR fixed this bug for me.

Could you please test this PR if you have time?

ihhub (approved):

ihhub:
@Districh-ru , thank you for the quick fix!

@LeHerosInconnu, please use master branch the latest commit for testing.

LeHerosInconnu:
Hello @Districh-ru and @ihhub,

> Could you please test this PR if you have time?

> @LeHerosInconnu, please use master branch the latest commit for testing.

I tested this and it seems to work correctly with version 1.1.2 9205.

Districh-ru:
@LeHerosInconnu, thank you for testing!

PR #9217, opened 2024-10-19
Fix invalid behavior and crash while interacting with whirlpools by heroes on boats

### LINKED ISSUE #9141: The boat wipes out part of the whirlpool when the hero who was in it disembarks
### Preliminary checks

### Platform

Windows

### Describe the bug

Found during testing #9140. On the `master` branch:

https://github.com/user-attachments/assets/7330fecc-44aa-4621-9c1c-59a110a4e733

After the hero disembarks from a boat located on one of the tiles of the whirlpool, the `OBJ_BOAT` replaces the `OBJ_WHIRLPOOL` on that tile. On the `master` branch, another hero is still able to teleport to that tile with the `OBJ_BOAT`.  If the hero gets back into the boat and sails away from the whirlpool, the whirlpool object on that tile is not being restored, but another hero still can teleport to it:

https://github.com/user-attachments/assets/5f6d66a2-6b57-4761-b843-78ba96aca4e4

With the fixes from #9140, the situation is a little different - `OBJ_BOAT` still replaces the `OBJ_WHIRLPOOL` upon disembarking, but another hero cannot teleport to that tile anymore (regardless of whether it's occupied with the boat or not):

https://github.com/user-attachments/assets/126e2320-9a2c-4ec6-bcd9-ece111e41564

### Save file

Test map:

[!!TEST1.zip](https://github.com/user-attachments/files/17121630/TEST1.zip)

### Additional info

_No response_

### LINKED ISSUE #9214: Game crashes during AI turn
### Preliminary checks

### Platform

Windows

### Describe the bug

The app closes during the AI turn. No assertion is triggered.

https://github.com/user-attachments/assets/758146fa-4d8c-42cf-8183-3e12044355be

### Save file

[Lattice_work_34.zip](https://github.com/user-attachments/files/17441486/Lattice_work_34.zip)

### Additional info

_No response_

ihhub:
- update object type after a hero boards a boat and removes the boat sprite from a tile (close #9141)
- fix a crash while a hero tries to disembark from a whirlpool (close #9214)

What was the problem:
1. A hero "owns" tile's object type while stepping on a tile. If a tile was for example marked as bush, then a hero class remembers this type and sets the tile's object type as a hero.
2. In case when a hero board a boat he remembers the type as "boat" and marks as a "hero" object type.
3. In situation when a hero travels by boat and "steps on" an object the tile is going to be marked as "hero" object type. More importantly, if an object was marked as main addon / main object part of that tile it will be moved to bottom layer objects (see `Maps::Tiles::setBoat()` method). This is the **first problem**.
4. However, when a hero standing on a tile with an object steps out of a boat the tile is marked as "boat" object type.
5. When a hero boards that boat based on the previous logic the tile is going to be marked as "empty" as no object exists (see `void ActionToBoat()` function and lines:
```cpp
hero.setObjectTypeUnderHero( MP2::OBJ_NONE ); <--- here we explicitly set that no object was under hero
destinationTile.resetObjectSprite(); <--- here we remove the boat information
```
This is the **second problem**.
6. When a hero steps out of the boat and a player makes a save the tile with a boat is going to be marked as "boat" object type. When the player loads this saves and we actually search for all tiles with Whirlpools that particular tile is not going to be marked as whirlpool since the object type is different. This is the **third problem**.

The solution:
1. To fix the first problem we introduce `Maps::getObjectPartByActionType()` function which search for any object part related to a specific object type. So, previously the logic was expecting whirlpool to be set as a main object part / addon but it could be moved by `Maps::Tiles::setBoat()` method to bottom layer. So, if a tile is marked as Whirlpool we search a whirpool part inside the main object part and within all objects on the bottom layer of the tile.
2. To fix the second problem we update tile's object type before a hero remember it but after we remove the boat:
```cpp
destinationTile.resetMainObjectPart(); <--- we remove boat's information from the tile
destinationTile.updateObjectType(); <--- update object type for this tile if any object previously existed
hero.setObjectTypeUnderHero( destinationTile.GetObject( true ) ); <--- set the updated object type for hero
destinationTile.SetObject( MP2::OBJ_HERO ); <-- set the tile's object type as hero
```
3. To fix the third problem we need to scan every tile on the map for existence of whirlpool object parts even if a tile is not marked as whirlpool. To solve this problem `Maps::getObjectParts()` function was introduced.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -29,6 +29,7 @@
 #include <memory>
 #include <ostream>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "ai_common.h"
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1729,9 +1729,9 @@ namespace
         if ( !hero.isObjectTypeVisited( objectType, Visit::GLOBAL ) ) {
             hero.SetVisited( tileIndex, Visit::GLOBAL );
 
-            const auto eyeMagiIndexes = Maps::getObjectParts( MP2::OBJ_EYE_OF_MAGI );
+            const auto & eyeMagiIndexes = world.getAllEyeOfMagiPositions();
             const uint32_t distance = GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES );
-            for ( const auto & [index, objectPart] : eyeMagiIndexes ) {
+            for ( const int32_t index : eyeMagiIndexes ) {
                 Maps::ClearFog( index, distance, hero.GetColor() );
             }
         }
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1730,7 +1730,7 @@ namespace
 
             const auto eyeMagiIndexes = Maps::getObjectParts( MP2::OBJ_EYE_OF_MAGI );
             const uint32_t distance = GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES );
-            for ( const auto& [index, objectPart] : eyeMagiIndexes ) {
+            for ( const auto & [index, objectPart] : eyeMagiIndexes ) {
                 Maps::ClearFog( index, distance, hero.GetColor() );
             }
         }
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1664,6 +1664,7 @@ namespace
         hero.setDirection( boatDirection );
 
         destinationTile.resetObjectSprite();
+        destinationTile.updateObjectType();
         hero.setObjectTypeUnderHero( destinationTile.GetObject( true ) );
         destinationTile.SetObject( MP2::OBJ_HERO );
 
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1728,9 +1728,9 @@ namespace
         if ( !hero.isObjectTypeVisited( objectType, Visit::GLOBAL ) ) {
             hero.SetVisited( tileIndex, Visit::GLOBAL );
 
-            const MapsIndexes eyeMagiIndexes = Maps::GetObjectPositions( MP2::OBJ_EYE_OF_MAGI );
+            const auto eyeMagiIndexes = Maps::getObjectParts( MP2::OBJ_EYE_OF_MAGI );
             const uint32_t distance = GameStatic::getFogDiscoveryDistance( GameStatic::FogDiscoveryType::MAGI_EYES );
-            for ( const int32_t index : eyeMagiIndexes ) {
+            for ( const auto& [index, objectPart] : eyeMagiIndexes ) {
                 Maps::ClearFog( index, distance, hero.GetColor() );
             }
         }
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1666,7 +1666,7 @@ namespace
         destinationTile.resetObjectSprite();
         hero.setObjectTypeUnderHero( destinationTile.GetObject( true ) );
         destinationTile.SetObject( MP2::OBJ_HERO );
-        
+
         hero.SetShipMaster( true );
 
         // Boat is no longer empty so we reset color to default
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1663,6 +1663,7 @@ namespace
         hero.setDirection( boatDirection );
 
         destinationTile.resetObjectSprite();
+        destinationTile.AddonsSort();
         destinationTile.updateObjectType();
         hero.setObjectTypeUnderHero( destinationTile.GetObject( true ) );
         destinationTile.SetObject( MP2::OBJ_HERO );
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1662,8 +1662,7 @@ namespace
         // Set the direction of the hero to the one of the boat as the boat does not move when boarding it
         hero.setDirection( boatDirection );
 
-        destinationTile.resetObjectSprite();
-        destinationTile.AddonsSort();
+        destinationTile.resetMainObjectPart();
         destinationTile.updateObjectType();
         hero.setObjectTypeUnderHero( destinationTile.GetObject( true ) );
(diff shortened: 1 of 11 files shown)


ihhub:
Hi @zenseii and @oleg-derevenetz , could you please check this pull request whether it fixes the mentioned issues?

ihhub:
> Hi @zenseii and @oleg-derevenetz , could you please check this pull request whether it fixes the mentioned issue?

Please hold on with the review. I found much bigger problem with the code.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1661,8 +1661,13 @@ namespace
 
         // Set the direction of the hero to the one of the boat as the boat does not move when boarding it
         hero.setDirection( boatDirection );
-        hero.setObjectTypeUnderHero( MP2::OBJ_NONE );
+
         destinationTile.resetObjectSprite();
+        destinationTile.AddonsSort();
+        destinationTile.updateObjectType();
```
oleg-derevenetz, line 1667: Eh, this looks a bit cryptic, so many tile-specific actions, very easy to miss something. Is there any way to simplify/wrap this for a user like me?

src/fheroes2/world/world.cpp
```diff
@@ -828,11 +834,23 @@ MapsIndexes World::GetWhirlpoolEndPoints( const int32_t index ) const
         return result;
     }
 
-    // The exit point from the destination whirlpool must match the entry point in the source whirlpool.
-    for ( const int32_t whirlpoolIndex : _allWhirlpools.at( entranceTile.getMainObjectPart()._imageIndex ) ) {
+    const Maps::TilesAddon * entranceObjectPart = Maps::getObjectPartByActionType( entranceTile, MP2::OBJ_WHIRLPOOL );
+    if ( entranceObjectPart == nullptr ) {
+        return result;
```
oleg-derevenetz, line 839: It is not entirely clear to me here: if, for instance, there is a Whirlpool tile with a Boat (or a Hero in a boat), will some `TilesAddon` be returned? Technically, the "Whirlpool part" is still here... If it should be returned, then why we do `assert( 0 )` for a Stone Liths tile above, but not here?

oleg-derevenetz (commented):
Hi @ihhub I just left a few questions, could you please take a look when possible?

ihhub:
> I've tested it on several maps and it works well.
> 
> I did notice in the save provided by LiquidYetti that the previously broken whirlpool tile will now contain an `ICN object type` as unknown. I didn't notice this happening in my fheroes2 map:
> 
> ```
> 19.10.2024 13:07:28: [DBG_DEVEL]        Interface::AdventureMap::mouseCursorAreaPressRight:
> ******* Tile info *******
> Tile index      : 7096, point: (76, 65)
> MP2 object type : 167 (Whirlpool)
> UID             : 0
> ICN object type : 0 (UNKNOWN)
> image index     : 255
> layer type      : 0 - Object layer
> is shadow       : no
> region          : 36
> ground          : Ocean (isRoad: 0)
> ground img index: 13, image flags: 4
> passable from   : center, top, top-right, right, bottom-right, bottom, bottom-left, left, top-left
> metadata value 1: 0
> metadata value 2: 0
> metadata value 3: 0
> --------- Level 1 --------
> UID             : 62
> ICN object type : 50 (OBJNWATR.ICN)
> image index     : 202
> layer type      : 1 - Background layer
> is shadow       : no
> --- Extra information ---
> ```
> 
> If this is expected then I'm ready to approve the PR.

Hi @zenseii , this is a leftover from the previous logic. A hero jumped into a boat before making this save file. If it is done with the proposed logic the main object remains whirlpool. However, the engine works even with this case as I added required changes for it.

Districh-ru:
Hi, @zenseii and @ihhub, it happens because a Summon Boat spell was used by the Purple AI player. And its behavior was not changed in this PR.
Technically it is OK - OBJNWATR.ICN object is now in the bottom layer addons, but yes - it looks a little weird that it is not moved back to the main addon. :)
We can update this logic in a new PR.

Branikolog:
Hi, @ihhub 
An assertion occurred, when I summon a boat standing on a lighthouse and board it.
[Assertion.zip](https://github.com/user-attachments/files/17445063/Assertion.zip)

Summon a boat with Astra and step on it.
image

Districh-ru:
> Hi, @ihhub An assertion occurred, when I summon a boat standing on a lighthouse and board in it.

@ihhub,
the `AddonsSort()` method moves the `MP2::OBJ_ICN_TYPE_FLAG32` to the `_mainAddon` which is probably not good.
We may change the behavior of `AddonsSort()` not to move `MP2::OBJ_NONE` objects to the `_mainAddon` by inserting between the 884 and 885 lines the next code:

```cpp
        if ( getObjectTypeByIcn( highestPriorityAddon._objectIcnType, highestPriorityAddon._imageIndex ) == MP2::OBJ_NONE ) {
            return;
        }
```

Or we should only don't move only `highestPriorityAddon._objectIcnType == MP2::OBJ_ICN_TYPE_FLAG32`?
What do you think?

src/fheroes2/ai/ai_hero_action.cpp
ihhub: You are being too polite. This logic looks horrible :)

ihhub:
Hi @Districh-ru , I removed object part sorting call for now as I don't think a solution would be that simple.

src/fheroes2/world/world.cpp
ihhub: We are dealing with 2 different cases:
1. If a tile is marked as any other object but it contains a part of Whirlpool / Stone Liths. In this case we have to run through each object part for each tile to get a list of objects. This is done for Whirlpool as this is the only object so far which could be "covered" by a boat.
2. If a tile is marked as Whirlpool / Stone Liths but it doesn't contains any associated object part. This is the case with Stone Liths since we are getting image index for the object part so it is important to verify that the object indeed exists on this tile.

ihhub:
Hi @Branikolog , I fixed the assertion. Could you please take a look again?

Districh-ru (approved):

src/fheroes2/world/world.cpp
ihhub: No, it is not a legitimate case. If a tile is marked as `OBJ_WHIRLPOOL` then one of object parts on the tile must be `OBJ_WHIRLPOOL` type. If a boat or a hero on that tile then the tile is not going to be marked as `OBJ_WHIRLPOOL`.

oleg-derevenetz (approved):

ihhub:
Hi @oleg-derevenetz and @Districh-ru , I updated the description of this pull request to explain in details the problems and how they were fixed. I hope now it clarifies many thing. Please let me know if something is not clear.

src/fheroes2/ai/ai_hero_action.cpp
ihhub: I added extra comments for clarification.

src/fheroes2/world/world.cpp
ihhub: You were right. I added extra assertions for Whirlpools and also added comments for them.

Branikolog (approved):

PR #9218, opened 2024-10-19
Various minor improvements

oleg-derevenetz:
This PR:

* Rewrites the `Rand::Queue` logic. The previous logic was weird and did many nonsense, there is a way to do the same much easier and without rewriting the contents of the queue (the whole process can be `const`);
* Simplifies the `BagArtifacts::exchangeArtifacts()`. Instead of passing bags and heroes separately, this method has been turned into `static` and accepts only exchanging heroes. Also a long-standing TODO regarding saving a slot for a spellbook has been implemented;
* Performs a minor modernization of the code related to `Army` and `Troop`.


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for these improvements!

PR #9219, opened 2024-10-19
Add u with two dots to buttons and translate and improve German buttons

zenseii:
u with diaeresis

Fix several buttons and write out the entire word without abbreviations.


src/fheroes2/gui/ui_font.cpp
```diff
@@ -5639,7 +5639,7 @@ namespace
             released[charCode - 32].setPosition( buttonFontOffset.x, buttonFontOffset.y - 3 );
         }
 
-        for ( const int & charCode : { 196, 197, 214 } ) {
+        for ( const int & charCode : { 196, 197, 214, 220 } ) {
```
ihhub, line 5642: Please remove `&` sign from this line. We don't need to get an integer by reference.

ihhub (changes requested):
Hi @zenseii , I suggested a tiny correction in the code. Can you please check it?

ihhub (approved):

ihhub:
@zenseii , many thanks for this fix!

PR #9222, opened 2024-10-21
Do not download the wing32.dll for HoMM2 demo version

oleg-derevenetz:
I have no idea why we are downloading some 3-rd party DLL on Windows along with the demo version. This was done back when BAT script was used, and I suppose that `H2DEMOW.EXE` somehow requires this DLL, but the main purpose of the demo version download script is to gain access to the demo version assets in order to run fheroes2 with them. It is not intended to provide the user with the opportunity to directly play the original HoMM2 demo. IMHO we should not download some questionable DLL files to the user's computer.


oleg-derevenetz:
> it's possible to get only specific files from the zip archive too

This would be suboptimal. ZIP archive has ~20 mb in size, while uncompressed `HEROES2.AGG` has ~42 mb in size, so it's better and faster to download the whole ZIP. But downloading some 3-rd party DLL is a bit too much.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this update! It's time to remove this file as demo isn't helpful anymore.

PR #9226, opened 2024-10-22
Rename Addon into Object Part

ihhub:
As highlighted in #9217 map tile object related code is a black box for most of developers. This code is very convoluted and hard to read. This pull request does simple renaming of **Addon** into **Object Part** to make it easier to read.

Other naming is welcome as well!


oleg-derevenetz (approved):

Districh-ru (approved):
👍

PR #9229, opened 2024-10-22
Fix the penalty calculation for enemy units that can evade the engagement during battle

### LINKED ISSUE #9225: Combat, AI tends to prioritize the strongest stack without considering its flying ability/higher speed
### Preliminary checks

https://github.com/user-attachments/assets/1fc1c0e3-59e5-431e-a8e0-9c43aba1dc1e

The current AI allows player to kite him using flying creatures.

I expect AI to focus on slower stack, especially if it is ranged troops. Unless AI can reach the flying one within the current move of course. 
Also AI can split his troops and focus several enemies at once: running with all his stacks after the genie or stack of Rocs has no sense - he never reaches it until player decides to go for melee himself.

### Additional info

_No response_

oleg-derevenetz:
close #9225

With this PR:

https://github.com/user-attachments/assets/cf40cbd8-be47-4535-b4e9-0ca60d96e123


src/fheroes2/ai/ai_battle.cpp
```diff
@@ -1551,10 +1540,34 @@ AI::BattleTargetPair AI::BattlePlanner::meleeUnitOffense( Battle::Arena & arena,
...
+                const uint32_t enemySpeed = enemy->GetSpeed( false, true );
+                // If the enemy unit is immovable, it will not be able to evade an engagement
+                if ( enemySpeed == Speed::STANDING ) {
+                    return baseDist;
+                }
+
+                const uint32_t penaltyDist = baseDist + Battle::Board::widthInCells + Battle::Board::heightInCells;
+
+                // Flying unit, even inferior in speed, can move around the entire battlefield and easily evade an engagement
+                if ( enemy->isFlying() ) {
```
nogov, line 1561: If a flying unit is slowed, it can't fly over the whole battlefield. Does it make sense to take this into account?

oleg-derevenetz: Hi @nogov you definitely should take a closer look at the implementation of this method.

Branikolog:
Hi, @oleg-derevenetz !
I've started testing this PR.
I've noticed, that AI tends to focus ranged troops no matter how strong they are now:
image
image

If my stack has lower speed, than enemy golems - part of them can focus my stack being not too far away from it (even if they are not close enough to directly attack during their current turn, which looks ok).
In a case my strongest stack is same or higher speed - enemy golems always focus my ranged stack, which sometimes looks not so optimal (1 elf is not a threat at all for the enemy)

Oh, speaking about the threat coming from my ranged troops... I've noticed that within this PR AI stopped calculating the threat from the ranged troops focusing the highest one on default:
image

Could you, please, dive into the logic once again to investigate if something went wrong? (AI is focused on the strongest stack in master branch.)

oleg-derevenetz:
Hi @Branikolog

> I've noticed, that AI tends to focus ranged troops no matter how strong they are now:

It's not quite like that. They will not try to "pursue" faster units, that's true, but only if they are not reachable immediately. If such a faster unit gets within attack range, it will be attacked:

https://github.com/user-attachments/assets/1db58a2d-276b-4022-a816-6b4f8b88cabe

The idea here is to get faster units to protect the shooters at the risk of destroying those shooters, rather than running around the field distracting attention and not getting damage while the shooters are shooting.

> In a case my strongest stack is same or higher speed - enemy golems always focus my ranged stack, which sometimes looks not so optimal (1 elf is not a threat at all for the enemy):

A melee unit of any size staying far away from your troops won't do you any damage, but a shooter always deals damage regardless of range, even if there is just a single shooter.

> I've noticed that within this PR AI stopped calculating the threat from the ranged troops focusing the highest one on default:

This should be fixed now:

https://github.com/user-attachments/assets/970e8f17-2f1a-420c-8172-43007063b497

Branikolog:
@oleg-derevenetz 

> It's not quite like that. They will not try to "pursue" faster units, that's true, but only if they are not reachable immediately. If such a faster unit gets within attack range, it will be attacked:

Yeah. This works fine, that's why I didn't mention such behaviour. 👍 

>A melee unit of any size staying far away from your troops won't do you any damage, but a shooter always deals damage regardless of range, even if there is just a single shooter.

Indeed, but I was thinking that in a case shooters do ridiculously low damage, comparing to opposing troops and comparing to other allied stacks focusing them looks not really effective. But OK. It's definitely out of the scope of this PR.

>This should be fixed now:

Great! Now Enemy tends to focus the strongest ranged stack. 👍 
image

In a case we have 100% same stacks enemy troops focus them both +/- equally:
image

And that's fine.

But if one stack is slightly stronger AI prefers focusing it more that the weaker one:
image
image

I believe AI could focus weaker (but closer)  ranged stack with one allied stack of dwarves at least,. Lets assume 50% weaker ranger could be focused with 1 dwarf stack in this situation...
If this issue is out of the PR, we can discuss it separately. :)

Also, with changes of the PR AI tends to ignore fliers a little more, than I expected:
image

Overall, AI performs better now.  👍 
But in the case flier is in the range of some allied enemy stack, It would be nice to allow several more stacks to focus the flier too (maybe if they can reach flier on the next turn so they can cover the ones fighting genie at the moment). Now all dwarves, that cannot reach Genie - rushing rangers. That leads to the situation, where genie fights with just 1-2 stacks, and the rest running towards rangers. I believe depending on the genies/rangers threat ratio and the distance to rangers dwarves could focus genies slightly more. Like 3-4 stacks focusing Genies, and 1-2 stacks crawling towards rangers. This also could be effective against fast walking troops. How do you think? 
I know, it's definitely out of the scope of this PR and the only thing that bothers me here is the issue with rangers threat estimation I mentioned above.

oleg-derevenetz:
> But if one stack is slightly stronger AI prefers focusing it more that the weaker one:

I switched the distance penalty calculation from logarithmic scale to linear, so it should be better now:

https://github.com/user-attachments/assets/a51788dd-42c5-4ec4-9a6c-12e530f08c28

> But in the case flier is in the range of some allied enemy stack, It would be nice to allow several more stacks to focus the flier too (maybe if they can reach flier on the next turn so they can cover the ones fighting genie at the moment). Now all dwarves, that cannot reach Genie - rushing rangers. That leads to the situation, where genie fights with just 1-2 stacks, and the rest running towards rangers. I believe depending on the genies/rangers threat ratio and the distance to rangers dwarves could focus genies slightly more. Like 3-4 stacks focusing Genies, and 1-2 stacks crawling towards rangers. This also could be effective against fast walking troops. How do you think?

I think it won't work very well. Flying units can move all over the field, so with the current mechanics (when each unit makes decisions independently based just on threat and distance) there always can be a situation when enemy flyer will fly a little closer to the stacks that are moving to rangers (while rangers are still relatively far away, so flyer will be evaluated as "more threatening"), distract them and make them step back, then they can fly a little closer to other stacks and distract them, and so on. As a result, no one will get to the shooters.

Branikolog:
@oleg-derevenetz 

>I think it won't work very well. Flying units can move all over the field, so with the current mechanics (when each unit makes decisions independently based just on threat and distance) there always can be a situation when enemy flyer will fly a little closer to the stacks that are moving to rangers (while rangers are still relatively far away, so flyer will be evaluated as "more threatening"), distract them and make them step back, then they can fly a little closer to other stacks and distract them, and so on. As a result, no one will get to the shooters.

My idea was in focusing those flyers (for faster walkers it is going to work even better) when they are **already** **in** the attacking range  of 1 AI troop at least:

image

The goal is not to leave only 1 stack fighting with flyers and have 1-2 additional stacks that can reach flyer the next turn guaranteed. So Flyers will have just several options:
- Fly far away - so Golems would focus ranger than.
- Fly a few hexes away but stay in a range of 1 stack at least and be attacked (few support stacks will follow as well)
- Continue fighting and suffer from 2-3 stacks in the end. 

With such logic we cannot bait slow AI troops harmlessly - we have keep flyer in the range, otherwise - they ignore flyer and focus on attacking other troops. Of course AI should also evaluate threat of the flyer and not focus flyer with another golems not able to reach it on the current turn (the same as it already works in this PR).


>I switched the distance penalty calculation from logarithmic scale to linear, so it should be better now:

Great! Works much better now! The ratio is not perfect, AI is focused on the higher stack 20/14:

image

But for 20/15 I feel the difference 👍 
image

While crawling over the battlefield diagonally golems waste 2 more turns to reach the upper ranger, so in this particular case they could be focused on a short-distance attack more....
But it's definitely a step forward!

oleg-derevenetz:
Hi @Branikolog

> The ratio is not perfect, AI is focused on the higher stack 20/14:

What is the "perfect" ratio then? There is a different algorithm in the `master` branch, but I doubt you'll want to go back to it:

https://github.com/user-attachments/assets/1f5f0eaf-cf3a-45e6-aa7c-dd9c29e8ca03

https://github.com/user-attachments/assets/1b799780-0ac4-4a54-a264-ae79a80b283e

> My idea was in focusing those flyers (against faster walkers it is going to work even better) when they are already in the attacking range of 1 AI troop at least:
> [...]
> The goal is not to leave only 1 stack fighting with flyers and have 1-2 additional stacks that can reach flyer the next turn guaranteed.

Even if some enemy unit is in the attacking range of some other unit, this does not automatically mean that it will necessarily be attacked on that unit's turn. We are assessing the situation that is "now", but that unit will attack only "then" (when it will have its turn): it may be slowed down in between, some other enemy unit may approach this unit and will be preferred as a target, some other friendly unit may block the way, and so on. We can try it, but it's better to do it in a separate experimental PR.

Branikolog:
@oleg-derevenetz 
Hi!

>What is the "perfect" ratio then? There is a different algorithm in the master branch, but I doubt you'll want to go back to it:

Sure, the PR is definitely improving the logic. I personally expected lower stack of golems (which needs like 2 more additional rounds to reach the upper ranger compared to the lower one) to focus the lower stack in a cases when its threat is around 50-60% compared to the upper one. But I don't know how the logic works here and if it easy to adjust in such way.

I'm OK with approving the current solution, since it makes better the situation comparing to master branch. 👍 

>We can try it, but it's better to do it in a separate experimental PR.

Only if it's not a complex quest. I believe, there're more urgent tasks for AI logic in battles (not mentioning other engine sections 😃)

oleg-derevenetz:
> But I don't know how the logic works here and if it easy to adjust in such way.

Currently there is a linear scaling for enemy threat depending on distance, namely `enemyThreat / distance`, just that simple. Before there was also a logarithmic scale, namely `enemyThreat / ( 1.0 + ln( distance )` it does not give a very good distribution of attacking stacks (just like in your and in my videos). If there are other suggestions, let's consider them.

> Only if it's not a complex quest.

Well, it's not that difficult, we just need to see first how it will work at all.

Branikolog:
@oleg-derevenetz 

> Currently there is a linear scaling for enemy threat depending on distance, namely `enemyThreat / distance`, just that simple. Before there was also a logarithmic scale, namely `enemyThreat / ( 1.0 + ln( distance ) )` it does not give a very good distribution of attacking stacks (just like in your and my videos). If there are other suggestions, let's consider them.

I think the distance should have a little more weight in the formula. But I can't suggest any good solution right now. Since the PR makes things better, I think it's better to approve it and do some fine tuning later.

Branikolog (approved):
@oleg-derevenetz 
Many thanks for solving this problem and making AI much more fun to fight with!

ihhub (approved):

ihhub:

oleg-derevenetz:
Hi @Branikolog

> My idea was in focusing those flyers (against faster walkers it is going to work even better) when they are already in the attacking range of 1 AI troop at least

I just implemented a "skeleton" of this logic for testing purposes, and I should say that there are some flaws in it. It is about those "now" and "then". Consider:

https://github.com/user-attachments/assets/dee5d0be-d184-4e1f-a1a3-7b5b70749e7f

I know that Golems at cell 62 are able to approach the cell 20, so I put the flying unit (Gargoyles) to that cell. Golems at cell 73 know too that Golems at cell 62 are "now" able to attack Gargoyles, so they are moving towards cell 20. But then happens "then", and, due to blocked path, Golems at cell 62 are not able to approach the cell 20 anymore! As a result, at least one stack of Golems was distracted by Gargoyles without dealing any damage. In the `master` branch (after this PR), such a situation is just impossible.

Branikolog:
@oleg-derevenetz , hi!

Indeed, you're completely right. My proposed logic works not always good without path planner for all AI troops. 
I've noticed before, that AI tends to create a mess in his army when troop **a** walks around troop **b**, which cause troop **b** walk around troop **a**.

I appreciate, that you've tried to implement my idea. I suppose to continue working on it (or any similar) we should try to fix the cases, when AI allied troops block paths of each other. If we manage to find the solution, that would help fixing a mess, that is on your video, when golems in bottom part of the screen started blocking each other.
Another side of this problem is when one of attacking troops takes the space near the target, which is the only available space for other allied troop, while the first had enough movement points to move a bit further and allow other allied troops to attack target as well.

But the most urgent problem in my opinion is the case, when AI troop is out of attacking range and any troop on the battlefield  is blocking the path to the target ranger (rangers are the priority now, right?) This causes AI troop to use not optimal walkaround, which could be easily abused by players:

https://github.com/user-attachments/assets/d468d4d2-6dab-4ecc-9aa6-be014759e340

PR #9232, opened 2024-10-24
Fix some code smells related to the resource management

oleg-derevenetz:
* `AnimationReference`: already has defaulted move assignment operator, no harm to have defaulted move constructor as well;
* `VecCastles`: just a non-owning vector of pointers, no restrictions on copy/move are needed;
* `GameArea`: already has the defaulted copy constructor, no harm to have defaulted move constructor, default assignment operators are always deleted because of the `BaseInterface & _interface` member;
* `VecHeroes`: just a non-owning vector of pointers, no restrictions on copy/move are needed;
* `Recruit`: just a POD type, no restrictions on copy/move are needed;
* `SecSkills`: just a vector of POD types, no restrictions on copy/move are needed;
* `Kingdom`: already has defaulted move assignment operator, no harm to have defaulted move constructor as well.

### DIFF
--- a/src/engine/image.h
+++ b/src/engine/image.h
@@ -37,6 +37,7 @@ namespace fheroes2
     public:
         Image() = default;
         Image( const int32_t width_, const int32_t height_ );
+
         Image( const Image & image_ );
         Image( Image && image_ ) noexcept;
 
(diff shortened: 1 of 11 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks again for the cleanup!

PR #9237, opened 2024-10-25
Standartize member naming for Maps::FileInfo structure

ihhub:
We have two different structures: Maps::FileInfo and BaseMapFormat. Both of them have a lot of similarities. To avoid any confusion this rename has been done. Also, in the future we are going to modify Maps::FileInfo so that BaseMapFormat can own this structure.


Districh-ru (approved):

src/fheroes2/maps/maps_fileinfo.h
```diff
@@ -185,14 +185,14 @@ namespace Maps
         std::array<uint8_t, maxNumOfPlayers> races;
         std::array<uint8_t, maxNumOfPlayers> unions;
 
-        uint8_t kingdomColors;
-        uint8_t colorsAvailableForHumans;
-        uint8_t colorsAvailableForComp;
+        uint8_t availablePlayerColors;
+        uint8_t humanPlayerColors;
+        uint8_t computerPlayerColors;
```
oleg-derevenetz, line 190: Now these names look like these colors are specifically used by human (or computer) players while in fact it may not be really the case, because the same colors may be available both for a human player and for AI. In this regard, "colorsAvailableForXXX" sounds more right IMHO. But this is of course a matter of taste.

ihhub:
Closing this pull request as it is just variable renaming.

PR #9238, opened 2024-10-25
SonarCloud: Ignore cpp:S106 (Standard outputs should not be used directly to log anything) for command-line tools

oleg-derevenetz:
I'm not 100% sure that this will work, because, apparently, this cannot be checked using a pull request.


ihhub (approved):

ihhub:

PR #9239, opened 2024-10-26
Remove all objects marked as shadow from top layer

ihhub:
Top object parts do not have their own layer. None of them can be marked as a shadow. This pull request removes top object parts that somehow are marked as shadow. How it was verified: inside `Maps::Tile::isTallObject()` method I added assertions to these loops:
```cpp
    for ( const auto & part : _topObjectPart ) {
        assert( !isObjectPartShadow( part ) );
        if ( part._uid != 0 ) {
            tileUIDs.emplace_back( part._uid );
        }
    }
...
        for ( const auto & part : topTile._topObjectPart ) {
            assert( !isObjectPartShadow( part ) );
            if ( part._uid == tileUID ) {
                return true;
            }
        }
```
This helped me to detect all wrongly marked objects. Also, based on their location in respective to other object parts they appear to be at top layer.

This change might affect passabilities of certain objects.


Districh-ru (commented):
Hi, @ihhub, this PR affects the passability of two objects (cactus and crater) and make their passability the same as in the original game:

Master branch:  изображение изображение

This PR: изображение изображение

But we also should update the `Maps::ObjectInfo` for these two objects to have the same passability. They were made for our editor according to the passability in the master branch.
Or we should not do this update of `Maps::ObjectInfo`. IMHO the passability in the maser branch better fits these objects.

What do you think, @Branikolog?

Branikolog:
Hi, @Districh-ru 
I think the passability of the cactus from the main branch is logical: not being too high, it allows you to go around it on a closer trajectory.
The passability of the lava cracks from the main branch is just my personal preference.

ihhub:
Hi @Districh-ru  and @Branikolog , what is the passability of these 2 objects in the original game?

Districh-ru:
> Hi @Districh-ru and @Branikolog , what is the passability of these 2 objects in the original game?

Hi, @ihhub, their passability in the original game is the same as in this PR.
These objects have fully transparent upper tiles which makes them taller and restricts some passability.
To properly play the original maps it is better to fully replicate the passability of the original game.
So IMHO we should approve this PR because it make our engine more close to the original game.

But we have our own editor wit a new map format.
In our editor we did not include these transparent tiles to these objects and their passability in this PR is not changed and looks like it is in the master build.
The question was: should we fully replicate the passability of original game in our maps despite that this passability is not the best and the already implemented one is better (IMHO)? :)

Branikolog:
I am personally OK with rolling back changes and making passability same to the original one. Since we have nice visual representation for passability it's OK to have it same to the original game in fh2 editor.

ihhub:
Hi @Branikolog and @Districh-ru , I suggest to keep the passability for the original maps as in the original game. For our maps we can use other passability for objects map makers put on a map. As mentioned by @Districh-ru , we don't include invisible parts of the objects, resulting in different passability. I believe it is acceptable to do.

If you agree with my statement please mark this pull request appropriately.

Districh-ru (approved):

PR #9242, opened 2024-10-27
Improve tile-independent passability calculations

ihhub:
We should verify all object part on a single tile, not only the main part.

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -660,31 +660,49 @@ int Maps::Tile::getBoatDirection() const
     return Direction::UNKNOWN;
 }
 
-int Maps::Tile::getOriginalPassability() const
+int Maps::Tile::getTileIndepentPassability() const
 {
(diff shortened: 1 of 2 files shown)


Districh-ru (approved):
I've checked the Great War II and Broken Alliance maps and did not notice any issues.

Branikolog (approved):

src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -660,39 +660,57 @@ int Maps::Tile::getBoatDirection() const
     return Direction::UNKNOWN;
 }
 
-int Maps::Tile::getOriginalPassability() const
+int Maps::Tile::getTileIndepentPassability() const
```
oleg-derevenetz, line 663:
```suggestion
int Maps::Tile::getTileIndependentPassability() const
```

ihhub: Fixed. Thanks for spotting it, @oleg-derevenetz !

PR #9252, opened 2024-11-03
Update initial tile passability calculations

ihhub:
We need to check bottom objects from in an opposite order for passability as action objects are considered special ones and all calculations for object parts below them are nullified.

### DIFF
--- a/src/fheroes2/maps/maps_tiles.h
+++ b/src/fheroes2/maps/maps_tiles.h
@@ -346,7 +346,7 @@ namespace Maps
 
         void _updateRoadFlag();
 
-        bool anyTallObjectOnTile() const;
+        bool isTallObject() const;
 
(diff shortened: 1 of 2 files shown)


src/fheroes2/maps/maps_tiles.h
```diff
@@ -346,7 +346,7 @@ namespace Maps
 
         void _updateRoadFlag();
 
-        bool isTallObject() const;
+        bool anyTallObjectOnTile() const;
```
oleg-derevenetz, line 349: Hi @ihhub I would rename it to `isAnyTallObjectOnTile` to maintain the style :)

Districh-ru (approved):

oleg-derevenetz (approved):

PR #9259, opened 2024-11-07
Update German translations related to the Price of Loyalty Campaign

felix-schneider:
Including scenario descriptions and strings relating to the Battle Garb of Anduran.

Notes:
- The scenario descriptions now closely mirror the original.
- Kraeger is spelled with `ae` in the German original as well, so I kept it. But the original wasn't able to draw an `ä`, and so substitutes `ae`. It's unclear, but I'm making the call to keep it the same as the English.
- The Empire is referred to consistently as "Imperium" in the cutscenes, not "Reich", so the texts reflect this.
- The descriptions for the Anduran artifacts did not render, as do all other descriptions of artifacts with simple stat boosts. I have unmarked these translations from `fuzzy`, causing them to render, even though the grammatical gender will be wrong most of the time.


felix-schneider:
Done.

ihhub (approved):

ihhub:
@felix-schneider , thank you so much for your contribution to the project!

PR #9264, opened 2024-11-09
Move Streams handling in Editor from "world" tiles to "Map_Format"

Districh-ru:
Relates to #8589

This PR updates Streams placing, removing and connecting with river deltas in Editor.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -230,7 +230,7 @@ namespace
         bool updateMapPlayerInformation = false;
 
         // Filter objects by group and remove them from '_mapFormat'.
-        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size() && !objectsUids.empty(); ++mapTileIndex ) {
+        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size(); ++mapTileIndex ) {
             Maps::Map_Format::TileInfo & mapTile = mapFormat.tiles[mapTileIndex];
 
             for ( auto objectIter = mapTile.objects.begin(); objectIter != mapTile.objects.end(); ) {
@@ -385,7 +385,7 @@ namespace
                     needRedraw = true;
                 }
                 else if ( objectIter->group == Maps::ObjectGroup::LANDSCAPE_MISCELLANEOUS ) {
-                    const int riverDeltaDirection = Maps::getRiverDeltaDirectionByIndex( static_cast<int32_t>( objectIter->index ) );
+                    const int riverDeltaDirection = Maps::getRiverDeltaDirectionByIndex( objectIter->group, static_cast<int32_t>( objectIter->index ) );
 
                     objectIter = mapTile.objects.erase( objectIter );
 
@@ -404,6 +404,10 @@ namespace
                 if ( objectsUids.empty() ) {
                     break;
                 }
+            }
+
+            if ( objectsUids.empty() ) {
+                break;
             }
         }
 
@@ -466,7 +470,7 @@ namespace
         }
         case Maps::ObjectGroup::LANDSCAPE_MISCELLANEOUS: {
             // River deltas are only objects that can be placed on water and on land.
-            if ( Maps::getRiverDeltaDirectionByIndex( objectType ) != Direction::UNKNOWN ) {
+            if ( Maps::isRiverDeltaObject( groupType, objectType ) ) {
                 // This is a river delta. Just don't check the terrain type.
             }
             else if ( !checkConditionForUsedTiles( Maps::getObjectInfo( groupType, objectType ), tilePos,
@@ -1400,7 +1404,7 @@ namespace Interface
 
             fheroes2::ActionCreator action( _historyManager, _mapFormat );
 
-            if ( Maps::addStreamToMap( _mapFormat, tileIndex ) ) {
+            if ( Maps::addStream( _mapFormat, tileIndex ) ) {
                 _redraw |= mapUpdateFlags;
 
                 action.commit();
@@ -1649,7 +1653,7 @@ namespace Interface
             }
 
             // For River Deltas we update the nearby Streams to properly connect to them.
-            if ( const int riverDeltaDirection = Maps::getRiverDeltaDirectionByIndex( objectType ); riverDeltaDirection != Direction::UNKNOWN ) {
+            if ( const int riverDeltaDirection = Maps::getRiverDeltaDirectionByIndex( groupType, objectType ); riverDeltaDirection != Direction::UNKNOWN ) {
                 Maps::updateStreamsToDeltaConnection( _mapFormat, tile.GetIndex(), riverDeltaDirection );
             }
 
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -230,7 +230,7 @@ namespace
         bool updateMapPlayerInformation = false;
 
         // Filter objects by group and remove them from '_mapFormat'.
-        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size(); ++mapTileIndex ) {
+        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size() && !objectsUids.empty(); ++mapTileIndex ) {
             Maps::Map_Format::TileInfo & mapTile = mapFormat.tiles[mapTileIndex];
 
             for ( auto objectIter = mapTile.objects.begin(); objectIter != mapTile.objects.end(); ) {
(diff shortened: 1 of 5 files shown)


Branikolog (approved):

src/fheroes2/maps/map_format_helper.h
```diff
@@ -53,6 +53,17 @@ namespace Maps
...
+    bool addStreamToMap( Map_Format::MapFormat & map, const int32_t tileId );
+
+    // Update the existing streams around the center tile to properly connect them to the center stream.
+    void updateStreamsAround( Map_Format::MapFormat & map, const int32_t centerTileId );
+
+    // Update the existing streams to connect them to the River Delta.
+    void updateStreamsToDeltaConnection( Map_Format::MapFormat & map, const int32_t tileId, const int deltaDirection );
+
+    // Returns 'Direction::UNKNOWN' if this index belongs not to the River Delta.
+    int getRiverDeltaDirectionByIndex( const int32_t objectIndex );
```
ihhub, line 65: This function's real return value is used only in 2 places while in other places we just verify that the value is valid. I suggest to create a separate function that will return a boolean variable to simply check if an object is a river delta. For example,
```cpp
bool isRiverDeltaObject( const ObjectGroup group, const uint32_t objectIndex );

...

bool isRiverDeltaObject( const ObjectGroup group, const uint32_t objectIndex )
{
    return getRiverDeltaDirectionByIndex( group, objectIndex ) != Direction::UNKNOWN;
}
```

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -230,7 +230,7 @@ namespace
         bool updateMapPlayerInformation = false;
 
         // Filter objects by group and remove them from '_mapFormat'.
-        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size(); ++mapTileIndex ) {
+        for ( size_t mapTileIndex = 0; mapTileIndex < mapFormat.tiles.size() && !objectsUids.empty(); ++mapTileIndex ) {
```
ihhub, line 233: Having several conditions in a foor loop might be problematic for some developers to read. Since we already have this code below:
```cpp
                if ( objectsUids.empty() ) {
                    break;
                }
```

I suggest to do the same for this loop as well.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -116,6 +121,166 @@ namespace
...
         }
     }
+
+    bool isStreamOnMapTile( const Maps::Map_Format::TileInfo & mapTile )
+    {
+        return std::any_of( mapTile.objects.cbegin(), mapTile.objects.cend(),
+                            []( const Maps::Map_Format::TileObjectInfo & object ) { return object.group == Maps::ObjectGroup::STREAMS; } );
+    }
+
+    bool isNeedStreamToDeltaConnection( Maps::Map_Format::MapFormat & map, const int32_t tileId, const int direction )
```
ihhub, line 131: Let's call this function as `isStreamToDeltaConnectionNeeded`.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -116,6 +121,166 @@ namespace
...
+    // Returns the direction vector bits from 'centerTileIndex' to the around tiles with streams.
+    int getStreamDirecton( Maps::Map_Format::MapFormat & map, const int32_t currentTileId )
+    {
+        assert( currentTileId >= 0 && map.tiles.size() > static_cast<size_t>( currentTileId ) );
+
+        // Stream includes also the Deltas. Here we need to check only streams excluding Deltas.
+        int streamDirection = ( isStreamOnMapTile( map.tiles[currentTileId] ) ) ? Direction::CENTER : 0;
+
+        // For streams we can check only the next four directions.
+        for ( const int direction : { Direction::LEFT, Direction::TOP, Direction::RIGHT, Direction::BOTTOM } ) {
```
ihhub, line 163: Is there a need to run this loop when **streamDirection** is 0?

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -116,6 +121,166 @@ namespace
...
+    void updateStreamObjectOnMapTile( Maps::Map_Format::MapFormat & map, const int32_t tileId, const bool forceStreamOnTile )
+    {
+        const int direction = getStreamDirecton( map, tileId );
+
+        if ( !forceStreamOnTile && hasNoBits( direction, Direction::CENTER ) ) {
+            // There will be no stream on this tile.
+            return;
+        }
+
+        const uint8_t objectIndex = getStreamIndex( forceStreamOnTile ? ( direction | Direction::CENTER ) : direction );
```
ihhub, line 253: I am thinking should we avoid magic numbers in the code like here and just return a boolean variable for the function while the real value can be assigned into a second input argument. What do you think?

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -116,6 +121,166 @@ namespace
...
+        auto & tileObjects = map.tiles[tileId].objects;
+
+        auto streamIter = std::find_if( tileObjects.begin(), tileObjects.end(),
+                                        []( const Maps::Map_Format::TileObjectInfo & object ) { return object.group == Maps::ObjectGroup::STREAMS; } );
+
+        if ( streamIter == tileObjects.end() ) {
+            // Add stream to this tile.
+            const auto & objectInfo = Maps::getObjectInfo( Maps::ObjectGroup::STREAMS, objectIndex );
+
+            if ( !Maps::setObjectOnTile( world.getTile( tileId ), objectInfo, true ) ) {
```
ihhub, line 270: Do we need to do this on a tile of Adventure Map since we refresh the whole map after?

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -390,6 +533,82 @@ namespace Maps
         }
     }
 
+    bool addStreamToMap( Map_Format::MapFormat & map, const int32_t tileId )
```
ihhub, line 536: We can skip `toMap` as we are in map format helper code.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -390,6 +533,82 @@ namespace Maps
...
+    {
+        assert( tileId >= 0 && map.tiles.size() > static_cast<size_t>( tileId ) );
+
+        Map_Format::TileInfo & thisTile = map.tiles[tileId];
+
+        if ( isStreamOnMapTile( thisTile ) || world.getTile( tileId ).isWater() ) {
```
ihhub, line 542: I am wondering should we check group type using `thisTile` instead since we want to eliminate usage of **world** here?

ihhub (changes requested):
Hi @Districh-ru , I added several comments and suggestions here. Could you please take a look at them once you have time?

src/fheroes2/maps/map_format_helper.cpp
Districh-ru: Hi, @ihhub, previously it was needed because only after the call of this function we checked the `forceStreamOnTile`  and the directions around were needed for the further logic.
Now this function is changed to accept `forceStreamOnTile` and if it is **false** and `streamDirection` is **0** the function returns this **0**.

src/fheroes2/maps/map_format_helper.cpp
Districh-ru: Now `forceStreamOnTile` is used in the other place and this place looks much better.

src/fheroes2/maps/map_format_helper.cpp
Districh-ru: Yes, because when placing a stream there is no whole map refresh.
The whole map refresh is currently done for Erase and Undo/Redo operations.
In future we should avoid the whole map refresh and erase/place only one object.

src/fheroes2/maps/map_format_helper.cpp
Districh-ru: Now we use `Ground::getGroundByImageIndex( thisTile.terrainIndex )`.
Thanks!)

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -385,6 +384,18 @@ namespace
                     objectIter = mapTile.objects.erase( objectIter );
                     needRedraw = true;
                 }
+                else if ( objectIter->group == Maps::ObjectGroup::LANDSCAPE_MISCELLANEOUS ) {
+                    const int riverDeltaDirection = Maps::getRiverDeltaDirectionByIndex( objectIter->group, static_cast<int32_t>( objectIter->index ) );
```
ihhub, line 388: Here we should use `isRiverDeltaObject` function call as we don't use **riverDeltaDirection** anywhere else.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -116,6 +121,167 @@ namespace
             }
         }
     }
+
+    bool isStreamOnMapTile( const Maps::Map_Format::TileInfo & mapTile )
```
ihhub, line 125: Since we are passing `mapTile` as argument we could name this function as **isStreamPresent**. What do you think?

ihhub (commented):
Hi @Districh-ru , I left 2 minor comments here. Would you mind to check them?

src/fheroes2/editor/editor_interface.cpp
Districh-ru: We use **riverDeltaDirection** at line 394:
`Maps::updateStreamsToDeltaConnection( mapFormat, static_cast<int32_t>( mapTileIndex ), riverDeltaDirection );`

src/fheroes2/maps/map_format_helper.cpp
Districh-ru: Sure.

ihhub (approved):

ihhub:
@Districh-ru , thank you so much for this change!

PR #9265, opened 2024-11-10
Improve & speed up the translation-related code a bit

oleg-derevenetz:
* Added more sanity checks when loading a MO file;
* Store the translated strings themselves instead of just pointers to parts of the buffer. Previously, it was assumed that the strings in the buffer always null-terminated (and that's how it should be according to the MO format specification), but if the MO file is corrupted, then this may not be the case. As a result, this may result in reading beyond the buffer boundaries.
* Added support for both big-endian and little-endian versions of MO files with endianness autodetection;
* Removed the TODO comment regarding the hash table and an explanatory comment was written instead;
* Fixed a suspicious place in `agg_image.cpp` in regard to endianness;
* Added the ability for `ROStreamBuf` to work in two modes: "view mode" (working on top of the external buffer - same as before) and "owning mode" (transferring the ownership of an external buffer to the `ROStreamBuf` through the move operation). `StreamFile::getStreamBuf()` now returns the owning `ROStreamBuf` instead of the `RWStreamBuf`;
* `ROStreamBuf` is now able to create const "views" of data inside itself (no copies -> no dynamic memory allocations):
  * `getRawView()` as a "copyless" alternative to `getRaw()`;
  * `getStringView()` as a "copyless" alternative to `getString()`;
* Removed some obsolete or trivial `static_assert()` messages in various places;
* Fixed the behavior of `StringSplit()` function - in the `master` branch, it produces wrong results in certain cases, e.g. `StringSplit( "", '#' )` produces `{}` (should be `{ "" }`), `StringSplit( "#", '#' )` produces `{ "" }` (should be `{ "", "" }`), etc. The logic of this method is incorrect from the moment the original fheroes2 sources are imported.

### DIFF
--- a/src/engine/agg_file.cpp
+++ b/src/engine/agg_file.cpp
@@ -41,13 +41,13 @@ namespace fheroes2
             return false;
         }
 
-        RWStreamBuf fileEntries = _stream.toStreamBuf( count * fileRecordSize );
+        RWStreamBuf fileEntries = _stream.getStreamBuf( count * fileRecordSize );
         const size_t nameEntriesSize = _maxFilenameSize * count;
         _stream.seek( size - nameEntriesSize );
-        RWStreamBuf nameEntries = _stream.toStreamBuf( nameEntriesSize );
+        RWStreamBuf nameEntries = _stream.getStreamBuf( nameEntriesSize );
 
         for ( size_t i = 0; i < count; ++i ) {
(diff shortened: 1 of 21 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for these changes and speedups!

PR #9270, opened 2024-11-13
Android: update the desugar_jdk_libs_nio dependency to 2.1.3

oleg-derevenetz:
Here is the changelog:

https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md

In theory, there are no critical changes that can break something.


ihhub (approved):

ihhub:

PR #9278, opened 2024-11-18
Show a warning dialog upon a map file loading failure

### LINKED ISSUE #9250: Modified PoL maps loading with SW/demo resources
### Preliminary checks

In a case we try to load PoL map manually renamed in SW format the game just jumps back to the main menu:

https://github.com/user-attachments/assets/0bed78fd-89a2-4885-9bf7-90331ca038a9

I suggest showing some pop up window noting that the engine can not load the map due to lacking of the full version.

I also suggest mentioning in the first startup dialogs or scenario selection dialog, that you will not able to run any PoL or fh2 map.
image

### Additional info

_No response_

ihhub:
close #9250

Since the PoL map is renamed that it could be considered as corrupted as it was modified manually. We don't need to show a very detailed explanation for this matter.


src/fheroes2/game/game_scenarioinfo.cpp
```diff
@@ -434,6 +434,7 @@ namespace
 
         if ( mapInfo.version == GameVersion::SUCCESSION_WARS || mapInfo.version == GameVersion::PRICE_OF_LOYALTY ) {
             if ( world.LoadMapMP2( mapInfo.filename, ( mapInfo.version == GameVersion::SUCCESSION_WARS ) ) ) {
+                fheroes2::showStandardTextMessage( _( "Warning" ), _( "The map is corrupted." ), Dialog::OK );
```
oleg-derevenetz, line 437: It looks like this warning (and the warning below) will be shown if map has been successfully loaded. Shouldn't it be the other way around?

ihhub: You are absolutely right!

mind-james (commented):
The new dialogue shows up for maps with renamed extensions as intended, but the menu screen is only partly drawn (looks cropped) as long as the dialogue is visible.

mind-james (approved):
Seems good now. Tested by deleting expansion data file, then renaming Arrax map extension to mp2. Error dialog comes up and there is no cropping of the menu.

PR #9279, opened 2024-11-18
Fix a few more code smells

oleg-derevenetz:
* Flatten the structure of `zzlib.cpp`: move methods out of `namespace Compression {}`, use `Compression::` syntax instead;
* `Compression::zipStreamBuf()`: add `const`;
* `Troops::GetFirstValid()`: add `const`.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for keeping the code clean!

PR #9281, opened 2024-11-18
Synchronize maximum allowed number of characters for input and Virtual Keyboard

ihhub:
- if a string has a limit on length when virtual keyboard invoked from this dialog will have the same limit
- all save files now have a limit of 255 characters. Previously, the limit was per dialog size.


src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -478,7 +478,8 @@ namespace
                     needRedraw = true;
                 }
                 else if ( !listboxEvent && le.isAnyKeyPressed()
-                          && ( !isTextLimit || fheroes2::Key::KEY_BACKSPACE == le.getPressedKeyValue() || fheroes2::Key::KEY_DELETE == le.getPressedKeyValue() )
+                          && ( filename.size() >= lengthLimit || fheroes2::Key::KEY_BACKSPACE == le.getPressedKeyValue()
```
Districh-ru, line 481: Hi,@ihhub, I'm not able to input any character in save game dialog.
Here we should check that there is no input limit: `filename` is lower than `lengthLimit `:
```suggestion
                          && ( filename.size() < lengthLimit || fheroes2::Key::KEY_BACKSPACE == le.getPressedKeyValue()
```

ihhub: Hmm... no idea why I did this...

Districh-ru (approved):

Branikolog:
Hi, @ihhub !
I've made a couple of tests and find an issue:

https://github.com/user-attachments/assets/6e3d161b-fb80-4f9f-b5a1-ff435672a4dc

Districh-ru:
Hi, @Branikolog, the text input cursor position setting should be fixed now.

Branikolog:
@Districh-ru thanks for fixing it!
I've found one more similar issue:

https://github.com/user-attachments/assets/ed0ef815-b881-4a52-bac0-5321f9fbbbcc

I've used several spaces in a row for the 5th line and cursor appears to be shifted in the same manner, as you've fixed above. 
Could you, please, try to fix it?

Districh-ru:
> @Districh-ru thanks for fixing it! I've found one more similar issue:
> 2024-11-27.10-21-43.mp4
> 
> I've used several spaces in a row for the 5th line and cursor appears to be shifted in the same manner, as you've fixed above. Could you, please, try to fix it?

Hi, @Branikolog,
Unfortunately it is not so easy to fix it quickly.
Lets fix this issue in a separate PR because it is not related to maximum input characters limit synchronization.

Branikolog (approved):
Everything else is fine!
Thanks!

PR #9283, opened 2024-11-18
Improve the castle reinforcement logic for AI (implement the long-standing TODO comment)

oleg-derevenetz:


src/fheroes2/ai/ai_planner_castle.cpp
```diff
@@ -307,24 +311,215 @@ void AI::Planner::updateKingdomBudget( const Kingdom & kingdom )
...
+        guestHeroArmy.UpgradeTroops( castle );
+        guestHeroArmy.MergeSameMonsterTroops();
+    }
+
+    // It is allowed to hire non-upgraded units even if an upgraded dwelling is built
+    static const std::array<uint32_t, 12> castleDwellings{ DWELLING_UPGRADE7, DWELLING_UPGRADE6, DWELLING_MONSTER6, DWELLING_UPGRADE5,
+                                                           DWELLING_MONSTER5, DWELLING_UPGRADE4, DWELLING_MONSTER4, DWELLING_UPGRADE3,
+                                                           DWELLING_MONSTER3, DWELLING_UPGRADE2, DWELLING_MONSTER2, DWELLING_MONSTER1 };
+
+    for ( const uint32_t dwelling : castleDwellings ) {
```
ihhub, line 352: Hi @oleg-derevenetz , I might don't fully understand the proposed logic. Let me explain how I read this code:
- run through all possible monsters in the castle
- if no monster exists or no available for hire then do nothing
- if it does and available for hire then try to hire them for castle's (garrison's) army
- if cannot and a hero is in the castle, try to move one slot from castle to army
- if can move then hire a new monster to the castle's army
- if cannot move then find a weakest troops among castle's and hero's armies and dismiss it, hire the monsters.

If the described logic is correct then I see one flaw: we don't try to hire monsters into hero's army like it is done inside `Castle::RecruitMonster()` method.

Could you please verify this?

ihhub (commented):
Hi @oleg-derevenetz , I left one comment here. Could you please check it?

src/fheroes2/ai/ai_planner_castle.cpp
oleg-derevenetz: Hi @ihhub

> If the described logic is correct then I see one flaw: we don't try to hire monsters into hero's army like it is done inside `Castle::RecruitMonster()` method.

Why do you think so? Local `recruitMonster()` lambda in turn calls the `Castle::RecruitMonster()` method, I even added a comment regarding this, so that the reader of this code does not think that troops are hired only in the garrison.

ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for this improvement in the AI code!

PR #9284, opened 2024-11-19
embed simpleson JSON lib and possibility to load monsters data from external jsons

CMiSSioN:
1. Added third party JSON parsing library
2. Possibility to load monsters data from external JSON files

### DIFF
--- a/VisualStudio/fheroes2/common.props
+++ b/VisualStudio/fheroes2/common.props
@@ -2,7 +2,7 @@
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemDefinitionGroup>
     <ClCompile>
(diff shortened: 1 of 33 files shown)


oleg-derevenetz:
Hi @CMiSSioN

For the sake of curiosity, I looked at the proposed library for parsing JSON, and I want to say that it is written... um... in a very original way. For instance, it is immediately clear that `json::parsing::decode_string()` will perform out-of-bounds read if its argument doesn't end with a quotation mark (`"`) (even if it is properly zero-terminated string!). Here is the [test sample](https://godbolt.org/z/fcshnqG8n) (I just replaced the library's own exceptions with `std::invalid_argument`).

CMiSSioN:
Hi @oleg-derevenetz 
Just forget it, @ihhub already explained me how wrong i was trying to contribute. Please, consider this PR never existed.

Issue #9260, opened 2024-11-07
Similar items share translation strings, which is inappropriate for languages with gendered nouns


felix-schneider:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Consider these two items from the OG:
![image](https://github.com/user-attachments/assets/18814e89-b82f-440f-a3b8-9ab46051df4b)
![image](https://github.com/user-attachments/assets/00123ea9-eea2-4094-836b-ce31f778993c)
In English these two items (the Dragon Sword of Dominion and the Power Axe of Dominion) use the same translation string: `The %{name} increases the hero's attack skill by %{count}`, but in German the Axe is grammatically female and the Sword is grammatically neutral, requiring a different translation. The OG has separate strings for these.

Maybe related #5248 

### Save file

N/A

ihhub:
Hi @felix-schneider and @zenseii ,

I believe to move work on this issue we need to add artifact name to each artifact description. What do you think?

felix-schneider:
That is the solution taken by the OG.

For the artifact descriptions themselves, the solution proposed above has been implemented, and works well: Add the article to the artifact name and remove it from the description.

It continues to be awkward in other contexts where the artifact is referred to, like when it's guarded by a monster, or locked behind skill, because it should be in a different case. We do some grammatical juggling to make it work. I don't think that removing the article from the artifact name again will help with that. In fact, nothing short of implementing multiple forms of the artifact name would work, and which forms are needed presumably varies by language, so I would understand very much if that was too much effort for too little benefit.

PR #9285, opened 2024-11-19
Update German translations related to artifacts

felix-schneider:
Notes:
- (German) As per the discussion in #9260, the definite article is now part of the artifact name.
  - (German) As part of this, normal grammar would demand that an adjective that is part of the artifact name would not be capitalized, as it is no longer at the start of the string (such as "Immervoller Erzkarren" vs. "Der Immervolle Erzkarren"). Looking at this in-game, this just looked weird to me, like it's just any old ever-full ore cart? It's THE ever-full ore cart! So I capitalized the adjectives as well. In cases of proper nouns, this is acceptable (c.f. the popular German TTRPG "Das Schwarze Auge").
- (English) Pickup text for medal of Valor refers to "a" Medal of Valor (note capitalization). In other texts, when artifacts are reffered to by name, they are called "the" artifact.
  - Same for Medal of Courage and Medal of Honor, but not Medal of Distinction
- (English) Ultimate Crown description refers to basic skills. Shoudn't they be called primary skills? (as opposed to the secondary skills) Basic skills could be confused with secondary skills at the basic level (so one could think the crown upgrades those to advanced)
  - Ultimate Scepter calls them attributes, which is also quite rare
- (English) In the Medal of Distinction pickup text, the word "Knights" is capitalized. Is this meant to refer to heros of the Knight faction?
- (English) The English artifact description usually refer to "the hero", whereas the German ones (in the OG also) refer to you directly.
- (English) The Skullcap description refers to "all mind-influencing spells", whereas it should be "all mind-influencing spells **you cast**", it does not make your enemy's spells cheaper. Also, it's just a skullcap? Like that is an actual thing that exists. It doesn't have any adjective to make it magical, you're just wearing a Kippa?
- (English) The Evil Eye has for some reason a different wording on its description than the Skullcap and the Snake-Ring. It says "reduced by half" and the others save "halves".
- (German) I went with the shorter translation of spell as "Zauber" instead of the "Zauberspruch" from the OG. Especially when used as part of a compound, it leads to really long words. Until we have some kind of dynamic hyphenation, the long words do not render very nicely.
- (German) While I'm not touching spells in this PR, the Wand of Negation mentions Dispel Magic. I never really like the translation of that spell as "Magie zerstreuen", which is really long, especially the mass version. I instead translated it as "Gegenzauber" - Counterspell. Side note: Even the translators of Magic the Gathering did not find an adequate equivalent of the word "dispel" in German. The Magic card "Dispel" is called "Bannzauber" - "Banish spell" in German.
- (Both) The resource-generating artifacts have a replacement string for the amount, but they all say `%{count} unit` (singular), so if that amount ever changes, the string needs to be edited anyway, so why bother having it?
- (English) The Staff of Wizardry says "boosts your spell power" whereas all other artifacts say "increases your spell power"
- (German) German does not use the same quotation marks as English. In English, you use upper quotation marks (`\u0022` or more correctly `\u201C` and `\u201D`) at the start and end of a direct quote. In German, at the start of a quotation, you would use the "low double comma" `\u201E` at the start and upper double quotes `\u201C` at the end. Both of those can be represented in CP1252 (`0x84` and `0x93`, respectively), but sadly it looks like the in-game font has no glyphs for them. The OG seems to actually use single quotes (actually apostrophes, `\u0027`) in some cases (e.g. the spell scroll). I have not followed this example and used `\u0022` instead.


felix-schneider:
After further discussion on discord, I decided to go back and compare the translations more closely against the original game. Please don't merge this yet.

files/lang/de.po
```diff
@@ -11139,71 +11099,63 @@ msgstr ""
...
 "hope to pay. The tax man takes pity and agrees to only take 250 gold a day "
 "from your account for life. Check here if you want one dollar to go to the "
 "presidential campaign election fund."
 msgstr ""
-"Euer überschwängliches Konsumverhalten hat Euch eine Steuerrechnung "
-"eingebracht, die Ihr niemals werdet zahlen können. Der Finanzbeamte hat "
-"Mitleid und willigt ein, lebenslang nur 250 Gold pro Tag von Eurem Konto "
-"abzuziehen. Unterschreibt hier, wenn Ihr für den Wahlkampffonds des "
-"Praesidenten spenden wollt."
+"Eure überschwängliches Prasserei hat Euch eine Steuerrechnung eingebracht, "
```
senneschall, line 11147: *überschwängliche

files/lang/de.po
```diff
@@ -7274,36 +7275,37 @@ msgid ""
...
+msgstr "Ihr habt ein Artefakt gefunden: "
 
 msgid ""
 "You've found the humble dwelling of a withered hermit. The hermit tells you "
 "that he is willing to give the %{art} to the first wise person he meets."
 msgstr ""
 "Ihr habt die bescheidene Behausung eines vertrockneten Einsiedlers gefunden. "
-"Der Einsiedler erzählt Euch, dass er bereit ist, die %{art} dem ersten "
-"weisen Menschen zu geben, den er trifft."
+"Der Einsiedler erzählt Euch, dass er im Besitz eines Artefakts ist, dass er "
```
senneschall, line 7289: "Der Einsiedler erzählt Euch, er ist im Besitz eines Artefakts, von dem er bereit ist, es dem ersten weisen Menschen zu geben, den er trifft:"

files/lang/de.po
```diff
@@ -7274,36 +7275,37 @@ msgid ""
...
 
 msgid ""
 "You've come across the spartan quarters of a retired soldier. The soldier "
 "tells you that he is willing to pass on the %{art} to the first true leader "
 "he meets."
 msgstr ""
 "Ihr seid auf das spartanische Quartier eines pensionierten Soldaten "
-"gestoßen. Der Soldat erzählt Euch, dass er bereit ist, die %{art} an den "
-"ersten wahren Anführer weiterzugeben, den er trifft."
+"gestoßen. Der Soldat erzählt Euch, dass er im Besitz eines Artefakts ist, "
```
senneschall, line 7298: "Der Soldat erzählt Euch, er ist im Besitz eines Artefakts, von dem er bereit ist, es an den ersten wahren Anführer weiterzugeben, den er trifft"

felix-schneider: That is not grammatically correct. My original translation also wasn't (second dass should be das), which I have corrected.

files/lang/de.po
felix-schneider: See below

felix-schneider:
Sorry, as it is so often, real life prevented me from working more. I have more or less finished it now.

Here are notes for the second pass:
Notes (part 2):
- Went back and matched the original strings more closely
  - While I don't consider them sacred scripture and can find better translations some of the time, I want to stick more or less close to the OG. This project is in no small part about nostalgia, and maybe someone remembers a particular description fondly.
- This has caused me to revert some translations, but not all of them.
  - Thunder Mace of Dominion - Donnerkolben der Herrschaft (was Donnerstab). It's neither a staff (Stab) nor a mace (Streitkolben), it's a flail (Flegel), but we already have one of those (see below)
  - Giant Flail of Dominion - Riesenflegel der Herrschaft (was Riesenpeitsche). It's technically a Mangual. It's definitely not a whip (Peitsche).
  - True Compass of Mobility - Der Unfehlbare Kompass der Mobilität (was Wahrer Kompass). It's not true because there is a False Compass of Mobility, it's true because it never fails to point north. There is no direct equivalent in German, so it's infallible (unfehlbar).
  - The German translators felt it necessary to end the description for the wizard's hat with an exclamation point. That's cute, so I kept it.
- The two types of mobility boots share a description. Luckily, they do not share this with other mobility items, like the compass, because the description would not fit: The boots are in plural, the compass is not.
- I adjusted some descriptions for (my subjective) clarity.
- The telescope description also has a placeholder for the number of squares, but the text is in singular, so if this is changed, the text would need to be changed anyway.
- In the German OG, each artifact actually has several versions of the name: The full name and a short name. In the hero screen, the tooltip is actually titled with the shortened name, with a subtitle explaining the bonus the artifact gives, but only for artifacts with a simple stat bonus. I quite like that, we should have it be like that. The game does not support this at this time.

files/lang/de.po
```diff
@@ -10901,189 +10860,179 @@ msgstr ""
...
 
 msgid "Wand of Negation"
-msgstr "Stab der Aufhebung"
+msgstr "Der Stab der Aufhebung"
 
-#, fuzzy
 msgid ""
 "The %{name} makes all of the hero's troops immune to dispel magic spells."
-msgstr "Der %{name} macht alle Eure Truppen immun gegen Todeszauber."
+msgstr "%{name} macht Eure Truppen immun gegen Gegenzauber."
```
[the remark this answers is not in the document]
felix-schneider: That probably appears quite often, I will have to check.

felix-schneider:
My suggestions for improvements of the English strings are in my comments above. If I have the time (no promises), I can make a PR for them.

I specifically made sure to cross-reference my translations with the strings from the OG German exe. So deviations from English should be the same as in the OG.

Off the top of my head, I don't remember any changes so significant as to completely change the meaning of the strings, so I don't see the problem here?

We can discuss about specific translations of course.

files/lang/de.po
```diff
@@ -11093,39 +11042,37 @@ msgid ""
...
-"gegen Reiseproviant ein. Sie lässt entfernte Objekte näher erscheinen, und "
-"er nennt sie...\n"
+"gegen Reiseproviant. Es lässt entfernte Objekte näher erscheinen, und er "
+"nennt es...\n"
 "\n"
 "ein Teleskop."
 
 msgid "Statesman's Quill"
-msgstr "Feder des Staatsmannes"
+msgstr "Die Feder des Staatsmanns"
```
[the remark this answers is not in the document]
felix-schneider: "Quill" means a feather, the picture is a feather, the original translation is feather and feathers were medieval writing instruments long before fountain pens (Federhalter). Why change it unnecessarily?

files/lang/de.po
```diff
@@ -11439,60 +11381,59 @@ msgid ""
...
+msgstr "Die Brosche der Abschirmung"
 
 msgid ""
 "The %{name} provides %{count} percent protection from Armageddon and "
 "Elemental Storm, but decreases spell power by 2."
 msgstr ""
-"Der %{name} bietet %{count} Prozent Schutz vor Armageddon und "
-"Elementarsturm, verringert aber die Zauberkraft um 2."
+"%{name} bietet %{count}% Schutz vor Armageddon und Elementarsturm, reduziert "
+"aber Eure Zauberkraft um 2."
```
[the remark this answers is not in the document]
felix-schneider: This was deliberate, as this one artifact description is irregular. Most other stat adjusting items refer to "your <attribute>".
Having a different wording seems only to make sense only of a different implementation applies, which is not the case here.

I guess this is a case where I am also advocating for a change on the English, but I don't see a reason to dogmatically adhere to the English wording (which we do not do in many other cases, such as artifact flavor texts). Consistency in wording is something that each translation should strive for imo.

files/lang/de.po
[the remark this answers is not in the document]
felix-schneider: I lean towards sticking with Feder:
1. It's clear (at least it always was to me) that it was meant as a writing instrument, not something you wear on your hat.
2. The use of Feder, not the more explicit Schreibfeder is quite typical, c.f. "aus der Feder von ..."
3. The OG says Feder

felix-schneider:
Re: you/your hero

For things like this, it would be good to have a localization Bible for each language, where conventions like these can be set down.
It also makes it clear what conventions are agreed upon policy and which are personal preference. I have no problem with one convention over the other in this case, but before I go through and edit all these things, I would like to be sure this is consensus and I will not be asked by the next reviewer to change it all back...

This change in particular would apply to probably almost all artifact descriptions, whether it's "raises your/your hero's Strength", "halves the cost of your/your hero's X spells", "increases your/your hero's movement over land" etc. They almost all refer to the effect they have on "you".

felix-schneider:
> Yes, I'm not saying this has to change in this PR, but at the least we shouldn't add more of them. Replacing them can be a gradual process. As stated in our translation guide, we prefer PRs to stay within a limited amount of changes.

Well, I'm touching all of the artifact descriptions anyway, so I might as well...

felix-schneider:
That went quicker than expected, I changed all the instances of "your xyz" to "your hero's xyz" in the artifact descriptions

felix-schneider:
Sorry for the delay, I hope this is correct.

ihhub (approved):

ihhub:
@felix-schneider , huge thanks for these changes!

PR #9293, opened 2024-11-21
Replace the clang-analyzer workflow with CodeChecker workflow

oleg-derevenetz:
[CodeChecker](https://codechecker.readthedocs.io/en/latest/) is a front-end for multiple analyzers, but in our case it is intended to be used instead of LLVM `scan-build` to run the Clang Analyzer.

Pros:

* It is able to perform the [cross translation unit analysis](https://clang.llvm.org/docs/analyzer/user-docs/CrossTranslationUnit.html) to better find bugs. `scan-build` has an experimental support for CTU at best and is not recommended for this.

Cons:

* CTU analysis is much slower than our `scan-build` analysis (approximately x2 slower).

Currently it found one issue in the LLVM internal header (seems legitimate BTW, but, naturally, there is nothing can be done with this, so I suppressed warnings for this header) and another issue in the implicit (compiler-generated) constructor (apparently copy or move constructor) for our own struct, which seems like a false positive (but probably it is related to that first issue in the LLVM header).

This PR is subject to discussion.


ihhub (approved):

ihhub:

I remove the previous GitHub Action stage requirements to be passed and I will set the new one after the release.

PR #9294, opened 2024-11-23
Add platform "MorphOS"

BeWorld2018:
MorphOS is PPC 32bit bigendian machine (like AmigaOS system)

To build it : make PLATFORM=morphos


oleg-derevenetz (commented):
Hi all,

@BeWorld2018 first of all, thanks for your contribution offer. Personally, I think that the proposed platform is quite exotic and have significant differences from POSIX in terms of filesystem structure, for instance. I'm not sure if it's worth adding support for such exotic platforms to the main codebase, because there will be issues with supporting this code - MorphOS is a proprietary platform for specific hardware, and I suspect that there are very few people who will be able to, say, test new code or changes in the existing code.  The matter is complicated by the fact that, as far as I can see, the proposed changes are not "universal" changes, they are rather "ad hoc" changes.

Alternatively, if MorphOS has its own package manager, I would suggest the author to make and maintain a package (or a separate port) that will contain the necessary patches without integrating them into main codebase.

@ihhub what do you think?

ihhub:
> Hi all,
> 
> @BeWorld2018 first of all, thanks for your contribution efforts. Personally, I think that the proposed platform is quite exotic and have significant differences from POSIX in terms of filesystem structure, for instance. I'm not sure if it's worth adding support for such exotic platforms to the main codebase, because there will be issues with supporting this code - MorphOS is a proprietary platform for specific hardware, and I suspect that there are very few people who will be able to, say, test new code or changes in the existing code. The matter is complicated by the fact that, as far as I can see, the proposed changes are not "universal" changes, they are rather "ad hoc" changes.
> 
> Alternatively, if MorphOS has its own package manager, I would suggest the author to make and maintain a package (or a separate port) that will contain the necessary patches without integrating them into main codebase.
> 
> @ihhub what do you think?

Hi @oleg-derevenetz , we are open to expand the range of supported platforms but as you noted it has to be done in a proper way that someone must maintain the code in a proper shape and up to date to avoid any issues and subsequent blame of the core team if a platform fails to support the engine.

@BeWorld2018 , could you please address the given comments. In order to accept the changes for the new platform we should have approved source code changes. Other things are missing in this pull request:
- lack of prove that the code is indeed compilable on MorphOS platform. Could you please do the research if any of GitHub actions support this platform? If yes how long it takes to compile fheroes2 using the actions?
- we need to expand the documentation to explain how to compile the engine for this platform. Please takes a look at https://github.com/ihhub/fheroes2/blob/master/docs/DEVELOPMENT.md

I would recommend to expand the first message in this pull request to include a comprehensive description as well as all steps mentioned by me as tasks, so it would be easier to address them.

I will review this pull request in a while as well. As of now I am marking it as "Draft".

src/Makefile
```diff
@@ -104,7 +107,12 @@ CPPFLAGS := $(CPPFLAGS) $(CPPFLAGS_FH2)
 ifdef FHEROES2_WITH_SYSTEM_SMACKER
 LIBS := $(LIBS) -lsmacker
 endif
+
+ifeq ($(PLATFORM),morphos)
+LIBS := $(LIBS) -lz_shared $(LDLIBS)
```
ihhub, line 112: Could you please explain why we need to use a shared variant of lz? Can we avoid it? Could you please point to a place where this is the only option available for this platform?

src/Makefile.morphos
```diff
@@ -0,0 +1,43 @@
+###########################################################################
+#   fheroes2: https://github.com/ihhub/fheroes2                           #
+#   Copyright (C) 2022 - 2024                                             #
```
ihhub, line 3: Please include only 2024 year.

src/Makefile.morphos
```diff
@@ -0,0 +1,43 @@
...
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+#   GNU General Public License for more details.                          #
+#                                                                         #
+#   You should have received a copy of the GNU General Public License     #
+#   along with this program; if not, write to the                         #
+#   Free Software Foundation, Inc.,                                       #
+#   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             #
+###########################################################################
+
+AR = ppc-morphos-gcc-ar-11
```
ihhub, line 21: Is it a default gcc version for MorphOS? Can we use the default one?

src/Makefile.morphos
```diff
@@ -0,0 +1,43 @@
...
+CC = ppc-morphos-gcc-11
+CXX = ppc-morphos-g++-11
+# Common flags (for both C and C++ compilers) for both third-party libraries and fheroes2
+AMIGADATE = $(shell date +"%-d.%-m.%Y")
+CCFLAGS := $(CCFLAGS) -noixemul -I/gg/usr/local/include -I/gg/usr/local/include/SDL2 -D__MORPHOS_SHAREDLIBS -D__STDC_LIMIT_MACROS -D__AMIGADATE__=\"$(AMIGADATE)\"
+LDFLAGS := -L/gg/usr/local/lib $(LDFLAGS) 
+# Common flags (for both C and C++ compilers) for fheroes2 only
+CCFLAGS_FH2 := -DTARGET_MORPHOS
+# -DNDEBUG
```
ihhub, line 30: Please remove this line as it is not in use anywhere. We have a specific flag for this matter.

src/Makefile.morphos
```diff
@@ -0,0 +1,43 @@
...
+
+LIBS := $(LIBS) -latomic
+
+ifdef FHEROES2_WITH_IMAGE
+LIBS := $(LIBS) -ljpeg -lwebp
```
ihhub, line 35: Doesn't `shell` command exist for Makefile on MorphOS? Why we do need to include JPEG and WEBJ libraries here?

src/engine/audio.cpp
```diff
@@ -73,7 +73,8 @@ namespace
         // Notice: Value 22050 causes music distortion on Windows.
         int frequency = 44100;
 #endif
-        uint16_t format = AUDIO_S16;
+        uint16_t format = AUDIO_S16SYS;
```
ihhub, line 76: This change might break other platforms. Have you tested it? Can we use the original value? What problems are you encountering by using the original value?

src/engine/serialize.h
```diff
@@ -105,6 +116,10 @@
 
 #include "math_base.h"
 
+#ifdef __MORPHOS__
+#undef tell
```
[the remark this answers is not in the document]
ihhub: Why is it done? o_O

ihhub (changes requested):
Hi @BeWorld2018 , I also put several comment here. Please take a look once you have time.

src/engine/audio.cpp
oleg-derevenetz: Hi @ihhub regarding this, it's more or less clear - our value means LSB byte order, while this value means platform native byte order, and this specific platform is big endian, i.e. has MSB byte order.

src/engine/serialize.h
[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub that's because somewhere inside the system headers of this platform, there is a preprocessor macro with the name `tell`. Macros have no idea about scopes, namespaces or class methods, which will break the build if your class has its own method with the same name, for instance. Of course, it cannot be said that this is a plus for the platform.

oleg-derevenetz:
@ihhub the issue mostly is that fheroes2 expects the target platform to be more or less POSIX-compatible, especially in terms of filesystem structure and file names, and this platform is not POSIX-compatible (at least if it has the same filesystem structure principles as the original Amiga has at the time).

src/Makefile
BeWorld2018: lz_shared use specific shared z library into us SDK. so specific to MorphOS.

BeWorld2018:
> Hi all,
> 
> @BeWorld2018 first of all, thanks for your contribution efforts. Personally, I think that the proposed platform is quite exotic and have significant differences from POSIX in terms of filesystem structure, for instance. I'm not sure if it's worth adding support for such exotic platforms to the main codebase, because there will be issues with supporting this code - MorphOS is a proprietary platform for specific hardware, and I suspect that there are very few people who will be able to, say, test new code or changes in the existing code. The matter is complicated by the fact that, as far as I can see, the proposed changes are not "universal" changes, they are rather "ad hoc" changes.
> 
> Alternatively, if MorphOS has its own package manager, I would suggest the author to make and maintain a package (or a separate port) that will contain the necessary patches without integrating them into main codebase.
> 
> @ihhub what do you think?

Hi, you certaintly right, MorphOS is little platform :-) and exotic
I can continue to port with my exotic code :-) on my git. no problem for me.

src/engine/audio.cpp
BeWorld2018: > Hi @ihhub regarding this, it's more or less clear - our value means LSB byte order, while this value means platform native byte order, and this specific platform is big endian, i.e. has MSB byte order.

You right, use AUDIO_S16SYS is better i think with LE / BE machine.

src/Makefile.morphos
BeWorld2018: worse, i dont build with this flags... FHEROES2_WITH_IMAGE:-) so i dont test this part... need to do.

oleg-derevenetz:
Hi @BeWorld2018 let's try to implement some improvements step-by-step. I just made a few changes to this PR to implement some "generalization" of the filesystem interaction logic to support non-quite-POSIX-compliant filesystems (which exactly correspond to the changes from #9296). Could you pull out these changes of mine in your PR and check if everything is working on your device (loading the original game resources, music playback, saving, loading and removing savefiles, map editor, and so on)?

src/fheroes2/game/fheroes2.cpp
```diff
@@ -289,9 +298,9 @@ int main( int argc, char ** argv )
     try {
         const fheroes2::HardwareInitializer hardwareInitializer;
         Logging::InitLog();
-
+#if !defined( TARGET_MORPHOS )
         COUT( GetCaption() )
-
+#endif
```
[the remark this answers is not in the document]
BeWorld2018: open shell directly with title :-)

src/fheroes2/game/fheroes2.cpp
```diff
@@ -80,6 +80,15 @@
 #include "ui_tool.h"
 #include "zzlib.h"
 
+#if defined( TARGET_MORPHOS )
+#include "version.h"
+#define STRINGIFY( x ) #x
+#define TOSTRING( x ) STRINGIFY( x )
+#define __AMIGAVER__ TOSTRING( MAJOR_VERSION ) "." TOSTRING( MINOR_VERSION ) "." TOSTRING( INTERMEDIATE_VERSION )
+unsigned long __stack = 1024 * 1024;
+static const std::string morphos_versions_tag = "$VER: fheroes2 " __AMIGAVER__ " (" __AMIGADATE__ ")";
```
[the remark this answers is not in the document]
BeWorld2018: Specific MorphOS to put version string into executable

src/Makefile.morphos
BeWorld2018: yes it's a last GCC available. but you right we can use ppc-morphos-gcc-ar (default)

src/engine/system.cpp
```diff
@@ -444,7 +464,11 @@ std::string System::GetDirname( std::string_view path )
 std::string System::GetBasename( std::string_view path )
 {
     if ( path.empty() ) {
+#if defined( TARGET_MORPHOS )
+        return { "PROGDIR:" };
```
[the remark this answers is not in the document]
BeWorld2018: . not working here, but maybe you can use SDL_GetBasePath() ?

PROGDIR:  = current directory in Amiga world
but i can use nothing "", and same effect because SDL2 Lock current directory

src/engine/zzlib.cpp
```diff
@@ -29,6 +29,10 @@
 #include <zconf.h>
 #include <zlib.h>
 
+#ifdef __MORPHOS__
+#undef bind
```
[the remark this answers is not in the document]
BeWorld2018: yeah... not nice :-) releated with zlib.h from system headers (shared library)

src/engine/system.cpp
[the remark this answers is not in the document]
oleg-derevenetz: > . not working here, but maybe you can use SDL_GetBasePath() ?

These are absolutely different things. `SDL_GetBasePath()` returns the directory where the application was run from. `System::GetBasename()` is intended to emulate the POSIX [`basename()`](https://man.archlinux.org/man/basename.3) call on systems that doesn't have it.

> PROGDIR: = current directory in Amiga world

IIRC that's not quite like this, because `PROGDIR:` is the directory where the application was run from, it's not quite the same as the current directory.

Anyway, I suggest to pull and try the latest [changes made by me](https://github.com/ihhub/fheroes2/pull/9294#issuecomment-2496238509) in this PR on your MorphOS device. All these changes are related to the filesystem access.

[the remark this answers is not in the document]
BeWorld2018: yes you right... i test that when i can, thanks

BeWorld2018:
> Hi @BeWorld2018 let's try to implement some improvements step-by-step. I just made a few changes to this PR to implement some "generalization" of the filesystem interaction logic to support non-quite-POSIX-compliant filesystems (which exactly correspond to the changes from #9296). Could you please pull out these changes of mine in your PR and check if everything is working on your device (loading the original game resources, music playback, saving, loading and removing savefiles, map editor, and so on)?

Yeah it's just perfect, just need to exclude MorphOS form GetCaseInsensitibePath, all datas are find with your changes.

oleg-derevenetz:
Hi @BeWorld2018

> Yeah it's just perfect, just need to exclude MorphOS form GetCaseInsensitibePath, all datas are find with your changes.

OK, that's good. The next release will be any day now, after it comes out, we can merge #9296 and #9297 into the `master` branch to minimize the set of changes required to support MorphOS. After that I'll rebase this PR on top of the `master` branch, the changeset in this PR of yours will be reduced, and we'll think about what else we can do.

BeWorld2018:
Nice, thanks you !

src/Makefile.morphos
ihhub: Let's use the default one. We want to avoid forcing people to install any other packages just to compile the engine.

oleg-derevenetz:
OK, the file system is more or less sorted out.

Now the question about fullscreen mode - what exactly is the problem with `SDL_WINDOW_FULLSCREEN_DESKTOP`? Yes, it does not cause video mode switching directly on the graphics card, and instead creates a full-screen borderless window with scaling, but that's exactly what it's supposed to do. This doesn't cause issues with multiple monitors and doesn't cause a mess with other windows or desktop icons due to changing the "real" video mode on the graphics card. What exactly is the problem? It doesn't look good on the screen somehow? Does this cause any specific problems when playing? More detailed explanations (perhaps with screenshots) are welcome :)

oleg-derevenetz:
> On MorphOS this flags SDL_WINDOW_FULLSCREEN_DESKTOP = false fullscreen mean Window take desktop size with borderless (and if you desktop size is 1920x1080... can be slow.. because MorphOS run on G4 / G5 PowerPC... so very old hardware)
and SDL_WINDOW_FULLSCREEN = create new screen with specific size.

So the real issue here is that `SDL_WINDOW_FULLSCREEN_DESKTOP` is slow on high display resolutions? Is this slowdown really being observed?

ihhub:
> No problem here.... On MorphOS this flags SDL_WINDOW_FULLSCREEN_DESKTOP = false fullscreen mean Window take desktop size with borderless (and if you desktop size is 1920x1080... can be slow.. because MorphOS run on G4 / G5 PowerPC... so very old hardware) and SDL_WINDOW_FULLSCREEN = create new screen with specific size.

The engine has FPS counter for this matter. Please enable it and see how it affects. Also, do it only on release build.

src/engine/screen.cpp
```diff
@@ -820,14 +820,14 @@ namespace
...
-#if defined( _WIN32 )
+#if defined( _WIN32 ) || defined( __MORPHOS__ )
                 if ( fheroes2::cursor().isSoftwareEmulation() ) {
                     flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
                 }
                 else {
                     flags |= SDL_WINDOW_FULLSCREEN;
                 }
-#else
+
```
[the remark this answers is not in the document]
ihhub: Hi @BeWorld2018 , please address this comment.

src/fheroes2/game/fheroes2.cpp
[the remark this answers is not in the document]
ihhub: @BeWorld2018 , is it true that logging just opens a shell command?

PR #9296, opened 2024-11-24
Use the std::filesystem and support non-quite-POSIX-compliant filesystems (such as Amiga FS)

oleg-derevenetz:
Related to https://github.com/ihhub/fheroes2/pull/9294

`std::filesystem` is designed to be aware of the target platform rules for working with filesystem paths.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these updates!

PR #9297, opened 2024-11-25
Use the platform-dependent byte order for the SDL audio format

oleg-derevenetz:
Related to #9294


ihhub:

PR #9312, opened 2024-12-03
AI::Planner::reinforceCastle() : paraphrase a few comments

oleg-derevenetz:
Follow-up to the #9283

Use the correct terminology: we don't necessarily need a **free** slot to hire units, a unit can be hired even if there are no free slots, but the same unit is already present in the army.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so these corrections!

PR #9319, opened 2024-12-05
Fix text input cursor setting for multi-line text input with many spaces at the line end

Districh-ru:
This PR fixes the issue with setting input cursor position when there are many spaces in multi-line text input: https://github.com/ihhub/fheroes2/pull/9281#issuecomment-2503107119

Now the multi-line text input field acts like in the original editor - is does not cut spaces at the line end to show user that there are many spaces there and to allow position the cursor at any place of such consequent spaces.

This PR also:
- moves `getTextInputCursorPosition()` to `ui_tool`;
- fixes horizontal position of the multi-font text line separated by `-`
master build:
bug
this PR:
nobug


Districh-ru:
Hi, @ihhub!
I've faced a strange code style check behavior: https://github.com/ihhub/fheroes2/actions/runs/12183834934/job/33986233552?pr=9319
The code was formatted using our `.clang-format` file.
And also the suggested change (https://github.com/ihhub/fheroes2/pull/9319/commits/ed153492cd40cdb63f275bdc3bb8c3689fc36fda) looks not good to me.

oleg-derevenetz:
> I've faced a strange code style check behavior: https://github.com/ihhub/fheroes2/actions/runs/12183834934/job/33986233552?pr=9319 The code was formatted using our `.clang-format` file. And also the suggested change ([ed15349](https://github.com/ihhub/fheroes2/commit/ed153492cd40cdb63f275bdc3bb8c3689fc36fda)) looks not good to me.

This formatting is because `clang-format` treats this as an assignment operation, and there is no knob yet to format assignment operations and pure virtual function declarations in a different way. Please see https://github.com/llvm/llvm-project/issues/57569 for details.

oleg-derevenetz:
Hi @Districh-ru

> I've updated this clang-format to the latest version (making a symlink to LLVM folder, VS does not look in the PATH environment variable folders) and now formatting goes like Code Style Check wants.

There is no need to create symlinks, MSVC can use a custom `clang-format` executable:

clang-format

src/fheroes2/gui/ui_text.cpp
```diff
@@ -706,6 +676,19 @@ namespace fheroes2
...
     }
 
+    void MultiFontText::getTextLineInfos( std::vector<TextLineInfo> & textLineInfos, const int32_t maxWidth, const int32_t rowHeight,
+                                          const bool ignoreSpacesAtTextEnd ) const
+    {
+        const size_t textsCount = _texts.size();
+        for ( size_t i = 0; i < textsCount; ++i ) {
+            const auto langugeSwitcher = getLanguageSwitcher( _texts[i] );
+
+            // To properly join multi-font text we must not ignore spaces at the text end if there is next text.
```
ihhub, line 686: A cosmetic suggestion:
```cpp
// To properly render a multi-font text we must not ignore spaces at the end of a text entry which is not the last one.
const bool isLastTextEntry = ( i == textsCount - 1 );

_texts[i].getTextLineInfos( textLineInfos, maxWidth, rowHeight, isLastTextEntry && ignoreSpacesAtTextEnd );
```

But your version is fine as well.

ihhub (changes requested):
Hi @Districh-ru , I left 2 comments here. Can you please take a look at them?

src/fheroes2/gui/ui_text.h
```diff
@@ -146,15 +164,27 @@ namespace fheroes2
...
         {
             return _language;
         }
 
-    protected:
-        bool _isUniformedVerticalAlignment{ true };
+        // Returns text lines parameters (in pixels) in 'offsets': x - horizontal line shift, y - vertical line shift.
+        // And in 'characterCount' - the number of characters on the line, in 'lineWidth' the width including the `offsetX` value.
+        virtual void getTextLineInfos( std::vector<TextLineInfo> & textLineInfos, const int32_t maxWidth, const int32_t rowHeight,
+                                       const bool ignoreSpacesAtTextEnd = true ) const
```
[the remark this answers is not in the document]
Districh-ru: Hi, @ihhub, I've updated the logic a bit and now `_ignoreSpacesAtTextEnd` is a member of Text class along with `_ignoreSpacesAtLineEnd`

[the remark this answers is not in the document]
ihhub: Hi @Districh-ru , my comment was to still keep the **ignoreSpacesAtTextEnd** input argument but not to make it to have the default value. **ignoreSpacesAtTextEnd** is used only within `getTextLineInfos()` functions so making it as an input argument is the best solution.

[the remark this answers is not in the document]
Districh-ru: Hi, @ihhub, sure, lets keep is as input argument. :)
I just thought to store all "text configuration" data in the Text class.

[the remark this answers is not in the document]
ihhub: We can store an argument as a member of class if it's being used internally. This variable is being used only within this function 🙂

ihhub (approved):

src/fheroes2/gui/ui_text.h
```diff
@@ -229,10 +253,20 @@ namespace fheroes2
             return _fontType;
         }
 
+        void setIgnoreSpacesAtLineEnd( const bool ignoreSpacesAtLineEnd )
```
ihhub, line 256: How about we change this method's name and the corresponding variable into:
```cpp
void keepTrailingSpaces()
{
    _keepTrailingSpaces = true;
}
```

We never use this method with an input argument as `true` in the original form.

src/fheroes2/gui/ui_text.h
```diff
@@ -229,10 +253,20 @@ namespace fheroes2
...
+        {
+            _ignoreSpacesAtLineEnd = ignoreSpacesAtLineEnd;
+        }
+
+        void getTextLineInfos( std::vector<TextLineInfo> & textLineInfos, const int32_t maxWidth, const int32_t rowHeight,
+                               const bool ignoreSpacesAtTextEnd ) const override;
```
ihhub, line 262: Instead of **ignoreSpacesAtTextEnd** I am suggesting to use **ignoreTrailingSpaces** or **keepTrailingSpaces**.

src/fheroes2/gui/ui_text.h
```diff
@@ -217,7 +241,7 @@ namespace fheroes2
 
         // This method modifies the underlying text and ends it with '...' if it is longer than the provided width.
         // By default it ignores spaces at the end of the text phrase.
```
ihhub, line 243: Please update this comment.

src/fheroes2/gui/ui_text.h
```diff
@@ -151,10 +169,16 @@ namespace fheroes2
             return _language;
         }
 
-    protected:
-        bool _isUniformedVerticalAlignment{ true };
+        // Returns text lines parameters (in pixels) in 'offsets': x - horizontal line shift, y - vertical line shift.
+        // And in 'characterCount' - the number of characters on the line, in 'lineWidth' the width including the `offsetX` value.
+        virtual void getTextLineInfos( std::vector<TextLineInfo> & textLineInfos, const int32_t maxWidth, const int32_t rowHeight,
```
ihhub, line 174: I don't think that we need to make this function virtual: the only place where this virtual function is being used in **getTextInputCursorPosition()** function which uses `Text` class. Therefore, this method can be a public method of `Text` class only while `MultiFontText` class should have a different private method.

Unless, we are going to use **getTextLineInfos()** method from `MultiFontText` class somewhere else in the future.

ihhub (changes requested):
Hi @Districh-ru , I took another round of review of this pull request as @Branikolog hasn't reviewed yet. I left few suggestions for improvements. Could you please take a look at them?

src/fheroes2/gui/ui_text.h
Districh-ru: Hi, @ihhub,  sure, but I've separated `_keepLineTrailingSpaces` and `keepTextTrailingSpaces`.
The first one tells the code to keep all spaces at every line end including the end of the whole text.
The second one tells to keep spaces only at the end of the whole text, but ignore spaces at the line end if it is not the last line.

src/fheroes2/gui/ui_text.h
Districh-ru: Ok

src/fheroes2/gui/ui_text.h
Districh-ru: Ow, i missed this.
Done!

src/fheroes2/gui/ui_text.h
Districh-ru: Yes, currently we do not plan to do some extra operations with the multi-font texts.

ihhub (approved):

ihhub:
@Districh-ru , I made the tests myself. Everything works perfect. Thank you!

PR #9320, opened 2024-12-06
Modernize the clang-format config file syntax

oleg-derevenetz:
Starting with `clang-format` 15, these options are not just Boolean options, but have multiple different sub-knobs. The old syntax is not valid anymore.


ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for keeping the checkers in the best shape!

PR #9323, opened 2024-12-07
Limit all resolutions to a currently selected resolution in the system

### LINKED ISSUE #9273: Game resolution on MacOS is a complete mess
### Preliminary checks

### Platform

macOS

### Describe the bug

I compiled the latest version of the engine on macOS Sonoma 14.6 and the resolution selection is obviously does not correspond to the reality. As you can see from the below screenshot 800x600 resolution is not as it should be:
Screenshot 2024-11-16 at 11 02 09 PM

Most of users will immediately try the highest resolution which leads to this:
Screenshot 2024-11-16 at 11 09 07 PM

You could imagine the frustration of players seeing this :)

I am using MacBook Air M3 for tests.

### Save file

No need.

### Additional info

_No response_

ihhub:
The engine should not show resolutions that are higher than the one currently set in the system for any of display. This is needed to avoid players facing a situation when they choose the highest resolution exceeding the current one and they are unable to change it later because the chosen resolution leads to have too small UI elements.

close #9273


oleg-derevenetz (commented):
Hi @ihhub IMHO, in a good way, it would be worth to get the resolution info about the display where the fheroes2 is currently running (using the `SDL_GetWindowDisplayIndex()`) and not just the first display. Regarding macOS, what do you think about choosing the minimum resolution among all displays (but maybe the one that is still not less than 640x480?) and using it as an upper bound? After all, the user can drag the game between the displays, can this lead to switching the display mode on macOS?

ihhub:
Hi @oleg-derevenetz , these are very good points. I will update the code based on your comment

ihhub:
Hi @oleg-derevenetz , could you please take care about the development of the project for the next 3 weeks as I need to take a personal break and I will not have access to any of my machines expect my phone?

oleg-derevenetz:
Hi @ihhub

> could you please take care about the development of the project for the next 3 weeks as I need to take a personal break and I will not have access to any of my machines expect my phone?

Going on vacation? :) Yes, well, I'll do my best to keep an eye on the project.

ihhub:
> Hi @ihhub
> 
> > could you please take care about the development of the project for the next 3 weeks as I need to take a personal break and I will not have access to any of my machines expect my phone?
> 
> Going on vacation? :) Yes, well, I'll do my best to keep an eye on the project.

Yes, my longest vacation in the last 10+ years :) I will still be active tomorrow and come back to the project on 15th April.

Thank you so much!

src/engine/screen.cpp
```diff
@@ -901,43 +901,86 @@ namespace
...
+            if ( displayCount < 1 ) {
                 return std::vector<fheroes2::ResolutionInfo>{ resolutionSet.rbegin(), resolutionSet.rend() };
-            }();
+            }
 
-            return filteredResolutions;
+            // We should limit all available resolutions to the one which is currently chosen on the system
+            // to avoid ending up having application window which is bigger than the screen resolution.
+            SDL_DisplayMode maxDisplayMode;
+            memset( &maxDisplayMode, 0, sizeof( SDL_DisplayMode ) );
```
oleg-derevenetz: We can use just this instead:

```suggestion
            SDL_DisplayMode maxDisplayMode{};
```

oleg-derevenetz (approved):
Hi @ihhub Looks good to me. I left just one optional suggestion, could you please take a look at your spare time?

PR #9325, opened 2024-12-07
Be specific in GitHub Actions about what we are building

ihhub:
A very tiny detail since we don't build any other app bundle.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -175,7 +175,7 @@ jobs:
       run: |
         make -j 2
       env: ${{ matrix.config.env }}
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

oleg-derevenetz (changes requested):
Hi @ihhub I suppose we can close this in favor of #9328, because there will be no need in separate step to create a bundle.

ihhub:
> Hi @ihhub I suppose we can close this in favor of #9328, because there will be no need in separate step to create a bundle.

Sure.

PR #9326, opened 2024-12-07
Fix a bug when picking up an object whose type does not match its sprite on the map

### LINKED ISSUE #9184: When picking up a floating chest from land, the game crashes
### Preliminary checks

### Platform

Windows

### Describe the bug

In more than one game, when pickup up a floating xp/gold chest, the game crashes.

### Save file

[save file](https://drive.google.com/file/d/11C11pEADActNIZJqodtgi7an61UYkpi2/view?usp=sharing)
Captură de ecran 2024-10-12 120156

Screen 2024-10-12 120044

### Additional info

_No response_

### LINKED ISSUE #9322: Crash on Assertion `objectUID != 0' failed.
### Preliminary checks

### Platform

Linux

### Describe the bug

fheroes2: ../fheroes2/heroes/heroes_action.cpp:241: void {anonymous}::runActionObjectFadeOutAnumation(const Maps::Tile&, MP2::MapObjectType): Assertion `objectUID != 0' failed.

To reproduce, just get all items in the middle of the ocean and you will see the crash

### Save file

[Alteris2_0006_bug_found.sav.zip](https://github.com/user-attachments/files/18044060/Alteris2_0006_bug_found.sav.zip)

### Additional info

_No response_

Districh-ru:
Fix  #9322 and fix #9184

This PR fixes such chest objects during map load and take into account a possible case when object type does not match the ICN while performing the fade-out animation.

~It also has some lines updated by the new clang-format. :)~

### DIFF
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -225,12 +225,13 @@ namespace
     {
         uint32_t objectUID = 0;
 
-        // There are hacked maps (or made with the original editor's bugs) that have not corresponding object type ant its sprite.
-        // So we also search object ID by checking if the object is action because on one tile there might be only one action object.
+        // There are original hacked maps (or made with exploitation of the original Editor's bugs) that have wrong object type in relation to its image sprite.
+        // So, we also search for object ID by checking if the object is an action type because one tile can have only one action object.
         MP2::MapObjectType actualObjectType = MP2::OBJ_NONE;
+        const bool isActionObject = MP2::isInGameActionObject( objectType );
 
         if ( const MP2::MapObjectType mainObjectType = Maps::getObjectTypeByIcn( tile.getMainObjectPart().icnType, tile.getMainObjectPart().icnIndex );
-             mainObjectType == objectType || MP2::isInGameActionObject( mainObjectType ) ) {
+             mainObjectType == objectType || ( isActionObject && MP2::isInGameActionObject( mainObjectType ) ) ) {
             objectUID = tile.getMainObjectPart()._uid;
 
             actualObjectType = mainObjectType;
@@ -239,7 +240,7 @@ namespace
             // In maps made by the original map editor the action object can be in the ground layer.
             for ( auto iter = tile.getGroundObjectParts().rbegin(); iter != tile.getGroundObjectParts().rend(); ++iter ) {
                 if ( const MP2::MapObjectType partObjectType = Maps::getObjectTypeByIcn( iter->icnType, iter->icnIndex );
-                     partObjectType == objectType || MP2::isInGameActionObject( partObjectType ) ) {
+                     partObjectType == objectType || ( isActionObject && MP2::isInGameActionObject( partObjectType ) ) ) {
                     objectUID = iter->_uid;
 
                     actualObjectType = partObjectType;
@@ -254,7 +255,7 @@ namespace
         I.getGameArea().runSingleObjectAnimation( std::make_shared<Interface::ObjectFadingOutInfo>( objectUID, tile.GetIndex(), objectType ) );
 
         // If there is a map bug the Object Animation destructor is not able to properly remove this object from the map.
-        if ( actualObjectType != objectType ) {
+        if ( actualObjectType != objectType && actualObjectType != MP2::OBJ_NONE ) {
             // Remove an object by its actual sprite type.
             removeObjectFromTileByType( tile, actualObjectType );
         }
--- a/src/fheroes2/heroes/heroes_action.cpp
+++ b/src/fheroes2/heroes/heroes_action.cpp
@@ -221,18 +221,28 @@ namespace
         I.setRedraw( Interface::REDRAW_RADAR );
     }
 
-    void runActionObjectFadeOutAnumation( const Maps::Tile & tile, const MP2::MapObjectType objectType )
+    void runActionObjectFadeOutAnimation( const Maps::Tile & tile, const MP2::MapObjectType objectType )
     {
         uint32_t objectUID = 0;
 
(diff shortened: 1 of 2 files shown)


Districh-ru:
Hi, @oleg-derevenetz, for your information:
I've (automatically) formatted the code with the new clang-format (v.19.1.5) but its formatting is not good for our Code style check: https://github.com/ihhub/fheroes2/actions/runs/12213646529/job/34073492599?pr=9326  :(

So I've additionally installed LLVM v.17 to match out Code style checker clang-format version.
(For the original clang-format v.12 it is needed to revert the latest changes in the `.clang-format` file.)

oleg-derevenetz:
Well, yes, different versions of `clang-format` may have a bit different formatting logic, so it's better to use the same version of `clang-format` on both sides.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -221,18 +221,28 @@ namespace
...
     {
         uint32_t objectUID = 0;
 
-        if ( Maps::getObjectTypeByIcn( tile.getMainObjectPart().icnType, tile.getMainObjectPart().icnIndex ) == objectType ) {
+        // There are hacked maps (or made with the original editor's bugs) that have not corresponding object type ant its sprite.
+        // So we also search object ID by checking if the object is action because on one tile there might be only one action object.
+        MP2::MapObjectType actualObjectType = MP2::OBJ_NONE;
+
+        if ( const MP2::MapObjectType mainObjectType = Maps::getObjectTypeByIcn( tile.getMainObjectPart().icnType, tile.getMainObjectPart().icnIndex );
+             mainObjectType == objectType || MP2::isInGameActionObject( mainObjectType ) ) {
```
ihhub, line 233: Is there a need to have `MP2::isInGameActionObject( mainObjectType )` check?

The original idea is to expand the functionality of this class to support removal of any object, not only an action type.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -243,6 +253,12 @@ namespace
         Interface::AdventureMap & I = Interface::AdventureMap::Get();
         I.getGameArea().runSingleObjectAnimation( std::make_shared<Interface::ObjectFadingOutInfo>( objectUID, tile.GetIndex(), objectType ) );
 
+        // If there is a map bug the Object Animation destructor is not able to properly remove this object from the map.
+        if ( actualObjectType != objectType ) {
+            // Remove an object by its actual sprite type.
+            removeObjectFromTileByType( tile, actualObjectType );
```
ihhub, line 259: Wouldn't we have an assertion while passing `actualObjectType` as **MP2::OBJ_NONE** is no actual object is found at the tile?

src/fheroes2/maps/maps_tiles_helper.cpp
```diff
@@ -2260,7 +2260,20 @@ namespace Maps
             assert( isFirstLoad );
 
             if ( tile.isWater() ) {
-                tile.setMainObjectType( MP2::OBJ_SEA_CHEST );
+                // On original map "Alteris 2" there is a treasure chest placed on the water and there might be other maps with such bug.
+                // If there is a bug then remove of the MP2::OBJ_TREASURE_CHEST will return 'true' and we can replace it with a Sea Chest object.
+                if ( removeObjectFromTileByType( tile, MP2::OBJ_TREASURE_CHEST ) ) {
+                    // WARNING. The 'seaChestObjectId' must match the Sea Chest object info position in 'Maps::ObjectGroup::ADVENTURE_WATER' group.
+                    const int32_t seaChestObjectId = 1;
+                    const auto & objectInfo = Maps::getObjectInfo( Maps::ObjectGroup::ADVENTURE_WATER, seaChestObjectId );
```
ihhub, line 2268: To avoid this we can add the following code:
```cpp
static int32_t seaChestObjectId = -1;
if ( seaChestObjectId < 0 ) {
    const auto & objects = Maps::getObjectsByGroup( Maps::ObjectGroup::ADVENTURE_WATER );
    for ( size_t i = 0; i < objects.size(); ++i ) {
        if ( objects[i].objectType == MP2::OBJ_SEA_CHEST ) {
            seaChestObjectId = static_cast<int32_t>( i );
            break;
        }
    }

    assert( seaChestObjectId >= 0 );
}
```

What do you think?

ihhub (changes requested):
Hi @Districh-ru , I left several comments in this pull request. Would you mind please to take a look once you have spare time?

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: Hi, @ihhub, yes, because if the map is hacked we check if the object is active. Currently the function is called `runActionObjectFadeOutAnimation` so it is meant that the object should be an action one. :)
And this change is described in the upper comment. Currently it is a way to fix such map hacks.

I've updated the code to support also non-action objects by adding `const bool isActionObject = MP2::isInGameActionObject( objectType );` and then checking `... || ( isActionObject && MP2::isInGameActionObject( mainObjectType ) )`.

src/fheroes2/heroes/heroes_action.cpp
Districh-ru: If it is so then we will break the previous assertion `assert( objectUID != 0 );`. :)
But lets be more safe - I've added `&& actualObjectType != MP2::OBJ_NONE`.

src/fheroes2/maps/maps_tiles_helper.cpp
Districh-ru: Sure!
I just wanted to make a faster code, but lets be more safe)
I've slightly modified your suggestion to place a Sea Chest object only if it is found and to do nothing otherwise.

oleg-derevenetz:
I suppose this PR should close #9184 as well.

Districh-ru:
@oleg-derevenetz, yes, you are absolutely right!

ihhub (approved):

ihhub:

PR #9328, opened 2024-12-07
Use a separate directory for binaries when building using GNU Make

### LINKED ISSUE #9324: Compilation of engine component using gcc leaves all objects files in the same directory
### Preliminary checks

While making changes in engine component of the project you can notice that object files are located in the same directory as the source files:
image

This is true using makefile approach.

We should place object files in a separate place not to mess with source files.

### Additional info

_No response_

oleg-derevenetz:
close #9324

* When using GNU Make, all object files, libraries, and binaries (including tools) will be created in `src/dist`;
* There will be no need to run `make bundle` to create macOS app bundle, just `make` is enough;
* Various minor improvements in Makefiles.


ihhub (approved):

ihhub:
@oleg-derevenetz , awesome! Thank you so much for these changes!

Issue #9331, opened 2024-12-09
No rule to make target '../engine/libengine.a'


Stiivi:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

macOS

### Describe the bug

When building 2804353ac3520742bcc04f6644b2965101de3768 on MacOS the `make` command fails with:

```
make[2]: *** No rule to make target '../engine/libengine.a', needed by 'fheroes2'.  Stop.
```

Did fresh pull and `make clean`.

Some possible pointers: the `libengine.a` is in `src/engine` not in `src/dist/engine`, same for `libsmacker.a`.

### Save file

(no save file needed)

oleg-derevenetz:
Well, it builds just fine both on our CI and on my macOS for instance. Regarding object files and libs in `src` and not in `src/dist`, these are leftovers from the previous build logic and should be removed manually, because there are no Makefiles anymore that will remove them automatically. I have an impression that this issue is related to `VPATH` logic and that the proper manual clean-up should fix it.

Stiivi:
Refreshing `src/` (removing it recursively and checking it out again) helped. Cheers.

PR #9332, opened 2024-12-10
Limit the use of VPATH for GNU Make

oleg-derevenetz:
Follow-up to #9328
Related to #9331

#9331 has occurred due to `VPATH` logic. When `VPATH` is used, all prerequisites which are not found at their paths are also looked at paths listed in `VPATH`. For instance, if we have a `Makefile` like this:

```make
VPATH := ../src/engine

all: libengine.a

[...]
```

then, if `libengine.a` is not located in the current directory, it will be looked in the `../src/engine` directory, and, if it's here (left from the previous build) and is up-to-date, it will not be rebuilt and will not be created at the current directory.

With this PR, I minimized the use of `VPATH` just to the `fheroes2` directory (there are no other options here because there are many subdirectories). For the rest of the cases, I have specified the paths to dependencies explicitly. I also removed variables that were used only once.

### DIFF
--- a/src/dist/engine/Makefile
+++ b/src/dist/engine/Makefile
@@ -22,20 +22,16 @@ ifndef FHEROES2_WITH_SYSTEM_SMACKER
 CCFLAGS := $(CCFLAGS) -I../../thirdparty/libsmacker
 endif
 
-SOURCEROOT := ../../engine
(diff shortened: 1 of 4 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for making the compilation even better!

PR #9343, opened 2024-12-14
Fix spell randomization logic for Artifacts and the Adventure Map objects

ihhub:
The original game does not have logic where it chooses whether a spell should be Combat or Adventure and only then does randomization based on the chosen category. Instead, it simply picks a random spell based on a level.

How it was discovered: I put 30 Shrine of the First Circle on a map and visited all of them by a hero. I counted how many spells belong to Adventure category while playing the map in the original game. It was only 3 spells out of 30 instead of expected ~15 per our old logic.

The old logic in the engine also explains why the Dimension Door spell was so often in Pyramids as the chance of getting this spell was wooping 25%.

### DIFF
--- a/src/fheroes2/spell/spell.cpp
+++ b/src/fheroes2/spell/spell.cpp
@@ -511,10 +511,9 @@ uint32_t Spell::weightForRace( const int race ) const
     return 10;
 }
 
-Spell Spell::Rand( const int level, const bool isAdventure )
+Spell Spell::getRandomSpell( const uint8_t level )
 {
-    std::vector<Spell> vec;
-    vec.reserve( 15 );
+    std::vector<Spell> validSpells;
 
     for ( int32_t spellId = NONE; spellId < PETRIFY; ++spellId ) {
         const Spell spell( spellId );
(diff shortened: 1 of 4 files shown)


src/fheroes2/spell/spell.cpp
```diff
@@ -511,10 +511,9 @@ uint32_t Spell::weightForRace( const int race ) const
     return 10;
 }
 
-Spell Spell::Rand( const int level, const bool isAdventure )
+Spell Spell::getRandomSpell( const uint8_t level )
 {
-    std::vector<Spell> vec;
-    vec.reserve( 15 );
+    std::vector<Spell> validSpells;
```
oleg-derevenetz, line 516: What do you think about reserving some space here - for `SPELL_COUNT` spells, for instance?

src/fheroes2/spell/spell.cpp
```diff
@@ -511,10 +511,9 @@ uint32_t Spell::weightForRace( const int race ) const
     return 10;
 }
 
-Spell Spell::Rand( const int level, const bool isAdventure )
+Spell Spell::getRandomSpell( const uint8_t level )
```
oleg-derevenetz, line 514: Why it's `uint8_t`? `Spell::Level()` returns just `int`, for instance.

oleg-derevenetz (commented):

src/fheroes2/spell/spell.cpp
ihhub: Sure, we can do this.

src/fheroes2/spell/spell.cpp
ihhub: I think it's a left over from my first idea. I will update this.

PR #9335, opened 2024-12-10
Add a new variant of Stone Liths object

ihhub:
All credits go to **PushPop** from Discord:

https://github.com/user-attachments/assets/10b18ce3-9974-4864-b2af-a441e82daa7e

PR #9344, opened 2024-12-15
Add ability to set the amount in Resource object on the Adventure Map

ihhub:
This is implemented only for Resource objects. Random resource object does **not** support editing as it is not obvious what to do with Gold resource plus random should really mean random.

image

This pull request must be merged immediately after #9335 to avoid map format version bumping.


Alucard648:
Gold is multiplied by 100 in H3 random resource object. So if you set amount of resource to 10, you may get euther 10 resources, for instance, 10 Mercury, or 1000 gold.

ihhub:
> Gold is multiplied by 100 in H3 random resource object. So if you set amount of resource to 10, you may get euther 10 resources, for instance, 10 Mercury, or 1000 gold.

Let's not repeat a non-trivial logic from HoMM3. This is not obvious for map makers as it seems.

ihhub:
@Alucard648 , HoMM4 Editor does not allow to edit random resources which makes more sense since (100 multiplier for Gold is not trivial as it seems):

image

src/fheroes2/editor/editor_object_popup_window.cpp
```diff
@@ -51,10 +53,19 @@ namespace
 
             return Maps::Ground::String( tile.GetGround() );
         case MP2::OBJ_RESOURCE: {
-            const Funds funds = getFundsFromTile( tile );
-            assert( funds.GetValidItemsCount() == 1 );
+            const auto & objects = Maps::getObjectsByGroup( Maps::ObjectGroup::ADVENTURE_TREASURES );
+            for ( const auto & info : objects ) {
```
oleg-derevenetz, line 57: Is the `objects` variable a bit redundant here?

```suggestion
            for ( const auto & info : Maps::getObjectsByGroup( Maps::ObjectGroup::ADVENTURE_TREASURES ) ) {
```

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -1012,6 +1012,23 @@ bool World::loadResurrectionMap( const std::string & filename )
                     tileData[0] = tileData[0] - 1U;
                 }
             }
+            else if ( map.version >= 7 && object.group == Maps::ObjectGroup::ADVENTURE_TREASURES ) {
```
oleg-derevenetz, line 1015: Is there any way not to use the version binding here and somehow manage the conversion when loading the map? IIRC maps are always "upconverted" to the latest supported format upon loading, right?

oleg-derevenetz (commented):
Hi @ihhub I just left one optional suggestion and one question, could you please take a look when you have time?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1290,6 +1291,32 @@ namespace Interface
                         action.commit();
                     }
                 }
+                else if ( objectInfo.objectType == MP2::OBJ_RESOURCE ) {
```
ihhub, line 1294: We do not delete metadata while deleting an object.

src/fheroes2/world/world_loadmap.cpp
ihhub: Yeah. I was thinking about this for a very long time. Checking version doesn't solve many all cases. I will drop the check.

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -1012,6 +1012,23 @@ bool World::loadResurrectionMap( const std::string & filename )
...
+                // Resource amount setup was implemented in map format version 7.
+                const auto & objectInfo = Maps::getObjectInfo( object.group, static_cast<int32_t>( object.index ) );
```
ihhub, line 1017: Cache Adventure Treasure objects like it is done under `// Read and populate objects.` comment.

src/fheroes2/maps/map_format_info.cpp
```diff
@@ -72,7 +72,7 @@ namespace
     constexpr uint16_t minimumSupportedVersion{ 2 };
 
     // Change the version when there is a need to expand map format functionality.
-    constexpr uint16_t currentSupportedVersion{ 6 };
+    constexpr uint16_t currentSupportedVersion{ 7 };
```
ihhub, line 75: Revert this change as we are not going to check it.

ihhub:
I swapped @zenseii with @Branikolog as @zenseii is very busy with other work at the moment.

ihhub:
Hi @oleg-derevenetz and @Districh-ru , this pull request is ready for your review.

oleg-derevenetz (approved):

Alucard648:
> @Alucard648 , HoMM4 Editor does not allow to edit random resources which makes more sense (since 100 multiplier for Gold is not trivial as it seems):
Heroes 3 allowed setting resource amount for piles. Check H5+ map editors.

Districh-ru (approved):

ihhub:
Hi @Branikolog , I've merged this pull request as we need to move forward with the release changes. Please use the master branch to test changes.

PR #9341, opened 2024-12-12
Document the mechanics of "drawing the angle of attack" in the README_android.md

oleg-derevenetz:
Related to #8464

We forgot to document this handy touchscreen mechanic anywhere. The wording can be adjusted if there are appropriate proposals.

PR #9346, opened 2024-12-15
Display a description of the mechanics of right-clicking on devices with touch input

oleg-derevenetz:
Related to https://github.com/ihhub/fheroes2/pull/9341#pullrequestreview-2504535675.

I would not like to make another dialog box at the first start of the game (we already have three (or even four) of them: "Select Game Language", "Greetings!", "Select Game Resolution" (not on all devices) and, finally, "Please Remember"), so I added this information to the "Greetings!" dialog box. This description will only be displayed if the user device has touch input support.

With this PR:

greetings

### DIFF
--- a/src/fheroes2/game/game_mainmenu.cpp
+++ b/src/fheroes2/game/game_mainmenu.cpp
@@ -232,40 +232,48 @@ fheroes2::GameMode Game::MainMenu( const bool isFirstGameRun )
 
     fheroes2::Display & display = fheroes2::Display::instance();
 
-    // image background
     fheroes2::drawMainMenuScreen();
+
     if ( isFirstGameRun ) {
         // Fade in Main Menu image before showing messages. This also resets the "need fade" state to have no fade-in after these messages.
         fheroes2::validateFadeInAndRender();
 
         fheroes2::selectLanguage( fheroes2::getSupportedLanguages(), fheroes2::getLanguageFromAbbreviation( conf.getGameLanguage() ), true );
 
-        if ( System::isHandheldDevice() ) {
+        {
+            std::string body = _( "Welcome to Heroes of Might and Magic II powered by fheroes2 engine!" );
+
+            if ( System::isTouchInputAvailable() ) {
+                body += _(
+                    "\n\nTo simulate a right-click with a touch to get info on various items, you need to first touch and keep touching on the item of interest and then touch anywhere else on the screen. You can then remove your first finger from the screen and keep viewing the info on the item." );
+            }
+
             // Handheld devices should use the minimal game's resolution. Users on handheld devices aren't asked to choose resolution.
-            fheroes2::showStandardTextMessage( _( "Greetings!" ), _( "Welcome to Heroes of Might and Magic II powered by fheroes2 engine!" ), Dialog::OK );
-        }
-        else {
-            fheroes2::showStandardTextMessage(
-                _( "Greetings!" ),
-                _( "Welcome to Heroes of Might and Magic II powered by the fheroes2 engine!\nBefore starting the game, please select a game resolution." ), Dialog::OK );
-            const bool isResolutionChanged = Dialog::SelectResolution();
-            if ( isResolutionChanged ) {
+            if ( !System::isHandheldDevice() ) {
+                body += _( "\n\nBefore starting the game, please select a game resolution." );
+            }
+
+            fheroes2::showStandardTextMessage( _( "Greetings!" ), body, Dialog::OK );
+
+            if ( !System::isHandheldDevice() && Dialog::SelectResolution() ) {
                 fheroes2::drawMainMenuScreen();
             }
         }
 
-        fheroes2::Text header( _( "Please Remember" ), fheroes2::FontType::normalYellow() );
+        {
+            fheroes2::Text header( _( "Please Remember" ), fheroes2::FontType::normalYellow() );
 
-        fheroes2::MultiFontText body;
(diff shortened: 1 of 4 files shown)


oleg-derevenetz:
BTW, what do you think about removing the "Please Remember" dialog box? At the time, the only way to open the Settings dialog was clicking on the door, which wasn't intuitive, hence the need for the "Please Remember" dialog box. But now we have the self-explanatory "Settings" button in the main menu...

src/fheroes2/game/game_mainmenu.cpp
```diff
@@ -232,40 +232,48 @@ fheroes2::GameMode Game::MainMenu( const bool isFirstGameRun )
...
-            fheroes2::showStandardTextMessage(
-                _( "Greetings!" ),
-                _( "Welcome to Heroes of Might and Magic II powered by the fheroes2 engine!\nBefore starting the game, please select a game resolution." ), Dialog::OK );
-            const bool isResolutionChanged = Dialog::SelectResolution();
-            if ( isResolutionChanged ) {
+            if ( !System::isHandheldDevice() ) {
+                body += _( "\n\nBefore starting the game, please select a game resolution." );
+            }
+
+            fheroes2::showStandardTextMessage( _( "Greetings!" ), body, Dialog::OK );
```
Districh-ru, line 256: We can move the `body` string not to make its copy.
```suggestion
            fheroes2::showStandardTextMessage( _( "Greetings!" ), std::move( body ), Dialog::OK );
```

src/fheroes2/game/game_mainmenu.cpp
```diff
@@ -232,40 +232,48 @@ fheroes2::GameMode Game::MainMenu( const bool isFirstGameRun )
...
 
         fheroes2::selectLanguage( fheroes2::getSupportedLanguages(), fheroes2::getLanguageFromAbbreviation( conf.getGameLanguage() ), true );
 
-        if ( System::isHandheldDevice() ) {
+        {
+            std::string body( _( "Welcome to Heroes of Might and Magic II powered by fheroes2 engine!" ) );
+
+            if ( System::isTouchInputAvailable() ) {
+                body += _(
+                    "\n\nTo simulate a right-click with a touch to get info on various items, you need to first touch and keep touching on the item of interest and then touch anywhere else on the screen. You can then remove your first finger from the screen and keep viewing the info on the item." );
```
Districh-ru, line 248: How about making the last sentence more concrete: `While holding the second finger you can remove the first one from the screen and keep viewing the information on the item.` ?

Districh-ru (changes requested):
Hi, @oleg-derevenetz, I left a couple of proposals. Could you please take a look when you have time?

src/fheroes2/game/game_mainmenu.cpp
oleg-derevenetz: Sure. For some reason, I thought that it takes it parameters by const reference, just like `fheroes2::showMessage()`, but it really turned out that this was not the case.

src/fheroes2/game/game_mainmenu.cpp
oleg-derevenetz: This phrase has been taken from the `README_android.md`, so we need to sync them in this case.

Districh-ru (approved):

Branikolog:
Hi, @oleg-derevenetz !
I agree to take away "please remember" dialog noting the settings too. It's pretty uselss.
But I think including the mechanics desription into the dialog named "greetings" is controversal. Players can just skip it. Also, there's too much text in a single dialog which could be another reason people to skip it instantly. :)
I personaly would prefer converting "Please remember" dialog into the new one with mechanics description, leaving same 3 dialogs in the beginning as it was implemented previously.
P.S. Unfortunately I couldn't download pre-built archives as there were only a few avaialbe (ubuntu and macos)

oleg-derevenetz:
Hi @Branikolog

> But I think including the mechanics desription into the dialog named "greetings" is controversal. Players can just skip it. Also, there's too much text in a single dialog which could be another reason people to skip it instantly. :)

Then why we show the "Greetings!" dialog at all if people just skip it without reading? And if people skip dialog windows, when they see a long sentence, then there is no point in showing these dialogs at all, let these people continue to suffer in regard to the right mouse button on the touchscreen.

> Unfortunately I couldn't download pre-built archives as there were only a few avaialbe (ubuntu and macos)

You should press on the "Latest #2" combo box at the upper right corner and select "Attempt #1" instead.

ihhub (approved):

Branikolog:
@oleg-derevenetz 
I just shared my experience after communicating with many people using the engine. :) Sometimes players literally instant-skip the dialogs, especially the first one with basic header "greetings", being charmed by the aesthetics and the music of the game. After that they could even look for an option to change resolution or the language back in panic.
But I 100% agree with you, it's 100% their fault. 😅 
Since we've taken away one dialog, I just proposed to show right-click description into a separate dialog with a warning header to catch players attention more on this substantial game interaction method.
I'm actually OK with mentioning it anywhere, but I just think it can look better in a way I suggested.
If everyone is fine with current solution I'm ready to approve the PR. 👍

Branikolog:
@oleg-derevenetz 
I'm personally OK with such greeting dialog. On the first start it looks relevant for me. I was bother about the fact, that together with introducing the engine in "greetings" dialog we start describing some game mechanics. Such description is more relevant for the dialog named like: "Please remember", "Please note" etc.
But it is just a matter of taste. If you are OK with that I'm happy with right click description noted anywhere in the GUI. :)

Branikolog (approved):

ihhub:

PR #9356, opened 2024-12-19
Editor: Reset the "tile under cursor" value, stored from a previously edited map

### LINKED ISSUE #9353: Assertion on loading the map after saved another one
### Preliminary checks

### Platform

Windows

### Describe the bug

I'm not sure, if this the problem with map versioning, but after I **save** changes in one particular map I cannot load some other maps.
If I just save map with no changes - no assertion.
If I just save changes and reload the same map - no assertion.
Some maps could be loaded with no assertion, but some fail.
Sometimes I make terrain changes and loading is fine. But in some other cases - assertion occurs.
Here's the way it could be reproduced in the most cases.

https://github.com/user-attachments/assets/4b960e3e-151a-4b63-a0d1-51683d41af2f

Edit: the assertion seem to occur more frequently, when I use only keyboard for navigating in the UI.

### Save file

[Assertoin maps.zip](https://github.com/user-attachments/files/18196948/Assertoin.maps.zip)

### Additional info

_No response_

Districh-ru:
Fix #9353
This PR force reset two cursor indicator parameters when loading or making a new map in editor.
Previously if you hold a cursor over game area when it has square cursor marker and press L to load a new map (or N to make a new one) the `_tileUnderCursor` index is stored and if a loaded/new map is smaller than this index you'll get an assertion because for the first frame it tries to render this marker beyond the map borders.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -783,6 +783,10 @@ namespace Interface
 
         _gameArea.SetUpdateCursor();
 
+        // The cursor parameters may contain values from a previously edited map that are not suitable for this one. Reset them.
+        _tileUnderCursor = -1;
+        _areaSelectionStartTileId = -1;
+
         uint32_t redrawFlags = REDRAW_GAMEAREA | REDRAW_RADAR | REDRAW_PANEL | REDRAW_STATUS | REDRAW_BORDER;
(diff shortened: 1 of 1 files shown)


Branikolog (approved):
Well done!

ihhub (approved):

ihhub:
@Districh-ru , many thanks for this fix!

PR #9357, opened 2024-12-19
Fix the mouse interaction in the Editor's spell selection and castle properties windows

### LINKED ISSUE #9354: Spell selection dialog, spell info is triggering by hovering over the icon with the right mouse button pressed.
### Preliminary checks

### Platform

Windows

### Describe the bug

Press right mouse button and move cursor over any spell icon:

https://github.com/user-attachments/assets/ea097c04-a513-4b90-81cc-3669c7c26d13

### Save file

None.

### Additional info

_No response_

oleg-derevenetz:
fix #9354


Branikolog:
Hi, @oleg-derevenetz !
I've checked the PR and the issue for shrines and pyramids is fixed. 👍 
I've noticed the same issue for castle window. Should I create a separate issue for that or these are similar problems and could be fixed within a single PR?
image

src/fheroes2/editor/editor_spell_selection.cpp
```diff
@@ -138,14 +138,17 @@ namespace
                 return false;
             }
 
-            if ( eventProcessor.MouseClickLeft() ) {
-                assert( static_cast<size_t>( spellIndex ) < _spellRoi.size() );
+            assert( static_cast<size_t>( spellIndex ) < _spellRoi.size() );
 
+            const fheroes2::Rect spellRoi = _spellRoi[spellIndex];
```
Districh-ru, line 143: Hi, @oleg-derevenetz, should we convert this to `const &` not to make a copy?
```suggestion
            const fheroes2::Rect & spellRoi = _spellRoi[spellIndex];
```

oleg-derevenetz: @Districh-ru we can do it, but I'm not sure that we should. `Rect` is 4 32-bit ints, which rougly correspond to 2 64-bit pointers, it's not a some structure of a few hundreds or thousands bytes in size, so I doubt that we will win significantly here in terms of performance. Also I prefer not to take a references to items of externally-managed vectors without any particular need because they can easily become dangling if reallocation of this vector will be caused somewhere between.

Districh-ru: @oleg-derevenetz, thanks for the explanation!
I'll take this into account.

oleg-derevenetz:
> Should I create a separate issue for that or these are similar problems and could be fixed within a single PR?

I (hopefully) fixed the castle properties window as well.

Districh-ru (approved):

ihhub (approved):

ihhub:

PR #9359, opened 2024-12-20
Fix the Linux ARM64 build

oleg-derevenetz:
An alternative to #9360.

`ubuntu-24.04` is now `ubuntu-latest`, and all this voodoo magic that was used to install ARM64 packages on the Ubuntu GitHub image no longer works, because `apt` now uses new configs of a different format, located in different places.

This PR fixes the build by downgrading the `ubuntu-latest` to `ubuntu-22.04` for all ARM64 targets.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -90,7 +90,7 @@ jobs:
           release_name: Ubuntu ARM64 (Linux) build with SDL2 (latest commit)
           release_tag: fheroes2-linux-arm-sdl2_dev
         - name: Linux ARM64 SDL2 Debug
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @oleg-derevenetz , this platform is one of the least downloaded versions on a par with Switch and Vita:
image

The platform occupies around 1.5% of total downloads. You can see statistics here: https://githubstats0.firebaseapp.com/

I suggest to keep this solution for now for the upcoming release as more and more devices use ARM processors and to support this platform we don't need to do extra steps (except what we've done so far).

ihhub (approved):

PR #9360, opened 2024-12-20
Remove the Linux ARM64 build

oleg-derevenetz:
An alternative to #9359.

`ubuntu-24.04` is now `ubuntu-latest`, and all this voodoo magic that was used to install ARM64 packages on the Ubuntu GitHub image no longer works, because `apt` now uses new configs of a different format, located in different places.

This PR fixes the build by removing all ARM64 targets.


ihhub:
I am closing this pull request as the alternative solution was taken for now.

PR #9375, opened 2024-12-23
Speed up GitHub Actions

ihhub:
According to the information about GitHub Runners which can be found [here](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners) Ubuntu runners have 4 cores. This means that we can speed up the compilation as previously every runner had only 2 cores.

What was changed:
- remove the hardcoded number of threads for Sonar Scanner. It can detect them automatically.

Before:
```bash
04:35:10.371 INFO  Available processors: 4
04:35:10.373 INFO  Using 2 threads for analysis, according to value of "sonar.cfamily.threads" property.
```

After:
```bash
02:09:57.107 INFO  Available processors: 4
02:09:57.109 INFO  Using 4 threads for analysis.
```

- use `$(nproc --all)` where possible for all Ubuntu runners. This approach makes us to avoid hardcoded values and also adjust to GitHub runner configuration.
- use `$(nproc --all)` where possible for all MacOS runners. This approach makes us to avoid hardcoded values and also adjust to GitHub runner configuration.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -54,7 +54,7 @@ jobs:
       if: ${{ github.event_name == 'push' && env.KEYSTORE_BASE64 != '' }}
       run: |
         cd android
-        ./gradlew assembleRelease bundleRelease --max-workers=4
+        ./gradlew assembleRelease bundleRelease
         cp app/build/outputs/apk/release/app-release.apk ../fheroes2.apk
         cp app/build/outputs/bundle/release/app-release.aab ../fheroes2.aab
       env:
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -37,7 +37,7 @@ jobs:
         sudo apt-get -y install gettext
     - name: Generate translations
       run: |
-        make -C files/lang -j $(nproc --all)
+        make -C files/lang -j "$(nproc)"
     - name: Create keystore
       if: ${{ github.event_name == 'push' && env.KEYSTORE_BASE64 != '' }}
       run: |
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -47,7 +47,7 @@ jobs:
       if: ${{ github.event_name == 'pull_request' || env.KEYSTORE_BASE64 == '' }}
       run: |
         cd android
-        ./gradlew assembleDebug bundleDebug --max-workers=4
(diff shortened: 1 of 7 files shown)


.github/workflows/codechecker.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - name: Analyze
       run: |
         export PATH="$HOME/.python-venv/bin:$PATH"
-        CodeChecker analyze --analyzers clangsa --ctu-all --ctu-ast-mode load-from-pch --output codechecker-output --ignore .codechecker-ignore build/compile_commands.json
+        CodeChecker analyze --jobs 4 --analyzers clangsa --ctu-all --ctu-ast-mode load-from-pch --output codechecker-output --ignore .codechecker-ignore build/compile_commands.json
```
oleg-derevenetz, line 30: CodeChecker autodetects the number of cores, so this parameter is redundant.

.github/workflows/android.yml
```diff
@@ -47,14 +47,14 @@ jobs:
       if: ${{ github.event_name == 'pull_request' || env.KEYSTORE_BASE64 == '' }}
       run: |
         cd android
-        ./gradlew assembleDebug bundleDebug
+        ./gradlew assembleDebug bundleDebug --max-workers=4
```
oleg-derevenetz, line 50: `--max-workers` is also [set by default to the number of processor cores](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_performance), so this setting is also redundant.

ihhub: I knew it would be the case :)

oleg-derevenetz (approved):
Hi @ihhub I made a few changes:

* I replaced `nproc --all` by just `nproc`. `nproc --all` may return invalid results on virtual machines, for example, on my home VM (which only has access to 6 CPU cores) `nproc` output looks like this:
  ```
  $ nproc
  6
  $ nproc --all
  128
  ```
* I use `sysctl -n hw.logicalcpu`, because it seems to be better suited for VMs according to [this manual](https://github.com/apple/darwin-xnu/blob/8f02f2a044b9bb1ad951987ef5bab20ec9486310/bsd/sys/sysctl.h#L1003).

ihhub:
@oleg-derevenetz , thank you for the corrections!

PR #9379, opened 2024-12-25
Add 3 new variants of Observation Tower object

ihhub:
All credits go to **PushPop** from Discord.

image

image


oleg-derevenetz (approved):

src/fheroes2/maps/map_object_info.cpp
```diff
@@ -4487,6 +4487,49 @@ namespace
...
+
+        // Observation Tower, snow terrain.
+        {
+            Maps::ObjectInfo object{ MP2::OBJ_OBSERVATION_TOWER };
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 238U, fheroes2::Point{ 0, 0 }, MP2::OBJ_OBSERVATION_TOWER, Maps::OBJECT_LAYER );
+
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 199U, fheroes2::Point{ -2, 0 }, MP2::OBJ_NONE, Maps::SHADOW_LAYER );
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 200U, fheroes2::Point{ -1, 0 }, MP2::OBJ_NONE, Maps::SHADOW_LAYER );
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 239U, fheroes2::Point{ 1, 0 }, MP2::OBJ_NON_ACTION_OBSERVATION_TOWER,
+                                                  Maps::BACKGROUND_LAYER );
```
Districh-ru, line 4526: Yes, there is a small pile of snow, but it does not block vertical movement on the right side of the tower.
изображение
Should we move the right part of this object to the `TERRAIN_LAYER`?
```suggestion
            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 239U, fheroes2::Point{ 1, 0 }, MP2::OBJ_NONE, Maps::TERRAIN_LAYER );
```

src/fheroes2/maps/map_object_info.cpp
```diff
@@ -4487,6 +4487,49 @@ namespace
...
+
+        // Observation Tower, desert terrain.
+        {
+            Maps::ObjectInfo object{ MP2::OBJ_OBSERVATION_TOWER };
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 236U, fheroes2::Point{ 0, 0 }, MP2::OBJ_OBSERVATION_TOWER, Maps::OBJECT_LAYER );
+
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 199U, fheroes2::Point{ -2, 0 }, MP2::OBJ_NONE, Maps::SHADOW_LAYER );
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 200U, fheroes2::Point{ -1, 0 }, MP2::OBJ_NONE, Maps::SHADOW_LAYER );
+            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 237U, fheroes2::Point{ 1, 0 }, MP2::OBJ_NON_ACTION_OBSERVATION_TOWER,
+                                                  Maps::BACKGROUND_LAYER );
```
Districh-ru, line 4511: The same comment:
```suggestion
            object.groundLevelParts.emplace_back( MP2::OBJ_ICN_TYPE_OBJNMUL2, 237U, fheroes2::Point{ 1, 0 }, MP2::OBJ_NONE, Maps::TERRAIN_LAYER );
```

Districh-ru (changes requested):
Hi, @ihhub,
I like how the new objects look! Many thanks, **PushPop**!

I left a comment regarding the passability of the Desert and Snow towers. Could you please take a look?

src/fheroes2/maps/map_object_info.cpp
ihhub: Hi @Districh-ru , it is intentional and also an action object cannot have terrain layer parts. An assertion at line 4955 (`assert( info.objectType == MP2::OBJ_NONE );`) will blow up because of this.

Districh-ru: Hi, @ihhub this assertion blows up if the `TERRAIN_LAYER` object part is not `MP2::OBJ_NONE`. So this change will not break the assertion:
изображение

IMHO the current passablilty is strange - the pile does not block the vertical movement, it can block the horizontal movement from the center to the right tile. But it is mine POV. :)

@Branikolog, @oleg-derevenetz and @zenseii, what do you think about these towers passability?
Should it be like for all other towers? Or the vertical movement at the right part of the tower should be blocked if there is a snow/sand pile part in the lower tile?

oleg-derevenetz: I really don't think that the same object should have different passability on different terrains.

ihhub: Hi @Districh-ru and @oleg-derevenetz , I've changed the logic to match the original tower passability.

Districh-ru (approved):

PR #9387, opened 2024-12-28
Add swap army and swap artifacts to the heroes meeting screen

zenseii:
This PR adds the swap army button and swap artifacts features to the heroes meeting screen.

Relates to #4084 

image

Button design by @FlorianBaba and me.
swap button

Swap artifacts was added to make the screen design more consistent and it has been requested by the community according to @Branikolog.

Rotation of the arrow parts was done with this ( sin270 = -1 and cos270 = 0): https://en.m.wikipedia.org/wiki/Rotation_matrix


src/fheroes2/army/army.cpp
```diff
@@ -1509,6 +1509,15 @@ void Army::MoveTroops( Army & from, const int monsterIdToKeep )
     moveTroops( false );
 }
 
+void Army::SwapTroops( Army & from )
+{
+    assert( this != &from );
+    const Troops temp = this->getTroops();
+    this->Reset();
```
oleg-derevenetz, line 1516: Hi @zenseii why there is a need to call `Army::Reset()`? It is intended mostly to reset the hero army to a default one, what is the purpose calling it here?

Branikolog:
Hi, @zenseii 
I've checked the PR.
Visually everything looks nice and absolutely naturally, as from design point.
I've noticed, that there's no pressed state for artifact swap button (there's no pressing animation for this only button).

Branikolog (approved):

oleg-derevenetz (approved):
Hi @zenseii I added an assertion to verify that hero will not get an empty army due to army exchange and also modernized some army-related code (while I'm here).

Districh-ru (approved):
Well done, @zenseii!
I have one suggestion that can be addressed in a new PR:
We should add the right-click info to help players fully understand the mechanics of all arrow-buttons in this dialog.

ihhub (approved):

ihhub:
@zenseii , huge thanks for implementing this!

Issue #7275, opened 2023-06-05
"File to Save:" window, allow a similar number of characters for the save file name between the standard keyboard and the mouse-operated keyboard


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the "File to Save:" window forces a maximum number of characters for the save file name.
But if you use the mouse-operated keyboard, there's no such limitation.
Personally, I'm in favor of a save file name with "no limit" of characters. :D
Related issue: https://github.com/ihhub/fheroes2/issues/1318.

In fheroes2:

![Save name length 001](https://github.com/ihhub/fheroes2/assets/43583994/85ba0483-98f4-4dd2-8985-2508bf1c8f72)

![Save name length 002](https://github.com/ihhub/fheroes2/assets/43583994/b317623c-6c42-41eb-b460-abf4fb7416ff)

### Save file

Here is an example of a save file name obtained via the mouse-operated keyboard:
[Broken_Alliance_00012345678901234567890123456789012345678901234567890123456789012345678901234567890.zip](https://github.com/ihhub/fheroes2/files/11655272/Broken_Alliance_00012345678901234567890123456789012345678901234567890123456789012345678901234567890.zip)

ihhub:
Relates to #1318

ihhub:
This issue was fixed by #7662.

Issue #1318, opened 2020-08-22
Limit on the number of characters accepted for the name of a save file


LeHerosInconnu:
In fheroes2, the number of characters you can use to name a save file is really low.
And it is not possible to change the position of the text cursor in the save name field with the arrow keys of the keyboard when we reach the character limit.
In the original game, the number of characters you can use to name a save file is really much higher, it is practically possible to write a novel. :)

ihhub:
I think this is not the case anymore as we allow to enter as many characters as you want (up to 255 if I am correct). Therefore I am closing this issue.

@LeHerosInconnu , please reopen this ticket if I am wrong here.

PR #9393, opened 2024-12-30
Implement scrollable text input

felix642:
Improved the editable text field in the save menu. When overflowing, the textbox would previously display the beginning of the text and complete with '...' for the overflowing text. The user could continue typing but would not get any feedback. With those changes the input box now scrolls when the text overflows.

[save.webm](https://github.com/user-attachments/assets/c6406c76-eb0d-48b9-b147-06156ff77a5d)

The changes were also applied to the virtual keyboard, but since this keyboard does not have any arrow key the user is not able to scroll left or right. This is still an improvement from the current implementation since the user it at least able to see what he is typing, but ideally we should also add some arrow keys or similar feature to be able to move the cursor left and right.

The changes are related to : #7275 and #1318 both issues are already fixed but with the current feedback it's hard to know the actual character limit.


src/fheroes2/gui/ui_text.h
```diff
@@ -257,14 +258,44 @@ namespace fheroes2
...
 
         bool _keepLineTrailingSpaces{ false };
     };
 
+    class TextInput final : public Text
+    {
+    public:
+        TextInput() = default;
+
+        TextInput( std::string text, const FontType fontType )
```
ihhub, line 274: Try to use `using Text::Text;` instead of redeclaring the constructor.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,46 @@ namespace fheroes2
...
+            return;
+        }
+
+        if ( _text.empty() ) {
+            return;
+        }
+
+        computeTextOffset( maxWidth );
+
+        const fheroes2::FontCharHandler charHandler( _fontType );
```
ihhub, line 387: We can reduce code duplication by:
```cpp
const fheroes2::FontCharHandler charHandler( _fontType );

computeTextOffset( maxWidth );
```
and remove line 402 as it won't be needed.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,46 @@ namespace fheroes2
...
+        assert( maxWidth > 0 );
+        if ( maxWidth <= 0 ) {
+            return;
+        }
+
+        if ( _text.empty() ) {
+            return;
+        }
+
+        computeTextOffset( maxWidth );
```
ihhub, line 385: Language switcher is missing here. The code won't work for non-English characters.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,46 @@ namespace fheroes2
...
+        if ( _cursorPosition < _textOffset ) {
+            // If the cursor is to the left of the textBox
+            _textOffset = _cursorPosition;
+            return;
+        }
+
+        const fheroes2::FontCharHandler charHandler( _fontType );
+
+        int32_t maxCharacterCount = getMaxCharacterCount( reinterpret_cast<const uint8_t *>( _text.data() + _textOffset ),
+                                                          static_cast<int32_t>( _text.size() - _textOffset ), charHandler, maxWidth );
```
ihhub, line 405: These lines are exact lines as 387-389. We could remove **computeTextOffset()** completely as just move the while loop below into the **fitToOneRow()** method.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,46 @@ namespace fheroes2
...
+        while ( _textOffset + maxCharacterCount <= _cursorPosition ) {
+            // If the cursor is to the right of the textBox
+            ++_textOffset;
+            maxCharacterCount = getMaxCharacterCount( reinterpret_cast<const uint8_t *>( _text.data() + _textOffset ), static_cast<int32_t>( _text.size() - _textOffset ),
+                                                      charHandler, maxWidth );
+        }
```
ihhub, line 411: Is it correct to say that the complexity of this loop is O(N * (N / 2)) where N is the number of characters present in the string? Is there a way to avoid recalculating all the character lengths?

ihhub (changes requested):
Hi @felix642 , I put few comments here. Would you mind to check them?

src/fheroes2/gui/ui_text.cpp
```diff
@@ -373,6 +373,8 @@ namespace fheroes2
 
     void TextInput::fitToOneRow( const int32_t maxWidth )
     {
+        constexpr size_t DISTANCE_TO_SIDE = 4;
```
Districh-ru, line 376: Hi, @felix642, let's name the variables according to https://github.com/ihhub/fheroes2/wiki/Code-style
And let's give it a more clear name, in example: `cursorToBorderDistance` or simply `cursorOffset`.

src/fheroes2/gui/ui_text.h
```diff
@@ -257,14 +258,38 @@ namespace fheroes2
...
         std::string _text;
 
         FontType _fontType;
 
         bool _keepLineTrailingSpaces{ false };
     };
 
+    class TextInput final : public Text
+    {
+        using Text::Text;
```
ihhub, line 271: Please put this statement within `public` section.

src/fheroes2/gui/ui_text.h
```diff
@@ -257,14 +258,38 @@ namespace fheroes2
...
+
+    public:
+        TextInput() = default;
```
ihhub, line 274: Do we need to have this constructor if we use `using Text::Text`?

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
         _text += truncatedEnding;
     }
 
+    void TextInput::fitToOneRow( const int32_t maxWidth )
+    {
+        constexpr size_t curosorToBorderDistance = 4;
```
ihhub, line 376: **curosor** is misspelled. Also, please declare this variable just before using it so it would be easier to identify where it is being used.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
...
+
+        const auto langugeSwitcher = getLanguageSwitcher( *this );
+        const fheroes2::FontCharHandler charHandler( _fontType );
+        const int32_t fullLineWidth
+            = getLineWidth( reinterpret_cast<const uint8_t *>( _text.data() ), static_cast<int32_t>( _text.size() ), charHandler, _keepLineTrailingSpaces );
+        if ( fullLineWidth < maxWidth ) {
+            return;
+        }
+
+        // If the cursor is to the left of the textBox
```
ihhub, line 391: Let's be consistent: some comments have a full stop at the end while others don't.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
...
+            const int32_t lineWidth = getLineWidth( reinterpret_cast<const uint8_t *>( _text.data() + _textOffset - 1 ),
+                                                    static_cast<int32_t>( _text.size() - _textOffset + 1 ), charHandler, _keepLineTrailingSpaces );
+            if ( lineWidth > maxWidth ) {
+                break;
+            }
+
+            --_textOffset;
+        }
+
+        // If the cursor is to the right of the Textbox
```
ihhub, line 405: `TextBox` is written here while at line 391 it is `textBox`.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
...
+        int32_t maxCharacterCount = getMaxCharacterCount( reinterpret_cast<const uint8_t *>( _text.data() + _textOffset ),
+                                                          static_cast<int32_t>( _text.size() - _textOffset ), charHandler, maxWidth );
+        while ( _textOffset + maxCharacterCount <= _cursorPosition + curosorToBorderDistance && _textOffset + maxCharacterCount < _text.size() ) {
```
ihhub, line 408: Could you please add parenthesis for these 2 conditions to separate them visually?

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
...
+        const size_t originalTextSize = _text.size();
+        _text = _text.substr( _textOffset, maxCharacterCount );
+
+        const std::string truncatedEnding( "..." );
+        const int32_t truncationSymbolWidth
+            = getLineWidth( reinterpret_cast<const uint8_t *>( truncatedEnding.data() ), static_cast<int32_t>( truncatedEnding.size() ), charHandler, true );
+
+        // Insert truncation symbol at the beginning if required.
+        if ( _textOffset != 0 ) {
+            const int32_t nbCharactersToReplace
```
ihhub, line 423: We do not use `nb` as abbreviation for `number` if it was the intention. If you would like to shorten the variable name you can name it as `charCountToReplace`.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
...
+                = getMaxCharacterCount( reinterpret_cast<const uint8_t *>( _text.data() ), static_cast<int32_t>( _text.size() ), charHandler, truncationSymbolWidth );
+            _text.erase( 0, nbCharactersToReplace );
+            _text.insert( 0, truncatedEnding );
+        }
+
+        // Insert truncation symbol at the end if required
+        if ( _text.size() + _textOffset < originalTextSize ) {
+            int totalWidth = 0;
+            int nbChars = 0;
```
ihhub, line 432: We do not use `nb` as abbreviation for `number` if it was the intention. If you would like to shorten the variable name you can name it as `charCount` or `charNumber`.

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,78 @@ namespace fheroes2
...
+
+        assert( maxWidth > 0 );
+        if ( maxWidth <= 0 || _text.empty() ) {
+            return;
+        }
+
+        const auto langugeSwitcher = getLanguageSwitcher( *this );
+        const fheroes2::FontCharHandler charHandler( _fontType );
+        const int32_t fullLineWidth
+            = getLineWidth( reinterpret_cast<const uint8_t *>( _text.data() ), static_cast<int32_t>( _text.size() ), charHandler, _keepLineTrailingSpaces );
```
ihhub, line 386: Is there a situation when we want not to keep trailing spaces? I believe for editing a string it is a must to show everything. If it is true then the code below to calculate the maximum number of characters can be reduce to just deduct a character width from the current one and do it again till we satisfy the requirements.

ihhub (changes requested):
Hi @felix642 , I left several comments here. Could you please take a look at them?

src/fheroes2/gui/ui_text.cpp
felix642: Hi @ihhub I'm not sure I understand properly what you're saying here. Are you referring to the fact that the current "view" is moved using the `getLineWidth()` method instead of using something like `FontCharHandler::getWidth()`?
If you could elaborate a bit more on your explanation I would greatly appreciate it.

ihhub: Answering your question in 2 parts:
- at this line we are calling `getLineWidth()` function with **_keepLineTrailingSpaces** parameter. Since `TextInput` class is designed to handle text input for users it is important to maintain whitespaces as we want to show a blinking cursor after them.
- in the code below this line we are trying to calculate a width of string iteratively excluding the last character from a string. If this is true then would it be more optimal to calculate the current width of a string and then deduct a single character's width from it?

src/fheroes2/gui/ui_text.cpp
```diff
@@ -371,6 +371,83 @@ namespace fheroes2
...
+
+        constexpr size_t cursorToBorderDistance = 4;
+
+        // If the cursor is to the left of the TextBox.
+        _textOffset = std::max( std::min( static_cast<int>( _textOffset ), static_cast<int>( _cursorPosition - cursorToBorderDistance ) ), 0 );
+
+        // If some characters were deleted and we have space for new characters.
+        int32_t currentWidth
+            = getLineWidth( reinterpret_cast<const uint8_t *>( _text.data() + _textOffset ), static_cast<int32_t>( _text.size() - _textOffset ), charHandler, true );
+        ;
```
ihhub, line 396: Please remove this line.

ihhub (changes requested):
Hi @felix642 , I put few more comments here. Can you please check them?

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -658,7 +658,21 @@ namespace fheroes2
         }
     }
 
-    size_t getTextInputCursorPosition( const std::string & text, const FontType fontType, const size_t currentTextCursorPosition, const int32_t pointerCursorXOffset,
+    size_t getTextInputCursorPosition( const TextInput & textInput, const std::string & fullText, const size_t currentTextCursorPosition,
```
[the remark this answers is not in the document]
felix642: Not exactly, the textInput might have some ellipsis (...) at the beginning which would return an invalid offset. So we have to use the original text to get the real location of the cursor.

src/fheroes2/gui/ui_text.h
```diff
@@ -257,14 +258,36 @@ namespace fheroes2
...
+        using Text::Text;
+
+        void setCursorPosition( const size_t position )
+        {
+            _cursorPosition = position;
+        }
+
+        void fitToOneRow( const int32_t maxWidth ) override;
+
+        size_t getOffset() const
```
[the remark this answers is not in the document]
felix642: This is the offset from the beginning of the text, so technically, an index offset in 1D array, do you want me to change it for something else ?

ihhub:
Hi @felix642 , can you please rebase this brach from master branch?

ihhub:
Hi @Districh-ru , could you please review this pull request again?

ihhub (approved):

Districh-ru:
> Hi @Districh-ru , could you please review this pull request again?

Hi @ihhub, sure, I'll do it a little later today.

Districh-ru:
This PR has an issue with fitting the text input field to one row:
изображение
изображение

I'm converting it to draft and will try to fix this.

Districh-ru:
Hi @felix642, thanks for fixing the width calculation!
Now it works good.

Districh-ru (approved):

oleg-derevenetz (approved):
Hi @Districh-ru I just made a few minor changes:

1. While the iterator comparison using `<` is valid in general and there are clear rules about this comparison, it may be slower than the `==` or `!=` because it implies the calculation of distance between two iterators, and this calculation may be fast or slow depending on the container. Usually it is fast for the random access containers, but I still replaced the comparison in a couple of places with `!=`;
2. We can break the iteration early in couple of places.

Please look at my changes, if everything is fine with them, then you are free to merge this PR.

P.S. Clang-Tidy suggests to replace the forward iteration by the range-based for, but I'll leave it as it is so that forward and reverse iteration match in style.

Districh-ru (approved):

PR #9398, opened 2024-12-31
Improved morale ui when all troops are undead

### LINKED ISSUE #2419: Hero morale indicator with undead army
While leading full undead army hero has neutral morale indicator. 
image
I suggest always show morale Indicator, as if hero had at least 1 alive creature in his army. This would help to predict the morale of any creature joined.
Ex: On a screenshot hero has expert leadership, granting +3 morale to all creatures. But Morale indicator shows neutral state, due to all undead creatures in hero's army. So if any alive creature join it would have +2 morale (As there's an undead troops in hero's army)
So my suggestion is to show +2 morale indicator for this case. Also, info should include modifiers: Expert leadership(+3) as well as undead troops in army (-1) 
While leading full undead army hero should have a negative (-1) morale indicator always without other modifiers.

felix642:
The morale modifier is now calculated when using an undead army. The previous behaviour was inconsistent depending on where the modifier was coming from (artifact vs hero).

Grayed out the morale icon if the hero is using an undead army. 

image

fixes #2419

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2024                                             *
+ *   Copyright (C) 2019 - 2025                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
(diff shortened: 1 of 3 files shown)


Branikolog (approved):
Works nice!

ihhub (approved):

ihhub:
@felix642 , many thanks for the fix!

PR #9401, opened 2025-01-01
Replace the macro-based include guards by #pragma once

oleg-derevenetz:
To be consistent with the new code. In addition, some old header files were renamed at the time, so their macro names do not match header file names anymore.


ihhub (approved):

ihhub:

PR #9402, opened 2025-01-01
Improve side tile passability logic

ihhub:
This is more a cosmetic change. The passability in general has not been changed. However, you will see less red corners for tiles.

Before the changes:
image

After the changes:
image

You can save 2 images and compare them to see what I mean by "red corners".

### DIFF
--- a/src/fheroes2/maps/maps_tiles.cpp
+++ b/src/fheroes2/maps/maps_tiles.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2024                                             *
+ *   Copyright (C) 2019 - 2025                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2009 by Andrey Afletdinov <fheroes2@gmail.com>          *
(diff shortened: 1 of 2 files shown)


src/fheroes2/maps/maps_tiles.cpp
```diff
@@ -735,16 +735,17 @@ void Maps::Tile::updatePassability()
...
         // This object "touches" the bottom part of the map. Mark is as inaccessible.
         _tilePassabilityDirections = 0;
         return;
     }
 
-    const Tile & bottomTile = world.getTile( GetDirectionIndex( _index, Direction::BOTTOM ) );
-    // If an object locates on land and the bottom tile is water mark the current tile as impassable. It's done for cases that a hero won't be able to
-    // disembark on the tile.
-    if ( !isWater() && bottomTile.isWater() ) {
+    const Tile & belowTile = world.getTile( GetDirectionIndex( _index, Direction::BOTTOM ) );
```
oleg-derevenetz, line 745: It's a bit of a controversial change IMHO, because we have `Direction::BOTTOM`, so it's natural to call this tile a "bottom tile" here and in other similar places, while "belowTile" rather looks like "something below the tile" (snail or earthworm maybe?...)

oleg-derevenetz (commented):
Hi @ihhub I just left one optional note related to the naming style :)

oleg-derevenetz (approved):

PR #9403, opened 2025-01-01
Add build target "web"

ololoken:
Add web platform support

Knows issues:
pthread support is unstable in emscripten and requires patching of sdk; build with enabled pthread is unstable and requires additional debugging

Plans:
- Add dialog to allow select `data` folder or zip archive with data
- Fix pthread support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,10 @@ android/app/src/main/assets/files/
 
 # Mac magic folders
 .DS_Store
+
+#web build artifacts
+*.wasm
+*.wasm.map
+fheroes2.js
+fheroes2.data
+src/dist/web/dist/*
--- a/Makefile.web
+++ b/Makefile.web
@@ -1,0 +1,37 @@
+###########################################################################
+#   fheroes2: https://github.com/ihhub/fheroes2                           #
+#   Copyright (C) 2021 - 2024                                             #
+#                                                                         #
+#   This program is free software; you can redistribute it and/or modify  #
+#   it under the terms of the GNU General Public License as published by  #
+#   the Free Software Foundation; either version 2 of the License, or     #
+#   (at your option) any later version.                                   #
+#                                                                         #
+#   This program is distributed in the hope that it will be useful,       #
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+#   GNU General Public License for more details.                          #
+#                                                                         #
+#   You should have received a copy of the GNU General Public License     #
+#   along with this program; if not, write to the                         #
+#   Free Software Foundation, Inc.,                                       #
+#   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             #
+###########################################################################
+
+# Options:
+#
+# FHEROES2_STRICT_COMPILATION: build in strict compilation mode (turns warnings into errors)
+# FHEROES2_WITH_DEBUG: build in debug mode
+# FHEROES2_DATA: set the built-in path to the fheroes2 data directory (e.g. /usr/share/fheroes2)
+
+PROJECT_NAME := fheroes2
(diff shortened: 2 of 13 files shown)


oleg-derevenetz:
> pthread support is unstable in emscripten and requires patching of sdk; build with enabled pthread is unstable and requires additional debugging

Eh, emscripten never changes. In principle, it is quite unstable, in my (admittedly, rather limited) experience, some things in it seem to start working, and then eventually break down again. In general, I'm not sure that it's worth adding its support to `master` yet, maybe it's worth waiting until the game starts working fully with it? I believe there is room for discussion here. What do you think @ihhub @Districh-ru @zenseii?

ihhub:
Hi @ololoken ,

We won't accept this method until the following requirement is going to be fulfilled:
- no assets from the original game are a part of the end product. Your solution and the link in the description is piracy. You cannot distribute the game. Every user must be able to use files on their machine like loading files from the local file system. If it is possible to do then we try to solve other problems.

ololoken:
ok, will add support of "non-piracy" workaround in fork first

ihhub:
@ololoken , you can still keep the pull request if you are going to update the code within the same branch. pthread support can be fixed by not allowing multithreading on the level of the engine. Also, check your solution for supporting video playback.

I didn't mean that won't accept emacripten support but it should be done in a way to comply with copyright 🙂 If we go with your solution then many people start pirating the game which is not we want.

oleg-derevenetz:
What I would also like to note is that the minimum set that should be present for merging to the `master` branch IMHO is as follows:

* Need to have a build recipe in our CI, just like it's done for PS Vita, for instance, so that, first of all, anyone should have a working build example in front of their eyes, and secondly, they should be able to download the build artifact as ZIP archive and run it on a local computer or web server (if it is necessary to "bake in" the files of the original game, then the demo version data files can be used, I believe it won't have any problems in terms of licensing);
* Need some documentation (like for PS Vita, Nintendo Switch or Android);
* Need to have a fully functional version (at least relatively). By "relatively" I mean that some tradeoffs are possible - for instance, background music may not resume from the remembered position, and `REWIND_AND_PLAY_INFINITE` playback mode may be used instead of `RESUME_AND_PLAY_INFINITE` (because it doesn't require the background thread, and there are endless issues with pthread in emscripten), but in general, background music should still work.

ololoken:
Hi, I will reopen MR a bit later.
Bundling of the data is not required. I used files from my original game just for demo. 
"Backing in" of the data from original game is bad idea not because it's bad and 'piracy', but because people may want to use set their own maps (mods?). 
Also there is problem with handling of huge data files in some browsers, so there is no option to embed full pack with all media data.

About pthread support, there are following issues:
1. emscripten sdk requires patching of ports system to resolve correct SDL_mixer dependency;
2. Disabled by default SharedArrayBuffer in browsers requires additional http server configuration or dirty hacks with service workers. Optionally user may enable the feature using browser flags, but it's not safe in scope of meltdown/spectre;
3. runtime hazenbugs with stackoverflows in js bindings;

My idea here is to reimplement asyncSoundManager using wasm workers api. But it's time consuming.

oleg-derevenetz:
> My idea here is to reimplement `asyncSoundManager` using wasm workers api. But it's time consuming.

`AsyncSoundManager` is not a vital thing per se. If the synchronous versions of the corresponding playback functions work fast enough in browser (this is especially related to MIDI because it's the main "retarder" on most systems), then it is possible not to use `AsyncSoundManager` at all. Another place is `MusicRestartManager` which can also be avoided by not using the `RESUME_AND_PLAY_INFINITE` playback mode with emscripten at all (what should be elementary achieved by forcing the `isMusicResumeSupported()` to always return `false` with emscripten).

ololoken:
[Demo](https://turch.in/fheroes2/public/) Asks user to provide required files.
Also I can add option to allow user download/upload saves and config.
@oleg-derevenetz Thanks, will take a look.

PR #9406, opened 2025-01-02
Add build target "web"

ololoken:
[Prev PR](https://github.com/ihhub/fheroes2/pull/9403) for reference
[Demo](https://turch.in/fheroes2/public/) no files bundled and user can provide own sets of maps, music, mods;

@oleg-derevenetz would you check how synchronous playback is done in AsyncManager::notifyWorker, smelly workaround but it works;

BTW ci/cd pipeline can be done using gh-pages

PR #9407, opened 2025-01-03
Temporary PR for Emscripten

oleg-derevenetz:
Related to #9406


ololoken:
@ihhub look
```cpp
 #ifdef __EMSCRIPTEN__
     EM_ASM( FS.syncfs( err = > err && console.warn( "FS.syncfs() error:", err ) ) );
 #endif
```
linter suggested you invalid js code: `err = > err && ....` is not ok 
so it's better disable clang-format checks for these lines

src/engine/serialize.cpp
```diff
@@ -410,6 +419,12 @@ bool StreamFile::open( const std::string & fn, const std::string & mode )
 void StreamFile::close()
 {
     _file.reset();
+
+#ifdef __EMSCRIPTEN__
+    // clang-format off
+    EM_ASM( FS.syncfs( err => err && console.warn( "FS.syncfs() error:", err ) ) );
```
oleg-derevenetz, line 425: I suppose that this call is no-op if there is nothing to sync, but if it leads to performance issues (since now it is called every time when `StreamFile` instance is closed and/or destroyed), it will be possible to think about how to optimize it.

src/engine/thread.cpp
```diff
@@ -54,13 +88,33 @@ namespace MultiThreading
...
+
+        while ( _runFlag ) {
+            const bool moreTasks = prepareTask();
+            if ( !moreTasks ) {
+                _runFlag = false;
+            }
+
+            {
+                // In accordance with the contract, the _mutex should NOT be acquired while calling the executeTask().
+                MutexUnlocker unlocker( _mutex );
```
oleg-derevenetz, line 110: I suppose that mutexes under Emscripten are "only for show", but it's better be safe than sorry, so we need to unlock the mutex if it is promised by the contract.

PR #9412, opened 2025-01-05

oleg-derevenetz:
* Don't create an empty anonymous namespace for non-Emscripten builds (code smell from SonarCloud);
* Limit the scope of the disabled Emscripten-specific `-Wvariadic-macro-arguments-omitted` warning.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this cleanup!

Issue #1329, opened 2020-08-24
Add option to not display movement path arrows


LeHerosInconnu:
Remaining:
In the original game, it is possible to not display the arrows of the hero's path when moving (see "show path" option in image).
Personally, I don't use this option.

In the original game:

![System Options window 01](https://user-images.githubusercontent.com/43583994/91023419-cf3fc580-e5f6-11ea-98f6-a0c410cfd668.png)

Issues done:
In the original game, it is possible to not display the movement of enemy heroes during the AI turn.
Personally, I use this option when I do tests, it saves time.

And the GUI mode can also be set to dynamic mode.
In this operating mode, the graphical interface is automatically adjusted according to the alignment of the castle played, the interface "Good" for the castles of the knight, the sorceress and the wizard, and the interface "Evil" for the castles of the barbarian, the warlock and the necromancer.

ihhub:
`Don't Show` option was added in the past.

ihhub:
The first option: Do Not Show Enemy was implemented in the past.

JoLiKMC:
I know this issue is a couple years old, now, but I'd like to add that I'm interested in seeing this "brought back", myself.  It seems like it'd be easy enough to implement – especially during the Campaign where the player is _distinctly_ playing under a "Good" prince or an "Evil" one.

I think LeHerosInconnu is right, too.  The starting castle _does_ determine the interface for the map.  A little judgmental of the game, perhaps, but I guess _someone's_ always gotta "be the bad guy", eh?

LeHerosInconnu:
Hello @zenseii,

> I personally don't see the merits of having the option to not show movement arrows added, aside from "the original game had it". If no one else does, then perhaps this issue should be closed?

I think that the option of not displaying the hero's path on the adventure map of the original game is there to offer the user a faster way to play (one click instead of two to move the hero), and it also allows for greater immersion in the game.
As I wrote, I don't use this feature, but other players might find it interesting.

ihhub:
Moving this ticket out of the scope of the project as none of us see a value in this feature.

ihhub:
Closing this issue. For the last 5 years none of players ever mentioned the need of this feature.

PR #9413, opened 2025-01-05
Add dynamic interface option

felix642:
Added dynamic interface (good / evil) based on the player's race. 
Relates to: #1329  (Despite the title, this issue also references the ability to hide the hero's path which was not added in this PR so we can't close it yet).


src/fheroes2/system/settings.cpp
```diff
@@ -883,9 +895,46 @@ bool Settings::isHideInterfaceEnabled() const
...
+    case InterfaceType::EVIL:
+        return true;
+    case InterfaceType::DYNAMIC: {
+        Player * player = Settings::Get().GetPlayers().GetCurrent();
+        if ( !player )
+            return false;
+
+        if ( player->isControlHuman() ) {
+            const int race = player->GetRace();
+            return race == Race::WRLK || race == Race::NECR || race == Race::BARB;
```
oleg-derevenetz, line 923: I suppose we can add special method for this in the `Race` namespace, say, `isEvilRace()` - similarly with existing `Race::isMagicalRace()`.

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -63,18 +63,30 @@ namespace
...
 
+        const fheroes2::Sprite * interfaceThemeIcon;
         std::string value;
-        if ( isEvilInterface ) {
-            value = _( "Evil" );
-        }
-        else {
+        switch ( interfaceType ) {
+        case DYNAMIC:
+            interfaceThemeIcon = &fheroes2::AGG::GetICN( ICN::SPANEL, 15 );
```
Districh-ru, line 72: We can make `uint32_t icnIndex;` and then set its value by the `interfaceType` and then use it in
`const fheroes2::Sprite & interfaceThemeIcon = fheroes2::AGG::GetICN( ICN::SPANEL, icnIndex );`
Doing this we will not duplicate the `GetICN()` call code and will not make a pointer to Sprite.
What do you think?

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -287,7 +299,7 @@ namespace fheroes2
                 windowType = showConfigurationWindow( saveConfiguration );
                 break;
             case SelectedWindow::InterfaceType:
-                conf.setEvilInterface( !conf.isEvilInterfaceEnabled() );
+                conf.setInterfaceType( static_cast<InterfaceType>( ( conf.getInterfaceType() + 1 ) % ( InterfaceType::DYNAMIC + 1 ) ) );
```
Districh-ru, line 302: To be more consistent with the other code we can add `InterfaceType::COUNT` as a last item of `InterfaceType` enumeration and use it here instead of `( InterfaceType::DYNAMIC + 1 )`.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -784,6 +784,13 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
                     // Reset environment sounds and music theme at the beginning of the human turn
                     AudioManager::ResetAudio();
 
+                    conf.SetCurrentColor( playerColor );
+                    if ( conf.getInterfaceType() == InterfaceType::DYNAMIC ) {
+                        reset();
+                        redraw( Interface::REDRAW_RADAR );
+                        redraw( Interface::REDRAW_ALL & ( ~Interface::REDRAW_RADAR ) );
+                    }
```
Districh-ru, line 792: In single-player game the race of the player is never changed so should we do this (`reset();` and radar redraw) only for the Hot Seat game?
And for the single-player game do only `reset();` after the line 737.
What do you think?

src/fheroes2/system/settings.cpp
```diff
@@ -235,8 +237,17 @@ bool Settings::Read( const std::string & filePath )
...
+        const std::string interfaceType = config.StrParams( "interface type" );
+        if ( interfaceType == "Good" ) {
+            setInterfaceType( InterfaceType::GOOD );
+        }
+        else if ( interfaceType == "Evil" ) {
+            setInterfaceType( InterfaceType::EVIL );
+        }
+        else {
+            setInterfaceType( InterfaceType::DYNAMIC );
+        }
```
Districh-ru, line 250: Should we set the **Good** interface by default (instead of Dynamic) if there is no configuration file or if it is corrupted?

Districh-ru (changes requested):
Hi, @felix642, I missed Dynamic interface option after the original game. Thanks for making this PR!
I left several comments to make it a bit better. Could you please check them when you have time?

src/fheroes2/system/settings.cpp
oleg-derevenetz: Personally I'm good with Dynamic interface by default BTW. Thus, people who do not like to regularly access the settings will be able to find out about its existence.

felix642: If I remember correctly this was the default behaviour in the original game hence why I chose to set the default value to dynamic. 
From a new player's perspective I feel like having a dynamic interface will give the best user experience and if for any reason they don't like it or prefer another option they can always change it and they will at least know that this option exists.

Districh-ru: @zenseii, @Branikolog, @ihhub what do you think?

Districh-ru: Sure, let's keep the Dynamic interface as default to correspond the original game!

Branikolog: > @zenseii, @Branikolog, @ihhub what do you think?

I'm fine with default dynamic interface. :)

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -63,18 +63,31 @@ namespace
...
+        case GOOD:
+            icnInx = 16;
             value = _( "Good" );
+            break;
+        case EVIL:
+            icnInx = 17;
+            value = _( "Evil" );
+            break;
+        default:
+            icnInx = 15;
```
Districh-ru, line 84: Hi, @felix642, to fix the CodeChecker issue we can initialize `icnInx` on the line 68 and remove lines 72 and 84. And also we should remove the line 85 by the same reason. The default sate of this switch should not happen normally, otherwise the assertion will blow up.

felix642: Thank you for the help @Districh-ru, it should be fixed now :)

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -65,12 +65,10 @@ namespace
         const Settings & conf = Settings::Get();
         const InterfaceType interfaceType = conf.getInterfaceType();
 
-        uint32_t icnInx;
-        std::string value;
+        uint32_t icnInx = 15;
+        std::string value = _( "Dynamic" );
```
Districh-ru, line 69: Hi, @felix642, while `15` is a constant, the `_( "Dynamic" )` is a translation function call.
So for the "Good" or "Evil" interface the translation will be always called for the "Dynamic" and then for "Good" or "Evil".

I suggest to only declare the `std::string value;` here and bring back `value = _( "Dynamic" );` to the DYNAMIC case.

Districh-ru (approved):
Well done!

src/fheroes2/game/game_startgame.cpp
```diff
@@ -735,6 +735,12 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
 
         // Fully update fog directions if there will be only one human player.
         Interface::GameArea::updateMapFogDirections();
+
+        if ( conf.getInterfaceType() == InterfaceType::DYNAMIC ) {
+            reset();
```
ihhub, line 740: This doesn't seem right. At line 702 we call reset and then we are calling it here. The code looks more like a hack. Could you please explain why we cannot do the same at line 702?

src/fheroes2/system/settings.h
```diff
@@ -64,6 +64,14 @@ enum class ZoomLevel : uint8_t
     ZoomLevel3 = 3, // Max zoom, but should only exists for debug builds
 };
 
+enum InterfaceType : uint8_t
+{
+    GOOD = 0,
+    EVIL = 1,
+    DYNAMIC = 2,
+    COUNT
```
ihhub, line 72: Either all entries have no value or all of them. SonarQube will complain about this warning.

src/fheroes2/system/settings.h
```diff
@@ -379,6 +389,7 @@ class Settings
     int ai_speed;
     int scroll_speed;
     int battle_speed;
+    InterfaceType _interfaceType;
```
ihhub, line 392: This member doesn't have default value.

src/fheroes2/system/settings.h
```diff
@@ -191,6 +199,9 @@ class Settings
     bool isHideInterfaceEnabled() const;
     bool isEvilInterfaceEnabled() const;
 
+    void setInterfaceType( InterfaceType type );
+    InterfaceType getInterfaceType() const;
```
ihhub, line 203: This method should be inlined.

src/fheroes2/system/settings.cpp
```diff
@@ -816,16 +839,6 @@ void Settings::setHideInterface( const bool enable )
     }
 }
 
-void Settings::setEvilInterface( const bool enable )
-{
-    if ( enable ) {
-        _gameOptions.SetModes( GAME_EVIL_INTERFACE );
```
ihhub, line 822: `GAME_EVIL_INTERFACE` must be renamed into `UNUSED_GAME_EVIL_INTERFACE` as we are no longer supporting it.

src/fheroes2/system/settings.cpp
```diff
@@ -883,9 +896,46 @@ bool Settings::isHideInterfaceEnabled() const
     return _gameOptions.Modes( GAME_HIDE_INTERFACE );
 }
 
+void Settings::setInterfaceType( InterfaceType type )
+{
+    assert( type >= InterfaceType::GOOD && type <= InterfaceType::DYNAMIC );
```
ihhub, line 901: Instead of avoid this assertion we simply can remove `InterfaceType::COUNT` from the enumeration.

src/fheroes2/system/settings.cpp
```diff
@@ -883,9 +896,46 @@ bool Settings::isHideInterfaceEnabled() const
...
 bool Settings::isEvilInterfaceEnabled() const
 {
-    return _gameOptions.Modes( GAME_EVIL_INTERFACE );
+    switch ( _interfaceType ) {
+    case InterfaceType::GOOD:
+        return false;
+    case InterfaceType::EVIL:
+        return true;
+    case InterfaceType::DYNAMIC: {
+        Player * player = Settings::Get().GetPlayers().GetCurrent();
```
ihhub, line 918: This method is not static. We should use its own methods within the body.

src/fheroes2/system/settings.cpp
```diff
@@ -883,9 +896,46 @@ bool Settings::isHideInterfaceEnabled() const
...
+        if ( !player )
+            return false;
```
ihhub, line 920: Is it even possible?

Also, please parenthesis for this if-body.

src/fheroes2/system/settings.cpp
```diff
@@ -883,9 +896,46 @@ bool Settings::isHideInterfaceEnabled() const
...
+        if ( player->isControlHuman() ) {
+            return Race::isEvilRace( player->GetRace() );
+        }
+
+        // Keep the UI of the last player during the AI turn
+        for ( auto iter = Settings::Get().GetPlayers().rbegin(); iter < Settings::Get().GetPlayers().rend(); ++iter ) {
+            if ( *iter && ( *iter )->isControlHuman() ) {
+                return Race::isEvilRace( ( *iter )->GetRace() );
+            }
+        }
```
ihhub, line 931: Do we need to check this? UI Interface changes only during player's turn.

src/fheroes2/system/settings.cpp
```diff
@@ -432,8 +443,20 @@ std::string Settings::String() const
...
+    os << std::endl << "# interface type (Good/Evil/Dynamic)" << std::endl;
+    switch ( _interfaceType ) {
+    case GOOD:
+        os << "interface type = Good" << std::endl;
+        break;
+    case EVIL:
+        os << "interface type = Evil" << std::endl;
+        break;
+    case DYNAMIC:
+        os << "interface type = Dynamic" << std::endl;
```
ihhub, line 455: None of settings have values starting from a capital letter. Please keep it the same way,.

src/fheroes2/system/settings.cpp
```diff
@@ -235,8 +237,17 @@ bool Settings::Read( const std::string & filePath )
         setBattleShowTurnOrder( config.StrParams( "battle turn order" ) == "on" );
     }
 
-    if ( config.Exists( "use evil interface" ) ) {
-        setEvilInterface( config.StrParams( "use evil interface" ) == "on" );
```
ihhub, line 239: Here we are breaking the existing settings players set. We should read the value of the previous setting.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -784,7 +790,15 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
...
                     AudioManager::ResetAudio();
 
+                    conf.SetCurrentColor( playerColor );
+
                     if ( isHotSeatGame ) {
+                        if ( conf.getInterfaceType() == InterfaceType::DYNAMIC ) {
+                            reset();
+                            redraw( Interface::REDRAW_RADAR );
+                            redraw( Interface::REDRAW_ALL & ( ~Interface::REDRAW_RADAR ) );
+                        }
```
ihhub, line 800: Do we want to rebuild everything even if the interface type is going to be the same?

src/fheroes2/dialog/dialog_interface_settings.cpp
```diff
@@ -287,7 +297,7 @@ namespace fheroes2
                 windowType = showConfigurationWindow( saveConfiguration );
                 break;
             case SelectedWindow::InterfaceType:
-                conf.setEvilInterface( !conf.isEvilInterfaceEnabled() );
+                conf.setInterfaceType( static_cast<InterfaceType>( ( conf.getInterfaceType() + 1 ) % InterfaceType::COUNT ) );
```
ihhub, line 300: This is the only place where `InterfaceType::COUNT` is really needed. We can simply do this for the code to avoid having an extra enumeration entry:
```cpp
if ( conf.getInterfaceType() == InterfaceType::DYNAMIC ) {
    conf.setInterfaceType( InterfaceType::GOOD );
}
else if ( conf.getInterfaceType() == InterfaceType::GOOD ) {
    conf.setInterfaceType( InterfaceType::BAD );
}
else {
    conf.setInterfaceType( InterfaceType::DYNAMIC );
}
```
Then we won't need checks in the code for non-existing types.

ihhub (changes requested):
Hi @felix642 , I put several comments here. Would you mind to take a look?

src/fheroes2/game/game_startgame.cpp
felix642: This is because when the game is first drawn (L. 702 - 710), the current player has not been selected yet (L. 734). 
This code was added to redraw the interface to make sure that it matches the current player, but you're right, there should be a way to only draw the interface once, I'll check if I can move some code around to make it happen.

src/fheroes2/system/settings.cpp
felix642: Might be related to my last comment where the current player might not be set at the start of the game. I'll remove this check if I update the code once I find a way to only draw the interface once at the beginning of the game.

src/fheroes2/system/settings.cpp
felix642: Some parts of the UI is updated during the AI's turn. 
Mainly the top part with the shield:
image

If I remove this code, the background of the shield will change color very rapidly based on the AI factions

ihhub:
Hi @felix642 , I converted this pull request into a draft. Once you update it please mark it as ready for review again.

src/fheroes2/system/settings.cpp
felix642: This is needed since this method is called in the Main Menu where there is no current Player.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -784,7 +785,15 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
                     // Reset environment sounds and music theme at the beginning of the human turn
                     AudioManager::ResetAudio();
 
+                    conf.SetCurrentColor( playerColor );
+
                     if ( isHotSeatGame ) {
+                        if ( conf.getInterfaceType() == InterfaceType::DYNAMIC && _isCurrentInterfaceEvil != conf.isEvilInterfaceEnabled() ) {
+                            reset();
```
ihhub, line 792: Could you please check whether `reset();` call is needed? It is used to recreate the whole object while here we don't need to do this.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -784,7 +785,15 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
...
+                            redraw( Interface::REDRAW_RADAR );
+                            redraw( Interface::REDRAW_ALL & ( ~Interface::REDRAW_RADAR ) );
```
ihhub, line 794: I think this call is not needed as we have line 804 which does the same.

src/fheroes2/game/game_startgame.cpp
```diff
@@ -804,8 +813,6 @@ fheroes2::GameMode Interface::AdventureMap::StartGame()
                         Game::DialogPlayers( playerColor, "", _( "%{color} player's turn." ) );
```
ihhub, line 813: Here is a problem: we are rendering a new dialog for a player. If you play a hotseat you don't want to show the Adventure Map before pressing OKAY on this dialog. Is it correct?

src/fheroes2/gui/ui_option_item.cpp
```diff
@@ -95,19 +95,28 @@ namespace fheroes2
...
+        case GOOD:
+            icnInx = 16;
             value = _( "Good" );
+            break;
+        case EVIL:
+            icnInx = 17;
+            value = _( "Evil" );
+            break;
+        default:
+            assert( 0 );
```
ihhub, line 115: Please add `break;` at the end of default clause as it is done for other places.

src/fheroes2/gui/ui_option_item.h
```diff
@@ -25,6 +25,7 @@
 
 #include "math_base.h"
 
+enum InterfaceType : uint8_t;
```
ihhub, line 28: Please add an empty line to separate forward declaration and namespace.

src/fheroes2/system/settings.cpp
```diff
@@ -235,7 +237,20 @@ bool Settings::Read( const std::string & filePath )
     }
 
     if ( config.Exists( "use evil interface" ) ) {
-        setEvilInterface( config.StrParams( "use evil interface" ) == "on" );
+        const bool isEvil = config.StrParams( "use evil interface" ) == "on";
```
ihhub, line 240: Please add a comment stating that this is done for the backward compatibility:
```cpp
// This code handles a configuration file's parameter made by an older versions of the engine.
// The original "use evil interface" parameter is not longer being set.
```

src/fheroes2/system/settings.cpp
```diff
@@ -880,9 +897,41 @@ bool Settings::isHideInterfaceEnabled() const
     return _gameOptions.Modes( GAME_HIDE_INTERFACE );
 }
 
+void Settings::setInterfaceType( InterfaceType type )
```
ihhub, line 900: This method can be inlined within the header file.

src/fheroes2/system/settings.cpp
```diff
@@ -880,9 +897,41 @@ bool Settings::isHideInterfaceEnabled() const
...
+        if ( !player ) {
+            return false;
+        }
+
+        if ( player->isControlHuman() ) {
+            return Race::isEvilRace( player->GetRace() );
+        }
+
+        // Keep the UI of the last player during the AI turn
+        for ( auto iter = Settings::Get().GetPlayers().rbegin(); iter < Settings::Get().GetPlayers().rend(); ++iter ) {
```
ihhub, line 923: Please cache `Settings::Get().GetPlayers()` call as it is not instant.

ihhub, line 923: The usual operation with iterators is `not equal` to compare with `end()`:
```cpp
iter != Settings::Get().GetPlayers().rend()
```

Not all containers guarantee support of `less than` opreator.

src/fheroes2/system/settings.cpp
```diff
@@ -880,9 +897,41 @@ bool Settings::isHideInterfaceEnabled() const
...
+            if ( *iter && ( *iter )->isControlHuman() ) {
+                return Race::isEvilRace( ( *iter )->GetRace() );
+            }
+        }
+        break;
+    }
+    default:
+        assert( 0 );
```
ihhub, line 931: Please add `break;` as it is done for other parts of the code.

src/fheroes2/system/settings.h
```diff
@@ -61,7 +61,14 @@ enum class ZoomLevel : uint8_t
     ZoomLevel0 = 0,
     ZoomLevel1 = 1,
     ZoomLevel2 = 2,
-    ZoomLevel3 = 3, // Max zoom, but should only exists for debug builds
+    ZoomLevel3 = 3 // Max zoom, but should only exists for debug builds
```
ihhub, line 64: Please revert this change. Nothing wrong by having a comma for end entry in an emuration as it implies that the future entries won't need to change this line of this code.

src/fheroes2/system/settings.h
```diff
@@ -379,6 +391,7 @@ class Settings
     int ai_speed;
     int scroll_speed;
     int battle_speed;
+    InterfaceType _interfaceType{ InterfaceType::GOOD };
```
ihhub, line 394: Since this is `uint8_t` type we want to avoid increasing the size of the class. Please put this member after `ZoomLevel _viewWorldZoomLevel{ ZoomLevel::ZoomLevel1 };`.

ihhub (changes requested):
Hi @felix642 , I added several more comments in this pull request. Would you mind to take a look at them?

src/fheroes2/game/game_startgame.cpp
felix642: This call is needed to update the interface.
The call to `reset()` will call `setPos()` on many objects in the interface. This is where the objects are assigned an icon and a position (You can check for instance the file `interface_buttons.cpp` `Interface::ButtonsPanel::SetPos`)
Wihtout any call to this method, some buttons will not display the proper theme.  

I could theoretically update the methods to update the theme in the `draw()` method, but I felt like this would be too complex for the small performance gain that we would get. What do you think?

src/fheroes2/game/game_startgame.cpp
felix642: This call is needed to update the interface.
The call to `reset()` will call `setPos()` on many objects in the interface. This is where the objects are assigned an icon and a position (You can check for instance the file `interface_buttons.cpp` `Interface::ButtonsPanel::SetPos`)
Wihtout any call to this method, some buttons will not display the proper theme.  

I could theoretically update the methods to update the theme in the `draw()` method, but I felt like this would be too complex for the small performance gain that we would get. What do you think?

src/fheroes2/system/settings.cpp
felix642: There was an invalid call to `Settings::Get()` since this method is not static and already part of the Settings class. I have now replaced the line with `GetPlayers().rend/rbegin();`.
AFAIK, this shouldn't add any overhead. Do you think it's still worth to cache the `GetPlayers()` call ?

ihhub: Caching `Settings::Get()` is a good step forward. I will take a look at the rest of the code.

src/fheroes2/game/game_startgame.cpp
ihhub: I am fine with the current approach. However, please verify that in the hot-seat mode we don't have any issues like incorrect interface type or seeing the interface type of the opponent before hitting OKAY button for their turn.

src/fheroes2/system/settings.cpp
```diff
@@ -429,8 +446,20 @@ std::string Settings::String() const
...
+        os << "interface type = good" << std::endl;
+        break;
+    case EVIL:
+        os << "interface type = evil" << std::endl;
+        break;
+    case DYNAMIC:
+        os << "interface type = dynamic" << std::endl;
+        break;
+    default:
+        assert( 0 );
```
ihhub, line 461: Please add missing `break;` call to be consistent with the rest of the code.

src/fheroes2/system/settings.cpp
```diff
@@ -429,8 +446,20 @@ std::string Settings::String() const
     os << std::endl << "# show turn order during battle: on/off" << std::endl;
     os << "battle turn order = " << ( _gameOptions.Modes( GAME_BATTLE_SHOW_TURN_ORDER ) ? "on" : "off" ) << std::endl;
 
-    os << std::endl << "# use evil interface style: on/off" << std::endl;
-    os << "use evil interface = " << ( _gameOptions.Modes( GAME_EVIL_INTERFACE ) ? "on" : "off" ) << std::endl;
+    os << std::endl << "# interface type (good/evil/dynamic)" << std::endl;
```
ihhub, line 449: I suggest to follow the comment structure as for other settings:
```cpp
os << std::endl << "# interface type: good/evil/dynamic" << std::endl;
```

src/fheroes2/system/settings.h
```diff
@@ -191,6 +198,15 @@ class Settings
     bool isHideInterfaceEnabled() const;
     bool isEvilInterfaceEnabled() const;
 
+    void setInterfaceType( InterfaceType type )
+    {
+        _interfaceType = type;
+    }
+    InterfaceType getInterfaceType() const
```
ihhub, line 205: Please add an empty line to visually separate 2 function bodies.

ihhub (changes requested):
Hi @felix642 , I left 3 very small comments here. Could you please check them?

ihhub (approved):

ihhub:
@felix642 , thank you so much for this feature!

PR #9414, opened 2025-01-05
Fix some of the HTML errors reported by SonarQube

oleg-derevenetz:
* Add the `<!DOCTYPE html>`;
* Remove the `text-stroke` CSS property. This property is non-standard and should be used with the `-webkit-` vendor prefix only (and property with the correct name is already in use).


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for these corrections!

PR #9416, opened 2025-01-06
Update the sources.props for tools

oleg-derevenetz:
After #9063, `engine/image_tool.cpp` depends on the `engine/agg.file.h`, which is not included in the `sources.props` of the corresponding tools. This doesn't break the build because the file itself is in place, but it still need to be added to the corresponding `sources.props` for the sake of convenience of working with relevant projects in the Visual Studio IDE.


ihhub (approved):

ihhub:
@oleg-derevenetz , well done!

PR #9417, opened 2025-01-06
Unify the Emscripten and non-Emscripten builds

oleg-derevenetz:
* The Emscripten build and the general build differ from each other in only one thing - in the name of the target file (`fheroes2.js` vs `fheroes2`). There is no need to have a separate `Makefile` in a separate directory just for Emscripten.
* `rm -f` works just fine in the Makefile `clean` targets, but the canonical way for GNU Make to ignore potential errors is to use the `-` prefix for the "potentially failing" commands.


Makefile.vita
```diff
@@ -41,7 +41,7 @@ fheroes2.vpk: eboot.bin param.sfo translations
 		--add files/data=files/data \
 		--add files/lang/vita_temp=files/lang \
 		fheroes2.vpk
-	rm -r files/lang/vita_temp
+	rm -rf files/lang/vita_temp
```
ihhub, line 44: Should we do the same (`-rm`) for this place as well?

ihhub: Actually, I realized that this is not for clean command so probably we should fail here as it is a part of a build command.

ihhub: Then looks good for me.

ihhub (approved):

ihhub:

Issue #9411, opened 2025-01-05
WASM version improvements


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

We've recently added WASM version of the engine in #9406. However, there are many things we need to do to make this version up to our standards.

### Describe the possible solution

This is a list of tasks which we will need to do:
- [x] add documentation as we've done for other platforms
- [ ] provide all JS files as a part of the package: no files should be downloaded from the Internet
- [ ] explicitly state in the page Heroes of Might and Magic II, not Heroes 2
- [ ] add a description at the bottom of the page explaining what this page is for and also why this version might crash/doesn't work/be slow. Important: state that all uploaded files are actually being uploaded on the same machine and being stored in a cache
- [x] wait till https://github.com/emscripten-core/emscripten/pull/23094 is merged
- [ ] deploy this version as a part of GitHub Actions to https://ihhub.github.io/fheroes2/
- [ ] change the main readme file to link the deployed version of the engine
- [ ] verify the WASM version on different browsers (possibly limit access only for desktop versions as we don't know at the moment how to handle file loading on mobile devices)
- [ ] add an option to download the demo version of game via a separate button as we did for Android
- [ ] add copyright info for index.html page
- [ ] once a player exits the game we should show the initial index.html page
- [x] fix logs being treated as errors
- [x] fix mouse being unresponsive after switching on V-Sync mode
- [ ] fix the fullscreen mode issues in different browsers

ihhub:
Hi @oleg-derevenetz , could you please look through the proposed list and comment if anything doesn't not make sense or I missed something?

Hi @zenseii , would you mind to take a look at possible changes in the documentation?

Hi @ololoken , could you please check if any of these tasks can be handled by you?

oleg-derevenetz:
> deploy this version as a part of GitHub Actions to https://ihhub.github.io/fheroes2/

I think we need to solve the issue with pthreads first. There is an unmerged (yet) PR to the upstream emscripten repo that should allow us to build fheroes2 with pthreads and throw away all these crutches for our `AsyncManager` (and probably should resolve the issue with the background music chokes).

> add an option to download the demo version of game via a separate button as we did for Android

[...]

PR #9421, opened 2025-01-06
[emscripten] additionaly build modularized version

ololoken:
Emscripten build will produce additional package where generated JavaScript is wrapped into a factory function. It's useful for ondemand wasm module instantiation.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -365,6 +365,7 @@ jobs:
       run: |
         # Translations and H2D files are already included in fheroes2.data
         7z a -bb1 -tzip -- fheroes2_emscripten.zip LICENSE changelog.txt fheroes2.data fheroes2.js fheroes2.wasm ./docs/README.txt ./files/emscripten/*
+        7z a -bb1 -tzip -- fheroes2_emscripten.module.zip LICENSE changelog.txt fheroes2.module.data fheroes2.module.js fheroes2.module.wasm ./docs/README.txt
     - uses: actions/upload-artifact@v4
       if: ${{ github.event_name == 'pull_request' }}
       with:
--- a/.gitignore
+++ b/.gitignore
@@ -101,5 +101,7 @@ android/app/src/main/assets/
 *.wasm
 *.wasm.map
 fheroes2.js
+fheroes2.module.js
 fheroes2.data
+fheroes2.module.data
 src/dist/web/dist/*
--- a/Makefile.emscripten
+++ b/Makefile.emscripten
@@ -33,6 +33,9 @@ fheroes2.js: translations
 	cp src/dist/fheroes2-ems/fheroes2.data .
 	cp src/dist/fheroes2-ems/fheroes2.js .
 	cp src/dist/fheroes2-ems/fheroes2.wasm .
+	cp src/dist/fheroes2-ems/fheroes2.module.data .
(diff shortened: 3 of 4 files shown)


oleg-derevenetz:
Hi @ololoken If you want to do this to improve the functionality of your launcher, there is no need at all to touch the project sources or file a PR. Just use this:

```
LDFLAGS="-sMODULARIZE -sEXPORTED_RUNTIME_METHODS=FS,ENV,noExitRuntime,addOnExit,run,callMain -sEXPORT_NAME=fheroes2" emmake make -f Makefile.emscripten
```

This works just fine (just tried with my changes from #9417).

Because today your launcher needs one set of exported symbols, and you file this PR. Tomorrow your (or some other) launcher will need another set of exported symbols, and you (or the author of that other launcher) will file another PR. Instead of all this nonsense, I suggest just to use the standard GNU Make mechanisms that allow you to specify the additional build parameters from the outside without having to touch the source code.

ololoken:
@oleg-derevenetz yes, but the plan is to download wasm module artifact into launcher on build

oleg-derevenetz:
> yes, but the plan is to download wasm module artifact into launcher on build

Well, as someone said, "imagine that this is done more than once". Imagine that some other guy decides to write some other 3-rd party launcher and he/she will need some different set of build options...

ololoken:
Not a problem, lets use EXPORT_ALL once and forever :/

oleg-derevenetz:
> Not a problem, lets use EXPORT_ALL once and forever :/

Please don't do this. There is no need to put some ad-hoc build parameters or build variants, that are intended just for your own 3-rd party launcher, into the main repo. Why don't you just want to build fheroes2 directly into your launcher's repo with the parameters you need, as I showed above? It can even be automated in future by monitoring the creation of new fheroes2 releases.

ololoken:
ok, got your point

ololoken:
BTW, let's consider if the project needs an official launcher?

oleg-derevenetz:
> BTW, let's consider if the project needs an official launcher?

As far as I understand, you decided to make your own launcher, which will be located [outside of this project](https://github.com/ihhub/fheroes2/issues/9411#issuecomment-2571728120) and you want to write it yourself, as you please, right?

ololoken:
I told it is [temporary solution](https://github.com/ihhub/fheroes2/blob/master/files/emscripten/index.html). It's not maintainable, created  just to check if wasm build works. 
Right now to run fheroes2.wasm you need:
1. serve generated files using http server
2. mount and populate virtual FS with original game data files

There is option to generate single html file. In that case *.js, *.wasm and *.data will be embedded as base64 encoded strings into large html. Theoretically, this file can be downloaded and opened in any browser locally. 

But I decided to develop standalone launcher as it's easier to maintain something implemented as standard web application.

oleg-derevenetz:
> But I decided to develop standalone launcher as it's easier to maintain something implemented as standard web application.

Sure. But please understand that constantly introducing changes in the `master` branch via pull requests for each brand new idea related to the launcher development doesn't look good. It's even elementary slow process. Instead of doing this, I suggest either to build the WASM with your settings in your launcher's repo in the manner shown above, where you can quickly make and test the changes you need at the moment, or, if you prefer, you can make the necessary changes in some branch of your own fheroes2 fork, and then, when launcher will be more or less ready and usable, offer the single "final" pull request.

PR #9418, opened 2025-01-06
Update background image for PSV version of the engine

ihhub:
To reflect the correct name of the project.

PR #9423, opened 2025-01-06
Set the PROJECT_VERSION for PSV in XX.YY format

oleg-derevenetz:
Related to https://github.com/ihhub/fheroes2/pull/9418#issuecomment-2573384981

Still can not detect some type of errors - e.g. the absence of `version.txt` itself (will set the version to 00.00 in this case), because GNU Make runs the shell in `sh` mode, which doesn't support the `bash`extensions - e.g. the `-o pipefail` option, but it's still better than no error handling at all.


Districh-ru:
Hi, @oleg-derevenetz, this PR works good on emulator:
изображение

I've asked **Pønytail** to test this build on his hardware.
If everything is OK, I'll approve this PR.

oleg-derevenetz:
> this PR works good on emulator

As far as I understand, on emulator everything works fine even without this PR? :)

Districh-ru:
> As far as I understand, on emulator everything works fine even without this PR? :)

Yes, the emulator reads the version string and displays it (as a string) when installing and then in the applications list:
изображение

The real hardware, I think, does some version analysis which needs `XX.YY` format.

ihhub (approved):

Districh-ru:
Hi, @lovenemesis, this PR should fix the issue you have encountered on PS Vita.
Could you please test it?
[fheroes2_psv.zip ](https://github.com/ihhub/fheroes2/actions/runs/12668859087/artifacts/2400738885)

oleg-derevenetz:
It looks like noone wants to get the fheroes2 running on PSV :) Is it worth to still support this platform? :)

ihhub:
@oleg-derevenetz ,

Well, yes and no :) However, we should reduce the priority for fixing bugs for this platform.

Districh-ru (approved):
**Pønytail** has reported that this build works and installs well on his PS Vita hardware and  he is very grateful for this fix!

ihhub:

PR #9444, opened 2025-01-11
Further fixes and improvements related to WebAssembly

oleg-derevenetz:
Related to #9411

* Allow to build the Wasm version with or without multithreading support, as desired. The advantages of the non-multithreaded version are that it does not require a web server to return the special set of HTTP headers.
* Enable LTO for the Emscripten builds.
* Build the Emscripten version in the [strict mode](https://emscripten.org/docs/tools_reference/settings_reference.html#strict).
* Improve the mechanics of music feeding during playback, fix the deadlock (see the comments in the corresponding source files).

I'm currently running Wasm build of fheroes2 (with threads) in the browser (Chromium-based) in the AI auto control mode for a few hours, and I should say that the situation with playing background music has improved significantly. ~~The only strange thing is that immediately after starting the game and when loading the first map, the memory consumption of the corresponding browser tab increases sharply from 500Mb to 5-6Gb for several seconds ( and during this time music is played intermittently). After a few seconds, the memory consumption drops back to 500Mb, and loading of other maps (or the same map) no longer causes such an effect, memory consumption is not growing anymore. This is very similar to the GC behavior, which means that some JavaScript code is most likely eating up memory (not the fheroes2 itself).~~

P.S. What's interesting, there is no such an effect in the Firefox.

P.P.S. The issue with memory consumption has been resolved.


src/dist/Makefile.emscripten
```diff
@@ -26,24 +26,45 @@ CCFLAGS := $(CCFLAGS) \
...
 	--preload-file ../../../files/lang/@/files/lang/ \
 	--preload-file ../../../files/timidity/@/etc/ \
 	-sALLOW_MEMORY_GROWTH \
 	-sASYNCIFY \
-	-sASYNCIFY_STACK_SIZE=32768 \
-	-sENVIRONMENT=web,worker \
+	-sASYNCIFY_STACK_SIZE=65536 \
+	-sINCOMING_MODULE_JS_API=canvas,preRun,setStatus \
 	-sINITIAL_MEMORY=128mb \
+	-sMEMORY_GROWTH_LINEAR_STEP=64mb \
```
ihhub, line 41: I would reduce it to 32 MB as I've barely seen the application to go beyond 120 MB on my machine so having something over 140 MB is highly unlikely. What do you think?

ihhub:
Hi @oleg-derevenetz , I tried your link but it doesn't work for me: I am pointing to the correct HoMM2 installation folder, the browser asks to upload 338 files and nothing happens after.

src/dist/Makefile.emscripten
```diff
@@ -26,24 +26,45 @@ CCFLAGS := $(CCFLAGS) \
...
 
 LDFLAGS := $(LDFLAGS) \
 	--preload-file ../../../files/data/resurrection.h2d@/files/data/resurrection.h2d \
 	--preload-file ../../../files/lang/@/files/lang/ \
 	--preload-file ../../../files/timidity/@/etc/ \
 	-sALLOW_MEMORY_GROWTH \
 	-sASYNCIFY \
-	-sASYNCIFY_STACK_SIZE=32768 \
-	-sENVIRONMENT=web,worker \
+	-sASYNCIFY_STACK_SIZE=65536 \
```
ihhub, line 38: Could you please explain what issue the enlarged stack size solves?

src/fheroes2/gui/interface_status.cpp
```diff
@@ -489,16 +489,18 @@ void Interface::StatusPanel::TimerEventProcessing()
 
 void Interface::StatusPanel::drawAITurnProgress( const uint32_t progressValue )
 {
+    // Even if there is no need to draw anything, we still need to pump the event queue to
+    // update the position of the software-emulated mouse cursor, feed the music player by
+    // another music chunk on some platforms (e.g. WebAssembly), etc.
+    LocalEvent::Get().HandleEvents( false );
```
ihhub, line 495: Since we are modifying this code I am suggesting to check the return value for the call to gracefully exist the application if possible:
```cpp
if ( !LocalEvent::Get().HandleEvents( false ) ) {
    return;
}
```

ihhub (changes requested):
Hi @oleg-derevenetz , I added just few comments here. Could you please take a look at them?

src/fheroes2/gui/interface_status.cpp
oleg-derevenetz: Hi @ihhub this method is called from the AI code and, even if we return from here without doing anything, the AI code will still continue, so what is the purpose of the proposed change?

ihhub: It's a fair point. Then leave it as it is.

oleg-derevenetz:
Hi @ihhub

> I tried your link but it doesn't work for me: I am pointing to the correct HoMM2 installation folder, the browser asks to upload 338 files and nothing happens after.

Well, this is a sort of a known issue. First of all, in my tests I use the "minimum" set of assets consisting of two AGG files and 2-3 maps of different size. Second, even if you "uploaded" the assets files, it doesn't mean that `fheroes2.wasm` and `fheroes2.data` are already downloaded and ready to run. Unfortunately, our current HTML launcher does not clearly enough reflect the current state of the application's readiness for launch. 

I also have the build with the "stock" launcher from Emscripten itself with built-in assets from the demo version, are you interested in the URL?

src/dist/Makefile.emscripten
oleg-derevenetz: OK, let's try to change it.

oleg-derevenetz:
Hi @ihhub I also fixed an issue with excessive memory consumption (and halved the size of the resulting wasm binary) - guys from the Emscripten team helped a lot. The updated version ia available at the Sourceforge - and there is also a build with demo version baked in, and I can give you a link (if you find this appropriate).

ihhub:
Hi @oleg-derevenetz , could you please temporary provide the link with the demo? I would like to see how it works.

oleg-derevenetz:
Hi @ihhub

> could you please temporary provide the link with the demo? I would like to see how it works.

Just add the `/demo` to the SourceForge URL provided above. Did you able to run the version with our custom launcher BTW? Since `fheroes2.wasm` is now just 7Mb instead of 20Mb, it's easier to start now because you don't have to wait for this file to load and be ready to run.

ihhub:
Hi @oleg-derevenetz , let me give a try to run locally. I haven't ran it these days as I barely touched GitHub due to being busy :)

oleg-derevenetz:
@ihhub

> let me give a try to run locally. I haven't ran it these days as I barely touched GitHub due to being busy :)

Unfortunately, SourceForge removed my website, probably they found it "suspicious" or like this, probably because doesn't have a repo with source files. I'll try to host it on the GitHub Pages. Unfortunately, there is no way to host the multithreaded version on GitHub Pages because it requires the custom HTTP headers.

ihhub:
Hi @oleg-derevenetz , this is what I received when I tried to upload a demo version locally:
image

What I did: downloaded a compile Artifact from GitHub, unzipped it, open with Mozilla (other browsers behave the same) and chose demo directory in fheroes2 which is available once you run scripts.

oleg-derevenetz:
> What I did: downloaded a compile Artifact from GitHub, unzipped it, open with Mozilla (other browsers behave the same) and chose demo directory in fheroes2 which is available once you run scripts.

Well, yes, `file://` URIs are not respected by CORS. Need to run "real" local Web server. Try this:

This is the custom build with the Emscripten launcher and demo assets baked in.

ihhub:
> Hi @ihhub
> 
> > What I did: downloaded a compile Artifact from GitHub, unzipped it, open with Mozilla (other browsers behave the same) and chose demo directory in fheroes2 which is available once you run scripts.
> 
> Well, yes, `file://` URIs are not respected by CORS. Need to run "real" local Web server (I use IIS for that on my Windows machine). Try this:
> 
> https://oleg-derevenetz.github.io/fheroes2/
> 
> This is the custom build with the Emscripten launcher and demo assets baked in.

V-Sync option breaks the game. We must disable it.

oleg-derevenetz:
> V-Sync option breaks the game. We must disable it. I am not able to mouse click on anything else after enabling it but keyboard works.

We'll have to see if we can fix it somehow. If the keyboard is still working, it means that the events are generally passing.

ihhub:
> > V-Sync option breaks the game. We must disable it. I am not able to mouse click on anything else after enabling it but keyboard works.
> 
> We'll have to see if we can fix it somehow. If the keyboard is still working, it means that the events are generally passing.

We can fix it in another pull request. Let me test for a while before merging this PR.

oleg-derevenetz:
> We can fix it in another pull request. Let me test for a while before merging this PR.

Sure. Moreover, I don't even know how to fix it yet and what exactly is the matter here :)

ihhub:
Another issues: logs are considered as errors. While this is true for real errors, it is wrong to text support mode. Do you know is there a differentiation of log events for WebAssembly?

oleg-derevenetz:
> Another issues: logs are considered as errors. While this is true for real errors, it is wrong to text support mode. Do you know is there a differentiation of log events for WebAssembly?

Our `COUT` macro is currently looking like this:

```
std::cerr << x << std::endl;
```

so no wonder that everything looks like errors :) We can make a custom macro for Emscripten with printing to `std::cout`.

ihhub:
> Another issues: logs are considered as errors. While this is true for real errors, it is wrong to text support mode. Do you know is there a differentiation of log events for WebAssembly?

I believe we should use something as `console.trace()` for our logs. However, we can fix it separately as well.

ihhub:
> > Another issues: logs are considered as errors. While this is true for real errors, it is wrong to text support mode. Do you know is there a differentiation of log events for WebAssembly?
> 
> Our `COUT` macro is currently looking like this:
> 
> ```
> std::cerr << x << std::endl;
> ```
> 
> so no wonder that everything looks like errors :) We can make a custom macro for Emscripten with printing to `std::cout`.

Let's do it separately. The rest looks good. Good job!

ihhub (approved):

ihhub:
Hi @oleg-derevenetz ,

Mouse event coordinates are shifted down for me in full screen mode. After reloading the game also works fine with enabled V-Sync but reloading is not what users expect us to do :)

ihhub:
By the way, if I switch off V-Sync the same happens so I suspect that something to do V-Sync toggle in SDL.

oleg-derevenetz:
> Mouse event coordinates are shifted down for me in full screen mode. After reloading the game also works fine with enabled V-Sync but reloading is not what users expect us to do :)

I mean, it looks more like a "video driver" problem than like events problem.

oleg-derevenetz:
For some reason, after destroying and recreating the texture and renderer, mouse events no longer arrive. It's not that they come with the wrong coordinates - they don't come at all.

PR #9449, opened 2025-01-15
WebAssembly: improve the logging

oleg-derevenetz:
Related to the https://github.com/ihhub/fheroes2/pull/9444#issuecomment-2592428569

Here is the version with our "custom" launcher (non-multithreaded build, because GitHub Pages doesn't allow to return the custom HTTP headers):

And here is the version with the "stock" launcher from Emscripten (and with bundled assets from the demo version):

https://oleg-derevenetz.github.io/fheroes2/demo/

I prefer to use the `emscripten_out()` and not JS `console.*` because it can be intercepted by the custom printing function in the launcher (see how it's done in the "stock" launcher above, it has a sort of "console" below the game canvas - it can be also useful for the text support mode, BTW). I don't use the `emscripten_dbg()` though, because its limitations (like "this symbol is only available in debug builds") do not look acceptable, because It's not too good to be bound by such restrictions when it comes to debugging - Emscripten debug builds are quite large and slow. If there is a need to quickly do some debugging in the console, then there is no reason to always build with `-O0`.

### DIFF
--- a/src/engine/logging.h
+++ b/src/engine/logging.h
@@ -143,7 +143,8 @@ namespace Logging
         _log_strstream << x;                                                                                                                                             \
         emscripten_out( _log_strstream.str().c_str() );                                                                                                                  \
     }
-#else // Default: log to STDERR
+#else
+// Default: log to stderr
 #define COUT( x )                                                                                                                                                        \
     {                                                                                                                                                                    \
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:

PR #9453, opened 2025-01-15
WebAssembly: fix the mouse issues when toggling the vsync mode and switching to full screen mode

oleg-derevenetz:
Related to https://github.com/ihhub/fheroes2/pull/9444#issuecomment-2592404037

This PR uses `SDL_RenderSetVSync()`, which is available since SDL 2.0.18 (that's why the Ubuntu version in the "Linux x86-64 SDL2 Release" workflow has been bumped from 20.04 to 22.04) instead of re-creating the renderer "on the fly" and also it removes a lot of cryptic code related to the SDL rendering.

Also it bumps the emsdk version to ~~4.0~~ latest and removes the workaround from our CI that was required to be able to build the multithreaded version of fheroes2.

As usual, you can see this PR running here:

https://oleg-derevenetz.github.io/fheroes2/demo/

There are still some issues left - for instance, when trying to change the in-game resolution when the game running in full screen mode, there are some inadequate switching, stretching, etc., but I don't know yet what to do about it anyway.


src/engine/screen.cpp
```diff
@@ -1171,7 +1132,57 @@ namespace
 
             _createPalette();
 
-            return _createRenderer( resolutionInfo.gameWidth, resolutionInfo.gameHeight );
+            // SDL_PIXELFORMAT_INDEX8 is not supported by SDL 2 even being available in the list of formats.
```
oleg-derevenetz, line 1135: Hi @ihhub could you please explain how this comment relates to the following lines? Maybe it should be moved somewhere else?

ihhub: Hi @oleg-derevenetz ,

It's an old comment and it's not valid anymore. You can remove it.

src/engine/screen.cpp
```diff
@@ -846,23 +838,17 @@ namespace
...
         bool isFullScreen() const override
         {
-            if ( _window == nullptr )
+            if ( _window == nullptr ) {
                 return BaseRenderEngine::isFullScreen();
+            }
 
-            const uint32_t flags = SDL_GetWindowFlags( _window );
-            return ( flags & SDL_WINDOW_FULLSCREEN ) != 0 || ( flags & SDL_WINDOW_FULLSCREEN_DESKTOP ) != 0;
+            return ( SDL_GetWindowFlags( _window ) & ( SDL_WINDOW_FULLSCREEN | SDL_WINDOW_FULLSCREEN_DESKTOP ) );
```
ihhub, line 851: I would still leave `const uint32_t flags = SDL_GetWindowFlags( _window );` as a separate string for debugging purposes as we have several issues and improvements related to rendering.

src/engine/screen.cpp
```diff
@@ -1257,82 +1258,38 @@ namespace
...
-            // it is important to lock the mouse in the application window area.
-            if ( isFullScreen() ) {
-                SDL_SetWindowGrab( _window, SDL_TRUE );
-            }
-            else {
-                SDL_SetWindowGrab( _window, SDL_FALSE );
-            }
-        }
-
-        bool _createRenderer( const int32_t width_, const int32_t height_ )
```
ihhub, line 1272: Can we please keep the original method **_createRenderer()** as it was created to separate this logic from others? We can though move `_toggleMouseCaptureMode();` out of the method.

src/engine/screen.cpp
```diff
@@ -1257,82 +1258,38 @@ namespace
...
-            if ( _texture == nullptr ) {
-                ERROR_LOG( "Failed to create a texture from a surface of " << width_ << " x " << height_ << " size. The error: " << SDL_GetError() )
-                clear();
-                return false;
-            }
-
-            if ( !_retrieveWindowInfo() ) {
-                clear();
-                return false;
+            if ( const int returnCode = SDL_RenderSetVSync( _renderer, _isVSyncEnabled ? SDL_ENABLE : SDL_DISABLE ); returnCode != 0 ) {
```
ihhub, line 1269: Here we don't actually need to have a variable declaration inside the if-statement as it is the only code within this method.

ihhub (changes requested):
Hi @oleg-derevenetz , I put several comments in this pull request. Could you please take a look at them and see if they make sense?

src/engine/screen.cpp
oleg-derevenetz: Hi @ihhub I did this on purpose. Since these mouse issues are caused by re-creating the renderer on the fly, I combined the methods so that no one would be tempted to try this again.

src/engine/screen.cpp
oleg-derevenetz: Hi @ihhub Well, yes, that's currently the case. But why not limit the variable scope, if there is a tool to enforce this? :)

src/engine/screen.cpp
```diff
@@ -830,14 +830,6 @@ namespace
 #else
                 flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 #endif
-
-                SDL_GetWindowSize( _window, &_windowedSize.width, &_windowedSize.height );
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub this trick caused issues when toggling the full screen mode in a web browser. I'll make a special build a bit later with this part reverted to demonstrate the issue.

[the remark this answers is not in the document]
oleg-derevenetz: @ihhub OK, here it is:

https://oleg-derevenetz.github.io/fheroes2/bug/

How to reproduce:

1. Run the game;
2. Press F4 button on the main screen to switch to the full-screen mode;
3. Press F4 button once again to switch back;
4. Notice that the game screen has been significantly shifted (and also probably resized). Also the mouse events have invalid coordinates.

src/engine/screen.cpp
oleg-derevenetz: @ihhub It's done.

src/engine/screen.cpp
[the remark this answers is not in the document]
ihhub: It works absolutely fine for me. I did exact steps as your described:
image

However, the code which was removed is easily reproducible on Windows if you don't use soft rendering mode.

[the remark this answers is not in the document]
oleg-derevenetz: @ihhub For me, it looks like this (after switching back from the full screen mode):

bug

I just run fheroes2 with default resolution (640x480), press the F4 button two times and get this :)

> However, the code which was removed is easily reproducible on Windows if you don't use soft rendering mode.

Yes, sure, I did a few tests and it looks like that SDL without that workaround just always use the `SDL_WINDOW_FULLSCREEN_DESKTOP` mode if the window has a non-standard (not supported by the hardware) resolution.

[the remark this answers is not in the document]
oleg-derevenetz: And in Firefox I get this:

bug

:)

oleg-derevenetz:
@ihhub what do you think about playability of the wasm version in general (except for some rough edges maybe like the full-screen mode)? Is it good enough to consider this version as "general availability" and publish it on the project's website?

ihhub:
> @ihhub what do you think about playability of the wasm version in general (except for some rough edges maybe like the full-screen mode)? Is it good enough to consider this version as "general availability" and publish it on the project's website as a "demo" (in a way)? I'm not talking about the launcher right now, but only about the game itself.

From the playability point of view, for me it looks good. I played for a while and didn't notice any issues with it, but I haven't played much.

I am still a bit sceptical about making it "general availability" right now, as the version needs polishing. We need to try to make the version as the same quality as others. However, it might not happen as it sits on top of browsers and we cannot control them. What we can do is to minimize the risks.

Therefore, I think a plan of actions should be the following:
- try to fix these edge cases. If we cannot fix some features and we shouldn't spend too much time on fixing them, we can disable them for now. This also would make players to use a proper desktop version of the engine since we can promote WebAssembly version as a "demo" like you stated.
- do a deep testing. What we found together is just based on little gameplay. We would need @Branikolog to do some work on this.
- ask the players to do the testing. We can do this within the Discord server.
- update the documentation and prepare it for the general availability but state that this version is unstable

I would refrain announcing about WebAssmebly in the upcoming 1.1.6 release. The same we did for Android in the past. We needed more time to test things and only then we made a public announcement.

ihhub:
@oleg-derevenetz , by the way, you've done a really great job bring this version to a stable state. Well done!

oleg-derevenetz:
Hi @ihhub IMHO, what we need most right now is a polished/reworked launcher. I think here's an approximate list of what should be in the GA version of the launcher:

* It should be more user friendly. Just two buttons without a clear explanation of what they do is not enough. It is necessary to explain to the user on the web page what exactly he needs to do in order to launch the game (probably even with localization).
* If it is waiting for something to load, then it must clearly display to the user what exactly it is waiting for (with the progress, if possible). If it is waiting for the data file to load, then it should display this. If it is waiting for the wasm file to load, then it also should display this. If it is waiting for the game assets, then it also should display this. Etc, etc.
* If there is an error, then it should be reported to the user. A more general error message can be displayed in the GUI, and a more technical one can be displayed in the browser console. It should not be that an error has occurred, but it is not shown to the user, as a result, the user is sitting and waiting for who knows what.
* It should be part of the fheroes2 project, because if the original author is unwilling or unable to support it, then someone else could do it (at least fix bugs).
* It should be as license-friendly as possible, and should not use third-party JS libraries or fonts with obscure licenses directly from the Internet (like `<script src="https://some.web.site.com/script.js">`).

If there was a person who would responsibly deal with this challenge, it would be good. Of course, I can do something simple, but I am completely unfamiliar with modern best practices in terms of creating web frontends, and to be honest, this is not for me. However, I've heard that there are a lot more front-end developers than C++ programmers :)

src/engine/screen.cpp
```diff
@@ -1257,82 +1280,38 @@ namespace
...
+            assert( _renderer != nullptr );
 
-            _texture = SDL_CreateTextureFromSurface( _renderer, _surface );
-            if ( _texture == nullptr ) {
-                ERROR_LOG( "Failed to create a texture from a surface of " << width_ << " x " << height_ << " size. The error: " << SDL_GetError() )
-                clear();
-                return false;
-            }
+#if !SDL_VERSION_ATLEAST( 2, 0, 18 )
+#error SDL_RenderSetVSync() is only supported since SDL 2.0.18
```
ihhub, line 1288: I assume that this will work only on systems that have an SDL version higher or equal to 2.0.18. In terms of support I have a feeling that there would be some systems where it won't be possible if packages were not being updated for a long time.  Also, having different SDL versions everywhere as was discovered in the past is a path for extra workload on our shoulders. What do you think about a solution where we somehow force certain systems to build a required SDL version as a part of our engine. This is not related to this pull request in general.

ihhub: Before doing any actions I will check what real minimum version we really need based on functions we utilise in the code.

oleg-derevenetz: @ihhub Do you mean that we just add a single `#if !SDL_VERSION_ATLEAST(...)` somewhere instead of multiple different ones?

ihhub: Yes but to do this we need to know the real minimum version of SDL we need to compile the engine.

oleg-derevenetz: @ihhub I suppose we can do this later in a separate PR. It looks like we don't have many places with `SDL_VERSION_ATLEAST` checks, less than 10 places in total.

ihhub:
Hi @Branikolog , please prioritize this pull request.

ihhub:
Hi @Branikolog , please complete your review by today.

Branikolog:
Hi! 
Everything works fine, but I've noticed a strange thing, once I've switched to window mode and back to Fullscreen. When turning on/off v-sync screen partially shifts (I think it's related to my machine), and the window mode icon switches for a moment.


https://github.com/user-attachments/assets/eee60878-f790-4108-bb38-610d66704f2a

Not so critical issue, though.

oleg-derevenetz:
> once I've switched to window mode and back to Fullscreen

Full screen is seriously broken in wasm (and it's broken differently in different browsers), and I don't know yet how to cure it in a "cross-platform" way. It may have to be disabled altogether in the wasm version.

Branikolog:
@oleg-derevenetz , hi.

> > once I've switched to window mode and back to Fullscreen
> 
> Full screen is seriously broken in wasm (and it's broken differently in different browsers), and I don't know yet how to cure it in a "cross-platform" way. It may have to be disabled altogether in the wasm version.

The issue I've filmed is valid for the PC version, since I'm not able to run application with browser. 

And I'm convinced, that Fullscreen shouldn't be an option for the browser.

oleg-derevenetz:
> The issue I've filmed is valid for the PC version, since I'm not able to run application with browser.

Yes, I tried to run the game in the "true full-screen" mode (I mean, on Windows with hardware mouse cursor), and yes, when the `SDL_RenderSetVSync()` is called, there is a single frame rendered with some invalid contents. Sometimes it is apparently some "remembered" contents from the windowed mode, and, if app was never running in the windowed mode - i.e. it was started in the full-screen mode - it is just a black frame. This is especially noticeable during the step-by-step execution in the debugger. To be honest, I do not know if anything can be done about it. This does not happen in windowed mode or in the `SDL_WINDOW_FULLSCREEN_DESKTOP` ("non-true-fullscreen") mode.

Branikolog:
@oleg-derevenetz 

Yeah, forgot to mention, that the issue occurs for hardware cursor.
From my point it's mostly a cosmetic issue. I just mentioned it just to let you know and to check if nothing is broken there. 👍 

Thanks for clarification!

ihhub:
Hi @oleg-derevenetz , let's merge it after 1.1.6 release.

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for the changes!

PR #9454, opened 2025-01-16
Fix overlap of FPS info and popup message in the Editor

Districh-ru:
Relates to #8589 

img


ihhub:

PR #9455, opened 2025-01-16
Add a button to toggle developer mode in debug build

Districh-ru:
Sometimes it is needed to debug with and without the developer mode.
In example: when you need to check the rendering (and you don't need the extra rendered passabilities and other) and also to check the tile info in console.

Without this PR you need to quit fheroes2, edit the fheroes2.cfg by setting the magic debug value and restart fheroes2.

This PR also adds **\`** (backquote) and **~** (tilde) to the processed keys and makes ` the default key to toggle the developer mode.

### DIFF
--- a/src/engine/localevent.cpp
+++ b/src/engine/localevent.cpp
@@ -613,6 +613,8 @@ namespace EventProcessing
                 return SDLK_AMPERSAND;
             case fheroes2::Key::KEY_QUOTE:
                 return SDLK_QUOTE;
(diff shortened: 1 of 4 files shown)


oleg-derevenetz (approved):
LGTM. Just added a few empty lines and sorted the event handlers according to the order of the event declarations.

ihhub (approved):

ihhub:
@Districh-ru , many thanks for this improvement!

PR #9456, opened 2025-01-16
Fix mine flag rendering and don't crop its sprite, remove unused flag setting logic

Districh-ru:
Relates to #5716

This PR:
- removes the logic for setting flags for pre v.1.0 experimental feature to capture Magic Garden, Windmill and Water Wheel;
- reduces the size of `ICN::FLAG32` sprites vector;
- fixes mine flag rendering when it is on (and mostly over) the top edge of the Adventure map:

Master build:

https://github.com/user-attachments/assets/aba99eeb-5597-4a09-b5a8-5b41bffeb7cb

This PR:

https://github.com/user-attachments/assets/02486932-3d6e-4fd7-8fbc-0f9212bd6578

### DIFF
--- a/src/fheroes2/world/world.cpp
+++ b/src/fheroes2/world/world.cpp
@@ -1585,8 +1585,8 @@ IStreamBase & operator>>( IStreamBase & stream, World & w )
         ++w.heroIdAsLossCondition;
     }
 
-    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_PRE_1_1106_RELEASE, "Remove the logic below." );
-    if ( Game::GetVersionOfCurrentSaveFile() < FORMAT_VERSION_PRE_1_1106_RELEASE ) {
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_PRE1_1106_RELEASE, "Remove the logic below." );
+    if ( Game::GetVersionOfCurrentSaveFile() < FORMAT_VERSION_PRE1_1106_RELEASE ) {
         // Update flags for Mine and Lighthouse captured objects.
         for ( const auto & [tileIndex, object] : w.map_captureobj ) {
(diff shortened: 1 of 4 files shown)


src/fheroes2/world/world.cpp
```diff
@@ -1584,6 +1585,36 @@ IStreamBase & operator>>( IStreamBase & stream, World & w )
         ++w.heroIdAsLossCondition;
     }
 
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_PRE1_1106_RELEASE, "Remove the logic below." );
+    if ( Game::GetVersionOfCurrentSaveFile() < FORMAT_VERSION_PRE1_1106_RELEASE ) {
+        // Update flags for Mine and Lighthouse captured objects.
+        for ( const auto & [tileIndex, object] : w.map_captureobj ) {
+            if ( object.objCol.first == MP2::OBJ_MINE ) {
```
ihhub, line 1592: We can do this way to ease code readability:
```cpp
const int color = object.GetColor();
if ( color == Color::NONE ) {
    // This object is not owned by anyone.
    continue;
}

if ( object.objCol.first == MP2::OBJ_MINE ) {
    ...
}
else if ( object.objCol.first == MP2::OBJ_LIGHTHOUSE ) {
    ...
}
```

ihhub (changes requested):
Hi @Districh-ru , I left few comments here. Would you mind to take a look at them?

src/fheroes2/world/world.cpp
Districh-ru: Hi, @ihhub, I did not cache `object.GetColor()` because it is inlined and is used second time only if the object is found.

ihhub: It makes sense. Thanks for explaining!

ihhub (approved):

Branikolog (approved):

ihhub:

PR #9462, opened 2025-01-18
WebAssembly: build both the multithreaded and non-multithreaded versions in parallel, optimize the workflows a bit

oleg-derevenetz:
In this PR:

* Both multithreaded and non-multithreaded WebAssembly binaries are now built in parallel;
* The WebAssembly release is not created/updated anymore, since it is mostly useless without a Web server - you can't just run it locally using the `file://` URL. In future there will be GitHub Pages version instead;
* Workflows are optimized a bit. Although there is no way to skip the matrix jobs using the job-level `if:` statement (because it is evaluated [before](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif) the matrix is applied), we still can skip the job steps depending on both external parameters (such as `github.event_name`) and matrix parameters. Thanks to this, now the jobs that should not be performed during the pushes to the `master` branch (e.g. any build jobs that don't create any artifacts to release them) will not be performed to save the GitHub Actions resources a bit;
* `matrix.config` is now replaced by the `matrix.include`, which is well-documented and allows to replace the `${{ matrix.config.param }}` substitutions by their shorter version - just `${{ matrix.param }}`.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -9,7 +9,7 @@ jobs:
   cmake:
     strategy:
       matrix:
-        config:
+        include:
         - name: Linux SDL2
           os: ubuntu-latest
           dependencies: libsdl2-dev libsdl2-mixer-dev libsdl2-image-dev gettext ninja-build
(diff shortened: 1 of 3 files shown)


ihhub:
Hi @oleg-derevenetz , should I manually delete [Emscripten build (latest commit)](https://github.com/ihhub/fheroes2/releases/tag/fheroes2-emscripten-sdl2_dev) build for now?

oleg-derevenetz:
Hi @ihhub

> should I manually delete [Emscripten build (latest commit)](https://github.com/ihhub/fheroes2/releases/tag/fheroes2-emscripten-sdl2_dev) build for now?

Once this PR will be merged, you can safely do this, because it will not be updated anymore.

ihhub (approved):

ihhub:

PR #9463, opened 2025-01-18
Do full image copy with one 'memcpy()' call

Districh-ru:
This PR does full `memcpy()` of image data if a full image copy is requested instead of calling `mamcpy()` in a loop for every image line.

### DIFF
--- a/src/engine/image.cpp
+++ b/src/engine/image.cpp
@@ -1291,6 +1291,12 @@ namespace fheroes2
 
     void Copy( const Image & in, Image & out )
     {
+        if ( !out.singleLayer() && !in.singleLayer() ) {
+            // Both images have transform layer. Copy using the assignment operator.
+            out = in;
+            return;
+        }
+
         const int32_t width = in.width();
         const int32_t height = in.height();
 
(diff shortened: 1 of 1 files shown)


src/engine/image.cpp
```diff
@@ -1291,8 +1291,26 @@ namespace fheroes2
...
+            memcpy( out.image(), in.image(), size );
+        }
+        else if ( in.singleLayer() ) {
+            // Copy image layer and set transform to non-transparent mode.
+            memcpy( out.image(), in.image(), size );
+            memset( out.transform(), static_cast<uint8_t>( 0 ), size );
+        }
+        else {
+            // Copy both image and transform layers.
+            memcpy( out.image(), in.image(), size * 2 );
```
ihhub, line 1312: I highly recommend to use assignment operator for this case as we could change the behavior of `Image::copy()` method in the future. We also want to avoid repeating the logic of Image class here:
```cpp
out = in;
```

ihhub (changes requested):
Hi @Districh-ru , I left one suggestion here. Could you please take a look at it?

src/engine/image.cpp
Districh-ru: Sure)

ihhub (approved):

ihhub:
@Districh-ru , a bunch of thanks for the improvement!

PR #9465, opened 2025-01-18
Remove temporary FindSDL2_* CMake scripts

oleg-derevenetz:
Modern OS distributions and package managers provide newer versions of `SDL_mixer` and `SDL_image` packages that properly register themselves and don't require custom CMake scripts to find them.


ihhub:
Hi @oleg-derevenetz , I assume this will work only on systems that have newest versions of SDL, won't it?

oleg-derevenetz:
Hi @ihhub

> I assume this will work only on systems that have newest versions of SDL, won't it?

Not quite. This change doesn't affect the SDL itself, the auxiliary script for the SDL itself was removed a few years ago. This change affects only the auxiliary SDL libs, such as `SDL_mixer` or `SDL_image`. Systems have a "new enough" versions of these libs for a quite some time, for instance, Ubuntu 22.04 already have them snd doesn't need any crutches of this kind, the same is for macOS with MacPorts or Homebrew.

ihhub (approved):

ihhub:

PR #9468, opened 2025-01-19
Fix a crash in View World mode for the Editor

ihhub:
Also, make zoom x1 available for the Editor.


Districh-ru (approved):
Well done, @ihhub!
I've also fixed borders in editor when the screen height is 480:
изображение
Please check the updated code if it is OK. :)

PR #9469, opened 2025-01-19
Fix the build with NDEBUG

oleg-derevenetz:
For instance, the Android release build is failing because of this:

https://github.com/ihhub/fheroes2/actions/runs/12853012354/job/35835681593


ihhub:
Oh, no! My bad...

ihhub (approved):

oleg-derevenetz:
> thanks for the quick fix! I was not sure why it wasn't failing before :(

Because #9466 has just been merged, and this issue is observed only during the build with `NDEBUG`, and in general, it is not customary for us to turn the assertions off.

PR #9479, opened 2025-01-22
Fix text rendering for the level up dialog

### LINKED ISSUE #9459: Adventure map, shortcut is not used for hero level up dialog
### Preliminary checks

### Platform

Windows

### Describe the bug

Some words in skills icons are too long:
Image

While right click info window works fine:

Image

### Save file

None.

### Additional info

_No response_

ihhub:
close #9459

Additionally, right click popup messages for LEARN buttons were added.


Branikolog:
Hi, @ihhub !
I've noticed, that text is aligned differently at lvlup window and hero window:
Lvlup:
image

hero/skill info windows:
image

The second one is shifted 1px to the left, while the first one is perfectly centered.

I suppose it is not related to the PR. Do you want me to open a separate issue for aligning texts on skills in hero/info windows?

Branikolog:
@ihhub hi!
I've checked the lvlup dialog and everything looks good and consistent for both window and info hints.
However, I've noticed, that for some texts a shorter form is not used for hero info dialog:
👍 
image

👎 
image

The initial issue related to lvlup dialog only. 
Should I open a separate issue for fixing hero info dialog?

Branikolog:
Hi, @Mr-Bajs !

> Hi, @ihhub I suggest that `You may learn either:\n%{skill1}\nor\n%{skill2}` into `The hero may learn either:\n%{skill1}\nor\n%{skill2}` To keep it consistent with the rest of the game.

There are two separate buttons and each of the button corresponds each skill:
image

I think it's better to mention only 1 skill for each corresponding button.
The only thing I would improve here is the texts alignment and headers. But since "hero" button also has same issue, I believe, it could be fixed in a separate PR for all three buttons.

ihhub:
Hi @Branikolog , let's focus here on what was the original issue.

Hi @Mr-Bajs , I would like to keep the changes minimal as the proposed change is not related to the original problem.

PR #9480, opened 2025-01-22
Show Artifact description window for clicking on it in the Editor

ihhub:
as it is still hard to remember what a particular artifact does.

image


Branikolog (approved):

Branikolog:
Hi, @ihhub !

A little offtop:

I've noticed that artifact icon info completely duplicates the initial information of the window:

https://github.com/user-attachments/assets/1e91617b-e89c-427a-b013-f04068a6cbdd

Such behaviour occurs not only for the implemented feature, but for all similar cases when we doulbe click over skills or artifacts inside a hero window. Do we need such useless hints for the icon which is already has same description at the current window? :)

ihhub:
@Branikolog , yes, we do.

PR #9482, opened 2025-01-23
Add description about 50% retaliation damage to the Blind spell

### LINKED ISSUE #9460: Spell book, halving retaliation is missing in the Blind spell description
### Preliminary checks

Blinded creature retaliates with only 50% of damage.

Image

 It would be nice to mention in the spell description that creature retaliates with half of its normal damage, so players clearly understand the difference between Paralyze and Blind.

### Additional info

_No response_

zenseii:
Fix #9460 

image


src/fheroes2/battle/battle_troop.cpp
```diff
@@ -516,8 +516,8 @@ uint32_t Battle::Unit::EstimateRetaliatoryDamage( const uint32_t damageTaken ) c
 
     const uint32_t retaliatoryDamage = unitsLeft * damagePerUnit;
 
-    // The retaliatory damage of a blinded unit is halved
-    return ( Modes( SP_BLIND ) ? retaliatoryDamage / 2 : retaliatoryDamage );
+    // The retaliatory damage of a blinded unit is reduced
+    return ( Modes( SP_BLIND ) ? retaliatoryDamage / ( 100 / Spell( Spell::BLIND ).ExtraValue() ) : retaliatoryDamage );
```
oleg-derevenetz, line 520: This formula looks a bit odd. Suppose that `ExtraValue` is 90 instead of 50 and `retaliatoryDamage` is 100. In this case, the resulting damage should be 10, right? But in your case it's `100 / ( 100 / 90 ) = 100`, because `100 / 90 = 1` (since it's all the integer operations) and `100 / 1 = 100`.

src/fheroes2/battle/battle_troop.cpp
```diff
@@ -565,7 +565,7 @@ uint32_t Battle::Unit::CalculateDamageUnit( const Unit & enemy, double dmg ) con
         // Petrified units cannot attack, respectively, there should be no retaliation
         assert( !enemy.Modes( SP_STONE ) );
 
-        dmg /= 2;
+        dmg *= ( Spell( Spell::BLIND ).ExtraValue() / 100.0 );
```
oleg-derevenetz, line 568: I suggest to avoid floating-point operations when it's not necessary (and it's not necessary there). In addition, this is also wrong, because when, for example, `dmg` is 100 and `ExtraValue` is 90, the result will be 90, but should be 10. I suppose that "reduces the damage when retaliating by %{count} percent" means that `%{count}` percent of the original value should be substracted from the original value, right?

oleg-derevenetz (changes requested):
Hi @zenseii I just left a few comments, could you please take a look when you have time?

Branikolog (approved):
Thanks for improving the description, @zenseii !

oleg-derevenetz (approved):
LGTM. I just added a few assertions to verify that the reduction percent will never accidentally exceed 100%.

ihhub (approved):

ihhub:
@zenseii , many thanks for this improvement!

PR #9485, opened 2025-01-23
Shorten hero name if name plus level is too long

zenseii:
Fix the hero + level being too long for the info box. We shorten the hero name and keep the level in case this is an enemy hero and we've cast Identify Hero. Knowing the level is more useful than the name because regardless of stats it can indicate other things.

Problem:
<details>

image
</details>

This PR:

image

Reported on Discord

Current translations are compatible but they need to be updated. I suggest doing this after the merge and after the bot has updated the POs.

### DIFF
--- a/src/fheroes2/dialog/dialog_quickinfo.cpp
+++ b/src/fheroes2/dialog/dialog_quickinfo.cpp
@@ -718,19 +718,22 @@ namespace
 
         const bool isActiveHero = ( activeHero != nullptr );
 
-        std::string message;
         // hero's name
+        const fheroes2::FontType smallWhite = fheroes2::FontType::smallWhite();
+        std::string heroName = hero.GetName();
+        fheroes2::Text text( heroName, smallWhite );
+        // hero's level
         if ( isFullInfo && isActiveHero ) {
-            message = _( "%{name} (Level %{level})" );
-            StringReplace( message, "%{name}", hero.GetName() );
-            StringReplace( message, "%{level}", activeHero->GetLevel() );
-        }
-        else {
-            message = hero.GetName();
+            std::string heroLevel = _( "heroQuickInfo|(Level %{level})" );
+            StringReplace( heroLevel, "%{level}", activeHero->GetLevel() );
+            const fheroes2::Text levelText( heroLevel, smallWhite );
+            // if Identify Hero has been cast then we want to know the hero's level rather than name.
+            const uint32_t boxShadowAndBorder = 39;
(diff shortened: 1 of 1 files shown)


src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -718,19 +718,21 @@ namespace
...
         if ( isFullInfo && isActiveHero ) {
-            message = _( "%{name} (Level %{level})" );
-            StringReplace( message, "%{name}", hero.GetName() );
-            StringReplace( message, "%{level}", activeHero->GetLevel() );
-        }
-        else {
-            message = hero.GetName();
+            std::string heroLevel = _( "heroQuickInfo|(Level %{level})" );
+            StringReplace( heroLevel, "%{level}", activeHero->GetLevel() );
+            const fheroes2::Text levelText( heroLevel, smallWhite );
```
ihhub, line 728: Since we include an additional space we should update `heroLevel` variable before passing to `fheroes2::Text` constructor:
```cpp
heroLevel.insert( 0, " " );
```

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -718,19 +718,21 @@ namespace
...
+            // if Identify Hero has been cast then we want to know the hero's level rather than name.
+            const int32_t boxShadowAndBorder = 39;
+            text.fitToOneRow( box.width() - boxShadowAndBorder - levelText.width() );
```
ihhub, line 731: To shorten the lifetime of the variable and also to avoid using extra memory we can do this way:
```cpp
text.fitToOneRow( box.width() - boxShadowAndBorder - fheroes2::Text{ heroLevel, smallWhite }.width() );
```

src/fheroes2/dialog/dialog_quickinfo.cpp
```diff
@@ -718,19 +718,21 @@ namespace
...
+            const std::string fittedText = text.text() + " " + heroLevel;
+            text.set( fittedText, smallWhite );
```
ihhub, line 733: We should do in one of 2 ways:
```cpp
std::string fittedText = text.text() + " " + heroLevel;
text.set( std::move( fittedText ), smallWhite );
```
or
```
text.set( text.text() + " " + heroLevel, smallWhite );
```
to avoid memory duplication.

ihhub (changes requested):
H @zenseii , I put few small comments here. Could you please take a look at them?

ihhub (approved):

ihhub:

PR #9492, opened 2025-01-26
[emscripten] allow override INCOMING_MODULE_JS_API

ololoken:
bump emscripten sdk to 4.0.1; 
fix NO_ALLOW_BLOCKING_ON_MAIN_THREAD => ALLOW_BLOCKING_ON_MAIN_THREAD=0; 
use -Oz in release build => 6.2mb wasm output instead of 9.7mb


.github/workflows/make.yml
```diff
@@ -335,7 +335,7 @@ jobs:
           package_name: fheroes2_emscripten_mt.zip
     name: Make (${{ matrix.name }})
     runs-on: ubuntu-latest
-    container: emscripten/emsdk:3.1.74
+    container: emscripten/emsdk:4.0.1
```
oleg-derevenetz, line 338: I use the `emsdk:latest` in #9453 and suggest to do the same here.

.github/workflows/make.yml
```diff
@@ -348,51 +348,10 @@ jobs:
...
-        +
-             build_dir = ports.clear_project_build('sdl2_mixer')
-             include_path = os.path.join(source_path, 'include')
-             includes = [
-        EOF
     - name: Build
       if: ${{ contains( matrix.on, github.event_name ) }}
       run: |
-        emmake make -f Makefile.emscripten -j "$(nproc)"
+        LDFLAGS="-sINCOMING_MODULE_JS_API=canvas,preRun,setStatus" emmake make -f Makefile.emscripten -j "$(nproc)"
```
oleg-derevenetz, line 354: In general, I'm not against this, but I suggest to move the `LDFLAGS` to `env`.

src/dist/Makefile
```diff
@@ -29,7 +29,7 @@ TARGET := fheroes2
 #
 
 # Common flags for both C and C++ compilers
-CCFLAGS := -fsigned-char -pthread
+CCFLAGS := $(CCFLAGS) -fsigned-char -pthread
```
oleg-derevenetz, line 32: What is the purpose of this change? `CCFLAGS` is not defined in the [GNU Coding Standards](https://www.gnu.org/prep/standards/standards.html) (unlike `CFLAGS`, `LDFLAGS` and so on), it's our own extension for internal use, and should not be defined externally.

src/dist/Makefile.emscripten
```diff
@@ -51,20 +50,19 @@ ifdef FHEROES2_WITH_DEBUG
 CCFLAGS := $(CCFLAGS) -gsource-map
 LDFLAGS := $(LDFLAGS) -O0 -gsource-map
 else
-CCFLAGS := $(CCFLAGS) -flto
-LDFLAGS := $(LDFLAGS) -O3 -flto
+CCFLAGS := $(filter-out -O3,$(CCFLAGS)) -Oz -flto
+LDFLAGS := $(LDFLAGS) -Oz -flto
```
oleg-derevenetz, line 54: Please remove this change. According to the [emcc documentation](https://emscripten.org/docs/tools_reference/emcc.html) `-O3` is focused on speed, while `-Oz` is focused on code size (in other words, it will not perform an optimization if it leads to an increase in code size, even if it leads to faster performance). This change can lead to performance degradation on slower machines, and should be introduced (and tested) separately.

src/dist/Makefile.emscripten
```diff
@@ -51,20 +50,19 @@ ifdef FHEROES2_WITH_DEBUG
...
-	-sALLOW_BLOCKING_ON_MAIN_THREAD \
 	-sENVIRONMENT=web,worker \
 	-sPTHREAD_POOL_SIZE=8
 else
 CCFLAGS := $(filter-out -pthread,$(CCFLAGS))
 
 LDFLAGS := $(filter-out -pthread,$(LDFLAGS)) \
 	-sENVIRONMENT=web \
-	-sNO_ALLOW_BLOCKING_ON_MAIN_THREAD
+	-sALLOW_BLOCKING_ON_MAIN_THREAD=0
```
oleg-derevenetz, line 67: What is the purpose of this change? `-sOPTION` is an equivalent to `-sOPTION=1` and `-sNO_OPTION` is an equivalent to `-sOPTION=0`, this is the documented feature. Since we already use the abbreviated forms (e.g. `-sASYNCIFY` and `-sNO_DISABLE_EXCEPTION_CATCHING`) we should use the abbreviated form here as well for consistency.

oleg-derevenetz (changes requested):
Hi @ololoken I just left a few notes, could you please take a look when you have time?

.github/workflows/make.yml
ololoken: bad idea, I'm sure there is not need to explain why.

src/dist/Makefile.emscripten
ololoken: Ok, but I prefer to use 0/1 toggles if option gets changes depending on env variables.

.github/workflows/make.yml
oleg-derevenetz: > bad idea, I'm sure there is not need to explain why.

1. No one will bother to bump this in future by hand;
2. If there are problems with a new emsdk version, they should be detected and fixed in time. That's the whole point of CI. We should not focus on some specific emsdk version, which then no one will bump.

src/dist/Makefile.emscripten
oleg-derevenetz: > To run wasm version you need machine capable to run modern browser, it's performance is more than enough in any case. But download size and initial wasm module initialisation depends on binary size.

How about the future operation of the application? We must first check that this change really will not cause any regression, especially on somewhat outdated hardware. Please roll back this change, it can be made later separately.

src/dist/Makefile.emscripten
```diff
@@ -51,14 +50,13 @@ ifdef FHEROES2_WITH_DEBUG
...
-CCFLAGS := $(CCFLAGS) -flto
-LDFLAGS := $(LDFLAGS) -O3 -flto
+CCFLAGS := $(filter-out -O3,$(CCFLAGS)) -Oz -flto
+LDFLAGS := $(LDFLAGS) -Oz -flto
 endif
 
 ifdef FHEROES2_WITH_THREADS
 LDFLAGS := $(LDFLAGS) \
 	-Wno-pthreads-mem-growth \
-	-sALLOW_BLOCKING_ON_MAIN_THREAD \
```
oleg-derevenetz, line 61: Why did you removed this line?

.github/workflows/make.yml
ololoken: 1. Why anybody will need to bump version at all if there are no pending sdk issues?
2. Right, point of CI is to build own code changes without unexpected drifts in dependencies and toolchains. As emscripten is far from stable state (especially in js api output) it better to lock sdk version.

src/dist/Makefile.emscripten
ololoken: mmm, -Oz applied only in case of emscripten build, no?

.github/workflows/make.yml
oleg-derevenetz: > Why anybody will need to bump version at all if there are no pending sdk issues?

Because other developers who want to work on this project will not install emsdk versions from half a year ago (or more). And if they install the latest emsdk and will not be able to build or run this app, they might think "what an outdated shit" (and they'd be right!) and don't want to work with this project at all.

> Right, point of CI is to build own code changes without unexpected drifts in dependencies and toolchains.

Nope, the point of CI to detect new issues in time.

src/dist/Makefile.emscripten
oleg-derevenetz: > mmm, -Oz applied only in case of emscripten build, no?

Yes, so what? This is a different optimization level that may cause regressions in a browser. We should verify first that there are no regressions.

src/dist/Makefile.emscripten
ololoken: Option is connected to pthread feature and can be changed in future, so it's better to leave default value.

oleg-derevenetz: fheroes2 currently _does_ block on the main thread, and there is a promise that this parameter "may become set to 0 by default in the future". If it will, then the app will crash (even if in reality this does not cause any issues). Please revert this change.

src/dist/Makefile.emscripten
ololoken: Trust me, there are no regression, only profits.

oleg-derevenetz: > Trust me

Sorry, this is not enough :) You even don't bother to check this, because "the user machine can't be slow because it should be capable to run modern browser". Let people really compare working on slow machines with and without this change. Once again - please revert this change. I will not approve this PR with this change anyway.

.github/workflows/make.yml
ololoken: There is no need to install "outdated" sdk, just use appropriate tool, use docker. The rest is such a far-fetched situation.

oleg-derevenetz: I'm sorry, but I've already seen a lot of projects that have problems building with modern versions of libraries and/or SDKs because they are locked to outdated versions of their dependencies. Especially a lot of this shit is written for Android. Once again:

1. This project should be able to be built and run with the latest versions of its dependencies;
2. No one will periodically bump the emsdk version by hand in future to verify this.

ololoken: Same situation, I saw tones of projects broken due to lack of version lock of dependencies. Especially a lot of shit made for web.

src/dist/Makefile.emscripten
ololoken: Right, this default value will be changed when it will be safe, so this is not an issue.

src/dist/Makefile.emscripten
ololoken: Are you seriously compare hardware requirement for modern browser and fheroes?

.github/workflows/make.yml
oleg-derevenetz: > Same situation, I saw tones of projects broken due to lack of version lock of dependencies. Especially a lot of shit made for web.

That's why issues with new dependencies should be detected and fixed in time, rather than a few months or years later, when you urgently need to upgrade to a new version of a dependency - for instance, because the old one contains a security hole. I am categorically against locking on a specific versions of dependencies, and this is not done for other platforms.

src/dist/Makefile.emscripten
oleg-derevenetz: > Right, this default value will be changed when it will be safe, so this is not an issue.

I don't think that this is "not an issue", and don't think that "it will be changed when it will be safe". Please revert this change.

src/dist/Makefile.emscripten
ihhub: @ololoken , we must rely on facts in development, not feelings. Any version of fheroes2 engine must be optimized no matter where it runs. As @oleg-derevenetz mentioned we support low performance machines such as Raspberry Pi and the engine must tun there smoothly.

oleg-derevenetz: Hi @ihhub BTW, here are two interesting videos. Here is the fheroes2 (demo version build) running in Chrome:

https://github.com/user-attachments/assets/35d614ac-a017-4ba4-8e77-45819ee6c92a

And this is the same demo version build running on the same device in Firefox:

https://github.com/user-attachments/assets/7c69989f-70c0-4618-b950-57e57db523e1

Can you guess the characteristics of this device? :) Here are they:

https://www.devicespecifications.com/en/model/a55332e4

It's the Android tablet with 2GB RAM(!) which was released in 2015, 10 years ago. And fheroes2 still works on this device in a browser, and it is quite possible to play it. At the same time, these "modern browsers" themselves are very slow on "modern websites". It's quite difficult to run a video player in it, for example. So yes, fheroes2 still should work on seriously outdated hardware, I consider this a sign of good taste. Modern frontenders don't care much about this :)

ihhub: That's cool. Any software must be optimized and not eat tons of RAM and CPU for no reason. This is what we are trying to do here.

ihhub:
@ololoken , I would to hear an explanation about https://yandex.ru/games/#app=432818 and especially the part why the information about the project and who contributed to it was blatantly removed.

FIY @oleg-derevenetz , @Districh-ru , @zenseii , @Branikolog and @idshibanov

ololoken:
Hey, initially it was experiment if it's possible to deploy such games to yagames at all. I disabled credits screen as there was links to 3rd party resources, that's disallowed by rules of platform. Now I can remove the game, or redeploy a version with enabled credits, in that case unfortunately links anyway must be hidden.

ihhub:
Hi @ololoken , thanks for the reply!

We do not restrict anyone to distribute the engine. There are several distributions of the engine done by other people on other platforms and we appreciate their effort. However, to apprise the effort of everyone who contributed to the project and also have a proper feedback form, it is extremely important to keep the credits and all the links and authors of the project to keep players know the origin of the project and who made this happen.

Secondly, the version published in the market violates Ubisoft copyright as it distributes the original game resources (aka piracy) and we would like not to do this nor to be associated with such an approach.

Thirdly, we haven't released fheroes2 in emscripter for a reason: it is not stable enough and some features were not working. This is to make sure that the quality of every single release of the engine is up to high standards and expectations from players.

Therefore, we have 2 ways:
- remove the game from the store (the easiest way)
- keep the game but comply with copyright issues (remove the usage of the original resources and only give a way to upload them from user's machines), bring back all functionality of the engine including full credits and the Editor, clearly state the origin of the engine and **links** to the official project web page (this is much harder way and this is how every distribution of the engine does)

ololoken:
ok, let's start from option b, if it fails, I will remove the game from platform. Will post update on Monday (23.06)

ihhub:
@ololoken , thank you!

ololoken:
@ihhub update:
1. full game functionality restored.
2. game data now is downloaded as standalone package from 3rd party resource, it's not a part of build anymore. I'm still considering way to add functionality to allow users upload own resources and bypass moderation.

There is also option to restrict access to the app for countries with strict regulations, to avoid problems related to intellectual property.

ihhub:
Hi @ololoken , I can see that there are some changes being made. However, there are still concerns which haven't been addressed:
- the engine **must** not be supplied / downloaded / uploaded with any form of original game resources. The only thing which you can do is to download resources from the publicly available demo version of the game. Any other ways are piracy and the team does not want to be associated with any sort of piracy or copyright issues
- the description of the game does not include any details about the project
- the title of the game should not be used as the original game. fheroes2 is not HoMM2: The Price of Loyalty. It is misleading

ololoken:
no worries I'm working on it
image
for now I revoked publication

ihhub:
@ololoken ,

You didn't fix anything. Take down this entry in the store now.

PR #9498, opened 2025-01-28
Increase maximum allowed map name length from the original 17 characters to 50

ihhub:
relates to #8589

image

image

### DIFF
--- a/src/fheroes2/editor/editor_save_map_window.cpp
+++ b/src/fheroes2/editor/editor_save_map_window.cpp
@@ -388,6 +388,7 @@ namespace Editor
                     mapName = std::move( editableMapName );
                     mapNameText.set( mapName, fheroes2::FontType::normalWhite(), language );
                     mapNameBackground.restore();
+                    mapNameText.fitToOneRow( mapNameRoi.width );
                     mapNameText.drawInRoi( mapNameRoi.x, mapNameRoi.y + 4, mapNameRoi.width, display, mapNameRoi );
 
                     display.render( mapNameRoi );
--- a/src/fheroes2/editor/editor_save_map_window.cpp
+++ b/src/fheroes2/editor/editor_save_map_window.cpp
@@ -445,7 +445,7 @@ namespace Editor
                 fheroes2::showStandardTextMessage( _( "Open Virtual Keyboard" ), _( "Click to open the Virtual Keyboard dialog." ), Dialog::ZERO );
             }
             else if ( le.isMouseRightButtonPressedInArea( mapNameRoi ) ) {
-                fheroes2::showStandardTextMessage( _( "Map Name" ), _( "Click to change your map name." ), Dialog::ZERO );
+                fheroes2::showStandardTextMessage( _( "Map Name" ), '\"' + mapName + "\"\n\n" + _( "Click to change your map name." ), Dialog::ZERO );
             }
 
             // Text input cursor blink.
--- a/src/fheroes2/editor/editor_save_map_window.cpp
+++ b/src/fheroes2/editor/editor_save_map_window.cpp
@@ -241,6 +241,7 @@ namespace Editor
         background.applyTextBackgroundShading( mapNameRoi );
         fheroes2::ImageRestorer mapNameBackground( display, mapNameRoi.x, mapNameRoi.y, mapNameRoi.width, mapNameRoi.height );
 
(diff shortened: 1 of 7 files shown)


ihhub:
Hi @zenseii , I've addressed your comment. Could you please take a look again?

ihhub:
Hi @zenseii , you are absolutely right! We actually have 2 places with the limit and one of them was in the map save dialog.

Branikolog:
Hi, @ihhub 
The issue found by @zenseii is still there.
Is it going to be fixed later in a separate PR?

Districh-ru (commented):
Hi, @ihhub,
LGTM, but I have one suggestion.
Since now we may have map names longer than can be displayed I suggest to show the full map name in the right mouse button press dialog, in example:
```cpp
fheroes2::showStandardTextMessage( mapName, _( "Click to change your map name." ), Dialog::ZERO );
```
or
```cpp
fheroes2::showStandardTextMessage( _( "Map Name" ), '\"' + mapName + "\"\n\n" + _( "Click to change your map name." ), Dialog::ZERO );
```
изображение

What do you think?

Also we can do such change to the load map dialog.

Branikolog:
Hi, @ihhub 
The issue with cutting map name is fixed now. 👍 
However, I've noticed that in some cases symbols are partially covering field edges:
image

[1x.zip](https://github.com/user-attachments/files/18600659/1x.zip)

src/fheroes2/editor/editor_save_map_window.cpp
```diff
@@ -241,6 +241,7 @@ namespace Editor
         background.applyTextBackgroundShading( mapNameRoi );
         fheroes2::ImageRestorer mapNameBackground( display, mapNameRoi.x, mapNameRoi.y, mapNameRoi.width, mapNameRoi.height );
 
+        mapNameText.fitToOneRow( mapNameRoi.width );
```
Districh-ru, line 244: Hi, @ihhub, to address the issue mentioned by @Branikolog we can reduce the maximum text width here:
```suggestion
        mapNameText.fitToOneRow( mapNameRoi.width - 6 );
```

src/fheroes2/editor/editor_save_map_window.cpp
```diff
@@ -389,6 +388,7 @@ namespace Editor
                     mapName = std::move( editableMapName );
                     mapNameText.set( mapName, fheroes2::FontType::normalWhite(), language );
                     mapNameBackground.restore();
+                    mapNameText.fitToOneRow( mapNameRoi.width );
```
Districh-ru, line 391: And here:
```suggestion
                    mapNameText.fitToOneRow( mapNameRoi.width - 6 );
```

Districh-ru (approved):

Branikolog:
@ihhub 

Hi! All mentioned issues are fixed! 👍 

Since we have enough space in scenario selection dialog, are we going to show longer name in this window?
image

Branikolog:
@ihhub 
I've also noticed, that information window for the map name does not show Cyrillic symbols for this particular case:
image

ihhub:
Hi @Branikolog , non-Latin letters should be properly displayed now.

Branikolog:
Hi, @zenseii !

Well, probably, expanding it to the left does not look really pleasant, but it still looks better, than empty space to the left, while showing three dots of some hidden text to the right:
image

Sure, when we have some new button text should be centered in that window, but for now expanding text to the right in a case is too long looks acceptable for me:

image

I suggest preserving the central alignment for the shorter names, but in a case name becomes too long, the whole map name should slightly move to left until it fills the entire bar and only then add three dots to the right.

ihhub:
> Hi, @zenseii !
> 
> Well, probably, expanding it to the left does not look really pleasant, but it still looks better, than empty space to the left, while showing three dots of some hidden text to the right: image
> 
> Sure, when we have some new button text should be centered in that window, but for now expanding text to the right in a case is too long looks acceptable for me:
> 
> image
> 
> I suggest preserving the central alignment for the shorter names, but in a case name becomes too long, the whole map name should slightly move to left until it fills the entire bar and only then add three dots to the right.

It is fixed.

ihhub:
@Branikolog , I am merging this pull request as we need to move forward with the release. Please use master branch for further testing of this feature.

PR #9499, opened 2025-01-28
Use a single button to switch between castle and town in the Editor dialog

### LINKED ISSUE #9478: Make Town and Castle buttons work like Min and Max buttons
### Preliminary checks

To avoid this problem:

Image

The fixed width of the buttons don't facilitate translations, but making both buttons text-adaptable won't work as there won't be enough room unless we widen the whole dialog.

I suggest to solve this as easy as possible (hopefully) by having the two buttons overlap and work like the Min Max buttons and the Standard/Campaign button in the high score screen.

I'm not sure how this will look if the buttons have different widths so perhaps we need to wait until we add a feature that two or more buttons will get the same width as the widest text on the buttons. That feature is necessary for many other menus.

### Additional info

_No response_

ihhub:
close #9478

image

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -1824,7 +1824,16 @@ namespace
         case ICN::BUTTON_CASTLE_EVIL: {
             _icnVsSprite[id].resize( 2 );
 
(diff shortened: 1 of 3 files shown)


ihhub:
Hi @zenseii , I've addressed your comments. Please take a look again.

Branikolog:
Hi, @ihhub !

I've noticed, that the shadow of the central button is way darker, than other buttons:
image

ihhub:
Hi @Branikolog , it should be fixed now.

Branikolog (approved):
👌

PR #9500, opened 2025-01-28
Make "Week of Monster" deterministic

### LINKED ISSUE #9439: Creatures are placed on different places each reloading "The month of XXX""
### Preliminary checks

As far as I understand, all creatures during "The month of" are generated according to the random seed of the particular machine and overall actions on the adventure map. I've noticed, that when I reload the same autosave at the day 7 and end turn (without any actions) all creatures appear on absolutely different places. All AI heroes seem to act absolutely the same during this turn. Even if I create a map without any AI player, unicorns tend to be placed randomly near the particular object of interest (at least in this map they appear near the same object though).

https://github.com/user-attachments/assets/5ea2540b-0cda-4c62-973c-03d389b5076b

Current behaviour seems to not correspond the idea of whole game repetitive outcome on reloading the same save file.

### Additional info

Save files for month of unicorn from the video attached and another save from actual gameplay, where trolls are generated completely random near different objects.
[Month of save.zip](https://github.com/user-attachments/files/18385204/Month.of.save.zip)

ihhub:
close #9439

### DIFF
--- a/src/fheroes2/world/world.cpp
+++ b/src/fheroes2/world/world.cpp
@@ -29,6 +29,7 @@
 #include <limits>
 #include <optional>
 #include <ostream>
+#include <random>
 #include <set>
 #include <tuple>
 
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (approved):

Branikolog (approved):
👍

PR #9510, opened 2025-01-29
Speed up loading MP2 maps by reading map info structure at once

ihhub:
In theory this approach is faster than just reading from a file many times. However, since my machine has a very good SSD I do not see any difference in performance especially when files are cached by the system. This should be tested on low end SSD or HDD devices to verify any performance impact.

relates to #6681

### DIFF
--- a/src/fheroes2/maps/maps_fileinfo.cpp
+++ b/src/fheroes2/maps/maps_fileinfo.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2024                                             *
+ *   Copyright (C) 2019 - 2025                                             *
  *                                                                         *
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub I really don't think that these changes will lead to better performance because of the page cache. If we don't pass flags like `O_DIRECT` when opening a file (and we do not), then all the I/O will be buffered, i.e. it will go through the page cache, which means that the minimal I/O size will be the page size (i.e. even if you read just a single byte, the whole page size will be in fact read - if it does not exceed the file size of course, and it's usually measured in kbytes, which is more than `MP2_MAP_INFO_SIZE`) and there also will be some OS-dependent read-ahead policy as well.

ihhub:
Hi @oleg-derevenetz , I was getting a hint about this this is why I wrote about caching.

shprotru:
Hello, @ihhub! Take a video:

https://github.com/user-attachments/assets/09171e24-49b2-4144-9ba0-1efb6d23aa52

Issue #2109, opened 2020-10-27
File Options window, additional action buttons and button layout


LeHerosInconnu:
I have already made suggestions for adding new action buttons to the File Options window:

- Add a "RESTART" button which allows to relaunch a scenario quickly: https://github.com/ihhub/fheroes2/issues/1528.
- Add a "QUICK SAVE" button which allows to save a scenario quickly: https://github.com/ihhub/fheroes2/issues/1529.

I suggest adding additional action buttons to facilitate access to the game save files:

- Add a "QUICK LOAD" button which allows to load the last quick save made by the player.
There is no need to select a file, the last quick save file is loaded directly.
- Add a "LOAD AUTOSAVE" button which allows to load the last autosave of the game.
There is no need to select a file, the last autosave file is loaded directly.

Layout of buttons in the window:

"NEW GAME" "RESTART"
"LOAD GAME" "SAVE GAME"
"QUICK LOAD" "QUICK SAVE" 
"LOAD AUTOSAVE" "QUIT"
"CANCEL"

In fheroes2:

![File Options window 01](https://user-images.githubusercontent.com/43583994/92213456-ddfa6800-ee93-11ea-826d-ada52b3086f4.png)

Branikolog:
I'd rather prefer to use quicksave and quickload only with keys on keyboard...
Adding so much buttons to this window wouldn't look so good... That's my personal point. )

LeHerosInconnu:
Hello @Branikolog,

> I'd rather prefer to use quicksave and quickload only with keys on keyboard...
> Adding so much buttons to this window wouldn't look so good... That's my personal point. )

All users are different, personally I prefer to do everything with the mouse because I already have my hand on it and it's simpler and more natural this way.

You also have to think about the future, and not all devices necessarily have a keyboard.

It can also be an option: "Game: Show extended interface for the "File Options" window".

PR #9512, opened 2025-01-29
Add return to main menu and play map buttons to the editor

### LINKED ISSUE #8720: Editor: Exit option should redirect back to game main menu
### Preliminary checks

EXIT button on editor menu popup (the one with NEW / LOAD / SAVE map) closes the application completely.
Since the editor is now a part of the game application, EXIT option could just redirect back to game main menu.

EXIT button in editor redirects to game MAIN MENU.

### Additional info

version 1.1.0 on Android 5.1.1

zenseii:
This adds a button to return to the main menu and a button to go straight to the map setup dialog.
The button to start the map was mainly added to keep symmetry of the dialog. 

image

I also added a method to the StandardWindow class that restores the display to before the standard window was drawn. This is used for the save dialog both in the editor and the adventure map.

Fix #8720


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1158,35 +1185,53 @@ namespace Interface
...
+            if ( le.MouseClickLeft( buttonPlayMap.area() ) ) {
+                // The map either hasn't been saved to a file or there aren't any human-playable colors.
+                if ( !conf.getCurrentMapInfo().name.empty() && conf.getCurrentMapInfo().colorsAvailableForHumans > 0 ) {
+                    if ( fheroes2::showStandardTextMessage( _( "Play Map" ), _( "Do you wish to leave the editor and play the current map?" ), Dialog::YES | Dialog::NO )
+                         == Dialog::YES ) {
+                        return fheroes2::GameMode::NEW_STANDARD;
+                    }
+                }
+                else {
+                    fheroes2::showStandardTextMessage( _( "Invalid Map" ), _( "This map is not valid. Try to save the latest changes." ), Dialog::OK );
```
Districh-ru, line 1230: Should we say about the need of 1 or more players on the map?
In example: "This map is not valid. The map should have at least one human player and be saved on your storage."

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1158,35 +1185,53 @@ namespace Interface
...
+                if ( fheroes2::showStandardTextMessage( _( "Main Menu" ), _( "Do you wish to return to the game's Main Menu? All unsaved changes will be lost." ),
+                                                        Dialog::YES | Dialog::NO )
+                     == Dialog::YES ) {
+                    return fheroes2::GameMode::MAIN_MENU;
+                }
+            }
+            if ( le.MouseClickLeft( buttonPlayMap.area() ) ) {
+                // The map either hasn't been saved to a file or there aren't any human-playable colors.
+                if ( !conf.getCurrentMapInfo().name.empty() && conf.getCurrentMapInfo().colorsAvailableForHumans > 0 ) {
+                    if ( fheroes2::showStandardTextMessage( _( "Play Map" ), _( "Do you wish to leave the editor and play the current map?" ), Dialog::YES | Dialog::NO )
```
Districh-ru, line 1224: Should we mention that all unsaved changes will be lost or allow to open the Save Map dialog?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1203,16 +1248,14 @@ namespace Interface
             else if ( le.isMouseRightButtonPressedInArea( buttonQuit.area() ) ) {
                 fheroes2::showStandardTextMessage( _( "Quit" ), _( "Quit out of the map editor." ), Dialog::ZERO );
             }
+            else if ( le.isMouseRightButtonPressedInArea( buttonMainMenu.area() ) ) {
+                fheroes2::showStandardTextMessage( _( "Main Menu" ), _( "Return to the game's Main Menu." ), Dialog::ZERO );
+            }
```
Districh-ru, line 1253: It would be nice if we add such dialog for the PLAY MAP button :)

Districh-ru (commented):
Hi, @zenseii, I left several suggestions. Could you please take a look when you have time?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1120,35 +1120,62 @@ namespace Interface
...
+        const bool isEvilInterface = conf.isEvilInterfaceEnabled();
+
+        fheroes2::Sprite mainMenuReleased;
+        fheroes2::Sprite mainMenuPressed;
+        fheroes2::getTextAdaptedButton( mainMenuReleased, mainMenuPressed, gettext_noop( "MAIN\nMENU" ),
+                                        isEvilInterface ? ICN::EMPTY_EVIL_BUTTON : ICN::EMPTY_GOOD_BUTTON, isEvilInterface ? ICN::STONEBAK_EVIL : ICN::STONEBAK );
+        fheroes2::Sprite playMapReleased;
+        fheroes2::Sprite playMapPressed;
+        fheroes2::getTextAdaptedButton( playMapReleased, playMapPressed, gettext_noop( "PLAY\nMAP" ), isEvilInterface ? ICN::EMPTY_EVIL_BUTTON : ICN::EMPTY_GOOD_BUTTON,
+                                        isEvilInterface ? ICN::STONEBAK_EVIL : ICN::STONEBAK );
```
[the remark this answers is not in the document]
ihhub: We want to avoid polluting `agg_file.cpp` file as it became too big to maintain. Having a function to do the work without increasing the number of ICN entries is a better solution.

@zenseii , I believe `fheroes2::getTextAdaptedButton()` function should be renamed into `fheroes2::getTextAdaptedButtonSprites()` function as it doesn't generate a button but only creates Sprites.

Going forward, we can create `fheroes2::getTextAdaptedButton()` function which will return `fheroes2::ButtonSprite` object so we can use it here. What do you think?

src/fheroes2/editor/editor_interface.cpp
ihhub, line 1224: Should `Editor` start from a capital letter?

src/fheroes2/editor/editor_interface.cpp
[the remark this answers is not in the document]
Districh-ru: So then it should be better to remove all other buttons from `agg_image.cpp` used in this dialog and generate them here to have the same code and to avoid questions like "Why this button image is made this way and that button - the other way?" ;)

Districh-ru:
> @Districh-ru. I noticed that when the Play Map button is used and you arrive at the scenario info dialog, if you then press cancel the screen will flash as if a fadeout was triggered. Are fade out's somehow cued? I'm not entirely sure where to look to solve this.

@zenseii, I've fixed this issue - added a check when rendering the Select Scenario dialog that also resets the need of fade effect.
Previously it was not needed because there was no way to open this dialog bypassing the main menu. :)

Districh-ru:
> > Looks nice. But would it look better if **Load Map** and **Save Map** where next to each other in the same row or column? Not that it really matters.
> 
> Hi, @Mr-Bajs. I wanted to mimic as much as possible how the original buttons were placed but I'm really impartial to it. I guess Main Menu should be where it is and quit should be lower right, the rest is the same for me.

I'd also prefer to swap **PLAY MAP** and **SAVE MAP** so there the upper row will be for the map manipulations (new/load/save) and the lower row for the quit from Editor (to new game/main menu/OS).

src/fheroes2/editor/editor_interface.cpp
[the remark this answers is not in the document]
ihhub: Hi @Districh-ru , not all buttons from `agg_file.cpp` file can be removed easily. Some of them have logic for button generation, others support original resources.

ihhub:
I am marking this pull request as high priority as it is a must for the upcoming release.

Branikolog:
Hi, @zenseii !

New menu looks great! 👍 👍 👍 

It would be nice to add an info hint on right click for the new "Play Map" button too.

I'm also slightly confused by the term "Invalid map" in this window:
image

It's a nice touch to warn a player, that he cannot play unsaved map, but in my personal opinion we shouldn't call this map as "invalid" as it misleads to the fact, that something is wrong with the map itself. Maybe it's better to point that user cannot play an unsaved map.

I also prefer Save and Load buttons to be placed on current different levels, as players can accidentally press wrong button and spoil his map. 👍 

I've also noticed, that gaps between buttons are not same:
image

src/fheroes2/editor/editor_interface.cpp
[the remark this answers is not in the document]
Districh-ru: Sure, let's deal with buttons generation in the other PR.

Branikolog:
@zenseii 

Current button placement is fine for me.

> We could say "Unplayable map" because any map that has changes that aren't saved is unplayable currently,

In my opinion there should be two separate cases, when the saved map cannot be played, because of incorrect game conditions (Invalid is ok then) and when the changes just haven't been saved yet.
For the last case there could be a message referring to unsaved process of the mapmaking, something like:
"You cannot play the map, until you save all changes. Save the current progress and try again."

LeHerosInconnu:
Hello everyone,

The button layout could be done this way:
Image from the screenshot of the original post, for explanation only.
So that the buttons are positioned "in pairs" (vertically from left to right).
- "NEW MAP" "PLAY MAP"
- "LOAD MAP" "SAVE MAP"
- "MAIN MENU" "QUIT".

Button layout 01 001

What do you think of "LAUNCH MAP" to replace "PLAY MAP"?

Related issue: https://github.com/ihhub/fheroes2/issues/2109 (for button layout in pairs).

LeHerosInconnu:
Hello @zenseii,

Yes, "START MAP" would be more appropriate.

Districh-ru:
> I'm Ok with both "Play" or "Start".

Hi, @zenseii, I'm also fine with any of these two variants. For the right click on this button we can display, in example, : `Leave the Editor and play the current map in Standard Game mode. However, all unsaved changes will be lost.`

src/fheroes2/gui/ui_window.cpp
```diff
@@ -487,4 +487,14 @@ namespace fheroes2
             }
         }
     }
+
+    void StandardWindow::restoreBackground()
+    {
+        _restorer.restore();
+        Display & display = Display::instance();
```
ihhub, line 494: I suggest not to do rendering within this method as the same logic is done inside the destructor `~StandardWindow()`. I can see that after every call of `restoreBackground()` method we actually return from a function implicitly calling the destructor of this class. This will ensure that we don't render the same frame twice. By changing this method we can end up having only one line of code here so we can make this method inlined within the header file.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1186,36 +1190,75 @@ namespace Interface
...
+            }
+            if ( le.MouseClickLeft( buttonMainMenu.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::EDITOR_TO_GAME_MAIN_MENU ) ) {
+                if ( fheroes2::showStandardTextMessage( _( "Main Menu" ),
+                                                        _( "Do you wish to return to the game's Main Menu? (Any unsaved changes to the current map will be lost.)" ),
+                                                        Dialog::YES | Dialog::NO )
+                     == Dialog::YES ) {
+                    return fheroes2::GameMode::MAIN_MENU;
+                }
+            }
+            if ( le.MouseClickLeft( buttonPlayMap.area() ) ) {
```
ihhub, line 1227: Let's add `else since the above condition will never be executed together with this condition:
```cpp
else if ( le.MouseClickLeft( buttonPlayMap.area() ) ) {
```

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1186,36 +1190,75 @@ namespace Interface
...
+                                                      _( "Do you wish to leave the Editor and start the map? (Any unsaved changes to the current map will be lost.)" ),
+                                                      Dialog::YES | Dialog::NO )
+                         == Dialog::YES ) {
+                        return fheroes2::GameMode::NEW_STANDARD;
+                    }
                 }
             }
-            else if ( le.MouseClickLeft( buttonCancel.area() ) || Game::HotKeyCloseWindow() ) {
-                break;
+            if ( le.MouseClickLeft( buttonCancel.area() ) || Game::HotKeyCloseWindow() ) {
```
ihhub, line 1260: Let's add `else since the above condition will never be executed together with this condition:
```cpp
else if ( le.MouseClickLeft( buttonCancel.area() ) || Game::HotKeyCloseWindow() ) {
```

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1186,36 +1190,75 @@ namespace Interface
...
+                        Get().saveMapToFile();
+                        isNameEmpty = conf.getCurrentMapInfo().name.empty();
+                        if ( isNameEmpty ) {
+                            display.render( background.totalArea() );
+                            continue;
+                        }
+                    }
+                    else {
+                        continue;
+                    }
```
ihhub, line 1244: To avoid complicated branching in the code we can do the opposite for the if-condition:
```cpp
if ( fheroes2::showStandardTextMessage( _( "Unsaved Changes" ),
     _( "This map has either terrain changes, undo history or has not yet been saved to a file.\n\nDo you wish to save the current map?" ), Dialog::YES | Dialog::NO ) == Dialog::NO ) {
    continue;
}

Get().saveMapToFile();
isNameEmpty = conf.getCurrentMapInfo().name.empty();
if ( isNameEmpty ) {
    display.render( background.totalArea() );
    continue;
}
````

**Review by ihhub — CHANGES_REQUESTED** [review]
Hi @zenseii , I put few comments here. Would you mind to take a look at them?

**Review by Branikolog — APPROVED** [review]
Perfect!

File: src/fheroes2/editor/editor_interface.cpp [hunk 9]
```diff
@@ -1186,36 +1190,71 @@ namespace Interface
...
+                if ( fheroes2::showStandardTextMessage(
+                         _( "Unsaved Changes" ),
+                         _( "This map has either terrain changes, undo history or has not yet been saved to a file.\n\nDo you wish to save the current map?" ),
+                         Dialog::YES | Dialog::NO )
+                     == Dialog::NO ) {
+                    continue;
+                }
+
+                Get().saveMapToFile();
+                bool isNameEmpty = conf.getCurrentMapInfo().name.empty();
```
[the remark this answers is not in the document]
ihhub: It is a valid suggestion.

ihhub (approved):

ihhub:
Hi @zenseii , could you please address clang-tidy and clang-format complaints?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1186,36 +1190,71 @@ namespace Interface
...
+                                                        _( "Do you wish to return to the game's Main Menu? (Any unsaved changes to the current map will be lost.)" ),
+                                                        Dialog::YES | Dialog::NO )
+                     == Dialog::YES ) {
+                    return fheroes2::GameMode::MAIN_MENU;
+                }
+            }
+            else if ( le.MouseClickLeft( buttonPlayMap.area() ) ) {
+                if ( fheroes2::showStandardTextMessage(
+                         _( "Unsaved Changes" ),
+                         _( "This map has either terrain changes, undo history or has not yet been saved to a file.\n\nDo you wish to save the current map?" ),
```
Districh-ru, line 1230: This dialog is displayed eve if I just saved the map.
Possibly we should leave a TODO to show it only if the map has changes from its saved version

src/fheroes2/heroes/heroes_dialog.cpp
```diff
@@ -127,8 +127,8 @@ int Heroes::OpenDialog( const bool readonly, const bool fade, const bool disable
     // Dialog title.
     const fheroes2::Rect titleRoi( dialogRoi.x + 60, dialogRoi.y + 1, 519, 17 );
 
-    auto drawTitleText = [&display, &titleRoi, &dialogRoi, &backgroundImage, this]( const std::string & heroName, const int heroRace, const bool restoreBackground ) {
-        if ( restoreBackground ) {
+    auto drawTitleText = [&display, &titleRoi, &dialogRoi, &backgroundImage, this]( const std::string & heroName, const int heroRace, const bool hideWindow ) {
+        if ( hideWindow ) {
```
Districh-ru: IMHO this change (and the same in castle.h and castle_well.cpp) should be reverted because this bool is to restore background of the part of the windows to render a text over it.

Districh-ru (approved):
Hi @zenseii , well done!
I just left one optioal comment to leave a TODO to show massage about warning for saving the map only if there are changes in the map (or possibly we can change the text).

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Or we can rephrase the text:
This map may have either terrain changes, undo history or has not yet been saved to a file

Districh-ru: @zenseii, yes, this dialog always appears after I press the SAVE MAP button.
I load a map, do no changes to it, open the menu and press START MAP button and this dialog appears :)


https://github.com/user-attachments/assets/6778db9b-d520-437a-b31b-39272c3f0bde

Districh-ru: > I see what happened. It was after we did the opposite check. I think we can solve it by checking for the isNameChanged, before showing the query dialog.

Sure, it'll be good for now.
Later we can expand this logic to check for changes in the map.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1186,36 +1190,70 @@ namespace Interface
...
+                            _( "This map has either terrain changes, undo history or has not yet been saved to a file.\n\nDo you wish to save the current map?" ),
+                            Dialog::YES | Dialog::NO )
+                            == Dialog::NO ) {
+                    continue;
+                }
+                if ( isNameEmpty ) {
+                    Get().saveMapToFile();
+                    isNameEmpty = conf.getCurrentMapInfo().name.empty();
+                }
+                if ( !isNameEmpty && conf.getCurrentMapInfo().colorsAvailableForHumans == 0 ) {
```
Districh-ru, line 1241: We can simply write:
```cpp
if ( isNameEmpty ) {
    continue;
}
```

instead of checking `!isNameEmpty` everytime.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1186,36 +1190,70 @@ namespace Interface
...
+                    fheroes2::showStandardTextMessage( _( "Unplayable Map" ),
+                                                       _( "This map is not playable. You need at least one human player for the map to be playable." ), Dialog::OK );
+                }
+                else if ( !isNameEmpty ) {
+                    if ( fheroes2::
```
Districh-ru, line 1246: To make SonarCloud happy we should combine such `is`s, but after we remove the `!isNameEmpty` check by the previous suggestion there will be only one condition left and all will be OK)

Districh-ru (approved):
@zenseii, now the save query dialog is shown for new (unnamed) maps, otherwise the other dialog informs that unsaved changes will be lost. 👍 

I left just one code readability optional suggestion above.

Districh-ru (approved):

ihhub:

Issue #1338, opened 2020-08-25
Additional message window with greater width


LeHerosInconnu:
Fheroes2 has remedied some flaws present in the original game regarding the information displayed in some windows.

For example, with the window displayed when a map event that gives more than two types of resources is activated.

In the original game:

![Additional message window with greater width 04](https://user-images.githubusercontent.com/43583994/91166813-3c288d80-e6d3-11ea-888f-d6ea90c56a69.png)

In fheroes2:

![Additional message window with greater width 01](https://user-images.githubusercontent.com/43583994/91166838-4480c880-e6d3-11ea-97e7-b7aece8a61d0.png)

Or, when the hero goes to the hill fort and there are more than two troops to upgrade.

In the original game:

![Additional message window with greater width 05](https://user-images.githubusercontent.com/43583994/91166874-53677b00-e6d3-11ea-8103-ba8c195b56e6.png)

In fheroes2:

![Additional message window with greater width 03](https://user-images.githubusercontent.com/43583994/91166886-58c4c580-e6d3-11ea-9c17-e675c483ee83.png)

However, as you can see, this is sometimes a problem.
The window displayed is not wide enough.

And if the text to display is very long, the top and bottom of the window are cut off.

![Additional message window with greater width 02](https://user-images.githubusercontent.com/43583994/91166909-5febd380-e6d3-11ea-86e8-e2f46ccd680a.png)

To remedy this, it would be necessary to have in addition a larger version of the message display window (or a version that adapts automatically) that would be used when these cases occur.

a1exsh:
Related: #1969

Branikolog:
Hill Fort message was already fixed long time ago.

PR #9514, opened 2025-01-30
Expand the functionality of the Adventure map event

### LINKED ISSUE #8731: Change Adventure Map event functionality to support extended format
### Preliminary checks

We made Adventure Map event format to have much more additional features than the existing event logic in the game. We need to add missing logic for it.

Besides just adding logic, we should also take care about serialization and backward compatibility of old saves.

### Additional info

_No response_

ihhub:
- the Adventure Map event now supports giving a Secondary Skill and also Experience
- the corresponding Editor dialog was reworked to add the new functionality
- many other places where we call a separate dialog to set amount of something were lacking needed images

image
image

close #8731

### DIFF
--- a/src/engine/image.h
+++ b/src/engine/image.h
@@ -188,11 +188,6 @@ namespace fheroes2
             _isRestored = true;
         }
 
-        Rect area() const
-        {
-            return { _x, _y, _width, _height };
-        }
-
     private:
         Image & _image;
         Image _copy;
--- a/src/engine/image.h
+++ b/src/engine/image.h
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2020 - 2024                                             *
+ *   Copyright (C) 2020 - 2025                                             *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
@@ -186,6 +186,11 @@ namespace fheroes2
         void reset()
         {
             _isRestored = true;
+        }
+
+        Rect area() const
+        {
+            return { _x, _y, _width, _height };
         }
 
     private:
--- a/src/fheroes2/editor/editor_event_details_window.cpp
+++ b/src/fheroes2/editor/editor_event_details_window.cpp
@@ -108,7 +108,7 @@ namespace Editor
         text.draw( messageRoi.x + ( messageRoi.width - text.width() ) / 2, offsetY, display );
 
         text.set( eventMetadata.message, fheroes2::FontType::normalWhite(), language );
-        text.draw( messageRoi.x + 5, messageRoi.y + 5, messageRoi.width - 10, display );
+        text.drawInRoi( messageRoi.x + 5, messageRoi.y + 5, messageRoi.width - 10, display, messageRoi );
 
         offsetY = startOffsetY;
         const int32_t secondColumnOffsetX = dialogRoi.x + sectionWidth + 2 * elementOffset;
@@ -351,7 +351,7 @@ namespace Editor
 
                     messageRoiRestorer.restore();
                     text.set( eventMetadata.message, fheroes2::FontType::normalWhite(), language );
-                    text.draw( messageRoi.x + 5, messageRoi.y + 5, messageRoi.width - 10, display );
+                    text.drawInRoi( messageRoi.x + 5, messageRoi.y + 5, messageRoi.width - 10, display, messageRoi );
                     isRedrawNeeded = true;
                 }
             }
@@ -419,7 +419,7 @@ namespace Editor
 
                 display.render( secondarySkillRoi );
             }
-            else if ( le.MouseClickLeft( experienceRoi ) ) {
+            else if ( le.MouseClickLeft( experienceRoiRestorer.area() ) ) {
                 const fheroes2::ExperienceDialogElement tempExperienceUI{ 0 };
                 int32_t tempValue{ eventMetadata.experience };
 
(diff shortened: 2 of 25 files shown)


ihhub:
Hi @zenseii , what do you think about this text: `Resources will be given as a reward.` ?

ihhub:
> Hi, @ihhub. This works very well. I have some comments. I believe we should check if the obtained artifact is a win condition and then trigger victory, just like we do for the sphinx.
> 
> Also we should change from `reward` to effect or event where I pointed out because the event can have negative effects.

Win condition is checked every time when a hero does an action.

ihhub:
Hi @zenseii , I think you refer to #1338

Branikolog:
Hi, @zenseii 

>I guess we all know it, but the best solution would probably be to make this type of window be able to extend in the width too.

Well, maybe this could work, but I think it's not possible to extend the frames (both evil and good) in these dimensions without help of the artist.

Showing several windows one by one is a common solution for the game currently. Like when you accept artifacts after the battle or lvlup for several levels in a single moment.

ihhub:
Hi @zenseii , I've changed the text for resources.

ihhub:
> Hi, @ihhub !
> 
> What if we add a special field for exp points and shift the star slightly to left so it does not look like a decoration, but an interactive element?
> 
> image
> 
> I do also suggest separating every gameplay aspect change into a separate window, as placing all together into a single window becomes too huge for the GUI:
> 
> image
> 
> I can't even press the button in this case. What if we show several windows in a row: text window, then resources window, then artifact window and after - skill window.

It is done.

ihhub:
Hi @zenseii  and @Branikolog , since we have debates about the dialog I decided to implement any solution for now and later we can decide what is the most suitable. Per my solution we are going to show 2 dialogs with the same text but different UI elements. All UI elements which are not fit for the first dialog are going to be displayed in the second.

ihhub:
Hi @zenseii , unfortunately this approach would lead to having 3 windows being shown. We also have other windows which are taller than the minimum resolution height. We can deal this issue outside this pull request.

Branikolog:
Hi, @ihhub 
I think it would be better to expand experience interaction area over the field with number too. Currently, I can click only on the experience icon, but clicking on the number shows no response. 
I'd also prefer shifting the experience block slightly to left:

image

But it's up to you. All three elements have different width, so placing them with same intervals could look not optimal... Maybe we should find the center of each element and align them according to the center of the sprite equally.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -2778,55 +2778,111 @@ namespace
...
+            // We have to split the message into 2 as it is too big.
+            std::vector<const fheroes2::DialogElement *> secondDialogUIElement;
+            while ( dialogHeight > displayHeight && !elementUI.empty() ) {
+                secondDialogUIElement.push_back( elementUI.back() );
+                elementUI.pop_back();
+                dialogHeight = fheroes2::getDialogHeight( emptyText, body, Dialog::OK, elementUI );
             }
+
+            fheroes2::showMessage( emptyText, body, Dialog::OK, elementUI );
+            fheroes2::showMessage( emptyText, body, Dialog::OK, secondDialogUIElement );
```
Districh-ru, line 2867: How about chowing only test in the first dialog and only UI elements in the second dialog?
For the case of very-very long texts the second dialog can be much higher than the first with the current code:
<details>

изображение


</details>

src/fheroes2/gui/ui_dialog.cpp
```diff
@@ -248,6 +248,75 @@ namespace fheroes2
...
+        if ( !rowHeight.empty() ) {
+            // UI elements are offset from the dialog body.
+            if ( bodyTextHeight > 0 ) {
+                elementHeight += textOffsetY;
+            }
+            elementHeight += textOffsetY;
+            elementHeight += rowHeight.back();
+        }
+
+        return overallTextHeight + elementHeight + ( ( buttons != 0 ) ? Dialog::FrameBox::getButtonAreaHeight() : 0 );
```
Districh-ru, line 317: Sadly, but `getDialogHeight()` returns only the dialog items (text + buttons) height excluding the dialog borders.
In example the next dialog height is 475:
<details>

изображение

</details>
We can try to fix it by taking border width into account:

```suggestion
        return overallTextHeight + elementHeight + ( ( buttons != 0 ) ? Dialog::FrameBox::getButtonAreaHeight() : 0 ) + fheroes2::borderWidthPx * 4;
```

The `* 4` is to show extra images on the dialog borders:
<details>

изображение

</details>

src/fheroes2/editor/editor_event_details_window.cpp
```diff
@@ -103,7 +110,79 @@ namespace Editor
         text.set( eventMetadata.message, fheroes2::FontType::normalWhite(), language );
         text.draw( messageRoi.x + 5, messageRoi.y + 5, messageRoi.width - 10, display );
```
Districh-ru, line 111: For the extra-long phrases we can do `drawInRoi()` to avoid rendering bugs:

<details>

изображение

</details>

```suggestion
        text.drawInRoi( messageRoi.x + 5, messageRoi.y + 5, messageRoi.width - 10, display, messageRoi );
```

Districh-ru (changes requested):
Hi, @ihhub, I left three comments regarding the case of very-very long event messages. Could please you take a look when you have time?

src/fheroes2/heroes/heroes_action.cpp
ihhub: Done.

ihhub:
Hi @Branikolog and @Districh-ru , I've addressed your comments. Could you please take a look?

src/engine/image.h
```diff
@@ -188,6 +188,11 @@ namespace fheroes2
             _isRestored = true;
         }
 
+        Rect area() const
+        {
+            return { _x, _y, _width, _height };
+        }
+
```
Districh-ru, line 195: This method duplicates the `rect()` method (lines 179-182).

Districh-ru (changes requested):
Hi, @ihhub, could you please check one issue I found in the code at first glance?

src/engine/image.h
ihhub: Removed.

Districh-ru (approved):

PR #9516, opened 2025-01-31
MonthOfMonstersAction(): make the generator dependent not just on the map seed, but also on the current month

oleg-derevenetz:
Follow-up to the #9500


ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for the improvement!

PR #9521, opened 2025-02-02
Add script to extract resources from CD images

Sarev0k:
I've verified that this extracts music and resources from my old Succession Wars and Price of Loyalty CDs.

### DIFF
--- a/script/homm2/extract_homm2_resources_from_cd_image.sh
+++ b/script/homm2/extract_homm2_resources_from_cd_image.sh
@@ -1,0 +1,166 @@
+#!/usr/bin/env bash
+
+###########################################################################
+#   fheroes2: https://github.com/ihhub/fheroes2                           #
+#   Copyright (C) 2025                                                    #
+#                                                                         #
+#   This program is free software; you can redistribute it and/or modify  #
+#   it under the terms of the GNU General Public License as published by  #
+#   the Free Software Foundation; either version 2 of the License, or     #
+#   (at your option) any later version.                                   #
+#                                                                         #
+#   This program is distributed in the hope that it will be useful,       #
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+#   GNU General Public License for more details.                          #
+#                                                                         #
+#   You should have received a copy of the GNU General Public License     #
+#   along with this program; if not, write to the                         #
+#   Free Software Foundation, Inc.,                                       #
+#   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             #
+###########################################################################
+
+set -e -o pipefail
+
+function echo_red {
+    echo -e "\033[0;31m$*\033[0m"
+}
+
+function echo_green {
+    echo -e "\033[0;32m$*\033[0m"
+}
+
+function echo_stage {
+    echo
+    echo_green "$*"
+    echo
+}
+
+echo_green "This script will extract and copy game resources from the original Heroes of Might and Magic II CD image."
+
+echo_stage "[1/4] determining the destination directory"
+
+DEST_PATH=""
+
+if [[ -n "$2" ]]; then
+    DEST_PATH="$2"
+elif [[ -f fheroes2 && -x fheroes2 ]]; then
+    DEST_PATH="."
+elif [[ -d ../../src ]]; then
+    # Special hack for developers running this script from the source tree
+    DEST_PATH="../.."
+fi
+
+if [[ -z "$DEST_PATH" || ( -d "$DEST_PATH" && ! -w "$DEST_PATH" ) ]]; then
+    if [[ "$(uname 2> /dev/null)" == "Linux" ]]; then
+        DEST_PATH="${XDG_CONFIG_HOME:-$HOME/.local/share}/fheroes2"
+    elif [[ -z "$2" ]]; then
+        DEST_PATH="$HOME/.fheroes2"
+    fi
+fi
+
+echo_green "Destination directory: $DEST_PATH"
+
+echo_stage "[2/4] determining the path to the HoMM2 CD image"
+
+if [[ "$#" -gt "0" ]]; then
+    HOMM2_CD_IMAGE_PATH="$1"
+else
+    read -e -p "Please enter the full path to the HoMM2 CD image (e.g. /home/user/homm2.cue or /home/user/homm2.iso): " HOMM2_CD_IMAGE_PATH
+fi
+
+STAGE_PATH=$(mktemp -d)
+trap 'rm -rf "$STAGE_PATH"' EXIT
+
+function stage_iso_assets {
+    if [[ -z "$(command -v fuseiso)" ]]; then
+        echo_red "fuseiso was not found in your system. Unable to mount CD image. Installation aborted."
+        exit 1
+    fi
+    if [[ -z "$(command -v fusermount)" ]]; then
+        echo_red "fusermount was not found in your system. Unable to unmount CD image. Installation aborted."
+        exit 1
+    fi
+
+    mkdir ROM
+    fuseiso "$1" ROM
+    trap 'fusermount -u ROM' RETURN
+
+    HOMM2_DIR=$(find ROM -type d -iname heroes2)
+    if [[ -z "$HOMM2_DIR" ]]; then
+        echo_red "Unable to locate 'heroes2' directory in CD image."
+        exit 1
+    fi
+
+    copy_directory "$HOMM2_DIR" "ANIM"
+    copy_directory "$HOMM2_DIR" "MAPS"
+    copy_directory "$HOMM2_DIR" "DATA"
+
+    fusermount -u ROM
+    trap - RETURN
+
+    rmdir ROM
+}
+
+function copy_directory() {
+    BASE_DIR=$1
+    DIR_NAME=$2
+
+    DIR_PATH=$(find "$BASE_DIR" -type d -iname "$DIR_NAME")
+    if [[ -n "$DIR_PATH" ]]; then
+        mkdir "$DIR_NAME"
+
+        cp -r "$DIR_PATH/." "$DIR_NAME"
+    fi
+}
+
+echo_stage "[3/4] staging game resources"
+cd "$STAGE_PATH"
+CD_IMAGE_FILE_NAME=$(basename "$HOMM2_CD_IMAGE_PATH")
+CD_IMAGE_EXTENSION=${CD_IMAGE_FILE_NAME##*.}
+if [[ "${CD_IMAGE_EXTENSION^^}" == "CUE" ]]; then
+    if [[ -z "$(command -v bchunk)" ]]; then
+        echo_red "bchunk was not found in your system. Unable to rip CD audio. Installation aborted."
+        exit 1
+    fi
+    # Parse the path of the bin file directly from the CUE sheet, so that this
+    # will continue to work even if the name of the cue sheet doesn't match the
+    # name of the bin file.
+    BIN_PATH="$(dirname "$HOMM2_CD_IMAGE_PATH")/$(cat "$HOMM2_CD_IMAGE_PATH" | grep FILE | grep -oE '"([^"]+)"' | tr -d '"')"
+    bchunk -w "$BIN_PATH" "$HOMM2_CD_IMAGE_PATH" tmp_ > /dev/null
+
+    if [[ -z "$(command -v flac)" ]]; then
+        echo_red "flac was not found in your system. Unable to reencode CD audio. Installation aborted."
+        exit 1
+    fi
+    mkdir MUSIC
+    for WAV_PATH in *.wav; do
+      # The music ripped from the CD starts from index 2, since the first index
+      # is occupied by the ISO data.  However, the engine expects the music to
+      # start from index zero, so we must subtract one from the index to meet
+      # this expectation.
+      WAV_FILE=$(basename "$WAV_PATH")
+      OLD_TRACK_NUM=$(echo "${WAV_FILE%.*}" | cut -d'_' -f2)
+      NEW_TRACK_NUM=$(printf '%02d' $((${OLD_TRACK_NUM#0}-1)))
+      NEW_FILE_NAME="homm2_$NEW_TRACK_NUM.flac"
+      # Flac generates a lot noise on the error stream, even when it's working
+      # as expected.  So we redirect the error output to quiet things down.
+      flac --delete-input-file -8 "$WAV_PATH" -o "MUSIC/$NEW_FILE_NAME" 2> /dev/null
+      echo_green "Completed ripping $NEW_FILE_NAME to 'MUSIC'."
+    done
+
+    stage_iso_assets "tmp_01.iso"
+    rm "tmp_01.iso"
+elif [[ "${CD_IMAGE_EXTENSION^^}" == "ISO" ]]; then
+    stage_iso_assets "$HOMM2_CD_IMAGE_PATH"
+else
+    echo_red "Unsupported CD image extension '$CD_IMAGE_EXTENSION'."
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


oleg-derevenetz (commented):
Hi @Sarev0k I just left a few questions regarding your PR, could you please take a look when you have time?

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,171 @@
+#!/bin/zsh
```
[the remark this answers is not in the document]
Sarev0k: My apologies, I had [this script](https://github.com/ihhub/fheroes2/blob/master/script/homm2/extract_homm2_resources_for_app_bundle.sh#L1) open at the time that I was using as an example.  I'll put out a new revision that migrates this to follow the example set in [this script](https://github.com/ihhub/fheroes2/blob/master/script/homm2/extract_homm2_resources.sh#L1).

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,171 @@
...
+
+###########################################################################
+#   fheroes2: https://github.com/ihhub/fheroes2                           #
+#   Copyright (C) 2021 - 2025                                             #
```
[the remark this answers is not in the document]
Sarev0k: I copied portions of [this script](https://github.com/ihhub/fheroes2/blob/master/script/homm2/extract_homm2_resources.sh#L5) that had a copyright header from 2021, so I was trying to preserve it.  Would you still like me to change the copyright header?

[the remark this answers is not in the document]
oleg-derevenetz: Well, since technically it's a new script anyway, I still recommend to se the copyright date to 2025 instead of the range. The range of years is used only for files that already existed in one way or another during the specified period.

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,171 @@
...
+    DIR_NAME=$2
+
+    DIR_PATH=$(find "$BASE_DIR" -type d -iname $DIR_NAME)
+    if [[ ! -z "$DIR_PATH" ]]; then
+        mkdir $DIR_NAME
+
+        for FILE_PATH in $DIR_PATH/*; do
+            # Depending on how the CD image was generated it's possible files
+            # will have lowercase names.  This is fine except the engine seems
+            # to expect data files to be uppercase.  We convert all file names
```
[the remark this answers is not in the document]
Sarev0k: When I leave the case sensitivity of the data directory alone, this dialog pops up when I try to start the engine.
Screenshot_20250202_080216

However, when I force the uppercase conversion this isn't the case.  From my testing in version 1.1.5, unless `HEROES2.AGG` is in upper case, that dialog will pop up.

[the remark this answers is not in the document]
oleg-derevenetz: I don't know by whom this dialog is displayed, but it's definitely not the fheroes2 engine. Most likely, the fheroes2 port or package does this in some way (by using some poorly designed startup script that "verifies prerequisites" in an incorrect way). fheroes2 engine itself doesn't have any issues with file name case - you can name this file `HEROES2.AGG`, `heroes2.agg` or even `HeRoEs2.aGg`, and it still should work just fine, even on the case-sensitive filesystem.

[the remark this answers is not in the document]
oleg-derevenetz: Of course, it's sad that such an incorrect check is being performed by something (most likely, some kind of auxiliary script as part of the fheroes2 package on this specific platform), and on the one hand, it might be worthwhile to forcibly change the filename case in order for this check to pass, but, on the other hand, I don't think we should adjust to such incorrect checks. IMHO these checks should be fixed by the package maintainer.

script/homm2/extract_homm2_resources_from_cd_image.sh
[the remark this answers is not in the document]
Sarev0k: Sure, I'll update this in the next revision.

script/homm2/extract_homm2_resources_from_cd_image.sh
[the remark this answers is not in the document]
Sarev0k: Thanks for the detailed explanation!

One reason I can think of that we may still want to keep this, is that it makes it simpler to run the script multiple times, once for each disk, and not have to worry about replicating two different copies of the same files across runs.

If you agree that forcefully manipulating the case would make sense for this use case, I can adjust the comment to explain that.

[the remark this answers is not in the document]
Sarev0k: Actually thinking this through more, there shouldn't be any overlap between the two disks.  I'll double check to verify this, and remove the filename manipulation if I can't find any potential conflicts.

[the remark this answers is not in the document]
Sarev0k: I'm definitely seeing some overlap between the original disk and expansion  for some maps and animations.  So it's potentially beneficial to force a case sensitivity between resource extraction assuming other folks could have ripped their images differently than I did.

I'll go ahead and remove the forcing  of case anyway though, since at least with my images, the case seems to match with the overlapping files.

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+fi
+
+echo_green "Destination directory: $DEST_PATH"
+
+echo_stage "[2/4] determining the path to the HoMM2 CD image"
+
+if [[ "$#" -gt "0" ]]; then
+    HOMM2_CD_IMAGE_PATH="$1"
+else
+    read "HOMM2_CD_IMAGE_PATH?Please enter the full path to the HoMM2 cd image (e.g. /home/user/homm2.cue or /home/user/homm2.iso): "
```
oleg-derevenetz, line 69: This syntax of the `read` command is valid for the `zsh`, but not for `bash`. I suggest to use the following syntax for `bash`:

```suggestion
    read -e -p "Please enter the full path to the HoMM2 cd image (e.g. /home/user/homm2.cue or /home/user/homm2.iso): " HOMM2_CD_IMAGE_PATH
```

`-e` allows `bash` to use the `readline` library to perform the file name completion using TAB character, which is handy when you need to enter the name of an existing file.

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+
+    HOMM2_DIR=$(find ROM -type d -iname heroes2)
+    if [[ -z "$HOMM2_DIR" ]]; then
+        echo_red "failed to locate heroes2 directory in cd image"
+        exit 1
+    fi
+
+    copy_directory $HOMM2_DIR "ANIM"
+    copy_directory $HOMM2_DIR "MAPS"
+    copy_directory $HOMM2_DIR "DATA"
```
oleg-derevenetz, line 97: In all the `copy_directory` calls above the `$HOMM_DIR` should be quoted (`"$HOMM_DIR"`) to avoid the word splitting or globing issues if it contains spaces or special characters like `*` or `?`.

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+    trap - RETURN
+
+    rmdir ROM
+}
+
+function copy_directory() {
+    BASE_DIR=$1
+    DIR_NAME=$2
+
+    DIR_PATH=$(find "$BASE_DIR" -type d -iname $DIR_NAME)
```
oleg-derevenetz, line 109:
```suggestion
    DIR_PATH=$(find "$BASE_DIR" -type d -iname "$DIR_NAME")
```

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+    if [[ ! -z "$DIR_PATH" ]]; then
```
oleg-derevenetz, line 110:
```suggestion
    if [[ -n "$DIR_PATH" ]]; then
```

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+        mkdir $DIR_NAME
```
oleg-derevenetz, line 111:
```suggestion
        mkdir "$DIR_NAME"
```

oleg-derevenetz (changes requested):
Hi @Sarev0k I left a few more comments - this time regarding the syntax of this script. Could you please take a look when you have time?

script/homm2/extract_homm2_resources_from_cd_image.sh
Sarev0k: Nice catch!  I'll fix this.

script/homm2/extract_homm2_resources_from_cd_image.sh
Sarev0k: Done

script/homm2/extract_homm2_resources_from_cd_image.sh
Sarev0k: Done.

script/homm2/extract_homm2_resources_from_cd_image.sh
Sarev0k: Done

script/homm2/extract_homm2_resources_from_cd_image.sh
Sarev0k: Done

oleg-derevenetz (approved):

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+    echo -e "\033[0;32m$*\033[0m"
+}
+
+function echo_stage {
+    echo
+    echo_green "$*"
+    echo
+}
+
+echo_green "This script will extract and copy game resources from the original Heroes of Might and Magic II CD images"
```
ihhub, line 39: `images` should be `image` as the script handles only one image file.

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+        exit 1
+    fi
+
+    mkdir ROM
+    fuseiso "$1" ROM
+    trap 'fusermount -u ROM' RETURN
+
+    HOMM2_DIR=$(find ROM -type d -iname heroes2)
+    if [[ -z "$HOMM2_DIR" ]]; then
+        echo_red "failed to locate heroes2 directory in cd image"
```
ihhub, line 91: Let's rephrase this sentence to match `extract_homm2_resources.sh` file:
```bash
"Unable to locate 'heroes2' directory in the CD image file."
```

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+fi
+
+echo_green "Destination directory: $DEST_PATH"
+
+echo_stage "[2/4] determining the path to the HoMM2 CD image"
+
+if [[ "$#" -gt "0" ]]; then
+    HOMM2_CD_IMAGE_PATH="$1"
+else
+    read -e -p "Please enter the full path to the HoMM2 cd image (e.g. /home/user/homm2.cue or /home/user/homm2.iso): " HOMM2_CD_IMAGE_PATH
```
ihhub, line 69: To be consistent with the rest of text: `cd image` --> `CD image`. The same comment for other places in the code below.

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+      # start from index zero, so we must subtract one from the index to meet
+      # this expectation.
+      WAV_FILE=$(basename "$WAV_PATH")
+      OLD_TRACK_NUM=$(echo "${WAV_FILE%.*}" | cut -d'_' -f2)
+      NEW_TRACK_NUM=$(printf '%02d' $((${OLD_TRACK_NUM#0}-1)))
+      NEW_FILE_NAME="homm2_$NEW_TRACK_NUM.flac"
+      # Flac generates a lot noise on the error stream, even when it's working
+      # as expected.  So we redirect the error output to quiet things down.
+      flac --delete-input-file -8 "$WAV_PATH" -o "MUSIC/$NEW_FILE_NAME" 2> /dev/null
+      echo_green "finished ripping $NEW_FILE_NAME to MUSIC"
```
ihhub, line 149: Let's be consistent and start every message from a capital letter if it is not an application name:
```
"Completed ripping $NEW_FILE_NAME into 'MUSIC'."
```

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+    fi
+}
+
+echo_stage "[3/4] staging game resources"
+cd "$STAGE_PATH"
+CD_IMAGE_FILE_NAME=$(basename "$HOMM2_CD_IMAGE_PATH")
+CD_IMAGE_EXTENSION=${CD_IMAGE_FILE_NAME##*.}
+if [[ "${CD_IMAGE_EXTENSION^^}" == "CUE" ]]; then
+    if [[ -z "$(command -v bchunk)" ]]; then
+        echo_red "bchunk was not found in your system. Unable to rip cd audio. Installation aborted."
```
ihhub, line 123: Let's follow the same way as the texts above:
```
bchunk was not found in your system. Unable to rip CD audio. Installation aborted.
```

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+        exit 1
+    fi
+    # Parse the path of the bin file directly from the CUE sheet, so that this
+    # will continue to work even if the name of the cue sheet doesn't match the
+    # name of the bin file.
+    BIN_PATH="$(dirname "$HOMM2_CD_IMAGE_PATH")/$(cat "$HOMM2_CD_IMAGE_PATH" | grep FILE | grep -oE '"([^"]+)"' | tr -d '"')"
+    bchunk -w "$BIN_PATH" "$HOMM2_CD_IMAGE_PATH" tmp_ > /dev/null
+
+    if [[ -z "$(command -v flac)" ]]; then
+        echo_red "flac was not found in your system. Unable to reencode cd audio. Installation aborted."
```
ihhub, line 133: Let's follow the same way as the texts above:
```
flac was not found in your system. Unable to convert CD audio. Installation aborted.
```

script/homm2/extract_homm2_resources_from_cd_image.sh
```diff
@@ -0,0 +1,166 @@
...
+      flac --delete-input-file -8 "$WAV_PATH" -o "MUSIC/$NEW_FILE_NAME" 2> /dev/null
+      echo_green "finished ripping $NEW_FILE_NAME to MUSIC"
+    done
+
+    stage_iso_assets "tmp_01.iso"
+    rm "tmp_01.iso"
+elif [[ "${CD_IMAGE_EXTENSION^^}" == "ISO" ]]; then
+    stage_iso_assets "$HOMM2_CD_IMAGE_PATH"
+else
+    echo_red "unsupported cd image extension $CD_IMAGE_EXTENSION."
```
ihhub, line 157: Let's be consistent and start every message from a capital letter if it is not an application name:
```
"Unsupported CD image extension '$CD_IMAGE_EXTENSION'."
```

ihhub (changes requested):
Hi @Sarev0k , I put several comments about the texts here. Could you please take a look at them?

ihhub (approved):

ihhub:
@Sarev0k , thank you so much for your contribution to the project!

Issue #9477, opened 2025-01-22
Expanding the button shine logic


zenseii:
### Preliminary checks

- [x] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

The current button shine logic for engine generated buttons is very bare-bones and therefore repetitive and easy to tell apart from the original buttons' shine effect.

Requirements:
The effect needs to be present on the sides and bottom side of buttons too for buttons that are big enough. Also the shine effect shouldn't be exactly the same for every button to keep variety as seen on original buttons.

### Describe the possible solution

Expand the function by:
1. Check the width and height of the button and from this determine if a button should have the effect on the sides and the bottom part and how many shines. The distinction will be between single lined text buttons and multilined ones.
2. To make the shine different between buttons of the same size, we can look at the button string value and make this value determine where and how many shines are added to the button. This makes the effect consistent for the same text and between loads of the engine.

### Additional info

Different types of shines might have to be added as images or functions that can be copied, or generated on the fly.

Issue #1843, opened 2020-09-30
Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, adjusting the position of the scroll bar and buttons


LeHerosInconnu:
I suggest a small modification for a better visual appearance for the scroll bar and buttons on the right at the bottom of the screen in order to leave a space of 1 pixel to the right of the scroll bar and buttons.

- move the block of scroll bar buttons and scroll bar by 1 pixel to the left.
- reduce the size of the frame of available creatures by 1 pixel on the right.
- move the "HEROES", "TOWN/CASTLES" and "EXIT" button block by 1 pixel to the left.
- reduce the dimension of the frame of mines and others by 1 pixel on the right.
- replace the yellow pixels at the bottom of the screen where the large vertical bar to the left of the "HEROES", "TOWN/CASTLES" and "EXIT" block joins the yellow frame of the screen.

![Kingdom Heroes visible frame H2 modif 02](https://user-images.githubusercontent.com/43583994/94682049-ac56ae80-0324-11eb-8861-d42426bcca03.png)

![Kingdom Heroes visible frame H2 modif coupé 02](https://user-images.githubusercontent.com/43583994/94682093-bbd5f780-0324-11eb-9dbf-57cb9867f382.png) ![Kingdom Heroes visible frame H2 modif ori coupé 02](https://user-images.githubusercontent.com/43583994/94682122-c7292300-0324-11eb-9f1b-f2eb98ce6c69.png)

![Kingdom Heroes visible frame H2 modif coupé 03](https://user-images.githubusercontent.com/43583994/94682135-cee8c780-0324-11eb-8bd7-66e1f673cc3e.png) ![Kingdom Heroes visible frame H2 modif ori coupé 03](https://user-images.githubusercontent.com/43583994/94683513-f5a7fd80-0326-11eb-83b2-eb1c486ad7b2.png)

Issue #1842, opened 2020-09-30
Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, wrong position of elements


LeHerosInconnu:
- Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, wrong horizontal position of column titles
- Kingdom Overview screen, Towns/Castles Overview screen, wrong horizontal and vertical position of town names
- Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, scroll bar incorrectly positioned horizontally
- Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, center horizontally all numbers of mines and all resource numbers
- Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, wrong vertical position of the "HEROES", "TOWNS/CASTLES" and "EXIT" buttons

**Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, wrong horizontal position of column titles**

In fheroes2, the column titles do not have the same horizontal position in their frame as in the original game.

In original game:

![Kingdom Heroes H2 01](https://user-images.githubusercontent.com/43583994/94677515-4fa3c580-031d-11eb-8f28-d67bf9c19e00.png)

In fheroes2:

![Kingdom Heroes fh2 01](https://user-images.githubusercontent.com/43583994/94677729-af01d580-031d-11eb-9c35-f050757ad786.png)

In original game:

![Kingdom Towns H2 01](https://user-images.githubusercontent.com/43583994/94677740-b5904d00-031d-11eb-83ee-575e4e32a4f1.png)

In fheroes2:

![Kingdom Towns fh2 01](https://user-images.githubusercontent.com/43583994/94677754-ba550100-031d-11eb-979a-3a7a0414c384.png)

Open the two images in different tabs in your browser, enlarge the images and switch between the two tabs alternately.
Heroes 2 first, fheroes2 second.

![Title column H2 01](https://user-images.githubusercontent.com/43583994/94677784-c640c300-031d-11eb-8d09-d6041f7c7095.png) ![Title column fh2 01](https://user-images.githubusercontent.com/43583994/94677819-d48edf00-031d-11eb-89a7-82bf93a8c563.png)

[...]

PR #9525, opened 2025-02-03
Rework button code

zenseii:
Adjusting to this change meant addressing a lot of long standing UI visual bugs.

What this PR does:

- Allows to generate buttons with fixed height and width. This is necessary for #9512.

- Allows to choose button background for fixed size generated buttons.

- Added palette conversion table for good buttons to evil buttons. This makes it easier than trying to make evil style buttons by hand for those button types that don't have that originally, and it fixed the wrong pressed state for map select button. This will also help to add evil style buttons for the editor.

- Improved button shine effect. Still not added for vertical buttons. It varies according to size of the button. Related to #9477  

- Switched to `createNormalButton()` in all but 1 set of buttons (vertical high scores buttons) in `agg_image.cpp`.

- Renamed `getTextAdaptedButton()` to `getTextAdaptedSprite()`

- Added black background so that we can use the smaller buttons near the well's status bar other places like in editor dialogs instead of the small buttons there.

- Added brown background to fix see through in the kingdom overview screen. Also fixed some wrong button placement in this dialog. Part of #1843 and #1842

- Added code to generate buttons that are shorter than 96px yet taller than 25px (needed for various dialogs like view world).

- Fix French min/max buttons transparency issues with M letter

- Generate French and Polish MIN and MAX buttons from their original assets.

- Remove all Italian specific button generations. The font used there is just a worse version of the one we generate.

- Removed Medium size empty button now that we can resize in height.

- Also a lot of button code tidying.

Other button info:

Text buttons that the engine needs to be able to make are these three:
1. fixed dimensions, provided by `getCustomNormalButton()` which is called by `makeButtonSprites()` in `ui_button.cpp` and `createNormalButton()` in `agg_image.cpp`. The only difference is that `makeButtonSprites()` doesn't translate text. Perhaps they could be combined.
2.  Text-adaptable dimensions, provided by `getTextAdaptedSprite()` in `ui_button.cpp`.
3. a group of buttons with equal dimensions, not provided yet but should be in `ui_button.cpp`.

Then there's a couple of calls to make/render buttons from various methods of `fheroes2::StandardWindow`. Ideally there should be a way to generate a window which adjusts its size according to the size of the buttons, to avoid any translations going outside of the window border.


Branikolog (approved):

Districh-ru (changes requested):
Hi @zenseii 
You did a great job!
Unfortunately the new button code has a little lower text width limit (maybe 1 pixel rounding issue) when generation some English buttons:
This PR:
изображение 
Master branch: 
изображение

The EXPANSION CAMPAIGN button issue is present in PR and master and can be fixed in the other PR: 
изображение

I was using Russian game assets to make engine generate English buttons.

Could you please check the CASTLES/TOWNS button issue when you have time.

Districh-ru:
> Hi, @Districh-ru. Thanks for looking at the PR. I must've remembered wrong then because I thought that button had the same issue on `master` when using any assets other than the English original. I will check to see with French assets switched to English. There are two points to be made about this:
> 
>     1. I find it unlikely that someone will use assets different from English when setting the language to English.
> 
>     2. There is a proposal in this issue [Kingdom Overview screen, Heroes Overview and Towns/Castles Overview screens, wrong position of elements #1842](https://github.com/ihhub/fheroes2/issues/1842) to extend the width of these buttons, which would anyways fix this problem if we're talking about 1 pixel.
> 
> 
> Personally I'm fine with leaving this as it is in this PR, but we could try to find a solution.

I agree that almost nobody will use non English assets with English language, but I mentioned this to show that there are differences in button text generation and the found issue may (it should not, but may :) ) blow up in some other translation.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1179,15 +1063,7 @@ namespace
...
-            for ( int32_t i = 0; i < static_cast<int32_t>( _icnVsSprite[id].size() ); ++i ) {
-                fheroes2::Sprite & out = _icnVsSprite[id][i];
-                out = fheroes2::AGG::GetICN( ICN::OVERVIEW, 2 + i );
-
-                // clean the button.
-                Fill( out, 6, 6 + i, 89 - i, 30, getButtonFillingColor( i == 0 ) );
-            }
-
-            renderTextOnButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "TOWNS/\nCASTLES" ), { 6, 5 }, { 5, 6 }, { 90, 34 }, fheroes2::FontColor::WHITE );
+            createNormalButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "TOWNS/\nCASTLES" ), false, ICN::BROWNBAK, { 89, 42 } );
```
Districh-ru, line 1066: Here is this one pixel difference. In `master build it is 90:
```suggestion
            createNormalButton( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "TOWNS/\nCASTLES" ), false, ICN::BROWNBAK, { 90, 42 } );
```

Districh-ru (approved):

ihhub (changes requested):
Hi @zenseii , I left 2 comments here. Would you mind to take a look at them?

ihhub (approved):

ihhub:
@zenseii , huge thanks for this tremendous amount of work!

PR #9534, opened 2025-02-06
Extract resources from Gold CD CAB

Sarev0k:
Instead of following the same pattern as Succession Wars and Price of Loyalty CDs, which maintained the maps and data resources in plain directories, the Gold CD stores these resources within a CAB file, extracted by the installer during installation.

This change extracts map and data resources from this CAB, when given a Gold CD image.

I've verified that this works as expected against the original Gold CD image.


ihhub:
@Sarev0k , thank you so much for this update!

PR #9557, opened 2025-02-11
Fix Editor buttons not being translated

### LINKED ISSUE #9542: Add translation support for strings in map editor's file dialog
### Preliminary checks

### Platform

Other

### Describe the bug

Relates to #9525 and #9512 

I realized we don't translate these buttons because `makeButtonSprites()` doesn't check supported font.
Trying to change to use the same system as in `getTextAdaptedSprites()` creates a mess in `ui_keyboard.cpp` because it needs to pass strings and not `const char *`.

I'm opening this issue to discuss a solution.

Currently I've only worked around it by adding this to `makeButtonSprites()`:
```cpp

const char * translatedText = _( text.c_str() );
const char * supportedText = fheroes2::isFontAvailable( translatedText, { fheroes2::FontSize::BUTTON_RELEASED, fontColor } ) ? translatedText : text.c_str();
```
One issue with this is that the shift button turns blank:

Image

### Save file

None needed

### Additional info

_No response_

ihhub:
close #9542

The solution is not perfect and the root cause of this case and any similar possible cases in the future is that a text translation is done implicitly for some functions while others don't. We need to avoid these confusion by removing translation code from any button generation code. Explicitly doing translation guarantees that everything works fine.


ihhub:
Hi @zenseii , could you please test this solution?

src/fheroes2/gui/ui_button.cpp
```diff
@@ -902,16 +902,33 @@ namespace fheroes2
...
+        const FontType releasedFont{ FontSize::BUTTON_RELEASED, fontColor };
+        const FontType pressedFont{ FontSize::BUTTON_PRESSED, fontColor };
+
+        const Text releasedText( text, releasedFont );
+        const Text pressedText( text, pressedFont );
 
-        const fheroes2::Text releasedText( text, { fheroes2::FontSize::BUTTON_RELEASED, fontColor } );
-        const fheroes2::Text pressedText( text, { fheroes2::FontSize::BUTTON_PRESSED, fontColor } );
+        const Size releasedTextSize( releasedText.width( buttonSize.width ), releasedText.height( buttonSize.width ) );
+        const Size pressedTextSize( pressedText.width( buttonSize.width ), pressedText.height( buttonSize.width ) );
```
Districh-ru: As i can see, the `width` field of the `Size` is never used and we call `.width( buttonSize.width )` for nothing.
So should we only call `...Text.height( buttonSize.width()` in the next two lines, what do you think?

src/fheroes2/gui/ui_button.cpp
```diff
@@ -891,8 +891,8 @@ namespace fheroes2
             addButtonShine( released, emptyButtonIcnID );
         }
 
-        const fheroes2::Size releasedTextSize( releasedText.width( textAreaWidth ), releasedText.height( textAreaWidth ) );
-        const fheroes2::Size pressedTextSize( pressedText.width( textAreaWidth ), pressedText.height( textAreaWidth ) );
+        const Size releasedTextSize( releasedText.width( textAreaWidth ), releasedText.height( textAreaWidth ) );
+        const Size pressedTextSize( pressedText.width( textAreaWidth ), pressedText.height( textAreaWidth ) );
```
Districh-ru: The same comment here.

Districh-ru (approved):
Hi @ihhub,
This PR work OK, but I found a couple places for optimization and pushed the changes:
I've removed the redundant calls to `.height( )` and also the calls to `.width( buttonSize.width )` which result was not used later and was stored for nothing.
Also a did a one small optimizationto cache `{textAreaWidth + backgroundBorders.width, textAreaHeight + backgroundBorders.height}`.

PR #9560, opened 2025-02-11
Add sort button and logic

OogaBoogaCode:
This is the reworked PR.

### DIFF
--- a/src/fheroes2/agg/agg_image.cpp
+++ b/src/fheroes2/agg/agg_image.cpp
@@ -399,17 +399,7 @@ namespace
         fheroes2::ApplyPalette( pressed, goodToEvilPalette );
     }
 
-    void createNormalButton( fheroes2::Sprite & released, fheroes2::Sprite & pressed, const char * untranslatedText, const bool isEvilInterface,
-                             const int backgroundIcnId, fheroes2::Size buttonSize )
-    {
-        // TODO: do not do translations for button generation. We shouldn't assume that we receive a non-translated string.
-        const fheroes2::FontColor buttonFontColor = isEvilInterface ? fheroes2::FontColor::GRAY : fheroes2::FontColor::WHITE;
-        const char * text = fheroes2::getSupportedText( untranslatedText, fheroes2::FontType{ fheroes2::FontSize::BUTTON_RELEASED, buttonFontColor } );
-
-        fheroes2::makeButtonSprites( released, pressed, text, buttonSize, isEvilInterface, backgroundIcnId );
-    }
-
-    void createCampaignButtonSet( const int campaignSetIcnId, const std::array<const char *, 5> & texts )
+    void createCampaignButtonSet( const int campaignSetIcnId, const std::array<const char *, 5> & untranslatedTexts )
     {
         int emptyButtonIcnID = 0;
         int buttonBackgroundICN = ICN::UNKNOWN;
@@ -432,9 +422,11 @@ namespace
             break;
         }
 
-        for ( size_t i = 0; i < texts.size(); ++i ) {
+        for ( size_t i = 0; i < untranslatedTexts.size(); ++i ) {
             const size_t icnIndex = 2 * i;
-            fheroes2::getTextAdaptedSprite( _icnVsSprite[campaignSetIcnId][icnIndex], _icnVsSprite[campaignSetIcnId][icnIndex + 1], texts[i], emptyButtonIcnID,
+
+            const char * translatedText = fheroes2::getSupportedText( untranslatedTexts[i], fheroes2::FontType::buttonReleasedWhite() );
+            fheroes2::getTextAdaptedSprite( _icnVsSprite[campaignSetIcnId][icnIndex], _icnVsSprite[campaignSetIcnId][icnIndex + 1], translatedText, emptyButtonIcnID,
                                             buttonBackgroundICN );
         }
     }
@@ -620,7 +612,8 @@ namespace
                 break;
             }
 
-            getTextAdaptedSprite( _icnVsSprite[id][0], _icnVsSprite[id][1], gettext_noop( "SELECT" ), ICN::EMPTY_MAP_SELECT_BUTTON, ICN::UNKNOWN );
+            const char * text = fheroes2::getSupportedText( gettext_noop( "SELECT" ), fheroes2::FontType::buttonReleasedWhite() );
+            getTextAdaptedSprite( _icnVsSprite[id][0], _icnVsSprite[id][1], text, ICN::EMPTY_MAP_SELECT_BUTTON, ICN::UNKNOWN );
 
             break;
         }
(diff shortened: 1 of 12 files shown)


src/fheroes2/system/settings.h
```diff
@@ -56,6 +56,12 @@ enum MusicSource
     MUSIC_EXTERNAL
 };
 
+enum SaveSortMode
```
ihhub, line 59: This should be class enumeration with no values:
```cpp
enum class SaveFileSortMethod : uint8_t
{
    SORT_BY_NAME,
    SORT_BY_DATE
};
```

src/fheroes2/system/settings.h
```diff
@@ -373,6 +385,7 @@ class Settings
 
     int sound_volume;
     int music_volume;
+    int save_sort_mode;
```
ihhub, line 388: Please use the correct camelCase variable naming convention:
```cpp
SaveFileSortMethod _saveFileSortMethod;
```
and since this is going to be `uint8_t` move it after `ZoomLevel _viewWorldZoomLevel{ ZoomLevel::ZoomLevel1 };`

Moreover, this member must have a default value:
```cpp
SaveFileSortMethod _saveFileSortMethod{ SaveFileSortMethod::SORT_BY_NAME };
```

src/fheroes2/system/settings.h
```diff
@@ -94,6 +100,11 @@ class Settings
         return _currentMapInfo;
     }
 
+    int SaveSortMode() const
```
ihhub, line 103: Please use the correct method naming:
```cpp
SaveFileSortMethod getSaveFileSortMethod() const
{
    return _saveFileSortMethod;
}
```

src/fheroes2/system/settings.h
```diff
@@ -257,6 +268,7 @@ class Settings
 
     void SetSoundVolume( int v );
     void SetMusicVolume( int v );
+    void SetSaveSortMode( int mode );
```
ihhub, line 271: This method should have the following signature:
```cpp
void setSaveFileSortMethod( const SaveFileSortMethod method )
{
    _saveFileSortMethod = method;
}
```

The method can be inlined since it is literally one line of code.

src/fheroes2/system/settings.cpp
```diff
@@ -390,6 +396,9 @@ std::string Settings::String() const
     os << std::endl << "# music volume: 0 - 10" << std::endl;
     os << "music volume = " << music_volume << std::endl;
 
+    os << std::endl << "# save sort mode: 1 - by name, 2 - by timestamp" << std::endl;
```
ihhub, line 399: We must not use any cryptic integer values and offer users more friend string approach:
```cpp
os << std::endl << "# file save sort method: name or date" << std::endl;
if ( _saveFileSortMethod == SaveFileSortMethod::SORT_BY_NAME ) {
    os << "file save sort method = name" << std::endl;
}
else {
    os << "file save sort method = date" << std::endl;
}
```

src/fheroes2/system/settings.cpp
```diff
@@ -190,6 +191,11 @@ bool Settings::Read( const std::string & filePath )
         SetMusicVolume( config.IntParams( "music volume" ) );
     }
 
+    // save sort mode
+    if ( config.Exists( "save sort mode" ) ) {
+        SetSaveSortMode( config.IntParams( "save sort mode" ) );
+    }
```
ihhub, line 197: Here we should handle strings based on the comment below. Also, please remove the comment as it brings more additional information.

src/fheroes2/gui/ui_window.h
```diff
@@ -89,6 +89,7 @@ namespace fheroes2
                                  const Padding padding );
         void renderButton( Button & button, const int icnId, const uint32_t releasedIndex, const uint32_t pressedIndex, const Point & offset, const Padding padding );
         void renderOkayCancelButtons( Button & buttonOk, Button & buttonCancel, const bool isEvilInterface );
+        void renderOkaySortCancelButtons( Button & buttonOk, Button & buttonSort, Point sortButtonOffset, Button & buttonCancel, const bool isEvilInterface );
```
ihhub, line 92: Please remove this method. Adding a new method which does exactly the same but for a very specific case is not a good design idea. Instead, you can use `renderButton()` method call for 3 buttons from your code.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -291,14 +297,13 @@ namespace
 
         const bool isEvilInterface = Settings::Get().isEvilInterfaceEnabled();
```
ihhub, line 298: Please cache `Settings::Get()` call and use it here and later in the newly added code.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -247,8 +247,14 @@ namespace
             }
         }
 
-        std::sort( mapInfos.begin(), mapInfos.end(), Maps::FileInfo::sortByFileName );
```
ihhub, line 250: `getSortedMapsFileInfoList()` function is literally used only in one place. We should just pass sorting method as an input parameter for this function.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -400,11 +410,21 @@ namespace
...
             buttonCancel.drawOnState( le.isMouseLeftButtonPressedInArea( buttonCancel.area() ) );
             if ( isEditing ) {
                 buttonVirtualKB->drawOnState( le.isMouseLeftButtonPressedInArea( buttonVirtualKB->area() ) );
             }
 
+            if ( le.MouseClickLeft( buttonSort.area() ) ) {
+                int const sortMode = Settings::Get().SaveSortMode();
+                Settings::Get().SetSaveSortMode( sortMode == 0 ? 1 : 0 );
+                lists = getSortedMapsFileInfoList();
+                listbox.SetListContent( lists );
```
ihhub, line 423: Do we need to call this function? What happens with a file selection? Does it preserved while switching between methods?

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -400,11 +410,21 @@ namespace
...
+                listbox.updateScrollBarImage();
+                listbox.Redraw();
+            }
+
             if ( le.MouseClickLeft( buttonCancel.area() ) || Game::HotKeyPressEvent( Game::HotKeyEvent::DEFAULT_CANCEL ) ) {
```
ihhub, line 428: Since we added the condition of mouse click above then we need to add else keyword here as we don't expect to execute the code.

ihhub (changes requested):
Hi @OogaBoogaCode , I left multiple comments here. Would you mind to take a look at them?

PR #9565, opened 2025-02-13
Avoid stacking units when defending archers

idshibanov:
If enemy army has units with area shot ability (Liches) AI should avoid stacking units to take extra damage when defending.


oleg-derevenetz:
Hi @idshibanov Maybe it's just not to use defensive tactics at all in such situations? It's quite difficult to cover friendly archers this way, because there will be holes vulnerable for flyers and melee units that are fast enough.

idshibanov:
> Hi @idshibanov Maybe it's just not to use defensive tactics at all in such situations? It's quite difficult to cover friendly archers this way, because there will be holes vulnerable for flyers and melee units that are fast enough.

Hello @oleg-derevenetz, I was curious to test it out and finally found some do so. 2 armies attacking 3 targets, 5 repeat battles with different seeds to account for battlefield random rolls (90 samples in total).

1. Walkers: (liches + hydras)
2. Normal: (liches as a part of regular necro army, no dragons)
3. Flyers: (liches with flyers)

Results:
- On master branch (no change): 
Sorcerer's: 65% casualties on average vs walkers, losing 100% to others
Barbarian's: 28% casualties vs walkers, 41.7% vs normal, 60% vs flyers

- No defensive tactics if area attack present (suggested): 
Sorcerer's: 50% casualties vs walkers, 46% vs normal, 90% vs flyers
Barbarian's: 17% casualties vs walkers, 22% vs normal, 41% vs flyers

- Defensive tactics but spread out (current change): 
Sorcerer's: 43% casualties vs walkers, 58% vs normal, 81% vs flyers
Barbarian's: 20% casualties vs walkers, 37% vs normal, 35% vs flyers

To me it's clear we have to account for area attack and either change improves the AI. Surprisingly, looks like spreading out units actually works better versus fliers. I think the defensive mode can lead to higher losses in some cases regardless of tactic that's why disabling it outright was optimal sometimes. We need to look at army compositions before turning the flag on.

**TLDR**: I prefer to leave the change as is and maybe make another PR for defensive mode improvements.

oleg-derevenetz (approved):
Hi @idshibanov I tuned the logic a bit to avoid situations like this:

https://github.com/user-attachments/assets/c2d81f7f-a7bb-44c8-80a2-f57f3843025b

With the latest changes it works like this:

https://github.com/user-attachments/assets/9e69729e-76f6-4964-bf98-dc2bec1d446e

Please take a look at the latest changes if possible, suggestions are welcome.

idshibanov:
Thanks for the clean up, I'm good to merge

idshibanov:
> @idshibanov and @oleg-derevenetz. Would it be sensible to add AoE spells like fireball, fireblast, cold ring and meteor shower to this logic in the same way how damaging spells are taken into account as "ranged" threat once the enemy has cast them?

Not a bad idea, but I think impact will be limited as you still would be able hit two stacks with one of these spells.

oleg-derevenetz:
> Would it be sensible to add AoE spells like fireball, fireblast, cold ring and meteor shower to this logic in the same way how damaging spells are taken into account as "ranged" threat once the enemy has cast them?

I suggest we do this later in a separate PR - if that makes sense, because mass damage spells, unlike the lich cloud, have a free aim, and they are more difficult to avoid.

ihhub (approved):

ihhub:
@idshibanov , thank you so much for this update!

PR #9567, opened 2025-02-14
Fix German original assets MIN button

zenseii:
This fixes a case where min button and max button's don't have the same shadow and are not at the same offset in dialogs when using German assets. This is due to #9525 where embedded shadows were removed from these buttons. The German Min button was forgotten.

The German min/max buttons are the same as the engine generated ones and equal to the English ones anyways.


ihhub (approved):

ihhub:

Issue #8105, opened 2023-12-04
AI misuses wirlpools


kitovyj:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Myra wants to attack Gem being just a few tiles away from her and uses a wirlpool instead of going around it.  That is very suboptimal in this case.

### Save file

[uses wirlpool wrong way.zip](https://github.com/ihhub/fheroes2/files/13549518/uses.wirlpool.wrong.way.zip)

oleg-derevenetz:
Technically the way through the Whirlpool is shorter - path through the Whirlpool takes ~700 movement points while path around the Whirlpool takes ~850 movement points, so, strictly speaking, this is not a bug. AI (in terms of pathfinding) always thinks that it will be lucky and it will always get into the right whirlpool, but this, of course, does not always happen. There are multiple ways to fix this, for example:

1. Let AI "cheat" with Whirlpools, just like it now "cheating" with Summon Boat spell, but with the Summon Boat it's "microcheat", while here it would be "real cheat";
2. Impose AI-only penalties for Whirlpools.

Each of these have its own "pros" and "cons". For now, I will reclassify this issue as a request for improvement.

kitovyj:
> Technically the way through the Whirlpool is shorter - path through the Whirlpool takes ~700 movement points while path around the Whirlpool takes ~850 movement points, so, strictly speaking, this is not a bug. AI (in terms of pathfinding) always thinks that it will be lucky and it will always get into the right whirlpool, but this, of course, does not always happen. There are multiple ways to fix this, for example:
> 
> 1. Let AI "cheat" with Whirlpools, just like it now "cheating" with Summon Boat spell, but with the Summon Boat it's "microcheat", while here it would be "real cheat";
> 2. Impose some AI-only penalties for Whirlpools.
> 
> Each of these have its own "pros" and "cons". For now, I will reclassify this issue as a request for improvement.

[...]

Issue #7139, opened 2023-05-08
Adventure map, AI should use spacebar to re-visit objects


Branikolog:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

To visit stone liths, for instance, AI has to move and then return back, while human players can press spacebar.

### Describe the possible solution

I think it would be fair, if AI hero can re-visit objects which he is standing on without wasting movement points or to not be stuck because of strong guardians near stone liths exit.

LeHerosInconnu:
Hello @Branikolog,

> I think it would be fair, if AI hero can re-visit objects which he is standing on without wasting movement points or to not be stuck because of strong guardians near stone liths exit.

I think the AI is going to freeze the game by using the stone liths in an endless movement loop... :D

Some scenarios in the original game are also based on the fact that the hero spends movement points to enter and exit stone liths.
In fheroes2, this free and infinite movement (between stone liths of the same type) for long distances and multiple different destinations easily chosen is in contradiction with the original game.
In the original game, it is not possible to choose the stone liths by which the hero will continue his travel without spending movement points by using multiple times stones liths of the same type.
Using the space bar for stone liths should consume movement points.

Related discussion: https://github.com/ihhub/fheroes2/discussions/5625.

oleg-derevenetz:
> I think the AI is going to freeze the game by using the stone liths in an endless movement loop... :D

Yes, exactly. The main reason why this is not done (except that there is no such mechanism in the original game) is that the number of hero movement points serves as a "control mechanism" that prevents any kind of looping due to the (potential) imperfections of AI logic.

Branikolog:
As a little offtop...
I was always hating stone liths operation in HoMM games, which forces a player to jump multiple times to reach the specific exit. (And forced to reload, in a case hero appeared in blocked exit or lost creatures in whirlpool)
I am about to suggest some day a new operating method of this object, where a player **can** select destination point. It could look alike to town portal window, where player can select a specific stone liths or choose some random (in a case he did not discovered all of exits). This would help to travel into target area without pressing space for several times. In the above particular map I could spend like a few minutes(!!) rolling for the exit I want which is inappropriate implementation in my opinion.
[...]

PR #9578, opened 2025-02-17
Allow AI to reuse teleporters that hero is standing on

idshibanov:
Relates to #7139. Relates to #8105 but doesn't fully resolve it.

Similarly to how a player may use the space bar to re-activate the stone liths AI should be able to reuse it. This is especially important on maps the exit is blocked on the other side that makes AI to play with one less hero (that is often impossible to reach). The feature is only enabled on Hard+ so normal campaign runs shouldn't be affected.


oleg-derevenetz:
Hi @idshibanov The main reason why I avoided implementing such logic is that the current AI implementation is prone to the cyclical repetition of the same actions - at least nothing prevents this in theory (and sometimes in practice). All other hero movement methods currently used by AI (including Dimension Door, Town Gate and Town Portal spells) spend the hero's movement points, which serves as a kind of safety valve, which ensures that the AI-controlled hero's turn ends eventually, even if the situation on the map is currently such that the AI is doing cyclically repeating actions. Reusing teleports without performing movements does not have such a safety valve, because hero doesn't spend any movement points, therefore, it is theoretically possible for a hero to move between two teleports indefinitely. What do you think?

idshibanov:
> Reusing teleports without performing movements does not have such a safety valve, because hero doesn't spend any movement points, therefore, it is theoretically possible for a hero to move between two teleporters indefinitely.

I understand the concern. I was thinking we could add a superficial movement cost (based on # of teleporter exits on the map) in the AI pathfinder to discourage AI from jumping back and forth. Planner cost only, no actual hero movement spent. Stone lith or whirlpool shouldn't be a destination point anyway.

There's also a discussion if we should make a new UI for players to select the exit portal. Giving player ability to use space bar to jump to any desired exit point defeated the purpose of randomness anyway. AI already knows which one.

idshibanov:
> In this case the strong Necro hero just stood in place on a stone lith and didn't move not matter how many days passed. This sort of behavior could be a problem for original maps and especially for the AI's ability to expand in the beginning.

This sounds like a bug/something I forgot to update. Will check it out later when got time. Thanks for testing!

PR #9584, opened 2025-02-19
Check the contents of the resurrection.h2d file for correctness

### LINKED ISSUE #9581: Add checks of the existence of a correct resurrection.h2d file
### Preliminary checks

Based on the #9580 case there are players who can forget to update resurrection.h2d file for a new fheroes2 engine version.

We would like to add check for existence of resurrection.h2d file and most importantly a correct version of it. We might need either to check presence of all needed files there or include versioning information.

### Additional info

_No response_

oleg-derevenetz:
close #9581


ihhub:
Hi @oleg-derevenetz , I am thinking to just update the format and add 4 byte integer stating a version. We will update the version only when changing the file. What do you think?

oleg-derevenetz:
Hi @ihhub I already thought about this and decided not to go this way:

* If we will auto-update the version each time the H2D file is updated, then this will be redundant if we just want to replace some existing file.
* If we will update the version manually, then we will regularly forget to do this.
* We will not be able to just decompose the H2D file into individual files and then recompose them back, we will need to do something with the version info.

The version info is fine when we have some user-created files (save files, maps, etc). In this case this looks redundant because H2D file is bundled with the engine an is an integral part of it, so we can just check that its contents is in sync with some sample. It will be difficult to forget to update something because it will be immediately detected at the first run.

ihhub (approved):

ihhub:

PR #9590, opened 2025-02-23
Move road placement logic directly to the map format

ihhub:
relates to #8589


ihhub:
Hi @Districh-ru , could you please take a look at the wrong road placement? I am not sure what is going on.

ihhub:
@Districh-ru , also I noticed that we still put terrain via `Maps::Tiles` and then save the info inside MapFormat. This leads to a fact that road placement does not properly update terrain but let's fix roads first.

Districh-ru:
Hi @ihhub, the roads placement and removal is fixed.
The terrain under the road is now also updated for the `world` tile.
I suggest to move terrain placement to the Map_Format in a new PR after this one is merged.

Could you please check the changes made when you have time.

Districh-ru (approved):

PR #9594, opened 2025-02-24
Fix the logic of transferring artifacts after a battle

### LINKED ISSUE #9592: Excess artifacts on a defeated hero should remain on that hero
### Preliminary checks

### Platform

Other

### Describe the bug

If a defeated hero has more artifacts than the winning hero can carry, then these artifacts should remain on the defeated hero if they are recruited.

This was reported by users as being the original behavior. I have not verified this myself.

### Save file

None provided. You could set up a map with 6 human playable knights with 1 castle each. Have two knight heroes fight on day 1 with both of them having lots of artifacts. According to the rule of guarantee of getting a same class hero day 1, this should make it easier to have the defeated hero become available.

### Additional info

There might be special rules for the ultimate artifacts concerning whether the defeated hero gets to keep that.

oleg-derevenetz:
fix #9592

If I understood correctly, this is exactly how it should work. Here is a video of the artifact transfer process (the source has been modified a bit to allow the artifact transfer on retreat or surrender):

https://github.com/user-attachments/assets/a3ffc057-92b0-45b4-8c76-36117e24208b

### DIFF
--- a/src/fheroes2/battle/battle_main.cpp
+++ b/src/fheroes2/battle/battle_main.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2024                                             *
+ *   Copyright (C) 2019 - 2025                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2010 by Andrey Afletdinov <fheroes2@gmail.com>          *
--- a/src/fheroes2/battle/battle_main.cpp
+++ b/src/fheroes2/battle/battle_main.cpp
@@ -61,7 +61,7 @@
 
 namespace
 {
-    bool isArtifactValidForTransfer( const Artifact & art )
+    bool isArtifactSuitableForTransfer( const Artifact & art )
     {
         return art.isValid() && art.GetID() != Artifact::MAGIC_BOOK;
     }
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
> From what I understand, the UA will never leave the defeated hero

According to my understanding, they either disappear or remain with the defeated hero. The winner never gets them. It works the same way here.

> I'm aware that in the original game the winning hero gets an empty inventory slot per UA on the defeated hero but this seems strange to me, which means that items are skipped.

Well, in this PR, empty slots for UA also remain in the winner's bag, just at the end, not in the middle. I don't think it's essential.

LeHerosInconnu:
Hello @zenseii and @oleg-derevenetz,

> Here is a video of the artifact transfer process (the source has been modified a bit to allow the artifact transfer on retreat or surrender):

It's only for demonstration purposes, right? :)

> @oleg-derevenetz . The logic looka correct apart from the transfer of ultimate artifacts. From what I understand, the UA will never leave the defeated hero, right @LeHerosInconnu ?

If the victorious hero has a free slot in his inventory, he tries to grab the ultimate artifact, but it disappears as soon as he tries.
The defeated hero retains an ultimate artifact only if the winning hero doesn't try to grab it, i.e. when his inventory no longer has a free slot to capture, or try to capture, the next artifact in the defeated hero's inventory.

> Meanwhile, I'm aware that in the original game the winning hero gets an empty inventory slot per UA on the defeated hero but this seems strange to me, which means that items are skipped. I'll ask for people's opinion on it in the issue thread.

Yes, it's weird and doesn't really make sense.

ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for the fix!

PR #9596, opened 2025-02-26
Update the Hungarian translation

big4billy:


files/lang/hu.po
```diff
@@ -7,9 +7,9 @@ msgstr ""
 "Project-Id-Version: fheroes2\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2025-02-22 01:59+0000\n"
-"PO-Revision-Date: 2023-09-04 21:55+0200\n"
+"PO-Revision-Date: 2025-02-26 22:14+0100\n"
 "Last-Translator: Berda Jenő <big4billy@gmail.com>\n"
-"Language-Team: Hungarian <hu@li.org>\n"
+"Language-Team: Hungarian <big4billy@gmail.com>\n"
```
ihhub, line 12: Please change this to:
```
"Last-Translator: fheroes2 team <fhomm2@gmail.com>\n"
"Language-Team: Hungarian <hu@li.org>\n"
```

We do not put any personal information in the source file of the engine to avoid situations of people taking ownership or reducing the impact of other people.

ihhub (changes requested):
Hi @big4billy , I left just one comment here. Could you please take a look at it?

files/lang/hu.po
big4billy: Dear Ihar,

It's far off from me to "take over" the ownership of any part of this software!

In 2023, before I restarted the Hungarian translation, I contacted the former, the "Last-Translator" via the given e-mail in .PO file. If it wouldn't have been there, I couldn't discuss with him the continuation of the translation work. I saved that file untached and I attach it now in "originalPO.zip" file. The state of the in 2014 stopped work was around 49% with lots of mistranslations and not proper wording. I started to update and make it even better ... yes, alone.

Neither me, or the former translator knows whose is that e-mail address (hu@li.org) and who maintains it therefore I cannot get any possible feedback from it about my job in translation. If the owner are you, okay I'm readily put it back.

Some words about your translator policy. I'm sad a bit that nowhere in the game is worth to mention who made the lot of voluntary work in localizations. These .PO text files are good opportunity to save not only the last translator, but the former ones as well. Just an example: I'm also working on Battle for Wesnoth game's Hungarian translation (alone too, because that work was also stopped years ago!). It contains lots of .PO-s, not only one. Every .PO contains all the history of translators with their active period between quotation marks in the first part of the file. I mean it's a nice and very little homage at their work.

[originalPO.zip](https://github.com/user-attachments/files/19076240/originalPO.zip)

ihhub: Hi @big4billy ,

My apologies about misinterpretation of the ownership word. Let me explain in details what I meant and why I asked to modify the code.

First of all, you have a solid point about `Language-Team:` and you are right about the weird email there. We need to sort this out and probably fix in all other files if needed. As of now you can leave this line as it is. @zenseii , any ideas about this?

Regarding personal information in the sources of the project. First of all, we have over 150 contributors for the project. All of them a listed in the main page of the project as well as in [contributors](https://github.com/ihhub/fheroes2/graphs/contributors) page. Every single person is being mentioned there. You can click on anyone's commits to see their contributions. For example, this is [yours](https://github.com/ihhub/fheroes2/commits?author=big4billy). Also, every source file a person modifies, contains a history of changes. For instance, for this translation file you can see the history [here](https://github.com/ihhub/fheroes2/commits/master/files/lang/hu.po) and also you can see who contributed to a specific file [here](https://github.com/ihhub/fheroes2/blame/master/files/lang/hu.po) by clicking **Blame** and seeing a list of contributors.

Having said that now we have the following situations:
- developers who contribute to specific source files must also have their contacts in these source files and imagine having to scroll 50 lines more for each file just to see the actual source code. None of developers is mentioned there just to avoid this situation (see the explanation above how to find anyone's contributions in the project). Additionally, a person could literally contribute just few lines of code by fixing an extremely rare issue in comparison with other developers who wrote thousands lines of code. There is no way of measuring the impact of their work. The same applies for translators.
- having the last translator is not a good idea. Imagine a person who changes Hungarian translation after you and literally modifies 1 line here. So, he will change the translation "ownership" (`Last-Translator`) to his name and vanishes forever. Then, another person who wants to continue would never ever get to know the state of the translation and what exactly the previous person did. The only way to prevent this is to have a single point of contact which was chosen to be the project's email, which is maintained and being accessed almost by every member of the core team, and also people should use git functionality to see who exactly modified even particular line of code.
- adding everyone's contact details is an additional work on shoulders not only contributors but also on people who review the changes.

As a result, measuring pros and cons for mentioning every single contributor or no one in the source files, the easiest and fair solution was chosen is to avoid showing contact details in the code.

I am open for discussions regarding what information we should actually put there.

ihhub: The current solution for these problem I can see is the following:
- leave `"Last-Translator: Berda Jenő <big4billy@gmail.com>\n"` as it is
- change `"Language-Team: Hungarian <hu@li.org>\n"` to `"Language-Team: \n"` or change it into `"Language-Team: fheroes2 team <fhomm2@gmail.com>\n" so that there would be at least 2 points of contact.

What do you think, @big4billy and @zenseii ?

big4billy:
That's fine for me, thx!

Ihar Hubchyk ***@***.***> ezt írta (időpont: 2025. márc. 8.,
Szo 11:19):

ihhub:
Hi @big4billy , since everyone agreed on the approach, could you please make the change?

big4billy:
Yes, but in the evenening soonest

ihhub (approved):

ihhub:
@big4billy , many thanks for this update!

Issue #9552, opened 2025-02-11
Elements shadows issues in several dialog windows


Branikolog:
### Preliminary checks

- [x] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [x] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

There're several issues with shadows generation for some dialogs:

![Image](https://github.com/user-attachments/assets/19ffee86-fa79-47c9-8fe0-f21e9078be8a)

OK button for difficulty selection dialog has too even shadow edges. 
On the contrary, checkmate icons have barely visible shadows. They also have too even edges.

![Image](https://github.com/user-attachments/assets/0eb35a5c-0141-4da8-ab1b-4a77f7a62c9e)

No shadows for translated buttons at creature info dialog.

![Image](https://github.com/user-attachments/assets/8fe33ecb-d9fa-43e1-9ba2-8f8713c99f8b)

Same dialog for the English version has shadows, but EXIT buttons seems to have the bottom shadow thinner for 1 pixel.

![Image](https://github.com/user-attachments/assets/070c3a25-4008-4b97-a0be-89321f8b47d4)

The shadows of both buttons in surrender dialog look like they are shifted downwards. Also they both seem to have too even edges.

### Save file

None.

Branikolog:
@zenseii 

I wasn't sure if these were the original shadows, that couldn't be fixed. So I haven't mentioned them here. 😅 

I personally love how the engine generates buttons shadows for recently implemented editor options dialog made by one skillful collaborator 😉 . 
It would be nice if each game button has such consistent shadow scheme. 👍

ihhub:
Hi @Branikolog , can you please update the original message in the topic to reflect what's left to do?

Branikolog:
Hi, @ihhub !

> Hi [@Branikolog](https://github.com/Branikolog) , can you please update the original message in the topic to reflect what's left to do?

Done! 👍

PR #9597, opened 2025-02-28
Add feature to generate symmetric buttons

zenseii:
This PR makes it possible to make a group of buttons in a dialog share the same size according to the widest and tallest buttons:
image

image

image

The dialog window will adjust the size according to the need of the buttons.

The castle/town button in the editor were changed to use this new function.

I also did some tidying in `agg_image.cpp` and in `icn.h`.

A bug in a resizing case was fixed.

I updated some languages that previously had to use full stop abbreviations or line breaks to fit the original button sizes.

This is an important step towards expanding button sizes for translations, which is especially needed for the main menu.

Implemented in these dialogs in this PR:
Auto combat dialog, File dialog (game), castle - town in editor

Also relevant for these dialogs:
File dialog (editor), main menu (load and new game), editor main menu, creature info dialog (upgrade/dismiss/exit), min-max and any other similar two-state buttons.

Later on it would be good to make button letters go closer if the texts are very wide, for now this is acceptable.

Related to and part of #9552 for the main menu.


src/fheroes2/gui/ui_window.cpp
```diff
@@ -343,6 +355,32 @@ namespace fheroes2
...
+
+        button.setICNInfo( icnId, releasedIndex, pressedIndex );
+        button.setPosition( pos.x, pos.y );
+        addGradientShadow( buttonSprite, _output, button.area().getPosition(), { -5, 5 } );
+        button.draw();
+    }
+
+    void StandardWindow::renderOkayCancelButtons( Button & buttonOk, Button & buttonCancel )
+    {
+        const Point buttonOffset( 20, 7 );
```
Districh-ru, line 373: So then we can use `buttonOffsetX` here too for the same constant.

src/fheroes2/gui/ui_window.cpp
```diff
@@ -378,30 +416,19 @@ namespace fheroes2
...
-
-        const int buttonCancelIcn = isEvilInterface ? ICN::BUTTON_SMALL_CANCEL_EVIL : ICN::BUTTON_SMALL_CANCEL_GOOD;
-        renderButton( buttonCancel, buttonCancelIcn, 0, 1, buttonOffset, Padding::BOTTOM_RIGHT );
+        const int32_t buttonsWidth = buttons.button( 0 ).area().width;
+        const int32_t buttonsHeight = buttons.button( 0 ).area().height;
+        for ( int row = 0; row < rows; row++ ) {
+            for ( int column = 0; column < columns; column++ ) {
+                buttons.button( column + 2 * row )
+                    .setPosition( _activeArea.x + column * buttonsWidth + buttonsOffset.x + column * 37,
+                                  _activeArea.y + ( row * ( buttonsHeight + 20 ) ) + buttonsOffset.y );
```
Districh-ru, line 427: And here too)

src/fheroes2/gui/ui_button.cpp
```diff
@@ -904,6 +939,45 @@ namespace fheroes2
...
+        std::vector<Text> buttonTexts;
+        buttonTexts.reserve( texts.size() );
+        for ( const char * text : texts ) {
+            buttonTexts.emplace_back( text, buttonFontType );
+        }
+
+        auto maxIter = std::max_element( buttonTexts.begin(), buttonTexts.end(), []( Text & a, Text & b ) { return a.width( a.width() ) < b.width( b.width() ); } );
+
+        const int32_t width = ( *maxIter ).width( ( *maxIter ).width() ) + 6;
+
```
Districh-ru, line 962: And what about setting the minimum width value?
изображение
The generated central button looks not so good as the Cancel button on the right side, IMHO.

Districh-ru (changes requested):
Hi @zenseii, great job!
I left several comments, would you mind to take a look in your spare time?

src/fheroes2/gui/ui_button.cpp
Districh-ru: So then we can pass `minWidth` as an input argument of the `makeSymmetricBackgroundSprites()` function.

src/fheroes2/gui/ui_window.cpp
Districh-ru: Since we are currently talking about `renderOkayCancelButtons()` method we can simply name this variable in the terms of this function - gaps from the edges. :)

And for the gaps between buttons in the ButtonGroup we can also pass the gaps (in example: `const fheroes2::Size gapBetweenButtons`) to the StandardWindow constructor and to the `renderSymmetricButtonGroup()`. - as we plan to use different gaps for different dialogs.
But, sadly, we'll have many same input arguments for the constructor and for methods. So in the future we can make a class for generating dialogs with only buttons that will store buttons parameters passed to the constructor and use StandardWindow to make dialog background and to render buttons...

src/fheroes2/gui/ui_window.cpp
```diff
@@ -67,6 +67,18 @@ namespace fheroes2
         render();
     }
 
+    StandardWindow::StandardWindow( const Size & buttonSize, const int columns, const int rows, const Size & windowPadding, Image & output )
+        : _output( output )
+        , _activeArea( ( output.width() - ( buttonSize.width * columns + 37 * ( columns - 1 ) + windowPadding.width ) ) / 2,
+                       ( output.height() - ( buttonSize.height * rows + windowPadding.height ) ) / 2,
+                       buttonSize.width * columns + 37 * ( columns - 1 ) + windowPadding.width, buttonSize.height * rows + 20 * ( rows - 1 ) + windowPadding.height )
```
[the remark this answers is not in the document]
Districh-ru: We should probably add a comment about why we do these calculations to make the code easier to maintain in the future.

ihhub:
Hi @zenseii , would we be able to review and merge this pull request within 5 days?

Districh-ru:
> @ihhub and @Districh-ru. This PR is now ready for your review.

@zenseii, I'll take a look at the code tomorrow and for now I have a suggestion to revert some of the changed translations (save/load map) since the Editor's File dialog is not changed in this PR and the text does not fit their width:
изображение

src/fheroes2/gui/ui_button.h
```diff
@@ -293,6 +293,11 @@ namespace fheroes2
 
         void drawShadows( Image & output = Display::instance() );
 
+        int32_t getButtonsCount()
```
Districh-ru, line 296: We can make this method `const`:
```suggestion
        int32_t getButtonsCount() const
```

src/fheroes2/gui/ui_button.cpp
```diff
@@ -955,26 +955,29 @@ namespace fheroes2
...
             buttonTexts.emplace_back( text, buttonFontType );
         }
+        // This max value is needed to make the width and height calculations correctly take into account that texts with \n are multi-lined.
+        const int32_t maxWidth = 200;
 
         auto maxIter = std::max_element( buttonTexts.begin(), buttonTexts.end(), []( Text & a, Text & b ) { return a.width( a.width() ) < b.width( b.width() ); } );
 
-        const int32_t width = ( *maxIter ).width( ( *maxIter ).width() ) + 6;
+        // We add 6 to have some extra horizontal margin.
+        const int32_t width = ( *maxIter ).width( maxWidth ) + 6;
```
Districh-ru, line 964: We can simplify the access to iterator's method :)
```suggestion
        const int32_t width = maxIter->width( maxWidth ) + 6;
```

src/fheroes2/gui/ui_button.cpp
```diff
@@ -955,26 +955,29 @@ namespace fheroes2
...
+        const int32_t finalWidth = std::clamp( width, minWidth, maxWidth );
 
-        maxIter = std::max_element( buttonTexts.begin(), buttonTexts.end(), [width]( Text & a, Text & b ) { return a.height( width ) < b.height( width ); } );
+        maxIter
+            = std::max_element( buttonTexts.begin(), buttonTexts.end(), [finalWidth]( Text & a, Text & b ) { return a.height( finalWidth ) < b.height( finalWidth ); } );
 
-        int32_t height = ( *maxIter ).height( width );
-        // The actual button background is 10 pixels taller than the text.
-        height += 10;
+        int32_t height = ( *maxIter ).height( finalWidth );
```
Districh-ru, line 970: The same here 
```suggestion
        int32_t height = maxIter->height( finalWidth );
```

src/fheroes2/gui/ui_window.cpp
```diff
@@ -67,16 +107,56 @@ namespace fheroes2
...
+    StandardWindow::StandardWindow( ButtonGroup & buttons, const bool isSingleColumn, const int32_t extraHeight, Image & output )
         : _output( output )
-        , _activeArea( ( output.width() - ( buttonSize.width * columns + 37 * ( columns - 1 ) + windowPadding.width ) ) / 2,
-                       ( output.height() - ( buttonSize.height * rows + windowPadding.height ) ) / 2,
-                       buttonSize.width * columns + 37 * ( columns - 1 ) + windowPadding.width, buttonSize.height * rows + 20 * ( rows - 1 ) + windowPadding.height )
+        , _activeArea( ( output.width() - getSymmetricDialogWidth( isSingleColumn, buttons.button( 0 ).area().width, buttons.getButtonsCount() ) ) / 2,
+                       ( output.height() - getSymmetricDialogHeight( isSingleColumn, extraHeight, buttons.button( 0 ).area().height, buttons.getButtonsCount() ) ) / 2,
+                       getSymmetricDialogWidth( isSingleColumn, buttons.button( 0 ).area().width, buttons.getButtonsCount() ),
+                       getSymmetricDialogHeight( isSingleColumn, extraHeight, buttons.button( 0 ).area().height, buttons.getButtonsCount() ) )
```
Districh-ru, line 115: `getSymmetricDialogWidth()` and `getSymmetricDialogHeight()` are called twice in this initializer with the same arguments.
What do you think about setting the `_activeArea` values in the constructor body (after the line 199)? First we can set `_activeArea.width` and `_activeArea.height` and then use it to set `_activeArea.x` and `_activeArea.y`

Districh-ru (commented):
Hi @zenseii, I did a quick review of the changed recently code and left some comments.
Could you please take a look when you have time?

src/fheroes2/gui/ui_window.cpp
Districh-ru: @zenseii, you are right, I missed that it is `const`
So let's leave the initialization as it is for now)

src/fheroes2/gui/ui_button.cpp
```diff
@@ -593,6 +601,26 @@ namespace fheroes2
...
+        for ( size_t i = 0; i < textCount; i++ ) {
+            createButton( 0, 0, sprites[i * 2], sprites[i * 2 + 1], static_cast<int>( i ) );
+        }
+    }
+
+    ButtonGroup::ButtonGroup( const int icnID )
+    {
+        // Button ICNs contain released and pressed states so we divide by two to find the number of buttons they correspond to.
+        const int32_t buttonCount = static_cast<int32_t>( fheroes2::AGG::GetICNCount( icnID ) ) / 2;
+        for ( int i = 0; i < buttonCount; i++ ) {
```
ihhub, line 619: Since **buttonCount** is `int32_t` then let's make **i** the same type.

src/fheroes2/gui/ui_button.cpp
```diff
@@ -593,6 +601,26 @@ namespace fheroes2
...
     }
 
+    ButtonGroup::ButtonGroup( const std::vector<const char *> & texts )
+    {
+        std::vector<Sprite> sprites;
+        const size_t textCount = texts.size();
+        sprites.resize( ( textCount * 2 ) );
+        makeSymmetricBackgroundSprites( sprites, texts, 86 );
+        for ( size_t i = 0; i < textCount; i++ ) {
+            createButton( 0, 0, sprites[i * 2], sprites[i * 2 + 1], static_cast<int>( i ) );
```
ihhub, line 611: I suggest to modify `ButtonGroup::createButton()` to avoid extra memory copying:
```cpp
    void ButtonGroup::createButton( const int32_t offsetX, const int32_t offsetY, Sprite released, Sprite pressed, const int returnValue )
    {
        _button.push_back( std::make_unique<ButtonSprite>( offsetX, offsetY, std::move( released ), std::move( pressed ) ) );
        _value.emplace_back( returnValue );
    }
```
and then in this function:
```cpp
createButton( 0, 0, std::move( sprites[i * 2] ), std::move( sprites[i * 2 + 1] ), static_cast<int>( i ) );
```
As a result we will not create a copy of Sprites and the old code will work as intented.

src/fheroes2/gui/ui_button.cpp
```diff
@@ -904,6 +939,48 @@ namespace fheroes2
...
+
+        std::vector<Text> buttonTexts;
+        buttonTexts.reserve( texts.size() );
+        for ( const char * text : texts ) {
+            buttonTexts.emplace_back( text, buttonFontType );
+        }
+        // This max value is needed to make the width and height calculations correctly take into account that texts with \n are multi-lined.
+        const int32_t maxWidth = 200;
+
+        auto maxIter = std::max_element( buttonTexts.begin(), buttonTexts.end(), []( Text & a, Text & b ) { return a.width( a.width() ) < b.width( b.width() ); } );
```
ihhub, line 961: Could you please explain why we use `a.width( a.width() )` call which is an equivalent to `a.width()`? Also, the input parameters should be `const` to avoid thinking that we modify the values:
```cpp
const Text & a, const Text & b
```

src/fheroes2/gui/ui_button.cpp
```diff
@@ -904,6 +939,48 @@ namespace fheroes2
...
+
+        // We add 6 to have some extra horizontal margin.
+        const int32_t width = maxIter->width( maxWidth ) + 6;
+        const int32_t finalWidth = std::clamp( width, minWidth, maxWidth );
+
+        maxIter
+            = std::max_element( buttonTexts.begin(), buttonTexts.end(), [finalWidth]( Text & a, Text & b ) { return a.height( finalWidth ) < b.height( finalWidth ); } );
```
ihhub, line 968: The input parameters should be const to avoid thinking that we modify the values:
```cpp
const Text & a, const Text & b
```

ihhub (changes requested):
Hi @zenseii , I went quickly through changes and left few comments. Would you mind to take a look at them?

Districh-ru (approved):
👍

ihhub:
Hi @zenseii , I made several changes in the code to speed up calculations and avoid any hypothetical crashes in the future. Could you please take a look whether the logic is still correct and I didn't break anything?

ihhub:
@zenseii , many thanks for this update!

PR #9599, opened 2025-03-02
Fix rendering of tall objects and extra objects (ghosts and flags) over the top map edge

Districh-ru:
Relates to #5716

This PR makes rendering of tall top layer objects after rendering the "ufit" sprites (heroes, monsters and other sprites that do not fit one tile). This fixes the issues: https://github.com/ihhub/fheroes2/issues/5716#issuecomment-1596156366, https://github.com/ihhub/fheroes2/issues/5716#issuecomment-1595729230, https://github.com/ihhub/fheroes2/issues/5716#issuecomment-1594354665
It also adds a check not to render flags and ghosts over the top map edge if the nearby tile is not revealed. This fixes this issue: https://github.com/ihhub/fheroes2/issues/5716#issuecomment-2678953114

The flags of enemy heroes, boats and monster's heads that are near the top map border and are under the fog are also not rendered now.

Ghosts for abandoned and haunted mines are now drawn over all other objects (they float high in the sky) to fix this issue:
https://github.com/ihhub/fheroes2/issues/5716#issuecomment-1946024656

This PR also does some rendering optimizations:
- makes renderer to skip the tiles fully covered with fog for top layer objects and for mine guardians;
- caches the `y * worldWidth` value not to calculate it for every tile in a row.

Master build:
изображение изображение изображение изображение

This PR:
изображение изображение изображение изображение

### DIFF
--- a/src/fheroes2/gui/interface_gamearea.cpp
+++ b/src/fheroes2/gui/interface_gamearea.cpp
@@ -510,9 +510,8 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
                 if ( isTileUnderFog && ( isUpperTileUnderFog || !hero->isShipMaster() ) ) {
                     // AI heroes can go out of fog so we have to render them one step earlier than getting out of fog.
                     if ( hero->isControlAI() && hero->isMoveEnabled() ) {
-                        const Route::Path & path = hero->GetPath();
                         // Check if the next AI hero path point will not be seen on map to skip it.
-                        if ( path.isValidForMovement() && ( world.getTile( path.GetFrontIndex() ).getFogDirection() == DIRECTION_ALL ) ) {
+                        if ( world.getTile( hero->GetPath().GetFrontIndex() ).getFogDirection() == DIRECTION_ALL ) {
                             continue;
                         }
                     }
--- a/src/fheroes2/gui/interface_gamearea.cpp
+++ b/src/fheroes2/gui/interface_gamearea.cpp
@@ -299,7 +299,8 @@ namespace
         const Maps::Tile & tileBelow = world.getTile( x, y + 1 );
 
         for ( const auto & lowerPart : tileBelow.getTopObjectParts() ) {
-            if ( lowerPart._uid == uid ) {
+            // NOTICE: Flags are put over the objects as a top object part and have the same UID but they do not make the object taller.
+            if ( lowerPart._uid == uid && lowerPart.icnType != MP2::OBJ_ICN_TYPE_FLAG32 ) {
                 // This is a tall object.
                 return true;
             }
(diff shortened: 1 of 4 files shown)


src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -103,6 +104,11 @@ namespace
                 }
             }
             else {
+                if ( offset.y == 0 && ( fogDirection & Direction::TOP ) == Direction::TOP ) {
+                    // Do not render object over the top map border if there is a nearby fog.
+                    continue;
+                }
```
ihhub, line 110: I was thinking about this problem. Imagine this situation: an object has 2 tiles above which objects exist. Now one tile has fog while another doesn't. Then we will see half of the object being rendered :) But, I think this solution is better than just render objects above fog.

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -254,6 +260,11 @@ namespace
                 }
             }
             else {
+                if ( heroPos.y == 0 && ( fogDirection & Direction::TOP ) == Direction::TOP ) {
+                    // Do not render hero's flag over the top map border if there is a nearby fog.
+                    continue;
+                }
```
ihhub, line 266: Should this check will just be as:
```cpp
if ( ( fogDirection & Direction::TOP ) == Direction::TOP ) {
    // Do not render hero's flag over the fog.
    continue;
}
```
Since, we shouldn't render hero flags anyway above fog?

src/fheroes2/gui/interface_gamearea.cpp
```diff
@@ -644,52 +644,67 @@ void Interface::GameArea::Redraw( fheroes2::Image & dst, int flag, bool isPuzzle
...
+
+                if ( areGhostsUnderFog ) {
+                    // Current tile and the top one are both fully under the fog.
+                    continue;
+                }
+
+                redrawFlyingGhostsOnMap( tile, dst, { x, y }, *this, isEditor );
+            }
+        }
+    }
```
ihhub, line 706: Here we are checking every tile to be rendered. Shouldn't we just check tiles that contain at least Mine or Abandoned Mine? We already have a loop to check at line 490.

ihhub (changes requested):
Hi @Districh-ru , I left several comments here. Would you mind please to check them in your spare time?

src/fheroes2/gui/interface_gamearea.cpp
Districh-ru: Hi @ihhub, I've also taken this problem into account (for the very high boats) and in the current code I do fix only the upper map row (`offset.y == 0`) to fix showing object's top part over the top map edge when the tile is not fully revealed.
This fix is not applied for the other tiles on the map.
This problem should be considered more carefully and fixed in the other PR.
We should decide what to do with the object partially covered with the fog: if the object's main part is under the fog - should we show all other parts? :)

src/fheroes2/gui/interface_gamearea.cpp
Districh-ru: A hero on a boat is 2 tiles high and it needs another approach to properly render this case. (Maybe we should extra checks for such high sprites.)
Currently I fix only the top map row (`heroPos.y == 0`).
This is a first step of fixing such rendering issues. :)

..later today I'll do more checks for heroes on boats. But lets focus on the highlighted issues: rendering of hero's flag over the top map border when the nearby tiles are not revealed and rendering of hero's flag over the top part of tall objects. And fix other issues in a new PR - step by step. :)

src/fheroes2/gui/interface_gamearea.cpp
Districh-ru: Sure, in example, we can make a vector for such like we do for the tall objects.
I'll try this approach later today.

Districh-ru: Done)

src/fheroes2/gui/interface_gamearea.cpp
ihhub: Tough decision. Let's keep this topic open for a while.

ihhub (approved):

ihhub:

PR #9607, opened 2025-03-05
Replace "gold only" chest dialog with "gold or exp" one in case when hero has no space for an artifact

### LINKED ISSUE #9546: Adventure map, treasure chest does not offer exp when hero has no space for an artifact
### Preliminary checks

### Platform

Windows

### Describe the bug

When trying to pick up the treasure chest containing an artifact instead of gold - game offers just a gold showing uncommon dialog:

Image

Image

For tha case of full hero the chest should offer a standard gold/exp choosing window.
The original game shows standard dialog for this case.
Offering the gold only looks like something is not normal here.

### Save file

[Chest Issue gold.zip](https://github.com/user-attachments/files/18734205/Chest.Issue.gold.zip)

### Additional info

_No response_

roman-yagodin:
This will resolve #9546.


src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1776,6 +1776,20 @@ namespace
         }
     }
 
+    uint32_t selectGoldOrExp( const std::string & hdr, uint32_t gold, Heroes & hero )
```
ihhub, line 1779: Let's modify the signature of this function:
```cpp
void selectGoldOrExp( const std::string & hdr, uint32_t & gold, Heroes & hero )
```

Then the body will be:
```cpp
        const uint32_t expr = gold > 500 ? gold - 500 : 500;
        std::string const msg = _(
            "After scouring the area, you fall upon a hidden treasure cache. You may take the gold or distribute the gold to the peasants for experience. Do you wish to keep the gold?" );

        if ( !Dialog::SelectGoldOrExp( hdr, msg, gold, expr, hero ) ) {
            gold = 0;
            hero.IncreaseExperience( expr );
        }
```

Which seem to be more logical as it is not easy to understand from the function signature what exactly it results.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1776,6 +1776,20 @@ namespace
...
+    {
+        const uint32_t expr = gold > 500 ? gold - 500 : 500;
+        std::string const msg = _(
```
ihhub, line 1782: Let's follow the same way of code as the rest of the code:
```cpp
const std::string msg
```

src/fheroes2/heroes/heroes_action.cpp
ihhub, line 1779: Since this is a helper function let's put it before the rest of functions just before `void ActionToMonster( Heroes & hero, int32_t dst_index )`.

src/fheroes2/heroes/heroes_action.cpp
```diff
@@ -1835,24 +1849,12 @@ namespace
             const Artifact & art = getArtifactFromTile( tile );
 
             if ( gold ) {
```
ihhub, line 1851: We need to do the same change for the AI in `void AIToTreasureChest()` function.

ihhub (changes requested):
Hi @roman-yagodin , I left few comments here. Could you please take a look when you have time?

roman-yagodin:
OK, sure.

src/fheroes2/heroes/heroes_action.cpp
ihhub: Hi @roman-yagodin , we haven't addressed this comment. We have to do the similar changes for the AI as well.

roman-yagodin: OK, I will look at it.

ihhub:
Hi @roman-yagodin , if you would be able to update the pull request by today then we can merge it to the upcoming release.

src/fheroes2/heroes/heroes_action.cpp
roman-yagodin: @ihhub, I've made some changes to `AIToTreasureChest()` function, but not sure I understand all ins and outs. Could you please check it out and give me some further directions? I wonder how it can be tested in first place.

ihhub (approved):

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -646,60 +646,67 @@ namespace
...
-                    gold = GoldInsteadArtifact( objectType );
+            if ( art.isValid() && !hero.PickupArtifact( art ) ) {
+                gold = GoldInsteadArtifact( objectType );
+                if ( gold ) {
+                    const uint32_t chance = 50;
+                    const uint32_t randomRoll = Rand::Get( 1, 100 );
+                    const uint32_t experience = gold > 500 ? gold - 500 : 500;
+                    if ( randomRoll <= chance ) {
+                        gold = 0;
+                        hero.IncreaseExperience( experience );
```
oleg-derevenetz, line 659: As far as I understand, this change is incorrect. Human players (according to the corresponding code in the `heroes_action.cpp` below) never get experience from water chests, right? If yes, then AI shouldn't either.

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -646,60 +646,67 @@ namespace
...
         else {
             const Artifact & art = getArtifactFromTile( tile );
 
-            if ( gold ) {
-                const uint32_t experience = gold > 500 ? gold - 500 : 500;
-                const Heroes::Role role = hero.getAIRole();
-                const int32_t kingdomGold = kingdom.GetFunds().gold;
+            if ( art.isValid() && !hero.PickupArtifact( art ) ) {
+                gold = GoldInsteadArtifact( objectType );
+            }
```
oleg-derevenetz, line 669: As far as I understand this logic, now if there is an artifact and the hero was able to pick it up, will he still have a certain chance to get minimum 500 experience? If I understood the logic correctly, then this behavior is wrong.

oleg-derevenetz (changes requested):
Hi @roman-yagodin I've left some comments, could you take a look at them as soon as you have some free time?

oleg-derevenetz (approved):

ihhub (approved):

ihhub:
@roman-yagodin and @oleg-derevenetz , thank you so much for the changes!

PR #9613, opened 2025-03-08
Fix object visiting marking logic

ihhub:
We had some "special" logic for wide objects which we were calling in an individual function. In the past we even had issues with updating their visited status which led to bugs. This pull request remove this special logic and adds a generic logic to check any object that might have multiple action object parts. This also applied for updating visited status for allies.

Before the changes we had no issues with these objects. However, if in the future (and most likely it's going to happen) we are going to add more objects with several action object parts then the current code is not flexible at all.


Districh-ru (approved):

src/fheroes2/heroes/heroes.cpp
```diff
@@ -1285,16 +1286,54 @@ bool Heroes::isObjectTypeVisited( const MP2::MapObjectType objectType, Visit::Ty
...
+    }
+
+    // An object could be bigger than 1 tile so we need to check all its tiles.
+    std::deque<int32_t> tileToVisit;
+    std::set<int32_t> visitedTiles;
+    visitedTiles.emplace( tileIndex );
+    tileToVisit.emplace_back( tileIndex );
+
+    while ( !tileToVisit.empty() ) {
+        const auto indexes = Maps::getAroundIndexes( tileToVisit.front() );
```
oleg-derevenetz, line 1315: Hi @ihhub are there any limits on object sizes on the map? I mean, if such limits exist, then it would be much more efficient just to iterate over the result of `Maps::getAroundIndexes( tileIndex, someMaxObjectSize )` and not use all these sets and deques.

ihhub: Hi @oleg-derevenetz , I think the biggest possible object size in terms of action parts is Whirlpool which is 3x2 tiles. Then the longest is Watering Hole - 4x1 tiles. Therefore, we can use 4x4 tiles as a maximum size ever for any object.

ihhub: @oleg-derevenetz , please take a look as I simplified the logic.

oleg-derevenetz (approved):
Hi @ihhub I made a few changes to make the logic a bit safer, could you please take a look when possible?

ihhub:
Hi @oleg-derevenetz , the logic looks good. Thank you so much for the improvements!

PR #9620, opened 2025-03-10
Make the AI use Guardian spells

ihhub:
You can throw rocks at me but the AI should use the spell if it can :)

Parameters and conditions are open for discussions.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -733,7 +733,7 @@ namespace
 
                 setColorOnTile( tile, hero.GetColor() );
 
-                AI::Planner::castAdventureSpellOnStandingObject( hero );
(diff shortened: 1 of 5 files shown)


src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -732,6 +732,8 @@ namespace
                 removeObjectProtection();
 
                 setColorOnTile( tile, hero.GetColor() );
+
+                AI::Planner::castAdventureSpellOnStandingObject( hero );
```
oleg-derevenetz, line 736: Hi @ihhub I suggest to rename it to `castAdventureSpellOnCapturedObject`, what do you think?

ihhub: Sure, it fits better.

oleg-derevenetz (approved):
LGTM.

Districh-ru (approved):
👍

PR #9628, opened 2025-03-13
Do not apply dynamic interface if the player is not ingame

### LINKED ISSUE #9625: Assertion triggered by having Dynamic Interface chosen when starting battle in Battle Only mode
### Preliminary checks

### Platform

Windows

### Describe the bug

Set the Dynamic Interface (Evil/Good) and start a battle in the battle only mode.

Assertion triggered here with `race` value being `0`: https://github.com/ihhub/fheroes2/blob/a6ccd233172aeb6aba645c34ac0778c050d4038d/src/fheroes2/kingdom/race.cpp#L129-L145

### Save file

None needed.

### Additional info

Related to #9413

felix642:
This is the proposed fix to #9625.
The issue is that we are trying to find the race of a player that is not in game. I've added an extra check to return the "good" interface in that case.

We might disagree and say that we should instead use the race of the selected hero in battle only mode, but I wasn't sure if this is really what we wanted and felt it might be safer to simply consider it outside of a game and review the code later if we want to change that logic.

cc: @ihhub 

Fixes #9625

### DIFF
--- a/src/fheroes2/system/settings.cpp
+++ b/src/fheroes2/system/settings.cpp
@@ -909,7 +909,7 @@ bool Settings::isEvilInterfaceEnabled() const
         return true;
     case InterfaceType::DYNAMIC: {
         const Player * player = GetPlayers().GetCurrent();
-        if ( !player ) {
+        if ( !player || !player->isPlay() ) {
             return false;
         }
 


ihhub (approved):

Districh-ru (approved):

ihhub:
@felix642 , thank you so much for the fix!

PR #9630, opened 2025-03-14
Do not set visited state if genie lamp is removed from the map

### LINKED ISSUE #9629: Assertion failed on latest snapshot build
### Preliminary checks

### Platform

Windows

### Describe the bug

I am playing the "Crowns of Eldoria" card from the Discord card set.
At one point after pressing the end of turn, such message is popped out.
Assertion failed in heroes.cpp, line 1328.
Error window and savefile (autosave) are attached.

### Save file

[AUTOSAVE.zip](https://github.com/user-attachments/files/19234246/AUTOSAVE.zip)

### Additional info

Image

Districh-ru:
Fix #9629 

If Genie Lamp is removed after all Genies are recruited AI should not mark this tile (without any object) as visited.

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1489,8 +1489,9 @@ namespace
             removeMainObjectFromTile( tile );
             resetObjectMetadata( tile );
         }
-
-        hero.SetVisited( dst_index, Visit::GLOBAL );
+        else {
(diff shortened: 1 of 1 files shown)


ihhub (approved):
So, it was an old bug since we didn't touch this code recently 🙂

ihhub:
@Districh-ru , many thanks for the fix!

PR #9644, opened 2025-03-16
SonarCloud: use the new SonarSource actions instead of the deprecated one

oleg-derevenetz:
The previously used action is marked as [deprecated](https://github.com/SonarSource/sonarcloud-github-c-cpp).


ihhub (approved):

ihhub:

Issue #9450, opened 2025-01-15
If no actions were made in the Editor disable Undo button


ihhub:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Undo button should be in a disabled state if no action was made in the Editor so a map maker would visually see whether any changes exist in the map since they opened it.

### Describe the possible solution

Disable the button is action queue is empty.

PR #9645, opened 2025-03-16
Add disable/enable logic for Editor's undo/redo buttons (#9450)

Furiiis:
Buttons redu and undo are disable if no actions were made in the Editor.
Also they become disable/enable if actions were made/revert in the Editor.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -810,8 +810,7 @@ namespace Interface
         reset();
 
         _historyManager.reset(
-                [this]( const bool isUndoAvailable, const bool isRedoAvailable ) {
-                    _editorPanel.updateUndoRedoButtonsStates(isUndoAvailable, isRedoAvailable); } );
+            [this]( const bool isUndoAvailable, const bool isRedoAvailable ) { _editorPanel.updateUndoRedoButtonsStates( isUndoAvailable, isRedoAvailable ); } );
 
         if ( isNewMap ) {
             _mapFormat = {};
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -831,6 +831,10 @@ namespace Interface
         _tileUnderCursor = -1;
         _areaSelectionStartTileId = -1;
 
+        _historyManager.setChangedCallback([this](const bool isUndoAvailable, const bool isRedoAvailable) {
+            _editorPanel._updateButtonStates(isUndoAvailable, isRedoAvailable);
+        });
+
         uint32_t redrawFlags = REDRAW_GAMEAREA | REDRAW_RADAR | REDRAW_PANEL | REDRAW_STATUS | REDRAW_BORDER;
         if ( conf.isEditorPassabilityEnabled() ) {
             redrawFlags |= REDRAW_PASSABILITIES;
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -809,7 +809,9 @@ namespace Interface
 
         reset();
 
-        _historyManager.reset();
+        _historyManager.reset(
+                [this]( const bool isUndoAvailable, const bool isRedoAvailable ) {
+                    _editorPanel.updateUndoRedoButtonsStates(isUndoAvailable, isRedoAvailable); } );
 
         if ( isNewMap ) {
             _mapFormat = {};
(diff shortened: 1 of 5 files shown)


ihhub:
Hi @Furiiis , please address code style issues.

src/fheroes2/editor/editor_interface_panel.cpp
```diff
@@ -894,6 +896,26 @@ namespace Interface
...
+
+        if ( isRedoAvailable ) {
+            _buttonRedo.enable();
+        }
+        else {
+            _buttonRedo.disable();
+        }
+
+        _buttonUndo.draw();
+        _buttonRedo.draw();
```
Districh-ru, line 916: To enable/disable button and draw it's new state only if it has changed we can slightly improve this logic:
```cpp
        auto updateButtonState = []( fheroes2::ButtonBase & button, const bool isAvailable ) {
            if ( isAvailable ) {
                if ( button.isDisabled() ) {
                    button.enable();
                    button.draw();
                }
            }
            else if ( button.isEnabled() ) {
                button.disable();
                button.draw();
            }
        };

        updateButtonState( _buttonUndo, isUndoAvailable );
        updateButtonState( _buttonRedo, isRedoAvailable );
```
What do you think?

src/fheroes2/editor/editor_interface_panel.h
```diff
@@ -120,6 +120,8 @@ namespace Interface
 
         static const char * getObjectGroupName( const Maps::ObjectGroup groupName );
 
+        void _updateButtonStates( const bool isUndoAvailable, const bool isRedoAvailable );
```
Districh-ru, line 123: Let's rename this method using the [code style naming suggestions](https://github.com/ihhub/fheroes2/wiki/Code-style):
`updateUndoRedoButtonsStates`
This name is longer, but it sais that only Undo and Redo button's states will be updated.

src/fheroes2/editor/history_manager.h
```diff
@@ -75,10 +76,16 @@ namespace fheroes2
     class HistoryManager
     {
     public:
+        void setChangedCallback( const std::function<void( const bool, const bool )> & callback )
+        {
+            _changesCallback = callback;
+        }
```
Districh-ru, line 82: When loading a map or starting a new one from the editor and having already some undo/redo history the buttons's states are not updated:

https://github.com/user-attachments/assets/f0ec840f-192b-4af7-87a9-4c4c66596565

We can update their states in this method when setting a callback function:
```suggestion
        void setChangedCallback( const std::function<void( const bool, const bool )> & callback )
        {
            _changesCallback = callback;

            if ( _changesCallback ) {
                _changesCallback( isUndoAvailable(), isRedoAvailable() );
            }
        }
```

I took this place to update the button's state as an example.

IMHO a better solution could be to place
```cpp
            if ( _changesCallback ) {
                _changesCallback( false, false );
            }
```
in the `reset()` method and to use it for setting the `_changesCallback` and remove the `setChangedCallback()` method:
```cpp
        void reset( const std::function<void( const bool, const bool )> & undoRedoCallback = nullptr )
        {
            _actions.clear();
            _lastActionId = 0;
            _changesCallback = undoRedoCallback ;

            if ( _changesCallback ) {
                _changesCallback( isUndoAvailable(), isRedoAvailable() );
            }
        }
```

What is your opinion?

Districh-ru (changes requested):
Hi @Furiiis, good approach!
I left several comments, would you mind to take a look when you have time?

ihhub (changes requested):
Hi @Furiiis , I left few small comments here. Could you please take a look when you have time?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -809,7 +809,8 @@ namespace Interface
 
         reset();
 
-        _historyManager.reset();
```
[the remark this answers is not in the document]
Furiiis: I changed the code as you recommended, but now `HistoryManager::reset()` doesn't reset the callback. First, we set the callback in the EditorInterface constructor, and then we reset this callback in fheroes2::GameMode EditorInterface::startEdit(const bool isNewMap).

src/fheroes2/editor/editor_interface.h
```diff
@@ -141,7 +141,8 @@ namespace Interface
             , _editorPanel( *this )
             , _warningMessage( *this )
         {
-            // Do nothing.
+            _historyManager.setStateCallback(
```
ihhub, line 144: A good development approach is to always reduce the scope (without over-engineering methods) of variables. Here we are passing **this** while we use only a single member of the class. I suggest that we pass only **_editorPanel** by reference. I also see that SonarQube might complain about `this` usage. What do you think?

src/fheroes2/editor/history_manager.h
```diff
@@ -75,10 +76,23 @@ namespace fheroes2
     class HistoryManager
     {
     public:
+        void setStateCallback( std::function<void( const bool, const bool )> stateCallback )
+        {
+            _stateCallback = std::move( stateCallback );
+
+            if ( _stateCallback ) {
+                _stateCallback( isUndoAvailable(), isRedoAvailable() );
+            }
```
ihhub, line 85: I recommend to remove this code and logically this method is designed only to set the callback but not to execute it. Since we already have a call of `updateUndoRedoButtonsStates( false, false );` within **EditorPanel** class this code will do nothing.

src/fheroes2/editor/editor_interface.cpp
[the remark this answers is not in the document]
ihhub: This is the correct approach.

ihhub (approved):

Districh-ru (approved):
Well done!

PR #9650, opened 2025-03-18
AI heroes should buy spellbook if there isn't one

idshibanov:
Fun fact: I found a commit from 2021 with this exact change but I don't think it was ever merged into main.

Test using Knight/Barbarian heroes.


oleg-derevenetz (approved):
LGTM. I just removed a few extra calls (to avoid doing the same thing multiple times) and renamed a few methods to make it more clear what they do exactly.

ihhub (approved):

ihhub:
@idshibanov , thank you so much for the improvement!

PR #9655, opened 2025-03-18
Campaign: allow to raise the game difficulty during the campaign

### LINKED ISSUE #9649: Allow increasing difficulty during campaign
### Preliminary checks

After some initial troubles I found the original campaigns at normal difficulty a bit too easy, and wanted to increase the difficulty. But the hard difficulty is disabled after the first mission, due to issues with score multiplier tracking (according to https://github.com/ihhub/fheroes2/discussions/9648#discussioncomment-12527062). I don't care about the scores, so having a whole difficulty level almost removed (locked behind a campaign restart) just for that feels arbitrary and frustrating.

Difficulty switching should be allowed, but the score modifier should only ever go down. Switching from a lower difficulty to a higher one should still be scored as playing on previous difficulty, but switching from a higher difficulty to a lower one should decrease the bonus accordingly. That way difficulty switching cannot be used to get unearned high scores, while those who don't care about scores can switch to any difficulty for any scenario.

There should be a confirmation popup with a warning about this when a player tries to make the switch.

### Additional info

_No response_

oleg-derevenetz:
close #9649

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -344,7 +344,6 @@ namespace
             assert( hero.GetIndex() == dstIndex );
 
             DEBUG_LOG( DBG_AI, DBG_INFO, hero.GetName() << " visits " << castle->GetName() )
-
             return;
         }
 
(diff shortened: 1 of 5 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for these changes!

Issue #9638, opened 2025-03-15
AI heroes get stuck in portals


dtcyganok:
### Preliminary checks

- [x] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [x] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Linux

### Describe the bug

Sometimes AI heroes get stuck in portals. They can teleport to other portals, but by some reason they decide not to move until the end of the game. I've encountered this strange behavior twice on the maps "Enroth" and "Sudden Siege".

![Image](https://github.com/user-attachments/assets/795e72a4-5210-4f60-8772-021cd7d6f51c)

### Save file

The stuck hero is in the bottom right corner of the map.

[Enroth_0031.zip](https://github.com/user-attachments/files/19266482/Enroth_0031.zip)

oleg-derevenetz:
> They can teleport to other portals

In reality, they cannot. Ability to teleport without movement (using the spacebar button) is an fheroes2 extension (in the original HoMM2 you cannot do that) and this extension is human-only, AI currently is not able to do this due to a number of reasons (mostly due to compatibility concerns and a few technical reasons). So this not a bug - if AI cannot move anywhere from the portal (due to too strong guard on a nearby tile, for instance), then it will not be able to teleport to other portals.

dtcyganok:
OK, probably there are compatibility concerns and technical reasons.

Still this behavior looks incorrect and strange at least.

dtcyganok:
> "protect" human players from AI using the random nature of teleports.

Interesting compromise. Though as a human player I use the random nature of teleports to jump to desirable location. So it might be more fair if AI players could do the same. My two pens)

oleg-derevenetz:
> By the way the AI of heroes 3 can do so if I remember it right.

HoMM3 has the "spacebar mechanics" from the very beginning, unlike HoMM2, which didn't have it, so there are original HoMM2 maps (I don't remember which ones) that use the HoMM2-specific teleport mechanics to make it difficult for heroes to move to certain places of the map (for instance, between the AI "zone" and the human player's "zone"). BTW, the same is related to Whirlpools.

minuspinus:
> > By the way the AI of heroes 3 can do so if I remember it right.
> 
[...]

PR #9661, opened 2025-03-20
Don't let heroes reuse the Stone Liths and Whirlpools while standing on them

oleg-derevenetz:


ihhub:
Hi @oleg-derevenetz , could you please create a separate pull request with your fixes for Code Checker since we have other PRs that have similar issues?

ihhub:
@oleg-derevenetz , alternatively I can disable a mandatory check for PR gates using Code Checker for now.

oleg-derevenetz:
Hi @ihhub

> could you please create a separate pull request with your fixes for Code Checker since we have other PRs that have similar issues?

These are experimental approaches I tried to fix the CodeChecker issue. I'll open a separate PR as soon as I settle this.

ihhub (approved):

ihhub:
I've added other reviewers to gather their opinion about this change.

Districh-ru:
Hello everyone!
I don't like this change as it leads to forcing player use many save/load (as they did in the original game and it was really annoying) to go to the wanted portal exit.
If some player don't want to use this ability then can not use the space bar - the engine does not force to use it.
And if you say that it should be disabled because AI cannot use this ability then along with this change, in example, we should not let heroes to dig for ultimate artifact because AI currently does not do it. :)

The ability to use the space bar does not break the map logic (instead of pressing the space bar you can do save and many "load and use portal" - that how it works in the original game). So technically this change will not drastically change the gameplay but sometimes this can make using many portals more annoying for a player who really wants to move their hero to a specific location.

If the community votes for this change then so be it, but I vote for keeping the master branch logic.

... and as a compromise the space bar reuse option can be disabled for Expert and Impossible difficulties and not only for portals (if you visit a dwelling at the 7th day and buy some troops and wants to stay for the next day on these difficulties you'll need to re-enter the dwelling without using the space bar, like in the original game).

oleg-derevenetz:
That's what I'm talking about :) It's not so easy to make such a change now, when everyone is already used to this HoMM3-style mechanics. Technically, it's nothing especially complicated, but from a psychological point of view, there is a lot of resistance to preserving the mechanics of the original HoMM2.

oleg-derevenetz:
> It is therefore obvious that we need, and players/map makers want, a portal which allows for selection of an exit, and both this and the original portals can exist together.

It's not entirely clear to me exactly how this will work. What if some of the exits of this portal are in the fog of war, how are they supposed to be chosen? Or is it assumed that they immediately highlight themselves (and perhaps some small area around them) for all players immediately from the beginning of the game? As far as I know, there are no similar mechanics in any (at least popular) game from the HoMM series.

Districh-ru:
> That's what I'm talking about :) It's not so easy to make such a change now, when everyone is already used to this HoMM3-style mechanics. Technically, it's nothing especially complicated, but from a psychological point of view, there is a lot of resistance to preserving the mechanics of the original HoMM2.

If we want to position this PR as a restoration of the original HoMM2 game mechanics then we should consider fully removing the reuse ability for heroes.
The solution in current PR makes the portal visiting to follow th original game logic, but all other action objects do not follow this rule: you can move a hero to a Windmill at the 7th day, press the space bar at the 1st day and move the hero having all his move points at the start. It gives a question - why the Windmill is better than a portal and is this a bug that I cannot reuse the portal while I can reuse all other objects. :)
The remove of reuse ability will force players to better plan their moves. And disabling this ability for all action objects looks like a complete follow-up to the original HoMM2 game rules, which I agree.

@oleg-derevenetz, @zenseii, @ihhub what do you think?

oleg-derevenetz:
> The solution in current PR makes the portal visiting to follow th original game logic, but all other action objects do not follow this rule

In fact, this PR is just a implementation of [this proposal](https://github.com/ihhub/fheroes2/issues/9638#issuecomment-2738304748) from @modo-lv. I decided to implement it and thereby initiate the discussion, because once again there were those who were dissatisfied with the current consensus between human player's UX improvements and following the logic of the original game.

LeHerosInconnu:
Hello everyone,

I support the @Districh-ru's point that the way the stone liths work in the original game brings frustration to the user.

In general, I'm of the opinion that if the user can bypass certain restrictions due to the game's operating design in any way (in this case saving/reloading the game before/after an action), because this allows him to remove a frustration generated by the game, one might as well directly remove the frustration generated by adjusting the game's operating design.

The way stone liths work in the original game can only be considered interesting from a player experience point of view when stone liths are not yet discovered on the adventure map (still under the fog of war).
This adds a certain amount of suspense to the course of the game at that precise moment (while the stone liths are still hidden).
After that, the player knows the possibilities of moving around with to the stone liths, and the suspense disappears from the game.
When the stone liths have not yet been revealed, the gameplay of the stone liths is more like an RPG (adventure and discovery), and after the stone liths have been revealed, the gameplay of the stone liths becomes more like a strategy game (moving from one point to another).

The problem that arises at this point (when the stone liths are revealed) is that their use continues to be an RPG type action when it should be a strategy game type action.
This brings frustration to the user, as there is something “wrong” with the way the game works.
Heroes of Might and Magic 2 is first and foremost a strategy game, with some RPG aspects, true, but it is first and foremost a strategy game.

It's also true that some of the game's original scenarios also make use of the random operation of stone liths, and the fact that having to leave the stone liths' position and then return to use it again consumes movement points to bring a strategic dimension to the moves while retaining an RPG feel; but in the end this only increases the time needed to complete the scenario.
The most important thing, in my opinion, is to develop a strategy to obtain a powerful hero with a substantial army supported by solid “resources” (cities for gold, creatures, spells, position, etc., mines for resources, etc.).
The randomness of movement with the stone liths, which may have seemed fun, ends up being a constraint.

For the rest, the fact that the hero has to leave a location and then return to enter it again, all the while expending movement points, is frustrating and, by the same token, poor game design.
Let's leave this useful enhancement of revisiting a site with the associated button and space bar brought by fheroes2 present.

Concerning the possible new types of stone liths with different functions proposed by @zenseii, I'm not in favor.
In Heroes 2, there's only one type of stone liths, no one way stone liths as in Heroes 3, etc.
Heroes 2 (fheroes2) is a simple game in that respect and should stay that way.
A single type of stone liths (in its functioning) is sufficient.
In Heroes 2, stone liths are more like shortcuts for moving around (strategy game aspect) than traps for moving around (random movement, or forced movement in the case of one way stone liths, etc., RPG game aspect).

Now that players are used to this enhancement to revisit a site, it's going to seem like a step backwards, bringing frustration, without a real addition to the game even if it is to function as the orinal game, as @oleg-derevenetz also points out.
Keeping the original way the game works just to keep the original way the game works doesn't really make sense, especially if things can be improved without drastically changing the original way the game works.

For the operation of stone liths, here's how it could work:
With a single stone liths.
Normally this shouldn't happen, this case should be reported to the scenario creator in the scenario editor when he saves his work so that he can remedy the situation.

With two stone liths.
In this case, there's no need to present a destination selection window.
This should remain fluid in use.

With three or more stone liths.
One stone liths revealed, the others hidden by the fog of war.
Here again, no need to present a selection window, the hero is sent to one of the other stone liths at random.

Several stone liths revealed, at least one hidden by the fog of war.
In this case, a selection window is presented (in the same way as for the town portal spell).
One entry is displayed for each stone liths already revealed.
Another entry, displayed first in the list and selected by default, indicating, for example, "Unrevealed location", sends the hero to the last stone liths or one of the other stone liths not yet revealed.

When all stone liths are revealed, it is no longer necessary to display the entry offering a stone lith not yet revealed.

With all this in mind, moving with stone liths shouldn't consume any movement points.
This is considered more of a strategic move from a location where the hero doesn't move (the hero stays on the same square, allowing him to move to the destination of his choice from among the possible ones).
Players who want a more RPG style movement can still do so by randomly choosing a destination from the list of available destinations.
Naturally, the AI ​​should use the ability to choose the destination that suits it best, whether it's to reveal the adventure map, or to move faster.

oleg-derevenetz:
> The problem that arises at this point (when the stone liths are revealed) is that their use continues to be an RPG type action when it should be a strategy game type action.
This brings frustration to the user, as there is something “wrong” with the way the game works.
Heroes of Might and Magic 2 is first and foremost a strategy game, with some RPG aspects, true, but it is first and foremost a strategy game.

In my opinion, people came here to play HoMM2 - as they remember it. I don't think there's a real need to introduce any mechanics that have fundamental differences from the original game, especially those that will be available on original maps one way or another. If the original game has a certain balance between the strategic component and the RPG component that was developed by its creators, then we need to keep it as is as much as possible, and not arbitrarily shift it to one side "just because we can".

modo-lv:
> I don't like this change as it leads to forcing player use many save/load (as they did in the original game and it was really annoying) to go to the wanted portal exit.

Getting a bad result in combat also leads players to save/load and try again to get a better result. That's player choice -- accept the result you got or savescum to get a better one. Players using save files to avoid risks is not a reason to remove those risks from the game.

> (if you visit a dwelling at the 7th day and buy some troops and wants to stay for the next day on these difficulties you'll need to re-enter the dwelling without using the space bar, like in the original game)

I don't agree that it's the same, as I explained in [my original comment](/ihhub/fheroes2/issues/9638#issuecomment-2738304748).

> to use it again consumes movement points to bring a strategic dimension to the moves while retaining an RPG feel; but in the end this only increases the time needed to complete the scenario.

It also increases the risk that the enemy (AI or human player) will gain advantage. While your hero is wasting movement points trying to get to the right portal exit, their heroes can spend theirs on exploring the map, acquiring resources etc.

I believe that the randomness, combined with having to move away and return, is/was a way to limit stone lith use, similar to how spell points limit teleportation spells like Dimension Door. Being able to just teleport across the map to any lith you want can a very powerful feature in a game where movement is a limited resource, so having some cost associated with it makes sense. I feel this was intentional by the original developers (as evidenced by maps specifically designed with this limitation in mind) and should function as in the original. Especially if it also makes human and AI gameplay rules more consistent.

As for the PR itself, I think that it is crucial to have an informative message pop up when a player tries to use spacebar on a portal, letting them know that the exception is intentional. I'm not seeing that in the PR, which could make this change come across as a bug or at the very least confusing for players unaware of these discussions.


P.S. Spacebar activation also didn't exist in the original HoMM 3, it was added with Armageddon's Blade expansion. I wonder if NWC developers had as much discussion as fheroes2 about portal balancing. :D

oleg-derevenetz:
> As for the PR itself, I think that it is crucial to have an informative message pop up when a player tries to use spacebar on a portal, letting them know that the exception is intentional. I'm not seeing that in the PR, which could make this change come across as a bug or at the very least confusing for players unaware of these discussions.

Regarding this: the game UI already has an indication of whether you can reuse the object you are standing on or not - the image on the "continue movement" button (with the horse) serves for this purpose. Therefore, if the question concerns whether the fact that the spacebar does not work on this object is intentional or not, then the state of this button should resolve all questions of this kind. Another question is _why_ this is intentional (i.e. the rationale). In this regard, I doubt that one or the other explanation of "why" in this case will be enough, because it will exceed the size of this discussion :) And even so, there will still be dissatisfied people, so I consider the explanation of "why" in a dialog window superfluous. As well as a dialog window that simply says "you cannot do this", which will simply duplicate the already existing information from an already existing button in the UI.

modo-lv:
Oh, and I'm still not opposed to making this a setting -- as a last resort, if that's what it takes. @oleg-derevenetz [explained why he's against it](https://github.com/ihhub/fheroes2/issues/9638#issuecomment-2738148533), and I completely agree that having a bunch of game mechanic variations is a bad idea, but maybe a rare exception could be justified for this, if a consensus can't be reached. As long as it is made very clear in the code and everywhere else that it is not done casually just because someone asked, but only for a controversial, long-debated issue where the original mechanics and those who prefer them directly and irreconcilably conflict with those who prefer the modernization.

modo-lv:
> Therefore, if the question concerns whether the fact that the spacebar does not work on this object is intentional or not, then the state of this button should resolve all questions of this kind.

In general you're right, I'm just thinking about the fact that players are used to the spacebar teleportation. If this PR goes through and the portal "spacing" stops working, it might lead to a bunch of superfluous bug reports, Discord questions, etc. A popup saying something like "you must step away from the liths and gather your scattered mind to be able to use them again" would leave zero doubt about the intentionality of the change.

oleg-derevenetz:
> If this PR goes through and the portal "spacing" stops working, it might lead to a bunch of superfluous bug reports, Discord questions, etc.

I have an impression that there will be a lot of "Discord questions, etc" in any case - "but why???". And to explain this "why" is not an easy task at all, there always will be dissatisfied people.

modo-lv:
> I have an impression that there will be a lot of "Discord questions, etc" in any case - "but why???". And to explain this "why" is not an easy task at all, there always will be dissatisfied people.

The "why"s will need links to discussions, yeah, there's no way to explain something that meta in-game. It just seems like a flavor text popup could help with transitioning to a new behavior process for players (and also for new players, who know/discover the spacebar first, and then think they've encountered a bug when they get to a portal). But of course it's not functionally necessary.

ihhub:
We need to solve 2 problems:
- make a decision about the logic of the game: either retain the original logic and it is good for the original maps; or keep the spacebar logic which some players used to. Remember, that players that use spacebar mostly PC games while on touch devices many players have no idea about this
- make the human player and AI players have the same conditions for the objects. At the moment humans have privilege of using Sone Liths and Whirlpools. The complexity of the implementation is also a huge factor

In any case some players is going to be dissatisfied so we need to make a weighted decision based on the above factors.

oleg-derevenetz:
> We need to solve 2 problems:
> - make a decision about the logic of the game: either retain the original logic and it is good for the original maps; or keep the spacebar logic which some players used to. Remember, that players that use spacebar mostly PC games while on touch devices many players have no idea about this
> - make the human player and AI players have the same conditions for the objects. At the moment humans have privilege of using Sone Liths and Whirlpools. The complexity of the implementation is also a huge factor
> 
> In any case some players is going to be dissatisfied so we need to make a weighted decision based on the above factors.

Or we can just keep the current consensus as is: AI is not allowed to re-use the Stone Liths and Whirlpools (allowing the AI to re-use them in fact means that AI will be able to teleport directly to a desired position, which will break the random nature of teleports itself as well as maps that are designed based on it), but human players are allowed to re-use them. In other words, we don't have to make any changes and just ignore the relevant complaints as well as suggestions for improvements that violate this consensus.

Districh-ru:
The **reuse** ability is not present in HoMM2 gameplay, it is a fact.
Talking about Stone Liths and Whirlpools @LeHerosInconnu separated interaction with them by two types:
a) Scouting. And this type differs between the original game and fheroes2. In fheroes2 you can enter a portal, reuse it as many as you want to scout all possible ways out and then return the portal you started from and go away without spending any move points to do scouting. This differs from the original game and it cannot be overcome by any save/load if you want the scouted area to stay on your map (if not - scout - you don't like this way - reload - scout once more).
b) Moving from point A to point B through a portal. You can go from a portal to the certain other one without spending any move points in fheroes2 by using the reuse ability and on original game by reloading a save that was done before entering a portal. So the result has no difference except that in original game it takes more time and actions and in non-critical situations players  mostly will not use this and use the randomly selected portal exit. (The game insists on following its rules :) )

All other action objects differs from the portals and there might be also two situations:
a) A player found that he need to do some other action in this move and revisit this object (in example, get more gold to bye more troops from dwelling). In original game he'll need to reload the previous save (autosave) and do the actions in the other order not to loose move points on revisiting the object while in fheroes2 you can just reuse object after doing all needed actions.
b) A player wants to start a day from visiting some action object. In original game he will need to move one tile to the object (or even move from the object and then move back), interact with the object and continue his move with less move points. While in fheroes2 a hero can start a day standing on the action object, reuse it and continue his move having all move points: in example, a hero finished his move on the 7th day on at the Windmill to reuse it at the 1st day ant start moving having all move points. This cannot be overcome in the original game.

So not only the portal interaction currently differs with the original game. And if we want to retain the original logic we should do it for all action objects not to have a mess: these objects act like in original game ant these are not.

But currently in fheroes2 a player can use on not use the "reuse" ability. The engine does not force him to reuse portal or any other action object and does not restricts him to move out from the action object and revisit it again. So it all depends on the player - what type of the gameplay he will use.

But when we will start implementing the multiplayer we should discuss it once more or even make some multiplayer rules because players would like to play a multiplayer game using the original game rules and the will want all the players in this game to follow the rules.

I don't like this PR because it does not make fheroes2 to fully follow the classic interaction with the map objects and for non-scouting portal use it will bring almost no change to the gameplay, but make it a little annoying.
Following all the action object interaction rules will look more finished and justified even if it leads to some annoyance in some places. Before fheroes2 we played the original game with these rules and we like the game and we can say the same about its rules and we know them.

This is only mine point of view and I respect the point of view of other people posted here.
And I'm OK to merge this PR as it is but for me it looks like unfinished in the way of following the original game mechanics... But we can start from this and maybe later discuss the way we want or not to have the reuse ability for the rest of the action objects.

Also I'd like to mention that #9638 will not be closed by this PR. The AI will still stuck in the case described in the issue.

oleg-derevenetz:
> Also I'd like to mention that https://github.com/ihhub/fheroes2/issues/9638 will not be closed by this PR. The AI will still stuck in the case described in the issue.

#9638 assumes that AI hero can re-use the portal to teleport to other portals, but in fact it's not really the case. This is certainly not a "bug" and I'm not sure that it's even an "issue". At best, this can be classified as a "request for improvement", effectively suggesting that the AI attack the guards, even if the AI hero's defeat is inevitable. In any case, this has nothing to do with the mechanics of the game as such.

Districh-ru:
> #9638 assumes that AI hero can re-use the portal to teleport to other portals, but in fact it's not really the case. This is certainly not a "bug" and I'm not sure that it's even an "issue". At best, this can be classified as a "request for improvement", effectively suggesting that the AI should attack the guards, even if the AI hero's defeat is inevitable. In any case, this has nothing to do with the mechanics of the game as such.

I was thinking the same that the only way to solve this AI stuck and portal block is to force AI to attack the monsters. And if we keep the reuse ability then we could allow AI to use it once _only_ in that case.

oleg-derevenetz:
@Districh-ru @zenseii BTW we can perform "whitelisting" instead of "blacklisting" - only allow re-use of certain objects that cannot be abused one way or another. If done this way, then what kind of objects could it be?

LeHerosInconnu:
Hello @ihhub,

> We need to solve 2 problems:
> 
> * make a decision about the logic of the game: either retain the original logic and it is good for the original maps; or keep the spacebar logic which some players used to. Remember, that players that use spacebar mostly PC games while on touch devices many players have no idea about this
> * make the human player and AI players have the same conditions for the objects. At the moment humans have privilege of using Sone Liths and Whirlpools. The complexity of the implementation is also a huge factor
> 
> In any case some players is going to be dissatisfied so we need to make a weighted decision based on the above factors.

I think users who don't use a keyboard can use the dedicated GUI button, yes?

Taking these constraints into account, I think it's preferable to allow the use of the dedicated button and the space bar to revisit a site, and to allow the AI to directly select the stone liths and whirlpool destination (which the human player can also do, albeit in a less than optimal way (re-use of the dedicated button or space bar instead of direct selection from a list)).
This would also solve the human player's problem of having to watch the AI's multiple moves in stone liths and wirlpools during the AI's turn (and also the creature losses engendered by the AI's multiple visits of whirlpools).

oleg-derevenetz:
> Taking these constraints into account, I think it's preferable to allow the use of the dedicated button and the space bar to revisit a site, and to allow the AI to directly select the stone liths and whirlpool destination (which the human player can also do, albeit in a less than optimal way (re-use of the dedicated button or space bar instead of direct selection from a list)).

If the developers of the original HoMM2 had wanted just such a mechanic, they would have implemented it, and would not have made moving through Stone Liths and Whirlpools random and consuming movement points. Anyway, it's part of the design of the original game, whether someone likes it or not, and it just so happens that there are maps that rely on just such mechanics.

oleg-derevenetz:
> For castles I use the re-visit key on Android quite often because it's more convenient than trying to click twice on the castle if the castle list is long.

OK, we have at least one case which I suppose even @Districh-ru won't argue about :)

LeHerosInconnu:
Hello everyone,

The choice adopted should take into account the satisfaction of as many types of player as possible.

The random operation of stone liths and whirlpools in the original game can be circumvented with a save/reload action before/after the action to be taken.
The loss of creatures in whirlpools can also be circumvented in the same way, or limited by creating a troop of one weak creature when possible.

From my point of view, this action of saving before an in-game action stems from a less-than-optimal design, since certain types of player feel the need to perform it.
Removing the need to save before an in-game action would therefore help to improve the game.

And the fact that the player or the AI can thus improve its movements during the game won't make it win the game at the same time.
This will, at most, enable it to complete the scenario more quickly (for both the human player and the AI).

modo-lv:
> The random operation of stone liths and whirlpools in the original game can be circumvented with a save/reload action before/after the action to be taken.
The loss of creatures in whirlpools can also be circumvented in the same way
[...]
From my point of view, this action of saving before an in-game action stems from a less-than-optimal design, since certain types of player feel the need to perform it.

Not knowing enemy hero skills, magic and exact unit counts (big difference between 20 and 49 "lots of" Cyclops) before combat can also be circumvented with a save/reload action (and many players feel the need to save before battles). Does that mean it is less-than-optimal design, and full exact information about the enemy hero should be visible before deciding to battle?

> After that, the player wants to use the stone liths to move around the adventure map more efficiently, and the randomness of the workings of the stone liths becomes a constraint.

Yes, paying for things you need is a constraint, it's part of the game balance. Troops cost resources, spells cost MP and liths cost movement points. Teleportation can be a huge advantage that saves days or even weeks of travel time (that's why every Door and Portal spell is level 4+, costs 10 MP and also needs movements points), and static locations + randomness + movement loss is the price. (Personally, I feel like the cost should be more -- MP, units or severely reduced movement next day, but this is primarily a recreation of a game, not an expansion, so we should keep to the original.)

Districh-ru:
> @Districh-ru @zenseii BTW we can perform "whitelisting" instead of "blacklisting" - only allow re-use of certain objects that cannot be abused one way or another. If done this way, then what kind of objects could it be?

@oleg-derevenetz, many action objects can be whitelisted: Castles/Towns, Oracle, Hill Fort, luck/morale power-ups, Arena,  ...

IMHO in the blacklist there should be:
- portals (Whirlpools and Stone Liths);
- weekly replenished sources that gives all resources/pover-ups to the visiting hero (Wind/Watermill, Magic Garden, Stables, Artesian Spring);
- dwellings.
It is a quick list and it can be discussed. :)

oleg-derevenetz:
> dwellings

By the way, what if, when hero visits a dwelling, there is no room in his army or not enough money to hire the desired number of troops? IIRC in the original game he will need to re-visit the dwelling (and spend the movement points) after dismissing some troops or trading at the marketplace.

modo-lv:
I disagree with blacklisting anything other than portals. Their randomness is what makes the movement cost matter. It's essentially gambling -- spend some movement points to try and win a lot more (by avoiding long travel times). Without such unpredictability, having an activation cost serves no mechanical/strategic purpose.

Districh-ru:
> Hello @Districh-ru,
> 
> What will this bring to the game?
> It will just annoy the player.
> Is it really necessary?

Hello @LeHerosInconnu, I just gave a list of objects that from mine point of view might have different interaction results (spending the move points) between the original game and fheroes2.
I used to use the reuse ability (sorry for the pun). So I'd like to keep the fheroes2 logic as it is. :)
But looking from the original game mechanics point of view - for some actions it looks like a small cheat.
If we are talking about possible following the rules of the original game then we should find out the differences and try to avoid them. Earlier I tried to describe that differences and, yes, it will be more annoying without the reuse ability as I pointed out earlier, but that are the rules of the HoMM2 game and sometimes a one tile more move can save the game.

I typed dwellings because one-two times I had a case: to get creatures from the dwelling in the first day and to run away from a strong enemy to a castle. Without the reuse ability the enemy would overtake me.
This is a very rare case, but it can happen and it cannot be bypassed by saving/loading.
In other cases the reuse ability will not affect the gameplay but the lack of it will annoy players like it said by @oleg-derevenetz.

So first of all we should decide do we want to strictly follow the rules of the original game?
Or maybe we can follow the original rules for some cases (where the original logic cannot be bypassed) and have flexibility for the other cases.
In example: do not allow to reuse dwellings and windmills only if hero is starting his move. And, possibly, do not allow to reuse portals until all of them are visited (exiting the portal will not mean that it is visited for this case).

> I disagree with blacklisting anything other than portals. Their randomness is what makes the movement cost matter. It's essentially gambling -- spend some movement points to try and win a lot more (by avoiding long travel times). Without such unpredictability, having an activation cost serves no mechanical/strategic purpose.

@modo-lv, you are right that it is applicable to any other action and after loosing a battle a player can always reload the game and try as many as he wants. But the battle is a more complicated action then entering a portal and a portal relies only on a random generator. Randomizing is a part of the game and it makes the game replayable. But (maybe sadly) this the only action while the battle is a sequence of random actions.
If there is no critical situation (run from strong enemy, run to protect your castle) then having no reuse will probably not annoy so much and most players will not reload (if the portal at the 2nd-3rd tile will travel a hero to the needed exit).
And if it is a critical situation then the randomness of the portals can be bypassed by reloads in the original game (if it is really needed).
I've added other objects to the blacklist because for some cases they give a difference in move points without having a random logic. But this is only a suggestion to discuss the final logic of fheroes2.

If a player wants a challenge then the original portal mechanic will give him this challenge, I agree. These players will not play on easy skill. But other players may want to reuse the portals. So as a compromise we can also consider not to blacklist portals for easy and maybe normal skills.

LeHerosInconnu:
@Districh-ru,

I understand your point of view, but the rules have to be simple to understand and to use in the game.
If some objects follow certain rules, and other objects follow other rules, and this is different again depending on conditions or the difficulty level of the game, the player will find it complicated.
All this just to penalize the player's movements for a not-so-great effect over the whole game and for the majority of games.
This means sacrificing a certain fluidity and enjoyment of the game for very little.
The player will spend an artificially increased amount of time countering what is imposed on him by the way the game works.
Exiting the location, returning to the location to activate it, etc., all of this generates dozens and dozens of clicks to fight against the game's interface, while the player could be spending his time deploying a strategy to move towards victory.
It becomes more of a labor than a thinking exercise.

dtcyganok:
I don't want to create additional noise, so I'll try to be concise. 
As @ihhub said we have 2 problems: 1) the logic of the game and 2) the equality of conditions.

This PR certainly solves the problem of the equality. Meanwhile The logic of the game and AI behavior are gigantic topics and should probably be spitted and debated separately. I don’t think mixing different, albeit related, topics helps to move forward. So let’s temporarily drop it.

ihhub:
Hi @oleg-derevenetz , @zenseii , @Districh-ru , @LeHerosInconnu @modo-lv ,

It took me a long time but I finally came back to these changes. In total, we have the following solutions:
1. Keep as it is. **Advantages**: no need to changes anything. **Disadvantages**: AI has unequal conditions. `It's a no for me since the AI must have the same conditions as human players.`
2. Revert to the original game's logic. **Advantages**: predictable and in line with the original game. **Disadvantages**: some players are unhappy, we already have fheroes2 maps utilizing this behavior so it would lead to broken maps behavior. `How many broken maps we have in total with this?`
3. The current solution (don't let heroes reuse the Stone Liths and Whirlpools while standing on them). **Advantages**: makes the logic for human players closer to the AI. **Disadvantages**: AI and human players still have different conditions to play, making this specific logic only for certain objects might lead to many questions from players.
4. Allow AI and human players to use space bar (or at least human players for now) but the action will cost a hero movement points. **Advantages**: human and AI players are going to have very similar conditions (not the same for now but much closer) and also none of them can abuse the gameplay too much. **Disadvantages**: some players are unhappy.

Note: I don't think that if we make any changes players will start using save file reloading. This technique is true for visiting other objects, picking resources, battles and etc. For some it might be annoying but for others it is not. If a player wants to get the best outcome they will use the reload.

I would like to hear other solutions / options but please keep it short that it is understandable without introduction of complex logic and most importantly not to confuse players. If we have no other options and we just vote and proceed with the solution.

oleg-derevenetz:
> The current solution (don't let heroes reuse the Stone Liths and Whirlpools while standing on them). **Advantages**: makes the logic for human players closer to the AI. **Disadvantages**: AI and human players still have different conditions to play, making this specific logic only for certain objects might lead to many questions from players.

Well, actually, it's not so bad about the different conditions. Yes, teleports are a rather delicate place in terms of consuming movement points, but for example, I don't see anything terrible in the fact that movement points will not be consumed when re-visiting dwellings or objects where a human player decides on a battle (such as a Graveyard or Utopia of Dragons). In the case of dwellings, AI (in principle) can also be trained to wait for the new week directly in them, this will not break the game balance in any way as well as it will not break any maps. In the case of objects, when the AI is going to visit such an object, it already knows for sure whether it is going to fight or just wants to take a shortcut, so this won't break anything either.

Districh-ru:
Hi, everybody!
I'm fine with 4th solution.
Taking penalty for reentering the teleport is almost the same as going outside the teleport and step back into it.
If AI will be able to use this then the issue #9638 will be closed: AI will return back from the portal. It will also make AI more closer to the real player because AI cannot load a save before entering the portal in the #9638 case. :)

oleg-derevenetz:
> I'm fine with 4th solution.

This variant is too much different from both HoMM2 and HoMM3. As far as I understand, there are no such mechanics anywhere else in games of HoMM series.

Districh-ru:
> This variant is too much different from both HoMM2 and HoMM3. As far as I understand, there are no such mechanics anywhere else in games of HoMM series.

Yes. It is like a compromise between HoMM2 and HoMM3 approaches.
The cost of re-entering the teleport can be equal to two moves. And also no one will forbid player to step outside the portal (to scout the area) and to go back for the same cost. But for many same portals when you want to go from teleport "A" to teleport "B" with the 4th solution it will be easier for player to do jumps before the randomizer will bring him to the teleport "B". And it will cost him the same movement points like in OG.

But after some thoughts ... we should not implement something much different from the known mechanics.

I'm also fine with the other solutions. But for the 3rd one (this PR) we should clearly tell the players why it was taken and why only teleports.

LeHerosInconnu:
Hello everyone,

Letting the user use the GUI button (and the space bar) is what I think is most desirable, in terms of conformity of operation with the other objects on the adventure map, practicality, simplicity, speed of interaction, etc.
A fixed operating cost should also be the most desirable, the easiest to understand (and accept), and also simpler to implement.
In my opinion, this operating cost should not be higher than the cost of moving on a road, which is 75 movement points.
For my part, an operating cost of 50 movement points seems to be the value to use.
It's easy to remember, it's not too penalizing so that it's acceptable to the greatest number of players (without "forcing" save/load), it's simple, and it doesn't overly modify the way the original game works.

tlsa:
> I would like to hear other solutions / options but please keep it short that it is understandable without introduction of complex logic and most importantly not to confuse players.

I would suggest keeping it exactly as it is now for human players. (That's the least confusing thing for the users and it's a popular feature.)

And for AI players I'd suggest letting them do exactly the same as human players, although it can be tied to game difficulty. Let the AI players play with all the efficiency of "space bar" at impossible difficulty, but limit them at easy difficulty to the old behavior. There is already a precedent for restricting AI abilities according to difficulty.

LeHerosInconnu:
Hello @tlsa,

> > I would like to hear other solutions / options but please keep it short that it is understandable without introduction of complex logic and most importantly not to confuse players.
> 
> I would suggest keeping it exactly as it is now for human players. (That's the least confusing thing for the users and it's a popular feature.)

I'd favor that too.

ihhub:
I can see that this discussion didn't go anywhere and we are still stuck in the the same state. Since many people have very different opinions about portal we should solve at least one very obvious case: whirlpools. I see no issue to make them not being accessible by a spacebar as players can move to a different tile of the same whirlpool. This is a very trivial change in this pull request as well.

Does anyone have very strong objection about this? @oleg-derevenetz , @Districh-ru , @zenseii and @LeHerosInconnu .

LeHerosInconnu:
Hello @ihhub,

> I can see that this discussion didn't go anywhere and we are still stuck in the the same state. Since many people have very different opinions about portal we should solve at least one very obvious case: whirlpools. I see no issue to make them not being accessible by a spacebar as players can move to a different tile of the same whirlpool. This is a very trivial change in this pull request as well.
> 
> Does anyone have very strong objection about this? @oleg-derevenetz , @Districh-ru , @zenseii and @LeHerosInconnu .

Honestly, the current system of using the space bar or the graphic button on the interface without any movement cost for the hero for stone liths and whirlpools is the better approach.
This would not favor human player, as the AI would also benefit from it.
Additionally, if the player wants to recreate the original game experience, he always has the possibility of not using the space bar or the graphic button.
Everyone is happy.

For the AI, choosing the desired destination point directly does not seem excessive to me.
In the end, that's what the human player will do.
This would also prevent the human player from viewing all their comings and goings during the AI players' observation phase.
All in all, there are many benefits to this.

ihhub:
> Hello @ihhub,
> 
> > I can see that this discussion didn't go anywhere and we are still stuck in the the same state. Since many people have very different opinions about portal we should solve at least one very obvious case: whirlpools. I see no issue to make them not being accessible by a spacebar as players can move to a different tile of the same whirlpool. This is a very trivial change in this pull request as well.
> > Does anyone have very strong objection about this? @oleg-derevenetz , @Districh-ru , @zenseii and @LeHerosInconnu .
> 
> Honestly, the current system of using the space bar or the graphic button on the interface without any movement cost for the hero for stone liths and whirlpools is the better approach. This would not favor human player, as the AI would also benefit from it. Additionally, if the player wants to recreate the original game experience, he always has the possibility of not using the space bar or the graphic button. Everyone is happy.
> 
> For the AI, choosing the desired destination point directly does not seem excessive to me. In the end, that's what the human player will do. This would also prevent the human player from viewing all their comings and goings during the AI players' observation phase. All in all, there are many benefits to this.

Let's keep the conversation short. Do you have any **strong objections** about my proposal?

Objection is an opinion based on facts, not preferences.

LeHerosInconnu:
@ihhub,

> Let's keep the conversation short. Do you have any **strong objections** about my proposal?
> 
> Objection is an opinion based on facts, not preferences.

As you wrote, the player can already move the hero from one square of the whirlpool to another.
Preventing the player from using the space bar or the graphic button will only impose a constraint on the player compared to the current functioning.
Other than that, I have no strong objections.

Districh-ru:
Hi @ihhub, I agree with @LeHerosInconnu position:
I have no strong objections against this change.
This is how original game operates and we played it for many years and we like the game.
fheroes2 added the _ability_ to use the space bar and I like it.
If we want to follow the OG in this case then we can start from the Whirlpools and see the community reaction on this change.

oleg-derevenetz:
> Since many people have very different opinions about portal we should solve at least one very obvious case: whirlpools. I see no issue to make them not being accessible by a spacebar as players can move to a different tile of the same whirlpool. This is a very trivial change in this pull request as well.

I have no objection to start to implement this logic from Whirlpool.

zenseii:
I agree with just implementing this for whirlpools.

ihhub:
Hi @oleg-derevenetz ,

Since no major objections and people agreed to give it a try please proceed with changes for Whirlpools only at the moment.

Districh-ru (approved):

ihhub:

PR #9662, opened 2025-03-20
Add French virtual keyboard

zenseii:
Part of #8826

In addition we stop using the French original assets fonts and use our engine generated because there are too many issues with the original font.

image

Had to add a bunch of missing symbols from the button font and small font. Also fixed a bunch of broken fonts with the `¨` sign.


ihhub (approved):

ihhub:
Hi @oleg-derevenetz , has Code Checker been updated recently? It seems to me now it complains about more stuff. This is why this pull request failed.

ihhub (approved):

Districh-ru:
> I've looked at the Polish version and we can just use the engine generated font there too. A lot of our letters and accents even look better than the original.
> 
> This leaves only Russian. @Districh-ru, do you think we can just force the use of our engine generated Cyrillic font when a user has Russian assets right now?

Hi @zenseii, I insist ( :) ) on using the engine generated fonts along with the engine generated buttons - they look the same or even better than in Russian assets.
And it'll also help us maintaining the engine because fonts in Russian assets sometime have a 1 pixel difference that sometimes may result in drastic difference on rendered game screen between engine generated font and the font from Russian assets.

ihhub:
@zenseii , a bunch of thanks for this update!

PR #9668, opened 2025-03-21
Add defense penalty to moat building info and adjust wording

### LINKED ISSUE #8808: Combat screen, Ballista info window, the movement penalty for troops in the moat is not currently active
### Preliminary checks

This movement penalty is a suggestion I made for the previous fheroes2 project, which had not yet been implemented at the time.
In the current fheroes2 project, the penalty is not active.
The text should be corrected to avoid confusion.

In fheroes2:

Ballista info window 005

This could be:

"The Moat stops the movement of troops entering it and reduces their defense skill by 3."

Instead of:

"The Moat reduces the defense skill of troops trapped in it by 3 and restricts their movement range."

> Hello @Killbrum,
> 
> In the original game, the moat information window indicates that the troops are slowed down and become more vulnerable to attacks.
> 
> In original game:
> 
> Moat penalty H2 01
> 
> In the original game, during combat, troops end their movement when they enter the moat.
> For vulnerability to attack, I never calculated if they really received more damage when they were in the moat.
> 
> Maybe @idshibanov could have more information about this.
> 
> The information text for the defense of the castle regarding the moat currently in fheroes2 (-3 to creature defense and halved movement) is one of my proposals for the previous fheroes2 project, but it does not seem to have been fully implemented in the previous project since troops do not suffer a movement penalty to get out of the moat; for the -3 defense penalty, I did not check either.
> 
> Since this proposal changes the original way the game works, it should only be included as an option.
> But this can wait for Beyond 1.0 version of the game.
> 
> In fheroes2:
> 
> Moat penalty fh2 01
> 
> Moat penalty fh2 02
> 
> _Originally posted by @LeHerosInconnu in https://github.com/ihhub/fheroes2/issues/471#issuecomment-725522473_

The text should be corrected to avoid confusion.
This could be:

"The Moat stops the movement of troops entering it and reduces their defense skill by 3."

Instead of:

"The Moat reduces the defense skill of troops trapped in it by 3 and restricts their movement range."

### Additional info

_No response_

modo-lv:
Fixes #8808.

I also slightly adjusted the wording to more clearly indicate that:
* Flying units aren't affected by the moat's movement penalty, but do still suffer the defense penalty if they land in it.
* The moat ends a unit's turn on *any* movement in it, not just entering.

### DIFF
--- a/src/fheroes2/castle/buildinginfo.cpp
+++ b/src/fheroes2/castle/buildinginfo.cpp
@@ -34,6 +34,7 @@
 #include "castle_building_info.h"
 #include "cursor.h"
 #include "dialog.h"
+#include "game_static.h"
 #include "icn.h"
 #include "localevent.h"
 #include "m82.h"
@@ -53,7 +54,6 @@
 #include "ui_constants.h"
 #include "ui_dialog.h"
 #include "ui_text.h"
-#include "game_static.h"
 
 namespace
 {
--- a/src/fheroes2/castle/buildinginfo.cpp
+++ b/src/fheroes2/castle/buildinginfo.cpp
@@ -53,6 +53,7 @@
 #include "ui_constants.h"
 #include "ui_dialog.h"
 #include "ui_text.h"
+#include "game_static.h"
 
 namespace
 {
@@ -311,6 +312,9 @@ std::string BuildingInfo::getBuildingDescription( const int race, const uint32_t
             StringReplace( description, "%{count}", Castle::GetGrownWel2() );
             break;
         case BUILD_CASTLE:
+        case BUILD_MOAT:
+            StringReplace( description, "%{count}", GameStatic::GetBattleMoatReduceDefense() );
+            break;
         case BUILD_STATUE:
         case BUILD_SPEC: {
             const Funds profit = ProfitConditions::FromBuilding( buildingId, race );
--- a/src/fheroes2/castle/buildinginfo.cpp
+++ b/src/fheroes2/castle/buildinginfo.cpp
@@ -311,10 +311,10 @@ std::string BuildingInfo::getBuildingDescription( const int race, const uint32_t
         case BUILD_WEL2:
             StringReplace( description, "%{count}", Castle::GetGrownWel2() );
             break;
-        case BUILD_CASTLE:
         case BUILD_MOAT:
             StringReplace( description, "%{count}", GameStatic::GetBattleMoatReduceDefense() );
             break;
+        case BUILD_CASTLE:
         case BUILD_STATUE:
         case BUILD_SPEC: {
             const Funds profit = ProfitConditions::FromBuilding( buildingId, race );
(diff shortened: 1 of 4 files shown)


ihhub:
Hi @modo-lv , please address code styling issues.

src/fheroes2/castle/buildinginfo.cpp
```diff
@@ -53,6 +53,7 @@
 #include "ui_constants.h"
 #include "ui_dialog.h"
 #include "ui_text.h"
+#include <game_static.h>
```
oleg-derevenetz, line 56: Hi @modo-lv please use double quotes instead of angle brackets for this header. This is a local header, so we don't need the [compiler to try to search for it](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html) in the system `include` directories and compiler's own ones.

modo-lv: Sorry, should have noticed the VS quick-fix used the wrong syntax. Fixed.

ihhub:
Hi @modo-lv , once you address comments from reviewers please either leave a message here or click on re-review icon for them so they could be notified about readiness of the changes :)

modo-lv:
> Hi @modo-lv , once you address comments from reviewers please either leave a message here or click on re-review icon for them so they could be notified about readiness of the changes :)

I don't remember seeing a "re-review" button anywhere, where is it?

oleg-derevenetz:
> I don't remember seeing a "re-review" button anywhere, where is it?

reviewers

But I'm not sure that this button is available to non-collaborators. Probably, ordinary contributors do not have it.

modo-lv:
So that's why I couldn't find it, I was looking the comment section. :D It is available and I see it now, thanks.

src/fheroes2/castle/buildinginfo.cpp
```diff
@@ -311,6 +312,9 @@ std::string BuildingInfo::getBuildingDescription( const int race, const uint32_t
             StringReplace( description, "%{count}", Castle::GetGrownWel2() );
             break;
         case BUILD_CASTLE:
+        case BUILD_MOAT:
```
oleg-derevenetz, line 315: This change doesn't look correct. The `fheroes2::getBuildingDescription()` for castle returns the following string: "The Castle improves the town's defense and increases its income to %{count} gold per day.", so it's incorrect to replace `%{count}` by the result of `GameStatic::GetBattleMoatReduceDefense()`. `case BUILD_CASTLE:` should be moved below and grouped with `BUILD_STATUE` and `BUILD_SPEC` as it was before this change.

oleg-derevenetz (changes requested):
Hi @modo-lv I just left one comment regarding the proposed changes, could you please take a look when you have time?

src/fheroes2/castle/buildinginfo.cpp
modo-lv: Fixed.

oleg-derevenetz (approved):

PR #9674, opened 2025-03-22
List item selection moves along with navigation keys

modo-lv:
Lists (save/load etc.) can be scrolled with arrow keys or navigation keys (page up/down, home, end), but the selection doesn't move with the latter, making them useless without additional mouse input and in conflict with arrows keys (list position jumps back to selection on arrow presses). This PR resolves that conflict by moving selection along with navigation keys, making all navigation keys useful and working together.

### DIFF
--- a/src/fheroes2/gui/interface_list.h
+++ b/src/fheroes2/gui/interface_list.h
@@ -340,31 +340,51 @@ namespace Interface
                 return false;
             }
 
-            if ( useHotkeys && le.isKeyPressed( fheroes2::Key::KEY_PAGE_UP ) && ( _topId > 0 ) ) {
-                needRedraw = true;
+            if ( useHotkeys && le.isKeyPressed( fheroes2::Key::KEY_PAGE_UP ) ) {
+                SetCurrentVisible();
+                if ( _topId > 0 ) {
+                    needRedraw = true;
 
-                if ( _topId > maxItems ) {
-                    _topId -= maxItems;
-                }
-                else {
-                    _topId = 0;
-                }
+                    const int prevTop = _topId;
+                    if ( _topId > maxItems ) {
+                        _topId -= maxItems;
+                    }
+                    else {
+                        _topId = 0;
+                    }
 
-                UpdateScrollbarRange();
-                _scrollbar.moveToIndex( _topId );
+                    UpdateScrollbarRange();
+                    _scrollbar.moveToIndex( _topId );
+                    _currentId = std::clamp( _currentId - ( prevTop - _topId ), _topId, _lastVisibleId() );
+                }
+                // The view is at the top of the list, but selection isn't
+                else if ( _currentId != 0 ) {
+                    _currentId = 0;
+                    needRedraw = true;
+                }
 
                 return true;
             }
-            if ( useHotkeys && le.isKeyPressed( fheroes2::Key::KEY_PAGE_DOWN ) && ( _topId + maxItems < _size() ) ) {
-                needRedraw = true;
+            if ( useHotkeys && le.isKeyPressed( fheroes2::Key::KEY_PAGE_DOWN ) ) {
+                SetCurrentVisible();
+                if ( _topId + maxItems < _size() ) {
+                    needRedraw = true;
 
-                _topId += maxItems;
-                if ( _topId + maxItems >= _size() ) {
-                    _topId = _size() - maxItems;
+                    const int prevTop = _topId;
+                    _topId += maxItems;
+                    if ( _topId + maxItems >= _size() ) {
+                        _topId = _size() - maxItems;
+                    }
+
+                    UpdateScrollbarRange();
+                    _scrollbar.moveToIndex( _topId );
+                    _currentId = std::clamp( _currentId + ( _topId - prevTop ), _topId, _lastVisibleId() );
                 }
-
-                UpdateScrollbarRange();
-                _scrollbar.moveToIndex( _topId );
+                // The view is at the bottom of the list, but the selection isn't
+                else if ( _currentId < _size() - 1 ) {
+                    _currentId = _size() - 1;
+                    needRedraw = true;
+                }
 
                 return true;
             }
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/interface_list.h
```diff
@@ -400,6 +419,7 @@ namespace Interface
                 needRedraw = true;
 
                 _topId = _size() - maxItems;
+                _currentId = std::clamp( _topId + maxItems, _currentId, _size() - 1 );
```
Districh-ru, line 422: By the end button we want to position on the last item in the list so we can simply:
```suggestion
                _currentId = _size() - 1;
```

src/fheroes2/gui/interface_list.h
```diff
@@ -340,31 +340,49 @@ namespace Interface
...
+                    if ( _topId > maxItems ) {
+                        _topId -= maxItems;
+                    }
+                    else {
+                        _topId = 0;
+                    }
+
+                    UpdateScrollbarRange();
+                    _scrollbar.moveToIndex( _topId );
+                    _currentId = std::clamp( _currentId - ( prevTop - _topId ), _topId, _currentId );
```
Districh-ru, line 357: If the currently selected item is not visible `_topId` can be bigger than `_currentId` (just scroll down the list after it is opened and press the Page Up key) and it leads to assertion in `clamp()`.
Also to focus an item in the visible part of the list we can limit cursor position only to a visible part of the list:
```suggestion
                    _currentId = std::clamp( _currentId - ( prevTop - _topId ), _topId, _topId + maxItems - 1 );
```

Another way is to move `_currentId` one page up and update the _topId` to have the selected item in the visible part of the list.

src/fheroes2/gui/interface_list.h
```diff
@@ -340,31 +340,49 @@ namespace Interface
...
+                    _topId += maxItems;
+                    if ( _topId + maxItems >= _size() ) {
+                        _topId = _size() - maxItems;
+                    }
 
-                UpdateScrollbarRange();
-                _scrollbar.moveToIndex( _topId );
+                    UpdateScrollbarRange();
+                    _scrollbar.moveToIndex( _topId );
+                    _currentId = std::clamp( _currentId + ( _topId - prevTop ), _topId, _size() - 1 );
```
Districh-ru, line 379: Almost the same as the previous comment - If the selected item is to the bottom of the visible part of the list then it will not be seen after doing Page Down.

Districh-ru (changes requested):
Hi @modo-lv, a good improvement!
I left three comments, could you please take a look when you have time?

src/fheroes2/gui/interface_list.h
modo-lv: Thanks, fixed.

src/fheroes2/gui/interface_list.h
modo-lv: I think somehow I got the Page Down code in the End section. Fixed.

src/fheroes2/gui/interface_list.h
modo-lv: Fixed.

modo-lv:
I've also made it so that if the player scrolls with the mouse but then uses navigation keys, they work the same as arrow keys (jump the viewport back to selection before moving up/down).

src/fheroes2/gui/interface_list.h
```diff
@@ -541,6 +563,11 @@ namespace Interface
             return content == nullptr ? 0 : static_cast<int>( content->size() );
         }
 
+        int _lastVisibleId()
```
Districh-ru, line 566: This method can be made `private` and `const`.

Districh-ru (changes requested):
@modo-lv, it works great.
I left one more comment, could you please check it.

src/fheroes2/gui/interface_list.h
modo-lv: Done.

Districh-ru (approved):

src/fheroes2/gui/interface_list.h
```diff
@@ -386,20 +406,22 @@ namespace Interface
...
                 UpdateScrollbarRange();
                 _scrollbar.moveToIndex( _topId );
 
                 return true;
             }
-            if ( useHotkeys && le.isKeyPressed( fheroes2::Key::KEY_END ) && ( _topId + maxItems < _size() ) ) {
+            if ( useHotkeys && le.isKeyPressed( fheroes2::Key::KEY_END ) && ( _topId + maxItems < _size() || _currentId < _lastVisibleId() ) ) {
                 needRedraw = true;
 
                 _topId = _size() - maxItems;
```
oleg-derevenetz, line 423: Suppose that we have just a single element in list. Before this PR, the control flow will never get here because `_topId + maxItems` will be always greater than `_size()`. But now we will get here (because `_currentId` (zero) will be less than `_topId + maxItems - 1`) and we will get negative `_topId` (because `maxItems` is usually more than 1). Yes, `Verify()` will clean it up somewhere later, but it still doesn't look very neat. What do you think?

modo-lv: I modified the `_lastVisibleId()` to never exceed list size (and even become `-1` if the list is empty). This should ensure that current ID can only be below last visible ID if there's at least two items in the list.

oleg-derevenetz: OK, now suppose that we have two elements in a list (which is less than `maxItems`), and the first one is currently selected. We press the End button and we will still get here due to `_currentId < _lastVisibleId()` (0 < 1). Now we have the same problem: `_topId` will become negative, because `_size()` is less than `maxItems`.

modo-lv: Getting into the code block is correct (pressing end *should* move the selection to the last item, even if no scrolling takes place), but you're right about `_topId` going negative. I have added a `0` minimum restriction to prevent it.

src/fheroes2/gui/interface_list.h
```diff
@@ -556,6 +578,11 @@ namespace Interface
         fheroes2::TimedEventValidator _timedButtonPgUp;
         fheroes2::TimedEventValidator _timedButtonPgDn;
 
+        int _lastVisibleId() const
+        {
+            return _topId + std::min( maxItems, _size() ) - 1;
```
Districh-ru, line 583: Hi @modo-lv, what about making this expression a bit more clear. When the list is smaller than the `maxItems` the `_topId` will be equal to `0` (but if for some reason it will be not 0) so we can rewrite it:
```suggestion
            return std::min( _topId + maxItems, _size() ) - 1;
```
So now it will say - we return the minimum position of last showed item (starting from the `_topId`) and the last item in the whole list.

modo-lv: Done.

oleg-derevenetz (approved):

oleg-derevenetz:
@modo-lv Thank you very much!

Issue #1530, opened 2020-09-04
Expressions for the quantity of creatures directly replaced by equivalent numbers


LeHerosInconnu:
I never remember what is Pack, Lots, Horde, and fight 10-19 Vampire Lords is not the same as 50-99. :)

In the original game, I replaced the expressions used to indicate the number of creatures with their equivalent in number using a hex editor.

It would be nice to add an option to the game that will make this easier.

"Game: Replace expressions with numbers for the quantity of creatures"

So replace:

Few by 1-4
Several by 5-9
Pack by 10-19
Lots by 20-49
Horde by 50-99
Throng by 100-249
Swarm by 250-499
Zounds by 500-999
Legion by 1000+

In original game:

![Creatures equivalent numbers 01](https://user-images.githubusercontent.com/43583994/92218236-18b2cf00-ee99-11ea-9348-6c23466f649b.png)

![Creatures equivalent numbers 05](https://user-images.githubusercontent.com/43583994/92218272-249e9100-ee99-11ea-81cd-673f47c316aa.png)

![Creatures equivalent numbers 06](https://user-images.githubusercontent.com/43583994/92218277-26685480-ee99-11ea-8ac5-7bc565d742ea.png)

Original game edited:

![Creatures equivalent numbers 04](https://user-images.githubusercontent.com/43583994/92218263-22d4cd80-ee99-11ea-9059-ea787a00ed7d.png)

![Creatures equivalent numbers 02](https://user-images.githubusercontent.com/43583994/92218242-1a7c9280-ee99-11ea-94f9-6f9bb6d6e9f2.png)

![Creatures equivalent numbers 03](https://user-images.githubusercontent.com/43583994/92218246-1c465600-ee99-11ea-9a30-c85b9eeba489.png)

modo-lv:
I'd like this as well, however I'm not sure we should completely replace the verbal amounts, as they've been a pretty fundamental part of the HoMM series, at least the NWC versions. Something like `lots (20-49)` could be a compromise that preserves the original wording while also providing the numeric ranges.

dtcyganok:
It'd be a really nice addition! I've always been confused by mysterious lots, horde etc. 

Why not consider it as default? The aforementioned compromised option `lots (20-49)` would be nice as well.

modo-lv:
> Why not consider it as default?

[...]

PR #9678, opened 2025-03-23
Add numeric army estimates (cfg file only)

modo-lv:
This PR adds a `numeric army size estimates` setting to `fheroes2.cfg` file. When changed to `on`, army estimates in game will display number ranges ("10-19") instead of the original verbal descriptions ("lots").

This has been requested in #1530, and I agree that it should be an option, however adding it to the UI turned out to be beyond my ability (at least for now). So this is not a prefect fix, useful for those who can manually edit their configuration files. But I believe this is better than nothing, and hopefully the UI parts can be added in the future.

image

image


src/fheroes2/army/army.cpp
```diff
@@ -82,6 +82,18 @@ namespace
         ARMY_LEGION = 1000
     };
 
+    std::unordered_map<ArmySize, std::string> TroopSizeNumbers{
```
oleg-derevenetz, line 85: This map can be made `const`, I suppose.

src/fheroes2/army/army.cpp
```diff
@@ -82,6 +82,18 @@ namespace
...
+        { ArmySize::ARMY_FEW, "1-4" },
+        { ArmySize::ARMY_SEVERAL, "5-9", },
```
oleg-derevenetz, line 87: Please remove these extra commas at the end, they break the formatting.

src/fheroes2/system/settings.cpp
```diff
@@ -80,7 +80,7 @@ namespace
         GAME_BATTLE_AUTO_RESOLVE = 0x04000000,
         GAME_BATTLE_AUTO_SPELLCAST = 0x08000000,
         GAME_AUTO_SAVE_AT_BEGINNING_OF_TURN = 0x10000000,
-        GAME_SCREEN_SCALING_TYPE_NEAREST = 0x20000000
+        GAME_SCREEN_SCALING_TYPE_NEAREST = 0x20000000,
```
oleg-derevenetz, line 83: Please remove this extra comma at the end, this is an unnecessary change (unless you don't want to add the bit flag).

src/fheroes2/system/settings.h
```diff
@@ -399,6 +409,7 @@ class Settings
     int game_type;
     ZoomLevel _viewWorldZoomLevel{ ZoomLevel::ZoomLevel1 };
     InterfaceType _interfaceType{ InterfaceType::GOOD };
+    bool _numericArmyEstimates{ false };
```
oleg-derevenetz, line 412: Historically, bit flags have been used for this. Is there any reason not to use a bit flag?

oleg-derevenetz (changes requested):
Hi @modo-lv I left a few comments regarding this PR, could you please to take a look when possible?

As for the idea of PR in general, we usually avoid introducing new settings if they cannot be done through the GUI, because on some platforms (that are considered first-class - for instance, on modern Android versions), manually editing the config is very difficult for an average user.

src/fheroes2/system/settings.h
modo-lv: I looked at the bit flags, but seeing as there's only 2 spaces left (`0x40000000` and `0x80000000`) and not sure what determines whether a setting should be a bit flag or a property, I opted to go with property. I guess it's more efficient storing booleans in flags, but with game settings being a singleton, does it really make that much difference?

modo-lv: I guess I'm asking whether we need the bit flags at all, instead of a more consistent layout where boolean settings are simple class members like all the others.

modo-lv: OK; setting changed to a bit flag.

src/fheroes2/army/army.cpp
modo-lv: Fixed.

src/fheroes2/system/settings.cpp
modo-lv: Indeed, this was a leftover mistake from early attempts; now the new setting does use the bit flag, so the comma is back.

I wouldn't intentionally add a comma just for the sake of adding it, but in general I prefer to have trailing commas for all elements in a listing. It allows for easy adding, removing and moving around items if necessary, without also having to separately add/remove the commas. Also keeps the last existing line unchanged when adding a new item, so it stays out of diffs it isn't part of.

oleg-derevenetz (approved):
Technically this PR looks OK, but I'd wait for @ihhub decision, because, as I already mentioned, there are [certain requirements](https://github.com/ihhub/fheroes2/wiki/Adding-a-new-option-into-the-game) for new options.

ihhub:
I will review this PR later.

Overall, I am not against this change as I've seen many players asking for this but I see that players won't like the idea of changing such a parameter inside a configuration file. It is not a user friendly approach. Therefore, this change must have a dedicated UI option. We don't need to implement it in this PR but at least have a brief idea how it should look like in the game and where to put it.

I am tagging @Branikolog for this matter.

modo-lv:
My idea was to make the interface settings panel a 3x3 instead of 2x2, and use these artifact icons for the setting. They have similar purple background to normal setting icons and are also similar enough to each other too look like two states of the same thing.

image
image

ihhub:
@modo-lv , icons for the option must be suitable, not some random images. We need to create new ones.

Having settled about the name of the option we should update the current source code as well. I think the option should be called something like "Army size view" and has 2 states: canonical and mumeric.

What do you think @zenseii , @oleg-derevenetz and @Branikolog ?

oleg-derevenetz:
> icons for the option must be suitable, not some random images. We need to create new ones.
> 
> Having settled about the name of the option we should update the current source code as well. I think the option should be called something like "Army size view" and has 2 states: canonical and numeric.

I agree.

Branikolog:
Hi!

>I think the option should be called something like "Army size view" and has 2 states: canonical and numeric.

Maybe we can stick to just two words: army size or army view. I think it would be enough.

Also, if we are going to expand settings window to 3x2 we need to add one more option there. 😅 

As for the icon itself, I personally think, that making a collage with several monsters sprites and toggling text (pack/"10-19") under them could work.

modo-lv:
Why not "army estimates"? Since it's not an exact size.

ihhub:
I can see that we are stuck here. Let me make the review and propose the needed changes for the future UI.

src/fheroes2/system/settings.h
```diff
@@ -269,6 +270,7 @@ class Settings
     void setAutoSaveAtBeginningOfTurn( const bool enable );
     void setBattleDamageInfo( const bool enable );
     void setHideInterface( const bool enable );
+    void setNumericEstimates( const bool enable );
```
ihhub, line 273: Let's call it `setNumericArmyEstimationView()`.

src/fheroes2/system/settings.h
```diff
@@ -399,6 +401,7 @@ class Settings
     int game_type;
     ZoomLevel _viewWorldZoomLevel{ ZoomLevel::ZoomLevel1 };
     InterfaceType _interfaceType{ InterfaceType::GOOD };
+    bool _numericArmyEstimates{ false };
```
ihhub, line 404: Please remove this member as we use `GAME_NUMERIC_ESTIMATES` for this matter.

src/fheroes2/system/settings.cpp
```diff
@@ -80,7 +80,8 @@ namespace
...
+        GAME_NUMERIC_ESTIMATES = 0x40000000,
```
ihhub, line 84: I suggest to call it `GAME_NUMERIC_ARMY_ESTIMATION_VIEW`.

usidedown:
How about this as a base for the icon in the options menu:
icon_blank
Maybe 2 different variations for each option, E.g. "A-Z" text for the classic way and some roman numerals for the new option.

ihhub:
Hi @zenseii , could you please take a look at the introduced configuration option and it's description and evaluate whether it is suitable in terms of the game?

src/fheroes2/army/army.cpp
```diff
@@ -147,39 +154,48 @@ namespace
...
-    case ArmySize::ARMY_LEGION:
-        str = _( "A legion of\n%{monster}" );
-        break;
-    default:
-        // Are you passing the correct value?
-        assert( 0 );
-        break;
+    // Numeric estimates
+    if ( Settings::Get().isArmyEstimationViewNumeric() ) {
+        str = _( "%{range}\n%{monster}" );
```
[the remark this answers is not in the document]
modo-lv: I called it "estimate" because I see it from an in-universe perspective: the heroes are estimating the size of the opposing army, without knowing the exact numbers. And it feels weird to describe singular terms like "several" and "lots" as ranges. As players we know (or can look up) the numeric range that a term represents, but the word itself isn't a range, it's a singular, inexact amount (an approximation).

As for the layout, I left it the same as canonical estimates because that's how it naturally reads: "a pack of orcs" is a natural English sentence, just like "ten to nineteen orcs". Putting the numbers underneath breaks that flow, making the numeric estimate read as a separate piece of information ("orcs. ten to nineteen."), which adds an unnecessary inconsistency with the canonical UI.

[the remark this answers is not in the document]
ihhub: Hi @modo-lv , we need to consider other languages as well. Usually having:
```
Orcs
10-19
```
implies that most of languages will need no changes. However, with the proposed
```
10-19
Orcs
```
we are facing a situation that we have to have some special cases for translations.

So, on one side as you highlighted we must be to true to the original atmosphere of the game but on the other hand we should be considered about translations. If we aren't sure we can take a look how in other games it is implemented.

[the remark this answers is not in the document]
zenseii: @ihhub has a very good point here. If I'm not mistaken languages like Polish change the plural of a word depending on what number is. In other words. _Orcs_ in "1-4" could be different from _Orcs_ in "20-49". Not sure if this would be avoided if we put the number range after.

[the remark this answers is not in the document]
ihhub: @modo-lv  and @zenseii , I made the change as this pull request was stuck for a very long time and the change is obvious and could be adjusted later. This unblocks us to proceed with accepting the PR and merging it.

zenseii:
Here's the icon gif as suggested by @PusshPop :
Army estimate

The "few" needs to be translated according to the selected language imo. Any translations we have should fit the dialog, but we can just use fit to row function as a precaution.

ihhub:
Hi @zenseii , are we 100% certain that we need this change? If yes let's go with config only solution at the moment and later add a new option. What do you think?

zenseii:
@ihhub, I had already made most of the necessary changes. As discussed in relation to #9971, adding this option was a sort of road-block for that PR and adding those options.

zenseii:
I swapped the phoenix with the centaur to avoid having cycling colors on the phoenix sprite because that made for too busy visuals.

I don't have strong opinions on what creature should be used.

image

zenseii:
@ihhub and @Districh-ru, I'd appreciate your reviews on this. The placement of the option in the dialog is going to change because of #9971 so this is only temporary.

zenseii (approved):
Approved now that the PR contains the changes I requested.

zenseii:
@ihhub, I thought it might fit better there too. One problem is that we already have a lot of options there and can't add another line in the main page of the options:
image

I could add it to the interface settings?
image

ihhub:
@zenseii , interface is a better place than the current one.

@Districh-ru , what do you think?

LeHerosInconnu:
Hello everyone,

Like @zenseii, I think "Army Range" would be more appropriate.

> I don't have strong opinions on what creature should be used.

Personally, I would have chosen this brave Peasant. :D

zenseii:
@LeHerosInconnu, while you're here, do you have any thoughts about the pixel-precision of the alignment of the creature and the army range? And yes I did prefer army range, but I'm willing to let it go to just get this PR wrapped up 😅

LeHerosInconnu:
Hello @zenseii,

> @LeHerosInconnu, while you're here, do you have any thoughts about the pixel-precision of the alignment of the creature and the army range? And yes I did prefer army range, but I'm willing to let it go to just get this PR wrapped up 😅

I would like to have a space of one pixel (or two) between the shadow of the chosen creature's sprite and the top of the numbers below, all centered horizontally and vertically within the frame area.

Position space army range 01

src/fheroes2/agg/agg_image.cpp
```diff
@@ -4958,6 +4958,15 @@ namespace
             }
             break;
         }
+        case ICN::EMPTY_OPTION_ICON_BACKGROUND: {
+            _icnVsSprite[id].resize( 1 );
+
+            fheroes2::Sprite & background = _icnVsSprite[id][0];
+            background.resize( 65, 65 );
```
Districh-ru, line 4988: Since the background image has no shading or transparency we can turn off the transparency layer to speed-up rendering:
`background._disableTransformLayer();`

And I concern if we rally need to have a copy of already loaded image that has no changes. Later in the code we make another copy of this copy and `Blit()` there creature sprite and text.

@zenseii what do you think if we do all the image processing for the two states of the new option in this file instead of doing this in `dialog_game_settings.cpp`. Also currently we have no ability to change this option in the game so (probably in the next PR) we should ass the same icon to `dialog_interface_settings.cpp`. And doing all processing in `agg_image.cpp` will reduce code duplication.

Districh-ru (changes requested):
> @zenseii , interface is a better place than the current one.
> 
> @Districh-ru , what do you think?

I agree that the Interface settings is a better place for the setting of UI indication of creatures' count.

Hi @zenseii, I left one comment with suggestion and question about the need of making `ICN::EMPTY_OPTION_ICON_BACKGROUND` and moving the image processing to `agg_image.cpp`. Could you please take a look when you have time?

PusshPop:
> I swapped the phoenix with the centaur to avoid having cycling colors on the phoenix sprite because that made for too busy visuals.>
@zenseii @Districh-ru 

Would this still be the case if the new color palette that Districh made in #10309 is applied?

Edit: maybe a Hydra is a better alternative? I chose a phoenix because it pops out against the purple background, but a yellow Hydra would do well too I think

zenseii:
@Districh-ru 

> I agree that the Interface settings is a better place for the setting of UI indication of creatures' count.

I will move the option and icon to the interface settings. Unless there are disagreements this means it will not be in the system options.

@PusshPop
> Would this still be the case if the new color palette that Districh made in #10309 is applied?

I'm not familiar with the changes done, but my guess would be no.

> Edit: maybe a Hydra is a better alternative? I chose a phoenix because it pops out against the purple background, but a yellow Hydra would do well too I think

I changed to the hydra. It looks good imo.

image

@LeHerosInconnu, I put 2 pixels distance between the creature sprite and the text to allow for accents in other languages.

src/fheroes2/agg/agg_image.cpp
zenseii: Moved the image generation to `generateDefaultImages()`

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1872,6 +1873,27 @@ namespace
...
+
+            const fheroes2::Sprite & creature = fheroes2::AGG::GetICN( ICN::MONS32, 34 );
+            const int32_t iconBackgroundSize = canonicalBackground.width();
+            fheroes2::Blit( creature, 0, 0, canonicalBackground, ( iconBackgroundSize - creature.width() ) / 2, 8, creature.width(), creature.height() );
+
+            // We need to copy the image before drawing text on it.
+            fheroes2::Sprite & numericBackground = _icnVsSprite[id][1];
+            numericBackground = canonicalBackground;
+
+            fheroes2::Text estimationMode( _( "army|Few" ), fheroes2::FontType( fheroes2::FontSize::SMALL, fheroes2::FontColor::WHITE ) );
```
ihhub: We mustn't put translation here. Once a button is being generated there is no way we can update it. Please check this scenario: run the game in English, then switch to another language and see if the text remains in English for this button.

ihhub: I suggest to generate only one image here and then inside the code add needed text to it.

Districh-ru: Hi @ihhub,  it works fine:

https://github.com/user-attachments/assets/9d026dbc-b784-473d-98b5-22797210d604

I've also updated the code to the code we use for buttons text.

ihhub: Thank you!

ihhub:
@modo-lv , @zenseii and @Districh-ru , thank you very much for making this feature happen!

PR #9690, opened 2025-03-26
Allow to use the hardware keyboard when virtual numpad is opened

oleg-derevenetz:
This is currently the only way to enter numbers using the hardware keyboard if there is more than one numeric input in a window:

https://github.com/user-attachments/assets/d720bbb9-6dc9-4a5f-9c93-2a51b73d5227

Also this PR fixes multiple places where closing the virtual keyboard window using the Enter key in the same time closes the outer window (which waits for the value from the virtual keyboard).


Districh-ru (commented):
Hi @oleg-derevenetz, do we plan to support `DEFAULT_LEFT`, `DEFAULT_RIGHT` and `KEY_DELETE` keys in virtual numpad?

oleg-derevenetz:
> do we plan to support `DEFAULT_LEFT`, `DEFAULT_RIGHT` and `KEY_DELETE` keys in virtual numpad?

May be later. They are not essential for the functionality that most people use. Currently they can use mouse or touch for positioning if they want to.

Districh-ru:
> May be later. They are not essential for the functionality that most people use. Currently they can use mouse or touch for positioning if they want to. 

OK.
The touch positioning is currently broken for this dialog :(
It assumes that text is center alligned.
I'll fix it soon in a new PR.

PR #9691, opened 2025-03-27
Fix close of the caller dialg when closing the Virtual keyboard dialog by a hotkey

Districh-ru:
This PR fixes the issue posted by @oleg-derevenetz in https://github.com/ihhub/fheroes2/issues/9672#issuecomment-2755582485


oleg-derevenetz (approved):
I usually fix this by changing the order in which events are handled, but as far as I understand, it's not always possible to change it if it's complicated enough.

oleg-derevenetz:
@Districh-ru Thank you very much!

PR #9698, opened 2025-03-27
Fix cursor positioning for the Virtual Keyboard

Districh-ru:
This PR fixes setting the cursor position by mouse (or screen tap) for the Virtual Keyboard and also fixes lag in screen update when doing this positioning.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -474,7 +474,7 @@ namespace
                     display.updateNextRenderRoi( { 0, 0, display.width(), display.height() } );
                 }
                 else if ( !filename.empty() && le.MouseClickLeft( textInputRoi ) ) {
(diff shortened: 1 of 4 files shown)


oleg-derevenetz:
Hi @Districh-ru regarding the `renderInputArea()`: there is also another suspicious place - `KeyboardRenderer::resize()` line 170, should there be also used `_output.render( renderInputArea() );` instead of just `renderInputArea();`? If so, then I suggest to start with changing the name for this method, because for an inexperienced person, it may really seem like it performs rendering to the output. Also I can suggest to add the `[[nodiscard]]` attribute to it so the compiler warns when its return value is discarded.

Districh-ru:
Hi @oleg-derevenetz, the `KeyboardRenderer::resize()` does not need to render image on the display because it is done later in the caller function. And the returned `fheroes2::Rect` is not needed in this method so I did not add the `[[nodiscard]]` attribute.
I've renamed the `renderInputArea()` method to `_redrawInputArea()` and added a new method to redraw the input area.

oleg-derevenetz (approved):

oleg-derevenetz:
@Districh-ru Thank you very much!

PR #9700, opened 2025-03-28
Fix gap between buttons in symmetric button dialog

zenseii:
A small mistake made in #9597.

With this PR:
image


oleg-derevenetz (approved):

oleg-derevenetz:
@zenseii Thank you very much!

PR #9702, opened 2025-03-28
Add stat bonus values to Battle Garb of Anduran info

modo-lv:
The description of the Battle Garb of Anduran says "combines the powers of the three Anduran artifacts", but does not specify what those powers are. This PR fixes that.


modo-lv:
The style checker wants me to reformat half the file, even though I only slightly changed one text line. Should I do that?

oleg-derevenetz (approved):
Well, considering that there is currently no universal mechanism for adding multiple bonuses to the artifact description, it will do.

I had doubts about the support of C++11 raw string literals in `xgettext`, but apparently they are supported correctly (it's been so many years, after all!), `fheroes2.pot` and PO files seems to be generated and updated correctly.

src/fheroes2/resource/artifact_info.cpp
```diff
@@ -40,556 +40,546 @@ namespace
...
+                  gettext_noop(
+                      "Rumors of a Griffin of unusual size preying upon the countryside lead you to its cave lair. A quick, brutal fight dispatches the beast, and a search of its foul nest turns up a huge black pearl." ),
+                  {},
+                  {} },
+
+                // Magic Book cannot be located on the original map. The Resurrection map format allows to place it on the map.
+                { gettext_noop( "Magic Book" ),
+                  gettext_noop( "The %{name} enables the hero to cast spells." ),
+                  gettext_noop(
+                      R"(A young man approaches you: "My Lord, allow me to show you my latest invention for spreading knowledge!" You follow the man into his workshop and immediately observe a large apparatus with levers and cranks. "This here is it!" he says eagerly, "The Printing Press." And before you get to say a word, he hands you a Magic Book.)" ),
```
ihhub, line 483: We need to take into an account that this will break the existing translations - we will need to update all of them. Additionally, I suggest to have the same style everywhere. If we are changing to C++11 style then we need to do the same for other places.

oleg-derevenetz: @ihhub why should it break something? It's just a notation, the string itself remains intact. If `xgettext`supports this notation - and apparently it does, at least more or less fresh versions - then nothing should change.

ihhub: I see. In this case I have no objections.

PR #9704, opened 2025-03-29
Cache the truncation symbol width ("...")

Districh-ru:
When displaying the truncated strings in edit fields they are redrawn for every cursor blink and for every string change and the size of the truncation symbol is recalculated each time.
This PR caches this width in std::map for the size of the given font.

### DIFF
--- a/src/fheroes2/gui/ui_text.cpp
+++ b/src/fheroes2/gui/ui_text.cpp
@@ -23,6 +23,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cstddef>
+#include <map>
 #include <memory>
 
 #include "agg_image.h"
@@ -34,6 +35,8 @@ namespace
     const uint8_t hyphenChar{ '-' };
 
     const uint8_t invalidChar{ '?' };
+
+    const std::string truncationSymbol( "..." );
 
     // Returns true if character is a line separator ('\n').
     bool isLineSeparator( const uint8_t character )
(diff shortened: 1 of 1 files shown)


src/fheroes2/gui/ui_text.cpp
```diff
@@ -287,6 +290,23 @@ namespace
...
+    int32_t getTruncationSymbolWidth( const fheroes2::FontType fontType )
+    {
+        // Symbol width depends on font size and not on its color.
+        static std::map<fheroes2::FontSize, int32_t> truncationSymbolWidth;
+
+        if ( auto iter = truncationSymbolWidth.find( fontType.size ); iter != truncationSymbolWidth.end() ) {
+            return iter->second;
+        }
+
+        auto [newIiter, isEmplaced] = truncationSymbolWidth.try_emplace( fontType.size, getLineWidth( reinterpret_cast<const uint8_t *>( truncationSymbol.data() ),
```
oleg-derevenetz, line 303: Hi @Districh-ru In general, it's faster to do an opposite - first, call `try_emplace()` with either value (e.g. zero), get the iterator and the sign of insertion. If our fake value has been inserted - OK, we got the iterator, just set the `iter->second` to the result of the `getLineWidth()` and then return the `iter->second`. If the value has not been inserted - just return the `iter->second` as is. This will save you one map lookup.

Districh-ru: Hi @oleg-derevenetz,  thanks for the optimization suggestion!

src/fheroes2/gui/ui_text.cpp
```diff
@@ -287,6 +290,22 @@ namespace
 
         return std::make_unique<fheroes2::LanguageSwitcher>( language.value() );
     }
+
+    int32_t getTruncationSymbolWidth( const fheroes2::FontType fontType )
+    {
+        // Symbol width depends on font size and not on its color.
+        static std::map<fheroes2::FontSize, int32_t> truncationSymbolWidth;
```
oleg-derevenetz, line 297: @Districh-ru I'm thinking about using of `std::unordered_map` here instead. On the one hand, it's usually faster - just `*buckets[hash(key)]`, i.e. constant complexity (if there are no hash collisions, of course), while `std::map` implies walking over red-black tree (logarithmic complexity). On the other hand, I suppose, there will be not so many elements... What do you think?

Districh-ru: @oleg-derevenetz 
There can be maximum 5 elements but normally for 1 or 2 this is used: small font and normal font. For buttons (pressed/released) and for the large font currently we don't use the truncation symbol.
So I choose map not to calculate hash every time.

oleg-derevenetz (approved):

PR #9708, opened 2025-03-29
Do not ask for additional confirmation if Auto Combat or Quick Combat have been started from the GUI dialog

### LINKED ISSUE #9707: Start auto/quick combat without additional confirmation
### Preliminary checks

Every time a user wants to enable auto/quick combat, he has to make 3 clicks (button "auto", choose the combat type and finally confirm). The step with confirmation seems quite redundant. Confirmation dialogs are typically used to prevent unwanted actions. But in our case the user already clicked the button "auto" and pays attention to what to choose.

Image

Image

Start the desirable combat type without further confirmation.

### Additional info

_No response_

oleg-derevenetz:
close #9707

### DIFF
--- a/src/fheroes2/battle/battle_interface.h
+++ b/src/fheroes2/battle/battle_interface.h
@@ -430,8 +430,8 @@ namespace Battle
         int GetBattleCursor( std::string & statusMsg ) const;
         int GetBattleSpellCursor( std::string & statusMsg ) const;
 
-        void startAutoCombat( const Unit & unit, Actions & actions );
-        void quickCombat( Actions & actions );
+        void _startAutoCombat( const Unit & unit, Actions & actions );
+        void _quickCombat( Actions & actions );
 
         Arena & arena;
         Dialog::FrameBorder border;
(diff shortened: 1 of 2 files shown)


src/fheroes2/battle/battle_interface.h
```diff
@@ -430,6 +430,9 @@ namespace Battle
         int GetBattleCursor( std::string & statusMsg ) const;
         int GetBattleSpellCursor( std::string & statusMsg ) const;
 
+        void startAutoCombat( const Unit & unit, Actions & actions );
+        void quickCombat( Actions & actions );
```
Districh-ru, line 434: Since these are private methods of the `Interface` class, let's add `_` to the beginning of their names.
The other private methods in this class do not start from `_` but they are from the old code and we will rename them later...

And what do you think about renaming `quickCombat()` to `_startQuickCombat()`?

Districh-ru (approved):
Hi @oleg-derevenetz, looks good to me.
I agree that one dialog is enough to make the player think about what he is doing. :)
I left one optional suggestion. Could you please check it.

src/fheroes2/battle/battle_interface.h
oleg-derevenetz: > And what do you think about renaming `quickCombat()` to `_startQuickCombat()`?

While "auto combat" can be both started and stopped, quick combat can only be "performed", so I don't think there is a need :) The same is for events: `EventStartAutoCombat()`, but `EventQuickCombat()`.

Districh-ru: Sure.

Issue #27, opened 2019-09-07
Replace (void) by empty () statement


ihhub:
Not sure what was an intention behind but original code is filled with functions containing `void` as input parameter. This brings no value to code and just complicates reading. We want to remove replace by empty parameter function code as `()`.

kgorecki:
As a specialist for that kind of tasks, I'll take it if no one else dares ;}

ihhub:
@kgorecki we already such pull request :)

kgorecki:
Oh boy, I was sleepy when I did that... :}

PR #9712, opened 2025-04-01
Swedish Translation Update (#27)

Mr-Bajs:
Keeping it up to date.

*Translated and updated changed and new strings.
*Changed resource generating artifacts descriptions from plural to singular due since they only add a single resource, except gold.


oleg-derevenetz:
@Mr-Bajs Thank you very much!

PR #9713, opened 2025-04-01
Android: set the application category to "game"

oleg-derevenetz:
Android 16 introduces feature "adaptive layouts" that will [ignore](https://developer.android.com/about/versions/16/behavior-changes-16?hl=en#adaptive-layouts) the orientation, resizability, and aspect ratio restrictions when app is running on screens with certain size, unless the app is a game. fheroes2 is not targeting the Android 16 yet, but it eventually will, so care should be taken to set the correct application category in advance.


ihhub:

PR #9719, opened 2025-04-05
Improve the input of numbers from the hardware keyboard

### LINKED ISSUE #1631: Separation of troops by entering the exact number: enhancement

oleg-derevenetz:
close #1631
related to #9672

Key changes compared to `master` branch:

1. The value starts to be entered from scratch when the corresponding dialog window is opened or when value has been changed not using keyboard events.
2. The Backspace and Delete keys both clear the input buffer completely (the value starts to be entered from scratch - exactly as in paragraph 1). This makes the behavior more sane in a number of scenarios (especially when the range of acceptable values does not include zero).
3. The minus sign can only be entered at the beginning of the number input. The corresponding key is ignored when the number is already partially entered from the keyboard. This makes the behavior more sane when the range of acceptable negative values does not match the range of acceptable positive values.
4. The input visually stops completely if the entered value does not fall within the valid range. For instance, if the range is `[0, 99]` (with the initial value `0`), and we press keys `1`, `0`, `0`, then the entered value will be as follows: `1`->`10`->`10` (not `99`). However, if the range is `[20, 1000]` (with the initial value `20`), and we press the same keys `1`, `0`, `0`, then the entered value will be as follows: `20`->`20`->`100`. In other words, you can now enter any valid value if you just type it completely on the keyboard. In the `master` branch, in certain cases, it is just impossible to enter a perfectly valid value using only the keyboard.

Still no perfect WYSIWYG, but IMHO it's much better than what we currently have in `master`.

### DIFF
--- a/src/fheroes2/gui/ui_tool.cpp
+++ b/src/fheroes2/gui/ui_tool.cpp
@@ -812,37 +812,18 @@ namespace fheroes2
     {
         assert( min <= max );
 
-        const auto parseValue = [min, max, &valueBuf = std::as_const( valueBuf )]() -> std::optional<int32_t> {
-            if ( valueBuf.empty() ) {
-                return {};
-            }
-
-            if ( valueBuf == "-" ) {
-                return {};
-            }
-
-            try {
-                if ( const int32_t value = std::stoi( valueBuf ); value == std::clamp( value, min, max ) ) {
-                    return value;
-                }
-            }
-            catch ( std::out_of_range & ) {
-                return {};
-            }
-
-            return {};
-        };
-
         const LocalEvent & le = LocalEvent::Get();
 
         if ( !le.isAnyKeyPressed() ) {
             return {};
         }
 
+        const int32_t zeroBufValue = std::clamp( 0, min, max );
+
         if ( le.isKeyPressed( fheroes2::Key::KEY_BACKSPACE ) || le.isKeyPressed( fheroes2::Key::KEY_DELETE ) ) {
             valueBuf.clear();
 
-            return std::clamp( 0, min, max );
+            return zeroBufValue;
         }
 
         if ( le.isKeyPressed( fheroes2::Key::KEY_MINUS ) || le.isKeyPressed( fheroes2::Key::KEY_KP_MINUS ) ) {
@@ -850,13 +831,13 @@ namespace fheroes2
                 return {};
             }
 
-            if ( !valueBuf.empty() ) {
+            if ( !std::all_of( valueBuf.begin(), valueBuf.end(), []( const char ch ) { return ( ch == '0' ); } ) ) {
                 return {};
             }
 
-            valueBuf.push_back( '-' );
+            valueBuf.insert( 0, 1, '-' );
 
-            return parseValue();
+            return zeroBufValue;
         }
 
         if ( const std::optional<char> newDigit = [&le]() -> std::optional<char> {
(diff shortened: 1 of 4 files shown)


oleg-derevenetz:
Hi @Districh-ru I would appreciate your opinion on this concept.

Districh-ru:
Hi @oleg-derevenetz, looks good to me!
Now the input is more clear to the user. I has no "auto-clamp" and acts mostly as in the master build: you open a dialog, type the number and press enter to confirm.
Clearing the number by Backspace/Del is also OK because values are mostly short so retyping them from the beginning will not be a problem. :)

I found one issue: in example if `max == 52` and you enter `55` only the first `5` is processed and shown, then next `5` is ignored and it is right. But then if I press `0` (to append it to `5` and have `50`) nothing happens because the string `55` is stored in `valueBuf`. What do you thing about removing the last digit from `valueBuf` if this value is more than `max` for positive values and less than `min` for negative values?

Also currently I don't see a good way to display the just typed `-` or an entered value that is less than `min` to indicate to the user that the keystroke was processed...

oleg-derevenetz:
Hi @Districh-ru

> Also currently I don't see a good way to display the just typed `-` or an entered value that is less than min to indicate to the user that the keystroke was processed...

With the current state of things (with `ValueSelectionDialogElement` using `int` internally) WYSIWYG is impossible. We need to pretty much rework the logic of how such elements work for this. I don't think it's necessary to do this now. Maybe later, if someone has the desire and ideas to do it.

src/fheroes2/gui/ui_tool.cpp
```diff
@@ -805,43 +808,79 @@ namespace fheroes2
...
+        if ( le.isKeyPressed( fheroes2::Key::KEY_MINUS ) || le.isKeyPressed( fheroes2::Key::KEY_KP_MINUS ) ) {
+            if ( min >= 0 ) {
+                return {};
+            }
+
+            if ( !std::all_of( valueBuf.begin(), valueBuf.end(), []( const char ch ) { return ( ch == '0' ); } ) ) {
+                return {};
+            }
+
+            valueBuf.insert( 0, 1, '-' );
```
Districh-ru, line 838: We get here only if `valueBuf` is empty or contains only zeros and we want to make the next input digit negative. What do you think about just doing `valueBuf = "-";` ?

Districh-ru (commented):
@oleg-derevenetz, LGTM.
I left one optional suggestion. Could you please check it when you have time.

src/fheroes2/gui/ui_tool.cpp
oleg-derevenetz: @Districh-ru Yes, I suppose it definitely won't get any worse.

Districh-ru (approved):

PR #9731, opened 2025-04-11
Text input cursor change from '_' to '|' (with zero width)

### LINKED ISSUE #9358: Message input window, blinking cursor makes the last symbol jump to the next line
### Preliminary checks

https://github.com/user-attachments/assets/6db9ff30-c15e-4bdb-9790-e1f80c60ea46

The last symbol in the line jumps to the next line.
This causes cursor appear on the next line, when we click on the right edge area of the line when we click at any place of the full text line.

There could be a reserved free space for the blinking cursor in the line or the cursor itself ( _ ) could be switched to a thinner one ( | ).

### Additional info

_No response_

### LINKED ISSUE #9490: Invalid character displayed for the cursor when setting the game language to french
### Preliminary checks

### Platform

Linux

### Describe the bug

Whenever the language of the game is changed to french. The (DEL -> \x7F) character displayed for the cursor in any input textbox is not valid.

https://github.com/user-attachments/assets/10188c39-04a7-41a3-92b5-f3d46e5a7021

  

### Save file

This bug can be reproduced on any map / save.

### Additional info

_No response_

Districh-ru:
Close #9490, close #9358

This PR changes the input cursor from `_` to `|`. It also removes the need to use the `'\x7F'` character for "displaying" the hidden cursor.
The new cursor has zero-sized width so it does not affect the real text width. And when animation its blinking the text is not fully redrawn - only the cursor area is restored or the cursor symbol is drawn again.

https://github.com/user-attachments/assets/05ed8ac8-443b-4fbd-8533-28487b7b77ca

https://github.com/user-attachments/assets/bbf72d2a-5e45-40a0-b2b1-03788825f9fa


src/fheroes2/editor/editor_save_map_window.cpp
```diff
@@ -455,37 +462,51 @@ namespace Editor
...
+            if ( !needFileNameRedraw && Game::validateAnimationDelay( Game::DelayType::CURSOR_BLINK_DELAY ) ) {
                 isCursorVisible = !isCursorVisible;
-                needFileNameRedraw = true;
+
+                if ( isCursorVisible ) {
+                    textInput.drawCursor( fileNameRoi.x, fileNameRoi.y + 4, display, fileNameRoi );
+                }
+                else {
+                    textCursorRestorer.restore();
+                }
```
ihhub, line 476: This seems like the code very similar to `MovableSprite` class. What do you think if we use it instead of the proposed code?

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -238,14 +237,35 @@ bool Dialog::inputString( const fheroes2::TextBase & title, const fheroes2::Text
...
+
+        if ( !isMutlilineText ) {
+            textPos.x = textInputArea.x + ( textInputArea.width - textInput.width() ) / 2;
+        }
+
+        textInput.drawInRoi( textPos.x, textPos.y, display, textInputArea );
+
+        const fheroes2::Rect & cursorRoi = textInput.getCursorArea();
+        textCursorRestorer.update( cursorRoi.x + textPos.x, cursorRoi.y + textPos.y, cursorRoi.width, cursorRoi.height );
+        textInput.drawCursor( textPos.x, textPos.y, display, textInputArea );
```
ihhub, line 264: Based on my comment below it feels like this is the same as `MovableSprite` class but with an extra logic related to sprite position based on the text. Should we utilize `MovableSprite` class (derived a child class from it) in order to remove duplicated code to ease code readability?

src/fheroes2/gui/ui_text.cpp
```diff
@@ -718,6 +634,248 @@ namespace fheroes2
...
+            // There is no need to fit the text. Reset text fitting values.
+            _textBeginPos = 0;
+
+            return;
+        }
+
+        constexpr int32_t cursorToBorderDistance = 4;
+
+        // If the cursor is to the left of the TextBox.
+        _textBeginPos = ( _cursorPosition > cursorToBorderDistance ) ? std::min( _textBeginPos, _cursorPosition - cursorToBorderDistance ) : 0;
```
ihhub, line 738: I am confused by this logic. This is the only place in the code where we set `_textBeginPos` to something not equal to 0. Let's analyze the code:
```cpp
if ( _cursorPosition > cursorToBorderDistance ) {
    _textBeginPos = std::min( _textBeginPos, _cursorPosition - cursorToBorderDistance );
}
else {
    _textBeginPos = 0;
}
```
in other words the only case when we set the value to something is not 0 is:
```cpp
if ( _cursorPosition > cursorToBorderDistance ) {
    _textBeginPos = std::min( _textBeginPos, _cursorPosition - cursorToBorderDistance );
}
```
so, we are setting to a negative value here. Is it intentional?

src/fheroes2/gui/ui_text.cpp
```diff
@@ -718,6 +634,248 @@ namespace fheroes2
...
+
+        if ( _textBeginPos > 0 && _textBeginPos < cursorToBorderDistance
+             && getLineWidth( textData, _textBeginPos, charHandler, true ) <= truncatedMaxWidth - currentWidth ) {
+            // The beginning of the text fits the given width. Update _textBeginPos to remove truncation at the beginning.
+            _textBeginPos = 0;
+            _textLength = getMaxCharacterCount( textData, originalTextSize, charHandler, truncatedMaxWidth );
+        }
+    }
+
+    int32_t TextInput::getSingleTruncationSymbolWidth() const
```
ihhub, line 812: This method should be `static` as it doesn't use any class' members. Moreover, we could just make `getTruncationSymbolWidth()` as a standalone function within `ui_text.h` without the need of having this method at all.

src/fheroes2/gui/ui_text.h
```diff
@@ -274,23 +274,114 @@ namespace fheroes2
     class TextInput final : public Text
```
ihhub, line 274: I am getting more convinced that this class should be moved to `ui_tool.h|cpp` files and utilize **MovableSprite** class as a member since we use `TextInput` class always together with `ImageRestorer`.

src/fheroes2/gui/ui_text.h
```diff
@@ -274,23 +274,114 @@ namespace fheroes2
...
+        int32_t getSingleTruncationSymbolWidth() const;
+
     private:
-        size_t _cursorPosition{ 0 };
-        size_t _textOffsetX{ 0 };
+        // Update the area of text occupied by cursor and fit the text if the `_autoFitToWidth` is > 0.
+        void _updateCursorArea();
+
+        fheroes2::Rect _cursorArea;
+        int32_t _cursorPosition{ 0 };
```
ihhub, line 380: Is **_cursorPosition** a character position related to the text? We have **_cursorArea** and **_cursorPosition** which sound very similar.

src/fheroes2/gui/ui_text.h
```diff
@@ -274,23 +274,114 @@ namespace fheroes2
...
+        int32_t _textBeginPos{ 0 };
```
ihhub, line 381: It seems like it is a position of the first visible character in the text. Is it correct? If yes please rename this member to have better understanding what this variable is for.

src/fheroes2/gui/ui_text.h
```diff
@@ -274,23 +274,114 @@ namespace fheroes2
...
+        int32_t _textLength{ 0 };
```
ihhub, line 382: Is it a visible text length?

src/fheroes2/gui/ui_text.h
```diff
@@ -274,23 +274,114 @@ namespace fheroes2
...
+        int32_t _autoFitToWidth{ 0 };
```
ihhub, line 383: Should this member be something like `_maxVisibleTextWidth`?

src/fheroes2/gui/ui_text.h
```diff
@@ -274,23 +274,114 @@ namespace fheroes2
...
+        int32_t _multilineMaxWidth{ 0 };
```
ihhub, line 384: Could you please explain what is the different between **_autoFitToWidth** and **_multilineMaxWidth**? It seems to be they both store maximum width to be shown. If yes and we still want to differentiate multi-line text and single-line text should we just use a boolean variable for this matter?

ihhub (changes requested):
Hi @Districh-ru , I put several comments in this pull request. Could you please check them when you have time?

src/fheroes2/dialog/dialog_selectcount.cpp
```diff
@@ -314,7 +334,7 @@ bool Dialog::inputString( const fheroes2::TextBase & title, const fheroes2::Text
 
         if ( le.MouseClickLeft( buttonVirtualKB.area() ) || ( isInGameKeyboardRequired && le.MouseClickLeft( textEditClickArea ) ) ) {
             if ( textLanguage.has_value() ) {
-                const fheroes2::LanguageSwitcher switcher( textLanguage.value() );
+                const fheroes2::LanguageSwitcher switcher( *textLanguage );
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @ihhub I suppose the reason is that `std::optional<T>::value()` performs an additional check that there is a value and throws an exception if there is no value, while `std::optional<T>::operator*()` performs no checks. Since we have already made sure that the value is in place (on the previous line), there is no point in checking this once again.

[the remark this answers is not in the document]
Districh-ru: Hi @ihhub,
@oleg-derevenetz is absolutely right. I did these changes because in the upper line we do check `textLanguage.has_value()` so there is no need to check it once more. And it is also visible that `textLanguage` is not a simple pointer (or iterator) because of this upper line. :)
I'll leave a comment about that we can simply deference the `std::optional` because it certainly has value as the check `.has_value()` was successfully passed.

src/fheroes2/gui/ui_text.cpp
Districh-ru: This is a first part of "text fitting": we determine were the visible text should begin to show the cursor and text around it.
I'll add a comment about this.
Also there will be no negative value because if `_cursorPosition > cursorToBorderDistance` then `_cursorPosition - cursorToBorderDistance` is positive non-zero value and otherwise we set `_textBeginPos` to zero.

src/fheroes2/gui/ui_text.h
Districh-ru: Sadly the the `TextInput` class uses some functions from the `ui_text.cpp` anonymous namespace so we cannot simply move it.

src/fheroes2/gui/ui_text.h
Districh-ru: Now it is `_cursorPositionInText`

src/fheroes2/gui/ui_text.h
Districh-ru: You are right, the variables' names are not concrete. Let it be `_visibleTextBeginPos`.

src/fheroes2/gui/ui_text.h
Districh-ru: Sure.
=> `_visibleTextLength`

src/fheroes2/gui/ui_text.h
Districh-ru: Sure, I've renamed the first member to `_maxTextWidth` and the second is now ` bool _isMultiLine`.
It is not absolutely the maximum visible widths for the text, text is also limited by the truncation symbols width so I did not add the `visible` word to the name of the first variable.

Districh-ru:
@ihhub,
I've added the `TextInputField` class to `ui_tool.h`/`.cpp` and did the test field and cursor rendering there.
It helped to remove some methods from `TextInput` class.
Also the `getTextInputCursorPosition()` funcion is now made a private member of `TextInputField` class - this made possible to make `getTextLineInfos()` protected.
The rendering can be optimized and I left a TODO to do all `cursorPosition` related and keyboard related operations in `TextInputField` class. But lets do it in a new PR because this one is big enough and let's focus on polishing it.

src/fheroes2/dialog/dialog_selectcount.cpp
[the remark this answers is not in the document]
ihhub: Thanks for the explanation! Let's keep the change then.

ihhub:
Hi @Districh-ru , could you please resolve merge conflicts?

ihhub:
Hi @Districh-ru , I modified the code to make TextInput class as a member of TextInputFiled class. I also did small codecleanup. Could you please take a look at the changes and check whether they make sense?

ihhub (approved):

Districh-ru:
> Hi @Districh-ru , I modified the code to make TextInput class as a member of TextInputFiled class. I also did small codecleanup. Could you please take a look at the changes and check whether they make sense?

Hi @ihhub, many thanks for updating the code, it's much better structured now!
I also did small clean-up.
After this PR is merged we can optimize rendering (do not redraw the text when the cursor is moved without changing the text and `_visibleTextBeginPos`. And we can add keyboard processing  and cursor positioning to the `TextInput` class in new PR.

ihhub:

Districh-ru:
Hi @zenseii , the DEL symbol is not used animore for the cursor, the original ASCII symbol '|' is used for it.
Probably this symbol is used in French localization ...
I'll take a look when I'll be home.

Districh-ru:
In future we would like to use more ASCII characters so IMHO the best solution is to change all "replaced" ASCII characters to their original ones from CP1252: #9795

PR #9738, opened 2025-04-15
GitHub Actions: run the CodeChecker and IWYU workflows only if some C/C++ files were changed

oleg-derevenetz:
Related to #9721

Since we have a complex structure of interdependent reusable workflows, we cannot just use the `paths` option as proposed in #9721. However, we can perform the relevant filtering ourselves by writing quite a bit of code.

Unlike build workflows, which may depend on files of a diverse kind (C/C++, Java, Gradle, shell, Python, translations, even documentation, which is bundled into resulting artifacts), CodeChecker and IWYU workflows should only be run if the source files related to C/C++ were changed, so this restriction looks relatively safe. Also the CodeChecker is the slowest workflow we have.

### DIFF
--- a/.github/workflows/codechecker.yml
+++ b/.github/workflows/codechecker.yml
@@ -2,12 +2,17 @@ name: CodeChecker
 
 on:
   workflow_call:
+    inputs:
+      run:
+        required: true
+        type: boolean
 
 permissions: {}
 
 jobs:
   codechecker:
     name: CodeChecker
+    if: ${{ inputs.run }}
     runs-on: ubuntu-24.04
     timeout-minutes: 60
     defaults:
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -2,12 +2,17 @@ name: IWYU
 
 on:
   workflow_call:
+    inputs:
+      run:
+        required: true
+        type: boolean
 
 permissions: {}
 
 jobs:
   iwyu:
     name: IWYU
+    if: ${{ inputs.run }}
     runs-on: ubuntu-24.04
     timeout-minutes: 30
     defaults:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -47,16 +47,20 @@ jobs:
     uses: ./.github/workflows/code_style_check.yml
   codechecker:
     name: CodeChecker
-    if: ${{ needs.dispatcher.outputs.runCxxAnalyzers == 'YES' }}
     needs:
+    - dispatcher
     - style
     uses: ./.github/workflows/codechecker.yml
+    with:
+      run: ${{ needs.dispatcher.outputs.runCxxAnalyzers == 'YES' }}
   iwyu:
     name: IWYU
-    if: ${{ needs.dispatcher.outputs.runCxxAnalyzers == 'YES' }}
     needs:
+    - dispatcher
     - style
     uses: ./.github/workflows/iwyu.yml
+    with:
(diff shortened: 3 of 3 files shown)


justinbburris:
Have you seen this? We use this path filter at my workplace and it works very nicely

https://github.com/dorny/paths-filter

oleg-derevenetz:
Hi @justinbburris

> Have you seen this? We use this path filter at my workplace and it works very nicely
> 
> https://github.com/dorny/paths-filter

To be honest, if there is a choice between writing a bit of code and using a 3rd-party Action, I'd prefer to write code myself. There were precedents when I even had to completely [rewrite](https://github.com/platisd/clang-tidy-pr-comments/pull/48) the third-party Actions to make such an Action work properly, and not everyone can be as kind and interested in fixing the issues as the author of that Action. It also happens that the authors no longer support their own projects for various reasons. I believe it's easier to maintain a small piece of my own code without extensive 3rd-party dependencies.

oleg-derevenetz:
Hi @ihhub if we go along with this PR, we will need to uncheck the "required" from CodeChecker and IWYU workflows, since they now will not be run for all PRs.

justinbburris:
> Hi @ihhub if we go along with this PR, we will need to uncheck the "required" mark from CodeChecker and IWYU workflows, since now they will no longer be run for all PRs.

FYI - It is possible to require those checks and skip the job runs

Per the [docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success". For more information, see [Using conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution).

See this [PR](https://github.com/justinbburris/fheroes2/pull/2) for an example. 

Here's an [example](https://github.com/justinbburris/fheroes2/pull/3) of changing the code and having them run (and fail) as well

### Example required check configuration
Screenshot 2025-04-17 at 7 28 37 PM

### Skipped required jobs but still green merge button 🙌 
Screenshot 2025-04-17 at 7 37 02 PM

### Failing required jobs and blocked merge button ❌ 
Screenshot 2025-04-17 at 7 55 46 PM

oleg-derevenetz:
Hi @justinbburris

> FYI - It is possible to require those checks and skip the job runs
> 
> Per the [docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks)
> 
>> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success". For more information, > see [Using conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-> execution).
> 
> See this https://github.com/justinbburris/fheroes2/pull/2 for an example.

Yes, it works better, thank you! However, it's a bit odd that jobs that were skipped due to conditionals are considered successful at a lower level (within the reusable workflow), but not at the higher level (within the parent workflow that runs the reusable workflow as a job). After all, they are all jobs. Apparently, this is some kind of overlook in the implementation of the reusable workflows.

ihhub:
@oleg-derevenetz , well, I like when things are resolved by themselves without me doing anything :)

ihhub (approved):

ihhub:
@oleg-derevenetz , many thanks for this improvement!

PR #9739, opened 2025-04-16
Clang-Tidy comments workflow: simplify the logic a bit

oleg-derevenetz:
While I'm here, let's modernize the `clang_tidy_comments.yml` workflow a bit and use a more terse syntax in some places.


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks for the changes!

PR #9741, opened 2025-04-19
Fix the resumption of music playback for Emscripten builds

oleg-derevenetz:
Currently, the resumption of music playback doesn't work in the non-multithreaded Emscripten builds (music tracks always start playing from the beginning), because the current logic in `master` branch relies on the use of inter-thread synchronization to ensure the correct order of actions. In the non-multithreaded Emscripten builds there are no threads, so this logic doesn't work as expected.


ihhub (approved):

ihhub:
@oleg-derevenetz , a bunch of thanks for the changes!

PR #9742, opened 2025-04-19
Update emscripten build to generate packaged webserver binary

justinbburris:
## Overview
This PR introduces a more user-friendly build system for the fheroes2 emscripten build, making it easier to distribute and play the game through a browser interface. The system includes a Node.js-based launcher that packages the game files and serves them locally, along with enhanced Emscripten build configurations.

## Status
This is being opened as a draft PR to demonstrate the concept of wrapping the app into a web-friendly build. The goal is to enable quick iteration and testing without manual file copying. Further testing and refinement may be needed, particularly for Linux and Windows platforms.

> [!IMPORTANT]
> This PR is in a draft state to show a high level proof of concept. Likely, the node pkg emscripten build should be broken apart from the emscripten app build and the PR would need to be cleaned up.

### HTTP Launcher
- A Node.js-based launcher that serves game files locally through a lightweight HTTP server
- Automatically opens the default browser to the game interface
- Handles port conflicts and provides clear error messages
- Packages all necessary game files and assets into standalone executables

### Build System
- Automated build scripts with proper CPU core detection
- macOS-specific build support via Brewfile for Emscripten dependencies
- Expanded file upload whitelist to include 'anim' and 'files/save' directories

### Prerequisites
- For macOS: Emscripten dependencies (automated via Brewfile)
- Node.js and npm for launcher packaging
- Emscripten SDK

### Build Commands
1. Build the Emscripten version:
   ```bash
   ./script/emscripten/build.sh
   ```

## How to Run
The packaged launcher will:
1. Start a local HTTP server on port 8000
2. Automatically open your default browser to `http://localhost:8000`
3. Serve the game files and handle asset loading

## Technical Details
- The launcher uses Node.js's `pkg` tool to create standalone executables
- Assets are bundled with the executable for easy distribution
- The server handles both game files and static assets from different directories
- Proper cleanup on server shutdown (Ctrl+C handling)

## Limitations and Trade-offs
- Currently spawns a local HTTP server and opens a browser window
- Port 8000 must be available for the server to start
- Limited testing on Linux and Windows platforms - build process suitability needs investigation

## Future Considerations
- Break out the "fully wrapped" app from the emscripten build
- Verify/Improve Windows and Linux build process compatibility
- Explore performance optimizations for the browser-based interface
- Add configuration options for server port and browser launch behavior

https://github.com/user-attachments/assets/9433ee63-1b7f-4c5d-b87e-4d4bd7bb605c

### DIFF
--- a/Makefile.emscripten
+++ b/Makefile.emscripten
@@ -25,9 +25,9 @@
 # FHEROES2_WITH_THREADS: build with multithreading support
 # FHEROES2_DATA: set the built-in path to the fheroes2 data directory (e.g. /usr/share/fheroes2)
 
-.PHONY: all clean translations
+.PHONY: all clean translations package-launchers
 
-all: fheroes2.js
+all: fheroes2.js package-launchers
 
 fheroes2.js: translations
 	$(MAKE) -C src/dist PLATFORM=emscripten
@@ -39,7 +39,14 @@ fheroes2.js: translations
 translations:
 	$(MAKE) -C files/lang
 
+package-launchers: fheroes2.js
+	@echo "Packaging HTTP launchers..."
+	@mkdir -p src/dist/http-launcher/
+	cd src/http-launcher && npm install && npm install -g pkg && pkg . --targets node16-win-x64,node16-macos-x64,node16-linux-x64 --output ../dist/http-launcher/fheroes2-launcher
+	@echo "Launcher packages created in src/dist/http-launcher/"
+
 clean:
 	$(MAKE) -C src/dist clean
 	$(MAKE) -C files/lang clean
 	-rm -f fheroes2.data fheroes2.js fheroes2.wasm*
+	-rm -rf src/dist/http-launcher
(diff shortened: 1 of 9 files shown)


oleg-derevenetz:
> Likely, the node pkg emscripten build should be broken apart from the emscripten app build and the PR would need to be cleaned up.

Yes, I also believe that there is no need to mix the WASM binary build process using the GNU Make infrastructure and the http server installation/setup. Perhaps someone really needs to run a local http server for tests, but in general this is a separate task from the build, and unlike the build, it is not always needed. Perhaps this should be moved to a separate script.

justinbburris:
> > Likely, the node pkg emscripten build should be broken apart from the emscripten app build and the PR would need to be cleaned up.
> 
> Yes, I also believe that there is no need to mix the WASM binary build process using the GNU Make infrastructure and the http server installation/setup. Perhaps someone really needs to run a local http server for tests, but in general this is a separate task from the build, and unlike the build, it is not always needed (I may already have an HTTP server for deploying WASM binaries). Perhaps this should be moved to a separate script.

Yeah we're in alignment there - cool. I'll shape that to be a bit more in the form of dev tooling, vs release tooling.

Issue #8795, opened 2024-06-01
Adventure map, sometimes the hero's movement animation is stopped when he enters the stone liths


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, sometimes the hero's movement animation is stopped when he enters the stone liths, then resumed, before the hero is transported to the other stone liths.

In fheroes2:

![vlcsnap-2024-06-01-20h37m06s251](https://github.com/ihhub/fheroes2/assets/43583994/9543c80d-ef9d-4ced-8361-c3224354689f)

https://github.com/ihhub/fheroes2/assets/43583994/01259c80-5fa6-4788-9adb-5052d480b0c5

### Save file

Here is the save file:
[Before stone liths 01.zip](https://github.com/user-attachments/files/15522931/Before.stone.liths.01.zip)

oleg-derevenetz:
The reason here is exactly the same as in the case of other similar delays, when some sound effect (teleport sound in this case) is waiting for the start of the new MIDI track playback (Swamp MIDI track playback in this case, since hero is briefly enters the Swamp tile before visiting the Stone Liths). There is no delay when music type is switched from MIDI to External.

Issue #8379, opened 2024-02-08
Combat screen, creature animation freeze during movement at start of combat


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, there is a creature animation freeze during movement at start of combat.

In fheroes2:

The second troop of wolves freezes as it moves.

https://github.com/ihhub/fheroes2/assets/43583994/b73ffd00-d9e3-4f95-9caa-347625438e5d

The troop of phoenixes freezes at the start of its movement.

https://github.com/ihhub/fheroes2/assets/43583994/534f23b7-301e-46af-9802-b5ae9e79c9c9

The troop of gargoyles freezes as it moves.

https://github.com/ihhub/fheroes2/assets/43583994/d5b61a99-b5f2-4a37-9bb2-6a9de211731e

The first troop of vampire lords freezes when attacking.

https://github.com/ihhub/fheroes2/assets/43583994/b5d319e5-2496-47c9-bfb0-12997cb72dc6

With all audio settings to "off":
In fheroes2:

![All audio off 01](https://github.com/ihhub/fheroes2/assets/43583994/79b6daf4-19c2-4efc-bf10-f12c3ba725de)

The first troop of vampire lords freezes at the start of its movement.

https://github.com/ihhub/fheroes2/assets/43583994/43bc232f-f05e-4f9c-82c1-e2c290e3225d

### Save file

Here are the save files:
[Creature animation freeze 01.zip](https://github.com/ihhub/fheroes2/files/14209720/Creature.animation.freeze.01.zip)

### Additional info

Related closed issue: https://github.com/ihhub/fheroes2/issues/7197.

oleg-derevenetz:
I'd say that the reason is the same as in #8376 and #8375 - WAV sound effect playback is started with a noticeable delay on this particular PC, which shouldn't be, because WAV playback is very cheap.

oleg-derevenetz:
Marking this as an audio-related issue but it's just a guess because I cannot reproduce this on my devices.

oleg-derevenetz:
[...]

Issue #8376, opened 2024-02-08
Adventure map, there is a slowdown before the window opens in some places


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Windows 7 SP1 - 64 bits

In fheroes2, there is a slowdown before the window opens in some places.
In fheroes2:

https://github.com/ihhub/fheroes2/assets/43583994/f470e296-575d-45f8-bf0e-98b9c3e25e32

https://github.com/ihhub/fheroes2/assets/43583994/48229de4-e41e-4e60-becb-c90c0c33f0c6

### Save file

Here are the save files:
[Before shrine 01.zip](https://github.com/ihhub/fheroes2/files/14208710/Before.shrine.01.zip)

### Additional info

Windows 7 SP1 - 64 bits

oleg-derevenetz:
Although I can't reproduce this problem on any of my devices, it seems to be due to the fact that for some reason the start of playing MP3/OGG/FLAC music effect file on this PC is so slow for some reason that it causes a noticeable delay. What happens if you switch the music from External to MIDI in the settings?

LeHerosInconnu:
Hello @oleg-derevenetz,

> Although I can't reproduce this problem on any of my devices, it seems to be due to the fact that for some reason the start of MP3/OGG/FLAC music effect playback on this PC is so slow that it causes a noticeable delay. What happens if you switch the music from External to MIDI in the game settings?

The "Music Type" setting is set to "MIDI" with the default fheroes2 soundfonts, and there are no MP3/OGG/FLAC files present when the problem described was detected.

oleg-derevenetz:
> How about setting the volume to zero? I believe that should skip playing any sounds so that we can verify if that is the issue.

IIRC WAV sound effects are played independently of audio effects volume settings. WAV playback itself is very cheap in comparison with MP3, not to mention MIDI.

oleg-derevenetz:
[...]

Issue #8375, opened 2024-02-08
Adventure map, sound of the horse's hooves continues after the horse has reached its destination


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Windows 7 SP1 - 64 bits

In fheroes2, the sound of the horse's hooves continues after the horse has reached its destination.

In fheroes2:
Activate the sound of the video.

https://github.com/ihhub/fheroes2/assets/43583994/95ced611-2334-4eaa-a164-f1d4a741376b

### Save file

Here is the save file:
[Horse sound after move 01.zip](https://github.com/ihhub/fheroes2/files/14208504/Horse.sound.after.move.01.zip)

### Additional info

Windows 7 SP1 - 64 bits

Branikolog:
Hi, @LeHerosInconnu 
Are you using MIDI? I believe this issue is valid only for MIDI sound, while for external music I cannot reproduce this problem.

LeHerosInconnu:
Hello @Branikolog,

> Hi, @LeHerosInconnu Are you using MIDI? I believe this issue is valid only for MIDI sound, while for external music I cannot reproduce this problem.

Yes, MIDI is used when the problem is detected.
I did a test with .ogg files in the MUSIC folder and "Music Type" set to "External" and the problem is no longer present.
There must have been a change in MIDI management at some point.

oleg-derevenetz:
> There must have been a change in MIDI management at some point.

This was always the case. MIDI playback (more precisely, MIDI playback startup) is slow in SDL_mixer. Here is the video of fheroes2 1.0.0 released Dec 21 2022 (more than a year ago):

https://github.com/ihhub/fheroes2/assets/32623900/9204a89e-f302-48f2-8614-a84acbc0dd73

As you can see, there is the same pause that caused by delay of playing the WAV musical effect for this Shrine caused by MIDI soundtrack change due to the change of terrain. This is the same place, but using the Windows built-in MIDI (not FluidSynth MIDI) - i removed the soundfont file to force SDL_mixer to switch to the Windows MIDI:

[...]

PR #9744, opened 2025-04-20
MIDI: allow to pick up custom timidity.cfg

oleg-derevenetz:
This PR is a part of the effort to fix the #8375,  #8376, #8379, #8795.

The idea is as follows: while Fluidsynth MIDI backend operates with heavy soundfont file (which it apparently loads every time it tries to play a MIDI file), Timidity MIDI backend operates with relatively small GUS patches (and apparently it does not load those instruments that are not needed to play a specific MIDI file or maybe uses other optimizations), which minimizes the delay at the start of the playback.

To test this idea, I made a special build of `SDL_mixer` with Timidity MIDI backend on Windows and used it along with this PR. The results are shown below.

Further roadmap:

* Merge this PR;
* Merge the https://github.com/microsoft/vcpkg/pull/45141 to add the Timidity backend support to the vcpkg port;
* Build the "transitional" dependency package for Windows both with Fluidsynth and Timidity backends;
* Allow the willing to compare the behavior of these backends (this can be easily done by simply deleting the `soundfonts` subdirectory from the files directory and adding the `timidity` directory instead, and then vice versa);
* If the test results are satisfactory (I think that's most likely what will happen), then we can remove the Fluidsynth backend from our Windows dependency package.

Please keep in mind that, although this replacement should help dramatically on Windows, we still don't control all platforms in the world, so on those platforms where Fluidsynth is used, MIDI playback can still start with a delay.

Effect from #8379 in the `master` branch build:

https://github.com/user-attachments/assets/5456f20f-6691-49a7-af0f-7418a70bba97

No effect from #8379 in the build based on this PR along with Timidity MIDI backend:

https://github.com/user-attachments/assets/a916086e-e094-4180-8def-e01bcaa600c6

Save file from #8795 loaded to the `master` branch build:

https://github.com/user-attachments/assets/98e4cbe4-66b6-497a-875a-3802e7ac111e

Save file from #8795 loaded to the build based on this PR along with Timidity MIDI backend:

https://github.com/user-attachments/assets/403a5766-425b-4456-8842-c7d5cfe69ced

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -104,7 +104,7 @@ tasks.register('copyTimidity', Copy) {
         include 'instruments/*.pat'
         include 'timidity.cfg'
     }
-    into 'src/main/assets'
+    into 'src/main/assets/files/timidity'
 }
 
 tasks.register('copyTranslations', Copy) {
--- a/src/dist/Makefile.emscripten
+++ b/src/dist/Makefile.emscripten
@@ -32,7 +32,7 @@ CCFLAGS := $(CCFLAGS) \
 LDFLAGS := $(LDFLAGS) \
 	--preload-file ../../../files/data/resurrection.h2d@/files/data/resurrection.h2d \
 	--preload-file ../../../files/lang/@/files/lang/ \
-	--preload-file ../../../files/timidity/@/etc/ \
+	--preload-file ../../../files/timidity/@/files/timidity/ \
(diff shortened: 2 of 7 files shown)


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for these changes! Hopefully, we can fully fix MIDI playback on Windows.

PR #9758, opened 2025-04-27
Code polish: standardize player Color

Districh-ru:
Currently in the code the Kingdom (castle, hero, army) color uses different format: `int`, `uint8_t` and sometimes `uint32_t` in temporary containers. It results in usage of `static_cast`s and checks for color type and value limit.
Also the same color method is used for player and barrier colors that are absolutely different.

This PR makes an attempt to standardize the color type in the code and introduces the `enum class` that helped to find all places of `Color` usage to update its type to the new one.

Alongside some code was also slightly polished.


src/fheroes2/kingdom/color.h
```diff
@@ -42,7 +44,7 @@ namespace Color
 {
     // !!! IMPORTANT !!!
     // Do NOT change the order of the items as they are used for the map format.
-    enum : uint8_t
+    enum class PlayerColor : uint8_t
```
oleg-derevenetz, line 47: What do you think about moving this `enum` out of namespace `Color`? `enum class` forms a namespace of itself, and `Color::PlayerColor` looks a bit long :) But of course, this is more of a philosophical question.

src/fheroes2/kingdom/color.h
```diff
@@ -55,42 +57,91 @@ namespace Color
         ALL = BLUE | GREEN | RED | YELLOW | ORANGE | PURPLE
     };
 
-    std::string String( const int color );
+    constexpr PlayerColor operator|( const PlayerColor lhs, const PlayerColor rhs )
+    {
+        return static_cast<PlayerColor>( static_cast<uint8_t>( lhs ) | static_cast<uint8_t>( rhs ) );
```
oleg-derevenetz, line 62: I would recommend using underlying type in cases like this, sort of:

```cpp
using PlayerColorUnderlyingType = std::underlying_type_t<PlayerColor>;
return static_cast<PlayerColor>( static_cast<PlayerColorUnderlyingType>( lhs ) | static_cast<PlayerColorUnderlyingType>( rhs ) );
```

The idea is that if you need to change the enum underlying type in the future, it will be enough to change it in just one place (not counting the forward declarations, of course).

oleg-derevenetz (commented):
Hi @Districh-ru I just left a couple of comments, could you please take a look when it's convenient for you?

src/fheroes2/kingdom/color.h
Districh-ru: Hi @oleg-derevenetz, sure let's do this.
`PlayerColor` already has "Color" in it's name. :)

src/fheroes2/ai/ai_common.cpp
```diff
@@ -376,21 +376,21 @@ void AI::shareObjectVisitInfoWithAllies( const Kingdom & kingdom, const int32_t
         return;
     }
 
-    const int friendColors = Players::GetPlayerFriends( kingdom.GetColor() );
-    if ( friendColors == 0 ) {
+    const PlayerColor friendColors = Players::GetPlayerFriends( kingdom.GetColor() );
```
ihhub, line 379: Does it make sense to return `PlayerColor` which is created to store **1** value while `friendColors` can store many colors?

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -81,15 +81,15 @@
 
 namespace
 {
-    uint32_t AIGetAllianceColors()
+    PlayerColor AIGetAllianceColors()
```
ihhub, line 84: Does it make sense to return `PlayerColor` which is created to store **1** value while this function can return many colors?

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -105,13 +105,13 @@ namespace
     }
 
     // Never cache the value of this function as it depends on hero's path and location.
-    bool AIIsShowAnimationForHero( const Heroes & hero, const uint32_t colors )
+    bool AIIsShowAnimationForHero( const Heroes & hero, const PlayerColor colors )
```
ihhub, line 108: The same comment as above: `colors` can contain multiple colors so we shouldn't even use an enumeration for this matter.

src/fheroes2/kingdom/color.h
```diff
@@ -38,59 +41,110 @@ namespace fheroes2
     const char * getTentColorName( const int color );
 }
 
-namespace Color
+// !!! IMPORTANT !!!
+// Do NOT change the order of the items as they are used for the map format.
+enum class PlayerColor : uint8_t
```
ihhub, line 46: Previously this enumeration was just a collection of bit values. Now it has a strict class type which means that this enumeration should be used describing something unique. However, as I pointed out in several comments we are using this enumeration to store multiple entries which might not even be present here. This will lead to confusion of developers who look in the code and also a debugger will not be able to interpret for example value 0x03 because it is not present in the list.

Therefore, I suggest to keep changes related to a standartization of unit8_t usage in the code but for multi-color values we should end up either using `uint32_t` or `int`.

What do you think?

ihhub (changes requested):
Hi @Districh-ru , I left just few comments here regarding the approach for code changes. Could you please take a look at them and see if they make sense?

src/fheroes2/kingdom/color.h
Districh-ru: Hi @ihhub, sure, but I'll make it `uint8_t` to be consistent with map and save file formats and `fogColors` - they use `uint8_t` to store colors.
I'll use `using PlayerColors = std::underlying_type_t<PlayerColor>;`  instead of simple `uint8_t`.
So the `PlayerColor` will be for the only color and `PlayerColors` for a combination of colors.

ihhub (approved):

src/fheroes2/kingdom/color.h
```diff
@@ -38,59 +41,126 @@ namespace fheroes2
...
+    GREEN = 0x02,
+    RED = 0x04,
+    YELLOW = 0x08,
+    ORANGE = 0x10,
+    PURPLE = 0x20,
+    UNUSED = 0x80,
+};
 
-    std::string String( const int color );
+using PlayerColors = std::underlying_type_t<PlayerColor>;
```
oleg-derevenetz, line 58: Hi @Districh-ru I also have the following (optional) suggestion: rename this type alias to the `PlayerColorsSet` or maybe you have better ideas. Rationale: `PlayerColor` and `PlayerColors` are very similar and can be easily confused with each other (and this mistake will be hard to spot with the eye).

Districh-ru:
> Hi @Districh-ru , could you please resolve merge conflict here so we can proceed with merging?

Hi @ihhub, done.

PR #9759, opened 2025-04-27

oleg-derevenetz:


src/fheroes2/gui/ui_text.h
```diff
@@ -274,8 +274,6 @@ namespace fheroes2
     class TextInput final : public Text
     {
     public:
-        TextInput() = delete;
```
oleg-derevenetz, line 277: If there is at least one user-declared constructor, implicit default constructors [are not generated](https://en.cppreference.com/w/cpp/language/default_constructor#Implicitly-declared_default_constructor) (see also an example below).

src/fheroes2/gui/ui_text.h
```diff
@@ -288,10 +286,6 @@ namespace fheroes2
             _updateCursorAreaInText();
         }
 
-        // TextInput should update text and cursor position in that text at the same time.
-        void set( std::string, const FontType ) = delete;
-        void set( std::string, const FontType, const std::optional<SupportedLanguage> ) = delete;
```
oleg-derevenetz, line 293: If the derived class has a method with the same name, then the corresponding methods in the base class are automatically hidden (and their signature does not matter):

https://godbolt.org/z/f1escEq61

Manually deleting them is pointless. If you need them to be visible in the derived class, you need to make them visible manually:

https://godbolt.org/z/8szsGcK8s

ihhub (approved):

ihhub:
@oleg-derevenetz , thank you so much for this code polishing!

PR #9760, opened 2025-04-30
Change encoding name from CP1258 to WINDOWS-1258 to avoid a msgfmt warning

ihhub:
According to msgfmt CP1258 is not a portable encoding name while WINDOWS-1258 is. Both are equal in terms of encoding so this is more a cosmetic change to make msgfmt happy.


ihhub:
It is a false fix. msgfmt is still very unhappy.

Issue #1434, opened 2020-08-29
Adventure map, Eye of the Magi, revealed area


LeHerosInconnu:
In fheroes2, the part of the adventure map revealed by the Eye of the Magi is not the same as in the original game.

In original game:

![Eye of the Magi revealed area 01](https://user-images.githubusercontent.com/43583994/91644384-cf4b2580-ea3b-11ea-9b19-c0bf17eae286.png)

In fheroes2:

![Eye of the Magi revealed area 02](https://user-images.githubusercontent.com/43583994/91644387-d3774300-ea3b-11ea-9d66-33e2346ad2cc.png)

In original game:

![Eye of the Magi revealed area 03](https://user-images.githubusercontent.com/43583994/91644391-d83bf700-ea3b-11ea-9b70-18cec8a3e2c6.png)

In fheroes2:

![Eye of the Magi revealed area 04](https://user-images.githubusercontent.com/43583994/91644395-dd00ab00-ea3b-11ea-8070-605ced960e0c.png)

ihhub:
Hi @LeHerosInconnu , yes it's not the same and it made intentionally. If you can see in the original game the reveal area is something close to hexagon while in our version is circle. Logically, reveal area must be close to circle as it's based on a radius.

Having said that, I'm making this issue as a question for further discussion.

LeHerosInconnu:
Hello @ihhub,

Yes, this is to be discussed because in the original game, the Eye of the Magi reveals more squares on the adventure map and this is to be taken into consideration. :)
Thanks for the precision.

LeHerosInconnu:
Another save for this.
Here the area revealed with an Observation Tower and an Eye of the Magi.

In original game:

![Eye tower area 001](https://user-images.githubusercontent.com/43583994/142216994-35fd5a06-ff92-48e7-9d71-f9f47020d6af.png)

In fheroes2:

![Eye tower area 002](https://user-images.githubusercontent.com/43583994/142217010-60c82a7f-3440-4259-9afa-77219e42ff7a.png)

Here are the save and the test scenario files:
[After eye tower 01.zip](https://github.com/ihhub/fheroes2/files/7555253/After.eye.tower.01.zip)

Ben80-hero:
If there is task to make revealing more authentic, then this is possible. The original algorithm is shown below. With an opening radius of more than 9 tiles, the opening area becomes a perfect circle.

  int minDelta;
  if(radius < 5)
    minDelta = 2;
  else
[...]

PR #9766, opened 2025-04-30
Add ability to set owner color for capturable objects in Editor

Districh-ru:
This PR implements the ability to set ownership of Mines, Sawmills, Alchemist's Labs and Lighthouses in the Editor.
It is only allowed to set the color for objects already placed on the map.

https://github.com/user-attachments/assets/73778840-73f4-4183-956c-72a8af849417


ihhub (changes requested):
Hi @Districh-ru , I left few comments here. Could you please take a look at them?

src/fheroes2/maps/map_format_info.h
```diff
@@ -330,6 +330,9 @@ namespace Maps::Map_Format
         std::map<uint32_t, AdventureMapEventMetadata> adventureMapEventMetadata;
 
         std::map<uint32_t, SelectionObjectMetadata> selectionObjectMetadata;
+
+        // This metadata contains the owner color in relation to tile ID.
```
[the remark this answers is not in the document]
Districh-ru: Hi @ihhub, initially I did use the `StandardObjectMetadata` but it led to many `static_casts` and checks.
All that we need is to set the owner color for the tile (there cannot be more than one action objects on the tile and more than one owner). So the best way is to store color ID for tile ID. So I decided to introduce `ownershipMetadata` to simplify operations with "capturing" the objects and to understand what data is really kept in the container.

And IMHO we should use less of these "magic" containers that contain different (uncertain) data.
It is a legacy of the original game map format, but we can try to do better. :)
The `StandardObjectMetadata` always needs a function to guess what the data is kept in there, check that the data is really the one we need and to determine how to use this data. Such container is a little hard to debug and maintain: you have a mess of different unknown data types in one container.

And we can go back to `StandardObjectMetadata`, I'm not against using it.

src/fheroes2/dialog/dialog_selectitems.cpp
```diff
@@ -1774,3 +1766,116 @@ int Dialog::selectAdventureMiscellaneousObjectType( const int objectType )
 
     return selectObjectType( objectType, objectInfo.size(), listbox );
 }
+
+uint8_t Dialog::selectPlayerColor( const uint8_t color, const uint8_t availableColors )
+{
+    fheroes2::Display & display = fheroes2::Display::instance();
+
+    const int32_t colorsCount = Color::Count( availableColors );
```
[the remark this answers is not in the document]
Districh-ru: If there are no players the dialog will bring only the Neutral color to select.
It is useless so the warning message is now shown for this case.

src/fheroes2/world/world.cpp
```diff
@@ -261,7 +261,8 @@ void CapturedObjects::ClearFog( const int colors ) const
         case MP2::OBJ_MINE:
         case MP2::OBJ_ALCHEMIST_LAB:
         case MP2::OBJ_SAWMILL:
-            scoutingDistance = 2;
+        case MP2::OBJ_LIGHTHOUSE:
+            scoutingDistance = 3;
```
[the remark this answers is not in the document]
Districh-ru: The scouting area in fheroes2 slightly differs from the original game (#1434) and this issue should be fixed in the other PR.

src/fheroes2/maps/map_format_info.h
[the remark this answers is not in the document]
ihhub: Let me review the code to see what could be the best.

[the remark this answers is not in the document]
Districh-ru: Sure, @ihhub, take your time.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -222,6 +222,13 @@ namespace
         return true;
     }
 
+    // Returns true if object can be owned, excluding Towns and Castles.
+    bool isCapturableObject( const Maps::ObjectGroup objectGroup, const MP2::MapObjectType objectType )
+    {
+        return ( objectGroup == Maps::ObjectGroup::ADVENTURE_MINES && objectType != MP2::OBJ_ABANDONED_MINE )
+               || ( objectGroup == Maps::ObjectGroup::ADVENTURE_MISCELLANEOUS && objectType == MP2::OBJ_LIGHTHOUSE );
```
ihhub, line 229: We can remove `objectGroup == Maps::ObjectGroup::ADVENTURE_MISCELLANEOUS` check as from this group we are checking only one object.

src/fheroes2/maps/map_format_info.h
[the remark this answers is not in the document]
ihhub: After further considerations I agree that putting this information in `standardMetadata` might not be a good idea. However, I worry that in the future we might expand metadata for Mines and other capturable objects. I suggest to create a separate structure for this matter. Right now it can be just a structure with one field. But in the future I foresee that we might end up adding guards for capturable objects (e.g. Mines). With the current approach we would have to write conversion code from one data type to a new structure. What do you think?

src/fheroes2/editor/editor_interface.cpp
ihhub: I also suggest just to have a switch-case and list all object types here. object group is not important here and it would also be clear for other developers what objects are considered capturable.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -286,6 +293,11 @@ namespace
                     continue;
                 }
 
+                // Remove ownership data for capturable objects.
+                if ( isCapturableObject( objectIter->group, Maps::getObjectsByGroup( objectIter->group )[objectIter->index].objectType ) ) {
```
ihhub, line 297: Since we have the similar code below and we want to avoid extra operations we can do the following:
```cpp
const auto & objects = Maps::getObjectsByGroup( objectIter->group );
assert( objectIter->index < objects.size() );
const auto objectType = objects[objectIter->index].objectType;

if ( isCapturableObject( objectType ) ) {
    mapFormat.ownershipMetadata.erase( static_cast<int32_t>( mapTileIndex ) );
}
```

and then in the code below remove the same code.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1574,6 +1586,37 @@ namespace Interface
 
                     fheroes2::showStandardTextMessage( MP2::StringObject( objectType ), std::move( str ), Dialog::OK );
                 }
+                else if ( isCapturableObject( object.group, objectType ) ) {
+                    if ( Color::Count( _mapFormat.availablePlayerColors ) == 0 ) {
+                        _warningMessage.reset( _( "There are no players on the map, so no one can own this object." ) );
```
ihhub, line 1591: Hi @zenseii , can you please review this sentence?

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1574,6 +1586,37 @@ namespace Interface
...
+                        return;
+                    }
+
+                    auto ownershipMetadata = _mapFormat.ownershipMetadata.find( tileIndex );
+                    const bool hasOwnershipMetadata = ownershipMetadata != _mapFormat.ownershipMetadata.end();
```
ihhub, line 1596: It would be great to add parenthesis for the code on the right.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -127,6 +127,14 @@ namespace
         }
     }
 
+    void updateWorldOwnership( const Maps::Map_Format::MapFormat & map )
```
ihhub, line 130: Let's call it `updateObjectsOwnership` as nobody owns the world :)

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -614,31 +622,28 @@ namespace Maps
...
+        updateWorldOwnership( map );
+
         return true;
     }
 
     bool readAllTiles( const Map_Format::MapFormat & map )
     {
         assert( map.size == world.w() && map.size == world.h() );
 
-        // We must clear all tiles before writing something on them.
```
ihhub, line 624: Could you please explain why we removed this? Tile can also contain some other information like metadata, hero ID and etc.

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -1255,6 +1255,11 @@ bool World::loadResurrectionMap( const std::string & filename )
 
     updateArtifactStats();
 
+    for ( const auto [tileIndex, color] : map.ownershipMetadata ) {
```
ihhub, line 1258: This code and `void updateWorldOwnership( const Maps::Map_Format::MapFormat & map )` do the same thing. Do we need to do it twice?

src/fheroes2/world/world.cpp
```diff
@@ -392,7 +393,7 @@ void World::generateBattleOnlyMap()
     }
 }
 
-void World::generateForEditor( const int32_t size )
+void World::generateForEditor( const int32_t size, const bool initTiles )
```
ihhub, line 396: I strongly suggest to just create a separate function for setting up tiles and having boolean flags inside a function to control the logic usually leads to a harder to read code.

ihhub (changes requested):
Hi @Districh-ru , I added few comments. Can you please take a look at them in your free time?

ihhub:
I am bumping the priority for this pull request as this is a must have feature for the upcoming release.

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Sure, done!

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Done)

src/fheroes2/editor/editor_interface.cpp
Districh-ru: OK

src/fheroes2/world/world_loadmap.cpp
Districh-ru: It is not done twice. `void updateWorldOwnership( const Maps::Map_Format::MapFormat & map )` was used only for loading maps in Editor.
Now the new function `Maps::updateWorldObjectsOwnership()` is called there and also here, when loading map for playing.

src/fheroes2/maps/map_format_info.h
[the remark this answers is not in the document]
Districh-ru: Sure, I made `CapturableObjectsMetadata` struct which we can expand later.

ihhub (approved):

src/fheroes2/maps/map_format_info.h
```diff
@@ -253,6 +253,11 @@ namespace Maps::Map_Format
         std::vector<int32_t> selectedItems;
     };
 
+    struct CapturableObjectsMetadata
```
ihhub, line 256: **CapturableObjectsMetadata** should be **CapturableObjectMetadata** in a singular form.

src/fheroes2/maps/map_format_info.h
```diff
@@ -330,6 +335,9 @@ namespace Maps::Map_Format
         std::map<uint32_t, AdventureMapEventMetadata> adventureMapEventMetadata;
 
         std::map<uint32_t, SelectionObjectMetadata> selectionObjectMetadata;
+
+        // This metadata contains the owner color in relation to tile ID.
```
ihhub, line 339: We shouldn't use tile index for metadata as it would block us in the future when we allow to place heroes over objects.

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -1582,6 +1591,37 @@ namespace Interface
 
                     fheroes2::showStandardTextMessage( MP2::StringObject( objectType ), std::move( str ), Dialog::OK );
                 }
+                else if ( isCapturableObject( objectType ) ) {
+                    if ( Color::Count( _mapFormat.availablePlayerColors ) == 0 ) {
+                        _warningMessage.reset( _( "There are no players on the map, so no one can own this object." ) );
+                        return;
+                    }
+
+                    auto ownershipMetadata = _mapFormat.capturableObjectsMetadata.find( tileIndex );
```
ihhub, line 1600: We should use object UID here. We can get this value from `objectIter`.

src/fheroes2/world/world_loadmap.cpp
```diff
@@ -709,6 +709,8 @@ bool World::loadResurrectionMap( const std::string & filename )
         return false;
     }
 
+    Maps::updateWorldObjectsOwnership( map );
```
ihhub, line 712: I suggest to remove this function and its call and use the loop below as we do for all other objects. Moreover, `isCapturableObject()` could be moved into **map_format_helper.h|cpp** files so we can use it in the loop below.

src/fheroes2/world/world.h
```diff
@@ -202,7 +202,11 @@ class World : protected fheroes2::Size
     // Generate 2x2 map for Battle Only mode.
     void generateBattleOnlyMap();
 
-    void generateForEditor( const int32_t size );
+    // Generates a map without initializing tiles. Use this function only before loading tiles.
+    void generateVoidMapForEditor( const int32_t size );
```
ihhub, line 206: `Void` and `Blank` can be easily interpret as synonyms by many people. I suggest the following names:
- **generateUninitializedMap** // Call this method only when reading a map from a file.
- **generateMapForEditor**

What do you think?

ihhub (changes requested):
Hi @Districh-ru , I left few more comments here. Would you mind to take a look at them?

src/fheroes2/maps/map_format_info.h
Districh-ru: Hi @ihhub, this will not block us from placing heroes over the objects: when hero captures a mine he is placed over the mine object and this is the same case. 
But there will be another issue: how the Details tool will determine what object (hero or a capturable object under him) we want to edit.

The owner can be set only for one tile (see `struct CapturedObjects`). There can not be many owned action objects at one tile. So the only data we need to capture any object on the map is the tile id and the owner color.

I've changed this container to use object ID as a `map` index, but IMHO now it is just overcomplicated.

But I can see one benefit from this change. If in the future we will allow to move objects on the map then the ownership will be "moved" with the object by its UID. :)

src/fheroes2/editor/editor_interface.cpp
Districh-ru: Done

src/fheroes2/maps/map_format_info.h
Districh-ru: OK

src/fheroes2/world/world_loadmap.cpp
Districh-ru: Sure, done.

src/fheroes2/world/world.h
Districh-ru: Thanks for the suggestion!
It was a bit hard for me to come up with a good name.

src/fheroes2/maps/map_format_helper.cpp
```diff
@@ -98,31 +98,49 @@ namespace
         }
     }
 
-    void updateWorldCastlesHeroes( const Maps::Map_Format::MapFormat & map )
+    void updateWorldCastlesHeroesAndCapturableObjects( const Maps::Map_Format::MapFormat & map )
```
ihhub, line 101: How about to called this function little more generic such as `updatePlayerRelatedObjects`?

ihhub (commented):
I only left one optional comment.

src/fheroes2/maps/map_format_helper.cpp
Districh-ru: Sure, let's rename this function.

ihhub (approved):

ihhub:
Hi @Districh-ru , I will merge this pull request together with another upcoming one from me as both of them are bumping up map version. This is to avoid an extra version in between releases.

ihhub:
@Districh-ru , many thanks for this feature!

PR #9770, opened 2025-05-02
Battlefield: fix incorrect pixels near the SKIP button when Hero's options are opened

### LINKED ISSUE #4471: Battle screen, dark pixels appear near skip button while pointing hero's options
### Preliminary checks

### Platform

Windows

### Describe the bug

"Eagle eye" mode - enabled. :D
Open hero's options screen ant then wave cursor over small window:
image
There's a small dark line between "SKIP" button and battle info bar.
Skip issue
Note that the dark mark disappear while we close hero's options.

### Save file

Any possible battle screen.

### Additional info

_No response_

Districh-ru:
fix #4471

The original status bar background image has incorrect pixels in the rightmost column that are automatically covered by the SKIP button when in the battlefield. And when the Hero's Options dialog is opened the SKIP button is not re-rendered so these pixels become visible.
This PR does crop the original image to remove the incorrect rightmost image column.
Probably this column is a leftover from the leftmost parts of two buttons (SKIP and WAIT?).

It also slightly optimizes:
- the `Status` class to use `width` initialized by `_upperBackground.width()` instead on calling `_upperBackground.width()`;
- removes `std::string statusMessage` initialization because it will be always unused and overwritten in the next loop.

This PR:
img

### DIFF
--- a/src/fheroes2/battle/battle_dialogs.cpp
+++ b/src/fheroes2/battle/battle_dialogs.cpp
@@ -1018,7 +1018,7 @@ int Battle::Arena::DialogBattleHero( HeroBase & hero, const bool buttons, Status
     // The Hero Screen is available for a Hero only (not Captain) and only when the corresponding player has a turn.
     Heroes * heroForHeroScreen = ( currentColor == hero.GetColor() ) ? dynamic_cast<Heroes *>( &hero ) : nullptr;
 
-    std::string statusMessage = _( "Hero's Options" );
+    std::string statusMessage;
 
     while ( le.HandleEvents() && !result ) {
         if ( btnCast.isEnabled() ) {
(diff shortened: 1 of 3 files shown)


src/fheroes2/battle/battle_dialogs.cpp
```diff
@@ -1109,7 +1109,7 @@ int Battle::Arena::DialogBattleHero( HeroBase & hero, const bool buttons, Status
         }
 
         if ( statusMessage != status.getMessage() ) {
-            status.setMessage( statusMessage, false );
+            status.setMessage( std::move( statusMessage ), false );
```
ihhub, line 1112: Aren't we going to break the logic in line 1111 by this? After move operation there is no guarantee that **statusMessage** would still have the same value which leads to a situation that in the next loop iteration `statusMessage != status.getMessage()` is going to be **true**.

ihhub (changes requested):
Hi @Districh-ru , I left one comment here. Can you please take a look at it?

src/fheroes2/battle/battle_dialogs.cpp
Districh-ru: Hi @ihhub, no since we always set the `statusMessage` text in lines 1037-1060 at the start of this loop and we never use this string after this line.
So if we did not do `std::move()` the `statusMessage` will be always overwritten (mostly with the same message).

ihhub:

PR #9780, opened 2025-05-03
Windows CMake workflow: use $VCPKG_INSTALLATION_ROOT instead of the hardcoded path

oleg-derevenetz:
It's better to be safe than sorry.


ihhub (approved):

ihhub:
@oleg-derevenetz , thank you very much for the improvement!

Issue #9782, opened 2025-05-04
MIDI music playback has significantly degraded in 1.1.8


none2003:
### Preliminary checks

- [x] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [x] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

Hi,

After updating to the latest version 1.1.8, I noticed that the sound quality of MIDI music playback has significantly degraded. The changelog mentions "Fix multiple MIDI-related issues on Windows," and the updated ZIP package of 1.1.8 has removed many MIDI-related DLL files. It seems that before version 1.1.8, the game used its own DLLs to play MIDI, while the new version 1.1.8 relies on Windows' MIDI output devices. The default "Microsoft GS Wavetable Synth" in Windows is a poor-quality MIDI device, which has led to the current situation.

Because my system usually has a third-party software wavetable, CoolSoft VirtualMIDISynth, installed, and I use CoolSoft MIDIMapper to change the default MIDI device in Windows from the Microsoft GS Wavetable Synth to VirtualMIDISynth #1, which is the CoolSoft VirtualMIDISynth software wavetable. When I directly play MIDI files using other playback software, the good sound quality of the software wavetable can be demonstrated. However, after entering fheroes2, I found that the sound quality is still the same as when using the Microsoft GS Wavetable Synth device, so I would like to ask if in the fheroes2 1.1.8 update, has the MIDI playback device been hardcoded to use the Microsoft GS Wavetable Synth?

If possible, please set the MIDI playback device of fheroes2 to use the default Windows MIDI device, so that users can improve the sound quality using third-party software synthesizers.

### Save file

No need save file.

oleg-derevenetz:
Hi @none2003, before 1.1.8 the game used fluidsynth backend to play MIDI on Windows. In 1.1.8 fluidsynth backend was replaced by TiMiDity backend (the same backend that is already used on Android, WebAssembly and often on Linux on versions prior to 1.1.8). This backend is much faster, but it uses different set of samples in comparison with fluidsynth.

[...]

PR #9785, opened 2025-05-04
Android: optimize the asset extraction process

oleg-derevenetz:
~~Related to #9782~~

~~This soundfont (in the GUS patches format that TiMidity understands) contains much higher quality samples, BUT it takes up more space accordingly (41Mb instead of 6Mb in the `master` branch). At the same time, it contains the same minimum required set of musical instruments as `master` branch (except drums), It's just that their samples are of higher quality.~~

~~Are we willing to sacrifice some data volume for the sake of MIDI sound quality?~~

Also this PR optimizes the asset extraction process on Android. Since the size of assets in this PR has increased significantly, it is no longer wise to extract them at each game launch.

P.S. OK, judging by the discussion in #9782, these samples don't look any better than the ones that are already included in fheroes2, except that they are bigger. I don't want to get involved in this discussion about musical tastes and will leave only the part that optimizes the process of unpacking assets in Android 😄


Districh-ru:
Providing the base set of small but good enough PAT files, like we do, is the best solution IMHO.

As a compromise we can leave an instruction (or a link to instruction) in the installation guide how to configure TiMidity and replace the PAT files (if needed) and also tell that the user can delete the `timidity` folder and use the system synthesizer on Windows system.

I always do the last (previously I was deleting the fheroes2.sf3) to use CoolSoft VirtualMIDISynth with my favorite soundfont file. :)

PS.
Currently we don't put the `timidity` folder to the PS Vita build and I have no MIDI music on the Vita3K emulator. But if I place this folder into `fheroes2.vpk` archive and install it I get the MIDI music playing on the emulator.
But there are no complaints about MIDI music on PS Vita...

oleg-derevenetz:
> As a compromise we can leave an instruction (or a link to instruction) in the installation guide how to configure TiMidity and replace the PAT files (if needed) and also tell that the user can delete the timidity folder and use the system synthesizer on Windows system.

I don't really know well enough how to configure TiMidity :) I'm just aware in general that there is a config file and it contains some sort of mapping between bank/instrument number and its samples file. But there are also parameters like `keep`, `amp`, `note` or `strip` and I have no idea what they mean.

> Currently we don't put the `timidity` folder to the PS Vita build and I have no MIDI music on the Vita3K emulator. But if I place this folder into `fheroes2.vpk` archive and install it I get the MIDI music playing on the emulator. But there are no complaints about MIDI music on PS Vita...

Nintendo Switch also doesn't get this folder and there are no bugs opened in this regard as well :)

Districh-ru:
> I don't really know well enough how to configure TiMidity :) I'm just aware in general that there is a config file and it contains some sort of mapping between bank/instrument number and its samples file. But there are also parameters like `keep`, `amp`, `note` or `strip` and I have no idea what they mean.

So as I :(
But we may leave a link, in example, to [description in Ubuntu manual](https://manpages.ubuntu.com/manpages/focal/man5/timidity.cfg.5.html) ... it can be discussed.

> Native Windows MIDI driver doesn't behave very well in terms of volume control. Also, as far as I understand, it can't resume MIDI tracks (or can it?)

Yup, it cannot resume MIDI tracks, but volume control works fine with VirtualMIDISynth.

ihhub (approved):

ihhub:
@oleg-derevenetz , huge thanks for this optimization!

PR #9787, opened 2025-05-06
Allow to use Up and Down hotkeys to change the value in value input dialogs

Districh-ru:
This PR adds `DEFAULT_UP` and `DEFAULT_DOWN` hotkeys processing to `ValueSelectionDialogElement` class that is used for value input fields in dialogs.


ihhub (approved):

ihhub:

PR #9793, opened 2025-05-07
Add ability to sort save files by name or date

usidedown:
Solves #1211

Introduces a new configuration `save file sorting` that changes the way sorting works in the load dialog

✅ Allows sorting by timestamp rather than filename
✅ Doesn't change the default behavior
~~❌No UI improvements~~
✅ Awesome UI thanks to @zenseii

### DIFF
--- a/src/engine/tools.h
+++ b/src/engine/tools.h
@@ -121,9 +121,6 @@ namespace fheroes2
             const CharType lc = std::tolower( *li, currentGlobalLocale );
             const CharType rc = std::tolower( *ri, currentGlobalLocale );
 
-            ++li;
-            ++ri;
-
             if ( lc < rc ) {
                 return true;
             }
@@ -131,6 +128,8 @@ namespace fheroes2
                 return false;
             }
             // The characters are "equal", so proceed to checking the next pair
+            ++li;
+            ++ri;
         }
 
         // We have reached the end of one (or both) of the strings, the first parameter is considered "less than" the second if it is shorter
--- a/src/engine/tools.h
+++ b/src/engine/tools.h
@@ -30,6 +30,7 @@
 #include <functional>
 #include <iomanip>
 #include <limits>
+#include <locale>
 #include <optional>
 #include <sstream>
 #include <string>
@@ -105,6 +106,36 @@ namespace fheroes2
 
     // Appends the given modifier to the end of the given string (e.g. "Coliseum +2")
     void appendModifierToString( std::string & str, const int mod );
+
+    // Performs case-insensitive string comparison, suitable for string sorting purposes. Returns true if the first parameter is
+    // "less than" the second, otherwise returns false.
+    template <typename CharType>
+    bool compareStringsCaseInsensitively( const std::basic_string<CharType> & lhs, const std::basic_string<CharType> & rhs )
+    {
+        typename std::basic_string<CharType>::const_iterator li = lhs.begin();
+        typename std::basic_string<CharType>::const_iterator ri = rhs.begin();
+
+        const std::locale currentGlobalLocale;
+
+        while ( li != lhs.end() && ri != rhs.end() ) {
+            const CharType lc = std::tolower( *li, currentGlobalLocale );
+            const CharType rc = std::tolower( *ri, currentGlobalLocale );
+
+            ++li;
+            ++ri;
+
+            if ( lc < rc ) {
+                return true;
+            }
+            if ( lc > rc ) {
+                return false;
+            }
+            // The characters are "equal", so proceed to checking the next pair
+        }
+
+        // We have reached the end of one (or both) of the strings, the first parameter is considered "less than" the second if it is shorter
+        return li == lhs.end() && ri != rhs.end();
+    }
 
     // Performs a checked conversion of an integer value of type From to an integer type To. Returns an empty std::optional<To> if
     // the source value does not fit into the target type.
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -486,7 +486,7 @@ namespace
                 const int currentId = listbox.getCurrentId();
 
                 settings.changeSaveFileSortingMethod();
-                (void)settings.Save( Settings::configFileName );
+                settings.Save( Settings::configFileName );
                 sortMapInfos( lists );
                 listUpdated = true;
 
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -31,7 +31,6 @@
 #include <memory>
 #include <sstream>
 #include <string>
-#include <tuple>
 #include <utility>
 #include <vector>
 
@@ -282,6 +281,32 @@ namespace
         return mapInfos;
     }
 
+    MapsFileInfoList::const_iterator findInMapInfos( const MapsFileInfoList::const_iterator & begin, const MapsFileInfoList::const_iterator & end,
+                                                     const std::string & lastChoice, const uint32_t lastChoiceTimestamp, const SaveFileSortingMethod sortingMethod )
+    {
+        if ( sortingMethod == SaveFileSortingMethod::FILENAME ) {
+            return std::lower_bound( begin, end, lastChoice, CompareByFilename() );
+        }
+
+        const auto [beginSameTimestamp, endSameTimestamp] = std::equal_range( begin, end, lastChoiceTimestamp, CompareByTimestamp() );
+        const MapsFileInfoList::const_iterator it
+            = std::find_if( beginSameTimestamp, endSameTimestamp, [&lastChoice]( const Maps::FileInfo & info ) { return info.filename == lastChoice; } );
+        if ( it != endSameTimestamp ) {
+            return it;
+        }
+        return end; // Not found.
+    }
+
+    MapsFileInfoList::const_iterator findInMapInfos( const MapsFileInfoList::const_iterator & begin, const MapsFileInfoList::const_iterator & end,
+                                                     const std::string & lastChoice, const SaveFileSortingMethod sortingMethod )
+    {
+        if ( sortingMethod == SaveFileSortingMethod::FILENAME ) {
+            return std::lower_bound( begin, end, lastChoice, CompareByFilename() );
+        }
+
+        return std::find_if( begin, end, [&lastChoice]( const Maps::FileInfo & info ) { return info.filename == lastChoice; } );
+    }
+
     std::string selectFileListSimple( const std::string & header, const std::string & lastfile, const bool isEditing )
     {
         // setup cursor
@@ -362,13 +387,7 @@ namespace
         if ( !lastfile.empty() ) {
             filename = System::GetStem( lastfile );
             charInsertPos = filename.size();
-            MapsFileInfoList::const_iterator it;
-            if ( settings.GetSaveFileSortingMethod() == SaveFileSortingMethod::FILENAME ) {
-                it = std::lower_bound( lists.cbegin(), lists.cend(), lastfile, CompareByFilename() );
-            }
-            else {
-                it = std::find_if( it, lists.cend(), [&lastfile]( const Maps::FileInfo & info ) { return info.filename == lastfile; } );
-            }
+            const MapsFileInfoList::const_iterator it = findInMapInfos( lists.cbegin(), lists.cend(), lastfile, fileSortingMethod );
 
             if ( it != lists.cend() ) {
                 listbox.SetCurrent( std::distance( lists.cbegin(), it ) );
@@ -519,19 +538,10 @@ namespace
 
                 // re-select the last selected file if any, unless we're typing in the list box
                 if ( !lastChoice.empty() ) {
-                    MapsFileInfoList::const_iterator it = lists.cbegin();
-                    MapsFileInfoList::const_iterator end = lists.cend();
-                    const SaveFileSortingMethod sortingMethod = settings.GetSaveFileSortingMethod();
+                    const MapsFileInfoList::const_iterator it
+                        = findInMapInfos( lists.cbegin(), lists.cend(), lastChoice, lastChoiceTimestamp, settings.GetSaveFileSortingMethod() );
 
-                    if ( sortingMethod == SaveFileSortingMethod::TIMESTAMP ) {
-                        std::tie( it, end ) = std::equal_range( lists.cbegin(), lists.cend(), lastChoiceTimestamp, CompareByTimestamp() );
-                        it = std::find_if( it, end, [&lastChoice]( const Maps::FileInfo & info ) { return info.filename == lastChoice; } );
-                    }
-                    else {
-                        it = std::lower_bound( lists.cbegin(), lists.cend(), lastChoice, CompareByFilename() );
-                    }
-
-                    if ( it != end ) {
+                    if ( it != lists.cend() ) {
                         const int newId = static_cast<int>( std::distance( lists.cbegin(), it ) );
                         if ( newId != currentId ) {
                             listbox.SetCurrent( newId );
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -235,7 +235,7 @@ namespace
             std::sort( mapInfos.begin(), mapInfos.end(), Maps::FileInfo::sortByFileName );
         }
         else {
-            assert( sortType == SaveFileSortType::LATEST );
+            assert( sortType == SaveFileSortType::TIMESTAMP );
             std::sort( mapInfos.begin(), mapInfos.end(), Maps::FileInfo::sortByTimestamp );
         }
     }
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -532,7 +532,7 @@ namespace
                     lastChoiceTimestamp = lists[currentId].timestamp;
                 }
 
-                settings.changeSaveFileSortingMethod();
+                settings.toggleSaveFileSortingMethod();
                 sortMapInfos( lists );
                 listUpdated = true;
 
--- a/src/fheroes2/dialog/dialog_selectfile.cpp
+++ b/src/fheroes2/dialog/dialog_selectfile.cpp
@@ -236,7 +236,7 @@ namespace
         }
         else {
             assert( sortType == SaveFileSortType::LATEST );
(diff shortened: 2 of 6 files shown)


usidedown:
@ihhub UI part is done, let me know how it looks.

I'm slightly worried about the performance of clicking the sort button. Not sure if it's worth improving though.

There are 2 "heavy" operations:
- the sorting itself should be NLOGN comparisons, but each comparison is not very cheap when comparing filenames. Fixing this is probably a bigger change and probably deserves its own PR.
- I also need to find the last selected item and re-selected in the new list. This is again N comparisons of file names. Maybe it's possible to compare pointers instead of strings to speed things up. Not sure it's worth doing.

What do you think?

usidedown:
> @usidedown , could you please also resolve merge conflicts to proceed with the review?

@ihhub Is the proposed UI design acceptable? I got the impression from discord that we want something different.

ihhub (changes requested):
Hi @usidedown , I put several comments in this pull request. Could you please take a look at them in your free time?

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -461,6 +482,33 @@ namespace
                     result = System::concatPath( Game::GetSaveDir(), filename + Game::GetSaveFileExtension() );
                 }
             }
+            else if ( le.MouseClickLeft( buttonSort.area() ) ) {
+                const int currentId = listbox.getCurrentId();
+
+                settings.CycleSaveFileSortType();
+                (void)settings.Save( Settings::configFileName );
+                sortMapInfos( lists );
```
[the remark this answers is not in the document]
usidedown: I think this one is going to be a problem.
We currently use a vector of _Maps::FileInfo_. So each sort we actually move these objects around and the address of the object changes.
The only unique field of FileInfo is _filename_. We technically can search for std::addressof(FileInfo.filename.c_str()). It'll work on some platforms but fail on other platforms (platforms with short string optimization and short filenames).

We can avoid all that by using a vector of pointers (or std::reference_wrapper) instead of vector of objects. But that would probably a large number of changes to the types ListBox uses and I don't think it's worth the hassle.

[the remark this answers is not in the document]
ihhub: You are right. I will think whether it is possible to simplify the code while keeping the best performance.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -521,8 +569,11 @@ namespace
             else if ( isEditing && le.isMouseRightButtonPressedInArea( buttonVirtualKB->area() ) ) {
                 fheroes2::showStandardTextMessage( _( "Open Virtual Keyboard" ), _( "Click to open the Virtual Keyboard dialog." ), Dialog::ZERO );
             }
+            else if ( le.isMouseRightButtonPressedInArea( buttonSort.area() ) ) {
+                fheroes2::showStandardTextMessage( _( "Sort" ), _( "Click to toggle sorting by name/date" ), Dialog::ZERO );
+            }
 
-            const bool needRedrawListbox = listbox.IsNeedRedraw();
+            const bool needRedrawListbox = listUpdated || listbox.IsNeedRedraw();
```
ihhub, line 576: I suggest to add a TODO comment here like this:
```cpp
// TODO: ListBox::SetCurrent() call should update needRedraw variable as it changes the internal UI view of the class.
```
We can address it in a separate pull request.

src/fheroes2/system/settings.h
```diff
@@ -71,6 +71,12 @@ enum class InterfaceType : uint8_t
     DYNAMIC = 2
 };
 
+enum class SaveFileSortingMethod : uint8_t
+{
+    FILENAME = 0,
```
ihhub, line 76: Please remove the values. Since we use `enum class` they are not needed.

ihhub (changes requested):
Hi @usidedown , I added few more comments here. Would you mind to take a look at them?

src/fheroes2/ai/ai_hero_action.cpp
```diff
@@ -1917,7 +1917,6 @@ void AI::HeroesAction( Heroes & hero, const int32_t dst_index )
     const bool isActionObject = MP2::isInGameActionObject( objectType, hero.isShipMaster() );
     if ( isActionObject ) {
         hero.SetModes( Heroes::ACTION );
-    }
```
usidedown, line 1920: @ihhub Can you take a look at this change?
It's unrelated to the pull request, but Github Actions were failing due to a bug here.
I think something is broken in master branch here.

ihhub: Thanks for spotting! @Districh-ru would need to take a look at this if it is the correct solution.

src/fheroes2/dialog/dialog_selectfile.cpp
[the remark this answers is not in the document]
usidedown: I modified the code to be similar to this, but with the following enhancement:
I search by timestamp when the list is sorted by it, and by name when appropriate.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -115,6 +116,26 @@ namespace
...
+        {
+            return info.timestamp > timestamp;
+        }
+        bool operator()( uint32_t timestamp, const Maps::FileInfo & info ) const
+        {
+            return timestamp > info.timestamp;
+        }
+    };
+
+    struct CompareByFilename
```
ihhub, line 131: Using a structure for iterator comparison is slower than having a lambda function for this matter. I recommend to declare a lambda function within the function where we use it. Also, it would reduce the function scope.

src/fheroes2/dialog/dialog_selectfile.cpp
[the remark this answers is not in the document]
ihhub: Hi @usidedown , now the code has become much more difficult to read. While the goal was the fastest implementation we sacrificed the readability a lot. I suggest to keep the code shorter and cleaner for now and the amount of changes in this pull request would be significantly smaller.

ihhub (changes requested):
Hi @usidedown , I added 2 more comments here. Would you mind to take a look at them?

src/fheroes2/dialog/dialog_selectfile.cpp
usidedown: Lambdas are syntactic sugar for structs, I'll be very surprised if you find performance differences. Don't you think it'll clutter the code more than necessary?

ihhub:
Hi @usidedown , are you able to complete the changes in 1-2 days? I would like to know whether we can include this change in the upcoming release.

usidedown:
> Hi @usidedown , are you able to complete the changes in 1-2 days? I would like to know whether we can include this change in the upcoming release.

I extracted the new logic to new functions. I believe it's considerably more readable then the last commit, and very similar to the code before the change.

usidedown:
@zenseii your design definitely looks better!
Is it a big change (code-wise) w.r.t the current PR?
I barely managed to put a button, not sure if I can handle this design 😅

zenseii:
> @zenseii your design definitely looks better! Is it a big change (code-wise) w.r.t the current PR? I barely managed to put a button, not sure if I can handle this design 😅

Thanks! I can push the needed changes as I did some of them already for the screenshot (it was less work than photoshopping it). I'll expand the height of the header cells too.

src/fheroes2/dialog/dialog_selectfile.cpp
oleg-derevenetz: Hi @ihhub while structures with `operator()` (functors) [have some minor disadvantages](https://godbolt.org/z/os9hxMhfW) compared to lambdas, I don't really think that functors are somehow noticeably slower. The advantage of functors is that they allow you to concentrate the similar logic in one place, while lambdas will be scattered throughout the code. Another working option could be to use overloaded functions, but this [will not work with templates](https://godbolt.org/z/fr1TjTrcP).

oleg-derevenetz (commented):
Hi @usidedown I just made a few changes:

* Renamed the `Settings::changeSaveFileSortingMethod()` to the `Settings::toggleSaveFileSortingMethod()` - it seems to me that this variant reflects its essence better.
* Moved the `CaseInsensitiveCompare()` to the `tools.h` (as `fheroes2::compareStringsCaseInsensitively()`) since this function is "relatively generic" and can be used in multiple places... But for now it is used just in `maps_fileinfo.cpp`.
* Moved all comparators to the single place so that there will be no confusion if there is a need to change something.
* Fixed the issue like this:

  https://github.com/user-attachments/assets/0aca4f5e-2cb1-4a99-93d5-653ed1dcad1d

  `std::lower_bound()` doesn't mean "equality" per se.
* Fixed the similar issue on the case-sensitive filesystems. Given that the current sorting is case-insensitive, there can be multiple "identical" file names with different case.
* Added a few assertions (for the debug builds only) to check that lists are properly sorted/partitioned before calling `std::lower_bound()` or `std::equal_range()` on them. Attempting to call these functions on an incorrectly sorted list causes undefined behavior.

Could you please take a look at these changes to make sure I didn't miss anything?

src/fheroes2/maps/maps_fileinfo.h
```diff
@@ -179,6 +175,39 @@ namespace Maps
             LOSS_OUT_OF_TIME = 3
         };
 
+        // This comparator performs the case-insensitive comparison
+        struct CompareByFileName
```
usidedown, line 179: I liked it better when the scope of the compare classes was smaller.
Maybe move this to its own .h file and include only from maps_fileinfo.cpp

This is a minor pet peeve, feel free to ignore!

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -244,11 +256,69 @@ namespace
...
+            return iter;
+        }
+
+        return end; // Not found.
+    }
+
+    MapsFileInfoList::const_iterator findInMapInfos( const MapsFileInfoList::const_iterator & begin, const MapsFileInfoList::const_iterator & end,
+                                                     const std::string & lastChoice, const SaveFileSortingMethod sortingMethod )
+    {
+        if ( sortingMethod == SaveFileSortingMethod::FILENAME ) {
```
usidedown: Consider handling the simpler case first

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -244,11 +256,69 @@ namespace
...
+    MapsFileInfoList::const_iterator findInMapInfos( const MapsFileInfoList::const_iterator & begin, const MapsFileInfoList::const_iterator & end,
+                                                     const std::string & lastChoice, const uint32_t lastChoiceTimestamp, const SaveFileSortingMethod sortingMethod )
+    {
+        if ( sortingMethod == SaveFileSortingMethod::FILENAME ) {
+#ifdef WITH_DEBUG
+            assert( std::is_sorted( begin, end, Maps::FileInfo::CompareByFileName{} ) );
+#endif
+
+            // With case-insensitive sorting, there may be several "identical" file names on the case-sensitive file system
+            const auto [beginSameFileName, endSameFileName] = std::equal_range( begin, end, lastChoice, Maps::FileInfo::CompareByFileName{} );
```
usidedown, line 280: The next few lines repeat 2 (actually 3) times.
Maybe it's worth an extra function. I trust your judgement.

usidedown (commented):
@oleg-derevenetz thank you so much for these fixes!

Code looks correct overall, but I have minor suggestions.

src/fheroes2/maps/maps_fileinfo.h
oleg-derevenetz: Hi @usidedown

> Maybe move this to its own .h file and include only from maps_fileinfo.cpp

These comparators are currently used in two files: in the `maps_fileinfo.cpp` and in the `dialog_selectfile.cpp`. Before, there was a separate comparator in each of these files (`CompareByXXX` in the `dialog_selectfile.cpp` and `sortByXXX()` in `maps_fileinfo.h/cpp`), which had the same logic, but were located in different places. Now the logic of all these comparators is located in just one place. Of course, we can put them in a separate module, but I'm not sure that it's worth making a separate module just for the sake of comparators. We need to include the `maps_fileinfo.h` in the `dialog_selectfile.cpp` anyway because `dialog_selectfile.cpp` needs to know the internals of `Maps::FileInfo` to work, so I'm not sure what we would win here by using a separate module for comparators.

src/fheroes2/dialog/dialog_selectfile.cpp
oleg-derevenetz: After some thought, I decided to replace these `if`'s with a `switch`. With `switch`, it's easier to keep track that all value variants will be processed.

src/fheroes2/dialog/dialog_selectfile.cpp
oleg-derevenetz: Actually 4 times :) Yes, that's a good idea.

zenseii:
Hi, @usidedown. I've added the header cells and radio buttons. Things are a bit more crammed in the dialog now. Ideally the way the size of the dialog is setup according to resolution and number of save files should be reworked, but for now I think this is good enough.

src/engine/tools.h
```diff
@@ -106,6 +107,36 @@ namespace fheroes2
...
+        typename std::basic_string<CharType>::const_iterator ri = rhs.begin();
+
+        const std::locale currentGlobalLocale;
+
+        while ( li != lhs.end() && ri != rhs.end() ) {
+            const CharType lc = std::tolower( *li, currentGlobalLocale );
+            const CharType rc = std::tolower( *ri, currentGlobalLocale );
+
+            ++li;
+            ++ri;
```
Districh-ru: We can move these two increments to the end of the loop body because we don't need to do increments if any of the next 2 conditions is true.

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -461,6 +608,14 @@ namespace
                     result = System::concatPath( Game::GetSaveDir(), filename + Game::GetSaveFileExtension() );
                 }
             }
+            else if ( le.MouseClickLeft( nameHeaderRoi ) && settings.GetSaveFileSortingMethod() != SaveFileSortingMethod::FILENAME ) {
+                listUpdated = true;
+                needRedraw = switchFileSorting( settings.GetSaveFileSortingMethod() == SaveFileSortingMethod::FILENAME );
```
Districh-ru: We can pass `false` here because in the line 611 `settings.GetSaveFileSortingMethod()` is already checked.
```suggestion
                needRedraw = switchFileSorting( false );
```

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -461,6 +608,14 @@ namespace
...
+            }
+            else if ( le.MouseClickLeft( dateHeaderRoi ) && settings.GetSaveFileSortingMethod() != SaveFileSortingMethod::TIMESTAMP ) {
+                listUpdated = true;
+                needRedraw = switchFileSorting( settings.GetSaveFileSortingMethod() == SaveFileSortingMethod::FILENAME );
```
Districh-ru: The same as previous comment, but `true` should be passed:
```suggestion
                needRedraw = switchFileSorting( true );
```

src/fheroes2/system/settings.h
```diff
@@ -354,6 +360,21 @@ class Settings
         _viewWorldZoomLevel = zoomLevel;
     }
 
+    SaveFileSortingMethod GetSaveFileSortingMethod() const
```
Districh-ru: Please start this method's name from the lower character:
```suggestion
    SaveFileSortingMethod getSaveFileSortingMethod() const
```

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -370,7 +455,68 @@ namespace
...
+        if ( fileSortingMethod == SaveFileSortingMethod::FILENAME ) {
+            fheroes2::Blit( mark, display, { markBackgroundNameRoi.x + 3, markBackgroundNameRoi.y + 3, mark.width(), mark.height() } );
+        }
+        else {
+            fheroes2::Blit( mark, display, { markBackgroundDateRoi.x + 3, markBackgroundDateRoi.y + 3, mark.width(), mark.height() } );
+        }
+
+        // Redraw sort radio buttons, sort file list in new method, and return whether to redraw the list.
+        auto switchFileSorting
+            = [&markBackgroundNameRoi, &markBackgroundDateRoi, &markBackground, &mark, &display, &listbox, &lists, &settings]( const bool isSortedByName ) {
```
Districh-ru: Lets rename the function argument to `doSortByDate`:
```suggestion
            = [&markBackgroundNameRoi, &markBackgroundDateRoi, &markBackground, &mark, &display, &listbox, &lists, &settings]( const bool doSortByDate ) {
```

src/fheroes2/system/settings.h
```diff
@@ -354,6 +360,21 @@ class Settings
...
+
+    void toggleSaveFileSortingMethod()
+    {
+        if ( _saveFileSortType == SaveFileSortingMethod::FILENAME ) {
+            _saveFileSortType = SaveFileSortingMethod::TIMESTAMP;
+        }
+        else {
+            _saveFileSortType = SaveFileSortingMethod::FILENAME;
+        }
+    }
```
Districh-ru: Lets make
```cpp
void setSaveFileSortingMethod( const SaveFileSortingMethod sortType ) {
    _saveFileSortType = sortType;
}
```
to explicitly set the needed sorting type.
It will also improve the flexibility if in the future we introduce new sort types (ascending/descending selection, in example).
What do you think?

src/fheroes2/dialog/dialog_selectfile.cpp
```diff
@@ -370,7 +455,68 @@ namespace
...
+
+                  const int currentId = listbox.getCurrentId();
+                  std::string lastChoice{};
+                  uint32_t lastChoiceTimestamp{};
+                  if ( currentId >= 0 && static_cast<size_t>( currentId ) < lists.size() ) {
+                      lastChoice = lists[currentId].filename;
+                      lastChoiceTimestamp = lists[currentId].timestamp;
+                  }
+
+                  settings.toggleSaveFileSortingMethod();
```
Districh-ru: The here we will do
```cpp
settings.setSaveFileSortingMethod( doSortByDate ? SaveFileSortingMethod::TIMESTAMP: SaveFileSortingMethod::FILENAME );
```

Districh-ru (approved):
Well done, @usidedown.

You can also mark in the PR topic that there are UI improvements. :)

ihhub (approved):

ihhub:

Issue #8796, opened 2024-06-01
Combat screen, giant's sprite is displayed on top of the crusader troop's creature count indicator


LeHerosInconnu:
### Preliminary checks

- [X] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [X] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Windows

### Describe the bug

In fheroes2, the giant's sprite is displayed on top of the crusader troop's creature count indicator.

In fheroes2:

![vlcsnap-2024-03-03-15h52m28s055](https://github.com/ihhub/fheroes2/assets/43583994/ca9ca14d-943f-4dd8-8597-877aea71c6b6)

https://github.com/ihhub/fheroes2/assets/43583994/cb1c0a1b-f47c-4301-9424-827e41229fa0

### Save file

Here is the save file:
[Sprite giants over count crusaders 01.zip](https://github.com/user-attachments/files/15522962/Sprite.giants.over.count.crusaders.01.zip)

Districh-ru:
Fixed by #9794

PR #9794, opened 2025-05-08
Battle Interface: drawing troop count after (moving) monster

usidedown:
solves #8796

Previously troop counts were drawn after the troops themselves.
This seems to cause the moving troops to override the troop count

old

I swapped the order and now the troop count should be in front of the moving monster

new

### DIFF
--- a/src/fheroes2/battle/battle_interface.cpp
+++ b/src/fheroes2/battle/battle_interface.cpp
@@ -1623,7 +1623,7 @@ void Battle::Interface::RedrawArmies()
                 }
                 else {
                     if ( isCellBefore ) {
-                        if ( _movingPos.y < currentCell->GetPos().y ) {
+                        if ( _movingPos.y <= currentCell->GetPos().y ) {
                             // The troop is moving to the upper row. We should render it prior to this row units.
                             upwardMovingTroopBeforeWall.emplace_back( unitOnCell );
                         }
@@ -1632,7 +1632,7 @@ void Battle::Interface::RedrawArmies()
                         }
                     }
                     else {
-                        if ( _movingPos.y < currentCell->GetPos().y ) {
+                        if ( _movingPos.y <= currentCell->GetPos().y ) {
                             // The troop is moving to the upper row. We should render it prior to this row units.
                             upwardMovingTroopAfterWall.emplace_back( unitOnCell );
                         }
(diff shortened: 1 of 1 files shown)


Districh-ru (approved):

ihhub:
Technically, everything is okay but I think we had some cases where troop count should be drawn before.

Hi @LeHerosInconnu, can you please test these changes?

Districh-ru (changes requested):
@ihhub, you are right.
When moving downwards we should draw the counter prior to the moving unit. In this PR the counter overlaps the moving downwards troop:

https://github.com/user-attachments/assets/02795a3a-5d49-497e-97c3-7669905c16c0

It can be done like we did for the `upwardMoovingTroop`.

Hi @usidedown, could you please check this issue when you have time.

usidedown:
Okay, It's now done in the exact same way as upwardMoovingTroop.
Now moving in the same row behaves exactly like upwardMoovingTroop, so it's technicially nonDownwardMovingTroop ;)


https://github.com/user-attachments/assets/046b8065-2c5d-4d65-ab1a-bd2dff338bd2

Districh-ru (changes requested):
Hi @usidedown,
Now the troop counts are draw correctly and I did not notice bugs in moving one troop around the other. Well done.

Could you please update the comments because no we have 2 cases:
- the troop is moving horizontally or upwards;
- the troop is moving downwards.

So we can also rename `upwardMovingTroop` to `MovingTroop` and the current `MovingTroop` to `downwardMovingTroop`. What do you think?

usidedown:
@Districh-ru Let me know how it looks now

Districh-ru (approved):
@usidedown, looks good to me!

LeHerosInconnu:
Hello @ihhub,

> Technically, everything is okay but I think we had some cases where troop count should be drawn before.
> 
> Hi @LeHerosInconnu, can you please test these changes?

I'll look into it when I have some time.
It may take a while.

ihhub (approved):

ihhub:
@usidedown , many thanks for this change!

LeHerosInconnu:
Hello @ihhub

> Hello @ihhub,
> 
> > Technically, everything is okay but I think we had some cases where troop count should be drawn before.
> > Hi @LeHerosInconnu, can you please test these changes?
> 
> I'll look into it when I have some time. It may take a while.

I did a quick test for this and everything looks good.
I'll be sure to keep an eye on this in my next test with a full game.

PR #9795, opened 2025-05-08
Update French language-specific characters on map load

### LINKED ISSUE #9490: Invalid character displayed for the cursor when setting the game language to french

Districh-ru:
fix #9490

This PR also fixes inability to display `lowercase i with circumflex` from the French translated maps:
изображение

https://github.com/user-attachments/assets/296ad9af-d286-4527-b28c-f3a7c2f6d8ac

TODO: update all texts from the save files with French language.

PS
I moved the "decoding" cases from `ui_font` to `tools` but while the comment differs with the character for the replacement (by it's comment).
The help of French-speaking players that have original assets is needed to check the correctness of characters update to the ones from [CP1252](https://www.ascii-code.com/).

### DIFF
--- a/src/fheroes2/maps/maps_fileinfo.cpp
+++ b/src/fheroes2/maps/maps_fileinfo.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *   fheroes2: https://github.com/ihhub/fheroes2                           *
- *   Copyright (C) 2019 - 2024                                             *
+ *   Copyright (C) 2019 - 2025                                             *
  *                                                                         *
  *   Free Heroes2 Engine: http://sourceforge.net/projects/fheroes2         *
  *   Copyright (C) 2009 by Andrey Afletdinov <fheroes2@gmail.com>          *
@@ -52,6 +52,7 @@
 #include "settings.h"
 #include "system.h"
 #include "tools.h"
+#include "ui_language.h"
 
 namespace
 {
@@ -90,6 +91,11 @@ namespace
         // create a list of unique maps (based on the map file name) and filter it by the preferred number of players
         std::map<std::string, Maps::FileInfo, std::less<>> uniqueMaps;
 
+        // For the original French version we update the language-specific characters to match CP1252.
+        const bool updateFrenchLangugeSpecificCharacters
+            = isOriginalMapFormat
+              && ( fheroes2::getCurrentLanguage() == fheroes2::SupportedLanguage::French || fheroes2::getResourceLanguage() == fheroes2::SupportedLanguage::French );
+
         for ( const std::string & mapFile : mapFiles ) {
             Maps::FileInfo fi;
 
@@ -122,6 +128,12 @@ namespace
                 if ( humanOnlyColorsCount == humanPlayerCount ) {
                     // The map has the exact number of human-only players. Make sure that the user cannot select any other players.
                     fi.removeHumanColors( fi.AllowCompHumanColors() );
+                }
+
+                // Update French language-specific characters to match CP1252.
+                if ( updateFrenchLangugeSpecificCharacters ) {
+                    fheroes2::updateFrenchLanguageSpecificCharactersForMaps( fi.name );
+                    fheroes2::updateFrenchLanguageSpecificCharactersForMaps( fi.description );
                 }
             }
 
--- a/src/engine/tools.cpp
+++ b/src/engine/tools.cpp
@@ -152,6 +152,76 @@ namespace fheroes2
         output.replace( output.size() - originalEndingSize, originalEndingSize, correctedEnding, correctedEndingSize );
     }
 
+    void updateFrenchLanguageSpecificCharactersForMaps( std::string & str )
+    {
+        for ( char & c : str ) {
+            switch ( c ) {
+            case 9:
+                // Lowercase i with circumflex
+                c = static_cast<char>( 238 );
+                break;
+            case 35:
+                //  Lowercase o with circumflex
+                c = static_cast<char>( 244 );
+                break;
+            case 36:
+                // Lowercase u with circumflex
+                c = static_cast<char>( 251 );
+                break;
+            case 38:
+                // Lowercase u with grave accent
+                c = static_cast<char>( 249 );
+                break;
+            case 42:
+                // Lowercase a with circumflex
+                c = static_cast<char>( 226 );
+                break;
+            case 60:
+                // Lowercase i with diaeresis
+                c = static_cast<char>( 239 );
+                break;
+            case 62:
+                // Lowercase i with circumflex <- Confirmed used in the OG Succession wars.
+                c = static_cast<char>( 238 );
+                break;
+            case 64:
+                // Lowercase a with grave accent
+                c = static_cast<char>( 224 );
+                break;
+            case 94:
+                // Lowercase c with cedilla
+                c = static_cast<char>( 231 );
+                break;
+            case 96:
+                // Lowercase e with grave accent
+                c = static_cast<char>( 232 );
+                break;
+            case 123:
+                // Lowercase i with diaeresis
+                c = static_cast<char>( 239 );
+                break;
+            case 124:
+                // Lowercase e with circumflex
+                c = static_cast<char>( 234 );
+                break;
+            case 125:
+                // Lowercase i with circumflex
+                c = static_cast<char>( 239 );
+                break;
+            case 126:
+                // Lowercase e with acute
+                c = static_cast<char>( 233 );
+                break;
+            case 127:
+                // Lowercase i with circumflex
+                c = static_cast<char>( 239 );
+                break;
+            default:
+                break;
+            }
+        }
+    }
+
     std::string abbreviateNumber( const int num )
     {
         if ( std::abs( num ) >= 1000000 ) {
--- a/src/engine/tools.h
+++ b/src/engine/tools.h
@@ -100,6 +100,12 @@ namespace fheroes2
 
     void replaceStringEnding( std::string & output, const char * originalEnding, const char * correctedEnding );
 
+    // The original French version replaces several ASCII special characters with language-specific characters.
+    // In the engine we use CP1252 for the French translation.
+    // This function replaces the special ASCII characters with the language-specific characters from CP1252
+    // to properly display texts for maps from French assets.
+    void updateFrenchLanguageSpecificCharactersForMaps( std::string & str );
+
     std::string abbreviateNumber( const int num );
 
     // Appends the given modifier to the end of the given string (e.g. "Coliseum +2")
--- a/src/fheroes2/castle/castle.cpp
+++ b/src/fheroes2/castle/castle.cpp
@@ -74,6 +74,7 @@
 #include "tools.h"
 #include "translations.h"
 #include "ui_dialog.h"
+#include "ui_language.h"
 #include "week.h"
 #include "world.h"
 
@@ -106,7 +107,7 @@ Castle::Castle( const int32_t posX, const int32_t posY, int race )
     // Do nothing.
 }
 
-void Castle::LoadFromMP2( const std::vector<uint8_t> & data )
+void Castle::LoadFromMP2( const std::vector<uint8_t> & data, const bool updateFrenchLanguageSpecificCharactersInName )
 {
     assert( data.size() == MP2::MP2_CASTLE_STRUCTURE_SIZE );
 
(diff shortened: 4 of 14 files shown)


Districh-ru:
Hi @zenseii and @felix642, I would appreciate your help in checking the correctness of the new `updateFrenchLanguageSpecificCharactersForMaps()` function.
This function returns the decimal code for a character from [CP1252](https://www.ascii-code.com/).
Sadly, I don't know French to check this, I can just technically say that the character are shown the same as in the master build.

felix642:
Hello @Districh-ru, from my understanding the method `updateFrenchLanguageSpecificCharactersForMaps()` only affect the map name / description.
I've skimmed through the map list to check if I could see any change in behaviour / weird characters in some descriptions and everything looks good!

Now there's one minor defect that I found, but the same behaviour exists on the master branch so I don't know if we want to address it in this PR or not.

Some maps have some weird characters at the end of the sentence but only when the french localization is chosen.
For example : 
image

The same behaviour can be seen in the dialogs too.
image

Anyway, let me know if this is indeed what you needed me to check or if I missed something.

Districh-ru:
Hello @felix642, many thanks for checking!
This PR relates to the strings loaded from map files: map name, description, events texts, heroes/castles names, Sphinx riddles.
And I asked you to check that French texts (especially non-ASCII characters) are shown correctly because I found inconsistency in descriptions in`generateFrenchAlphabet()` function that was used to replace some ASCII characters with French language-specific CP1252 ones (for maps) and the CP1252 characters that are mapped in this function. The function I made to update all map strings is done using the logic of `generateFrenchAlphabet()` function.
So I cannot definitely say what is wrong - the comments or the CP1252 character's codes in this function. :(

I'll also try to do a comparison of this PR with French GOG version later today.

Districh-ru:
I've checked the French GOG version and fixed recurrence in CP1252 characters mapping to fully correspond this official version.

Districh-ru:
@felix642, sadly in the official French GOG version the character `<` is replaced with `ï` and character `>` - with `î`, while these characters (`<` and `>`) can be used (and used) by map makers. And when starting a map with French assets these ASCII characters will always be replaced.
Thus is a bug of the official assets. And we support the official assets including the maps so we have to reproduce this behavior so the French speaking players can enjoy playing the official campaign and maps in French.
The only way to properly display the maps that were not created for the French GOG version in French is currently:
1. Have the non-French assets (to have all original ASCII symbols).
2. Start the map in with any language except French (e.g. setting English language in fheroes2) to the engine will not remap these characters.
3. After the map has started you can change the language back to French.

Sadly there is no way to detect the language of MP2/MX2 maps to properly deal with them. :(

Districh-ru:
Possibly not all if the "replaced" characters are used in the official maps, some of them might be used only in the game UI, so we can reduce the amount of the "remaps" (including `<` and `>`). But we need to check all 143 maps' strings for presence of these characters...

Districh-ru:
> @Districh-ru, thanks for checking all this! As you can see, the French original encoding is a mess, especially because the SW French maps/campaigns use some values for accented characters, while the PoL French uses other.
> 
> As for detecting what language the mp2/mx2 maps are in, wouldn't it be safe enough to assume that if we have French assets (crc32 checked), that the maps also will be from the original French assets?

@zenseii, yeah it look like an ordinary mess of non-official translations. :) And this mess affects our GUI and other maps.
I've thought that we should check only for the original French assets and do not check for the selected language, but there is no guarantee that there are no other maps made for this uncommon codepage.
But if the maps with this codepage are only provided with the French assets than we can do strings conversion for maps only for the French assets (by crc32) and only for the list of maps present in the official French GOG version not to corrupt the other maps, like it is shown upper: https://github.com/ihhub/fheroes2/pull/9795#issuecomment-2867999902 - that map is not from the official assets and has symbols that are corrupted by the changed fonts.
Hi @ihhub, I would appreciate your opinion on the approach how to use CP1252 without additional ASCII characters hack for French maps and also with the ability to properly display all characters in these maps.

ihhub:
Hi @Districh-ru ,

I took a look at changes. If I understood correctly your approach is to avoid having a dedicated French language assets but try to fix special characters while loading MP2 maps and within save files. If this is correct I can see a problem that it is not possible to detect that a map is made using French assets at the moment.

Districh-ru:
> Hi @Districh-ru ,
> 
> I took a look at changes. If I understood correctly your approach is to avoid having a dedicated French language assets but try to fix special characters while loading MP2 maps and within save files. If this is correct I can see a problem that it is not possible to detect that a map is made using French assets at the moment.

Hi @ihhub, yes, my approach is to make 0-127 ASCII characters the same for all languages. So we'll have no problems with text input cursor and other ASCII characters (*, &, $, # <, >, and others). Any map maker is able to use all ASCII characters and if you load such fh2m map with French language selected you will have about 13 characters different (corrupted) from the other languages.

In future we may get bug reports like #9490
So we should not ever call `generateFrenchAlphabet()` that replaces the ASCII characters with CP1252 ones and also in future we should restore ASCII characters for the French assets.

But we need to support the official French GOG version maps without modifying the map files.
So the only way I see is to update all strings in map files.

Currently my approach gives the same result as the master build: for French assets and when French language is selected all MP2 and MPX files will have the replaced ASCII characters and these characters for non-official maps will be "corrupted", like it is shown earlier for `<` and `>`.

And we may take @zenseii's approach and do this conversion only when the French assets are found because for the other assets there will be no maps that needs the replacement of ASCII characters. After some thoughts I  would like to use it.

Additionally we exactly know the list of maps included in the French GOG version so we may do the characters update only for these maps and only when French assets are found. It should remove the "corruption" of ASCII characters in the other maps.
These extra ASCII characters were placed in maps by using some hex editor and not the original map editor (the 0x09 and 0x7F could not be put by the editor). Tomorrow I'll check the abilities of map editor in French GOG version.

By this PR I want to reduce the replacement of ASCII characters not to interfere with fheroes2 GUI and fh2m maps.

PS
Yes, to fix the cursor issue we may do a patch for the case of French assets, or use the other character for the cursor, or even do not use font character to display text cursor and simply generate a sprite for it.
But we'll have issues with these symbols in all other maps and GUI unfixed:
just launch fheroes2, select French language and try to enter a name for hero: `Richie & com $$$` and you will get 
изображение  instead of wanted изображение

ihhub:
Hi @Districh-ru , assuming we chose this path. I think we also need to solve the problem that all characters are shifted by 32 symbols and this causes us additional maintenance and headache. I don't think that extra 1.2 KB of space required for 32 empty images is worth complicating our code. However, this is not directly related to this pull request.

ihhub:
Hi @Districh-ru  and @zenseii , can we with almost 100% guarantee identify that a map is made by a French version of the HoMM2 Editor? Are there any characters that neve being used by other languages?

Districh-ru:
> Hi @Districh-ru , assuming we chose this path. I think we also need to solve the problem that all characters are shifted by 32 symbols and this causes us additional maintenance and headache. I don't think that extra 1.2 KB of space required for 32 empty images is worth complicating our code. However, this is not directly related to this pull request.

@ihhub, it is OK that font characters are shifted by 32 symbols - we don't need to print the "unprintable control codes" :) But to simplify the code we can add 32 empty images to every font, but that's a matter of the other PR.

> Hi @Districh-ru and @zenseii , can we with almost 100% guarantee identify that a map is made by a French version of the HoMM2 Editor? Are there any characters that neve being used by other languages?

As I remember the original game does not allow to input many non-latin characters (in example `{`, `}`). I'll check it tomorrow.

Districh-ru:
> I checked too with French GOG Dos release and I was able to type these letters `` *#|$&><`@~ `` as seen here: 
I managed to get them by randomly pressing ctrl and alt and alt gr and the number keys. Not sure if this has to do with how French keyboard are the AZERTY variant? Nevertheless I don't know if it matters.

@zenseii, thanks for rechecking!
I only tried Shift + number keys. In English version this combination works while in French it returns only the digits.

The only possible way to deal with this uncommon codepage, IMHO, is:
1. Restore the original ASCII characters for French assets (we will do it in the other PR);
2. Use CP1252 for French language;
3. Update the characters in MP2/MX2 strings _only_ when French assets are found and when French language is selected.


What do you think?

src/fheroes2/gui/ui_language.cpp
```diff
@@ -273,10 +273,15 @@ namespace fheroes2
...
+        // The original French assets replaces several ASCII special characters with language-specific characters.
+        // In the engine we use CP1252 for these characters so we need to generate CP1252 alphabet
+        // to properly draw these characters when English language is selected with French assets.
+        const bool loadFrenchAlphabet = ( language == SupportedLanguage::English ) && ( resourceLanguage == SupportedLanguage::French );
         const bool isOriginalResourceLanguage
-            = ( language == SupportedLanguage::English ) || ( language == resourceLanguage && SupportedLanguage::French != resourceLanguage );
+            = !loadFrenchAlphabet && ( ( language == SupportedLanguage::English ) || ( language == resourceLanguage && resourceLanguage != SupportedLanguage::French ) );
 
-        AGG::updateLanguageDependentResources( language, isOriginalResourceLanguage );
+        AGG::updateLanguageDependentResources( loadFrenchAlphabet ? resourceLanguage : language, isOriginalResourceLanguage );
```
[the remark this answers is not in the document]
Districh-ru: Hi @zenseii, thanks for checking!
I've updated the logic and comments in this place.

src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -90,6 +91,12 @@ namespace
         // create a list of unique maps (based on the map file name) and filter it by the preferred number of players
         std::map<std::string, Maps::FileInfo, std::less<>> uniqueMaps;
 
+        // For the original French version we update the language-specific characters
```
ihhub, line 94: Let's do this way:
```cpp
// Maps made by the original French version Editor or hacked maps could contain
// special ASCII characters that are not supposed to be there.
// While reading the original maps we attempt to fix these characters
// with an assumption that it might alter the original map maker text but to be true to the original game allowed characters.
const bool fixSpecialCharacters
            = isOriginalMapFormat
              && ( fheroes2::getCurrentLanguage() == fheroes2::SupportedLanguage::French && fheroes2::getResourceLanguage() == fheroes2::SupportedLanguage::French );
```
Another question is why do we care about these special characters for other than French language? What consequences we would have by applying this fix irrespective of the game's language?

src/fheroes2/maps/maps_fileinfo.cpp
```diff
@@ -671,6 +684,14 @@ IStreamBase & Maps::operator>>( IStreamBase & stream, FileInfo & fi )
     // Only the filename part of the path to the map file is loaded
     stream >> fi.filename >> fi.name >> fi.description >> fi.width >> fi.height >> fi.difficulty >> kingdommax;
 
+    static_assert( LAST_SUPPORTED_FORMAT_VERSION < FORMAT_VERSION_1109_RELEASE, "Remove the logic below." );
+    if ( Game::GetVersionOfCurrentSaveFile() < FORMAT_VERSION_1109_RELEASE ) {
```
ihhub, line 688: Ideally, we have to check map info version and apply the fix only to maps made by the original editor. This can be done outside stream reading function and inside `Game::Load()` function. We can create dedicated methods for each class to be later removed with bumped save version. If we have such methods then we won't need to pass an extra argument `updateFrenchLanguageSpecificCharacters` but just to call these methods to do the job within `World::LoadMapMP2()` method. The same applies for all other places. In this case then number of places where we apply such fixes would be minimal. What do you think?

ihhub (changes requested):

src/fheroes2/maps/maps_fileinfo.cpp
Districh-ru: Hi @ihhub, 
These special characters contains two printable: `$` and `&`.
These characters may be used in maps and are printable by the original editor in English GOG version

<details><summary>Screenshot</summary>
<p>

изображение

</p>
</details> 

These characters and others that do not have sprites and are represented by dot (`.`) in the original editor (<, >, #, ^, @ and others) can be typed: https://github.com/ihhub/fheroes2/pull/9795#issuecomment-2867999902
So it is safer to do fix only for French assets.

src/fheroes2/maps/maps_fileinfo.cpp
Districh-ru: I introduced these methods but they will stay forever since they are needed to load MP2 maps made with French editor.

ihhub:
@Districh-ru , a bunch of thanks for this change!

PR #9802, opened 2025-05-11
Website Performance Optimization - sub 220ms load times

justinbburris:
## Overview
This PR introduces several optimizations to improve website performance, accessibility, and user experience. The changes focus on modernizing the website's assets and implementing best practices for web development.

### Performance Optimizations
- Converted all font files from TTF to WOFF2 format for better compression and faster loading
- Added `font-display: swap` to all font declarations to improve text rendering performance
- Converted PNG screenshots to WebP format for better compression while maintaining quality
- Added `loading="lazy"` to images for better page load performance
- Implemented proper viewport meta tag for better mobile responsiveness
- Added `defer` attribute to JavaScript files to improve page load time
- Separated font CSS into its own file and implemented preload for critical CSS

### Accessibility Improvements
- Added proper alt text to badges
- Improved image aspect ratios for better visual consistency

### UI/UX Improvements
- Adjusted font sizes for better readability
- Improved responsive layout handling
- Enhanced image showcase layout with proper aspect ratios

# Performance Comparison

My total load times are regularly approaching 220ms, with no cache. However, I also used webpagetest.org to run repeatable benchmarks with timing breakdowns.

### Web test links
[Current performance](https://www.webpagetest.org/result/250511_AiDcVK_63K/)
[New performance](https://www.webpagetest.org/result/250511_AiDcX9_641/)

```
Connectivity: 1600/768 Kbps, 300ms Latency
```
### Results

| Test | Current | New |
|--------|--------|--------|
| LCP | 2.623s | 2.152s |
| First view | 6.057s | 5.110s |
| CLS | 1.047s | 1.062s | 
| Page Weight | 646KB | 562KB |

### Web test links
[Current performance](https://www.webpagetest.org/result/250511_BiDcAP_5R3/)
[New performance](https://www.webpagetest.org/result/250511_AiDcGB_5YQ/)

```
Connectivity: 5000/1000 Kbps, 28ms Latency
```

### Results

| Test | Current | New |
|--------|--------|--------|
| LCP | 1.13s | 0.296s |
| First view | 1.507s | 0.333s |
| CLS | 0.004s | 1.026s | 
| Page Weight | 646KB | 562KB |

### Web test links
[Current performance](https://www.webpagetest.org/result/250511_AiDcVQ_5ZA/)
[New performance](https://www.webpagetest.org/result/250511_BiDc6R_5Z8/)

```
Connectivity: Native
```
### Results

| Test | Current | New |
|--------|--------|--------|
| LCP | 0.474s | 0.286s |
| First view | 0.431s | 0.264s |
| CLS | 0s | 1.036s | 
| Page Weight | 646KB | 562KB |

# Future performance benefits

These are performance benefits that are still available, which haven't been implemented with this work. Mostly because they would require a build step.

1. Inline critical styles into the HTML to reduce CLS (cumulative layout shift) in initial page load (via tool like critical) and minify the rest
2. Minify Javascripts using build process (via tool like terser)
3. Implement a real CDN in front of assets with a longer TTL for static assets (Github pages has a SHORT TTL and this makes the CDN less effective).


oleg-derevenetz:
> Implement a real CDN in front of assets with a longer TTL for static assets (Github pages has a SHORT TTL and this makes the CDN less effective).

As far as I understand, this will require using another hosting instead of GitHub Pages, because GitHub Pages does not allow to manipulate HTTP headers, right? I suppose we would like to avoid this.

justinbburris:
> > Implement a real CDN in front of assets with a longer TTL for static assets (Github pages has a SHORT TTL and this makes the CDN less effective).
> 
> As far as I understand, this will require using another hosting instead of GitHub Pages, because GitHub Pages does not allow to manipulate HTTP headers, right? I suppose we would like to avoid this.

Yep - let me expand on this point

1. Github pages don't have great TTLs for their headers. This is _mostly_ fine but it's unfortunate in that we'll get constant cache misses when they time out. So it's fine for a session, but not great for multiple sessions. This may not be a concern (and considering we're already doing great with our performance, it probably is)
2. If it was a big enough concern, we have 2 paths forward
  a. Serve via another hosting provider
  b. Use a domain to proxy to GH pages

Both options are likely non-starters, but I figured it'd be good to call it out as another optimization that could be done in the future.

oleg-derevenetz (changes requested):
Hi @justinbburris technically looks good to me, except for one thing: you need to synchronize changes between `docs/README.md` and `README.md`. Otherwise, the new webp images will not be displayed in the repository description on GitHub itself and the wording will be different.

justinbburris:
> Hi @justinbburris technically looks good to me, except for one thing: you need to synchronize changes between `docs/README.md` and `README.md`. Otherwise, the new webp images will not be displayed in the repository description on GitHub itself and the wording will be different.

Gah! I had that fixed in a commit I dropped (to clear out some other unrelated changes) 

Sorry for the sloppy submission. I'll get that fixed. 

🙇

docs/README.md
```diff
@@ -41,7 +43,7 @@ Initially, the project was developed on [**sourceforge**](https://sourceforge.ne
 
 ## Contribution and Development
 
-This repository is a place for everyone. If you want to contribute, please read more [**here**](https://github.com/ihhub/fheroes2/wiki/F.A.Q.#q-how-can-i-contribute-to-the-project).
+This repository is a place for everyone. If you want to contribute, please read more to learn [**how to contribute**](https://github.com/ihhub/fheroes2/wiki/F.A.Q.#q-how-can-i-contribute-to-the-project).
```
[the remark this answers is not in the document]
justinbburris: love it 🙌

justinbburris:
> Hi @justinbburris , I have unrelated to this pull request question: what do you think about generation of web pages from wiki section of this project? Users shouldn't be "redirected " to dev related pages when they browse the project's website.

Agreed - and this is tricky. The wiki feature, afaik, cannot be directly embedded into the Github Pages site as is. What we could look into is one of three paths. I've ordered this list from what I'd recommend, given the tradeoffs.

### Option 1. Completely move away from Github Wikis and move into docs in repo (and thus natively into Github Pages)

This _could_ be generally good as the wiki is locked to changes to a subset of users. Putting the content in the repo, the number of contributors would naturally rise, without going through a central source. Some pages could be converted to other Github features, like projects, etc.

### Option 2. Port specific pages into Github pages

If the page is linked form the doc site, then they should live in the docs

### Option 3. Use a tool to convert our wiki into pages embedded into the doc site at build time.

This would require a build step (we're talking about doing this anyway) and would be a bit tricky to implement in such a way that the devex makes sense, but we could go this path.

docs/README.md
```diff
@@ -61,18 +63,34 @@ To build the [website](https://ihhub.github.io/fheroes2/) from source, please fo
...
+    <img loading="lazy" width="61" height="20" src="https://img.shields.io/badge/Facebook-blue.svg" alt="Facebook" />
+</a>
+<a href="https://vk.com/fheroes2">
+    <img loading="lazy" width="25" height="20" src="https://img.shields.io/badge/VK-blue.svg" alt="VK" />
+</a>
+<a href="https://discord.gg/xF85vbZ">
+    <img loading="lazy"
+        width="93"
+        height="20"
+        src="https://img.shields.io/discord/733093692860137523.svg?label=&amp;logo=discord&amp;logoColor=ffffff&amp;color=7389D8&amp;labelColor=6A7EC2"
```
oleg-derevenetz, line 91: Hi @justinbburris are you sure that in this URL should be `&amp;` and not `&`?

justinbburris: Great Catch! 🙇 

It shouldn't be written like this.

I'm not certain how this happened. I copied the HTML straight from the browser, which did not have the URL encodings. But then the commit did for sure. I'll do some testing tonight and see if my editor "helped" me with the URL.

justinbburris:
Oh Good catch on the alt-text. Nice to make that improvement everywhere.

PR #9804, opened 2025-05-12
Update the German translation

do-ut-des:
This PR enables various German translations after they had been reviewed and fixes some translation issues.


files/lang/de.po
```diff
@@ -176,29 +176,26 @@ msgstr ""
 msgid "QUIT"
 msgstr "BEENDEN"
 
-#, fuzzy
 msgid ""
 "LOAD\n"
 "MAP"
 msgstr ""
 "SPIEL\n"
 "LADEN"
```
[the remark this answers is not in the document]
felix-schneider: No it's not.
Load map = Karte laden
Load game = Spiel laden

files/lang/de.po
```diff
@@ -176,29 +176,26 @@ msgstr ""
...
 "SPIEL"
 
-#, fuzzy
 msgid ""
 "SAVE\n"
 "MAP"
 msgstr ""
 "SPIEL\n"
-"SPEICH."
+"SPEICHERN"
```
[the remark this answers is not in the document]
felix-schneider: Same answer as above,
The abbreviation is because of space on the button, is this resolved in fheroes2?

files/lang/de.po
```diff
@@ -2022,21 +1980,18 @@ msgid ""
 "The main hero will have %{count} %{monster} at the start of the scenario."
 msgstr "Der Hauptheld hat zu Beginn des Szenarios %{count} %{monster}."
 
-#, fuzzy
 msgid ""
 "The main hero will have the \"%{spell}\" spell at the start of the scenario."
-msgstr "Der Hauptheld hat zu Beginn des Szenarios einen \"%{spell}\"-Zauber."
+msgstr "Der Hauptheld hat zu Beginn des Szenarios den Zauberspruch \"%{spell}\"."
```
felix-schneider, line 1985: See note above about Zauber vs. Zauberspruch

files/lang/de.po
```diff
@@ -2066,33 +2021,29 @@ msgid "For this action it is necessary to build a castle first."
...
+"%{name} kann nicht gebaut werden, da kein Gewässer in der Nähe ist."
 
 msgid "This building has been disabled."
-msgstr ""
+msgstr "Dieses Gebäude wurde deaktiviert."
 
-#, fuzzy
 msgid "Cannot afford the %{name}."
-msgstr "%{name} ist zu teuer."
+msgstr "Für %{name} fehlen Euch die Resourcen."
```
felix-schneider, line 2040: "Ihr könnt euch %{name} nicht leisten"?

files/lang/de.po
```diff
@@ -2722,17 +2660,14 @@ msgstr ""
...
 "pro Woche."
 
-#, fuzzy
 msgid "The Statue increases the town's income by %{count} gold per day."
 msgstr "Die Statue erhöht das Einkommen Eurer Stadt um %{count} Gold pro Tag."
 
 msgid "The Left Turret provides extra firepower during castle combat."
-msgstr ""
-"Der linker Geschützturm bietet zusätzliche Feuerkraft während der Belagerung."
+msgstr "Der linke Geschützturm bietet zusätzliche Feuerkraft während der Belagerung."
```
felix-schneider, line 2667: während _einer_ Belagerung?

files/lang/de.po
```diff
@@ -2743,23 +2678,23 @@ msgstr ""
...
+"Burggraben aufhält, wird ihre Verteidigungsfähigkeit um %{count} reduziert."
 
-#, fuzzy
 msgid ""
 "The Castle improves the town's defense and increases its income to %{count} "
 "gold per day."
 msgstr ""
 "Die Burg verbessert die Stadtverteidigung und erhöht das Einkommen auf "
-"%{count} Gold pro Tag."
+"%{count} Goldstücke pro Tag."
```
felix-schneider, line 2697: The statue says gold, not gold pieces, I would stick with that.

files/lang/de.po
```diff
@@ -2876,19 +2811,17 @@ msgstr "Verstreute Formation"
 msgid "Grouped Formation"
 msgstr "Gruppierte Formation"
 
-#, fuzzy
 msgid "Click to show the next town."
-msgstr "Klicken um die nächste Stadt anzuzeigen."
+msgstr "Klicke, um die nächste Stadt anzuzeigen."
```
felix-schneider, line 2815: The game adresses the player as "Ihr", so this is not correct. I find the instruction to click a bit unnecessary, so maybe the shorter "Nächste Stadt anzeigen"?

files/lang/de.po
```diff
@@ -2970,10 +2903,10 @@ msgstr ""
 "bereit, Euch in einer Fähigkeit Eurer Wahl zu unterrichten."
 
 msgid "Attack Skill"
-msgstr "Angriff"
+msgstr "Angriffsfähigkeit"
 
 msgid "Defense Skill"
-msgstr "Verteidigung"
+msgstr "Verteidigungsfähigkeit"
```
felix-schneider, line 2909: I was chewing on this during my translation of artifacts.
Verify this in the UI, this may change the display in the hero screen or exceed the allotted space in the creature screen.

While many artifact descriptions refer to "Angriffsfähigkeit", it may be better to stick with "Angriff" and "Verteidigung" here.

files/lang/de.po
```diff
@@ -3417,12 +3347,11 @@ msgstr "schon gelernt"
 msgid "treeOfKnowledge|free"
 msgstr "Baum des Wissens"
 
-#, fuzzy
 msgid "already claimed"
-msgstr "schon gelernt"
+msgstr "schon beansprucht"
```
felix-schneider, line 3351: The translation is correct, but sounds a bit awkward? I think the meaning here is claimed as in a reward. Is this for the witch's hut? In that case, "gelernt" would be entirely appropriate.

files/lang/de.po
```diff
@@ -3439,9 +3368,8 @@ msgstr "(nicht besucht)"
 msgid "%{color} Barrier"
 msgstr "%{color} Barriere"
 
-#, fuzzy
 msgid "(tent visited)"
-msgstr "(nicht besucht)"
+msgstr "(Zelt besucht)"
 
 msgid "%{color} Tent"
 msgstr "%{color} Zelt"
```
felix-schneider, line 3375: How is the color translated? "Rot Zelt" would be incorrect. Would need to check all the colors, but we may be lucky and `%{color}es Zelt` is correct for all of them?

files/lang/de.po
```diff
@@ -3612,36 +3536,29 @@ msgstr "Held auswählen:"
...
 
-#, fuzzy
 msgid "Click to start placing the selected hero."
-msgstr "Klicken um die ausgewählte Auflösung anzuwenden."
+msgstr "Klickt, um die Platzierung des ausgewählten Helden zu beginnen."
 
-#, fuzzy
 msgid "%{color} %{race}"
-msgstr "Spieler %{color}"
+msgstr "%{race} %{color}"
```
felix-schneider, line 3543: "Totenbeschwörer Rot"? Normally the color should go first, but then it needs to be adjusted to the class name :/

files/lang/de.po
```diff
@@ -3612,36 +3536,29 @@ msgstr "Held auswählen:"
...
 
-#, fuzzy
 msgid "Select Treasure:"
-msgstr "Held auswählen:"
+msgstr "Wählt den Schatz:"
 
-#, fuzzy
 msgid "Select Ocean Object:"
-msgstr "Ozean-Objekte"
+msgstr "Wählt das Ozeanobjekt"
```
felix-schneider, line 3558: Original ends in :

files/lang/de.po
```diff
@@ -3671,26 +3585,22 @@ msgstr "Spieler %{color}"
...
 msgid "Mine placing"
-msgstr ""
+msgstr "Minenplatzierung"
 
-#, fuzzy
 msgid ""
 "Resource\n"
 "type:"
-msgstr "Ressource"
+msgstr "Ressourcentyp:"
```
felix-schneider, line 3600: If "Resource type" had to have a newline in it (why is the newline hardcoded in? Other buttons wrap automatically?)
then "Ressourcentyp" will almost certainly not fit. "Ressource" is adequate.

files/lang/de.po
```diff
@@ -4189,31 +4097,29 @@ msgstr "Eine vorhandene Karte laden."
 msgid "No Secondary Skill will be given by the event."
 msgstr ""
 
-#, fuzzy
 msgid "Secondary Skill"
-msgstr "Fähigkeit auswählen:"
+msgstr "Sekundäre Fertigkeit"
```
felix-schneider, line 4101: Sekundärfertigkeit

files/lang/de.po
```diff
@@ -4189,31 +4097,29 @@ msgstr "Eine vorhandene Karte laden."
...
 msgid ""
 "If this checkbox is checked, the event will trigger only once. If not "
 "checked, the event will trigger every time one of the specified players "
 "crosses the event tile."
 msgstr ""
 
-#, fuzzy
 msgid "%{objects} cannot be placed on water."
-msgstr "Monster können nicht auf Wasser plaziert werden."
+msgstr "%{objects} kann nicht auf Wasser platziert werden."
```
felix-schneider, line 4122: Unclear here: The placeholder implies a plural and the English is ambiguous. This would need two translation strings.

files/lang/de.po
```diff
@@ -10006,19 +9912,19 @@ msgid "Iron Golem"
...
-msgstr "Stahlgoleme"
+msgstr "Stahlgolems"
 
 msgid "Roc"
-msgstr "Roc"
+msgstr "Roch"
 
 msgid "Rocs"
-msgstr "Rocs"
+msgstr "Rochs"
```
felix-schneider, line 9927: The OG says Roc, I would not change it.

files/lang/de.po
```diff
@@ -10084,16 +9990,16 @@ msgid "Vampire Lords"
...
-msgstr "Liches"
+msgstr "Leichen"
 
 msgid "Power Lich"
-msgstr "Meister-Lich"
+msgstr "Meister-Leiche"
 
 msgid "Power Liches"
-msgstr "Meister-Liches"
+msgstr "Meister-Leichen"
```
felix-schneider, line 10002: Hard disagree on Leichen, they're not corpses.
The original uses the English plural, which could be improved: "Ein Lich", "Zwei Liche", but definitely not Leichen.

D&D coined the term, the German translations of D&D use "Lich":
https://de.wikipedia.org/wiki/Lich_(Untoter)
https://mythical.ink/de/rpg-tools/dnd-monsters?q=Lich

files/lang/de.po
```diff
@@ -10123,10 +10029,10 @@ msgid "Genie"
 msgstr "Dschinn"
 
 msgid "Genies"
-msgstr "Dschinne"
+msgstr "Dschinns"
```
felix-schneider, line 10032: According to Wikipedia, the plural is "Dschinn": https://de.wikipedia.org/wiki/Dschinn
I don't speak Arabic, so I cannot confirm.

felix-schneider (commented):
As usual, I have a lot of thoughts about language 😅
Some of the comments actually concern the English strings as well @zenseii

files/lang/de.po
```diff
@@ -10282,56 +10188,46 @@ msgid "Soul Eater"
...
 
-#, fuzzy
 msgid "Fire creature"
-msgstr "Kreaturen rekrutieren"
+msgstr "Feuerwesen"
 
-#, fuzzy
 msgid "Water creature"
-msgstr "See"
+msgstr "Wasserwesen"
```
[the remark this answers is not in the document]
felix-schneider: So why are they water creatures and not water elementals?

files/lang/de.po
```diff
@@ -10282,56 +10188,46 @@ msgid "Soul Eater"
...
 
-#, fuzzy
 msgid "Double damage from Water creatures"
-msgstr "Entfernt alle Zaubersprüche von allen Kreaturen."
+msgstr "Doppelter Schaden von Wasserwesen."
 
 msgid "Lightning"
 msgstr "Blitz"
 
 msgid "% immunity to "
```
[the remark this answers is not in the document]
felix-schneider: Other strings with numbers in them (even when that number is always one, like with the resource producing items) have placeholders, so why not this one? Also after "to"

files/lang/de.po
felix-schneider: Yeah...
image
image

files/lang/de.po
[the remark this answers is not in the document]
felix-schneider: No linguistic problem in this one, but with how aggressively placeholders were used in the artifacts, I find it strange that they aren't here.

files/lang/de.po
```diff
@@ -1066,29 +1036,26 @@ msgstr "Mensch"
...
-msgstr ""
+msgstr "Auf Standardeinstellungen setzen."
 
 msgid "Please select another hero."
 msgstr "Bitte wählt einen anderen Helden aus."
 
-#, fuzzy
 msgid "%{race1} %{name1} vs %{race2} %{name2}"
-msgstr "%{name1} trifft %{name2}"
+msgstr "%{race1} %{name1} vs %{race2} %{name2}"
```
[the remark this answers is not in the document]
felix-schneider: I checked the game, I think this string is for the Battle Mode added in fheroes2, where you can just have a standalone combat. In that case, I would prefer "gegen". The title of the trading screen is a different string that says "trifft" and does not name the classes.

One problem with <class> <name> is female heroes of factions with a male title (all except Sorceress): "Zauberer Myra", "Totenbeschwörer Darlana", "Barbar Jezebel". It's not **that** bad, but not strictly correct either. Then again, I guess "Jezebel (Barbar)" isn't much better. I will leave it up to you.

files/lang/de.po
```diff
@@ -1687,26 +1650,21 @@ msgstr " werden immer vor Eurer Armee weglaufen."
...
 
-#, fuzzy
 msgid "\"%{artifact}\" artifact will be carried over in the campaign."
-msgstr "Das Artefakt \"%{artifact}\" wird in das Szenario übernommen."
+msgstr "Das Artefakt \"%{artifact}\" wird in die Kampagne übernommen."
 
-#, fuzzy
 msgid "The army will be carried over in the campaign."
-msgstr "Die Armee wird über das Szenario mitgenommen."
+msgstr "Die Armee wird idn die Kampagne übernommen."
```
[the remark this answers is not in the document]
felix-schneider: No, the army is carried over within the campaign from scenario to scenario (although I think this is in roland 7, and it doesn't carry further than a single scenario).

I don't know where the original strings are from, seems like they are some kind of automatic translation or fuzzy matching of existing strings. Many of them blatantly do not say the same thing as the English.

files/lang/de.po
```diff
@@ -470,9 +461,8 @@ msgstr ""
 msgid "You have already cast a spell this round."
 msgstr "Ihr habt in dieser Runde bereits einen Zauberspruch ausgesprochen."
 
-#, fuzzy
 msgid "That spell will have no effect!"
-msgstr "Dieser Zauber wirkt sich auf niemanden aus!"
+msgstr "Dieser Zauberspruch wird keinen Effekt haben!"
```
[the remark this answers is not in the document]
felix-schneider: Re: Zauber vs Zauberspruch: Yes there can be subtle distinctions, my preference for Zauber is purely practical: The game often does not leave a lot of room for text and does not automatic hyphenation, so I prefer the shorter word where possible.

Re: Mana: Mana not Manna https://de.wikipedia.org/wiki/Mana

files/lang/de.po
felix-schneider: 1001 Nights is originally in Arabic, the transliteration into Latin characters is not exact. (German) Wikipedia lists Roch, Roc, Rokh, Ruch and Rock as alternate spellings. In English, I saw even more variants.

The English version of the game uses Roc in both Heroes 2 and 3, not ending in a /x/ sound and the German translators kept it, for both Heroes 2 and 3. I would not go around changing that. This project often takes a conservative approach to these sorts of "fixes" because people are familiar with the original and it is strange to see the name suddenly be different.

files/lang/de.po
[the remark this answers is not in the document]
felix-schneider: Where is this string even used? The description window for the Elementals does not use this string, they only take double damage from certain spells, not from elementals of the opposite type?

files/lang/de.po
```diff
@@ -2876,19 +2811,17 @@ msgstr "Verstreute Formation"
...
+msgstr "Klicke, um die nächste Stadt anzuzeigen."
 
-#, fuzzy
 msgid "Click to show the previous town."
-msgstr "Klicken um die vorherige Stadt anzuzeigen."
+msgstr "Klicke, um die vorherige Stadt anzuzeigen."
 
 msgid "Set garrison combat formation to 'Spread'"
-msgstr "Garnisonskampfformation auf \"Ausbreitend\" einstellen"
+msgstr "Die Kampfformation der Garnison auf \"Ausbreiten\" einstellen"
```
felix-schneider, line 2821: Just checked the hero screen and it says "verstreute Formation"

files/lang/de.po
```diff
@@ -963,43 +940,37 @@ msgstr "Der %{spell} fügt einer lebenden Kreatur %{damage} Schaden zu."
 msgid "The %{spell} does %{damage} damage to all living creatures."
 msgstr "Der %{spell} verursacht %{damage} Schaden an allen lebenden Kreaturen."
 
-#, fuzzy
 msgid "The %{attacker}'s attack blinds the %{target}!"
 msgid_plural "The %{attacker}' attack blinds the %{target}!"
-msgstr[0] "%{target} wegen des Angriffs von %{attacker} geblendet!"
-msgstr[1] "%{target} wegen des Angriffs von %{attacker} geblendet!"
+msgstr[0] "Des %{attacker}s Angriff blendet %{target}!"
+msgstr[1] "Der %{attacker} Angriff blendet %{target}!"
```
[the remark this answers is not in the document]
felix-schneider: I'm with you on not changing the intent, although it is also possible to read too much into a message like this.

I'm not opposed to the Genitive, but the correct Genitive is "Der Angriff des Einhorns blendet X". The original is not being overly poetic, so there is no reason the translation should be.

files/lang/de.po
felix-schneider: You can visit the tree by refuse to pay the cost.

files/lang/de.po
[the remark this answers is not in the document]
felix-schneider: Mana is a very common term in Fantasy games. As you saw with "Angriffsfertigkeit", sometimes you have to make a compromise between the correct translation and the UI.

But we can stick with this translation until a specifc UI issue appears on which do discuss it.

files/lang/de.po
```diff
@@ -963,43 +940,37 @@ msgstr "Der %{spell} fügt einer lebenden Kreatur %{damage} Schaden zu."
...
 
-#, fuzzy
 msgid "The %{attacker}'s gaze turns the %{target} to stone!"
 msgid_plural "The %{attacker}' gaze turns the %{target} to stone!"
-msgstr[0] "Der Blick des %{attacker} verwandelt das %{target} in Stein!"
-msgstr[1] "Der Blick des %{attacker} verwandelt das %{target} in Stein!"
+msgstr[0] "Des %{attacker} Blick verwandelt das %{target} in Stein!"
+msgstr[1] "Der %{attacker} Blicke verwandeln das %{target} in Stein!"
```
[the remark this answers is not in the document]
felix-schneider: Ok, if you prefer the longer one that's ok. I think it's "zu Stein verwandeln" though? Not "in"?

files/lang/de.po
felix-schneider: That looks even worse, maybe "Zelt (Orange)"?

felix-schneider: Are you mistaking the object for something else? It's a tent that is orange, It's not a color that's on the map.

files/lang/de.po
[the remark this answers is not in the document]
felix-schneider: I don't understand, what is wrong with
"Der Angriff des Einhorns blendet X"?
This is correct German, uses the Genitive and matches the English meaning exactly. The word order is not the same which is normal for possessive constructs.

files/lang/de.po
```diff
@@ -963,43 +941,37 @@ msgstr "Der %{spell} fügt einer lebenden Kreatur %{damage} Schaden zu."
...
-msgstr[1] "Die %{target} sind durch den %{attacker} gelähmt!"
+msgstr[0] "%{target} ist durch den %{attacker} paralysiert!"
+msgstr[1] "Die %{target} sind durch den %{attacker} paralysiert!"
 
-#, fuzzy
 msgid "The %{attacker} dispels all good spells on your %{target}!"
 msgid_plural "The %{attacker} dispel all good spells on your %{target}!"
 msgstr[0] ""
-"Der %{attacker} verbannt alle guten Zaubersprüche auf Eurem %{target}!"
+"%{attacker} verbannt alle guten Zaubersprüche von Eurem %{target}!"
```
felix-schneider, line 967: Only just noticed this:
The dispel magic spell is "Magie zerstreuen". To clarify that this is the same effect, this should probably also use "zerstreuen"

files/lang/de.po
felix-schneider: Ah, no, these are the tents which unlock the corresponding colored gates. They can be visited by each player, but are not flagged.

files/lang/de.po
felix-schneider: In #9285 I proposed "Gegenzauber", which would not address the verb here.

I guess we can keep the translation here, but we will have to be careful to find all these usages when we do change the translation of the spell. Best to search for the English text for usages.

felix-schneider:
I think that's all of my discussions addressed.

ihhub (approved):

ihhub:
@do-ut-des , thank you so much for you contribution to the project!

PR #9806, opened 2025-05-12
Implement the infrastructure for bundling FH2M maps

### LINKED ISSUE #9803: Modify package installer scripts to include maps directory with fh2m maps
### Preliminary checks

Starting from 1.1.9 release we are going gradually to add user created maps into the fheroes2 package. As of now our scripts do not accommodate this. We need to make preparations for the future files.

Modify existing scripts to include **maps** directory to be copied on the user's machines.

### Additional info

_No response_

oleg-derevenetz:
close #9803

I added the `empty.fh2m` map to have some kind of map for testing purposes. As soon as any real maps are added, it can be deleted.


Districh-ru (approved):
LGTM.
I've checked Windows, Android and Vita packages - they all contain `empty.fh2m` and this map is installed in the `maps` folder located next to the `files` folder.
But before merging this PR the test `empty.fh2m` should be removed.

oleg-derevenetz:
> But before merging this PR the test `empty.fh2m` should be removed.

In general, I thought to keep it in the repository a bit longer so that more willing people could test the bundling of maps on more devices, and remove it as soon as at least one "real" map appears, or immediately before the next release.

ihhub:
Hi @oleg-derevenetz , we are going to ads 1-2 maps in the upcoming days so we can use them instead of the dummy one.

oleg-derevenetz:
Hi @ihhub

> we are going to ads 1-2 maps in the upcoming days so we can use them instead of the dummy one.

No problem - we'll delete the temporary map as soon as the real maps are added, right in the same PR. And until that happens, let the dummy map stay for a while.

ihhub:
@oleg-derevenetz , I absolutely agree with this.

ihhub (approved):

ihhub:

PR #9814, opened 2025-05-17
Add an optional field for map creator's notes

ihhub:
This field is optional and requires to be filled for map makers providing maps to be a part of the fheroes2 bundle. This is a much more elegant solution rather than proving contact details in the description of the map or in a separate file.

image

### DIFF
--- a/src/fheroes2/editor/editor_map_specs_window.cpp
+++ b/src/fheroes2/editor/editor_map_specs_window.cpp
@@ -2262,6 +2262,10 @@ namespace Editor
         background.renderTextAdaptedButtonSprite( buttonLanguage, translatedText, { 20 + buttonRumors.area().width + buttonEvents.area().width + 2 * 10, 6 },
                                                   fheroes2::StandardWindow::Padding::BOTTOM_LEFT );
 
+        fheroes2::ButtonSprite buttonCreatorNotes;
+        translatedText = fheroes2::getSupportedText( gettext_noop( "INFO" ), fheroes2::FontType::buttonReleasedWhite() );
+        background.renderTextAdaptedButtonSprite( buttonCreatorNotes, translatedText, { 21, 12 }, fheroes2::StandardWindow::Padding::TOP_RIGHT );
+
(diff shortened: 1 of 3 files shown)


src/fheroes2/editor/editor_map_specs_window.cpp
```diff
@@ -2262,6 +2262,10 @@ namespace Editor
         background.renderTextAdaptedButtonSprite( buttonLanguage, translatedText, { 20 + buttonRumors.area().width + buttonEvents.area().width + 2 * 10, 6 },
                                                   fheroes2::StandardWindow::Padding::BOTTOM_LEFT );
 
+        fheroes2::ButtonSprite buttonCreatorNotes;
+        translatedText = fheroes2::getSupportedText( gettext_noop( "INFO" ), fheroes2::FontType::buttonReleasedWhite() );
```
Districh-ru, line 2266: We should use another word or "editorMap|INFO" not to interfere with "INFO" in Adventure Options:
изображение

In example for better understanding in Russian it was translated like not "Info", but "Scena-`\n`rio" because it shows the scenario data.

Districh-ru (changes requested):
Hi @ihhub, I left one comment. Could you please take a look when you have time?

PS
And a question: will we allow (in future, not in this PR) the players to see this INFO? In example, in scenario select dialog by pressing some new button or in the right click dialog.

src/fheroes2/editor/editor_map_specs_window.cpp
ihhub: How about the following:
- **ABOUT**
- **NOTES** (very easy to confuse with events or rumors)

I am fine with your **editorMap|INFO** proposal as well. @zenseii , what do you think?

Districh-ru: @ihhub and @zenseii,
For me **ABOUT** suits more because there will be a text about map and it's creators.
And I'm fine with **editorMap|INFO**.

ihhub: I changed to `ABOUT`.

ihhub:
Hi @zenseii , could you please also review the new text?

Districh-ru (approved):

PR #9817, opened 2025-05-17
Fixed behavior of buttons when pressing down on them

usidedown:
Buttons continue to behave as if pressed even when mouse moves away:

button

This PR adds a check that the mouse is still within the button boundaries:
button_new

This behavior is more consistent with the original game and the way buttons function.

It may look like a big change, but it's not bad when reviewed commit-by-commit.
Most of the changes are to drawOnState of buttons - I kept all of them in a single commit.
I also fixed a couple of other buttons I encountered when playing - they are in separate commits.

I intentionally introduced a new function rather than override all the usages of `isMouseLeftButtonPressedInArea`.
1. Some of the usages are already correct.
2. It's easier to see I didn't break something unintentionally.

### DIFF
--- a/src/fheroes2/gui/interface_list.h
+++ b/src/fheroes2/gui/interface_list.h
@@ -333,8 +333,8 @@ namespace Interface
         {
             LocalEvent & le = LocalEvent::Get();
 
-            _buttonPgUp.drawOnState( le.isMouseLeftButtonPressedInArea( _buttonPgUp.area() ) );
-            _buttonPgDn.drawOnState( le.isMouseLeftButtonPressedInArea( _buttonPgDn.area() ) );
+            _buttonPgUp.drawOnState( le.isMouseLeftButtonPressedAndHeldInArea( _buttonPgUp.area() ) );
+            _buttonPgDn.drawOnState( le.isMouseLeftButtonPressedAndHeldInArea( _buttonPgDn.area() ) );
 
             if ( !IsValid() ) {
                 return false;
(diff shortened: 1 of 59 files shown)


src/fheroes2/gui/interface_list.h
```diff
@@ -448,19 +448,19 @@ namespace Interface
...
 
                 UpdateScrollbarRange();
 
-                if ( le.isMouseLeftButtonPressedInArea( _scrollbar.getArea() ) && ( _size() > maxItems ) ) {
+                if ( le.isMouseLeftButtonPressedAndHeldInArea( _scrollbar.getArea() ) && ( _size() > maxItems ) ) {
                     _scrollbar.moveToPos( mousePosition );
                 }
 
-                if ( le.isMouseLeftButtonPressedInArea( rtAreaItems ) ) {
+                if ( le.isMouseLeftButtonPressedAndHeldInArea( rtAreaItems ) ) {
```
Districh-ru, line 463: Hi @usidedown, I suggest to revert the changes related to the scrollbar and list dragging.

The scroll bar is thin to always keep cursor right over it: in the OS if you start scrolling by the scrollbar you can move your cursor out for some range but still continue scrolling.
Yes, the fheroes2 scroll bar has no such range limit to stop scrolling when the cursor is too far from the scroll bar, but lets add it in a new PR if it is really needed (IMHO it is not needed). :)

Also about dragging: with `isMouseLeftButtonPressedInArea()` after you start dragging you may scroll the list more than only inside the visible items area, like in most Android applications (in example, my Contacts app continue drag-scrolling even when my finger goes over the top or bottom buttons).

Districh-ru (approved):
Well done!

ihhub:
Thank you so much for these great changes, @usidedown !

PR #9818, opened 2025-05-17
Add Play Online Feature to Website

justinbburris:
## Overview
This PR introduces a new "Play Online" feature to our website, allowing users to play fheroes2 directly in their browser. This is implemented through Emscripten integration and requires changes to our site deployment process to support it.

> [!IMPORTANT]
> Try and [play online](https://justinbburris.github.io/fheroes2/play-online)! 

## Core Changes
- Added new `/play-online` section to the website
- Integrated Emscripten build into the site structure
- Upload Emscripten builds into latest-commit releases
- Created GitHub Actions workflow to automate the build and deployment process

## Technical Implementation
The implementation required changes to our deployment process to support the Emscripten build:
- New GitHub Actions workflow (`.github/workflows/pages.yml`) that:
  - Builds the Jekyll site
  - Downloads and integrates the latest Emscripten release
  - Deploys the combined site to GitHub Pages

## Required Repository Changes
To enable this new deployment method, the following changes are needed in the repository settings:

1. Go to Settings > Pages
2. Under "Build and deployment":
   - Source: Select "GitHub Actions"
   - Branch: This will be handled by the workflow

## Testing
The play-online feature has been tested with:
- Latest Emscripten build integration
- Proper site structure and navigation
- Automatic deployment process

## Migration Steps
1. Update repository settings as described above
2. Merge this PR
3. Monitor the first automatic deployment
4. Verify the play-online feature is accessible at https://ihhub.github.io/fheroes2/play-online/

## Notes
- The pages site build is automatically updated with each push to master
- The workflow includes proper concurrency handling to prevent deployment conflicts
- All necessary permissions are explicitly defined in the workflow

## Testing
- Tested successfully on macOS Chrome
- Tested semi-successfully on iPhone
  - Game plays but no audio. 🤔

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,62 @@
+name: Build and deploy website
+
+on:
+  push:
+    branches: [ master ]
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Pages
+        id: pages
+        uses: actions/configure-pages@v5
+
+      - name: Build with Jekyll
+        uses: actions/jekyll-build-pages@v1
+        with:
+          source: ./docs
+          destination: ./_site
+
+      - name: Download Emscripten release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release download fheroes2-emscripten-sdl2_dev -D ./dist
+
+      - name: Extract and prepare site
+        working-directory: ./dist
+        run: |
+          mkdir -p ./site/play-online
+          unzip -q fheroes2_emscripten.zip -d site/play-online
+          cp -r ../_site/* ./site/
+          rm -f fheroes2_emscripten.zip
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: "./dist/site"
+
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{steps.deployment.outputs.page_url}}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -16,6 +16,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     defaults:
       run:
         shell: bash
@@ -45,6 +46,7 @@ jobs:
     needs:
     - build
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     defaults:
       run:
         shell: bash
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -6,8 +6,8 @@ on:
 
 permissions:
   contents: write
-  pages: write
   id-token: write
+  pages: write
 
 concurrency:
   group: "pages"
@@ -19,23 +19,19 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v5
-
       - name: Build with Jekyll
         uses: actions/jekyll-build-pages@v1
         with:
           source: ./docs
           destination: ./_site
-
       - name: Download Emscripten release
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           gh release download fheroes2-emscripten-sdl2_dev -D ./dist
-
       - name: Extract and prepare site
         working-directory: ./dist
         run: |
@@ -43,12 +39,10 @@ jobs:
           unzip -q fheroes2_emscripten.zip -d site/play-online
           cp -r ../_site/* ./site/
           rm -f fheroes2_emscripten.zip
-
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v3
         with:
           path: "./dist/site"
-
   deploy:
     environment:
       name: github-pages
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -15,6 +15,7 @@ concurrency:
 
 jobs:
   build:
+    name: Build website
     runs-on: ubuntu-latest
     timeout-minutes: 10
     defaults:
@@ -43,6 +44,7 @@ jobs:
         with:
           path: ./dist/site
   deploy:
+    name: Deploy website
     needs:
     - build
     runs-on: ubuntu-latest
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -40,7 +40,7 @@ jobs:
           rm -f fheroes2_emscripten.zip
       - uses: actions/upload-pages-artifact@v3
         with:
-          path: "./dist/site"
+          path: ./dist/site
   deploy:
     needs:
     - build
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -10,20 +10,19 @@ permissions:
   pages: write
 
 concurrency:
-  group: "pages"
+  group: pages
   cancel-in-progress: false
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
     steps:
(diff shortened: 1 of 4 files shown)


justinbburris:
cc @oleg-derevenetz 🙇 
Step 1 of 3 :)

.github/workflows/pages.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/jekyll-build-pages@v1
+        with:
+          source: ./docs
+          destination: ./_site
+
+      - name: Download Emscripten release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release download fheroes2-emscripten-sdl2_dev -D ./dist
```
oleg-derevenetz, line 38: Hi @justinbburris where exactly does this release should come from? At the moment, as far as I remember, there is no special WebAssembly release in the Releases section. Moreover, the WebAssembly build runs only on pull requests.

justinbburris: 🤦 This is what I get for trying to work on this while my kids hang on me. One moment, I have the release work in the original cloudflare branch, let me pull it in.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,64 @@
+name: Build and deploy website
+
+on:
+  workflow_run:
+    workflows: ["Push"]
+    types:
+      - completed
```
oleg-derevenetz, line 7: I believe it's worth doing this within the existing paradigm by replacing `workflow_run:` with `workflow_call:` and calling this workflow from the `push.yml` via the `needs:` dependency - just like it's done for other similar workflows (like the `translation_update.yml`, for instance). But if it's more convenient for you to leave this as is for now - for testing purposes, for instance - then it's not urgent at all. What do you think?

justinbburris: I debated this myself, quite a bit, and I ultimately went with this approach because I liked how it reacts in response to the completion of this workflow, but its not directly related to it. Right now we're deploying the site whenever we build a new version, but we might want to setup similar rules for whenever we tag a release (so that we'd have something like beta-play-online and play-online). 

In that situation, having the logic decoupled from the core workflow would make sense.

I don't have a strong preference here, I was mostly thinking about where we might go and what might be more maintainable in that situation.

GHA is pretty flexible, and the complexity is low, so we can embed it into the job directly.

oleg-derevenetz: What I don't like in `workflow_run` is that while with the `workflow_call` the target workflow runs in the context of a calling workflow, with the `workflow_run` the target workflow always runs in a privileged context (it has all the access to the repo secrets, etc) regardless of the context and privileges of a calling workflow. Although in this case it doesn't seem very dangerous (since the `Push` workflow runs in privileged context anyway), I have a habit of not giving by default more privileges than absolutely necessary :)

justinbburris: That makes sense!

justinbburris: I'll update this later tonight, and mark as ready for review - if we have no objection to using the latest build in the website.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,55 @@
...
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
```
oleg-derevenetz, line 21: It is worth to specify `shell` explicitly, because, if the shell is unspecified (even if it is `bash` by default), it will run in a [less strict mode](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell) compared to the explicitly specified shell.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/configure-pages@v5
+      - uses: actions/jekyll-build-pages@v1
```
oleg-derevenetz, line 25: Usually, if a step uses an action with a self-explanatory name (like here), we don't add the explicit name for it.

oleg-derevenetz:
> * Tested semi-successfully on iPhone
>   * Game plays but no audio. 🤔

It looks like [this](https://github.com/emscripten-core/emscripten/issues/19763) issue. Browsers usually block audio from an HTML element (e.g. `<video>` or `<audio>` elements) until the user interacts with them somehow. It looks like the audio continued to be blocked in Safari on iPhone even if user interacted with wasm app using touch. IIRC audio works just fine on Android in similar case.

.github/workflows/pages.yml
justinbburris: oh nice TIL - thank you!

.github/workflows/pages.yml
```diff
@@ -0,0 +1,59 @@
...
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    name: Build website
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
```
justinbburris, line 20: yep good call

.github/workflows/pages.yml
justinbburris, line 1: I wonder if we can implement a linter rule that captures these style items, so you don't have to cleanup after me 😅

docs/deployments/WEB_DEPLOYMENT.md
```diff
@@ -0,0 +1,73 @@
...
+    PagesAction->>GitHubPages: Deploys site
+    EndUser->>GitHubPages: Plays game via browser
+```
+
+## Deployment Pipeline Overview
+
+### Trigger
+
+* Deployment is triggered **when a new Git commit is pushed**.
+* A GitHub Action workflow waits for the completion of the build-and-release process (`build.yml` workflow).
```
justinbburris, line 41: Should this be like this?

```suggestion
* A GitHub Action workflow waits for the completion of the build-and-release process (`make.yml` workflow).
```

docs/deployments/WEB_DEPLOYMENT.md
```diff
@@ -0,0 +1,73 @@
...
+
+### make.yml
+
+* Triggered on `push` to branch
+* Builds the application
+* Publishes release assets
+
+### pages.yml
+
+* Triggered on completion of `build.yml`
```
justinbburris, line 65: Similar question as above
```suggestion
* Triggered on completion of `push.yml`
```

justinbburris:
@oleg-derevenetz just 2 questions in the docs - otherwise lookin 😎

docs/deployments/WEB_DEPLOYMENT.md
oleg-derevenetz: Yes, of course, thank you! It's a typo.

docs/deployments/WEB_DEPLOYMENT.md
oleg-derevenetz: Yes, but `make.yml`.`push.yml` is an enclosing workflow, it doesn't complete yet when `pages.yml` is triggered.

ihhub:
Hi @justinbburris  and @oleg-derevenetz , the changes look good to me. However, the description of this pull request states that we will have a link for users while in reality we don't. So, nobody would know that there is a way to even play onlline. It is intentional?

Second problem which should be addressed in a separate pull request is the visual look of the launcher. It's okaish for now but definitely different from the rest of the website. It is good to have a consistent design.

oleg-derevenetz:
Hi @ihhub I suppose that this is a preliminary pull request to make some infrastructure for further launcher development. That's why there is no link to it so far from the main website.

ihhub:
> Hi @ihhub I suppose that this is a preliminary pull request to make some infrastructure for further launcher development. That's why there is no link to it so far from the main website. @justinbburris am I right?

This works for me.

justinbburris:
@ihhub

@oleg-derevenetz is spot on that the primary purpose of this is to setup the tooling to deploy the site with the launcher. We don't want to link to it, yet, as there are the styling issues you mentioned and the tooling gaps that need to be addressed in future work. 

Additionally this decouples the build risks from a public launch so any kinks, with the build/release tooling can be worked out, without the stress that people might be relying upon this online version to play. 

Finally this unblocks the exploration of how to perform translation of the docs site, as I suspect we'll need this kind of build process to accomplish that. 

Sorry I didn't include these thoughts in the description!

ihhub:
Thank you so much for your hard work, @justinbburris ! Looking forward for more great things to happen!

justinbburris:
@oleg-derevenetz thank you for the through investigation. This could also potentially present an issue, as the branch deployment would temporarily remove the `play-online` route url, while the Github Actions deployment waits to build.

ihhub:
Hi @oleg-derevenetz , now it is done:

IMG_4966

Issue #9805, opened 2025-05-12
Inconsistencies in bad luck between OG and fh2


zenseii:
### Preliminary checks

- [x] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.
- [x] I've checked that this issue is reproducible on the [**latest snapshot build**](https://ihhub.github.io/fheroes2/INSTALL.html#snapshots-latest-builds).

### Platform

Other

### Describe the bug

Reported by a user. See videos below.
* Bad luck animation in fheroes2 is different
* Bad luck should trigger more often (?)
* Bad luck in fheroes2 can do 0 damage (correct?)

fheroes2:
https://www.youtube.com/watch?v=cnW1NJxflls

OG:
https://www.youtube.com/watch?v=JjNLN4NhOIs

### Save file

None provided. Bad luck can only be had through visiting cleared pyramids iirc.

usidedown:
Regarding 0 damage:

Looks like the culprit is `Battle::Unit::GetDamage`.
`Battle::Unit::CalculateDamageUnit` makes sure the minimum damage is 1, but GetDamage does not.

```c++
...
  else if ( Modes( LUCK_BAD ) ) {
      res /= 2;
  }
  
  return res;;
}
```
If `res` is already 1, then it'll become 0.
Maybe final line should be changed to
```c++
return max(res, 1);
```

Regarding luck probability:

Luck appears to be decided by `Battle::Unit::SetRandomLuck`
```c++
void Battle::Unit::SetRandomLuck( Rand::DeterministicRandomGenerator & randomGenerator )
{
    const int32_t luck = GetLuck();
    const int32_t chance = static_cast<int32_t>( randomGenerator.Get( 1, 24 ) );

    if ( luck > 0 && chance <= luck ) {
        SetModes( LUCK_GOOD );
    }
    else if ( luck < 0 && chance <= -luck ) {
        SetModes( LUCK_BAD );
    }

    // Bless, Curse and Luck do stack
}
```
if randomGenerator is uniform, for maximum luck of 3 (or minimum bad luck -3) we get 12.5% for it to trigger.
I am not sure what is the desired behavior.

oleg-derevenetz:
[...]

PR #9819, opened 2025-05-18
Do not allow units to cause zero damage in case of bad luck event

oleg-derevenetz:
Related to #9805

Also this PR adds the implementation of the `checkedCast()` for floating-point -> integer conversions.


ihhub (approved):

ihhub:

PR #9820, opened 2025-05-18
Cache resource language calculations to improve the performance

ihhub:
Reading resources from an AGG file and calculating CRC32 value from it is not an instance operation. Also, since we want to minimize I/O operations from file storage it is important to cache it. This should improve the startup time of the application as well as FPS value for the Select Game Language dialog.

Just small facts:
- the engine startup calls `getResourceLanguage()` function **5** times
- the Select Game Language dialog needs wooping **52** calls of the same function

### DIFF
--- a/src/fheroes2/gui/ui_language.cpp
+++ b/src/fheroes2/gui/ui_language.cpp
@@ -25,6 +25,7 @@
 #include <cstdint>
 #include <functional>
 #include <map>
+#include <optional>
 #include <set>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


ihhub:
Hi @zenseii , could you please test the performance on PS Vita with these changes for the Select Game Language dialog whether it is any improvement?

Districh-ru (approved):

ihhub:
> Much better. Tested at 640x480 with soft cursor off and got 20 fps now.

Nice. Thank you for the check! If you remember what was the performance before this change?

ihhub:
@zenseii , yay! So, for this release it is a 5x improvement.

PR #9824, opened 2025-05-19
Clean Editor-related code: remove the unused mine color property

Districh-ru:
Relates to #9766 

This PR removes the specially coded color for mines that assumed to have many map objects in the database for each player color.
The color for mines is now set in `CapturableObjectMetadata`.

### DIFF
--- a/src/fheroes2/dialog/dialog_selectitems.cpp
+++ b/src/fheroes2/dialog/dialog_selectitems.cpp
@@ -1446,8 +1446,6 @@ int32_t Dialog::selectMineType( const int32_t type )
     int32_t offsetY = area.y + 10;
     text.draw( area.x + ( area.width - text.width() ) / 2, offsetY, display );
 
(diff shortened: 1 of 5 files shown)


ihhub (approved):

oleg-derevenetz:
Hi @Districh-ru could you please take a look at this bug found by Sonar:

https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=ihhub_fheroes2&open=AZbtyXKnZ3PbnynhlS6z

? It seems that after this PR the same value is returned regardless of the condition. Either this condition is no longer needed, or there is a typo somewhere.

Districh-ru:
Hi @oleg-derevenetz, thanks for pointing this out!
This condition is not needed any more after we removed the planned "color" from mines objects database.
PR to fix it is opened:  #9831

PR #9829, opened 2025-05-20

oleg-derevenetz:
* SFINAE: `std::numeric_limits<T>::max_exponent` is an [exclusive upper-bound](https://en.cppreference.com/w/cpp/types/numeric_limits/max_exponent) of the exponent and not an inclusive one (the maximum allowed exponent for a finite value is `max_exponent - 1` and not `max_exponent` itself);
* Deduplicated a few lines of code by simplification of the `if constexpr` in both `checkedCast()` implementations;
* Added `const` according to the Sonar hint;
* Merged a few `if` statements according to the Sonar hint.


ihhub (approved):

ihhub:

PR #9830, opened 2025-05-20
Only run GitHub Pages workflow in the base repo

oleg-derevenetz:
Otherwise this workflow fails on forks (because GitHub Pages are usually not enabled/configured on them).

Also this PR simplifies the `pages.yml` workflow a bit by avoiding extra copy operations.


ihhub (approved):

ihhub:

PR #9834, opened 2025-05-21
Minor battlefield cell related code speed-up

Districh-ru:
This PR:
- inlines `isNearIndexes()` and `isValidIndex()` methods;
- adds `Battle::Board::GetCell( const int32_t position )` not to use `Battle::Cell * Battle::Board::GetCell( const int32_t position, const int dir = CENTER )` because for this case we don't need to check direction and to get index by direction;
- uses `operator[]` instead of `at()` in release version because the index is already checked in the upper lines;
- moves precalculations that are not needed for all `case`s.


ihhub (approved):

ihhub:
@Districh-ru , thank you so much for these improvements!

PR #9835, opened 2025-05-21
Tune the probability of the Good Luck and Bad Luck events

### LINKED ISSUE #9805: Inconsistencies in bad luck between OG and fh2

oleg-derevenetz:
close #9805

I did not find a specific formula for the probability of luck events for HoMM2, so I used the formula for HoMM1 from this description:

https://handbookhmm.ru/1-moral-luck-ppb

According to it, the probability of good luck event (if luck is positive) is `R % 12 < L`, where `R` is some random number and `L` is a luck of the unit. If `R` is always in range `[1..12]`, then it means than the probability is `L / 12`. The same probability is for a bad luck event (if `L` is negative): `-L / 12` (if we take the version of the formula with the OG bug fixed, see the original article).

In the `master` branch the probability is `L / 24` and `-L / 24` respectively. This PR changes the probabilities to `|L| / 12`, but honestly I'm not so sure that this is the correct probability. For some reason, it seems to me that at least bad luck event in HoMM2 is much less common.

### DIFF
--- a/src/fheroes2/battle/battle_troop.cpp
+++ b/src/fheroes2/battle/battle_troop.cpp
@@ -318,7 +318,7 @@ void Battle::Unit::SetRandomMorale( Rand::DeterministicRandomGenerator & randomG
 void Battle::Unit::SetRandomLuck( Rand::DeterministicRandomGenerator & randomGenerator )
 {
     const int32_t luck = GetLuck();
-    const int32_t chance = static_cast<int32_t>( randomGenerator.Get( 1, 24 ) );
+    const int32_t chance = static_cast<int32_t>( randomGenerator.Get( 1, 12 ) );
 
     if ( luck > 0 && chance <= luck ) {
         SetModes( LUCK_GOOD );


oleg-derevenetz:
> IMHO the current formula is correct.

So the question is, which formula should be used - from HoMM1 or HoMM3 (since we don't have an original formula for HoMM2)? @zenseii in #9805 states that

> Bad luck should trigger more often

Let's hear other opinions.

usidedown:
> Hi everyone.
> 
> From my tests, bad luck (at -2) triggered less often than good luck, and iirc bad morale has higher chance of triggering than good morale. This logic might be worth using for bad luck too. This sort of makes the pyramid into a high risk, high reward sort of object.

Bad luck (-2) triggers less often than luck (+2)?
Does Bad Luck (-1) trigger at all? Maybe HOMM2 has the exact same bug as HOMM1.
This indicates that bad luck triggers even less often than in fheroes2...

usidedown:
> Bad luck (-1) would have to be tested by combining a pyramid and a luck artifact.

I took the save files fro
[Before bad luck 01.zip](https://github.com/user-attachments/files/20374470/Before.bad.luck.01.zip)
m #9805 and added luck skill to all heroes.

The ensuing fight is (-1) luck vs. (1) luck. I let it run on auto. 
(1) luck does trigger.
(-1) luck does not trigger

oleg-derevenetz:
@zenseii @usidedown How about the following then:

* We close this PR (without merging) as well as #9805, effectively leaving the luck event probabilities as is (`|L| / 24`);
* When we have some specific probability numbers that we can focus on (taken from some kind of documentation or some reliable statistical research), then will we open a separate issue?

oleg-derevenetz:
OK, closing this PR then.

PR #9836, opened 2025-05-21
Use the double-precision floating-point values for unit threat evaluation

oleg-derevenetz:
After implementing the `checkedCast()` for floating-point -> integer conversions in #9819 it became obvious that unit threat evaluation logic with conversions from `double` to `int32_t` wasn't a good idea. For instance, although this battle works just fine in this PR:

https://github.com/user-attachments/assets/28f835a3-3441-4ab2-a67a-969e3e7f5705

it crashes in the `master` branch due to `std::bad_optional_access` exception here:

https://github.com/ihhub/fheroes2/blob/4d3b56f8bde4d54335cfb2febfa9c6444ea85a17/src/fheroes2/battle/battle_troop.cpp#L1160

since the result of the `attackerThreat * 100` expression doesn't fit into `int32_t`, so in this PR I decided to remove all these extra `double` -> `int32_t` -> `double` conversions (since most of the threat evaluation code in the `AI::BattlePlanner` already works with `double` values), and just use `double` everywhere.

### DIFF
--- a/src/fheroes2/ai/ai_battle.cpp
+++ b/src/fheroes2/ai/ai_battle.cpp
@@ -23,11 +23,11 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
-#include <cfloat>
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <limits>
 #include <map>
 #include <numeric>
 #include <optional>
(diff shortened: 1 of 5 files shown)


ihhub (approved):

ihhub:
Awesome work, @oleg-derevenetz !

PR #9838, opened 2025-05-22
AI: move the guardian spell cast logic from the AI planner to the anonymous namespace

oleg-derevenetz:
`AI::Planner` class is not quite a suitable place for this logic, since this logic has nothing to do with _planning_ 🙂 and is not used somewhere outside of the AI code. Since it is used only by `ai_hero_action.cpp`, I moved the corresponding function to the anonymous namespace in that file.


ihhub (approved):

ihhub:
@oleg-derevenetz , thanks a lot for the change!

PR #9778, opened 2025-05-03
Fix map center positioning when AI interacts with action object

Districh-ru:
fix #4141, fix #5146

This PR adds extra centering map on the hero when he interacts with an action object:

https://github.com/user-attachments/assets/348e11c7-9d79-4796-9f0f-e50e0deadb36

https://github.com/user-attachments/assets/2184e9d7-e684-48bf-950e-0b24abe06396

PR #9842, opened 2025-05-24
Compilation fix

ihhub:
regression from #9778

Thanks @usidedown for spotting it!

### DIFF
--- a/src/fheroes2/ai/ai_hero_action.cpp
+++ b/src/fheroes2/ai/ai_hero_action.cpp
@@ -1917,11 +1917,11 @@ void AI::HeroesAction( Heroes & hero, const int32_t dst_index )
     const bool isActionObject = MP2::isInGameActionObject( objectType, hero.isShipMaster() );
     if ( isActionObject ) {
         hero.SetModes( Heroes::ACTION );
-    }
 
-    if ( AIIsShowAnimationForHero( hero, AIGetAllianceColors() ) ) {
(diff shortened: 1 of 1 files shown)


oleg-derevenetz:
Hi @ihhub apparently we have to enable this setting:

latest_code

I believe that this problem arose due to the fact that although after some changes in the `master` the merging of a new pull request was technically conflict-free, but something shifted there.

oleg-derevenetz:
Hi @ihhub

> doesn't this feature require that every branch must be rebased before being merged?

Yes, and the idea is that all the necessary checks should pass on that rebased branch. The situation when the merge occurs cleanly but generates non-buildable code will be detected before merging.

ihhub:
Hi @oleg-derevenetz ,

Such an approach is inconvenient when we need to merge multiple PRs in a row.

I think here we are trying to solve a very rare case. We literally had 5-10 cases (essentially 5-10 PRs) with compilation issues after merging them. If we compare this number with the number of merged PRs then these are very rare cases for us.

oleg-derevenetz:
@ihhub

> Such an approach is inconvenient when we need to merge multiple PRs in a row.

There is also a merge queue mechanism, but it is available only in repositories [owned by organizations](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

Issue #3411, opened 2021-05-15
Combat, wrong positioning of the Hydras to attack the Ogre Lord (0.9.3 3370)


LeHerosInconnu:
In this attack configuration, Hydras' troop must move two hexes, thus positioning on the two hexes of the Wolf's corpse, to attack the Ogre Lord.

In fheroes2:

![Hydra wrong move fh2 01](https://user-images.githubusercontent.com/43583994/118363157-507b2600-b593-11eb-96c1-4a896426ba2b.png)

https://user-images.githubusercontent.com/43583994/118363164-596bf780-b593-11eb-91f0-0b22d1b74a16.mp4

Kastore attacks Maximus.

Here is the save file:
[Before hydra attack position.zip](https://github.com/ihhub/fheroes2/files/6487485/Before.hydra.attack.position.zip)

oleg-derevenetz:
Hi @LeHerosInconnu,

> In this attack configuration, Hydras' troop must move two hexes, thus positioning on the two hexes of the Wolf's corpse, to attack the Ogre Lord.

Unlike original game, where "main" cell of a wide creature is a tail cell, in fheroes2 "main" cell is a head cell. That's why this difference in mechanics.

LeHerosInconnu:
@oleg-derevenetz.

> Hi @LeHerosInconnu,
> 
> > In this attack configuration, Hydras' troop must move two hexes, thus positioning on the two hexes of the Wolf's corpse, to attack the Ogre Lord.
> 
> Unlike original game, where "main" cell of a wide creature is a tail cell, in fheroes2 "main" cell is a head cell. That's why this difference in mechanics.

It's annoying, the Hydras should advance as far as possible before turning to attack.
This can make the difference between winning and losing a combat. :)
Players should have the same feeling, so with the same behavior, of the original game.

oleg-derevenetz:
It's in fact the same as in #495 and #501, caused by difference in mechanics between OG and fheroes2: different "main cell" of wide creatures and different attack vector (while in OG cell is attacked "from a direction", in fheroes2 "one cell attacks another cell").

oleg-derevenetz:
Implemented in #9847.

PR #9847, opened 2025-05-26
Support 8-directions of attack for wide creatures

usidedown:
Solves #3411 

Adds mouse cursor for top/bottom attacks:
vlcsnap-2025-05-26-08h54m27s175
vlcsnap-2025-05-26-08h53m34s489

Adds attacks originating from "tail" direction:
vlcsnap-2025-05-26-08h54m32s402


usidedown:
> Hi @usidedown what worries me here is that two concepts are mixed up here: the concept of "vertical attack" in the battle UI (which is OK) and the concept of the "vertical direction of the neighboring cell", which doesn't really exist, because, if you take a look at the battlefield, there is in fact no neighboring cells with a strictly vertical direction. For instance, the `Battle::Board::GetIndexDirection()` will not be able to return anything sane for such a direction. We IMHO need to separate these concepts somehow, rather than create some kind of "fake" cell directions that nothing corresponds to.

Yes you're right, attack directions and cell directions are mixed all around. I'll take a look.



> Also I suppose it is impossible to use the [swipe mechanics](https://github.com/ihhub/fheroes2/pull/8464) (intended for relatively small touchscreen devices) with these additional directions, right?

Right again! You can attack in 6 directions using swipe. Top and bottom require a little bit more finesse. I'll see what I can do.

usidedown:
@oleg-derevenetz I introduced CellDirection and AttackDirection types. They are similar but CellDirection includes only the 6 directions whereas AttackDirections allows for Top and Bottom attacks.
Let me know what you think.

Swipe controls currently have trouble with Top/Bottom because they are hex based. Fixing this would require a fundamental change in the way swipe works. I think it should be done in a separate PR.

src/fheroes2/battle/battle_cell.h
```diff
@@ -49,6 +49,75 @@ namespace Battle
...
+
+    inline bool isRightSide( const CellDirection d )
+    {
+        return static_cast<int>( d ) & static_cast<int>( CellDirection::RIGHT_SIDE );
+    }
+
+    inline bool isAround( const CellDirection d )
+    {
+        return static_cast<int>( d ) & static_cast<int>( CellDirection::AROUND );
+    }
```
oleg-derevenetz, line 75: Is it really used somewhere? I understand that it is not used, and it can probably be deleted?

oleg-derevenetz (commented):
Hi @usidedown Overall it looks good. I have made some minimal corrections (such as fixing typos, adding assertions or adding `const` to arguments in function prototypes - I'm aware that this doesn't do anything in prototypes from the point of view of mangling, but in this project it is customary to indicate the constness in prototypes).

I also asked a few questions, could you please take a look?

> Swipe controls currently have trouble with Top/Bottom because they are hex based. Fixing this would require a fundamental change in the way swipe works. I think it should be done in a separate PR.

OK. It may not be possible to do this in a convenient way at all, because the hexes are relatively large and the space to indicate the top/bottom attack is relatively small.

usidedown:
@oleg-derevenetz Thank you so much for your fixes!

> OK. It may not be possible to do this in a convenient way at all, because the hexes are relatively large and the space to indicate the top/bottom attack is relatively small.

I'm still thinking about it but instead of using source and destination Hexes we can use pixel coordinates.
A source + dest points give a vector. It can then be compared to the 8 directions. This way I think it's possible to create 8 equal directions.
It's all very preliminary though.

src/fheroes2/battle/battle_cell.cpp
```diff
@@ -276,68 +276,72 @@ void Battle::Cell::SetArea( const fheroes2::Rect & area )
...
-            return tail && tail->isPassable( true ) && isPassable( true );
-        }
+    if ( dir == CellDirection::LEFT || dir == CellDirection::RIGHT ) {
+        return isPassable( true ) || _index == unit.GetTailIndex();
+    }
 
-        case LEFT:
-        case RIGHT:
-            return isPassable( true ) || _index == unit.GetTailIndex();
+    const CellDirection sideDir = ( dir == CellDirection::TOP_LEFT || dir == CellDirection::BOTTOM_LEFT ) ? CellDirection::LEFT : CellDirection::RIGHT;
```
oleg-derevenetz, line 338: It seems to me that something is wrong here. In the `master` branch:

```cpp
const bool reflect = ( ( BOTTOM_LEFT | TOP_LEFT ) & dir ) != 0;
const Cell * tail = Board::GetCell( _index, reflect ? RIGHT : LEFT );
```

So if `dir` is `TOP_LEFT` or `BOTTOM_LEFT`, then the `sideDir` (or maybe the better name is `tailDir`?) should be `RIGHT`, and not `LEFT`, right?

IIRC the logic here is as follows: if the wide unit wants to perform a diagonal move to the left side, then its tail should be on the right (because the unit can't move "tail first", right?) and, accordingly, vice versa - when moving diagonally to the right, the tail should be on the left.

usidedown:
> Hi @usidedown I just made a few more changes, mainly related to improving the clarity of the code by using more "descriptive" names and comments, and by streamlining the logic a bit in a few places. Could you please take a look when you have time?

@oleg-derevenetz Thank you so much for your improvements!
Looks good to me.

Districh-ru (commented):
Hi @usidedown, I did a quick test and found a case for wide creature with 2-cells attack: the creature attack is showed by hex highlighting in a reverse way (from right to left), the damage is applied also in this way while the attack animation is shown for non-reversed attack (from left to right).

https://github.com/user-attachments/assets/739a1fd9-9075-4584-93b4-0b48bb022f54

Could you please check it when you have time.

And also a wishlist for the future (not this PR :) ): It'll be good if we can choose the direction of the 2-cell up/down attack: to attack the upper(lower) cell and left _or_ right cell.

usidedown:
Hi @Districh-ru.
Thank you for taking the time to test this!

I'm afraid I didn't understand the problem, the behavior I see in the video isn't changed from the master branch. The only difference is the cursor which is now pointing upwards instead of top-left.

You can see that with the (new) tail attack the behavior is also as expected:

https://github.com/user-attachments/assets/446ad37d-a904-458d-98c5-4aa653bb3f90

Districh-ru:
@usidedown, in the master branch the Phoenix flies to the cell (form left to right), then reverses (turns its hear from right to left) and performs diagonal attack from left to right.
In this PR Phoenix does not turn around before attacking so by the animation he performs diagonal attack from left to right (burns the bushes) while the damage is applied to the Archers that are not under the fire of Phoenix (by the shown animation).

usidedown:
@Districh-ru 
Perhaps I am missing something? This is the behavior I get with 1.1.8 release:

https://github.com/user-attachments/assets/95d16807-0216-4425-b558-04552eb5da10

usidedown:
@Districh-ru thank you so much for taking the time to review this

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -3363,51 +3464,105 @@ void Battle::Interface::MouseLeftClickBoardAction( const int themes, const Cell
...
         case Cursor::SWORD_BOTTOMLEFT:
         case Cursor::SWORD_LEFT: {
+            const Unit * unitOnCell = cell.GetUnit();
+            if ( !unitOnCell ) {
+                break;
+            }
+
             if ( !isConfirmed ) {
                 break;
             }
```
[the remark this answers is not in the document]
oleg-derevenetz: Hi @Districh-ru These are my edits, and yes, it's probably better this way. I'll change it now.

Districh-ru (approved):

ihhub:
Hi @usidedown , could you please resolve merge conflicts for this pull request?

src/fheroes2/battle/battle_cell.cpp
```diff
@@ -276,68 +276,74 @@ void Battle::Cell::SetArea( const fheroes2::Rect & area )
...
 
-Battle::CellDirection Battle::Cell::GetTriangleDirection( const fheroes2::Point & dst ) const
+Battle::AttackDirection Battle::Cell::GetTriangleDirection( const fheroes2::Point & dst ) const
 {
     const fheroes2::Point pt( infl * dst.x, infl * dst.y );
 
+    const fheroes2::Point one_half( ( _coord[1].x + _coord[2].x ) / 2, ( _coord[1].y + _coord[2].y ) / 2 );
+    const fheroes2::Point two_half( ( _coord[2].x + _coord[3].x ) / 2, ( _coord[2].y + _coord[3].y ) / 2 );
+    const fheroes2::Point four_half( ( _coord[4].x + _coord[5].x ) / 2, ( _coord[4].y + _coord[5].y ) / 2 );
+    const fheroes2::Point five_half( ( _coord[5].x + _coord[6].x ) / 2, ( _coord[5].y + _coord[6].y ) / 2 );
```
ihhub: Please use camelCase for variable naming.

src/fheroes2/battle/battle_cell.cpp
```diff
@@ -276,68 +276,74 @@ void Battle::Cell::SetArea( const fheroes2::Rect & area )
...
+    assert( tailDir == CellDirection::RIGHT || dir == CellDirection::TOP_RIGHT || dir == CellDirection::BOTTOM_RIGHT );
 
-        default:
-            break;
-        }
-    }
+    const Cell * tail = Board::GetCell( _index, tailDir );
 
-    return isPassable( true );
+    return tail && tail->isPassable( true ) && isPassable( true );
```
ihhub, line 346: Is it correct that we expect that `tail` can be **nullptr**?

ihhub (changes requested):
Hi @usidedown , I left 2 small comments. Would you mind to check them?

src/fheroes2/battle/battle_cell.cpp
usidedown: I can't think of a situation where nullptr is expected, but this check was before I refactored this code.

ihhub: My point is if the tail is always expected it is good to assert it :)

ihhub:
I am moving this pull request to the next release as we made code freeze for any features for the upcoming release.

src/fheroes2/battle/battle_cell.cpp
oleg-derevenetz: Hi @ihhub @usidedown since this method is used to "test hypotheses" during the pathfinding, it can hypothetically be asked, for instance, "can the wide unit move from cell 32 to cell 21?" here:

image

This is a quite legitimate question in general and the answer in this case should be "no" instead of a crash due to failed assertion.

usidedown: @oleg-derevenetz Did you manage to actually get `tail == nullptr` in this case?
I tried moving a champion from 32 to 21 and `tail` was always valid.

oleg-derevenetz: Hi @usidedown

> I tried moving a champion from 32 to 21 and tail was always valid.

In reality, you cannot move a champion from 32 to 21 _directly_ :) Consider (in slow-mo 😂):

https://github.com/user-attachments/assets/e0ad61b3-7f67-46cd-9c2a-dc412597efcd

The champion turns back first (to 31), then it goes to 20, and only then it goes to 21 (by turning back once again). Pathfinder will not let him to just go from 32 to 21 in one step. Technically, champion makes one move in terms of spent "movement points", but in fact he makes several moves. It's just that some of them are "free", but they still have to be made.

oleg-derevenetz: I just want to say that functions, the whole point of which is to perform some check, should simply answer "no" if the check fails. They can apply some preconditions to their parameters in form of assertions, but if their parameters are generally adequate (and in the example above they definitely are), these functions should not crash.

usidedown: @oleg-derevenetz 
Everything you say makes sense but I put the assertion there and couldn't get it to trigger 🤷

oleg-derevenetz: > Everything you say makes sense but I put the assertion there and couldn't get it to trigger 🤷

You cannot trigger it because there are no conditions for this - noone asks for ability to move this way so far :) But you can trigger it manually. I added this assertion here:

```cpp
assert( tail != nullptr );
```

and then added the following code in the `Battle::Arena::ApplyActionMove()`:

```cpp
    Unit * unit = GetTroopUID( uid ); // This line already was here

    if ( unit->isWide() ) {
        Board::GetCell( 21 )->isPassableFromAdjacent( *unit, *Board::GetCell( 32 ) );
    }
```

and voila, this assertion is triggered when I try to move my only Champion in the Battle Only mode. But if you swap 21 and 32 in the line above, it will never be triggered.

usidedown:
Hi @ihhub, could you take a look at this PR when you have the time?
Is there more work to be done here?

ihhub (approved):

ihhub:
@usidedown , huge thanks for this feature!

zenseii:
@usidedown and @oleg-derevenetz a user reported a bug that 2 hex creatures can't attack in the down left direction if a creature is standing to their left. The user says it's version 1.1.9 but I suppose it's a snapshot build from after this PR got merged because you can see a sword pointing straight downwards (?).

https://youtu.be/ZBNs2IPo1zc?si=4Kp0lspH9ZMZYVus

No save is provided but I can request one if needed.

Edit: They've confirmed that this is a snapshot build.

oleg-derevenetz:
> a user reported a bug that 2 hex creatures can't attack in the down left direction if a creature is standing to their left. The user says it's version 1.1.9 but I suppose it's a snapshot build from after this PR got merged because you can see a sword pointing straight downwards (?).
> 
> https://youtu.be/ZBNs2IPo1zc?si=4Kp0lspH9ZMZYVus
> 
> No save is provided but I can request one if needed.
> 
> Edit: They've confirmed that this is a snapshot build.

Yeah, there would be an assertion failure in debug build. In regular build, the attack just won't work :)

PR #9850, opened 2025-05-27
Reduce map reloads while erasing objects or placing terrain in editor

Districh-ru:
This PR speed-ups the Editor by removing the need of full map reload after placing terrain on the map or removing objects by the Eraser tool.


ihhub:
@Districh-ru , many thanks for these changes!

Issue #344, opened 2020-02-01
Window doesn't scroll near minimap


Branikolog:
Cannot scroll window up near minimap in the corner. (In a window mode)
![Безымянный](https://user-images.githubusercontent.com/55348946/73589675-30307080-44e2-11ea-8496-f04546ba9314.jpg)

ihhub:
It's a limitation of SDL. Putting this issue to low priority

PR #9852, opened 2025-05-28
Fix scroll cursor ("arrow") position on screen edges for software rendered cursor

Districh-ru:
Relates to #344

This PR fixes "arrow cursor" the going out of screen bounds when software cursor rendering is on. For this case we can update the position where the cursor is going to be rendered.

img

For the hardware case it relies on OS cursor renderer. One of the ways to fix the issue is to update the cursor image center but it leads to many computations including that in this case the cursor is rendered as is and the game screen can be scaled by SDL.

This PR also slightly optimizes calculations in `Display::render()` method and `getActiveArea()` function.


ihhub (approved):

ihhub:

Issue #1228, opened 2020-08-15
Position of the program window in the workspace


LeHerosInconnu:
With the original program, in windowed mode, the position of the program window in the workspace is remembered when the user quits the program and relaunches it later.

![Position fenetre 01 recadré](https://user-images.githubusercontent.com/43583994/90309713-2b199880-deeb-11ea-8b7d-4f6eecc94cfe.png)

With fheroes2, this position is not memorized.

![Position fenetre 02 recadré](https://user-images.githubusercontent.com/43583994/90309724-44bae000-deeb-11ea-943c-42ba2235626a.png)

celliott113:
We would need to setup logic to allow for the window to save it's current position before closing. Something in the form of when the program is about to close, it will save it's current window position to a file or a setting storage.

When the program is started back up, it will read from that file or setting storage and set the window to that previous position. Will look into this. I can see that each time the window is opened, it opens in the center of the users primary display.

ihhub:
Hi @celliott113 ,
This can be done via an SDL call. See `SDL_GetWindowPosition()` and `SDL_SetWindowPosition` functions. `Settings` class is responsible for settings storage.

PR #9853, opened 2025-05-28
feat: remember position of game window

usidedown:
Solves #1228 

Saves the position of the window in fheroes2.conf.
The position is captured every time the window is moved, and saved to settings when the game exits.
It's possible to capture only when the game exits, but then we might not save the correct position if the user swaps to Fullscreen mode before quitting.

### DIFF
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -971,10 +971,9 @@ namespace
             if ( _window == nullptr ) {
                 return { -1, -1 };
             }
-            int x = 0;
-            int y = 0;
-            SDL_GetWindowPosition( _window, &x, &y );
-            return { x, y };
+            fheroes2::Point result;
+            SDL_GetWindowPosition( _window, &result.x, &result.y );
+            return result;
         }
 
         void setWindowPos( fheroes2::Point pos ) override
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -960,6 +960,27 @@ namespace
             _toggleVSync();
         }
 
+        fheroes2::Point getWindowPos() const override
+        {
+            if ( _window == nullptr ) {
+                return { -1, -1 };
+            }
+            int x = 0;
+            int y = 0;
+            SDL_GetWindowPosition( _window, &x, &y );
+            return { x, y };
+        }
+
+        void setWindowPos( fheroes2::Point pos ) override
+        {
+            if ( pos == fheroes2::Point{ -1, -1 } ) {
+                _prevWindowPos = { SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED };
+            }
+            else {
+                _prevWindowPos = pos;
+            }
+        }
+
     private:
         SDL_Window * _window{ nullptr };
         SDL_Surface * _surface{ nullptr };
@@ -1361,6 +1382,11 @@ namespace fheroes2
         Image::reset();
 
         _screenSize = { info.screenWidth, info.screenHeight };
+    }
+
+    void Display::setWindowPos( Point point )
+    {
+        _engine->setWindowPos( point );
     }
 
     Display & Display::instance()
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -831,7 +831,7 @@ namespace
                 flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 #endif
 
-                // update the window position, in case we get queried for it
+                // Update the window position, in case we get queried for it
                 SDL_GetWindowPosition( _window, &_prevWindowPos.x, &_prevWindowPos.y );
 
                 // If the window size has been manually changed to one that does not match any of the resolutions supported by the display, then after switching
--- a/src/engine/screen.cpp
+++ b/src/engine/screen.cpp
@@ -1139,15 +1139,9 @@ namespace
             }
 
             std::optional<bool> isInDisplay = isWindowInAnyDisplay( resolutionInfo, _prevWindowPos );
-            if ( !isInDisplay.has_value() ) {
-                ERROR_LOG( "Failed to check if the previous window position is within any display." )
 
-                clear();
-                return false;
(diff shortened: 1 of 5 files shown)


src/engine/screen.cpp
```diff
@@ -954,6 +957,30 @@ namespace
...
+        fheroes2::Point getWindowPos() const override
+        {
+            if ( isFullScreen() ) {
+                return _prevWindowPos;
+            }
+            if ( _window == nullptr ) {
+                return { -1, -1 };
+            }
+            int x = 0;
+            int y = 0;
```
ihhub, line 969: `fheroes2::Point` is `std::PointBase2D<int32_t>` so we can use the class directly without the need of additional vaiables.

src/engine/screen.cpp
```diff
@@ -954,6 +957,30 @@ namespace
...
+            SDL_GetWindowPosition( _window, &x, &y );
+            return { x, y };
+        }
+
+        void setWindowPos( fheroes2::Point pos ) override
```
ihhub, line 974: This input argument can be `const`.

src/engine/screen.cpp
```diff
@@ -954,6 +957,30 @@ namespace
...
+        {
+            if ( pos == fheroes2::Point{ -1, -1 } ) {
+                _prevWindowPos = { SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED };
+            }
+            else {
+                _prevWindowPos = pos;
+            }
+        }
```
ihhub, line 982: We must have a check whether the position is within the screen area. Moreover, the position plus the size of the windows must be within the screen.

ihhub (changes requested):
Hi @usidedown , I left few comments in this pull request. Would you mind to take a look at them when you have time?

src/engine/screen.cpp
usidedown: I'm on the fence regarding these checks. They seem to be mostly useful if the user messes with fheroes2.cfg.

My main issue is that when setWindowPos is called _currentScreenResolution isn't set yet. Even if I postpone the check to engine::allocate, _currentScreenResolution is still not set.
Moreover, it's legal to be in a position where pos + size exceeds the screen size. Probably not useful even if you have multiple displays, but still valid.
What do you think?

ihhub:
We will merge this pull request after the upcoming release.

src/engine/screen.cpp
```diff
@@ -825,6 +825,9 @@ namespace
                 flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 #endif
 
+                // update the window position, in case we get queried for it
```
ihhub: [Very minor] Since every other comment starts from a capital letter I suggest to do the same here.

src/engine/screen.cpp
ihhub: Assuming the following situation:
- make the resolution 1920x1080
- move the window to the left part of the display
- change resolution to 640x480

As a result, the whole window is going to be invisible and outside the display. At very least we need to have a mechanism if a user messes up and they would always blame us :)

Or I might be wrong as SDL handles such cases so we don't need to do anything.

src/fheroes2/system/settings.cpp
```diff
@@ -398,6 +403,9 @@ std::string Settings::String() const
     os << std::endl << "# video mode: in-game width x in-game height : on-screen width x on-screen height" << std::endl;
     os << "videomode = " << display.width() << "x" << display.height() << ":" << display.screenSize().width << "x" << display.screenSize().height << std::endl;
 
+    os << std::endl << "# position of the game's window" << std::endl;
+    os << "main window position = [ " << _windowPos.x << ", " << _windowPos.y << " ]" << std::endl;
```
ihhub: Should we name the parameter as `game window position` as well?

src/engine/screen.cpp
usidedown: You're right!
I've implemented the check.
I believe it is better to check when the window is created rather then when the position is set to avoid race conditions.

usidedown:
Hi @ihhub, could you take a look at this PR when you have the time?
Is there more work to be done here?

src/engine/screen.cpp
```diff
@@ -1080,6 +1132,20 @@ namespace
 #endif
             }
 
+            std::optional<bool> isInDisplay = isWindowInAnyDisplay( resolutionInfo, _prevWindowPos );
+            if ( !isInDisplay.has_value() ) {
+                ERROR_LOG( "Failed to check if the previous window position is within any display." )
+
+                clear();
+                return false;
+            }
```
ihhub: I suggest to still fall back to default display ID with no position specified in this case. There is no guarantee that `SDL_GetNumVideoDisplays()` function works on all platforms.

ihhub (approved):

ihhub:

Killbrum:
Hi guys. Just a generic warning. In the past, I already had different cases with such functionality. The end user can mistakenly change the window position to a very pity one (e.g. far corner). Don't ask why, mainly it just happens. Generic user expectation that on the next run - the window will be restored properly.

Let's go through a very generic case
- End user works with multiple windows. By mistake fheroes2 window is moved to be bottom corner and the window title is barely visible
- The end user cannot find fheroes2 windows, so the game is forcefully closed 
- On the next run the window is restored to the same position (e.g. not playable)

Typically it's handled by counting window size and in case endpoints (X/Y) have higher values than screen size - restore to the default screen and centered

Another typical case. An end user has 2 monitors (e.g. connected TV). On the next run, the TV (or any other second monitor) is disconnected. What will happen then? Was that test scenario checked?

usidedown:
Hi @Killbrum, thanks for your input!

We ended up with the following:
- The window is only saved on a proper exit.
- The game checks that the window is fully contained in the display, otherwise the default position is used
- If we failed to get the display information, the default position is used.

I believe the behavior is correct in the scenarios you outlined.

PR #9854, opened 2025-05-28
Inline some methods in image.h ans some `fheroes2::Sprite`-related code clean-up

Districh-ru:
This PR:
- inlines some short methods ~and functions~ in `image.h`;
- adds the `Sprite( Image && image, const int32_t x = 0, const int32_t y = 0 )` constructor;
- ~uses `std::move()` instead of `std::swap` in assignment constructor and operator for `Image` and `Sprite` classes~.

It also removes the unused function `void ApplyAlpha( const Image & in, Image & out, const uint8_t alpha )`.


src/engine/image.cpp
```diff
@@ -607,8 +566,8 @@ namespace fheroes2
 
         Image::operator=( std::move( sprite ) );
 
-        std::swap( _x, sprite._x );
-        std::swap( _y, sprite._y );
+        _x = sprite._x;
```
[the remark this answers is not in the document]
Districh-ru: False positive issue: `_x` and `_y` of `Sprite` are not moved while "moving" to `Image` operator.

src/engine/image.cpp
```diff
@@ -440,45 +440,27 @@ namespace
...
         }
 
-        std::swap( _width, image_._width );
-        std::swap( _height, image_._height );
-        std::swap( _data, image_._data );
-        std::swap( _singleLayer, image_._singleLayer );
+        _width = image._width;
+        _height = image._height;
+        _data = std::move( image._data );
+        _singleLayer = image._singleLayer;
```
oleg-derevenetz, line 463: Hi @Districh-ru

Personally, I usually use the state exchange via `std::swap()` in such cases, rather than simple assignment, because this leaves the moved-from object in an absolutely valid state, where all values are sane and are in sync. This is not the case in your code: `_data` will become an empty `std::unique_ptr`, but `_width` or `_height` will be non-zero.

Since there is no "destructive move" in C++, there is a non-zero probability that the object will be (accidentally?) reused after being "moved from". In the case of a complete exchange of states, it is quite safe. In your case, it's not safe, the most that can be done with the object is to destroy it safely. 

Not that this is a big problem in itself, because reuse of a "moved from" object is usually suspicious. But if it is possible to increase the safety of working with such an object for almost free, I usually try to do this :)

Is there (and in other similar places below) any real gain from leaving a "moved from" object in an invalid state instead of a state exchange?

Districh-ru: Hi @oleg-derevenetz,
I just wanted to reduce the `std::move()`s count (one instead of three for `std::swap()`) thinking that the r-value `image` image will never be used.
And if the moved-from object will be accidentally used after the `std::swap()` it will contain not it's previous data but the data from the object it was moved-to. For me it looked a little strange. :)
But thanks to your explanations now I understand why it was done this way.

Can we just set `_width` and `_height` to `0` and `_singleLayer` to `false` for the moved-from object so it will be definitely zero-sized meaning that all data is moved from it?
It will be like after the `Image()` constructor without arguments.

If it is not good I'm fine to revert these changes.

oleg-derevenetz (commented):
Technically looks good to me, I asked just one question, please take a look when you have time.

BTW, here is an article from Herb Sutter regarding `std::move()`:

https://herbsutter.com/2020/02/17/move-simply/

Although there are no specific techniques mentioned, it is mentioned that invariants of the "moved-from" object should hold and such an object should be safe for a "regular use" (not just for a destruction!). Otherwise Herb considered the object implementation as buggy. Personally, I agree with him, although in general, as I already said, it is considered that the reuse of the "moved-from" object is suspicious, and Clang-Tidy has a special warning about this.

src/engine/image.h
```diff
@@ -36,15 +36,23 @@ namespace fheroes2
     {
     public:
         Image() = default;
-        Image( const int32_t width_, const int32_t height_ );
 
-        Image( const Image & image_ );
-        Image( Image && image_ ) noexcept;
+        Image( const int32_t width, const int32_t height )
+        {
+            Image::resize( width, height );
```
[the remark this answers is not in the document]
Districh-ru: As I understand, `Image::` is placed here to explicitly say not to use the overloaded variant of `resize()` virtual method if the `Image()` constructor is called from a derived class.

https://rules.sonarsource.com/cpp/type/Code%20Smell/RSPEC-1699/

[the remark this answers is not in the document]
oleg-derevenetz: Ah, it's virtual, OK then.

src/engine/image.h
```diff
@@ -328,4 +353,52 @@ namespace fheroes2
     void Transpose( const Image & in, Image & out );
 
     void updateShadow( Image & image, const Point & shadowOffset, const uint8_t transformId, const bool connectCorners );
+
+    // The next inlined functions are implemented to simplify calls of some functions declared above.
+
+    // Replace a particular pixel value by transparency value (transform layer value will be 1)
+    inline void AddTransparency( Image & image, const uint8_t valueToReplace )
```
ihhub: These changes breaks the order in which functions were added. I don't think that we need to sacrifice readability of the code for the sake of one additional function call.

Districh-ru: Hi @ihhub,
Sure, I moved `AddTransparency()` to `agg_image.cpp` because it is used only in that code and I put other inlined functions in the alphabetical order.

Also I removed the check 
```cpp
        if ( inPos.x < 0 || inPos.y < 0 ) {
            return;
        }
```
in `void Blit( const Image & in, const Point & inPos, Image & out, const Point & outPos, const Size & size, const bool flip )` because in the called `Blit()` or `Copy()` function such check is performed by:
```CPP
if ( !Verify( in, inX, inY, out, outX, outY, width, height ) ) {
    return;
}
```

So there is no need to duplicate the same check.

src/engine/image.h
```diff
@@ -209,42 +244,75 @@ namespace fheroes2
...
     void ApplyTransform( Image & image, int32_t x, int32_t y, int32_t width, int32_t height, const uint8_t transformId );
 
-    // draw one image onto another
-    void Blit( const Image & in, Image & out, const bool flip = false );
-    void Blit( const Image & in, Image & out, const Rect & outRoi, const bool flip = false );
-    void Blit( const Image & in, Image & out, int32_t outX, int32_t outY, const bool flip = false );
+    // Draw one image on another taking into account the transparency and shadows data in the transform layer.
     void Blit( const Image & in, int32_t inX, int32_t inY, Image & out, int32_t outX, int32_t outY, int32_t width, int32_t height, const bool flip = false );
 
+    inline void Blit( const Image & in, Image & out, const bool flip = false )
```
ihhub: With these changes the code looks harder to read. Previously:
```cpp
    void Blit( const Image & in, Image & out, const bool flip = false );
    void Blit( const Image & in, Image & out, const Rect & outRoi, const bool flip = false );
    void Blit( const Image & in, Image & out, int32_t outX, int32_t outY, const bool flip = false );
```
now:
```cpp
    void Blit( const Image & in, int32_t inX, int32_t inY, Image & out, int32_t outX, int32_t outY, int32_t width, int32_t height, const bool flip = false );

    inline void Blit( const Image & in, Image & out, const bool flip = false )
    {
        Blit( in, 0, 0, out, 0, 0, in.width(), in.height(), flip );
    }

    inline void Blit( const Image & in, Image & out, const Rect & outRoi, const bool flip = false )
    {
        Blit( in, 0, 0, out, outRoi.x, outRoi.y, outRoi.width, outRoi.height, flip );
    }

    inline void Blit( const Image & in, Image & out, const int32_t outX, const int32_t outY, const bool flip = false )
    {
        Blit( in, 0, 0, out, outX, outY, in.width(), in.height(), flip );
    }
```

As you can see previously all variants of the same function can be easily differentiated. Now it is much harder to read because of `inline` keyword.

This is why I pointed out that we are sacrificing code readability for one extra function call. Do we want to go this extra optimization path or keep the code more readable?

Districh-ru: Hi @ihhub,  thanks for detailed explanation about code readability.
I restored the original declaration order of functions in `image.h`.
But moved the `inline` implementations to the end of H file.
What do you think about this?
If it is not good I'll move all implementations to CPP file like it is in the master build.

src/tools/h2dmgr.cpp
```diff
@@ -254,7 +254,10 @@ namespace
                     return EXIT_FAILURE;
                 }
 
-                if ( !fheroes2::writeImageToH2D( writer, std::filesystem::path( inputFileName ).filename().replace_extension( "image" ).string(), image ) ) {
+                // TODO: Add ability to set sprite position and ability to determine colors in original image
+                // that are used for transparency and shadows.
+                if ( !fheroes2::writeImageToH2D( writer, std::filesystem::path( inputFileName ).filename().replace_extension( "image" ).string(),
```
ihhub: Would it be better to make `image` as `fheroes2::Sprite` rather than creating a temporary object here?

src/fheroes2/gui/ui_button.cpp
```diff
@@ -35,21 +35,18 @@
 
 namespace
 {
-    fheroes2::Image resizeButton( const fheroes2::Image & original, const fheroes2::Size & buttonSize )
```
ihhub: The idea behind this function is to return just an image without coordinates. Now a caller of this function might think that it returns some coordinates. Since we already have `explicit Sprite( Image && image, const int32_t x = 0, const int32_t y = 0 )` we have almost no performance drop in returning `Image` and assigning it to `Sprite`.

src/fheroes2/battle/battle_interface.cpp
```diff
@@ -6862,8 +6862,8 @@ void Battle::PopupDamageInfo::_makeDamageImage()
     const fheroes2::Rect borderRect( isLeftSidePopup ? ( x - w - unitRect.width - borderWidth ) : x, y, w, h );
 
     const fheroes2::Sprite & backgroundImage = fheroes2::AGG::GetICN( ICN::CELLWIN, 1 );
-    _damageImage = fheroes2::Stretch( backgroundImage, 0, 0, backgroundImage.width(), backgroundImage.height(), borderRect.width, borderRect.height );
-    _damageImage.setPosition( borderRect.x, borderRect.y );
+    _damageImage = fheroes2::Sprite{ fheroes2::Stretch( backgroundImage, 0, 0, backgroundImage.width(), backgroundImage.height(), borderRect.width, borderRect.height ),
+                                     borderRect.x, borderRect.y };
```
ihhub: We could still use the old code as it differs from the current one just by one function call. The current code has become harder to read as
```cpp
_damageImage = fheroes2::Sprite{ fheroes2::Stretch(...
```
is much difficult to understand and debug later.

src/fheroes2/agg/agg_image.cpp
```diff
@@ -3997,7 +4001,7 @@ namespace
                 // Make a sprite for EDITOR_ANY_ULTIMATE_ARTIFACT used only in Editor for the special victory condition.
                 // A temporary solution: apply the blur effect originally used for the Holy Shout spell and the purple palette.
                 fheroes2::Sprite & targetImage = _icnVsSprite[id][83];
-                targetImage = CreateHolyShoutEffect( _icnVsSprite[id][91], 1, 0 );
+                targetImage = fheroes2::Sprite{ CreateHolyShoutEffect( _icnVsSprite[id][91], 1, 0 ) };
```
ihhub: We don't need to explicitly call a constructor here as `explicit Sprite( Image && image, const int32_t x = 0, const int32_t y = 0 )` does what we need.

This is a comment for all other places where we added `fheroes2::Sprite{}` call.

Unless the code is not compilable I suggest not to do this and rely on a constructor call rather than compilating the existing code.

ihhub (changes requested):
Hi @Districh-ru , I left several more comments in this pull request. Could you please take a look at them and see if they make any sense?

src/tools/h2dmgr.cpp
Districh-ru: Sure, done!

src/fheroes2/gui/ui_button.cpp
Districh-ru: OK, I returned to `Image`.

src/fheroes2/battle/battle_interface.cpp
Districh-ru: Done

src/fheroes2/agg/agg_image.cpp
Districh-ru: I've added `Sprite & operator=( Image && image )` to make code compilable and removed all added `fheroes2::Sprite{}` calls.

ihhub (approved):

src/engine/image.h
ihhub: I think the best case is to keep the code for these functions as before since now we have some functions being placed in one area and then others in another. Ideally a developer should see all functions at once. I don't think that added performance is worth the reduction of code readability. What do you think?

Districh-ru: I agree with your vision, I just tried to find a compromise between readability and performance.
I returned function implementations back to CPP file.

ihhub:
@Districh-ru , awesome work!

PR #9856, opened 2025-05-29
Rework New Game, Load Game and Editor menus

### LINKED ISSUE #7100: Rename CONFIG button into SETTINGS
### Preliminary checks

Have a button with title CONFIG and its description as Game Settings seems not very logical. I understand that CONFIG button is taken from network setup which works well there but not exactly suits Game Settings.

image

Rename the button and possibly adjust the translations.

### Additional info

_No response_

zenseii:
Close #7100 
Part of #9552 and #1893

This PR moves these dialogs over to `StandardWindow` to allow for Evil interface, variable size buttons for translations, and improved shadows. A lot of tidying has been done too.

The rationale for merging all the dialog states into one `StandardWindow` was to remove repeated code and unnecessary render and draw calls. All these separate dialogs were basically doing the same thing, just with different buttons.

This should also make it easier when implementing the original animating background of the lamps in the main menu (they should animate even with dialogs opened).

Battle only is now opened without having to press Hot Seat first, because I didn't see how that is relevant when battle only defaults to player vs ai and there is no distinct plan for network play for battle only.

Notes on some changes/decisions:
* Buttons are generated in one go like `BUTTONS_NEW_GAME_MENU_GOOD` because they need to have the same dimensions.
* Symmetric button rendering was extracted from the constructor to a separate method because it was necessary to have an empty dialog restorer.
* The network-related stuff for the main menu was removed because it would've been more work to change it, and considering that, as far as I know, nothing's been planned regarding the menu design (e.g. should it be divided into Host menu and Guest menu). Either way, adding new buttons later on when network is being worked on will be easy.

image image


Districh-ru:
Hi @zenseii, you did a great work!
I'll check the code a little later (probably tomorrow).
I did a quick test and fund that in File Options dialog the translated buttons are not "evil" when Evil interface is selected:
изображение

src/fheroes2/agg/agg_image.cpp
```diff
@@ -1306,16 +1295,27 @@ namespace
...
+            }
 
-            convertToEvilButtonBackground( _icnVsSprite[id][0], _icnVsSprite[id][1], goodButtonIcnID );
             break;
         }
         case ICN::BUTTONS_FILE_DIALOG_EVIL:
         case ICN::BUTTONS_FILE_DIALOG_GOOD: {
             _icnVsSprite[id].resize( 8 );
 
+            const bool isEvilInterface = id == ICN::BUTTONS_FILE_DIALOG_EVIL;
```
Districh-ru, line 1316: For better readability we can add parentheses:
```suggestion
            const bool isEvilInterface = (id == ICN::BUTTONS_FILE_DIALOG_EVIL);
```

Districh-ru:
As I can see, this PR will close #7100

Districh-ru (changes requested):
Hi @zenseii, all menus works as expected.
I like how you combined the main and editor menu.
I left two comments, could you please take a look in your spare time?

Also what do you think of making the behavior for Campaign and Multi-player menus act like in Editor: to have BACK button instead of CANCEL?
It will be different from the original game but IMHO more convenient: You open the New Game or Load Game menu and CANCEL actually returns you back the Main Menu. So I suggest to do the same action for the other menus - just return back one step.
And, yes, we can do this in a new PR after we discuss the need of such change.

Districh-ru (changes requested):
Hi @zenseii, I rechecked the code and found one more place where we could slightly optimize the code. Could you please take a look when you have time?

ihhub:
I am moving this pull request to the next release.

ihhub:
Hi @zenseii , since we made Interface independent windows in the Main Menu should we add back an option to change Interface Type? Of course it can be done in a separate pull request.

ihhub (approved):

zenseii:
> Hi @zenseii , since we made Interface independent windows in the Main Menu should we add back an option to change Interface Type? Of course it can be done in a separate pull request.

@ihhub, I would definitely like to have the interface option in the main menu settings. I suppose we would need to add an option together with it to match the column and rows? Another PR for it makes sense.

ihhub:
Hi @zenseii , yes we would need to add one more option to expand the window to have 3x3 option grid.

ihhub:

PR #9860, opened 2025-06-01
Remove rumor about latest update

### LINKED ISSUE #2199: Castle screen, Tavern, fheroes2 "rumor"
I find that this "rumor" breaks the immersion in the game.
I think it's okay to display it up to version 1.0 of fheroes2.

However, it would be better if it appears only in scenarios in which the scenario creator has not added rumors in the scenario editor, as this may interfere with some scenarios where rumors are important because they give necessary hints to the player.

I also suggest adding "the" before "game":
"Check the newest version of the game at..."

Instead of:
"Check the newest version of game at..."

In fheroes2:

Tavern fheroes2 rumor 01

zenseii:
This fheroes2-specific rumor makes it less likely for other useful rumors to appear, like those giving the general area of the UA, and this diminishes the original value of the Tavern. It's also more likely that a player will see the download link in the Credits or in the docs than randomly in the Tavern.

Fix #2199

### DIFF
--- a/src/fheroes2/world/world.cpp
+++ b/src/fheroes2/world/world.cpp
@@ -671,7 +671,7 @@ void World::MonthOfMonstersAction( const Monster & mons )
 
 fheroes2::LocalizedString World::getCurrentRumor() const
 {
-    const uint32_t standardRumorCount = 10;
+    const uint32_t standardRumorCount = 9;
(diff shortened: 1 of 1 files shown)


ihhub (approved):

ihhub:

PR #9866, opened 2025-06-03
Detect doubled objects' UIDs in fh2m maps in debug build

Districh-ru:
This PR adds simple checks for repeat of objects' UIDs during `fh2m` map load.
If the repeats are detected the information about this will be output to the console and log file.
The shown information may help to fix the already made maps by the very first versions of editor and to check the editor correctness in the future while testing editor-related PRs.

### DIFF
--- a/src/fheroes2/maps/map_format_helper.cpp
+++ b/src/fheroes2/maps/map_format_helper.cpp
@@ -1213,14 +1213,65 @@ namespace Maps
         auto sortObjects = []( const IndexedObjectInfo & left, const IndexedObjectInfo & right ) { return left.info->id < right.info->id; };
         std::multiset<IndexedObjectInfo, decltype( sortObjects )> sortedObjects( sortObjects );
 
+#if defined( WITH_DEBUG )
+        std::map<uint32_t, IndexedObjectInfo> objectsUIDs;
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @Districh-ru , do you think it is possible to make a correction code for all types of versions (release or debug) when it is obvious to fix the map when it is being open in the Editor? This might solve problems (not all) when broken maps due to our bugs in the Editor in the past.

Districh-ru:
Hi @ihhub, I can try to add a simple code that updates the incorrect UIDs in the case they are not present in any metadata, but I'll remove the `const` from `const Map_Format::MapFormat & map` argument to allow changes in map while loading without using `const_cast` as I did to fix 2 maps.

But this code will slow down map loading (we can measure the loading times for big maps with check and without).
Do we need to do this fix on every map start in custom game or only on map load in Editor?
And since we fixed the 2 found maps with this bug of all maps that we have in `map-sharing` is it really needed to run the check+fix in release version?
The source of this bug was fixed a year ago to I assume that the earlier maps are already published or abandoned...
Then should we do the fix only for old map versions?

I'm not against making a code to fix such maps (we can even check the object type and fix some metadata by type), but I don't like it to run always if it may be mostly unneeded. :)

ihhub:
@Districh-ru , let's got with this solution first.

ihhub:
@Districh-ru , many thanks for the changes!

PR #9870, opened 2025-06-06
Update Russian translation

Districh-ru:
Recently added strings translation.


ihhub:
@Districh-ru , any thanks for these updates!

PR #9872, opened 2025-06-06
Add "7 Deserts" map as a part of the fheroes2 bundle

ihhub:
Expanding the set of user maps.


oleg-derevenetz:
Hi @ihhub is there any way to exclude spaces in the names of map files? You see, some GNU Make constructs (such as `$(patsubst)`) may have issues with files that have spaces in their names.

P.S. Probably, parentheses are also undesirable, because they have a special meaning in `bash`, which is used by GNU Make (and it apparently does not escape special characters when calling shell). However, maybe if the parentheses are not at the beginning of the file name, then it doesn't matter - it needs to be checked separately, but it's better to be safe than sorry.

ihhub:
Hi @oleg-derevenetz , how about now? :)

oleg-derevenetz (approved):
Yes, now it seems OK. I think in the future it will be necessary to produce some kind of guide for the developer of the maps supplied with the engine, where, among other things(?), the naming scheme/requirements for the maps will be described.

oleg-derevenetz:
> Maybe it's intentional, but we don't automatically copy `fh2m` files into the build destination's `maps` folder. Personally I think we should do this, as we do that for original maps (mp2, mx2 etc.).

We copy them:

https://github.com/ihhub/fheroes2/blob/9d6d6b3295074b5ec5cdedac214933121188cb00/VisualStudio/fheroes2/common.props#L22-L26

BUT only if the destination folder does not exist yet. This logic has been here for a very long time and has not changed since then.

oleg-derevenetz:
@zenseii

> Also, the name of the file should use an underscore and not dash: `7_deserts...`

By the way, would you like to write a guide for map developers that will be suitable for delivery as part of the engine (including the map file naming scheme)? Perhaps it would make sense to highlight some other points in this guide that I am not aware of?

zenseii:
> By the way, would you like to write a guide for map developers that will be suitable for delivery as part of the engine (including the map file naming scheme)? Perhaps it would make sense to highlight some other points in this guide that I am not aware of?

We already have a guide on the Discord server, which is the only place where we accept map submissions right now, but we can add the point about the file name limitations.

Btw what happens on Linux systems if you try to save a fh2m file with these characters in its name with the editor?

oleg-derevenetz:
> Btw what happens on Linux systems if you try to save a fh2m file with these characters in its name with the editor?

These limitations are not because of filesystem or engine itself. Engine can work with file names containing spaces, parentheses, etc just fine. These limitations are because of build system that builds the engine and bundles the map files into the auxiliary data files (this is especially related to the WASM/Emscripten build). GNU Make just doesn't like the filenames of this kind.

P.S. There is the corresponding [bug](https://savannah.gnu.org/bugs/?712) at least 23 years old - but this is just a bug from that date, and in fact GNU Make has not worked with such files normally since its birth.

PR #9873, opened 2025-06-06
Fix showing one 1st level troop in hero's army diring hero fade-out animation after loosing the battle

### LINKED ISSUE #9869: Defeated hero has 1 monster of level 1 before disappearing
### Preliminary checks

### Platform

Windows

### Describe the bug

https://github.com/user-attachments/assets/b9fb172e-265d-4367-8428-b47eb301ebff

### Save file

Play 7 Deserts map to check this.

### Additional info

_No response_

Districh-ru:
fix #9869 

The incorrect 1st level troop is shown because before dismissing the defeated hero from the kingdom his army is always reset to one 1st level troop according to hero's race.
This PR removes the reset of the defeated army right after the battle because the reset is also done when dismissing the defeated hero from the kingdom right after the hero fade-out animation.

https://github.com/user-attachments/assets/9bc3e4ab-9dfa-4d69-833e-2ff86ec5f017

This PR also slightly updates the logic for `Troops::isValid()`, `Troops::GetOccupiedSlotCount()` and `Army::drawMultipleMonsterLines()`.

### DIFF
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -346,12 +346,18 @@ double Troops::getReinforcementValue( const Troops & reinforcement ) const
 
 bool Troops::isValid() const
 {
-    return std::any_of( begin(), end(), []( const Troop * troop ) { return ( troop != nullptr ) && troop->isValid(); } );
+    return std::any_of( begin(), end(), []( const Troop * troop ) {
+        assert( troop != nullptr );
+        return troop->isValid();
+    } );
 }
 
 uint32_t Troops::GetOccupiedSlotCount() const
 {
-    return static_cast<uint32_t>( std::count_if( begin(), end(), []( const Troop * troop ) { return ( troop != nullptr ) && troop->isValid(); } ) );
+    return static_cast<uint32_t>( std::count_if( begin(), end(), []( const Troop * troop ) {
+        assert( troop != nullptr );
+        return troop->isValid();
+    } ) );
 }
 
 bool Troops::areAllTroopsUnique() const
--- a/src/fheroes2/army/army.cpp
+++ b/src/fheroes2/army/army.cpp
@@ -346,21 +346,12 @@ double Troops::getReinforcementValue( const Troops & reinforcement ) const
 
 bool Troops::isValid() const
 {
-    for ( const_iterator it = begin(); it != end(); ++it ) {
-        if ( ( *it )->isValid() )
-            return true;
-    }
-    return false;
+    return std::any_of( begin(), end(), []( const Troop * troop ) { return ( troop != nullptr ) && troop->isValid(); } );
 }
 
 uint32_t Troops::GetOccupiedSlotCount() const
 {
-    uint32_t total = 0;
-    for ( const_iterator it = begin(); it != end(); ++it ) {
-        if ( ( *it )->isValid() )
-            ++total;
-    }
-    return total;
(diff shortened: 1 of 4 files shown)


src/fheroes2/battle/battle_main.cpp
```diff
@@ -475,14 +475,6 @@ Battle::Result Battle::Loader( Army & army1, Army & army2, int32_t mapsindex )
     army1.resetInvalidMonsters();
     army2.resetInvalidMonsters();
 
-    // Reset the hero's army to the minimum army if the hero retreated or was defeated
-    if ( commander1 && commander1->isHeroes() && ( !army1.isValid() || ( result.army1 & RESULT_RETREAT ) ) ) {
-        army1.Reset( false );
-    }
-    if ( commander2 && commander2->isHeroes() && ( !army2.isValid() || ( result.army2 & RESULT_RETREAT ) ) ) {
-        army2.Reset( false );
-    }
```
Districh-ru, line 484: Doing this `Reset()` we always set one 1st level troop in the army and the Status area shows it correctly. But there is no need to reset army here because after defeat or retreat `hero.Dismiss()` is always called. And in this method the army is reset once again but to the default state.

src/fheroes2/army/army.cpp
```diff
@@ -346,21 +346,12 @@ double Troops::getReinforcementValue( const Troops & reinforcement ) const
 
 bool Troops::isValid() const
 {
-    for ( const_iterator it = begin(); it != end(); ++it ) {
-        if ( ( *it )->isValid() )
-            return true;
-    }
-    return false;
+    return std::any_of( begin(), end(), []( const Troop * troop ) { return ( troop != nullptr ) && troop->isValid(); } );
```
oleg-derevenetz: As far as I remember, the case when `troop == nullptr` is normally impossible and the corresponding check can be replaced by an assertion.

src/fheroes2/army/army.cpp
```diff
@@ -346,21 +346,12 @@ double Troops::getReinforcementValue( const Troops & reinforcement ) const
...
 
 uint32_t Troops::GetOccupiedSlotCount() const
 {
-    uint32_t total = 0;
-    for ( const_iterator it = begin(); it != end(); ++it ) {
-        if ( ( *it )->isValid() )
-            ++total;
-    }
-    return total;
+    return static_cast<uint32_t>( std::count_if( begin(), end(), []( const Troop * troop ) { return ( troop != nullptr ) && troop->isValid(); } ) );
```
oleg-derevenetz: The `troop != nullptr` check here can also be replaced by an assertion.

src/fheroes2/battle/battle_main.cpp
oleg-derevenetz: As far as I understand, the player will still see the "real" state of the hero's army after the battle, right? Including the case of an empty army.

oleg-derevenetz (changes requested):
Hi @Districh-ru I just left a couple of suggestions and a question, could you please take a look when you have time?

src/fheroes2/battle/battle_main.cpp
Districh-ru: Yes.
When just dismissing a hero the player sees his current army during hero's fade-out.
After surrendering and retreat will be the same - while showing the fade-out animation the current army state (after the battle) will be shown.
When hero is defeated there will be shown only the background (empty army) like on the video.

src/fheroes2/army/army.cpp
Districh-ru: Agree.
Done.

ihhub (approved):

ihhub:

Issue #9867, opened 2025-06-05
Random Number Generation Improvements


usidedown:
### Preliminary checks

- [x] I've checked that there aren't [**other open issues**](https://github.com/ihhub/fheroes2/issues?q=is%3Aissue) on the same topic.

### Describe the problem requiring a solution

Randomness is an important part of the game, and I believe there are 2 aspects that we can improve on:
- Performance - I believe there are low hanging fruits available to make RNG faster and use less memory.
- Reproducibility - It's very nice when you re-load the game and the results remain the same. The game does this very well, unless you switch platforms. This isn't a problem for game play (maybe a little for debugging) but could be a fundamental problem with the editor when it begins generating random maps.

### One-off Random Engine
Let's look at the following code:
```c++
uint32_t Rand::GetWithSeed( uint32_t from, uint32_t to, uint32_t seed )
{
    if ( from > to ) {
        std::swap( from, to );
    }

    std::uniform_int_distribution<uint32_t> distrib( from, to );
    std::mt19937 seededGen( seed );

    return distrib( seededGen );
```

This function from _rand.cpp_ is called many times during a battle, and each time creates its own engine.
Creating this engine is a relatively expensive operation (see below).  Preferably we should re-use the same engine.
We have the similar `Rand::GetWithGen` which does re-use the engine.

There are additional oddities, such as manually keeping the state and re-seeding the engine

```c++
uint32_t Rand::DeterministicRandomGenerator::Get( const uint32_t from, const uint32_t to /* = 0 */ )
{
    ++_currentSeed;
    return Rand::GetWithSeed( from, to, _currentSeed );
}
```
The engine already keeps its state internally. I am not sure what benefit there is to use a seed for each generation.

### Mersenne-Twister
mt19937 is an okay engine...But it's outdated and has some big disadvantages.
For once, its size:
```c++
static_assert(sizeof(std::mt19937) == 5000);
```
The state of std::mt19937 uses 624 values of internal state. That's actually pretty big!
Additionally, mt is considered slow when compared to alternatives.

On the bright side, it is identical on all platforms (see below)

## Reproducibility

Let me start by saying the current code does a great job at reproducibility. A deterministic PRNG is used. Seeds are used and saved. Unfortunately, the c++ standard doesn't guarantee reproducibility between different platforms. And in fact we'll get different numbers on different c++ implementations thanks to `uniform_int_distribution`. 

### One-off Random Engine
This one requires refactoring existing code. No magic here :(.
[...]

PR #9874, opened 2025-06-06
Change RNG algorithms

usidedown:
Solves #9867 

- Introduces PCG engine instead of mt19937 based on pcg-cpp library
This should improve performance.

- Replaces std::shuffle with pcg-cpp:shuffle
- Replace uniform_int_distribution with custom implementation

These should ensure compatibility between platforms

### DIFF
--- a/src/dist/engine/Makefile
+++ b/src/dist/engine/Makefile
@@ -20,6 +20,7 @@
 
 ifndef FHEROES2_WITH_SYSTEM_SMACKER
 CCFLAGS := $(CCFLAGS) -I../../thirdparty/libsmacker
+CCFLAGS := $(CCFLAGS) -I../../thirdparty/pcg-cpp/include
 endif
 
 SOURCEDIR := ../../engine
--- a/src/dist/engine/Makefile
+++ b/src/dist/engine/Makefile
@@ -20,8 +20,9 @@
 
 ifndef FHEROES2_WITH_SYSTEM_SMACKER
 CCFLAGS := $(CCFLAGS) -I../../thirdparty/libsmacker
-CCFLAGS := $(CCFLAGS) -I../../thirdparty/pcg-cpp/include
 endif
+
+CCFLAGS := $(CCFLAGS) -isystem ../../thirdparty/pcg-cpp/include
 
 SOURCEDIR := ../../engine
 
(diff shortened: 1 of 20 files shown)


usidedown:
All the warnings in clang-tidy come from pcg-cpp, so I believe it's okay.

src/dist/engine/Makefile
```diff
@@ -22,6 +22,8 @@ ifndef FHEROES2_WITH_SYSTEM_SMACKER
 CCFLAGS := $(CCFLAGS) -I../../thirdparty/libsmacker
 endif
 
+CCFLAGS := $(CCFLAGS) -isystem ../../thirdparty/pcg-cpp/include
```
oleg-derevenetz, line 25: This is somewhat alarming. Is everything really so bad there that you had to use `-isystem`? Also I took a closer look at this library and it looks like it is somewhat abandoned, the last commit was there about 3 years ago.

usidedown: We can probably use 
-I as most warnings had to be fixed for Android 
(And for consistency change cmake as well)
I try to avoid making changes in 3rd party libraries, warnings were not that bad. 

The library was maintained for 8 years, I don't think we'll find a better pcg implementation.

oleg-derevenetz: As far as I understand, the problem with reproducibility of random sequences on different platforms/compilers is not so much related to the implementation of the Mersenne Twister engine, but rather to differences in implementation of `std::uniform_int_distribution`, right? Can we limit ourselves to our own implementation of this distribution for now? As I understand it, this does not require the involvement of third-party libraries.

usidedown: Yes, pcg is used for performance reasons and not required for compatibility. 

We can implement distribution and shuffle(std variant uses uniform int distribution) ourselves.

We can also implement pcg. If only pcg32 is used it shouldn't be too difficult (and it shouldn't require int128)

oleg-derevenetz: > We can implement distribution and shuffle(std variant uses uniform int distribution) ourselves.

I suggest we limit ourselves to this for now (we can implement them within the scope of our existing `rand` module within `engine`) and then we'll see.

usidedown: Okay.
I'll close this PR and start a smaller one just with the relevant changes without all the clutter.

PR #9876, opened 2025-06-06
Android: use SHA512 hash instead of MD5 during the generateAssetsDigest build step to further minimize the likelihood of hash collisions

oleg-derevenetz:
The digest file is generated during the build, so the potential difference in hash function performance is not so important, and comparing digests during game startup will not suffer, because they are relatively small anyway.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -143,8 +143,8 @@ tasks.register('generateAssetsDigest') {
 
             final byte[] itBuf = Files.readAllBytes(itPath)
 
-            sb.append(String.format('%032x % 12d %s\n',
-                    new BigInteger(1, MessageDigest.getInstance('MD5').digest(itBuf)),
+            sb.append(String.format('%0128x % 12d %s\n',
+                    new BigInteger(1, MessageDigest.getInstance('SHA-512').digest(itBuf)),
                     itBuf.length, assetsDirPath.relativize(itPath)))
         }
 


usidedown:
I am curious as to what brought this up.
MD5sum has 128 bit, so a collision is expected after roughly 2<sup>64</sup> invocations.
If someone is maliciously generating hashes to cause a collision, it's a different story and md5sum can't be used.

oleg-derevenetz:
> MD5sum has 128 bit, so a collision is expected after roughly 2^64 invocations.

SHA512 has 512 bits, and the digest file size is still quite small and quick for compare. It's just that if there is an opportunity to further reduce the likelihood of a collision without any noticeable negative consequences, then why not do it? What are the disadvantages?

> If someone is maliciously generating hashes to cause a collision, it's a different story and md5sum can't be used.

It's not about that. We control both files and hash generation. Of course, if the user has generated something for himself, then this is their problem :)

usidedown:
I actually like this change, I'm just curious how it came up 😄 

> What are the disadvantages
Only disadvantages are the longer time to compute sha512 and the fact that it takes more space.
These are of course negligible unless a large number of computations is performed.

> It's not about that. We control both files and hash generation. Of course, if the user has generated something for himself, then this is their problem :)

IMHO the security argument is good. It protects users from malicious contributors.

I am not really sure what this hash covers anyway 😅

oleg-derevenetz:
> Only disadvantages are the longer time to compute sha512 and the fact that it takes more space.

These computations are only performed during the build process, not at runtime.

> I am not really sure what this hash covers anyway 😅

On Android, the auxiliary files (like `resurrection.h2d`, translations, TiMidity instruments samples, etc) are supplied as a part of an APK file as so-called "assets". These assets are stored inside the APK and are not available as "regular" files. In order for the engine to work with these assets as with regular files, we first need to "extract" them to the "real" Android filesystem. At the time, they were extracted at every game run, but this is somewhat redundant, because they only may change when the engine version is updated. Optimization has been introduced - the build system prepares a digest file with hashes, sizes, and file names, and asset files are updated only if this digest itself has changed (the digest from the APK and the digest stored on the file system after the previous asset extraction are compared).

usidedown:
Thank you so much for taking the time to explain 🙏 

> These computations are only performed during the build process, not at runtime.

As you previously stated, it would also impact the startup time when the comparisons are made (the digests need to be calculated over the existing files and compared)
The difference between md5sum and sha512sum is probably negligible. The IO involved is probably the bigger part.

oleg-derevenetz:
> As you previously stated, it would also impact the startup time when the comparisons are made (the digests need to be calculated over the existing files and compared)

No, it doesn't work this way. The comparison is performed between previous digests generated by build system (these digests are on the filesystem of the Android device) and new digests generated by build system (these digests are inside the APK). No digest calculation over actual files is performed during runtime. If we were doing this, it would be easier not to calculate any digests at all, but just always overwrite existing files.

oleg-derevenetz:
@usidedown The only potential problem that worries me here is that somehow both the hash and the size of the old version of the file will completely match the hash and size of the new version of the same file during the APK build in our repo. Therefore, I want to minimize the likelihood of this by reasonable means, hence this PR. The fact that the end user can tamper with the old digest and/or old files (e.g. manually adjusting them to the hash and size of the new file version) on _their own device_ (e.g. using ADB) doesn't bother me at all - it's their own device, and it won't affect anyone but them.

ihhub (approved):

ihhub:

PR #9879, opened 2025-06-07
Fix tiles initialization when creating a new map from scratch

Districh-ru:
This PR fixes tiles initialization when creating a new map. Previously only the top row was initialized.

It also renames the variable `size` to `width` because it actually is the width of the map which is the same as the map height because we support only squared maps.

### DIFF
--- a/src/fheroes2/editor/editor_interface.cpp
+++ b/src/fheroes2/editor/editor_interface.cpp
@@ -2015,7 +2015,7 @@ namespace Interface
 
         _mapFormat.width = mapWidth;
 
-        // The square maps are only supported so map height is the same as width.
+        // Only square maps are supported so map height is the same as width.
         const int32_t tilesCount = mapWidth * mapWidth;
 
         _mapFormat.tiles.resize( tilesCount );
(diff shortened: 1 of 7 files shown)


src/fheroes2/editor/editor_interface.cpp
```diff
@@ -2005,21 +2005,22 @@ namespace Interface
...
             assert( 0 );
 
             return false;
         }
 
-        _mapFormat.size = size;
+        _mapFormat.width = mapWidth;
 
-        const int32_t tilesCount = size * size;
+        // The square maps are only supported so map height is the same as width.
```
zenseii:
```suggestion
        // Only square maps are supported so map height is the same as width.
```

src/fheroes2/editor/editor_interface.cpp
```diff
@@ -2005,21 +2005,22 @@ namespace Interface
...
+        _mapFormat.width = mapWidth;
 
-        const int32_t tilesCount = size * size;
+        // Only square maps are supported so map height is the same as width.
+        const int32_t tilesCount = mapWidth * mapWidth;
 
         _mapFormat.tiles.resize( tilesCount );
 
-        for ( int32_t i = 0; i < size; ++i ) {
+        for ( int32_t i = 0; i < tilesCount; ++i ) {
```
ihhub, line 2023: Is it correct to say that this was the place where the original bug was introduced?

Districh-ru: Yes.
The upper code also shows that the `tilesCount` should be used.
My bad.

ihhub (approved):

ihhub:
@Districh-ru , many thanks for the quick fix of the problem!

PR #9884, opened 2025-06-08
Feature/reproducibility

usidedown:
Solves reproducibility issues raised in #9867 

- Replace uniform_int_distribution with custom implementation
- Replaces std::shuffle with custom implementation

Doesn't address the performance concerns raised in the issue above.

### DIFF
--- a/src/engine/rand.cpp
+++ b/src/engine/rand.cpp
@@ -27,7 +27,7 @@
 
 namespace
 {
-    // implementation of Fast Random Integer Generation in an Interval (https://arxiv.org/abs/1805.10941)
+    // Implementation of Fast Random Integer Generation in an Interval (https://arxiv.org/abs/1805.10941)
     // NOTE: we can't use std::uniform_int_distribution here because it behaves differently on different platforms
     uint32_t uniformIntInInterval( const uint32_t range, std::mt19937 & gen )
     {
@@ -71,7 +71,7 @@ uint32_t Rand::uniformIntDistribution( const uint32_t from, const uint32_t to, s
 
     assert( from < to );
     const uint32_t rangeExclusive = to - from;
-    // if the range is the entire uint32_t (from 0 to 2**32-1), we can just return a random number
+    // If the range is the entire uint32_t (from 0 to 2**32-1), we can just return a random number
     if ( rangeExclusive == std::numeric_limits<uint32_t>::max() ) {
         static_assert( std::mt19937::min() == 0 && std::mt19937::max() == std::numeric_limits<uint32_t>::max() );
 
--- a/src/engine/rand.cpp
+++ b/src/engine/rand.cpp
@@ -39,7 +39,7 @@ namespace
         assert( range > 0 );
 
         // Our implementation assumes our RNG can use the entire range of uint32_t
-        assert( gen.min() == 0 && gen.max() == std::numeric_limits<uint32_t>::max() );
+        static_assert( std::mt19937::min() == 0 && std::mt19937::max() == std::numeric_limits<uint32_t>::max() );
 
         uint32_t generated = gen();
         uint64_t mult = ( static_cast<uint64_t>( generated ) * range );
@@ -82,7 +82,7 @@ uint32_t Rand::uniformIntDistribution( const uint32_t from, const uint32_t to, s
     const uint32_t rangeExclusive = to - from;
     // if the range is the entire uint32_t (from 0 to 2**32-1), we can just return a random number
     if ( rangeExclusive == std::numeric_limits<uint32_t>::max() ) {
-        assert( gen.min() == 0 && gen.max() == std::numeric_limits<uint32_t>::max() );
+        static_assert( std::mt19937::min() == 0 && std::mt19937::max() == std::numeric_limits<uint32_t>::max() );
 
         return gen();
     }
--- a/src/engine/rand.cpp
+++ b/src/engine/rand.cpp
@@ -25,44 +25,39 @@
 
 #include <numeric>
 
-namespace
-{
-
 #pragma warning( push )
 #pragma warning( disable : 4146 ) // suppress warning C4146: unary minus operator applied to unsigned type, result still unsigned
 
-    // implementation of Fast Random Integer Generation in an Interval (https://arxiv.org/abs/1805.10941)
-    // NOTE: we can't use std::uniform_int_distribution here because it behaves differently on different platforms
-    uint32_t uniformIntDistribution( const uint32_t from, const uint32_t to, std::mt19937 & gen )
-    {
-        if ( from == to ) {
-            return from;
-        }
+// implementation of Fast Random Integer Generation in an Interval (https://arxiv.org/abs/1805.10941)
+// NOTE: we can't use std::uniform_int_distribution here because it behaves differently on different platforms
+uint32_t Rand::uniformIntDistribution( const uint32_t from, const uint32_t to, std::mt19937 & gen )
+{
+    if ( from == to ) {
+        return from;
+    }
 
-        const uint32_t range = to - from + 1;
+    const uint32_t range = to - from + 1;
 
-        uint32_t generated = gen();
-        uint64_t mult = ( static_cast<uint64_t>( generated ) * range );
-        uint32_t lowerPart = static_cast<uint32_t>( mult );
-        if ( lowerPart >= range ) {
-            const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
-            return from + upperPart;
-        }
-
-        // ( -range ) % range is the same as (2**32 - range) % range
-        const uint32_t discardBound = ( -range ) % range;
-
-        while ( lowerPart < discardBound ) {
-            generated = gen();
-            mult = ( static_cast<uint64_t>( generated ) * range );
-            lowerPart = static_cast<uint32_t>( mult );
-        }
+    uint32_t generated = gen();
+    uint64_t mult = ( static_cast<uint64_t>( generated ) * range );
+    uint32_t lowerPart = static_cast<uint32_t>( mult );
+    if ( lowerPart >= range ) {
         const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
         return from + upperPart;
     }
-#pragma warning( pop )
 
+    // ( -range ) % range is the same as (2**32 - range) % range
+    const uint32_t discardBound = ( -range ) % range;
+
+    while ( lowerPart < discardBound ) {
+        generated = gen();
+        mult = ( static_cast<uint64_t>( generated ) * range );
+        lowerPart = static_cast<uint32_t>( mult );
+    }
+    const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
+    return from + upperPart;
 }
+#pragma warning( pop )
 
 std::mt19937 & Rand::CurrentThreadRandomDevice()
 {
--- a/src/engine/rand.cpp
+++ b/src/engine/rand.cpp
@@ -47,7 +47,7 @@ namespace
 
         if ( lowerPart >= range ) {
             const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
-            assert( upperPart >= 0 && upperPart <= range );
+            assert( /* upperPart >= 0 && */ upperPart <= range );
 
             return upperPart;
         }
@@ -62,7 +62,7 @@ namespace
         }
 
         const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
-        assert( upperPart >= 0 && upperPart <= range );
+        assert( /* upperPart >= 0 && */ upperPart <= range );
 
         return upperPart;
     }
--- a/src/engine/rand.cpp
+++ b/src/engine/rand.cpp
@@ -47,7 +47,7 @@ namespace
             return upperPart;
         }
 
-        // This is the same as (2**32 - range) % range in the two’s complement representation
+        // This is the same as (2**32 - range) % range in the two's complement representation
         const uint32_t discardBound = static_cast<uint32_t>( ~( range - 1 ) ) % range;
 
         while ( lowerPart < discardBound ) {
--- a/src/engine/rand.cpp
+++ b/src/engine/rand.cpp
@@ -40,18 +40,25 @@ namespace
         }
 
         const uint32_t range = to - from + 1;
+
+        uint32_t generated = gen();
+        uint64_t mult = ( static_cast<uint64_t>( generated ) * range );
+        uint32_t lowerPart = static_cast<uint32_t>( mult );
+        if ( lowerPart >= range ) {
+            const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
+            return from + upperPart;
+        }
+
         // ( -range ) % range is the same as (2**32 - range) % range
         const uint32_t discardBound = ( -range ) % range;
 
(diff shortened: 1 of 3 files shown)


usidedown:
Not sure why IWYU suggests string, it's probably a false positive.

oleg-derevenetz:
> Not sure why IWYU suggests string, it's probably a false positive.

There are two options:

* Just include the suggested file (I usually do this, if this is a false positive, it can be fixed later in a new version of IWYU and it will suggest to remove this header eventually). There can be no any real harm.
* Use the `no_include` [pragma](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-no_include). In this case, if this header still needs to be included due to some future edits, then IWYU will not offer it anyway, which may turn into a problem (for example, on some platforms the build will be successful, and on some it will not). I would use this pragma only as a last resort.

src/engine/rand.cpp
```diff
@@ -25,6 +25,44 @@
 
 #include <numeric>
 
+#if defined( _WIN32 )
```
oleg-derevenetz: Whether these pragmas work or not depends not on the platform, but on the compiler. Someone can use gcc to build fheroes2 for win32, and the build will fail. I would use the `_MSC_VER` [instead](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170).

src/engine/rand.h
```diff
@@ -35,6 +35,29 @@
 
 namespace Rand
 {
+    uint32_t uniformIntDistribution( const uint32_t from, const uint32_t to, std::mt19937 & gen );
+
+    // Fisher-Yates shuffle AKA Knuth shuffle, probably the same as std::shuffle.
+    // NOTE: we can't use std::shuffle here because it uses std::uniform_int_distribution which behaves differently on different platforms.
+    template <class Iter, class Random>
+    void shuffle( Iter first, Iter last, Random && gen ) // NOLINT(cppcoreguidelines-missing-std-forward): same signature as std::shuffle
```
oleg-derevenetz: There is actually no need for `NOLINT`, you can just ignore the Clang-Tidy suggestions if you really know what you're doing. These are just optional hints.

src/engine/rand.h
```diff
@@ -35,6 +35,29 @@
...
+    {
+        if ( first == last ) {
+            return;
+        }
+
+        // change last from once-past-the-end to last element
+        --last;
+        uint32_t remainingSwaps = static_cast<uint32_t>( last - first );
+        while ( remainingSwaps > 1 ) {
+            using std::swap;
```
oleg-derevenetz, line 66: What is the purpose of this line?

src/engine/rand.h
```diff
@@ -35,6 +35,29 @@
...
+    template <class Iter, class Random>
+    void shuffle( Iter first, Iter last, Random && gen ) // NOLINT(cppcoreguidelines-missing-std-forward): same signature as std::shuffle
+    {
+        if ( first == last ) {
+            return;
+        }
+
+        // change last from once-past-the-end to last element
+        --last;
+        uint32_t remainingSwaps = static_cast<uint32_t>( last - first );
```
oleg-derevenetz: I would at least add an assertion that `last` is not less than `first`.

src/engine/rand.h
```diff
@@ -35,6 +35,29 @@
...
+        while ( remainingSwaps > 1 ) {
```
oleg-derevenetz: I suppose if a set to be shuffled consists of only two elements, then there is no chance that they will ever be swapped?

oleg-derevenetz (commented):
Hi @usidedown I just left a few questions and suggestions. Could you please take a look when you have time?

src/engine/rand.h
usidedown: I think we best keep it so it won't fail for the next time the file is modified.

src/engine/rand.h
usidedown: Added explanation inline. Let me know if it's still unclear.

oleg-derevenetz: I'm asking this because ADL [will not work](https://godbolt.org/z/Ybb3MPvos) if you then call `std::swap(...)`. You should call the unqualified version - i.e. just `swap(...)`.

usidedown: Could've sworn I used unqualified, thanks for the catch! 
I'll fix when I'm in front of my computer.

src/engine/rand.h
oleg-derevenetz: If we think about it, then this `shuffle()` is always called with `std::mt19937` as the third argument, so we can just turn its signature into

```suggestion
    void shuffle( Iter first, Iter last, std::mt19937 & gen )
```

and avoid this issue with universal reference and `std::forward` completely. We are not implementing libstdc++ after all and we are not obliged to replicate the signatures if we don't need this. What do you think?

usidedown: I like having the same interface to allow drop in replacement. One less line change when we change random engine 😉 
I don't see any disadvantage to keeping the forwarding reference as the code is very simple. 
That being said this is immaterial and I'm willing to accept the suggestion if you prefer it that way.

oleg-derevenetz: No, it's not necessary in general, it's just a way to get rid of unnecessary comments that silence warnings. In addition, it looks funny that a function template that accepts an argument of generic type then passes this argument to the function `uniformIntDistribution` that accepts an argument of only a specific type. If we change `std::mt19937` to something else in the future, then there will be quite a lot of such replacements, and I doubt that one more replacement will matter :) But this is optional.

src/engine/rand.cpp
```diff
@@ -25,6 +25,55 @@
...
+// NOTE: we can't use std::uniform_int_distribution here because it behaves differently on different platforms
+uint32_t Rand::uniformIntDistribution( const uint32_t from, const uint32_t to, std::mt19937 & gen )
+{
+    if ( from == to ) {
+        return from;
+    }
+
+    assert( from < to );
+
+    const uint32_t range = to - from + 1;
```
oleg-derevenetz: I suppose this won't work on the entire range (i.e. if `from == 0 && to == UINT32_MAX`), am I right? I'm not sure if this is really required somewhere in our code, but in a good way, a valid implementation should work on the entire range. If this is difficult to fix, then we can add an assertion for now.

src/engine/rand.h
```diff
@@ -29,12 +29,39 @@
...
+        --last;
+        uint32_t remainingSwaps = static_cast<uint32_t>( last - first );
+        while ( remainingSwaps > 0 ) {
+            // Allow argument-dependant lookup (ADL) for swap: first try in the namespace of the type, then in the std namespace.
+            using std::swap;
+            const uint32_t index = uniformIntDistribution( 0, remainingSwaps, gen );
+            swap( *last, *( first + index ) );
+            --last;
+            --remainingSwaps;
+        }
```
oleg-derevenetz: How about something like this:

```suggestion
        // last points to the past-the-end element that should not participate in the shuffling process
        for ( auto i = last - first - 1; i > 0; --i ) {
            using std::swap;
            swap( *( first + i ), *( first + uniformIntDistribution( 0, static_cast<uint32_t>( i ), gen ) ) );
        }
```

The same idea, but looks a bit shorter (and allows to declare `first` and `last` as `const`).

oleg-derevenetz (commented):
Hi @usidedown I added a few assertions just in case and left a couple of questions and suggestions, would you mind to take a look at your spare time?

src/engine/rand.h
usidedown: Looks less clear to me. Having things on separate lines with comments was easier to read IMHO.
I'll change if you find the new implementation more understandable.

oleg-derevenetz: As for me, this variant has the following objective advantages:

1. There is no need to handle the special case when `last == first`;
2. There is no need to additionally check that `last` is not less than `first`;
3. Both `first` and `last` can be `const`.

And there is also a subjective advantage that this variant is much shorter. As for clarity, verbosity is not always equivalent to clarity. However, if you prefer your variant, then I do not insist.

usidedown: >There is no need to handle the special case when last == first;
There is no need to additionally check that last is not less than first;

All true, but come at the cost of using int64 arithmetic rather than uint32_t. Probably doesn't matter on 64-bit CPU though ¯\\\_(ツ)\_/¯

> Both first and last can be const
I actually like const 😄 . Didn't seem to matter much as arguments are passed by value anyway.

I think clarity and readability is the most important consideration here.

I'll change if you think the for loop is clearer.

src/engine/rand.cpp
```diff
@@ -25,6 +25,62 @@
...
+
+        if ( lowerPart >= range ) {
+            const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
+            assert( /* upperPart >= 0 && */ upperPart <= range );
+
+            return upperPart;
+        }
+
+        // This is the same as (2**32 - range) % range in the two's complement representation
+        const uint32_t discardBound = static_cast<uint32_t>( ~( range - 1 ) ) % range;
```
oleg-derevenetz, line 51: I should note that both `-range % range` and `~(range - 1) % range` variants are in general [vulnerable to integer promotion-related issues](https://godbolt.org/z/oqhPcETTT), so I decided to emphasize this by adding an explicit `static_cast`. Also I decided to emphasize in the comment that all this will work only on two's complement systems (consider an imaginary one's complement system with 64-bit `int`... fortunately, such general-purpose systems are now almost never encountered).

usidedown: Thank you so much for your help and fixes!
A couple of notes:

1. I find ~ ( range -1 ) to be less clear than ( -range ). This is very minor 😅 .
2. Didn't know integer promotion kicks in for unary operators, cool!
3. I find the two's complement comment confusing. range is unsigned. unsigned arithmetic is well defined to be modulo 2**32. This is guaranteed by the standard and works for all platforms.
You are of course correct about signed arithmetic, but that shouldn't apply in this case.

oleg-derevenetz: > You are of course correct about signed arithmetic, but that shouldn't apply in this case.

For illustrative purposes, consider some imaginary platform where `uint32_t` is 4 bit in size and `int64_t` (AKA `int` on this imaginary platform) is 8 bit-one's-complement :) In this case `range` will be promoted to the signed `int` value. Suppose that `range` is 0. In this case, `range - 1` will become `11111110` (`-1` in 8 bit-one's-complement representation), and `~( range - 1 )` will become 1, right? And this is not what the result of the expression `2^4 - 0` should be.

usidedown: What about going back to -range (with the cast)
```c++
const uint32_t discardBound = static_cast<uint32_t>( -range  ) % range;
```
Shouldn't that work even if range is promoted?

oleg-derevenetz: > Shouldn't that work even if range is promoted?

Unfortunately or fortunately, modern general-purpose one's complement architectures are practically non-existent anymore and are unlikely to come back to life in the future. At least on the machines where fheroes2 will be running this is for sure :) Since C++20, the two's complement is the only permitted representation for signed integers (see [P1236R1](https://open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2131r0.html)).

src/engine/rand.cpp
```diff
@@ -25,6 +25,62 @@
...
+        // Our implementation assumes our RNG can use the entire range of uint32_t
+        static_assert( std::mt19937::min() == 0 && std::mt19937::max() == std::numeric_limits<uint32_t>::max() );
+
+        uint32_t generated = gen();
+        uint64_t mult = ( static_cast<uint64_t>( generated ) * range );
+        uint32_t lowerPart = static_cast<uint32_t>( mult );
+
+        if ( lowerPart >= range ) {
+            const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
+            assert( /* upperPart >= 0 && */ upperPart <= range );
```
ihhub: Please the commented out code.

src/engine/rand.cpp
```diff
@@ -25,6 +25,62 @@
 
 #include <numeric>
 
+namespace
+{
+    // implementation of Fast Random Integer Generation in an Interval (https://arxiv.org/abs/1805.10941)
```
ihhub: Since all other comments start from a capital letter please do the same here.

src/engine/rand.cpp
```diff
@@ -25,6 +25,62 @@
...
+        const uint32_t discardBound = static_cast<uint32_t>( ~( range - 1 ) ) % range;
+
+        while ( lowerPart < discardBound ) {
+            generated = gen();
+            mult = ( static_cast<uint64_t>( generated ) * range );
+            lowerPart = static_cast<uint32_t>( mult );
+        }
+
+        const uint32_t upperPart = static_cast<uint32_t>( mult >> 32 );
+        assert( /* upperPart >= 0 && */ upperPart <= range );
```
ihhub: Please remove the commented out code.

src/engine/rand.cpp
```diff
@@ -25,6 +25,62 @@
...
+
+uint32_t Rand::uniformIntDistribution( const uint32_t from, const uint32_t to, std::mt19937 & gen )
+{
+    if ( from == to ) {
+        return from;
+    }
+
+    assert( from < to );
+    const uint32_t rangeExclusive = to - from;
+    // if the range is the entire uint32_t (from 0 to 2**32-1), we can just return a random number
```
ihhub: Since all other comments start from a capital letter please do the same here.

src/engine/rand.h
ihhub: Let's not do a very generic implementation for the sake of a potential future which usually never comes. As @oleg-derevenetz suggested we should use `std::mt19937` over `Random` and remove linter specific comments.

src/engine/rand.h
```diff
@@ -27,14 +27,48 @@
 #include <cassert>
 #include <cstdint>
 #include <cstdlib>
+// IWYU pragma: no_include <exception>
```
ihhub, line 32: What's the problem if we include `<exception>` here? We should always try not to add comments specific to some tools. The source code should remain self explanatory without the need of understanding of magical comments.

src/engine/rand.h
```diff
@@ -27,14 +27,48 @@
...
+    template <class Iter, class Random>
+    void shuffle( Iter first, Iter last, Random && gen ) // NOLINT(cppcoreguidelines-missing-std-forward): same signature as std::shuffle
+    {
+        if ( first == last ) {
+            return;
+        }
+
+        assert( first < last );
+
+        // change last from once-past-the-end to last element
```
ihhub: Since all other comments start from a capital letter please do the same here.

ihhub (changes requested):
Hi @usidedown , I left several comments here. Can you please check them?

src/engine/rand.h
oleg-derevenetz: Hi @ihhub the problem in this very case is that if we include `<exception>`, then IWYU will ask to remove it and include `<string>` instead, then vice versa and so on ad infinitum. That's an issue in IWYU. Please see [my comment](https://github.com/ihhub/fheroes2/pull/9884#pullrequestreview-2916756827 ) for details.

ihhub: Thanks for the explanation!

ihhub (approved):

ihhub:

